PR #2458, opened 2020-03-22
make taskkill argument process same as what windows do

kernelbin:
## Purpose

make the argument process of cmd cmdutil "taskkill" behave same as what windows do.

now options are nolonger detected as parameters when placed after options that accept one.
like ```taskkill /im /f``` will nolonger regard ```/f``` as a process name

now ```/?``` and ```/f``` appear more than once will report not allowed more than once
 like ```taskkill /f /pid 1000 /f```

now only one option ```/f``` without ```/pid``` and ```/im``` will report must specify ```/pid``` or ```/im```.
like ```taskkill /f```

I failed to find an issue match this on JIRA.

## Proposed changes

fix the problem that the argument process of cmd cmdutil "taskkill" behave differs from what windows do.

## TODO

support argument ```/t``` and  argument ```/fi```
#### also, translation needed in .rc files
I just translate it into Chinese and English, other language's translation needed


binarymaster:
You can `git push --force-with-lease` instead of closing and opening new PR ;)

kernelbin:
> You can `git push --force-with-lease` instead of closing and opening new PR ;)

got it. I will use this next time.😅(In fact I'm not quite familiar with git)

PR #1872, opened 2019-08-23
Implement versioning system for DLLs

tkreuzer:
## Purpose

This PR implement a system that allows to make DLL exports version specific at runtime, based on the applications compatibility setting. It still allows ReactOS DLLs to statically import functions from higher versions, even if they are hidden to the application.

JIRA issue: [CORE-11288](https://jira.reactos.org/browse/CORE-11288)

## Proposed changes

- Add a feature to spec2def to generate a RosCompat descriptor (contained in a special section) pointing to a mask-array, that specifies for each named export on which Windows versions it is present.
- Add a feature to the loader in NTDLL that checks for a RosCompat descriptor, and if present, uses it to patch the export table, splitting it into a public and private part.
- Modify the loader to allow ReactOS DLLs (those that contain a RosCompat descriptor, even if it is empty) to import from private/hidden exports.

## TODO
- [ ] Allow to disable this feature with a CMake switch to allow building DLLs that are Windows 2003 compatible.
- [ ] Code cleanup (move some code to a shared library, fix TODOs)
- [ ] Implement tests for different Windows versions
- [ ] Run on testbot

PR #2461, opened 2020-03-22
[FONTS] Add Selawik as a substitution for Segoe UI

wjk:
## Purpose

Many applications written for Windows Vista and later hardcode the font name “Segoe UI” into their binaries. We cannot distribute Segoe UI ourselves due to license issues. However, there is a metric-compatible font called Selawik that is freely licensed. Therefore, this PR adds it to this repo, and also inserts a FontSubstitution Registry entry so that apps that reference Segoe UI by name will use Selawik transparently instead.

## TODO

- [ ] I have only substituted the regular font, not Bold or Italic, because I don't know how to properly substitute alternate styles. If someone could please let me know how, I'll add the other styles.
- [ ] Since I only speak English, I have not edited any of the other languages’ font substitution tables in `muifonts.h`. If anyone could tell me what (if anything) is needed here, I’ll gladly add it.
- [ ] If we want to hold this PR until we are ready to start adding Vista+ APIs into ReactOS core, for example until after #1872 is merged, that's fine with me. I just wanted to get this code into my fork now.


gonzoMD:
@katahiromz didn't you want to get rid of all those substitutions?

wjk:
Unfortunately, I don't know of any other good way to do this. If we do nothing, then apps that hardcode Segoe UI _will_ break. (This includes the MSI internal UI, as it has no way to dynamically determine the theme font, nor to resize controls to match it.) If I add a special-case in win32ss somewhere, then (a) IMO that seems excessive when we already have a mechanism right here and (b) if the user does install the actual Segoe UI font (either from a license-violating GitHub repo or by installing a Microsoft program that bundles it), then the correct font will not be used. With this solution, such a user can simply delete the substitution with regedit and then reboot. If anyone has any other ideas, I'd love to hear them. Thanks!

Mouvedia:
What you want is [Selawik](https://github.com/Microsoft/Selawik) which is the open source replacement of Segoe UI.

wjk:
@Mouvedia That's actually a very good idea. While the repo you linked is abandonware, and ugly abandonware at that, I have found [this fork of Selawik](https://github.com/vv-monsalve/Selawik/tree/QA-Selawik) that actually looks quite good at the point sizes used in ReactOS. (While it does look very different from Segoe UI when enlarged, at 9pt and 12pt these differences are hard to perceive.) And it is guaranteed metric compatible with Segoe UI.

I will add the Selawik font files and license to this PR. However, for full compatibility with Windows, we will still need the font substitution entry, or apps that hardcode Segoe UI would still break. Thanks so much!

PR #2462, opened 2020-03-23
Introduce configuration for gitpod

zefklop:
## Purpose

I played a bit with gitpod and their online IDE (theia) and I have been pleased by the experience. The PR is about having the possibility to develop and debug ReactOS (in qemu) with a one-click solution.

The proposed configuration creates a custom docker image, and has the tasks configured to build, launch & debug both bootcd & livecd (using GDB).

I fixed a few hitches while I were at it, mostly in the build system and in KDGDB.

## Proposed changes

Nothing much, those are a bunch of config files, but you can have a look there :
https://gitpod.io#snapshot/4b28a6aa-3a64-4a6e-8bdb-5a7c190b2e10
If you want to have nice C/C++ integration, you have to choose the build configuration in the "C/C++ build config" button at the bottom right of the screen.
NB : in the "open ports" tab, you can open the browser for port 6080. There you will have a VNC display pointing to an openbox desktop. This is where qemu will show the instance being debugged.

### DIFF
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -1,0 +1,23 @@
+FROM gitpod/workspace-full-vnc
+                    
+USER gitpod
+
+# Install custom tools, runtime, etc. using apt-get
+# For example, the command below would install "bastet" - a command line tetris clone:
+#
+# RUN sudo apt-get -q update && #     sudo apt-get install -yq bastet && #     sudo rm -rf /var/lib/apt/lists/*
+#
+# More information: https://www.gitpod.io/docs/config-docker/
+RUN sudo apt-get -q update && \
+    sudo apt-get -yq upgrade && \
+    sudo apt-get -yq install qemu-system-x86 qemu-utils gdb-mingw-w64 && \
+    sudo rm -rf /var/lib/apt/lists/*
+
+RUN wget https://svn.reactos.org/amine/RosBEBinFull.tar.gz && \
+    sudo tar -xzf RosBEBinFull.tar.gz -C /usr/local && \
+    sudo mv /usr/local/RosBEBinFull /usr/local/RosBE && \
+    rm -f RosBEBinFull.tar.gz
+
+RUN echo 'export PATH=/usr/local/RosBE/i386/bin:$PATH' >> /home/gitpod/.profile
+
+RUN qemu-img create -f qcow2 reactos_hdd.qcow 10G
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,18 @@
+tasks:
+  - before: > 
+      brew install cmake ninja
+    init: >
+      mkdir -p /workspace/reactos/build &&
+      cd /workspace/reactos/build &&
+      cmake .. -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DGDB:BOOL=TRUE -DSEPARATE_DBG:BOOL=TRUE -D_WINKD_:BOOL=TRUE -DKDBG:BOOL=FALSE -DENABLE_ROSTESTS:BOOL=TRUE -DCMAKE_TOOLCHAIN_FILE=toolchain-gcc.cmake -G "Ninja" &&
+      ninja xdk psdk 
+    command: >
+      cd /workspace/reactos/build &&
+      ninja all
+
+image:
+  file: .gitpod.Dockerfile
+
+vscode:
+  extensions:
+    - ms-vscode.cpptools@0.27.0-insiders3:Djj3Csw0GXjmueWAPWvTsg==
--- a/.theia/launch.json
+++ b/.theia/launch.json
@@ -1,0 +1,29 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "type": "cppdbg",
+      "request": "launch",
+      "name": "livecd (qemu)",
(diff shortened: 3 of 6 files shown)


drivers/base/kdgdb/kdpacket.c
```diff
@@ -379,14 +389,17 @@ KdReceivePacket(
     {
         DBGKD_MANIPULATE_STATE64* State = (DBGKD_MANIPULATE_STATE64*)MessageHeader->Buffer;
 
+        KDDBGPRINT("State manipulation: ");
```
HBelusca: Not sure this dprint is interesting, because we always get a corresponding print just afterwards: when the if-block is taken, and when it is not.

drivers/base/kdgdb/kdgdb.h
```diff
@@ -19,7 +19,7 @@
 
 #include <pstypes.h>
 
-// #define KDDEBUG /* uncomment to enable debugging this dll */
+#define KDDEBUG /* uncomment to enable debugging this dll */
```
HBelusca: Maybe comment this out again when finalizing the PR?

drivers/base/kdgdb/kdcom.c
```diff
@@ -131,7 +131,12 @@ KdpPortInitialize(IN ULONG ComPortNumber,
     }
     else
     {
+        UCHAR Dummy;
+
         KdComPortInUse = KdComPort.Address;
+        /* Consume what could already be there */
+        while(CpGetByte(&KdComPort, &Dummy, FALSE, FALSE) == CP_GET_SUCCESS)
```
HBelusca: `while<space>`

boot/CMakeLists.txt
```diff
@@ -15,11 +15,12 @@ else()
     message(FATAL_ERROR "Unknown ARCH '" ${ARCH} "', cannot generate a valid UEFI boot filename.")
 endif()
 
-add_custom_target(efisys
```
HBelusca: I suppose these modifications fix CMake dependencies?

drivers/base/kdgdb/kdpacket.c
zefklop: It's a contextualisation dprint. If you look closely, there is no trailing new line here. It helped me to understand an issue I had, so I keep it there.

drivers/base/kdgdb/kdgdb.h
zefklop: Sure

drivers/base/kdgdb/kdcom.c
zefklop: Sure

boot/CMakeLists.txt
zefklop: Not really. A custom target with a command is always considered out of date. Using the custom_command / custom_target couple makes the target up-to-date as long as its dependencies are up-to-date.

This avoids triggering mkisofs command each time you want to build the bootcd while it is already up-to-date.

CMakeLists.txt
```diff
@@ -249,7 +249,11 @@ else()
     endif()
 
     # For MSVC builds, this puts all debug symbols file in the same directory.
-    set(CMAKE_PDB_OUTPUT_DIRECTORY "${REACTOS_BINARY_DIR}/msvc_pdb")
+    if (MSVC)
+        set(CMAKE_PDB_OUTPUT_DIRECTORY "${REACTOS_BINARY_DIR}/msvc_pdb")
+    elseif(SEPARATE_DBG)
```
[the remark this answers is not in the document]
zefklop: It must be defined on cmake invocation. It's used when you want to produce separate debug symbols. I think I might be the only one to use it.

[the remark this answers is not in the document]
HBelusca: When you first invoke configure.cmd/sh it first tries to find the available compiler it's going to use, before creating the directories etc... so that's why I was thinking there may exist a cmake variable that gets initialized / from which one can derive a new cmake variable with a shortened version of the compiler name (msvc/gcc/clang/...), that we then can use here to create the name of the symbols directory.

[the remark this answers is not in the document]
zefklop: Ah OK. I don't think suh a variable exists, I might be wrong though.

drivers/base/kdgdb/kdcom.c
zefklop: In fact, this is a leftover which can be deleted

CMakeLists.txt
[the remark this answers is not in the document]
Extravert-ir: I think the separation should be done by symbols format. Something like `symbols_pdb` and `symbols_dwarf`.
Does `CMAKE_PDB_OUTPUT_DIRECTORY` called like that for all compilers?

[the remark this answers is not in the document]
Extravert-ir: @HBelusca we define such variables ourselves https://github.com/reactos/reactos/blob/2928084e2f86f1b2b39214dfaf872e76c0561650/sdk/cmake/config.cmake#L39-L50

zefklop:
The cmake changes are now the object of another PR, and the rest was directly committed to master.

CMakeLists.txt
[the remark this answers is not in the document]
zefklop: Out of this PR

drivers/base/kdgdb/kdpacket.c
zefklop: Out of this PR

zefklop:
If nobody disagrees, I am willing to merge this in by tomorrow.

PR #2463, opened 2020-03-24
[WINETESTS] Remove forced _WIN32_WINNT defines

SergeGautherie:
Only comctl32 is affected.

Import
https://source.winehq.org/git/wine.git/commit/be3cda6ec0856d143d4737ee9306844569b76e15

### DIFF
--- a/modules/rostests/winetests/comctl32/subclass.c
+++ b/modules/rostests/winetests/comctl32/subclass.c
@@ -17,8 +17,6 @@
  * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
  */
 
-#define _WIN32_WINNT 0x0501 /* For SetWindowSubclass/etc */
-
 #include <assert.h>
 #include <stdarg.h>
 
--- a/modules/rostests/winetests/kernel32/sync.c
+++ b/modules/rostests/winetests/kernel32/sync.c
@@ -18,9 +18,6 @@
  * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
  */
 
-#ifndef __REACTOS__
-#define _WIN32_WINNT 0x500
-#endif
 #include <stdarg.h>
 #include <stdlib.h>
 #include <stdio.h>
--- a/modules/rostests/winetests/kernel32/thread.c
+++ b/modules/rostests/winetests/kernel32/thread.c
@@ -18,11 +18,6 @@
  * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
  */
 
-/* Define _WIN32_WINNT to get SetThreadIdealProcessor on Windows */
-#ifndef __REACTOS__
-#define _WIN32_WINNT 0x0600
-#endif
-
 #include <assert.h>
 #include <stdarg.h>
 #include <stdio.h>
--- a/modules/rostests/winetests/kernel32/timer.c
+++ b/modules/rostests/winetests/kernel32/timer.c
@@ -18,10 +18,6 @@
  * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
  */
 
-#ifndef __REACTOS__
-#define _WIN32_WINNT 0x0501
-#endif
-
 #include "wine/test.h"
 #include "winbase.h"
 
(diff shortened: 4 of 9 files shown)


modules/rostests/winetests/comctl32/subclass.c
```diff
@@ -17,7 +17,10 @@
  * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
  */
 
+#ifdef __REACTOS__
+/* TODO: Wine remnant. Upgrade to 0x0502/0x0600. */
```
learn-more: Is this a note for yourself or are others supposed to understand it?

SergeGautherie: Both. Feel free to suggest a better comment, if need be.

Extravert-ir:
Can we just merge the changes with a next winesync?

SergeGautherie:
> next winesync?

I guess that means unknown delay before I might continue to work on this define matter. Choice is yours...

modules/rostests/winetests/comctl32/subclass.c
ThFabba: Mark's point here was, and I agree, that this doesn't adequately explain what's wrong here or how to fix it.
The ifdef indicates that Wine already removed this define here. But the comment suggests using 0x0502 instead, which is strange because that's our default. So why can't this simply be removed?

learn-more: Yeah sorry, I should try to elaborate more instead of throwing in half a sentence myself :)

SergeGautherie: > Mark's point here was, and I agree, that this doesn't adequately explain what's wrong here or how to fix it.

(Already answered.)

> The ifdef indicates that Wine already removed this define here.

Well, ifdef itself indicates it's ROS specific, not whether the define came from Wine or was added by ROS.

> But the comment suggests using 0x0502 instead, which is strange because that's our default. 

That is not obvious here:
doesn't
`remove_definitions(-D_WIN32_WINNT=0x502 -D_WIN32_IE=0x600)`
make default be `0x0600`?

> why can't this simply be removed?

Of course, if you prefer that change to be made right away, that's even better :-)

ThFabba: > > But the comment suggests using 0x0502 instead, which is strange because that's our default.
> 
> That is not obvious here:
> doesn't
> `remove_definitions(-D_WIN32_WINNT=0x502 -D_WIN32_IE=0x600)`
> make default be `0x0600`?

That's not the default, that's an override. And it's also the answer to my question (and essentially what the comment should say): the CMakeLists.txt for this module removes our default, and that's why we still need this here.


> > why can't this simply be removed?
> 
> Of course, if you prefer that change to be made right away, that's even better :-)

If whatever the CMake was fixing is now fixed in the source, then yes, we should remove that line from CMakeLists.txt rather than work around it in the code.
This is a "if it builds, it's correct" type of situation, so it should hopefully be straightforward to answer.

SergeGautherie: > That's not the default, that's an override.

I was thinking `#define` is the main override here. Anyway.

> And it's also the answer to my question (and essentially what the comment should say): the CMakeLists.txt for this module removes our default, and that's why we still need this here.

Well, I feel like I get "too long" complains when I add long comments and "too short" complains on short comments: I am sorry I don't write comments "right" on first try.

> If whatever the CMake was fixing is now fixed in the source, then yes, we should remove that line from CMakeLists.txt rather than work around it in the code.

As written on a previous PR, I'll check CMakeLists.txt files separately/later:
first, I am checking the `#define` overrides..

> This is a "if it builds, it's correct" type of situation, so it should hopefully be straightforward to answer.

Then, direct port (= upgrade to 0x0600) is correct, as far I can tell.

ThFabba: Yes, your definition of straightforward certainly often seems to be very different from mine.

Dealing with CMake in a separate PR can definitely make sense. But if you have to add a comment, only for the next PR to delete that comment and fix the issue properly, that seems overcomplicated to me.

Anyway, what you have now -- removing stuff that was removed by Wine, and things still build as they should -- looks like a straightforward improvement, so thank you. If the CMake fix is similarly simple, this is certainly a win.

SergeGautherie: > overcomplicated

I thought it may ease this/similar PR: bad idea :-<

PR #2465, opened 2020-03-25
[VSSAPI] Import Wine commits "Stub CreateVssBackupComponentsInternal + ?CreateVssBackupComponents@@YGJPAPAVIVssBackupComponents@@@z" CORE-9511

oleg-dubinskiy:
## Purpose

Import the two following Wine commits in vssapi:
https://source.winehq.org/git/wine.git/commit/ce9b5a7e1782828f0a84d31eff73cf643ea4fe85 vssapi: Add CreateVssBackupComponentsInternal stub.
https://source.winehq.org/git/wine.git/commit/767c94bc5c0b9cdef0465b0fec555d846331fb03
vssapi: Add ?CreateVssBackupComponents@@YGJPAPAVIVssBackupComponents@@@Z stub (depends on the 1st one).
Also import some headers and definitions from Wine, which are needed by these stubs.
Required by Recuva 1.53 to start successfully. Although now it still crashes anyway, but it at least starts at all, and its (and our) debugger doesn't interfere with the program's work. It is possible simply minimize them and continue to use the program.
P. S.: As I see, our vssapi is already synced to Wine 4.18. This version already contains that stubs, but our one for some reason doesn't have them. Maybe @AmineKhaldi missed something during the last winesync... :thinking: 

JIRA issue: [CORE-9511](https://jira.reactos.org/browse/CORE-9511)

## Proposed changes

- Replace our vswriter.h by Wine's vswriter.idl, since the first one does not contain all required definitions and Wine's vsbackup.idl (which is needed too) depends on it (has it hardcoded).
- Add some missing definitions into vss.idl from Wine's one, they are required by vsbackup.idl header.
- Import vsbackup.idl from Wine, it's required by the stubs.
- Import the commits itself with the stubs.

## Result

Before:
Recuva_before

After:
Recuva_after

Works (finds and recovers deleted files) as well:
Recuva_working


HBelusca:
I don't know why, but all the official MS PSDK do not have vswriter.idl, but only vswriter.h and according to its header it hasn't been autogenerated by any IDL tool (contrary to e.g. vss.h that is generated from the (also included file) vss.idl).
Therefore I have no idea why Wine thought it was great to break PSDK compatibility there.

HBelusca:
Please fix the compilation error:
```
[00:03:32] vssapi.exp : error LNK2001: unresolved external symbol "long __stdcall CreateVssBackupComponents(class IVssBackupComponents * *)" (?CreateVssBackupComponents@@YGJPAPAVIVssBackupComponents@@@Z)
[00:03:32] dll\win32\vssapi\vssapi.dll : fatal error LNK1120: 1 unresolved externals
[00:03:32] LINK failed. with 1120
```

oleg-dubinskiy:
> I don't know why, but all the official MS PSDK do not have vswriter.idl, but only vswriter.h and according to its header it hasn't been autogenerated by any IDL tool (contrary to e.g. vss.h that is generated from the (also included file) vss.idl).
Therefore I have no idea why Wine thought it was great to break PSDK compatibility there.
Also the vswriter.h can (and does, in MS PSDK) contain C++ definitions too.

Thanks, didn't know about that.
So... what I need to do? Restore vswriter.h, add all missing definitions there and ROS-specific diff in vsbackup.idl? Or keep it as is? :thinking: 
> Please fix the compilation error:

Hmm, for me it compiles successfully as is, even the whole bootcd. Maybe it's MSVC-specific error then.

learn-more:
You might wanna add a wrapper in `msvc.s`

learn-more:
yeah works fine.
`DEFINE_THISCALL_WRAPPER ?CreateVssBackupComponents@@YGJPAPAVIVssBackupComponents@@@Z, _VSSAPI_CreateVssBackupComponents@4`

dll/win32/vssapi/main.c
```diff
@@ -94,3 +95,19 @@ HRESULT __thiscall VSSAPI_CVssWriter_Unsubscribe( struct CVssWriter *writer )
     return S_OK;
 }
 DEFINE_THISCALL_WRAPPER( VSSAPI_CVssWriter_Unsubscribe, 4 )
```
learn-more: you probably need those too for the 2 new functions?

oleg-dubinskiy:
Committed. Thanks @learn-more. :+1: 
Now MSVC compilation should succeed.

learn-more:
Sorry, I did not pay enough attention, that is still wrong :(
I'll commit a fix for this on your branch.

PR #2466, opened 2020-03-25
[REACTOS] Fix '\NLS\' typos

SergeGautherie:
Use `\Nls\`, as everywhere else.

NB:
1 case remains in `dll/win32/kernel32/winnls/string/lang.c`: to be WineSync'ed...

### DIFF
--- a/base/setup/lib/mui.c
+++ b/base/setup/lib/mui.c
@@ -403,7 +403,7 @@ AddCodepageToRegistry(
 
     // Open the nls codepage key
     RtlInitUnicodeString(&KeyName,
-                         L"SYSTEM\\CurrentControlSet\\Control\\NLS\\CodePage");
+                         L"SYSTEM\\CurrentControlSet\\Control\\Nls\\CodePage");
     InitializeObjectAttributes(&ObjectAttributes,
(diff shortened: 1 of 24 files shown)


ThFabba:
Registry keys are case-insensitive, so this is a non-issue.
[I'm not going to close this since it's not wrong, but I'm also not willing to spend time on this.]

SergeGautherie:
> Registry keys are case-insensitive, so this is a non-issue.

Yeah, nonetheless using same (expected) case everywhere improves consistency, and makes searching much simpler, for example.

> [I'm not going to close this since it's not wrong, but I'm also not willing to spend time on this.]

(This PR is similar to a1c13a964f56626dd704cd07c63d03699b6f5043, for example.)

learn-more:
It is not really similar,
that PR changes strings that are next to each other for consistency,
this PR change strings all over the codebase.

SergeGautherie:
> It is not really similar

Well, it probably depends on which similarity/difference one looks at :-|
I didn't expect this trivial search and replace to create a debate :-(

ThFabba:
Enough bikeshedding

PR #2471, opened 2020-03-26
[SDK][LIB] Add structures for the arbiter.

vgalnt:
The main purpose of this PR is to add a definition of the structures for the HW resource arbiter.

In addition to this, the IopInitializeArbiters() function was implemented in the PNP manager and stub was added for ArbInitializeArbiterInstance().

The directory "XXX" is chosen because the arbiter can be used not only by the kernel, but also by pci and acpi bus drivers, as well as legacy HAL (PIC).

### DIFF
--- a/ntoskrnl/io/pnpmgr/arbs.c
+++ b/ntoskrnl/io/pnpmgr/arbs.c
@@ -1,0 +1,123 @@
+/*
+ * PROJECT:         ReactOS Kernel
+ * COPYRIGHT:       GPL-2.0+ (https://spdx.org/licenses/GPL-2.0+)
+ * FILE:            ntoskrnl/io/pnpmgr/arbs.c
+ * PURPOSE:         Root arbiters of the PnP manager
+ * PROGRAMMERS:     Copyright 2020 Vadim Galyant <vgal@rambler.ru>
+ */
+
+/* INCLUDES ******************************************************************/
+
+#include <ntoskrnl.h>
+
+#define NDEBUG
+#include <debug.h>
+
+/* GLOBALS *******************************************************************/
+
+extern ARBITER_INSTANCE IopRootBusNumberArbiter;
+extern ARBITER_INSTANCE IopRootIrqArbiter;
+extern ARBITER_INSTANCE IopRootDmaArbiter;
+extern ARBITER_INSTANCE IopRootMemArbiter;
+extern ARBITER_INSTANCE IopRootPortArbiter;
+
+/* DATA **********************************************************************/
+
+/* FUNCTIONS *****************************************************************/
+
+/* BusNumber arbiter */
+
+NTSTATUS
+NTAPI
+IopBusNumberInitialize(VOID)
+{
+    NTSTATUS Status;
+
+    DPRINT("IopRootBusNumberArbiter %p\n", &IopRootBusNumberArbiter);
+
+    Status = ArbInitializeArbiterInstance(&IopRootBusNumberArbiter,
+                                          NULL,
+                                          CmResourceTypeBusNumber,
+                                          L"RootBusNumber",
+                                          L"Root",
+                                          NULL);
+    return Status;
+}
+
+/* Irq arbiter */
+
+NTSTATUS
+NTAPI
+IopIrqInitialize(VOID)
+{
+    NTSTATUS Status;
+
+    DPRINT("IopRootIrqArbiter %p\n", &IopRootIrqArbiter);
+
+    Status = ArbInitializeArbiterInstance(&IopRootIrqArbiter,
+                                          NULL,
+                                          CmResourceTypeInterrupt,
+                                          L"RootIRQ",
+                                          L"Root",
+                                          NULL);
+    return Status;
+}
+
+/* Dma arbiter */
+
+NTSTATUS
+NTAPI
+IopDmaInitialize(VOID)
+{
+    NTSTATUS Status;
+
+    DPRINT("IopRootDmaArbiter %p\n", &IopRootDmaArbiter);
+
+    Status = ArbInitializeArbiterInstance(&IopRootDmaArbiter,
+                                          NULL,
+                                          CmResourceTypeDma,
+                                          L"RootDMA",
+                                          L"Root",
+                                          NULL);
+    return Status;
+}
+
+/* Memory arbiter */
+
+NTSTATUS
+NTAPI
+IopMemInitialize(VOID)
+{
+    NTSTATUS Status;
+
+    DPRINT("IopRootMemArbiter %p\n", &IopRootMemArbiter);
+
+    Status = ArbInitializeArbiterInstance(&IopRootMemArbiter,
+                                          NULL,
+                                          CmResourceTypeMemory,
+                                          L"RootMemory",
+                                          L"Root",
+                                          NULL);
+    return Status;
+}
+
+/* Port arbiter */
+
+NTSTATUS
+NTAPI
+IopPortInitialize(VOID)
+{
+    NTSTATUS Status;
+
+    DPRINT("IopRootPortArbiter %p\n", &IopRootPortArbiter);
+
+    Status = ArbInitializeArbiterInstance(&IopRootPortArbiter,
+                                          NULL,
+                                          CmResourceTypePort,
+                                          L"RootPort",
+                                          L"Root",
+                                          NULL);
+    return Status;
+}
+
+/* EOF */
--- a/ntoskrnl/io/pnpmgr/pnpinit.c
+++ b/ntoskrnl/io/pnpmgr/pnpinit.c
@@ -24,6 +24,18 @@ PUNICODE_STRING PiInitGroupOrderTable;
 USHORT PiInitGroupOrderTableCount;
 INTERFACE_TYPE PnpDefaultInterfaceType;
 
+ARBITER_INSTANCE IopRootBusNumberArbiter;
+ARBITER_INSTANCE IopRootIrqArbiter;
+ARBITER_INSTANCE IopRootDmaArbiter;
+ARBITER_INSTANCE IopRootMemArbiter;
+ARBITER_INSTANCE IopRootPortArbiter;
+
+NTSTATUS NTAPI IopPortInitialize(VOID);
+NTSTATUS NTAPI IopMemInitialize(VOID);
+NTSTATUS NTAPI IopDmaInitialize(VOID);
+NTSTATUS NTAPI IopIrqInitialize(VOID);
+NTSTATUS NTAPI IopBusNumberInitialize(VOID);
+
 /* FUNCTIONS ******************************************************************/
 
 INTERFACE_TYPE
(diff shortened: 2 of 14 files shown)


sdk/lib/arbiter/arbiter.h
```diff
@@ -0,0 +1,255 @@
...
+    ULONGLONG CurrentMinimum;
+    ULONGLONG CurrentMaximum;
+    PARBITER_LIST_ENTRY Entry;
+    PARBITER_ALTERNATIVE CurrentAlternative;
+    ULONG AlternativeCount;
+    PARBITER_ALTERNATIVE Alternatives;
+    USHORT Flags;
+    UCHAR RangeAttributes;
+    UCHAR RangeAvailableAttributes;
+    ULONG WorkSpace;
```
ThFabba: Should be ULONG_PTR

ntoskrnl/io/pnpmgr/pnpinit.c
```diff
@@ -24,6 +24,18 @@ PUNICODE_STRING PiInitGroupOrderTable;
...
+ARBITER_INSTANCE IopRootIrqArbiter;
+ARBITER_INSTANCE IopRootDmaArbiter;
+ARBITER_INSTANCE IopRootMemArbiter;
+ARBITER_INSTANCE IopRootPortArbiter;
+
+NTSTATUS NTAPI IopPortInitialize();
+NTSTATUS NTAPI IopMemInitialize();
+NTSTATUS NTAPI IopDmaInitialize();
+NTSTATUS NTAPI IopIrqInitialize();
+NTSTATUS NTAPI IopBusNumberInitialize();
```
ThFabba: `()` here means "any number of parameters. Should be `(VOID)`.

sdk/lib/arbiter/arbiter.h
```diff
@@ -0,0 +1,255 @@
...
+typedef NTSTATUS
+(NTAPI * PARB_OVERRIDE_CONFLICT)(
+    _In_ PARBITER_INSTANCE Arbiter
+);
+
+typedef struct _ARBITER_INSTANCE
+{
+    ULONG Signature;
+    PKEVENT MutexEvent;
+    PUSHORT Name;
```
ThFabba: `PCWSTR` (or at least `PWSTR`) seems appropriate here.

sdk/lib/arbiter/arbiter.h
```diff
@@ -0,0 +1,255 @@
...
+);
+
+typedef NTSTATUS
+(NTAPI * PARB_QUERY_CONFLICT)(
+    _In_ PARBITER_INSTANCE Arbiter
+);
+
+typedef NTSTATUS
+(NTAPI * PARB_ADD_RESERVED)(
+    _In_ PARBITER_INSTANCE Arbiter
```
ThFabba: This function seems to also take a `PIO_RESOURCE_DESCRIPTOR` and a `PCM_PARTIAL_RESOURCE_DESCRIPTOR`

sdk/lib/arbiter/arbiter.h
```diff
@@ -0,0 +1,255 @@
...
+);
+
+typedef NTSTATUS
+(NTAPI * PARB_QUERY_ARBITRATE)(
+    _In_ PARBITER_INSTANCE Arbiter
+);
+
+typedef NTSTATUS
+(NTAPI * PARB_QUERY_CONFLICT)(
+    _In_ PARBITER_INSTANCE Arbiter
```
ThFabba: It seems the functions marked as "not used yet" below don't have correct prototypes yet.
I'd suggest one of:
* Add a TODO or FIXME comment above the typedef stating that it's not correct yet
* Use PVOID in the `ARBITER_INSTANCE` structure instead for the time being
* Figure out these types and complete them.

sdk/lib/arbiter/arbiter.h
```diff
@@ -0,0 +1,255 @@
...
+);
+
+NTSTATUS
+NTAPI
+ArbInitializeArbiterInstance(
+    _Inout_ PARBITER_INSTANCE Arbiter,
+    _In_ PDEVICE_OBJECT BusDeviceObject,
+    _In_ CM_RESOURCE_TYPE ResourceType,
+    _In_ PWSTR ArbiterName,
+    _In_ PWSTR OrderName,
```
ThFabba: It seems like it would probably be safe to use `PCWSTR` for these?

sdk/lib/arbiter/arbiter.h
```diff
@@ -0,0 +1,255 @@
...
+#include <stdarg.h>
+#include <stddef.h>
+#include <stdlib.h>
+
+/* PSDK/NDK Headers */
+#define WIN32_NO_STATUS
+#define _INC_WINDOWS
+#define COM_NO_WINDOWS_H
+#define COBJMACROS
+#define CONST_VTABLE
```
ThFabba: I think these are all Win32-specific so shouldn't be needed here.

sdk/lib/arbiter/arbiter.h
```diff
@@ -0,0 +1,255 @@
...
+#include <ndk/iofuncs.h>
+#include <ndk/kefuncs.h>
+#include <ndk/ldrfuncs.h>
+#include <ndk/mmfuncs.h>
+#include <ndk/obfuncs.h>
+#include <ndk/psfuncs.h>
+#include <ndk/rtlfuncs.h>
+#include <ndk/setypes.h>
+#include <ndk/sefuncs.h>
+#include <ndk/umfuncs.h>
```
ThFabba: Since this is the public header for the library, I would suggest only including headers here that are needed to make the definitions below work correctly (which I'm hoping would mean NDK headers aren't needed here at all; and maybe you can get by with just wdm.h instead of ntifs.h).
You could then have another header internal to the library itself that's used as the PCH and includes everything needed for the implementation.

sdk/lib/arbiter/arbiter.h
```diff
@@ -0,0 +1,255 @@
+#ifndef _ARBITER_H
+#define _ARBITER_H
+
+#ifndef _NTOSKRNL_
+#include <ntifs.h>
+#include <wdm.h>
```
ThFabba: This is not necessary, `ntifs.h` already includes `ntddk.h`, which includes `wdm.h`.

sdk/lib/arbiter/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+
+include_directories(${REACTOS_SOURCE_DIR}/sdk/lib/rtl)
+
+list(APPEND SOURCE
+     ${REACTOS_SOURCE_DIR}/ntoskrnl/io/arbiter/arbiter.c
```
ThFabba: Why not have arbiter.c live in this directory and link ntoskrnl with arblib?

ntoskrnl/io/pnpmgr/pnpinit.c
```diff
@@ -38,10 +50,46 @@ NTSTATUS
...
 IopInitializeArbiters(VOID)
 {
-     /* FIXME: TODO */
-    return STATUS_SUCCESS;
+    NTSTATUS Status;
+
+    Status = IopPortInitialize();
+    if (!NT_SUCCESS(Status))
+    {
+        DPRINT1("IopInitializeArbiters: Status %X\n", Status);
```
ThFabba: Nit: I'd prefer if we printed the name of the failing function instead.

sdk/lib/arbiter/arbiter.h
vgalnt: Yes, this is my x64 error...

ntoskrnl/io/pnpmgr/pnpinit.c
vgalnt: Ok.

sdk/lib/arbiter/arbiter.h
vgalnt: Yes.

sdk/lib/arbiter/arbiter.h
vgalnt: I choose the first.

sdk/lib/arbiter/CMakeLists.txt
vgalnt: I see that in the file "ntos.cmake" all files are within the boundaries of "${REACTOS_SOURCE_DIR}/ntoskrnl/"...

ntoskrnl/io/pnpmgr/pnpinit.c
vgalnt: Ok

sdk/lib/arbiter/arbiter.h
vgalnt: Initially, I did not plan to add this. Only after the errors of AppVeyor and Travis did I do this. Therefore, I will try to remove it all. In RoSBE and in MSVS for x86, I have no errors...

vgalnt: It looks like a bug when building in RoSBE and in MSVS goes fine, and in AppVeyor and Travis fails. This looks like a different compilation order or a different set of keys to compile. In any case, I expect the identity of the processes for creating ReactOS ...

sdk/lib/arbiter/CMakeLists.txt
HBelusca: Well then why the "arbiter.c" file would not be part of the compiled "arbiter" library that is in the SDK/lib ? (look e.g. at cmlib that is used also by ntoskrnl)

ntoskrnl/io/pnpmgr/arbs.c
```diff
@@ -0,0 +1,123 @@
+/*
+ * PROJECT:         ReactOS Kernel
+ * COPYRIGHT:       GPL-2.0+ (https://spdx.org/licenses/GPL-2.0+)
+ * FILE:            ntoskrnl/io/debug.c
+ * PURPOSE:         Root arbiters of the PnP manager
+ * PROGRAMMERS:     ReactOS Portable Systems Group
```
HBelusca, line 6: Please correct this :)

ntoskrnl/io/pnpmgr/arbs.c
```diff
@@ -0,0 +1,123 @@
+/*
+ * PROJECT:         ReactOS Kernel
+ * COPYRIGHT:       GPL-2.0+ (https://spdx.org/licenses/GPL-2.0+)
+ * FILE:            ntoskrnl/io/debug.c
```
HBelusca, line 4: Update

sdk/lib/drivers/arbiter/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+
+include_directories(${REACTOS_SOURCE_DIR}/sdk/lib/rtl)
+
+list(APPEND SOURCE
+    arbiter.c
+    arbiter.h)
+
+    add_library(arbiter ${SOURCE})
```
HBelusca, line 8: Remove indentation for this line and the next ones.

sdk/lib/drivers/arbiter/arbiter.c
```diff
@@ -0,0 +1,41 @@
+/*
+ * PROJECT:         ReactOS Kernel
+ * COPYRIGHT:       GPL-2.0+ (https://spdx.org/licenses/GPL-2.0+)
+ * FILE:            ntoskrnl/io/debug.c
+ * PURPOSE:         Arbiter of hardware resources library
+ * PROGRAMMERS:     ReactOS Portable Systems Group
```
HBelusca, line 6: Please update!

sdk/lib/drivers/arbiter/arbiter.c
```diff
@@ -0,0 +1,41 @@
+/*
+ * PROJECT:         ReactOS Kernel
+ * COPYRIGHT:       GPL-2.0+ (https://spdx.org/licenses/GPL-2.0+)
+ * FILE:            ntoskrnl/io/debug.c
```
HBelusca, line 4: Update!

PR #2472, opened 2020-03-27
Improve cdrom targets

zefklop:
## Purpose
Rework how our cdrom targets are generated, with proper dependency tracking, thanks to properties maintained at target level and avoiding calling functions at the end of the party.

Maybe we should move the add_cd_file function to boot/CmakeFiles.txt to make it clearer where everything comes from.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -265,9 +265,6 @@ else()
     # Create the registry hives
     create_registry_hives()
 
-    # Create {bootcd, livecd, bootcdregtest}.lst
-    create_iso_lists()
-
     file(MAKE_DIRECTORY ${REACTOS_BINARY_DIR}/sdk/include/reactos)
 
     add_dependency_footer()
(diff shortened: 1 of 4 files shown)


learn-more:
Maybe cmake is too old for this specific generator expression?

zefklop:
> 
> 
> Maybe cmake is too old for this specific generator expression?

Yes, this will probably have to wait for RosBE 2.2. I tested it with cmake 3.17, I was a bit too edgy.

learn-more:
If you update this with the minimal version for your generator expression,
it will fail with a message stating cmake is too old instead of some weird generator expression error:
https://git.reactos.org/?p=reactos.git;a=blob;f=CMakeLists.txt;hb=ae0e0569d5015f31b79a599a903bb9fb43b74e95#l2

ColinFinck:
RosBE 2.2 is out and required now. Time to retry? :)

PR #2473, opened 2020-03-28
[WORDPAD] Restore code commas in no-NO.rc

SergeGautherie:
".../no-NO.rc:81: syntax error"

NB:
This blocks enabling NO_NO.

### DIFF
--- a/base/applications/wordpad/lang/no-NO.rc
+++ b/base/applications/wordpad/lang/no-NO.rc
@@ -78,9 +78,9 @@ BEGIN
     POPUP "F&ormat"
     BEGIN
         MENUITEM "&Skrift...",               ID_FONTSETTINGS
-        MENUITEM "&Punktmerking"             ID_BULLET
-        MENUITEM "&Avsnitt..."               ID_PARAFORMAT
-        MENUITEM "&Tabulatorer..."           ID_TABSTOPS
+        MENUITEM "&Punktmerking",            ID_BULLET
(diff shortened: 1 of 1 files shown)


Harteex:
> This obviously looks good.
> But the fact that `LANGUAGE_NO_NO` is not enabled by default shows a bigger problem: it's actually supposed to be `nn-NO.rc` / `LANGUAGE_NN_NO`.

According to the .rc file, it is SUBLANG_NORWEGIAN_BOKMAL which should be nb-NO, not nn-NO.

nb: Bokmål
nn: Nynorsk

ThFabba:
> 
> 
> > This obviously looks good.
> > But the fact that `LANGUAGE_NO_NO` is not enabled by default shows a bigger problem: it's actually supposed to be `nn-NO.rc` / `LANGUAGE_NN_NO`.
> 
> According to the .rc file, it is SUBLANG_NORWEGIAN_BOKMAL which should be nb-NO, not nn-NO.
> 
> nb: Bokmål
> nn: Nynorsk

Indeed, thanks!
`LANGUAGE_NB_NO` is used in lots of places, so that's probably the right thing to replace `LANGUAGE_NO_NO` with (we don't have any .rc files that use `SUBLANG_NORWEGIAN_NYNORSK`).

The usual pattern seems to be
```
#ifdef LANGUAGE_NB_NO
    #include "lang/no-NO.rc"
#endif
```
which is a little confusing, but at least it's used everywhere (we don't have a single nb-NO.rc or nn-NO.rc)

PR #2476, opened 2020-03-28
[RTL] Introduce RtlpImageNtHeader

learn-more:
which implements the required functionality.
ntdll and ntoskrnl now have a wrapper for this, with SEH.
This protects the function against malformed / bad images,
whilst still being able to use the code in freeldr et al.
Idea from Thomas.
CORE-14857

JIRA issue: [CORE-14857](https://jira.reactos.org/browse/CORE-14857)

### DIFF
--- a/dll/ntdll/rtl/libsupp.c
+++ b/dll/ntdll/rtl/libsupp.c
@@ -505,6 +505,49 @@ RtlpGetAtomEntry(PRTL_ATOM_TABLE AtomTable, ULONG Index)
    return NULL;
 }
 
+/* Ldr SEH-Protected access to IMAGE_NT_HEADERS */
+
+/* Rtl SEH-Free version of this */
+NTSTATUS
+NTAPI
+RtlpImageNtHeaderEx(
+    _In_ ULONG Flags,
+    _In_ PVOID Base,
+    _In_ ULONG64 Size,
+    _Out_ PIMAGE_NT_HEADERS *OutHeaders);
+
+
+/*
+ * @implemented
+ * @note: This is here, so that we do not drag SEH into rosload, freeldr and bootmgfw
+ */
+NTSTATUS
+NTAPI
+RtlImageNtHeaderEx(
+    _In_ ULONG Flags,
+    _In_ PVOID Base,
+    _In_ ULONG64 Size,
+    _Out_ PIMAGE_NT_HEADERS *OutHeaders)
+{
+    NTSTATUS Status;
+
+    /* Assume failure. This is also done in RtlpImageNtHeaderEx, but this is guarded by SEH. */
+    if (OutHeaders != NULL)
+        *OutHeaders = NULL;
+
+    _SEH2_TRY
+    {
+        Status = RtlpImageNtHeaderEx(Flags, Base, Size, OutHeaders);
+    }
+    _SEH2_EXCEPT(EXCEPTION_EXECUTE_HANDLER)
+    {
+        /* Fail with the SEH error */
+        Status = _SEH2_GetExceptionCode();
+    }
+    _SEH2_END;
+
+    return Status;
+}
 
 /*
  * Ldr Resource support code
--- a/ntoskrnl/rtl/libsupp.c
+++ b/ntoskrnl/rtl/libsupp.c
@@ -691,6 +691,49 @@ RtlpGetAtomEntry(PRTL_ATOM_TABLE AtomTable, ULONG Index)
    return Entry;
 }
 
+/* Ldr SEH-Protected access to IMAGE_NT_HEADERS */
+
+/* Rtl SEH-Free version of this */
+NTSTATUS
+NTAPI
+RtlpImageNtHeaderEx(
+    _In_ ULONG Flags,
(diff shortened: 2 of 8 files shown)


boot/freeldr/freeldr/lib/mm/meminit.c
```diff
@@ -243,7 +243,7 @@ MmCheckFreeldrImageFile(VOID)
     PIMAGE_OPTIONAL_HEADER OptionalHeader;
 
     /* Get the NT headers */
-    NtHeaders = RtlImageNtHeader(&__ImageBase);
+    NtHeaders = RtlpImageNtHeader(&__ImageBase);
```
ThFabba: I know at least freeldr also has a `libsupp.c`, so you could also add the Rtl- versions there, and therefore avoid changing callers. That might also work for bootmgfw/rosload?
(you also wouldn't need `RtlpImageNtHeader` at all in that case)

dll/ntdll/rtl/libsupp.c
```diff
@@ -505,6 +505,62 @@ RtlpGetAtomEntry(PRTL_ATOM_TABLE AtomTable, ULONG Index)
...
+    _In_ ULONG Flags,
+    _In_ PVOID Base,
+    _In_ ULONG64 Size,
+    _Out_ PIMAGE_NT_HEADERS *OutHeaders)
+{
+    NTSTATUS Status;
+
+    _SEH2_TRY
+    {
+        Status = RtlpImageNtHeaderEx(Flags, Base, Size, OutHeaders);
```
ThFabba: Since `RtlpImageNtHeaderEx` sets `*OutHeaders = NULL` as its first operation, I guess we're fine wrt making sure that always happens in case of an exception.
We'll have slightly different behavior here in the case where `OutHeaders` is an invalid but non-null pointer (we'll catch the exception and return, whereas the Windows implementation probably just crashes). Could be worth doing an explicit `*OutHeaders = NULL` in the exception case, or could be useless code that's never gonna be relevant :p

ntoskrnl/rtl/libsupp.c
```diff
@@ -691,6 +691,67 @@ RtlpGetAtomEntry(PRTL_ATOM_TABLE AtomTable, ULONG Index)
...
+NTAPI
+RtlpImageNtHeaderEx(
+    _In_ ULONG Flags,
+    _In_ PVOID Base,
+    _In_ ULONG64 Size,
+    _Out_ PIMAGE_NT_HEADERS *OutHeaders);
+
+#ifndef RTL_IMAGE_NT_HEADER_EX_FLAG_NO_RANGE_CHECK
+#define RTL_IMAGE_NT_HEADER_EX_FLAG_NO_RANGE_CHECK          0x00000001
+#endif
```
ThFabba: Do we not always have NDK included here?

dll/ntdll/rtl/libsupp.c
learn-more: This is a cornercase, but also something that might be very hard to find if it causes problems.
How about I set the OutHeaders to `NULL` before entering SEH?

ntoskrnl/rtl/libsupp.c
learn-more: The symbol was not available, would you prefer I try to include the correct header?

boot/freeldr/freeldr/lib/mm/meminit.c
learn-more: This sounds like a cleaner solution.

ntoskrnl/rtl/libsupp.c
learn-more: It is defined under `NTOS_MODE_USER`

learn-more: this also causes a problem in freeldr

dll/ntdll/rtl/libsupp.c
ThFabba: Yeah that would work too. Just need to handle the `OutHeaders == NULL` case, too, I suppose.

ntoskrnl/rtl/libsupp.c
ThFabba: Hrm. Since this function clearly exists in kernel mode, it should be fine to move the define out of `NTOS_MODE_USER` in NDK.
But if that causes trouble elsewhere, the redefine here seems reasonable.

dll/ntdll/rtl/libsupp.c
```diff
@@ -505,6 +505,49 @@ RtlpGetAtomEntry(PRTL_ATOM_TABLE AtomTable, ULONG Index)
...
+RtlpImageNtHeaderEx(
+    _In_ ULONG Flags,
+    _In_ PVOID Base,
+    _In_ ULONG64 Size,
+    _Out_ PIMAGE_NT_HEADERS *OutHeaders);
+
+
+/*
+ * @implemented
+ * @note: This is here, so that we do not drag SEH into rosload, freeldr and bootmgfw
```
HBelusca: Sentence ends with full stop :P

PR #2478, opened 2020-03-28
[SDK] Align PROCESS_PRIORITY_CLASS of 4-bytes (ULONG)

Fraizeraust:
More information can be found in the commit description.
**BEFORE:**
Capture
**AFTER:**
Capture2

### DIFF
--- a/sdk/include/ndk/pstypes.h
+++ b/sdk/include/ndk/pstypes.h
@@ -902,7 +902,7 @@ typedef struct _PROCESS_SESSION_INFORMATION
 
 #endif
 
-typedef struct _PROCESS_PRIORITY_CLASS
+typedef struct DECLSPEC_ALIGN(4) _PROCESS_PRIORITY_CLASS
 {
     BOOLEAN Foreground;
     UCHAR PriorityClass;


Fraizeraust:
As this PR deals with a kernel mode structure, it'd be appreciated if this PR could be ran in a testman to check for possible regressions. More safer this way than being sorry.

sdk/include/ndk/pstypes.h
```diff
@@ -906,7 +906,7 @@ typedef struct _PROCESS_PRIORITY_CLASS
 {
     BOOLEAN Foreground;
     UCHAR PriorityClass;
-} PROCESS_PRIORITY_CLASS, *PPROCESS_PRIORITY_CLASS;
+}__attribute__((aligned(sizeof(ULONG)))) PROCESS_PRIORITY_CLASS, *PPROCESS_PRIORITY_CLASS;
```
katahiromz: Is __attribute__ available in MSVC?

Fraizeraust: I don't know, this structure is not available publicly as it's part of NDK. But with the investigations being done and the testcase pushed to master, it seems like Windows uses the very same attribute.

ThFabba: It's not. You want `DECLSPEC_ALIGN` instead for the cross-platform version.
Also, check out this comment for correct placement of the attribute: https://git.reactos.org/?p=reactos.git;a=blob;f=sdk/include/xdk/ntbasedef.h;hb=ae0e0569d5015f31b79a599a903bb9fb43b74e95#l241

sdk/include/ndk/pstypes.h
```diff
@@ -906,7 +906,7 @@ typedef struct _PROCESS_PRIORITY_CLASS
...
 
 typedef struct _PROCESS_FOREGROUND_BACKGROUND
```
ThFabba: This structure likely has the same problem. You could look at that in a follow-up if you wanted.

Fraizeraust: Sure thing, I can take a look at this structure as well. But I'd rather create a separate PR for that to avoid cluttering this one with unrelated changes.

Extravert-ir: That's fine, I think

Fraizeraust:
Comments addressed. When someone initiates a testman for this please put the link results here.

HBelusca:
Hi you should potentially fix also https://git.reactos.org/?p=reactos.git;a=blob;f=sdk/include/reactos/wine/winternl.h;hb=99f9c1666de320517ddd2564c8ec3dbb1f892087#l1321

HBelusca:
oops

PR #2485, opened 2020-03-30
[SPEC2DEF] Omit skipped exports from ordinal numbering.

learn-more:
JIRA issue: [CORE-16769](https://jira.reactos.org/browse/CORE-16769)

### DIFF
--- a/sdk/tools/spec2def/spec2def.c
+++ b/sdk/tools/spec2def/spec2def.c
@@ -1332,7 +1332,7 @@ ApplyOrdinals(EXPORT* pexports, unsigned cExports)
     /* Pass 2: apply available ordinals */
     for (i = 0, j = 1; i < cExports; i++)
     {
-        if ((pexports[i].uFlags & FL_ORDINAL) == 0)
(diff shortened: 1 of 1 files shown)


ThFabba:
> JIRA issue: [CORE-16769](https://jira.reactos.org/browse/CORE-16769)

This is just cleanup and doesn't actually fix any bug, right?

learn-more:
> > JIRA issue: [CORE-16769](https://jira.reactos.org/browse/CORE-16769)
> 
> This is just cleanup and doesn't actually fix any bug, right?

It fixes that application failing to snap an ordinal import.
By pure accident one of the ordinals it needs fell in a gap.

ThFabba:
But if it imports by ordinal, and we don't explicitly export the right function by ordinal, it's probably going to get a random incorrect function pointer anyway?

learn-more:
> But if it imports by ordinal, and we don't explicitly export the right function by ordinal, it's probably going to get a random incorrect function pointer anyway?

yeah,
just like on win10 where it gets something completely different as well.

PR #2487, opened 2020-03-31
[NTOS:MM] Make the definitions and macros for x86 more human-readable.

vgalnt:
Purpose: make the definitions and macros for x86 MM more human-readable.
Now definitions have become more universal for PAE mode.

### DIFF
--- a/ntoskrnl/include/internal/i386/mm.h
+++ b/ntoskrnl/include/internal/i386/mm.h
@@ -140,7 +140,7 @@
 /* Maximum number of PDEs */
 #define PDE_PER_SYSTEM (PD_COUNT * PDE_PER_PAGE)
 
-/* TODO: It seems this variable is not needed for x86 */
+/* TODO: It seems this constant is not needed for x86 */
 #define PPE_PER_PAGE 1
 
 /* Maximum number of pages for 4 GB of virtual space */
@@ -154,13 +154,13 @@
 #define MI_SYSTEM_PTE_BASE (PVOID)MiAddressToPte(NULL)
 
 /* Base addreses for page directories */
-#define PDE_BASE (ULONG_PTR)MiAddressToPte(PTE_BASE)
+#define PDE_BASE (ULONG_PTR)MiPteToPde(PTE_BASE)
 #define PDE_TOP  (ULONG_PTR)(PDE_BASE + (PDE_PER_SYSTEM * sizeof(MMPDE)) - 1)
 #define PDE_MASK (PDE_TOP - PDE_BASE)
 
 /* The size of the virtual memory area that is mapped using a single PDE */
 #ifndef PDE_MAPPED_VA
-#define PDE_MAPPED_VA  (PTE_PER_PAGE * PAGE_SIZE)
+#define PDE_MAPPED_VA (PTE_PER_PAGE * PAGE_SIZE)
 #endif
 
 /* Maps the virtual address to the corresponding PTE */
@@ -169,7 +169,7 @@
 
 /* Maps the virtual address to the corresponding PDE */
 #define MiAddressToPde(Va) \
-    ((PMMPDE)(PDE_BASE + ((MiAddressToPdeOffset(Va)) * sizeof(MMPTE))))
+    ((PMMPDE)(PDE_BASE + ((MiAddressToPdeOffset(Va)) * sizeof(MMPDE))))
 
 /* Takes the PTE index (for one PD page) from the virtual address */
 #define MiAddressToPteOffset(Va) \
--- a/ntoskrnl/include/internal/i386/mm.h
+++ b/ntoskrnl/include/internal/i386/mm.h
@@ -28,18 +28,7 @@
 #define MI_DEBUG_MAPPING                        (PVOID)0xFFBFF000
 #define MI_HIGHEST_SYSTEM_ADDRESS               (PVOID)0xFFFFFFFF
 
-/* FIXME: These are different for PAE */
-#define PTE_BASE    0xC0000000
-#define PDE_BASE    0xC0300000
-#define PDE_TOP     0xC0300FFF
-#define PTE_TOP     0xC03FFFFF
-
-#define PTE_PER_PAGE 0x400
-#define PDE_PER_PAGE 0x400
-#define PPE_PER_PAGE 1
-
 /* Misc address definitions */
-#define MI_SYSTEM_PTE_BASE                      (PVOID)MiAddressToPte(NULL)
 #define MM_HIGHEST_VAD_ADDRESS \
     (PVOID)((ULONG_PTR)MM_HIGHEST_USER_ADDRESS - (16 * PAGE_SIZE))
 #define MI_MAPPING_RANGE_START              (ULONG)HYPER_SPACE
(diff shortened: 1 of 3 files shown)


ntoskrnl/include/internal/i386/mm.h
```diff
@@ -128,21 +117,68 @@
...
+
+/* Base addresses for page tables */
+#define PTE_BASE (ULONG_PTR)0xC0000000
+#define PTE_TOP  (ULONG_PTR)(PTE_BASE + (MI_MAX_PAGES * sizeof(MMPTE)) - 1)
+#define PTE_MASK (PTE_TOP - PTE_BASE)
+
+#define MI_SYSTEM_PTE_BASE (PVOID)MiAddressToPte(NULL)
+
+/* Base addreses for page directories */
+#define PDE_BASE (ULONG_PTR)MiAddressToPte(PTE_BASE)
```
zefklop: Using MiAddressToPte for defining a PDE is confusing. Either MiAddressToPde(NULL) or MiPteToPde(PTE_BASE)

ntoskrnl/include/internal/i386/mm.h
```diff
@@ -128,21 +117,68 @@
...
+#define PDE_TOP  (ULONG_PTR)(PDE_BASE + (PDE_PER_SYSTEM * sizeof(MMPDE)) - 1)
+#define PDE_MASK (PDE_TOP - PDE_BASE)
+
+/* The size of the virtual memory area that is mapped using a single PDE */
+#ifndef PDE_MAPPED_VA
```
zefklop: Where else would it be defined ?

ntoskrnl/include/internal/i386/mm.h
```diff
@@ -128,21 +117,68 @@
...
+#define PDE_MAPPED_VA  (PTE_PER_PAGE * PAGE_SIZE)
+#endif
+
+/* Maps the virtual address to the corresponding PTE */
+#define MiAddressToPte(Va) \
+    ((PMMPTE)(PTE_BASE + ((((ULONG_PTR)(Va)) / PAGE_SIZE) * sizeof(MMPTE))))
+
+/* Maps the virtual address to the corresponding PDE */
+#define MiAddressToPde(Va) \
+    ((PMMPDE)(PDE_BASE + ((MiAddressToPdeOffset(Va)) * sizeof(MMPTE))))
```
zefklop: sizeof(MMPDE) please.

ntoskrnl/include/internal/i386/mm.h
```diff
@@ -128,21 +117,68 @@
...
-/* Convert an address to a corresponding PTE offset/index */
-#define MiAddressToPteOffset(x) \
-    ((((ULONG)(x)) << 10) >> 22)
+/* Takes the PDE offset (within all PDs pages) from the virtual address */
+#define MiAddressToPdeOffset(Va) (((ULONG_PTR)(Va)) / PDE_MAPPED_VA)
 
-/* Convert an address to a corresponding PDE offset/index */
-#define MiAddressToPdeOffset(x) \
-    (((ULONG)(x)) / (1024 * PAGE_SIZE))
+/* TODO: Free this variable (for offset from the pointer to the PDE) */
```
zefklop: I don't understand the comment. Do you mean "Get rid of this define" ?

ntoskrnl/include/internal/i386/mm.h
vgalnt: https://github.com/reactos/reactos/blob/d37799ea3e52a74493eae40f861b4a5f6a3b8b8f/ntoskrnl/mm/ARM3/miarm.h#L22
Without this ifndef, hal failed compilation.

ntoskrnl/include/internal/i386/mm.h
vgalnt: Ok

ntoskrnl/include/internal/i386/mm.h
HBelusca: Incidentally there also exists a `PDE_PER_PAGE`  in our code base.

ntoskrnl/include/internal/i386/mm.h
vgalnt: I want to use MiAddressToPdeOffset() wherever MiGetPdeOffset() is currently used. Thus freeing this name for use where the parameter is a pointer to PDE (MiGetPdeOffset(_Pde)) instead of MiAddressToPdeOffset (Va).

ntoskrnl/include/internal/i386/mm.h
vgalnt: I agree, this looks clearer.

ntoskrnl/include/internal/i386/mm.h
vgalnt: But it wasn’t used before in ntoskrnl/include/internal/i386/mm.h ...

HBelusca: Maybe because someone made a copy-paste error and forgot to replace where needed, PTE by PDE?

vgalnt: There is no error, I just added PDE_MAPPED_VA in macros that are used by HAL.
https://git.reactos.org/?p=reactos.git;a=blob;f=hal/halx86/include/halp.h;hb=cefc378b0d64a30d246f0ff7a8f71e97de337f78#l530

ntoskrnl/include/internal/i386/mm.h
```diff
@@ -128,21 +117,68 @@
...
+#define MI_SYSTEM_PTE_BASE (PVOID)MiAddressToPte(NULL)
+
+/* Base addreses for page directories */
+#define PDE_BASE (ULONG_PTR)MiPteToPde(PTE_BASE)
+#define PDE_TOP  (ULONG_PTR)(PDE_BASE + (PDE_PER_SYSTEM * sizeof(MMPDE)) - 1)
+#define PDE_MASK (PDE_TOP - PDE_BASE)
+
+/* The size of the virtual memory area that is mapped using a single PDE */
+#ifndef PDE_MAPPED_VA
+#define PDE_MAPPED_VA (PTE_PER_PAGE * PAGE_SIZE)
```
HBelusca: Then, why not using "PDE_PER_PAGE" here?

HBelusca: Basically also, in that miarm.h file:
https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/mm/ARM3/miarm.h;hb=d37799ea3e52a74493eae40f861b4a5f6a3b8b8f#l21
the fact that there is
`#define PDE_MAPPED_VA  (PTE_PER_PAGE * PAGE_SIZE)`
but also in the same file, `PDE_PER_PAGE` is also used is disturbing for me. Is it really intentional to use the `PTE_PER_PAGE` for defining `PDE_MAPPED_VA` or is it an error?

vgalnt: /* The size of the virtual memory area that is mapped using a single PDE */

For "normal" (non-PAE) mode:
1 PDE points to 1 page directory (PD)
1 PD contains 1024 page table entries (PTEs) ==  (PTE_PER_PAGE)
1 PTE corresponds to 1 page
PDE_MAPPED_VA == (1024 * 4096) == (PTE_PER_PAGE * PAGE_SIZE) == 4Mb

For PAE mode:
1 PDE points to 1 page directory (PD)
1 PD contains 512 page table entries (PTEs) ==  (PTE_PER_PAGE)
1 PTE corresponds to 1 page
PDE_MAPPED_VA == (512 * 4096) == (PTE_PER_PAGE * PAGE_SIZE) == 2Mb

Mouvedia:
https://github.com/reactos/reactos/blob/893a3c9d030fd8b078cbd747eeefd3f6ce57e560/ntoskrnl/kd64/i386/kdx86.c#L421
https://github.com/reactos/reactos/blob/893a3c9d030fd8b078cbd747eeefd3f6ce57e560/ntoskrnl/mm/ARM3/session.c#L135
https://github.com/reactos/reactos/blob/893a3c9d030fd8b078cbd747eeefd3f6ce57e560/ntoskrnl/mm/ARM3/iosup.c#L69

Can you take a look at these?

vgalnt:
Mouvedia, I think PAE is not a top priority right now. First need to make not PAE mode functional. But "cleaning up" in the header files will avoid additional changes in the future.

ntoskrnl/include/internal/i386/mm.h
zefklop: Then maybe remove it from miarm.h while we're at it ? At some point, this header will have to disappear and be merged with the other ones.

ntoskrnl/include/internal/i386/mm.h
```diff
@@ -128,21 +117,69 @@
...
+/*  Translating virtual addresses to physical addresses
+        (See: "Intel� 64 and IA-32 Architectures Software Developer�s Manual
+              Volume 3A: System Programming Guide, Part 1, CHAPTER 4 PAGING")
+    Page directory (PD) and Page table (PT) definitions
+    Page directory entry (PDE) and Page table entry (PTE) definitions
+*/
+
+/* Maximum number of page directories pages */
+#ifndef _X86PAE_
+#define PD_COUNT 1        /* Only one page directory page */
```
tkreuzer, line 129: Please don't add these back! I'm very happy that they were removed. There is PDE_PER_PAGE, and that is an "official" definition. This one is pointless, since it just a different and less useful definition of PDE_PER_PAGE.

PR #2489, opened 2020-03-31
[APITEST][NTDLL] Implement RtlGetNtProductType testcase

Fraizeraust:
**Windows Server 2003:**
Capture
**Windows XP:**
Capture2
**ReactOS:**
Capture4

### DIFF
--- a/modules/rostests/apitests/ntdll/RtlGetNtProductType.c
+++ b/modules/rostests/apitests/ntdll/RtlGetNtProductType.c
@@ -1,0 +1,97 @@
+/*
+ * PROJECT:         ReactOS API tests
+ * LICENSE:         GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:         Tests for the RtlGetNtProductType API
+ * COPYRIGHT:       Copyright 2020 Bișoc George <fraizeraust99 at gmail dot com>
+ */
+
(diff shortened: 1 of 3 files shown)


modules/rostests/apitests/ntdll/precomp.h
```diff
@@ -11,5 +11,7 @@
 #include <apitest_guard.h>
 #include <ndk/ntndk.h>
 #include <strsafe.h>
+#include <winbase.h>
+#include <winreg.h>
```
HBelusca: Alternatively they can be placed in the .c file that actually needs them, i.e. RtlGetNtProductType.c .

modules/rostests/apitests/ntdll/RtlGetNtProductType.c
```diff
@@ -0,0 +1,97 @@
...
+
+    /* Query the product type normally from the Registry */
+    Ret = ReturnNtProduct(&ProductNtType);
+    if (!Ret)
+    {
+        ok(Ret, "Failed to query the product type value!\n");
+    }
+
+    /* Now, get the product type from the NTDLL system call */
+    Status = RtlGetNtProductType(&ProductType);
```
ThFabba, line 94: In NT land we tend to reserve the name `Status` for `NTSTATUS` variables.
You could just re-use `Ret` here.

PR #2491, opened 2020-03-31
Releases/0.4.13

DhruvDoshi:
## Purpose
The installation with the React os for the compatible dies isn't working in my virtual machine 

## Proposed changes
Considering the problem we should develop the specific version of the React os in which we have VMware and Virtual box specific tools

## TODO

Currently I am developing the tools which could be used to develop the drivers connection to the installation part of the operating system


ColinFinck:
Invalid pull request

PR #2493, opened 2020-04-01
[STOBJECT] Change eject hardware dialog buttons from OK/Cancel to OK

JaredSmudde:
Change the eject hardware information dialog buttons from OK/Cancel to just OK. It's just an information dialog that needs one button, it's not actually doing different actions based on user input.

Comparison


JaredSmudde:
Not sure why this is requesting a code review from learn-more.

PR #2494, opened 2020-04-01
[RPCRT4] Fully WineSync spec file. CORE-16441

SergeGautherie:
## Purpose

Addendum to WineSync 4.18.
Plus a new commit.

JIRA issue: [CORE-16441](https://jira.reactos.org/browse/CORE-16441)

## TODO

- [ ] Review numbering.


dll/win32/rpcrt4/rpcrt4.spec
```diff
@@ -267,7 +267,8 @@
 267 stdcall NdrSendReceive(ptr ptr)
 268 stdcall NdrServerCall2(ptr)
 269 stdcall NdrServerCall(ptr)
-270 stdcall NdrServerContextMarshall(ptr ptr long)
+# @ stdcall NdrServerCallAll(ptr)
```
SergeGautherie, line 270: Wine uses `@` in its whole file. ReactOS did too, until a775d6728645bbb710b6ea7c497fdf0700663662.
Should I restore use of `@` (now)? Or which number should this function (and all the others) have?

HBelusca: For the time being, keep the explicit ordinals.
NdrServerContextMarshall is ordinal 270
NdrServerCallAll does not exist in Windows 2003; it has been introduced only on Windows 7+.

HBelusca: As a developer you should know that MSDN (when it comes especially to this "Minimum supported" section) is 90% wrong, so what should be looked at are the actual files, and when you take the rpcrt4.dll from all Windows <(=) Vista you note that there is no export whatsoever called "NdrServerCallAll".

HBelusca: For x64-specific exports, use `@ stdcall -arch=x86_64 ExportName`
For x86-specific exports, use `@ stdcall -arch=i386 ExportName`

HBelusca:
Please fix your rebasing (currently the PR is polluted with unrelated commits that already exist in master).

SergeGautherie:
> fix your rebasing

Done. (Not sure what happened: was all right locally.)

dll/win32/rpcrt4/rpcrt4.spec
SergeGautherie: > actual files

I know, but I often have no (easy) access to these.

> Windows <(=) Vista

For the record, Vista x86 does not have this function either.

> see comments above

Ah, indeed: it's good you noticed it! (I would have expected an `#ifdef` :-/)

> -arch=x86_64

Replaced. Or, in this case, should it be `-arch=win64`?

HBelusca: It's called "x86_64" in our spec files (just check the examples in our code base)

SergeGautherie: >  "x86_64"

For what it's worth, ROS has both cases:
https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=\-arch\%3Dx86_64&sr=1
https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=\-arch\%3Dwin64&sr=1

PR #2495, opened 2020-04-01
[SYSDM] Cumulative update 2020-04-01 for SMBIOS parser

binarymaster:
Totally not an April Fools day joke. :ok_hand:

### DIFF
--- a/dll/cpl/sysdm/smbios.c
+++ b/dll/cpl/sysdm/smbios.c
@@ -3,7 +3,7 @@
  * LICENSE:     GPL - See COPYING in the top level directory
  * FILE:        dll/cpl/sysdm/smbios.c
  * PURPOSE:     Retrieve device or motherboard name identifier from DMI/SMBIOS
- * COPYRIGHT:   Copyright 2018-2019 Stanislav Motylkov <x86corez@gmail.com>
+ * COPYRIGHT:   Copyright 2018-2020 Stanislav Motylkov <x86corez@gmail.com>
  *
  */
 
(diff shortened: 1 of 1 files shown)


dll/cpl/sysdm/smbios.c
```diff
@@ -430,6 +465,26 @@ BOOL TrimNonPrintable(PCHAR DmiString)
...
+{
+    PCHAR c;
+
+    if (!DmiString)
+    {
+        return FALSE;
+    }
+
+    /* Special handling for HP with broken revision */
+    c = strstr(DmiString, "(\xFF\xFF");
```
HBelusca, line 479: Oh they really have a NULL-terminated string with FF bytes??

binarymaster: Yes this happens in some hardware setups / SMBIOS dumps :)

HBelusca: Why does it need to be a special case? Isn't this already handled by TrimNonPrintable?

binarymaster: It didn't worked as expected for some reason, already forgot why... it's about this case:

image

HBelusca: I would have expected that the TrimNonPrintable would parse the full-device name string from the beginning, then reach the opening parenthesis (still OK), then reach the character 0xFF (displayed with "ya" in the screencapture), see it's invalid and NULL-terminates the string there?

binarymaster: Yes that's correct. The idea behind this additional function is to change return value to `FALSE` as if the string was not trimmed. Preprocessed string is passed to `TrimNonPrintable` and it returns `FALSE`. If we pass the string as is, it would return `TRUE`, and based on this value later the logic decides to use board strings instead of system strings, but they are more usable in this exact case. That's why additional function was introduced.

HBelusca: ok then you should defintively add a comment in the code explaining that.

binarymaster: Added a comment ;)

PR #2500, opened 2020-04-05
added a flag for mandatory password

erkinalp:
## Purpose

Just added a simple flag for mandatory password in syssetup.

JIRA issue: [CORE-8038](https://jira.reactos.org/browse/CORE-8038)

### DIFF
--- a/dll/win32/syssetup/wizard.c
+++ b/dll/win32/syssetup/wizard.c
@@ -939,7 +939,7 @@ ComputerPageDlgProc(HWND hwndDlg,
                         return TRUE;
                     }
 
-#if 0
+#if PASSWORDS_MANDATORY
                     /* Check if admin passwords have been entered */
                     if ((GetDlgItemText(hwndDlg, IDC_ADMINPASSWORD1, Password1, 128) == 0) ||
                         (GetDlgItemText(hwndDlg, IDC_ADMINPASSWORD2, Password2, 128) == 0))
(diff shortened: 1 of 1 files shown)


dll/win32/syssetup/wizard.c
```diff
@@ -939,7 +939,7 @@ ComputerPageDlgProc(HWND hwndDlg,
                         return TRUE;
                     }
 
-#if 0
+#if PASSWORDS_MANDATORY
```
binarymaster, line 942: I don't see where this is defined :thinking: https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=PASSWORDS_MANDATORY

erkinalp: I am going to add it shortly.

ThFabba: Yes, that's the point. This doesn't enable requiring passwords, it just allows you to enable it using a define.

binarymaster:
```suggestion
#ifdef PASSWORDS_MANDATORY
```
Then maybe something like that. And `//#define PASSWORDS_MANDATORY` commented definition somewhere.

Fraizeraust:
Didn't Eric Kohl talk about the reason of the mandatory password chunk of code being disabled in compile time?
https://jira.reactos.org/browse/CORE-8038?focusedCommentId=57546&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-57546

You should include mandatory password filling only for release builds.

erkinalp:
@Feizeraust You would have a build configuraiton with `PASSWORDS_MANDATORY=1` defined in release builds.

PR #2504, opened 2020-04-06
[BOOTVID][NTOSKRNL][HALX86] Move hardcoded coordinates of bootvid into one place

disean:
## Purpose

Reduce the hardcoded values. Easier way to add custom resolutions. No visual changes.

### DIFF
--- a/ntoskrnl/include/internal/inbv.h
+++ b/ntoskrnl/include/internal/inbv.h
@@ -1,5 +1,8 @@
 #pragma once
 
+// Native definitions from BOOTVID (Boot Video Driver).
+#include "bootvid/bootvid.h"
+
 //
 // Driver Initialization
 //
--- a/sdk/include/reactos/drivers/bootvid/display.h
+++ b/sdk/include/reactos/drivers/bootvid/display.h
@@ -1,0 +1,12 @@
+/*
+ * PROJECT:     ReactOS Boot Video Driver
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Provides display settings for BOOTVID, HAL and NTOSKRNL.
+ * COPYRIGHT:   Copyright 2020 Dmitry Borisov (di.sean@protonmail.com)
+ */
+
+#pragma once
+
+/* For default VGA */
+#define SCREEN_WIDTH    640
+#define SCREEN_HEIGHT   480
--- a/drivers/base/bootvid/precomp.h
+++ b/drivers/base/bootvid/precomp.h
@@ -11,12 +11,6 @@
 #define BOOTCHAR_HEIGHT 13
 #define BOOTCHAR_WIDTH  8 // Each character line is encoded in a UCHAR.
 
-/*
- * Screen resolution (for default VGA)
- */
-#define SCREEN_WIDTH    640
-#define SCREEN_HEIGHT   480
-
 #ifndef _M_ARM
 #include "vga.h"
 #endif /* _M_ARM */
--- a/hal/halx86/generic/misc.c
+++ b/hal/halx86/generic/misc.c
@@ -9,6 +9,8 @@
 /* INCLUDES *******************************************************************/
 
 #include <hal.h>
+#include <drivers/bootvid/display.h>
+
 #define NDEBUG
 #include <debug.h>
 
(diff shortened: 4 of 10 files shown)


sdk/include/reactos/drivers/bootvid/display.h
```diff
@@ -0,0 +1,39 @@
...
+ * COPYRIGHT:   Copyright 2020 Dmitry Borisov (di.sean@protonmail.com)
+ */
+
+#pragma once
+
+/* For default VGA */
+#define SCREEN_WIDTH    640
+#define SCREEN_HEIGHT   480
+
+#define VID_SCROLL_AREA_LEFT       32
```
HBelusca, line 14: Are you going to use these VID_ defines elsewhere than in NTOSKRNL? If not I would strongly suggest you move them back to some header in ntoskrnl/inbv.

binarymaster: Yes moving this header to inbv would be better. It looks a bit strange when kernel and HAL import header from some driver, and not vice versa. :)

sdk/include/reactos/drivers/bootvid/display.h
```diff
@@ -0,0 +1,39 @@
...
+#define VID_SKU_AREA_RIGHT        454
+#define VID_SKU_AREA_BOTTOM       256
+
+#define VID_SKU_SAVE_AREA_LEFT    413
+#define VID_SKU_SAVE_AREA_TOP     237
+
+#define VID_SKU_TEXT_LEFT         180
+#define VID_SKU_TEXT_TOP          121
+
+#define VID_FOOTER_BGD_TOP        (SCREEN_HEIGHT - 59)
```
binarymaster, line 39:
```suggestion
#define VID_FOOTER_BKG_TOP        (SCREEN_HEIGHT - 59)
```
I think BKG is more common/recognizable abbreviation for background.

binarymaster: Or even just "BG", that's even more common. :)

sdk/include/reactos/drivers/bootvid/display.h
```diff
@@ -0,0 +1,12 @@
...
+ * PROJECT:     ReactOS video boot driver
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Provides display settings for BOOTVID, NTOSKRNL and HAL
+ * COPYRIGHT:   Copyright 2020 Dmitry Borisov (di.sean@protonmail.com)
+ */
+
+#pragma once
+
+/* For default VGA */
+#define SCREEN_WIDTH    640
```
binarymaster, line 11: I'd move these to `inbv.h` as well to avoid including driver header in kernel and HAL (as I commented earlier).

HBelusca: But if you do that we then completely loose the point of this PR?
These width & height are the only sizes currently imposed by bootvid that then need to be known by the other components using it (NTOSKRNL, HAL and BOOTVID itself), so I think they are fine here.

ntoskrnl/include/internal/inbv.h
```diff
@@ -128,3 +128,34 @@ InbvPortInitialize(
     OUT PULONG PortId,
     IN BOOLEAN IsMMIODevice
 );
+
```
HBelusca, line 131: So for these values we have to think more about part of a "bootscreen theme", if you see what I mean... Thus I would say they could be placed in a separate header file in the kernel (and in the ntoskrn/inbv directory?)

sdk/include/reactos/drivers/bootvid/display.h
binarymaster: Why? Just create a separate header file in `/inbv/` and include it where needed, just to make it clear - inbv controls the width and height, and bootvid driver imports header from it.

HBelusca: Not sure I understand completely; the bootvid is the main limiter in this case and imposes the full screen width/height, that everything else using it gets, so it's more natural to have a global bootvid header that exposes these values, and code that uses bootvid also including this header.

binarymaster: If you modify the header for different width/height and rebuild bootvid along with kernel and HAL, it would work in ReactOS case. But it won't work for Windows XP/2003, as its kernel will still use its own defined values, which makes me think the kernel (inbv component) is better place for such definitions. The bootvid driver has no way to tell the kernel its own preferred screen resolution, so we assume it's hardcoded, and if it's really hardcoded, it's done in the ntos:inbv.

binarymaster: Just think of it like... the ntos:inbv is a "port", and bootvid is a "miniport". :)

HBelusca: Not really. ntos:inbv is just a wrapper around bootvid, and the latter imposes its screen size.

binarymaster: Alright then.

spetreolle: hardcoded.h

PR #2505, opened 2020-04-06
[MSGINA] gui: Remove superfluous DLG_DATA.TimerID

SergeGautherie:
And use a define, with a positive value.
And report SetTimer() failure.


HBelusca:
Does this fix **anything** or not?

SergeGautherie:
> fix

No, it is code improvement.

dll/win32/msgina/gui.c
```diff
@@ -12,6 +12,8 @@
 #include <winnls.h>
 #include <winreg.h>
 
+#define IDT_BAR 1
```
binarymaster, line 15: `IDT_BAR` should probably go to resource header file.

SergeGautherie: It can. Done.

JoachimHenze:
SergeGautherie, What do you think about moving
#define IDT_BAR 1
into gui.c?
Beside that, LGTM.

dll/win32/msgina/gui.c
```diff
@@ -191,7 +190,10 @@ StatusDialogProc(
 
             if (DlgData_LoadBitmaps(pDlgData))
             {
-                pDlgData->TimerID = SetTimer(hwndDlg, -1, 20, NULL);
+                if (SetTimer(hwndDlg, IDT_BAR, 20, NULL) == 0)
+                {
+                    ERR("SetTimer(IDT_BAR) failed: %d\n", GetLastError());
```
HBelusca, line 195: Not sure whether a thing such as the failure to create a timer to make cute animation bar rolling around, is worth an error in the debug log, especially knowing that if such function fails it's because your system is completely out of resources....

SergeGautherie:
> What do you think about moving

It is already there, in 1st (850e710) commit.
I think it is technically better in .c but logically fine in .h: I let reviewer choose.

dll/win32/msgina/gui.c
SergeGautherie: I concur this is debug log only and unlikely to happen. (And harmless, visually noticeable, ...)
Nonetheless, I think it is a good pattern to use: I let reviewer decide.

PR #2507, opened 2020-04-07
[SHIMLIB] Fix gcc8 build crashing when using shims

learn-more:
JIRA issue: [CORE-16514](https://jira.reactos.org/browse/CORE-16514)
There seems to be some padding before `_shim_end`, causing the loop to continue forever.

### DIFF
--- a/dll/appcompat/shims/shimlib/shimlib.c
+++ b/dll/appcompat/shims/shimlib/shimlib.c
@@ -100,7 +100,7 @@ BOOL ShimLib_StrAEqualsWNC(PCSTR szString, PCWSTR wszString)
 #endif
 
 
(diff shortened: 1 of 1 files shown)


dll/appcompat/shims/shimlib/shimlib.c
```diff
@@ -112,7 +112,7 @@ PHOOKAPI WINAPI ShimLib_GetHookAPIs(IN LPCSTR szCommandLine, IN LPCWSTR wszShimN
 {
     PSHIMREG ps = &_shim_start;
     ps++;
```
HBelusca, line 114: Why is the first address skipped?

learn-more: Because it's an empty entry

PR #2508, opened 2020-04-08
WIP. [WINETESTS] 'generated.c' tests are compile-time only checks. ROSTESTS-251

SergeGautherie:
## Purpose

Running them works, but is useless: no actual tests, results=0.

JIRA issue: [ROSTESTS-251](https://jira.reactos.org/browse/ROSTESTS-251)

### DIFF
--- a/modules/rostests/winetests/gdi32/testlist.c
+++ b/modules/rostests/winetests/gdi32/testlist.c
@@ -10,7 +10,9 @@ extern void func_dc(void);
 extern void func_dib(void);
 extern void func_font(void);
 extern void func_gdiobj(void);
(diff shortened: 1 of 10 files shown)


learn-more:
What problem does this solve?

It introduces at least one new problem:
Now we have to think about if this is not eliminated by a compiler due to it being unreachable code.

ThFabba:
There's no need for the ifdef -- our testlists.c files should look exactly like the ones auto-generated for Wine.
If those include `generated`, let's include it everywhere unconditionally. If they don't let's remove it everywhere.
As far as I can tell, Wine's testlists do include generated, so let's make sure we also include it everywhere.

SergeGautherie:
I'll rework this...

PR #2509, opened 2020-04-08
[CMAKE][CONFIGURE] Spread RUNTIME_CHECKS across all supported compilers

Extravert-ir:
Guard all current RUNTIME_CHECKS usages with MSVC compiler check.
Now RUNTIME_CHECKS switch can be used on all platforms (but so far only
MSVC adds actual checks with it).
This commit also removes RTC switch from configure.cmd, now MSVC runtime
checks are supposed to be enabled as a CMake define (-DRUNTIME_CHECKS=1)

I'm making this PR to ensure that developers are OK with removing `RTC` option from `configure.cmd` in it's current form

### DIFF
--- a/boot/environ/CMakeLists.txt
+++ b/boot/environ/CMakeLists.txt
@@ -103,9 +103,12 @@ set_entrypoint(bootmgfw EfiEntry)
 
 target_link_libraries(bootmgfw bootlib cportlib cmlib rtl libcntpr)
 
+# dynamic analysis switches
 if(STACK_PROTECTOR)
     target_link_libraries(bootmgfw gcc_ssp)
-elseif(RUNTIME_CHECKS)
+endif()
+
+if(RUNTIME_CHECKS)
     target_link_libraries(bootmgfw runtmchk)
 endif()
 
@@ -148,9 +151,12 @@ endif()
 
 target_link_libraries(rosload bootlib cportlib cmlib rtl libcntpr)
 
+# dynamic analysis switches
 if(STACK_PROTECTOR)
     target_link_libraries(rosload gcc_ssp)
-elseif(RUNTIME_CHECKS)
+endif()
+
+if(RUNTIME_CHECKS)
     target_link_libraries(rosload runtmchk)
 endif()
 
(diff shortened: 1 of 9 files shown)


boot/environ/CMakeLists.txt
```diff
@@ -105,8 +105,12 @@ target_link_libraries(bootmgfw bootlib cportlib cmlib rtl libcntpr)
 
 if(STACK_PROTECTOR)
     target_link_libraries(bootmgfw gcc_ssp)
-elseif(RUNTIME_CHECKS)
-    target_link_libraries(bootmgfw runtmchk)
+endif()
+
+if(RUNTIME_CHECKS)
+    if(MSVC)
```
learn-more, line 111: Having to do this check everywhere is messy?
I'd just add a function: `ROS_ADD_RUNTIME_CHECKS(target)` and in that function you check if runtime checks are enabled, if they are not ofc do nothing,
if they are enabled you check compiler, and change compiler specific crap.
This way when adding this, we dont have to edit the entire tree, but can simply update this cmake function for new checks.

learn-more: Optionally, can add this function in `sdk/cmake/gcc.cmake` etc, so that each compiler specific file can add it's own checks.

HBelusca: (and no need to prefix the function with "ROS_": all of our own cmake functions/macros don't have this name as well)

ThFabba: Or maybe we could add an empty runtmchk library for non-MSVC targets so that it can be used unconditionally.

ThFabba:
Alternatively you could simply add a new variable for the additional sanitizer(s) you're adding. We already have STACK_PROTECTOR and RUNTIME_CHECKS separately, so could add UB_SANITIZER/ADDRESS_SANITIZER etc.

Extravert-ir:
> Alternatively you could simply add a new variable for the additional sanitizer(s) you're adding. We already have STACK_PROTECTOR and RUNTIME_CHECKS separately, so could add UB_SANITIZER/ADDRESS_SANITIZER etc.

Hm, this may be better in future. Because some sanitizers cannot work with each other. And MSVC has ASan support in beta

Extravert-ir:
> but would prefer to make it the default at some point.

I always run builds with it (and afaik @learn-more too). What stops us from doing it now? :)

ThFabba:
> I always run builds with it (and afaik @learn-more too). What stops us from doing it now? :)

Nothing. Feel free.

PR #2512, opened 2020-04-08
Add code to SxS manifest parser to parse XML namespaces

wjk:
Some application manifest files, such as those emitted by Microsoft mt.exe, use namespace prefixes heavily. Without this change, xml_elem_cmp() would choke on elements with names such as `<ms_compatibility:supportedOS>`. I ran into this problem while debugging `dotnet.exe` from the .NET Core 3.1 SDK on ReactOS. An apitest is included, and passing.

If you want me to squash this PR into fewer commits before merging, just let me know. Enjoy!


sdk/lib/rtl/actctx.c
```diff
@@ -9,7 +9,7 @@
  *                  Jacek Caban for CodeWeavers
  *                  Alexandre Julliard
  *                  Stefan Ginsberg (stefan.ginsberg@reactos.org)
- *                  Samuel Serapi�n
+ *                  Samuel Serapión
```
learn-more, line 12: Unintended change?

learn-more:
Do those manifests work correctly in wine?
Most of this code is synced to their implementation.

sdk/lib/rtl/actctx.c
wjk: Sort of. My editor replaced the `ó` [with a replacement character](https://github.com/reactos/reactos/blob/457165eaad137e6a6aa0e7149c5d49dead19b441/sdk/lib/rtl/actctx.c#L12) for some reason, so I replaced the original letter in d68cbf6. I didn't see it in the diff when I prepared the PR for some reason.

wjk:
@learn-more Wine handles XML namespaces properly (it recognizes xmlns prefixes instead of ignoring them); however, I could not get a simple copy/paste to function properly. This implementation is actually my fourth attempt at solving this problem. (Attempt one was pasting Wine’s code; attempts two and three were overcomplicated and completely broken attempts at building a DOM and walking the manifest XML that way.)

wjk:
Also, IMHO the ability to properly recognize namespace prefixes is not required to correctly parse manifest files. Despite the extensive use of different namespaces in the documentation, the base element names are actually unique enough that we can still recognize what they are without needing to differentiate between different XML namespaces. Furthermore, the differences between asm.v1, asm.v2, and asm.v3 are so slight Wine treats the three namespaces as equivalent.

tkreuzer:
I would prefer to sync this file with wine, if possible.

learn-more:
This seems to introduce new regressions in kernel32:actctx.

wjk:
I have now managed to successfully port Wine’s implementation of activation-context parsing (as of commit https://github.com/wine-mirror/wine/commit/a455ff61b40ff73b48d0ccc9c1f14679bb65ab8d) into our `actctx.c`. Therefore, I have force-pushed to remove my previous attempt, while keeping the unit test. I’m still looking into the test regressions. Thanks!

learn-more:
Awesome, good to see you have not given up!

wjk:
@learn-more After considerable testing, I have come no closer to finding the cause of the test failures in the kernel32 apitest suites DefaultActCtx or FindActCtxSectionStringW. In fact, when I run those two tests on the head of master (without my changes), they fail the exact same way as with on the branch with changes applied! I am now very confused. Could you please clarify what failures you were talking about in your post on April 26? If my changes actually introduce new regressions, I want to fix them, but at this point I’m pretty sure that those two apitest suites are _not_ regressions, since they don’t work on master either. Thanks!

learn-more:
https://reactos.org/testman/compare.php?ids=73592,73594
It seems that there are no additional failures anymore indeed.
So either your latest changes fixed them, or I compared the wrong things

wjk:
@learn-more Is there anything else I should do or is this PR ready to merge? Thanks!

learn-more:
> @learn-more Is there anything else I should do or is this PR ready to merge? Thanks!

I simply have not had the time to look in-depth at this, sorry for the delay.
I'll try to do it soon-ish, but at first glance it seems that you did everything correct.

tkreuzer (changes requested):
Please place the definitions in actctx.c where they are in the wine code, so we will keep at least a little chance to sync this up in the future.

sdk/lib/rtl/actctx.c
```diff
@@ -43,18 +43,36 @@ BOOLEAN RtlpNotAllowingMultipleActivation;
...
-
+#define MAX_NAMESPACES 64
 
 typedef struct
 {
     const WCHAR        *ptr;
     unsigned int        len;
 } xmlstr_t;
 
+static const xmlstr_t empty_xmlstr;
```
tkreuzer, line 54: Place this below actctx_loader

tkreuzer (approved):
Nevermind, looks like everything except that one definition is at the right place.

tkreuzer:
I ran a test on buildbot. Results are good: https://reactos.org/testman/compare.php?ids=74634,74636,74638

PR #2513, opened 2020-04-08
[APISETS] Forward _seh_filter_{dll,exe}

wjk:
## Purpose

Since these functions are forwarded to a different function name, update.py misses them. The choice of target [comes from Wine](https://github.com/wine-mirror/wine/blob/a98ca88dd1139ad8e16e63fbb349c2fc5e9ad332/dlls/ucrtbase/ucrtbase.spec#L1877-L1878).

With this change applied, calls to other stub functions from certain binaries (namely, dotnet.exe from .NET Core) now "correctly" crash, invoking drwtsn32, instead of simply exiting.

### DIFF
--- a/dll/apisets/api-ms-win-crt-runtime-l1-1-0.spec
+++ b/dll/apisets/api-ms-win-crt-runtime-l1-1-0.spec
@@ -66,8 +66,8 @@
 @ stub _register_onexit_function
 @ stub _register_thread_local_exe_atexit_callback
 @ stdcall _resetstkoflw() msvcrt._resetstkoflw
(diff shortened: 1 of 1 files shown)


SergeGautherie:
Does this help [CORE-16700](https://jira.reactos.org/browse/CORE-16700)?
It is suggested there that these functions may be at lest `-version=0x600+`. Are they (not)?

HBelusca:
@learn-more : since it is said that update.py misses them, would it be possible to also add a list of "exceptions" where these mappings (using different names) are specified?

learn-more:
no, this script is going away hopefully.

PR #2518, opened 2020-04-09
[MSGINA] Redraw only the animation bar scroll

Fraizeraust:


dll/win32/msgina/gui.c
```diff
@@ -193,6 +194,9 @@ StatusDialogProc(
             {
                 pDlgData->TimerID = SetTimer(hwndDlg, -1, 20, NULL);
             }
+
+            /* Get the animation bar control */
+            pDlgData->hWndBarCtrl = GetDlgItem(hwndDlg, IDC_BAR);
```
SergeGautherie, line 199: Move it inside the previous `if`.

dll/win32/msgina/gui.c
```diff
@@ -197,6 +198,9 @@ StatusDialogProc(
                 {
                     ERR("SetTimer(IDT_BAR) failed: %d\n", GetLastError());
                 }
+
+                /* Get the animation bar control */
+                pDlgData->hWndBarCtrl = GetDlgItem(hwndDlg, IDC_BAR);
```
SergeGautherie, line 203: Now, add an `else` too :-|

HBelusca: For what?

PR #2520, opened 2020-04-09
[MSGINA] gui: Improve DlgData_LoadBitmaps()

SergeGautherie:
And let bar timer depend on the bar image only.

### DIFF
--- a/dll/win32/msgina/gui.c
+++ b/dll/win32/msgina/gui.c
@@ -56,28 +56,34 @@ DlgData_Create(HWND hwndDlg, PGINA_CONTEXT pgContext)
     return pDlgData;
 }
 
-static BOOL
-DlgData_LoadBitmaps(PDLG_DATA pDlgData)
+static VOID
+DlgData_LoadBitmaps(_Inout_ PDLG_DATA pDlgData)
 {
     BITMAP bm;
 
     if (!pDlgData)
-        return FALSE;
+    {
+        return;
+    }
 
     pDlgData->hLogoBitmap = LoadImageW(pDlgData->pgContext->hDllInstance,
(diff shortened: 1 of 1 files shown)


SergeGautherie:
Addendum to 623dd26ccea5e79001aee769c230fb4f8e304fc8: Cc @katahiromz

dll/win32/msgina/gui.c
```diff
@@ -56,28 +56,34 @@ DlgData_Create(HWND hwndDlg, PGINA_CONTEXT pgContext)
     return pDlgData;
 }
 
-static BOOL
-DlgData_LoadBitmaps(PDLG_DATA pDlgData)
+static void
```
Fraizeraust, line 59: How would declaring the function as VOID be considered as an improvement? Could you elaborate?

SergeGautherie: Case fixed.
The current return value is not what is wanted, then it becomes unused.

dll/win32/msgina/gui.c
```diff
@@ -56,28 +56,34 @@ DlgData_Create(HWND hwndDlg, PGINA_CONTEXT pgContext)
     return pDlgData;
 }
 
-static BOOL
-DlgData_LoadBitmaps(PDLG_DATA pDlgData)
+static VOID
+DlgData_LoadBitmaps(_Out_ PDLG_DATA pDlgData)
```
learn-more, line 60: Technically this is `_Inout_`

SergeGautherie: Right, I missed to account for the `pDlgData->pgContext->hDllInstance` read. Fixed.

PR #2521, opened 2020-04-09
[DXDIAG] ddtest: Fix timer handling

SergeGautherie:
Most notably, add 2 KillTimer() in DDOffscreenBufferTest().


learn-more:
And while at it, also change all timer related code just for the heck of it?

SergeGautherie:
>  just for the heck of it?

Reasons are similar/worse to #2505:
superfluous variable/code, no define, negative value. no SetTimer() failure check, missing KillTimer(), ...

I am sad you prefer poor/broken code :-(

binarymaster:
> I am sad you prefer poor/broken code :-(

We prefer no code bikeshedding.

PR #2522, opened 2020-04-10
Update zh-CN.rc

robsean:
Fixing translation in ReactOS setup GUI program.


HBelusca:
Hi! Do you have a latin transliteration of your name so that we can use it in the commit author?

robsean:
> 
> 
> Hi! Do you have a latin transliteration of your name so that we can use it in the commit a

> 
> 
> Hi! Do you have a latin transliteration of your name so that we can use it in the commit author?

robsean

PR #2526, opened 2020-04-10
[IPHLPAPI] GetAdaptersAddresses(): Add 1 free() in ERROR_NO_DATA case

SergeGautherie:
And fix 4 comments.

Follow-up to 68d94581ad181efe860a168dc6938846c87638fc.

### DIFF
--- a/dll/win32/iphlpapi/iphlpapi_main.c
+++ b/dll/win32/iphlpapi/iphlpapi_main.c
@@ -3186,7 +3186,10 @@ DWORD WINAPI DECLSPEC_HOTPATCH GetAdaptersAddresses(ULONG Family,ULONG Flags,PVO
 
     ret = openTcpFile(&tcpFile, FILE_READ_DATA);
     if (!NT_SUCCESS(ret))
+    {
+        free(indexTable);
         return ERROR_NO_DATA;
+    }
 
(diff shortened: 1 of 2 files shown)


SergeGautherie:
While here, I noticed:
* `getInterfaceIndexTableInt()` uses `calloc()`, not `HeapAlloc()`.
* `GetTcpTable()` and `GetUdpTable()` parameters are different from Windows ones.

Should we do something about theses, separately?

ThFabba:
>     * `getInterfaceIndexTableInt()` uses `calloc()`, not `HeapAlloc()`.

Nothing wrong with `calloc`. It's even safer since HeapAlloc since it catches integer overflow for you!


>     * `GetTcpTable()` and `GetUdpTable()` parameters are different from Windows ones.

Different how? DWORD vs ULONG? They're the same.

SergeGautherie:
> It's even safer than HeapAlloc

I am simply wondering why we don't use the same function everywhere, either one...
As in, was `s/free/HeapFree/` better than `s/HeapAlloc/calloc/` (or simply `malloc`)?

> They're the same

Indeed, sorry: I wrongly looked at `getTcpTable(CLASS_TABLE Class)`.

PR #2528, opened 2020-04-11
Create .editorconfig

Saibamen:
## Purpose

Add .editorconfig file to automate some files-params while coding

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,15 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*]
+end_of_line = crlf
+insert_final_newline = true
+charset = utf-8
+indent_size = 4
+indent_style = space
+trim_trailing_whitespace = true
+
+[*.{json,yml,md}]
+indent_size = 2


.editorconfig
```diff
@@ -0,0 +1,15 @@
...
+
+[*]
+end_of_line = crlf
+insert_final_newline = true
+charset = utf-8
+indent_size = 4
+indent_style = space
+trim_trailing_whitespace = true
+
+[*.{json,yml,md}]
```
HBelusca, line 14: Do we use json/yml in our codebase?

SergeGautherie: yml is used for appveyor, gitpod and travis, at least.

.editorconfig
```diff
@@ -0,0 +1,15 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*]
+end_of_line = crlf
```
ThFabba, line 7: This is not accurate, most files should use native line endings. See `.gitattributes`.

HBelusca: According to
https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#end_of_line
there is not yet a value to translate "native line endings"
so I will remove it from the editorconfig.

HBelusca: Fixed by commit deebc5848e4e1cff9cbf3706c5ea370076eeb160.

PR #2534, opened 2020-04-11
Delete gitignored files

Saibamen:
## Purpose

This is just result of run this commands:
```
git rm -r --cached .
git add .
```


ThFabba:
The right fix here is probably to adjust .gitignore.

Saibamen:
So we need those files?
Please check CI for this PR. AppVeyor and Travis has builded successfully

ThFabba:
I don't think we use those files, but they're included to keep the structure of the upstream library.

Saibamen:
I have made alternative solution: https://github.com/reactos/reactos/pull/2538

Extravert-ir:
Looks like `build` folder was added to gitignore for no reason. I don't see anything which creates it

Extravert-ir:
Closed in favor of c42e124

PR #2535, opened 2020-04-11
Avrf progress

learn-more:
avrf progress,
adding some noisy stubs instead of nullptr's.

This allows running avrf 4.0.665.0
Not everything works yet, and we need a workaround for wine's hacked critical sections.

https://testbot.winehq.org/JobDetails.pl?Key=69391


binarymaster (approved):
There are two minor style nits (that I wouldn't post, unless you want them), and a typo in the 2nd commit: `fro` -> `for`; Otherwise, LGTM! 👍

tkreuzer (approved):

PR #2540, opened 2020-04-12
[USBCCGP] Fix some Driver Verifier issues

ThFabba:


drivers/usb/usbccgp/fdo.c
```diff
@@ -119,7 +119,7 @@ FDO_DeviceRelations(
     if (IoStack->Parameters.QueryDeviceRelations.Type != BusRelations)
     {
         /* FDO always only handles bus relations */
-        return USBCCGP_SyncForwardIrp(FDODeviceExtension->NextDeviceObject, Irp);
+        return STATUS_SUCCESS;
```
Extravert-ir, line 122: Why shouldn't we pass IRP to the next DO here?

jeditobe:
XBOX fanboys can't wait for this anymore lol

PR #2543, opened 2020-04-12
[CHARMAP] Updates to behaviour and UI

giawa:
## Purpose

This pull request updates charmap to look a bit better (removes gap at bottom of the window), removes the blank space character (0x0020) from the charmap, and also modifies the behaviour of when a larger glyph is rendered (allowing the user to select a new glyph by holding down the mouse button).  This better mimics the charmap.exe that is bundled by Microsoft.

## Proposed changes

- Remove gap where the advanced button is normally rendered when compiled with REMOVE_ADVANCED (which is the default behaviour)
- Skip over the blank space character
- Change behaviour of rendering large glyphs to allow mouse move, and to hide on double click
- Optimize search for glyph under the mouse by using the cellSize instead of PtInRect

## Before

before

## After

after

### DIFF
--- a/base/applications/charmap/charmap.c
+++ b/base/applications/charmap/charmap.c
@@ -276,6 +276,8 @@ ChangeView(HWND hWnd)
     RECT rcCharmap;
 #ifndef REMOVE_ADVANCED
     RECT rcAdvanced;
+#else
+    RECT rcCopy;
 #endif
     RECT rcPanelExt;
     RECT rcPanelInt;
@@ -284,10 +286,16 @@ ChangeView(HWND hWnd)
     UINT xPos, yPos;
     UINT Width, Height;
     UINT DeskTopWidth, DeskTopHeight;
+#ifdef REMOVE_ADVANCED
+    HWND hCopy;
+#endif
 
     GetClientRect(hCharmapDlg, &rcCharmap);
 #ifndef REMOVE_ADVANCED
     GetClientRect(hAdvancedDlg, &rcAdvanced);
+#else
+    hCopy = GetDlgItem(hCharmapDlg, IDC_COPY);
+    GetClientRect(hCopy, &rcCopy);
 #endif
     GetWindowRect(hWnd, &rcPanelExt);
     GetClientRect(hWnd, &rcPanelInt);
@@ -312,6 +320,10 @@ ChangeView(HWND hWnd)
 #ifndef REMOVE_ADVANCED
     if (Settings.IsAdvancedView)
         Height += rcAdvanced.bottom;
+#else
+    /* The lack of advanced button leaves an empty gap at the bottom of the window.
+       Shrink the window height a bit here to accomodate for that lost control. */
+    Height = rcCharmap.bottom + rcCopy.bottom + 10;
 #endif
     if ((xPos + Width) > DeskTopWidth)
         xPos += DeskTopWidth - (xPos + Width);
--- a/base/applications/charmap/map.c
+++ b/base/applications/charmap/map.c
@@ -256,7 +256,7 @@ SetFont(PMAP infoPtr,
                          GGI_MARK_NONEXISTING_GLYPHS) != GDI_ERROR)
     {
         j = 0;
-        for (i = 0; i < MAX_GLYPHS; i++)
+        for (i = ' ' + 1; i < MAX_GLYPHS; i++)
         {
             if (out[i] != 0xffff)
             {
--- a/base/applications/charmap/map.c
+++ b/base/applications/charmap/map.c
@@ -324,12 +324,12 @@ OnClick(PMAP infoPtr,
         if (infoPtr->pActiveCell)
         {
             /* invalidate normal cells, required when
-                * moving a small active cell via keyboard */
+             * moving a small active cell via keyboard */
             if (!infoPtr->pActiveCell->bLarge)
             {
                 InvalidateRect(infoPtr->hMapWnd,
-                                &infoPtr->pActiveCell->CellInt,
-                                TRUE);
+                               &infoPtr->pActiveCell->CellInt,
+                               TRUE);
             }
 
             infoPtr->pActiveCell->bActive = FALSE;
@@ -556,8 +556,8 @@ MapWndProc(HWND hwnd,
             if (wParam & MK_LBUTTON)
             {
                 OnClick(infoPtr,
-                    LOWORD(lParam),
-                    HIWORD(lParam));
+                        LOWORD(lParam),
+                        HIWORD(lParam));
             }
             break;
         }
--- a/base/applications/charmap/map.c
+++ b/base/applications/charmap/map.c
@@ -312,64 +312,38 @@ OnClick(PMAP infoPtr,
         WORD ptx,
         WORD pty)
 {
-    POINT pt;
     INT x, y;
 
-    pt.x = ptx;
-    pt.y = pty;
+    x = ptx / max(1, infoPtr->CellSize.cx);
+    y = pty / max(1, infoPtr->CellSize.cy);
 
-    for (x = 0; x < XCELLS; x++)
-    for (y = 0; y < YCELLS; y++)
+    /* if the cell is not already active */
+    if (!infoPtr->Cells[y][x].bActive)
     {
-        if (PtInRect(&infoPtr->Cells[y][x].CellInt,
-                     pt))
+        /* set previous active cell to inactive */
+        if (infoPtr->pActiveCell)
         {
-            /* if the cell is not already active */
-            if (!infoPtr->Cells[y][x].bActive)
+            /* invalidate normal cells, required when
+                * moving a small active cell via keyboard */
+            if (!infoPtr->pActiveCell->bLarge)
             {
-                /* set previous active cell to inactive */
-                if (infoPtr->pActiveCell)
-                {
-                    /* invalidate normal cells, required when
-                     * moving a small active cell via keyboard */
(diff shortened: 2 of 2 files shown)


giawa:
Okay, good to know.  It took me a few tries, but I think I've successfully added my name to those commits.

base/applications/charmap/map.c
```diff
@@ -260,6 +260,7 @@ SetFont(PMAP infoPtr,
         {
             if (out[i] != 0xffff)
             {
+                if (ch[i] == 0x0020) continue;  // skip blank space character
```
learn-more, line 263: `' '` Would be more readable (and not require a comment?)

binarymaster:
Hello! It would be nice if you also fill your full name in GitHub [profile settings](https://github.com/settings/profile), because sometimes name in the commits gets replaced with the user's nick on merging (that's known GitHub problem).

giawa:
Thanks, I've added my name in my profile as well.

base/applications/charmap/map.c
giawa: Sure, that sounds good.  I included the full 0x0020 to identify ch[i] as a 16-bit value, but the implicit cast from char would definitely be more readable.  What's the recommended way of modifying this in the PR?  Should I just add a new commit?  Or would you like me to rebase and modify this commit only to remove multiple modifications from this line?

binarymaster: If you're aware of git tricks for rebase and force push with lease, it's the best way to do it.

giawa: Perfect, sounds good!  I've made that modification and pushed it.

giawa:
Thanks for the feedback.  I've modified the commit to use ' ' instead of 0x0020.

base/applications/charmap/map.c
giawa: Ahhh cool!  I had never heard of force-with-lease.  I'll have to keep that in mind for next time I try something like this out.

base/applications/charmap/map.c
```diff
@@ -260,6 +260,7 @@ SetFont(PMAP infoPtr,
         {
             if (out[i] != 0xffff)
             {
+                if (ch[i] == ' ') continue;
```
ThFabba, line 263: This is acceptable for simplicity. But really, we know none of the characters before space are printable either, so we could just change the loop boundaries to start at `' ' + 1`.

base/applications/charmap/map.c
```diff
@@ -312,64 +313,53 @@ OnClick(PMAP infoPtr,
         WORD ptx,
         WORD pty)
 {
-    POINT pt;
     INT x, y;
 
-    pt.x = ptx;
-    pt.y = pty;
+    x = ptx / infoPtr->CellSize.cx;
+    y = pty / infoPtr->CellSize.cy;
```
ThFabba, line 319: Nice catch!

base/applications/charmap/map.c
```diff
@@ -312,64 +313,53 @@ OnClick(PMAP infoPtr,
...
+        infoPtr->pActiveCell->bLarge = TRUE;
+        if (infoPtr->hLrgWnd)
+            MoveLargeCell(infoPtr);
+        else
+            CreateLargeCell(infoPtr);
+    }
+    else
+    {
+        /* flick between large and small */
+        /*if (infoPtr->pActiveCell->bLarge)
```
ThFabba, line 352: If this is no longer needed, please remove it. If there's some reason that commenting it out is better, there should be a comment explaining why.

base/applications/charmap/charmap.c
```diff
@@ -312,6 +312,10 @@ ChangeView(HWND hWnd)
 #ifndef REMOVE_ADVANCED
     if (Settings.IsAdvancedView)
         Height += rcAdvanced.bottom;
+#else
+    /* The lack of advanced button leaves an empty gap at the bottom of the window.
+       Shrink the window height a bit here to accomodate for that lost control. */
+    Height -= 30;
```
ThFabba, line 318: Can this be calculated based on the controls? Since the dialog layout is defined in the resources, a different size could be appropriate for different languages.
Something like `IDC_CHECK_ADVANCED.Bottom - IDC_COPY.Bottom`?

base/applications/charmap/map.c
giawa: Sounds good, I've removed this code

base/applications/charmap/map.c
giawa: Sure, I've started the loop at ' ' + 1

base/applications/charmap/charmap.c
giawa: Sort of.  The advanced dialog is removed from the control.  I can find the bottom of the IDC_COPY (by getting rcCharmap.bottom + IDC_COPY.Bottom) and then add some padding, but that padding is kind of a magic number (see new commit).  Perhaps that's good enough?  Let me know what you think.

Also worth noting that I now have two assignments to Height.  The original, and then the new calculation.  I left the original for now so that DeY is used, since the warning if it isn't used will trigger a compilation error.  Otherwise there will be a bunch of code to either eliminate DeY when compiling with REMOVE_ADVANCED, or I'll have to use it with another magic number.  Something like Height = rcCharmap.bottom + rcCopy.bottom + DeY - 29

ThFabba: Seems good to me. Thanks!

giawa:
Feedback from ThFabba has been committed.  Thanks for taking a look!

base/applications/charmap/map.c
```diff
@@ -312,64 +312,38 @@ OnClick(PMAP infoPtr,
         WORD ptx,
         WORD pty)
 {
-    POINT pt;
     INT x, y;
 
-    pt.x = ptx;
-    pt.y = pty;
+    x = ptx / infoPtr->CellSize.cx;
+    y = pty / infoPtr->CellSize.cy;
```
HBelusca, line 318: Are we sure that infoPtr->CellSize.cx and cy will never ever be == 0 ?

base/applications/charmap/map.c
```diff
@@ -256,7 +256,7 @@ SetFont(PMAP infoPtr,
                          GGI_MARK_NONEXISTING_GLYPHS) != GDI_ERROR)
     {
         j = 0;
-        for (i = 0; i < MAX_GLYPHS; i++)
+        for (i = ' ' + 1; i < MAX_GLYPHS; i++)
```
HBelusca, line 259: Hmm... This should be possibly investigated more. Suppose I try to view a font for which I have printable glyphs for the first characters (0x00 to 0x20); in theory I would like to see these as well?

base/applications/charmap/charmap.c
```diff
@@ -284,10 +286,14 @@ ChangeView(HWND hWnd)
     UINT xPos, yPos;
     UINT Width, Height;
     UINT DeskTopWidth, DeskTopHeight;
+    HWND hCopy;
```
HBelusca, line 289: Only used when REMOVE_ADVANCED is defined.

base/applications/charmap/map.c
ThFabba: With such a hypothetical font, 0x20 might be a non-space printable character then? How would we tell that case apart from regular space?
If you have such a font to test with I'd be very interested to see how Windows charmap displays it. If this is purely hypothetical, then I think this solution is reasonable?

HBelusca: Yes here it was more like "something to think about" / test on Windows / find a nice solution.
For the time being what has been done here doesn't block any merging.

base/applications/charmap/map.c
giawa: The cell size is computed on WM_CREATE and is dependent on the client rect.  So cx and cy will be non-zero as long as the client rect is calculated correctly (comes from the rc I believe).  That being said, I can protect against this if you would like, and bail out if either of them are zero.

base/applications/charmap/map.c
giawa: Fair enough, and I'm not 100% sure of the answer here.  I would like to note that the loop from 0 to ' ' actually fixed more issues with some fonts having 'carriage return' as a valid glyph.  So I do like this new loop approach better than simply skipping ' '.

base/applications/charmap/map.c
HBelusca: You can manage them to be greater or equal to '1' instead.

giawa: Do you mean just doing something like this?

`x = ptx / max(1, infoPtr->CellSize.cx);`

HBelusca: yes, for example. Or alternatively fixing all the places where the CellSize.cx/cy is initialized.

giawa: Okay, what I'll probably do is implement as above (using max) for this commit.  Then I'll prepare a new pull request that can protect against CellSize containing 0 and perhaps some other charmap fixes.  Does that sound okay?

base/applications/charmap/charmap.c
giawa: Sounds good, I'll prepare an update a bit later today to address this.  Thanks for the comments!

base/applications/charmap/map.c
HBelusca: Sounds good to me!

giawa: Should be done in the latest commit.  Thanks!

giawa:
Pushed new commit to resolve issues highlighted by HBelusca.  Thanks for taking a look!

PR #2550, opened 2020-04-14
[FREELDR][INSTALL] Include installfreeldr in LiveCD and HybridCD

archeYR:
It might be useful to have installfreeldr on LiveCD/HybridCD as well.


HBelusca:
Hi since this is a freeldr tool, there is also another one worth including too:
https://git.reactos.org/?p=reactos.git;a=blob;f=boot/freeldr/fdebug/CMakeLists.txt;hb=feeeca26959404884c8b7115259f90bf58fda1bb#l5

archeYR:
> 
> 
> Hi since this is a freeldr tool, there is also another one worth including too:
> https://git.reactos.org/?p=reactos.git;a=blob;f=boot/freeldr/fdebug/CMakeLists.txt;hb=feeeca26959404884c8b7115259f90bf58fda1bb#l5

Done.

PR #2551, opened 2020-04-15
[COMCTL32 and SHELL32] Add basic snap-to-grid

giawa:
## Purpose

This pull request improves the behaviour of the listview used by the shell32 file explorer to more closely mimic what a user would expect (do not allow icons to be moved when auto-arrange is enabled).  This pull request also provides a (dumb) version of snap-to-grid for the listview control and hooks up the 'line up icons' context menu item in shell32.

## Proposed changes

- [COMCTL32 listview] Do not allow user to move icons when auto-arrange is enabled.
- [COMCTL32 listview] Implement a basic snap to grid layout scheme (moves icon to closest grid location, does not deal with stacked icons for now).
- [SHELL32] Hook up 'Line up Icons' context menu item and have it fire Arrange(LVA_SNAPTOGRID).

## TODO

I think the pull request as-is (after review and any fixes) takes a good step in the right direction, and I would like to continue to improve snap-to-grid and shell32 icon layout in general with future pull requests.  Here is what I am thinking, and would love feedback on:

- Discuss possibility of saving auto-arrange state to registry somewhere.  Where?
- Improve snap-to-grid functionality to deal with multiple icons on the same grid location.
- To perfectly mimic Windows the user should be able to move icons even when auto-arrange is enabled, they should just automatically snap to grid and the grid should be the minimum size.
- Actually auto-arrange based on different sort criteria.

## Before 'Line Up Icons'
before

## After 'Line Up Icons'
after

### DIFF
--- a/dll/win32/comctl32/listview.c
+++ b/dll/win32/comctl32/listview.c
@@ -8992,6 +8992,12 @@ static BOOL LISTVIEW_SetItemPosition(LISTVIEW_INFO *infoPtr, INT nItem, const PO
     if (!pt || nItem < 0 || nItem >= infoPtr->nItemCount ||
 	!(infoPtr->uView == LV_VIEW_ICON || infoPtr->uView == LV_VIEW_SMALLICON)) return FALSE;
 
+#ifdef __REACTOS__
+    /* FIXME:  This should really call snap to grid if auto-arrange is enabled
+       and limit the size of the grid to nItemCount elements */
+    if (is_autoarrange(infoPtr)) return FALSE;
+#endif
+
     Pt = *pt;
     LISTVIEW_GetOrigin(infoPtr, &Origin);
 
--- a/dll/win32/comctl32/listview.c
+++ b/dll/win32/comctl32/listview.c
@@ -2723,11 +2723,16 @@ static DWORD LISTVIEW_MapIndexToId(const LISTVIEW_INFO *infoPtr, INT iItem)
  * PARAMETER(S):
  * [I] infoPtr : valid pointer to the listview structure
  * [O] lpPos : will get the current icon position
+ * [I] nItem : item id to get position for
  *
  * RETURN:
  * None
  */
+#ifdef __REACTOS__
+static void LISTVIEW_NextIconPosTop(LISTVIEW_INFO *infoPtr, LPPOINT lpPos, INT nItem)
+#else
 static void LISTVIEW_NextIconPosTop(LISTVIEW_INFO *infoPtr, LPPOINT lpPos)
+#endif
 {
     INT nListWidth = infoPtr->rcList.right - infoPtr->rcList.left;
     
@@ -2749,11 +2754,16 @@ static void LISTVIEW_NextIconPosTop(LISTVIEW_INFO *infoPtr, LPPOINT lpPos)
  * PARAMETER(S):
  * [I] infoPtr : valid pointer to the listview structure
  * [O] lpPos : will get the current icon position
+ * [I] nItem : item id to get position for
  *
  * RETURN:
  * None
  */
+#ifdef __REACTOS__
+static void LISTVIEW_NextIconPosLeft(LISTVIEW_INFO *infoPtr, LPPOINT lpPos, INT nItem)
+#else
 static void LISTVIEW_NextIconPosLeft(LISTVIEW_INFO *infoPtr, LPPOINT lpPos)
+#endif
 {
     INT nListHeight = infoPtr->rcList.bottom - infoPtr->rcList.top;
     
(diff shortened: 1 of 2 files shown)


HBelusca:
For COMCTL32 it would be interesting if you could also send your patch to Wine (via the wine-devel mailing list), so that if accepted, we could have it synchronized too.

dll/win32/comctl32/listview.c
```diff
@@ -2766,7 +2769,42 @@ static void LISTVIEW_NextIconPosLeft(LISTVIEW_INFO *infoPtr, LPPOINT lpPos)
...
+    INT nMaxRows = nListHeight / infoPtr->nItemHeight;
+    POINT oldPosition;
+
+    // get the existing x and y position and then snap to the closest grid square
+    oldPosition.x = (LONG_PTR)DPA_GetPtr(infoPtr->hdpaPosX, nItem);
+    oldPosition.y = (LONG_PTR)DPA_GetPtr(infoPtr->hdpaPosY, nItem);
+
+    // FIXME: This could should deal with multiple icons in the same grid square
+    (*lpPos).x = max(0, (LONG)round((float)oldPosition.x / infoPtr->nItemWidth) * infoPtr->nItemWidth);
+    (*lpPos).y = max(0, (LONG)round((float)oldPosition.y / infoPtr->nItemHeight) * infoPtr->nItemHeight);
```
HBelusca, line 2800: I would suggest (in order to avoid floats as much as possible) the following, schematically:
```c
((oldPosition + (nItemSize/2)) / nItemSize) * nItemSize;
```
with some potential optimization of the formula.... But let's see also what the wine ppl say about this.

dll/win32/comctl32/listview.c
```diff
@@ -130,6 +130,7 @@
 #include <stdlib.h>
 #include <stdarg.h>
 #include <stdio.h>
+#include <math.h>
```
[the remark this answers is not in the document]
giawa: Exactly, this was for "round" only.

dll/win32/comctl32/listview.c
giawa: Sure, I could take a look at a method to get the equivalent of round without using floats.  I just went for the quick approach.  Is there someone at wine I should reach out to first?  Or would you like me to start making this change?

HBelusca: Sure you can write to the Wine guys, here is the mailing list: https://www.winehq.org/mailman/listinfo/wine-devel

HBelusca: Another formula:
```c
oldPosition + (nItemSize>>1)-(oldPosition+(nItemSize>>1)) % nItemSize;
```

dll/win32/comctl32/listview.c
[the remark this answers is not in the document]
giawa: I've removed this include in that latest commit.

dll/win32/comctl32/listview.c
giawa: Nice, no division!  I've updated the PR to include this suggested formula.

giawa: Nice, no division required.  I've updated the code you suggested.  I'll reach out to Wine as well.  Will my headway with wine affect the ability for me to continue working on this in reactos?

Mouvedia:
> Discuss possibility of saving auto-arrange state to registry somewhere. Where?

You should investigate how Layout.dll does it.

http://stevehealy.org/techblog/?p=162
https://www.watchingthenet.com/restore-desktop-icon-layout-when-windows-changes-lauout.html

<sup>disclaimer I dunno if these are secure<sup>

dll/win32/comctl32/listview.c
HBelusca: There is no problem sending patches to Wine as well if you find that code from them that we use in ReactOS has a bug or is incomplete. Many of the developers there already did that, and it eases code synchronizations for later.
In case you send more substantial patches targeting some Wine code they will usually ask you to write a unit test to test whether what you wrote behaves similarly than on Windows.

HBelusca: Yet another formula that works ok for negative positions:
```c
newPosition = ( newPosition = oldPosition + (1|-(oldPosition<0))*(nItemSize>>1) ) - (newPosition%nItemSize);
```
(with some inspiration from http://graphics.stanford.edu/~seander/bithacks.html#CopyIntegerSign )
But I think at the end Wine might prefer using the floats with round()...

giawa:
I'll try taking a look, but it looks like Layout.dll is a closed source Microsoft library?  Maybe someone out there has done some research on how icons were placed in older versions of Windows.  In modern versions they appear to be in HKCU\Software\Microsoft\Windows\Shell\Bags\1\Desktop

HBelusca:
Standard layout saving is managed via these so-called shell property bags (and stored in per-user registry keys as you mentioned above).
Regarding that layout.dll, this is some 3rd-party shell extension that adds shell menu entries (as shown by the screenshots in the attached links) that allows the user to explicitly control when to save the icons layout and to restore it.

giawa:
I've updated comctl32 to wrap the changes with `#ifdef __REACTOS__`.

yagoulas:
nice work

PR #2552, opened 2020-04-16
[RAPPS] Use _beginthreadex instead of CreateThread

kernelbin:
## Purpose

use _beginthreadex instead of CreateThread

## Proposed changes

use _beginthreadex instead of CreateThread

### DIFF
--- a/base/applications/rapps/loaddlg.cpp
+++ b/base/applications/rapps/loaddlg.cpp
@@ -41,6 +41,7 @@
 
 #include <rosctrls.h>
 #include <windowsx.h>
+#include <process.h>
 #undef SubclassWindow
 
 #include "rosui.h"
@@ -355,7 +356,7 @@ public:
     static VOID Add(DownloadInfo info);
     static VOID Download(const DownloadInfo& DLInfo, BOOL bIsModal = FALSE);
     static INT_PTR CALLBACK DownloadDlgProc(HWND Dlg, UINT uMsg, WPARAM wParam, LPARAM lParam);
-    static DWORD WINAPI ThreadFunc(LPVOID Context);
+    static unsigned WINAPI ThreadFunc(LPVOID Context);
     static VOID LaunchDownloadDialog(BOOL);
 };
 
@@ -433,7 +434,7 @@ INT_PTR CALLBACK CDownloadManager::DownloadDlgProc(HWND Dlg, UINT uMsg, WPARAM w
         // Start download process
         DownloadParam *param = new DownloadParam(Dlg, AppsToInstallList, szCaption);
         DWORD ThreadId;
-        HANDLE Thread = CreateThread(NULL, 0, ThreadFunc, (LPVOID) param, 0, &ThreadId);
+        HANDLE Thread = (HANDLE)_beginthreadex(NULL, 0, ThreadFunc, (void *) param, 0, (unsigned *)&ThreadId);
 
         if (!Thread)
         {
@@ -507,7 +508,7 @@ VOID CDownloadManager::UpdateProgress(
     }
 }
 
-DWORD WINAPI CDownloadManager::ThreadFunc(LPVOID param)
+unsigned WINAPI CDownloadManager::ThreadFunc(LPVOID param)
 {
     ATL::CStringW Path;
     PWSTR p, q;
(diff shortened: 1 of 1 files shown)


Fraizeraust:
How does using `_beginthreadex` improves or fixes anything? Care to provide a detailed description of what is you're trying to do?

kernelbin:
> How does using `_beginthreadex` improves or fixes anything? Care to provide a detailed description of what is you're trying to do?

A thread created by CreateThread may cause some potential bugs when calling some C/C++ runtime functions. _beginthread and _beginthreadex will use Thread Local Storage to protect the thread.

And using CreateThread is also not suggested by msdn. 
see this: 

https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createthread#remarks

「A thread in an executable that calls the C run-time library (CRT) should use the _beginthreadex and _endthreadex functions for thread management rather than CreateThread and ExitThread;」

base/applications/rapps/loaddlg.cpp
```diff
@@ -433,7 +434,7 @@ INT_PTR CALLBACK CDownloadManager::DownloadDlgProc(HWND Dlg, UINT uMsg, WPARAM w
         // Start download process
         DownloadParam *param = new DownloadParam(Dlg, AppsToInstallList, szCaption);
         DWORD ThreadId;
-        HANDLE Thread = CreateThread(NULL, 0, ThreadFunc, (LPVOID) param, 0, &ThreadId);
+        HANDLE Thread = (HANDLE)_beginthreadex(NULL, 0, ThreadFunc, (void *) param, 0, (unsigned *)&ThreadId);
```
learn-more, line 437: If you use the correct type for ThreadId, there is no need to cast that.

HBelusca:
Using the CRT beginthread & co is only useful if, from inside it, you are using CRT functions that may use shared resources but at the same time do not support reentrance, like... printf & the like.
If the thread does nothing but use those CRT functions memcpy, strlen etc. then there is no gain of using beginthread.

Also MSDN are at best innacurate, at most written by interns. There are for example articles telling you that calling a documented native function is "obsolete and should be replaced by whatever Win32 counterpart function", which makes obviously nonsense because: 1. if you already wrote a pure win32 app you certainly used that win32 function, and 2. if you write instead a native app, you cannot use win32 functions anyway.

kernelbin:
> Using the CRT beginthread & co is only useful if, from inside it, you are using CRT functions that may use shared resources but at the same time do not support reentrance, like... printf & the like.
> If the thread does nothing but use those CRT functions memcpy, strlen etc. then there is no gain of using beginthread.
> 
> Also MSDN are at best innacurate, at most written by interns. There are for example articles telling you that calling a documented native function is "obsolete and should be replaced by whatever Win32 counterpart function", which makes obviously nonsense because: 1. if you already wrote a pure win32 app you certainly used that win32 function, and 2. if you write instead a native app, you cannot use win32 functions anyway.

Yes, you are right. But we don't know what will be used in this thread in the future. If someone uses some function like that (it's possible because this is a C++ program), and don't know the thread is created by CreateThread, it will cost a lot of time to find the bug. So I think always use _beginthreadex instead of CreateThread in C/C++ code is a good habit.

base/applications/rapps/loaddlg.cpp
```diff
@@ -355,7 +356,7 @@ class CDownloadManager
     static VOID Add(DownloadInfo info);
     static VOID Download(const DownloadInfo& DLInfo, BOOL bIsModal = FALSE);
     static INT_PTR CALLBACK DownloadDlgProc(HWND Dlg, UINT uMsg, WPARAM wParam, LPARAM lParam);
-    static DWORD WINAPI ThreadFunc(LPVOID Context);
+    static unsigned WINAPI ThreadFunc(LPVOID Context);
```
SergeGautherie, line 359: `unsigned int`, to be explicit?

learn-more:
> > Using the CRT beginthread & co is only useful if, from inside it, you are using CRT functions that may use shared resources but at the same time do not support reentrance, like... printf & the like.
> > If the thread does nothing but use those CRT functions memcpy, strlen etc. then there is no gain of using beginthread.
> > Also MSDN are at best innacurate, at most written by interns. There are for example articles telling you that calling a documented native function is "obsolete and should be replaced by whatever Win32 counterpart function", which makes obviously nonsense because: 1. if you already wrote a pure win32 app you certainly used that win32 function, and 2. if you write instead a native app, you cannot use win32 functions anyway.
> 
> Yes, you are right. But we don't know what will be used in this thread in the future. If someone uses some function like that (it's possible because this is a C++ program), and don't know the thread is created by CreateThread, it will cost a lot of time to find the bug. So I think always use _beginthreadex instead of CreateThread in C/C++ code is a good habit.

I was going to type something like this.
Indeed in a case like this, it certainly does do no harm (afaik) to use _beginthreadex.

PR #2553, opened 2020-04-16
[USETUP][MUI] Loop over next MUI entries with the same ID

GeoB99:
Currently it's WIP, more to come.

### DIFF
--- a/base/setup/usetup/mui.c
+++ b/base/setup/usetup/mui.c
@@ -325,6 +325,7 @@ MUIClearText(
     IN INT TextID)
 {
     const MUI_ENTRY * entry;
+    ULONG Index = 0;
 
     /* Get the MUI entry */
     entry = MUIGetEntry(Page, TextID);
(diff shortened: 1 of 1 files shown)


tkreuzer:
Is this still alive?

binarymaster:
> Is this still alive?

Needs review.

learn-more:
Considering this PR did not get a review yet, can you please elaborate a bit on why you changed this?

Fraizeraust:
On some locales, take German as an example, some text string IDs have to be used twice to denote a single text description string, in this case of format partition prompt.
https://github.com/reactos/reactos/blob/master/base/setup/usetup/lang/de-DE.h#L1310
And the commit to add further information into the context -- https://github.com/reactos/reactos/commit/7ed0701983125b64bd6895ca6244f212c24db84d

With that said, currently the code doesn't loop over other existing IDs which is what this PR is aiming for. With this, MUI functions such as `MUIClearStyledText` can remove text strings properly of the same ID.

Here's a screenshot of what roughly happens, on clean master:
Capture

HBelusca (commented):
Note that this code supposes that, once the entry with the given ID has been found (with MUIGetEntry), this entry, **including all the next ones that may or may not have the same ID** need to be cleared.
Is it what is really intended?

GeoB99:
>or may not have the same ID

Wrong. The ID that we're talking about corresponds to the text string ID of the specific entry, where `MUIGetEntry` returns the address of the entry given the correct text ID.
See https://github.com/reactos/reactos/blob/2013064a34e1aeb18f68fe108531fd2515203e25/base/setup/usetup/mui.c#L294
And https://github.com/reactos/reactos/blob/master/base/setup/usetup/lang/de-DE.h#L1306

With that said, the code in this patch loops the entries based on the ID given that is the text ID itself. As there are two ID entries of `TEXT_ID_FORMAT_PROMPT` in German locale, the code loops, clears the text and increments the index, until there's no entry left with `TEXT_ID_FORMAT_PROMPT`.

HBelusca (approved):

PR #2575, opened 2020-04-18
[LOADCONFIG_APITEST] Tell GCC that the .def is not just included for the nice comments

learn-more:


modules/rostests/apitests/loadconfig/CMakeLists.txt
```diff
@@ -13,4 +13,10 @@ add_importlibs(loadconfig_apitest msvcrt kernel32 ntdll)
 #add_pch(loadconfig_apitest loadconfig.h SOURCE)
 add_rostests_file(TARGET loadconfig_apitest)
 
+# Tell GCC again that we are really interested in exporting symols (who would have figured that we want to use the .def we specified before?)
```
SergeGautherie, line 16: Typo: 'symols'.

Add an "if(GCC)"?

PR #2582, opened 2020-04-18
[LWIP] Attempt to fix CORE-16868

sdever:
## Purpose

Try to fix CORE-16868 by calling tcp_close when we get shutdown with both tx and rx enabled.

JIRA issue: [CORE-16868](https://jira.reactos.org/browse/CORE-16868)

### DIFF
--- a/sdk/lib/drivers/lwip/src/rostcp.c
+++ b/sdk/lib/drivers/lwip/src/rostcp.c
@@ -647,11 +647,18 @@ LibTCPShutdownCallback(void *arg)
      * PCB without telling us if we shutdown TX and RX. To avoid these problems, we'll clear the
      * socket context if we have called shutdown for TX and RX.
      */
-    if (msg->Input.Shutdown.shut_rx) {
-        msg->Output.Shutdown.Error = tcp_shutdown(pcb, TRUE, FALSE);
+    if (msg->Input.Shutdown.shut_rx != msg->Input.Shutdown.shut_tx) {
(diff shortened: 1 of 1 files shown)


sdk/lib/drivers/lwip/src/rostcp.c
```diff
@@ -647,11 +647,18 @@ LibTCPShutdownCallback(void *arg)
      * PCB without telling us if we shutdown TX and RX. To avoid these problems, we'll clear the
      * socket context if we have called shutdown for TX and RX.
      */
-    if (msg->Input.Shutdown.shut_rx) {
-        msg->Output.Shutdown.Error = tcp_shutdown(pcb, TRUE, FALSE);
+    if (msg->Input.Shutdown.shut_rx != msg->Input.Shutdown.shut_tx) {
```
learn-more, line 650: This this also hold up if both are `0` ?

sdever: If both are 0, nothing should be done I think (and there should be a test case for this ^^)

KyleKatarn31:
I've been testing this PR with latest release of SUMo (5.10.16) and it solves ALL TCP-relatest exceptions without any noticeable regression.
Very good work. I strongly support integration of this PR.

sdk/lib/drivers/lwip/src/rostcp.c
sdever: Yeah could be an option, this case shouldn't even happen

PR #2589, opened 2020-04-19
Add Search Hidden Files and Folders Registry Entry for HKCU

Doug-Lyons:
## Add Missing Explorer - SearchHidden Registry Entry in HKCU

_Update default ReactOS HKCU file._

JIRA issue: [CORE-16882](https://jira.reactos.org/browse/CORE-16882)

### DIFF
--- a/boot/bootdata/hivedef.inf
+++ b/boot/bootdata/hivedef.inf
@@ -1883,6 +1883,7 @@ HKCU,"SOFTWARE\Classes\Applications",,0x00000012
 HKCU,"SOFTWARE\Microsoft\Windows",,0x00000012
 HKCU,"SOFTWARE\Microsoft\Windows\CurrentVersion",,0x00000012
 HKCU,"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer",,0x00000012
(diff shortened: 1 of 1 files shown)


KyleKatarn31:
I finally have a doubt that Windows XP creates these keys at 0x0 prior to 1st use of the function. Have you checked ?
Not a big issue anyway...

Doug-Lyons:
@yagoulas, Thank you for helping with the commit. Doug Lyons

Doug-Lyons:
> I finally have a doubt that Windows XP creates these keys at 0x0 prior to 1st use of the function. Have you checked ?
> Not a big issue anyway...

@KyleKatarn31, Actually I did check and you are correct that on a clean Windows XP install, this registry entry does not appear until you have used the "Search" button the first time, but I think having it prior is actually better, like we have it now. Thanks.

KyleKatarn31:
> 
> 
> > I finally have a doubt that Windows XP creates these keys at 0x0 prior to 1st use of the function. Have you checked ?
> > Not a big issue anyway...
> 
> @KyleKatarn31, Actually I did check and you are correct that on a clean Windows XP install, this registry entry does not appear until you have used the "Search" button the first time, but I think having it prior is actually better, like we have it now. Thanks.

I agree !

PR #2590, opened 2020-04-19
[BROWSEUI] Attempt absolute path on relative path fail

giawa:
## Purpose

This PR will attempt to parse an absolute path even in the case of an unexpected failure while computing the relative path of a directory entered into the address bar of the file explorer.  An unexpected error could occur when `BindToObject` is unimplemented, or when `BindToObject` was attempted on the Desktop.

## Proposed changes

- Attempt to parse an absolute path as long as `psfDesktop` is valid.

## More Information

To reproduce this issue:

1.  Boot up a copy of ReactOS and navigate the file explorer to the desktop.
2.  Try to enter a valid path and navigate to it using the address bar.
3.  You will be unable to navigate using the address bar.  The error being returned is E_INVALIDARG by `BindToObject` because the Desktop is the current path.

stuckondesktop

### DIFF
--- a/dll/win32/browseui/addresseditbox.cpp
+++ b/dll/win32/browseui/addresseditbox.cpp
@@ -146,11 +146,11 @@ HRESULT STDMETHODCALLTYPE CAddressEditBox::ParseNow(long paramC)
 
     hr = pbs->GetPidl(&pidlCurrent);
     if (FAILED_UNEXPECTEDLY(hr))
-        goto cleanup;
+        goto parseabsolute;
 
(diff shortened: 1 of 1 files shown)


Mouvedia:
Please check CORE-10989 and CORE-9120 once you are done with this.

giawa:
> 
> 
> Please check CORE-10989 and CORE-9120 once you are done with this.

Thanks for pointing those out!  I've submitted a PR for CORE-9120 separately, and will also look at CORE-10989.

PR #2594, opened 2020-04-19
[RAPPS] Proxy fields default state fix (CORE-16853)

KyleKatarn31:
## Purpose

_ When a proxy was selected then "default settings" is applied, the state of the proxy text fields is unduly kept enabled because the test code in the dialog initialization section is only able to "re-enable" these fields with the wrong assumption that they have been disabled beforehand

JIRA issue: [CORE-16853](https://jira.reactos.org/browse/CORE-16853)

## Proposed changes

_  Change in InitSettingsControls in order to Disable these 2 text fields when not necessary.

### DIFF
--- a/base/applications/rapps/settingsdlg.cpp
+++ b/base/applications/rapps/settingsdlg.cpp
@@ -71,13 +71,13 @@ namespace
 
         if (Info->Proxy == 0)
         {
-        	EnableWindow(GetDlgItem(hDlg, IDC_PROXY_SERVER), TRUE);
-        	EnableWindow(GetDlgItem(hDlg, IDC_NO_PROXY_FOR), TRUE);
+            EnableWindow(GetDlgItem(hDlg, IDC_PROXY_SERVER), TRUE);
+            EnableWindow(GetDlgItem(hDlg, IDC_NO_PROXY_FOR), TRUE);
         }
         else
         {
-        	EnableWindow(GetDlgItem(hDlg, IDC_PROXY_SERVER), FALSE);
-        	EnableWindow(GetDlgItem(hDlg, IDC_NO_PROXY_FOR), FALSE);
+            EnableWindow(GetDlgItem(hDlg, IDC_PROXY_SERVER), FALSE);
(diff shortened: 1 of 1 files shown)


base/applications/rapps/settingsdlg.cpp
```diff
@@ -69,15 +69,15 @@ namespace
 
         CheckRadioButton(hDlg, IDC_PROXY_DEFAULT, IDC_USE_PROXY, IDC_PROXY_DEFAULT + Info->Proxy);
 
-		if (IDC_PROXY_DEFAULT + Info->Proxy == IDC_USE_PROXY)
+        if (IDC_PROXY_DEFAULT + Info->Proxy == IDC_USE_PROXY)
```
sdever: I'm not a big fan about this, I think this would be more readable.
```suggestion
        if (Info->Proxy == 0)
```
And as a bonus suggestion, you can define an enum to get rid of those magic consts :P

sdever: Also, instead of creating a commit to fix indentation, you'd better
```
git add <your files>
git commit --amend
```
to merge your new modifications into the existing commit (and then force-push)

KyleKatarn31: ok

KyleKatarn31: I can't add you change (marked "outdated"). Is it intentionnal ?

sdever: The last commit didn't have the line where I wanted to suggest changes (hence my suggestion to amend your commit instead creating a new one for indentation fixes)

KyleKatarn31: ok

KyleKatarn31:
@sdever : Should be OK now

base/applications/rapps/settingsdlg.cpp
```diff
@@ -71,8 +71,13 @@ namespace
 
         if (IDC_PROXY_DEFAULT + Info->Proxy == IDC_USE_PROXY)
```
sdever: Still the old code

PR #2598, opened 2020-04-19
[EXPLORER] Keep processing messages while waiting for a startup task

learn-more:
JIRA issue: [CORE-16909](https://jira.reactos.org/browse/CORE-16909)

### DIFF
--- a/base/shell/explorer/startup.cpp
+++ b/base/shell/explorer/startup.cpp
@@ -85,8 +85,37 @@ static int runCmd(LPWSTR cmdline, LPCWSTR dir, BOOL wait, BOOL minimized)
     TRACE("Successfully ran command\n");
 
     if (wait)
-    {   /* wait for the process to exit */
-        WaitForSingleObject(info.hProcess, INFINITE);
+    {
+        HANDLE Handles[] = { info.hProcess };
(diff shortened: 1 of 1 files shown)


base/shell/explorer/startup.cpp
```diff
@@ -85,8 +85,37 @@ static int runCmd(LPWSTR cmdline, LPCWSTR dir, BOOL wait, BOOL minimized)
...
+        DWORD nCount = _countof(Handles);
+        DWORD dwWait;
+        MSG msg;
+
+        /* wait for the process to exit */
+        for (;;)
+        {
+            /* We need to keep processing messages,
+               otherwise we will hang anything that is trying to send a message to us */
+            dwWait = MsgWaitForMultipleObjects(nCount, Handles, FALSE, INFINITE, QS_ALLINPUT);
```
HBelusca, line 99: Why not just using
```c
MsgWaitForMultipleObjects(1, &info.hProcess, FALSE, INFINITE, QS_ALLINPUT);
```
since you're waiting on a single handle?

learn-more: I like this better.

PR #2601, opened 2020-04-19
[KDCOM][NTOS:INBV][FREELDR] Decrease default baud rate for PC-98

disean:
## Purpose

Attempt to fix bug described in that Twitter post: https://twitter.com/simk98l/status/1251340347491282945.

## Proposed changes

According to PC-9801 Bible p. 50, divisor for PIT will become unsupported in some cases after having removed the fractional part. 
- Replace 19200 value with 9600 which is supported by both 10 MHz and 8 MHz machines.

## TODO

- [x] Make a real hardware test (with screen debug enabled) to confirm that patch works.

### DIFF
--- a/boot/freeldr/freeldr/lib/debug.c
+++ b/boot/freeldr/freeldr/lib/debug.c
@@ -43,8 +43,14 @@ static UCHAR DbgChannels[DBG_CHANNELS_COUNT];
 ULONG DebugPort = RS232;
 
 /* Serial debug connection */
-ULONG ComPort  = 0; // The COM port initializer chooses the first available port starting from COM4 down to COM1.
+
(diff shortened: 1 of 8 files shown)


boot/freeldr/freeldr/lib/debug.c
```diff
@@ -43,8 +43,14 @@ static UCHAR DbgChannels[DBG_CHANNELS_COUNT];
 ULONG DebugPort = RS232;
 
 /* Serial debug connection */
-ULONG ComPort  = 0; // The COM port initializer chooses the first available port starting from COM4 down to COM1.
+
```
binarymaster, line 46:
```suggestion
```

PR #2614, opened 2020-04-21
[SYSDM] User Profiles fix of OnDestroy (CORE-16921)

KyleKatarn31:
## Purpose

_ User Profiles window incorrectly release Heap zone in its OnDestroy() function, leading to undue closure of its parent Window

JIRA issue: [CORE-16921](https://jira.reactos.org/browse/CORE-16921)

## Proposed changes

_ Correct initialization of "Item" variable before using it
_ Protection against invalid Handle usage
_ Consistency of variable naming accross the window

### DIFF
--- a/dll/cpl/sysdm/userprofile.c
+++ b/dll/cpl/sysdm/userprofile.c
@@ -828,10 +828,8 @@ OnDestroy(
         Item.iSubItem = 0;
         if (ListView_GetItem(hwndListView, &Item))
         {
-            if (Item.lParam != 0)
-            {
+            if (Item.lParam != 0)            
                 HeapFree(GetProcessHeap(), 0, (LPVOID)Item.lParam);
-            }
         }
     }
 }
--- a/dll/cpl/sysdm/userprofile.c
+++ b/dll/cpl/sysdm/userprofile.c
@@ -820,7 +820,6 @@ OnDestroy(
         return;
     
     nItems = ListView_GetItemCount(hwndListView);
-
     for (i = 0; i < nItems; i++)
     {
         ZeroMemory(&Item, sizeof(LVITEM));    
--- a/dll/cpl/sysdm/userprofile.c
+++ b/dll/cpl/sysdm/userprofile.c
@@ -828,7 +828,7 @@ OnDestroy(
         Item.iSubItem = 0;
         if (ListView_GetItem(hwndListView, &Item))
         {
-            if (Item.lParam != 0)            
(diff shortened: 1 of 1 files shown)


katahiromz:
Using `LVN_DELETEITEM` notification is the better way.

KyleKatarn31:
> Using `LVN_DELETEITEM` notification is the better way.

I understand. However, i'd prefer to "maintain" the current code design principle and "fix what is broken" rather than redesigning with WM_NOTIFY + LVN_DELETEITEM.

Would you accept to keep current structure (fixed as propose), or would you only accept a redesigned code based on WM_NOTIFY with LVN_DELETEITEM ?

dll/cpl/sysdm/userprofile.c
```diff
@@ -645,6 +645,7 @@ UpdateButtonState(
         iSelected = ListView_GetNextItem(hwndListView, -1, LVNI_SELECTED);
         if (iSelected != -1)
         {
+            ZeroMemory(&Item, sizeof(LVITEM));
             Item.mask = LVIF_PARAM;
             Item.iItem = iSelected;
             Item.iSubItem = 0;
```
SergeGautherie, line 651: This line becomes redundant.

KyleKatarn31: Why redundant ? It allows the proper initialization to 0 of all fields but the ones used for search. Am I wrong ?

KyleKatarn31: Oups, my mistake.
Indeed "Item.iSubItem = 0;" is useless after a ZeroMemory. This code pattern is used 4 times in this file (2 times already like this prior to the change).
Would you suggest that I remove the line setting iSubItem to 0 all 4 times ?

KyleKatarn31: Oups, my mistake.
Indeed "Item.iSubItem = 0;" is useless after a ZeroMemory. This code pattern is used 4 times in this file (2 times already like this prior to the change).
Would you suggest that I remove the line setting iSubItem to 0 all 4 times ?

KyleKatarn31:
@SergeGautherie : Now hat you comments are integrated, OK for approval ?

KyleKatarn31:
@katahiromz @sdever : New implementation based on LVN_DELETEITEM as suggested.

Tested OK

If OK for you, would you please approve and merge ?

KyleKatarn31:
Thanks !

PR #2607, opened 2020-04-20
[MEDIA][FONTS] Fix French glyphs of Tahoma font

katahiromz:
## Purpose
Fix French glyphs (`U+00EC`, `U+00EE`, and `U+00EF`) of Tahoma font.
Version 0.018 khmz

JIRA issue: [CORE-16899](https://jira.reactos.org/browse/CORE-16899)

AFTER:
after

PR #2615, opened 2020-04-21
[GDI32_APITEST] Add Tahoma font to GetTextMetrics apitest set

Doug-Lyons:
## Add Tahoma font to GetTextMetrics apitest set.

_Tahoma is available in Windows and ReactOS both, so this allows us to compare these fonts in both systems._

### DIFF
--- a/modules/rostests/apitests/gdi32/GetTextMetrics.c
+++ b/modules/rostests/apitests/gdi32/GetTextMetrics.c
@@ -373,6 +373,57 @@ static const TEST_ENTRY g_MSMINCHO[] =
     { __LINE__, -400, 0, 3000, 400, 400, 344, 56, 0, 0 },
 };
 #define g_MSMINCHO_count _countof(g_MSMINCHO)
+
+static const TEST_ENTRY g_TAHOMA[] =
+{
+    { __LINE__, 400, 0, 1953, 400, 400, 332, 68, 0, 0 },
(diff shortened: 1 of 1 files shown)


KyleKatarn31:
Please keep in mind that some glyphs are mising in ROS's Tahoma compared to Windows
image

But your PR is interesting in particular with regard to the recent issues faced here : https://github.com/reactos/reactos/pull/2607

modules/rostests/apitests/gdi32/GetTextMetrics.c
```diff
@@ -464,6 +517,8 @@ START_TEST(GetTextMetrics)
         {
             const TEST_ENTRY *test = &font->tests[k];
 
+            printf("Testing '%s'.\n", font->entry_name);
```
learn-more, line 520: please use `trace` here, which is a builtin winetest function (and can be configured with test parameters)

learn-more:
https://testbot.winehq.org/JobDetails.pl?Key=70155

modules/rostests/apitests/gdi32/GetTextMetrics.c
Doug-Lyons: Done and moved to a better location as well. Thanks.

Doug-Lyons:
Thanks for the commit, @katahiromz.

PR #2622, opened 2020-04-22
[USETUP] Update now-redirected 'http' and 'www.reactos.org' URLs

SergeGautherie:
NB:
http://www.reactos.org/uk/
is not updated, as it returns
`Error 404 - Page not found`.


Saibamen:
In the past, old ROS website have translations like http://www.reactos.org/pl/
Please update those addresses too

PR #2625, opened 2020-04-22
[SHELL32] Fix for incorrect init of hide/show hidden files & folders (CORE-16932)

KyleKatarn31:
## Purpose

_ In current implementation, init value for hide/show hidden files & folders is set to 1 (Show) while "default" button sets it to 2 (Hide). "Default" button should set to "Show" (1).

JIRA issue: [CORE-16932](https://jira.reactos.org/browse/CORE-16932)

## Proposed changes

_ Change of Advanced Files & Folders Settings action of Default button to select "Show"
_ Change is made by flipping the HKLM default values

image

### DIFF
--- a/boot/bootdata/hivedef.inf
+++ b/boot/bootdata/hivedef.inf
@@ -1895,9 +1895,9 @@ HKCU,"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\CLSID\{20D04FE0-3AEA-10
 HKCU,"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\Advanced","ListviewShadow",0x00010003,0x00000001
 HKCU,"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\Advanced","HideFileExt",0x00010003,0x00000000
 HKCU,"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\Advanced","StartMenuLogoff",0x00010003,0x00000001
-HKCU,"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\Advanced","Hidden",0x00010003,1 
-HKCU,"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\Advanced","ShowSuperHidden",0x00010003,0
 ; "Hidden" to be changed to 2 if we later want to have "Hide hidden files by default"
+HKCU,"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\Advanced","Hidden",0x00010003,1
+HKCU,"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\Advanced","ShowSuperHidden",0x00010003,0
 
 ; ComDlg32
 HKCU,"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\ComDlg32",,0x00000012
--- a/boot/bootdata/hivedef.inf
+++ b/boot/bootdata/hivedef.inf
@@ -1895,8 +1895,9 @@ HKCU,"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\CLSID\{20D04FE0-3AEA-10
 HKCU,"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\Advanced","ListviewShadow",0x00010003,0x00000001
 HKCU,"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\Advanced","HideFileExt",0x00010003,0x00000000
 HKCU,"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\Advanced","StartMenuLogoff",0x00010003,0x00000001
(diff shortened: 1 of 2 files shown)


Extravert-ir:
I'd like the default to remain as it is

KyleKatarn31:
> 
> 
> I'd like the default to remain as it is

Reason for change :
Current init :
- Is not the safest option for users and is not aligned with Windows practices
- Is not aligned with the "Default" button on the advanced settings page (Files & Folders options)

learn-more:
We are nowhere near a place that we can update features to be convenient for users.
At the moment it makes sense that they are convenient for developers.

KyleKatarn31:
> 
> 
> We are nowhere near a place that we can update features to be convenient for users.
> At the moment it makes sense that they are convenient for developers.

Yes. That's the purpose of the "new" proposed design.

SergeGautherie:
Add a comment somewhere, to remember to change it back in the future?

KyleKatarn31:
> 
> 
> Add a comment somewhere, to remember to change it back in the future?

I thought that you all wanted this to be "definitive". Do you want me to refer PR+CORE-xxxx along with the impacted lines in comments?

KyleKatarn31:
> 
> 
> Add a comment somewhere, to remember to change it back in the future?

Comments added. Would you please Review & Approve if OK ?

boot/bootdata/hivedef.inf
```diff
@@ -1895,8 +1895,9 @@ HKCU,"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\CLSID\{20D04FE0-3AEA-10
 HKCU,"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\Advanced","ListviewShadow",0x00010003,0x00000001
 HKCU,"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\Advanced","HideFileExt",0x00010003,0x00000000
 HKCU,"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\Advanced","StartMenuLogoff",0x00010003,0x00000001
-HKCU,"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\Advanced","Hidden",0x00010003,1
+HKCU,"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\Advanced","Hidden",0x00010003,1
```
SergeGautherie, line 1898: Why add an ending space?

boot/bootdata/hivesft.inf
```diff
@@ -317,19 +317,21 @@ HKLM,"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\Advanced\Folder\Hidden\
 HKLM,"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\Advanced\Folder\Hidden\NOHIDDEN","Type",0x00000000,"radio"
 HKLM,"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\Advanced\Folder\Hidden\NOHIDDEN","CheckedValue",0x00010001,0x00000002
 HKLM,"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\Advanced\Folder\Hidden\NOHIDDEN","ValueName",0x00000000,"Hidden"
-HKLM,"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\Advanced\Folder\Hidden\NOHIDDEN","DefaultValue",0x00010001,0x00000002
+HKLM,"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\Advanced\Folder\Hidden\NOHIDDEN","DefaultValue",0x00010001,0x00000001
 HKLM,"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\Advanced\Folder\Hidden\NOHIDDEN","HKeyRoot",0x00010001,0x80000001
 HKLM,"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\Advanced\Folder\Hidden\NOHIDDEN","HelpID",0x00000000,"shell.hlp#51104"
+; "DefaultValue" to be changed to 2 if we later want to have "Hide hidden files by default"
```
SergeGautherie, line 323: Comments are usually inserted just before the line they apply to, don't they?

boot/bootdata/hivedef.inf
KyleKatarn31: OK, i fix

boot/bootdata/hivesft.inf
KyleKatarn31: OK, i fix

Extravert-ir:
Unless we are not going to add documentation for all registry variables here, I don't see a point in having comments only for some of them.
Default settings is actually distributive-dependent. Windows Server and Workstation have different settings, for example

KyleKatarn31:
> 
> 
> Unless we are not going to add documentation for all registry variables here, I don't see a point in having comments only for some of them.
> Default settings is actually distributive-dependent. Windows Server and Workstation have different settings, for example

I understood the request from other dev as a must since we do something one way while ReactOS is mainly for dev while on the long run we want to be able to come back to the other way with minimum efforts.

learn-more:
For the one merging this: Please note that the commit message needs to be fixed...

PR #2626, opened 2020-04-22
[SHELL32] Correctly fail if shortcut target not found

katahiromz:
## Purpose
Dereferencing the shell link of invalid target must fail correctly.
JIRA issue: [CORE-16816](https://jira.reactos.org/browse/CORE-16816)

BEFORE:
before
It won't fail and file operation occurs.

AFTER:
after
It fails. File operation won't happen.

### DIFF
--- a/dll/win32/shell32/CShellLink.cpp
+++ b/dll/win32/shell32/CShellLink.cpp
@@ -3162,7 +3162,21 @@ HRESULT STDMETHODCALLTYPE CShellLink::Drop(IDataObject *pDataObject,
     DWORD dwKeyState, POINTL pt, DWORD *pdwEffect)
 {
     TRACE("(%p)\n", this);
-    HRESULT hr = S_OK;
+    HRESULT hr;
+
+    WCHAR szPath[MAX_PATH];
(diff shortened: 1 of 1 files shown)


katahiromz:
WinXP:
resolvepath1-winxp
resolvepath2-winxp

katahiromz:
The UI of these dialogs seems implemented in `IShellLink::Resolve`.

HBelusca:
Here is the implementation of this interface's method:
https://git.reactos.org/?p=reactos.git;a=blob;f=dll/win32/shell32/CShellLink.cpp;hb=5a5ba1ad107d1af37241d444c567abc6f7cb6eb3#l1390

As the documentation tells:
https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-ishelllinka-resolve

katahiromz:
Do you know Levenshtein distance?

katahiromz:
> Nearest match based on size, date, and type

Levenshtein distance is unrelated?

katahiromz:
Lacking materials... Closed.

PR #2627, opened 2020-04-22
[TFTPD] Fixes possible out of bound access in tftpd

qarmin:
## Purpose

Just small fix for possible array overflow(array have only 32 elements)
https://github.com/reactos/reactos/blob/893a3c9d030fd8b078cbd747eeefd3f6ce57e560/base/services/tftpd/tftpd.h#L146

### DIFF
--- a/base/services/tftpd/tftpd.cpp
+++ b/base/services/tftpd/tftpd.cpp
@@ -543,7 +543,7 @@ void processRequest(void *lpParam)
 #ifdef __REACTOS__
                 for (MYWORD j = 0; j < _countof(cfig.hostRanges) && cfig.hostRanges[j].rangeStart; j++)
 #else
-                for (int j = 0; j <= 32 && cfig.hostRanges[j].rangeStart; j++)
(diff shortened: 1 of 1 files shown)


base/services/tftpd/tftpd.cpp
```diff
@@ -543,7 +543,7 @@ void processRequest(void *lpParam)
 #ifdef __REACTOS__
                 for (MYWORD j = 0; j < _countof(cfig.hostRanges) && cfig.hostRanges[j].rangeStart; j++)
 #else
-                for (int j = 0; j <= 32 && cfig.hostRanges[j].rangeStart; j++)
+                for (int j = 0; j < 32 && cfig.hostRanges[j].rangeStart; j++)
```
SergeGautherie, line 546: Did you miss to read line 543?

qarmin: Oh, this is thirdparty code?

Extravert-ir: Probably a very old one, but yeah, it's treated like that.
So the part inside `#ifdef __REACTOS__` compiles, which is fine

PR #2630, opened 2020-04-22
[ZIPFLDR][ZIPFLDR_APITEST] Fix date on extraction, add tests for new features

learn-more:
Needed to test the 'EnumParentDir' stuff,
Had the rest already laying around.

[CORE-16908](https://jira.reactos.org/browse/CORE-16908)


learn-more:
@yagoulas main focus point is the `EnumParentDir` test.
The rest is for later

yagoulas:
So if I understand it right we return the correct attributes for the zip file, right?

learn-more:
> So if I understand it right we return the correct attributes for the zip file, right?

As discussed earlier today:
We are very far from returning the correct attributes,
it is even returned as file instead of as folder atm.

learn-more:
This test needs to be updated slightly:
Instead of hardcoding the order of files, it needs to take into account that filesystem can return out of order.

PR #2633, opened 2020-04-22
[README] Update now-redirected 'www.reactos.org' URLs

SergeGautherie:
NB:
https://www.reactos.org/joining/faqs
is not updated, as it returns
`Error 404 - Page not found`.

### DIFF
--- a/boot/bootdata/readme.txt
+++ b/boot/bootdata/readme.txt
@@ -10,7 +10,7 @@ ReactOS™ is an Open Source effort to develop a quality operating system that i
 compatible with applications and drivers written for the Microsoft® Windows™ NT
 family of operating systems (NT4, 2000, XP, 2003, Vista, Seven).
 
-More information is available at: https://www.reactos.org
+More information is available at: https://reactos.org/
 
 
 2. Relationship with the WINE project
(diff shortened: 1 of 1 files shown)


boot/bootdata/readme.txt
```diff
@@ -10,7 +10,7 @@ ReactOS™ is an Open Source effort to develop a quality operating system that i
 compatible with applications and drivers written for the Microsoft® Windows™ NT
 family of operating systems (NT4, 2000, XP, 2003, Vista, Seven).
```
KyleKatarn31, line 11: Not necessarily true beyond XP (eg : some Vista specific API are not supported, limiting supported applications. Current wording may suggest that any application written for these systems will run in ReactOS

I know it's not the core purpose of the PR but might deserve a fix...

boot/bootdata/readme.txt
```diff
@@ -66,7 +66,7 @@ ReactOS - amongst other topics. Contributors to the project are always welcome.
 -----------------------------
 
 Some of your questions may be answered in: https://www.reactos.org/joining/faqs
```
binarymaster, line 68:
```suggestion
Some of your questions may be answered in: https://reactos.org/faq/
```

boot/bootdata/readme.txt
SergeGautherie: > Current wording [...] will [...]

That is worded as a goal, so future, hence all right.
(Although I agree it can be a bit misleading, as current alpha state is obviously not there yet.)

> I know it's not the core purpose of the PR but might deserve a fix...

Rewriting the text is not my purpose at all!
Please, avoid such unrelated comments.

PR #2634, opened 2020-04-22
[FREELDR] Update now-redirected 'www.reactos.org' URLs

SergeGautherie:


Prince-Kassad:
The code this is used in is leftover from pre-0.3.0 ReactOS and is only called if you manually change MinimalUI from Yes to No in freeldr.ini. Do we want to keep this, given we're already hitting size limits for freeldr on FAT volumes...?

SergeGautherie:
> This is a visual thing, not a clickable link,

Even so, I see no reason not to update it. Your choice anyway.

> I dont think it makes sense to change this.

I think there is no possible change to this PR. You will either commit or refuse it, as is.

boot/freeldr/freeldr/include/ver.h
```diff
@@ -22,7 +22,7 @@
 /* Just some stuff */
 #define VERSION         "FreeLoader v3.0"
 #define COPYRIGHT       "Copyright (C) 1998-2007 ReactOS Team"
-#define AUTHOR_EMAIL    "<www.reactos.org>"
+#define AUTHOR_EMAIL    "<reactos.org>"
```
HBelusca, line 25: Actually both URLs do work, so I don't know why we should change it.

tkreuzer:
Not a reasonable change

PR #2635, opened 2020-04-22
[SYSDM] Update now-redirected 'http' and 'www.reactos.org' URLs

SergeGautherie:


binarymaster:
@SergeGautherie how dare you to spam the repository with PRs? :rage: Why not doing it in one?

SergeGautherie:
> Why not doing it in one?

As usual, I feel like "always getting the opposite complain: global = too big, per component = too small" :-(
Then, you will want all remaining URL commits in 1 PR?

binarymaster:
> Then, you will want all remaining URL commits in 1 PR?

I'm okay with small changes, but I'm against spamming the repo. That means it would be better if you open new PRs not earlier than previous one is merged.

SergeGautherie:
> I'm okay with small changes, but I'm against spamming the repo.

I am not sure what exactly you call 'spam', then.

> not earlier than previous one is merged

Really? It seems to me that would totally defeat the purpose of working asynchronously, not counting unreviewed PRs for example.
Where is that rule that a contributor is limited to 1 PR at a time?

PR #2644, opened 2020-04-23
[CONTROL] Fix hang while calling RunControlPanel() CORE-11373

gonzoMD:
## Purpose

Fix JIRA issue [CORE-11373](https://jira.reactos.org/browse/CORE-11373)

## Proposed changes

- patch by Jira user "wimh"
- also addressed Hermes' suggestion to use ARRAYSIZE instead of MAX_PATH

## TODO

You have to tell me, if there's something  ;)

### DIFF
--- a/base/applications/control/control.c
+++ b/base/applications/control/control.c
@@ -57,18 +57,12 @@ OpenShellFolder(LPWSTR lpFolderCLSID)
 static INT
 RunControlPanel(LPCWSTR lpCmd)
 {
-    /*
-     * Old method:
-     *
     WCHAR szParameters[MAX_PATH];
-    wcscpy(szParameters, L"shell32.dll,Control_RunDLL ");
-    wcscat(szParameters, lpCmd);
-    return RUNDLL(szParameters);
-     */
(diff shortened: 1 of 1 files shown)


base/applications/control/control.c
```diff
@@ -57,18 +57,12 @@ OpenShellFolder(LPWSTR lpFolderCLSID)
...
-    return RUNDLL(szParameters);
-     */
+    StringCchCopyW(szParameters, ARRAYSIZE(szParameters), L"shell32.dll,Control_RunDLL ");
+    if (FAILED(StringCchCatW(szParameters, ARRAYSIZE(szParameters), lpCmd)))
+        return 0;
 
-    /* New method: */
-    Control_RunDLLW(GetDesktopWindow(), 0, lpCmd, SW_SHOW);
-    return 1;
+    return RUNDLL(szParameters);
```
learn-more, line 65: So now it's back at the 'old method'.

gonzoMD: Yes, but the bug is fixed and the original author of the new method approved the fix.

PR #2645, opened 2020-04-24
[SHELL32] Distinguish floppy and removable drives

katahiromz:
## Purpose

Distinguish floppy drive and non-floppy removable drive in icon and description.

JIRA issue: [CORE-10221](https://jira.reactos.org/browse/CORE-10221)

after

### DIFF
--- a/dll/win32/shell32/dialogs/drvdefext.cpp
+++ b/dll/win32/shell32/dialogs/drvdefext.cpp
@@ -283,32 +283,17 @@ CDrvDefExt::PaintStaticControls(HWND hwndDlg, LPDRAWITEMSTRUCT pDrawItem)
     }
 }
 
-typedef NTSTATUS (NTAPI *FN_NtQueryVolumeInformationFile)(HANDLE, PIO_STATUS_BLOCK,
-                                                          PVOID, ULONG, FS_INFORMATION_CLASS);
-
 // https://stackoverflow.com/questions/3098696/get-information-about-disk-drives-result-on-windows7-32-bit-system/3100268#3100268
 static BOOL
 GetDriveTypeAndCharacteristics(HANDLE hDevice, DEVICE_TYPE *pDeviceType, ULONG *pCharacteristics)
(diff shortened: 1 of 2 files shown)


dll/win32/shell32/dialogs/drvdefext.cpp
```diff
@@ -278,6 +283,115 @@ CDrvDefExt::PaintStaticControls(HWND hwndDlg, LPDRAWITEMSTRUCT pDrawItem)
...
+    IO_STATUS_BLOCK IoStatusBlock;
+    FILE_FS_DEVICE_INFORMATION DeviceInfo;
+
+    ntdll = GetModuleHandleA("ntdll");
+    if (ntdll == NULL)
+        return FALSE;
+
+    pNtQueryVolumeInformationFile =
+        (FN_NtQueryVolumeInformationFile)
+            GetProcAddress(ntdll, "NtQueryVolumeInformationFile");
```
learn-more, line 305: We can probably call this directly here, without the need for GetProcAddress?

katahiromz: OK.

katahiromz:
Can you test, somebody? I don't have removable media.

binarymaster:
@Extravert-ir I remember once you told to distinguish drive types, PnP-aware disk driver is required, can you check it?

yagoulas:
I think I had a look at this some months ago. I remember seeing that GetDriveTypeW in ros doesn't return the correct type and when it is fixed we can use it to keep things clear. Could you have a look at it and if it works properly in ros?

katahiromz:
I think the return value `DRIVE_REMOVABLE` of `GetDriveTypeW` is vague in Windows, between floppy drive type and non-floppy removable drive type.

katahiromz:
I cannot contact the reporter of this issue. I cannot test this PR.

katahiromz:
Assuming correct...

PR #2647, opened 2020-04-24
[THEMES] Update now-redirected 'http' and 'www.reactos.org' URLs

SergeGautherie:


ColinFinck:
I'm merging this for the https:// change. I don't really care about www.reactos.org vs. reactos.org - and we shouldn't have PRs just for the sake of changing this.

SergeGautherie:
> I'm merging this for the https:// change. I don't really care about [www.reactos.org](http://www.reactos.org) vs. reactos.org - and we shouldn't have PRs just for the sake of changing this.

As well do both, isn't it?

PR #2651, opened 2020-04-24
[MSPAINT] Add to recent documents

katahiromz:
## Purpose

JIRA issue: [CORE-3588](https://jira.reactos.org/browse/CORE-3588)

- Add to recent documents upon file save.

after

### DIFF
--- a/base/applications/mspaint/dib.cpp
+++ b/base/applications/mspaint/dib.cpp
@@ -69,6 +69,8 @@ SaveDIBToFile(HBITMAP hBitmap, LPTSTR FileName, HDC hDC, LPSYSTEMTIME time, int 
     // TODO: update hRes and vRes
 
     CloseHandle(hFile);
+
+    registrySettings.SetMostRecentFile(FileName);
 }
 
 void ShowFileLoadError(LPCTSTR name)


base/applications/mspaint/dib.cpp
```diff
@@ -69,6 +71,13 @@ SaveDIBToFile(HBITMAP hBitmap, LPTSTR FileName, HDC hDC, LPSYSTEMTIME time, int
     // TODO: update hRes and vRes
 
     CloseHandle(hFile);
+
+    if (bAlreadyExists)
+        SHChangeNotify(SHCNE_UPDATEITEM, SHCNF_PATHW, FileName, NULL);
```
ThFabba, line 76: This cannot possibly be right. Every application that modifies files does not have to call this function. The filesystem is responsible for that. Please test this on Windows / against Windows's mspaint

base/applications/mspaint/precomp.h
```diff
@@ -3,10 +3,7 @@
 
 #include <stdarg.h>
 
-#include <windef.h>
-#include <winbase.h>
-#include <winuser.h>
-#include <wingdi.h>
+#include <windows.h>
```
ThFabba, line 6: Please don't just go around changing these, this was done for build performance reasons.

base/applications/mspaint/dib.cpp
katahiromz: Reverted.

base/applications/mspaint/precomp.h
katahiromz: Reverted.

PR #2563, opened 2020-04-17
[SDK] Wine sync _mbtowc_l code

Fraizeraust:
[CORE-16885](https://jira.reactos.org/browse/CORE-16885)

PR #2652, opened 2020-04-24
[CRT_APITEST] Add testcase for mbtowc()

Fraizeraust:
**Windows XP**
Capture2
**Windows Server 2003**

**ReactOS (with #2563 applied to build)**
Capture

### DIFF
--- a/modules/rostests/apitests/crt/mbtowc.c
+++ b/modules/rostests/apitests/crt/mbtowc.c
@@ -1,0 +1,61 @@
+/*
+ * PROJECT:         ReactOS API tests
+ * LICENSE:         GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:         Tests for mbtowc
+ * COPYRIGHT:       Copyright 2020 Bișoc George <fraizeraust99 at gmail dot com>
+ */
+
+#include <apitest.h>
+#include <apitest_guard.h>
+
+#define WIN32_NO_STATUS
+#include <stdio.h>
+#include <stdlib.h>
+
+START_TEST(mbtowc)
+{
+    int Length;
+    wchar_t BufferDest[3];
+    char *ch;
+
+    ch = AllocateGuarded(sizeof(ch));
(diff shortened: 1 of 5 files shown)


modules/rostests/apitests/crt/mbtowc.c
```diff
@@ -0,0 +1,46 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+
+#define ok_wchar(x, y) ok_int(x, y)
+
+START_TEST(mbtowc)
+{
+    size_t Length;
+    wchar_t BufferDest[3];
+    char ch = 'A';
```
ThFabba, line 20: A possible future improvement would be to allocate a buffer for this using `AllocateGuarded`, to show that only the first character is ever accessed.

modules/rostests/apitests/crt/mbtowc.c
```diff
@@ -0,0 +1,46 @@
...
+    /* The multibyte argument is invalid */
+    Length = mbtowc(BufferDest, NULL, 0);
+    ok(Length == 0, "Expected 0 characters to be converted but got %u.\n", Length);
+
+    /* The multibyte argument is invalid but count number for examination is correct */
+    Length = mbtowc(BufferDest, NULL, MB_CUR_MAX);
+    ok(Length == 0, "Expected 0 characters to be converted but got %u.\n", Length);
+
+    /* Convert the character and validate the output that we should get */
+    Length = mbtowc(BufferDest, &ch, MB_CUR_MAX);
```
ThFabba, line 43: Another useful test would be this with a NULL wchar parameter, to show that you can get the length without receiving output.

Fraizeraust:
Both MSVCRT and NTDLL apitest testcases work flawlessly for this function testcase in Windows Server 2003 and ReactOS, although with the exception of NTDLL testcase unit in Windows XP as `ntdll_crt_apitest.exe` prematurely fails with a missing entry point ` _vscwprintf` in ntdll.dll which by default doesn't exist in XP.

PR #2654, opened 2020-04-24
[INTL] Translate hardcoded string

KyleKatarn31:
## Purpose

_ In Intl / Advanced CPL, there is an hardcoded string, not available for translation ("Unable to install a new language for programs don't support unicode!")

JIRA issue: [CORE-16951](https://jira.reactos.org/browse/CORE-16951)

## Proposed changes

_ String added to the localized ressource files.


dll/cpl/intl/lang/en-US.rc
```diff
@@ -214,3 +214,8 @@ BEGIN
     IDS_ERROR_NLS_CODE_REG "Problem opening key: HKLM\\SYSTEM\\CurrentControlSet\\Control\\NLS\\CodePage"
     IDS_ERROR_INPUT_DLL "Unable to start input.dll"
 END
+
+STRINGTABLE
```
ThFabba, line 218: Any reason to add a new table rather than including this string in the previous one?

KyleKatarn31: No specific reason. Would you prefer the previous table ? Another one ? What's the impact ?

Saibamen:
Polish translation: https://github.com/reactos/reactos/pull/2665

PR #2657, opened 2020-04-24
[BTRFS] sha256.c: Restore upstream code. CORE-16679

SergeGautherie:
## Purpose

2 commits, so you can choose.

Addendum to 3ba423996c41f9d8a6284a0aac1a44aa332815a6.
JIRA issue: [CORE-16679](https://jira.reactos.org/browse/CORE-16679)

### DIFF
--- a/drivers/filesystems/btrfs/compress.c
+++ b/drivers/filesystems/btrfs/compress.c
@@ -888,7 +888,7 @@ NTSTATUS write_compressed(fcb* fcb, uint64_t start_data, uint64_t end_data, void
     void* csum = NULL;
 #ifdef __REACTOS__
     int32_t i2;
-    uint32_t i3, j;
+    uint32_t i3;
 #endif // __REACTOS__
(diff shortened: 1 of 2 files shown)


ColinFinck:
We are compiling with C99 enabled now, so this should no longer be necessary.
Can you check if we can restore the original code without any `#ifdef __REACTOS__`?

SergeGautherie:
> Can you check

Not yet, as a VS2015 env is still missing...

SergeGautherie:
Not `WIP` anymore.

Extravert-ir:
It makes sense to either restore all of the code to original (in places where we changed the style to C89), or just wait for a next release.
I prefer the latter

ColinFinck:
Closing this PR then as we're going to fix this when importing a new WinBtrfs release.
Which shouldn't be too far away considering the current daily commits to the winbtrfs upstream repository.

PR #2658, opened 2020-04-25
Drop Visual Studio 2010-2013

Extravert-ir:
This year we celebrate a 10th anniversary of Visual Studio 2010 IDE.
Good time make this pull request :D

### DIFF
--- a/configure.cmd
+++ b/configure.cmd
@@ -50,14 +50,11 @@ if defined ROS_ARCH (
     cl 2>&1 | find "x86" > NUL && set ARCH=i386
     cl 2>&1 | find "x64" > NUL && set ARCH=amd64
     cl 2>&1 | find "ARM" > NUL && set ARCH=arm
-    cl 2>&1 | find "16.00." > NUL && set VS_VERSION=10
-    cl 2>&1 | find "17.00." > NUL && set VS_VERSION=11
-    cl 2>&1 | find "18.00." > NUL && set VS_VERSION=12
     cl 2>&1 | find "19.00." > NUL && set VS_VERSION=14
     cl 2>&1 | findstr /R /c:"19\.1.\." > NUL && set VS_VERSION=15
     cl 2>&1 | findstr /R /c:"19\.2.\." > NUL && set VS_VERSION=16
     if not defined VS_VERSION (
-        echo Error: Visual Studio version too old ^(before 10 ^(2010^)^) or version detection failed.
+        echo Error: Visual Studio version too old ^(before 14 ^(2015^)^) or version detection failed.
         goto quit
     )
     set BUILD_ENVIRONMENT=VS
--- a/sdk/cmake/msvc.cmake
+++ b/sdk/cmake/msvc.cmake
@@ -50,19 +50,17 @@ endif()
 
 # HACK: for VS 11+ we need to explicitly disable SSE, which is off by
 # default for older compilers. See CORE-6507
-if(MSVC_VERSION GREATER 1699 AND ARCH STREQUAL "i386")
+if(ARCH STREQUAL "i386")
     add_compile_flags("/arch:IA32")
 endif ()
 
 # VS 12+ requires /FS when used in parallel compilations
-if(MSVC_VERSION GREATER 1799 AND NOT MSVC_IDE)
+if(NOT MSVC_IDE)
     add_compile_flags("/FS")
 endif ()
 
 # VS14+ tries to use thread-safe initialization
-if(MSVC_VERSION GREATER 1899)
-    add_compile_flags("/Zc:threadSafeInit-")
-endif ()
+add_compile_flags("/Zc:threadSafeInit-")
 
 # HACK: Disable use of __CxxFrameHandler4 on VS 16.3+ (x64 only)
 # See https://developercommunity.visualstudio.com/content/problem/746534/visual-c-163-runtime-uses-an-unsupported-api-for-u.html
@@ -112,8 +110,8 @@ add_compile_flags("/wd4018")
 add_compile_flags("/we4013 /we4020 /we4022 /we4028 /we4047 /we4098 /we4101 /we4113 /we4129 /we4133 /we4163 /we4229 /we4311 /we4312 /we4313 /we4477 /we4603 /we4700 /we4715 /we4716")
 
 # - C4189: local variable initialized but not referenced
-# Not in Release mode and not with MSVC 2010
-if((NOT CMAKE_BUILD_TYPE STREQUAL "Release") AND (NOT MSVC_VERSION LESS 1700))
+# Not in Release mode
+if(NOT CMAKE_BUILD_TYPE STREQUAL "Release")
     add_compile_flags("/we4189")
 endif()
 
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ The code of ReactOS is licensed under [GNU GPL 2.0](https://github.com/reactos/r
 To build the system it is strongly advised to use the _ReactOS Build Environment (RosBE)._
 Up-to-date versions for Windows and for Unix/GNU-Linux are available from our download page at: ["Build Environment"](https://reactos.org/wiki/Build_Environment).
 
-Alternatively one can use Microsoft Visual C++ (MSVC) version 2010+. Building with MSVC is covered here: ["Visual Studio or Microsoft Visual C++"](https://reactos.org/wiki/CMake#Visual_Studio_or_Microsoft_Visual_C.2B.2B).
+Alternatively one can use Microsoft Visual C++ (MSVC) version 2015+. Building with MSVC is covered here: ["Visual Studio or Microsoft Visual C++"](https://reactos.org/wiki/CMake#Visual_Studio_or_Microsoft_Visual_C.2B.2B).
 
 ### Binaries
 
--- a/sdk/cmake/gcc.cmake
+++ b/sdk/cmake/gcc.cmake
@@ -52,7 +52,7 @@ if(NOT GCC_VERSION VERSION_LESS 4.8)
 endif()
 
 if(CMAKE_C_COMPILER_ID STREQUAL "Clang")
-    add_compile_flags_language("-std=gnu89 -Wno-microsoft" "C")
+    add_compile_flags_language("-std=gnu99 -Wno-microsoft" "C")
     set(CMAKE_LINK_DEF_FILE_FLAG "")
     set(CMAKE_STATIC_LIBRARY_SUFFIX ".a")
     set(CMAKE_LINK_LIBRARY_SUFFIX "")
@@ -70,7 +70,6 @@ if(DBG)
     if(NOT CMAKE_C_COMPILER_ID STREQUAL "Clang")
         add_compile_flags_language("-Wold-style-declaration" "C")
     endif()
-    add_compile_flags_language("-Wdeclaration-after-statement" "C")
 endif()
 
 add_compile_flags_language("-fno-rtti -fno-exceptions" "CXX")
@@ -351,8 +350,8 @@ endif()
 
 function(fixup_load_config _target)
     get_target_property(PEFIXUP native-pefixup IMPORTED_LOCATION_NOCONFIG)
-    add_custom_command(TARGET ${_target} POST_BUILD 
-        COMMAND "${PEFIXUP}" 
+    add_custom_command(TARGET ${_target} POST_BUILD
+        COMMAND "${PEFIXUP}"
                 "$<TARGET_FILE:${_target}>"
         COMMENT "Patching in LOAD_CONFIG")
 endfunction()
--- a/sdk/include/reactos/libs/pseh/pseh3.h
+++ b/sdk/include/reactos/libs/pseh/pseh3.h
@@ -16,12 +16,6 @@
 extern "C" {
 #endif
 
-#ifdef __cplusplus
-#define PRAGMA_DIAGNOSTIC_IGNORED_DECLARATION_AFTER_STATEMENT
-#else
-#define PRAGMA_DIAGNOSTIC_IGNORED_DECLARATION_AFTER_STATEMENT _Pragma("GCC diagnostic ignored \"-Wdeclaration-after-statement\"")
-#endif
-
 /* CLANG must safe non-volatiles, because it uses a return-twice algorithm */
 #if defined(__clang__) && !defined(_SEH3$_FRAME_ALL_NONVOLATILES)
 #define _SEH3$_FRAME_ALL_NONVOLATILES 1
@@ -359,9 +353,6 @@ _Pragma("GCC diagnostic pop") \
         (void)&&_SEH3$_l_BeforeFilterOrFinally; \
         (void)&&_SEH3$_l_FilterOrFinally; \
 \
-_Pragma("GCC diagnostic push") \
-PRAGMA_DIAGNOSTIC_IGNORED_DECLARATION_AFTER_STATEMENT \
-\
         /* Count the try level. Outside of any __try, _SEH3$_TryLevel is 0 */ \
         enum { \
             _SEH3$_PreviousTryLevel = _SEH3$_TryLevel, \
@@ -373,8 +364,6 @@ PRAGMA_DIAGNOSTIC_IGNORED_DECLARATION_AFTER_STATEMENT \
 \
         /* Allocate a registration frame */ \
         volatile SEH3$_REGISTRATION_FRAME _SEH3$_AUTO_CLEANUP _SEH3$_TrylevelFrame; \
-\
-_Pragma("GCC diagnostic pop") \
 \
         goto _SEH3$_l_BeforeTry; \
         /* Silence warning */ goto _SEH3$_l_AfterTry; \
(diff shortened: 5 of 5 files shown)


JoachimHenze:
Are you crazy? Why? Keeping our code compatible to multiple compilers is a big plus!
I really dislike the idea to simply drop the support, just because it is considered as "convenient".
Especially VS2010 is a really important compiler for us still.

sdk/cmake/msvc.cmake
```diff
@@ -113,7 +111,7 @@ add_compile_flags("/we4013 /we4020 /we4022 /we4028 /we4047 /we4098 /we4101 /we41
 
 # - C4189: local variable initialized but not referenced
 # Not in Release mode and not with MSVC 2010
```
SergeGautherie, line 113: Here and previously: clean up comments too?

SergeGautherie:
It certainly is surprising to have such a non-draft PR pop up without notice.

> compatible to multiple compilers

Question is: what actual advantages does v2010(-2013) still have over v2015(+)?

> VS2010 is a really important compiler for us still

I would kind of hope something like this, like the upgrade to GCC 8.
Nonetheless, as the latter, it would require to upgrade build environments (ReactOS, AppVeyor, ...) first.

Doug-Lyons:
The latest version of MS Visual Studio that I own is VS2013. I am not a professional developer and in the past I have paid out of my own pocket for this and previous development software's such as Visual Basic 6. Up until VS2015 MS offered an "upgrade" path so that if you had owned a previous version of one of their development packages or a development package from another vendor, you could get a much reduced price for a new VS2013. As I remember I paid about US$ 150 for mine. But with VS2015, MS discontinued this practice and I was looking at paying about US$ 500 for a new package. For my use as a recreational programmer with only occasional use, this did not seem like a good investment for me, so I did not purchase it and have never upgraded. So I have Visual Studio 2013 and I am staying with it for what I consider a good reason. If you abandon it, then I can certainly go with the free versions offered, but they are not full-featured as you know. I hope that you keep support at least back to VS2013 because it was the last reasonably priced VS version. Thanks.

sdever:
@Doug-Lyons The VS Community just has the same features than Profesional edition (this is not the limited "Express" editions). The only difference is the license restrictions on how you can use it to create software/in your company, but for occasional/open-source use it's perfectly fine.

binarymaster:
> Keeping our code compatible to multiple compilers is a big plus!

Maintaining code diffs of 3rd-party components (like WinBtrfs) which do not support old compilers is a big minus.

Extravert-ir:
> Because it is considered as "convenient".

In order to support this compiler, I need to write all code in C89 style, which is not pleasing at all. Moreover, it scares new-comers (including me some time ago). It could be a good trade-off, but by supporting this compiler we receive nothing except dozens of "Fix build" commits (because really most of the developers use something more modern)

And as @binarymaster mentioned, take a look at drivers/filesystems/btrfs folder - it has 76 mentions of `#if(n)def __REACTOS__` and half of them are just to support MSVC 2010. This is a real pain in the ass to maintain.

> Keeping our code compatible to multiple compilers is a big plus!

This is not the case here. Supporting more *versions* of a compiler does not make much sense from a developer point of view - latest version always have all features from previous ones.

Actually, this was discussed already, I don't understand why people are surprised https://reactos.org/project-news/december-2018-meeting-minutes/

JoachimHenze:
Making modern MSVC 2015+ to generate any XP/2k3 compatible binary (==current ros compatible) is pita.
Does the VS2015 toolchain even run on ros at all?
Argumentation based on 3rd party BTRFS is a bit problematic, because it existing in our sources is a pure workaround, as we do not have any NTFS yet. Why not helping to improve official BTRFS codebase to be VS2010 compatible instead? I mean the work has even been done for that already in our tree.

And I stick with my opinion: Usually supporting multiple compiler versions is an indicator for having high quality code.

learn-more:
> Making modern MSVC 2015+ to generate any XP/2k3 compatible binary (==current ros compatible) is pita.

Selecting another toolchain is what is needed in 2015, that is all.
PS: I build ros with 2015 for years already.

> Argumentation based on 3rd party BTRFS is a bit problematic, because it existing in our sources is a pure workaround, as we do not have any NTFS yet. Why not helping to improve official BTRFS codebase to be VS2010 compatible instead? I mean the work has even been done for that already in our tree.

I don't think he accepts such patches, and for good reason, it means lowering the code quality.

> And I stick with my opinion: Usually supporting multiple compiler versions is an indicator for having high quality code.

usually, yes, but in this case that means sticking to outdated / older patterns that are not giving us anything except for trouble

cjsthompson:
> because it existing in our sources is a pure workaround, as we do not have any NTFS yet.

Btrfs is not a workaround, it's a feature and a major one. While NTFS apparently has some bits and pieces required for some Windows features, Btrfs is a next generation file system with a slew of features NTFS for Windows 10 is nowhere close to have.

HBelusca:
> it means lowering the code quality.

I don't know what he could lower further.

HBelusca:
> latest version always have all features from previous ones

When you see the number of updates MSVC 2017/2019 receives because of numerous bugs in the compiled code it generates, one would prefer to use a compiler that has passed the test of time!

ThFabba:
You probably need a change to `appveyor.yml` too (currently uses 2010). Not sure if all it needs is to change the Visual Studio path or if you somehow need to configure it to use a different builder type.

Extravert-ir:
> because of numerous bugs in the compiled code it generates

Are you sure about that? AFAIK they just add more C++ features to STL these days

SergeGautherie:
> WinBtrfs

Note that, to fully support its shellext code as is, C++17 from VS 2017(+) would be needed even.

> surprised

Indeed: 'the decision is to drop VS 2010 and 2013'.
But unlike the GCC upgrade: no Jira issue, no build env set up, ...

NB:
It also reports 'The decision was made to work towards moving to GCC 6.4',
though it seems 8.3 is the actual target, so one can wonder how accurate this report (still) is.

PS:
In any case, the fact that this PR became a forum discussion means decision/communication process went wrong somehow.

> run on ros

Quick glance: it would seem VS 2012+ do not run on XP (and S03?), so that would probably need ROS compatibility layer...
@Doug-Lyons, does your VS 2013 run on ROS?

Additional question is:
while ROS project can use VS 2015 as its oldest builder (as in 'official support'),
does that mean it is already planned to use 2015-only features? Or would 2013(-) still be usable by contributors, for some time at least?
In other words, how much clean up is required by this very PR?
NB: Quick glance, it seems VS 2013 added more parts of C99 core, then VS 2015 added more C99 library support.

> different builder type

VS 2015 should be available on both 2015 (currently used) and 2017 images.
I would suggest to upgrade both, in either 1 or 2 steps.

Doug-Lyons:
@SergeGautherie, On ReactOS running VS2013, I am blocked on the install by not being able to install Net Framework 4.5.1 as you can see here:

ReactOS_VS2013_Blocked

vgalnt:
Why do need Btrfs if RoS is an operating system for virtual machines. Is FAT not enough for virtual machines?
The next step will be to abandon x86?

ColinFinck:
Visual Studio 2015+ comes in a free Community Edition for open-source projects that has all the features of the Professional Edition. Hence no reason to keep support for older versions just because some people happen to have a license.

Visual Studio 2015+ currently doesn't work under ReactOS, but neither does Visual Studio 2010.
Hence this is no reason either. And if you actually want to work on Visual Studio support in ReactOS, you can just try to install the VS Build Tools instead of the full-blown IDE. This package is not available for older Visual Studio versions either.

Supporting all kinds of old compilers and operating systems is no sign of quality, it's just unthankful work. I'm saying that as someone who just spent a shitload of time patching our RosBE 2.2.0 until it became compatible with Windows XP again. It doesn't scale to keep doing that for even more compilers.
But you see, even if we drop support for Visual Studio < 2015, you can still build ReactOS on Windows XP using GCC 8.

WinBtrfs is just one example of a third-party project using modern C++ features requiring new compilers. The number of projects doing that is only going to grow, and you know as good as me that nobody will volunteer to "fix" them.
And _we_ actually want to use modern C++ features as well. By wrapping every heap allocation into a `std::unique_ptr`, I can finally stop tracking all exit paths of a function to not miss a `HeapFree`.
I can also use the safer `std::wstring` everywhere where I had to use `WCHAR` buffers previously, because these things are finally interchangeably in C++17.

Finally, this change is just the beginning for more to come: Victor and me have planned to import a proper C++ STL for quite some time. Right now, we build using the host STL on GCC and with an abandoned stlport on MSVC. This setup requires utter hacks and is the reason for several funky problems in our C++ binaries.
And a new STL is obviously going to depend on modern compiler features.

All of this is why I fully support this proposal to drop VS support for anything < VS 2015!

Zero3K:
> 
> 
> Why do need Btrfs if RoS is an operating system for virtual machines. Is FAT not enough for virtual machines?
> The next step will be to abandon x86?

BTRFS doesn't have as many issues as FAT32 does in ReactOS.

vgalnt:
I would prefer fixing FAT32 problems in ReactOS. But the best option I see is NTFS. When FAT32 and NTFS will not have problems in ReactOS, then you can already add other "non-native" file systems (if there is enough time for this). Why has Linux already had NTFS-3g (which supports recording and formatting), but not ReactOS?

binarymaster:
> Why has Linux already had NTFS-3g (which supports recording and formatting), but not ReactOS?

Vadim, you worked hard and deep in our kernel, you should know the answer. 😅

vgalnt:
I only know that "if nothing is doing, then and nothing will happen" (c) <Ordinary human>

SergeGautherie:
Doug, thanks for the tests/confirmation. (File Jira issues, if need be.)

Colin, thanks for the detailed explanation!

Saibamen:
@ThFabba: I changed AppVeyor paths in https://github.com/reactos/reactos/pull/2721

JoachimHenze:
> BTRFS doesn't have as many issues as FAT32 does in ReactOS.
Zero3k, Your statement is as wrong as a statement can get.

> In any case, the fact that this PR became a forum discussion means decision/communication process went wrong somehow.
Definitely!

MakiseKurisu:
TBH I'd like to see 2015 being dropped as well. [2017](https://github.com/mesonbuild/meson/blob/131459bd05d8d5fbc118892872293704dd6f93a1/mesonbuild/environment.py#L196) introduced 2 new environment variables to ease the detection of host and target architecture. It also introduces [vswhere](https://renenyffenegger.ch/notes/Windows/development/Visual-Studio/environment-variables/index) command so we can find `vcvarsall.bat` automatically so `configure.cmd` can set up environment by itself with setlocal/endlocal.

HBelusca:
We could as well split our codebase further, with (say) ReactOS "Core" being the very basic of NT kernel and user dll + the very base of win32, that we "enforce" to be compilable with say MSVC 2010, and a separate repo where there is everything else, that we can only ensure it's compilable with the latest MSVC.

MakiseKurisu:
I have long dreamed a ReactOS pacman for package management. If someone wants to make it happen I'm happy to help.

ColinFinck:
> We could as well split our codebase further, with (say) ReactOS "Core" being the very basic of NT kernel and user dll + the very base of win32, that we "enforce" to be compilable with say MSVC 2010

No reason to make hard operating system development even harder by keeping eternal compatibility to ancient compilers...

vgalnt:
ReactOS development reminds me of building a multi-story building for free housing. Despite the fact that the third part of the foundation is missing, for some reason everyone is engaged in the construction of walls and even the roof of this house. But despite being free, no one plans to live in this house. And even few people believe that this house will ever be completed. And even if the roof is completed, the foundation will remain in a deplorable state ...

binarymaster:
Complaining about the kernel won't help with its development or magically make all people working on it. Also you probably know we have bugreport-driven development.

@vgalnt and what about answering the questions in the kernel roadmap ticket you created? [CORE-16841](https://jira.reactos.org/browse/CORE-16841)

ColinFinck:
@vgalnt I'm fully with you on that, but dropping VS2010 support is not about fancy user-mode features. If we had this discussion 30 years ago, people would argue for writing our kernel in pure assembler instead of that hipster C language. In the same way, there is no reason today to limit us to the artificial boundaries of C89.
If people think that maintaining compatibility with all kinds of ancient compilers is so easy, they shall have a look at GNU `configure` scripts. It needs exactly that amount of preprocessor definitions to properly handle all cases, and definitely makes development harder for no compelling reason.

vgalnt:
@binarymaster  I would like to have magical powers so that at least half of all people start working on the kernel, but yes, I'm not a magician, alas. That’s why I’m trying to say that until the kernel is completed, until ReactOS is actually not a real operating system, but just a toy in a virtual machine ...
And that is why I created CORE-16841. But apparently this does not interest either the developers themselves or the ReactOS community. In fact, not a single message was about the ROADMAP, and I see no substantive questions.
I know that you can tell me the kernel of the OS is very complex and no one understands it. I do not agree with you. Usually people who have a great desire, they do, despite the difficulties. And people who have no desire, they are looking for any excuses ... I am sure that among the ReactOS developers there is someone who could make the kernel. But apparently, just do not want ...

@ColinFinck From the outside, this may seem like a conversation between the dumb and the deaf. I’m trying to tell you that first they finish building the foundation, and then the walls begin, and especially the roof. To which you answer me that you agree, and add that recently they brought modern nails from higher strength steel and that it is better to build a roof with them )

It may be clearer on the example of a car ...
Imagine that you decided 30 years ago to make a car similar to ... for example, LADA Niva 4x4 (it doesn’t matter). You make this car in the garage with your buddies. And the engine is already starting up, but after working for a short time, it stalls. And yes, only the first gear works on a manual gearbox. If you do not understand how the engine works and do not do it, then on this car you will not be able to ride your grandchildren.

JoachimHenze:
For all that still do not believe me when I say that supporting multiple compilers *will* improve your code quality, here is the living proof, just a few days after our discussion:
https://git.reactos.org/?p=reactos.git;a=commit;h=4cbb21761cde71af7ad47573099ca92d966c8920
For sure the code is better after this commit, compared to the before-state.
I would not have spotted this, if I would not build with 0.4.14RC with RosBE2.1.6 and ISO C90 restriction.

configure.cmd
```diff
@@ -50,14 +50,11 @@ if defined ROS_ARCH (
     cl 2>&1 | find "x86" > NUL && set ARCH=i386
     cl 2>&1 | find "x64" > NUL && set ARCH=amd64
     cl 2>&1 | find "ARM" > NUL && set ARCH=arm
-    cl 2>&1 | find "16.00." > NUL && set VS_VERSION=10
-    cl 2>&1 | find "17.00." > NUL && set VS_VERSION=11
-    cl 2>&1 | find "18.00." > NUL && set VS_VERSION=12
```
JoachimHenze, line 55: I request to undo this check. Maybe you can generate a warning instead a hard failure.
The warning could say that VS2015 is our main compiler now and that automated builds do not cover VS<2015 anymore, but that PRs with fixes for maintaining/restoring older compiler support down to VS2010 would be welcome.

ColinFinck: No, they won't be welcome. They would actively counter our efforts to remove code duplication/hacks for VS2010 (already happened) and bring in a modern C++ library (as Dominik2 is doing right now), which will then lead to a lot more readable code and resolving the even bigger mess that our C++ build environment is right now.
No C++ developer of today wants to be bound by the limits of ancient C++03, and as soon as the first C++11 code is in, there is simply no way for VS2010 support without maintaining two versions of the same application.

If you want ReactOS to become a thing in a lifetime, you at least have to give developers the best tools available!

ColinFinck:
@vgalnt You still seem to think that C99/C++11 support is not relevant to the foundation (kernel) at all. It highly is! People like @janderwald have been writing fundamental audio and USB drivers in C++ for years. Third-party drivers like WinBtrfs make use of these features too, and already need to be hacked for ReactOS.
Again, if you want a bigger kernel team, why can't that team use the best tools available? Why does it have to restrict itself to decades-old standards? What's so nice about a 10-year-old compiler?

To me as one of the RosBE maintainers, the move to a modern C++ standard library solves a fundamental problem that's even below the kernel, it's in the build environment. If building ReactOS is already based on assumptions, forget about everything else. And right now it is: We mix RosBE and ReactOS components when linking any C++ module (GCC) or rely on an unmaintained dead-end C++ library (MSVC). Things work and crash by chance.
The only way forward is the import of a maintained C++ standard library supported by both compilers. But all modern implementations are for C++11 and higher, and therefore require compilers newer than VS2010.

I can't stress enough how relying on this particular version of Visual Studio prevents progress on ReactOS fundamentals!

@JoachimHenze You don't really want to argue with this one-character change, do you?
I could as well use it to support my arguments, because a duplicate semicolon there could in no way be misunderstood by any sane compiler. We have a lot more of those in our code (cf. https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=%3B%3B), and nobody seems to bother. Maybe people rather want to write actual code than caring about each and every compiler?

HBelusca:
> People like @janderwald have been writing fundamental audio and USB drivers in C++ for years. Third-party drivers like WinBtrfs make use of these features too, and already need to be hacked for ReactOS.

Code that is otherwise unreadable, buggy, and needs to be partly changed.

Extravert-ir:
> "let's break VS 2010 build of everything (user+kernel-mode included) because 2 or 3 3rd-party things we want in kernel mode want to use C++xx

@HBelusca nope. As a developer I want to write code in C99, because this it more readable. It's not my fault that VS2010 doesn't support code as simple as
```
for (int i = 0; i < n; i++) {}
```
I'm not a C++ guy (yet), but for them this is even more relevant. C++14 is almost completely different language from what we have to use now

vgalnt:
I can’t understand what the target OS version for ReactOS is. Maybe Windows 10? Then I can understand. But if it is Windows 2003 ... then it looks like roughly, as if your child cannot walk, but you teach him to ride a bicycle ...
I have XP (and so far I don’t see critical reasons to switch), and apparently 0.4.14 will be the last one for me that I can build on MS. Ok ...

PR #2660, opened 2020-04-26
[SHELL32] Update name in CDefView::LV_RenameItems

katahiromz:
## Purpose
By specifying `LPSTR_TEXTCALLBACKW`, we explicitly update the text/name of the listview item of Explorer.
JIRA issue: [CORE-16960](https://jira.reactos.org/browse/CORE-16960)


yagoulas:
So when a file gets renamed the name we show was not shown?

PR #2665, opened 2020-04-26
Intl polish update

Saibamen:
After https://github.com/reactos/reactos/pull/2654


binarymaster:
Can you please update your name in [GitHub profile settings](https://github.com/settings/profile) to your full name `Adam Stachowicz`?

I think it would be better than committing under a nickname.

Saibamen:
@binarymaster Done

PR #2668, opened 2020-04-26
[SHLWAPI] Improve SHEvaluateSystemCommandTemplate export CORE-14742

oleg-dubinskiy:
## Purpose

Improve export for SHEvaluateSystemCommandTemplate funtion in shlwapi by using `stdcall -stub` instead of `stub`.
Required by TuneUP service from TuneUP Utilites 2014. After my changes, it does no longer crash, and main app now works correctly. Since this stub it enough to avoid that crash, I guess it is not necessarily to add a full code stub.

JIRA issue: [CORE-14742](https://jira.reactos.org/browse/CORE-14742)

## Proposed changes

- Replace `stub` by `stdcall -stub`;
- Add function arguments according to MSDN, since using them is required by `stdcall -stub` unlike `stub`.

## Result

Before:
0 4 14-dev-1531-ga2f5283

After:
MS_shlwapi

### DIFF
--- a/dll/win32/shlwapi/shlwapi.spec
+++ b/dll/win32/shlwapi/shlwapi.spec
@@ -549,7 +549,7 @@
 549 stdcall -noname SHCoCreateInstanceAC(ptr ptr long ptr ptr)
 550 stub -noname GetTemplateInfoFromHandle
 551 stub -noname IShellFolder_CompareIDs
-552 stdcall -stub -noname SHEvaluateSystemCommandTemplate(wstr ptr ptr ptr)
+552 stdcall -stub -noname -version=0x501-0x502 SHEvaluateSystemCommandTemplate(wstr ptr ptr ptr)
 553 stdcall IsInternetESCEnabled()
 554 stdcall -noname -stub SHGetAllAccessSA()
 555 stdcall AssocQueryStringByKeyA(long long ptr str ptr ptr)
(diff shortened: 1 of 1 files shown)


dll/win32/shlwapi/shlwapi.spec
```diff
@@ -549,7 +549,7 @@
 549 stdcall -noname SHCoCreateInstanceAC(ptr ptr long ptr ptr)
 550 stub -noname GetTemplateInfoFromHandle
 551 stub -noname IShellFolder_CompareIDs
-552 stub -noname SHEvaluateSystemCommandTemplate
+552 stdcall -stub -noname SHEvaluateSystemCommandTemplate(wstr ptr ptr ptr)
```
SergeGautherie, line 552:
```suggestion
552 stdcall -stub -noname -version=0x501-0x502 SHEvaluateSystemCommandTemplate(wstr ptr ptr ptr)
```

oleg-dubinskiy:
Oops, I was wrong. SHEvaluateSystemCommandTemplate function appeared earlier, since XP SP2, so it's not NT6+. It was only moved into shell32 in Vista. And as ordinal it was retired since Vista+. The only difference is an NT5 version of this function was not documented, unlike NT6 one. See https://geoffchappell.com/studies/windows/shell/shlwapi/api/index.htm, https://www.geoffchappell.com/studies/windows/shell/shlwapi/history/ords600.htm and the following @SergeGautherie's comment: https://jira.reactos.org/browse/CORE-14742?focusedCommentId=122605&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-122605.

dll/win32/shlwapi/shlwapi.spec
oleg-dubinskiy: Committed, thanks.

PR #2679, opened 2020-04-27
Update CODEOWNERS - translation section

Saibamen:
Add missing translation teams

`*` is, because we have other file extension for translation somewhere like here: https://github.com/reactos/reactos/tree/master/base/setup/usetup/lang

Some language teams appears here more than one, like lang-chinese or portuguese, but I belive this list is not finished yet.

Lang codes: http://www.lingoes.net/en/translator/langcode.htm

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -327,7 +327,22 @@ amd64/                                      @tkreuzer
 # Translations
 # This is the list of translation teams in ReactOS GitHub organization.
 # If you want to be part of one - hit us at https://chat.reactos.org/
-de-DE.rc    @reactos/lang-german
-fr-FR.rc    @reactos/lang-french
-nl-NL.rc    @reactos/lang-dutch
-ru-RU.rc    @reactos/lang-russian
+de-DE.*    @reactos/lang-german
+es-ES.*    @reactos/lang-spanish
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -327,7 +327,22 @@ amd64/                                      @tkreuzer
 # Translations
 # This is the list of translation teams in ReactOS GitHub organization.
 # If you want to be part of one - hit us at https://chat.reactos.org/
-de-DE.rc    @reactos/lang-german
-fr-FR.rc    @reactos/lang-french
-nl-NL.rc    @reactos/lang-dutch
-ru-RU.rc    @reactos/lang-russian
+de-DE.*    @reactos/lang-german
```
KyleKatarn31, line 330: Why .* instead of .rc ?

HBelusca: See the explanation at the top of the PR.

KyleKatarn31: Right.

SergeGautherie: Would there be a usable pattern/mean to explicitly list `h` and `rc` only?

binarymaster: There are also `.mc` and files without extension.

PR #2683, opened 2020-04-28
[FDC] Calling add_driver_inf() is a TODO, for the time being

SergeGautherie:
Addendum to 019f21ee1d764dff8586acc847a299714e83977a.

The `.inf` file was not moved along.
@Extravert-ir answered me: there is probably issues with that driver currently.
In any case, the latter is commented out...


Extravert-ir:
The driver is not being built. This way it will generate a cmake error, which will be another reminder to a person who will work with it in future

SergeGautherie:
Oh, I didn't know generating errors was a way to go...

PR #2653, opened 2020-04-24
[SERVMAN] UI update and Error Management

KyleKatarn31:
## Purpose

_ Current design does not warn user nor logs DEBUG traces when Service Start/Stop command fails or reach timeout
_ Current Service Start/Stop progress window are WS_EX_TOOLWINDOW which reduce lisibility, is a ReactOS specificity without good reason

This is to be considered as a pre-requisite for https://jira.reactos.org/browse/CORE-16950 resolution, which will imply message pump to be added + interaction with running threads

## Proposed changes

_ DPRINT1 traces added on failure cases
_ Error Message box presented to user upon failure with explicit root cause identification
_ Change Dialog definition to standard Windows

image

PR #2684, opened 2020-04-28
[RAPPS] Add error messages on all download failure cases (CORE-16851)

KyleKatarn31:
## Purpose

- Current RAPPS has many failure conditions during download process where process stops without user awareness nor logging, making root cause analysis difficult / impossible

JIRA issue: [CORE-16851](https://jira.reactos.org/browse/CORE-16851)

## Proposed changes

- Systematic error messages presented to the users when user awareness is required
- Additionnal logging when no additionnal user awareness is required (cancellation,...)
- All messages added to localization files

image

### DIFF
--- a/base/applications/rapps/loaddlg.cpp
+++ b/base/applications/rapps/loaddlg.cpp
@@ -514,21 +514,21 @@ VOID ShowLastError(
     HWND hWndOwner,
     DWORD dwLastError)
 {
-    LPTSTR lpMsg;
+    LPWSTR lpMsg;
     
-    if (!FormatMessage(FORMAT_MESSAGE_ALLOCATE_BUFFER |
-                       FORMAT_MESSAGE_FROM_SYSTEM |
-                       FORMAT_MESSAGE_IGNORE_INSERTS,
-                       NULL,
-                       dwLastError,
-                       LANG_USER_DEFAULT,
-                       (LPTSTR)&lpMsg,
-                       0, NULL))
+    if (!FormatMessageW(FORMAT_MESSAGE_ALLOCATE_BUFFER |
+                        FORMAT_MESSAGE_FROM_SYSTEM |
+                        FORMAT_MESSAGE_IGNORE_INSERTS,
+                        NULL,
+                        dwLastError,
+                        LANG_USER_DEFAULT,
+                        (LPWSTR)&lpMsg,
+                        0, NULL))
     {
         return;
     }
 
-    MessageBox(hWndOwner, lpMsg, NULL, MB_OK | MB_ICONERROR);
+    MessageBoxW(hWndOwner, lpMsg, NULL, MB_OK | MB_ICONERROR);
     LocalFree(lpMsg);
 }
 
--- a/base/applications/rapps/loaddlg.cpp
+++ b/base/applications/rapps/loaddlg.cpp
@@ -39,6 +39,8 @@
 #include <wininet.h>
 #include <shellutils.h>
 
+#include <debug.h>
+
 #include <rosctrls.h>
 #include <windowsx.h>
 #include <process.h>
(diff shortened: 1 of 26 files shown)


KyleKatarn31:
@kernelbin : for review.

base/applications/rapps/loaddlg.cpp
```diff
@@ -640,7 +651,10 @@ unsigned int WINAPI CDownloadManager::ThreadFunc(LPVOID param)
         }
 
         if (!hOpen)
+        {
+            MessageBox_LoadString(hMainWnd, IDS_UNABLE_TO_INTACCESS);
```
binarymaster, line 655: You can use some API function that translates error code into localized text message, so you don't need to add all these messages in app.

KyleKatarn31: Which API do you have in mind ? Most messages here are implementation-specific, not simply GetLastError(). Did I miss something ?

katahiromz: FormatMessage with HMODULE.

KyleKatarn31: Sorry but i'm missing the purpose.... what would be the benefit here ? 1 single IDS_UNABLE_TO like "Error occured : %d" ? Tailored error message looks more explicit for end user as cause may differ (configuration of path, disk capacity, proxy settings....)

binarymaster: > 1 single IDS_UNABLE_TO like "Error occured : %d" ?

It's applicable to all error strings you added.

KyleKatarn31: Sorry but I still don't undersand what you expect... In any case, i'll have the same number of strings to be translated and more code.... May I have a very simple PoC of what you suggest ?

kernelbin:
I saw you used a lot of ```goto end;``` and something like this.
why don't use SEH instead?

KyleKatarn31:
> 
> 
> I saw you used a lot of `goto end;` and something like this.
> why don't use SEH instead?

Purpose here is to improve user awareness upon failure, no full RAPPS download management rede-sign. That's more a topic reserved for GSoC work :)

kernelbin:
> > I saw you used a lot of `goto end;` and something like this.
> > why don't use SEH instead?
> 
> Purpose here is to improve user awareness upon failure, no full RAPPS download management rede-sign. That's more a topic reserved for GSoC work :)

ok   : )

base/applications/rapps/loaddlg.cpp
KyleKatarn31: Same design solution as for https://github.com/reactos/reactos/pull/2653

base/applications/rapps/loaddlg.cpp
```diff
@@ -508,6 +510,27 @@ VOID CDownloadManager::UpdateProgress(
     }
 }
 
+VOID ShowLastError(HWND hWndOwner,
+                        DWORD dwLastError)
```
binarymaster, line 514:
```suggestion
VOID ShowLastError(
    HWND hWndOwner,
    DWORD dwLastError)
```
Formatting.

KyleKatarn31: Fixed, thanks !

base/applications/rapps/loaddlg.cpp
```diff
@@ -508,6 +510,28 @@ VOID CDownloadManager::UpdateProgress(
...
+                       dwLastError,
+                       LANG_USER_DEFAULT,
+                       (LPTSTR)&lpMsg,
+                       0, NULL))
+    {
+        return;
+    }
+
+    MessageBox(hWndOwner, lpMsg, NULL, MB_OK | MB_ICONERROR);
+    LocalFree(lpMsg);
```
kernelbin, line 532: I suggest using Wide Char. ( FormatMessageW, MessageBoxW, .... )

KyleKatarn31: Yes of course. I'll fix

KyleKatarn31:
Same comment to be applied on : https://github.com/reactos/reactos/pull/2653/ (i'll use wide char too)

KyleKatarn31:
All comments are now fully integrated.
Last review round before merge ?

KyleKatarn31:
Following today's discussion in Dev channel, I will not modify this code.
Either you merge as is, or you close and discard.

KyleKatarn31:
Thanks !

PR #2688, opened 2020-04-29
[GITHUB][TRAVIS][APPVEYOR] Move all CI to GitHub Actions

Extravert-ir:
Example: https://github.com/Extravert-ir/reactos/actions/runs/90613908

Note: currently it builds with VS 2019, because I wasn't able to make it work with VS 2015. So this is something to be done later (either on Buildbot or here)


appveyor.yml
```diff
@@ -1,59 +0,0 @@
...
-        call "C:\PROGRA~2\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" amd64
-      ) else (
-        call "C:\Program Files (x86)\Microsoft Visual Studio 10.0\VC\vcvarsall.bat" x86
-      )
-  - cmake --version
-  - md c:\ros_build
-  - cd c:\ros_build
-  - call %APPVEYOR_BUILD_FOLDER%\configure.cmd %clang_configure_option% -DENABLE_ROSTESTS=1
-  - ps: >-
-      & ninja -k0 2>&1 | select-string -pattern "\[\d+\/\d+\] " -NotMatch | %{$_.Line}
```
SergeGautherie, line 39: Is `-k 0` option the default now, or is it dropped?
What about the log cleaning part?

SergeGautherie:
Quick look:
G.A. is a 3rd CI, much like AppVeyor (Linux and Windows).
Then, while adding it seems fine, why remove AppVeyor and Travis support files?

>  I wasn't able to make it work with VS 2015

Quick search:
It seems VS choice is only between 2017 and 2019,
so closest env would be to use one of them with its v140 toolchain.

> Example

For what it's worth, as current setups are different,
* Linux:
AppVeyor: 18 min
Travis: 24 min, (but runs in parallel to AppVeyor)
GitHub A.: 37 min, (but runs in parallel to Windows)
* Windows:
AppVeyor: 2 * 11 min
GitHub A.: 21 and 26 min, but run in parallel

.github/workflows/build.yml
```diff
@@ -0,0 +1,128 @@
+name: Build
+on: [push, pull_request]
+
+jobs:
+  build-gcc-i386:
+    name: GCC (i386)
+    runs-on: ubuntu-latest
```
SergeGautherie, line 7: Linux and windows: use explicit versions, to prevent from unexpected changes, ...?

Extravert-ir:
> why remove AppVeyor and Travis support files?

Because we don't need them anymore? They rely on archives from svn.reactos.org, which will become old soon

> For what it's worth, as current setups are different,

The build slowdown is likely because PCH is turned off for CMake 3.17 atm. We will fix this as soon as we drop support for RosBE 2.1

Extravert-ir:
> AppVeyor: 2 * 11 min

Before building an image within 2 * 11 min, you have to wait a couple of hours in the queue

SergeGautherie:
> Because we don't need them anymore?

Yeah. As usual: who are 'we'? What about people who may want to (still) use them? (I am.)
Additionally, wrt AppVeyor, what about real VS 2015 support?

In any case, why not let them continue to run for a while after G.A. is activated? (To compare in real life.)

> They rely on archives from svn.reactos.org, which will become old soon

I assume they could (easily) use the new archives, couldn't they?

> The build slowdown is likely because PCH is turned off for CMake 3.17 atm.

Indeed, that is one of (major) current differences.

> wait a couple of hours in the queue

Not true as is: it depends on the queue length. (Without a queue, I would say within 1 min.)
It is true that, even for free accounts, G.A.seems to offer largely increased parallel builds and timeout delays :-)

.github/workflows/build.yml
Saibamen: 1. Why sticking on older OS versions? Lets have support for older builders like Visual Studio 2015, not needed to support older Operating Systems too.. Perfect will be to run these jobs on both VS 2015 and latest (2019 for now)
2. Windows have only one version to choose from: https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners

.github/workflows/build.yml
```diff
@@ -0,0 +1,125 @@
...
+        uses: actions/upload-artifact@v1
+        with:
+          name: reactos-gcc-i386-${{github.sha}}
+          path: build/bootcd.iso
+      - name: Upload livecd
+        uses: actions/upload-artifact@v1
+        with:
+          name: reactos-gcc-i386-${{github.sha}}
+          path: build/livecd.iso
+  build-msvc-i386:
```
HBelusca, line 40:
```suggestion

  build-msvc-i386:
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,125 @@
...
+      uses: actions/upload-artifact@v1
+      with:
+        name: reactos-msvc-i386-${{github.sha}}
+        path: build/bootcd.iso
+    - name: Upload livecd
+      uses: actions/upload-artifact@v1
+      with:
+        name: reactos-msvc-i386-${{github.sha}}
+        path: build/livecd.iso
+  build-msvc-amd64:
```
HBelusca, line 81:
```suggestion

  build-msvc-amd64:
```

Saibamen:
> 
> 
> Does the workflow file accepts newlines (as suggested) to remove clutter & better visually separate the different jobs in the file?

Yes, this is YAML file

.github/workflows/build.yml
SergeGautherie: > older OS

I did not ask for 'old', I asked for 'explicit'...

> Windows have only one version

Yeah, it is confusing, as
https://help.github.com/en/actions/reference/software-installed-on-github-hosted-runners
lists S16 and S19...

Saibamen:
@Extravert-ir: What do you changed here: https://github.com/reactos/reactos/pull/2688#event-3298614313 ?

Please use separate commits so Code Reviewers can track your changes

Saibamen:
Already merged? We are missing second edge case testing which is VS 2015

.github/workflows/build.yml
Saibamen: OK, 2016 lives again: https://github.com/actions/virtual-environments/issues/68#issuecomment-602652021

SergeGautherie:
> The build slowdown is likely because PCH is turned off for CMake 3.17 atm.

For the record, MSVC 2015 x86/x64 on AppVeyor:
* RosBE 2.1.x with PCH: 9-12 min.
* RosBE 2.2.0 without PCH: 21-30 min.

appveyor.yml
SergeGautherie: Submitted #2868.

PR #2690, opened 2020-04-29
[MMSYS] Fix of "Sound" properties page (combo box and icon) rendering (CORE-16935)

KyleKatarn31:
## Purpose

- Current design show an incorrect combo box style, making it almost unusable, associated to a play/stop buton which is incorrectly rendered

JIRA issue: [CORE-16935](https://jira.reactos.org/browse/CORE-16935)

CORE-16935
image

## Proposed changes

- Correct Combo box style
- Use of ico (based on another ROS ico) with the proper design (size, transparency)
- Fix in all lang files

image


KyleKatarn31:
Thanks !

PR #2697, opened 2020-04-29
[SMSS] Fix PROCESSOR_IDENTIFIER environment variable assignment

KyleKatarn31:
## Purpose

- Current PROCESSOR_IDENTIFIER Environment Variable is "truncated" and does not correctly show the Vendor name. This is due to an incorrect string concatenation.

HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Session Manager\Environment" is truncated while HKEY_LOCAL_MACHINE\HARDWARE\Description\System\CentralProcessor\0 is correct and contains the full, relevant data (VendorIdentifier)

JIRA issue: [CORE-16986](https://jira.reactos.org/browse/CORE-16986)

image

## Proposed changes

- Use a correct concatenation to have the full text stored in registry and Environement Variable

image

### DIFF
--- a/base/system/smss/sminit.c
+++ b/base/system/smss/sminit.c
@@ -1819,9 +1819,9 @@ SmpCreateDynamicEnvironmentVariables(VOID)
     if (NT_SUCCESS(Status))
     {
         /* To combine it into a single string */
-        swprintf((PWCHAR)PartialInfo->Data + wcslen((PWCHAR)PartialInfo->Data),
-                 L", %S",
-                 PartialInfo2->Data);
+        ((PWCHAR)(PartialInfo->Data))[PartialInfo->DataLength / sizeof(WCHAR)] = L'\0';        
+        wcsncat((PWCHAR)PartialInfo->Data, L", ", 2 * sizeof(WCHAR));
+        wcsncat((PWCHAR)PartialInfo->Data, (PWCHAR)PartialInfo2->Data, PartialInfo2->DataLength / sizeof(WCHAR));		
     }
 
     /* So that we can set this as the PROCESSOR_IDENTIFIER variable */


base/system/smss/sminit.c
```diff
@@ -1819,9 +1819,8 @@ SmpCreateDynamicEnvironmentVariables(VOID)
     if (NT_SUCCESS(Status))
     {
         /* To combine it into a single string */
-        swprintf((PWCHAR)PartialInfo->Data + wcslen((PWCHAR)PartialInfo->Data),
-                 L", %S",
-                 PartialInfo2->Data);
+        wcsncat((PWCHAR)PartialInfo->Data, L", ",2 * sizeof(WCHAR));
```
HBelusca, line 1822:
```suggestion
        wcsncat((PWCHAR)PartialInfo->Data, L", ", 2 * sizeof(WCHAR));
```

KyleKatarn31:
> %.*S

@HBelusca ok but using the code below leads to the same situation where only the 1st character ("G") is appended.

swprintf((PWCHAR)PartialInfo->Data + wcslen((PWCHAR)PartialInfo->Data),
		         L", %.*S",
		         ResultLength,
		         PartialInfo2->Data);

Do you have any other iteam in mind ?

learn-more:
> > %.*S
> 
> @HBelusca ok but using the code below leads to the same situation where only the 1st character ("G") is appended.
> 
> swprintf((PWCHAR)PartialInfo->Data + wcslen((PWCHAR)PartialInfo->Data),
> L", %.*S",
> ResultLength,
> PartialInfo2->Data);
> 
> Do you have any other iteam in mind ?

It is not a null-terminated string.
You use wcslen.
How is that supposed to work?

KyleKatarn31:
> 
> 
> > > %.*S
> > 
> > 
> > @HBelusca ok but using the code below leads to the same situation where only the 1st character ("G") is appended.
> > swprintf((PWCHAR)PartialInfo->Data + wcslen((PWCHAR)PartialInfo->Data),
> > L", %.*S",
> > ResultLength,
> > PartialInfo2->Data);
> > Do you have any other iteam in mind ?
> 
> It is not a null-terminated string.
> You use wcslen.
> How is that supposed to work?

It's the current design i'm trying to fix :-)
So "you" means "ROS dev before me".

I'll force 0 char at the right position, following value given by NtQueryValueKey

base/system/smss/sminit.c
KyleKatarn31: ok

base/system/smss/sminit.c
```diff
@@ -1819,9 +1819,8 @@ SmpCreateDynamicEnvironmentVariables(VOID)
...
+        wcsncat((PWCHAR)PartialInfo->Data, (PWCHAR)PartialInfo2->Data, ResultLength);
```
HBelusca, line 1823: You should not use `ResultLength` , but instead `PartialInfo2->DataLength` (which is in bytes, so if the corresponding data is a UNICODE string, you would need to use `PartialInfo2->DataLength / sizeof(WCHAR)` for the length in characters).

KyleKatarn31: Done

KyleKatarn31:
> 
> 
> > > %.*S
> > 
> > 
> > @HBelusca ok but using the code below leads to the same situation where only the 1st character ("G") is appended.
> > swprintf((PWCHAR)PartialInfo->Data + wcslen((PWCHAR)PartialInfo->Data),
> > L", %.*S",
> > ResultLength,
> > PartialInfo2->Data);
> > Do you have any other iteam in mind ?
> 
> It is not a null-terminated string.
> You use wcslen.
> How is that supposed to work?

Now null-terminated to be safe.

base/system/smss/sminit.c
```diff
@@ -1819,9 +1819,9 @@ SmpCreateDynamicEnvironmentVariables(VOID)
     if (NT_SUCCESS(Status))
     {
         /* To combine it into a single string */
-        swprintf((PWCHAR)PartialInfo->Data + wcslen((PWCHAR)PartialInfo->Data),
-                 L", %S",
-                 PartialInfo2->Data);
+        ((PWCHAR)(PartialInfo->Data))[PartialInfo->DataLength / sizeof(WCHAR)] = L'\0';
```
SergeGautherie, line 1822: No need for extra parentheses.
`UNICODE_NULL` is usually added after filling the string.

base/system/smss/sminit.c
```diff
@@ -1819,9 +1819,9 @@ SmpCreateDynamicEnvironmentVariables(VOID)
     if (NT_SUCCESS(Status))
     {
         /* To combine it into a single string */
-        swprintf((PWCHAR)PartialInfo->Data + wcslen((PWCHAR)PartialInfo->Data),
-                 L", %S",
-                 PartialInfo2->Data);
+        ((PWCHAR)(PartialInfo->Data))[PartialInfo->DataLength / sizeof(WCHAR)] = L'\0';        
+        wcsncat((PWCHAR)PartialInfo->Data, L", ", 2 * sizeof(WCHAR));
```
SergeGautherie, line 1823: Simply use `wcscpy()`.

KyleKatarn31: How is it superior to wcsncat when the intended action is combining strings ?
wcsncat based implementation works perfectly, while previous design based on swprintf fails.
What would be the benefit from doing this just in a different manner ?

KyleKatarn31:
Any futher comment on this fix ?

base/system/smss/sminit.c
SergeGautherie: These 2 lines are overly complicated/confusing:
writing a `UNICODE_NULL` is useless as you overwrite it immediately after,
using `wcsncat` for a fixed length string is overkill,
moreover it expects a count of characters, not bytes,
...

As you were told, `swprintf` itself was just fine, the issue was `wcslen` use.

KyleKatarn31: "overly complicated/confusing" : Looks more like a self-esteem problem given the simplicity of the proposed design :-)

All jokes appart, i'll rework again with swprintf if this is a "must" to get it merged into master.

SergeGautherie: > jokes

Maybe that is part of the problem: it seems you rewrote these lines multiple times, yet new code is (still) incorrect! And you don't seem to get the point :-(

KyleKatarn31: If you were thinking about 
swprintf((PWCHAR)PartialInfo->Data + PartialInfo->DataLength/sizeof(WCHAR)-1,
           		L", %*S",
         		PartialInfo2->DataLength/sizeof(WCHAR),
			PartialInfo2->Data);

It does not work %*S just adds "blanks" before PartialInfo2->Data first character and %S just prints the 1st character, as current broken design.

Since the proposed design if fit-for-purpose, simple, safe and with an adequate level of performance with regard to the need, I suggest to go with it.

Should a dev prefer to overwrite and commit a proposal of it's own based on swprintf or any different design, with a proper justification, i'm perfectly fine.

So @SergeGautherie, @HBelusca either
- You approve my code and merge it into master 
- You commit a different code and merge it into master 
- You place this PR on hold and keep the current broken code, which is in any way worse than any above options.

KyleKatarn31:
Following today's discussion in Dev channel, I will not modify this code.
Either you merge as is, or you close and discard.

base/system/smss/sminit.c
```diff
@@ -1819,9 +1819,9 @@ SmpCreateDynamicEnvironmentVariables(VOID)
...
+        wcsncat((PWCHAR)PartialInfo->Data, (PWCHAR)PartialInfo2->Data, PartialInfo2->DataLength / sizeof(WCHAR));
```
HBelusca, line 1824: So the alternative:
```suggestion
        swprintf(((PWCHAR)PartialInfo->Data) + PartialInfo->DataLength / sizeof(WCHAR),
                 L", %.*S",
                 PartialInfo2->Data, PartialInfo2->DataLength / sizeof(WCHAR));
```
would not work?

KyleKatarn31: I let you test and commit. I will no longer work on ReactOS PR's until further notice.

HBelusca: Note that I've used small s instead of capital S in the swprintf string formatter, because according to all the comments made so far it appears that the string retrieved in the PartialInfo2 is already in unicode (and not in ANSI as was supposed in the original code, hence the truncation that appeared to happen).

HBelusca, line 1824: So the alternative:
```suggestion
        swprintf(((PWCHAR)PartialInfo->Data) + PartialInfo->DataLength / sizeof(WCHAR),
                 L", %.*s",
                 PartialInfo2->DataLength / sizeof(WCHAR), PartialInfo2->Data);
```
would not work?

KyleKatarn31: Does not work, only 1 string is displayed : 
image

KyleKatarn31:
> 
> 
> Use StringCbCatW or StringCchCatW instead please.

Request is of course for @HBelusca.

HBelusca:
Superseded by minimal fix in commit 75615293.

KyleKatarn31:
Problem confirmed fixed.

PR #2699, opened 2020-04-30
[DESK] Set some flags for Browse for Wallpaper dialog

JaredSmudde:
This patch makes some changes to the browse for wallpaper dialog. It sets and uses the lpstrInitialDir flag and sets it to the My Pictures folder. Please note Server 2003 doesn't have a My Pictures folder and instead sets the default folder to My Documents. This patch also adds the OFN_EXPLORER so the Open file dialog uses the XP/2003 style dialog instead of the old style dialog.

Photos below!
Before
after

### DIFF
--- a/dll/cpl/desk/background.c
+++ b/dll/cpl/desk/background.c
@@ -635,7 +635,7 @@ OnBrowseButton(HWND hwndDlg, PBACKGROUND_DATA pData)
     OPENFILENAME ofn;
     TCHAR filename[MAX_PATH];
     TCHAR fileTitle[256];
-	TCHAR initialDir[MAX_PATH];
+    TCHAR initialDir[MAX_PATH];
     LPTSTR filter;
     LPTSTR extensions;
     BackgroundItem *backgroundItem = NULL;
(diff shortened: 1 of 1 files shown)


JaredSmudde:
Excellent idea! Hopefully the latest commit is ok.

dll/cpl/desk/background.c
```diff
@@ -652,6 +653,7 @@ OnBrowseButton(HWND hwndDlg, PBACKGROUND_DATA pData)
 
     hwndBackgroundList = GetDlgItem(hwndDlg, IDC_BACKGROUND_LIST);
     himl = ListView_GetImageList(hwndBackgroundList, LVSIL_SMALL);
+    SHGetFolderPathW(NULL, CSIDL_MYPICTURES, NULL, 0, initialDir);
```
KyleKatarn31, line 656: What if SHGetFolderPathW fails ?

dll/cpl/desk/background.c
```diff
@@ -652,6 +653,7 @@ OnBrowseButton(HWND hwndDlg, PBACKGROUND_DATA pData)
 
     hwndBackgroundList = GetDlgItem(hwndDlg, IDC_BACKGROUND_LIST);
     himl = ListView_GetImageList(hwndBackgroundList, LVSIL_SMALL);
```
KyleKatarn31, line 655: what if himl is NULL ?

KyleKatarn31:
katahiromz : you approved the change while I have open comments on the code. Does this mean that you disregard them ?

KyleKatarn31:
@binarymaster @katahiromz : Why rushing into merge while some comments on lack or error management on SHGetFolderPathW and ListView_GetImageList ? Code fails if NULL is returned.

I see a much lower level of expectation here compared to PR i'm proposing....

dll/cpl/desk/background.c
JaredSmudde: Would this be acceptable? I found that desk uses this exact method a little further down in the code.
if (FAILED(SHGetFolderPath(NULL, CSIDL_MYPICTURES, NULL, 0, initialDir)))
    {
        return;
    }

dll/cpl/desk/background.c
JaredSmudde: Not sure how I would error check this, I've found other ROS code that doesn't take this into account.

PR #2702, opened 2020-04-30
[SYSSETUP] Improve SetupQueryRegisteredOsComponent export CORE-16784

oleg-dubinskiy:
## Purpose

Improve an export for SetupQueryRegisteredOsComponent function in syssetup by using `stdcall -stub` instead of `stub`. Also, since this function is completely undocumented from MS side and there is no even any unofficial prototype, we don't know the function's arguments at all, so just use `()` without arguments then.
Required by Windows Media Encoder 9 installer from Rapps. This fix avoids the rundll32 crashes during WME9 installation. So now it installs without any errors.
Please note that it will install successfully only without VirtualBox Guest Additions installed. If they are installed, then installer will hang at the start of the installation due to another unrelated problem in netapi32. See the ticket description for details.

JIRA issue: [CORE-16784](https://jira.reactos.org/browse/CORE-16784)

## Result

Before:
wme9_no-GA_1
After:
wme9_no-GA_3_success

### DIFF
--- a/dll/win32/syssetup/syssetup.spec
+++ b/dll/win32/syssetup/syssetup.spec
@@ -62,7 +62,7 @@
 @ stub SetupOobeInitPostServices
 @ stub SetupOobeInitPreServices
 @ stub SetupPidGen3
-@ stdcall -stub SetupQueryRegisteredOsComponent()
+@ stdcall -stub SetupQueryRegisteredOsComponent(ptr ptr ptr)
 @ stub SetupQueryRegisteredOsComponentsOrder
 @ stub SetupReadPhoneList
 @ stub SetupSetAdminPassword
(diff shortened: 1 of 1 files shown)


dll/win32/syssetup/syssetup.spec
```diff
@@ -62,7 +62,7 @@
 @ stub SetupOobeInitPostServices
 @ stub SetupOobeInitPreServices
 @ stub SetupPidGen3
-@ stub SetupQueryRegisteredOsComponent
+@ stdcall -stub SetupQueryRegisteredOsComponent()
```
learn-more, line 65:
```suggestion
@ stdcall -stub SetupQueryRegisteredOsComponent(ptr ptr ptr)
```
This might not be entirely correct, but at least the amount of arguments should be fine :)

oleg-dubinskiy: Committed. Thanks @learn-more.
If this is more correct, I'm OK with this change. You know better. :)

PR #2703, opened 2020-04-30
[SHELL32] Return ERROR_INVALID_DRIVE

giawa:
## Purpose

Shell32 currently returns S_OK when browsing to a drive that does not exist.  This causes browseui to allow browsing to non-available drive letters.

beforePatch

The purpose of this PR is to address this issue by returning ERROR_INVALID_DRIVE if the drive is either unknown or unmounted.

## Proposed changes

Check to see if the drive type is unknown or unmounted by making a call to `GetDriveTypeW` while validating arguments in `CDrivesFolder`.  I'm not 100% sure if this is the correct place to do this check, so I'm definitely open to suggestions on changing how/where this is done.

## Working Screenshot

afterPatch

## TODO

- Invalid network locations continue to behave incorrectly.  I think network support in general is limited, so modifying this will take some future work.

### DIFF
--- a/dll/win32/shell32/folders/CDrivesFolder.cpp
+++ b/dll/win32/shell32/folders/CDrivesFolder.cpp
@@ -573,6 +573,7 @@ HRESULT WINAPI CDrivesFolder::ParseDisplayName(HWND hwndOwner, LPBC pbc, LPOLEST
     HRESULT hr = E_INVALIDARG;
     LPCWSTR szNext = NULL;
     LPITEMIDLIST pidlTemp = NULL;
+    WCHAR volumePathName[MAX_PATH];
 
     TRACE("(%p)->(HWND=%p,%p,%p=%s,%p,pidl=%p,%p)\n", this,
           hwndOwner, pbc, lpszDisplayName, debugstr_w (lpszDisplayName),
@@ -588,6 +589,13 @@ HRESULT WINAPI CDrivesFolder::ParseDisplayName(HWND hwndOwner, LPBC pbc, LPOLEST
 
     if (PathGetDriveNumberW(lpszDisplayName) < 0)
         return E_INVALIDARG;
+
+    /* check if this drive actually exists */
+    if (!GetVolumePathNameW(lpszDisplayName, volumePathName, _countof(volumePathName)) ||
+        GetDriveTypeW(volumePathName) < DRIVE_REMOVABLE)
+    {
+        return HRESULT_FROM_WIN32(ERROR_INVALID_DRIVE);
+    }
 
     pidlTemp = _ILCreateDrive(lpszDisplayName);
     if (!pidlTemp)


dll/win32/shell32/folders/CDrivesFolder.cpp
```diff
@@ -573,6 +573,7 @@ HRESULT WINAPI CDrivesFolder::ParseDisplayName(HWND hwndOwner, LPBC pbc, LPOLEST
     HRESULT hr = E_INVALIDARG;
     LPCWSTR szNext = NULL;
     LPITEMIDLIST pidlTemp = NULL;
+    CComHeapPtr<WCHAR> volumePathName;
```
katahiromz, line 576: `WCHAR volumePathName[MAX_PATH];`

giawa: Sounds good!  I've forced pushed that change.

giawa:
Updated after feedback from katahiromz.

dll/win32/shell32/folders/CDrivesFolder.cpp
Fraizeraust: What benefits does an array of fixed size of MAX_PATH? Wasn't a pointer just enough?
I may be wrong but nowadays certain paths can exceed above the size of a MAX_PATH. Dynamic buffer allocation is more concise in this matter.

giawa: I'm happy either way, although the original allocation I had was for MAX_PATH as well.  In this case, volumePathName only stores the path to the volume, and ignores the remainder of the path.  So this code will only fail if the volume name is larger than MAX_PATH (also since we pass MAX_PATH into GetVolumePathNameW it should return an error code if the buffer is not large enough).  The safest would be to get the length of the volume name some how, but I'm not sure if there's a good way to do that.

A future refactor should be to ensure we support up to 32,768 character paths when using the Unicode functions, since I think the Unicode versions support much longer path names.

dll/win32/shell32/folders/CDrivesFolder.cpp
```diff
@@ -589,6 +590,13 @@ HRESULT WINAPI CDrivesFolder::ParseDisplayName(HWND hwndOwner, LPBC pbc, LPOLEST
     if (PathGetDriveNumberW(lpszDisplayName) < 0)
         return E_INVALIDARG;
 
+    /* check if this drive actually exists */
+    if (!GetVolumePathNameW(lpszDisplayName, volumePathName, MAX_PATH) ||
```
learn-more, line 594: Since you are using a buffer now, it would make sense to use `_countof(volumePathName)` or `ARRAYSIZE` (depending what works in this binary.

giawa: `_countof` did the trick!  I've pushed new code

JoachimHenze:
giawa, someone identified a regression introduced by this commit, see https://jira.reactos.org/browse/CORE-17041
Eventually merger Katayama can help.

giawa:
@JoachimHenze  Thanks for the heads up.  I'm looking into why this caused a regression now.  More info to follow.

giawa:
I cannot reproduce this locally (used executable provided in JIRA).  I'll reach out to the reporter directly to try to get more information and will follow up here.  The installer never even triggered the code that was introduced by this git commit when I attempted it.  I wonder if it is actually happening at the 'history conversion' step, and since I don't have a prior install that affects it?  I tried installing it again over top of the current install and that didn't change anything.

PR #2706, opened 2020-05-01
[SYSDM] Virtual Memory panel : Complementary fix of PR2597

KyleKatarn31:
## Purpose

- PR 2597 did fix change some thresholds related to VM settings. This was not reflected in all language files, which could have been misleading. However, this would NOT have any any functional impact since the range of acceptable values had been extended.

JIRA issue: [CORE-12486 ](https://jira.reactos.org/browse/CORE-12486 )

## Proposed changes

- Change of 16 MB to 2 MB in all lang files.
- Some complementary translation enhancement to be made by individual translators


Mas4hmad:
can you add ...`not be smaller than 2 MB,  must not be larger than 4095 MB and must`... for other language? i think it's for consistency. maybe you can reset it all to english.
in the side, your code is legit.

KyleKatarn31:
> 
> 
> can you add ...`not be smaller than 2 MB, must not be larger than 4095 MB and must`... for other language? i think it's for consistency. maybe you can reset it all to english.
> in the side, your code is legit.

Sorry but I don't talk all the language of ReactOS :-)
That's why i've mentionned "Some complementary translation enhancement to be made by individual translators"

KyleKatarn31:
thanks

PR #2708, opened 2020-05-01
[SDK][STDLIB][STRING] Wine-sync wcsrtombs_l and _mbstowcs_l

Fraizeraust:
`wctomb` is not addressed in this commit, it crashes our current testcase unit we have for that function.
Here are the test results for `mbstowcs` and `wcstombs`.
Capture
Capture2

### DIFF
--- a/sdk/lib/crt/stdlib/mbstowcs.c
+++ b/sdk/lib/crt/stdlib/mbstowcs.c
@@ -1,7 +1,7 @@
 #include <precomp.h>
 
 /*********************************************************************
- * _mbstowcs_l
+ *		_mbstowcs_l (MSVCRT.@)
  */
 size_t CDECL _mbstowcs_l(wchar_t *wcstr, const char *mbstr,
         size_t count, _locale_t locale)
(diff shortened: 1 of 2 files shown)


sdk/lib/crt/stdlib/mbstowcs.c
```diff
@@ -1,19 +1,35 @@
 #include <precomp.h>
 
 /*********************************************************************
- * _mbstowcs_l
+ *		_mbstowcs_l (MSVCRT.@)
```
HBelusca, line 4: the tab is also from wine?

Fraizeraust: Yes.

sdk/lib/crt/string/wcs.c
```diff
@@ -257,7 +257,8 @@ INT CDECL wctomb( char *dst, wchar_t ch )
 /*********************************************************************
  * wcsrtombs_l (INTERNAL)
  */
-static size_t CDECL wcsrtombs_l(char *mbstr, const wchar_t **wcstr, size_t count, _locale_t locale)
+static size_t CDECL wcsrtombs_l(char *mbstr, const wchar_t **wcstr,
+            size_t count, _locale_t locale)
 {
     MSVCRT_pthreadlocinfo locinfo;
     size_t tmp = 0;
```
HBelusca, line 264: It is interesting to note that in this function Wine does no check for wcstr being NULL.

PR #2710, opened 2020-05-02
[VIDEOPRT] Remove confusing 'else' after 'return'. CORE-10146

SergeGautherie:
## Purpose

Revert a hunk of b82bf8ce167ae96dff2868461e82136b90b8c45f.

JIRA issue: [CORE-10146](https://jira.reactos.org/browse/CORE-10146)


Extravert-ir:
This is not "confusing" at all

SergeGautherie:
> not "confusing"

Well,
`else` is redundant/useless after a `return`,
the first `if` is obviously not part of the following ones which act as a `switch`.

PR #2712, opened 2020-05-02
[FREELDR] machpc.c: Increase serial port length to 8 from 7

SergeGautherie:
See for example:

* [I/O Ports and Devices](https://www.pearsonitcertification.com/articles/article.aspx?p=1681059) Table 3-1.
* https://en.wikipedia.org/wiki/COM_(hardware_interface)#I/O_addresses
`This IC has seven internal 8-bit registers which hold information and configuration data about which data is to be sent or was received, the baud rate, interrupt configuration and more. In the case of COM1, these registers can be accessed by writing to or reading from the I/O addresses 0x3F8 to 0x3FF.`
* https://www.tldp.org/HOWTO/Serial-HOWTO-8.html#ss8.9
`The addresses shown below represent the first address of an 8-byte range. For example 3f8 is really the range 3f8-3ff.`


SergeGautherie:

binarymaster:
Merged in commit https://github.com/reactos/reactos/commit/468b8f3719aa6794e4976b746956b1753b7be2eb. Thanks!

PR #2715, opened 2020-05-02
[RPCRT4] Export NdrCreateServerInterfaceFromStub as `stub`

oleg-dubinskiy:
## Purpose

Add export for NdrCreateServerInterfaceFromStub function in our rpcrt4.
Required by MS ole32.dll. Since this function is undocumented, and we don't know its exact arguments, use just `stub` for an export. Moreover, it's enough to allow MS ole32 no longer fail.

JIRA issue: [CORE-15395](https://jira.reactos.org/browse/CORE-15395), [CORE-17004](https://jira.reactos.org/browse/CORE-17004).

## Result

Before:
MS_ole32_before

After:
MS_ole32_after

### DIFF
--- a/dll/win32/rpcrt4/rpcrt4.spec
+++ b/dll/win32/rpcrt4/rpcrt4.spec
@@ -179,7 +179,7 @@
 179 stdcall NdrCorrelationFree(ptr)
 180 stdcall NdrCorrelationInitialize(ptr ptr long long)
 181 stdcall NdrCorrelationPass(ptr)
-# NdrCreateServerInterfaceFromStub
+182 stub NdrCreateServerInterfaceFromStub
 183 stub NdrDcomAsyncClientCall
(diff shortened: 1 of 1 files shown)


SergeGautherie:
>  we don't know its exact arguments

Not quite:
PSDK S03 already has it
```
RPC_STATUS RPC_ENTRY
NdrCreateServerInterfaceFromStub(
            IN struct IRpcStubBuffer* pStub,
            IN OUT RPC_SERVER_INTERFACE *pServerIf );
```

oleg-dubinskiy:
Thanks @SergeGautherie. Didn't know that.
So it is possible to use `stdcall -stub` or even add a code stub, but if just `stub` is enough to fix the crash, I think it is not necessarily to do more.

SergeGautherie:
> not necessarily to do more

Agreed: that is usual ROS policy.

PR #2716, opened 2020-05-02
[FREELDR] pchw.c: Split floppy port length into 6+1 from 8

SergeGautherie:
`0x03F6` is IDE control port.


SergeGautherie:

SergeGautherie:
> Screenshot please.

Added.

disean:
This PR breaks compatibility with NTLDR and does not fix anything.

> Confirmed on XP x86, on VPC 2004.

You should learn about IRP_MN_FILTER_RESOURCE_REQUIREMENTS.

PR #2718, opened 2020-05-02
[FREELDR] machpc.c: Increase parallel port length to 8 from 3

SergeGautherie:
I could not find any "explicit" documentation about this change :-|

### DIFF
--- a/boot/freeldr/freeldr/arch/i386/pc/machpc.c
+++ b/boot/freeldr/freeldr/arch/i386/pc/machpc.c
@@ -809,7 +809,7 @@ DetectParallelPorts(PCONFIGURATION_COMPONENT_DATA BusKey)
         PartialDescriptor->Flags = CM_RESOURCE_PORT_IO;
         PartialDescriptor->u.Port.Start.LowPart = Base;
         PartialDescriptor->u.Port.Start.HighPart = 0x0;
-        PartialDescriptor->u.Port.Length = 3;
+        PartialDescriptor->u.Port.Length = 8;
 
         /* Set Interrupt */
         if (Irq[i] != (ULONG) - 1)


SergeGautherie:

learn-more:
> I could not find any "explicit" documentation about this change :-|

So, then why is this changed?
What does it fix?
How did you find this?

SergeGautherie:
> So, then why is this changed?
> What does it fix?
> How did you find this?

Check the XP picture?

mifritscher:
Please be careful: 

It seems that a standard (SPP) LPT has indeed only 3 ports: https://www.tldp.org/HOWTO/pdf/IO-Port-Programming.pdf / https://wiki.osdev.org/Parallel_port .

But a EPP or ECP LPT has indeed more ports: https://www.sciencedirect.com/topics/engineering/parallel-port .

The parport driver from linux seems to get the extend (=length) automatically: https://tldp.org/LDP/khg/HyperNews/get/devices/parport.html

Here is the source: https://github.com/torvalds/linux/blob/9aa900c8094dba7a60dc805ecec1e9f720744ba1/drivers/parport/parport_pc.c right at the beginning you see the port assignment.

At this code, look into the function parport_pc_probe_port() (line 2003). Here you can see the logic to to get the mode. Attention, in most cases it first request the extended region(line 2074), but releases it later if the port has no EPP capabilites (line 2167).

disean:
This PR breaks compatibility with NTLDR and does not fix anything.

> Confirmed on XP x86, on VPC 2004.

You should learn about IRP_MN_FILTER_RESOURCE_REQUIREMENTS.

tkreuzer (changes requested):
Don't try to "fix" things that you don't understand.

PR #2723, opened 2020-05-02
[REGAPI] Import regapi.dll from Wine Staging 5.7 CORE-16458

oleg-dubinskiy:
## Purpose

Import regapi.dll stub from Wine Staging 5.7. There are no any differences between Devel and Staging versions of regapi, so both of them are identical.
Required by MS winlogon with Win32SS replaced. I double-rechecked: now it doesn't fail due to missing regapi.dll in system32 or separate missing functions in this dll, exactly the same as with MS regapi. In pair with my profmap.dll (which was merged recently), now it fails due to InitializeUserProfile() stub in our userenv.dll (if to replace msgina too).
Also I compared the exports of MS regapi.dll and Wine's one. Win2k3 version has all exports that Wine version has and some others (which looks like were dropped in NT6). But those exports which already exist, are enough to allow MS winlogon no longer fail.

JIRA issue: [CORE-16458](https://jira.reactos.org/browse/CORE-16458)

## Proposed changes

- Add dll/win32/regapi directory with CMakeLists.txt and Wine's main.c and regapi.spec source files;
- Enable all this stuff in dll/win32/CMakeLists.txt;
- Add regapi entry in media/doc/README.WINE;
- Additionally, add a comments for some other missing exports, which Win2k3's regapi has, but Wine's regapi hasn't.

Win2k3 has this dll out of the box as well:
regapi_win2k3

DLL purpose: Registry Configuration APIs.


SergeGautherie:
> Win2k3 version has all exports that Wine version has

What is the policy: enable all stubs or only the needed ones?

> and some others (which looks like were dropped in NT6).

Optionally, maybe add these other ones (as comments)?

oleg-dubinskiy:
> What is the policy: enable all stubs or only the needed ones?

Ideally I'd prefer to add all stubs (at least as comments), to match our exports with 2k3 ones, but since this stuff is winesynced, and all current stubs are enough to avoid the failure, probably it will be better to keep it as-is. Or we don't sync *.spec files with Wine, only *.c? :thinking:

SergeGautherie:
> Or we don't sync *.spec files with Wine, only *.c?

Indeed, .spec files are special cases: they are ReactOS specific actually, though they can be (partially) "sync'ed" as need be.

oleg-dubinskiy:
> .spec files are special cases: they are ReactOS specific actually, though they can be (partially) "sync'ed" as need be.

Oh, OK. Thanks for the tip. Then I'll add all missing exports (as comments).

oleg-dubinskiy:
Oops, I was a bit wrong. :sweat_smile: As it appeared during my recent tests, even with imported Wine's regapi, after replacing userenv.dll, MS winlogon still fails due to regapi RegWinStationQueryW() stub. So problem is still not fixed, but at least the situation now improved. Probably using `stdcall -stub` may help, but I'm not sure about that. I will do it in a separate PR if needed. Sorry for my mistake.
UPD: Oh, I found it. MS winlogon needs only 3 functions from regapi.dll:
MS_winlogon
So using `stdcall -stub` for them most of all may fix the problem.

PR #2725, opened 2020-05-03
[NTDLL] Improve EtwRegisterTraceGuidsW() stub and related, based on Wine

SergeGautherie:
`(dll/ntdll/etw/trace.c:125) EtwRegisterTraceGuidsW stub()`
is actually called:
see logs on CORE-13878 and CORE-16996, for example.

Then, improve it somewhat:
* Adapt Wine stub code.
* Additions based on MS docs.


learn-more:
'based on wine'

wmi.spec is touched for the sole purpose of making it different from wine?

PR #2728, opened 2020-05-04
[SHELL32] Let the run dialog use the newer style browse dialog

JaredSmudde:
Apply the OFN_EXPLORER flag to the Run dialog so the dialog uses the newer style browse dialog.

Why? Because it's a whole lot easier and quicker to browse for files with the newer style dialog. :)

Before
after


SergeGautherie:
For the record, is that what S03 uses?

JaredSmudde:
Yes it is.

PR #2730, opened 2020-05-04
[OSK] Handle the warning dialog box in its own thread

GeoB99:


learn-more:
What problem is solved here?

Fraizeraust:
The warning dialog box is the first to be displayed and as such the user has to interact with the said dialog (i.e. click the OK button) for the main OSK dialog to be shown. In Windows XP's OSK, both the warning box and the main OSK window are displayed at the same time unlike in our OSK.

I guess I should've worded the commit description way better.

base/applications/osk/main.c
```diff
@@ -808,10 +826,14 @@ int WINAPI wWinMain(HINSTANCE hInstance,
...
     LoadSettings();
 
-    /* If the member of the struct (bShowWarning) is set then display the dialog box */
-    if (Globals.bShowWarning)
+    /* If for whatever reason the thread fails to be created then handle the dialog box in main thread... */
+    if (CreateThread(NULL, 0, (LPTHREAD_START_ROUTINE)OSK_WarningDlgThread, (PVOID)Globals.hInstance, 0, NULL) == NULL)
     {
-        DialogBoxW(Globals.hInstance, MAKEINTRESOURCEW(IDD_WARNINGDIALOG_OSK), Globals.hMainWnd, OSK_WarningProc);
+        /* If the member of the struct (bShowWarning) is set then display the dialog box */
+        if (Globals.bShowWarning)
```
HBelusca, line 833: You may be able to simplify slightly more all this, and just have one single `if (Globals.bShowWarning)` that if it's true, then you either launch the thread or display the dialog from main thread.
This would also avoid the `bShowWarning` check in the thread implementation.

learn-more:
If you remove the parent from the dialogbox, and launch it with `CreateDialog` instead of `DialogBox`, does that also solve your problem?

Fraizeraust:
> If you remove the parent from the dialogbox, and launch it with CreateDialog instead of DialogBox, does that also solve your problem?

Creating the dialog with CreateDialog doesn't solve the problem, actually the dialog doesn't even show up at all despite invoking `ShowWindow` (note that the handle is given). 

I have to change lots of stuff here to make it work with CreateDialog so honestly I'd prefer the simpler approach with threads.

HBelusca:
CreateDialog would indeed also require to run manually a message pump.

base/applications/osk/main.c
```diff
@@ -811,7 +825,11 @@ int WINAPI wWinMain(HINSTANCE hInstance,
     /* If the member of the struct (bShowWarning) is set then display the dialog box */
     if (Globals.bShowWarning)
     {
-        DialogBoxW(Globals.hInstance, MAKEINTRESOURCEW(IDD_WARNINGDIALOG_OSK), Globals.hMainWnd, OSK_WarningProc);
+        /* If for whatever reason the thread fails to be created then handle the dialog box in main thread... */
+        if (CreateThread(NULL, 0, (LPTHREAD_START_ROUTINE)OSK_WarningDlgThread, (PVOID)Globals.hInstance, 0, NULL) == NULL)
```
ThFabba, line 829: Please do not cast function pointers.

Fraizeraust:
I revisited back this PR and tried to properly create the modeless dialog box using `CreateDialog` with message pump and all. The welcome dialog box still takes precedence despite the parent dialog not being created with `DialogBox` now, thus, this solution actually doesn't solve the problem.

PR #2732, opened 2020-05-04
[COMDLG32] Add some more places to the places bar

JaredSmudde:
Add Recent Documents and My Network Places to the places bar in the open file dialog. There is an issue with label wrapping, I've looked at it, appears to be a bug deeper in comctl32 I think.

comdlg32

### DIFF
--- a/dll/win32/comdlg32/filedlg.c
+++ b/dll/win32/comdlg32/filedlg.c
@@ -305,9 +305,17 @@ static void filedlg_collect_places_pidls(FileOpenDlgInfos *fodInfos)
 {
     static const int default_places[] =
     {
+#ifdef __REACTOS__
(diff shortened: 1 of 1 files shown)


KyleKatarn31:
Maybe to be proposed to WineHQ as well (unless it's already implemented in Wine 5.7) ?

HBelusca:
Yep I was going to suggest sending the patch to Wine.

HBelusca:
NOTE: Technically this "places bar" can be customized:
https://www.tenforums.com/tutorials/126153-change-places-bar-items-common-dialog-box-windows.html
https://www.sevenforums.com/tutorials/85487-common-file-dialog-box-customize-places-bar.html

PR #2734, opened 2020-05-04
[ADVAPI32] Fix undue debug print in nominal case

KyleKatarn31:
## Purpose

- In current implementation, when Regedit opens HKCR root key, an error is logged
err:(dll/win32/advapi32/reg/hkcr.c:964) Returning 259.
This is not correct as code 259 is for ERROR_NO_MORE_ITEMS which is the nominal return value when end of enumeration is reached.

## Proposed changes

Err is not displayed when ERROR_NO_MORE_ITEMS is returned.


KyleKatarn31:
Thanks

PR #2735, opened 2020-05-05
[ADVAPI32] Remove now-redundant comment

SergeGautherie:
Addendum to 5dfe3455db5d8c21a326178006ae8dab9c71fc4b.


KyleKatarn31:
This comment was left intentionnally since it justifies the specific if condition.

SergeGautherie:
> intentionnally

Your commit message is rather explicit/enough about that.

KyleKatarn31:
> 
> 
> > intentionnally
> 
> Your commit message is rather explicit/enough about that.

OK clear.

PR #2740, opened 2020-05-05
[NTDLL_APITEST] Expand the NtSetInformationProcess testcase with some tests

Fraizeraust:
`ProcessRaisePriority` class expects a buffer information of **KPRIORITY** type.
**Windows Server 2003**
Capture2
**Windows XP**
Capture4
**Windows 10 Home 1909**
Capture
**ReactOS**
Capture6

### DIFF
--- a/modules/rostests/apitests/ntdll/NtSetInformationProcess.c
+++ b/modules/rostests/apitests/ntdll/NtSetInformationProcess.c
@@ -9,6 +9,147 @@
 
 static
 void
+Test_ProcBasePriorityClass(void)
+{
+    NTSTATUS Status;
+
+    /*
+     * Assign a priority of HIGH_PRIORITY (see pstypes.h).
+     * The function will fail with a status of STATUS_PRIVILEGE_NOT_HELD
+     * as the process who executed the caller doesn't have the requested
+     * privileges to perform the operation.
+    */
+    KPRIORITY BasePriority = HIGH_PRIORITY;
+
+    /* Everything is NULL */
+    Status = NtSetInformationProcess(NULL,
+                                     ProcessBasePriority,
+                                     NULL,
+                                     0);
+    ok_hex(Status, STATUS_INFO_LENGTH_MISMATCH);
+
+    /* Set the base priority to an invalid process handle */
+    Status = NtSetInformationProcess(NULL,
+                                     ProcessBasePriority,
+                                     &BasePriority,
+                                     sizeof(KPRIORITY));
+    ok_hex(Status, STATUS_INVALID_HANDLE);
+
+    /* Don't input anything to the caller but the length information is valid */
+    Status = NtSetInformationProcess(NtCurrentProcess(),
+                                     ProcessBasePriority,
+                                     NULL,
+                                     sizeof(KPRIORITY));
+    ok_hex(Status, STATUS_ACCESS_VIOLATION);
+
+    /* Give the base priority input to the caller but length is invalid */
+    Status = NtSetInformationProcess(NtCurrentProcess(),
+                                     ProcessBasePriority,
+                                     &BasePriority,
+                                     0);
+    ok_hex(Status, STATUS_INFO_LENGTH_MISMATCH);
+
+    /* The input buffer is misaligned and length information invalid */
+    Status = NtSetInformationProcess(NtCurrentProcess(),
+                                     ProcessBasePriority,
+                                     (PVOID)1,
+                                     0);
+    ok_hex(Status, STATUS_INFO_LENGTH_MISMATCH);
+
+    /* The input buffer is misaligned */
+    Status = NtSetInformationProcess(NtCurrentProcess(),
+                                     ProcessBasePriority,
+                                     (PVOID)1,
+                                     sizeof(KPRIORITY));
+    ok_hex(Status, STATUS_DATATYPE_MISALIGNMENT);
+
+    /* The input buffer is misaligned (try with an alignment of 2) */
+    Status = NtSetInformationProcess(NtCurrentProcess(),
+                                     ProcessBasePriority,
+                                     (PVOID)2,
+                                     sizeof(KPRIORITY));
+    ok_hex(Status, STATUS_DATATYPE_MISALIGNMENT);
+
+    /* Set the base priority but we have lack privileges to do so */
+    Status = NtSetInformationProcess(NtCurrentProcess(),
+                                     ProcessBasePriority,
+                                     &BasePriority,
+                                     sizeof(KPRIORITY));
+    ok_hex(Status, STATUS_PRIVILEGE_NOT_HELD);
+
+    /*
+     * Assign a random priority value this time and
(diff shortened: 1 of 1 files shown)


modules/rostests/apitests/ntdll/NtSetInformationProcess.c
```diff
@@ -7,6 +7,162 @@
...
 #include "precomp.h"
 
+static
+void
+Test_ProcBasePriorityClass(void)
+{
+    NTSTATUS Status;
+    KPRIORITY *BasePriority;
+
+    BasePriority = HeapAlloc(GetProcessHeap(), HEAP_ZERO_MEMORY, sizeof(KPRIORITY));
```
HBelusca, line 17: Why not just using a regular buffer (on the stack)?

modules/rostests/apitests/ntdll/NtSetInformationProcess.c
```diff
@@ -7,6 +7,162 @@
...
+}
+
+static
+void
+Test_ProcRaisePriorityClass(void)
+{
+    NTSTATUS Status;
+    KPRIORITY *RaisePriority;
+
+    RaisePriority = HeapAlloc(GetProcessHeap(), HEAP_ZERO_MEMORY, sizeof(KPRIORITY));
```
HBelusca, line 103: Ditto.

HBelusca: Isn't it also a ULONG here?

Fraizeraust: No? **ProcessRaisePriority** wants a buffer of KPRIORITY which is a LONG but at the same time it wants a length information of ULONG.

HBelusca: Then it's simply a ULONG that it wants 😄

HBelusca:
For notice, our kernel code:
- for ProcessBasePriority:
https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/ps/query.c;hb=1f3cb13fb97e618b53e55db0f455fe7259694326#l1525

- for ProcessRaisePriority:
https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/ps/query.c;hb=1f3cb13fb97e618b53e55db0f455fe7259694326#l1589

appears to do what's necessary already.

HBelusca:
The error codes that the test running in ReactOS does mention:
```
/error 0xc0000004:
    STATUS_INFO_LENGTH_MISMATCH

/error 0xc0000008:
    STATUS_INVALID_HANDLE

/error 0x80000002:
    STATUS_DATATYPE_MISALIGNMENT
```

HBelusca:
Basically, what the test run shows on ReactOS, is that the `NtSetInformationProcess()` function misses some ProbeForRead calls, similar to what is done here:
https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/ps/query.c;hb=1f3cb13fb97e618b53e55db0f455fe7259694326#l88
(this was originally what was intended by the commented-out call to `DefaultSetInfoBufferCheck()`).

Also I've checked, and it appears as well that `NtSetInformationThread()` misses these probings too

Fraizeraust:
bf493b9 is the commit where `ps_i.h` tables were removed and c4f9542 where the probing checks were disabled.

modules/rostests/apitests/ntdll/NtSetInformationProcess.c
```diff
@@ -7,6 +7,140 @@
...
+    Status = NtSetInformationProcess(NtCurrentProcess(),
+                                     ProcessBasePriority,
+                                     (PVOID)1,
+                                     0);
+    ok_hex(Status, STATUS_INFO_LENGTH_MISMATCH);
+
+    /* The input buffer is misaligned */
+    Status = NtSetInformationProcess(NtCurrentProcess(),
+                                     ProcessBasePriority,
+                                     (PVOID)1,
```
ThFabba, line 62: For completeness, test `2` as well to show that the alignment is indeed four and not two (which is unlikely, but still...)

modules/rostests/apitests/ntdll/NtSetInformationProcess.c
```diff
@@ -7,6 +7,140 @@
...
+                                     ProcessBasePriority,
+                                     &BasePriority,
+                                     sizeof(KPRIORITY));
+    ok_hex(Status, STATUS_PRIVILEGE_NOT_HELD);
+
+    /*
+     * Assign a random priority value this time and
+     * set the base priority to the current process.
+    */
+    BasePriority = 8;
```
ThFabba, line 77: For the future: the interesting test cases are boundary values -- i.e. what is the last value that succeeds and what is the first that returns `STATUS_PRIVILEGE_NOT_HELD`?

modules/rostests/apitests/ntdll/NtSetInformationProcess.c
```diff
@@ -7,6 +7,140 @@
...
+    Status = NtSetInformationProcess(NtCurrentProcess(),
+                                     ProcessBasePriority,
+                                     &BasePriority,
+                                     sizeof(KPRIORITY));
+    ok_hex(Status, STATUS_SUCCESS);
+}
```
ThFabba, line 83: Because these tests run one after the other in the same process, each test should leave the process in the state it found it in (i.e. this function should query the priority at the beginning, then reset it back at the end). Otherwise test results might start to depend on things done by previous tests, and changing the order would change results.

modules/rostests/apitests/ntdll/NtSetInformationProcess.c
```diff
@@ -7,6 +7,147 @@
...
+    Status = NtSetInformationProcess(NtCurrentProcess(),
+                                     ProcessRaisePriority,
+                                     NULL,
+                                     sizeof(ULONG));
+    ok_hex(Status, STATUS_ACCESS_VIOLATION);
+
+    /* The input buffer is misaligned */
+    Status = NtSetInformationProcess(NtCurrentProcess(),
+                                     ProcessRaisePriority,
+                                     (PVOID)1,
```
ThFabba, line 139: Add `(PVOID)2` test?

modules/rostests/apitests/ntdll/NtSetInformationProcess.c
```diff
@@ -7,6 +7,147 @@
...
+
+    /* The input buffer is misaligned and length information invalid */
+    Status = NtSetInformationProcess(NtCurrentProcess(),
+                                     ProcessRaisePriority,
+                                     (PVOID)1,
+                                     0);
+    ok_hex(Status, STATUS_INFO_LENGTH_MISMATCH);
+
+    /* The information length is invalid */
+    Status = NtSetInformationProcess(NtCurrentProcess(),
```
ThFabba, line 123: This doesn't seem to add anything over the one just above.

PR #2750, opened 2020-05-08
[OPENGLCFG] Empty list boxes if registry key is missing

Mondgestein:
## Purpose
If the registry key that holds the names of installed
opengl drivers is missing, all list boxes in openglcfg
are empty. This is a minor code rearrangement to fix
this behavior.

### DIFF
--- a/dll/cpl/openglcfg/general.c
+++ b/dll/cpl/openglcfg/general.c
@@ -10,11 +10,11 @@ static VOID InitSettings(HWND hWndDlg)
 {
     HKEY hKeyRenderer;
     HKEY hKeyDrivers;
+    DWORD dwType = 0;
+    DWORD dwSize = MAX_KEY_LENGTH;
     WCHAR szBuffer[MAX_KEY_LENGTH];
     WCHAR szBultin[MAX_KEY_LENGTH];
     WCHAR szDriver[MAX_KEY_LENGTH];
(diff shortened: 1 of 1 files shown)


dll/cpl/openglcfg/general.c
```diff
@@ -42,6 +33,12 @@ static VOID InitSettings(HWND hWndDlg)
     LoadString(hApplet, IDS_RENDERER_RSWR, (LPTSTR)szBultin, 127);
     SendDlgItemMessageW(hWndDlg, IDC_RENDERER, CB_ADDSTRING, 0, (LPARAM)szBultin);
 
+    if (RegCreateKeyExW(HKEY_CURRENT_USER, KEY_RENDERER, 0, NULL, 0, MAXIMUM_ALLOWED, NULL, &hKeyRenderer, NULL) != ERROR_SUCCESS)
+    {
+        RegCloseKey(hKeyDrivers);
```
KyleKatarn31, line 38: Closekey called before Open key (since bloc from original L19 is moved later in code) !!

KyleKatarn31: To be resolved ?

Mondgestein: I have added another commit if you mean that.
Or should the first one modified instead?

PR #2743, opened 2020-05-06
[UTILMAN] Use the ShellAboutW API routine to display the about dialog

Fraizeraust:
## Purpose
As the dialog simply displays the copyright names and nothing particular it makes sense to use `ShellAboutW()` instead.

PR #2751, opened 2020-05-08
[RAPPS] Use the ShellAboutW API routine to display the about dialog

KyleKatarn31:
## Purpose

image

Similar to https://github.com/reactos/reactos/pull/2743 by @HBelusca 

## Proposed changes

image


base/applications/rapps/gui.cpp
```diff
@@ -1469,6 +1469,19 @@ class CMainWindow :
...
 
+    VOID ShowAboutDlg()
+    {
+        ATL::CStringW szApp;
+        ATL::CStringW szAuthors;
+        HICON hIcon;
+
+        szApp.LoadStringW(IDS_APPTITLE);
+        szAuthors.LoadStringW(IDS_APP_AUTHORS);
+        hIcon = (HICON) LoadIconW(hInst, MAKEINTRESOURCEW(IDI_MAIN));
```
Fraizeraust, line 1480: Does it make sense to cast the function as `HICON` where such function returns a handle to a icon?

base/applications/rapps/gui.cpp
```diff
@@ -1469,6 +1469,19 @@ class CMainWindow :
...
+        ShellAboutW(m_hWnd, szApp, szAuthors, hIcon);
+        DeleteObject(hIcon);
```
Fraizeraust, line 1482: Icons have their own function to destroy the icon and free up the handle from the memory space, `DestroyIcon()`. Use that instead.

base/applications/rapps/gui.cpp
KyleKatarn31: ok

base/applications/rapps/gui.cpp
KyleKatarn31: ok

base/applications/rapps/gui.cpp
```diff
@@ -1605,7 +1618,7 @@ class CMainWindow :
             break;
 
         case ID_ABOUT:
-            ShowAboutDialog();
+            ShowAboutDlg();
```
learn-more, line 1621: Why rename the function?

KyleKatarn31: Because I copy / pasted the code pattern from @HBelusca PR.
Do you want me to keep the original function name from this module ?

KyleKatarn31:
Thanks

PR #2754, opened 2020-05-08
[SNDREC32] Use the ShellAboutW API routine to display the about dialog

KyleKatarn31:
## Purpose

image

Similar to #2743 by @HBelusca

## Proposed changes

image

And correct a wrong error message text by opportunity.


Fraizeraust:
Why don't you do this in a single PR where you push it with each commit the "about" dialog related changes. Also do we have to do this for every single built-in ROS application?

KyleKatarn31:
> 
> 
> Why don't you do this in a single PR where you push it with each commit the "about" dialog related changes. Also do we have to do this for every single built-in ROS application?

I thought that the rule was "1 PR per module" with exceptions for translations only.
Isn't it better to have the standard ReactOS "About" dialog for built-in ReactOS apps ? It's already the case on vast majority of them.

Matriks404:
I don't think repeated **ReactOS Team** is needed here, as it specified by the generic text above (look at the explorer about window).

KyleKatarn31:
> 
> 
> I don't think repeated **ReactOS Team** is needed here, as it specified by the generic text above (look at the explorer about window).

Any better name/text to suggest as "authors" ?

Prince-Kassad:
Blank (pass NULL)?

Note that Windows actually passes the current sound settings into the About box for this particular application, but that's probably more a nice-to-have than a necessity.

KyleKatarn31:
> 
> 
> Blank (pass NULL)?
> 
> Note that Windows actually passes the current sound settings into the About box for this particular application, but that's probably more a nice-to-have than a necessity.

I agree that passing the sound setting here is an unnecessary feature.
OK for L"" as Author name in this case. I'll commit this change asap.

base/applications/sndrec32/sndrec32.cpp
```diff
@@ -619,8 +598,12 @@ WndProc(HWND hWnd,
                     break;
 
                 case ID_ABOUT:
-                    DialogBox(hInst, MAKEINTRESOURCE(IDD_ABOUTBOX), hWnd, AboutDlgProc);
-                    return TRUE;
+                    LoadStringW(hInst, IDS_APP_TITLE, szAppName, sizeof(szAppName) / sizeof(WCHAR));                      
+                    LoadStringW(hInst, IDS_APP_AUTHORS, szAppAuthors, sizeof(szAppAuthors) / sizeof(WCHAR));
```
learn-more, line 602: Why load an empty string and pass it to ShellAbout, instead of passing `NULL`?

KyleKatarn31: I initially populated "IDS_APP_AUTHORS" with ReactOS Team, then got the comment that it was not useful. Therefore I set it to an empty string but kept the code, in case with later have a proper Author to quote for this module.

Would you prefer to have this part of code removed, replaced by NULL and IDS_APP_AUTHORS definition discarded as well ?

learn-more: Yes, adding dead code to the codebase 'in case someone wants to change something later' is not really a good idea.

KyleKatarn31: OK, i'll refactor.

KyleKatarn31: Done !

base/applications/sndrec32/sndrec32.cpp
```diff
@@ -118,8 +94,8 @@ _tWinMain(HINSTANCE hInstance,
     s_info.cbSize = sizeof( NONCLIENTMETRICS );
 
     InitCommonControls();
-	
-	switch (GetUserDefaultUILanguage())
+    
+    switch (GetUserDefaultUILanguage())
```
HBelusca, line 98:
```suggestion

    switch (GetUserDefaultUILanguage())
```

KyleKatarn31:
Thanks

PR #2755, opened 2020-05-08
[NDK][XDK] Add 3 CM_RESOURCE_INTERRUPT_* defines

SergeGautherie:
* Add 3 defines from SDK 10.
* Sync 2 defines from XDK.

Noticed while checking CORE-9922.

NB:
I don't know whether any of these need to be duplicated. Better safe than sorry...


SergeGautherie:
> windows 8+

All I can tell is none of these were in PSDK S03R2.

sdk/include/xdk/cmtypes.h
```diff
@@ -259,6 +259,9 @@ typedef enum _CM_SHARE_DISPOSITION {
 #define CM_RESOURCE_INTERRUPT_LATCHED         0x0001
 #define CM_RESOURCE_INTERRUPT_MESSAGE         0x0002
 #define CM_RESOURCE_INTERRUPT_POLICY_INCLUDED 0x0004
+#define CM_RESOURCE_INTERRUPT_ALLOW_RESERVED_IDT    0x0008
```
ThFabba, line 262: The others are fine, this isn't in MS's wdm.h. Only in `miniport.h`.

SergeGautherie: Shall I comment it out, with a comment? Or simply remove it?
In both files?

HBelusca: @ThFabba : I think we would need something like `$ifndef (_WDMDDK_)` or similar, but that hpp command doesn't exist.

ThFabba: There's no other XDK-generated header that it does go in. So an `ifdef` doesn't seem to make sense.
I'd say just remove it from here and keep it in NDK.

SergeGautherie: Removed from XDK.

PR #2756, opened 2020-05-08
[USETUP] Compute the exact column line count when highlighting the te…

GeoB99:
…xt in boot loader page

**BEFORE**
VirtualBox_ReactOS_08_05_2020_15_26_45
**AFTER**
Capture

### DIFF
--- a/base/setup/usetup/mui.c
+++ b/base/setup/usetup/mui.c
@@ -304,6 +304,140 @@ MUIGetEntry(
 }
 
 /**
+ * @MUIGetEntryTextFromCoord
+ *
+ * Retrieves the portion of text from a MUI entry given the X and Y coordinate points.
+ *
+ * @param[in]   Page
+ *     The MUI (Multilingual User Interface) entry page number, as a unsigned long integer.
+ *
+ * @param[in]   X
+ *      The X coordinate where the text belongs.
+ *
+ * @param[in]   Y
+ *      The Y coordinate where the text belongs.
+ *
+ * @return
+ *     Returns a pointer to a constant string pointed by the MUI entry.
+ *
+ */
+LPCSTR
+MUIGetEntryTextFromCoord(
+    IN ULONG Page,
+    IN SHORT X,
+    IN SHORT Y)
+{
+    const MUI_ENTRY * entry;
+    ULONG index;
+
+    /* Retrieve the entries of a MUI page */
+    entry = FindMUIEntriesOfPage(Page);
+    if (!entry)
+    {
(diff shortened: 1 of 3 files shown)


binarymaster:
Suggestion: Loop through all menu items, calculate maximum length, and use it.

Fraizeraust:
I'd rather prefer to fill the text with white background based on its length (similarly done when you choose a FS in USETUP for format) than unnecessarily filling more white space for shorter text like the `skip bootloader install` option.

base/setup/usetup/mui.c
```diff
@@ -303,6 +303,138 @@ MUIGetEntry(
...
+    IN SHORT Y,
+    IN SHORT Row)
+{
+    LPCSTR Buffer;
+    ULONG Length;
+
+    Buffer = MUIGetEntryTextFromCoord(Page, X, Y);
+    if (!Buffer)
+    {
+        DPRINT("MUIHighlightText(): Couldn't obtain a buffer from the string text!\n");
```
KyleKatarn31, line 389: Why this DPRINT since MUIGetEntryTextFromCoord already DPRINTs all failure conditions ?
Why don't we "return" here as using NULL buffer on next seps is likely not a good idea ? Or use a "default" value for Length ?

base/setup/usetup/mui.c
```diff
@@ -303,6 +303,138 @@ MUIGetEntry(
...
+    IN SHORT Y,
+    IN SHORT Row)
+{
+    LPCSTR Buffer;
+    ULONG Length;
+
+    Buffer = MUIGetEntryTextFromCoord(Page, X, Y);
+    if (!Buffer)
+    {
+        DPRINT("MUINormalText(): Couldn't obtain a buffer from the string text!\n");
```
KyleKatarn31, line 431: Why this DPRINT since MUIGetEntryTextFromCoord already DPRINTs all failure conditions ?
Why don't we "return" here as using NULL buffer on next seps is likely not a good idea ? Or use a "default" value for Length ?

base/setup/usetup/mui.c
Fraizeraust: It's my own way to add debug print functions to log debug information as much as possible, verbose-wise. Although I do agree that this function should bail out if no buffer could be captured.

KyleKatarn31: ok

KyleKatarn31:
OK, no more comments from me.
Any reviewer and integration for this one ?

GeoB99:
I'm going to cannibalize this PR, Hermes will take care with his VBR/MBR work.

PR #2760, opened 2020-05-09
[SNDREC32] Fix incorrect opaque text rendering

KyleKatarn31:
## Purpose

- SNDREC32 currently use an ugly text rendering with incorrect OPAQUE background to white, while DC is set to be transparent.

image

## Proposed changes

Remove "Opaque" parameter so that the transparency can be correctly used.

image


KyleKatarn31:
Thanks !

PR #2764, opened 2020-05-09
[CMAKE] Implement PCH builds with target_precompile_headers. CORE-16905

AmineKhaldi:
JIRA issue: [CORE-16905](https://jira.reactos.org/browse/CORE-16905)


dll/directx/wine/d3dcompiler_43/CMakeLists.txt
```diff
@@ -15,19 +15,21 @@ list(APPEND SOURCE
...
     hlsl.yy.c
     ppy.tab.c
-    ppl.yy.c
+    ppl.yy.c)
+
+add_library(d3dcompiler_43 MODULE
+    ${SOURCE}
+    ${PCH_SKIP_SOURCE}
     version.rc
+    ${CMAKE_CURRENT_BINARY_DIR}/d3dcompiler_43_stubs.c
```
ThFabba, line 32: In dhcpcsvc you have the stubs file in PCH_SKIP_SOURCE. I suppose technically that's better?

base/applications/winhlp32/CMakeLists.txt
```diff
@@ -11,16 +11,24 @@ list(APPEND SOURCE
...
 # macro.lex.yy.c has been generated with relative file paths...
 set_source_files_properties(macro.lex.yy.c PROPERTIES COMPILE_FLAGS "-UREACTOS_SOURCE_DIR -DREACTOS_SOURCE_DIR=\"\\\".\\\"\"")
 
 add_rc_deps(rsrc.rc ${CMAKE_CURRENT_SOURCE_DIR}/res/winhelp.ico)
-add_executable(winhlp32 ${SOURCE} macro.lex.yy.c string.c rsrc.rc)
+
+add_executable(winhlp32
+    ${SOURCE}
+    ${PCH_SKIP_SOURCE}
+    rsrc.rc)
```
ThFabba, line 28: CMake's PCH doesn't apply force-includes to .rc files, does it?

dll/win32/msgina/CMakeLists.txt
```diff
@@ -10,28 +10,27 @@ include_directories(
...
     msgina.rc
-    ${CMAKE_CURRENT_BINARY_DIR}/msgina_stubs.c
     ${CMAKE_CURRENT_BINARY_DIR}/msgina.def)
 
 set_module_type(msgina win32dll UNICODE)
 target_link_libraries(msgina wine uuid ${PSEH_LIB})
 add_delay_importlibs(msgina secur32)
 add_importlibs(msgina advapi32 user32 gdi32 powrprof userenv msvcrt kernel32 ntdll)
-add_pch(msgina msgina.h CPP_SOURCE)
+add_pch(msgina msgina.h "${PCH_SKIP_SOURCE}")
```
ThFabba, line 35: This is interesting -- the PCH was previously only applied to the .cpp file. I suppose that was a mistake?

dll/win32/msxml3/CMakeLists.txt
```diff
@@ -64,17 +68,15 @@ set_source_files_properties(version.rc PROPERTIES OBJECT_DEPENDS "${msxml3_rc_de
 
 add_library(msxml3 MODULE
     ${SOURCE}
-    xslpattern.tab.c
-    xslpattern.yy.c
-    factory.c
-    uuid.c
+    ${PCH_SKIP_SOURCE}
     version.rc
+    ${CMAKE_CURRENT_BINARY_DIR}/msxml3_stubs.c
```
ThFabba, line 73: Move to PCH_SKIP_SOURCE?

dll/win32/ole32/CMakeLists.txt
```diff
@@ -58,31 +58,33 @@ list(APPEND SOURCE
...
     ${CMAKE_CURRENT_BINARY_DIR}/ole32_stubs.c
-    ${CMAKE_CURRENT_BINARY_DIR}/proxy.dlldata.c
+    ${CMAKE_CURRENT_BINARY_DIR}/proxy.dlldata.c)
+
+add_library(ole32 MODULE
+    ${SOURCE}
+    ${PCH_SKIP_SOURCE}
+    ole32res.rc
+    ${CMAKE_CURRENT_BINARY_DIR}/dcom_i.c
+    ${CMAKE_CURRENT_BINARY_DIR}/irot_c.c
```
ThFabba, line 81: What does taking these out of SOURCE achieve? Previously that would mean they wouldn't get the PCH auto-included -- now being here just means the same as being in SOURCE, doesn't it?

dll/win32/oleacc/CMakeLists.txt
```diff
@@ -29,15 +32,14 @@ set_source_files_properties(oleacc.rc PROPERTIES OBJECT_DEPENDS "${oleacc_rc_dep
 
 add_library(oleacc MODULE
     ${SOURCE}
-    guid.c
+    ${PCH_SKIP_SOURCE}
     oleacc.rc
-    ${CMAKE_CURRENT_BINARY_DIR}/oleacc_classes_p.c
-    ${CMAKE_CURRENT_BINARY_DIR}/oleacc_stubs.c
+    ${CMAKE_CURRENT_BINARY_DIR}/proxy.dlldata.c
```
ThFabba, line 37: What makes the dlldata file different from the other stubs (for which we disable PCH)?

dll/win32/scrrun/CMakeLists.txt
```diff
@@ -21,13 +22,14 @@ set_source_files_properties(scrrun.rc PROPERTIES OBJECT_DEPENDS "${scrrun_rc_dep
 
 add_library(scrrun MODULE
     ${SOURCE}
-    guid.c
+    ${PCH_SKIP_SOURCE}
     scrrun.rc
+    ${CMAKE_CURRENT_BINARY_DIR}/scrrun_stubs.c
```
ThFabba, line 27: Move to PCH_SKIP_SOURCE?

dll/win32/setupapi/CMakeLists.txt
```diff
@@ -29,19 +29,21 @@ list(APPEND SOURCE
...
-    ${CMAKE_CURRENT_BINARY_DIR}/pnp_c.c)
+    rpc.c)
+
+list(APPEND PCH_SKIP_SOURCE
+    ${CMAKE_CURRENT_BINARY_DIR}/setupapi_stubs.c)
 
 add_library(setupapi MODULE
     ${SOURCE}
+    ${PCH_SKIP_SOURCE}
+    ${CMAKE_CURRENT_BINARY_DIR}/pnp_c.c
```
ThFabba, line 40: Move to PCH_SKIP_SOURCE?

dll/win32/sti/CMakeLists.txt
```diff
@@ -11,19 +11,21 @@ add_rpcproxy_files(sti_wia.idl)
...
+list(APPEND PCH_SKIP_SOURCE
+    guid.c)
 
 add_library(sti MODULE
     ${SOURCE}
-    guid.c
+    ${PCH_SKIP_SOURCE}
     sti.rc
+    ${CMAKE_CURRENT_BINARY_DIR}/proxy.dlldata.c
+    ${CMAKE_CURRENT_BINARY_DIR}/sti_wia_p.c
```
ThFabba, line 24: Move to PCH_SKIP_SOURCE?

CMakeLists.txt
```diff
@@ -139,7 +139,7 @@ else()
         add_definitions(-D_WINKD_)
     endif()
 
-    if(CMAKE_VERSION MATCHES "ReactOS" AND CMAKE_VERSION VERSION_LESS 3.3.0)
+    if(CMAKE_VERSION MATCHES "ReactOS")
```
ThFabba, line 142: Make this `AND NOT CMAKE_VERSION VERSION_LESS 3.16.0`?
That way we don't have to 100% time the merge of this with the RosBE switch.

dll/directx/wine/d3d9/CMakeLists.txt
```diff
@@ -18,18 +18,20 @@ list(APPEND SOURCE
...
+
+list(APPEND PCH_SKIP_SOURCE
+    guid.c)
 
 add_library(d3d9 MODULE
     ${SOURCE}
-    guid.c
+    ${PCH_SKIP_SOURCE}
     version.rc
     ${CMAKE_CURRENT_BINARY_DIR}/d3d9_stubs.c
```
ThFabba, line 30: Move to PCH_SKIP_SOURCE?

dll/directx/wine/d3dcompiler_43/CMakeLists.txt
AmineKhaldi: Basically the idea back then was that we'd make use of windef.h (from stubs.h) being part of the PCH almost always, except when we get compile issues from including the stub files. What do you think?

ThFabba: Ah. Yep that makes sense!

base/applications/winhlp32/CMakeLists.txt
AmineKhaldi: Yeap, I checked before letting them in :)

dll/win32/msgina/CMakeLists.txt
AmineKhaldi: Yeah, I had the exact same reaction! I deemed it an oversight because we had one c++ file and many c files so it makes sense to leverage PCH for the c ones :)

dll/win32/ole32/CMakeLists.txt
AmineKhaldi: With the CMake built-in PCH functionality only exposing the ability to skip source files, I leveraged the existing framework to account for that, and seamlessly continue working in all the modules we're not skipping in. SOURCE was always meant to indicate source level files, as we put the autogenerated ones in the add_x() calls to make them visible.
Finally, SOURCE in the add_pch() calls that do not skip anything, are just that, the string "SOURCE" and so we're like "sure, I'll skip this source file named SOURCE from PCH when I see it" :)

CMakeLists.txt
AmineKhaldi: Done.

dll/win32/msgina/CMakeLists.txt
learn-more: iirc, sharing PCH between c and c++ caused build problems.

CMakeLists.txt
Extravert-ir: Are we keeping `cmake_minimum_required` where it is?

sdk/cmake/CMakeMacros.cmake
```diff
@@ -974,3 +974,13 @@ function(add_rostests_file)
         endif()
     endif()
 endfunction()
+
+if(PCH AND (NOT ENABLE_CCACHE) AND (NOT CMAKE_HOST_APPLE))
```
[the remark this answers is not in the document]
AmineKhaldi: Hmm it was done back then this way to allow anyone to experiment with both co-existing. I wouldn't see any reason to prevent that. What do you think?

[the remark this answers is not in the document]
Extravert-ir: Does it make sense to use precompiled headers with ccache? From my understanding, it shouldn't increase performance in this case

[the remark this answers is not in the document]
AmineKhaldi: For me it doesn't make sense to mix them at all, but what I meant above is that I wouldn't restrict anyone from the possibility just because of that. It's not harming us in any way to leave that available.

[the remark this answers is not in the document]
Extravert-ir: My point is simple: it's easier when pch build is enabled with only one cmake swith - `PCH`. If I want to turn it on or off, I just need to pass this switch manually and that's it. In this case, I need to also do something with `ENABLE_CCACHE` and `CMAKE_HOST_APPLE`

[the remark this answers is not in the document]
Extravert-ir: My suggestion is to make
```
if(ENABLE_CCACHE OR CMAKE_HOST_APPLE)
set(PCH 0)
endif()
```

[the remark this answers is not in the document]
AmineKhaldi: I see what you mean. I moved the ccache check up around the pch check.

.github/workflows/build.yml
```diff
@@ -11,13 +11,27 @@ jobs:
           wget https://svn.reactos.org/storage/vperevertkin/rosbe-ci.tar.zst
           mkdir ${{github.workspace}}/rosbe
           tar -I zstd -xvf rosbe-ci.tar.zst --directory ${{github.workspace}}/rosbe
+      - name: Install other packages
+        run: sudo apt install ccache
```
ThFabba, line 15: Do the build containers get reused?
I think the main benefit of ccache comes from building ROS a second time when a previous build has been cached -- so if we get a fresh container each build, I would imagine it to not be useful.

Extravert-ir:
Here are the results: on my laptop, I've got 9m 15s vs 13m 51s build time (GCC)
But there is one consequence: GCC build folder now consumes almost 17gb of space (instead of 1.5). RosBE 2.1 PCH build occupied only half of that (about 8gb)

@AmineKhaldi made some research and looks like it's just GCC8 generates more code. We will probably fill a bug report.

For MSVC everything is fine, build folder size is 7gb.

I think maybe we should turn the PCH build off by default on GCC for now. What do you think?

PR #2766, opened 2020-05-10
[USER32] Adjusted the column width in IDC_PICKICON_LIST

zecarlos1957:
Adjusted the width of the columns in  IDC_PICKICON_LIST list

The list of icons in IDC_PICKICON_LIST only shows half an icon in the right hand column on all lines.

Proposal:
Adjust the width of each column to show a full number of icons.


zecarlos1957:
Before
After

katahiromz:
The PR title has two typos.
I doubt whether this PR is a generic solution...

zecarlos1957:
let me think about. Thanks.

Matriks404:
But Windows also shows 7 full icons and about a 1/3 of the next icon, is there any reason we don't want to behave like Windows here?

PR #2767, opened 2020-05-11
[FTP] Don't hardcode C: drive

katahiromz:
## Purpose

### DIFF
--- a/base/applications/network/ftp/main.c
+++ b/base/applications/network/ftp/main.c
@@ -227,10 +227,16 @@ int main(int argc, const char *argv[])
 		(void) strcpy(home, pw->pw_dir);
 	}
 #endif
-	if (GetEnvironmentVariableA("SystemDrive", home, ARRAYSIZE(home)))
+	cp = getenv("SystemDrive");
+	if (cp != NULL && *cp != 0)
+	{
+		strcpy(home, cp);
 		strcat(home, "/");
+	}
 	else
+	{
 		strcpy(home, "C:/");
+	}
 	if (argc > 0) {
 		if (setjmp(toplevel))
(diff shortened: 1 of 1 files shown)


base/applications/network/ftp/main.c
```diff
@@ -227,7 +227,10 @@ int main(int argc, const char *argv[])
 		(void) strcpy(home, pw->pw_dir);
 	}
 #endif
-        strcpy(home, "C:/");
+	if (GetEnvironmentVariableA("SystemDrive", home, ARRAYSIZE(home)))
+		strcat(home, "/");
+	else
+		strcpy(home, "C:/");
```
KyleKatarn31, line 233: Maybe worth logging a DPRINT1 with GetLastError here as GetEnvironmentVariableA not supposed to fail for SystemDrive

HBelusca: We should avoid using macros that resolve to ntdll exports in these sort of utilities.

HBelusca, line 233: Since this whole application is coded in POSIX way, please use instead `getenv()`:
```suggestion
        cp = getenv("SystemDrive");
        if ((cp != NULL) && *cp) {
                strcpy(home, cp);
                strcat(home, "/");
        } else {
                strcpy(home, "C:/");
        }
```
(I also fixed the indentation and kept the ambient C code style so as to match with the rest of the code.)

Extravert-ir: The file uses tabs and you used spaces ;)

HBelusca: oh yes indeed ^^ (actually a mixture)

PR #2768, opened 2020-05-11
[SHELL32] Don't hardcode C: drive

katahiromz:
## Purpose

### DIFF
--- a/dll/win32/shell32/folders/CRecycleBin.cpp
+++ b/dll/win32/shell32/folders/CRecycleBin.cpp
@@ -191,13 +191,17 @@ CRecycleBinEnum::~CRecycleBinEnum()
 
 HRESULT WINAPI CRecycleBinEnum::Initialize(DWORD dwFlags)
 {
-    static LPCWSTR szDrive = L"C:\\";
+    WCHAR szDrive[8];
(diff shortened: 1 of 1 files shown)


dll/win32/shell32/folders/CRecycleBin.cpp
```diff
@@ -191,13 +191,17 @@ CRecycleBinEnum::~CRecycleBinEnum()
 
 HRESULT WINAPI CRecycleBinEnum::Initialize(DWORD dwFlags)
 {
-    static LPCWSTR szDrive = L"C:\\";
+    WCHAR szDrive[8];
+    if (GetEnvironmentVariableW(L"SystemDrive", szDrive, _countof(szDrive)))
```
learn-more, line 195: This probably should take into account what drives the parent recycle bin is activated for?
But we probably don't support that at all...

katahiromz: Registry setting?

learn-more: No, probably needs to be passed from wherever this was invoked from,
but I am not sure we handle recycle bin on a non-system drive at all...

PR #2777, opened 2020-05-13
[SHELL32] Don't hardcode C: drive again

katahiromz:
## Purpose

- Avoid buffer overrun.
- Don't hardcode C: drive.

### DIFF
--- a/dll/win32/shell32/folders/CRecycleBin.cpp
+++ b/dll/win32/shell32/folders/CRecycleBin.cpp
@@ -192,7 +192,7 @@ CRecycleBinEnum::~CRecycleBinEnum()
 HRESULT WINAPI CRecycleBinEnum::Initialize(DWORD dwFlags)
 {
     WCHAR szDrive[8];
-    if (GetEnvironmentVariableW(L"SystemDrive", szDrive, _countof(szDrive)))
+    if (GetEnvironmentVariableW(L"SystemDrive", szDrive, _countof(szDrive) - 1))
         PathAddBackslashW(szDrive);
     else
         StringCbCopyW(szDrive, sizeof(szDrive), L"C:\\");
@@ -367,7 +367,7 @@ HRESULT WINAPI CRecycleBinItemContextMenu::InvokeCommand(LPCMINVOKECOMMANDINFO l
         Context.pFileDetails = _ILGetRecycleStruct(apidl);
         Context.bFound = FALSE;
 
-        if (GetEnvironmentVariableW(L"SystemDrive", szDrive, _countof(szDrive)))
+        if (GetEnvironmentVariableW(L"SystemDrive", szDrive, _countof(szDrive) - 1))
             PathAddBackslashW(szDrive);
         else
             StringCbCopyW(szDrive, sizeof(szDrive), L"C:\\");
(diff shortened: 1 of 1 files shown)


dll/win32/shell32/folders/CRecycleBin.cpp
```diff
@@ -823,14 +823,18 @@ HRESULT WINAPI CRecycleBin::InvokeCommand(LPCMINVOKECOMMANDINFO lpcmi)
...
 
     if (LOWORD(lpcmi->lpVerb) == iIdEmpty)
     {
-        // FIXME
-        // path & flags
-        hr = SHEmptyRecycleBinW(lpcmi->hwnd, L"C:\\", 0);
+        if (GetEnvironmentVariableW(L"SystemDrive", szDrive, _countof(szDrive) - 1))
+            PathAddBackslashW(szDrive);
+        else
+            StringCbCopyW(szDrive, sizeof(szDrive), L"C:\\");
```
Extravert-ir, line 835: Shouldn't we just fail if there is no such environment variable?
This fallback looks like a hack

HBelusca: Also there is the possibility that the recycle bin is on a mounted volume that doesn't have any drive letter, and theoretically they should still be displayed in the recycle bin thing.

dll/win32/shell32/folders/CRecycleBin.cpp
```diff
@@ -192,7 +192,7 @@ CRecycleBinEnum::~CRecycleBinEnum()
 HRESULT WINAPI CRecycleBinEnum::Initialize(DWORD dwFlags)
 {
     WCHAR szDrive[8];
-    if (GetEnvironmentVariableW(L"SystemDrive", szDrive, _countof(szDrive)))
+    if (GetEnvironmentVariableW(L"SystemDrive", szDrive, _countof(szDrive) - 1))
```
SergeGautherie, line 195: More explicitly, like:
```suggestion
    if (GetEnvironmentVariableW(L"SystemDrive", szDrive, _countof(szDrive) - _countof(L"\")))
```

Moreover, should we foolproof the result check: `0`, `< size`, `> size`?

HBelusca: _countof(L"\") won't compile (also if it's for what you thought, it's equal to 2.)

dll/win32/shell32/folders/CRecycleBin.cpp
katahiromz: ???

dll/win32/shell32/folders/CRecycleBin.cpp
```diff
@@ -192,10 +192,10 @@ CRecycleBinEnum::~CRecycleBinEnum()
 HRESULT WINAPI CRecycleBinEnum::Initialize(DWORD dwFlags)
 {
     WCHAR szDrive[8];
-    if (GetEnvironmentVariableW(L"SystemDrive", szDrive, _countof(szDrive)))
-        PathAddBackslashW(szDrive);
-    else
-        StringCbCopyW(szDrive, sizeof(szDrive), L"C:\\");
+    if (!GetEnvironmentVariableW(L"SystemDrive", szDrive, _countof(szDrive) - 1))
+        return E_FAIL;
```
learn-more, line 196: A DPRINT or FAILED_UNEXPECTEDLY or something like it would be nice, but other than that I agree that failing is the correct thing to do here.

katahiromz: `ERR`'s are inserted.

PR #2779, opened 2020-05-13
[SERVICES][DCOMLAUNCH] Don't hardcode C: drive

katahiromz:
## Purpose

### DIFF
--- a/base/services/dcomlaunch/network.c
+++ b/base/services/dcomlaunch/network.c
@@ -45,7 +45,7 @@ CookupNodeId(UCHAR * NodeId)
     MEMORYSTATUS MemoryStatus;
     LUID Luid;
     DWORD SectorsPerCluster, BytesPerSector, NumberOfFreeClusters, TotalNumberOfClusters;
-    CHAR szDrive[8];
+    WCHAR szDrive[8];
 
     /* Initialize node id */
     memset(NodeId, 0x11, SEED_BUFFER_SIZE * sizeof(UCHAR));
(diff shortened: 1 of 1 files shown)


base/services/dcomlaunch/network.c
```diff
@@ -104,8 +105,14 @@ CookupNodeId(UCHAR * NodeId)
         *NodeMiddle = *NodeMiddle ^ Luid.HighPart;
     }
 
+    /* Get system drive */
+    if (GetEnvironmentVariableW(L"SystemDrive", szDrive, ARRAYSIZE(szDrive) - 1))
```
HBelusca, line 109: For services, it is more robust and secure to obtain the systemdrive letter not via the environment prompt, but instead by calling the [GetSystemDirectory](https://docs.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getsystemdirectoryw) function and extracting the drive letter out of it.

base/services/dcomlaunch/network.c
```diff
@@ -104,8 +105,14 @@ CookupNodeId(UCHAR * NodeId)
...
+        lstrcatW(szDrive, L"\\");
+    else
+        lstrcpyW(szDrive, L"C:\\");
+
     /* And finally with free disk space */
-    if (GetDiskFreeSpaceA("c:\\", &SectorsPerCluster, &BytesPerSector, &NumberOfFreeClusters, &TotalNumberOfClusters))
+    if (GetDiskFreeSpaceW(szDrive, &SectorsPerCluster, &BytesPerSector, &NumberOfFreeClusters, &TotalNumberOfClusters))
```
HBelusca, line 115: Actually we can do better here, and use [GetDiskFreeSpaceExW](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getdiskfreespaceexw) that removes the need to provide a root directory. (Suggested by @Extravert-ir )

PR #2783, opened 2020-05-13
[SHLWAPI] Implement SHIsLowMemoryMachine

katahiromz:
## Purpose

Implement `shlwapi!SHIsLowMemoryMachine` function.

JIRA issue: N/A

### DIFF
--- a/dll/win32/shlwapi/ordinal.c
+++ b/dll/win32/shlwapi/ordinal.c
@@ -4324,16 +4324,28 @@ HRESULT WINAPI SHGetInverseCMAP(LPDWORD dest, DWORD dwSize)
  * Determine if the current computer has low memory.
  *
  * PARAMS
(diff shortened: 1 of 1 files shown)


ThFabba:
What does this fix? Introducing ROS-Diffs in Wine code without application benefits is counterproductive.

KyleKatarn31:
https://docs.microsoft.com/en-us/windows/win32/api/shlwapi/nf-shlwapi-shislowmemorymachine
"Note  Always returns FALSE under Windows XP."

Therefore, why would we need the 16 MB threshold for ROS ? (even for compatibility for low end apps eventuelly calling this API)

HBelusca:
> "Note Always returns FALSE under Windows XP"

But not on Win2k3 (which is our target).

KyleKatarn31:
OK !

PR #2785, opened 2020-05-13
[SHELL32] Implement SheShortenPathA

katahiromz:
## Purpose

Implement `shell32!SheShortenPathA` function.

JIRA issue: N/A


SergeGautherie:
Should this function be moved out of stubs.cpp, then?

PR #2791, opened 2020-05-14
[MSGINA] PL-pl FIX

pithwz:
msgina.dll PL-pl
-FIXED: dialogs, buttons,
msgina_before_after

### DIFF
--- a/dll/win32/msgina/lang/pl-PL.rc
+++ b/dll/win32/msgina/lang/pl-PL.rc
@@ -5,7 +5,7 @@
  *    UTF-8 conversion by Caemyr (May, 2011)
  *    Updated by Saibamen - Adam Stachowicz (saibamenppl@gmail.com) (April, 2014)
  *    Updated by wojo664 - Wojtek Kozłowski (wojo664@gmail.com, wojtekkozlo664@op.pl) (April, 2014)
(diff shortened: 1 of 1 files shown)


learn-more:
Fixed how?

HBelusca:
> Fixed how?

The size of the UI elements.

learn-more:
> > Fixed how?
> 
> The size of the UI elements.

Obviously, but the last commit to this file was also from him, and also changed sizes.
So I was wondering what changed all of a sudden?

binarymaster:
Yeah I'd like to see "before" and "after" too.

Other PRs are changing element sizes as well, if these are just cosmetic changes, they should be applied (synced) to all translations.

HBelusca:
Sometimes UI object sizes need to be changed per language (if strings are longer than e.g. in english...)

gonzoMD:
According to the images, I see no much difference. No cutoff strings etc. Why are these changes needed?

binarymaster (changes requested):
I see only cosmetic changes here like moving icons and labels around. If you want this change merged, you need to perform the same tweaking for all languages/translations, otherwise this is a no go.

pithwz:
end

PR #2792, opened 2020-05-15
[RAPPS] Move database source URL to settings dialog

kernelbin:
## Purpose

move Rapps database URL to settings dialog, enable users to edit the source URL
and display a different 「Updating Database… (Unofficial source)」 when updating db from an unofficial source

now the dialog looks like this
image

## TODO

- check if the URL is valid when clicking OK in the dialog


base/applications/rapps/available.cpp
```diff
@@ -301,8 +301,42 @@ BOOL CAvailableApps::UpdateAppsDB()
...
+        0, &AppDataBaseURLLength, 0);
+        
+        ApplicationDataBaseURL = (WCHAR *)HeapAlloc(GetProcessHeap(), 0, AppDataBaseURLLength);
+        
+        if(!ApplicationDataBaseURL) return FALSE;
+        
+        UrlCombineW(APPLICATION_DATABASE_BASEURL,
+        APPLICATION_DATABASE_FILENAME,
+        ApplicationDataBaseURL, &AppDataBaseURLLength, 0);
+    }
```
learn-more, line 334: I think it would be better if people have to specify the full url.
That way they can also specify a script or whatever type of web application that dynamically generates a file.

base/applications/rapps/loaddlg.cpp
```diff
@@ -76,13 +84,15 @@ struct DownloadInfo
...
-        : szUrl(AppInfo.m_szUrlDownload)
+        : DownloadType(DLTYPE_APPLICATION)
+        , szUrl(AppInfo.m_szUrlDownload)
         , szName(AppInfo.m_szName)
         , szSHA1(AppInfo.m_szSHA1)
         , SizeInBytes(AppInfo.m_SizeBytes)
     {
     }
 
+    int DownloadType;
```
learn-more, line 95:
```suggestion
    DownloadType DownloadType;
```

base/applications/rapps/available.cpp
kernelbin: I was thinking about probably some other files (for example, a file standalone to record dependencies between software, or an update for rapps itself) probably will be added in the future.
You are right, but I think people can still specify a web application to dynamically generate for ```rappmgr.cab``` if they want, just a little bit of trouble.

that's just my opinion, feel free to tell me if you still think full URL is better.😉

base/applications/rapps/loaddlg.cpp
kernelbin: ah, thanks! I will correct it.

base/applications/rapps/available.cpp
learn-more: Yes, for now there is one 'cab' file to be downloaded.
If we need a seperate file for dependencies (which I don't think), it can also be stored in this cab file.

kernelbin:
not ready to merge yet, some bugs are still investigating

base/applications/rapps/lang/en-US.rc
```diff
@@ -100,6 +104,16 @@ BEGIN
...
+IDD_ABOUT_DIALOG DIALOGEX 22, 16, 190, 66
+STYLE DS_SHELLFONT | WS_BORDER | WS_DLGFRAME | WS_SYSMENU | DS_MODALFRAME
+CAPTION "About"
+FONT 8, "MS Shell Dlg"
+BEGIN
+    ICON IDI_MAIN, IDC_STATIC, 10, 10, 7, 30
+    LTEXT "ReactOS Applications Manager\nCopyright (C) 2009\nby Dmitry Chapyshev (dmitry@reactos.org)", IDC_STATIC, 48, 7, 130, 39
+    DEFPUSHBUTTON "Close", IDOK, 133, 46, 50, 14
+END
+
```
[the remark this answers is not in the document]
kernelbin: I don't know..... ? probably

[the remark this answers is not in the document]
learn-more: Yeah, that got recently removed, and it looks like you re-added it.
https://github.com/reactos/reactos/commit/a6084d7440c260eb0ecb352f3b20f8c3a89162a2#diff-7f26a0bce2f6b7dc7e30785e464d5cde

base/applications/rapps/settingsdlg.cpp
```diff
@@ -50,6 +50,31 @@ BOOL ChooseFolder(HWND hwnd)
     return bRet;
 }
 
+BOOL IsUrlValid(const WCHAR * Url)
+{
+    URL_COMPONENTSW UrlComponmentInfo = { 0 };
+    UrlComponmentInfo.dwStructSize = sizeof(UrlComponmentInfo);
+    UrlComponmentInfo.dwSchemeLength = 1;
```
learn-more, line 57: This is a vista feature according to the documentation, does it work on 2k3?

learn-more: Seems to work fine in 2k3.

base/applications/rapps/include/winmain.h
```diff
@@ -15,6 +16,9 @@ struct SETTINGS_INFO
     INT Top;
     INT Width;
     INT Height;
+    /* Software source settings */
+    BOOL bUseSource;
+    WCHAR szSourceURL[INTERNET_MAX_URL_LENGTH];
```
learn-more, line 21: Adding data in the middle of the settings does break all settings below it when loading it from the registry,
And the bUseSource will probably be 0 (depending on what `Proxy` was before.
However I doubt anyone will load rapps on a machine that previously ran the old version of rapps...

kernelbin: Ah, I thought ROS and RAPPS are in alpha state, do not need to think about updating ...
I just tried to make the order here correspond with the order in the dialog

learn-more: It is used by almost everyone using ReactOS, however it is almost always in a fresh install.
Either way it would make sense to store the settings in a human readable format (one value per option) some day, which also make it easier to introduce new options without having to worry about random things changing.

kernelbin: almost everyone always uses rapps in a fresh install, that's why I think we can ignore this problem (at least now) 😂
one value per option does sound good. I can improve it later (probably another PR, not this one)

PR #2793, opened 2020-05-15
[NTDLL_APITEST] Add NtSetInformationThread testcase

Fraizeraust:
It seems like both XP and Server 2003 do privilege probe checks when setting a base priority to a thread whereas ReactOS doesn't. I've to fix that later.
**Windows XP**
Capture
**Windows Server 2003**
Capture4
**ReactOS**
Capture2

### DIFF
--- a/modules/rostests/apitests/ntdll/NtSetInformationThread.c
+++ b/modules/rostests/apitests/ntdll/NtSetInformationThread.c
@@ -1,0 +1,100 @@
+/*
+ * PROJECT:         ReactOS API tests
+ * LICENSE:         GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:         Tests for the NtSetInformationThread API
+ * COPYRIGHT:       Copyright 2020 George Bișoc <george.bisoc@reactos.org>
+ */
+
+#include "precomp.h"
+
+static
+void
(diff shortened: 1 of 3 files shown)


HBelusca:
Yep this is the same thing as what I mentioned for NtSetInformationProcess last time.

modules/rostests/apitests/ntdll/NtSetInformationThread.c
```diff
@@ -0,0 +1,100 @@
...
+#include "precomp.h"
+
+static
+void
+Test_ThreadPriorityClass(void)
+{
+    NTSTATUS Status;
+    KPRIORITY *Priority;
+
+    Priority = HeapAlloc(GetProcessHeap(), HEAP_ZERO_MEMORY, sizeof(KPRIORITY));
```
HBelusca, line 17: So, why using a heap allocation instead of stack?

Fraizeraust: Sometimes I just like heap allocation pointers.  😄 
But sure, I'll allocate the variable directly to the stack later.

learn-more: For this test it makes sense, since that way the stack is not corrupted, and the heap can include guard pages around it.

PR #2803, opened 2020-05-17
[SHLWAPI] Some functions are NT5.2 only. CORE-9708

SergeGautherie:
http://www.geoffchappell.com/studies/windows/shell/shlwapi/history/ords600.htm
I checked 552-566 range only.

JIRA issue: [CORE-9708](https://jira.reactos.org/browse/CORE-9708)

### DIFF
--- a/dll/win32/shlwapi/shlwapi.spec
+++ b/dll/win32/shlwapi/shlwapi.spec
@@ -559,11 +559,11 @@
 559 stub -noname SHGetSignatureInfo
 560 stdcall -stub -noname SHWindowsPolicyGetValue(ptr ptr ptr)
 561 stub -noname AssocGetUrlAction
-562 stub -noname SHGetPrivateProfileInt
-563 stdcall -stub -noname SHGetPrivateProfileSection(wstr ptr long ptr)
-564 stub -noname SHGetPrivateProfileSectionNames
-565 stub -noname SHGetPrivateProfileString
-566 stub -noname SHGetPrivateProfileStruct
+562 stub -noname -version=0x502 SHGetPrivateProfileInt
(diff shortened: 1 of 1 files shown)


dll/win32/shlwapi/shlwapi.spec
```diff
@@ -551,19 +551,19 @@
 551 stub -noname IShellFolder_CompareIDs
 552 stdcall -stub -noname -version=0x501-0x502 SHEvaluateSystemCommandTemplate(wstr ptr ptr ptr)
 553 stdcall IsInternetESCEnabled()
-554 stdcall -noname -stub SHGetAllAccessSA()
+554 stdcall -stub -noname -version=0x502 SHGetAllAccessSA()
```
ThFabba, line 554: Not sure where you're getting your information from but my Win2003 version of shlwapi has `_SHGetAllAccessSA@0` as ordinal 554 according to symbols. I will assume the same is true for all the other functions unless you have proof otherwise.

ThFabba: Oh. I just realized you're marking them as 5.2.
Then this seems merely pointless, not wrong.

SergeGautherie: > pointless

There are already some cases:
https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=version%3D0x50

Would that not help with "AppCompat" someday? At least as documentation. Especially such functions/ordinals which differ between Windows versions.

ThFabba:
> Would that not help with "AppCompat" someday?

ReactOS has 10000 _actual_ bugs. What "might" help "some day" is a waste of time now. Please be mindful of the time it takes to review and merge changes.

PR #2806, opened 2020-05-18
[ROSAUTOTEST] Add /t parameter for repeating tests

Extravert-ir:

### DIFF
--- a/modules/rostests/rosautotest/CConfiguration.cpp
+++ b/modules/rostests/rosautotest/CConfiguration.cpp
@@ -19,6 +19,7 @@ CConfiguration::CConfiguration()
     : m_CrashRecovery(false),
       m_IsInteractive(false),
       m_PrintToConsole(true),
+      m_RepeatCount(1),
       m_Shutdown(false),
       m_Submit(false)
 {
@@ -52,10 +53,17 @@ CConfiguration::ParseParameters(int argc, wchar_t* argv[])
     {
         if(argv[i][0] == '-' || argv[i][0] == '/')
         {
+            unsigned long tmp_RepeatCount;
+
             switch(argv[i][1])
             {
                 case 'c':
                     ++i;
+                    if (i >= argc)
+                    {
+                        throw CInvalidParameterException();
+                    }
+
                     m_Comment = UnicodeToAscii(argv[i]);
                     break;
 
@@ -73,6 +81,23 @@ CConfiguration::ParseParameters(int argc, wchar_t* argv[])
 
                 case 'w':
                     m_Submit = true;
+                    break;
+
+                case 't':
+                    ++i;
+                    if (i >= argc)
+                    {
+                        throw CInvalidParameterException();
+                    }
+
+                    tmp_RepeatCount = wcstoul(argv[i], NULL, 10);
+
+                    if (tmp_RepeatCount == 0 || tmp_RepeatCount > 10000)
+                    {
+                        throw CInvalidParameterException();
+                    }
+
+                    m_RepeatCount = tmp_RepeatCount;
                     break;
 
                 default:
--- a/modules/rostests/rosautotest/CConfiguration.h
+++ b/modules/rostests/rosautotest/CConfiguration.h
@@ -12,6 +12,7 @@ private:
     bool m_IsInteractive;
     bool m_IsReactOS;
     bool m_PrintToConsole;
+    unsigned long m_RepeatCount;
     bool m_Shutdown;
     bool m_Submit;
     string m_Comment;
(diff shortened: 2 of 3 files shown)


modules/rostests/rosautotest/CConfiguration.cpp
```diff
@@ -75,6 +78,12 @@ CConfiguration::ParseParameters(int argc, wchar_t* argv[])
                     m_Submit = true;
                     break;
 
+                case 't':
+                    ++i;
+                    tmp_RepeatCount = _wtoi(argv[i]);
+                    m_RepeatCount = tmp_RepeatCount == 0 ? 1 : tmp_RepeatCount;
```
KyleKatarn31, line 84: "== 0" to be replaced by "< 1" in case negative number is used (robustness case)

learn-more: please explain how an unsigned number can be negative.

KyleKatarn31: Sorry, I looked at _wtoi rather than tmp_RepeatCount. Disregard.

modules/rostests/rosautotest/CConfiguration.cpp
```diff
@@ -75,6 +78,12 @@ CConfiguration::ParseParameters(int argc, wchar_t* argv[])
                     m_Submit = true;
                     break;
 
+                case 't':
+                    ++i;
+                    tmp_RepeatCount = _wtoi(argv[i]);
```
ColinFinck, line 83: Crashes here when running "rosautotest /t" (without any number)

modules/rostests/rosautotest/CConfiguration.h
```diff
@@ -35,6 +36,7 @@ class CConfiguration
     bool IsReactOS() const { return m_IsReactOS; }
     const string& GetComment() const { return m_Comment; }
     const wstring& GetModule() const { return m_Module; }
+    const unsigned char& GetRepeatCount() const { return m_RepeatCount; }
```
ColinFinck, line 39: Primitive types can be copied at no cost. It's best practice to just return the copied value instead of a `const` reference for getting the count.

The return value will also be moved to the 32-bit EAX register, so using an `unsigned char` here doesn't really save anything. An `int` or `long` type is more common here.
It's probably the best to store `m_RepeatCount` as an `unsigned long` and use `wcstoul` above instead of `_wtoi` (which overflows an `unsigned char` for values > 255).

modules/rostests/rosautotest/CConfiguration.cpp
KyleKatarn31: Yes, because of "++i" then argv[i], with an incorrect assumption on the array size.

modules/rostests/rosautotest/CConfiguration.cpp
KyleKatarn31: Any need to protect again a max value or do we allow full range of var type ?

modules/rostests/rosautotest/main.cpp
```diff
@@ -77,8 +77,27 @@ wmain(int argc, wchar_t* argv[])
...
+        {
+            CWineTest WineTest;
+
+            if (Configuration.GetRepeatCount() > 1)
+            {
+                stringstream ss;
+                ss << "[ROSAUTOTEST] Running attempt #" << i+1 << endl;
+                StringOut(ss.str());
+            }
+            WineTest.Run();
```
HBelusca, line 99: Strange that this function is used like that. I would have expected it's instead some kind of static function of the CWineTest class ^^ (to be used like `CWineTest::Run()`).

learn-more: then how is it supposed to access member variables?

learn-more: besides, Run is virtual...

modules/rostests/rosautotest/main.cpp
```diff
@@ -77,8 +77,27 @@ wmain(int argc, wchar_t* argv[])
...
+        {
+            stringstream ss1;
+
+            ss1 << "[ROSAUTOTEST] The test will be repeated " << Configuration.GetRepeatCount() << " times" << endl;
+            StringOut(ss1.str());
+        }
+
         /* Run the tests */
-        WineTest.Run();
+        for (int i = 0; i < Configuration.GetRepeatCount(); i++)
```
ThFabba, line 89: Should make `i` match the return type of `GetRepeatCount` in signedness

modules/rostests/rosautotest/CConfiguration.cpp
ThFabba, line 83: Not sure I see the point in using the `tmp_` variable. Since `m_RepeatCount` is going to be overwritten unconditionally in this case and we're not examining the previous value, might as well write there directly?

ThFabba: The same is probably true of the comment case above then?

Extravert-ir: The idea here was to truncate input. I don't want the whole `int` range

ThFabba: I'd argue it's better to fail than to silently truncate. But if you feel strongly, the temp var is fine.

PR #2807, opened 2020-05-18
[DEVMGR] Use the ShellAboutW API routine to display the about dialog and French translation update

KyleKatarn31:
## Purpose

image

Similar to #2743 by @HBelusca

## Proposed changes

image

### DIFF
--- a/dll/win32/devmgr/devmgmt/MainWindow.cpp
+++ b/dll/win32/devmgr/devmgmt/MainWindow.cpp
@@ -654,11 +654,15 @@ CDeviceManager::OnCommand(_In_ WPARAM wParam,
             CAtlStringW szAppName;
             CAtlStringW szAppAuthors;
             HICON hIcon;
-            szAppName.LoadStringW(g_hThisInstance, IDS_APPNAME);
-            szAppAuthors.LoadStringW(g_hThisInstance, IDS_APP_AUTHORS);
+
+            if (!szAppName.LoadStringW(g_hThisInstance, IDS_APPNAME))
+                szAppName = L"ReactOS Device Manager";
+            if (!szAppAuthors.LoadStringW(g_hThisInstance, IDS_APP_AUTHORS))
+                szAppAuthors = L"";
             hIcon = LoadIconW(g_hThisInstance, MAKEINTRESOURCEW(IDI_MAIN_ICON));
             ShellAboutW(m_hMainWnd, szAppName, szAppAuthors, hIcon);
-            DestroyIcon(hIcon);
+            if (hIcon)
+                DestroyIcon(hIcon);
 
             // Set focus back to the treeview
             m_DeviceView->SetFocus();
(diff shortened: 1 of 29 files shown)


dll/win32/devmgr/devmgmt/MainWindow.cpp
```diff
@@ -652,11 +651,14 @@ CDeviceManager::OnCommand(_In_ WPARAM wParam,
 
         case IDM_ABOUT:
         {
-            // Apportion blame
-            MessageBoxW(m_hMainWnd,
-                        L"ReactOS Device Manager\r\nCopyright Ged Murphy 2015",
-                        L"About",
-                        MB_OK | MB_APPLMODAL);
+            CAtlStringW szAppName;
```
gedmurphy, line 654: No error checking?
Looks like `szAppAuthors` can be null, and `DestroyIcon` can handle a null ptr. Didn't look into `szAppName`

KyleKatarn31: szAppName is the only pointer which can not be NULL for ShellAboutW.
However, I understand that CAtlStringW is "safe" with regard to this case. Am I wrong ?

gedmurphy: I wouldn't say `DestroyIcon` can be null, it just so happens that if you drill into win32k it has some protection there. There's nothing in the docs to suggest you can pass null.
Regarding `CAtlStringW` being safe, it's unrelated. An empty string will be null, so you're passing null into the API. 

You should really be null checking `szAppName` and `hIcon` for null to avoid crashing the app.

KyleKatarn31: OK for checking hIcon for safe `DestroyIcon`.
For szAppName, I understand that `CAtlStringW `would produce a null string (L"\0"), but not a null pointer. But i'll add robustness case for that as well.

gedmurphy: Oh it does? It's been so long since I worked with atl...
The fix looks much better now anyway 👍

KyleKatarn31: Anyway, thank you very much for this quick and efficient review and feedback !
Happy to update and complement the very good work you did in the past on this module

KyleKatarn31:
@binarymaster : please add "translation" tag too (I took the opportunity to fix the French translation in the same files)

KyleKatarn31:
Ready for merge ?

dll/win32/devmgr/devmgmt/MainWindow.cpp
SergeGautherie: I kind of remember @ThFabba telling me code should not include defaults for resources...
Does that not apply here?

KyleKatarn31: Sorry i've not seen any explicit design rule for that and @gedmurphy asked me to change.

PR #2814, opened 2020-05-19
[SHELLFIND] Fix column sorting. CORE-16825

b-r-o-c-k:
## Purpose

Fix column sorting in search results folder.

JIRA issue: [CORE-16825](https://jira.reactos.org/browse/CORE-16825)

### DIFF
--- a/dll/win32/browseui/shellfind/CFindFolder.cpp
+++ b/dll/win32/browseui/shellfind/CFindFolder.cpp
@@ -538,7 +538,20 @@ STDMETHODIMP CFindFolder::BindToStorage(PCUIDLIST_RELATIVE pidl, LPBC pbcReserve
 
 STDMETHODIMP CFindFolder::CompareIDs(LPARAM lParam, PCUIDLIST_RELATIVE pidl1, PCUIDLIST_RELATIVE pidl2)
 {
-    return m_pisfInner->CompareIDs(lParam, _ILGetFSPidl(pidl1), _ILGetFSPidl(pidl2));
+    WORD wColumn = LOWORD(lParam);
+    switch (wColumn)
+    {
+    case 0: // Name
+        break;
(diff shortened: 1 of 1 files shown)


dll/win32/browseui/shellfind/CFindFolder.cpp
```diff
@@ -538,7 +538,20 @@ STDMETHODIMP CFindFolder::BindToStorage(PCUIDLIST_RELATIVE pidl, LPBC pbcReserve
...
+        break;
+    case 1: // Path
+        return MAKE_COMPARE_HRESULT(StrCmpW(_ILGetPath(pidl1), _ILGetPath(pidl2)));
+    case 2: // Relevance
+        return E_NOTIMPL;
+    default: // Default columns
+        wColumn -= _countof(g_ColumnDefs) - 1;
+        break;
+    }
+    return m_pisfInner->CompareIDs(HIWORD(lParam) | wColumn, _ILGetFSPidl(pidl1), _ILGetFSPidl(pidl2));
```
KyleKatarn31, line 554: The basic problem was on File Size... I don't see anything here which would adress the issue. Am I missing something ?

KyleKatarn31:
Why merged ?? Haven't you seen my comment ? This does nothing wrt to the reported issue !

dll/win32/browseui/shellfind/CFindFolder.cpp
b-r-o-c-k: File size is a default column that is already handled by the parent shell folder class. The problem was caused by this function not correctly calling the parent `CompareIDs()` function to handle the default columns.

KyleKatarn31: Ok thanks, very clear.

PR #2816, opened 2020-05-19
[SHLWAPI] lang/*.rc : Enforce aa-BB filenames

SergeGautherie:


binarymaster:
Is this synced with 🍷 ?

SergeGautherie:
> Is this synced with 🍷 ?

Not these parts:
https://source.winehq.org/git/wine.git/blob/HEAD:/dlls/shlwapi/shlwapi.rc

HBelusca:
These (old) file names are those that come out when you generate the corresponding language resource files out of the *.po for shlwapi. The same thing should also happen for all the other wine-synced dlls.

SergeGautherie:
> all the other wine-synced dlls

Yes, first search returns about 20+ similar cases.

dll/win32/shlwapi/shlwapi.rc
```diff
@@ -30,77 +30,77 @@
...
 #ifdef LANGUAGE_RU_RU
-    #include "lang/shlwapi_Ru.rc"
-#endif
-#ifdef LANGUAGE_SK_SK
-    #include "lang/shlwapi_Sk.rc"
+    #include "lang/ru-RU.rc"
 #endif
 #ifdef LANGUAGE_SL_SI
-    #include "lang/shlwapi_Si.rc"
+    #include "lang/si-SI.rc"
```
ThFabba, line 87: The file should be called `sl-SI` (es el, ES EYE)

SergeGautherie: Good catch :-) Fixed.

SergeGautherie:
> I'm not against this but I don't think it's worth spending time on.

Then, I filed [CORE-17110](https://jira.reactos.org/browse/CORE-17110).

ThFabba:
> > I'm not against this but I don't think it's worth spending time on.
> 
> Then, I filed [CORE-17110](https://jira.reactos.org/browse/CORE-17110).

I changed my mind. I am now actively against making more of these changes.

PR #2818, opened 2020-05-19
[CODEOWNERS] Add some translation teams

SergeGautherie:
Commented out, by default. Should some/all be enabled right away?


Extravert-ir:
This feature doesn't work anyway)

KyleKatarn31:
"# If you want to be part of one - hit us at https://chat.reactos.org/" => How can I join the FR one ?

Saibamen:
@KyleKatarn31: Just send request at **Development** channel

ThFabba:
It's obvious what to add once those teams actually exist, so comments don't add value.

PR #2821, opened 2020-05-20
[NTOS:INBV] Fix screen flash and more

feel-the-dz3n:
JIRA issue: [CORE-16786](https://jira.reactos.org/browse/CORE-16786)

### DIFF
--- a/ntoskrnl/inbv/inbv.c
+++ b/ntoskrnl/inbv/inbv.c
@@ -1284,7 +1284,7 @@ DisplayBootBitmap(IN BOOLEAN TextMode)
             /* Save previous screen pixels to buffer */
             InbvScreenToBufferBlt(Buffer, 0, 0, 22, 9, 24);
             /* Draw the progress bar bit */
-            InbvBitBlt(Bar, 0, 0);
+            BitBltPalette(Bar, TRUE, 0, 0);
             /* Store it in global buffer */
             InbvScreenToBufferBlt(RotBarBuffer, 0, 0, 22, 9, 24);
             /* Restore screen pixels */
--- a/ntoskrnl/inbv/inbv.c
+++ b/ntoskrnl/inbv/inbv.c
@@ -1279,7 +1279,7 @@ DisplayBootBitmap(IN BOOLEAN TextMode)
 #endif
 
 #ifdef INBV_ROTBAR_IMPLEMENTED
-        if (Bar)
+        if ((TempRotBarSelection == RB_SQUARE_CELLS) && Bar)
         {
             /* Save previous screen pixels to buffer */
             InbvScreenToBufferBlt(Buffer, 0, 0, 22, 9, 24);


ntoskrnl/inbv/inbv.c
```diff
@@ -1279,7 +1279,7 @@ DisplayBootBitmap(IN BOOLEAN TextMode)
 #endif
 
 #ifdef INBV_ROTBAR_IMPLEMENTED
-        if (Bar)
+        if (ROT_BAR_DEFAULT_MODE == RB_SQUARE_CELLS && Bar)
```
HBelusca, line 1282:
```suggestion
        if ((ROT_BAR_DEFAULT_MODE == RB_SQUARE_CELLS) && Bar)
```

binarymaster: TempRotBarSelection variable should have the value

ntoskrnl/inbv/inbv.c
```diff
@@ -1284,7 +1284,7 @@ DisplayBootBitmap(IN BOOLEAN TextMode)
             /* Save previous screen pixels to buffer */
             InbvScreenToBufferBlt(Buffer, 0, 0, 22, 9, 24);
             /* Draw the progress bar bit */
-            InbvBitBlt(Bar, 0, 0);
+            BitBltPalette(Bar, TRUE, 0, 0);
```
HBelusca, line 1287: When we don't use the rotbar (as it should be by default), do we still get the blink?

ntoskrnl/inbv/inbv.c
feel-the-dz3n: My mistake. Should I use additional brackets with TempRotBarSelection too?

HBelusca: code style says preferably yes.

ntoskrnl/inbv/inbv.c
feel-the-dz3n: If you are talking about using with this condition
```
if (ROT_BAR_DEFAULT_MODE == RB_SQUARE_CELLS && Bar)
```
then we don't have any blink.

Blink happens only with InbvBitBlt, because it draws a palette of Bar image, then we doing fade animation.

ntoskrnl/inbv/inbv.c
```diff
@@ -1279,12 +1279,12 @@ DisplayBootBitmap(IN BOOLEAN TextMode)
 #endif
 
 #ifdef INBV_ROTBAR_IMPLEMENTED
-        if (Bar)
+        if ((TempRotBarSelection == RB_SQUARE_CELLS) && Bar)
```
feel-the-dz3n, line 1282: Fixed here

ntoskrnl/inbv/inbv.c
feel-the-dz3n: I also tested RB_SQUARE_CELLS, everything seems to be fine

HBelusca: Oh yes indeed; and actually the whole mechanism was in place already ^^

PR #2822, opened 2020-05-20
[REACTOS] Remove executable bits on source files

SergeGautherie:
New iteration of 3c7e9bb33243c59261b7214d3c55f1cbdd5a8558.


SergeGautherie:
There are only a few other files, which are .bat/.cmd/.sh, so not touching them.

HBelusca:
> @HBelusca, there are only a few other files, which are .bat/.cmd/.sh, so not touching them.

OK these are fine.

PR #2823, opened 2020-05-21
[NTOSKRNL] Place INIT_FUNCTION before the return type

SergeGautherie:
Follow-up to 71fefa32db013317842fd9224b63bcf5f72f2e32.

### DIFF
--- a/ntoskrnl/cc/view.c
+++ b/ntoskrnl/cc/view.c
@@ -1529,8 +1529,8 @@ CcGetFileObjectFromSectionPtrs (
     return NULL;
 }
 
-VOID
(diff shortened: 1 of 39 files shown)


HBelusca:
There are more with the same "problem".

ThFabba:
According to 71fefa32db013317842fd9224b63bcf5f72f2e32, this is actually required by the compiler in some circumstances, so I'd argue it's not a formatting change. Though since things seem to be compiling as-is, I'm not sure what these circumstances are.

SergeGautherie:
> whole kernel

Done :-)
NB: I checked the whole tree and found no others.

PR #2825, opened 2020-05-21
[SYSSETUP] Make status dialog look like msgina status dialog

feel-the-dz3n:
This patch makes syssetup status dialog look like a msgina status dialog. It makes user feel like it's a system dialog. User won't pay as much attention to the rotating line, as it was with a contrasting progress bar, so user's waiting will be much calmer.

Here is almost the same dialog from Win 2k/XP:

image

image

But it lacks for a progress-showing control, since ReactOS is not so stable. It also shows a branding logo, which is noticeable and memorable for a user.

Screenshot:

image

[YouTube Video](https://youtu.be/-HkEGckt_xk)


dll/win32/syssetup/install.c
```diff
@@ -531,17 +541,108 @@ StatusMessageWindowProc(
...
+                                            IMAGE_BITMAP, 0, 0, LR_DEFAULTCOLOR);
+            GetObject(pDlgData->hBarBitmap, sizeof(bm), &bm);
+            pDlgData->BarWidth = bm.bmWidth;
+            pDlgData->BarHeight = bm.bmHeight;
+
+            if (pDlgData->hLogoBitmap && pDlgData->hBarBitmap)
+            {
+                if (SetTimer(hwndDlg, IDT_BAR, 20, NULL) == 0)
+                {
+                    // ERR("SetTimer(IDT_BAR) failed: %d\n", GetLastError());
```
feel-the-dz3n, line 578: What should I use here?

feel-the-dz3n:
> An idea: you probably can avoid resource duplication in `syssetup.dll` and just load them from `msgina.dll`

It's a good idea but let's keep compatibility with XP - it dupblicates resources too

learn-more:
why closed?

feel-the-dz3n:
> why closed?
accidentally

dll/win32/syssetup/install.c
```diff
@@ -531,17 +541,108 @@ StatusMessageWindowProc(
...
             if (!LoadStringW(hDllInstance, IDS_STATUS_INSTALL_DEV, szMsg, ARRAYSIZE(szMsg)))
                 return FALSE;
             SetDlgItemTextW(hwndDlg, IDC_STATUSLABEL, szMsg);
+
+            return TRUE;
+        }
+
+        case WM_TIMER:
+        {
+            if (pDlgData->hBarBitmap)
```
ThFabba, line 595: If allocating `pDlgData` failed we'll have null here as well. Perhaps `if (uMsg != WM_INITDIALOG && pDlgData == NULL) return FALSE;` or some such at the top, to avoid having to add null checks in each individual case?

feel-the-dz3n:
Fixed

gonzoMD:
> > An idea: you probably can avoid resource duplication in `syssetup.dll` and just load them from `msgina.dll`
> 
> It's a good idea but let's keep compatibility with XP - it dupblicates resources too

Our syssetup is already incompatible in this regard. I believe that, if you want to compatibility, you have at least to use the same resource ID as windows.

However, this is a useful and way too long rotting PR. :(
I think I'll merge this until the end of next week if nothing happens until then.

PR #2830, opened 2020-05-21
[WINSPOOL] GetPrinterA(): Refactor failure handling. CORE-16622

SergeGautherie:
Addendum to ecde3768254664ca40a3f2fa49d1342d4bd92c02.

JIRA issue: [CORE-16622](https://jira.reactos.org/browse/CORE-16622)


JoachimHenze:
Does this fix anything user-observable? If not, then please do not file it against that ticket (which already contains a lot of commits and has been resolved with a fixVer in the past!)

SergeGautherie:
> Does this fix anything user-observable?

I don't know, though it certainly should in case of error.

> do not file it against that ticket

Well, that is the very/only ticket it applies too, as far as I know...

PR #2304, opened 2020-02-01
[WINSPOOL] EnumPrintersA(): Fix failure handling on/after EnumPrintersW(). CORE-16622

SergeGautherie:
Split off from #2298.

Addendum to e8b1778.
JIRA issue: [CORE-16622](https://jira.reactos.org/browse/CORE-16622)

PR #2832, opened 2020-05-22
[WINSPOOL] GetPrinterDriverA(): Refactor failure handling. CORE-16622

SergeGautherie:
Addendum to d90beaeed26863985486874aaa2a1c3422fd6d9f.

JIRA issue: [CORE-16622](https://jira.reactos.org/browse/CORE-16622)

### DIFF
--- a/win32ss/printing/base/winspool/precomp.h
+++ b/win32ss/printing/base/winspool/precomp.h
@@ -42,7 +42,7 @@ SPOOLER_HANDLE, *PSPOOLER_HANDLE;
 extern HANDLE hProcessHeap;
 
 // utils.c
-extern BOOL UnicodeToAnsiInPlace(PWSTR pwszField);
+BOOL UnicodeToAnsiInPlace(PWSTR pwszField);
 
 // devmode.c
 extern void RosConvertAnsiDevModeToUnicodeDevmode(PDEVMODEA pDevModeInput, PDEVMODEW pDevModeOutput);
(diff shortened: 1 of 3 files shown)


JoachimHenze:
Can we avoid linking this to CORE-16622? The issue has a fixVer in the past, and I don't see how the proposed changes would improve that usecase.

SergeGautherie:
> Can we avoid linking this to CORE-16622?

Already answered on #2830.
NB:
I may understand your tester point of view, yet I have a code writer point of view.
Obviously, it would have been much better to fix all these PRs before committing them...

> Will it improve https://jira.reactos.org/browse/CORE-16715 instead maybe?

Already answered on #2304.

SergeGautherie:
> should result in the same

Agreed. Done, then.

> let `UnicodeToAnsiInPlace` return a `DWORD`

I concur: I had thought about it, but I do not want to make all these changes.
This is a minimal fix only.

> especially hard to understand

Well, I simply tried to respect that new way of doing things...

ColinFinck:
> > let `UnicodeToAnsiInPlace` return a `DWORD`
> 
> I concur: I had thought about it, but I do not want to make all these changes.
> This is a minimal fix only.

It's as simple as changing `return TRUE` to `return ERROR_SUCCESS` and `return FALSE` to `return ERROR_NOT_ENOUGH_MEMORY` in `UnicodeToAnsiInPlace` :)

This PR is already making a few things better - thank you for that. Let's also do the last step and make it perfect!
This saves us a discussion in a subsequent PR :)

SergeGautherie:
> > This is a minimal fix only.
> 
> as simple

Alright, added UnicodeToAnsiInPlace commit.

ColinFinck:
Thanks a lot! Looks exactly as I envisioned it :)

PR #2836, opened 2020-05-23
[NTOS:KE] Avoid SEH in trap handlers. CORE-15723

ThFabba:
If SEH is used in a C trap handler, the exception frame will be registered before the call to KiEnterTrap, which means we save the wrong exception frame. We'll therefore also restore this wrong frame for the excepting code, resulting in a stale SEH chain.

We avoid this problem by saving the handler in the assembly trap entry code. However, SEH in a C trap handler function will also not be correctly unwound, since the function never exits (it uses KiEoiHelper instead). This may result in finally handlers not running or other unexpected behavior. Therefore we still forbid doing this through asserts in the C KiEnterTrap variants.

JIRA issue: [CORE-15723](https://jira.reactos.org/browse/CORE-15723)

### DIFF
--- a/ntoskrnl/include/internal/i386/asmmacro.S
+++ b/ntoskrnl/include/internal/i386/asmmacro.S
@@ -207,6 +207,10 @@ set_sane_segs:
         mov fs, ax
     endif
 
+    /* Save ExceptionList, since the C handler might use SEH */
+    mov eax, fs:[KPCR_EXCEPTION_LIST]
+    mov [esp + KTRAP_FRAME_EXCEPTION_LIST], eax
+
 #if DBG
     /* Keep the frame chain intact */
     mov eax, [esp + KTRAP_FRAME_EIP]
--- a/ntoskrnl/ke/i386/traphdlr.c
+++ b/ntoskrnl/ke/i386/traphdlr.c
@@ -1216,56 +1216,11 @@ KiTrap0DHandler(IN PKTRAP_FRAME TrapFrame)
     KiTrapReturn(TrapFrame);
 }
 
-DECLSPEC_NORETURN
-VOID
+BOOLEAN
 FASTCALL
-KiTrap0EHandler(IN PKTRAP_FRAME TrapFrame)
+KiCheckForSListFault(PKTRAP_FRAME TrapFrame)
 {
-    PKTHREAD Thread;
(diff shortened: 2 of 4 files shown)


ntoskrnl/ke/i386/traphdlr.c
```diff
@@ -1216,53 +1216,9 @@ KiTrap0DHandler(IN PKTRAP_FRAME TrapFrame)
     KiTrapReturn(TrapFrame);
 }
 
-DECLSPEC_NORETURN
 VOID
-FASTCALL
-KiTrap0EHandler(IN PKTRAP_FRAME TrapFrame)
+KiCheckForSListFault(PKTRAP_FRAME TrapFrame)
```
HBelusca, line 1220: Not fastcall as well?

ntoskrnl/ke/i386/traphdlr.c
```diff
@@ -1357,7 +1313,56 @@ KiTrap0EHandler(IN PKTRAP_FRAME TrapFrame)
             KiEoiHelper(TrapFrame);
```
HBelusca, line 1313: Does this correctly unwind the two function calls KiCheckForSListFault() and KiTrap0EHandler() ?

ntoskrnl/include/internal/i386/asmmacro.S
```diff
@@ -207,6 +207,9 @@ set_sane_segs:
         mov fs, ax
     endif
 
+    mov eax, fs:[KPCR_EXCEPTION_LIST]
```
HBelusca, line 210: A comment to recall why we are doing this may be in order...

ntoskrnl/ke/i386/traphdlr.c
ThFabba: Good catch, will move it outside this function.

ntoskrnl/ke/i386/traphdlr.c
ThFabba: Sure, why not

ntoskrnl/include/internal/i386/asmmacro.S
ThFabba: Will add.

PR #2527, opened 2020-04-11
Implement kernel sanitizers

Extravert-ir:
Jira issue: [CORE-15771](https://jira.reactos.org/browse/CORE-15771)

TODO:
- [x] UndefinedBehaviorSanitizer (UBSan, `-fsanitize=undefined`) 
- [ ] `-fsanitize=nullability*` group of checks
- [ ] KernelAddressSanitizer (KASan, `-fsanitize=address`)
- [ ] Test on 64bit builds

PR #2846, opened 2020-05-24
[CMAKE] gcc.cmake: Remove unwanted ${MODULE} in an add_compile_flags()

SergeGautherie:
Addendum to 87daca2bea39195687569ae47d51bf186768658d (r64527).


SergeGautherie:
This is also silently embedded in draft #2527...

Extravert-ir:
This goes to #2527

PR #2847, opened 2020-05-24
[FORMATTING] *cmake*: Remove space in 'if ('

SergeGautherie:


SergeGautherie:
For some reason, `sdk/cmake/gcc.cmake` and `sdk/cmake/msvc.cmake` were not committed.
@ColinFinck, could you double-check?

PR #2848, opened 2020-05-24
[NTDLL] Use lowest OS version GUID in the application manifest

tkreuzer:
Use lowest OS version GUID in the application manifest, instead of highest

### DIFF
--- a/dll/ntdll/ldr/ldrinit.c
+++ b/dll/ntdll/ldr/ldrinit.c
@@ -1636,8 +1636,13 @@ LdrpInitializeProcessCompat(PVOID pProcessActctx, PVOID* pOldShimData)
     if (ContextCompatInfo->ElementCount == 0)
         return;
 
-    /* Search for known GUID's, starting from newest to oldest. */
-    for (cur = 0; cur < RTL_NUMBER_OF(KnownCompatGuids); ++cur)
+    /* Search for known GUIDs, starting from oldest to newest.
+       Note that on Windows it is somewhat reversed, starting from the latest known
+       version, going down. But we are not Windows, trying to allow a lower version,
+       we are ReactOS trying to fake a higher version. So we interpret what Windows
+       does as "try the closest version to the actual version", so we start with the
+       lowest version, which is closest to Windows 2003, which we mostly are. */
+    for (cur = RTL_NUMBER_OF(KnownCompatGuids) - 1; cur != -1; --cur)
     {
         for (n = 0; n < ContextCompatInfo->ElementCount; ++n)
         {
@@ -1671,9 +1676,11 @@ LdrpInitializeProcessCompat(PVOID pProcessActctx, PVOID* pOldShimData)
                     *pOldShimData = pShimData;
                 }
 
-                /* Store the highest found version, and bail out. */
+                /* Store the lowest found version, and bail out. */
                 pShimData->dwRosProcessCompatVersion = KnownCompatGuids[cur].Version;
-                DPRINT1("LdrpInitializeProcessCompat: Found guid for winver 0x%x\n", KnownCompatGuids[cur].Version);
+                DPRINT1("LdrpInitializeProcessCompat: Found guid for winver 0x%x in manifest from %wZ\n",
+                        KnownCompatGuids[cur].Version,
+                        &(NtCurrentPeb()->ProcessParameters->ImagePathName));
                 return;
             }
         }


dll/ntdll/ldr/ldrinit.c
```diff
@@ -1636,8 +1636,13 @@ LdrpInitializeProcessCompat(PVOID pProcessActctx, PVOID* pOldShimData)
     if (ContextCompatInfo->ElementCount == 0)
         return;
 
-    /* Search for known GUID's, starting from newest to oldest. */
-    for (cur = 0; cur < RTL_NUMBER_OF(KnownCompatGuids); ++cur)
+    /* Search for known GUID's, starting from oldest to newest.
+       Note that on Windows it is somewhat reversed, starting from the latest known
+       vesion, going down. But we are not Windows, trying to allow a lower version,
+       we are ReactOS trying to fake a higher version. So we interpret what Windows
+       does as "try the closes version to the actual version", so we start with the
```
SergeGautherie, line 1643:
```suggestion
       does as "try the closest version to the actual version", so we start with the
```

dll/ntdll/ldr/ldrinit.c
```diff
@@ -1673,7 +1678,9 @@ LdrpInitializeProcessCompat(PVOID pProcessActctx, PVOID* pOldShimData)
 
                 /* Store the highest found version, and bail out. */
```
SergeGautherie, line 1679:
```suggestion
                /* Store the lowest found version, and bail out. */
```

dll/ntdll/ldr/ldrinit.c
HBelusca, line 1643: "closes" --> "closest".

dll/ntdll/ldr/ldrinit.c
```diff
@@ -1636,8 +1636,13 @@ LdrpInitializeProcessCompat(PVOID pProcessActctx, PVOID* pOldShimData)
...
 
-    /* Search for known GUID's, starting from newest to oldest. */
-    for (cur = 0; cur < RTL_NUMBER_OF(KnownCompatGuids); ++cur)
+    /* Search for known GUID's, starting from oldest to newest.
+       Note that on Windows it is somewhat reversed, starting from the latest known
+       vesion, going down. But we are not Windows, trying to allow a lower version,
+       we are ReactOS trying to fake a higher version. So we interpret what Windows
+       does as "try the closes version to the actual version", so we start with the
+       lowest version, which is closest to Windows 2003, which we mostly are. */
+    for (cur = RTL_NUMBER_OF(KnownCompatGuids) - 1; cur != -1; --cur)
```
SergeGautherie, line 1645: Suggestion: reverse order in `KnownCompatGuids`, instead.

dll/ntdll/ldr/ldrinit.c
```diff
@@ -1636,8 +1636,13 @@ LdrpInitializeProcessCompat(PVOID pProcessActctx, PVOID* pOldShimData)
     if (ContextCompatInfo->ElementCount == 0)
         return;
 
-    /* Search for known GUID's, starting from newest to oldest. */
-    for (cur = 0; cur < RTL_NUMBER_OF(KnownCompatGuids); ++cur)
+    /* Search for known GUID's, starting from oldest to newest.
```
ColinFinck, line 1639: nitpick:
```suggestion
    /* Search for known GUIDs, starting from oldest to newest.
```

dll/ntdll/ldr/ldrinit.c
```diff
@@ -1636,8 +1636,13 @@ LdrpInitializeProcessCompat(PVOID pProcessActctx, PVOID* pOldShimData)
...
+       Note that on Windows it is somewhat reversed, starting from the latest known
+       vesion, going down. But we are not Windows, trying to allow a lower version,
```
ColinFinck, line 1641: nitpick:
```suggestion
       version, going down. But we are not Windows, trying to allow a lower version,
```

dll/ntdll/ldr/ldrinit.c
ColinFinck, line 1645: This is fine, however the most popular pattern for a reverse loop in C seems to be:
```suggestion
    for (cur = RTL_NUMBER_OF(KnownCompatGuids) - 1; --cur >= 0;)
```

Change at will.

tkreuzer: This wouldn't work, because cur is unsigned. It's also broken, because it starts with ```cur = RTL_NUMBER_OF(KnownCompatGuids) - 2```. Additionally it breaks the logic in combination with continue and break. And I find it ugly :)

PR #2849, opened 2020-05-24
[REACTOS] Lower case 'RunTmChk'

SergeGautherie:
Addendum to 28644220497bedab1aaccbe0a7424b175a765a16 (r64808).
Addendum to 400a2c65cf78ce95c30d33146adaa219c608b32d (r64816).


SergeGautherie:
The RUNTMCHK commit is also silently embedded in draft #2527...

Extravert-ir:
I don't want to spoil commit log with this, everything is to be done in #2527

PR #2852, opened 2020-05-25
[NTDLL_APITEST] Add ProcessWx86Information testcase for NtXxxInformat…

Fraizeraust:
…ionProcess

Provide a testcase for `ProcessWx86Information` class to ensure its proper alignment size as well as some other stuff. I've to figure out later how can I successfully assign the VDM power to a process.
**Windows XP**
Capture
**Windows Server 2003**
Capture2
**ReactOS**
Capture4

### DIFF
--- a/modules/rostests/apitests/ntdll/NtQueryInformationProcess.c
+++ b/modules/rostests/apitests/ntdll/NtQueryInformationProcess.c
@@ -253,6 +253,76 @@ Test_ProcessPriorityClassAlignment(void)
     free(ProcPriority);
 }
 
+static
+void
+Test_ProcessWx86Information(void)
+{
+    NTSTATUS Status;
+    ULONG VdmPower = 1, ReturnLength;
+
+    /* Everything is NULL */
+    Status = NtQueryInformationProcess(NULL,
+                                       ProcessWx86Information,
+                                       NULL,
+                                       0,
+                                       NULL);
+    ok_hex(Status, STATUS_INFO_LENGTH_MISMATCH);
+
+    /* Given an invalid process handle */
+    Status = NtQueryInformationProcess(NULL,
+                                       ProcessWx86Information,
+                                       &VdmPower,
+                                       sizeof(VdmPower),
+                                       NULL);
+    ok_hex(Status, STATUS_INVALID_HANDLE);
+
+    /* Don't query anything */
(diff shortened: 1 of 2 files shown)


modules/rostests/apitests/ntdll/NtQueryInformationProcess.c
```diff
@@ -253,6 +253,76 @@ Test_ProcessPriorityClassAlignment(void)
...
+
+    /* Query the VDM power */
+    Status = NtQueryInformationProcess(NtCurrentProcess(),
+                                       ProcessWx86Information,
+                                       &VdmPower,
+                                       sizeof(VdmPower),
+                                       &ReturnLength);
+    ok_hex(Status, STATUS_SUCCESS);
+    ok(ReturnLength != 0, "ReturnLength shouldn't be 0!\n");
+    ok(VdmPower == 0 || VdmPower == 1, "The VDM power value must be within the boundary between 0 and 1, not anything else! Got %lu\n", VdmPower);
```
ThFabba, line 319: Should init `VdmPower` before the `NtQueryInformationProcess` to make the test more robust.

modules/rostests/apitests/ntdll/NtQueryInformationProcess.c
```diff
@@ -253,6 +253,76 @@ Test_ProcessPriorityClassAlignment(void)
...
+    ok_hex(Status, STATUS_DATATYPE_MISALIGNMENT);
+
+    /* Query the VDM power */
+    Status = NtQueryInformationProcess(NtCurrentProcess(),
+                                       ProcessWx86Information,
+                                       &VdmPower,
+                                       sizeof(VdmPower),
+                                       &ReturnLength);
+    ok_hex(Status, STATUS_SUCCESS);
+    ok(ReturnLength != 0, "ReturnLength shouldn't be 0!\n");
```
ThFabba, line 318: Does the value vary, or why is the check so generic? I'd assume it'd always be `sizeof(VdmPower)`?

Fraizeraust: My idea is to inspect the internal behaviour of `NtQueryInformationXxx` when returning the length in the last argument call across different information classes. In this case, the length of the requested information is the same like `sizeof(VdmPower)`.

If that test check is too redundant I can simply remove it from the testcase code.

PR #2855, opened 2020-05-25
[MMSYS] OnHScroll(): Fix wrong '&&' operator. CORE-17079

SergeGautherie:
JIRA issue: [CORE-17079](https://jira.reactos.org/browse/CORE-17079)


HBelusca:
It is clear from the logic that || was indeed the intended operator (aka. if Id is outside of the numerical range, then return).

PR #2856, opened 2020-05-25
[NFSD] readdir_copy_shortname(): Add needed brackets to '*name_size_out'. CORE-17078

SergeGautherie:
Obviously, increase the value, not move the pointer!

JIRA issue: [CORE-17078](https://jira.reactos.org/browse/CORE-17078)

### DIFF
--- a/base/services/nfsd/readdir.c
+++ b/base/services/nfsd/readdir.c
@@ -165,7 +165,11 @@ static void readdir_copy_shortname(
     /* GetShortPathName returns number of characters, not including \0 */
     *name_size_out = (CCHAR)GetShortPathNameW(name, name_out, 12);
     if (*name_size_out) {
+#ifndef __REACTOS__
         *name_size_out++;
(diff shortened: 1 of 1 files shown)


base/services/nfsd/readdir.c
```diff
@@ -165,7 +165,11 @@ static void readdir_copy_shortname(
     /* GetShortPathName returns number of characters, not including \0 */
     *name_size_out = (CCHAR)GetShortPathNameW(name, name_out, 12);
     if (*name_size_out) {
+#ifndef __REACTOS__
```
binarymaster, line 168: Is this 3rd party code? Care to upstream the fix?

SergeGautherie: Yes. Not now.
NB:
First, that import needs to be added to `media/doc/README.FSD`: upstream source, forked or sync'ed, ...

PR #2860, opened 2020-05-25
[NTOS:INBV] Boot screen 😎

feel-the-dz3n:
### Goal

To refresh the existing boot screen. An attempt to make it more modern.

### Preview

ros-new-boot-screen

And post-shutdown screen:

reactos-po

[And GIF (outdated)](https://user-images.githubusercontent.com/25367511/82846177-44b0d280-9ef0-11ea-9208-ac92f9f20de1.gif)

### Downloads

Go to the checks tab and download artifacts there


HBelusca:
To make everybody happy one needs to write **modular** boot "themes".

feel-the-dz3n:
> To make everybody happy one needs to write **modular** boot "themes".

And perhaps a modular blue screen?

amber8706:
HBelusca: "everybody" can make their build ROS with themes, boot screen, blue screen etc. For this not need create "pull request".

minecraftchest1:
I think there should be a spinner, such as the one in windows 10. I like it.

feel-the-dz3n:
> I think there should be a spinner, such as the one in windows 10. I like it.

https://www.google.com/search?q=spinner

minecraftchest1:
Forgive me, I mean the circle that spins during boot up on windows 10.

feel-the-dz3n:
> Forgive me, I mean the circle that spins during boot up on windows 10.

There's something like this when using square cells progress bar type.

minecraftchest1:
I've seen it, a d it isn't the same.

Fraizeraust:
Do you realize GitHub is not for chattering? Go with that elsewhere and not here.

myslivets:
Not bad! I like it.

sziberov:
Removing everything at bottom can make it even more modern.

osdeverr:
Looks fresh, I like it.### re

tkreuzer:
Wasn't there some magic that prevents people from exchanging the boot screen? Or has this been removed?

HBelusca:
What kind of "magic" were you thinking about?

tkreuzer:
> 
> 
> What kind of "magic" were you thinking about?

Alex Ionescu type of haxxory in the kernel and hal.

feel-the-dz3n:
[image](https://www.youtube.com/watch?v=U7CZcd-UYmU)

JoachimHenze:
feel-the-dz3n what is the binary size of ntoskrnl before and after this PR?

JoachimHenze (approved):
binary size comparison ntoskrnl.exe for GCC8.4.0 dbg i386 linux:
4.845.568Bytes 0.4.15-dev-2937-g1078340-x86-gcc-lin-dbg vanilla master head, standard bootlogo
4.837.376Bytes with the new dz3n bootlogo
Since it does decrease the binary size, I do give my personal thumbs-up for this PR!

ColinFinck (changes requested):
The important thing about design changes is that they need to be consistent. There is no point in changing a single bitmap and calling the job done. In fact, any such action, even with good intentions, makes the overall design look worse, because you are now mixing totally different graphics for the same thing.

---

ReactOS 0.4.13 still has a single ReactOS logo everywhere:
grafik
grafik

It's not a coincidence at all that this also matches the logo on our website:

grafik

---

Current ReactOS HEAD already got a bit messy. A new msgina/winver bitmap was added in #3621 - a PR that went through without any mutual consensus:

grafik

Neither the text is properly aligned here, nor "OS" has the right color. I also wonder where the logo was taken from considering that the proportions of the "ball" and its surroundings don't match those of the other logos.

Next come the fancy shutdown and logoff dialogs, which use a variant of the simplified logo:

grafik

I see that this logo fits well to the design of the fancy dialog, and possibly to one of our themes. But that still makes it look foreign in the context of the remaining ReactOS graphics.

---

Your PR adds the ["stamp" logo](https://github.com/reactos/press-media/blob/master/Art/Logo/Orb/Simplified/ReactOS_Orb_Stamp.svg) on top of the logo mess we already have. As the name says, this logo is only used for the official ReactOS stamp and nothing else. I had asked the community in 2019 and a majority preferred the [simplified logo](https://github.com/reactos/press-media/blob/master/Art/Shirts%20and%20Hoodies/ReactOS_Logo_Black.svg)
Hence, this logo is also used for all products in our merchandising shop: https://shop.spreadshirt.net/reactos/

I see that none of this mess is your fault. We are in fact in a limbo situation between the traditional ReactOS logo (that has been used almost everywhere) and the simplified one (that is the current fashion, but also looks more appropriate for some situations).
But we must not make this situation even worse. Which is why I need to reject this PR as long as we haven't come to a common understanding about the ReactOS design.

Designers are welcome to join the discussion.

HBelusca:
> The important thing about design changes is that they need to be consistent. There is no point in changing a single bitmap and calling the job done. In fact, any such action, even with good intentions, makes the overall design look worse, because you are now mixing totally different graphics for the same thing.

Especially when the changed logo look only makes sense for **only one** associated win32 UI theme (lautus or mizu).

feel-the-dz3n:
That makes sense. Thanks.

пн, 19 июл. 2021 г., 21:50 Hermès BÉLUSCA - MAÏTO ***@***.***

feel-the-dz3n:
I am wondering, what if I use the official logo here instead of the flat one? If so, this branch will introduce only a new rounded progress bar (which is actually an inconsistent progress bar design too 🤔) and a few minor tweaks. What do you think?

@ColinFinck @HBelusca @JoachimHenze

HBelusca:
Another alternative is to not touch what is already fine.

feel-the-dz3n:
Sure, it's a good alternative.

I feel a strong reluctance to these changes in your words. Just in case.
Let's clarify the meaning of the changes to avoid misunderstandings. I
understand that it works even without any changes, but this is not about
the boot screen implementation. I believe that the layout change will make
not problems, but only positive emotions, if any user notice this. I also
understand that current version of this patch makes interface parts
inconsistent. So, according to the target of the pull request (to refresh
20 years old boot screen layout), I am looking for possible solutions, and
I am wondering could something of already made changes be used, if the
original logo would be used too.

Thanks!

вт, 20 июл. 2021 г., 23:07 Hermès BÉLUSCA - MAÏTO ***@***.***

HBelusca:
To clarify: there wouldn't be much discussion if these changes could be selected at runtime, much like any usual UI theme. However here, they are static (since compiled in the kernel).

feel-the-dz3n:
In addition to this I want to say, that I respect the existing, as told
before, already working layout. But, I think some parts are looking
outdated, and I described my suggestions in the patch.


ср, 21 июл. 2021 г., 00:05 Ярослав Кибыш ***@***.***>:

feel-the-dz3n:
Great idea!

ср, 21 июл. 2021 г., 00:10 Hermès BÉLUSCA - MAÏTO ***@***.***

GeoB99:
Seeing the state of this PR and remarks the developers have given here in regard of a custom "modernized" boot splash screen, I would safely say that the pull request can be closed. On my consideration I appreciate your design work on this but truth ought be told here, it only makes consistency worse and I side with Colin on this one (not that current look is any perfect either).

What I can think of is a general discussion meetup between developers and potential designer artists who can come up with a new design idea for ReactOS (boot screen, logo, default theme style, whatever) and maybe your work can be considered. And from there we can judge what will the final outlook of ReactOS will be.

Although right now it's best to just not touch the boot screen or anything else at all. I hope my action done to this PR won't be taken with bad faith.

PR #2857, opened 2020-05-25
[CSRSRV] CsrGetProcessLuid(): Check 1st NtQueryInformationToken() res…

SergeGautherie:
…ult too

Detected by Cppcheck: redundantAssignment.

PR #2861, opened 2020-05-25
[REACTOS] Properly check 1st NtQueryInformationToken() result

SergeGautherie:
(Everywhere) We don't want to continue with `size = 0`, do we?

Follow-up to #2857.

### DIFF
--- a/dll/win32/advapi32/reg/reg.c
+++ b/dll/win32/advapi32/reg/reg.c
@@ -3494,6 +3494,12 @@ ReadTokenSid:
                                      &RequiredLength);
     if (Status != STATUS_BUFFER_TOO_SMALL)
     {
+        if (NT_SUCCESS(Status))
+        {
+            ERR("NtQueryInformationToken(0) succeeded unexpectedly (Status 0x%08lx)\n", Status);
+            return ERROR_NO_TOKEN;
+        }
+
         /* NOTE - as opposed to all other registry functions windows does indeed
                   change the last error code in case the caller supplied a invalid
                   handle for example! */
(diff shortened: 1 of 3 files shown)


dll/win32/advapi32/reg/reg.c
```diff
@@ -3492,7 +3492,7 @@ RegOpenUserClassesRoot(IN HANDLE hToken,
                                      NULL,
                                      0,
                                      &RequiredLength);
-    if (!NT_SUCCESS(Status) && (Status != STATUS_BUFFER_TOO_SMALL))
+    if (Status != STATUS_BUFFER_TOO_SMALL)
```
ThFabba, line 3495: This is bogus. You're now returning success here but no valid key handle.

SergeGautherie: Agreed. Fixed.

dll/win32/advapi32/reg/reg.c
```diff
@@ -3492,8 +3492,14 @@ RegOpenUserClassesRoot(IN HANDLE hToken,
...
     {
+        if (Status == STATUS_SUCCESS)
```
ThFabba, line 3497: There are about 2 billion other success status codes.
I'd suggest
```
if (NT_SUCCESS(Status))
{
    ...
}
else if (Status != STATUS_BUFFER_TOO_SMALL)
{
    ...
}
```

SergeGautherie: > `NT_SUCCESS(Status)`

Obviously. Fixed.

> `if else if`

While that is somewhat clearer to read, I feel it is better to have only one block of "error" handling, especially as we always expect `STATUS_BUFFER_TOO_SMALL` result.

SergeGautherie:
> requested changes

@ThFabba, please update status.

PR #2862, opened 2020-05-25
[CSRSRV] CsrSetProcessSecurity(): Check 1st NtQueryInformationToken()…

SergeGautherie:
… result

Also:
* Add 1 NtClose(hToken), in an error case.
* Do not call RtlFreeHeap(..., ..., NULL).

Follow-up to #2857.

### DIFF
--- a/subsystems/win32/csrsrv/init.c
+++ b/subsystems/win32/csrsrv/init.c
@@ -74,12 +74,18 @@ CsrSetProcessSecurity(VOID)
     if (!NT_SUCCESS(Status)) goto Quickie;
 
     /* Get the Token User Length */
-    NtQueryInformationToken(hToken, TokenUser, NULL, 0, &Length);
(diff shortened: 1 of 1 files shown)


subsystems/win32/csrsrv/init.c
```diff
@@ -153,7 +159,7 @@ CsrSetProcessSecurity(VOID)
     /* Free the memory and return */
 Quickie:
     if (ProcSd) RtlFreeHeap(CsrHeap, 0, ProcSd);
-    RtlFreeHeap(CsrHeap, 0, TokenInfo);
+    if (TokenInfo) RtlFreeHeap(CsrHeap, 0, TokenInfo);
```
Saibamen, line 162:
```suggestion
    if (TokenInfo)
        RtlFreeHeap(CsrHeap, 0, TokenInfo);
```

HBelusca: Let's not modify in an non-consistent manner the formatting, please.

PR #2864, opened 2020-05-26
[CSRSRV] CsrUnhandledExceptionFilter: Check RtlAdjustPrivilege() result

SergeGautherie:
And remove unused NtRaiseHardError() assignment.

Detected by Cppcheck: redundantAssignment.
Addendum to d221bdfbb1581227c0933e7cf93c7de9cdead01d (r55647).

### DIFF
--- a/subsystems/win32/csrsrv/server.c
+++ b/subsystems/win32/csrsrv/server.c
@@ -650,7 +650,6 @@ CsrUnhandledExceptionFilter(IN PEXCEPTION_POINTERS ExceptionInfo)
                                         TRUE,
                                         TRUE,
                                         &OldValue);
-
             /* Use the Process token if that failed */
             if (Status == STATUS_NO_TOKEN)
             {
@@ -658,6 +657,11 @@ CsrUnhandledExceptionFilter(IN PEXCEPTION_POINTERS ExceptionInfo)
                                             TRUE,
                                             FALSE,
                                             &OldValue);
+            }
+            if (!NT_SUCCESS(Status))
+            {
+                DPRINT1("CsrUnhandledExceptionFilter(): RtlAdjustPrivilege(SE_SHUTDOWN_PRIVILEGE) failed, Status = 0x%08lx\n", Status);
+                goto NoPrivilege;
             }
 
             /* Initialize our Name String */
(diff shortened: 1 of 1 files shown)


subsystems/win32/csrsrv/server.c
```diff
@@ -153,12 +153,11 @@ CsrLoadServerDll(IN PCHAR DllString,
...
 
             /* Send a hard error */
-            NtRaiseHardError(Status,
-                             2,
-                             3,
-                             Parameters,
-                             OptionOk,
-                             &Response);
+            Status2 = NtRaiseHardError(Status, 2, 3, Parameters, OptionOk, &Response);
+            if (!NT_SUCCESS(Status2))
```
HBelusca, line 157: It doesn't really make sense to look for the returned status of NtRaiseHardError, if it's just for dprinting stuff in the debug log. The idea here was to just raise the hard-error, and then the code will continue execution.

subsystems/win32/csrsrv/server.c
```diff
@@ -676,8 +679,13 @@ CsrUnhandledExceptionFilter(IN PEXCEPTION_POINTERS ExceptionInfo)
                                       ErrorParameters,
                                       OptionShutdownSystem,
                                       &Response);
+            if (!NT_SUCCESS(Status))
```
HBelusca, line 682: What for? The CSRSS is killed just afterwards, so there is no real point in telling why NtRaiseHardError could have failed. (It's not like if we wanted to then do actions depending on the returned status.)

subsystems/win32/csrsrv/server.c
```diff
@@ -153,12 +153,11 @@ CsrLoadServerDll(IN PCHAR DllString,
             RtlInitUnicodeString(&ErrorString, L"Default Load Path");
 
             /* Send a hard error */
-            NtRaiseHardError(Status,
-                             2,
-                             3,
-                             Parameters,
-                             OptionOk,
-                             &Response);
+            Status2 = NtRaiseHardError(Status, 2, 3, Parameters, OptionOk, &Response);
```
HBelusca, line 156: Please restore the formatting (and we don't really need the returned Status code, see below).

SergeGautherie: Reverted.

subsystems/win32/csrsrv/server.c
SergeGautherie: Reverted, then.

subsystems/win32/csrsrv/server.c
SergeGautherie: Reverted, then.

PR #2867, opened 2020-05-27
[LABEL] wmain(): Initialize 'szBuffer'. CORE-9444

SergeGautherie:
Detected by Cppcheck: uninitvar.
Addendum to 6bfe4f68aff2011f2a840ddfe16cd42067033d8e.
JIRA issue: [CORE-9444](https://jira.reactos.org/browse/CORE-9444)

### DIFF
--- a/base/applications/cmdutils/label/label.c
+++ b/base/applications/cmdutils/label/label.c
@@ -142,17 +142,14 @@ PromptYesNo(VOID)
 int wmain(int argc, WCHAR *argv[])
 {
     WCHAR szRootPath[] = L" :\\";
-    WCHAR szBuffer[80];
-    WCHAR szLabel[80];
+    WCHAR szBuffer[80] = L"";
+    WCHAR szLabel[80] = L"";
     WCHAR szOldLabel[80];
     DWORD dwSerialNr;
     INT len, i;
 
     /* Initialize the Console Standard Streams */
     ConInitStdStreams();
-
-    /* set empty label string */
-    szLabel[0] = UNICODE_NULL;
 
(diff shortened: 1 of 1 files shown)


JoachimHenze:
It does not make any sense to flag that as an addendum to 6bfe4f6, or to link it against CORE-9444, because in that commit, the code was not changed, just moved. Creating such false references only spams the JIRA tracker and erroneously links commits to ancient tickets and bugreports where they have absolutely no relation to. Please stop that practice (of referencing "git blame commit" of latest touch) if code was only moved.

HBelusca:
That's why in the commit to master I removed these references.

JoachimHenze:
Good hbelusca.
Still it would help to keep JIRA in shape, if SergeGautherie would follow that pattern when doing future PRs.
Because JIRA does still see that stuff as related erroneously.

SergeGautherie:
> not changed, just moved

@JoachimHenze, read that commit again: old code was _changed_ and especially that variable is _new_!

JoachimHenze:
You are right, SergeGautherie, the whole function was newly introduced by that commit:
int wmain(int argc, WCHAR *argv[])
and therefore also that buffer. So your relation "as addendum to commit xyz" was legit, I agree.
Still we should try to not reuse JIRA tickets that are in resolved state already and have set a fixVer in the past.

SergeGautherie:
> JIRA tickets that are in resolved state already and have set a fixVer in the past.

@JoachimHenze, check CORE-9444 again: it is _not_ resolved and has _no_ fixVer!

PR #2869, opened 2020-05-27
[CMAKE] Group files in the project file based on directory layout

learn-more:
Result:
Image Pasted at 2020-5-27 10-39

Before:

Image Pasted at 2020-5-27 15-40


HBelusca:
Looks like the `[MODULE]` thing would be instead `[NTOSKRNL]`.

learn-more:
> Note: this is CMake 3.8+

`cmake version 3.17.1-ReactOS`
We are in the clear then :)

PR #2870, opened 2020-05-28
[NETCFGX] LoadDNSSettings(): Improve RegQueryValueExW() calls a bit

SergeGautherie:
Re-initialize size when needed (only).

### DIFF
--- a/dll/win32/netcfgx/tcpipconf_notify.c
+++ b/dll/win32/netcfgx/tcpipconf_notify.c
@@ -2719,6 +2719,8 @@ LoadDNSSettings(
     {
         dwSize = sizeof(DWORD);
         RegQueryValueExW(hKey, L"RegisterAdapterName", NULL, NULL, (LPBYTE)&This->pCurrentConfig->pDNS->RegisterAdapterName, &dwSize);
+
+        dwSize = sizeof(DWORD);
         RegQueryValueExW(hKey, L"RegistrationEnabled", NULL, NULL, (LPBYTE)&This->pCurrentConfig->pDNS->RegistrationEnabled, &dwSize);
 
         dwSize = sizeof(This->pCurrentConfig->pDNS->szDomain);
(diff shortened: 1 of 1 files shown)


dll/win32/netcfgx/tcpipconf_notify.c
```diff
@@ -2732,7 +2734,6 @@ LoadDNSSettings(
         dwSize = sizeof(DWORD);
         RegQueryValueExW(hKey, L"UseDomainNameDevolution", NULL, NULL, (LPBYTE)&This->pCurrentConfig->pDNS->UseDomainNameDevolution, &dwSize);
 
-        dwSize = 0;
```
ThFabba, line 2735: It's still good practice to do this, even though it's not technically necessary.

dll/win32/netcfgx/tcpipconf_notify.c
```diff
@@ -2719,6 +2719,8 @@ LoadDNSSettings(
     {
         dwSize = sizeof(DWORD);
         RegQueryValueExW(hKey, L"RegisterAdapterName", NULL, NULL, (LPBYTE)&This->pCurrentConfig->pDNS->RegisterAdapterName, &dwSize);
+
+        dwSize = sizeof(This->pCurrentConfig->pDNS->RegistrationEnabled);
```
ThFabba, line 2723: Stay consistent. Everywhere else uses `sizeof(DWORD)`

dll/win32/netcfgx/tcpipconf_notify.c
ThFabba: The function takes `*lpcbData` as an input. We pass a buffer pointer that is not valid, so we should indicate that it has zero writable bytes in the parameter intended for this. The documentation for this function says that this is not necessary -- but passing a value of zero makes the code so obviously correct that you don't need to even look at the documentation.
Hundreds and hundreds of C functions work this way, and `NULL, 0` or `NULL, &variableWithZeroValue` will work for all of them.
If you find this confusing, that's an indication of your inexperience with common C programming paradigms, not an a problem with this code.

dll/win32/netcfgx/tcpipconf_notify.c
ThFabba: This function uses `sizeof(DWORD)` everywhere, except the case where the buffer size is nontrivial.
Your change made this simple DWORD case look like the complicated string case below, instead of looking like all the other cases that are actually like it.

This is what style is useful for: making similar things look similar, so that differences stand out.

HBelusca:
Committed in 62b4c61f82b4fb29faac651414951a06969cf17d.

dll/win32/netcfgx/tcpipconf_notify.c
SergeGautherie: P.S.:
Well, I assumed ReactOS would be a place where people would know/check how Windows functions work.
Anyway, I can understand your explanation and never said that was causing a bug.

PR #2871, opened 2020-05-28
[RAPPS] add a layer of window above the richedit

kernelbin:
## Purpose

I add a window layer, now the richedit does not belong to the main window directly.

why I am doing so is because:
- the main window has undertaken too many tasks.
- later I want to add a window at the left side of the richedit in order to display a snapshot or an icon

### DIFF
--- a/base/applications/rapps/gui.cpp
+++ b/base/applications/rapps/gui.cpp
@@ -391,11 +391,13 @@ public:
             
             ShellExecuteW(m_hWnd, L"open", pLink, NULL, NULL, SW_SHOWNOACTIVATE);
             HeapFree(GetProcessHeap(), 0, pLink);
+			pLink = NULL;
             break;
 
         case ID_COPY_LINK:
             CopyTextToClipboard(pLink);
             HeapFree(GetProcessHeap(), 0, pLink);
+			pLink = NULL;
             break;
 
         }
--- a/base/applications/rapps/gui.cpp
+++ b/base/applications/rapps/gui.cpp
@@ -1,4 +1,4 @@
-﻿/*
+/*
  * PROJECT:     ReactOS Applications Manager
  * LICENSE:     GPL-2.0+ (https://spdx.org/licenses/GPL-2.0+)
  * FILE:        base/applications/rapps/gui.cpp
@@ -53,14 +53,16 @@ INT GetSystemColorDepth()
     return ColorDepth;
 }
 
-class CAppRichEdit: 
+LRESULT CALLBACK AppInfoWndProc(HWND hwnd, UINT message, WPARAM wParam, LPARAM lParam);
+
+class CAppRichEdit :
     public CUiWindow<CRichEdit>
 {
 private:
     VOID LoadAndInsertText(UINT uStringID,
-                           const ATL::CStringW& szText,
-                           DWORD StringFlags,
-                           DWORD TextFlags)
+        const ATL::CStringW& szText,
+        DWORD StringFlags,
+        DWORD TextFlags)
     {
         ATL::CStringW szLoadedText;
         if (!szText.IsEmpty() && szLoadedText.LoadStringW(uStringID))
@@ -71,7 +73,7 @@ private:
     }
 
     VOID LoadAndInsertText(UINT uStringID,
-                                       DWORD StringFlags)
+        DWORD StringFlags)
     {
         ATL::CStringW szLoadedText;
         if (szLoadedText.LoadStringW(uStringID))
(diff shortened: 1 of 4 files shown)


base/applications/rapps/gui.cpp
```diff
@@ -257,6 +259,164 @@ class CAppRichEdit:
...
+
+    HWND Create(HWND hwndParent)
+    {
+        if (!bClassRegistered)
+        {
+            WNDCLASSW wndclass = { 0 };
+            wndclass.style = CS_HREDRAW | CS_VREDRAW;
+            wndclass.lpfnWndProc = AppInfoWndProc;
+            wndclass.cbClsExtra = 0;
+            wndclass.cbWndExtra = sizeof(CAppInfoDisplay);
```
learn-more, line 296:
```suggestion
            wndclass.cbWndExtra = sizeof(CAppInfoDisplay*);
```

base/applications/rapps/gui.cpp
```diff
@@ -257,6 +259,164 @@ class CAppRichEdit:
     }
 };
 
+class CAppInfoDisplay :
+    public CUiWindow<>
+{
+    BOOL bClassRegistered = FALSE;
```
learn-more, line 265:
```suggestion
    static BOOL bClassRegistered = FALSE;
```

base/applications/rapps/gui.cpp
```diff
@@ -257,6 +259,164 @@ class CAppRichEdit:
...
+    }
+
+    VOID OnLink(ENLINK* Link)
+    {
+        switch (Link->msg)
+        {
+        case WM_LBUTTONUP:
+        case WM_RBUTTONUP:
+        {
+            if (pLink) HeapFree(GetProcessHeap(), 0, pLink);
```
learn-more, line 325: This whole `pLink` thing could use some love, I am pretty sure there are some bugs due to that...
(Not something for now, but to keep in mind that if you work on it, evaluate if the whole thing can be made simpler)

base/applications/rapps/gui.cpp
```diff
@@ -257,6 +259,164 @@ class CAppRichEdit:
...
+    VOID OnCommand(WPARAM wParam, LPARAM lParam)
+    {
+        WORD wCommand = LOWORD(wParam);
+
+        switch (wCommand)
+        {
+        case ID_OPEN_LINK:
+            
+            ShellExecuteW(m_hWnd, L"open", pLink, NULL, NULL, SW_SHOWNOACTIVATE);
+            HeapFree(GetProcessHeap(), 0, pLink);
```
learn-more, line 354: For example, stuff like this.
Now pLink is a dead pointer...

base/applications/rapps/gui.cpp
```diff
@@ -257,6 +259,164 @@ class CAppRichEdit:
     }
 };
 
+class CAppInfoDisplay :
+    public CUiWindow<>
```
learn-more, line 263: If you don't empty out the template arguments, but use something that is a `CWindow` or even `CWindow` itself,
you can use the ATL mechanism of creating the window, registering the class etc.
This removes the need of you toying around with the pointer, associating it with your class etc.

learn-more: See for example the `RichEdit` you use.

kernelbin: > If you don't empty out the template arguments, but use something that is a `CWindow` or even `CWindow` itself,
> you can use the ATL mechanism of creating the window, registering the class etc.
> This removes the need of you toying around with the pointer, associating it with your class etc.

I'm not familiar with ATL, so at first, I create the window in the most traditional way 😅. now corrected

base/applications/rapps/gui.cpp
```diff
@@ -257,6 +257,156 @@ class CAppRichEdit:
     }
 };
 
+class CAppInfoDisplay :
+    public CUiWindow<CWindowImpl<CAppInfoDisplay>>
+{
+    static BOOL bClassRegistered;
```
learn-more, line 263: This is not needed anymore :)

kernelbin: now deleted

base/applications/rapps/gui.cpp
```diff
@@ -257,6 +257,154 @@ class CAppRichEdit:
...
+    {
+        WORD wCommand = LOWORD(wParam);
+
+        switch (wCommand)
+        {
+        case ID_OPEN_LINK:
+            
+            ShellExecuteW(m_hWnd, L"open", pLink, NULL, NULL, SW_SHOWNOACTIVATE);
+            HeapFree(GetProcessHeap(), 0, pLink);
+			pLink = NULL;
```
learn-more, line 394: You seem to have introduced tabs here, instead of spaces

kernelbin: ah yes. my bad habit. corrected.

HBelusca: You should consider this: https://stackoverflow.com/questions/30140595/show-whitespace-characters-in-visual-studio-code 😄

kernelbin: > You should consider this: https://stackoverflow.com/questions/30140595/show-whitespace-characters-in-visual-studio-code 😄

got it. actually I'm using Visual Studio and I've just opened the option to insert spaces instead of tabs. hope I will no longer make this mistake.

KyleKatarn31:
Before/After screenshots please.

kernelbin:
> Before/After screenshots please.

for this PR, the UI is not changed at all. I just add one more window acting as a logic layer

KyleKatarn31:
> 
> 
> > Before/After screenshots please.
> 
> for this PR, the UI is not changed at all. I just add one more window acting as a logic layer

Ok, so disregard my comment.

PR #2874, opened 2020-05-29
[NTOS:PS] Probe the information class validity based on class table

GeoB99:
## Purpose
Bring `PsProcessInfoClass` and `PsThreadInfoClass` tables back that were removed in bf493b9 commit which should fix a number of issues with datatype misalignment. This lets us enabling the probe class validity code in `NtSetInformationProcess/Thread` functions. Further testcases for the rest of the classes have to be implement to ensure the alignment values are correct so this PR will remain a draft for the moment.

While I've fixed some assertions and other issues brought by `ps_i.h` it could be great if someone can run a testman on this to see how this patch goes.

**EDIT (the description above is obsolete):** As #2906 is now merged, it's time to enable the alignment probing checks back.


GeoB99:
Rebased the PR and enabled the alignment probing in the kernel code. Here's the current test results:

**NtSetInformationThread / NtSetInformationProcess**
Capture
**NtQueryInformationThread / NtQueryInformationProcess**
Capture2

Compared to the current results in master:
Capture3


The alignment test failures have been squashed and dropped significantly, most of these test failures you see caught by the probe library is due to unimplemented cases in kernel. Overall ReactOS works nicely without problems but I'm going to do further tests to make sure ROS doesn't explode like the Reactor 4 from Chernobyl.

GeoB99:
Ready

modules/rostests/apitests/ntdll/NtQueryInformationProcess.c
```diff
@@ -351,7 +351,7 @@ Test_ProcQueryAlignmentProbe(void)
         QuerySetProcessValidator(QUERY,
                                  InfoClass,
                                  (PVOID)(ULONG_PTR)PsProcessInfoClass[InfoClass].AlignmentQUERY,
-                                 PsProcessInfoClass[InfoClass].RequiredSizeQUERY - 1,
+                                 0,
```
zefklop, line 354: This looks like a valid test case to me. Why do you remove this ?

GeoB99: Because this test literally fails for whatever bizarre reason in ReactOS, albeit it works fine in W2k3. And as it turns out, the `RequiredSizeQUERY` value is not changed even though it should be subtracted. I dunno what's the likely cause of it so I just changed the value to 0 as it's still valid since we don't supply the correct size anyway and `STATUS_INFO_LENGTH_MISMATCH` is properly returned.

If you think you've a better suggestion or solution, I'd be glad to hear it. Thanks

zefklop: Well, that looks like a bug to me, and you remove the very test that points to it.

We must simply check sizes in NtQueryInformationProcess before probing input pointer.

ntoskrnl/include/internal/probe.h
```diff
@@ -31,7 +31,12 @@ DefaultSetInfoBufferCheck(ULONG Class,
 
         if (NT_SUCCESS(Status))
         {
-            if (PreviousMode != KernelMode)
+            /*
+             * An alignment of 1 (CHAR) means no alignment probe required.
+             * Some classes do not need alignment probing.
+             */
+            if (ClassList[Class].AlignmentSET != sizeof(CHAR) ||
+                PreviousMode != KernelMode)
```
zefklop, line 39: This is wrong. ProbeForRead fails for any address which is above highest user mode address, so any kernel mode call will fail in case Alignement != 1

zefklop (changes requested):
- [ ] Fix NtQueryInformationProcess instead of deleting valid test
- [ ] Fix KernelMode address probing.

modules/rostests/apitests/ntdll/NtQueryInformationProcess.c
zefklop: Maybe just relax the check for ICIF_QUERY_SIZE_VARIABLE & ICIF_SET_SIZE_VARIABLE in DefaultQueryInfoBufferCheck & DefaultSetInfoBufferCheck functions. I think this would make sense since those variable arrays must at least hold one value to have any effect.

GeoB99: Your suggestion actually brings more problems than it solves, also sizes are already checked in NtQueryInformationProcess and friends already. I further checked again on why the test fails with `PsProcessInfoClass[InfoClass].RequiredSizeQUERY - 1` and it seems that it fails with STATUS_ACCESS_VIOLATION within `ProbeForWrite`, on that line respectively.
https://github.com/reactos/reactos/blob/master/ntoskrnl/ex/exintrin.c#L180

This case seems to happen only in NtQueryInformationProcess and NtQueryInformationThread respectively. I'm still trying to find a solution for this but dunno if I can keep going because it's so confusing as hell.

ntoskrnl/include/internal/probe.h
HBelusca, line 39: Confirming, this breaks MS NT compatibility.

modules/rostests/apitests/ntdll/NtQueryInformationProcess.c
zefklop: Yes, because the test provides a non-zero length, which triggers the `Length != 0` path in ProbeForWrite. As the test provides an invalid buffer, this naturally fails.

The test actually shows that the buffer length gets tested *before* the buffer is probed, and that length check is skipped because of this line here: https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/include/internal/probe.h

What happens if you comment out that "if" ?

GeoB99:
Addressed the comments and the test failures regarding STATUS_INFO_LENGTH_MISMATCH. These are the current results:
**NtQueryInformationProcess/Thread**
Capture
**NtSetInformationProcess**
Capture2
**NtSetInformationThread**
Capture4

zefklop:
Now this begs for a test bot run :wink:

modules/rostests/apitests/ntdll/probelib.c
```diff
@@ -31,21 +31,31 @@ QuerySetProcessValidator(
                     break;
                 }
 
+                /*
+                 * NOTE: Buffer validation checks are relaxed on Windows Server 2003 SP2,
+                 * that is, the system only checks if the pointer is properly aligned in
+                 * NtQueryInformationProcess function and friends but it doesn't check if the
+                 * buffer is valid (that is up for each information class to care about that, whatever
+                 * that is). On ReactOS however we're way aggressive regarding this.
```
HBelusca, line 39: Isn't what you were implementing in e.g. DefaultQueryInfoBufferCheck() with this ICIF_QUERY_SIZE_VARIABLE ?
If so you could always move that check after the probing done in DefaultQueryInfoBufferCheck() ?

modules/rostests/apitests/ntdll/probelib.c
```diff
@@ -31,21 +31,31 @@ QuerySetProcessValidator(
...
+                 */
                 case ProcessHandleTracing:
                 {
+#ifdef __REACTOS__
+                    SpecialStatus = STATUS_ACCESS_VIOLATION;
+#else
                     SpecialStatus = STATUS_INVALID_PARAMETER;
+#endif
```
zefklop, line 47: Are you changing expected test results based on whether we're compiling this in Ros tree or not ?

HBelusca: looks like it does.

GeoB99: I was experimenting some stuff, apologies for having amended that into commit in advance.

modules/rostests/apitests/ntdll/probelib.c
GeoB99: Dunno what exactly you want with this but your suggestion actually breaks stuff even more. `ICIF_QUERY_SIZE_VARIABLE` means the size length of an information class is arbitrary, whereas the lengths whose sizes are fixed mustn't have this flag. As this flag and similar were badly misused in the tables, that reason alone is why the tests failed previously.

Moving this flag check after the buffer probing is done, as you suggested, the buffer is always probed first and size length is checked after and that makes the tests fail again. Btw, I removed that comment in code, so forget about it.

HBelusca: ok, thanks for the explanation.

GeoB99:
Amended the commit again and here's the final result, the only difference being:
Capture
27 equates to `ProcessImageFileName` which is the only information class whose size length is arbitrary so far. As a matter of fact STATUS_INFO_LENGTH_MISMATCH would never get triggered. For this test failure to get fixed there needs some improvements to the ICIF code thing and a few stuff to be changed in tests but I'm pretty much burned out with this already so I'll do that in a future PR (or everyone is welcome to fix it anyway).

Tomorrow I'll run a testbot against this PR to see how it goes with alignment enabled.

GeoB99:
Test bot results

Test KVM -> https://build.reactos.org/#/builders/6/builds/29178
Test Vbox -> https://build.reactos.org/#/builders/3/builds/16765

zefklop:
> Test bot results
> 
> Test KVM -> https://build.reactos.org/#/builders/6/builds/29178
> Test Vbox -> https://build.reactos.org/#/builders/3/builds/16765

A link to testman showing the changed test results would help.

GeoB99:
Re-ran a new testman instance thanks to Timo (I'm not able to log in to buildbots unfortunately). Here are the test results:

**NtQueryInformationProcess** -> https://reactos.org/testman/detail.php?id=53061561&prev=0
**NtQueryInformationThread** -> https://reactos.org/testman/detail.php?id=53061562&prev=0
**NtSetInformationProcess** -> https://reactos.org/testman/detail.php?id=53061571&prev=0
**NtSetInformationThread** -> https://reactos.org/testman/detail.php?id=53061572&prev=0

GeoB99:
I'm gonna share the comparison test results here, so that it doesn't get lost in the Development channel. @zefklop (and everyone else) may take a look at it.
https://reactos.org/testman/compare.php?ids=77464,77466

tkreuzer (approved):

PR #2880, opened 2020-05-31
[APISETS] api-ms-win-core-localization-l1-n-n: Are NT6.1++ actually. CORE-17095

SergeGautherie:
[WTB 72525](https://testbot.winehq.org/JobDetails.pl?Key=72525)
[WTB 72526](https://testbot.winehq.org/JobDetails.pl?Key=72526)
[WTB 72527](https://testbot.winehq.org/JobDetails.pl?Key=72527)
+
https://docs.microsoft.com/en-us/windows/win32/apiindex/windows-8-api-sets
http://www.geoffchappell.com/studies/windows/win32/apisetschema/history/sets62.htm?tx=3

JIRA issue: [CORE-17095](https://jira.reactos.org/browse/CORE-17095)


SergeGautherie:
According to
https://docs.microsoft.com/en-us/windows/win32/apiindex/windows-8-api-sets
v1.2.0 would be NT6.2. (Which would seem consistent.)
Would you like me to make that change?

PR #2853, opened 2020-05-25
[NTOSKRNL] Local Inter-Process Communication (LPC) improvements

belliash:
## Purpose

Several changes, fixes and improvements to the Local Inter-Process Communication (LPC) facility provided by the kernel. This as well includes some missing parts of LPC implementation and makes it more complete.

JIRA issue: [CORE-17075](https://jira.reactos.org/browse/CORE-17075)

## Proposed changes
- Add more debugging messages,
- Add more sanity checks (eg. to make sure reply message is on the reply chain),
- Implement and make use of LpcWaitablePortObjectType,
- Implement NtReplyWaitReplyPort(),
- Implement NtQueryInformationPort(),
- Implement callbacks support,
- Implement Thread Semaphore support,
- Unify the way of checking message type by using LpcpGetMessageType(),
- Several small fixes and improvements

PR #2881, opened 2020-06-01
[HAL] Implement HalProcessorIdle() for MP

belliash:
Just a quick implementation. This function should do almost same work as its UP equivalent (sti + halt). However, on MP, processor should not be stopped as that might cause SNOOP to slow down.

### DIFF
--- a/hal/halx86/mp/processor_mp.c
+++ b/hal/halx86/mp/processor_mp.c
@@ -134,7 +134,8 @@ VOID
 NTAPI
 HalProcessorIdle(VOID)
 {
-    UNIMPLEMENTED;
+    /* Enable interrupts */
+    _enable();
 }
 
 /* EOF */


ThFabba:
Can you explain the SNOOP slowdown comment in more detail? As far as I can tell, all versions of `HalProcessorIdle` also do a `HLT`.
Thanks.

belliash:
This is a part of cache coherency and depends on architecture. Snooping-based cache coherence seems to hurt when CPU gets halted as a side-effect. For more details its best to read about snooping in C1 state, especially how it can temporarily leave the C1 (HALT) state to deal with requests coming from the CPU external bus.

I'm not quite sure if HAL should force CPU to halt in MP envs. I believe not.

ColinFinck:
Closing this PR due to violation of our contribution guidelines (https://github.com/reactos/reactos/pull/2853#issuecomment-650745493)

PR #2882, opened 2020-06-01
[OSK] Updated german on-screen keyboard translation

FritzsHero:
## Purpose

Updated german on-screen keyboard translation. Synchronized translation of "Enhanced keyboard" with "Standard keyboard".


binarymaster:
Cc @reactos/lang-german

base/applications/osk/lang/de-DE.rc
```diff
@@ -238,69 +238,69 @@ BEGIN
...
     CONTROL         "",IDC_LED_SCROLL,"Static",SS_CENTER|SS_NOTIFY,424,15,4,3
 END
 
 IDR_OSK_MENU MENU DISCARDABLE
 BEGIN
-    POPUP "File"
+    POPUP "Datei"
     BEGIN
-        MENUITEM "&Exit", IDM_EXIT
+        MENUITEM "&Beenden", IDM_EXIT
```
gonzoMD, line 254: Put the Ampersand after the B?

base/applications/osk/lang/de-DE.rc
```diff
@@ -238,69 +238,69 @@ BEGIN
...
 STYLE DS_SHELLFONT | DS_MODALFRAME | DS_CENTER | WS_POPUP | WS_CAPTION
-CAPTION "On-Screen Keyboard"
+CAPTION "Bildschirmtastatur"
 FONT 8, "MS Shell Dlg", 400, 0, 0x1
 BEGIN
     DEFPUSHBUTTON "OK", IDOK, 193, 76, 50, 14
     ICON IDI_OSK, IDC_STATIC, 7, 17, 20, 20
-    LTEXT "On-Screen Keyboard provides a minimum level of functionality for mobility-impaired users. Mobility-impaired users will need a utility program with higher functionality for daily use.", IDC_STATIC, 36, 7, 207, 33
-    CONTROL "Do not show this message again", IDC_SHOWWARNINGCHECK, "Button",
+    LTEXT "Bildschirmtastatur stellt ein mindestmaß an Funktionalität für verhinderte Nutzer zur Verfügung. Solche Benutzer werden ein Program mit größerem Funktionsumfang brauchen für den täglichen Gebrauch.", IDC_STATIC, 36, 7, 207, 33
```
gonzoMD, line 296: Eingeschränkt instead of verhindert.

Diese Nutzer werden, für den täglichen Gebrauch,... then the rest of the sentence

base/applications/osk/lang/de-DE.rc
```diff
@@ -238,69 +238,69 @@ BEGIN
...
     END
 
-    POPUP "Settings"
+    POPUP "Einstellungen"
     BEGIN
-        MENUITEM "Always on Top", IDM_ON_TOP, CHECKED
+        MENUITEM "Immer im Vordergrund", IDM_ON_TOP, CHECKED
         MENUITEM SEPARATOR
-        MENUITEM "&Use Click Sound", IDM_CLICK_SOUND
+        MENUITEM "&Benutze Klick-ton", IDM_CLICK_SOUND
```
gonzoMD, line 274: Klick-Ton ben&utzen

base/applications/osk/lang/de-DE.rc
```diff
@@ -238,69 +238,69 @@ BEGIN
...
-CAPTION "On-Screen Keyboard"
+CAPTION "Bildschirmtastatur"
 FONT 8, "MS Shell Dlg", 400, 0, 0x1
 BEGIN
     DEFPUSHBUTTON "OK", IDOK, 193, 76, 50, 14
     ICON IDI_OSK, IDC_STATIC, 7, 17, 20, 20
-    LTEXT "On-Screen Keyboard provides a minimum level of functionality for mobility-impaired users. Mobility-impaired users will need a utility program with higher functionality for daily use.", IDC_STATIC, 36, 7, 207, 33
-    CONTROL "Do not show this message again", IDC_SHOWWARNINGCHECK, "Button",
+    LTEXT "Bildschirmtastatur stellt ein mindestmaß an Funktionalität für verhinderte Nutzer zur Verfügung. Solche Benutzer werden ein Program mit größerem Funktionsumfang brauchen für den täglichen Gebrauch.", IDC_STATIC, 36, 7, 207, 33
+    CONTROL "Zeige diese Nachricht nicht mehr", IDC_SHOWWARNINGCHECK, "Button",
```
gonzoMD, line 297: Diese Nachricht nicht mehr anzeigen.

base/applications/osk/lang/de-DE.rc
```diff
@@ -162,67 +162,67 @@ BEGIN
...
-    PUSHBUTTON      "ret",SCAN_CODE_43,262,60,38,15,BS_ICON
-    PUSHBUTTON      "shift",SCAN_CODE_44,3,79,57,15,BS_ICON|BS_PUSHLIKE|BS_AUTOCHECKBOX
-    PUSHBUTTON      "z",SCAN_CODE_46,64,79,15,15
-    PUSHBUTTON      "x",SCAN_CODE_47,83,79,15,15
-    PUSHBUTTON      "c",SCAN_CODE_48,102,79,15,15
-    PUSHBUTTON      "v",SCAN_CODE_49,121,79,15,15
-    PUSHBUTTON      "b",SCAN_CODE_50,140,79,15,15
-    PUSHBUTTON      "n",SCAN_CODE_51,159,79,15,15
-    PUSHBUTTON      "m",SCAN_CODE_52,178,79,15,15
+    PUSHBUTTON      "Q",SCAN_CODE_17,42,41,15,15
```
binarymaster, line 169: Why did you changed letters case? It does not belong to translation, but if you want to change, it should be done in a separate PR.

base/applications/osk/lang/de-DE.rc
FritzsHero: I agree with you. It would probably make sense to keep the shortcuts as close as possible to the "original" english translation. We then should also consider to unify this in all application with a german translation. In "taskmgr" and "dxdiag" the translation also has &Beenden instead of B&eenden . Maybe this should be updated too.

base/applications/osk/lang/de-DE.rc
FritzsHero: Excellent question!
Actually the "Enhanced keyboard" was already translated with upper cases, while "standard keyboard" was not translated and used still lower cases. Now they both use upper case letters, which, to me, look more appealing and clean.
However when we look at the english resource file we see that both keyboard versions over there use lower case letters instead of upper case letters.

I can revert this. Can I force push on my branch or do you prefer another commit?

We should however decide if we either want to use lower or upper case letters across all translations, to keep things unified. With that we can better ensure that the user experience and application behavior across all translations stays the same. Maybe we should open a discussion issue about that?

Fraizeraust: Revert this change to keep the originality with the English resource of OSK (and if you want for the other resources but in a next PR please). In the future when I plan and have time (I can't promise anything) to re-create On-Screen Keyboard it'll no longer use resources and the upper/lower case conversion will be done programmatically.

PR #2883, opened 2020-06-01
[KERNEL32] FatalAppExitW: Add an 'UNREFERENCED_LOCAL_VARIABLE(Status)'

SergeGautherie:
Addendum to 21d2accad37825f47e3caa78ff36153751652238 (r71793).

### DIFF
--- a/dll/win32/kernel32/client/proc.c
+++ b/dll/win32/kernel32/client/proc.c
@@ -1609,6 +1609,8 @@ FatalAppExitW(IN UINT uAction,
 #if DBG
     /* Give the user a chance to abort */
     if ((NT_SUCCESS(Status)) && (Response == ResponseCancel)) return;
+#else
+    UNREFERENCED_LOCAL_VARIABLE(Status);
 #endif
 
     /* Otherwise kill the process */


dll/win32/kernel32/client/proc.c
```diff
@@ -1588,13 +1588,18 @@ FatalAppExitW(IN UINT uAction,
...
     NTSTATUS Status;
+#endif
 
     /* Setup the string to print out */
     RtlInitUnicodeString(&UnicodeString, lpMessageText);
 
     /* Display the hard error no matter what */
-    Status = NtRaiseHardError(STATUS_FATAL_APP_EXIT | HARDERROR_OVERRIDE_ERRORMODE,
+#if DBG
+    Status =
```
ThFabba, line 1600: `UNREFERENCED_LOCAL_VARIABLE(Status)` seems nicer.

HBelusca: or DBG_UNREFERENCED_LOCAL_VARIABLE?

ThFabba: DBG_UNREFERENCED_LOCAL_VARIABLE means the code referencing the variable is not implemented and it will be fixed later - it has no relation to the DBG macro.

SergeGautherie: > `UNREFERENCED_LOCAL_VARIABLE(Status)` seems nicer.

I feel that would be an abuse of that macro and I would need to duplicate the `NtRaiseHardError()` call,
so I think my proposal is clearer.

ThFabba: Duplicate? Not sure I follow. Here's what I'm thinking:
```
NTSTATUS Status;
Status = NtRaiseHardError(...)
#if DBG
if (!NT_SUCCESS(Status)) { ... }
#else
UNREFERENCED_LOCAL_VARIABLE(Status);
#endif
```

SergeGautherie: Yeah, I guessed that and I feel it is an abuse of that macro. Anyway.
Changed. As you prefer.

PR #2887, opened 2020-06-02
[LOG2LINES] Fix 'getFmt()'

SergeGautherie:
This fixes 'd' case.

Detected by Cppcheck: uselessAssignmentPtrArg.

### DIFF
--- a/sdk/tools/log2lines/util.c
+++ b/sdk/tools/log2lines/util.c
@@ -117,6 +117,7 @@ basename(char *path)
     return path;
 }
 
+static
 const char *
 getFmt(const char *a)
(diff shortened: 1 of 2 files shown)


SergeGautherie:
As there was no issue filed during 10+ years, I suggest to drop explicit `'d'` case and simply use `%i` instead.
What do you think?

learn-more:
so to fix a very severe *ahum* cppcheck issue, you also refactor the function?
This is the reason why your PR's stay open this long, since there is always the case that we have to ask to revert 3 other things unrelated to what you are 'fixing'.

PR #2890, opened 2020-06-02
[CMAKE] MSVC RUNTIME_CHECKS is a 'Debug'-only feature

SergeGautherie:
"cl : Command line error D8016 : '/Ox' and '/RTC1' command-line options are incompatible"

Addendum to 92dfec219d8c53a84c68ca069abbc170fc8bdb49.

### DIFF
--- a/sdk/cmake/config-amd64.cmake
+++ b/sdk/cmake/config-amd64.cmake
@@ -44,5 +44,10 @@ set(USERMODE TRUE CACHE BOOL
 if(MSVC)
 set(_PREFAST_ FALSE CACHE BOOL
 "Whether to enable PREFAST while compiling.")
-    option(RUNTIME_CHECKS "Whether to enable runtime checks on MSVC" ON)
+    if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+        option(RUNTIME_CHECKS "Whether to enable runtime checks on MSVC" ON)
+    else()
+        # RTC are incompatible with compiler optimizations.
+        set(RUNTIME_CHECKS FALSE CACHE BOOL "Whether to enable runtime checks on MSVC")
+    endif()
 endif()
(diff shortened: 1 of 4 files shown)


sdk/cmake/config-amd64.cmake
```diff
@@ -44,5 +44,10 @@ set(USERMODE TRUE CACHE BOOL
 if(MSVC)
 set(_PREFAST_ FALSE CACHE BOOL
 "Whether to enable PREFAST while compiling.")
-    option(RUNTIME_CHECKS "Whether to enable runtime checks on MSVC" ON)
+    if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+        option(RUNTIME_CHECKS "Whether to enable runtime checks on MSVC" ON)
+    else()
+        # RTC are incompatible with compiler optimizations.
+        set(RUNTIME_CHECKS FALSE CACHE BOOL "Whether to enable runtime checks on MSVC")
```
ThFabba, line 51: Why does this not use `option` like the regular case?

SergeGautherie: I felt like `set()` would be a bit more explicit about never changing the value.
If you prefer `option()`, that should work fine too.

ThFabba: Sounds like a perfect use case for https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html#module:CMakeDependentOption

SergeGautherie: Nice suggestion.
Done.

SergeGautherie:
Cc @Extravert-ir, as regression author.

Extravert-ir:
Merged a2a6038e562e1936d737bac6cb623cbd74b0f1e2

PR #2891, opened 2020-06-04
[POWRPROF_APITEST] Fix 2+1 Cppcheck reports

SergeGautherie:
Note that both tests are currently skipped...

### DIFF
--- a/modules/rostests/apitests/powrprof/power.c
+++ b/modules/rostests/apitests/powrprof/power.c
@@ -60,7 +60,7 @@ void test_CallNtPowerInformation(void)
    /* LastSleepTime tests */
    retval = CallNtPowerInformation(LastSleepTime, 0, 0, 0, 0);
    ok(retval == STATUS_SUCCESS, "function result wrong expected STATUS_SUCCESS but got %d\n", (UINT)retval);
-   retval = CallNtPowerInformation(LastSleepTime, &atime, sizeof(sizeof(ULONGLONG)), 0, 0);
+   retval = CallNtPowerInformation(LastSleepTime, &atime, sizeof(ULONGLONG), 0, 0);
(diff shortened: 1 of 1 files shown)


modules/rostests/apitests/powrprof/power.c
```diff
@@ -4620,7 +4620,8 @@ pPP_original->user.OptimizeForPowerDc=1;
    ret = ValidatePowerPolicies(&gpp,&pp);
    ok(ret, "function was expected to succeed error %i\n",(UINT)GetLastError());
    ok(globalcompare(gpp,gpp_compare),"Difference Found\n");
-   pp_compare.user.MaxSleepAc=4;
+   // FIXME: Duplicate line.
+   // pp_compare.user.MaxSleepAc=4;
```
binarymaster, line 4624: Considering the code above where variable is initialized, you can safely delete this line.

SergeGautherie: Deleted.

modules/rostests/apitests/powrprof/power.c
```diff
@@ -60,7 +60,7 @@ void test_CallNtPowerInformation(void)
    /* LastSleepTime tests */
    retval = CallNtPowerInformation(LastSleepTime, 0, 0, 0, 0);
    ok(retval == STATUS_SUCCESS, "function result wrong expected STATUS_SUCCESS but got %d\n", (UINT)retval);
-   retval = CallNtPowerInformation(LastSleepTime, &atime, sizeof(sizeof(ULONGLONG)), 0, 0);
+   retval = CallNtPowerInformation(LastSleepTime, &atime, sizeof(atime), 0, 0);
```
ThFabba, line 63: I agree that `sizeof(variable)` is generally preferable to `sizeof(TYPE)` but it's still a style choice, and the #1 rule of style is to be consistent, so please use `sizeof(ULONGLONG)`. Then this becomes a trivial fix.

SergeGautherie: Reverted.

PR #2892, opened 2020-06-05
[CRT] atexit.c: Fix memory handling

SergeGautherie:
And related improvements.

Addendum to b30bf6de315557f80cfa16574689563dfe08eab5 (r54651).

### DIFF
--- a/sdk/lib/crt/stdlib/atexit.c
+++ b/sdk/lib/crt/stdlib/atexit.c
@@ -69,7 +69,7 @@ _onexit_t CDECL _onexit(_onexit_t func)
   {
     _onexit_t *newtable;
     TRACE("expanding table\n");
-    newtable = calloc(atexit_table_size + 32, sizeof(void *));
+    newtable = calloc(atexit_table_size + 32, sizeof(_onexit_t));
     if (!newtable)
     {
       TRACE("failed!\n");
--- a/sdk/lib/crt/stdlib/atexit.c
+++ b/sdk/lib/crt/stdlib/atexit.c
@@ -76,7 +76,7 @@ _onexit_t CDECL _onexit(_onexit_t func)
       UNLOCK_EXIT;
       return NULL;
     }
-    memcpy (newtable, atexit_table, atexit_table_size);
+    memcpy (newtable, atexit_table, atexit_table_size*sizeof(_onexit_t));
     atexit_table_size += 32;
     free (atexit_table);
     atexit_table = newtable;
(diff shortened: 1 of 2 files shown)


learn-more:
What does it 'fix' other than breaking the previously zero-initialized table?

SergeGautherie:
> What does it 'fix' other than breaking the previously zero-initialized table?

How does it break?
Especially, what good does zeroing do? As the first part is overwritten and the second part is not accessed...

sdk/lib/crt/stdlib/atexit.c
```diff
@@ -69,16 +79,14 @@ _onexit_t CDECL _onexit(_onexit_t func)
   {
     _onexit_t *newtable;
     TRACE("expanding table\n");
-    newtable = calloc(atexit_table_size + 32, sizeof(void *));
+    newtable = realloc(atexit_table, (atexit_table_size + 32) * sizeof(_onexit_t));
```
ThFabba, line 82: Missing zero-init

SergeGautherie: Again, what is zero-init needed for?

ThFabba: My mistake, didn't see there's `atexit_registered` in addition to `atexit_table_size`.

ThFabba:
Your branch name mentions cppcheck -- so is there an actual cppcheck warning you're fixing here? If so please quote/explain it. If not, this seems entirely unnecessary.

SergeGautherie:
> Wine sync

Wine code is different, though I could try to upstream bits that would apply there...
https://source.winehq.org/git/wine.git/blob/17529582402ebe27ef975fc7dcb8353f4f95e629:/dlls/msvcrt/exit.c#l65

> actual cppcheck warning

* https://qarmin.gitlab.io/-/reactos/-/jobs/577525577/artifacts/report/1011.html#line-72
* '[CRT] _onexit(): Use the explicit '_onexit_t' type' commit.

ThFabba:
> Wine code is different, though I could try to upstream bits that would apply there...
> https://source.winehq.org/git/wine.git/blob/17529582402ebe27ef975fc7dcb8353f4f95e629:/dlls/msvcrt/exit.c#l65

Wine code has simply changed over time.
https://source.winehq.org/git/wine.git/commitdiff/83f8b226948d40a81917a649b7ef83e2ea46a845


>     * https://qarmin.gitlab.io/-/reactos/-/jobs/577525577/artifacts/report/1011.html#line-72
>     * '[CRT] _onexit(): Use the explicit '_onexit_t' type' commit.

So cppcheck is smart enough to know that `newtable` is pointer-sized but not that `*newtable` is pointer-sized? That sounds like a flat-out bug in their analyzer.
In any case, it may be worth silencing the warning... for which that single commit looks just fine. Doesn't warrant a complete refactor of upstream code.

SergeGautherie:
> code has simply changed

Thanks for the link. I'll look at that separately.

> just fix the bug

Removed 2 refactor commits.

PR #2894, opened 2020-06-05
[RAPPS] Screenshot preview and other trivial fixes

kernelbin:
## Purpose

this PR is mainly for adding a window for snapshot preview. (In the future, people can click this window to view other snapshots)

currently, it only supports showing an image from a file ( path_to_rapps\rapps\rapps\snapshots\ )

and some other fixes are done in this PR too
- always use rappmgr.cab as the filename when downloading DB (otherwise may cause bug when downloading from a specified URL)
- use PathXXX API to operate path
- delete some unused/obsolete var

## TODO
 - snapshot URL support


base/applications/rapps/gui.cpp
```diff
@@ -21,11 +21,28 @@
...
+using namespace Gdiplus;
 
 #define SEARCH_TIMER_ID 'SR'
 #define LISTVIEW_ICON_SIZE 24
 #define TREEVIEW_ICON_SIZE 24
 
+#define SNPSHTPREV_EMPTY    0 // show nothing
+#define SNPSHTPREV_LOADING  1 // image is loading (most likely downloading)
+#define SNPSHTPREV_FILE     2 // display image from a file
+#define SNPSHTPREV_FAILED   3 // image can not be shown (download failure or wrong image)
```
learn-more, line 36: Is there any reason not to use an enum?

base/applications/rapps/gui.cpp
```diff
@@ -257,6 +274,250 @@ class CAppRichEdit :
...
+        break;
+        }
+    }
+
+    float GetLoadingDotWidth(int width, int height)
+    {
+        return min(width, height) / 20.0;
+    }
+
+    float GetFrameDotShift(int Frame, int width, int height)
```
learn-more, line 415: I don't understand this function name?

base/applications/rapps/include/available.h
```diff
@@ -48,17 +64,17 @@ struct CAvailableApplicationInfo
     ATL::CStringW m_szInstalledVersion;
 
     // Create an object from file
-    CAvailableApplicationInfo(const ATL::CStringW& sFileNameParam);
+    CAvailableApplicationInfo(const ATL::CStringW& sFileNameParam, AvailableStrings m_Strings);
 
     // Load all info from the file
-    VOID RefreshAppInfo();
+    VOID RefreshAppInfo(AvailableStrings m_Strings);
```
learn-more, line 70:
```suggestion
    VOID RefreshAppInfo(AvailableStrings& m_Strings);
```
You probably want to use a reference here, otherwise you create a copy of the object.

base/applications/rapps/include/available.h
```diff
@@ -48,17 +64,17 @@ struct CAvailableApplicationInfo
     ATL::CStringW m_szInstalledVersion;
 
     // Create an object from file
-    CAvailableApplicationInfo(const ATL::CStringW& sFileNameParam);
+    CAvailableApplicationInfo(const ATL::CStringW& sFileNameParam, AvailableStrings m_Strings);
```
learn-more, line 67:
```suggestion
    CAvailableApplicationInfo(const ATL::CStringW& sFileNameParam, AvailableStrings& m_Strings);
```

base/applications/rapps/include/available.h
```diff
@@ -71,7 +87,7 @@ struct CAvailableApplicationInfo
     inline BOOL GetString(LPCWSTR lpKeyName, ATL::CStringW& ReturnedString);
 
     // Lazily load general info from the file
-    VOID RetrieveGeneralInfo();
+    VOID RetrieveGeneralInfo(AvailableStrings m_Strings);
```
learn-more, line 90:
```suggestion
    VOID RetrieveGeneralInfo(AvailableStrings& m_Strings);
```

base/applications/rapps/winmain.cpp
```diff
@@ -13,6 +13,8 @@
 
 #include <atlcom.h>
 
+#include<gdiplus.h>
```
learn-more, line 16:
```suggestion
#include <gdiplus.h>
```

base/applications/rapps/include/available.h
```diff
@@ -35,8 +50,9 @@ struct CAvailableApplicationInfo
     ATL::CStringW m_szSize;
     ATL::CStringW m_szUrlSite;
     ATL::CStringW m_szUrlDownload;
-    ATL::CStringW m_szCDPath;
     ATL::CSimpleArray<LCID> m_LanguageLCIDs;
+    ATL::CStringW m_szSnapshotFilename[MAX_SNAPSHOT_NUM];
```
learn-more, line 54: Why not use an `ATL:CSimpleArray` instead of a fixed-size array?

base/applications/rapps/gui.cpp
```diff
@@ -304,9 +569,17 @@ class CAppInfoDisplay :
         return FALSE;
     }
 
+    VOID ResizeChildren(int Width, int Height)
+    {
+        int SnpshtWidth = SnpshtPrev->GetRequestedWidth(Height);
+        ::MoveWindow(SnpshtPrev->m_hWnd, 0, 0, SnpshtWidth, Height, TRUE);
+        ::MoveWindow(RichEdit->m_hWnd, SnpshtWidth, 0, Width - SnpshtWidth, Height, TRUE);
```
learn-more, line 576: When resizing multiple child windows, it's better to use `BeginDeferWindowPos`, `DeferWindowPos` and `EndDeferWindowPos`.
This will move all windows at once.
See for an example MainWnd_OnSize from fontview:
https://git.reactos.org/?p=reactos.git;a=blob;f=base/applications/fontview/fontview.c;hb=5d88487cd18058baad1199277777e148c8a1081a#l402

base/applications/rapps/available.cpp
```diff
@@ -18,23 +18,23 @@
 #include <atlstr.h>
 
  // CAvailableApplicationInfo
-CAvailableApplicationInfo::CAvailableApplicationInfo(const ATL::CStringW& sFileNameParam)
+CAvailableApplicationInfo::CAvailableApplicationInfo(const ATL::CStringW& sFileNameParam, AvailableStrings m_Strings)
```
learn-more, line 21: It's better to not name a function argument like we name our member variables (with the `m_` prefix)

base/applications/rapps/available.cpp
```diff
@@ -231,11 +261,20 @@ AvailableStrings::AvailableStrings()
     //FIXME: maybe provide a fallback?
     if (GetStorageDirectory(szPath))
     {
-        szAppsPath = szPath + L"\\rapps\\";
+        szAppsPath = szPath;
+        PathAppendW(szAppsPath.GetBuffer(MAX_PATH), L"rapps");
+        PathAddBackslashW(szAppsPath.GetBuffer());
```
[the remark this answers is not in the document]
kernelbin: It's because some other places are still using the original way to handle path, without this backslash will lead to some bugs. I haven't change them all, so I add this temporary.

base/applications/rapps/CMakeLists.txt
```diff
@@ -39,7 +39,7 @@ add_rc_deps(rapps.rc ${rapps_rc_deps})
 add_executable(rapps ${SOURCE} rapps.rc)
 set_module_type(rapps win32gui UNICODE)
 target_link_libraries(rapps uuid wine)
-add_importlibs(rapps advapi32 comctl32 gdi32 wininet user32 shell32 shlwapi ole32 setupapi msvcrt kernel32 ntdll)
+add_importlibs(rapps advapi32 comctl32 gdi32 wininet user32 shell32 shlwapi ole32 setupapi msvcrt kernel32 ntdll gdiplus)
```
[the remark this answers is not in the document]
kernelbin: OK, I've changed it. but... may I ask why? 🤣

base/applications/rapps/available.cpp
```diff
@@ -209,6 +225,20 @@ BOOL CAvailableApplicationInfo::HasUpdate() const
     return (m_szInstalledVersion.Compare(m_szVersion) < 0) ? TRUE : FALSE;
 }
 
+BOOL CAvailableApplicationInfo::RetrieveSnapshot(int Index,ATL::CStringW& SnapshotFileName) const
+{
+    if (Index >= MAX_SNAPSHOT_NUM)
```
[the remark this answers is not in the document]
kernelbin: now I changed it to UINT, then there's no need to check ```< 0```

base/applications/rapps/CMakeLists.txt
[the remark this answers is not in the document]
learn-more: Thomas explained this nicely in mattermost:

> The order is important because if ntdll comes first it'll satisfy all the CRT functions, so you're linking against the NT CRT instead of the Win32 one
And because msvcrt includes msvcrtex, which may need stuff from kernel32 and ntdll
Having them last... Hmm. It actually breaks something (at link time) if you don't, but I don't remember what
(but not in all modules)
But the underlying concept is: linking has an order, and you add new things on the left, not the right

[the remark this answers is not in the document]
kernelbin: now I see ... thanks!!!

base/applications/rapps/gui.cpp
kernelbin: ok, now I improved this.

base/applications/rapps/include/available.h
kernelbin: I think it's almost certain that we are going to limit the maximum number of snapshots. Displaying an infinite number of screenshots makes no sense. As 16 (MAX_SNAPSHOT_NUM) is not big, I think it's ok to use a fixed-size array.

why 16? I don't know 😝 I think it's enough to store snapshots of a normal software and As a programmer, it looks more like a round number than 10 or 20, so I chose it. Anyway, it's a macro, we can change it at any time.

kernelbin:
> This is certainly making good progress 👍
> 
> Some minor additions, however for the array of snapshots: Maintenance / handling / range checking on them is going to be generally easier when you store them in one of the atl collection classes instead of in a fixed array. (They are available, we are writing c++, so why not use those?)

yes, it makes sense. now I've already improved it.

base/applications/rapps/gui.cpp
```diff
@@ -304,9 +655,95 @@ class CAppInfoDisplay :
...
+        else
+        {
+            dwError = GetLastError();
+        }
+
+
+        if (dwError != ERROR_SUCCESS)
+        {
+            // TODO: error handling (I have no idea how to handle)
+        }
```
learn-more, line 711:
```suggestion
#if DBG
        ATLASSERT(dwError == ERROR_SUCCESS);
#endif
```

PR #2895, opened 2020-06-05
[KERNEL32][RTL] Improve _module_name_from_addr() and PrintStackTrace()

SergeGautherie:


ThFabba:
The bytes vs chars thing... not an issue, because the array in question is an array of chars.
The MAXDWORD thing... is dead code because this function is only ever called with the size of a static (128 byte) buffer.

So I gather there's no actual bug fix hiding in here somewhere?

SergeGautherie:
> not an issue

I know, yet the code is wrong/confused.

> dead code

I know (as it is currently), at least the existing "fix" is unmatched and would uselessly disable the feature.
Then, there should be a better fix... Or use an `ASSERT()` instead?

> no actual bug fix

As code is currently, yes, you can call it "refactor".

learn-more:
the code is not wrong/confused, and I would not call this an 'improvement'

PR #2896, opened 2020-06-05
[NTOSKRNL] Add 'DECLSPEC_NORETURN' to a few 'KeBugCheck*()'

SergeGautherie:


boot/environ/lib/misc/rtlcompat.c
```diff
@@ -102,6 +102,7 @@ DbgPrint (
     return 0;
 }
 
+// FIXME: DECLSPEC_NORETURN
```
HBelusca, line 105: Why is it FIXME? Does the function sometimes return?

HBelusca: EDIT: Oh ok that's not the kernel one.

PR #2897, opened 2020-06-06
[GDI32_APITEST] GetDIBits.c: Properly handle 'malloc()/free()' and re…

SergeGautherie:
…lated

Detected by Cppcheck: memleak.

### DIFF
--- a/modules/rostests/apitests/gdi32/GetDIBits.c
+++ b/modules/rostests/apitests/gdi32/GetDIBits.c
@@ -19,8 +19,22 @@ Test_GetDIBits_xBpp(
     ULONG cjSizeImage, cColors;
     HDC hdc;
 
-    hdc = GetDC(0);
+    hdc = GetDC(NULL);
+    ok(hdc != NULL, "GetDC failed\n");
+    if (hdc == NULL)
+    {
+        skip("No hdc\n");
+        return;
+    }
+
     hbmp = CreateBitmap(3, 3, 1, cBitsPixel, NULL);
+    ok(hbmp != NULL, "CreateBitmap failed\n");
+    if (hbmp == NULL)
+    {
+        skip("No hbmp\n");
+        ok(ReleaseDC(NULL, hdc) == 1, "ReleaseDC failed\n");
+        return;
+    }
 
     /* Fill in the size field */
     ZeroMemory(pbmi, sizeof(ajBuffer));
@@ -66,7 +80,6 @@ Test_GetDIBits_xBpp(
     ok_int(pbmi->bmiHeader.biClrUsed, cColors);
     ok_int(pbmi->bmiHeader.biClrImportant, cColors);
 
-
 #if 0
     /* Get info including the color table */
     ZeroMemory(pbmi, sizeof(ajBuffer));
@@ -90,11 +103,11 @@ Test_GetDIBits_xBpp(
     ok_int(GetDIBits(hdc, hbmp, 0, 4, pvBits, pbmi, DIB_RGB_COLORS), 3);
 #endif
 
-    DeleteObject(hbmp);
-    ReleaseDC(NULL, hdc);
+    ok(DeleteObject(hbmp), "DeleteObject failed\n");
+    ok(ReleaseDC(NULL, hdc) == 1, "ReleaseDC failed\n");
 }
 
-void Test_GetDIBits()
+void Test_GetDIBits(void)
 {
     HDC hdcScreen, hdcMem;
     HBITMAP hbmp;
@@ -109,15 +122,31 @@ void Test_GetDIBits()
 
     bisize = sizeof(BITMAPV5HEADER) + 256 * sizeof(DWORD);
     pbi = malloc(bisize);
+    ok(pbi != NULL, "malloc failed\n");
+    if (pbi == NULL)
+    {
+        skip("No pbi\n");
+        return;
+    }
+
     pbch = (PVOID)pbi;
     pbV5Header = (PVOID)pbi;
 
     hdcScreen = GetDC(NULL);
-    ok(hdcScreen != 0, "GetDC failed, skipping tests\n");
-    if (hdcScreen == NULL) return;
+    ok(hdcScreen != NULL, "GetDC failed\n");
+    if (hdcScreen == NULL)
+    {
+        skip("No hdcScreen\n");
+        free(pbi);
+        return;
+    }
 
     hbmp = CreateCompatibleBitmap(hdcScreen, 16, 16);
     ok(hbmp != NULL, "CreateCompatibleBitmap failed\n");
+    if (hbmp == NULL)
+    {
+        goto noBmp1;
+    }
 
     /* misc */
     SetLastError(ERROR_SUCCESS);
@@ -135,8 +164,6 @@ void Test_GetDIBits()
     SetLastError(ERROR_SUCCESS);
     ok(GetDIBits((HDC)2345, hbmp, 0, 15, NULL, pbi, 0) == 0, "\n");
     ok_err(ERROR_INVALID_PARAMETER);
-
-
 
     /* null hdc */
     SetLastError(ERROR_SUCCESS);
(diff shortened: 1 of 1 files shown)


modules/rostests/apitests/gdi32/GetDIBits.c
```diff
@@ -19,8 +19,22 @@ Test_GetDIBits_xBpp(
     ULONG cjSizeImage, cColors;
     HDC hdc;
 
-    hdc = GetDC(0);
+    hdc = GetDC(NULL);
+    ok(hdc != NULL, "GetDC failed\n");
```
binarymaster, line 23: Handles are not pointers, I'm not sure it makes sense to use `NULL`.

HBelusca: Yes actually it does make sense. Handles are similar to pointers (have the same bit size, etc.)

modules/rostests/apitests/gdi32/GetDIBits.c
```diff
@@ -19,8 +19,22 @@ Test_GetDIBits_xBpp(
...
+        skip("No hdc\n");
+        return;
+    }
+
     hbmp = CreateBitmap(3, 3, 1, cBitsPixel, NULL);
+    ok(hbmp != NULL, "CreateBitmap failed\n");
+    if (hbmp == NULL)
+    {
+        skip("No hbmp\n");
+        ok(ReleaseDC(NULL, hdc) == 1, "ReleaseDC failed\n");
```
HBelusca, line 35: ReleaseDC (and DeleteObject, see below) are functions, while ok() is a macro. You should never call a function within a parameter of a macro, because nothing guarantees that the macro won't evaluate more than one time its parameters, and thus, here, calling the function more than once.
You should therefore call the function and retrieve its result, and then use the ok() macro using only the result.

Separately from this general remark, I'm not sure it makes real sense to check whether the release/delete of DCs (or other **valod** objects) has succeeded, since releasing/deleting a valid object must succeed (as well, a specific test for these Create/Select/Release/Delete functions where these checks are done, should already exist, so we should know whether the functions in question behave correctly, already).

modules/rostests/apitests/gdi32/GetDIBits.c
```diff
@@ -90,11 +103,11 @@ Test_GetDIBits_xBpp(
     ok_int(GetDIBits(hdc, hbmp, 0, 4, pvBits, pbmi, DIB_RGB_COLORS), 3);
 #endif
 
-    DeleteObject(hbmp);
-    ReleaseDC(NULL, hdc);
+    ok(DeleteObject(hbmp), "DeleteObject failed\n");
+    ok(ReleaseDC(NULL, hdc) == 1, "ReleaseDC failed\n");
```
HBelusca, line 107: See comment above.

modules/rostests/apitests/gdi32/GetDIBits.c
```diff
@@ -239,11 +266,18 @@ void Test_GetDIBits()
     pbi->bmiHeader.biSize = sizeof(BITMAPINFOHEADER);
     ok_int(GetDIBits(hdcScreen, hbmp, 234, 43, NULL, pbi, DIB_RGB_COLORS), 1);
 
-    DeleteObject(hbmp);
+    ok(DeleteObject(hbmp), "DeleteObject failed\n");
```
HBelusca, line 269: See comment above.

modules/rostests/apitests/gdi32/GetDIBits.c
```diff
@@ -431,12 +473,21 @@ void Test_GetDIBits()
     pbi->bmiHeader.biBitCount = 5;
     ok_int(GetDIBits(hdcScreen, hbmp, 0, 4, pvBits, pbi, DIB_RGB_COLORS), 0);
 
-    DeleteObject(hbmp);
     HeapFree(GetProcessHeap(), 0, pvBits);
 
+noBits1:
+    ok(DeleteObject(hbmp), "DeleteObject failed\n");
```
HBelusca, line 479: See comment above.

modules/rostests/apitests/gdi32/GetDIBits.c
```diff
@@ -452,34 +503,56 @@ void Test_GetDIBits()
...
+
+    hdcMem = CreateCompatibleDC(NULL);
+    ok(hdcMem != NULL, "CreateCompatibleDC failed\n");
+    if (hdcMem == NULL)
+    {
+        skip("No hdcMem\n");
+    }
+    else
+    {
+        ok_int(GetDIBits(hdcMem, hbmp, 0, 4, pvBits, pbi, DIB_RGB_COLORS), 0);
```
HBelusca, line 524: See comment above. (Concerning these three successive ok() tests.)

modules/rostests/apitests/gdi32/GetDIBits.c
```diff
@@ -452,34 +503,56 @@ void Test_GetDIBits()
...
+        ok(DeleteDC(hdcMem), "DeleteDC failed\n");
+    }
 
     HeapFree(GetProcessHeap(), 0, pvBits);
-    DeleteDC(hdcMem);
-    ReleaseDC(NULL, hdcScreen);
+
+noBits2:
+noBmp3:
+    ok(ReleaseDC(NULL, hdcScreen) == 1, "ReleaseDC failed\n");
```
HBelusca, line 534: See comment above.

modules/rostests/apitests/gdi32/GetDIBits.c
ThFabba: I don't agree, unfortunately ;)

While the macro thing may be generally a good rule, `ok` is well-defined for this purpose. There are hundreds of places where it's called this way. The only reason this isn't good practice is that in case of failure you can't easily see the actual result for debugging.

As for testing unnecessary calls... this can be annoying if the "unrelated" feature is unreliable, as it causes failures in unrelated tests; but it can also find real bugs and broken assumptions in the test. Since there's no clear answer, my advice would simply be, as so often: don't change it in existing code.

modules/rostests/apitests/gdi32/GetDIBits.c
SergeGautherie: > Handles are not pointers

No: they are defined as pointers.

modules/rostests/apitests/gdi32/GetDIBits.c
SergeGautherie: Wow! You could just write 'apply everywhere', you know...

modules/rostests/apitests/gdi32/GetDIBits.c
SergeGautherie: I disagree on both remarks:
1. What Thomas answered.
I might even argue that unsafe macros are bad...
2. Checking cleanup functions is not about testing them, it's about (fully) checking how the current test/run behaves.
Thus, I don't understand that last advice: I think tests should explicitly check and report, as much as possible! (Especially while developing an alpha-stage OS...)

HBelusca: Well then, why not testing the printf and the free() calls as well in every unrelated tests? 🤣 🤣

SergeGautherie: ??why not??

Sure, go ahead!
And let me know if you find other `GDI32` related functions...

modules/rostests/apitests/gdi32/GetDIBits.c
```diff
@@ -452,34 +503,56 @@ void Test_GetDIBits()
...
     Test_GetDIBits_xBpp(1);
     Test_GetDIBits_xBpp(4);
     Test_GetDIBits_xBpp(8);
     Test_GetDIBits_xBpp(16);
     Test_GetDIBits_xBpp(24);
     Test_GetDIBits_xBpp(32);
+
     Test_GetDIBits();
 }
-
```
learn-more, line 485: why....

SergeGautherie: Removed...

PR #2901, opened 2020-06-06
[CMAKE] Update to 3.17 policies

ThFabba:
JIRA issue: [CORE-17109](https://jira.reactos.org/browse/CORE-17109)

Due Date: June 20th


CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 
-cmake_minimum_required(VERSION 3.2.1)
+cmake_minimum_required(VERSION 3.17.0)
 cmake_policy(VERSION 3.2.1)
 
 if(NOT CMAKE_VERSION MATCHES "ReactOS")
```
SergeGautherie, line 5: Is this check still true/needed? For non-MSVC?

sdk/cmake/host-tools.cmake
```diff
@@ -27,17 +27,23 @@ function(configure_host_tools HOST_TOOLS_DIR)
 
     set_property(SOURCE host_tools PROPERTY SYMBOLIC 1)
 
+    include(${HOST_TOOLS_DIR}/ImportExecutables.cmake)
+    include(${HOST_TOOLS_DIR}/TargetList.cmake)
+
+    foreach(_target ${NATIVE_TARGETS})
+        get_target_property(_target_location native-${_target} IMPORTED_LOCATION_DEBUG)
+        list(APPEND _target_locations ${_target_location})
```
learn-more, line 35: Does cmake cache the `_target_locations` in between runs?
e.g. does it grow each re-configure?

ThFabba: No, otherwise the same would be true of `SOURCE` in each directory.

CMakeLists.txt
```diff
@@ -1,20 +1,14 @@
 
-cmake_minimum_required(VERSION 3.2.1)
-cmake_policy(VERSION 3.2.1)
```
Extravert-ir, line 3: Looks like, `cmake_policy` can be omitted https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html

CMakeLists.txt
binarymaster: https://chat.reactos.org/reactos/pl/7weqy76ocfn55p9m6snmpzcddc

SergeGautherie: > https://chat.reactos.org/reactos/pl/7weqy76ocfn55p9m6snmpzcddc

Hum... It is written there that:
* (RosBE 2.2) `3.17-ReactOS` custom CMake has no special features, except being compiled to run on XP too.
* Want to keep the warning, with regard to possible 3rd party custom CMake.

Then, I would suggest to:
* Rephrase it, at least to add that official CMake is fully fine too.
* Update GitHub Actions MSVC builds to use RosBE CMake, not to trigger the (confusing) warning.

sdk/cmake/host-tools.cmake
```diff
@@ -27,17 +27,25 @@ function(configure_host_tools HOST_TOOLS_DIR)
...
     set_property(SOURCE host_tools PROPERTY SYMBOLIC 1)
 
+    include(${HOST_TOOLS_DIR}/ImportExecutables.cmake)
+    include(${HOST_TOOLS_DIR}/TargetList.cmake)
+
+    foreach(_target ${NATIVE_TARGETS})
+        get_target_property(_target_configs native-${_target} IMPORTED_CONFIGURATIONS)
+        list(GET _target_configs 0 _target_config)
+        get_target_property(_target_location native-${_target} IMPORTED_LOCATION_${_target_config})
+        list(APPEND _target_locations ${_target_location})
```
ThFabba, line 37: This is gross. Hoping to use generator expressions or something to avoid it.

sdk/cmake/host-tools.cmake
```diff
@@ -27,17 +27,23 @@ function(configure_host_tools HOST_TOOLS_DIR)
...
+    endforeach()
+
     # Make a host-tools target so it'll be built when needed
     # custom target + symbolic output prevents cmake from running
     # the command multiple times per build
     add_custom_command(
         COMMAND ${CMAKE_COMMAND} --build ${HOST_TOOLS_DIR}
-        OUTPUT host_tools)
+        OUTPUT host_tools
+        BYPRODUCTS ${_target_locations})
```
learn-more, line 44: nice

Extravert-ir:
Are you sure we need so many commits for this? :)

learn-more:
each commit can be reverted like this, which seems nicer than squashing everything on a pile.

ThFabba:
Also, some of these changes (`Use CMake 3.17 default policies.`) come with a bit of a regression risk since we might start to build things differently. I'd like to be able to have a regress-test point at something specific.
So, we don't _need_ this many commits, but I think they might be useful.

PR #2909, opened 2020-06-08
[VIDEOPRT] IntVideoPortCreateAdapterDeviceObject(): 1+1 trivial fixes

SergeGautherie:
Detected by Cppcheck: funcArgOrderDifferent, redundantAssignment.

### DIFF
--- a/win32ss/drivers/videoprt/videoprt.c
+++ b/win32ss/drivers/videoprt/videoprt.c
@@ -90,10 +90,10 @@ IntVideoPortDeferredRoutine(
 NTSTATUS
 NTAPI
 IntVideoPortCreateAdapterDeviceObject(
-    IN PDRIVER_OBJECT DriverObject,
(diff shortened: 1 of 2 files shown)


Extravert-ir:
This PR does not fix anything.
It will make sense only if you format the whole module

ThFabba:
Explain what the `funcArgOrderDifferent` warning is and how you're fixing it please. You don't seem to be changing anything about argument order, just annotations.

SergeGautherie:
> format the whole module

What do you mean?

--

> just annotations

Indeed, here (and at a few other (unfixed) places), Cppcheck is confused by unmatched `OPTIONAL`.

ThFabba:
I tend to consider at least
* SAL2 annotations
* braces-around-all-blocks
* printf formats
* StringSafe functions
things that can be done piecemeal without reformatting the whole file. So I think this is okay (especially because it fixes a static analysis issue).
But we should probably find a consensus on things like that and formalize it in Victor's new contribution guidelines.

PR #2910, opened 2020-06-09
[GDI32] MirrorRgnByWidth(): Fix 'phrgn' parameter handling

SergeGautherie:
Detected by Cppcheck: uselessAssignmentPtrArg.
Addendum to 794e7aa (r31523) and e647530 (r65893).

### DIFF
--- a/win32ss/gdi/gdi32/objects/region.c
+++ b/win32ss/gdi/gdi32/objects/region.c
@@ -85,7 +85,7 @@ FASTCALL
 MirrorRgnByWidth(
     _In_ HRGN hrgn,
     _In_ INT Width,
-    _In_ HRGN *phrgn)
+    _Out_opt_ HRGN *phrgn)
 {
     INT cRgnDSize, Ret = 0;
     PRGNDATA pRgnData;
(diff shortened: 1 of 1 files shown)


Extravert-ir:
Note: this code path is not used anywhere (all usages are commented out)

SergeGautherie:
> Note: this code path is not used anywhere (all usages are commented out)

I saw that. Fixing it should help potential callers...

ThFabba:
This should be a straightforward fix, but the unnecessary changes are turning it into something questionable.
Keep the bug fix. Because you're already touching this code, fixing the wrong annotations is also good (it would not be worth doing on its own). Revert everything else.

peterooch:
Just for note: most DC mirroring code is broken so I don't think this will change much.

SergeGautherie:
> it would not be worth doing on its own

I have a different point of view: wrong annotation adds confusion.

> Revert everything else.

Trying a simpler version...

SergeGautherie:
> As you may or may not have noticed, your test was not successful. 

(I am not sure what you meant.)

> get rid of the static and the comment

Reverted.

PR #2912, opened 2020-06-11
[TASKMGR] CachedGetUserFromSid(): Fix user name length on cache hit

SergeGautherie:
Return string length, not buffer size.

### DIFF
--- a/base/applications/taskmgr/perfdata.c
+++ b/base/applications/taskmgr/perfdata.c
@@ -144,16 +144,16 @@ CachedGetUserFromSid(
         if (EqualSid((PSID)&pEntry->Data, pSid))
         {
             wcsncpy(pUserName, pEntry->pszName, cwcUserName);
-            *pcwcUserName = cwcUserName;
+            *pcwcUserName = wcslen(pUserName);
             return;
         }
     }
 
     /* We didn't find the SID in the list, get the name conventional */
     SidToUserName(pSid, pUserName, cwcUserName);
+    *pcwcUserName = wcslen(pUserName);
 
     /* Allocate a new entry */
-    *pcwcUserName = wcslen(pUserName);
     cwcUserName = *pcwcUserName + 1;
     cbSid = GetLengthSid(pSid);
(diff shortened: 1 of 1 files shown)


base/applications/taskmgr/perfdata.c
```diff
@@ -144,16 +145,16 @@ CachedGetUserFromSid(
         if (EqualSid((PSID)&pEntry->Data, pSid))
         {
             wcsncpy(pUserName, pEntry->pszName, cwcUserName);
-            *pcwcUserName = cwcUserName;
+            *pcwcUserName = wcslen(pUserName);
```
SergeGautherie, line 148: As this out value is (currently) unused, a simplification would be to change this parameter to an in-only one.
What do you think?

HBelusca: Let's not introduce additional changes unrelated to the bug you wanted to fix.

SergeGautherie: > additional changes unrelated to the bug

That would be a different/final fix/optimization, not an additional change.
That seems very related to this code...

ThFabba:
Please explain the bug you're fixing. What does "on hit" mean? Please remove unrelated and cosmetic changes.

SergeGautherie:
> Please explain the bug you're fixing. What does "on hit" mean? Please remove unrelated and cosmetic changes.

PR description added. Cache hit, != miss. Reverted.

base/applications/taskmgr/perfdata.h
```diff
@@ -33,6 +33,8 @@ typedef struct _PERFDATA
 	WCHAR				ImageName[MAX_PATH];
 	HANDLE				ProcessId;
 	WCHAR				UserName[MAX_PATH];
+// TODO: Store actual length, to avoid wcslen() on CachedGetUserFromSid() cache hit.
+//	ULONG				UserNameLength;
```
ThFabba, line 37: This is not necessary unless someone finds that it becomes a bottleneck.

SergeGautherie: I think TaskMgr, running continuously, should use a little resource as possible. Anyway.
Reverted.

PR #2913, opened 2020-06-11
[TASKMGR] Check 'Heap*Alloc()' results

SergeGautherie:


base/applications/taskmgr/dbgchnl.c
```diff
@@ -243,20 +255,37 @@ static int enum_channel(HANDLE hProcess, EnumChannelCB ce, void* user, unsigned
                         if (!wcscmp(cache[j], buffer + 1)) break;
                     if (j != used_cache) continue;
                     if (used_cache == num_cache)
+                    {
                         cache = HeapReAlloc(GetProcessHeap(), 0, cache, (num_cache *= 2) * sizeof(WCHAR*));
```
HBelusca, line 259: There is a bug:
If HeapReAlloc fails, it will return NULL, but
> [...] the original memory is not freed, and the original handle and pointer are still valid.

Therefore here in case of failure you are leaking the original "cache" buffer.

SergeGautherie: Indeed, I missed to update this case too.
Fixed.

base/applications/taskmgr/perfdata.c
```diff
@@ -211,12 +215,16 @@ void PerfDataRefresh(void)
 
     /* Get processor time information */
     SysProcessorTimeInfo = (PSYSTEM_PROCESSOR_PERFORMANCE_INFORMATION)HeapAlloc(GetProcessHeap(), 0, sizeof(SYSTEM_PROCESSOR_PERFORMANCE_INFORMATION) * SystemBasicInfo.NumberOfProcessors);
+    if (!SysProcessorTimeInfo)
```
ThFabba, line 218: I'd prefer just skipping the work related to this (i.e. the filling of the buffer up here and the `Save system processor time info` section below) but continuing on with the rest.

base/applications/taskmgr/perfdata.c
```diff
@@ -229,13 +237,17 @@ void PerfDataRefresh(void)
     {
         BufferSize += 0x10000;
         SysHandleInfoData = (LPBYTE)HeapAlloc(GetProcessHeap(), 0, BufferSize);
+        if (!SysHandleInfoData)
```
ThFabba, line 240: Same here, this can just be skipped.

base/applications/taskmgr/dbgchnl.c
```diff
@@ -207,20 +207,32 @@ typedef int (*EnumChannelCB)(HANDLE, void*, WCHAR*, void*);
  */
 static int enum_channel(HANDLE hProcess, EnumChannelCB ce, void* user, unsigned unique)
```
ThFabba, line 208: This function (and related code) is Wine-specific (i.e. dead code for us) and synced to a 15 year old version of Wine (i.e. won't in fact work, even if run in Wine today).
We should do one of four things:
* Leave it as is and don't spend time on it
* Entirely remove the feature
* Sync it to current Wine, such that it actually works if run in Wine
* Find a way to make it work on ROS (we have debug channels, but not libwine.so, so it's going to take some rethinking... or it may not be possible)

base/applications/taskmgr/perfdata.c
```diff
@@ -325,6 +343,10 @@ void PerfDataRefresh(void)
 
     /* Now alloc a new PERFDATA array and fill in the data */
     pPerfData = (PPERFDATA)HeapAlloc(GetProcessHeap(), HEAP_ZERO_MEMORY, sizeof(PERFDATA) * ProcessCount);
+    if (!pPerfData)
+    {
+        goto noPerfData;
```
ThFabba, line 348: Labels in this function start with a capital letter.

base/applications/taskmgr/perfdata.c
```diff
@@ -247,6 +259,12 @@ void PerfDataRefresh(void)
     {
         BufferSize += 0x10000;
         pBuffer = (LPBYTE)HeapAlloc(GetProcessHeap(), 0, BufferSize);
+        if (!pBuffer)
```
ThFabba, line 262: Failing in the case of this or pPerfData is reasonable.

SergeGautherie: Made this case skippable and some others too

base/applications/taskmgr/perfdata.c
```diff
@@ -245,51 +266,77 @@ void PerfDataRefresh(void)
...
+        pSysProcessInfo = NULL;
+    }
 
     EnterCriticalSection(&PerfDataCriticalSection);
 
-    /*
-     * Save system performance info
-     */
-    memcpy(&SystemPerfInfo, &SysPerfInfo, sizeof(SYSTEM_PERFORMANCE_INFORMATION));
+    // Save system performance information.
```
ThFabba, line 296: Also revert comment style changes.

Extravert-ir:

PR #2914, opened 2020-06-12
[RDBSSLIB] RxCommonWrite(): 'BreakpointsSave' is 'DBG'-only

SergeGautherie:
MSVC 'Release' build reports:
'sdk\lib\drivers\rdbsslib\rdbss.c(4651): error C4101: 'BreakpointsSave': unreferenced local variable'

Addendum to f8e1661 (r75398).


Extravert-ir:
See the answer in #2916

PR #2916, opened 2020-06-12
[RTL] actctx.c: Some variables are 'DBG'-only

SergeGautherie:
MSVC 'Release' build reports multiple:
'sdk\lib\rtl\actctx.c(...): error C4101: '...': unreferenced local variable'

Addendum to 69f20a9 (r45845).

### DIFF
--- a/sdk/lib/rtl/actctx.c
+++ b/sdk/lib/rtl/actctx.c
@@ -1670,8 +1670,10 @@ static BOOL parse_nummethods(const xmlstr_t *str, struct entity *entity)
             num = num * 10 + *curr - '0';
         else
         {
+#if DBG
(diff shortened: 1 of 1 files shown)


Extravert-ir:
MSVC release builds should not turn this warning into an error, that's what we do in GCC release builds

SergeGautherie:
> MSVC release builds should not turn this warning into an error,

'Release' does not do anything special:
'sdk/cmake/msvc.cmake' enables it for all builds:
`add_compile_flags("... /we4101 ...")`

> that's what we do in GCC release builds

Do you want code copied from there? Which one?

>  closed

Then, how is this going to be fixed?

PR #2918, opened 2020-06-12
[MSI][OLE32] 1+1 removals, related to wine/asm.h

SergeGautherie:
MSVC build reports:
{{
dll\win32\msi\msvchelper.h(3): warning C4005: '__ASM_DEFINE_FUNC': macro redefinition
sdk\include\reactos\wine/asm.h(60): note: see previous definition of '__ASM_DEFINE_FUNC'
dll\win32\msi\custom.c(547): warning C4003: not enough actual parameters for macro '__ASM_DEFINE_FUNC'
}}

Addendum to 476c99b and 9efafd6.

### DIFF
--- a/dll/win32/msi/msvchelper.h
+++ b/dll/win32/msi/msvchelper.h
@@ -1,6 +1,5 @@
 
 #ifdef __i386__
-#define __ASM_DEFINE_FUNC(name,suffix,code)
 
 typedef unsigned int (__stdcall *__MSVC__MsiCustomActionEntryPoint)(unsigned int);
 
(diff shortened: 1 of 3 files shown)


dll/win32/msi/msvchelper.h
```diff
@@ -1,6 +1,6 @@
 
 #ifdef __i386__
-#define __ASM_DEFINE_FUNC(name,suffix,code)
+#define __ASM_DEFINE_FUNC(name,code)
```
ThFabba, line 3: There's no need to redefine it at all, since this is how it's already defined in `wine/asm.h`. So please just delete this line.

SergeGautherie: Well, with regard to fixing the warnings only, an additional sync change would rather be to add `#if !defined(_MSC_VER)`...
Nonetheless, if you confirm the local (re)definition is unwanted, because the full global definition should always apply, then I would concur to remove it.

ThFabba: I have no idea what you're talking about. This file is entirely ROS-specific, and is used only in MSVC builds. The global definition matches your new definition exactly, therefore this line is pointless and should be deleted.

SergeGautherie: Right, obviously.
Fully removed.

PR #2919, opened 2020-06-13
Add FDPRINT and use filtered debugging in usbstor

Extravert-ir:
As we have debug filters properly implemented now (thanks @HBelusca!), I decided to alter our `debug.h` by adding `FDPRINT_` macro along with `ERR_/WARN_` etc.
`usbstor` driver is an example of usage here, feel free to suggest other ideas.

This PR also adds a custom component ID to debugger. This is an open discussion - we need a way to add our own components to ease the debugging experience.

After all questions figured out, I'll convert `usbport`, `usbhub` and `scsiport` drivers to filtered debugging too


drivers/usb/usbstor/dbghlp.h
```diff
@@ -0,0 +1,18 @@
...
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Debug helpers
+ * COPYRIGHT:   2020 Victor Perevertkin (victor.perevertkin@reactos.org)
+ */
+
+#define NDEBUG
+#include <debug.h>
+
+#define DBGLVL_DISK (1 << 31)
+#define DBGLVL_PNP (1 << 30)
```
HBelusca, line 12: Since you are using bitfield, you should (best practice) OR these with DPFLTR_MASK when you use these defines.
Alternatively you can directly use the bitfield shift number (30 and 31 directly) and this should work fine as well.
> Another easy way to use the Level parameter is to always use explicit bitfields. If you choose this method, you may wish to use a bitwise OR of the value DPFLTR_MASK with your bitfield. This value ensures that you do not accidentally use a value less than 32.

(see https://docs.microsoft.com/en-us/windows-hardware/drivers/devtest/reading-and-filtering-debugging-messages )

ntoskrnl/kdbg/kdb_cli.c
```diff
@@ -301,6 +301,10 @@ ComponentTable[] =
     { "XSAVE",          DPFLTR_XSAVE_ID         },
     { "SE",             DPFLTR_SE_ID            },
     { "DRIVEEXTENDER",  DPFLTR_DRIVEEXTENDER_ID },
+//
+// ReactOS-specific components
+//
+    { "ROSUSBPORT",     DPFLTR_ROSUSBPORT_ID    },
```
HBelusca, line 307: Why does this need a separate component?

Extravert-ir: The current set of components has only `USBSTOR` for USB-related stuff. I can assign `usbport.sys` to some ID which does not have "USB" in it's name, but I think it may be confusing

Extravert-ir: Same for `usbhub.sys`

drivers/usb/usbstor/pdo.c
```diff
@@ -551,14 +548,14 @@ USBSTOR_PdoHandlePnp(
                break;
            }
 
-           DPRINT1("USBSTOR_PdoHandlePnp: IRP_MN_QUERY_ID IdType %x unimplemented\n", IoStack->Parameters.QueryId.IdType);
+           WARN("IRP_MN_QUERY_ID IdType %x unimplemented\n", IoStack->Parameters.QueryId.IdType);
```
tkreuzer, line 551: Not displayed by default.

tkreuzer (approved):

PR #2921, opened 2020-06-13
Revert "[FONTS] Add Selawik as a substitution for Segoe UI" CORE-17122

oleg-dubinskiy:
## Purpose

Revert commit 97fb49a3d2bfdbbf5f03e193f922a80dd39ca9ba. It will fix regression with incorrectly displayed Cyrillic text in all apps that use Segoe UI font.

JIRA issue: [CORE-17122](https://jira.reactos.org/browse/CORE-17122)

## Proposed changes

Completely revert the guilty commit. It seems to me the best solution for now. 

- delete the substitution for Selawik;
- delete the fonts itself and their license files.

## Result

Before:
0 4 15-dev-31-g97fb49a

After:
0 4 15-dev-30-ga7e4d66


ThFabba:
> Original Segoe UI font also has no Cyrillic support

But presumably it works on Windows? Does that mean Windows picks a different font, and therefore so should ROS?

oleg-dubinskiy:
Oops, sorry. I was wrong. :sweat_smile: 
Just checked : original Segoe UI fonts from Windows Vista, same as from 10, **has** support for Cyrillic glyphs:
segoeui_cyrillic
I checked probably broken version of Segoe UI earlier.
So not surprising that it works on Windows as it should.
Then probably better solution indeed will be to implement those glyphs in Selawik also.
Although Windows 10 has no Segoe UI substitution at all:
VirtualBox_Windows 10_14_06_2020_12_03_35
But nevertheless, Segoe UI is displayed there completely correctly.

PR #2925, opened 2020-06-15
[RAPPS] snapshot url support

kernelbin:
## Purpose

support specifying an URL for a snapshot

`KI7 `3DZ_D {RLN }B6R15

### DIFF
--- a/base/applications/rapps/asyncinet.cpp
+++ b/base/applications/rapps/asyncinet.cpp
@@ -5,13 +5,170 @@
  * COPYRIGHT:   Copyright 2020 He Yang            (1160386205@qq.com)
  */
 
-
 #include "rapps.h"
 #include <windows.h>
 #include <wininet.h>
-#include <strsafe.h>
+#include <atlbase.h>
 #include "asyncinet.h"
 
+BOOL AsyncInetAcquire(pASYNCINET AsyncInet);
+VOID AsyncInetRelease(pASYNCINET AsyncInet);
+int AsyncInetPerformCallback(pASYNCINET AsyncInet,
+    ASYNC_EVENT Event,
+    WPARAM wParam,
+    LPARAM lParam
+    );
+VOID CALLBACK AsyncInetStatusCallback(
+    HINTERNET hInternet,
+    DWORD_PTR dwContext,
+    DWORD dwInternetStatus,
+    LPVOID lpvStatusInformation,
+    DWORD dwStatusInformationLength
+    );
+VOID AsyncInetReadFileLoop(pASYNCINET AsyncInet);
+BOOL AsyncInetCleanUp(pASYNCINET AsyncInet);
+VOID AsyncInetFree(pASYNCINET AsyncInet);
+
+pASYNCINET AsyncInetDownload(LPCWSTR lpszAgent,
+    DWORD   dwAccessType,
+    LPCWSTR lpszProxy,
+    LPCWSTR lpszProxyBypass,
+    LPCWSTR lpszUrl,
+    BOOL bAllowCache,
+    ASYNCINET_CALLBACK Callback,
+    VOID* Extension
+    ) // start a download task
+{
+    pASYNCINET AsyncInet = NULL;
+    BOOL bSuccess = FALSE;
+    DWORD InetOpenUrlFlag = 0;
+    INTERNET_STATUS_CALLBACK OldCallbackFunc;
+
+    AsyncInet = (pASYNCINET)HeapAlloc(GetProcessHeap(), HEAP_ZERO_MEMORY, sizeof(ASYNCINET));
+    if (!AsyncInet)
+    {
+        OutputDebugStringA("At File: " __FILE__ " HeapAlloc returned 0\n");
+        return 0;
+    }
+
+    AsyncInet->Callback = Callback;
+    AsyncInet->Extension = Extension;
+
+
+    AsyncInet->hEventHandleCreated = CreateEvent(NULL, FALSE, FALSE, NULL);
+
+    InitializeCriticalSection(&(AsyncInet->CriticalSection));
+    AsyncInet->ReferenceCnt = 0;
+    AsyncInet->hEventRefZero = CreateEvent(NULL, TRUE, TRUE, NULL);
+    AsyncInet->hEventHandleClose = CreateEvent(NULL, FALSE, FALSE, NULL);
+
+    if (AsyncInet->hEventHandleCreated && AsyncInet->hEventRefZero && AsyncInet->hEventHandleClose)
+    {
+        AsyncInet->hInternet = InternetOpenW(lpszAgent, dwAccessType, lpszProxy, lpszProxyBypass, INTERNET_FLAG_ASYNC);
+
+        if (AsyncInet->hInternet)
+        {
+            OldCallbackFunc = InternetSetStatusCallbackW(AsyncInet->hInternet, AsyncInetStatusCallback);
+            if (OldCallbackFunc != INTERNET_INVALID_STATUS_CALLBACK)
+            {
+                InetOpenUrlFlag |= INTERNET_FLAG_KEEP_CONNECTION | INTERNET_FLAG_PASSIVE | INTERNET_FLAG_RESYNCHRONIZE;
+                if (!bAllowCache)
+                {
+                    InetOpenUrlFlag |= INTERNET_FLAG_DONT_CACHE | INTERNET_FLAG_PRAGMA_NOCACHE | INTERNET_FLAG_RELOAD;
+                }
+
+                if (AsyncInetAcquire(AsyncInet))
+                {
+                    AsyncInet->hInetFile = InternetOpenUrlW(AsyncInet->hInternet, lpszUrl, 0, 0, InetOpenUrlFlag, (DWORD_PTR)AsyncInet);
+
+                    if (AsyncInet->hInetFile)
+                    {
+                        // operate complete synchronously
+                        bSuccess = TRUE;
+                        AsyncInetReadFileLoop(AsyncInet);
+                        AsyncInetRelease(AsyncInet);
+                    }
+                    else
+                    {
+                        if (GetLastError() == ERROR_IO_PENDING)
+                        {
+                            // everything fine. waiting for handle created
+                            switch (WaitForSingleObject(AsyncInet->hEventHandleCreated, INFINITE))
+                            {
+                            case WAIT_OBJECT_0:
+                                bSuccess = TRUE;
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    if (!bSuccess)
+    {
+        AsyncInetFree(AsyncInet);
+        AsyncInet = NULL;
+    }
+
+    return AsyncInet;
+}
+
+BOOL AsyncInetCancel(pASYNCINET AsyncInet) // mark as cancelled (this will send a cancel notificaion at last) and do graceful cleanup.
+{
+    if (AsyncInet)
+    {
+        AsyncInet->bCancelled = TRUE;
+        return AsyncInetCleanUp(AsyncInet);
+    }
+    else
+    {
+        return FALSE;
+    }
+}
+
+BOOL AsyncInetAcquire(pASYNCINET AsyncInet) // try to increase refcnt by 1. if returned FALSE, AsyncInet should not be used anymore
+{
+    BOOL bResult = FALSE;
+    if (AsyncInet)
+    {
+        EnterCriticalSection(&(AsyncInet->CriticalSection));
+        if (!(AsyncInet->bCleanUp))
+        {
+            AsyncInet->ReferenceCnt++;
+            ResetEvent(AsyncInet->hEventRefZero); // no longer zero
+            bResult = TRUE;
+        }
+        // otherwise (AsyncInet->bCleanUp == TRUE)
+        // AsyncInetAcquire will return FALSE. 
+        // In this case, any thread should no longer use this AsyncInet
+
+        LeaveCriticalSection(&(AsyncInet->CriticalSection));
+    }
+
+    return bResult;
+}
+
+VOID AsyncInetRelease(pASYNCINET AsyncInet) // try to decrease refcnt by 1
+{
+    if (AsyncInet)
+    {
+        EnterCriticalSection(&(AsyncInet->CriticalSection));
+        if (AsyncInet->ReferenceCnt)
+        {
+            AsyncInet->ReferenceCnt--;
+            if (AsyncInet->ReferenceCnt == 0)
+            {
+                SetEvent(AsyncInet->hEventRefZero);
+            }
+        }
+        else
+        {
+            ATLASSERT(FALSE); // should be always non-negative, can not decrease anymore.
+        }
+        LeaveCriticalSection(&(AsyncInet->CriticalSection));
+    }
+}
 
 int AsyncInetPerformCallback(pASYNCINET AsyncInet,
     ASYNC_EVENT Event,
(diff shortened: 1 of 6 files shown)


base/applications/rapps/asyncinet.cpp
```diff
@@ -0,0 +1,258 @@
...
+                        break;
+                    }
+                }
+            }
+        }
+        break;
+        case ERROR_INVALID_HANDLE:
+            if (AsyncInet->bIsCancelled)
+            {
+                // this is most likely not an error. when canceling, using closed/invalid handle may lead to this
```
learn-more, line 142: I would say this is up to the class calling the asyncinet?
In some cases this might be true, in other cases this will be an actual error during transmission.

base/applications/rapps/asyncinet.cpp
```diff
@@ -0,0 +1,258 @@
...
+    VOID* Extension
+    )
+{
+    pASYNCINET AsyncInet = NULL;
+    BOOL bSuccess = FALSE;
+    DWORD InetOpenUrlFlag = 0;
+    INTERNET_STATUS_CALLBACK OldCallbackFunc;
+
+    AsyncInet = (pASYNCINET)HeapAlloc(GetProcessHeap(), HEAP_ZERO_MEMORY, sizeof(ASYNCINET));
+    if (!AsyncInet)goto cleanup;
```
learn-more, line 175: At this point, return NULL would be easier.
And it would be nice to include a debug log that this failed, since it would be quite unexpected.

base/applications/rapps/asyncinet.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+    AsyncInet->Callback = Callback;
+    AsyncInet->Extension = Extension;
+
+    AsyncInet->hInternet = InternetOpenW(lpszAgent, dwAccessType, lpszProxy, lpszProxyBypass, INTERNET_FLAG_ASYNC);
+    if (!AsyncInet->hInternet)goto cleanup;
```
learn-more, line 181: It would be cleaner imo to either nest a few levels like this:
```
if (AsyncInet->hInternet)
{
    OldCallbackFunc = InternetSetStatusCallbackW(AsyncInet->hInternet, AsyncInetStatusCallback);
    if (OldCallbackFunc != INTERNET_INVALID_STATUS_CALLBACK)
    {
        //
    }
}
```

base/applications/rapps/asyncinet.cpp
```diff
@@ -0,0 +1,258 @@
...
+        if (GetLastError() != ERROR_IO_PENDING)
+        {
+            goto cleanup;
+        }
+    }
+    else
+    {
+        // TODO: If I remember it correctly, sometimes the file is cache before
+        // thus may lead to InternetOpenUrlW return immediately with the handle.
+        // more investigation required. And if it's true, it should be handled correctly.
```
learn-more, line 203: Most image hosts (like imgur) use image caching,
otherwise something like this might work: https://refreshyourcache.com/en/cache-test/

base/applications/rapps/asyncinet.cpp
```diff
@@ -0,0 +1,258 @@
...
+{
+    if (!AsyncInet)
+    {
+        return FALSE;
+    }
+
+    AsyncInet->bIsCancelled = TRUE;
+
+    // AsyncInet may got freed while handling INTERNET_STATUS_HANDLE_CLOSING.
+    // so store the handle first, then close them
```
learn-more, line 244: That seems like a race condition then?

base/applications/rapps/gui.cpp
```diff
@@ -541,37 +652,88 @@ class CAppSnapshotPreview :
...
+                DownloadParam->DownloadFileName.ReleaseBuffer();
+                delete DownloadParam;
+                DisplayFailed();
+                return FALSE;
+            }
+            DownloadParam->DownloadFileName.ReleaseBuffer();
+
+            DownloadParam->hFile = CreateFileW(DownloadParam->DownloadFileName.GetString(),
+                GENERIC_WRITE, 0, NULL, CREATE_ALWAYS, FILE_ATTRIBUTE_NORMAL, NULL);
+            if (INVALID_HANDLE_VALUE == DownloadParam->hFile)
```
learn-more, line 710: We don't generally swap the order of these so that the constant is to the left

base/applications/rapps/asyncinet.cpp
kernelbin: I think yes. we don't know which will happen first, download complete or download cancelled

base/applications/rapps/asyncinet.cpp
kernelbin: yes and now I add code to verify AsyncInet->bIsCancelled. Is this enough?

base/applications/rapps/asyncinet.cpp
kernelbin: It's not about caching, it about will InternetOpenUrl still has a chance to directly return with the handle (instead of return 0 with GetLastError() == IO_PENDING) even if I specify INTERNET_FLAG_ASYNC in InternetOpen (the entry of InternetOpenUrl in msdn didn't elaborate this, but I do encountered this situation years ago). I just need to seek for confirmation

learn-more: yeah this should be taken into account.

kernelbin: Great, I have found the evidence.
https://docs.microsoft.com/zh-cn/windows/win32/wininet/common-functions#setting-up-asynchronous-operations
> Therefore, depending on the request, the operation might complete synchronously or asynchronously. The application should check the return code. If a function returns FALSE or NULL, and GetLastError returns ERROR_IO_PENDING, the request has been made asynchronously

I'm going to improve this now

base/applications/rapps/asyncinet.cpp
kernelbin: ok, I've improved this.

base/applications/rapps/asyncinet.cpp
kernelbin: I prefer handling cleanup using stuff like __try{}__finally{}, but nesting some levels seems tidy & clear too 👍
I've already improved this !

kernelbin:
I found some bugs. I'm still investigating, so I turn this PR to a draft.

base/applications/rapps/asyncinet.cpp
```diff
@@ -0,0 +1,273 @@
...
+{
+    pASYNCINET AsyncInet = NULL;
+    BOOL bSuccess = FALSE;
+    DWORD InetOpenUrlFlag = 0;
+    INTERNET_STATUS_CALLBACK OldCallbackFunc;
+
+    AsyncInet = (pASYNCINET)HeapAlloc(GetProcessHeap(), HEAP_ZERO_MEMORY, sizeof(ASYNCINET));
+    if (!AsyncInet)
+    {
+        OutputDebugStringA("At File: " __FILE__ " HeapAlloc returned 0");
```
learn-more, line 188: Missing \n

base/applications/rapps/asyncinet.cpp
learn-more: Both `AsyncInetDownload` and this function close these handles, but one of them cleans up the object.
This seems weird?

base/applications/rapps/asyncinet.cpp
kernelbin: oops..

kernelbin: sorry for that, now fixed.

base/applications/rapps/asyncinet.cpp
```diff
@@ -0,0 +1,405 @@
+/*
+ * PROJECT:     ReactOS Applications Manager
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Async Internet operation using WinINet
+ * COPYRIGHT:   Copyright 2020 He Yang            (1160386205@qq.com)
+ */
+
+#include "rapps.h"
+#include <windows.h>
```
learn-more, line 9: I am pretty sure that windows.h is covered by `rapps.h`

base/applications/rapps/asyncinet.cpp
```diff
@@ -0,0 +1,405 @@
...
+
+    AsyncInet->Callback = Callback;
+    AsyncInet->Extension = Extension;
+
+    AsyncInet->hEventHandleCreated = CreateEvent(NULL, FALSE, FALSE, NULL);
+
+    InitializeCriticalSection(&(AsyncInet->CriticalSection));
+    AsyncInet->ReferenceCnt = 0;
+    AsyncInet->hEventRefZero = CreateEvent(NULL, TRUE, TRUE, NULL);
+    AsyncInet->hEventHandleClose = CreateEvent(NULL, FALSE, FALSE, NULL);
```
learn-more, line 68: What about converting `AsyncInet` to a class?
```
class CAsyncInet:
    public CComObjectRootEx<CComMultiThreadModelNoCS>,
    public IUnknown
{
public:
~CAsyncInet()
{
    // cleanup every handle / pointer here
}
// all async functions / properties here.
};
```
Now you can create it like this:
```
CComObject<CAsyncInet>* pAsync;
HRESULT hr = CComObject<CAsyncInet>::CreateInstance(&pAsync);
if (FAILED_UNEQPECTEDLY(hr))
    return hr;
```
(Please note, at this point reference count is 0, so someone needs to AddRef it!)

And everyone that wants to use it, uses it like:
```
pAsync->AddRef();
// do something
pAsync->Release();
```

Or, since we are using c++:
```
void doSomething(pAsync)
{
    CComPtr<CAsyncInet> pRef = pAsync;   // Assigning to the smart pointer will call AddRef()
    // do something with pRef
}  // When the smart pointer is destroyed here, it will call Release()
```

base/applications/rapps/asyncinet.cpp
```diff
@@ -0,0 +1,405 @@
...
+        {
+            if (AsyncInet->bCancelled)
+            {
+                AsyncInetPerformCallback(AsyncInet, ASYNCINET_CANCELLED, 0, 0);
+            }
+            AsyncInetFree(AsyncInet);
+            return TRUE;
+        }
+        default:
+            return FALSE;
```
learn-more, line 362: `bCleanUp` is still 1 here, so it will never be cleaned up

base/applications/rapps/include/asyncinet.h
```diff
@@ -0,0 +1,68 @@
...
+{
+    HINTERNET hInternet;
+    HINTERNET hInetFile;
+
+    HANDLE hEventHandleCreated;
+
+    UINT ReferenceCnt;
+    CRITICAL_SECTION CriticalSection;
+    HANDLE hEventRefZero; // TODO: Use conditional variable instead when one day ROS has conditional variable
+    HANDLE hEventHandleClose;
```
learn-more, line 42: If you use reference counting correctly, you would not need those handles, and probably also not the critical section.
Basically:

Before doing anything, a thread (or function) increases reference count.
When starting a thread or sending something to another thread, (thread a creates thread b):

1. Thread a increments reference count
2. Thread a starts thread b
3. Thread b is responsible of decrementing the reference count when it is done with the object.

base/applications/rapps/include/asyncinet.h
```diff
@@ -0,0 +1,68 @@
...
+
+    BOOL bIsOpenUrlComplete;
+
+    BOOL bCleanUp;
```
learn-more, line 46: bCleanup is probably not needed,
when decrementing a reference count to 0, cleanup is implied (and noone else should have a reference to this object anymore!)
a function / thread should not receive an object with reference count 0, you could even assert on that (since someone handed you a pointer that was not 'safe' anymore, since they forgot to take a reference.

base/applications/rapps/asyncinet.cpp
learn-more: The first call to Release where reference count reaches 0 will destroy the object.

base/applications/rapps/asyncinet.cpp
```diff
@@ -0,0 +1,454 @@
...
+    AsyncInet->hInternet = NULL;
+    AsyncInet->hInetFile = NULL;
+
+    AsyncInet->Callback = Callback;
+    AsyncInet->Extension = Extension;
+
+    AsyncInet->hEventHandleCreated = CreateEvent(NULL, FALSE, FALSE, NULL);
+
+    InitializeCriticalSection(&(AsyncInet->CriticalSection));
+    AsyncInet->ReferenceCnt = 0;
```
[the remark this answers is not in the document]
kernelbin: yes, ```InternetOpen``` and ```InternetSetStatusCallbackW``` never run async.
as for the initial ReferenceCnt, it's increased at line ```85``` for the function itself, and at line ```125``` for the caller

base/applications/rapps/asyncinet.cpp
```diff
@@ -0,0 +1,454 @@
...
+
+BOOL AsyncInetAcquire(pASYNCINET AsyncInet) // try to increase refcnt by 1. if returned FALSE, AsyncInet should not be used anymore
+{
+    BOOL bResult = FALSE;
+    if (AsyncInet)
+    {
+        EnterCriticalSection(&(AsyncInet->CriticalSection));
+        if (!(AsyncInet->bCleanUp))
+        {
+            AsyncInet->ReferenceCnt++;
```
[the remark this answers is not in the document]
kernelbin: ah, yes this sounds good.

base/applications/rapps/asyncinet.cpp
```diff
@@ -0,0 +1,454 @@
...
+
+VOID AsyncInetRelease(pASYNCINET AsyncInet) // try to decrease refcnt by 1
+{
+    BOOL bCleanUp = FALSE;
+    if (AsyncInet)
+    {
+        EnterCriticalSection(&(AsyncInet->CriticalSection));
+        if (AsyncInet->ReferenceCnt)
+        {
+            AsyncInet->ReferenceCnt--;
```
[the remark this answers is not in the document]
kernelbin: may I ask why assert on AsyncInet? It's already checked a few lines above.

base/applications/rapps/asyncinet.cpp
```diff
@@ -0,0 +1,454 @@
...
+            if (AsyncInet->ReferenceCnt == 0)
+            {
+                bCleanUp = TRUE;
+                AsyncInet->bCleanUp = TRUE;
```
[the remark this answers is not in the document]
kernelbin: currently yes. 
should I set AsyncInet->bCleanUp as ```TRUE``` when AsyncInetCancel is called?

base/applications/rapps/asyncinet.cpp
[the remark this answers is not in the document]
learn-more: I meant AsyncInet->ReferenceCnt, my bad.

base/applications/rapps/asyncinet.cpp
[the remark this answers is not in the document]
learn-more: Well, if there is nothing else setting bCleanUp to TRUE, then bCleanup can be removed, and checks for bCleanup == TRUE can be changed to ReferenceCnt == 0

base/applications/rapps/asyncinet.cpp
```diff
@@ -0,0 +1,454 @@
...
+    }
+    return AsyncInet;
+}
+
+BOOL AsyncInetCancel(pASYNCINET AsyncInet) // mark as cancelled (this will send a cancel notificaion at last) and do graceful cleanup.
+{
+    if (AsyncInet)
+    {
+        EnterCriticalSection(&(AsyncInet->CriticalSection));
+        AsyncInet->bCancelled = TRUE;
```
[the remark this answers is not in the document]
kernelbin: I'm thinking about this:
when cancelling, other places may still using this AsyncInet->hInetFile (most likely in AsyncInetStatusCallback).
so just set it to zero IMO may lead to some error (I haven't verified this)
am I right? 🤣 should I use something like another critical section to protect it?

base/applications/rapps/asyncinet.cpp
```diff
@@ -0,0 +1,454 @@
...
+
+            AsyncInetReadFileLoop(AsyncInet);
+        }
+        break;
+        case ERROR_INVALID_HANDLE:
+        case ERROR_INTERNET_OPERATION_CANCELLED:
+        case ERROR_CANCELLED:
+            if (AsyncInet->bCleanUp || AsyncInet->bCancelled)
+            {
+                AsyncInetRelease(AsyncInet);
```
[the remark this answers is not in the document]
kernelbin: yes, I think there's no need to check bCleanUp

PR #2926, opened 2020-06-15
[FILESYSTEMS] Fix pool memory disclosure in NtQueryAttributesFile handlers

khanhnt2:
## Purpose

Fix structure alignment issue leads to pool memory disclosure. 

## Reproduce

Use MSVC 14.0 x86 (Visual Studio 2015) to compile ReactOS with the latest/HEAD commit.

## PoC

```c
#include <Windows.h>
#include <SubAuth.h>
#include <stdio.h>

typedef struct _IO_STATUS_BLOCK {
    union {
        NTSTATUS Status;
        PVOID    Pointer;
    } DUMMYUNIONNAME;
    ULONG_PTR Information;
} IO_STATUS_BLOCK, *PIO_STATUS_BLOCK;

typedef struct _FILE_BASIC_INFORMATION {
    LARGE_INTEGER CreationTime;
    LARGE_INTEGER LastAccessTime;
    LARGE_INTEGER LastWriteTime;
    LARGE_INTEGER ChangeTime;
    ULONG         FileAttributes;
} FILE_BASIC_INFORMATION, *PFILE_BASIC_INFORMATION;

/* Object Attributes */
typedef struct _OBJECT_ATTRIBUTES {
    ULONG Length;
    HANDLE RootDirectory;
    PUNICODE_STRING ObjectName;
    ULONG Attributes;
    PVOID SecurityDescriptor;
    PVOID SecurityQualityOfService;
} OBJECT_ATTRIBUTES, *POBJECT_ATTRIBUTES;

#define InitializeObjectAttributes(p, n, a, r, s) { \
     (p)->Length = sizeof(OBJECT_ATTRIBUTES); \
     (p)->RootDirectory = r; \
     (p)->Attributes = a; \
     (p)->ObjectName = n; \
     (p)->SecurityDescriptor = s; \
     (p)->SecurityQualityOfService = NULL; \
     }

typedef NTSTATUS(WINAPI *NTQUERYATTRIBUTESFILE)(POBJECT_ATTRIBUTES, PFILE_BASIC_INFORMATION);

void dump_bytes(PBYTE arr, DWORD length) {
    for (int i = 0; i < length; i++) {
        if (arr[i] < 16) {
            printf("0%x ", arr[i]);
        }
        else {
            printf("%x ", arr[i]);
        }
        if ((i + 1) % 16 == 0) {
            printf("\n");
        }
    }
    printf("\n");
}

int main()
{
    HMODULE ntdll = GetModuleHandleA("ntdll");
    NTQUERYATTRIBUTESFILE NtQueryAttributesFile = (NTQUERYATTRIBUTESFILE)GetProcAddress(ntdll, "NtQueryAttributesFile");
    BYTE infor[0x28] = { 0 }; // equal to FILE_BASIC_INFORMATION file_info = { 0 };
    infor[0x24] = 1;
    infor[0x25] = 1;
    infor[0x26] = 1;
    infor[0x27] = 1;
    IO_STATUS_BLOCK iosb = { 0 };
    OBJECT_ATTRIBUTES obat = { 0 };
    UNICODE_STRING filename;
    filename.Buffer = (PWCH)L"\\??\\C:\\";
    filename.Length = lstrlenW(filename.Buffer) * 2;
    filename.MaximumLength = 255;

    InitializeObjectAttributes(&obat, &filename, 0x40, NULL, NULL);
    NtQueryAttributesFile(&obat, (PFILE_BASIC_INFORMATION)infor);

    dump_bytes(infor, sizeof(FILE_BASIC_INFORMATION));
}
```

The expected result is the last 4 bytes of `infor` are different from `01 01 01 01`.

Stack trace:

```
 #0  0x805817bd ((001817bd) ntoskrnl.exe!memcpy+3d)
 #1  0x80476a88 ((00076a88) ntoskrnl.exe!IopParseDevice+2138 [h:\project\reactos\ntoskrnl\io\iomgr\file.c @ 1276])
 #2  0x8050d2fa ((0010d2fa) ntoskrnl.exe!ObpLookupObjectName+a3a [h:\project\reactos\ntoskrnl\ob\obname.c @ 1003])
 #3  0x805054db ((001054db) ntoskrnl.exe!ObOpenObjectByName+16b [h:\project\reactos\ntoskrnl\ob\obhandle.c @ 2609])
 #4  0x80476fa5 ((00076fa5) ntoskrnl.exe!IopQueryAttributesFile+205 [h:\project\reactos\ntoskrnl\io\iomgr\file.c @ 2404])
 #5  0x80478856 ((00078856) ntoskrnl.exe!NtQueryAttributesFile+16 [h:\project\reactos\ntoskrnl\io\iomgr\file.c @ 3985])
 #6  0x8054d99b ((0014d99b) ntoskrnl.exe!KiSystemCallTrampoline+1b [h:\project\reactos\ntoskrnl\include\internal\i386\ke.h @ 766])
 #7  0x8054b808 ((0014b808) ntoskrnl.exe!KiSystemServiceHandler+278 [h:\project\reactos\ntoskrnl\ke\i386\traphdlr.c @ 1846])
 #8  0x80403ea5 ((00003ea5) ntoskrnl.exe!KiFastCallEntry+96)
```

 ## Analysis

The root cause is the structure `_FILE_BASIC_INFORMATION` are added 4 padding bytes after field `FileAttributes` so `sizeof(_FILE_BASIC_INFORMATION)` is 0x28, not 0x24. And it is used to allocate pool memory at https://github.com/reactos/reactos/blob/393a2c42fec59341271a53d4f87c253a13aabf9c/ntoskrnl/io/iomgr/file.c#L1262-L1272 After that, filesystem driver will use this memory to fill file basic information but the last 4 bytes (padding bytes) aren't used. This vulnerability affects all filesystem drivers which support `FILE_INFORMATION_CLASS.FileBasicInformation` query.

### DIFF
--- a/drivers/filesystems/ext2/src/fileinfo.c
+++ b/drivers/filesystems/ext2/src/fileinfo.c
@@ -137,6 +137,7 @@ Ext2QueryFileInformation (IN PEXT2_IRP_CONTEXT IrpContext)
             }
 
             FileBasicInformation = (PFILE_BASIC_INFORMATION) Buffer;
+            RtlZeroBytes(FileBasicInformation, sizeof(FILE_BASIC_INFORMATION));
 
             FileBasicInformation->CreationTime = Mcb->CreationTime;
             FileBasicInformation->LastAccessTime = Mcb->LastAccessTime;
--- a/drivers/filesystems/ext2/src/fileinfo.c
+++ b/drivers/filesystems/ext2/src/fileinfo.c
@@ -137,7 +137,6 @@ Ext2QueryFileInformation (IN PEXT2_IRP_CONTEXT IrpContext)
             }
 
             FileBasicInformation = (PFILE_BASIC_INFORMATION) Buffer;
-            RtlZeroBytes(FileBasicInformation, sizeof(FILE_BASIC_INFORMATION));
 
             FileBasicInformation->CreationTime = Mcb->CreationTime;
             FileBasicInformation->LastAccessTime = Mcb->LastAccessTime;
--- a/ntoskrnl/io/iomgr/file.c
+++ b/ntoskrnl/io/iomgr/file.c
@@ -1264,7 +1264,6 @@ IopParseDevice(IN PVOID ParseObject,
                                                       TAG_IO);
                 if (FileBasicInfo)
                 {
-                    RtlZeroMemory(FileBasicInfo, sizeof(*FileBasicInfo));
                     /* Do the query */
                     Status = IoQueryFileInformation(FileObject,
                                                     FileBasicInformation,
--- a/ntoskrnl/io/iomgr/file.c
+++ b/ntoskrnl/io/iomgr/file.c
@@ -1262,9 +1262,9 @@ IopParseDevice(IN PVOID ParseObject,
                 FileBasicInfo = ExAllocatePoolWithTag(NonPagedPool,
                                                       sizeof(*FileBasicInfo),
                                                       TAG_IO);
-                RtlZeroMemory(FileBasicInfo, sizeof(*FileBasicInfo));
                 if (FileBasicInfo)
                 {
+                    RtlZeroMemory(FileBasicInfo, sizeof(*FileBasicInfo));
                     /* Do the query */
                     Status = IoQueryFileInformation(FileObject,
                                                     FileBasicInformation,
--- a/ntoskrnl/io/iomgr/file.c
+++ b/ntoskrnl/io/iomgr/file.c
@@ -1262,6 +1262,7 @@ IopParseDevice(IN PVOID ParseObject,
                 FileBasicInfo = ExAllocatePoolWithTag(NonPagedPool,
                                                       sizeof(*FileBasicInfo),
                                                       TAG_IO);
+                RtlZeroMemory(FileBasicInfo, sizeof(*FileBasicInfo));
                 if (FileBasicInfo)
                 {
                     /* Do the query */
--- a/drivers/filesystems/cdfs/fileinfo.c
+++ b/drivers/filesystems/cdfs/fileinfo.c
@@ -899,6 +899,8 @@ Return Value:
     //  We only support creation, last modify and last write times on Cdfs.
     //
 
+    RtlZeroMemory(Buffer, sizeof(FILE_BASIC_INFORMATION));
+
     Buffer->LastWriteTime.QuadPart =
     Buffer->CreationTime.QuadPart =
     Buffer->ChangeTime.QuadPart = Fcb->CreationTime;
--- a/drivers/filesystems/fastfat/finfo.c
+++ b/drivers/filesystems/fastfat/finfo.c
@@ -293,6 +293,8 @@ VfatGetBasicInformation(
     if (*BufferLength < sizeof(FILE_BASIC_INFORMATION))
         return STATUS_BUFFER_OVERFLOW;
 
+    RtlZeroMemory(BasicInfo, sizeof(FILE_BASIC_INFORMATION));
+
     if (vfatVolumeIsFatX(DeviceExt))
     {
         FsdDosDateTimeToSystemTime(DeviceExt,
--- a/drivers/filesystems/ntfs/finfo.c
+++ b/drivers/filesystems/ntfs/finfo.c
@@ -107,6 +107,8 @@ NtfsGetBasicInformation(PFILE_OBJECT FileObject,
     if (*BufferLength < sizeof(FILE_BASIC_INFORMATION))
         return STATUS_BUFFER_TOO_SMALL;
 
+    RtlZeroMemory(BasicInfo, sizeof(FILE_BASIC_INFORMATION));
+
     BasicInfo->CreationTime.QuadPart = FileName->CreationTime;
     BasicInfo->LastAccessTime.QuadPart = FileName->LastAccessTime;
     BasicInfo->LastWriteTime.QuadPart = FileName->LastWriteTime;
--- a/drivers/filesystems/reiserfs/src/fileinfo.c
+++ b/drivers/filesystems/reiserfs/src/fileinfo.c
@@ -3,7 +3,7 @@
  * PROJECT:          ReiserFs file system driver for Windows NT/2000/XP/Vista.
  * FILE:             fileinfo.c
  * PURPOSE:          
- * PROGRAMMER:       Mark Piper, Matt Wu, Bo Brantén.
+ * PROGRAMMER:       Mark Piper, Matt Wu, Bo BrantÃ©n.
  * HOMEPAGE:         
  * UPDATE HISTORY: 
  */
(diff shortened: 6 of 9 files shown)


SergeGautherie:
At first glance, I would say the fix is wrong and the behavior is correct.
Do you get a different behavior when compiled and run on Windows (XP/S03)?

learn-more:
This static assertion works fine when building a driver with the windows ddk:
```
    static_assert(sizeof(FILE_BASIC_INFORMATION) == 5 * 8, "Wrong size!");
```
So this approach is most likely not correct.
What the solution could be, is zeroing out the memory after allocating it, using the same size as the allocation used.

khanhnt2:
Which compiler did you use? I use msvc 14.0 x86 to compile. And the 
```c
    static_assert(sizeof(FILE_BASIC_INFORMATION) == 5 * 8, "Wrong size!");
```
works. When I added that `static_assert` at reactos/ntoskrnl/io/iomgr/file.c:1265 ( after `FileBasicInfo = ExAllocatePoolWithTag(NonPagedPool, 
                                       sizeof(*FileBasicInfo), 
                                       TAG_IO);`) the compile progress will be interrupt and throw error `..\ntoskrnl\io\iomgr\file.c(1265): error C2338: Wrong size!`

khanhnt2:
> 
> 
> At first glance, I would say the fix is wrong and the behavior is correct.
> Do you get a different behavior when compiled and run on Windows (XP/S03)?

You can compile the PoC on windows then copy PE file to ReactOS and run to see the result.

SergeGautherie:
> You can compile the PoC on windows then copy PE file to ReactOS and run to see the result.

Hum, _you_ are expected to test it on _Windows_ (no ReactOS code involved) and report the result...

ntoskrnl/io/iomgr/file.c
```diff
@@ -1262,6 +1262,7 @@ IopParseDevice(IN PVOID ParseObject,
                 FileBasicInfo = ExAllocatePoolWithTag(NonPagedPool,
                                                       sizeof(*FileBasicInfo),
                                                       TAG_IO);
+                RtlZeroMemory(FileBasicInfo, sizeof(*FileBasicInfo));
```
SergeGautherie, line 1265: Maybe, but not at this line...

khanhnt2: Which line I should use RtlZeroMemory?

ThFabba: Inside the `if (FileBasicInfo)` block, since here it could be NULL.

khanhnt2: I changed, thanks.

ThFabba:
Hmm. So looking a bit at how Windows does this, it turns out this might be the responsibility of the filesystem driver.
E.g. https://git.reactos.org/?p=reactos.git;a=blob;f=drivers/filesystems/fastfat_new/fileinfo.c;hb=5f076e5558f585eb8428aa2b323eb2ed4d10bf00#l1083

This is probably something we could write a test for to verify...

khanhnt2:
> 
> 
> Hmm. So looking a bit at how Windows does this, it turns out this might be the responsibility of the filesystem driver.
> E.g. https://git.reactos.org/?p=reactos.git;a=blob;f=drivers/filesystems/fastfat_new/fileinfo.c;hb=5f076e5558f585eb8428aa2b323eb2ed4d10bf00#l1083
> 
> This is probably something we could write a test for to verify...

Can you explain your idea? I have an idea to verify structure alignment can happen but I can't determine whether it could lead to memory disclosure.

ThFabba:
> Can you explain your idea? I have an idea to verify structure alignment can happen but I can't determine whether it could lead to memory disclosure.

So the way to test this scenario is to trigger this code path by using `NtQueryAttributesFile` with a specially-crafted file system driver that e.g. asserts that the buffer contains random (in the simplest case: not-all-zero) data. This is not a 100% reliable test (because ExAllocatePoolWithTag _can_ return a buffer that's filled with all zeroes, that's just somewhat unlikely), but seeing it fail once would be enough. In any case, that test is rather involved.

I've done some research and found that indeed it is the file system driver's responsibility to zero the buffer here.
So let's make a deal: I'll create the test to prove this, and also add some instrumentation to our kernel that will help catch this FS bug. Since this will take me some time (on the order of months), you can assume this is the correct way forward and fix the bug in our FS drivers (fastfat, ntfs, npfs and msfs are the most likely suspects, but other filesystems like btrfs and ext2 may also have this bug, even though they were developed against Windows) instead of in the kernel.

SergeGautherie:
> how Windows does this

For what it's worth, to support some kind of zeroing fix,
https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-exallocatepoolwithtag
`Memory that ExAllocatePoolWithTag allocates is uninitialized. A kernel-mode driver must first zero this memory if it is going to make it visible to user-mode software (to avoid leaking potentially privileged contents).`

SergeGautherie:
Upstream synced codes need `#ifdef __REACTOS__`...

Extravert-ir:
I think I've missed that in previous changes, but shouldn't we user `RtlSecureZeroMemory` here?

drivers/filesystems/ext2/src/fileinfo.c
```diff
@@ -137,6 +137,7 @@ Ext2QueryFileInformation (IN PEXT2_IRP_CONTEXT IrpContext)
             }
 
             FileBasicInformation = (PFILE_BASIC_INFORMATION) Buffer;
+            RtlZeroBytes(FileBasicInformation, sizeof(FILE_BASIC_INFORMATION));
```
Extravert-ir, line 140: The `Buffer` is zeroed out already. Why zeroing for the second time is needed? If the error was triggered here, maybe we have a case where `Length` and `sizeof(FILE_BASIC_INFORMATION)` don't match?
Or `RtlZeroMemory(Buffer, Length)` just got optimized out?

Extravert-ir:
By the way, why do we care only about `FileBasicInformation`? Isn't it better to just erase a `SystemBuffer` in `QueryInformation` routine? Sounds like a more bug-prone solution

khanhnt2:
> 
> 
> By the way, why do we care only about `FileBasicInformation`? Isn't it better to just erase a `SystemBuffer` in `QueryInformation` routine? Sounds like a more bug-prone solution

I used to think about that. Some drivers clear `SystemBuffer` before using, other one clear buffer in specify function like this pull request. I'm still confuse which solution is better.

ThFabba:
> I think I've missed that in previous changes, but shouldn't we use `RtlSecureZeroMemory` here?

That should not be necessary. The Secure version makes sure the memory gets zeroed even when it appears unnecessary to the compiler due to optimizations. Here, the memory is accessed, so a compiler that tries to optimize away the zero would be broken.

drivers/filesystems/ext2/src/fileinfo.c
khanhnt2: I removed that redundant function.

SergeGautherie:
> Upstream synced codes need `#ifdef __REACTOS__`...

Why were these not added?

PR #2927, opened 2020-06-16
Add Restart in NT Native Mode Option to msgina

QuadlessPyjack:
## Purpose

This PR proposes the addition of a new shutdown dialogue action:
"Restart in NT Native mode" in order to allow booting into a future NT Native CMD shell.
[Demo video on YouTube](https://www.youtube.com/watch?v=fYkXU2tAgf4)

_This is my first PR for ReactOS_ - I did this because I've always been fascinated by the NT architecture and because I figured this would be a great way for me to get familiarised with OS development in general and the NT Native API in particular.

I was told that a draft PR is the best way to get feedback on my idea and code so, here I am :)
I figured developing a native NT shell is not a high priority (if there even are any plans for such a thing) so it would give me enough room to learn and experiment without affecting anyone.

Now, going forward - if having a native NT shell is something you might consider useful, I would like to ask for your feedback also on how it would be best for me to proceed: should I try to work on and expand TinyKrnl's Native Shell project or should I base my work on a different codebase or start completely from scratch?

## Proposed changes

- Added a new shutdown menu option in msgina and shell32 : _Restart in NT native mode_
- Added localised strings in Romanian (my native language) as well. English for the rest.
- Added registry manipulation logic in msgina that does the following after the user has opted to restart in NT native mode:
   - Checks in Session Manager if the native shell args are already written in BootExecute (if yes, it just reboots)
   - Writes the shell args required to boot into the native shell in BootExecute if not present

## TODO

~~Currently, to debug my registry manipulation code, I'm running it in a console app on Windows 10 - Is there a better way to do this? Asking because once every two reboots, something odd happens that causes Regedit to crash on load and I'm not sure how to debug this. On Windows 10, the registry changes look alright.~~ Fixed.
- ~~Review Coding Style and apply changes as needed.~~


dll/win32/msgina/shutdown.c
```diff
@@ -15,6 +15,7 @@
 #define WLX_SHUTDOWN_STATE_LOGOFF       0x01
 #define WLX_SHUTDOWN_STATE_POWER_OFF    0x02
 #define WLX_SHUTDOWN_STATE_REBOOT       0x04
+#define WLX_SHUTDOWN_STATE_NT_MODE      0x05
```
learn-more, line 18: This does not seem correct, since it is used as bitflag (See `return WLX_SHUTDOWN_STATE_POWER_OFF | WLX_SHUTDOWN_STATE_REBOOT | WLX_SHUTDOWN_STATE_NT_MODE;`)

dll/win32/msgina/shutdown.c
```diff
@@ -336,6 +348,17 @@ ShutdownOnInit(
             SendMessageW(hwndList, CB_SETITEMDATA, idx, WLX_SAS_ACTION_SHUTDOWN_REBOOT);
     }
 
+    /* Restart in NT Native Mode */
+    if (pContext->ShutdownOptions & WLX_SHUTDOWN_STATE_NT_MODE)
```
learn-more, line 352: This also responds to reboot and logoff now (see bitflag definition)

dll/win32/msgina/shutdown.c
```diff
@@ -538,6 +561,89 @@ ShutdownDialog(
...
+VOID SetRegistryToBootInNTNativeMode(VOID)
+{
+    // prepare the registry for booting into the NT Native Shell
+    LONG lRet;
+    HKEY hKey;
+    DWORD dwType = 0, dwCurrentValueSize = 0;
+    WCHAR *pszCurrentValue = NULL, *pszFinalValue = NULL;
+    UINT cchLatest = 0;
+    const UINT wcharSize = sizeof(WCHAR);
+#define SHELL_ARG_LENGTH 22
```
learn-more, line 580: 22 because?

dll/win32/msgina/shutdown.c
```diff
@@ -538,6 +561,89 @@ ShutdownDialog(
...
+ */
+VOID SetRegistryToBootInNTNativeMode(VOID)
+{
+    // prepare the registry for booting into the NT Native Shell
+    LONG lRet;
+    HKEY hKey;
+    DWORD dwType = 0, dwCurrentValueSize = 0;
+    WCHAR *pszCurrentValue = NULL, *pszFinalValue = NULL;
+    UINT cchLatest = 0;
+    const UINT wcharSize = sizeof(WCHAR);
```
learn-more, line 579: can just use `sizeof(WCHAR)` in your code, will be replaced with a const by the compiler.

dll/win32/msgina/shutdown.c
```diff
@@ -538,6 +561,89 @@ ShutdownDialog(
...
+        goto Cleanup;
+    }
+    memset(pszFinalValue, '\0', (cchLatest / wcharSize));
+
+    wcscpy(&pszFinalValue[0], &pszCurrentValue[0]);
+    lstrcat((LPWSTR)pszFinalValue, pszNativeShellArg);
+
+    // MULTI_SZ uses null terminators as string delimiters
+    *(pszFinalValue + uInitialStringLength) = '\0';
+    *(pszFinalValue + uInitialStringLength + SHELL_ARG_LENGTH - 2) = '\0';
```
learn-more, line 636: Please let the compiler handle stuff like this (and the magic 22 ;)
```
    WCHAR pszNativeShellArg1[] = L"0native Hello World!0";
    WCHAR pszNativeShellArg2[] = L"\0native Hello World!\0";

    printf("1: %u\n",_countof(pszNativeShellArg1));
    printf("2: %u\n",_countof(pszNativeShellArg2));
```
Will output:
```
1: 22
2: 22
```

dll/win32/msgina/shutdown.c
QuadlessPyjack: Damn! Rookie mistake right there. Fixing now.

JoachimHenze:
Thanks for your contribution. I still downvoted this PR, because I see no reasoning behind this.
Adding something "just for fun" does not seem to be a good idea in my eyes, because everything we add, requires to be maintained and groomed in the future.

QuadlessPyjack:
I think there 

> 
> 
> Thanks for your contribution. I still downvoted this PR, because I see no reasoning behind this.
> Adding something "just for fun" does not seem to be a good idea in my eyes, because everything we add, requires to be maintained and groomed in the future.

I think there's a miscommunication error on my part here - as stated in the PR description above and based on my question there, I want to develop this further, specifically, as a continuation, I'm interested in delving deeper into implementing a native NT shell.
Grooming and maintaining this implementation *is* the fun for me, otherwise I wouldn't be doing it :)

QuadlessPyjack:
Question, because I'm not entirely sure how the PR progression flow works here:
Is there anything else I need to do?

Are those thumbs up / thumbs down indicators just there for cosmetic value or do they actually serve a purpose in approval count?

JoseSanchezz:
I have mixed thoughts about this change: it's interesting and fun to play with but at the same time seems impractical to include. Considering how few people will use this feature vs how many people will use that shutdown dialog (everyone), I don't see why we would want that option there: it'll be distracting and confusing to the vast majority of users. Is there a use case for this that is common enough to justify the maintenance burden for us and extra attention cycles for users?

I think booting into native mode shell (or any other app) should be exposed only as a hidden feature for advanced users, which the existing BootExecute registry key already accomplishes well.

QuadlessPyjack:
I agree that the native mode shell is a power user feature but the idea behind this feature was to provide an easier, cleaner, integrated way of entering this mode.

Perhaps I can change this so that this mode is only shown if any of the following are true:
- User is an Administrator
- SKU is Server
- CTRL key was held down when the Shutdown button was pressed in the Start Menu (like how Windows 10 exposes access to the Boot Immersive Menu - bootim.exe).

dll/win32/msgina/shutdown.c
```diff
@@ -15,6 +15,7 @@
 #define WLX_SHUTDOWN_STATE_LOGOFF       0x01
 #define WLX_SHUTDOWN_STATE_POWER_OFF    0x02
 #define WLX_SHUTDOWN_STATE_REBOOT       0x04
+#define WLX_SHUTDOWN_STATE_NT_MODE      0x08
 // 0x08
```
ThFabba, line 19: You'd want to remove this comment, since you filled in the bit.

dll/win32/msgina/shutdown.c
```diff
@@ -538,6 +561,116 @@ ShutdownDialog(
...
+    {
+        TRACE("Unable to open the Session Manager key, error %d", GetLastError());
+        return;
+    }
+
+    ReturnCode = RegQueryValueEx(HiveKey, L"BootExecute", NULL, &RegValueType, NULL, &InitialRegValueSize);
+
+    if (ReturnCode != ERROR_SUCCESS || RegValueType != REG_MULTI_SZ)
+    {
+        TRACE("Unable to grab BootExecute, error %d", GetLastError());
```
ThFabba, line 604: Leaking `HiveKey`.

dll/win32/msgina/shutdown.c
```diff
@@ -538,6 +561,116 @@ ShutdownDialog(
...
+        goto Cleanup;
+    }
+
+    InitialRegValue = (WCHAR*)HeapAlloc(GetProcessHeap(), 0, InitialRegValueSize);
+    if (!InitialRegValue)
+    {
+        TRACE("HeapAlloc failed to allocate %d bytes", InitialRegValueSize);
+        return;
+    }
+    memset(InitialRegValue, '\0', InitialRegValueLength);
```
ThFabba, line 632: This needs a byte count, you're passing a character count (but you should really just use `HEAP_ZERO_MEMORY`)

dll/win32/msgina/shutdown.c
```diff
@@ -538,6 +561,116 @@ ShutdownDialog(
...
+        return;
+    }
+
+    const UINT InitialRegValueLength = InitialRegValueSize / sizeof(WCHAR);
+
+    /* BootExecute is empty (there should be at least autochk in there) - add our string and get out */
+    if (InitialRegValueSize == 0)
+    {
+        FinalRegValueSize = ShellArgsLength * sizeof(WCHAR) + sizeof(WCHAR);
+        FinalRegValue = (WCHAR*)HeapAlloc(GetProcessHeap(), 0, FinalRegValueSize);
```
ThFabba, line 614: Casting HeapAlloc's return value is unnecessary in C.

dll/win32/msgina/shutdown.c
```diff
@@ -538,6 +561,116 @@ ShutdownDialog(
...
+    ReturnCode = RegQueryValueExW(HiveKey, L"BootExecute", NULL, &RegValueType, (LPBYTE)InitialRegValue, &InitialRegValueSize);
+
+    FinalRegValueSize = InitialRegValueSize - sizeof(WCHAR) + ShellArgsLength * sizeof(WCHAR);
+
+    /* check if we haven't set the key already */
+    WCHAR* StringIt = InitialRegValue;
+    UINT CharCount = 0;
+    while (CharCount < InitialRegValueLength)
+    {
+        if (wcsstr(StringIt, (NativeShellArguments + 1)) != NULL)
```
ThFabba, line 643: `NativeShellArguments + 1` is `L"ative Hello World!"`. That doesn't seem like what you want.

dll/win32/msgina/shutdown.c
```diff
@@ -538,6 +561,116 @@ ShutdownDialog(
...
+        FinalRegValueSize = ShellArgsLength * sizeof(WCHAR) + sizeof(WCHAR);
+        FinalRegValue = (WCHAR*)HeapAlloc(GetProcessHeap(), 0, FinalRegValueSize);
+        if (!FinalRegValue)
+        {
+            TRACE("HeapAlloc failed to allocate %d bytes\n", FinalRegValueSize);
+            goto Cleanup;
+        }
+
+        memset(FinalRegValue, '\0', (FinalRegValueSize / sizeof(WCHAR)));
+        memcpy(FinalRegValue, NativeShellArguments, ShellArgsLength * sizeof(WCHAR));
```
ThFabba, line 622: What is this code for? You're jumping to `Cleanup` right after so the memory will be freed right away without doing anything with its contents.

dll/win32/msgina/shutdown.c
```diff
@@ -538,6 +561,116 @@ ShutdownDialog(
...
+        CharCount += CurrentStringLength != 0 ? CurrentStringLength : 1;
+    }
+
+    FinalRegValue = (WCHAR*)HeapAlloc(GetProcessHeap(), 0, FinalRegValueSize);
+    if (!FinalRegValue)
+    {
+        TRACE("HeapAlloc failed to allocate %d bytes\n", FinalRegValueSize);
+        goto Cleanup;
+    }
+    memset(FinalRegValue, '\0', (FinalRegValueSize / sizeof(WCHAR)));
```
ThFabba, line 659: Use `HEAP_ZERO_MEMORY` instead.

dll/win32/msgina/shutdown.c
```diff
@@ -538,6 +561,116 @@ ShutdownDialog(
...
+
+    ReturnCode = RegQueryValueEx(HiveKey, L"BootExecute", NULL, &RegValueType, NULL, &InitialRegValueSize);
+
+    if (ReturnCode != ERROR_SUCCESS || RegValueType != REG_MULTI_SZ)
+    {
+        TRACE("Unable to grab BootExecute, error %d", GetLastError());
+        return;
+    }
+
+    const UINT InitialRegValueLength = InitialRegValueSize / sizeof(WCHAR);
```
ThFabba, line 608: I don't think `Length` vs `Size` conveys the difference well. I'd suggest either:
* Use Hungarian notation (`cb` vs `cch` is one of the few cases where that's actually useful) -- "the Win32 way"
* Use a descriptive name, e.g. `InitialRegValueBytes / InitialRegValueChars` or `InitialRegValueByteCount / InitialRegValueCharCount` etc.
* Keep all variables in bytes, divide by `sizeof(WCHAR)` where needed -- "the NT way"

dll/win32/msgina/resource.h
```diff
@@ -99,9 +99,11 @@
...
 #define IDS_SHUTDOWN_SLEEP            50003
 #define IDS_SHUTDOWN_HIBERNATE        50004
+#define IDS_SHUTDOWN_NT_MODE          50005
 /* Shut down descriptions */
-#define IDS_SHUTDOWN_SHUTDOWN_DESC    50005
-#define IDS_SHUTDOWN_LOGOFF_DESC      50006
-#define IDS_SHUTDOWN_RESTART_DESC     50007
-#define IDS_SHUTDOWN_SLEEP_DESC       50008
-#define IDS_SHUTDOWN_HIBERNATE_DESC   50009
+#define IDS_SHUTDOWN_SHUTDOWN_DESC    50006
```
ThFabba, line 104: Do these resource ids match up with Windows? If so, you'd want to avoid moving them around and instead add your new one at the end.

dll/win32/msgina/shutdown.c
```diff
@@ -336,6 +348,17 @@ ShutdownOnInit(
...
+    {
+        LoadStringW(pgContext->hDllInstance, IDS_SHUTDOWN_NT_MODE, szBuffer, _countof(szBuffer));
+        idx = SendMessageW(hwndList, CB_ADDSTRING, 0, (LPARAM)szBuffer);
+        if (idx != CB_ERR)
+        {
+            SendMessageW(hwndList, CB_SETITEMDATA, idx, WLX_SAS_ACTION_SHUTDOWN_NT_MODE);
+        }
+    }
+
     // if (pContext->ShutdownOptions & 0x08) {}
```
ThFabba, line 362: This is you. Remove please.

dll/win32/msgina/shutdown.c
```diff
@@ -15,6 +15,7 @@
 #define WLX_SHUTDOWN_STATE_LOGOFF       0x01
 #define WLX_SHUTDOWN_STATE_POWER_OFF    0x02
 #define WLX_SHUTDOWN_STATE_REBOOT       0x04
+#define WLX_SHUTDOWN_STATE_NT_MODE      0x08
```
ThFabba, line 18: It kind of sounds like these values are matching Windows functionality and 0x08/0x20/0x80 exist but we don't implement them. In that case it might be better to add the new flag at the end (0x100), if the data type used to store these is large enough.

dll/win32/msgina/shutdown.c
```diff
@@ -538,6 +561,116 @@ ShutdownDialog(
...
+/*************************************************************************
+ * SetRegistryToBootInNTNativeMode
+ *
+ * NOTES
+ *     Used to append the NT Native shell start-up arguments to the
+ *     BootExecute registry value so that the system will boot into it.
+ */
+VOID SetRegistryToBootInNTNativeMode(VOID)
+{
+    #if 1
```
ThFabba, line 573: The dead code is going to confuse compilers & static analyzers. You could e.g. `#if 0` out the contents of the function instead.

dll/win32/msgina/shutdown.c
QuadlessPyjack: Replaced return with `goto Cleanup` statement.

dll/win32/msgina/shutdown.c
QuadlessPyjack: Inverted #if logic. Ideally I could add a check in GetAllowedShutdownOptions() to check if the user is an administrator and/or the CTRL key is pressed but that probably warrants its own separate discussion I think.

dll/win32/msgina/shutdown.c
QuadlessPyjack: Moved to 0x100

dll/win32/msgina/shutdown.c
QuadlessPyjack: N/A due to SAS_ACTION value changed to 0x100 (see comment below)

dll/win32/msgina/resource.h
QuadlessPyjack: Reverted IDS values to originals and moved the new entries last. Sorry about that.

dll/win32/msgina/shutdown.c
QuadlessPyjack: Replaced suffixes: Length --> CharCount | Size --> ByteCount

dll/win32/msgina/shutdown.c
QuadlessPyjack: N/A. Changed #define value to 0x100

PR #2928, opened 2020-06-16
[FASTFAT_NEW] Fix GCC 8.4 build, Refactor ReactOS workarounds

SergeGautherie:


drivers/filesystems/fastfat_new/fatdata.h
```diff
@@ -263,11 +263,19 @@ extern LONG FatDebugTraceIndent;
 }
 #endif
 
+#ifndef __REACTOS__
+#define DebugUnwind(X) {                                                      \
+    if (AbnormalTermination()) {                                              \
```
HBelusca, line 268: Then, following the trend chosen in this PR, why not having instead some sort of
```c
#define AbnormalTermination() _SEH2_AbnormalTermination()
```
somewhere?

SergeGautherie: > initial import commit

I had a look: it will be worth a separate PR.

> `#define AbnormalTermination()`

I don't know how to do that for all SEH2 parts, especially `_SEH2_END`,
so better leave those parts "forked".
I reverted my related change here.

drivers/filesystems/fastfat_new/fatprocs.h
```diff
@@ -40,6 +40,13 @@ Module Name:
 
 
 #ifdef __REACTOS__
+// Downgrade unsupported NT6.2+ features.
+#define MdlMappingNoExecute 0
+#define NonPagedPoolNx NonPagedPool
+#define NonPagedPoolNxCacheAligned NonPagedPoolCacheAligned
+#define POOL_NX_ALLOCATION 0
```
ThFabba, line 47: Can you `#undef POOL_NX_ALLOCATION` (and `MdlMappingNoExecute`) please, so that it doesn't break once our DDK headers get these defines?

drivers/filesystems/fastfat_new/fatprocs.h
```diff
@@ -40,6 +40,13 @@ Module Name:
...
+
+// Moved upper: needed in 'fatstruc.h'.
```
ThFabba, line 49: Nit: Moved up*

drivers/filesystems/fastfat_new/fatprocs.h
SergeGautherie: Added `#if`.

drivers/filesystems/fastfat_new/fatprocs.h
SergeGautherie: Fixed.

drivers/filesystems/fastfat_new/fatdata.h
```diff
@@ -234,7 +234,7 @@ extern LONG FatDebugTraceIndent;
     }                                                       \
 }
 
-#ifdef _MSC_VER
+#if !defined(__REACTOS__) || defined(_MSC_VER)
```
Extravert-ir, line 237: I don't get the idea here. Do we have this warning disabled already?

SergeGautherie: The idea is simply to be explicit about code that does (not) exist upstream.
I don't understand your question. Behavior is unchanged.

drivers/filesystems/fastfat_new/fatprocs.h
Extravert-ir: Should be `#undef`, not `#if`

SergeGautherie: I saw the request, but I don't know/understand why.
What is wrong with `#if NTDDI_VERSION < NTDDI_WIN8`?
What would be wrong with `#ifndef ...`?

ThFabba: ReactOS will add these defines and enum members to its DDK headers sooner or later, to be compatible with the real WDK. The point was to change this code such that it wouldn't break at that point. Your change doesn't achieve that.

SergeGautherie: > sooner or later

Now, before this PR, may be the right time?

NB:
I don't have WDK 8+, so can't look at that myself.

> this driver would use the real values and compile but stop working

Ah, you plan to update the PSDK in a way ROS Core will not support: I couldn't guess that.

SergeGautherie: > match the latest public version

I understand and agree with that, though it may be a bit tricky, as NT6.2+ MS headers do not support NT5.2 anymore, do they?
Anyway, no point debating this here: new headers were not added, PR is merged now.

> build things written against MS headers

It seems this PR/driver is such a case, and well...

> compile & behave correctly when built against MS's headers

Does ROS build system have an option to do that? It would help working on that goal.

PR #2930, opened 2020-06-17
[PSDK][KERNEL32] Export functions needed for Syzkaller

Freakston:
This Pull Request has the required modifications to run Syzkaller.

### DIFF
--- a/sdk/include/psdk/winbase.h
+++ b/sdk/include/psdk/winbase.h
@@ -1362,11 +1362,11 @@ typedef enum _COMPUTER_NAME_FORMAT {
 	ComputerNamePhysicalDnsFullyQualified,
 	ComputerNameMax
 } COMPUTER_NAME_FORMAT;
-#endif /* (_WIN32_WINNT >= 0x0500) */
-
-#if (_WIN32_WINNT >= 0x0600)
+ /* (_WIN32_WINNT >= 0x0500) */
+//Modified by Freakston for Syzkaller
 typedef RTL_SRWLOCK SRWLOCK, *PSRWLOCK;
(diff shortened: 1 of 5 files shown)


SergeGautherie:
Wow! NT6 functions in NT5.2 `kernel32`?
Should use `kernel32_vista` instead?

Freakston:
> Wow! NT6 functions in NT5.2 `kernel32`?
> Should use `kernel32_vista` instead?

That isn't working as expected and this particular build will only be used for fuzzing purposes. @SergeGautherie

sdk/include/psdk/winbase.h
```diff
@@ -1362,11 +1362,11 @@ typedef enum _COMPUTER_NAME_FORMAT {
 	ComputerNamePhysicalDnsFullyQualified,
 	ComputerNameMax
 } COMPUTER_NAME_FORMAT;
-#endif /* (_WIN32_WINNT >= 0x0500) */
-
-#if (_WIN32_WINNT >= 0x0600)
+ /* (_WIN32_WINNT >= 0x0500) */
+//Modified by Freakston for Syzkaller
```
ThFabba, line 1366: If you're building it within our build system... could you just define _WIN32_WINNT to be a higher version and link to kernel32_vista?

SergeGautherie:
> this particular build

Ah, this PR is not meant to be committed?
Then, you should be more explicit about its actual purpose and mark it as draft?

binarymaster:
> Ah, this PR is not meant to be committed?

You probably didn't noticed that this PR is special GSoC PR:
image

SergeGautherie:
> special GSoC PR

What difference does that make?

PR #2931, opened 2020-06-18
[SDK][SVCHOST] Move all global svchost declarations to public header

oleg-dubinskiy:
## Purpose

Move all svchost declarations which should be global, from private svchost.h header to public svc.h header.
It will allow to use those declarations globally where they are required (e. g. for SvchostPushServiceGlobals functions in some services: https://www.geoffchappell.com/studies/windows/win32/services/svchost/dll/svchostpushserviceglobals.htm?tx=76).
That public header probably will be ROS-specific, since MS did not document it, see https://www.geoffchappell.com/studies/windows/win32/services/svchost/process/globaldata.htm?tx=78 instead.
No functional changes.

JIRA issue: None.

## Proposed changes

- Create a public sdk/include/reactos/svc.h header;
- Move there all declarations which should be public, from private base/services/svchost/svchost.h;
- Update the includes in svchost.h accordingly. Move all previous includes to the public header, since they are required by global svchost declarations. But they are still used by private declarations also, because I included the public header there too.

### DIFF
--- a/base/services/svchost/svchost.h
+++ b/base/services/svchost/svchost.h
@@ -15,11 +15,9 @@
 #include <rpc.h>
 #include <ndk/rtlfuncs.h>
 #include <ndk/kdtypes.h>
+#include <dpfilter.h>
 
-//
-// FIXME: Should go in public headers
-//
-#define DPFLTR_SVCHOST_ID 28
+#include <svc.h>
 
 //
 // This prints out a SVCHOST-specific debug print, with the PID/TID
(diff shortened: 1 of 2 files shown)


ThFabba:
A header not present in Microsoft's SDK shouldn't live in `psdk`. Can you move it to `sdk/include/reactos` please?

oleg-dubinskiy:
Fixed @ThFabba. Thanks. Didn't know about that.
Compilation still succeeds for me.

sdk/include/reactos/svc.h
```diff
@@ -0,0 +1,60 @@
...
+/* See https://www.geoffchappell.com/studies/windows/win32/services/svchost/process/globaldata.htm?tx=78 */
+
+#ifndef __SVC_H
+#define __SVC_H
+
+#include <rpc.h>
+#include <ndk/rtlfuncs.h>
+#include <ndk/kdtypes.h>
+
+#define DPFLTR_SVCHOST_ID 28
```
HBelusca, line 17: So that one should not necessary, because it is already declared in psdk/dpfilter.h. (But then our svchost.h should include that latter header...)

oleg-dubinskiy: Fixed. Thanks @HBelusca. I missed it.
This will be indeed better, because the public header doesn't use that declaration anywhere, only private header does it, so it makes sense to include dpfilter.h only there.
Compilation still succeeds for me.

sdk/include/reactos/svc.h
```diff
@@ -0,0 +1,58 @@
...
+
+#include <rpc.h>
+#include <ndk/rtlfuncs.h>
+#include <ndk/kdtypes.h>
+
+//
+// This is the callback that a hosted service can register for stop notification
+//
+typedef VOID
+    (CALLBACK *PSVCHOST_STOP_CALLBACK) (
```
HBelusca, line 21:
```suggestion
(CALLBACK *PSVCHOST_STOP_CALLBACK) (
```

sdk/include/reactos/svc.h
```diff
@@ -0,0 +1,58 @@
...
+ * PROGRAMMERS: ReactOS Portable Systems Group
+ */
+/* See https://www.geoffchappell.com/studies/windows/win32/services/svchost/process/globaldata.htm?tx=78 */
+
+#ifndef __SVC_H
+#define __SVC_H
+
+#include <rpc.h>
+#include <ndk/rtlfuncs.h>
+#include <ndk/kdtypes.h>
```
HBelusca, line 15: The generic rule is that shared headers for interfaces should not include "unrelated" headers.
Here, rpc.h and the ndk headers were included in svchost because the latter uses also functions and types defined in these headers for its own implementation.
However an unrelated service that implements an interface to svchost (by including this reactos/svc.h) may not need at all rpc.h and these ndk headers.
This is why, these headers may remain instead in svchost.

sdk/include/reactos/svc.h
oleg-dubinskiy: Fixed.

sdk/include/reactos/svc.h
oleg-dubinskiy: Fixed. Thanks @HBelusca. :+1: You're right here too.
Moreover, the includes ordering is important too, because when I included svc.h before rpc.h, I've got the compilation error! :worried: . Only after including it **after** rpc.h, the build succeeded for me again. I wonder why it happens?... :thinking:

base/services/svchost/svchost.h
```diff
@@ -12,15 +12,12 @@
 #define WIN32_NO_STATUS
 #define WIN32_LEAN_AND_MEAN
 
+#include <dpfilter.h>
 #include <rpc.h>
+#include <svc.h>
 #include <ndk/rtlfuncs.h>
 #include <ndk/kdtypes.h>
```
HBelusca, line 19: Just keep the original ordering:
```suggestion
#include <rpc.h>
#include <ndk/rtlfuncs.h>
#include <ndk/kdtypes.h>
#include <dpfilter.h>

#include <svc.h>
```

sdk/include/reactos/svc.h
HBelusca: Because maybe rpc.h includes itself other headers that define stuff that svc.h uses.
The usual rule is:
1. include the standard NDK + PSDK headers.
2. then include the "utilities" headers (e.g. strsafe.h, ...) and the "interfaces" headers (like that svc.h)

base/services/svchost/svchost.h
oleg-dubinskiy: Fixed, thanks.

PR #2933, opened 2020-06-18
[WINMM] Improve implementation of timeGetTime, timeBeginPeriod, timeEndPeriod

Tom-III:
## Purpose

Improve Windows Multimedia timer implementations.  Many games that use a frame-limiter to keep their framerate locked rely on these functions giving millisecond accurate timing readings.  The previous implementation of wrapping GetTickCount only offers accuracy of tens of milliseconds leading to unstable or slow framerates and gameplay.

JIRA issue: [CORE-17121](https://jira.reactos.org/browse/CORE-17121)

## Proposed changes

- Improve timeGetTime by using the more accurate QueryPerformanceCounter function when a high-resolution timer was requested.
- timeBeginPeriod with a resolution of 5ms or less will switch timeGetTime from using GetTickCount to QueryPerformanceCounter.
- timeEndPeriod with a resolution of 5ms or less will switch timeGetTime back to GetTickCount.

### DIFF
--- a/dll/win32/winmm/time.c
+++ b/dll/win32/winmm/time.c
@@ -423,12 +423,14 @@ MMRESULT WINAPI timeGetDevCaps(LPTIMECAPS lpCaps, UINT wSize)
 MMRESULT WINAPI timeBeginPeriod(UINT wPeriod)
 {
     if (wPeriod < MMSYSTIME_MININTERVAL || wPeriod > MMSYSTIME_MAXINTERVAL)
-		return TIMERR_NOCANDO;
+        return TIMERR_NOCANDO;
 
     /*  High resolution timer requested, use QPC */
-    if (wPeriod <= 5 && TIME_qpcFreq.QuadPart == 0) 
+    if (wPeriod <= 5 && TIME_qpcFreq.QuadPart == 0)
+    {
         if(QueryPerformanceFrequency(&TIME_qpcFreq))
             TIME_qpcFreq.QuadPart /= 1000;
+    }
 
     if (wPeriod > MMSYSTIME_MININTERVAL)
     {
@@ -444,7 +446,7 @@ MMRESULT WINAPI timeBeginPeriod(UINT wPeriod)
 MMRESULT WINAPI timeEndPeriod(UINT wPeriod)
 {
     if (wPeriod < MMSYSTIME_MININTERVAL || wPeriod > MMSYSTIME_MAXINTERVAL)
-		return TIMERR_NOCANDO;
+        return TIMERR_NOCANDO;
 
     /*  High resolution timer no longer requested, stop using QPC */
     if (wPeriod <= 5 && TIME_qpcFreq.QuadPart != 0)
@@ -463,7 +465,7 @@ MMRESULT WINAPI timeEndPeriod(UINT wPeriod)
  */
 DWORD WINAPI timeGetTime(void)
 {
-	LARGE_INTEGER perfCount;
+    LARGE_INTEGER perfCount;
 #if defined(COMMENTOUTPRIORTODELETING)
     DWORD       count;
 
@@ -474,11 +476,11 @@ DWORD WINAPI timeGetTime(void)
     if (pFnRestoreThunkLock) pFnRestoreThunkLock(count);
 #endif
     /* Use QPC if a high-resolution timer was requested (<= 5ms) */
-    if (TIME_qpcFreq.QuadPart != 0) 
+    if (TIME_qpcFreq.QuadPart != 0)
     {
         QueryPerformanceCounter(&perfCount);
         return (DWORD)(perfCount.QuadPart / TIME_qpcFreq.QuadPart);
     }
-	/* Otherwise continue using GetTickCount */
+    /* Otherwise continue using GetTickCount */
     return GetTickCount();
 }
--- a/dll/win32/winmm/time.c
+++ b/dll/win32/winmm/time.c
@@ -47,6 +47,7 @@ static    LPWINE_TIMERENTRY 	TIME_TimersList;
 static    HANDLE                TIME_hKillEvent;
 static    HANDLE                TIME_hWakeEvent;
 static    BOOL                  TIME_TimeToDie = TRUE;
+static    LARGE_INTEGER         TIME_qpcFreq;
 
 /*
  * Some observations on the behavior of winmm on Windows.
@@ -422,7 +423,12 @@ MMRESULT WINAPI timeGetDevCaps(LPTIMECAPS lpCaps, UINT wSize)
 MMRESULT WINAPI timeBeginPeriod(UINT wPeriod)
 {
     if (wPeriod < MMSYSTIME_MININTERVAL || wPeriod > MMSYSTIME_MAXINTERVAL)
-	return TIMERR_NOCANDO;
+		return TIMERR_NOCANDO;
+
+    /*  High resolution timer requested, use QPC */
+    if (wPeriod <= 5 && TIME_qpcFreq.QuadPart == 0) 
+        if(QueryPerformanceFrequency(&TIME_qpcFreq))
+            TIME_qpcFreq.QuadPart /= 1000;
 
     if (wPeriod > MMSYSTIME_MININTERVAL)
     {
@@ -438,7 +444,11 @@ MMRESULT WINAPI timeBeginPeriod(UINT wPeriod)
 MMRESULT WINAPI timeEndPeriod(UINT wPeriod)
 {
     if (wPeriod < MMSYSTIME_MININTERVAL || wPeriod > MMSYSTIME_MAXINTERVAL)
-	return TIMERR_NOCANDO;
+		return TIMERR_NOCANDO;
+
+    /*  High resolution timer no longer requested, stop using QPC */
+    if (wPeriod <= 5 && TIME_qpcFreq.QuadPart != 0)
+        TIME_qpcFreq.QuadPart = 0;
 
     if (wPeriod > MMSYSTIME_MININTERVAL)
     {
@@ -453,6 +463,7 @@ MMRESULT WINAPI timeEndPeriod(UINT wPeriod)
  */
 DWORD WINAPI timeGetTime(void)
 {
+	LARGE_INTEGER perfCount;
 #if defined(COMMENTOUTPRIORTODELETING)
     DWORD       count;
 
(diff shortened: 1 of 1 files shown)


dll/win32/winmm/time.c
```diff
@@ -422,7 +423,15 @@ MMRESULT WINAPI timeGetDevCaps(LPTIMECAPS lpCaps, UINT wSize)
...
     if (wPeriod < MMSYSTIME_MININTERVAL || wPeriod > MMSYSTIME_MAXINTERVAL)
-	return TIMERR_NOCANDO;
+		return TIMERR_NOCANDO;
+
+	/*  High resolution timer requested, use QPC */
+	if (wPeriod <= 5 && TIME_qpcFreq.QuadPart == 0) {
+		if(QueryPerformanceFrequency(&TIME_qpcFreq))
+			TIME_qpcFreq.QuadPart /= 1000;
+		else
+			TIME_qpcFreq.QuadPart == 0;
```
Extravert-ir, line 433: This statement does not do anything. Typo?

dll/win32/winmm/time.c
```diff
@@ -438,7 +447,11 @@ MMRESULT WINAPI timeBeginPeriod(UINT wPeriod)
 MMRESULT WINAPI timeEndPeriod(UINT wPeriod)
 {
     if (wPeriod < MMSYSTIME_MININTERVAL || wPeriod > MMSYSTIME_MAXINTERVAL)
-	return TIMERR_NOCANDO;
+		return TIMERR_NOCANDO;
+
+	/*  High resolution timer request finished, stop using QPC */
+	if (wPeriod <= 5 && TIME_qpcFreq.QuadPart != 0)
+		TIME_qpcFreq.QuadPart = 0;
```
Extravert-ir, line 454: Do we need to check `wPeriod` here?

Extravert-ir: Actually, `QueryPerformanceFrequency` result does not change across reboots so I don't see a reason to clear it

dll/win32/winmm/time.c
```diff
@@ -422,7 +423,15 @@ MMRESULT WINAPI timeGetDevCaps(LPTIMECAPS lpCaps, UINT wSize)
 MMRESULT WINAPI timeBeginPeriod(UINT wPeriod)
 {
     if (wPeriod < MMSYSTIME_MININTERVAL || wPeriod > MMSYSTIME_MAXINTERVAL)
-	return TIMERR_NOCANDO;
+		return TIMERR_NOCANDO;
+
+	/*  High resolution timer requested, use QPC */
+	if (wPeriod <= 5 && TIME_qpcFreq.QuadPart == 0) {
```
ThFabba, line 429: Why 5 and/or what happens between 2 and 4?

dll/win32/winmm/time.c
```diff
@@ -422,7 +423,15 @@ MMRESULT WINAPI timeGetDevCaps(LPTIMECAPS lpCaps, UINT wSize)
...
+		if(QueryPerformanceFrequency(&TIME_qpcFreq))
+			TIME_qpcFreq.QuadPart /= 1000;
```
ThFabba, line 431: So the value calculated here doesn't depend on wPeriod. Should it?

dll/win32/winmm/time.c
ThFabba: Technically there should be a stack of periods that we remember, and we only set it back to zero if the stack is empty. But this seems fine to start with.

dll/win32/winmm/time.c
```diff
@@ -462,6 +476,11 @@ DWORD WINAPI timeGetTime(void)
     if (pFnReleaseThunkLock) pFnReleaseThunkLock(&count);
     if (pFnRestoreThunkLock) pFnRestoreThunkLock(count);
 #endif
-
+	/* Use QPC if a high-resolution timer is requested (<= 5ms) */
+	if (TIME_qpcFreq.QuadPart != 0) {
+		QueryPerformanceCounter(&perfCount);
```
kernelbin, line 481: normally ```QueryPerformanceCounter``` and ```QueryPerformanceFrequency``` won't fail.

dll/win32/winmm/time.c
Tom-III: QueryPerformanceFrequency is used both to cache the frequency and as a flag to indicate weather QueryPerformanceFrequency should be used or GetTickCount, if you set it to 0 it reverts back to GetTickCount.

dll/win32/winmm/time.c
Tom-III: Because the default precision on NT is 5 milliseconds or more according to the documentation for timeGetTime, meaning any time an application requests a timer lower than that it should switch to the high-precision timer.

dll/win32/winmm/time.c
Tom-III: A timer value less than 5 milliseconds (the default on NT) is a request for a high-accuracy timer.  This commit is meant to be an improvement over simply wrapping GetTickCount.

dll/win32/winmm/time.c
Tom-III: Additionally the official timeGetTime affects the system scheduler and the system as a whole, whereas this patch is application local.

dll/win32/winmm/time.c
Tom-III: I see, since there already was a if 0 check above, fixed it.

dll/win32/winmm/time.c
Tom-III: Only one check is actually needed (at line 430) since if QPF fails TIME_qpcFreq.QuadPart will remain 0 and thus this code will be skipped preventing the need for a second check.

dll/win32/winmm/time.c
```diff
@@ -422,7 +423,12 @@ MMRESULT WINAPI timeGetDevCaps(LPTIMECAPS lpCaps, UINT wSize)
 MMRESULT WINAPI timeBeginPeriod(UINT wPeriod)
 {
     if (wPeriod < MMSYSTIME_MININTERVAL || wPeriod > MMSYSTIME_MAXINTERVAL)
-	return TIMERR_NOCANDO;
+		return TIMERR_NOCANDO;
+
+    /*  High resolution timer requested, use QPC */
+    if (wPeriod <= 5 && TIME_qpcFreq.QuadPart == 0) 
+        if(QueryPerformanceFrequency(&TIME_qpcFreq))
+            TIME_qpcFreq.QuadPart /= 1000;
```
HBelusca, line 431:
```suggestion
    if (wPeriod <= 5 && TIME_qpcFreq.QuadPart == 0)
    {
        if (QueryPerformanceFrequency(&TIME_qpcFreq))
            TIME_qpcFreq.QuadPart /= 1000;
    }
```

dll/win32/winmm/time.c
```diff
@@ -453,6 +463,7 @@ MMRESULT WINAPI timeEndPeriod(UINT wPeriod)
  */
 DWORD WINAPI timeGetTime(void)
 {
+	LARGE_INTEGER perfCount;
```
HBelusca, line 466:
```suggestion
    LARGE_INTEGER perfCount;
```

dll/win32/winmm/time.c
```diff
@@ -462,6 +473,12 @@ DWORD WINAPI timeGetTime(void)
     if (pFnReleaseThunkLock) pFnReleaseThunkLock(&count);
     if (pFnRestoreThunkLock) pFnRestoreThunkLock(count);
 #endif
-
+    /* Use QPC if a high-resolution timer was requested (<= 5ms) */
+    if (TIME_qpcFreq.QuadPart != 0)
```
HBelusca, line 477:
```suggestion
    if (TIME_qpcFreq.QuadPart != 0)
```

dll/win32/winmm/time.c
```diff
@@ -462,6 +473,12 @@ DWORD WINAPI timeGetTime(void)
...
     if (pFnRestoreThunkLock) pFnRestoreThunkLock(count);
 #endif
-
+    /* Use QPC if a high-resolution timer was requested (<= 5ms) */
+    if (TIME_qpcFreq.QuadPart != 0) 
+    {
+        QueryPerformanceCounter(&perfCount);
+        return (DWORD)(perfCount.QuadPart / TIME_qpcFreq.QuadPart);
+    }
+	/* Otherwise continue using GetTickCount */
```
HBelusca, line 482:
```suggestion
    /* Otherwise continue using GetTickCount */
```

dll/win32/winmm/time.c
```diff
@@ -438,7 +444,11 @@ MMRESULT WINAPI timeBeginPeriod(UINT wPeriod)
 MMRESULT WINAPI timeEndPeriod(UINT wPeriod)
 {
     if (wPeriod < MMSYSTIME_MININTERVAL || wPeriod > MMSYSTIME_MAXINTERVAL)
-	return TIMERR_NOCANDO;
+		return TIMERR_NOCANDO;
```
HBelusca, line 447:
```suggestion
        return TIMERR_NOCANDO;
```

dll/win32/winmm/time.c
```diff
@@ -422,7 +423,12 @@ MMRESULT WINAPI timeGetDevCaps(LPTIMECAPS lpCaps, UINT wSize)
 MMRESULT WINAPI timeBeginPeriod(UINT wPeriod)
 {
     if (wPeriod < MMSYSTIME_MININTERVAL || wPeriod > MMSYSTIME_MAXINTERVAL)
-	return TIMERR_NOCANDO;
+		return TIMERR_NOCANDO;
```
HBelusca, line 426:
```suggestion
        return TIMERR_NOCANDO;
```

dll/win32/winmm/time.c
```diff
@@ -422,7 +425,16 @@ MMRESULT WINAPI timeGetDevCaps(LPTIMECAPS lpCaps, UINT wSize)
...
 {
     if (wPeriod < MMSYSTIME_MININTERVAL || wPeriod > MMSYSTIME_MAXINTERVAL)
-	return TIMERR_NOCANDO;
+        return TIMERR_NOCANDO;
+
+#ifdef __REACTOS__
+    /*  High resolution timer requested, use QPC */
+    if (wPeriod <= 5 && TIME_qpcFreq.QuadPart == 0)
+    {
+        if(QueryPerformanceFrequency(&TIME_qpcFreq))
```
HBelusca, line 434:
```suggestion
        if (QueryPerformanceFrequency(&TIME_qpcFreq))
```

PR #2934, opened 2020-06-19
[NTOS:IO] Move device node functions from pnpmgr.c to devnode.c

Extravert-ir:

### DIFF
--- a/ntoskrnl/io/pnpmgr/devnode.c
+++ b/ntoskrnl/io/pnpmgr/devnode.c
@@ -1,0 +1,400 @@
+/*
+ * PROJECT:     ReactOS Kernel
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     PnP manager device tree functions
+ * COPYRIGHT:   Casper S. Hornstrup (chorns@users.sourceforge.net)
+ *              2007 Hervé Poussineau (hpoussin@reactos.org)
+ *              2010-2012 Cameron Gutman (cameron.gutman@reactos.org)
+ */
+
+/* INCLUDES ******************************************************************/
+
+#include <ntoskrnl.h>
+#define NDEBUG
+#include <debug.h>
+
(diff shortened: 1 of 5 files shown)


ntoskrnl/io/pnpmgr/devnode.c
```diff
@@ -0,0 +1,396 @@
...
+ *              2007 Hervé Poussineau (hpoussin@reactos.org)
+ *              2010-2012 Cameron Gutman (cameron.gutman@reactos.org)
+ */
+
+
+#include <ntoskrnl.h>
+#define NDEBUG
+#include <debug.h>
+
+PDEVICE_NODE IopRootDeviceNode;
```
ThFabba, line 15: Wow, kernel source code looks weird without the `/* FUNCTIONS ***********/` thing :D

Extravert-ir: If we want this back, good time to do it :D
But honestly I haven't made my mind about these headers

ThFabba: Yeah I guess I'd prefer adding these comments here for consistency with the rest of the kernel. Less surprise when reading the code = better ;)

Extravert-ir: What is the difference between `GLOBALS` and `DATA`?

ThFabba: I don't think there is a difference. GLOBALS seems more common in ntoskrnl *shrug*

ntoskrnl/io/pnpmgr/devnode.c
```diff
@@ -0,0 +1,400 @@
+/*
```
ThFabba, line 1: Diff of "just move all the functions to this file" -> what's actually in the file:
https://gist.github.com/ThFabba/1cedf75456857db56d94cd99c05081f8

Extravert-ir: I indicated this in the commit message!

PR #2936, opened 2020-06-20
[BOOTVID] Fix drawing strings on top of the bitmap

disean:
## Purpose

 Fix drawing strings on top of the bitmap.

JIRA issue: [CORE-15896](https://jira.reactos.org/browse/CORE-15896)

## Proposed changes

- Fix VGA pixel drawing
- Minor improvements
  - Annotate some functions to avoid confusing
  - Fix BitBlt behavior
  - PC-98
    - Improve text scrolling performance
    - Fix failure handling
    - Reduce memory mapping that's not needed
    - Fix screen buffer copying code

### DIFF
--- a/drivers/base/bootvid/common.c
+++ b/drivers/base/bootvid/common.c
@@ -55,9 +55,9 @@ BitBlt(
     _In_ ULONG BitsPerPixel,
     _In_ ULONG Delta)
 {
-    ULONG sx, dx, dy;
-    UCHAR color;
-    ULONG offset = 0;
+    ULONG X, Y, Pixel;
+    UCHAR Colors;
+    PUCHAR InputBuffer;
     const ULONG Bottom = Top + Height;
     const ULONG Right = Left + Width;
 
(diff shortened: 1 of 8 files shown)


drivers/base/bootvid/i386/xbox/bootvid.c
```diff
@@ -430,7 +430,7 @@ VidScreenToBufferBlt(
         PUCHAR Buf = Buffer + y * Delta;
 
         /* Start the X inner loop */
-        for (ULONG x = 0; x < Width; x += 2)
+        for (ULONG x = 0; x < Width + 1; x += sizeof(USHORT))
```
HBelusca, line 433: Why is the loop going from 0 up to "Width" included ? Shouldn't it be from 0 to Width-1 included (aka. to Width excluded) ?

drivers/base/bootvid/i386/pc98/bootvid.c
```diff
@@ -421,16 +423,17 @@ VidScreenToBufferBlt(
...
     /* Clear the destination buffer */
     RtlZeroMemory(Buffer, Delta * Height);
 
     for (Y = 0; Y < Height; Y++)
     {
         OutputBuffer = Buffer + Y * Delta;
+        PixelsPosition = (PUSHORT)(FrameBuffer + FB_OFFSET(Left, Top + Y));
 
-        for (X = 0; X < Width; X += 2)
+        for (X = 0; X < Width + 1; X += sizeof(USHORT))
```
HBelusca, line 436: Why is the loop going from 0 up to "Width" included ? Shouldn't it be from 0 to Width-1 included (aka. to Width excluded) ?

drivers/base/bootvid/i386/xbox/bootvid.c
disean: Because the loop definrd inside function iterates from `0` to `Width+1` https://git.reactos.org/?p=reactos.git;a=blob;f=drivers/base/bootvid/common.c;h=7979ea42a0481fed85baf98c6a80791ba44a0aac;hb=dd6eb40381be1a090e4130e6d5cdd9fea141aac9#l84

HBelusca: Well, it's obviously wrong (buggy), because suppose that you specify a BitBlt with an image of width == 0, then you would expect the code to **NOT** touch any pixels, yet your rewrite (and that old do-while loop) does modify the pixels.

drivers/base/bootvid/i386/pc98/bootvid.c
HBelusca: (Same comment reply as above)

drivers/base/bootvid/common.c
```diff
@@ -82,24 +82,28 @@ BitBlt(
...
+             ++X, ++Pixel)
+        {
+            if (Pixel % 2)
+            {
+                SetPixel(X, Y, Colors & 0x0F);
+            }
+            else
+            {
+                /* Extract colors at every two pixels */
+                Colors = *InputBuffer++;
```
HBelusca, line 100: Suggestion (so that we get to see first the first case):
```c
            if ((Pixel % 2) == 0)
            {
                /* Extract colors for two pixels at once */
                Colors = *InputBuffer++;
                /* Set the first pixel... */
                SetPixel(X, Y, Colors >> 4);
            }
            else
            {
                /* ... then the second pixel */
                SetPixel(X, Y, Colors & 0x0F);
            }
```

PR #2938, opened 2020-06-21
[FASTFAT_NEW] Sync to upstream 16022c5

SergeGautherie:


Extravert-ir:
I think this commit: https://github.com/microsoft/Windows-driver-samples/commit/16022c593787aef040b8a37ef697316e07f1322e should be put as a sync-commit, that makes more sense

SergeGautherie:
> should be put as a sync-commit

I am not sure what you mean...

Extravert-ir:
I mean the commit hash which is written in `README.FSD`

SergeGautherie:
> commit hash

I kind of concur, but note current hash is not a FastFAT one either.
What is the rule, then? (Global or by component?)

Extravert-ir:
No rules here, anarchy!
So in order to reduce the chaos, I'd like to introduce one: we should sync with the latest commit touching the target code

SergeGautherie:
> latest commit

Very fine with me.
NB: 'cdfs' and 'floppy_new' hashes are already fine.

PR #2939, opened 2020-06-21
[RAPPS] rename variables

kernelbin:
## Purpose

- Use "Scrnshot" instead of "Snapshot"
*Include changing the folder name used to store screenshots.*
- some other minor rename

nothing is changed functionally in this PR

### DIFF
--- a/base/applications/rapps/available.cpp
+++ b/base/applications/rapps/available.cpp
@@ -236,13 +236,13 @@ BOOL CAvailableApplicationInfo::HasUpdate() const
     return (m_szInstalledVersion.Compare(m_szVersion) < 0) ? TRUE : FALSE;
 }
 
-BOOL CAvailableApplicationInfo::RetrieveScrnshot(UINT Index,ATL::CStringW& ScrnshotFileName) const
+BOOL CAvailableApplicationInfo::RetrieveScrnshot(UINT Index,ATL::CStringW& ScrnshotLocation) const
 {
     if (Index >= (UINT)m_szScrnshotLocation.GetSize())
     {
         return FALSE;
     }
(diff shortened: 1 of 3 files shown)


base/applications/rapps/available.cpp
```diff
@@ -54,32 +54,32 @@ VOID CAvailableApplicationInfo::RetrieveGeneralInfo(AvailableStrings& AvlbString
...
+    for (int i = 0; i < MAX_SCRNSHOT_NUM; i++)
     {
-        WCHAR SnapshotField[sizeof("Snapshot") + 4];
-        wsprintfW(SnapshotField, L"Snapshot%d", i + 1);
-        ATL::CStringW SnapshotLocation;
-        if (!GetString(SnapshotField, SnapshotLocation))
+        WCHAR ScrnshotField[sizeof("Scrnshot") + 4];
+        wsprintfW(ScrnshotField, L"Scrnshot%d", i + 1);
+        ATL::CStringW ScrnshotLocation;
+        if (!GetString(ScrnshotField, ScrnshotLocation))
```
learn-more, line 63:
```suggestion
        CStringW ScrnshotField;
        ScrnshotField.Format(L"Screenshot%d", i + 1);
        CStringW ScrnshotLocation;
        if (!GetString(ScrnshotField, ScrnshotLocation))
```
There are multiple changes worth noting:
* We can use the CString's capability of formatting text
* The ATL:: prefix is not needed in this case
* I think it is better to use the non-abbreviated version 'screenshot' because this is something that is going to be typed out by people.

kernelbin: yes, you are right🤣

PR #2941, opened 2020-06-21
[RAPPS] move icon to a field in .txt file

kernelbin:
## Purpose

move icon to a field in .txt file, no longer use the software's name

using the software's name may lead to some problems like the software name is too long, or contain some illegal chars (for example, ```Code::Blocks```. in a normal file name ```:``` is not allowed. ```:``` is used for alternate data streams in NTFS?)

some code is also added to check if ```../``` is used for directory escape

### DIFF
--- a/base/applications/rapps/available.cpp
+++ b/base/applications/rapps/available.cpp
@@ -72,14 +72,16 @@ VOID CAvailableApplicationInfo::RetrieveGeneralInfo(AvailableStrings& AvlbString
         }
         else
         {
-            // TODO: Does the filename contain anything stuff like "\\" ".." ":" "<" ">" ?
+            // TODO: Does the filename contain anything stuff like ":" "<" ">" ?
             // these stuff may lead to security issues
-
             ATL::CStringW ScrnshotName = AvlbStrings.szAppsPath;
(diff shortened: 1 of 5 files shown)


base/applications/rapps/available.cpp
```diff
@@ -83,6 +83,19 @@ VOID CAvailableApplicationInfo::RetrieveGeneralInfo(AvailableStrings& AvlbString
         }
     }
 
+    // TODO: are we going to support specify an URL for an icon ?
+    ATL::CStringW IconLocation;
+    if (GetString(L"Icon", IconLocation))
+    {
+        // TODO: Does the filename contain anything stuff like "\\" ".." ":" "<" ">" ?
```
learn-more, line 90: `wcspbrk`, with `L"\\/:*?\"<>|"` will probably get you started :)

kernelbin: I think "\\" and ".." should be acceptable as long as it's not used for dir escape (as we might want to classify those icons/screenshots into different folders). Now I've already done some work about this.

and about stuff like ```<>?*|:```, I think it over and thought they won't cause some actual damage. the worst thing could be is the file can not be opened. (If I'm wrong and too optimistic, please tell me). So I decided to just leave it there. 🤣

PR #2945, opened 2020-06-23
[TASKMGR] CachedGetUserFromSid(): Properly support NULL 'pSid'

SergeGautherie:
With regard to `SystemUserSid == NULL`...

### DIFF
--- a/base/applications/taskmgr/perfdata.c
+++ b/base/applications/taskmgr/perfdata.c
@@ -125,13 +125,26 @@ static void SidToUserName(PSID Sid, LPWSTR szBuffer, DWORD BufferSize)
 VOID
 WINAPI
 CachedGetUserFromSid(
-    PSID pSid,
-    LPWSTR pUserName,
-    PULONG pcwcUserName)
+    _In_opt_ PSID pSid,
+    _Out_ LPWSTR pUserName,
+    _Inout_ PULONG pcwcUserName)
 {
     PLIST_ENTRY pCur;
     PSIDTOUSERNAME pEntry;
     ULONG cbSid, cwcUserName;
+
+    // It seems NT5.2- does not check the username arguments, especially 'NULL' and '0'.
+
+    if (!pSid)
+    {
+        // TODO: Probably make this string a localizable resource.
+        static const WCHAR szUserNameFromNull[] = L"(unknown)";
+
+        wcsncpy(pUserName, szUserNameFromNull, *pcwcUserName - 1);
+        pUserName[*pcwcUserName - 1] = UNICODE_NULL;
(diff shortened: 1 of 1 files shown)


Extravert-ir:
Can you reproduce the problem while using the app?

SergeGautherie:
> Can you reproduce the problem while using the app?

On LiveCD,

RealLife:
I don't know how to trigger this edge case.
All processes are displayed with `Username == 'SYSTEM'`.

Test:
Skip `AllocateAndInitializeSid()` call.
'System' and 'System Idle Process' processes use new code path and are displayed with empty 'Username'.

base/applications/taskmgr/perfdata.c
```diff
@@ -125,14 +128,22 @@ static void SidToUserName(PSID Sid, LPWSTR szBuffer, DWORD BufferSize)
 VOID
 WINAPI
 CachedGetUserFromSid(
```
ThFabba, line 130: On Windows, `CachedGetUserFromSid` is a function exported from `utildll.dll`. Can you write a test that shows how it behaves with a NULL Sid? Thanks.

base/applications/taskmgr/perfdata.c
```diff
@@ -125,14 +128,22 @@ static void SidToUserName(PSID Sid, LPWSTR szBuffer, DWORD BufferSize)
...
+    _Out_ LPWSTR pUserName,
+    _Inout_ PULONG pcwcUserName)
 {
     PLIST_ENTRY pCur;
     PSIDTOUSERNAME pEntry;
     ULONG cbSid, cwcUserName;
 
+    if (!pSid)
+    {
+        DPRINT("No pSid\n");
```
ThFabba, line 141: The only way for this to become null is if `PerfDataInitialize` fails to allocate memory. So it would make more sense to print there (once) than here (every loop).

base/applications/taskmgr/perfdata.c
SergeGautherie: Thanks for the information.
Done: #2954.

base/applications/taskmgr/perfdata.c
SergeGautherie: `PerfDataInitialize()` is (already) taken care of on #2946.
`GetSecurityInfo()` documentation says it could fail too: I'll move this trace there...

base/applications/taskmgr/perfdata.c
SergeGautherie: > `utildll.dll`

Shall we also start creating this dll? And plan to move `CachedGetUserFromSid()` from here there?

Extravert-ir:

PR #2946, opened 2020-06-23
[TASKMGR] Properly handle start failures

SergeGautherie:
Especially, do not leave unreleased resources.


Extravert-ir:

PR #2947, opened 2020-06-23
Move ROSAPPS to system32

learn-more:
JIRA issue: [CORE-17097](https://jira.reactos.org/browse/CORE-17097)


jimtabor:
Need more information on "Why"? We should keep it separated. Only allow System32 type files in System32.

PR #2949, opened 2020-06-25
[KMTESTS:EX] ExUuid: Fix version check, Add variant check

SergeGautherie:
Version copypasta: actually check whole 4-bit value, not 1 bit only.

Addendum to bbfb3ca.

[WTB 74181](https://testbot.winehq.org/JobDetails.pl?Key=74181)

### DIFF
--- a/modules/rostests/kmtests/ntos_ex/ExUuid.c
+++ b/modules/rostests/kmtests/ntos_ex/ExUuid.c
@@ -3,6 +3,7 @@
  * LICENSE:         LGPLv2+ - See COPYING.LIB in the top level directory
  * PURPOSE:         Kernel-Mode Test Suite UUIDs test
  * PROGRAMMER:      Pierre Schweitzer <pierre@reactos.org>
+ *                  Serge Gautherie <reactos-git_serge_171003@gautherie.fr>
  */
 
 #include <kmt_test.h>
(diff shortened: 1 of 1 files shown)


SergeGautherie:
If you are interested in the FIXME, I would suggest to do a try run then support both results.

ThFabba:
Can you please explain how it was broken before and put that in the commit message and the PR description? Thanks.

SergeGautherie:
> explain

Added.

modules/rostests/kmtests/ntos_ex/ExUuid.c
```diff
@@ -14,8 +15,32 @@ START_TEST(ExUuid)
...
+    else if ((V & 0b1110) == 0b1100)
+    {
+        V = 2;
+    }
+    else // if ((V & 0b1110) == 0b1110)
+    {
+        // Reserved. Fake a value.
+        V = 255;
+    }
+    ok(V == 1, "UUID variant: expected 1, got %u\n", V);
```
ThFabba, line 45: This seems a bit excessive. Failure is not expected, so we don't need it particularly user-friendly. I think as a rule, it's good to handle failure safely (e.g. exit on allocation failure) and actionably (e.g. print out the failing values so that somebody can compare and investigate).
Am I understanding correctly that simply changing the original check to `(Uuid.Data3 & 0xF000) == 0x1000` would address this issue (while still adhering to these goals)? If so, I would prefer that solution to adding this much extra code.

SergeGautherie: I understand what you wrote, but it does not seem to apply in this context.
Did you miss that this second part is about 'variant', not 'version'?

ThFabba: Okay, so you're adding an additional test, too. The same applies there -- make it one line.

SergeGautherie: > one line

Done.

ThFabba:
Superseded by 1481df94aba.

SergeGautherie:
> Superseded by [1481df9](https://github.com/reactos/reactos/commit/1481df94abad5cc75a7f5bdcd57273700b7c7dd2).

PS:
Is 1,000 times necessary? Would 100 or 10 be enough?

PR #2952, opened 2020-06-28
Initial implementation of Wireless Zero Configuration Interface

oleg-dubinskiy:
## Purpose

Implement initial version of Wireless Zero Configuration Interface, which includes: Wireless Zero Configuration Dialogs stub (wzcdlg.dll), Wireless Zero Configuration API DLL (wzcsapi.dll) and Wireless Zero Configuration Service (wzcsvc.dll).
Required by many apps which work with WiFi. Most of them use WZC service instead of WlanSvc. For example, WeFi (which was tested in the ticket), Xirus WiFi Inspector etc.

JIRA issue: [CORE-17148](https://jira.reactos.org/browse/CORE-17148)

## Proposed changes

- First add the wzcsapi.h header with most of declarations that MS header has, except all MS comments and Windows CE specific diffs. Order all declarations separately: definitions, then enumerations and functions.
- Add the directories with the WZC components sources itself and enable them in CMakeLists.
- Almost implement WZC service: all service functions, and stub Svchost support (SvchostPushServiceGlobals, which uses svc.h header with global Svchost declarations added in one of my previous PRs). Also add the registry entries for it: service key in HKLM\SYSTEM\CurrentControlSet\Services\ with all required values, and the svchost string in HKLM\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Svchost\, since it uses svchost.exe for its work, same as in Windows.
- Stub all documented WZC functions in WZC API DLL, export all other undocumented as `stub` only in .spec file. Also implement MIDL_user_allocate(free) in this dll. Add a hack in .spec for WZCEapolUIResponse export with required amount of `int64` parameters, to avoid compilation problems, because the 2nd parameter of it has very big real symbol name, since it uses the struct passed by value.
- Stub the WZC Dialogs DLL: export all functions as `stub` only, since all of them are completely undocumented.
- Also implement DllMain function in each dll.
- Add .rc files with internal dll names, descriptions and version numbers.

## TODO

- [x] Investigate why the service (wzcsvc.dll) almost works in Windows, but does not in ReactOS (same as MS one btw). Although in Win2k3 SP2 it doesn't start automatically (most of all due to unimplemented Svchost support, I see the svchost crash after system startup with dll replaced), but I still can start it manually, and it works without any problems, not worse than MS dll. But in ReactOS it doesn't start: it hand endlessly in "Starting" state (with error 1056), and moreover, it causes fail of some other ROS system services (Schedule, Winmgmt, W32Time, Wuauserv), which fail to start, it their turn, with error 232 (the pipe is being closed). See [CORE-17151](https://jira.reactos.org/browse/CORE-17151) for the reference (is now fixed).


base/services/wzcsvc/wzcsvc.c
```diff
@@ -0,0 +1,162 @@
...
+#include <stdarg.h>
+#include <windef.h>
+#include <winbase.h>
+#include <winreg.h>
+#include <winsvc.h>
+
+#define NDEBUG
+#include <debug.h>
+
+#include <svc.h>
```
learn-more, line 23: How does this build for you, where does it come from?

oleg-dubinskiy: This header contains all global svchost declarations which is my work from my #2931 PR.
This is required e. g. by some services, which implement SvchostPushServiceGlobals function.

PR #2955, opened 2020-06-30
[REACTOS] Fix 'CppCheck: unreachableCode'

SergeGautherie:
Follow-up to a72b785.


Extravert-ir:
Another useless change which does not improve anything except removing a static checker warning.
Unreachable code is put there just in case something goes wrong. UDFS driver deserves a huge refactoring, changing two lines there does not make it work better not improve code readability.

`[HALX86] HalpTrap06(): Remove unreachable 'UNREACHABLE'` This should be put on a quotes page

PR #2957, opened 2020-06-30
[DDK] FltAllocatePoolAlignedWithTag(): Wants 'NonPagedPool', …

SergeGautherie:
…not 'NonPagedPoolNx'

Based on WDK 7.1...
`NonPagedPoolNx` does not exist (yet) on ReactOS headers.


Extravert-ir:
We sync our SDK with the latest version of SDKs from Microsoft so this is correct

SergeGautherie:
> We sync our SDK with the latest version of SDKs from Microsoft so this is correct

Yeah, this is a current case related to the discussion in #2928:
how does it work right, since `NonPagedPoolNx` does not exist (yet) on ReactOS headers?

PR #2958, opened 2020-06-30
[DBGENG] Import dbgeng.dll from Wine Staging 4.18 CORE-17153

oleg-dubinskiy:
## Purpose

Import dbgeng.dll from Wine Staging 4.18. There was no any major improvements in dbgeng since version 4.15, so I decided to use the version to which we are currently synced. And moreover, there are no any Staging patches for dbgeng, so Devel and Staging versions of dbgeng are completely identical.
Required by ICQ New 10.0.40117. Now, after importing that dll, it starts successfully.

JIRA issue: [CORE-17153](https://jira.reactos.org/browse/CORE-17153)

## Proposed changes

- First, import dbgeng.h header from Wine (Staging, 4.18 too) into sdk/include/psdk, since it is required by the dll sources itself.
- Next, import the dll itself: add the separate directory, create CMakeLists.txt, copy Wine's .c and .spec files.
- To fix the compilation, add the following ROS-specific diffs:
    - Remove invalid export in .spec file, since it does not export anything, and made only for matching Windows 10 exports.
    - Add some diffs in the .c file: 
        - Fix the path of winternl.h include, because otherwise it uses our winternl.h instead of Wine's, which shouldn't be used at all, according to its remarks.
        - Force the types of some parameters in some function calls, since looks like we have a bit different declarations of those functions unlike Wine, so they are conflicting with those different parameters types.

## Result

Before:
ICQ_New_ROS_before

After:
ICQ_New_ROS_after

This dll is also present in Win2k3 out of the box:
dbgeng_Win2k3

DLL purpose: Windows Symbolic Debugger Engine.


dll/win32/dbgeng/dbgeng.c
```diff
@@ -0,0 +1,4133 @@
...
+                WARN("Failed to get module information, error %d.\n", GetLastError());
+                continue;
+            }
+
+            target->modules.info[i].params.Base = (ULONG_PTR)info.lpBaseOfDll;
+            target->modules.info[i].params.Size = info.SizeOfImage;
+            target->modules.info[i].params.TimeDateStamp = debug_target_get_module_timestamp(target, modules[i]);
+
+            GetModuleFileNameExA(target->handle, modules[i], target->modules.info[i].image_name,
+#ifdef __REACTOS__
```
ThFabba, line 183: You should define `__WINESRC__` in CMakeLists.txt. That should make `ARRAY_SIZE` available here.

oleg-dubinskiy: Fixed. Thanks @ThFabba. I missed that.
Still compilable for me.

Extravert-ir:
Why are you syncing with Wine 4.18 and not with something more new?

oleg-dubinskiy:
>Why are you syncing with Wine 4.18 and not with something more new?

Because:
1) there was no any improvements ~ after 4.18 (even earlier) **in this component**, see https://source.winehq.org/git/wine.git/history/HEAD:/dlls/dbgeng. There was just one refactoring commit after that release, which is associated with Unix-specific Wine components and doesn't improve the functionality: https://source.winehq.org/git/wine.git/commitdiff/44a230937b6dc320aad8b18828060e3e916eee03. So why we need to sync with the newer versions if there are no any differences?...
2) just because we are currently synced to this Wine version, so, based on my first argument, I decided to keep the same version as of the most of our other winesynced components.

dll/win32/dbgeng/dbgeng.c
```diff
@@ -0,0 +1,4129 @@
...
+{
+    FIXME("%p, %p, %d.\n", iface, context, size);
+    return E_NOTIMPL;
+}
+
+#ifdef __REACTOS__
+static IDebugClient7Vtbl debugclientvtbl =
+#else
+static const IDebugClient7Vtbl debugclientvtbl =
+#endif
```
learn-more, line 1056: Why is this required?

oleg-dubinskiy: @learn-more I made this change (and same others in similar structures) to fix some compilation errors. But looks like now, after defining `__WINESRC__` in CMakeLists.txt, they are gone, even when I removed these diffs. So I'll revert them all, if it will not broke the compilation.
Thanks for noticing! :+1:

oleg-dubinskiy: Fixed.
Still compilable for me.

learn-more (approved):

PR #2959, opened 2020-06-30
[CDFS] WindowsSync to f064dfb

SergeGautherie:
Trivial.


Extravert-ir:
Fill in a pull request when there will be actual code changes

PR #2962, opened 2020-07-02
[CMAKE] Remove obsolete 'GCC_VERSION' checks

SergeGautherie:
following upgrade to GCC 8.4.


SergeGautherie:
Should we add a minimal version check in `configure`, as there is for `cl`?

PR #2963, opened 2020-07-02
[CMAKE] Remove obsolete '_ROSBE_VERSION' checks

SergeGautherie:
following upgrade to RosBE (2.1 and now) 2.2.


SergeGautherie:
Would it be worth adding a minimal version check somewhere, as there is for `cl` and `cmake`?

Saibamen:
> Would it be worth adding a minimal version check somewhere, as there is for `cl` and `cmake`?

IMHO yes

Saibamen (approved):

SergeGautherie:
> > Would it be worth adding a minimal version check somewhere, as there is for cl and cmake?
> 
> IMHO yes

PS: I am not going to do it, anyone feels free to follow-up.

PR #2967, opened 2020-07-05
[SDK] Fix host-tools build from visual studio ide in Release

learn-more:
When building from the IDE in Release, the host-tools would be built in Debug,
and the targets using them expect to find them in the Release folder.

### DIFF
--- a/sdk/cmake/host-tools.cmake
+++ b/sdk/cmake/host-tools.cmake
@@ -38,8 +38,9 @@ function(configure_host_tools HOST_TOOLS_DIR)
     # Make a host-tools target so it'll be built when needed
     # custom target + symbolic output prevents cmake from running
     # the command multiple times per build
+    # Specify the --config option, so the Release/Debug setting from the IDE can be used
     add_custom_command(
-        COMMAND ${CMAKE_COMMAND} --build ${HOST_TOOLS_DIR}
+        COMMAND ${CMAKE_COMMAND} --build ${HOST_TOOLS_DIR} --config $<CONFIG>
         OUTPUT host_tools
         BYPRODUCTS ${_target_locations})
     add_custom_target(build-host-tools ALL DEPENDS host_tools)


sdk/cmake/host-tools.cmake
```diff
@@ -35,11 +35,15 @@ function(configure_host_tools HOST_TOOLS_DIR)
         list(APPEND _target_locations ${_target_location})
     endforeach()
 
+    # When building for MSVC, inhibit the Release/Debug build currently selected in the ide
```
ThFabba, line 38: You mean inherit? :)

sdk/cmake/host-tools.cmake
```diff
@@ -35,11 +35,15 @@ function(configure_host_tools HOST_TOOLS_DIR)
...
+    if (MSVC_IDE)
```
ThFabba, line 39: Any reason we shouldn't just do this unconditionally?

sdk/cmake/host-tools.cmake
learn-more: heh yes.

sdk/cmake/host-tools.cmake
learn-more: `$(ConfigurationName)` is an msvc specific macro

learn-more:
> On a related note... maybe we should just always build the release version of host tools (unless overridden). Potentially faster build speed and whatnot.

Sure, someone is free to do that if they feel it would profit us,
but that requires more changes and I wanted something working now.

sdk/cmake/host-tools.cmake
ThFabba: I think you can just use a generator expression instead, e.g. `$<CONFIG>`

Extravert-ir:
> On a related note... maybe we should just always build the release version of host tools (unless overridden). Potentially faster build speed and whatnot.

I thought about that in the context of freeloader as well

sdk/cmake/host-tools.cmake
learn-more: Thanks, didnt know that!

PR #2969, opened 2020-07-06
[WIN32SS] Fix UserSendKeyboardInput() KEYEVENTF_SCANCODE use case and behaviour of meta key in the OSK

arjav2002:
Small change made to make hotkeys work and the start button work as a modifier on the OSK.

JIRA issue: https://jira.reactos.org/browse/CORE-17144

Changed all the language .rc files to fit the modifier key behaviour of the start button and one line in keyboard.c


learn-more:
Furthermore, you are changing two completely unrelated things in one commit,
where one change (the osk) is not even mentioned in the commit at all...

arjav2002:
So, I should add that too?

arjav2002:
Mentioned the change in the commit. They're really small changes both pertaining to fixing of the meta key (and other hotkeys in the process) so I thought it would be fine to put them both in the same commit.

SergeGautherie:
NB: This PR is a duplicate of #2968. (I don't know what was wrong with the latter...)

arjav2002:
I didn't know how to ammend the commit message properly and messed up.

arjav2002:
Oh alright so two different pull requests? You reckon I close this one?

learn-more:
For future reference:
That is not what I asked for, I requested 2 commits, not 2 PR's
Also, there is no need to close every PR and create a new one, you can simply update a PR with new changes by pushing to the same branch the PR is created from.

arjav2002:
Yes, of course, I have learnt that.
In my defense, you said they didn't explicitly say that I needed to make
two commits in the same PR, your precise words were "these two changes
don't belong in a single commit" and I got a bit confused :P

PR #2970, opened 2020-07-06
[RAPPS] listview refactor

kernelbin:
### some refactor about listview

- add a layer of window (I named it as table view), it contains ListView and AppInfoDisplay. now MainWindow no longer interacts with native listview directly.
- installed apps are now a class (```CInstalledApplicationInfo```) and all the installed apps info are in class ```and CInstalledApps```
- some changes are applied to ```CAvailableApplicationInfo```. For example, now info about whether the app is selected is no longer stored inside it. (now ```CAvailableApplicationInfo``` just store some info purely about the app itself)

### several trivial fixes/improvements

- fix some memory leak when destroying window. (although currently, the program will always exit after the main window is destroyed)
- In the old version, if you select all apps under some subclass in the tree-view under available apps and switch to another subclass and click select all again, it won't work. This bug is fixed in this PR

*nothing important is changed in UI*

## TODO

I comment out some code at the beginning to make it easier to refactor. so now some function is missing
I need to uncomment them or rewrite them.

now the tab-order seems broken. I should fix it


base/applications/rapps/installed.cpp
```diff
@@ -12,149 +12,226 @@
...
 {
-    LPCWSTR szModify = L"ModifyPath";
-    LPCWSTR szUninstall = L"UninstallString";
-    DWORD dwType, dwSize;
-    WCHAR szPath[MAX_PATH];
+    DWORD dwSize = 0;
+    String.Empty();
+
+    // retrieve the size of value first.
+    // TODO: I assume the type as REG_SZ. but I think REG_EXPAND_SZ should be handled correctly too.
```
learn-more, line 76: Please specify / check for REG_SZ :)

base/applications/rapps/installed.cpp
```diff
@@ -12,149 +12,226 @@
...
+        NULL,
+        NULL,
+        (LPBYTE)String.GetBuffer(),
+        &dwSize) != ERROR_SUCCESS)
     {
+        String.ReleaseBuffer();
+        String.Empty();
         return FALSE;
     }
+    String.GetBuffer()[dwSize / sizeof(WCHAR)] = L'0'; // ensure zero terminated
```
learn-more, line 103:
```suggestion
    String.GetBuffer()[dwSize / sizeof(WCHAR)] = L'\0'; // ensure zero terminated
```

base/applications/rapps/installed.cpp
```diff
@@ -12,149 +12,226 @@
...
+                        Info->GetApplicationString(L"RegOwner", Info->szRegOwner);
+                        Info->GetApplicationString(L"ProductID", Info->szProductID);
+                        Info->GetApplicationString(L"HelpLink", Info->szHelpLink);
+                        Info->GetApplicationString(L"HelpTelephone", Info->szHelpTelephone);
+                        Info->GetApplicationString(L"Readme", Info->szReadme);
+                        Info->GetApplicationString(L"Contact", Info->szContact);
+                        Info->GetApplicationString(L"URLUpdateInfo", Info->szURLUpdateInfo);
+                        Info->GetApplicationString(L"URLInfoAbout", Info->szURLInfoAbout);
+                        Info->GetApplicationString(L"Comments", Info->szComments);
+                        Info->GetApplicationString(L"InstallDate", Info->szInstallDate);
```
learn-more, line 176: In the registry it's possible that this is not a string but a DWORD.

For example, `xampp` has a DWORD here:
`1546720283`, which (according to the control panel) translates to `08-dec-19`

base/applications/rapps/installed.cpp
```diff
@@ -12,149 +12,226 @@
...
-        Info.szKeyName.ReleaseBuffer();
-        if (RegOpenKeyW(hKey, Info.szKeyName.GetString(), &Info.hSubKey) == ERROR_SUCCESS)
+        DWORD dwSize = MAX_PATH;
+        HKEY hKey, hSubKey;
+        LONG ItemIndex = 0;
+        ATL::CStringW szKeyName;
+        //Info.hRootKey = IsUserKey ? HKEY_CURRENT_USER : HKEY_LOCAL_MACHINE;
+
+        if (RegOpenKeyW(RootKeyEnum[i],
+            L"Software\\Microsoft\\Windows\\CurrentVersion\\Uninstall",
```
learn-more, line 136: When running as 32 bit app, this is redirected to `Computer\HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall`,
you will have to read 2 extra keys (Which was done in `GetInstalledVersion` before, using the `KEY_WOW64_32KEY` and `KEY_WOW64_64KEY` flags)

base/applications/rapps/gui.cpp
```diff
@@ -1332,296 +1371,717 @@ class CAppsListView :
...
+        }
+    }
+
+    // this function is called when a item of listview is checked/unchecked
+    // CallbackParam is the param passed to listview when adding the item (the one getting focus now).
+    BOOL ItemCheckStateChanged(BOOL bChecked, LPVOID CallbackParam);
+    
+};
+
+class CSideTreeView :
```
learn-more, line 1882: It might make sense to move a few of those to their own cpp file, instead of dropping everything in gui.cpp

base/applications/rapps/gui.cpp
```diff
@@ -1707,15 +2148,23 @@ class CMainWindow :
         return b;
     }
 
+    VOID LayoutCleanup()
+    {
+        if (m_TreeView) delete m_TreeView;
```
learn-more, line 2153: Same here, don't need the `if`

base/applications/rapps/gui.cpp
```diff
@@ -2272,17 +2583,36 @@ class CMainWindow :
...
+                    if (FocusedApps)
+                    {
+                        if (DownloadApplication(FocusedApps, FALSE))
+                        {
+                            UpdateApplicationsList(-1);
+                        }
+                    }
+                    else
+                    {
+                        // TODO: popup a messagebox telling user to select/check some app first
```
learn-more, line 2613: Can we disable the button instead?
That is nicer than the user clicking install and getting slapped around with a messagebox

base/applications/rapps/include/rosui.h
```diff
@@ -495,7 +495,10 @@ class CUiWindow :
 
     virtual ~CUiWindow()
     {
-        T::DestroyWindow();
+        if (::IsWindow(T::m_hWnd))
```
learn-more, line 498:
```suggestion
        if (T::IsWindow())
```

learn-more: On second thought:
Are you sure this is needed?

base/applications/rapps/available.cpp
```diff
@@ -400,71 +400,129 @@ BOOL CAvailableApps::ForceUpdateAppsDB()
...
-    do
+    else
     {
-        // loop for all the cached entries
-        POSITION CurrentListPosition = m_InfoList.GetHeadPosition();
-        CAvailableApplicationInfo* Info = NULL;
+        HANDLE hFind = INVALID_HANDLE_VALUE;
+        WIN32_FIND_DATAW FindFileData;
+
+        hFind = FindFirstFileW(m_Strings.szSearchPath.GetString(), &FindFileData);
```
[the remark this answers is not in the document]
kernelbin: actually this ```CAvailableApps::Enum``` is not written by me, I just add some code to handle ```ENUM_CAT_SELECTED``` separately 🤣. As it's working so I didn't look into the origin code. If you want I can refactor it too

base/applications/rapps/installed.cpp
kernelbin: hmmm, a UNIX-timestamp?

base/applications/rapps/gui.cpp
kernelbin: of course, having their own cpp file is better but this means they need their own .h file too (some classes here are referencing each other). It's not a small work and this PR is already too big IMO🤣
I would like to do this, probably next PR but not this🤣

base/applications/rapps/include/rosui.h
kernelbin: I think yes. because one can destroy the window directly without calling ~CUiWindow. this will lead to ```m_hWnd``` an invalid hwnd. And if delete is performed after the window is destroyed, trying to delete an invalid hwnd will cause an assert

kernelbin:
> Some initial comments while reviewing the code.
> As always, if my comments are not clear enough, feel free to ask me to elaborate :)

ok, thank you very much😃

base/applications/rapps/include/rosui.h
learn-more: yeah but you probably shouldnt destroy the window without the CUiWindow...

kernelbin: sometimes it's just out of your control I think....? 🤣
for example, a parent window will automatically destroy all child window when it's destroyed.
and for top-level window, if you click the close button (unless you handle WM_CLOSE yourself) the system will destroy it too.

base/applications/rapps/gui.cpp
```diff
@@ -65,6 +65,16 @@ enum SCRNSHOT_STATUS
 
 #define PI 3.1415927
 
+// retrieve the value using a mask
+#define GET_MASKED_VALUE(Value, Mask) (((Value) & (Mask)) / ((Mask) - ((Mask) & (Mask - 1))))
```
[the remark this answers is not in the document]
kernelbin: yes, this looks better

base/applications/rapps/gui.cpp
kernelbin: yes, we can. But it requires writing more code. I will update the TODO and left it here🤣 (after all, there was even no MessageBox in the old version).

base/applications/rapps/available.cpp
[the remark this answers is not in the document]
learn-more: Let's keep it for now

base/applications/rapps/installed.cpp
kernelbin: ok, I think I fixed this now~

base/applications/rapps/gui.cpp
```diff
@@ -250,41 +262,37 @@ class CAppRichEdit :
...
-        GET_INFO(L"URLInfoAbout", IDS_INFO_INFOABOUT, CFE_BOLD, CFM_LINK);
-        GET_INFO(L"Comments", IDS_INFO_COMMENTS, CFE_BOLD, 0);
-        GET_INFO(L"InstallDate", IDS_INFO_INSTALLDATE, CFE_BOLD, 0);
-        GET_INFO(L"InstallLocation", IDS_INFO_INSTLOCATION, CFE_BOLD, 0);
-        GET_INFO(L"InstallSource", IDS_INFO_INSTALLSRC, CFE_BOLD, 0);
-        GET_INFO(L"UninstallString", IDS_INFO_UNINSTALLSTR, CFE_BOLD, 0);
-        GET_INFO(L"InstallSource", IDS_INFO_INSTALLSRC, CFE_BOLD, 0);
-        GET_INFO(L"ModifyPath", IDS_INFO_MODIFYPATH, CFE_BOLD, 0);
+        LoadAndInsertText(StringID, Text, StringFlags, TextFlags); \
+    }
```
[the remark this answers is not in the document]
kernelbin: no problem!

base/applications/rapps/gui.cpp
```diff
@@ -1332,296 +1371,720 @@ class CAppsListView :
...
+                LISTVIEW_ICON_SIZE,
+                LISTVIEW_ICON_SIZE,
+                LR_LOADFROMFILE);
+        }
 
-        m_TreeView->AddCategory(TVI_ROOT, IDS_SELECTEDFORINST, IDI_SELECTEDFORINST);
+        if (!hIcon || GetLastError() != ERROR_SUCCESS)
+        {
+            /* Load default icon */
+            hIcon = (HICON)LoadIconW(hInst, MAKEINTRESOURCEW(IDI_MAIN));
```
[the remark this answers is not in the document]
kernelbin: Of course yes. But soon I'm going to rewrite this listview (by WM_DRAWITEM or whatever), and this ImageList probably will be no longer used at that time🤣
Can I just leave it here? I will keep this in mind. If this ImageList will still be used (I cannot foresee it), I will make it up later😅

base/applications/rapps/available.cpp
```diff
@@ -400,71 +400,129 @@ BOOL CAvailableApps::ForceUpdateAppsDB()
...
+        POSITION Position = m_SelectedList.Find(AvlbInfo);
+        if (Position)
+        {
+            m_SelectedList.RemoveAt(Position);
+            return TRUE;
         }
-    } while (FindNextFileW(hFind, &FindFileData) != 0);
+        return FALSE;
+    }
+}
```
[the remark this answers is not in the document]
kernelbin: ok. I separate them into 2 functions now

PR #2971, opened 2020-07-07
[REACTOS] *.spec: Add some '-ret64'

SergeGautherie:
And fix calling convention in apisets.


dll/apisets/api-ms-win-core-crt-l1-1-0.spec
```diff
@@ -2,7 +2,7 @@
 # This file is autogenerated by update.py
 
 @ stdcall __isascii() msvcrt.__isascii
-@ stdcall _atoi64() msvcrt._atoi64
+@ cdecl -ret64 _atoi64() msvcrt._atoi64
```
ThFabba, line 5: This does not seem useful for forwarders. Do you expect this to achieve anything?

SergeGautherie: At least, code consistency with ReactOS/Wine and `cdecl`/`stub` cases.
If you confirm `-ret64` is useless and unwanted for forwarders, I'll remove them.

learn-more: A forwarder does not contain any code, it is simply a string in the dll specifying the target dll + function name

SergeGautherie: Removed.

dll/apisets/api-ms-win-crt-convert-l1-1-0.spec
```diff
@@ -7,13 +7,13 @@
 @ stub _atof_l
 @ stub _atoflt
 @ stub _atoflt_l
-@ stdcall _atoi64() msvcrt._atoi64
-@ stub _atoi64_l
+@ cdecl _atoi64() msvcrt._atoi64
+@ stub -ret64 _atoi64_l
```
ThFabba, line 11: Does this flag have any effect on stubs?

dll/apisets/api-ms-win-core-crt-l1-1-0.spec
```diff
@@ -2,7 +2,7 @@
 # This file is autogenerated by update.py
 
 @ stdcall __isascii() msvcrt.__isascii
-@ stdcall _atoi64() msvcrt._atoi64
+@ cdecl _atoi64() msvcrt._atoi64
```
ThFabba, line 5: This change is not useful on a forwarder either.

dll/apisets/api-ms-win-crt-convert-l1-1-0.spec
tkreuzer: No, it's only useful for wine relay tracing, which we don't support.

dll/apisets/api-ms-win-core-crt-l1-1-0.spec
tkreuzer: More useful than the -ret64 thing.

tkreuzer (commented):
IMO completely pointless.

dll/apisets/api-ms-win-crt-convert-l1-1-0.spec
learn-more: actually, we do ;)
https://reactos.org/wiki/Debugging#Log_api_calls_to_a_specific_ReactOS_dll

PR #2972, opened 2020-07-07
[DRIVERS:TCPIP] DispTdiQueryInformation NULL checks

Tom-III:
## Purpose

Implementing a patch submitted by hater which adds additional NULL checks to DispTdiQueryInformation, which return STATUS_INVALID_PARAMETER.

JIRA issue: [CORE-12274](https://jira.reactos.org/browse/CORE-12274)

## Proposed changes

-  Two additional NULL checks in DispTdiQueryInformation which cause STATUS_INVALID_PARAMETER to be returned instead of a NULL pointer dereference taking place in Kernel Mode.

### DIFF
--- a/drivers/network/tcpip/tcpip/dispatch.c
+++ b/drivers/network/tcpip/tcpip/dispatch.c
@@ -711,6 +711,12 @@ NTSTATUS DispTdiQueryInformation(
         switch ((ULONG_PTR)IrpSp->FileObject->FsContext2) {
           case TDI_TRANSPORT_ADDRESS_FILE:
             AddrFile = (PADDRESS_FILE)TranContext->Handle.AddressHandle;
+            if (AddrFile == NULL)
+            {
+                TI_DbgPrint(MIN_TRACE, ("FIXME: No address file object.\n"));
+                ASSERT(AddrFile != NULL);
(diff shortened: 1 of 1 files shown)


Tom-III:
Thanks, fixed commit message.

binarymaster:
Nobody else reviewed this PR so far. :slightly_frowning_face: 

I added "needs tests" label, which means you'll need to create API tests that shows this is correct behavior on Windows XP / 2003. Note that the tests might already exist, and in this case you just need to demonstrate output of the tests for "before" and "after" this patch.

learn-more:
In theory a null check is good, but not if this thing is never supposed to be NULL...
I simply don't know enough about this code to judge that :)

BenNottelling:
> In theory a null check is good, but not if this thing is never supposed to be NULL...
> I simply don't know enough about this code to judge that :)

Perhaps merging it with a ```//FIXME: NULL checks added for stability, but is this supposed to happen?```
I think avoiding a BSoD is a good idea

BenNottelling (approved):

ThFabba (approved):

ThFabba:
I solved this dilemma by adding asserts in these cases. Now it won't go unnoticed, but you can also continue execution instead of crashing.

PR #2973, opened 2020-07-07
[OSK] Meta key modifier behaviour added

arjav2002:
To make the meta key in the OSK have modifier key behaviour.
JIRA issue: [CORE-171611](https://jira.reactos.org/browse/CORE-17161)

Changed all the .rc files in the OSK so that the meta key doesn't release on it's own and needs a second click to release.


arjav2002:
That means?

SergeGautherie:
NB: Split off from #2969.

arjav2002:
> NB: Split off from #2969.

Branch updated.

PR #2974, opened 2020-07-07
[NTUSER] Fix UserSendKeyboardInput() KEYEVENTF_SCANCODE use case

arjav2002:
When KEYEVENTF_SCANCODE is supplied, the edited routine is supposed to ignore the supplied VK and hence procure it itself from the given scancode. The word VK is getting values > 0xFE from the keyboard table which is incorrect (VK > 0xFE do not exist)

JIRA issue: [CORE-17144](https://jira.reactos.org/browse/CORE-17144)

Turns out, simply ```&= 0xFF``` makes the VK correspond to the correct value.

### DIFF
--- a/modules/rostests/apitests/user32/keybd_event.c
+++ b/modules/rostests/apitests/user32/keybd_event.c
@@ -1,0 +1,42 @@
+/*
+ * PROJECT:     ReactOS api tests
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Test for keyboard inputs with flags KEYEVENTF_EXTENDEDKEY and KEYEVENTF_SCANCODE
+ * COPYRIGHT:   Copyright 2021 Arjav Garg <arjavgarg@gmail.com>
+ */
+
+#include "precomp.h"
+
+static void testScancodeExtendedKey(BYTE wVk, BYTE scanCode)
+{
+    trace("wVK: %x\tScancode: %x\n", wVk, scanCode);
+
+    keybd_event(0, scanCode, KEYEVENTF_SCANCODE | KEYEVENTF_EXTENDEDKEY, 0);
+    SHORT winKeyState = GetAsyncKeyState(wVk);
+    ok(winKeyState & 0x8000, "VK=%x should be detected as key down.\n", wVk);
+
+    keybd_event(0, scanCode, KEYEVENTF_SCANCODE | KEYEVENTF_EXTENDEDKEY | KEYEVENTF_KEYUP, 0);
+    winKeyState = GetAsyncKeyState(wVk);
+    ok(!(winKeyState & 0x8000), "VK=%x should be detected as key up.\n", wVk);
+}
+
+/* https://docs.microsoft.com/en-us/windows/win32/inputdev/about-keyboard-input#extended-key-flag */
+START_TEST(keybd_event)
+{
+    testScancodeExtendedKey(VK_RWIN, 0x5C);
+    testScancodeExtendedKey(VK_LWIN, 0x5B);
+    testScancodeExtendedKey(VK_RMENU, 0x38);
+    testScancodeExtendedKey(VK_RCONTROL, 0x1D);
+    testScancodeExtendedKey(VK_INSERT, 0x52);
+    testScancodeExtendedKey(VK_DELETE, 0x53);
+    testScancodeExtendedKey(VK_HOME, 0x47);
+    testScancodeExtendedKey(VK_END, 0x4f);
+    testScancodeExtendedKey(VK_PRIOR, 0x49);
+    testScancodeExtendedKey(VK_NEXT, 0x51);
+    testScancodeExtendedKey(VK_UP, 0x48);
+    testScancodeExtendedKey(VK_RIGHT, 0x4d);
+    testScancodeExtendedKey(VK_LEFT, 0x4b);
+    testScancodeExtendedKey(VK_DOWN, 0x50);
+    testScancodeExtendedKey(VK_DIVIDE, 0x35);
+    testScancodeExtendedKey(VK_RETURN, 0x1C);
+}
--- a/win32ss/user/ntuser/keyboard.c
+++ b/win32ss/user/ntuser/keyboard.c
@@ -1020,8 +1020,11 @@ UserSendKeyboardInput(KEYBDINPUT *pKbdInput, BOOL bInjected)
         }
         else
         {
-            wVk = pKbdInput->wVk & 0xFF;
+            wVk = pKbdInput->wVk;
         }
+
+        /* Remove all virtual key flags (KBDEXT, KBDMULTIVK, KBDSPECIAL, KBDNUMPAD) */
+        wVk &= 0xFF;
     }
 
     /* If time is given, use it */
--- a/modules/rostests/apitests/user32/CMakeLists.txt
+++ b/modules/rostests/apitests/user32/CMakeLists.txt
@@ -23,6 +23,7 @@ list(APPEND SOURCE
     GetUserObjectInformation.c
     GetWindowPlacement.c
     InitializeLpkHooks.c
+    keybd_event.c
     LoadImage.c
     LookupIconIdFromDirectoryEx.c
     MessageStateAnalyzer.c
--- a/modules/rostests/apitests/user32/testlist.c
+++ b/modules/rostests/apitests/user32/testlist.c
@@ -25,6 +25,7 @@ extern void func_GetSystemMetrics(void);
 extern void func_GetUserObjectInformation(void);
 extern void func_GetWindowPlacement(void);
 extern void func_InitializeLpkHooks(void);
+extern void func_keybd_event(void);
 extern void func_LoadImage(void);
 extern void func_LookupIconIdFromDirectoryEx(void);
 extern void func_MessageStateAnalyzer(void);
@@ -76,6 +77,7 @@ const struct test winetest_testlist[] =
     { "GetUserObjectInformation", func_GetUserObjectInformation },
     { "GetWindowPlacement", func_GetWindowPlacement },
     { "InitializeLpkHooks", func_InitializeLpkHooks },
+    { "keybd_event", func_keybd_event },
     { "LoadImage", func_LoadImage },
     { "LookupIconIdFromDirectoryEx", func_LookupIconIdFromDirectoryEx },
     { "MessageStateAnalyzer", func_MessageStateAnalyzer },


SergeGautherie:
NB: Split off from #2969.

arjav2002:
> NB: Split off from #2969.

Branch updated.

GeoB99:
@arjav2002 What's the status of this, have you got any ready tests to provide for this PR?

arjav2002:
> @arjav2002 What's the status of this, have you got any ready tests to provide for this PR?

Unfourtunately, as of now, none. I hope to get to it at the end of the semester.

arjav2002:
Update on research regarding this PR.
Calls/notable steps on pressing keyboard ROS key: `UserProcessKeyboardInput` -> call to `IntVscToVk` -> return to `UserProcessKeyboardInput` and do `&= 0xFF` on the `wVk` -> call to `UserSendKeyboardInput` ->  `.....`
For the osk (the build of which works on windows 10): call to `NtUserSendInput` with `KEYEVENTF_SCANCODE` (which works on windows, so can't/don't want to change this) -> `UserSendKeyboardInput` -> call to `IntVscToVk` -> return to `UserSendKeyboardInput` -> `.....`

If you dig deeper into the call stacks, you'll see that both reach `UpdateAsyncKeyState`. Here the `wVk` is being divided by 8 and being used to index into an array `gafAsyncKeyStateRecentDown` of size `0xFF / 0x8 = 0x20`. Therefore, if we assume that the osk is cprrect in calling with the flag `KEYEVENTF_SCANCODE`, then it resulting in a call to `UpdateAsyncKeyState` with the `wVk = 0x15B`. On dividing by 8, this results in 0x2B which is an out of bounds index for the array.

arjav2002:
I  could  change the commit to `&= 0xFF` only in this else block:
```
if (pKbdInput->dwFlags & KEYEVENTF_UNICODE)
    {
        ....
    }
    else
    {
        ....
    }
```

win32ss/user/ntuser/keyboard.c
```diff
@@ -1020,10 +1020,12 @@ UserSendKeyboardInput(KEYBDINPUT *pKbdInput, BOOL bInjected)
         }
         else
         {
-            wVk = pKbdInput->wVk & 0xFF;
+            wVk = pKbdInput->wVk;
         }
     }
 
+    wVk &= 0xFF;
```
HBelusca, line 1027: This would AND `VK_PACKET` with 0xFF, in case the `if (pKbdInput->dwFlags & KEYEVENTF_UNICODE)` is taken. Is it what you want?

arjav2002: That is an unintended effect, and I was in fact about to propose that I move this line to something like this:
```
if (pKbdInput->dwFlags & KEYEVENTF_UNICODE)
    {
        /* Generate WM_KEYDOWN msg with wParam == VK_PACKET and
           high order word of lParam == pKbdInput->wScan */
        wVk = VK_PACKET;
    }
    else
    {
        wScanCode &= 0x7F;
        if (pKbdInput->dwFlags & KEYEVENTF_SCANCODE)
        {
            /* Don't ignore invalid scan codes */
            wVk = IntVscToVk(wScanCode | (bExt ? 0xE000 : 0), pKbdTbl);
            if (!wVk) /* use 0xFF if vsc is invalid */
                wVk = 0xFF;
        }
        else
        {
            wVk = pKbdInput->wVk;
        }
        wVk &= 0xFF;
    }
```

HBelusca: this would be better indeed.

HBelusca:
So, what's missing in this PR, is to write a rostest for proving that.

arjav2002:
Reason for using hardcoded scancode values: [CORE-17770](https://jira.reactos.org/browse/CORE-17770)

modules/rostests/apitests/user32/keybd_event.c
```diff
@@ -0,0 +1,33 @@
...
+	testScancodeExtendedKey(VK_END, 0x4f);
+	testScancodeExtendedKey(VK_PRIOR, 0x49);
+	testScancodeExtendedKey(VK_NEXT, 0x51);
+	testScancodeExtendedKey(VK_UP, 0x48);
+	testScancodeExtendedKey(VK_RIGHT, 0x4d);
+	testScancodeExtendedKey(VK_LEFT, 0x4b);
+	testScancodeExtendedKey(VK_DOWN, 0x50);
+	testScancodeExtendedKey(VK_DIVIDE, 0x35);
+	testScancodeExtendedKey(VK_RETURN, 0x1C);
+}
```
katahiromz, line 33: Insert a newline at end of file.

modules/rostests/apitests/user32/keybd_event.c
```diff
@@ -0,0 +1,33 @@
...
+    SHORT winKeyState = GetAsyncKeyState(wVk);
+    ok(winKeyState & 0x8000, "VK=%x should be detected as key down.\n", wVk);
+
+    keybd_event(0, scanCode, KEYEVENTF_SCANCODE | KEYEVENTF_EXTENDEDKEY | KEYEVENTF_KEYUP, 0);
+    winKeyState = GetAsyncKeyState(wVk);
+    ok(!(winKeyState & 0x8000), "VK=%x should be detected as key up.\n", wVk);
+}
+
+/* https://docs.microsoft.com/en-us/windows/win32/inputdev/about-keyboard-input#extended-key-flag */
+START_TEST(keybd_event) {
```
katahiromz, line 16: Insert a newline after paren.

modules/rostests/apitests/user32/keybd_event.c
```diff
@@ -0,0 +1,33 @@
+#include "precomp.h"
+
+void testScancodeExtendedKey(BYTE wVk, BYTE scanCode) {
```
katahiromz, line 3: Ditto.

katahiromz:
The screenshots of BEFORE and AFTER please!

win32ss/user/ntuser/keyboard.c
```diff
@@ -1020,8 +1020,9 @@ UserSendKeyboardInput(KEYBDINPUT *pKbdInput, BOOL bInjected)
         }
         else
         {
-            wVk = pKbdInput->wVk & 0xFF;
+            wVk = pKbdInput->wVk;
         }
+        wVk &= 0xFF;
```
learn-more, line 1025: None of your testcases showcases this behavior, since you use a byte for you VK?

arjav2002:
before fix:

https://user-images.githubusercontent.com/22453292/133794532-abc62e3c-52ef-4dfa-ae26-a05301f22619.mp4

after fix:

https://user-images.githubusercontent.com/22453292/133794649-e3771769-9ae2-4a7b-9509-c9db5edd5f29.mp4

win32ss/user/ntuser/keyboard.c
arjav2002: The difference this change is making is reflected when the `if (pKbdInput->dwFlags & KEYEVENTF_SCANCODE)` is executed. Since the else block also had the same operating being done, I basically clubbed both in the ANDings into a single line. Should I change it to AND only in the if block?

arjav2002: I also want to mention that in the case of extended keys with `KEYEVENTF_SCANCODE` flag, the `IntVscToVk` does in fact return a word `E0xx`.

modules/rostests/apitests/user32/keybd_event.c
```diff
@@ -0,0 +1,35 @@
+#include "precomp.h"
```
ThFabba, line 1: Can you add a copyright header please? You can copy/adapt it from any other test. https://reactos.org/wiki/Coding_Style#File_Structure also has an example (and that's the latest version of the header, which is preferred).

modules/rostests/apitests/user32/keybd_event.c
```diff
@@ -0,0 +1,35 @@
...
+
+void testScancodeExtendedKey(BYTE wVk, BYTE scanCode)
+{
+	trace("wVK: %x\tScancode: %x\n", wVk, scanCode);
+
+	keybd_event(0, scanCode, KEYEVENTF_SCANCODE | KEYEVENTF_EXTENDEDKEY, 0);
+    SHORT winKeyState = GetAsyncKeyState(wVk);
```
ThFabba, line 8: Your indentation suddenly switches from 2 to 4 spaces here. Please keep it consistent (ideally at 4 everywhere).

win32ss/user/ntuser/keyboard.c
ThFabba: I would in fact suggest keeping the `& 0xFF` in both branches in this case -- mainly so that you can put a descriptive comment above it in the `KEYEVENTF_SCANCODE` case that explains why it needs to be done there (in particular -- why does the lookup table contain values with the high bits set? Is that needed for a different code path, but not here?). You can also mention that E0xx example there.
If you have a good understanding of what's going on here, it could in fact be useful to add a comment to the non-scancode case as well, explaining why the masking is needed there.

modules/rostests/apitests/user32/keybd_event.c
```diff
@@ -0,0 +1,35 @@
...
+    ok(winKeyState & 0x8000, "VK=%x should be detected as key down.\n", wVk);
+
+    keybd_event(0, scanCode, KEYEVENTF_SCANCODE | KEYEVENTF_EXTENDEDKEY | KEYEVENTF_KEYUP, 0);
+    winKeyState = GetAsyncKeyState(wVk);
```
ThFabba, line 12: So, looking at `ProcessKeyEvent`, the async key state does seem to be a good way to tell what's happening in `UserSendKeyboardInput`. A low-level keyboard hook would be a slightly better test, because it would show a 1:1 reflection of what's happening in input processing, and the keycode passed to it doesn't go through `IntSimplifyVk`.
Similarly, `SendInput` would make a slightly better test for this than `keybd_event`, because it goes through less abstraction. With the test as is, I could ask "How do you know that it's not some magic inside `keybd_event` and/or `GetAsyncKeyState` that's making this test behave this way.
However, I think it's close enough to reasonable proof that your change is in the correct place -- now let's just make sure it doesn't cause any regressions.

ThFabba (commented):
One more note: can you please put the Jira ticket in the commit message? That'll allow Jira to automatically link to this PR as well as the commit, once it's in master.

win32ss/user/ntuser/keyboard.c
arjav2002: Correction: The word returned in the case mentioned is `0x01XX`, not `0xE000`.

arjav2002: Yes, these high bits in the lookup table are required. `UserProcessKeyboardInput` uses this to set the `dwFlags`. 
image

One can see that, an ANDing action similar to my changes is performed here also, before sending the `wVk` to `UserSendKeyboardInput`.

arjav2002:
Winetests: https://testbot.winehq.org/JobDetails.pl?Key=98280

win32ss/user/ntuser/keyboard.c
arjav2002: I think the `&= 0xFF` operation is done in both the cases for the same reason, to remove all the virtual key flags.

ThFabba: If the explanation is the same, then keeping it outside the if makes sense to me.

arjav2002: I have added the comment alongwith keeping it outside the if, apart from the other notes. Do let me know if something else is still left.

ThFabba:
Testbot run: https://reactos.org/testman/compare.php?ids=78767,79047
Recent master for flaky tests comparison: https://reactos.org/testman/compare.php?ids=79037,79040,79043
LGTM

ThFabba (approved):

ThFabba:
I just ran this test on Windows and noticed the following:
* on Windows 10 20H2 it left the start menu open after the test was finished
* on Windows Server 2003 SP2, it seemed to effectively send "up arrow -> backspace -> slash -> return" to my command prompt, therefore running itself again, except with `keybd_even/` as the test name

I don't think these are blockers, and will merge regardless. But perhaps you could look into ways of improving the test such that it doesn't affect system state as much, especially not in a variable way (e.g. by explicitly setting focus to a particular window)

PR #2490, opened 2020-03-31
WIP. [MSGINA] gui: Improve DLG_DATA handling. CORE-16775

SergeGautherie:
## Purpose

Trying to make it a bit clearer/safer.

JIRA issue: [CORE-16775](https://jira.reactos.org/browse/CORE-16775)

## TODO

- [ ] @katahiromz, you added `LogoWidth` and `LogoHeight` on 623dd26ccea5e79001aee769c230fb4f8e304fc8.
Can they be removed? Or what are they intended for?
- [ ] @julcar, could you test this PR? Thanks.

PR #2976, opened 2020-07-11
[MSGINA] gui: Disable unused 'LogoWidth' and 'LogoHeight' in '_DLG_DATA'

SergeGautherie:
Addendum to 623dd26.

Part of #2490.
See discussion there:
* `LogoHeight` would be used for CJK support, but there seems to be no such code.
* `LogoWidth` is still assumed to be useless.

### DIFF
--- a/dll/win32/msgina/gui.c
+++ b/dll/win32/msgina/gui.c
@@ -12,6 +12,9 @@
 #include <winnls.h>
 #include <winreg.h>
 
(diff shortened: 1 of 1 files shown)


ThFabba:
If you don't know the future direction of the code, don't mess with it. Just leave it up to the people who put that code there.
Removing unused code can be good to simplify things -- an ifdef does the opposite.

PR #2980, opened 2020-07-12
[USER32] Implement checking for animated mouse cursors in CURSORICON_LoadFileFromW CORE-14166

oleg-dubinskiy:
## Purpose

Implement checking for available animated mouse cursors (.ani), which are set via main.cpl, in internal CURSORICON_LoadFileFromW function. My implementation simply assigns `CURSORICON_GetCursorDataFromANI` call with all required parameters, to `hCurIcon` variable, similarly to as it done in Wine: https://source.winehq.org/git/wine.git/blob/95ac10e526f9949802c8b76a969a942a9a4c2727:/dlls/user32/cursoricon.c#l1108.

JIRA issue: [CORE-14166](https://jira.reactos.org/browse/CORE-14166)

## Proposed changes

- Implement the checking for .ani cursors in according code path of CURSORICON_LoadFileFromW function.
- Update copyrights.

## TODO

- [ ] Investigate why it still doesn't work (due to another issue in win32k or something is still wrong in my implementation instead). Now, after my changes, it spams about fail in UserGetCurIconObject function from win32k for some reason. See the ticket comment for details: https://jira.reactos.org/browse/CORE-14166?focusedCommentId=124356&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-124356.

### DIFF
--- a/win32ss/user/user32/windows/cursoricon.c
+++ b/win32ss/user/user32/windows/cursoricon.c
@@ -3,7 +3,8 @@
  * COPYRIGHT:       GPL - See COPYING in the top level directory
  * FILE:            win32ss/user/user32/windows/cursoricon.c
  * PURPOSE:         cursor and icons implementation
- * PROGRAMMER:      Jérôme Gardou (jerome.gardou@reactos.org)
+ * PROGRAMMERS:     Jérôme Gardou (jerome.gardou@reactos.org)
+ *                  Oleg Dubinskiy (oleg.dubinskij2013@yandex.ua)
  */
 
 #include <user32.h>
@@ -1354,7 +1355,7 @@ CURSORICON_LoadFromFileW(
     /* Check for .ani. */
     if (memcmp( bits, "RIFF", 4 ) == 0)
     {
-        UNIMPLEMENTED;
+        hCurIcon = (HANDLE)CURSORICON_GetCursorDataFromANI(&cursorData, bits, filesize, fuLoad);
         goto end;
     }
 


oleg-dubinskiy:
I found the source of problem! :smiley: UserGetCurIconObject does no longer fail for me after I changed the type of `CURSORICON_GetCursorDataFromANI` function to `HANDLE`. Should I rewrite/winesync that function then? :thinking: In Wine, it's already of `HCURSOR` type, which is, in essence, a `HANDLE` to a cursor: https://source.winehq.org/git/wine.git/blob/HEAD:/dlls/user32/cursoricon.c#l1473

jimtabor:
Doesn't work? Maybe it's not implemented yet??

oleg-dubinskiy:
>Maybe it's not implemented yet??

Not completely (properly) implemented, but it doesn't mean that not implemented at all (if you mean my part of the code).
But the `CURSORICON_GetCursorDataFromANI`, which has been called, is already implemented, but looks like not (completely) correctly too (because of `BOOL` type of the function): https://git.reactos.org/?p=reactos.git;a=blob;f=win32ss/user/user32/windows/cursoricon.c;hb=64abd9fca8b6ebaba8769ca249b7fba2dde716a2#l890.

win32ss/user/user32/windows/cursoricon.c
```diff
@@ -1354,7 +1355,7 @@ CURSORICON_LoadFromFileW(
     /* Check for .ani. */
     if (memcmp( bits, "RIFF", 4 ) == 0)
     {
-        UNIMPLEMENTED;
+        hCurIcon = (HANDLE)CURSORICON_GetCursorDataFromANI(&cursorData, bits, filesize, fuLoad);
```
katahiromz, line 1358: Why `(HANDLE)` cast needed?

oleg-dubinskiy: I think it's indeed not needed to cast it, since it doesn't work properly. I guess I should just check the `CURSORICON_GetCursorDataFromANI` function with all required parameters and then do something with `hCurIcon` variable, something like
`if(!CURSORICON_GetCursorDataFromANI(&cursorData, bits, filesize, fuLoad)`
`{`
`    /* do something if the function returns FALSE */`
`}`
`else`
`{`
`    /* do something with hCurIcon if it returns TRUE */`
`}`
But I wonder what it should be exactly. This turned out to be more difficult than I thought at first. Sorry. :sweat_smile: 
I did the cast earlier only because it has been done in Wine by this way. I think that the converting `CURSORICON_GetCursorDataFromANI` into `HANDLE` type is also wrong.

learn-more:
@oleg-dubinskiy what is the status of this PR?

PR #2982, opened 2020-07-15
[RTL] vDbgPrintExWithPrefixInternal(): Make 'strncpy()' length more u…

SergeGautherie:
…sual

Fix GCC 'RelWithDebInfo'
'.../debug.c:84:9: error: 'strncpy' specified bound depends on the length of the source argument [-Werror=stringop-overflow=]'


Extravert-ir:
This code is wrong. `stringop-overflow` check has too many false positives

SergeGautherie:
> This code is wrong.

How is it wrong?

> `stringop-overflow` check has too many false positives

How to fix this/them, then?

PR #2984, opened 2020-07-16
[REACTOS] C99 'inline' needs 'static'

SergeGautherie:
Fix GCC 'MinSizeRel' build.

### DIFF
--- a/drivers/filesystems/cdfs/deviosup.c
+++ b/drivers/filesystems/cdfs/deviosup.c
@@ -253,6 +253,9 @@ Arguments:
 }
 
 
+#ifdef __REACTOS__
+static
+#endif
 __inline
 TRACK_MODE_TYPE
 CdFileTrackMode (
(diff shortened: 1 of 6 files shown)


Extravert-ir:
If you want to fix errors in some special build, please make all changes required for that in a single pull request

SergeGautherie:
> all changes

Added workaround for `RelWithDebInfo` and `MinSizeRel` builds.
With this PR, all 4 GCC builds succeed.

Extravert-ir:
This is too easy :) Would be nice to have this bug fixed as well https://jira.reactos.org/browse/CORE-16793

SergeGautherie:
> This is too easy :)

That is the workaround already used for 'Release'.

--

> Would be nice to have this bug fixed as well

Not this PR, though I did already start/try to work on these warnings.

> CORE-16793

Please, answer my questions on #2982...

Extravert-ir (changes requested):
I will accept the change with `static inline` thing, but only if you fix all occurrences of that in the codebase

SergeGautherie:
> fix all occurrences of that in the codebase

Which other occurrences?

Extravert-ir:
Better just override the `__inline` definition with `static inline`

PR #2993, opened 2020-07-18
[WHOAMI] WhoamiGetTokenInfo(): Make it cleaner and safer

SergeGautherie:
JIRA issue: [CORE-11079](https://jira.reactos.org/browse/CORE-11079)

### DIFF
--- a/base/applications/cmdutils/whoami/whoami.c
+++ b/base/applications/cmdutils/whoami/whoami.c
@@ -80,35 +80,43 @@ BOOL WhoamiFree(VOID* Buffer)
 
 VOID* WhoamiGetTokenInfo(TOKEN_INFORMATION_CLASS TokenType)
 {
-    HANDLE hToken = 0;
-    DWORD dwLength = 0;
-    VOID* pTokenInfo = 0;
+    HANDLE hToken;
+    DWORD dwLength;
+    VOID* pTokenInfo = NULL;
+    BOOL bRet;
 
     if (OpenProcessToken(GetCurrentProcess(), TOKEN_READ, &hToken))
     {
-        GetTokenInformation(hToken,
-                            TokenType,
-                            NULL,
-                            dwLength,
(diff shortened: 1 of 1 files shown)


SergeGautherie:
> Please mention @vicmarcal in the commit message. 

Added, though pretty obvious.

> And the message itself have to be more "clear" about what's done here and why.

Replaced with more details.

base/applications/cmdutils/whoami/whoami.c
```diff
@@ -80,35 +80,43 @@ BOOL WhoamiFree(VOID* Buffer)
...
-                            TokenType,
-                            NULL,
-                            dwLength,
-                            &dwLength);
+        bRet = GetTokenInformation(hToken,
+                                   TokenType,
+                                   NULL,
+                                   0,
+                                   &dwLength);
+        if (bRet || GetLastError() != ERROR_INSUFFICIENT_BUFFER)
```
learn-more, line 95: Did you ever test this code?

SergeGautherie:
> Did you ever test this code?

Tested on LiveCD: results look unchanged.
What do you think is wrong?

learn-more:
If it succeeds you abort?

SergeGautherie:
> If it succeeds you abort?

Yes: as in 98bbe83.

base/applications/cmdutils/whoami/whoami.c
SergeGautherie: Tested on LiveCD: results look unchanged.
What do you think is wrong?

SergeGautherie:
> @Extravert-ir requested changes

Please, update status.

GeoB99:
This is pure bikesheeding -- `NtQueryInformationToken` will always return `STATUS_BUFFER_TOO_SMALL` if you provide a buffer length of 0 (and a NULL pointer). I see no benefits coming form your changes.

PR #2997, opened 2020-07-19
[REACTOS] Fix GetTokenInformation() usage

SergeGautherie:
…return value checks

### DIFF
--- a/dll/win32/netid/netid.c
+++ b/dll/win32/netid/netid.c
@@ -179,14 +179,11 @@ IsUserAdmin(VOID)
                           &hToken))
         goto done;
 
-    dwSize = 0;
-    GetTokenInformation(hToken,
-                        TokenGroups,
-                        NULL,
-                        0,
-                        &dwSize);
-    if (dwSize == 0)
+    if (GetTokenInformation(hToken, TokenGroups, NULL, 0, &dwSize) ||
+        GetLastError() != ERROR_INSUFFICIENT_BUFFER)
+    {
         goto done;
+    }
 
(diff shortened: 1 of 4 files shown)


Extravert-ir:
You're inverting the logic here. Can you elaborate why previous code is wrong and we should use your version.

And please, do similar changes in one pull request!

SergeGautherie:
> why previous code is wrong

Obviously, current code errors out as soon as `GetTokenInformation()` fails, so `GetLastError()` is not checked. In other words, the rest of the function is dead code.

> And please, do similar changes in one pull request!

Merged 3 previous PRs.

dll/win32/netid/netid.c
```diff
@@ -180,13 +180,15 @@ IsUserAdmin(VOID)
         goto done;
 
     dwSize = 0;
-    GetTokenInformation(hToken,
-                        TokenGroups,
-                        NULL,
-                        0,
-                        &dwSize);
-    if (dwSize == 0)
+    if (GetTokenInformation(hToken,
```
Fraizeraust, line 183: `GetTokenInformation` returns a BOOL value. With your change the code literally hits the check when this function successfully retrieves details about the token....

Extravert-ir: This is how it supposed to be used: first you call `GetTokenInformation` with NULL buffer and expect it to return `ERROR_INSUFFICIENT_BUFFER`. Then you make a proper call

Extravert-ir, line 183: While you're here, please format this condition as others.

PR #3001, opened 2020-07-20
[CLASS2]: Set DiskSize member of deviceExtension->DiskGeometry in Scs…

coderTrevor:
…iClassReadDriveCapacity(). Consolidate code and comment for setting media capacity in bytes. CORE-17166

## Purpose

This resolves the issue with the size of disk drives not being detected & reported properly, which leads to BSOD's when ReactOS is installed and missing partitions when booting from a livecd. This resolves CORE-17166.

JIRA issue: [CORE-17166](https://jira.reactos.org/browse/CORE-17166)

## Proposed changes

We need to set the DiskSize member of the DISK_GEOMETRY_EX struct deviceExtension->DiskGeometry.

### DIFF
--- a/drivers/storage/class/class2/class2.c
+++ b/drivers/storage/class/class2/class2.c
@@ -828,19 +828,19 @@ Retry:
             lastSector + 1));
 
         //
-        // Calculate media capacity in bytes.
-        //
-
-        deviceExtension->PartitionLength.QuadPart = (LONGLONG)(lastSector + 1);
-
-        //
         // Calculate number of cylinders.
         //
 
         deviceExtension->DiskGeometry->Geometry.Cylinders.QuadPart = (LONGLONG)((lastSector + 1)/(DEFAULT_SECTORS_PER_TRACK * DEFAULT_TRACKS_PER_CYLINDER));
 
+        //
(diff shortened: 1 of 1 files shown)


drivers/storage/class/class2/class2.c
```diff
@@ -828,19 +828,19 @@ Return Value:
...
 
         //
-        // Calculate number of cylinders.
+        // Calculate media capacity in bytes.
         //
 
-        deviceExtension->DiskGeometry->Geometry.Cylinders.QuadPart = (LONGLONG)((lastSector + 1)/(DEFAULT_SECTORS_PER_TRACK * DEFAULT_TRACKS_PER_CYLINDER));
-
+        deviceExtension->PartitionLength.QuadPart = (LONGLONG)(lastSector + 1);
         deviceExtension->PartitionLength.QuadPart =
```
Extravert-ir, line 841: Any idea why this is not just `(lastSector + 1) << deviceExtension->SectorShift`?
Anyway, I'm merging this because the days of this driver are counted

coderTrevor: No good ideas. I just kept it that way to keep the changes minimal.

PR #3006, opened 2020-07-23
[REGEDIT] Improve importing registry files messagebox CORE-15494

oleg-dubinskiy:
## Purpose

Improve the messagebox responsible for importing the data from the *.reg file in Regedit.
Now it looks same as in Windows. The only difference is the translations, but I think it is not necessarily to make them identical to MS ones.

JIRA issue: [CORE-15494](https://jira.reactos.org/browse/CORE-15494)

## Proposed changes

- Remove "Cancel" button, since "No" and "Cancel" buttons do exactly the same thing. They both close the messagebox without applying the .reg file content and without terminating all other messageboxes if there is more than one.
Also remove IDCANCEL case for this.
- Disable Close button by setting the messagebox's window class style to `CS_NOCLOSE`, same as it done in Windows Regedit. Now it is not clickable, but still visible.
- Change the messagebox's icon type from warning icon to question icon, because this is done in Windows Regedit too.

## Result

Before:
regedit_before

- The "Cancel" button is here.
- The Close button is enabled (clickable).
- The message icon type is warning (exclamation).

After:
regedit_after

- There is no more "Cancel" button, only "Yes" and "No".
- The Close button is now disabled (not clickable).
- The message icon type is question.

### DIFF
--- a/base/applications/regedit/regedit.c
+++ b/base/applications/regedit/regedit.c
@@ -140,16 +140,31 @@ BOOL PerformRegAction(REGEDIT_ACTION action, LPWSTR s, BOOL silent)
         {
             WCHAR szText[512];
             WCHAR filename[MAX_PATH];
+            LPWSTR command_line = s;
+            UINT count = 0, mbType = MB_YESNO;
             FILE *fp;
 
-            get_file_name(&s, filename);
-            if (!filename[0])
+            get_file_name(&command_line, filename);
+            while (filename[0])
+            {
+                count++;
+                get_file_name(&command_line, filename);
+            }
+
+            if (count == 0)
             {
                 InfoMessageBox(NULL, MB_OK | MB_ICONINFORMATION, NULL, L"No file name is specified.");
                 InfoMessageBox(NULL, MB_OK | MB_ICONINFORMATION, szTitle, usage);
                 exit(4);
             }
 
+            if (count > 1)
+            {
+                /* Enable three buttons if we have more than one file */
+                mbType = MB_YESNOCANCEL;
+            }
+
+            get_file_name(&s, filename);
             while (filename[0])
             {
                 /* Request import confirmation */
@@ -159,7 +174,7 @@ BOOL PerformRegAction(REGEDIT_ACTION action, LPWSTR s, BOOL silent)
 
                     LoadStringW(hInst, IDS_IMPORT_PROMPT, szText, COUNT_OF(szText));
 
-                    choice = InfoMessageBox(NULL, MB_YESNOCANCEL | MB_ICONWARNING, szTitle, szText, filename);
+                    choice = InfoMessageBox(NULL, mbType | MB_ICONQUESTION, szTitle, szText, filename);
 
                     switch (choice)
                     {


HBelusca:
Please consider the comment that was present in the code:
```c
/* The cancel case is useful if the user is importing more than one registry file
   at a time, and wants to back out anytime during the import process. This way, the
   user doesn't have to resort to ending the regedit process abruptly just to cancel
   the operation. */
```

oleg-dubinskiy:
@HBelusca I already saw that comment, but my tests prove otherwise. Regarding to them, the "No" and "Cancel" buttons do exactly the same work. Clicking as the first, same the second one terminates only one (current) messagebox and one Regedit process. So looks like that thing is no longer actual. Did you see my latest two comments and video proof in the ticket?

learn-more:
You will see a change if you call regedit manually with the 2 files like this:
```
regedit 1.reg 2.reg
```

oleg-dubinskiy:
Oh, exactly! Thanks @learn-more!
When to run Regedit with those two reg files from CMD instead, the buttons indeed have a different effect!
Although such launching seems to be less familiar for the most of users, and MS Regedit has no this, is it still necessarily to restore it? Or it will be better to keep it as-is? :thinking:

binarymaster:
Please restore this functionality. When I looked at the changes, even without reading next comments I thought about applying multiple registry files in a loop where "No" skips one file, and "Cancel" cancels the operation.

binarymaster:
Also you may consider not showing "Cancel" button if there is just one file specified in arguments.

Prince-Kassad:
If my Windows knowledge doesn't fail me there are similar message boxes in other places that have the options "Yes", "Yes to All" and "No", but you can hold Ctrl and click No to get the effect "No to All". Maybe something similar could be implemented here...?

oleg-dubinskiy:
>"Yes", "Yes to All" and "No"... "No to All"

It seems to me that such buttons are used in the different message boxes (probably only in shell32 file operations dialogs), so we can't use them here, in standard message boxes.

base/applications/regedit/regedit.c
```diff
@@ -158,8 +158,9 @@ BOOL PerformRegAction(REGEDIT_ACTION action, LPWSTR s, BOOL silent)
                     int choice;
 
                     LoadStringW(hInst, IDS_IMPORT_PROMPT, szText, COUNT_OF(szText));
+                    SetClassLongPtrW(NULL, GCL_STYLE, GetClassLongPtrW(NULL, GCL_STYLE) | CS_NOCLOSE);
```
binarymaster, line 161: I have no idea how did you get a message box with a disabled close button using this trick :zany_face: Because `InfoMessageBox` function is specific to regedit, and it internally calls `MessageBoxW`, whose window style you cannot control. :thinking:

binarymaster: And it doesn't work for me when I apply the patch locally, as I expected. There are some hacky solutions available though: https://stackoverflow.com/questions/13189791/disable-x-button-icon-on-the-top-of-the-right-in-messagebox-using-c-win32-api (but I probably gonna just revert this)

PR #3007, opened 2020-07-24
[NTDLL_APITEST] Add NtXxxSystemEnvironmentValueEx testcases

Fraizeraust:
## Purpose
As `NtQuery/Set` functions aren't implemented in Windows Server 2003 either, the APITEST confirms this case.
**Windows Server 2003**
Capture
**ReactOS**
Capture2


Extravert-ir:
Does it make sense to test these functions if they are not implemented? :)

Fraizeraust:
You're actually kind of right, guess I'll just close the PR.

PR #3009, opened 2020-07-27
[IMM32] Stub ImmRegisterClient CORE-16479

oleg-dubinskiy:
## Purpose

Add a stub for `ImmRegisterClient` function in imm32.dll.
This function is currently undocumented by MS.
Based on the prototypes that we already have in
https://git.reactos.org/?p=reactos.git;a=blob;f=win32ss/user/user32/include/user32p.h;hb=9d2ecd3a13143609a30aca7480b195fa5f56df18#l39
and
https://git.reactos.org/?p=reactos.git;a=blob;f=win32ss/user/user32/misc/imm.c;hb=9d2ecd3a13143609a30aca7480b195fa5f56df18#l44.
Also I looked the following unofficial documentation: https://reactos.org/wiki/User:Alvinhochun/Localization/IME/IMM.
Although the function data type is `int` instead of `BOOL` there, those docs are currently very incomplete, so they may be wrong. I'm sure that our prototype is more correct.
It allows MS user32.dll no longer crash in ReactOS, so now ROS win32k Frankenstein is booting into desktop successfully even without replacing imm32.dll.

JIRA issue: [CORE-16479](https://jira.reactos.org/browse/CORE-16479)

## Result

Before:
imm32_before

After:
imm32_after


SergeGautherie:
Handle is a pointer, so
`@ stdcall ImmRegisterClient(ptr ptr)`

oleg-dubinskiy:
Fixed. Thanks @SergeGautherie.

katahiromz:
Sorry, there is conflict.

learn-more:
@oleg-dubinskiy the linked code suggests that the first argument is of the type `SHAREDINFO*`, which comes from `win32ss/include/ntuser.h`

oleg-dubinskiy:
@katahiromz @learn-more thanks. Everything is fixed.

PR #3011, opened 2020-07-30
[NTOS:MM] Replace ASSERT by UNIMPLEMENTED_ONCE macro for unimplemented code path CORE-16478

oleg-dubinskiy:
## Purpose

Replace the `ASSERT` by `UNIMPLEMENTED_ONCE` macro with checking the data of `ProtoPte` variable for unimplemented code path in `MiRemoveMappedPtes` function of our MM.
It seems to me that asserts should not be used in unimplemented code paths, so I replaced it by the correct macro instead.
It allows ReactOS win32k Frankenstein no longer hang (in debug builds) when using any apps which require layered window support (e. g. RocketDock 1.3.5) and also in some other cases, like when trying to use VirtualBox display integration mode.

JIRA issue: [CORE-16478](https://jira.reactos.org/browse/CORE-16478)

## Result

RocketDock 1.3.5:
RocketDock_MS_Win32SS_fixed

VirtualBox 5.1.38 display integration mode:
VBox_mode_after

### DIFF
--- a/ntoskrnl/mm/ARM3/section.c
+++ b/ntoskrnl/mm/ARM3/section.c
@@ -2303,7 +2303,7 @@ MiRemoveMappedPtes(IN PVOID BaseAddress,
                 ProtoPte = MiProtoPteToPte(&PteContents);
 
                 /* We don't support anything else atm */
-                ASSERT(ProtoPte->u.Long == 0);
+                if (ProtoPte->u.Long != 0) UNIMPLEMENTED_ONCE;
             }
         }
 


SergeGautherie:
1)
```
-                ASSERT(ProtoPte->u.Long == 0);
+                if (ProtoPte->u.Long != 0) UNIMPLEMENTED_ONCE;
```

2) Would `UNIMPLEMENTED` be enough or spam too much?

oleg-dubinskiy:
1. Thanks @SergeGautherie. :+1: This works for me too. So it's not required to comment out `ProtoPte` then.
2. As I see, yes, too much, to affect the performance in debug mode. 320 lines of the spam appeared after simple launching the dock, moving mouse cursor on it a few seconds and exiting. So I think `UNIMPLEMENTED_ONCE` will be better in this case. With the last one, there is no spam at all.

tkreuzer (changes requested):
ASSERTs exist for a reason. If it gets triggered, it usually means there is something seriously broken. And this is kernel code, so silencing the issue is a really bad idea. Please investigate where the problem is: What are the contents of the PTE when it hits? Use ```!pte <pte address>``` in WinDbg.

GeoB99:
I agree with Timo regarding this. Dismissing the assert is not only the bad way to fix this problem but is also a hack. Judging by the comment in the code I highly assume that specific part in `MiRemoveMappedPtes` has to be implemented.

Considering this is a stale PR I'm going to close this. Feel free to come back working on this in a proper way, let me know when you're ready as I'll open the PR back.

PR #3015, opened 2020-08-01
[NTOS] Improve bugcheck frontend

feel-the-dz3n:
## Purpose

Improve bugcheck frontend!

## Proposed changes

- Less text for descriptions;
- Better palette for eyes;
- Show a symbolic name of the code. (added in Windows since Vista!)

## TODO

- [ ] Resync with master branch to include a [bootvid fix](https://github.com/reactos/reactos/pull/2936);
- [ ] Review the ``bug.c`` code;
  - [ ] Make sure that current implementation of a symbolic name display is good...
- [ ] Modify all cases from ``bugcodes.mc``.


ntoskrnl/ke/bug.c
```diff
@@ -334,6 +334,420 @@ KiInitializeBugCheck(VOID)
     }
 }
 
+PCHAR
+NTAPI
+KeGetBugMessageName(IN ULONG MessageId)
```
feel-the-dz3n, line 339: I am totally sure that this incorrect way to get bugcheck code name. This code is almost two years old and I lost that tool source code lol.

ntoskrnl/ke/bug.c
```diff
@@ -610,29 +1024,57 @@ KiDumpParameterImages(IN PCHAR Message,
...
 
     /* Check if bootvid is installed */
     if (InbvIsBootDriverInstalled())
     {
         /* Acquire ownership and reset the display */
         InbvAcquireDisplayOwnership();
         InbvResetDisplay();
+    }
+
+    /* Load bugcheck bitmap */
```
HBelusca, line 1039: bugcheck bitmap for setting the palette :D

ntoskrnl/ke/bug.c
learn-more: You probably Want to use RtlFindMessage, like what is used here:
https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/ex/init.c;hb=89e716a7f71c57d5c46718feec6f9f562fc121d4#l1754

GeoB99: `RtlFindMessage` would only work if the bugcheck message description does only have the actual bugcheck title itself, which not all of them have it unfortunately. Consider this e.g.:
https://git.reactos.org/?p=reactos.git;a=blob;f=sdk/include/reactos/mc/bugcodes.mc;h=329b75a2e4e94631d5b102a1c13386e8d7256fa2;hb=HEAD#l1718

So you'd end up with `A critical kernel security check failed` instead of the bugcheck itself. Maybe adding an additional type block like `BugcheckTitle` in .mc and then implement a function like `RtlFindBugcheckTitle` that is similar to `RtlFindMessage` would help?

HBelusca: > So you'd end up with `A critical kernel security check failed` instead of the bugcheck itself. Maybe adding an additional type block like `BugcheckTitle` in .mc and then implement a function like `RtlFindBugcheckTitle` that is similar to `RtlFindMessage` would help?

This does not make sense. The .mc files are compiled and generate binary message resources (see how these look like when you open e.g. ntoskrnl.exe in a resource viewer/editor) and corresponding header files containing the list of #defines for their values. These keywords (SymbolicName etc...) are interpreted by the message compiler (mc.exe in Visual C++, windmc.exe in mingw etc.), and at best, adding a new "block" would do nothing at all, at worst it would emit an error saying the "block" is not recognized and fail. In any case you would not gain any thing at all.

ntoskrnl/ke/bug.c
```diff
@@ -334,6 +334,420 @@ KiInitializeBugCheck(VOID)
     }
 }
 
+PCHAR
+NTAPI
+KeGetBugMessageName(IN ULONG MessageId)
+{
+    /* Generated with https://github.com/feel-the-dz3n/kebughtoc */
```
HBelusca, line 341: I've just discovered that the `-x` [option of the message compiler](https://learn.microsoft.com/en-us/windows/win32/wes/message-compiler--mc-exe-#arguments-specific-to-message-text-files) (works for both MC.EXE and GNU windres.exe), allows to generate a separate source file containing a C-style mapping array of values in a manner similar to what you want to achieve:
```c
struct testbugcodesSymbolicName
{
  ULONG MessageId;
  char *SymbolicName;
} testbugcodesSymbolicNames[] =
{
  {(ULONG) WINDOWS_NT_BANNER, "WINDOWS_NT_BANNER" },
  {(ULONG) WINDOWS_NT_CSD_STRING, "WINDOWS_NT_CSD_STRING" },
....
  { (ULONG) 0xffffffff, NULL }
};
```
To find the string, one would have to write a for/while loop, instead of a switch.

PR #3016, opened 2020-08-03
[TASKMGR] Fix display of process command-line string

powpingdone:
## Purpose
Fixes issue [CORE-17115](https://jira.reactos.org/browse/CORE-17115)

## Analysis
The `CMD_LINE_CACHE` index (`idx`) does not update when programs are closed. The following fix is implemented: compare it with the cache's process id (`pid`) and the process id determined by the index in the same cache.

### DIFF
--- a/base/applications/taskmgr/perfdata.c
+++ b/base/applications/taskmgr/perfdata.c
@@ -5,6 +5,7 @@
  *
  *  Copyright (C) 1999 - 2001  Brian Palmer                <brianp@reactos.org>
  *  Copyright (C)        2014  Ismael Ferreras Morezuelas  <swyterzone+ros@gmail.com>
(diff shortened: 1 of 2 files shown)


learn-more:
> 
> 
> > Add initial support for linux-msvc on wine
> 
> Please avoid adding unrelated changes into PR, create another separate PR instead.

it's probably because the branch was created from his `master` branch...

powpingdone:
yeah whoops thats my fault

powpingdone:
sorry about that I *thought* I had created a different branch to work in. 

note to self: *always check* your branch

JoachimHenze:
Unfortunately this patch is still faulty and introduced a regression, see https://jira.reactos.org/browse/CORE-17115?focusedCommentId=124665&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-124665
We have to rework!

PR #3019, opened 2020-08-04
[WINESYNC] Fix syntax error in winesync.py

oleg-dubinskiy:
## Purpose

Fix a syntax error in winesync.py script.
It allows to run this script properly in Unix-based systems (at least on my Ubuntu 16.04.6 LTS with Python 2.7 and 3.5 versions, with all required dependencies already installed).

JIRA issue: None

## Proposed changes

- Use the double quotation marks instead of single ones at line 230, around "git commit --amend" debug print.

## TODO

- [ ] Check does it still work on Windows. Feel free to test it for yourself.

## Result

Before:
winesync py_before

After:
winesync py_after


oleg-dubinskiy:
Oops, I was a bit wrong. Sorry. :sweat_smile: 
The fstrings appeared starting from Python 3.6: http://zetcode.com/python/fstring/, so they are correct. No need to fix them.
I just used an outdated Python 3.5, which doesn't support them. Changing them in wrong way may broke the winesync.py code.
After upgrading my Python version, there appeared only one syntax error. When I fixed it, the script became working for me again. See updated PR description.
P. S.: I double-rechecked now: before my changes, I got that error with all Python versions from 3.6 to 3.9. After fixing it, the script runs properly for me with all of them. So that was indeed wrong.

oleg-dubinskiy:
Oh no, that line already has the double quotation marks in master. Looks like I mixed up something. My bad. :sweat_smile: 
Then that script is OK as-is and my PR makes no sense. Closing it. Sorry again.

PR #3021, opened 2020-08-06
[Win32SS][GDI] Fix pool memory disclosure in NtGdiGetGlyphOutline

khanhnt2:
## Purpose

Fix pool memory disclosure in NtGdiGetGlyphOutline.

## Analysis

The value of `pvBuf` will be initialized in function `ftGdiGetGlyphOutline`. But in some cases like https://github.com/reactos/reactos/blob/52fb8c1a8d1d1ef2ddb46c017b461011b26188ee/win32ss/gdi/ntgdi/freetype.c#L4016-L4017 `ftGdiGetGlyphOutline` will return without initializing `pvBuf` so it will lead to memory disclosure when execute this https://github.com/reactos/reactos/blob/50510538db49b58b3364a0938a82879bc08890a3/win32ss/gdi/ntgdi/font.c#L767-L768

I guess this is a bug pattern when initializer function does a comparison then return immediately without initializing.
## TODO

~~- [ ] Remove redundant RtlZeroMemory functions in `ftGdiGetGlyphOutline`~~

### DIFF
--- a/win32ss/gdi/ntgdi/font.c
+++ b/win32ss/gdi/ntgdi/font.c
@@ -743,13 +743,12 @@ NtGdiGetGlyphOutline(
 
   if (UnsafeBuf && cjBuf)
   {
-     pvBuf = ExAllocatePoolWithTag(PagedPool, cjBuf, GDITAG_TEXT);
+     pvBuf = ExAllocatePoolZero(PagedPool, cjBuf, GDITAG_TEXT);
      if (!pvBuf)
      {
         EngSetLastError(ERROR_NOT_ENOUGH_MEMORY);
         goto Exit;
      }
-     RtlZeroMemory(pvBuf, cjBuf);
   }
 
   Ret = ftGdiGetGlyphOutline( dc,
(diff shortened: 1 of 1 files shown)


khanhnt2:
I doubt that my fix is correct. Should I use RtlZeroMeory in the function `ftGdiGetGlyphOutline` to fix? For example, the fix will look like

```
 if (needed > cjBuf) {
     RtlZeroMemory(pvBuf, cjBuf); 
     return GDI_ERROR; 
 }
```

But there are a lot of comparison like that in the function `ftGdiGetGlyphOutline`.

win32ss/gdi/ntgdi/font.c
```diff
@@ -749,6 +749,7 @@ NtGdiGetGlyphOutline(
         EngSetLastError(ERROR_NOT_ENOUGH_MEMORY);
         goto Exit;
      }
+     RtlZeroMemory(pvBuf, cjBuf);
```
Extravert-ir, line 752: Let's use `ExAllocatePoolZero` since we have it in our sdk now

khanhnt2: Thanks for your advice, I have changed that function.

khanhnt2:
I realize that ftGdiGetGlyphOutline is not only used by NtGdiGetGlyphOutline, so I guess I shouldn't remove redundant RtlZeroMemory in ftGdiGetGlyphOutline.

learn-more (approved):

PR #3029, opened 2020-08-08
[NTOS:RTL] Implement the kernel-mode version of RtlGetNtProductType

Fraizeraust:
Implement `RtlGetNtProductType` in the kernel mode RTL and let `NtSetSystemPowerState` use it (this has to be investigated in the future why and how is the product type value used). 

The documentation of this function can be found here -> https://fraizeraust.github.io/pages/doc/ntoskrnl/rtl/prodver.html

### DIFF
--- a/ntoskrnl/rtl/misc.c
+++ b/ntoskrnl/rtl/misc.c
@@ -19,6 +19,7 @@ extern ULONG NtGlobalFlag;
 extern ULONG NtMajorVersion;
 extern ULONG NtMinorVersion;
 extern ULONG NtOSCSDVersion;
+#define PRODUCT_TAG 'iPtR'
 
 /* FUNCTIONS *****************************************************************/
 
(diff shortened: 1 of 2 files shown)


ntoskrnl/rtl/misc.c
```diff
@@ -62,6 +62,123 @@ RtlGetVersion(IN OUT PRTL_OSVERSIONINFOW lpVersionInformation)
...
+RtlGetNtProductType(OUT PNT_PRODUCT_TYPE ProductType)
+{
+    HANDLE Key;
+    ULONG ReturnedLength;
+    OBJECT_ATTRIBUTES ObjectAttributes;
+    NTSTATUS Status;
+    UCHAR Buffer[sizeof(PKEY_VALUE_PARTIAL_INFORMATION) + 256 * sizeof(WCHAR)];
+    PKEY_VALUE_PARTIAL_INFORMATION PartialInformation = (PVOID)Buffer;
+    UNICODE_STRING WorkstationProduct;
+    UNICODE_STRING LanManProduct;
```
learn-more, line 97: Why not:
```
static UNICODE_STRING LanManProduct = RTL_CONSTANT_STRING(L"LanmanNT"); 
```

ntoskrnl/rtl/misc.c
```diff
@@ -62,6 +62,121 @@ RtlGetVersion(IN OUT PRTL_OSVERSIONINFOW lpVersionInformation)
...
+ * @return
+ *      The function returns TRUE when the call successfully returned the type product of the system.
+ *      It'll return FALSE on failure otherwise. In the latter case the function will return WinNT
+ *      as the default product type.
+ *
+ * @remarks
+ *      The call expects to be called at PASSIVE_LEVEL. This is the base implementation of the function
+ *      as there's stuff still left to be investigated. For instance the function uses ExAllocatePoolWithTag
+ *      with the tag as "iPtR". In addition to that the function checks if the product type is actually valid
+ *      by checking the "ProductTypeIsValid" member of _KUSER_SHARED_DATA structure. Currently we do not
```
learn-more, line 83: Would it be an idea to implement this code?
So that this function is 'complete' as soon as the path setting the ProductTypeIsValid to true is implemented?

ntoskrnl/rtl/misc.c
```diff
@@ -62,6 +62,121 @@ RtlGetVersion(IN OUT PRTL_OSVERSIONINFOW lpVersionInformation)
...
+    {
+        /* The following product is an OS Advanced Server */
+        *ProductType = NtProductLanManNt;
+        ZwClose(Key);
+        return TRUE;
+    }
+    else
+    {
+        /* The following product is an OS Server */
+        *ProductType = NtProductServer;
```
learn-more, line 168: Are you sure everything else is `NtProductServer`, and not `NtProductWinNt`?

ntoskrnl/rtl/misc.c
```diff
@@ -62,6 +62,121 @@ RtlGetVersion(IN OUT PRTL_OSVERSIONINFOW lpVersionInformation)
...
+    }
+
+    /* Now it's time to query the value from the key to check what kind of product the OS is */
+    Status = ZwQueryValueKey(Key,
+                             &ValueNtProduct,
+                             KeyValuePartialInformation,
+                             PartialInformation,
+                             sizeof(Buffer),
+                             &ReturnedLength);
+    if (!NT_SUCCESS(Status))
```
learn-more, line 126: If you close the key before doing this check, you can remove the `ZwClose` from all branches below

ntoskrnl/po/power.c
```diff
@@ -1118,6 +1119,16 @@ NtSetSystemPowerState(IN POWER_ACTION SystemAction,
         break;
     }
 
+    /*
+     * TODO: Investigate in the future why Windows Server 2003
+     * needs to get the NT product type during a system power state
+     * action.
+    */
+    if (!RtlGetNtProductType(&ProductType))
```
JoachimHenze, line 1127: As long as we do not know why MS does this call, I see absolutely no benefit in adding it here.
Stuff that is called without any reason is wasted time, bloated binary size, and additional risk just, especially in the kernel.
If you absolutely feel the need to add it to rememeber that MS does call it here, then please at least comment **all the code** that you added within power.c!

JoachimHenze (changes requested):

ntoskrnl/po/power.c
ThFabba: Agreed, we don't need dead code.
FYI, the reason Windows does this seems to have something to do with writing event log entries during hibernation.

Fraizeraust: Yeah, apologies for that. I'll address the comment reviews in the next force push.

ntoskrnl/rtl/misc.c
```diff
@@ -62,6 +63,140 @@ RtlGetVersion(IN OUT PRTL_OSVERSIONINFOW lpVersionInformation)
...
+    if (BufferKey->Type != REG_SZ)
+    {
+        /* We've got something else, so bail out */
+        DPRINT1("RtlGetNtProductType(): An invalid value type has been found!\n");
+        Success = FALSE;
+        goto Exit;
+    }
+
+    /* Initialize the Unicode string with the data from the registry value */
+    RtlInitUnicodeString(&NtProductType, (PWCHAR)BufferKey->Data);
```
learn-more, line 161: You can't assume this is null terminated.
It is probably better to use `DataLength` (which _can_ include the null terminator)

ntoskrnl/rtl/misc.c
```diff
@@ -62,6 +63,141 @@ RtlGetVersion(IN OUT PRTL_OSVERSIONINFOW lpVersionInformation)
...
+    {
+        /* We've got something else, so bail out */
+        DPRINT1("RtlGetNtProductType(): An invalid value type has been found!\n");
+        Success = FALSE;
+        goto Exit;
+    }
+
+    /* Initialise the Unicode string with the data from the registry value */
+    NtProductType.Length = NtProductType.MaximumLength = BufferKey->DataLength;
+    NtProductType.Buffer = (PWCHAR)BufferKey->Data;
```
learn-more, line 162:
```suggestion
    NtProductType.Buffer = (PWCHAR)BufferKey->Data;
    if (NtProductType.Length > sizeof(WCHAR) && NtProductType.Buffer[NtProductType.Length / sizeof(WCHAR) -1] == UNICODE_NULL)
        NtProductType.Length -= sizeof(WCHAR)
```

ntoskrnl/rtl/misc.c
```diff
@@ -62,6 +63,147 @@ RtlGetVersion(IN OUT PRTL_OSVERSIONINFOW lpVersionInformation)
...
+
+    if (NtProductType.Length > sizeof(WCHAR) && NtProductType.Buffer[NtProductType.Length /
+        sizeof(WCHAR) - 1] == UNICODE_NULL)
+    {
+        NtProductType.Length -= sizeof(WCHAR);
+    }
+
+    /* Now it's time to get the product based on the value data */
+    if (RtlCompareUnicodeString(&NtProductType, &WorkstationProduct, FALSE) == 0)
```
learn-more, line 171: Should this check be done case sensitive?

JoachimHenze (approved):
My concern has been addressed now.

ntoskrnl/rtl/misc.c
```diff
@@ -62,6 +63,147 @@ RtlGetVersion(IN OUT PRTL_OSVERSIONINFOW lpVersionInformation)
     return STATUS_SUCCESS;
 }
 
+/**
+ * @RtlGetNtProductType
+ *
```
binarymaster, line 68:
```suggestion
 * @brief
```

ntoskrnl/rtl/misc.c
```diff
@@ -62,6 +63,147 @@ RtlGetVersion(IN OUT PRTL_OSVERSIONINFOW lpVersionInformation)
...
+ * @return
+ *      The function returns TRUE when the call successfully returned the type product of the system.
+ *      It'll return FALSE on failure otherwise. In the latter case the function will return WinNT
+ *      as the default product type.
+ *
+ * @remarks
+ *      The call expects to be called at PASSIVE_LEVEL. The function firstly checks if the product type is
+ *      actually valid by checking the "ProductTypeIsValid" member of _KUSER_SHARED_DATA structure.
+ *      Currently we do not implement code that is responsible for the management of this member, yet.
+*/
```
binarymaster, line 84:
```suggestion
 *
 */
```

ntoskrnl/rtl/misc.c
```diff
@@ -62,6 +63,147 @@ RtlGetVersion(IN OUT PRTL_OSVERSIONINFOW lpVersionInformation)
...
+    NTSTATUS Status;
+    PKEY_VALUE_PARTIAL_INFORMATION BufferKey;
+    ULONG BufferKeyLength = sizeof(PKEY_VALUE_PARTIAL_INFORMATION) + (256 * sizeof(WCHAR));
+    UNICODE_STRING NtProductType;
+    static UNICODE_STRING WorkstationProduct = RTL_CONSTANT_STRING(L"WinNT");
+    static UNICODE_STRING LanManProduct = RTL_CONSTANT_STRING(L"LanmanNT");
+    static UNICODE_STRING ServerProduct = RTL_CONSTANT_STRING(L"ServerNT");
+    static UNICODE_STRING KeyProduct = RTL_CONSTANT_STRING(L"\\Registry\\Machine\\SYSTEM\\CurrentControlSet\\Control\\ProductOptions");
+    static UNICODE_STRING ValueNtProduct = RTL_CONSTANT_STRING(L"ProductType");
+    PAGED_CODE();
```
binarymaster, line 102:
```suggestion

    PAGED_CODE();
```

ntoskrnl/rtl/misc.c
```diff
@@ -62,6 +63,147 @@ RtlGetVersion(IN OUT PRTL_OSVERSIONINFOW lpVersionInformation)
...
+    {
+        DPRINT1("RtlGetNtProductType(): The ZwQueryValueKey() function has failed! (Status: 0x%lx)\n", Status);
+        Success = FALSE;
+        goto Exit;
+    }
+
+    /*
+     * Do a sanity check before we get the product type of the operating system
+     * so that we're sure the type of the value we've got is actually correct.
+    */
```
binarymaster, line 151:
```suggestion
     */
```

binarymaster (commented):
Nitpicks. See also https://reactos.org/wiki/Documentation_Guidelines#API_Documentation

tkreuzer:
@Fraizeraust I don't understand why this needs to fetch it from the registry, instead of pulling it from SharedUserData. The latter should just be the cached value of this.

Fraizeraust:
@tkreuzer When I stumbled upon the function calls of this routine like `ZwQueryValueKey` et al in a Windows Server 2003 checked build, I thought at the first glance the first thing this function ever does is to query the registry to gather the NT product type value whereas the user-mode equivalent of this function in NTDLL is the one that fetches the said value from `SharedUserData`.

tkreuzer:
This is done in the kernel already. 
ExpInitializeExecutive -> CmGetSystemControlValues -> Queries the value and puts it into MmProductType (stores the first 4 bytes as an ULONG, which is ridiculous, but that's how we do it). This is then evaluated in MmArmInitSystem, which sets SharedUserData->NtProductType appropriately. So the only thing that this code would change is the behavior, when you change the product type in the registry at runtime. And that doesn't even seem to be possible. At least on Windows 2k3 it's not possible as admin, on Win 10 it's not even possible as system. Both result in access violation error. So my conclusion is: all of this code is useless.

PR #3030, opened 2020-08-08
[SHELL32] Optimize change notification

katahiromz:
## Purpose

JIRA issue: [CORE-13950](https://jira.reactos.org/browse/CORE-13950)

- Keep the directory lists only.
- Don't remember file sizes and normal file paths.


katahiromz:
AFTER:
after
Under investigation...

katahiromz:
Win10:
Win10
The testcase is wrong...

katahiromz:
The test results are no change. I believe this PR is improvement.

PR #3034, opened 2020-08-10
[NTUSER] Add more QUERY_WINDOW_* values

katahiromz:
## Purpose

- Add `QUERY_WINDOW_*` constant values for `NtUserQueryWindow` funciton.


katahiromz:
I added them for future coders.

learn-more:
You might consider adding a trace there, showing that it needs to be implemented?
That way it is easier to notice when they are called and not handled correctly

PR #3038, opened 2020-08-11
[SHELL32_APITEST] Strengthen SHChangeNotify testcase more

katahiromz:
## Purpose

JIRA issue: [CORE-13950](https://jira.reactos.org/browse/CORE-13950)

- Add tests on `SHCNRF_InterruptLevel` and `SHCNRF_RecursiveInterrupt`.
- Fix and improve codes.
- Don't use clipboard for data exchange.
- Close Explorer Cabinets before tests.

PR #3042, opened 2020-08-12
[XDK] Fix compilation with MSVC version 19.28

wjk:
## Purpose

This version of MSVC (included in VS 16.8 preview 1) introduces a change wherein it treats the definition of any intrinsic function (such as those in xdk/interlocked.h) as a fatal error. Use of `/Oi-` to suppress intrinsics use does not help here. I have verified that this code is still valid under GCC.

JIRA issue: [CORE-17190](https://jira.reactos.org/browse/CORE-17190)

## Proposed changes

- `#if` out the offending code.

### DIFF
--- a/sdk/include/xdk/interlocked.h
+++ b/sdk/include/xdk/interlocked.h
@@ -299,6 +299,7 @@
 #define InterlockedXor64NoFence __NF_(_InterlockedXor64)
 #endif /* _M_ARM */
 
+#if !(defined(_MSC_VER) && _MSC_VER >= 1928)
 #ifdef _M_IX86
 
 FORCEINLINE
@@ -535,4 +536,5 @@ _InterlockedBitTestAndComplement64(
 }
 
 #endif /* M_IA64 */
+#endif /* MSVC >= 1928 */
 


tkreuzer:
@wjk Did you test x64 build?

wjk:
[Actions did](https://github.com/reactos/reactos/pull/3042/checks?check_run_id=973953634), and it seems to be OK.

wjk:
Then again, maybe not. While the build succeeds I downloaded the MSVC x64 ISOs, and neither of them boots. 🤦 It gets as far as the `KdpPrintBanner` function; then, the screen freezes while the serial output is filled with gibberish. (First, the line of dashes is printed over and over — which is how I identified the function to begin with — and then garbage characters spewed out until I killed QEMU.)

Unless this is expected (as I know amd64 MSVC builds are very buggy), my change broke something badly. Unfortunately, without this change ReactOS won't compile at all, as detailed in the issue. I'm not certain what to do.

wjk:
However, the i386 MSVC livecd doesn't do this. The serial output is that single line repeated ~10 times, but after that no more output is printed, and the livecd successfully boots to desktop. My first guess is that MSVC builds are programmed to expect Microsoft's binary KD protocol, and I am that is what I am seeing on the serial port.

wjk:
@tkreuzer As far as I can tell (and currently that's not that far, as neither VirtualBox nor VMware Workstation are currently connecting to WinDBG), this gibberish-on-COM1 is completely expected behavior. From what I can tell reading the kdcom.dll sources, after sending a data packet (e.g. the one containing that banner string), kdcom it waits for an acknowledgment packet from WinDBG. It will loop continuously while listening for such a packet, and since WinDBG is not connected, it will loop forever.

Furthermore, I tried downloading the MSVC-AMD64 build of #3038 from GitHub Actions’ cache, and it exhibited the exact same behavior as the MSVC-AMD64 builds of this PR. Therefore, I believe the behavior I described in my above posts are **not** caused by the changes in this PR.

Also, while you’re here, I tried to see if I could get your version-management code in #1872 to work, and it works flawlessly with no changes needed (at least under GCC; I was testing MSVC when I ran into this problem). I also wrote a (working) unit test, which your PR lacks. Seeing as #1872 has seemingly stalled (no activity since May), I would like to send in my changes as a second PR, superseding yours. I just wanted to let you know about this before I opened it, to avoid any misunderstandings. Thanks so much!

binarymaster:
9605110 was committed by @ThFabba, is this PR still actual?

ThFabba:
Uh, sorry I totally didn't see this PR. Jira didn't show it linked in the ticket for some reason.
Yes, my comment supersedes this. Thanks for working on it, though!

learn-more:
> Uh, sorry I totally didn't see this PR. Jira didn't show it linked in the ticket for some reason.
> Yes, my comment supersedes this. Thanks for working on it, though!

Jira only shows the ticket if it is mentioned in the commit comment, not when it is mentioned in the PR description afaik.

PR #3055, opened 2020-08-15
Fix build with Clang-Cl and add it back to CI

ThFabba:
JIRA issue: [CORE-17202](https://jira.reactos.org/browse/CORE-17202)


learn-more (approved):

sdk/cmake/msvc.cmake
```diff
@@ -51,7 +51,12 @@ endif()
 # HACK: for VS 11+ we need to explicitly disable SSE, which is off by
 # default for older compilers. See CORE-6507
 if(ARCH STREQUAL "i386")
-    add_compile_flags("/arch:IA32")
+    # Clang's IA32 means i386, which doesn't have cmpxchg8b
```
HBelusca, line 54: So what kind of value Clang is actually using (what's the value of ${OARCH}) when RosBE's specified ARCH is i386?

sdk/include/xdk/ntdef.template.h
```diff
@@ -220,7 +220,7 @@ char _RTL_CONSTANT_STRING_type_check(const void *s);
 #endif
 #define RTL_CONSTANT_STRING(s) { \
     sizeof(s)-sizeof((s)[0]), \
-    sizeof(s) / sizeof(_RTL_CONSTANT_STRING_type_check(s)), \
+    sizeof(s) / (sizeof(_RTL_CONSTANT_STRING_type_check(s))), \
```
HBelusca, line 223: What kind of error does Clang emit here? (these extra parentheses look really superfluous)

learn-more: See the commit message?
It includes the full output

HBelusca: Hm yeah; the construct is actually quite strange. Was it based on what MS' WDK does have?

sdk/cmake/msvc.cmake
Extravert-ir: > which doesn't have cmpxchg8b

But this instruction is required for ReactOS to work, isn't it?

ThFabba: It defaults to `pentium` but can be overridden. https://github.com/reactos/reactos/blob/master/sdk/cmake/config.cmake#L6

Yes, `cmpxchg8b` is required, which is exactly why things won't work with `i386` and require `pentium`.

sdk/include/xdk/ntdef.template.h
ThFabba: Yes, it's based on WDK headers.
It's simply a `/ 1`, i.e. has no effect -- but the `_RTL_CONSTANT_STRING_type_check` makes it type-safe in C++.

PR #3057, opened 2020-08-16
[TASKMGR] Fix race issue(s) brought about by pr #3016

powpingdone:
## Purpose

Fix issues created in pull request #3016 and/or previously created in the taskmgr executable.

JIRA issue: [CORE-17115](https://jira.reactos.org/browse/CORE-17115)

## TODO

- [x] Fix race issue causing deallocation of memory during high process volume.

### DIFF
--- a/base/applications/taskmgr/perfdata.c
+++ b/base/applications/taskmgr/perfdata.c
@@ -572,6 +572,33 @@ BOOL PerfDataGetCommandLine(ULONG Index, LPWSTR lpCommandLine, ULONG nMaxCount)
     /* [A] Search for a string already in cache? If so, use it. If not, check if invalid and free it if so. */
     while (cache && cache->pnext != NULL)
     {
+        EnterCriticalSection(&PerfDataCriticalSection);
+
+        /* Because threading *might* free the cache we're reading from, check if we still exist. */
+        if (cache == NULL)
+        {
+            if (old_cache != NULL)
+            {
+                /* Use the previous cache that wasnt deallocated. */
+                cache = old_cache;
+            }
+            else if (global_cache != NULL)
+            {
+                /* Start at the beginning, because we have no idea where we are. */
+                cache = global_cache;
+            }
+            else
+            {
+                /* Everything is screwed, lets pretend that this never happened. */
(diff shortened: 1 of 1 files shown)


JoachimHenze:
Hi powpingdone, thanks for working on it. I would recommend to just limit to fixing that single regression within this PR, that was introduced yesterday. And leave all the other TODOs (that atm still unchecked anyway) for another PR. This way we will get this regression-fix faster into our codebase. And things are easier to review then. Thank you!

powpingdone:
Gotcha, splitting pr.

base/applications/taskmgr/perfdata.c
```diff
@@ -572,21 +572,48 @@ BOOL PerfDataGetCommandLine(ULONG Index, LPWSTR lpCommandLine, ULONG nMaxCount)
...
+            }
+            else if (global_cache != NULL)
+            {
+                /* Start at the beginning, because we have no idea where we are. */
+                cache = global_cache;
+            }
+            else
+            {
+                /* Everything is screwed, lets pretend that this never happened. */
+                break;
```
binarymaster, line 593: Isn't `LeaveCriticalSection(&PerfDataCriticalSection);` needed here before break?

powpingdone:
That's going to be another fix, separate pr as per requested by @JoachimHenze but do you want it as part of this pr?

binarymaster:
Let's wait for his answer then.

base/applications/taskmgr/perfdata.c
powpingdone: Yes, yes it is.

JoachimHenze:
Well of course, when I said: "Let's fix what broke by PR3016", I meant "all that broke by PR3016". Obviously the symptom of it "stops updating its process list" did not happen before PR3016+3057. So yes, fixing that should be part of the solution as well.

After all, just plain reverting 3016 and then trying to fix CORE-17115 from scratch, might be the test way to do, because then we have all the time we need and are not in any hurry.
So that's what I suggest now: just revert 3016, and then let powpingdone try with a completely fresh approach to the initial problem. We can avoid stumbling from regression to regression that way.

powpingdone:
Gotcha. Ill close this then.

PR #3066, opened 2020-08-17
[TASKMGR] PerfData*CommandLine*(): Do not skip 'global_cache' last entry

SergeGautherie:
Avoids:
* Get: Playing hide and seek.
* Exit: Leaking.

Addendum to f43bb8d (r65469).

### DIFF
--- a/base/applications/taskmgr/perfdata.c
+++ b/base/applications/taskmgr/perfdata.c
@@ -566,7 +566,7 @@ BOOL PerfDataGetCommandLine(ULONG Index, LPWSTR lpCommandLine, ULONG nMaxCount)
     PCMD_LINE_CACHE cache = global_cache;
 
     /* [A] Search for a string already in cache? If so, use it */
-    while (cache && cache->pnext != NULL)
+    while (cache)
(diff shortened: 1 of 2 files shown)


JoachimHenze:
SergeGautherie did never claim it would, but since this chat comment https://chat.reactos.org/reactos/pl/9i9ujyiu6tyxfdy7xn6bfjqdnc implied that this fix might fix the crash, that was introduced by 0.4.15-dev-523-g535e262 (and mentioned in CORE-17115),
I tested that and I can say for sure, that it does not. This PR here is not enough to fix that crash-regression!

That does not mean that Serge Gautheries patch is wrong. I just say that it does not help for what people hoped it would.
And my position is, that we should fix the regression first instead, before deciding whether we touch anything above that by merging Serges PR.

binarymaster (changes requested):
These changes are based on outdated head (stuff got reverted), please resolve conflicts or close PR.

Extravert-ir:
Please open a new PR for taskmgr with all changes you want put there

PR #3068, opened 2020-08-17
[TASKMGR] PerfDataGetCommandLine(): Remove useless 'cache->str' check

SergeGautherie:
`str` is always assigned and non-NULL.

Addendum to f43bb8d (r65469).


Extravert-ir:

PR #3069, opened 2020-08-17
[TASKMGR] PerfDataGetCommandLine(): Fix return value when failed

SergeGautherie:
Addendum to f43bb8d (r65469).


Extravert-ir:

PR #3071, opened 2020-08-18
[RAPPS] cmdline package name

kernelbin:
## Purpose

Original, rapps is using software's display name as a identifier in command-line.
(but the display name might be too long, contains strange chars. More over,display name might varies in different locale).
so I made this change.
currently, the package-name of the software corresponds with the ini file name in DB (without ```.txt``` extension)

this example shows the difference now to install codeblocks before/after this PR

**before**
```rapps /INSTALL "Code::Blocks (without compiler)"```

**after**
```rapps /INSTALL codeblocks```

## TODO
- [x] display package name in GUI too.

### CODE REVIEW

File: base/applications/rapps/available.cpp [hunk 1]
```diff
@@ -38,6 +38,13 @@ VOID CAvailableApplicationInfo::RetrieveGeneralInfo(AvailableStrings& AvlbString
 {
     m_Parser = new CConfigParser(m_sFileName);
 
+    // TODO: I temporarily use the file name (without suffix) as package name.
+    // It should be better to put this in a field of ini file.
+    // consider write a converter to do this and write a github action for rapps-db to ensure package_name is unique.
```
learn-more, line 43: As noted before, I am in favor of keeping this the filename, and only adding a key when needed. :)

base/applications/rapps/unattended.cpp
```diff
@@ -22,13 +22,12 @@ BOOL UseCmdParameters(LPWSTR lpCmdLine)
         return FALSE;
     }
 
-    // TODO: use DB filenames as names because they're shorter
-    ATL::CSimpleArray<ATL::CStringW> arrNames;
+    ATL::CSimpleArray<ATL::CStringW> PkgNameList;
     if (!StrCmpIW(argv[1], CMD_KEY_INSTALL))
     {
         for (INT i = 2; i < argc; ++i)
```
learn-more, line 28: You probably want to abort this loop (maybe in a later pr?) when you encounter another `/` or `-` prefixed option

base/applications/rapps/unattended.cpp
```diff
@@ -43,12 +42,12 @@ BOOL UseCmdParameters(LPWSTR lpCmdLine)
         INFCONTEXT Context;
         if (SetupFindFirstLineW(InfHandle, L"RAPPS", L"Install", &Context))
```
learn-more, line 43: This branch is for passing a .inf (or .ini) file to rapps

learn-more: (and most likely what inspired the filename)

base/applications/rapps/unattended.cpp
kernelbin: yes, I'd like to improve command-line parsing in next PR😉

base/applications/rapps/available.cpp
kernelbin: So I left it here🤣
but I think adding ```m_szPkgName``` is still needed (as it will be used more than once and I don't want to parse the filename to remove suffix again and again😝)

base/applications/rapps/unattended.cpp
kernelbin: hmm which filename?

learn-more: unattend.cpp

PR #3075, opened 2020-08-18
[SHELL32] Enable CopyTo/MoveTo textbox functionality

katahiromz:
## Purpose

JIRA issue: [CORE-11132](https://jira.reactos.org/browse/CORE-11132)

- The user can enter the full path of a directory in `"Copy To Folder"` and `"Move To Folder"` features.
- Fix two bugs in `SHBrowseForFolder` function. One is disability of text box to enter a full path. Another one is disability to choose the root item in `BFFM_SETSELECTION`.
- Add `ES_AUTOHSCROLL` style to text box. This enables the user to enter a long text.


dll/win32/shell32/CCopyToMenu.cpp
```diff
@@ -45,7 +45,9 @@ HRESULT _GetCidlFromDataObject(IDataObject *pDataObject, CIDA** ppcida)
 CCopyToMenu::CCopyToMenu() :
     m_idCmdFirst(0),
     m_idCmdLast(0),
-    m_idCmdCopyTo(-1)
+    m_idCmdCopyTo(-1),
+    m_fnOldWndProc(NULL),
+    m_bInSettingText(FALSE)
```
HBelusca, line 50: What is this flag for?

katahiromz: For checking/locking text update status. See `WindowProc` and `BrowseCallbackProc`.

katahiromz: If `m_bInSettingText` is TRUE, then it is in processing of `EN_CHANGE` or `BFFM_SELCHANGED`.

katahiromz: Renamed the variable as `m_bIgnoreTextBoxChange` and simplified the code.

dll/win32/shell32/CMoveToMenu.cpp
```diff
@@ -33,6 +36,36 @@ WindowProc(HWND hwnd, UINT uMsg, WPARAM wParam, LPARAM lParam)
...
+                    {
+                        if (!this_->m_bIgnoreTextBoxChange)
+                        {
+                            // get the text
+                            GetDlgItemTextW(hwnd, IDC_BROWSE_FOR_FOLDER_FOLDER_TEXT, szPath, _countof(szPath));
+                            StrTrimW(szPath, L" \t");
+
+                            // update OK button
+                            BOOL bValid = !PathIsRelative(szPath) && PathIsDirectoryW(szPath);
+                            SendMessageW(hwnd, BFFM_ENABLEOK, 0, bValid);
```
learn-more, line 56: Instead of adding an additional path for this,
would it be an idea to convert it to a pidl, and set this pidl as if it was selected in the treeview? (and possibly even select it)
In that way, it might only need some code in the EN_CHANGE here, and the rest could magically work.

katahiromz: After `BFFM_SELCHANGED` callback, `OK` button will be enabled or disabled by the system. `PostMessageW` will be overwrite the `OK` button enabled/disabled status by message queueing.

katahiromz: We have to distinguish between the relative path and the absolute path. Setting the pidl will store the file title (not absolute path) into the text box. Setting the pidl will cause text box change, text box change will set the pidl, then an infinite recursive loop will happen.

katahiromz: >would it be an idea to convert it to a pidl, and set this pidl as if it was selected in the treeview? (and possibly even select it)

It needs another flag to avoid the infinite loop...

katahiromz: Setting a PIDL will store the file title. This behavior is inevitable.

PR #3080, opened 2020-08-19
[VIDEOPRT] Define debug variable

disean:
## Purpose

Add useful variable for debugging video drivers. See https://docs.microsoft.com/ru-ru/previous-versions/ff570170(v=vs.85)

### DIFF
--- a/win32ss/drivers/videoprt/videoprt.c
+++ b/win32ss/drivers/videoprt/videoprt.c
@@ -31,6 +31,8 @@
 
 /* GLOBAL VARIABLES ***********************************************************/
 
+ULONG VideoDebugLevel = 0;
+
 BOOLEAN VpBaseVideo = FALSE;
 BOOLEAN VpNoVesa = FALSE;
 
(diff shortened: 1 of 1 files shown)


win32ss/drivers/videoprt/videoprt.c
```diff
@@ -794,6 +798,9 @@ VideoPortDebugPrint(
 {
     va_list ap;
 
+    if (VideoDebugLevel >= DebugPrintLevel)
+        DebugPrintLevel = Error;
```
binarymaster, line 802: Shouldn't it be something like...
```suggestion
    if (VideoDebugLevel != DebugPrintLevel)
        DebugPrintLevel = VideoDebugLevel;
```
Because in your code `DebugPrintLevel` is always assigned with same value. Or I'm not understanding something here. :sweat_smile:

win32ss/drivers/videoprt/videoprt.c
```diff
@@ -294,6 +296,8 @@ IntVideoPortFindAdapter(
                                       SystemBasicInfo.PageSize;
     }
 
+    // FIXME: Check the adapter key and update VideoDebugLevel variable.
```
binarymaster, line 299: It would be nice to fix this thing in this PR :) Probably using `RtlQueryRegistryValues` function.

win32ss/drivers/videoprt/videoprt.c
disean: There's nothing wrong here. The value should be updated on each boot (See FIXME).

PR #3084, opened 2020-08-20
[ADVAPI32] SaferComputeTokenFromLevel(): Stub-plement 'SAFER_TOKEN_NULL_IF_EQUAL' case

SergeGautherie:
Testcase:
LiveCD > XP cmd.exe > regtest.cmd (Twice, to better see.)

For the record, it now triggers an additional log:
`(/dll/win32/kernel32/client/vdm.c:1375) CmdBatNotification() is UNIMPLEMENTED!`

JIRA issue: [CORE-14015](https://jira.reactos.org/browse/CORE-14015)

### DIFF
--- a/dll/win32/advapi32/sec/safer.c
+++ b/dll/win32/advapi32/sec/safer.c
@@ -155,6 +155,26 @@ SaferComputeTokenFromLevel(
     _Inout_opt_ PVOID pReserved)
 {
     FIXME("(%p, %p, %p, 0x%lx, %p) stub\n", LevelHandle, InAccessToken, OutAccessToken, dwFlags, pReserved);
+
+    if (!InAccessToken)
+    {
+        FIXME("(In=NULL): Get thread/process token\n");
+    }
+
+    if (dwFlags == SAFER_TOKEN_NULL_IF_EQUAL)
+    // TODO: if (dwFlags & SAFER_TOKEN_NULL_IF_EQUAL != 0)
+    {
+        // TODO: Compare should eventually be done in other cases too.
+        FIXME("(F=SAFER_TOKEN_NULL_IF_EQUAL): Compare tokens\n");
+
+        // Default to 'not more restrictive' result.
+        *OutAccessToken = NULL;
+        TRACE("(F=SAFER_TOKEN_NULL_IF_EQUAL) returns TRUE\n");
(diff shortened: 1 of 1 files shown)


dll/win32/advapi32/sec/safer.c
```diff
@@ -27,7 +27,8 @@ SaferCreateLevel(
     _Outptr_ SAFER_LEVEL_HANDLE *pLevelHandle,
     _Reserved_ PVOID pReserved)
 {
-    FIXME("(%lu, %lu, %lu, %p, %p) stub\n", dwScopeId, dwLevelId, OpenFlags, pLevelHandle, pReserved);
+    FIXME("%s(%lu, %lu, %lu, %p, %p) stub\n",
+          __FUNCTION__, dwScopeId, dwLevelId, OpenFlags, pLevelHandle, pReserved);
```
learn-more, line 31: Why the function prefix everywhere?
That does not add anything but noise...

GeoB99:
To be frank, none of that makes much sense (I wonder how would you even assign `0xdeadbeefdeadbeefULL` to the newly returned restricted token by the call). Also the function does a lot more when `SAFER_TOKEN_NULL_IF_EQUAL` is supplied to the flags parameter, that is, it retrieves token information supplied to the call, creates a security descriptor, sets such descriptor to the owner and finally it duplicates the given token. Returning NULL to the output resulting token is not enough in order to achieve a correct "stub-implementation". 

I'll close the PR for being stale and for that reason, when you come back with a proper implementation I'll open it and address with some review. If you're curious how is this function implemented in ADVAPI32, here's a stack list of used routines according to public PDBs:

SaferComputeTokenFromLevel():

- IsSaferDisabled
- RtlEnterCriticalSection
- CodeAuthzLevelObjpLookupByLevelId
- NtQueryInformationToken
- CodeAuthzpComputeAccessTokenFromCodeAuthzObject
- CodeAuthzpCompareCodeAuthzLevelWithToken
- RtlLeaveCriticalSection
- CodeAuthzpGetTokenInformation
- RtlCreateSecurityDescriptor
- RtlSetOwnerSecurityDescriptor
- NtDuplicateToken
- LocalFree
- BaseSetLastNTError
- SetLastError

PR #3086, opened 2020-08-21
[NTOS:EX] Fix uninitialized memory in ExUuidCreate

khanhnt2:
## Purpose

Fix uninitialized memory in ExUuidCreate.

## Analysis

In the function ExUuidCreate, the code is used to initialize value of Uuid is
```c
            RtlCopyMemory(&Uuid->Data4[0],
                          &ExpUuidCachedValues.NodeId[0],
                          SEED_BUFFER_SIZE);
```

```c
typedef struct _GUID
{
#ifdef _MSC_VER
    unsigned long  Data1;
#else
    unsigned int   Data1;
#endif
    unsigned short Data2;
    unsigned short Data3;
    unsigned char  Data4[ 8 ];
} GUID;
```

The size of `Uuid->Data` is 8 whereas the value of `SEED_BUFFER_SIZE` is 6. So `Data4[6]` and `Data[7]` are not initialized. This bug affects videoprt driver of Win32SS at https://github.com/reactos/reactos/blob/18a51dc7f5f66108fc8050e604debaa53ab587cb/win32ss/drivers/videoprt/registry.c#L395-L404

The function `RtlStringFromGUID` converts UUID to string and it uses Data[6] and Data[7] too.


ntoskrnl/ex/uuid.c
```diff
@@ -400,9 +400,10 @@ ExUuidCreate(OUT UUID *Uuid)
         {
             Time.QuadPart = ExpUuidCachedValues.Time;
 
+            C_ASSERT(sizeof(UUID_CACHED_VALUES_STRUCT) - FIELD_OFFSET(UUID_CACHED_VALUES_STRUCT, ClockSeqHiAndReserved) == sizeof(Uuid->Data4));
```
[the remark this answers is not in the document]
khanhnt2: I changed to `RTL_SIZEOF_THROUGH_FIELD(UUID_CACHED_VALUES_STRUCT, NodeId)` and compile progress was still failed. I think copying 2 padding bytes into Uuid isn't good idea so I just only changed `==` to `>=`.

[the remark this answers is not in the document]
SergeGautherie: `==` is actually wanted here.
Which code did you try? Which 3 values do you get?

ntoskrnl/ex/uuid.c
```diff
@@ -400,9 +400,10 @@ ExUuidCreate(OUT UUID *Uuid)
         {
             Time.QuadPart = ExpUuidCachedValues.Time;
 
+            C_ASSERT(sizeof(UUID_CACHED_VALUES_STRUCT) - FIELD_OFFSET(UUID_CACHED_VALUES_STRUCT, ClockSeqHiAndReserved) >= sizeof(Uuid->Data4));
             RtlCopyMemory(&Uuid->Data4[0],
-                          &ExpUuidCachedValues.NodeId[0],
-                          SEED_BUFFER_SIZE);
+                          &ExpUuidCachedValues.ClockSeqHiAndReserved,
+                          sizeof(Uuid->Data4));
```
binarymaster, line 406:
```
In file included from ../src/sdk/include/crt/mingw32/intrin.h:98,
                 from ../src/sdk/include/crt/intrin.h:1017,
                 from sdk/include/ddk/wdm.h:70,
                 from sdk/include/ddk/ntddk.h:38,
                 from sdk/include/ddk/ntifs.h:35,
                 from ../src/ntoskrnl/include/ntoskrnl.h:34,
                 from ../src/ntoskrnl/ex/uuid.c:13:
In function 'memcpy',
    inlined from 'ExUuidCreate' at ../src/ntoskrnl/ex/uuid.c:404:13:
../src/sdk/include/crt/mingw32/intrin_x86.h:76:12: error: 'memmove' offset [14, 20] from the object at 'ExpUuidCachedValues' is out of the bounds of referenced subobject 'ClockSeqHiAndReserved' with type 'unsigned char' at offset 12 [-Werror=array-bounds]
     return memmove(dest, source, num);
            ^~~~~~~~~~~~~~~~~~~~~~~~~~
```

SergeGautherie: Add a `union`, adding `Data4`?

ntoskrnl/ex/uuid.c
```diff
@@ -400,9 +400,10 @@ ExUuidCreate(OUT UUID *Uuid)
         {
             Time.QuadPart = ExpUuidCachedValues.Time;
 
+            C_ASSERT(sizeof(UUID_CACHED_VALUES_STRUCT) - FIELD_OFFSET(UUID_CACHED_VALUES_STRUCT, ClockSeqHiAndReserved) >= sizeof(Uuid->Data4));
```
Extravert-ir, line 403: I don't like the assertion placement. Can we move it to the GLOBALS section on the top and put some comment along with it?

Extravert-ir (changes requested):
Please, fix build issues

ntoskrnl/ex/uuid.c
khanhnt2: I have moved assertion to global.

khanhnt2:
Btw, I don't know why my pull request is failed by GCC, my added assertion is used to check out of bound but compiler still warning out of bound issue. Is that compiler bug?

ThFabba:
> Btw, I don't know why my pull request is failed by GCC, my added assertion is used to check out of bound but compiler still warning out of bound issue. Is that compiler bug?

The assertion verifies this for humans, but it's not going to help the compiler. I like the union idea suggested by Serge. Something like:
```
typedef struct _UUID_CACHED_VALUES_STRUCT
{
    ULONGLONG Time;
    LONG AllocatedCount;
    union
    {
        struct
        {
            UCHAR ClockSeqHiAndReserved;
            UCHAR ClockSeqLow;
            UCHAR NodeId[SEED_BUFFER_SIZE];
        };
        UCHAR GuidInit[8];
    };
} UUID_CACHED_VALUES_STRUCT, *PUUID_CACHED_VALUES_STRUCT;
```

ntoskrnl/ex/uuid.c
```diff
@@ -34,6 +34,7 @@
 #pragma alloc_text(INIT, ExLuidInitialization)
 #endif
 
+C_ASSERT(sizeof(UUID_CACHED_VALUES_STRUCT) - FIELD_OFFSET(UUID_CACHED_VALUES_STRUCT, ClockSeqHiAndReserved) >= sizeof(((GUID *)0)->Data4));
```
ThFabba, line 37: `RTL_FIELD_SIZE(GUID, Data4)` would make the second part more readable.
I also personally liked this better inside the function close to where we make use of this (in)equality fact.

HBelusca:
@khanhnt2 : Any news on this?

tkreuzer:
Superseded by https://github.com/reactos/reactos/pull/3193

PR #3094, opened 2020-08-25
[NTOS:PNP] Fix and improve PnP notifications code

Extravert-ir:
- BUGFIX: do not call IoGetRelatedTargetDevice while guarded mutex is acquired
  (the function issues an APC, but they are disabled inside a critical section)
- BUGFIX: only the beginning of a structure for GUID_PNP_CUSTOM_NOTIFICATION was copied and queued.
  Just pass it as-is to a subscriber, without copying
- Don't convert event GUID to string, store and compare GUID struct itself
- Split IopNotifyPlugPlayNotification into 3 functions for each type of notification
  (less stack usage and for future changes)
- Move initialization code for notifications into a separate routine
- Use separate lists and locks for every type of notification
- Put "TargetDeviceChange" notifications into their place inside DEVICE_NODE

### DIFF
--- a/ntoskrnl/io/pnpmgr/pnpnotify.c
+++ b/ntoskrnl/io/pnpmgr/pnpnotify.c
@@ -1,11 +1,11 @@
 /*
- * PROJECT:         ReactOS Kernel
- * COPYRIGHT:       GPL - See COPYING in the top level directory
- * FILE:            ntoskrnl/io/pnpmgr/pnpnotify.c
- * PURPOSE:         Plug & Play notification functions
- * PROGRAMMERS:     Filip Navara (xnavara@volny.cz)
- *                  Hervé Poussineau (hpoussin@reactos.org)
- *                  Pierre Schweitzer
+ * PROJECT:     ReactOS Kernel
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Plug & Play notification functions
+ * COPYRIGHT:   Copyright 2003 Filip Navara <xnavara@volny.cz>
+ *              Copyright 2005-2006 HervÃ© Poussineau <hpoussin@reactos.org>
+ *              Copyright 2010 Pierre Schweitzer <pierre@reactos.org>
+ *              Copyright 2020 Victor Perevertkin <victor.perevertkin@reactos.org>
  */
 
 /* INCLUDES ******************************************************************/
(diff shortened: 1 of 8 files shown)


ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -341,20 +357,28 @@ IoRegisterPlugPlayNotification(IN IO_NOTIFICATION_EVENT_CATEGORY EventCategory,
         case EventCategoryTargetDeviceChange:
         {
             Entry->FileObject = (PFILE_OBJECT)EventCategoryData;
+            Status = IoGetRelatedTargetDevice((PFILE_OBJECT)EventCategoryData, &Entry->FileObjectPDO);
```
Extravert-ir, line 360: Can this return different TargetDevices for the same FileObject?

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -279,58 +326,52 @@ IoRegisterPlugPlayNotification(IN IO_NOTIFICATION_EVENT_CATEGORY EventCategory,
...
         case EventCategoryDeviceInterfaceChange:
         {
-            Status = RtlStringFromGUID(EventCategoryData, &Entry->Guid);
-            if (!NT_SUCCESS(Status))
+            if (EventCategoryFlags & PNPNOTIFY_DEVICE_INTERFACE_INCLUDE_EXISTING_INTERFACES)
             {
-                ExFreePoolWithTag(Entry, TAG_PNP_NOTIFY);
-                ObDereferenceObject(DriverObject);
-                return Status;
+                DEVICE_INTERFACE_CHANGE_NOTIFICATION NotificationInfos;
```
tkreuzer, line 340: This case is pretty large, consider giving it it's own function.

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -19,210 +19,259 @@
...
 
-    /* Loop through procedures registred in PnpNotifyListHead
-     * list to find those that meet some criteria.
-     */
-    ListEntry = PnpNotifyListHead.Flink;
-    while (ListEntry != &PnpNotifyListHead)
+    /* Call entry into new allocated memory */
+    DPRINT("IopNotifyPlugPlayNotification(): found suitable callback %p\n", Entry);
+
+    KeReleaseGuardedMutex(&PiNotificationInProgressLock);
```
tkreuzer, line 112: Why is this lock released here? Or better: Why is this function called with this lock held? And why does the name imply that it's held while the notification is in progress, rather than protecting the list, which afaics is the only thing it does.

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -341,20 +382,30 @@ IoRegisterPlugPlayNotification(IN IO_NOTIFICATION_EVENT_CATEGORY EventCategory,
...
+            ObDereferenceObject(DriverObject);
+            return STATUS_NOT_SUPPORTED;
         }
     }
 
-    KeAcquireGuardedMutex(&PnpNotifyListLock);
-    InsertHeadList(&PnpNotifyListHead,
-                   &Entry->PnpNotifyList);
-    KeReleaseGuardedMutex(&PnpNotifyListLock);
+    ExInterlockedInsertTailList(&PiAddedNotifyListHead,
```
tkreuzer, line 406: I don't really understand the point of having 2 lists. Why do you need to add it to a different list, when you don't hold any locks while processing an entry from the main list? If you want to add a new entry, just add it the tail and it will be processed at the end of the loop that walks the list, instead of at the head as it was done previously.

tkreuzer (changes requested):
I am not convinced that we need 2 lists here. Instead using a single list, a single *list*-lock and inserting new entries at the tail instead of the head should be sufficient.

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -19,210 +19,259 @@
...
+    }
 
-        if (CallCurrentEntry)
+    *notifStruct = (DEVICE_INTERFACE_CHANGE_NOTIFICATION) {
+        .Version = 1,
+        .Size = sizeof(DEVICE_INTERFACE_CHANGE_NOTIFICATION),
+        .Event = *Event,
+        .InterfaceClassGuid = *InterfaceClassGuid,
+        .SymbolicLinkName = SymbolicLinkName
+    };
```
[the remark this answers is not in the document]
andrew-boyarshin: That's designated initializers from C99. Pretty standard, and supported e.g. with MSVC 2013+.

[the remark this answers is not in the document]
HBelusca: Please don't use that in the kerne and keep the current code stylel; it should remain C89-compatible.

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -19,210 +19,259 @@
...
 
+INIT_FUNCTION
 VOID
-IopNotifyPlugPlayNotification(
-    IN PDEVICE_OBJECT DeviceObject,
-    IN IO_NOTIFICATION_EVENT_CATEGORY EventCategory,
-    IN LPCGUID Event,
-    IN PVOID EventCategoryData1,
-    IN PVOID EventCategoryData2)
+PipInitializeNotifications()
```
HBelusca, line 47:
```suggestion
PipInitializeNotifications(VOID)
```

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -19,210 +19,259 @@
...
+VOID
+PipProcessSingleNotification(
+    _In_ PPNP_NOTIFY_ENTRY Entry,
+    _In_ IO_NOTIFICATION_EVENT_CATEGORY EventCategory,
+    _In_ PVOID NotificationStructure,
+    _In_opt_ PDEVICE_OBJECT DeviceObject)
+{
+    PAGED_CODE();
+
+    // remove deleted entries
```
HBelusca, line 65:
```suggestion
    /* Remove deleted entries */
```

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -19,210 +19,259 @@
...
+    {
+        RemoveEntryList(&Entry->PnpNotifyList);
+        ObDereferenceObject(Entry->DriverObject);
+        ExFreePoolWithTag(Entry, TAG_PNP_NOTIFY);
+        return;
+    }
 
-    KeAcquireGuardedMutex(&PnpNotifyListLock);
-    if (IsListEmpty(&PnpNotifyListHead))
+    // skip ones not corresponding to our category
```
HBelusca, line 74:
```suggestion
    /* Skip ones not corresponding to our category */
```

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -19,210 +19,259 @@
...
-    /* Loop through procedures registred in PnpNotifyListHead
-     * list to find those that meet some criteria.
-     */
-    ListEntry = PnpNotifyListHead.Flink;
-    while (ListEntry != &PnpNotifyListHead)
+    /* Call entry into new allocated memory */
+    DPRINT("IopNotifyPlugPlayNotification(): found suitable callback %p\n", Entry);
+
+    KeReleaseGuardedMutex(&PiNotificationInProgressLock);
+    (Entry->PnpNotificationProc)(NotificationStructure, Entry->Context);
```
HBelusca, line 113:
```suggestion
    Entry->PnpNotificationProc(NotificationStructure, Entry->Context);
```

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -19,210 +19,259 @@
...
+    _In_opt_ PDEVICE_OBJECT DeviceObject)
+{
+    PLIST_ENTRY ListEntry;
+    PPNP_NOTIFY_ENTRY ChangeEntry;
+
+    PAGED_CODE();
+
+    KeAcquireGuardedMutex(&PiNotificationInProgressLock);
+
+    // first, process the main list
```
HBelusca, line 131:
```suggestion
    /* First, process the main list */
```

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -19,210 +19,259 @@
...
-            default:
-            {
-                DPRINT1("IopNotifyPlugPlayNotification(): unknown EventCategory 0x%x UNIMPLEMENTED\n", EventCategory);
-                break;
-            }
-        }
+        PipProcessSingleNotification(ChangeEntry, EventCategory, NotificationStruct, DeviceObject);
+    }
+
+    // Now process added elements (if any)
```
HBelusca, line 142:
```suggestion
    /* Now process added elements (if any) */
```

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -19,210 +19,259 @@
...
+    ListEntry = PiAddedNotifyListHead.Flink;
+    while ((ListEntry = ExInterlockedRemoveHeadList(&PiAddedNotifyListHead, &PiNotifyAdditionLock)))
+    {
+        // append the entry to PiNotifyListHead while we're holding PiNotificationInProgressLock
```
HBelusca, line 146:
```suggestion
        /* Append the entry to PiNotifyListHead while we're holding PiNotificationInProgressLock */
```
OK the code is already self-explaining, so no need to say that?

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -19,210 +19,259 @@
...
+ *             EventCategoryHardwareProfileChange PnP event
+ *
+ * @param[in]  Event  The PnP event GUID
+ */
+VOID
+PiNotifyHardwareProfileChange(
+    _In_ LPCGUID Event)
+{
+    PHWPROFILE_CHANGE_NOTIFICATION notifStruct;
+    notifStruct = ExAllocatePoolWithTag(PagedPool, sizeof(*notifStruct), TAG_PNP_NOTIFY);
```
HBelusca, line 205:
```suggestion

    notifStruct = ExAllocatePoolWithTag(PagedPool, sizeof(*notifStruct), TAG_PNP_NOTIFY);
```

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -19,210 +19,259 @@
...
+ * @param[in]  CustomNotification  Pointer to a custom notification for GUID_PNP_CUSTOM_NOTIFICATION
+ */
+VOID
+PiNotifyTargetDeviceChange(
+    _In_ LPCGUID Event,
+    _In_ PDEVICE_OBJECT DeviceObject,
+    _In_opt_ PTARGET_DEVICE_CUSTOM_NOTIFICATION CustomNotification)
+{
+    PVOID notificationStruct;
+    // just in case our device is removed during the operation
```
HBelusca, line 237:
```suggestion

    /* Reference the device just in case it is removed during the operation */
```

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -19,210 +19,259 @@
...
-                ChangeEntry);
-
-            KeReleaseGuardedMutex(&PnpNotifyListLock);
-            (ChangeEntry->PnpNotificationProc)(NotificationStructure,
-                                               ChangeEntry->Context);
-            KeAcquireGuardedMutex(&PnpNotifyListLock);
+            return;
         }
 
+        *notifStruct = (TARGET_DEVICE_REMOVAL_NOTIFICATION) {
```
HBelusca, line 252: See @tkreuzer comment above.

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -279,58 +326,52 @@ IoRegisterPlugPlayNotification(IN IO_NOTIFICATION_EVENT_CATEGORY EventCategory,
...
+                                  sizeof(GUID));
+                    NotificationInfos.SymbolicLinkName = &SymbolicLinkU;
+
+                    for (SymbolicLink = SymbolicLinkList;
+                         *SymbolicLink;
+                         SymbolicLink += (SymbolicLinkU.Length / sizeof(WCHAR)) + 1)
+                    {
+                        RtlInitUnicodeString(&SymbolicLinkU, SymbolicLink);
+                        DPRINT("Calling callback routine for %S\n", SymbolicLink);
+                        (*CallbackRoutine)(&NotificationInfos, Context);
```
HBelusca, line 367:
```suggestion
                        CallbackRoutine(&NotificationInfos, Context);
```

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -279,58 +326,52 @@ IoRegisterPlugPlayNotification(IN IO_NOTIFICATION_EVENT_CATEGORY EventCategory,
...
+                    }
+
+                    ExFreePool(SymbolicLinkList);
```
HBelusca, line 370: Can we use a ExFreePoolWithTag here? Or is the tag actually not known?

ntoskrnl/io/pnpmgr/pnpnotify.c
Extravert-ir: The lock is released because we need to call the routine with APCs enabled. It's not stated explicitly in the docs but it's "what windows does" and in this case such behaviour should be kept, I think.
I'll add a comment if we keep the lock in the end

ntoskrnl/io/pnpmgr/pnpnotify.c
Extravert-ir: The idea was to make adding and removing operations lock-free.

> just add it the tail and it will be processed at the end of the loop that walks the list

How can it work if the new element is added while we process the last element? How to be with deletions?

ntoskrnl/io/pnpmgr/pnpnotify.c
Extravert-ir: Looks like I've found a problem here. The element can be deleted while we're inside `PnpNotificationProc` (by another thread) because the lock is not held.

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -19,210 +19,274 @@
...
-    IO_NOTIFICATION_EVENT_CATEGORY EventCategory;
     PVOID Context;
-    UNICODE_STRING Guid;
-    PFILE_OBJECT FileObject;
     PDRIVER_OBJECT DriverObject;
     PDRIVER_NOTIFICATION_CALLBACK_ROUTINE PnpNotificationProc;
+    union
+    {
+        GUID Guid; // for EventCategoryDeviceInterfaceChange
+        struct // for EventCategoryTargetDeviceChange
```
tkreuzer, line 28: Is there any reason this is a struct?

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -19,210 +19,274 @@
...
+CODE_SEG("PAGE")
+VOID
+PiFreeTargetDeviceNotifyList(
+    _In_ PDEVICE_NODE DeviceNode)
+{
+    PAGED_CODE();
 
-    KeAcquireGuardedMutex(&PnpNotifyListLock);
-    if (IsListEmpty(&PnpNotifyListHead))
+    for (PLIST_ENTRY entry = DeviceNode->TargetDeviceNotify.Flink;
```
tkreuzer, line 64: Declaration should be on top the function / scope.

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -19,210 +19,274 @@
...
+    KeInitializeGuardedMutex(&IopTargetDeviceNotifyLock);
+    KeInitializeGuardedMutex(&IopDeviceClassNotifyLock);
+    KeInitializeGuardedMutex(&IopHwProfileNotifyLock);
+    InitializeListHead(&PiNotifyHwProfileListHead);
+    InitializeListHead(&PiNotifyDeviceInterfaceListHead);
+}
 
-    ASSERT(DeviceObject);
+CODE_SEG("PAGE")
+VOID
```
tkreuzer, line 58: I suggest adding a ```_Requires_lock_held_(...)``` annotation.

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -19,210 +19,274 @@
...
+    _In_ IO_NOTIFICATION_EVENT_CATEGORY EventCategory,
+    _In_ PVOID NotificationStructure,
+    _In_ PKGUARDED_MUTEX Lock)
+{
+    PDEVICE_INTERFACE_CHANGE_NOTIFICATION notifyStruct = NotificationStructure;
+
+    PAGED_CODE();
+
+    // actually remove "deleted" entries
+    if (InterlockedAnd(&Entry->Deleted, 1))
```
tkreuzer, line 91: What's the point of ```InterlockedAnd``` here?

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -19,210 +19,274 @@
...
-     */
-    ListEntry = PnpNotifyListHead.Flink;
-    while (ListEntry != &PnpNotifyListHead)
+    // release the lock because the notification routine has to be called without any
+    // limitations regarding APCs
+    KeReleaseGuardedMutex(Lock);
+
+    // for verifying the driver behaviour
+#if DBG
+    KIRQL oldIrql = KeGetCurrentIrql();
```
tkreuzer, line 113: Declaration should be on top of scope.

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -19,210 +19,274 @@
...
-        CallCurrentEntry = FALSE;
+        return;
+    }
 
-        if (ChangeEntry->EventCategory != EventCategory)
-        {
-            ListEntry = ListEntry->Flink;
-            continue;
-        }
+    *notifyStruct = (DEVICE_INTERFACE_CHANGE_NOTIFICATION) {
```
tkreuzer, line 148: Not standard-conforming.

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -19,210 +19,274 @@
...
+    _In_ LPCGUID Event)
+{
+    PHWPROFILE_CHANGE_NOTIFICATION notifyStruct;
+    notifyStruct = ExAllocatePoolWithTag(PagedPool, sizeof(*notifyStruct), TAG_PNP_NOTIFY);
+    if (!notifyStruct)
+    {
+        return;
+    }
+
+    *notifyStruct = (HWPROFILE_CHANGE_NOTIFICATION) {
```
tkreuzer, line 195: non-standard

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -19,210 +19,274 @@
...
 } PNP_NOTIFY_ENTRY, *PPNP_NOTIFY_ENTRY;
 
-KGUARDED_MUTEX PnpNotifyListLock;
-LIST_ENTRY PnpNotifyListHead;
+KGUARDED_MUTEX IopTargetDeviceNotifyLock;
+KGUARDED_MUTEX IopDeviceClassNotifyLock;
+KGUARDED_MUTEX IopHwProfileNotifyLock;
+
+LIST_ENTRY PiNotifyHwProfileListHead;
+LIST_ENTRY PiNotifyDeviceInterfaceListHead;
```
ThFabba, line 42: You can use `_Guarded_by_` annotations here to make the relationship to the locks obvious.

Extravert-ir:
@tkreuzer regarding non-standard stuff: this is not logical. We need to either enforce `-std=c89`, or use all features our compilers support. I'll make this code compile with MSVC2010/C89 if we decide to keep that compatibility in the mailing list voting

HBelusca:
> @tkreuzer regarding non-standard stuff: this is not logical. We need to either enforce `-std=c89`, or use all features our compilers support. I'll make this code compile with MSVC2010/C89 if we decide to keep that compatibility in the mailing list voting

It looks to me like it's about keeping C89 compatibility for the kernel.

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -19,210 +19,274 @@
...
+    }
+
+    KeAcquireGuardedMutex(&IopTargetDeviceNotifyLock);
+
+    for (PLIST_ENTRY entry = deviceNode->TargetDeviceNotify.Flink;
+         entry != &deviceNode->TargetDeviceNotify;
+         entry = entry->Flink)
+    {
+        PPNP_NOTIFY_ENTRY nEntry = CONTAINING_RECORD(entry, PNP_NOTIFY_ENTRY, PnpNotifyList);
+        ((PTARGET_DEVICE_REMOVAL_NOTIFICATION)notificationStruct)->FileObject = nEntry->FileObject;
```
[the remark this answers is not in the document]
HBelusca: Should it be done here in the code, or more closely around the definition of the involved structures?

ntoskrnl/io/pnpmgr/pnpnotify.c
Extravert-ir: Does it make sense to use `ExFreePoolWithTag(ptr, 0)` instead of `ExFreePool(ptr)`?

ntoskrnl/io/pnpmgr/pnpnotify.c
Extravert-ir: I used it as an interlocked load operation

ntoskrnl/io/pnpmgr/pnpnotify.c
ColinFinck: @tkreuzer: Frankly asking, what are the benefits?
As far as I know, C89 requiring this is no "feature", but a limitation of the compilers of that time.
However, our currently used compilers are no longer bounded by the artificial limits of C89.

I only see downsides here: More lines of code and cluttered scope definitions. A perfect example is: https://github.com/reactos/reactos/blob/30144179190c11519add8c1f4fe13afed52cb997/ntoskrnl/io/iomgr/file.c#L324-L361

JoachimHenze (changes requested):
I do request changes until all the VS2010 issues were addressed.that were pointed to by HBelusca and Timo.

ntoskrnl/io/pnpmgr/pnpnotify.c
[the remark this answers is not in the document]
vgalnt: Maybe this will immediately solve all the problems? :)

/* Windows NT target version 10.19H1 */
#define VER_PRODUCTMAJORVERSION  10
#define VER_PRODUCTMINORVERSION   7

add_definitions(-DWINVER=0xA00
                        -D_WIN32_IE=0xA00
                        -D_WIN32_WINNT=0xA00
                        -D_WIN32_WINDOWS=0xA00
                        -D_SETUPAPI_VER=0xA00
                        -DMINGW_HAS_SECURE_API=1)

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -19,210 +19,274 @@
...
+ *
+ * @param[in]  Event  The PnP event GUID
+ */
+CODE_SEG("PAGE")
+VOID
+PiNotifyHardwareProfileChange(
+    _In_ LPCGUID Event)
+{
+    PHWPROFILE_CHANGE_NOTIFICATION notifyStruct;
+    notifyStruct = ExAllocatePoolWithTag(PagedPool, sizeof(*notifyStruct), TAG_PNP_NOTIFY);
```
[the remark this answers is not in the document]
Extravert-ir: This doesn't make sense for a current implementation, but for a future one (which may use a separate working thread) this will be handy. I'm not sure if I should leave it like this now through

ntoskrnl/io/pnpmgr/pnpnotify.c
Extravert-ir: We shouldn't need a lock here

ntoskrnl/io/pnpmgr/pnpnotify.c
Extravert-ir: Why? For saving stack space?

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -14,215 +14,309 @@
...
-     */
-    ListEntry = PnpNotifyListHead.Flink;
-    while (ListEntry != &PnpNotifyListHead)
+    // the notification may be unregistered inside the procedure
+    // thus reference the driver so we may proceed
+    ObReferenceObject(Entry->DriverObject);
+
+    // release the lock because the notification routine has to be called without any
+    // limitations regarding APCs
+    KeReleaseGuardedMutex(Lock);
```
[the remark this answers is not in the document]
ThFabba: On that note, we should also do what Win7+ does with `IoUnregisterPlugPlayNotificationEx` and guarantee that no callbacks are running (or will run) after the unregister call returns.
This is safe to do for the non-Ex version as well (since it doesn't specify this behavior), and is simply the better architecture.

ntoskrnl/io/pnpmgr/pnpnotify.c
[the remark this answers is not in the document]
ThFabba: Sure, I think it's fine to leave it this way.

ntoskrnl/io/pnpmgr/pnpnotify.c
ThFabba: More importantly, I don't think this provides a readability benefit over assigning the members individually, and in fact makes the code two lines longer.

Extravert-ir: It also zeroes all non-assigned members, so you're saving `RtlZeroMemory` call here

ntoskrnl/io/pnpmgr/pnpnotify.c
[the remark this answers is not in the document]
Extravert-ir: Here we go, now there is `PiReferencePnpNotifyEntry`. Now after the register call returns we may have two situations: either it frees the notification structure (and no callbacks can be run after that obviously) or we're in the middle of a callback and cleanup will occur upon its return. I guess, this matches `IoUnregisterPlugPlayNotificationEx` behavior

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -366,25 +472,46 @@ IoRegisterPlugPlayNotification(IN IO_NOTIFICATION_EVENT_CATEGORY EventCategory,
...
+        default:
+            UNREACHABLE;
+            return STATUS_NOT_SUPPORTED;
+    }
 
-    ExFreePoolWithTag(Entry, TAG_PNP_NOTIFY);
+    KeAcquireGuardedMutex(Lock);
+    if (!Entry->Deleted)
+    {
+        Entry->Deleted = TRUE; // so it can't be unregistered two times
```
ThFabba, line 507: Does Windows protect against this, too? Otherwise I wouldn't add extra code just to support incorrect behavior.

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -14,215 +14,308 @@
...
+
+    KeAcquireGuardedMutex(&PiNotifyDeviceInterfaceLock);
+
+    PLIST_ENTRY entry = PiNotifyDeviceInterfaceListHead.Flink;
+    while (entry != &PiNotifyDeviceInterfaceListHead)
     {
-        ChangeEntry = CONTAINING_RECORD(ListEntry, PNP_NOTIFY_ENTRY, PnpNotifyList);
-        CallCurrentEntry = FALSE;
+        PPNP_NOTIFY_ENTRY nEntry = CONTAINING_RECORD(entry, PNP_NOTIFY_ENTRY, PnpNotifyList);
+        entry = entry->Flink; // the list entry may be freed
```
ThFabba, line 187: So, now this just became unsafe. We're not holding a reference to the "next" entry, but `PiProcessSingleNotification` still releases the lock so it could become invalid.
We could:
* take a reference to the next entry here (but then would need extra processing to handle the case where that ends up with a zero refcount)
* have `PiProcessSingleNotification` save the next entry before it derefs the current one and return that
* take the reference to the current entry here rather than in `PiProcessSingleNotification`, so that we can grab the next item before the deref

I'm leaning towards the second option.

ntoskrnl/io/pnpmgr/pnpnotify.c
```diff
@@ -14,215 +14,308 @@
...
+    KeInitializeGuardedMutex(&PiNotifyDeviceInterfaceLock);
+    InitializeListHead(&PiNotifyHwProfileListHead);
+    InitializeListHead(&PiNotifyDeviceInterfaceListHead);
+}
 
-    ASSERT(DeviceObject);
+static
+CODE_SEG("PAGE")
+VOID
+PiDereferencePnpNotifyEntry(
```
ThFabba, line 67: I know we can't annotate/assert a specific lock required here, since multiple ones are possible. But since they're all guarded mutexes, maybe `_Requires_lock_held_(_Global_critical_region_)` + `ASSERT(KeAreAllApcsDisabled())`?
Same for the Reference function of course.

ntoskrnl/io/pnpmgr/pnpnotify.c
Extravert-ir: Have I understood you right, it may become invalid if the next element is freed while the lock is held?
And in such case we want to keep the "current" entry to get the updated `Flink`

ntoskrnl/io/pnpmgr/pnpnotify.c
Extravert-ir: Yeah, it has the protection

PR #3096, opened 2020-08-25
[OpenGL32] Remove Err Log Spam from wglMakeCurrent in nominal cases CORE-17224

KyleKatarn31:
## Purpose

- wglCurrentContext raises the following error :
err:(dll/opengl/opengl32/wgl.c:747) Error: hdc is not a DC handle!
In particular during ROS installation, after declining Wine Gecko installation.
- Looking at the code and inspecting wgl.c from Wine HEAD : https://source.winehq.org/git/wine.git/blob_plain/HEAD:/dlls/opengl32/wgl.c
it appears that this "Err" is not an error, since we will naturally fall into this branch in particular when wglMakeCurrent(NULL, NULL) is called (from DLL_THREAD_DETACH or DLL_PROCESS_DETACH,....)
- This Err is not present in Wine's head (where the wglMakeCurrent is also much lighter)

JIRA: https://jira.reactos.org/browse/CORE-17224

## Proposed changes

- Remove unecssary ERR macro call.

### DIFF
--- a/dll/opengl/opengl32/wgl.c
+++ b/dll/opengl/opengl32/wgl.c
@@ -744,6 +744,7 @@ BOOL WINAPI wglMakeCurrent(HDC hdc, HGLRC hglrc)
         /* Winetest conformance */
         if (GetObjectType( hdc ) != OBJ_DC && GetObjectType( hdc ) != OBJ_MEMDC)
         {
+            if (hdc) ERR( "Error: hdc is not a DC handle!\n");
             SetLastError( ERROR_INVALID_HANDLE );
             return FALSE;
         }
--- a/dll/opengl/opengl32/wgl.c
+++ b/dll/opengl/opengl32/wgl.c
@@ -744,7 +744,10 @@ BOOL WINAPI wglMakeCurrent(HDC hdc, HGLRC hglrc)
         /* Winetest conformance */
         if (GetObjectType( hdc ) != OBJ_DC && GetObjectType( hdc ) != OBJ_MEMDC)
         {
-            if (hdc) ERR( "Error: hdc is not a DC handle!\n");
+            if (hdc)
+            {
+                ERR("hdc (%p) is not a DC handle!\n", hdc);
+            }
             SetLastError( ERROR_INVALID_HANDLE );
             return FALSE;
         }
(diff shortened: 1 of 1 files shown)


SergeGautherie:
> naturally fall into this branch

Does that apply to `NULL` only or actual values too?

KyleKatarn31:
> > naturally fall into this branch
> 
> Does that apply to `NULL` only or actual values too?

It's a test conformance case (extreme robustness case) but I assume yes. You're indeed right, this deserves a modification.

dll/opengl/opengl32/wgl.c
```diff
@@ -744,6 +744,7 @@ BOOL WINAPI wglMakeCurrent(HDC hdc, HGLRC hglrc)
         /* Winetest conformance */
         if (GetObjectType( hdc ) != OBJ_DC && GetObjectType( hdc ) != OBJ_MEMDC)
         {
+            if (hdc) ERR( "Error: hdc is not a DC handle!\n");
```
SergeGautherie, line 747:
```suggestion
            if (hdc)
            {
                ERR("hdc (%p) is not a DC handle!\n", hdc);
            }
```

dll/opengl/opengl32/wgl.c
```diff
@@ -744,7 +744,10 @@ BOOL WINAPI wglMakeCurrent(HDC hdc, HGLRC hglrc)
         /* Winetest conformance */
         if (GetObjectType( hdc ) != OBJ_DC && GetObjectType( hdc ) != OBJ_MEMDC)
```
SergeGautherie, line 745: While here, merge redundant `GetObjectType` calls?

dll/opengl/opengl32/wgl.c
```diff
@@ -744,7 +744,10 @@ BOOL WINAPI wglMakeCurrent(HDC hdc, HGLRC hglrc)
...
         {
-            ERR( "Error: hdc is not a DC handle!\n");
+            if (hdc)
+            {
+                ERR("hdc (%p) is not a DC handle!\n", hdc);
```
SergeGautherie, line 749: NB: If this remaining case is actually triggered, then we may want to log `GetObjectType(hdc)` result too.

KyleKatarn31: yes, I add this to the ERR macro.

dll/opengl/opengl32/wgl.c
KyleKatarn31: yes, i make this change.

KyleKatarn31:
All comments taken into account. Ready to go ?

learn-more (approved):

PR #3098, opened 2020-08-25
[RAPPS] improve when to call  ConInitStdStreams

kernelbin:
## Purpose

trivial improvement about when to call ConInitStdStreams

### DIFF
--- a/base/applications/rapps/unattended.cpp
+++ b/base/applications/rapps/unattended.cpp
@@ -274,6 +274,8 @@ BOOL ParseCmdAndExecute(LPWSTR lpCmdLine, BOOL bIsFirstLaunch, int nCmdShow)
 
         if (hMutex)
             CloseHandle(hMutex);
+
+        return TRUE;
     }
     else if (MatchCmdOption(argv[1], CMD_KEY_INSTALL))
     {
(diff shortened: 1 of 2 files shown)


HBelusca:
Hmmm....
Wouldn't it be possible to have some "cmd_main" or "console_main" (whatever...) function that is called when RAPPS in text(command)-mode is triggered, that would do this ConInit thing once ?

kernelbin:
> Hmmm....
> Wouldn't it be possible to have some "cmd_main" or "console_main" (whatever...) function that is called when RAPPS in text(command)-mode is triggered, that would do this ConInit thing once ?

Not until we parsed the cmdline did we know it's going to start in CLI or not🤣
so having a ```console_main``` or whatever means I need to parse cmdline twice, or passing a lot of information retrieved from cmdline to ```console_main``` ...
It's possible but IMO it's possibly worse than current desing🤣

base/applications/rapps/unattended.cpp
```diff
@@ -283,18 +283,22 @@ BOOL ParseCmdAndExecute(LPWSTR lpCmdLine, BOOL bIsFirstLaunch, int nCmdShow)
     }
     else if (MatchCmdOption(argv[1], CMD_KEY_FIND))
```
learn-more, line 284:
```suggestion
    ConInitStdStreams();
    if (MatchCmdOption(argv[1], CMD_KEY_FIND))
```
You'd also need to add a return in the first if() case

learn-more: then you can remove the init from all

PR #3099, opened 2020-08-26
[SHELLEXT] Trivial cleanup by removing unused/unnecessary command lines from localization files (.rc) 

KyleKatarn31:
## Purpose

- ShellExt .rc file contain command line elements which are neither a good design solution, nor language-dependant nor even used in the code at all.

## Proposed changes

- removal of these unused elements


learn-more:
This is done for compatibility with 2k3, which also has this.

KyleKatarn31:
What kind of compatibility if strings are not used ?

PR #3100, opened 2020-08-26
[RAMDISK] Add setup information file for the ramdisk driver

archeYR:
Before:
qemu-system-i386_h2HKvuLfpn

After:
qemu-system-i386_syiPvVUnEy


drivers/storage/class/ramdisk/ramdisk.inf
```diff
@@ -0,0 +1,69 @@
...
+[RamdiskVolume_Inst.NT.Services]
+AddService = , 2
+
+;-------------------------------- STRINGS -------------------------------
+
+[Strings]
+ReactOS = "ReactOS Team"
+RamdiskClassName = "Ramdisk"
+RamdiskBus.Desc = "ReactOS RAM Disk Controller"
+RamdiskVolume.Desc = "ReactOS RAM Disk Device (volume)"
```
binarymaster, line 61: The controller and volume devices are not ReactOS-specific, so better name them like that:
```suggestion
RamdiskBus.Desc = "RAM disk controller"
RamdiskVolume.Desc = "RAM disk volume"
```
(lowercase names for consistency with other standard device names)

binarymaster (commented):
Also I think maybe disk volume should have different class, but not sure.

archeYR:
> 
> 
> Also I think maybe disk volume should have different class, but not sure.

In Windows (XP/2003 at least), they share the same class.

drivers/storage/class/ramdisk/ramdisk.inf
HBelusca: On Windows:
image
(Translation:
"Pilote" == "Driver"
"Contrôleur de disque virtuel Windows" == "Windows Virtual Disk Controller"
"Périphérique de disque virtuel Windows (volume)" == "Windows Virtual Disk Peripheral (volume)"
)

binarymaster: Ok then a second suggestion :smile: Feel free to choose between my first one and this one
```suggestion
RamdiskBus.Desc = "RAM disk controller"
RamdiskVolume.Desc = "RAM disk device (volume)"
```

drivers/storage/class/ramdisk/ramdisk.inf
```diff
@@ -0,0 +1,69 @@
...
+Ramdisk.ServiceDesc = "ReactOS RAM Disk Driver"
+
+[Strings.0415]
```
KyleKatarn31, line 64: I see that .0415 is directly added in this PR. May I suggest for French ?
ReactOS = "Equipe ReactOS"
RamdiskClassName = "Ramdisk"
RamdiskBus.Desc = "Contrôleur de disque RAM"
RamdiskVolume.Desc = "Disque RAM (volume)"

drivers/storage/class/ramdisk/ramdisk.inf
binarymaster: image

drivers/storage/class/ramdisk/ramdisk.inf
```diff
@@ -0,0 +1,69 @@
...
+ReactOS = "Zespół ReactOS"
+RamdiskClassName = "Ramdisk"
```
KyleKatarn31, line 66: is the classname really language dependant ?

drivers/storage/class/ramdisk/ramdisk.inf
KyleKatarn31: Ramdisk.ServiceDesc = "Pilote de disque RAM"

binarymaster: Additional translations are added later, because things in the PR tend to change.

drivers/storage/class/ramdisk/ramdisk.inf
archeYR: For some languages (eg. those that use different character sets) probably yeah. Even though it's not really needed for that language, personally I would like to keep the same amount of strings for each language.

binarymaster (approved):

PR #3101, opened 2020-08-26
[RAPPS] settings read/write refactor

kernelbin:
## Purpose

refactor code about settings
now settings are no longer stored in registery in binary form.
but a more human-friendly way.

A0QU5)G%5@7XLJ~J0WECV9K

### DIFF
--- a/base/applications/rapps/include/settings.h
+++ b/base/applications/rapps/include/settings.h
@@ -1,0 +1,33 @@
+#pragma once
+#include <windef.h>
+#include <wininet.h>
+
+struct SETTINGS_INFO
+{
+    BOOL bSaveWndPos;
+    BOOL bUpdateAtStart;
+    BOOL bLogEnabled;
+    WCHAR szDownloadDir[MAX_PATH];
+    BOOL bDelInstaller;
+    /* Window Pos */
+    BOOL Maximized;
+    INT Left;
+    INT Top;
+    INT Width;
+    INT Height;
+    /* Proxy settings */
+    INT Proxy;
+    WCHAR szProxyServer[MAX_PATH];
+    WCHAR szNoProxyFor[MAX_PATH];
+    /* Software source settings */
+    BOOL bUseSource;
+    WCHAR szSourceURL[INTERNET_MAX_URL_LENGTH];
+};
+
+typedef SETTINGS_INFO *PSETTINGS_INFO;
+
+BOOL LoadSettings();
+VOID SaveSettings(HWND hwnd);
+VOID FillDefaultSettings(PSETTINGS_INFO pSettingsInfo);
+
+extern SETTINGS_INFO SettingsInfo;
--- a/base/applications/rapps/include/settings.h
+++ b/base/applications/rapps/include/settings.h
@@ -41,8 +41,8 @@ typedef struct _SettingsRegEntry
     const WCHAR *RegKeyName; // The key name of this field in registery.
 }SETTINGS_REG_ENTRY;
 
-BOOL LoadSettings();
-VOID SaveSettings(HWND hwnd);
+BOOL LoadSettings(PSETTINGS_INFO pSettingsInfo);
+VOID SaveSettings(HWND hwnd, PSETTINGS_INFO pSettingsInfo);
 VOID FillDefaultSettings(PSETTINGS_INFO pSettingsInfo);
 
 extern SETTINGS_INFO SettingsInfo;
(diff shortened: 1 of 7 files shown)


base/applications/rapps/include/settings.h
```diff
@@ -0,0 +1,48 @@
...
+enum SETTINGS_FIELDTYPE
+{
+    SettingsFieldInt,
+    SettingsFieldBool,
+    SettingsFieldString,
+};
+
+typedef struct _SettingsRegEntry
+{
+    size_t Offset;  // Offset of the field in SETTINGS_INFO struct
```
learn-more, line 38: If you change this to a union something like this:
```
union
{
    BOOL* bValue;
    INT* iValue;
    WCHAR* wValue;
};
```
you can create pointers to the settings instead of using the `OFFSET_OF` macro

kernelbin: I probably didn't gey your means.... then how should I know where's the field I want to visit? 🤣

base/applications/rapps/settings.cpp
```diff
@@ -0,0 +1,247 @@
...
+    infoFields.AddTail(new SettingsFieldInt(&(settings.Top), L"WindowPosTop"));
+    infoFields.AddTail(new SettingsFieldInt(&(settings.Width), L"WindowPosWidth"));
+    infoFields.AddTail(new SettingsFieldInt(&(settings.Height), L"WindowPosHeight"));
+    infoFields.AddTail(new SettingsFieldInt(&(settings.Proxy), L"ProxyMode"));
+    infoFields.AddTail(new SettingsFieldString((settings.szProxyServer), MAX_PATH, L"ProxyServer"));
+    infoFields.AddTail(new SettingsFieldString((settings.szNoProxyFor), MAX_PATH, L"NoProxyFor"));
+    infoFields.AddTail(new SettingsFieldBool(&(settings.bUseSource), L"bUseSource"));
+    infoFields.AddTail(new SettingsFieldString((settings.szSourceURL), INTERNET_MAX_URL_LENGTH, L"SourceURL"));
+
+    return;
```
learn-more, line 122: This is not really needed (but does not hurt)

base/applications/rapps/settings.cpp
```diff
@@ -0,0 +1,247 @@
...
+BOOL LoadSettings(PSETTINGS_INFO pSettingsInfo)
+{
+    ATL::CRegKey RegKey;
+    if (RegKey.Open(HKEY_CURRENT_USER, L"Software\\ReactOS\\rapps", KEY_READ) != ERROR_SUCCESS)
+    {
+        return FALSE;
+    }
+
+    BOOL bRet = LoadAllSettings(RegKey, *pSettingsInfo);
+    RegKey.Close();
```
learn-more, line 215: This is automatically done by the class, hence why using classes like this is more convenient :)

base/applications/rapps/settings.cpp
```diff
@@ -0,0 +1,247 @@
...
+                && (wp.flags & WPF_RESTORETOMAXIMIZED)));
+    }
+
+    if (RegKey.Create(HKEY_CURRENT_USER, L"Software\\ReactOS\\rapps", NULL,
+        REG_OPTION_NON_VOLATILE, KEY_WRITE, NULL, NULL) != ERROR_SUCCESS)
+    {
+        return FALSE;
+    }
+
+    BOOL bRet = SaveAllSettings(RegKey, *pSettingsInfo);
```
learn-more, line 244: you can just `return SaveAllSettiongs(....` here, because the CRegKey will handle closing when going out of scope

learn-more (approved):
Some minor details, but even without those addressed we could merge this :)

base/applications/rapps/settings.cpp
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    virtual BOOL Save(CRegKey &key) override
+    {
+        return key.SetStringValue(m_RegName, m_pStringStore) == ERROR_SUCCESS;
+    }
+    virtual BOOL Load(CRegKey &key) override
+    {
+        ULONG nChar = m_StringLen - 1; // make sure the terminating L'\0'
+        LONG lResult = key.QueryStringValue(m_RegName, m_pStringStore, &nChar);
```
learn-more, line 93: This one might overwrite the default value, but still return something else than ERROR_SUCCESS.

learn-more (approved):

PR #3102, opened 2020-08-26
[CRT] Remove remnant '__globallocalestatus' code

SergeGautherie:
Addendum to 183b263 (r55869).
JIRA issue: [CORE-14042](https://jira.reactos.org/browse/CORE-14042)


HBelusca:
`__globallocalestatus` is something that is present (but used as well) in the MS CRT, so I expect it should be present in our CRT as well. It's just that we don't implement yet the per-thread locale functionality yet.

SergeGautherie:
Thanks for the details!

Possible options, until used:
- Remove broken code.
- Keep it, but improve ifdef-out.
- Hack-re-add the actual variable somewhere.
- ...

Extravert-ir:
I choose the option: fix the code, don't mess with hacking hacks

PR #3103, opened 2020-08-27
[NTOS:PS] Initial implementation of VDM Ps support

Fraizeraust:
- Create the "psvdm.c" file and implement PspVdmInitialize()
- Use PspVdmInitialize in PspInitPhase0
- Move the PspDeleteVdmObjects() code into the new file and implement the rest parts of it

Further documentation details of `PspDeleteVdmObjects()` and `PspVdmInitialize()` can be found in [this page](https://fraizeraust.github.io/pages/doc/ntoskrnl/ps/vdm.html).

### DIFF
--- a/ntoskrnl/ps/i386/psvdm.c
+++ b/ntoskrnl/ps/i386/psvdm.c
@@ -1,0 +1,73 @@
+/*
+ * PROJECT:     ReactOS Kernel
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Virtual DOS Machine (VDM) support routines
+ * COPYRIGHT:   Copyright Stefan Ginsberg (stefan.ginsberg@reactos.org)
+ *              Copyright 2020 George Bișoc (george.bisoc@reactos.org)
+ */
+
+/* INCLUDES ******************************************************************/
+
+#include <ntoskrnl.h>
+#define NDEBUG
+#include <debug.h>
+
+/* GLOBALS *******************************************************************/
+
+ERESOURCE VdmIoListCreationResource;
+
+/* FUNCTIONS *****************************************************************/
+
+/**
+ * @brief
+ * Deletes Virtual DOS Machine (DOS) objects, given an Executive process (EPROCESS).
+ * 
+ * @param[in] Process
+ * A pointer to an Executive process.
+ * 
+ * @return
+ * Nothing.
+ * 
+ */
+VOID
+NTAPI
+PspDeleteVdmObjects(IN PEPROCESS Process)
+{
+    /* If there are no VDM objects, just exit */
+    if (Process->VdmObjects == NULL)
+    {
+        DPRINT1("PspDeleteVdmObjects(): No VDM objects could be found, exiting...\n");
+        return;
+    }
+
+    /* Otherwise return the quota it was taking up and free them */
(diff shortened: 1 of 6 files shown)


HBelusca:
Is this (private) function going to be used somewhere else right now? Maybe you could just continue working on the VDM kernel interface before marking this PR as non-draft?

Fraizeraust:
As I said on Mattermost, in order to have a fully complete VDM kernel interface it'd require a huge amount of work which I am comfortable providing separate PRs with code in a gradual manner. In addition to what I've said, I must first and foremost publish documentation details of VDM/Ps VDM related routines before I start writing code.

To answer your question -- no, I'm not going to make this PR as draft. Furthermore both `PspVdmInitialize` and `PspDeleteVdmObjects` are used explicitly in code.

ntoskrnl/ps/i386/psvdm.c
```diff
@@ -0,0 +1,73 @@
...
+
+    /* Otherwise return the quota it was taking up and free them */
+    PsReturnProcessNonPagedPoolQuota(Process->VdmObjects, sizeof(Process->VdmObjects));
+    ExFreePoolWithTag(Process->VdmObjects, '  eK');
+    Process->VdmObjects = NULL;
+
+    /* Remove the VDM resource from the system's resources list and get our process quota */
+    ExDeleteResourceLite(&VdmIoListCreationResource);
+    PsReturnProcessPagedPoolQuota(Process, sizeof(Process));
+    DPRINT1("PspDeleteVdmObjects(): Successfully deleted the VDM objects!\n");
```
tkreuzer, line 51: Why do you delete the global resource on per-process cleanup? And why return the quota for the process here? Non of this seems to make any sense.

tkreuzer (changes requested):

ntoskrnl/ps/i386/psvdm.c
```diff
@@ -0,0 +1,73 @@
...
+{
+    /* If there are no VDM objects, just exit */
+    if (Process->VdmObjects == NULL)
+    {
+        DPRINT1("PspDeleteVdmObjects(): No VDM objects could be found, exiting...\n");
+        return;
+    }
+
+    /* Otherwise return the quota it was taking up and free them */
+    PsReturnProcessNonPagedPoolQuota(Process->VdmObjects, sizeof(Process->VdmObjects));
```
ThFabba, line 44: The first argument should be the `Process` to charge, not VdmObjects.
For the second argument, `VdmObjects` is a pointer so you'll only get `sizeof(PVOID)` here. You'll need the actual struct size.

ntoskrnl/ps/psmgr.c
```diff
@@ -540,7 +540,10 @@ PspInitPhase0(IN PLOADER_PARAMETER_BLOCK LoaderBlock)
     PspCidTable = ExCreateHandleTable(NULL);
     if (!PspCidTable) return FALSE;
 
-    /* FIXME: Initialize LDT/VDM support */
+#if defined(_X86_)
+    /* FIXME: Initialize LDT support */
+    PspVdmInitialize();
```
ThFabba, line 545: If the return value doesn't get checked, why does the function return it?

ThFabba (commented):
Any particular reason you're working on this? Our ntvdm doesn't rely on V86 mode, and there are no plans to change that AFAIK. So this might end up being dead code.

PR #3106, opened 2020-08-27
[KMTESTS:RTL] clang: Ignore a warning on a 'RtlFillMemory()' call

SergeGautherie:

### DIFF
--- a/modules/rostests/apitests/crt/sprintf.c
+++ b/modules/rostests/apitests/crt/sprintf.c
@@ -21,9 +21,7 @@
 #pragma GCC diagnostic ignored "-Wformat"
 #pragma GCC diagnostic ignored "-Wformat-zero-length"
 #pragma GCC diagnostic ignored "-Wnonnull"
-#if __GNUC__ >= 7
 #pragma GCC diagnostic ignored "-Wformat-overflow"
-#endif
 #endif
 
 /* NOTE: This test is not only used for all the CRT apitests, but also for
--- a/modules/rostests/kmtests/rtl/RtlMemory.c
+++ b/modules/rostests/kmtests/rtl/RtlMemory.c
@@ -426,12 +426,12 @@ START_TEST(RtlMemory)
     KeLowerIrql(Irql);
     Status = STATUS_SUCCESS;
     _SEH2_TRY {
-#if defined(__GNUC__)
+#if defined(__GNUC__) || defined(__clang__)
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wmemset-transposed-args"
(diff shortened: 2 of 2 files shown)


modules/rostests/apitests/crt/sprintf.c
```diff
@@ -21,10 +21,8 @@
 #pragma GCC diagnostic ignored "-Wformat"
 #pragma GCC diagnostic ignored "-Wformat-zero-length"
 #pragma GCC diagnostic ignored "-Wnonnull"
-#if __GNUC__ >= 7
```
JoachimHenze, line 24: Why removing that? It will not help for Clang!

modules/rostests/kmtests/rtl/RtlMemory.c
```diff
@@ -426,12 +426,12 @@ START_TEST(RtlMemory)
     KeLowerIrql(Irql);
     Status = STATUS_SUCCESS;
     _SEH2_TRY {
-#if defined(__GNUC__) && __GNUC__ >= 5
+#if defined(__GNUC__) || defined(__clang__)
```
JoachimHenze, line 429: Why not #if defined(__GNUC__) && __GNUC__ >= 5 || defined(__clang__)

modules/rostests/kmtests/rtl/RtlMemory.c
```diff
@@ -426,12 +426,12 @@ START_TEST(RtlMemory)
...
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wmemset-transposed-args"
 #endif
         RtlFillMemory(NULL, 0, 0x55);
-#if defined(__GNUC__) && __GNUC__ >= 5
+#if defined(__GNUC__) || defined(__clang__)
```
JoachimHenze, line 434: Why not #if defined(__GNUC__) && __GNUC__ >= 5 || defined(__clang__)

SergeGautherie:
@JoachimHenze, why keep obsolete code?

JoachimHenze:
Because the ros codebase is large, and we are within a major transition phase when switching the compiler. And responsible devs know, that such a transition usually takes a long time (up to many years for large projects), and within that timeframe still new regressions might pop up, Wise to not unnecessarily break any compatibility during that timeframe therefore.

SergeGautherie:
> unnecessarily break any compatibility

I thought that timeframe was (mostly) over, after the actual switch.
Moreover, parts of old gcc support have already been removed, especially in cmake files...

learn-more:
Yeah, I am done asking every time to remove unrelated changes.

PR #3105, opened 2020-08-27
[GITHUB] clang-cl and msvc-amd64: Have CD uploads, but disabled

SergeGautherie:
Pave the way, but don't waste resources.

JIRA issue: [CORE-17202](https://jira.reactos.org/browse/CORE-17202)

PR #3116, opened 2020-08-31
[GITHUB] clang-cl: Upload CDs

SergeGautherie:
Disabled by default, until they are usable/useful.

CORE-17202

(Remaining part of #3105.)


Extravert-ir:
Thank you for help, but we're clever enough to copy that code from another part of the file when it's needed

PR #3117, opened 2020-08-31
[ROSTESTS] Remove obsolete '#if __GNUC__ >= ...'s

SergeGautherie:
After upgrade to GCC 8.4.

(Split off from #3106.)


Extravert-ir:
There are more than 3500 issues on Jira, why are you bothering us with so minor stuff instead of fixing something which is an issue or a task at least?

PR #3118, opened 2020-09-01
Improve precompiled headers usage

zefklop:
## Purpose

Improve our usage of precompiled headers

## Proposed changes

- Fix the add_pch cmake macro
- Add a global header "reactos_pch.h" which defines __REACTOS_USE_PCH__
- Use this define as a guard to not include the selected header, as cmake includes it through the -include (GCC) or /FI (MSVC) compiler switch

This way we really open the header file _once_ and parse it _once_

Included are a few commits for some modules. With time, we can change all of them.


dll/win32/shell32/wine/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+    shellpath.c
+    shellreg.c
+    shellstring.c
+    shlmenu.c
+    shpolicy.c
+    precomp.h)
+
+add_library(shell32_wine OBJECT ${SOURCE})
+add_pch(shell32_wine precomp.h)
+target_compile_definitions(shell32_wine PRIVATE UNICODE _UNICODE _WINE)
```
zefklop, line 24:
```suggestion
target_compile_definitions(shell32_wine PRIVATE UNICODE _UNICODE _WINE)
add_dependencies(shell32_wine psdk)
```

Extravert-ir:
Two thoughts here:
- Should it be a separate header, maybe it's enough to use `target_compile_definitions` instead?
- As `target_precompile_headers` adds a PCH to every file, maybe we should do that manually when PCH is disabled? (by using `target_sources`) That way we could avoid dealing with conditional `#include` statements

zefklop:
>     * Should it be a separate header, maybe it's enough to use `target_compile_definitions` instead?
I thought about it, but then it adds a define to all compilation units. It's fine in most cases, but sometimes we skip the precompiled header for some files depending on the compiler, so having a define only for the files using the PCH is better.

>     * As `target_precompile_headers` adds a PCH to every file, maybe we should do that manually when PCH is disabled? (by using `target_sources`) That way we could avoid dealing with conditional `#include` statements
Maybe we can do that. But then we have to be careful about dependency check. I'll see if I can get this right.

ThFabba:
I seem to recall learning in the past that compilers are smart enough to avoid opening the file an extra time... however I have nothing to back that up.
Do you have a reference that says this is a good idea? Do you see any actual improvements in compilation times or disk I/O during build?
I'm hesitant about adding these ifdefs to tons of source files if the the benefit isn't very noticeable.

zefklop:
I just réalisés that this can be achieved using #pragma once in the related header file.

I'll come up with something better in another PR.

PR #3120, opened 2020-09-01
[KMTESTS:RTL] clang: Ignore a warning on a 'RtlFillMemory()' call

SergeGautherie:
'...\RtlMemory.c(433,29): warning: 'size' argument to memset is '0'; did you mean to transpose the last two arguments? [-Wmemset-transposed-args]'

(Split off from #3106.)


Extravert-ir:
The change does not make sense without fixing other clang warnings

PR #3125, opened 2020-09-03
[EXPLORER] Implement HSHELL_ACTIVATESHELLWINDOW

katahiromz:
## Purpose

JIRA issue: N/A

- Implement `HSHELL_ACTIVATESHELLWINDOW` shell hook.

### DIFF
--- a/base/shell/explorer/taskswnd.cpp
+++ b/base/shell/explorer/taskswnd.cpp
@@ -1533,6 +1533,10 @@ public:
             break;
 
         case HSHELL_ACTIVATESHELLWINDOW:
+            ::SwitchToThisWindow(m_Tray->GetHWND(), TRUE);
+            ::SetForegroundWindow(m_Tray->GetHWND());
+            break;
+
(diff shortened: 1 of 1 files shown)


base/shell/explorer/taskswnd.cpp
```diff
@@ -1533,6 +1533,10 @@ class CTaskSwitchWnd :
             break;
 
         case HSHELL_ACTIVATESHELLWINDOW:
+            ::SwitchToThisWindow(m_Tray->GetHWND(), TRUE);
```
HBelusca, line 1536: Two naive questions:
- What would happen if FALSE was used here instead?
- If the window was minimized, is the HSHELL_ACTIVATESHELLWINDOW intended also to restore the explorer window to its original size? If so, do these two functions do that? Or should a ShowWindow call also be needed?

katahiromz: FALSE means Alt+Esc that is rarely used.

PR #3127, opened 2020-09-04
[FONTEXT] Initial implementation of CFontExt::DoGetFontTitle

katahiromz:
## Purpose

JIRA issue: [CORE-12861](https://jira.reactos.org/browse/CORE-12861)

This PR is a preparation of fonts folder implementation.

not-working-yet
Not working yet.

## Proposed Changes
- Use  `GetFontResourceInfoW` to get the font name.

### DIFF
--- a/dll/shellext/fontext/CFontExt.cpp
+++ b/dll/shellext/fontext/CFontExt.cpp
@@ -615,7 +615,7 @@ CFontExt::DoGetFontTitle(IN LPCWSTR pszFontPath, OUT CStringW& strFontName)
     strFontName.ReleaseBuffer();
     if (ret)
     {
-        TRACE("pszFontName: %S\n", pszBuffer);
+        TRACE("pszFontName: %S\n", (LPCWSTR)strFontName);
         return S_OK;
     }
 
--- a/dll/shellext/fontext/CFontExt.cpp
+++ b/dll/shellext/fontext/CFontExt.cpp
@@ -7,6 +7,7 @@
  */
 
 #include "precomp.h"
+#include "undocgdi.h" // for GetFontResourceInfoW
 
 WINE_DEFAULT_DEBUG_CHANNEL(fontext);
 
@@ -563,8 +564,8 @@ HRESULT CFontExt::DoInstallFontFile(LPCWSTR pszFontPath, LPCWSTR pszFontsDir, HK
     WCHAR szDestFile[MAX_PATH];
     LPCWSTR pszFileTitle = PathFindFileName(pszFontPath);
 
-    WCHAR szFontName[512];
-    if (!DoGetFontTitle(pszFontPath, szFontName, sizeof(szFontName)))
+    CStringW strFontName;
+    if (!DoGetFontTitle(pszFontPath, strFontName))
         return E_FAIL;
 
     RemoveFontResourceW(pszFileTitle);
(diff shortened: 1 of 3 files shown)


dll/shellext/fontext/CFontExt.cpp
```diff
@@ -568,7 +564,7 @@ HRESULT CFontExt::DoInstallFontFile(LPCWSTR pszFontPath, LPCWSTR pszFontsDir, HK
     LPCWSTR pszFileTitle = PathFindFileName(pszFontPath);
 
     WCHAR szFontName[512];
-    if (!DoGetFontTitle(pszFontPath, szFontName))
+    if (!DoGetFontTitle(pszFontPath, szFontName, sizeof(szFontName)))
```
learn-more, line 567: can we improve this to use CString instead of a fixed size buffer?

dll/shellext/fontext/CFontExt.cpp
```diff
@@ -601,8 +597,31 @@ HRESULT CFontExt::DoInstallFontFile(LPCWSTR pszFontPath, LPCWSTR pszFontsDir, HK
...
 
-HRESULT CFontExt::DoGetFontTitle(LPCWSTR pszFontPath, LPCWSTR pszFontName)
+typedef BOOL (WINAPI *FN_GetFontResourceInfoW)(LPCWSTR, DWORD *, void*, DWORD);
+
+HRESULT
+CFontExt::DoGetFontTitle(IN LPCWSTR pszFontPath, OUT LPWSTR pszFontName,
+                         IN DWORD cbFontName)
 {
-    // TODO:
+    HMODULE hGdi32 = LoadLibraryA("gdi32");
```
learn-more, line 606: Why not define the prototype for the function and directly call it, why LoadLibrary/GetProcAddress?

dll/shellext/fontext/CFontExt.cpp
katahiromz: OK.

dll/shellext/fontext/CFontExt.cpp
katahiromz: OK.

dll/shellext/fontext/CFontExt.cpp
```diff
@@ -589,7 +586,8 @@ HRESULT CFontExt::DoInstallFontFile(LPCWSTR pszFontPath, LPCWSTR pszFontsDir, HK
     }
 
     DWORD cbData = (wcslen(pszFileTitle) + 1) * sizeof(WCHAR);
-    LONG nError = RegSetValueExW(hkeyFonts, szFontName, 0, REG_SZ, (const BYTE *)szFontName, cbData);
+    LONG nError = RegSetValueExW(hkeyFonts, strFontName, 0, REG_SZ,
+                                 (const BYTE *)(LPCWSTR)strFontName, cbData);
```
learn-more, line 590: Why is the length calculated on something else than the actual string that is stored here?

learn-more: probably the argument to be written should be `pszFileTitle` ?

katahiromz: Fixed.

learn-more (approved):

sdk/include/reactos/undocgdi.h
```diff
@@ -26,19 +26,24 @@ typedef struct GDI_DRAW_STREAM_TAG
...
-#endif
\ No newline at end of file
+EXTERN_C BOOL WINAPI
+GetFontResourceInfoW(
+    _In_z_ LPCWSTR lpFileName,
+    _Inout_ DWORD *pdwBufSize,
+    _Out_writes_to_opt_(*pdwBufSize, 1) PVOID lpBuffer,
+    _In_ DWORD dwType);
+
+#endif
```
HBelusca: It should not be necessary to specify "EXTERN_C" everytime.
The whole header must be made c++-compatible by putting everything enclosed inside a
```
#ifdef __cplusplus
extern "C" {
#endif

// Here are all the functions, vars, etc.

// and at the end of the file:
#ifdef __cplusplus
}
#endif
```
This is done in all PSDK headers etc. see an example here too: https://git.reactos.org/?p=reactos.git;a=blob;f=sdk/lib/conutils/screen.h;hb=HEAD

katahiromz: OK.

PR #3131, opened 2020-09-05
[DESK] Fix crash when selecting 3D Text screen saver CORE-8217

zecarlos1957:
[DESK] Fix crash when selecting 3D Text screen saver
JIRA issue: [CORE-8217](https://jira.reactos.org/browse/CORE-8217)

CreateProcess needs dwCreationFlags parameter


JoachimHenze:
There were concerns rised within CORE-8217 whether this fix is the correct way to approach the problem. And even after the fix it may hang (moving the parent window). I don't think it is right to touch desk.cpl at all here! I do vote for not committing this PR.

SergeGautherie:
Is this issue/fix reproducible on Windows XP/S03?

zecarlos1957:
This is not the correct  approach. So i close this PR

zecarlos1957:
This is not the correct  approach

PR #3134, opened 2020-09-06
[SHELLFIND] Encoding-aware file search

katahiromz:
## Purpose

JIRA issue: [CORE-17250](https://jira.reactos.org/browse/CORE-17250)

- Consider UTF-8/UTF-16/UTF-16BE encodings in file search.
- Optimize for speed.

TestData:
TestData

BEFORE:
before

AFTER:
after

## Proposed Changes

- Look the BOM (byte order mark) or NUL characters at the beginning of the file if any and find the text.
- If found, it became a match of a file item.

### DIFF
--- a/dll/win32/browseui/shellfind/CFindFolder.cpp
+++ b/dll/win32/browseui/shellfind/CFindFolder.cpp
@@ -199,7 +199,11 @@ static BOOL SearchFile(LPCWSTR lpFilePath, _SearchData *pSearchData)
     }
 
     if (size == INVALID_FILE_SIZE)
-        size = 32 * 1024 * 1024; // Use first 32 MB
+    {
+        MEMORYSTATUS status = { sizeof(status) };
+        GlobalMemoryStatus(&status);
+        size = status.dwAvailPhys * 2 / 3;
+    }
 
     HANDLE hFileMap = CreateFileMappingW(hFile, NULL, PAGE_READONLY, 0, size, NULL);
     CloseHandle(hFile);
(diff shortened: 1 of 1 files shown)


dll/win32/browseui/shellfind/CFindFolder.cpp
```diff
@@ -213,14 +213,47 @@ static UINT SearchFile(LPCWSTR lpFilePath, _SearchData *pSearchData)
         return 0;
 
     UINT uMatches = 0;
-    // Check for UTF-16 BOM
-    if (size >= 2 && lpFileContent[0] == 0xFF && lpFileContent[1] == 0xFE)
+    if (size >= 2 &&
```
HBelusca, line 216: See https://git.reactos.org/?p=reactos.git;a=blob;f=base/applications/cmdutils/more/more.c;h=161818a5e6ba6d0219e84e887d23467df5aabe9d;hb=HEAD#l149 for more details.

katahiromz: Perfect encoding detection is expensive in this case.

Extravert-ir:
This is a good example of a code with bad design, which will work in very rare cases.

1. Why the charset detection is made inside `SearchFile` function? You're mixing up the logic here
2. Why writing your own function at all, while there is `IsTextUnicode` and what @hbelusca pointed out?
3. In real life there are almost no files with BOM. Even less files are in anything other than UTF-8 or ANSI. So this will be a "dead code" in 99% of cases

Let's not give reasons for trolls to blame us for poor quality code

PR #126, opened 2017-11-15
Improve font charset penalty system

katahiromz:
## Purpose

JIRA issue: [CORE-12419](https://jira.reactos.org/browse/CORE-12419)

PR #3136, opened 2020-09-06
[EXPLORER] Registry key not closed on error (CORE-14519)

KyleKatarn31:
## Purpose

- If goto cleanup is called, the key won't be closed.

JIRA issue: [CORE-14519](https://jira.reactos.org/browse/CORE-14519)

## Proposed changes

- Close key as 1st cleanup step (since neither nominal case which close the key, nor the failed key opening would lead there)

### DIFF
--- a/base/shell/explorer/shellservice.cpp
+++ b/base/shell/explorer/shellservice.cpp
@@ -92,6 +92,7 @@ HRESULT InitShellServices(HDPA * phdpa)
         if (FAILED_UNEXPECTEDLY(hr))
         {
             ERR("CLSIDFromString failed %08x.\n", hr);
+            RegCloseKey(hkey);
             goto cleanup;
         }
 
@@ -99,6 +100,7 @@ HRESULT InitShellServices(HDPA * phdpa)
         if (FAILED_UNEXPECTEDLY(hr))
         {
             ERR("CoCreateInstance failed %08x.\n", hr);
+            RegCloseKey(hkey);
             goto cleanup;
         }
 
@@ -112,6 +114,7 @@ HRESULT InitShellServices(HDPA * phdpa)
     {
         ERR("RegEnumValueW failed %08x.\n", ret);
         hr = HRESULT_FROM_WIN32(GetLastError());
+        RegCloseKey(hkey);
         goto cleanup;
     }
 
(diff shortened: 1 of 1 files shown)


base/shell/explorer/shellservice.cpp
```diff
@@ -112,6 +114,7 @@ HRESULT InitShellServices(HDPA * phdpa)
     {
         ERR("RegEnumValueW failed %08x.\n", ret);
         hr = HRESULT_FROM_WIN32(GetLastError());
+        RegCloseKey(hkey);
```
binarymaster, line 117: Just do it on cleanup if `hkey` is not NULL?

KyleKatarn31: You're right and no need to check since both previous RegCloseKey and failed RegOpenKeyW lead to a return condition preventing entering Goto cleanup.

Change done.

base/shell/explorer/shellservice.cpp
```diff
@@ -126,6 +126,7 @@ HRESULT InitShellServices(HDPA * phdpa)
     return count > 0 ? S_OK : S_FALSE;
 
 cleanup:
+    RegCloseKey(hkey);
```
SergeGautherie, line 129: Not right: check lines 118 and 123...

Doug-Lyons: @SergeGautherie, I believe that Line 118 just needs to be deleted after the RegCloseKey() is inserted under the Cleanup: label. Or looking at it another way the RegCloseKey() in Line 118 needs to be moved under the Cleanup: label. Thanks.

SergeGautherie: @Doug-Lyons, no: check line 126.

Doug-Lyons: @SergeGautherie, You are correct. Also, Line #126 should probably be as follows:
hr = count > 0 ? S_OK : S_FALSE;
Do you agree that these three changes should handle all of the obvious conditions now?

SergeGautherie: @Doug-Lyons, no...

KyleKatarn31: OK 118 just need to come after 123. Nice catch @SergeGautherie

KyleKatarn31:
@SergeGautherie @Doug-Lyons : Is it better / good enougth now ?

Doug-Lyons:
@KyleKatarn31, I believe that what you have now will work well. Thanks.

PR #3137, opened 2020-09-06
[GDI32] Fix double cx's in call to HANDLE_METADC in win32ss painting.c

Doug-Lyons:
## Typo Fix for win32ss//gdi/gdi32/objects/painting.c

_Fix 2 cases of cx used instead of cy._

No JIRA related issue

I stumbled onto this looking into fixing the StretchBlt function.
I discussed this with Jim Tabor and he agreed it was a typo error.


HBelusca (approved):

JoachimHenze (approved):
Wow! Nice catch!

PR #3139, opened 2020-09-06
Second take on DLL export versioning system

wjk:
## Purpose

This PR is a follow-up copy of #1872, since work on that PR seems to have stopped. I have fixed the merge conflicts in the original PR, addressed review comments from there, and wrote a unit test that exercises the functionality added.

JIRA issue: [CORE-11288](https://jira.reactos.org/browse/CORE-11288)

## Proposed changes

- Add a feature to spec2def to generate a RosCompat descriptor (contained in a special section) pointing to a mask-array, that specifies for each named export on which Windows versions it is present.
- Add a feature to the loader in NTDLL that checks for a RosCompat descriptor, and if present, uses it to patch the export table, splitting it into a public and private part.
- Modify the loader to allow ReactOS DLLs (those that contain a RosCompat descriptor, even if it is empty) to import from private/hidden exports.
- Add a CMake setting, `ENABLE_DLL_VERSIONING`, that defaults to ON. If enabled, the RosCompat descriptors will be emitted, and all APIs (up to and including Windows 10 APIs) will be compiled. If disabled, the RosCompat descriptors will not be emitted, and only NT5.2 APIs will be available. (The RosCompat parsing code will still be compiled in this case, but will do nothing.)

## Open Questions

- [ ] Should we remove the `*_vista` DLLs?
- [ ] Do we have enough unit tests? The current test is a program that exercises `kernel32!InitOnceExecuteOnce`. This API is available in NT6.0+. Two copies of this version are compiled: `versioning_win10.exe` (whose manifest contains the Windows 10 compat GUID), which works; and `versioning_xp.exe` (whose manifest contains no compat GUIDs), which is expected to crash when run because InitOnceExecuteOnce is hidden by the loader.


binarymaster:
> since work on that PR seems to have stopped

I don't think so, but maybe it's better to ask in chat about such things...

dll/apisets/api-ms-win-core-localization-l1-1-0.spec
```diff
@@ -5,7 +5,7 @@
 @ stdcall -version=0x601+ FindNLSString() kernel32.FindNLSString
 @ stdcall -version=0x601+ FindNLSStringEx() kernel32.FindNLSStringEx
 @ stdcall -version=0x601+ GetACP() kernel32.GetACP
-@ stdcall -version=0x601+ GetCPFileNameFromRegistry() kernel32.GetCPFileNameFromRegistry
+@ stdcall -version=0x501-0x600 GetCPFileNameFromRegistry() kernel32.GetCPFileNameFromRegistry
```
SergeGautherie, line 8: Unwanted change?

dll/win32/kernel32/kernel32.spec
```diff
@@ -32,7 +32,7 @@
 @ stub -version=0x600+ BaseGenerateAppCompatData
 @ stdcall BaseInitAppcompatCacheSupport()
 @ stdcall BaseIsAppcompatInfrastructureDisabled() IsShimInfrastructureDisabled
-@ stdcall -version=0x501-0x502 BaseProcessInitPostImport()
+@ stdcall BaseProcessInitPostImport()
```
SergeGautherie, line 35: Unwanted change?

wjk: No; see my commit comment for 84de3f2235675b5d73850dca0be61d91b3344b2c for why this is needed.

dll/apisets/api-ms-win-core-localization-l1-1-0.spec
wjk: This was intentional as well; see my commit comment on a4713a054901971bdc85045b28efb38d6fe68276 for detaials.

SergeGautherie: Whatever the error was, that is a wrong fix, as apisets are all NT6+ at least.

dll/win32/kernel32/kernel32.spec
SergeGautherie: That looks like a wrong fix as well:
either this function/export actually exists on NT6+ and this change should be a separate PR,
or it does not and there should be no change.

learn-more: does not really matter, but reactos components should be able to bypass this mechanism so either way should work,
removing this version or leaving it in place.

tkreuzer: They should all be 0x600+, so that the loader can identify DLLs that have only Vista+ exports and fail when trying to load the DLL, emulating the non-existance of the DLL for apps that check this to see, whether it's running on Vista.

tkreuzer:
Instead of assuming it was abandoned, you could have asked. I am working on it.

dll/win32/kernel32/kernel32.spec
SergeGautherie: > They should all be 0x600+

Which 'they' are you referring to? Did you mean to comment about the apisets instead?
As changing `BaseProcessInitPostImport` to `0x600+` from `0x501-0x502` seems even more unexpected...

wjk:
Superseded by #3239. Please accept my apologies for my stepping on Timo's toes by opening this PR.

PR #3144, opened 2020-09-07
[RAPPS] Display custom applications icons for installed applications CORE-17257

oleg-dubinskiy:
## Purpose

Allow Rapps to display custom app icons for installed software instead of just default Rapps icon.
In my opinion, default icon for each application looks monotonously and not informatively, because it doesn't display what it should. So displaying the custom app's icon instead will make it less hard to understand which app is actually installed.
Moreover, it already has been done for available apps (during GSoC 2020).

JIRA issue: [CORE-17257](https://jira.reactos.org/browse/CORE-17257)

## Proposed changes

- Implement `RetrieveIcon` helper function in `CInstalledApplicationInfo` class, which retrueves the current app's icon from registrty, same as it done for `CAvailableApplicationInfo`.
- Use it for loading the icon in `CAppsListView::AddInstalledApplication` function, via `ExtractIconW`. Load default Rapps icon only when the app has no its custom icon.
- Retrieve `DisplayIcon` value from registry in `CInstalledApps::Enum` function, same as other registry values (like app name, description, etc).Store it in `szDisplayIcon` string, which is used in `CInstalledApplicationInfo::RetrieveIcon` for retrieving the data of that value.
- Increase `LISTVIEW_ICON_SIZE` macro from 24 to 32, so 32x32 icon size is now used instead of 24x24. This makes displayed icons more accurate, since most of apps contain 32x32 icon, so they look a bit distorted with 24x24 size.

## TODO

- [x] Display the icons also for apps, whose strings in registry are terminated by zero. Currently, my patch implements displaying the icons only for apps, which have direct string. E. g. for: `C:\AppPath\AppName.exe` it displays the icon correctly, but for `C:\AppPath\AppName.exe,0` it does not. :frowning_face: 

## Result

Before:
rapps_icons_before

After:
rapps_icons_after

### DIFF
--- a/base/applications/rapps/appview.cpp
+++ b/base/applications/rapps/appview.cpp
@@ -1407,7 +1407,7 @@ BOOL CAppsListView::AddInstalledApplication(CInstalledApplicationInfo *InstAppIn
     if (!hIcon)
     {
         /* Load default icon */
-        hIcon = (HICON)LoadIconW(hInst, MAKEINTRESOURCE(IDI_MAIN));
+        hIcon = LoadIconW(hInst, MAKEINTRESOURCEW(IDI_MAIN));
     }
 
     int IconIndex = ImageList_AddIcon(m_hImageListView, hIcon);
--- a/base/applications/rapps/appview.cpp
+++ b/base/applications/rapps/appview.cpp
@@ -1387,11 +1387,11 @@ BOOL CAppsListView::AddInstalledApplication(CInstalledApplicationInfo *InstAppIn
     ATL::CStringW szIconPath;
     if (InstAppInfo->RetrieveIcon(szIconPath))
     {
-        INT Ret = PathParseIconLocationW((LPWSTR)szIconPath.GetString());
+        PathParseIconLocationW((LPWSTR)szIconPath.GetString());
 
         hIcon = ExtractIconW(hInst,
                              szIconPath.GetString(),
-                             Ret);
+                             0);
     }
 
     if (!hIcon)
--- a/base/applications/rapps/appview.cpp
+++ b/base/applications/rapps/appview.cpp
@@ -1398,9 +1398,9 @@ BOOL CAppsListView::AddInstalledApplication(CInstalledApplicationInfo *InstAppIn
         }
         else if (!Ret)
         {
-            hIcon = (HICON)ExtractIconW(hInst,
-                szIconPath.GetString(),
-                0);
+            hIcon = ExtractIconW(hInst,
+                                 szIconPath.GetString(),
+                                 0);
         }
     }
 
--- a/base/applications/rapps/appview.cpp
+++ b/base/applications/rapps/appview.cpp
@@ -1389,19 +1389,9 @@ BOOL CAppsListView::AddInstalledApplication(CInstalledApplicationInfo *InstAppIn
     {
         INT Ret = PathParseIconLocationW((LPWSTR)szIconPath.GetString());
 
-        /* Check if icon location is zero-terminated */
-        if (Ret == 0)
-        {
-            hIcon = ExtractIconW(hInst,
-                                 szIconPath.GetString(),
-                                 Ret);
-        }
-        else if (!Ret)
-        {
-            hIcon = ExtractIconW(hInst,
-                                 szIconPath.GetString(),
-                                 0);
-        }
+        hIcon = ExtractIconW(hInst,
+                             szIconPath.GetString(),
+                             Ret);
     }
 
     if (!hIcon)
--- a/base/applications/rapps/appview.cpp
+++ b/base/applications/rapps/appview.cpp
@@ -1382,7 +1382,33 @@ BOOL CAppsListView::AddInstalledApplication(CInstalledApplicationInfo *InstAppIn
         return FALSE;
     }
 
-    HICON hIcon = (HICON)LoadIconW(hInst, MAKEINTRESOURCEW(IDI_MAIN));
+    /* Load icon from registry */
+    HICON hIcon = NULL;
+    ATL::CStringW szIconPath;
+    if (InstAppInfo->RetrieveIcon(szIconPath))
+    {
+        INT Ret = PathParseIconLocationW((LPWSTR)szIconPath.GetString());
+
+        /* Check if icon location is zero-terminated */
+        if (Ret == 0)
+        {
+            hIcon = (HICON)ExtractIconW(hInst,
+                szIconPath.GetString(),
+                Ret);
+        }
+        else if (!Ret)
+        {
+            hIcon = (HICON)ExtractIconW(hInst,
+                szIconPath.GetString(),
+                0);
+        }
+    }
+
+    if (!hIcon)
+    {
+        /* Load default icon */
+        hIcon = (HICON)LoadIconW(hInst, MAKEINTRESOURCE(IDI_MAIN));
+    }
 
     int IconIndex = ImageList_AddIcon(m_hImageListView, hIcon);
     DestroyIcon(hIcon);
--- a/base/applications/rapps/appview.cpp
+++ b/base/applications/rapps/appview.cpp
@@ -1392,9 +1392,9 @@ BOOL CAppsListView::AddInstalledApplication(CInstalledApplicationInfo *InstAppIn
         /* Check if icon location is zero-terminated */
         if (Ret == 0)
         {
-            hIcon = (HICON)ExtractIconW(hInst,
-                szIconPath.GetString(),
-                Ret);
+            hIcon = ExtractIconW(hInst,
+                                 szIconPath.GetString(),
+                                 Ret);
         }
         else if (!Ret)
         {
@@ -2050,4 +2050,3 @@ BOOL CApplicationView::ItemCheckStateChanged(BOOL bChecked, LPVOID CallbackParam
     return TRUE;
 }
 // **** CApplicationView ****
-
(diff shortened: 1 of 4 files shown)


base/applications/rapps/appview.cpp
```diff
@@ -1382,7 +1382,21 @@ BOOL CAppsListView::AddInstalledApplication(CInstalledApplicationInfo *InstAppIn
...
+    HICON hIcon = NULL;
+    ATL::CStringW szIconPath;
+    if (InstAppInfo->RetrieveIcon(szIconPath))
+    {
+        hIcon = (HICON)ExtractIconW(hInst,
+            szIconPath.GetString(),
+            0);
+    }
+
+    if (!hIcon || GetLastError() != ERROR_SUCCESS)
```
learn-more, line 1395: GetLastError check should not be here.

base/applications/rapps/appview.cpp
```diff
@@ -1382,7 +1382,21 @@ BOOL CAppsListView::AddInstalledApplication(CInstalledApplicationInfo *InstAppIn
         return FALSE;
     }
 
-    HICON hIcon = (HICON)LoadIconW(hInst, MAKEINTRESOURCEW(IDI_MAIN));
+    /* Load icon from registry */
+    HICON hIcon = NULL;
+    ATL::CStringW szIconPath;
+    if (InstAppInfo->RetrieveIcon(szIconPath))
+    {
+        hIcon = (HICON)ExtractIconW(hInst,
```
learn-more, line 1390: You might want to play with `PathParseIconLocationW` ;)

learn-more: or just extract it manually from the CStringW

learn-more: If the string also starts with an `@` you might consider passing it to `SHLoadIndirectString`

base/applications/rapps/appview.cpp
oleg-dubinskiy: Fixed, thanks.
Btw I added `GetLastError` check only because it has been done also for available apps:
https://github.com/reactos/reactos/blob/master/base/applications/rapps/appview.cpp#L1418
If that check is not required, shouldn't it also be removed there?

base/applications/rapps/appview.cpp
oleg-dubinskiy: Fixed, thanks @learn-more! :+1: 
`PathParseIconLocationW` works as well for me with zero-terminated strings.
But I don't remember any software, which has `@` at the start of the string from `DisplayIcon` value. Is there any? :thinking:

base/applications/rapps/appview.cpp
```diff
@@ -1382,7 +1382,33 @@ BOOL CAppsListView::AddInstalledApplication(CInstalledApplicationInfo *InstAppIn
...
+    if (InstAppInfo->RetrieveIcon(szIconPath))
+    {
+        INT Ret = PathParseIconLocationW((LPWSTR)szIconPath.GetString());
+
+        /* Check if icon location is zero-terminated */
+        if (Ret == 0)
+        {
+            hIcon = (HICON)ExtractIconW(hInst,
+                szIconPath.GetString(),
+                Ret);
```
HBelusca, line 1397:
```suggestion
            hIcon = ExtractIconW(hInst,
                                 szIconPath.GetString(),
                                 Ret);
```

base/applications/rapps/appview.cpp
```diff
@@ -1382,7 +1382,33 @@ BOOL CAppsListView::AddInstalledApplication(CInstalledApplicationInfo *InstAppIn
...
+        }
+        else if (!Ret)
+        {
+            hIcon = (HICON)ExtractIconW(hInst,
+                szIconPath.GetString(),
+                0);
```
HBelusca, line 1403:
```suggestion
            hIcon = ExtractIconW(hInst,
                                 szIconPath.GetString(),
                                 0);
```

base/applications/rapps/appview.cpp
```diff
@@ -1382,7 +1382,33 @@ BOOL CAppsListView::AddInstalledApplication(CInstalledApplicationInfo *InstAppIn
...
+        }
+    }
+
+    if (!hIcon)
+    {
+        /* Load default icon */
+        hIcon = (HICON)LoadIconW(hInst, MAKEINTRESOURCE(IDI_MAIN));
```
HBelusca, line 1410:
```suggestion
        hIcon = LoadIconW(hInst, MAKEINTRESOURCEW(IDI_MAIN));
```

HBelusca (changes requested):

oleg-dubinskiy:
Fixed. Thanks @HBelusca.
I just followed the same coding style as in all other code of this file, but if it's better, I'm OK with these changes.

base/applications/rapps/include/appview.h
```diff
@@ -20,7 +20,7 @@
 
 using namespace Gdiplus;
 
-#define LISTVIEW_ICON_SIZE 24
```
kernelbin, line 23: may I ask why this is changed 🤣

kernelbin (approved):

base/applications/rapps/include/appview.h
binarymaster: According to the [Oleg's comment in VK](https://vk.com/wall-1086956_82280?reply=82283&thread=82282), he changed it because almost all existing apps provide either 16x16 or 32x32 icons. RAPPS itself used 24x24 icon which was the same for all entries, and when we try to use unique icons, due to unsupported size, they are resized to 24x24 which makes them looking bad. Oleg decided to select 32x32 size because he thought 16x16 would be too small.

base/applications/rapps/appview.cpp
```diff
@@ -1382,7 +1382,33 @@ BOOL CAppsListView::AddInstalledApplication(CInstalledApplicationInfo *InstAppIn
...
-    HICON hIcon = (HICON)LoadIconW(hInst, MAKEINTRESOURCEW(IDI_MAIN));
+    /* Load icon from registry */
+    HICON hIcon = NULL;
+    ATL::CStringW szIconPath;
+    if (InstAppInfo->RetrieveIcon(szIconPath))
+    {
+        INT Ret = PathParseIconLocationW((LPWSTR)szIconPath.GetString());
+
+        /* Check if icon location is zero-terminated */
+        if (Ret == 0)
```
HBelusca, line 1393: I don't understand this (and the "else if (!Ret)" below).
You are basically doing this:
```
if (Ret == 0)
{
    hIcon = ExtractIcon(..., Ret /* which is == 0 */);
}
else if (Ret != 0) // Note that just using "else" would have been sufficient...
{
    hIcon = ExtractIcon(..., 0);
}
```
In other words in both cases, you did the very same call `hIcon = ExtractIcon(..., 0);`
Did you want instead to use the actual returned icon index in the ExtractIcon call (and not just always '0')? If so you don't need to use all this complicated ifs, but just need to use `hIcon = ExtractIcon(..., Ret);` .
And instead, if you always want to use icon 0, just ignore the returned value of the PathParseIconLocationW() call and directly call `hIcon = ExtractIcon(..., 0);` .

HBelusca (changes requested):

base/applications/rapps/appview.cpp
oleg-dubinskiy: Fixed. Thanks for the tip @HBelusca. :+1: 
It still displays all icons with the first way: just use return value from `PathParseIconLocationW` and don't do any checks of it.
But the 2nd way proposed by you is not suitable in this case, because without parsing the string from registry, it displays only the icons from strings without icon index at the end (as I already said in the PR description).

HBelusca: >  But the 2nd way proposed by you is not suitable in this case, because without parsing the string from registry, it displays only the icons from strings without icon index at the end (as I already said in the PR description).

In the 2nd way I still was proposing to use `PathParseIconLocationW()`, but to not take care about the returned value that you store into Ret; in other words the 2nd way would have been just:
```c
PathParseIconLocationW((LPWSTR)szIconPath.GetString());

hIcon = ExtractIconW(hInst,
                     szIconPath.GetString(),
                     0);
```
(So, this would have been only if you wanted to be able to parse these registry strings with the index appended, but not take into account that index and use instead always zero.)

HBelusca (approved):

base/applications/rapps/appview.cpp
```diff
@@ -1382,7 +1382,23 @@ BOOL CAppsListView::AddInstalledApplication(CInstalledApplicationInfo *InstAppIn
         return FALSE;
     }
 
-    HICON hIcon = (HICON)LoadIconW(hInst, MAKEINTRESOURCEW(IDI_MAIN));
+    /* Load icon from registry */
+    HICON hIcon = NULL;
+    ATL::CStringW szIconPath;
+    if (InstAppInfo->RetrieveIcon(szIconPath))
+    {
+        PathParseIconLocationW((LPWSTR)szIconPath.GetString());
```
learn-more, line 1390: You'd need the return from this to get the icon index

HBelusca: Depends what @oleg-dubinskiy really wants to use as the icon (either hardcodes always icon zero, or actually uses whatever icon index was specified in the icon path)

oleg-dubinskiy: Yes, I think hardcoded zero will be better, since according to my analysis, all apps provide executables with the main icon as first in the index (detected as `0`), so we don't need other icons here (which have `1`, `2`, `3` etc. index numbers).

oleg-dubinskiy: Return value is required only for the random suitable index value.

learn-more: In that case it might make sense to add a comment explaining that you deliberately ignore this?

learn-more (approved):

PR #3146, opened 2020-09-07
[PCIIDEX] Add missing exported function

disean:
## Purpose

Useful for debugging IDE controller minidrivers.

JIRA issue: [CORE-17256](https://jira.reactos.org/browse/CORE-17256)

AFTER:
drv

### DIFF
--- a/drivers/storage/ide/pciidex/miniport.c
+++ b/drivers/storage/ide/pciidex/miniport.c
@@ -11,6 +11,9 @@
 #define NDEBUG
 #include <debug.h>
 
+/** @brief Global debugging level. Valid values are between 0 (Error) and 3 (Trace). */
+ULONG PciIdeDebug = 0;
+
 static DRIVER_DISPATCH PciIdeXForwardOrIgnore;
 static NTSTATUS NTAPI
 PciIdeXForwardOrIgnore(
@@ -88,6 +91,32 @@ PciIdeXPnpDispatch(
 		return PciIdeXFdoPnpDispatch(DeviceObject, Irp);
 	else
 		return PciIdeXPdoPnpDispatch(DeviceObject, Irp);
+}
+
+/**
+ * @brief Prints the given string with printf-like formatting to the kernel debugger.
+ * @param[in] DebugPrintLevel Level of the debug message.
+ *                            Valid values are between 0 (Error) and 3 (Trace).
+ * @param[in] DebugMessage    Format of the string/arguments.
+ * @param[in] ...             Variable number of arguments matching the format
+ *                            specified in \a DebugMessage.
+ * @sa PciIdeDebug
+ */
+VOID
+PciIdeXDebugPrint(
+    _In_ ULONG DebugPrintLevel,
+    _In_z_ _Printf_format_string_ PCCHAR DebugMessage,
+    ...)
+{
+    va_list ap;
+
+    /* Check if we can print anything */
+    if (DebugPrintLevel <= PciIdeDebug)
+        DebugPrintLevel = 0;
+
+    va_start(ap, DebugMessage);
+    vDbgPrintEx(DPFLTR_PCIIDE_ID, DebugPrintLevel, DebugMessage, ap);
+    va_end(ap);
 }
 
 NTSTATUS NTAPI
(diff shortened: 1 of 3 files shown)


drivers/storage/ide/pciidex/pciidex.spec
```diff
@@ -1,3 +1,4 @@
+@ cdecl PciIdeXDebugPrint(long ptr)
```
SergeGautherie, line 1:
```suggestion
@ varargs PciIdeXDebugPrint(long ptr)
```

drivers/storage/ide/pciidex/miniport.c
```diff
@@ -90,6 +93,30 @@ PciIdeXPnpDispatch(
...
+ */
+VOID
+PciIdeXDebugPrint(
+    _In_ ULONG DebugPrintLevel,
+    _In_z_ _Printf_format_string_ PCCHAR DebugMessage,
+    ...)
+{
+    va_list ap;
+
+    if (PciIdeDebug >= DebugPrintLevel)
```
SergeGautherie, line 112:
```suggestion
    if (DebugPrintLevel <= PciIdeDebug)
```

drivers/storage/ide/pciidex/miniport.c
```diff
@@ -90,6 +93,30 @@ PciIdeXPnpDispatch(
...
+ *                             Valid values are between 0 (Error) and 3 (Trace).
+ * \param[in] DebugMessage     Format of the string/arguments.
+ * \param[in] ...              Variable number of arguments matching the format
+ *                             specified in \a DebugMessage.
+ * \sa PciIdeDebug
+ */
+VOID
+PciIdeXDebugPrint(
+    _In_ ULONG DebugPrintLevel,
+    _In_z_ _Printf_format_string_ PCCHAR DebugMessage,
```
SergeGautherie, line 107: Also sync https://git.reactos.org/?p=reactos.git;a=blob;f=drivers/storage/inc/ide.h;hb=HEAD#l478.

drivers/storage/ide/pciidex/miniport.c
```diff
@@ -90,6 +93,30 @@ PciIdeXPnpDispatch(
...
+    ...)
+{
+    va_list ap;
+
+    if (DebugPrintLevel <= PciIdeDebug)
+        DebugPrintLevel = 0;
```
SergeGautherie, line 113: Add a comment to explain what this block is doing?

SergeGautherie (commented):
> `#define PciIdeXDebugPrint(x) PciIdeXDebugPrint x`

While there, could you double-check what this define is doing, as in needed for? Thanks.

Extravert-ir (approved):
1. Looks like this file uses tabs for indentation (damn, somebody needs to make a proper clang-format file for us)
2. We use `/**` and `@param` syntax for Doxygen comments

Everything else LGTM

disean:
> While there, could you double-check what this define is doing, as in needed for?

Similar to videoprt implementation
https://git.reactos.org/?p=reactos.git;a=blob;f=sdk/include/ddk/video.h;hb=92f4746845727e5f21e8184f56db580eceef1c6d#l73

> Looks like this file uses tabs for indentation

I'll fix the formatting issues later.

binarymaster (approved):

drivers/storage/ide/pciidex/miniport.c
SergeGautherie: Well, then, if I guess correctly, why not rather early return here and use 0 directly in the following call?
And use/add `#define DPFLTR_ERROR_LEVEL 0`, instead of a magic value.

drivers/storage/ide/pciidex/pciidex.spec
```diff
@@ -1,3 +1,4 @@
+@ varargs PciIdeXDebugPrint(long ptr)
```
SergeGautherie, line 1:
```suggestion
@ varargs PciIdeXDebugPrint(long str)
```
Sorry, missed it the first time.

SergeGautherie (changes requested):
> Similar to videoprt implementation

Well, not obvious (to me):
`  73 #define VideoDebugPrint(x) VideoPortDebugPrint x` is an alias,
whereas
`486 #define PciIdeXDebugPrint(x) PciIdeXDebugPrint x` refers to itself.
Bear with me, could you explain what this syntax actually do?

PR #3147, opened 2020-09-08
[SHELLFIND] Split encoding detection code

katahiromz:
## Purpose

Split encoding detection code to improve code quality. @Extravert-ir This is follow-up of #3134 (7376cdebb6fc945798b29212c9f3239ea52cf89f).

JIRA issue: [CORE-17250](https://jira.reactos.org/browse/CORE-17250)

## Proposed changes

- Split encoding detection code as newly added `IsDataUnicode` function from `SearchFile` function.

### DIFF
--- a/dll/win32/browseui/shellfind/CFindFolder.cpp
+++ b/dll/win32/browseui/shellfind/CFindFolder.cpp
@@ -340,6 +340,9 @@ static BOOL SearchFile(LPCWSTR lpFilePath, _SearchData *pSearchData)
     if (!pbContents)
         return FALSE;
 
+    if (size > 4 * 1024) // limit of 4 KB
+        size = 4 * 1024;
+
     ENCODING encoding;
     IsDataUnicode(pbContents, size, &encoding, NULL);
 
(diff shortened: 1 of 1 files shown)


katahiromz:
It's slow. I can't recommend.

HBelusca:
> It's slow. I can't recommend.

Perhaps (have you done actual tests?), but if it detects the file encoding more accurately (with less false positives) than before, then it's better.

learn-more:
> > It's slow. I can't recommend.
> 
> Perhaps (have you done actual tests?), but if it detects the file encoding more accurately (with less false positives) than before, then it's better.

Not if it takes weeks to search a floppy drive. 
Might does up by just looking at the first 1k bytes or so?

Extravert-ir:
> Might speed up by just looking at the first 1k bytes or so?

4k aka one page, I think

katahiromz:
`RtlIsTextUnicode` function in `sdk/lib/rtl/unicode.c` has 256 bytes limit.

PR #3149, opened 2020-09-09
[KERNEL32_VISTA][NTDLL_VISTA][RTL_VISTA] Move Vista Rtl functions from kernel32_vista and ntdll_vista to rtl_vista

tkreuzer:
* Move RtlRunOnce functions from kernel32_vista to rtl_vista and export them from ntdll_vista
* Move condvar.c and srw.c from ntdll_vista to rtl_vista
* Move ntdll_vista build script to ntdll

The RtlRunOnce functions are taken from wine, completely unmodified.
The code that was in kernel32_vista had change that used a global keyed_event handle, but was never initialized, so we were still passing NULL thus using the global ExpCritSecOutOfMemoryEvent.

Testbot results: https://reactos.org/testman/compare.php?ids=74165,74166,74170,74172,74178


sdk/lib/rtl/runonce.c
```diff
@@ -0,0 +1,130 @@
+
```
HBelusca, line 1: may need some "header" telling this comes from wine.

HBelusca (approved):
LGTM.

binarymaster (approved):

sdk/lib/rtl/runonce.c
```diff
@@ -0,0 +1,130 @@
...
+/* PSDK/NDK Headers */
```
ThFabba, line 2: rtl uses `rtl.h` as a PCH, so you should include that here to avoid differences between PCH and non-PCH builds.

dll/win32/ntdll_vista/CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ list(APPEND SOURCE
     DllMain.c
     condvar.c
     srw.c
+    ${REACTOS_SOURCE_DIR}/sdk/lib/rtl/runonce.c
```
ThFabba, line 14: Can we just link ntdll_vista with the `rtl` library to make this simpler?
Alternatively, `runonce.c` could just live here like the other files do, instead of being part of rtl?

tkreuzer: You can't link rtl, since ntdll_vista already links to ntdll, which will create a mess. But this will soon be exported by ntdll (export will be hidden unless the appcompat setting specifies vista, but our own dlls will be able to link to it secretly).

sdk/lib/rtl/runonce.c
tkreuzer: Says it below: ```/* Taken from Wine ntdll/sync.c */```

sdk/lib/rtl/runonce.c
tkreuzer: Fixed.

tkreuzer:
I made a few changes:
1. use rtl_vista instead of manually including the files
2. move condvar.c and srw.c from ntdll_vista to rtl_vista, too
3. move the buildscripts for ntdll_vista into a subfolder of ntdll

learn-more (approved):

katahiromz (approved):

PR #3152, opened 2020-09-10
[REACTOS] Do not use LoadLibraryEx() NT6+ flags

SergeGautherie:
These modules are not compiled as NT6+.

JIRA issue: [CORE-12004](https://jira.reactos.org/browse/CORE-12004)


learn-more (changes requested):
Please revert the comment/winetest (UNRELATED) changes.

learn-more (approved):
Thanks.

ThFabba:
So do these calls actually fail on Server 2003, or is this pure speculation?!

SergeGautherie:
> do these calls actually fail on Server 2003

If I understand correctly, they do!
https://reactos.org/testman/detail.php?id=49569557&prev=0
https://git.reactos.org/?p=reactos.git;a=blob;f=modules/rostests/winetests/kernel32/loader.c;hb=HEAD#l890

HBelusca:
Yes they do (e.g. in eventvwr -- not in the PR because it was already done -- I had to disable this `LOAD_LIBRARY_AS_IMAGE_RESOURCE` flag, because otherwise when testing on Win2003 I couldn't load the resource dll I wanted).

PR #3157, opened 2020-09-10
[winsrv] improve zh-CN translation

chirsz-ever:
The new word is more in line with PRC people's usage habits, and it
is actually used in the Simplified Chinese translation of Windows.

### DIFF
--- a/win32ss/user/winsrv/consrv/frontends/gui/lang/zh-CN.rc
+++ b/win32ss/user/winsrv/consrv/frontends/gui/lang/zh-CN.rc
@@ -5,7 +5,7 @@ BEGIN
     IDS_EDIT "编辑"
     IDS_MARK "标记"
     IDS_COPY "复制\tEnter"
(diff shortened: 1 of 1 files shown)


chirsz-ever:
I have another translation suggestions, but I'm not sure how to do with it, because it is in `dll/win32/comctl32/lang/comctl_Zh.rc`, which is part of code sharing with Wine. Wine seems to combine all translations to [a single file](https://source.winehq.org/git/wine.git/blob/HEAD:/po/zh_CN.po), will ReactOS fork this action? Or I should just send two patches to here and Wine?

binarymaster:
Cc @reactos/lang-chinese

kernelbin:
> ems to combine all tran

I think contribute to Wine directly would be fine.... ? ( I'm not sure )

robsean:
I think that send two patches to here and Wine is find.

In this way, we can check whether the translation is correct

PR #3164, opened 2020-09-12
[SHELL32_APITEST] Strengthen ShellExecCmdLine testcase

katahiromz:
## Purpose

`ShellExecCmdLine` should support opening the special folders ("My Documents" and "Control Panel" etc.).

JIRA issue: [CORE-16939](https://jira.reactos.org/browse/CORE-16939)

## Proposed changes

- Add tests for "My Documents" and "Control Panel".
- Add some comments for usability.

Win10:
Win10
Successful.

ReactOS:
ReactOS
There are some failures.

### DIFF
--- a/modules/rostests/apitests/shell32/ShellExecCmdLine.cpp
+++ b/modules/rostests/apitests/shell32/ShellExecCmdLine.cpp
@@ -13,9 +13,6 @@
 #include <debug.h>
 #include <stdio.h>
 
-#define ShellExecCmdLine ShellExecCmdLine // In testing on ReactOS
-//#undef ShellExecCmdLine // In testing on Vista+
-
 #ifndef SECL_NO_UI
     #define SECL_NO_UI          0x2
     #define SECL_LOG_USAGE      0x8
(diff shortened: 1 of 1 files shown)


modules/rostests/apitests/shell32/ShellExecCmdLine.cpp
```diff
@@ -13,7 +13,8 @@
 #include <debug.h>
 #include <stdio.h>
 
-//#define ShellExecCmdLine ShellExecCmdLine
+#define ShellExecCmdLine ShellExecCmdLine // In testing on ReactOS
+//#undef ShellExecCmdLine // In testing on Vista+
```
learn-more, line 17: ?

katahiromz: ShellExecCmdLine is Vista+ private function. So WinXP doesn't export it. Defining ShellExecCmdLine macro will enable the below synced code. We can't automatically switch the behavior.

learn-more: If you completely remove the define,
rename the function (inside the apitest) to `proxy_ShellExecCmdLine` (or something along those lines),
then if the GetProcAddress failes, you fill the `g_pShellExecCmdLine` with the `proxy_ShellExecCmdLine` instead of the actual function, it should work for both?
unless I am missing something

katahiromz: OK. Done.

PR #3166, opened 2020-09-12
[AMD64] Get some stuff merged

tkreuzer:
This PR contains a few commits from the amd64 branch that can be merged.
All of the code is pure x64 code and doesn't affect x86 at all.


ntoskrnl/ke/amd64/stubs.c
```diff
@@ -72,8 +72,8 @@ KiDpcInterruptHandler(VOID)
     }
 
     /* Go back to old irql and disable interrupts */
```
HBelusca, line 74:
```suggestion
    /* Disable interrupts and go back to old irql */
```

HBelusca (approved):
Looks reasonable to me :)

ntoskrnl/ps/query.c
```diff
@@ -1902,7 +1902,7 @@ NtSetInformationProcess(IN HANDLE ProcessHandle,
             /* Only supported on x86 */
 #if defined (_X86_)
             Ke386SetIOPL();
-#else
+#elif !defined(_M_AMD64)
             Status = STATUS_NOT_IMPLEMENTED;
 #endif
```
binarymaster, line 1907:
```suggestion
#elif defined(_M_AMD64)
            NOTHING;
#else
            Status = STATUS_NOT_IMPLEMENTED;
#endif
```

binarymaster (approved):
One nit-pick, otherwise LGTM.

hal/halx86/apic/rtctimer.c
```diff
@@ -33,7 +33,7 @@ ULONG
 RtcClockRateToIncrement(UCHAR Rate)
 {
     ULONG Freqency = ((32768 << 1) >> Rate);
-    return (1000000 + (Freqency/2)) / Freqency;
+    return (10000000 + (Freqency/2)) / Freqency;
```
ThFabba, line 36: I certainly wouldn't mind a comment for this function, since I personally can't reason about this change at all.

ntoskrnl/ps/query.c
```diff
@@ -1902,7 +1902,7 @@ NtSetInformationProcess(IN HANDLE ProcessHandle,
             /* Only supported on x86 */
 #if defined (_X86_)
             Ke386SetIOPL();
-#else
+#elif !defined(_M_AMD64)
```
ThFabba, line 1905: Is this because
* we have stuff (incorrectly) relying on this SetInfo call to succeed,
* AMD64 should support this call, but it's unimplemented in ROS and we want to fake success, or
* AMD64 supports this call but does nothing (i.e. this is the correct implementation)?

sdk/lib/crt/float/amd64/fpreset.S
```diff
@@ -6,6 +6,7 @@
 FUNC _fpreset
     .endprolog
     fninit
+    ret
```
ThFabba, line 9: lol nice

ntoskrnl/ke/amd64/stubs.c
```diff
@@ -330,6 +330,9 @@ KiSystemCallHandler(
     TrapFrame->TrapFrame = (ULONG64)Thread->TrapFrame;
     Thread->TrapFrame = TrapFrame;
 
+    /* We don't have an exception frame yet */
```
ThFabba, line 333: Does "yet" here refer to an unimplemented feature, i.e. we should be doing something like the x86 version
```
1754     TrapFrame->ExceptionList = KeGetPcr()->NtTib.ExceptionList;
1755     KeGetPcr()->NtTib.ExceptionList = EXCEPTION_CHAIN_END;
```
? Or does it simply mean we'll do this at some other point in the code?

ThFabba (approved):

hal/halx86/apic/apictrap.S
```diff
@@ -20,7 +20,7 @@ TRAP_ENTRY HalpProfileInterrupt, (TF_VOLATILES OR TF_SEND_EOI)
 
 PUBLIC ApicSpuriousService
 ApicSpuriousService:
-    iret
```
Extravert-ir, line 23: Just curious, shouldn't `iret` do the same thing as `iretq` when compiled as amd64?

Extravert-ir (approved):

hal/halx86/apic/apictrap.S
tkreuzer: See https://www.felixcloutier.com/x86/iret:iretd

ntoskrnl/ke/amd64/stubs.c
tkreuzer: You are confusing ExceptionList with Exception frame. The former only exists on x86 as all other architectures use table based SEH. The latter is to store non-volatile register information on an exception. On non-x86 this isn't done on a syscall to avoid the performance overhead of saving all the registers, when it's not needed (with one exception: NtSetContextThread, which needs to restore non-volatiles from the context when returning, but for that there is a workaround)

ntoskrnl/ps/query.c
tkreuzer: On 2k3 it returns STATUS_SUCCESS, on Vista and later it returns STATUS_NOT_IMPLEMENTED. I'll add a comment.

hal/halx86/apic/rtctimer.c
tkreuzer: I added some comments. Hope that makes it clearer.

ntoskrnl/ps/query.c
tkreuzer: I added an appcompat version check (doesn't work yet, but will be once the appcompat stuff is merged) to behave appropriately based on the appcompat version.

boot/bootdata/txtsetup.sif
```diff
@@ -19,7 +19,8 @@ _1 = %cdname%,%cdtagfile%,,""
 1 = %cdname%,%cdtagfilei%,,\reactos
 
 [SourceDisksNames.amd64]
-1 = %cdname%,%cdtagfilea%,,\amd64
+;1 = %cdname%,%cdtagfilea%,,\amd64
+1 = %cdname%,%cdtagfilei%,,\reactos
```
learn-more, line 23: This is supposed to be fixed later?

sdk/lib/rtl/process.c
```diff
@@ -493,3 +493,19 @@ RtlGetCurrentProcessorNumber(VOID)
...
+    /* Get the current PEB */
+    PPEB Peb = RtlGetCurrentPeb();
+    if (Peb == NULL)
+    {
+        /* Default to Server 2003 */
+        return _WIN32_WINNT_WS03;
+    }
+
+    /* Calculate OS version from PEB fields */
+    return (Peb->OSMajorVersion << 8) | Peb->OSMinorVersion;
```
learn-more, line 510: I am assuming this is supposed to be substituted by the selected appcompat version later?

boot/bootdata/txtsetup.sif
tkreuzer: Possibly? I don't care enough to fix it.

sdk/lib/rtl/process.c
tkreuzer: Well I plan to actually patch the PEB.

ThFabba, line 510: I think in kernel mode we need to probe and SEH-protect this.

hal/halx86/apic/rtctimer.c
ThFabba: Much, thanks!

sdk/lib/rtl/process.c
HBelusca: Correct.

boot/bootdata/txtsetup.sif
HBelusca: This is currently done like this because, we have separate bootcds for 32-bit and 64-bit ReactOS (due to the way we separately do builds for x86 and x64). If one day we want to support one bootcd that can contain both the 32-bit and the 64-bit versions of ReactOS, then we will have to put the corresponding installation files in these i386 / amd64 subdirectories instead.

sdk/lib/rtl/process.c
tkreuzer: There is no need to "probe" it, since the pointer comes from kernel mode.
```
PPEB
NTAPI
RtlGetCurrentPeb(VOID)
{
    return ((PEPROCESS)(KeGetCurrentThread()->ApcState.Process))->Peb;
}
```
I also don't see why SEH would be needed, since the Peb is a "NoChange" allocation and you can't "free" it or change it's protection (see MiCreatePebOrTeb()).

zefklop: Can we make it not look like it's an "official" RTL function?

tkreuzer: Renamed it to RtlRosGetAppcompatVersion(). Everyone knows: if there is a "Ros" in there, it's a hack :)

PR #3169, opened 2020-09-12
[CABMAN] Add commandline support for creating a cab with folders

learn-more:
Update rapps example scripts for new usage

JIRA issue: [CORE-17230](https://jira.reactos.org/browse/CORE-17230)

We need to create 2 .cab files, since rapps cannot handle extracting folders without a code change.
To work around this, the first file does not include the icons (but rapps cannot extract it anyway),
and the second file (with a new name) does include the icons (and since rapps needs an update to change the filename, folder extraction can be added in the same update)

### DIFF
--- a/base/applications/rapps/CreateCabFile.bat
+++ b/base/applications/rapps/CreateCabFile.bat
@@ -1,12 +1,32 @@
 @echo off
 
-cd ..\..\..\media
+IF "%1"=="" GOTO show_usage
+IF "%2"=="" GOTO show_usage
+IF "%3"=="" GOTO show_usage
 
-mkdir rapps\utf16
+SET CABMAN_CMD=%1
+SET UTF16LE_CMD=%2
+SET RAPPSDB_PATH=%3
 
-for %%f in (rapps\*.txt) do (
-     ..\output-MinGW-i386\host-tools\utf16le.exe "rapps\%%~nf.txt" "rapps\utf16\%%~nf.txt"
+mkdir "%RAPPSDB_PATH%\utf16"
+
+echo Converting txt files to utf16
+for %%f in (%RAPPSDB_PATH%\*.txt) do (
+     %UTF16LE_CMD% "%RAPPSDB_PATH%\%%~nf.txt" "%RAPPSDB_PATH%\utf16\%%~nf.txt"
 )
 
-..\output-MinGW-i386\host-tools\cabman.exe -M mszip -S rapps\rappmgr.cab rapps\utf16\*.txt
-rmdir /s /q rapps\utf16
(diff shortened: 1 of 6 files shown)


base/applications/rapps/CreateCabFile.bat
```diff
@@ -1,12 +1,33 @@
...
+for %%f in (%RAPPSDB_PATH%\*.txt) do (
+     %UTF16LE_CMD% "%RAPPSDB_PATH%\%%~nf.txt" "%RAPPSDB_PATH%\utf16\%%~nf.txt"
 )
 
-..\output-MinGW-i386\host-tools\cabman.exe -M mszip -S rapps\rappmgr.cab rapps\utf16\*.txt
-rmdir /s /q rapps\utf16
+echo Building rappmgr.cab
+%CABMAN_CMD% -M mszip -S "%RAPPSDB_PATH%\rappmgr.cab" "%RAPPSDB_PATH%\utf16\*.txt"
+
+
```
HBelusca, line 21:
```suggestion
```

base/applications/rapps/CreateCabFile.sh
```diff
@@ -1,9 +1,30 @@
...
 done
-cd ..
-../../host-tools/cabman -M mszip -S rapps/rappmgr.cab rapps/utf16/*.txt
-rm -r rapps/uft16
\ No newline at end of file
+
+echo Building rappmgr.cab
+$CABMAN_CMD -M mszip -S "$RAPPSDB_PATH/rappmgr.cab" "$RAPPSDB_PATH/utf16/*.txt"
+
+
```
HBelusca, line 23:
```suggestion
```

HBelusca:
@learn-more : So, if I understand correctly, the .cab file themselves contain the flattened file list, and the embedded inf-like info at their beginning contain the subdirectories where the files have to be decompressed?

learn-more:
> @learn-more : So, if I understand correctly, the .cab file themselves contain the flattened file list, and the embedded inf-like info at their beginning contain the subdirectories where the files have to be decompressed?

no, it is simply the trick to create a filename like 'icons\ico1.ico' and it is up to the viewer (winrar, rapps or whatever) to actually form 'folders' out of that.
image

unless that is what you mean with 'inf-like' info, then yes

ColinFinck (approved):
Looks okay-ish. I don't have time for a thorough review of each line, so if there is something particular you want me to look at, please let me know. Otherwise approved.

It's sad to see that we still need to implement filesystem stuff twice for Windows and Unix.
Maybe `std::filesystem` can help here in a subsequent PR.

PR #3172, opened 2020-09-13
[SHELL32][SHLWAPI] Fix ShellMessageBoxW use of fixed sized buffer

KyleKatarn31:
## Purpose

- ShellMessageBoxW truncates mesages longer than 100 chars due to use of fixed size buffer

JIRA issue: [CORE-17271](https://jira.reactos.org/browse/CORE-17271)
ShellMessageBoxW

## Proposed changes

- Fix ShellMessageBoxW implementation
- Proper redirect of ShellMessageBoxWrapW to ShellMessageBoxW
- Fixes in ShellMessageBoxA implementation

ShellMessageBoxW_fix

### DIFF
--- a/dll/win32/shell32/wine/shellord.c
+++ b/dll/win32/shell32/wine/shellord.c
@@ -371,7 +371,8 @@ int ShellMessageBoxW(
             if (szText) LoadStringW(hInstance, LOWORD(lpText), szText, len + 1);
         }
         pszText = szText;
-        if (!pszText) {
+        if (!pszText)
+        {
             WARN("Failed to load id %d\n", LOWORD(lpText));
             __ms_va_end(args);
             return 0;
@@ -446,7 +447,8 @@ int ShellMessageBoxA(
             if (szText) LoadStringA(hInstance, LOWORD(lpText), szText, len + 1);;
         }
         pszText = szText;
-        if (!pszText) {
+        if (!pszText)
+        {
             WARN("Failed to load id %d\n", LOWORD(lpText));
             __ms_va_end(args);
             return 0;
--- a/dll/win32/shell32/wine/shellord.c
+++ b/dll/win32/shell32/wine/shellord.c
@@ -345,36 +345,51 @@ int ShellMessageBoxW(
 	UINT uType,
 	...)
 {
-	WCHAR	szText[100],szTitle[100];
-	LPCWSTR pszText = szText, pszTitle = szTitle;
-	LPWSTR  pszTemp;
-	__ms_va_list args;
-	int	ret;
+    WCHAR *szText = NULL, szTitle[100];
+    LPCWSTR pszText, pszTitle = szTitle;
+    LPWSTR pszTemp;
+    __ms_va_list args;
+    int ret;
 
-	__ms_va_start(args, uType);
-	/* wvsprintfA(buf,fmt, args); */
+    __ms_va_start(args, uType);
 
-	TRACE("(%p,%p,%p,%p,%08x)\n",
-	    hInstance,hWnd,lpText,lpCaption,uType);
+    TRACE("(%p,%p,%p,%p,%08x)\n", hInstance, hWnd, lpText, lpCaption, uType);
 
-	if (IS_INTRESOURCE(lpCaption))
-	  LoadStringW(hInstance, LOWORD(lpCaption), szTitle, sizeof(szTitle)/sizeof(szTitle[0]));
-	else
-	  pszTitle = lpCaption;
+    if (IS_INTRESOURCE(lpCaption))
+        LoadStringW(hInstance, LOWORD(lpCaption), szTitle, _countof(szTitle));
+    else
+        pszTitle = lpCaption;
 
-	if (IS_INTRESOURCE(lpText))
-	  LoadStringW(hInstance, LOWORD(lpText), szText, sizeof(szText)/sizeof(szText[0]));
-	else
-	  pszText = lpText;
+    if (IS_INTRESOURCE(lpText))
+    {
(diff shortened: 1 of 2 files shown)


dll/win32/shlwapi/ordinal.c
```diff
@@ -4812,51 +4812,7 @@ DWORD WINAPI GetUIVersion(void)
...
-                   pszText, 0, 0, (LPWSTR)&pszTemp, 0, &args);
-
-    __ms_va_end(args);
-
-    ret = MessageBoxW(hWnd, pszTemp, pszTitle, uType);
-
-    HeapFree(GetProcessHeap(), 0, szText);
-    LocalFree(pszTemp);
-    return ret;
+    return ShellMessageBoxW(hInstance, hWnd, lpText, lpCaption, uType);
```
learn-more, line 4815: This is not a 'proper' redirect, you are omitting all vararg arguments.
Also, it looks like you are redirecting exactly the other way around as proposed.

dll/win32/shell32/wine/shellord.c
```diff
@@ -403,36 +418,52 @@ int ShellMessageBoxA(
...
+      pszTitle = lpCaption;
 
-	if (IS_INTRESOURCE(lpText))
-	  LoadStringA(hInstance, LOWORD(lpText), szText, sizeof(szText));
-	else
-	  pszText = lpText;
+    if (IS_INTRESOURCE(lpText))
+    {
+        const char *ptr;
+        UINT len = LoadStringA(hInstance, LOWORD(lpText), (LPSTR)&ptr, 0);
```
learn-more, line 441: Didn't know the A version could do this...

learn-more (changes requested):
Please ensure your code at least works before proposing a PR...

dll/win32/shell32/wine/shellord.c
```diff
@@ -345,36 +345,51 @@ int ShellMessageBoxW(
...
 {
-	WCHAR	szText[100],szTitle[100];
-	LPCWSTR pszText = szText, pszTitle = szTitle;
-	LPWSTR  pszTemp;
-	__ms_va_list args;
-	int	ret;
+    WCHAR *szText = NULL, szTitle[100];
+    LPCWSTR pszText, pszTitle = szTitle;
+    LPWSTR pszTemp;
+    __ms_va_list args;
```
katahiromz, line 351: Can we delete __ms_?

dll/win32/shell32/wine/shellord.c
```diff
@@ -345,36 +345,51 @@ int ShellMessageBoxW(
...
+        const WCHAR *ptr;
+        UINT len = LoadStringW(hInstance, LOWORD(lpText), (LPWSTR)&ptr, 0);
+
+        if (len)
+        {
+            szText = HeapAlloc(GetProcessHeap(), 0, (len + 1) * sizeof(WCHAR));
+            if (szText) LoadStringW(hInstance, LOWORD(lpText), szText, len + 1);
+        }
+        pszText = szText;
+        if (!pszText) {
```
katahiromz, line 374: Insert newline after paren?

dll/win32/shell32/wine/shellord.c
KyleKatarn31: https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-loadstringa ?

dll/win32/shlwapi/ordinal.c
KyleKatarn31: Yes but looking at other wrapper, that's the usual way to do it for other functions. @HBelusca : Did you have a special reason for proposing the other way around ?

KyleKatarn31: > This is not a 'proper' redirect, you are omitting all vararg arguments.
> Also, it looks like you are redirecting exactly the other way around as proposed.

How could I pass varags ?

dll/win32/shell32/wine/shellord.c
learn-more: Seems that MSDN is not always right :)
The A version does not do this (how could it, the string is not stored as Ansi version in the resources?),
only the W version is doing this.

dll/win32/shlwapi/ordinal.c
learn-more: He proposed it this way, since that is what windows is doing.
Windows directly forwards this in the imports, which also saves having to deal with the varargs.

HBelusca: I said that Windows redirects shell32!ShellMessageBoxW directly to shlwapi!ShellMessageBoxWrapW !! and NOT the other way around as you are doing in this commit.

HBelusca (changes requested):

KyleKatarn31:
OK sorry, in this case, i'll close the PR and let somebody else do the fix.
All I can do is a simpler fix that chirurgically deal with the flaw in ShellMessageBoxW but i prefer to let the redirect and ShellMessageBoxA modifications to a more experienced dev.

@learn-more @HBelusca : Would you prefer 2 steps (1 do a quick and simple one first to fix the bug, then somebody else do a larger redesign later) or would one of you take care of it all-in-one quickly ?

dll/win32/shell32/wine/shellord.c
KyleKatarn31: Oups :-)

dll/win32/shell32/wine/shellord.c
KyleKatarn31: Wrong code on my side (missing args management)

PR #3173, opened 2020-09-13
[INPORT] Add driver for bus mouse devices

disean:
## Purpose

Basic input support. Driver also supports PC (Untested, but should work).

AFTER:
input

## Proposed changes

- Install mouse driver on NEC PC-98 series
- Add driver for bus mouse devices

### DIFF
--- a/drivers/input/inport/hardware.c
+++ b/drivers/input/inport/hardware.c
@@ -1,0 +1,433 @@
+/*
+ * PROJECT:     ReactOS InPort (Bus) Mouse Driver
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Hardware support code
+ * COPYRIGHT:   Copyright 2020 Dmitry Borisov (di.sean@protonmail.com)
+ */
+
+/* Note: Some code was taken from Linux */
+
+/* INCLUDES *******************************************************************/
+
+#include "inport.h"
+
+#define NDEBUG
+#include <debug.h>
+
+/* GLOBALS ********************************************************************/
+
+#ifdef ALLOC_PRAGMA
+#pragma alloc_text(PAGE, InPortInitializeMouse)
+#endif
+
+#define READ_MOUSE(DeviceExtension, Port) \
+    READ_PORT_UCHAR((DeviceExtension)->IoBase + (Port))
+
+#define WRITE_MOUSE(DeviceExtension, Port, Data) \
+    WRITE_PORT_UCHAR((DeviceExtension)->IoBase + (Port), (Data))
+
+/*
+ * NEC
+ */
+#define NEC_BM_DATA          0x00
+
+#define NEC_BM_CONTROL       0x04
+    #define NEC_INT_ENABLE       0x00
+    #define NEC_INT_DISABLE      0x10
+
+    #define NEC_READ_X_LOW       0x00
+    #define NEC_READ_X_HIGH      0x20
+    #define NEC_READ_Y_LOW       0x40
+    #define NEC_READ_Y_HIGH      0x60
+
+    #define NEC_INPUT_CAPTURE    0x00
+    #define NEC_INPUT_HOLD       0x80
+
+#define NEC_BM_CONFIG        0x06
+    #define NEC_PPI_INT_ENABLE   0x08
+    #define NEC_PPI_INT_DISABLE  0x09
+    #define NEC_PPI_HC_NO_CLEAR  0x0E
+    #define NEC_PPI_HC_CLEAR     0x0F
+    #define NEC_PPI_DEFAULT_MODE 0x93
+
+#define NEC_BM_INT_RATE      0x4002
+    #define NEC_RATE_120_HZ      0x00
+    #define NEC_RATE_60_HZ       0x01
+    #define NEC_RATE_30_HZ       0x02
+    #define NEC_RATE_15_HZ       0x03
+
+#define NEC_BM_HIRESO_BASE   (PUCHAR)0x61
+
+#define NEC_BUTTON_RIGHT    0x20
+#define NEC_BUTTON_LEFT     0x80
+
+/*
+ * Microsoft InPort
+ */
+#define MS_INPORT_CONTROL    0x00
+    #define INPORT_REG_BTNS    0x00
+    #define INPORT_REG_X       0x01
+    #define INPORT_REG_Y       0x02
+    #define INPORT_REG_MODE    0x07
+    #define INPORT_RESET       0x80
+
+#define MS_INPORT_DATA       0x01
+    #define INPORT_MODE_IRQ    0x01
+    #define INPORT_MODE_BASE   0x10
+    #define INPORT_MODE_HOLD   0x20
+
+#define MS_INPORT_SIGNATURE  0x02
+
+#define MS_BUTTON_MIDDLE   0x01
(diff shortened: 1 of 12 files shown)


ntoskrnl/io/pnpmgr/pnpmgr.c
```diff
@@ -1065,7 +1065,12 @@ IopEnumerateDetectedDevices(
    UNICODE_STRING HardwareIdKeyboard = RTL_CONSTANT_STRING(L"*PNP0303\0");
    static ULONG DeviceIndexKeyboard = 0;
    const UNICODE_STRING IdentifierMouse = RTL_CONSTANT_STRING(L"PointerController");
+   // FIXME: IopEnumerateDetectedDevices() should be rewritten.
```
binarymaster, line 1068: Suggestion to make this comment more detailed, based on [this message](https://chat.reactos.org/reactos/pl/5e1xra5hqpgx88a11a6phgbmsh).
```suggestion
   /* FIXME: IopEnumerateDetectedDevices() enumerates legacy devices using hardcoded PnP identifiers.
    * It should be rewritten by implementing an actual parser of LegacyXlate sections of driver INF files. */
```

drivers/input/inport/hardware.c
```diff
@@ -0,0 +1,418 @@
...
+    ULONG InputDataConsumed;
+    INPORT_RAW_DATA RawData;
+    ULONG ButtonDiff;
+
+    UNREFERENCED_PARAMETER(Dpc);
+    UNREFERENCED_PARAMETER(Irp);
+    UNREFERENCED_PARAMETER(Context);
+
+    /* Copy raw data */
+    RawData = DeviceExtension->RawData;
```
ThFabba, line 130: It's generally not safe to access the same data in the ISR and the DPC. Should this use `KeAcquireInterruptSpinLock` or similar?

drivers/input/inport/pnp.c
```diff
@@ -0,0 +1,285 @@
...
+
+    IoReleaseRemoveLockAndWait(&DeviceExtension->RemoveLock, Irp);
+
+    if (DeviceExtension->InterruptObject)
+        IoDisconnectInterrupt(DeviceExtension->InterruptObject);
+
+    /* Wait for DPC */
+    KeDelayExecutionThread(KernelMode, FALSE, &Delay);
+
+    InPortStopMouse(DeviceExtension);
```
ThFabba, line 218: You should tell the device to stop issuing interrupts before calling `IoDisconnectInterrupt`, not after.

ThFabba, line 218: You can get `IRP_MN_REMOVE_DEVICE` even if `IRP_MN_START_DEVICE` was never called. So `DeviceExtension->IoBase` may not have been initialized here.

HBelusca:
Build error
`D:\a\reactos\reactos\src\drivers\input\inport\wmi.c(163): error C4189: 'IrpSp': local variable is initialized but not referenced`

binarymaster:
> error C4189: 'IrpSp': local variable is initialized but not referenced

`UNREFERENCED_PARAMETER(IrpSp);` probably should help.

drivers/input/inport/hardware.c
```diff
@@ -0,0 +1,418 @@
...
+    _In_opt_ PVOID Context)
+{
+    UCHAR Temp;
+    CHAR DeltaX, DeltaY;
+    PINPORT_DEVICE_EXTENSION DeviceExtension = (PINPORT_DEVICE_EXTENSION)Context;
+
+    _Analysis_assume_(DeviceExtension != NULL);
+    UNREFERENCED_PARAMETER(Interrupt);
+
+    switch (DeviceExtension->MouseType)
```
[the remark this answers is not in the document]
disean: These devices don't have any status registers.

drivers/input/inport/hardware.c
```diff
@@ -0,0 +1,424 @@
...
+NTAPI
+InPortIsr(
+    _In_ PKINTERRUPT Interrupt,
+    _In_ PVOID Context)
+{
+    UCHAR Temp;
+    CHAR DeltaX, DeltaY;
+    PINPORT_DEVICE_EXTENSION DeviceExtension = Context;
+
+    UNREFERENCED_PARAMETER(Interrupt);
```
binarymaster, line 241:
```suggestion
    /* Bus mouse devices don't have a status register to check
     * whether this interrupt is indeed for us */
    UNREFERENCED_PARAMETER(Interrupt);
```

drivers/input/inport/hardware.c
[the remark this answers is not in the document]
binarymaster: I think it's worth to have a comment in the code about it.

binarymaster (approved):
Minor suggestion about the comment, otherwise looks good to me. :+1:

drivers/input/inport/hardware.c
```diff
@@ -0,0 +1,428 @@
...
+    if (ButtonDiff != 0)
+    {
+        switch (DeviceExtension->MouseType)
+        {
+            case NecBusMouse:
+            {
+                if (ButtonDiff & NEC_BUTTON_LEFT)
+                {
+                    if (RawData.Buttons & NEC_BUTTON_LEFT)
+                        DeviceExtension->MouseInputData.ButtonFlags |= MOUSE_LEFT_BUTTON_UP;
```
ThFabba, line 150: The NEC and Logitech mouse seem to be up when the flag is set, whereas the MS mouse is down. Is this a bug?
If not, it could make sense to have the flag names reflect that they mean "button is not pressed", or have a comment about this.

drivers/input/inport/hardware.c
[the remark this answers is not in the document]
ThFabba: One possible thing to do could be to calculate the `ButtonDiff`, and return false if it's zero. But I guess that might not be reliable either, so your call...

drivers/input/inport/inport.c
```diff
@@ -0,0 +1,239 @@
...
+    _In_ PDEVICE_OBJECT DeviceObject,
+    _Inout_ PIRP Irp)
+{
+    NTSTATUS Status;
+    PINPORT_DEVICE_EXTENSION DeviceExtension = DeviceObject->DeviceExtension;
+
+    DPRINT("%s(%p, %p) %X\n", __FUNCTION__, DeviceObject, Irp,
+           IoGetCurrentIrpStackLocation(Irp)->MinorFunction);
+
+    PoStartNextPowerIrp(Irp);
```
ThFabba, line 188: Not calling this _immediately_ before `IoCompleteRequest`/`IoSkipCurrentIrpStackLocation` seems odd, but I can't give a concrete reason why it would be. The documentation doesn't seem to specify this in much detail.

drivers/input/inport/pnp.c
```diff
@@ -0,0 +1,283 @@
...
+    if (IsStarted)
+    {
+        KeSynchronizeExecution(DeviceExtension->InterruptObject,
+                               InPortStopMouse,
+                               DeviceExtension);
+
+        /* Flush DPC for ISR */
+        KeFlushQueuedDpcs();
+
+        IoDisconnectInterrupt(DeviceExtension->InterruptObject);
```
ThFabba, line 215: I think this should come before `KeFlushQueuedDpcs`. First we want to guarantee that the ISR is not running (since it could queue a new DPC); _then_ we want to make sure no DPCs are running.

binarymaster (approved):

ThFabba (approved):

PR #3176, opened 2020-09-13
[SHELL32] Improve Format dialog debug log (trivial)

KyleKatarn31:
## Purpose

- While looking at https://jira.reactos.org/browse/CORE-17272 some logging could be improved to better manage such situation

## Proposed changes

- Logging of name for unknown fs
- Fix some trace content as not only HDD could be affected

### DIFF
--- a/dll/win32/shell32/dialogs/drive.cpp
+++ b/dll/win32/shell32/dialogs/drive.cpp
@@ -444,7 +444,7 @@ FormatExCB(
         case FSNOTSUPPORTED:
         case CLUSTERSIZETOOSMALL:
             bSuccess = FALSE;
-            FIXME("\n");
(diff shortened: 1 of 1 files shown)


dll/win32/shell32/dialogs/drive.cpp
```diff
@@ -329,7 +329,7 @@ InsertDefaultClusterSizeForFs(HWND hwndDlg, PFORMAT_DRIVE_CONTEXT pContext)
     }
     else
     {
-        FIXME("unknown fs\n");
+        FIXME("unknown fs : %ls\n", wszBuf);
```
HBelusca, line 332:
```suggestion
        FIXME("Unknown filesystem: %ls\n", wszBuf);
```

HBelusca (commented):
Minor improvement.

PR #3177, opened 2020-09-13
[FORMAT] Fix of output text overlap in console when finished

KyleKatarn31:
## Purpose

- At the end of the format operation, text in console gets overwritten and leftovers from previous line remain displayed.

formatbug

## Proposed changes

- Output the missing linefeed.

formatbugfix

### DIFF
--- a/base/system/format/format.c
+++ b/base/system/format/format.c
@@ -562,6 +562,7 @@ int wmain(int argc, WCHAR *argv[])
     FormatEx(RootDirectory, media, FileSystem, Label, QuickFormat,
              ClusterSize, FormatExCallback);
     if (Error) return -1;
+    ConPuts(StdOut, L"\n");
     ConResPuts(StdOut, STRING_FMT_COMPLETE);
 
     //


HBelusca:
I suppose that the original intent (that caused the overlap) was to "remove" the "100% formatted" line, and replace it by "Formatting terminated".

KyleKatarn31:
> I suppose that the original intent (that caused the overlap) was to "remove" the "100% formatted" line, and replace it by "Formatting terminated".

I prefer to use a newline instead. In particular when format process fails it allows to see at which point the process got interrupted.

HBelusca:
> > I suppose that the original intent (that caused the overlap) was to "remove" the "100% formatted" line, and replace it by "Formatting terminated".
> 
> I prefer to use a newline instead. In particular when format process fails it allows to see at which point the process got interrupted.

Definitively, if an error happens, keeping the percentage on screen is fine. The replacement of the 100% into "Formatting finish" should better be done only when the formatting correctly finishes.

KyleKatarn31:
> > > I suppose that the original intent (that caused the overlap) was to "remove" the "100% formatted" line, and replace it by "Formatting terminated".
> > 
> > 
> > I prefer to use a newline instead. In particular when format process fails it allows to see at which point the process got interrupted.
> 
> Definitively, if an error happens, keeping the percentage on screen is fine. The replacement of the 100% into "Formatting finish" should better be done only when the formatting correctly finishes.

The code is not structured to do that in an easy way.
Would this fix be OK for you or do you prefer the more complex solution ? In this case, i'd close this PR and open a Ticket for somebody else.

HBelusca:
Putting the thing simply in newline at first would be fine.

KyleKatarn31:
OK then it's ready to go.

PR #3179, opened 2020-09-13
[SHELL32][SHLWAPI] Forward shell32.ShellMessageBoxW directly to shlwapi.ShellMessageBoxWrapW.

HBelusca:
And:
[PSDK] Explicily use WINAPIV for the variadic ShellMessageBoxA/W functions.
[INCLUDE/REACTOS] Add ShellMessageBoxWrapW in shlwapi_undoc.h .

## Purpose

JIRA issue: [CORE-17271](https://jira.reactos.org/browse/CORE-17271)
See also PR #3172 by @KyleKatarn31.

## Proposed changes

This makes ShellMessageBoxW use the correct implementation where the
text buffer size is dynamic, instead of having a too small hardcoded
size.

However we cannot straightforwardly implement ShellMessageBoxA around
ShellMessageBoxW, by converting some parameters from ANSI to UNICODE,
because there may be some variadic ANSI strings, associated with '%s'
printf-like formatters inside the format string, that would also need
to be converted; however there is no way for us to find these and perform
the conversion ourselves.
Therefore, we re-implement ShellMessageBoxA by doing a copy-paste ANSI
adaptation of the shlwapi.ShellMessageBoxWrapW function.

Note that, on Vista+ onwards, shlwapi implements both ShellMessageBoxA/W,
and shell32 directly forwards these exports to shlwapi, thus avoiding
these workarounds.

### DIFF
--- a/dll/win32/shell32/precomp.h
+++ b/dll/win32/shell32/precomp.h
@@ -45,6 +45,19 @@
 #undef ShellExecute
 #include <undocshell.h>
 
+/*
+ * For versions < Vista+, redefine ShellMessageBoxW to ShellMessageBoxWrapW
+ * (this is needed to avoid a linker error). On Vista+ onwards, shell32.ShellMessageBoxW
+ * redirects to shlwapi.ShellMessageBoxW so the #define should not be needed.
+ *
+ * However our shell32 is built with _WIN32_WINNT set to 0x600 (Vista+),
+ * yet its exports (especially regarding ShellMessageBoxA/W) are Win2003
+ * compatible. So the #define is still needed, and the check be disabled.
+ */
+// #if (_WIN32_WINNT < 0x0600)
+#define ShellMessageBoxW ShellMessageBoxWrapW
+// #endif
+
 #include <browseui_undoc.h>
 
 #include <shellutils.h>
--- a/dll/win32/shell32/wine/shellord.c
+++ b/dll/win32/shell32/wine/shellord.c
@@ -331,12 +331,28 @@ BOOL WINAPI RegisterShellHook(
  *
  * See ShellMessageBoxA.
  *
+ */
+#ifdef __REACTOS__
+/*
+ * shell32.ShellMessageBoxW directly redirects to shlwapi.ShellMessageBoxWrapW,
+ * while shell32.ShellMessageBoxA is a copy-paste ANSI adaptation of the
+ * shlwapi.ShellMessageBoxWrapW function.
+ *
+ * From Vista+ onwards, all the implementation of ShellMessageBoxA/W that
+ * were existing in shell32 has been completely moved to shlwapi, so that
+ * shell32.ShellMessageBoxA and shell32.ShellMessageBoxW are redirections
+ * to the corresponding shlwapi functions.
+ *
+ */
+#else // !__REACTOS__
+/*
  * NOTE:
  * shlwapi.ShellMessageBoxWrapW is a duplicate of shell32.ShellMessageBoxW
  * because we can't forward to it in the .spec file since it's exported by
  * ordinal. If you change the implementation here please update the code in
  * shlwapi as well.
  */
+// Wine version, broken.
 int ShellMessageBoxW(
 	HINSTANCE hInstance,
 	HWND hWnd,
@@ -358,12 +374,12 @@ int ShellMessageBoxW(
 	    hInstance,hWnd,lpText,lpCaption,uType);
 
 	if (IS_INTRESOURCE(lpCaption))
-	  LoadStringW(hInstance, LOWORD(lpCaption), szTitle, sizeof(szTitle)/sizeof(szTitle[0]));
+	  LoadStringW(hInstance, LOWORD(lpCaption), szTitle, ARRAY_SIZE(szTitle));
 	else
 	  pszTitle = lpCaption;
 
 	if (IS_INTRESOURCE(lpText))
-	  LoadStringW(hInstance, LOWORD(lpText), szText, sizeof(szText)/sizeof(szText[0]));
+	  LoadStringW(hInstance, LOWORD(lpText), szText, ARRAY_SIZE(szText));
 	else
 	  pszText = lpText;
 
@@ -376,6 +392,7 @@ int ShellMessageBoxW(
         LocalFree(pszTemp);
 	return ret;
 }
+#endif
 
 /*************************************************************************
  * ShellMessageBoxA				[SHELL32.183]
@@ -395,6 +412,18 @@ int ShellMessageBoxW(
  * NOTES
  *     Exported by ordinal
  */
+#ifdef __REACTOS__
+/*
+ * Note that we cannot straightforwardly implement ShellMessageBoxA around
+ * ShellMessageBoxW, by converting some parameters from ANSI to UNICODE,
+ * because there may be some variadic ANSI strings, associated with '%s'
+ * printf-like formatters inside the format string, that would also need
+ * to be converted; however there is no way for us to find these and perform
+ * the conversion ourselves.
+ * Therefore, we re-implement ShellMessageBoxA by doing a copy-paste ANSI
+ * adaptation of the shlwapi.ShellMessageBoxWrapW function.
+ */
+#endif
 int ShellMessageBoxA(
 	HINSTANCE hInstance,
 	HWND hWnd,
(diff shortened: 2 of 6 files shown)


dll/win32/shell32/precomp.h
```diff
@@ -45,6 +45,10 @@
 #undef ShellExecute
 #include <undocshell.h>
 
+// #if (_WIN32_WINNT >= 0x0600)
```
HBelusca, line 48: This is commented for the moment, because:
- our shell32 is built with _WIN32_WINNT set to 0x600 (Vista+),
- but our exports (especially regarding ShellMessageBoxA/W) are Win2003-compatible (they change starting Vista+, as explained in the commit log and the code).

dll/win32/shell32/precomp.h
```diff
@@ -45,6 +45,10 @@
...
+#define ShellMessageBoxW ShellMessageBoxWrapW
```
HBelusca, line 49: This is done so as to avoid modifying and replacing all places in the shell32 code where ShellMessageBoxW is used, into ShellMessageBoxWrapW.

dll/win32/shell32/wine/shellord.c
```diff
@@ -358,12 +374,12 @@ int ShellMessageBoxW(
 	    hInstance,hWnd,lpText,lpCaption,uType);
 
 	if (IS_INTRESOURCE(lpCaption))
-	  LoadStringW(hInstance, LOWORD(lpCaption), szTitle, sizeof(szTitle)/sizeof(szTitle[0]));
+	  LoadStringW(hInstance, LOWORD(lpCaption), szTitle, ARRAY_SIZE(szTitle));
```
HBelusca, line 377: These ARRAY_SIZE changes are because I also synchronized with latest Wine (where the code hasn't changed at all except for these).

dll/win32/shlwapi/ordinal.c
```diff
@@ -4823,7 +4837,7 @@ INT WINAPIV ShellMessageBoxWrapW(HINSTANCE hInstance, HWND hWnd, LPCWSTR lpText,
     TRACE("(%p,%p,%p,%p,%08x)\n", hInstance, hWnd, lpText, lpCaption, uType);
 
     if (IS_INTRESOURCE(lpCaption))
-        LoadStringW(hInstance, LOWORD(lpCaption), szTitle, sizeof(szTitle)/sizeof(szTitle[0]));
+        LoadStringW(hInstance, LOWORD(lpCaption), szTitle, ARRAY_SIZE(szTitle));
```
HBelusca, line 4840: See comment about ARRAY_SIZE above.

dll/win32/shell32/wine/shellord.c
```diff
@@ -403,6 +432,58 @@ int ShellMessageBoxA(
...
+    }
+    else
+        pszText = lpText;
+
+    FormatMessageA(FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_STRING,
+                   pszText, 0, 0, (LPSTR)&pszTemp, 0, &args);
+
+    __ms_va_end(args);
+
+    ret = MessageBoxA(hWnd, pszTemp, pszTitle, uType | MB_SETFOREGROUND);
```
HBelusca, line 478: I detected that the message boxes triggered by ShellMessageBoxA/W on Windows have the setforeground flag always set (checked by testing).

dll/win32/shlwapi/ordinal.c
```diff
@@ -4852,7 +4866,11 @@ INT WINAPIV ShellMessageBoxWrapW(HINSTANCE hInstance, HWND hWnd, LPCWSTR lpText,
 
     __ms_va_end(args);
 
+#ifdef __REACTOS__
     ret = MessageBoxW(hWnd, pszTemp, pszTitle, uType);
+#else
+    ret = MessageBoxW(hWnd, pszTemp, pszTitle, uType | MB_SETFOREGROUND);
```
HBelusca, line 4872: See comment above regarding the setforeground flag.

sdk/include/reactos/shlwapi_undoc.h
```diff
@@ -120,6 +120,16 @@ INT WINAPI SHUnicodeToAnsi(LPCWSTR lpSrcStr, LPSTR lpDstStr, INT iLen);
...
+int
+WINAPIV
+ShellMessageBoxWrapW(
+  _In_opt_ HINSTANCE hAppInst,
+  _In_opt_ HWND hWnd,
+  _In_ LPCWSTR lpcText,
+  _In_opt_ LPCWSTR lpcTitle,
+  _In_ UINT fuStyle,
+  ...);
+
```
HBelusca, line 132: Inspired by @learn-more 's private patch, but I've put it in the shlwapi_undoc header because this is a shlwapi-specific thing.

HBelusca (commented):
Some PR comments to explain why I did some stuff like that.

dll/win32/shell32/precomp.h
HBelusca: Note that without this change, there is a linker error where the linker is unable to find `_ShellMessageBoxW`.

dll/win32/shell32/wine/shellord.c
```diff
@@ -403,6 +432,58 @@ int ShellMessageBoxA(
...
+
+    HeapFree(GetProcessHeap(), 0, szText);
```
KyleKatarn31, line 480: if (szText) ....

dll/win32/shlwapi/ordinal.c
```diff
@@ -4852,7 +4866,11 @@ INT WINAPIV ShellMessageBoxWrapW(HINSTANCE hInstance, HWND hWnd, LPCWSTR lpText,
...
+#endif
 
     HeapFree(GetProcessHeap(), 0, szText);
```
KyleKatarn31, line 4875: if (szText) ...

dll/win32/shell32/wine/shellord.c
learn-more: no.

dll/win32/shlwapi/ordinal.c
learn-more: no

dll/win32/shell32/precomp.h
learn-more: So what is the #ifdef doing here?
Even in 2k3 mode shell should use the Wrap function internally?

dll/win32/shell32/wine/shellord.c
```diff
@@ -403,6 +432,58 @@ int ShellMessageBoxA(
...
+        LoadStringA(hInstance, LOWORD(lpCaption), szTitle, ARRAY_SIZE(szTitle));
+    else
+        pszTitle = lpCaption;
+
+    if (IS_INTRESOURCE(lpText))
+    {
+        /* Retrieve the maximum possible length of the corresponding
+         * Unicode string (not counting any possible NULL-terminator). */
+        const WCHAR *ptr;
+        UINT len = LoadStringW(hInstance, LOWORD(lpText), (LPWSTR)&ptr, 0);
```
learn-more, line 456: It's better to use the A here as well.

dll/win32/shlwapi/ordinal.c
learn-more: nit: you can add:
```
#ifdef __REACTOS__
uType |= MB_SETFOREGROUND
#endif
```
and reduce the diff.

learn-more (changes requested):
One thing that should be addressed (calling the W to get the length),
other than that lgtm.

dll/win32/shell32/precomp.h
HBelusca: So yes, in 2k3 mode it implicitly uses the Wrap function internally. But in Vista+, where shlwapi exports named versions of ShellMessageBoxA and W, the shell32 ones directly redirects to these shlwapi exports, and it doesn't use the Wrap version internally anymore.

dll/win32/shell32/wine/shellord.c
HBelusca: It's actually not possible to call the LoadStringA with this set of parameters (both on Windows and ReactOS) in order to just retrieve the size of the string, without using a buffer to also retrieve the string contents (not the pointer).

learn-more: yes, so give it a length of 1 or so, and a dummy buffer so it can return the actual length?
The length in WCHAR is most likely not the same as the length in Ansi..

learn-more: or alternatively, convert the string yourself (and calculate resulting length),
but then you'd have to do this the same way as `LoadStringA` is doing.

HBelusca: Documentation says:

> If the function succeeds, the return value is one of the following:
>
> - **The number of characters copied into the buffer (if cchBufferMax is non-zero), not including the terminating null character.**
>
> - The number of characters in the string resource that lpBuffer points to (if cchBufferMax is zero). The string resource is not guaranteed to be null-terminated in the module's resource table, and you can use this value to determine where the string resource ends.
>
> - Zero if the string resource does not exist.
>
> To get extended error information, call GetLastError.

Our implementation of LoadStringA (which is similar to the one of Windows) is:
https://git.reactos.org/?p=reactos.git;a=blob;f=win32ss/user/user32/misc/resources.c;hb=f3bc1b396aa5471cc4eb820ce4586c0ca5b6d1a8#l77

So first of all it expects a non-zero cchBufferMax (the buflen parameter; note that it doesn't check for the buffer pointer to be NULL); then, when the string is found, it translates it in ANSI into the passed buffer, taking into account its size from the buflen parameter, and returns in retval the size of the copied and converted string part.

KyleKatarn31: Yes, that's the discussion we had in my previous PR. Looks like our implementation of LoadStringA is not as per MSDN

HBelusca: Our implementation has to be on par with what Windows actually does, and not what MSDN may say if it's wrong.

learn-more (approved):

dll/win32/shell32/wine/shellord.c
HBelusca: (For record, there are winetests that check this behaviour:
https://git.reactos.org/?p=reactos.git;a=blob;f=modules/rostests/winetests/user32/resource.c;hb=5077e400d5a92d15516c0dd1c7fa4cdb9399dd12#l125
)

dll/win32/shell32/precomp.h
HBelusca: The #if should actually read `#if (_WIN32_WINNT < 0x0600)`

dll/win32/shell32/wine/shellord.c
HBelusca: I calculate the more correct necessary character number using the WideCharToMultiByte() call below.

PR #3181, opened 2020-09-13
[KERNEL32_VISTA] Cleanup stuff

tkreuzer:
- Move kernel32_vista into a subfolder of kernel32
- Add existing vista.c to kernel32_vista
- Some cleanup

Testman results: https://reactos.org/testman/compare.php?ids=74701,74703,74706,74708


dll/win32/kernel32/kernel32_vista/vista.c
```diff
@@ -860,6 +732,7 @@ GetUserPreferredUILanguages(
 /*
  * @unimplemented
  */
+#if 0 // Tis is Windows 7+
```
SergeGautherie, line 735:
```suggestion
#if 0 // This is Windows 7+
```

SergeGautherie: Also sync
https://git.reactos.org/?p=reactos.git;a=blob;f=sdk/include/psdk/winnls.h;hb=HEAD#l1103

PR #3184, opened 2020-09-14
[KDBG] Fixed handle leak in KdbpSymLoadModuleSymbols() when RosSymCreateFromFile() fails

TheDeadFish:
## Purpose
kdbg leaks file handle when it fails to load the symbols from a module. This fixes that leak

## Proposed changes
Reorder operations so that file handle is always closed

### DIFF
--- a/ntoskrnl/kdbg/kdb_symbols.c
+++ b/ntoskrnl/kdbg/kdb_symbols.c
@@ -363,6 +363,7 @@ KdbpSymLoadModuleSymbols(
     HANDLE FileHandle;
     NTSTATUS Status;
     IO_STATUS_BLOCK IoStatusBlock;
+    BOOLEAN Result;
 
     /* Allow KDB to break on module load */
     KdbModuleLoaded(FileName);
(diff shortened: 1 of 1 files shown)


ThFabba (approved):

ntoskrnl/kdbg/kdb_symbols.c
```diff
@@ -404,14 +405,15 @@ KdbpSymLoadModuleSymbols(
 
     DPRINT("Loading symbols from %wZ...\n", FileName);
 
-    if (!RosSymCreateFromFile(&FileHandle, RosSymInfo))
+    Result = RosSymCreateFromFile(&FileHandle, RosSymInfo);
```
HBelusca, line 408: A possible alternative to save from the usage of this new variable, would be to do something like:
```c
    Status = (RosSymCreateFromFile(&FileHandle, RosSymInfo) ? STATUS_SUCCESS : STATUS_UNSUCCESSFUL);
```
and using `if (!NT_SUCCESS(Status))` (or `if (Status != STATUS_SUCCESS)`.

TheDeadFish: I personally think this is really bad coding style, it obscures the true function of the code. I don't understand why such contortions would be considered more acceptable than an extra variable.

HBelusca: i agree.

PR #3185, opened 2020-09-14
[RAPPS] Fix of unhandled exception when trying to modify/uninstall without selection

KyleKatarn31:
## Purpose

- When tying to modify/uninstall and empty selection, RAPPS crashes due to unhandled exception. This is due to the result of GetFocusedItemData() behind used without being checked non NULL.

image

JIRA issue: [CORE-17279](https://jira.reactos.org/browse/CORE-17279)

## Proposed changes

- Return FALSE and do nothing (as expected) when GetFocusedItemData() is NULL.

image

### DIFF
--- a/base/applications/rapps/gui.cpp
+++ b/base/applications/rapps/gui.cpp
@@ -278,16 +278,20 @@ BOOL CMainWindow::RemoveSelectedAppFromRegistry()
     if (MessageBoxW(szMsgText, szMsgTitle, MB_YESNO | MB_ICONQUESTION) == IDYES)
     {
         CInstalledApplicationInfo *InstalledApp = (CInstalledApplicationInfo *)m_ApplicationView->GetFocusedItemData();
-        LSTATUS Result = InstalledApp->RemoveFromRegistry();
-        if (Result != ERROR_SUCCESS)
+        if (InstalledApp)
         {
-            // TODO: popup a messagebox telling user it fails somehow
-            return FALSE;
-        }
+            LSTATUS Result = InstalledApp->RemoveFromRegistry();
+            if (Result != ERROR_SUCCESS)
+            {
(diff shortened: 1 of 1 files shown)


base/applications/rapps/gui.cpp
```diff
@@ -299,7 +303,7 @@ BOOL CMainWindow::UninstallSelectedApp(BOOL bModify)
         return FALSE;
 
     CInstalledApplicationInfo *InstalledApp = (CInstalledApplicationInfo *)m_ApplicationView->GetFocusedItemData();
-
+    if(!InstalledApp) return FALSE;
```
HBelusca, line 306:
```suggestion
    if (!InstalledApp) return FALSE;
```

base/applications/rapps/gui.cpp
```diff
@@ -278,16 +278,20 @@ BOOL CMainWindow::RemoveSelectedAppFromRegistry()
     if (MessageBoxW(szMsgText, szMsgTitle, MB_YESNO | MB_ICONQUESTION) == IDYES)
     {
         CInstalledApplicationInfo *InstalledApp = (CInstalledApplicationInfo *)m_ApplicationView->GetFocusedItemData();
-        LSTATUS Result = InstalledApp->RemoveFromRegistry();
-        if (Result != ERROR_SUCCESS)
+        if (InstalledApp)
```
HBelusca, line 281: Just handle the `if (!InstalledApp)` case here first; this will allow to not have this indentation increase (and reduce the patch).

KyleKatarn31: Good point !

KyleKatarn31:
@HBelusca : Ready to go ?

learn-more (approved):

PR #3187, opened 2020-09-14
[PSDK][XDK] *.h: *ProbeForRead() does not read memory (anymore)

SergeGautherie:
Addendum to cd22887 (r51126).


ThFabba:
Annotations describe contract, not behavior. Our annotations match those in WDK.

PR #3188, opened 2020-09-14
[NTOS:EX] ProbeFor*(): Log failures

SergeGautherie:
JIRA issue: [CORE-16757](https://jira.reactos.org/browse/CORE-16757)


HBelusca:
No need to cripple the code with dprints for each and every failure. Only error/failure code paths that should ideally never happen may require DPRINTs; otherwise for other code paths like e.g. this one (where the probing is done on user-mode pointers, and therefore it is **expected** that there can be problems) DPRINTs are completely useless.
At best this could be just a local patch of yours if you wanted to do some tests and detect problems (where e.g. the caller of the function passes garbage unintentionally). But this should not be sent upstream.

PR #3189, opened 2020-09-14
[NTOS:PS] NtQueryInformationProcess(): Make use of ProbeForRead() mor…

SergeGautherie:
…e explicit

Addendum to 24834c0 (r72532).


ThFabba:
This is what `git blame` is for.
If you want to do something useful, write a test that shows this.

PR #3197, opened 2020-09-16
[RAPPS] Use `RegDeleteKeyExW` for deleting the apps strings from registry only if it is available in advapi32

oleg-dubinskiy:
## Purpose

Use `RegDeleteKeyW` function instead of `RegDeleteKeyExW` for deleting the aspplications strings from registry, if the 2nd one is not available in advapi32.dll, on Windows versions whose don't have the Ex function (XP SP3 x32, Server 2003 SP0 and earlier). So now, on XP x64/Server 2003 SP1+, it will use Ex function, and on earlier versions - non-Ex one.
It will fix a regression with Rapps startup on these systems, due to that missing function.

JIRA issue: [CORE-17264](https://jira.reactos.org/browse/CORE-17264)

## Proposed changes

- First retrieve the advapi32 handle.
- It it has been found, get the address of `RegDeleteKeyExW` function.
- If the address has been found too, call that function from advapi32.dll with all required parameters.
- Otherwise, if it has not been found, call `RegDeleteKeyW` instead.
- Add/update the comments accordingly.

## Result

Before:
rapps_before

After:
rapps_after

### DIFF
--- a/base/applications/rapps/installed.cpp
+++ b/base/applications/rapps/installed.cpp
@@ -131,16 +131,34 @@ BOOL CInstalledApplicationInfo::UninstallApplication(BOOL bModify)
     return StartProcess(bModify ? szModifyPath : szUninstallString, TRUE);
 }
 
+typedef LSTATUS (WINAPI *RegDeleteKeyExWProc)(HKEY, LPCWSTR, REGSAM, DWORD);
+
 LSTATUS CInstalledApplicationInfo::RemoveFromRegistry()
 {
     ATL::CStringW szFullName = L"Software\\Microsoft\\Windows\\CurrentVersion\\Uninstall\\" + szKeyName;
+    HMODULE hMod = GetModuleHandleW(L"advapi32.dll");
+    RegDeleteKeyExWProc pRegDeleteKeyExW;
 
-    // TODO: if there are subkeys inside, simply RegDeleteKeyExW will fail
+    // TODO: if there are subkeys inside, simply RegDeleteKeyExW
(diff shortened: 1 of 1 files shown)


learn-more (approved):

KyleKatarn31:
Do you confirm that we don't have the same issue in dll/win32/msi/action.c:3009 & 3013 ?

oleg-dubinskiy:
>Do you confirm that we don't have the same issue in dll/win32/msi/action.c:3009 & 3013 ? (Unless it's for the specific purpose of using RAPPS in Windows and not for our MSI module)

I guess it's a (possible) problem just of our msi module instead and for start it needs to test our msi components on XP. If the similar fail due to missing Ex function will happen on XP (e. g. during installing msi installers), then yes, it probably can be fixed by the same way.

KyleKatarn31:
> > Do you confirm that we don't have the same issue in dll/win32/msi/action.c:3009 & 3013 ? (Unless it's for the specific purpose of using RAPPS in Windows and not for our MSI module)
> 
> I guess it's a (possible) problem just of our msi module instead and for start it needs to test our msi components on XP. If the similar fail due to missing Ex function will happen on XP (e. g. during installing msi installers), then yes, it probably can be fixed by the same way.

OK, then i let you see if it deserves a second PR or can be fixed here too.
Thanks !

learn-more:
Please refrain from using github as a discussion platform for completely unrelated code.

KyleKatarn31:
That's what I'm doing, referred code is exactly related to the bug fix of this PR

PR #3199, opened 2020-09-17
[WIN32SS] Repaint the whole window frame on some style bits set

Doug-Lyons:
## Fix regression causing Calculator buttons to disappear when changing number base.

_Code change to recognize Calculator Window Style and handle it like before the regression._

JIRA issue: [CORE-16827](https://jira.reactos.org/browse/CORE-16827)


jimtabor (approved):
Once all the details are worked out. Sometime later, a re rewrite of paint will be done.

The changes look okay,

PR #3200, opened 2020-09-17
[MSCONFIG_NEW] Initialize(): Correctly handle no semaphore case

SergeGautherie:


SergeGautherie:
@HBelusca, Could you review this trivial PR, which has stalled for nearly 3 years? Thanks.

DarkFire01:
Going to close this for now, feel free to reopen if you feel like continuing.

base/applications/msconfig_new/msconfig.c
```diff
@@ -358,25 +358,29 @@ BOOL Initialize(HINSTANCE hInstance)
...
-
-        /*
-         * A semaphore with the same name already exist. It should have been
-         * created by another instance of MSConfig. Try to find its window
-         * and bring it to front.
-         */
-        if ( (hSingleWnd && IsWindow(hSingleWnd))                         ||
-             ( (hSingleWnd = FindWindowW(L"#32770", szAppName)) != NULL ) ||
-             (!bIsPreVistaOSVersion ? ( (hSingleWnd = FindWindowW(L"#32770", lpszVistaAppName)) != NULL ) : FALSE) )
+        if (hSemaphore)
```
HBelusca, line 361: Actually the PR wasn't convincing.
The idea of the original code was this:

1. if the semaphore already existed because of some other MSConfig instance [(CreateSemaphore returned non-zero handle BUT the last error == ERROR_ALREADY_EXISTS as per MSDN docs)](https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-createsemaphorew#return-value), the code tries to find another existing MSConfig window and bring it to front if so;

2. if the semaphore creation (CreateSemaphoreW call) failed, this actually means that something unexpected failed (so the semaphore didn't exist prior, and doesn't exist now, or, it could be that the semaphore already exists but somehow we don't have correct accesses to it: may correspond to an MSConfig opened with Admin access or whatnot), then the idea of the code was STILL to try to find if there's an existing MSConfig window somewhere and bring it upfront : this was here some sort of last-chance attempt.

The PR modifications would defeat the purpose of point 2.

PR #3202, opened 2020-09-17
[NTOS:KE] KeDelayExecutionThread(): Sync annotations

SergeGautherie:


Extravert-ir:
Either sync all the Ke subsystem + SDK, or don't do such changes.
It doesn't make sense to sync _only one_ function

PR #3109, opened 2020-08-30
[GDI32_APITEST] Create StretchBlt Regression Test

Doug-Lyons:
## Add new Regression Unit Test for StretchBlt functions to gdi32_apitest.

_Added new code to gdi32_apitest for testing StretchBlt functions._

JIRA issue: [CORE-16642](https://jira.reactos.org/browse/CORE-16642)

PR #2942, opened 2020-06-22
[WIN32K] Add StretchBlt function ability to flip images

Doug-Lyons:
## Fix StretchBlt function to allow flipped images

_Modify dib\dibxxbpp.c programs to understand flipped images._

JIRA issue: [CORE-16642](https://jira.reactos.org/browse/CORE-16642)
                  and [CORE-16984](https://jira.reactos.org/browse/CORE-16984)
                  and [CORE-13273](https://jira.reactos.org/browse/CORE-13273)
                  and [CORE-14671](https://jira.reactos.org/browse/CORE-14671)

SpinaThyr_Good

Trurlg_Good

Lazarus_PeaZip_OK

PR #3203, opened 2020-09-17
[GDI32_APITEST] Add tests for SetLayout and its effects

peterooch:
## Purpose

As part of the efforts to fix mirroring (Right-to-left UI), this apitest should help determine what is wrong currently in how certain functions behave when the DC is mirrored.

This test focuses mostly on BitBlt, StretchBlt and LPtoDP.

This apitest was inspired by PR #3109

I have figured a solution for the LPtoDP issue (based on how Wine does it)  but I'm currently waiting for PR #2942 (Horizontal flipping is one of the features of mirroring) so I could build upon that.

JIRA issue: [CORE-17298](https://jira.reactos.org/browse/CORE-17298)

Tests results:
Win10 version 2004
results win10
Win2003
results win2003
ReactOS
results ros

### DIFF
--- a/modules/rostests/apitests/gdi32/SetLayout.c
+++ b/modules/rostests/apitests/gdi32/SetLayout.c
@@ -1,0 +1,268 @@
+/*
+ * PROJECT:         ReactOS api tests
+ * LICENSE:         GPL - See COPYING in the top level directory
+ * PURPOSE:         Tests for SetLayout and its effects on other gdi functions
+ *                  such as StretchBlt, BitBlt, LPtoDP, DPtoLP
+ * PROGRAMMERS:     Baruch Rutman
+ *                  Inspired by the StretchBlt test
+ */
+
+#include "precomp.h"
+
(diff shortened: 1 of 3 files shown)


modules/rostests/apitests/gdi32/SetLayout.c
```diff
@@ -0,0 +1,241 @@
...
+}
+
+#define BLACK_PIXEL 0x000000
+#define BLUE_PIXEL  0x0000FF
+#define GREEN_PIXEL 0x00FF00
+#define RED_PIXEL   0xFF0000
+#define WHITE_PIXEL 0xFFFFFF
+
+#if 0
+/* Draw the bitmap as colored letters (if console supports color escape codes) */
```
learn-more, line 28: With the function `SetConsoleTextAttribute` you can change the colors in a windows console :)

peterooch: Good point, did that and now it works in ReactOS as well.
Thanks for the tip.

binarymaster (approved):

PR #3204, opened 2020-09-17
[WIN32SS] Scrollbars - Integer underflow/wraparound

KyleKatarn31:
## Purpose

The Scrollinfo member nPage is an unsigned integer (UINT). 
If nPage is 0, subtracting 1 will give 4294967295. This may cause a scrollbar not to be hidden or disabled.

JIRA issue: [CORE-16491](https://jira.reactos.org/browse/CORE-16491)

## Proposed changes

- explicit cast to int in order to be robust to undeflow/wraparound
Patch proposed by "I_Kill_Bugs" in Jira

### DIFF
--- a/win32ss/user/ntuser/scrollbar.c
+++ b/win32ss/user/ntuser/scrollbar.c
@@ -597,7 +597,8 @@ co_IntSetScrollInfo(PWND Window, INT nBar, LPCSCROLLINFO lpsi, BOOL bRedraw)
    if (lpsi->fMask & (SIF_RANGE | SIF_PAGE | SIF_DISABLENOSCROLL))
    {
       new_flags = Window->pSBInfo->WSBflags;
-      if (Info->nMin >= (int)(Info->nMax - max(Info->nPage - 1, 0)))
+      /* Cast Info->nPage to int to prevent wraparound/underflow. Fixes CORE-16491 */
+      if (Info->nMin >= (int)(Info->nMax - max((int)Info->nPage - 1, 0)))
       {
          /* Hide or disable scroll-bar */
          if (lpsi->fMask & SIF_DISABLENOSCROLL)


KyleKatarn31:
@HBelusca : Is it "pending" because or expect something "more" from me on this PR ?

HBelusca:
Because I wanted to have time to think for myself whether all these casts are actually necessary, if we could rewrite this condition in another way.

KyleKatarn31:
> Because I wanted to have time to think for myself whether all these casts are actually necessary, if we could rewrite this condition in another way.

HBelusca:
You can keep this PR open by the way.

KyleKatarn31:
Sorry, my mistake :-)

win32ss/user/ntuser/scrollbar.c
```diff
@@ -597,7 +597,8 @@ co_IntSetScrollInfo(PWND Window, INT nBar, LPCSCROLLINFO lpsi, BOOL bRedraw)
    if (lpsi->fMask & (SIF_RANGE | SIF_PAGE | SIF_DISABLENOSCROLL))
    {
       new_flags = Window->pSBInfo->WSBflags;
-      if (Info->nMin >= (int)(Info->nMax - max(Info->nPage - 1, 0)))
+      /* Cast Info->nPage to int to prevent wraparound/underflow. Fixes CORE-16491 */
+      if (Info->nMin >= (int)(Info->nMax - max((int)Info->nPage - 1, 0)))
```
HBelusca, line 601:
```suggestion
      if (Info->nMin + (int)max(Info->nPage, 1) > Info->nMax)
```
Explanations:
- `max((int)Info->nPage - 1, 0)` is equivalent to `(int)max(Info->nPage, 1) - 1` : Indeed, in the case the max() originally returned 0, it's because `nPage - 1 <= 0`, or in other words, because `nPage <= 1` . If on the contrary the original max() returned `(int)nPage - 1`, it's because `(int)nPage - 1 > 0`, or in other words, `(int)nPage > 1`. Therefore, the new comparisons are: `nPage <= 1` and we return `0 == 1 - 1`, or `(int)nPage > 1` and we return `(int)nPage - 1`. Hence the new proposed max() syntax.
- The inequality therefore becomes: `if (Info->nMin >= (int)(Info->nMax - (int)max(Info->nPage, 1) + 1))`. Since `nMax` and the max() are now ints, the outer `(int)` cast is now unnecessary. We get: `if (Info->nMin >= Info->nMax - (int)max(Info->nPage, 1) + 1)`.
- One can rewrite this inequality by moving part of the RHS to the LHS: `if (Info->nMin + (int)max(Info->nPage, 1) - 1 >= Info->nMax)`. Let us remove that " - 1" part: One can be convinced from this trivial example ( `(F) == FALSE, (T) == TRUE`), since we are working with integers:
```
etc.
1 - 1 >= 2 (F) ---> 1 > 2 (F)
2 - 1 >= 2 (F) ---> 2 > 2 (F)
3 - 1 >= 2 (T) ---> 3 > 2 (T)
4 - 1 >= 2 (T) ---> 4 > 2 (T)
etc.
```
that removing the "- 1" will change the "greater or equal" to a "strictly greater". Hence the final ansatz: `if (Info->nMin + (int)max(Info->nPage, 1) > Info->nMax)`.
Alternatively this could also be written as: `if (Info->nMin > Info->nMax - (int)max(Info->nPage, 1))`

You can test whether this new suggestion actually solves CORE-16491 and does not trigger any wraparound/underflow/overflow.

KyleKatarn31:
Looks correct. Thanks @HBelusca for the code optimization and justification.

KyleKatarn31:
@binarymaster : please tag "bug fix"

katahiromz:
@ghost cannot respond because the user is dead.

HBelusca (approved):

GeoB99:
Superseded by [`ffea515`](https://github.com/reactos/reactos/commit/ffea5152e607512f09efdb0bd02499e481877f59).

PR #3207, opened 2020-09-18
[IMM32][INCLUDE] Add immtable.h and modify imm32.spec

katahiromz:
## Purpose

Start implementing IMM (Input Method Manager) support.

## Proposed changes

- Add `immtable.h` header into `win32ss/user/user32/misc/`.
- Modify `imm32.spec` file.


katahiromz:
`HIMC` is a pointer.
`HIMCC` is a pointer.
`HWND` is a pointer.

binarymaster (approved):
`immtable.h` is added but not included anywhere, otherwise looks ok.

PR #3208, opened 2020-09-18
Disabled ECDSA in mbedtls.dll

qbancoffee:
After observing that I could not download JDK8 via RAPPS because
authentication via ECDSA_P256 is not implemented, I decided to do
a wireshark capture of the handshaking between ReactOS(via mbedtls)
and cdn.zulu.com(the server), specifically

https://cdn.azul.com/zulu/bin/zulu8.48.0.53-ca-jdk8.0.265-win_i686.msi

The "Client Hello" reports authentication methods that are not yet
implemented and I guess the server attempts to use the latest and
greatest from the list. When an error is generated, the download
stops instead of the server and client trying to negotiate another
method authentication/encryption.

I simply disbabled ECDSA based stuff in mbedtls so that the
unimplemented ciphers are not reported as implemented and the client
and server can negotiate a method that is implemented.

## Purpose

Disabled ECDSA authentication  in mbedtls because it is not implemented in bcrypt.

JIRA issue: [CORE-16741](https://jira.reactos.org/browse/CORE-16741)

## Proposed changes

Temporarily disable ECDSA based ciphers so that they are not reported as available
during "Client Hello" TLS handshaking.

## TODO

- [ ] 
- [ ]


JoachimHenze:
Can we make bcrypt properly use those ciphers instead Thomas? Pleasy-please-please!

binarymaster (changes requested):
Hello,

```
```

```
```

Regards,

qbancoffee:
O.k. I've updated with my real name for this repo and the email is my real email. I amended the commit and forced pushed it.
Here it is so you can compare
Pedro Pena <qbancoffee@yahoo.com>

binarymaster:
image

@qbancoffee thanks, it looks fixed now. But your GitHub profile name is not updated yet, so if one of developers decide to squash & merge the PR, GitHub will use your nickname as author field instead. It's recommended to set full name profile as well.

qbancoffee:
Done

JoseSanchezz:
Wouldn't it be preferable to allow it to fail until the algorithm in question is implemented? Sure, it won't work, but it won't be using a less secure (potentially inadequate) substitute either.

qbancoffee:
I guess it's a matter of whether you want certain links to download or not from rapps.  I think the best thing is for all of the 
ciphers to be implemented. One weird thing I noticed is that some downloads that shouldn't work actually do work,
like all of the sourceforge links. I'm not sure what cipher is used because I can't seem to see it in the Server Hello portion.

I spent a couple of hours testing each cipher until I came up with a list with the ciphers that are actually implemented in bcrypt
and found a better way of only allowing those to be advertised as working ciphers. 

I think I might ammend this pull request and do that instead of actually disabling ECDSA_P256

JoachimHenze:
This PR was superseded by https://github.com/reactos/reactos/pull/4233

PR #3209, opened 2020-09-19
[RAPPS] update help command

kernelbin:
## Purpose

update help command


KyleKatarn31:
Would you please give more details about the change you are pushing here ? (Purpose, reason for change, descrption) ?

PR #3210, opened 2020-09-19
[MORE] Implement 'Q' key for 'Quit'

katahiromz:
## Purpose

This PR will implement the 'Q' key on the command line tool 'MORE'.

JIRA issue: [CORE-4019](https://jira.reactos.org/browse/CORE-4019)

## Proposed changes

- Implement 'Quit' action for keyboard key 'Q'.

BEFORE:
before
It goes next page.

AFTER:
after
It quits.

### DIFF
--- a/base/applications/cmdutils/more/more.c
+++ b/base/applications/cmdutils/more/more.c
@@ -8,6 +8,7 @@
  * PROGRAMMERS:     Paolo Pantaleo
  *                  Timothy Schepens
  *                  Hermes Belusca-Maito (hermes.belusca@sfr.fr)
+ *                  Katayama Hirofumi MZ (katayama.hirofumi.mz@gmail.com)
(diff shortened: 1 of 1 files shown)


base/applications/cmdutils/more/more.c
```diff
@@ -143,6 +145,16 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
         return FALSE;
     }
 
+    /* 'Q': Quit */
```
HBelusca, line 148:
```suggestion
    /* 'Q': Quit */
    // FIXME: Available only when command extensions are enabled.
```

HBelusca (approved):
Please add the FIXME comment in the code, I will handle that later on.
Otherwise, patch is OK :)

PR #3216, opened 2020-09-20
[IPHLPAPI] Fix for hasArp which crashes some iphlpapi tests.

Doug-Lyons:
## Fix hasArp function.

JIRA issue: [CORE-16513](https://jira.reactos.org/browse/CORE-16513)

## Change return value from testing on "&" to testing on "==".

_Fix iphlpapi regressiion tests which tend to cause multiple ASSERTS at present._

The comment is included to help speed the review and should be removed or updated on the commit or afterwards.

### DIFF
--- a/dll/win32/iphlpapi/ifenum_reactos.c
+++ b/dll/win32/iphlpapi/ifenum_reactos.c
@@ -149,8 +149,7 @@ BOOL hasArp( HANDLE tcpFile, TDIEntityID *arp_maybe ) {
                               NULL );
     if( !NT_SUCCESS(status) ) return FALSE;
 
-    /* This was previously treated incorrectly as a single bitmapped flag.
-     * But this is not correct because AT_ARP is 0x280. */
+    /* AT_ARP corresponds to an individual TDI entity type */
     return (type == AT_ARP);
 }
 
--- a/dll/win32/iphlpapi/ifenum_reactos.c
+++ b/dll/win32/iphlpapi/ifenum_reactos.c
@@ -149,7 +149,9 @@ BOOL hasArp( HANDLE tcpFile, TDIEntityID *arp_maybe ) {
                               NULL );
     if( !NT_SUCCESS(status) ) return FALSE;
 
-    return (type & AT_ARP);
+    /* This was previously treated incorrectly as a single bitmapped flag.
+     * But this is not correct because AT_ARP is 0x280. */
+    return (type == AT_ARP);
 }
 
 static NTSTATUS getInterfaceInfoSet( HANDLE tcpFile,


HBelusca:
MSDN:
https://docs.microsoft.com/en-us/windows/win32/api/tcpioctl/ni-tcpioctl-ioctl_tcp_query_information_ex
> These flag values identify the type of the specified entity. Once again, the Context member of the input structure is ignored.
> The possible type-flag values that can be returned are shown in the following table.
> AT_ARP | Entity implements ARP (Address Resolution Protocol).

So why is the existing flag-type comparison not working?

Doug-Lyons:
@HBelusca, The previous code returned a BOOL (type & AT_ARP). So look at some of the other values here:

'''
#define	ER_ENTITY   0x380
#define	IF_ENTITY   0x200
#define	AT_ARP      0x280
'''
You can see that both of the first two entities will also return a BOOL of true even though they are not AT_ARP values.

HBelusca (approved):

dll/win32/iphlpapi/ifenum_reactos.c
```diff
@@ -149,7 +149,9 @@ BOOL hasArp( HANDLE tcpFile, TDIEntityID *arp_maybe ) {
                               NULL );
     if( !NT_SUCCESS(status) ) return FALSE;
 
-    return (type & AT_ARP);
+    /* This was previously treated incorrectly as a single bitmapped flag.
+     * But this is not correct because AT_ARP is 0x280. */
```
SergeGautherie, line 153: Remove the comment.

SergeGautherie (approved):

learn-more (approved):

dll/win32/iphlpapi/ifenum_reactos.c
Doug-Lyons: @SergeGautherie, I have no problem with the comment being removed, but I would prefer that the person who commits this do it. That way others will have the benefit of seeing it during the review cycle. Thanks.

HBelusca, line 153: A suggestion for the comment:
```suggestion
    /* AT_ARP corresponds to an individual TDI entity type */
```

Doug-Lyons: @HBelusca, My comment will not make sense after this commit, but I included it to help explain what was happening in hopes that it would speed the commit. I was thinking that it would be quickly deleted and/or changed and your suggestion certainly look like a good substitute. Again, maybe the committer can implement this for us. Thanks.

PR #3217, opened 2020-09-21
[REACTOS] Fix parameter types on some exports

SergeGautherie:


dll/ntdll/def/ntdll.spec
```diff
@@ -425,15 +425,15 @@
 @ stdcall NtQueryEaFile(long ptr ptr long long ptr long ptr long)
 @ stdcall NtQueryEvent(long long ptr long ptr)
 @ stdcall NtQueryFullAttributesFile(ptr ptr)
-@ stdcall NtQueryInformationAtom(long long ptr long ptr)
+@ stdcall NtQueryInformationAtom(ptr long ptr long ptr)
```
tkreuzer, line 428: This is wrong, RTL_ATOM is a USHORT.

tkreuzer (changes requested):

dll/ntdll/def/ntdll.spec
SergeGautherie: Indeed.
Fixed, plus another case.

GeoB99 (approved):

PR #3211, opened 2020-09-19
[CRYPTUI] Fix attempt to Destroy already destroyed ImageList

KyleKatarn31:
## Purpose

- When closing CertUI, (or sub dialogs) attempts are made to Destroy ImageList that have already been destroyed. This is due to explicit call ImageList_Destroy while these ImageList are also linked to a listview object which also tries to delete these ImageList in LISTVIEW_NCDestroy

JIRA issue: [CORE-17290](https://jira.reactos.org/browse/CORE-17290)

## Proposed changes

- Remove unneeded calls to ImageList_Destroy (on ImageList which are passed to LVM_SETIMAGELIST

PR #3220, opened 2020-09-21
[RAPPS] Fix attempt to delete an already destroyed ImageList

KyleKatarn31:
## Purpose

- When closing RAPPS, attempt is made to destroy ImageList that have already been destroyed. This is due to explicit call ImageList_Destroy while these ImageList are also linked to a CMainToolbar object which also tries to delete these ImageList in TOOLBAR_Destroy (inherited from comctl32 toolbar)

JIRA issue: [CORE-17295](https://jira.reactos.org/browse/CORE-17295)

## Proposed changes

- Remove unneeded calls to ImageList_Destroy

### DIFF
--- a/base/applications/rapps/appview.cpp
+++ b/base/applications/rapps/appview.cpp
@@ -141,8 +141,7 @@ HWND CMainToolbar::Create(HWND hwndParent)
         return FALSE;
     }
 
-    //ImageList_Destroy(SetImageList(hImageList));
-	SetImageList(hImageList);
+    SetImageList(hImageList);
 
     AddButtons(_countof(Buttons), Buttons);
 
--- a/base/applications/rapps/appview.cpp
+++ b/base/applications/rapps/appview.cpp
@@ -141,7 +141,8 @@ HWND CMainToolbar::Create(HWND hwndParent)
         return FALSE;
     }
 
-    ImageList_Destroy(SetImageList(hImageList));
+    //ImageList_Destroy(SetImageList(hImageList));
+	SetImageList(hImageList);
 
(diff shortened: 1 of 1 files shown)


base/applications/rapps/appview.cpp
```diff
@@ -141,7 +141,7 @@ HWND CMainToolbar::Create(HWND hwndParent)
         return FALSE;
     }
 
-    ImageList_Destroy(SetImageList(hImageList));
+    SetImageList(hImageList);
```
learn-more, line 144: This destroys the previous imagelist.

KyleKatarn31: Yes but TOOLBAR_Destroy will later try to destroy the same Imagelist, as set as component Internal Imagelist by SetImageList (inherited from CToolbar).
Same problem as for https://github.com/reactos/reactos/pull/3211 (where the issue came from the use of a ListView)

learn-more: Please explain how TOOLBAR_Destroy will delete the imagelist that is being leaked here.
Does it keep a list of previously returned imagelists?

KyleKatarn31: - hImageList get referenced as himlInt in the Toolbar component when SetImageList is called
- then destroyed L144 by ImageList_Destroy
- and himlInt gets deleted again when TOOLBAR_Destroy is called (when component is destroyed at RAPPS closure)

Step 2 is a bug.

KyleKatarn31:
Close as I can't reproduce with 0.4.15-dev-989-gb4e07e

PR #3223, opened 2020-09-22
[NTDLL_APITEST] RtlBitmapApi: Fix OS version checks, and more

SergeGautherie:
Follow-up to b217d8b (#3221): additional fixes to 465745b, plus a couple related improvements to be more explicit.
Especially, fix expected results on NT6.0 and partially on NT5.1.

[WTB 79117](https://testbot.winehq.org/JobDetails.pl?Key=79117)

### DIFF
--- a/modules/rostests/apitests/ntdll/RtlBitmap.c
+++ b/modules/rostests/apitests/ntdll/RtlBitmap.c
@@ -597,8 +597,7 @@ Test_RtlFindLongestRunClear(void)
 {
 }
 
-
-START_TEST(RtlBitmap)
+START_TEST(RtlBitmapApi)
 {
     /* Windows 2003 has broken bitmap code that modifies the buffer */
     if (!IsWindows7OrGreater() && !IsReactOS())
@@ -627,4 +626,3 @@ START_TEST(RtlBitmap)
     Test_RtlFindClearRuns();
     Test_RtlFindLongestRunClear();
 }
-
--- a/modules/rostests/apitests/ntdll/RtlBitmap.c
+++ b/modules/rostests/apitests/ntdll/RtlBitmap.c
@@ -600,7 +600,7 @@ Test_RtlFindLongestRunClear(void)
 START_TEST(RtlBitmapApi)
 {
     /* Windows 2003 has broken bitmap code that modifies the buffer */
-    if (!IsWindows7OrGreater() && !IsReactOS())
+    if (!IsWindowsVistaOrGreater() && !IsReactOS())
     {
         IsBroken = TRUE;
     }
--- a/modules/rostests/apitests/ntdll/RtlBitmap.c
+++ b/modules/rostests/apitests/ntdll/RtlBitmap.c
@@ -580,21 +580,29 @@ Test_RtlFindNextForwardRunClear(void)
 void
 Test_RtlFindFirstRunClear(void)
 {
+  todo_if(TRUE)
+    ok(FALSE, "%s() is UNIMPLEMENTED\n", __FUNCTION__);
 }
 
 void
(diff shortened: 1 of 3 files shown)


SergeGautherie:
Cc @tkreuzer

tkreuzer:
Please provide a proper description of what this is trying to fix, instead of throwing a link to wine test results at us.

SergeGautherie:
> proper description

Added. Is that what you expected?

sdk/include/psdk/versionhelpers.h
```diff
@@ -43,93 +43,111 @@ IsWindowsVersionOrGreater(WORD wMajorVersion, WORD wMinorVersion, WORD wServiceP
...
 VERSIONHELPERAPI
-IsWindowsXPSP3OrGreater()
+IsWindowsXPSP3OrGreater(void)
 {
     return IsWindowsVersionOrGreater(5, 1, 3);
 }
 
 VERSIONHELPERAPI
-IsWindowsVistaOrGreater()
+IsWindowsServer2003OrGreater(void)
```
ThFabba, line 70: This is a PSDK header. MS's PSDK does not have this, so neither should ours.

SergeGautherie: Right. What would you suggest?
* Add `#ifdef __REACTOS__`, as `IsReactOS()` has.
* "Duplicate" needed code to somewhere else.
* Use "XP+ && Server && !Vista+" instead, which should probably be good enough.
* ...

modules/rostests/apitests/ntdll/RtlBitmap.c
```diff
@@ -580,30 +580,43 @@ Test_RtlFindNextForwardRunClear(void)
 void
 Test_RtlFindFirstRunClear(void)
 {
+  todo_if(TRUE)
+    ok(FALSE, "%s() is UNIMPLEMENTED\n", __FUNCTION__);
```
GeoB99, line 584: `trace` can also work for this.

sdk/include/psdk/versionhelpers.h
```diff
@@ -43,93 +43,111 @@ IsWindowsVersionOrGreater(WORD wMajorVersion, WORD wMinorVersion, WORD wServiceP
 }
 
 VERSIONHELPERAPI
-IsWindowsXPOrGreater()
+IsWindowsXPOrGreater(void)
```
GeoB99, line 46:
```suggestion
IsWindowsXPOrGreater(VOID)
```
Same as for other prototype calls.

sdk/include/psdk/versionhelpers.h
GeoB99: Wrapping code with `__REACTOS__`, in my humble opinion, could work for this case. That is, if everyone agrees with that of course.

modules/rostests/apitests/ntdll/RtlBitmap.c
```diff
@@ -580,30 +580,43 @@ Test_RtlFindNextForwardRunClear(void)
...
 void
 Test_RtlFindLongestRunClear(void)
 {
+  todo_if(TRUE)
+    ok(FALSE, "%s() is UNIMPLEMENTED\n", __FUNCTION__);
 }
 
-
-START_TEST(RtlBitmap)
+START_TEST(RtlBitmapApi)
```
GeoB99, line 608: As you're changing the test name entry point code, as a general rule, the file name must have the same name too.

learn-more: why is this renamed?

GeoB99:
Closing this since the author hasn't provided any further input on this PR.

PR #3225, opened 2020-09-22
[PORTCLS] fixed audio stutter with rosapps AC97 driver

TheDeadFish:
## Purpose
Fix stutter that occurs with waveout mode in winamp using official ac97 driver

## Proposed changes
replace call to AllocatedBufferSize(), with BufferSize()


JoachimHenze:
I do confirm that this fix did fix for me the Winamp 5.666 AC97 playback on VBox4.3.12 with the #define USE_MMIXER_LIB being defined in wdmaud.c on 0.4.15-dev-1045-gd8b5c00. The stutter did happen without this patch when Winamps WaveOut Output plugin was in use. I tested that scenario with our AC97 driver from rapps "Avance Logic 5.10.0.3610", not with the one we do intend to include into the source tree.
Another very nice patch from theDeadFish! Thank you!

PR #3227, opened 2020-09-22
[comctl32] Wrong return value for ImageList_Remove

KyleKatarn31:
## Purpose

- Comctl32 ImageList_Remove "ReactOS hack" returns "-1" while expected retun type is BOOL

## Proposed changes

- Replace "-1" by FALSE
No need to open a WineHQ ticket as the change is only in ReactOS section (so called "The big bad reactos image list hack!")

### DIFF
--- a/dll/win32/comctl32/imagelist.c
+++ b/dll/win32/comctl32/imagelist.c
@@ -4196,7 +4196,7 @@ ImageList_Remove (HIMAGELIST himl, INT i)
 {
     IImageList2* piml = IImageList2_from_impl(himl);
     if (!piml)
-        return -1;
+        return FALSE;
 
     return (piml->lpVtbl->Remove(piml, i) == S_OK) ? TRUE : FALSE;
 }


SergeGautherie (changes requested):
Well found!

Could you add a test for this case?
modules/rostests/winetests/comctl32/imagelist.c

KyleKatarn31:
> Well found!
> 
> Could you add a test for this case?
> modules/rostests/winetests/comctl32/imagelist.c

Thanks 👍 
Yes, in a separate PR because I want to do a global screening on ImageList tests

katahiromz:
@ghost cannot respond because the user is dead.

SergeGautherie (approved):

SergeGautherie:
Additional tests submitted to Wine.
[WTB 85239](https://testbot.winehq.org/JobDetails.pl?Key=85239)

GeoB99:
Superseded by [`1cf1892`](https://github.com/reactos/reactos/commit/1cf1892b14b53a586b45e6b7da81fa9ec57977e9).

PR #3228, opened 2020-09-22
[OPENGL32_APITEST] Fix incorrectly failing test

BenNottelling:
```Test failed: Expected version 1.1.0, got "1.1".```
This is the same, and not something that a test should fail for. 
https://reactos.org/testman/detail.php?id=50067026&prev=0

### DIFF
--- a/modules/rostests/apitests/opengl32/sw_extensions.c
+++ b/modules/rostests/apitests/opengl32/sw_extensions.c
@@ -61,7 +61,7 @@ START_TEST(sw_extensions)
 
     /* Get the version */
     output = (const char*)glGetString(GL_VERSION);
-    ok(strcmp(output, "1.1.0") == 0, "Expected version 1.1.0, got \"%s\".\n", output);
+    ok(strcmp(output, "1.1") == 0, "Expected version 1.1, got \"%s\".\n", output);
 
     /* Get the extensions list */
     output = (const char*)glGetString(GL_EXTENSIONS);

### CODE REVIEW

**Comment by SergeGautherie** [comment]
Hum, why fix the test rather than ROS?
Does your change even succeed on Windows?

File: modules/rostests/apitests/opengl32/sw_extensions.c [hunk 1]
```diff
@@ -2,7 +2,7 @@
  * PROJECT:         ReactOS api tests
  * LICENSE:         BSD - See COPYING.ARM in the top level directory
  * PURPOSE:         Tests extensions exposed by the software implementation
- * PROGRAMMERS:     J�r�me Gardou
+ * PROGRAMMERS:     Jérôme Gardou
```
binarymaster, line 5: Unnecessary change, probably encoding.

BenNottelling:
> Hum, why fix the test rather than ROS?
> Does your change even succeed on Windows?

ReactOS isn't broken, Mesa3D just reports the version as 1.1 instead of 1.1.0
https://github.com/reactos/reactos/blob/c3e4b5fec6ff51e04e95c3c6908b1b99aee49023/dll/opengl/mesa/misc.c#L316

So I guess another solution would be to change that version to 1.1.0, or modify the check in this test to allow anything 1.1.X, but it seems to me this test want's only 1.1.0 so this seemed more appropriate 

But again, nothing is actually broken. The reported / checked version is effectively the same, just that one expects a ```0``` at the end and the other doesn't provide.

modules/rostests/apitests/opengl32/sw_extensions.c
BenNottelling: Yup my bad, this is resolved now.

KyleKatarn31:
If the problem is in Mesa, the fix should be there too...

GeoB99:
The fix is so incorrect, as per the screenshot:
Capture

It'd be better if you actually fixed the issue on the ROS end instead, as this is not a legitimate fix by any means. Considering this PR is stale, I'm going to close it for this and the other reason. If you plan to provide a proper patch with fix, I'll open the PR back.

BenNottelling:
> The fix is so incorrect, as per the screenshot:
> Capture
> 
> It'd be better if you actually fixed the issue on the ROS end instead, as this is not a legitimate fix by any means. Considering this PR is stale, I'm going to close it for this and the other reason. If you plan to provide a proper patch with fix, I'll open the PR back.

Ahh, ok. I see why these were different to begin with. Test checks for ```1.1.0``` as that's what windows reports, but our inbuilt mesa reports ```1.1```. As such, it "works" on windows, but not ReactOS.
Really, is broken. Idk what would be preferable, changing mesa to report ```1.1.0``` or modifying the check so both ```1.1.0``` and ```1.1``` are both considered a valid return. My apologies for not checking what windows returns by default

SergeGautherie:
> what would be preferable

Update Mesa to match Windows, unless there is a known reason to use a different value.

PR #3231, opened 2020-09-23
[BROWSEUI][BROWSEUI_APITEST] SHExplorerParseCmdLine(): Sync and improve

SergeGautherie:
NB:
Major drop in executed checks, simply as `Padding` is commented out.

JIRA issue: [ROSTESTS-302](https://jira.reactos.org/browse/ROSTESTS-302)

Cc @katahiromz

### DIFF
--- a/dll/win32/browseui/browseui.spec
+++ b/dll/win32/browseui/browseui.spec
@@ -4,7 +4,7 @@
 @ stdcall -private DllCanUnloadNow()
 105 stdcall -noname SHCreateSavedWindows()
 106 stdcall -noname SHCreateFromDesktop(ptr)
-107 stdcall -noname SHExplorerParseCmdLine(wstr)
+107 stdcall -noname SHExplorerParseCmdLine(ptr)
 @ stdcall -private DllGetClassObject(ptr ptr ptr)
 @ stdcall -private DllGetVersion(ptr)
 @ stdcall -private DllInstall(long wstr)
--- a/dll/win32/browseui/browseui.spec
+++ b/dll/win32/browseui/browseui.spec
@@ -3,7 +3,7 @@
 103 stdcall -noname SHOpenNewFrame(ptr ptr long long)
 @ stdcall -private DllCanUnloadNow()
 105 stdcall -noname SHCreateSavedWindows()
-106 stdcall -noname SHCreateFromDesktop(long)
+106 stdcall -noname SHCreateFromDesktop(ptr)
 107 stdcall -noname SHExplorerParseCmdLine(wstr)
 @ stdcall -private DllGetClassObject(ptr ptr ptr)
 @ stdcall -private DllGetVersion(ptr)
--- a/modules/rostests/apitests/browseui/SHExplorerParseCmdLine.c
+++ b/modules/rostests/apitests/browseui/SHExplorerParseCmdLine.c
@@ -201,19 +201,26 @@ _Out_opt_ PUINT PWriteEnd)
 
     for (i = 0; i < sizeof(Info) / sizeof(DWORD); i++)
     {
-        switch (i*4)
+        switch (i * sizeof(DWORD))
         {
-        case 0x00: // strPath
-        case 0x04: // pidlPath
-        case 0x08: // dwFlags
-        case 0x20: // pidlRoot
-        case 0x34: // guidInproc (1/4)
-        case 0x38: // guidInproc (2/4)
-        case 0x3C: // guidInproc (3/4)
-        case 0x40: // guidInproc (4/4)
-            break;
-        default:
-            ok(InfoWords[i] == 0x55555555, "Line %lu: Word 0x%02lx has been set to 0x%08lx\n", TestLine, i * 4, InfoWords[i]);
+            case FIELD_OFFSET(EXPLORER_CMDLINE_PARSE_RESULTS, strPath):
+            case FIELD_OFFSET(EXPLORER_CMDLINE_PARSE_RESULTS, pidlPath):
+            case FIELD_OFFSET(EXPLORER_CMDLINE_PARSE_RESULTS, dwFlags):
+            // TODO: 'case FIELD_OFFSET(EXPLORER_CMDLINE_PARSE_RESULTS, nCmdShow):'?
+            case FIELD_OFFSET(EXPLORER_CMDLINE_PARSE_RESULTS, pidlRoot):
+            // TODO: 'case FIELD_OFFSET(EXPLORER_CMDLINE_PARSE_RESULTS, clsid):'?
+            case FIELD_OFFSET(EXPLORER_CMDLINE_PARSE_RESULTS, guidInproc) + (0 * sizeof(DWORD)):
+            case FIELD_OFFSET(EXPLORER_CMDLINE_PARSE_RESULTS, guidInproc) + (1 * sizeof(DWORD)):
+            case FIELD_OFFSET(EXPLORER_CMDLINE_PARSE_RESULTS, guidInproc) + (2 * sizeof(DWORD)):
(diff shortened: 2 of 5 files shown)


modules/rostests/apitests/browseui/SHExplorerParseCmdLine.c
```diff
@@ -22,21 +23,7 @@
...
-    PIDLIST_ABSOLUTE pidl;
-    DWORD dwFlags;
-    ULONG Unknown1[5];
-    PIDLIST_ABSOLUTE pidlRoot;
-    CLSID clsid;
-    GUID guidInproc;
-    ULONG Padding[PADDING_SIZE];
-} EXPLORER_INFO, *PEXPLORER_INFO;
-
-UINT_PTR (WINAPI *SHExplorerParseCmdLine)(_Out_ PEXPLORER_INFO Info);
```
learn-more, line 39: Why would you remove this?
It is useful that the test has a 'known good' struct, so it can be compared against the header one?
Otherwise the test depends on the header / struct it is testing...

SergeGautherie: > Why would you remove this?

Is it not what Thomas and Giannis suggested on Jira?

> compared against the header one

Well, the test file is marked as testing the function (only), not investigating the struct (too).
My point of view would be the opposite: do test (with) the actual struct, not a (different) local one...

HBelusca: yes (sure) why (not)

learn-more: is (this) struct (documented) somewhere?
Otherwise (we) lose (a) major (source) of (documentation),
and (in) a (few) years (people) will ask how we came by this struct.

SergeGautherie: > is (this) struct (documented) somewhere?

Not that I know of.

> Otherwise (we) lose (a) major (source) of (documentation),

Which documentation?
Sorry, I don't understand how having two different structs helps anything.

> and (in) a (few) years (people) will ask how we came by this struct.

Is it not already the case, somehow?
What is your proposal, then?

dll/win32/browseui/browseui.spec
```diff
@@ -3,8 +3,8 @@
 103 stdcall -noname SHOpenNewFrame(ptr ptr long long)
 @ stdcall -private DllCanUnloadNow()
 105 stdcall -noname SHCreateSavedWindows()
-106 stdcall -noname SHCreateFromDesktop(long)
-107 stdcall -noname SHExplorerParseCmdLine(wstr)
+106 stdcall -noname SHCreateFromDesktop(ptr)
```
katahiromz, line 6: Yes, the parameter is pointer.

dll/win32/browseui/browseui.spec
```diff
@@ -3,8 +3,8 @@
...
+107 stdcall -noname SHExplorerParseCmdLine(ptr)
```
katahiromz, line 7: Yes, the parameter is pointer.

modules/rostests/apitests/browseui/SHExplorerParseCmdLine.c
```diff
@@ -191,34 +188,40 @@ _Out_opt_ PUINT PWriteEnd)
...
-        case 0x08: // dwFlags
-        case 0x20: // pidlRoot
-        case 0x34: // guidInproc (1/4)
-        case 0x38: // guidInproc (2/4)
-        case 0x3C: // guidInproc (3/4)
-        case 0x40: // guidInproc (4/4)
-            break;
-        default:
-            ok(InfoWords[i] == 0x55555555, "Line %lu: Word 0x%02lx has been set to 0x%08lx\n", TestLine, i * 4, InfoWords[i]);
+            case FIELD_OFFSET(EXPLORER_CMDLINE_PARSE_RESULTS, strPath):
```
katahiromz, line 207: Nice code.

modules/rostests/apitests/browseui/SHExplorerParseCmdLine.c
```diff
@@ -433,10 +436,10 @@ START_TEST(SHExplorerParseCmdLine)
     WCHAR winDir[MAX_PATH];
 
     HMODULE browseui = LoadLibraryA("browseui.dll");
-    SHExplorerParseCmdLine = (UINT_PTR (__stdcall *)(PEXPLORER_INFO))GetProcAddress(browseui, MAKEINTRESOURCEA(107));
-    if (!SHExplorerParseCmdLine)
+    pSHExplorerParseCmdLine = (UINT_PTR (WINAPI *)(PEXPLORER_CMDLINE_PARSE_RESULTS))GetProcAddress(browseui, MAKEINTRESOURCEA(107));
```
katahiromz, line 439: Why didn't use function typedef?

katahiromz (commented):
Did you test this PR on WinXP or Win2k3? Any screenshots?

modules/rostests/apitests/browseui/SHExplorerParseCmdLine.c
SergeGautherie: Added typedef.

SergeGautherie:
> Did you test this PR on WinXP or Win2k3? Any screenshots?

Yes: [WTB 85155](https://testbot.winehq.org/JobDetails.pl?Key=85155)

HBelusca:
> > Did you test this PR on WinXP or Win2k3? Any screenshots?
> 
> Yes: [WTB 85155](https://testbot.winehq.org/JobDetails.pl?Key=85155)

From WIneHQ for almost all the bots of interest (all WIndows <= 2003):
> No WineTest results are available to detect new errors
> Timeout

That's a bit weak for testing...

SergeGautherie:
> > No WineTest results are available to detect new errors
> > Timeout
> 
> That's a bit weak for testing...

As far as I know, ReactOS project offers no testing facility at all, so even "weak" is at least not so bad!
These screenshots are the "best" solution I came up with.
What better solution do you suggest?

HBelusca:
> What better solution do you suggest?

Try to see whether by submitting a trivially simple test to them, they don't timeout, then retry submitting the browseui test. If it times out again, and in a systematical manner, try it yourself on Windows (in a virtual machine for example....) to see whether the timeout is reproducible again there (something hangs? or takes "too much time" ? ). If so, that would mean there's a problem with the test itself.

SergeGautherie:
> don't timeout

This timeout is unrelated to the test: my script does an explicit ending `pause`.
As this very test needs its executable only, I agree using my packaging script, out of habit, is overkill/unwanted.

[WTB 85158](https://testbot.winehq.org/JobDetails.pl?Key=85158)

katahiromz (approved):

modules/rostests/apitests/browseui/SHExplorerParseCmdLine.c
```diff
@@ -22,21 +23,8 @@
...
-    DWORD dwFlags;
-    ULONG Unknown1[5];
-    PIDLIST_ABSOLUTE pidlRoot;
-    CLSID clsid;
-    GUID guidInproc;
-    ULONG Padding[PADDING_SIZE];
-} EXPLORER_INFO, *PEXPLORER_INFO;
-
-UINT_PTR (WINAPI *SHExplorerParseCmdLine)(_Out_ PEXPLORER_INFO Info);
+typedef UINT_PTR (WINAPI * SHExplorerParseCmdLine_Type)(PEXPLORER_CMDLINE_PARSE_RESULTS);
```
HBelusca, line 26:
```suggestion
typedef UINT_PTR (WINAPI *SHExplorerParseCmdLine_Type)(PEXPLORER_CMDLINE_PARSE_RESULTS);
```

HBelusca (approved):

PR #3234, opened 2020-09-24
[KMDF] Add Kernel-Mode Driver Framework to SDK

Extravert-ir:
This is needed to build a "newer" cdrom_new driver :)

I've taken v1.17 because it is almost identical to v1.15 (which is the earliest version open-sourced)
Not all files and functions are included, some stuff is stubbed due to the fact ReactOS is not Win10-compatible :)
KMDF can only be statically linked to drivers for now, proper dynamic linking needs a helper wdfldr.sys driver.

NB: if you want to review my changes, look at commits a3a83e9f91a643d720ecc2e389747d8246998806 and 3cf714a0669c96a41457c66a3f3b47d1b32e94a4


HBelusca:
Having the choice of statically link a driver with the KMDF (in lib mode) would allow to then use that driver on a Windows NT OS without having also to install the kmdf wdfXXX.sys driver.

Extravert-ir:
> Having the choice of statically link a driver with the KMDF (in lib mode) would allow to then use that driver on a Windows NT OS without having also to install the kmdf wdfXXX.sys driver.

Yeah, but it increases a driver binary size for about 650kb :)
But I agree with you, it would be possible

HBelusca:
A file seems to be missing:
`/home/runner/work/reactos/reactos/src/sdk/lib/wdf/shared/core/coreprivshared.hpp:80:10: fatal error: FxIrpQueue.hpp: No such file or directory`

Extravert-ir:
Files are there, it's a string case issue. The real file name is `fxirpqueue.hpp`

Extravert-ir:
How do you think, should the framework be left in sdk/lib or it's better to move it to sdk/lib/drivers?

HBelusca:
The framework seems to be needed for uniquely writing drivers (be they kernel-mode, or, if we also had the corresponding UMDF thing, drivers in user-mode). So I think lib/drivers would be a better place. Things in lib/ (not lib/drivers/) seems to be usable in other parts of the code, not only in drivers.

Extravert-ir:
I haven't put it into `sdk/lib/drivers` because of high nesting level...
BTW, `wdfldr.sys` will be in the same folder eventually

ColinFinck:
I also think, as WDF is a drivers-only thing, it should therefore go into `sdk/lib/drivers`.
This would also simplify a possible future separation of kernel and user-mode components.

PR #3236, opened 2020-09-26
[NTOS] Remove an old hack

tkreuzer:
This removes an old hack, that isn't needed anymore.

### DIFF
--- a/ntoskrnl/io/iomgr/driver.c
+++ b/ntoskrnl/io/iomgr/driver.c
@@ -1133,9 +1133,6 @@ IopInitializeBootDrivers(VOID)
      * See PiQueueDeviceAction function
      */
     PnPBootDriversLoaded = TRUE;
-
-    /* In old ROS, the loader list became empty after this point. Simulate. */
-    InitializeListHead(&KeLoaderBlock->LoadOrderListHead);
 }
 
 INIT_FUNCTION


learn-more (approved):

erkinalp (approved):

ThFabba (approved):

binarymaster (approved):

HBelusca (approved):

erkinalp:
Why was it needed in the first place?

HBelusca:
> Why was it needed in the first place?

How difficult is it to look at the git history (using e.g. the "blame" functionality) so as to find which commit introduced it: https://github.com/reactos/reactos/commit/380f89c205d72db68dfdfa93ac61b19029879fa8 , and why?

erkinalp:
I looked at that commit and no obvious reason. Just that it happens after the last driver loaded.

HBelusca:
> I have looked at that commit and seen no obvious reason. Just that it happens after the last driver loaded.

The logic deduction is that there was some code later that was looking at that list again and supposed that if it was not empty, there were drivers again that needed to be loaded (but this was not the case), and thus, to not introduce a bug in that old code (that was obviously not fixed by that commit), the small workaround was to "reinitialize" the list.

erkinalp:
By doing that, @ionescu007 went further than Microsoft does in terms of backwards compatibility. You are marvellous in creating a binary-compatible Windows alternative.

HBelusca:
I'm not sure you understood my point @erkinalp , but OK....

Extravert-ir (approved):

ionescu007:
Nice :)

PR #3237, opened 2020-09-26
[NTOS:MM] Fix paged pool initialization on x64

tkreuzer:
Fix paged pool initialization on x64

### DIFF
--- a/ntoskrnl/mm/ARM3/mminit.c
+++ b/ntoskrnl/mm/ARM3/mminit.c
@@ -1755,7 +1755,7 @@ MiBuildPagedPool(VOID)
     MMPDE TempPde = ValidKernelPde;
     PFN_NUMBER PageFrameIndex;
     KIRQL OldIrql;
-    SIZE_T Size;
+    SIZE_T Size, NumberOfPages, NumberOfPdes;
     ULONG BitMapSize;
 #if (_MI_PAGING_LEVELS >= 3)
     MMPPE TempPpe = ValidKernelPpe;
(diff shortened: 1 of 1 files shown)


ntoskrnl/mm/ARM3/mminit.c
```diff
@@ -1819,12 +1819,12 @@ MiBuildPagedPool(VOID)
...
     // Now check how many PTEs will be required for these many pages.
     //
-    Size = (Size + (1024 - 1)) / 1024;
+    Size = (Size + (PTE_PER_PAGE - 1)) / PTE_PER_PAGE;
 
     //
     // Recompute the page-aligned size of the paged pool, in bytes and pages.
     //
-    MmSizeOfPagedPoolInBytes = Size * PAGE_SIZE * 1024;
+    MmSizeOfPagedPoolInBytes = Size * PAGE_SIZE * PTE_PER_PAGE;
```
ThFabba, line 1827: This seems strange. Why does "size of paged pool in bytes" depend on how many PTEs fit in a page?

ThFabba (approved):

ntoskrnl/mm/ARM3/mminit.c
tkreuzer: The comments are misleading. Obviously this was written by Alex. And obviously he had no idea, what he was doing 😄 It's not calculating "how many PTEs will be required for these many pages", but how many PDEs are required. And somehow the logic is that paged pool is allocated at PDE granularity (there is an ASSERT below). Don't ask me why, ask Alex, maybe he knows why, but probably not, and probably newer Windows versions don't do this crazy stuff. Long story short: "It's what Windows does". I'll fix the comments.

- [ ] Fix comments.

tkreuzer: I fixed the comments as well as introduce additional variables other than "Size" to make stuff a little bit more readable,

ThFabba (approved):
Thanks! Much more obvious.

ntoskrnl/mm/ARM3/mminit.c
```diff
@@ -1814,17 +1814,17 @@ MiBuildPagedPool(VOID)
...
     //
-    Size = (Size + (1024 - 1)) / 1024;
+    NumberOfPdes = (NumberOfPages + (PTE_PER_PAGE - 1)) / PTE_PER_PAGE;
 
     //
-    // Recompute the page-aligned size of the paged pool, in bytes and pages.
+    // Recompute the PDE-aligned size of the paged pool, in bytes and pages.
     //
-    MmSizeOfPagedPoolInBytes = Size * PAGE_SIZE * 1024;
+    MmSizeOfPagedPoolInBytes = NumberOfPdes * PTE_PER_PAGE * PAGE_SIZE;
```
learn-more, line 1827: Using both Pde and PTE here (note: I have no clue if that is a bug or not, it just seems weird)

tkreuzer: Each PDE refers to one page of PTEs. Multiplying number of PDEs with PTEs per page equals total number of PTEs, i.e. total number of pages.

PR #3239, opened 2020-09-26
Appcompat dll export magic

tkreuzer:
## Purpose

This PR implement a system that allows to make DLL exports version specific at runtime, based on the applications compatibility setting. It still allows ReactOS DLLs to statically import functions from higher versions, even if they are hidden to the application.

JIRA issue: [CORE-11288](https://jira.reactos.org/browse/CORE-11288)

## Proposed changes

- Add a feature to spec2def to generate a RosCompat descriptor (contained in a special section) pointing to a mask-array, that specifies for each named export on which Windows versions it is present.
- Add a feature to the loader in NTDLL that checks for a RosCompat descriptor, and if present, uses it to patch the export table, splitting it into a public and private part.
- Modify the loader to allow ReactOS DLLs (those that contain a RosCompat descriptor, even if it is empty) to import from private/hidden exports.

## TODO
- [ ] Run on testbot


dll/apisets/api-ms-win-crt-private-l1-1-0.spec
```diff
@@ -40,48 +41,50 @@
...
 @ stub __dcrt_initial_narrow_environment
-@ stub __intrinsic_abnormal_termination
+@ stub -arch=i386 __intrinsic_abnormal_termination
 @ stub __intrinsic_setjmp
 @ stub __intrinsic_setjmpex
 @ stub __processing_throw
 @ stub __report_gsfailure
 @ stub __std_exception_copy
 @ stub __std_exception_destroy
+@ stub -arch=x86_65 __std_terminate
```
Extravert-ir, line 51:
```suggestion
@ stub -arch=x86_64 __std_terminate
```

wjk:
My 2¢ on the problem listed in the OP: How often do applications legitimately (i.e., not testing against or depending on undocumented or semi-documented behavior) load a DLL using `MapViewOfFile()`? If it isn't that common a use-case, I would suggest option 1 until option 4 can be implemented (possibly on top of the SxS redirection system).

Also, what do you mean by “if an app looks up the entry manually”? If that means parsing the PE file format by hand in-memory to find, relocate, and call export addresses, this looks to me like an even rarer use-case (except maybe for heavily obfuscated software e.g. malware). The feasibility of this approach depends on how likely it is that normal, correctly written applications engage in these two behaviors, and you would have more knowledge of this than I. Hope this helps!

andrew-boyarshin:
I propose taking the current approach for now (neither of the 1-3), dropping the hack that makes the loader tests pass ok, open an issue to implement kernel-mode FS redirection approach (4) instead, add the issue number to the loader test to be printed before marking test as failed. Make it an explicitly acknowledged regression so that there will be an incentive to fix it.

jschueller:
is this good to go ? (except for the x86_65 typo :)

sdk/tools/spec2def/spec2def.c
```diff
@@ -1365,7 +1400,70 @@ void usage(void)
...
+    if (result == 0)
+    {
+        return (pexpLeft->strName.len < pexpRight->strName.len) ? -1 : 1;
+    }
+
+    return result;
+}
+
+void
+Output_RosCompatDescriptor(FILE *file, EXPORT *pexports, unsigned int cExports)
```
zefklop, line 1427: I'd suggest generating the regular export directory, not use the .def file and call the C file <module>_exports.c

That would pave the way for relay tracing :-)

zefklop (commented):
Why don't you simply merge xxx and xxx_vista modules ?

zefklop (commented):
Wouldn't it make sense to have spec2def export every function, regardless of the target version ?

sdk/lib/crt/crt.cmake
```diff
@@ -368,6 +368,7 @@ if(ARCH STREQUAL "i386")
     list(APPEND CRT_ASM_SOURCE
         except/i386/chkesp.s
         except/i386/prolog.s
+        except/i386/ehandler-asm.s
```
zefklop, line 371: amd64 ehandler-asm.s missing

sdk/lib/crt/msvcrtex.cmake
```diff
@@ -67,7 +69,10 @@ if(ARCH STREQUAL "i386")
         math/i386/cisqrt.c)
 elseif(ARCH STREQUAL "amd64")
     list(APPEND MSVCRTEX_ASM_SOURCE
-        except/amd64/chkstk_ms.s)
+        except/amd64/chkstk_ms.s
+        except/amd64/ehandler-asm.s)
```
zefklop, line 73: msvcrtex is not linked into msvcrt.dll

jschueller:
hello, is it still in progress ?

jschueller:
hello, any news ?

dll/ntdll/ldr/appcompat.c
```diff
@@ -0,0 +1,390 @@
...
+    }
+    else
+    {
+        /* Patch the PEB */
+        PPEB Peb = NtCurrentPeb();
+        Peb->OSMajorVersion = AppCompatVersion >> 8;
+        Peb->OSMinorVersion = AppCompatVersion & 0xFF;
+    }
+
+    DPRINT("roscompat: Patching eat of %wZ for 0x%x\n", &LdrEntry->BaseDllName, AppCompatVersion);
```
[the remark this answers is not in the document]
learn-more: EAT (Export Address Table)

IniKiwi:
any news ?

tkreuzer:
Will be superceded by new simpler mechanism soon.

PR #3241, opened 2020-09-26
[FASTFAT] Fix create for DOT and DOT-DOT leaving bad directory entry

Doug-Lyons:
## Fix Create for DOT and DOT-DOT leaving bad directory entry.

_Add testing for these values and return appropriately._

JIRA issue: [CORE-11325](https://jira.reactos.org/browse/CORE-11325)

## Fixes to create.c and finfo.c in fastfat.

### DIFF
--- a/drivers/filesystems/fastfat/create.c
+++ b/drivers/filesystems/fastfat/create.c
@@ -360,9 +360,7 @@ VfatOpenFile(
         return STATUS_CANNOT_DELETE;
     }
 
-    if ((vfatFCBIsRoot(Fcb) ||
-         (Fcb->LongNameU.Length == sizeof(WCHAR) && Fcb->LongNameU.Buffer[0] == L'.') ||
-         (Fcb->LongNameU.Length == 2 * sizeof(WCHAR) && Fcb->LongNameU.Buffer[0] == L'.' && Fcb->LongNameU.Buffer[1] == L'.')) &&
+    if ((vfatFCBIsRoot(Fcb) || IsDotOrDotDot(&Fcb->LongNameU)) &&
         BooleanFlagOn(RequestedOptions, FILE_DELETE_ON_CLOSE))
     {
         // we cannot delete a '.', '..' or the root directory
@@ -791,6 +789,13 @@ VfatCreateFile(
                 Attributes |= FILE_ATTRIBUTE_ARCHIVE;
             }
             vfatSplitPathName(&PathNameU, NULL, &FileNameU);
+
+            if (IsDotOrDotDot(&FileNameU))
+            {
+                vfatReleaseFCB(DeviceExt, ParentFcb);
+                vfatAddToStat(DeviceExt, Fat.FailedCreates, 1);
+                return STATUS_OBJECT_NAME_INVALID;
+            }
             Status = VfatAddEntry(DeviceExt, &FileNameU, &pFcb, ParentFcb, RequestedOptions,
                                   Attributes, NULL);
             vfatReleaseFCB(DeviceExt, ParentFcb);
--- a/drivers/filesystems/fastfat/finfo.c
+++ b/drivers/filesystems/fastfat/finfo.c
@@ -380,9 +380,7 @@ VfatSetDispositionInformation(
         return STATUS_CANNOT_DELETE;
     }
 
-    if (vfatFCBIsRoot(FCB) ||
-        (FCB->LongNameU.Length == sizeof(WCHAR) && FCB->LongNameU.Buffer[0] == L'.') ||
-        (FCB->LongNameU.Length == 2 * sizeof(WCHAR) && FCB->LongNameU.Buffer[0] == L'.' && FCB->LongNameU.Buffer[1] == L'.'))
+    if (vfatFCBIsRoot(FCB) || IsDotOrDotDot(&FCB->LongNameU))
     {
         /* we cannot delete a '.', '..' or the root directory */
         return STATUS_ACCESS_DENIED;
@@ -803,6 +801,12 @@ VfatSetRenameInformation(
     DPRINT("Old dir: %wZ, Old file: %wZ\n", &SourcePath, &SourceFile);
     vfatSplitPathName(&NewName, &NewPath, &NewFile);
     DPRINT("New dir: %wZ, New file: %wZ\n", &NewPath, &NewFile);
+
+    if (IsDotOrDotDot(&NewFile))
+    {
+        Status = STATUS_OBJECT_NAME_INVALID;
+        goto Cleanup;
+    }
 
     if (vfatFCBIsDirectory(FCB) && !IsListEmpty(&FCB->ParentListHead))
     {
--- a/drivers/filesystems/fastfat/string.c
+++ b/drivers/filesystems/fastfat/string.c
@@ -3,7 +3,8 @@
  * PROJECT:          ReactOS kernel
  * FILE:             drivers/fs/vfat/string.c
  * PURPOSE:          VFAT Filesystem
- * PROGRAMMER:       Jason Filby (jasonfilby@yahoo.com)
+ * PROGRAMMERS:      Jason Filby (jasonfilby@yahoo.com)
+ *                   Doug Lyons (douglyons at douglyons dot com)
  *
  */
 
@@ -23,4 +24,11 @@ vfatIsLongIllegal(
     WCHAR c)
 {
     return wcschr(long_illegals, c) ? TRUE : FALSE;
+}
+
+BOOLEAN
+IsDotOrDotDot(PCUNICODE_STRING Name)
+{
+    return ((Name->Length == sizeof(WCHAR) && Name->Buffer[0] == L'.') ||
+        (Name->Length == 2 * sizeof(WCHAR) && Name->Buffer[0] == L'.' && Name->Buffer[1] == L'.'));
 }
--- a/drivers/filesystems/fastfat/vfat.h
+++ b/drivers/filesystems/fastfat/vfat.h
@@ -1221,6 +1221,10 @@ BOOLEAN
 vfatIsLongIllegal(
     WCHAR c);
 
+BOOLEAN
+IsDotOrDotDot(
+    PCUNICODE_STRING Name);
+
 /* volume.c */
 
 NTSTATUS


drivers/filesystems/fastfat/finfo.c
```diff
@@ -66,6 +66,12 @@ const char* FileInformationClassNames[] =
     "FileMaximumInformation"
 };
 
+BOOLEAN IsDots(UNICODE_STRING File)
```
ThFabba, line 69: I think I'd prefer if this were named even more explicitly (such that nobody reading the calling code has to look up what this function does), e.g. `IsDotOrDotDot`

drivers/filesystems/fastfat/create.c
```diff
@@ -791,6 +791,11 @@ VfatCreateFile(
                 Attributes |= FILE_ATTRIBUTE_ARCHIVE;
             }
             vfatSplitPathName(&PathNameU, NULL, &FileNameU);
+
+            if ((FileNameU.Length == sizeof(WCHAR) && FileNameU.Buffer[0] == L'.') ||
```
ThFabba, line 795: It would be nice to also use the function here. You could put it in `string.c` to make it clear it's a global utility.
There is also another place (`VfatOpenFile`) where it could make sense to use this function.

drivers/filesystems/fastfat/finfo.c
```diff
@@ -831,6 +835,10 @@ VfatSetRenameInformation(
                              (vfatFCBIsDirectory(FCB) ?
                               FILE_NOTIFY_CHANGE_DIR_NAME : FILE_NOTIFY_CHANGE_FILE_NAME),
                              FILE_ACTION_RENAMED_OLD_NAME);
+
+            if (IsDots(NewFile))
```
ThFabba, line 839: A single check right after the DPRINT in L805/809 above should do the trick I think.
`vfatReportChange` is a side-effect that we don't want to see unless the file is actually going to get moved.

drivers/filesystems/fastfat/create.c
Doug-Lyons: @ThFabba, Which string.c? There are many in the code. Thanks.

ThFabba: `drivers/filesystems/fastfat/string.c` near `vfatIsLongIllegal` (with a declaration in vfat.h, also near `vfatIsLongIllegal`)

Doug-Lyons: Got it. Thanks.

drivers/filesystems/fastfat/finfo.c
Doug-Lyons: Great idea. Done.

drivers/filesystems/fastfat/create.c
Doug-Lyons: Done. I did not add to VfatOpenFile yet. I will review. Thanks.

drivers/filesystems/fastfat/finfo.c
Doug-Lyons: Done. Thanks.

drivers/filesystems/fastfat/create.c
Doug-Lyons: OK. I have used the IsDotOrDotDot inside VfatOpenFile now.

drivers/filesystems/fastfat/string.c
```diff
@@ -24,3 +25,10 @@ vfatIsLongIllegal(
 {
     return wcschr(long_illegals, c) ? TRUE : FALSE;
 }
+
+BOOLEAN
+IsDotOrDotDot(UNICODE_STRING File)
```
ThFabba, line 30: Oops, forgot this the first time around, sorry. Passing `UNICODE_STRING`s by value is a bit strange since it's a structure larger than pointer size. I'd normally expect this function to take a `PCUNICODE_STRING` as its argument.

drivers/filesystems/fastfat/finfo.c
```diff
@@ -804,6 +802,9 @@ VfatSetRenameInformation(
     vfatSplitPathName(&NewName, &NewPath, &NewFile);
     DPRINT("New dir: %wZ, New file: %wZ\n", &NewPath, &NewFile);
 
+            if (IsDotOrDotDot(NewFile))
```
ThFabba: Indent is off here

drivers/filesystems/fastfat/finfo.c
```diff
@@ -864,6 +865,10 @@ VfatSetRenameInformation(
                              (vfatFCBIsDirectory(FCB) ?
                               FILE_NOTIFY_CHANGE_DIR_NAME : FILE_NOTIFY_CHANGE_FILE_NAME),
                              (DeletedTarget ? FILE_ACTION_REMOVED : FILE_ACTION_RENAMED_OLD_NAME));
+
+            if (IsDotOrDotDot(NewFile))
```
ThFabba: You no longer need this since it was already checked above.

drivers/filesystems/fastfat/finfo.c
```diff
@@ -920,6 +925,10 @@ VfatSetRenameInformation(
                          (vfatFCBIsDirectory(FCB) ?
                           FILE_NOTIFY_CHANGE_DIR_NAME : FILE_NOTIFY_CHANGE_FILE_NAME),
                          FILE_ACTION_REMOVED);
+
+        if (IsDotOrDotDot(NewFile))
```
ThFabba: No longer need this.

drivers/filesystems/fastfat/create.c
```diff
@@ -791,6 +789,10 @@ VfatCreateFile(
                 Attributes |= FILE_ATTRIBUTE_ARCHIVE;
             }
             vfatSplitPathName(&PathNameU, NULL, &FileNameU);
+
+            if (IsDotOrDotDot(FileNameU))
+                return STATUS_OBJECT_NAME_INVALID;
```
ThFabba, line 794: You'll need to clean up here and release the FCB (and, I suppose, add to statistics). See the failure path a few lines above.

drivers/filesystems/fastfat/finfo.c
```diff
@@ -804,6 +802,9 @@ VfatSetRenameInformation(
     vfatSplitPathName(&NewName, &NewPath, &NewFile);
     DPRINT("New dir: %wZ, New file: %wZ\n", &NewPath, &NewFile);
 
+            if (IsDotOrDotDot(NewFile))
+                return STATUS_OBJECT_NAME_INVALID;
```
ThFabba: You should `goto Cleanup` here, like other failure paths, to make sure we don't leak anything.

drivers/filesystems/fastfat/finfo.c
Doug-Lyons: Should be fixed now. Thanks.

drivers/filesystems/fastfat/finfo.c
Doug-Lyons: Removed. Thanks.

drivers/filesystems/fastfat/finfo.c
Doug-Lyons: Removed. Thanks.

drivers/filesystems/fastfat/string.c
Doug-Lyons: Fixed. Thanks.

drivers/filesystems/fastfat/create.c
Doug-Lyons: Fixed. Thanks.

drivers/filesystems/fastfat/finfo.c
Doug-Lyons: Fixed. Thanks.

drivers/filesystems/fastfat/string.c
```diff
@@ -24,3 +25,10 @@ vfatIsLongIllegal(
 {
     return wcschr(long_illegals, c) ? TRUE : FALSE;
 }
+
+BOOLEAN
+IsDotOrDotDot(PUNICODE_STRING File)
```
ThFabba, line 30: This could be a `const UNICODE_STRING *` or `PCUNICODE_STRING` since the string isn't modified.

ThFabba (approved):

drivers/filesystems/fastfat/string.c
Doug-Lyons: Sorry, I do not know what the differences might be. Please tell me your preference or if it can be left as is. These C/C++ types of questions are still magic to me.

Doug-Lyons: @ThFabba, Thanks for the approval. My testing shows that this will eliminate over 1500 lines showing the "disk repair" warning and shave about a minute off of the total regression set test time.

ThFabba: The addition of `const` will make it a compiler error for this function to try to modify the data pointed to by `File`. Pointers-to-const are generally preferred, unless the data needs to be modified.
So I would recommend the `PCUNICODE_STRING` here (which is a typedef for `const UNICODE_STRING *`).

Doug-Lyons: OK. I have made the changes and this does not appear to have changed the calling convention. Thanks.

Doug-Lyons:
See Testbot run compares here:
https://jira.reactos.org/browse/CORE-11325?focusedCommentId=125707&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-125707

drivers/filesystems/fastfat/string.c
```diff
@@ -24,3 +25,10 @@ vfatIsLongIllegal(
 {
     return wcschr(long_illegals, c) ? TRUE : FALSE;
 }
+
+BOOLEAN
+IsDotOrDotDot(PCUNICODE_STRING File)
```
HBelusca, line 30: I would prefer the name "FileName" instead of just "File", but it's just my own opinion. Otherwise good to go!

HBelusca (approved):

drivers/filesystems/fastfat/string.c
Doug-Lyons: How about just "Name" since this could be a File Name or a Directory Name? If you agree, then "fixed". Thanks.

drivers/filesystems/fastfat/vfat.h
```diff
@@ -1221,6 +1221,10 @@ BOOLEAN
 vfatIsLongIllegal(
     WCHAR c);
 
+BOOLEAN
+IsDotOrDotDot(
```
SergeGautherie, line 1225: Would it be worth to be inline or even a macro?

Doug-Lyons: @SergeGautherie, I followed the same pattern as the function above which is "vfatIsLongIllegal". This makes it consistent with what was done previously. If any of the other developers believes that this is beneficial, then maybe it can be added later as a follow-up to this PR. My preference is not to make changes to something that already has two approvals. Thanks.

Doug-Lyons:
@ThFabba, @HBelusca, Could one of you please do a squash and merge this for me at your convenience. The last time I tried to do a squash it did not work out so well. Thanks lots.

Extravert-ir:
I haven't figured out how this case is handled in other FS drivers, but what concerns me is that I don't see any similar code anywhere else.
Looks like in other drivers this check is more generalized. But it may still be OK for fastfat

Doug-Lyons:
> I haven't figured out how this case is handled in other FS drivers, but what concerns me is that I don't see any similar code anywhere else.
> Looks like in other drivers this check is more generalized. But it may still be OK for fastfat

@Extravert-ir, Do you know how the other FS's handle our kmtest:IoFilesystem test? Maybe they fail as well and do not handle this case either?

HBelusca:
For example the official MS fastfat_new driver doesn't have such checks, at least in an obvious manner.

ThFabba:
fastfat_new does a few calls to `FsRtlIsFatDbcsLegal` like this one, which will fail for `"."` and `".."`:
https://git.reactos.org/?p=reactos.git;a=blob;f=drivers/filesystems/fastfat_new/create.c;hb=0689455bb3e66f89b64aecefc2367efd454b9c16#l1564
However I can't find the specific call that would prevent the creation of new "." or ".." files. That might be because there will always be an existing DIRENT for these files, though if that's the case I don't know how we arrive at STATUS_OBJECT_NAME_INVALID.
Could be worth running the test on Windows with the fastfat sample driver and stepping through to see how it ends up failing. But I think the change proposed here is a good start in any case, and we can fix it more "correctly" later -- or, you know, discard the old fastfat driver. Whichever happens first ;)

binarymaster:
> Could one of you please do a squash and merge this for me at your convenience.

Done.

Doug-Lyons:
Thanks very much, @binarymaster.

PR #3246, opened 2020-09-27
Add NT6+ style support to uxtheme

wjk:
## Purpose

This PR adds support for uxtheme classes, parts, and styles added in NT6.0. I use these parts in my applications to draw text using the Vista "heading text" style (12-point #003399 Segoe UI), and to draw the "task dialog footer area" background, both while automatically handling style changes. Under Windows 10, this only happens when High Contrast mode is toggled. With ReactOS, theme changes can happen much more often. By pulling values from uxtheme rather than hardcoding fonts/sizes/colors, I make my application more resilient against this changes.

## TODO

- [ ] None of the msstyles files in this repo have any support for the classes added in this PR. This is partially due to the fact that adding the code for every part and state in every INI file in every theme would be very time-consuming, and partially due to the fact that some parts may need images created, and I have no graphics skills.


binarymaster:
Any showcases? :slightly_smiling_face:

dll/win32/uxtheme/stylemap.c
```diff
@@ -40,7 +40,11 @@ typedef struct _MSSTYLES_CLASS_MAP {
 
 typedef struct _MSSTYLES_CLASS_NAME {
     const MSSTYLES_CLASS_MAP *lpMap;
+#if __REACTOS__
```
HBelusca, line 43: Please fix all these `#if __REACTOS__` to `#ifdef __REACTOS__`

HBelusca (changes requested):

wjk:
@binarymaster I don't quite understand what you’re asking. If you mean showcasing code that uses these styles, here is [one example](https://github.com/microsoft/PowerToys/blob/4d1046d6e3dfb565b6961a6da28afddf167bbb65/src/modules/fastdelete/FastDelete.ShellExtension/HeaderLabel.cs), [and another](https://github.com/microsoft/PowerToys/blob/4d1046d6e3dfb565b6961a6da28afddf167bbb65/src/modules/fastdelete/FastDelete.ShellExtension/FooterPanel.cs). If you mean what these parts look like when used on Windows, [there are examples of Aero Wizards and Task Dialogs on this page](https://docs.microsoft.com/en-us/windows/win32/uxguide/win-dialog-box).

dll/win32/uxtheme/stylemap.c
wjk: This has been done in f552e84.

binarymaster:
> If you mean what these parts look like when used on Windows

Yes, I meant this, and also how it will look like in ReactOS (if it's possible to show)

milawynsrealm:
https://www.mctrl.org/tools.php
There is a tool called Theme Explorer that might help you.

wjk:
I'll work on integrating the new parts into one of the themes (probably Modern/Light) so I can show you them running on ReactOS, but it’ll take some time. Once that looks good, I’ll copy the code into the other inbox msstyles files. Adding proper support for the Aero Wizard and Task Dialog controls (so the styles can be shown in their native locations) is outside the scope of this PR, but I’ll probably work on submitting those features in separate PRs sometime.

One other note: All third-party ReactOS msstyles files (including those in RAPPS) will _not_ contain these parts, because those files were designed for (or taken from) Windows XP, and the parts were added in Windows Vista. However, as I understand it this will not cause anything to crash; instead, uxtheme.dll will look in the msstyles file, not find what it is looking for, and return `E_PROP_ID_UNSUPPORTED` from the function called. How much of a compatibility issue this creates (as compared to Windows, where these parts are always available), and any potential mitigations thereto, is beyond my ability to determine. Thanks!

wjk:
Closed PR by accident. Sorry about that!

dll/win32/uxtheme/stylemap.c
```diff
@@ -1005,8 +1009,75 @@ static const MSSTYLES_CLASS_MAP classWindow[] = {
     {0,0,{'\0'}}
 };
 
+#ifdef __REACTOS__
+static const MSSTYLES_CLASS_MAP classAeroWizard[] = {
+    {AW_TITLEBAR,0,L"TitleBar"},
```
katahiromz, line 1014: space after comma

wjk: I was following the style of the rest of the file, which uses no spaces throughout.

katahiromz: OK.

sdk/include/psdk/tmschema.h
```diff
@@ -1247,6 +1247,91 @@ BEGIN_TM_PART_STATES(VERTTHUMB)
     TM_STATE(4, VTS, DISABLED)
 END_TM_PART_STATES()
 
+/* AEROWIZARD parts */
```
HBelusca, line 1250: I wonder whether it would be of interest to surround the NT6+ stuff within `#if (NTDDI_VERSION >= NTDDI_VISTA) ... #endif`

dll/win32/uxtheme/stylemap.c
```diff
@@ -1005,8 +1009,75 @@ static const MSSTYLES_CLASS_MAP classWindow[] = {
     {0,0,{'\0'}}
 };
 
+#ifdef __REACTOS__
```
HBelusca, line 1012: Same comment regarding the NTDDI_VERSION as below.

PR #3247, opened 2020-09-28
[BROWSEUI] Fix Edit_BackWord function

katahiromz:
## Purpose

Improve deletion of characters about `Ctrl+Back`.
JIRA issue: [CORE-1419](https://jira.reactos.org/browse/CORE-1419)

## Proposed changes

- Fix `Edit_BackWord` function by using `GetStringTypeW`.

### DIFF
--- a/dll/win32/browseui/CAutoComplete.cpp
+++ b/dll/win32/browseui/CAutoComplete.cpp
@@ -299,13 +299,12 @@ static void Edit_BackWord(HWND hwndEdit)
     if (GetWindowTextW(hwndEdit, pszText, cchText + 1) <= 0 || cchText < (size_t)iStart)
         return;
 
-    WORD wType1, wType2;
+    WORD types[2];
     for (--iStart; 0 < iStart; --iStart)
     {
-        GetStringTypeW(CT_CTYPE1, &pszText[iStart - 1], 1, &wType1);
-        GetStringTypeW(CT_CTYPE1, &pszText[iStart], 1, &wType2);
-        if (((wType1 & C1_PUNCT) && !(wType2 & C1_SPACE)) ||
-            ((wType1 & C1_SPACE) && (wType2 & (C1_ALPHA | C1_DIGIT))))
+        GetStringTypeW(CT_CTYPE1, &pszText[iStart - 1], 2, types);
+        if (((types[0] & C1_PUNCT) && !(types[1] & C1_SPACE)) ||
+            ((types[0] & C1_SPACE) && (types[1] & (C1_ALPHA | C1_DIGIT))))
         {
             SendMessageW(hwndEdit, EM_SETSEL, iStart, iEnd);
             SendMessageW(hwndEdit, EM_REPLACESEL, TRUE, (LPARAM)L"");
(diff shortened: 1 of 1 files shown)


katahiromz:
https://jira.reactos.org/secure/attachment/58057/chartype.zip
This is a test program for `GetStringTypeW`.

dll/win32/browseui/CAutoComplete.cpp
```diff
@@ -302,12 +302,15 @@ static void Edit_BackWord(HWND hwndEdit)
...
+    for (--iStart; 0 < iStart; --iStart)
     {
-        WCHAR ch1 = pszText[iStart - 1];
-        WCHAR ch2 = pszText[iStart];
-        if ((wcschr(L"\\/.:;", ch1) && ch2 && !IsCharSpaceW(ch2)) ||
-            (IsCharSpaceW(ch1) && IsCharAlphaNumericW(ch2)))
+        WCHAR& ch1 = pszText[iStart - 1];
+        WCHAR& ch2 = pszText[iStart];
+        GetStringTypeW(CT_CTYPE1, &ch1, 1, &wType1);
+        GetStringTypeW(CT_CTYPE1, &ch2, 1, &wType2);
```
learn-more, line 311: Why not directly pass `pszText + iStart` into `GetStringTypeW` ?
It seems pretty weird to make `ch1` and `ch2` a reference to an entry in an array?

dll/win32/browseui/CAutoComplete.cpp
```diff
@@ -302,12 +302,15 @@ static void Edit_BackWord(HWND hwndEdit)
         return;
     }
 
-    for (; 0 < iStart; --iStart)
+    WORD wType1, wType2;
+    for (--iStart; 0 < iStart; --iStart)
```
learn-more, line 306: is `iStart` guaranteed to be smaller than `cchText` ?

katahiromz: I think `iStart` is `<= cchText`.

dll/win32/browseui/CAutoComplete.cpp
katahiromz: Fixed. OK.

dll/win32/browseui/CAutoComplete.cpp
katahiromz: I added a check `(cchText < iStart)`. OK.

dll/win32/browseui/CAutoComplete.cpp
```diff
@@ -290,7 +290,7 @@ static void Edit_BackWord(HWND hwndEdit)
     if (iStart != iEnd)
         return;
 
-    DWORD cchText = GetWindowTextLengthW(hwndEdit);
+    INT cchText = GetWindowTextLengthW(hwndEdit);
```
learn-more, line 293: It `should` probably never happen, but theoretically cchText can be negative.

f.e. riched20 can return `E_INVALIDARG`, from `WM_GETTEXTLENGTH`, which translates to `0x80070057L`, which will be negative when interpreted as signed int.

katahiromz: Made `cchText` a `size_t`. Added `iStart` and `iEnd` sign checks. OK.

PR #3248, opened 2020-09-28
WIP - [UMPNPMGR] Investigate bug crash. CORE-16103

HBelusca:
Actually use linked-lists. Reverts part of commit 043a98dd (see also commit b2aeafca).

## Purpose

Investigation of the bug crash reported at
JIRA issue: [CORE-16103](https://jira.reactos.org/browse/CORE-16103)

## Proposed changes

I investigate the usage of regular (doubly) linked lists with explicit mutex locking.

### DIFF
--- a/base/services/umpnpmgr/install.c
+++ b/base/services/umpnpmgr/install.c
@@ -22,7 +22,7 @@
  * FILE:             base/services/umpnpmgr/install.c
  * PURPOSE:          Device installer
  * PROGRAMMER:       Eric Kohl (eric.kohl@reactos.org)
- *                   Hervé Poussineau (hpoussin@reactos.org)
+ *                   HervÃ© Poussineau (hpoussin@reactos.org)
  *                   Colin Finck (colin@reactos.org)
  */
 
@@ -40,7 +40,9 @@ HANDLE hUserToken = NULL;
 HANDLE hInstallEvent = NULL;
 HANDLE hNoPendingInstalls = NULL;
 
-SLIST_HEADER DeviceInstallListHead;
+/* Device-install event list */
+HANDLE hDeviceInstallListMutex;
+LIST_ENTRY DeviceInstallListHead;
 HANDLE hDeviceInstallListNotEmpty;
 
 
@@ -354,7 +356,7 @@ DWORD
 WINAPI
 DeviceInstallThread(LPVOID lpParameter)
 {
-    PSLIST_ENTRY ListEntry;
+    PLIST_ENTRY ListEntry;
     DeviceInstallParams* Params;
     BOOL showWizard;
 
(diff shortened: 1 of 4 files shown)


HBelusca:
@JoachimHenze : please test this PR and tell me the results.

JoachimHenze (dismissed):
The fix does not work for me yet to fix https://jira.reactos.org/browse/CORE-16103
I tried after the first commit 6741c3f. see
https://jira.reactos.org/browse/CORE-16103?focusedCommentId=125732&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-125732
for log and screenshot.

SergeGautherie:
> I tried after the first commit [6741c3f](https://github.com/reactos/reactos/commit/6741c3f446560a0aa2144b87359f40bfe7f8f290).

That kind of confirms the order issue.
Did you try with `RemoveHeadList()` instead?

HBelusca:
> > I tried after the first commit [6741c3f](https://github.com/reactos/reactos/commit/6741c3f446560a0aa2144b87359f40bfe7f8f290).
> 
> That kind of confirms the order issue.
> Did you try with `RemoveHeadList()` instead?

Yes that's what I'm thinking about too.

base/services/umpnpmgr/install.c
```diff
@@ -366,7 +367,10 @@ DeviceInstallThread(LPVOID lpParameter)
 
     while (TRUE)
     {
-        ListEntry = InterlockedPopEntrySList(&DeviceInstallListHead);
+        WaitForSingleObject(hDeviceInstallListMutex, INFINITE);
+        ListEntry = (IsListEmpty(&DeviceInstallListHead)
+                        ? NULL : /*RemoveHeadList*/RemoveTailList(&DeviceInstallListHead));
```
HBelusca, line 372: @JoachimHenze : Please retest with this modification:
```suggestion
                        ? NULL : RemoveHeadList(&DeviceInstallListHead));
```

JoachimHenze:
Okay, I do confirm based on 20 of 20 successfuly installation attempts which did not expose the issue a single time, that changing that line to use RemoveHeadList() instead of the former RemoveTailList(), does successfully fix the bug.
Let's get that merged into master quickly, so I can backport into 0.4.14. I will change the state of my Review from "Changes Requested" to "Approved" after that change is visible in github, not earlier. Just for safety-reasons.
After that change was done, you can still immediately merge, no need to wait for me changing my state before that. No need for additional delay.

base/services/umpnpmgr/install.c
binarymaster: Ship it :)

SergeGautherie:
> I do confirm

Thanks!

> Let's get that merged into master quickly

Well, it seems to me using a mutex was part of the test too:
final code should rather use `ExInterlockedRemoveHeadList()`...

HBelusca:
@SergeGautherie : we don't need your confirmations when the situation is already clear and understood, especially if it's for saying something wrong too (your last sentence).

SergeGautherie:
> we don't need your confirmations

Wow...

> wrong too (your last sentence).

What is wrong (with the `ExInterlocked*List()`)?

PR #3250, opened 2020-09-29
[BROWSEUI] Accept TypedURLs to CLSID_ACLCustomMRU

katahiromz:
## Purpose

Related to #3249. `IACLCustomMRU` has to load the `TypedURLs` key correctly.
The `TypedURLs` key consists of the registry values of `"url1", "url2", "url3" etc` instead of `"MRUList"`, `"a"`, `"b"` etc.

## Proposed changes

- In `CACLCustomMRU::Initialize` method, if the key was `TypedURLs`, then load the `TypedURLs` registry key.
- If the key was `TypedURLs`, then don't save the entries.
- Implement `IEnumString` and `IACList`.

## TODOs

- [x] Reduce failures.

BEFORE:
before
2 failures, 1 skipped.

AFTER:
after
Successful.


katahiromz:
Ready to go!

PR #3251, opened 2020-09-30
[PORTCLS] Fix races, fix WavePci Port Drivers

TheDeadFish:
## Purpose
Fix WavePci Port Drivers, (like ac97 sample driver)
Also fix race conditions

## Proposed changes
Split buffers on page to prevent non-contiguous memory being passed to driver
Protect GetMappingWithTag, ReleaseMappingWithTag with spinlock to prevent race conditions
(GetMapping, ReleaseMapping do not need spinlock, they are only called from a service routine)
Remove HasLastMappingFailed(), it never worked properly and can never work

Additional changes:
Move some variables to class level from IRP data, (they did not need to be in IRP data as only one set is active at a time)
Remove m_NumDataAvailable, it was not implemented correctly and was only used for debugging info, if that info is really needed it can be recreated on demand by walking the IRP list. (Not implemented yet)
Removed all remnants of volatile and interlocked operations, they are not needed and were not correct anyway


drivers/wdm/audio/backpln/portcls/irpstream.cpp
```diff
@@ -14,6 +14,50 @@
 
 #include <debug.h>
 
+static
+KSDDKAPI
+PIRP
+NTAPI
```
HBelusca, line 20: Not sure it's needed to have a locally-used-only function marked as if it was an exported public API.

drivers/wdm/audio/backpln/portcls/irpstream.cpp
```diff
@@ -14,6 +14,50 @@
...
+        {
+
+            /* remove irp from list */
+            RemoveEntryList(&Irp->Tail.Overlay.ListEntry);
+
+            /* return irp */
+            return Irp;
+        }
+
+    }while(TRUE);
```
HBelusca, line 56:
```suggestion
    for (CurEntry = QueueHead->Flink; CurEntry != QueueHead; CurEntry = CurEntry->Flink)
    {
        /* Get the IRP offset */
        Irp = (PIRP)CONTAINING_RECORD(CurEntry, IRP, Tail.Overlay.ListEntry);

        /* Remove the cancel routine */
        if (IoSetCancelRoutine(Irp, NULL))
        {
            /* Remove the IRP from the list and return it */
            RemoveEntryList(&Irp->Tail.Overlay.ListEntry);
            return Irp;
        }
    }
```

drivers/wdm/audio/backpln/portcls/irpstream.cpp
```diff
@@ -14,6 +14,50 @@
 
 #include <debug.h>
 
+static
+KSDDKAPI
+PIRP
+NTAPI
+RemoveHeadList_IRP(
+    IN  OUT PLIST_ENTRY QueueHead)
```
HBelusca, line 22:
```suggestion
    IN OUT PLIST_ENTRY QueueHead)
```

drivers/wdm/audio/backpln/portcls/irpstream.cpp
```diff
@@ -185,15 +234,23 @@ CIrpQueue::AddMapping(
...
+            HeaderLength = Header->FrameExtent;
         }
 
-        /* move to next header */
+        // increment available data
+        TotalStreamData += HeaderLength;
+
+        // append page count
+        StreamPageCount += ADDRESS_AND_SIZE_TO_SPAN_PAGES(
+            MmGetMdlByteOffset(Mdl), HeaderLength);
```
HBelusca, line 250:
```suggestion
        StreamPageCount += ADDRESS_AND_SIZE_TO_SPAN_PAGES(
                               MmGetMdlByteOffset(Mdl), HeaderLength);
```

drivers/wdm/audio/backpln/portcls/irpstream.cpp
```diff
@@ -322,6 +365,16 @@ CIrpQueue::GetMapping(
         // get a fresh new irp from the queue
         m_Irp = Irp = KsRemoveIrpFromCancelableQueue(&m_IrpList, &m_IrpListLock, KsListEntryHead, KsAcquireAndRemoveOnlySingleItem);
         m_CurrentOffset = Offset = 0;
+
+
```
HBelusca, line 369:
```suggestion
```

drivers/wdm/audio/backpln/portcls/irpstream.cpp
```diff
@@ -496,7 +534,8 @@ NTAPI
 CIrpQueue::NumData()
 {
     // returns the amount of audio stream data available
-    return m_NumDataAvailable;
+    UNIMPLEMENTED;
+    return 0;
```
HBelusca, line 538: It's not possible now to know this data? Or is NumData() going to be deprecated?

drivers/wdm/audio/backpln/portcls/irpstream.cpp
```diff
@@ -534,84 +570,115 @@ CIrpQueue::GetMappingWithTag(
...
+    ULONG Size;
+    LPBYTE Data;
 
     /* sanity checks */
     PC_ASSERT(PhysicalAddress);
     PC_ASSERT(VirtualAddress);
     PC_ASSERT(ByteCount);
     PC_ASSERT(Flags);
 
+
```
HBelusca, line 583:
```suggestion
```

drivers/wdm/audio/backpln/portcls/irpstream.cpp
```diff
@@ -534,84 +570,115 @@ CIrpQueue::GetMappingWithTag(
...
+    KeAcquireSpinLock(&m_IrpListLock, &OldLevel);
```
HBelusca, line 584: Why does the spin lock for the IRP list need to be acquired until the very end of this function? Isn't it possible to use again these ExInterlocked* functions as before? (and doing the explicit KeAcquire/ReleaseSpinLock calls only in RemoveHeadList_IRP ?)

drivers/wdm/audio/backpln/portcls/irpstream.cpp
```diff
@@ -534,84 +570,115 @@ CIrpQueue::GetMappingWithTag(
...
+    // setup mapping
+    Data = (LPBYTE)StreamData->Data[m_StreamHeaderIndex] + m_CurrentOffset;
+    *VirtualAddress = Data;
 
-        // decrement num data available
-        m_NumDataAvailable -= StreamData->CurStreamHeader->FrameExtent;
-    }
+    // get byte count
+    *ByteCount = (LPBYTE)ROUND_TO_PAGES(Data+1)-Data;
+    if(*ByteCount > (Size - m_CurrentOffset))
```
HBelusca, line 640:
```suggestion
    if (*ByteCount > (Size - m_CurrentOffset))
```

drivers/wdm/audio/backpln/portcls/irpstream.cpp
```diff
@@ -627,82 +694,63 @@ CIrpQueue::ReleaseMappingWithTag(
...
-        // now check if there are already used mappings
-        StreamData = (PKSSTREAM_DATA)m_Irp->Tail.Overlay.DriverContext[STREAM_DATA_OFFSET];
+    KeAcquireSpinLock(&m_IrpListLock, &OldLevel);
 
-        if (StreamData->StreamHeaderIndex)
+    // check if used list empty
+    if (IsListEmpty(&m_FreeIrpList))
+    {
+        // get current irp
+        if(m_Irp == NULL)
```
HBelusca, line 705:
```suggestion
        if (!m_Irp)
```

drivers/wdm/audio/backpln/portcls/irpstream.cpp
```diff
@@ -627,82 +694,63 @@ CIrpQueue::ReleaseMappingWithTag(
...
     // get stream data
     StreamData = (PKSSTREAM_DATA)Irp->Tail.Overlay.DriverContext[STREAM_DATA_OFFSET];
 
-    // sanity check
-    PC_ASSERT(StreamData->StreamHeaderIndex == StreamData->StreamHeaderCount);
-
-    // check if the released mapping is one of these
-    for(Index = 0; Index < StreamData->StreamHeaderCount; Index++)
+    // release oldest in use mapping
+    for(Index = 0; Index < StreamData->nTags; Index++)
```
HBelusca, line 729:
```suggestion
    for (Index = 0; Index < StreamData->nTags; Index++)
```

drivers/wdm/audio/backpln/portcls/irpstream.cpp
```diff
@@ -627,82 +694,63 @@ CIrpQueue::ReleaseMappingWithTag(
...
-            (StreamData->Tags[Index].Used != FALSE))
+        if (StreamData->Tags[Index].Used != FALSE)
         {
-            // mark mapping as released
-            StreamData->Tags[Index].Tag = NULL;
             StreamData->Tags[Index].Used = FALSE;
 
-            // done
+            // Warn if wrong mapping released
+            if(StreamData->Tags[Index].Tag != Tag)
```
HBelusca, line 736:
```suggestion
            if (StreamData->Tags[Index].Tag != Tag)
```

drivers/wdm/audio/backpln/portcls/irpstream.cpp
```diff
@@ -627,82 +694,63 @@ CIrpQueue::ReleaseMappingWithTag(
...
-            // therefore if the current mapping is not the searched one, it must have been already
-            // released
-            //
-            ASSERT(StreamData->Tags[Index].Tag == NULL);
-            ASSERT(StreamData->Tags[Index].Used == FALSE);
-        }
+    }
+
+    // current IRP, do not complete
+    if(Irp == m_Irp)
```
HBelusca, line 746:
```suggestion
    if (Irp == m_Irp)
```

drivers/wdm/audio/backpln/portcls/irpstream.cpp
```diff
@@ -627,82 +694,63 @@ CIrpQueue::ReleaseMappingWithTag(
...
-            // we assume that mappings are released in the same order as they have been acquired
-            // therefore if the current mapping is not the searched one, it must have been already
-            // released
-            //
-            ASSERT(StreamData->Tags[Index].Tag == NULL);
-            ASSERT(StreamData->Tags[Index].Used == FALSE);
-        }
+    }
+
+    // current IRP, do not complete
```
HBelusca, line 745:
```suggestion
    // If this is the current IRP, do not complete
```

drivers/wdm/audio/backpln/portcls/irpstream.cpp
```diff
@@ -322,6 +365,16 @@ CIrpQueue::GetMapping(
         // get a fresh new irp from the queue
         m_Irp = Irp = KsRemoveIrpFromCancelableQueue(&m_IrpList, &m_IrpListLock, KsListEntryHead, KsAcquireAndRemoveOnlySingleItem);
         m_CurrentOffset = Offset = 0;
+
+
+        if(m_Irp)
```
HBelusca, line 370:
```suggestion
        if (m_Irp)
```

drivers/wdm/audio/backpln/portcls/irpstream.cpp
TheDeadFish: Its not, I just have no idea what those mean, I copied that function from elsewhere and modified it to suit my needs.

TheDeadFish:
@HBelusca Is there any tool to find these basic formatting changes or are you just doing it all by eye?

drivers/wdm/audio/backpln/portcls/irpstream.cpp
HBelusca: You can then remove the `KSDDKAPI` and `NTAPI` thingies there.

HBelusca:
> @HBelusca Is there any tool to find these basic formatting changes or are you just doing it all by eye?

I see them by eye very easily. But if you want to "automatize" you can search for "common patterns": `if(`, `for(`, `while(` etc... (basically those where people are most susceptible to forget the formatting spacing).

drivers/wdm/audio/backpln/portcls/irpstream.cpp
TheDeadFish: The entire function needs to be atomic not just the individual variables. The existing lock-less code was definitely wrong. It was wrong in the first few lines, if two threads ran the code below at the same time two IRPs could end up being pulled from the queue and one of them would be lost since m_Irp is a shared variable.

if (!m_Irp) { m_Irp = KsRemoveIrpFromCancelableQueue(); }

drivers/wdm/audio/backpln/portcls/irpstream.cpp
TheDeadFish: The value can be computed on demand pretty easily, I can code that up now if you want. I just wanted to get the ball rolling with this PR before I just kept messing around with it forever.

drivers/wdm/audio/backpln/portcls/irpstream.cpp
TheDeadFish:
```suggestion
PIRP
```

drivers/wdm/audio/backpln/portcls/irpstream.cpp
TheDeadFish: How did you arrive at that indentation depth it seems arbitrary?

HBelusca: That was to put the function/macro's parameters below the other function-related stuff (usually we do:
```c
returnValue = FunctionCall(first_param,
                           ...);
```
when the first line is not too long already).

TheDeadFish: Ok I understand and appreciate the aesthetic of keeping the function arguments further to the right that the function name. (I still think full alignment as given in your example is disgusting, but the suggestion is good)

janderwald (commented):
Why is RemoveHeadList_Irp requiered? Please use KsRemoveIrpFromCancellableQueue.

HBelusca:
> Why is RemoveHeadList_Irp requiered? Please use KsRemoveIrpFromCancellableQueue.

@janderwald : @TheDeadFish explained to me in a previous comment:
https://github.com/reactos/reactos/pull/3251#discussion_r498355229
> The entire function needs to be atomic not just the individual variables. The existing lock-less code was definitely wrong. It was wrong in the first few lines, if two threads ran the code below at the same time two IRPs could end up being pulled from the queue and one of them would be lost since m_Irp is a shared variable.
>
> if (!m_Irp) { m_Irp = KsRemoveIrpFromCancelableQueue(); }

janderwald:
> > Why is RemoveHeadList_Irp requiered? Please use KsRemoveIrpFromCancellableQueue.
> 
> @janderwald : @TheDeadFish explained to me in a previous comment:
> [#3251 (comment)](https://github.com/reactos/reactos/pull/3251#discussion_r498355229)
> 
> > The entire function needs to be atomic not just the individual variables. The existing lock-less code was definitely wrong. It was wrong in the first few lines, if two threads ran the code below at the same time two IRPs could end up being pulled from the queue and one of them would be lost since m_Irp is a shared variable.
> > if (!m_Irp) { m_Irp = KsRemoveIrpFromCancelableQueue(); }

Please have a look at that KsRemove... function. It should already do that. When the behaviour is not atomic, then it should be fixed

Zero3K:
> > > Why is RemoveHeadList_Irp requiered? Please use KsRemoveIrpFromCancellableQueue.
> > 
> > 
> > @janderwald : @TheDeadFish explained to me in a previous comment:
> > [#3251 (comment)](https://github.com/reactos/reactos/pull/3251#discussion_r498355229)
> > > The entire function needs to be atomic not just the individual variables. The existing lock-less code was definitely wrong. It was wrong in the first few lines, if two threads ran the code below at the same time two IRPs could end up being pulled from the queue and one of them would be lost since m_Irp is a shared variable.
> > > if (!m_Irp) { m_Irp = KsRemoveIrpFromCancelableQueue(); }
> 
> Please have a look at that KsRemove... function. It should already do that. When the behaviour is not atomic, then it should be fixed

What is the difference between the two?

TheDeadFish:
> > > Why is RemoveHeadList_Irp requiered? Please use KsRemoveIrpFromCancellableQueue.
> > 
> > 
> > @janderwald : @TheDeadFish explained to me in a previous comment:
> > [#3251 (comment)](https://github.com/reactos/reactos/pull/3251#discussion_r498355229)
> > > The entire function needs to be atomic not just the individual variables. The existing lock-less code was definitely wrong. It was wrong in the first few lines, if two threads ran the code below at the same time two IRPs could end up being pulled from the queue and one of them would be lost since m_Irp is a shared variable.
> > > if (!m_Irp) { m_Irp = KsRemoveIrpFromCancelableQueue(); }
> 
> Please have a look at that KsRemove... function. It should already do that. When the behaviour is not atomic, then it should be fixed
KsRemoveIrpFromCancelableQueue() takes ownership of the IRP, and if we were to then process and complete the IRP right there then that would be fine. But we do not complete the IRP right there, we place the IRP into a shared variable m_irp and then operate on that same IRP over multiple calls to GetMappingWithTag, those calls can occur simultaneously from multiple threads, so the entire state associated with processing that IRP must be protected by a lock or the state would get messed up. The reason I use, RemoveHeadList_Irp is because KsRemoveIrpFromCancellableQueue acquires and releases the spinlock, but I need to keep the spinlock throughout the entire function.

PR #3252, opened 2020-09-30
[RAPPS][RAPPS.COM] Introduce a rapps console shim

learn-more:
rapps.com will be launched when running 'rapps' from cmd or
from run, and since it is a console application, explorer/ cmd will wait for it.
This rapps.com passes the commandline to rapps.exe,
and wait for rapps.exe to exit before closing.

JIRA issue: [CORE-17281](https://jira.reactos.org/browse/CORE-17281)


base/applications/CMakeLists.txt
```diff
@@ -30,6 +30,7 @@ add_subdirectory(network)
 add_subdirectory(notepad)
 add_subdirectory(osk)
 add_subdirectory(rapps)
+add_subdirectory(rapps_com)
```
HBelusca, line 33: Idea: Could you put this stub console shim as a subdirectory of rapps instead? (for self-containedness)

learn-more: Not sure if that's an improvement, the applications folder already has quite some arbitrary folders (cmdutils, screensavers, ..?)
And this will only make it harder to find stuff

base/applications/rapps_com/main.c
```diff
@@ -0,0 +1,73 @@
+/*
+ * PROJECT:     rapps.com
```
HBelusca, line 2:
```suggestion
 * PROJECT:     ReactOS Applications Manager Command-Line Launcher (rapps.com)
```

base/applications/rapps_com/main.c
```diff
@@ -0,0 +1,73 @@
...
+        wchar_t* ptr = cmdline;
+        size_t cchRemaining = arglen;
+
+        StringCchPrintfExW(ptr, cchRemaining, &ptr, &cchRemaining, 0, L"\"%s\"", RappsExe);
+
+        for (int n = 1; n < argc; ++n)
+        {
+            StringCchPrintfExW(ptr, cchRemaining, &ptr, &cchRemaining, 0, L" \"%s\"", argv[n]);
+        }
+    }
```
kernelbin, line 67: why not just use ```GetCommandLineW``` instead of jointing strings from argv🤣

learn-more: Because then I have to manually account for quoted arguments.

PR #3254, opened 2020-10-01
[MSGINA] Implement ShellTurnOffDialog CORE-17313

oleg-dubinskiy:
## Purpose

Implement undocumented `ShellTurnOffDialog` function based on the (most of all correct) following prototype: http://diendan.congdongcviet.com/threads/t13622::tim-handle-cua-cua-so-tat-may-tren-windows-nhu-the-nao.cpp?p=69284#post69284 (line 32) and return ShellShutdownDialog from it with all required parameters.
It fixes the crash when trying to open the shutdown dialog from the Start Menu with shell32.dll (and other required dlls) from Windows Server 2003 SP2 (when installing ReactOS as Workstation). So now it opens successfully, and it is possible to restart or shutdown the system properly from Start Menu without logging off (with replaced shell).

JIRA issue: [CORE-17313](https://jira.reactos.org/browse/CORE-17313)

## Result

Before:
MS_shell32_shudtdown_dlg_bug

After:
MS_shell32_shudtdown_dlg_fixed

### DIFF
--- a/dll/win32/msgina/msgina.spec
+++ b/dll/win32/msgina/msgina.spec
@@ -6,7 +6,7 @@
 6 stub -noname ShellEnableFriendlyUI    ; (long)
 7 stub -noname ShellEnableMultipleUsers ; (long)
 8 stub -noname ShellEnableRemoteConnections ; (long)
-9 stub -noname ShellTurnOffDialog ; (long)
+9 stdcall -noname ShellTurnOffDialog(ptr)
 10 stub -noname ShellIsMultipleUsersEnabled ; ()
 11 stub -noname ShellACPIPowerButtonPressed ; (long long long)
 12 stub -noname ShellIsSingleUserNoPassword ; (wstr wstr)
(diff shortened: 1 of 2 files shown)


HBelusca:
LGTM!

Out of curiosity, could you try to see whether you can trigger a different behaviour on Windows XP and 2003, by using this function and comparing with the ShellShutdownDialog, if you play around with some registry values (that we also support in ReactOS: https://github.com/reactos/reactos/commit/2edcb58e6550fd6283cf3a7d332ee61849704f3d ) ?

oleg-dubinskiy:
OK @HBelusca. I'll try that later.

oleg-dubinskiy:
Done @HBelusca.
I tried to change `LogonType` and `ForceFriendlyUI`, as proposed in the reference links from the commit, but it didn't change anything for me. :frowning_face: There is still exactly the same shutdown UI, as in XP, same in 2003.
Probably it also depends on `ProductOptions` parameters, whose obviously can't be changed in Windows, for licensing reasons. :thinking:

erkinalp:
There is an issue with button layout. The icons that are supposed to appear, do not appear.

learn-more:
> There is an issue with button layout. The icons that are supposed to appear, do not appear.

unthemed buttons...

PR #3255, opened 2020-10-01
[USETUP] Fix usetup screen transition. CORE-17312

HBelusca:
## Purpose

Fix the visual "blinking" that occurs due to screen clearing, whenever a screen transition is done.

JIRA issue: [CORE-17312](https://jira.reactos.org/browse/CORE-17312)

## Proposed changes

Hide everything under the same foreground & background colors, so that
the actual color and text blanking reset does not create a visual "blinking".
Then, blank the text and finally reset the actual foreground &
background colors.

We do this because we cannot do the screen scrolling trick that would
allow to change both the text and the colors at the same time (the
function is currently not available in our console "emulation" layer).

### DIFF
--- a/base/setup/usetup/consup.c
+++ b/base/setup/usetup/consup.c
@@ -235,17 +235,32 @@ CONSOLE_ClearScreen(VOID)
     coPos.X = 0;
     coPos.Y = 0;
 
+    /*
+     * Hide everything under the same foreground & background colors, so that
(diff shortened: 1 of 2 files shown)


JoachimHenze (approved):
I did test this fix (after the 2nd commit 434fa56) applied manually on top of 0.4.15-dev-1076-g0733d96 and I do confirm that it properly fixed CORE-17312. No wishes left from my side. Thank you very much!

PR #3256, opened 2020-10-02
[MSGINA] Improve exports for all stubbed functions

oleg-dubinskiy:
## Purpose

Use `stdcall -stub` instead of `stub` and enable the functions's parameters, whose are already added as comments (except `ShellTurnOffDialog` export, since this function I already implemented in my previous #3254 PR, so I didn't touch that, to avoid the possible conflicts; please merge that first :pray: ).
Since those functions are completely undocumented, and unlikely will be implemented in ROS, stubbing them only in spec file seems to be a better solution than adding a (most of all) wrong prototypes.
This allows to properly load our msgina.dll in Windows XP/2003. :smiley: Although these systems still don't login into desktop with our dll replaced, but nevertheless the system crash which has been caused by stubbed functions does no longer happen.

JIRA issue: None.

## Result

Attached GIFs show the test of our msgina.dll on Windows Server 2003 Standard Edition SP2, before and after my changes.

Before:
ros_msgina_before

After:
ros_msgina_after


Saibamen:
Do you think we need to change all `stub` into `stdcall -stub`: https://github.com/reactos/reactos/search?q=%20stub%20 ?

Or create a proxy for `stub` to call `stdcall -stub` instead?

oleg-dubinskiy:
>Do you think we need to change all stub into stdcall -stub: https://github.com/reactos/reactos/search?q=%20stub%20 ?
Or create a proxy for stub to call stdcall -stub instead?

I guess it's not necessarily. In many cases, just `stub` is enough to fix the problem, so we don't need to use `stdcall -stub` everywhere. Moreover if the parameters of calling function are unknown.

PR #3257, opened 2020-10-03
[DBGHELP] Accept NULL symbol path as reset-to-default. CORE-17073

ThFabba:
Fixes heap corruption when running kernrate.
This is to collect some feedback before/while I send this upstream.

JIRA issue: [CORE-17073](https://jira.reactos.org/browse/CORE-17073)

### DIFF
--- a/dll/win32/dbghelp/dbghelp.c
+++ b/dll/win32/dbghelp/dbghelp.c
@@ -191,6 +191,36 @@ struct cpu* cpu_find(DWORD machine)
     return NULL;
 }
 
+static WCHAR* make_default_search_path(void)
+{
+    WCHAR*      search_path;
+    unsigned    size;
+    unsigned    len;
+
+    search_path = HeapAlloc(GetProcessHeap(), 0, (len = MAX_PATH) * sizeof(WCHAR));
+    while ((size = GetCurrentDirectoryW(len, search_path)) >= len)
+        search_path = HeapReAlloc(GetProcessHeap(), 0, search_path, (len *= 2) * sizeof(WCHAR));
+    search_path = HeapReAlloc(GetProcessHeap(), 0, search_path, (size + 1) * sizeof(WCHAR));
+
+    len = GetEnvironmentVariableW(L"_NT_SYMBOL_PATH", NULL, 0);
+    if (len)
+    {
+        search_path = HeapReAlloc(GetProcessHeap(), 0, search_path, (size + 1 + len + 1) * sizeof(WCHAR));
+        search_path[size] = ';';
+        GetEnvironmentVariableW(L"_NT_SYMBOL_PATH", search_path + size + 1, len);
+        size += 1 + len;
+    }
+    len = GetEnvironmentVariableW(L"_NT_ALTERNATE_SYMBOL_PATH", NULL, 0);
+    if (len)
+    {
+        search_path = HeapReAlloc(GetProcessHeap(), 0, search_path, (size + 1 + len + 1) * sizeof(WCHAR));
+        search_path[size] = ';';
+        GetEnvironmentVariableW(L"_NT_ALTERNATE_SYMBOL_PATH", search_path + size + 1, len);
+    }
+
+    return search_path;
+}
+
 /******************************************************************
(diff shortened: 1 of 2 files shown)


dll/win32/dbghelp/dbghelp.c
```diff
@@ -219,12 +256,18 @@ BOOL WINAPI SymSetSearchPath(HANDLE hProcess, PCSTR searchPath)
...
 
-    len = MultiByteToWideChar(CP_ACP, 0, searchPath, -1, NULL, 0);
-    if ((sp = HeapAlloc(GetProcessHeap(), 0, len * sizeof(WCHAR))))
+    if (searchPath)
     {
+        len = MultiByteToWideChar(CP_ACP, 0, searchPath, -1, NULL, 0);
+        sp = HeapAlloc(GetProcessHeap(), 0, len * sizeof(WCHAR));
+        if (!sp) return FALSE;
         MultiByteToWideChar(CP_ACP, 0, searchPath, -1, sp, len);
+    }
```
learn-more, line 265:
```suggestion
    }
    else
    {
        sp = NULL;
    }
```

JoachimHenze:
I do recommend to change in function SymSetSearchPath: the line
    WCHAR*      sp;
to
    WCHAR*      sp = NULL;

to avoid the following compiler warning, that I got only on carrier-releaser interestingly,

../dll/win32/dbghelp/dbghelp.c: In function 'SymSetSearchPath':
../dll/win32/dbghelp/dbghelp.c:244:9: error: 'sp' may be used uninitialized in this function [-Werror=uninitialized]

but I did *not* get that error on compilation when building this locally on "GCC4.7.2 RosBE2.1.6" or "GCC8.4.0 RosBE2.21", or "VS2010", or with the buildbits/testbots. Weird!
I have no clue, why carrier releaser does behave different here at all upon compilation!

Extravert-ir:
@JoachimHenze Mark's comment is about the same issue, and his solution is better :)

dll/win32/dbghelp/dbghelp.c
```diff
@@ -219,12 +256,18 @@ BOOL WINAPI SymSetSearchPath(HANDLE hProcess, PCSTR searchPath)
...
+
+    ret = SymSetSearchPathW(hProcess, sp);
 
-        ret = SymSetSearchPathW(hProcess, sp);
+    if (searchPath)
```
Extravert-ir, line 269:
```suggestion
    if (sp)
```

Extravert-ir (changes requested):

dll/win32/dbghelp/dbghelp.c
```diff
@@ -191,21 +191,68 @@ struct cpu* cpu_find(DWORD machine)
...
+    search_path = HeapAlloc(GetProcessHeap(), 0,
+                            (1 + sym_path_len + alt_sym_path_len + 1) * sizeof(WCHAR));
+    if (!search_path) return NULL;
+
+    p = search_path;
+    *p++ = L'.';
+    if (sym_path_len)
+    {
+        *p++ = L';';
+        GetEnvironmentVariableW(L"_NT_SYMBOL_PATH", p, sym_path_len);
```
learn-more, line 216: In theory there is a TOCTOU problem here, because if the env-var would be updated to be longer than `sym_path_len`, the result in the `p` buffer would be undefined according to msdn.
Not sure if this should be take into consideration at all, probably not tbh.
https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-getenvironmentvariable

learn-more (approved):

dll/win32/dbghelp/dbghelp.c
ThFabba: Yeah, I'd expect if somebody is messing with environment variables in another thread, it's the program's own fault. Especially since dbghelp documentation specifically says that it's single-threaded.
We won't crash or anything, so I think this is probably good enough. Could potentially zero the heap allocation so that we won't get uninitialized garbage in the buffer if one of the calls fails.

ThFabba:
Superseded by Wine import 4601d948013712974d9fd6990800c0ecfb36ed36

PR #3258, opened 2020-10-03
[NTDLL_APITEST] Check whether changing the product type in the regist…

tkreuzer:
Check whether changing the product type in the registry has an effect on RtlGetNtProductType

### DIFF
--- a/modules/rostests/apitests/ntdll/RtlGetNtProductType.c
+++ b/modules/rostests/apitests/ntdll/RtlGetNtProductType.c
@@ -68,11 +68,66 @@ ReturnNtProduct(PDWORD ProductNtType)
     return FALSE;
 }
 
+static
+BOOLEAN
+ChangeNtProductType(DWORD NtProductType)
+{
+    PWSTR ProductTypeString;
+    LONG Result;
+    HKEY Key;
+    DWORD dwSize;
+
+    if (NtProductType == NtProductWinNt)
+    {
+        ProductTypeString = L"WinNT";
+    }
+    else if (NtProductType == NtProductLanManNt)
+    {
+        ProductTypeString = L"LanmanNT";
+    }
+    else if (NtProductType == NtProductServer)
+    {
+        ProductTypeString = L"ServerNT";
+    }
+    else
+    {
+        ok(FALSE, "Passed invalid product type to CHangeNtProduct: %lu", NtProductType);
+    }
+
+    Result = RegOpenKeyExW(HKEY_LOCAL_MACHINE,
+                           L"SYSTEM\\CurrentControlSet\\Control\\ProductOptions",
+                           0,
+                           KEY_QUERY_VALUE,
+                           &Key);
+    if (Result != ERROR_SUCCESS)
+    {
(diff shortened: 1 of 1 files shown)


modules/rostests/apitests/ntdll/RtlGetNtProductType.c
```diff
@@ -94,4 +149,19 @@ START_TEST(RtlGetNtProductType)
     Ret = RtlGetNtProductType(&ProductType);
     ok(Ret == TRUE, "Expected a valid product type value (and TRUE as returned success code) but got %u as status.\n", Ret);
     ok(ProductNtType == ProductType, "Expected the product type value to be the same but got %lu (original value pointed by RtlGetNtProductType() is %d).\n", ProductNtType, ProductType);
+
+    /* CHange product type in the registry */
```
HBelusca, line 153:
```suggestion
    /* Change product type in the registry */
```

modules/rostests/apitests/ntdll/RtlGetNtProductType.c
```diff
@@ -68,11 +68,66 @@ ReturnNtProduct(PDWORD ProductNtType)
...
+                           0,
+                           KEY_QUERY_VALUE,
+                           &Key);
+    if (Result != ERROR_SUCCESS)
+    {
+        ok(FALSE, "RegOpenKeyExW failed with 0x%lx\n", Result);
+        return FALSE;
+    }
+ 
+    dwSize = ((DWORD)wcslen(ProductTypeString) + 1) * sizeof(WCHAR);
```
HBelusca, line 108:
```suggestion
    dwSize = (DWORD)((wcslen(ProductTypeString) + 1) * sizeof(WCHAR));
```
In your way the wcslen was DWORD converted, but the doubling via the sizeof could wrap the size value in case it would have exceeded the maximum value that could be contained in DWORD.
In the suggestion, the doubling is done on size_t value, then it's truncated down to DWORD.
(Pick whatever you prefer ^^)

modules/rostests/apitests/ntdll/RtlGetNtProductType.c
```diff
@@ -68,11 +68,66 @@ ReturnNtProduct(PDWORD ProductNtType)
...
+                            (PBYTE)ProductTypeString,
+                            dwSize);
+    if (Result != ERROR_SUCCESS)
+    {
+        ok(FALSE, "RegSetValueExW failed with 0x%lx\n", Result);
+        RegCloseKey(Key);
+        return FALSE;
+    }
+
+    RegCloseKey(Key);
```
HBelusca, line 122: Maybe move the RegCloseKey call just before checking for the Result?

modules/rostests/apitests/ntdll/RtlGetNtProductType.c
tkreuzer: LOL. That makes absolutely no difference.

HBelusca:
On Windows, isn't that registry subkey/value permanently monitored for any changes (and changes being reverted when they appear)? If so, the apitest wouldn't detect anything sensible 🤔

HBelusca (approved):

PR #458, opened 2018-04-01
[WINLOGON] Implement BaseInitiateSystemShutdownEx().

alpha5056:
## Purpose

Implement BaseInitiateSystemShutdownEx().

JIRA issue: [CORE-8568](https://jira.reactos.org/browse/CORE-8568)

## Proposed changes

Add a dialog resource which shuts down the computer after a specified timeout.

- Add dialog resource
- Make BaseInitiateSystemShutdownEx() start a new shutdown thread
- Create thread function that simply calls DialogBox()
- Implement countdown timer inside the DialogProc

## TODO

A couple FIXMEs:
- Whether the caller actually has permission is not verified
- The date/time format does not conform to the locale
- More translation work is needed

## Legal Notice

As this is my first contribution, I suppose this would be an appropriate place for the affirmation:
> I hereby swear that I have not used nor seen the source code to any version of the Windows operating system nor any Microsoft product that may be related to the proposed project that is under a license incompatible with contribution to ReactOS, including but not limited to the leaked Windows 2000 source code and the Windows Research Kernel.

PR #3259, opened 2020-10-03
[WINLOGON] Display the shutdown message popup dialog on the current input desktop.

HBelusca:
## Purpose

Display the shutdown message popup dialog on the current input desktop.

JIRA issue: [CORE-17050](https://jira.reactos.org/browse/CORE-17050)

## Proposed changes

Display the shutdown message popup dialog on the current input desktop,
and periodically monitor for any change of the input desktop. When the
latter changes, close the dialog and recreate it on the new input desktop.

In addition, retain the current dialog position and restore it when the
dialog is recreated on the new desktop.

### DIFF
--- a/base/system/winlogon/shutdown.c
+++ b/base/system/winlogon/shutdown.c
@@ -3,7 +3,8 @@
  * PROJECT:         ReactOS Winlogon
  * FILE:            base/system/winlogon/shutdown.c
  * PURPOSE:         System shutdown dialog
- * PROGRAMMERS:     alpha5056 <alpha5056@users.noreply.github.com>
+ * PROGRAMMERS:     Edward Bronsten <tanki.alpha5056@gmail.com>
+ *                  Eric Kohl
  */
 
 /* INCLUDES ******************************************************************/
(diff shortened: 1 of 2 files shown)


base/system/winlogon/shutdown.c
```diff
@@ -4,6 +4,7 @@
  * FILE:            base/system/winlogon/shutdown.c
  * PURPOSE:         System shutdown dialog
  * PROGRAMMERS:     alpha5056 <alpha5056@users.noreply.github.com>
```
binarymaster, line 6:
```suggestion
 * PROGRAMMERS:     Edward Bronsten <tanki.alpha5056@gmail.com>
```
Extracted author data from PR #458, you probably can fix this in formatting commit.

JoachimHenze (approved):
I do confirm that the first 2 commits of this PR (after a155c28) applied on top of 0.4.15-dev-1086-g5bb36e9 does properly fix CORE-17050.
I also checked: neither CORE-11933 nor CORE-16548 gets reintroduced by it.
Fine.
My only critics is unneeded beautification being mixed with the relevant regression-fix, which may lead to backporting being more complicated/risky than needed.

binarymaster (approved):

PR #3261, opened 2020-10-04
[NTOS:KE] KiInitializePcr(): Zero the actual PCR size, not PAGE_SIZE

SergeGautherie:
## Purpose

Better safe than sorry: no overrun.

## Proposed changes

For example, on i386: 4064 < 4096.

Follow-up to 597c140 (r55405).


ThFabba:
Your definition of "safe" is broken. What's safe here is to zero the entire page, because not doing so assumes that somebody else has already done it (which is probably true, but unless you can point to the code that did, you're guessing, which is not acceptable).

PR #3262, opened 2020-10-05
[KEYBOARD][NTDLL_APITEST] Remove 'set_property(... C_STANDARD 99)'

SergeGautherie:
## Purpose

Cleanup.

## Proposed changes

C99 is the default now.

Follow-up to 95483b4.


Extravert-ir:
This kind of changes are not accepted without something meaningful

SergeGautherie:
> something meaningful

What do you mean?

Extravert-ir:
This is a cosmetic change without any other purpose. We don't accept such changes if they are not from maintainers

PR #3264, opened 2020-10-05
[BROWSEUI] Remove '\1' from AutoComplete list

katahiromz:
## Purpose

Delete "backslash one" (indicates `SW_` values) from auto-completion list.

BEFORE:
before

AFTER:
after

### DIFF
--- a/dll/win32/browseui/ACLCustomMRU.cpp
+++ b/dll/win32/browseui/ACLCustomMRU.cpp
@@ -35,7 +35,7 @@ STDMETHODIMP CACLCustomMRU::Next(ULONG celt, LPWSTR *rgelt, ULONG *pceltFetched)
 
     CStringW str = m_MRUData[m_ielt];
 
-    if (!m_bTypedURLs && !PathFileExistsW(str))
+    if (!m_bTypedURLs)
     {
         // Erase the last "\\1" etc. (indicates SW_* value)
         INT ich = str.ReverseFind(L'\\');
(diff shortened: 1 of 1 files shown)


dll/win32/browseui/ACLCustomMRU.cpp
```diff
@@ -33,12 +33,22 @@ STDMETHODIMP CACLCustomMRU::Next(ULONG celt, LPWSTR *rgelt, ULONG *pceltFetched)
...
-    size_t cb = (m_MRUData[m_ielt].GetLength() + 1) * sizeof(WCHAR);
+    CStringW str = m_MRUData[m_ielt];
+
+    if (!m_bTypedURLs)
+    {
+        // Erase the last "\\1" etc. (indicates SW_* value)
+        INT ich = str.ReverseFind(L'\\');
+        if (ich >= 0)
+            str = str.Left(ich);
+    }
```
HBelusca, line 44: Suppose that the value is instead: `some_directory\notepad.exe` (instead of, `notepad.exe\1`). How would this situation behave in Windows? Would it work as well with your new modification?

katahiromz: I confirmed Windows was smarter than me. I added file existence check. OK.

katahiromz: Case 1:
case1
Case 2:
case2

katahiromz: Case 1 contains `"\1"`. That works well.
Case 2 doesn't contain `"\1"`. That is merged with `"C:\Windows"`. (There is no `C:\WinDows`)

HBelusca: Windows is case insensitive by default, so `Windows` or `WinDows` is the same for it.

katahiromz: Reverted.

PR #3265, opened 2020-10-05
[WIN32SS] Fix CF_DIB format not being saved to clipboard on Print Screen key

Doug-Lyons:
##  Send CF_DIB data to clipboard when the PrntScrn key is hit

_Add memory allocation for CF_DIB formatted data on clipboard_

JIRA issue: [CORE-17318](https://jira.reactos.org/browse/CORE-17318)

## Add code to Allocate and Free pool memory for use by CF_DIB formatted data

### DIFF
--- a/win32ss/user/ntuser/object.c
+++ b/win32ss/user/ntuser/object.c
@@ -221,7 +221,9 @@ static PVOID AllocSysObjectCB(
     ASSERT(Size > sizeof(HEAD));
 
     /* Allocate the clipboard data */
-    Object = ExAllocatePoolZero(PagedPool, Size, USERTAG_CLIPBOARD);
+    // FIXME: This allocation should be done on the current session pool;
+    // however ReactOS' MM doesn't support session pool yet.
+    Object = ExAllocatePoolZero(/* SESSION_POOL_MASK | */ PagedPool, Size, USERTAG_CLIPBOARD);
     if (!Object)
     {
         ERR("ExAllocatePoolZero failed. No object created.\n");
--- a/win32ss/user/ntuser/object.c
+++ b/win32ss/user/ntuser/object.c
@@ -207,10 +207,43 @@ static PVOID AllocSysObject(
     return Object;
 }
 
+_Success_(return!=NULL)
+static PVOID AllocSysObjectCB(
+    _In_ PDESKTOP pDesk,
+    _In_ PTHREADINFO pti,
+    _In_ SIZE_T Size,
+    _Out_ PVOID* ObjectOwner)
+{
+    PVOID Object;
+
+    UNREFERENCED_PARAMETER(pDesk);
+    UNREFERENCED_PARAMETER(pti);
+    ASSERT(Size > sizeof(HEAD));
+
+    /* Allocate the clipboard data */
+    Object =  ExAllocatePoolWithTag(PagedPool, Size, USERTAG_CLIPBOARD);
+
+    if (!Object)
+    {
+        ERR("Object is NULL-No Object Created.\n");
+        return NULL;
+    }
+
+    *ObjectOwner = NULL;
+    RtlZeroMemory(Object, Size);
+    return Object;
+}
+
 static void FreeSysObject(
     _In_ PVOID Object)
 {
     UserHeapFree(Object);
+}
+
+static void FreeSysObjectCB(
+    _In_ PVOID Object)
+{
+    ExFreePoolWithTag(Object, USERTAG_CLIPBOARD);
 }
 
 static const struct
@@ -226,7 +259,7 @@ static const struct
     { AllocProcMarkObject,      IntDestroyCurIconObject,    FreeCurIconObject },    /* TYPE_CURSOR */
     { AllocSysObject,           /*UserSetWindowPosCleanup*/NULL, FreeSysObject },   /* TYPE_SETWINDOWPOS */
     { AllocDeskThreadObject,    IntRemoveHook,              FreeDeskThreadObject }, /* TYPE_HOOK */
-    { AllocSysObject,           /*UserClipDataCleanup*/NULL,FreeSysObject },        /* TYPE_CLIPDATA */
+    { AllocSysObjectCB,         /*UserClipDataCleanup*/NULL,FreeSysObjectCB },      /* TYPE_CLIPDATA */
     { AllocDeskProcObject,      DestroyCallProc,            FreeDeskProcObject },   /* TYPE_CALLPROC */
     { AllocProcMarkObject,      UserDestroyAccelTable,      FreeProcMarkObject },   /* TYPE_ACCELTABLE */
     { NULL,                     NULL,                       NULL },                 /* TYPE_DDEACCESS */
--- a/win32ss/user/ntuser/object.c
+++ b/win32ss/user/ntuser/object.c
@@ -221,16 +221,14 @@ static PVOID AllocSysObjectCB(
     ASSERT(Size > sizeof(HEAD));
 
     /* Allocate the clipboard data */
-    Object = ExAllocatePoolWithTag(PagedPool, Size, USERTAG_CLIPBOARD);
-
+    Object = ExAllocatePoolZero(PagedPool, Size, USERTAG_CLIPBOARD);
     if (!Object)
     {
-        ERR("ExAllocatePoolWithTag failed, no object created.\n");
+        ERR("ExAllocatePoolZero failed. No object created.\n");
         return NULL;
     }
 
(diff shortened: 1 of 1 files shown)


win32ss/user/ntuser/object.c
```diff
@@ -207,12 +207,45 @@ static PVOID AllocSysObject(
...
+    _Out_ PVOID* ObjectOwner)
+{
+    PVOID Object;
+
+    UNREFERENCED_PARAMETER(pDesk);
+    UNREFERENCED_PARAMETER(pti);
+    ASSERT(Size > sizeof(HEAD));
+
+    /* Allocate the clipboard data */
+    Object =  ExAllocatePoolWithTag(PagedPool, Size, USERTAG_CLIPBOARD);
```
binarymaster, line 224:
```suggestion
    Object = ExAllocatePoolWithTag(PagedPool, Size, USERTAG_CLIPBOARD);
```
Whitespace fix.

win32ss/user/ntuser/object.c
```diff
@@ -207,12 +207,45 @@ static PVOID AllocSysObject(
...
+
+    if (!Object)
+    {
+        ERR("Object is NULL-No Object Created.\n");
```
binarymaster, line 228:
```suggestion
        ERR("ExAllocatePoolWithTag failed, no object created.\n");
```
I've had hard time trying to understand this text :sweat_smile:

win32ss/user/ntuser/object.c
Doug-Lyons: Corrected. Thanks.

win32ss/user/ntuser/object.c
Doug-Lyons: Improved. Thanks.

win32ss/user/ntuser/object.c
```diff
@@ -207,12 +207,45 @@ static PVOID AllocSysObject(
...
+    PVOID Object;
+
+    UNREFERENCED_PARAMETER(pDesk);
+    UNREFERENCED_PARAMETER(pti);
+    ASSERT(Size > sizeof(HEAD));
+
+    /* Allocate the clipboard data */
+    Object = ExAllocatePoolWithTag(PagedPool, Size, USERTAG_CLIPBOARD);
+
+    if (!Object)
```
HBelusca, line 226:
```suggestion
    Object = ExAllocatePoolZero(PagedPool, Size, USERTAG_CLIPBOARD);
    if (!Object)
```

win32ss/user/ntuser/object.c
```diff
@@ -207,12 +207,45 @@ static PVOID AllocSysObject(
...
+    {
+        ERR("ExAllocatePoolWithTag failed, no object created.\n");
+        return NULL;
+    }
+
+    *ObjectOwner = NULL;
+    RtlZeroMemory(Object, Size);
```
HBelusca, line 233:
```suggestion
```
Now unneeded thanks to the usage of `ExAllocatePoolZero` :)

win32ss/user/ntuser/object.c
```diff
@@ -207,12 +207,45 @@ static PVOID AllocSysObject(
...
+    UNREFERENCED_PARAMETER(pDesk);
+    UNREFERENCED_PARAMETER(pti);
+    ASSERT(Size > sizeof(HEAD));
+
+    /* Allocate the clipboard data */
+    Object = ExAllocatePoolWithTag(PagedPool, Size, USERTAG_CLIPBOARD);
+
+    if (!Object)
+    {
+        ERR("ExAllocatePoolWithTag failed, no object created.\n");
```
HBelusca, line 228:
```suggestion
        ERR("ExAllocatePoolZero failed, no object created.\n");
```

HBelusca (commented):
I find it strange, though, that the simple change of memory allocator "fixes" the problem. Can you explain to me why the other one, that uses UserHeapAlloc (allocates an object on the win32 desktop heap so that it can also be accessed in user-mode), didn't work?

Doug-Lyons:
> I find it strange, though, that the simple change of memory allocator "fixes" the problem. Can you explain to me why the other one, that uses UserHeapAlloc (allocates an object on the win32 desktop heap so that it can also be accessed in user-mode), didn't work?

@HBelusca, This starts in keyboard.c in function SnapWindow which calls UserCloseClipboard which then calls IntAddSynthesizedFormats. Now if you look at the code in clipboard.c and the function IntAddSynthesizedFormats you will see that there is a call to UserCreateObject. This calls ObjectCallbacks[type].ObjectAlloc which is mapped to AllocSysObject. This eventually calls UserHeapAlloc with the Size of the object. Say we are using 800 x 600, then this will be 800 x 600 x 4 (words per pixel) which is almost 2 MB. If you follow UserHeapAlloc this goes through RtlAllocateHeap referencing the GlobalUserHeap and the same Size. In RtlAllocateHeap the Size is compared through a series of 3 if statements as Index which is the Size/8. The first is to see if this is less than HEAP_FREELISTS (0x80). The next is to see if this is less than Heap->VirtualMemoryThreshold (0xfe00) and the third is to see if the heap is growable (the GlobalUserHeap is not). Our need is for 1,920,000 bytes / 8 (0x3a980), so we are not less than either of the first two numbers and not growable, so this allocation request fails causing the whole chain to fail.
Hopefully, this answers your question. There is another problem here with the original CF_BITMAP allocation where a handle to the bitmap (hbm) is passed to the clipboard to be used for the clipboard data, but this is never reserved in memory. I believe that this leads to problems such as we see in CORE-16586, but this has not been addressed yet. Thanks.

win32ss/user/ntuser/object.c
Doug-Lyons: Changed. Thanks.

win32ss/user/ntuser/object.c
Doug-Lyons: Removed now because of different call. Thanks.

win32ss/user/ntuser/object.c
Doug-Lyons: Revised Working to reflect different call and better English. Thanks.

encodedpr (approved):

katahiromz:
Option 1: Use `ExAllocatePoolWithTag` (or `ExAllocatePoolZero`) instead of `UserHeapAlloc` and `RtlAllocateHeap`.
Option 2: Use `HEAP_GROWABLE` flag for `RtlAllocateHeap`.

Doug-Lyons:
> Option 1: Use `ExAllocatePoolWithTag` (or `ExAllocatePoolZero`) instead of `UserHeapAlloc` and `RtlAllocateHeap`.
> Option 2: Use `HEAP_GROWABLE` flag for `RtlAllocateHeap`.

@katahiromz, I did use 'ExAllocatePoolZero' in my code. Did I mistake your meaning here? Thanks for your comments. Doug Lyons

katahiromz:
@HBelusca How do you think the difference of `ExAllocatePoolWithTag` and `RtlAllocateHeap` with `HEAP_GROWABLE` flag?

Doug-Lyons:
> @HBelusca How do you think the difference of `ExAllocatePoolWithTag` and `RtlAllocateHeap` with `HEAP_GROWABLE` flag?

@katahiromz, eventually you run into this in win32ss\user\ntuser\main.c
```
/* Create the global USER heap */
    GlobalUserHeap = UserCreateHeap(&GlobalUserHeapSection,
                                    &GlobalUserHeapBase,
                                    1 * 1024 * 1024); /* FIXME: 1 MB for now... */
```
We need more memory than this for our larger screens.

HBelusca (approved):

win32ss/user/ntuser/object.c
```diff
@@ -207,12 +207,43 @@ static PVOID AllocSysObject(
...
+    _Out_ PVOID* ObjectOwner)
+{
+    PVOID Object;
+
+    UNREFERENCED_PARAMETER(pDesk);
+    UNREFERENCED_PARAMETER(pti);
+    ASSERT(Size > sizeof(HEAD));
+
+    /* Allocate the clipboard data */
+    Object = ExAllocatePoolZero(PagedPool, Size, USERTAG_CLIPBOARD);
```
HBelusca, line 224:
```suggestion
    // FIXME: This allocation should be done on the current session pool;
    // however ReactOS' MM doesn't support session pool yet.
    Object = ExAllocatePoolZero(/* SESSION_POOL_MASK | */ PagedPool, Size, USERTAG_CLIPBOARD);
```

Doug-Lyons:
@HBelusca, Thanks for your help in  committing this.

PR #3266, opened 2020-10-05
[BOOTDATA] Comment out biosinfo.inf

archeYR:
This entry is unused as there is no biosinfo.inf anywhere yet, and it only breaks boot with NTLDR

Before:
vmplayer_sCAi8sUNEl

After:
vmplayer_3VBA87p7UE

### DIFF
--- a/boot/bootdata/hivesys.inf
+++ b/boot/bootdata/hivesys.inf
@@ -5,7 +5,7 @@ Signature = "$Windows NT$"
 
 HKLM,"SYSTEM\CurrentControlSet\Control","CurrentUser",2,"USERNAME"
 HKLM,"SYSTEM\CurrentControlSet\Control","WaitToKillServiceTimeout",2,"20000"
-HKLM,"SYSTEM\CurrentControlSet\Control\Biosinfo","InfName",2,"biosinfo.inf"
+;HKLM,"SYSTEM\CurrentControlSet\Control\Biosinfo","InfName",2,"biosinfo.inf"
 HKLM,"SYSTEM\CurrentControlSet\Control\PnP",,0x00000012
 HKLM,"SYSTEM\CurrentControlSet\Control\Video",,0x00000012
 


binarymaster:
By the way, how `biosinfo.inf` file is generated?

Zero3K:
biosinfo.inf isn't generated. Its included in a Windows Server 2003 install. I am attaching it as an archive.

[biosinfo.zip](https://github.com/reactos/reactos/files/5328025/biosinfo.zip)

HBelusca (approved):
I'm fine with this; when the framework for handling the HW workarounds using this file become implemented, we can uncomment it back.

boot/bootdata/hivesys.inf
```diff
@@ -5,7 +5,7 @@ Signature = "$Windows NT$"
 
 HKLM,"SYSTEM\CurrentControlSet\Control","CurrentUser",2,"USERNAME"
 HKLM,"SYSTEM\CurrentControlSet\Control","WaitToKillServiceTimeout",2,"20000"
-HKLM,"SYSTEM\CurrentControlSet\Control\Biosinfo","InfName",2,"biosinfo.inf"
+;HKLM,"SYSTEM\CurrentControlSet\Control\Biosinfo","InfName",2,"biosinfo.inf"
```
SergeGautherie, line 8:
```suggestion
; HKLM,"SYSTEM\CurrentControlSet\Control\Biosinfo","InfName",2,"biosinfo.inf" ; UNIMPLEMENTED
```

HBelusca: You won't start adding UNIMPLEMENTED to all commented lines in all .inf files?

SergeGautherie: > `git blame`

(It's a pain, especially as https://git.reactos.org/ does not support it.) Anyway.

PR #3267, opened 2020-10-05
Repro sync 10-5-2020

minecraftchest1:
Repro sync


minecraftchest1:
Woopse wring direction.

PR #3271, opened 2020-10-06
[NTOS:MM] ARM3/sysldr.c: Fix string length handling

SergeGautherie:
## Purpose

Be consistent: work right.

## Proposed changes

* Always set/use consistent Length and MaximumLength.
* Set intended UNICODE_NULL on a FullDllName.

### DIFF
--- a/ntoskrnl/mm/ARM3/sysldr.c
+++ b/ntoskrnl/mm/ARM3/sysldr.c
@@ -822,11 +822,9 @@ MiSnapThunk(IN PVOID DllBase,
             /* Now assume failure in case the forwarder doesn't exist */
             Status = STATUS_DRIVER_ENTRYPOINT_NOT_FOUND;
 
-            /* Build the forwarder name */
+            /* Build the forwarder name, including the dot */
             DllName.Buffer = (PCHAR)Address->u1.Function;
-            DllName.Length = (USHORT)(strchr(DllName.Buffer, '.') -
-                                      DllName.Buffer) +
-                                      sizeof(ANSI_NULL);
+            DllName.Length = (USHORT)(strchr(DllName.Buffer, '.') - DllName.Buffer + strlen("."));
             DllName.MaximumLength = DllName.Length;
 
             /* Convert it */
@@ -869,7 +867,7 @@ MiSnapThunk(IN PVOID DllBase,
                                                         TAG_LDR_WSTR);
                     if (!ForwardName) break;
 
-                    /* Copy the data */
+                    /* Copy the data, including the ANSI_NULL */
                     RtlCopyMemory(&ForwardName->Name[0],
                                   DllName.Buffer + DllName.Length,
                                   ForwardLength);
@@ -1175,6 +1173,7 @@ CheckDllState:
         if (!ImportBase)
         {
             /* Setup the import DLL name */
+            DllName.Length = 0;
             DllName.MaximumLength = NameString.Length +
                                     ImageFileDirectory->Length +
                                     sizeof(UNICODE_NULL);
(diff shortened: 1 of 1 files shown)


ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -822,11 +822,9 @@ MiSnapThunk(IN PVOID DllBase,
             /* Now assume failure in case the forwarder doesn't exist */
             Status = STATUS_DRIVER_ENTRYPOINT_NOT_FOUND;
 
-            /* Build the forwarder name */
+            /* Build the forwarder name, including the dot */
```
HBelusca, line 825: Why "including the dot"? The forwarder name is the name of the dll without the extension (i.e. without the first dot and what follows it) (Aside note: actually the extension should start at the latest dot, so I don't know whether using the strchr here is intended or not, i.e. whether windows does this or not).

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -822,11 +822,9 @@ MiSnapThunk(IN PVOID DllBase,
...
             DllName.Buffer = (PCHAR)Address->u1.Function;
-            DllName.Length = (USHORT)(strchr(DllName.Buffer, '.') -
-                                      DllName.Buffer) +
-                                      sizeof(ANSI_NULL);
+            DllName.Length = (USHORT)(strchr(DllName.Buffer, '.') - DllName.Buffer + strlen("."));
```
HBelusca, line 827: Wrong.

SergeGautherie: What is wrong?

ntoskrnl/mm/ARM3/sysldr.c
SergeGautherie: I am not sure what your point is: I am simply documenting existing code (with regard to strings), which seems consistent as far as I looked.
If you think it does not behave as it should, file an issue...

ntoskrnl/mm/ARM3/sysldr.c
tkreuzer: Changing kernel code for the sake of changing it? using strlen here is retarded.

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -2255,17 +2255,19 @@ MiInitializeLoadedModuleList(IN PLOADER_PARAMETER_BLOCK LoaderBlock)
         }
 
         /* Set the base name */
+        // NewEntry->BaseDllName.Length = LdrEntry->BaseDllName.Length; // Already.
+        NewEntry->BaseDllName.MaximumLength = NewEntry->BaseDllName.Length + sizeof(UNICODE_NULL);
         NewEntry->BaseDllName.Buffer = (PVOID)(NewEntry + 1);
 
-        /* Copy the full and base name */
+        /* Copy and null-terminate the full and base names */
```
tkreuzer, line 2262: You seem to be determined to change as many lines in the kernel as possible. I really don't want git blame to one day tell me that you wrote the entire kernel.

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -3275,6 +3274,7 @@ MmLoadSystemImage(IN PUNICODE_STRING FileName,
             UnicodeTemp = PrefixName;
             UnicodeTemp.Buffer += 11;
             UnicodeTemp.Length -= (11 * sizeof(WCHAR));
+            UnicodeTemp.MaximumLength -= (11 * sizeof(WCHAR));
```
tkreuzer, line 3277: Pointless.

tkreuzer (changes requested):
I don't see anything in here that fixes anything, but plenty of random changes with the potential to cause subtle breakages.

Extravert-ir:
The purpose of this PR is not clear: what does it fix?

PR #3274, opened 2020-10-06
[NTOS:MM] ARM3/sysldr.c: Improve 3 format strings

SergeGautherie:
## Purpose

Be consistent and explicit: not confusing.

## Proposed changes

Improve 3 format strings

### DIFF
--- a/ntoskrnl/mm/ARM3/sysldr.c
+++ b/ntoskrnl/mm/ARM3/sysldr.c
@@ -173,7 +173,7 @@ MiLoadImageSection(IN OUT PVOID *SectionPtr,
 
     /* The driver is here */
     *ImageBase = DriverBase;
-    DPRINT1("Loading: %wZ at %p with %lx pages\n", FileName, DriverBase, PteCount);
+    DPRINT1("Loading: %wZ at %p with 0x%lx pages\n", FileName, DriverBase, PteCount);
 
     /* Lock the PFN database */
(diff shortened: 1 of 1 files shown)


Extravert-ir:
I always wondered how do you choose a place for your "be consistent" fixes?
There are hundreds of other places with the same "inconsistency". Why don't you fix them? Does it make sense to fix only 2 out of hundreds places like that?

HBelusca:
I was about to write the same comment ^

SergeGautherie:
> I always wondered how do you choose a place for your "be consistent" fixes?

I don't "choose", I "notice":
for example, '10 pages' here...

> There are hundreds of other places with the same "inconsistency". Why don't you fix them? Does it make sense to fix only 2 out of hundreds places like that?

If you know these places, why don't you tell where or even fix them?
Aren't you aware that I repeatedly get orders to make smaller PRs?
If I understand correctly, your hint to fix more is to start by not fixing any: that does not make sense to me.

tkreuzer:
If you are concerned with our format strings, focus on those that create warnings or are clearly misleading, because nobody cares for anything else. And if you do, please cover a larger set than 1 file, because it's a waste of time. If your PR was something like "Fix 23 format warnings in the kernel", people like me might actually look into it. This is a waste of time for many people, who could spend their time way better with fixing actual issues.

PR #3277, opened 2020-10-06
[RAPPS] (Q.O.L.) Alphabetical order categories.

Mou-Ikkai:
## Purpose

_As the title suggests, this puts all categories into alphabetical order... Except 'Other' is at the bottom. If wanted, I can put it in alphabetical as well._

## Proposed changes

- I find it a bit... Unneeded to have categories to be scattered, and I prefer a clean list.
- This is mainly a QOL/Optional change. And yes, this does work.


binarymaster:
> As the title suggests, this puts all categories into alphabetical order... 

uhm... what about different languages (translations)?

Mou-Ikkai:
> > As the title suggests, this puts all categories into alphabetical order...
> 
> uhm... what about different languages (translations)?

That one, I can't get a grasp on... Sorting alphabetically with each language would probably require a script to sort them by their letters... That one I don't even know I can do. Since it's all a string table...

learn-more:
Yeah, this might work for english but breaks it again for another language.

PR #3278, opened 2020-10-07
[NTOS:MM] MiWriteProtectSystemImage(): Leave image header as R/W, on …

SergeGautherie:
…non-Debug builds

## Purpose

Minimal workaround, compared to 'MmEnforceWriteProtection = FALSE' until now.

JIRA issue: [CORE-16387](https://jira.reactos.org/browse/CORE-16387)

## TODO

- [ ] Test whether this hack actually works around this issue. (I can't do it.)


JoachimHenze:
Closed. See PR3280 for the why. This is not better than how we used to workaround.

SergeGautherie:
> This is not better than how we used to workaround.

Well, do as you prefer, though I don't understand how disabling the whole feature is better than a small part only.

PR #3280, opened 2020-10-07
[NTOS:MM] MmLoadSystemImage(): Leave whole image as R/W, on non-Debug…

SergeGautherie:
… builds

## Purpose

Minimal workaround, compared to 'MmEnforceWriteProtection = FALSE' until now.

JIRA issue: [CORE-16449](https://jira.reactos.org/browse/CORE-16449)

## Proposed changes

And add a couple DPRINT1(), to automatically log data.

### DIFF
--- a/ntoskrnl/mm/ARM3/sysldr.c
+++ b/ntoskrnl/mm/ARM3/sysldr.c
@@ -16,6 +16,10 @@
 #define MODULE_INVOLVED_IN_ARM3
 #include <mm/ARM3/miarm.h>
 
+#if !defined(CORE_16449_IS_FIXED) && DBG
+static BOOLEAN MWPSI_IsCore16449 = FALSE;
+#endif
+
 static
 inline
 VOID
@@ -2533,6 +2537,13 @@ MiWriteProtectSystemImage(
         Protection = (Section->Characteristics & IMAGE_SCN_PROTECTION_MASK);
 
         /* Update the protection for this section */
+#if !defined(CORE_16449_IS_FIXED) && DBG
+    if (MWPSI_IsCore16449)
+    {
+        DPRINT1("(CORE-16449) Section %lu, Calling MiSetSystemCodeProtection(%p, %p, 0x%08lx)\n",
+                i, FirstPte, LastPte, Protection);
+    }
+#endif
         MiSetSystemCodeProtection(FirstPte, LastPte, Protection);
     }
 
@@ -3253,12 +3264,16 @@ LoaderScan:
     if (wcscmp(LdrEntry->BaseDllName.Buffer, L"ftfd.dll"  ) == 0 ||
         wcscmp(LdrEntry->BaseDllName.Buffer, L"win32k.sys") == 0)
     {
+        MWPSI_IsCore16449 = TRUE;
         DPRINT1("(CORE-16449) Calling MiWriteProtectSystemImage(%p) for %wZ\n",
                 LdrEntry->DllBase, &LdrEntry->FullDllName);
     }
 #endif
 #if defined(CORE_16449_IS_FIXED) || DBG
     MiWriteProtectSystemImage(LdrEntry->DllBase);
+#endif
+#if !defined(CORE_16449_IS_FIXED) && DBG
+    MWPSI_IsCore16449 = FALSE;
 #endif
 
     /* Check if notifications are enabled */
--- a/ntoskrnl/mm/ARM3/sysldr.c
+++ b/ntoskrnl/mm/ARM3/sysldr.c
@@ -3249,7 +3249,9 @@ LoaderScan:
     /* FIXME: Call driver verifier's loader function */
 
     /* Write-protect the system image */
+#if defined(CORE_16449_IS_FIXED) || DBG
     MiWriteProtectSystemImage(LdrEntry->DllBase);
+#endif
 
     /* Check if notifications are enabled */
     if (PsImageNotifyEnabled)
--- a/ntoskrnl/mm/ARM3/sysldr.c
+++ b/ntoskrnl/mm/ARM3/sysldr.c
@@ -3250,8 +3250,12 @@ LoaderScan:
 
     /* Write-protect the system image */
 #if !defined(CORE_16449_IS_FIXED) && DBG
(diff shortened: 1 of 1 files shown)


ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -3242,7 +3257,13 @@ MmLoadSystemImage(IN PUNICODE_STRING FileName,
     /* FIXME: Call driver verifier's loader function */
 
     /* Write-protect the system image */
+#if !defined(CORE_16449_IS_FIXED) && DBG
+    DPRINT1("(CORE-16449) Calling MiWriteProtectSystemImage() for %wZ (%wZ)\n",
+            &LdrEntry->BaseDllName, &LdrEntry->FullDllName);
+#endif
+#if defined(CORE_16449_IS_FIXED) || DBG
```
tkreuzer, line 3264: Why only calling this on DBG? The code above should work on release builds.

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -2409,7 +2409,22 @@ MiSetSystemCodeProtection(
         TempPte = *PointerPte;
 
         /* Make sure it's valid */
+#if !defined(CORE_16449_IS_FIXED) && DBG
```
tkreuzer, line 2412: Why special handling for DBG? Why not just always DPRINT1 and continue?

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -3242,7 +3257,13 @@ MmLoadSystemImage(IN PUNICODE_STRING FileName,
     /* FIXME: Call driver verifier's loader function */
 
     /* Write-protect the system image */
+#if !defined(CORE_16449_IS_FIXED) && DBG
+    DPRINT1("(CORE-16449) Calling MiWriteProtectSystemImage() for %wZ (%wZ)\n",
```
tkreuzer, line 3261: Full name is enough, no need to print the base name. I'd rather print the image base.

ntoskrnl/mm/ARM3/sysldr.c
SergeGautherie: 'I will never let that protection feature reach any release until [...]' (c) Joachim

ntoskrnl/mm/ARM3/sysldr.c
SergeGautherie: (Because I often get complains about touching code/behavior I should not... That was before Thomas hijacked this part.)

ntoskrnl/mm/ARM3/sysldr.c
SergeGautherie: Changed.

ntoskrnl/mm/ARM3/sysldr.c
JoachimHenze: >> if (wcscmp(LdrEntry->BaseDllName.Buffer, L"ftfd.dll"  ) == 0 || wcscmp(LdrEntry->BaseDllName.Buffer, L"win32k.sys") == 0)
Woah is that horrible! And what about pfmodnt?
Do you really think it's beneficial to add such ugly exception-per-driver-hacks into the kernel?

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -3249,7 +3260,21 @@ MmLoadSystemImage(IN PUNICODE_STRING FileName,
     /* FIXME: Call driver verifier's loader function */
 
     /* Write-protect the system image */
+#if !defined(CORE_16449_IS_FIXED) && DBG
+    if (wcscmp(LdrEntry->BaseDllName.Buffer, L"ftfd.dll"  ) == 0 ||
+        wcscmp(LdrEntry->BaseDllName.Buffer, L"win32k.sys") == 0)
```
learn-more, line 3265: This is a terrible hack, but at least use the unicode functions?

learn-more: (`RtlEqualUnicodeString` or the likes)

ntoskrnl/mm/ARM3/sysldr.c
SergeGautherie: > horrible

??

> And what about pfmodnt?

What is 'pfmodnt'?

> Do you really think it's beneficial to add such ugly exception-per-driver-hacks into the kernel?

On one PR I get complain about adding too much DPRINT1, on another I get complain about not adding enough: usual nonsense, as I see it...

Why don't you submit a better PR to investigate and fix CORE-16449 (for example), rather than saying 'no' so often?

ntoskrnl/mm/ARM3/sysldr.c
SergeGautherie: > unicode functions

That seems needlessly complicated here to me, though I could, if you confirm.

ntoskrnl/mm/ARM3/sysldr.c
binarymaster: > usual nonsense, as I see it...

Do you know what the real nonsense is? I will tell you:

- You have [**53 open PRs**](https://github.com/reactos/reactos/pulls/SergeGautherie) in the repo, is this normal for just one person?
- As a result, all our existing manpower is being ~~used~~ abused for reviewing your mostly low-quality and untested Pull Requests

> Why don't you submit a better PR to investigate and fix

And after all this you even dare to ask somebody to spend even more time? Hopefully some of devs already decided to ignore any PRs and changes done by you, which includes me (_you might have noticed I stopped adding labels to your PRs_), so we still have some manpower to handle PRs from other people.

The more PRs you open, the more I think you do it intentionally to just abuse our manpower and ruin the project. So at this point the best thing you can do to save the project is to **stop sending any PRs**.

After this your comment I really want to kick you from our repository, but I'm not going to close all your PRs because this will result in spamming all devs e-mails, and so I don't recommend anyone doing this including you. Just leave. Thanks for understanding.

Extravert-ir:
Such workaround is not acceptable. `MmEnforceWriteProtection = FALSE` is better than this

PR #3281, opened 2020-10-07
[NTOS:MM] MiWriteProtectSystemImage(): Add an explicit text to a DPRI…

SergeGautherie:
…NT1()

## Purpose

Draw more attention to this case.

JIRA issue: [CORE-16387](https://jira.reactos.org/browse/CORE-16387)

### DIFF
--- a/ntoskrnl/mm/ARM3/sysldr.c
+++ b/ntoskrnl/mm/ARM3/sysldr.c
@@ -2533,7 +2533,7 @@ MiWriteProtectSystemImage(
     if (ALIGN_UP_POINTER_BY(SectionEnd, PAGE_SIZE) !=
         Add2Ptr(ImageBase, NtHeaders->OptionalHeader.SizeOfImage))
     {
-        DPRINT1("ImageBase 0x%p ImageSize 0x%lx Section %u VA 0x%lx Raw 0x%lx virt 0x%lx\n",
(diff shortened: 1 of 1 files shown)


ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -2533,7 +2533,7 @@ MiWriteProtectSystemImage(
     if (ALIGN_UP_POINTER_BY(SectionEnd, PAGE_SIZE) !=
         Add2Ptr(ImageBase, NtHeaders->OptionalHeader.SizeOfImage))
     {
-        DPRINT1("ImageBase 0x%p ImageSize 0x%lx Section %u VA 0x%lx Raw 0x%lx virt 0x%lx\n",
+        DPRINT1("WARNING: Image does not end with its last section. (ImageBase 0x%p ImageSize 0x%lx Section %lu VA 0x%lx Raw 0x%lx virt 0x%lx)\n",
```
learn-more, line 2536: So there is data behind the last section (usually called an overlay).

This is a warning because?

SergeGautherie: > This is a warning because?

I assumed so because, i.e., this `DPRINT1()` is a demoted `NT_ASSERT()`: 3410b25.
If it is not even meant as a warning, then it should be further demoted to `DPRINT()`, as a trace only...

ThFabba: Or... maybe it's neither a warning nor does it need to be demoted, because it's simply a debugging artifact that may be useful when troubleshooting other issues but is harmless in the general case?
Your default assumption should be to trust the original code authors, not to assume you know better.

PR #3286, opened 2020-10-10
[NTOS:IO] Complete the IRQ associated with a hard-error in case of rundown.

HBelusca:
## Purpose

Fix few bugs encountered in IoRaiseHardError() while discussing with @Extravert-ir and @ThFabba.

JIRA issue: [CORE-14037](https://jira.reactos.org/browse/CORE-14037).

## Proposed changes

~~- Fix buggy retrieval of the current calling Irp->Tail.Overlay.Thread.~~

~~- The 4th argument (KernelRoutine) to the KeInitializeApc() is **NOT**
  optional; however its 5th argument (RundownRoutine) is.~~

~~So use the mandatory routine for freeing the allocated APC, and
  use the optional routine to perform the rundown (when a thread
  exits while the APC is being executed) in order to just complete
  the associated IRP request.~~

~~- Check whether the ExAllocatePoolWithTag() call failed or not before
  queueing the allocated APC.~~

Use the optional routine to perform the rundown (when a thread
exits while the APC is being executed) in order to complete the
associated IRP request.

### DIFF
--- a/ntoskrnl/io/iomgr/error.c
+++ b/ntoskrnl/io/iomgr/error.c
@@ -484,6 +484,24 @@ IopLogWorker(IN PVOID Parameter)
     ExFreePoolWithTag(Message, TAG_IO);
 }
 
+/* See also ntoskrnl/io/iomgr/irp.c!IopCompleteRequest */
+static
+VOID
+NTAPI
+IopAbortApc(IN PKAPC Apc)
+{
+    PAGED_CODE();
+
+    /* Complete the associated request and free the APC */
+    IopCompleteRequest(Apc,
+                       &Apc->NormalRoutine,
+                       &Apc->NormalContext,
+                       &Apc->SystemArgument1,
+                       &Apc->SystemArgument2);
+
+    ExFreePoolWithTag(Apc, TAG_APC);
+}
+
 static
 VOID
 NTAPI
@@ -694,7 +712,7 @@ IoRaiseHardError(IN PIRP Irp,
                     &Thread->Tcb,
                     Irp->ApcEnvironment,
                     IopFreeApc,
-                    NULL,
+                    IopAbortApc,
                     IopRaiseHardError,
                     KernelMode,
                     Irp);


binarymaster (approved):
> JIRA issue: None.

Actually it's [CORE-14037](https://jira.reactos.org/browse/CORE-14037), and with the patch from PR applied the issue is entirely fixed for me! :tada: 

Please update the PR description and commit message with Jira ticket ID.

See the video demo: https://chat.reactos.org/files/ob9qfxufm3ypijkwe9wjpkbcqr/public?h=23-IDSjKpwuCumKtoVROLRIUnDLRAUP4K0bIY2CDIWQ

HBelusca:
Will do the updates once I get the review from Viktor and Thomas :)

ntoskrnl/io/iomgr/error.c
```diff
@@ -669,20 +689,29 @@ IoRaiseHardError(IN PIRP Irp,
         return;
     }
 
-    /* Setup an APC */
-    ErrorApc = ExAllocatePoolWithTag(NonPagedPool,
-                                     sizeof(KAPC),
-                                     TAG_APC);
+    // TODO: Do the APC only if this is for an asynchronous operation;
```
ThFabba, line 692: I wouldn't quite say it that way. The APC is the normal case, the work item is for very special circumstances (cleanup/fsctl IRPs for paging I/O only)

ThFabba (approved):

ntoskrnl/io/iomgr/error.c
```diff
@@ -484,15 +484,35 @@ IopLogWorker(IN PVOID Parameter)
     ExFreePoolWithTag(Message, TAG_IO);
 }
 
-VOID
+/* See also ntoskrnl/io/iomgr/irp.c!IopCompleteRequest */
+static VOID
```
Extravert-ir, line 488: Aren't we usually do
```suggestion
static
VOID
```
?

ntoskrnl/io/iomgr/error.c
Extravert-ir: The documentation says it always sends an APC and I don't any room for other cases there
https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/ntddk/nf-ntddk-ioraiseharderror

Extravert-ir (approved):

ntoskrnl/io/iomgr/error.c
HBelusca: Dunno; the former takes slightly less place ^^

ntoskrnl/io/iomgr/error.c
HBelusca: The documentation doesn't say "always"; it talks in generic terms.

ntoskrnl/io/iomgr/error.c
Extravert-ir: Why not `static VOID NTAPI` then?)

ntoskrnl/io/iomgr/error.c
```diff
@@ -484,15 +484,35 @@ IopLogWorker(IN PVOID Parameter)
...
+IopAbortApc(IN PKAPC Apc)
 {
+    PAGED_CODE();
+
+    IopCompleteRequest(Apc,
+                       &Apc->NormalRoutine,
+                       &Apc->NormalContext,
+                       &Apc->SystemArgument1,
+                       &Apc->SystemArgument2);
+}
```
tkreuzer, line 499: Free the Apc?

HBelusca: indeed!

HBelusca: Hmm Do you know whether there some special treatment by IopCompleteRequest() that may require the Apc pointer to be still valid after the call?

ThFabba: I think this might be a bug in Windows. They use `IopAbortRequest` here, which is the rundown routine that corresponds to IopCompleteRequest (where we don't use one). In that case, the APC doesn't need to be freed because it's part of the IRP structure. But that's not the case here.

On the other hand, Windows 10 still does this the same way so I may be missing something subtle.

HBelusca: Note that when we call `KeInitializeApc()` in `IoRaiseHardError()`, we associate an Irp to it (the last parameter); so that later, if the APC is being ran down (and this callback is called), it should still be part of this IRP, and this PR adds this new `IopAbortApc()` that calls `IopCompleteRequest()` (so it sounds very similar to what you describe for windows) to complete the associated IRP (our `IopCompleteRequest` also retrieves the IRP associated to this APC).

ntoskrnl/io/iomgr/error.c
HBelusca: I've reformulated the comment based on yours from mattermost, please check :)

ntoskrnl/io/iomgr/error.c
ThFabba: The last parameter is simply a context for the APC routines.
The APC itself is allocated from pool in IoRaiseHardError and nothing seems to free it in the rundown case.

jeditobe:
Я вся чешуся

HBelusca:
I made PR #3302 which is the same as this one, but without the rundown handling. That latter PR should be OK for merging, while I keep this one for investigating the rundown feature.

PR #3288, opened 2020-10-10
[PSDK] dsrole.h: Define DSROLE_PRIMARY_DS_READONLY

SergeGautherie:
I need it to enhance netapi32_apitest.


SergeGautherie:
For the record:
not yet in PSDK S2003R2, not at all in WDK 7.1. present in SDK 10.

HBelusca:
It's correct it's not in the PSDK of WS2003, but it is in PSDK 7.1 :
image

PR #3289, opened 2020-10-11
[SHELL32_APITEST] Add ShellHook testcase

katahiromz:
## Purpose

Add a testcase named `ShellHook` for shell hook.

JIRA issue: [CORE-17330](https://jira.reactos.org/browse/CORE-17330)

This testcase will test `HSHELL_WINDOWCREATED` especially. `HSHELL_WINDOWCREATED` affects display of task bar panes.

WinXP:
WinXP
Successful.

Win2k3:
Win2k3
Successful.

Win10:
Win10
Successful.


Doug-Lyons:
@katahiromz, This test does not find the regression introduced in CORE-17330. Can you add a test for this type of failure that would show a failure in what we have today and then not fail after the fix? Thanks.

katahiromz:
BEFORE `double_taskbar_fix.patch`:
before
AFTER `double_taskbar_fix.patch`:
after

Doug-Lyons:
@katahiromz, Hopefully this test would be able to identify between the two revisions listed below. I have tested reverting the changes made by the guilty revision and there are still 56 failures. Ideally there should be less before the guilty commit. Thanks.

reactos-bootcd-0.4.14-dev-368-gfa26425-x86-gcc-lin-dbg.iso ok
reactos-bootcd-0.4.14-dev-369-g141cf04-x86-gcc-lin-dbg.iso affected GUILTY

Doug-Lyons:
@katahiromz, My testing on W2K3SP2 shows 27 failures now. Can this be fixed? Thanks.

Shell32_apitest_ShellHook_Failures1

katahiromz:
BEFORE `double_taskbar_fix.patch`:
before

AFTER `double_taskbar_fix.patch`:
after

katahiromz:
Ready to go.

katahiromz:
@Doug-Lyons How about W2K3SP2 now?

Doug-Lyons:
@katahiromz, It's perfect. Thanks.
Shell32_apitest_ShellHook_W2K3SP2

PR #3292, opened 2020-10-11
[FASTFAT] Ensure that deferred write IRP contexts are not touched.

ThFabba:
Cc may decide to process deferred writes any time, so the context might
already be freed by the time we return from CcDeferWrite.
Also mark the IRPs as pending.

JIRA issue: [CORE-17328](https://jira.reactos.org/browse/CORE-17328)

### DIFF
--- a/drivers/filesystems/fastfat/misc.c
+++ b/drivers/filesystems/fastfat/misc.c
@@ -130,7 +130,7 @@ VfatDispatchRequest(
             break;
 
         case IRP_MJ_WRITE:
-            Status = VfatWrite(IrpContext);
+            Status = VfatWrite(&IrpContext);
             break;
 
         case IRP_MJ_FILE_SYSTEM_CONTROL:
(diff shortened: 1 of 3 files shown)


drivers/filesystems/fastfat/misc.c
```diff
@@ -182,37 +182,39 @@ VfatDispatchRequest(
...
+            /* Unless the IRP was queued, always free the IRP context */
+            VfatFreeIrpContext(IrpContext);
+        }
     }
 
     FsRtlExitFileSystem();
 
     return Status;
 }
-
```
HBelusca, line 215: hm?

HBelusca:
Read operations on the contrary do not need a similar treatment?

drivers/filesystems/fastfat/misc.c
ThFabba: Will fix.

ThFabba:
> Read operations on the contrary do not need a similar treatment?

There's no CcDeferRead, no :)

drivers/filesystems/fastfat/rw.c
```diff
@@ -999,13 +1000,15 @@ VfatWrite(
         Retrying = BooleanFlagOn(IrpContext->Flags, IRPCONTEXT_DEFERRED_WRITE);
         SetFlag(IrpContext->Flags, IRPCONTEXT_DEFERRED_WRITE);
 
+        IoMarkIrpPending(IrpContext->Irp);
```
SergeGautherie, line 1003: Indeed, fastfat_new does it inside its `FatPrePostIrp()`.

drivers/filesystems/fastfat/rw.c
```diff
@@ -873,8 +873,9 @@ VfatRead(
 
 NTSTATUS
 VfatWrite(
-    PVFAT_IRP_CONTEXT IrpContext)
+    PVFAT_IRP_CONTEXT *pIrpContext)
```
SergeGautherie, line 876: (For what it's worth, fastfat may need this change, though fastfat_new does not have it...)

PR #3294, opened 2020-10-12
[NTUSER] Fix Shell Hook HSHELL_WINDOWCREATED

katahiromz:
## Purpose

Fix Shell Hook and reduce failures of `shell32_apitest ShellHook` testcase.

JIRA issue: [CORE-17330](https://jira.reactos.org/browse/CORE-17330)

## Proposed changes

- Fix the condition of notifying shell hook `HSHELL_WINDOWCREATED` in `ntuser/winpos.c`.

BEFORE:
before
Some failures.

AFTER:
after
Successful.

### DIFF
--- a/win32ss/user/ntuser/winpos.c
+++ b/win32ss/user/ntuser/winpos.c
@@ -1904,18 +1904,13 @@ co_WinPosSetWindowPos(
    }
    else if (WinPos.flags & SWP_SHOWWINDOW)
    {
-      if (Window->spwndOwner == NULL ||
-          !(Window->spwndOwner->style & WS_VISIBLE) ||
-          (Window->spwndOwner->ExStyle & WS_EX_TOOLWINDOW))
+      if ((Window->ExStyle & (WS_EX_APPWINDOW | WS_EX_TOOLWINDOW)) >= WS_EX_APPWINDOW ||
+          (!(Window->ExStyle & WS_EX_TOOLWINDOW) && !Window->spwndOwner &&
+           (!Window->spwndParent || UserIsDesktopWindow(Window->spwndParent))))
       {
-         if (UserIsDesktopWindow(Window->spwndParent) &&
-             (!(Window->ExStyle & WS_EX_TOOLWINDOW) ||
-              (Window->ExStyle & WS_EX_APPWINDOW)))
-         {
-            co_IntShellHookNotify(HSHELL_WINDOWCREATED, (WPARAM)Window->head.h, 0);
-            if (!(WinPos.flags & SWP_NOACTIVATE))
-               UpdateShellHook(Window);
-         }
+         co_IntShellHookNotify(HSHELL_WINDOWCREATED, (WPARAM)Window->head.h, 0);
+         if (!(WinPos.flags & SWP_NOACTIVATE))
+            UpdateShellHook(Window);
       }
 
       Window->style |= WS_VISIBLE; //IntSetStyle( Window, WS_VISIBLE, 0 );
--- a/win32ss/user/ntuser/winpos.c
+++ b/win32ss/user/ntuser/winpos.c
@@ -1904,7 +1904,7 @@ co_WinPosSetWindowPos(
    }
    else if (WinPos.flags & SWP_SHOWWINDOW)
    {
-      if ((Window->ExStyle & (WS_EX_APPWINDOW | WS_EX_TOOLWINDOW)) >= WS_EX_APPWINDOW ||
+      if ((Window->ExStyle & WS_EX_APPWINDOW) ||
           (!(Window->ExStyle & WS_EX_TOOLWINDOW) && !Window->spwndOwner &&
            (!Window->spwndParent || UserIsDesktopWindow(Window->spwndParent))))
       {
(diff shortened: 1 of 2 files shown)


JoachimHenze:
Katayama, Is this PR meant to be committed **instead** or **additionally** to the patch that DougLyons proposed (https://jira.reactos.org/secure/attachment/58152/double_taskbar_fix.patch) in CORE-17330?

katahiromz:
I mean instead.

Doug-Lyons:
@katahiromz, You did a great job in fixing both the actual problem and getting the test results to succeed. BUT, even if you make no changes to the "ntuser\window.c" you get the same results of not seeing the double taskbar entries and the tests all still succeed. I suggest that you need to take out your changes to window.c or add some test cases that actually exercise this code path. Thanks. Doug Lyons

katahiromz:
Now I reverted `window.c`.

Doug-Lyons:
Thanks, @katahiromz. This looks GREAT.

win32ss/user/ntuser/winpos.c
```diff
@@ -1904,18 +1904,13 @@ co_WinPosSetWindowPos(
    }
    else if (WinPos.flags & SWP_SHOWWINDOW)
    {
-      if (Window->spwndOwner == NULL ||
-          !(Window->spwndOwner->style & WS_VISIBLE) ||
-          (Window->spwndOwner->ExStyle & WS_EX_TOOLWINDOW))
+      if ((Window->ExStyle & (WS_EX_APPWINDOW | WS_EX_TOOLWINDOW)) >= WS_EX_APPWINDOW ||
```
HBelusca, line 1907: For this first test, why not just looking at `(Window->ExStyle & WS_EX_APPWINDOW) == WS_EX_APPWINDOW`, since:
WS_EX_APPWINDOW is == 0x40000 ; WS_EX_TOOLWINDOW == 128, so your original test would be true only when at least the WS_EX_APPWINDOW bit is set? (so basically at this stage the test result doesn't really change even if you didn't look at the WS_EX_TOOLWINDOW part).

katahiromz: Be careful. It is not `==` but `>=`.

HBelusca: Yes I know, when you do `ExStyle & (WS_EX_APPWINDOW | WS_EX_TOOLWINDOW)` the obtained result will be either 0x40000 (just the appwindow bit set) or 0x80 (== 128 , just the toolwindow bit set), or the "sum" 0x40080 (both bits set), and then, you do your greater-or-equal comparison with WS_EX_APPWINDOW (which is 0x40000).
So the only way this test can be TRUE, is when the `ExStyle & (WS_EX_APPWINDOW | WS_EX_TOOLWINDOW)` is equal to 0x40000 or 0x40080, i.e. in both cases the appwindow bit is set, and basically you don't care about the toolwindow.
Hence, in my opinion, this test can just be simplified to checking whether the appwindow bit is set, as I proposed above.

katahiromz: Simplified about `WS_EX_APPWINDOW`.

Doug-Lyons: @HBelusca, Nice Simplification. Thanks.

JoachimHenze:
> @katahiromz, You did a great job in fixing both the actual problem and getting the test results to succeed. BUT, even if you make no changes to the "ntuser\window.c" you get the same results of not seeing the double taskbar entries and the tests all still succeed. I suggest that you need to take out your changes to window.c or add some test cases that actually exercise this code path. Thanks. Doug Lyons
> Addition: Since this part of the code in window.c is originally written by you, I see no reason that your new changes should not be made until they can be proven by some programs later or new tests to be added later. Thanks for your work here.

**I could imagine** (unverified) that the changes in "ntuser\window.c" may (or may not) be needed to get a consistent behavior regarding the taskswitcher later? We should keep the behavior of the taskswitcher in mind at least when we test that later.

katahiromz:
One of `window.c` generates an `HSHELL_WINDOWCREATED` notification messsage when a window is being created and shown in some conditions in general.

Doug-Lyons:
I added an ERR print to window.c as shown below and I have not been able to trigger it yet regardless of what I try:
window_c_Shell_Notify

katahiromz:
Might be invalid and unneeded code...

katahiromz:
In File `window.c`:
```c
   /* Notify the shell that a new window was created */
   if (UserIsDesktopWindow(Window->spwndParent) &&
       Window->spwndOwner == NULL &&
       (Window->style & WS_VISIBLE) &&
       (!(Window->ExStyle & WS_EX_TOOLWINDOW) ||
        (Window->ExStyle & WS_EX_APPWINDOW)))
   {
      co_IntShellHookNotify(HSHELL_WINDOWCREATED, (WPARAM)hWnd, 0);
   }
```
@Doug-Lyons Can you delete this part and retest again? Thanks.

katahiromz:
VirtualBox_ReactOS_13_10_2020_03_02_29
I found the notification code in `window.c` is unneeded code. Deleted.

Doug-Lyons:
@katahiromz, I was not sure about whether this code in window.c was ever called, so I ran a full suite of rosautotest's on my local computer and this test actually showed this call being done and executed four (4) times. Two occurrences are in comctl32:misc, one is in user32:class and one is in user32:msg_messages.  I have attached the results of these tests here for your review.

[window_c_ShellNotify.zip](https://github.com/reactos/reactos/files/5367135/window_c_ShellNotify.zip)

With us now being able to see where they might be used, I think that we need to do a bit more investigation before deleting this code. Thanks. Doug Lyons

PR #3296, opened 2020-10-14
[ATL] Implement AtlComModuleGetClassObject and call it in CComModule::GetClassObject

oleg-dubinskiy:
## Purpose

Add `AtlComModuleGetClassObject` implementation adapted from dll/win32/atl/atl.c and use it in `CComModule::GetClassObject` function.
This will allow to properly get class object from the dlls whose use ATL/COM. Their `DllGetClassObject` implementations directly call `CComModule::GetClassObject`, which, in its turn, uses `AtlComModuleGetClassObject`.
I found this unimplemented code path during investigating browseui bug from CORE-17337. Our browseui.dll fails to load some classes from registry, even if they persist there.
Although it does not fix the browseui problem, but nevertheless it improves the situation a bit.

JIRA issue: [CORE-17337](https://jira.reactos.org/browse/CORE-17337)

### DIFF
--- a/sdk/lib/atl/atlbase.h
+++ b/sdk/lib/atl/atlbase.h
@@ -868,8 +868,7 @@ public:
         }
         if (hResult == S_OK && *ppv == NULL)
         {
-            // FIXME: call AtlComModuleGetClassObject
-            hResult = CLASS_E_CLASSNOTAVAILABLE;
+            hResult = AtlComModuleGetClassObject(&_AtlComModule, rclsid, riid, ppv);
         }
         return hResult;
     }
@@ -1802,6 +1801,30 @@ inline HRESULT WINAPI AtlComModuleUnregisterServer(_ATL_COM_MODULE *mod, BOOL bU
     return hResult;
 }
 
+// Adapted from dll/win32/atl/atl.c
+inline HRESULT WINAPI AtlComModuleGetClassObject(_ATL_COM_MODULE *pComModule, REFCLSID rclsid, REFIID riid, void **ppv)
+{
+    _ATL_OBJMAP_ENTRY **iter;
+    IUnknown* unk;
+    HRESULT hr;
+
+    if (!pComModule)
+        return E_INVALIDARG;
+
+    for (iter = pComModule->m_ppAutoObjMapFirst; iter < pComModule->m_ppAutoObjMapLast; iter++)
+    {
+        if (IsEqualCLSID(*(*iter)->pclsid, rclsid) && (*iter)->pfnGetClassObject)
+        {
+            if (!(*iter)->pCF)
+                hr = (*iter)->pfnGetClassObject((void*)(*iter)->pfnCreateInstance, IID_IUnknown, (void**)&(*iter)->pCF);
+            if ((*iter)->pCF)
+                hr = unk->QueryInterface(riid, ppv);
+            return hr;
+        }
+    }
+
+    return CLASS_E_CLASSNOTAVAILABLE;
+}
 
 // Adapted from dll/win32/atl/atl.c
 inline HRESULT WINAPI AtlComModuleRegisterClassObjects(_ATL_COM_MODULE *module, DWORD context, DWORD flags)


sdk/lib/atl/atlbase.h
```diff
@@ -1802,6 +1801,30 @@ inline HRESULT WINAPI AtlComModuleUnregisterServer(_ATL_COM_MODULE *mod, BOOL bU
...
+        return E_INVALIDARG;
+
+    for (iter = pComModule->m_ppAutoObjMapFirst; iter < pComModule->m_ppAutoObjMapLast; iter++)
+    {
+        if (IsEqualCLSID(*(*iter)->pclsid, rclsid) && (*iter)->pfnGetClassObject)
+        {
+            if (!(*iter)->pCF)
+                hr = (*iter)->pfnGetClassObject((void*)(*iter)->pfnCreateInstance, IID_IUnknown, (void**)&(*iter)->pCF);
+            if ((*iter)->pCF)
+                hr = unk->QueryInterface(riid, ppv);
```
katahiromz, line 1821: `unk` is NULL.

sdk/lib/atl/atlbase.h
```diff
@@ -1802,6 +1801,30 @@ inline HRESULT WINAPI AtlComModuleUnregisterServer(_ATL_COM_MODULE *mod, BOOL bU
...
+
+    if (!pComModule)
+        return E_INVALIDARG;
+
+    for (iter = pComModule->m_ppAutoObjMapFirst; iter < pComModule->m_ppAutoObjMapLast; iter++)
+    {
+        if (IsEqualCLSID(*(*iter)->pclsid, rclsid) && (*iter)->pfnGetClassObject)
+        {
+            if (!(*iter)->pCF)
+                hr = (*iter)->pfnGetClassObject((void*)(*iter)->pfnCreateInstance, IID_IUnknown, (void**)&(*iter)->pCF);
```
katahiromz, line 1819: Can we use `IID_PPV_ARG` macro?

sdk/lib/atl/atlbase.h
oleg-dubinskiy: Fixed, thanks.

oleg-dubinskiy: Oops, looks like it breaks MSVC. :sweat_smile:

learn-more:
> Although it does not fix the browseui problem, but nevertheless it improves the situation a bit.

Do you have a specific case that is improved?

oleg-dubinskiy:
> > Although it does not fix the browseui problem, but nevertheless it improves the situation a bit.
> 
> Do you have a specific case that is improved?

Looks like no @learn-more. Sorry if I was wrong.
From the start, I expected that it will fix browseui DllGetClassObject error, but it doesn't. Although I didn't test yet any other dll(s) that use(s) ATL (any other specific cases when it may fail with the similar error), so it still might fix another similar potential errors.

learn-more:
> > > Although it does not fix the browseui problem, but nevertheless it improves the situation a bit.
> > 
> > 
> > Do you have a specific case that is improved?
> 
> Looks like no @learn-more. Sorry if I was wrong.
> From the start, I expected that it will fix browseui DllGetClassObject error, but it doesn't. Although I didn't test yet any other dll(s) that use(s) ATL (any other specific cases when it may fail with the similar error), so it still might fix another similar potential errors.

I have not checked the code in-depth, but I suspect this just re-visits the same classes that were already checked before.
As for https://jira.reactos.org/browse/CORE-17337 , there is no code in reactos currently implementing that interface:
https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&sr=1&s=603d3800
https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&sr=1&s=CLSID_ShellTaskScheduler
https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&sr=1&s=CLSID_BackgroundTaskScheduler
So either way this PR won't make that available.

oleg-dubinskiy:
> > > > Although it does not fix the browseui problem, but nevertheless it improves the situation a bit.
> > > 
> > > 
> > > Do you have a specific case that is improved?
> > 
> > 
> > Looks like no @learn-more. Sorry if I was wrong.
> > From the start, I expected that it will fix browseui DllGetClassObject error, but it doesn't. Although I didn't test yet any other dll(s) that use(s) ATL (any other specific cases when it may fail with the similar error), so it still might fix another similar potential errors.
> 
> I have not checked the code in-depth, but I suspect this just re-visits the same classes that were already checked before.
> As for https://jira.reactos.org/browse/CORE-17337 , there is no code in reactos currently implementing that interface:
> https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&sr=1&s=603d3800
> https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&sr=1&s=CLSID_ShellTaskScheduler
> https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&sr=1&s=CLSID_BackgroundTaskScheduler
> So either way this PR won't make that available.

Sadly. :frowning_face: Then this PR at least implements some stuff in the ATL itself, which is an improvement anyway.
I wonder is there any MS or 3rd-party documentation for this... If yes, it might be possible to at least stubplement that. But for now, I didn't find any docs about Background/ShellTaskScheduler in browseui.

learn-more:
In it's current state this adds nothing,
it reads members that are not initialized to something other than NULL,
the code uses uninitialized variables...

PR #3299, opened 2020-10-16
[SHELL32] Resizable for SHBrowseForFolder

katahiromz:
## Purpose

Make the dialog box of `SHBrowseForFolder` function resizable.

JIRA issue: [CORE-17341](https://jira.reactos.org/browse/CORE-17341)

## Proposed changes

- Add `WS_SIZEBOX` window style for resource dialog `IDD_BROWSE_FOR_FOLDER_NEW`.
- Add `layout.h` (I wrote) into `sdk/include/reactos`.
- Use `layout.h` to resize the dialog box.
- Use `WM_GETMINMAXINFO` rather than `WM_WINDOWPOSCHANGING` to limit the window size.

## TODOs
- [x] Do test.
- [x] Optimize speed.
- [x] Optimize memory.
- [x] Consider file location. 
- [x] Confirm license condition.

AFTER:
after
Resizable.


binarymaster:
> Use `cresize` and `c-vector` (License: MIT) to resize the dialog box.

Third-party stuff... :thinking: Should probably go to sdk/lib or some other common place.

katahiromz:
The control layout info can be statically stored. Maybe `c-vector` is unnecessary.

HBelusca (dismissed):
> Use cresize and c-vector (License: MIT) to resize the dialog box.đ

Please don't.

JoachimHenze:
I think I could be able to write this by hand'n'feet. The dialog is not so complex actually, so it is clear where everything should be moved to. Would other devs also prefer a hand-knitted solution instead of using this 3rd party lib?

katahiromz:
File `cvector.h` is removed. File `cresize.h` is now under LGPL 2.0.

katahiromz:
OK. 3rd party code is removed.

sdk/include/reactos/layout.h
```diff
@@ -0,0 +1,232 @@
+/*
+ * PROJECT:     ReactOS headers
+ * LICENSE:     LGPL-2.0-or-later (https://spdx.org/licenses/LGPL-2.0-or-later)
+ * PURPOSE:     The layout engine of resizable dialog boxes / windows
```
learn-more, line 4: There is also `base/applications/rapps/include/rosui.h` which was supposed to be used in more places than only rapps, but it might not apply here.
Just something to keep in mind :)

HBelusca: This layout.h file is also nice, in case one wants to do some layout calculation in a C program 😂

katahiromz:
I cannot wait forever.

katahiromz:
Timeout. I will merge PR.

HBelusca:
> I cannot wait forever.

People have also a real-life life and job, in other words they don't work on ReactOS full time. So it is expected that PRs can be reviewed every other next week(-end).

PR #3302, opened 2020-10-16
[NTOS:IO] Some fixes for IoRaiseHardError().

HBelusca:
## Purpose

Fix few bugs encountered in IoRaiseHardError() while discussing with @Extravert-ir and @ThFabba.

JIRA issue: [CORE-14037](https://jira.reactos.org/browse/CORE-14037).

## Proposed changes

- Fix buggy retrieval of the current calling Irp->Tail.Overlay.Thread.

- The 4th argument (KernelRoutine) to the KeInitializeApc() is **NOT**
  optional; however its 5th argument (RundownRoutine) is.
  So use the mandatory routine for freeing the allocated APC instead.
  We don't use the rundown routine yet.

- Check whether the ExAllocatePoolWithTag() call failed or not before
  queueing the allocated APC.

### DIFF
--- a/ntoskrnl/io/iomgr/error.c
+++ b/ntoskrnl/io/iomgr/error.c
@@ -484,14 +484,22 @@ IopLogWorker(IN PVOID Parameter)
     ExFreePoolWithTag(Message, TAG_IO);
 }
 
+static
 VOID
 NTAPI
-IopFreeApc(IN PKAPC Apc)
+IopFreeApc(IN PKAPC Apc,
+           IN OUT PKNORMAL_ROUTINE* NormalRoutine,
+           IN OUT PVOID* NormalContext,
+           IN OUT PVOID* SystemArgument1,
+           IN OUT PVOID* SystemArgument2)
 {
+    PAGED_CODE();
+
     /* Free the APC */
-    ExFreePool(Apc);
+    ExFreePoolWithTag(Apc, TAG_APC);
 }
 
+static
 VOID
 NTAPI
 IopRaiseHardError(IN PVOID NormalContext,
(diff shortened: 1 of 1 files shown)


binarymaster (approved):
Tested, works just fine as previous PR :)

Extravert-ir (approved):

ntoskrnl/io/iomgr/error.c
```diff
@@ -669,21 +677,29 @@ IoRaiseHardError(IN PIRP Irp,
...
+        IoCompleteRequest(Irp, IO_DISK_INCREMENT);
+        return;
+    }
+
     KeInitializeApc(ErrorApc,
                     &Thread->Tcb,
                     Irp->ApcEnvironment,
-                    NULL,
                     IopFreeApc,
+                    NULL,
```
tkreuzer, line 697: Now it can leak. You need both.

ntoskrnl/io/iomgr/error.c
```diff
@@ -657,7 +665,7 @@ IoRaiseHardError(IN PIRP Irp,
                  IN PVPB Vpb,
                  IN PDEVICE_OBJECT RealDeviceObject)
 {
-    PETHREAD Thread = (PETHREAD)&Irp->Tail.Overlay.Thread;
```
tkreuzer, line 660: I just want to note that this line was authored by Alex Ionescu.

tkreuzer (approved):

ThFabba (approved):

ntoskrnl/io/iomgr/error.c
HBelusca, line 697: (For information the allocated APC won't leak even if there is no RundownRoutine, because when such a function is absent, the kernel will then proceed to free the allocated APC:
https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/ps/kill.c;hb=7790670dd80c5d4cf3c4b68f8327fe97126c6229#l880
)

PR #3305, opened 2020-10-18
[FASTFAT] Implement the overflow queue. CORE-17344

ThFabba:
This avoids blocking all Ex worker threads in fastfat, thereby making Cc
unable to issue the lazy writes that would unblock those workers.
This is more or less directly taken from fastfat_new.

JIRA issue: [CORE-17344](https://jira.reactos.org/browse/CORE-17344)

### DIFF
--- a/drivers/filesystems/fastfat/misc.c
+++ b/drivers/filesystems/fastfat/misc.c
@@ -336,12 +336,53 @@ static
 VOID
 NTAPI
 VfatDoRequest(
-    PVOID IrpContext)
+    PVOID Context)
 {
+    PVFAT_IRP_CONTEXT IrpContext = Context;
+    PDEVICE_EXTENSION DeviceExt;
+    KIRQL OldIrql;
+
     InterlockedDecrement(&QueueCount);
-    DPRINT("VfatDoRequest(IrpContext %p), MajorFunction %x, %d\n",
-           IrpContext, ((PVFAT_IRP_CONTEXT)IrpContext)->MajorFunction, QueueCount);
-    VfatDispatchRequest((PVFAT_IRP_CONTEXT)IrpContext);
+
+    if (IrpContext->Stack->FileObject != NULL)
+    {
+        DeviceExt = IrpContext->Stack->DeviceObject->DeviceExtension;
+        ObReferenceObject(DeviceExt->VolumeDevice);
+    }
+
+    do
+    {
+        DPRINT("VfatDoRequest(IrpContext %p), MajorFunction %x, %d\n",
+               IrpContext, IrpContext->MajorFunction, QueueCount);
+        VfatDispatchRequest(IrpContext);
+        IrpContext = NULL;
+
+        /* Now process any overflow items */
+        if (DeviceExt != NULL)
+        {
+            KeAcquireSpinLock(&DeviceExt->OverflowQueueSpinLock, &OldIrql);
+            if (DeviceExt->OverflowQueueCount != 0)
+            {
+                IrpContext = CONTAINING_RECORD(RemoveHeadList(&DeviceExt->OverflowQueue),
+                                               VFAT_IRP_CONTEXT,
+                                               WorkQueueItem.List);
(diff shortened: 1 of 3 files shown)


drivers/filesystems/fastfat/misc.c
```diff
@@ -336,19 +336,64 @@ static
...
+
+    do
+    {
+        DPRINT("VfatDoRequest(IrpContext %p), MajorFunction %x, %d\n",
+               IrpContext, IrpContext->MajorFunction, QueueCount);
+        VfatDispatchRequest(IrpContext);
+        IrpContext = NULL;
+
+        /* Now process any overflow items */
+        if (DeviceExt != NULL)
```
learn-more, line 361: Would it make sense to also check OverflowQueueCount here outside the lock?
Theoretically this would prevent grabbing the spinlock when there is no work to be done.

drivers/filesystems/fastfat/misc.c
```diff
@@ -357,10 +402,41 @@ VfatQueueRequest(
...
     IoMarkIrpPending(IrpContext->Irp);
-    ExInitializeWorkItem(&IrpContext->WorkQueueItem, VfatDoRequest, IrpContext);
-    ExQueueWorkItem(&IrpContext->WorkQueueItem, CriticalWorkQueue);
+
+    /* We should not block more than two worker threads per volume,
+     * or we might stop Cc from doing the work to unblock us.
+     * Add additional requests into the overflow queue instead and process
+     * them all in an existing worker thread (see VfatDoRequest above).
+     */
+    if (IrpContext->Stack->FileObject != NULL)
```
learn-more, line 414: Did you mean to check `DeviceObject` here, or is there a relation between FileObject and DeviceObject?

drivers/filesystems/fastfat/misc.c
```diff
@@ -336,19 +336,64 @@ static
...
+    }
+
+    do
+    {
+        DPRINT("VfatDoRequest(IrpContext %p), MajorFunction %x, %d\n",
+               IrpContext, IrpContext->MajorFunction, QueueCount);
+        VfatDispatchRequest(IrpContext);
+        IrpContext = NULL;
+
+        /* Now process any overflow items */
```
HBelusca, line 360: If I understand correctly the code, it's processing just one item at a time?
I was wondering why you are not doing something like this instead:
```c
while (!IsListEmpty(&DeviceExt->OverflowQueue))
{
    IrpContext = CONTAINING_RECORD(RemoveHeadList(&DeviceExt->OverflowQueue),
                                   VFAT_IRP_CONTEXT,
                                   WorkQueueItem.List);

    // Proceed as in the rest...
}
```

drivers/filesystems/fastfat/misc.c
ThFabba: We do need to do work in either case, since `PostedRequestCount` is also protected by the lock.
I don't think it's a safe shortcut to take, since the queuing thread must also have a guarantee about the fact that there's a worker thread active that will handle its overflow request. So if we determine that `OverflowQueueCount` is 0, we must reset `PostedRequestCount` atomically with that check, or we risk letting an item sit in the queue without a thread to pick it up.

drivers/filesystems/fastfat/misc.c
ThFabba: The first item is the one passed to the WorkItem via the context. After that it processes items from the overflow queue (which may not exist, if there is no FileObject).
I could separate the first item from the loop if you feel strongly that that would be clearer, e.g.
```
VfatDispatchRequest();

if (DeviceExt)
{
    while (!IsListEmpty())
    {
        VfatDispatchRequest();
    }
}
```

drivers/filesystems/fastfat/misc.c
ThFabba: Yeah I don't totally understand this part, but FileObject is right -- it's what fastfat_new does.

tkreuzer (approved):
lgtm

PR #3306, opened 2020-10-19
[REACTOS] *FastIoDispatch: Remove redundant zero'ing

SergeGautherie:
## Purpose

Cleanup.

JIRA issue: [CORE-17342](https://jira.reactos.org/browse/CORE-17342)

### DIFF
--- a/drivers/base/null/null.c
+++ b/drivers/base/null/null.c
@@ -202,7 +202,6 @@ DriverEntry(IN PDRIVER_OBJECT DriverObject,
     DriverObject->DriverUnload = NullUnload;
 
     /* Initialize the fast I/O dispatch table */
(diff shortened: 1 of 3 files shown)


drivers/base/null/null.c
```diff
@@ -202,7 +202,6 @@ DriverEntry(IN PDRIVER_OBJECT DriverObject,
     DriverObject->DriverUnload = NullUnload;
 
     /* Initialize the fast I/O dispatch table */
-    RtlZeroMemory(&FastIoDispatch, sizeof(FastIoDispatch));
```
HBelusca, line 205: Why do you think it's redundant? Have you checked the default contents of the structure when the code is compiled with runtime checks enabled in MSVC?

tkreuzer:
Pointless and dangerous.

PR #3307, opened 2020-10-19
[KMTESTS:NTOS] Report FastIO calls, Set SizeOfFastIoDispatch

SergeGautherie:
## Purpose

Be more explicit and correct.

JIRA issue: [CORE-17342](https://jira.reactos.org/browse/CORE-17342)

## Notes

- CcCopyRead: number of test varies between 188/200/212 before, add 10 after.
I am not checking why it is not constant.
- NtCreateSection: a few new failures on ROS+FASTFAT.
To be investigated when existing failures are fixed. (Simple fallout? Maybe ROS-FAT vs MS-NTFS? ...)

### DIFF
--- a/modules/rostests/kmtests/ntos_cc/CcCopyRead_drv.c
+++ b/modules/rostests/kmtests/ntos_cc/CcCopyRead_drv.c
@@ -36,6 +36,9 @@ FastIoRead(
     _Out_ PIO_STATUS_BLOCK IoStatus,
     _In_ PDEVICE_OBJECT DeviceObject)
 {
+    // Report whether this FastIO function is actually called or not.
+    ok(TRUE, "\n");
+
     IoStatus->Status = STATUS_NOT_SUPPORTED;
     return FALSE;
 }
--- a/modules/rostests/kmtests/ntos_cc/CcCopyRead_drv.c
+++ b/modules/rostests/kmtests/ntos_cc/CcCopyRead_drv.c
@@ -65,9 +65,9 @@ TestEntry(
     KmtRegisterIrpHandler(IRP_MJ_CREATE, NULL, TestIrpHandler);
     KmtRegisterIrpHandler(IRP_MJ_READ, NULL, TestIrpHandler);
 
+    TestFastIoDispatch.SizeOfFastIoDispatch = sizeof(TestFastIoDispatch);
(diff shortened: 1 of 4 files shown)


modules/rostests/kmtests/ntos_cc/CcCopyRead_drv.c
```diff
@@ -36,6 +36,9 @@ FastIoRead(
     _Out_ PIO_STATUS_BLOCK IoStatus,
     _In_ PDEVICE_OBJECT DeviceObject)
 {
+    // Report whether this FastIO function is actually called or not.
+    ok(TRUE, "\n");
```
HBelusca, line 40: Why does this need to just emit an extra newline?

SergeGautherie:
> Why does this need to just emit an extra newline?

The newline is a safety and reminder. Text is useless in this usecase.
What would you prefer instead?

HBelusca:
A safety for what, exactly?

modules/rostests/kmtests/ntos_cc/CcCopyRead_drv.c
SergeGautherie: The newline is a safety and reminder. Text is useless in this usecase.
What would you prefer instead?

SergeGautherie:
> A safety for what, exactly?

In case it gets used somehow.
Sorry, I don't understand what is bothering you exactly:
What would you prefer instead?

modules/rostests/kmtests/ntos_cc/CcCopyRead_drv.c
```diff
@@ -62,10 +65,10 @@ TestEntry(
     KmtRegisterIrpHandler(IRP_MJ_CREATE, NULL, TestIrpHandler);
     KmtRegisterIrpHandler(IRP_MJ_READ, NULL, TestIrpHandler);
 
+    TestFastIoDispatch.SizeOfFastIoDispatch = sizeof(TestFastIoDispatch);
```
ThFabba, line 68: This is a test. The test shows that the kernel code being tested doesn't check `SizeOfFastIoDispatch`.

GeoB99:
I'm going to close this since this PR as is been sitting here for quite a long time with not further input. Feel free however to poke me though in case you want to continue working on this.

PR #3308, opened 2020-10-20
[BROWSEUI] CCommonBrowser class should support aggregation

oleg-dubinskiy:
## Purpose

Fix aggregation support in `CCommonBrowser` class of our browseui.dll. According to https://www.geoffchappell.com/studies/windows/ie/browseui/classes/ccommonbrowser.htm, it should support aggregation.
It will remove the following error from the log when use MS shell32.dll in ReactOS:
`fixme:(dll/win32/ole32/compobj.c:3454) Class {af604efe-8897-11d1-b944-00a0c90312e1} does not support aggregation`.

JIRA issue: [CORE-17345](https://jira.reactos.org/browse/CORE-17345)

## Proposed changes

- Use `DECLARE_AGGREGATABLE` macro instead of `DECLARE_NOT_AGGREGATABLE` one, according to https://docs.microsoft.com/en-us/cpp/atl/reference/aggregation-and-class-factory-macros?view=vs-2019. So now it can be aggregated.


learn-more (approved):
lgtm.

I would however love some tests (in another PR) proving that our `DECLARE_AGGREGATABLE` implementation is behaving as it should (there is currently only one class using it, and I honestly don't know if it uses aggregation at all).
An example would be to write a test against the object here:
https://docs.microsoft.com/en-us/windows/win32/com/aggregation
and then rewrite the test using our COM classes.

PR #3309, opened 2020-10-21
[BROWSEUI][PSDK] Stubplement UserAssist CORE-17345

oleg-dubinskiy:
## Purpose

Add initial implementation of `UserAssist` class, with `IUserAssist` interface, according to https://www.geoffchappell.com/studies/windows/ie/browseui/classes/userassist.htm.
Required by MS shell32.dll from our browseui.dll, in order to work properly.

JIRA issue: [CORE-17345](https://jira.reactos.org/browse/CORE-17345)

## Proposed changes

- Add the `CUserAssist` class declaration and stubplementation, GUID is based on https://git.reactos.org/?p=reactos.git;a=blob;f=modules/rostests/apitests/com/com_apitest.h;hb=ba09834c5ebdaf8d3e2a518c7fb78eaa7d89bb50#l98.
- Add the `IUserAssist` interface declaration and stub all methods.
- Enable using them for browseui's `DllGetClassObject`, which uses ATL.
- Add rgs resource with required registry data, which is required for proper loading the class, and include it in the main resource file. It can be added into registry by registering browseui via `regsvr32 /i browseui.dll` command.
- Update precompiled header and CMake stuff accordingly.

## TODO

- [ ] Separate revised method from the original ones if needed, since it has a different GUID, according to the reference. How I can do this properly?
- [x] Make it working. Before my changes, it did not load, since the interface was missing, and it still doesn't work, but now ole COM factory returns `E_NOINTERFACE` what means "No such interface supported", although now the interface is present. Either it requires the proper implementation, or something is wrong with the stub implementation itself. Maybe the CLSID declaration is wrong somewhere? :thinking: All GUIDs seem to be correct, according to the documentation link.


sdk/include/psdk/shobjidl.idl
```diff
@@ -4768,4 +4768,40 @@ interface IItemNameLimits : IUnknown
         [out] int *piMaxNameLen);
 };
 
+/*****************************************************************************
+ * IUserAssist interface
+ */
+[
+    uuid(dd313e04-feff-11d1-8ecd-0000f87a470c),
```
learn-more, line 4775: This is CLSID_UserAssist, it should be IID_IUserAssist, which is `DD313E05-FEFF-11D1-8ECD-0000F87A470C` according to https://www.geoffchappell.com/studies/windows/ie/browseui/interfaces/iuserassist.htm

oleg-dubinskiy: Indeed. Thanks @learn-more. :+1: 
Now that error does no longer appear for me in the log.
But it still doesn't work properly.
Judging by the log, now it needs `ShellTaskSheduler`, same as MS shimgvw.dll from CORE-17337.
I think I can send that in another PR. Since that one is documented by MS: https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nn-shobjidl_core-ishelltaskscheduler, unlike `UserAssist`, it will be more easy to stubplement that.

learn-more (approved):

PR #3311, opened 2020-10-22
Fix Clang-CL build

zefklop:
## Purpose

Link apisets only agains import libraries
Make CRTDLL be a standalone CRT library
Improve msvcrtex <-> msvcrt handling a bit


katahiromz:
Still failing.

zefklop:
> Still failing.

But not on the same module

katahiromz:
Successful. To go.

dll/apisets/CMakeLists.txt
```diff
@@ -30,7 +30,9 @@ function (add_apiset apiset_name baseaddress)
         target_compile_options(${apiset_name} PRIVATE /wd4026 /wd4273)
     endif()
 
-    add_importlibs(${apiset_name} ${ARGN} ntdll)
+    # link against our "raw" import libs, otherwise add_importlibs brings msvcrtex in
+    list(TRANSFORM ARGN PREPEND lib OUTPUT_VARIABLE liblist)
+    target_link_libraries(${apiset_name} ${liblist} libntdll)
```
learn-more, line 35: There are also some other macro's called in `add_importlibs`, namely:
`add_dependency_node(${_module})`, `add_dependencies(${_module} lib${LIB})`, `add_dependency_edge(${_module} ${LIB})`
Those are not needed?

learn-more: `msvcrtex` is only brought in on `msvcrt`, so would it make sense to add an optional argument on `add_importlibs` to specify that msvcrtex should be skipped?

dll/win32/crtdll/CMakeLists.txt
```diff
@@ -19,12 +19,7 @@ list(APPEND SOURCE
 
 add_library(crtdll MODULE ${SOURCE})
 set_module_type(crtdll win32dll ENTRYPOINT DllMain 12)
-target_link_libraries(crtdll crt wine)
-
-if(MSVC)
-    add_importlibs(crtdll msvcrt)
-    target_link_libraries(crtdll oldnames)
-endif()
+target_link_libraries(crtdll crt wine ${PSEH_LIB})
```
learn-more, line 22: `oldnames` is not needed anymore?

dll/win32/crtdll/crtdll.spec
```diff
@@ -1,528 +1,528 @@
-# Old C runtime library. All functions provided by msvcrt
+# Old C runtime library. All functions provided by
```
learn-more, line 1: provided by?

This effectively changes crtdll from a forwarded to a full-blown (more or less) copy of msvcrt?

dll/apisets/CMakeLists.txt
zefklop: right, we should probably take this, too

dll/win32/crtdll/CMakeLists.txt
zefklop: no, crtdll doesn't import anything anymore

dll/win32/crtdll/crtdll.spec
zefklop: Provided by itself, will fix.

Yes, more or less. But this is important : a program compiled with crtdll (if you find any) is not supposed to mess around with msvcrt state, and our libraries should not be at the mercy of a program using crtdll messing around with it.

Not that we should rely much on MSVCRT state neither, but this mitigates one risk :-/

dll/apisets/CMakeLists.txt
Extravert-ir: `add_dependency_node` and `add_dependency_edge` are needed only for building an xml file with dependencies (we're bypassing that in so many places - does it make sense at all?)
As of `add_dependencies`, it seems that it is not needed when used with `target_link_libraries`: https://stackoverflow.com/questions/27282938/target-link-libraries-and-add-dependencies

Giving all that, do we need `add_importlibs` at all?

zefklop: Fixed :-)

dll/win32/crtdll/crtdll.spec
Extravert-ir: Please, put a reference to a previous commit in a commit message f0c91ae971ca9affd819aa10cf65ad925217932b

sdk/cmake/CMakeMacros.cmake
```diff
@@ -533,7 +533,8 @@ function(add_importlibs _module)
     foreach(LIB ${ARGN})
         if("${LIB}" MATCHES "msvcrt")
             target_compile_definitions(${_module} PRIVATE _DLL __USE_CRTIMP)
-            target_link_libraries(${_module} msvcrtex)
+            # set_module_type might not have been called at this point, so use a generator expression
+            target_link_libraries(${_module} "$<$<IN_LIST:$<TARGET_PROPERTY:${_module},REACTOS_MODULE_TYPE>,win32gui;win32cui;win32dll;win32ocx;cpl>:msvcrtex>")
```
tkreuzer, line 537: Can't we do ```target_link_libraries(msvcrt INTERFACE msvcrtex)``` and get all of this automatically?

dll/apisets/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ function (add_apiset apiset_name baseaddress)
         ${CMAKE_CURRENT_BINARY_DIR}/${apiset_name}.def)
 
     add_dependencies(${apiset_name} xdk)
-    set_module_type(${apiset_name} win32dll ENTRYPOINT 0 UNICODE IMAGEBASE ${baseaddress})
+    set_module_type(${apiset_name} module UNICODE IMAGEBASE ${baseaddress})
```
tkreuzer, line 25: While I totally agree with this change, this doesn't actually address the real issue (which is a broken CRT). I have an additional fix for the CRT ready soon.

dll/win32/crtdll/crtdll.spec
learn-more: allright, thanks for the explanation!

tkreuzer (approved):
While the crtdll change will clearly increase our iso size, I guess it's the correct thing to do.

dll/win32/crtdll/crtdll.spec
zefklop: > Please, put a reference to a previous commit in a commit message [f0c91ae](https://github.com/reactos/reactos/commit/f0c91ae971ca9affd819aa10cf65ad925217932b)

Done

sdk/cmake/CMakeMacros.cmake
zefklop: Nope; this would link msvcrtex after libmsvcrt, and we want it to be before...

CMakeLists.txt
```diff
@@ -115,6 +115,13 @@ else()
     set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE NEVER)
 
     # Add our own target properties
+    # General module definitions
+    define_property(TARGET PROPERTY REACTOS_MODULE_TYPE
+        BRIEF_DOCS "The type of this module"
+        FULL_DOCS [[
+he type of this module.
```
HBelusca, line 122:
```suggestion
The type of this module.
```

CMakeLists.txt
```diff
@@ -115,6 +115,13 @@ else()
...
+One of "nativecui", "nativedll", "kernelmodedriver", "wdmdriver", "kerneldll", "win32cui", "win32gui", "win32dll", "win32ocx", "cpl" or "module"]])
```
HBelusca, line 123: Can you tell me (or refer) what "module" (in that list) means? i.e. What sort of subsystem/default libs/.... values it describes?

Extravert-ir:
> While the crtdll change will clearly increase our iso size, I guess it's the correct thing to do.

The crtdll.dll has been increased by about 2 times. GCC: 653kb vs 313kb, MSVC: 253kb vs 119kb
Btw, this was done like that already 12 years ago :D

Extravert-ir:
What's funny, the build failure originally caused by CMake upgrade from 3.18.3 to 3.18.4
With this change being guilty https://gitlab.kitware.com/cmake/cmake/-/commit/f5d3da091b8e5952f8e880afbd09ee2f18e8ea7f

CMakeLists.txt
zefklop: Just that : a DLL which is not meant to have a particular entry point, nor anything fancy

sdk/cmake/CMakeMacros.cmake
zefklop: Actually, we can, with a bit of trickery :-)

tkreuzer: If we can mark everything that comes from msvcrtex, that is also in msvcrt as private, the order shouldn't matter.

dll/win32/msvcrt/CMakeLists.txt
```diff
@@ -27,7 +35,7 @@ target_link_libraries(msvcrt crt wine ${PSEH_LIB})
 if(MSVC)
     # export of deleting destructor "name"
     add_target_link_flags(msvcrt "/ignore:4102")
-    add_target_property(libmsvcrt STATIC_LIBRARY_FLAGS "/ignore:4102")
+    add_target_property(libmsvcrt_real STATIC_LIBRARY_FLAGS "/ignore:4102")
```
Extravert-ir, line 38: Change to `STATIC_LIBRARY_OPTIONS` while we're here?

sdk/cmake/CMakeMacros.cmake
zefklop: True, but then you have to link msvcrtex with libmsvcrt anyway, so you end up with a libmsvcrt msvcrtex libmsvcrt linking order

dll/win32/msvcrt/CMakeLists.txt
zefklop: Oh yeah. And target_link_options also

PR #3312, opened 2020-10-22
[CRT] Remove duplicate atexit/_onexit

tkreuzer:
This improves the CRT situation a bit by removing the duplicated stuff and use the same code in msvcrtex as is exported from msvcrt. It also adds a simple apitest to validate that it works correctly.

### DIFF
--- a/modules/rostests/apitests/crt/atexit.c
+++ b/modules/rostests/apitests/crt/atexit.c
@@ -1,0 +1,93 @@
+/*
+* PROJECT:         ReactOS API tests
+* LICENSE:         LGPLv2.1+ - See COPYING.LIB in the top level directory
+* PURPOSE:         Test for atexit
+* PROGRAMMER:      Timo Kreuzer <timo.kreuzer@reactos.org>
+*/
+
+#include <apitest.h>
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <ntndk.h>
+
+int g_sequence = 0;
+HANDLE g_hSemaphore;
+
+void exitfunc1(void)
+{
+    ok_int(g_sequence, 1);
+    g_sequence++;
+    ReleaseSemaphore(g_hSemaphore, 1, NULL);
+}
+
+void exitfunc2(void)
+{
+    ok_int(g_sequence, 2);
+    g_sequence++;
+    ReleaseSemaphore(g_hSemaphore, 1, NULL);
+}
+
+void exitfunc3(void)
+{
+    ok_int(g_sequence, 0);
+    g_sequence++;
+    ReleaseSemaphore(g_hSemaphore, 1, NULL);
+    printf("exitfunc3\n");
+}
+
+typedef int (__cdecl *PFN_atexit)(void (__cdecl*)(void));
+
+void Test_atexit()
+{
+    HMODULE hmod;
+    PFN_atexit patexit;
+
+    /* Open the named sempahore to count atexit callbacks */
+    g_hSemaphore = OpenSemaphoreA(SEMAPHORE_ALL_ACCESS, FALSE, "atext_apitest_sempahore");
+    ok(g_hSemaphore != NULL, "couldn't open semaphore.\n");
+
+    /* Load atexit from msvcrt.dll */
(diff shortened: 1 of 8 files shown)


modules/rostests/apitests/crt/atexit.c
```diff
@@ -0,0 +1,69 @@
...
+
+void exitfunc2(void)
+{
+    ok_int(g_sequence, 2);
+    g_sequence++;
+}
+
+void exitfunc3(void)
+{
+    ok_int(g_sequence, 0);
```
learn-more, line 29: This checks the sequence, but there is no verification that they are _actually_ called

sdk/lib/crt/startup/atonexit.c
```diff
@@ -24,13 +24,31 @@
...
 extern _onexit_t (__cdecl * __MINGW_IMP_SYMBOL(_onexit)) (_onexit_t func);
 
+/* INTERNAL: call atexit functions */
+void __call_atexit(void)
+{
+    /* Note: should only be called with the exit lock held */
+    _PVFV *first, *last;
+
+    first =  (_PVFV *)_decode_pointer(__onexitbegin);
+    last = (_PVFV *)_decode_pointer(__onexitend);;
```
learn-more, line 37: nit: double comma

sdk/lib/crt/startup/crtexe.c
```diff
@@ -324,6 +326,7 @@ __tmainCRTStartup (void)
 #endif
     mainret = main (argc, argv, envp);
 #endif
+    __call_atexit();
```
learn-more, line 329: isnt the dll responsible for calling this?
or is the crtexe for the case where the crt is linked into the exe?

modules/rostests/apitests/crt/atexit.c
tkreuzer: Any suggestion how to fix that?

sdk/lib/crt/startup/crtexe.c
tkreuzer: atexit is executed on a per-module base. First the registered entries within the exe are processed (from the code tat is linked into the exe), then msvcrt (and any other loaded dll) processes it's own entries (te corresponding __call_atexit is in crtdll.c). This is why exitfunc2 is last in the sequence, since it was registered through msvcrt.dll, not in the current module.

sdk/lib/crt/startup/atonexit.c
```diff
@@ -24,13 +24,31 @@
...
+extern _onexit_t __cdecl __dllonexit (_onexit_t, _PVFV**, _PVFV**);
 extern _onexit_t (__cdecl * __MINGW_IMP_SYMBOL(_onexit)) (_onexit_t func);
 
+/* INTERNAL: call atexit functions */
+void __call_atexit(void)
+{
+    /* Note: should only be called with the exit lock held */
+    _PVFV *first, *last;
+
+    first =  (_PVFV *)_decode_pointer(__onexitbegin);
```
HBelusca, line 36:
```suggestion
    first = (_PVFV *)_decode_pointer(__onexitbegin);
```

sdk/lib/crt/startup/atonexit.c
```diff
@@ -24,13 +24,31 @@
...
+    last = (_PVFV *)_decode_pointer(__onexitend);;
+
+    if (!first) return;
+
+    while (--last >= first)
+        if (*last)
+            (**last)();
```
HBelusca, line 43:
```suggestion
    while (--last >= first)
    {
        if (*last)
            (**last)();
    }
```
unless this was coming from 3rd party code (wine...)?

modules/rostests/apitests/crt/atexit.c
learn-more: Signalling a named event or something from the handler is probably the easiest,
You can check that it's signalled after the process exits

tkreuzer: I used a named sempahore to count the callbacks

learn-more (approved):

PR #3313, opened 2020-10-23
[NTOSKRNL] Use MI_IS_WRITE_ACCESS(), instead of magic values

SergeGautherie:
Explicit macros vs magic values.

### DIFF
--- a/ntoskrnl/include/internal/amd64/mm.h
+++ b/ntoskrnl/include/internal/amd64/mm.h
@@ -118,9 +118,16 @@
 #endif
 
 /* Macros to identify the page fault reason from the error code */
-#define MI_IS_NOT_PRESENT_FAULT(FaultCode) !BooleanFlagOn(FaultCode, 0x1)
-#define MI_IS_WRITE_ACCESS(FaultCode) BooleanFlagOn(FaultCode, 0x2)
-#define MI_IS_INSTRUCTION_FETCH(FaultCode) BooleanFlagOn(FaultCode, 0x10)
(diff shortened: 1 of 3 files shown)


Extravert-ir (changes requested):
1. I don't think this variable is needed at all
2. You're not consistent: not all usages are fixed

SergeGautherie:
>     1. I don't think this variable is needed at all

Removed.

>     2. You're not consistent: not all usages are fixed

Which other usages?

Extravert-ir:
`ntoskrnl/kdbg/kdb.c` for example

SergeGautherie:
> `ntoskrnl/kdbg/kdb.c`

Thanks for the hint!
Done.

tkreuzer (approved):

ntoskrnl/include/internal/arm/mm.h
```diff
@@ -91,9 +91,16 @@
...
+#define MI_IS_NOT_PRESENT_FAULT(FaultCode)  TRUE
+#define MI_IS_WRITE_ACCESS(FaultCode)       TRUE
+// 0x00000004: user-mode access.
+// TODO: Confirm which value to use.
+#define MI_IS_RESERVED_BIT_FAULT(FaultCode) FALSE
+#define MI_IS_INSTRUCTION_FETCH(FaultCode)  FALSE
+// 0x00000020: protection-key violation.
+// Bits 6-14: reserved.
+// 0x00008000: violation of SGX-specific access-control requirements.
+// Bits 16-31: reserved.
```
ThFabba, line 103: These bits are x86-specific, so this is nonsensical.

SergeGautherie: Right.
Replaced with a better TODO.

PR #3314, opened 2020-10-23
[CMAKE] Make unattended bootcd configurable via cmake

zefklop:
Instead of messing around with a file in the source tree and risk commiting it

use cmake -DUNATTENDED_BOOTCD=yes to enable it, and use the UNATTENDED_BOOTCD_* cache variables to configure

### DIFF
--- a/boot/bootdata/CMakeLists.txt
+++ b/boot/bootdata/CMakeLists.txt
@@ -1,5 +1,6 @@
 
 add_subdirectory(packages)
+add_subdirectory(bootcd)
 
 # Common hives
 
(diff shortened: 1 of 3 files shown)


learn-more:
This file is meant to be locally configured beyond a simple few parameters, for example I set the resolution of my vm in this thing.

If you are scared of committing it accidentally, why not use something like
```git update-index --assume-unchanged boot/bootdata/bootcd/unattend.inf```

binarymaster:
This is very questionable improvement, I personally prefer just modifying the file, since re-configure takes some system resources and time.

You should probably ask developers or at least make a poll in the chat before committing this...

PR #3320, opened 2020-10-25
[GDI32] Fix up GdiFixUpHandle for x64

tkreuzer:
Adds a test for GdiFixUpHandle and fix GdiFixUpHandle accordingly

### DIFF
--- a/modules/rostests/apitests/gdi32/GetObject.c
+++ b/modules/rostests/apitests/gdi32/GetObject.c
@@ -344,6 +344,21 @@ Test_Brush(void)
 
     ok(GetObjectW((HANDLE)GDI_OBJECT_TYPE_BRUSH, sizeof(LOGBRUSH), &logbrush) == 0, "\n");
     ok(GetLastError() == ERROR_INVALID_PARAMETER, "expected ERROR_INVALID_PARAMETER, got %ld\n", GetLastError());
+
+    /* Test handle fixup */
+    hBrush = (HBRUSH)((ULONG_PTR)GetStockObject(WHITE_BRUSH) & 0xFFFF);
+    ok(GetObjectW(hBrush, sizeof(LOGBRUSH), &logbrush) == sizeof(LOGBRUSH),
+       "GetObject(0x%p, ...) failed.\n", hBrush);
+
+#ifdef _WIN64
+    /* Test upper 32 bits */
+    hBrush = (HBRUSH)((ULONG64)GetStockObject(WHITE_BRUSH) | 0xFFFFFFFF00000000ULL);
+    ok(GetObjectW(hBrush, sizeof(LOGBRUSH), &logbrush) == sizeof(LOGBRUSH),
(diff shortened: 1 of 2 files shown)


win32ss/gdi/gdi32/misc/misc.c
```diff
@@ -745,7 +745,7 @@ GdiFixUpHandle(HGDIOBJ hGdiObj)
     Entry = GdiHandleTable + GDI_HANDLE_GET_INDEX(hGdiObj);
 
     /* Rebuild handle for Object */
-    return hGdiObj = (HGDIOBJ)(((LONG_PTR)(hGdiObj)) | (Entry->Type << GDI_ENTRY_UPPER_SHIFT));
+    return hGdiObj = (HGDIOBJ)(((ULONG_PTR)(hGdiObj)) | (Entry->Type << GDI_ENTRY_UPPER_SHIFT));
```
ThFabba, line 748: Nit: Could remove the pointless assignment while you're here.

modules/rostests/apitests/gdi32/GetObject.c
```diff
@@ -344,6 +344,21 @@ Test_Brush(void)
 
     ok(GetObjectW((HANDLE)GDI_OBJECT_TYPE_BRUSH, sizeof(LOGBRUSH), &logbrush) == 0, "\n");
     ok(GetLastError() == ERROR_INVALID_PARAMETER, "expected ERROR_INVALID_PARAMETER, got %ld\n", GetLastError());
+
+    /* Test handle fixup */
+    hBrush = (HBRUSH)((ULONG_PTR)GetStockObject(WHITE_BRUSH) & 0xFFFF);
```
ThFabba, line 349: Since it's exported, you could also test calling GdiFixUpHandle directly?

HBelusca, line 349: Maybe a comment to "confirm" that it is expected that the stock object result holds in the first 16 bits? (unless this is a "known" fact?)

win32ss/gdi/gdi32/misc/misc.c
ThFabba, line 748: I'm not sure I understand this change, or how it relates to the newly added tests.
Somehow `(Entry->Type << GDI_ENTRY_UPPER_SHIFT)` got sign-extended before? What are the consequences of this, especially given that your test seems to show that the upper 32 bits are ignored?

HBelusca, line 748: I'm not sure the `hGdiObj =` part is really necessary since only the result of the operation is returned and nowhere else after the "return" within this function is the hGdiObj variable used yet again.

HBelusca (approved):
Looks ok otherwise :)

modules/rostests/apitests/gdi32/GetObject.c
tkreuzer: I'm too lazy to write a new test

PR #3323, opened 2020-10-25
[RAPPS] Restore event logging of install/uninstall actions

learn-more:
Rapps used to have this, it got lost during refactoring.
Does it make sense to restore this?
Do we optionally want to increase verbosity? (e.g. log install success/failure, maybe download start/failure/stop?)

### DIFF
--- a/base/applications/rapps/installed.cpp
+++ b/base/applications/rapps/installed.cpp
@@ -138,6 +138,9 @@ BOOL CInstalledApplicationInfo::RetrieveIcon(ATL::CStringW& IconLocation)
 
 BOOL CInstalledApplicationInfo::UninstallApplication(BOOL bModify)
 {
+    if (!bModify)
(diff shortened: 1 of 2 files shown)


SergeGautherie:
Maybe make use of `MSG_SUCCESS_INSTALL` eventually?

learn-more:
> Maybe make use of `MSG_SUCCESS_INSTALL` eventually?

Not 'eventually',
but yes I copypasted too much.

binarymaster (approved):

kernelbin:
Oh It's possibly my mistake losting the code to logging 😥
Sorry for that

learn-more:
> Oh It's possibly my mistake losting the code to logging 😥
> Sorry for that

No problem 👍

But considering the lack of input on this issue, does not seem that many people care about this logging, so maybe we should remove it entirely.

PR #3325, opened 2020-10-27
[SHELL32] Disable text when no user is registered

rileyjamesbell:
Disable the "This version of ReactOS is registered to:" text in the about dialogue
when no user is registered (like on the liveCD).

## Purpose

Have the about dialogue make more sense.

## Proposed changes

Disable the "This version of ReactOS is registered to:" text in the about dialogue when no user is registered (like on the liveCD).


HBelusca (commented):
1. Modifications to wine-synced code should go within
```c
#ifdef __REACTOS__
...
#endif
```
blocks.

2. But then if this text is hidden, which other clues are offered to the ReactOS user that his/her ReactOS installation is not "registered"?

rileyjamesbell:
This is my first time contributing to a large project like this so sorry if I'm doing something stupid or just being a nuisance.

I'll be sure to use the "#ifdef __REACTOS__" next time.
>But then if this text is hidden, which other clues are offered to the ReactOS user that his/her ReactOS installation is not "registered"?

I'm not really sure what a registered version of ReactOS is supposed to be. But it doesn't make much sense to say "This version of ReactOS is registered to:" and just some blank lines.
Maybe replace the text with something like "This version of ReactOS is not registered"?

PS. Would maintainers be able to modify the commit with these revisions or would I have to make a new pull request?

HBelusca:
> Maybe replace the text with something like "This version of ReactOS is not registered"?

That could work too (but then this would require adding new translations for this string).

> PS. Would maintainers be able to modify the commit with these revisions or would I have to make a new pull request?

You can always yourself add commits to the PR by committing and pushing into the branch from which you made the PR.

binarymaster:
> I'll be sure to use the "#ifdef REACTOS" next time.

@rileyjamesbell it seems you are not very active, so to avoid your PR stalling here forever, I took over your patch and simplified it to minimize the final diff. Check it out.

rileyjamesbell:
Cool! This change isn't specific to ReactOS, perhaps it should be upstreamed to Wine?

binarymaster:
> This change isn't specific to ReactOS, perhaps it should be upstreamed to Wine?

How so? The text label we're hiding clearly belongs to ReactOS :slightly_smiling_face:

PR #3326, opened 2020-10-27
[BROWSEUI][PSDK] Stubplement ShellTaskScheduler CORE-17337

oleg-dubinskiy:
## Purpose

Add initial implementation of `ShellTaskScheduler` class with `IShellTaskScheduler` interface methods, according to https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nn-shobjidl_core-ishelltaskscheduler. That class seems to have only one appropriate interface.
Required by MS Picture and Fax Viewer (shimgvw.dll). Now it starts properly without replacing any other system files. So adding only a basic stubs is enough to fix the problem.

JIRA issue: [CORE-17337](https://jira.reactos.org/browse/CORE-17337)

## Proposed changes

- Add the `CShellTaskScheduler` class declaration and stubplementation, CLSID declaration is copied from https://git.reactos.org/?p=reactos.git;a=blob;f=modules/rostests/apitests/com/com_apitest.h;hb=b2cf5a4c03a31b48b6167aeba3465ff33b866c5f#l94.
- Add the `IShellTaskScheduler` interface declaration and stub all methods. Btw it's already declared globally in https://git.reactos.org/?p=reactos.git;a=blob;f=sdk/include/psdk/shobjidl.idl;hb=b2cf5a4c03a31b48b6167aeba3465ff33b866c5f#l4614.
- Enable using them for browseui's `DllGetClassObject`, which uses ATL.
- Add rgs resource with required registry data, which is required for proper loading the class, and include it in the main resource file. It can be added into registry by registering browseui via `regsvr32 /i browseui.dll` command.
- Update precompiled header and CMake stuff accordingly.

## Result

After my chages, it starts as well when trying to open any image (after doing all reproducing steps):
MS_shimgvw_fixed

But it does not open the images yet.
According to my analysis, it also requires `NamespaceWalker` class from our shell32.dll (which is currently missing), and the proper implementation of `INamespaceWalk*` methods (just stubs don't work).
I guess it will be another issue.


binarymaster (approved):

dll/win32/browseui/CShellTaskScheduler.h
```diff
@@ -0,0 +1,36 @@
...
+{
+private:
+public:
+    CShellTaskScheduler();
+    ~CShellTaskScheduler();
+
+    // *** IShellTaskScheduler methods ***
+    STDMETHODIMP AddTask(IRunnableTask *pTask, REFGUID rtoid, DWORD_PTR lParam, DWORD dwPriority);
+    STDMETHODIMP RemoveTasks(REFGUID rtoid, DWORD_PTR lParam, BOOL fWaitIfRunning);
+    virtual UINT STDMETHODCALLTYPE CountTasks(REFGUID rtoid);
```
learn-more, line 23: Why not `STDMETHODIMP_(UINT)` instead of `virtual UINT STDMETHODCALLTYPE` ?

oleg-dubinskiy: Because I even didn't know that `STDDMETHODIMP` calling conversion support multiple data types. Sorry.
I'll follow up this in new PR then. :ok_hand:

learn-more: > Because I even didn't know that `STDMETHODIMP` calling conversion supports multiple data types. Sorry.
> I'll follow up this in new PR then. 👌

There is an underscore in that name, that is different from the name without the underscore.

oleg-dubinskiy: Ah, indeed. :sweat_smile: So this is a different conversion. Thanks.

PR #3330, opened 2020-10-28
[CMD][TRANSLATION] update zh-CN translation (partially)

kernelbin:
update cmd translation (zh-CN) partially

help needed:

- I translated ```auxiliary device``` as ```外置设备```，and do not know is this translation good enough. 

- I think ```alternate data streams``` refers to NTFS multiple data streams, so I translated it so. But I'm not sure about this.

BTW I think manual about CTTY command in English is confusing, possibly
```To return control to the standard console, type: CTTY CON on the auxiliary\n\```
should be changed to 
```To return control to the default console, type: CTTY CON on the auxiliary\n\```

English is not my mother tongue, so I'm not sure about this.

### CODE REVIEW

File: base/shell/cmd/lang/zh-CN.rc [hunk 1]
```diff
@@ -148,7 +149,7 @@ DELAY [/m]n\n\n\
                G  分组目录在前                    -  反转顺序的前缀\n\
   /P          在每屏幕信息输出后停顿。\n\
   /Q          显示文件的所有者。\n\
-  /R          Displays alternate data streams of the files.\n\
+  /R          显示文件的 NTFS 数据流。\n\
```
HBelusca, line 152: They indeed refer to these, but the concept of alternate data streams is not specific just for NTFS. So having some translation that does not refer to "NTFS" would be better.

kernelbin: ok, I will change a way to translate it🤣

PR #3334, opened 2020-10-30
[HALX86] Implement HalQueryAMLIIllegalIOPortAddresses case in HaliQuerySystemInformation CORE-17359

oleg-dubinskiy:
## Purpose

Implement `HalQueryAMLIIllegalIOPortAddresses` case in `HaliQuerySystemInformation`.
Based on the patch by @vgalnt.

JIRA issue: [CORE-17359](https://jira.reactos.org/browse/CORE-17359)

### DIFF
--- a/hal/halx86/generic/sysinfo.c
+++ b/hal/halx86/generic/sysinfo.c
@@ -4,6 +4,7 @@
  * FILE:            hal/halx86/generic/sysinfo.c
  * PURPOSE:         HAL Information Routines
  * PROGRAMMERS:     Alex Ionescu (alex.ionescu@reactos.org)
+ *                  Vadim Galyant (vgal@rambler.ru)
  */
 
 /* INCLUDES *******************************************************************/
@@ -12,7 +13,42 @@
 #define NDEBUG
 #include <debug.h>
 
+HAL_AMLI_BAD_IO_ADDRESS_LIST HalALMIBadIOAddressList[] =
+{
+    { 0x0000, 0x10, 1, NULL }, // DMA controller
+    { 0x0020, 0x02, 0, NULL }, // Programmable Interrupt Controller (8259A)
+    { 0x0040, 0x04, 1, NULL }, // System Timer 1
+    { 0x0048, 0x04, 1, NULL }, // System Timer 2 failsafe
+    { 0x0070, 0x02, 1, NULL }, // Real-time clock
+    { 0x0074, 0x03, 1, NULL }, // Extended CMOS
+    { 0x0081, 0x03, 1, NULL }, // DMA 1 page registers
+    { 0x0087, 0x01, 1, NULL }, // DMA 1 Ch 0 low page
+    { 0x0089, 0x01, 1, NULL }, // DMA 2 page registers
+    { 0x008A, 0x02, 1, NULL }, // DMA 2 page registers
+    { 0x008F, 0x01, 1, NULL }, // DMA 2 low page refresh
+    { 0x0090, 0x02, 1, NULL }, // Arbitration control
+    { 0x0093, 0x02, 1, NULL }, // Reserved system board setup
+    { 0x0096, 0x02, 1, NULL }, // POS channel select
+    { 0x00A0, 0x02, 0, NULL }, // Cascaded PIC
+    { 0x00C0, 0x20, 1, NULL }, // ISA DMA
+    { 0x04D0, 0x02, 0, NULL }, // PIC edge/level registers
+    { 0x0CF8, 0x08, 1, &HaliHandlePCIConfigSpaceAccess }, // PCI configuration space
+    { 0x0000, 0x00, 0, NULL } // Reserved
+};
+
 /* FUNCTIONS ******************************************************************/
+
+NTSTATUS
+NTAPI
+HaliHandlePCIConfigSpaceAccess(_In_ BOOLEAN IsRead,
+                               _In_ ULONG Port,
+                               _In_ ULONG Length,
+                               _Inout_ PULONG Buffer)
+{
+    DPRINT1("HaliHandlePCIConfigSpaceAccess: IsRead %X, Port 0x%X, Length %u, Buffer %p\n", IsRead, Port, Length, Buffer);
+    //ASSERT(FALSE);
+    return STATUS_NOT_IMPLEMENTED;
+}
 
 NTSTATUS
 NTAPI
@@ -44,7 +80,22 @@ HaliQuerySystemInformation(IN HAL_QUERY_INFORMATION_CLASS InformationClass,
 		REPORT_THIS_CASE(HalCmcLogInformation);
 		REPORT_THIS_CASE(HalCpeLogInformation);
 		REPORT_THIS_CASE(HalQueryMcaInterface);
-		REPORT_THIS_CASE(HalQueryAMLIIllegalIOPortAddresses);
+		case HalQueryAMLIIllegalIOPortAddresses:
+		{
+			ULONG Size = sizeof(HalALMIBadIOAddressList);
+			NTSTATUS Status = STATUS_INFO_LENGTH_MISMATCH;
+
+			if (BufferSize >= Size)
+			{
+				RtlCopyMemory(Buffer, HalALMIBadIOAddressList, Size);
+				Status = STATUS_SUCCESS;
+			}
+
+			*ReturnedLength = Size;
+			KeFlushWriteBuffer();
+			return Status;
+		}
+
 		REPORT_THIS_CASE(HalQueryMaxHotPlugMemoryAddress);
 		REPORT_THIS_CASE(HalPartitionIpiInterface);
 		REPORT_THIS_CASE(HalPlatformInformation);
--- a/hal/halx86/include/halp.h
+++ b/hal/halx86/include/halp.h
@@ -318,6 +318,15 @@ HalpUnmapVirtualAddress(
 /* sysinfo.c */
 NTSTATUS
 NTAPI
+HaliHandlePCIConfigSpaceAccess(
+    _In_ BOOLEAN IsRead,
+    _In_ ULONG Port,
+    _In_ ULONG Length,
+    _Inout_ PULONG Buffer
+);
+
+NTSTATUS
+NTAPI
 HaliQuerySystemInformation(
     IN HAL_QUERY_INFORMATION_CLASS InformationClass,
     IN ULONG BufferSize,


hal/halx86/generic/sysinfo.c
```diff
@@ -44,7 +44,11 @@ HaliQuerySystemInformation(IN HAL_QUERY_INFORMATION_CLASS InformationClass,
 		REPORT_THIS_CASE(HalCmcLogInformation);
 		REPORT_THIS_CASE(HalCpeLogInformation);
 		REPORT_THIS_CASE(HalQueryMcaInterface);
-		REPORT_THIS_CASE(HalQueryAMLIIllegalIOPortAddresses);
+		case HalQueryAMLIIllegalIOPortAddresses:
+		{
+			/* FIXME: TODO */
+			return STATUS_NOT_IMPLEMENTED;
```
ThFabba, line 50: I'm unclear on how this can fix anything. `REPORT_THIS_CASE` uses `break`, which will exit the `switch` and also return `STATUS_NOT_IMPLEMENTED`. So the only difference seems to be debug prints?

oleg-dubinskiy: Probably the difference is I didn't use `break` in my stub, so it fixed the problem? :thinking:

binarymaster: But return code is the same anyway, so the only difference is missing `UNIMPLEMENTED` :raised_eyebrow: That's strange

HBelusca: To test the hypothesis of the DPRINT vs. no DPRINT stuff @oleg-dubinskiy , I suggest you try this PR but where you add an explicit `UNIMPLEMENTED` or a `DPRINT1("Not implemented\n")` command just before your `return STATUS_NOT_IMPLEMENTED;` . If it crashes again systematically we will know that for some reason (still not understood) the DPRINT triggers the BSOD.

ThFabba (changes requested):
I'm pretty sure this cannot be related to the bug check. I think it's more likely that:
* this is a "fourth stage" situation (e.g. we create the registry values that acpi is missing in 3rd stage, therefore you get the error if you replace acpi after 2nd stage but you don't get it if you first boot 3rd stage, and then replace it on the next boot),
* ROS did not boot in ACPI mode in the case where it worked
* there's a race condition where we do create the necessary registry value, but we create it at a point in time that could sometimes come before acpi.sys is loaded and sometimes after

hal/halx86/generic/sysinfo.c
oleg-dubinskiy: No @HBelusca. :smiley: Still booting successfully for me, as with `UNIMPLEMENTED`, same with `DPRINT1("Not implemented\n")`. And what is the most surprising, is I still don't see that debug print from the log at all, in both cases.
So most of all, the problem is in something other. It seems to me that MS implementation of acpi.sys just needs a proper definition of that case, without any macros.

oleg-dubinskiy:
Looks like you are right @ThFabba regarding the 1st and 2nd things.
I just tried to replace acpi.sys before 3rd setup stage (with ACPI Uniprocessor setup), using Linux/ReactOS LiveCD, and it indeed doesn't boot correctly at 3rd setup stage, fails exactly with the same BSOD. And with `DPRINT1` added, now I see the following in the log before BSOD:
`(hal/halx86/generic/sysinfo.c:50) HalQueryAMLIIllegalIOPortAddresses not implemented!`.
This happens most of all due to that missing registry data, because I also see the following lines:
`(/ntoskrnl/config/cmvalche.c:665) Caller requested unknown class: 4`.
Sadly that it still doesn't fix the actual problem properly. :cry: 
So I doubt I should close this PR, or there is still a reason don't do that?... :thinking:

hal/halx86/generic/sysinfo.c
HBelusca: But if the macro just resolves to the very exact same code as the one written in "expanded" form, this does not make any sense, and I would tend to agree with @ThFabba that it is actually a result of some sort of race condition.

vgalnt:
Why make another stub instead of one?
https://jira.reactos.org/browse/CORE-17359

hal/halx86/generic/sysinfo.c
oleg-dubinskiy: OK, now I updated the PR by the proper implementation of this, instead of stub. But it still doesn't fix CORE-17358. It even didn't improve the situation at all! :frowning_face: 
So I linked CORE-17359 instead.
Thanks to @vgalnt for the provided patch with this implementation. :+1:

hal/halx86/generic/sysinfo.c
```diff
@@ -12,8 +12,50 @@
...
 /* FUNCTIONS ******************************************************************/
 
+NTSTATUS
+NTAPI
+HaliHandlePCIConfigSpaceAccess(BOOLEAN IsRead,
+                               ULONG Port,
+                               ULONG Length,
+                               PVOID Buffer)
+{
+    DPRINT1("HaliHandlePCIConfigSpaceAccess: IsRead %X, Port %X, Length %X, Buffer %p\n", IsRead, Port, Length, Buffer);
```
binarymaster, line 54: Improve debug print, but I'm not sure about boolean format.
```suggestion
    DPRINT1("HaliHandlePCIConfigSpaceAccess: IsRead %X, Port 0x%X, Length %u, Buffer %p\n", IsRead, Port, Length, Buffer);
```

hal/halx86/generic/sysinfo.c
```diff
@@ -24,6 +66,24 @@ HaliQuerySystemInformation(IN HAL_QUERY_INFORMATION_CLASS InformationClass,
...
+                *ReturnedLength = Size;
+                KeFlushWriteBuffer();
+                return STATUS_INFO_LENGTH_MISMATCH;
+            }
+
+            RtlCopyMemory(Buffer, AMLIIllegalIOPortAddresses, Size);
+
+            *ReturnedLength = Size;
+            KeFlushWriteBuffer();
+            return STATUS_SUCCESS;
```
binarymaster, line 84:
```suggestion
            NTSTATUS Status = STATUS_INFO_LENGTH_MISMATCH;

            if (BufferSize >= Size)
            {
                RtlCopyMemory(Buffer, AMLIIllegalIOPortAddresses, Size);
                Status = STATUS_SUCCESS;
            }

            *ReturnedLength = Size;
            KeFlushWriteBuffer();
            return Status;
```
Refactoring.

hal/halx86/generic/sysinfo.c
```diff
@@ -44,7 +104,6 @@ HaliQuerySystemInformation(IN HAL_QUERY_INFORMATION_CLASS InformationClass,
 		REPORT_THIS_CASE(HalCmcLogInformation);
 		REPORT_THIS_CASE(HalCpeLogInformation);
 		REPORT_THIS_CASE(HalQueryMcaInterface);
-		REPORT_THIS_CASE(HalQueryAMLIIllegalIOPortAddresses);
```
binarymaster, line 47: I'd suggest to move the implementation to this line, instead of just removing it.

binarymaster (changes requested):
Since you're now contributing a patch from another person, you need to sign the commit with proper author field.

```
git commit --amend --no-edit --author "Vadim Galyant <vgal@rambler.ru>"
```

hal/halx86/include/halp.h
```diff
@@ -185,6 +185,14 @@ typedef struct _HalAddressUsage
     } Element[];
 } ADDRESS_USAGE, *PADDRESS_USAGE;
 
+typedef struct _AMLI_ILLEGAL_IO_PORT_ADDRESSES
```
ThFabba, line 188: Please use the public structure `HAL_AMLI_BAD_IO_ADDRESS_LIST` from `ntddk.h` instead.

hal/halx86/generic/sysinfo.c
learn-more: `typedef unsigned char BOOLEAN, *PBOOLEAN;`

learn-more: `typedef int BOOL;`

hal/halx86/generic/sysinfo.c
```diff
@@ -12,8 +12,50 @@
 #define NDEBUG
 #include <debug.h>
 
+NTSTATUS
+NTAPI
+HaliHandlePCIConfigSpaceAccess(BOOLEAN IsRead,
+                               ULONG Port,
+                               ULONG Length,
+                               PVOID Buffer);
```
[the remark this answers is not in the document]
oleg-dubinskiy: No, there's no any: https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=HaliHandlePCIConfigSpaceAccess.
Maybe move it in some header file then? :thinking:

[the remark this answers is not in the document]
binarymaster: Yes, that's what I meant in this case.

> Maybe move it in some header file then?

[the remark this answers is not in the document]
ThFabba: You don't need this declaration at all.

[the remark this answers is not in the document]
oleg-dubinskiy: Oh, OK. I'll delete this. I wonder why Vadim added it at all.

hal/halx86/generic/sysinfo.c
```diff
@@ -12,8 +12,50 @@
...
+    { 0x0089, 0x01, 1, NULL },
+    { 0x008A, 0x02, 1, NULL },
+    { 0x008F, 0x01, 1, NULL },
+    { 0x0090, 0x02, 1, NULL },
+    { 0x0093, 0x02, 1, NULL },
+    { 0x0096, 0x02, 1, NULL },
+    { 0x00A0, 0x02, 0, NULL },
+    { 0x00C0, 0x20, 1, NULL },
+    { 0x04D0, 0x02, 0, NULL },
+    { 0x0CF8, 0x08, 1, &HaliHandlePCIConfigSpaceAccess },
```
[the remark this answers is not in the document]
oleg-dubinskiy: > `0xCF8` is really PCI config I/O port, there might be a define for it.

Probably these?
https://git.reactos.org/?p=reactos.git;a=blob;f=hal/halx86/include/bus.h;hb=d829bcc629b36ecf38c8a99f2f4098296b071c8a#l164
https://git.reactos.org/?p=reactos.git;a=blob;f=hal/halx86/include/bus.h;hb=d829bcc629b36ecf38c8a99f2f4098296b071c8a#l170
(one of them, most of all the first)

hal/halx86/generic/sysinfo.c
```diff
@@ -12,8 +12,50 @@
...
 #include <debug.h>
 
+NTSTATUS
+NTAPI
+HaliHandlePCIConfigSpaceAccess(BOOLEAN IsRead,
+                               ULONG Port,
+                               ULONG Length,
+                               PVOID Buffer);
+
+AMLI_ILLEGAL_IO_PORT_ADDRESSES AMLIIllegalIOPortAddresses[19] =
```
ThFabba, line 22: You don't need to specify the 19, the compiler will use the length of the initializer.

hal/halx86/generic/sysinfo.c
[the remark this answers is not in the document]
ThFabba: We won't have constants for all of these ports so I think the most valuable thing to do would be to keep the number in the table and add comments for what these are. https://wiki.osdev.org/I/O_Ports#The_list can be a good starting point for that.

hal/halx86/generic/sysinfo.c
oleg-dubinskiy: Committed, thanks.

hal/halx86/generic/sysinfo.c
oleg-dubinskiy: Committed, thanks @binarymaster. :+1:

hal/halx86/generic/sysinfo.c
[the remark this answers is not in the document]
oleg-dubinskiy: Oops. :sweat_smile: Looks like this declaration is badly required, because it is used in the structure below, for `&HaliHandlePCIConfigSpaceAccess` member. After removing this, I got the following compilation error: `../hal/halx86/generic/sysinfo.c:35:25: error: 'HaliHandlePCIConfigSpaceAccess' undeclared here (not in a function); did you mean 'HalHandlerForConfigSpace'?
     { 0x0CF8, 0x08, 1, &HaliHandlePCIConfigSpaceAccess }, // PCI configuration space
                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                         HalHandlerForConfigSpace`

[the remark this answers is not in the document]
oleg-dubinskiy: So I'll keep it as-is.

hal/halx86/generic/sysinfo.c
```diff
@@ -24,6 +66,22 @@ HaliQuerySystemInformation(IN HAL_QUERY_INFORMATION_CLASS InformationClass,
 #define REPORT_THIS_CASE(X) case X: DPRINT1("Unhandled case: %s\n", #X); break
 	switch (InformationClass)
 	{
+        case HalQueryAMLIIllegalIOPortAddresses:
```
ThFabba, line 69: The indentation from the parent block here appears to be 6 spaces. Please either make it 4 (our standard) or 8 (to match the `REPORT_THIS_CASE` lines).

hal/halx86/generic/sysinfo.c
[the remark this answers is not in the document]
ThFabba: Fine with me. You could also move it to before the `HaliQuerySystemInformation` declaration in halp.h.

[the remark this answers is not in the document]
oleg-dubinskiy: Fixed, thanks.

hal/halx86/generic/sysinfo.c
oleg-dubinskiy: Fixed.

hal/halx86/include/halp.h
oleg-dubinskiy: Fixed, thanks.
Also, since there was incompatible pointer types between the 4th parameter of `PHALIOREADWRITEHANDLER`, which is used in `HAL_AMLI_BAD_IO_ADDRESS_LIST` and the 4th parameter of `HaliHandlePCIConfigSpaceAccess`, I got the compilation error. So I changed the parameter of the last one from `PVOID` to `PULONG`. I guess it will be more correct, in case if that array should use the public structure instead of private one declared by Vadim.

hal/halx86/generic/sysinfo.c
oleg-dubinskiy: Fixed.

hal/halx86/generic/sysinfo.c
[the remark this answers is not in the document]
oleg-dubinskiy: Fixed. Thanks @ThFabba.
Btw, I also found the similar reference in our APICA code: https://git.reactos.org/?p=reactos.git;a=blob;f=drivers/bus/acpi/acpica/hardware/hwvalid.c;hb=d829bcc629b36ecf38c8a99f2f4098296b071c8a#l70.

hal/halx86/generic/sysinfo.c
oleg-dubinskiy: Fixed, thanks.
The indentation looked inaccurate because Vadim used the spaces in his code, but in the whole other file, the tabs instead of spaces are used. Now I added the tabs in his code, so now it looks OK.

binarymaster (approved):
LGTM. :+1: But I think the API test in #3339 should be merged first.

ThFabba (approved):
Test results look good.
```
[ROSAUTOTEST] System uptime 108.37 seconds
(ntoskrnl\mm\ARM3\sysldr.c:176) Loading: \??\C:\ReactOS\bin\kmtest_drv.sys at EEF20000 with 129 pages
(sdk\lib\rtl\heap.c:2243) HEAP: Trying to free an invalid address FFFFFFFF!
Running Wine Test, Module: kmtest, Test: HalSystemInfo
err:(dll\win32\advapi32\service\scm.c:2946) RStartServiceW() failed (Error 1056)
(modules\rostests\kmtests\kmtest_drv\kmtest_drv.c:371) DriverIoControl. Starting test HalSystemInfo
(hal\halx86\generic\sysinfo.c:48) HaliHandlePCIConfigSpaceAccess: IsRead 1, Port 0xCF7, Length 1, Buffer EF058AC8
(modules\rostests\kmtests\kmtest_drv\kmtest_drv.c:373) DriverIoControl. Finished test HalSystemInfo
(sdk\lib\rtl\heap.c:2243) HEAP: Trying to free an invalid address FFFFFFFF!
HalSystemInfo.c:143: Test failed: [17] IOHandler returned 0xc0000002
HalSystemInfo: 89 tests executed (0 marked as todo, 1 failures), 0 skipped.
Test HalSystemInfo completed in 0.15 seconds.
```

hal/halx86/include/hal.h
```diff
@@ -26,6 +26,7 @@
 #endif
 
 /* IFS/DDK/NDK Headers */
+#include <ntddk.h>
```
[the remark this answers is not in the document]
oleg-dubinskiy: OK, fixed.

hal/halx86/include/halp.h
```diff
@@ -321,6 +321,15 @@ HalpUnmapVirtualAddress(
 );
 
 /* sysinfo.c */
+NTSTATUS
+NTAPI
+HaliHandlePCIConfigSpaceAccess(
+    BOOLEAN IsRead,
+    ULONG Port,
+    ULONG Length,
+    PULONG Buffer
```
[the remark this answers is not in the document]
oleg-dubinskiy: Done.

PR #3335, opened 2020-10-30
[CMAKE] Introduce libntdll_crt and use it for modules linking to ntdl…

zefklop:
…l but not msvcrt

ROSTESTS-362

## Purpose

Make clear to the linker which imports come from which module instead of relying on linking order (which got broken in 23e04ae1eec666118a2359a17fc442e0332c258b


dll/ntdll/CMakeLists.txt
```diff
@@ -3,6 +3,12 @@ add_subdirectory(nt_0600)
 
 spec2def(ntdll.dll def/ntdll.spec ADD_IMPORTLIB)
 
+# Use a CRT-only import lib for modules needing it
+generate_import_lib(libntdll_crt ntdll.dll def/ntdll_crt.spec)
+
+# We link ntdll_crt to modules not linking to msvcrt
+target_link_libraries(libntdll INTERFACE "$<$<NOT:$<IN_LIST:libmsvcrt,$<TARGET_PROPERTY:LINK_LIBRARIES>>>:libntdll_crt>")
```
tkreuzer, line 10: With this, people could still forget to link msvcrt and would automatically get ntdll's crt. IMO ntdll_crt should only be linked into native dlls/exes, baed on the module type.

tkreuzer (changes requested):
- [ ] Please keep the order of exports in ntdll as it is
- [ ] To make it work, we need to handle -private in spec2def, otherwise it won't work

zefklop:
Made something eles, thanks to an idea from Thomas

PR #3338, opened 2020-10-30
[WIN32K:NTUSER] Fix gradient caption bars with low color depth

disean:
## Purpose

This fixes a problem when the window title bar is dithered down to the 256/16-color palette.
See also [0705f303d054ac636a53ca04d9e41172fd7c36bc](https://git.reactos.org/?p=reactos.git;a=commit;h=0705f303d054ac636a53ca04d9e41172fd7c36bc).

JIRA issue: none

BEFORE:
before

AFTER:
after

### DIFF
--- a/win32ss/user/ntuser/sysparams.c
+++ b/win32ss/user/ntuser/sysparams.c
@@ -1583,7 +1583,17 @@ SpiGetSet(UINT uiAction, UINT uiParam, PVOID pvParam, FLONG fl)
             return SpiSetUserPref(UPM_LISTBOXSMOOTHSCROLLING, pvParam, fl);
 
         case SPI_GETGRADIENTCAPTIONS:
-            return SpiGetUserPref(UPM_GRADIENTCAPTIONS, pvParam, fl);
+        {
+            if (NtGdiGetDeviceCaps(ScreenDeviceContext, BITSPIXEL) <= 8)
+            {
+                INT iValue = 0;
+                return SpiGetInt(pvParam, &iValue, fl);
+            }
+            else
+            {
+                return SpiGetUserPref(UPM_GRADIENTCAPTIONS, pvParam, fl);
+            }
+        }
 
         case SPI_SETGRADIENTCAPTIONS:
             return SpiSetUserPref(UPM_GRADIENTCAPTIONS, pvParam, fl);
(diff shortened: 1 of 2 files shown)


HBelusca:
Does Windows behave in a similar manner?

disean:
Yes, here's my test program
```
int _tmain(int argc, const TCHAR *argv[])
{
    BOOL bGradientCaptions;
    if (SystemParametersInfoW(SPI_GETGRADIENTCAPTIONS, 0, &bGradientCaptions, 0))
        wprintf(L"Return: %u\n", bGradientCaptions);
    else
        wprintf(L"Request failed with %lu\n", GetLastError());
    return 0;
}
```

binarymaster:
I wonder if we have existing API test for this :thinking: 

Should be probably somewhere here: https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=SystemParametersInfoW

image

win32ss/user/ntuser/sysparams.c
```diff
@@ -1583,7 +1583,17 @@ SpiGetSet(UINT uiAction, UINT uiParam, PVOID pvParam, FLONG fl)
             return SpiSetUserPref(UPM_LISTBOXSMOOTHSCROLLING, pvParam, fl);
 
         case SPI_GETGRADIENTCAPTIONS:
-            return SpiGetUserPref(UPM_GRADIENTCAPTIONS, pvParam, fl);
+        {
+            if (NtGdiGetDeviceCaps(ScreenDeviceContext, BITSPIXEL) <= 8)
```
ThFabba, line 1587: Calling NtGdi functions while holding the user lock is a slippery slope. But in this particular case it seems that this is safe.

ThFabba (commented):
The Win32k change looks okay to me. Haven't really looked at the test.

binarymaster (approved):
Confirmed, approved! :+1:
### BEFORE
```
SystemParametersInfo.c:238: Test failed: SPI_GETGRADIENTCAPTIONS got unexpected value (1 instead of 0)

SystemParametersInfo: 92 tests executed (0 marked as todo, 18 failures), 0 skipped.
```
### AFTER
```
SystemParametersInfo: 92 tests executed (0 marked as todo, 17 failures), 0 skipped.
```
### Test on Windows XP
```
SystemParametersInfo: 113 tests executed (0 marked as todo, 27 failures), 0 skipped.
```
(no mention of `SPI_GETGRADIENTCAPTIONS` in log, so the test is correct)

PR #3343, opened 2020-11-01
[CMAKE] Add architecture to our build string

tkreuzer:
This introduces a new CMake variable WINARCH, which is what Windows usually uses ("x86" instead of "i386") e.g. in sxs names.


binarymaster (approved):

CMakeLists.txt
```diff
@@ -32,6 +29,16 @@ endif()
 # for more information.
 string(TOLOWER ${ARCH} ARCH)
 
+# alternative architecture string
```
HBelusca, line 32:
```suggestion
# Alternative WinNT-compatible architecture string
```

HBelusca (approved):

learn-more (approved):

ThFabba (approved):

Extravert-ir:
Will leaving `i386` actually break anything? Maybe we can avoid another abstraction here :)

PR #3346, opened 2020-11-01
[KERNEL32] Fix register initialization in BaseInitializeContext

tkreuzer:
Get rid of BaseThreadStartupThunk and BaseProcessStartThunk asm wrappers and go to the C functions directly (home space is allocated on the stack by the kernel)


HBelusca (commented):
Maybe you need to explain why this is however needed in other architectures, say, in x86 32 bits?

tkreuzer:
> 
> 
> Maybe you need to explain why this is however needed in other architectures, say, in x86 32 bits?

No, I don't think I need to explain why other people wrote their code how they wrote their code.

PR #3348, opened 2020-11-01
[NTOS:CC] Raise status, when CcMapData fails

tkreuzer:
Also add a check in fastfat, that asserts, when CcMapData returns FALSE.


ntoskrnl/cc/pin.c
```diff
@@ -427,7 +427,9 @@ CcMapData (
             if (iBcb == NULL)
             {
                 CcRosReleaseVacb(SharedCacheMap, Vacb, TRUE, FALSE, FALSE);
+                *pBcb = NULL; // If you ever remove this for compat, make sure to review all callers for using an unititialized value
```
HBelusca, line 430: 😆

zefklop:
I'm in the process of rewriting all of this. Please hold your horses before merging

SergeGautherie:
What is left of this PR, after switch to fastfat_new?

zefklop:
This is still needed.

PR #3349, opened 2020-11-02
[CMAKE] Turn import libs into regular C STATIC libraries for GCC builds

zefklop:
## Purpose

This lets us embed msvcrtex into libmsvcrt and keep library linking order, somethiong which INTERFACE libraries doesn't let us do

JIRA issue: [ROSTESTS-361](https://jira.reactos.org/browse/ROSTESTS-361)

## Proposed changes

Use DLLTOOL to produce object files instead of an import library, and add them into a regular static library.


base/applications/network/ftp/CMakeLists.txt
```diff
@@ -12,12 +10,14 @@ list(APPEND SOURCE
     precomp.h)
 
 add_executable(ftp ${SOURCE} ftp.rc)
+target_compile_definitions(ftp PRIVATE lint)
+
 set_module_type(ftp win32cui)
 add_importlibs(ftp ws2_32 iphlpapi msvcrt kernel32)
+target_link_libraries(ftp oldnames)
```
ThFabba, line 17: I think as a rule, we want `target_link_libraries` before `add_importlibs` (since libraries will often depend on things from dlls)

dll/win32/msvcrt/CMakeLists.txt
```diff
@@ -1,14 +1,12 @@
...
-# The msvcrt <-> msvcrtex trick
-generate_import_lib(libmsvcrt_real msvcrt.dll msvcrt.spec)
-add_library(libmsvcrt INTERFACE)
+# Let consumers of msvcrt have the right defines
 target_compile_definitions(libmsvcrt INTERFACE _DLL __USE_CRTIMP)
-# if the linked module is one of win32gui;win32cui;win32dll;win32ocx;cpl link it with msvcrtex, which itself is linked to libmsvcrt_real
-# Otherwise, just link to libmsvcrt_real
-target_link_libraries(libmsvcrt INTERFACE "$<IF:$<IN_LIST:$<TARGET_PROPERTY:REACTOS_MODULE_TYPE>,win32gui;win32cui;win32dll;win32ocx;cpl>,msvcrtex,libmsvcrt_real>")
+# Embed msvcrtex into libmsvcrt
+target_sources(libmsvcrt PRIVATE $<TARGET_OBJECTS:msvcrtex>)
```
ThFabba, line 8: Nice!

sdk/cmake/gcc.cmake
```diff
@@ -315,21 +315,72 @@ function(fixup_load_config _target)
...
-    set_target_properties(${_libname} PROPERTIES LINKER_LANGUAGE "IMPLIB" PREFIX "")
-
-    # Create delayed importlib
-    _add_library(${_libname}_delayed STATIC EXCLUDE_FROM_ALL ${CMAKE_CURRENT_BINARY_DIR}/${_libname}_implib.def)
-    set_target_properties(${_libname}_delayed PROPERTIES LINKER_LANGUAGE "IMPLIB_DELAYED" PREFIX "")
+    # With this, we let DLLTOOL create the object files
+    set(LIBRARY_PRIVATE_DIR ${CMAKE_CURRENT_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/${_libname}.dir)
+    add_custom_command(
+        OUTPUT ${LIBRARY_PRIVATE_DIR}/${_libname}t.o ${LIBRARY_PRIVATE_DIR}/${_libname}h.o
+        COMMAND ${CMAKE_COMMAND} -E rm -f ${LIBRARY_PRIVATE_DIR}/${_libname}s*.o # Remove all previously generated files
```
ThFabba, line 335: Should these be specified as `BYPRODUCTS`?
Also, why do we have to delete these, does dlltool not overwrite them?

sdk/cmake/gcc.cmake
```diff
@@ -315,21 +315,72 @@ function(fixup_load_config _target)
...
+        COMMAND ${CMAKE_DLLTOOL} --def ${CMAKE_CURRENT_BINARY_DIR}/${_libname}_implib.def --kill-at --output-lib=${NULDEV} -t ${_libname} -n -n # Produce obj files
+        DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/${_libname}_implib.def
+        WORKING_DIRECTORY ${LIBRARY_PRIVATE_DIR})
+    set_source_files_properties(
+        ${LIBRARY_PRIVATE_DIR}/${_libname}t.o
+        ${LIBRARY_PRIVATE_DIR}/${_libname}s.o
```
ThFabba, line 341: Don't you mean `h.o`?

sdk/cmake/gcc.cmake
```diff
@@ -315,21 +315,72 @@ function(fixup_load_config _target)
...
-
-    # Create delayed importlib
-    _add_library(${_libname}_delayed STATIC EXCLUDE_FROM_ALL ${CMAKE_CURRENT_BINARY_DIR}/${_libname}_implib.def)
-    set_target_properties(${_libname}_delayed PROPERTIES LINKER_LANGUAGE "IMPLIB_DELAYED" PREFIX "")
+    # With this, we let DLLTOOL create the object files
+    set(LIBRARY_PRIVATE_DIR ${CMAKE_CURRENT_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/${_libname}.dir)
+    add_custom_command(
+        OUTPUT ${LIBRARY_PRIVATE_DIR}/${_libname}t.o ${LIBRARY_PRIVATE_DIR}/${_libname}h.o
+        COMMAND ${CMAKE_COMMAND} -E rm -f ${LIBRARY_PRIVATE_DIR}/${_libname}s*.o # Remove all previously generated files
+        COMMAND ${CMAKE_DLLTOOL} --def ${CMAKE_CURRENT_BINARY_DIR}/${_libname}_implib.def --kill-at --output-lib=${NULDEV} -t ${_libname} -n -n # Produce obj files
```
ThFabba, line 336: > -n --no-delete            Keep temp files (repeat for extra preservation)

Starting to sound like we should write our own dlltool. That might fix delay imports too ;)

sdk/cmake/msvc.cmake
```diff
@@ -321,13 +321,11 @@ function(generate_import_lib _libname _dllname _spec_file)
     if(MSVC_IDE)
         add_library(${_libname} STATIC EXCLUDE_FROM_ALL ${_asm_stubs_file}.obj)
         set_source_files_properties(${_asm_stubs_file}.obj PROPERTIES EXTERNAL_OBJECT TRUE)
-        set_target_properties(${_libname} PROPERTIES LINKER_LANGUAGE "C")
     else()
         # NOTE: as stub file and def file are generated in one pass, depending on one is like depending on the other
         add_library(${_libname} STATIC EXCLUDE_FROM_ALL ${_asm_stubs_file})
-        # set correct "link rule"
-        set_target_properties(${_libname} PROPERTIES LINKER_LANGUAGE "IMPLIB")
```
ThFabba, line 329: If we're no longer using the IMPLIB language, you can remove these: https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=_IMPLIB_

sdk/cmake/msvc.cmake
```diff
@@ -321,13 +321,11 @@ function(generate_import_lib _libname _dllname _spec_file)
...
     endif()
+    set_target_properties(${_libname} PROPERTIES LINKER_LANGUAGE "C")
```
ThFabba, line 328: This will make clang-cl on GitHub Actions unhappy. It defaults to using `llvm-lib` with the latest CMake. We have an explicit override to use `link /lib` for IMPLIB linking, which was how I fixed this.

With this, we may need to override `CMAKE_C_CREATE_STATIC_LIBRARY` on Clang-CL (or all MSVC) builds.

win32ss/gdi/gdi32/CMakeLists.txt
```diff
@@ -57,10 +57,9 @@ set_module_type(gdi32
 
 target_link_libraries(gdi32
     winegdi
-    wine 
+    wine
     win32ksys
     dxguid
-    msvcrtex
```
ThFabba, line 63: Wait, so this was unnecessary this whole time?

sdk/lib/crt/msvcrtex.cmake
```diff
@@ -93,12 +93,9 @@ endif()
 set_source_files_properties(${MSVCRTEX_ASM_SOURCE} PROPERTIES COMPILE_DEFINITIONS "_DLL;_MSVCRTEX_")
 add_asm_files(msvcrtex_asm ${MSVCRTEX_ASM_SOURCE})
 
-add_library(msvcrtex ${MSVCRTEX_SOURCE} ${msvcrtex_asm})
+add_library(msvcrtex OBJECT ${MSVCRTEX_SOURCE} ${msvcrtex_asm})
```
ThFabba, line 96: Making msvcrtex an object library caused problems in gdi32/user32 for me (duplicate symbols, since all the objects are now linked in unconditionally). I just reverted to a static library to fix that.

sdk/cmake/CMakeMacros.cmake
```diff
@@ -577,8 +577,11 @@ function(set_module_type MODULE TYPE)
         add_target_link_flags(${MODULE} "-Wl,--major-os-version,5 -Wl,--minor-os-version,01")
     endif()
 
-    # Set unicode definitions
+    # Set unicode property
     if(__module_UNICODE)
+        set_target_properties(${MODULE} PROPERTIES UNICODE_MODULE TRUE)
+        # This breaks IDL macros
```
HBelusca, line 583: What kind of error(s) does it give?

sdk/cmake/gcc.cmake
HBelusca: We should definitively fix the whole binutils stuff for Windows targets anyway ^^

base/applications/network/ftp/CMakeLists.txt
zefklop: Yes, but stuff from msvcrtex depends on it :-(

sdk/cmake/msvc.cmake
zefklop: it's still used by oldnames. Will fix

win32ss/gdi/gdi32/CMakeLists.txt
zefklop: looks like it, yes

sdk/cmake/CMakeMacros.cmake
zefklop: Something like "generator expressions can't be used for custom command"

PR #3353, opened 2020-11-04
[BOOTDATA] Add new Setup Debug/Screen boot entries in BootCD/HybridCD/PC98 boot menus.

HBelusca:
## Purpose

Based on the idea proposed by @vgalnt in:
JIRA issue: [CORE-17350](https://jira.reactos.org/browse/CORE-17350)

Makes testing easier on real hardware (especially for these configs
where debugging can only be done via screen mode).

Add also an experimental Setup ACPI APIC entry in bootcd.ini for testing
purposes, along the lines of commit 5ee0925 .
In the similar LiveCD entry, use instead the /HAL= option.

### DIFF
--- a/boot/bootdata/bootcd.ini
+++ b/boot/bootdata/bootcd.ini
@@ -26,6 +26,21 @@ TimeText=Seconds until highlighted choice will be started automatically:
 
 [Operating Systems]
 Setup="Setup"
+Setup_Debug="Setup (Debug)"
+Setup_Aacpi="Setup ACPI APIC (Debug)"
+Setup_Screen="Setup (Screen)"
 
 [Setup]
 BootType=ReactOSSetup
+
+[Setup_Debug]
+BootType=ReactOSSetup
+Options=/DEBUG /DEBUGPORT=COM1 /BAUDRATE=115200 /NOGUIBOOT /SIFOPTIONSOVERRIDE
+
+[Setup_Aacpi]
+BootType=ReactOSSetup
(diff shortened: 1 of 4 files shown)


boot/bootdata/bootcd.ini
```diff
@@ -1,6 +1,6 @@
 [FREELOADER]
 DefaultOS=Setup
```
binarymaster, line 2: Shouldn't default choice be `Setup_Debug` to avoid changing the current behavior? Or it's overridden by `txtsetup.sif` by default anyway?

HBelusca: Actually it's currently already done by txtsetup.sif, because both the OsLoadOptions + SetupDebugOptions values are combined together automatically, due to the fact that do that systematically when compiling freeldr in DBG mode.
But there is also a difference: the SetupDebugOptions currently active use /KDSERIAL, which is nice for the test bots, but not when you want to interact yourself with the debugger on real hardware, iirc.

vgalnt: There is also no interactivity when ReactOS boot from bootcd in Virtual Box (5.1.30-32, XP Host).
If "Options = / NOGUIBOOT / DEBUG / DEBUGPORT = COM1 / BAUDRATE = 115200 / FIRSTCHANCE" is used then I can debug bootcd.

HBelusca: When using DEBUGPORT=COMx, the /KDSERIAL means that the debugger input should come from the host computer (that receives the data from serial for display) instead of the target computer being debugged.

HBelusca:
Cc @disean too :)

binarymaster (approved):
Didn't tested yet, but it looks correct. :+1:

disean (approved):

PR #3354, opened 2020-11-05
[WIN32K:NTUSER] Update USER_TIMER_M*IMUM handling to current target

SergeGautherie:
which is WS03 SP2.

### DIFF
--- a/win32ss/user/ntuser/sysparams.c
+++ b/win32ss/user/ntuser/sysparams.c
@@ -1490,14 +1490,19 @@ SpiGetSet(UINT uiAction, UINT uiParam, PVOID pvParam, FLONG fl)
             return SpiGetInt(pvParam, &gspv.iMouseHoverTime, fl);
 
         case SPI_SETMOUSEHOVERTIME:
-           /* See http://msdn2.microsoft.com/en-us/library/ms724947.aspx
-            * copy text from it, if some agument why xp and 2003 behovir diffent
-            * only if they do not have SP install
-            * " Windows Server 2003 and Windows XP: The operating system does not
-            *   enforce the use of USER_TIMER_MAXIMUM and USER_TIMER_MINIMUM until
-            *   Windows Server 2003 SP1 and Windows XP SP2 "
-            */
+        {
+            /* Windows XP SP2 and Windows Server 2003 SP1 new behaviour */
+            if (uiParam < USER_TIMER_MINIMUM)
+            {
+               uiParam = USER_TIMER_MINIMUM;
+            }
+            else if (uiParam > USER_TIMER_MAXIMUM)
+            {
+               uiParam = USER_TIMER_MAXIMUM;
+            }
+
             return SpiSetInt(&gspv.iMouseHoverTime, uiParam, KEY_MOUSE, VAL_HOVERTIME, fl);
+        }
 
         case SPI_GETWHEELSCROLLLINES:
             return SpiGetInt(pvParam, &gspv.iWheelScrollLines, fl);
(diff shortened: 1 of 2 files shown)


win32ss/user/ntuser/sysparams.c
```diff
@@ -1490,14 +1490,19 @@ SpiGetSet(UINT uiAction, UINT uiParam, PVOID pvParam, FLONG fl)
...
+        {
+            /* Windows XP SP2 and Windows Server 2003 SP1 new behaviour */
+            if (uiParam < USER_TIMER_MINIMUM)
+            {
+               uiParam = USER_TIMER_MINIMUM;
+            }
+            else if (uiParam > USER_TIMER_MAXIMUM)
+            {
+               uiParam = USER_TIMER_MAXIMUM;
+            }
```
HBelusca, line 1502:
```suggestion
            uiParam = min(max(uiParam, USER_TIMER_MINIMUM), USER_TIMER_MAXIMUM);
```

win32ss/user/ntuser/timer.c
```diff
@@ -185,28 +185,18 @@ IntSetTimer( PWND Window,
...
-     TRACE("Adjusting uElapse\n");
+     TRACE("Adjusting uElapse to USER_TIMER_MINIMUM\n");
      Elapse = USER_TIMER_MINIMUM; // 1024hz .9765625 ms, set to 10.0 ms (+/-)1 ms
   }
+  else if (Elapse > USER_TIMER_MAXIMUM)
+  {
+     /* Windows XP SP2 and Windows Server 2003 new behaviour */
+     TRACE("Adjusting uElapse to USER_TIMER_MAXIMUM\n");
+     Elapse = USER_TIMER_MAXIMUM;
+  }
```
HBelusca, line 199: Not sure we really care about these traces, in which case doing a
```Elapse = min(max(Elapse, USER_TIMER_MINIMUM), USER_TIMER_MAXIMUM);```
would be more "straightforward".

HBelusca:
Otherwise you can use things like:
```#if (NTDDI_VERSION >= NTDDI_WS03) ... #endif```; that would convey the very same information as the added comments, and in a compile-time manner.

win32ss/user/ntuser/sysparams.c
SergeGautherie: I am not immediately convinced: good as 1-liner, but `min` and `max` never apply together.

win32ss/user/ntuser/timer.c
SergeGautherie: TRACE: I don't know, so I am not removing them (yet).
min/max: See reply to other comment.

SergeGautherie (commented):
> `NTDDI_WS03

That does not seem wanted, as old behaviour shall never be used again.

HBelusca:
What do you mean by "min and max do not apply together" ?

SergeGautherie:
> What do you mean by "min and max do not apply together" ?

A value can't be both too low and too high.

HBelusca:
> > What do you mean by "min and max do not apply together" ?
> 
> A value can't be both too low and too high.

Then you haven't understood what the min and max macros do.

SergeGautherie:
> you haven't understood

Really? Please, explain.

ThFabba:
> > you haven't understood
> 
> Really? Please, explain.

This is tiresome.
Hermès provided a code sample, you're claiming it doesn't work. Prove it.

SergeGautherie:
> you're claiming it doesn't work

No, I do not: I simply wonder why "max and min" is better than "if_< or if_>" in this case(s).

PR #3358, opened 2020-11-08
[RTL] Fix RtlImageDirectoryEntryToData for 32 bit images on x64

tkreuzer:
We can't run 32 bit images on x64 yet, but we should at least be able to map them and retrieve their resources, etc.

### DIFF
--- a/sdk/lib/rtl/image.c
+++ b/sdk/lib/rtl/image.c
@@ -289,17 +289,38 @@ RtlImageDirectoryEntryToData(
     if (NtHeader == NULL)
         return NULL;
 
-    if (Directory >= SWAPD(NtHeader->OptionalHeader.NumberOfRvaAndSizes))
-        return NULL;
+    if (NtHeader->OptionalHeader.Magic == IMAGE_NT_OPTIONAL_HDR64_MAGIC)
+    {
+        PIMAGE_OPTIONAL_HEADER64 OptionalHeader = (PIMAGE_OPTIONAL_HEADER64)&NtHeader->OptionalHeader;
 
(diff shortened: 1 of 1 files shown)


HBelusca (approved):

binarymaster (approved):

ThFabba (approved):

sdk/lib/rtl/image.c
```diff
@@ -289,17 +289,40 @@ RtlImageDirectoryEntryToData(
     if (NtHeader == NULL)
         return NULL;
 
-    if (Directory >= SWAPD(NtHeader->OptionalHeader.NumberOfRvaAndSizes))
-        return NULL;
+#ifdef _WIN64
+    if (NtHeader->OptionalHeader.Magic == IMAGE_NT_OPTIONAL_HDR64_MAGIC)
```
ThFabba, line 293: Actually... if you call the user mode version of this function, does it recognize images that don't match your process's bitness?

learn-more: https://pastebin.com/v4awm9LE

Test succeeds for me as 32 bit test, so at least the 32 bit version can recognize a 64 bit (in-memory) image

learn-more: image
2k3

zefklop (approved):

sdk/lib/rtl/image.c
tkreuzer: @learn-more Your paste is gone. Can you repaste is somewhere or commit it?

learn-more: Added #3383

PR #3359, opened 2020-11-08
[NTOS:OB] ObReferenceObjectByHandle(): Add 2 DRIVER_VERIFIER_DETECTED_VIOLATION cases

SergeGautherie:
Commit Thomas's patch, plus a few tweaks.
Especially, make it depend on a define, until the D.V. management feature is added.

### DIFF
--- a/ntoskrnl/include/config.h
+++ b/ntoskrnl/include/config.h
@@ -1,4 +1,7 @@
 #pragma once
 
+// Enable Driver Verifier checks in ObReferenceObjectByHandle(). (CORE-10207)
+// #define ENABLE_DRIVER_VERIFIER_OBREFERENCEOBJECTBYHANDLE
+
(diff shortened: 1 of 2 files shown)


GeoB99:
I fairly believe Thomas has provided the patch as a mean to help out hunting bugs in the calls made within the kernel and not something to be merged to master. 

The proper way to fix the aforementioned bug in the ticket is to actually find out all the respective calls that don't use the `OBJ_KERNEL_HANDLE` attribute correctly.

GeoB99:
Anyways, I stand by my comment. Thomas' patch is not meant to be merged in master that way, use it at your advantage for research to catch bugs in the kernel. 

When Driver Verifier is fully and properly implemented in ReactOS, this can be re-addressed later. With that said, I'm closing this.

PR #3360, opened 2020-11-08
[NTOS:MM] MmCreateSection(), CC hack: Enforce using a kernel handle

SergeGautherie:
3rd part of this CC hack...


SergeGautherie:
It looks like #3361 will remove this whole `! NEWCC` hack...

PR #3366, opened 2020-11-11
[AFD] LockHandles(): Enforce using a kernel handle

SergeGautherie:

### DIFF
--- a/drivers/network/afd/afd/lock.c
+++ b/drivers/network/afd/afd/lock.c
@@ -311,6 +311,8 @@ PAFD_HANDLE LockHandles( PAFD_HANDLE HandleArray, UINT HandleCount ) {
     PAFD_HANDLE FileObjects;
     UINT i;
     NTSTATUS Status;
+    PVOID Object;
+    HANDLE KernelHandle;
 
     FileObjects = ExAllocatePoolZero(NonPagedPool,
                                      HandleCount * sizeof(AFD_HANDLE),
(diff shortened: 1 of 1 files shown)


drivers/network/afd/afd/lock.c
```diff
@@ -328,17 +330,55 @@ PAFD_HANDLE LockHandles( PAFD_HANDLE HandleArray, UINT HandleCount ) {
...
+            AFD_DbgPrint(MIN_TRACE,
+                         ("ObReferenceObjectByHandle(%p, UserMode) failed: 0x%08lx\n",
+                          (HANDLE)HandleArray[i].Handle, Status));
+
+            UnlockHandles(FileObjects, i);
+            return NULL;
+        }
+
+        // Create a kernel handle from the pointer.
+        Status = ObOpenObjectByPointer(Object,
```
ThFabba, line 351: Your logic here is
* reference handle U to get object O
* create handle K from object O
* reference handle K to get object O

That's entirely pointless, just save the object the first time.

SergeGautherie: > pointless

Existing code uses 'KernelMode, FILE_ALL_ACCESS'...
Now replaced with 'UserMode, 0'.

drivers/network/afd/afd/lock.c
```diff
@@ -308,38 +308,44 @@ VOID UnlockBuffers( PAFD_WSABUF Buf, UINT Count, BOOL Address ) {
 /* Produce a kernel-land handle array with handles replaced by object
  * pointers.  This will allow the system to do proper alerting */
 PAFD_HANDLE LockHandles( PAFD_HANDLE HandleArray, UINT HandleCount ) {
```
HBelusca, line 310: I have a problem with this function (and its caller in the AFD driver): There doesn't seem to be any sort of validation of the HandleCount value, with respect to the reported buffer size passed to the IOCTL handler, and thus no guard that the loop for copying the handles won't do a buffer overread.

HBelusca (changes requested):

PR #3368, opened 2020-11-12
Clang update

zefklop:
## Purpose

Enable compilation with CLang

## Proposed changes

I've had this branch for quite some time, so let's make CLang great again.

## Why does it matter

A lot of tools use CLang nowadays to perform syntax highlighting, checks, etc... Making sure that CLang is able to compile ReactOS is the way to go if we want to use those tools out of the box.
Also, CLang is the only non-MSVC compiler able to produce PDB symbols... One day maybe, we could make use of that.

## Does it work yet ?

Compiling with CLang : yes. Running ReactOS compiled with CLang : No.

## Will you push this further

No, it's at the point that I want it to be. But if someone volunteers to make a bootable ReactOS, use clang static analysis tools and what not, this is the starting point.


sdk/cmake/config.cmake
```diff
@@ -40,7 +40,12 @@ if(CMAKE_C_COMPILER_ID STREQUAL "GNU")
     set(GCC TRUE CACHE BOOL "The compiler is GCC")
     set(CLANG FALSE CACHE BOOL "The compiler is LLVM Clang")
 elseif(CMAKE_C_COMPILER_ID STREQUAL "Clang")
-    set(GCC FALSE CACHE BOOL "The compiler is GCC")
+    # We can use LLVM Clang mimicking CL or GCC. Account for this
+    if (MSVC)
+        set(GCC FALSE CACHE BOOL "The compiler is GCC")
```
Extravert-ir, line 45: I have a feeling this would complicate things too much. Is there any better solution?

dll/win32/msvcrt/stubs.c
```diff
@@ -96,6 +96,7 @@ typedef struct __crt_locale_data_public
                unsigned int _locale_lc_codepage;
 } __crt_locale_data_public;
 
+static
```
Extravert-ir, line 99: Should these be just `static`, without `__inline`?

Extravert-ir (approved):

sdk/cmake/config.cmake
zefklop: Not setting GCC would complicate things more. As clang-cl does with MSVC, clang aims to be CLI-compatible with GCC.

While building, I didn't notice much difference there.

dll/win32/msvcrt/stubs.c
zefklop: Probably. Didn't care enough to check :-)

sdk/cmake/gcc.cmake
```diff
@@ -423,30 +431,41 @@ add_compile_options("$<$<COMPILE_LANGUAGE:CXX>:$<IF:$<BOOL:$<TARGET_PROPERTY:WIT
 # We disable exceptions, unless said so
 add_compile_options("$<$<COMPILE_LANGUAGE:CXX>:$<IF:$<BOOL:$<TARGET_PROPERTY:WITH_CXX_EXCEPTIONS>>,-fexceptions,-fno-exceptions>>")
 
+# G++ shipped with ROSBE uses sjlj exceptions. Tell Clang it is so
```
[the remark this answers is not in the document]
zefklop: this is for C++ exceptions.
And yes, CLang supports SEH, but it doesn't work right now...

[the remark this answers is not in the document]
Extravert-ir: > this is for C++ exceptions

I know, I put the `--enable-sjlj-exceptions` setting into the toolchain myself :)
MSVC uses SEH for C++ exceptions, so should Clang, shouldn't it?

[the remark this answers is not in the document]
zefklop: I didn't push the investigation very far. Without this, C++ code has references to undefined symbols. I think this would be related to SJLJ vs DWARF exceptions unwinding, not SEH.

sdk/include/crt/xmmintrin.h
```diff
@@ -47,57 +47,70 @@ __m128 _mm_xor_ps(__m128 a, __m128 b);
 __m128 _mm_div_ps(__m128 a, __m128 b);
 
 #ifdef _MSC_VER
-unsigned int _mm_getcsr(void);
+ unsigned int _mm_getcsr(void);
```
HBelusca, line 50: Please revert.

sdk/lib/crt/math/i386/exp2_asm.s
```diff
@@ -0,0 +1,38 @@
+#include <asm.inc>
```
HBelusca, line 1: Missing copyright notice?

drivers/network/ndis/ndis/buffer.c
```diff
@@ -10,7 +10,7 @@
 
 #include <ndissys.h>
 
-__inline ULONG SkipToOffset(
+FORCEINLINE ULONG SkipToOffset(
```
HBelusca, line 13:
```suggestion
FORCEINLINE
ULONG
SkipToOffset(
```

ntoskrnl/CMakeLists.txt
```diff
@@ -12,6 +12,11 @@ set(NTKRNLMP_ASM_SOURCE ${ASM_SOURCE})
 spec2def(ntoskrnl.exe ntoskrnl.spec ADD_IMPORTLIB)
 add_asm_files(ntoskrnl_asm ${NTOSKRNL_ASM_SOURCE})
 
+if (GCC AND CLANG)
+    #Clang optimises strcmp calls to memcmp.
```
HBelusca, line 16:
```suggestion
    # Clang optimises strcmp calls to memcmp.
```

HBelusca: I'm not sure this sort of optimization is correct, to begin with.

modules/rosapps/applications/net/ncftp/ncftp/main.c
```diff
@@ -24,7 +24,7 @@
 	WSADATA wsaData;
 	int wsaInit = 0;
 
-	__inline void DisposeWinsock(int aUNUSED) { if (wsaInit > 0) WSACleanup(); wsaInit--; }
+	static __inline void DisposeWinsock(int aUNUSED) { if (wsaInit > 0) WSACleanup(); wsaInit--; }
```
HBelusca, line 27: Is the underscored inline here correct? (since you removed these in a previous commit)

zefklop: I'm not even sure this is our code...

sdk/lib/crt/math/i386/exp2_asm.s
zefklop: This is a copy of exp code, which is also left without copyright.

ntoskrnl/CMakeLists.txt
zefklop: It doesn't always do it. Mainly when one of the two comparands is known at compile time.

sdk/lib/crt/stdlib/rot.c
```diff
@@ -19,8 +19,16 @@
...
 
-unsigned int _rotr( unsigned int value, int shift );
-unsigned long _lrotr(unsigned long value, int shift);
+#if defined (__clang__) && !defined(_MSC_VER)
+#define ASM_ALIAS __asm__
+#else
+#define ASM_ALIAS(x)
+#endif
+
+unsigned int _rotr( unsigned int value, int shift ) ASM_ALIAS("__rotr");
```
ThFabba, line 28: I'm not clear on why this is needed. Isn't `__rotr` the default cdecl name for a C function called `_rotr`?

sdk/lib/atl/atlpath.h
```diff
@@ -23,10 +23,10 @@ class CPathT
     // const
     inline BOOL PathFileExistsX(LPCSTR pszPath) { return PathFileExistsA(pszPath); }
     inline BOOL PathFileExistsX(LPCWSTR pszPath) { return PathFileExistsW(pszPath); }
-    inline int PathFindExtensionX(LPCSTR pszPath) { return PathFindExtensionA(pszPath); }
-    inline int PathFindExtensionX(LPCWSTR pszPath) { return PathFindExtensionW(pszPath); }
-    inline int PathFindFileNameX(LPCSTR pszPath) { return PathFindFileNameA(pszPath); }
-    inline int PathFindFileNameX(LPCWSTR pszPath) { return PathFindFileNameW(pszPath); }
+    inline PVOID PathFindExtensionX(LPCSTR pszPath) { return PathFindExtensionA(pszPath); }
```
ThFabba, line 26: I'm surprised the old code worked with any compiler... but I'm also surprised that this works. The function returns a pointer-to-const, so converting it to PVOID would need a cast.
Any reason we're not just using the correct return type here?

base/applications/rapps/include/rosui.h
```diff
@@ -96,9 +96,13 @@ class CPointerArray
 
     BOOL RemoveAt(INT i)
     {
-        T* ptr = (T*) DPA_GetPtr(m_hDpa, i);
-        OnRemoveItem(ptr);
-        return DPA_DeletePtr(m_hDpa, i);
+        T* ptr = DPA_DeletePtr(m_hDpa, i);
```
ThFabba, line 99: `DPA_DeletePtr` returns PVOID so this needs `static_cast<T*>`. How does this compile?

ntoskrnl/CMakeLists.txt
```diff
@@ -43,7 +43,7 @@ if(MSVC)
     add_target_link_flags(ntoskrnl "/SECTION:.rsrc,!DP") #Accessed from bugcheck code
     add_target_link_flags(ntoskrnl "/SECTION:INIT,ERWD")
 else()
-    if(GDB)
+    if(GDB AND NOT CLANG)
```
ThFabba, line 46: Why? What's the consequence of not optimizing? Do we get size issues like in freeldr or something?

sdk/include/crt/mingw32/intrin.h
```diff
@@ -30,12 +30,14 @@
...
 #ifndef RC_INVOKED
 
+#ifndef __INTRIN_INLINE
 #ifdef __clang__
-#define __INTRIN_INLINE __forceinline
 #define HAS_BUILTIN(x) __has_builtin(x)
+#define __ATTRIBUTE_ARTIFICIAL
 #else
-#define HAS_BUILTIN(x) 0
+#define HAS_BUILTIN(x)
```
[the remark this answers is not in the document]
zefklop: It's unintended. Will fix.

sdk/include/crt/mingw32/intrin.h
```diff
@@ -30,12 +30,14 @@
 
 #ifndef RC_INVOKED
 
+#ifndef __INTRIN_INLINE
```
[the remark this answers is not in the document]
zefklop: leftover from when this was defined here, I think. Will fix with HAS_BUILTIN instead.

sdk/include/crt/mingw32/intrin.h
```diff
@@ -30,12 +30,14 @@
...
 #ifdef __clang__
-#define __INTRIN_INLINE __forceinline
 #define HAS_BUILTIN(x) __has_builtin(x)
+#define __ATTRIBUTE_ARTIFICIAL
 #else
-#define HAS_BUILTIN(x) 0
+#define HAS_BUILTIN(x)
 #define __ATTRIBUTE_ARTIFICIAL __attribute__((artificial))
+#endif
```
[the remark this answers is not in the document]
zefklop: Sure. Indentation will help also.

sdk/include/crt/mingw32/intrin_x86.h
```diff
@@ -1696,12 +1730,14 @@ __INTRIN_INLINE void __writecr4(unsigned int Data)
 	__asm__("mov %[Data], %%cr4" : : [Data] "r" (Data) : "memory");
 }
 
+#ifndef __clang__
```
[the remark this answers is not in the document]
zefklop: No CLang claims that cr8 is not available on this platform, IIRC.

sdk/lib/crt/stdlib/rot.c
zefklop: those functions are builtins in CLang, and it rejects redefinig them. Look up in the file, _rotr is defined as __function_rotr.
Defining there symbol name with an ASM alias is fine, though.

sdk/lib/atl/atlpath.h
zefklop: I guess nobody cared until now because that code is not instantaited/optimized away ?

base/applications/rapps/include/rosui.h
zefklop: it's still better than what it was before ;-)

A quick check on godbolt indicates that one must pass -fpermissive flag to G++ to have this compile.

ntoskrnl/CMakeLists.txt
zefklop: PSEH3 uses too much registers in this case. I don't remember the exact error message, but one of the asm statement clobbers half of the x86 register set and CLang complains.

sdk/include/crt/mingw32/intrin_x86.h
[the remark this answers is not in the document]
ThFabba: Oh. Well it would be correct, wouldn't it? It's an x64-only register...

sdk/lib/crt/stdlib/rot.c
ThFabba: Oh. That makes sense, thanks.
Alternatively, perhaps `-fno-builtin=_rotr` could make this more obvious?

Also, is this okay for x64, or do we need to remove the extra underscore there?

ntoskrnl/CMakeLists.txt
ThFabba: Interesting. Mind adding a small comment?

drivers/filesystems/cdfs/CMakeLists.txt
```diff
@@ -38,3 +38,9 @@ target_link_libraries(cdfs ${PSEH_LIB} memcmp)
 add_importlibs(cdfs ntoskrnl hal)
 add_cd_file(TARGET cdfs DESTINATION reactos/system32/drivers NO_CAB FOR all)
 add_registry_inf(cdfs_reg.inf)
+
+if (CLANG AND GCC)
+    # Make __inline work as expected
+    target_compile_options(cdfs PRIVATE -std=gnu89)
```
ThFabba, line 44: I wonder if we should change this default globally or something...

sdk/lib/atl/atlpath.h
ThFabba: It should be perfectly legal here. You can't overload _only_ the return type. But as long as parameter types differ between overloads, you can change the return type, too.

ThFabba (approved):

sdk/include/crt/mingw32/intrin.h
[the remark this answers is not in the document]
zefklop: after reading this again, this block is where __INTRIN_INLINE is defined, so it makes sense.

I will add a bit of indentation to make this clearer

sdk/include/crt/mingw32/intrin_x86.h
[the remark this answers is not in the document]
zefklop: I guess yes. I will just drop that, it's defined in the x86_64 case up in this file anyway

sdk/lib/crt/stdlib/rot.c
zefklop: I don't know, but I guess this will have to be fixed.

dll/win32/msvcrt/stubs.c
zefklop: Done

drivers/filesystems/cdfs/CMakeLists.txt
zefklop: No idea, really...

PR #3371, opened 2020-11-14
[REGEDIT] Display finished find messagebox after finishing the search via "Find Next" (F3) CORE-17368

oleg-dubinskiy:
## Purpose

Display the notification messagebox if no more items were found when search via "Edit" -> "Find Next" (or just F3 hotkey).
Also move the messagebox into separate function and call it in all appropriate places: in `FindDialog`, where it was from the start, and in the `ID_EDIT_FINDNEXT`, which is responsible for "Find Next" action.
So now it is more understandable whether the search was finished.
It repeats behaviour of Windows Regedit.

JIRA issue: [CORE-17368](https://jira.reactos.org/browse/CORE-17368)

### DIFF
--- a/base/applications/regedit/framewnd.c
+++ b/base/applications/regedit/framewnd.c
@@ -1230,7 +1230,7 @@ static BOOL _CmdWndProc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam)
         FindDialog(hWnd);
         break;
     case ID_EDIT_FINDNEXT:
-        FindNext(hWnd);
+        FindNextMessageBox(hWnd);
         break;
     case ID_EDIT_COPYKEYNAME:
         CopyKeyName(hWnd, hKeyRoot, keyPath);
--- a/base/applications/regedit/find.c
+++ b/base/applications/regedit/find.c
@@ -810,18 +810,23 @@ static INT_PTR CALLBACK FindDialogProc(HWND hDlg, UINT uMsg, WPARAM wParam, LPAR
     return iResult;
 }
 
+void FindNextMessageBox(HWND hWnd)
+{
+    if (!FindNext(hWnd))
+    {
+        WCHAR msg[128], caption[128];
+
+        LoadStringW(hInst, IDS_FINISHEDFIND, msg, COUNT_OF(msg));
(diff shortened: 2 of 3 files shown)


base/applications/regedit/framewnd.c
```diff
@@ -1230,7 +1230,14 @@ static BOOL _CmdWndProc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam)
         FindDialog(hWnd);
         break;
     case ID_EDIT_FINDNEXT:
-        FindNext(hWnd);
+        if (!FindNext(hWnd))
```
binarymaster, line 1233: This basically duplicates the existing code. I think you should refactor it into a separate function.

https://github.com/reactos/reactos/blob/6aae5f4fdd87efe805f7bd44c601044b0cbe47c2/base/applications/regedit/find.c#L818

oleg-dubinskiy: Done. Thanks @binarymaster. :+1: 
Still works for me without any issues.

HBelusca: But why is then the function used only once?

oleg-dubinskiy: Why I need call it two or more times @HBelusca? What this will improve? It works for me as-is, the behaviour is the same as in 2k3 Regedit. What's wrong? :confused:

binarymaster (approved):
Looks good to me, except I'd call this new function `FindNextMessageBox`.

oleg-dubinskiy:
>I'd call this new function `FindNextMessageBox`.

Fixed. :ok_hand:

base/applications/regedit/framewnd.c
HBelusca: I understood @binarymaster 's comment as the fact that this code chunk was duplicated at least two times, hence needing to be placed instead in a function in order to remove the code duplication. But if that code chunk was actually present just once, I don't see why moving it inside a function would help.

binarymaster: I see what you mean, it's possible to move message box creation into `FindNext()`, but I think in current condition of this PR the code is more readable.

PR #3373, opened 2020-11-15
[NTOS:KD] KdpDebugLogInit(): Close the thread handle

SergeGautherie:
And assert being in kernel mode
to be explicit that using Nt*() is safe.

### DIFF
--- a/ntoskrnl/kd/kdio.c
+++ b/ntoskrnl/kd/kdio.c
@@ -110,6 +110,8 @@ KdpLoggerThread(PVOID Context)
     ULONG beg, end, num;
     IO_STATUS_BLOCK Iosb;
 
+    ASSERT(ExGetPreviousMode() == KernelMode);
+
     KdpLoggingEnabled = TRUE;
 
     while (TRUE)
--- a/ntoskrnl/kd/kdio.c
+++ b/ntoskrnl/kd/kdio.c
@@ -296,6 +296,8 @@ KdpDebugLogInit(PKD_DISPATCH_TABLE DispatchTable,
                                ThreadPriority,
                                &Priority,
                                sizeof(Priority));
+
+        ZwClose(ThreadHandle);
     }
 }
 


ntoskrnl/kd/kdio.c
```diff
@@ -184,15 +184,15 @@ KdpLoggerThread(PVOID Context)
 
         if (end > beg)
         {
-            NtWriteFile(KdpLogFileHandle, NULL, NULL, NULL, &Iosb,
+            ZwWriteFile(KdpLogFileHandle, NULL, NULL, NULL, &Iosb,
```
ThFabba, line 187: I'm not strongly against this but also not super in favor of it.
An alternative that doesn't kill the optimization would be to `ASSERT(ExGetPreviousMode() == KernelMode)` to show that using Nt* is safe in these functions.

HBelusca: The logger thread is a system thread so it should normally run in kernel mode context only. (So the Nt* functions are safe to call.)

SergeGautherie: > I'm not strongly against this but also not super in favor of it.

Yeah, I wondered too.

> `ASSERT(ExGetPreviousMode() == KernelMode)`

Replaced :-)

SergeGautherie: > The logger thread is a system thread

Yes, this commit is only about being explicit/extra-safe.

HBelusca: Well then you don't need that. It's not an "extra-safe" feature, but it's an under-optimization (now going through the system-call table etc.)

ntoskrnl/kd/kdio.c
```diff
@@ -256,6 +258,8 @@ KdpDebugLogInit(PKD_DISPATCH_TABLE DispatchTable,
     KPRIORITY Priority;
     SIZE_T MemSizeMBs;
 
+    ASSERT(ExGetPreviousMode() == KernelMode);
```
HBelusca, line 261: Useless since anyone understanding the code would know that `KdpDebugLogInit` is called by `KdInitSystem`, which is only called from the kernel and by the kernel, so it's definitively running kernel mode context only.

ntoskrnl/kd/kdio.c
```diff
@@ -160,6 +160,8 @@ KdpLoggerThread(PVOID Context)
     ULONG beg, end, num;
     IO_STATUS_BLOCK Iosb;
 
+    ASSERT(ExGetPreviousMode() == KernelMode);
```
HBelusca, line 163: Same reason as below by transitivity.

HBelusca (changes requested):

ntoskrnl/kd/kdio.c
SergeGautherie: > Well then you don't need that. It's not an "extra-safe" feature, but it's an under-optimization (now going through the system-call table etc.)

I am not sure what you want to debate further: I think my answer to Thomas and my commit update are enough here.

ntoskrnl/kd/kdio.c
SergeGautherie: > it's definitively running kernel mode context only.

Obviously. This is not about how current code behaves.
Your point of view is clear. If Thomas doesn't want the asserts either, I'll simply remove them.

> anyone understanding the code

Hum... Could you tell me why, for example,
https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=KdbpCliInit
uses `Zw*()`?

SergeGautherie: Eventually, 7caf9e9 does what I initially submitted...

SergeGautherie:
Cc @zefklop, who is working on this code.

PR #3374, opened 2020-11-15
[WIN32K:NTUSER] IntCreateDesktop(): Update a 'KernelMode' remnant

SergeGautherie:
Addendum to 765f094 (r57632) then 1abeb90.

--

NB:
Currently, `AccessMode == UserMode`.
@HBelusca, Is it worth having this value as a parameter? // Yes, for future use.

### DIFF
--- a/win32ss/user/ntuser/desktop.c
+++ b/win32ss/user/ntuser/desktop.c
@@ -2368,7 +2368,7 @@ IntCreateDesktop(
     Status = ObReferenceObjectByHandle(hDesk,
                                        0,
                                        ExDesktopObjectType,
-                                       KernelMode,
+                                       AccessMode,
                                        (PVOID*)&pdesk,
                                        NULL);
     if (!NT_SUCCESS(Status))


win32ss/user/ntuser/desktop.c
```diff
@@ -2368,7 +2368,7 @@ IntCreateDesktop(
     Status = ObReferenceObjectByHandle(hDesk,
                                        0,
                                        ExDesktopObjectType,
-                                       KernelMode,
+                                       AccessMode,
```
HBelusca, line 2371: Why? The hDesk handle was obtained from kernel mode, via the previous call; it didn't come from e.g. user mode.

HBelusca:
> Currently, AccessMode == UserMode

Yes I also want the IntCreateDesktop function to be available for being called from previous mode == KernelMode (i.e. by internal calls in win32k).

win32ss/user/ntuser/desktop.c
SergeGautherie: > always use `UserMode` here

That would currently work, but not match the rest of current code either, before and after this call.

> try this

No thanks: my goal is to fix, at least improve, current code, not to guess what future calls with `KernelMode` you are planning would look like.
That should be checked, as the rest of the code, when there is an actual case.

ThFabba: You are making a change and Hermès says it may not be correct, and suggests a fix. It seems to be a response of "No thanks" means "please close this PR, I'm not interested in making it correct".

ThFabba (changes requested):

win32ss/user/ntuser/desktop.c
SergeGautherie: > not interested

Not at all: I merely replied that 'try this' about some future call is too vague (for me) as 'an actual [test]case'...

SergeGautherie: @HBelusca, could you either accept this PR as is, or give an explicit testcase? Thanks.

PR #3375, opened 2020-11-16
[HALX86] Update the PCI hardware IDs database, to v2020.11.14

SergeGautherie:
Import
https://github.com/pciutils/pciids/blob/master/pci.ids


HBelusca:
Committed in master.

PR #3377, opened 2020-11-19
[UNIATA] Workaround a regression

SergeGautherie:
Fix random crashes on Virtual PC (2004, at least).

- Workaround regressions.

JIRA issue: [CORE-17371](https://jira.reactos.org/browse/CORE-17371)

### DIFF
--- a/drivers/storage/ide/uniata/CMakeLists.txt
+++ b/drivers/storage/ide/uniata/CMakeLists.txt
@@ -17,7 +17,7 @@ list(APPEND SOURCE
     id_queue.cpp
     id_sata.cpp
     ros_glue/ros_glue.cpp
-    stdafx.h)
+    stdafx.h) # Not stdafx.cpp (which only includes the .h), to fix PCH.
 
 add_library(uniata MODULE ${SOURCE} idedma.rc)
 
--- a/drivers/storage/ide/uniata/CMakeLists.txt
+++ b/drivers/storage/ide/uniata/CMakeLists.txt
@@ -17,7 +17,7 @@ list(APPEND SOURCE
     id_queue.cpp
     id_sata.cpp
     ros_glue/ros_glue.cpp
-    stdafx.h) # Not stdafx.cpp (which only includes the .h), to fix PCH.
+    stdafx.h)
 
 add_library(uniata MODULE ${SOURCE} idedma.rc)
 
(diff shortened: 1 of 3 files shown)


SergeGautherie:
@Alter-1, could you double-check and explain? Thanks.

drivers/storage/ide/uniata/CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ list(APPEND SOURCE
     id_queue.cpp
     id_sata.cpp
     ros_glue/ros_glue.cpp
-    stdafx.h)
+    stdafx.h) # Not stdafx.cpp (which only includes the .h), to fix PCH.
```
learn-more, line 20: Why comment that you should not add 'stdafx.cpp' here?
Why not also comment that you should not include 'RandomOtherFile.cpp' there?

SergeGautherie: > Why comment that you should not add 'stdafx.cpp' here?

To prevent from the mistake being actually made again.

> Why not also comment that you should not include 'RandomOtherFile.cpp' there?

??

learn-more: > ??

Trying to convey a message with sarcasm doesnt seem to be processed by you, so I'll rephrase it:

There is no need to add random 'Dont do xyz here' when there is no suggestion anywhere to actually do 'xyz'

mifritscher: Actually, I can understand @SergeGautherie here, because the other files are *.cpp files as well - which can lead to wrong "fixes".

SergeGautherie: > There is no need to add random 'Dont do xyz here' when there is no suggestion anywhere to actually do 'xyz'

Agreed.
Reverted, though that is not the case here...

minecraftchest1: I don't see a problem.

binarymaster (changes requested):
Upstream diff fix has been addressed in 88d36c93. I cannot update your branch, please allow edits from maintainers.
```
remote: Resolving deltas: 100% (12/12), completed with 7 local objects.
To https://github.com/SergeGautherie/reactos
 ! [remote rejected]       pr_branch -> SergeGautherie/CORE-17371_uniata_revert-cur_cdev (permission denied)
error: failed to push some refs to 'https://github.com/SergeGautherie/reactos'
```

binarymaster (changes requested):
Please remove these commits by a force push with lease:

> - [UNIATA] CMakeLists.txt: Add a comment about PCH
> - Revert sarcasm?

So your PR should contain only these two commits:

> - [UNIATA] BusMasterAdapters[]: 8086:7111 is really PIIX4, not PIIX3
> - [UNIATA] UniataQueueRequest(): Work around random crashes on Virtual PC 2004

PR #3378, opened 2020-11-19
[CMAKE] Overhaul CDs and reactos.cab creation and dependency tracking - try 2

zefklop:
- Use target properties to track dependencies and file names & destination for the relevant targets
 - Use generator expressions to generate the relevant files

## Purpose

Make the whole process cleaner, instead of calling some macro at the end of the process.


sdk/cmake/CMakeMacros.cmake
```diff
@@ -250,6 +250,17 @@ function(add_cd_file)
         endif()
     endif()
 
+    # If we use a byproduct of a target, then we can use this as a dependency specified in DEPENDS argument of add_custom_command
+    # whereas if we use a custom target, we must use target-level dependency
+    # see CMake documentation of add_custom_command
```
HBelusca, line 255:
```suggestion
    # If we use a byproduct of a target, then we can use this as a dependency specified in DEPENDS argument
    # of add_custom_command, whereas if we use a custom target, we must use target-level dependency.
    # See CMake documentation of add_custom_command,
```

boot/CMakeLists.txt
```diff
@@ -15,10 +15,12 @@ else()
     message(FATAL_ERROR "Unknown ARCH '" ${ARCH} "', cannot generate a valid UEFI boot filename.")
 endif()
 
-add_custom_target(efisys
+add_custom_command(
+    OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/efisys.bin
     COMMAND native-fatten ${CMAKE_CURRENT_BINARY_DIR}/efisys.bin -format 2880 EFIBOOT -boot ${CMAKE_CURRENT_BINARY_DIR}/freeldr/bootsect/fat.bin -mkdir EFI -mkdir EFI/BOOT -add $<TARGET_FILE:bootmgfw> EFI/BOOT/boot${EFI_PLATFORM_ID}.efi
     DEPENDS native-fatten fat bootmgfw
     VERBATIM)
+add_custom_target(efisys DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/efisys.bin)
```
HBelusca, line 23: Why isn't CMake clever enough to understand that the former (existing) code should have the same behaviour as the new one?

boot/CMakeLists.txt
```diff
@@ -59,106 +61,112 @@ set(ISO_VOLNAME      "ReactOS")             # For both the Volume ID and the Vol
...
+add_custom_command(
+    OUTPUT ${REACTOS_BINARY_DIR}/bootcd.iso
     COMMAND native-mkisofs -quiet -o ${REACTOS_BINARY_DIR}/bootcd.iso -iso-level 4
         -publisher ${ISO_MANUFACTURER} -preparer ${ISO_MANUFACTURER} -volid ${ISO_VOLNAME} -volset ${ISO_VOLNAME}
         -eltorito-boot loader/isoboot.bin -no-emul-boot -boot-load-size 4 -eltorito-alt-boot -eltorito-platform efi -eltorito-boot loader/efisys.bin -no-emul-boot -hide boot.catalog
         -sort ${CMAKE_CURRENT_BINARY_DIR}/bootfiles.sort
         -no-cache-inodes -graft-points -path-list ${CMAKE_CURRENT_BINARY_DIR}/bootcd.$<CONFIG>.lst
-    COMMAND native-isohybrid -b ${_isombr_file} -t 0x96 ${REACTOS_BINARY_DIR}/bootcd.iso
-    DEPENDS isombr native-isohybrid native-mkisofs
+    DEPENDS isombr native-isohybrid native-mkisofs $<GENEX_EVAL:$<TARGET_PROPERTY:bootcd,ISO_DEPENDENCIES>>
```
HBelusca, line 110: It seems that the correct native-isohybrid command call just disappeared ... :(
 (while it remains for the livecd and the others, as expected)

boot/CMakeLists.txt
zefklop: It doesn't. add_custom_target(COMMAND) creates an always out-of-date target, since you don't provide info about what the command generates, whereas add_custom_target(DEPENDS) properly creates a target which is as up-to-date as the command output.

Same goes for the cdrom targets.

boot/CMakeLists.txt
zefklop: Right. Copy-paste fail here.

Extravert-ir (approved):
Tested, works for me

PR #3380, opened 2020-11-22
[REGEDIT] Prevent launching multiple instances of application CORE-17379

oleg-dubinskiy:
## Purpose

Prevent the possibility to launch multiple instances of our Regedit at the same time.
This repeats MS behaviour. XP/2003 Regedit allows to launch only one instance of it.
Also make the window of current instance active and show it when trying to launch the app second time.

JIRA issue: [CORE-17379](https://jira.reactos.org/browse/CORE-17379)

## Result

Tested our Regedit on Win2k3, before and after my changes.

Before:
regedit_before_2k3

After:
regedit_after_2k3

MS Regedit (for comparison):
MS_regedit_2k3

### DIFF
--- a/base/applications/regedit/main.c
+++ b/base/applications/regedit/main.c
@@ -194,6 +194,7 @@ int APIENTRY wWinMain(HINSTANCE hInstance,
                       int       nCmdShow)
 {
     MSG msg;
+    HANDLE hMutex;
     HACCEL hAccel;
 
     UNREFERENCED_PARAMETER(hPrevInstance);
@@ -217,6 +218,29 @@ int APIENTRY wWinMain(HINSTANCE hInstance,
         default:
             break;
     }
+
+    /* Prevent multiple instances of application */
+    hMutex = CreateMutexW(NULL, FALSE, L"Regedit");
+
+    if (hMutex)
+    {
+        if (GetLastError() == ERROR_ALREADY_EXISTS)
+        {
+            /* The application's instance is already running */
+            HWND hCurrentWnd = FindWindowW(szFrameClass, szTitle);
+
+            if (hCurrentWnd)
+            {
+                /* Activate and show the window of current instance */
+                SetForegroundWindow(hCurrentWnd);
+                ShowWindow(hCurrentWnd, nCmdShow);
+            }
+
+            CloseHandle(hMutex);
+            return 0;
+        }
+    }
+
     /* Store instance handle in our global variable */
     hInst = hInstance;
 
@@ -239,6 +263,13 @@ int APIENTRY wWinMain(HINSTANCE hInstance,
     }
 
     ExitInstance(hInstance);
+
+    if (hMutex)
+    {
+        ReleaseMutex(hMutex);
+        CloseHandle(hMutex);
+    }
+
     return (int)msg.wParam;
 }
 


base/applications/regedit/main.c
```diff
@@ -217,6 +218,29 @@ int APIENTRY wWinMain(HINSTANCE hInstance,
         default:
             break;
     }
+
+    /* Prevent multiple instances of application */
+    hMutex = CreateMutexW(NULL, FALSE, L"Regedit");
```
HBelusca, line 223: Is it compatible with what MS does? (Namely, what happens on Windows when you try to open both MS regedit and ROS regedit at the same time?)

base/applications/regedit/main.c
```diff
@@ -217,6 +218,29 @@ int APIENTRY wWinMain(HINSTANCE hInstance,
...
+
+    if (hMutex)
+    {
+        if (GetLastError() == ERROR_ALREADY_EXISTS)
+        {
+            /* The application's instance is already running */
+            HWND hCurrentWnd = FindWindowW(szFrameClass, szTitle);
```
HBelusca, line 230: What if I start a russian-translated ROS regedit on an english Windows installation that is also running an english-translated regedit?

HBelusca (commented):
Finally as @learn-more asked in the JIRA comment, is this "limitation" actually desirable from a user perspective?

base/applications/regedit/main.c
oleg-dubinskiy: >Is it compatible with what MS does?

Yes @HBelusca. Acording to MSDN article for `WinMain` function: https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-winmain#parameters (see `hPrevInstance` parameter reference), this should be done using a Mutex:
> If you need to detect whether another instance already exists, create a uniquely named mutex using the CreateMutex function.

And then I did as described here (using Mutex method): https://stackoverflow.com/a/8799795.
So it definitely should be compatible.

>Namely, what happens on Windows when you try to open both MS regedit and ROS regedit at the same time?

Nothing suspicious, as I see. They are functioning independent from each other. Is this wrong? :thinking:

HBelusca: I was not asking about how one could potentially achieve this. I was asking how MS regedit **actually** does this, in order to be able to "interact" with it (i.e. detect that if MS regedit is already opened, and you open ROS regedit at the same time, then the latter does not show up but instead re-focuses the MS regedit window).

oleg-dubinskiy: Ah, I understood what you mean.
As I see, MS Regedit most of all asks for its other instances by enumerating all Regedit processes. I tried to launch two identical copies of MS Regedit from the different paths, and... launching the second copy indeed redirects me to the current running instace of Regedit. So looks like the Mutex method doesn't work with MS Regedit as expected (if it does not redirect to MS Regedit after launching our one).
Also in case when I launch our Regedit first, and then MS one, it same doesn't redirect me to the instance of our one. Is it a problem of our Regedit too, or it depends only on MS Regedit in this case? :thinking:

SergeGautherie:
For what it's worth, you may want to check #2946 for similar TaskMgr code.

base/applications/regedit/main.c
```diff
@@ -239,6 +263,13 @@ int APIENTRY wWinMain(HINSTANCE hInstance,
     }
 
     ExitInstance(hInstance);
+
+    if (hMutex)
+    {
+        ReleaseMutex(hMutex);
```
ThFabba, line 269: You never acquire the mutex, so you shouldn't release it.

Valeriy-Zainullin:
> Finally as @learn-more asked in the JIRA comment, is this "limitation" actually desirable from a user perspective?

If we allow multiple instances, then we will need to add refreshing to the registry editor. Otherwise one of windows may display outdated registry information. Maybe automatic, if we have manual refreshing. Should we make automatic refreshing instead? Then user will be able to edit many registry items at the same time using registry editor.

HBelusca:
A good way for doing the refreshing automatically would be for example, for regedit to register a callback for monitoring registry key changes (which is not implemented in ROS yet, but exists on Windows).
Or alternatively, to do a "manual" automated refresh on the transition from regedit's unfocused window to focused window.

learn-more:
Regedit also does not refresh when another application changes a random key, which seems fine.

PR #3381, opened 2020-11-25
[NTOS:WMI] Don't free the buffer on success in IoWMIWriteEvent CORE-17384

oleg-dubinskiy:
## Purpose

Don't call `ExFreePool` when `IoWMIWriteEvent` returns success.
It will fix several asserts when trying to boot ReactOS with acpi.sys from XP/2003.
Although the system still doesn't boot properly, but nevertheless this improves the situation a bit.
Based on @ThFabba's comment: https://jira.reactos.org/browse/CORE-17384?focusedCommentId=126435&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-126435

JIRA issue: [CORE-17384](https://jira.reactos.org/browse/CORE-17384)

### DIFF
--- a/ntoskrnl/wmi/wmi.c
+++ b/ntoskrnl/wmi/wmi.c
@@ -111,10 +111,6 @@ IoWMIWriteEvent(IN PVOID WnodeEventItem)
     DPRINT1("IoWMIWriteEvent() called for WnodeEventItem %p, returning success\n",
         WnodeEventItem);
 
-    /* Free the buffer if we are returning success */
-    if (WnodeEventItem != NULL)
-        ExFreePool(WnodeEventItem);
-
     return STATUS_SUCCESS;
 }
 


HBelusca:
Why shouldn't we free our own allocated buffer? Does it get freed later on by ourselves too? or?
(basically can you see who else is doing the other freeing?)

ThFabba:
It's not "our own allocated buffer", it's a buffer owned and passed in by the caller.
The documentation for this function says that the buffer is freed, however reality is more complicated than than.
Most importantly, ACPI passes in a stack buffer, which is why we get asserts and crashes if we try to free it.
Until we have a better understanding of how this function is supposed to behave and/or some tests, "sometimes leaking" seems better than "sometimes crashing".

SergeGautherie:
Should we `#if` out this code and/or add a reminder comment, instead of a removal?

SergeGautherie:
To be closed: superseded.

PR #3382, opened 2020-11-25
[CLB] Enable the dll CORE-17385

oleg-dubinskiy:
## Purpose

Enable Column List Box dll (clb.dll) in CMake rules.
Required by MS Regedit, together with ulib.dll. It works as well with our clb.dll.

JIRA issue: [CORE-17385](https://jira.reactos.org/browse/CORE-17385)

## Proposed changes

- Enable the clb subdirectory in the regedit's CMakeLists.txt.
- Add reactos/system32 destination for clb.dll in clb's CMakeLists.txt, so during building the iso, it will be added into reactos.cab and copied into system32 during installation or LiveCD booting.

## TODO

- [ ] Move the dll into dll/win32 if needed, in order to not mix system dlls with executable applications (probably in another PR).

## Result

Current state (without copying any additional dlls):
MS_Regedit_current

Before (after copying MS ulib.dll):
MS_Regedit_before

After:
MS_Regedit_after

This dll also exists on XP/2003:
clb_win2k3

DLL purpose: Column List Box.


JoachimHenze:
My personal gutfeeling: I would postpone committing that until it covers both: clb.dll and ulib.dll.
Otherwise it's no real win imho. Or unless we identified any arbitrary app which would be happy with just having clb.dll.
I do not insist on waiting though. It's just my gutfeeling.

learn-more (approved):
The code is there, we might as well just include it.

PR #3385, opened 2020-12-09
Sync wininet to wine 6.0rc1

zefklop:
## Purpose

When running rosautotest, I could see that there was a loop of wininet trying to delete URL cache entries or something. Maybe it is what stops the test bot to succeed, maybe not, but a wine sync wouldn't hurt anyway.


Extravert-ir:
I think we should fix testbots first

zefklop:
> I think we should fix testbots first

Sure.

PR #3390, opened 2020-12-15
[KS] Fix length of the media categories path in KspReadMediaCategory CORE-17361

oleg-dubinskiy:
## Purpose

Fix length of the media categories path in `KspReadMediaCategory`.
It will allow our Kernel Streaming driver to properly load MS audio stack (portcls.sys, sysaudio.sys and wdmaud.sys from Windows XP/2003). :smiley: 

## Analysis

Currently, it's just set to zero.
Actually, it contains the path to the media categories registry key plus key name for each media category.
So it should have the appropriate length.

## Result

Before:
MS_audio_stack_before

After:
MS_audio_stack_after

Although the sound does no logner work with such replacement (at least with AC97 VirtualBox driver), but nevertheless it's booting.
According to the log, MS audio drivers need `KSEVENTSETID_Sysaudio` event set from our Kernel Streaming, which should be available from `KsGernerateEventList` function, which, in its turn, uses `KsGenerateEvents`. These functions are currently unimplemented on our side.

### DIFF
--- a/drivers/ksfilter/ks/connectivity.c
+++ b/drivers/ksfilter/ks/connectivity.c
@@ -248,7 +248,7 @@ KspReadMediaCategory(
         return Status;
 
     /* allocate buffer for the registry key */
-    Path.Length = 0;
+    Path.Length = MediaPath.Length + GuidString.Length;
     Path.MaximumLength = MediaPath.MaximumLength + GuidString.MaximumLength;
     Path.Buffer = AllocateItem(NonPagedPool, Path.MaximumLength);
     if (!Path.Buffer)


drivers/ksfilter/ks/connectivity.c
```diff
@@ -283,7 +283,7 @@ KspReadMediaCategory(
     }
 
     /* query the name size */
-    Status = ZwQueryValueKey(hKey, &Name, KeyValuePartialInformation, NULL, 0, &Size);
+    Status = ZwQueryValueKey(hKey, &Name, KeyValuePartialInformation, NULL, 0, 0);
```
binarymaster, line 286: I don't understand the fix... `ULONG Size;` is a local variable. Yes, it's not initialized, but it's passed as a pointer to get the size...

You probably want to initialize Size to 0.

binarymaster (changes requested):

binarymaster:
> We're currently using uninitialized `Size` variable which corresponds for `KeyValueInformation` buffer size

Wat? :raised_eyebrow: It's 6th function argument, so it corresponds to `ResultLength`.

drivers/ksfilter/ks/connectivity.c
oleg-dubinskiy: > You probably want to initialize Size to 0.

I already tried that, but for some reason, it does not work for me. The system still fails with bugcheck. :frowning_face:

oleg-dubinskiy:
> it corresponds to ResultLength

Yes, it is. I just meant that it corresponds to the size of the buffer of `KeyValueInformation`, but not to that parameter directly. And I forgot to say that it's a pointer to that.
Sorry if I expressed my thoughts a bit wrongly. :sweat_smile:

binarymaster:
So, you pass NULL as the last parameter, `Size` variable remains uninitialized, and then it allocates the buffer with this uninitialized value... how it is supposed to work? :confused:

gonzoMD:
> Although the sound does no logner work with such replacement (at least with AC97 VirtualBox driver), but nevertheless it's booting.

Oo. Another regression. Joachim will go nuts. I hope this can be improved soon.

drivers/ksfilter/ks/connectivity.c
HBelusca: > /* query the name size */
>    Status = ZwQueryValueKey(hKey, &Name, KeyValuePartialInformation, NULL, 0, 0);

Then how and where do you retrieve the size? :trollface: 
No wonder the "new" code will be buggier than before.

From MSDN:
> [out] Pointer to a variable that receives the size, in bytes, of the key information. If the ZwQueryValueKey routine returns STATUS_SUCCESS, callers can use the value of this variable to determine the amount of data returned. If the routine returns STATUS_BUFFER_OVERFLOW or STATUS_BUFFER_TOO_SMALL, callers can use the value of this variable to determine the size of buffer required to hold the key information.

SergeGautherie:
As already pointed out, analysis and fix are both wrong.
Could you show relevant before/after debug log, with (updated) #3391? Thanks.

oleg-dubinskiy:
> Oo. Another regression.

No, it's not. Probably I said wrong again. :sweat_smile: 
I just meant it does not work **only after** replacing those drivers. It works without replacing them, and does not with replaced ones. As earlier, same now.
It cannot be a regression.

drivers/ksfilter/ks/connectivity.c
oleg-dubinskiy: OK @HBelusca. You are right. This fix is indeed wrong.
I just did this according to that OSR article, but it appeared to be the different problem from that one.
I found the actual source of problem instead and made a better fix.
The buffer is too small only because the length of the string stored in the buffer is set to `0`: https://github.com/reactos/reactos/blob/889eab78ca7c0f17a2ada9ff20de5807a293cdfe/drivers/ksfilter/ks/connectivity.c#L251
`Path` string contains `MediaPath`, which is a path of the registry key to be opened, plus `GuidString`, which is a string of media category, whose size is querying, converted from its input GUID.
So I decided that its length should be appropriate.
Indeed: after setting `Path.Length` to `MediaPath.Length + GuidString.Length`, same as it done for `Path.MaximumLength`, the system is booting for me again with MS audio stack. :smiley: 
So it's not needed to keep the buffer uninitialized.
I will force-push my branch by this fix soon. Allow me some time.

oleg-dubinskiy: Fixed. :ok_hand:

drivers/ksfilter/ks/connectivity.c
```diff
@@ -248,7 +248,7 @@ KspReadMediaCategory(
         return Status;
 
     /* allocate buffer for the registry key */
-    Path.Length = 0;
+    Path.Length = MediaPath.Length + GuidString.Length;
```
HBelusca, line 251: This is wrong.
The `Path.Length` value is correctly updated by the `RtlAppendUnicodeStringToString` calls from below.
See also its implementation https://git.reactos.org/?p=reactos.git;a=blob;f=sdk/lib/rtl/unicode.c;hb=889eab78ca7c0f17a2ada9ff20de5807a293cdfe#l209

HBelusca (changes requested):

drivers/ksfilter/ks/connectivity.c
oleg-dubinskiy: OK, @HBelusca. I found out that this fix is indeed wrong. As it appeared, the problem is not caused by our Kernel Streaming (ks.sys) at all. The actual source of problem seems to be in PnP instead. See my new #3510 PR for details.

oleg-dubinskiy:
Closing as invalid, since this fix is indeed wrong. Opened a new PR with the better one.

PR #3391, opened 2020-12-15
[KS] Add more diagnostic DPRINTs in KspReadMediaCategory CORE-17361

oleg-dubinskiy:
## Purpose

Add some more diagnostic `DPRINT`s in two `ZwQueryValueKey` calls of `KspReadMediaCategory`, when querying the size of the buffer and when reading the media categories names from registry. For both cases: on success and on failure.
They helped me to track the actual problem with fail to loading MS audio stack in ReactOS (portcls.sys, sysaudio.sys and wdmaud.sys from Windows XP/2003). Btw, I already fixed it in my previous #3390 PR. :slightly_smiling_face: 

### DIFF
--- a/drivers/ksfilter/ks/connectivity.c
+++ b/drivers/ksfilter/ks/connectivity.c
@@ -284,9 +284,13 @@ KspReadMediaCategory(
 
     /* query the name size */
     Status = ZwQueryValueKey(hKey, &Name, KeyValuePartialInformation, NULL, 0, &Size);
+
+    DPRINT("ZwQueryValueKey() status 0x%08lx %wZ\n", Status, &Name);
+
     if (!NT_SUCCESS(Status) && Status != STATUS_BUFFER_TOO_SMALL)
     {
         /* failed to query for name key */
+        DPRINT1("ZwQueryValueKey() failed with status 0x%08lx\n", Status);
         ZwClose(hKey);
         return Status;
     }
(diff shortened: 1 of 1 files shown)


drivers/ksfilter/ks/connectivity.c
```diff
@@ -284,9 +284,13 @@ KspReadMediaCategory(
 
     /* query the name size */
     Status = ZwQueryValueKey(hKey, &Name, KeyValuePartialInformation, NULL, 0, &Size);
+
+    DPRINT("ZwQueryValueKey() status 0x%08lx %wZ\n", Status, &Name);
```
Fraizeraust, line 288: As much as you say the `DPRINT` calls helped you tracking down the problem you were investigating, would it make sense to log in the NTSTATUS code even on a success case? Why?

SergeGautherie: Why `Name`, which is already known? And not `Size`, which is queried?

> status is always successful

That feels rather odd: `STATUS_BUFFER_TOO_SMALL` seems expected in this pattern...

drivers/ksfilter/ks/connectivity.c
```diff
@@ -303,12 +307,15 @@ KspReadMediaCategory(
     /* now read the info */
     Status = ZwQueryValueKey(hKey, &Name, KeyValuePartialInformation, (PVOID)KeyInfo, Size, &Size);
 
+    DPRINT("ZwQueryValueKey() status 0x%08lx %wZ\n", Status, &Name);
+
```
GeoB99, line 311: I see no point in logging the status code here as most of the time it is `STATUS_SUCCESS` anyway.

GeoB99 (changes requested):

PR #3392, opened 2020-12-16
[WINSXS] Add some amd64 manifests

learn-more:
Based on b394e5a9a70845aec9e402c5b7571cf89f054228


dll/apisets/CMakeLists.txt
```diff
@@ -4,6 +4,12 @@
 
 project(apisets)
 
+if (ARCH STREQUAL "amd64")
+    set(SXSARCH "amd64")
+else()
+    set(SXSARCH "x86")
+endif()
+
```
HBelusca, line 12: Why introducing yet another variable?
https://git.reactos.org/?p=reactos.git;a=blob;f=CMakeLists.txt;hb=784725143e9b75d41888252b787b24d6f6b2213f#l34
There is `WINARCH` already by @tkreuzer 
> This introduces a new CMake variable WINARCH, which what Windows usually uses ("x86" instead of "i386") e.g. **in sxs names**.

learn-more: Because I copied this code from his commit :)

HBelusca (approved):

PR #3394, opened 2020-12-19
Implementation APIC ACPI HALs initialization.

vgalnt:
WIP. JIRA issue: [CORE-17378](https://jira.reactos.org/browse/CORE-17378)


hal/halx86/legacy/halmps.rc
```diff
@@ -0,0 +1,5 @@
+#define REACTOS_VERSION_DLL
+#define REACTOS_STR_FILE_DESCRIPTION  "Hardware Abstraction Layer DLL"
```
binarymaster, line 2:
```suggestion
#define REACTOS_STR_FILE_DESCRIPTION  "x86 APIC/MPS Multiprocessor Hardware Abstraction Layer"
```

hal/halx86/legacy/halapic.rc
```diff
@@ -0,0 +1,5 @@
+#define REACTOS_VERSION_DLL
+#define REACTOS_STR_FILE_DESCRIPTION  "Hardware Abstraction Layer DLL"
```
binarymaster, line 2:
```suggestion
#define REACTOS_STR_FILE_DESCRIPTION  "x86 APIC/MPS Uniprocessor Hardware Abstraction Layer"
```

hal/halx86/legacy/hal.rc
```diff
@@ -0,0 +1,5 @@
+#define REACTOS_VERSION_DLL
+#define REACTOS_STR_FILE_DESCRIPTION  "Hardware Abstraction Layer DLL"
```
binarymaster, line 2:
```suggestion
#define REACTOS_STR_FILE_DESCRIPTION  "x86 PIC Uniprocessor Hardware Abstraction Layer"
```

hal/halx86/acpi/halmacpi.rc
```diff
@@ -0,0 +1,5 @@
+#define REACTOS_VERSION_DLL
+#define REACTOS_STR_FILE_DESCRIPTION  "Hardware Abstraction Layer DLL"
```
binarymaster, line 2:
```suggestion
#define REACTOS_STR_FILE_DESCRIPTION  "x86 APIC/ACPI Multiprocessor Hardware Abstraction Layer"
```

hal/halx86/acpi/halacpi.rc
```diff
@@ -1,5 +1,5 @@
 #define REACTOS_VERSION_DLL
-#define REACTOS_STR_FILE_DESCRIPTION  "X86 Uniprocessor Hardware Abstraction Layer"
-#define REACTOS_STR_INTERNAL_NAME     "halacpi"
+#define REACTOS_STR_FILE_DESCRIPTION  "Hardware Abstraction Layer DLL"
```
binarymaster, line 2:
```suggestion
#define REACTOS_STR_FILE_DESCRIPTION  "x86 PIC/ACPI Uniprocessor Hardware Abstraction Layer"
```

hal/halx86/acpi/halaacpi.rc
```diff
@@ -0,0 +1,5 @@
+#define REACTOS_VERSION_DLL
+#define REACTOS_STR_FILE_DESCRIPTION  "Hardware Abstraction Layer DLL"
```
binarymaster, line 2:
```suggestion
#define REACTOS_STR_FILE_DESCRIPTION  "x86 APIC/ACPI Uniprocessor Hardware Abstraction Layer"
```

hal/halx86/CMakeLists.txt
```diff
@@ -67,27 +67,32 @@ if(ARCH STREQUAL "i386")
...
+    # Legacy HALs
+    add_hal(hal SOURCES legacy/hal.rc COMPONENTS generic legacy hal)
+    add_hal(halapic SOURCES legacy/halapic.rc COMPONENTS generic legacy halapic)
+    add_hal(halmps SOURCES legacy/halmps.rc COMPONENTS generic legacy halmps)
+
+    # ACPI HALs
+    add_hal(halacpi SOURCES acpi/halacpi.rc COMPONENTS generic acpi halacpi)
+    add_hal(halaacpi SOURCES acpi/halaacpi.rc COMPONENTS generic acpi halaacpi)
+    add_hal(halmacpi SOURCES acpi/halmacpi.rc COMPONENTS generic acpi halmacpi)
+
```
binarymaster, line 92: I'm not a fan of this way of cmake scripting, I appreciate the previous way, so it's something like...
```c
    # Legacy HALs
    add_hal(hal SOURCES legacy/hal.rc COMPONENTS generic legacy up pic)
    add_hal(halapic SOURCES legacy/halapic.rc COMPONENTS generic legacy up apic)
    add_hal(halmps SOURCES legacy/halmps.rc COMPONENTS generic legacy smp apic)

    # ACPI HALs
    add_hal(halacpi SOURCES acpi/halacpi.rc COMPONENTS generic acpi up pic)
    add_hal(halaacpi SOURCES acpi/halaacpi.rc COMPONENTS generic acpi up apic)
    add_hal(halmacpi SOURCES acpi/halmacpi.rc COMPONENTS generic acpi smp apic)
```
This way MPS definitions should be somewhere in `legacy` module. Note that it's not code suggestion, just a thought of how it should look like.

hal/halx86/apic/apic.c
```diff
@@ -85,6 +85,28 @@ HalVectorToIRQL[16] =
       29, /* E1 IPI_LEVEL / EF POWER_LEVEL */
       31, /* FF HIGH_LEVEL */
 };
+
+/* For 0x50..0xBF vectors IRQLs values saves dinamically in HalpAllocateSystemInterruptVector() */
```
binarymaster, line 89: Typo.
```suggestion
/* For 0x50..0xBF vectors IRQLs values saves dynamically in HalpAllocateSystemInterruptVector() */
```

hal/halx86/legacy/halmps.rc
vgalnt: From the point of view of compatibility with NT, all HALs have the same name ...

hal/halx86/CMakeLists.txt
vgalnt: I am not a fan :)
I just need the ability to individually compile a piece of code for a single HAL.
Example: halaacpi and halmacpi are almost the same except for spinlocks. This requires a separate compilation with the CONFIG_SMP key.
Another example: HaliHaltSystem() function - for ACPI HALs (halacpi, halaacpi, halmacpi) one option is used, for hal.dll another option, for halapic + halmps - this function is absent.
Another example: the arbiter is used only in hal.dll
It's like making LEGO from just 5 types of figures)
But I need a different type of figures...

vgalnt: MPS definitions are not needed separately, as this code will also be the same for halapic and halmps. The only difference is that spinlock.c will compile with or without CONFIG_SMP.

hal/halx86/apic/apic.c
vgalnt: Ok

Extravert-ir:
Is this PR finished? If not, mark it as a draft, otherwise I don't know whether I should start reviewing or not.

What I can say now:
- I don't like the idea of having 6 HALs, we can avoid this maintenance burden.
Unless somebody can point me to the hardware which strictly requires only it, I'd like to remove `halapic` and `halmps`. `halaacpi` is also questionable, but it doesn't need additional code, so I'm fine with it for now
- ACPI structure definitions should probably go to SDK, where we already have some. There is also `drivers/bus/acpica/include` which likely already has everything we need

vgalnt:
In this PR I am planning to implement initialization for halaacpi and halmacpi only. The code for them is actually the same.
Indeed, starting from NT6 there are only two HALs halacpi and halmacpi, and starting from NT6.3 only one halmacpi remains.
Since the current target OS version is NT 5.2, it is logical to leave all 6 HALs (but personally I do not plan to deal with legacy).
When (and if) the target OS version is changed to for example NT 6.1, then of course the other 4 HALs will not make sense.

vgalnt:
This PR is not yet complete.

vgalnt:
I suggest making the functional part of the code a priority (at this stage). All that concerns code formatting (code-equilibristics, etc.), then this can be done when RoS is already working on hardware :)

I would not like to sync headers with drivers/bus/acpica/include. At least for now, while WINVER is hardcoded to 0x502.

Extravert-ir:
> At least for now, while WINVER is hardcoded to 0x502.

How is that related?

vgalnt:
Windows 2003 was released on April 24, 2003
Windows 2003 R2 was released on December 6, 2005

The first revision of the ACPI specification was released in December 1996
The revision 2.0 was released in August 2000
The revision 3.0 was released in September 2004
The revision 4.0 was released in June 2009

Thus, while WINVER is hardcoded to 0x502, the RoS target revision of ACPI is hardcoded 3.0.

As far as I know "acpica" is synchronized with the latest revision of ACPI. Apparently there is no 100% guarantee that compatibility with 3.0 is not lost (or will not be lost in the future).
In addition, attempts to run the RoS driver ACPI.SYS on Windows 2003 were unsuccessful.

ColinFinck:
> Thus, while WINVER is hardcoded to 0x502, the RoS target revision of ACPI is hardcoded 3.0.

Vadim, ReactOS is not about creating an _exact_ reimplementation of Windows Server 2003. Don't mistake our current kernel target of NT 5.2 for that.

Whenever we can go with time without sacrificing compatibility, we should. ACPI is an example for that. It's complex enough that it makes no sense to write our own interpreter when Intel offers the mature ACPICA reference implementation. With that being used by Linux and FreeBSD as well, I believe we should be pretty safe from sudden incompatibilities to older systems.

On a related note, we're also not limiting our schannel.dll to the now useless TLS 1.1 just because Windows Server 2003 didn't support anything newer. We work on NT6+ application support and EFI booting, because that is needed in 2020. Even if Windows Server 2003 didn't come with it.


> In addition, attempts to run the RoS driver ACPI.SYS on Windows 2003 were unsuccessful.

That however likely is a real issue that should be investigated. Maybe @ThFabba knows more here.

SergeGautherie:
> > RoS target revision of ACPI is hardcoded 3.0.
> 
> ACPI is an example for that.

See, for example,
https://git.reactos.org/?p=reactos.git;a=blob;f=boot/freeldr/freeldr/include/arch/pc/pcbios.h;hb=HEAD

vgalnt:
Colin, I have nothing against Intel, ACPICA, Linux, FreeBSD, or any fancy stuff))

Since I really look at things, for me personally, the most important thing is the ability to run thousands of programs written for NT on those computers that run 5.1 or 5.2, respectively. When this "minimum bar" is "taken", you can "raise the height" to NT 6.x. But so far, this is far from it.

vgalnt:
https://github.com/SergeGautherie, I meant:

if (WINVER == 0x502)
  MAX_APCI = 3.0
else if (WINVER == 0x600)
  MAX_ACPI = 4.0
  ...
  ...

hal/halx86/legacy/halmps.rc
HBelusca: Since no program looks at the file description (i.e. there won't be any compatibility problem), we can give more descriptive names to indicate for what purpose given hal is.

vgalnt: @HBelusca, 
Are you 100% sure about that?
I am 100% sure that if I do the same as in NT, then there will definitely be no problems.
Is there any real use for these names?

HBelusca: >  Is there any real use for these names?

I don't think there is any real use for these names, except for the user who would be better informed about what that HAL dll is about.

SergeGautherie:
> @SergeGautherie, I meant:
> 
> if (WINVER == 0x502)
> MAX_APCI = 3.0
> else if (WINVER == 0x600)
> MAX_ACPI = 4.0

I simply gave; for what it's worth, an example of ACPI version not being restricted in FreeLdr.
I don't understand what your are trying to tell me...

vgalnt:
@SergeGautherie, when I said "Thus, while WINVER is hardcoded to 0x502, the RoS target revision of ACPI is hardcoded 3.0",
I meant that if RoS not uses a construction like
if (WINVER == 0x502)
MAX_APCI = 3.0
else if (WINVER == 0x600)
MAX_ACPI = 4.0
...
...
then the ACPI version should be limited. It will be more safe.

RepOfCATV95:
> @SergeGautherie, when I said "Thus, while WINVER is hardcoded to 0x502, the RoS target revision of ACPI is hardcoded 3.0",
> I meant that if RoS not uses a construction like
> if (WINVER == 0x502)
> MAX_APCI = 3.0
> else if (WINVER == 0x600)
> MAX_ACPI = 4.0
> ...
> ...
> then the ACPI version should be limited. It will be more safe.



> @SergeGautherie, when I said "Thus, while WINVER is hardcoded to 0x502, the RoS target revision of ACPI is hardcoded 3.0",
> I meant that if RoS not uses a construction like
> if (WINVER == 0x502)
> MAX_APCI = 3.0
> else if (WINVER == 0x600)
> MAX_ACPI = 4.0
> ...
> ...
> then the ACPI version should be limited. It will be more safe.

That will be impossible to change WINVER as long as both ACPI revision and WINVER are actually hardcoded.

SergeGautherie:
> the ACPI version should be limited. It will be more safe.

I understood what you wrote/suggested.
Again, I simply pointed out a FreeLdr (!= ROS itself) example.
Then, I don't understand what you expect me to do with your replies...

Extravert-ir:
This pull request is too large already, please consider finishing it.
After this one is reviewed and merged, we can continue on adding more code

vgalnt:
Unfortunately, at the moment I do not have enough free time.
I plan to finish this PR in 2-3 weeks (approximately).

vgalnt:
Two ISOs (livecd and bootcd) are available for tests
http://vgal.ru.com/downloads/live-ReactOS-0.4.15-apicacpi.4-vgal.iso.7z
http://vgal.ru.com/downloads/bootcd-ReactOS-0.4.15-apicacpi.4-vgal.iso.7z

rox4d:
I was try to boot this live-cd from IODD Virtual CD in a three PC, and in most recent PC (FX8320e CPU in ASRock 970 Extreme3 R2.0 board) it was only successful while halacpi selected in boot menu (booting other two PC was failed before selection menu), not shure, but possible i can provide more useful info (how)?

vgalnt:
Thank you, rox4d!

So far, it makes no sense to expect positive results on the real computer. To do this, at least perform this (or similar) roadmap [https://jira.reactos.org/browse/CORE-16841].

This PR is only one (but necessary) step towards using ROS on the HW.

By the way, if you test in VBox, you need to include ACPI in the virtual machine settings
VBox-APIC

Extravert-ir:
x64 boot regressed for me with an infinite recursion in `HalpInitPhase0`. That should be fixed

vgalnt:
Unfortunately, I do not have the possibility of developing and debugging for x64. Therefore, I will be glad any help (logs, trouble finding, etc.).

hal/halx86/acpi/busemul.c
```diff
@@ -14,8 +14,26 @@
...
+extern ULONG HalpShutdownContext;
+extern ULONG HalpPicVectorRedirect[HAL_PIC_VECTORS];
+
 /* PRIVATE FUNCTIONS **********************************************************/
 
+VOID
+NTAPI
+HalpCheckPowerButton(VOID)
+{
+    if ((!KiBugCheckData[0] && !InbvCheckDisplayOwnership()) ||
```
HBelusca, line 26: What this function is expected to do?
(And since you've added the Inbv check, does that mean it is also supposed to display some text?)

Extravert-ir:
I'm in the process of reviewing the pull request.
Two points I'd like to say now:

1. Please, leave only `halaacpi` and `halmacpi` and remove everything else. I described my point above already, other HALs are useless at this time, and we for sure don't have the manpower to support them.
2. Remove all not implemented functions too. They don't add any value without the code or a comment at least

Extravert-ir:
I rearranged files in `halx86` folder, similar to how you did here. So now this PR should have smaller diff, after rebase

hal/halx86/acpi/busemul.c
vgalnt: NT 5.2 ACPI Driver and Acpi Hal have an interface:
https://jira.reactos.org/browse/CORE-17378?focusedCommentId=127590&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-127590

When calling HaliAcpiMachineStateInit(), ACPI driver transmits ACPI HAL information, which is used including to fill HalpShutdownContext. It is used to call the HaliAcpiSleep() function under certain circumstances (for example, when HalpDisableHibernate is TRUE or HalpDisableS5Hibernation is TRUE).

ACPI HAL calls InbvAcquireDisplayOwnership() in two cases:
from HalHandleNMI(), which ends with halt;
from HalReturnToFirmware(), which then causes HalpReboot().

Thus, this function is performed when
if (KiBugCheckData[0] || InbvCheckDisplayOwnership()) &&
    HalpShutdownContext != 0)
{
   ...
}

This function reads and writes to registers
PM1a_EVT_BLK, PM1b_EVT_BLK, PM1a_CNT_BLK, PM1b_CNT_BLK (see Fixed ACPI Description Table (FADT) decription in ACPI specification).

Until there is an interface with the ACPI driver, it makes no sense to implement this function. It's just a trap for the future.

vgalnt:
Extravert-ir, I am not a supporter of removing the previous HAL.
In one issue of the Russian intellectual television game "What? Where? When?" connoisseurs lost to viewers when asked about loose sweets in stores:
"There are two shops near the school. They sell the same sweets for the same weight by weight. But children always buy sweets in only one store. Why?"
The answer turned out to be:
“In one store, the seller put a lot of sweets on the scales, and then removed the excess to the desired weight. In another, the seller put a few sweets on the scales and added to the desired weight.
The children chose the store where sweets are added to them. "

To leave the previous HAL types, and at the same time not waste time on them, I suggest leaving them as they are now. You can just add a new directory to reactos_dir\hal\... It could be "halx86_new" or "halmacpi" or whatever. In this new folder, you can develop halaacpi and halmacpi without the old unnecessary code.

Extravert-ir:
You didn't get my point, probably. I'm talking about the code you added in this pull request.
MPS code does not exist at all in our code base (except old stuff in `mp` folder, but it is not being compiled for a long time), let's not add it. Existing `hal` and `halacpi` are not touched.
So in the end there will be 4 HALS:
- PIC HAL / hal.dll
- ACPI PIC HAL / halacpi.dll
- ACPI APIC HAL / halaacpi.dll <-- the usefulness of this one is questionable, but let's keep it for now
- ACPI APIC SMP HAL / halmacpi.dll
- At some point UEFI HAL will also be added
PIC HALs are for i386, APIC ones are for both i386 and amd64, UEFI HAL is for amd64 only

All other stuff is possible, but will be considered "a spin-off" platform, like dedicated HALs for xbox and PC-98. These HALs are supported as soon as there are maintainers for them (@binarymaster and @disean respectively). Are you going to support and test MPS HALs on a real hardware?

Here at ReactOS we need to narrow our goals, not widen them even more.

> You can just add a new directory to reactos_dir\hal...

No, enough *_new folders. It took quite some time to eliminate most of them in our code base
> The children chose the store where sweets are added to them

Code is not sweets. Less code - less bugs, less work to do.

hal/halx86/acpi/halacpi.c
```diff
@@ -35,15 +36,339 @@ PACPI_BIOS_MULTI_NODE HalpAcpiMultiNode;
...
+        {
+            TotalStall += StallCounter;
+
+            _mm_pause();
+
+            do
+            {
+                StallCounter--;
+            }
+            while (StallCounter);
```
Extravert-ir, line 278: This needs to be refactored

hal/halx86/acpi/halacpi.c
```diff
@@ -35,15 +36,339 @@ PACPI_BIOS_MULTI_NODE HalpAcpiMultiNode;
...
+    OutPerfCount->QuadPart = TimerInfo.PerformanceCounter.QuadPart + QueryTimer();
+
+    if (OutPerfFrequency) {
+        OutPerfFrequency->QuadPart = PMTimerFreq;
+    }
+}
+
+ULONGLONG
+__cdecl
+HalpQueryPerformanceCounter(VOID)
```
Extravert-ir, line 195: I'd place ACPI PM timer stuff into a separate file. This one is getting big

APIC HAL uses ACPI PM timer instead of TSC counter like we do? Or this is just for ACPI PIC HAL?

I'd say, timer stuff is a big topic by itself. We should implement a common way of switching between clock sources (PM timer, TSC, HPET, PIT). Deserves a separate PR...

Extravert-ir (changes requested):

hal/halx86/generic/usage.c
```diff
@@ -32,71 +32,18 @@ USHORT HalpComPortIrqMapping[5][2] =
 
 ADDRESS_USAGE HalpComIoSpace =
 {
-    NULL, CmResourceTypePort, IDT_INTERNAL,
+    NULL, CmResourceTypePort, IDT_DEVICE,
     {
-        {0x2F8,   0x8},     /* COM 1 */
+        {0x2F8,   0x8},     /* COM 2 */
```
HBelusca, line 37: I confirm this change (see e.g. boot/freeldr/freeldr/lib/comm/rs232.c line 43; boot/freeldr/freeldr/ntldr/headless.c line 125+; ntoskrnl/inbv/inbvport.c line 136+).
And I also wonder, why does the ComIoSpace only contains info for the COM2 and not the COM1?

hal/halx86/acpi/halacpi.c
```diff
@@ -785,6 +905,60 @@ HaliAcpiTimerInit(IN ULONG TimerPort,
...
+        return;
+    }
+
+    DPRINT1("HalaAcpiTimerInit: HalpBrokenAcpiTimer\n");
+    DbgBreakPoint();
+}
+
+VOID
+NTAPI
+HaliAcpiTimerInit(_In_ PULONG TimerPort,
```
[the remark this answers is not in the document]
vgalnt: I do not have learning, development, testing and debugging for x64. Therefore, I try to not change the code for x64 as much as possible.

hal/halx86/acpi/halacpi.c
```diff
@@ -843,7 +1022,11 @@ HalpSetupAcpiPhase0(IN PLOADER_PARAMETER_BLOCK LoaderBlock)
     }
 
     /* Setup the ACPI timer */
+#ifdef _M_AMD64
     HaliAcpiTimerInit(0, 0);
+#else
+    HaliAcpiTimerInit(NULL, FALSE);
```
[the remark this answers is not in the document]
vgalnt: x64...

hal/halx86/acpi/halacpi.c
```diff
@@ -1112,4 +1295,114 @@ HalReportResourceUsage(VOID)
...
+    Value.QuadPart = TimerInfo.AcpiTimeValue.QuadPart + TimerInfo.ValueExt;
+    Value.QuadPart += ((Value.LowPart ^ Time) & TimerInfo.ValueExt);
+
+    TimerInfo.TimerCarry = Value.HighPart;
+    TimerInfo.AcpiTimeValue.QuadPart = Value.QuadPart;
+}
+
+VOID
+NTAPI
+HalAcpiBrokenPiix4TimerCarry(VOID)
```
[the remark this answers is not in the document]
vgalnt: This is necessary for some chipsets. In fact, this is a black list in the registry (HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\HAL\...). 
Control_HAL
This variable is set in Halppiix4Detect() if Bit0 is set to the registry value for this chipset.
Since to work with HW, first of all, need to change the ACPI driver, I did not add Halppiix4Detect() (and a few more subfunctions).

hal/halx86/generic/halinit.c
```diff
@@ -96,44 +75,20 @@ HalInitSystem(IN ULONG BootPhase,
...
-        /* Initialize the clock */
-        HalpInitializeClock();
-
-        /*
-         * We could be rebooting with a pending profile interrupt,
-         * so clear it here before interrupts are enabled
-         */
-        HalStopProfileInterrupt(ProfileTime);
-
+#if !defined(SARCH_PC98) && !defined(SARCH_XBOX)
```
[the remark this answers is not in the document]
vgalnt: Ok

hal/halx86/generic/usage.c
```diff
@@ -32,71 +32,18 @@ USHORT HalpComPortIrqMapping[5][2] =
...
         {0,0},
     }
 };
 
+#ifdef _M_AMD64
```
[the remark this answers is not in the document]
vgalnt: x64... I will try to transfer it. Please check for 64-bit

hal/halx86/include/halp.h
```diff
@@ -173,6 +173,7 @@ typedef struct
     UCHAR BusReleativeVector;
 } IDTUsage;
 
+#pragma pack(push, 1)
```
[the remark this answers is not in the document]
vgalnt: Remove

hal/halx86/include/halp.h
```diff
@@ -514,6 +531,34 @@ HalpInitProcessor(
     IN PLOADER_PARAMETER_BLOCK LoaderBlock
 );
 
+CODE_SEG("INIT")
```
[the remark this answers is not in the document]
vgalnt: Ok

hal/halx86/apic/apicacpi.c
```diff
@@ -0,0 +1,616 @@
...
+    if (Phase == 0 && HalpProcLocalApicTable == NULL)
+    {
+        /* First initialization */
+
+        Header = (PACPI_SUBTABLE_HEADER)&ApicTable[1];
+        TableEnd = (ULONG_PTR)ApicTable + ApicTable->Header.Length;
+
+        HalpProcLocalApicTable = HalpStaticProcLocalApicTable;
+
+        if ((ULONG_PTR)Header < TableEnd)
```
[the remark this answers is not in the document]
vgalnt: Ok

hal/halx86/generic/usage.c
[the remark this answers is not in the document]
vgalnt: By the way, the Flags field values are distinguished for them (IDT_INTERNAL, IDT_DEVICE). Therefore ...

hal/halx86/generic/halinit.c
```diff
@@ -14,31 +14,10 @@
...
-        /* Read the command line */
-        CommandLine = LoaderBlock->LoadOptions;
-
-        /* Check if PCI is locked */
-        if (strstr(CommandLine, "PCILOCK")) HalpPciLockSettings = TRUE;
-
-        /* Check for initial breakpoint */
-        if (strstr(CommandLine, "BREAK")) DbgBreakPoint();
-    }
-}
```
ColinFinck, line 40: Why do we have this function duplicated 3 times in minihal, pic, and xbox HAL now instead of keeping it here?

hal/halx86/acpi/acpitimer.c
```diff
@@ -97,85 +101,86 @@ NTAPI
...
     StallCounter = StallLoopValue;
 
     while (TRUE)
     {
+        /* Do one microsecond stall */
         while (TRUE)
         {
-            TotalStall += StallCounter;
-
-            _mm_pause();
```
disean: YieldProcessor()

hal/halx86/acpi/halacpi.c
vgalnt: Now it in 'acpitimer.c'

hal/halx86/acpi/halacpi.c
vgalnt: Done

hal/halx86/generic/usage.c
vgalnt: HalpComIoSpace is used if KD is using a COM port
HalpComIoSpace

In the time of NT5.1 | 2, computers motherboards have had a back connector for COM ports
COM-connectors

In those days, the COM1 was usually used for modems
U S  Robotics V  Everything Modem Model 34538 Courier 56K Business

I think that in order not to touch the COM1 port, the null-modem cable has been connected to the COM2 port by default.

hal/halx86/generic/halinit.c
vgalnt: Because I added an variant for this function (for APIC) and a conflict appeared
HalpGetParameters

hal/halx86/acpi/acpitimer.c
vgalnt: This is changed after the function refactoring.

hal/halx86/include/ioapic.h
```diff
@@ -74,6 +84,35 @@ typedef struct _IOAPIC_INFO
...
+FORCEINLINE
+VOID
+IoApicWrite(_In_ PIO_APIC_REGISTERS IoApicRegs,
+            _In_ UCHAR RegisterIdx,
+            _In_ ULONG Value)
+{
+    //DbgPrint("IoApicWrite: RegisterIdx %X, Value %X\n", RegisterIdx, Value);
+
+    *(volatile PUCHAR)((ULONG_PTR)IoApicRegs + IOAPIC_IOREGSEL) = RegisterIdx;
+    *(volatile PULONG)((ULONG_PTR)IoApicRegs + IOAPIC_IOWIN) = Value;
```
HBelusca, line 113: Not using WRITE_REGISTER_ULONG or stuff?

hal/halx86/include/ioapic.h
```diff
@@ -74,6 +84,35 @@ typedef struct _IOAPIC_INFO
...
+#ifndef _M_AMD64
+FORCEINLINE
+ULONG
+IoApicRead(_In_ PIO_APIC_REGISTERS IoApicRegs,
+           _In_ UCHAR RegisterIdx)
+{
+    ULONG Result;
+
+    *(volatile PUCHAR)((ULONG_PTR)IoApicRegs + IOAPIC_IOREGSEL) = RegisterIdx;
+    Result = *(volatile PULONG)((ULONG_PTR)IoApicRegs + IOAPIC_IOWIN);
```
HBelusca, line 98: Not using READ_REGISTER_ULONG ?

sdk/include/ndk/halfuncs.h
```diff
@@ -152,20 +152,34 @@ HalEnableSystemInterrupt(
...
 VOID
 NTAPI
 HalEndSystemInterrupt(
     _In_ KIRQL Irql,
-    _In_ PKTRAP_FRAME TrapFrame
+    _In_ PHAL_INTERRUPT_CONTEXT IntContext
 );
 #else
+/* NT actually uses the stack to place the pointer to the TrapFrame (really the third parameter),
+   but ... HalEndSystemInterrupt() is defined with only two parameters ...
```
HBelusca, line 173: Which version of NT? XP? 2003?

hal/halx86/include/ioapic.h
disean: It is done to avoid function call

HBelusca: But when used within the HAL itself, these are not functions:
https://git.reactos.org/?p=reactos.git;a=blob;f=sdk/include/ddk/ioaccess.h;hb=40f7e863472d2b0a662c06a95316997ab7c62d2b#l31

hal/halx86/include/ioapic.h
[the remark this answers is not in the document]
HBelusca: Please use the READ/WRITE_REGISTER_xxxx macros (see my other comment).

binarymaster:
image

Hmm... didn't indented to do this, only deleted my own comments :thinking:

hal/halx86/include/ioapic.h
disean: I've checked halpc98 DLL using dependency walker and I see *_REGISTER_UCHAR in the list.

HBelusca: Yes of course, macros forms are used within the HAL, but then it also needs to export these as functions for the other kernel modules to use them.

HBelusca: Interesting, so the NTOSKRNL exports the WRITE/READ_REGISTER_xxx, but it's the HAL that exports the WRITE/READ_PORT_xxx 😆 .
Anyway, if you use the header file I mentioned above, you should be able to use these things as macros.

Extravert-ir:
At the moment we're in a kind of stuck situation with this PR. Merging the PR in its current situation is a no-go, it basically creates two code paths for i586 and amd64 in a place where it is not needed. After merging, it will be way more hard to resolve this issue.

Somebody needs to merge this code with the old APIC code (and with amd64 code). I honestly don't know what can be the best way to handle this.
I have two suggestions:
- ACPI Timer can be merged separately, in a different pull request
- Volunteer opens a second PR based on this one, which includes changes for amd64 platform too. We will continue the review there

hal/halx86/include/ioapic.h
vgalnt: Yes, with "ioaccess.h"it ok.

sdk/include/ndk/halfuncs.h
vgalnt: 2003

sdk/include/ndk/halfuncs.h
```diff
@@ -152,20 +152,34 @@ HalEnableSystemInterrupt(
...
 NTHALAPI
 VOID
 NTAPI
 HalEndSystemInterrupt(
     _In_ KIRQL Irql,
-    _In_ PKTRAP_FRAME TrapFrame
+    _In_ PHAL_INTERRUPT_CONTEXT IntContext
 );
 #else
+/* NT actually uses the stack to place the pointer to the TrapFrame (really the third parameter),
```
binarymaster, line 172:
```suggestion
/* NT 5.2 actually uses the stack to place the pointer to the TrapFrame (really the third parameter),
```

vgalnt:
Extravert-ir, I want to note that this PR was made based on my code, which was debugged and tested, including on real computers. But I had to adapt it (I had it a separate, independent HAL) ...

My opinion is that due to the small number of developers, you shouldn't be sprayed with 32 and 64 bits at the same time.
"Better to have a titmouse in hand than pie in the sky."
The solution usually depends on the final task at hand ...
You can, for example, strive to create an "ideal" code, or you can strive for this code to work on real hardware (even if it won't be so beautiful).

hal/halx86/apic/apic.c
```diff
@@ -800,6 +1370,36 @@ KfRaiseIrql(
...
+KfRaiseIrql(_In_ KIRQL NewIrql)
+{
+    PKPCR Pcr = KeGetPcr();
+    KIRQL OldIrql;
+
+    OldIrql = Pcr->Irql;
+    Pcr->Irql = NewIrql;
+
+    return OldIrql;
+}
```
Extravert-ir, line 1402: Our current APIC code writes a value to the TPR register here (see `ApicLowerIrql` and `ApicSetIrql` from above), and that is a correct behavior (aka "what Windows does").

What's going on here, and why have you changed that?

vgalnt: Let's start by not using this code for amd64 now...
It is also a fact that the current x86 APIC HALs do not work.
Unlike my variant.
I think the statement "and that is a correct behavior (aka" what Windows does ")." is controversial. It is just more suitable for my code)

hal/hal.spec
```diff
@@ -18,7 +18,7 @@
 @ stdcall HalDisableSystemInterrupt(long long)
 @ stdcall HalDisplayString(str)
 @ stdcall HalEnableSystemInterrupt(long long long)
-@ stdcall -arch=i386,arm HalEndSystemInterrupt(long long)
+@ stdcall -arch=i386,arm HalEndSystemInterrupt(long long) #NT use nonstandard parameters calling
```
HBelusca, line 21: What kind of nonstandard calling it uses?

sdk/include/ndk/halfuncs.h
```diff
@@ -152,20 +152,34 @@ HalEnableSystemInterrupt(
 );
 
 #ifdef __REACTOS__
+/* For APIC, the interrupt vector must also be transmitted. */
+typedef struct _HAL_INTERRUPT_CONTEXT
+{
+    PKTRAP_FRAME TrapFrame;
+    KIRQL Irql;
```
HBelusca, line 159: Isn't that redundant with the `Irql` parameter of `HalEndSystemInterrupt()`?

hal/halx86/apic/apic.c
```diff
@@ -671,87 +858,470 @@ HalDisableSystemInterrupt(
...
+                       _In_ KIRQL OldIrql)
+{
+    KIRQL Irql = HalpVectorToIRQL[IntContext->Vector >> 4];
+
+    if (Irql < KeGetCurrentIrql())
+    {
+        HalpLowerIrqlHardwareInterrupts(Irql);
+    }
+
+    if (ApicRead(APIC_PPR) != (IntContext->Vector & 0xF0))
```
Extravert-ir, line 1261: From the Intel SDM (Vol. 3A, 10-29)
Screenshot_20210412_022213


So you're touching an undocumented area here. What do you expect to see? Does it work on all platforms?

hal/halx86/apic/apic.c
```diff
@@ -671,87 +858,470 @@ HalDisableSystemInterrupt(
...
+
+static
+VOID
+NTAPI
+HalpEndSystemInterrupt(_In_ PHAL_INTERRUPT_CONTEXT IntContext,
+                       _In_ KIRQL OldIrql)
+{
+    KIRQL Irql = HalpVectorToIRQL[IntContext->Vector >> 4];
+
+    if (Irql < KeGetCurrentIrql())
```
Extravert-ir, line 1256: This seem to be the only place where `Vector` field is used (for which you needed to change how the interrupts are processed in the kernel https://github.com/reactos/reactos/pull/3394/commits/420be339b28ea4ec178623593141bf366ac5ca12)

When does this happen? If an interrupt with a higher IRQL appeared while we're processing this one? What if the purpose of `HalpLowerIrqlHardwareInterrupts` then?

hal/halx86/apic/apic.c
```diff
@@ -671,87 +858,470 @@ HalDisableSystemInterrupt(
...
+
+        _enable();
+        return TRUE;
+    }
+
+    pPrcbVector = (PUCHAR)KeGetCurrentPrcb()->HalReserved;
+
+    Idx = pPrcbVector[0];
+    pPrcbVector[Idx + 1] = SystemVector;
+    pPrcbVector[0] = Idx + 1;
```
Extravert-ir, line 1242: You seem to use `HalReserved` field of the PKRCB. What is the meaning of the first byte and the next `UINT32`s?

hal/hal.spec
vgalnt: https://jira.reactos.org/browse/CORE-17378?focusedCommentId=127852&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-127852

hal/halx86/apic/apic.c
vgalnt: If the IRQL (which corresponds to the incoming (new) SystemVector) is less than the current IRQL, then this vector is stored in the "interrupt stack" (in HalReserved). This will subsequently be fetched (in a loop) and executed by HalpGenerateInterrupt (Vector) (asm - "int Vector").
By the way, don't confuse HalReserved in Pcr and Prcb.

first byte - current position in "interrupt stack"
next - stack values (SystemVector`s)

hal/halx86/apic/apic.c
vgalnt: In case the HAL must be NT compatible (no __REACTOS__) "Vector" is required. When the case is __REACTOS__ (like the current one), then theoretically it is possible. If you have an option for how to do this, then you can suggest it.

HalpLowerIrqlHardwareInterrupts(), as its name implies, is used to handle pending interrupts that were requested during the current interrupt (and whose IRQL was lower than the current interrupt IRQL).
If the incoming interrupt has a higher priority, then the processor is switched over to handle it.
If it is lower, then we need to remember it and process it later, after the completion of the current one.

hal/halx86/apic/apic.c
vgalnt: Fixed

sdk/include/ndk/halfuncs.h
vgalnt: Removed

hal/halx86/apic/apic.c
Extravert-ir: What was changed?
`(ApicRead(APIC_PPR) & 0xF0) != (IntContext->Vector & 0xF0)` still reads reserved area in PPR

vgalnt: 1. The reserved area is no longer used for comparison.
2. Does it say somewhere that this area cannot be read?

PR #3395, opened 2020-12-19
[FASTFAT] Add Optimize Speed patch to fastfat.

Doug-Lyons:
## Optimize fcb.c to speed up fastfat

_Optimize fcb.c for speed. Simplification of Optimize-Speed-3.patch._

JIRA issue: [CORE-14826](https://jira.reactos.org/browse/CORE-14826)

### DIFF
--- a/drivers/filesystems/fastfat/fcb.c
+++ b/drivers/filesystems/fastfat/fcb.c
@@ -370,7 +370,7 @@ _vfatReleaseFCB(
         ASSERT(pFCB->RefCount > 0);
         RefCount = --pFCB->RefCount;
 
-        if (RefCount == 1 && BooleanFlagOn(pFCB->Flags, FCB_CACHE_INITIALIZED))
+        if (RefCount == 0 && BooleanFlagOn(pFCB->Flags, FCB_CACHE_INITIALIZED))
         {
             PFILE_OBJECT tmpFileObject;
             tmpFileObject = pFCB->FileObject;


SergeGautherie:
> test for closing file to test for 0 handles open

Not exactly: this block looks more like cache uninitialization and reference count.

While this change looks fine, it seems to be a workaround (keeping "unused" cache around) until delayed close is (fully) supported,
so I would suggest to use an `#ifdef CORE_14917_IS_FIXED`.

Doug-Lyons:
> > test for closing file to test for 0 handles open
> 
> Not exactly: this block looks more like cache uninitialization and reference count.
> 
> While this change looks fine, it seems to be a workaround (keeping "unused" cache around) until delayed close is (fully) supported,
> so I would suggest to use an `#ifdef CORE_14917_IS_FIXED`.

@SergeGautherie, You may be correct here. Since I have not analyzed this in depth, so I changed the description to just be more generic. Thanks.

SergeGautherie:
> whether this is basically a partial revert

According to Pierre's commit comment back then, `RefCount` cannot reach `0` while FO/cache exists.
Then, your change should disable Pierre's new block actually, thus restoring most of the old behavior.

To (double-)check whether that still holds true, he suggests to test 'directories intensive operation (like extraction the ReactOS source code in ReactOS ;-))'...

Doug-Lyons:
> > whether this is basically a partial revert
> 
> According to Pierre's commit comment back then, `RefCount` cannot reach `0` while FO/cache exists.
> Then, your change should disable Pierre's new block actually, thus restoring most of the old behavior.
> 
> To (double-)check whether that still holds true, he suggests to test 'directories intensive operation (like extraction the ReactOS source code in ReactOS ;-))'...

@SergeGautherie, I have changed nothing in how the RefCount is incremented or decremented. This change only changes what happens when the RefCount gets to 1 (it disables part of the functionality) and when it gets to 0 (it performs functions that used to occur when the RefCount got to 1). 

If we use Pierre Switzer's "FastFAT KDBG Extensions" (see fastfat\iface.c Line # 157) with the new patch I get 211 entries and in https://jira.reactos.org/browse/CORE-14557 at https://jira.reactos.org/secure/attachment/45919/refs_at_boot.txt he had 390 entries.
```
kdb:> ? fat.vols
(drivers/filesystems/fastfat/kdbg.c:57) Volume: B0EE2018 with VCB: B0EE20D0
kdb:> ?fat.files B0EE2018
(drivers/filesystems/fastfat/kdbg.c:117) FCB B0F241D8 (ref: 4, oc: 0 U C ND) for FO 00000000 with path: \
...
```
So, based on what we see in this comparison, we are much smaller and this COULD indicate that we are not holding these unnecessary cached FCB's in memory.
Thanks. Doug Lyons

Extravert-ir:
This is obsolete now, with the adoption of the new driver

PR #3396, opened 2020-12-20
[NTOS:PNP][NTOS:IO] Refactoring of the driver initialization code

Extravert-ir:
Purpose: make the code more clean, remove duplication and add a bit of new features

- Do not create a device object + node on every driver load
- Remove the usage of IopCreateDeviceNode and change it to PipAllocateDeviceNode where required
- Use consolidated registry.inf for generating SETUPREG.HIV (for having all `add_registry_inf` data there too)
- Remove IopCreateDriver and put its code into IoCreateDriver and IopInitializeDriverModule. It's hard to extract a meaningful common part from it
- Refactor IopInitializeDriverModule. Extend and put the DriverName generation logic into it. Now this function frees the ModuleObject in case of failure and returns STATUS_FAILED_DRIVER_ENTRY in case of DriverInit failure (will be used later)
- Split IopLoadUnloadDriver into IopLoadDriver and calling DriverUnload
- Schedule the worker for (un)loading driver in a separate routine (IopDoLoadUnloadDriver) this allows IopLoadDriver to be called separately (if we are sure that we're in the system process)
- Do not hold the IopDriverLoadResource while trying to reference a driver object (but still acquire it when we actually need to load a driver)
- Change IopLoadDriver and IopInitializeDriverModule to use registry handle instead of a service name string and/or full registry path
- Do not try to reference a driver object inside IopLoadDriver. It's supposed to be done before the function call
- Move the driver's name obtaining logic into the IopGetDriverNames function
- Create a new PiCallDriverAddDevice instead of PipCallDriverAddDevice and move it to pnpmgr/devaction.c file. Move around all its internal helpers too
- Support a proper Windows-compatible driver loading order (lower filters, device servers, upper filters, etc.)
- Set a correct Problem for the DeviceNode, in case of an error during driver loading
- Check the Start Type for all drivers before loading 
- Do not try to load drivers during the early boot stage when there is no disk subsystem initialized
- Start loading drivers from the disk only after SystemRoot is initialized


ntoskrnl/io/pnpmgr/devnode.c
```diff
@@ -84,7 +129,7 @@ PipAllocateDeviceNode(
  *
  * @return     Status, indicating the result of an operation
  */
-
+#if 0
```
HBelusca, line 132: Is that the thing about the legacy drivers that won't completely be supported now?

Extravert-ir: Yeah, I left it because of that. But tbh, I'm not sure if those entries are enough for legacy devices to work properly

binarymaster:
Regresso introduced: https://jira.reactos.org/browse/CORE-17532

binarymaster:
Another regresso introduced: https://jira.reactos.org/browse/CORE-18725

PR #3397, opened 2020-12-20
[KMTESTS:PS] Implement the basic kernel suite tests for process quota…

GeoB99:
… routines

## Purpose
Implement the basic KMTEST for process quota stuff. Currently the tests are covered for `PsCharge*` kernel routines for now, further tests will be added when needed.
**Windows Server 2003**
Capture
**ReactOS**
Capture2


GeoB99:
I realised it's not quite the right way to use the function(s) to charge the quota, so I'm gonna close this. I'll be back into it when I have time.

PR #3398, opened 2020-12-23
[SHELL32] Determine if the FS is of type FATX

GeoB99:
## Purpose
Do not outlaw FATX as an unknown filesystem when it's already proposed in the combo listbox.
**JIRA Issue:** [CORE-17272](https://jira.reactos.org/browse/CORE-17272)

### DIFF
--- a/dll/win32/shell32/dialogs/drive.cpp
+++ b/dll/win32/shell32/dialogs/drive.cpp
@@ -81,6 +81,27 @@ GetDefaultClusterSize(LPWSTR szFs, PDWORD pClusterSize, PULARGE_INTEGER TotalNum
         else
             return FALSE;
     }
+    else if (!wcsicmp(szFs, L"FATX"))
+    {
+        if (TotalNumberOfBytes->QuadPart <= (16 * 1024 * 1024))
+            ClusterSize = 2048;
+        else if (TotalNumberOfBytes->QuadPart <= (32 * 1024 * 1024))
+            ClusterSize = 512;
+        else if (TotalNumberOfBytes->QuadPart <= (64 * 1024 * 1024))
+            ClusterSize = 1024;
+        else if (TotalNumberOfBytes->QuadPart <= (128 * 1024 * 1024))
+            ClusterSize = 2048;
+        else if (TotalNumberOfBytes->QuadPart <= (256 * 1024 * 1024))
+            ClusterSize = 4096;
+        else if (TotalNumberOfBytes->QuadPart <= (8192LL * 1024LL * 1024LL))
+            ClusterSize = 2048;
+        else if (TotalNumberOfBytes->QuadPart <= (16384LL * 1024LL * 1024LL))
+            ClusterSize = 8192;
+        else if (TotalNumberOfBytes->QuadPart <= (32768LL * 1024LL * 1024LL))
+            ClusterSize = 16384;
+        else
+            return FALSE;
+    }
     else if (!wcsicmp(szFs, L"NTFS"))
     {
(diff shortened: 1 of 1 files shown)


dll/win32/shell32/dialogs/drive.cpp
```diff
@@ -41,6 +41,7 @@ GetDefaultClusterSize(LPWSTR szFs, PDWORD pClusterSize, PULARGE_INTEGER TotalNum
     DWORD ClusterSize;
 
     if (!wcsicmp(szFs, L"FAT16") ||
+        !wcsicmp(szFs, L"FATX") ||
```
learn-more, line 44: Also a reactos hack?

GeoB99: No. The boot sector system string is always `FATX` regardless of what kind of format should be performed internally, FATX16 or FATX32 (as you can see [here](https://git.reactos.org/?p=reactos.git;a=blob;f=sdk/lib/fslib/vfatxlib/fatx.c;hb=8248f02ff46abbc5b9a1cec8703c88655cf1813a#l359)).

And now that I'm talking about it I also did a shitty mistake, SHELL32 mustn't care about FATX16/FATX32 differences. I'll refactor this PR as soon as I can.

dll/win32/shell32/dialogs/drive.cpp
```diff
@@ -81,6 +81,27 @@ GetDefaultClusterSize(LPWSTR szFs, PDWORD pClusterSize, PULARGE_INTEGER TotalNum
         else
             return FALSE;
     }
+    else if (!wcsicmp(szFs, L"FATX"))
+    {
+        if (TotalNumberOfBytes->QuadPart <= (16 * 1024 * 1024))
```
katahiromz, line 86: From what did you take these values?
Comment a URL.

GeoB99: The values are taken from both FAT and FAT32 code condition path cases, if you actually pay attention to the code above.

katahiromz: Ok.

katahiromz (approved):

dll/win32/shell32/dialogs/drive.cpp
```diff
@@ -252,6 +273,28 @@ InsertDefaultClusterSizeForFs(HWND hwndDlg, PFORMAT_DRIVE_CONTEXT pContext)
...
+            hDlgCtrl = GetDlgItem(hwndDlg, 28680);
+            SendMessageW(hDlgCtrl, CB_RESETCONTENT, 0, 0);
+            lIndex = SendMessageW(hDlgCtrl, CB_ADDSTRING, 0, (LPARAM)wszBuf);
+            if (lIndex != CB_ERR)
+                SendMessageW(hDlgCtrl, CB_SETITEMDATA, lIndex, (LPARAM)ClusterSize);
+            SendMessageW(hDlgCtrl, CB_SETCURSEL, 0, 0);
+        }
+
+        SendMessageW(GetDlgItem(hwndDlg, 28675), BM_SETCHECK, BST_UNCHECKED, 0);
+        EnableWindow(GetDlgItem(hwndDlg, 28675), FALSE);
```
HBelusca, line 296: A general remark for this file (when I see these): It seems these blocks are copied/pasted identically for each filesystem we can support in the GUI; for NTFS there seems to be a minor variation. How hard would it be to simplify all these to have a common code? (something for thought, not for this PR.)

GeoB99: I can write a small function helper that simplifies and avoids code duplication here and there in the code. I can do that for the next PR, not of a big deal.

PR #3400, opened 2020-12-26
[NTOS:CC] Implement CcRemapBcb and CcSetLogHandleForFile CORE-17409

oleg-dubinskiy:
## Purpose

Add implementations of `CcRemapBcb` and `CcSetLogHandleForFile` routines in our cache controller (CC), (partially) based on the ones from new CC.
Since in new CC implementations of `CcRemapBcb` and  `CcRepinBcb` are nearly identical, and in MSDN they are described similarly, I decided to do the same in our current CC too. And it worked for me. :slightly_smiling_face: Also the implementation of `CcSetLogHandleForFile` is similar to the one in new CC, but the difference is using the internal structures of the current CC, whose are appropriate to the ones from new CC.

JIRA issue: [CORE-17409](https://jira.reactos.org/browse/CORE-17409)

## TODO

- [ ] Handle failure case in `CcRemapBcb` if needed. This is probably also required, but currently is not implemented in new CC.

## Result

This allows NTFS driver from Windows XP and 2003 properly recognize (and work with) NTFS volumes in ReactOS. :smiley: (in pair with @ThFabba's patch from [CORE-10147](https://jira.reactos.org/browse/CORE-10147))

Before:
MS_ntfs_in_ReactOS_before

After:
MS_ntfs_in_ReactOS_wo_USB

Read/write support works as well.

### DIFF
--- a/ntoskrnl/cc/cacheman.c
+++ b/ntoskrnl/cc/cacheman.c
@@ -154,7 +154,7 @@ CcRemapBcb (
     IN PVOID Bcb
     )
 {
-    PINTERNAL_BCB iBcb = Bcb;
+    PINTERNAL_BCB iBcb = CONTAINING_RECORD(Bcb, INTERNAL_BCB, PFCB);
 
     CCTRACE(CC_API_DEBUG, "Bcb=%p\n", Bcb);
 
--- a/ntoskrnl/cc/cacheman.c
+++ b/ntoskrnl/cc/cacheman.c
@@ -137,8 +137,16 @@ CcGetFlushedValidData (
 	return i;
 }
 
-/*
- * @implemented
+/**
+ * @brief Remaps a buffer control block (BCB).
+ * 
+ * Maps data in a cache file for reading, similarly to CcMapData.
+ * This makes the data mapped, but not pinned, so it can't be modified.
+ * 
+ * @param [in] Bcb
+ * Pointer to the BCB.
+ * 
+ * @return Read-only BCB pointer.
  */
 PVOID
 NTAPI
--- a/ntoskrnl/cc/cacheman.c
+++ b/ntoskrnl/cc/cacheman.c
@@ -6,6 +6,7 @@
  *
  * PROGRAMMERS:     David Welch (welch@cwcom.net)
  *                  Pierre Schweitzer (pierre@reactos.org)
+ *                  Oleg Dubinskiy (oleg.dubinskij2013@yandex.ua)
  */
 
 /* INCLUDES *****************************************************************/
@@ -137,7 +138,7 @@ CcGetFlushedValidData (
 }
 
 /*
- * @unimplemented
+ * @implemented
  */
 PVOID
 NTAPI
@@ -145,9 +146,12 @@ CcRemapBcb (
     IN PVOID Bcb
     )
 {
-	UNIMPLEMENTED;
+    PINTERNAL_BCB iBcb = Bcb;
 
-    return 0;
+    CCTRACE(CC_API_DEBUG, "Bcb=%p\n", Bcb);
+
+    iBcb->RefCount++;
+    return Bcb;
 }
 
 /*
--- a/ntoskrnl/cc/fs.c
+++ b/ntoskrnl/cc/fs.c
@@ -346,8 +346,20 @@ CcSetFileSizes (
     KeReleaseSpinLock(&SharedCacheMap->CacheMapLock, oldirql);
 }
 
-/*
- * @implemented
+/**
+ * @brief Sets a log handle for file.
+ * 
+ * @param [in] FileObject
+ * Pointer to the file object of file for which will be stored a log handle.
+ * 
+ * @param [in] LogHandle
+ * Pointer to the log handle which will be stored.
+ * 
+ * @param [in] FlushToLsnRoutine
+ * Pointer to a flush callback routine of the log file, which should be called
+ * before flushing buffers of the file.
+ * 
+ * @return None
  */
 VOID
 NTAPI
--- a/ntoskrnl/cc/fs.c
+++ b/ntoskrnl/cc/fs.c
@@ -5,6 +5,7 @@
  * PURPOSE:         Implements cache managers functions useful for File Systems
  *
  * PROGRAMMERS:     Alex Ionescu
+ *                  Oleg Dubinskiy
  */
 
 /* INCLUDES ******************************************************************/
@@ -346,7 +347,7 @@ CcSetFileSizes (
 }
 
 /*
- * @unimplemented
+ * @implemented
  */
 VOID
 NTAPI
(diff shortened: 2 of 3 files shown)


ntoskrnl/cc/cacheman.c
```diff
@@ -137,17 +138,20 @@ CcGetFlushedValidData (
 }
 
 /*
- * @unimplemented
+ * @implemented
```
HBelusca, line 141: That's not really what I would call an "implementation" 🤣

ntoskrnl/cc/fs.c
```diff
@@ -346,7 +347,7 @@ CcSetFileSizes (
 }
 
 /*
- * @unimplemented
+ * @implemented
```
HBelusca, line 350: Same comment.

ntoskrnl/cc/cacheman.c
oleg-dubinskiy: So what's this then? :confused: I should keep `unimplemented` there?

binarymaster: Two options: either just remove this comment block, or (better) write proper SAL annotation for it, so it can be properly Doxyged'ed.

oleg-dubinskiy: Is there any example how to write SAL annotation properly? :thinking:

ntoskrnl/cc/cacheman.c
```diff
@@ -137,17 +138,20 @@ CcGetFlushedValidData (
...
     IN PVOID Bcb
     )
 {
-	UNIMPLEMENTED;
+    PINTERNAL_BCB iBcb = Bcb;
+
+    CCTRACE(CC_API_DEBUG, "Bcb=%p\n", Bcb);
 
-    return 0;
+    iBcb->RefCount++;
```
Extravert-ir, line 153: Every `RefCount` increase must have a corresponding `RefCount` decrease.
Where is the decrease for this one?

Looks like this stub can just be improved by returning `Bcb` instead of `NULL`

ntoskrnl/cc/fs.c
```diff
@@ -355,10 +356,15 @@ CcSetLogHandleForFile (
...
 {
+    PROS_SHARED_CACHE_MAP Map = FileObject->SectionObjectPointer->SharedCacheMap;
+
     CCTRACE(CC_API_DEBUG, "FileObject=%p LogHandle=%p FlushToLsnRoutine=%p\n",
         FileObject, LogHandle, FlushToLsnRoutine);
 
-    UNIMPLEMENTED;
+    if (!Map) return;
+
+    Map->LogHandle = LogHandle;
```
Extravert-ir, line 366: This seem to be correct, but does it make sense to add these variables to the structure, before we have a code which uses them?

Extravert-ir (changes requested):

ntoskrnl/cc/fs.c
oleg-dubinskiy: I think it does, because when we'll add that code, it will not needed to fix this implementation later.

ntoskrnl/cc/cacheman.c
oleg-dubinskiy: It probably should be done on failure case, which I didn't implement yet.

oleg-dubinskiy: No @Extravert-ir , just returning `Bcb` instead of `0` does not work for me. The system fails again for me when trying to write anything to NTFS partition. :frowning_face: The only difference is some asserts, whose don't appear if to return `0`.
And moreover, if to look at `CcRepinBcb` implementation (which is similar to `CcRemapBcb`), it also has only increase, but does not have any decrease. And it's not used by anything other (in the kernel) that has appropriate decrease, same as my one. Either this routine indeed shouldn't be used by any other ones, or those ones are just not implemented yet. But in any case, if it works at least for NTFS driver, then my implementation is definitely correct.
So your argument seems to be wrong in this case.

zefklop: The MSDN documentation is unclear about it : should a call to CcRemapBcb be paired with CcUnpinData ?

oleg-dubinskiy: I found the decrease one for `CcRepinBcb`. It is called many times it some our FS drivers, e. g. fastfat_new, for example here: https://git.reactos.org/?p=reactos.git;a=blob;f=drivers/filesystems/fastfat_new/allocsup.c;hb=8ee88d3bd5bb2464424490bcbca8510d2116df51#l4117, and each time, after that, it calls `CcUnpinData`, whose implementation returns `CcUnpinDataForThread`, which, in its turn, contains a decrease for reference count of `Bcb`: https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/cc/pin.c;hb=8ee88d3bd5bb2464424490bcbca8510d2116df51#l617.
It's probably done by the same way in MS NTFS driver too. First it calls `CcRemapBcb` somewhere, and then some appropriate routine for it, which decrements the reference count (probably `CcUnpinData` or other). :slightly_smiling_face: 
So looks like the decrease is done by another function (currently implemented or not), and it's not needed to add it into my implementation.
See https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/cc/pin.c;hb=8ee88d3bd5bb2464424490bcbca8510d2116df51#l155 also (called by `CcUnpinData`).

oleg-dubinskiy: > The MSDN documentation is unclear about it : should a call to CcRemapBcb be paired with CcUnpinData ?

Probably not, but there definitely should be some routine for decrease. Judging by MSDN documentation, and our code, it's `CcUnpinRepinnedBcb`: https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/cc/pin.c;hb=8ee88d3bd5bb2464424490bcbca8510d2116df51#l642.

ntoskrnl/cc/cacheman.c
oleg-dubinskiy: Done. :ok_hand:

ntoskrnl/cc/fs.c
oleg-dubinskiy: This too.

zefklop (changes requested):
Please add tests with the following scenarii:

- CcRemapBcb
  - Incoming Bcb comes from CcMapData
  - Incoming Bcb comes from CcPinRead
  - Calling twice CcMapData
  - Is the returned value different from the provided Bcb ? If yes, in which cases ?

Take an example of CcPinRead tests for reference.

zefklop (changes requested):
Please correct Pointer arithmetics

ntoskrnl/cc/cacheman.c
```diff
@@ -136,18 +137,29 @@ CcGetFlushedValidData (
...
+ * @return Read-only BCB pointer.
  */
 PVOID
 NTAPI
 CcRemapBcb (
     IN PVOID Bcb
     )
 {
-	UNIMPLEMENTED;
+    PINTERNAL_BCB iBcb = Bcb;
```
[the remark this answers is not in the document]
oleg-dubinskiy: Fixed. :ok_hand:

GeoB99:
I'm going to close this since this PR is been sitting here for quite a long time with not further input. Feel free however to poke me though in case you want to continue working on this.

PR #3401, opened 2020-12-27
[TASKMGR] Fix CORE-17115 by removing caching

powpingdone:
The command line caching system was faulty. Indexes do not persist the actual program across different programs. PID is also not a solution because it would eventually collide as well given enough opening and closing of processes. After timing the function, I've determined it to be fast enough to not require caching. 

[CORE-17115](https://jira.reactos.org/browse/CORE-17115)

### DIFF
--- a/base/applications/taskmgr/perfdata.c
+++ b/base/applications/taskmgr/perfdata.c
@@ -48,8 +48,6 @@ SYSTEM_HANDLE_INFORMATION                  SystemHandleInfo;
 PSYSTEM_PROCESSOR_PERFORMANCE_INFORMATION  SystemProcessorTimeInfo = NULL;
 PSID                                       SystemUserSid = NULL;
 
-PCMD_LINE_CACHE global_cache = NULL;
-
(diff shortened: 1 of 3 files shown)


powpingdone:
Should I add the copyright info to the top of each of these files i've edited?

HBelusca:
Copyright notices are for when you have done substantial changes in the concerned files. If it's just making one or two line changes it's not needed.

powpingdone:
Then I should only add it to perfdata.c?

powpingdone:
Nvm, no need.

SergeGautherie:
I would think having a cache is good and fixing it would be better than dropping it.

powpingdone:
The potential problem I see with fully fixing the caching system just over complicates the whole command line fetching system. 

When I previously "solved" this bug in #3016, I brought about racing issues in taskmgr. This eventually culminated in multiple other commits that ballooned into a much more complicated system than the previous broken cache system. I just don't think its worth over-engineering this.

SergeGautherie:
> over-engineering

I would guess caching can be somewhat redesigned as need be. I don't know if the result would over-engineered, hopefully not.

PR #3404, opened 2020-12-28
[UNIATA] Emulate SCSI INQUIRY to properly recognize Xbox DVD drives

binarymaster:
## Purpose

Emulate SCSI INQUIRY to properly recognize Xbox DVD drives and allow booting from them on a real OG Xbox.

JIRA issue: [CORE-16692](https://jira.reactos.org/browse/CORE-16692)

## Proposed changes

Explicitly check known drives by their names using ATA IDENTIFY PACKET fields. It should be possible to send INQUIRY anyway and then check for `SCSI_SENSE_ILLEGAL_REQUEST` error code, but I found it too hard to implement, so I'd let @Alter-1 implement this feature in upstream if he wants.

## Notes from Linux kernel source

These notes were found by @Extravert-ir while investigating this problem. Maybe this could be an argument for adding it officially to UniATA driver.
```
In Linux, if the DEVICE exists and the SCSI INQUIRY fails (e.g. because
       the SG_IO ioctl is not supported) then an ATA IDENTIFY (PACKET)  DEVICE
       is tried. If it succeeds then device identification strings are output.
       The --raw and --hex options can be used to manipulate the  output.   If
       the  --ata  option  is given then the SCSI INQUIRY is not performed and
       the DEVICE is assumed to be ATA (or ATAPI).
```

### DIFF
--- a/drivers/storage/ide/uniata/id_ata.cpp
+++ b/drivers/storage/ide/uniata/id_ata.cpp
@@ -9499,6 +9499,82 @@ reject_srb:
                    (Srb->Cdb[0] != SCSIOP_ATA_PASSTHROUGH)/* &&
                    (Srb->Cdb[0] != SCSIOP_REPORT_LUNS)*/) {
                     KdPrint3((PRINT_PREFIX "Try ATAPI send %x\n", Srb->Cdb[0]));
+#ifdef __REACTOS__
+                    status = SRB_STATUS_BUSY;
+
+                    if (Srb->Cdb[0] == SCSIOP_INQUIRY &&
+                        (LunExt->DeviceFlags & DFLAGS_ATAPI_DEVICE) &&
+                        (LunExt->IdentifyData.DeviceType == ATAPI_TYPE_CDROM ||
+                        LunExt->IdentifyData.DeviceType == ATAPI_TYPE_OPTICAL) &&
+                        LunExt->IdentifyData.ModelNumber[0])
+                    {
+                        ULONG j;
+                        CCHAR vendorId[26];
+
+                        // Attempt to identify known broken CD/DVD drives
+                        for (j = 0; j < sizeof(vendorId); j += 2)
+                        {
+                            // Build a buffer based on the identify data.
+                            MOV_DW_SWP(vendorId[j], ((PUCHAR)LunExt->IdentifyData.ModelNumber)[j]);
+                        }
+
+                        // Emulate INQUIRY support for broken CD/DVD drives (e.g. Microsoft Xbox).
+                        // Currently we implement it by explicitly checking the drive name from ATA IDENTIFY PACKET.
+                        if (!AtapiStringCmp(vendorId, "THOMSON-DVD", 11) ||
+                            !AtapiStringCmp(vendorId, "PHILIPS XBOX DVD DRIVE", 22) ||
+                            !AtapiStringCmp(vendorId, "PHILIPS J5 3235C", 16) ||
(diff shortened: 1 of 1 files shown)


HBelusca:
Cc @Alter-1 too :)

HBelusca (commented):
Should be fine.... I suppose 🤣 (ahh, these kilometer-long functions 😂 😂 )

drivers/storage/ide/uniata/id_ata.cpp
```diff
@@ -9499,6 +9499,82 @@ AtapiStartIo__(
...
+                            !AtapiStringCmp(vendorId, "PHILIPS XBOX DVD DRIVE", 22) ||
+                            !AtapiStringCmp(vendorId, "PHILIPS J5 3235C", 16) ||
+                            !AtapiStringCmp(vendorId, "SAMSUNG DVD-ROM SDG-605B", 24))
+                        {
+                            // TODO:
+                            // Better send INQUIRY and then check for chan->ReturningMediaStatus >> 4 == SCSI_SENSE_ILLEGAL_REQUEST
+                            // in AtapiInterrupt__() and emulate the response only in this case.
+
+                            // If this hack stays for long enough, consider adding Xbox 360 drive names to the condition,
+                            // as they are affected by the same problem.
```
zefklop, line 9533: I don't get the point about not adding controller known to be flaky to the list.

drivers/storage/ide/uniata/id_ata.cpp
```diff
@@ -9499,6 +9499,82 @@ AtapiStartIo__(
...
+                    {
+                        ULONG j;
+                        CCHAR vendorId[26];
+
+                        // Attempt to identify known broken CD/DVD drives
+                        for (j = 0; j < sizeof(vendorId); j += 2)
+                        {
+                            // Build a buffer based on the identify data.
+                            MOV_DW_SWP(vendorId[j], ((PUCHAR)LunExt->IdentifyData.ModelNumber)[j]);
+                        }
```
zefklop, line 9519: What a strangely packed data...

Given how often you have to perform this, can you check whether uniata has a helper function here ? If not, nevermind, don't be too intrusive.

zefklop (commented):
> ```
> In Linux, if the DEVICE exists and the SCSI INQUIRY fails (e.g. because
>        the SG_IO ioctl is not supported) then an ATA IDENTIFY (PACKET)  DEVICE
>        is tried. If it succeeds then device identification strings are output.
>        The --raw and --hex options can be used to manipulate the  output.   If
>        the  --ata  option  is given then the SCSI INQUIRY is not performed and
>        the DEVICE is assumed to be ATA (or ATAPI).
> ```

Can you provide a link to this, please ?

drivers/storage/ide/uniata/id_ata.cpp
binarymaster: Right, but I don't know their exact names. OG Xbox DVD drives on the other hand are very well documented: https://xboxdevwiki.net/Porting_an_Operating_System_to_the_Xbox_HOWTO#DVD_Drive_Bugs

drivers/storage/ide/uniata/id_ata.cpp
binarymaster: Yeah uniata don't have a helper function for this. The similar constructs are used in different places, and this one is basically copypasted :)

binarymaster (commented):
> Can you provide a link to this, please ?

Sure: https://linux.die.net/man/8/sg_inq

zefklop (approved):
Fine for me

PR #3412, opened 2021-01-04
flpydisk [pl-pl] update

pithwz:
flpydisk [pl-pl] update

### DIFF
--- a/media/inf/flpydisk.inf
+++ b/media/inf/flpydisk.inf
@@ -82,7 +82,7 @@ Floppy.DeviceDesc = "Disquetera genérica"
 ReactOS = "Zespół ReactOS"
 
 FloppyClassName = "Stacje dyskietek"
-GenericMfg = "(Standardowe stacje dyskietek)"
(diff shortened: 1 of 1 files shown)


media/inf/flpydisk.inf
```diff
@@ -82,7 +82,7 @@ Floppy.DeviceDesc = "Disquetera genérica"
 ReactOS = "Zespół ReactOS"
 
 FloppyClassName = "Stacje dyskietek"
-GenericMfg = "(Standardowe stacje dyskietek)"
+GenericMfg = "(Standardowe sterowniki stacji dyskietek)"
```
binarymaster, line 85: Original translation is correct. See english translation for reference:

> (Generic floppy disk drives)

**Drives**, not drivers.

pithwz: its description from WinNT 5.1+ :)

binarymaster: We are not copying translations from Windows as is, because it's copyrighted material.

PR #3414, opened 2021-01-04
[NTOS:MM] Stubplement cookie generation for drivers

learn-more:
Drivers compiled with a newer SDK check that their cookie is non-default.
This is not the same algorithm as Windows.

### DIFF
--- a/ntoskrnl/mm/ARM3/sysldr.c
+++ b/ntoskrnl/mm/ARM3/sysldr.c
@@ -50,6 +50,12 @@ PMMPTE MiKernelResourceStartPte, MiKernelResourceEndPte;
 ULONG_PTR ExPoolCodeStart, ExPoolCodeEnd, MmPoolCodeStart, MmPoolCodeEnd;
 ULONG_PTR MmPteCodeStart, MmPteCodeEnd;
 
+#ifdef _WIN64
+#define DEFAULT_SECURITY_COOKIE 0x00002B992DDFA232ll
+#else
+#define DEFAULT_SECURITY_COOKIE 0xBB40E64E
+#endif
+
 /* FUNCTIONS ******************************************************************/
 
 PVOID
@@ -2807,6 +2813,84 @@ Fail:
     return Status;
 }
 
+
+PVOID
+NTAPI
+LdrpFetchAddressOfSecurityCookie(PVOID BaseAddress, ULONG SizeOfImage)
+{
+    PIMAGE_LOAD_CONFIG_DIRECTORY ConfigDir;
+    ULONG DirSize;
+    PVOID Cookie = NULL;
+
+    /* Check NT header first */
+    if (!RtlImageNtHeader(BaseAddress)) return NULL;
+
+    /* Get the pointer to the config directory */
+    ConfigDir = RtlImageDirectoryEntryToData(BaseAddress,
+                                             TRUE,
+                                             IMAGE_DIRECTORY_ENTRY_LOAD_CONFIG,
+                                             &DirSize);
+
+    /* Check for sanity */
+    if (!ConfigDir ||
+        DirSize < FIELD_OFFSET(IMAGE_LOAD_CONFIG_DIRECTORY, SEHandlerTable) ||  /* SEHandlerTable is after SecurityCookie */
+        (ConfigDir->Size != DirSize))
+    {
+        /* Invalid directory*/
+        return NULL;
+    }
+
+    /* Now get the cookie */
+    Cookie = (PVOID)ConfigDir->SecurityCookie;
+
+    /* Check this cookie */
+    if ((PCHAR)Cookie <= (PCHAR)BaseAddress ||
+        (PCHAR)Cookie >= (PCHAR)BaseAddress + SizeOfImage)
+    {
+        Cookie = NULL;
+    }
+
+    /* Return validated security cookie */
+    return Cookie;
+}
+
+PVOID
+NTAPI
+LdrpInitSecurityCookie(PLDR_DATA_TABLE_ENTRY LdrEntry)
+{
+    PULONG_PTR Cookie;
+    ULONG_PTR NewCookie;
+
+    /* Fetch address of the cookie */
+    Cookie = LdrpFetchAddressOfSecurityCookie(LdrEntry->DllBase, LdrEntry->SizeOfImage);
+
+    if (Cookie)
+    {
+        /* Check if it's a default one */
+        if ((*Cookie == DEFAULT_SECURITY_COOKIE) ||
+            (*Cookie == 0))
+        {
+            LARGE_INTEGER Counter = KeQueryPerformanceCounter(NULL);
+            /* The address should be unique */
+            NewCookie = (ULONG_PTR)Cookie;
+
+            /* We just need a simple tick, don't care about precision and whatnot */
+            NewCookie ^= (ULONG_PTR)Counter.LowPart;
+
+            /* If the result is 0 or the same as we got, just add one to the default value */
+            if ((NewCookie == 0) || (NewCookie == *Cookie))
+            {
+                NewCookie = DEFAULT_SECURITY_COOKIE + 1;
+            }
+
+            /* Set the new cookie value */
+            *Cookie = NewCookie;
+        }
+    }
+
+    return Cookie;
+}
+
 NTSTATUS
 NTAPI
 MmLoadSystemImage(IN PUNICODE_STRING FileName,
@@ -3251,6 +3335,9 @@ LoaderScan:
 
     /* Write-protect the system image */
     MiWriteProtectSystemImage(LdrEntry->DllBase);
+
+    /* Initialize the security cookie (Win7 is not doing this yet!) */
+    LdrpInitSecurityCookie(LdrEntry);
 
     /* Check if notifications are enabled */
     if (PsImageNotifyEnabled)


ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -2814,6 +2820,80 @@ MmCheckSystemImage(IN HANDLE ImageHandle,
...
+PVOID
+NTAPI
+LdrpFetchAddressOfSecurityCookie(PVOID BaseAddress, ULONG SizeOfImage)
+{
+    PIMAGE_LOAD_CONFIG_DIRECTORY ConfigDir;
+    ULONG DirSize;
+    PVOID Cookie = NULL;
+
+    /* Check NT header first */
+    if (!RtlImageNtHeader(BaseAddress)) return NULL;
```
HBelusca, line 2833:
```suggestion
    if (!RtlImageNtHeader(BaseAddress))
        return NULL;
```

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -2814,6 +2820,80 @@ MmCheckSystemImage(IN HANDLE ImageHandle,
...
+
+    /* Get the pointer to the config directory */
+    ConfigDir = RtlImageDirectoryEntryToData(BaseAddress,
+        TRUE,
+        IMAGE_DIRECTORY_ENTRY_LOAD_CONFIG,
+        &DirSize);
```
HBelusca, line 2839:
```suggestion
    ConfigDir = RtlImageDirectoryEntryToData(BaseAddress,
                                             TRUE,
                                             IMAGE_DIRECTORY_ENTRY_LOAD_CONFIG,
                                             &DirSize);
```

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -2814,6 +2820,80 @@ MmCheckSystemImage(IN HANDLE ImageHandle,
...
+
+    /* Check for sanity */
+    if (!ConfigDir ||
+        (DirSize != 64 && ConfigDir->Size != DirSize) ||
+        (ConfigDir->Size < 0x48))
+        return NULL;
```
HBelusca, line 2845:
```suggestion
    if (!ConfigDir ||
        (DirSize != 64 && ConfigDir->Size != DirSize) ||
        (ConfigDir->Size < 0x48))
    {
        return NULL;
    }
```

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -2814,6 +2820,80 @@ MmCheckSystemImage(IN HANDLE ImageHandle,
...
+NTAPI
+LdrpInitSecurityCookie(PLDR_DATA_TABLE_ENTRY LdrEntry)
+{
+    PULONG_PTR Cookie;
+    ULONG_PTR NewCookie;
+
+    /* Fetch address of the cookie */
+    Cookie = LdrpFetchAddressOfSecurityCookie(LdrEntry->DllBase, LdrEntry->SizeOfImage);
+
+    if (Cookie)
```
HBelusca, line 2871:
```suggestion
    if (!Cookie)
        return NULL;
```
may be done similarly for the other if, and you gain one or two levels of indentation for the main code block below.

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -3259,6 +3339,9 @@ MmLoadSystemImage(IN PUNICODE_STRING FileName,
     /* Write-protect the system image */
     MiWriteProtectSystemImage(LdrEntry->DllBase);
 
+    /* Initialize the security cookie (Win7 is not doing this yet!) */
```
HBelusca, line 3342: You mean, it does it, but later?

ntoskrnl/mm/ARM3/sysldr.c
HBelusca: Also, 0x48 == 72== `sizeof(IMAGE_LOAD_CONFIG_DIRECTORY)` for its 32-bit version (this is by the way platform specific, with `IMAGE_LOAD_CONFIG_DIRECTORY32` for 32-bits and `IMAGE_LOAD_CONFIG_DIRECTORY64` for 64 bits).

HBelusca: And 64 == `FIELD_OFFSET(IMAGE_LOAD_CONFIG_DIRECTORY, SecurityCookie)`

HBelusca: See also https://git.reactos.org/?p=reactos.git;a=blob;f=modules/rostests/apitests/loadconfig/common.c;hb=68182b2a52858eb0011df1467444e66e0b056f74#l47 that does a similar thing.
And these magic values should also be fixed in this function: https://git.reactos.org/?p=reactos.git;a=blob;f=dll/ntdll/ldr/ldrinit.c;hb=68182b2a52858eb0011df1467444e66e0b056f74#l410

ntoskrnl/mm/ARM3/sysldr.c
learn-more: Win10 is doing this, didnt check win8

Extravert-ir:
Would be cool to enable security cookie in classpnp
https://github.com/reactos/reactos/blob/master/drivers/storage/class/classpnp/class.c#L231-L236

learn-more:
> Nitpick for commit message module:
> 
> ```diff
> -[NTOS]
> +[NTOS:MM]
> ```

I think there is already enough nitpicking in this PR, guess it's better to make a huge pr because that will receive no comments.

ntoskrnl/mm/ARM3/sysldr.c
learn-more: No, following the style of the surrounding code.

ColinFinck (approved):
LGTM. Just two remarks:
* This commit introduces architecture-specific `__rdtsc` in a non-architecture-specific file. The official alternative seems to be `NtGetTickCount`. However, the corresponding `KiGetTickCountHandler` trap handler currently always returns zero as the tick count, so this function isn't really an alternative right now.
* This commit also introduces `Ldrp` prefixes in an MM source file. I see that you copied most of the code from ntdll ldr, which is fine. Just not sure if we better rename the functions then to match the destination `Mi` prefix or keep things as is.

HBelusca:
> LGTM. Just two remarks:
> 
>     * This commit introduces architecture-specific `__rdtsc` in a non-architecture-specific file. The official alternative seems to be `NtGetTickCount`. However, the corresponding `KiGetTickCountHandler` trap handler currently always returns zero as the tick count, so this function isn't really an alternative right now.
> 
>     * This commit also introduces `Ldrp` prefixes in an MM source file. I see that you copied most of the code from ntdll ldr, which is fine. Just not sure if we better rename the functions then to match the destination `Mi` prefix or keep things as is.

The corresponding user-mode Ldrp.... function does this:
https://git.reactos.org/?p=reactos.git;a=blob;f=dll/ntdll/ldr/ldrinit.c;hb=fd8080b0942e88502ca1735a0590f31e6d072e32#l464
i.e. calls NtQueryPerformanceCounter().
Since here we are in kernel-mode, and we can avoid probing etc.... we may as well call KeQueryPerformanceCounter() which does the correct job and abstract out the CPU-specific rdtsc .

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -2814,6 +2820,83 @@ MmCheckSystemImage(IN HANDLE ImageHandle,
...
+    {
+        /* Check if it's a default one */
+        if ((*Cookie == DEFAULT_SECURITY_COOKIE) ||
+            (*Cookie == 0))
+        {
+            /* The address should be unique */
+            NewCookie = (ULONG_PTR)Cookie;
+
+            /* We just need a simple tick, don't care about precision and whatnot */
+            NewCookie ^= (ULONG_PTR)__rdtsc();
```
ColinFinck, line 2884:
```suggestion
            NewCookie ^= (ULONG_PTR)KeQueryPerformanceCounter(NULL);
```

One-click fix based on @HBelusca's suggestion.

learn-more: I looked at that function, but wasnt sure about all the stuff it was doing...

zefklop:
Can we move forward with this ? I see no reason to hold on.

learn-more:
> Can we move forward with this ? I see no reason to hold on.

I did not re-test with the suggested change from colin.
If you want, feel free to apply that fix and re-test this.

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -2814,6 +2820,80 @@ MmCheckSystemImage(IN HANDLE ImageHandle,
...
+
+            /* If the result is 0 or the same as we got, just add one to the default value */
+            if ((NewCookie == 0) || (NewCookie == *Cookie))
+            {
+                NewCookie = DEFAULT_SECURITY_COOKIE + 1;
```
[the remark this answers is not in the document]
ColinFinck: This situation is unlikely enough to not warrant a restart with `(ULONG_PTR)Cookie + 1` in a loop that could (theoretically) run infinitely.
Let's leave this part as is.

ColinFinck:
> Can we move forward with this ? I see no reason to hold on.

Let's replace `__rdtsc` by `KeQueryPerformanceCounter` as suggested, try it, and if everything works, this PR should be good to go.

ntoskrnl/mm/ARM3/sysldr.c
learn-more: `KeQueryPerformanceCounter` returns a `LARGE_INTEGER` so the one-click won't compile.

PR #3415, opened 2021-01-05
Sync D3DX9 modules to wine 6.0-rc5

zefklop:
## Purpose

Sync d3dx9 modules to wine 6.0-rc5

## Proposed changes

This is the last step to make d3dx9 independent of wined3d, making it easier to sync d3d9, d3d8, ddraw & wined3d as a next step.


Zero3K:
It works for me. All graphics display properly in Melty Blood: Actress Again - Current Code.

PR #3416, opened 2021-01-05
[BOOTSECT][FREELDR] Support booting from NTFS partitions

sdever:
## Purpose

Implement NTFS boot sector that loads FreeLdr from a NTFS partition

Jira issue: [CORE-17474](https://jira.reactos.org/browse/CORE-17474)

## Proposed changes

 - Implement VBR for NTFS
 - Fix a bug in FreeLdr's NTFS driver preventing it to find freeldr.inf
 
## TODO

- Handle "weird" NTFS partition with uncommon sector sizes / sectors per cluster / sectors per index record
- Better implementation for file loading
- 64-bit LCN support
- More testing


JoachimHenze:
Great! How can we test that?

sdever:
> Great! How can we test that?

I manually copied the ntfs.bin into the partition (with the command `dd if=boot/freeldr/bootsect/ntfs.bin of=/dev/loop0p1 bs=1 skip=84 seek=84` on Linux), and copied the ReactOS bootcd and freeldr.sys on the mounted NTFS partition

I didn't change usetup yet to install the VBR on NTFS partition, as someone else said he worked on NTFS formatting/install on Mattermost.

boot/freeldr/bootsect/ntfs.S
```diff
@@ -0,0 +1,804 @@
...
+        pushad
+        call  ReadNonResidentAttribute
+        popad
+        mov   bx, fs
+        add   bx, HEX(100)
+        mov   fs, bx
+        xor   bx, bx
+        inc   ecx
+        cmp   ecx, edx
+        jbe   FreeLdrLoad
```
HBelusca, line 389: Is the extra level of indentation on purpose?

boot/freeldr/bootsect/ntfs.S
```diff
@@ -0,0 +1,804 @@
...
+BiosCHSDriveSize:
+    .long 0
+
+.org 509 // Pad to 509 bytes
+BootPartition:
+    .byte 1
+
+BootFlagSignature:
+    .word HEX(0aa55)    // BootSector signature
+
```
HBelusca, line 304: Please add an information block along the lines of what is done in the FAT32 bootsector:
https://git.reactos.org/?p=reactos.git;a=blob;f=boot/freeldr/bootsect/fat32.S;h=ff9c40af1211a7f7112b0a74d2bc7cdd876a8a0d;hb=HEAD#l317

Also, please check whether on Windows the NTFS boot sector is stored on disk contiguously or not. If not, please also add a comment similar to the one written in the FAT32 bootsector, concerning the second part that is (in FAT32 case) at sector 14.

boot/freeldr/bootsect/ntfs.S
```diff
@@ -0,0 +1,804 @@
...
+PrintFreeldrError:
+    mov  si, offset msgFreeldr
+    call PutChars
+    jmp  Reboot
+
+FreeLdr:
+    .word 'F', 'R', 'E', 'E', 'L', 'D', 'R', '.', 'S', 'Y', 'S'
+msgFreeldr:
+    .ascii "FreeLdr not found, cannot boot", CR, LF, NUL
+
```
HBelusca, line 801: Maybe add and use a "Loading FreeLdr..." message as we do with our other boot sectors, see example here:
https://git.reactos.org/?p=reactos.git;a=blob;f=boot/freeldr/bootsect/fat32.S;h=ff9c40af1211a7f7112b0a74d2bc7cdd876a8a0d;hb=HEAD#l530

boot/freeldr/bootsect/ntfs.S
```diff
@@ -0,0 +1,804 @@
...
+SectorsPerFat:
+    .word 0             // Unused by NTFS
+SectorsPerTrack:
+    .word 0             // Should contain disk geometry
+NumberOfHeads:
+    .word 0             // Should contain disk geometry
+HiddenSectors:
+    .long 0
+TotalSectorsBig:
+    .long 0
```
HBelusca, line 48: Unused by NTFS ?

HBelusca (commented):
I know that Windows' NTFS bootsector code complains if the boot loader is on a compressed NTFS volume. Do you intend to implement such check (and complain in this situation)?

boot/freeldr/bootsect/ntfs.S
sdever: Those values are used by NTFS, but can't be computed at compile time (since it depends of partition scheme). The labels are here only to get correct addresses to access those values in the code.

sdever:
> I know that Windows' NTFS bootsector code complains if the boot loader is on a compressed NTFS volume. Do you intend to implement such check (and complain in this situation)?

This is on my todo-list yes (I forgot to add it in PR todo list)

boot/freeldr/bootsect/ntfs.S
sdever: Boot sector is stored on disk contiguously on my Windows partitions: it's represented by the $Boot file (inode number 7), and uses the two first clusters in the first partition as marked in the run lists. Also, linux-ntfs documentation states this file cannot be relocated: https://flatcap.org/linux-ntfs/ntfs/files/boot.html.

I'll add what I wrote here as comment for the boot sector

boot/freeldr/bootsect/ntfs.S
HBelusca: My comment was only about "TotalSectorsBig"; according to documentation (e.g. Starman's MBR/VBR stuff, and another website) the TotalSectorsBig doesn't seem to be used by NTFS.

sdever: Yes, TotalSectorsBig is set to 0 on NTFS partitions formatted by Windows (and I was misled by Github display of PRs :p)

boot/freeldr/bootsect/ntfs.S
```diff
@@ -0,0 +1,829 @@
...
+BootDrive:
+    .byte HEX(80)
+CurrentHead:
+    .byte 0
+BootSignature:
+    .byte HEX(80)
+Unused:
+    .byte 0
+VolumeSectorCount:
+    .long 0
```
Extravert-ir, line 59: There is `.quad` macro for 64-bit numbers

boot/freeldr/bootsect/ntfs.S
```diff
@@ -810,7 +822,8 @@ msgFreeldr:
     .ascii "FreeLdr not found, cannot boot", CR, LF, NUL
 msgLoading:
     .ascii "Loading FreeLoader...", CR, LF, NUL
-
+msgFreeldrCompressed:
+    .ascii "freeldr.sys is a sparse file, compressed or encrypted, cannot boot", CR, LF, NUL
```
mzji, line 826: `freeldr.sys is a sparse, compressed or encrypted file, cannot boot` ?

boot/freeldr/bootsect/ntfs.S
```diff
@@ -302,6 +301,12 @@ BootPartition:
 BootFlagSignature:
     .word HEX(0aa55)    // BootSector signature
 
+// End of Bootsector
+// Next sector starts as sector 2, since boot sector is 
+// as a file under NTFS ($Boot, which has inode number 7)
+// and takes the two first clusters of the volume (which means)
+// 8 sectors (checked on Windows and mkfs.ntfs from ntfsutils)
+
```
mzji, line 309: IIRC cluster size can vary between different NTFS instances? Thus two clusters don't really means 8 sectors in all cases, right? Or only NTFS instances those have 1 cluster = 4 sectors can be used for boot? BTW 1 sector on PC HDD usually is 512 bytes in the old days, and a NTFS cluster by default is 4KiB, that means a NTFS cluster by default equals 8 sectors, not 4 sectors.

sdever: For the moment, it only boots on sectors with 512 bytes, as it's the most common NTFS settings, see the TODO list in the pull request.
Also I messed up with my calculations, it's 16 sectors for two clusters, thanks for pointing it :)

boot/freeldr/bootsect/ntfs.S
```diff
@@ -0,0 +1,829 @@
...
+    mov   cl, byte ptr [SectsPerCluster]
+    movzx ecx, cx
+    mul   ecx
+    mov   dword ptr MFTStartSector, eax
+
+    // Compute size of MFT entry in sectors
+    mov   al, byte ptr [ClustersPerMftRecord]
+    movzx ax, al
+    test  al, HEX(80)
+    jnz   NegativeOffset
```
Extravert-ir, line 326: I would put a comment here, this for example https://flatcap.org/linux-ntfs/ntfs/files/boot.html

> This can be negative, which means that the size of the MFT/Index record is smaller than a cluster. In this case the size of the MFT/Index record in bytes is equal to 2^(-1 * Clusters per MFT/Index record). So for example if Clusters per MFT Record is 0xF6 (-10 in decimal), the MFT record size is 2^(-1 * -10) = 2^10 = 1024 bytes.

Just a nit, this feels more straightforward:
```
test al, al
js NegativeOffset
```

boot/freeldr/bootsect/ntfs.S
```diff
@@ -0,0 +1,829 @@
...
+    mov   cx, word ptr [SectsPerCluster]
+    mul   cx
+    mov   word ptr [SectsPerMFT], cx
+    jmp   SearchNext
+    NegativeOffset:
+    not   al
+    mov   cl, al
+    sub   cl, 8
```
Extravert-ir, line 334: what is `8` here?

boot/freeldr/bootsect/ntfs.S
```diff
@@ -0,0 +1,829 @@
...
+    push  si
+    push  di
+    sub   sp, HEX(10)
+    mov   word ptr [bp - fileRecordBase], es
+    mov   word ptr [bp - fileRecord], ax
+
+    mov   ebx, ATTRIBUTE_INDEX_ROOT
+    call  FindAttributeHdr
+    test  ax, ax
+    jz    near ptr PrintFileSystemError         // fail if no INDEX_ROOT
```
Extravert-ir, line 438: The `near` can be omitted?

boot/freeldr/bootsect/ntfs.S
```diff
@@ -0,0 +1,829 @@
...
+    jnz   near ptr PrintFileSystemError         // fail if attribute is non-resident
+
+    add   si, word ptr es:[si + HEX(14)]
+    cmp   dword ptr es:[si], HEX(30)
+    jnz   PrintFileSystemError                  // fail if FILE_NAME attribute isn't indexed
+
+    mov   word ptr [bp - indexRootData], si
+    mov   ax, word ptr [bp - fileRecord]
+
+    test  byte ptr es:[si + 12], 1
```
Extravert-ir, line 451: Hex and Dec numbers are mixed

boot/freeldr/bootsect/ntfs.S
```diff
@@ -0,0 +1,829 @@
...
+    pop   di
+    pop   si
+    mov   sp, bp
+    pop   bp
+    ret
+
+// [ES:SI] - the source file name
+// [DS:DI] - the destination file name
+// CX - compare length
+CompareStr:
```
Extravert-ir, line 536: Case-insensitive compare

boot/freeldr/bootsect/ntfs.S
```diff
@@ -0,0 +1,829 @@
...
+    pop bp
+    ret
+
+// Reads a NTFS cluster
+// AX     : cluster number to read
+// ES:BX  : address to read
+ReadCluster:
+    push  edx
+    push  ecx
+    mov   cl, byte ptr [SectsPerCluster]   // Convert sectors to clusters
```
Extravert-ir, line 574: Maybe just store a multiplied number too?

boot/freeldr/bootsect/ntfs.S
```diff
@@ -0,0 +1,829 @@
...
+    movzx ecx, cx
+    mul   ecx
+    add   eax, dword ptr [HiddenSectors]   // Add offset from the disk beginning
+    call  ReadSectors
+    pop   ecx
+    pop   edx
+    ret
+
+// Reads a MFT entry
+// AX     : MFT inode
```
Extravert-ir, line 584: Just make it `eax`, that way you can avoid `movzx eax, ax` stuff (it is already called like that actually)

boot/freeldr/bootsect/ntfs.S
```diff
@@ -0,0 +1,829 @@
...
+    push  edx
+    push  ecx
+    push  si
+    push  di
+
+    mov   cx, word ptr [SectsPerMFT]       // Get the correct number of sectors for the FILE entry
+    mul   cx
+    movzx eax, ax
+    mov   ecx, dword ptr [MFTStartSector]  // Add it to the start of the MFT
+    add   eax, ecx
```
Extravert-ir, line 596:
```suggestion
    add   eax, dword ptr [MFTStartSector]  // Add it to the start of the MFT
```

Extravert-ir: And `cx` will not be touched

boot/freeldr/bootsect/ntfs.S
```diff
@@ -0,0 +1,829 @@
...
+    add   eax, dword ptr [HiddenSectors]   // Add disk offset
+    mov   cx, word ptr [SectsPerMFT]
+    movzx ecx, cx
+    call  ReadSectors
+    
+    cmp   dword ptr es:[bx], HEX(454c4946) // Ensure we get a valid FILE record
```
Extravert-ir, line 602: I don't know NTFS that well. What is this?

boot/freeldr/bootsect/ntfs.S
```diff
@@ -0,0 +1,829 @@
...
+// Returns: ebx: the unpacked number
+UnpackLen:
+    push  cx
+    movzx ax, al
+
+    // Read the whole DWORD and then compute a mask to remove
+    // unneeded bytes to get correct size
+    xor   ebx, ebx
+    mov   ebx, dword ptr es:[si]
+    add   si, ax
```
Extravert-ir, line 750: This is a return value as well

boot/freeldr/bootsect/ntfs.S
```diff
@@ -0,0 +1,829 @@
...
+        jmp   FixupLoop
+    EndFixupLoop:
+
+    pop di
+    pop si
+    ret
+
+// Reads a non-resident attribute
+// ES:AX : Address of the data runs
+// FS:BX : Address to write to
```
Extravert-ir, line 652: This looks a bit unusual. Why not `DS:AX` and `ES:BX` (to fit the `ReadCluster`)?

boot/freeldr/bootsect/ntfs.S
```diff
@@ -0,0 +1,829 @@
...
+// ECX    : LCN to read
+ReadNonResidentAttribute:
+    push  bp
+    mov   bp, sp
+    sub   sp, HEX(10)
+    push  dx
+    push  si
+    mov   dword ptr [bp - 4], 0          // Store the current LCN
```
Extravert-ir, line 660: Wasting bytes :)

Extravert-ir (dismissed):
Good job!
I didn't dive that much into NTFS-specific stuff here.
Please, answer the comments and format functions like I did in the BTRFS VBR:
```
// Searches a key in a BTRFS tree
// INPUT:
// - [bp+4] BTRFS key to find
// - EDX:EAX logical address of root header
// - CL leaf node level
// OUTPUT:
// - ES:[SI] pointer to found item's key
// - ES:[BX] pointer to found item
// - ECX item length
// LOCALS:
// - [bp-8] - block number/current node offset
// - [bp-9] - current level
SearchTree:
```

boot/freeldr/bootsect/ntfs.S
HBelusca: It's `ELIF`  (aka. `FILE` written in little endian).

boot/freeldr/bootsect/ntfs.S
```diff
@@ -0,0 +1,829 @@
...
+
+#include <asm.inc>
+#include <freeldr/include/arch/pc/x86common.h>
+
+#define BP_REL(x) [bp+x-offset start]
+
+.code16
+
+//ORG HEX(7c00)
+
```
HBelusca, line 19: (Regarding a previous comment by @Extravert-ir ): Yes the publicly available documentation on the format of the NTFS boot sector shows that all these fields are indeed part of the NTFS BPB.

boot/freeldr/bootsect/ntfs.S
```diff
@@ -0,0 +1,829 @@
...
+#define ATTRIBUTE_INDEX_ROOT HEX(90)
+#define ATTRIBUTE_INDEX_ALLOCATION HEX(A0)
+
+StartSearch:
+    // Compute MFT start sector
+    mov   eax, dword ptr [MftLocation]
+    mov   cl, byte ptr [SectsPerCluster]
+    movzx ecx, cx
+    mul   ecx
+    mov   dword ptr MFTStartSector, eax
```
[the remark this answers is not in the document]
sdever: I should learn how to perform 64-bit calculations in real-mode assembly before  (which is quite a PITA compared to protected mode), so I'll handle 64 bit in a future iteration

boot/freeldr/bootsect/ntfs.S
sdever: MSVC build whines if not present

boot/freeldr/bootsect/ntfs.S
sdever: Because I didn't want to touch DS, since it was useful to print debug messages on that part

boot/freeldr/bootsect/ntfs.S
sdever: I should make a define here yes

Zero3K:
I tried it on my 2006 Macbook. I get a ? Folder icon instead of it booting to ReactOS.

Zero3K:
I have an issue with booting NTFS on a 2006 Macbook after applying this and the other two patches needed to get ReactOS to use MS' NTFS driver. I wrote the bootsector via dd under Slax and copied over the ReactOS directory and boot files from the USB that I copied them to (from a VHD that I installed the modified ReactOS to. When booting via the SSD after doing that, Loading FreeLdr... keeps being spammed.

boot/freeldr/bootsect/ntfs.S
HBelusca: This would then be our first asm code with near ptr.
Could you see whether by removing `near ptr` it works then? (that's how all of our other code does it; alternatively it uses `short`)

boot/freeldr/bootsect/ntfs.S
aaSSfxxx: We can't, since we don't know what cluster we'll need to read

boot/freeldr/bootsect/ntfs.S
sdever: For some reason MSVC stopped whining on rhis and compiles fine without.

boot/freeldr/bootsect/ntfs.S
sdever: ReadCluster takes an arbitrary cluster number, and it needs to be converted to sectors then, so it can't be precomputed.

boot/freeldr/bootsect/ntfs.S
HBelusca: Maybe it's part of their new set of bugs they "introduced" in MASM when it compiles 16-bit code?
See another instance of this there: https://developercommunity.visualstudio.com/t/MASMML-x86:-Inconsistent-large-mov/1385890

boot/freeldr/bootsect/ntfs.S
```diff
@@ -0,0 +1,829 @@
...
+PrintDiskError:
+    mov  si, offset msgDiskError        // Bad boot disk message
+    call PutChars                       // Display it
+
+    jmp  Reboot
+
+// Displays a file system error message
+// And reboots
+PrintFileSystemError:
+    mov  si, offset msgFileSystemError  // FreeLdr not found message
```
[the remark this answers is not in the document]
sdever: This is weird, without the offset, the emitted code dereferences the symbol instead of taking its address...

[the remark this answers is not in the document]
HBelusca: What's the MASM/ML version you are using? If it's one from VS 2019, could you try testing with an older VS version if possible? Maybe it's yet another one of their new "bugs"... 🤔

HBelusca:
**Rebased** PR on master after PR #3775 being committed.

binarymaster:
@JoachimHenze reopening would only make sense if you're going to handle the PR yourself (and self-assign it).

It's not assigned to anyone so far, so it will be closed once again in 4 months.

JoachimHenze:
We have one devs approval so far. That might encourage a seocnd one. At least that is my hope. Otherwise I will accordingly to our guidelines push the SquashnMerge button in a week from now.
The only critics in here so far is missing documentation (extravert). That could be added retrospectively as well. And imho doesn't justify blocking the PR.

JoachimHenze (approved):

boot/freeldr/freeldr/lib/fs/ntfs.c
```diff
@@ -733,6 +733,7 @@ static BOOLEAN NtfsLookupFile(PNTFS_VOLUME_INFO Volume, PCSTR FileName, PNTFS_MF
             TRACE("- Failed\n");
             return FALSE;
         }
+        CurrentMFTIndex &= NTFS_MFT_MASK;
```
HBelusca, line 736: Not necessary anymore, since `NtfsFindMftRecord` applies the mask (see PR #3775).
```suggestion
```

boot/freeldr/bootsect/ntfs.S
```diff
@@ -0,0 +1,890 @@
...
+        pushad
+        call  ReadNonResidentAttribute
+        popad
+        mov   bx, fs
+        add   bx, HEX(100)
+        mov   fs, bx
+        xor   bx, bx
+        inc   ecx
+        cmp   ecx, edx
+        jbe   FreeLdrLoad
```
HBelusca, line 413: Those three lines seem to cause a loading problem (takes ~ 40 seconds); @DarkFire01 and @archeYR have a patch for it that we're currently reviewing.

boot/freeldr/bootsect/ntfs.S
```diff
@@ -0,0 +1,890 @@
...
+    jnz   PrintFileSystemError  // jump if not valid
+    pop   bx
+    
+    mov   ax, fs
+    mov   es, ax
+    call  ApplyFixups
+    pop   es
+    ret
+
+// Finds an attribute header intho the MFT
```
HBelusca, line 842:
```suggestion
// Finds an attribute header into the MFT
```

boot/freeldr/bootsect/ntfs.S
HBelusca: @sdever Are you able to reproduce that problem I evoked above? Or does freeldr load fast for you?

boot/freeldr/bootsect/ntfs.S
```diff
@@ -0,0 +1,890 @@
...
+    mov   cx, word ptr [BytesPerSector]
+    div   cx
+    mov   word ptr [SectsPerMFT], ax
+
+    SearchNext:
+    mov   bp, sp
+    sub   sp, HEX(10)
+
+    // Read Root Directory MFT
+    mov   ax, HEX(2000)
```
disean, line 350: Can you document the location and size of the disk buffer? See https://git.reactos.org/?p=reactos.git;a=blob;f=boot/freeldr/bootsect/ext2.S;h=bd3119c635feafba68ee36677098b4aa3efdea0d;hb=refs/heads/master#l129

disean (commented):
Minor suggestions for reducing code size

HBelusca:
@sdever I think that we'll merge this PR soon, in order to not block it more (if you don't see anything against it).
Those extra TODOs can be dealt with in separate PRs.

boot/freeldr/bootsect/ntfs.S
DarkFire01: This has been resolved I think, I formatted a drive on XP and it worked as you’d expect. Unless I’m missing something can be resolved.

boot/freeldr/freeldr/lib/fs/ntfs.c
```diff
@@ -1,7 +1,7 @@
 /*
  *  FreeLoader NTFS support
  *  Copyright (C) 2004  Filip Navara  <xnavara@volny.cz>
- *  Copyright (C) 2009-2010  Herv� Poussineau
+ *  Copyright (C) 2009-2010  Herv� Poussineau
```
tkreuzer, line 4: Please revert this change.

tkreuzer (approved):

boot/freeldr/freeldr/lib/fs/ntfs.c
JoachimHenze: Timo, This is no intentional change by sdever. It is a bug in the github website, **which always chokes on Herves** name when showing PRs. That is why I recommended in the past to change his name to H. Poussineau to workaround that github bug. You can verify my claim by putting a .patch behind the URL
https://github.com/reactos/reactos/pull/3416.patch
The change is not visible in there.

tkreuzer: Huh? GitHub doesn't make up changes in unmodified files!
Also what is this?
`
  *  Copyright (C) 2004  Filip Navara  <xnavara@volny.cz>
- *  Copyright (C) 2009-2010  Herv� Poussineau
+ *  Copyright (C) 2009-2010  Herv� Poussineau
  *
`

JoachimHenze: You are right. It **is** within the patch. I have seen this exact same thing on many PRs before though.

binarymaster: It's not a problem and it doesn't need a change. File was encoded in Win-1252, and now it's UTF-8.

JoachimHenze: Thank you for double-checking this specific file @binarymaster .
We should double-check all others too with "Poussineau".

HBelusca (approved):

sdever:
> @sdever I think that we'll merge this PR soon, in order to not block it more (if you don't see anything against it). Those extra TODOs can be dealt with in separate PRs.

Yes, it can be merged as is

HBelusca (approved):

binarymaster (approved):
🚢 :shipit:

DarkFire01 (approved):

HBelusca:
Thanks for your work @sdever ! :tada:

PR #3426, opened 2021-01-15
[AFD] Don't mark a closed socket as Overread

learn-more:
## Purpose

No clue if this is correct, but it fixes the test (tm)

JIRA issue: [CORE-17425](https://jira.reactos.org/browse/CORE-17425)

### DIFF
--- a/modules/rostests/apitests/ws2_32/recv.c
+++ b/modules/rostests/apitests/ws2_32/recv.c
@@ -1,8 +1,8 @@
 /*
  * PROJECT:         ReactOS api tests
  * LICENSE:         GPL - See COPYING in the top level directory
- * PURPOSE:         Test for recv
- * PROGRAMMERS:     Colin Finck
+ * PURPOSE:         Test for recv with MSG_PEEK
+ * PROGRAMMERS:     Copyright 2008 Colin Finck <colin@reactos.org>
  */
 
 #include "ws2_32.h"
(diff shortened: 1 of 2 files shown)


modules/rostests/apitests/ws2_32/recv.c
```diff
@@ -113,6 +113,21 @@ int Test_recv()
     ok(readIosb.Information == RECV_BUF, "Short read\n");
 
     NtClose(readEvent);
+
+    /* Now keep reading until the end (CORE-17425) */
+    int ret;
+    char recv_data[100] = { 0 };
```
SergeGautherie, line 119: Is the initialization needed?

modules/rostests/apitests/ws2_32/recv.c
```diff
@@ -113,6 +113,21 @@ int Test_recv()
     ok(readIosb.Information == RECV_BUF, "Short read\n");
 
     NtClose(readEvent);
+
+    /* Now keep reading until the end (CORE-17425) */
+    int ret;
```
HBelusca, line 118: You could resuse the existing `iResult` variable (that is not used as well anymore after that in the function).

modules/rostests/apitests/ws2_32/recv.c
HBelusca, line 119: You could reuse one of these `szBuf1` buffers already declared in the function.
This would also allow the test to continue compiling in my setup and me not asking to move variables at the beginning of the function.

encodedpr (approved):

Jookia:
but why

PR #1701, opened 2019-06-29
[i8042prt] Initialize mice correctly and remove the need for device hacks

BelleNottelling:
## Purpose

Correctly initialize mice correctly to fix issues on some laptops, then initialize the wheel again to prevent it no longer working.

JIRA issue: [CORE-6901](https://jira.reactos.org/browse/CORE-6901)
JIRA issue: [CORE-12663](https://jira.reactos.org/browse/CORE-12663)
JIRA issue: [CORE-12434](https://jira.reactos.org/browse/CORE-12434)

## Proposed changes

- During mouse detection set it into stream mode (which is what we use) and then disable error reporting (as described [here](https://isdaman.com/alsos/hardware/mouse/ps2interface.htm))
- Initialize the mouse wheel (Microsoft IntelliMouse compatible devices) by setting the polling rate to `200`, `100`, and then `80`
- If the device reports that it's a Microsoft IntelliMouse compatible device enable the scroll wheel

## TODO
- Test and verify this patch doesn't break HW (In progress)

## Tests performed
 _Checkmark means it passed_

- \[x] Latitude E6400
- \[x] Virtualbox with PS2 mouse enabled in settings
- \[x] Qemu

PR #3429, opened 2021-01-18
[i8042prt] Fixed mouse type detection regression on Compaq N600c

arjav2002:
https://github.com/reactos/reactos/pull/1701
Fixes the regression on @Extravert-ir 's Compaq N600c laptop

@BenNottelling is the co-author of this pull request. I have added one write to the `i8042prt` which makes sure the data/error reporting is enabled at the end of the routine added in the original pull request. Routines related to registries have remained untouched unlike in the original pull request


BenNottelling:
As per the documentation [here](https://isdaman.com/alsos/hardware/mouse/ps2interface.htm) (which is what my changes were based off of) it states

> If the mouse is in Stream mode, the host should disable data reporting (command 0xF5) before sending any other commands...

So I question if turning it back on actually makes sense..

drivers/input/i8042prt/pnp.c
```diff
@@ -263,6 +263,50 @@ i8042DetectMouse(
     INFO_(I8042PRT, "Mouse not detected\n");
 }
 
+static void
+i8042DetectMouseType(
+    IN PPORT_DEVICE_EXTENSION DeviceExtension)
+{
+    int chkcnt = 10;
```
BenNottelling, line 270: If you're going re-write my loop, you need to delete the variable from it, as well

arjav2002:
> As per the documentation [here](https://isdaman.com/alsos/hardware/mouse/ps2interface.htm) (which is what my changes were based off of) it states
> 
> > If the mouse is in Stream mode, the host should disable data reporting (command 0xF5) before sending any other commands...
> 
> So I question if turning it back on actually makes sense..

Yes, so it is only enabling it again after the detection of the mouse type is done.

BenNottelling (commented):
I don't really understand why you're applying my changes to avoid hardware hacks & the extra mouse reset, but then still leave all of those in-place.

arjav2002:
> I don't really understand why you're applying my changes to avoid hardware hacks & the extra mouse reset, but then still leave all of those in-place.

Okay, I'll review it and make sure only things pertaining to the registry are not deleted.

drivers/input/i8042prt/pnp.c
arjav2002: okay

arjav2002:
> I don't really understand why you're applying my changes to avoid hardware hacks & the extra mouse reset, but then still leave all of those in-place.

So I looked at it, and it looks like hwhacks.c only has registry related code with (SMBiosTables etc), which shouldn't be removed as there's no driver for it yet.

arjav2002:
Removed the second reset @BenNottelling

Extravert-ir:
> 
> 
> > I don't really understand why you're applying my changes to avoid hardware hacks & the extra mouse reset, but then still leave all of those in-place.
> 
> So I looked at it, and it looks like hwhacks.c only has registry related code with (SMBiosTables etc), which shouldn't be removed as there's no driver for it yet. @Extravert-ir back me up on this?

The registry stuff should be kept, but hardware hacks should be removed ofc. It's a bit confusing SMBios stuff was put there

arjav2002:
Correct me if I'm wrong, but `i8042InitializeHwHacks` only seems to do smbios things.

> > > I don't really understand why you're applying my changes to avoid hardware hacks & the extra mouse reset, but then still leave all of those in-place.
> > 
> > 
> > So I looked at it, and it looks like hwhacks.c only has registry related code with (SMBiosTables etc), which shouldn't be removed as there's no driver for it yet. @Extravert-ir back me up on this?
> 
> The registry stuff should be kept, but hardware hacks should be removed ofc. It's a bit confusing SMBios stuff was put there

Extravert-ir:
Add `Co-authored-by: Benjamin Aerni <daprogramerforbatch@gmail.com>` string to the end of the commit message

arjav2002:
> Add `Co-authored-by: Benjamin Aerni <daprogramerforbatch@gmail.com>` string to the end of the commit message

done

drivers/input/i8042prt/pnp.c
```diff
@@ -263,6 +263,49 @@ i8042DetectMouse(
     INFO_(I8042PRT, "Mouse not detected\n");
 }
 
+static void
+i8042DetectMouseType(
+    IN PPORT_DEVICE_EXTENSION DeviceExtension)
+{
+    UCHAR Value;
```
HBelusca, line 270:
```suggestion
    UCHAR Value;
    INT i;

```

drivers/input/i8042prt/pnp.c
```diff
@@ -263,6 +263,49 @@ i8042DetectMouse(
...
+    i8042IsrWritePort(DeviceExtension, 0xC8, CTRL_WRITE_MOUSE);
+
+    i8042IsrWritePort(DeviceExtension, MOU_DEF_SAMPLE_RATE, CTRL_WRITE_MOUSE);
+    i8042IsrWritePort(DeviceExtension, 0x64, CTRL_WRITE_MOUSE);
+
+    i8042IsrWritePort(DeviceExtension, MOU_DEF_SAMPLE_RATE, CTRL_WRITE_MOUSE);
+    i8042IsrWritePort(DeviceExtension, 0x50, CTRL_WRITE_MOUSE);
+
+    i8042IsrWritePort(DeviceExtension, MOU_READ_DEV_TYPE, CTRL_WRITE_MOUSE);
+    for(int i = 0; i < 10; i++)
```
HBelusca, line 286:
```suggestion
    for (i = 0; i < 10; i++) 
```
Also a comment why the loop is done 10 times would be welcome.

drivers/input/i8042prt/pnp.c
```diff
@@ -263,6 +263,49 @@ i8042DetectMouse(
...
+    i8042IsrWritePort(DeviceExtension, 0x64, CTRL_WRITE_MOUSE);
+
+    i8042IsrWritePort(DeviceExtension, MOU_DEF_SAMPLE_RATE, CTRL_WRITE_MOUSE);
+    i8042IsrWritePort(DeviceExtension, 0x50, CTRL_WRITE_MOUSE);
+
+    i8042IsrWritePort(DeviceExtension, MOU_READ_DEV_TYPE, CTRL_WRITE_MOUSE);
+    for(int i = 0; i < 10; i++) 
+    {
+        i8042ReadDataWait(DeviceExtension, &Value);
+        if(Value == 0x03)
```
HBelusca, line 289:
```suggestion
        if (Value == 0x03)
```

drivers/input/i8042prt/pnp.c
```diff
@@ -263,6 +263,49 @@ i8042DetectMouse(
...
+        {
+            DeviceExtension->MouseExtension->MouseType = Intellimouse;
+            break;
+        }
+        else if(Value == 0x00)
```
HBelusca, line 294:
```suggestion
        else if (Value == 0x00)
```

Extravert-ir:
> So I question if turning it back on actually makes sense..

Adding that `MOU_ENAB` command at least made the touchpad work on my laptop.
Maybe the proper fix should be done some other way, but it was not suggested :)

drivers/input/i8042prt/pnp.c
arjav2002: @BenNottelling selected this number in his PR.

Extravert-ir:
> > Add `Co-authored-by: Benjamin Aerni <daprogramerforbatch@gmail.com>` string to the end of the commit message
> 
> done

https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors#creating-co-authored-commits-on-the-command-line

drivers/input/i8042prt/pnp.c
BenNottelling: I simply picked it because it should loop through the mouse messages enough times to read everything.
Well, I kind assumed so

drivers/input/i8042prt/pnp.c
arjav2002: done

arjav2002:
> > > Add `Co-authored-by: Benjamin Aerni <daprogramerforbatch@gmail.com>` string to the end of the commit message
> > 
> > 
> > done
> 
> https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors#creating-co-authored-commits-on-the-command-line

done

drivers/input/i8042prt/pnp.c
```diff
@@ -470,6 +516,8 @@ StartProcedure(
         DeviceExtension->Flags & MOUSE_STARTED &&
         !(DeviceExtension->Flags & MOUSE_INITIALIZED))
     {
+        /* Detect the mouse type here, rather than in i8042DetectMouse, fixes lockup on BootCD */
+        i8042DetectMouseType(DeviceExtension);
```
HBelusca, line 520: As already mentioned in the original PR by @BenNottelling , the mouse type is also set in the `i8042MouResetIsr()` function, see https://git.reactos.org/?p=reactos.git;a=blob;f=drivers/input/i8042prt/mouse.c;hb=6d238e451345dfd4f5b8178d4d515801546fe95d#l629
Hence I have a question: When is this `i8042MouResetIsr()` function called, relative to this new `i8042DetectMouseType()` of yours? Before? After?
And would it be then possible to not duplicate the functionality of setting the MouseType?

drivers/input/i8042prt/pnp.c
```diff
@@ -263,6 +263,52 @@ i8042DetectMouse(
     INFO_(I8042PRT, "Mouse not detected\n");
 }
 
+static void
+i8042DetectMouseType(
+    IN PPORT_DEVICE_EXTENSION DeviceExtension)
+{
+    UCHAR Value;
+    INT i;
```
HBelusca, line 271: Please add a newline after this variable to separate the variables from the rest of the code.

drivers/input/i8042prt/pnp.c
arjav2002: `i8042DetectMouseType()` is called right before the interrupt for the mouse is connected. The mouse ISR is the one calling `i8042MouResetIsr`, hence before.

arjav2002:
A thing that I found weird is that in @Extravert-ir 's Compaq, it works after enabling the mouse at the end in the routine, but suddenly stops working if this is not done, **even when the same thing is done right after `i8042DetectMouseType` in `i8042MouInitialize` in `i8042ConnectMouseInterrupt`. It is the second write to i8042prt.**

drivers/input/i8042prt/pnp.c
arjav2002: > As already mentioned in the original PR by @BenNottelling , the mouse type is also set in the `i8042MouResetIsr()` function, see https://git.reactos.org/?p=reactos.git;a=blob;f=drivers/input/i8042prt/mouse.c;hb=6d238e451345dfd4f5b8178d4d515801546fe95d#l629
> Hence I have a question: When is this `i8042MouResetIsr()` function called, relative to this new `i8042DetectMouseType()` of yours? Before? After?
> And would it be then possible to not duplicate the functionality of setting the MouseType?

I would be willing to explore the possibility of not having to duplicate. What I'm given to understand is that the second reset does that. This causes the trackpad on some dell laptops to not work. This is where hwhacks was introduced. So if  someone is willing to give me debug logs for the D latitude series Dell laptops, that would be great. I can try to get the trackpad to work even with the second reset.

Or we can simply make a quick fix like it is happening currently.

JoachimHenze:
A regression? Which revision did regress your mouse Extravert-ir ?

BenNottelling:
> A regression? Which revision did regress your mouse Extravert-ir ?

No regression happened in master.
It's related to my PR#1701

arjav2002:
``And would it be then possible to not duplicate the functionality of setting the MouseType?``
@HBelusca an attempt to do it. Works on @Extravert-ir 's  Compaq on which PR [#1701](https://github.com/reactos/reactos/pull/1701) regressed. Also works on @Doug-Lyons 's Dell D620a.

Extravert-ir (changes requested):
Can you explain the changes? What's being done now that was not done before?
This stuff definitely requires some comments both in code and commit changes

arjav2002:
> Can you explain the changes? What's being done now that was not done before?
> This stuff definitely requires some comments both in code and commit changes

Removed all hardware related stuff in hwhacks, Changed some hex values for their corresponding ``#define``s. @BenNottelling 's PR #1701 had code for enabling wheels. So that was a clue that it had to integrated with the already existing  case label ``EnableWheel``. ``master`` branch already had code for enabling the wheels. PR #1701 had executed two more commands (``MOU_SET_STREAM_MODE`` and ``MOU_NO_ERROR_REPORT``) before enabling the wheels. So that is what has been added into ``EnableWheel`` case label.

A way to keep track of number of ACKs received (as in the while loop in #1701) was also needed, for which a new variable in the DeviceExtension was created. So, if it was an Intellimouse, the already existing code was allowed to handle that case, else AckCount is incremented in a for loop fashion. 

Then there was my fix to PR #1701 which made it work on @Extravert-ir 's Compaq. The ``MOU_ENAB`` command. That was added at the end of setting the default sampling rate. This was also done in the fix to #1701. 

Lastly, the if block in pnp.c was removed because all the hardware related stuff was also removed.

drivers/input/i8042prt/mouse.c
```diff
@@ -655,6 +655,7 @@ i8042MouResetIsr(
 
 	DeviceExtension->MouseTimeoutState = TimeoutStart;
 	PortDeviceExtension = DeviceExtension->Common.PortDeviceExtension;
+	DPRINT1("Value: %x\tResetState: %x\n", Value, DeviceExtension->MouseResetState);
```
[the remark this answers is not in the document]
arjav2002: Done.

Extravert-ir:
Cool, now reformat it for a commit message.
Don't refer to the previous PR and people there, it doesn't make sense. The commit message should be self-explanatory

arjav2002:
> Cool, now reformat it for a commit message.
> Don't refer to the previous PR and people there, it doesn't make sense. The commit message should be self-explanatory

done

ColinFinck:
Really nice @arjav2002 and @BenNottelling! I have been looking forward to see these ugly and non-scalable hardware hacks removed for a long time.
Code looks fine from my side. I just wonder if it has been tested on enough old and new systems (considering that every chipset and device vendor has its own dialect of the standard, which brought us this mess in the first place..)

arjav2002:
Currently fails on Dell Latitude E6400, @BenNottelling is helping me fix it.

BenNottelling:
I would like to make a note that [CORE-17470](https://jira.reactos.org/browse/CORE-17470) is getting in the way of debugging on my Latitude E6400. I have tried reverting both the work that's referenced as being responsible for this regression and MM work in a local branch to no luck. I need the PNP work Extravert did so I can install from USB, currently I have no way to write to a DVD/CD. I do intend to try reverting again incase I did something wrong.

@arjav2002 
I noticed ```MouseResetState``` 1004 is blank. The result is the code path you've created doesn't send a ```MOU_DEF_SAMPLE_RATE``` and it tries to send two mouse polling rates one after another. This isn't correct and doesn't follow spec, as such it causes a regression in [V86](https://copy.sh/v86/).
I expect Qemu & VirtualBox to experience the same issue, but I'm having issues with both at the moment and haven't figured out if it's a recent commit or something on my end.

Fixing this so the driver doesn't miss the ```MOU_DEF_SAMPLE_RATE``` allows the mouse to work currently in V86 once again. Still broken on my laptop.
As a trial, I removed the command to re-enable error reporting as that's how I had it, this didn't fix anything and actually caused it to break again in V86. 

I'm probably going to stop for the night, but if I can tomorrow while I'm at work I'll try to go over the current code path and see if there's anything I think that may case issues. 

Also, complaint. You go case 1004, 1006, and then 1005. This bothers me and also means you can't read top down in the code
image

arjav2002:
Thanks for noticing!

arjav2002:
> I would like to make a note that [CORE-17470](https://jira.reactos.org/browse/CORE-17470) is getting in the way of debugging on my Latitude E6400. I have tried reverting both the work that's referenced as being responsible for this regression and MM work in a local branch to no luck. I need the PNP work Extravert did so I can install from USB, currently I have no way to write to a DVD/CD. I do intend to try reverting again incase I did something wrong.
> 
> @arjav2002 
> I noticed ```MouseResetState``` 1004 is blank. The result is the code path you've created doesn't send a ```MOU_DEF_SAMPLE_RATE``` and it tries to send two mouse polling rates one after another. This isn't correct and doesn't follow spec, as such it causes a regression in [V86](https://copy.sh/v86/).

Thanks for noticing!

arjav2002:
> @arjav2002
> I noticed `MouseResetState` 1004 is blank. The result is the code path you've created doesn't send a `MOU_DEF_SAMPLE_RATE` and it tries to send two mouse polling rates one after another. This isn't correct and doesn't follow spec, as such it causes a regression in [V86](https://copy.sh/v86/).

No see, the case is empty so it falls through into 1006, sends the MOU_DEF_SAMPLE_RATE and then increments the `MouseResetState` to 1005. Then 1005 and then 1006. I'll change it anyway though.

arjav2002:
> @arjav2002
> I noticed `MouseResetState` 1004 is blank. The result is the code path you've created doesn't send a `MOU_DEF_SAMPLE_RATE` and it tries to send two mouse polling rates one after another. This isn't correct and doesn't follow spec, as such it causes a regression in [V86](https://copy.sh/v86/).
> I expect Qemu & VirtualBox to experience the same issue, but I'm having issues with both at the moment and haven't figured out if it's a recent commit or something on my end.

I'm not too sure about what you mean. Is the code path correct as of now? If not, it would be helpful if you could quote where exactly it starts going wrong.

BenNottelling:
> I'm not too sure about what you mean. Is the code path correct as of now? If not, it would be helpful if you could quote where exactly it starts going wrong.

I don't know either, I think I made a comment miss understanding what you did and then you changed the code to be more obvious so I don't think that comment is relevant

arjav2002:
> I would like to make a note that [CORE-17470](https://jira.reactos.org/browse/CORE-17470) is getting in the way of debugging on my Latitude E6400. I have tried reverting both the work that's referenced as being responsible for this regression and MM work in a local branch to no luck. I need the PNP work Extravert did so I can install from USB, currently I have no way to write to a DVD/CD. I do intend to try reverting again incase I did something wrong.

Does this mean it is a "cyclic" problem? https://jira.reactos.org/browse/CORE-17722
image

PR #3433, opened 2021-01-23
[ROSAUTOTEST] Allow rosautotest in same directory as other files and show time at end.

Doug-Lyons:
## rosautotest improvements

_Add ability for rosautotest to reside in the same directory as its files and show duration at end._

Typical results:
[ROSAUTOTEST] System uptime at start was 412412.86 seconds

[ROSAUTOTEST] System uptime at end was 412412.88 seconds
[ROSAUTOTEST] Duration was 0.00 minutes

The main reason that I created this was to allow me to be able to copy a single folder named "bin" into a downloaded ISO install of ReactOS. These have no ReactOS tests included with them. I wanted to be able to get the Regression Test results to use as a base and I originally had to merge my external "bin" folder with the new install and then copy over the rosautotest.exe file into the C:\ReactOS\system32 subdirectory. I wanted to be able to simplify this action, so the idea to make rosautotest.exe exclude itself from trying to run itself as a test was born. Also, these tests take a long time and scroll by hundreds of lines and I wanted to be able to see at the end of the tests just how long that they had taken. I created this PR so that others could benefit from my efforts. Thanks.


modules/rostests/rosautotest/main.cpp
```diff
@@ -100,6 +102,22 @@ wmain(int argc, wchar_t* argv[])
             WineTest.Run();
         }
 
+        /* Clear the stringstream */
+        ss.str("");
+        ss.clear();
+
+        /* Show the beginning time */
+        ss << "[ROSAUTOTEST] System uptime at start was " << setprecision(2) << fixed;
+        ss << ((float)TestStartTime/1000) << " seconds" << endl;
```
SergeGautherie, line 111: Add spaces around operators:
```suggestion
        ss << ((float)TestStartTime / 1000) << " seconds" << endl;
```

modules/rostests/rosautotest/main.cpp
```diff
@@ -100,6 +102,22 @@ wmain(int argc, wchar_t* argv[])
...
+
+        /* Show the time now so that we can see how long it took. */
+        ss << endl
+           << "[ROSAUTOTEST] System uptime at end was " << setprecision(2) << fixed;
+        ss << ((float)GetTickCount()/1000) << " seconds" << endl;
+        ss << "Duration was " << (((float)GetTickCount()/1000) - ((float)TestStartTime/1000))/60;
```
SergeGautherie, line 117:
```suggestion
        ss << "[ROSAUTOTEST] Duration was " << (((float)GetTickCount()/1000) - ((float)TestStartTime/1000))/60;
```

modules/rostests/rosautotest/main.cpp
```diff
@@ -100,6 +102,22 @@ wmain(int argc, wchar_t* argv[])
             WineTest.Run();
         }
 
+        /* Clear the stringstream */
+        ss.str("");
+        ss.clear();
+
+        /* Show the beginning time */
+        ss << "[ROSAUTOTEST] System uptime at start was " << setprecision(2) << fixed;
```
SergeGautherie, line 110: Are the 3 time needed? Do they need to be on 3 separate lines? And/Or should 'start' time be written at actual start?

modules/rostests/rosautotest/main.cpp
Doug-Lyons: I will make this change. Thanks.

modules/rostests/rosautotest/main.cpp
Doug-Lyons: I will make this change. Thanks.

modules/rostests/rosautotest/CWineTest.cpp
```diff
@@ -8,6 +8,7 @@
 #include "precomp.h"
 
 static const DWORD ListTimeout = 10000;
+static WCHAR TestName[] = L"rosautotest.exe";
```
HBelusca, line 11: What if I happen to have copied and renamed this file to some other name, but otherwise try the setup you are trying to fix?

Doug-Lyons: See latest commit and let me know what you think. Thank.

modules/rostests/rosautotest/main.cpp
```diff
@@ -8,6 +8,7 @@
 #include "precomp.h"
 #include <cstdio>
 DWORD TestStartTime;
+WCHAR TestName[];
```
SergeGautherie, line 11: Missing length.

modules/rostests/rosautotest/main.cpp
```diff
@@ -52,6 +53,9 @@ wmain(int argc, wchar_t* argv[])
 {
     int ReturnValue = 1;
 
+    wcscpy(TestName, argv[0]);
+    wcsncat (TestName, L".exe", 4);
```
SergeGautherie, line 57: These calls are unsafe.

modules/rostests/rosautotest/main.cpp
SergeGautherie: Even assuming it does not hurt to report the 3 values, 1 log line should be enough for that purpose.

modules/rostests/rosautotest/main.cpp
HBelusca: Also you should use `GetModuleFileNameW(NULL, TestName, _countof(TestName));`

modules/rostests/rosautotest/main.cpp
Doug-Lyons: This was done purposefully here to indicate that this is only a forward define and not the primary definition which is actually in precomp.h.

modules/rostests/rosautotest/main.cpp
Doug-Lyons: I like the GetModuleFineNameW idea and changed the code to reflect this method. Please have a look. Also, I used a string safe function to copy the truncated file name to TestName.

modules/rostests/rosautotest/main.cpp
```diff
@@ -52,6 +53,9 @@ wmain(int argc, wchar_t* argv[])
 {
     int ReturnValue = 1;
 
+    wcscpy(TestName, argv[0]);
```
learn-more: `argv[0]` entirely depends on how the application was started, it could be `rosautotest.exe`, `c:\tests\rosautotest.exe` or even something like `..\rosautotest.exe`.
It would be more defined to use something like `GetModuleFileNameW(NULL, path, _countof(path))` and only use the filename of that

Doug-Lyons: Thanks, @learn-more. I did not realized this limitation of the argv[0] feature. I don't think that I have ever tried to use it programmatically before, but was aware that it returned the first part of the command line. The code has been modified to use the GetModuleFileNameW now.

HBelusca:
@ColinFinck hi, any thoughts about this feature?

modules/rostests/rosautotest/precomp.h
```diff
@@ -80,6 +80,6 @@ string StringOut(const string& String, bool forcePrint = true);
 string UnicodeToAscii(PCWSTR UnicodeString);
 string UnicodeToAscii(const wstring& UnicodeString);
 
-extern WCHAR TestName[50];
+extern WCHAR TestName[250];
```
HBelusca, line 83: Why not choosing `MAX_PATH` ?

modules/rostests/rosautotest/main.cpp
```diff
@@ -100,6 +110,22 @@ wmain(int argc, wchar_t* argv[])
             WineTest.Run();
         }
 
+        /* Clear the stringstream */
+        ss.str("");
+        ss.clear();
+
+        /* Show the beginning time */
```
HBelusca, line 117: maybe "Show again the beginning time" ; with "again" implicitly referring to the fact that it was already shown somewhere else (i.e. when starting the tests).

modules/rostests/rosautotest/main.cpp
```diff
@@ -100,6 +110,22 @@ wmain(int argc, wchar_t* argv[])
...
+        ss << "[ROSAUTOTEST] System uptime at start was " << setprecision(2) << fixed;
+        ss << ((float)TestStartTime / 1000) << " seconds" << endl;
+
+        /* Show the time now so that we can see how long it took. */
```
HBelusca, line 121:
```suggestion
        /* Show the time now so that we can see how long the tests took */
```

modules/rostests/rosautotest/main.cpp
```diff
@@ -100,6 +110,22 @@ wmain(int argc, wchar_t* argv[])
...
+        ss << endl
+           << "[ROSAUTOTEST] System uptime at end was " << setprecision(2) << fixed;
+        ss << ((float)GetTickCount()/1000) << " seconds" << endl;
```
HBelusca, line 124: It may be of interest to store the "GetTickCount()" value in a variable.
Because the function is again called below and there certainly have some ticks that have elapsed between there and below.

modules/rostests/rosautotest/main.cpp
```diff
@@ -7,6 +7,9 @@
 
 #include "precomp.h"
 #include <cstdio>
+#include <strsafe.h>
+DWORD TestStartTime;
```
HBelusca, line 11: Move the TestStartTime variable inside the wmain() since it's only used there.

modules/rostests/rosautotest/main.cpp
```diff
@@ -7,6 +7,9 @@
...
+WCHAR TestName[];
```
HBelusca, line 12: missing the array size.

DarkFire01 (approved):

learn-more:
Was this ran against testbots?
If not, we should probably do that because they might be particular about the output format.

modules/rostests/rosautotest/precomp.h
learn-more: `MAX_PATH` is 260 so this needs to either be changed to `260` or to `MAX_PATH`

Doug-Lyons:
Replaced by PR #7823 because of problems with rebase.

PR #3435, opened 2021-01-28
[CMAKE] Generate Flex and Bison files at build time

Extravert-ir:
We have Flex and Bison distributed along with RosBE 2.2 so this should not be a problem

Supersedes #2148


SergeGautherie:
5375e33:
It looks like the `set_source_files_properties()` line in `dll/win32/msi/CMakeLists.txt` was missed/leftover...

From #2148:
Why not update `base/applications/winhlp32/*` too?

New:
Maybe remove `sdk/tools/widl/readme.txt` too, now?

PR #3437, opened 2021-01-29
[NTOS:PNP] Fix regression of HAL recognition in Device Manager

binarymaster:
Send `GUID_DEVICE_ENUMERATED` event for root devices.

According to Victor, Windows Server 2003 does not do this, however this fixes the regression with least changes in the kernel.

- Also initialize device description registry value, as it's required by New device installer (newdev.dll) to install driver from INF file.

JIRA issues: [CORE-17212](https://jira.reactos.org/browse/CORE-17212) [CORE-17398](https://jira.reactos.org/browse/CORE-17398)

PR #3438, opened 2021-01-30
[CMIPCI][DOC][GREEN][INF] Fix 7 DriverVer values 

SergeGautherie:
Follow-up to #3437.


binarymaster (approved):
Ideally, the INF files with `-WaveRT` should have driver version `1.2.6-WaveRT` in consistency with the [driver source code](https://git.reactos.org/?p=reactos.git;a=blob;f=drivers/wdm/audio/drivers/CMIDriver/mintopo.cpp;hb=0edaeba452c21a7e8b3a3e157577303fe18f6da6#l1360). However it seems INF files [have a strict format](https://docs.microsoft.com/en-us/windows-hardware/drivers/install/inf-driverver-directive) for the `DriverVer` directive, so it's the only way to write it correctly.

PR #3440, opened 2021-01-30
[TASKLIST] Implement tasklist command-line utility

kernelbin:
## Purpose

```tasklist``` is missing in ReactOS currently.
and It might be used by some programs or scripts

### screenshot:
TI1({{G5X_6E6)VA$IZ ~BR

It works fine on Windows as well

## Proposed changes

add ```tasklist``` for ReactOS

## TODO

- [x] no header param support (```/nh``` param)

### CODE REVIEW

File: base/applications/cmdutils/tasklist/tasklist.c [hunk 1]
```diff
@@ -0,0 +1,398 @@
...
+{
+    // Load ntdll.dll in order to use NtQuerySystemInformation
+    HMODULE hNtDLL = LoadLibraryW(L"Ntdll.dll");
+    if (!hNtDLL)
+    {
+        ConResMsgPrintf(StdOut, 0, IDS_ENUM_FAILED);
+        return FALSE;
+    }
+
+    NT_QUERY_SYSTEM_INFORMATION PtrNtQuerySystemInformation =
```
HBelusca, line 206: Do you consider this program to be able to run on any Windows NT (including of course 2000, 2003/XP, Vista+), but not on Windows 9x?
If so you could just link tasklist at compile time with ntdll, and directly use NtQuerySystemInformation() without doing that runtime delayloading.

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,398 @@
...
+        (NT_QUERY_SYSTEM_INFORMATION)GetProcAddress(hNtDLL, "NtQuerySystemInformation");
+
+    if (!PtrNtQuerySystemInformation)
+    {
+        ConResMsgPrintf(StdOut, 0, IDS_ENUM_FAILED);
+        FreeLibrary(hNtDLL);
+        return FALSE;
+    }
+
+    // call NtQuerySystemInformation for the process information
```
HBelusca, line 216:
```suggestion
    // Call NtQuerySystemInformation for the process information
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,398 @@
...
+    ULONG ProcessInfoBufferLength = 0;
+    ULONG ResultLength = 0;
+    PBYTE ProcessInfoBuffer = 0;
+
+    // Get the buffer size we need
+    NTSTATUS Status = PtrNtQuerySystemInformation(SystemProcessInformation, NULL, 0, &ResultLength);
+
+    for (int Retry = 0; Retry < NT_SYSTEM_QUERY_MAX_RETRY; Retry++)
```
HBelusca, line 224: What if NT_SYSTEM_QUERY_MAX_RETRY are not enough to retrieve the complete list of processes?

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,398 @@
...
+    if (!NT_SUCCESS(Status))
+    {
+        // tried NT_SYSTEM_QUERY_MAX_RETRY times, or failed with some other reason
+        ConResMsgPrintf(StdOut, 0, IDS_ENUM_FAILED);
+        HeapFree(GetProcessHeap(), 0, ProcessInfoBuffer);
+        FreeLibrary(hNtDLL);
+        return FALSE;
+    }
+
+    // print header
```
HBelusca, line 276: Comments like these should not be added.

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,398 @@
...
+// MaxWidth is the width for printing.
+INT PrintString(LPCWSTR String, int MaxWidth)
+{
+    return ConPrintf(StdOut, L"%-*.*ls", MaxWidth, MaxWidth, String);
+}
+
+// Print a string, aligned to right.
+// MaxWidth is the width for printing.
+// return FALSE if number length is longer than MaxWidth
+BOOL PrintNum(long long Number, int MaxWidth)
```
HBelusca, line 109: Please use NT types (ULONGLONG, INT etc...) or things like INT32, INT64 ...

base/applications/cmdutils/tasklist/tasklist.c
kernelbin: it's advised by msdn to load this dynamically, as it might be altered in the future.

I think stability is more important, and I don't think there's too much difference on the performance😂

base/applications/cmdutils/tasklist/tasklist.c
kernelbin: ok!  I will fix it tomorrow

base/applications/cmdutils/tasklist/tasklist.c
HBelusca: MSDN advices and information are for 3rd-party coders writing programs. Not for developers writing core utilities and the OS itself, like we do here in ReactOS.

base/applications/cmdutils/tasklist/tasklist.c
kernelbin: the required length is returned by ResultLength, retrying is just in case any new process/thread are created during this period.

should I increase this value? how much do you think is enough?😂😂

or should I change the process/thread priority to a higher level during querying？

base/applications/cmdutils/tasklist/lang/bg-BG.rc
```diff
@@ -0,0 +1,20 @@
...
+    IDS_INVALID_SYNTAX, "Error: Invalid syntax.\n"
+    IDS_INVALID_OPTION, "Error: Unknown or invalid command line option specified.\n"
+    IDS_ENUM_FAILED, "Error: Unable to enumerate the process list.\n"
+    IDS_OUT_OF_MEMORY, "Error: Out of memory.\n"
+    IDS_HEADER_IMAGENAME, "Image Name"
+    IDS_HEADER_PID, "PID"
+    IDS_HEADER_SESSION, "Session#"
+    IDS_HEADER_MEMUSAGE, "Mem Usage"
+END
+
```
GeoB99, line 20: What's with the redundant whitespace here?

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,398 @@
+/*
+ * PROJECT:     ReactOS Tasklist Command
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Displays a list of currently running processes on the computer.
+ * COPYRIGHT:   Copyright 2020 He Yang (1160386205@qq.com)
+ */
+
+#include <stdio.h>
```
GeoB99, line 8: Headers inclusion, defines and other stuff should be on their specific header of the application.

base/applications/cmdutils/tasklist/tasklist.c
HBelusca: ok then this should be fine. Another way could be to do what taskmgr already does (or tlist or ctm commands in the rosapps directory).

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,398 @@
...
+{
+    int i;
+    BOOL bHasHelp = FALSE, bHasVerbose = FALSE;
+    for (i = 1; i < argc; i++)
+    {
+        int Argument = GetArgumentType(argv[i]);
+
+        switch (Argument)
+        {
+        case OP_PARAM_HELP:
```
GeoB99, line 332: Missing space for all the case bodies here within the switch.

base/applications/cmdutils/tasklist/tasklist.c
kernelbin: ah, do you mean I should create a ```tasklist.h``` and put all these stuff inside it ? 🤣

base/applications/cmdutils/tasklist/tasklist.c
kernelbin: fixed now, thanks 🤣

base/applications/cmdutils/tasklist/lang/bg-BG.rc
kernelbin: removed now!

base/applications/cmdutils/tasklist/tasklist.c
GeoB99: Yes.

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,405 @@
...
+
+VOID PrintSplitLine(INT Length)
+{
+    for (INT i = 0; i < Length; i++)
+    {
+        ConPrintf(StdOut, L"=");
+    }
+}
+
+VOID PrintHeader()
```
HBelusca, line 57:
```suggestion
VOID PrintHeader(VOID)
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,405 @@
...
+    LoadStringW(GetModuleHandle(NULL), IDS_HEADER_PID, lpstrPID, HEADER_STR_MAXLEN);
+    LoadStringW(GetModuleHandle(NULL), IDS_HEADER_SESSION, lpstrSession, HEADER_STR_MAXLEN);
+    LoadStringW(GetModuleHandle(NULL), IDS_HEADER_MEMUSAGE, lpstrMemUsage, HEADER_STR_MAXLEN);
+
+    ConPrintf(
+        StdOut, L"%-*.*ls %*.*ls %*.*ls %*.*ls",
+        COLUMNWIDTH_IMAGENAME, COLUMNWIDTH_IMAGENAME, lpstrImageName,
+        COLUMNWIDTH_PID, COLUMNWIDTH_PID, lpstrPID,
+        COLUMNWIDTH_SESSION, COLUMNWIDTH_SESSION, lpstrSession,
+        COLUMNWIDTH_MEMUSAGE, COLUMNWIDTH_MEMUSAGE, lpstrMemUsage);
```
HBelusca, line 74:
```suggestion
    ConPrintf(StdOut, L"%-*.*ls %*.*ls %*.*ls %*.*ls",
              COLUMNWIDTH_IMAGENAME, COLUMNWIDTH_IMAGENAME, lpstrImageName,
              COLUMNWIDTH_PID, COLUMNWIDTH_PID, lpstrPID,
              COLUMNWIDTH_SESSION, COLUMNWIDTH_SESSION, lpstrSession,
              COLUMNWIDTH_MEMUSAGE, COLUMNWIDTH_MEMUSAGE, lpstrMemUsage);
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,405 @@
...
+
+    // Call NtQuerySystemInformation for the process information
+    ULONG ProcessInfoBufferLength = 0;
+    ULONG ResultLength = 0;
+    PBYTE ProcessInfoBuffer = 0;
+
+    // Get the buffer size we need
+    NTSTATUS Status = PtrNtQuerySystemInformation(SystemProcessInformation, NULL, 0, &ResultLength);
+
+    for (INT Retry = 0; Retry < NT_SYSTEM_QUERY_MAX_RETRY; Retry++)
```
HBelusca, line 231: Following our previous discussion about that max_retry thing, could you add a comment that the purpose of this loop is to "mitigate" the case where new processes could appear between the NtQuerySystemInformation() call for size and the call for the actual data?

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,405 @@
...
+    {
+        // (Re)allocate buffer
+        ProcessInfoBufferLength = ResultLength;
+        ResultLength = 0;
+        if (ProcessInfoBuffer)
+        {
+            PBYTE NewProcessInfoBuffer = HeapReAlloc(GetProcessHeap(), 0, ProcessInfoBuffer, ProcessInfoBufferLength);
```
HBelusca, line 238:
```suggestion
            PBYTE NewProcessInfoBuffer = HeapReAlloc(GetProcessHeap(), 0,
                                                     ProcessInfoBuffer,
                                                     ProcessInfoBufferLength);
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,405 @@
...
+                FreeLibrary(hNtDLL);
+                return FALSE;
+            }
+        }
+
+        // Query information
+        Status = PtrNtQuerySystemInformation(
+            SystemProcessInformation, ProcessInfoBuffer, ProcessInfoBufferLength, &ResultLength);
+
+        if (Status != STATUS_INFO_LENGTH_MISMATCH)
```
HBelusca, line 268:
```suggestion
        Status = PtrNtQuerySystemInformation(SystemProcessInformation,
                                             ProcessInfoBuffer,
                                             ProcessInfoBufferLength,
                                             &ResultLength);
        if (Status != STATUS_INFO_LENGTH_MISMATCH)
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,405 @@
...
+        {
+            ConResMsgPrintf(StdOut, 0, IDS_INVALID_SYNTAX);
+            ConResMsgPrintf(StdOut, 0, IDS_USAGE);
+            return FALSE;
+        }
+        else
+        {
+            ConResMsgPrintf(StdOut, 0, IDS_USAGE);
+            ConResMsgPrintf(StdOut, 0, IDS_DESCRIPTION);
+            exit(0);
```
HBelusca, line 385:
```suggestion
            return FALSE;
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,405 @@
...
+    }
+    return TRUE;
+}
+
+int wmain(int argc, WCHAR *argv[])
+{
+    /* Initialize the Console Standard Streams */
+    ConInitStdStreams();
+
+    if(!ProcessArguments(argc, argv))
```
HBelusca, line 400:
```suggestion
    if (!ProcessArguments(argc, argv))
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,405 @@
...
+        HeapFree(GetProcessHeap(), 0, ProcessInfoBuffer);
+        FreeLibrary(hNtDLL);
+        return FALSE;
+    }
+
+    // print header
+    PrintHeader();
+
+    PSYSTEM_PROCESS_INFORMATION pSPI;
+    pSPI = (PSYSTEM_PROCESS_INFORMATION)ProcessInfoBuffer;
```
HBelusca, line 287: Instead of all these intermediate buffers, you could just have used a single pSPI pointer and get rid of your ProcessInfoBuffer. Then in the allocation loop from above you could have just reallocated pSPI up to its correct size.

kernelbin: pSPI will be changed when enumerating process, and we still need a pointer pointing to the original buffer in order to free it finally🤣

HBelusca: oh, ok then, fine :)

HBelusca:
Please add a newline before displaying the headers.

base/applications/cmdutils/tasklist/tasklist.c
kernelbin: good idea!

base/applications/cmdutils/tasklist/tasklist.c
kernelbin: OK, comment added! 👌

base/applications/cmdutils/tasklist/tasklist.c
kernelbin: Do you mean by this program can't run on windows 9x ... ?

HBelusca: > Do you mean by this program can't run on windows 9x ... ?

Yes this was also a question: Since I'm not sure this tasklist and the like are designed to run on Win9x (that we don't care since that tool is part of an NT system), there should be no need to also runtime-load ntdll and its api.

kernelbin:
> Please add a newline before displaying the headers.
> So that it can look like e.g.
> 
> ```
> C:\Users\TheUser>tasklist
> 
> Image Name                     PID Session Name        Session#    Mem Usage
> ========================= ======== ================ =========== ============
> ```

oh I even didn't noticed theres a new line above the headers!🤣🤣
now added

base/applications/cmdutils/tasklist/tasklist.c
kernelbin: now I made it linked to ```ntdll.dll``` statically. 🤣🤣 
but I still wonder under what situation should I link an application to ntdll statically .... ?

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,354 @@
...
+    WCHAR lpstrPID[HEADER_STR_MAXLEN];
+    WCHAR lpstrSession[HEADER_STR_MAXLEN];
+    WCHAR lpstrMemUsage[HEADER_STR_MAXLEN];
+
+    LoadStringW(GetModuleHandle(NULL), IDS_HEADER_IMAGENAME, lpstrImageName, HEADER_STR_MAXLEN);
+    LoadStringW(GetModuleHandle(NULL), IDS_HEADER_PID, lpstrPID, HEADER_STR_MAXLEN);
+    LoadStringW(GetModuleHandle(NULL), IDS_HEADER_SESSION, lpstrSession, HEADER_STR_MAXLEN);
+    LoadStringW(GetModuleHandle(NULL), IDS_HEADER_MEMUSAGE, lpstrMemUsage, HEADER_STR_MAXLEN);
+
+    PrintString(lpstrImageName, COLUMNWIDTH_IMAGENAME, TRUE);
```
learn-more, line 148: Maybe create a `PrintString` that accepts a resource ID?

kernelbin: That sounds great!

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,354 @@
...
+VOID PrintHeader(VOID)
+{
+    ConPrintf(StdOut, L"\n");
+
+    WCHAR lpstrImageName[HEADER_STR_MAXLEN];
+    WCHAR lpstrPID[HEADER_STR_MAXLEN];
+    WCHAR lpstrSession[HEADER_STR_MAXLEN];
+    WCHAR lpstrMemUsage[HEADER_STR_MAXLEN];
+
+    LoadStringW(GetModuleHandle(NULL), IDS_HEADER_IMAGENAME, lpstrImageName, HEADER_STR_MAXLEN);
```
katahiromz, line 143: Use `GetModuleHandleW` and call it only once.

kernelbin: That makes sense. already adopted, thanks! 🤣

base/applications/cmdutils/tasklist/tasklist.c
kernelbin: I created a ```PrintResString``` to do this

base/applications/cmdutils/tasklist/tasklist.h
```diff
@@ -0,0 +1,36 @@
+#pragma once
```
HBelusca, line 1: Please add a file header notice like the one present in tasklist.c

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,356 @@
...
+}
+
+// Print a string from resource
+// if bAlignLeft == TRUE then aligned to left, otherwise aligned to right
+// MaxWidth is the width for printing.
+// The string will be truncated if it's longer than RES_STR_MAXLEN
+INT PrintResString(HINSTANCE hInstance, UINT uid, INT MaxWidth, BOOL bAlignLeft)
+{
+    WCHAR StringBuffer[RES_STR_MAXLEN];
+    LoadStringW(hInstance, uid, StringBuffer, RES_STR_MAXLEN);
```
HBelusca, line 44:
```suggestion
    LoadStringW(hInstance, uid, StringBuffer, _countof(StringBuffer));
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,356 @@
...
+        if (pSPI->NextEntryOffset == 0)
+            break;
+        pSPI = (PSYSTEM_PROCESS_INFORMATION)((LPBYTE)pSPI + pSPI->NextEntryOffset);
+    }
+
+    HeapFree(GetProcessHeap(), 0, ProcessInfoBuffer);
+    return TRUE;
+}
+
+INT GetArgumentType(WCHAR* argument)
```
HBelusca, line 260: Technically, here the argument is a NULL-terminated string (`PCWSTR`)

base/applications/cmdutils/tasklist/tasklist.h
```diff
@@ -0,0 +1,36 @@
...
+#define RES_STR_MAXLEN 64
+
+static WCHAR opHelp[] = L"?";
+static WCHAR opVerbose[] = L"v";
+static PWCHAR opList[] = {opHelp, opVerbose};
+
+#define OP_PARAM_INVALID -1
+
+#define OP_PARAM_HELP 0
+#define OP_PARAM_VERBOSE 1
```
HBelusca, line 30:
```suggestion
#define OP_PARAM_INVALID    -1
#define OP_PARAM_HELP       0
#define OP_PARAM_VERBOSE    1
```

base/applications/cmdutils/tasklist/tasklist.h
```diff
@@ -0,0 +1,36 @@
...
+#include <conutils.h>
+
+#include "resource.h"
+
+#define NT_SYSTEM_QUERY_MAX_RETRY 5
+
+#define COLUMNWIDTH_IMAGENAME 25
+#define COLUMNWIDTH_PID 8
+#define COLUMNWIDTH_SESSION 11
+#define COLUMNWIDTH_MEMUSAGE 12
```
HBelusca, line 19:
```suggestion
#define COLUMNWIDTH_IMAGENAME   25
#define COLUMNWIDTH_PID         8
#define COLUMNWIDTH_SESSION     11
#define COLUMNWIDTH_MEMUSAGE    12
```

base/applications/cmdutils/tasklist/tasklist.h
```diff
@@ -0,0 +1,36 @@
...
+#define OP_PARAM_INVALID -1
+
+#define OP_PARAM_HELP 0
+#define OP_PARAM_VERBOSE 1
+
+typedef NTSTATUS(NTAPI *NT_QUERY_SYSTEM_INFORMATION)(
+    SYSTEM_INFORMATION_CLASS SystemInformationClass,
+    PVOID SystemInformation,
+    ULONG InformationLength,
+    PULONG ResultLength);
```
HBelusca, line 36: Not needed anymore.
```suggestion
```

base/applications/cmdutils/tasklist/tasklist.h
```diff
@@ -0,0 +1,36 @@
...
+#define COLUMNWIDTH_IMAGENAME 25
+#define COLUMNWIDTH_PID 8
+#define COLUMNWIDTH_SESSION 11
+#define COLUMNWIDTH_MEMUSAGE 12
+
+#define RES_STR_MAXLEN 64
+
+static WCHAR opHelp[] = L"?";
+static WCHAR opVerbose[] = L"v";
+static PWCHAR opList[] = {opHelp, opVerbose};
```
HBelusca, line 25: Since you are not also using the "opXXX" variables individually explicitly in the rest of the code, maybe just consider doing something like:
```suggestion
static PCWSTR opList[] = {L"?", L"v"};
```

base/applications/cmdutils/tasklist/tasklist.h
HBelusca: Maybe add also a comment telling that these values should follow the same order as the elements in opList.

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,359 @@
...
+
+    ConPrintf(StdOut, L"\n");
+}
+
+BOOL EnumProcessAndPrint(BOOL bNoHeader)
+{
+    // Call NtQuerySystemInformation for the process information
+    ULONG ProcessInfoBufferLength = 0;
+    ULONG ResultLength = 0;
+    PBYTE ProcessInfoBuffer = 0;
```
katahiromz, line 174:
```suggestion
    PBYTE ProcessInfoBuffer = NULL;
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,359 @@
...
+    PrintSplitLine(COLUMNWIDTH_MEMUSAGE);
+
+    ConPrintf(StdOut, L"\n");
+}
+
+BOOL EnumProcessAndPrint(BOOL bNoHeader)
+{
+    // Call NtQuerySystemInformation for the process information
+    ULONG ProcessInfoBufferLength = 0;
+    ULONG ResultLength = 0;
```
katahiromz, line 173:
```suggestion
    ULONG ProcessInfoBufferLength = 0, ResultLength = 0;
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,359 @@
...
+        if (pSPI->NextEntryOffset == 0)
+            break;
+        pSPI = (PSYSTEM_PROCESS_INFORMATION)((LPBYTE)pSPI + pSPI->NextEntryOffset);
+    }
+
+    HeapFree(GetProcessHeap(), 0, ProcessInfoBuffer);
+    return TRUE;
+}
+
+INT GetArgumentType(WCHAR* argument)
```
katahiromz, line 263:
```suggestion
INT GetArgumentType(LPCWSTR argument)
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,359 @@
...
+    {
+        if (!wcsicmp(opList[i], argument))
+        {
+            return i;
+        }
+    }
+    return OP_PARAM_INVALID;
+}
+
+BOOL ProcessArguments(INT argc, WCHAR *argv[])
```
katahiromz, line 283:
```suggestion
BOOL ProcessArguments(INT argc, WCHAR **argv)
```

base/applications/cmdutils/tasklist/tasklist.h
katahiromz: OK.

base/applications/cmdutils/tasklist/tasklist.h
```diff
@@ -0,0 +1,36 @@
...
+
+#define WIN32_NO_STATUS
+#include <windows.h>
+#include <ntndk.h>
+
+#include <conutils.h>
+
+#include "resource.h"
+
+#define NT_SYSTEM_QUERY_MAX_RETRY 5
```
katahiromz, line 21: `NT_SYSTEM_QUERY_MAX_RETRY` is concealable detail. Move it to `tasklist.c`.

base/applications/cmdutils/tasklist/tasklist.h
```diff
@@ -0,0 +1,36 @@
...
+
+#define COLUMNWIDTH_IMAGENAME   25
+#define COLUMNWIDTH_PID         8
+#define COLUMNWIDTH_SESSION     11
+#define COLUMNWIDTH_MEMUSAGE    12
```
katahiromz, line 26: They are concealable.

base/applications/cmdutils/tasklist/tasklist.h
```diff
@@ -0,0 +1,36 @@
...
+
+#define RES_STR_MAXLEN 64
+
+// the strings in opList are the command-line options.
+// should always correspond with the defines below, in sequence (except OP_PARAM_INVALID)
+static PCWSTR opList[] = {L"?", L"nh"};
```
katahiromz, line 32: Don't place static variable in header. Move it into `tasklist.c`.

base/applications/cmdutils/tasklist/tasklist.h
```diff
@@ -0,0 +1,36 @@
...
+#include "resource.h"
+
+#define NT_SYSTEM_QUERY_MAX_RETRY 5
+
+#define COLUMNWIDTH_IMAGENAME   25
+#define COLUMNWIDTH_PID         8
+#define COLUMNWIDTH_SESSION     11
+#define COLUMNWIDTH_MEMUSAGE    12
+
+#define RES_STR_MAXLEN 64
```
katahiromz, line 28: Ditto.

base/applications/cmdutils/tasklist/tasklist.h
```diff
@@ -0,0 +1,36 @@
...
+
+// the strings in opList are the command-line options.
+// should always correspond with the defines below, in sequence (except OP_PARAM_INVALID)
+static PCWSTR opList[] = {L"?", L"nh"};
+
+#define OP_PARAM_INVALID    -1
+#define OP_PARAM_HELP       0
+#define OP_PARAM_NOHEADER    1
```
katahiromz, line 36: Concealable.

katahiromz (commented):
Some concealable definitions there are.

kernelbin:
> Some concealable definitions there are.

ah .... may I ask what does concealable means? Do you mean I should move these defines into ```tasklist.c```

base/applications/cmdutils/tasklist/tasklist.c
HBelusca: This is optional formatting.

base/applications/cmdutils/tasklist/tasklist.c
HBelusca: or PCWSTR if the rest of the code tends to use the latter, or its other versions: PWSTR, PWCHAR etc.

base/applications/cmdutils/tasklist/tasklist.h
HBelusca: Everything in tasklist.h is "concealable" in that sense, then, because it's only used within tasklist.h 🤣 
Actually having these defines here means one could "configure"/customize the app like those that have these config.h for that purpose ^^

katahiromz:
'Concealable' means 'it can hide'.

kernelbin:
> 'Concealable' means 'it can hide'. If the info should be private, then you have to hide.

I've already moved those things I think is concealable into ```tasklist.h```  (those static variables is an accident) 🤣

as for those ```COLUMNWIDTH_*``` defines, I agree with HBelusca. tasklist is not a library, and the purpose of ```tasklist.h``` is to configure tasklist, so there's no need to conceal it.
So I still left it there unmoved for now. 🤣 if you still think so we can discuss it further

katahiromz:
Ok, ready to go.

base/applications/cmdutils/tasklist/tasklist.h
```diff
@@ -0,0 +1,26 @@
...
+
+#pragma once
+
+#include <stdio.h>
+#include <stdlib.h>
+
+#define WIN32_NO_STATUS
+#include <windows.h>
+#include <ntndk.h>
+
```
katahiromz, line 16: Delete this empty line.

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,370 @@
...
+
+    do
+    {
+        Tmp /= 10;
+        PrintLength += 1;
+    } while (Tmp);
+
+    if (PrintLength > MaxWidth)
+    {
+        return FALSE;
```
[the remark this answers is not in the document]
kernelbin: assert sounds good! 🤣
but what should I return instead of a ```BOOL```?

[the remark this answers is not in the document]
katahiromz: VOID.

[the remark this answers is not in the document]
HBelusca: Otherwise, failing gracefully maybe? anyone?

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,370 @@
...
+    return TRUE;
+}
+
+// Print memory size using appropriate unit, with comma-separated number, aligned to right.
+// MaxWidth is the width for printing.
+// StartingUnit is the minimum memory unit used, 0 for Byte, 1 for KB, 2 for MB ...
+// return FALSE when failed to format.
+BOOL PrintMemory(SIZE_T MemorySizeByte, INT MaxWidth, INT StartingUnit)
+{
+    WCHAR *MemoryUnit[] = {L"B", L"K", L"M", L"G", L"T", L"P"};
```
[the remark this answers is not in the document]
HBelusca: PCWSTR

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,370 @@
...
+
+                INT Mod = 1;
+                for (INT i = 0; i < PrintLength - 1; i++)
+                {
+                    Mod *= 10;
+                }
+
+                for (INT i = PrintLength - 1; i >= 0; i--)
+                {
+                    ConPrintf(StdOut, L"%d", MemorySize / Mod);
```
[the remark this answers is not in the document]
HBelusca: That assert had to be done before (when Mod is calculated)

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,370 @@
...
+    } while (Tmp);
+
+    if (PrintLength > MaxWidth)
+    {
+        return FALSE;
+    }
+    ConPrintf(StdOut, L"%*lld", MaxWidth, Number);
+
+    return TRUE;
+}
```
HBelusca, line 87:
```suggestion
VOID PrintNum(LONGLONG Number, INT MaxWidth)
{
    /* NOTE: The maximum length of the printed number will be 20 characters
     * (including the sign for signed, or with another digit for unsigned). */
    ConPrintf(StdOut, L"%*lld", MaxWidth, Number);
}
```
Also as a user I don't want a tool that crashes on me just because it's unable to display long yet valid information.

base/applications/cmdutils/tasklist/tasklist.h
```diff
@@ -0,0 +1,26 @@
...
+#include <conutils.h>
+
```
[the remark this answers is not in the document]
HBelusca: There is no need for wine/debug, this is not a Wine tool.

base/applications/cmdutils/tasklist/tasklist.c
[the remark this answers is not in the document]
HBelusca: Also as a user I don't want a tool that crashes on me because it cannot display its information.

HBelusca (changes requested):

base/applications/cmdutils/tasklist/tasklist.c
kernelbin: there's a problem with this. It won't truncate the number if it's longer than ```MaxWidth```. should I write some code to truncate the number?

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,370 @@
...
+    if (MaxWidth <= 0)
+    {
+        return FALSE;
+    }
+
+    SIZE_T MemorySize = MemorySizeByte;
+
+    for (INT i = 0; i < _countof(MemoryUnit); i++)
+    {
+        if (i >= StartingUnit)
```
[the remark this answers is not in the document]
kernelbin: then I also have to do ```MemorySize >>= 10;``` before continuing, I don't want to write this twice 🤣 (should I write this in the for loop instead?)
by the way, I'm wondering what does the unit ```K``` in tasklist refers to?
AFAIN it can refer to ```KB``` or ```KiB```, and that's two different units.....

katahiromz:
Use DPRINT1 instead of assert?

base/applications/cmdutils/tasklist/tasklist.c
HBelusca: If it truncates the number it won't report accurate information as well. Not sure this is wanted.

base/applications/cmdutils/tasklist/tasklist.c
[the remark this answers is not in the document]
HBelusca: The notation of these prefixes is also language/locale-dependent.

base/applications/cmdutils/tasklist/tasklist.c
kernelbin: I observed ```tasklist``` on windows under TABLE mode (```/fo table```, also the default format), the column width is big enough that the number is almost impossible to need to be truncated.

But the image name column does truncate the file name if it's too long (while won't in ```list``` and ```csv``` format) 

so I think possibly it's more important to keep it well-formatted under table mode

that's why I think truncate the number is better, does this make sense? 😂

kernelbin:
> Use DPRINT1 instead of assert?

ah, may I ask what's ```DPRINT1```?

base/applications/cmdutils/tasklist/tasklist.c
HBelusca: While truncating strings can make sense, truncating numbers don't.

kernelbin: then fine 😂
after all, it's almost impossible to have these number big enough to need to be truncated

base/applications/cmdutils/tasklist/tasklist.c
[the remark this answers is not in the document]
kernelbin: prefixes ?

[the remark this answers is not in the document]
HBelusca: Errm I meant these suffixes (the K, M, G, etc...)

base/applications/cmdutils/tasklist/tasklist.c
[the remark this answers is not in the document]
kernelbin: actually, it's impossible for ```Mod``` to be zero (at least inside the for-loop) so there is no need to assert I think😂

base/applications/cmdutils/tasklist/tasklist.c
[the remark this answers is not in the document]
kernelbin: the unit is locale-dependent?! 
Oh, I don't know this before...
should I move them into the resources file or what?

base/applications/cmdutils/tasklist/tasklist.h
kernelbin: Ah, I'm trying to classify these headers 🤣. conutils.h, AFAIR is a ReactOS stuff? so I put it alone.

HBelusca: > conutils.h, AFAIR is a ReactOS stuff?

Yes it is.

base/applications/cmdutils/tasklist/tasklist.c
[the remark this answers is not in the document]
katahiromz: `<shlwapi.h>` has `StrFormatByteSizeW` function.
https://docs.microsoft.com/ja-jp/windows/win32/api/shlwapi/nf-shlwapi-strformatbytesizew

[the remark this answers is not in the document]
HBelusca: Be careful however to not have core command-line utilities depend on optional GUI shell features though.

katahiromz:
`DPRINT1` is `<debug.h>`'s debugging function of `printf`-like syntax.

base/applications/cmdutils/tasklist/tasklist.c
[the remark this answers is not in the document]
katahiromz: StrFormatByteSizeW has no GUI. Should we use StrFormatByteSizeW? Otherwise what we use?

base/applications/cmdutils/tasklist/tasklist.h
katahiromz: OK.

base/applications/cmdutils/tasklist/tasklist.h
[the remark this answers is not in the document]
katahiromz: Ok.

base/applications/cmdutils/tasklist/tasklist.c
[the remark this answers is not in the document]
kernelbin: ah, I've recently checked the resources that the  ```tasklist``` on windows have included, and I have found this:
```
    253, "%s K"
```
guess this is what tasklist used to format the memory usage string? BTW I didn't found any other strings like M,G,T.
possibly it's supposed to only display memory in KB/KiB?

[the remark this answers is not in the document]
HBelusca: Yes do that. I would suggest (because some people could invoke "compatibility" with the Windows tasklist and tools that could parse its output) to default to only displaying the numbers in KB (multiple of 1024), and only if you specify a ROS-specific switch on the command-line, would you format the output in KB/MB/GB etc... depending on the size.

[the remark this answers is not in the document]
kernelbin: > Yes do that. I would suggest (because some people could invoke "compatibility" with the Windows tasklist and tools that could parse its output) to default to only displaying the numbers in KB (multiple of 1024), and only if you specify a ROS-specific switch on the command-line, would you format the output in KB/MB/GB etc... depending on the size.

Ok, I will refactor it tonight.

kernelbin:
Is it looks ok now?🤣

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,348 @@
...
+    if (MaxWidth <= 0 || !hInstance)
+    {
+        assert(0);
+        return FALSE;
+    }
+
+    SIZE_T MemorySize = MemorySizeByte >> 10;
+
+    WCHAR NumberString[27] = { 0 }; // length 26 is enough to display ULLONG_MAX in decimal with comma, one more for zero-terminated.
+    assert(sizeof(SIZE_T) <= 8);
```
HBelusca, line 87:
```suggestion
    C_ASSERT(sizeof(SIZE_T) <= 8);
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,348 @@
...
+            return i;
+        }
+    }
+    return OP_PARAM_INVALID;
+}
+
+BOOL ProcessArguments(INT argc, WCHAR **argv)
+{
+    BOOL bHasHelp = FALSE, bHasNoHeader = FALSE;
+    for (INT i = 1; i < argc; i++)
```
HBelusca, line 276:
```suggestion

    for (INT i = 1; i < argc; i++)
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,348 @@
...
+    if (!NT_SUCCESS(Status))
+    {
+        // tried NT_SYSTEM_QUERY_MAX_RETRY times, or failed with some other reason
+        ConResMsgPrintf(StdOut, 0, IDS_ENUM_FAILED);
+        HeapFree(GetProcessHeap(), 0, ProcessInfoBuffer);
+        return FALSE;
+    }
+
+    HINSTANCE hInstance = GetModuleHandleW(NULL);
+
```
HBelusca, line 224:
```suggestion
    HINSTANCE hInstance = GetModuleHandleW(NULL);
    assert(hInstance);

```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,348 @@
...
+// if bAlignLeft == TRUE then aligned to left, otherwise aligned to right
+// MaxWidth is the width for printing.
+// The string WILL be truncated if it's longer than RES_STR_MAXLEN
+INT PrintResString(HINSTANCE hInstance, UINT uid, INT MaxWidth, BOOL bAlignLeft)
+{
+    if (!hInstance)
+    {
+        assert(0);
+        return 0;
+    }
```
HBelusca, line 58:
```suggestion
        return 0;
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,348 @@
...
+// Print memory size using KB as unit, with comma-separated number, aligned to right.
+// MaxWidth is the width for printing.
+// the number WILL be truncated if it's longer than MaxWidth
+BOOL PrintMemory(SIZE_T MemorySizeByte, INT MaxWidth, HINSTANCE hInstance)
+{
+    if (MaxWidth <= 0 || !hInstance)
+    {
+        assert(0);
+        return FALSE;
+    }
```
HBelusca, line 82:
```suggestion
        return FALSE;
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,348 @@
...
+    return TRUE;
+}
+
+VOID PrintHeader(HINSTANCE hInstance)
+{
+    if (!hInstance)
+    {
+        assert(0);
+        return;
+    }
```
HBelusca, line 137:
```suggestion
        return;

```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,348 @@
...
+#define OP_PARAM_HELP       0
+#define OP_PARAM_NOHEADER    1
+
+// the max string length PrintResString can handle
+#define RES_STR_MAXLEN 64
+
+// Print split line
+VOID PrintSplitLine(INT Length)
+{
+    for (INT i = 0; i < Length; i++)
```
HBelusca, line 25:
```suggestion
    for (; Length; Length--)
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,348 @@
+/*
+ * PROJECT:     ReactOS Tasklist Command
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Displays a list of currently running processes on the computer.
+ * COPYRIGHT:   Copyright 2020 He Yang (1160386205@qq.com)
+ */
+
+#include "tasklist.h"
+
+
```
HBelusca, line 10:
```suggestion
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,349 @@
...
+// the max string length PrintResString can handle
+#define RES_STR_MAXLEN 64
+
+// Print split line
+VOID PrintSplitLine(INT Length)
+{
+    for (INT i = 0; i < Length; i++)
+    {
+        ConPrintf(StdOut, L"=");
```
HBelusca, line 27:
```suggestion
        ConPuts(StdOut, L"=");
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,349 @@
...
+    }
+}
+
+// Print spaces
+VOID PrintSpace(INT Length)
+{
+    for (; Length; Length--)
+    {
+        ConPrintf(StdOut, L" ");
```
HBelusca, line 36:
```suggestion
        ConPuts(StdOut, L" ");
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,349 @@
...
+    }
```
HBelusca, line 37: Note that only when filling using spaces, the shorter printf expression works too:
```suggestion
    ConPrintf(StdOut, L"%*s", Length, L"");
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,349 @@
...
+
+#define OP_PARAM_INVALID    -1
+#define OP_PARAM_HELP       0
+#define OP_PARAM_NOHEADER    1
+
+// the max string length PrintResString can handle
+#define RES_STR_MAXLEN 64
+
+// Print split line
+VOID PrintSplitLine(INT Length)
```
HBelusca, line 23: Can your lengths be actually negative?

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,349 @@
...
+    }
+    PrintResString(hInstance, IDS_HEADER_IMAGENAME, COLUMNWIDTH_IMAGENAME, TRUE);
+    PrintSpace(1);
+    PrintResString(hInstance, IDS_HEADER_PID,       COLUMNWIDTH_PID,       FALSE);
+    PrintSpace(1);
+    PrintResString(hInstance, IDS_HEADER_SESSION,   COLUMNWIDTH_SESSION,   FALSE);
+    PrintSpace(1);
+    PrintResString(hInstance, IDS_HEADER_MEMUSAGE,  COLUMNWIDTH_MEMUSAGE,  FALSE);
+
+    ConPrintf(StdOut, L"\n");
```
HBelusca, line 146:
```suggestion
    ConPuts(StdOut, L"\n");
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,349 @@
...
+
+    PrintSplitLine(COLUMNWIDTH_IMAGENAME);
+    PrintSpace(1);
+    PrintSplitLine(COLUMNWIDTH_PID);
+    PrintSpace(1);
+    PrintSplitLine(COLUMNWIDTH_SESSION);
+    PrintSpace(1);
+    PrintSplitLine(COLUMNWIDTH_MEMUSAGE);
+
+    ConPrintf(StdOut, L"\n");
```
HBelusca, line 156:
```suggestion
    ConPuts(StdOut, L"\n");
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,349 @@
...
+                                                     ProcessInfoBuffer,
+                                                     ProcessInfoBufferLength);
+            if (NewProcessInfoBuffer)
+            {
+                ProcessInfoBuffer = NewProcessInfoBuffer;
+            }
+            else
+            {
+                // out of memory ?
+                ConResMsgPrintf(StdOut, 0, IDS_OUT_OF_MEMORY);
```
HBelusca, line 188:
```suggestion
                ConResMsgPrintf(StdErr, 0, IDS_OUT_OF_MEMORY);
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,349 @@
...
+                return FALSE;
+            }
+        }
+        else
+        {
+            ProcessInfoBuffer = HeapAlloc(GetProcessHeap(), 0, ProcessInfoBufferLength);
+            if (!ProcessInfoBuffer)
+            {
+                // out of memory ?
+                ConResMsgPrintf(StdOut, 0, IDS_OUT_OF_MEMORY);
```
HBelusca, line 199:
```suggestion
                ConResMsgPrintf(StdErr, 0, IDS_OUT_OF_MEMORY);
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,349 @@
...
+        if (Status != STATUS_INFO_LENGTH_MISMATCH)
+        {
+            break;
+        }
+    }
+
+    if (!NT_SUCCESS(Status))
+    {
+        // tried NT_SYSTEM_QUERY_MAX_RETRY times, or failed with some other reason
+        ConResMsgPrintf(StdOut, 0, IDS_ENUM_FAILED);
```
HBelusca, line 218:
```suggestion
        ConResMsgPrintf(StdErr, 0, IDS_ENUM_FAILED);
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,349 @@
...
+        HeapFree(GetProcessHeap(), 0, ProcessInfoBuffer);
+        return FALSE;
+    }
+
+    HINSTANCE hInstance = GetModuleHandleW(NULL);
+
+    ConPrintf(StdOut, L"\n");
```
HBelusca, line 226:
```suggestion
    ConPuts(StdOut, L"\n");
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,349 @@
...
+    {
+        PrintString(pSPI->UniqueProcessId ? pSPI->ImageName.Buffer : L"System Idle Process", COLUMNWIDTH_IMAGENAME, TRUE);
+        PrintSpace(1);
+        PrintNum((LONGLONG)(INT_PTR)pSPI->UniqueProcessId, COLUMNWIDTH_PID);
+        PrintSpace(1);
+        PrintNum((ULONGLONG)pSPI->SessionId, COLUMNWIDTH_SESSION);
+        PrintSpace(1);
+        PrintMemory(pSPI->WorkingSetSize, COLUMNWIDTH_MEMUSAGE, hInstance);
+
+        ConPrintf(StdOut, L"\n");
```
HBelusca, line 245:
```suggestion
        ConPuts(StdOut, L"\n");
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,350 @@
...
+            }
+        }
+    }
+
+    if (bHasHelp)
+    {
+        if (argc > 2) // any parameters other than -? is specified
+        {
+            ConResMsgPrintf(StdOut, 0, IDS_INVALID_SYNTAX);
+            ConResMsgPrintf(StdOut, 0, IDS_USAGE);
```
HBelusca, line 323:
```suggestion
            ConResMsgPrintf(StdErr, 0, IDS_INVALID_SYNTAX);
            ConResMsgPrintf(StdErr, 0, IDS_USAGE);
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,350 @@
...
+                    return FALSE;
+                }
+                bHasNoHeader = TRUE;
+                break;
+            }
+            case OP_PARAM_INVALID:
+            default:
+            {
+                ConResMsgPrintf(StdOut, 0, IDS_INVALID_OPTION);
+                ConResMsgPrintf(StdOut, 0, IDS_USAGE);
```
HBelusca, line 312:
```suggestion
                ConResMsgPrintf(StdErr, 0, IDS_INVALID_OPTION);
                ConResMsgPrintf(StdErr, 0, IDS_USAGE);
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,350 @@
...
+                bHasHelp = TRUE;
+                break;
+            }
+            case OP_PARAM_NOHEADER:
+            {
+                if (bHasNoHeader)
+                {
+                    // -nh already specified
+                    ConResMsgPrintf(StdOut, 0, IDS_PARAM_TOO_MUCH, argv[i], 1);
+                    ConResMsgPrintf(StdOut, 0, IDS_USAGE);
```
HBelusca, line 302:
```suggestion
                    ConResMsgPrintf(StdErr, 0, IDS_PARAM_TOO_MUCH, argv[i], 1);
                    ConResMsgPrintf(StdErr, 0, IDS_USAGE);
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,350 @@
...
+
+        switch (Argument)
+        {
+            case OP_PARAM_HELP:
+            {
+                if (bHasHelp)
+                {
+                    // -? already specified
+                    ConResMsgPrintf(StdOut, 0, IDS_PARAM_TOO_MUCH, argv[i], 1);
+                    ConResMsgPrintf(StdOut, 0, IDS_USAGE);
```
HBelusca, line 290:
```suggestion
                    ConResMsgPrintf(StdErr, 0, IDS_PARAM_TOO_MUCH, argv[i], 1);
                    ConResMsgPrintf(StdErr, 0, IDS_USAGE);
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,350 @@
...
+            PBYTE NewProcessInfoBuffer = HeapReAlloc(GetProcessHeap(), 0,
+                                                     ProcessInfoBuffer,
+                                                     ProcessInfoBufferLength);
+            if (NewProcessInfoBuffer)
+            {
+                ProcessInfoBuffer = NewProcessInfoBuffer;
+            }
+            else
+            {
+                // out of memory ?
```
HBelusca, line 187:
```suggestion
                // out of memory
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,350 @@
...
+                HeapFree(GetProcessHeap(), 0, ProcessInfoBuffer);
+                return FALSE;
+            }
+        }
+        else
+        {
+            ProcessInfoBuffer = HeapAlloc(GetProcessHeap(), 0, ProcessInfoBufferLength);
+            if (!ProcessInfoBuffer)
+            {
+                // out of memory ?
```
HBelusca, line 198:
```suggestion
                // out of memory
```

base/applications/cmdutils/tasklist/tasklist.c
kernelbin: Ah, I thought ```ConPuts``` will append a linefeed after the string just like what ```puts``` does but it actually didn't 🤣
I'm misled by its name

HBelusca: It's like the "fputs" (The idea behind is that all these f*** CRT functions: fputs, fprintf... act on an explicit stream, much like these Con*** functions, and the fputs doesn't add an extra newline.)

base/applications/cmdutils/tasklist/tasklist.c
kernelbin: Brilliant! I'm just thinking about is calling ```ConPrintf``` ```n``` times too performance costing

base/applications/cmdutils/tasklist/tasklist.c
kernelbin: hmmm, does ```tasklist``` on windows actually printing these messages into StdErr?

base/applications/cmdutils/tasklist/tasklist.c
HBelusca: yep! ConPuts in loop is less costly; however that trick I mentioned only works when you fill with spaces. For any other characters a loop of some sort is necessary.

base/applications/cmdutils/tasklist/tasklist.c
kernelbin: oh yes it's true

kernelbin:
there are much more imperfects in my code than I imaged, I should check it more carefully next time 🤣🤣🤣

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,338 @@
...
+// Print spaces
+VOID PrintSpace(UINT Length)
+{
+    ConPrintf(StdOut, L"%*ls", (INT)Length, L"");
+}
+
+// Print a string.
+// if bAlignLeft == TRUE then aligned to left, otherwise aligned to right
+// MaxWidth is the width for printing.
+INT PrintString(LPCWSTR String, INT MaxWidth, BOOL bAlignLeft)
```
HBelusca, line 39:
```suggestion
INT PrintString(LPCWSTR String, UINT MaxWidth, BOOL bAlignLeft)
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,338 @@
...
+{
+    return ConPrintf(StdOut, bAlignLeft ? L"%-*.*ls" : L"%*.*ls", MaxWidth, MaxWidth, String);
```
HBelusca, line 41: Maybe casts are needed now for MaxWidth? (Does the compiler give warnings if they stay UINT?)

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,338 @@
...
+}
+
+// Print a string from resource
+// if bAlignLeft == TRUE then aligned to left, otherwise aligned to right
+// MaxWidth is the width for printing.
+// The string WILL be truncated if it's longer than RES_STR_MAXLEN
+INT PrintResString(HINSTANCE hInstance, UINT uid, INT MaxWidth, BOOL bAlignLeft)
```
HBelusca, line 48:
```suggestion
INT PrintResString(HINSTANCE hInstance, UINT uID, UINT MaxWidth, BOOL bAlignLeft)
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,338 @@
...
+{
+    if (!hInstance)
+        return 0;
+    
+    WCHAR StringBuffer[RES_STR_MAXLEN];
+    LoadStringW(hInstance, uid, StringBuffer, _countof(StringBuffer));
```
HBelusca, line 54:
```suggestion
    LoadStringW(hInstance, uID, StringBuffer, _countof(StringBuffer));
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,338 @@
...
+    return PrintString(StringBuffer, MaxWidth, bAlignLeft);
+}
+
+// Print a number, aligned to right.
+// MaxWidth is the width for printing.
+// the number WILL NOT be truncated if it's longer than MaxWidth
+VOID PrintNum(LONGLONG Number, INT MaxWidth)
```
HBelusca, line 61:
```suggestion
VOID PrintNum(LONGLONG Number, UINT MaxWidth)
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,338 @@
...
+{
+    ConPrintf(StdOut, L"%*lld", MaxWidth, Number);
+}
+
+// Print memory size using KB as unit, with comma-separated number, aligned to right.
+// MaxWidth is the width for printing.
+// the number WILL be truncated if it's longer than MaxWidth
+BOOL PrintMemory(SIZE_T MemorySizeByte, INT MaxWidth, HINSTANCE hInstance)
```
HBelusca, line 69:
```suggestion
BOOL PrintMemory(SIZE_T MemorySizeByte, UINT MaxWidth, HINSTANCE hInstance)
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,338 @@
...
+
+    SIZE_T MemorySize = MemorySizeByte >> 10;
+
+    WCHAR NumberString[27] = { 0 }; // length 26 is enough to display ULLONG_MAX in decimal with comma, one more for zero-terminated.
+    C_ASSERT(sizeof(SIZE_T) <= 8);
+
+    PWCHAR pNumberStr = NumberString;
+
+    // calculate the length
+    INT PrintLength = 0;
```
HBelusca, line 82:
```suggestion
    UINT PrintLength = 0;
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,338 @@
...
+    SIZE_T Tmp = MemorySize;
+    do
+    {
+        Tmp /= 10;
+        PrintLength += 1;
```
HBelusca, line 87:
```suggestion
        PrintLength++;
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,338 @@
...
+{
+    if (argument[0] != L'/' && argument[0] != L'-')
+    {
+        return OP_PARAM_INVALID;
+    }
+    argument++;
+
+    for (INT i = 0; i < _countof(opList); i++)
+    {
+        if (!wcsicmp(opList[i], argument))
```
HBelusca, line 254:
```suggestion
        if (!_wcsicmp(opList[i], argument))
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,338 @@
...
+        MemorySize %= Mod;
+        pNumberStr++;
+        
+        if (i && i % 3 == 0)
+        {
+            *pNumberStr = L',';
+            pNumberStr++;
+        }
+        Mod /= 10;
+    }
```
HBelusca, line 108: By the way I want to mention something: This formatting is also locale-dependent. For example when running in French locale:
image
Note: No comma for thousand delimiters, and the unit is indeed localized (re. our past discussion, so I think that if in tasklist resources they just have "%s K" it may be a fallback)
See:
https://docs.microsoft.com/en-us/globalization/locale/number-formatting-in-win32
https://docs.microsoft.com/en-us/windows/win32/api/winnls/nf-winnls-getnumberformatw
and
https://git.reactos.org/?p=reactos.git;a=blob;f=base/applications/mscutils/eventvwr/eventvwr.c;hb=ea26bef01de20d5c140a426397627089892356a0#l1113

HBelusca: If this looks too complicated this could wait for a separate PR.

base/applications/cmdutils/tasklist/tasklist.c
HBelusca: Please implement this suggestion as well.

HBelusca (commented):
Last-minute considerations, otherwise it's LGTM!

base/applications/cmdutils/tasklist/tasklist.c
kernelbin: compiler gives no warnings, but the document (printf's) says if it's negative, it will be ignored. 🤣
(IMO, even if it's not negative, like 0x7fffffff, it might already is a disaster)

kernelbin: possibly just left it here and add a comment saying this function will ignore this param if it's negative?

HBelusca: Then you can just go ahead and convert all your width variables to unsigned ones. This will make the code self-documented (namely that no width can be negative)

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,338 @@
+/*
+ * PROJECT:     ReactOS Tasklist Command
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Displays a list of currently running processes on the computer.
+ * COPYRIGHT:   Copyright 2020 He Yang (1160386205@qq.com)
```
katahiromz, line 5:
```suggestion
 * COPYRIGHT:   Copyright 2021 He Yang (1160386205@qq.com)
```

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,370 @@
...
+// StartingUnit is the minimum memory unit used, 0 for Byte, 1 for KB, 2 for MB ...
+// return FALSE when failed to format.
+BOOL PrintMemory(SIZE_T MemorySizeByte, INT MaxWidth, INT StartingUnit)
+{
+    WCHAR *MemoryUnit[] = {L"B", L"K", L"M", L"G", L"T", L"P"};
+
+    MaxWidth -= 2; // a space and a unit will occupy 2 length.
+    if (MaxWidth <= 0)
+    {
+        return FALSE;
```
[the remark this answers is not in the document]
katahiromz: Rejected. OK.

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,338 @@
...
+        PrintLength++;
+    } while (Tmp);
+
+    INT Mod = 1;
+    for (INT i = 0; i < PrintLength - 1; i++)
+    {
+        Mod *= 10;
+    }
+
+    for (INT i = PrintLength - 1; i >= 0; i--)
```
katahiromz, line 96: for (UINT i = PrintLength - 1; i > 0;)
{
i--;

base/applications/cmdutils/tasklist/tasklist.c
kernelbin: ok!

base/applications/cmdutils/tasklist/tasklist.c
kernelbin: I'm also planning to add other format support later (`/fo` option), possibly I can fix this in that PR together🤣

base/applications/cmdutils/tasklist/tasklist.c
kernelbin: `UINT` makes sense, but may I ask why changing the for loop? 🤣

kernelbin: oh, I got it! otherwise, it will loop infinitely 🤣🤣🤣

base/applications/cmdutils/tasklist/tasklist.c
```diff
@@ -0,0 +1,335 @@
...
+            pNumberStr++;
+        }
+    }
+
+    WCHAR FormatStr[RES_STR_MAXLEN];
+    LoadStringW(hInstance, IDS_MEMORY_STR, FormatStr, _countof(FormatStr));
+
+    WCHAR String[RES_STR_MAXLEN + _countof(NumberString)];
+    
+    swprintf(String, FormatStr, NumberString);
```
katahiromz, line 112: `swprintf` with no buffer overflow check is not secure.

base/applications/cmdutils/tasklist/tasklist.c
HBelusca: At that stage yes a separate PR for the other new features would be better.

base/applications/cmdutils/tasklist/tasklist.c
katahiromz:
```
int swprintf_s(
   wchar_t *buffer,
   size_t sizeOfBuffer,
   const wchar_t *format,
   ...
);
```

kernelbin: actually, I'm confused when I wrote this down yesterday.
this is the declaration of `swprint` I found in UCRT (and all other documents I can found)
```
int swprintf( wchar_t *buffer, size_t bufsz,
              const wchar_t* format, ... );
```

you can see the second param is `bufsz`, controlling maximum character could be written

but this is what I found in ReactOS: (at `reactos\sdk\include\crt\stdio.h`)
```
int __cdecl
    swprintf(
    _Out_ wchar_t*,
    _Printf_format_string_ const wchar_t*,
    ...);
```

the second param is missing, and I found another function pretty similar to `swprintf` I found elsewhere called `_swprintf_c`, which I failed to find any documents about it (possibly it's a ReactOS stuff?)

do you know the reason for this? 🤣 Is this a mistake when writing CRT for ReactOS, or does this having any purpose?

HBelusca: > swprintf conforms to the ISO C Standard, which requires the second parameter, count, of type size_t. To force the old nonstandard behavior, define _CRT_NON_CONFORMING_SWPRINTFS. In a future version, the old behavior may be removed, so code should be changed to use the new conformant behavior.

(from MSDN)

Our code uses the old style (from WIn2003 and below), i.e. the so-called "nonstandard behaviour".

kernelbin: I see...  then I guess using `swprintf_s`, just as what Hirofumi said, is better 🤣
seems this function doesn't have a diverge on standard/non-standard behaviour

katahiromz:
```txt
/base/applications/cmdutils/tasklist/tasklist.c:112: undefined reference to `swprintf_s'
```
Sorry, compilation is failing...

HBelusca: > Sorry, compilation is failing... C11 is requirement.

Huh please NO. Just include the correct headers and it'll be fixed.

HBelusca: You can use as well _snwprintf() (I would say I would prefer that one since it remains CRT function).

kernelbin: actually, I think it doesn't matter whether if the function is in CRT or not, we've already used tons of WinAPI even NT API in this program 🤣

HBelusca (approved):

katahiromz (approved):

PR #3441, opened 2021-01-31
[SCSIPORT] Append GEN_SCSIADAPTER compatible ID for legacy adapters

binarymaster:
Fixes UniATA root SCSI devices detection in the Device Manager. **IMPORTANT:** This depends on PR #3437.

image

Based on the description of `GEN_SCSIADAPTER` from Windows `pnpscsi.inf` (see also [this reference](https://community.osr.com/discussion/41967/installing-isa-scsi-miniport-driver-through-f6-on-fresh-install-of-windows-2000-problem)):
```
;*********************************
;Generic scsi adpater - compatible ID match on GEN_SCSIADAPTER
;as a method of making all non-pnp miniports have their adapters
;show up as scsi adapters in devmgr. Scsiport adds this id to
;legacy detected adapters.
```

See also chat messages:
- https://chat.reactos.org/reactos/pl/qii1w36wu7yrxyukh9b5dzwnje
- https://chat.reactos.org/reactos/pl/f5tanc9nfjdb5m17nhrq33994a

JIRA issue: [CORE-17398](https://jira.reactos.org/browse/CORE-17398)

### DIFF
--- a/drivers/storage/port/scsiport/fdo.c
+++ b/drivers/storage/port/scsiport/fdo.c
@@ -705,6 +705,51 @@ FdoHandleDeviceRelations(
     return IoCallDriver(PortExtension->Common.LowerDevice, Irp);
 }
 
+static
+NTSTATUS
+FdoHandleQueryCompatibleId(
+    _Inout_ PZZWSTR* PwIds)
+{
+    static WCHAR GenScsiAdapterId[] = L"GEN_SCSIADAPTER";
+    PWCHAR Ids = *PwIds, NewIds;
+    ULONG Length = 0;
+
+    if (Ids)
+    {
+        /* Calculate the length of existing MULTI_SZ value line by line */
+        while (*Ids)
+        {
+            Ids += wcslen(Ids) + 1;
+        }
+        Length = Ids - *PwIds;
+        Ids = *PwIds;
+    }
+
+    /* New MULTI_SZ with added identifier and finalizing zeros */
+    NewIds = ExAllocatePoolZero(PagedPool,
+                                Length * sizeof(WCHAR) + sizeof(GenScsiAdapterId) + sizeof(UNICODE_NULL),
+                                TAG_SCSIPORT);
+    if (!NewIds)
+    {
+        return STATUS_INSUFFICIENT_RESOURCES;
+    }
+
+    if (Length)
+    {
+        RtlCopyMemory(NewIds, Ids, Length * sizeof(WCHAR));
+    }
+    RtlCopyMemory(&NewIds[Length], GenScsiAdapterId, sizeof(GenScsiAdapterId));
+
+    /* Finally replace identifiers */
+    if (Ids)
+    {
+        ExFreePool(Ids);
+    }
+    *PwIds = NewIds;
+
+    return STATUS_SUCCESS;
+}
+
 NTSTATUS
 FdoDispatchPnp(
     _In_ PDEVICE_OBJECT DeviceObject,
@@ -731,6 +776,17 @@ FdoDispatchPnp(
         case IRP_MN_QUERY_DEVICE_RELATIONS:
         {
             return FdoHandleDeviceRelations(portExt, Irp);
+        }
+        case IRP_MN_QUERY_ID:
+        {
+            if (ioStack->Parameters.QueryId.IdType == BusQueryCompatibleIDs)
+            {
+                Irp->IoStatus.Information = 0;
+                IoForwardIrpSynchronously(portExt->Common.LowerDevice, Irp);
+                status = FdoHandleQueryCompatibleId((PZZWSTR*)&Irp->IoStatus.Information);
+                break;
+            }
+            // otherwise fall through the default case
         }
         default:
         {
--- a/media/inf/scsi.inf
+++ b/media/inf/scsi.inf
@@ -26,6 +26,7 @@ DefaultDestDir = 12
 %GenericMfg% = GenericMfg
 
 [GenericMfg]
+%GEN_SCSIADAPTER.DeviceDesc% = NO_DRV,,GEN_SCSIADAPTER
 %PCI\VEN_104B&CC_0100.DeviceDesc% = BusLogic_Inst,PCI\VEN_104B&CC_0100
 
 ;----------------------------- ScsiPort Driver ----------------------------
@@ -39,6 +40,13 @@ StartType     = 0
 ErrorControl  = 0
 ServiceBinary = %12%\scsiport.sys
 LoadOrderGroup = SCSI Port
+
+;---------------------------- NO DRIVER REQ -----------------------------
+
+[NO_DRV]
+
+[NO_DRV.Services]
+AddService = , 0x00000002
 
 ;----------------------------- BusLogic Driver ----------------------------
 
@@ -66,6 +74,7 @@ ReactOS = "ReactOS Team"
 SCSIClassName = "SCSI and RAID Controllers"
 
 GenericMfg = "(Standard SCSI and RAID controllers)"
+GEN_SCSIADAPTER.DeviceDesc = "SCSI/RAID Host Controller"
 PCI\VEN_104B&CC_0100.DeviceDesc = "BusLogic SCSI Controller"
 
 [Strings.0405]
@@ -120,6 +129,7 @@ ReactOS = "Команда ReactOS"
 SCSIClassName = "SCSI и RAID контроллеры"
 
 GenericMfg = "(Стандартные SCSI и RAID контроллеры)"
+GEN_SCSIADAPTER.DeviceDesc = "SCSI/RAID хост-контроллер"
 PCI\VEN_104B&CC_0100.DeviceDesc = "Контроллер BusLogic SCSI"
 
 [Strings.041B]


media/inf/scsi.inf
```diff
@@ -26,6 +26,7 @@ DefaultDestDir = 12
 %GenericMfg% = GenericMfg
 
 [GenericMfg]
+%GEN_SCSIADAPTER.DeviceDesc% = NO_DRV,,GEN_SCSIADAPTER
 %PCI\VEN_104B&CC_0100.DeviceDesc% = BusLogic_Inst,PCI\VEN_104B&CC_0100
```
Extravert-ir, line 30: I'd say this is not a place for BusLogic driver

Extravert-ir (changes requested):

drivers/storage/port/scsiport/fdo.c
```diff
@@ -705,6 +705,98 @@ FdoHandleDeviceRelations(
...
+    }
+    if (Length > 0)
+    {
+        /* Add zero separator to the length */
+        Length++;
+    }
+
+    /* New MULTI_SZ with added identifier and finalizing zeros */
+    NewIds = ExAllocatePoolZero(PagedPool,
+                                (Length + wcslen(GenScsiAdapterId) + 2) * sizeof(WCHAR),
```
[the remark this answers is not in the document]
Extravert-ir: and +4 ofc

drivers/storage/port/scsiport/fdo.c
```diff
@@ -705,6 +705,98 @@ FdoHandleDeviceRelations(
...
+    ULONG i;
+
+    Ids = (PWCHAR)Irp->IoStatus.Information;
+    if (Ids)
+    {
+        /* Calculate the length of existing MULTI_SZ value */
+        i = 0;
+        while (Ids[i] != UNICODE_NULL || Ids[i + 1] != UNICODE_NULL)
+        {
+            i++;
```
Extravert-ir, line 765: Just use `Length`?

drivers/storage/port/scsiport/fdo.c
```diff
@@ -705,6 +705,98 @@ FdoHandleDeviceRelations(
...
+    if (!NewIds)
+    {
+        return STATUS_INSUFFICIENT_RESOURCES;
+    }
+
+    if (Length)
+    {
+        RtlCopyMemory(NewIds, Ids, Length * sizeof(WCHAR));
+    }
+    RtlCopyMemory(&NewIds[Length], GenScsiAdapterId, wcslen(GenScsiAdapterId) * sizeof(WCHAR));
```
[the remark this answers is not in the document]
HBelusca: The sizeof now counts the NULL-terminator too. Is this wanted?

drivers/storage/port/scsiport/fdo.c
```diff
@@ -705,6 +705,98 @@ FdoHandleDeviceRelations(
...
+NTSTATUS
+FdoHandleQueryCompatibleId(
+    _Inout_ PIRP Irp)
+{
+    PWCHAR GenScsiAdapterId = L"GEN_SCSIADAPTER";
+    PWCHAR Ids, NewIds;
+    ULONG Length = 0;
+    ULONG i;
+
+    Ids = (PWCHAR)Irp->IoStatus.Information;
```
HBelusca, line 758: Is it really safe to assume that this always points to a safe MULTI_SZ and not to a potential corrupted buffer?

drivers/storage/port/scsiport/fdo.c
```diff
@@ -705,6 +705,98 @@ FdoHandleDeviceRelations(
...
+    if (Ids)
+    {
+        /* Calculate the length of existing MULTI_SZ value */
+        i = 0;
+        while (Ids[i] != UNICODE_NULL || Ids[i + 1] != UNICODE_NULL)
```
HBelusca, line 763: Why not using instead:
```c
Length = 0;
while (*Ids)
{
    Length += wcslen(Ids) + 1;
    Ids += Length;
}
```

drivers/storage/port/scsiport/fdo.c
binarymaster: Well, it should be either NULL, or a correct MULTI_SZ string.

drivers/storage/port/scsiport/fdo.c
binarymaster: This one counts each "line" of MULTI_SZ, and could allow getting rid of `i` variable, good idea!

drivers/storage/port/scsiport/fdo.c
binarymaster: `Ids[i]` looks better than `Ids[Length]`, but when applying Hermes' idea above, should look good too.

drivers/storage/port/scsiport/fdo.c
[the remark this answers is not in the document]
binarymaster: In this case counting NULL-terminator should be okay.

drivers/storage/port/scsiport/fdo.c
```diff
@@ -705,6 +705,98 @@ FdoHandleDeviceRelations(
...
+    if (Length)
+    {
+        RtlCopyMemory(NewIds, Ids, Length * sizeof(WCHAR));
+    }
+    RtlCopyMemory(&NewIds[Length], GenScsiAdapterId, wcslen(GenScsiAdapterId) * sizeof(WCHAR));
+
+    /* Finally replace identifiers */
+    if (Irp->IoStatus.Information)
+    {
+        ExFreePoolWithTag((PWCHAR)Irp->IoStatus.Information, TAG_SCSIPORT);
```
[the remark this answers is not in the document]
binarymaster: Is it okay to use just `ExPoolFree` in this case?

media/inf/scsi.inf
binarymaster: Agree, it should be in driver directory INF file, but it's out of this PR scope, so for the later.

drivers/storage/port/scsiport/fdo.c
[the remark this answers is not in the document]
binarymaster: Should be +2, as `sizeof(GenScsiAdapterId)` already contains unicode NULL-terminator.

binarymaster (commented):
Review addressed.

drivers/storage/port/scsiport/fdo.c
```diff
@@ -705,6 +705,53 @@ FdoHandleDeviceRelations(
     return IoCallDriver(PortExtension->Common.LowerDevice, Irp);
 }
 
+static
+NTSTATUS
+FdoHandleQueryCompatibleId(
+    _Inout_ PIRP Irp)
```
HBelusca, line 711: Since it seems this function actually acts just on the `Irp->IoStatus.Information` part, why not just make it take an
`_Inout_ PZZWSTR* Ids` (a pointer to a (pointer-to-)double-NULL-terminated string, which is exactly what a MULTI_SZ is) (this allows to have the function self-documented), then you work on that, and you call it like:
`status = FdoHandleQueryCompatibleId((PZZWSTR*)&Irp->IoStatus.Information);`

binarymaster: Makes sense, thanks, done.

drivers/storage/port/scsiport/fdo.c
```diff
@@ -705,6 +705,51 @@ FdoHandleDeviceRelations(
...
+    PWCHAR Ids = *PwIds, NewIds;
+    ULONG Length = 0;
+
+    if (Ids)
+    {
+        /* Calculate the length of existing MULTI_SZ value line by line */
+        while (*Ids)
+        {
+            Length += wcslen(Ids) + 1;
+            Ids += Length;
```
Extravert-ir, line 723: You need to add to `PwIds`, not to `Ids`, otherwise it's a buffer overflow

drivers/storage/port/scsiport/fdo.c
```diff
@@ -732,6 +777,16 @@ FdoDispatchPnp(
         {
             return FdoHandleDeviceRelations(portExt, Irp);
         }
+        case IRP_MN_QUERY_ID:
+        {
+           if (ioStack->Parameters.QueryId.IdType == BusQueryCompatibleIDs)
+           {
+               IoForwardIrpSynchronously(portExt->Common.LowerDevice, Irp);
```
Extravert-ir, line 784: Set the `Irp->IoStatus.Information` to 0, in case `IoForwardIrpSynchronously` fails

Extravert-ir (changes requested):

drivers/storage/port/scsiport/fdo.c
HBelusca: Are you sure? The PwIds is also used for the output.

drivers/storage/port/scsiport/fdo.c
```diff
@@ -705,6 +705,51 @@ FdoHandleDeviceRelations(
...
+        {
+            Length += wcslen(Ids) + 1;
+            Ids += Length;
+        }
+        Ids -= Length;
+    }
+
+    /* New MULTI_SZ with added identifier and finalizing zeros */
+    NewIds = ExAllocatePoolZero(PagedPool,
+                                Length * sizeof(WCHAR) + sizeof(GenScsiAdapterId) + 2,
```
HBelusca, line 730: Is this "2" a `sizeof(UNICODE_NULL)` ?

drivers/storage/port/scsiport/fdo.c
```diff
@@ -705,6 +705,51 @@ FdoHandleDeviceRelations(
...
+
+    if (Ids)
+    {
+        /* Calculate the length of existing MULTI_SZ value line by line */
+        while (*Ids)
+        {
+            Length += wcslen(Ids) + 1;
+            Ids += Length;
+        }
+        Ids -= Length;
```
HBelusca, line 725: Or `Ids = *PwIds;` again... (your choice)

drivers/storage/port/scsiport/fdo.c
```diff
@@ -732,6 +777,16 @@ FdoDispatchPnp(
         {
             return FdoHandleDeviceRelations(portExt, Irp);
         }
+        case IRP_MN_QUERY_ID:
+        {
+           if (ioStack->Parameters.QueryId.IdType == BusQueryCompatibleIDs)
```
HBelusca, line 782: What is happening to the indentation space here?

drivers/storage/port/scsiport/fdo.c
binarymaster: Yes.

drivers/storage/port/scsiport/fdo.c
binarymaster: Why? `Ids` is a local variable which holds a copy of the string pointer.

Extravert-ir: `Length` contains the whole length processed. So if there are (for example) 3 strings there, you will add the length the first one 3 times

HBelusca: If `*PwIds` ( == original `Ids` before the loop) points to a MULTI_SZ buffer having 3 strings (+ the ending NULL terminator), after the while loop `Length` will contain the total number of characters of the whole MULTI_SZ buffer, excepting the very last NULL-terminator.
When the code below copies the data, it will copy the first 3 strings (total number of characters with their own NULLs, but NOT the last terminating NULL char of the MULTI_SZ), followed by a new 4th string the `GenScsiAdapterId` one (its sizeof counts its own NULL-terminator), and then, due to the size specification in the `ExAllocatePoolZero()` call, the last char is already zero and corresponds to the final ending UNICODE_NULL.

binarymaster: Ah I see now, `Ids += Length;` will increase the pointer more than needed, as it contains entire length. Blame @HBelusca :trollface:

Extravert-ir: You're like counting an arithmetic progression here

HBelusca: Oh lol yes you're correct !
(usually what I do with that thing is:
```c
while (Ids)
{
    Ids += wcslen(Ids) + 1;
}
without having the Length thing, and then later I do: `Length = Ids - <the_old_Ids>`

File: drivers/storage/port/scsiport/fdo.c [hunk 14]
```diff
@@ -732,6 +777,20 @@ FdoDispatchPnp(
         {
             return FdoHandleDeviceRelations(portExt, Irp);
         }
+        case IRP_MN_QUERY_ID:
+        {
+            if (ioStack->Parameters.QueryId.IdType == BusQueryCompatibleIDs)
+            {
+                status = IoForwardIrpSynchronously(portExt->Common.LowerDevice, Irp);
```
Extravert-ir, line 784: ACHTUNG!
`IoForwardIrpSynchronously` returns `BOOLEAN` :D

drivers/storage/port/scsiport/fdo.c
```diff
@@ -732,6 +777,20 @@ FdoDispatchPnp(
...
+                if (!NT_SUCCESS(status))
+                {
+                    Irp->IoStatus.Information = 0;
+                }
```
Extravert-ir, line 788:
```suggestion
                Irp->IoStatus.Information = 0;
                IoForwardIrpSynchronously(portExt->Common.LowerDevice, Irp);
```

HBelusca: You don't look at the returned value of IoForwardIrpSynchronously?

drivers/storage/port/scsiport/fdo.c
binarymaster: Fixed now. Well, having all this discusso here makes me think my first approach with a loop checking for two consequence `UNICODE_NULL`s was much simpler. :smile:

drivers/storage/port/scsiport/fdo.c
binarymaster: ULTRAKEK :rofl:

drivers/storage/port/scsiport/fdo.c
binarymaster: Yep, we continue execution (appending the ID) regardless of it.

Extravert-ir (approved):

HBelusca (approved):

PR #3442, opened 2021-01-31
Restore ReactOS sysdm.cpl layout

ColinFinck:
## Purpose

This PR restores the ReactOS sysdm.cpl layout we had before 2021-01-20 that has defined ReactOS' personality for over a decade. There is no point in blatantly copying the Windows XP layout here!

I'm well aware that this brings back [CORE-17429](https://jira.reactos.org/browse/CORE-17429), but at this point let me be frank:
I'd rather like to have a cut-off version string than losing our identity!
There are definitely ways to fix that string later without redesigning the dialog altogether (e.g. by changing the overall dialog size). This code should give a better base to start from.

I'm also aware that this leaves no space for future OEM logos/manufacturer strings either.
Let me again be honest here: This is no critical feature that justifies taking the fun out of ReactOS.

In the interest of simplifying maintenance and even reducing the binary size, I have moved the `IDS_DEVS` string to the `LANG_NEUTRAL` resources.
This should also be updated in the future, maybe even automatically. But again, this code should give a better base to start from.

As the previous change to sysdm.cpl directly to master has already sent tempers flaring, I'm opening a PR.
That should have happened in the first place!


dll/cpl/sysdm/sysdm.rc
```diff
@@ -29,6 +29,11 @@ RC_LICENSE RTDATA "resources/gpl.txt"
...
...haposhnikov\nAmine Khaldi\nAndrew Greenwood\nAndrey Korotaev\nArt Yerkes\nBenedikt Freisen\nCameron Gutman\nChristoph von Wittich\nColin Finck\nDaniel Reimer\nDavid Quintana\nDmitry Chapyshev\nEric Kohl\nGed Murphy\nGiannis Adamopoulos\nGregor Brunmar\nHermès Bélusca-Maïto\nHervé Poussineau\nJames Tabor\nJeffrey Morlan\nJérôme Gardou\nJohannes Anderwald\nKamil Hornícek\nKatayama Hirofumi MZ\nKJK::Hyperion\nMaarten Bosma\nMagnus Olsen\nMarc Piulachs\nMark Jansen\nMatthias Kupfer\nMike Nordell\nPeter Hater\nPeter Ward\nPierre Schweitzer\nSaveliy Tretiakov\nStanislav Motylkov\nStefan Ginsberg\nSylvain Petreolle\nThomas Blümel\nThomas Faber\nTimo Kreuzer\nVadim Galyant\n\nAlex Ionescu\nFilip Navara\nGunnar Dalsnes\nMartin Fuchs\nRoyce Mitchell III\nBrandon Turner\nBrian Palmer\nCasper Hornstrup\nDavid Welch\nEmanuele Aliberti\nGé van Geldorp\nGregor Anich\nJason Filby\nJens Collin\nMichael Wirth\nNathan Woods\nRobert Dickenson\nRex Jolliff\nVizzini\n\nRelease Engineers\n\nAmine Khaldi\nColin Finck\nJoachim Henze\nThomas Faber\nZ98\n\nWebsite Team\n\nColin Finck\nJaix Bly\nKlemens Friedl\nZ98\n\nMedia Team\n\nMindflyer\nWierd_W\n\nfurther thanks go to\n\nall Contributors\nWine Team\n\n"
```
learn-more, line 34: To simplify maintenance you could even make this a multi-line string directly in the code.

learn-more (approved):

GeoB99 (approved):

dll/cpl/sysdm/sysdm.rc
Extravert-ir: As we're here, maybe we can add some people too. Me, for example :)

Extravert-ir (approved):

dll/cpl/sysdm/sysdm.rc
ColinFinck: Sounds like a plan as soon as we autogenerate that string!
For now, I know how to reliably encode UTF-8 in resource files, but not in plain C code files. And we have some developer names containing UTF-8 characters.

binarymaster (approved):
While we're at it, maybe we can also fix version string by simply resizing the main tab of system properties.

JoachimHenze:
Since I have been added for review: I am fine with these changes, if they come paired with something that preserves the string displaying. I will can then port it back myself into the older releases even. No need for you to waste a whole day with that additionally.
Two change requests though:
1.) Have the commit message say "that it's the teams consensus that OEM branding can not be implemented anymore with that solution, and that the whole team agrees that we will never want to do that", so that future devs are aware of that deliberate decision.
2.) Fix the version string displaying to guarantee it not being cut off.

learn-more:
> Since I have been added for review: I am fine with these changes, if they come paired with something that preserves the string displaying. I will can then port it back myself into the older releases even. No need for you to waste a whole day with that additionally.
> Two change requests though:
> 1.) Have the commit message say "that it's the teams consensus that OEM branding can not be implemented anymore with that solution, and that the whole team agrees that we will never want to do that", so that future devs are aware of that deliberate decision.
> 2.) Fix the version string displaying to guarantee it not being cut off.

If someone wants to implement OEM branding for whatever reason, they can always just add a button that pops an OEM dialog?

JoachimHenze:
>>If someone wants to implement OEM branding for whatever reason, they can always just add a button that pops an OEM dialog?
Ok, sounds reasonable. What I meant to preserve as information for the future was "make it obvious to future devs that changing this dialog has a political dimension and will not be tolerated,"
Then we just need a fix the string displaying.

dll/cpl/sysdm/sysdm.rc
HBelusca: Do we still have a project coordinator? 🐸

HBelusca (approved):

JoachimHenze:
Attention:
Your change in da-DK will for sure cut off that string 80->32:
LTEXT "Operativsystem:", IDC_STATIC, 114, 16, 80, 9    ->    LTEXT "Operativsystem:", IDC_STATIC, 4, 116, 32, 9
Why not leaving that controls length the maximum that was used in any language?

JoachimHenze:
the width of 80 there should be sufficient for all languages.

ColinFinck:
This is the final dialog now:

grafik

- A revision string without date has been introduced for this (`KERNEL_VERSION_REVISION_RC`, `KERNEL_VERSION_REVISION_STR`)
- Labels have been realigned and resized to prevent cut-offs

JoachimHenze:
yes, label length 100 is good as well. Thanks! Good to go!

JoachimHenze (approved):

HBelusca (approved):

binarymaster (approved):

PR #3444, opened 2021-02-03
[THEMES] Fix Lunar theme

Venca24:
## Purpose

Fixes a mistake from #3424 where is no reference to the new file with Czech translation in Lunar theme. Everything else is OK.


binarymaster (changes requested):
Hello,

```
```

```
```

Regards,

binarymaster (approved):
Now should be ok, but only for rebase and merge. Cc @gonzoMD

PR #3445, opened 2021-02-03
[NTOS:SE] Do not use a global lock for tokens

GeoB99:
Instead of using a global lock for tokens, initialise the lock locally every time a token is about to be created or duplicated (as in Windows Server 2003).

### DIFF
--- a/ntoskrnl/se/token.c
+++ b/ntoskrnl/se/token.c
@@ -27,7 +27,6 @@ typedef struct _TOKEN_AUDIT_POLICY_INFORMATION
 /* GLOBALS ********************************************************************/
 
 POBJECT_TYPE SeTokenObjectType = NULL;
-ERESOURCE SepTokenLock; // FIXME: Global lock!
 
 TOKEN_SOURCE SeSystemTokenSource = {"*SYSTEM*", {0}};
 LUID SeSystemAuthenticationId = SYSTEM_LUID;
@@ -82,6 +81,59 @@ static const INFORMATION_CLASS_INFO SeTokenInformationClass[] = {
 };
 
 /* FUNCTIONS *****************************************************************/
+
+/**
+ * @brief
+ * Creates a lock for the token.
+ * 
+ * @param[in,out] Token
+ * A token which lock has to be created.
+ *
+ * @return
+ * STATUS_SUCCESS if the pool allocation and resource initialisation have
+ * completed successfully, otherwise STATUS_INSUFFICIENT_RESOURCES on a
+ * pool allocation failure.
+ */
+static
+NTSTATUS
+SepCreateTokenLock(
+    _Inout_ PTOKEN Token)
+{
+    PAGED_CODE();
+
+    Token->TokenLock = ExAllocatePoolWithTag(NonPagedPool,
+                                             sizeof(ERESOURCE),
+                                             TAG_SE_TOKEN_LOCK);
+    if (Token->TokenLock == NULL)
+    {
+        DPRINT1("SepCreateTokenLock(): Failed to allocate memory!\n");
+        return STATUS_INSUFFICIENT_RESOURCES;
+    }
+
+    ExInitializeResourceLite(Token->TokenLock);
+    return STATUS_SUCCESS;
+}
+
+/**
+ * @brief
+ * Deletes a lock of a token.
+ *
+ * @param[in,out] Token
+ * A token which contains the lock.
+ *
+ * @return
+ * Nothing.
+ */
+static
+VOID
+SepDeleteTokenLock(
+    _Inout_ PTOKEN Token)
+{
+    PAGED_CODE();
+
+    ExDeleteResourceLite(Token->TokenLock);
+    ExFreePoolWithTag(Token->TokenLock, TAG_SE_TOKEN_LOCK);
+}
 
 static NTSTATUS
 SepCompareTokens(IN PTOKEN FirstToken,
@@ -477,7 +529,13 @@ SepDuplicateToken(
     AccessToken->TokenType = TokenType;
     AccessToken->ImpersonationLevel = Level;
 
-    AccessToken->TokenLock = &SepTokenLock; // FIXME: Global lock!
+    /* Initialise the lock for the access token */
+    Status = SepCreateTokenLock(AccessToken);
+    if (!NT_SUCCESS(Status))
+    {
+        ObDereferenceObject(AccessToken);
+        return Status;
+    }
 
     /* Copy the immutable fields */
     RtlCopyLuid(&AccessToken->TokenSource.SourceIdentifier,
@@ -490,7 +548,6 @@ SepDuplicateToken(
     AccessToken->ParentTokenId = Token->ParentTokenId;
     AccessToken->ExpirationTime = Token->ExpirationTime;
     AccessToken->OriginatingLogonSession = Token->OriginatingLogonSession;
-
 
     /* Lock the source token and copy the mutable fields */
     SepAcquireTokenLockExclusive(Token);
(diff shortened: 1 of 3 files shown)


ntoskrnl/se/token.c
```diff
@@ -83,6 +82,64 @@ static const INFORMATION_CLASS_INFO SeTokenInformationClass[] = {
...
+ */
+FORCEINLINE
+NTSTATUS
+SepCreateTokenLock(
+    _Inout_ PTOKEN Token)
+{
+    PAGED_CODE();
+
+    /* Allocate non paged block of memory for the resource */
+    Token->TokenLock = (PERESOURCE)ExAllocatePoolWithTag(NonPagedPool,
```
Extravert-ir, line 106: The cast is unnecessary

ntoskrnl/se/token.c
```diff
@@ -83,6 +82,64 @@ static const INFORMATION_CLASS_INFO SeTokenInformationClass[] = {
...
+ * @param[in,out] Token
+ * A token which lock has to be created.
+ *
+ * @return
+ * STATUS_SUCCESS if the pool allocation and resource initialisation have
+ * completed successfully, otherwise STATUS_INSUFFICIENT_RESOURCES on a
+ * pool allocation failure.
+ *
+ */
+FORCEINLINE
```
Extravert-ir, line 98: `static` should be enough here, I think. `FORCEINLINE` is mostly for headers

ntoskrnl/se/token.c
```diff
@@ -477,7 +534,13 @@ SepDuplicateToken(
     AccessToken->TokenType = TokenType;
     AccessToken->ImpersonationLevel = Level;
 
-    AccessToken->TokenLock = &SepTokenLock; // FIXME: Global lock!
+    /* Initialise the lock for the access token */
+    Status = SepCreateTokenLock(AccessToken);
+    if (!NT_SUCCESS(Status))
+    {
+        DPRINT1("SepCreateTokenLock() failed (Status 0x%lx)\n", Status);
```
zefklop, line 541: Shouldn't we perform some cleanup there ?

ntoskrnl/se/token.c
```diff
@@ -976,7 +1040,13 @@ SepCreateToken(
     AccessToken->TokenType = TokenType;
     AccessToken->ImpersonationLevel = ImpersonationLevel;
 
-    AccessToken->TokenLock = &SepTokenLock; // FIXME: Global lock!
+    /* Initialise the lock for the access token */
+    Status = SepCreateTokenLock(AccessToken);
+    if (!NT_SUCCESS(Status))
+    {
+        DPRINT1("SepCreateTokenLock() failed (Status 0x%lx)\n", Status);
+        return Status;
```
zefklop, line 1048: Clean up here, also.

zefklop (changes requested):
Please implement clean-up paths or add a comment as why it's not necessary.

zefklop (approved):
Looks good to me, now.

ntoskrnl/se/token.c
```diff
@@ -83,6 +82,64 @@ static const INFORMATION_CLASS_INFO SeTokenInformationClass[] = {
...
+ *
+ */
+static
+NTSTATUS
+SepCreateTokenLock(
+    _Inout_ PTOKEN Token)
+{
+    PAGED_CODE();
+
+    /* Allocate non paged block of memory for the resource */
```
HBelusca, line 105: Code is already self-explanatory.

ntoskrnl/se/token.c
```diff
@@ -83,6 +82,64 @@ static const INFORMATION_CLASS_INFO SeTokenInformationClass[] = {
...
+    Token->TokenLock = ExAllocatePoolWithTag(NonPagedPool,
+                                             sizeof(ERESOURCE),
+                                             TAG_SE_TOKEN_LOCK);
+    if (Token->TokenLock == NULL)
+    {
+        DPRINT1("SepCreateTokenLock(): Failed to allocate into the memory pool!\n");
+        return STATUS_INSUFFICIENT_RESOURCES;
+    }
+
+    /* Now initialise the resource lock */
```
HBelusca, line 115: ditto

ntoskrnl/se/token.c
```diff
@@ -83,6 +82,64 @@ static const INFORMATION_CLASS_INFO SeTokenInformationClass[] = {
...
+ *
+ */
+static
+VOID
+SepDeleteTokenLock(
+    _In_ PTOKEN Token)
+{
+    PAGED_CODE();
+
+    /* Delete the token resource lock and free the memory pool */
```
HBelusca, line 138: ditto

ntoskrnl/se/token.c
```diff
@@ -83,6 +82,64 @@ static const INFORMATION_CLASS_INFO SeTokenInformationClass[] = {
...
+ * Creates a lock for the token.
+ * 
+ * @param[in,out] Token
+ * A token which lock has to be created.
+ *
+ * @return
+ * STATUS_SUCCESS if the pool allocation and resource initialisation have
+ * completed successfully, otherwise STATUS_INSUFFICIENT_RESOURCES on a
+ * pool allocation failure.
+ *
```
HBelusca, line 96:
```suggestion
```

ntoskrnl/se/token.c
```diff
@@ -83,6 +82,64 @@ static const INFORMATION_CLASS_INFO SeTokenInformationClass[] = {
...
+/**
+ * @brief
+ * Deletes a lock of a token.
+ *
+ * @param[in] Token
+ * A token which contains the lock.
+ *
+ * @return
+ * Nothing.
+ *
```
HBelusca, line 129:
```suggestion
```

ntoskrnl/se/token.c
```diff
@@ -83,6 +82,64 @@ static const INFORMATION_CLASS_INFO SeTokenInformationClass[] = {
...
+ */
+static
+VOID
+SepDeleteTokenLock(
+    _In_ PTOKEN Token)
```
HBelusca, line 134: `_Inout_` ?

ntoskrnl/se/token.c
```diff
@@ -83,6 +82,64 @@ static const INFORMATION_CLASS_INFO SeTokenInformationClass[] = {
...
+{
+    PAGED_CODE();
+
+    /* Allocate non paged block of memory for the resource */
+    Token->TokenLock = ExAllocatePoolWithTag(NonPagedPool,
+                                             sizeof(ERESOURCE),
+                                             TAG_SE_TOKEN_LOCK);
+    if (Token->TokenLock == NULL)
+    {
+        DPRINT1("SepCreateTokenLock(): Failed to allocate into the memory pool!\n");
```
HBelusca, line 111:
```suggestion
        DPRINT1("SepCreateTokenLock(): Failed to allocate memory!\n");
```

HBelusca: Besides, since this SepCreateTokenLock() is called at two places and in both places you also DPRINT when it fails, there is then no need to also add yet another DPRINT there.

ntoskrnl/se/token.c
```diff
@@ -83,6 +82,64 @@ static const INFORMATION_CLASS_INFO SeTokenInformationClass[] = {
...
+ * @param[in,out] Token
+ * A token which lock has to be created.
+ *
+ * @return
+ * STATUS_SUCCESS if the pool allocation and resource initialisation have
+ * completed successfully, otherwise STATUS_INSUFFICIENT_RESOURCES on a
+ * pool allocation failure.
+ *
+ */
+static
```
HBelusca, line 98:
```suggestion
static inline
```

ntoskrnl/se/token.c
```diff
@@ -83,6 +82,64 @@ static const INFORMATION_CLASS_INFO SeTokenInformationClass[] = {
...
+ * Deletes a lock of a token.
+ *
+ * @param[in] Token
+ * A token which contains the lock.
+ *
+ * @return
+ * Nothing.
+ *
+ */
+static
```
HBelusca, line 131:
```suggestion
static inline
```

HBelusca (commented):
Mostly style, otherwise LGTM

ntoskrnl/se/token.c
Extravert-ir: No

HBelusca: Since these functions are short and used only once or twice they can be made inline.

Extravert-ir: `static inline` does not have any difference to `static` in this case. Only wasting lines

PR #3447, opened 2021-02-05
[NTOS:SE] Complete the SepCompareTokens implementation

GeoB99:
## Purpose
This implements the rest of `SepCompareTokens()` code which makes `NtCompareTokens()` able to properly distinguish the tokens.
## Proposed Changes
* Implement SepCompareSidAndAttributesFromTokens and SepComparePrivilegeAndAttributesFromTokens functions for array elements comparison
* Implement the token comparison code in SepCompareTokens function
* Add a missing PAGED_CODE() in SepCompareTokens as most of the token comparison code is paged
* Use SAL annotations for SepCompareTokens and NtCompareTokens
## TODO
- [x] Implement a testcase for `NtCompareTokens()` to exercise the correctness of the code's behaviour

### DIFF
--- a/modules/rostests/apitests/ntdll/NtCompareTokens.c
+++ b/modules/rostests/apitests/ntdll/NtCompareTokens.c
@@ -1,0 +1,109 @@
+/*
+ * PROJECT:         ReactOS API tests
+ * LICENSE:         GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:         Tests for the NtCompareTokens API
+ * COPYRIGHT:       Copyright 2021 George Bișoc <george.bisoc@reactos.org>
+ */
+
+#include "precomp.h"
+
+static
+HANDLE
+GetTokenFromCurrentProcess(VOID)
+{
+    BOOL Success;
+    HANDLE Token;
+
+    Success = OpenProcessToken(GetCurrentProcess(),
+                               TOKEN_READ | TOKEN_ADJUST_PRIVILEGES | TOKEN_DUPLICATE,
+                               &Token);
+    if (!Success)
+    {
+        skip("OpenProcessToken() has failed to get the process' token (error code: %lu)!\n", GetLastError());
+        return NULL;
+    }
+
+    return Token;
+}
+
+static
+HANDLE
+GetDuplicateToken(_In_ HANDLE Token)
+{
+    BOOL Success;
+    HANDLE ReturnedToken;
+
+    Success = DuplicateToken(Token, SecurityIdentification, &ReturnedToken);
+    if (!Success)
+    {
+        skip("DuplicateToken() has failed to get the process' token (error code: %lu)!\n", GetLastError());
+        return NULL;
+    }
+
+    return ReturnedToken;
+}
+
+static
+VOID
+DisableTokenPrivileges(_In_ HANDLE Token)
+{
+    BOOL Success;
+
+    Success = AdjustTokenPrivileges(Token, TRUE, NULL, 0, NULL, NULL);
+    if (!Success)
+    {
+        skip("AdjustTokenPrivileges() has failed to disable the privileges (error code: %lu)!\n", GetLastError());
+        return;
+    }
+}
+
+START_TEST(NtCompareTokens)
+{
+    NTSTATUS Status;
+    HANDLE ProcessToken = NULL;
+    HANDLE DuplicatedToken = NULL;
+    BOOLEAN IsEqual = FALSE;
+
+    /* Obtain some tokens from current process */
+    ProcessToken = GetTokenFromCurrentProcess();
+    DuplicatedToken = GetDuplicateToken(ProcessToken);
+
+    /*
+     * Give invalid token handles and don't output
+     * the returned value in the last parameter.
+     */
+    Status = NtCompareTokens(NULL, NULL, NULL);
+    ok_hex(Status, STATUS_ACCESS_VIOLATION);
+
+    /*
+     * Token handles are valid but don't output
+     * the returned value.
+     */
+    Status = NtCompareTokens(ProcessToken, ProcessToken, NULL);
+    ok_hex(Status, STATUS_ACCESS_VIOLATION);
+
+    /* The tokens are the same */
+    Status = NtCompareTokens(ProcessToken, ProcessToken, &IsEqual);
+    ok_hex(Status, STATUS_SUCCESS);
+    ok(IsEqual == TRUE, "Equal tokens expected but they aren't (current value: %u)!\n", IsEqual);
+
+    /* A token is duplicated with equal SIDs and privileges */
+    Status = NtCompareTokens(ProcessToken, DuplicatedToken, &IsEqual);
+    ok_hex(Status, STATUS_SUCCESS);
+    ok(IsEqual == TRUE, "Equal tokens expected but they aren't (current value: %u)!\n", IsEqual);
+
+    /* Disable all the privileges for token. */
+    DisableTokenPrivileges(ProcessToken);
+
+    /*
+     * The main token has privileges disabled but the
+     * duplicated one has them enabled still.
+     */
+    Status = NtCompareTokens(ProcessToken, DuplicatedToken, &IsEqual);
+    ok_hex(Status, STATUS_SUCCESS);
+    ok(IsEqual == FALSE, "Tokens mustn't be equal (current value: %u)!\n", IsEqual);
+
+    /* We finished our tests, close the tokens */
+    CloseHandle(ProcessToken);
+    CloseHandle(DuplicatedToken);
+}
--- a/ntoskrnl/se/token.c
+++ b/ntoskrnl/se/token.c
@@ -135,12 +135,173 @@ SepDeleteTokenLock(
     ExFreePoolWithTag(Token->TokenLock, TAG_SE_TOKEN_LOCK);
 }
 
-static NTSTATUS
-SepCompareTokens(IN PTOKEN FirstToken,
-                 IN PTOKEN SecondToken,
-                 OUT PBOOLEAN Equal)
+/**
+ * @brief
+ * Compares the elements of SID arrays provided by tokens.
+ * The elements that are being compared for equality are
+ * the SIDs and their attributes.
+ *
+ * @param[in] SidArrayToken1
+ * SID array from the first token.
+ *
+ * @param[in] CountSidArray1
+ * SID count array from the first token.
+ *
+ * @param[in] SidArrayToken2
+ * SID array from the second token.
+ *
+ * @param[in] CountSidArray2
+ * SID count array from the second token.
+ * 
+ * @return
+ * Returns TRUE if the elements match from either arrays,
+ * FALSE otherwise.
+ */
+static
+BOOLEAN
+SepCompareSidAndAttributesFromTokens(
+    _In_ PSID_AND_ATTRIBUTES SidArrayToken1,
+    _In_ ULONG CountSidArray1,
+    _In_ PSID_AND_ATTRIBUTES SidArrayToken2,
+    _In_ ULONG CountSidArray2)
+{
+    ULONG FirstCount, SecondCount;
+    PSID_AND_ATTRIBUTES FirstSidArray, SecondSidArray;
+    PAGED_CODE();
+
+    /* Bail out if index counters provided are not equal */
+    if (CountSidArray1 != CountSidArray2)
+    {
+        DPRINT("SepCompareSidAndAttributesFromTokens(): Index counters are not the same!\n");
+        return FALSE;
+    }
+
+    /* Loop over the SID arrays and compare them */
+    for (FirstCount = 0; FirstCount < CountSidArray1; FirstCount++)
+    {
+        for (SecondCount = 0; SecondCount < CountSidArray2; SecondCount++)
+        {
+            FirstSidArray = &SidArrayToken1[FirstCount];
+            SecondSidArray = &SidArrayToken2[SecondCount];
+
+            if (RtlEqualSid(FirstSidArray->Sid, SecondSidArray->Sid) &&
+                FirstSidArray->Attributes == SecondSidArray->Attributes)
+            {
+                break;
+            }
+        }
+
+        /* We've exhausted the array of the second token without finding this one */
+        if (SecondCount == CountSidArray2)
+        {
+            DPRINT("SepCompareSidAndAttributesFromTokens(): No matching elements could be found in either token!\n");
+            return FALSE;
+        }
+    }
+
+    return TRUE;
+}
+
+/**
+ * @brief
+ * Compares the elements of privilege arrays provided by tokens.
+ * The elements that are being compared for equality are
+ * the privileges and their attributes.
+ *
+ * @param[in] PrivArrayToken1
+ * Privilege array from the first token.
+ *
+ * @param[in] CountPrivArray1
+ * Privilege count array from the first token.
+ *
+ * @param[in] PrivArrayToken2
+ * Privilege array from the second token.
+ *
+ * @param[in] CountPrivArray2
+ * Privilege count array from the second token.
+ *
+ * @return
+ * Returns TRUE if the elements match from either arrays,
+ * FALSE otherwise.
+ */
+static
+BOOLEAN
+SepComparePrivilegeAndAttributesFromTokens(
+    _In_ PLUID_AND_ATTRIBUTES PrivArrayToken1,
+    _In_ ULONG CountPrivArray1,
+    _In_ PLUID_AND_ATTRIBUTES PrivArrayToken2,
+    _In_ ULONG CountPrivArray2)
+{
+    ULONG FirstCount, SecondCount;
+    PLUID_AND_ATTRIBUTES FirstPrivArray, SecondPrivArray;
+    PAGED_CODE();
+
+    /* Bail out if index counters provided are not equal */
+    if (CountPrivArray1 != CountPrivArray2)
+    {
+        DPRINT("SepComparePrivilegeAndAttributesFromTokens(): Index counters are not the same!\n");
+        return FALSE;
+    }
+
+    /* Loop over the privilege arrays and compare them */
+    for (FirstCount = 0; FirstCount < CountPrivArray1; FirstCount++)
+    {
+        for (SecondCount = 0; SecondCount < CountPrivArray2; SecondCount++)
+        {
+            FirstPrivArray = &PrivArrayToken1[FirstCount];
+            SecondPrivArray = &PrivArrayToken2[SecondCount];
+
+            if (RtlEqualLuid(&FirstPrivArray->Luid, &SecondPrivArray->Luid) &&
+                FirstPrivArray->Attributes == SecondPrivArray->Attributes)
+            {
+                break;
+            }
+        }
+
+        /* We've exhausted the array of the second token without finding this one */
+        if (SecondCount == CountPrivArray2)
+        {
+            DPRINT("SepComparePrivilegeAndAttributesFromTokens(): No matching elements could be found in either token!\n");
+            return FALSE;
+        }
+    }
+
+    return TRUE;
+}
+
+/**
+ * @brief
+ * Compares tokens if they're equal based on all the following properties. If all
+ * of the said conditions are met then the tokens are deemed as equal.
+ *
+ * - Every SID that is present in either token is also present in the other one.
+ * - Both or none of the tokens are restricted.
+ * - If both tokens are restricted, every SID that is restricted in either token is
+ *   also restricted in the other one.
+ * - Every privilege present in either token is also present in the other one.
+ * 
+ * @param[in] FirstToken
+ * The first token.
+ *
+ * @param[in] SecondToken
+ * The second token.
+ *
+ * @param[out] Equal
+ * The retrieved value which determines if the tokens are
+ * equal or not.
+ *
+ * @return
+ * Returns STATUS_SUCCESS.
+ */
+static
+NTSTATUS
+SepCompareTokens(
+    _In_ PTOKEN FirstToken,
+    _In_ PTOKEN SecondToken,
+    _Out_ PBOOLEAN Equal)
 {
     BOOLEAN Restricted, IsEqual = FALSE;
+    PAGED_CODE();
 
     ASSERT(FirstToken != SecondToken);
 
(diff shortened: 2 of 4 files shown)


GeoB99:
Test results
**Windows Server 2003**
Capture2
**ReactOS**
Capture

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,222 @@ SepDeleteTokenLock(
...
+                        goto Quit;
+                    }
+                }
+            }
         }
 
-        /* FIXME: Check if every privilege that is present in either token is also present in the other one */
-        DPRINT1("FIXME: Pretending tokens are equal!\n");
-        IsEqual = TRUE;
+        /* Check if every privilege that is present in either token is also present in the other one */
```
HBelusca, line 333: Can you confirm that you want to look at the privileges and attributes from the tokens only if **just the first token** is restricted, but don't care about the second one? (That sounds broken.)

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,222 @@ SepDeleteTokenLock(
...
+ * @param[in] IndexOfSecondToken
+ * The index of the second token.
+ *
+ * @return
+ * Returns TRUE if both the privileges and the attributes are equal,
+ * FALSE otherwise.
+ */
+static
+BOOLEAN
+SepComparePrivilegesAndAttributesFromTokens(
```
HBelusca, line 160: Why not inlining this function? It is used only once.

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,222 @@ SepDeleteTokenLock(
...
+ *
+ * @param[in] IndexOfSecondToken
+ * The index of the second token.
+ *
+ * @return
+ * Returns TRUE if both the SIDs and the attributes are equal,
+ * FALSE otherwise.
+ */
+static
+BOOLEAN
```
HBelusca, line 200: Why not inlining this function? It is used only once.

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,222 @@ SepDeleteTokenLock(
...
+ *
+ * @param[in] IndexOfSecondToken
+ * The index of the second token.
+ *
+ * @return
+ * Returns TRUE if both the restricted SIDs and are equal,
+ * FALSE otherwise.
+ */
+static
+BOOLEAN
```
HBelusca, line 243: Same question

ntoskrnl/se/token.c
GeoB99: I don't see much point in inlining the functions since they're already statically declared as any other private routine of this file, but fine by me I guess. I'll make those functions inlined then.

zefklop: Please do not use inline for static functions which are not part of a header. 

1. Let the compiler be smarter than you on this topic.
2. Let the debugger help you when things go wrong.

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,243 @@ SepDeleteTokenLock(
...
+{
+    BOOLEAN IsSidRestrictedEqual;
+    PAGED_CODE();
+
+    /* Are the restricted SIDs from either token equal? */
+    IsSidRestrictedEqual = RtlEqualSid(FirstToken->RestrictedSids[IndexInFirstToken].Sid, SecondToken->RestrictedSids[IndexInSecondToken].Sid);
+    if (!IsSidRestrictedEqual)
+        return FALSE;
+
+    return TRUE;
```
Extravert-ir, line 258: Why so complicated?
```suggestion
    return RtlEqualSid(FirstToken->RestrictedSids[IndexInFirstToken].Sid,
                       SecondToken->RestrictedSids[IndexInSecondToken].Sid);
```

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,243 @@ SepDeleteTokenLock(
...
+    /* Are the SIDs from either token equal? */
+    IsSidEqual = RtlEqualSid(FirstToken->UserAndGroups[IndexInFirstToken].Sid, SecondToken->UserAndGroups[IndexInSecondToken].Sid);
+    if (!IsSidEqual)
+        return FALSE;
+
+    /* Are the attributes from either SID from token equal? */
+    if (FirstToken->UserAndGroups[IndexInFirstToken].Attributes != SecondToken->UserAndGroups[IndexInSecondToken].Attributes)
+        return FALSE;
+
+    return TRUE;
```
Extravert-ir, line 219: Keep it simple
```suggestion
    PAGED_CODE();
    
    PSID_AND_ATTRIBUTES token1 = &FirstToken->UserAndGroups[IndexInFirstToken];
    PSID_AND_ATTRIBUTES token2 = &FirstToken->UserAndGroups[IndexInSecondToken];
    
    return RtlEqualSid(token1->Sid, token2->Sid) && token1->Attributes == token2->Attributes;
```

Extravert-ir:
Will these compare functions be used anywhere else?
If not, I don't think it's worth making them, they are simple enough to be just put right where used

GeoB99:
The comparison functions are used exclusively for `SepCompareTokens/NtCompareTokens`. I can simply move the comparison code to `SepCompareTokens` private function if that's what you want, not a big deal for me.

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,138 @@ SepDeleteTokenLock(
     ExFreePoolWithTag(Token->TokenLock, TAG_SE_TOKEN_LOCK);
 }
 
-static NTSTATUS
-SepCompareTokens(IN PTOKEN FirstToken,
-                 IN PTOKEN SecondToken,
-                 OUT PBOOLEAN Equal)
+/**
+ * @brief
+ * Compares tokens if they're equal given their properties.
```
HBelusca, line 140: You may add a bit more explanation there.
To quote the documentation:
```
The NtCompareTokens function compares two access tokens and determines whether they are equivalent with respect to a call to the AccessCheck function.

[...]

Two access control tokens are considered to be equivalent if all of the following conditions are true:

- Every security identifier (SID) that is present in either token is also present in the other token.
- Neither or both of the tokens are restricted.
- If both tokens are restricted, every SID that is restricted in one token is also restricted in the other token.
- Every privilege present in either token is also present in the other token.
```

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,138 @@ SepDeleteTokenLock(
...
+
+                    IsEqual = TRUE;
+                    DPRINT1("SepCompareTokens(): Tokens are equal!\n");
+                }
+            }
         }
+    }
+    else
+    {
+        /* If none of the tokens are restricted then just begin checking for every privilege present */
```
HBelusca, line 242: This code is the same as the one above (that I mark with "****"), so you can group all that together.

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,138 @@ SepDeleteTokenLock(
...
+                    if (!RtlEqualSid(FirstToken->RestrictedSids[FirstTokenIndex].Sid,
+                                     SecondToken->RestrictedSids[SecondTokenIndex].Sid))
+                    {
+                        DPRINT1("SepCompareTokens(): Restricted SIDs are not equal!\n");
+                        goto Quit;
+                    }
+                }
+            }
+
+            /* Check if every privilege that is present in either token is also present in the other one */
```
HBelusca, line 213: ***** Code in common with the one below.

modules/rostests/apitests/ntdll/NtCompareTokens.c
```diff
@@ -0,0 +1,107 @@
...
+    HANDLE ProcessToken, DuplicatedToken, TokenNoPrivileges = NULL;
+    BOOLEAN IsEqual = FALSE;
+
+    /* Obtain some tokens from current process */
+    ProcessToken = GetTokenFromCurrentProcess();
+    TokenNoPrivileges = GetTokenFromCurrentProcess();
+    DuplicatedToken = GetDuplicateToken(ProcessToken);
+
+    /* Disable all the privileges for token */
+    DisableTokenPrivileges(TokenNoPrivileges);
```
HBelusca, line 72: Isn't that going to implicitly also disable the privileges of `ProcessToken` ? @ThFabba @tkreuzer , any idea about that?

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,138 @@ SepDeleteTokenLock(
...
+
+            /* Are the attributes from either SID from tokens equal? */
+            if (FirstToken->UserAndGroups[FirstTokenIndex].Attributes !=
+                SecondToken->UserAndGroups[SecondTokenIndex].Attributes)
+            {
+                DPRINT1("SepCompareTokens(): SID attributes are not equal!\n");
+                goto Quit;
+            }
+        }
+    }
```
zefklop, line 192: Is this expected that all the members of the UserAndGroups array are the same ?

I'd guess not. So please :

- Shortcut the loop if `FirstToken->UserAndGroupCount != SecondToken->UserAndGroupCount`
- Think again about why this is broken.

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,138 @@ SepDeleteTokenLock(
...
+            for (FirstTokenIndex = 0; FirstTokenIndex < FirstToken->PrivilegeCount; FirstTokenIndex++)
+            {
+                for (SecondTokenIndex = 0; SecondTokenIndex < SecondToken->PrivilegeCount; SecondTokenIndex++)
+                {
+                    /* Are the privileges from either token equal? */
+                    if (FirstToken->Privileges[FirstTokenIndex].Luid.LowPart !=
+                        SecondToken->Privileges[SecondTokenIndex].Luid.LowPart)
+                    {
+                        DPRINT1("SepCompareTokens(): Privileges are not equal!\n");
+                        goto Quit;
```
zefklop, line 223: continue

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,138 @@ SepDeleteTokenLock(
...
+                    }
+
+                    /* Are the privilege attributes from either token equal? */
+                    if (FirstToken->Privileges[FirstTokenIndex].Attributes !=
+                        SecondToken->Privileges[SecondTokenIndex].Attributes)
+                    {
+                        DPRINT1("SepCompareTokens(): Privilege attributes are not equal!\n");
+                        goto Quit;
```
zefklop, line 231: continue

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,138 @@ SepDeleteTokenLock(
...
+                    }
+
+                    IsEqual = TRUE;
+                    DPRINT1("SepCompareTokens(): Tokens are equal!\n");
```
zefklop, line 235: break

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,138 @@ SepDeleteTokenLock(
...
+                                     SecondToken->RestrictedSids[SecondTokenIndex].Sid))
+                    {
+                        DPRINT1("SepCompareTokens(): Restricted SIDs are not equal!\n");
+                        goto Quit;
+                    }
+                }
+            }
+
+            /* Check if every privilege that is present in either token is also present in the other one */
+            for (FirstTokenIndex = 0; FirstTokenIndex < FirstToken->PrivilegeCount; FirstTokenIndex++)
```
zefklop, line 214: Let me help for this one...

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,138 @@ SepDeleteTokenLock(
...
+            {
```
zefklop, line 215: `IsEqual = FALSE;`

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,138 @@ SepDeleteTokenLock(
...
+                    if (FirstToken->Privileges[FirstTokenIndex].Attributes !=
+                        SecondToken->Privileges[SecondTokenIndex].Attributes)
+                    {
+                        DPRINT1("SepCompareTokens(): Privilege attributes are not equal!\n");
+                        goto Quit;
+                    }
+
+                    IsEqual = TRUE;
+                    DPRINT1("SepCompareTokens(): Tokens are equal!\n");
+                }
```
zefklop, line 236: `if (!IsEqual)
{
    DPRINT1("Whatever you see fit.\n");
    goto Quit;
}`

zefklop (changes requested):
Please rewrite the broken loops.

ntoskrnl/se/token.c
HBelusca: > Is this expected that all the members of the UserAndGroups array are the same ?

To answer this question, a rostest that exactly builds two tokens that have some but not all such UserAndGroups attributes being the same, would be welcome.

GeoB99: I made some research in regard of this and what Windows Server 2003 does is checks first if both the `UserAndGroupCount` counter member from either tokens are equivalent in order to proceed looping for SIDs and their attributes. Otherwise W2k3 simply bails out if the counters are not the same. Such case applies for privileges too.

modules/rostests/apitests/ntdll/NtCompareTokens.c
GeoB99: Indeed, since `TokenNoPrivileges` is an opened token handle from the main process that'd implicitly disable the privilege(s) for the main process token as well. I didn't specifically explain that in the comments so apologies. Either way I should get rid of it within tests in the next push.

ntoskrnl/se/token.c
GeoB99: By using your suggestion, the code just simply jumps to the second loop with the first one never firing up to loop over the first array... 

Furthermore according to `SepComparePrivilegeAndAttributeArrays`'s behaviour if privileges and/or attributes are not equal it doesn't continue looping the arrays but instead it bails out and `NtCompareTokens` deems the tokens as not equivalent (same goes for SIDs and their attributes).

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,147 @@ SepDeleteTokenLock(
...
+    if (FirstToken->UserAndGroupCount != SecondToken->UserAndGroupCount)
+    {
+        DPRINT1("SepCompareTokens(): User & Group index counters are not the same!\n");
+        goto Quit;
+    }
+
+    /* Check if every SID that is present in either token is also present in the other one */
+    for (FirstTokenIndex = 0; FirstTokenIndex < FirstToken->UserAndGroupCount; FirstTokenIndex++)
+    {
+        for (SecondTokenIndex = FirstTokenIndex; SecondTokenIndex < SecondToken->UserAndGroupCount; SecondTokenIndex++)
```
zefklop, line 189: OK, let me tell you why this is wrong with an example (struct members simplified for clarity)

Say Token is the following : 
`{{AdminGroup, Attribute1}, {AdminUser, Attribute2}}`
Now apply your double loop to this token:

1. Token[0].Sid == Token[0].Sid --> OK
2. Token[0].Sid != Token[1].Sid --> Quit Loop

And your token is not equal to itself.

zefklop (changes requested):
Fix the loops !

GeoB99:
I rewrote the loops again, please let me know if there's something missing or else.

zefklop (changes requested):
FIX THE LOOOOOOOOOOPS

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,183 @@ SepDeleteTokenLock(
...
+            }
+
+            if (FirstToken->UserAndGroups[FirstTokenIndex].Attributes !=
+                SecondToken->UserAndGroups[SecondTokenIndex].Attributes)
+            {
+                DPRINT("SepCompareTokens(): SID attributes are not equal!\n");
+                continue;
+            }
+
+            goto UserGroupsLoop;
```
zefklop, line 206:
```suggestion
            break;
```

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,183 @@ SepDeleteTokenLock(
...
+
+        /*
+         * If we're here then we've exhausted the arrays
+         * and the elements are not equal to the either token.
+         */
+        IsEqual = FALSE;
+        goto Quit;
+
+        /* Go to the next iteration until we exhaust the arrays */
+        UserGroupsLoop:;
```
zefklop, line 217:
```suggestion
         if (SecondTokenIndex == SecondToken->UserAndGroupCount)
         {
              /* We've exhausted the array of the second token without finding this one */
              IsEqual = FALSE;
              goto Quit;
          }
```

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,183 @@ SepDeleteTokenLock(
...
+ */
+static
+NTSTATUS
+SepCompareTokens(
+    _In_ PTOKEN FirstToken,
+    _In_ PTOKEN SecondToken,
+    _Out_ PBOOLEAN Equal)
 {
-    BOOLEAN Restricted, IsEqual = FALSE;
+    BOOLEAN Restricted, IsEqual;
```
zefklop, line 169: Initialize IsEqual to FALSE here, and set it to TRUE at the end of the function, instead of setting it over the whole place.

ntoskrnl/se/token.c
zefklop, line 217: Labels inside loops are ugly.

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,183 @@ SepDeleteTokenLock(
...
+        DPRINT("SepCompareTokens(): User & Group index counters are not the same!\n");
+        IsEqual = FALSE;
+        goto Quit;
+    }
 
+    /* Check if every SID that is present in either token is also present in the other one */
+    for (FirstTokenIndex = 0; FirstTokenIndex < FirstToken->UserAndGroupCount; FirstTokenIndex++)
+    {
+        for (SecondTokenIndex = 0; SecondTokenIndex < SecondToken->UserAndGroupCount; SecondTokenIndex++)
+        {
```
zefklop, line 191: Please factor this loop in SepFindSidAndAttributeInArray or something. Repeating the same code over & over is terible.

zefklop (changes requested):
- [x] Fix loop styles (no labels inside of them, please)
- [x] Factor out common code

Otherwise, the logic now looks better.

zefklop (approved):
Looks good to me, now.

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,232 @@ SepDeleteTokenLock(
...
+
+    /* Loop over the SID arrays and compare them */
+    for (FirstCount = 0; FirstCount < CountSidArray1; FirstCount++)
+    {
+        for (SecondCount = 0; SecondCount < CountSidArray2; SecondCount++)
+        {
+            if (!RtlEqualSid(SidArrayToken1[FirstCount].Sid,
+                             SidArrayToken2[SecondCount].Sid))
+            {
+                DPRINT("SepCompareSidAndAttributesFromTokens(): SIDs are not equal!\n");
```
HBelusca, line 186: As the search loop is designed, this is not a failure case so the DPRINT is not needed.

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,232 @@ SepDeleteTokenLock(
...
+                continue;
+            }
+
+            if (SidArrayToken1[FirstCount].Attributes !=
+                SidArrayToken2[SecondCount].Attributes)
+            {
+                DPRINT("SepCompareSidAndAttributesFromTokens(): SID attributes are not equal!\n");
```
HBelusca, line 193: same

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,232 @@ SepDeleteTokenLock(
...
+
+    /* Loop over the privilege arrays and compare them */
+    for (FirstCount = 0; FirstCount < CountPrivArray1; FirstCount++)
+    {
+        for (SecondCount = 0; SecondCount < CountPrivArray2; SecondCount++)
+        {
+            if (PrivArrayToken1[FirstCount].Luid.LowPart !=
+                PrivArrayToken2[SecondCount].Luid.LowPart)
+            {
+                DPRINT("SepComparePrivilegeAndAttributesFromTokens(): Privileges are not equal!\n");
```
HBelusca, line 259: same

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,232 @@ SepDeleteTokenLock(
...
+                continue;
+            }
+
+            if (PrivArrayToken1[FirstCount].Attributes !=
+                PrivArrayToken2[SecondCount].Attributes)
+            {
+                DPRINT("SepComparePrivilegeAndAttributesFromTokens(): Privilege attributes are not equal!\n");
```
HBelusca, line 266: same

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,232 @@ SepDeleteTokenLock(
...
+    {
+        goto Quit;
+    }
 
+    /* Are both tokens restricted? */
     Restricted = SeTokenIsRestricted(FirstToken);
     if (Restricted == SeTokenIsRestricted(SecondToken))
     {
         if (Restricted)
         {
```
HBelusca, line 338: As a side comment, it would appear that these two if-s could be merged into a single one:
```
```suggestion
    Restricted = SeTokenIsRestricted(FirstToken);
    if (Restricted && (Restricted == SeTokenIsRestricted(SecondToken)))
    {
```

HBelusca (commented):
Some minor comments, but looks good otherwise!

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,232 @@ SepDeleteTokenLock(
...
+                                              SecondToken->UserAndGroupCount))
+    {
+        goto Quit;
+    }
 
+    /* Are both tokens restricted? */
     Restricted = SeTokenIsRestricted(FirstToken);
     if (Restricted == SeTokenIsRestricted(SecondToken))
     {
         if (Restricted)
```
HBelusca, line 337: The documentation explicitly says that the two tokens are considered equal if, amongst all other conditions that need to be satisfied, the following one is also TRUE:
```
Neither or both of the tokens are restricted.
```
In other words here, the code should be modified as follows:
```suggestion
    Restricted = SeTokenIsRestricted(FirstToken);
    if (Restricted != SeTokenIsRestricted(SecondToken))
    {
        /* The tokens do not have the same restrictions, fail */
        goto Quit;
    }
    /* If both tokens are restricted, check for SID restrictions in both tokens */
    if (Restricted)
    {
        // Here do the SepCompareSidAndAttributesFromTokens() call.
    }
```

HBelusca (changes requested):
Something to fix. @zefklop could you confirm?

ntoskrnl/se/token.c
GeoB99: Ah yeah, crap. I got confused by the wording of that part of the documentation. I'll fix that one asap.

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,235 @@ SepDeleteTokenLock(
...
+        return FALSE;
+    }
+
+    /* Loop over the privilege arrays and compare them */
+    for (FirstCount = 0; FirstCount < CountPrivArray1; FirstCount++)
+    {
+        for (SecondCount = 0; SecondCount < CountPrivArray2; SecondCount++)
+        {
+            if (PrivArrayToken1[FirstCount].Luid.LowPart !=
+                PrivArrayToken2[SecondCount].Luid.LowPart)
```
HBelusca, line 255: Why just looking at the LowPart? Why not using RtlEqualLuid?

HBelusca (approved):

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,235 @@ SepDeleteTokenLock(
...
+                continue;
+            }
+
+            if (SidArrayToken1[FirstCount].Attributes !=
+                SidArrayToken2[SecondCount].Attributes)
+            {
+                continue;
+            }
+
+            break;
```
Extravert-ir, line 195: Others may not agree with me, but I feel like this writing is much more clear
```
            PSID_AND_ATTRIBUTES Token1 = &SidArrayToken1[FirstCount];
            PSID_AND_ATTRIBUTES Token2 = &SidArrayToken2[SecondCount];

            if (RtlEqualSid(Token1->Sid, Token2->Sid) && Token1->Attributes == Token2->Attributes)
            {
                break;
            }
```
Same with the second function

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,227 @@ SepDeleteTokenLock(
...
+    /* Loop over the privilege arrays and compare them */
+    for (FirstCount = 0; FirstCount < CountPrivArray1; FirstCount++)
+    {
+        for (SecondCount = 0; SecondCount < CountPrivArray2; SecondCount++)
+        {
+            FirstPrivArray = &PrivArrayToken1[FirstCount];
+            SecondPrivArray = &PrivArrayToken2[SecondCount];
+
+            if (RtlEqualLuid(&FirstPrivArray->Luid, &SecondPrivArray->Luid) &&
+                             FirstPrivArray->Attributes == SecondPrivArray->Attributes)
```
Extravert-ir, line 255:
```suggestion
                FirstPrivArray->Attributes == SecondPrivArray->Attributes)
```

ntoskrnl/se/token.c
```diff
@@ -135,34 +135,227 @@ SepDeleteTokenLock(
...
+    /* Loop over the SID arrays and compare them */
+    for (FirstCount = 0; FirstCount < CountSidArray1; FirstCount++)
+    {
+        for (SecondCount = 0; SecondCount < CountSidArray2; SecondCount++)
+        {
+            FirstSidArray = &SidArrayToken1[FirstCount];
+            SecondSidArray = &SidArrayToken2[SecondCount];
+
+            if (RtlEqualSid(FirstSidArray->Sid, SecondSidArray->Sid) &&
+                            FirstSidArray->Attributes == SecondSidArray->Attributes)
```
Extravert-ir, line 188:
```suggestion
                FirstSidArray->Attributes == SecondSidArray->Attributes)
```

Extravert-ir (changes requested):

Extravert-ir (approved):

PR #3452, opened 2021-02-11
[NOTEPAD] Implement File->New Window

gonzoMD:
## Purpose

Let the user open another window without closing the current one.

## Proposed changes

- implement DIALOG_FileNewWindow() and use it

### DIFF
--- a/base/applications/notepad/dialog.c
+++ b/base/applications/notepad/dialog.c
@@ -415,6 +415,13 @@ VOID DIALOG_FileNew(VOID)
     }
 }
 
+VOID DIALOG_FileNewWindow(VOID)
+{
+    TCHAR pszNotepadExe[MAX_PATH];
+    GetModuleFileName(NULL, pszNotepadExe, ARRAYSIZE(pszNotepadExe));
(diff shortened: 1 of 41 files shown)


base/applications/notepad/dialog.c
```diff
@@ -415,6 +415,13 @@ VOID DIALOG_FileNew(VOID)
     }
 }
 
+VOID DIALOG_FileNewWindow(VOID)
+{
+    TCHAR pszNotepadExe[MAX_PATH];
+    GetModuleFileName(NULL, pszNotepadExe, MAX_PATH);
```
HBelusca, line 421:
```suggestion
    GetModuleFileName(NULL, pszNotepadExe, _countof(pszNotepadExe));
```
(or `ARRAYSIZE(pszNotepadExe)` in case `_countof` is not available).

HBelusca (commented):
Apart from the suggestion, LGTM.
You should perhaps add in the commit log that Windows does indeed start a new own's instance too when clicking on the New Window.
(I've checked this behaviour by looking at taskmgr, and testing notepad.exe that I renamed into something else.)

base/applications/notepad/dialog.c
gonzoMD: done

PR #3456, opened 2021-02-13
[PSDK] winbase.h: Update LoadLibraryEx() dwFlags value defines

SergeGautherie:
Add another Vista check.

Addendum to 986cc5e.


SergeGautherie:
Cc @katahiromz

katahiromz:
That is undocumented value. So, ...

PR #3459, opened 2021-02-14
[SHIMGVW] Fix wrong text into the title bar.

carlo-bramini:


gonzoMD:
Hey Carlo, thanks for the fix. Can you please take a look at line 1135 of shimgvw.c? The sizeof() stuff is wrong because it still uses sizeof(szBuf) / sizeof(TCHAR).
Use ARRAYSIZE(szBuf) instead

HBelusca:
This also shows that not all the functions in the code were explicitly Unicode, so maybe you can change these as well.

learn-more (approved):

katahiromz:
Make CallWindowProc and DefWindowProc calls Unicode.

PR #3462, opened 2021-02-15
[TIMEDATE] Improve analog clock appearance

katahiromz:
## Purpose

Improve analog clock appearance of Control Panel Applet `timedate`.

## Proposed changes

- Totally rewrite the clock-related code.
- Emulates line width by `WidenPath` trick.

BEFORE:
before
AFTER:
VirtualBox_ReactOS_15_02_2021_22_28_00

### DIFF
--- a/dll/cpl/timedate/clock.c
+++ b/dll/cpl/timedate/clock.c
@@ -48,7 +48,7 @@ SetClockCoodinates(LPPOINT pPoint, POINT Center, INT Radius, INT Angle)
 }
 
 static inline VOID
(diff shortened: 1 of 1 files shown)


katahiromz:
`timedate.cpl` on Win2k3:
2k3

katahiromz:
We can change the colors as we like. For example, red, blue or darkgreen etc.

binarymaster (commented):
Would be better do make it switchable, as some people might prefer previous style instead of the new one, or vice versa.

katahiromz:
Closed.

PR #3471, opened 2021-02-20
[SHIMGVW] Use _countof instead of ARRAYSIZE and use UINT

katahiromz:
## Purpose

Improve code for standardization and erasing warning.

JIRA issue: N/A

## Proposed changes

- Use `_countof` macro instead of non-standard `ARRAYSIZE` macro.
- Use `UINT` type for array index instead of `INT` type, for erasing signed/unsigned comparison warning.

### DIFF
--- a/dll/win32/shimgvw/shimgvw.c
+++ b/dll/win32/shimgvw/shimgvw.c
@@ -243,7 +243,7 @@ BOOL Anime_Step(DWORD *pdwDelay)
 
 static void ZoomInOrOut(BOOL bZoomIn)
 {
-    INT i;
+    SIZE_T i;
 
     if (image == NULL)
         return;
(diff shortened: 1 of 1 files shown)


dll/win32/shimgvw/shimgvw.c
```diff
@@ -243,20 +243,20 @@ BOOL Anime_Step(DWORD *pdwDelay)
 
 static void ZoomInOrOut(BOOL bZoomIn)
 {
-    INT i;
+    SIZE_T i;
```
binarymaster, line 246: I'm not sure about using `SIZE_T` for a loop iterator variable, however let's see what anyone else can say about it.

HBelusca: On the one side you remove a compiler's warning; on the other side these arrays have very few elements so that the count of these elements can hold even in a int8 or int16 😂

katahiromz: `UINT` is better?

PR #3473, opened 2021-02-21
[SHIMGVW] Improve zooming feature and realize image edit

carlo-bramini:


dll/win32/shimgvw/shimgvw.c
```diff
@@ -1098,6 +1174,8 @@ ImageView_CreateWindow(HWND hwnd, LPCWSTR szFileName)
 
     InitCommonControlsEx(&Icc);
 
+    CoInitializeEx(NULL, COINIT_APARTMENTTHREADED | COINIT_DISABLE_OLE1DDE);
```
binarymaster, line 1177: Hm... I don't see how you use COM here yet. :thinking:

dll/win32/shimgvw/shimgvw.c
```diff
@@ -921,9 +939,9 @@ ImageView_InitControls(HWND hwnd)
 
     if (shiSettings.Maximized) ShowWindow(hwnd, SW_MAXIMIZE);
 
-    hDispWnd = CreateWindowEx(WS_EX_CLIENTEDGE, WC_STATIC, _T(""),
-                              WS_CHILD | WS_VISIBLE,
-                              0, 0, 0, 0, hwnd, NULL, hInstance, NULL);
+    hDispWnd = CreateWindowExW(WS_EX_CLIENTEDGE, WC_STATIC, _T(""),
```
GeoB99, line 942: You're explicitly using the Unicode variant of `CreateWindow(Ex)` yet in the `lpWindowName` parameter you're agnostic about which type of character string to use.
```suggestion
    hDispWnd = CreateWindowExW(WS_EX_CLIENTEDGE, WC_STATIC, L"",
```

dll/win32/shimgvw/shimgvw.c
carlo-bramini: https://docs.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-shellexecuteexa

dll/win32/shimgvw/shimgvw.c
carlo-bramini: Argh, you are right. Fixed.

dll/win32/shimgvw/shimgvw.c
carlo-bramini: immagine

binarymaster: Ah I see, alright then.

dll/win32/shimgvw/shimgvw.c
```diff
@@ -241,9 +242,41 @@ BOOL Anime_Step(DWORD *pdwDelay)
...
+    /* Check if a zoom button of the toolbar must be grayed */
+    bEnableZoomIn = TRUE;
+    bEnableZoomOut = TRUE;
+
+    if (NewZoom >= MAX_ZOOM)
+    {
+        bEnableZoomIn = FALSE;
+    }
+    else
+    if (NewZoom <= MIN_ZOOM)
```
katahiromz, line 264:
```suggestion
    else if (NewZoom <= MIN_ZOOM)
```

dll/win32/shimgvw/shimgvw.c
```diff
@@ -241,9 +242,41 @@ BOOL Anime_Step(DWORD *pdwDelay)
...
+
+    /* If zoom has not been changed, ignore it */
+    if (ZoomPercents == NewZoom)
+        return;
+
+    ZoomPercents = NewZoom;
+
+    /* Check if a zoom button of the toolbar must be grayed */
+    bEnableZoomIn = TRUE;
+    bEnableZoomOut = TRUE;
```
katahiromz, line 257:
```suggestion
    bEnableZoomIn = bEnableZoomOut = TRUE;
```

dll/win32/shimgvw/shimgvw.c
```diff
@@ -959,6 +977,58 @@ ImageView_OnSize(HWND hwnd, UINT state, INT cx, INT cy)
...
+{
+    DPRINT1("ImageView_Delete: unimplemented.\n");
+    return 0;
+}
+
+static LRESULT CALLBACK
+ImageView_Modify(HWND hwnd)
+{
+    int nChars = GetFullPathNameW(currentFile->FileName, 0, NULL, NULL);
+    WCHAR *lpFPN;
```
katahiromz, line 991:
```suggestion
    LPWSTR pszPathName;
```

dll/win32/shimgvw/shimgvw.c
```diff
@@ -959,6 +977,58 @@ ImageView_OnSize(HWND hwnd, UINT state, INT cx, INT cy)
...
+    SHELLEXECUTEINFOW sei;
+
+    if (!nChars)
+    {
+        DPRINT1("ImageView_Modify: failed to get full path name.\n");
+        return 1;
+    }
+
+    lpFPN = (WCHAR *)malloc(nChars * sizeof(WCHAR));
```
katahiromz, line 1000:
```suggestion
    pszPathName = HeapAlloc(GetProcessHeap(), HEAP_ZERO_MEMORY, nChars * sizeof(WCHAR));
```

dll/win32/shimgvw/shimgvw.c
```diff
@@ -959,6 +977,58 @@ ImageView_OnSize(HWND hwnd, UINT state, INT cx, INT cy)
...
+    if (lpFPN == NULL)
+    {
+        DPRINT1("malloc() failed in ImageView_Modify()\n");
+        return 1;
+    }
+
+    GetFullPathNameW(currentFile->FileName,
+                     nChars,
+                     lpFPN,
+                     NULL);
```
katahiromz, line 1010:
```suggestion
    GetFullPathNameW(currentFile->FileName, nChars, pszPathName, NULL);
```

dll/win32/shimgvw/shimgvw.c
```diff
@@ -959,6 +977,58 @@ ImageView_OnSize(HWND hwnd, UINT state, INT cx, INT cy)
...
+    sei.lpDirectory = NULL;
+    sei.nShow = SW_SHOWNORMAL;
+    sei.hInstApp = NULL;
+
+    if (!ShellExecuteExW(&sei))
+    {
+        DPRINT1("ImageView_Modify: ShellExecuteExW() failed with code %08X\n", (int)GetLastError());
+    }
+
+    free(lpFPN);
```
katahiromz, line 1027:
```suggestion
    HeapFree(GetProcessHeap(), 0, pszPathName);
```

dll/win32/shimgvw/shimgvw.c
```diff
@@ -959,6 +977,58 @@ ImageView_OnSize(HWND hwnd, UINT state, INT cx, INT cy)
     }
 }
 
+static LRESULT CALLBACK
```
katahiromz, line 980:
```suggestion
static LRESULT
```
This is not callback.

dll/win32/shimgvw/shimgvw.c
```diff
@@ -959,6 +977,58 @@ ImageView_OnSize(HWND hwnd, UINT state, INT cx, INT cy)
...
+ImageView_Delete(HWND hwnd)
+{
+    DPRINT1("ImageView_Delete: unimplemented.\n");
+    return 0;
+}
+
+static LRESULT CALLBACK
```
katahiromz, line 987:
```suggestion
static LRESULT
```
This is not callback.

dll/win32/shimgvw/shimgvw.c
```diff
@@ -959,6 +977,58 @@ ImageView_OnSize(HWND hwnd, UINT state, INT cx, INT cy)
...
+    if (!nChars)
+    {
+        DPRINT1("ImageView_Modify: failed to get full path name.\n");
+        return 1;
+    }
+
+    lpFPN = (WCHAR *)malloc(nChars * sizeof(WCHAR));
+    if (lpFPN == NULL)
+    {
+        DPRINT1("malloc() failed in ImageView_Modify()\n");
```
katahiromz, line 1003:
```suggestion
        DPRINT1("HeapAlloc() failed in ImageView_Modify()\n");
```

dll/win32/shimgvw/shimgvw.c
```diff
@@ -959,6 +977,58 @@ ImageView_OnSize(HWND hwnd, UINT state, INT cx, INT cy)
...
+    GetFullPathNameW(currentFile->FileName,
+                     nChars,
+                     lpFPN,
+                     NULL);
+
+    sei.cbSize = sizeof(sei);
+    sei.fMask = 0;
+    sei.hwnd = NULL;
+    sei.lpVerb = L"edit";
+    sei.lpFile = lpFPN;
```
katahiromz, line 1016:
```suggestion
    sei.lpFile = pszPathName;
```

katahiromz (commented):
Good job.

dll/win32/shimgvw/shimgvw.c
carlo-bramini: The whole source is using malloc() so, when I had written this patch, I prefered to use that instead of HeapAlloc(). And why HEAP_ZERO_MEMORY? The buffer is allocated of the exact size we need, from the information we got from previous call to GetFullPathNameW(). There is absolutely no need to do so.

dll/win32/shimgvw/shimgvw.c
carlo-bramini: Why? Since `ImageView_WndProc` calls it as `return ImageView_Delete(hwnd);` (and `ImageView_Modify`) is the same, the compiler can generate better code if that function is also `__stdcall`.

dll/win32/shimgvw/shimgvw.c
katahiromz: HEAP_ZERO_MEMORY is unnecessary. Sorry.

dll/win32/shimgvw/shimgvw.c
carlo-bramini: I could do it, but this is against the coding rules ("Each statement should get its own line.").

binarymaster: Coding Style wiki page does not cover `else if` part, however... https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=else+if

dll/win32/shimgvw/shimgvw.c
katahiromz: So that you might use STDCALL.

dll/win32/shimgvw/shimgvw.c
learn-more: besides, `else if` is considered as 1 statement.

dll/win32/shimgvw/shimgvw.c
learn-more: I doubt this will make the compiler generate better code,
and in cases like this making an assumption like that can even cause the compiler to generate worse code.

Considering this is not going to be a hot path, those optimizations (and especially, optimizations based on guesses) are better not added.

dll/win32/shimgvw/shimgvw.c
katahiromz: ‘CoUninitialize` call and return value check is needed.

dll/win32/shimgvw/shimgvw.c
```diff
@@ -921,9 +937,9 @@ ImageView_InitControls(HWND hwnd)
 
     if (shiSettings.Maximized) ShowWindow(hwnd, SW_MAXIMIZE);
 
-    hDispWnd = CreateWindowEx(WS_EX_CLIENTEDGE, WC_STATIC, _T(""),
-                              WS_CHILD | WS_VISIBLE,
-                              0, 0, 0, 0, hwnd, NULL, hInstance, NULL);
+    hDispWnd = CreateWindowExW(WS_EX_CLIENTEDGE, WC_STATIC, L"",
```
katahiromz, line 940:
```suggestion
    hDispWnd = CreateWindowExW(WS_EX_CLIENTEDGE, WC_STATIC, NULL,
```
Technically `NULL` for window text is valid.

learn-more: but is it an improvement?

dll/win32/shimgvw/shimgvw.c
```diff
@@ -959,6 +975,55 @@ ImageView_OnSize(HWND hwnd, UINT state, INT cx, INT cy)
...
+ImageView_Delete(HWND hwnd)
+{
+    DPRINT1("ImageView_Delete: unimplemented.\n");
+    return 0;
+}
+
+static LRESULT
+ImageView_Modify(HWND hwnd)
+{
+    int nChars = GetFullPathNameW(currentFile->FileName, 0, NULL, NULL);
```
HBelusca, line 988: Maybe it's cleaner to make the function call where you just below it check for failure, not many lines before.

dll/win32/shimgvw/shimgvw.c
```diff
@@ -1177,6 +1255,10 @@ ImageView_CreateWindow(HWND hwnd, LPCWSTR szFileName)
     Anime_FreeInfo();
 
     GdiplusShutdown(gdiplusToken);
+
+    /* Release COM resources */
+    CoUninitialize();
```
katahiromz, line 1260:
```suggestion
    if (SUCCEEDED(hComRes))
        CoUninitialize();
```

carlo-bramini: Actually, if `CoInitializeEx` failed, then `CoUninitialize` does nothing, but ok.

learn-more: > To close the COM library gracefully, each successful call to **CoInitialize** or **CoInitializeEx**, including those that return S_FALSE, must be balanced by a corresponding call to CoUninitialize.

From https://docs.microsoft.com/en-us/windows/win32/api/objbase/nf-objbase-coinitialize

katahiromz (approved):
Let's go.

JoachimHenze:
@Carlo-bramini what is the OLE32 CoInitializeEx()/CoUnitialize() stuff used for? Has that something to do with drag/drop-support? Can you elaborate please what we would lose without that?

HBelusca:
You have certainly noticed that this dll also registers as a COM server and also calls Shell API functions? (that requires COM initialization)

JoachimHenze:
Yes, so do I read/interpret correctly that the OLE32 CoInitializeEx()/CoUnitialize() stuff is solely used for the _modify-feature, but is not needed at all for improving the zooming-feature?

carlo-bramini:
immagine

PR #3474, opened 2021-02-21
[USETUP] Display the "check the FS" section as a MUI page

GeoB99:
`Check the filesystem` is the only section of USETUP that doesn't belong to its own MUI page, `CHECK_FILE_SYSTEM_PAGE`, but instead the contents inside that section are hardcoded in `CheckFileSystemPage` page function. With that said, display such section as a MUI page so that stuff can be localised.

### DIFF
--- a/base/setup/usetup/mui.h
+++ b/base/setup/usetup/mui.h
@@ -124,7 +124,6 @@ MUISetStyledText(
 #define STRING_NONFORMATTEDSYSTEMPART    62
 #define STRING_NONFORMATTEDOTHERPART     63
 #define STRING_INSTALLONPART             10
-#define STRING_CHECKINGPART              11
 #define STRING_CONTINUE                  12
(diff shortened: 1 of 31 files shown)


base/setup/usetup/mui.h
```diff
@@ -102,67 +102,66 @@ MUISetStyledText(
 /* MUI Text IDs */
 
 /* Static MUI Text */
-#define TEXT_ID_STATIC (-1)
+#define TEXT_ID_STATIC                        (-1)
```
HBelusca, line 105: Please avoid reformatting the whole file for the moment (as it may collide with other mui.h changes in other PRs, and definitively collides with other changes I have in the GPT support improvement branches.)
Just do the change of adding or removing the string/page IDs , and without shifting the other values.

HBelusca (changes requested):

HBelusca (approved):

PR #3476, opened 2021-02-22
[NTOS:OB] Do not close the handle if it's granted access to ObpAccess…

GeoB99:
…ProtectCloseBit
## Purpose

As of now the Object Manager private service, ObpCloseHandleTableEntry, looks for OBJ_PROTECT_CLOSE attribute if a handle should not be closed. However, in ObDuplicateObject if an attribute of OBJ_PROTECT_CLOSE is found as it's been filled to the caller (see L2466) this attribute is removed from the attributes list of the new handle and ObpAccessProtectCloseBit access is granted to the newly duplicated object handle.

With that being said ObpCloseHandleTableEntry indiscriminately closes the object handle albeit it shouldn't do so. As a matter of fact in Windows Server 2003 SP2 this service indeed checks for ObpAccessProtectCloseBit flag bit and if the condition is met then it returns STATUS_HANDLE_NOT_CLOSABLE as it should. Therefore we should do the same.

Now NtClose can properly warn the calling thread the object handle can't be closed which fixes a testcase failure within NtDuplicateObject NTDLL APITEST where this function gives handle close protection bit as requested by the caller.

### DIFF
--- a/ntoskrnl/include/internal/ob.h
+++ b/ntoskrnl/include/internal/ob.h
@@ -54,6 +54,11 @@
                                                          OBJ_AUDIT_OBJECT_CLOSE)
 
 //
+// Handle Access Protection Close Flag
+//
+#define ObpAccessProtectCloseBit 0x02000000L
+
+//
 // Identifies a Kernel Handle
 //
 #ifdef _WIN64
--- a/ntoskrnl/ob/obhandle.c
+++ b/ntoskrnl/ob/obhandle.c
@@ -18,7 +18,6 @@
 #include <debug.h>
 
 PHANDLE_TABLE ObpKernelHandleTable = NULL;
-ULONG ObpAccessProtectCloseBit = MAXIMUM_ALLOWED;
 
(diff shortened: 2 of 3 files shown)


GeoB99:
**BEFORE**
Capture
**AFTER**
Capture2
**Windows Server 2003**
Capture3

GeoB99:
I took a quick glance at the references of this global variable and it seems it's used in some other places in W2k3 SP2 whereas in ReactOS it's used so far in `ObpSetHandleAttributes`, `ObDuplicateObject` and finally in `ObpCloseHandleTableEntry` with this PR.

Although I'd rather work on that in a different PR regarding the other cases in our code. This PR is for the sake of fixing `ObpCloseHandleTableEntry` so that it won't close a protected object handle (and thus making the APITEST to stop moaning about the aforementioned failure).

HBelusca:
My comment was so that the protected close bit get actually handled consistently everywhere ^^

PR #3477, opened 2021-02-22
[NTOS:MM] Fail NP pool allocations when running out of pages

zefklop:
## Purpose

Instead of blindly asking for pages when allocating and happily crash, just fail the allocation and let the caller handle that.

### DIFF
--- a/ntoskrnl/mm/ARM3/expool.c
+++ b/ntoskrnl/mm/ARM3/expool.c
@@ -1446,14 +1446,14 @@ ExGetPoolTagInfo(IN PSYSTEM_POOLTAG_INFORMATION SystemInformation,
 }
 
 _IRQL_requires_(DISPATCH_LEVEL)
+static
 BOOLEAN
-NTAPI
-ExpExpandBigPageTable(
-    _In_ _IRQL_restores_ KIRQL OldIrql)
+ExpReallocateBigPageTable(
+    _In_ _IRQL_restores_ KIRQL OldIrql,
+    _In_ BOOLEAN Shrink)
 {
-    ULONG OldSize = PoolBigPageTableSize;
-    ULONG NewSize = 2 * OldSize;
-    ULONG NewSizeInBytes;
+    SIZE_T OldSize = PoolBigPageTableSize;
+    SIZE_T NewSize, NewSizeInBytes;
     PPOOL_TRACKER_BIG_PAGES NewTable;
     PPOOL_TRACKER_BIG_PAGES OldTable;
     ULONG i;
@@ -1465,11 +1465,41 @@ ExpExpandBigPageTable(
     ASSERT(KeGetCurrentIrql() == DISPATCH_LEVEL);
 
     /* Make sure we don't overflow */
-    if (!NT_SUCCESS(RtlULongMult(2,
-                                 OldSize * sizeof(POOL_TRACKER_BIG_PAGES),
-                                 &NewSizeInBytes)))
+    if (Shrink)
     {
-        DPRINT1("Overflow expanding big page table. Size=%lu\n", OldSize);
+        NewSize = OldSize / 2;
+
+        /* Make sure we don't shrink too much. */
+        ASSERT(NewSize >= ExpPoolBigEntriesInUse);
+
+        NewSize = ALIGN_UP_BY(NewSize, PAGE_SIZE / sizeof(POOL_TRACKER_BIG_PAGES));
+        ASSERT(NewSize <= OldSize);
+
+        /* If there is only one page left, then keep it around. Not a failure either. */
+        if (NewSize == OldSize)
+        {
+            ASSERT(NewSize == (PAGE_SIZE / sizeof(POOL_TRACKER_BIG_PAGES)));
+            KeReleaseSpinLock(&ExpLargePoolTableLock, OldIrql);
+            return TRUE;
+        }
+    }
+    else
+    {
+        if (!NT_SUCCESS(RtlSIZETMult(2, OldSize, &NewSize)))
+        {
+            DPRINT1("Overflow expanding big page table. Size=%lu\n", OldSize);
+            KeReleaseSpinLock(&ExpLargePoolTableLock, OldIrql);
+            return FALSE;
+        }
+
+        /* Make sure we don't stupidly waste pages */
+        NewSize = ALIGN_DOWN_BY(NewSize, PAGE_SIZE / sizeof(POOL_TRACKER_BIG_PAGES));
+        ASSERT(NewSize > OldSize);
+    }
+
+    if (!NT_SUCCESS(RtlSIZETMult(sizeof(POOL_TRACKER_BIG_PAGES), NewSize, &NewSizeInBytes)))
+    {
+        DPRINT1("Overflow while calculating big page table size. Size=%lu\n", OldSize);
         KeReleaseSpinLock(&ExpLargePoolTableLock, OldIrql);
         return FALSE;
     }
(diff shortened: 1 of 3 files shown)


ntoskrnl/mm/ARM3/expool.c
```diff
@@ -62,7 +62,7 @@ ULONGLONG MiLastPoolDumpTime;
  * pool code, but only for checked builds.
  *
  * As of Vista, however, an MSDN Blog entry by a Security Team Manager indicates
- * that these checks are done even on retail builds, due to the increasing
+ * that these checks are done even on retail builds, dNonPagedPoolDescriptorue to the increasing
```
Doug-Lyons, line 65: There appears to be an obvious copy-paste error here in the middle of 'due'.

zefklop: Fixed, thanks !

X547:
Does it fix kernel crash when application is out of memory ([test program](https://jira.reactos.org/secure/attachment/57450/AvalibleMem%20%28heap%29.c))?

ntoskrnl/mm/ARM3/expool.c
```diff
@@ -1465,11 +1465,45 @@ ExpExpandBigPageTable(
     ASSERT(KeGetCurrentIrql() == DISPATCH_LEVEL);
 
     /* Make sure we don't overflow */
-    if (!NT_SUCCESS(RtlULongMult(2,
-                                 OldSize * sizeof(POOL_TRACKER_BIG_PAGES),
-                                 &NewSizeInBytes)))
+    if (!Shrink)
```
Extravert-ir, line 1468: Maybe just change the name to `Expand`? Or swap the cases

ntoskrnl/mm/ARM3/expool.c
```diff
@@ -1704,8 +1739,20 @@ ExpFindAndRemoveTagBigPages(IN PVOID Va,
     // the lock and return the tag that was located
     //
     InterlockedIncrement((PLONG)&Entry->Va);
```
Extravert-ir, line 1741: We're holding the `ExpLargePoolTableLock`, is the atomic operation still needed?

ntoskrnl/mm/ARM3/pool.c
```diff
@@ -849,6 +849,19 @@ MiAllocatePoolPages(IN POOL_TYPE PoolType,
...
 
+    /* Check that we have enough available pages for this request */
+    if (MmAvailablePages < SizeInPages)
+    {
+        MiReleasePfnLockFromDpcLevel();
+        KeReleaseQueuedSpinLock(LockQueueMmNonPagedPoolLock, OldIrql);
+
+        MiReleaseSystemPtes(StartPte, SizeInPages, NonPagedPoolExpansion);
+
+        DPRINT1("OUT OF AVAILABLE PAGES!\n");
```
Extravert-ir, line 860: I'd put `MmAvailablePages` and `SizeInPages` here

ntoskrnl/mm/ARM3/expool.c
```diff
@@ -1465,11 +1465,45 @@ ExpExpandBigPageTable(
...
+        NewSize = ALIGN_DOWN_BY(NewSize, PAGE_SIZE / sizeof(POOL_TRACKER_BIG_PAGES));
+        ASSERT(NewSize > OldSize);
+    }
+    else
     {
-        DPRINT1("Overflow expanding big page table. Size=%lu\n", OldSize);
+        NewSize = OldSize / 2;
+
+        /* Make sure we don't shrink too much. This is not a failure. */
+        if (NewSize < ExpPoolBigEntriesInUse)
```
Extravert-ir, line 1486: With the only usage inside a condition `if (ExpPoolBigEntriesInUse < (PoolBigPageTableSize / 8))`, this shouldn't happen, right?

ntoskrnl/mm/ARM3/expool.c
```diff
@@ -1465,11 +1465,45 @@ ExpExpandBigPageTable(
...
+        if (!NT_SUCCESS(RtlULongMult(2, OldSize, &NewSize)))
+        {
+            DPRINT1("Overflow expanding big page table. Size=%lu\n", OldSize);
+            KeReleaseSpinLock(&ExpLargePoolTableLock, OldIrql);
+            return FALSE;
+        }
+
+        /* Make sure we don't stupidly waste pages */
+        NewSize = ALIGN_DOWN_BY(NewSize, PAGE_SIZE / sizeof(POOL_TRACKER_BIG_PAGES));
+        ASSERT(NewSize > OldSize);
```
Extravert-ir, line 1479: This will trigger only if `NewSize` is equal to `OldSize`, aka both `NewSize` and `OldSize` are less than `PAGE_SIZE / sizeof(POOL_TRACKER_BIG_PAGES)`.
Can this happen?

ntoskrnl/mm/ARM3/expool.c
```diff
@@ -1446,14 +1446,14 @@ ExGetPoolTagInfo(IN PSYSTEM_POOLTAG_INFORMATION SystemInformation,
...
-ExpExpandBigPageTable(
-    _In_ _IRQL_restores_ KIRQL OldIrql)
+ExpReallocateBigPageTable(
+    _In_ _IRQL_restores_ KIRQL OldIrql,
+    _In_ BOOLEAN Shrink)
 {
     ULONG OldSize = PoolBigPageTableSize;
-    ULONG NewSize = 2 * OldSize;
-    ULONG NewSizeInBytes;
+    ULONG NewSize, NewSizeInBytes;
```
Extravert-ir, line 1456: Not really related to the PR, but shouldn't the size of the table be `size_t`?

ntoskrnl/mm/ARM3/expool.c
```diff
@@ -1704,8 +1739,20 @@ ExpFindAndRemoveTagBigPages(IN PVOID Va,
     // the lock and return the tag that was located
     //
     InterlockedIncrement((PLONG)&Entry->Va);
-    InterlockedDecrementUL(&ExpPoolBigEntriesInUse);
-    KeReleaseSpinLock(&ExpLargePoolTableLock, OldIrql);
+
+    ExpPoolBigEntriesInUse--;
+
+    /* If reaching 12.5% of the size (or whatever integer rounding gets us to),
+     * halve the allocation size, which will get us to 25% of space used. */
```
Extravert-ir, line 1746: Are these numbers connected with 75% from above?
I think it makes sense to create a define so the value can be easily tuned. Somebody may want to do a benchmark for that and find the best value :D

zefklop: Well, this is just what happens when you play with halving & doubling sizes.

The idea is to keep the occupation rate in the 25% - 75% range. This could as well be 33 - 66, 12 - 88 or whatever.

I'll see to have some defines.

ntoskrnl/mm/ARM3/pool.c
zefklop: Makes sense, yes.

ntoskrnl/mm/ARM3/expool.c
zefklop: Alright.

ntoskrnl/mm/ARM3/expool.c
zefklop: Alright.

ntoskrnl/mm/ARM3/expool.c
zefklop: No, hence the assertion.

ntoskrnl/mm/ARM3/expool.c
zefklop: Hmmm, I remember putting an ASSERT here and triggering it, and this made sense to turn it into a check back then. Let me look again.

ntoskrnl/mm/ARM3/expool.c
zefklop: No. And setting bit 0 like this is bizarre.

ntoskrnl/mm/ARM3/expool.c
zefklop: Done

ntoskrnl/mm/ARM3/expool.c
zefklop: Fixed.

PR #3481, opened 2021-02-23
[NTOS:PNP] Implement PlugPlayControlStartDevice control class and use it

Extravert-ir:
This adds another action for starting devices after usermode PnP service enables the driver for it.
`PlugPlayControlResetDevice` is wrong in this case, because it returns the device node back to `DeviceNodeUninitialized` state, which is an overkill here (initialization phase is completed successfully, no need to repeat it)

Should fix: [CORE-17490](https://jira.reactos.org/browse/CORE-17490), [CORE-17463](https://jira.reactos.org/browse/CORE-17463)

### DIFF
--- a/base/setup/usetup/devinst.c
+++ b/base/setup/usetup/devinst.c
@@ -42,11 +42,11 @@ static BOOLEAN
 ResetDevice(
     IN LPCWSTR DeviceId)
 {
-    PLUGPLAY_CONTROL_RESET_DEVICE_DATA ResetDeviceData;
+    PLUGPLAY_CONTROL_DEVICE_CONTROL_DATA ResetDeviceData;
     NTSTATUS Status;
 
     RtlInitUnicodeString(&ResetDeviceData.DeviceInstance, DeviceId);
-    Status = NtPlugPlayControl(PlugPlayControlResetDevice, &ResetDeviceData, sizeof(PLUGPLAY_CONTROL_RESET_DEVICE_DATA));
+    Status = NtPlugPlayControl(PlugPlayControlResetDevice, &ResetDeviceData, sizeof(PLUGPLAY_CONTROL_DEVICE_CONTROL_DATA));
     if (!NT_SUCCESS(Status))
     {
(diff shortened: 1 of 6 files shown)


binarymaster (commented):
Tested, achieved two improvements, one remains the same :)

> [CORE-17490](https://jira.reactos.org/browse/CORE-17490)

:heavy_check_mark: Tested on VirtualBox with LiveCD - `ISAPNP read data port` now appears in Device Manager.

> [CORE-17458](https://jira.reactos.org/browse/CORE-17458)

:heavy_check_mark: Networking works now both in LiveCD and in 2nd stage setup.

> [CORE-17463](https://jira.reactos.org/browse/CORE-17463)

:heavy_multiplication_x: PS/2 keyboard and mouse in 3rd stage with Standard HAL still not working for me, the same error code in Device Manager.

base/setup/usetup/devinst.c
```diff
@@ -39,20 +39,25 @@ typedef struct
...
-    RtlInitUnicodeString(&ResetDeviceData.DeviceInstance, DeviceId);
-    Status = NtPlugPlayControl(PlugPlayControlResetDevice, &ResetDeviceData, sizeof(PLUGPLAY_CONTROL_RESET_DEVICE_DATA));
-    if (!NT_SUCCESS(Status))
+    RtlInitUnicodeString(&PlugPlayData.DeviceInstance, DeviceId);
+    PlugPlayData.Operation = PNP_GET_DEVICE_STATUS;
+
+    Status = NtPlugPlayControl(PlugPlayControlDeviceStatus, &PlugPlayData, sizeof(PlugPlayData));
+    if (NT_SUCCESS(Status))
+    {
+        return (_Bool)((PlugPlayData.DeviceStatus & DN_DRIVER_LOADED) &&
```
HBelusca, line 54:
```suggestion
        return (BOOLEAN)((PlugPlayData.DeviceStatus & DN_DRIVER_LOADED) &&
```
(you can use the same return type as the function.)

PR #3485, opened 2021-02-24
[CMAKE:GCC] Fixup of driver sections

zefklop:
## Purpose

GCC (or, more precisely, LD) is unable to set some section characteristics, such as discardable or not-paged flag. 

JIRA issue: [CORE-17494](https://jira.reactos.org/browse/CORE-17494)

## Proposed changes

Use pefixup and linker scripts to do the right thing

- Put INIT section just before .reloc, and not after (as does MSVC)
- Have pefixup utility add the relevant flags to well-known sections in order to make them (non-)pageable, discardable.

## TODO

None so far.


HBelusca (changes requested):
This is quite an invasive hack that also touches files used with MSVC builds (that have nothing to do with GCC/LD). The only proper patch is to fix LD.

zefklop:
> This is quite an invasive hack that also touches files used with MSVC builds (that have nothing to do with GCC/LD). The only proper patch is to fix LD.

No it doesn't. pefixup is only used in GCC build.

ntoskrnl/ntkrnlmp/CMakeLists.txt
```diff
@@ -19,35 +19,16 @@ add_executable(ntkrnlmp
...
-    set_entrypoint(ntkrnlmp KiSystemStartup)
-endif()
-set_subsystem(ntkrnlmp native)
-
-if(MSVC)
-    set_image_base(ntkrnlmp 0x00400000)
-    add_target_link_flags(ntkrnlmp "/SECTION:.rsrc,!DP") #Accessed from bugcheck code
-else()
-    set_image_base(ntkrnlmp 0x80800000)
-endif()
```
HBelusca, line 38: Not sure it makes sense to move these things away, because there is just "one" kernel, so no need to create module type just for it. (Same comment as for the ntoskrnl cmakefile)

zefklop: If you look closely, there is already /SECTION:INIT,xxxx missing for ntkrnlmp

CMakeMacros.cmake, gcc.cmake & msvc.cmake were made to hide toolchain specifics, avoid duplicating configuration & the like, which is exactly what the PR is doing. What's your point in being so nit-picky ?

Extravert-ir:
How are we going to avoid [ROSBE-154](https://jira.reactos.org/browse/ROSBE-154)?

zefklop:
> How are we going to avoid [ROSBE-154](https://jira.reactos.org/browse/ROSBE-154)?

By putting the INIT section before .reloc, instead of after. Like link.exe does

sdk/cmake/CMakeMacros.cmake
```diff
@@ -642,18 +626,33 @@ function(set_module_type MODULE TYPE)
...
         endif()
-    elseif((${TYPE} STREQUAL kernelmodedriver) OR (${TYPE} STREQUAL wdmdriver) OR (${TYPE} STREQUAL kerneldll))
-        set_image_base(${MODULE} 0x00010000)
+    elseif(TYPE IN_LIST KERNEL_MODULE_TYPES)
+        # special case for kernel
+        if (TYPE STREQUAL kernel)
+            if (MSVC)
+                set_image_base(${MODULE} 0x00400000)
+            else()
+                set_image_base(${MODULE} 0x80800000)
```
Extravert-ir, line 635: The `if(GDB AND NOT CLANG)` case used to set `0x00800000`. Is that intended?

sdk/cmake/init-section.lds
```diff
@@ -8,5 +8,4 @@ SECTIONS
     __init_end__ = . ;
   }
 }
-/*INSERT AFTER .reloc;*//*This is what we actually want to do, but do not dare due to LD bug ROSBE-154*/
-INSERT AFTER .rsrc; /*This is what we do instead to at least fix CORE-14683*/
+INSERT BEFORE .reloc; /* .reloc is always at the end */
```
Extravert-ir, line 11: In our MSVC binaries and in Windows drivers, I see the following order: `INIT`, `.rsrc`, `.reloc`. Unless there are other limitations, I suggest to unify the behavior

sdk/tools/pefixup.c
```diff
@@ -121,26 +127,99 @@ static int add_loadconfig(unsigned char *buffer, PIMAGE_NT_HEADERS nt_header)
...
+        if ((strncasecmp((char*)Section->Name, "INIT", 4) == 0)
+            /* Bugcheck code needs this, so don't discard it for kernels */
+            || ((strncasecmp((char*)Section->Name, ".rsrc", 4) == 0) && (mode != MODE_KERNEL)))
+        {
+            Section->Characteristics |= IMAGE_SCN_MEM_DISCARDABLE;
+            continue;
+        }
+
+        /* Known sections which can be paged */
+        if ((strncasecmp((char*)Section->Name, "PAGE", 4) == 0)
```
Extravert-ir, line 147: `.edata`, `INIT`, `.rsrc` should also be here (even for the kernel)

sdk/cmake/gcc.cmake
```diff
@@ -280,13 +280,18 @@ function(set_image_base MODULE IMAGE_BASE)
...
         if(${TYPE} STREQUAL "wdmdriver")
-            add_target_link_flags(${MODULE} "-Wl,--wdmdriver")
+            target_link_options(${MODULE} PRIVATE "-Wl,--wdmdriver")
         endif()
-        #Disabled due to LD bug: ROSBE-154
-        #add_linker_script(${MODULE} ${REACTOS_SOURCE_DIR}/sdk/cmake/init-section.lds)
+        # Place INIT section at the tail of the module
+        add_linker_script(${MODULE} ${REACTOS_SOURCE_DIR}/sdk/cmake/init-section.lds)
+        # Fixup section characteristiscs
+        add_custom_command(TARGET ${MODULE} POST_BUILD
```
Extravert-ir, line 291: I think it makes sense to leave a comment here about which characteristics are being added, so people don't need to look inside pefixup

sdk/cmake/gcc.cmake
```diff
@@ -280,13 +280,18 @@ function(set_image_base MODULE IMAGE_BASE)
...
+        target_link_options(${MODULE} PRIVATE -Wl,--exclude-all-symbols,-file-alignment=0x1000,-section-alignment=0x1000)
         if(${TYPE} STREQUAL "wdmdriver")
-            add_target_link_flags(${MODULE} "-Wl,--wdmdriver")
+            target_link_options(${MODULE} PRIVATE "-Wl,--wdmdriver")
         endif()
-        #Disabled due to LD bug: ROSBE-154
-        #add_linker_script(${MODULE} ${REACTOS_SOURCE_DIR}/sdk/cmake/init-section.lds)
+        # Place INIT section at the tail of the module
+        add_linker_script(${MODULE} ${REACTOS_SOURCE_DIR}/sdk/cmake/init-section.lds)
+        # Fixup section characteristiscs
```
Extravert-ir, line 290:
```suggestion
        # Fixup section characteristics
```

Extravert-ir (changes requested):

sdk/cmake/CMakeMacros.cmake
zefklop: Yes. Actually, 0x00400000 works for everyone now. This is a remnant of old RosBE.

sdk/cmake/init-section.lds
zefklop: .rsrc is non-discardable for ntoskrnl.

sdk/tools/pefixup.c
zefklop: INIT and .rsrc are discardable, not sure about .edata.

sdk/cmake/gcc.cmake
zefklop: Sure

sdk/tools/pefixup.c
Extravert-ir: Windows' `ntoskrnl.exe`: `.edata` is pageable
image

sdk/cmake/init-section.lds
Extravert-ir: I just don't like that we would have different layout for MSVC and GCC builds

zefklop: Done

sdk/cmake/gcc.cmake
zefklop: Done

sdk/tools/pefixup.c
zefklop: OK, this is what is in now :

- Remove alignment bits that LD adds and that are not needed
- Remove write-access from .rsrc
- Remove initialized bit for code sections (LD adds it although it's not needed)
- Add the non-paged bit, except for PAGE, .rsrc and .edata

Extravert-ir (approved):

sdk/tools/pefixup.c
Extravert-ir: Why link.exe does not put the alignment though?

zefklop: I'm not sure. My guess would be that it may be important for COFF, but PE doesn't care since it has the SectionAlignment value anyway.

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -1510,9 +1510,8 @@ MiFindInitializationCode(OUT PVOID *StartVa,
             /* Assume failure */
             InitFound = FALSE;
 
-            /* Is this the INIT section or a discardable section? */
-            if ((strncmp((PCCH)Section->Name, "INIT", 5) == 0) ||
-                ((Section->Characteristics & IMAGE_SCN_MEM_DISCARDABLE)))
+            /* Is this a discardable section? */
+            if (Section->Characteristics & IMAGE_SCN_MEM_DISCARDABLE)
```
HBelusca, line 1514: The removal of the check for "INIT" section breaks Windows compatibility (you can check that by testing whether a section named `"INIT"` gets indeed discarded, even if it doesn't have the `IMAGE_SCN_MEM_DISCARDABLE` flag set).

HBelusca: Also, there is extra treatment for other sections with special names, as explained in
https://community.osr.com/discussion/comment/162734/#Comment_162734

sdk/tools/pefixup.c
```diff
@@ -24,6 +24,12 @@
 static const char* g_ApplicationName;
 static const char* g_Target;
 
+#define MODE_LOADCONFIG 0
+#define MODE_KERNELDRIVER 1
+#define MODE_WDMDRIVER 2
+#define MODE_KERNELDLL 3
+#define MODE_KERNEL 4
```
HBelusca, line 31:
```suggestion
typedef enum _MODE_FIXUP
{
    MODE_LOADCONFIG,
    MODE_KERNELDRIVER,
    MODE_WDMDRIVER,
    MODE_KERNELDLL,
    MODE_KERNEL,
} MODE_FIXUP;
```

sdk/tools/pefixup.c
```diff
@@ -121,26 +127,113 @@ static int add_loadconfig(unsigned char *buffer, PIMAGE_NT_HEADERS nt_header)
     return 1;
 }
 
+static int driver_fixup(int mode, unsigned char *buffer, PIMAGE_NT_HEADERS nt_header)
```
HBelusca, line 130:
```suggestion
static int driver_fixup(MODE_FIXUP mode, unsigned char *buffer, PIMAGE_NT_HEADERS nt_header)
```

sdk/tools/pefixup.c
```diff
@@ -121,26 +127,113 @@ static int add_loadconfig(unsigned char *buffer, PIMAGE_NT_HEADERS nt_header)
...
+}
+
 int main(int argc, char **argv)
 {
     FILE* file;
     long len;
     unsigned char *buffer;
     PIMAGE_DOS_HEADER dos_header;
     int result = 1;
+    int mode;
```
HBelusca, line 197:
```suggestion
    MODE_FIXUP mode;
```

sdk/cmake/msvc.cmake
```diff
@@ -249,13 +249,20 @@ function(set_image_base MODULE IMAGE_BASE)
...
+        # Mark INIT section as Executable Read Write Discardable
+        target_link_options(${MODULE} PRIVATE /SECTION:INIT,ERWD)
+
+        if(TYPE STREQUAL kernelmodedriver)
+            target_link_options(${MODULE} PRIVATE /DRIVER)
+        elseif(TYPE STREQUAL wdmdriver)
+            target_link_options(${MODULE} PRIVATE /DRIVER:WDM)
+        elseif (TYPE STREQUAL kernel)
+            # Mark .rsrc section as non-disposable non-pageable, as bugcheck code needs to access it
+            target_link_options(${MODULE} PRIVATE /SECTION:.rsrc,!DP )
```
HBelusca, line 264:
```suggestion
            target_link_options(${MODULE} PRIVATE /SECTION:.rsrc,!DP)
```

sdk/tools/pefixup.c
```diff
@@ -121,26 +127,113 @@ static int add_loadconfig(unsigned char *buffer, PIMAGE_NT_HEADERS nt_header)
...
+void
+print_usage(void)
+{
+    printf("Usage: %s <mode> <filename>\n", g_ApplicationName);
+    printf("Where <mode> is on of the following:\n");
+    printf("  --loadconfig:         Fix the LOAD_CONFIG directory entry\n");
+    printf("  --kernelmodedriver:   Fix code and data sections for driver images\n");
+    printf("  --wdmdriver:          Fix code and data sections for WDM drivers\n");
+    printf("  --kerneldll:          Fix code and data sections for Kernel-Mode DLLs\n");
+    printf("  --kernel:             Fix code and data sections for kernels\n");
```
HBelusca, line 187:
```suggestion
    printf("  --loadconfig:         Fix the LOAD_CONFIG directory entry\n");
    printf("  --kernelmodedriver    Fix code and data sections for driver images\n");
    printf("  --wdmdriver           Fix code and data sections for WDM drivers\n");
    printf("  --kerneldll           Fix code and data sections for Kernel-Mode DLLs\n");
    printf("  --kernel              Fix code and data sections for kernels\n");
```
(Since the options do not require a semicolon following them for working.)

sdk/cmake/gcc.cmake
```diff
@@ -280,13 +280,22 @@ function(set_image_base MODULE IMAGE_BASE)
 endfunction()
 
 function(set_module_type_toolchain MODULE TYPE)
-    if((${TYPE} STREQUAL "kernelmodedriver") OR (${TYPE} STREQUAL "wdmdriver"))
-        add_target_link_flags(${MODULE} "-Wl,--exclude-all-symbols,-file-alignment=0x1000,-section-alignment=0x1000")
+    if (TYPE IN_LIST KERNEL_MODULE_TYPES)
```
HBelusca, line 283:
```suggestion
    if(TYPE IN_LIST KERNEL_MODULE_TYPES)
```

sdk/cmake/CMakeMacros.cmake
```diff
@@ -642,18 +626,29 @@ function(set_module_type MODULE TYPE)
...
     # Now do some stuff which is specific to each type
-    if((${TYPE} STREQUAL kernelmodedriver) OR (${TYPE} STREQUAL wdmdriver) OR (${TYPE} STREQUAL kerneldll))
+    if(TYPE IN_LIST KERNEL_MODULE_TYPES)
         add_dependencies(${MODULE} bugcodes xdk)
         if((${TYPE} STREQUAL kernelmodedriver) OR (${TYPE} STREQUAL wdmdriver))
             set_target_properties(${MODULE} PROPERTIES SUFFIX ".sys")
         endif()
     endif()
 
+    if (TYPE STREQUAL kernel)
```
HBelusca, line 646:
```suggestion
    if(TYPE STREQUAL kernel)
```

sdk/cmake/CMakeMacros.cmake
```diff
@@ -642,18 +626,29 @@ function(set_module_type MODULE TYPE)
         else()
             message(STATUS "${MODULE} has no base address")
         endif()
-    elseif((${TYPE} STREQUAL kernelmodedriver) OR (${TYPE} STREQUAL wdmdriver) OR (${TYPE} STREQUAL kerneldll))
-        set_image_base(${MODULE} 0x00010000)
+    elseif(TYPE IN_LIST KERNEL_MODULE_TYPES)
+        # special case for kernel
+        if (TYPE STREQUAL kernel)
```
HBelusca, line 631:
```suggestion
        if(TYPE STREQUAL kernel)
```

ntoskrnl/CMakeLists.txt
```diff
@@ -32,46 +32,26 @@ add_executable(ntoskrnl
...
 if(STACK_PROTECTOR)
     target_sources(ntoskrnl PRIVATE $<TARGET_OBJECTS:gcc_ssp_nt>)
-elseif(RUNTIME_CHECKS)
-    target_link_libraries(ntoskrnl runtmchk)
 endif()
 
 add_importlibs(ntoskrnl hal kdcom bootvid)
 add_pch(ntoskrnl ${REACTOS_SOURCE_DIR}/ntoskrnl/include/ntoskrnl.h "${PCH_SKIP_SOURCE}")
-add_dependencies(ntoskrnl psdk bugcodes asm)
+add_dependencies(ntoskrnl psdk asm)
```
HBelusca, line 54: So, where does the mandatory bugcodes dependency gets inherited from, now?

ntoskrnl/mm/ARM3/sysldr.c
zefklop: Right. Commit reverted.

ntoskrnl/CMakeLists.txt
zefklop: CMakeMacros.cmake - line 640

PR #3488, opened 2021-02-26
[RTL] Support ENABLE_SYSTEMWIDE_DPH define

SergeGautherie:
## Purpose

To enable it, run "configure.*" with "-DENABLE_SYSTEMWIDE_DPH=1".

JIRA issue: [CORE-12020](https://jira.reactos.org/browse/CORE-12020)

Co-Authored-By: Thomas Faber <thomas.faber@reactos.org>

## Proposed changes

This is the basic part only, which matches
https://reactos.org/wiki/Debugging#Debug_Page_Heap_.28DPH.29


sdk/lib/rtl/CMakeLists.txt
```diff
@@ -5,6 +5,10 @@ add_definitions(
     -D_NTSYSTEM_
     -D_NTDLLBUILD_)
 
+if(ENABLE_SYSTEMWIDE_DPH)
+    add_definitions(-DENABLE_SYSTEMWIDE_DPH)
+endif()
```
zefklop, line 10: Please use `target_compile_definitions` on rtl

zefklop:
Can we enable this globally for debug build ?

I think an option would be in order
https://cmake.org/cmake/help/latest/command/option.html

`option(ENABLE_SYSTEMWIDE_DPH "Globally enable Debug-Paged-Heap" ${DBG})`

Somewhere in sdk/cmake/config.cmake

sdk/lib/rtl/CMakeLists.txt
SergeGautherie: Done.

SergeGautherie:
> Can we enable this globally for debug build ?

I would support setting up some kind of DPH testing.
Nonetheless, your suggestion, as is, is not well suited for that, as this feature makes the build very slow (to install).

Short term: compile a separate build.
Long term: research a way to make this feature dynamically configurable, maybe a registry key...

sdk/lib/rtl/heap.c
```diff
@@ -1289,6 +1293,13 @@ RtlCreateHeap(ULONG Flags,
         Flags &= HEAP_CREATE_VALID_MASK;
     }
 
+#ifdef ENABLE_SYSTEMWIDE_DPH
+    if (!Addr)
+    {
+        Flags |= HEAP_FLAG_PAGE_ALLOCS;
+    }
+#endif
+
```
zefklop, line 1302: This should not be needed. Look at RtlpPageHeapCreate implementation.

zefklop (changes requested):
Please explain why force-setting HEAP_FLAG_PAGE_ALLOCS is necessary

sdk/lib/rtl/heap.c
learn-more: Just setting `FLG_HEAP_PAGE_ALLOCS` on `NtGlobalFlag` should be enough, no need for all these hacks everywhere?

zefklop: NtGlobalFlag is per application, so it would be tedious to enable it globally.

learn-more: There is also a global `NtGlobalFlag`, that all applications inherit

zefklop: Right, it's in Session Manager\GlobalFlag Somehow I missed it.

@SergeGautherie maybe setting it in regtest.cmd would do what you want to achieve.

SergeGautherie:
>  this feature makes ReactOS very slow (to install)

Let's wait for #3517 first...

zefklop:
> > this feature makes ReactOS very slow (to install)
> 
> Let's wait for #3517 first...

This won't change anything, DPH is not using the code paths that #3517 solves

sdk/lib/rtl/heap.c
SergeGautherie: > Just setting `FLG_HEAP_PAGE_ALLOCS` on `NtGlobalFlag` should be enough, no need for all these hacks everywhere?

Thanks for the hint.
I tried to do it on NTOS:EX and NTDLL:LDR, but that did not succeed.
Could you tell me where/how to do that exactly?

--

> regtest.cmd

That might be a good idea for running tests, but it is not the system-wide case (from ISO start).

SergeGautherie (commented):
> Please explain why force-setting HEAP_FLAG_PAGE_ALLOCS is necessary

Before looking into that later, I would simply remind that I just copied what the wiki suggests since 8 years ago.

GeoB99:
Stale PR.

PR #3491, opened 2021-02-27
[COMCTL32] SysListView32  headers disappears

zecarlos1957:
In a SysListView32 object with adjustable width header,
 with the number of elements greater than the size of a page and
 with the index of the first visible element greater than zero
 (by scrolling down one or more elements)
 if in these circumstances we adjust the width of a column, the entire header disappears.

Changes:
          In the file comctl32 / listview.c function LISTVIEW_UpdateVScroll and
	  LISTVIEW_UpdateHScroll (lines 2103, 2106 and 2161)
	  we do not want to change the size of the header, nor the size of the list.
         These functions are about scrolling 

 https://jira.reactos.org/browse/CORE-16988

Teste on:
         Control Panel/Administrative Tools/System Config/Services 
       or Control Panel/User Accounts/Users (after adding 6 new users)
       or Control Panel/User Accounts/Groups

### DIFF
--- a/dll/win32/comctl32/listview.c
+++ b/dll/win32/comctl32/listview.c
@@ -2095,6 +2095,7 @@ static INT LISTVIEW_UpdateHScroll(LISTVIEW_INFO *infoPtr)
     dx -= SetScrollInfo(infoPtr->hwndSelf, SB_HORZ, &horzInfo, TRUE);
     TRACE("horzInfo=%s\n", debugscrollinfo(&horzInfo));
 
+#ifndef __REACTOS__
     /* Update the Header Control */
     if (infoPtr->hwndHeader)
     {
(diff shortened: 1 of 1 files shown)


yagoulas:
is it possible to write a test for this behavior? are there any related winetests?

zecarlos1957:
@yagoulas  There is a program with dozens of tests in modules/winetests/comctl32/listview.c.
I don't know if this fulfills your request?

HBelusca:
Besides, it would be interested to check whether this happens in Wine, somehow, since this code comes from them.

zecarlos1957:
Fixed by https://github.com/reactos/reactos/commit/a5365194f4b7f85ab0b3e086ad2e25a9e9b29694
Thanks to Doug Lyons who picked the fix for us from Wine 6.0

PR #3501, opened 2021-03-03
[GEN_BASEADDRESS] Add traces, Reduce width

SergeGautherie:
Minor improvements.

PR #3502, opened 2021-03-03
[CMAKE] baseaddress*.cmake: Update to 0.4.15-dev-2119-g6170b57

SergeGautherie:
Last update was on 2019-12-27.

NB:
Module name width depends on #3501.


HBelusca:
How did you generate these files?

SergeGautherie:
> How did you generate these files?

I added steps to GitHub builds.

GeoB99:
A lot of things have changed since you created this PR so there's no point on trying to update the base addresses (which themselves are also outdated). Plus it makes sense for a person who has commit push access to update such addresses.

PR #3505, opened 2021-03-05
[SHELL32] Avoid code duplication on FAT and EXT2/BtrFS related FSs in…

GeoB99:
…sertion

This is a slight refactor of `InsertDefaultClusterSizeForFs` code with the goal to avoid having the same code for FAT-related and EXT2/BtrFS filesystems.

### DIFF
--- a/dll/win32/shell32/dialogs/drive.cpp
+++ b/dll/win32/shell32/dialogs/drive.cpp
@@ -41,7 +41,7 @@ GetDefaultClusterSize(LPWSTR szFs, PDWORD pClusterSize, PULARGE_INTEGER TotalNum
     DWORD ClusterSize;
 
     if (!wcsicmp(szFs, L"FAT16") ||
-        !wcsicmp(szFs, L"FAT")) //REACTOS HACK
+        !wcsicmp(szFs, L"FAT")) // REACTOS HACK
     {
         if (TotalNumberOfBytes->QuadPart <= (16 * 1024 * 1024))
             ClusterSize = 2048;
@@ -208,6 +208,8 @@ static VOID
 InsertDefaultClusterSizeForFs(HWND hwndDlg, PFORMAT_DRIVE_CONTEXT pContext)
 {
     WCHAR wszBuf[100] = {0};
+    WCHAR wszDefaultSize[100] = {0};
+    PCWSTR pwszFsSizeLimit;
     WCHAR szDrive[] = L"C:\\";
     INT iSelIndex;
     ULARGE_INTEGER FreeBytesAvailableUser, TotalNumberOfBytes;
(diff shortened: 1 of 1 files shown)


dll/win32/shell32/dialogs/drive.cpp
```diff
@@ -299,7 +270,7 @@ InsertDefaultClusterSizeForFs(HWND hwndDlg, PFORMAT_DRIVE_CONTEXT pContext)
     {
         if (!GetDefaultClusterSize(wszBuf, &ClusterSize, &TotalNumberOfBytes))
         {
-            TRACE("NTFS is not supported on drive larger than 2TB current %lu\n", TotalNumberOfBytes.QuadPart);
+            TRACE("NTFS is not supported on drive larger than 2TB, current size -> %lu\n", TotalNumberOfBytes.QuadPart);
```
HBelusca, line 273:
```suggestion
            TRACE("NTFS is not supported on drive larger than 2TB, current size: %lu\n", TotalNumberOfBytes.QuadPart);
```

dll/win32/shell32/dialogs/drive.cpp
```diff
@@ -328,32 +299,12 @@ InsertDefaultClusterSizeForFs(HWND hwndDlg, PFORMAT_DRIVE_CONTEXT pContext)
...
-        EnableWindow(GetDlgItem(hwndDlg, 28675), TRUE);
-    }
-    else if (!wcsicmp(wszBuf, L"BtrFS"))
+    else if (!wcsicmp(wszBuf, L"EXT2") ||
+             !wcsicmp(wszBuf, L"BtrFS"))
     {
         if (!GetDefaultClusterSize(wszBuf, &ClusterSize, &TotalNumberOfBytes))
         {
-            TRACE("BtrFS is not supported on drive larger than 16E current %lu\n", TotalNumberOfBytes.QuadPart);
+            TRACE("%S is not supported on drive larger than %S, current size -> %lu\n", wszBuf, wszFsLimit, TotalNumberOfBytes.QuadPart);
```
HBelusca, line 307:
```suggestion
            TRACE("%S is not supported on drive larger than %S, current size: %lu\n", wszBuf, wszFsLimit, TotalNumberOfBytes.QuadPart);
```

dll/win32/shell32/dialogs/drive.cpp
```diff
@@ -230,54 +231,24 @@ InsertDefaultClusterSizeForFs(HWND hwndDlg, PFORMAT_DRIVE_CONTEXT pContext)
...
-            lIndex = SendMessageW(hDlgCtrl, CB_ADDSTRING, 0, (LPARAM)wszBuf);
-            if (lIndex != CB_ERR)
-                SendMessageW(hDlgCtrl, CB_SETITEMDATA, lIndex, (LPARAM)ClusterSize);
-            SendMessageW(hDlgCtrl, CB_SETCURSEL, 0, 0);
-        }
-
-        SendMessageW(GetDlgItem(hwndDlg, 28675), BM_SETCHECK, BST_UNCHECKED, 0);
-        EnableWindow(GetDlgItem(hwndDlg, 28675), FALSE);
-    }
+        StringCchCopyW(wszFsLimit, _countof(wszFsLimit), L"4G");
```
HBelusca, line 234: Just have `wszFsLimit` actually be a pointer to string (a `PCWSTR`, and also rename that variable...) so that you can just need to do:
```c
pwszFsLimit = L"4G";
```
etc... and then use that `pwszFsLimit` in parameters of these TRACE calls.

dll/win32/shell32/dialogs/drive.cpp
```diff
@@ -230,54 +231,24 @@ InsertDefaultClusterSizeForFs(HWND hwndDlg, PFORMAT_DRIVE_CONTEXT pContext)
 
     if (!wcsicmp(wszBuf, L"FAT16") ||
         !wcsicmp(wszBuf, L"FAT")) //REACTOS HACK
-    {
```
HBelusca, line 233: Since your if conditions take more than one line, and there are in if... else if ... pattern, please keep the braces for each case.

dll/win32/shell32/dialogs/drive.cpp
```diff
@@ -328,32 +299,12 @@ InsertDefaultClusterSizeForFs(HWND hwndDlg, PFORMAT_DRIVE_CONTEXT pContext)
...
+    else if (!wcsicmp(wszBuf, L"EXT2") ||
+             !wcsicmp(wszBuf, L"BtrFS"))
     {
         if (!GetDefaultClusterSize(wszBuf, &ClusterSize, &TotalNumberOfBytes))
         {
-            TRACE("BtrFS is not supported on drive larger than 16E current %lu\n", TotalNumberOfBytes.QuadPart);
+            TRACE("%S is not supported on drive larger than %S, current size -> %lu\n", wszBuf, wszFsLimit, TotalNumberOfBytes.QuadPart);
             SendMessageW(hDlgCtrl, CB_DELETESTRING, iSelIndex, 0);
             return;
         }
```
HBelusca, line 310: Isn't the following LoadString etc... block the same as in the FAT... case?
(Sorry I cannot place comments at arbitrary places in this github webinterface).

dll/win32/shell32/dialogs/drive.cpp
```diff
@@ -229,86 +231,76 @@ InsertDefaultClusterSizeForFs(HWND hwndDlg, PFORMAT_DRIVE_CONTEXT pContext)
...
+    {
+        pwszFsSizeLimit = L"32T";
+    }
+    else
+    {
+        pwszFsSizeLimit = L"16E";
+    }
+
+    if (!wcsicmp(wszBuf, L"FAT16") ||
+        !wcsicmp(wszBuf, L"FAT") || //REACTOS HACK
```
HBelusca, line 256:
```suggestion
        !wcsicmp(wszBuf, L"FAT") || // REACTOS HACK
```

dll/win32/shell32/dialogs/drive.cpp
```diff
@@ -229,86 +231,76 @@ InsertDefaultClusterSizeForFs(HWND hwndDlg, PFORMAT_DRIVE_CONTEXT pContext)
...
-
-        SendMessageW(GetDlgItem(hwndDlg, 28675), BM_SETCHECK, BST_UNCHECKED, 0);
-        EnableWindow(GetDlgItem(hwndDlg, 28675), FALSE);
     }
     else if (!wcsicmp(wszBuf, L"NTFS"))
     {
         if (!GetDefaultClusterSize(wszBuf, &ClusterSize, &TotalNumberOfBytes))
         {
-            TRACE("NTFS is not supported on drive larger than 2TB current %lu\n", TotalNumberOfBytes.QuadPart);
+            TRACE("NTFS is not supported on drive larger than 2TB, current size: %lu\n", TotalNumberOfBytes.QuadPart);
```
HBelusca, line 294: By the way I have 8TB drives with NTFS and these work great.

dll/win32/shell32/dialogs/drive.cpp
```diff
@@ -229,86 +231,76 @@ InsertDefaultClusterSizeForFs(HWND hwndDlg, PFORMAT_DRIVE_CONTEXT pContext)
...
             SendMessageW(hDlgCtrl, CB_SETCURSEL, 0, 0);
         }
 
         SendMessageW(GetDlgItem(hwndDlg, 28675), BM_SETCHECK, BST_UNCHECKED, 0);
-        EnableWindow(GetDlgItem(hwndDlg, 28675), FALSE);
-    }
-    else if (!wcsicmp(wszBuf, L"FATX"))
-    {
-        if (!GetDefaultClusterSize(wszBuf, &ClusterSize, &TotalNumberOfBytes))
+        if (!wcsicmp(wszBuf, L"EXT2") ||
```
HBelusca, line 280: Since these dialog item resources still don't have human-readable IDs (but just numbers....) it would be nice if a comment explains what gets enabled/disabled.

dll/win32/shell32/dialogs/drive.cpp
```diff
@@ -229,86 +231,76 @@ InsertDefaultClusterSizeForFs(HWND hwndDlg, PFORMAT_DRIVE_CONTEXT pContext)
...
         }
-
-        SendMessageW(GetDlgItem(hwndDlg, 28675), BM_SETCHECK, BST_UNCHECKED, 0);
-        EnableWindow(GetDlgItem(hwndDlg, 28675), FALSE);
     }
     else if (!wcsicmp(wszBuf, L"NTFS"))
     {
         if (!GetDefaultClusterSize(wszBuf, &ClusterSize, &TotalNumberOfBytes))
         {
-            TRACE("NTFS is not supported on drive larger than 2TB current %lu\n", TotalNumberOfBytes.QuadPart);
```
HBelusca, line 302: All that part (until the `ClusterSize = 512;` much below) seems to also be in common with the other file systems, could you confirm? If so it may be interesting to make it indeed common with the code above, and then have the specific `if (!wcsicmp(wszBuf, L"NTFS"))` just for the ClusterSize thing.

HBelusca (approved):

PR #3506, opened 2021-03-05
[MSGINA][SHELL32] Polish fancy log off dialog box fix + other fixes

arnavbhatt288:
## Purpose

- Fix the Polish fancy log off dialog box.
- Fix the way the "default" button was determined for the enter key.

## Proposed changes

- Made the fancy log off dialog box longer along with other adjustments for Polish.
- Now the subclass proc will determine which button should be "default". (based on whether the owner draw buttons lost their keyboard focus or not)

## Result
Before:
Screenshot from 2021-03-05 22-11-12

After:
Screenshot from 2021-03-05 22-12-42


HBelusca:
Is Polish so far the only dialog affected by the size?

arnavbhatt288:
> Is Polish so far the only dialog affected by the size?

@HBelusca Yes, rest are fine.

GeoB99 (approved):
You need to rebase and reword the title of your commit, otherwise LGTM.

PR #3509, opened 2021-03-06
[FLOPPY][FLOPPY_NEW] Switch to MS driver

SergeGautherie:
JIRA issue: [CORE-17411](https://jira.reactos.org/browse/CORE-17411)

## TODO

- [ ] Test it.

### DIFF
--- a/drivers/storage/CMakeLists.txt
+++ b/drivers/storage/CMakeLists.txt
@@ -1,5 +1,6 @@
 add_subdirectory(class)
-add_subdirectory(floppy)
+# add_subdirectory(floppy)
+add_subdirectory(floppy_new)
 add_subdirectory(ide)
 add_subdirectory(mountmgr)
 add_subdirectory(partmgr)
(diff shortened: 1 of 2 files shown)


ColinFinck:
As Eric correctly pointed out in CORE-17411, the imported "floppy_new" is not a replacement for our floppy.sys at all.
"floppy_new" supports floppy drives communicating via SCSI-style commands (connected to SCSI or USB buses), but not the traditional floppy drives communicating via I/O ports.
In fact, "floppy_new" should have never been imported under this name.

I'm not saying our floppy.sys is perfect. It was originally developed as a substitute for NT4's floppy.sys, but MS refactored floppy support into fdc.sys and flpydisk.sys starting with NT5.
Eric has already started work on fdc.sys, though our flpydisk.sys is still non-existent.

Someone may also reimport the latest Windows driver sample under its proper name "sfloppy", possibly remove some ReactOS hacks that are no longer needed, and verify that this brings support for USB floppy drives.

SergeGautherie:
> As Eric correctly pointed out in CORE-17411

Yes, that was after I created this PR.

> the imported "floppy_new" is not a replacement for our floppy.sys at all.

Thanks for the detailed explanation!

PR #3511, opened 2021-03-07
[SHELL32] Do not format the partition if it's a system drive

GeoB99:
Implement a sanity check helper which determines if the partition is a system drive or not based on the `%SystemDrive%` environment variable, preventing the user from nuking accidentally the partition with ReactOS system files installed. 😝

**GIF for demonstration (sorry for shitty quality)**

### DIFF
--- a/dll/win32/shell32/dialogs/drive.cpp
+++ b/dll/win32/shell32/dialogs/drive.cpp
@@ -35,6 +35,35 @@ typedef struct
 EXTERN_C HPSXA WINAPI SHCreatePropSheetExtArrayEx(HKEY hKey, LPCWSTR pszSubKey, UINT max_iface, IDataObject *pDataObj);
 HPROPSHEETPAGE SH_CreatePropertySheetPage(LPCSTR resname, DLGPROC dlgproc, LPARAM lParam, LPWSTR szTitle);
 
+/*
+ * TODO: In Windows the Shell doesn't know by itself if a drive is
+ * a system one or not but rather a packet message is being sent by
+ * FMIFS library code and further translated into specific packet
+ * status codes in the Shell, the packet being _FMIFS_PACKET_TYPE.
+ *
+ * With that being said, most of this code as well as FMIFS library code
+ * have to be refactored in order to comply with the way Windows works.
+ *
+ * See the enum definition for more details:
+ * https://github.com/microsoft/winfile/blob/master/src/fmifs.h#L23
+ */
+static BOOL
+IsSystemDrive(PFORMAT_DRIVE_CONTEXT pContext)
+{
+    WCHAR wszDriveLetter[6], wszSystemDrv[6];
+
+    wszDriveLetter[0] = pContext->Drive + L'A';
+    StringCchCatW(wszDriveLetter, _countof(wszDriveLetter), L":");
+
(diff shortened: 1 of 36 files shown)


HBelusca:
Does Windows (XP/Server 2003) show a similar warning?

GeoB99:
It doesn't, at least not in a similar fashion like the way ReactOS does with this implementation. In W2k3 it does it in two different ways.

This is without the "Quick Format" option enabled:
Capture

And this one is with "Quick Format" enabled:
Capture2

This suggests to me that in Windows the Shell definitely acknowledges the given partition for format as a system one but it doesn't explicitly warn the user that it cannot format it because it's a system drive (the former case looks so weird to me but whatever). I seriously despise the way's done in Windows.

HBelusca:
To me it doesn't look the shell "knows" it's the system partition; instead it goes on but the underlying format routines cannot lock the system partition for formatting but fail, and the shell just translates the error code into these human-readable errors.

GeoB99:
According to WinDBG, the message boxes are created by `shell32!FormatCallback`. With further analysis it seems a respective argument passed to that call is of `FMIFS_PACKET_TYPE` data type, presumably an enum (luckily such enumeration definition is defined in [WinFile -- the open sourced file manager by Microsoft](https://github.com/microsoft/winfile/blob/master/src/fmifs.h#L23)). The IFS packet message that's passed to that function is brought by `FormatEx`.

As none of that is done in any way in ReactOS, the responsible code that manages the format dialog box internal format procedure as well as the FMIFS code library in its entirety have to be rewritten eventually with the respect of FMIFS_PACKET_TYPE and whatever is done in Windows. This also means getting totally rid of WINE's code in Shell32 regarding formatting stuff (which is not a trivial task, plus I'm not that interested about it).

With all that being said, let the Shell know if a drive is system or not for now until someone is adventurous enough to work on that. 😛

dll/win32/shell32/dialogs/drive.cpp
```diff
@@ -737,7 +766,16 @@ SHFormatDrive(HWND hwnd, UINT drive, UINT fmtID, UINT options)
...
-    result = DialogBoxParamW(shell32_hInstance, MAKEINTRESOURCEW(IDD_FORMAT_DRIVE), hwnd, FormatDriveDlg, (LPARAM)&Context);
+    if (!IsSystemDrive(&Context))
+    {
+        result = DialogBoxParamW(shell32_hInstance, MAKEINTRESOURCEW(IDD_FORMAT_DRIVE), hwnd, FormatDriveDlg, (LPARAM)&Context);
+    }
+    else
+    {
+        result = SHFMT_ERROR;
+        ShellMessageBoxW(shell32_hInstance, hwnd, MAKEINTRESOURCEW(IDS_NO_FORMAT), MAKEINTRESOURCEW(IDS_NO_FORMAT_TITLE), MB_OK | MB_ICONWARNING);
+        TRACE("SHFormatDrive(): The provided drive for format is a system partition! Aborting...\n");
```
HBelusca, line 777:
```suggestion
        TRACE("SHFormatDrive(): The provided drive for format is a system volume! Aborting...\n");
```

HBelusca (commented):
minor comments, but otherwise OK.

PR #3512, opened 2021-03-08
[SFLOPPY] Move and rename floppy_new to class/sfloppy

SergeGautherie:
Supersedes #3509.

JIRA issue: [CORE-17411](https://jira.reactos.org/browse/CORE-17411)


boot/bootdata/hivesys.inf
```diff
@@ -1394,7 +1394,7 @@ HKLM,"SYSTEM\CurrentControlSet\Control\ServiceGroupOrder","List",0x00010000, \
  "System Bus Extender", \
  "SCSI Miniport", \
  "Port", \
- "Primary Disk", \
+ "Primary disk", \
```
ColinFinck, line 1397: This is just wrong, and unrelated.
Why are you mixing this into a PR about the sfloppy driver??

ColinFinck (changes requested):

boot/bootdata/hivesys.inf
SergeGautherie: Indeed, sorry, reverted that one.
Because floppy_new has that mismatch...

boot/bootdata/hivesys.inf
```diff
@@ -1773,7 +1773,7 @@ HKLM,"SYSTEM\CurrentControlSet\Services\usbuhci","Type",0x00010001,0x00000001
 
 ; USB storage driver
 HKLM,"SYSTEM\CurrentControlSet\Services\usbstor","ErrorControl",0x00010001,0x00000001
-HKLM,"SYSTEM\CurrentControlSet\Services\usbstor","Group",0x00000000,"Primary Disk"
+HKLM,"SYSTEM\CurrentControlSet\Services\usbstor","Group",0x00000000,"Primary disk"
```
HBelusca, line 1776: Please revert.

drivers/storage/floppy/floppy/floppy_reg.inf
```diff
@@ -1,7 +1,7 @@
 ; Floppy driver
 [AddReg]
 HKLM,"SYSTEM\CurrentControlSet\Services\Floppy","ErrorControl",0x00010001,0x00000000
-HKLM,"SYSTEM\CurrentControlSet\Services\Floppy","Group",0x00000000,"Primary Disk"
+HKLM,"SYSTEM\CurrentControlSet\Services\Floppy","Group",0x00000000,"Primary disk"
```
HBelusca, line 4: Please revert.

drivers/storage/class/sfloppy/sfloppy_reg.inf
```diff
@@ -0,0 +1,6 @@
+; Microsoft Super Floppy (sfloppy) Storage Class Driver
+[AddReg]
+HKLM,"SYSTEM\CurrentControlSet\Services\Sfloppy","ErrorControl",0x00010001,0x00000000
+HKLM,"SYSTEM\CurrentControlSet\Services\Sfloppy","Group",0x00000000,"Primary disk"
```
HBelusca, line 4:
```suggestion
HKLM,"SYSTEM\CurrentControlSet\Services\Sfloppy","Group",0x00000000,"Primary Disk"
```

HBelusca (changes requested):

Extravert-ir:
sfloppy is a class (classpnp) driver, so it should reside in a `drivers/class` folder

boot/bootdata/hivesys.inf
Extravert-ir: The right writing is `Primary Disk`. Please fill a separate PR with that

SergeGautherie: Moved to #3516...

SergeGautherie:
> `drivers/class` folder

Moved.

SergeGautherie:
Added 'Tag' key.

HBelusca:
The definitive correct folder is `drivers/storage/class`. Indeed:
1. It's a class driver for the storage stack, not for whatever generic driver.
2. All the other storage class drivers we have (some of which, as you know, coming from the MS samples) are already in the existing `drivers/storage/class`.

Extravert-ir:
It is `drivers/storage/class` ofc, I mistyped
BTW, as we're here, it makes sense to write an inf file for the driver (in order for it to be complete). It should take `GenSFloppy`, `USBSTOR\GenSFloppy` and some LS-120/SuperDrive IDs

HBelusca:
For completeness, at least on Windows 10 the sfloppy INF data stuff that @Extravert-ir mentions, is present inside a `flpydisk.inf` file.
We happen to have this file already:
https://git.reactos.org/?p=reactos.git;a=blob;f=media/inf/flpydisk.inf;hb=0c34c067c461d00c290e76e8066b226e47eeec43

Extravert-ir:
I want it to be a separate file, otherwise we will have to always ship both flpydisk and sfloppy

SergeGautherie:
> `drivers/storage/class` ofc

Hum, isn't that what I had submitted initially? Or do you mean you do not want the `/sfloppy` sub-directory?

> write an inf file for the driver

I would prefer to only move and update existing files,
and leave further dev and possibly enabling the driver to a follow-up PR.

ColinFinck:
> I want it to be a separate file, otherwise we will have to always ship both flpydisk and sfloppy together

Do we know what "flpydisk" does exactly, and why NT4's "floppy" driver was split into "fdc" and "flpydisk"?

I haven't checked the details yet, but I could imagine that "flpydisk" provides common code for the different interfaces that support floppies, like "fdc" or "sfloppy".
This is pure speculation though. Some enlightenment would be very helpful here.

SergeGautherie:
> I could imagine that "flpydisk" provides common code

For what it's worth, that looks likely:

http://dll-repair.com/flpydisk-sys/
'flpydisk.sys, also written as Floppy Driver, is used to support fdc, fdc.sys and sfloppy, sfloppy.sys [...]'

http://www.delphigroups.info/2/7c/502169.html
'We did download the Win2k DDK and found no floppy driver but 1. fdc.sys + 2. flpydisk.sys instead. [...]'

PR #3514, opened 2021-03-08
Define _DEBUG for better ATL debugging

katahiromz:
## Purpose

Currently `ATLASSERT` macro won't work in GCC compilers, due to lack of `_DEBUG` macro.
Microsoft-style debugging needs `_DEBUG` macro definition.

## Proposed changes

- Define `_DEBUG` macro as follows in some precompiled header that uses ATL module.

```c
#if !defined(_DEBUG) && !defined(NDEBUG)
    #define _DEBUG
#endif
```
- Fix `sdk/lib/atl/atlcom.h`.
- Fix `sdk/lib/atl/statreg.h`.

## TODO

- [x] Can build
- [x] Can test successful
- [ ] Get consensus of team

### DIFF
--- a/sdk/lib/atl/statreg.h
+++ b/sdk/lib/atl/statreg.h
@@ -68,19 +68,19 @@ public:
 
     HRESULT STDMETHODCALLTYPE QueryInterface(const IID & /* riid */, void ** /* ppvObject */ )
     {
-        ATLASSERT(_T("statically linked in CRegObject is not a com object. Do not callthis function"));
+        ATLASSERT(FALSE && TEXT("statically linked in CRegObject is not a com object. Do not call this function"));
         return E_NOTIMPL;
     }
 
     ULONG STDMETHODCALLTYPE AddRef()
     {
-        ATLASSERT(_T("statically linked in CRegObject is not a com object. Do not callthis function"));
+        ATLASSERT(FALSE && TEXT("statically linked in CRegObject is not a com object. Do not callthis function"));
         return 1;
     }
 
     ULONG STDMETHODCALLTYPE Release()
     {
-        ATLASSERT(_T("statically linked in CRegObject is not a com object. Do not callthis function"));
+        ATLASSERT(FALSE && TEXT("statically linked in CRegObject is not a com object. Do not callthis function"));
         return 0;
     }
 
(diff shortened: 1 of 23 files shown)


SergeGautherie:
Could this define live in some ATL header(s), instead of being duplicated in every module than uses it?

katahiromz:
Nope. Microsoft manual says that we have to define `_DEBUG` explicitly, so we follow their way.

sdk/lib/atl/statreg.h
```diff
@@ -68,19 +68,19 @@ class CRegObject : public IRegistrarBase
 
     HRESULT STDMETHODCALLTYPE QueryInterface(const IID & /* riid */, void ** /* ppvObject */ )
     {
-        ATLASSERT(_T("statically linked in CRegObject is not a com object. Do not callthis function"));
+        ATLASSERT(FALSE && TEXT("statically linked in CRegObject is not a com object. Do not call this function"));
```
SergeGautherie, line 71: `call this`: fix the 2 others too.

SergeGautherie:
> `#if !defined(_DEBUG) && !defined(NDEBUG)`

Why not check GCC/!MSVC, as you reported this is compiler dependent?
Especially, isn't your proposal wrong on MSVC?

> Microsoft manual says that we have to define `_DEBUG` explicitly, so we follow their way.

Which manual? What does it say exactly?

https://docs.microsoft.com/en-us/cpp/c-runtime-library/debug?view=msvc-160
`The compiler defines _DEBUG when you specify the /MTd or /MDd option. These options specify debug versions of the C run-time library.`
It looks like this define is meant to be checked, not to be set by code...
As in, maybe define it at cmake level, if that applies?

katahiromz:
There are many NDEBUG-defined projects in ReactOS. So we can't define _DEBUG globally at CMake settings.

Thus, we have to define _DEBUG macro in specific projects individually. If so, we can either:

1. Define it at CMakeLists.txt of individual projects. Or:
2. Define it at precompiled (or top-level) header of individual projects.

katahiromz:
Defining _DEBUG in CMakeLists.txt is better.
But I can't write the mechanism of CMake.

PR #3516, opened 2021-03-09
[FLOPPY][FLOPPY_NEW][HIVESYS] Inf: Use 'd' for 'Primary disk'

SergeGautherie:
as 'Group' value.

NB:
Only hivesys '...\ServiceGroupOrder' list has a 'D'.

## Purpose

Match WXP/WS03.

### DIFF
--- a/boot/bootdata/hivesys.inf
+++ b/boot/bootdata/hivesys.inf
@@ -1773,7 +1773,7 @@ HKLM,"SYSTEM\CurrentControlSet\Services\usbuhci","Type",0x00010001,0x00000001
 
 ; USB storage driver
 HKLM,"SYSTEM\CurrentControlSet\Services\usbstor","ErrorControl",0x00010001,0x00000001
-HKLM,"SYSTEM\CurrentControlSet\Services\usbstor","Group",0x00000000,"Primary Disk"
+HKLM,"SYSTEM\CurrentControlSet\Services\usbstor","Group",0x00000000,"Primary disk"
 HKLM,"SYSTEM\CurrentControlSet\Services\usbstor","ImagePath",0x00020000,"system32\drivers\usbstor.sys"
 HKLM,"SYSTEM\CurrentControlSet\Services\usbstor","Start",0x00010001,0x00000000
 HKLM,"SYSTEM\CurrentControlSet\Services\usbstor","Type",0x00010001,0x00000001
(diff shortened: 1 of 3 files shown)


Extravert-ir (changes requested):
The correct writing is with the capital 'D'. Need to change other places

SergeGautherie:
> The correct writing is with the capital 'D'. Need to change other places

Again. Yes, this list (only) does have a 'D',
nonetheless, all the other occurrences on my WXP/WS03 MS VMs have a 'd'.
Is that not the case for you?

WXP example:
```
Windows Registry Editor Version 5.00

[HKEY_LOCAL_MACHINE\SYSTEM\ControlSet001\Services\Sfloppy]
"DependOnGroup"=hex(7):53,00,43,00,53,00,49,00,20,00,6d,00,69,00,6e,00,69,00,\
  70,00,6f,00,72,00,74,00,00,00,00,00
"ErrorControl"=dword:00000000
"Group"="Primary disk"
"Start"=dword:00000001
"Tag"=dword:00000004
"Type"=dword:00000001

[HKEY_LOCAL_MACHINE\SYSTEM\ControlSet001\Services\Sfloppy\Enum]
"Count"=dword:00000000
"NextInstance"=dword:00000000
"INITSTARTFAILED"=dword:00000001
```

HBelusca (changes requested):
Not sure it actually makes sense to have a different casing for the groups here in the drivers, compared to the correct one in the service groups list. Our casing was made consistent, and you break it.

SergeGautherie:
> made consistent

Well, it does not look like fully consistent to me:
https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=Primary+disk&sr=1

Then, consistent with which OS, ReactOS or Windows?

Extravert-ir:
The `SYSTEM\CurrentControlSet\Control\ServiceGroupOrder` key is the source for these strings because, that is what a driver's `Group` value is compared with. `Primary disk` seem to be just a mistake, because all other groups have all first letters capitalized.
Having this string "same as in Windows" won't help anything, comparison is case-insensitive anyway

SergeGautherie:
> "same as in Windows" won't help anything

Ok...

> Need to change other places

Then, shall I update the 2 key names instead, or just close this PR?

Extravert-ir:
Superseded by d220ea3e3c4b2a61fb3125c36c252965b44b1267

PR #3519, opened 2021-03-10
[CMAKE] Define _DEBUG for better ATL debugging (Retry)

katahiromz:
## Purpose

Currently `ATLASSERT` macro won't work in some environments, due to lack of `_DEBUG` macro.
Microsoft-style debugging needs `_DEBUG` macro definition. Thanks to @zefklop.

## Proposed changes

- `if(DBG) target_compile_definitions(atl_classes INTERFACE _DEBUG) endif(DBG)` in `sdk/lib/atl`.
- Fix `sdk/lib/atl/statreg.h` (misuse of `ATLASSERT`).
- Fix `sdk/lib/atl/atlcom.h` (non-NULL check for non-NULL).
- Fix `base/applications/rapps/include/defines.h` (redefinition of `_DEBUG`).

## TODO

- [ ] Fix build.
- [x] @zefklop confirmed.

### DIFF
--- a/base/applications/rapps/include/defines.h
+++ b/base/applications/rapps/include/defines.h
@@ -4,7 +4,6 @@
 #define _INC_WINDOWS
 #define COM_NO_WINDOWS_H
 #define COBJMACROS
-#define _DEBUG
 #include <tchar.h>
 #include <stdarg.h>
 
--- a/sdk/lib/atl/CMakeLists.txt
+++ b/sdk/lib/atl/CMakeLists.txt
@@ -1,5 +1,6 @@
 
 add_library(atl_classes INTERFACE)
+target_compile_definitions(atl_classes INTERFACE _DEBUG)
 
 target_include_directories(atl_classes INTERFACE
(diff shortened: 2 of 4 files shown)


SergeGautherie:
Supersedes #3514.
This looks much better.

I still wonder if it could be "inherited" from ATL (or CRT?), instead of added to each target module?

katahiromz:
Our repository has non-debuggable modules (that might define NDEBUG internally). That cannot define `_DEBUG` there; otherwise debugging status will be compromised.

zefklop (changes requested):
Please instead use `target_compile_definitions(atl_classes INTERFACE _DEBUG)` wrapped into a `if (DBG)` logical block

We tend not to use generator expressions unless it's really necessary.

katahiromz:
I see. `INTERFACE` is useful.

zefklop (approved):

SergeGautherie:
> `INTERFACE`

Perfect :-)

PR #3527, opened 2021-03-11
[ATL][SHELL32][EXPLORER] Disable ATLASSERT

katahiromz:
## Purpose

We are not ready for enabling `ATLASSERT`. Enabling ATL assertions takes time to realize.

## Proposed changes

- Disable `ATLASSERT` by undefining `_DEBUG`.
- Revert currently non-fixable codes.

### DIFF
--- a/sdk/lib/atl/CMakeLists.txt
+++ b/sdk/lib/atl/CMakeLists.txt
@@ -1,7 +1,7 @@
 
 add_library(atl_classes INTERFACE)
 if(DBG)
-    #target_compile_definitions(atl_classes INTERFACE _DEBUG)
+    #target_compile_definitions(atl_classes INTERFACE _DEBUG) # HACK & FIXME: CORE-17505
 endif()
 
 target_include_directories(atl_classes INTERFACE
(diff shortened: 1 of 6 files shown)


sdk/lib/atl/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 
 add_library(atl_classes INTERFACE)
 if(DBG)
-    target_compile_definitions(atl_classes INTERFACE _DEBUG)
+    #target_compile_definitions(atl_classes INTERFACE _DEBUG) # HACK & FIXME: CORE-17505
```
SergeGautherie, line 4: Does this need to be disabled globally, or could it be done on the affected files/modules only?

katahiromz: If you have any other idea, please create your Pull Request.

katahiromz: Pinpoint fix is preferred but we can't.

katahiromz:
That's for user's convenience. We shouldn't disappoint iso downloader.

PR #3529, opened 2021-03-11
[ATL] CWindowImplBaseT::WindowProc(): Narrow 2 ATLASSERT() down

SergeGautherie:
Addendum to e076e05.


learn-more:
What problem is this trying to solve?

SergeGautherie:
> What problem is this trying to solve?

Triggering 3 different assertions if `pThis == NULL`, for example.

PR #3534, opened 2021-03-14
[NTOS:SE] Create two system anonymous logon tokens on boot

GeoB99:
On Windows three kind of system tokens are being created -- 1) system process token (which we already do that in ReactOS, 2) anonymous logon token not including the "Everyone SID group" and 3) anonymous logon token that includes such SID group.

The latter two tokens are created during phase 0 of the initialisation process of the SRM kernel component and they're needed for eventual impersonation of such tokens and such. With that being said, this is a next step closer to the implementation of `NtImpersonateAnonymousToken`.
## Proposed Changes
* Implement `SepCreateSystemAnonymousLogonToken` and `SepCreateSystemAnonymousLogonTokenNoEveryone` functions (they're equivalent to `SeMakeAnonymousLogonToken` and `SeMakeAnonymousLogonTokenNoEveryone` on Windows).
* Set up a security descriptor and an access control list for the anonymous logon.
* [Additional] `SeMakeSystemToken` is set in the INIT code segment section on Windows so do that for `SepCreateSystemProcessToken` as well. Also include the documentation header.

### DIFF
--- a/ntoskrnl/se/semgr.c
+++ b/ntoskrnl/se/semgr.c
@@ -15,6 +15,8 @@
 
 /* GLOBALS ********************************************************************/
 
+PTOKEN SeAnonymousLogonToken = NULL;
+PTOKEN SeAnonymousLogonTokenNoEveryone = NULL;
 PSE_EXPORTS SeExports = NULL;
 SE_EXPORTS SepExports;
 ULONG SidInTokenCalls = 0;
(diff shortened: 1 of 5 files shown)


HBelusca:
It would be nice if these were actually used in ReactOS. Because at the moment you introduce this code that creates them, great, but that's all.

GeoB99:
And here's where I said `With that being said, this is a next step closer to the implementation of NtImpersonateAnonymousToken.` which is my primary aim to do that in the next PR where those tokens will be used. I don't like cluttering the PR with massive gazillion of changes at once.

ntoskrnl/se/semgr.c
```diff
@@ -122,6 +124,14 @@ SepInitializationPhase0(VOID)
     ObInitializeFastReference(&PsGetCurrentProcess()->Token, NULL);
     ObInitializeFastReference(&PsGetCurrentProcess()->Token,
                               SepCreateSystemProcessToken());
+
+    /* Initialise the anonymous logon tokens */
+    SeAnonymousLogonToken = SepCreateSystemAnonymousLogonToken();
+    if (!SeAnonymousLogonToken) return FALSE;
```
Extravert-ir, line 130:
```suggestion
    if (!SeAnonymousLogonToken)
        return FALSE;
```
We don't allow that with our current coding style (you are touching a damn old code!)

Extravert-ir (changes requested):
Not that much to review actually.
I just say that non-exported functions should not have a calling convention set, and you seem to use some different line limit (less than 100 characters)

ntoskrnl/se/semgr.c
GeoB99: I wrote this chunk of code that way for the sake of preserving the existing code style of the function. But fine.

PR #3536, opened 2021-03-15
[SHELL32] Hide edit control on 'Copy to' and 'Move to'

katahiromz:
## Purpose

'Copy to' and'Move to' are features of Explorer for beginners. Displaying the textbox is confusing for beginners.
This PR hides the textbox in actions of 'Copy To' and 'Move To' features.

JIRA issue: N/A

## Proposed changes

- Use `BIF_NEWDIALOGSTYLE` style rather than `BIF_USENEWUI`.

## Comparison

WinXP:
WinXP
Win2k3:
Win2k3
BEFORE:
before
AFTER:
after


amber8706:
What for? If this is not present in Windows , then it is bad for Windows. I like the ability to copy the path to a folder and just use it to move or copy it.

amber8706:
Anyway, Windows 7+ has this field: 
image

This looks like an improvement on the dialog box (compared to XP).

PR #3539, opened 2021-03-17
Limit the use of FPU in win32k

zefklop:
## Purpose
Like it should be.
Far from being complete, but that's a start.


win32ss/gdi/eng/i386/floatobj.c
```diff
@@ -0,0 +1,22 @@
+/*
+ * PROJECT:         ReactOS win32 kernel mode subsystem
+ * LICENSE:         GPL - See COPYING in the top level directory
+ * FILE:            win32ss/gdi/eng/i386/floatobj.c
+ * PURPOSE:         FLOATOBJ API
+ * PROGRAMMERS:     Jérôme Gardou
+ */
```
GeoB99, line 7: Small nit-pick -- we no longer use such kind of header like this. See [Coding Style](https://reactos.org/wiki/Coding_Style#File_Structure).

zefklop:
And some tests are even fixed !

PR #3548, opened 2021-03-23
[HHPCOMP] Add ARM (32-bit) types

konradybcio:
ARM32 seems to use the same type defines as i386 and friends.
Let's allow the thing to be compiled again!

Signed-off-by: Konrad Dybcio <konrad.dybcio@somainline.org>

## Purpose

One step closer to getting ARM back on the track

JIRA issue: [CORE-17517](https://jira.reactos.org/browse/CORE-17517) seems the closest

## Proposed changes

Add type definitions for ARM

### DIFF
--- a/sdk/tools/hhpcomp/chmc/chm.h
+++ b/sdk/tools/hhpcomp/chmc/chm.h
@@ -49,7 +49,8 @@ typedef unsigned __int64        UInt64;
 /* Sparc        */
 /* MIPS         */
 /* PPC          */
-#elif __i386__ || __sun || __sgi || __ppc__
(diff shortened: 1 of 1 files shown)


HBelusca (approved):

binarymaster (approved):
Corrected PR title a bit, would be nice to fix the commit message as well; otherwise LGTM :+1:

SergeGautherie:
> JIRA issue: CORE-17517 seems the closest

Note that, while the ARM32 goal is the same, this PR targets non-MSVC builds, so actually separate.

PR #3550, opened 2021-03-24
Make .rsrc sections of kernel & drivers discardable

zefklop:
## Purpose

Align GCC & MSVC builds on this matter & make kernel handle its own resources properly.


zefklop:
Not this one.

learn-more:
ntoskrnl.exe from 2k3 does not have `.rsrc` as discardable:
image

PR #3555, opened 2021-03-26
[NTOS:SE] Implement anonymous logon token impersonation

GeoB99:
This implements the `NtImpersonateAnonymousToken` system call function, so that we can finally impersonate the logon token whether by calling the system call or `ImpersonateAnonymousToken` user mode service from ADVAPI32 library.
## Proposed Changes
* Add `EveryoneIncludesAnonymous` registry value to Lsa (Local Security Authority) registry key
* Add the prototype of `SepRegQueryHelper` to the internal kernel header, making it public for other kernel code to use, and also annotate its parameters and include a documentation header
* Implement `SepImpersonateAnonymousToken`

### DIFF
--- a/ntoskrnl/se/srm.c
+++ b/ntoskrnl/se/srm.c
@@ -74,14 +74,38 @@ PSEP_LOGON_SESSION_TERMINATED_NOTIFICATION SepLogonNotifications = NULL;
 
 /* PRIVATE FUNCTIONS **********************************************************/
 
+/**
+ * @brief
+ * A private registry helper that returns the desired value
+ * data based on the specifics requested by the caller.
+ *
+ * @param[in] KeyName
+ * Name of the key.
+ *
+ * @param[in] ValueName
+ * Name of the registry value.
+ *
+ * @param[in] ValueType
+ * The type of the registry value.
+ *
+ * @param[in] DataLength
+ * The data length, in bytes, representing the size of the registry value.
+ *
+ * @param[out] ValueData
+ * The requested value data provided by the function.
+ *
+ * @return
+ * Returns STATUS_SUCCESS if the operations have completed successfully,
+ * otherwise a failure NTSTATUS code is returned.
+ */
 NTSTATUS
 NTAPI
 SepRegQueryHelper(
-    PCWSTR KeyName,
-    PCWSTR ValueName,
-    ULONG ValueType,
-    ULONG DataLength,
-    PVOID ValueData)
+    _In_ PCWSTR KeyName,
+    _In_ PCWSTR ValueName,
+    _In_ ULONG ValueType,
+    _In_ ULONG DataLength,
+    _Out_ PVOID ValueData)
 {
     UNICODE_STRING ValueNameString;
     UNICODE_STRING KeyNameString;
@@ -127,7 +151,6 @@ SepRegQueryHelper(
         Status = STATUS_OBJECT_TYPE_MISMATCH;
         goto Cleanup;
     }
-
 
     if (ValueType == REG_BINARY)
     {
--- a/ntoskrnl/se/token.c
+++ b/ntoskrnl/se/token.c
@@ -363,6 +363,119 @@ Quit:
     return STATUS_SUCCESS;
 }
 
+/**
+ * @brief
+ * Private function that impersonates the system's anonymous logon token.
+ * The major bulk of the impersonation procedure is done here.
+ *
+ * @param[in] Thread
+ * The executive thread object that is to impersonate the client.
+ *
+ * @param[in] PreviousMode
+ * The access processor mode, indicating if the call is executed
+ * in kernel or user mode.
+ *
+ * @return
+ * Returns STATUS_SUCCESS if the impersonation has succeeded.
+ * STATUS_UNSUCCESSFUL is returned if the primary token couldn't be
+ * obtained from the current process to perform additional tasks.
+ * STATUS_ACCESS_DENIED is returned if the process' primary token is
+ * restricted, which for this matter we cannot impersonate onto a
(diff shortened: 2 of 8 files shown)


GeoB99:
**Windows Server 2003**
Capture
**ReactOS**
Capture2

ntoskrnl/se/token.c
```diff
@@ -363,6 +363,156 @@ SepCompareTokens(
...
+    ProcessToken = PsReferencePrimaryToken(PsGetCurrentProcess());
+    if (!ProcessToken)
+    {
+        DPRINT1("SepImpersonateAnonymousToken(): Couldn't be able to get the process' primary token, bail out...\n");
+        return STATUS_UNSUCCESSFUL;
+    }
+
+    /* Now, is the token from the current process restricted? */
+    IsTokenRestricted = SeTokenIsRestricted(ProcessToken);
+    if (IsTokenRestricted)
```
disean, line 494:
```suggestion
    if (SeTokenIsRestricted(ProcessToken))
```

ntoskrnl/se/token.c
```diff
@@ -363,6 +363,155 @@ SepCompareTokens(
...
+ */
+static
+NTSTATUS
+SepIncludeEveryoneGroupAnonymous(
+    _Out_ PBOOLEAN IncludeEveryone)
+{
+    NTSTATUS Status;
+    ULONG IncludeEveryoneValueData;
+    static UNICODE_STRING LsaKey = RTL_CONSTANT_STRING(L"\\Registry\\Machine\\SYSTEM\\CurrentControlSet\\Control\\Lsa");
+    static UNICODE_STRING IncludeEveryoneValue = RTL_CONSTANT_STRING(L"EveryoneIncludesAnonymous");
```
Extravert-ir, line 389: You seem to use only `Buffer` fields from these. Why do you need a `UNICODE_STRING` then?

ntoskrnl/se/token.c
```diff
@@ -363,6 +363,155 @@ SepCompareTokens(
...
+    PAGED_CODE();
+
+    /* Invoke the registry helper to return the data value for us */
+    Status = SepRegQueryHelper(LsaKey.Buffer,
+                               IncludeEveryoneValue.Buffer,
+                               REG_DWORD,
+                               sizeof(ULONG),
```
Extravert-ir, line 396:
```suggestion
                               sizeof(IncludeEveryoneValueData),
```
(just a more future-proof way to write)

ntoskrnl/se/token.c
```diff
@@ -363,6 +363,155 @@ SepCompareTokens(
...
+                               &IncludeEveryoneValueData);
+    if (!NT_SUCCESS(Status))
+    {
+        DPRINT1("SepIncludeEveryoneGroupAnonymous(): Failed to query the registry value (Status 0x%lx)\n", Status);
+        return Status;
+    }
+
+    /* Cache the gathered value to the argument parameter */
+    *IncludeEveryone = IncludeEveryoneValueData;
```
Extravert-ir, line 405: I'm not sure should we care about that tbh, but this will convert everything which is >0xFF to FALSE

ntoskrnl/se/srm.c
```diff
@@ -74,14 +74,38 @@ PSEP_LOGON_SESSION_TERMINATED_NOTIFICATION SepLogonNotifications = NULL;
...
+ * @param[out] ValueData
+ * The requested value data provided by the function.
+ *
+ * @return
+ * Returns STATUS_SUCCESS if the operations have completed successfully,
+ * otherwise a failure NTSTATUS code is returned.
+ */
 NTSTATUS
 NTAPI
 SepRegQueryHelper(
```
Extravert-ir, line 103: This helper actually does a very similar job to `IopGetRegistryValue`. Maybe we should make a couple of "global" helpers for registry. Should probably go to `Rtlp`

Extravert-ir (changes requested):

ntoskrnl/se/srm.c
GeoB99: That's definitely a neat idea but that is outside the scope of this PR so let's keep it as is for now. 😛
This is something that can be doable in a different PR.

ntoskrnl/se/token.c
```diff
@@ -4304,15 +4417,67 @@ NtFilterToken(IN HANDLE ExistingTokenHandle,
...
+    if (!NT_SUCCESS(Status))
+    {
+        DPRINT1("NtImpersonateAnonymousToken(): Failed to impersonate the token (Status 0x%lx)\n", Status);
+        ObDereferenceObject(Thread);
+        return Status;
+    }
+
+    /* Impersonation accomplished with success */
+    ObDereferenceObject(Thread);
+    return Status;
```
Extravert-ir, line 4480: These are identical branches

Extravert-ir (changes requested):
Otherwise everything is OK

ntoskrnl/se/token.c
```diff
@@ -4304,15 +4417,67 @@ NtFilterToken(IN HANDLE ExistingTokenHandle,
...
+    if (!NT_SUCCESS(Status))
+    {
+        DPRINT1("NtImpersonateAnonymousToken(): Failed to impersonate the token (Status 0x%lx)\n", Status);
+        ObDereferenceObject(Thread);
+        return Status;
+    }
+
+    /* Impersonation accomplished with success */
+    ObDereferenceObject(Thread);
+    return STATUS_SUCCESS;
```
Extravert-ir, line 4480:
```suggestion
    }

    ObDereferenceObject(Thread);
    return Status;
```

Extravert-ir (approved):

PR #3562, opened 2021-03-29
[UNIATA] DriverEntry(): Fix 'Paramaters' typo, in reg key names

SergeGautherie:
Read the expected/actual keys, otherwise results are always nonexistent/false.

JIRA issue: [CORE-17524](https://jira.reactos.org/browse/CORE-17524)

While here, fix typo elsewhere too.

### DIFF
--- a/drivers/storage/ide/uniata/id_ata.cpp
+++ b/drivers/storage/ide/uniata/id_ata.cpp
@@ -10754,11 +10754,19 @@ DriverEntry(
     }
 
     if(WinVer_Id() >= WinVer_2k) {
+#ifndef __REACTOS__
         if(AtapiRegCheckParameterValue(NULL, L"Paramaters\\PnpInterface", L"1", 0)) {
+#else
+        if(AtapiRegCheckParameterValue(NULL, L"Parameters\\PnpInterface", L"1", 0)) {
+#endif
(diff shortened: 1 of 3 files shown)


drivers/storage/ide/uniata/id_ata.cpp
```diff
@@ -10754,11 +10754,19 @@ DriverEntry(
     }
 
     if(WinVer_Id() >= WinVer_2k) {
+#ifndef __REACTOS__
         if(AtapiRegCheckParameterValue(NULL, L"Paramaters\\PnpInterface", L"1", 0)) {
+#else
+        if(AtapiRegCheckParameterValue(NULL, L"Parameters\\PnpInterface", L"1", 0)) {
+#endif
```
GeoB99, line 10761: Is it worth enclosing this part of code in ifndef cases just for a typo? Can't be that sent upstream?

SergeGautherie: Yes. As far as I know, there is no upstream repository, and @Alter-1 has been inactive for quite a while, sadly.

HBelusca (approved):

binarymaster (commented):
Please change commit title from `[TRANSLATION]` to `[CMD]` according to the module. Otherwise LGTM.

PR #3567, opened 2021-03-30
[WIN32K:NTUSER] Add WindowArrangementActive flag support

sykhro:
## Purpose

This PR adds support for disabling window snappiung via WindowArrangementActive in `HKEY_CURRENT_USER\Control Panel\Desktop`

## Proposed changes

- Add the WindowArrangementActive flag and set it active by default (matches current ReactOS behaviour)
- Add a global and initialize it along with other parameters in `sysparams` (would `ntuser` be a better choice?)
- Check the flag in `nosclient` before performing the snap

### DIFF
--- a/win32ss/user/ntuser/desktop.c
+++ b/win32ss/user/ntuser/desktop.c
@@ -37,6 +37,7 @@ HDC ScreenDeviceContext = NULL;
 PTHREADINFO gptiDesktopThread = NULL;
 HCURSOR gDesktopCursor = NULL;
 PKEVENT gpDesktopThreadStartedEvent = NULL;
+BOOL g_AllowSnappingWindows = TRUE; // Default ReactOS behaviour is to snap windows
 
 /* OBJECT CALLBACKS **********************************************************/
 
@@ -259,6 +260,16 @@ InitDesktopImpl(VOID)
     KeInitializeEvent(gpDesktopThreadStartedEvent,
                       SynchronizationEvent,
                       FALSE);
+
+    /* Check if user has disabled window snapping */
+    DWORD EntryValue = 0;
+    if (RegReadUserSetting(L"Control Panel\\Desktop",
+                           L"WindowArrangementActive",
+                           REG_SZ, &EntryValue, sizeof(EntryValue)))
+    {
+        EntryValue = atoi((PCHAR)&EntryValue);
+        g_AllowSnappingWindows = EntryValue != 0;
+    }
 
     return STATUS_SUCCESS;
 }
--- a/boot/bootdata/hivedef.inf
+++ b/boot/bootdata/hivedef.inf
@@ -82,6 +82,7 @@ HKCU,"Control Panel\Desktop","PowerOffTimeOut",2,"0"
 HKCU,"Control Panel\Desktop","MenuShowDelay",2,"400"
 HKCU,"Control Panel\Desktop","WheelScrollLines",2,"3"
 HKCU,"Control Panel\Desktop","WheelScrollChars",2,"3"
+HKCU,"Control Panel\Desktop","WindowArrangementActive",2,"1"
 
 HKCU,"Control Panel\Desktop\WindowMetrics","AppliedDPI",0x00010003,0x00000060
 HKCU,"Control Panel\Desktop\WindowMetrics","ScrollWidth",2,"16"
--- a/win32ss/user/ntuser/desktop.h
+++ b/win32ss/user/ntuser/desktop.h
@@ -87,6 +87,7 @@ extern HDC ScreenDeviceContext;
 extern PTHREADINFO gptiForeground;
 extern PTHREADINFO gptiDesktopThread;
 extern PKEVENT gpDesktopThreadStartedEvent;
+extern BOOL g_AllowSnappingWindows;
 
 typedef struct _SHELL_HOOK_WINDOW
 {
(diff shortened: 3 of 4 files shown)


win32ss/user/ntuser/sysparams.c
```diff
@@ -371,6 +372,15 @@ NtUserUpdatePerUserSystemParameters(
         g_PaintDesktopVersion = FALSE;
     bResult = TRUE;
 
+    DWORD dwValue = 0;
+    NTSTATUS Status = RegReadUserSetting(L"Control Panel\\Desktop",
+                                             L"WindowArrangementActive",
+                                             REG_SZ, &dwValue, sizeof(dwValue));
```
GeoB99, line 378: This is wrong, `RegReadUserSetting` returns a boolean value and not a NTSTATUS one. Also there's extra space in the last two lines.

See https://git.reactos.org/?p=reactos.git;a=blob;f=win32ss/user/ntuser/misc/registry.c;hb=4c731adc041e611622c2dbb39cd67d49bd48ed19#l158

win32ss/user/ntuser/sysparams.c
```diff
@@ -371,6 +372,15 @@ NtUserUpdatePerUserSystemParameters(
...
+    if(NT_SUCCESS(Status)) {
+        dwValue = atoi((char *)&dwValue);
```
GeoB99, line 380:
```suggestion
        dwValue = atoi((PCHAR)&dwValue);
```

GeoB99 (changes requested):

sykhro:
Not knowing how windows exactly does this (early win10 added a syscall which seems related but the flag is present since win7) I've decided to rework it and set the flag inside of the private desktop init function instead of `NtUserUpdatePerUserSystemParameters` which wouldn't match windows' behaviour. An alternative would be to create a dedicated init function and call it at the bottom of win32k's `DriverEntry`, not sure if it's worth it though (given that it's just a single check)

win32ss/user/ntuser/desktop.c
```diff
@@ -260,6 +261,16 @@ InitDesktopImpl(VOID)
                       SynchronizationEvent,
                       FALSE);
 
+    /* Check if user has disabled window snapping */
+    DWORD dwValue = 0;
+    BOOL ret = RegReadUserSetting(L"Control Panel\\Desktop",
+                                            L"WindowArrangementActive",
+                                            REG_SZ, &dwValue, sizeof(dwValue));
+    if(ret) {
```
learn-more, line 269: Please follow the style of the surrounding code.

win32ss/user/ntuser/desktop.c
```diff
@@ -260,6 +261,16 @@ InitDesktopImpl(VOID)
...
+        dwValue = atoi((PCHAR)&dwValue);
+        g_AllowSnappingWindows = dwValue == 1;
```
learn-more, line 271: Only on 1 it should be active?
I would assume that windows uses something more like `g_AllowSnappingWindows = dwValue != 0`

sykhro: Just checked out of curiosity (win10 20h2), if the registry entry is a number then it does indeed work if `dwValue != 0`; otherwise it remains disabled. I'll change it to `g_AllowSnappingWindows = dwValue != 0`, can switch to `strtol` if you want me to match windows' behavior accurately

win32ss/user/ntuser/desktop.c
sykhro: Could you give me an example? I'm not sure what to change, should I drop the hungarian notation and use something like `Ret` and `EntryValue`?

binarymaster:
```suggestion
    if (ret)
    {
```
See https://reactos.org/wiki/Coding_Style

win32ss/user/ntuser/desktop.c
```diff
@@ -260,6 +261,17 @@ InitDesktopImpl(VOID)
                       SynchronizationEvent,
                       FALSE);
 
+    /* Check if user has disabled window snapping */
+    DWORD EntryValue = 0;
+    BOOL HasEntry = RegReadUserSetting(L"Control Panel\\Desktop",
+                                       L"WindowArrangementActive",
+                                       REG_SZ, &EntryValue, sizeof(EntryValue));
+    if (HasEntry)
```
HBelusca, line 269:
```suggestion
    if (RegReadUserSetting(L"Control Panel\\Desktop",
                           L"WindowArrangementActive",
                           REG_SZ, &EntryValue, sizeof(EntryValue)))
```

HBelusca: Also, why isn't this done in SpiUpdatePerUserSystemParameters() ?

HBelusca (changes requested):

win32ss/user/ntuser/desktop.c
sykhro: I was told on IRC it would be more appropriate to do it here since we're not sure how Windows does it, I can move it there if it's more appropriate

binarymaster (changes requested):
I'm doubling the change request by Hermes above. It indeed should be done in `SpiUpdatePerUserSystemParameters()` and in other places where SPI variables are being set or read. See MSDN documentation about **SPI_GETWINARRANGING**:

https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-systemparametersinfoa

image

Here you can see the places where it should be added (took `VAL_SCRLLCHARS` as an example):
- https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=VAL_SCRLLCHARS
- https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=uiWheelScrollChars

PR #3570, opened 2021-03-31
[WALLPAPER] Move wallpaper folder from modules/ to media/

Seplx:
## Purpose
This is more intuitive to me rather than the modules/ folder used before, it's also consistent with the other material in the media/ folder such as fonts and themes

## Proposed changes
move modules/wallpapers/ to media/wallpapers/.
Edit CMake list of of both modules/ and media/ folders to correspond to this change.


Seplx:
> Things in `modules/` are those that can be removed or added during build configuration, and that are super-optional, contrary to the fonts.

Ok thanks the naming makes sense now, closing this then.

PR #2868, opened 2020-05-27
[GITHUB:WORKFLOWS] Restore 2 features, from AppVeyor/Travis

SergeGautherie:
Addendum to a0283ecea3c10d081320a6fab0821a03cda6b7f8.

PR #3576, opened 2021-04-02
[GITHUB] Downgrade to windows-2016 from windows-latest

SergeGautherie:
Temporary, until MS VS2019 itself is fixed.

JIRA issue: [CORE-17533](https://jira.reactos.org/browse/CORE-17533)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -105,7 +105,7 @@ jobs:
 
   build-clang-cl-i386:
     name: Clang-CL (i386)
-    runs-on: windows-latest
+    runs-on: windows-2016
     steps:
     - name: Install packages
       run: |
@@ -142,7 +142,7 @@ jobs:
 
   build-msvc-i386:
     name: MSVC (i386)
-    runs-on: windows-latest
+    runs-on: windows-2016
     steps:
     - name: Install packages
       run: choco install ninja -y
@@ -186,7 +186,7 @@ jobs:
 
   build-msvc-amd64:
     name: MSVC (amd64)
-    runs-on: windows-latest
+    runs-on: windows-2016
     steps:
     - name: Install packages
       run: choco install ninja -y
@@ -231,7 +231,7 @@ jobs:
 
   build-msbuild-i386:
     name: MSBuild (i386)
-    runs-on: windows-latest
+    runs-on: windows-2016
     steps:
     - name: Install Flex and Bison
       run: |
@@ -251,6 +251,6 @@ jobs:
       run: |
         mkdir build
         cd build
-        cmake -G "Visual Studio 16 2019" -A Win32 -DCMAKE_TOOLCHAIN_FILE:FILEPATH=toolchain-msvc.cmake -DARCH:STRING=i386 -DENABLE_ROSTESTS=1 -DENABLE_ROSAPPS=1 ${{github.workspace}}\src
+        cmake -G "Visual Studio 15 2017" -A Win32 -DCMAKE_TOOLCHAIN_FILE:FILEPATH=toolchain-msvc.cmake -DARCH:STRING=i386 -DENABLE_ROSTESTS=1 -DENABLE_ROSAPPS=1 ${{github.workspace}}\src
     - name: Build
       run: cmake --build ${{github.workspace}}\build --target bootcd --target livecd


.github/workflows/build.yml
```diff
@@ -105,152 +105,152 @@ jobs:
...
-      working-directory: ${{github.workspace}}\build
-      run: |
-        cmake --build . --target bootcd
-        cmake --build . --target livecd
+      - name: Install packages
+        run: |
+          choco install ninja -y
+          choco install --x86 -y llvm
+      - name: Install Flex and Bison
+        run: |
```
HBelusca, line 115: what's the matter with all these whitespace changes?

SergeGautherie: Ease comparing, copy-pasting, ...
At least, use the same indentation for Linux and Windows...

ColinFinck: I second that. YAML is ill-defined enough that there is no standard regarding whitespaces.
If the original maintainer decided for the current whitespace style, don't change that as part of a PR that is meant to change something else entirely.

SergeGautherie: > Formatting should be left at the discretion of the maintainer(s) of the respective file.

I am not sure what your point is:
this file has self-inconsistencies and is not listed in CODEOWNERS.

> All you did here is just to create meaningless noise with such whitespace changes,

I argue that all this inconsistency does is to create pain when I work on this file, especially when comparing/editing both Linux and Windows builds.

> so please revert them and keep only the changes that are in the context with this PR.

Arguable, considering (worth) 8b8108ddeefd84d527a43f3eabab4a0134638636 for example.

--

> YAML is ill-defined enough that there is no standard regarding whitespaces.

"Agreed", but I want to fix ROS file, not debate YAML definition.

> If the original maintainer decided for the current whitespace style,

Initial (a0283ecea3c10d081320a6fab0821a03cda6b7f8) committer is @Extravert-ir.
What is the rationale for using different indentations for different builds?

> don't change that as part of a PR that is meant to change something else entirely.

By popular demand, I'll split that part to a separate PR...

learn-more: It is very simple:

1. You are not the maintainer of this file
2. You have been asked numerous times not to change all kinds of random stuff unrelated to the PR.
3. Every time we have more or less the same discussion

Should we just close every PR where you do this?

SergeGautherie: > It is very simple:

I wish it would be.

>     1. You are not the maintainer of this file

Obviously, but, as often, you don't tell who is, nor answer the other questions...

>     2. You have been asked numerous times not to change all kinds of random stuff unrelated to the PR.

True, and the very _opposite_ too:
see, for examples,
https://github.com/reactos/reactos/pull/3066#issuecomment-715343541
where @Extravert-ir closed all my "splitted" TaskMgr PRs, asking for an all-in-one instead,
or
https://github.com/reactos/reactos/pull/3512#issuecomment-795594098
where @Extravert-ir asked to do more dev within an "unrelated" PR,
NB:
No special blame to @Extravert-ir here, only examples from file initial committer.

Also, I would debate whether formatting this file is actually unrelated to updating it.

>     3. Every time we have more or less the same discussion

Mostly, yes, because either way I try to submit my PRs, it seems there is often a (different) dev who disagree :-(

> Should we just close every PR where you do this?

Sadly, that often happens no matter what.
I still "wonder" why a contributor like me can't submit a simple PR with 2 separate commits like this one,
while a dev can commit a single commit with separate formatting changes merged with code changes...

learn-more: > > It is very simple:
> 
> I wish it would be.
> 
> > ```
> > 1. You are not the maintainer of this file
> > ```
> 
> Obviously, but, as often, you don't tell who is, nor answer the other questions...

As it's not listed, I use the magic of git to see who is most active in this file, and that appears to be @Extravert-ir 

> 
> > ```
> > 2. You have been asked numerous times not to change all kinds of random stuff unrelated to the PR.
> > ```
> 
> True, and the very _opposite_ too:
> see, for examples,
> [#3066 (comment)](https://github.com/reactos/reactos/pull/3066#issuecomment-715343541)
> where @Extravert-ir closed all my "splitted" TaskMgr PRs, asking for an all-in-one instead,
> or

I sincerely hope you understand the difference between multiple pr's to change a few bits here and there,
and multiple PR's where you re-format an entire file.

> [#3512 (comment)](https://github.com/reactos/reactos/pull/3512#issuecomment-795594098)
> where @Extravert-ir asked to do more dev within an "unrelated" PR,
> NB:
> No special blame to @Extravert-ir here, only examples from file initial committer.
> 
> Also, I would debate whether formatting this file is actually unrelated to updating it.
> 

Here you are moving files around, and his suggestion is that if there is a commit that touches all lines with no changes, it would make sense to do this other action as well, since otherwise there are always 2 commits to skip past while looking for the history.

> > ```
> > 3. Every time we have more or less the same discussion
> > ```
> 
> Mostly, yes, because either way I try to submit my PRs, it seems there is often a (different) dev who disagree :-(
> 

Yeah, most of us are not paid devs, most of us do this in their spare time, and there are no rules set in stone for a lot of things.
So depending on the component being changed, other people will chime in.

> > Should we just close every PR where you do this?
> 
> Sadly, that often happens no matter what.
> I still "wonder" why a contributor like me can't submit a simple PR with 2 separate commits like this one,
> while a dev can commit a single commit with separate formatting changes merged with code changes...

It is preferred that a dev does not do this either,
but a maintainer (or regular dev) of a certain part is more or less in charge of formatting code / keeping it tidy.

GeoB99:
Will you continue working on this (and also revert the style changes)?

.github/workflows/build.yml
SergeGautherie: > As it's not listed, I use the magic of git to see who is most active in this file, and that appears to be @Extravert-ir

Hum, why not simply update CODEOWNERS?
Also, I do remember I used to look at history (to pick reviewers if I remember correctly) and got complain that doing that was not appropriate...

> I sincerely hope you understand the difference between multiple pr's to change a few bits here and there,
> and multiple PR's where you re-format an entire file.

Yes and no, I am not sure what your point/example(s) is.
I mostly feel someone often finds some sort of (different/varying) reason to say 'no'...

> Here you are moving files around, and his suggestion is that if there is a commit that touches all lines with no changes, it would make sense to do this other action as well, since otherwise there are always 2 commits to skip past while looking for the history.

#3512: Well, my intent is the opposite, not moving and changing things at he same time.
History: Yes, but having separate commits ease sorting things out, like (regression) testing...

> Yeah, most of us are not paid devs, most of us do this in their spare time, and there are no rules set in stone for a lot of things.

I know...

> So depending on the component being changed, other people will chime in.

That is somewhat a problem.
For example with this PR, if Victor is in charge, why do other people comment without waiting for him first?
In other words, if different devs/modules use different rules, then why not make the laters more explicit, to everyone?

> It is preferred that a dev does not do this either,
> but a maintainer (or regular dev) of a certain part is more or less in charge of formatting code / keeping it tidy.

That is part of the issue: it is not always obvious why a dev commits as he does then why code can't be "fixed".
Here (#3580 now), why is whitespace style not uniform across the file (initial commit) and why can't it be improved to make (local) dev'ing easier?

SergeGautherie:
> Will you continue working on this (and also revert the style changes)?

What other work are you thinking of?
I would agree to a different style fix, but, in any case, I want to understand what the style rule of this file is (to be) first...

.github/workflows/build.yml
learn-more: > why do other people comment without waiting for him first?

We are no machines?

SergeGautherie: > We are no machines?

??

dreimer1986:
> > Will you continue working on this (and also revert the style changes)?
> 
> What other work are you thinking of?
> I would agree to a different style fix, but, in any case, I want to understand what the style rule of this file is (to be) first...

Well, it's quite easy. You open a file you wanna change for a PR and you see a code style. Perfect! Now do what you were planning to do in that file and then save and close it. Next you make the PR. The coding style is no object of a fix or feature PR as it's neither a fix nor a feature. If you think the style is bad, then make a entirely separate PR for that and wait for the replies raining upon ya to that one.

Extravert-ir:
@SergeGautherie 
As you may have noticed, most of your recent PRs are just ignored by people. How do you think, maybe there is a reason for that?
90% of the time you're doing nitpicking around the code, which either don't change anything regarding bugfixes or features, or can potentially introduce a new bug without any benefit.
ReactOS is a project in an Alpha state, and this is simply not the right time to deal with fixing comments or indentation. Ideally, this stuff should just be automated (and you can help us here, that would be much more appreciated)

What we appreciate here are PRs with new features and bugfixing, everything else will be taken with much less amount of enthusiasm (or may be ignored at all)

> In other words, if different devs/modules use different rules, then why not make the laters more explicit, to everyone?

We are developers, we like to write code, not rules. If you want to formalize them, prepare a draft and then welcome to our chat for discussion. If it looks reasonable, we can make it "official"

> Here (#3580 now), why is whitespace style not uniform across the file (initial commit) and why can't it be improved to make (local) dev'ing easier?

The indent is wrong, but it's not a big deal. I'll fix it next time I do something in this file. Thank you for pointing that out, but I will not accept the change
Same for switching to "windows-2016" btw. The idea of this step is to catch issues with the latest versions of software. For getting more "stable" builds, there is https://iso.reactos.org

SergeGautherie:
> If you think the style is bad, then make a entirely separate PR for that and wait for the replies raining upon ya to that one.

Ok, let's try that again: #3645 and #3663.

--

> Ideally, this stuff should just be automated (and you can help us here, that would be much more appreciated)

Maybe. How?

> We are developers, we like to write code, not rules. If you want to formalize them, prepare a draft and then welcome to our chat for discussion. If it looks reasonable, we can make it "official"

Well, there is obviously no way I can guess which (different) rules each dev wants...

> I'll fix it next time I do something in this file. Thank you for pointing that out, but I will not accept the change

I understand, but that doesn't make sense to me, as pain continues meanwhile.
Anyway. (Superseded by 246dbd3.)

> Same for switching to "windows-2016" btw. The idea of this step is to catch issues with the latest versions of software. For getting more "stable" builds, there is https://iso.reactos.org

Well, catching is fine, but what then? I obviously assumed CI is meant to actually test ROS, which requires usable artifatcs.
As far as I know, https://iso.reactos.org cannot be used to build/test any (PR) code...

Again, this PR is refused, then superseded by #3647 which is committed.
That is exactly what I say: two different devs, two opposite actions...
Another example? That very same PR restored one of the features that were refused on #2868!
(Hopefully, the other one will eventually be restored too...)

PR #3579, opened 2021-04-03
[CMAKE] Store and re-use M4 and BISON_PKGDATADIR env variables

learn-more:
## Purpose

Configure time environment variables are not stored / re-used by cmake,
and bison/flex need environment variables to function.

JIRA issue: [ROSBE-174](https://jira.reactos.org/browse/ROSBE-174)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,6 +92,41 @@ endif()
 find_package(BISON REQUIRED)
 find_package(FLEX REQUIRED)
 
+if(MSVC_IDE)
+    # Bison needs M4 and BISON_PKGDATADIR set at build time,
+    # but visual studio is hardly ever opened from the configure-time environment.
+    # Since cmake does not support setting env variables for a custom command,
+    # we have to write a wrapper that sets the variables and then executes bison.
+    # Idea taken from https://stackoverflow.com/a/35032051/4928207
+    if(DEFINED ENV{M4})
+        # Store this environment variable for configure re-runs from withing visual studio.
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -84,6 +84,41 @@ endif()
...
+    if(DEFINED ENV{M4})
+        # Store this environment variable for configure re-runs from withing visual studio.
+        SET(ROS_SAVED_M4 "$ENV{M4}" CACHE INTERNAL "")
+    endif()
+    if(DEFINED ENV{BISON_PKGDATADIR})
+        SET(ROS_SAVED_BISON_PKGDATADIR "$ENV{BISON_PKGDATADIR}" CACHE INTERNAL "")
+    endif()
+
+    # Tell the user about a misconfigured environment
+    if("x${ROS_SAVED_M4}x" STREQUAL "xx" OR "x${ROS_SAVED_BISON_PKGDATADIR}x" STREQUAL "xx")
```
HBelusca, line 102: Why testing the strings without the 'x' guards would be insufficient?

learn-more: This is what some cmake modules did.

CMakeLists.txt
```diff
@@ -92,6 +92,41 @@ endif()
...
+        message(FATAL_ERROR "\nM4 or BISON_PKGDATADIR environment variables not set, cannot continue!\n"
```
Extravert-ir, line 111: This check should be moved outside `if(MSVC_IDE)`

learn-more: As discussed in MM:
This will stay in place for now, since there is no wiki page for general MSVC building yet.

PR #3580, opened 2021-04-03
[FORMATTING] GitHub: Enforce 2-space indentation within 'steps:'

SergeGautherie:
Split off from #3576.

## Purpose

Ease comparing, copy-pasting, ...,
especially Linux and Windows builds within build.yml.

## Proposed changes

I prefer 2-space indentation.
If you prefer no indentation, just ask for it.
In any case, please, choose an identical indentation...

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -107,150 +107,150 @@ jobs:
     name: Clang-CL (i386)
     runs-on: windows-latest
     steps:
-    - name: Install packages
-      run: |
-        choco install ninja -y
-        choco install --x86 -y llvm
-    - name: Install Flex and Bison
-      run: |
-        curl -O https://svn.reactos.org/storage/vperevertkin/flexbison.7z
-        7z x flexbison.7z -O${{github.workspace}}\bin
-        echo "${{github.workspace}}\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
-        echo "BISON_PKGDATADIR=${{github.workspace}}\bin\share\bison" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
-        echo "M4=${{github.workspace}}\bin\m4.exe" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
-    - name: Add CL to PATH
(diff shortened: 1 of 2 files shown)


Extravert-ir:
OK, maybe it's not clear from our current contribution guidelines, so let me say it (I feel like it was told many times already, but still)
**We don't accept pull requests with fixing minor style-related things which don't fix anything.**
This belongs to grammar mistakes in comments, indentation, code style.

Why don't we accept those?

1. They pollute commit log without any reason, such things are fixed while working on the code
2. They take too much time from team members for (again) no reason. Look at the #3576 - 5 team members spent time to looking at it, and writing comments.

Respect your and our time - fix real bugs, leave code organization to us

SergeGautherie:
> **We don't accept pull requests with fixing minor style-related things which don't fix anything.**

In this very PR case, I "disagree": this commit is a fix, not a style thing only.
As in,
while mismatching indentation in C (for example) usually has no impact on compilation,
mismatching spaces before - in YAML (this file) do break its parsing at build time.
In other words,
while current script is not broken, syn'cing its indentation does prevent introducing **actual failures**.

>     1. They pollute commit log without any reason, such things are usually fixed while working on the code

First, I wonder what `[FORMATTING]` is for, if such commits are seen as pollution.
Then, I do (now) try to follow this rule: I submitted this commit inside #3576 (where I do work on this very code), refused, so I split it off, refused.
From my point of view, that is so contradictory...

>     2. They take too much time from team members for (again) no reason. Look at the #3576 - 5 team members spent time looking at it and writing comments.

(Again) I disagree.
(I am also sad about "team members vs contributors", but that is another debate.)

> Respect your and our time - fix real bugs, leave code organization to us

I do tell you this real issue is very much loosing my time indeed:
each time I work on this file, and not only when I submit a PR about it, I need to check where I need to apply which indentation, especially when copy-pasting/moving around/global editing.
As expectable, even after spending time to do that, I sometimes miss some cases, get a build failure and spend some more time doule-checking/fixing it.
Do such "mistake" happen to me only? I wonder.

I wish it would be much simpler to make working with that code just easier...

SergeGautherie:
Of course, please, reopen.

SergeGautherie:
Eventually superseded by 246dbd3384716c045c0.

PR #3582, opened 2021-04-03
[BROWSEUI][BROWSEUI_APITEST] Add CLSID_ACLHistory stubs

katahiromz:
## Purpose
Just a preparation for implementation of `CLSID_ACLHistory` that realizes URL history auto-completion.

## Proposed changes

- Add IACLHistory testcase into `browseui_apitest`.
- Initial implement `CLSID_ACLHistory` as stub.

## Comparison
BEFORE:
SHACF_URLHISTORY-Unknown-error-0x80040154
AFTER:
after


katahiromz:
WinXP:
winxp
Win2k3:
win2k3
Win10:
win10
ReactOS:
after

dll/win32/browseui/aclhistory.h
```diff
@@ -0,0 +1,34 @@
...
+ * PURPOSE:     Implement CLSID_ACLHistory for auto-completion
+ * COPYRIGHT:   Copyright 2021 Katayama Hirofumi MZ <katayama.hirofumi.mz@gmail.com>
+ */
+
+#pragma once
+
+class CACLHistory
+    : public CComCoClass<CACLHistory, &CLSID_ACLHistory>
+    , public CComObjectRootEx<CComMultiThreadModelNoCS>
+    , public IEnumString
```
learn-more, line 13: It would be interesting (probably not for this PR) to see if atl has a helper class for IEnum(String), since we are writing a lot of duplicate code now.

dll/win32/browseui/res/autocompletehistory.rgs
```diff
@@ -0,0 +1,13 @@
+HKCR
+{
+	NoRemove CLSID
+	{
+		ForceRemove {00BB2764-6A77-11D0-A535-00C04FD7D062} = s 'Microsoft History AutoComplete List'
```
learn-more, line 5: Is there anything that depends on the `Microsoft` here?
If not, I would prefer it to not include Microsoft in the name.

modules/rostests/apitests/browseui/IACLHistory.cpp
```diff
@@ -0,0 +1,56 @@
...
+#define _UNICODE
+#define UNICODE
+#include <apitest.h>
+#include <shlobj.h>
+#include <atlbase.h>
+#include <atlcom.h>
+#include <stdio.h>
+#include <shellutils.h>
+
+struct CCoInit
```
learn-more, line 17: Might be worth moving this to `apitest.h` or a common header like that?
(Does not have to be done in this PR, but something to keep in mind)

learn-more (approved):
Some minor comments, lgtm with or without those fixed. (However I would like the Microsoft thing to be changed if that does not cause problems)

dll/win32/browseui/res/autocompletehistory.rgs
HBelusca: If you want to change it, maybe replacing "Microsoft" by "ReactOS"?

HBelusca (commented):
lgtm after the MS change.

dll/win32/browseui/res/autocompletehistory.rgs
katahiromz: OK.

PR #3510, opened 2021-03-07
[NTOS:IO] Implement IoGetDeviceInterfaceAlias CORE-17361

oleg-dubinskiy:
## Purpose

Implement `IoGetDeviceInterfaceAlias` routine based on the one from the following commit: https://git.reactos.org/?p=reactos.git;a=commit;h=444c59cb3f83ea3ad6bc929dfe4fa2c47c69f220, with a lot of improvements and refactoring.
It's required by MS audio drivers, so now ReactOS is nearly able to properly load the audio stack from Windows XP/2003 (ks.sys, portcls.sys, swenum.sys, sysaudio.sys and wdmaud.sys). :smiley: 
Although it still doesn't completely fix the problem, but it's almost done and much more improves the situation.

JIRA issue: [CORE-17361](https://jira.reactos.org/browse/CORE-17361)

Supersedes #3874.

## Proposed changes

Differences from the original implementation:

- Add a lot of status failure checks.
- Improve enumerating loop.
- Properly allocate the buffer for some unicode strings.
- Simplify the initialization code for some unicode strings.
- Fix some uninitialized handles.
- Add cleanup section and move all freeing calls there. Call it in all failure checks, to properly free no longer needed stuff in case of failure.
- Additional fix: create only non-volatile subkey(s) in `OpenRegistryHandlesFromSymbolicLink`, since the parent key(s) whose are created in `IoOpenDeviceInterfaceRegistryKey` are non-volatile too. So then created subkey(s) cannot be volatile. It will fix status 0xc0000181 (STATUS_CHILD_MUST_BE_VOLATILE) when calling `IoGetDeviceInterfaceAlias` implementation.

## TODO

- [x] Add more comments in the code.
- [x] Fix possible remaining bug(s) in the code. Figure out why MS sysaudio still fails to free the `AliasSymbolicLinkName` string when it's no longer needed (what seems to happen, judging by a backtrace, see it in the last comment of the ticket).
- [x] Check whether `IoOpenDeviceInterfaceRegistryKey` fix is correct. If no, fix the `IoGetDeviceInterfaceAlias` implementation or find another solution instead.
- [x] This PR relies on https://github.com/reactos/reactos/pull/4292

PR #3583, opened 2021-04-04
[NTOS:IO] OpenRegistryHandlesFromSymbolicLink: Use REG_OPTION_NON_VOLATILE CORE-17361

oleg-dubinskiy:
## Purpose

Use `REG_OPTION_NON_VOLATILE` instead of `REG_OPTION_VOLATILE` in all `ZwCreateKey` calls of `OpenRegistryHandlesFromSymbolicLink`, since the keys created/opened by this function, should be non-volatile (in other words, saved after reboot).
Also `Device Parameters` subkey that is created in `IoOpenDeviceInterfaceRegistryKey` (which uses that routine as well), is non-volatile too, so the parent keys whose contain it, cannot be volatile.
It will fix an error with status `0xc0000181` (`STATUS_CHILD_MUST_BE_VOLATILE`) occuring during loading kernel mode audio drivers from Windows XP/2003, especially checked (debug) versions, with my `IoGetDeviceInterfaceAlias` implementation from previous PR #3510. Also it may fix other error cases.

### DIFF
--- a/ntoskrnl/io/iomgr/deviface.c
+++ b/ntoskrnl/io/iomgr/deviface.c
@@ -103,7 +103,7 @@ OpenRegistryHandlesFromSymbolicLink(IN PUNICODE_STRING SymbolicLinkName,
                          &ObjectAttributes,
                          0,
                          NULL,
-                         REG_OPTION_VOLATILE,
+                         REG_OPTION_NON_VOLATILE,
                          NULL);
     ZwClose(ClassesKey);
     if (!NT_SUCCESS(Status))
@@ -154,7 +154,7 @@ OpenRegistryHandlesFromSymbolicLink(IN PUNICODE_STRING SymbolicLinkName,
                          &ObjectAttributes,
                          0,
                          NULL,
-                         REG_OPTION_VOLATILE,
+                         REG_OPTION_NON_VOLATILE,
                          NULL);
     if (!NT_SUCCESS(Status))
     {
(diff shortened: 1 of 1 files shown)


GeoB99 (approved):
I can confirm on Windows (Server 2003 SP2) that such keys are indeed created as non volatile.

zefklop:
can you add a KM test for IoOpenDeviceInterfaceRegistryKey that shows this behaviour ?
See here for a way: https://stackoverflow.com/questions/4432977/is-there-a-way-to-verify-if-a-windows-registry-key-is-volatile-reg-option-volat

oleg-dubinskiy:
Done @zefklop.
I wrote the new `IoDeviceInterface` kmtest, based on Mark's patch (with `IoOpenDeviceInterfaceRegistryKey` testcase), and when it tries to create a subkey, after opening the interface key on Windows XP and 2003, it returns success and displays received handle to the registry key, like this:
`(modules/rostests/kmtests/ntos_io/IoDeviceInterface.c:89) IoOpenDeviceInterfaceRegistryKey(): success: 0 8000063C
(modules/rostests/kmtests/ntos_io/IoDeviceInterface.c:113) ZwCreateKey(): successfully created subkey: 0 800004D8
(modules/rostests/kmtests/ntos_io/IoDeviceInterface.c:89) IoOpenDeviceInterfaceRegistryKey(): success: 1 80000534
(modules/rostests/kmtests/ntos_io/IoDeviceInterface.c:113) ZwCreateKey(): successfully created subkey: 1 80000504
(modules/rostests/kmtests/ntos_io/IoDeviceInterface.c:89) IoOpenDeviceInterfaceRegistryKey(): success: 2 80000408
(modules/rostests/kmtests/ntos_io/IoDeviceInterface.c:113) ZwCreateKey(): successfully created subkey: 2 80000528`
As I expected. So George's analysis was indeed correct. :slightly_smiling_face: 
PR with that test will be follow soon.

tkreuzer (approved):

PR #3584, opened 2021-04-04
[NTOS:IO] OpenRegistryHandlesFromSymbolicLink: Add missing OBJ_OPENIF CORE-17361

oleg-dubinskiy:
## Purpose

Add missing `OBJ_OPENIF` flags in all `ZwCreateKey` calls of `OpenRegistryHandlesFromSymbolicLink`.
It will prevent this function from forced (re)creating the keys and possible overwriting data, in case they're already exist in registry. If they were already created before, they just should be opened (which seems to be more correct).
It also will improve the situation when loading Windows XP/2003 audio drivers in ReactOS, because that routine is used for loading them as well in my current implementation of PR #3510.

### DIFF
--- a/ntoskrnl/io/iomgr/deviface.c
+++ b/ntoskrnl/io/iomgr/deviface.c
@@ -95,7 +95,7 @@ OpenRegistryHandlesFromSymbolicLink(IN PUNICODE_STRING SymbolicLinkName,
 
     InitializeObjectAttributes(&ObjectAttributes,
                                &GuidString,
-                               OBJ_CASE_INSENSITIVE | OBJ_KERNEL_HANDLE,
+                               OBJ_CASE_INSENSITIVE | OBJ_KERNEL_HANDLE | OBJ_OPENIF,
                                ClassesKey,
                                NULL);
     Status = ZwCreateKey(GuidKeyRealP,
@@ -146,7 +146,7 @@ OpenRegistryHandlesFromSymbolicLink(IN PUNICODE_STRING SymbolicLinkName,
 
     InitializeObjectAttributes(&ObjectAttributes,
                                &SubKeyName,
-                               OBJ_CASE_INSENSITIVE | OBJ_KERNEL_HANDLE,
+                               OBJ_CASE_INSENSITIVE | OBJ_KERNEL_HANDLE | OBJ_OPENIF,
                                *GuidKeyRealP,
                                NULL);
     Status = ZwCreateKey(DeviceKeyRealP,
@@ -164,7 +164,7 @@ OpenRegistryHandlesFromSymbolicLink(IN PUNICODE_STRING SymbolicLinkName,
 
     InitializeObjectAttributes(&ObjectAttributes,
                                &ReferenceString,
-                               OBJ_CASE_INSENSITIVE | OBJ_KERNEL_HANDLE,
+                               OBJ_CASE_INSENSITIVE | OBJ_KERNEL_HANDLE | OBJ_OPENIF,
                                *DeviceKeyRealP,
                                NULL);
     Status = ZwCreateKey(InstanceKeyRealP,


HBelusca:
https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-zwcreatekey

> If the key specified by ObjectAttributes does not exist, the routine attempts to create the key. For this attempt to succeed, the new key must be a direct subkey of the key that is referred to by RootDirectory, and the key that RootDirectory refers to must have been opened for KEY_CREATE_SUB_KEY access.
>
> If the specified key already exists, it is opened and its value is not affected in any way.

HBelusca:
The actual problem of our `OpenRegistryHandlesFromSymbolicLink()` function seems that, if the subkeys don't exist, it always want to create them, even for an "open" operation like when it is called by `IoOpenDeviceInterfaceRegistryKey()` (which can be possible: from that latter function's documentation https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-ioopendeviceinterfaceregistrykey , it can return STATUS_OBJECT_NAME_NOT_FOUND or STATUS_OBJECT_PATH_NOT_FOUND )

oleg-dubinskiy:
> https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-zwcreatekey
> 
> > If the key specified by ObjectAttributes does not exist, the routine attempts to create the key. For this attempt to succeed, the new key must be a direct subkey of the key that is referred to by RootDirectory, and the key that RootDirectory refers to must have been opened for KEY_CREATE_SUB_KEY access.
> > If the specified key already exists, it is opened and its value is not affected in any way.

OK @HBelusca, I checked: nearly everything in that code is correct, according to this MSDN part (which you quoted). So if we still want to create the key in case it doesn't exist, then it needs only to add `KEY_CREATE_SUB_KEY` to `AccessMask` parameter.

> The actual problem of our `OpenRegistryHandlesFromSymbolicLink()` function seems that, if the subkeys don't exist, it always want to create them, even for an "open" operation like when it is called by `IoOpenDeviceInterfaceRegistryKey()` (which can be possible: from that latter function's documentation https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-ioopendeviceinterfaceregistrykey , it can return STATUS_OBJECT_NAME_NOT_FOUND or STATUS_OBJECT_PATH_NOT_FOUND )

But if we need to just open the key instead (as you said in this 2nd comment), and just return the appropriate status if it's not found, then the only way to do this is using `ZwOpenKey` instead.
But unfortunately, I already tried to do this, and it makes the situation worse, at least with checked builds of MS ks.sys, portcls.sys, swenum.sys, sysaudio.sys, wdmaud.sys (although the release ones are still loading successfully): if they don't find the alias GUID key, and then its (seems `NULL`) handle is passed to the next call, it moreover doesn't open, and in such case, it causes a system failure (in my case - BSOD 0x7F).
So another option might be to check availability only of GUID key (we already have `IopOpenInterfaceKey` for that) in `IoGetDeviceInterfaceAlias`, and then don't use the (invalid) handle anywhere else in this rotuine, if the GUID key doesn't exist. But in this case, kmtest results are wrong again. :frowning_face: 
The mandatory criteria of the alias is the availability of the `Device Parameters` subkey in its instance key. But checking it without the failure of opening the parent subkeys is problematical, if any of them is missing. That's probably a reason why we're using `ZwCreateKey` here.
Another thing is, we're using this routine only for opening non-alias symbolic links, whose seem to be always exist. Then it doesn't cause any problems.

HBelusca:
> > The actual problem of our `OpenRegistryHandlesFromSymbolicLink()` function seems that, if the subkeys don't exist, it always want to create them, even for an "open" operation like when it is called by `IoOpenDeviceInterfaceRegistryKey()` (which can be possible: from that latter function's documentation https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-ioopendeviceinterfaceregistrykey , it can return STATUS_OBJECT_NAME_NOT_FOUND or STATUS_OBJECT_PATH_NOT_FOUND )
> 
> But if we need to just open the key instead (as you said in this 2nd comment), and just return the appropriate status if it's not found, then the only way to do this is using `ZwOpenKey` instead.
> But unfortunately, I already tried to do this, and it makes the situation worse, at least with checked builds of MS ks.sys, portcls.sys, swenum.sys, sysaudio.sys, wdmaud.sys (although the release ones are still loading successfully): if they don't find the alias GUID key, and then its (seems `NULL`) handle is passed to the next call, it moreover doesn't open, and in such case, it causes a system failure (in my case - BSOD 0x7F).
[...]

In the second usage of `OpenRegistryHandlesFromSymbolicLink()` however (in the other IoSet*** function), we'll want to have the keys created.

oleg-dubinskiy:
OK, I did more investigations and more deep analysis.
On Windows, this routine calls another function for creating/opening the last two subkeys, which has a BOOLEAN parameter, that allows to either create, or just open registry keys (without any creation). And this is used differently by different functions (that's called either with `TRUE`, or with `FALSE`).
We also should do the similar thing, so just using `OBJ_OPENIF` is not enough in this situation.
I already added it in my new #3874, so assuming that PR supersedes this one.

PR #3566, opened 2021-03-30
WIP: make clang build bootable to desktop

Valeriy-Zainullin:
At least, make some progress in that direction.

## Purpose

_Make bootcd built with clang bootable to the desktop._

JIRA issue: [CORE-8516](https://jira.reactos.org/browse/CORE-8516)

## Proposed changes

To be added.

## TODO

To be added.

PR #3588, opened 2021-04-06
[NTOS:KE] Rewrite KiSystemCallTrampoline in assembly

zefklop:
Instead of making assumptions about what the compiler does with forced-inline functions

GCC version doesn't properly clobber ecx, edx & eax.
MSVC version assumes that the compiler puts return value of forced-inline functions into eax.
CLANG just blows up.
See PR #3566

### DIFF
--- a/ntoskrnl/ke/i386/trap.s
+++ b/ntoskrnl/ke/i386/trap.s
@@ -220,4 +220,38 @@ _KiConvertToGuiThread@0:
     /* return to the caller */
     ret
 
+/*
+NTSTATUS
+NTAPI
+KiSystemCallTrampoline(IN PVOID Handler,
+                       IN PVOID Arguments,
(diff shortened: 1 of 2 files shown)


ntoskrnl/ke/i386/trap.s
```diff
@@ -220,4 +220,39 @@ _KiConvertToGuiThread@0:
...
     ret
 
+/*
+NTSTATUS
+NTAPI
+KiSystemCallTrampoline(IN PVOID Handler,
+                       IN PVOID Arguments,
+                       IN ULONG StackBytes);
+*/
+
```
HBelusca, line 230:
```suggestion
```

HBelusca (approved):

Extravert-ir (approved):

PR #3593, opened 2021-04-08
[WINLOGON] Show "Restaring..." while restarting.

arnavbhatt288:
## Purpose

Instead of just showing "ReactOS is shutting down" while restarting, show "ReactOS is restarting".

## Proposed changes

- Added a new string ID named `IDS_REACTOSISRESTARTING`.
- Now `HandleShutdown()` will determine whether the system is shutting down or restarting and will use required string ID.

# Before:
Screenshot_vm1_2021-04-08_12:39:48

# After:
Screenshot_win2k3-2_2021-04-18_21:18:30


GeoB99 (approved):
You can just squash your two commits into one, there's no point in having an extra commit for the copyright credit thingy. Otherwise the changes look OK to me.

HBelusca:
Technically, in both cases, ReactOS **is** shutting down and doesn't do anything different.
It is your **computer** however, that either restarts or powers off. This distinction can be also seen in the naming of the APIs.

arnavbhatt288:
> Technically, in both cases, ReactOS **is** shutting down and doesn't do anything different.
> It is your **computer** however, that either restarts or powers off. This distinction can be also seen in the naming of the APIs.

How about we write "Your PC is restarting" instead the current sentence or similar?

SergeGautherie:
Does Windows have this?
Maybe something like "ROS shut down. Then, computer will restart."

gonzoMD:
> > Technically, in both cases, ReactOS **is** shutting down and doesn't do anything different.
> > It is your **computer** however, that either restarts or powers off. This distinction can be also seen in the naming of the APIs.
> 
> How about we write "Your PC is restarting" instead the current sentence or similar?

IMO 'shutting down' and 'restarting' should be enough. Users don't really care if 'reactos' or 'computer' is displayed

gonzoMD:
> > Technically, in both cases, ReactOS **is** shutting down and doesn't do anything different.
> > It is your **computer** however, that either restarts or powers off. This distinction can be also seen in the naming of the APIs.
> 
> How about we write "Your PC is restarting" instead the current sentence or similar?

IMO 'shutting down' and 'restarting' should be enough. Users don't really care if 'reactos' or 'computer' is displayed

base/system/winlogon/sas.c
```diff
@@ -1046,7 +1047,12 @@ HandleShutdown(
     BOOLEAN Old;
 
     // SwitchDesktop(Session->WinlogonDesktop);
-    DisplayStatusMessage(Session, Session->WinlogonDesktop, IDS_REACTOSISSHUTTINGDOWN);
+    
+    /* If the system is rebooting, show IDS_REACTOSISRESTARTING string instead */
```
binarymaster, line 1051:
```suggestion
    /* If the system is rebooting, show the according string */
```

andrew-boyarshin: _According_ feels out of place here (according to what? to whom?). Maybe use _appropriate_ instead?

binarymaster: Thanks, corrected

base/system/winlogon/lang/es-ES.rc
```diff
@@ -46,6 +46,7 @@ BEGIN
     IDS_RUNNINGLOGONSCRIPTS "Ejecutando los scripts de inicio de sesión..."
     IDS_LOADINGYOURPERSONALSETTINGS "Cargando su configuración personal..."
     IDS_CLOSINGNETWORKCONNECTIONS "Cerrando las conexiones de red..."
+    IDS_REACTOSISRESTARTING "Restarting..."
     IDS_REACTOSISSHUTTINGDOWN "ReactOS se está apagando..."
```
binarymaster, line 50: Since `IDS_REACTOSISSHUTTINGDOWN` now has new meaning in your PR, you need to set it the same in all translations.

binarymaster (changes requested):

binarymaster (approved):

base/system/winlogon/sas.c
```diff
@@ -1046,7 +1047,12 @@ HandleShutdown(
     BOOLEAN Old;
 
     // SwitchDesktop(Session->WinlogonDesktop);
-    DisplayStatusMessage(Session, Session->WinlogonDesktop, IDS_REACTOSISSHUTTINGDOWN);
+    
+    /* If the system is rebooting, show the appropriate string */
+    if (wlxAction == WLX_SAS_ACTION_SHUTDOWN_REBOOT) 
+        DisplayStatusMessage(Session, Session->WinlogonDesktop, IDS_REACTOSISRESTARTING);
+    else
+        DisplayStatusMessage(Session, Session->WinlogonDesktop, IDS_REACTOSISSHUTTINGDOWN);
```
SergeGautherie, line 1055: Style: Add braces or use `? :`?

HBelusca: It's fine as it is.

base/system/winlogon/lang/es-ES.rc
JoachimHenze: That was a very weird recommendation. It would have been better, if he wouldn't have touched IDS_REACTOSISSHUTTINGDOWN for any localization.

PR #3594, opened 2021-04-08
Fixup of resource sections in GCC builds

disean:
## Proposed changes

- GCC: Make resource sections discardable for kernel-mode drivers and DLLs
- Fix pciide and pcix drivers

JIRA issue: [CORE-17401](https://jira.reactos.org/browse/CORE-17401)

### DIFF
--- a/drivers/bus/pcix/CMakeLists.txt
+++ b/drivers/bus/pcix/CMakeLists.txt
@@ -48,6 +48,14 @@ add_library(pcix MODULE
 
 set_module_type(pcix kernelmodedriver)
 target_link_libraries(pcix arbiter)
+if(MSVC)
+    target_link_options(pcix PRIVATE /SECTION:.rsrc,!D)
+else()
+    add_custom_command(
+        TARGET pcix POST_BUILD
+        COMMAND native-pefixup --section:.rsrc,!D $<TARGET_FILE:pcix>
+        VERBATIM)
+endif()
 add_importlibs(pcix ntoskrnl hal)
 add_pch(pcix pci.h "${PCH_SKIP_SOURCE}")
 add_dependencies(pcix pciclass)
--- a/drivers/storage/ide/pciide/CMakeLists.txt
+++ b/drivers/storage/ide/pciide/CMakeLists.txt
@@ -1,5 +1,13 @@
 
 add_library(pciide MODULE pciide.c pciide.rc)
 set_module_type(pciide kernelmodedriver)
+if(MSVC)
+    target_link_options(pciide PRIVATE /SECTION:.rsrc,!D)
+else()
+    add_custom_command(
+        TARGET pciide POST_BUILD
+        COMMAND native-pefixup --section:.rsrc,!D $<TARGET_FILE:pciide>
+        VERBATIM)
+endif()
 add_importlibs(pciide pciidex ntoskrnl)
 add_cd_file(TARGET pciide DESTINATION reactos/system32/drivers NO_CAB FOR all)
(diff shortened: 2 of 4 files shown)


HBelusca:
No please don't do that. We don't need new CMake module types for each and every particular cases.
The correct fix is to keep what is currently present in master, and just add your resource section override linker flags in **your own** cmakefiles for the given module (pciide and pcix for example).

disean:
Ok, I'll try to improve it.

sdk/tools/pefixup.c
```diff
@@ -144,9 +144,15 @@ static int driver_fixup(int mode, unsigned char *buffer, PIMAGE_NT_HEADERS nt_he
...
-        /* For some reason, .rsrc is made writable by windres */
         if (strncasecmp((char*)Section->Name, ".rsrc", 5) == 0)
         {
+            /* .rsrc is discardable for driver images, WDM drivers, and Kernel-Mode DLLs */
+            if (mode == MODE_KERNELDRIVER || mode == MODE_WDMDRIVER || mode == MODE_KERNELDLL)
+            {
+                Section->Characteristics |= IMAGE_SCN_MEM_DISCARDABLE;
+            }
+
+            /* For some reason, .rsrc is made writable by windres */
```
HBelusca, line 155: Why does this need an extra handling there?

disean: I follow the loop logic "continue if the given section can be paged".

drivers/storage/ide/pciide/CMakeLists.txt
```diff
@@ -1,5 +1,11 @@
 
 add_library(pciide MODULE pciide.c pciide.rc)
 set_module_type(pciide kernelmodedriver)
+if(MSVC)
+    target_link_options(pciide PRIVATE /SECTION:.rsrc,!D)
+else()
+    # Not implemented
```
HBelusca, line 7:
```suggestion
    add_custom_command(
        TARGET pciide POST_BUILD
        COMMAND objcopy --set-section-flags .rsrc=readonly $<TARGET_FILE:pciide>
        VERBATIM)
```
(Note: if "objcopy" doesn't work, try using `${CMAKE_OBJCOPY}` instead.)

drivers/bus/pcix/CMakeLists.txt
```diff
@@ -48,6 +48,11 @@ add_library(pcix MODULE
 
 set_module_type(pcix kernelmodedriver)
 target_link_libraries(pcix arbiter)
+if(MSVC)
+    target_link_options(pcix PRIVATE /SECTION:.rsrc,!D)
+else()
+    # Not implemented
```
HBelusca, line 54:
```suggestion
    add_custom_command(
        TARGET pcix POST_BUILD
        COMMAND objcopy --set-section-flags .rsrc=readonly $<TARGET_FILE:pcix>
        VERBATIM)
```
(Note: if "objcopy" doesn't work, try using `${CMAKE_OBJCOPY}` instead.)

sdk/tools/pefixup.c
HBelusca: My comment was misplaced and was for the handling of the discardable flag.

disean: MSVC linker does that by default.

HBelusca: Oh sorry you were fixing also another problem in the same PR.
Ok then keep it, and adjust the comment there too: https://git.reactos.org/?p=reactos.git;a=blob;f=sdk/cmake/gcc.cmake;hb=2587d72ef7c0673a619a7ad4202a82f926e8a6ab#l300

drivers/storage/ide/pciide/CMakeLists.txt
disean: objcopy is broken
result

HBelusca: huh yeah indeed definitively. It does much more changes (and sometimes harmful ones) than intended. Let's not use it then.

HBelusca:
@disean : Before I merge your PR, could you also have a look at whether the resource section, as a generic rule for kernel drivers, is to be paged or non-paged?

disean:
The resource section is pageable.
By the way, CORE-17401 unrelated to the pcix driver.

PR #3595, opened 2021-04-08
[KMTESTS:IO] Rewrite IoDeviceInterface kmtest CORE-17361

oleg-dubinskiy:
## Purpose

Add new kernel-mode test for device interface functions, based on the patch by Mark Jansen.
The previous one was buggy and crashed even on Windows.
The new one doesn't fail, and works without bugs. Although there is missing one test case (for `IoRegisterDeviceInterface`).
Also perhaps `IoRegister`/`UnregisterPlugPlayNotification` testcase should be kept, if it's also corresponding to device interface functions.
This kmtest helped me to test the behaviour of some IoDeviceInterface* routines (and understand a bit more how they're functioning), whose I already improved in my previous PRs: #3510, #3583 and #3584.

## Proposed changes

Add a new kmtest with testcases for:
- `IoGetDeviceInterfaceAlias`.
- `IoGetDeviceInterfaces`.
- `IoOpenDeviceInterfaceRegistryKey`.
- `IoSetDeviceInterfaceState`.

## TODO

- [x] Maybe restore `IoRegister`/`UnregisterPlugPlayNotification` testcase if it also corresponds to device interface functions.
- [ ] Maybe add the `IoRegisterDeviceInterface` testcase, but I seem to have not enough skills for that (since that function works with device objects and other stuff like that, not only with registry and symbolic links).

## Result

The test in its current state completely succeeds on Windows XP and 2003. :slightly_smiling_face: 

kmtest_XP

### DIFF
--- a/modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
+++ b/modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
@@ -1,11 +1,13 @@
 /*
- * PROJECT:         ReactOS kernel-mode tests
- * LICENSE:         LGPLv2+ - See COPYING.LIB in the top level directory
- * PURPOSE:         Kernel-Mode Test Suite Device Interface functions test
- * PROGRAMMER:      Filip Navara <xnavara@volny.cz>
+ * PROJECT:     ReactOS kernel-mode tests
+ * LICENSE:     LGPL-2.1-or-later (https://spdx.org/licenses/LGPL-2.1-or-later)
+ * PURPOSE:     Test for Device Interface functions
+ * COPYRIGHT:   Copyright 2011 Filip Navara <xnavara@volny.cz>
+ *              Copyright 2021 Mark Jansen <mark.jansen@reactos.org>
+ *              Copyright 2021-2022 Oleg Dubinskiy <oleg.dubinskij2013@yandex.ua>
  */
 
-/* TODO: what's with the prototypes at the top, what's with the if-ed out part? Doesn't process most results */
+/* TODO: Add IoRegisterDeviceInterface testcase */
 
 #include <kmt_test.h>
 #include <poclass.h>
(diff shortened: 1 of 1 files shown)


GeoB99:
>Also perhaps IoRegister/UnregisterPlugPlayNotification testcase should be kept, if it's also corresponding to device interface functions.

You could have just enclosed the respective calls in a `#if 0` macro in `START_TEST` and let someone else to fix the chunks of code responsible for crashing the KMTEST, unless the code is so broken in such a way that a total rewrite is better.

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
```diff
@@ -1,292 +1,221 @@
...
-    Buffer = KmtAllocateGuarded(BufferSize);
-    if (!skip(Buffer != NULL, "Failed to allocate %lu bytes\n", BufferSize))
+    PWSTR SymbolicLinkList, SymbolicLinkListPtr;
+    UNICODE_STRING GuidString;
+
+    RtlStringFromGUID(guid, &GuidString);
+
+    Status = IoGetDeviceInterfaces(guid, NULL, DEVICE_INTERFACE_INCLUDE_NONACTIVE, &SymbolicLinkList);
+
+    ok(NT_SUCCESS(Status), "IoGetDeviceInterfaces failed with status 0x%X for '%wZ'\n", (unsigned int)Status, &GuidString);
```
GeoB99, line 193: What's the point of the type cast here?

oleg-dubinskiy: Oh, looks like I didn't check this part of the code yet, since it was wrote by @learn-more. But I'm sure that this cast is just not needed and can be safely removed.

oleg-dubinskiy:
> > Also perhaps IoRegister/UnregisterPlugPlayNotification testcase should be kept, if it's also corresponding to device interface functions.
> 
> You could have just enclosed the respective calls in a `#if 0` macro in `START_TEST` and let someone else to fix the chunks of code responsible for crashing the KMTEST, unless the code is so broken in such a way that a total rewrite is better.

Indeed, agreed. But looks like I figured out why the previous test has crashed. The failure on Windows most of all causes only an old `IoRegisterDeviceInterface` testcase. After copying and enabling it in the new kmtest, it crashed on Windows too. So looks like this testcase is indeed a source of problem.
So perhaps I can keep `IoRegister`/`UnregisterPlugPlayNotification` one, and it will work without the issues.

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
learn-more: probably gcc crying about crap

oleg-dubinskiy: Thanks. Just removed it, and at least GCC build still doesn't fail for me. So it indeed seems to be not needed.

oleg-dubinskiy: Fixed. :ok_hand:

oleg-dubinskiy:
Yep, an old `IoRegister`/`UnregisterPlugPlayNotification` testcase doesn't cause any failures on Windows and works as well there. So I restored it as well. :slightly_smiling_face:

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
```diff
@@ -1,125 +1,216 @@
...
+
+        NTSTATUS Status = IoOpenDeviceInterfaceRegistryKey(&SymbolicLinkName, KEY_CREATE_SUB_KEY, &DeviceInterfaceKey);
+
+        if (NT_SUCCESS(Status))
+        {
+            DPRINT1("IoOpenDeviceInterfaceRegistryKey(): success: %d %p\n", n, DeviceInterfaceKey);
+        }
+        else
+        {
+            DPRINT1("IoOpenDeviceInterfaceRegistryKey(): fail: %d 0x%x\n", n, Status);
```
GeoB99, line 99: Can't those prints be promoted to `trace()` (same applies for other DPRINT calls in the kernel test code)?

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
```diff
@@ -1,125 +1,216 @@
...
+        }
+
+        InitializeObjectAttributes(
+            &ObjectAttributes,
+            &KeyName,
+            OBJ_CASE_INSENSITIVE | OBJ_KERNEL_HANDLE,
+            DeviceInterfaceKey,
+            NULL);
```
GeoB99, line 107:
```suggestion
        InitializeObjectAttributes(&ObjectAttributes,
                                   &KeyName,
                                   OBJ_CASE_INSENSITIVE | OBJ_KERNEL_HANDLE,
                                   DeviceInterfaceKey,
                                   NULL);
```
Idem for the other call below.

GeoB99 (commented):
Nothing much to review here from my part, just two nit-picks to consider. But otherwise OK from me.

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
oleg-dubinskiy: Yes, they can. But the only difference between `DPRINT` and `trace` is the 2nd one prints the debug output directly in the command prompt instead of COM port. It seems to be more comfortable, but I doubt do we want to completely disable the debug output from the COM.

oleg-dubinskiy: OK, replaced.
If debug output from COM port is not so important, then I'm OK with this change. :slightly_smiling_face:

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
oleg-dubinskiy: Fixed.

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
```diff
@@ -1,125 +1,214 @@
...
-   OUT PWSTR *SymbolicLinkList);
-#endif /* 0 */
-
-static VOID DeviceInterfaceTest_Func()
+/* Predefined GUIDs are required for IoGetDeviceInterfaceAlias and IoOpenDeviceInterfaceRegistryKey.
+ * Only they can provide the aliases and the needed subkeys, unlike manually declared test GUIDs.
+ * Since IoRegisterDeviceInterface testcase is missing, it is not possible to register the new device interface
+ * and get an alias/key handle of it using this test. */
+/* From https://docs.microsoft.com/en-us/windows-hardware/drivers/stream/ksproperty-topology-categories */
+#define STATIC_KSCATEGORY_BRIDGE \
```
ThFabba, line 23: These seem to be defined in `<ks.h>`. Is it possible to include that here instead of duplicating the definitions?

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
```diff
@@ -1,125 +1,214 @@
...
+        else
+        {
+            trace("IoSetDeviceInterfaceState(): failed to enable interface: %d 0x%x\n", n, Status);
+        }
+    }
+}
+
+static
+VOID
+Test_IoGetDeviceInterfaces(const GUID* guid)
```
ThFabba, line 182: Nit: parameter shouldn't start with a lowercase letter.

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
```diff
@@ -1,125 +1,214 @@
...
-    // 1. DeviceNode = NULL
-    DeviceObjectExtension.DeviceNode = NULL;
-    Status = IoRegisterDeviceInterface(&DeviceObject, &Guid, NULL,
-        &SymbolicLinkName);
+    Status = IoGetDeviceInterfaces(guid, NULL, DEVICE_INTERFACE_INCLUDE_NONACTIVE, &SymbolicLinkList);
 
-    ok(Status == STATUS_INVALID_DEVICE_REQUEST,
-        "IoRegisterDeviceInterface returned 0x%08lX\n", Status);
+    ok(NT_SUCCESS(Status), "IoGetDeviceInterfaces failed with status 0x%X for '%wZ'\n", Status, &GuidString);
+    if (!NT_SUCCESS(Status))
```
ThFabba, line 193: You can use `if (skip(NT_SUCCESS(Status), "No device interface list\n"))` to issue a skip message in this case. That helps make it more obvious that parts of the test couldn't be executed.

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
```diff
@@ -1,125 +1,214 @@
...
-    DeviceNode.InstancePath.Length = 0;
-    Status = IoRegisterDeviceInterface(&DeviceObject, &Guid, NULL,
-        &SymbolicLinkName);
+    trace("IoGetDeviceInterfaces '%wZ' results:\n", &GuidString);
+    RtlFreeUnicodeString(&GuidString);
 
-    ok(Status == STATUS_INVALID_DEVICE_REQUEST,
-        "IoRegisterDeviceInterface returned 0x%08lX\n", Status);
+    for (SymbolicLinkListPtr = SymbolicLinkList;
+        SymbolicLinkListPtr[0] != 0 && SymbolicLinkListPtr[1] != 0;
```
ThFabba, line 202: The `SymbolicLinkListPtr[1] != 0` here would make you skip any strings that are only one character long.
While that's not an expected outcome here, it seems unnecessary to provide such an arbitrary limitation.

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
```diff
@@ -1,125 +1,214 @@
...
+VOID
+Test_IoSetDeviceInterfaceState(PWSTR SymbolicLink)
+{
+    UNICODE_STRING SymbolicLinkName;
+
+    RtlInitUnicodeString(&SymbolicLinkName, SymbolicLink);
+
+    for (size_t n = 0; n < RTL_NUMBER_OF(Types); ++n)
+    {
+        NTSTATUS Status = IoSetDeviceInterfaceState(&SymbolicLinkName, TRUE);
```
ThFabba, line 167: You can't just enable a device interface for a device that you don't own. This test has to be done with a device created by the test.

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
```diff
@@ -1,125 +1,214 @@
...
+        {
+            trace("IoOpenDeviceInterfaceRegistryKey(): fail: %d 0x%x\n", n, Status);
+        }
+
+        InitializeObjectAttributes(&ObjectAttributes,
+                                   &KeyName,
+                                   OBJ_CASE_INSENSITIVE | OBJ_KERNEL_HANDLE,
+                                   DeviceInterfaceKey,
+                                   NULL);
+        Status = ZwCreateKey(&DeviceInterfaceSubKey,
```
ThFabba, line 107: What's the purpose of creating a subkey? If you just want to validate that the handle is valid, `ObReferenceObjectByHandle` with the handle and `CmKeyObjectType` will do the job. Or otherwise some read-only operation, like reading the default value or enumerating values/subkeys.

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
```diff
@@ -1,125 +1,214 @@
...
-   ExFreePool(SymbolicLinkList);
+    RtlInitUnicodeString(&SymbolicLinkName, SymbolicLink);
+    RtlInitUnicodeString(&KeyName, L"ReactOS_kmtest");
+
+    for (size_t n = 0; n < RTL_NUMBER_OF(Types); ++n)
+    {
+        HANDLE DeviceInterfaceKey, DeviceInterfaceSubKey;
+        OBJECT_ATTRIBUTES ObjectAttributes;
+
+        NTSTATUS Status = IoOpenDeviceInterfaceRegistryKey(&SymbolicLinkName, KEY_CREATE_SUB_KEY, &DeviceInterfaceKey);
```
ThFabba, line 91: It's worth a comment that, counter to what the documentation says, it's okay to call this from a user process's thread -- since that's what we're doing here.

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
```diff
@@ -1,125 +1,214 @@
...
+
+        if (NT_SUCCESS(Status))
+        {
+            trace("IoOpenDeviceInterfaceRegistryKey(): success: %d %p\n", n, DeviceInterfaceKey);
```
ThFabba, line 95: Why are there only traces and no `ok()`s in this function? Is it normal for the function call to fail on some devices?

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
```diff
@@ -1,125 +1,214 @@
...
+        }
+        else
+        {
+            trace("IoOpenDeviceInterfaceRegistryKey(): fail: %d 0x%x\n", n, Status);
```
ThFabba, line 99: You're missing a `continue` or similar in this else branch -- otherwise you'll be referencing an invalid handle later on.

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
```diff
@@ -1,125 +1,214 @@
...
+    &KSCATEGORY_MIXER,
+    &KSCATEGORY_RENDER,
+    &KSCATEGORY_SPLITTER,
+};
 
-#if 0
-   DPRINT("[PnP Test] Trying to get aliases\n");
+static
+VOID
+Test_IoOpenDeviceInterfaceRegistryKey(PWSTR SymbolicLink)
```
ThFabba, line 79: Nit: Seems like you can use `PCWSTR` in all of these functions

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
```diff
@@ -1,125 +1,214 @@
...
+    RtlInitUnicodeString(&SymbolicLinkName, SymbolicLink);
+
+    for (size_t n = 0; n < RTL_NUMBER_OF(Types); ++n)
+    {
+        UNICODE_STRING AliasSymbolicLinkName;
+        NTSTATUS Status = IoGetDeviceInterfaceAlias(&SymbolicLinkName, Types[n], &AliasSymbolicLinkName);
+
+        if (NT_SUCCESS(Status))
+        {
+            trace("IoGetDeviceInterfaceAlias(): success: %d %wZ\n", n, &AliasSymbolicLinkName);
```
ThFabba, line 146: Like above, you're not testing much here.
Does the call succeed if you pass in the same GUID that was originally passed to `IoGetDeviceInterfaces`? If so, then you can pass the GUID (or the `n`) through to this function, and if they're the same, expect success. And presumably we'd expect the Alias name to be the same string as the original in that case?

If the function call fails, what status do we expect? Presumably `STATUS_OBJECT_NAME_NOT_FOUND` or `STATUS_OBJECT_PATH_NOT_FOUND`?
If you pass in a custom GUID as the second parameter, do you get `STATUS_INVALID_HANDLE` like the documentation seems to suggest?

These are all things you can codify into `ok()`s to make the test actually verify something, instead of just tracing.

ThFabba:
Oh, and yes, the `IoRegisterDeviceInterface` testcase is pretty broken -- it's a completely bogus idea to create a `DEVICE_OBJECT` on the stack. It should use `IoCreateDevice` instead (and even that might not work because these functions expect PDOs, not just any DOs).
I'm okay with deleting that test case.

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
```diff
@@ -1,125 +1,214 @@
...
-         DPRINT("[PnP Test] Alias: %S\n", AliasSymbolicLink.Buffer);
-      }
-   }
-#endif
-
-   ExFreePool(SymbolicLinkList);
+    RtlInitUnicodeString(&SymbolicLinkName, SymbolicLink);
+    RtlInitUnicodeString(&KeyName, L"ReactOS_kmtest");
+
+    for (size_t n = 0; n < RTL_NUMBER_OF(Types); ++n)
```
JoachimHenze, line 86: Declare variable n at top of function already to be C89-compatible, please.

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
```diff
@@ -1,125 +1,214 @@
...
-    DEVICE_OBJECT DeviceObject;
-    EXTENDED_DEVOBJ_EXTENSION DeviceObjectExtension;
-    DEVICE_NODE DeviceNode;
     UNICODE_STRING SymbolicLinkName;
-    NTSTATUS Status;
 
-    RtlInitUnicodeString(&SymbolicLinkName, L"");
+    RtlInitUnicodeString(&SymbolicLinkName, SymbolicLink);
+
+    for (size_t n = 0; n < RTL_NUMBER_OF(Types); ++n)
```
JoachimHenze, line 139: Same here, move declaration of n upwards please.

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
```diff
@@ -1,125 +1,214 @@
...
+
+static
+VOID
+Test_IoSetDeviceInterfaceState(PWSTR SymbolicLink)
+{
+    UNICODE_STRING SymbolicLinkName;
+
+    RtlInitUnicodeString(&SymbolicLinkName, SymbolicLink);
+
+    for (size_t n = 0; n < RTL_NUMBER_OF(Types); ++n)
```
JoachimHenze, line 165: And same here.

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
```diff
@@ -181,112 +270,11 @@ Test_IoRegisterPlugPlayNotification(VOID)
...
-        KmtEndSeh(STATUS_SUCCESS)
-        ok_eq_hex(Status, STATUS_OBJECT_NAME_NOT_FOUND);
-        KmtFreeGuarded(Buffer);
-    }
-
-    RtlInitUnicodeString(&SymbolicLinkName, L"\\??\\#aaaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa}");
-    BufferSize = SymbolicLinkName.Length;
-    Buffer = KmtAllocateGuarded(BufferSize);
-    if (!skip(Buffer != NULL, "Failed to allocate %lu bytes\n", BufferSize))
+    for (size_t n = 0; n < RTL_NUMBER_OF(Types); ++n)
```
JoachimHenze, line 275: same here

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
oleg-dubinskiy: No, it isn't. `ks.h` is not available in kmtests environment, that's the problem. I tried to include it, but then this failed to compile for me. Maybe it is possible to adapt the kmtests for using those headers somehow, but I don't know how. :frowning_face:

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
oleg-dubinskiy: The purpose of the subkey is to check the create options of the parent key: volatile or non-volatile. It was required for creating a proper fix for #3583 PR. If the first call didn't fail, the handle obviously seems to be valid I think. At least in case with `IoOpenDeviceInterfaceRegistryKey`, as said in MS documentation.

ThFabba: Gotcha. You're right, looks like creating a nonvolatile subkey is the only way to verify whether a key is volatile.
Just remember to use `ok()` to actually test something, then.

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
oleg-dubinskiy: Fixed, thanks.

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
oleg-dubinskiy: OK, added.

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
oleg-dubinskiy: Removed.

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
oleg-dubinskiy: OK, added a comment and an `ok()` call.

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
oleg-dubinskiy: Thanks. Added `ok()` macros for all testcases. However, I kept the `NTSTATUS` checks for `continue` and success traces.

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
oleg-dubinskiy: Thanks. Added a `continue` for all failure cases in the loops.

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
oleg-dubinskiy: Done.

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
oleg-dubinskiy: Done. Thanks @JoachimHenze. :+1:

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
oleg-dubinskiy: Done.

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
oleg-dubinskiy: Fixed.

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
oleg-dubinskiy: Fixed.

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
oleg-dubinskiy: So how should I create the device? Maybe you mean device interface? If it's possible only via `IoRegisterDeviceInterface` testcase, then that needs to be implemented too. But I doubt how to implement it properly. In case it's not correct to use this testcase without `IoRegisterDeviceInterface`, then should I disable it for now? :thinking:

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
oleg-dubinskiy: >Does the call succeed if you pass in the same GUID that was originally passed to IoGetDeviceInterfaces? If so, then you can pass the GUID (or the n) through to this function, and if they're the same, expect success. And presumably we'd expect the Alias name to be the same string as the original in that case?

As you might notice, I'm always passing the same list of GUIDs into all functions, independently from `IoGetDeviceInterfaces`. So they should use those GUIDs separately.
What I'm only passing from `IoGetDeviceInterfaces`, is actually **symbolic link**, not a GUID, since MSDN says that the symbolic links should be obtained from `IoGetDeviceInterfaces` or `IoRegisterDeviceInterface` first.
However, I've also added a case which passes alias symbolic links into `IoOpenDeviceInterfaceRegistryKey`, since the documentation says that the symbolic link for that function can be also obtained from `IoGetDeviceInterfaceAlias`. And additionally, I added a NULL GUID to the list of GUIDs, which can be considered as invalid or, as you said, custom. Or custom GUID should be only declared by myself and shouldn't exist anywhere else? :thinking: 
See an updated code for details. :slightly_smiling_face:

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
oleg-dubinskiy: Ah, just noticed your comment below that it can be done via `IoCreateDevice`. So maybe I can implement that, but I'm not sure. :slightly_frowning_face:

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
oleg-dubinskiy: OK, added a comment. Thanks. :slightly_smiling_face:

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
oleg-dubinskiy: >If the function call fails, what status do we expect? Presumably STATUS_OBJECT_NAME_NOT_FOUND or STATUS_OBJECT_PATH_NOT_FOUND?

At least when I'm passing `GUID_NULL`, it fails with 0xc0000034 (`STATUS_OBJECT_NAME_NOT_FOUND`), obviously because it was not found in registry. But the GUID itself seems to be valid (as parameter).
test_results
However I didn't check yet what happens if one of the parameters (either symbolic link or GUID) is really invalid (`NULL`).

oleg-dubinskiy: OK, just added an invalid parameter testcases.
At least when I'm passing an invalid (`NULL`) symbolic link, it returns `STATUS_INVALID_PARAMETER`, not `STATUS_INVALID_HANDLE`. With invalid GUID the same value should be returned I guess, because MSDN says that it returns identical value if one of both parameters is invalid. So I should return the same in my implementation. :slightly_smiling_face: 
invalid_parameter
But I still don't understand why MSDN lies about the return value itself. Maybe it has been changed if the future Windows versions, but why then there is no any note about that then?... :thinking: Although these are related return codes, but nevertheless...

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
HBelusca: What kind of errors do you get?

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
```diff
@@ -1,125 +1,231 @@
...
+    RtlInitUnicodeString(&SymbolicLinkName, SymbolicLink);
+    RtlInitUnicodeString(&KeyName, L"ReactOS_kmtest");
+
+    for (n = 0; n < RTL_NUMBER_OF(Types); ++n)
+    {
+        HANDLE DeviceInterfaceKey, DeviceInterfaceSubKey;
+        OBJECT_ATTRIBUTES ObjectAttributes;
+
+        /* It's okay to call this from a user process's thread */
+        NTSTATUS Status = IoOpenDeviceInterfaceRegistryKey(&SymbolicLinkName, KEY_CREATE_SUB_KEY, &DeviceInterfaceKey);
```
HBelusca, line 98: This function call seems to be independent of the current `Types[n]`˙, so could it be done only once outside the for-loop?

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
```diff
@@ -1,125 +1,231 @@
...
+    {
+        return;
+    }
+
+    trace("IoGetDeviceInterfaces '%wZ' results:\n", &GuidString);
+    RtlFreeUnicodeString(&GuidString);
+
+    for (SymbolicLinkListPtr = SymbolicLinkList;
+        SymbolicLinkListPtr[0] != 0;
+        SymbolicLinkListPtr += wcslen(SymbolicLinkListPtr) + 1)
```
HBelusca, line 220:
```suggestion
    for (SymbolicLinkListPtr = SymbolicLinkList;
         SymbolicLinkListPtr[0] != UNICODE_NULL;
         SymbolicLinkListPtr += wcslen(SymbolicLinkListPtr) + 1)
```

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
```diff
@@ -1,125 +1,231 @@
...
+    &KSCATEGORY_MIXER,
+    &KSCATEGORY_RENDER,
+    &KSCATEGORY_SPLITTER,
+};
 
-#if 0
-   DPRINT("[PnP Test] Trying to get aliases\n");
+static
+VOID
+Test_IoOpenDeviceInterfaceRegistryKey(PCWSTR SymbolicLink)
```
HBelusca, line 84:
```suggestion
Test_IoOpenDeviceInterfaceRegistryKey(
    _In_opt_ PCWSTR SymbolicLink)
```

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
```diff
@@ -1,125 +1,231 @@
...
+
+            ZwClose(DeviceInterfaceKey);
+	    }
+    }
 }
 
 static
 VOID
-Test_IoRegisterDeviceInterface(VOID)
+Test_IoGetDeviceInterfaceAlias(PCWSTR SymbolicLink)
```
HBelusca, line 143:
```suggestion
Test_IoGetDeviceInterfaceAlias(
    _In_opt_ PCWSTR SymbolicLink)
```

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
```diff
@@ -1,125 +1,231 @@
...
+    }
+}
 
-    // 1. DeviceNode = NULL
-    DeviceObjectExtension.DeviceNode = NULL;
-    Status = IoRegisterDeviceInterface(&DeviceObject, &Guid, NULL,
-        &SymbolicLinkName);
+static
+VOID
+Test_IoSetDeviceInterfaceState(PCWSTR SymbolicLink)
```
HBelusca, line 174:
```suggestion
Test_IoSetDeviceInterfaceState(
    _In_opt_ PCWSTR SymbolicLink)
```

modules/rostests/kmtests/ntos_io/IoDeviceInterface.c
```diff
@@ -1,125 +1,231 @@
...
+            trace("IoSetDeviceInterfaceState(): successfully enabled interface: %d %wZ\n", n, &SymbolicLinkName);
+	    }
+    }
+}
 
-    ok(Status == STATUS_INVALID_DEVICE_REQUEST,
-        "IoRegisterDeviceInterface returned 0x%08lX\n", Status);
+static
+VOID
+Test_IoGetDeviceInterfaces(const GUID* Guid)
```
HBelusca, line 199:
```suggestion
Test_IoGetDeviceInterfaces(
    _In_ const GUID* Guid)
```

PR #3596, opened 2021-04-09
Add Visual Studio Static Analysis to github actions

zefklop:
## Purpose

Put ntoskrnl under analysis on a weekly basis.

JIRA issue: [CORE-7444](https://jira.reactos.org/browse/CORE-7444) I guess

## Proposed changes

Run MsBuild with static analysis enabled

## TODO

Fix all the "warning C28251: Inconsistent annotation for 'xxx'" would be a good start
Maybe turn warnings into errors
Later, analyze more than ntoskrnl

### DIFF
--- a/sdk/include/ndk/iofuncs.h
+++ b/sdk/include/ndk/iofuncs.h
@@ -247,6 +247,7 @@ NtLoadDriver(
     _In_ PUNICODE_STRING DriverServiceName
 );
 
+__kernel_entry
 NTSYSCALLAPI
 NTSTATUS
 NTAPI
@@ -292,6 +293,7 @@ NtNotifyChangeDirectoryFile(
     _In_ BOOLEAN WatchTree
 );
 
+__kernel_entry
 NTSYSCALLAPI
 NTSTATUS
 NTAPI
@@ -535,14 +537,15 @@ NtSetIoCompletion(
     _In_ ULONG CompletionInformation
 );
 
+__kernel_entry
 NTSYSCALLAPI
 NTSTATUS
 NTAPI
 NtSetQuotaInformationFile(
-    HANDLE FileHandle,
-    PIO_STATUS_BLOCK IoStatusBlock,
-    PVOID Buffer,
-    ULONG BufferLength
+    _In_ HANDLE FileHandle,
+    _Out_ PIO_STATUS_BLOCK IoStatusBlock,
+    _In_reads_bytes_(BufferLength) PVOID Buffer,
+    _In_ ULONG BufferLength
 );
 
 __kernel_entry
@@ -703,6 +706,7 @@ ZwDeleteBootEntry(
     _In_ PUNICODE_STRING EntryValue
 );
 
+_IRQL_requires_max_(PASSIVE_LEVEL)
 NTSYSAPI
 NTSTATUS
 NTAPI
@@ -798,6 +802,7 @@ ZwNotifyChangeDirectoryFile(
     _In_ BOOLEAN WatchTree
 );
 
+_IRQL_requires_max_(PASSIVE_LEVEL)
 NTSYSAPI
 NTSTATUS
 NTAPI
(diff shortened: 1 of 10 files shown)


sdk/include/ndk/iofuncs.h
```diff
@@ -247,6 +247,7 @@ NtLoadDriver(
     _In_ PUNICODE_STRING DriverServiceName
 );
 
+__kernel_entry
```
HBelusca, line 250: Would be nice if somehow this was part of the definition of `NTSYSCALLAPI`.

sdk/include/xdk/ntdef.template.h
```diff
@@ -117,9 +117,6 @@ extern "C" {
 $define(UCHAR=UCHAR)
 $include(ntbasedef.h)
 
-typedef _Return_type_success_(return >= 0) LONG NTSTATUS;
-typedef NTSTATUS *PNTSTATUS;
```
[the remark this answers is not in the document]
zefklop: yes

sdk/include/ndk/iofuncs.h
zefklop: I'm not sure. This would probably require including specstrings.h before defining it, and changing include order always lead to too much complication

tkreuzer:
Is there a reason to use msbuild? Does that help to make /analyze any better? Because in my experience it's pretty pointless, as it fails to detect even the most simple errors and instead just creates a lot of noise.

zefklop:
> Is there a reason to use msbuild? Does that help to make /analyze any better? Because in my experience it's pretty pointless, as it fails to detect even the most simple errors and instead just creates a lot of noise.

Well, this lets msbuild put the right compile flag for our projects, without any other effort than putting '/p:RunCodeAnalysis=true' '/p:CodeAnalysisRuleSet=NativeRecommendedRules.ruleset' at the end of the build command, instead of tinkering with cmake options that nobody will be willing to maintain.

What kind of noise did you end up having ? If you look closely, the script produces parses the log & provides a CSV file of the detected warnings that you can download as an artefact.

tkreuzer:
> 
> Well, this lets msbuild put the right compile flag for our projects, without any other effort than putting '/p:RunCodeAnalysis=true' '/p:CodeAnalysisRuleSet=NativeRecommendedRules.ruleset' at the end of the build command, instead of tinkering with cmake options that nobody will be willing to maintain.

We already support this since a long time by adding `-D_VS_ANALYSIS_=1` to the configure command. And nobody maintained anything :) We don't specify a ruleset, so it should use all rules, but I'm not 100 sure about that. We could add (a) custom ruleset(s), if we want though.

> 
> What kind of noise did you end up having ? If you look closely, the script parses the log & provides a CSV file of the detected warnings that you can download as an artefact.

Generally my experience was, that it's great at detecting inconsistent annotations, and that's pretty much it. It fails to detect the most simple buffer overruns, or IRQL violations, but instead gives you a bunch of false positives (which is partly due to broken annotation sin MS DDK/SDK headers).
Here's a bunch of broken functions, that do not generate any warnings:
```
int FooBar1(_In_ _Notnull_ int* Ptr) noexcept { *Ptr = 0; return 0; } // writing to _In_ pointer
int FooBar2(_In_opt_ int* Ptr) noexcept { if (Ptr != 0) *Ptr = 0; return 0; } // at least it verifies the _opt_ part here
int FooBar3(_Out_bytecap_(1) int* Ptr) { *Ptr = 0; return 0; } // bytecap is too small? don't care.
int FooBar4(_In_bytecount_(1) int* Ptr) noexcept { Ptr[42] = 7; return 0; } // Yeah, looks legit.
int FooBar5(_Out_bytecap_(Count) int* Ptr, int Count) { (void)Count; Ptr[0] = 7; return 0; } // No need to check anything
int FooBar6(_Out_bytecap_(Count) int* Ptr, int Count) { (void)Count; Ptr[42] = 7; return 0; } // this is also fine

```

PR #3597, opened 2021-04-10
[CMAKE][CRT] Fix a few minor copypastas

SergeGautherie:
Be explicit...

### DIFF
--- a/sdk/lib/crt/math/exp2.c
+++ b/sdk/lib/crt/math/exp2.c
@@ -8,6 +8,6 @@ exp2(
     _In_ double x)
 {
     /* This below avoids clang to optimize our pow call to exp2 */
-    static const TWO = 2.0;
+    static const double TWO = 2.0;
     return pow(TWO, x);
 }
--- a/sdk/cmake/config.cmake
+++ b/sdk/cmake/config.cmake
@@ -47,7 +47,7 @@ elseif(CMAKE_C_COMPILER_ID STREQUAL "Clang")
         set(GCC TRUE CACHE BOOL "The compiler is GCC")
     endif()
     set(CLANG TRUE CACHE BOOL "The compiler is LLVM Clang")
-elseif(MSVC) # aka CMAKE_C_COMPILER_ID STEQUAL "MSVC"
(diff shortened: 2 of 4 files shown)


sdk/lib/crt/math/exp2.c
```diff
@@ -8,6 +8,6 @@ exp2(
     _In_ double x)
 {
     /* This below avoids clang to optimize our pow call to exp2 */
-    static const TWO = 2.0;
+    static const double TWO = 2.0;
```
learn-more, line 11: Does this still avoid clang to optimize the call?
Why did you change this?

ThFabba: To be honest, I don't know how the old version even compiles. It's not like types are optional in C.
(but yes, please verify that the generated assembly code is the same on all compilers)

ThFabba: Spoiler alert: this change does change the generated code.
To be fair, the new code looks better... but the gist is, this needs a test.

HBelusca: missing type?

learn-more (dismissed):
Please show that the functional change is correct, and does not break what the workaround aimed to fix.

sdk/lib/crt/math/exp2.c
zefklop: Oh, my bad :-( I don't even know how I could have missed that.

zefklop:
> Please show that the functional change is correct, and does not break what the workaround aimed to fix.

The fact that clang build is not broken is already a good sign in itself. This on can go in, as far as I'm concerned.

zefklop (approved):

sdk/lib/crt/math/exp2.c
zefklop: > Oh, my bad :-( I don't even know how I could have missed that.

-wImplicit is enabled by -Wall in both GCC & clang builds. But only clang actually uses this file, and -Werror is disabled here, that's how things slip through.

binarymaster (changes requested):
Please resolve conflicting file.

image

PR #3598, opened 2021-04-11
PEFIXUP: Improvements and add support for overriding section attributes

HBelusca:
## Purpose

1. Some improvements:
   - Only recalculate the PE checksum if the image is not from a reproducible build (with original checksum == 0).
   - Section names are case-sensitive, so we can just use strncmp() instead of the non-standard strncasecmp() for names comparisons.
   - Avoid multi-level nesting of code with the error handling done at the very end. Use size_t variables for file sizes.

2. A fix:
   - Use vfprintf() with the va_args list.

3. The new functionality:
Add support for overriding the attributes of specific PE sections.
Use a syntax identical to the one from the EDITBIN tool:
`--section:name[=newname][,[[!]{CDEIKOMPRSUW}][A{1248PTSX}]]`
https://docs.microsoft.com/en-us/cpp/build/reference/section-editbin?view=msvc-160

Should also help with PR #3594

### DIFF
--- a/sdk/tools/pefixup.c
+++ b/sdk/tools/pefixup.c
@@ -26,6 +26,7 @@ static const char* g_Target;
 
 enum fixup_mode
 {
+    MODE_NONE = 0,
     MODE_LOADCONFIG,
     MODE_KERNELDRIVER,
     MODE_WDMDRIVER,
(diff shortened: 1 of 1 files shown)


learn-more:
#3594 already implements this partially

disean:
Ok, rebased.

PR #3599, opened 2021-04-11
[HHPCOMP] Add ARM64/AARCH64 types

konradybcio:
These were missing.. not anymore!

Tested on a PinePhone of all things..

Big hack that allows it to build (on Linux for Linux.. arm-windows-winpe in llvm when!?):

`
mkdir build
cd build
cmake .. -DARCH=arm -DWINARCH=arm -DCMAKE_CXX_COMPILER=g++ -DCMAKE_C_COMPILER=gcc -DCMAKE_C_COMPILER_TARGET=aarch64-linux-eabi -DCMAKE_CXX_COMPILER_TARGET=aarch64-linux-eabi -DCMAKE_C_COMPILER_FORCED=TRUE -DCMAKE_CXX_COMPILER_FORCED=TRUE -DLD_EXECUTABLE=ld
make -j4 -k
`

The apps that managed to build do run on linux/arm64.

### DIFF
--- a/sdk/tools/hhpcomp/chmc/chm.h
+++ b/sdk/tools/hhpcomp/chmc/chm.h
@@ -59,9 +59,11 @@ typedef unsigned long           UInt32;
 typedef long long               Int64;
 typedef unsigned long long      UInt64;
 
-/* x86-64 */
+/* x86-64        */
+/* IA-64         */
+/* ARM64/AArch64 */
 /* Note that these may be appropriate for other 64-bit machines. */
-#elif __x86_64__ || __ia64__
+#elif __x86_64__ || __ia64__ || __aarch64__
 typedef unsigned char           UChar;
 typedef short                   Int16;
 typedef unsigned short          UInt16;


binarymaster:
I think it would be nice to document this way of building on [ARM Port](https://reactos.org/wiki/ARM_Port) and [AArch64](https://reactos.org/wiki/AArch64) wiki pages.

sdk/tools/hhpcomp/chmc/chm.h
```diff
@@ -60,8 +60,9 @@ typedef long long               Int64;
 typedef unsigned long long      UInt64;
 
 /* x86-64 */
+/* arm64/aarch64 */
 /* Note that these may be appropriate for other 64-bit machines. */
```
binarymaster, line 64: Suggestion: either leave only `/* x86-64 */` here, or update it with all architectures. I'd vote for leaving only first one here, as the comment below explains "these may be appropriate for other 64-bit machines".

learn-more:
`-DCMAKE_C_COMPILER_TARGET=aarch64-linux-eabi -DCMAKE_CXX_COMPILER_TARGET=aarch64-linux-eabi`
You are instructing cmake to create linux binaries?

learn-more:
> `-DCMAKE_C_COMPILER_TARGET=aarch64-linux-eabi -DCMAKE_CXX_COMPILER_TARGET=aarch64-linux-eabi`
> You are instructing cmake to create linux binaries?

Ah, for a host tool.
This is not really going to scale, something like #3602 but then for linux

SergeGautherie:
> CORE-17518

Note that, while the ARM64 goal is the same, this PR targets non-MSVC builds, so actually separate.
Furthermore, it looks like the related defines will be wrong for gcc/clang x64 builds on Windows (not Linux)...

PR #3603, opened 2021-04-12
[FRAMEDYN] Only compile x86 ASM on x86

konradybcio:
ARM compilers do not like this.

### DIFF
--- a/dll/win32/framedyn/CMakeLists.txt
+++ b/dll/win32/framedyn/CMakeLists.txt
@@ -14,7 +14,7 @@ set_module_type(framedyn win32dll UNICODE)
 add_importlibs(framedyn oleaut32 msvcrt kernel32 ntdll)
 add_cd_file(TARGET framedyn DESTINATION reactos/system32/wbem FOR all)
 
-if(NOT MSVC)
+if((NOT MSVC) AND (ARCH STREQUAL "i386"))
     add_asm_files(framedynex_asm alias_gcc.s)
     add_library(framedynex ${framedynex_asm})
     set_target_properties(framedynex PROPERTIES LINKER_LANGUAGE ASM)


dll/win32/framedyn/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ set_module_type(framedyn win32dll UNICODE)
 add_importlibs(framedyn oleaut32 msvcrt kernel32 ntdll)
 add_cd_file(TARGET framedyn DESTINATION reactos/system32/wbem FOR all)
 
-if(NOT MSVC)
+if((NOT MSVC) AND (ARCH STREQUAL "i386"))
     add_asm_files(framedynex_asm alias_gcc.s)
     add_library(framedynex ${framedynex_asm})
     set_target_properties(framedynex PROPERTIES LINKER_LANGUAGE ASM)
```
binarymaster, line 20: I think this will be more correct, as this assembler file should probably be used for both x86 and amd64 architectures:
```suggestion
if(NOT MSVC)
    if (ARCH STREQUAL "i386" OR ARCH STREQUAL "amd64")
        add_asm_files(framedynex_asm alias_gcc.s)
        add_library(framedynex ${framedynex_asm})
        set_target_properties(framedynex PROPERTIES LINKER_LANGUAGE ASM)
    endif()
```
(would be nice if somebody else could confirm it)

learn-more: The main target of this file is to create mangled names, so a correct(er) approach would be to add an arm-version of this as well. (Possibly just change the jmp in the existing file?)

binarymaster:
> so a correct(er) approach would be to add an arm-version of this as well. (Possibly just change the jmp in the existing file?)

Agree with this idea. So @konradybcio you need to change [alias_gcc.s](https://github.com/reactos/reactos/blob/master/dll/win32/framedyn/alias_gcc.s) file to something like this:
```asm
#if (defined(_M_IX86) || defined(_M_AMD64))

#include <asm.inc>

.code
.align 4

MACRO(DEFINE_ALIAS, alias, alias2, origstring)
PUBLIC alias
.weakref alias2, origstring
alias:
    jmp alias2
ENDM

#elif defined(_M_ARM)

#include <kxarm.h>

... snip (your ARM code here) ...

#else
#error "Unsupported platform."
#endif
```
I took [oldnames-msvcrt.S](https://github.com/reactos/reactos/blob/master/sdk/lib/crt/oldnames-msvcrt.S) as an example.

binarymaster:
It seems this PR is actually superseded with PR #3636 and 061c7ecb1ae9a752bf84ed039765a43f5c1b3abe commit by @zefklop.

PR #3605, opened 2021-04-12
[NTOS:SE][NTOS:PS] Ensure that client impersonation is possible

GeoB99:
`PsImpersonateClient` blindly impersonates the requested client even though it doesn't know if the actual token given to the call can be impersonated for the thread of the client which we are going to begin impersonation. 

In the case where impersonation is not possible, make a copy of the given token and assign the newly one for impersonation instead.

**JIRA Issue:** [CORE-17539](https://jira.reactos.org/browse/CORE-17539)

### DIFF
--- a/ntoskrnl/ps/security.c
+++ b/ntoskrnl/ps/security.c
@@ -614,8 +614,10 @@ PsImpersonateClient(IN PETHREAD Thread,
                     IN SECURITY_IMPERSONATION_LEVEL ImpersonationLevel)
 {
     PPS_IMPERSONATION_INFORMATION Impersonation, OldData;
-    PTOKEN OldToken = NULL;
+    PTOKEN OldToken = NULL, ProcessToken = NULL;
+    PACCESS_TOKEN NewToken, ImpersonationToken;
     PEJOB Job;
+    NTSTATUS Status;
 
     PAGED_CODE();
     PSTRACE(PS_SECURITY_DEBUG, "Thread: %p, Token: %p\n", Thread, Token);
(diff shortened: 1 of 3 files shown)


ntoskrnl/se/token.c
```diff
@@ -81,6 +81,32 @@ static const INFORMATION_CLASS_INFO SeTokenInformationClass[] = {
...
+ * @param[in] SecondTokenSource
+ * The second token source.
+ *
+ * @return
+ * Returns TRUE if both token sources are equal amongst
+ * their properties, FALSE otherwise.
+ */
+static
+BOOLEAN
+SepEqualTokenSource(
```
Extravert-ir, line 100: Well, I'd say not all functions need descriptions. This one is just wasting 13 lines of space :)

GeoB99: So you're just saying I can totally remove the documentation header for this function altogether?

ntoskrnl/se/token.c
```diff
@@ -2152,6 +2152,104 @@ SeTokenIsWriteRestricted(IN PACCESS_TOKEN Token)
...
+ * @param[in] ImpersonationLevel
+ * Security impersonation level grade.
+ *
+ * @return
+ * Returns TRUE if the conditions checked are met for token impersonation,
+ * FALSE otherwise.
+ */
+BOOLEAN
+NTAPI
+SeTokenCanImpersonate(
```
Extravert-ir, line 2176: This method is exported?

ntoskrnl/ps/security.c
```diff
@@ -670,22 +673,55 @@ PsImpersonateClient(IN PETHREAD Thread,
...
+         * token to the impersonation information.
+         */
+        TokenImpersonation = Token;
+
+        /* Obtain a token from the process */
+        ProcessToken = PsReferencePrimaryToken(Thread->ThreadsProcess);
+
+        /* Make sure we can impersonate */
+        CanImpersonate = SeTokenCanImpersonate(ProcessToken, Token, ImpersonationLevel);
+        if (!CanImpersonate)
```
Extravert-ir, line 690: Don't be afraid of putting boolean functions inside `if()` operator

ntoskrnl/se/token.c
GeoB99: No, the function is private and only in use for the kernel. PE Explorer demonstrates that.
Capture

ntoskrnl/se/token.c
```diff
@@ -2265,6 +2265,104 @@ SeTokenIsWriteRestricted(IN PACCESS_TOKEN Token)
...
+    /*
+     * Make sure the tokens aren't diverged in terms of
+     * restrictions, that is, the token is restricted
+     * but the other one isn't.
+     */
+    if (SeTokenIsRestricted(ProcessToken) !=
+        SeTokenIsRestricted(TokenToImpersonate))
+    {
+        /*
+         * At least one token is restricted so we cannot
```
ThFabba, line 2346: Actually, _exactly_ one token is restricted here.

ntoskrnl/se/token.c
```diff
@@ -2265,6 +2265,104 @@ SeTokenIsWriteRestricted(IN PACCESS_TOKEN Token)
...
+    {
+        /* They aren't, bail out */
+        DPRINT("SeTokenCanImpersonate(): Tokens SIDs are not equal!\n");
+        CanImpersonate = FALSE;
+        goto Quit;
+    }
+
+    /*
+     * Make sure the tokens aren't diverged in terms of
+     * restrictions, that is, the token is restricted
```
ThFabba, line 2339: `s/the/one/`?

ntoskrnl/ps/security.c
```diff
@@ -670,22 +672,61 @@ PsImpersonateClient(IN PETHREAD Thread,
...
 
-        /* FIXME: If the process token can't impersonate, we need to make a copy instead */
+        /*
+         * Assign the token we get from the caller first. The reason
+         * we have to do that is because we're unsure if we can impersonate
+         * in the first place. In the scenario where we cannot then the
+         * last resort is to make a copy of the token and assign that newly
+         * token to the impersonation information.
+         */
+        TokenImpersonation = Token;
```
ThFabba, line 682: Would `ImpersonationToken` be a better name? `TokenImpersonation` appears to be an enum member.

PR #3537, opened 2021-03-16
[HALX86] Sort out MP HALs, Rearrange files

SergeGautherie:
## Purpose

Additional steps forward, per @vgalnt's comments.

JIRA issue: [CORE-17378](https://jira.reactos.org/browse/CORE-17378)

## Proposed changes

- [HALX86] CMakeLists.txt: Sort out MP HALs
Addendum to a064c5d.
- [HALX86] Rearrange files
Addendum to 7601fb5.
- [HALX86] generic/spinlock.c: Make #ifdef spans more explicit

PR #3611, opened 2021-04-14
[HALX86] generic/spinlock.c: Make #ifdef spans more explicit

SergeGautherie:
Especially, properly mark unreachable code.

Split off from #3537.


Extravert-ir:
When `CONFIG_SMP` is implemented, these ifdefs will be removed anyway

PR #3612, opened 2021-04-15
[CMAKE] Eliminate the use of GCC and CLANG variables

Extravert-ir:
Now, these should be used:
MSVC - for compilers with MSVC-like API
NOT MSVC - for compilers with GNU-like API

CMAKE_C_COMPILER_ID for everything else


dll/3rdparty/mbedtls/CMakeLists.txt
```diff
@@ -95,6 +95,8 @@ add_cd_file(TARGET mbedtls DESTINATION reactos/system32 FOR all)
 
 if(NOT MSVC)
     target_compile_options(mbedtls PRIVATE -Wno-pointer-sign -Wno-unused-function)
-elseif(USE_CLANG_CL)
+endif()
+
+if(MAKE_C_COMPILER_ID STREQUAL "Clang")
```
andrew-boyarshin, line 100: CMAKE?

base/services/nfsd/CMakeLists.txt
```diff
@@ -48,7 +48,7 @@ list(APPEND SOURCE
 
 add_executable(nfsd ${SOURCE} nfsd.rc)
 
-if(MSVC AND (NOT USE_CLANG_CL))
+if(CMAKE_C_COMPILER_ID STREQUAL "MSVC")
```
learn-more, line 51: strequal is case sensitive, is this always uppercase?

learn-more: seems so.

learn-more:
What do we gain by removing those variables and resorting to STREQUAL for the checks?

dll/3rdparty/mbedtls/CMakeLists.txt
Extravert-ir: Thanks!

Extravert-ir:
These variables were confusing:
`MSVC` is a cmake built-in variable, and is set for MSVC and clang-cl
`GCC` is set for GCC and clang in GCC mode
`CLANG` is set for clang and clang-cl

So `if(GCC)` does not mean the condition is true only for GCC, which looks weird (for me at least)

I thought about introducing something like `COMPILER_API_GNU` and `COMPILER_API_MSVC`, but for our set of compilers it would basically set to `NOT MSVC` and `MSVC` respectively

Extravert-ir:
The usage of `USE_CLANG_CL` should also be reduced btw

sdk/include/xdk/ntbasedef.h
```diff
@@ -788,7 +788,11 @@ extern "C++" { \
 #endif /* _M_AMD64 */
 
 /* C_ASSERT Definition */
-#define C_ASSERT(expr) extern char (*c_assert(void)) [(expr) ? 1 : -1]
+#ifndef _Static_assert
+ #define C_ASSERT(expr) typedef char c_assert [(expr) ? 1 : -1]
+#else
+ #define C_ASSERT(expr) _Static_assert(expr, "(" #expr ") failed")
+#endif
```
HBelusca, line 795: Why is this a problem with Clang? Does it fail to compile even if the static assertion is true?

Extravert-ir: As far as I understand, it just doesn't like how we define our `C_ASSERT`

sdk/tools/xml2sdb/xml2sdb.cpp
```diff
@@ -16,18 +16,14 @@ using tinyxml2::XMLText;
...
-C_ASSERT(sizeof(wchar_t) == 2);
-C_ASSERT(sizeof(TAG) == 2);
-C_ASSERT(sizeof(TAGID) == 4);
+static_assert(sizeof(GUID) == 16, "GUID has wrong size");
+static_assert(sizeof(ULONG) == 4, "ULONG has wrong size");
+static_assert(sizeof(LARGE_INTEGER) == 8, "LARGE_INTEGER has wrong size");
+static_assert(sizeof(WCHAR) == 2, "WCHAR has wrong size");
+static_assert(sizeof(wchar_t) == 2, "wchar_t has wrong size");
+static_assert(sizeof(TAG) == 2, "TAG has wrong size");
+static_assert(sizeof(TAGID) == 4, "TAGID has wrong size");
```
HBelusca, line 26: That "has wrong size" thing repeated over & over is much copy-pasting. Maybe, for host-tools, we could have one centralized definition of C_ASSERT, exactly as done as in the XDK, and use that everywhere; that would save the copypaste effort you've done there.

sdk/include/xdk/ntbasedef.h
HBelusca: The way we do it is the same as the MS PSDK or something like that 🤔

PR #3615, opened 2021-04-16
Fix a few MSVC amd64 C4334 warnings

SergeGautherie:
Trivial.

'...(...): warning C4334: '<<': result of 32-bit shift implicitly converted to 64 bits (was 64-bit shift intended?)'

### DIFF
--- a/sdk/lib/rtl/compress.c
+++ b/sdk/lib/rtl/compress.c
@@ -59,7 +59,15 @@ static PUCHAR lznt1_decompress_chunk(UCHAR *dst, ULONG dst_size, UCHAR *src, ULO
 
                 /* find length / displacement bits */
                 for (displacement_bits = 12; displacement_bits > 4; displacement_bits--)
+#ifndef __REACTOS__
                     if ((1 << (displacement_bits - 1)) < dst_cur - dst) break;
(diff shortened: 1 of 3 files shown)


sdk/lib/rtl/compress.c
```diff
@@ -59,7 +59,10 @@ static PUCHAR lznt1_decompress_chunk(UCHAR *dst, ULONG dst_size, UCHAR *src, ULO
 
                 /* find length / displacement bits */
                 for (displacement_bits = 12; displacement_bits > 4; displacement_bits--)
-                    if ((1 << (displacement_bits - 1)) < dst_cur - dst) break;
+                {
```
ThFabba, line 62: This appears to be unmodified Wine code, so please keep it in sync.

SergeGautherie: Submitted to Wine, stalled there...

tkreuzer: It's not a bug, it's a false positive. We can silence it with a compiler option. Please revert this part.

tkreuzer (changes requested):
- [ ] Revert change to wine code

PR #3618, opened 2021-04-19
[MSGINA] Determine which kind of dialog box is before terminating it

GeoB99:
Do a sanity check onto the dialog box type before terminating it. As it stands now, a call of `PostQuitMessage` is invoked even when the dialog box is modal. This is illegal due to the fact that the shutdown dialog box is initiated by the "Security" main window thus `WlxDialogBoxParam`.

A call to `PostQuitMessage` onto a modal dialog box leads to a undefined behaviour, as it'll not just terminate the thread but also eventually killing the Winlogon process whose the thread belongs to. This fixes [CORE-17535](https://jira.reactos.org/browse/CORE-17535).

### DIFF
--- a/dll/win32/msgina/shutdown.c
+++ b/dll/win32/msgina/shutdown.c
@@ -61,6 +61,7 @@ typedef struct _SHUTDOWN_DLG_CONTEXT
     BOOL bReasonUI;
     BOOL bFriendlyUI;
     BOOL bIsButtonHot[NUMBER_OF_BUTTONS];
+    BOOL bIsDialogModal;
     WNDPROC OldButtonProc;
 } SHUTDOWN_DLG_CONTEXT, *PSHUTDOWN_DLG_CONTEXT;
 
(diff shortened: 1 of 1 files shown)


HBelusca:
My concern about that (the PostQuitMessage that Arnav introduced) is that, before his PR, they weren't needed and things used to work. So my question there is, why are these PostQuitMessage now sometimes still needed?

GeoB99:
It's very simple, as you can see the screenshot which showcases the specific chunk of code:
Capture
Before the Arnav's PR, the dialog box was always created as modal regardless of the condition if it was met or not (is Winlogon present in the context, are the Winlogon functions present, etc etc), `DialogBoxParamW` was used instead of `CreateDialogParamW`.

Now with the introduction of the shutdown fancy dialog, such dialog is created by `CreateDialogParamW` which is a modeless dialog box. The system handles modeless and modal dialog boxes differently and thus we've to use different cleanup methods as well. Therefore in the case where a modeless shutdown dialog box is created we must use the combination of `DestroyWindow` and `PostQuitMessage` calls.

PR #3621, opened 2021-04-23
[SHELL32] Show Product version on About dialog bmp

jeffbox12:
## Purpose

The About dialog banner will now show the correspondent Product version selected on install.
This will make it easier to tell which edition of ReactOS is installed.

## Proposed changes

- Changes all banners with a new design. (msgina, syssetup & shell32)
- The About dialog will now show the Product version, if it's Workstation or Server.
  _Only the About dialog will show the Product version for now,_
- Includes svg file for easier year modification. (rosbanner.svg)

## Preview

When Workstation is selected:
workstation

When Server is selected:
server


milawynsrealm:
The color scheme is great. However, the text for ReactOS and the product type (Workstation, or server) is visually disconnected and separate. This is passable as is, but it would look better if they were closer together. You don't have to change the font size, but at least look at the logo banner used for Windows XP as a reference (like shown in the winver.exe tool). If the two-line approach doesn't work, see if it works on a single line.

dll/win32/shell32/wine/shell32_main.c
```diff
@@ -1135,6 +1135,47 @@ INT_PTR CALLBACK AboutAuthorsDlgProc( HWND hWnd, UINT msg, WPARAM wParam, LPARAM
...
+                            &dwSize);
+    RegCloseKey(hKey);
+
+    if (lRet != ERROR_SUCCESS || dwType != REG_DWORD || dwValue != 0x300)
+    {
+        // Allow only on Workstation
+        return FALSE;
+    }
+
+    if (lRet != ERROR_SUCCESS || dwType != REG_DWORD)
```
peterooch, line 1173: The conditions are already checked in the `if`  above.

dll/win32/shell32/wine/shell32_main.c
```diff
@@ -1135,6 +1135,47 @@ INT_PTR CALLBACK AboutAuthorsDlgProc( HWND hWnd, UINT msg, WPARAM wParam, LPARAM
...
+    lRet = RegOpenKeyExW(HKEY_LOCAL_MACHINE,
+                         L"SYSTEM\\CurrentControlSet\\Control\\Windows",
+                         0,
+                         KEY_QUERY_VALUE,
+                         &hKey);
+    if (lRet != ERROR_SUCCESS)
+        return FALSE;
+
+    // Check product version number
+    dwValue = 0;
```
GeoB99, line 1157: What's the point of assigning `dwValue` to 0 again when it's already assigned before?

dll/win32/shell32/wine/shell32_main.c
```diff
@@ -1159,6 +1200,18 @@ static INT_PTR CALLBACK AboutDlgProc( HWND hWnd, UINT msg, WPARAM wParam, LPARAM
 
                 // Preload the ROS bitmap
                 hLogoBmp = (HBITMAP)LoadImage(shell32_hInstance, MAKEINTRESOURCE(IDB_REACTOS), IMAGE_BITMAP, 0, 0, LR_DEFAULTCOLOR);
+#ifdef __REACTOS__
+                if(IsVersionWorkstation() == TRUE)
```
GeoB99, line 1204:
```suggestion
                if (IsVersionWorkstation())
```

dll/win32/shell32/wine/shell32_main.c
```diff
@@ -1159,6 +1200,18 @@ static INT_PTR CALLBACK AboutDlgProc( HWND hWnd, UINT msg, WPARAM wParam, LPARAM
...
+                {
+                   // Load Workstation Bitmap
+                   hLogoBmp = (HBITMAP)LoadImage(shell32_hInstance, MAKEINTRESOURCE(IDB_REACTOS_WORKSTATION), IMAGE_BITMAP, 0, 0, LR_DEFAULTCOLOR);
+                }
+                else
+                {
+                   // Load Server Bitmap
+                   hLogoBmp = (HBITMAP)LoadImage(shell32_hInstance, MAKEINTRESOURCE(IDB_REACTOS_SERVER), IMAGE_BITMAP, 0, 0, LR_DEFAULTCOLOR);
+                }
+#endif
```
GeoB99, line 1214: At the code above the ReactOS logo bitmap is loaded by the `LoadImage` call yet after that you're discarding and overwriting it with a new one based on Workstation/Server setup condition. This is a waste of time, why don't you just load the logo bitmap based on the condition alone? Also why is this chunk of code wrapped within `__REACTOS__` (same as the other one)?

dll/win32/shell32/wine/shell32_main.c
```diff
@@ -1135,6 +1135,47 @@ INT_PTR CALLBACK AboutAuthorsDlgProc( HWND hWnd, UINT msg, WPARAM wParam, LPARAM
 /*************************************************************************
  * AboutDlgProc            (internal)
  */
+#ifdef __REACTOS__
```
GeoB99, line 1138: Ditto.

dll/win32/shell32/wine/shell32_main.c
```diff
@@ -1135,6 +1135,47 @@ INT_PTR CALLBACK AboutAuthorsDlgProc( HWND hWnd, UINT msg, WPARAM wParam, LPARAM
...
+    dwSize = sizeof(dwValue);
+    lRet = RegQueryValueExW(hKey,
+                            L"CSDVersion",
+                            NULL,
+                            &dwType,
+                            (LPBYTE)&dwValue,
+                            &dwSize);
+    RegCloseKey(hKey);
+
+    if (lRet != ERROR_SUCCESS || dwType != REG_DWORD || dwValue != 0x300)
```
GeoB99, line 1167: Not of a big deal but having a defined macro for `0x300` which explicitly tells its purpose would be better rather than passing such magic value as is.

GeoB99 (changes requested):
I personally find these bitmap logos looking great, although some stuff have to be addressed in code.

dll/win32/shell32/wine/shell32_main.c
jeffbox12: "why is this chunk of code wrapped within __REACTOS__"

Because this is a file from Wine, which you aren't supposed to modify. I was told that when adding new code to it, it should be wrapped with it. You can see more of these throughout the code.

"you are discarding and overwriting the code above with a new one, why don't you just load the logo bitmap based on the condition alone?"

To preserve the original Wine code. I wasn't sure what to do with it.

jeffbox12: I'll leave it commented for now.

dll/win32/shell32/wine/shell32_main.c
```diff
@@ -1158,7 +1199,19 @@ static INT_PTR CALLBACK AboutDlgProc( HWND hWnd, UINT msg, WPARAM wParam, LPARAM
                 WCHAR szAuthorsText[20];
 
                 // Preload the ROS bitmap
-                hLogoBmp = (HBITMAP)LoadImage(shell32_hInstance, MAKEINTRESOURCE(IDB_REACTOS), IMAGE_BITMAP, 0, 0, LR_DEFAULTCOLOR);
+                //hLogoBmp = (HBITMAP)LoadImage(shell32_hInstance, MAKEINTRESOURCE(IDB_REACTOS), IMAGE_BITMAP, 0, 0, LR_DEFAULTCOLOR);
```
GeoB99, line 1202: This doesn't make any sense. Whilst the whole file's code comes from Wine, without a doubt, this specific line is definitely not part of Wine (Wine doesn't even show the ROS logo originally, which I hope this is self-explanatory). 

Thoroughly searching within Git history using `git blame`, the code was added by this commit -- [`6ced137`](https://github.com/reactos/reactos/commit/6ced137ce2f6f415c2106ce8b57d936002cf5941). Just removing this line should be enough as it was added by us, not Wine.

GeoB99 (changes requested):

dll/win32/shell32/wine/shell32_main.c
binarymaster: From [WINESYNC.txt](https://github.com/reactos/reactos/blob/master/media/doc/WINESYNC.txt)
```
dll/win32/shell32             # Forked at Wine-20071011
```
So there is no much point of using `#ifdef __REACTOS__` in shell32.

dll/win32/shell32/wine/shell32_main.c
binarymaster: > Because this is a file from Wine, which you aren't supposed to modify

See my previous comment, our shell32 is forked, and not being synched with Wine.

dll/win32/shell32/wine/shell32_main.c
jeffbox12: I didn't know this was a forked file. Thanks for pointing it out.

dll/win32/shell32/wine/shell32_main.c
```diff
@@ -1158,7 +1197,16 @@ static INT_PTR CALLBACK AboutDlgProc( HWND hWnd, UINT msg, WPARAM wParam, LPARAM
                 WCHAR szAuthorsText[20];
 
                 // Preload the ROS bitmap
-                hLogoBmp = (HBITMAP)LoadImage(shell32_hInstance, MAKEINTRESOURCE(IDB_REACTOS), IMAGE_BITMAP, 0, 0, LR_DEFAULTCOLOR);
+                if (IsVersionWorkstation())
```
learn-more, line 1200: Why not use `IsWindowsServer` from `versionhelpers.h`?

jeffbox12: Great Suggestion! That cleaned up most of the code.

katahiromz (approved):

katahiromz:
Please fix conflict.

katahiromz:
Which is better?
無題
無題2

milawynsrealm:
@katahiromz - The bottom image is an improvement. The logo and the product type are better connected and make it easier for the user to make that connection. But since graphic design is more of an art than an exact science, I leave it up to the submitter on what is the better layout. If you don't know what to do next, here is a link that might help ( https://www.printwand.com/blog/basic-alignment-principles-in-graphic-design-with-examples ).

RepOfCATV95:
@milawynsrealm And what is about graphic design means for ReactOS banner image?

milawynsrealm:
@RepOfCATV95 - Just some tips for the alignment of the text. If you don't think it's necessary, then don't worry about it. It looks fine as is.

PR #3625, opened 2021-04-25
[CMDUTILS][FC] Implement text file comparison

katahiromz:
## Purpose

Implement text file comparison by using file mappings (both Unicode and ANSI).
JIRA issue: [CORE-17500](https://jira.reactos.org/browse/CORE-17500)

shot

## Proposed changes

- We split text into lines, and then compare lines.
- If two text file are identical, return `FCRET_IDENTICAL`.
- If different, show the difference and return `FCRET_DIFFERENT`.
- If any error occurred, then return `FCRET_INVALID`.

## Comparison

BEFORE:
before
61 failures.
AFTER:
after
7 failures.

## TODO

- [x] `/C`: Ignores the letter case.
- [ ] `/LBn`: Sets the number of lines for the internal line buffer to N. The default length of the line buffer is 100 lines. If the files that you are comparing have more than 100 consecutive differing lines, fc cancels the comparison.
- [x] `/N`: Displays the line numbers during an ASCII comparison.
- [ ] `/nnnn`: Specifies the number of consecutive lines that must match following a mismatch, before fc considers the files to be resynchronized. If the number of matching lines in the files is less than nnnn, fc displays the matching lines as differences. The default value is 2.
- [x] `/T`: Prevents fc from converting tabs to spaces.
- [x] `/U`: Compares files as Unicode text files.
- [x] `/W`: Compresses white space (that is, tabs and spaces) during the comparison. If a line contains many consecutive spaces or tabs, /w treats these characters as a single space. When used with /w, fc ignores white space at the beginning and end of a line.
- [x] Do test.

## See also

https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/fc

### DIFF
--- a/base/applications/cmdutils/fc/CMakeLists.txt
+++ b/base/applications/cmdutils/fc/CMakeLists.txt
@@ -1,6 +1,6 @@
 include_directories(${REACTOS_SOURCE_DIR}/sdk/lib/conutils)
 
-add_executable(fc fc.c fc.rc)
+add_executable(fc fc.c texta.c textw.c fc.rc)
 set_module_type(fc win32cui UNICODE)
 target_link_libraries(fc conutils ${PSEH_LIB})
 add_importlibs(fc msvcrt user32 kernel32)
--- a/base/applications/cmdutils/fc/CMakeLists.txt
+++ b/base/applications/cmdutils/fc/CMakeLists.txt
@@ -2,6 +2,6 @@ include_directories(${REACTOS_SOURCE_DIR}/sdk/lib/conutils)
 
 add_executable(fc fc.c texta.c textw.c fc.rc)
 set_module_type(fc win32cui UNICODE)
-target_link_libraries(fc conutils ${PSEH_LIB})
-add_importlibs(fc msvcrt user32 kernel32)
+target_link_libraries(fc conutils wine ${PSEH_LIB})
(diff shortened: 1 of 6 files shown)


katahiromz:
For your convenience:
katahiromz/fc
https://github.com/katahiromz/fc

Vicc2008:
I have a Samsung Galaxy Tab E SM-T113NU with spreadtrum SC8830 and I want to port Windows XP Tablet PC Edition to arm processors

katahiromz:
Time out. I will merge.

base/applications/cmdutils/fc/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 include_directories(${REACTOS_SOURCE_DIR}/sdk/lib/conutils)
 
-add_executable(fc fc.c fc.rc)
+add_executable(fc fc.c texta.c textw.c fc.rc)
 set_module_type(fc win32cui UNICODE)
-target_link_libraries(fc conutils ${PSEH_LIB})
-add_importlibs(fc msvcrt user32 kernel32)
+target_link_libraries(fc conutils wine ${PSEH_LIB})
```
HBelusca, line 5: wth do we need "wine" there?

katahiromz: I will delete it in #3640.

PR #3627, opened 2021-04-25
[WINDOWSCODECS_WINETEST] tiffformat.c: Sync with Wine Staging 6.7

SergeGautherie:
## Purpose

Fix Clang-Cl
'...\modules\rostests\winetests\windowscodecs\tiffformat.c(1172,110): warning: use of logical '||' with constant operand [-Wconstant-logical-operand]'

## Proposed changes

Copied Wine file.
There is no Wine Staging patch.

And properly mark unsupported 96bppRGBFloat GUID.

--

ROS current result: 740 / 0 / 0

ROS new result: 829 / 2 / 0.
WTB WXP/WS03 results: 685 / 0 / 0

### DIFF
--- a/modules/rostests/winetests/windowscodecs/tiffformat.c
+++ b/modules/rostests/winetests/windowscodecs/tiffformat.c
@@ -402,7 +402,7 @@ static HRESULT create_decoder(const void *image_data, UINT image_size, IWICBitma
     return hr;
 }
 
-static HRESULT get_pixelformat_info(const GUID *format, UINT *bpp, UINT *channels, BOOL *trasparency)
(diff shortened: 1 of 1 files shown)


zefklop:
Please configure & use the winesync script for this and #3628

SergeGautherie:
> Please configure & use the winesync script for this and #3628

As discussed by email, it looks like requirements, especially pygit2, are not (easily) supported on WXP, which is what I can use.

I fully like that script, but in this cases:
this PR: simple enough that detailed commits seem overkill;
#3628: that is not even intended as a full sync.

PR #3630, opened 2021-04-26
[NTOS:MM] Implement proper refcounting of page tables on amd64

zefklop:
## Purpose

Count page table references on amd64 so that we can recursevely erase PDEs, PPEs & PXEs when erasing PTEs

JIRA issue: [CORE-17552](https://jira.reactos.org/browse/CORE-17552)

## Proposed changes

Use the relevant fields of the Soft PTE to count this.


ntoskrnl/mm/ARM3/virtual.c
```diff
@@ -734,21 +739,48 @@ MiDeleteVirtualAddresses(IN ULONG_PTR Va,
...
+        /* Release the lock */
+        MiReleasePfnLock(OldIrql);
+
         /* Check remaining PTE count (go back 1 page due to above loop) */
         if (MiQueryPageTableReferences((PVOID)(Va - PAGE_SIZE)) == 0)
         {
             if (PointerPde->u.Long != 0)
             {
-                /* Delete the PTE proper */
+                OldIrql = MiAcquirePfnLock();
```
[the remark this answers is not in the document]
zefklop: This whole function locks too much already. I'll have a try at making it lock-less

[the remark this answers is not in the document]
zefklop: Well, in fact MiDeletePte should not require to have the PFN lock held. This will be for an other PR.

ntoskrnl/mm/ARM3/virtual.c
```diff
@@ -2456,6 +2501,7 @@ MiMakePdeExistAndMakeValid(IN PMMPDE PointerPde,
        //
        ASSERT(KeAreAllApcsDisabled() == TRUE);
 
+#if _MI_PAGING_LEVELS == 4
```
tkreuzer, line 2504: Wait, how was this even working before?

ntoskrnl/mm/balance.c
```diff
@@ -320,32 +313,6 @@ MiBalancerThread(PVOID Unused)
         {
             ULONG InitialTarget = 0;
 
-#if (_MI_PAGING_LEVELS == 2)
```
tkreuzer, line 323: Does this destroy anything? Or is the whole paging thing fake anyway?

ntoskrnl/mm/marea.c
```diff
@@ -548,73 +548,4 @@ MiRosCleanupMemoryArea(
     ASSERT(NT_SUCCESS(Status));
 }
 
-VOID
```
tkreuzer, line 551: What's this? Getting rid of old unused stuff?

ntoskrnl/mm/balance.c
zefklop: This function is never called outside the "balancer thread"

ntoskrnl/mm/ARM3/miarm.h
```diff
@@ -2466,6 +2433,154 @@ MiSynchronizeSystemPde(PMMPDE PointerPde)
...
+
+    /* Make sure we're locked */
+    ASSERT(PsGetCurrentThread()->OwnsProcessWorkingSetExclusive);
+
+    /* If we're decreasing refcount, then it must be valid! */
+    ASSERT(PointerPde->u.Hard.Valid == 1);
+
+    /* This lies on the PFN */
+    Pfn = MiGetPfnEntry(PFN_FROM_PDE(PointerPde));
+    Pfn->OriginalPte.u.Soft.UsedPageTableEntries--;
```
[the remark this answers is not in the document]
zefklop: The idea is that if the page gets written to disk, the PDE/PPE/PXE gets the right from OrgininalPte without having to think of it.

The MMPFN struct lost the UsedPageTableEntries field from Win8. From my POV, it's only a waste of space not to use the available bits of OriginalPte...

ntoskrnl/mm/ARM3/miarm.h
```diff
@@ -2466,6 +2433,154 @@ MiSynchronizeSystemPde(PMMPDE PointerPde)
...
+    /* Check if PXE or PPE have references first. */
+#if _MI_PAGING_LEVELS == 4
+    PointerPxe = MiAddressToPxe(Address);
+    if ((PointerPxe->u.Hard.Valid == 1) || (PointerPxe->u.Soft.Transition == 1))
+    {
+        Pfn = MiGetPfnEntry(PFN_FROM_PXE(PointerPxe));
+        if (Pfn->OriginalPte.u.Soft.UsedPageTableEntries == 0)
+            return 0;
+    }
+    else if (PointerPxe->u.Soft.UsedPageTableEntries == 0)
```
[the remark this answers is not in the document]
zefklop: To know if it's valid or not. Legacy Mm relies on this...

ntoskrnl/mm/ARM3/miarm.h
```diff
@@ -2466,6 +2433,154 @@ MiSynchronizeSystemPde(PMMPDE PointerPde)
...
+#if _MI_PAGING_LEVELS == 4
+    PMMPXE PointerPxe;
+#endif
+    PMMPFN Pfn;
+
+    /* Make sure we're locked */
+    ASSERT((PsGetCurrentThread()->OwnsProcessWorkingSetExclusive) || (PsGetCurrentThread()->OwnsProcessWorkingSetShared));
+
+    /* We must not be holding the PFN lock */
+    ASSERT(KeGetCurrentIrql() < DISPATCH_LEVEL);
```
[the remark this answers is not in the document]
zefklop: Good question. Removing.

zefklop:
Test results: https://reactos.org/testman/compare.php?ids=77364,77366

ntoskrnl/mm/ARM3/virtual.c
tkreuzer: Ahh, right.

ntoskrnl/mm/ARM3/miarm.h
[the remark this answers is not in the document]
tkreuzer: Now I understand why. It's because you "unified" page.c and the new code now handles paging and uses this, while the old code didn't handle that. There is a problem though: it's called with the PFN lock held all over the place. No wonder, since it's very unintuitive that it works so differently than the other 2 functions above. I'd rather use something like `MiIsUserPageTablePresent()`, where the new behavior is needed and keep this function without the paging code.

ntoskrnl/mm/ARM3/virtual.c
```diff
@@ -2034,12 +2099,18 @@ MiIsEntireRangeCommitted(IN ULONG_PTR StartingAddress,
             }
             else
             {
+                /* Is the entire VAD committed? If not, fail */
+                if (!Vad->u.VadFlags.MemCommit) return FALSE;
```
tkreuzer, line 2100: Question: Can page tables be decommitted, if the entire range of PTEs is decommitted?

ntoskrnl/mm/ARM3/session.c
```diff
@@ -477,7 +477,7 @@ MiSessionInitializeWorkingSetList(VOID)
 
     /* Fill out the two pointers */
     MmSessionSpace->Vm.VmWorkingSetList = WorkingSetList;
-    MmSessionSpace->Wsle = (PMMWSLE)WorkingSetList->UsedPageTableEntries;
+    MmSessionSpace->Wsle = (PMMWSLE)((&WorkingSetList->VadBitMapHint) + 1);
```
tkreuzer, line 480: Sure this is correct for x64? Shouldn't it rather be ->CommittedPageDirectoryParents?

ntoskrnl/mm/ARM3/miarm.h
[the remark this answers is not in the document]
tkreuzer: I guess you deleted them all now.

ntoskrnl/mm/marea.c
```diff
@@ -547,76 +547,4 @@ MiRosCleanupMemoryArea(
...
-
-#ifndef _M_AMD64
-    OldIrql = MiAcquireExpansionLock();
-    RemoveEntryList(&Process->MmProcessLinks);
-    MiReleaseExpansionLock(OldIrql);
-#endif
-    MmLockAddressSpace(&Process->Vm);
-
-#if (_MI_PAGING_LEVELS == 2)
-    {
```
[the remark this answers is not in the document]
zefklop: Yes, because now we properly delete the PDE each time its PageTableReference reaches zero. See page.c

ntoskrnl/mm/ARM3/virtual.c
```diff
@@ -428,9 +434,16 @@ MiDeletePte(IN PMMPTE PointerPte,
             /* Drop the reference on the page table. */
             MiDecrementShareCount(MiGetPfnEntry(Pfn1->u4.PteFrame), Pfn1->u4.PteFrame);
 
+            /* In case of shared page, the prototype PTE must be in transition, not the process one */
             ASSERT(Pfn1->u3.e1.PrototypePte == 0);
 
-            /* Make the page free. For prototypes, it will be made free when deleting the section object */
+            /* Delete the PFN */
+            MI_SET_PFN_DELETED(Pfn1);
```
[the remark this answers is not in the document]
zefklop: The PFN must be unconditionally made free. There is no way for the (future) MPW to know that it is in fact writing a stale PFN if we don't set this.

I don't remember why I put it in this specific PR though...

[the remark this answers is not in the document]
zefklop: Geez, this commit message doesn't make sense... I'll rebase properly.

ntoskrnl/mm/ARM3/session.c
zefklop: Never sure of anything :wink: 
In case of the session address space, I guess that HighestUserAddress & the like don't make sense, but Page Table committed state does. This is true for process VM as well, where Wsle is set to (VmWorkingSetList + 1) while CommittedPageDirectoryParents is variable length array.
I guess we need a helper function giving the WorkingSetList size.

zefklop:
Locally checked that each commit get properly built.

ntoskrnl/mm/ARM3/virtual.c
```diff
@@ -46,157 +46,134 @@ MiCalculatePageCommitment(IN ULONG_PTR StartingAddress,
...
+    BOOLEAN OnPpeBoundary = TRUE;
+#if _MI_PAGING_LEVELS == 4
+    PMMPXE PointerPxe;
+    BOOLEAN OnPxeBoundary = TRUE;
+#endif
+#endif
 
-    /* Compute starting and ending PTE and PDE addresses */
-    PointerPde = MiAddressToPde(StartingAddress);
+    /* Make sure this all makes sense */
```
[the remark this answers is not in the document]
zefklop: Sure.

ntoskrnl/mm/ARM3/virtual.c
zefklop: Yes. Page tables must be accounted for when calculating commit charges. So each time we commit a PTE, we must check if the PDE/PPE/PXE supporting it is, and if not, act accordingly.
Same thing when decommiting, but there it gets a bit more complicated, one PDE can be underneath more than one committed range.

ntoskrnl/mm/ARM3/mdlsup.c
```diff
@@ -341,28 +339,16 @@ MiUnmapLockedPagesInUserSpace(
         PageTablePage = PointerPde->u.Hard.PageFrameNumber;
         MiDecrementShareCount(MiGetPfnEntry(PageTablePage), PageTablePage);
 
+        if (MiDecrementPageTableReferences(BaseAddress) == 0)
+        {
+            MiDeletePde(PointerPde, Process);
```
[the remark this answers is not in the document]
zefklop: Sure.

ntoskrnl/mm/ARM3/session.c
tkreuzer: Sounds good. And some comments documenting how it works would be great.

ntoskrnl/mm/ARM3/virtual.c
tkreuzer: Related to this: https://github.com/tkreuzer/reactos/commit/d002afa2cae8439a200cbdb88bea6aee3117108b#r51788177

ntoskrnl/mm/ARM3/mdlsup.c
```diff
@@ -341,28 +339,17 @@ MiUnmapLockedPagesInUserSpace(
...
-            ASSERT(PointerPde->u.Hard.Valid == 1);
-            if (MiQueryPageTableReferences(BaseAddress) == 0)
-            {
-                ASSERT(PointerPde->u.Long != 0);
-                MiDeletePte(PointerPde,
-                            MiPteToAddress(PointerPde),
-                            Process,
-                            NULL);
-            }
-        }
```
zefklop, line 365: Am I the only one seeing how broken this was ? How did it even work ?

tkreuzer: Yeah, it's checking the wrong page table references. Would probably never have deleted the PDEs, or even worse, deleted a PDE, which still had references.

tkreuzer (approved):

ntoskrnl/mm/ARM3/mdlsup.c
zefklop: This bug is the perfect random f*ck. Would only be wrong if freed range is one page short from the beginning of the PDE and the next PDE is on a completely free range.

ntoskrnl/mm/ARM3/virtual.c
```diff
@@ -719,36 +710,26 @@ MiDeleteVirtualAddresses(IN ULONG_PTR Va,
                     /* The PTE was never mapped, just nuke it here */
                     MI_ERASE_PTE(PointerPte);
                 }
+
+                if (MiDecrementPageTableReferences((PVOID)Va) == 0)
+                {
+                    ASSERT(PointerPde->u.Long != 0);
+                    ASSERT(MiIsPteOnPdeBoundary(PointerPte + 1) || ((Va + PAGE_SIZE) > EndingAddress));
```
tkreuzer, line 717: This fails on x64. And it doesn't seem correct. There can be 0 PTEs at the end of a page table.

tkreuzer (changes requested):
Please fix assert in MiDeleteVirtualAddresses

tkreuzer (approved):

ntoskrnl/mm/ARM3/virtual.c
```diff
@@ -719,40 +710,30 @@ MiDeleteVirtualAddresses(IN ULONG_PTR Va,
                     /* The PTE was never mapped, just nuke it here */
                     MI_ERASE_PTE(PointerPte);
                 }
+
+                if (MiDecrementPageTableReferences((PVOID)Va) == 0)
+                {
+                    ASSERT(PointerPde->u.Long != 0);
+                    /* Delete the PDE proper */
+                    MiDeletePde(PointerPde, CurrentProcess);
+                    /* Jump */
```
tkreuzer, line 719: Lazy late night comment? 😄

zefklop: Oh well...

zefklop:
All bots green. @Extravert-ir waiting for your review.

PR #3633, opened 2021-04-29
[NTOS:OB] ObQueryDeviceMapInformation(): Fix annotations

SergeGautherie:
Be consistent and correct...

### DIFF
--- a/ntoskrnl/ob/devicemap.c
+++ b/ntoskrnl/ob/devicemap.c
@@ -536,9 +536,10 @@ ObInheritDeviceMap(IN PEPROCESS Parent,
 
 NTSTATUS
 NTAPI
-ObQueryDeviceMapInformation(IN PEPROCESS Process,
-                            IN PPROCESS_DEVICEMAP_INFORMATION DeviceMapInfo,
-                            IN ULONG Flags)
+ObQueryDeviceMapInformation(
+    _In_opt_ PEPROCESS Process,
+    _Out_ PPROCESS_DEVICEMAP_INFORMATION DeviceMapInfo,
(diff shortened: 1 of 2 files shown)


ntoskrnl/ob/devicemap.c
```diff
@@ -536,9 +536,10 @@ ObInheritDeviceMap(IN PEPROCESS Parent,
 
 NTSTATUS
 NTAPI
-ObQueryDeviceMapInformation(IN PEPROCESS Process,
-                            IN PPROCESS_DEVICEMAP_INFORMATION DeviceMapInfo,
-                            IN ULONG Flags)
+ObQueryDeviceMapInformation(
+    _In_opt_ PEPROCESS Process,
+    _Out_ PPROCESS_DEVICEMAP_INFORMATION DeviceMapInfo,
+    _In_ ULONG Flags)
```
GeoB99, line 542: Are you intending to introduce SAL annotations for the whole file or just that function?

SergeGautherie: This function only, for the time being. (Noticed while checking #3634.)

Extravert-ir: What's the point in annotating only one function in a file?

SergeGautherie: It's a start... Does that cause any harm?

GeoB99:
> It's a start... Does that cause any harm?

This time I'll give it a pass but in the future you can make a PR that annotates the functions with SAL of the whole file instead of making million of PRs (and by the way, you have like 44 open PRs left that you must address yourself, don't ya think?)

PR #3637, opened 2021-05-03
[GDI32] Fix wrong ordering of parameters in NtGdiDdGetDC call

oleg-dubinskiy:
## Purpose

Place `pColorTable` after `pSurfaceLocal->hDDSurface`, instead of before it, since such ordering is used in the calling `NtGdiDdGetDC`. See https://docs.microsoft.com/en-us/windows/win32/devnotes/-dxgkernel-ntgdiddgetdc for the reference.
It allows to properly pass the DirectDraw surface handle (and palette entry pointer) from MS ddraw into win32k.
Otherwise, they are passing into the wrong parameters of actual `NtGdi`* function, and due to this, since they're detected as invalid, they become `NULL`, and that function does not work correctly.
Required by MS DirectDraw stack (ddraw.dll & dxg.sys). Although it does not completely fix the issue, but it at least improves the situation.

### DIFF
--- a/win32ss/gdi/gdi32/misc/gdientry.c
+++ b/win32ss/gdi/gdi32/misc/gdientry.c
@@ -1924,7 +1924,7 @@ DdGetDC(LPDDRAWI_DDRAWSURFACE_LCL pSurfaceLocal,
         LPPALETTEENTRY pColorTable)
 {
     /* Call win32k directly */
-    return NtGdiDdGetDC(pColorTable, (HANDLE) pSurfaceLocal->hDDSurface);
+    return NtGdiDdGetDC((HANDLE) pSurfaceLocal->hDDSurface, pColorTable);
 }
 
 /*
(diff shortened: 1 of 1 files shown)


win32ss/gdi/gdi32/misc/gdientry.c
```diff
@@ -1924,7 +1924,7 @@ DdGetDC(LPDDRAWI_DDRAWSURFACE_LCL pSurfaceLocal,
         LPPALETTEENTRY pColorTable)
 {
     /* Call win32k directly */
-    return NtGdiDdGetDC(pColorTable, (HANDLE) pSurfaceLocal->hDDSurface);
+    return NtGdiDdGetDC((HANDLE) pSurfaceLocal->hDDSurface, pColorTable);
```
HBelusca, line 1927:
```suggestion
    return NtGdiDdGetDC((HANDLE)pSurfaceLocal->hDDSurface, pColorTable);
```

PR #3645, opened 2021-05-06
[GITHUB] build.yml: Add build-msbuild-amd64

SergeGautherie:
Add the missing build,
even though a couple issues remain to fix.

**ToDo: Choose one of the two hacks, or suggest a better solution.**


zefklop:
Can you rebase this with only the commit enabling the build ?

SergeGautherie:
> Can you rebase this with only the commit enabling the build ?

Why? I assume that would still break this new build...

PR #3646, opened 2021-05-07
[BCRYPT] Sync with Wine-6.7. Added mbedtls.c stubs of key operation functions.

mande806:
## Purpose

Updates BCRYPT to the latest Wine version. I plan to submit mbedtls-based ECC cert support in a future PR after this is merged.
This is my first PR for ReactOS, so please let me know if I'm doing something incorrectly. Thank you.

## Proposed changes

- Updates BCRYPT to the latest Wine version. This version:
    - splits out the key operation functions (which need an external library such as gnuTLS or mbedtls) into a separate interface.
    - has built-in support for many of the hash functions which our old copy needed mbedtls for.

- because of the separate interface, there are now only 3 __REACTOS__ #ifdefs! 🎉

- After this is merged successfully, I will eventually submit a PR with implementations for the portions of the mbedtls.c file needed for ECC cert support (so we can download from GitHub with rapps again 😄)

## TODO

- [X] passes bcrypt_winetest.exe successfully:
image
- [X] non-ECC cert https links download fine in rapps
-  please help test this and let me know if I did this right, as this is my first pull request for reactos

### DIFF
--- a/dll/win32/bcrypt/CMakeLists.txt
+++ b/dll/win32/bcrypt/CMakeLists.txt
@@ -7,6 +7,10 @@ spec2def(bcrypt.dll bcrypt.spec ADD_IMPORTLIB)
 
 list(APPEND SOURCE
     bcrypt_main.c
+    mbedtls.c
+    md2.c
+    sha256.c
+    sha512.c
     version.rc
     ${CMAKE_CURRENT_BINARY_DIR}/bcrypt_stubs.c
     ${CMAKE_CURRENT_BINARY_DIR}/bcrypt.def)
--- a/dll/win32/bcrypt/bcrypt.spec
+++ b/dll/win32/bcrypt/bcrypt.spec
@@ -1,56 +1,58 @@
-@ stub BCryptAddContextFunction
-@ stub BCryptAddContextFunctionProvider
+@ stdcall BCryptAddContextFunction(long wstr long wstr long)
+@ stdcall BCryptAddContextFunctionProvider(long wstr long wstr wstr long)
 @ stdcall BCryptCloseAlgorithmProvider(ptr long)
 @ stub BCryptConfigureContext
 @ stub BCryptConfigureContextFunction
 @ stub BCryptCreateContext
 @ stdcall BCryptCreateHash(ptr ptr ptr long ptr long long)
-@ stub BCryptDecrypt
+@ stdcall BCryptDecrypt(ptr ptr long ptr ptr long ptr long ptr long)
 @ stub BCryptDeleteContext
-@ stub BCryptDeriveKey
+@ stdcall BCryptDeriveKey(ptr wstr ptr ptr long ptr long)
+@ stdcall BCryptDeriveKeyCapi(ptr ptr ptr long long)
+@ stdcall BCryptDeriveKeyPBKDF2(ptr ptr long ptr long int64 ptr long long)
 @ stdcall BCryptDestroyHash(ptr)
-@ stdcall -stub BCryptDestroyKey(ptr)
-@ stub BCryptDestroySecret
-@ stub BCryptDuplicateHash
-@ stub BCryptDuplicateKey
-@ stub BCryptEncrypt
+@ stdcall BCryptDestroyKey(ptr)
+@ stdcall BCryptDestroySecret(ptr)
+@ stdcall BCryptDuplicateHash(ptr ptr ptr long long)
+@ stdcall BCryptDuplicateKey(ptr ptr ptr long long)
+@ stdcall BCryptEncrypt(ptr ptr long ptr ptr long ptr long ptr long)
 @ stdcall BCryptEnumAlgorithms(long ptr ptr long)
 @ stub BCryptEnumContextFunctionProviders
-@ stub BCryptEnumContextFunctions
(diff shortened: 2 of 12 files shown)


SergeGautherie:
> That "fix" describes itself as "the wrong way forward".

Then, for the time being, please add an explicit note in commit summary, and/or the WINESYNC file, that/why W.S. patches are intentionally ignored.

> That is my personal opinion

(Not my decision to make either.)

mande806:
> Then, for the time being, please add an explicit note in commit summary, and/or the WINESYNC file, that/why W.S. patches are intentionally ignored.

I specifically documented in WINESYNC.txt that this was synced with WINE-6.7, not Wine-Staging. I tried to follow the pattern already used by other modules to show this.
If this is not sufficient, please let me know what you are thinking of as a solution.

SergeGautherie:
> I specifically documented in WINESYNC.txt that this was synced with WINE-6.7, not Wine-Staging

Yes, but my point is to add a reminder about _why_.

> If this is not sufficient, please let me know what you are thinking of as a solution.

What I wrote in my previous comment.

mande806:
> > I specifically documented in WINESYNC.txt that this was synced with WINE-6.7, not Wine-Staging
> 
> Yes, but my point is to add a reminder about _why_.
> 
> > If this is not sufficient, please let me know what you are thinking of as a solution.
> 
> What I wrote in my previous comment.

None of the other modules record *why* they were synced/forked from either Wine or Wine-Staging, so if that is desirable, I need to know what you're looking for.

SergeGautherie:
> None of the other modules record _why_ they were synced/forked from either Wine or Wine-Staging

That is exactly my point: to remember why some (yours here) are Wine only instead of currently expected W.S..

> I need to know what you're looking for.

1+ words/sentences, like your reply in this PR.

Anyway, don't reply again: committer will do what he wants...

zefklop:
I like this, but I would like this better if you'd use the winesync script :smile: 

I agree that applying wine-staging patches here is a bad idea. Would you mind making that optional in the script ? If not, I can find the time to do that.

mande806:
> I like this, but I would like this better if you'd use the winesync script 😄
> 
> I agree that applying wine-staging patches here is a bad idea. Would you mind making that optional in the script ? If not, I can find the time to do that.

I will take a look at doing that when I get a chance. I suppose you'll want me to make a winesync .cfg for bcrypt? I agree winesync by script will probably be useful going forward to preserve our ReactOS-specific changes.
The primary issue with automated winesyncing the current bcrypt is that *literally all* of our existing ReactOS-specific changes are not relevant to the latest version (which includes the basic hashing algorithms internally), and the codebase has changed so much that an automated merge attempt would be very unnecessarily messy.

zefklop:
> > I like this, but I would like this better if you'd use the winesync script smile
> > I agree that applying wine-staging patches here is a bad idea. Would you mind making that optional in the script ? If not, I can find the time to do that.
> 
> I will take a look at doing that when I get a chance. I suppose you'll want me to make a winesync .cfg for bcrypt? I agree winesync by script will probably be useful going forward to preserve our ReactOS-specific changes.
Yes
> The primary issue with automated winesyncing the current bcrypt is that _literally all_ of our existing ReactOS-specific changes are not relevant to the latest version (which includes the basic hashing algorithms internally), and the codebase has changed so much that an automated merge attempt would be very unnecessarily messy.

Well, I know that this is tedious, but this makes regression testing much easier in the end. Believe me, it's better to take the pain of merging wine-commits one by one with the help of the winesync script and then be done with it, rather than having a monster commit, mother of all regressions and wonder which tiny part of it is the real root cause. If none of our specific change is needed anymore, then it's fine. Each relevant wine commit will supersede them in the end, and we will have the relevant commit log if all hell breaks loose in the end.

JoachimHenze:
mande806 you mentioned: "After this is merged successfully, I will eventually submit a PR with implementations for the portions of the mbedtls.c file needed for ECC cert support (so we can download from GitHub with rapps again 😄)"
Is this just a plan? Or do you have something working already, that we could play with for real?
This would ofc give also much more opportunities to really test the existing PR3646.

DarkFire01:
Hello! It seems the author of this PR is long gone but I do believe this is worth rekindling the subject.

I attempted to winesync bcrypt just to realize the.. current state of affairs.

as the PR author said our usage of embedtls ends up sort of.. depreciated. I believe the sane set of operations should be this.

1) Let's go ahead with this PR, and get it merged as is.
This includes the bulk of the logic we need to get off the ground for a good chunk of APIs which maybe vista+ apps start to use more aggressively. 

2) update the winetests for bcrypt accordingly to the version specified here.

3) if we need more we can update to the latest wine which opts for using the vista+ exports from ntdll along with some internal logic for the bare minimum of what bcrypt does.  (Which is all we allow with embedtls atm anyway lol)

A more long term solution to ncrypt, bcrypt and some other components would be a correct usage of
https://github.com/microsoft/SymCrypt/tree/main 

Which is what windows 10+ uses for these dlls.


On a another note to @mande806 
Sorry about how this PR ended up! I hope we can get your work merged :) to some degree.

tkreuzer:
The hash functions already exist in cryptlib.

tkreuzer:
Oh and that macos.c file can be removed. And gnutls.c probably also.

PR #3651, opened 2021-05-08
[KMTESTS:SE] Remove erroneous tests regarding impersonation level

GeoB99:
These tests didn't make much sense. The captured token from the subject context is not an impersonation token and both Windows and ReactOS ultimately fail on those tests.


GeoB99:
**BEFORE**
Capture3
**AFTER**
Capture2
**Windows Server 2003 SP2**
Capture

PR #3658, opened 2021-05-12
[MORE][CONUTILS] Implement missing features of the MORE command.

katahiromz:
## Purpose

Implement missing features of the `MORE` command.

JIRA issue: [CORE-4019](https://jira.reactos.org/browse/CORE-4019)

## Proposed changes

- Extend `conutils` module.
- Implement all the subcommands of `MORE`.
- Handle long lines.
- Modify `IDS_USAGE` resource string.
- Add `IDS_BAD_FLAG`, `IDS_CONTINUE_OPTIONS`, `IDS_CONTINUE_LINES` and `IDS_CONTINUE_LINE_AT` resource strings.
- Handle command line options.

## TODO

- [x] Do test.
- [x] `/C` option.
- [x]  `/P` option.
- [x] `/S` option.
- [x] `/Tn` option.
- [x] `+n` option.
- [x] `P` subcommand.
- [x] `S` subcommand.
- [x] `F` subcommand.
- [x] `Q` subcommand.
- [x] `=` subcommand.
- [x]  `?` subcommand.
- [x] `<Space>` subcommand.
- [x] `<Enter>` subcommand.
- [x] Considering double-width characters.
- [x] Tabs

### DIFF
--- a/base/applications/cmdutils/more/lang/bg-BG.rc
+++ b/base/applications/cmdutils/more/lang/bg-BG.rc
@@ -24,13 +24,14 @@ at ""-- Continue --"" prompt:\n\
    =        Show line number.\n\
    ?        Show help line.\n\
    <Space>  Display next page.\n\
-   <Enter>  Display next line.\n"
+   <Enter>  Display next line.\n\
+   <Esc>    Same as Q.\n"
 
     IDS_CONTINUE "-- Продължаване --"
     IDS_CONTINUE_PROGRESS "-- Продължаване (%d%%) --"
     IDS_FILE_ACCESS "Няма достъп до файл %s.\n"
     IDS_BAD_FLAG "Invalid argument - '%ls'\n"
-    IDS_CONTINUE_OPTIONS "-- Continue (%d%%) -- [Options: psfq=<Space><Enter>]"
+    IDS_CONTINUE_OPTIONS "-- Continue (%d%%) -- [Options: psfq=<Space><Enter><Esc>]"
     IDS_CONTINUE_LINES "-- Continue (%d%%) -- Lines: "
     IDS_CONTINUE_LINE_AT "-- Continue (%d%%) -- [Line:%d]"
 END
--- a/base/applications/cmdutils/more/lang/bg-BG.rc
+++ b/base/applications/cmdutils/more/lang/bg-BG.rc
@@ -17,15 +17,11 @@ Options:\n\
   Command                    A command, whose output shall be displayed.\n\n\
 If extended features are enabled, the following commands are available\n\
 at ""-- Continue --"" prompt:\n\
-   P n      Display n lines.\n\
-   S n      Skip n lines.\n\
-   F        Display next file.\n\
-   Q        Quit.\n\
-   =        Show line number.\n\
-   ?        Show help line.\n\
-   <Space>  Display next page.\n\
-   <Enter>  Display next line.\n\
-   <Esc>    Same as Q.\n"
+   P n      Display n lines.        <Space>  Display next page.\n\
+   S n      Skip n lines.           <Enter>  Display next line.\n\
+   F        Display next file.      <Esc>    Same as Q.\n\
+   Q        Quit.                   =        Show line number.\n\
+   ?        Show help line.\n"
 
     IDS_CONTINUE "-- Продължаване --"
     IDS_CONTINUE_PROGRESS "-- Продължаване (%d%%) --"
--- a/base/applications/cmdutils/more/lang/bg-BG.rc
+++ b/base/applications/cmdutils/more/lang/bg-BG.rc
@@ -24,14 +24,13 @@ at ""-- Continue --"" prompt:\n\
    ?        Show help line.\n\
    =        Show line number.\n\
    <Space>  Display next page.\n\
-   <Enter>  Display next line.\n\
-   <Esc>    Same as Q.\n"
+   <Enter>  Display next line.\n"
 
     IDS_CONTINUE "-- Продължаване --"
     IDS_CONTINUE_PROGRESS "-- Продължаване (%d%%) --"
     IDS_FILE_ACCESS "Няма достъп до файл %s.\n"
     IDS_BAD_FLAG "Invalid argument - '%ls'\n"
-    IDS_CONTINUE_OPTIONS "-- Continue (%d%%) [Options: psfq=<Space><Enter><Esc>] --"
+    IDS_CONTINUE_OPTIONS "-- Continue (%d%%) [Options: psfq=<Space><Enter>] --"
     IDS_CONTINUE_LINES "-- Continue (%d%%) -- Lines: "
     IDS_CONTINUE_LINE_AT "-- Continue (%d%%) [Line:%d] --"
 END
--- a/base/applications/cmdutils/more/lang/bg-BG.rc
+++ b/base/applications/cmdutils/more/lang/bg-BG.rc
@@ -31,7 +31,7 @@ at ""-- Continue --"" prompt:\n\
     IDS_CONTINUE_PROGRESS "-- Продължаване (%d%%) --"
     IDS_FILE_ACCESS "Няма достъп до файл %s.\n"
     IDS_BAD_FLAG "Invalid argument - '%ls'\n"
-    IDS_CONTINUE_OPTIONS "-- Continue (%d%%) -- [Options: psfq=<Space><Enter><Esc>]"
+    IDS_CONTINUE_OPTIONS "-- Continue (%d%%) [Options: psfq=<Space><Enter><Esc>] --"
     IDS_CONTINUE_LINES "-- Continue (%d%%) -- Lines: "
-    IDS_CONTINUE_LINE_AT "-- Continue (%d%%) -- [Line:%d]"
+    IDS_CONTINUE_LINE_AT "-- Continue (%d%%) [Line:%d] --"
 END
--- a/base/applications/cmdutils/more/lang/bg-BG.rc
+++ b/base/applications/cmdutils/more/lang/bg-BG.rc
@@ -17,11 +17,15 @@ Options:\n\
   Command                    A command, whose output shall be displayed.\n\n\
 If extended features are enabled, the following commands are available\n\
 at ""-- Continue --"" prompt:\n\
-   P n      Display n lines.        <Space>  Display next page.\n\
-   S n      Skip n lines.           <Enter>  Display next line.\n\
-   F        Display next file.      <Esc>    Same as Q.\n\
-   Q        Quit.                   =        Show line number.\n\
-   ?        Show help line.\n"
+   P n      Display n lines.\n\
+   S n      Skip n lines.\n\
+   F        Display next file.\n\
+   Q        Quit.\n\
+   ?        Show help line.\n\
+   =        Show line number.\n\
+   <Space>  Display next page.\n\
+   <Enter>  Display next line.\n\
+   <Esc>    Same as Q.\n"
 
     IDS_CONTINUE "-- Продължаване --"
     IDS_CONTINUE_PROGRESS "-- Продължаване (%d%%) --"
--- a/base/applications/cmdutils/more/more.c
+++ b/base/applications/cmdutils/more/more.c
@@ -738,7 +738,7 @@ static BOOL IsFlag(PCWSTR param)
         pch = param + 1;
         if (*pch)
         {
-            wcstol(pch, &endptr, 10);
+            (void)wcstol(pch, &endptr, 10);
             return (*endptr == 0);
         }
     }
--- a/base/applications/cmdutils/more/more.c
+++ b/base/applications/cmdutils/more/more.c
@@ -61,7 +61,7 @@ BOOL bEnableExtensions = TRUE; // FIXME: By default, it should be FALSE.
 #define FLAG_PLUSn (1 << 6)
 
 static DWORD s_dwFlags = 0;
-static DWORD s_nTabWidth = 8;
+static LONG s_nTabWidth = 8;
 static DWORD s_nNextLineNo = 0;
 static BOOL s_bPrevLineIsBlank = FALSE;
 static UINT s_nPromptID = IDS_CONTINUE_PROGRESS;
@@ -728,20 +728,19 @@ LoadRegistrySettings(HKEY hKeyRoot)
 
 static BOOL IsFlag(LPCWSTR param)
 {
+    LPCWSTR pch;
+    LPWCH endptr;
+
     if (param[0] == L'/')
         return TRUE;
 
     if (param[0] == L'+')
     {
-        LPCWSTR pch = param + 1;
-        if (L'0' <= *pch && *pch <= L'9')
+        pch = param + 1;
+        if (*pch)
         {
-            do
-            {
-                ++pch;
-            } while (L'0' <= *pch && *pch <= L'9');
-
-            return (*pch == 0);
+            wcstol(pch, &endptr, 10);
+            return (*endptr == 0);
         }
     }
     return FALSE;
@@ -749,7 +748,7 @@ static BOOL IsFlag(LPCWSTR param)
 
 static BOOL ParseArgument(LPCWSTR arg, BOOL *pbHasFiles)
 {
-    LPWSTR endptr;
+    LPWCH endptr;
 
     if (arg[0] == L'/')
     {
@@ -791,10 +790,10 @@ static BOOL ParseArgument(LPCWSTR arg, BOOL *pbHasFiles)
                 }
                 break;
             case L'T':
-                if (L'0' <= arg[2] && arg[2] <= L'9')
+                if ((L'0' <= arg[2] && arg[2] <= L'9') || (arg[2] == L'-'))
                 {
                     s_dwFlags |= FLAG_Tn;
-                    s_nTabWidth = wcstoul(&arg[2], &endptr, 10);
+                    s_nTabWidth = wcstol(&arg[2], &endptr, 10);
                     if (*endptr == 0)
                         return TRUE;
                 }
@@ -805,10 +804,10 @@ static BOOL ParseArgument(LPCWSTR arg, BOOL *pbHasFiles)
     }
     else if (arg[0] == L'+')
     {
-        if (L'0' <= arg[1] && arg[1] <= L'9')
+        if ((L'0' <= arg[1] && arg[1] <= L'9') || (arg[1] == L'-'))
         {
             s_dwFlags |= FLAG_PLUSn;
-            s_nNextLineNo = wcstoul(&arg[1], &endptr, 10) + 1;
+            s_nNextLineNo = wcstol(&arg[1], &endptr, 10) + 1;
             if (*endptr == 0)
                 return TRUE;
         }
@@ -830,7 +829,8 @@ static BOOL ParseArgument(LPCWSTR arg, BOOL *pbHasFiles)
 static BOOL ParseMoreVariable(BOOL *pbHasFiles)
 {
     BOOL ret = TRUE;
-    LPWSTR psz, pch;
+    LPWSTR psz;
+    LPWCH pch;
     DWORD cch;
 
     cch = GetEnvironmentVariableW(L"MORE", NULL, 0);
--- a/base/applications/cmdutils/more/more.c
+++ b/base/applications/cmdutils/more/more.c
@@ -63,6 +63,8 @@ static DWORD s_dwFlags = FLAG_E;
 static DWORD s_nTabWidth = 8;
 static DWORD s_nNextLineNo = 0;
 static BOOL s_bPrevLineIsBlank = FALSE;
+static UINT s_nPromptID = IDS_CONTINUE_PROGRESS;
+static WCHAR s_chSubCommand = 0;
 
 static inline BOOL IsFlag(LPCWSTR param)
 {
@@ -217,8 +219,6 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
     BOOL fCtrl;
     DWORD nLines;
     WCHAR ch;
-    static UINT s_nPromptID = IDS_CONTINUE_PROGRESS;
-    static WCHAR s_chSubCommand = 0;
 Restart:
     nLines = 0;
 
@@ -1001,6 +1001,8 @@ int wmain(int argc, WCHAR* argv[])
     for (i = 1; i < argc; i++)
     {
         s_bPrevLineIsBlank = FALSE;
+        s_nPromptID = IDS_CONTINUE_PROGRESS;
+        s_chSubCommand = 0;
 
         if (IsFlag(argv[i]))
             continue;
--- a/base/applications/cmdutils/more/more.c
+++ b/base/applications/cmdutils/more/more.c
@@ -790,13 +790,10 @@ static BOOL ParseArgument(LPCWSTR arg, BOOL *pbHasFiles)
                 }
                 break;
             case L'T':
-                if ((L'0' <= arg[2] && arg[2] <= L'9') || (arg[2] == L'-'))
-                {
-                    s_dwFlags |= FLAG_Tn;
-                    s_nTabWidth = wcstol(&arg[2], &endptr, 10);
-                    if (*endptr == 0)
-                        return TRUE;
-                }
+                s_dwFlags |= FLAG_Tn;
+                s_nTabWidth = wcstol(&arg[2], &endptr, 10);
+                if (*endptr == 0)
+                    return TRUE;
                 break;
             default:
                 break;
@@ -804,13 +801,10 @@ static BOOL ParseArgument(LPCWSTR arg, BOOL *pbHasFiles)
     }
     else if (arg[0] == L'+')
     {
-        if ((L'0' <= arg[1] && arg[1] <= L'9') || (arg[1] == L'-'))
-        {
-            s_dwFlags |= FLAG_PLUSn;
-            s_nNextLineNo = wcstol(&arg[1], &endptr, 10) + 1;
-            if (*endptr == 0)
-                return TRUE;
-        }
+        s_dwFlags |= FLAG_PLUSn;
+        s_nNextLineNo = wcstol(&arg[1], &endptr, 10) + 1;
+        if (*endptr == 0)
+            return TRUE;
     }
 
     if (IsFlag(arg))
--- a/base/applications/cmdutils/more/more.c
+++ b/base/applications/cmdutils/more/more.c
@@ -408,8 +408,8 @@ Restart:
     /* If extended features are available */
     if (s_dwFlags & FLAG_E)
     {
-        /* 'Q': Quit */
-        if (KeyEvent.wVirtualKeyCode == L'Q')
+        /* 'Q' or [Esc]: Quit */
+        if (KeyEvent.wVirtualKeyCode == L'Q' || KeyEvent.wVirtualKeyCode == VK_ESCAPE)
         {
             /* We break, output a newline */
             WCHAR ch = L'\n';
--- a/base/applications/cmdutils/more/more.c
+++ b/base/applications/cmdutils/more/more.c
@@ -24,6 +24,7 @@
 
 #include <stdio.h>
 #include <stdlib.h>
+#include <assert.h>
 
 #include <windef.h>
 #include <winbase.h>
@@ -300,8 +301,9 @@ Restart:
             continue;
         }
 
-        /* Ignore any unsupported subcommand */
-        if (s_chSubCommand != L'P' && s_chSubCommand != L'S')
+        assert(s_chSubCommand == L'P' || s_chSubCommand == L'S' || s_chSubCommand == 0);
+
+        if (s_chSubCommand == 0)
             break;
 
         ch = KeyEvent.uChar.UnicodeChar;
--- a/base/applications/cmdutils/more/more.c
+++ b/base/applications/cmdutils/more/more.c
@@ -24,7 +24,6 @@
 
 #include <stdio.h>
 #include <stdlib.h>
-#include <assert.h>
 
 #include <windef.h>
 #include <winbase.h>
@@ -201,7 +200,6 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
     KEY_EVENT_RECORD KeyEvent;
     BOOL fCtrl;
     DWORD nLines;
-    WCHAR ch;
     WCHAR chSubCommand = 0;
 
 Restart:
@@ -260,6 +258,7 @@ Restart:
     {
         INPUT_RECORD ir = {0};
         DWORD dwRead;
+        WCHAR ch;
 
         do
         {
@@ -267,8 +266,16 @@ Restart:
         }
         while ((ir.EventType != KEY_EVENT) || (!ir.Event.KeyEvent.bKeyDown));
 
-        /* Got our key, return to caller */
+        /* Got our key */
         KeyEvent = ir.Event.KeyEvent;
+
+        /* Ignore any unsupported keyboard press */
+        if ((KeyEvent.wVirtualKeyCode == VK_SHIFT) ||
+            (KeyEvent.wVirtualKeyCode == VK_MENU)  ||
+            (KeyEvent.wVirtualKeyCode == VK_CONTROL))
+        {
+            continue;
+        }
 
         /* Ctrl key is pressed? */
         fCtrl = !!(KeyEvent.dwControlKeyState & (LEFT_CTRL_PRESSED | RIGHT_CTRL_PRESSED));
@@ -281,17 +288,9 @@ Restart:
             break;
         }
 
-        /* Ignore any unsupported keyboard press */
-        if ((KeyEvent.wVirtualKeyCode == VK_SHIFT) ||
-            (KeyEvent.wVirtualKeyCode == VK_MENU)  ||
-            (KeyEvent.wVirtualKeyCode == VK_CONTROL))
-        {
-            continue;
-        }
-
-        assert(chSubCommand == L'P' || chSubCommand == L'S' || chSubCommand == 0);
-
-        if (chSubCommand == 0)
+        /* If extended features are unavailable, or no
+         * pending commands, don't do more processing. */
+        if (!(s_dwFlags & FLAG_E) || (chSubCommand == 0))
             break;
 
         ch = KeyEvent.uChar.UnicodeChar;
@@ -304,7 +303,7 @@ Restart:
             ConStreamWrite(Pager->Screen->Stream, &ch, 1);
             continue;
         }
-        if (KeyEvent.wVirtualKeyCode == VK_RETURN)
+        else if (KeyEvent.wVirtualKeyCode == VK_RETURN)
         {
             if (nLines == 0)
             {
@@ -354,6 +353,16 @@ Restart:
      */
     ConClearLine(Pager->Screen->Stream);
 
+    /* Ctrl+C or Ctrl+Esc: Control Break */
+    if (fCtrl && ((KeyEvent.wVirtualKeyCode == VK_ESCAPE) ||
+                  (KeyEvent.wVirtualKeyCode == L'C')))
+    {
+        /* We break, output a newline */
+        WCHAR ch = L'\n';
+        ConStreamWrite(Pager->Screen->Stream, &ch, 1);
+        return FALSE;
+    }
+
     switch (chSubCommand)
     {
         case L'P':
@@ -371,39 +380,19 @@ Restart:
             break;
     }
 
-    /* Ctrl+C or Ctrl+Esc: Control Break */
-    if (fCtrl && ((KeyEvent.wVirtualKeyCode == VK_ESCAPE) ||
-                  (KeyEvent.wVirtualKeyCode == L'C')))
-    {
-        /* We break, output a newline */
-        WCHAR ch = L'\n';
-        ConStreamWrite(Pager->Screen->Stream, &ch, 1);
-        return FALSE;
-    }
-
-    if (fCtrl)
-    {
-        chSubCommand = 0;
-        goto Restart;
-    }
-
-    /* [Space] key: One page */
-    if (KeyEvent.wVirtualKeyCode == VK_SPACE)
-    {
-        if (s_dwFlags & FLAG_C)
-        {
-            /* Clear the screen */
-            ConClearScreen(Pager->Screen);
-        }
-        Pager->ScrollRows = Pager->ScreenRows - 1;
-        return TRUE;
-    }
-
     /* If extended features are available */
     if (s_dwFlags & FLAG_E)
     {
+        /* Ignore any key presses if Ctrl is pressed */
+        if (fCtrl)
+        {
+            chSubCommand = 0;
+            goto Restart;
+        }
+
         /* 'Q' or [Esc]: Quit */
-        if (KeyEvent.wVirtualKeyCode == L'Q' || KeyEvent.wVirtualKeyCode == VK_ESCAPE)
+        if (KeyEvent.wVirtualKeyCode == L'Q' ||
+            KeyEvent.wVirtualKeyCode == VK_ESCAPE)
         {
             /* We break, output a newline */
             WCHAR ch = L'\n';
@@ -425,10 +414,22 @@ Restart:
             goto Restart;
         }
 
-        /* [Enter] key: One line */
+        /* [Enter] key: Display one line */
         if (KeyEvent.wVirtualKeyCode == VK_RETURN)
         {
             Pager->ScrollRows = 1;
+            return TRUE;
+        }
+
+        /* [Space] key: Display one page */
+        if (KeyEvent.wVirtualKeyCode == VK_SPACE)
+        {
+            if (s_dwFlags & FLAG_C)
+            {
+                /* Clear the screen */
+                ConClearScreen(Pager->Screen);
+            }
+            Pager->ScrollRows = Pager->ScreenRows - 1;
             return TRUE;
         }
 
@@ -448,17 +449,23 @@ Restart:
             goto Restart;
         }
 
-        /* '=': Show current line */
+        /* '=': Show current line number */
         if (KeyEvent.uChar.UnicodeChar == L'=')
         {
             s_nPromptID = IDS_CONTINUE_LINE_AT;
             goto Restart;
         }
-    }
 
-    s_nPromptID = IDS_CONTINUE_PROGRESS;
-    chSubCommand = 0;
-    goto Restart;
+        s_nPromptID = IDS_CONTINUE_PROGRESS;
+        chSubCommand = 0;
+        goto Restart;
+    }
+    else
+    {
+        /* Extended features are unavailable: display one page */
+        Pager->ScrollRows = Pager->ScreenRows - 1;
+        return TRUE;
+    }
 }
 
 /*
--- a/base/applications/cmdutils/more/more.c
+++ b/base/applications/cmdutils/more/more.c
@@ -365,11 +365,9 @@ Restart:
     switch (chSubCommand)
     {
         case L'P':
-            chSubCommand = 0;
             Pager->ScrollRows = nLines;
             return TRUE;
         case L'S':
-            chSubCommand = 0;
             s_dwFlags |= FLAG_PLUSn;
             s_nNextLineNo = Pager->lineno + nLines;
             Pager->ScrollRows = Pager->ScreenRows - 1;
--- a/base/applications/cmdutils/more/more.c
+++ b/base/applications/cmdutils/more/more.c
@@ -144,7 +144,8 @@ MorePagerExpandTab(
 
     /* Store to buffer */
     iColumn = Pager->iColumn;
-    for (ich1 = ich2 = 0; ich1 < cch && ich2 < cch2; ++ich1)
+    // NOTE: Also, it would have been: ich2 <= cch2
+    for (ich1 = ich2 = 0; ich1 < cch /*&& ich2 < cch2*/; ++ich1)
     {
         if (line[ich1] == L'\t')
         {
@@ -278,7 +279,7 @@ Restart:
     // FIXME: Does not support TTY yet!
     for (;;)
     {
-        INPUT_RECORD ir;
+        INPUT_RECORD ir = {0};
         DWORD dwRead;
 
         do
@@ -885,9 +886,9 @@ static BOOL ParseArgument(LPCWSTR arg, BOOL *pbHasFiles)
 
 static BOOL ParseMoreVariable(BOOL *pbHasFiles)
 {
+    BOOL ret = TRUE;
     LPWSTR psz, pch;
     DWORD cch;
-    BOOL ret;
 
     cch = GetEnvironmentVariableW(L"MORE", NULL, 0);
     if (cch == 0)
@@ -936,7 +937,7 @@ int wmain(int argc, WCHAR* argv[])
     PVOID FileCacheBuffer = NULL;
     PWCHAR StringBuffer = NULL;
     DWORD StringBufferLength = 0;
-    DWORD dwReadBytes, dwReadChars;
+    DWORD dwReadBytes = 0, dwReadChars = 0;
 
     TCHAR szFullPath[MAX_PATH];
 
--- a/base/applications/cmdutils/more/more.c
+++ b/base/applications/cmdutils/more/more.c
@@ -49,7 +49,7 @@ HANDLE hStdIn, hStdOut;
 HANDLE hKeyboard;
 
 /* Enable/Disable extensions */
-BOOL s_bEnableExtensions = TRUE; // FIXME: By default, it should be FALSE.
+BOOL s_bEnableExtensions = FALSE;
 
 #define FLAG_HELP (1 << 0)
 #define FLAG_E (1 << 1)
@@ -749,17 +749,20 @@ FileGetString(
     return TRUE;
 }
 
-static VOID
-LoadRegistrySettings(HKEY hKeyRoot)
+static BOOL
+LoadRegistrySettings(HKEY hKeyRoot, BOOL bDefault)
 {
     LONG lRet;
     HKEY hKey;
-    DWORD dwType, len;
+    DWORD dwType, dwSize;
+    BOOL ret = bDefault;
     /*
      * Buffer big enough to hold the string L"4294967295",
      * corresponding to the literal 0xFFFFFFFF (MAXULONG) in decimal.
      */
-    DWORD Buffer[6];
+    WCHAR Buffer[11];
+    typedef char assertion_1[(sizeof(Buffer) >= sizeof(L"4294967295")) ? 1 : -1];
+    typedef char assertion_2[(sizeof(Buffer) >= sizeof(DWORD)) ? 1 : -1];
 
     lRet = RegOpenKeyExW(hKeyRoot,
                          L"Software\\Microsoft\\Command Processor",
@@ -767,26 +770,27 @@ LoadRegistrySettings(HKEY hKeyRoot)
                          KEY_QUERY_VALUE,
                          &hKey);
     if (lRet != ERROR_SUCCESS)
-        return;
+        return ret;
 
-    len = sizeof(Buffer);
+    dwSize = sizeof(Buffer);
     lRet = RegQueryValueExW(hKey,
                             L"EnableExtensions",
                             NULL,
                             &dwType,
                             (LPBYTE)&Buffer,
-                            &len);
+                            &dwSize);
     if (lRet == ERROR_SUCCESS)
     {
         /* Overwrite the default setting */
         if (dwType == REG_DWORD)
-            s_bEnableExtensions = !!*(PDWORD)Buffer;
+            ret = (*(PDWORD)Buffer != 0);
         else if (dwType == REG_SZ)
-            s_bEnableExtensions = (_wtol((PWSTR)Buffer) == 1);
+            ret = (_wtol(Buffer) != 0);
     }
     // else, use the default setting set globally.
 
     RegCloseKey(hKey);
+    return ret;
 }
 
 static BOOL ParseArgument(LPCWSTR arg, BOOL *pbHasFiles)
@@ -945,9 +949,17 @@ int wmain(int argc, WCHAR* argv[])
     }
 
     /* Load the registry settings */
-    LoadRegistrySettings(HKEY_LOCAL_MACHINE);
-    LoadRegistrySettings(HKEY_CURRENT_USER);
-    s_dwFlags = (s_bEnableExtensions ? FLAG_E : 0);
+    s_dwFlags = 0;
+    if (LoadRegistrySettings(HKEY_LOCAL_MACHINE, FALSE))
+    {
+        if (LoadRegistrySettings(HKEY_CURRENT_USER, TRUE))
+            s_dwFlags = FLAG_E;
+    }
+    else
+    {
+        if (LoadRegistrySettings(HKEY_CURRENT_USER, FALSE))
+            s_dwFlags = FLAG_E;
+    }
 
     // TODO: First, load the "MORE" environment variable and parse it,
     // then parse the command-line parameters.
--- a/base/applications/cmdutils/more/more.c
+++ b/base/applications/cmdutils/more/more.c
@@ -82,87 +82,16 @@ static BOOL IsBlankLine(IN PCWCH line, IN DWORD cch)
 }
 
 static BOOL __stdcall
-MorePagerExpandTab(
-    IN OUT PCON_PAGER Pager,
-    IN PCWCH line,
-    IN DWORD cch,
-    IN OUT PDWORD pdwFlags)
-{
-    PWCHAR psz;
-    DWORD ich1, ich2, cch2, spaces;
-    BOOL ret;
-    DWORD iColumn = Pager->iColumn;
-
-    /* Calculate expanded length */
-    for (ich1 = ich2 = 0; ich1 < cch; ++ich1)
-    {
-        if (line[ich1] == L'\t')
-        {
-            spaces = s_nTabWidth - (iColumn % s_nTabWidth);
-            iColumn += spaces;
-            ich2 += spaces;
-        }
-        else if (line[ich1] == L'\n')
-        {
-            iColumn = 0;
-            ++ich2;
-        }
-        else
-        {
-            ++iColumn;
-            ++ich2;
-        }
-    }
-
-    cch2 = ich2;
-    psz = malloc(cch2 * sizeof(WCHAR));
-    if (!psz)
-        return FALSE;
-
-    /* Store to buffer */
-    iColumn = Pager->iColumn;
-    for (ich1 = ich2 = 0; ich1 < cch; ++ich1)
-    {
-        if (line[ich1] == L'\t')
-        {
-            spaces = s_nTabWidth - (iColumn % s_nTabWidth);
-            iColumn += spaces;
-            while (spaces-- > 0)
-            {
-                psz[ich2++] = L' ';
-            }
-        }
-        else if (line[ich1] == L'\n')
-        {
-            iColumn = 0;
-            psz[ich2++] = line[ich1];
-        }
-        else
-        {
-            ++iColumn;
-            psz[ich2++] = line[ich1];
-        }
-    }
-
-    /* Do output */
-    ret = Pager->DefPagerLine(Pager, psz, ich2, pdwFlags);
-    free(psz);
-    Pager->iColumn = iColumn;
-    return ret;
-}
-
-static BOOL __stdcall
 MorePagerLine(
     IN OUT PCON_PAGER Pager,
     IN PCWCH line,
-    IN DWORD cch,
-    IN OUT PDWORD pdwFlags)
+    IN DWORD cch)
 {
     if (s_dwFlags & FLAG_PLUSn)
     {
         if (Pager->lineno < s_nNextLineNo)
         {
-            *pdwFlags &= ~CON_PAGER_LINE_FLAG_NEWLINE;
+            Pager->dwFlags &= ~CON_PAGER_FLAG_NEWLINE;
             s_bPrevLineIsBlank = IsBlankLine(line, cch);
             return TRUE; /* Don't output */
         }
@@ -175,7 +104,7 @@ MorePagerLine(
         {
             if (s_bPrevLineIsBlank)
             {
-                *pdwFlags &= ~CON_PAGER_LINE_FLAG_NEWLINE;
+                Pager->dwFlags &= ~CON_PAGER_FLAG_NEWLINE;
                 return TRUE; /* Don't output */
             }
             s_bPrevLineIsBlank = TRUE;
@@ -187,7 +116,7 @@ MorePagerLine(
     }
 
     s_nNextLineNo = 0;
-    return MorePagerExpandTab(Pager, line, cch, pdwFlags);
+    return Pager->DefPagerLine(Pager, line, cch);
 }
 
 static BOOL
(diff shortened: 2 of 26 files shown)


katahiromz:
無題
無題2
Working well. Check time.

katahiromz:
On ReactOS:
VirtualBox_ReactOS_12_05_2021_12_07_00
VirtualBox_ReactOS_12_05_2021_12_07_28
It works.

katahiromz:
Check time.

sdk/lib/conutils/pager.c
```diff
@@ -31,7 +33,86 @@
...
+    DWORD ScrollRows = ScreenRows - 1, ichLast = ich, MaxRows = ScrollRows;
+    UINT nWidthOfChar = 1, nCodePage = GetConsoleOutputCP();
+    BOOL IsCJK = FALSE, IsDoubleWidthCharTrailing = FALSE;
+
+    if (ich >= cch)
+        return TRUE;
+
+    switch (nCodePage)
+    {
+        case 936: case 950: case 932: case 949:
```
SergeGautherie, line 60: Sort them? One per line?

HBelusca (commented):
Note: I haven't finished my review yet.

sdk/lib/conutils/pager.c
```diff
@@ -31,7 +33,86 @@
 // Temporary HACK
 #define CON_STREAM_WRITE    ConStreamWrite
 
+static inline INT GetWidthOfCharCJK(UINT nCodePage, WCHAR ch)
```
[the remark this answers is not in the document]
katahiromz: OK.

sdk/lib/conutils/pager.c
```diff
@@ -31,7 +33,86 @@
...
+}
+
+static BOOL ConPagerAction(PCON_PAGER Pager)
+{
+    PCTCH TextBuff = Pager->TextBuff;
+    DWORD ich = Pager->ich, cch = Pager->cch, iLine = Pager->iLine, iColumn;
+    DWORD ScreenColumns = Pager->ScreenColumns, ScreenRows = Pager->ScreenRows;
+    DWORD ScrollRows = ScreenRows - 1, ichLast = ich, MaxRows = ScrollRows;
+    UINT nWidthOfChar = 1, nCodePage = GetConsoleOutputCP();
+    BOOL IsCJK = FALSE, IsDoubleWidthCharTrailing = FALSE;
```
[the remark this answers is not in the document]
katahiromz: OK.

sdk/lib/conutils/pager.c
```diff
@@ -31,7 +33,86 @@
...
+
+    switch (nCodePage)
+    {
+        case 936: case 950: case 932: case 949:
+        {
+            /* Chinese, Japanese or Korean. They uses double-width characters. */
+            IsCJK = TRUE;
+            break;
+        }
+    }
```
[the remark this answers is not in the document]
katahiromz: OK.

base/applications/cmdutils/more/more.c
```diff
@@ -156,6 +156,20 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
...
 
+    /* [Enter] key: One line go */
+    if ((KeyEvent.wVirtualKeyCode == VK_RETURN) &&
+        !(KeyEvent.dwControlKeyState & (LEFT_CTRL_PRESSED | RIGHT_CTRL_PRESSED)))
+    {
+        Pager->PagerAction = CON_PAGER_ACTION_SHOW_LINE;
+        return TRUE;
+    }
+
+    if (KeyEvent.dwControlKeyState & (LEFT_CTRL_PRESSED | RIGHT_CTRL_PRESSED))
```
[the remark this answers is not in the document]
katahiromz: Simplified by using `fCtrlPressed` variable.

katahiromz:
If the pager actions were implemented in `conutils`, ` conutils` as a static library becomes bloated and the total size of `cmdutils` increases. @HBelusca How do you think?

HBelusca:
> If the pager actions were implemented in `conutils`, ` conutils` as a static library becomes bloated and the total size of `cmdutils` increases. @HBelusca How do you think?

I was told that normally, the compiler is intelligent enough at link time to see which functions from the library are being used and those that are not, and those that are not are being rid of from the resulting executable. So for example, a console app that uses some conutils features but, say, not the pager functionality, would at the end not contain the pager code.

katahiromz:
Paging function is used in `cmd` module:
https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=Con%5BA-Za-z%5D%2BPaging&sr=1

HBelusca:
> The paging function is used in `cmd` module:
> https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=Con%5BA-Za-z%5D%2BPaging&sr=1

Well yes I know that, since it's me who moved it from there into a library since it was also going to be used e.g. in MORE and others...

katahiromz:
無題

katahiromz:
Check time.

katahiromz:
Can we disable extended features of `MORE` command?

sdk/lib/conutils/pager.c
```diff
@@ -31,7 +33,115 @@
...
+    IsCJK = IsCJKCodePage(nCodePage);
+
+    for (; ich < cch && iLine < ScrollRows; ++ich)
+    {
+        Pager->lineno = lineno;
+        if (IsCJK)
+        {
+            nWidthOfChar = GetWidthOfCharCJK(nCodePage, TextBuff[ich]);
+            IsDoubleWidthCharTrailing = (nWidthOfChar == 2) &&
+                                        (iColumn + 1 == ScreenColumns);
```
HBelusca, line 102:
```suggestion
                                        (iColumn + 1 >= ScreenColumns);
```

base/applications/cmdutils/more/more.c
```diff
@@ -134,28 +312,103 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
...
+            WCHAR ch = L'\n';
+            ConStreamWrite(Pager->Screen->Stream, &ch, 1);
+            return FALSE;
+        }
+
+        /* 'F': Next file */
+        if ((KeyEvent.wVirtualKeyCode == L'F') && !fCtrl)
+        {
+            Pager->PagerAction = MorePageActionNextFile;
+            return TRUE;
```
HBelusca, line 359: When you do: `MORE file1 file2` and you are viewing `file1` pressing "F" for going to next file should:
- stop the paging of the current file (so, `PagePrompt` should return FALSE),
- detect in the pager's caller (in main()) that paging didn't stop prematurely because we want to quit, but because we want to go to the next file
- go to the next file.

base/applications/cmdutils/more/more.c
```diff
@@ -116,11 +227,78 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
...
+                        if (nLines == 0 && ch == L'0')
+                            continue;
+                        nLines *= 10;
+                        nLines += ch - L'0';
+                        ConStreamWrite(Pager->Screen->Stream, &ch, 1);
+                        continue;
+                    }
+                    if (KeyEvent.wVirtualKeyCode == VK_RETURN)
+                    {
+                        if (s_chSubCommand != L'P' && s_chSubCommand != L'S')
```
HBelusca, line 265: This is already FALSE

base/applications/cmdutils/more/more.c
```diff
@@ -134,28 +312,103 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
...
+            return TRUE;
+        }
+    }
+
+    /* [Space] key: One page */
+    if ((KeyEvent.wVirtualKeyCode == VK_SPACE) && !fCtrl)
+    {
+        Pager->ScrollRows = Pager->ScreenRows - 1;
+        return TRUE;
     }
```
HBelusca, line 409: Please move this before the treatment of the extended features,

base/applications/cmdutils/more/more.c
```diff
@@ -500,8 +754,102 @@ int wmain(int argc, WCHAR* argv[])
         return 1;
     }
 
+    /* Parse flags */
+    HasFiles = FALSE;
+    for (i = 1; i < argc; i++)
+    {
+        if (argv[i][0] == L'/' || argv[i][0] == L'-')
```
HBelusca, line 761: We are creating a clone of a Windows command, not a linux one.
```suggestion
        if (argv[i][0] == L'/')
```
example on Windows:
```
C:\Temp>more -e
Cannot access file C:\Temp\-e
```

base/applications/cmdutils/more/more.c
```diff
@@ -500,8 +754,102 @@ int wmain(int argc, WCHAR* argv[])
...
+        }
+        else if (argv[i][0] == L'+')
+        {
+            if (L'0' <= argv[i][1] && argv[i][1] <= L'9')
+            {
+                LPWSTR endptr;
+                s_dwFlags |= FLAG_PLUSn;
+                s_nNextLineNo = wcstoul(&argv[i][1], &endptr, 10);
+                if (*endptr == 0)
+                    continue;
```
HBelusca, line 822: Does that make it behave as follows?
```
C:\Temp>more /e +5d somefile.txt
Cannot access file C:\Temp\+5d
```

base/applications/cmdutils/more/more.c
```diff
@@ -500,8 +754,102 @@ int wmain(int argc, WCHAR* argv[])
...
+                    }
+                    break;
+                case L'T':
+                    if (L'0' <= argv[i][2] && argv[i][2] <= L'9')
+                    {
+                        LPWSTR endptr;
+                        s_dwFlags |= FLAG_Tn;
+                        s_nTabWidth = wcstoul(&argv[i][2], &endptr, 10);
+                        if (*endptr == 0)
+                            continue;
```
HBelusca, line 807: Does that make it behave as follows?
```
C:\Temp>more /e /T5d somefile.txt
Invalid parameter - /T5d
```

base/applications/cmdutils/more/more.c
```diff
@@ -500,8 +754,102 @@ int wmain(int argc, WCHAR* argv[])
...
+    {
+        ConResPuts(StdOut, IDS_USAGE);
+        return 0;
+    }
+
+    if (s_dwFlags & FLAG_C)
+    {
+        /* Clear the screen */
+        ConClearScreen(&Screen);
+    }
```
HBelusca, line 847: This has to be done, not here, but **every time** a new page (not one line) is to be displayed.

base/applications/cmdutils/more/more.c
```diff
@@ -134,28 +312,103 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
...
+            return TRUE;
+        }
+
+        /* 'P': Display n lines */
+        if (KeyEvent.wVirtualKeyCode == L'P' && !fCtrl)
+        {
+            Pager->PagerAction = MorePageActionDoNothing;
+            s_nPromptID = IDS_CONTINUE_LINES;
+            s_chSubCommand = L'P';
+            return TRUE;
```
HBelusca, line 383: Just jump back at the "top" of the PagePrompt function.

base/applications/cmdutils/more/more.c
```diff
@@ -134,28 +312,103 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
...
+        }
+
+        /* 'S': Skip n lines */
+        if (KeyEvent.wVirtualKeyCode == L'S' && !fCtrl)
+        {
+            Pager->PagerAction = MorePageActionDoNothing;
+            s_nPromptID = IDS_CONTINUE_LINES;
+            s_chSubCommand = L'S';
+            return TRUE;
```
HBelusca, line 392: Just jump back at the "top" of the PagePrompt function.

base/applications/cmdutils/more/more.c
```diff
@@ -134,28 +312,103 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
...
+        }
+
+        /* '=': Show current line */
+        if (KeyEvent.uChar.UnicodeChar == L'=' && !fCtrl)
+        {
+            Pager->PagerAction = MorePageActionDoNothing;
+            s_nPromptID = IDS_CONTINUE_LINE_AT;
+            return TRUE;
```
HBelusca, line 400: Just jump back at the "top" of the PagePrompt function.

base/applications/cmdutils/more/more.c
```diff
@@ -134,28 +312,103 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
...
+            Pager->PagerAction = MorePageActionNextFile;
+            return TRUE;
+        }
+
+        /* '?': Show Options */
+        if (KeyEvent.uChar.UnicodeChar == L'?' && !fCtrl)
+        {
+            Pager->PagerAction = MorePageActionDoNothing;
+            s_nPromptID = IDS_CONTINUE_OPTIONS;
+            return TRUE;
```
HBelusca, line 367: Just jump back at the "top" of the PagePrompt function.

base/applications/cmdutils/more/more.c
```diff
@@ -22,6 +22,7 @@
  *     use window size instead of buffer size.
  */
 
+#include <stdlib.h>
 #include <stdio.h>
```
HBelusca, line 26:
```suggestion
#include <stdio.h>
#include <stdlib.h>
```

HBelusca (commented):
Please check commit 90b3a9c for getting the default settings for the enable-extensions functionality.

HBelusca:
Also, the MORE utility supports retrieving default settings from a `"MORE"` environment variable, that is a list of whitespace-separated command switches.

base/applications/cmdutils/more/more.c
```diff
@@ -116,11 +227,78 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
...
+                        SetConsoleCursorPosition(hOutput, csbi.dwCursorPosition);
+                        continue;
+                    }
+                    break;
+                }
+                goto skip;
+            }
+        }
+    }
+skip:
```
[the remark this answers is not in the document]
katahiromz: OK.

base/applications/cmdutils/more/more.c
```diff
@@ -134,28 +312,103 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
...
-    if ((KeyEvent.wVirtualKeyCode == L'Q') &&
-        !(KeyEvent.dwControlKeyState & (LEFT_CTRL_PRESSED | RIGHT_CTRL_PRESSED)))
+    if (s_dwFlags & FLAG_E) /* If extended features are available */
     {
-        /* We break, output a newline */
-        WCHAR ch = L'\n';
-        ConStreamWrite(Pager->Screen->Stream, &ch, 1);
-        return FALSE;
+        /* 'Q': Quit */
+        if ((KeyEvent.wVirtualKeyCode == L'Q') && !fCtrl)
```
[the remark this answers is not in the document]
katahiromz: OK.

sdk/lib/conutils/pager.c
```diff
@@ -70,48 +173,43 @@ ConWritePaging(
...
     /* Make sure the user doesn't have the screen too small */
-    if (ScreenLines < 4)
+    if (Pager->ScreenRows < 4)
     {
         CON_STREAM_WRITE(Pager->Screen->Stream, szStr, len);
         return TRUE;
     }
 
-    while (i < len)
+    while (!(*Pager->PagerAction)(Pager))
```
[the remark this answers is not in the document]
katahiromz: OK.

sdk/lib/conutils/pager.c
```diff
@@ -70,48 +173,43 @@ ConWritePaging(
...
+        Pager->ScrollRows = Pager->ScreenRows - 1;
+
+    if (len == 0 || szStr == NULL)
+        return TRUE;
 
     /* Make sure the user doesn't have the screen too small */
-    if (ScreenLines < 4)
+    if (Pager->ScreenRows < 4)
     {
         CON_STREAM_WRITE(Pager->Screen->Stream, szStr, len);
```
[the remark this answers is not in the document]
katahiromz: OK.

sdk/lib/conutils/pager.h
```diff
@@ -26,27 +26,39 @@
...
+    DWORD ScrollRows;
+    CON_PAGER_ACTION_FN PagerAction; /* The action function */
+    CON_PAGER_LINE_FN PagerLine; /* The line function */
+    CON_PAGER_LINE_FN DefPagerLine; /* Default line function */
+    PCTCH TextBuff; /* The text buffer */
+    DWORD ich; /* The current index of character */
+    DWORD cch; /* The total number of characters */
+    DWORD iColumn; /* The current index of column */
+    DWORD iLine; /* The output line count */
+    DWORD lineno; /* The line number */
```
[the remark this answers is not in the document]
katahiromz: `lineno` is logical line number of the actual file.
`iLine` is physical outputted line count of the screen.

```c
            if (dwFlags & CON_PAGER_LINE_FLAG_NEWLINE)
                ++iLine;
            if (TextBuff[ich] == TEXT('\n'))
                ++lineno;
```

sdk/lib/conutils/pager.c
katahiromz: `iColumn + 2 == ScreenColumns` is not the case of this.

HBelusca: With the `>=` this makes the code more robust.

base/applications/cmdutils/more/more.c
katahiromz: OK.

base/applications/cmdutils/more/more.c
katahiromz: OK.

base/applications/cmdutils/more/more.c
katahiromz: OK.

base/applications/cmdutils/more/more.c
katahiromz: OK.

base/applications/cmdutils/more/more.c
katahiromz: OK.

base/applications/cmdutils/more/more.c
katahiromz: OK.

base/applications/cmdutils/more/more.c
```diff
@@ -116,11 +227,78 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
...
+                        ch = L' ';
+                        ConStreamWrite(Pager->Screen->Stream, &ch, 1);
+                        SetConsoleCursorPosition(hOutput, csbi.dwCursorPosition);
+                        continue;
+                    }
+                    break;
+                }
+                goto skip;
+            }
+        }
```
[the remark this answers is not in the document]
katahiromz: OK.

base/applications/cmdutils/more/more.c
```diff
@@ -46,6 +47,109 @@ HANDLE hFile = INVALID_HANDLE_VALUE;
...
+        if (!(wType & (C1_BLANK | C1_SPACE)))
+            return FALSE;
+    }
+    return TRUE;
+}
+
+static BOOL CALLBACK
+MorePagerExpandTab(PCON_PAGER Pager, LPCWSTR line, DWORD cch, DWORD *pdwFlags)
+{
+    WCHAR sz2[512];
```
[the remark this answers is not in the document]
katahiromz: OK.

base/applications/cmdutils/more/more.c
katahiromz: How to detect?

base/applications/cmdutils/more/more.c
katahiromz: 無題

base/applications/cmdutils/more/more.c
katahiromz: 無題

base/applications/cmdutils/more/more.c
katahiromz: Moved.

base/applications/cmdutils/more/more.c
katahiromz: OK.

sdk/lib/conutils/pager.c
katahiromz: If `iColumn + 2 == ScreenColumns`, then the double-width character will fit the screen edge and it doesn't overlap.

base/applications/cmdutils/more/more.c
HBelusca: This `PagePrompt()` function could set some global (in more.c) `BOOL bDoNextFile;` variable to TRUE.
Then, in the `main()` function, at the place where the following check is done:
```c
            /* If we Ctrl-C/Ctrl-Break, stop everything */
            if (!bContinue)
            {
                CloseHandle(hFile);
                goto Quit;
            }
```
you could modify it into something like:
```c
            /* Check whether we should stop displaying this file */
            if (!bContinue)
            {
                if (bDoNextFile)
                {
                    /* Bail out and continue with the other files */
                    bDoNextFile = FALSE; // Reset the flag.
                    break;
                }

                /* We Ctrl-C/Ctrl-Break, stop everything */
                CloseHandle(hFile);
                goto Quit;
            }
```

base/applications/cmdutils/more/more.c
katahiromz: 無題
OK.

base/applications/cmdutils/more/more.c
```diff
@@ -134,28 +356,113 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
...
+            Pager->PagerAction = MorePageActionNextFile;
+            return TRUE;
+        }
+
+        /* '?': Show Options */
+        if (KeyEvent.uChar.UnicodeChar == L'?')
+        {
+            Pager->PagerAction = MorePageActionDoNothing;
+            s_nPromptID = IDS_CONTINUE_OPTIONS;
+            return TRUE;
```
HBelusca, line 431: Does the following work? :
```suggestion
            s_nPromptID = IDS_CONTINUE_OPTIONS;
            goto top;
```

base/applications/cmdutils/more/more.c
```diff
@@ -46,6 +47,159 @@ HANDLE hFile = INVALID_HANDLE_VALUE;
...
+            return FALSE;
+    }
+    return TRUE;
+}
+
+static BOOL CALLBACK
+MorePagerExpandTab(PCON_PAGER Pager, LPCWSTR line, DWORD cch, DWORD *pdwFlags)
+{
+    LPWSTR psz;
+    DWORD ich1, ich2, cch2, spaces, iColumn = Pager->iColumn;
```
HBelusca, line 111:
```suggestion
    DWORD ich1, ich2, cch2, spaces,
    DWORD iColumn = Pager->iColumn;
```

base/applications/cmdutils/more/more.c
```diff
@@ -54,6 +208,13 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
     HANDLE hInput = ConStreamGetOSHandle(StdIn);
     DWORD dwMode;
     KEY_EVENT_RECORD KeyEvent;
+    BOOL fCtrl;
+    DWORD nLines;
+    WCHAR ch;
+    static UINT s_nPromptID = IDS_CONTINUE_PROGRESS;
+    static WCHAR s_chSubCommand = 0;
```
HBelusca, line 215: I'm not a fan of having these as static variables **within** the function.
Indeed, if paging display is cancelled early for this file and another file is going to be displayed, how are you going to reset these variables to their default values?

base/applications/cmdutils/more/more.c
```diff
@@ -54,6 +208,13 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
...
+top:
```
HBelusca, line 216:
```suggestion

top:
```
Also, rename "top" into "Restart" or something more understandable / less generic.

base/applications/cmdutils/more/more.c
katahiromz: OK.

base/applications/cmdutils/more/more.c
katahiromz: OK.

base/applications/cmdutils/more/more.c
katahiromz: OK.

base/applications/cmdutils/more/more.c
katahiromz: OK.

base/applications/cmdutils/more/more.c
katahiromz: OK.

sdk/lib/conutils/pager.c
HBelusca: ok thanks for the explanation.

katahiromz:
Check time.

boot/bootdata/hivesft.inf
```diff
@@ -469,6 +469,7 @@ HKCU,"SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\BitBucket\Volume",,0x00
 
 ; CMD Settings
 HKLM,"SOFTWARE\Microsoft\Command Processor","AutoRun",0x00020000,""
+HKLM,"SOFTWARE\Microsoft\Command Processor","EnableExtensions",0x00010003,1
```
HBelusca, line 472: Please don't, yet, because this will currently have side-effects on CMD.EXE that I haven't yet fully resolved.

base/applications/cmdutils/more/more.c
```diff
@@ -945,9 +949,17 @@ int wmain(int argc, WCHAR* argv[])
...
+    if (LoadRegistrySettings(HKEY_LOCAL_MACHINE, FALSE))
+    {
+        if (LoadRegistrySettings(HKEY_CURRENT_USER, TRUE))
+            s_dwFlags = FLAG_E;
+    }
+    else
+    {
+        if (LoadRegistrySettings(HKEY_CURRENT_USER, FALSE))
+            s_dwFlags = FLAG_E;
+    }
```
HBelusca: The previous code was behaving correctly, so why did you change it?

base/applications/cmdutils/more/more.c
```diff
@@ -749,44 +749,48 @@ FileGetString(
...
+    DWORD dwType, dwSize;
+    BOOL ret = bDefault;
     /*
      * Buffer big enough to hold the string L"4294967295",
      * corresponding to the literal 0xFFFFFFFF (MAXULONG) in decimal.
      */
-    DWORD Buffer[6];
+    WCHAR Buffer[11];
+    typedef char assertion_1[(sizeof(Buffer) >= sizeof(L"4294967295")) ? 1 : -1];
+    typedef char assertion_2[(sizeof(Buffer) >= sizeof(DWORD)) ? 1 : -1];
```
HBelusca, line 763: wtf is that.

HBelusca, line 763: wtf is that.

base/applications/cmdutils/more/more.c
```diff
@@ -49,7 +49,7 @@ HANDLE hStdIn, hStdOut;
 HANDLE hKeyboard;
 
 /* Enable/Disable extensions */
-BOOL s_bEnableExtensions = TRUE; // FIXME: By default, it should be FALSE.
+BOOL s_bEnableExtensions = FALSE;
```
HBelusca: See comment below regarding having EnableExtensions in the registry, yet.

base/applications/cmdutils/more/more.c
```diff
@@ -749,44 +749,48 @@ FileGetString(
...
-            s_bEnableExtensions = !!*(PDWORD)Buffer;
+            ret = (*(PDWORD)Buffer != 0);
         else if (dwType == REG_SZ)
-            s_bEnableExtensions = (_wtol((PWSTR)Buffer) == 1);
+            ret = (_wtol(Buffer) != 0);
     }
     // else, use the default setting set globally.
 
     RegCloseKey(hKey);
+    return ret;
```
HBelusca: Not good, please revert (except for the `DWORD Buffer[6];` --> `WCHAR Buffer[11];` that is correct).

base/applications/cmdutils/more/lang/bg-BG.rc
```diff
@@ -17,15 +17,11 @@ Options:\n\
...
-   =        Show line number.\n\
-   ?        Show help line.\n\
-   <Space>  Display next page.\n\
-   <Enter>  Display next line.\n\
-   <Esc>    Same as Q.\n"
+   P n      Display n lines.        <Space>  Display next page.\n\
+   S n      Skip n lines.           <Enter>  Display next line.\n\
+   F        Display next file.      <Esc>    Same as Q.\n\
+   Q        Quit.                   =        Show line number.\n\
+   ?        Show help line.\n"
```
HBelusca, line 24: Previous formatting might have been better, because it's not in all languages that these descriptions remain as short as that, and now for these languages one would need to change that formatting.

boot/bootdata/hivesft.inf
katahiromz: Reverted.

base/applications/cmdutils/more/more.c
katahiromz: Reverted.

base/applications/cmdutils/more/more.c
katahiromz: OK.

base/applications/cmdutils/more/more.c
katahiromz: Reverted.

base/applications/cmdutils/more/more.c
katahiromz: Reverted.

base/applications/cmdutils/more/lang/bg-BG.rc
katahiromz: Reverted, too.

katahiromz:
Check time.

base/applications/cmdutils/more/more.c
```diff
@@ -553,8 +971,27 @@ int wmain(int argc, WCHAR* argv[])
         return 1;
     }
 
+    // First, load the "MORE" environment variable and parse it,
+    // then parse the command-line parameters.
```
HBelusca, line 975:
```suggestion
    /* First, load the "MORE" environment variable and parse it,
     * then parse the command-line parameters. */
```

base/applications/cmdutils/more/more.c
```diff
@@ -529,12 +951,9 @@ int wmain(int argc, WCHAR* argv[])
...
     LoadRegistrySettings(HKEY_CURRENT_USER);
+    if (bEnableExtensions)
+        s_dwFlags |= FLAG_E;
 
-    // TODO: First, load the "MORE" environment variable and parse it,
-    // then parse the command-line parameters.
-
-    // FIXME: Parse all the remaining parameters.
-    // Then the file list can be found at the very end.
     // FIXME2: Use the PARSER api that can be found in EVENTCREATE.
```
HBelusca, line 957: Regarding that FIXME2: don't touch it, just keep the current command-line parsing as you are doing here.

base/applications/cmdutils/more/more.c
```diff
@@ -456,7 +761,9 @@ LoadRegistrySettings(HKEY hKeyRoot)
      * Buffer big enough to hold the string L"4294967295",
      * corresponding to the literal 0xFFFFFFFF (MAXULONG) in decimal.
      */
-    DWORD Buffer[6];
+    WCHAR Buffer[11];
+    C_ASSERT(sizeof(Buffer) >= sizeof(L"4294967295"));
```
HBelusca, line 765: The following might be better:
```suggestion
    WCHAR Buffer[sizeof("4294967295")];
```
It automatically ensures the correct size (the sizeof will also count the null-terminator) and makes the intent more concise.

base/applications/cmdutils/more/more.c
```diff
@@ -456,7 +761,9 @@ LoadRegistrySettings(HKEY hKeyRoot)
...
+    C_ASSERT(sizeof(Buffer) >= sizeof(DWORD));
```
HBelusca, line 766: ok

base/applications/cmdutils/more/more.c
```diff
@@ -121,11 +285,69 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
...
+        if ((KeyEvent.wVirtualKeyCode == VK_SHIFT) ||
+            (KeyEvent.wVirtualKeyCode == VK_MENU)  ||
+            (KeyEvent.wVirtualKeyCode == VK_CONTROL))
+        {
+            continue;
+        }
+
+        /* Ignore any unsupported subcommand */
+        if (s_chSubCommand != L'P' && s_chSubCommand != L'S')
+            break;
```
HBelusca, line 310: Technically if your code runs fine, the `s_chSubCommand` variable must only take the values '0' (NULL character), 'P' or 'S' and nothing more, so that might just need to become some sort of assertion on the value of `s_chSubCommand` (see also comments below).

base/applications/cmdutils/more/more.c
```diff
@@ -121,11 +285,69 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
...
+            if (csbi.dwCursorPosition.X > 0)
+                csbi.dwCursorPosition.X = csbi.dwCursorPosition.X - 1;
+            SetConsoleCursorPosition(hOutput, csbi.dwCursorPosition);
+
+            ch = L' ';
+            ConStreamWrite(Pager->Screen->Stream, &ch, 1);
+            SetConsoleCursorPosition(hOutput, csbi.dwCursorPosition);
+            continue;
+        }
+    }
```
HBelusca, line 350:
```suggestion
    }

```

base/applications/cmdutils/more/more.c
```diff
@@ -121,11 +285,69 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
...
+        }
+        if (KeyEvent.wVirtualKeyCode == VK_RETURN)
+        {
+            break;
+        }
+        else if (KeyEvent.wVirtualKeyCode == VK_ESCAPE)
+        {
+            s_chSubCommand = L'-';
+            break;
+        }
```
HBelusca, line 330: If your intent is to just reset the paging prompt without doing anything else (which seems to be the case with Windows' MORE, only when the current prompt is the "OPTIONS"/help one), then it should be better to just reset `s_chSubCommand` to zero, remain within this PagePrompt function, refresh the displayed prompt and wait for keyboard, without using this sort of "hack" of setting the command to '-' and using the hackish MorePageActionDoNothing stuff below.

base/applications/cmdutils/more/more.c
```diff
@@ -139,28 +361,112 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
...
+        case L'S':
+            s_chSubCommand = 0;
+            s_dwFlags |= FLAG_PLUSn;
+            s_nNextLineNo = Pager->lineno + nLines;
+            Pager->ScrollRows = Pager->ScreenRows - 1;
+            return TRUE;
+        case L'-':
+            s_chSubCommand = 0;
+            Pager->PagerAction = MorePageActionDoNothing;
+            return TRUE;
```
HBelusca, line 379: See my comment above about that '-' hack and the VK_ESCAPE handling.

base/applications/cmdutils/more/more.c
```diff
@@ -139,28 +361,112 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
...
+    if (fCtrl)
     {
-        /* We break, output a newline */
-        WCHAR ch = L'\n';
-        ConStreamWrite(Pager->Screen->Stream, &ch, 1);
-        return FALSE;
+        /* Do nothing */
+        Pager->PagerAction = MorePageActionDoNothing;
+        return TRUE;
+    }
```
HBelusca, line 398: I'm not convinced by the pertinence of this part of code. The original idea of the PagePrompt() function was that it does all the prompt management and there is no need at all to quit that function and tell the paging support to "not do anything".

base/applications/cmdutils/more/more.c
```diff
@@ -139,28 +361,112 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
...
+
+        /* '=': Show current line */
+        if (KeyEvent.uChar.UnicodeChar == L'=')
+        {
+            s_nPromptID = IDS_CONTINUE_LINE_AT;
+            goto Restart;
+        }
     }
 
+    Pager->PagerAction = MorePageActionDoNothing;
```
HBelusca, line 469: See comment above.

base/applications/cmdutils/more/more.c
katahiromz: OK.

base/applications/cmdutils/more/more.c
katahiromz: OK.

base/applications/cmdutils/more/more.c
katahiromz: OK.

base/applications/cmdutils/more/more.c
katahiromz: OK.

base/applications/cmdutils/more/more.c
katahiromz: OK.

base/applications/cmdutils/more/more.c
katahiromz: Deleted pager actions. OK.

base/applications/cmdutils/more/more.c
katahiromz: OK.

base/applications/cmdutils/more/more.c
katahiromz: OK.

base/applications/cmdutils/more/more.c
katahiromz: Insert an assertion. OK.

katahiromz:
Check time.

HBelusca:
Adding separate commits: they have been tested, and it's faster than pressing every time the "suggestion" stuff in github. You can look at them separately, and keep them as they are, or modify them, etc.

base/applications/cmdutils/more/lang/bg-BG.rc
```diff
@@ -3,13 +3,35 @@ LANGUAGE LANG_BULGARIAN, SUBLANG_DEFAULT
...
+   =        Show line number.\n\
+   <Space>  Display next page.\n\
+   <Enter>  Display next line.\n\
+   <Esc>    Same as Q.\n"
+
     IDS_CONTINUE "-- Продължаване --"
     IDS_CONTINUE_PROGRESS "-- Продължаване (%d%%) --"
     IDS_FILE_ACCESS "Няма достъп до файл %s.\n"
+    IDS_BAD_FLAG "Invalid argument - '%ls'\n"
+    IDS_CONTINUE_OPTIONS "-- Continue (%d%%) [Options: psfq=<Space><Enter><Esc>] --"
```
HBelusca, line 34: Here the format of the prompts is compatible with MS ones.

sdk/lib/conutils/pager.c
```diff
@@ -31,7 +33,127 @@
...
+        ret = 2;
+    return ret;
+}
+
+static BOOL __stdcall
+ConDefaultPagerLine(
+    IN OUT PCON_PAGER Pager,
+    IN PCTCH line,
+    IN DWORD cch,
+    IN OUT PDWORD pdwFlags)
```
HBelusca, line 66: (This function and the other ones as well): Parameter list formatted as in the rest of the conutils library.

HBelusca: Also, using "PCTCH" as in the rest of the lib: pointer to const "TCHAR" (note: there is no "PCTCHAR" type), as the buffer is NOT a null-terminated string (like what's understood with "PCTSTR" and the like), but a buffer of (T)chars.

sdk/lib/conutils/pager.h
```diff
@@ -26,27 +26,40 @@
...
 
-
 // #include <wincon.h>
 
+struct _CON_PAGER;
+typedef BOOL (__stdcall *CON_PAGER_LINE_FN)(
+    IN OUT struct _CON_PAGER *Pager,
+    IN PCTCH line,
+    IN DWORD cch,
+    IN OUT PDWORD pdwFlags);
```
HBelusca, line 36: This has been modified in accordance too.

base/applications/cmdutils/more/more.c
```diff
@@ -51,6 +52,144 @@ HANDLE hKeyboard;
...
+    }
+    return TRUE;
+}
+
+static BOOL __stdcall
+MorePagerExpandTab(
+    IN OUT PCON_PAGER Pager,
+    IN PCWCH line,
+    IN DWORD cch,
+    IN OUT PDWORD pdwFlags)
```
HBelusca, line 89: Similarly done as in the conutils (and the rest of MORE, as it originally was).

base/applications/cmdutils/more/more.c
```diff
@@ -51,6 +52,144 @@ HANDLE hKeyboard;
...
+{
+    PWCHAR psz;
```
HBelusca, line 91: This is again a character buffer, not a (NULL)-terminated string.

base/applications/cmdutils/more/more.c
```diff
@@ -51,6 +52,144 @@ HANDLE hKeyboard;
...
+        }
+        else
+        {
+            ++iColumn;
+            ++ich2;
+        }
+    }
+
+    cch2 = ich2;
+    psz = malloc((cch2 + 1) * sizeof(WCHAR));
```
HBelusca, line 118: In principle, no need to reserve for a terminating NULL, since the pager functions take counted buffers.

base/applications/cmdutils/more/more.c
```diff
@@ -51,6 +52,144 @@ HANDLE hKeyboard;
...
+    if (!psz)
+        return FALSE;
+
+    /* Store to buffer */
+    iColumn = Pager->iColumn;
+    // NOTE: Also, it would have been: ich2 <= cch2
+    for (ich1 = ich2 = 0; ich1 < cch /*&& ich2 < cch2*/; ++ich1)
```
HBelusca, line 125: Normally, from the previous for-loop, your ich2 variable can go up to (and including) cch2. So either you keep the explicit condition in this for loop (but it then needs to be fixed), or, you can remove it altogether since it appears this for-loop parses in the same way as in the previous one (that calculated cch2).

base/applications/cmdutils/more/more.c
```diff
@@ -51,6 +52,144 @@ HANDLE hKeyboard;
...
+        }
+        else if (line[ich1] == L'\n')
+        {
+            iColumn = 0;
+            psz[ich2++] = line[ich1];
+        }
+        else
+        {
+            ++iColumn;
+            psz[ich2++] = line[ich1];
```
HBelusca, line 144: Note: For some reason, code analyzer says here:
```
Warning C6386: Buffer overrun while writing to 'psz':  the writable size is '((cch2+1))*sizeof(WCHAR)' bytes, but '4' bytes might be written.
```
My confidence in this is quite low, since this for-loop appears to exactly do the same thing as the one above, and the psz buffer should be correctly sized.

base/applications/cmdutils/more/more.c
```diff
@@ -59,6 +198,12 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
     HANDLE hInput = ConStreamGetOSHandle(StdIn);
     DWORD dwMode;
     KEY_EVENT_RECORD KeyEvent;
+    BOOL fCtrl;
+    DWORD nLines;
+    WCHAR chSubCommand = 0;
```
HBelusca, line 203: The chSubCommand can now be moved within the function that uses it.

base/applications/cmdutils/more/more.c
```diff
@@ -106,25 +253,92 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
...
+                      (KeyEvent.wVirtualKeyCode == L'C')))
+        {
+            chSubCommand = 0;
+            break;
+        }
+
+        /* If extended features are unavailable, or no
+         * pending commands, don't do more processing. */
+        if (!(s_dwFlags & FLAG_E) || (chSubCommand == 0))
+            break;
```
HBelusca, line 294: This now breaks from the loop early, if the extensions were disabled.

base/applications/cmdutils/more/more.c
```diff
@@ -106,25 +253,92 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
...
+        WCHAR ch;
+
         do
         {
             ReadConsoleInput(hInput, &ir, 1, &dwRead);
         }
         while ((ir.EventType != KEY_EVENT) || (!ir.Event.KeyEvent.bKeyDown));
 
-        /* Got our key, return to caller */
+        /* Got our key */
```
HBelusca, line 269: Comment is fixed.

base/applications/cmdutils/more/more.c
```diff
@@ -106,25 +253,92 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
...
         KeyEvent = ir.Event.KeyEvent;
+
+        /* Ignore any unsupported keyboard press */
+        if ((KeyEvent.wVirtualKeyCode == VK_SHIFT) ||
+            (KeyEvent.wVirtualKeyCode == VK_MENU)  ||
+            (KeyEvent.wVirtualKeyCode == VK_CONTROL))
+        {
+            continue;
+        }
```
HBelusca, line 278: This has been moved up compared to what was in your code, to have a behaviour closer to the original (original in master) one.

base/applications/cmdutils/more/more.c
```diff
@@ -140,28 +354,118 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
...
         WCHAR ch = L'\n';
         ConStreamWrite(Pager->Screen->Stream, &ch, 1);
         return FALSE;
     }
 
-    /* 'Q': Quit */
-    // FIXME: Available only when command extensions are enabled.
-    if ((KeyEvent.wVirtualKeyCode == L'Q') &&
-        !(KeyEvent.dwControlKeyState & (LEFT_CTRL_PRESSED | RIGHT_CTRL_PRESSED)))
+    switch (chSubCommand)
```
HBelusca, line 366: I moved this subCommand block after the Ctrl-C handling.
Also it could even be moved in the block below when extensions are enabled. (Up to you whether you prefer this or not.)

base/applications/cmdutils/more/more.c
```diff
@@ -140,28 +354,118 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
...
-    return TRUE;
+    /* If extended features are available */
+    if (s_dwFlags & FLAG_E)
+    {
+        /* Ignore any key presses if Ctrl is pressed */
+        if (fCtrl)
+        {
+            chSubCommand = 0;
+            goto Restart;
+        }
```
HBelusca, line 391: Manual tests show that keys with Ctrl press are ignored only when extended features are enabled.

base/applications/cmdutils/more/more.c
```diff
@@ -140,28 +354,118 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
...
+
+        /* [Enter] key: Display one line */
+        if (KeyEvent.wVirtualKeyCode == VK_RETURN)
+        {
+            Pager->ScrollRows = 1;
+            return TRUE;
+        }
+
+        /* [Space] key: Display one page */
```
HBelusca, line 424: Manual tests show that specific Space handling is done (with the support of the /C per-page screen clearing) only when extensions are enabled.

base/applications/cmdutils/more/more.c
```diff
@@ -140,28 +354,118 @@ PagePrompt(PCON_PAGER Pager, DWORD Done, DWORD Total)
...
+        s_nPromptID = IDS_CONTINUE_PROGRESS;
+        chSubCommand = 0;
+        goto Restart;
+    }
+    else
+    {
+        /* Extended features are unavailable: display one page */
+        Pager->ScrollRows = Pager->ScreenRows - 1;
+        return TRUE;
+    }
```
HBelusca, line 468: When extensions are disabled, any key press (except Ctrl/Alt/Shift) move down one page.

base/applications/cmdutils/more/more.c
```diff
@@ -486,6 +790,141 @@ LoadRegistrySettings(HKEY hKeyRoot)
     RegCloseKey(hKey);
 }
 
+static inline BOOL IsFlag(LPCWSTR param)
```
HBelusca, line 793: This has been moved here, together with the other command-line parsing functions.

base/applications/cmdutils/more/lang/bg-BG.rc
katahiromz: OK.

sdk/lib/conutils/pager.c
katahiromz: OK!

base/applications/cmdutils/more/more.c
katahiromz: OK.

base/applications/cmdutils/more/more.c
katahiromz: OK.

sdk/lib/conutils/pager.h
katahiromz: OK.

base/applications/cmdutils/more/more.c
katahiromz: OK.

base/applications/cmdutils/more/more.c
katahiromz: OK.

base/applications/cmdutils/more/more.c
katahiromz: OK.

base/applications/cmdutils/more/more.c
katahiromz: OK.

base/applications/cmdutils/more/more.c
katahiromz: OK.

base/applications/cmdutils/more/more.c
katahiromz: OK.

base/applications/cmdutils/more/more.c
katahiromz: Simplified the for-loop. OK.

katahiromz:
I have introduced a bug for long lines. Debugging now...

base/applications/cmdutils/more/more.c
katahiromz: OK.

base/applications/cmdutils/more/more.c
katahiromz: OK.

base/applications/cmdutils/more/more.c
katahiromz: OK.

katahiromz:
The bug is fixed.

katahiromz:
`iColumn` variable in `ConPagerDefaultAction` has to consider tab widths...

katahiromz:
Tab expansion is fixed. Done. Check time.

If there is no objection, I will merge two days later.

sdk/lib/conutils/pager.c
```diff
@@ -31,7 +33,146 @@
...
+        }
+
+        if (TextBuff[ich] == TEXT('\t') &&
+            (Pager->dwFlags & CON_PAGER_FLAG_EXPAND_TABS))
+        {
+            ConCallPagerLine(Pager, &TextBuff[ichLast], ich - ichLast);
+            ichLast = ++ich;
+            Pager->nSpacePending += nTabWidth - (iColumn % nTabWidth);
+            goto ExpandTab;
+        }
```
HBelusca, line 141: Care needs to be taken in this block: What happens when `nTabWidth` == 0 ?
With a text containing the string (here in C format): `"a\tb\tc\td\t\te\t\tf"` , running WIndows' MORE:
```
E:\>more /T0 test_tabs.txt
abcdef

E:\>more /T1 test_tabs.txt
a b c d  e  f

E:\>more /T2 test_tabs.txt
a b c d   e   f

E:\>more /T3 test_tabs.txt
a  b  c  d     e     f

E:\>more /T4 test_tabs.txt
a   b   c   d       e       f

E:\>more /T5 test_tabs.txt
a    b    c    d         e         f

E:\>more /T6 test_tabs.txt
a     b     c     d           e           f

E:\>more /T7 test_tabs.txt
a      b      c      d             e             f

E:\>more /T8 test_tabs.txt
a       b       c       d               e               f
```

katahiromz: 無題
Division by Zero happens.

HBelusca: indeed

katahiromz: Fixed.

HBelusca: Do the other cases (tab width == 1, 2, etc...) work as in the example above?

katahiromz: The positive cases are OK. The negative cases are beyond my idle.

HBelusca: All negative cases behave exactly as the positive cases where the tab width is larger or equal to the current console screenbuffer width.

katahiromz: 無題1
無題2
It's difficult for me to emulate it...

katahiromz: Making the system FIFO can simplify this problem but it needs dynamic memory.

HBelusca: > Making the system FIFO (first-in, first-out) can simplify this problem but it needs dynamic memory.

what???

katahiromz: Making the buffering (with tab expansion etc.) dynamic FIFO will fix this problem.

HBelusca: Just watch how Windows' MORE works. If the code works when handling very large (>= width of the console screenbuffer) tab spaces, then it is trivial to implement the observed behaviour with negative space.

katahiromz: Sorry, It's impossible for me.

HBelusca: I will give it a try.

sdk/lib/conutils/pager.h
```diff
@@ -26,27 +26,42 @@
...
 {
     PCON_SCREEN Screen;
-
-    // TODO: Add more properties. Maybe those extra parameters
-    // of PAGE_PROMPT could go there?
-
-    /* Used to count number of lines since last pause */
-    DWORD LineCount;
+    DWORD ScreenColumns;
+    DWORD ScreenRows;
```
HBelusca, line 45: These two variables is actually already accessible via the Screen structure.

sdk/lib/conutils/pager.c
HBelusca: There is another bug in the tab-expansion, where, when a text line, after tab-expansion, takes more than one line and the tabs are not correctly positioned.

Example (the vertical bars are NOT part of the string being displayed; they just represent the borders of the console window):
This PR:
```
|a                             b                             c                   |
|          d                   e                             f                   |
```

Windows' MORE:
```
|a                             b                             c                   |
|          d                             e                             f         |
```

katahiromz: The second bug is fixed.

katahiromz:
Check time. I will merge three days later.

HBelusca:
> Check time. I will merge three days later.

Nice! Thanks to your latest fixes, I'm able to get my small patch for handling the tab width >= max console width or tab width negative, working as the MS' MORE command.

sdk/lib/conutils/pager.c
```diff
@@ -31,7 +32,188 @@
...
+
+        /* FORM-FEED character */
+        if (TextBuff[ich] == TEXT('\f') &&
+            (Pager->dwFlags & CON_PAGER_FLAG_EXPAND_FF))
+        {
+            /* Skip the form-feed and clear until the end of the screen */
+            ConCallPagerLine(Pager, &TextBuff[ichLast], ich - ichLast);
+            ichLast = ich + 1;
+            // FIXME: Should we handle CON_PAGER_FLAG_DONT_OUTPUT ?
```
HBelusca, line 157: I would like to know your advice on that.

sdk/lib/conutils/pager.c
```diff
@@ -31,7 +32,188 @@
...
+            while (iLine < ScrollRows)
+            {
+                ConCallPagerLine(Pager, L"\n", 1);
+                // CON_STREAM_WRITE(Pager->Screen->Stream, TEXT("\n"), 1);
+                // if (!(Pager->dwFlags & CON_PAGER_FLAG_DONT_OUTPUT))
```
HBelusca, line 162: Same: Should we use CON_STREAM_WRITE or ConCallPagerLine ? (Here the aim when expanding the form-feed character is to print out newlines for "filling" the screen up to the bottom)

sdk/lib/conutils/pager.c
```diff
@@ -31,7 +32,188 @@
...
+    UINT nWidthOfChar = 1;
+    BOOL IsDoubleWidthCharTrailing = FALSE;
+
+    if (ich >= cch)
+        return TRUE;
+
+    nCodePage = GetConsoleOutputCP();
+    IsCJK = IsCJKCodePage(nCodePage);
+
+    /* Normalize the tab width: if negative or too large,
```
HBelusca, line 96: That's the fix for the tab width being negative or "too large".

sdk/lib/conutils/pager.c
```diff
@@ -44,74 +226,65 @@ ConWritePaging(
...
-            if (!PagePrompt(Pager, from, len))
-                return FALSE;
+            Pager->ScreenColumns = csbi.srWindow.Right - csbi.srWindow.Left + 1;
+            Pager->ScreenRows = csbi.srWindow.Bottom - csbi.srWindow.Top + 1;
+        }
 
-            // TODO: Recalculate 'ScreenLines' in case the user redimensions
-            // the window during the prompt.
+        /* Reset to display one page by default */
+        Pager->ScrollRows = Pager->ScreenRows - 1;
```
HBelusca, line 282: I find this: resetting the ScrollRows value to its default, to be advantageous, so that the PagePrompt callback doesn't need to care about the "correct" value to set there, when it wants to tell the pager to display a full screen page. The PagePrompt callback resets the ScrollRows value to a different value only when it wants, for the next paging action, to change the number of lines.

HBelusca (commented):
So here are the main fix (the tab negative value support) and the implementation for `/P` (expand form-feeds).

Regarding `/P`:
I have used this test file (call it `"test.txt"`) (contains also tabs):
```
a	b	c	d	e	f
aa	bb	cc	dd	ee	ff
aaa	bbb	ccc	ddd	eee	fff
aaaa	bbbb	cccc	dddd	eeee	ffff
This is aformfeed
After formfeed
blabla
foobar
blabla
foobar
blabla
<many more foobar / blabla>
```
(between `a`, and `form`, and `form` and `feed`, there is one form-feed character `\f` , HEX=0x0C, inserted).

Note that MS' MORE.COM command has a bug: when doing `more /P test.txt`, it will display everything until `This is a`, then fill the remaining screen with newline and prompt (OK), and if you press either ENTER or SPACE, it displays garbage instead of displaying `form`; then at the next ENTER or SPACE keypress it will again display garbage instead of displaying `feed` (this is the bug).
Then it will display the remaining contents of the file as usual (OK).

For example:
```
This is a



















 ȼ     <---- garbage 1

 ȼ     <---- garbage 2
After formfeed
-- More (34%) --
```
(using a 80x25 console, and I've pressed SPACE twice at the prompt.)
With our MORE command however (this PR) we behave correctly:
```
aaaa    bbbb    cccc    dddd    eeee    ffff
This is a



















form    <---- OK
feed    <---- OK
After formfeed
-- Continue (34%) --
```

sdk/lib/conutils/pager.c
katahiromz: For the future that extends the pager flags, we'd better not to use `CON_STREAM_WRITE` here directly.

sdk/lib/conutils/pager.c
katahiromz: We should count the logical line number, so we have to call `ConCallPagerLine` even if no output.

sdk/lib/conutils/pager.c
katahiromz: `// if (!(Pager->dwFlags & CON_PAGER_FLAG_DONT_OUTPUT))`
Uncomment this.

HBelusca: If this gets uncommented there's the risk of not increasing iLine and therefore entering an infinite while loop.

sdk/lib/conutils/pager.c
HBelusca: Why increasing `lineno` here? This is unrelated to an actual newline in the file.

katahiromz: Sorry.
https://github.com/reactos/reactos/pull/3658/files#diff-01bb03740bc237b04360f7d7054804371263fecc66b7e2b9048fea5f3e7c9a37R91-R94
```c
    if (s_dwFlags & FLAG_PLUSn) /* Skip lines */
    {
        if (Pager->lineno < s_nNextLineNo)
        {
```
Flag `FLAG_PLUSn` is watching `lineno`.

sdk/lib/conutils/pager.c
katahiromz:
```c
for (n = 0; n < ScrollRows - iLine; ++n)
{
    ConCallPagerLine(Pager, L"\n", 1);
    if (!(Pager->dwFlags & CON_PAGER_FLAG_DONT_OUTPUT))
        ++iLine;
}
iColumn = 0;
continue;
```

HBelusca: I'll check this; nevertheless now this looks slightly "overengineered" (just to output consecutive newlines).

HBelusca:
Miscellaneous fixes for problems encountered during testing.
(NOTE: MORE output to file can be tested with e.g.: `more some_file.txt > output_file.txt`)

base/applications/cmdutils/more/lang/bg-BG.rc
```diff
@@ -3,13 +3,35 @@ LANGUAGE LANG_BULGARIAN, SUBLANG_DEFAULT
...
+at ""-- Continue --"" prompt:\n\
+   P n      Display n lines.\n\
+   S n      Skip n lines.\n\
+   F        Display next file.\n\
+   Q        Quit.\n\
+   ?        Show help line.\n\
+   =        Show line number.\n\
+   <Space>  Display next page.\n\
+   <Enter>  Display next line.\n\
+   <Esc>    Same as Q.\n"
```
katahiromz, line 28:
```suggestion
   <Enter>  Display next line.\n"
```

base/applications/cmdutils/more/lang/bg-BG.rc
katahiromz, line 34:
```suggestion
    IDS_CONTINUE_OPTIONS "-- Continue (%d%%) [Options: psfq=<Space><Enter>] --"
```

base/applications/cmdutils/more/lang/bg-BG.rc
katahiromz: Fixed.

base/applications/cmdutils/more/lang/bg-BG.rc
katahiromz: Fixed.

katahiromz:
To go, or not to go?

HBelusca (approved):
OK let's ship it. I have other improvements I'm working on; this will be done in a separate PR.

PR #3661, opened 2021-05-12
[MSGINA] Implement Product version Banners

jeffbox12:
## Purpose

This is a follow-up of my Product version Banner implementation based on version selection on install.
Now the Logon, Shut down and Please wait dialogs will now show the Product version.

## Proposed changes

- `msgina` will now show the correct Product version selected
- New banner design based on feedback
- New banners for VGA mode
- Each banner has now it's own .svg file for easy copyright year change automation

## Preview

123635829-67609980-d81c-11eb-96ef-027d57fc6520


dll/win32/msgina/gui.c
```diff
@@ -66,14 +67,23 @@ DlgData_LoadBitmaps(_Inout_ PDLG_DATA pDlgData)
...
+        pDlgData->hLogoBitmap = LoadImageW(pDlgData->pgContext->hDllInstance,
+                                           MAKEINTRESOURCEW(IDI_ROSLOGO_WORKSTATION), IMAGE_BITMAP,
+                                           0, 0, LR_DEFAULTCOLOR);
+    }
     if (pDlgData->hLogoBitmap)
     {
         GetObject(pDlgData->hLogoBitmap, sizeof(bm), &bm);
         pDlgData->LogoWidth = bm.bmWidth;
-        pDlgData->LogoHeight = bm.bmHeight;
+        pDlgData->LogoHeight = bm.bmHeight;
```
HBelusca, line 86:
```suggestion
        pDlgData->LogoHeight = bm.bmHeight;
```

dll/win32/msgina/resource.h
```diff
@@ -77,6 +77,9 @@
 #define IDI_ROSLOGO_FLAG         20002
 #define IDI_DLG_BG               20003
 #define IDB_IMAGE_STRIP          20004
+#define IDI_ROSLOGO_WORKSTATION  20005
+#define IDI_ROSLOGO_SERVER       20006
+
```
HBelusca, line 82:
```suggestion
```

dll/win32/msgina/shutdown.c
```diff
@@ -1029,9 +1030,18 @@ ShutdownDialogProc(
...
+                pContext->hBitmap =
+                    LoadImageW(pContext->pgContext->hDllInstance, MAKEINTRESOURCEW(IDI_ROSLOGO_SERVER), IMAGE_BITMAP, 0, 0, LR_DEFAULTCOLOR);
+                return TRUE;
+            }
+            else
+            {
+                pContext->hBitmap =
+                    LoadImageW(pContext->pgContext->hDllInstance, MAKEINTRESOURCEW(IDI_ROSLOGO_WORKSTATION), IMAGE_BITMAP, 0, 0, LR_DEFAULTCOLOR);
+                return TRUE;
+            }
```
HBelusca, line 1044:
```suggestion
            if (IsWindowsServer())
            {
                pContext->hBitmap = LoadImageW(pContext->pgContext->hDllInstance,
                                               MAKEINTRESOURCEW(IDI_ROSLOGO_SERVER), IMAGE_BITMAP,
                                               0, 0, LR_DEFAULTCOLOR);
            }
            else
            {
                pContext->hBitmap = LoadImageW(pContext->pgContext->hDllInstance,
                                               MAKEINTRESOURCEW(IDI_ROSLOGO_WORKSTATION), IMAGE_BITMAP,
                                               0, 0, LR_DEFAULTCOLOR);
            }
            return TRUE;
```

binarymaster: Ideally it would be abstract this code somewhere, since we already have it in shell32 right? :)

HBelusca: Besides the fact that msgina should be able to run in the absence of shell32...
On Vista+ there's a specific resource-only branding dll just for this kind of thing.

dll/win32/msgina/gui.c
```diff
@@ -66,9 +67,36 @@ DlgData_LoadBitmaps(_Inout_ PDLG_DATA pDlgData)
...
-    pDlgData->hLogoBitmap = LoadImageW(pDlgData->pgContext->hDllInstance,
-                                       MAKEINTRESOURCEW(IDI_ROSLOGO), IMAGE_BITMAP,
-                                       0, 0, LR_DEFAULTCOLOR);
+    if (IsWindowsServer())
+    {
+        pDlgData->hLogoBitmap = LoadImageW(pDlgData->pgContext->hDllInstance,
+                                           MAKEINTRESOURCEW(IDI_ROSLOGO_SERVER), IMAGE_BITMAP,
+                                           0, 0, LR_DEFAULTCOLOR);
+
+        GetObject(pDlgData->hLogoBitmap, sizeof(BITMAP), &bm);
```
HBelusca, line 76:
```suggestion
        GetObject(pDlgData->hLogoBitmap, sizeof(bm), &bm);
```

dll/win32/msgina/gui.c
```diff
@@ -66,9 +67,36 @@ DlgData_LoadBitmaps(_Inout_ PDLG_DATA pDlgData)
...
+                                               0, 0, LR_DEFAULTCOLOR);
+        }
+    }
+    else
+    {
+        pDlgData->hLogoBitmap = LoadImageW(pDlgData->pgContext->hDllInstance,
+                                           MAKEINTRESOURCEW(IDI_ROSLOGO_WORKSTATION), IMAGE_BITMAP,
+                                           0, 0, LR_DEFAULTCOLOR);
+
+        GetObject(pDlgData->hLogoBitmap, sizeof(BITMAP), &bm);
```
HBelusca, line 91:
```suggestion
        GetObject(pDlgData->hLogoBitmap, sizeof(bm), &bm);
```

HBelusca (commented):
Shouldn't we have the same in shell32's winver dialog?

jeffbox12:
> Shouldn't we have the same in shell32's winver dialog?

Yes, thats already implemented.

dll/win32/msgina/gui.c
```diff
@@ -282,6 +310,18 @@ StatusDialogProc(
             DlgData_Destroy(hwndDlg);
             return TRUE;
         }
+
+        case WM_PAINT:
```
learn-more, line 314: Why is it now painted manually instead of just adding the image to the control?

jeffbox12: The Normal and VGA bitmap are two separate Bitmaps. It is painted manually so it can display the correct bitmap.

learn-more: You can just activate the correct bitmap in the control using `STM_SETIMAGE` or something like that?

jeffbox12: From what I've seen throughout this code, this is exactly how the Bitmap is painted on each dialog.

ColinFinck (changes requested):
This PR adds a few drop shadows that we don't use anywhere else. It leaves the "ReactOS" text misaligned and "OS" in a different color, with different proportions between the ReactOS logo and text compared to our ReactOS Corporate Identity everywhere else.

I don't mind changing our design, but it needs to be done consistently, and needs to be backed by a majority of the team. I have written down my observations about design inconsistencies from 0.4.13 until today in all detail here: https://github.com/reactos/reactos/pull/2860#pullrequestreview-709871345

Let's first get a common understanding about how ReactOS shall eventually look like before we change bitmaps one by one. Otherwise, we're going to replace bitmaps over and over again, without ever getting to an optimal state.

JoachimHenze:
A question regarding code-maintability: Can we remove the year strings within the copyright text on those bitmaps, so that we won't have to groom them in many different places year for year? What do others think about that?

HBelusca:
> A question regarding code-maintability: Can we remove the year strings within the copyright text on those bitmaps, so that we won't have to groom them in many different places year for year? What do others think about that?

Sounds fine to me, but then, the "1996" year would disappear from them as well?

learn-more:
> A question regarding code-maintability: Can we remove the year strings within the copyright text on those bitmaps, so that we won't have to groom them in many different places year for year? What do others think about that?

Or write some code that paints that string over it dynamically

jeffbox12:
PR converted to Draft until work is done.

PR #3662, opened 2021-05-12
[PC98VID] Fix MSVC 2015 build for PC-98 target

disean:
## Proposed changes

Addendum to 3adf45086743d5f075e4270af7af64017eb01d63

Also isolate the constant structures into page section
and restore the previous sections for start IO and entrypoint routines.

### DIFF
--- a/win32ss/drivers/miniport/pc98vid/pc98vid.c
+++ b/win32ss/drivers/miniport/pc98vid/pc98vid.c
@@ -11,6 +11,7 @@
 
 /* GLOBALS ********************************************************************/
 
+DATA_SEG("PAGECONS")
 const VIDEOMODE VideoModes[] =
 {
     {640, 480, GRAPH_HF_31KHZ, GDC2_CLOCK1_5MHZ, GDC2_CLOCK2_5MHZ,
@@ -18,7 +19,8 @@ const VIDEOMODE VideoModes[] =
      {0, 80, 12, 2, 4, 4, 6, 480, 37}, {0, 80, 12, 2, 4, 132, 6, 480, 37}}
 };
 
-static VIDEO_ACCESS_RANGE LegacyRangeList[] =
+DATA_SEG("PAGECONS")
+const VIDEO_ACCESS_RANGE LegacyRangeList[] =
 {
     { {{0x60,  0}}, 0x00000001, 1, 1, 1, 0 },
     { {{0x62,  0}}, 0x00000001, 1, 1, 1, 0 },
@@ -72,7 +74,7 @@ Pc98VidFindAdapter(
 
     Status = VideoPortVerifyAccessRanges(DeviceExtension,
                                          RTL_NUMBER_OF(LegacyRangeList),
-                                         LegacyRangeList);
+                                         (PVIDEO_ACCESS_RANGE)LegacyRangeList);
     if (Status != NO_ERROR)
     {
         VideoDebugPrint((Error, "%s() Resource conflict was found\n", __FUNCTION__));
@@ -193,6 +195,15 @@ Pc98VidGetVideoChildDescriptor(
     return ERROR_NO_MORE_DEVICES;
 }
 
+#if defined(_MSC_VER)
+/*
+ * Avoid C2983 error for MSVC 2015. There is no such thing
+ * as DRIVER_INITIALIZE for video miniport drivers.
+ */
+#pragma alloc_text(INIT, DriverEntry)
+#else
+CODE_SEG("INIT")
+#endif
 ULONG
 NTAPI
 DriverEntry(
@@ -227,7 +238,7 @@ DriverEntry(
         InitData.HwGetVideoChildDescriptor = Pc98VidGetVideoChildDescriptor;
     }
 
-    InitData.HwLegacyResourceList = LegacyRangeList;
+    InitData.HwLegacyResourceList = (PVIDEO_ACCESS_RANGE)LegacyRangeList;
     InitData.HwLegacyResourceCount = RTL_NUMBER_OF(LegacyRangeList);
 
     InitData.AdapterInterfaceType = Isa;
--- a/win32ss/drivers/miniport/pc98vid/pc98vid.h
+++ b/win32ss/drivers/miniport/pc98vid/pc98vid.h
@@ -22,6 +22,10 @@
 #define READ_PORT_UCHAR(p) VideoPortReadPortUchar(p)
 #include <drivers/pc98/video.h>
 
+#if defined(_MSC_VER)
+#pragma section("PAGECONS", read)
+#endif
+
 #define MONITOR_HW_ID 0x1033FACE /* Dummy */
 
 typedef struct _VIDEOMODE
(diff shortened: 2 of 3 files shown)


win32ss/drivers/miniport/pc98vid/pc98vid.c
```diff
@@ -193,6 +193,15 @@ Pc98VidGetVideoChildDescriptor(
...
 
+#if defined(_MSC_VER)
+/*
+ * Avoid C2983 error for MSVC 2015. There is no such thing
+ * as DRIVER_INITIALIZE for video miniport drivers.
+ */
+#pragma alloc_text(INIT, DriverEntry)
+#else
+CODE_SEG("INIT")
+#endif
```
HBelusca, line 204: I don't see how the INIT segment thing is related to DRIVER_INITIALIZE .

win32ss/drivers/miniport/pc98vid/pc98vid.c
```diff
@@ -193,6 +193,15 @@ Pc98VidGetVideoChildDescriptor(
     return ERROR_NO_MORE_DEVICES;
 }
 
+#if defined(_MSC_VER)
```
SergeGautherie, line 196:
```suggestion
#if defined(_MSC_VER) && _MSC_VER < 1910
```

win32ss/drivers/miniport/pc98vid/pc98vid.c
```diff
@@ -11,14 +11,14 @@
 
 /* GLOBALS ********************************************************************/
 
-const VIDEOMODE VideoModes[] =
+DATA_SEG("PAGECONS") VIDEOMODE VideoModes[] =
```
zefklop, line 14: this still needs to be marked as const

win32ss/drivers/miniport/pc98vid/pc98vid.c
zefklop: Yes, Just put the CODE_SEG declaration

win32ss/drivers/miniport/pc98vid/pc98vid.h
```diff
@@ -54,37 +58,43 @@ NTAPI
...
+CODE_SEG("PAGE")
 VP_STATUS
 NTAPI
 Pc98VidSetPowerState(
     _In_ PVOID HwDeviceExtension,
     _In_ ULONG HwId,
     _In_ PVIDEO_POWER_MANAGEMENT VideoPowerControl);
 
-extern const VIDEOMODE VideoModes[];
+DATA_SEG("PAGECONS")
```
zefklop, line 97: Please keep it const

win32ss/drivers/miniport/pc98vid/pc98vid.c
```diff
@@ -11,14 +11,14 @@
...
 {
     {640, 480, GRAPH_HF_31KHZ, GDC2_CLOCK1_5MHZ, GDC2_CLOCK2_5MHZ,
      GDC2_MODE_LINES_800, 60,
      {0, 80, 12, 2, 4, 4, 6, 480, 37}, {0, 80, 12, 2, 4, 132, 6, 480, 37}}
 };
 
-static VIDEO_ACCESS_RANGE LegacyRangeList[] =
+DATA_SEG("PAGECONS") VIDEO_ACCESS_RANGE LegacyRangeList[] =
```
zefklop, line 21: Please keep it const

zefklop (changes requested):
- [ ] Unconditionally use CODE_SEG for DriverEntry
- [ ] Keep the const qualifier for constant data

win32ss/drivers/miniport/pc98vid/pc98vid.c
disean: Maybe DRIVER_INITIALIZE is a special linker thing, I had this error:
<details>
D:\a\reactos\reactos\src\win32ss\drivers\miniport\pc98vid\pc98vid.c(210): error C2983: 'DriverEntry': all declarations must have an identical __declspec(code_seg(...))
D:\a\reactos\reactos\src\sdk\include\ddk\video.h(62): note: see declaration of 'DriverEntry'
  
</details>

This declaration doesn't help
```c
CODE_SEG("INIT")
ULONG
NTAPI
DriverEntry(
    _In_ PVOID Context1,
    _In_ PVOID Context2);
```

SergeGautherie: You need to sync
win32ss/drivers/miniport/pc98vid/pc98vid.c

HBelusca: > I had this error:
> D:\a\reactos\reactos\src\win32ss\drivers\miniport\pc98vid\pc98vid.c(210): error C2983: 'DriverEntry': all declarations must have an identical __declspec(code_seg(...)) D:\a\reactos\reactos\src\sdk\include\ddk\video.h(62): note: see declaration of 'DriverEntry'

Because in our video.h it is declared like that:
https://git.reactos.org/?p=reactos.git;a=blob;f=sdk/include/ddk/video.h;h=1500e03bb9bfac8827e03e2e6a9ab9482df6c822;hb=HEAD#l59
```c
  60 ULONG
  61 NTAPI
  62 DriverEntry(
  63   PVOID Context1,
  64   PVOID Context2);
  65 
```
without any CODE_SEG thing, which is actually legit because this is in the public DDK and it cannot enforce you to put the thing in a INIT section, but you want to, and with MSVC 2015 when using that CODE_SEG thing it would appear it requires as well the prototype to have the same CODE_SEG.

HBelusca: We shouldn't add any CODE_SEG to the video.h's DriverEntry prototype, because otherwise we would break compatibility with the official MS DDK/WDK.

disean: Yes, that's why I added ` #if defined(_MSC_VER) `. I really don't understand why WDM stuff works there without CODE_SEG thing https://git.reactos.org/?p=reactos.git;a=blob;f=sdk/include/xdk/iotypes.h;hb=eba00d1b5ed678c5557271649d7cb41dd4ee47f7#l2229

SergeGautherie: It looks like the other miniports do not have this.
Need to update them all or _none_, due to a common declaration?
https://git.reactos.org/?p=reactos.git;a=blob;f=sdk/include/ddk/video.h;hb=HEAD#l59

HBelusca: > Yes, that's why I added `#if defined(_MSC_VER)`. I really don't understand why WDM stuff works there without CODE_SEG thing.

Because in that header for example, this is a (function) typedef that is being declared, and not an actual function prototype.

disean: Sure :sweat_smile:

SergeGautherie: Then, forget about it or add it for 2017+ only.

zefklop: Maybe we should just have VS2015 have CODE_SEG being a NOP. We already spent way too much time on this...

Extravert-ir: What do we need MSVC 2015 for? The bug was unnoticed for quite some time, nobody cared.
Now we have to change our practice and include CODE_SEG even in headers, just for this particular compiler

HBelusca: The `#pragma alloc_text(...)` technique Microsoft always used (and still use), e.g. in the WDK, etc.., always worked, from MSVC 6 up to now.

Extravert-ir: Yeah, but that doesn't work for GCC

SergeGautherie: > The bug was unnoticed for quite some time

Not fully true: Jira tickets were/are filed about VS2015 regressions.
Of course, dropping VS2015 CI, while claiming support, did not help!

> nobody cared.

That, on the other hand...

zefklop (approved):

PR #3663, opened 2021-05-13
[GITHUB] build.yml: Sync' syntax

SergeGautherie:
No bug, yet more consistent.


zefklop:
Next time you can add such trivial changes to another PR, like #3645

SergeGautherie:
> Next time you can add such trivial changes to another PR, like #3645

Yeah: some other devs complain when I do that... Anyway, let's not reopen that debate here.

PR #3665, opened 2021-05-15
[KERNEL32][NTDLL] Implement SetSearchPathMode and RtlSetSearchPathMode

oleg-dubinskiy:
## Purpose

Implement `SetSearchPathMode` in kernel32 and call an appropriate `RtlSetSearchPathMode` from ntdll inside it, which I also implemented (partially based on Wine's one).
Although from the start these APIs were introduced since Windows 7/Server 2008 R2, they are also available for XP/2003 via update, so they exist as well on XPSP3/2K3SP2.
Required by KeePassX 0.4.4 to start successfully.

JIRA issue: [CORE-17586](https://jira.reactos.org/browse/CORE-17586)

## TODO

- [ ] Properly obtain the search mode value from registry in failure case. I'm not sure about where it should be implemented. Or it just should be done by the caller itself, judging by MSDN?... :thinking: 

## Result

Although my implementation is still incomplete, nevertheless it is enough to allow KeePassX to start successfully, as visible on screenshots. :slightly_smiling_face: 

Before:
keepassx-error

After:
keepassx-ros

XP:
keepassx-xp

### DIFF
--- a/dll/ntdll/def/ntdll.spec
+++ b/dll/ntdll/def/ntdll.spec
@@ -1158,6 +1158,7 @@
 @ stub -version=0x600+ RtlSetProcessDebugInformation
 @ cdecl RtlSetProcessIsCritical(long ptr long)
 @ stdcall RtlSetSaclSecurityDescriptor(ptr long ptr long)
+@ stdcall RtlSetSearchPathMode(long)
 @ stdcall RtlSetSecurityDescriptorRMControl(ptr ptr)
 @ stdcall RtlSetSecurityObject(long ptr ptr ptr ptr)
 @ stdcall RtlSetSecurityObjectEx(long ptr ptr long ptr ptr)
--- a/dll/win32/kernel32/client/path.c
+++ b/dll/win32/kernel32/client/path.c
@@ -1120,7 +1120,7 @@ GetFullPathNameW(IN LPCWSTR lpFileName,
  */
 BOOL
 WINAPI
-SetSearchPathMode(IN DWORD dwFlags)
+SetSearchPathMode(_In_ DWORD dwFlags)
 {
     NTSTATUS Status;
 
--- a/dll/win32/kernel32/client/path.c
+++ b/dll/win32/kernel32/client/path.c
@@ -4,6 +4,7 @@
  * FILE:            dll/win32/kernel32/client/path.c
  * PURPOSE:         Handles path APIs
  * PROGRAMMERS:     Alex Ionescu (alex.ionescu@reactos.org)
+ *                  Oleg Dubinskiy (oleg.dubinskij2013@yandex.ua)
  */
 
 /* INCLUDES *******************************************************************/
@@ -1112,6 +1113,26 @@ GetFullPathNameW(IN LPCWSTR lpFileName,
                                 nBufferLength * sizeof(WCHAR),
                                 lpBuffer,
                                 lpFilePart) / sizeof(WCHAR);
+}
+
+/*
+ * @implemented
+ */
+BOOL
+WINAPI
+SetSearchPathMode(IN DWORD dwFlags)
+{
+    NTSTATUS Status;
+
+    /* Call Rtl to do the work */
+    Status = RtlSetSearchPathMode(dwFlags);
+    if (!NT_SUCCESS(Status))
+    {
+        SetLastError(RtlNtStatusToDosError(Status));
+        return FALSE;
+    }
+
+    return NT_SUCCESS(Status);
 }
 
 /*
--- a/dll/win32/kernel32/client/path.c
+++ b/dll/win32/kernel32/client/path.c
@@ -1128,7 +1128,7 @@ SetSearchPathMode(IN DWORD dwFlags)
     Status = RtlSetSearchPathMode(dwFlags);
     if (!NT_SUCCESS(Status))
     {
-        SetLastError(RtlNtStatusToDosError(Status));
+        BaseSetLastNTError(Status);
         return FALSE;
     }
 
--- a/dll/win32/kernel32/client/path.c
+++ b/dll/win32/kernel32/client/path.c
@@ -1132,7 +1132,7 @@ SetSearchPathMode(IN DWORD dwFlags)
         return FALSE;
     }
 
-    return NT_SUCCESS(Status);
+    return TRUE;
 }
 
 /*
--- a/sdk/include/psdk/winbase.h
+++ b/sdk/include/psdk/winbase.h
@@ -3166,6 +3166,7 @@ BOOL WINAPI SetProcessAffinityMask(_In_ HANDLE, _In_ DWORD_PTR);
 BOOL WINAPI SetProcessPriorityBoost(_In_ HANDLE, _In_ BOOL);
 BOOL WINAPI SetProcessShutdownParameters(DWORD,DWORD);
 BOOL WINAPI SetProcessWorkingSetSize(_In_ HANDLE, _In_ SIZE_T, _In_ SIZE_T);
+BOOL WINAPI SetSearchPathMode(_In_ DWORD);
 #if (_WIN32_WINNT >= 0x0600)
 VOID WINAPI SetSecurityAccessMask(SECURITY_INFORMATION,LPDWORD);
 #endif
--- a/sdk/lib/rtl/path.c
+++ b/sdk/lib/rtl/path.c
@@ -2177,17 +2177,17 @@ RtlSetSearchPathMode(_In_ ULONG Flags)
 
     switch (Flags)
     {
-    case BASE_SEARCH_PATH_ENABLE_SAFE_SEARCHMODE:
-        RetVal = 1;
-        break;
-    case BASE_SEARCH_PATH_DISABLE_SAFE_SEARCHMODE:
-        RetVal = 0;
-        break;
-    case BASE_SEARCH_PATH_ENABLE_SAFE_SEARCHMODE | BASE_SEARCH_PATH_PERMANENT:
-        InterlockedExchange((LONG *)&RtlPathSafeMode, 2);
-        return STATUS_SUCCESS;
-    default:
-        return STATUS_INVALID_PARAMETER;
+        case BASE_SEARCH_PATH_ENABLE_SAFE_SEARCHMODE:
+            RetVal = 1;
+            break;
+        case BASE_SEARCH_PATH_DISABLE_SAFE_SEARCHMODE:
+            RetVal = 0;
+            break;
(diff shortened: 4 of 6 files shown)


dll/win32/kernel32/client/path.c
```diff
@@ -1114,6 +1115,26 @@ GetFullPathNameW(IN LPCWSTR lpFileName,
...
+
+    /* Call Rtl to do the work */
+    Status = RtlSetSearchPathMode(dwFlags);
+    if (!NT_SUCCESS(Status))
+    {
+        SetLastError(RtlNtStatusToDosError(Status));
+        return FALSE;
+    }
+
+    return NT_SUCCESS(Status);
```
SergeGautherie, line 1135:
```suggestion
    return TRUE;
```

sdk/lib/rtl/path.c
```diff
@@ -2160,6 +2166,44 @@ NTSTATUS NTAPI RtlNtPathNameToDosPathName(IN ULONG Flags,
     return STATUS_SUCCESS;
 }
 
+/*
+ * @implemented
+ */
+NTSTATUS
+NTAPI
+RtlSetSearchPathMode(IN ULONG Flags)
```
SergeGautherie, line 2174:
```suggestion
RtlSetSearchPathMode(_In_ ULONG Flags)
```

sdk/include/psdk/winbase.h
```diff
@@ -3166,6 +3166,7 @@ BOOL WINAPI SetProcessAffinityMask(_In_ HANDLE, _In_ DWORD_PTR);
 BOOL WINAPI SetProcessPriorityBoost(_In_ HANDLE, _In_ BOOL);
 BOOL WINAPI SetProcessShutdownParameters(DWORD,DWORD);
 BOOL WINAPI SetProcessWorkingSetSize(_In_ HANDLE, _In_ SIZE_T, _In_ SIZE_T);
+BOOL WINAPI SetSearchPathMode(DWORD);
```
SergeGautherie, line 3169:
```suggestion
BOOL WINAPI SetSearchPathMode(_In_ DWORD);
```

dll/win32/kernel32/client/path.c
```diff
@@ -1114,6 +1115,26 @@ GetFullPathNameW(IN LPCWSTR lpFileName,
                                 lpFilePart) / sizeof(WCHAR);
 }
 
+/*
+ * @implemented
+ */
+BOOL
+WINAPI
+SetSearchPathMode(DWORD dwFlags)
```
SergeGautherie, line 1123:
```suggestion
SetSearchPathMode(_In_ DWORD dwFlags)
```

sdk/lib/rtl/path.c
oleg-dubinskiy: Is this change necessary? It only makes the difference in coding style. Does it change anything else?

dll/win32/kernel32/client/path.c
oleg-dubinskiy: Is this correct? Shouldn't this function receive the appropriate status from the actual Rtl one instead? :thinking:

dll/win32/kernel32/client/path.c
oleg-dubinskiy: Is it necessary? In MSDN prototype, this parameter is not marked as input at all. Although technically it seems to be correct, but does it change anything?

HBelusca: Except that there is an annotation when you look at the function prototype in MS PSDK' winbase.h.

oleg-dubinskiy: OK, then it seems to be indeed required.

sdk/include/psdk/winbase.h
oleg-dubinskiy: Committed.

dll/win32/kernel32/client/path.c
oleg-dubinskiy: Committed, thanks.

sdk/lib/rtl/path.c
SergeGautherie: What is the point of using old-style `IN` in *.c, while using new-style `_In_` in *.h?

dll/win32/kernel32/client/path.c
SergeGautherie: Sorry, I don't understand what you are asking for.

HBelusca: @oleg-dubinskiy : Since you've handled the case where `NT_SUCCESS(Status)` is FALSE above (and you unconditionally return FALSE in that case), then now when the code flow is reaching that `return` statement, `NT_SUCCESS(Status)` can only be TRUE. That's why Serge suggested this change.

dll/win32/kernel32/client/path.c
```diff
@@ -1114,6 +1115,26 @@ GetFullPathNameW(IN LPCWSTR lpFileName,
...
+WINAPI
+SetSearchPathMode(IN DWORD dwFlags)
+{
+    NTSTATUS Status;
+
+    /* Call Rtl to do the work */
+    Status = RtlSetSearchPathMode(dwFlags);
+    if (!NT_SUCCESS(Status))
+    {
+        SetLastError(RtlNtStatusToDosError(Status));
```
HBelusca, line 1131:
```suggestion
        BaseSetLastNTError(Status);
```

dll/win32/kernel32/client/path.c
oleg-dubinskiy: OK, thanks @HBelusca. Then this is indeed correct. Sorry if that.

oleg-dubinskiy: Committed.

dll/win32/kernel32/client/path.c
oleg-dubinskiy: Committed, thanks.

sdk/lib/rtl/path.c
oleg-dubinskiy: As I see, such style is used for all other functions in these *.c and *.h, so I just use the general style.

SergeGautherie: Obviously not in the 2 modified .h...

oleg-dubinskiy: OK, committed.

sdk/lib/rtl/path.c
```diff
@@ -51,6 +52,11 @@ static const UNICODE_STRING RtlpDefaultExtension = RTL_CONSTANT_STRING(L".DLL");
 static const UNICODE_STRING RtlpDotLocal = RTL_CONSTANT_STRING(L".Local\\");
 static const UNICODE_STRING RtlpPathDividers = RTL_CONSTANT_STRING(L"\\/");
 
+/* Search mode used by RtlDosSearchPath_Ustr() and RtlSetSeatchPathMode() */
```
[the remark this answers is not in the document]
oleg-dubinskiy: Fixed, thanks.

dll/win32/kernel32/client/path.c
```diff
@@ -1114,6 +1115,26 @@ GetFullPathNameW(IN LPCWSTR lpFileName,
                                 lpFilePart) / sizeof(WCHAR);
 }
 
+/*
+ * @implemented
+ */
+BOOL
+WINAPI
+SetSearchPathMode(IN DWORD dwFlags)
```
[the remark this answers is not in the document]
oleg-dubinskiy: Now definitely committed. :slightly_smiling_face:

sdk/lib/rtl/path.c
```diff
@@ -2160,6 +2166,44 @@ NTSTATUS NTAPI RtlNtPathNameToDosPathName(IN ULONG Flags,
...
+ * @implemented
+ */
+NTSTATUS
+NTAPI
+RtlSetSearchPathMode(_In_ ULONG Flags)
+{
+    LONG RetVal;
+
+    switch (Flags)
+    {
```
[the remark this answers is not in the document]
oleg-dubinskiy: Done. :slightly_smiling_face:

SergeGautherie (approved):
Looks good to me.

learn-more:
@oleg-dubinskiy Why did you forward it to `RtlSetSearchPathMode` ?
In our implementation, all path handling exists in `kernel32\client\path.c`, so it would look more like what is done with `BaseDllOrderCurrent` in that file.

sdk/lib/rtl/path.c
```diff
@@ -51,6 +52,11 @@ static const UNICODE_STRING RtlpDefaultExtension = RTL_CONSTANT_STRING(L".DLL");
 static const UNICODE_STRING RtlpDotLocal = RTL_CONSTANT_STRING(L".Local\\");
 static const UNICODE_STRING RtlpPathDividers = RTL_CONSTANT_STRING(L"\\/");
 
+/* Search mode used by RtlDosSearchPath_Ustr() and RtlSetSearchPathMode() */
```
learn-more, line 55: `RtlDosSearchPath_Ustr` gets a path passed in, it does not calculate it's own path.

dll/ntdll/def/ntdll.spec
```diff
@@ -1158,6 +1158,7 @@
 @ stub -version=0x600+ RtlSetProcessDebugInformation
 @ cdecl RtlSetProcessIsCritical(long ptr long)
 @ stdcall RtlSetSaclSecurityDescriptor(ptr long ptr long)
+@ stdcall RtlSetSearchPathMode(long)
```
HBelusca, line 1161: This is Vista+

JoachimHenze: @HBelusca  please read the description of JIRA ticket https://jira.reactos.org/browse/CORE-17586 ! Does that answer/invalidate your concern?

JoachimHenze: In my XPSP3 kernel32.dll there is an export for SetSearchPathMode:
image

But in my XPSP3 ntdll.dll there is **no** export for RtlSetSearchPathMode
image

So we must find a solution here, that adds the one in kernel32.dll, but not the one in ntdll.dll

xmine64:
> Properly obtain the search mode value from registry in failure case. I'm not sure about where it should be implemented.

Based on https://learn.microsoft.com/en-us/windows/win32/api/processenv/nf-processenv-searchpathw the value is read from the following address:

```
HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Session Manager\SafeProcessSearchMode
```

PR #3667, opened 2021-05-17
[GITPOD] change extract directory to automatically extract to RosBE, …

puntillol59:
## Purpose

change extract directory to automatically extract to RosBE, and set tar to verbose mode shorting the waiting time for the user and have verbose mode activated to show if there were any issues with the file itself. therefore helping the user and developer.

JIRA issue: [CORE-XXXX](https://jira.reactos.org/browse/CORE-XXXX)

## Proposed changes

- verbose mode flag activated
- tar directory flags changed


zefklop:
Maybe you could have this build RosBE-Unix from its latest release instead. Or even better, use the PPA :-)

PR #3671, opened 2021-05-20
[NTOS] SpinLock debugging

zefklop:
## Purpose

Make sure we don't already own spinlocks when acquiring them and avoid embarassing moments like with 374fef2d59f39b714ff3152f1590f6af843c8bf5


zefklop:
Not mergeable. Let's open a new one.

PR #3674, opened 2021-05-20
[COMCTL32] Sync datetime.c to Wine-6.0

Doug-Lyons:
## Fix automatic toggling of date display on second stage setup

_Update comctrl from Wine with fix._

JIRA issue: [CORE-17199](https://jira.reactos.org/browse/CORE-17199)

### DIFF
--- a/dll/win32/comctl32/datetime.c
+++ b/dll/win32/comctl32/datetime.c
@@ -44,7 +44,6 @@
 #include "commctrl.h"
 #include "comctl32.h"
 #include "wine/debug.h"
-#include "wine/unicode.h"
 
 WINE_DEFAULT_DEBUG_CHANNEL(datetime);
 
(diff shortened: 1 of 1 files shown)


JoachimHenze:
I do confirm that this PR is binary 1:1 to what I tested in https://jira.reactos.org/browse/CORE-17199?focusedCommentId=128218&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-128218
Reduces failure count in winetests:datetime from 1 to 0.
And properly fixed https://jira.reactos.org/browse/CORE-17199 for me when I did apply it locally.

JoachimHenze (approved):

dll/win32/comctl32/datetime.c
```diff
@@ -127,12 +126,12 @@ extern int MONTHCAL_CalculateDayOfWeek(SYSTEMTIME *date, BOOL inplace);
 
 static BOOL DATETIME_SendSimpleNotify (const DATETIME_INFO *infoPtr, UINT code);
 static BOOL DATETIME_SendDateTimeChangeNotify (const DATETIME_INFO *infoPtr);
-static const WCHAR allowedformatchars[] = {'d', 'h', 'H', 'm', 'M', 's', 't', 'y', 'X', 0};
+static const WCHAR allowedformatchars[] = L"dhHmMstyX";
```
learn-more, line 129: wine finally stopped obfuscating their strings?

HBelusca: yes since few months now.

SergeGautherie:
Would you be able to use the winesync script?

HBelusca (commented):
My review would be "LGTM", barred the fact that I don't like the named descriptors in the structures, so I will neither approve or reject.

Doug-Lyons:
> > Sync datetime.c to Wine-6.0
> 
> Why not to Wine Staging?
> Would you be able to use the winesync script?

@SergeGautherie, My understanding is that Wine Staging has been side-lined and its code is being integrated over time into Wine by a Wine Developer. This may not be exactly correct, but it is just my understanding of what happened.
I am not familiar with how to use the wine-sync script, but I think that this works for where there are many exceptions (#ifdef's) in ReactOS to the Wine code and would do little good when we are in perfect sync which is the case here. Thanks.

Doug-Lyons:
@learn-more, Is there any chance that you might be able to help with getting this committed? It seems there has been sufficient time for review. It is a direct Wine sync and has been shown to provide benefits. Thanks. Doug Lyons

SergeGautherie:
> @SergeGautherie, My understanding is that Wine Staging has been side-lined and its code is being integrated over time into Wine by a Wine Developer. This may not be exactly correct, but it is just my understanding of what happened.

Ah? I didn't know. Did Wine make some kind of announcement or the like?

> I am not familiar with how to use the wine-sync script, but I think that this works for where there are many exceptions (#ifdef's) in ReactOS to the Wine code and would do little good when we are in perfect sync which is the case here.

Not sure: I think this new script is about better tracking sync/regressions, no matter `__REACTOS__` or not.

Doug-Lyons:
> > My understanding is that Wine Staging has been side-lined and its code is being integrated over time into Wine by a Wine Developer.
> 
> Ah? I didn't know. Did Wine make some kind of announcement or the like?
> 
> > I am not familiar with how to use the wine-sync script, but I think that this works for where there are many exceptions (#ifdef's) in ReactOS to the Wine code and would do little good when we are in perfect sync which is the case here.
> 
> Not sure: I think this new script is about better tracking sync/regressions, no matter `__REACTOS__` or not.

@SergeGautherie, If you look here: https://en.wikipedia.org/wiki/Wine_(software) you will see this statement:

Since January 2017, patches in wine-staging begins to be actively merged into the WineHQ upstream as wine-compholio transferred the project to Alistair Leslie-Hughes, a key WineHQ developer.

I am not sure how exactly to interpret this, so I leave that up to you.

I have no comments on the wine-sync script since I am totally unfamiliar with it. Maybe someone who understands its purpose could provide a Wiki page about how to use it.

Thanks, Doug Lyons

PR #3676, opened 2021-05-22
[NTOS:SE] Do some cleanup on ICIF flags in token information classes

GeoB99:
A few of these classes have fixed size lengths, the rest are arbitrary. Also the TokenAuditPolicy class hasn't a size length type specified in the table, which is wrong (and move the corresponding TOKEN_AUDIT_POLICY_INFORMATION structure into the private header).


zefklop (approved):
I checked in the Query & Set functions and this looks good to me

PR #3677, opened 2021-05-23
[NTOS:SE] Implement logon session deletion

GeoB99:
This implements `SepRmDeleteLogonSession`. Also declare a prototype for `SepCleanupLUIDDeviceMapDirectory` and annotate it with SAL.

### DIFF
--- a/ntoskrnl/se/srm.c
+++ b/ntoskrnl/se/srm.c
@@ -35,6 +35,11 @@ SepRmCommandServerThread(
 
 static
 NTSTATUS
+SepCleanupLUIDDeviceMapDirectory(
+    _In_ PLUID LogonLuid);
+
+static
+NTSTATUS
 SepRmCreateLogonSession(
     PLUID LogonLuid);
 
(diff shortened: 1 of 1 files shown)


ntoskrnl/se/srm.c
```diff
@@ -376,17 +381,96 @@ SepRmCreateLogonSession(
...
+    for (SessionToDelete = SepLogonSessions;
+         SessionToDelete != NULL;
+         SessionToDelete = SessionToDelete->Next)
+    {
+        /*
+         * Does the actual logon session exist in the
+         * saved logon sessions database with the LUID
+         * provided?
+         */
+        if (!RtlEqualLuid(&SessionToDelete->LogonId, LogonLuid))
```
wiktorek140, line 424: It might cause troubles when multiple logon session will be set, and if we iterate on all of this, one can be different.
Then we skip other in list, and targeted one might be missed

GeoB99: Yeah, I realized quitting the loop aggressively which could leave behind potential logon sessions with the expected LUID can definitely lead to problems. I'll correct it asap.

ntoskrnl/se/srm.c
```diff
@@ -376,17 +381,96 @@ SepRmCreateLogonSession(
...
+    Status = STATUS_SUCCESS;
+
+Leave:
+    /* Release the database lock */
+    KeReleaseGuardedMutex(&SepRmDbLock);
+
+    if (NT_SUCCESS(Status))
+    {
+        ExFreePoolWithTag(SessionToDelete, SEP_LOGON_SESSION_TAG);
+    }
```
HBelusca, line 471: Shouldn't this be moved before the "Leave" label?

zefklop (approved):
Nothing obviously wrong, though I'm not too familiar with this.

ntoskrnl/se/srm.c
```diff
@@ -376,17 +381,100 @@ SepRmCreateLogonSession(
...
+ * STATUS_NO_SUCH_LOGON_SESSION is returned if the logon session with the submitted
+ * LUID doesn't exist. STATUS_BAD_LOGON_SESSION_STATE is returned if the logon session
+ * is still in use and we're not allowed to delete it. Otherwise a NTSTATUS failure code
+ * is returned.
+ */
 static
 NTSTATUS
 SepRmDeleteLogonSession(
-    PLUID LogonLuid)
+    _In_ PLUID LogonLuid)
```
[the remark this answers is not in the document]
GeoB99: As far as I'm concerned, yes. Windows doesn't check if the logon ID provided matches with one of these two internal logon sessions, presumably because they'd never pass such LUIDs to that function. Although we can do way better than that, I could add a check myself for it just in case but I'd like to hear you an opinion from this.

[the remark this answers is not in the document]
HBelusca: > As far as I'm concerned, yes. Windows doesn't check if the logon ID provided matches with one of these two internal logon sessions

And how do you know that? :trollface: Have you written a test that validates this?

PR #3679, opened 2021-05-24
[NTOS:CC] Fix use of unintialized variable (caught by RTC1)

tkreuzer:
Fix use of unintialized variable (caught by RTC1)

### DIFF
--- a/ntoskrnl/cc/view.c
+++ b/ntoskrnl/cc/view.c
@@ -1045,7 +1045,7 @@ CcFlushCache (
         {
             if (vacb->Dirty)
             {
-                IO_STATUS_BLOCK VacbIosb;
+                IO_STATUS_BLOCK VacbIosb = { 0 };
                 Status = CcRosFlushVacb(vacb, &VacbIosb);
                 if (!NT_SUCCESS(Status))
                 {


SergeGautherie:
Where/How is it used uninitialized exactly?

Code after line 397 does look suspicious!

binarymaster:
> Where/How is it used uninitialized exactly?

`CcRosFlushVacb()` does not modify `VacbIosb` because it's just input parameter.
```
NTSTATUS
CcRosFlushVacb (
    _In_ PROS_VACB Vacb,
    _In_ PIO_STATUS_BLOCK Iosb)
```

binarymaster (approved):

binarymaster:
This code part looks suspicious though, because... why it's needed if we just use zeroed-out variable.
```c
                if (IoStatus)
                    IoStatus-&gt;Information += VacbIosb.Information;
```

SergeGautherie:
> `CcRosFlushVacb()` does not modify `VacbIosb` because it's just input parameter.

Could you double-check that?
As far as I can see, this annotation on `CcRosFlushVacb()` and `MmFlushSegment()` is wrong:
it should be `_Out_opt_`.

> This code part looks suspicious though, because... why it's needed if we just use zeroed-out variable.

This comment does not make sense to me, then.

binarymaster:
> Could you double-check that?

You can't check the function code yourself? :roll_eyes:

SergeGautherie:
> You can't check the function code yourself? 🙄

So you confirm
```
4679     if (Iosb)
4680         Iosb->Information = 0;
```
looks like an input param to you? 🙄

HBelusca (commented):
Interesting....
Such a similar situation happens for the MmFlushSegment call, just below (lines 1065 to 1085) but the code verifier / RTC doesn't catch it as potential variable misusage?

tkreuzer:
Guys, I fixed a bug, I don't have any intend to review all that code to validate everything related to it is correct. If you think there is something else broken, you are free to create your own PR. Comments like this are the reason people don't want to go through PRs for small changes.

ThFabba:
Seriously though, I don't see how this could be uninitialized.
* If `CcRosFlushVacb` fails, `CcFlushCache` does `goto quit` and the variable goes out of scope
* In `CcRosFlushVacb`, the `Iosb` parameter is only referenced in the call to `MmFlushSegment`. The only exit point before that call returns a failure status
* In `MmFlushSegment`, we first have an exit opportunity, but it returns failure. We then initialize `Iosb->Information` (which must be the culprit because no other member is ever read); and every return path in the function also sets `Iosb->Status`.

I see two reads, one `+=` of `Information` in `MmFlushSegment`, and the read of `Information` in `CcFlushCache`. It appears neither of them can be uninitialized.
I'm worried there might have been some stack corruption that overwrote a `Status` return value to success, leading us to falsely returning success somewhere, and indeed accessing `Information` uninitialized.

Am I missing something? Do you have any additional info about where this RTC fault happened / what the other locals' values were?

SergeGautherie:
> I fixed a bug

Well, you never gave any details about assumed bug...

> If you think there is something else broken, you are free to create your own PR.

Sure, these were only suggestions here:
I have already created a PR for annotations, I'll create another one for the other case...

PR #3682, opened 2021-05-24
Merge a bunch of commits from amd64-bringup branch

tkreuzer:
This is stuff that I consider ready to merge. Everythig is x64 only!

### DIFF
--- a/ntoskrnl/ke/amd64/stubs.c
+++ b/ntoskrnl/ke/amd64/stubs.c
@@ -260,10 +260,7 @@ NtSyscallFailure(void)
 
 PVOID
 KiSystemCallHandler(
-    _In_ ULONG64 ReturnAddress,
-    _In_ ULONG64 P2,
-    _In_ ULONG64 P3,
-    _In_ ULONG64 P4)
+    VOID)
 {
     PKTRAP_FRAME TrapFrame;
     PKSERVICE_TABLE_DESCRIPTOR DescriptorTable;
@@ -274,12 +271,6 @@ KiSystemCallHandler(
 
     /* Get a pointer to the trap frame */
     TrapFrame = (PKTRAP_FRAME)((PULONG64)_AddressOfReturnAddress() + 1 + MAX_SYSCALL_PARAMS);
-
-    /* Save some values in the trap frame */
-    TrapFrame->Rip = ReturnAddress;
-    TrapFrame->Rdx = P2;
-    TrapFrame->R8 = P3;
-    TrapFrame->R9 = P4;
 
     /* Increase system call count */
     __addgsdword(FIELD_OFFSET(KIPCR, Prcb.KeSystemCalls), 1);
@@ -388,11 +379,11 @@ KiSystemCallHandler(
                 break;
 
             default:
-                __debugbreak();
+                ASSERT(FALSE);
                 break;
         }
     }
-    _SEH2_EXCEPT(1)
+    _SEH2_EXCEPT(EXCEPTION_EXECUTE_HANDLER)
     {
         TrapFrame->Rax = _SEH2_GetExceptionCode();
         return (PVOID)NtSyscallFailure;
--- a/ntoskrnl/ke/amd64/trap.S
+++ b/ntoskrnl/ke/amd64/trap.S
@@ -747,10 +747,10 @@ EXTERN KiSystemCallHandler:PROC
  *--*/
 PUBLIC KiSystemCallEntry64
 .PROC KiSystemCallEntry64
-
-    /* Old stack pointer is in rcx, lie and say we saved it in rbp */
-    .setframe rbp, 0
-    .endprolog
+    /* The unwind info pretends we have a machine frame */
+    .PUSHFRAME
+    .ALLOCSTACK (KTRAP_FRAME_LENGTH + MAX_SYSCALL_PARAM_SIZE - MachineFrameLength)
+    .ENDPROLOG
 
     /* Swap gs to kernel, so we can access the PCR */
     swapgs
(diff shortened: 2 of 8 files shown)


ntoskrnl/ke/amd64/kiinit.c
```diff
@@ -32,7 +32,7 @@ KIPCR KiInitialPcr;
 UCHAR DECLSPEC_ALIGN(16) P0BootStackData[KERNEL_STACK_SIZE] = {0};
 UCHAR DECLSPEC_ALIGN(16) KiDoubleFaultStackData[KERNEL_STACK_SIZE] = {0};
 ULONG_PTR P0BootStack = (ULONG_PTR)&P0BootStackData[KERNEL_STACK_SIZE];
-ULONG_PTR KiDoubleFaultStack = (ULONG_PTR)&KiDoubleFaultStackData[KERNEL_STACK_SIZE];
+ULONG_PTR KiDoubleFaultStack = (ULONG_PTR)&KiDoubleFaultStackData[KERNEL_STACK_SIZE - 8];
```
HBelusca, line 35: Hmmm...
Why does such shift seems "necessary" there, but not in x86? https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/ke/i386/kiinit.c;hb=2047cf361327fe1efc46ef58e2ac0e8595d7d468#l19

ntoskrnl/ke/amd64/stubs.c
```diff
@@ -423,4 +423,29 @@ NtVdmControl(IN ULONG ControlCode,
...
 
+    /* Save the volatile register context in the trap frame */
+    KeContextToTrapFrame(Context,
+                         NULL,
+                         TrapFrame,
+                         Context->ContextFlags,
+                         RequestorMode);
+
+    /* The target frame is 0x80 bytes before the trap frame */
+    TargetFrame = (ULONG64)TrapFrame - 0x80;
```
HBelusca, line 447: Isn't there some "sizeof(...)"/"FIELD_OFFSET" or some define to use instead of this hardcoded value?

ntoskrnl/ke/amd64/kiinit.c
tkreuzer: x86 probably doesn't use SSE instructions in the kernel that need 16 byte alignment, so this doesn't apply. It also depends on how the trap entry allocates stack. But after double checking I think this change is actually wrong. I just don't get too many double fault aborts these days to notice 😄 
Double faults push an error code on the stack, so the machine frame is 48 bytes large. The handler allocates 0x160 bytes (same on Windows) of stack, which should leave the stack 16 byte aligned.
I'll revert that part.

ntoskrnl/ke/amd64/stubs.c
tkreuzer: I'll have a look, but this commit breaks compilation anyway, so it will be deferred.

ntoskrnl/ke/amd64/usercall.c
```diff
@@ -253,8 +253,11 @@ KeUserModeCallback(
     /* Enter a SEH Block */
     _SEH2_TRY
     {
-        /* Calculate and align the stack size */
-        UserArguments = (PUCHAR)ALIGN_DOWN_POINTER_BY(OldStack - ArgumentLength, sizeof(PVOID));
+        /* Calculate and align the stack. This is unaligned by 8 bytes, since the following
+           UCALLOUT_FRAME compensates for that and on entry we already have a full stack
+           frame with home space for the next call, i.e. we are already inside the function
+           body and the stack needs to be 16 byte aligned. */
+        UserArguments = (PUCHAR)ALIGN_DOWN_POINTER_BY(OldStack - ArgumentLength, 16) - 8;
```
zefklop, line 260: Maybe we should have UCALLOUT_FRAME have a 16-bytes aligned size.

ntoskrnl/ke/amd64/trap.S
```diff
@@ -611,7 +611,7 @@ PUBLIC KiRetireDpcListInDpcStack
 
     /* Switch stack and call the function */
     mov rsp, rdx
-    sub rsp, 40
+    sub rsp, 48
```
zefklop, line 614: Why do we need to reserve this much stack anyway ? How did it work before ?

tkreuzer: The caller needs to reserve 32 bytes for the callee (home space). Since on function entry the stack alignment is off by 8 bytes, a non-leaf function must allocate at least 40 bytes to make the stack aligned in the function body. This is what I replicated here first, but the new stack is aligned on function entry. 32 is probably fine, but those extra 16 bytes won't hurt anyone.

zefklop: Well, maybe add a comment there. Someone might see this code and wonder why we need 16 bytes more...

ntoskrnl/ke/amd64/usercall.c
tkreuzer: It's based on information in ksamd64.inc from the Windows SDK:
```
;
;  Usermode Callout User Frame Definitions
;

CkBuffer equ 00020H
CkLength equ 00028H
CkApiNumber equ 0002CH
CkRsp equ 00048H
CkRip equ 00030H
CalloutFrameLength equ 00058H
```
This matches this result (from reactos x64):
```
kd> dt -r _UCALLOUT_FRAME
nt!_UCALLOUT_FRAME
   +0x000 P1Home           : Uint8B
   +0x008 P2Home           : Uint8B
   +0x010 P3Home           : Uint8B
   +0x018 P4Home           : Uint8B
   +0x020 Buffer           : Ptr64 Void
   +0x028 Length           : Uint4B
   +0x02c ApiNumber        : Uint4B
   +0x030 MachineFrame     : _MACHINE_FRAME
      +0x000 Rip              : Uint8B
      +0x008 SegCs            : Uint2B
      +0x00a Fill1            : [3] Uint2B
      +0x010 EFlags           : Uint4B
      +0x014 Fill2            : Uint4B
      +0x018 Rsp              : Uint8B
      +0x020 SegSs            : Uint2B
      +0x022 Fill3            : [3] Uint2B

```

zefklop: Alright then.

ntoskrnl/ke/amd64/trap.S
```diff
@@ -747,29 +747,34 @@ PUBLIC KiSystemCallEntry64
...
+    mov rax, cr8
+    test eax, eax
+    jz KiSystemCall64Again
+    int HEX(2C)
+#endif
 
-    /* Old stack pointer is in rcx, lie and say we saved it in rbp */
-    .setframe rbp, 0
-    .endprolog
+KiSystemCall64Again LABEL PROC
```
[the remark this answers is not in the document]
tkreuzer: Ah, interesting, didn't know of that one.

sdk/lib/crt/crt.cmake
```diff
@@ -469,7 +469,7 @@ elseif(ARCH STREQUAL "amd64")
         math/amd64/log.S
         math/amd64/log10.S
         math/amd64/pow.S
-        math/amd64/sqrt.S
+        # math/amd64/sqrt.S
```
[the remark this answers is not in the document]
tkreuzer: Ìn fact nothing. I thought it was broken, but I just tested it and it seems to work.

sdk/lib/crt/math/amd64/sqrt.S
```diff
@@ -35,7 +35,7 @@ sqrt:
     ja short x_is_nan
 
     /* All is well, calculate the sqrt */
-    sqrtpd xmm0, xmm0
+    sqrtsd xmm0, xmm0
```
zefklop, line 38: Well, this works, but it needs a fix anyway ;-)

zefklop (approved):

PR #3683, opened 2021-05-24
[NTOS:CC][NTOS:MM] 2 functions: Fix Iosb annotation

SergeGautherie:
Noticed while checking #3679.

### DIFF
--- a/ntoskrnl/cc/view.c
+++ b/ntoskrnl/cc/view.c
@@ -158,7 +158,7 @@ CcRosTraceCacheMap (
 NTSTATUS
 CcRosFlushVacb (
     _In_ PROS_VACB Vacb,
-    _In_ PIO_STATUS_BLOCK Iosb)
+    _Out_opt_ PIO_STATUS_BLOCK Iosb)
 {
     NTSTATUS Status;
     BOOLEAN HaveLock = FALSE;
--- a/ntoskrnl/include/internal/cc.h
+++ b/ntoskrnl/include/internal/cc.h
@@ -310,7 +310,10 @@ CcMdlWriteComplete2(
 );
 
 NTSTATUS
-CcRosFlushVacb(PROS_VACB Vacb, PIO_STATUS_BLOCK Iosb);
+CcRosFlushVacb(
+    _In_ PROS_VACB Vacb,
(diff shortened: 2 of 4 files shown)


ntoskrnl/cc/view.c
```diff
@@ -158,7 +158,7 @@ CcRosTraceCacheMap (
 NTSTATUS
 CcRosFlushVacb (
     _In_ PROS_VACB Vacb,
-    _In_ PIO_STATUS_BLOCK Iosb)
+    _Out_opt_ PIO_STATUS_BLOCK Iosb)
```
binarymaster, line 161: Show me where in this function `Iosb` parameter gets written into.

binarymaster (dismissed):

ntoskrnl/cc/view.c
SergeGautherie: Really?
https://github.com/reactos/reactos/blob/c3dbcde1e45ab4c5c2d977a92fda1160bb7571e6/ntoskrnl/cc/view.c#L178-L181

binarymaster: For some reason I don't think it's a write operation.

SergeGautherie: What do you think it is, then?

GeoB99: `MmFlushSegment` writes data into the I/O status block the information field and status code respectively, only if such block is actually filled to the call. 
https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/mm/section.c;hb=d07203ecb700a11c21608448d2326ea6a6b3f368#l4662

In the case of `CcRosFlushVacb` we write back to the I/O status block whatever stuff and the caller can do whatever it wants with this variable after it's been written to it. With that said it'd certainly make sense to treat this argument as `_Out_`.

binarymaster: That was not clear for me, thanks for the explanation. :+1:

GeoB99 (approved):
LGTM but take my review with a slight grain of salt regarding whether a I/O status block is mandatory on return when flushing the VACB structure or not. Maybe @zefklop can help us deciding that.

tkreuzer (approved):

PR #3684, opened 2021-05-24
[WIN32K][USER32] Change mechanism for storing dialog info pointer

tkreuzer:
DLGWINDOWEXTRA is 30 (both on win32 and win64). This has storage for the following entries: DWLP_MSGRESULT (0), DWLP_DLGPROC (8), DWLP_USER (16)
We used to store the dialog info pointer using SetWindowLongPtr (DWLP_ROS_DIALOGINFO == DWLP_USER+sizeof(ULONG_PTR) == 24), which was fine on win32, but failed on win64, since there wasn't enough space left (24 + 8 = 32 > 30).
Rewrite the way the DLGINFO pointer is stored, by adding an additional field to the WND structure and set it using NtUserxSetDialogPointer (which is what it is for).
Also fix too small cbWndExtra for the button class.


win32ss/user/ntuser/simplecall.c
```diff
@@ -798,9 +798,9 @@ NtUserCallHwndParam(
             UserRefObjectCo(pWnd, &Ref);
 
             if (pWnd->head.pti->ppi == PsGetCurrentProcessWin32Process() &&
-                pWnd->cbwndExtra == DLGWINDOWEXTRA &&
```
HBelusca, line 801: Please keep this check (could be "improved" to `pWnd->cbwndExtra >= DLGWINDOWEXTRA` because it is possible to ask for more than this size) for compatibility with Windows, that is such that when the size is larger than `DLGWINDOWEXTRA` it indicates it's a dialog (even if the Win32 dialog data structure is not stored in there).

Check https://devblogs.microsoft.com/oldnewthing/20031113-00/?p=41843 and
https://stackoverflow.com/questions/866692/if-you-add-extra-data-space-to-a-dialogbox-class-be-accessed-by-getwindowlongptr

win32ss/user/user32/windows/dialog.c
```diff
@@ -141,31 +139,21 @@ DIALOGINFO *DIALOG_get_info( HWND hWnd, BOOL create )
        return NULL;
     }
 
-    dlgInfo = (DIALOGINFO *)GetWindowLongPtrW( hWnd, DWLP_ROS_DIALOGINFO );
+    dlgInfo = pWindow->DialogPointer;
 
     if (!dlgInfo && create)
     {
-        if (pWindow && pWindow->cbwndExtra >= DLGWINDOWEXTRA)
```
HBelusca, line 148: Similarly this check might be kept for the same reasons as I mentioned above ("validate that it's intended to be an actual dialog window").

win32ss/user/ntuser/simplecall.c
tkreuzer: "That's not what Windows does!"™

HBelusca: I'm *not* saying that they store the extra dialog data in that space (they don't); I'm just saying there's a check of the cbwndExtra size just to *validate* it's a dialog.

tkreuzer: Changed accordingly.

win32ss/user/user32/windows/dialog.c
```diff
@@ -141,31 +139,21 @@ DIALOGINFO *DIALOG_get_info( HWND hWnd, BOOL create )
...
-        {
-            if (!(dlgInfo = HeapAlloc( GetProcessHeap(), HEAP_ZERO_MEMORY, sizeof(*dlgInfo) )))
-                return NULL;
+        if (!(dlgInfo = HeapAlloc( GetProcessHeap(), HEAP_ZERO_MEMORY, sizeof(*dlgInfo) )))
```
katahiromz, line 146: Indent.

win32ss/include/ntuser.h
```diff
@@ -726,6 +726,8 @@ typedef struct _WND
     PSBINFOEX pSBInfoex; // convert to PSBINFO
     /* Entry in the list of thread windows. */
     LIST_ENTRY ThreadListEntry;
+
+    PVOID DialogPointer;
```
zefklop, line 730: Shouldn't we use a specific struct for Dialogs (or just say that this is behind it) ? You're making every allocated WND bigger but only dialogs use it. Or would it be erased for e.g. buttons ?

tkreuzer: The DIALOGINFO structure is only defined in dialog.c. Windows also stores it in the WND structure.

zefklop: Fine

PR #3691, opened 2021-05-27
[GITHUB] build.yml: Properly set and use LLVM_VERSION

SergeGautherie:
- Restore actually requesting v12, not default
- Avoid gcc configure
```
CMake Warning:
  Manually-specified variables were not used by the project:

    CLANG_VERSION
```

Addendum to a66214c.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,6 +37,7 @@ jobs:
         wget https://apt.llvm.org/llvm.sh
         chmod +x llvm.sh
         sudo ./llvm.sh $LLVM_VERSION
+        echo "D_CLANG_VERSION=-DCLANG_VERSION=$LLVM_VERSION" >> $GITHUB_ENV
     - name: Source checkout
       uses: actions/checkout@v2
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -30,13 +30,16 @@ jobs:
...
+      if: ${{ matrix.compiler == 'clang' }}
+      run: echo "LLVM_VERSION=12" >> $GITHUB_ENV
     - name: Install LLVM
       if: ${{ matrix.compiler == 'clang' }}
       run: |
-        echo "LLVM_VERSION=12" >> $GITHUB_ENV
         wget https://apt.llvm.org/llvm.sh
         chmod +x llvm.sh
         sudo ./llvm.sh $LLVM_VERSION
+        echo "D_CLANG_VERSION=-DCLANG_VERSION=$LLVM_VERSION" >> $GITHUB_ENV
```
zefklop, line 42: If I'm not mistaken, LLVM_VERSION is not used anywhere else than on configure step. Just set D_CLANG_VERSION there and this is it, no need to add more steps.

Maybe rename this to something more generic, like "CMAKE_EXTRA_ARGS" or something.

SergeGautherie: > LLVM_VERSION is not used anywhere else than on configure step.

It is used on this very `Install LLVM` step: hence the current regression, since the step merge.

> Maybe rename this to something more generic, like "CMAKE_EXTRA_ARGS" or something.

Maybe, though I would prefer to keep it obvious for the time being: not having to support concatenation/... on this var.

.github/workflows/build.yml
```diff
@@ -30,13 +30,16 @@ jobs:
...
     - name: Install ccache
       run: sudo apt install ccache
+    - name: Set LLVM version
+      if: ${{ matrix.compiler == 'clang' }}
+      run: echo "LLVM_VERSION=12" >> $GITHUB_ENV
     - name: Install LLVM
       if: ${{ matrix.compiler == 'clang' }}
       run: |
-        echo "LLVM_VERSION=12" >> $GITHUB_ENV
         wget https://apt.llvm.org/llvm.sh
```
zefklop, line 39:
```suggestion
        export LLVM_VERSION=12
        wget https://apt.llvm.org/llvm.sh
```

.github/workflows/build.yml
zefklop: With the added suggestion, you only have one step and you set version once

.github/workflows/build.yml
SergeGautherie: Done.

.github/workflows/build.yml
SergeGautherie: Right.

zefklop:
Thanks

PR #3693, opened 2021-05-28
[USBPORT] Make unload work

tkreuzer:
This PR makes the USBPORT driver unload, when it fails to load.
This is needed for x64, where USB doesn't work yet.

### DIFF
--- a/drivers/usb/usbport/usbport.c
+++ b/drivers/usb/usbport/usbport.c
@@ -1390,6 +1390,11 @@ USBPORT_WorkerThread(IN PVOID StartContext)
                               FALSE,
                               NULL);
 
+        if (FdoExtension->Flags & USBPORT_FLAG_WORKER_THREAD_EXIT)
+        {
+            break;
+        }
+
         KeQuerySystemTime(&NewTime);
 
         KeAcquireSpinLock(&FdoExtension->WorkerThreadEventSpinLock, &OldIrql);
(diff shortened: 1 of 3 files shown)


Extravert-ir (approved):

drivers/usb/usbport/usbport.c
```diff
@@ -1390,6 +1390,11 @@ USBPORT_WorkerThread(IN PVOID StartContext)
                               FALSE,
                               NULL);
 
+        if (FdoExtension->Flags & USBPORT_FLAG_WORKER_THREAD_EXIT)
```
zefklop, line 1393: Then what's the purpose of the while condition at line 1420 ?

Extravert-ir: Hm yeah, feels like this can be done via existing flag `USBPORT_FLAG_WORKER_THREAD_ON`

Extravert-ir: Huh, and the logic of this flag seem to be inverted...

tkreuzer: Yeah, don't ask me. I added something that works.

ThFabba: Yeah this flag is definitely misnamed.
Similar commit from an old branch of mine: https://github.com/ThFabba/reactos/commit/993037e0701dfcb527bb95239405f98f2357a120
Feel free to ignore it though, yours looks fine.

tkreuzer: Ah, the good old patches from 2 years ago. Almost fresh in comparison to some of my older ones 😄 I'll commit this one, which works good enough for the x64 branch. I'll leave it to you or anyone else to improve on it.

PR #3695, opened 2021-05-28
[MPR][MSXML3][USP10] Remove 3 *_ros.diff

SergeGautherie:
These are obsolete.
Clean the way for further work.


GeoB99:
Have you contacted the responsible author(s) of these diffs if you can remove them or not?

SergeGautherie:
> Have you contacted the responsible author(s) of these diffs if you can remove them or not?

I have not: as far as I can see/know, these 2 authors are long gone.
NB: That may explain why these files are unmaintained.

PR #3698, opened 2021-05-29
Amd64/set trap context

tkreuzer:
This PR implements the following functions for x64:
- `RtlpCaptureNonVolatileContextPointers`, which walks the stack and captures the addresses of all non-volatile registers on the stack, when they have been saved first. This is needed to be able to fix up the non-volatile registers on a system call, which doesn't capture non-volatiles, but relies on them to be restored by the callees.
- `RtlSetUnwindContext`, which uses RtlpCaptureNonVolatileContextPointers to set the non-volatile registers in the the stack. They will be picked up, when returning back or unwinding, e.g. to the system call handler.
- `KiSetTrapContext`, which is an asm wrapper around RtlSetUnwindContext, which first stores an exception frame to assure that all non-volatile registers were put on the stack, then calls RtlSetUnwindContext to update their first saving positions on the stack and finally restore the exception frame to potentially load any updated registers, that haven't been saved elsewhere on the stack.

Finally it fixes `PspGetOrSetContextKernelRoutine` to use the above mechanism to change user mode context in the trap frame as a result of a call to `NtSetContextThread` / `PsSetContextThread`

I hope this sounds complicated enough to avoid too many irrelevant comments 😏


ntoskrnl/ps/amd64/psctx.c
```diff
@@ -13,83 +13,65 @@
...
-//        TrapFrame = (PKTRAP_FRAME)((ULONG_PTR)Thread->InitialStack -
-//                                   ROUND_UP(sizeof(KTRAP_FRAME), KTRAP_FRAME_ALIGN) -
-//                                   sizeof(FX_SAVE_AREA));
+        /* Get the thread's base trap frame */
+        TrapFrame = KeGetTrapFrame(KeGetCurrentThread());
     }
 
     /* Check if it's a set or get */
-    if (Apc->SystemArgument1)
+    if (Apc->SystemArgument1 != 0)
```
HBelusca, line 60:
```suggestion
    if (Apc->SystemArgument1 != NULL)
```

ntoskrnl/ke/amd64/trap.S
```diff
@@ -1049,6 +1049,31 @@ KiConvertToGuiThreadFailed:
 
 ENDFUNC
 
+
+EXTERN KiSetTrapContextInternal:PROC
+
+/*
+ * VOID
+ * KiSetTrapContext (
+ *     _In_ PCONTEXT Context);
```
HBelusca, line 1058: Isn't the prototype
```
VOID
KiSetTrapContext(
    _Out_ PKTRAP_FRAME TrapFrame,
    _In_ PCONTEXT Context,
    _In_ KPROCESSOR_MODE RequestorMode);
```
instead?

sdk/lib/rtl/amd64/unwind.c
```diff
@@ -687,3 +687,71 @@ RtlRaiseException(IN PEXCEPTION_RECORD ExceptionRecord)
     RtlRaiseStatus(Status);
 }
 
+VOID
+RtlCaptureNonVolatileContextPointers(
```
HBelusca, line 691: Is this a name from official MS pdb symbols? If not, and if this used only internally, wouldn't it be better to use the `Rtlp` prefix instead?

ntoskrnl/ke/amd64/context.c
```diff
@@ -292,3 +292,29 @@ KeTrapFrameToContext(IN PKTRAP_FRAME TrapFrame,
...
+    _In_ KPROCESSOR_MODE RequestorMode)
+{
+    ULONG64 TargetFrame;
+
+    /* Save the volatile register context in the trap frame */
+    KeContextToTrapFrame(Context,
+        NULL,
+        TrapFrame,
+        Context->ContextFlags,
+        RequestorMode);
```
HBelusca, line 313:
```suggestion
    KeContextToTrapFrame(Context,
                         NULL,
                         TrapFrame,
                         Context->ContextFlags,
                         RequestorMode);
```

ntoskrnl/ke/amd64/context.c
```diff
@@ -292,3 +292,29 @@ KeTrapFrameToContext(IN PKTRAP_FRAME TrapFrame,
...
+
+    /* The target frame is 0x80 bytes before the trap frame */
+    TargetFrame = (ULONG64)TrapFrame - 0x80;
```
HBelusca, line 316: So, no #define or sthg for the "magic" 0x80 ?

ntoskrnl/ps/amd64/psctx.c
tkreuzer: Tn this case I actually prefer 0, since it's not a pointer, but an option stored in a pointer.

HBelusca: Dunno... I see the type of the variable being PVOID 🤷

ntoskrnl/ke/amd64/context.c
tkreuzer: Fine, you'll get it.

ntoskrnl/ps/amd64/psctx.c
tkreuzer: Yeah, but I see what it's meaning is.

sdk/lib/rtl/amd64/unwind.c
tkreuzer: Changed.

ntoskrnl/ke/amd64/context.c
tkreuzer: kk

ntoskrnl/ke/amd64/context.c
```diff
@@ -292,3 +292,29 @@ KeTrapFrameToContext(IN PKTRAP_FRAME TrapFrame,
     if (OldIrql < APC_LEVEL) KeLowerIrql(OldIrql);
 }
 
+VOID
+RtlSetUnwindContext(
+    _In_ PCONTEXT Context,
+    _In_ DWORD64 TargetFrame);
```
zefklop, line 298: why DWORD64 ? PKTRAP_FRAME became a bad word ?

zefklop: Errm, I now see it's a call site. Whaterver, then.

ntoskrnl/ps/amd64/psctx.c
```diff
@@ -13,83 +13,65 @@
...
-{   
-    PAGED_CODE();
- 
-    /* Convert the trap frame to a context */
-    KeTrapFrameToContext(TrapFrame, NULL, Context);
-}
+KiSetTrapContext(
+    _Out_ PKTRAP_FRAME TrapFrame,
+    _In_ PCONTEXT Context,
+    _In_ KPROCESSOR_MODE RequestorMode);
```
zefklop, line 20: can we have this in a header ?

sdk/lib/rtl/amd64/unwind.c
```diff
@@ -687,3 +687,72 @@ RtlRaiseException(IN PEXCEPTION_RECORD ExceptionRecord)
...
+    KPROCESSOR_MODE Mode;
+
+    /* Zero out the nonvolatile context pointers */
+    RtlZeroMemory(NonvolatileContextPointers, sizeof(*NonvolatileContextPointers));
+
+    /* Capture the current context */
+    RtlCaptureContext(&Context);
+
+    /* Use the highest bit in RIP as mode flag */
+    Mode = Context.Rip >> 63;
```
zefklop, line 710: RtlpGetMode() ? This looks fishy.

sdk/lib/rtl/amd64/unwind.c
```diff
@@ -687,3 +687,72 @@ RtlRaiseException(IN PEXCEPTION_RECORD ExceptionRecord)
...
+                         ImageBase,
+                         Context.Rip,
+                         FunctionEntry,
+                         &Context,
+                         &HandlerData,
+                         &EstablisherFrame,
+                         NonvolatileContextPointers);
+
+        /* Continue until we reached the target frame or user mode */
+    } while ((EstablisherFrame < TargetFrame) &&
```
zefklop, line 729: Are we sure that the frame will always go upwards ? What if there is some strange stuff done with setjmp/longjmp ?

tkreuzer: longjmp is an unwind and actually goes through RtlUnwindEx (on Windows). Strange stuff with the stack is not allowed on x64.

zefklop: OK.

sdk/lib/rtl/amd64/unwind.c
tkreuzer: I simplified it and added some ASSERTs instead.

ntoskrnl/ps/amd64/psctx.c
tkreuzer: Done

PR #3700, opened 2021-05-30
[WIN32SS][NTUSER] ShowWindow.SW_MINIMIZE should show window

katahiromz:
## Purpose
Improve compatibility of `user32!ShowWindow` function.
JIRA issue: [CORE-15669](https://jira.reactos.org/browse/CORE-15669)

## Proposed changes

- `user32!ShowWindow.SW_MINIMIZE` should show the window.
- Fix the return value of `ShowWindow` function on invalid parameter.

## TODO

- [x] Do test in depth.

## Comparison

BEFORE:
before-1
There were some failures.
AFTER:
after-1
Successful.

BEFORE:
before-2
The setup program of DVD Write Now won't show the taskbar pane.
AFTER:
after-2
The setup program of DVD Write Now shows the taskbar pane.
Win2k3:
Win2k3

### DIFF
--- a/win32ss/user/ntuser/winpos.c
+++ b/win32ss/user/ntuser/winpos.c
@@ -2551,9 +2551,8 @@ co_WinPosShowWindow(PWND Wnd, INT Cmd)
          Swp |= SWP_NOACTIVATE | SWP_NOZORDER;
          /* Fall through. */
       case SW_SHOWMINIMIZED:
+      case SW_MINIMIZE: /* CORE-15669: SW_MINIMIZE also shows */
          Swp |= SWP_SHOWWINDOW;
-         /* Fall through. */
-      case SW_MINIMIZE:
          {
             Swp |= SWP_NOACTIVATE;
             if (!(style & WS_MINIMIZE))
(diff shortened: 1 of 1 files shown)


win32ss/user/ntuser/winpos.c
```diff
@@ -2648,7 +2647,7 @@ co_WinPosShowWindow(PWND Wnd, INT Cmd)
 
       default:
          //ERR("co_WinPosShowWindow Exit Good 4\n");
-         return WasVisible;
+         return FALSE;
```
HBelusca, line 2650: Is this codepath also tested via the (existing) tests?

katahiromz: Yes. See BEFORE and AFTER comparison above.  `user32_apitest ShowWindow`

HBelusca (approved):

katahiromz:
I will wait for response of @JoachimHenze.

katahiromz:
https://build.reactos.org/#/builders/5/builds/102

JoachimHenze (approved):
I tested the PR3700 a bit the fix on top of 0.4.15-dev-2620-g5ce9e2e,
(So I implicitly tested also the already merged PR3697 as well at the same time)
good: it works for CORE-15669, it satisfies some more tests.I also retested some other related tickets and could not quicky identify any case that would behave worse than before. I think we should commit it.

While testing I had an interesting finding CORE-17611, but that in no way should make us believe, this PR here wouldn't be good. Let's commit it! Approved!

PR #3711, opened 2021-06-01
[SHELL32] Improve and fix ShellExecCmdLine and ShellExecuteEx

katahiromz:
## Purpose

Improve Run dialog and fix CORE-17351. 
JIRA issue: [CORE-17351](https://jira.reactos.org/browse/CORE-17351), [CORE-16898](https://jira.reactos.org/browse/CORE-16898), [CORE-17612](https://jira.reactos.org/browse/CORE-17612)

## Proposed changes

- Simplify `ShellExecCmdLine` function.
- Accept URLs that starts with `www.`.
- Use `PathGetArgsW`, `PathRemoveArgsW`, and `PathUnquoteSpacesW` instead of `SplitParams`.
- Add extra search on `ShellExecuteEx`.

## TODO

- [ ] Do test in depth.

## Comparison

before-0
BEFORE:
before-1
AFTER:
after-1

### DIFF
--- a/dll/win32/shell32/shlexec.cpp
+++ b/dll/win32/shell32/shlexec.cpp
@@ -3,7 +3,7 @@
  *
  * Copyright 1998 Marcus Meissner
  * Copyright 2002 Eric Pouech
- * Copyright 2018-2019 Katayama Hirofumi MZ
+ * Copyright 2018-2021 Katayama Hirofumi MZ
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
@@ -780,6 +780,16 @@ static UINT SHELL_FindExecutable(LPCWSTR lpPath, LPCWSTR lpFile, LPCWSTR lpVerb,
         lpFile = xlpFile;
         /* The file was found in one of the directories in the system-wide search path */
     }
+#ifdef __REACTOS__
+    else
+    {
+        lstrcpynW(xlpFile, lpFile, _countof(xlpFile));
+        if (PathFindOnPathExW(xlpFile, NULL, 0xFF))
+            lpFile = xlpFile;
+        else
+            xlpFile[0] = 0;
+    }
+#endif
 
     attribs = GetFileAttributesW(lpFile);
     if (attribs != INVALID_FILE_ATTRIBUTES && (attribs & FILE_ATTRIBUTE_DIRECTORY))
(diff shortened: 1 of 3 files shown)


katahiromz:
Evidence: https://jira.reactos.org/secure/attachment/59582/w2k3-api-monitor-google-without-https.txt

JoachimHenze:
The after-picture with the entry-point might require us to create a fresh ticket, looks like a regression, although I don't want to imply that it was caused by your commit ofc.

Regarding your commit: Let us experiment a bit more with that before committing: A few test recommendations:
* type "command.com" in the rundlg. Does it start the interpreter like it should? Or does it (now) interpret as a website unintentionally?
* place a file C:\reactos\hello.bat with content just "pause" and type just "hello" without .bat extension in the rundlg. Does it start like it should?
* to test spaces in the command line, test with "regsvr32 /u zipfldr.dll", does it work as expected?

HBelusca:
> Accept URLs without http:// and https://.

This isn't what happens on Windows (Windows will complain, e.g. for `google.com` without the http(s) prefix, that the file `google.com` doesn't exist). It works only **when** there is a `www.` in front.

katahiromz:
Win2k3:
win2k3-command-com
ReactOS:
reactos-command-com
Successful.

Win2k3:
reactos-hello-bat
ReactOS:
win2k3-hello-bat
Failed.

ReactOS:
reactos-regsvr32
Win2k3:
win2k3-regsvr32
Successful.

katahiromz:
I can remember the WHERE command. If `ShellExecuteEx` could search the path to execute, then the search is unnecessary.

HBelusca:
Regarding the hello.bat failure on reactos, did this happen before?

JoachimHenze:
> Regarding the hello.bat failure on reactos, did this happen before?

Yes that case does fail also on master head, I just checked and created https://jira.reactos.org/browse/CORE-17612
This does not invalidate this PR. It can be committed also without passing that test!

JoachimHenze:
Regarding the failure of the "www.google.de" test...
Does it succeed **with your patch** if you install Firefox 28 beforehand from rapps and configure it to be your default browser?
Because that fails in vanilla master also still with "file not found" (and is regression CORE-17351 actually. I tested back then: It regressed even for Firefox with 0.4.10-dev-419-gbfcbda2).
If it would work with FF28 with your patch, then I would consider the patch to be ok and worth committing.

In both cases: An unrelated ticket should be created for the "procedure entry point thing" in the Gecko-iexplore-case. I suspect that to be a regression, and I think it is unrelated to your changes!

katahiromz:
VirtualBox_ReactOS_02_06_2021_06_43_21
It works on Forefox 28.

katahiromz:
I will fix also file execution. Please wait.

katahiromz:
`hello.bat`:
hello
`iexplore`:
iexplore

JoachimHenze:
Wow, seems like you really aim for the "one-fixes-all"-fix now. I added CORE-17612 to the description since you seem to be able to handle that case as well now after your most recent updates on the PR.
I admit, I am impressed about what you reached so far, but I hope, we are not too brave here and start to touch too much at once.
Can you attach a before and after comparison with the testbots, while I'd do run some more manual tests with the current state?

katahiromz:
https://build.reactos.org/#/builders/9/builds/29753
https://build.reactos.org/#/builders/9/builds/29755

JoachimHenze:
Well then it would be best to try to solve those additional test failures first before investing more time into additional manual testing.
For me it is also okay to have a smaller patch just, that might fix just one or two of the regressions (and not all 3 tickets), if we could have something that wouldn't increase test failures.
I am more interested in the regression fixes, then into the hello-improvement e.g.

Thanks for your endurance, I do follow your work here with great interest!

JoachimHenze:
The "Test WHS" builder seems to strengthen our trust into how shell32:ShellExecCmdLine is currently implemented.
The tests are most likely not entirely wrong, as that bot (which runs 2k3sp2) does pass all of that suite atm:
https://reactos.org/testman/compare.php?ids=77429

katahiromz:
https://github.com/reactos/reactos/commit/f37b94ba391b58f0466993a1483ab0942771c365 is committed to `master`.

katahiromz:
Closed due to reconstruction.
I will retry.

PR #3714, opened 2021-06-02
Amd64/apic hal

tkreuzer:
This fixes interrupt allocation on x64 and makes the network adapter work.


hal/halx86/acpi/busemul.c
```diff
@@ -93,10 +93,18 @@ HalpGetSystemInterruptVector_Acpi(IN ULONG BusNumber,
                                  OUT PKIRQL Irql,
                                  OUT PKAFFINITY Affinity)
 {
+#ifdef _M_AMD64
```
zefklop, line 96: don't we have HAL_APIC define or something ?

tkreuzer: No, it's all compiled as libraries, that are then linked together appropriately for each hal. I think this function should go to the PIC library and both should have the same name.

hal/halx86/apic/apictrap.S
```diff
@@ -22,23 +22,6 @@ PUBLIC ApicSpuriousService
 ApicSpuriousService:
     iretq
 
-PUBLIC HackEoi
```
HBelusca, line 25: But does that mean, then, that if one were to test ROS with HAL APIC on "old" vbox, where in principle, Windows 2003 with APIC HAL does work, we would again get problems?

hal/halx86/apic/apic.c
```diff
@@ -94,6 +94,7 @@ ULONG
 IOApicRead(UCHAR Register)
 {
     /* Select the register, then do the read */
+    ASSERT(Register <= 0x3F);
     *(volatile UCHAR *)(IOAPIC_BASE + IOAPIC_IOREGSEL) = Register;
```
Extravert-ir, line 98:
```suggestion
    *(volatile ULONG *)(IOAPIC_BASE + IOAPIC_IOREGSEL) = Register;
```

hal/halx86/apic/apic.c
```diff
@@ -372,8 +352,69 @@ HalpAllocateSystemInterrupt(
...
+        /* Calculate IRQL */
+        NT_ASSERT(HalpVectorToIndex[Vector] == BusInterruptLevel);
+        *OutIrql = HalpVectorToIrql(Vector);
+    }
+    else
+    {
+        ULONG Offset;
+
+        /* Outer loop to find alternative slots, when all IRQLs are in use */
+        for (Offset = 0; Offset < 15; Offset++)
```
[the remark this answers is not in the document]
HBelusca: Oh please... This is going to break compilation in some compilers.

hal/halx86/apic/apictrap.S
tkreuzer: w2k3 apic hal uses the same EOI mechanism. And yes, on old VBox it won't work. But who cares?

HBelusca: What version(s) are what you call "old VBox"? Is VBox 4.x or VBox 5.x in that range?

HBelusca: Also regarding the EOI mechanism, VGal in his last HAL PR has the mechanism Win2k3 actually uses.

hal/halx86/CMakeLists.txt
```diff
@@ -61,6 +61,7 @@ if(ARCH STREQUAL "i386")
     add_hal(hal SOURCES pic/halpic.rc COMPONENTS generic legacy up pic)
     add_hal(halacpi SOURCES acpi/halacpi.rc COMPONENTS generic acpi up pic)
     add_hal(halaacpi SOURCES apic/halaacpi.rc COMPONENTS generic acpi up apic)
+    add_hal(halapic SOURCES apic/halapic.rc COMPONENTS generic legacy up apic)
```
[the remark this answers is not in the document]
tkreuzer: I don't know, who tests the other ones on Hw? Or in a VM? Or at all? I'm fine with removing it though. Windows has it, but that doesn't mean we need to have it, either. I am not aware of any real use case of this anyway, except validating that the different components all work together.

hal/halx86/apic/apic.c
```diff
@@ -372,8 +352,69 @@ HalpAllocateSystemInterrupt(
...
+    ApicWriteIORedirectionEntry(Irq, ReDirReg);
+
+    /* Save irq in the table */
+    HalpVectorToIndex[Vector] = Irq;
+
+    return Vector;
+}
+
+ULONG
+NTAPI
```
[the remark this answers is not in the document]
tkreuzer: Huh?

hal/halx86/apic/apictrap.S
tkreuzer: I don't know the versions. Feel free to test it. Regarding VGal's PR: It's a huge clusterfuck. All that code is falling from the sky and probably mostly straight out of IDA or rewriting code for the sake of rewriting it. It also doesn't change the EOI mechanism, because the one used here is the one used by Windows, as it is the only way to do it.

hal/halx86/pic/pic.c
```diff
@@ -1233,6 +1233,35 @@ HalpDispatchInterrupt2(VOID)
     return NULL;
 }
 
+ULONG
+NTAPI
```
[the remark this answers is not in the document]
tkreuzer: Same as above. Why?

hal/halx86/apic/apic.c
HBelusca: Alternatively,
```suggestion
    WRITE_REGISTER_ULONG(IOAPIC_BASE + IOAPIC_IOREGSEL, Register);
```

binarymaster: I heard one does not simply `xxx_REGISTER_xxx` or `xxx_PORT_xxx` inside critical parts of HAL :ok_hand:

HBelusca: I've heard that when code uses the correct header, these become inlined macros that do the correct thing.
https://git.reactos.org/?p=reactos.git;a=blob;f=sdk/include/ddk/ioaccess.h;hb=46c84779b6b07f1b9a172c6b357eba458770d1a2#l36

hal/halx86/apic/apic.c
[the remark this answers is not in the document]
tkreuzer: I generally don't like this for debugging reasons. When a variable goes out of scope, and then you hit an ASSERT or something, you won't know what the last value of the variable was.

hal/halx86/apic/apic.c
```diff
@@ -94,17 +94,19 @@ ULONG
...
 
 FORCEINLINE
 VOID
 IOApicWrite(UCHAR Register, ULONG Value)
 {
     /* Select the register, then do the write */
-    *(volatile UCHAR *)(IOAPIC_BASE + IOAPIC_IOREGSEL) = Register;
-    *(volatile ULONG *)(IOAPIC_BASE + IOAPIC_IOWIN) = Value;
+    ASSERT(Register <= 0x3F);
+    WRITE_REGISTER_ULONG((PULONG)(IOAPIC_BASE + IOAPIC_IOREGSEL), Register);
```
disean, line 108: Please include the `ioaccess.h` header.

tkreuzer: ioaccess.h is for display drivers, that don't include wdm.h.

disean: It will inline `*_REGISTER_*` functions everywhere. Otherwise you'll have to call the kernel each time you access the APIC registers. `halarm` uses this header.

PR #3685, opened 2021-05-24
[GITHUB] build.yml: Extend MSVC ARM build to more applications

SergeGautherie:
## Purpose

Almost complete Jira ticket, as filed.

JIRA issue: [CORE-17517](https://jira.reactos.org/browse/CORE-17517)

## Proposed changes

- Build all base apps.
Plus, build all ros apps.
- Upload artifacts.
- Prepare a bit more for follow-up ARM64 PR.

PR #3716, opened 2021-06-02
[DRWTSN32] PrintStackBacktrace(): Do not default to i386

SergeGautherie:
(Split off from #3685 )

Sync with other cases.

JIRA issue: [CORE-17605](https://jira.reactos.org/browse/CORE-17605)

### DIFF
--- a/base/applications/drwtsn32/stacktrace.cpp
+++ b/base/applications/drwtsn32/stacktrace.cpp
@@ -38,7 +38,7 @@ void PrintStackBacktrace(FILE* output, DumpData& data, ThreadData& thread)
     DWORD MachineType;
     STACKFRAME64 StackFrame = { { 0 } };
 
-#ifdef _M_X64
+#if defined(_M_AMD64)
     MachineType = IMAGE_FILE_MACHINE_AMD64;
     StackFrame.AddrPC.Offset = thread.Context.Rip;
     StackFrame.AddrPC.Mode = AddrModeFlat;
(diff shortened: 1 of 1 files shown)


learn-more:
And once again: Dont just change the code for the lulz.
Fix a bug or leave it alone.

SergeGautherie:
> And once again: Dont just change the code for the lulz.
> Fix a bug or leave it alone.

?? This PR seems very much part of the actual fix to me.
Unless you would explain why this case has to be different than the other ones: as in what is so great about implicitly defaulting to i386 here?

learn-more:
> > And once again: Dont just change the code for the lulz.
> > Fix a bug or leave it alone.
> 
> ?? This PR seems very much part of the actual fix to me.
> Unless you would explain why this case has to be different than the other ones: as in what is so great about implicitly defaulting to i386 here?

How does this fix anything?
It just exchanges one error for another.

SergeGautherie:
> How does this fix anything?
> It just exchanges one error for another.

It stops trying to compile i386 code on other architectures, like arm/arm64.
It gives 1 explicit error instead of multiple misleading ones, and compilation stops at once instead of getting the actual error on the next case.
If you find the current misbehavior so fine, then why not make the others cases like it?

Extravert-ir:
We don't support arm/arm64 right now. And adding nitpicks which don't really fix the issue don't make the situation any better.
Mark has already said that - fix the real bug, or don't play on our nerves. We prefer real fixes to that chicanery

SergeGautherie:
Eventually included in 1f44552d36f41397991d4594edada88cd062a107.

PR #3718, opened 2021-06-02
[CMAKE] msvc.cmake: Add arm64 support for spec2def

SergeGautherie:
though spec2def does not support it yet.

JIRA issue: [CORE-17615](https://jira.reactos.org/browse/CORE-17615)


Extravert-ir:
This is to be added along with actual support for arm64 in spec2def

SergeGautherie:
> This is to be added along with actual support for arm64 in spec2def

I don't see why one must depend on the other.
NB:
Happy implicit default to i386 :-(

PR #3719, opened 2021-06-03
[DRWTSN32] Build it on supported architectures only

SergeGautherie:
(Split off from #3685 )

Support for other architectures has to be added as needed.

JIRA issue: [CORE-17605](https://jira.reactos.org/browse/CORE-17605)


binarymaster:
No need to disable building a module just for the sake of successful build.

And to make it clear, [CORE-17517](https://jira.reactos.org/browse/CORE-17517) was created by me and resolved as fixed, which means I'm satisfied with the solution, no need for more action, unless actual fixes or implementations.

SergeGautherie:
> No need to disable building a module just for the sake of successful build.

?? It's for the sake of building all the other base applications!
Moreover, what good does it do to try and fail building on unsupported architectures, like arm/arm64?

> And to make it clear, [CORE-17517](https://jira.reactos.org/browse/CORE-17517) was created by me and resolved as fixed, which means I'm satisfied with the solution, no need for more action, unless actual fixes or implementations.

Good for you, while it seems inconsistent to me:
that ticket asked for all base apps, while rapps only was completed yet.
Mark enabled tests too, while that module(s) seems far from ready...

Then, yeah, I feel devs like to commit things mixed up / confusing, then keep them that way as long as possible :-(
And, for some reason, I am not allowed to do the "same": see #3718 for example.

PR #3721, opened 2021-06-03
[SHELL32] Add missing flags in RegGetValueW call

oleg-dubinskiy:
## Purpose

`SHEmptyRecycleBinW` should actually check for `REG_EXPAND_SZ` value of sound event (as it's provided by registry), not only `REG_SZ`.
Our mmsys.cpl always sets the `REG_EXPAND_SZ` values for sound events, same as they are in Windows by default. So the function fails to play the sound event if the value is not `REG_SZ`.
Moreover, it checks for the mismatch of the value type after calling `RegGetValueW`, and fails on this.
It allows to play Recycle Bin sound event properly when user empties the trash (after doing some winesync in winmm also).
Also fix the check for value type mismatch and expand environment variable from the string if it exists there.

## TODO

- [x] Perhaps call `ExpandEnvironmentStringsW` and pass its output to `PlaySoundW`, since the string is expandable. I doubt is this badly required in this case, if the sound event is now playing properly even without doing that.

## Result

For the test, I used Recycle Bin sound from Windows Vista, since ReactOS currently has no appropriate sound for it, and not each sound plays correctly at all, due to limitations of AC97 VirtualBox driver.
As visible in the video, the sound event is now playing successfully after my changes. :slightly_smiling_face: 

https://user-images.githubusercontent.com/26385117/120683234-0c01ec80-c4a6-11eb-954a-fed3808c178f.mp4

### DIFF
--- a/dll/win32/shell32/folders/CRecycleBin.cpp
+++ b/dll/win32/shell32/folders/CRecycleBin.cpp
@@ -1125,14 +1125,14 @@ HRESULT WINAPI SHEmptyRecycleBinW(HWND hwnd, LPCWSTR pszRootPath, DWORD dwFlags)
         ret = RegGetValueW(HKEY_CURRENT_USER,
                            L"AppEvents\\Schemes\\Apps\\Explorer\\EmptyRecycleBin\\.Current",
                            NULL,
-                           RRF_RT_REG_SZ,
+                           RRF_RT_REG_SZ | RRF_RT_REG_EXPAND_SZ | RRF_NOEXPAND,
                            &dwType,
                            (PVOID)szPath,
                            &dwSize);
         if (ret != ERROR_SUCCESS)
             return S_OK;
 
-        if (dwType != REG_EXPAND_SZ) /* type dismatch */
+        if (dwType != REG_EXPAND_SZ || dwType != REG_SZ) /* type dismatch */
             return S_OK;
 
         szPath[_countof(szPath)-1] = L'\0';
--- a/dll/win32/shell32/folders/CRecycleBin.cpp
+++ b/dll/win32/shell32/folders/CRecycleBin.cpp
@@ -1132,7 +1132,7 @@ HRESULT WINAPI SHEmptyRecycleBinW(HWND hwnd, LPCWSTR pszRootPath, DWORD dwFlags)
         if (ret != ERROR_SUCCESS)
             return S_OK;
 
-        if (dwType != REG_EXPAND_SZ || dwType != REG_SZ) /* type dismatch */
+        if (dwType != REG_SZ || dwType != REG_EXPAND_SZ) /* Check whether the type is valid */
             return S_OK;
 
         szPath[_countof(szPath)-1] = L'\0';
--- a/dll/win32/shell32/folders/CRecycleBin.cpp
+++ b/dll/win32/shell32/folders/CRecycleBin.cpp
@@ -1137,7 +1137,7 @@ HRESULT WINAPI SHEmptyRecycleBinW(HWND hwnd, LPCWSTR pszRootPath, DWORD dwFlags)
 
         if (dwType == REG_EXPAND_SZ)
         {
-            if (!ExpandEnvironmentStringsW(szPath, szDest, MAX_PATH))
+            if (!ExpandEnvironmentStringsW(szPath, szDest, _countof(szDest)))
                 return S_OK;
         }
 
--- a/dll/win32/shell32/folders/CRecycleBin.cpp
+++ b/dll/win32/shell32/folders/CRecycleBin.cpp
@@ -1028,7 +1028,7 @@ HRESULT WINAPI SHEmptyRecycleBinA(HWND hwnd, LPCSTR pszRootPath, DWORD dwFlags)
 
 HRESULT WINAPI SHEmptyRecycleBinW(HWND hwnd, LPCWSTR pszRootPath, DWORD dwFlags)
 {
-    WCHAR szPath[MAX_PATH] = {0}, szBuffer[MAX_PATH];
+    WCHAR szPath[MAX_PATH] = {0}, szBuffer[MAX_PATH], szDest[MAX_PATH];
     DWORD dwSize, dwType, count;
     LONG ret;
     IShellFolder *pDesktop, *pRecycleBin;
@@ -1132,11 +1132,24 @@ HRESULT WINAPI SHEmptyRecycleBinW(HWND hwnd, LPCWSTR pszRootPath, DWORD dwFlags)
         if (ret != ERROR_SUCCESS)
             return S_OK;
 
-        if (dwType != REG_SZ || dwType != REG_EXPAND_SZ) /* Check whether the type is valid */
+        if (dwType != REG_SZ && dwType != REG_EXPAND_SZ) /* Check whether the type is valid */
             return S_OK;
 
-        szPath[_countof(szPath)-1] = L'\0';
-        PlaySoundW(szPath, NULL, SND_FILENAME);
+        ERR("szPath is %ls, size %d\n", szPath, dwSize);
+
+        if (dwType == REG_EXPAND_SZ)
+        {
+            if (!ExpandEnvironmentStringsW(szPath, szDest, MAX_PATH))
+            {
+                ERR("ExpandEnvironmentStringsW failed for %ls\n", szPath);
+                return S_OK;
+            }
+        }
+
+        ERR("szDest is %ls\n", szDest);
+
+        szDest[_countof(szDest)-1] = L'\0';
+        PlaySoundW(szDest, NULL, SND_FILENAME);
     }
     return S_OK;
 }
--- a/dll/win32/shell32/folders/CRecycleBin.cpp
+++ b/dll/win32/shell32/folders/CRecycleBin.cpp
@@ -1028,7 +1028,7 @@ HRESULT WINAPI SHEmptyRecycleBinA(HWND hwnd, LPCSTR pszRootPath, DWORD dwFlags)
 
 HRESULT WINAPI SHEmptyRecycleBinW(HWND hwnd, LPCWSTR pszRootPath, DWORD dwFlags)
 {
-    WCHAR szPath[MAX_PATH] = {0}, szBuffer[MAX_PATH], szDest[MAX_PATH];
+    WCHAR szPath[MAX_PATH] = {0}, szBuffer[MAX_PATH];
     DWORD dwSize, dwType, count;
     LONG ret;
     IShellFolder *pDesktop, *pRecycleBin;
@@ -1127,7 +1127,7 @@ HRESULT WINAPI SHEmptyRecycleBinW(HWND hwnd, LPCWSTR pszRootPath, DWORD dwFlags)
                            NULL,
                            RRF_RT_REG_SZ | RRF_RT_REG_EXPAND_SZ | RRF_NOEXPAND,
                            &dwType,
-                           (PVOID)szPath,
+                           (PVOID)szBuffer,
                            &dwSize);
         if (ret != ERROR_SUCCESS)
             return S_OK;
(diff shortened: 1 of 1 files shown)


dll/win32/shell32/folders/CRecycleBin.cpp
```diff
@@ -1125,7 +1125,7 @@ HRESULT WINAPI SHEmptyRecycleBinW(HWND hwnd, LPCWSTR pszRootPath, DWORD dwFlags)
         ret = RegGetValueW(HKEY_CURRENT_USER,
                            L"AppEvents\\Schemes\\Apps\\Explorer\\EmptyRecycleBin\\.Current",
                            NULL,
-                           RRF_RT_REG_SZ,
+                           RRF_RT_REG_EXPAND_SZ | RRF_NOEXPAND,
```
HBelusca, line 1128: And what if the value was REG_SZ instead? And who is going to expand the string by the way?

oleg-dubinskiy: Hmm, perhaps I also should consider this case.

oleg-dubinskiy: As MSDN says, any of these flags can be just combined if we need to retrieve more than one type of data, so I think if to specify both `RRF_RT_REG_SZ` and `RRF_RT_REG_EXPAND_SZ`, it will search for the value of both of these types.

oleg-dubinskiy: Fixed, thanks @HBelusca.

dll/win32/shell32/folders/CRecycleBin.cpp
```diff
@@ -1125,14 +1125,14 @@ HRESULT WINAPI SHEmptyRecycleBinW(HWND hwnd, LPCWSTR pszRootPath, DWORD dwFlags)
...
-                           RRF_RT_REG_SZ,
+                           RRF_RT_REG_SZ | RRF_RT_REG_EXPAND_SZ | RRF_NOEXPAND,
                            &dwType,
                            (PVOID)szPath,
                            &dwSize);
         if (ret != ERROR_SUCCESS)
             return S_OK;
 
-        if (dwType != REG_EXPAND_SZ) /* type dismatch */
+        if (dwType != REG_EXPAND_SZ || dwType != REG_SZ) /* type dismatch */
```
HBelusca, line 1135:
```suggestion
        if (dwType != REG_SZ || dwType != REG_EXPAND_SZ) /* Check whether the type is valid */
```

HBelusca:
> Perhaps call ExpandEnvironmentStringsW and pass its output to PlaySoundW, since the string is expandable. I doubt is this badly required in this case, if the sound event is now playing properly even without doing that.

If PlaySoundW does already the expansion, then it's not necessary to do it there.
Check on Windows, whether passing a non-expanded path to PlaySound, works OK or not.

dll/win32/shell32/folders/CRecycleBin.cpp
oleg-dubinskiy: Committed. :ok_hand:

oleg-dubinskiy:
> > Perhaps call ExpandEnvironmentStringsW and pass its output to PlaySoundW, since the string is expandable. I doubt is this badly required in this case, if the sound event is now playing properly even without doing that.
> 
> If PlaySoundW does already the expansion, then it's not necessary to do it there.
> Check on Windows, whether passing a non-expanded path to PlaySound, works OK or not.

Done @HBelusca.
At least on XP `PlaySoundW` returns success when trying to pass the string without previous `ExpandEnvironmentSringsW` call.
So your guessing seems to be correct. :slightly_smiling_face: 

playsound_test

HBelusca:
> > > Perhaps call ExpandEnvironmentStringsW and pass its output to PlaySoundW, since the string is expandable. I doubt is this badly required in this case, if the sound event is now playing properly even without doing that.

That test you show is obviously not working, since your backslashes in the path string are not escaped, and you are showing me a path that doesn't contain an environment variable to be expanded. So it's also not testing what I was asking.

oleg-dubinskiy:
> > Done @HBelusca.
> > At least on XP `PlaySoundW` returns success when trying to pass the string without previous `ExpandEnvironmentSringsW` call.
> > So your guessing seems to be correct. slightly_smiling_face
> 
> That test you show is obviously not working, since your backslashes in the path string are not escaped, and you are showing me a path that doesn't contain an environment variable to be expanded. So it's also not testing what I was asking.
> 
> EDIT: I've tested the following on Windows 10, please double-check on Windows XP (supposing your system is installed in `C:\Windows` and has the example file):
> 
> ```c
> PlaySoundW(L"C:\\Windows\\Media\\Windows Logon.wav", NULL, SND_FILENAME);
> ```
> 
> This will correctly play whatever sound this file is.
> However, the following (that contains a non-expanded environment variable, like e.g. strings that can be REG_EXPAND_SZ):
> 
> ```c
> PlaySoundW(L"%SystemRoot%\\Media\\Windows Logon.wav", NULL, SND_FILENAME);
> ```
> 
> will **NOT** play the wav sound, but instead will play the default "beep" sound. This is clearly not what is intended.
> This is also what is told in the MSDN documentation:
> 
> > SND_FILENAME  The pszSound parameter is a file name. If the file cannot be found, the function plays the default sound unless the SND_NODEFAULT flag is set.

Confirmed: on Windows XP is the same situation.
So it indeed needs to expand the string before passing it to `PlaySoundW`.

oleg-dubinskiy:
However, after expanding that string and passing a result as a sound path, the sound is playing successfully again. :slightly_smiling_face: 

expanded_string

So I should do the same.

oleg-dubinskiy:
Done @HBelusca. :slightly_smiling_face:

dll/win32/shell32/folders/CRecycleBin.cpp
```diff
@@ -1125,18 +1125,24 @@ HRESULT WINAPI SHEmptyRecycleBinW(HWND hwnd, LPCWSTR pszRootPath, DWORD dwFlags)
...
         if (ret != ERROR_SUCCESS)
             return S_OK;
 
-        if (dwType != REG_EXPAND_SZ) /* type dismatch */
+        if (dwType != REG_SZ && dwType != REG_EXPAND_SZ) /* Check whether the type is valid */
             return S_OK;
 
-        szPath[_countof(szPath)-1] = L'\0';
-        PlaySoundW(szPath, NULL, SND_FILENAME);
+        if (dwType == REG_EXPAND_SZ)
```
HBelusca, line 1138: In this whole code you could use `szBuffer` instead of `szPath`, and then do the expansion into `˙szPath`.
This would save space in the stack.

dll/win32/shell32/folders/CRecycleBin.cpp
```diff
@@ -1125,18 +1125,24 @@ HRESULT WINAPI SHEmptyRecycleBinW(HWND hwnd, LPCWSTR pszRootPath, DWORD dwFlags)
...
+        {
+            if (!ExpandEnvironmentStringsW(szPath, szDest, MAX_PATH))
```
HBelusca, line 1140:
```suggestion
            if (!ExpandEnvironmentStringsW(szPath, szDest, _countof(szDest)))
```

oleg-dubinskiy: Fixed.

dll/win32/shell32/folders/CRecycleBin.cpp
oleg-dubinskiy: OK, so as I understand, I can get rid of `szDest` at all.

HBelusca: yep.

oleg-dubinskiy: Done, thanks.
Sound event still works correctly for me after this change.

dll/win32/shell32/folders/CRecycleBin.cpp
```diff
@@ -1125,18 +1125,29 @@ HRESULT WINAPI SHEmptyRecycleBinW(HWND hwnd, LPCWSTR pszRootPath, DWORD dwFlags)
...
+                return S_OK;
+
+            szPath[_countof(szPath)-1] = L'\0';
+            PlaySoundW(szPath, NULL, SND_FILENAME);
+        }
+        else if (dwType == REG_SZ)
+        {
+            szBuffer[_countof(szBuffer)-1] = L'\0';
+            PlaySoundW(szBuffer, NULL, SND_FILENAME);
+        }
```
learn-more, line 1150:
```suggestion
        }
        szBuffer[_countof(szBuffer)-1] = L'\0';
        PlaySoundW(szBuffer, NULL, SND_FILENAME);
```

dll/win32/shell32/folders/CRecycleBin.cpp
```diff
@@ -1125,18 +1125,29 @@ HRESULT WINAPI SHEmptyRecycleBinW(HWND hwnd, LPCWSTR pszRootPath, DWORD dwFlags)
         ret = RegGetValueW(HKEY_CURRENT_USER,
                            L"AppEvents\\Schemes\\Apps\\Explorer\\EmptyRecycleBin\\.Current",
                            NULL,
-                           RRF_RT_REG_SZ,
+                           RRF_RT_REG_SZ | RRF_RT_REG_EXPAND_SZ | RRF_NOEXPAND,
                            &dwType,
-                           (PVOID)szPath,
+                           (PVOID)szBuffer,
```
SergeGautherie, line 1130: Sync line 1124 too.

dll/win32/shell32/folders/CRecycleBin.cpp
```diff
@@ -1125,18 +1125,29 @@ HRESULT WINAPI SHEmptyRecycleBinW(HWND hwnd, LPCWSTR pszRootPath, DWORD dwFlags)
...
 
-        if (dwType != REG_EXPAND_SZ) /* type dismatch */
+        if (dwType != REG_SZ && dwType != REG_EXPAND_SZ) /* Check whether the type is valid */
             return S_OK;
 
-        szPath[_countof(szPath)-1] = L'\0';
-        PlaySoundW(szPath, NULL, SND_FILENAME);
+        if (dwType == REG_EXPAND_SZ)
+        {
+            if (!ExpandEnvironmentStringsW(szBuffer, szPath, _countof(szPath)))
```
SergeGautherie, line 1140: This is a wrong way of checking this return value.

dll/win32/shell32/folders/CRecycleBin.cpp
```diff
@@ -1125,18 +1125,29 @@ HRESULT WINAPI SHEmptyRecycleBinW(HWND hwnd, LPCWSTR pszRootPath, DWORD dwFlags)
         ret = RegGetValueW(HKEY_CURRENT_USER,
                            L"AppEvents\\Schemes\\Apps\\Explorer\\EmptyRecycleBin\\.Current",
                            NULL,
-                           RRF_RT_REG_SZ,
+                           RRF_RT_REG_SZ | RRF_RT_REG_EXPAND_SZ | RRF_NOEXPAND,
```
SergeGautherie, line 1128: Does Windows actually support both types for this value?

dll/win32/shell32/folders/CRecycleBin.cpp
```diff
@@ -1125,18 +1125,29 @@ HRESULT WINAPI SHEmptyRecycleBinW(HWND hwnd, LPCWSTR pszRootPath, DWORD dwFlags)
...
             return S_OK;
 
-        szPath[_countof(szPath)-1] = L'\0';
-        PlaySoundW(szPath, NULL, SND_FILENAME);
+        if (dwType == REG_EXPAND_SZ)
+        {
+            if (!ExpandEnvironmentStringsW(szBuffer, szPath, _countof(szPath)))
+                return S_OK;
+
+            szPath[_countof(szPath)-1] = L'\0';
```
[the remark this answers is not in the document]
learn-more: it is not tho.

PR #3725, opened 2021-06-04
[SOUNDS] Add new Recycle Bin system sound

oleg-dubinskiy:
## Purpose

Add new sound for Recycle Bin sound event.
It's the modern (XP equivalent) of two Recycle Bin sounds I implemented. The 2nd one is a classic (9x-2000) version equivalent, attached in the ticket as well.
Since this sound event is now working properly after some fixes in shell32 code from my previous #3721 PR, it's now possible for user to set the sound for this sound event manually.
But don't enable it by default, since it is also disabled on Windows Server 2003.

JIRA issue: [CORE-15868](https://jira.reactos.org/browse/CORE-15868)

## Proposed changes

- Add the sound file itself.
- Enable it in media/CMakeLists.txt.

## TODO

- [ ] Probably choose a license for this sound file, since it's made by myself, and I didn't decide yet. (which is better? or I should just use the same license as of original source (MIT)?) Link to the original sound: https://freesound.org/s/426815/.

## Result

As visible in the attached video, this sound is playing successfully when user empties Recycle Bin, if he set it in mmsys.cpl previously. :slightly_smiling_face: 

https://user-images.githubusercontent.com/26385117/120822189-d4a34680-c55e-11eb-9ff9-54fa1fdf4e6c.mp4

(watch in Chrome)


JoachimHenze:
Don't get me wron, but it sounds a bit like a Diablo II necromancer that is raising skelletons. ;)
Can we make it sound a bit more like rumpling **paper** ?

oleg-dubinskiy:
> Don't get me wron, but it sounds a bit like a Diablo II necromancer that is raising skelletons. ;)
> Can we make it sound a bit more like rumpling **paper** ?

I doubt. Maybe it reminds something other different from paper at first glance, but I found the good (in my opinion) recorded sound of real paper. For me, it seems to be the most similar to original XP sound.
Maybe I just did never play DIablo II, so don't know how that sounds. :smile: 
OK, if needed, I can try to find another similar sound on freesound.org, where I also found the source of this one.

binarymaster (commented):
This PR doesn't seem to add `ReactOS_Recycle.wav` to the actual sound scheme, it only adds the file itself and CMake listing. Please fix it.

oleg-dubinskiy:
@binarymaster: I purposely didn't include it in default sound scheme and enable by default, because it isn't done on Windows Server 2003: the default sound scheme contains only logon sound there (same as in ReactOS). Other sounds can be set manually.
And at all, it also depends on appearance settings set by user (all sounds are enabled or not).
So is it necessarily switch to XP behaviour? I think it is better to keep 2003 one.

binarymaster:
This stuff should depend on server/workstation choice then I think.

oleg-dubinskiy:
> This stuff should depend on server/workstation choice then I think.

Yes, and seems should be done in another PR.

binarymaster:
> Yes, and seems should be done in another PR.

There is no much point for two separate PRs that are intended for the same thing, especially when it's so trivial.

oleg-dubinskiy:
> > Yes, and seems should be done in another PR.
> 
> There is no much point for two separate PRs that are intended for the same thing, especially when it's so trivial.

OK, then how I can change it only for Workstation here? :thinking:

learn-more:
This sounds like a thing to be done for the component that also allows switching between workstation and server?
Execute `server.inf` or `workstation.inf` or so...

oleg-dubinskiy:
I know! I just can implement setting sound events in registry in syssetup, since it is responsible for choosing Server/Workstation settings at 2nd setup stage. It seems to be the best place to do that.
I guess it still is better to submit it in a separate PR.

learn-more:
That makes sense.

binarymaster (approved):

PR #3726, opened 2021-06-05
[CPL][APPWIZ] Make gecko download cancellable by keyboard

katahiromz:
## Purpose

JIRA issue: [CORE-17550](https://jira.reactos.org/browse/CORE-17550), [CORE-5737](https://jira.reactos.org/browse/CORE-5737)

## Proposed changes

- Move the focus by sending `WM_NEXTDLGCTL` before making the focus control disabled or hidden.

## Screenshots

Start download of gecko by pressing `Enter` key...
VirtualBox_ReactOS_05_06_2021_17_12_50
And cancel by keyboard `Esc` key.
VirtualBox_ReactOS_05_06_2021_17_13_03
OK.

### DIFF
--- a/dll/cpl/appwiz/addons.c
+++ b/dll/cpl/appwiz/addons.c
@@ -388,10 +388,17 @@ static DWORD WINAPI download_proc(PVOID arg)
 
 static INT_PTR CALLBACK installer_proc(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lParam)
 {
-    HWND hwndInstallButton;
+    HWND hwndProgress, hwndInstallButton;
     switch(msg) {
     case WM_INITDIALOG:
-        ShowWindow(GetDlgItem(hwnd, ID_DWL_PROGRESS), SW_HIDE);
+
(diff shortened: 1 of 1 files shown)


JoachimHenze (approved):
For the record: It is the download that cannot be canceled by keyboard.
That initial question beforehand (Do you want to DL+install Gecko can be answered with Cancel by pressing tab+left-arrow+enter). Therefore I tweaked the description slightly,
But aside from that nitpicking, I think your PR is good.

julenuri:
Is this going to fix [CORE-5737](https://jira.reactos.org/browse/CORE-5737)?
Could be interesting for the only use of keyboard.

katahiromz:
Also fixes CORE-5737.
VirtualBox_ReactOS_05_06_2021_18_12_46

katahiromz:
Thus, we shouldn't disable nor hide the control with focus.

learn-more:
@katahiromz it might be that win32k is actually supposed to move focus to another control in a case like this,
it would be worth investigating if there is already a wine/api test that covers this, and otherwise create one.

PR #3729, opened 2021-06-05
Atl cstring operator !=

learn-more:
## Purpose

Implement 'operator !='

### DIFF
--- a/sdk/lib/atl/cstringt.h
+++ b/sdk/lib/atl/cstringt.h
@@ -533,6 +533,43 @@ public:
         return str2.GetLength() == 1 && str2[0] == ch1;
     }
 
+    friend bool operator!=(const CStringT& str1, const CStringT& str2) throw()
+    {
+        return str1.Compare(str2) != 0;
+    }
+
+    friend bool operator!=(const CStringT& str1, PCXSTR psz2) throw()
+    {
+        return str1.Compare(psz2) != 0;
+    }
+
+    friend bool operator!=(const CStringT& str1, PCYSTR psz2) throw()
+    {
+        CStringT tmp(psz2, str1.GetManager());
+        return tmp.Compare(str1) != 0;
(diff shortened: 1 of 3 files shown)


sdk/lib/atl/cstringt.h
```diff
@@ -533,6 +533,43 @@ class CStringT :
         return str2.GetLength() == 1 && str2[0] == ch1;
     }
 
+    friend bool operator!=(const CStringT& str1, const CStringT& str2) throw()
```
HBelusca, line 536: Hmm... Why not just calling the corresponding `operator==` and NOT-ing the result?

learn-more: because this was easier.

HBelusca: Copy-pasting the code from operator== and manually changing every condition to its negative was "easier"??

learn-more: yes.

sdk/lib/atl/cstringt.h
```diff
@@ -533,6 +533,43 @@ class CStringT :
...
+
+    friend bool operator!=(PCYSTR psz1, const CStringT& str2) throw()
+    {
+        CStringT tmp(psz1, str2.GetManager());
+        return tmp.Compare(str2) != 0;
+    }
+
+    friend bool operator!=(XCHAR ch1, const CStringT& str2) throw()
+    {
+        return str2.GetLength() != 1 || str2[0] != ch1;
```
zefklop, line 570: `return str2[0] != ch || str2[1} != '\0'`
feels more effcient.

learn-more: It does, but some string implementations will assert when you access an item out of range, and I did want to make clear that there was not chance of that.

PR #3732, opened 2021-06-06
[USER32_APITEST] Add SetFocus testcase

katahiromz:
## Purpose

Investigate about setting keyboard focus.
JIRA issue: [CORE-17550](https://jira.reactos.org/browse/CORE-17550)
See also: #3726

## Proposed changes

- Add a testcase for `SetFocus`/`GetFocus`.

## TODO

- [x] Do tests.

## Comparison

WinXP:
winxp
Win2k3:
win2k3
Win10:
win10


modules/rostests/apitests/user32/SetFocus.c
```diff
@@ -0,0 +1,209 @@
...
+    return gui.hwndFocus;
+}
+
+static unsigned __stdcall thread_proc_0(void *arg)
+{
+    HWND hwnd = arg;
+
+    ok_int(GetFocus() == NULL, TRUE);
+    ok_int(GetMainThreadFocus() == GetDlgItem(hwnd, IDOK), TRUE);
+    PostMessageA(hwnd, WM_NEXTDLGCTL, FALSE, FALSE);
```
learn-more, line 29: Why do you use postmessage here?
If you use SendMessage, the message will be processed before the function returns.

katahiromz: Thanks.

PR #3734, opened 2021-06-06
[NTOSKRNL] Use IQS_NONE for non existent classes

GeoB99:

### DIFF
--- a/ntoskrnl/include/internal/icif.h
+++ b/ntoskrnl/include/internal/icif.h
@@ -3,7 +3,7 @@
  * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
  * PURPOSE:     Internal header for information classes info interface
  * COPYRIGHT:   Copyright ???
- *              Copyright 2020 George Bișoc <george.bisoc@reactos.org>
+ *              Copyright 2020-2021 George Bișoc <george.bisoc@reactos.org>
  */
 
 #pragma once
@@ -38,4 +38,4 @@ typedef struct _INFORMATION_CLASS_INFO
   { sizeof(TypeQuery), sizeof(AlignmentQuery), sizeof(TypeSet), sizeof(AlignmentSet), Flags }
 
 #define IQS_NONE \
-  { 0, 0, 0, 0, ICIF_NONE }
+  { 0, sizeof(CHAR), 0, sizeof(CHAR), ICIF_NONE }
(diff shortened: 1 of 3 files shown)


ntoskrnl/include/internal/icif.h
```diff
@@ -38,4 +38,4 @@ typedef struct _INFORMATION_CLASS_INFO
   { sizeof(TypeQuery), sizeof(AlignmentQuery), sizeof(TypeSet), sizeof(AlignmentSet), Flags }
 
 #define IQS_NONE \
-  { 0, 0, 0, 0, ICIF_NONE }
+  { sizeof(CHAR), sizeof(CHAR), sizeof(CHAR), sizeof(CHAR), ICIF_NONE }
```
learn-more, line 41: Would it make sense to implement this macro using `IQS_SAME` for example?
Please note that there is one case that was already using `IQS_NONE`, so you are giving that other values now! (`ThreadDescriptorTableEntry` in non-x86)

ntoskrnl/se/token.c
```diff
@@ -32,7 +32,7 @@ static GENERIC_MAPPING SepTokenMapping = {
 static const INFORMATION_CLASS_INFO SeTokenInformationClass[] = {
 
     /* Class 0 not used, blame MS! */
-    IQS_SAME(0, 0, 0),
+    IQS_NONE,
```
learn-more, line 35: This is now something else?
Because IQS_NONE is no longer 0,0

ntoskrnl/include/internal/icif.h
GeoB99: In principle `IQS_NONE` means there's no query/set information for the respective class, so it's deemed as non existent. Whilst the macro is used in non-x86 `ThreadDescriptorTableEntry`, yes, but if you look at the diff of the PR there are several non existent classes that use `IQS_SAME` with the very same values, which is inconsistent and  `IQS_SAME` shouldn't be used, there's `IQS_NONE` for that purpose and so that we can use `IQS_NONE` in a consistent manner.

Now to answer your question in the other comment, the basic thumb of the rule is -- use `IQS_NONE` if a class doesn't exist and `IQS_SAME` for a class that exists and there's no distinction between Query/Set data values. Now that we're talking about it, I think maybe it'd make sense logically to have a size type value to 0 but keeping CHAR for alignment requirement because an alignment of 1 == CHAR means no alignment is required.

learn-more: That makes sense.

PR #3735, opened 2021-06-07
Implement MmFlushImageSection(MmFlushForDelete)

zefklop:
## Purpose

Fix case when you can't delete an executable file after having executed it.

JIRA issue: [CORE-17544](https://jira.reactos.org/browse/CORE-17544)

## Proposed changes

Actually free the pages lying around when all image sections are deleted and MmFlushImageSection(MmFlushForDelete) is called. This makes the image section drops all its refcount and finally be deleted.

### DIFF
--- a/ntoskrnl/include/internal/mm.h
+++ b/ntoskrnl/include/internal/mm.h
@@ -52,6 +52,11 @@ struct _MM_RMAP_ENTRY;
 typedef ULONG_PTR SWAPENTRY;
 
 //
+// Special IRQL value (found in assertions)
+//
+#define MM_NOIRQL ((KIRQL)0xFFFFFFFF)
+
+//
 // MmDbgCopyMemory Flags
 //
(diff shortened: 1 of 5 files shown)


ntoskrnl/include/internal/mm.h
```diff
@@ -51,6 +51,11 @@ struct _EPROCESS;
 struct _MM_RMAP_ENTRY;
 typedef ULONG_PTR SWAPENTRY;
 
+//
+// Special IRQL value (found in assertions)
+//
+#define MM_NOIRQL (KIRQL)0xFFFFFFFF
```
HBelusca, line 57:
```suggestion
#define MM_NOIRQL ((KIRQL)0xFFFFFFFF)
```

HBelusca:
Does this also fix driver unloading?

zefklop:
> Does this also fix driver unloading?

No. Why would it ?

PR #3738, opened 2021-06-09
[HIDPARSER] Don't include pshpack1.h before including other headers

tkreuzer:
NEVER DO THIS! It is guaranteed to be wrong. Instead always individually pack single structures that need packing.
This fixes USB mouse on 64 bit builds.

### DIFF
--- a/sdk/lib/drivers/hidparser/parser.h
+++ b/sdk/lib/drivers/hidparser/parser.h
@@ -2,7 +2,6 @@
 #define _HIDPARSER_H_
 
 #include <wdm.h>
-#include <pshpack1.h>
 #define _HIDPI_
 #define _HIDPI_NO_FUNCTION_MACROS_
 #include <hidpddi.h>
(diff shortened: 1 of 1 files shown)


sdk/lib/drivers/hidparser/parser.h
```diff
@@ -126,6 +127,7 @@ typedef struct
 
 }MAIN_ITEM_DATA, *PMAIN_ITEM_DATA;
 
+//#include <pshpack1.h>
```
HBelusca, line 130: Is this because you're not sure? It's certainly not a hardware structure (it contains embedded pointers), so I think it's safe to completely remove that packing there.

If not then all the data before should be in a own (packed) structure, and that global item thing would then include that data as one member, then the embedded pointer.

HBelusca (approved):

sdk/lib/drivers/hidparser/parser.h
tkreuzer: Oops, that is a test remainder :) Will remove

binarymaster (approved):

katahiromz (approved):

PR #3744, opened 2021-06-13
[DBGHELP] Fix compilation error on clang version 11.0.1

ohhmm:
## Purpose

Fixed minor compilation issue (clang).

JIRA issue: none

## Proposed changes

Explicit type cast.

## TODO

-

### DIFF
--- a/dll/win32/dbghelp/compat.c
+++ b/dll/win32/dbghelp/compat.c
@@ -174,7 +174,7 @@ BOOL __ReadFile(HANDLE file, PVOID buffer, DWORD len, PDWORD outlen, PVOID overl
 
     read = fread(buffer, 1, len, file);
 
-    if (ferror(file) != 0)
+    if (ferror((FILE*)file) != 0)
         return FALSE;
 
     *outlen = (DWORD)read;


.gitignore
```diff
@@ -10,5 +10,6 @@ modules/[Aa][Hh][Kk]_[Tt]ests
 .cproject
 .project
 .settings
+build/
```
binarymaster, line 13: gitignore change should be done in a separate commit

binarymaster (changes requested):
Please fix commit message according to our [commit message style](https://github.com/reactos/reactos/blob/master/.gitmessage). I already fixed PR title for you.

dll/win32/dbghelp/compat.c
```diff
@@ -174,7 +174,7 @@ BOOL __ReadFile(HANDLE file, PVOID buffer, DWORD len, PDWORD outlen, PVOID overl
 
     read = fread(buffer, 1, len, file);
 
-    if (ferror(file) != 0)
+    if (ferror((FILE*)file) != 0)
```
HBelusca, line 177: Why would this cast be needed only there and not in all the other f*** function calls where `file` HANDLE is passed while the parameter type is actually a `FILE*` ?

HBelusca (changes requested):

dll/win32/dbghelp/compat.c
ohhmm: This commit is for dbghelp only

HBelusca: Yes of course it is for dbghelp, which is precisely this module. Just look at the rest of this file and you'll understand what I mean... (For example, the fread function call just above also takes this `file` parameter, which is in principle a `FILE*`, etc etc.)

SergeGautherie:
> clang version 11.0.1

What is the exact error?

--

For what it's worth, on GitHub:
* Current build requests `LLVM_VERSION=12` (v12.0.0).
* Trying `LLVM_VERSION=11` (v11.1.0) succeeds too.

Then, do we care about v11.0.1?

PR #3745, opened 2021-06-14
[NTOS:MM] Fix placing entries into MmNonPagedPoolFreeListHead

Extravert-ir:
When freeing pages, free page entries with pages num == 3 were incorrectly treated as entries with pages num >= 4 and thus their re-insertion was not triggered. That lead to non paged pool fragmentation (can be triggered by kmtest:ExPools, for example)

Also, altered the index acquisition code for MmNonPagedPoolFreeList entries so it looks more clear

Kudos go to @ThFabba for finding this bug :tada:

### DIFF
--- a/ntoskrnl/mm/ARM3/pool.c
+++ b/ntoskrnl/mm/ARM3/pool.c
@@ -680,8 +680,7 @@ MiAllocatePoolPages(IN POOL_TYPE PoolType,
     //
     // Allocations of less than 4 pages go into their individual buckets
     //
-    i = SizeInPages - 1;
-    if (i >= MI_MAX_FREE_PAGE_LISTS) i = MI_MAX_FREE_PAGE_LISTS - 1;
+    i = min(SizeInPages, MI_MAX_FREE_PAGE_LISTS) - 1;
 
     //
     // Loop through all the free page lists based on the page index
@@ -738,8 +737,7 @@ MiAllocatePoolPages(IN POOL_TYPE PoolType,
                 if (FreeEntry->Size != 0)
                 {
                     /* Check which list to insert this entry into */
-                    i = FreeEntry->Size - 1;
-                    if (i >= MI_MAX_FREE_PAGE_LISTS) i = MI_MAX_FREE_PAGE_LISTS - 1;
+                    i = min(FreeEntry->Size, MI_MAX_FREE_PAGE_LISTS) - 1;
 
                     /* Insert the entry into the free list head, check for prot. pool */
                     if (MmProtectFreedNonPagedPool)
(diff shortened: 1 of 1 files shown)


ThFabba (approved):

HBelusca (approved):

binarymaster (approved):
This makes testbots green, so this makes me happy :D

ntoskrnl/mm/ARM3/pool.c
```diff
@@ -1176,11 +1174,11 @@ MiFreePoolPages(IN PVOID StartingVa)
         }
 
         //
-        // Check if the entry is small enough to be indexed on a free list
+        // Check if the entry is small enough (1-3 pages) to be indexed on a free list
         // If it is, we'll want to re-insert it, since we're about to
         // collapse our pages on top of it, which will change its count
         //
-        if (FreeEntry->Size < (MI_MAX_FREE_PAGE_LISTS - 1))
+        if (FreeEntry->Size < MI_MAX_FREE_PAGE_LISTS)
```
zefklop, line 1181: This one is the bug, right ?

zefklop (approved):
Looks good to me.
NB I'm not too fond of using "min" in kernel code, but no blocking from my side.

ntoskrnl/mm/ARM3/pool.c
```diff
@@ -680,8 +680,7 @@ MiAllocatePoolPages(IN POOL_TYPE PoolType,
     //
     // Allocations of less than 4 pages go into their individual buckets
     //
-    i = SizeInPages - 1;
-    if (i >= MI_MAX_FREE_PAGE_LISTS) i = MI_MAX_FREE_PAGE_LISTS - 1;
+    i = min(SizeInPages, MI_MAX_FREE_PAGE_LISTS) - 1;
```
zefklop, line 683: Why do you change this ? It's not a functional change, is it ?

HBelusca:
"min" is defined as a macro by the way.

zefklop:
> "min" is defined as a macro by the way.

That's why I don't like it. min(x++,y++) doesn't do what you think it should do.

HBelusca:
> > "min" is defined as a macro by the way.
> 
> That's why I don't like it. min(x++,y++) doesn't do what you think it should do.

Indeed, and that's why when you use it, it's only "for simple things" like sticking constants in it, or a single variable without doing any operations on it (which is precisely the case there).

ntoskrnl/mm/ARM3/pool.c
Extravert-ir: Yeah, it is

ntoskrnl/mm/ARM3/pool.c
Extravert-ir: While doing changes, I spent more time understanding what's going on here than it should be (according to me, at least). And decided to change it to a more clear code :)

PR #3747, opened 2021-06-15
[NTOS:CC] CcRosFlushVacb(): 3 improvements related to its uses

SergeGautherie:
Revert #3679 and follow it up.

### DIFF
--- a/ntoskrnl/cc/view.c
+++ b/ntoskrnl/cc/view.c
@@ -1045,7 +1045,7 @@ CcFlushCache (
         {
             if (vacb->Dirty)
             {
-                IO_STATUS_BLOCK VacbIosb = { 0 };
+                IO_STATUS_BLOCK VacbIosb;
                 Status = CcRosFlushVacb(vacb, &VacbIosb);
                 if (!NT_SUCCESS(Status))
                 {
--- a/ntoskrnl/cc/view.c
+++ b/ntoskrnl/cc/view.c
@@ -412,10 +412,9 @@ CcRosFlushDirtyPages (
         if (--SharedCacheMap->OpenCount == 0)
             CcRosDeleteFileCache(SharedCacheMap->FileObject, SharedCacheMap, &OldIrql);
 
-        if (!NT_SUCCESS(Status) && (Status != STATUS_END_OF_FILE) &&
-            (Status != STATUS_MEDIA_WRITE_PROTECTED))
+        if (!NT_SUCCESS(Status))
         {
-            DPRINT1("CC: Failed to flush VACB.\n");
+            DPRINT1("CcRosFlushVacb() failed. Status 0x%08x\n", Status);
         }
         else
         {
--- a/ntoskrnl/cc/view.c
+++ b/ntoskrnl/cc/view.c
@@ -155,6 +155,7 @@ CcRosTraceCacheMap (
 #endif
 }
 
+static
 NTSTATUS
 CcRosFlushVacb (
     _In_ PROS_VACB Vacb,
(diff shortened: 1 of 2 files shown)


ntoskrnl/cc/view.c
```diff
@@ -411,8 +412,7 @@ CcRosFlushDirtyPages (
         if (--SharedCacheMap->OpenCount == 0)
             CcRosDeleteFileCache(SharedCacheMap->FileObject, SharedCacheMap, &OldIrql);
 
-        if (!NT_SUCCESS(Status) && (Status != STATUS_END_OF_FILE) &&
-            (Status != STATUS_MEDIA_WRITE_PROTECTED))
+        if (!NT_SUCCESS(Status))
```
ThFabba, line 415: I see what you're fixing here, but it doesn't seem to relate to the call in `CcFlushCache`. Are you saying Timo fixed the wrong instance?

ntoskrnl/cc/view.c
```diff
@@ -411,8 +412,7 @@ CcRosFlushDirtyPages (
         if (--SharedCacheMap->OpenCount == 0)
             CcRosDeleteFileCache(SharedCacheMap->FileObject, SharedCacheMap, &OldIrql);
 
-        if (!NT_SUCCESS(Status) && (Status != STATUS_END_OF_FILE) &&
-            (Status != STATUS_MEDIA_WRITE_PROTECTED))
```
ThFabba, line 415: These checks that you call "obsolete" were probably there to avoid the DPRINT in case of common failure codes.
Have you tested this? Do you see any debug spam after removing the checks?

ntoskrnl/include/internal/cc.h
```diff
@@ -309,9 +309,6 @@ CcMdlWriteComplete2(
     IN PMDL MdlChain
 );
 
-NTSTATUS
-CcRosFlushVacb(PROS_VACB Vacb, PIO_STATUS_BLOCK Iosb);
-
```
ThFabba, line 314: The original kernel coding style appears to have been "use unique names rather than static functions". I don't think there's consensus either for or against that style, so please leave it alone.

SergeGautherie: I took example after the 3 existing `static CcRos*()` in ntoskrnl/cc/view.c.
Obviously, I can remove that commit, meant to make code clearer, if you confirm it is unwanted.

ntoskrnl/cc/view.c
SergeGautherie: > Are you saying

No, the 3 commits are independent.
I concur with what you wrote on his PR and I have no additional clue about that.
I simply noticed this separate case while reviewing the other one.

ntoskrnl/cc/view.c
SergeGautherie: > avoid the DPRINT

That is what I thought at first, which would mean they would have been misplaced.

> Have you tested this?

Nothing to test here, as these statuses cannot happen since @zefklop's `CcRosFlushVacb()` change.

ntoskrnl/cc/view.c
```diff
@@ -411,8 +412,7 @@ CcRosFlushDirtyPages (
...
+        if (!NT_SUCCESS(Status))
         {
             DPRINT1("CC: Failed to flush VACB.\n");
```
zefklop, line 417: Can you add the status to the debug print while you're at it? In its current state, it's just noise.

ntoskrnl/include/internal/cc.h
zefklop: I prefer static functions. But that's no official coding style or whatever, so well...

ntoskrnl/cc/view.c
SergeGautherie: Added.

GeoB99:
Have you tested this? Also since you are reverting a change committed by Timo, you should better wait for his input on this one.

PR #3750, opened 2021-06-17
[PCI] Filter out devices with null vendor and device ID

binarymaster:
This is how Windows pci.sys works. Fixes [CORE-17636](https://jira.reactos.org/browse/CORE-17636)

### DIFF
--- a/drivers/bus/pci/fdo.c
+++ b/drivers/bus/pci/fdo.c
@@ -151,6 +151,12 @@ FdoEnumerateDevices(
                    PciConfig.VendorID,
                    PciConfig.DeviceID);
 
+            if (PciConfig.VendorID == 0 && PciConfig.DeviceID == 0)
+            {
+                DPRINT("Filter out devices with null vendor and device ID\n");
+                continue;
+            }
+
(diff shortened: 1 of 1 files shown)


drivers/bus/pci/fdo.c
```diff
@@ -151,6 +151,12 @@ FdoEnumerateDevices(
                    PciConfig.VendorID,
                    PciConfig.DeviceID);
 
+            if (PciConfig.VendorID == 0 && PciConfig.DeviceID == 0)
```
HBelusca, line 154: Out of curiosity, could you enable the DPRINTs that are above (line 147, 134, 121) and test that on VPC 2007 to see how the other values come out when this PCI device gets detected?

binarymaster: Here's what I get with debugging enabled:
```
(drivers/bus/pci/fdo.c:124) Bus 0  Device  7  Func 2
(drivers/bus/pci/fdo.c:134) Size 64
(drivers/bus/pci/fdo.c:152) Bus 0  Device  7  Func 2  VenID 0x0000  DevID 0x0000
(drivers/bus/pci/fdo.c:156) Filter out devices with null vendor and device ID
```

tkreuzer (approved):

PR #3751, opened 2021-06-17
[NTOS:SE] Implement logon session termination notification

GeoB99:
**Windows Server 2003 SP2**
Capture
**ReactOS**
Capture2

### DIFF
--- a/modules/rostests/kmtests/ntos_se/SeLogonSession.c
+++ b/modules/rostests/kmtests/ntos_se/SeLogonSession.c
@@ -1,0 +1,35 @@
+/*
+ * PROJECT:     ReactOS kernel-mode tests
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Kernel mode tests for logon session manipulation API
+ * COPYRIGHT:   Copyright 2021 George Bișoc <george.bisoc@reactos.org>
+ */
+
+#include <kmt_test.h>
+#include <ntifs.h>
+
+#define IMAGINARY_LOGON {0x001, 0}
+
+static
+VOID
+LogonMarkTermination(VOID)
+{
+    NTSTATUS Status;
+    LUID SystemLogonID = SYSTEM_LUID;
+    LUID NonExistentLogonID = IMAGINARY_LOGON;
+
+    /* Mark the system authentication logon for termination */
+    Status = SeMarkLogonSessionForTerminationNotification(&SystemLogonID);
+    ok_irql(PASSIVE_LEVEL);
+    ok_eq_hex(Status, STATUS_SUCCESS);
+
+    /* We give a non existent logon */
+    Status = SeMarkLogonSessionForTerminationNotification(&NonExistentLogonID);
+    ok_irql(PASSIVE_LEVEL);
+    ok_eq_hex(Status, STATUS_NOT_FOUND);
+}
+
+START_TEST(SeLogonSession)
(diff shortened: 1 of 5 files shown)


modules/rostests/kmtests/ntos_se/SeLogonSession.c
```diff
@@ -0,0 +1,35 @@
...
+    LUID SystemLogonID = SYSTEM_LUID;
+    LUID NonExistentLogonID = IMAGINARY_LOGON;
+
+    /* Mark the system authentication logon for termination */
+    Status = SeMarkLogonSessionForTerminationNotification(&SystemLogonID);
+    ok_irql(PASSIVE_LEVEL);
+    ok_eq_hex(Status, STATUS_SUCCESS);
+
+    /* We give a non existent logon */
+    Status = SeMarkLogonSessionForTerminationNotification(&NonExistentLogonID);
```
HBelusca, line 27: I'm wondering, whether it is possible to artificially open, then close a small logon session, in order to verify whether the notification will be raised.

sdk/include/xdk/setypes.h
```diff
@@ -675,6 +675,10 @@ typedef struct _SID_AND_ATTRIBUTES_HASH {
 #define NETWORKSERVICE_LUID  {0x3e4, 0x0}
 #define IUSER_LUID           {0x3e3, 0x0}
 
+/* Logon session reference flags */
+
+#define LOGON_SESSION_MARK_FOR_EVENTUAL_TERMINATION   0x0001
```
HBelusca, line 680:
```suggestion
#define SEP_LOGON_SESSION_TERMINATION_NOTIFY    0x0001
```
Calling the function you wrote is the act of "marking" for termination by setting the flag, but the flag name indicates the action that needs to be done.

modules/rostests/kmtests/ntos_se/SeLogonSession.c
GeoB99: I doubt one can simply artificially open/create a logon session, since the internal part of the system is up to create sessions (and the LSA auth package for that matter but that is in user mode though) and not anybody else. But even then, session termination notification is far from being fully complete in ReactOS.

When a session is marked for termination, Windows notifies both LSA and filesystems that a session is about to terminate in logon dereference routine (that is `SepRmDereferenceLogonSession` in our case). This is done by using several worker queuing kernel routines like `ExQueueWorkItem` and such, which the kernel gives the opportunity to both LSA and FSDs to do whatever stuff they have to do once a logon is about to terminate. Our `SepRmDereferenceLogonSession` is not fully implemented because of that.

Given how this stuff works in Windows, my plan is to have it implemented accordingly in ReactOS thorough different PRs for the sake of easier review. First implement this `SeMarkLogonSessionForTerminationNotification` and then fully implement the logon dereference function which notifies LSA and FSDs thorough respective internal subroutines that deal with this stuff.

HBelusca: You may consider writing to Eric in order to obtain advice on how creating/deleting logon sessions as tests can be achieved.

HBelusca: Alternatively I'm wondering whether this is possible from kernel-mode.

GeoB99: I talked with Eric and no, there's no other alternative possible to simulate a logon creation because of how logon implementation is done in Windows.

PR #3752, opened 2021-06-18
Alias SAL annotations to GCC attributes

zefklop:
## Purpose

I want to use `_Must_inspect_result_` on some functions in Mm, and I want to do it right. Since we don't really care about warnings emitted by MSVC, but we do about ones emitted by GCC, let's have some fun and alias `_Must_inspect_result_` to `__attribute__((__warn_unused_result__))` which is equivalent and errors out on compilation.

JIRA issue: [CORE-17637](https://jira.reactos.org/browse/CORE-17637)

## Proposed changes

- Move sal.h to sdk/include/crt
- Introduce a GCC specific sal.h overriding existing definitions
- Fix those in rtl & ntoskrnl
- Fix compilation of WDK, as it appears that GCC doesn't ignore mismatched attributes between declaration & definition anymore (?!)

## TODO

Fix all remaining issues pointed in CORE-17637 so that we can enable the GCC warning  and live happily.


Extravert-ir:
I was also thinking about mapping `_Out_` to `nonnull` attribute, but that may be harder to do

zefklop:
> I was also thinking about mapping `_Out_` to `nonnull` attribute, but that may be harder to do

That may be possible. At least, you have the infrastructure to do that easily.

ntoskrnl/config/i386/cmhardwr.c
```diff
@@ -472,42 +472,48 @@ CmpInitializeMachineDependentConfiguration(IN PLOADER_PARAMETER_BLOCK LoaderBloc
...
+                    if (NT_SUCCESS(RtlAnsiStringToUnicodeString(&Data, &TempString, TRUE)))
+                    {
+                        /* Add it to the registry */
+                        RtlInitUnicodeString(&ValueName, L"ProcessorNameString");
+                        Status = NtSetValueKey(KeyHandle,
+                                            &ValueName,
+                                            0,
+                                            REG_SZ,
+                                            Data.Buffer,
+                                            Data.Length + sizeof(UNICODE_NULL));
```
SergeGautherie, line 484:
```suggestion
                        Status = NtSetValueKey(KeyHandle,
                                               &ValueName,
                                               0,
                                               REG_SZ,
                                               Data.Buffer,
                                               Data.Length + sizeof(UNICODE_NULL));
```
And below too...

sdk/lib/rtl/sysvol.c
```diff
@@ -511,12 +513,15 @@ RtlpSysVolTakeOwnership(IN PUNICODE_STRING DirectoryPath,
...
-                                NULL);
+        /* Disable privileges that we had to enable, whetever the result was. */
+        NTSTATUS Status2 = ZwAdjustPrivilegesToken(hToken,
+                                                   FALSE,
+                                                   &TokenPrivileges,
+                                                   0,
+                                                   NULL,
+                                                   NULL);
+        /* This must succeed */
+        ASSERT(NT_SUCCESS(Status2));
```
SergeGautherie, line 524: Does `Status2` need `#if DBG` or the like?

sdk/lib/drivers/wdf/shared/core/fxdevice.cpp
```diff
@@ -1332,6 +1332,7 @@ FxDevice::_AcquireOptinRemoveLock(
 
 _Must_inspect_result_
 NTSTATUS
+STDCALL
```
Extravert-ir, line 1335: Why nobody complains about that right now? Are these functions not used?

zefklop: My guess is that adding more attributes leads to better checks.

sdk/lib/rtl/sysvol.c
zefklop: I'll `(void)` it, that will be simpler.

base/applications/fltmc/fltmc.cpp
```diff
@@ -220,7 +220,8 @@ PrintVolumeInfo(_In_ PVOID Buffer)
         VolName[FilterVolInfo->FilterVolumeNameLength] = UNICODE_NULL;
     }
 
-    (void)FilterGetDosName(VolName, DosName, 16);
```
HBelusca, line 223: Interesting: Usually this `(void)` thing is precisely used to shut up compiler warnings about the non-checking of the returned result (see GCC, MSVC, where otherwise you would get the warning). But somehow this gets ignored?

base/applications/network/net/cmdUser.c
```diff
@@ -188,7 +188,11 @@ GetTimeInSeconds(VOID)
     GetSystemTimeAsFileTime(&FileTime);
     Time.u.LowPart = FileTime.dwLowDateTime;
     Time.u.HighPart = FileTime.dwHighDateTime;
-    RtlTimeToSecondsSince1970(&Time, &dwSeconds);
+    if (!RtlTimeToSecondsSince1970(&Time, &dwSeconds))
+    {
+        /* Uh ? */
```
HBelusca, line 193: 😄

ntoskrnl/ex/work.c
```diff
@@ -262,13 +263,19 @@ ExpCreateWorkerThread(WORK_QUEUE_TYPE WorkQueueType,
...
+                                  THREAD_ALL_ACCESS,
+                                  NULL,
+                                  NULL,
+                                  NULL,
+                                  ExpWorkerThreadEntryPoint,
+                                  UlongToPtr(Context));
+    if (!NT_SUCCESS(Status))
+    {
+        /* Well... */
+        DPRINT1("Failed to crate worker thread: 0x%08x\n", Status);
```
HBelusca, line 276:
```suggestion
        DPRINT1("Failed to create worker thread: 0x%08x\n", Status);
```

HBelusca: It's strange that there's no complaint on the contrary for the `ObReferenceObjectByHandle` call above where no return value check has been done whatsoever.

ntoskrnl/ex/work.c
```diff
@@ -576,13 +584,17 @@ ExpInitializeWorkerThreads(VOID)
...
+                                  NULL,
+                                  ExpWorkerThreadBalanceManager,
+                                  NULL);
+    if (!NT_SUCCESS(Status))
+    {
+        KeBugCheckEx(PHASE1_INITIALIZATION_FAILED, Status, 0, 0, 0);
+    }
 
     /* Get a pointer to it for the shutdown process */
     ObReferenceObjectByHandle(ThreadHandle,
```
HBelusca, line 600: But here there's no complaint about the absence of return value check? 🤔🤔

ntoskrnl/include/internal/io.h
```diff
@@ -80,6 +80,10 @@
 // Unable to create symbolic link pointing to the RAM disk device
 //
 #define RD_SYMLINK_CREATE_FAILED 5
+//
+// Unable to create sysroot path when creating the ram disk
```
HBelusca, line 84:
```suggestion
// Unable to create system root path when creating the RAM disk
```

base/applications/fltmc/fltmc.cpp
zefklop: If you think your call succeeds anyway, then you either don't put the `warn_unused_result` or `_Must_inspect_result_` attribute in the first place, or you ASSERT on it in debug builds. I think that putting (void) is lazy and stupid, and it turns out that the GCC developers think like I do on this matter.

ntoskrnl/ob/obname.c
```diff
@@ -259,7 +259,7 @@ ObpCreateDosDevicesDirectory(VOID)
      * DosDevices directory \?? , for backward compatibility with
      * Windows NT-2000 systems.
      */
-    RtlCreateUnicodeString(&LinkName, L"\\DosDevices");
+    RtlInitUnicodeString(&LinkName, L"\\DosDevices");
```
HBelusca, line 262: o_O so that was a leaking allocation?!

sdk/include/crt/sal.h
```diff
@@ -0,0 +1,16 @@
...
+ * PROJECT:     ReactOS PSDK
+ * LICENSE:     MIT (https://spdx.org/licenses/MIT)
+ * PURPOSE:     Standard Annotation Language (SAL) definitions
+ * COPYRIGHT:   2021 - Jérôme Gardou
+ */
+
+#pragma once
+
+/* Include MS version first */
+#include "msc/sal.h"
```
HBelusca, line 11: Having moved the sal.h headers out of include/psdk, breaks our psdk compatibility with MS.

sdk/lib/drivers/wdf/shared/core/fxdevice.cpp
```diff
@@ -1546,6 +1547,7 @@ DispatchWorker(
...
```
SergeGautherie, line 1550: Add `#ifdef __REACTOS__` to all MS WDF changes?

base/applications/fltmc/fltmc.cpp
```diff
@@ -220,7 +220,8 @@ PrintVolumeInfo(_In_ PVOID Buffer)
...
+    if (!SUCCEEDED(FilterGetDosName(VolName, DosName, 16)))
+        DosName[0] = L'\0';
```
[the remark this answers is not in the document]
zefklop: I don't like ANSI_NULL and UNICODE_NULL. NULL is for pointers, period.
`0`of `'\0'` is clear to anyone who knows what a C string is like.

SergeGautherie (commented):
> we don't really care about warnings emitted by MSVC

Could this check be enabled as error on MSVC too?

base/applications/fltmc/fltmc.cpp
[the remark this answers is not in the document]
SergeGautherie: Agreed, though my comment is about "magic values". Anyway.

HBelusca:
> let's have some fun and alias _Must_inspect_result_ to __attribute__((__warn_unused_result__)) which is equivalent 

Does this GCC attribute means to check not only for the "direct" return value, but also for any other returned value (e.g. via parameter pointer) ?

ntoskrnl/ex/work.c
zefklop: Because nobody added the `_Must_inspect_result_` annotation to the declaration of `ObReferenceObjectByHandle`

ntoskrnl/ex/work.c
zefklop: See the other comment

ntoskrnl/ob/obname.c
zefklop: Yessir

zefklop:
> Does this GCC attribute means to check not only for the "direct" return value, but also for any other returned value (e.g. via parameter pointer) ?

Only return value. See https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html

HBelusca:
> > Does this GCC attribute means to check not only for the "direct" return value, but also for any other returned value (e.g. via parameter pointer) ?
> 
> Only return value. See https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html

OK, so it's more closely related to the MSVC `_Check_return_` one, according to https://docs.microsoft.com/en-us/cpp/code-quality/annotating-function-behavior?view=msvc-160 .
The `_Must_inspect_result_` appears also to control the returned values by pointer parameters.

zefklop:
> > > Does this GCC attribute means to check not only for the "direct" return value, but also for any other returned value (e.g. via parameter pointer) ?
> > 
> > 
> > Only return value. See https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html
> 
> OK, so it's more closely related to the MSVC `_Check_return_` one, according to https://docs.microsoft.com/en-us/cpp/code-quality/annotating-function-behavior?view=msvc-160 .
> The `_Must_inspect_result_` appears also to control the returned values by pointer parameters.

Right. Although `_Must_inspect_result_` implies `_Check_return_`. I'll update the PR.

sdk/include/crt/sal.h
zefklop: Alright. Although I would have preferred to have this in a compiler-specific directory, but so be it.

zefklop:
> > we don't really care about warnings emitted by MSVC
> 
> Could this check be enabled as error on MSVC too?

This requires static analysis on MSVC build. I opened PR #3596 about this, but nobody seems to like it :disappointed:

sdk/include/crt/sal.h
HBelusca: We can imagine a hybrid solution where the include/psdk also contains compiler-specific subdirectories, and the "master" sal.h is in include/psdk . Then it includes the compiler-specific sal headers that are in include/psdk/msc or gcc. 🤔

zefklop: well, for now there would only be ms_sal.h & gcc_sal.h so I think this can stay as it is. If more files need splitting, we can create a subdirectory, yes.

base/applications/network/nslookup/nslookup.c
```diff
@@ -827,7 +828,12 @@ int main( int argc, char* argv[] )
 
     HeapFree( ProcessHeap, 0, pNetInfo );
 
-    WSAStartup( MAKEWORD(2,2), &wsaData );
+    ret = WSAStartup( MAKEWORD(2,2), &wsaData );
+    if (ret != 0)
+    {
+        _tprintf(_T("Winsock initialization failed: %d\n"), ret );
```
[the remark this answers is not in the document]
zefklop: Corrected, with a style which is closer to the one of surrounding code.

sdk/lib/drivers/wdf/shared/core/fxdevice.cpp
zefklop: Not done previously. I'll let this decision to @Extravert-ir

base/applications/network/net/cmdUser.c
```diff
@@ -188,7 +188,11 @@ GetTimeInSeconds(VOID)
     GetSystemTimeAsFileTime(&FileTime);
     Time.u.LowPart = FileTime.dwLowDateTime;
     Time.u.HighPart = FileTime.dwHighDateTime;
-    RtlTimeToSecondsSince1970(&Time, &dwSeconds);
+    if (!RtlTimeToSecondsSince1970(&Time, &dwSeconds))
```
HBelusca, line 191: Regarding `RtlTimeToSecondsSince1970`, and the related `RtlTimeToSecondsSince1980` and `RtlTimeFieldsToTime`: in the latest Windows WDK (see its ntifs.h file), these functions are not marked with `_Must_inspect_result_` (or `_Check_result_`): they are just marked with `_Success_(return != 0)` . So I don't know why in ReactOS they've got that extra `_Must_inspect_result_` thing.

HBelusca: Actually, in ReactOS, `RtlTimeToSecondsSince1980` doesn't have the `_Must_inspect_result_` annotation, exactly as in the MS WDK. So whoever added it for `RtlTimeToSecondsSince1970`, besides breaking compatibility with MS WDK, also added an inconsistency.

sdk/include/psdk/gcc_sal.h
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+#pragma once
+
+#ifndef __GNUC__
+#error "Not for your compiler!"
+#endif
+
+#ifndef __has_attribute
+#warn "GCC without __has_attribute, no SAL niceties for you"
```
wjk, line 15: I think you want `#warning` here.

wjk (changes requested):

base/applications/network/net/cmdUser.c
zefklop: Should we remove those annotations, though ? Maybe we can add some define ?
`__REACTOS_NO_EXTRA_SAL_ANNOTATIONS__` maybe. We will have trouble with fastfat_new otherwise, and I don't feel like adding extra code just for this.

sdk/cmake/gcc.cmake
```diff
@@ -102,6 +102,7 @@ add_compile_options(-Wno-char-subscripts -Wno-multichar -Wno-unused-value)
 add_compile_options(-Wno-unused-const-variable)
 add_compile_options(-Wno-unused-local-typedefs)
 add_compile_options(-Wno-deprecated)
+add_compile_options(-Wno-unused-result) # FIXME To be removed when CORE-17637 is resolved
```
tkreuzer, line 105: What about allowing this warning instead of disabling it?

zefklop: I enable it for rtl & ntoskrnl. There 350 more occurrences to fix, I'd propose a per subdirectory approach. I already started with base/applications

base/applications/fltmc/fltmc.cpp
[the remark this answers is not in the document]
SergeGautherie: > NULL is for pointers, period.

PS: ROS code style does read to use `*_NULL`...

sdk/lib/rtl/priv.c
```diff
@@ -387,9 +387,12 @@ RtlReleasePrivilege(IN PVOID ReturnedState)
...
     {
         /* Otherwise, restore old state */
-        ZwAdjustPrivilegesToken(State->Token, FALSE,
-                                State->OldPrivileges, 0, NULL, NULL);
-
+        Status = ZwAdjustPrivilegesToken(State->Token, FALSE,
+                                         State->OldPrivileges, 0, NULL, NULL);
+        if (!NT_SUCCESS(Status))
+        {
+            RtlRaiseStatus(Status);
```
HBelusca, line 394: Not sure the `RtlReleasePrivilege` function is expected to raise an exception.

sdk/lib/drivers/wdf/shared/targets/general/fxiotargetremote.cpp
```diff
@@ -27,7 +27,6 @@ extern "C" {
 #include <initguid.h>
 #include "wdmguid.h"
 
-
```
HBelusca, line 30: Please revert (it's a 3rd party file)

sdk/include/wdf/kmdf/1.17/wdf.h
```diff
@@ -40,7 +40,7 @@ Revision History:
 #endif
 
 // reactos start
-#include <psdk/sal.h>
+#include <sal.h>
```
HBelusca, line 43: Can be reverted.

ntoskrnl/mm/ARM3/section.c
```diff
@@ -1927,11 +1927,13 @@ MmGetFileNameForAddress(IN PVOID Address,
     if (NT_SUCCESS(Status))
     {
         /* Init modulename */
-        RtlCreateUnicodeString(ModuleName, ModuleNameInformation->Name.Buffer);
+        if(!RtlCreateUnicodeString(ModuleName, ModuleNameInformation->Name.Buffer))
```
HBelusca, line 1930:
```suggestion
        if (!RtlCreateUnicodeString(ModuleName, ModuleNameInformation->Name.Buffer))
```

ntoskrnl/config/i386/cmhardwr.c
```diff
@@ -763,22 +776,25 @@ CmpInitializeMachineDependentConfiguration(IN PLOADER_PARAMETER_BLOCK LoaderBloc
...
-                                   Data.Length + sizeof(UNICODE_NULL));
-
-            /* ROS: Save a copy for bugzilla reporting */
-            RtlCreateUnicodeString(&KeRosVideoBiosDate, Data.Buffer);
-
-            /* Free the string */
-            RtlFreeUnicodeString(&Data);
+            if (NT_SUCCESS(RtlAnsiStringToUnicodeString(&Data, &TempString, TRUE)))
+            {
+
```
HBelusca, line 781:
```suggestion
```

ntoskrnl/config/i386/cmhardwr.c
```diff
@@ -732,7 +744,8 @@ CmpInitializeMachineDependentConfiguration(IN PLOADER_PARAMETER_BLOCK LoaderBloc
                                        TotalLength);
 
                 /* ROS: Save a copy for bugzilla reporting */
-                RtlCreateUnicodeString(&KeRosBiosVersion, (PWCH)BiosVersion);
+                if(!RtlCreateUnicodeString(&KeRosBiosVersion, (PWCH)BiosVersion))
```
HBelusca, line 747:
```suggestion
                if (!RtlCreateUnicodeString(&KeRosBiosVersion, (PWCH)BiosVersion))
```

ntoskrnl/config/i386/cmhardwr.c
```diff
@@ -672,7 +679,10 @@ CmpInitializeMachineDependentConfiguration(IN PLOADER_PARAMETER_BLOCK LoaderBloc
                                            Data.Length + sizeof(UNICODE_NULL));
 
                     /* ROS: Save a copy for bugzilla reporting */
-                    RtlCreateUnicodeString(&KeRosBiosDate, Data.Buffer);
+                    if (!RtlCreateUnicodeString(&KeRosBiosDate, Data.Buffer))
+                    {
+                        KeRosBiosDate.Length = 0;
+                    }
```
HBelusca, line 685:
```suggestion
                        KeRosBiosDate.Length = 0;
```

ntoskrnl/config/i386/cmhardwr.c
```diff
@@ -763,22 +776,25 @@ CmpInitializeMachineDependentConfiguration(IN PLOADER_PARAMETER_BLOCK LoaderBloc
...
-                                   REG_SZ,
-                                   Data.Buffer,
-                                   Data.Length + sizeof(UNICODE_NULL));
-
-            /* ROS: Save a copy for bugzilla reporting */
-            RtlCreateUnicodeString(&KeRosVideoBiosDate, Data.Buffer);
-
-            /* Free the string */
-            RtlFreeUnicodeString(&Data);
+            if (NT_SUCCESS(RtlAnsiStringToUnicodeString(&Data, &TempString, TRUE)))
```
HBelusca, line 779: Why not using
```suggestion
            Status = RtlAnsiStringToUnicodeString(&Data, &TempString, TRUE);
            if (NT_SUCCESS(Status))
```
(and same question for other changes above and below)

base/applications/network/net/cmdUser.c
HBelusca: Not sure it's very important to have `_Must_inspect_result_` for `RtlTimeToSecondsSince1970` just for ReactOS...

sdk/lib/rtl/priv.c
zefklop: It raises just above. 
This PR is about killing the warnings while keeping the existing as it is. If this raises concerns about the existing, then let's merge this PR and open new ones for specific issues, otherwise we will never move forward.

base/applications/network/net/cmdUser.c
zefklop: Yeah, let's ditch it.

ntoskrnl/config/i386/cmhardwr.c
zefklop: Dunno. That whole function is a mess anyway.

sdk/lib/drivers/wdf/shared/targets/general/fxiotargetremote.cpp
zefklop: Oops :-)

HBelusca (approved):

PR #3753, opened 2021-06-18
Amd64/more memory

tkreuzer:
Allows to use more than 4 GB (it was actually only 1 GB) on x64 builds.

### DIFF
--- a/boot/freeldr/freeldr/include/mm.h
+++ b/boot/freeldr/freeldr/include/mm.h
@@ -51,7 +51,12 @@ typedef struct _FREELDR_MEMORY_DESCRIPTOR
 #define MM_PAGE_SIZE    4096
 #define MM_PAGE_MASK    0xFFF
 #define MM_PAGE_SHIFT    12
-#define MM_MAX_PAGE        0xFFFFF
+#if defined(_X86PAE_)
+#define MM_MAX_PAGE        0x3FFFFFF /* 26 bits for the PFN */
+#else
+#define MM_MAX_PAGE        0xFFFFF /* 20 bits for the PFN */
+#endif
+#define MM_MAX_PAGE_LOADER 0xFFFFF /* 4 GB flat address range */
 
 #define MM_SIZE_TO_PAGES(a)  \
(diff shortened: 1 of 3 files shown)


binarymaster (approved):
LGTM

ntoskrnl/include/internal/mm.h
```diff
@@ -1282,7 +1290,11 @@ MmFindRegion(
 #define IS_DIRTY_SSE(E)          ((E) & 2)
 #define WRITE_SSE(E)             ((E) | 4)
 #define IS_WRITE_SSE(E)          ((E) & 4)
+#ifdef _WIN64
```
HBelusca, line 1293:
```suggestion
#ifdef _M_AMD64
```

HBelusca (approved):
see suggestion, but LGTM otherwise.

ntoskrnl/include/internal/mm.h
tkreuzer: Well the whole SSE stuff is horrible, but it's generally not architecture dependent, except for 32 vs 64 bits, so using _WIN64 is intentionally here. Should work with any 64 bit system.

boot/freeldr/freeldr/include/mm.h
```diff
@@ -51,7 +51,12 @@ typedef struct _FREELDR_MEMORY_DESCRIPTOR
 #define MM_PAGE_SIZE    4096
 #define MM_PAGE_MASK    0xFFF
 #define MM_PAGE_SHIFT    12
-#define MM_MAX_PAGE        0xFFFFF
+#if defined(_X86PAE_)
```
tkreuzer, line 54: Just for the record: I don't think that we will actually compile freeloader with that definition, and I don't think freeloader should bother with this, but PAE is not really a concern for me other than as a design principle, so it basically serves as a documentation, how stuff should work, and an actual correct implementation would probably need to either make sure Mm can deal with too high physical pages or freeloader needs to determine the kernel that is loaded and decide the margin based upon that.

HBelusca: Especially since it is known, that the same NTLDR on Windows can manage both the non-PAE and the PAE ntoskrnl.

PR #3756, opened 2021-06-19
[NTOS:PS] Fix several issues within info classes in AMD64 build

GeoB99:
This fixes several issues regarding process/thread information classes where they fail with a size mismatch status error in AMD64 build.

## TODO
- [x] Keep investigating for other potential issues until we're assured that things work fine as they should

### DIFF
--- a/ntoskrnl/include/internal/ps_i.h
+++ b/ntoskrnl/include/internal/ps_i.h
@@ -142,13 +142,16 @@ static const INFORMATION_CLASS_INFO PsProcessInfoClass[] =
         ICIF_QUERY | ICIF_SET | ICIF_SET_SIZE_VARIABLE
     ),
 
-    /* ProcessUserModeIOPL */
-    IQS_SAME
+    /* ProcessUserModeIOPL is only implemented in x86 */
+#if defined (_X86_)
+    IQS_NO_TYPE_LENGTH
     (
-        UCHAR,
         ULONG,
         ICIF_SET
     ),
+#else
+    IQS_NONE,
+#endif
 
     /* ProcessEnableAlignmentFaultFixup */
     IQS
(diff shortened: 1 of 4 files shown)


ntoskrnl/ps/query.c
```diff
@@ -1172,7 +1172,7 @@ NtSetInformationProcess(IN HANDLE ProcessHandle,
         case ProcessWx86Information:
 
             /* Check buffer length */
-            if (ProcessInformationLength != sizeof(HANDLE))
+            if (ProcessInformationLength != sizeof(ULONG))
```
HBelusca, line 1175: Are you sure you change the correct case? This is `ProcessWx86Information`, not `ProcessWow64Information`

HBelusca: EDIT: indeed you're correct here, for `ProcessWx86Information` this is indeed a ULONG.

ntoskrnl/include/internal/ps_i.h
```diff
@@ -142,13 +142,17 @@ static const INFORMATION_CLASS_INFO PsProcessInfoClass[] =
         ICIF_QUERY | ICIF_SET | ICIF_SET_SIZE_VARIABLE
     ),
 
-    /* ProcessUserModeIOPL */
+    /* ProcessUserModeIOPL is only implemented in x86 */
+#if defined (_X86_)
     IQS_SAME
     (
-        UCHAR,
+        0,
```
tkreuzer, line 149: This field is for the type and expands to `sizeof(Type)`, i.e. `sizeof(0)` == `sizeof(int), which is probably not what you want. I suggest keeping it as it is.

GeoB99: The thing is, according to NT/2000 API reference documentation `ProcessUserModeIOPL` doesn't require a size length of any type hence it should be 0. Keeping UCHAR as it's is not correct either. Now that I think about it, are you OK with me having a new IQS macro specifically for these kind of particular classes like  ProcessUserModeIOPL, `IQS_NO_TYPE_LENGTH`, where the class has its own alginment but no type size length?

HBelusca: Having a flag like you propose makes sense.

PR #63, opened 2017-10-18
CORE-13048. [I8042PRT] keyboard.c: Restore 1 "Irql =" (which was lost in r30000).

SergeGautherie:
[CORE-13048](https://jira.reactos.org/browse/CORE-13048)
- keyboard.c: Restore 1 "Irql =" (which was lost in r30000).

PR #3758, opened 2021-06-20
[WIN32K] CreateDIBPalette SEH Simplification

Doug-Lyons:
## Simplify the SEH testing previously commited.

_Enter an SEH2_TRY one time and do all testing within it instead of entering the SEH_TRY multiple times._

JIRA issue: [CORE-17626](https://jira.reactos.org/browse/CORE-17626)

## Proposed from Mark Jansen's improvement suggestions.

### DIFF
--- a/win32ss/gdi/ntgdi/dibobj.c
+++ b/win32ss/gdi/ntgdi/dibobj.c
@@ -45,8 +45,6 @@ CreateDIBPalette(
 {
     PPALETTE ppal;
     ULONG i, cBitsPixel, cColors;
-    RGBQUAD rgb;
-    NTSTATUS Status;
 
     if (pbmi->bmiHeader.biSize < sizeof(BITMAPINFOHEADER))
     {
@@ -132,32 +130,30 @@ CreateDIBPalette(
 
             // FIXME: do we need to handle PALETTEINDEX / PALETTERGB macro?
 
-            /* Loop all color indices in the DIB */
-            for (i = 0; i < cColors; i++)
+            /* Use SEH to verify we can READ all data successfully */
+            _SEH2_TRY
             {
-                /* User SEH to verify READ success */
-                Status = STATUS_SUCCESS;
-                _SEH2_TRY
-                {
-                    rgb = prgb[i];
-                }
-                _SEH2_EXCEPT(EXCEPTION_EXECUTE_HANDLER)
-                {
-                    Status = _SEH2_GetExceptionCode();
-                    /* On Read Failure, put zero in Palette */
-                    PALETTE_vSetRGBColorForIndex(ppal, i, 0);
-                }
-                _SEH2_END
-
-                if(NT_SUCCESS(Status))
+                /* Loop all color indices in the DIB */
+                for (i = 0; i < cColors; i++)
                 {
                     /* Get the color value and translate it to a COLORREF */
+                    RGBQUAD rgb = prgb[i];
                     COLORREF crColor = RGB(rgb.rgbRed, rgb.rgbGreen, rgb.rgbBlue);
 
                     /* Set the RGB value in the palette */
                     PALETTE_vSetRGBColorForIndex(ppal, i, crColor);
                 }
             }
+            _SEH2_EXCEPT(EXCEPTION_EXECUTE_HANDLER)
+            {
+                /* Set all palette entries to zero */
+                for (i = 0; i < cColors; i++)
+                {
+                        PALETTE_vSetRGBColorForIndex(ppal, i, 0);
+                }
+            }
+            _SEH2_END
+
         }
         else
         {
--- a/win32ss/gdi/ntgdi/dibobj.c
+++ b/win32ss/gdi/ntgdi/dibobj.c
@@ -152,7 +152,7 @@ CreateDIBPalette(
                     PALETTE_vSetRGBColorForIndex(ppal, i, 0);
                 }
             }
-            _SEH2_END
+            _SEH2_END;
 
         }
         else
--- a/win32ss/gdi/ntgdi/dibobj.c
+++ b/win32ss/gdi/ntgdi/dibobj.c
@@ -149,7 +149,7 @@ CreateDIBPalette(
                 /* Set all palette entries to zero */
                 for (i = 0; i < cColors; i++)
                 {
-                        PALETTE_vSetRGBColorForIndex(ppal, i, 0);
+                    PALETTE_vSetRGBColorForIndex(ppal, i, 0);
                 }
             }
             _SEH2_END
--- a/win32ss/gdi/ntgdi/dibobj.c
+++ b/win32ss/gdi/ntgdi/dibobj.c
@@ -140,7 +140,7 @@ CreateDIBPalette(
             {
               /* Do Nothing */
             }
-            _SEH2_END
+            _SEH2_END;
 
             for (i = 0; i < cColors; ++i)
             {
--- a/win32ss/gdi/ntgdi/dibobj.c
+++ b/win32ss/gdi/ntgdi/dibobj.c
@@ -134,10 +134,7 @@ CreateDIBPalette(
             /* Use SEH to verify we can READ prgb[] succesfully */
             _SEH2_TRY
             {
-                for(i = 0; i < cColors; ++i)
-                {
-                    colors[i] = prgb[i];
-                }
+                RtlCopyMemory(colors, prgb, cColors * sizeof(colors[0]));
             }
             _SEH2_EXCEPT(EXCEPTION_EXECUTE_HANDLER)
             {
--- a/win32ss/gdi/ntgdi/dibobj.c
+++ b/win32ss/gdi/ntgdi/dibobj.c
@@ -127,33 +127,33 @@ CreateDIBPalette(
         {
             /* The colors are an array of RGBQUAD values */
             RGBQUAD *prgb = (RGBQUAD*)((PCHAR)pbmi + pbmi->bmiHeader.biSize);
+            RGBQUAD colors[256] = {0};
 
             // FIXME: do we need to handle PALETTEINDEX / PALETTERGB macro?
 
-            /* Use SEH to verify we can READ all data successfully */
+            /* Use SEH to verify we can READ prgb[] succesfully */
             _SEH2_TRY
             {
-                /* Loop all color indices in the DIB */
-                for (i = 0; i < cColors; i++)
+                for(i = 0; i < cColors; ++i)
                 {
-                    /* Get the color value and translate it to a COLORREF */
-                    RGBQUAD rgb = prgb[i];
-                    COLORREF crColor = RGB(rgb.rgbRed, rgb.rgbGreen, rgb.rgbBlue);
-
-                    /* Set the RGB value in the palette */
-                    PALETTE_vSetRGBColorForIndex(ppal, i, crColor);
+                    colors[i] = prgb[i];
                 }
             }
             _SEH2_EXCEPT(EXCEPTION_EXECUTE_HANDLER)
             {
-                /* Set all palette entries to zero */
-                for (i = 0; i < cColors; i++)
-                {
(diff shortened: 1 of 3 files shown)


win32ss/gdi/ntgdi/dibobj.c
```diff
@@ -132,32 +130,30 @@ CreateDIBPalette(
...
                     /* Set the RGB value in the palette */
                     PALETTE_vSetRGBColorForIndex(ppal, i, crColor);
                 }
             }
+            _SEH2_EXCEPT(EXCEPTION_EXECUTE_HANDLER)
+            {
+                /* Set all palette entries to zero */
+                for (i = 0; i < cColors; i++)
+                {
+                        PALETTE_vSetRGBColorForIndex(ppal, i, 0);
```
HBelusca, line 152: indentation is off.

win32ss/gdi/ntgdi/dibobj.c
```diff
@@ -132,32 +130,30 @@ CreateDIBPalette(
...
+                }
+            }
+            _SEH2_END
```
HBelusca, line 155:
```suggestion
            _SEH2_END;
```

Doug-Lyons: In this file (dibobj.c) there are four instances of _SEH2_END that do not have the following semi-colon and two that do have the following semi-colon. Can you point me to some authoritative documentation on its syntax?

win32ss/gdi/ntgdi/dibobj.c
Doug-Lyons: Fixed. Thanks.

win32ss/gdi/ntgdi/dibobj.c
HBelusca: It's the syntax we use everywhere else.

Doug-Lyons: Even here https://reactos.org/wiki/PSEH there are examples that are shown both with and without the trailing semicolon.

HBelusca: only one over all the other ones.

Doug-Lyons: OK. You comment is authoritative enough for me. Fixed. Thanks.

learn-more: There are only very few that use this without `;`:
https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=_SEH2_END%24&sr=1

JoachimHenze (approved):

zefklop:
I'd prefer having an array of 256 values on the stack, copy the input array in a `_SEH2_TRY` block with the proper check, and then set the palette entries outside of the SEH block from this copied array. This would also avoid the strange reset loop in the `_SEH2_EXCEPT` block

Doug-Lyons:
> I'd prefer having an array of 256 values on the stack, copy the input array in a `_SEH2_TRY` block with the proper check, and then set the palette entries outside of the SEH block from this copied array. This would also avoid the strange reset loop in the `_SEH2_EXCEPT` block

Sorry, I do not know how to put these onto the stack. I have no concept of when the C-Program will use the stack versus not using it. Can you give me some pseudo-code to help me understand what you are asking? Or point me to another existing example to follow. Thanks.

zefklop:
> > I'd prefer having an array of 256 values on the stack, copy the input array in a `_SEH2_TRY` block with the proper check, and then set the palette entries outside of the SEH block from this copied array. This would also avoid the strange reset loop in the `_SEH2_EXCEPT` block
> 
> Sorry, I do not know how to put these onto the stack. I have no concept of when the C-Program will use the stack versus not using it. Can you give me some pseudo-code to help me understand what you are asking? Or point me to another existing example to follow. Thanks.

The stack is the space that is allocated for your functions to run. So each time you write something like
```
void some_func(void)
{
    BYTE var[256];
    do_something(var);
}
```
The array `var` is allocated on the stack. See this: https://stackoverflow.com/questions/79923/what-and-where-are-the-stack-and-heap

Doug-Lyons:
> > > I'd prefer having an array of 256 values on the stack, copy the input array in a `_SEH2_TRY` block with the proper check, and then set the palette entries outside of the SEH block from this copied array. This would also avoid the strange reset loop in the `_SEH2_EXCEPT` block
> > 
> > 
> > Sorry, I do not know how to put these onto the stack. I have no concept of when the C-Program will use the stack versus not using it. Can you give me some pseudo-code to help me understand what you are asking? Or point me to another existing example to follow. Thanks.
> 
> The stack is the space that is allocated for your functions to run. So each time you write something like
> 
> ```
> void some_func(void)
> {
>     BYTE var[256];
>     do_something(var);
> }
> ```
> 
> The array `var` is allocated on the stack. See this: https://stackoverflow.com/questions/79923/what-and-where-are-the-stack-and-heap

OK. Thanks. Let me work on this a bit.

Doug-Lyons:
> > > > I'd prefer having an array of 256 values on the stack, copy the input array in a `_SEH2_TRY` block with the proper check, and then set the palette entries outside of the SEH block from this copied array. This would also avoid the strange reset loop in the `_SEH2_EXCEPT` block
> > > 
> > > 
> > > Sorry, I do not know how to put these onto the stack. I have no concept of when the C-Program will use the stack versus not using it. Can you give me some pseudo-code to help me understand what you are asking? Or point me to another existing example to follow. Thanks.
> > 
> > 
> > The stack is the space that is allocated for your functions to run. So each time you write something like
> > ```
> > void some_func(void)
> > {
> >     BYTE var[256];
> >     do_something(var);
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The array `var` is allocated on the stack. See this: https://stackoverflow.com/questions/79923/what-and-where-are-the-stack-and-heap
> 
> OK. Thanks. Let me work on this a bit.

I believe that I have made the updates as requested. Thanks.

win32ss/gdi/ntgdi/dibobj.c
```diff
@@ -129,34 +127,32 @@ CreateDIBPalette(
...
+            _SEH2_TRY
             {
-                /* User SEH to verify READ success */
-                Status = STATUS_SUCCESS;
-                _SEH2_TRY
-                {
-                    rgb = prgb[i];
-                }
-                _SEH2_EXCEPT(EXCEPTION_EXECUTE_HANDLER)
+                for(i = 0; i < cColors; ++i)
```
zefklop, line 137: is `cColors` guaranteed to be less than 256 ? 
Also, `RtlCopyMemory(colors, prgb, cColors * sizeof(colors[0]))` would look better to me.

win32ss/gdi/ntgdi/dibobj.c
```diff
@@ -129,34 +127,32 @@ CreateDIBPalette(
...
-                    Status = _SEH2_GetExceptionCode();
-                    /* On Read Failure, put zero in Palette */
-                    PALETTE_vSetRGBColorForIndex(ppal, i, 0);
+                    colors[i] = prgb[i];
                 }
-                _SEH2_END
+            }
+            _SEH2_EXCEPT(EXCEPTION_EXECUTE_HANDLER)
+            {
+              /* Do Nothing */
```
zefklop, line 144: Error handling ? Return FALSE or something ? Otherwise you're just zeroing the palette in the code below.

win32ss/gdi/ntgdi/dibobj.c
Doug-Lyons: cColors will be less than or equal to 256. Looking at the code above in Line #60 we have this:
```
    if (cBitsPixel <= 8)
```
Which is a prerequisite of reaching our code at all. So the maximum value of cBitsPixel is 8.
Next in Line #63, we have the following:
```
        cColors = 1 << cBitsPixel;
```
So taking the maximum value of '8' for cBitsPixel and shifting 1 left this '8' times gives us 256.
This guarantees that the cColors value will never be greater than 256.

I will change from using the for look to your RtlCopyMemory since it looks better to you.

HBelusca: > Now in Line#78 we have this:
>            cColors = min(cColors, pbmi->bmiHeader.biClrUsed);

This by itself doesn't guarantee anything, as biClrUsed can be anything (it's a DWORD value).

win32ss/gdi/ntgdi/dibobj.c
Doug-Lyons: Nothing is needed or wanted to be returned here. Our new statement #130 is our initializer and is as follows:
```
            RGBQUAD colors[256] = {0};
```
This effectively sets all or our 256 output RGBQUAD colors[] values to { 0, 0, 0, 0} which is what we want as our default values.
If our _SEH2_TRY succeeds in changing the value of some colors[] to something else, then we want these new values.
Now when we do the RtlCopyMemory we get the updated colors[] value copied into the 'ppal' palette.
Otherwise we want to keep the defaults and copy these into the 'ppal' and continue with our logic as a graceful fallback.

Don't you agree? Thanks, Doug Lyons

win32ss/gdi/ntgdi/dibobj.c
Doug-Lyons: @HBelusca, a 'min' function must take the lowest value, so the maximum of any one value must be the maximum for the 'min' function. So if we have min(256, AnythingElse), we know that this cannot return anything greater than 256. Correct?

HBelusca: Ah yes, indeed, if either one of the "parameters" of the min macro is already ensured to be not greater than 256, then it should return a value <= 256.

Doug-Lyons: Thanks for noting your agreement with me on this. :)

win32ss/gdi/ntgdi/dibobj.c
zefklop: I would prefer to return a NULL palette in this case, otherwise we are just creating a black DIB. A quick check shows that callers handle this, except here https://git.reactos.org/?p=reactos.git;a=blob;f=win32ss/gdi/ntgdi/dibobj.c;hb=c8ce0cc4340224bf93f4f34050deaf504d598bc3#l1460 which would need to be reworked anyway.

Doug-Lyons: If I understand what you are asking, you want something like this:
```
            /* Use SEH to verify we can READ prgb[] succesfully */
            _SEH2_TRY
            {
                RtlCopyMemory(colors, prgb, cColors * sizeof(colors[0]));
            }
            _SEH2_EXCEPT(EXCEPTION_EXECUTE_HANDLER)
            {
                Status = _SEH2_GetExceptionCode();
            }
            _SEH2_END

            if(!NT_SUCCESS(Status))
            {
                SetLastNtError(Status);
                return NULL;
            }
```

and when I tested this code, I received the following results:

GDIProg_Bad_Exitpng

This is definitely NOT what Windows Server 2003 SP2 does. Do I understand your request correctly? Thanks.

zefklop: Interesting. You can leave it as it is, then.

Doug-Lyons: Great! Thanks.

zefklop (approved):

JoachimHenze:
That _SEH2_END is now without the trailing semicolon again (most likely due to frequent changes).

Doug-Lyons:
> That _SEH2_END is now without the trailing semicolon again (most likely due to frequent changes).

Should be fixed now. Thanks.

JoachimHenze:
I just committed the final content of this PR by hand from my local workspace by 0.4.15-dev-2775-gc596fd3 https://git.reactos.org/?p=reactos.git;a=commit;h=c596fd3ef6933c9eec95a47c807d01d3776bbcc0 . The PR can be closed therefore. It is merged.

PR #3761, opened 2021-06-22
[CMD] Improve performance

zefklop:
## Purpose

Sledgehammer deletion of weird outdated debug stuff, because I am tired of seeing cmd:batch taking way too much time when I run the test suite.

## Proposed changes

Do not use custom alloc/free functions. If you need this, use DPH.
Do not free resources at process exit.
 - You're doing it brick by brick while the OS does that with a bulldozer better than you do.

### DIFF
--- a/base/shell/cmd/batch.c
+++ b/base/shell/cmd/batch.c
@@ -75,7 +75,6 @@ BOOL bEcho = TRUE;  /* The echo flag */
 /* Buffer for reading Batch file lines */
 TCHAR textline[BATCH_BUFFSIZE];
 
-
 /*
  * Returns a pointer to the n'th parameter of the current batch file.
  * If no such parameter exists returns pointer to empty string.
@@ -527,47 +526,31 @@ VOID AddBatchRedirection(REDIRECTION **RedirList)
  */
 BOOL BatchGetString(LPTSTR lpBuffer, INT nBufferLength)
 {
-    LPSTR lpString;
     INT len = 0;
-#ifdef _UNICODE
-    lpString = cmd_alloc(nBufferLength);
-    if (!lpString)
-    {
-        WARN("Cannot allocate memory for lpString\n");
-        error_out_of_memory();
-        return FALSE;
-    }
-#else
-    lpString = lpBuffer;
-#endif
+
     /* read all chars from memory until a '\n' is encountered */
     if (bc->mem)
     {
-        for (; (bc->mempos < bc->memsize  &&  len < (nBufferLength-1)); len++)
-        { 
-            lpString[len] = bc->mem[bc->mempos++];
-            if (lpString[len] == '\n' )
+        for (; ((bc->mempos + len) < bc->memsize  &&  len < (nBufferLength-1)); len++)
+        {
+#ifndef _UNICODE
+            lpBuffer[len] = bc->mem[bc->mempos + len];
+#endif
+            if (bc->mem[bc->mempos + len] == '\n')
             {
                 len++;
                 break;
             }
         }
+#ifdef _UNICODE
+        nBufferLength = MultiByteToWideChar(OutputCodePage, 0, &bc->mem[bc->mempos], len, lpBuffer, nBufferLength);
+        lpBuffer[nBufferLength] = L'\0';
+        lpBuffer[len] = '\0';
+#endif
+        bc->mempos += len;
     }
 
-    if (!len)
-    {
-#ifdef _UNICODE
-        cmd_free(lpString);
-#endif
-        return FALSE;
-    }
-
-    lpString[len++] = '\0';
-#ifdef _UNICODE
-    MultiByteToWideChar(OutputCodePage, 0, lpString, -1, lpBuffer, len);
-    cmd_free(lpString);
-#endif
-    return TRUE;
+    return len != 0;
 }
 
 /*
--- a/base/shell/cmd/cmd.c
+++ b/base/shell/cmd/cmd.c
@@ -2333,6 +2333,17 @@ static VOID Cleanup(VOID)
         ParseCommandLine(_T("\\cmdexit.bat"));
     }
 
+    /* Remove ctrl break handler */
+    RemoveBreakHandler();
+
+    /* Restore the default console mode */
+    SetConsoleMode(ConStreamGetOSHandle(StdIn),
+                   ENABLE_PROCESSED_INPUT | ENABLE_LINE_INPUT | ENABLE_ECHO_INPUT);
+    SetConsoleMode(ConStreamGetOSHandle(StdOut),
+                   ENABLE_PROCESSED_OUTPUT | ENABLE_WRAP_AT_EOL_OUTPUT);
+
+
+#ifdef _DEBUG_MEM
 #ifdef FEATURE_DIRECTORY_STACK
     /* Destroy directory stack */
     DestroyDirectoryStack();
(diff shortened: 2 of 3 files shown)


base/shell/cmd/cmd.c
```diff
@@ -2333,28 +2333,11 @@ static VOID Cleanup(VOID)
...
-    /* Remove ctrl break handler */
-    RemoveBreakHandler();
-
     /* Restore the default console mode */
     SetConsoleMode(ConStreamGetOSHandle(StdIn),
                    ENABLE_PROCESSED_INPUT | ENABLE_LINE_INPUT | ENABLE_ECHO_INPUT);
     SetConsoleMode(ConStreamGetOSHandle(StdOut),
                    ENABLE_PROCESSED_OUTPUT | ENABLE_WRAP_AT_EOL_OUTPUT);
-
-    DeleteCriticalSection(&ChildProcessRunningLock);
```
learn-more, line 2357: Could this interfere with application verifier?
There are some stops for critical sections not being released, but I am not sure if that also  is for the main thread.

zefklop: As far as I know, application verifier checks that you properly call 'DeleteCriticalSection' before freeing its underlying memory. In this case, this would be when Mm frees all pages from the mapped executable, when application verifier is way out of play.

HBelusca (dismissed):
Just for you know, the wrapped cmddbg things allow to see that e.g. there can be memory leaks sometimes when certain execution paths during batch file execution are taken. So I would suggest, that you keep them there, but just disable the DEBUG_MEM flag compilation by default. So that the cmd_alloc and co. would just compile out to the usual malloc etc.

base/shell/cmd/cmd.c
```diff
@@ -2333,28 +2333,11 @@ static VOID Cleanup(VOID)
...
-#ifdef FEATURE_HISTORY
-    CleanHistory();
-#endif
-
-    /* Free GetEnvVar's buffer */
-    GetEnvVar(NULL);
-
-    /* Remove ctrl break handler */
-    RemoveBreakHandler();
-
```
HBelusca, line 2350: Why?

zefklop: Because this ends up passing a lot of time in the heap implementation instead of just not caring like it can. The kernel frees the virtual memory when a process dies, and this is all that is needed to be done.

learn-more:
> Just for you know, the wrapped cmddbg things allow to see that e.g. there can be memory leaks sometimes when certain execution paths during batch file execution are taken. So I would suggest, that you keep them there, but just disable the DEBUG_MEM flag compilation by default. So that the cmd_alloc and co. would just compile out to the usual malloc etc.

When you are building with msvc, just use `_CRTDBG_MAP_ALLOC` with `_CrtDumpMemoryLeaks()` ?

I tend to agree with:
> You're doing it brick by brick while the OS does that with a bulldozer better than you do.

Where you can replace OS with crt at your convenience.

HBelusca:
> because I am tired of seeing cmd:batch taking way too much time when I run the test suite.

Out of curiosity, do you have some comparison of timing numbers "before" vs. "after"?

HBelusca:
> When you are building with msvc, just use `_CRTDBG_MAP_ALLOC` with `_CrtDumpMemoryLeaks()` ?

It would be interesting to have this in ReactOS (when I build cmd I use the project created from our configure step via cmake, and this will use our CRT, not MS' one).

zefklop:
> > because I am tired of seeing cmd:batch taking way too much time when I run the test suite.
> 
> Out of curiosity, do you have some comparison of timing numbers "before" vs. "after"?

image

base/shell/cmd/batch.c
```diff
@@ -72,10 +72,9 @@ BOOL bBcEcho = TRUE;
 
 BOOL bEcho = TRUE;  /* The echo flag */
 
-/* Buffer for reading Batch file lines */
+/* Buffers for reading Batch file lines */
```
HBelusca, line 75: 🤔 It looks like one single buffer there.

base/shell/cmd/batch.c
```diff
@@ -527,47 +526,32 @@ VOID AddBatchRedirection(REDIRECTION **RedirList)
...
-        for (; (bc->mempos < bc->memsize  &&  len < (nBufferLength-1)); len++)
-        { 
-            lpString[len] = bc->mem[bc->mempos++];
-            if (lpString[len] == '\n' )
+        for (; ((bc->mempos + len) < bc->memsize  &&  len < (nBufferLength-1)); len++)
+        {
+#ifndef _UNICODE
+            lpBuffer[len] = bc->mem[bc->mempos + len];
+#endif
+            if (bc->mem[bc->mempos + len] == '\n' )
```
HBelusca, line 539:
```suggestion
            if (bc->mem[bc->mempos + len] == '\n')
```

HBelusca (commented):
This batch line commit is nice :)

HBelusca:
> > > because I am tired of seeing cmd:batch taking way too much time when I run the test suite.
> > 
> > Out of curiosity, do you have some comparison of timing numbers "before" vs. "after"?

and before? like ~ 1 minute or so?

base/shell/cmd/parser.c
```diff
@@ -471,7 +471,7 @@ AllocCommand(
         SIZE_T CmdHeadLen = _tcslen(CmdHead) + 1;
         SIZE_T CmdTailLen = _tcslen(CmdTail) + 1;
 
-        Cmd = cmd_alloc(FIELD_OFFSET(PARSED_COMMAND,
+        Cmd = malloc(FIELD_OFFSET(PARSED_COMMAND,
                                      Command.First[CmdHeadLen + CmdTailLen]));
```
HBelusca, line 475:
```suggestion
                                  Command.First[CmdHeadLen + CmdTailLen]));
```

zefklop:
> > > > because I am tired of seeing cmd:batch taking way too much time when I run the test suite.
> > > 
> > > 
> > > Out of curiosity, do you have some comparison of timing numbers "before" vs. "after"?
> 
> and before? like ~ 1 minute or so?
Yes, on my machine. But testbot says it better:

https://reactos.org/testman/compare.php?ids=77843,77850

image

base/shell/cmd/parser.c
zefklop: reverted.

base/shell/cmd/cmd.c
HBelusca: OK since you now just disable the custom heap allocator by just removing the define, while keeping its code present, i.e. one can compile it back by adding the flag back; please restore these cleaning commands but surround them inside `#ifdef _DEBUG_MEM`, so that when the heap allocator is enabled, it can correctly count the cleanup there (and not telling "you've memory leaks").

zefklop: Done

base/shell/cmd/cmd.c
```diff
@@ -2333,6 +2333,7 @@ static VOID Cleanup(VOID)
         ParseCommandLine(_T("\\cmdexit.bat"));
     }
 
+#ifdef _DEBUG_MEM
```
HBelusca, line 2336: As usual, github's interface cannot allow adding comments/suggestions in code parts that are outside of the +-3 lines around the current diff'ed lines, so...

My suggestion here: since the console mode state is something that exists across different processes living in said console, the two SetConsoleMode calls below are necessary.
So you need to end the first `_DEBUG_MEM` block with an `#endif _DEBUG_MEM` just after the `RemoveBreakHandler();`, and then restart an `#ifdef _DEBUG_MEM` just before the `DeleteCriticalSection()` call.

HBelusca (approved):
LGTM.

PR #3765, opened 2021-06-24
[NTOS:KE] Rewrite KeZeroPages in assembly & add SSE version

zefklop:
We spend a lot of time in this function, let's optimize it.
Noticed while running heavy memory tests like mshtml_winetest

### DIFF
--- a/ntoskrnl/ke/amd64/zeropage.S
+++ b/ntoskrnl/ke/amd64/zeropage.S
@@ -1,0 +1,45 @@
+/*
+ * PROJECT:     ReactOS Kernel
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Fast zeroing of pages
+ * COPYRIGHT:   Copyright 2021 Jérôme Gardou <jerome.gardou@reactos.org>
+ */
+
+#include <asm.inc>
+
+/* FUNCTIONS ****************************************************************/
+.code
+
+/* Benchmarking from Timo on some AMD machine:
+    rep movsq : 128
+    movaps 175
+    movnti 620
+    movntdq: 620
+    movntps: 620
+    MS KeZeroPages (movnti unrolled): 883
+    MS KeZeroSinglePage (mov): 346
+
+    whole discussion in https://github.com/reactos/reactos/pull/3765
+    We stick with rep stosq.
+*/
+
+/*
+ * VOID
+ * KeZeroPages(PVOID Ptr, ULONG Size);
+ */
+PUBLIC KeZeroPages
+FUNC KeZeroPages
+    push rdi
+    .PUSHREG rdi
+    .ENDPROLOG
+
+    mov rdi, rcx
+    mov ecx, edx
+    shr ecx, 3
+    xor rax, rax
+    rep stosq
+    pop rdi
+    ret
+ENDFUNC
+
+END
--- a/ntoskrnl/ke/i386/zeropage.S
+++ b/ntoskrnl/ke/i386/zeropage.S
@@ -1,0 +1,32 @@
+/*
+ * PROJECT:     ReactOS Kernel
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Fast zeroing of pages
+ * COPYRIGHT:   Copyright 2021 Jérôme Gardou <jerome.gardou@reactos.org>
+ */
+
+#include <asm.inc>
+
+/* FUNCTIONS ****************************************************************/
+.code
+
+/*
+ * VOID
+ * FASTCALL
+ * KeZeroPages(void* ptr, ULONG Size)
+ */
+PUBLIC @KeZeroPages@8
+FUNC @KeZeroPages@8
+    FPO 0, 0, 0, 0, 0, FRAME_FPO
+
+    push edi
+    mov edi, ecx
+    mov ecx, edx
+    shr ecx, 2
+    xor eax, eax
+    rep stosd
+    pop edi
+    ret
+ENDFUNC
+
+END
--- a/ntoskrnl/ke/amd64/stubs.c
+++ b/ntoskrnl/ke/amd64/stubs.c
@@ -88,16 +88,6 @@ KiDpcInterruptHandler(VOID)
     KeLowerIrql(OldIrql);
 }
 
-
-VOID
-FASTCALL
-KeZeroPages(IN PVOID Address,
-            IN ULONG Size)
-{
-    /* Not using XMMI in this routine */
-    RtlZeroMemory(Address, Size);
-}
-
 PVOID
 KiSwitchKernelStackHelper(
     LONG_PTR StackOffset,
--- a/ntoskrnl/ke/i386/cpu.c
+++ b/ntoskrnl/ke/i386/cpu.c
@@ -1134,15 +1134,6 @@ KeInvalidateAllCaches(VOID)
 }
 
 VOID
-FASTCALL
-KeZeroPages(IN PVOID Address,
-            IN ULONG Size)
-{
-    /* Not using XMMI in this routine */
-    RtlZeroMemory(Address, Size);
-}
-
-VOID
 NTAPI
 KiSaveProcessorState(IN PKTRAP_FRAME TrapFrame,
                      IN PKEXCEPTION_FRAME ExceptionFrame)
--- a/ntoskrnl/mm/ARM3/hypermap.c
+++ b/ntoskrnl/mm/ARM3/hypermap.c
@@ -150,6 +150,10 @@ MiMapPagesInZeroSpace(IN PMMPFN Pfn1,
     PointerPte += (Offset + 1);
     TempPte = ValidKernelPte;
 
+    /* Disable cache. Write through */
+    MI_PAGE_DISABLE_CACHE(&TempPte);
+    MI_PAGE_WRITE_THROUGH(&TempPte);
+
     /* Make sure the list isn't empty and loop it */
     ASSERT(Pfn1 != (PVOID)LIST_HEAD);
     while (Pfn1 != (PVOID)LIST_HEAD)
--- a/ntoskrnl/mm/ARM3/pagfault.c
+++ b/ntoskrnl/mm/ARM3/pagfault.c
@@ -674,6 +674,11 @@ MiResolveDemandZeroFault(IN PVOID Address,
             PageFrameNumber = MiRemoveAnyPage(Color);
             NeedZero = TRUE;
         }
+        else
+        {
+            /* Page guaranteed to be zero-filled */
+            NeedZero = FALSE;
+        }
     }
     else
     {
@@ -688,6 +693,8 @@ MiResolveDemandZeroFault(IN PVOID Address,
         {
             /* System wants a zero page, obtain one */
             PageFrameNumber = MiRemoveZeroPage(Color);
+            /* No need to zero-fill it */
+            NeedZero = FALSE;
         }
     }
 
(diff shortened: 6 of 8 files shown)


HBelusca:
LGTM. I wonder how the timings compare there.

HBelusca (approved):

ntoskrnl/ke/i386/kiinit.c
```diff
@@ -159,7 +159,25 @@ KiInitMachineDependent(VOID)
...
-            /* FIXME: Implement and enable XMM Page Zeroing for Mm */
+            /* Patch the KeZeroPages function to enable fast zeroing */
+            PUCHAR Instr = (PUCHAR)KeZeroPages;
+            switch(*Instr)
+            {
+                case 0xE9:
+                    /* 32-bit jmp */
+                    Instr[4] = 0x90; //NOP
+                    Instr[3] = 0x90; //NOP
+                    Instr[2] = 0x90; //NOP
```
HBelusca, line 170: I think just one mention of "nop" is sufficient 😅 
```suggestion
                    Instr[4] = 0x90; // NOP
                    Instr[3] = 0x90;
                    Instr[2] = 0x90;
```

ntoskrnl/ke/i386/kiinit.c
```diff
@@ -159,7 +159,25 @@ KiInitMachineDependent(VOID)
             CpuCount = KeNumberProcessors;
             KeIpiGenericCall(Ki386EnableXMMIExceptions, (ULONG_PTR)&CpuCount);
 
-            /* FIXME: Implement and enable XMM Page Zeroing for Mm */
+            /* Patch the KeZeroPages function to enable fast zeroing */
+            PUCHAR Instr = (PUCHAR)KeZeroPages;
+            switch(*Instr)
```
HBelusca, line 164:
```suggestion
            switch (*Instr)
```

HBelusca (approved):

Extravert-ir:
Add SSE version on amd64 as you're writing it anyway

ntoskrnl/ke/i386/kiinit.c
```diff
@@ -159,7 +159,25 @@ KiInitMachineDependent(VOID)
             CpuCount = KeNumberProcessors;
             KeIpiGenericCall(Ki386EnableXMMIExceptions, (ULONG_PTR)&CpuCount);
 
-            /* FIXME: Implement and enable XMM Page Zeroing for Mm */
+            /* Patch the KeZeroPages function to enable fast zeroing */
+            PUCHAR Instr = (PUCHAR)KeZeroPages;
```
Extravert-ir, line 163: Isn't it possible to patch the symbol itself?

ntoskrnl/ke/i386/zeropage.S
```diff
@@ -0,0 +1,45 @@
...
+
+PUBLIC @KeZeroPages@8
+@KeZeroPages@8:
+    /* Kernel will overwrite this with nop if SSE is available */
+    jmp KeZeroPagesNoSSE
+    /* Save XMM0 & zero it */
+    mov eax, esp
+    and eax, HEX(FFFFFFF0)
+    sub eax, 16
+    movaps xmmword ptr [eax], xmm0
```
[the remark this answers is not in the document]
zefklop: Indeed.

ntoskrnl/ke/i386/zeropage.S
```diff
@@ -0,0 +1,45 @@
...
+    rep stosd
+    pop edi
+    ret
+
+PUBLIC @KeZeroPages@8
+@KeZeroPages@8:
+    /* Kernel will overwrite this with nop if SSE is available */
+    jmp KeZeroPagesNoSSE
+    /* Save XMM0 & zero it */
+    mov eax, esp
```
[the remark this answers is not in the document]
zefklop: I first thought there would be no interrupts there, but you're right, in the zero paging thread this can be the case.

ntoskrnl/ke/i386/kiinit.c
zefklop: > Isn't it possible to patch the symbol itself?

This function is called from inside the kernel. Patching the symbol table (is that waht you mean ?) wouldn't help

zefklop: > I'm guessing a function pointer wouldn't incur any noticeable perf cost given we always zero an entire page.

Yes, but then you add complexity to the code for little benefit. I guess it's a matter of taste :-)

ntoskrnl/ke/i386/zeropage.S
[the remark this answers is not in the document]
HBelusca: This observation possibly needs to be added as comment in code.

zefklop:
> Add SSE version on amd64 as you're writing it anyway

Sure. Done.

ntoskrnl/ke/i386/kiinit.c
ThFabba: You think a function pointer is more complex than runtime patching assembly? Interesting take! I guess it is a matter of taste then :)

ntoskrnl/ke/i386/zeropage.S
```diff
@@ -0,0 +1,49 @@
...
+ * COPYRIGHT:   Copyright 2021 Jérôme Gardou <jerome.gardou@reactos.org>
+ */
+
+#include <asm.inc>
+
+/* FUNCTIONS ****************************************************************/
+.code
+
+KeZeroPagesNoSSE:
+    push edi
```
ThFabba, line 14: Maybe add FPO information in case we ever crash in here?

ntoskrnl/ke/i386/zeropage.S
```diff
@@ -0,0 +1,49 @@
...
+    rep stosd
+    pop edi
+    ret
+
+PUBLIC @KeZeroPages@8
+@KeZeroPages@8:
+    /* Kernel will overwrite this with nop if SSE is available */
+    jmp KeZeroPagesNoSSE
+    /* Save XMM0 & zero it */
+    sub esp, 31
```
ThFabba, line 28: I know x86 doesn't have a stack alignment requirement, but this seems like it would throw off pretty much any other code looking at this stack pointer (e.g. the aforementioned interrupt). Do you think it's safe?

ntoskrnl/ke/amd64/zeropage.S
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+#include <asm.inc>
+
+/* FUNCTIONS ****************************************************************/
+.code
+
+PUBLIC KeZeroPages
+KeZeroPages:
+    pxor xmm0, xmm0
```
ThFabba, line 15: Might want a `.ENDPROLOG` at least?

JoachimHenze:
This PR sounds like it could be a potential fix for https://jira.reactos.org/browse/CORE-11856 .
If so, then it makes sense to mention the ticket in the description.

ThFabba (approved):
> LGTM. I wonder how the timings compare there.

```
KeZeroPages (using RtlZeroMemory):
(ntoskrnl\mm\ARM3\zeropage.c:182) Zeroed 497499 pages (1943 MB) in 1386237807 ns (815450618 for zero) @ 2786 ns/page (total)
(ntoskrnl\mm\ARM3\zeropage.c:182) Zeroed 497499 pages (1943 MB) in 1361019525 ns (799070887 for zero) @ 2735 ns/page (total)

KiXMMIZeroPagesNoSave movntps with sfence, no unroll:
[1 page]
(ntoskrnl\mm\ARM3\zeropage.c:182) Zeroed 497502 pages (1943 MB) in 1285018547 ns (745964152 for zero) @ 2582 ns/page (total)
(ntoskrnl\mm\ARM3\zeropage.c:182) Zeroed 497502 pages (1943 MB) in 1281508604 ns (746905333 for zero) @ 2575 ns/page (total)
```

I did a bunch more measurements but these are the most important. It seems to be about 10-20% faster in plain execution time, not accounting for the avoided cache pollution.
My conclusion was that you get much more speed-up by zeroing more than one page at a time, compared to using movntps -- but it's certainly still worth it.

ThFabba:
> This PR sounds like it could be a potential fix for https://jira.reactos.org/browse/CORE-11856 .
> If so, then it makes sense to mention the ticket in the description.

It should improve things, but note that the expected speed-up is no greater than 20%, so don't get your hopes up too much. ;)

ntoskrnl/ke/amd64/zeropage.S
HBelusca: your branch was nice, I wonder why it didn't make into PR state 🤔

ntoskrnl/ke/i386/zeropage.S
zefklop: Indeed. I don't even know why I just dont sub 32 there...

ntoskrnl/ke/amd64/zeropage.S
zefklop: doh, forgot that.

ntoskrnl/ke/i386/kiinit.c
tkreuzer: Please use a function pointer. There no additional complexity. You could also declare the function as dllimport and then have a pointer `PVOID __imp_KeZeroPages;`, which is the same thing.

ntoskrnl/ke/amd64/zeropage.S
```diff
@@ -0,0 +1,29 @@
...
+    /* Calculate buffer end */
+    add rdx, rcx
+    /* Loop */
+do_zero:
+    movntps xmmword ptr [rcx], xmm0
```
tkreuzer, line 20: I checked, whether unrolling could improve performance, but turns out, it doesn't seem to make much of a difference. At least not on my AMD machine. It stayed at around 630 cycles per page. RtlZeroMemory was 350 cycles, RtlSecureZeroMemory (using a simple `rep stosq`) was 250!

ThFabba: Yeah, my tests (on Intel) never showed any gain from unrolling either.
Btw, modern processors heavily optimize `rep stosb`, so it's typically worth trying that as well.

tkreuzer: Since we want to zero pages, i.e. we have both aligned pointer and size and a large size, we can just use `rep stosd` and `rep stosq`, which can be written in C with intrinsics and is most likely the fastest on any system. Maybe AVX 512 can do it faster, but SSE seemingly can't.

ThFabba: Even if `movntps` is "only" up to par with `stos`, that probably means it's going to make other code on the system noticeably faster by avoiding unnecessary cache evictions, right?

tkreuzer: Did some better benchmarking:
`` rep movsq`` : 128
`` movaps``  175
`` movnti`` 535
`` movntdq``: 620
`` movntps``: 620

So `` movntps``  and ``movnti``  are much slower than `` rep movsq`` . Does that help with cache? Different comments here https://stackoverflow.com/questions/10274355/cycles-cost-for-l1-cache-hit-vs-register-on-x86, suggest the difference between L1 and L2 is 6-8 cycles. So 6 * 64 = 384, 8 * 64 = 512. Versus 620 - 128 = 492 cycles slower zeroing. But chances are you are evicting cache lines, you don't need anymore. I guess you would need to write a more complex benchmark.

ThFabba: My measurements on Intel typically had `movntps` at the top. But zeroing multiple pages at once provided a much more noticeable gain than any optimization of the zeroing function.

I wonder if we could simply make `MiMapPagesInZeroSpace` disable caching on those pages entirely, to avoid the cache evictions, and then just stick with `stos` for simplicity (and apparently on some processors, performance).

zefklop: Let's stick with stosd/stosq

Extravert-ir (approved):
I'd say it makes sense to put part of the benchmarking discussions in the code as a comment, so it won't get lost (e.g. why we've chosen `rep stosd` here)

PR #3768, opened 2021-06-26
[NTOS:CACHE] _MmSetPageEntrySectionSegment(): Optimize if() conditions

SergeGautherie:
## Purpose

Addendum to b3e9c89.
JIRA issue: [CORE-17544](https://jira.reactos.org/browse/CORE-17544)

## Proposed changes

More like next `if()`.


ThFabba:
This is not an optimization, nor does it help readability.

SergeGautherie:
> This is not an optimization, nor does it help readability.

I think adding the `else`, at the very least, would be more explicit. Anyway.

PR #3773, opened 2021-06-29
[SCSIPORT] Use a work item to send SCSI sense request

zefklop:
This doesn't work from a DPC, since IoAllocateAsynchronousFsdRequest requires IRQL <= APC


zefklop:
IoAllocateAsynchronousFsdRequest requires IRQL <= APC_LEVEL when the buffer is not from non-paged pool, and it is in this case. So this is fine.

PR #3775, opened 2021-06-29
[FREELDR] Allow booting from NTFS volume.

vgalnt:
Minor changes to Freeloader that will allow booting from NTFS volume.

### DIFF
--- a/boot/freeldr/freeldr/include/fs/ntfs.h
+++ b/boot/freeldr/freeldr/include/fs/ntfs.h
@@ -65,6 +65,8 @@
 #define NTFS_FILE_NAME_DOS            2
 #define NTFS_FILE_NAME_WIN32_AND_DOS        3
 
+#define NTFS_MFT_MASK 0x0000FFFFFFFFFFFFULL
+
 #include <pshpack1.h>
 typedef struct
 {
@@ -115,6 +117,8 @@ typedef struct
     ULONG        BytesAllocated;
     ULONGLONG        BaseMFTRecord;
     USHORT        NextAttributeInstance;
+    USHORT        Padding;                     // Align to 4 UCHAR boundary (NTFS 3.1+ (Windows XP and above))
+    ULONG        MFTRecordNumber;             // Number of this MFT Record (NTFS 3.1+ (Windows XP and above))
 } NTFS_MFT_RECORD, *PNTFS_MFT_RECORD;
 
 typedef struct
@@ -185,7 +189,7 @@ typedef struct
     USHORT        Reserved;
     UCHAR        FileNameLength;
     UCHAR        FileNameType;
-    WCHAR        FileName[0];
+    WCHAR        FileName[1];
 } NTFS_FILE_NAME_ATTR, *PNTFS_FILE_NAME_ATTR;
 
 typedef struct
@@ -197,7 +201,7 @@ typedef struct
     ULONGLONG    StartingVCN;
     ULONGLONG    BaseFileRef;
     USHORT        AttrId;
-    PWCHAR        Name;
+    WCHAR        Name[1];
 } NTFS_ATTR_LIST_ATTR, *PNTFS_ATTR_LIST_ATTR;
 
 typedef struct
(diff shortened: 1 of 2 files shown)


HBelusca:
Please see also PR #3416.

boot/freeldr/freeldr/include/fs/ntfs.h
```diff
@@ -65,6 +65,8 @@
 #define NTFS_FILE_NAME_DOS            2
 #define NTFS_FILE_NAME_WIN32_AND_DOS        3
 
+#define NTFS_MFT_MASK 0x0000FFFFFFFFFFFFULL
+
```
binarymaster, line 69: Checked the code in both PRs, only this line crosses more or less, others are ok :)

HBelusca: The others (where the `NTFS_MFT_MASK` mask is individually applied to `IndexedFile` only when needed) is closer to what is done in our ntfs driver, than what the other PR I mentioned does.

vgalnt: Btw
https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=NTFS_MFT_MASK
This is duplicated in "drivers/filesystems/ntfs/ntfs.h"
Your suggestions?

HBelusca: It's fine to have a copy of this value in freeldr, since we don't share FS driver code.

boot/freeldr/freeldr/include/fs/ntfs.h
```diff
@@ -115,6 +117,8 @@ typedef struct
     ULONG        BytesAllocated;
     ULONGLONG        BaseMFTRecord;
     USHORT        NextAttributeInstance;
+    USHORT        Padding;                     // Align to 4 UCHAR boundary (NTFS 3.1+ (Windows XP and above))
```
learn-more, line 120: If this is required, it makes sense to validate it with C_ASSERT (So it cannot be broken accidentally)

boot/freeldr/freeldr/lib/fs/ntfs.c
```diff
@@ -680,7 +680,7 @@ static BOOLEAN NtfsFindMftRecord(PNTFS_VOLUME_INFO Volume, ULONGLONG MFTIndex, P
                     if (NtfsCompareFileName(FileName, IndexEntry))
                     {
                         TRACE("File found\n");
-                        *OutMFTIndex = IndexEntry->Data.Directory.IndexedFile;
+                        *OutMFTIndex = (IndexEntry->Data.Directory.IndexedFile & NTFS_MFT_MASK);
                         FrLdrTempFree(BitmapData, TAG_NTFS_BITMAP);
```
SergeGautherie, line 684: Remove superfluous parentheses?

boot/freeldr/freeldr/include/fs/ntfs.h
```diff
@@ -185,7 +189,7 @@ typedef struct
     USHORT        Reserved;
     UCHAR        FileNameLength;
     UCHAR        FileNameType;
-    WCHAR        FileName[0];
```
Extravert-ir, line 188: I don't like this, [1] must only be used on structures for "compatibility". This confuses all static analysis tools. Better fix code the relies on it

boot/freeldr/freeldr/include/fs/ntfs.h
Extravert-ir: As far as I can see, we can safely remove packed attribute from all structures here (except NTFS_BOOTSECTOR), and there will be no need in padding fields

boot/freeldr/freeldr/include/fs/ntfs.h
HBelusca: These static analysis tools should then be more clever / be fixed 🤣

Extravert-ir: Array with variable size is defined by `[]` or `[0]` in C, `[1]` means array with one element and nothing else

HBelusca: In **portable** C89 the empty arrays were not possible.

Extravert-ir: And we use **portable** C99 here

PR #3781, opened 2021-07-01
[HALX86:APIC] apic.c: Use new defines some more

SergeGautherie:
Follow-up to b7a149f.


DarkFire01 (approved):
I was asked to look at it, looks fine to me.

PR #3785, opened 2021-07-02
[ROSTESTS] Disable some tests on x64 for now

tkreuzer:
Try to make Test WHS x64 happy
See CORE-17664


binarymaster (approved):
Tester looks happy and so me. :smile: 

image

modules/rostests/apitests/localspl/tests.c
```diff
@@ -210,6 +210,14 @@ _RunRemoteTest(const char* szTestName)
 
 START_TEST(fpEnumPrinters)
 {
+#if defined(_M_AMD64)
+    if (!winetest_interactive)
+    {
+        skip("ROSTESTS-366: Skipping localspl_apitest:fpEnumPrinters because it hangs on Windows Server 2003 x64-Testbot. Set winetest_interactive to run it anyway.\n");
```
ColinFinck, line 216: All localspl API tests hang on the Win2003 x64 Testbot the same way. It's just that you won't notice now as we're running tests under a restricted user account.

modules/rostests/kmtests/ntos_ex/ExFastMutex.c
```diff
@@ -310,6 +310,14 @@ TestFastMutexConcurrent(
 
 START_TEST(ExFastMutex)
 {
+#if defined(_M_AMD64)
+    if (TRUE)
+    {
+        skip(FALSE, "ROSTESTS-367: Skipping kmtest:ExFastMutex because it hangs on Windows Server 2003 x64-Testbot. Set winetest_interactive to run it anyway.\n");
```
ColinFinck, line 316: Misleading comment, `winetest_interactive` won't change anything for this kmtest

modules/rostests/apitests/localspl/tests.c
tkreuzer: Ok, will skip all of them.

PR #3787, opened 2021-07-03
[CMAKE][SDK] Support NDEBUG_GLOBAL

SergeGautherie:
in addition to existing NDEBUG which is used locally.
Using 2 separate defines avoids redefinitions.

CMake can define NDEBUG by default for some compilers, like Clang.
It feels like having NDEBUG and NDEBUG_LOCAL would be better than
NDEBUG_GLOBAL and NDEBUG, but the latter is the current minimal change.
It also means ReactOS cannot rely on default global NDEBUG (anymore).

As hinted by @tkreuzer.
JIRA issue: [CORE-7885](https://jira.reactos.org/browse/CORE-7885)


Extravert-ir:
I don't see a problem in having those warnings for release builds. They don't hurt anyone
Moreover, we should shift from that NDEBUG thing away to using DbgPrintEx

PR #3788, opened 2021-07-03
[IMM32] Rewrite IMCC handling functions

katahiromz:
## Purpose

## Proposed changes

- Rewrite `ImmCreateIMCC`, `ImmDestroyIMCC`, `ImmReSizeIMCC`, `ImmGetIMCCLockCount`, `ImmUnlockIMCC`, `ImmLockIMCC`, `ImmGetIMCCSize` functions.

### DIFF
--- a/dll/win32/imm32/imm.c
+++ b/dll/win32/imm32/imm.c
@@ -2959,7 +2959,13 @@ DWORD WINAPI ImmGetIMCLockCount(HIMC hIMC)
 */
 HIMCC  WINAPI ImmCreateIMCC(DWORD size)
 {
+#ifdef __REACTOS__
+    if (size < 4)
+        size = 4;
+    return LocalAlloc(LHND, size);
+#else
     return GlobalAlloc(GMEM_ZEROINIT | GMEM_MOVEABLE, size);
+#endif
 }
 
 /***********************************************************************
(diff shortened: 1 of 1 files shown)


dll/win32/imm32/imm.c
```diff
@@ -2959,55 +2959,95 @@ DWORD WINAPI ImmGetIMCLockCount(HIMC hIMC)
 */
 HIMCC  WINAPI ImmCreateIMCC(DWORD size)
 {
+#ifdef __REACTOS__
+    if (size < 4)
+        size = 4;
+    return LocalAlloc(LHND, size);
```
learn-more, line 2965: What is the advantage of `LocalAlloc` over `GlobalAlloc` ?

katahiromz: I don't know well.

learn-more: https://docs.microsoft.com/en-us/windows/win32/memory/global-and-local-functions
There isn't all that much difference, so it might make sense to check if you actually need to change this,
and if so, I would assume it are mainly differences in options / flags.

HBelusca: On 9x systems, `GlobalAlloc()` makes the allocated block accessible system-wide, while `LocalAlloc()` makes it only available for the given process. On NT, both *Alloc make the block available only for the given process. They are both wrappers for `HeapAlloc()`.
Windows APIs that originally had the concept of allocating something to be accessible system-wide (e.g. the clipboard functions) will typically tell you they have allocated a memory buffer with `GlobalAlloc()` and that you may need to free it with `GlobalFree()`. Some other APIs that have the concept of allocating the buffer to be accessible only by the calling process, e.g. `FormatMessage()`, will tell you they have allocated the buffer with `LocalAlloc()` and you need to release it with `LocalFree()`.

katahiromz: This affects speed performance a little.

katahiromz:
Check time. I will merge 2 days later.

PR #3792, opened 2021-07-03
[WORDPAD] Fix icons for toolbar and formatbar

Doug-Lyons:
## Fix Icons for Wordpad's Toolbar and Formatbar to work in all bit depths

_Restored old Toolbar and create new Formatbar to correct problems._

JIRA issue: [CORE-5823](https://jira.reactos.org/browse/CORE-5823)


JoachimHenze (approved):
I have tested the binary wordpad.exe outcome of DougLyons proposal for all 32bpp/24bpp/16bpp/8bpp for all 3:
ros ok
2ksp2 ok
xpsp3 ok

This is better than the set of files that was proposed by Jared 10years ago, and that was still integrated into mayn of our releases (iirc ~0.3.17 until 0.4.7).
If the halfbaken version was good enough for integration, then this one is for sure as well.

Getting this synced to Wine will be tough, because they will most likely ask for someone to fix their SVG source as well. And nobody is really willing to do that.
I really want to merge this for us right now, I don't want to wait another 10years here.
Dougs solution works really perfectly!

JoachimHenze:
comparison shots in https://jira.reactos.org/browse/CORE-5823?focusedCommentId=128740&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-128740
I will merge this in t minus 2 hours.

SergeGautherie:
> wordpad_ros.diff

Please, rename it: `_ros.diff` was/is used for actual diffs.

Doug-Lyons:
> > wordpad_ros.diff
> 
> Please, rename it: `_ros.diff` was/is used for actual diffs.

@SergeGautherie, I appreciate your comments. I believe that if you look around more you will see that there are other places were this format is used and it is not an "actual diff" such as ole32_ros.diff. So my preference is to leave it for now until we have some type of official standard to follow. Thanks.

HBelusca:
We don't care about syncing back binary resources with Wine anyway.

SergeGautherie:
> it is not an "actual diff" such as ole32_ros.diff.

?? https://git.reactos.org/?p=reactos.git;a=blob;f=dll/win32/ole32/ole32_ros.diff;hb=HEAD does look like a diff to me...

PR #3795, opened 2021-07-04
[HALX86][NTOS:KE] Add the fundamental implementation for halmacpi

DarkFire01:
## Purpose

Implement the first half of Halmacpi and the generic SMP logic. 
Implement the x86 version of KeStartAllProcessors and the generic logic for the x64 variant. 

## Proposed changes

- Implement x86 KeStartAllProcessors
- Add the ability for ReactOS to start other physical processors
- Bringup x86 cores to protected mode and halt
- halt x64 processors right at bootstub 
- Create the other needed logic to continue forward


hal/halx86/smp/i386/apentry.S
```diff
@@ -0,0 +1,73 @@
...
+
+    mov   eax, cr0
+    or    eax, HEX(00000001)
+    mov   cr0, eax
+
+    ljmp16 HEX(08), (HEX(4000) + _APEntryEnd - _APEntry)
+
+    hlt
+
+gdtptr:
```
tkreuzer, line 40: Why is this using it's own GDT? Why doesn't it reuse the one from the BSP?

hal/halx86/generic/up.c
```diff
@@ -0,0 +1,31 @@
+/*
+ * PROJECT:     ReactOS Kernel
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Core source file for UP alternative functions
+ * PROGRAMMERS:  Copyright 2021 Justin Miller <justinmiller100@gmail.com>
+ */
```
Extravert-ir, line 6: Here and in other files:
```suggestion
/*
 * PROJECT:     ReactOS Kernel
 * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
 * PURPOSE:     Core source file for UP alternative functions
 * COPYRIGHT:   Copyright 2021 Justin Miller <justinmiller100@gmail.com>
 */
```

hal/halx86/generic/up.c
```diff
@@ -0,0 +1,31 @@
...
+HalRequestIpi(KAFFINITY TargetProcessors)
+{
+    /* This should never be called in UP mode */
+    __debugbreak();
+}
+
+BOOLEAN
+NTAPI
+HalStartNextProcessor(IN PLOADER_PARAMETER_BLOCK LoaderBlock,
+                      IN PKPROCESSOR_STATE ProcessorState)
```
Extravert-ir, line 27: Here and in other declarations/definitions:
```suggestion
HalStartNextProcessor(
    _In_ PLOADER_PARAMETER_BLOCK LoaderBlock,
    _In_ PKPROCESSOR_STATE ProcessorState)
```

ntoskrnl/ke/amd64/mp.c
```diff
@@ -0,0 +1,95 @@
...
+    HighestPhysicalAddress.QuadPart = UINT64_MAX;
+    KDESCRIPTOR GdtDesc, IdtDesc;
+    __sgdt(&GdtDesc.Limit);
+    __sidt(&IdtDesc.Limit);
+    do
+    {
+        ProcessorCount++;
+        /* Attempt to allocate memory for its new setup */
+        SIZE_T APInfo = sizeof(struct APInfo) + 2; 
+        PVOID PAPInfo = MmAllocateContiguousMemory(APInfo, HighestPhysicalAddress);
```
tkreuzer, line 44: Why is this not allocated from non-paged pool?

ntoskrnl/ke/amd64/mp.c
```diff
@@ -0,0 +1,95 @@
...
+    BOOLEAN ApplicationProcessorsEnabled = FALSE;
+    HighestPhysicalAddress.QuadPart = UINT64_MAX;
+    KDESCRIPTOR GdtDesc, IdtDesc;
+    __sgdt(&GdtDesc.Limit);
+    __sidt(&IdtDesc.Limit);
+    do
+    {
+        ProcessorCount++;
+        /* Attempt to allocate memory for its new setup */
+        SIZE_T APInfo = sizeof(struct APInfo) + 2;
```
tkreuzer, line 43: Please call this APInfoSize or something.

ntoskrnl/ke/amd64/mp.c
```diff
@@ -0,0 +1,95 @@
...
+    do
+    {
+        ProcessorCount++;
+        /* Attempt to allocate memory for its new setup */
+        SIZE_T APInfo = sizeof(struct APInfo) + 2; 
+        PVOID PAPInfo = MmAllocateContiguousMemory(APInfo, HighestPhysicalAddress);
+
+        if (!PAPInfo)
+        {
+            ASSERT("KeStartAllProcessors: Memory Allocation has failed");
```
tkreuzer, line 48: This is not how an ASSERT works!

hal/halx86/smp/i386/apentry.S
```diff
@@ -0,0 +1,73 @@
...
+
+      /* 16-bit real mode DS */
+     .word HEX(FFFF)
+     .word HEX(0000)
+     .word HEX(9200)
+     .word HEX(0000)
+gdtend:
+    hlt
+_APEntryEnd:
+END
```
Extravert-ir, line 73: The file should end with a new line (here and in other files)

hal/halx86/smp/i386/apentry.S
```diff
@@ -0,0 +1,73 @@
...
+PUBLIC _APEntry
+PUBLIC _APEntryEnd
+
+.code16
+_APEntry:
+    cli
+
+    xor     ax, ax
+	mov		ds, ax
+	mov		ss, ax
```
Extravert-ir, line 12: Indentation

ntoskrnl/ke/amd64/mp.c
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        RtlCopyMemory((PVOID)((ULONG_PTR)PAPInfo + sizeof(KPCR) + sizeof(KTSS)), (PVOID)GdtDesc.Base, GdtDesc.Limit + 1);
```
tkreuzer, line 51: There is a structure definition for it, so use that.

hal/halx86/smp/i386/apentry.S
```diff
@@ -0,0 +1,73 @@
+#include <asm.inc>
```
Extravert-ir, line 1: ASM files also need to have a header with licence

Extravert-ir (changes requested):
OK, coding style stuff first

ntoskrnl/ke/amd64/mp.c
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        /* Prep a new loaderblock for AP */
+        KeLoaderBlock->KernelStack = (ULONG_PTR)KernelStack;
+        KeLoaderBlock->Prcb = (ULONG_PTR)pKPcr;
+        KeLoaderBlock->Thread = (ULONG_PTR)InitialThread;
+
+        ApplicationProcessorsEnabled = HalStartNextProcessor(KeLoaderBlock, &ProcessorState);
+
+        if (ApplicationProcessorsEnabled == FALSE){
```
tkreuzer, line 84: Opening brace on new line.

ntoskrnl/ke/amd64/mp.c
```diff
@@ -0,0 +1,95 @@
...
+        PKTHREAD InitialThread = (PKTHREAD)((ULONG_PTR) PAPInfo + sizeof(KPCR) + sizeof(KTSS));
+
+        /* Prep some stacks for the AP */
+        PVOID KernelStack;
+        PVOID DPCStack;
+
+        KernelStack = MmCreateKernelStack(FALSE, 0);
+        if (!KernelStack)
+        {
+            ASSERT("KeStartAllProcessors: Could not create Kernel Stack for an AP");
```
tkreuzer, line 68: This is not how ASSERT() works

hal/halx86/smp/i386/apentry.S
```diff
@@ -0,0 +1,73 @@
...
+#else
+    lgdt cs:[gdtptr - _APEntry]
+    lidt cs:[idtdesc - _APEntry]
+#endif
+
+    mov   eax, cr0
+    or    eax, HEX(00000001)
+    mov   cr0, eax
+
+    ljmp16 HEX(08), (HEX(4000) + _APEntryEnd - _APEntry)
```
Extravert-ir, line 29: What's `HEX(4000)`?

ntoskrnl/ke/amd64/mp.c
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        DPCStack = MmCreateKernelStack(FALSE, 0);
+        if (!DPCStack)
+        {
+            MmDeleteKernelStack(KernelStack, FALSE);
```
tkreuzer, line 74: Maybe goto exit and really clean up?

ntoskrnl/ke/i386/mp.c
```diff
@@ -0,0 +1,140 @@
...
+        ApplicationProcessorsEnabled = HalStartNextProcessor(KeLoaderBlock, &ProcessorState);
+
+        if (ApplicationProcessorsEnabled == FALSE){
+            ProcessorCount--;
+            /* Started means the AP itself is online, this doesn't mean it's seen by the kernel */
+            DPRINT1("HalStartNextProcessor: Sucessful AP startup count is %X\n", ProcessorCount);
+            /* We have finished starting processors, Time to cleanup! */
+            MmFreeContiguousMemory(PAPInfo);
+            MmDeleteKernelStack(KernelStack, FALSE);
+            MmDeleteKernelStack(DPCStack, FALSE);
```
Extravert-ir, line 136: This is better to keep outside the loop

ntoskrnl/ke/i386/mp.c
```diff
@@ -0,0 +1,140 @@
...
+        /* Prep a new loaderblock for AP */
+        KeLoaderBlock->KernelStack = (ULONG_PTR)KernelStack;
+        KeLoaderBlock->Prcb = (ULONG_PTR)pKPcr;
+        KeLoaderBlock->Thread = (ULONG_PTR)InitialThread;
+
+        /* Fully initalize AP's TSS */
+        Ki386InitializeTss(pKTss, pIdt, pGdt);
+
+        /* Finally call HalStartNextProcessor with the prepped values */
+        ApplicationProcessorsEnabled = HalStartNextProcessor(KeLoaderBlock, &ProcessorState);
```
Extravert-ir, line 127: The variable is not needed here, just put `HalStartNextProcessor` inside the while() condition

ntoskrnl/ke/i386/mp.c
```diff
@@ -0,0 +1,140 @@
...
+    KDESCRIPTOR GdtDesc, IdtDesc;
+    __sgdt(&GdtDesc.Limit);
+    __sidt(&IdtDesc.Limit);
+    do
+    {
+        ProcessorCount++;
+
+        /* Attempt to allocate memory for its new setup */
+        SIZE_T APInfoSize = sizeof(struct APInfo) + 2;
+        PVOID PAPInfo = MmAllocateContiguousMemory(APInfoSize, HighestPhysicalAddress);
```
Extravert-ir, line 45: Are you sure HighestPhysicalAddress is OK here? As far as I understand, this stuff will be used in AP boot code, shouldn't it be within the first 1MB?

ntoskrnl/ke/i386/mp.c
```diff
@@ -0,0 +1,140 @@
...
+
+        if (!PAPInfo)
+        {
+            ASSERT("KeStartAllProcessors: Memory Allocation has failed");
+        }
+
+        RtlCopyMemory((PVOID)((ULONG_PTR)PAPInfo + sizeof(KPCR) + sizeof(KTSS)), (PVOID)GdtDesc.Base, GdtDesc.Limit + 1);
```
Extravert-ir, line 52: Treat `PAPInfo` like a pointer to a structure, you won't need `RtlCopyMemory` that way

Extravert-ir (changes requested):

ntoskrnl/ke/i386/mp.c
DarkFire01: This isn't used in the AP boot code, this is far after from what I understand.

ntoskrnl/include/internal/amd64/ke.h
```diff
@@ -455,6 +455,13 @@ KiSetTrapContext(
     _In_ PCONTEXT Context,
     _In_ KPROCESSOR_MODE RequestorMode);
 
+VOID
+NTAPI
+KiInitializePcr(IN PKIPCR Pcr,
```
tkreuzer, line 460: Why is this needed? What's wrong with initializing it in KiSystemStartup?

ntoskrnl/ke/amd64/mp.c
```diff
@@ -0,0 +1,95 @@
...
+        if (!PAPInfo)
+        {
+            ASSERT("KeStartAllProcessors: Memory Allocation has failed");
+        }
+
+        RtlCopyMemory((PVOID)((ULONG_PTR)PAPInfo + sizeof(KPCR) + sizeof(KTSS)), (PVOID)GdtDesc.Base, GdtDesc.Limit + 1);
+        RtlCopyMemory((PVOID)((ULONG_PTR)PAPInfo + sizeof(KPCR) + sizeof(KTSS) + GdtDesc.Limit + 1), (PVOID)IdtDesc.Base, IdtDesc.Limit + 1);
+
+        /* Prep processorstate */
+        KPROCESSOR_STATE ProcessorState;
```
tkreuzer, line 55: Since we already have a PCR, you can use the processor state from Pcr->Prcb.ProcessorState;

ntoskrnl/ke/amd64/mp.c
```diff
@@ -0,0 +1,95 @@
...
+
+        /* Allocate PCR - InitalThread */
+        PKPCR pKPcr = (PKPCR)PAPInfo;
+        PKTHREAD InitialThread = (PKTHREAD)((ULONG_PTR) PAPInfo + sizeof(KPCR) + sizeof(KTSS));
```
tkreuzer, line 59: Use the struct, Luke.

ntoskrnl/ke/i386/mp.c
```diff
@@ -0,0 +1,140 @@
...
+PHYSICAL_ADDRESS HighestPhysicalAddress;
+ULONG ProcessorCount = 0;
+
+struct APInfo
+{
+    KPCR pcr;
+    KTSS tss;
+    ETHREAD thread;
+    KDESCRIPTOR GdtDesc;
+    KDESCRIPTOR IdtDesc;
```
tkreuzer, line 26: These are part of KPROCESSOR_STATE::SpecialRegisters

ntoskrnl/ke/i386/mp.c
```diff
@@ -0,0 +1,140 @@
...
+        }
+
+        DPCStack = MmCreateKernelStack(FALSE, 0);
+        if (!DPCStack)
+        {
+            MmDeleteKernelStack(KernelStack, FALSE);
+        }
+
+        /* Initalize a new PCR for the specific AP */
+        KiInitializePcr(ProcessorCount,
```
tkreuzer, line 110: Why does this need to be initialized? What's wrong with doing that in KiSystemStartup, like done for the BSP?

ntoskrnl/ke/i386/mp.c
```diff
@@ -0,0 +1,140 @@
...
+    KTSS tss;
+    ETHREAD thread;
+    KDESCRIPTOR GdtDesc;
+    KDESCRIPTOR IdtDesc;
+} APInfo;
+
+/* FUNCTIONS *****************************************************************/
+VOID
+NTAPI
+KeStartAllProcessors()
```
tkreuzer, line 32: A lot of this function is not architecture dependent. Please split into an architecture independent version, which calls architecture dependent helper function to initialize the processor state.

ntoskrnl/ke/amd64/mp.c
```diff
@@ -0,0 +1,95 @@
...
+    {
+        ProcessorCount++;
+        /* Attempt to allocate memory for its new setup */
+        SIZE_T APInfo = sizeof(struct APInfo) + 2; 
+        PVOID PAPInfo = MmAllocateContiguousMemory(APInfo, HighestPhysicalAddress);
+
+        if (!PAPInfo)
+        {
+            ASSERT("KeStartAllProcessors: Memory Allocation has failed");
+        }
```
GeoB99, line 49: You're not using the ASSERT macro consistently. If you want to break out in the debugger if the predicate in the assertion condition is not TRUE (the predicate being your `PAPInfo` obviously) then use `ERROR_FATAL()`.
https://git.reactos.org/?p=reactos.git;a=blob;f=sdk/include/reactos/debug.h;hb=fb8edf90a002cefd0d99adb4316f12e6d6840418#l238

tkreuzer (changes requested):

ntoskrnl/ke/i386/mp.c
DarkFire01: https://github.com/reactos/reactos/blob/master/ntoskrnl/ke/i386/kiinit.c#L755-L756

ntoskrnl/ke/i386/mp.c
DarkFire01: I got a few different ideas for this subject, I'm going to leave this comment open so we can all figure it out.
I have written both versions, where I have done this and where I haven't here is my 2 cents:
- A lot of this function IS architecture dependent, and even the size of for example the TSS is different for x64. 
- I don't have much of an opinion to splitting off the processor state functions but if it's different for every architecture and we are using different implementations of the function per architecture then Is there still a point to splitting it off to another function?
- I feel like separating the implementations is cleaner and easier to work with, it would take a lot of helper functions for selective tasks and/or a lot of preprocessor definitions to get it working as a single implementation. 
- and finally the main reason: This is what my mentor suggested me to do so I PRed this version

Im totally willing to change it if we decide it's a better approach as the file is already written though.

hal/halx86/smp/smp.c
```diff
@@ -0,0 +1,50 @@
...
+NTAPI
+HalStartNextProcessor(
+    IN PLOADER_PARAMETER_BLOCK LoaderBlock,
+    IN PKPROCESSOR_STATE ProcessorState)
+{
+    if (MaxAPCount > StartedProcessorCount)
+    {
+        /* We only a create a new pagetable once */
+        if (StartedProcessorCount == 1)
+            HalpInitializeAPPageTables(HalpLowStub);
```
tkreuzer, line 37: This shouldn't be done here, but in the achitecture dependent code in ntoskrnl: You allocate a page, map it, copy the lower half of the current page table root, unmap it, store the PFN in ProcessorState.SpecialRegister.Cr3.

hal/halx86/smp/i386/apentry.S
DarkFire01: It does eventually reuse the one from the BSP, but when you are initializing an application processor you need to create a temporary GDT that can be accessed in the real mode stub.

hal/halx86/smp/smp.c
DarkFire01: Interesting... Okay ill remove this stub in this PR, and we can touch on this subject again over PM? before i submit the next one. Id really like to try this idea

ntoskrnl/ke/amd64/mp.c
HBelusca: And what is that structure? I don't see any structure like:
```c
typedef struct _some_struct
{
    KPCR Pcrb;
    KTSS Tss;
#ifdef _M_AMD64
    // Here: DoubleFaultStack
    // Here: MachineCheckStack
    // Here: NMIStack
#endif
    ETHREAD InitialThread;
#ifdef _M_IX86
    KTSS DoubleFaultTSS;
    KTSS NmiTSS;
    // Here: ExceptionStack;
#endif
    // Here: The full GDT for the given processor.
    // Here: The full IDT for the given processor.
} my_struct;
```
which is the one necessary for holding the full per-CPU data information, as explained in https://www.geoffchappell.com/studies/windows/km/ntoskrnl/inc/ntos/kpcr.htm .

ntoskrnl/ke/i386/mp.c
HBelusca: See comment above.

ntoskrnl/ke/i386/mp.c
HBelusca: Windows does it like that too.

ntoskrnl/ke/i386/mp.c
```diff
@@ -0,0 +1,127 @@
...
+
+/* INCLUDES *****************************************************************/
+
+#include <ntoskrnl.h>
+#define NDEBUG
+#include <debug.h>
+
+/* GLOBALS *******************************************************************/
+
+#define PHYSICAL_ADDRESS  LARGE_INTEGER
```
HBelusca, line 16: This definition already exists elsewhere, and should be already available in the kernel.

ntoskrnl/ke/amd64/mp.c
HBelusca: It's the global per-processor state structure, see https://www.geoffchappell.com/studies/windows/km/ntoskrnl/inc/ntos/kpcr.htm ,
so "obviously" all that must be available at all times and therefore cannot be paged.

tkreuzer: That is why I asked "Why is this not allocated from **non-paged** pool?"

HBelusca: Isn't the memory allocated with `MmAllocateContiguousMemory` in non-paged pool too?

ntoskrnl/ke/amd64/mp.c
tkreuzer: 1. My comment has nothing to do with arcitecture specific fields.
2. There is no problem with having a architecture specific definition of that structure.
3. You wouldn't need any of that, because you can put all of that into the processor state.

HBelusca: The `PROCESSOR_STATE` structure contains "pointers" to the actual data that needs to be stored somewhere (the stacks, the ETHREAD, the KPCR , for that given processor), and this is that actual data that needs to be allocated and initialized.

ntoskrnl/ke/i386/mp.c
learn-more: `psdk/ntdef.h`

ntoskrnl/ke/i386/mp.c
tkreuzer: See my blog post from 6 years ago.

HBelusca: an URL to it could help :P

ntoskrnl/ke/i386/mp.c
tkreuzer: Ok so here are my arguments:
1. A lot of this function IS architecture independent
2. Splitting it would result in 1 relatively simple loop + 2 relatively simple flat functions, vs 2 functions with large loop bodies that work partly the same, but partly not and when there is a bug fixed in one, is it gonna be fixed in the other one?
3. No, it doesn't need a lot of helper functions, it only needs one main function and one small helper function per architecture.
4. You should probably address this last, when everything else is addressed.
4. Your mentor better shows up here and explains himself 😛

hal/halx86/apic/apicsmp.c
```diff
@@ -84,13 +85,10 @@ ApicRequestGlobalInterrupt(
 
 /* SMP SUPPORT FUNCTIONS ******************************************************/
 
-// Should be called by SMP version of HalRequestIpi
 VOID
-NTAPI
+NTAPI /* HalRequestIpi (CONFIG_SMP) */
```
zefklop, line 89: What does this comment mean ?

hal/halx86/generic/up.c
```diff
@@ -0,0 +1,32 @@
...
+#include <debug.h>
+
+/* FUNCTIONS *****************************************************************/
+
+VOID
+NTAPI
+HalRequestIpi(KAFFINITY TargetProcessors)
+{
+    /* This should never be called in UP mode */
+    __debugbreak();
```
zefklop, line 21: Do we really need this symbol ? If it's never called, it shouldn't even be there, right ?

hal/halx86/generic/up.c
```diff
@@ -0,0 +1,32 @@
...
+}
+
+BOOLEAN
+NTAPI
+HalStartNextProcessor(IN PLOADER_PARAMETER_BLOCK LoaderBlock,
```
zefklop: same here. Why do we need it ?

hal/halx86/generic/up.c
DarkFire01: This and HalStartNextProcessor gets built with every single one of our Hals, Instead of having them in a processor.c thats associated with every different Hal, I threw these functions in this specific file so we easily build either the Uniprocessor or Multiprocessor count part.

This is the same functionality as before I just moved the functions into a better spot so we can actually build the CONFIG_SMP counterparts.

hal/halx86/generic/up.c
DarkFire01: I don't really know if i should copy and past the message above so i will just link it instead for the take of organization, 
https://github.com/reactos/reactos/pull/3795#discussion_r663883849

hal/halx86/apic/apicsmp.c
DarkFire01: Only called by HalRequestIpi when using the CONFIG_SMP definition. This was just for the sake of organization, I can remove it's a bit of a weird comment. :) just lemme know if you'd like me to take it out

hal/halx86/generic/up.c
zefklop: I see they are in the exports of the hal
Unless we want to restore the preprocessed spec thingie (I personnally don't), I guess we have to live with that.

Wouldn't a bugcheck be what we want there ? We surely don't want a SMP kernel calling into a MP hal, do we ?

hal/halx86/apic/apicsmp.c
zefklop: Then "Only called by SMP version of HalRequestIpi" would work. Those telegraphic comments don't help to understand.

hal/halx86/smp/mpmemory.c
```diff
@@ -0,0 +1,40 @@
...
+
+/* GLOBALS *******************************************************************/
+
+/* AP spinup stub specific */
+#if 0
+extern PVOID APEntry;
+extern PVOID APEntryEnd;
+extern PVOID APSpinup;
+extern PVOID APSpinupEnd;
+#endif
```
zefklop: Please remove this if the variables are already declared in headers.

hal/halx86/smp/smp.c
```diff
@@ -10,23 +10,48 @@
...
+    if(MaxAPCount > StartedProcessorCount)
+    {
+        /* We only a create a new pagetable once */
+        if(StartedProcessorCount == 1)
+        {
+            //HalpInitializeAPPageTables(HalpLowStub);
+        }
+        
+        /* Start an AP */
+        HalpInitializeAPStub(HalpLowStub);
```
zefklop, line 36: Maybe call this only when `StartedProcessorCount == 1` ? Or on HAL entry point ?

hal/halx86/smp/amd64/apspinup.S
```diff
@@ -0,0 +1,10 @@
+#include <asm.inc>
+
+PUBLIC APSpinup
+PUBLIC APSpinupEnd
+
+.code64
+APSpinup:
+    hlt
+APSpinupEnd:
+END
```
zefklop: Missing new line

hal/halx86/smp/mpmemory.c
```diff
@@ -17,24 +17,20 @@
...
 extern PVOID APSpinup;
 extern PVOID APSpinupEnd;
-#endif
 
 /* FUNCTIONS *****************************************************************/
 
 VOID
 HalpInitializeAPStub(PVOID APStubLocation)
 {
-    #if 0
```
zefklop: Please add this code in the commit actually using it.

hal/halx86/smp/mpmemory.c
```diff
@@ -17,24 +17,20 @@
...
     PVOID APStubSecondPhaseLoc;
 
     /* Copy over the bootstub for specific AP */
     APStubSecondPhaseLoc = (PVOID)((ULONG_PTR)APStubLocation + ((ULONG_PTR)&APEntryEnd  - (ULONG_PTR)&APEntry));
     RtlCopyMemory(APStubLocation, &APEntry,  ((ULONG_PTR)&APEntryEnd - (ULONG_PTR)&APEntry));
     RtlCopyMemory(APStubSecondPhaseLoc, &APSpinup,  ((ULONG_PTR)&APSpinupEnd - (ULONG_PTR)&APSpinup));
```
zefklop: Can you explain why this is needed ?

ntoskrnl/ke/amd64/mp.c
```diff
@@ -0,0 +1,84 @@
...
+
+/* INCLUDES *****************************************************************/
+
+#include <ntoskrnl.h>
+#define NDEBUG
+#include <debug.h>
+
+/* GLOBALS *******************************************************************/
+
+#define PHYSICAL_ADDRESS  LARGE_INTEGER
```
zefklop, line 16: oh not this in amd64 please 😭

ntoskrnl/ke/amd64/mp.c
zefklop: Nope, it takes space from the "system space" where e.g. drivers are mapped

hal/halx86/smp/mpmemory.c
DarkFire01: We copy over the boot stub code to a location under a megabyte, as the processors start in real mode.

ntoskrnl/ke/amd64/mp.c
DarkFire01: Huh? Is there a better alternative?

hal/halx86/smp/i386/apentry.S
```diff
@@ -22,11 +29,12 @@ _APEntry:
     lidt cs:[idtdesc - _APEntry]
 #endif
 
-    mov   eax, cr0
-    or    eax, HEX(00000001)
-    mov   cr0, eax
+    mov eax, cr0
+    or  eax, HEX(00000001)
+    mov cr0, eax
```
zefklop, line 34: Small comment like "enable protected mode" maybe ? Also, there is CR0_PE constant.

hal/halx86/smp/smp.c
DarkFire01: At the stub expands into the next PR, we overwrite the stub per AP as we use different values per application processor.

ntoskrnl/ke/i386/mp.c
```diff
@@ -33,10 +41,10 @@ KeStartAllProcessors()
         ProcessorCount++;
 
         /* Attempt to allocate memory for its new setup */
-        SIZE_T APInfo = sizeof(KPCR) + sizeof(KTSS) + sizeof(ETHREAD) + GdtDesc.Limit + 1 + IdtDesc.Limit + 1; 
-        PVOID PAPInfo = MmAllocateContiguousMemory(APInfo, HighestPhysicalAddress);
+        SIZE_T APInfoSize = sizeof(struct APInfo) + 2;
```
zefklop: where does this `+2` come from ?

DarkFire01: GDT limit + 1 
IDT limit +1,  
You want a slight offset to the size of the table.

ntoskrnl/ke/i386/mp.c
```diff
@@ -118,7 +126,10 @@ KeStartAllProcessors()
         /* Finally call HalStartNextProcessor with the prepped values */
         ApplicationProcessorsEnabled = HalStartNextProcessor(KeLoaderBlock, &ProcessorState);
 
-        if(ApplicationProcessorsEnabled == FALSE){
+        if (ApplicationProcessorsEnabled == FALSE){
+            ProcessorCount--;
+            /* Started means the AP itself is online, this doesn't mean it's seen by the kernel */
+            DPRINT1("HalStartNextProcessor: Sucessful AP startup count is %X\n", ProcessorCount);
```
zefklop: I'm confused about what this code is about. Is it failure or success ?

hal/halx86/smp/smp.c
zefklop: 👍

ntoskrnl/ke/i386/mp.c
DarkFire01: It's an intended failure. We continue the loop until processors are no longer being successfully started

hal/halx86/smp/mpmemory.c
DarkFire01: I'm entirely why github is saying that these are disabled. But in the actual branch they're enabled. 
Never the less these are never declared in headers.

hal/halx86/smp/i386/apentry.S
DarkFire01: Are you asking me to redefine the CR0_PE constant? 
I'm not entirely sure how id pull that into the bootstub in a way an application processor can access it.

hal/halx86/smp/mpmemory.c
zefklop: oh, I did a per-commit review. Well, I guess it's fine, but I find it strange to write disabled code only to enable it in the next commit.

ntoskrnl/ke/amd64/mp.c
zefklop: Using ExAllocatePoolWithTag, as Timo pointed out ;-)

hal/halx86/smp/mpmemory.c
DarkFire01: I'm actually extremely terrible at git, so i probably did that on accident sorry about that!

hal/halx86/smp/i386/apentry.S
zefklop: no, I mean instead of `HEX(00000001)`, use `CR0_PE` from ks386.h

ntoskrnl/ke/i386/mp.c
zefklop: Then I think I don't understand the comment above.

ntoskrnl/ke/i386/mp.c
zefklop: You're writing into the ETHREAD part of PAPInfo.

ntoskrnl/ke/i386/mp.c
DarkFire01: Which comment? also this If condition isn't in the file anymore

ntoskrnl/ke/i386/mp.c
Extravert-ir: I'd like to postpone the split to later.

1. Currently, we can't check if the functions works properly and I have a feeling it will need more things to be added. When there is a _working_ implementation of AP bootup, we can start refactorings like that. 
2. What's common here? HalStartNextProcessor call, MmCreateKernelStack and KeLoaderBlock initialization? I feel like it's not enough to do the split, currently you can see all actions done in one place, with your approach small functions will be spread across files (so harder to read code)
3. @DarkFire01 Don't put arguments like "This is what my mentor suggested me to do so I PRed this version", that's not constructive. Of you haven't understood my arguments, ask me :)

DarkFire01: Understood I will switch to the other one!

ntoskrnl/ke/i386/mp.c
```diff
@@ -0,0 +1,127 @@
...
+    KPROCESSOR_STATE ProcessorState;
+    KDESCRIPTOR GdtDesc, IdtDesc;
+
+    HighestPhysicalAddress.QuadPart = UINT64_MAX;
+    __sgdt(&GdtDesc.Limit);
+    __sidt(&IdtDesc.Limit);
+
+    /* Attempt to allocate memory for the new setup */
+    SIZE_T APInfoSize = sizeof(struct APInfo) + 2;
+    PVOID PAPInfo = MmAllocateContiguousMemory(APInfoSize, HighestPhysicalAddress);
```
Extravert-ir, line 46: If you think these fields won't be used in AP boot code, that should be a simple NonPagedPool

ntoskrnl/ke/i386/mp.c
Extravert-ir: What is that needed for? That should be inside the structure

ntoskrnl/ke/i386/mp.c
```diff
@@ -0,0 +1,127 @@
...
+
+#include <ntoskrnl.h>
+#define NDEBUG
+#include <debug.h>
+
+/* GLOBALS *******************************************************************/
+
+#define PHYSICAL_ADDRESS  LARGE_INTEGER
+PHYSICAL_ADDRESS HighestPhysicalAddress;
+ULONG ProcessorCount = 0;
```
Extravert-ir, line 18: What is the difference between this variable and `StartedProcessorCount` from HAL?

hal/halx86/smp/i386/apentry.S
tkreuzer: Nevermind, in fact each CPU needs it's own GDT (for FS and TSS)

ntoskrnl/ke/i386/mp.c
```diff
@@ -0,0 +1,127 @@
...
+
+        /* Clear GS */
+        ProcessorState.ContextFrame.SegGs = 0;
+        /* Set FS to PCR */
+        ProcessorState.ContextFrame.SegFs = KGDT_R0_PCR;
+
+        /* Load TSR */
+        ProcessorState.SpecialRegisters.Tr = KGDT_TSS;
+
+        ProcessorState.ContextFrame.Eip = (ULONG_PTR)KiSystemStartup;
```
tkreuzer, line 81: I don't think any of this is needed here, since the values are hardcoded and can be loaded in APSpinup as well.

HBelusca: The NT kernel is responsible for telling the HAL what state to initialize the application processors. The HAL just starts the processors and place them in whatever state the kernel told it to use. In particular the kernel knows what protected mode entrypoint the application processors need to execute ultimately.

ntoskrnl/ke/i386/mp.c
tkreuzer: Seems a bit strange, but ok.

ntoskrnl/ke/i386/mp.c
```diff
@@ -0,0 +1,127 @@
...
+
+    } while (HalStartNextProcessor(KeLoaderBlock, &ProcessorState));
+
+    ProcessorCount--;
+    /* Started means the AP itself is online, this doesn't mean it's seen by the kernel */
+    DPRINT1("HalStartNextProcessor: Sucessful AP startup count is %X\n", ProcessorCount);
+    /* We have finished starting processors, Time to cleanup! */
+    MmFreeContiguousMemory(PAPInfo);
+    MmDeleteKernelStack(KernelStack, FALSE);
+    MmDeleteKernelStack(DPCStack, FALSE);
```
tkreuzer, line 126: This shouldn't be freed on success. Each CPU needs the stuff in there.

ntoskrnl/ke/i386/mp.c
tkreuzer: And it should be allocated for each CPU.

ntoskrnl/ke/i386/mp.c
DarkFire01: This is ran an extra time for a CPU that doesn't exist, 
if HalStartNextProcessor returns false, we didn't successfully start an AP.
So what should we do with this memory that never got an AP assigned to it?

HBelusca: Only in case of failure to start an AP, should that per-CPU memory be deleted.

DarkFire01: That's whats happening here is it not?

HBelusca: It would have been more obvious if the code was looking like:
```c
while (TRUE /* or some hardcoded limitation on the number of CPUs */)
{
    // Initialize the per-CPU data...
    // Call KiInitializePcr() and the other helpers...

    if (!HalStartNextProcessor(KeLoaderBlock, &ProcessorState))
    {
        /* Cannot start anymore CPUs: cleanup and bail out */
        // Delete the two allocated stacks.
        // Delete PAPInfo
        break;
    }

    /* The CPU started: perform additional initialization */

    // Wait for the CPU to actually startup into KiSystemStartup
    while (/* some condition */)
    {
        // Do something....
    }
    if (/* CPU still not started */)
    {
        // Tell it somehow to stop whatever it is doing
        // Do some cleanup for this CPU, but continue the main loop.
    }
}
```

ntoskrnl/ke/amd64/mp.c
```diff
@@ -0,0 +1,75 @@
...
+    PVOID KernelStack;
+    KPROCESSOR_STATE ProcessorState;
+    KDESCRIPTOR GdtDesc, IdtDesc;
+
+    HighestPhysicalAddress.QuadPart = UINT64_MAX;
+    __sgdt(&GdtDesc.Limit);
+    __sidt(&IdtDesc.Limit);
+
+    /* Attempt to allocate memory for the new setup */
+    SIZE_T APInfoSize = sizeof(struct APInfo) + 2;
```
tkreuzer, line 44: `+ 2`? Is that your memory donation for a charitable cause?

ntoskrnl/ke/amd64/mp.c
tkreuzer: I was primarily referring to the use of pointer arithmetic. `There is a structure definition for it` means there is one in line 20 of this file, so there it should be used instead of `(ULONG_PTR)PAPInfo + sizeof(KPCR) + ...`
The comment about using the PROCESSOR_STATE was related to copying the KDESCRIPTOR, which is part of KPROCESSOR_STATE. But what is done here is even more crazy, as it copies the entire GDT into a KDESCPRIPTOR field.

ntoskrnl/ke/i386/mp.c
tkreuzer: Yeah, and I don't think the entire GDT fits into the KDESCRIPTOR field either.

ntoskrnl/ke/i386/mp.c
tkreuzer: After having another look, I got aware that what has been tried to do here was to copy the GDT, not tthe KDESCRIPTOR. So the structure should rather contain a `KGDTENTRY Gdt[128];`

HBelusca: Something like this, except that the actual number of entries should be deduced from the BSP' GDT.Limit value.

ntoskrnl/ke/i386/mp.c
tkreuzer: Yeah, I left a comment about that. adding 2 bytes and then copying the entire GDT into the descriptor doesn't really work. 😆

ntoskrnl/ke/i386/mp.c
tkreuzer: Ah, so this is Windows compatibility thing. Fair enough.

ntoskrnl/ke/i386/mp.c
tkreuzer: According to a comment in freeldr, the GDT is always 128 entries. And that is what I've seen on Windows as well. No idea why, but "that's what Windows does".

ntoskrnl/ke/amd64/mp.c
DarkFire01: Im going to remember i did this and cringe at my completely ignorance.
fixed it. ill push it next time i push everything

it was a donation to the Justin is a little special foundation

ntoskrnl/ke/i386/mp.c
DarkFire01: Well ones keeping track of how many times the KeStartAllProcessors loop has been ran, the other is to actually eventually associate an NTProcessorNumber to ApicID. 
Though this could be cleaned probably.

ntoskrnl/ke/mp.c
```diff
@@ -0,0 +1,127 @@
...
+ULONG ProcessorCount = 0;
+
+struct APInfo
+{
+    KPCR pcr;
+    KTSS tss;
+    ETHREAD thread;
+    KDESCRIPTOR GdtDesc;
+    KDESCRIPTOR IdtDesc;
+} APInfo;
```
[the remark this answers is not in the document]
HBelusca: Why moving the Tss member in another place?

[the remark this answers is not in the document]
tkreuzer: Kindof seperating the "portable" part from the architecture specific part. It's not important, I just felt it would be better placed there.

ntoskrnl/ke/mp.c
```diff
@@ -34,35 +32,27 @@ KeStartAllProcessors()
...
     KPROCESSOR_STATE ProcessorState;
     KDESCRIPTOR GdtDesc, IdtDesc;
 
-    HighestPhysicalAddress.QuadPart = UINT64_MAX;
     __sgdt(&GdtDesc.Limit);
     __sidt(&IdtDesc.Limit);
 
     /* Attempt to allocate memory for the new setup */
-    SIZE_T APInfoSize = sizeof(struct APInfo);
+    SIZE_T APInfoSize = sizeof(APInfo);
```
tkreuzer, line 43: This gives you the size of APInfo, which is a pointer, so the result is 4/8. What you want is `sizeof(*APInfo)` or `sizeof(APINFO)`.

ntoskrnl/ke/mp.c
```diff
@@ -0,0 +1,119 @@
...
+{
+    KIPCR Pcr;
+    KTHREAD Thread;
+    KGDTENTRY Gdt[128];
+    KIDTENTRY Idt;
+    KTSS Tss;
+} APINFO, *PAPINFO;
+
+/* FUNCTIONS *****************************************************************/
+VOID
```
disean, line 28:
```suggestion

CODE_SEG("INIT")
VOID
```

sdk/include/ndk/kefuncs.h
```diff
@@ -73,6 +73,12 @@ KiDeliverApc(
 //
 // Process/Thread Functions
 //
+VOID
```
disean, line 76:
```suggestion
CODE_SEG("INIT")
VOID
```

disean (commented):
You might want to make this function discardable

hal/halx86/smp/i386/apspinup.S
```diff
@@ -0,0 +1,77 @@
...
+    mov esp, [ecx + HEX(4096)] 
+
+    /* Load the GDT based on BSP */
+
+    /* TODO: replace this with something that's going to actually work */
+    xor eax, eax
+    mov eax, ([ebx] + [APCr3 - _APSpinup])
+    mov cr3, eax
+
+    /* Load some ProcessorState values into proper area */
```
Extravert-ir, line 32: The full ProcessorState must be loaded here (control, debug, segment, general purpose registers, etc.)

Extravert-ir (changes requested):

hal/halx86/smp/i386/apentry.S
```diff
@@ -0,0 +1,104 @@
...
+    /* bitshift 4 to get the address */
+    mov bx, cs
+    shl ebx, HEX(4)
+    add ebx, (gdt - _APEntry)
+    mov dword ptr cs:[gdtoffset - _APEntry], ebx
+
+    /* Load the GDT and blank IDT */
+#ifdef _USE_ML
+    lgdt fword ptr cs:[gdtptr - _APEntry]
+    lidt fword ptr cs:[idtdesc - _APEntry]
```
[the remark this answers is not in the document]
DarkFire01: Don't we load that in the protected mode area though? along with the GDT based on the BSPs one

hal/halx86/smp/mpmemory.c
```diff
@@ -0,0 +1,110 @@
...
+}
+#elif _M_IX86
+VOID
+HalpWriteProcessorState(PVOID APStubLocation, 
+                        PKPROCESSOR_STATE ProcessorState)
+{
+    PVOID APProcessorStateLoc;
+    APSTUB APStub;
+    APProcessorStateLoc = (PUSHORT)(((ULONG_PTR)APStubLocation) + ((ULONG_PTR)&APProcessorStateStruct - (ULONG_PTR)&APSpinup) + ((ULONG_PTR)&APEntryEnd  - (ULONG_PTR)&APEntry));
+    APStub.StructAPCr0 = ProcessorState->SpecialRegisters.Cr0;
```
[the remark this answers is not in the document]
DarkFire01: I don't see why id be able to, I don't use the entire KPROCESSOR_STATE, what we have here is just a reference of one initialized earlier with a specific set of values.

ntoskrnl/ke/mp.c
```diff
@@ -0,0 +1,140 @@
...
+        ProcessorState->SpecialRegisters.Cr4 = __readcr4();
+        ProcessorState->ContextFrame.Rip = (ULONG_PTR)KiSystemStartup;
+    }
+#else //_M_IX86
+VOID
+NTAPI
+KxInitAPProcessorState(
+    _Out_ PKPROCESSOR_STATE ProcessorState)
+    {
+        /* Prep Cr Regsters */
```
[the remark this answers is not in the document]
DarkFire01: What do you mean by "zero out ProcessorState" in this context?
And yup Debug registers will be saved too!

hal/halx86/smp/i386/apentry.S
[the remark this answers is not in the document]
DarkFire01: Actually thinking about it again; I will just remove that segment.
 As it gets loaded again later like i said above, so it doesn't need to be done twice.

hal/halx86/smp/mpmemory.c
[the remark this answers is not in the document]
Extravert-ir: But you *need* to use the entire KPROCESSOR_STATE :D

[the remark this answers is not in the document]
DarkFire01: How exactly?

ntoskrnl/ke/mp.c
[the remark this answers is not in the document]
DarkFire01: Alright, understood! thank you!

hal/halx86/smp/i386/apspinup.S
Extravert-ir: You need to prepare an AP for calling `KiSystemStartup`. In this particular case, I'd say GP registers are needed to be zeroed it least

DarkFire01: But that has nothing to do with processorstate, also don't i need to push a parameter on the stack? that already removes atleast 1 one from being zeroed before this occurs

hal/halx86/smp/mpmemory.c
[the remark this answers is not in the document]
DarkFire01: Alright, actually just consider it done. I was under the impression we don't actually use the entire ProcessorState. Even if we don't copying the entire thing over doesn't make much of a difference to me. 
Will do.

ntoskrnl/ke/mp.c
```diff
@@ -0,0 +1,140 @@
...
+CODE_SEG("INIT")
+VOID
+NTAPI
+KeStartAllProcessors()
+{
+    PVOID KernelStack;
+    PVOID DPCStack;
+    PAPINFO APInfo;
+    KPROCESSOR_STATE ProcessorState;
+    LOADER_PARAMETER_BLOCK KeLoaderBlock;
```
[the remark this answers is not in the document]
Extravert-ir: So `KeLoaderBlock` keep the state of the latest started CPU?

hal/halx86/smp/mpmemory.c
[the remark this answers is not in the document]
Extravert-ir: The thing is, it's easier to just copy the whole state. On i386 you really don't need GP registers, but for amd64 for example, a parameter for `KiSystemStartup` will go to a register

ntoskrnl/ke/mp.c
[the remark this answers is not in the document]
HBelusca: Yes, this should be visible e.g. on MP windows and looking at what it contains.

hal/halx86/smp/i386/apspinup.S
Extravert-ir: ProcessorState contains the whole processor state, that's what it has to do :)
You can just be dumb here and load what it has unconditionally, and let KeStartAllProcessors decide what should be put there. You may suggest a different solution, but I feel like that would be a good separation of responsibility areas

ntoskrnl/ke/mp.c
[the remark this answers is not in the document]
Extravert-ir: Then this must be removed completely

binarymaster:
PR is rekt, please fix

ntoskrnl/ke/mp.c
```diff
@@ -0,0 +1,307 @@
...
+            MM_PAGE_SIZE - 1, TYPE_DATA, 0, 2);
+        KiSetGdtEntry(KiGetGdtEntry(&APInfo->Gdt, KGDT_DF_TSS), (ULONG_PTR)&APInfo->TssDoubleFault,
+            0xFFFF, 0x09, 0, 2);
+        KiSetGdtEntry(KiGetGdtEntry(&APInfo->Gdt, KGDT_NMI_TSS), (ULONG_PTR)&APInfo->TssNMI,
+            0xFFFF, TYPE_CODE, 0, 2);
+        
+        BootStubPTEPhy = MmGetPhysicalAddress(BootStubPTE);
+        ProcessorState.ContextFrame.Eax = (ULONG_PTR)BootStubPTE;
+        ProcessorState.ContextFrame.Ecx = BootStubPTEPhy.QuadPart;
+        ProcessorState.ContextFrame.Ebx = PageTablePhysicalLoc.QuadPart;
```
Extravert-ir, line 106: What is this?

Extravert-ir (changes requested):

ntoskrnl/ke/mp.c
```diff
@@ -0,0 +1,307 @@
...
+NTAPI
+KeStartAllProcessors()
+{
+    ProcessorCount = 0;
+    while (TRUE)
+    {
+        SIZE_T APInfoSize = sizeof(APINFO);
+        PHARDWARE_PTE PDE, BootStubPTE, GDTPTE, IDTPTE;
+        APInfo = ExAllocatePool(NonPagedPool, APInfoSize);
+        PDE = ExAllocatePool(NonPagedPool, 100000); //TODO replace the silly size place holder
```
Extravert-ir, line 38: One PDE has a size of exactly one page (4096 bytes)

hal/halx86/smp/mpmemory.c
```diff
@@ -0,0 +1,175 @@
...
+    /* Calculate and Write Locations for jumps */
+    APJumpOffset = (UINT16)(((ULONG_PTR)APStubLocation * 4) + ((ULONG_PTR)&APEntryEnd  - (ULONG_PTR)&APEntry));
+    APFinalOffset = (UINT16)(((ULONG_PTR)APStubLocation * 4) + ((ULONG_PTR)&APEntryEnd  - (ULONG_PTR)&APEntry) + ((ULONG_PTR)&APSpinupEnd - (ULONG_PTR)&APSpinup));
+    RtlCopyMemory(APJumppLoc, &APJumpOffset,  sizeof(APJumpOffset));
+    RtlCopyMemory(APFinalLoc, &APFinalOffset,  sizeof(APFinalOffset));
+
+    /* Copy GDT Descriptor into area */
+    GdtLoc = (PKDESCRIPTOR)((ULONG_PTR)APStubLocation + ((ULONG_PTR)&APGDT - (ULONG_PTR)&APSpinup) + (ULONG_PTR)&APEntryEnd  - (ULONG_PTR)&APEntry);
+    GdtLoc->Limit = FinalGdt.Limit;
+    GdtLoc->Base = FinalGdt.Base;
```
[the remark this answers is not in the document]
HBelusca: If the low stub only needs the GDT and IDT descriptor values, there's no point in copying the whole processorstate? 🤔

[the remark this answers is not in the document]
Extravert-ir: It needs CR, DR, segment registers, descriptor tables, stack, IP register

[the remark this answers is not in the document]
Extravert-ir: I explained that already several messages before

DarkFire01:
Super seeded by 
https://github.com/reactos/reactos/pull/4140

PR #3798, opened 2021-07-05
Fix locking of Working set in various places

zefklop:
## Purpose

Avoid some race conditions, non-serialized read & writes on the VAD tree, etc.

JIRA issues: [CORE-17595](https://jira.reactos.org/browse/CORE-17595) [CORE-17642](https://jira.reactos.org/browse/CORE-17642) 

## Proposed changes

- Get rid of old MmAccessFault, use MmArm3Accessfault instead (renamed to MmAccessFault)
- From MmAccessFault, handle Working Set locking
  - Do not lock when not coming from a trap
  - Fix callers accordingly
- From MmAccessFault, dispatch to old Mm if needed
- Misc related fixes


modules/CMakeLists.txt
```diff
@@ -1,4 +1,8 @@
 
+option(ENABLE_ROSAPPS "Enable building the ReactOS extra applications" OFF)
+option(ENABLE_ROSTESTS "Enable building the ReactOS test suite" OFF)
+option(ENABLE_WALLPAPERS "Integrate wallpapers into the built ISOs" OFF)
+
```
Doug-Lyons, line 5: This looks like a stray commit that has already been merged into master.

Doug-Lyons: Please check the ability to use DPH after this patch. It seems to have broken this for me when I use the following:
gflags /p /enable regedit.exe /full. After this if I run regedit.exe, I get a memory access fault (0xC0000005). Thanks.

zefklop: I don't have such behaviour

learn-more: @Doug-Lyons if you enable break on first chance, you can investigate what it is.

JoachimHenze:
zefklop, you mentioned this PR in https://jira.reactos.org/browse/CORE-17698 . So it sounds in JIRA as if it would fix that. Understood right?
Why don't you add https://jira.reactos.org/browse/CORE-17698 and https://jira.reactos.org/browse/CORE-17690 to the JIRA issues it addresses in the PRs description then?

JoachimHenze:
With your PR I can not longer reproduce https://jira.reactos.org/browse/CORE-17690 see https://jira.reactos.org/browse/CORE-17690?focusedCommentId=129089&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-129089 for a successful log with the patch.
I guess you should mention that ticket in the description also! It happifies me for CORE-17690.

JoachimHenze:
But this ticket does NOT fix CORE-17595 as the tickets description currently implies. I retested with the artifacts iso. So please remove that JIRA-ID from the PRs description! It's a false promise!
https://jira.reactos.org/browse/CORE-17595?focusedCommentId=129090&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-129090

modules/CMakeLists.txt
zefklop: I am able to reproduce. Investigating.

ntoskrnl/include/internal/mm.h
```diff
@@ -46,6 +46,8 @@ extern SIZE_T MmtotalCommitLimitMaximum;
 extern PVOID MiDebugMapping; // internal
 extern PMMPTE MmDebugPte; // internal
 
+extern KSPIN_LOCK MmPfnLock;
```
tkreuzer, line 49: Is this for the annotations? IIRC those don't care about what you put there. You can probably use _Requires_lock_not_held_(LockQueuePfnLock)

ntoskrnl/include/internal/mm.h
```diff
@@ -822,17 +818,27 @@ NTAPI
...
     IN ULONG FaultCode,
+    IN PMMSUPPORT AddressSpace,
+    IN PMEMORY_AREA MemoryArea,
     IN PVOID Address,
     IN KPROCESSOR_MODE Mode,
     IN PVOID TrapInformation
 );
 
+/* pagfault.c *********************************************************************/
+NTSTATUS
```
HBelusca, line 834:
```suggestion

NTSTATUS
```

ntoskrnl/include/internal/mm.h
```diff
@@ -940,22 +946,32 @@ MmGetSectionAssociation(PFN_NUMBER Page,
                         PLARGE_INTEGER Offset);
 
 /* freelist.c **********************************************************/
-
+_IRQL_raises_(DISPATCH_LEVEL)
```
HBelusca, line 949:
```suggestion

_IRQL_raises_(DISPATCH_LEVEL)
```

ntoskrnl/include/internal/mm.h
```diff
@@ -1476,10 +1492,18 @@ MmUnsharePageEntrySectionSegment(PMEMORY_AREA MemoryArea,
                                  BOOLEAN PageOut,
                                  ULONG_PTR *InEntry);
 
+_When_(OldIrql == MM_NOIRQL, _IRQL_requires_max_(DISPATCH_LEVEL))
+_When_(OldIrql == MM_NOIRQL, _Requires_lock_not_held_(MmPfnLock))
+_When_(OldIrql != MM_NOIRQL, _Requires_lock_held_(MmPfnLock))
+_When_(OldIrql != MM_NOIRQL, _Releases_lock_(MmPfnLock))
+_When_(OldIrql != MM_NOIRQL, _IRQL_restores_(OldIrql))
+_When_(OldIrql != MM_NOIRQL, _IRQL_requires_(DISPATCH_LEVEL))
```
HBelusca, line 1500: 🤣

ntoskrnl/mm/ARM3/mdlsup.c
```diff
@@ -1179,39 +1192,22 @@ MmProbeAndLockPages(IN PMDL Mdl,
...
+            (PointerPxe->u.Hard.Valid == 0) ||
 #endif
 #if (_MI_PAGING_LEVELS >= 3)
-               (PointerPpe->u.Hard.Valid == 0) ||
+            (PointerPpe->u.Hard.Valid == 0) ||
 #endif
-               (PointerPde->u.Hard.Valid == 0) ||
-               (PointerPte->u.Hard.Valid == 0))
+            (PointerPde->u.Hard.Valid == 0) ||
+            (PointerPte->u.Hard.Valid == 0))
```
HBelusca, line 1205: indent was fine originally (with respect to the while)

ntoskrnl/mm/ARM3/mdlsup.c
```diff
@@ -1179,39 +1192,22 @@ MmProbeAndLockPages(IN PMDL Mdl,
...
-            else
-            {
-                /* Release process working set */
-                MiUnlockProcessWorkingSet(CurrentProcess, PsGetCurrentThread());
-            }
-
-            //
-            // Access the page
-            //
+            /* Access the page if this not non-paged pool */
```
HBelusca, line 1207:
```suggestion
            /* Access the page if this is not non-paged pool */
```

ntoskrnl/mm/ARM3/pagfault.c
```diff
@@ -291,16 +291,25 @@ MiCheckVirtualAddress(IN PVOID VirtualAddress,
...
             ASSERT(Vad->u2.VadFlags2.ExtendableFile == 0);
 
             /* Return the proto VAD */
             *ProtoVad = Vad;
 
-            /* Get the prototype PTE for this page */
-            PointerPte = (((ULONG_PTR)VirtualAddress >> PAGE_SHIFT) - Vad->StartingVpn) + Vad->FirstPrototypePte;
-            ASSERT(PointerPte != NULL);
-            ASSERT(PointerPte <= Vad->LastContiguousPte);
+            /* Check for MEMORY_AREA thingie */
```
HBelusca, line 299: what is this thingie?

ntoskrnl/mm/ARM3/section.c
```diff
@@ -836,24 +835,14 @@ MiUnmapViewOfSection(IN PEPROCESS Process,
...
-
     /* Check if we should attach to the process */
     if (CurrentProcess != Process)
     {
         /* The process is different, do an attach */
         KeStackAttachProcess(&Process->Pcb, &ApcState);
         Attached = TRUE;
     }
-
+
```
HBelusca, line 845:
```suggestion

```

ntoskrnl/mm/i386/page.c
```diff
@@ -198,12 +198,11 @@ MmGetPfnForProcess(PEPROCESS Process,
     /* And for our process */
     ASSERT(Process == PsGetCurrentProcess());
 
-    /* Lock for reading */
-    MiLockProcessWorkingSetShared(Process, PsGetCurrentThread());
+    /* Make sure we're locking it right. */
+    ASSERT(PsGetCurrentThread()->OwnsProcessWorkingSetExclusive || PsGetCurrentThread()->OwnsProcessWorkingSetShared);
```
HBelusca, line 202:
```suggestion
    ASSERT(PsGetCurrentThread()->OwnsProcessWorkingSetExclusive ||
           PsGetCurrentThread()->OwnsProcessWorkingSetShared);
```

ntoskrnl/mm/i386/page.c
```diff
@@ -266,18 +263,17 @@ MmDeleteVirtualMapping(PEPROCESS Process, PVOID Address,
         }
 
         /* Only for current process !!! */
-        ASSERT(Process = PsGetCurrentProcess());
-        MiLockProcessWorkingSetUnsafe(Process, PsGetCurrentThread());
+        ASSERT(Process == PsGetCurrentProcess());
```
HBelusca, line 266: ouch that bug!

ntoskrnl/mm/i386/page.c
```diff
@@ -399,21 +368,19 @@ MmIsPagePresent(PEPROCESS Process, PVOID Address)
     ASSERT(Process != NULL);
     ASSERT(Process == PsGetCurrentProcess());
 
-    MiLockProcessWorkingSetShared(Process, PsGetCurrentThread());
+    /* Must hold some lock */
+    ASSERT(PsGetCurrentThread()->OwnsProcessWorkingSetExclusive || PsGetCurrentThread()->OwnsProcessWorkingSetShared);
```
HBelusca, line 372:
```suggestion
    ASSERT(PsGetCurrentThread()->OwnsProcessWorkingSetExclusive ||
           PsGetCurrentThread()->OwnsProcessWorkingSetShared);
```

ntoskrnl/mm/i386/page.c
```diff
@@ -442,12 +409,11 @@ MmIsDisabledPage(PEPROCESS Process, PVOID Address)
         ASSERT(Process != NULL);
         ASSERT(Process == PsGetCurrentProcess());
 
-        MiLockProcessWorkingSetShared(Process, PsGetCurrentThread());
+        NT_ASSERT(PsGetCurrentThread()->OwnsProcessWorkingSetExclusive || PsGetCurrentThread()->OwnsProcessWorkingSetShared);
```
HBelusca, line 412:
```suggestion
        NT_ASSERT(PsGetCurrentThread()->OwnsProcessWorkingSetExclusive ||
                  PsGetCurrentThread()->OwnsProcessWorkingSetShared);
```
Why using an NT_ASSERT here, but in other places where similar check is done, a simple ASSERT was used instead?

ntoskrnl/mm/i386/page.c
```diff
@@ -482,12 +445,11 @@ MmIsPageSwapEntry(PEPROCESS Process, PVOID Address)
     ASSERT(Process != NULL);
     ASSERT(Process == PsGetCurrentProcess());
 
-    MiLockProcessWorkingSetShared(Process, PsGetCurrentThread());
+    NT_ASSERT(PsGetCurrentThread()->OwnsProcessWorkingSetExclusive || PsGetCurrentThread()->OwnsProcessWorkingSetShared);
```
HBelusca, line 448:
```suggestion
    NT_ASSERT(PsGetCurrentThread()->OwnsProcessWorkingSetExclusive ||
              PsGetCurrentThread()->OwnsProcessWorkingSetShared);
```

ntoskrnl/mm/i386/page.c
```diff
@@ -517,13 +477,11 @@ MmGetPageFileMapping(PEPROCESS Process, PVOID Address, SWAPENTRY* SwapEntry)
 
     ASSERT(Process != NULL);
     ASSERT(Process == PsGetCurrentProcess());
-
-    MiLockProcessWorkingSetShared(Process, PsGetCurrentThread());
+    ASSERT(PsGetCurrentThread()->OwnsProcessWorkingSetExclusive || PsGetCurrentThread()->OwnsProcessWorkingSetShared);
```
HBelusca, line 480:
```suggestion
    ASSERT(PsGetCurrentThread()->OwnsProcessWorkingSetExclusive ||
           PsGetCurrentThread()->OwnsProcessWorkingSetShared);
```

ntoskrnl/mm/i386/page.c
```diff
@@ -705,12 +653,11 @@ MmGetPageProtect(PEPROCESS Process, PVOID Address)
 
         ASSERT(Process == PsGetCurrentProcess());
 
-        MiLockProcessWorkingSetShared(Process, PsGetCurrentThread());
+        ASSERT(PsGetCurrentThread()->OwnsProcessWorkingSetExclusive || PsGetCurrentThread()->OwnsProcessWorkingSetShared);
```
HBelusca, line 656:
```suggestion
        ASSERT(PsGetCurrentThread()->OwnsProcessWorkingSetExclusive ||
               PsGetCurrentThread()->OwnsProcessWorkingSetShared);
```

ntoskrnl/mm/marea.c
```diff
@@ -300,11 +300,9 @@ MmFreeMemoryArea(
...
         PEPROCESS Process = MmGetAddressSpaceOwner(AddressSpace);
 
-        if (Process != NULL &&
-                Process != CurrentProcess)
-        {
-            KeAttachProcess(&Process->Pcb);
-        }
+        ASSERT((Process == NULL) || (Process == CurrentProcess));
+        NT_ASSERT(((Process == NULL) && PsGetCurrentThread()->OwnsSystemWorkingSetExclusive) || 
+            ((Process != NULL) && PsGetCurrentThread()->OwnsProcessWorkingSetExclusive));
```
HBelusca, line 305:
```suggestion
        NT_ASSERT(((Process == NULL) && PsGetCurrentThread()->OwnsSystemWorkingSetExclusive) || 
                  ((Process != NULL) && PsGetCurrentThread()->OwnsProcessWorkingSetExclusive));
```

ntoskrnl/mm/section.c
```diff
@@ -1599,31 +1592,34 @@ MmNotPresentFaultSectionView(PMMSUPPORT AddressSpace,
...
+        MmDeletePageFileMapping(Process, Address, &SwapEntry);
         if (SwapEntry == MM_WAIT_ENTRY)
         {
-            MmUnlockAddressSpace(AddressSpace);
-            YieldProcessor();
-            MmLockAddressSpace(AddressSpace);
+            /* Someone else is handling this page fault.
+             * Restore this. */
+            MmCreatePageFileMapping(Process, Address, MM_WAIT_ENTRY);
+            /* Let the other one a chance to finish */
```
HBelusca, line 1601:
```suggestion
            /* Let the other one have a chance to finish */
```

ntoskrnl/include/internal/mm.h
zefklop: No, it's for the assert below.

ntoskrnl/mm/ARM3/pagfault.c
zefklop: the hack where a VAD is a proxy to a MEMORY_AREA from legacy Mm

ntoskrnl/mm/i386/page.c
zefklop: NT_ASSERT uses annotations, while ASSERT is just a macro forwarding to RtlAssert

ntoskrnl/mm/ARM3/virtual.c
```diff
@@ -2308,16 +2255,43 @@ MiProtectVirtualMemory(IN PEPROCESS Process,
...
-        /* Check if data or page file mapping protection PTE is compatible */
-        if (!Vad->ControlArea->u.Flags.Image)
+        /* Handle legacy MM case */
+        if (Vad->u.VadFlags.Spare)
         {
-            /* Not yet */
-            DPRINT1("Fixme: Not checking for valid protection\n");
+            PMEMORY_AREA MemoryArea = (PMEMORY_AREA)Vad;
+
+            if (MemoryArea->DeleteInProgress)
```
tkreuzer, line 2316: Can we move this to `` MmDeleteVirtualMapping`` ?

ntoskrnl/mm/i386/page.c
tkreuzer: We define ASSERT to NT_ASSERT, so no need to rewrite all ASSERTs 😉

ntoskrnl/mm/section.c
zefklop: https://www.linguee.fr/anglais-francais/traduction/give+a+chance.html ;-)

HBelusca: ok I was perturbed due to the "let"

ntoskrnl/mm/i386/page.c
zefklop: Eh...

ntoskrnl/mm/ARM3/pagfault.c
```diff
@@ -1919,19 +1917,22 @@ _WARN("Session space stuff is not implemented yet!")
...
+        {
+
+            KeRaiseIrql(APC_LEVEL, &LockIrql);
+            MiLockWorkingSet(CurrentThread, WorkingSet);
+        }
+        else
+        {
+            /* IRQL was already checked above */
+            ASSERT(IsSessionAddress ? PsGetCurrentThread()->OwnsSessionWorkingSetExclusive : PsGetCurrentThread()->OwnsSystemWorkingSetExclusive);
+        }
```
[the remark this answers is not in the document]
zefklop: Two things:
1. I wouldn't say that passing bogus pointers to a function is fine.
2. This avoids unlocking and locking again each time we call MmAccessFault. And when you do such things, you're supposed to read again any paging structure that you read before, because the stuff you just read might have disappear under the action of the modified page writer, the working set trimmer or anything else freeing or allocating virtual Memory. So "proper locking & unlocking" is full of traps in this case.

The hack added in `MiMakeSystemAddressValid` is removed in a later commit.

ntoskrnl/mm/ARM3/pagfault.c
```diff
@@ -2230,16 +2230,9 @@ _WARN("Session space stuff is not implemented yet!")
 
         if (ProtectionCode == MM_NOACCESS)
         {
-#if (_MI_PAGING_LEVELS == 2)
-            /* Could be a page table for paged pool */
-            MiCheckPdeForPagedPool(Address);
-#endif
```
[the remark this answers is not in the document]
zefklop: This function is only useful for kernel mode addresses, we're in the UM code path.

ntoskrnl/mm/ARM3/mdlsup.c
zefklop: Comment fixed, I originally wrote it for a different code.

ntoskrnl/mm/section.c
```diff
@@ -991,41 +991,33 @@ VOID
...
-        HaveLock = TRUE;
         OldIrql = MiAcquirePfnLock();
     }
 
     if (InterlockedDecrement64(Segment->ReferenceCount) > 0)
     {
         /* Nothing to do yet */
-        if (HaveLock)
-            MiReleasePfnLock(OldIrql);
+        MiReleasePfnLock(OldIrql);
```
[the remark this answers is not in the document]
zefklop: Discussion in #3850

ntoskrnl/mm/marea.c
```diff
@@ -317,11 +322,27 @@ MmFreeMemoryArea(
...
             {
-                MmDeleteVirtualMapping(Process, (PVOID)Address, &Dirty, &Page);
+                if (MmDeleteVirtualMapping(Process, (PVOID)Address, &Dirty, &Page) && Process)
+                {
+                    if (MiDecrementPageTableReferences((PVOID)Address) == 0)
+                    {
+                        KIRQL OldIrql = MiAcquirePfnLock();
+                        MiDeletePde(MiAddressToPde(Address), Process);
+                        MiReleasePfnLock(OldIrql);
+                    }
```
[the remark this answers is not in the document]
zefklop: Because there are cases where you don't want MmDeleteVirtualMapping to dereference the PDE. For instance when swapping pages for a COW mapping.

ntoskrnl/mm/ARM3/pagfault.c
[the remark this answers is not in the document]
tkreuzer: Let's see. These are the places, where MmAccessFault is called from:

- Page fault handler: unlocked
- MmProbeAndLockPages (twice): unlocked
- MiMakeSystemAddressValid: unlocks and relocks
- MiMakeSystemAddressValidPfn: unlocks and relocks
- MiLockVirtualMemory: Aquires lock, releases it for MmAccessFault

Here's what MmAccessFault looks like:
```
MmAccessFautlt(...)
{
    if (Address >= MmSystemRangeStart)
    {
        MiSynchronizeSystemPde()
        MiCheckPdeForPagedPool()

        // Shouldn't be done for user PDEs!
        if (PointerPde->u.Hard.Valid == 0)
        {
            KeBugCheck();
        }


        TempPte = *PointerPte;
        if (TempPte.u.Hard.Valid == 1)
        {
            // not relevant
        }

        if (MI_IS_SESSION_PTE(Address))
        {
            MiCheckPdeForSessionSpace
        }

        if (MI_IS_PAGE_TABLE_OR_HYPER_ADDRESS(Address))
        {
            goto UserFault;
        }

        //Get correct workingset/process (session/normal)
        MiLockWorkingSet(CurrentThread, WorkingSet);

        // All the things locked kernel mode

        MiUnlockWorkingSet(CurrentThread, WorkingSet);
        return Status;
     }

UserFault:

    MiLockProcessWorkingSet(CurrentProcess, CurrentThread);

    // All the things locked user mode

    MiUnlockProcessWorkingSet(CurrentProcess, CurrentThread);
    return Status
}
```
There is actually a bug: We call MiMakeSystemAddressValid to page in user mode PDEs etc. This goes through the kernel mode path in MmAccessFault, which then bugchecks, when higher level PDEs are not present. This is wrong.

Suggested code change:

1. Factor out "All the things locked" for both kernel and user
2. Call the kernel thing from MiMakeSystemAddressValid (probably needs the extra 
3. Introduce MiMakePageTableAddressValid and call the user path from that
4. Call the user thing from MiLockVirtualMemory

This would make the page fault handler code much more readable, fix the paging in of PDEs, and remove unnecessary code pathes.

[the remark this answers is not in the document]
zefklop: >     * Page fault handler: unlocked
OK
>     * MmProbeAndLockPages (twice): unlocked
This is wrong. The WS lock must be held all along, otherwise the pages at the beginning of the probed range can be paged out by the time we're done with probing and all hell will break loose when actually locking the pages.
>     * MiMakeSystemAddressValid: unlocks and relocks
And warns caller about lock state change.
>     * MiMakeSystemAddressValidPfn: unlocks and relocks
Ditto
>     * MiLockVirtualMemory: Aquires lock, releases it for MmAccessFault
Changed in this PR.

ntoskrnl/mm/ARM3/pagfault.c
```diff
@@ -2226,6 +2256,46 @@ _WARN("Session space stuff is not implemented yet!")
...
+    }
+
+    /* Does this come from old Mm */
+    if (Vad && Vad->u.VadFlags.Spare)
+    {
+        /* Increment before. Ros Mm relies on this. */
+        if (TempPte.u.Long == 0)
+        {
+            MiIncrementPageTableReferences(Address);
+        }
```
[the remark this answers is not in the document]
zefklop: Sure.

ntoskrnl/mm/ARM3/virtual.c
zefklop: MmProtectSectionView ? Sure.

ntoskrnl/mm/section.c
```diff
@@ -4407,6 +4428,12 @@ MmMapViewInSystemSpaceEx (
     MmUnlockSectionSegment(Segment);
     MmUnlockAddressSpace(AddressSpace);
 
+    if (NT_SUCCESS(Status))
+    {
+        /* Map PDEs */
+        MiFillSystemPageDirectory(*MappedBase, *ViewSize);
```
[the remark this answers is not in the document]
zefklop: MiFillSystemPageDirectory uses the PFN lock internally.

ntoskrnl/mm/ARM3/pagfault.c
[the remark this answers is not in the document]
tkreuzer: Ok, but the refactoring I suggested can be used for those functions as well.

ntoskrnl/mm/ARM3/miarm.h
```diff
@@ -2339,18 +2339,40 @@ MiDeletePte(
...
-    IN KIRQL OldIrql
+_Must_inspect_result_
+FORCEINLINE
+BOOLEAN
+MiMakeSystemAddressValid(
+    _In_ PVOID PageTableVirtualAddress,
+    _In_ PEPROCESS CurrentProcess)
+{
+    BOOLEAN ret = _MiMakeSystemAddressValid(PageTableVirtualAddress, CurrentProcess);
+    KeMemoryBarrierWithoutFence();
```
tkreuzer, line 2355: Why would this be needed? A function call is a sequence point and everything (as visible to the current thread) is evaluated before the return. If the concern is, that this function is inlined on release builds and the write to the PTE is reordered to after access to the page, then we should rather add a memory barrier to the `` MI_WRITE_VALID_PTE``  and ``MI_UPDATE_VALID_PTE`` functions.

tkreuzer: Correction: In standard C, there is a sequence point between the return and the start of the next full expression, so `` MiMakeSystemAddressValid(PageTableVirtualAddress, CurrentProcess) + *(PUCHAR)PageTableVirtualAddress``  wouldn't have a sequence point, but hopefully nobody would write that code :).

zefklop: It's not about "being evaluated before the return", it's about compiler optimizing memory accesses before or after the call to such functions. Only volatiles are guaranteed to be read or written between sequence points.
I believe that an optimizing compiler can make the following actually access *PointerPte before the call to MiMakeSystemAddressValid :
```
PMMPTE PointerPte = MiAddressToPte(addr);
MiMakeSystemAddressValid(DifferentPointerPteWithinPDRange);
MMPTE TempPte =*PointerPte;
```

tkreuzer: In that case all code that maps any PTE is prone to this. Therefore I suggest adding the memory barrier to the end of MI_WRITE_VALID_P*E and MI_UPDATE_VALID_P*E, as well as to the beginning of MI_WRITE_INVALID_PTE.

ntoskrnl/mm/ARM3/pagfault.c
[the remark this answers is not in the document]
zefklop: working on it. This is not trivial :slightly_smiling_face:

ntoskrnl/mm/ARM3/mdlsup.c
```diff
@@ -1259,28 +1238,10 @@ MmProbeAndLockPages(IN PMDL Mdl,
...
-                            MiUnlockProcessWorkingSet(CurrentProcess, PsGetCurrentThread());
-                        }
-
                         //
                         // Access the page
                         //
-
-                        //HACK: Pass a placeholder TrapInformation so the fault handler knows we're unlocked
-                        Status = MmAccessFault(TRUE, Address, KernelMode, (PVOID)(ULONG_PTR)0xBADBADA3BADBADA3ULL);
+                        Status = MmAccessFault(TRUE, Address, KernelMode, NULL);
```
Extravert-ir, line 1244: Stupid remark, but you pass TRUE/FALSE values to an ULONG field.
Shouldn't we define a set of flags somewhere, because currently I see only a set of `MI_IS_*` set of macroses which use raw constants inside.

ntoskrnl/mm/ARM3/miarm.h
Extravert-ir: LTO may inline all the functions and start reordering code which once was in different ones.
Two questions: why not putting the barrier inside the function, and shouldn't here be something more hard than a compiler barrier? If that's important, how can we be sure CPU won't reorder stuff?

Extravert-ir (changes requested):

ntoskrnl/mm/ARM3/miarm.h
zefklop:
```
why not putting the barrier inside the function ?
```
The compiler won't see it on caller side if it's inside the function / in a different compilation unit.
```
If that's important, how can we be sure CPU won't reorder stuff?
```
We can't. But the CPU won't trigger a PF until it's "sure" it has to, and we would have the instructions actually making the PT valid in the pipeline.

ntoskrnl/mm/ARM3/mdlsup.c
zefklop: Yes, I have to make up some names.

ntoskrnl/mm/ARM3/pagfault.c
```diff
@@ -1676,608 +1762,352 @@ MiDispatchFault(IN ULONG FaultCode,
...
-        /* Not yet implemented in ReactOS */
-        ASSERT(MI_IS_PAGE_LARGE(PointerPde) == FALSE);
+        UNIMPLEMENTED_DBGBREAK("No support for hyper space in page fault handler!\n");
+        return STATUS_ACCESS_VIOLATION;
     }
 
-    /* Now capture the PTE. */
-    TempPte = *PointerPte;
+    /* PDE was just brought in. Barrier. */
+    KeMemoryBarrierWithoutFence();
```
tkreuzer, line 2106: Shouldn't be needed

ntoskrnl/mm/ARM3/pagfault.c
```diff
@@ -2401,31 +2219,15 @@ _WARN("Session space stuff is not implemented yet!")
...
          */
-        if (Address <= MM_HIGHEST_USER_ADDRESS
-#if _MI_PAGING_LEVELS >= 3
-            || MiIsUserPte(Address)
-#if _MI_PAGING_LEVELS == 4
-            || MiIsUserPde(Address)
-#endif
-#endif
-        )
+        if (Address <= MM_HIGHEST_USER_ADDRESS)
```
tkreuzer, line 2230: Can't this be asserted? I don't see how this could be a page table address here now.

ntoskrnl/mm/ARM3/pagfault.c
```diff
@@ -2434,6 +2236,7 @@ _WARN("Session space stuff is not implemented yet!")
         /* Is this a guard page? */
         if ((ProtectionCode & MM_PROTECT_SPECIAL) == MM_GUARDPAGE)
```
tkreuzer, line 2237: Do guard pages have a zero PTE? This whole path seems useless.

ntoskrnl/mm/ARM3/pagfault.c
```diff
@@ -2446,19 +2249,20 @@ _WARN("Session space stuff is not implemented yet!")
...
+            MiLockProcessWorkingSetForFault(CurrentProcess, CurrentThread, WsSafe, WsShared);
+            return Status;
         }
 
         /* Did we get a prototype PTE back? */
         if (!ProtoPte)
         {
             /* Is this PTE actually part of the PDE-PTE self-mapping directory? */
-            if (PointerPde == MiAddressToPde(PTE_BASE))
+            if (MiAddressToPde(Address) == MiAddressToPde(PTE_BASE))
```
tkreuzer, line 2265: This looks broken. Should probably be something like `if (MiIsPageTableAddress(Address))`. But like the above block, I think it's useless.

ntoskrnl/mm/ARM3/pagfault.c
```diff
@@ -2641,9 +2446,168 @@ _WARN("Session space stuff is not implemented yet!")
...
+            /* PXE/PPE/PDE (still) not valid, kill the system */
+            KeBugCheckEx(PAGE_FAULT_IN_NONPAGED_AREA,
+                         (ULONG_PTR)Address,
+                         FaultCode,
+                         (ULONG_PTR)TrapInformation,
+                         2);
+        }
+
+        /* Barrier between PDE check & data read */
+        KeMemoryBarrierWithoutFence();
```
tkreuzer, line 2560: Move to `MiIsPdeValid`

ntoskrnl/mm/ARM3/pagfault.c
```diff
@@ -1676,608 +1762,352 @@ MiDispatchFault(IN ULONG FaultCode,
...
+    if (TempPte.u.Hard.Valid == 1)
+    {
+        /* Bugcheck in case of invalid write or execute. No COW */
+        MiCheckBadWriteOrExecute(FaultCode, Address, TempPte, TrapInformation, FALSE);
 
-        /* Not yet implemented in ReactOS */
-        ASSERT(MI_IS_PAGE_LARGE(PointerPde) == FALSE);
-        ASSERT((!MI_IS_NOT_PRESENT_FAULT(FaultCode) && MI_IS_PAGE_COPY_ON_WRITE(PointerPte)) == FALSE);
+        /* We're good */
+        return STATUS_SUCCESS;
```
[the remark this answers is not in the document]
zefklop: We don't do CoW.

ntoskrnl/mm/ARM3/pagfault.c
zefklop: I'm unclear whether a page fault on hyper space address could occur for now. But this was checkd above, so I guess we can assert it.

ntoskrnl/mm/ARM3/miarm.h
tkreuzer: The CPU doesn't reorder reads and writes on the same CPU core/thread (exception is speculative execution and that doesn't cause faults).
If it's in a different compilation unit and not inlined, then the compiler cannot move memory accesses around it, because it doesn't know the side effects of that function. Otherwise no locking function would ever work without an explicit memory barrier after it.
So the only reason to add that memory barrier is to prevent the compiler from shuffling things around, when it inlines the function and (wrongly) determines that there is no dependency between them . So adding the memory barrier in the function does exactly what is needed.

ntoskrnl/mm/ARM3/miarm.h
```diff
@@ -2337,19 +2368,54 @@ MiDeletePte(
...
-NTAPI
+FORCEINLINE
+BOOLEAN
 MiMakeSystemAddressValidPfn(
-    IN PVOID VirtualAddress,
-    IN KIRQL OldIrql
-);
+    _In_ PVOID VirtualAddress,
+    _Inout_ PKIRQL OldIrql)
+{
```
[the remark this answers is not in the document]
zefklop: Yes. Fixed.

ntoskrnl/mm/ARM3/miarm.h
```diff
@@ -2337,19 +2368,54 @@ MiDeletePte(
...
+        MiUnlockWorkingSet(CurrentThread, &MmSystemCacheWs);
+
+        /* This flag will be useful later when we do better locking */
+        LockChange = TRUE;
+
+        /* Lock the PFN database again */
+        *OldIrql = MiAcquirePfnLock();
+    }
+
+    KeMemoryBarrierWithoutFence();
```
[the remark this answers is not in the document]
zefklop: Made it inline, like the others.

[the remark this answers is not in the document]
zefklop: Actually, MiMakeSystemAddressValidPfn still need a memory barrier.

ntoskrnl/mm/ARM3/pagfault.c
```diff
@@ -479,14 +478,101 @@ MiCheckPdeForPagedPool(IN PVOID Address)
...
+{
+    return
+#if _MI_PAGING_LEVELS == 4
+        MiAddressToPxe(Address)->u.Hard.Valid &&
+#endif
+#if _MI_PAGING_LEVELS >= 3
+        MiAddressToPpe(Address)->u.Hard.Valid &&
+#endif
+        MiAddressToPde(Address)->u.Hard.Valid;
+}
```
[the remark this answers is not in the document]
zefklop: Done

ntoskrnl/mm/ARM3/pagfault.c
```diff
@@ -479,14 +478,101 @@ MiCheckPdeForPagedPool(IN PVOID Address)
...
+        if (!(Pte.u.Long & PTE_READWRITE) &&
+            !(Pfn1->OriginalPte.u.Soft.Protection & MM_READWRITE) &&
+            (!AllowCopyOnWrite || !MI_IS_PAGE_COPY_ON_WRITE(&Pte)))
+        {
+            /* Crash with distinguished bugcheck code */
+            KeBugCheckEx(ATTEMPTED_WRITE_TO_READONLY_MEMORY,
+                            (ULONG_PTR)Address,
+                            Pte.u.Long,
+                            (ULONG_PTR)TrapInformation,
+                            11);
```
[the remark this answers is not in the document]
zefklop: Done

ntoskrnl/mm/ARM3/pagfault.c
```diff
@@ -1676,608 +1762,352 @@ MiDispatchFault(IN ULONG FaultCode,
...
+        /* PXE/PPE/PDE (still) not valid, kill the system */
+        KeBugCheckEx(PAGE_FAULT_IN_NONPAGED_AREA,
+                        (ULONG_PTR)Address,
+                        FaultCode,
+                        (ULONG_PTR)TrapInformation,
+                        2);
+    }
+
+    /* Barrier between PDE check & data read. */
+    KeMemoryBarrierWithoutFence();
```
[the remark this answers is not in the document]
zefklop: Done

ntoskrnl/mm/ARM3/pagfault.c
```diff
@@ -1676,608 +1762,352 @@ MiDispatchFault(IN ULONG FaultCode,
...
-            {
-                /* Bad boy, bad boy, whatcha gonna do, whatcha gonna do when ARM3 comes for you! */
-                KeBugCheckEx(DRIVER_CAUGHT_MODIFYING_FREED_POOL,
-                             (ULONG_PTR)Address,
-                             FaultCode,
-                             Mode,
-                             4);
-            }
+    /* Grab the PTE. Put a barrier between PDE validity check and us. */
+    KeMemoryBarrierWithoutFence();
```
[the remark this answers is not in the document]
zefklop: Done

ntoskrnl/mm/ARM3/pagfault.c
```diff
@@ -1676,608 +1762,352 @@ MiDispatchFault(IN ULONG FaultCode,
...
     {
-        /* Right now, we only handle scenarios where the PPE is totally empty */
-        ASSERT(PointerPpe->u.Long == 0);
+        DPRINT("Faultcode 0x%x denied because protection is 0x%x\n", FaultCode, *ProtectionCode);
+        return STATUS_ACCESS_VIOLATION;
+    }
 
-        /* This is only possible for user mode addresses! */
-        ASSERT(PointerPte <= MiHighestUserPte);
+    /* Right now, we only handle scenarios where the PXE is totally empty */
```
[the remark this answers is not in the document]
zefklop: Done

ntoskrnl/mm/ARM3/pagfault.c
zefklop: I'm not sure. MiBringInPageTable is not forced inline.

ntoskrnl/mm/ARM3/pagfault.c
```diff
@@ -1676,608 +1762,352 @@ MiDispatchFault(IN ULONG FaultCode,
...
-            MiGetPfnEntry(PointerPde->u.Hard.PageFrameNumber)->CallSite = _ReturnAddress();
-#endif
-        /* We should come back with APCs enabled, and with a valid PDE */
-        ASSERT(KeAreAllApcsDisabled() == TRUE);
-        ASSERT(PointerPde->u.Hard.Valid == 1);
+        return MiBringInPageTable(FaultCode,
+                                  MiPteToPde(Address),
+                                  Address,
+                                  CurrentProcess,
+                                  _MI_PAGING_LEVELS, /* Let caller handle references */
```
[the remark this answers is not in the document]
zefklop: This avoid MiIncrementPageTableReferences to be called. Will turn that into a BOOLEAN.

ntoskrnl/mm/ARM3/pagfault.c
zefklop: I'll have to check what NtAllocateVirtualMemory does.

zefklop: Actually, it looks like NtAllocateVirtualMemory allows this if you use MEM_COMMIT without specifying a base address.

ntoskrnl/mm/ARM3/pagfault.c
zefklop: Done

ntoskrnl/mm/ARM3/pagfault.c
zefklop: Moved in the PDE specific code

ntoskrnl/mm/ARM3/pagfault.c
```diff
@@ -1676,608 +1762,352 @@ MiDispatchFault(IN ULONG FaultCode,
...
-#if (_MI_PAGING_LEVELS >= 3)
-    /* Check if the PPE is valid */
-    if (PointerPpe->u.Hard.Valid == 0)
+    if (!MiIsAccessAllowed(*ProtectionCode, MI_IS_WRITE_ACCESS(FaultCode), MI_IS_INSTRUCTION_FETCH(FaultCode)))
     {
-        /* Right now, we only handle scenarios where the PPE is totally empty */
-        ASSERT(PointerPpe->u.Long == 0);
+        DPRINT("Faultcode 0x%x denied because protection is 0x%x\n", FaultCode, *ProtectionCode);
+        return STATUS_ACCESS_VIOLATION;
+    }
```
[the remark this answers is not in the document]
tkreuzer: Oh, nvm, now I'm getting it.

ntoskrnl/mm/ARM3/pagfault.c
```diff
@@ -1676,608 +1748,347 @@ MiDispatchFault(IN ULONG FaultCode,
...
-        }
-
-        /* Resolve a demand zero fault */
-        MiResolveDemandZeroFault(PointerPpe,
-                                 PointerPxe,
-                                 MM_EXECUTE_READWRITE,
-                                 CurrentProcess,
-                                 MM_NOIRQL);
+    /* Check if the PDE is valid */
+    if (PointerPde->u.Hard.Valid)
```
tkreuzer, line 1976: Use `IsPteValid` and remove the memory barrier above.

ntoskrnl/mm/ARM3/pagfault.c
tkreuzer: Wow, yet another of these awesome features: reserve and commit an entire region of guard pages :D

ntoskrnl/mm/ARM3/miarm.h
Extravert-ir: > The CPU doesn't reorder reads and writes on the same CPU core/thread

As far as I understand, this is true only on x86. Not sure if that's important for us right now though

ntoskrnl/mm/ARM3/miarm.h
```diff
@@ -2387,13 +2409,107 @@ MiRosUnmapViewInSystemSpace(
...
+ */
+
+FORCEINLINE
+BOOLEAN
+MiIsPteValid(PMMPTE PointerPte)
+{
+    BOOLEAN ret;
+    KeMemoryBarrierWithoutFence();
+    ret = PointerPte->u.Hard.Valid;
+    KeMemoryBarrierWithoutFence();
```
Extravert-ir, line 2446: Looks like we're trying to emulate `atomic_load()` from C11 here

ntoskrnl/mm/ARM3/pagfault.c
zefklop: Somewhere out there, there may be JIT compiler implementing garbage collection using this feature.

ntoskrnl/mm/ARM3/miarm.h
zefklop: Doesn't `atomic_load` involve some specific CPU instruction ? Here the Memory Barrier is for compiler only.

zefklop:
For those trying to figuring out what this memory barrier stuff is about:
https://gcc.godbolt.org/z/9fzGWj7W5

Adapted from a code sample from @tkreuzer

ntoskrnl/mm/ARM3/miarm.h
Extravert-ir: On x86, it doesn't do anything special except being a compiler barrier
https://godbolt.org/z/h6Yo7Y4z5

JoachimHenze:
[~zefklop] I downloaded the latest artifacts again today from https://github.com/reactos/reactos/pull/3798 after the most recent changes. It was the gcc 386 dbg build, which identified itself as
ReactOS 0.4.15-x86-dev (Build 20210805-57cd3d8) (Commit 57cd3d895a89675756d49c365c343367a047f46c)
(x) and even that did fail with the strong-name bug at 2nd try.
(x) https://github.com/reactos/reactos/pull/3798 does still not fix CORE-17595.

And CORE-17642 also has been reported to be fixed already in master head.

In sum this means that this PR does currently fix no known JIRA-ticket!
I don't want to say, that it is wrong therefore, but we should not give any false promises or keep wrong relations if we ultimately intend to commit something from within here!

Extravert-ir (approved):

Extravert-ir:
Can't we make MMPTE or some parts of it `volatile`? That should eliminate the need in explicit compiler barriers

Extravert-ir:
I've got a problem while testing this PR inside a clang-cl build. It triggers an IRQL check in `MiLockWorkingSet`. Here is a stack trace:
```
kd> kp
 # ChildEBP RetAddr  
00 (Inline) -------- nt!MiLockWorkingSet+0x4d [C:\rosgit\ntoskrnl\mm\ARM3\miarm.h @ 1315] 
01 f2741e28 804a467a nt!MmProbeAndLockPages(struct _MDL * Mdl = 0xf30c4f88, char AccessMode = 0n0 '', _LOCK_OPERATION Operation = IoReadAccess (0))+0x826 [C:\rosgit\ntoskrnl\mm\ARM3\mdlsup.c @ 1179] 
02 f2741e88 f7a4c9f9 nt!IoBuildAsynchronousFsdRequest(unsigned long MajorFunction = 3, struct _DEVICE_OBJECT * DeviceObject = 0xf2f52a08 Device for "\Driver\UniATA", void * Buffer = 0xf3072fe8, unsigned long Length = 0x12, union _LARGE_INTEGER * StartingOffset = 0xf2741eb0 {0x1}, struct _IO_STATUS_BLOCK * IoStatusBlock = 0x00000000)+0x25a [C:\rosgit\ntoskrnl\io\iomgr\irp.c @ 824] 
03 f2741ef0 f7a4becd scsiport!SpiSendRequestSense(struct _SCSI_PORT_LUN_EXTENSION * LunExtension = 0xf2f52ac0, struct _SCSI_REQUEST_BLOCK * InitialSrb = 0xf2f74fc0)+0xf9 [C:\rosgit\drivers\storage\port\scsiport\scsi.c @ 555] 
04 f2741f30 f7a4b44e scsiport!SpiProcessCompletedRequest(struct _SCSI_PORT_DEVICE_EXTENSION * DeviceExtension = 0xf2d0cb30, struct _SCSI_REQUEST_BLOCK_INFO * SrbInfo = 0x00000000, unsigned char * NeedToCallStartIo = 0xf2741f57 "")+0x89d [C:\rosgit\drivers\storage\port\scsiport\scsi.c @ 951] 
05 f2741f80 804d0bf1 scsiport!ScsiPortDpcForIsr(struct _KDPC * Dpc = 0xf2d0caec, struct _DEVICE_OBJECT * DpcDeviceObject = 0xf2d0ca78 Device for "\Driver\UniATA", struct _IRP * DpcIrp = 0x00000000, void * DpcContext = 0xf2d0cb30)+0x32e [C:\rosgit\drivers\storage\port\scsiport\scsi.c @ 1381] 
06 f2741ff4 80645e7f nt!KiRetireDpcList(struct _KPRCB * Prcb = 0xffdff120)+0x1c1 [C:\rosgit\ntoskrnl\ke\dpc.c @ 633] 
07 f2741ff8 f26f5740 nt!KiRetireDpcListInDpcStack+0xa
```

So we're in scsiport's `DpcForIsr` handler, so the IRQL is DISPATCH_LEVEL. It calls `IoBuildAsynchronousFsdRequest`, which calls `MmProbeAndLockPages` with a MDL address (which itself is nonpaged), so the usage of `MmProbeAndLockPages` at DISPATCH_LEVEL should be fine at this case.
What's wrong then, and why it is not triggered on MSVC?

Extravert-ir:
Ok, I've got a bit further. Inside `MmProbeAndLockPages`, this code path goes wrong way:
```
/* Check how we should lock */
if (MI_IS_SESSION_ADDRESS(Base))
{
    WorkingSet = &MmSessionSpace->GlobalVirtualAddress->Vm;
}
else if (MI_IS_NON_PAGED_POOL_ADDRESS(Base))
{
    UsePfnLock = TRUE;
    OldIrql = MiAcquirePfnLock();
}
else
{
    WorkingSet = &MmSystemCacheWs;
}
```

In this case, `Base` address is the one which comes from `IoAllocateMdl`. It was a small allocation, so it was taken from a lookaside buffer (`LookasideMdlList`):
https://github.com/reactos/reactos/blob/911fc3cf5b06610167c8729116131d9699fb80d2/ntoskrnl/io/iomgr/iomdl.c#L53-L54

`MI_IS_NON_PAGED_POOL_ADDRESS(Base)` returns `FALSE` for such address, looks like this is what's wrong here

Extravert-ir (changes requested):
Ok, I've got a bit further. Inside `MmProbeAndLockPages`, this code path goes wrong way:
```
/* Check how we should lock */
if (MI_IS_SESSION_ADDRESS(Base))
{
    WorkingSet = &MmSessionSpace->GlobalVirtualAddress->Vm;
}
else if (MI_IS_NON_PAGED_POOL_ADDRESS(Base))
{
    UsePfnLock = TRUE;
    OldIrql = MiAcquirePfnLock();
}
else
{
    WorkingSet = &MmSystemCacheWs;
}
```

In this case, `Base` address is the one which comes from `IoAllocateMdl`. It was a small allocation, so it was taken from a lookaside buffer (`LookasideMdlList`):
https://github.com/reactos/reactos/blob/911fc3cf5b06610167c8729116131d9699fb80d2/ntoskrnl/io/iomgr/iomdl.c#L53-L54

`MI_IS_NON_PAGED_POOL_ADDRESS(Base)` returns `FALSE` for such address, looks like this is what's wrong here

zefklop:
> Ok, I've got a bit further. Inside `MmProbeAndLockPages`, this code path goes wrong way:
> 
> ```
> /* Check how we should lock */
> if (MI_IS_SESSION_ADDRESS(Base))
> {
>     WorkingSet = &MmSessionSpace->GlobalVirtualAddress->Vm;
> }
> else if (MI_IS_NON_PAGED_POOL_ADDRESS(Base))
> {
>     UsePfnLock = TRUE;
>     OldIrql = MiAcquirePfnLock();
> }
> else
> {
>     WorkingSet = &MmSystemCacheWs;
> }
> ```
> 
> In this case, `Base` address is the one which comes from `IoAllocateMdl`. It was a small allocation, so it was taken from a lookaside buffer (`LookasideMdlList`):
> 
> https://github.com/reactos/reactos/blob/911fc3cf5b06610167c8729116131d9699fb80d2/ntoskrnl/io/iomgr/iomdl.c#L53-L54

Why would one probe&lock the MDL ? Shouldn't that be the buffer ?

> 
> `MI_IS_NON_PAGED_POOL_ADDRESS(Base)` returns `FALSE` for such address, looks like this is what's wrong here
> The address in my case is `0xf2e95fe8` which (according to table) belongs to `0xEB000000 - 0xF7BE0000 System PTE Space`

Thanks for the analysis. Indeed, this should take PTE space into account, I'll see how to correct this.

Extravert-ir:
Sorry, I was wrong at some things, don't know how had I overlooked that :)

It is reproducible with both MSVC and Clang when __Special Pool is enabled__. Sorry for a confusion. (I sometimes enable it without taking much attention). I guess for GCC it will trigger that too (can't check right now).

> Base is the buffer passed to IoBuildAsynchronousFsdRequest. Why should that be in system PTE space ?

I've found the allocation actually, it comes from cdrom. Nothing interesting:
https://github.com/reactos/reactos/blob/db8dd3b35e83cb797547a7f2499cebe9ac0fc54b/drivers/storage/class/cdrom/scratch.c#L261-L263

What's interesting to me is that according to log, special pool has this range:
```
(ntoskrnl\mm\ARM3\special.c:196) Special pool start F272B000 - end F2B2A000
```
But the address allocated is higher than that area - `0xf300dfe8`. Is it a mistake in DPRINT?

binarymaster:
The branch [`zefklop:CORE-17595`](https://github.com/zefklop/reactos/tree/CORE-17595) got synced with ROS master, but without PR commits applied on top, and this led to PR close... perhaps by accident? :eyes:

PR #3801, opened 2021-07-05
[CMAKE] msvc: Properly fix condition to use /we4189

SergeGautherie:
## Purpose

Addendum to d10728a.
JIRA issue: [CORE-17656](https://jira.reactos.org/browse/CORE-17656)

## Proposed changes

Use `<$CONFIG:...>` generator expression.


learn-more:
This is not the same

PR #3807, opened 2021-07-08
[VIDEOPRT] Fix IRP_MN_FILTER_RESOURCE_REQUIREMENTS case

SergeGautherie:
## Purpose

Addendum to 25eae4b (r54237).
JIRA issue: [CORE-17651](https://jira.reactos.org/browse/CORE-17651) (Point 1)

## Proposed changes

Return requirements, or free them.

### DIFF
--- a/win32ss/drivers/videoprt/dispatch.c
+++ b/win32ss/drivers/videoprt/dispatch.c
@@ -1029,7 +1029,6 @@ IntVideoPortDispatchFdoPnp(
             if (NT_SUCCESS(Status) && NT_SUCCESS(Irp->IoStatus.Status))
                 Status = IntVideoPortFilterResourceRequirements(DeviceObject, Irp);
             Irp->IoStatus.Status = Status;
-            Irp->IoStatus.Information = 0;
             IoCompleteRequest(Irp, IO_NO_INCREMENT);
             break;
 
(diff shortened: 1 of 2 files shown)


win32ss/drivers/videoprt/dispatch.c
```diff
@@ -1029,7 +1029,11 @@ IntVideoPortDispatchFdoPnp(
             if (NT_SUCCESS(Status) && NT_SUCCESS(Irp->IoStatus.Status))
                 Status = IntVideoPortFilterResourceRequirements(DeviceObject, Irp);
             Irp->IoStatus.Status = Status;
-            Irp->IoStatus.Information = 0;
+            if (!NT_SUCCESS(Status) && Irp->IoStatus.Information != 0)
```
tkreuzer, line 1032: This shouldn't be done here, but in whatever function has allocated it.

tkreuzer (changes requested):
Please move cleanup to where the allocation happened.

win32ss/drivers/videoprt/dispatch.c
SergeGautherie: Moved.

win32ss/drivers/videoprt/resource.c
```diff
@@ -93,7 +93,14 @@ IntVideoPortFilterResourceRequirements(
 
     Status = IntVideoPortGetLegacyResources(DriverExtension, DeviceExtension, &AccessRanges, &AccessRangeCount);
     if (!NT_SUCCESS(Status))
+    {
```
Extravert-ir, line 96: I feel like we shouldn't touch `OldResList` at all on failure

Extravert-ir (changes requested):

Extravert-ir:
Fixed upstream

PR #3808, opened 2021-07-08
[GDI32] Revise IntTMWFixUp Raster Font List

Doug-Lyons:
## Update IntTMWFixUp Raster Font List

_This list had fonts listed that ReactOS does not support and was missing fonts._

JIRA issues: [CORE-17662](https://jira.reactos.org/browse/CORE-17662) and [CORE-17723](https://jira.reactos.org/browse/CORE-17723)

## Add Two Fonts and Delete Two Fonts

_This fixes the text shown in the toolbar for Pidgin and Geany when using the Lunar theme ._
While we are editing this file, we will alphabetize the font names as well.

Here is my test program for proving that CORE-1091 is still fixed:

Before first patch:

VB6_Install_Bad

After 1st patch and current:

VB6_Install_OK

### DIFF
--- a/win32ss/gdi/ntgdi/text.c
+++ b/win32ss/gdi/ntgdi/text.c
@@ -42,14 +42,13 @@ IntTMWFixUp(
      * out the problematic TrueType and Vector bits.
      * Our list below checks for Raster Font Facenames. */
     DPRINT("Font Facename is '%S'.\n", lf.lfFaceName);
-    if ((wcsicmp(lf.lfFaceName, L"Helv") == 0) ||
-        (wcsicmp(lf.lfFaceName, L"Courier") == 0) ||
+    if ((wcsicmp(lf.lfFaceName, L"Courier") == 0) ||
+        (wcsicmp(lf.lfFaceName, L"FixedSys") == 0) ||
+        (wcsicmp(lf.lfFaceName, L"Helv") == 0) ||
         (wcsicmp(lf.lfFaceName, L"MS Sans Serif") == 0) ||
-        (wcsicmp(lf.lfFaceName, L"MS Serif") == 0) ||
-        (wcsicmp(lf.lfFaceName, L"Times New Roman") == 0) ||
-        (wcsicmp(lf.lfFaceName, L"MS Shell Dlg") == 0) ||
         (wcsicmp(lf.lfFaceName, L"System") == 0) ||
-        (wcsicmp(lf.lfFaceName, L"Terminal") == 0))
+        (wcsicmp(lf.lfFaceName, L"Terminal") == 0) ||
+        (wcsicmp(lf.lfFaceName, L"Tms Rmn") == 0))
     {
         ptm->TextMetric.tmPitchAndFamily &= ~(TMPF_TRUETYPE | TMPF_VECTOR);
     }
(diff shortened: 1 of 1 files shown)


cernodile (approved):
LGTM.

tkreuzer:
I don't think fonts should be deleted from here. We use substitutes, so if one of the fonts is requested, that we substitute, we probably still want this. I haven't looked at the internals of this though.

Doug-Lyons:
> I don't think fonts should be deleted from here. We use substitutes, so if one of the fonts is requested, that we substitute, we probably still want this. I haven't looked at the internals of this though.

These fonts are only those that are actually raster fonts in W2K3SP2. I deleted the "MS Serif" one because ReactOS does not have this font. I will add it back to the list if you think that it likely could be used. My thoughts are that it would be a rare case where this would happen, but I will definitely defer to your judgement on this. The other two that were deleted (MS Shell Dlg and Times New Roman) was just wrong and these are definitely TrueType fonts in W2KPS2. Thanks.

JoachimHenze:
Can anyone retest https://jira.reactos.org/browse/CORE-1091 with this PR?

Doug-Lyons:
> Can anyone retest https://jira.reactos.org/browse/CORE-1091 with this PR.

@JoachimHenze, I have included example screen captures now in the header above. Thanks.

learn-more:
> I don't think fonts should be deleted from here. We use substitutes, so if one of the fonts is requested, that we substitute, we probably still want this. I haven't looked at the internals of this though.

This entire thing is a hack until we have support for raster fonts, so if a change like this improves shit, it is fine.

Doug-Lyons:
> > I don't think fonts should be deleted from here. We use substitutes, so if one of the fonts is requested, that we substitute, we probably still want this. I haven't looked at the internals of this though.
> 
> These fonts are only those that are actually raster fonts in W2K3SP2. I deleted the "MS Serif" one because ReactOS does not have this font. I will add it back to the list if you think that it likely could be used. My thoughts are that it would be a rare case where this would happen, but I will definitely defer to your judgement on this. The other two that were deleted were MS Shell Dlg and Times New Roman. Times New Roman was just wrong. This is a TrueType font in W2KPS2. The MS Shell Dlg is in the FontSubstitutes registry entry and it is mapped to "Microsoft Sans Serif" which is a TrueType font and therefore wrong. Thanks.

I have added "MS Serif" back into the raster font list based on your comment. Thanks.

Doug-Lyons:
@tkreuzer, Maybe one last review and hopefully you can commit this for us. The only items deleted were those definitely in error.

Doug-Lyons:
@katahiromz, I believe that you merged the previous version of this code in PR#2689. Can you review and help with this now? Thanks. Doug Lyons

Doug-Lyons:
@cernodile, Can you please check that this still fixes your problem after my last changes? Thanks.

cernodile:
@Doug-Lyons my approval from testing standpoint stands. Built with latest changes, works.

learn-more (approved):

Doug-Lyons:
@learn-more, Can you please help with getting this committed for me. Thanks.

learn-more:
@Doug-Lyons Thanks for your contribution!

PR #3811, opened 2021-07-10
Fix msvc release build

tkreuzer:
This PR fixes MSVC Release builds for both x86 and x64 and adds them as build types to GitHub actions.
The x64 Release build doesn't boot, the x86 one crashes in 2nd stage.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -78,6 +78,7 @@ jobs:
     strategy:
       matrix:
         arch: [i386, amd64]
+        buildtype: [Debug,Release]
         toolset: ['14.2', '14.1'] # VS 2019, 2017
         include:
         - arch: i386 # Not compiling on amd64 prompt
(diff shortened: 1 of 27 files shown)


.github/workflows/build.yml
```diff
@@ -111,7 +112,7 @@ jobs:
       with:
         path: src
     - name: Configure
-      run: cmake -S src -B build -G Ninja -DCMAKE_TOOLCHAIN_FILE:FILEPATH=toolchain-msvc.cmake -DARCH:STRING=${{matrix.arch}} -DENABLE_ROSTESTS=1 -DENABLE_ROSAPPS=1
+      run: cmake -S src -B build -G Ninja -DCMAKE_TOOLCHAIN_FILE:FILEPATH=toolchain-msvc.cmake -DARCH:STRING=${{matrix.arch}} -DENABLE_ROSTESTS=1 -DENABLE_ROSAPPS=1 -DCMAKE_BUILD_TYPE=${{matrix.buildtype}}
```
Extravert-ir, line 115: Should we do `-DENABLE_ROSTESTS=1 -DENABLE_ROSAPPS=1` for a release build?

tkreuzer: Why not?

HBelusca (approved):
Looks quite reasonable.

PR #3814, opened 2021-07-11
[NTOS:MM] Fix MiSetSystemCodeProtection

tkreuzer:
CORE-16387 CORE-17638
Windows only makes the header RO for session space mapped drivers (win32k, dx and display drivers)


HBelusca (approved):
Does Windows 10 still do this R/W protection?

tkreuzer:
> 
> 
> Does Windows 10 still do this R/W protection?

Yes, obviously. It's most likely way more strict. It even protects pages with a hypervisor and 2nd level page table (if enabled), so that you can't even mess with the PTEs.

PR #3816, opened 2021-07-12
[SHELL32] Add File Extension for BAT and CMD to auto tests for files to be executed

Doug-Lyons:
## Allow files with extensions of BAT and CMD to be executed without having to add their extensions

_Extend the file extensions tested for auto execution by adding BAT and CMD._

JIRA issue: [CORE-17612](https://jira.reactos.org/browse/CORE-17612)

## Add additional programming lines to test for BAT and CMD to automatically tested file extensions.

Also while we are changing this code, we adjust the testing order in which the extensions are evaluated.
This order should be COM, EXE, BAT, and then CMD (first to last).
This is the fallback order even if the environment variable PATHEXT does not exist.

Testbot Results:

CORE-17612 JID59850 shlexec_cpp_02.patch on top of aea4cfb

KVM:  https://reactos.org/testman/compare.php?ids=78259,78265 LGTM
 VBox: https://reactos.org/testman/compare.php?ids=78261,78263 LGTM

### DIFF
--- a/dll/win32/shell32/shlexec.cpp
+++ b/dll/win32/shell32/shlexec.cpp
@@ -28,6 +28,8 @@ WINE_DEFAULT_DEBUG_CHANNEL(exec);
 static const WCHAR wszOpen[] = L"open";
 static const WCHAR wszExe[] = L".exe";
 static const WCHAR wszCom[] = L".com";
+static const WCHAR wszBat[] = L".bat";
+static const WCHAR wszCmd[] = L".cmd";
 
 #define SEE_MASK_CLASSALL (SEE_MASK_CLASSNAME | SEE_MASK_CLASSKEY)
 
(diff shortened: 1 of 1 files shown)


katahiromz:
This ordering is seen at `shlwapi!PathFileExistsDefExtW`. Shouldn't we use `shlwapi` functions to search?

dll/win32/shell32/shlexec.cpp
```diff
@@ -28,6 +28,8 @@ WINE_DEFAULT_DEBUG_CHANNEL(exec);
 static const WCHAR wszOpen[] = L"open";
 static const WCHAR wszExe[] = L".exe";
 static const WCHAR wszCom[] = L".com";
+static const WCHAR wszBat[] = L".bat";
+static const WCHAR wszCmd[] = L".cmd";
```
HBelusca, line 32: Shouldn't all this be covered by the `PathResolveW` API @katahiromz finally implemented recently?

Doug-Lyons:
> This ordering is seen at `shlwapi!PathFileExistsDefExtW`. Shouldn't we use `shlwapi` functions to search?

This might work well. Can you propose a patch and attach it to the original issue at [CORE-17612](https://jira.reactos.org/browse/CORE-17612). Thanks,

dll/win32/shell32/shlexec.cpp
Doug-Lyons: I tried working with the PathResolveW function, but I was not able to easily create workable code myself with it. Can you attach a patch in the original issue [CORE-17612](https://jira.reactos.org/browse/CORE-17612) that we can discuss? Thanks.

HBelusca:
> This ordering is seen at `shlwapi!PathFileExistsDefExtW`. Shouldn't we use `shlwapi` functions to search?

Or PathResolve? (that calls that PathFileExistsDefExtW function internally)

Doug-Lyons:
Superseded by https://github.com/reactos/reactos/pull/4932 which has already been merged.

PR #3835, opened 2021-07-17
[CRT] Refactor cmake files

tkreuzer:
This is an attempt to make the mess a little more readable.


sdk/lib/crt/misc/misc.cmake
```diff
@@ -0,0 +1,29 @@
...
+    misc/getargs.c
+    misc/i10output.c
+    misc/initterm.c
+    misc/lock.c
+    misc/purecall.c
+    misc/stubs.c
+    misc/tls.c
+)
+
+add_library(getopt misc/getopt.c)
```
HBelusca, line 27: Wouldn't it make sense to keep this add_library into the "root" crt cmake file? Since somehow this is a separate module.

tkreuzer: Eventually I would like to replace these file collections with object libraries to avoid compiling the exact same stuff twice. That would put stuff like this into the folder where it resides in. I didn't want to do that right away, because we still use CMake defines to compille code differently between libcntpr and crt in a number of places, so this needs to be fixed first. And stuff like getopt and atan2 are IMO "hack- libraries" rather than real libraries.

HBelusca: It is sometimes necessary to recompile things "twice" or more because they get compiled with different defines enabled, e.g. the win32 CRT versus the NT user-mode (or even kernel-mode) CRT.

tkreuzer: I agree, but in those cases it might be better to have 2 different source files. I.e. "foobar.c" and "foobar_nt.c", where the latter adds some defines and includes he former. If we stuck to this principle, we wouldn't recompile plenty of code without any changes. It add a little bit of required thought, but helps reducing compile time. I already added this principle to hal, so we wouldn't recompile the entire hal for every different configuration. I would eventuallly like to add it to the kernel as well. When 99& of the code i exactly the same, it's better to have a few extra wrapper files that include other files, so you don't have to recompile the whole thing.

tkreuzer:
Yo, I merged, LOL!

PR #1987, opened 2019-10-21
[MSI_WINETEST] Sync with Wine Staging 4.18. CORE-16441

AmineKhaldi:

PR #3836, opened 2021-07-19
[MSI] Fix MS Office 2000/2003 Installs by Wine-6.8 Sync of MSI_OpenPackageW

Doug-Lyons:
## Fix MS Office 2000/2003 Install Failures by updating MSI_OpenPackageW

_Partial Wine-6.8 sync of package.c for MSI_OpenPackageW._

JIRA issue: [CORE-17693](https://jira.reactos.org/browse/CORE-17693)

### DIFF
--- a/dll/win32/msi/package.c
+++ b/dll/win32/msi/package.c
@@ -1498,6 +1498,8 @@ UINT MSI_OpenPackageW(LPCWSTR szPackage, MSIPACKAGE **pPackage)
         r = get_local_package( file, localfile );
         if (r != ERROR_SUCCESS || GetFileAttributesW( localfile ) == INVALID_FILE_ATTRIBUTES)
         {
+            DWORD localfile_attr;
+
             r = msi_create_empty_local_file( localfile, dotmsi );
             if (r != ERROR_SUCCESS)
             {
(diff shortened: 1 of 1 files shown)


Extravert-ir:
I don't quite like picking up individual files, that puts our Wine-synced packages into an uncertain state. How about syncing the whole MSI package?
There are still #1986 and #1987 pull requests laying around (ofc we shouldn't sync with 4.18 these days, but with the latest one)

Doug-Lyons:
> I don't quite like picking up individual files, that puts our Wine-synced packages into an uncertain state. How about syncing the whole MSI package?
> There are still #1986 and #1987 pull requests laying around (ofc we shouldn't sync with 4.18 these days, but with the latest one)

@Extravert-ir, I did not even sync this function because its parameter count was changed from two to three and this would require touching lots of code. My intent was just to fix what we know is broken and do this with minimum changes to our existing code. I am not willing to get involved in such a large project as syncing a whole section of code.
This fixes what was broken in recent code changes and makes this work again in a reasonable way I believe. Thanks.

Getequ:
@Extravert-ir  "ofc we shouldn't sync with 4.18 these days, but with the latest one" - having mix of many versions not a good idea i think (because of possible compatibility issues). better to update all possible to 4.18 and after that move forward

learn-more (approved):
In general I would would say a complete winesync is preferred,
but in this case it will fix (a lot of?) setups when starting them from an .iso

HBelusca:
Do we know approximately which wine commit(s) these new code lines correspond to?

Doug-Lyons:
> Do we know approximately which wine commit(s) these new code lines correspond to?

Yes. https://source.winehq.org/git/wine.git/commit/e830975806df9ef283c89f7153b06b145dfb2cec?f=dlls/msi/package.c

HBelusca:
> > Do we know approximately which wine commit(s) these new code lines correspond to?
> 
> Yes. [...] 

Then I would strongly suggest you tell in the commit log that you're precisely importing this commit.

Doug-Lyons:
@HBelusca, Sorry I do not know how to add a statement about this wine commit in the commit log. Maybe this could be added when this is committed. Thanks.

Doug-Lyons:
Thanks @learn-more and @binarymaster for your help in getting this merged.

PR #3837, opened 2021-07-19
Remove FFS/UFS and ReiserFS driver from the tree

Extravert-ir:
These file systems are almost not seen in real life, upstream projects are not maintained.
And anybody will barely test them except basic work.
We better focus on UDFS, NTFS, BTRFS and EXT2/3/4 development as these file systems are actually used in real life.


HBelusca:
For reminders, please add a comment in https://jira.reactos.org/browse/CORE-11040 and https://jira.reactos.org/browse/CORE-11005 that these FSes are being removed, once this PR is merged in master.

HBelusca (approved):

binarymaster (approved):

JoachimHenze:
Foremost I do *not* oppose these removals (as Windows does also not natively support those FS'es), but since you destroy the work of someone else here (Pierres):
I still want to say that BTRFS is actually a much bigger burden for ros than reiser and FFS ever were.
And gust ftr: reiserfs is our only partly-journaling FS atm.

I would like to get to know your motivation for performing that step right now extravert-ir, are these 2FS'es holding you back somehow with some newer VisualStudio? Are they broken in any way? Are they incompatible to the new storage stack, that was introduced during 0.4.15dev'ing?

Extravert-ir:
I don't even know if they work right now, nor I am going to check that.
This is just dead code and that is the main reason why I want to remove these file systems. Not connected to any other change.

> BTRFS is actually a much bigger burden for ros than reiser and FFS ever were

BTRFS is can be used in ReactOS, ReiserFS and FFS support is "nominal": we can't even format a partition

Extravert-ir:
Kept fs_rec support for these file systems

PR #3825, opened 2021-07-14
[GITHUB] build.yml: Separate Debug/Release artifacts, Add 2 i386-Release builds

SergeGautherie:
## Purpose

Addendum to a9a22ae.

## Proposed changes

- Fix Dbg/Rel artifact handling.
- Complete supported i386-Release build list.

PR #3838, opened 2021-07-21
[GITHUB] Build Release builds for all compilers

Extravert-ir:


SergeGautherie:
What about #3825?

learn-more:
> What about #3825?

What about it?


> What about #3825?

What about it?

tkreuzer:
Unrelated to this change, but I noticed it's uploading ISOs. Why would we ever want to have ISOs from PRs?

tkreuzer (approved):

zefklop (approved):

Extravert-ir:
> Unrelated to this change, but I noticed it's uploading ISOs. Why would we ever want to have ISOs from PRs?

Sometimes it may be useful for testing. But generally, I don't disable it because it costs nothing

sdk/cmake/msvc.cmake
```diff
@@ -279,6 +283,12 @@ function(set_module_type_toolchain MODULE TYPE)
             # Mark .rsrc section as non-disposable non-pageable, as bugcheck code needs to access it
             target_link_options(${MODULE} PRIVATE /SECTION:.rsrc,!DP)
         endif()
+
+        # Clang may insert the call during some optimizations,
+        # but i386 ntoskrnl doesn't export memcmp (before NT 6.2)
+        if(CMAKE_C_COMPILER_ID STREQUAL "Clang" AND ARCH STREQUAL "i386")
+            target_compile_options(${MODULE} PRIVATE -fno-builtin-memcmp)
+        endif()
```
zefklop, line 291: https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/CMakeLists.txt;h=6849bcf009987f137755d9a2fe26df69374d6ac0;hb=HEAD#l33

Both are fine, but we should be consistent.

sdk/lib/crt/msvcrtex.cmake
```diff
@@ -43,19 +43,24 @@ else()
...
-    endif()
+endif()
+
+if(CMAKE_C_COMPILER_ID STREQUAL "Clang")
+    # Clang performs some optimizations requiring those funtions
+    list(APPEND MSVCRTEX_SOURCE
+        math/exp2.c
+        math/exp2f.c
+        )
 endif()
```
zefklop, line 54: :+1: for CMAKE_C_COMPILER_ID

tkreuzer (approved):

SergeGautherie:
> > What about #3825?
> 
> What about it?

Why ignore/hijack it?

PR #3839, opened 2021-07-22
[VIDEOPRT] Fix updating of new registry path values

tkreuzer:
JIRA issue: [CORE-17688](https://jira.reactos.org/browse/CORE-17688)

When a new driver is installed for the same device (like VBoxVideo), it uses the same hardware enum registry key and thus reuses the same DisplayId and the same display registry key. Therefore we need to update the setting in that key, even when the key already exists.

This seems to work good and not cause any issues, but testing indicated that on Windows some values are only updated, when the driver has changed. If neccessary, this can be achieved by updating and querying the ActiveService value in the device enum key (e.g. HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Enum\PCI\VEN_80EE&DEV_BEEF&SUSYS_00000000&REV_00\3&267a616a&0&10\Control: ActiveService). If that doesn't match the current device name (from DriverExtension->RegistryPath) the values should be copied over.

And also fixes an unrelated uninitialized variable introduced 3 weeks ago, which makes MSVC builds constantly break, but nobody seems to care about.

### DIFF
--- a/win32ss/drivers/videoprt/registry.c
+++ b/win32ss/drivers/videoprt/registry.c
@@ -498,41 +498,41 @@ IntCreateNewRegistryPath(
             ERR_(VIDEOPRT, "Failed create key '%wZ'\n", &DeviceExtension->NewRegistryPath);
             return Status;
         }
-
-        /* Open the new key */
-        InitializeObjectAttributes(&ObjectAttributes,
-                                   &DeviceExtension->NewRegistryPath,
-                                   OBJ_KERNEL_HANDLE | OBJ_CASE_INSENSITIVE,
-                                   NULL,
-                                   NULL);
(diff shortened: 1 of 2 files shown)


HBelusca (approved):
lgtm.
I wonder whether this helps also fixing the problem happening when one uninstalls the VBox additions (and therefore, the VBox display driver)

binarymaster (approved):

tkreuzer:
> 
> 
> lgtm.
> I wonder whether this helps also fixing the problem happening when one uninstalls the VBox additions (and therefore, the VBox display driver)

Uninstalling leads to an Mm ASSERT. https://jira.reactos.org/browse/CORE-17698 That's Jerome's turf.

PR #3840, opened 2021-07-23
[COMCTL32] Allow only one instance to 'Properties' dialog boxes

zecarlos1957:
[COMCTL32] Allow only one instance to 'Properties' dialog boxes

Expected behavior:
  As in Windows, only one instance of the 'Properties' dialogs box is possible.

Because psInfo->strPropertiesFor is only affected when creating the dialog in PROPSHEET_CreateDialog and we need this data before this stage, we implemented the helper function ExtractPropSheetStr(PropSheetInfo* psInfo) to get the Title "Properties for %s' from the IDD_PROPSHEET DLGTEMPLATE structure.

### DIFF
--- a/dll/win32/comctl32/propsheet.c
+++ b/dll/win32/comctl32/propsheet.c
@@ -2822,6 +2822,56 @@ static INT do_loop(const PropSheetInfo *psInfo)
     return ret;
 }
 
+#ifdef __REACTOS__
+static LPWSTR load_string( HINSTANCE instance, LPCWSTR str );
+
+VOID ExtractPropSheetStr(PropSheetInfo* psInfo)
+{
+    const DLGTEMPLATE* pTemplate;
+    const WORD*  p;
+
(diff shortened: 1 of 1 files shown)


binarymaster (commented):
Is this a new feature or a bug fix? (is this also behavior on Windows?) It would be nice so see tests results comparison as well.

zecarlos1957:
As in Windows, only one instance of the 'Properties' dialogs box is possible.

learn-more (changes requested):
I very much doubt that comctl32 is blocking a duplicate property sheet like that.
If anything, I'd say that the shell itself is managing this.

image

zecarlos1957:
@learn-more  Seems to be a bigger problem than me.
Any complaints if I delete this PR?

learn-more:
If you think a PR is no longer needed, it is fine to close it

PR #3844, opened 2021-07-24
[CRT] Fix _matherror and _setusermatherror

tkreuzer:
We previously used 2 different versions: one from wine and one from mingw-w64.
The former was used in msvcrt, the latter was statically compiled into the executable.

When using MS libs, there is only one _matherr, which is statically linked into the executable and does nothing (it's not really a function for users to be called).
_setusermatherror should only be exist in msvcrt and not statically, which wouldn't work at all.

I just ripped the old stuff out and replaced it completely, so nobody gets the great idea to sync the stuff back from wine/mingw-w64.
Note that the code isn't used yet, but I intend to change that with importing some math functions from amd libm.


binarymaster (approved):

sdk/lib/crt/math/usermatherr.c
```diff
@@ -0,0 +1,34 @@
...
+ * PURPOSE:     Implementation of __setusermatherr and _invoke_user_matherr
+ * COPYRIGHT:   Copyright 2021 Timo Kreuzer <timo.kreuzer@reactos.org>
+ */
+
+#include <math.h>
+
+/* MS headers have this in corecrt_startup.h */
+typedef int (*_UserMathErrorFunctionPointer)(struct _exception *);
+
+_UserMathErrorFunctionPointer user_matherr = 0;;
```
HBelusca, line 13:
```suggestion
_UserMathErrorFunctionPointer user_matherr = NULL;
```

sdk/lib/crt/math/usermatherr.c
```diff
@@ -0,0 +1,34 @@
...
+__setusermatherr(_UserMathErrorFunctionPointer func)
+{
+    user_matherr = func;
+}
+
+int
+__cdecl
+_invoke_user_matherr(struct _exception *e)
+{
+    if (user_matherr != 0)
```
HBelusca, line 26:
```suggestion
    if (user_matherr != NULL)
```

ColinFinck (approved):
If the MS version of `_matherr` indeed does nothing while mingw-w64 and WINE have an implementation, this seems to be the cleanest solution.

However, I'm afraid a few years from now, someone will look into mingw-w64/WINE anyway, see that they have an implementation where we only have a stub, and open a PR to change that. Watch closely and possibly add even more comments into your code to emphasize that this must not be synced.

sdk/lib/crt/math/usermatherr.c
zefklop: `static` please

tkreuzer: Fixed

sdk/lib/crt/math/usermatherr.c
tkreuzer: Fixed

sdk/lib/crt/math/usermatherr.c
```diff
@@ -0,0 +1,36 @@
...
+void
+__cdecl
+__setusermatherr(_UserMathErrorFunctionPointer func)
+{
+    user_matherr = func;
+}
+
+int
+__cdecl
+_invoke_user_matherr(struct _exception *e)
```
zefklop, line 26: When is this one used ?

tkreuzer: When math functions encounter an error like division by zero. We don't use it yet, but that will change when I import some math functions from AMD's libm.

zefklop (approved):

PR #3849, opened 2021-07-26
[ROSTESTS] Implement white-listing feature

tkreuzer:
## Purpose

Add a feature to allow white-listing tests.

## Proposed changes

This adds a tool that generates tables for the test code to check for white-listed tests. The tables are generated from a text whitelist, where each entry contains the hash of the original commit, in which the test was added, the file name, the line and a reason. To get the data use
"git blame -L <line>,<line> --show-number <file name>"
The tool uses git to translate this back to the current line, even when the file is changed.

## TODO
- [ ] Detect when a line gets deleted
- [ ] Add functionality to add entries with the tool


sdk/tools/whitelister/whitelister.cpp
```diff
@@ -0,0 +1,301 @@
+
```
HBelusca, line 1: Standard source file header block? https://reactos.org/wiki/Coding_Style#File_Structure

sdk/cmake/host-tools.cmake
```diff
@@ -118,3 +118,17 @@ function(setup_host_tools)
         add_dependencies(native-${_module} host-tools ${INSTALL_DIR}/bin/${HOST_EXTRA_DIR}${_module}${HOST_MODULE_SUFFIX})
     endforeach()
 endfunction()
+
+function(generate_whitelist_code _whitelist_file)
+
```
HBelusca, line 123:
```suggestion
```

sdk/cmake/host-tools.cmake
```diff
@@ -118,3 +118,17 @@ function(setup_host_tools)
...
+    # Error out on anything else than txt
+    if(NOT ${_whitelist_file} MATCHES ".*\\.txt")
```
HBelusca, line 125: is the match string a regex? or some windows-like expression? (looks strange at first sight, I would have expected something like `"*.txt"`)

sdk/include/reactos/wine/test.h
```diff
@@ -213,6 +213,26 @@ extern void __winetest_cdecl winetest_print(const char* msg, ...);
 # define S5(x) (x)
 #endif
 
+enum WHITELIST_ATTRIBUTES
```
HBelusca, line 216:
```suggestion
#ifdef USE_WHITELIST

enum WHITELIST_ATTRIBUTES
```

sdk/include/reactos/wine/test.h
```diff
@@ -213,6 +213,26 @@ extern void __winetest_cdecl winetest_print(const char* msg, ...);
...
+    int Line;
+    const char* Reason;
+} WHITELIST_ENTRY, *PWHITELIST_ENTRY;
+
+typedef struct _WHITELIST_FILE_ENTRY
+{
+    const char* File;
+    PWHITELIST_ENTRY Entries;
+    int NumEntries;
+} WHITELIST_FILE_ENTRY, *PWHITELIST_FILE_ENTRY;
```
HBelusca, line 235:
```suggestion
} WHITELIST_FILE_ENTRY, *PWHITELIST_FILE_ENTRY;

#endif // USE_WHITELIST
```

sdk/tools/whitelister/whitelister.cpp
```diff
@@ -0,0 +1,301 @@
...
+    // Get the reason
+    start += length + 1;
+    length = Line.length() - start - 2;
+    string reason = Line.substr(start + 1, length);
+
+    // Translate to head
+    string translated = git_rev_blame(oldHash, oldFile, oldLine);
+
+    fprintf(stderr, "Translated: \"%s\"\n", translated.c_str());
```
HBelusca, line 126: using `cerr` ? (and in all other places where `fprintf(stderr` is used)

sdk/tools/whitelister/whitelister.cpp
```diff
@@ -0,0 +1,301 @@
...
+    {
+        Usage();
+        return 1;
+    }
+
+    if (string(argv[1]) == "--add-entry")
+    {
+        if (argc != 5)
+        {
+            printf("Usage: whitelister --add-entry <whitelist-file> <source-file> <line>\n");
```
HBelusca, line 269: use `cout` ?

sdk/tools/whitelister/whitelister.cpp
```diff
@@ -0,0 +1,301 @@
...
+    string code = GenerateWhitelistCode(whitelist);
+    ofstream outputFile(OutputFileName);
+    outputFile << code;
+    return 0;
+}
+
+void Usage(void)
+{
+    printf("Usage: whitelister --add-entry <whitelist-file> <source-file> <line>\n");
+    printf("Usage: whitelister --gen-tables <whitelist-file> <output-file> <reactos source root>\n");
```
HBelusca, line 250:
```suggestion
    cout << "Usage: whitelister --add-entry <whitelist-file> <source-file> <line>\n"
         << "       whitelister --gen-tables <whitelist-file> <output-file> <reactos source root>\n";
```

sdk/tools/whitelister/whitelister.cpp
```diff
@@ -0,0 +1,301 @@
...
+    {
+        if (newFileName == wlEntry.FullFileName)
+        {
+            wlEntry.Lines.push_back(WL_LINE_ENTRY{ newLine, reason });
+            fprintf(stderr, "!! size of found entry = %Iu\n", wlEntry.Lines.size());
+            return;
+        }
+    }
+
+    // Get thhe short file name
```
HBelusca, line 153:
```suggestion
    // Get the short file name
```

sdk/tools/whitelister/whitelister.cpp
```diff
@@ -0,0 +1,301 @@
...
+{
+    return ExecuteCmd("git blame -L %u,%u --show-number --show-name %s",
+        Line, Line, File.c_str());
+}
+
+string
+git_rev_blame(string &OldHash, string &OldFile, int OldLine)
+{
+    return ExecuteCmd("git blame -L %u,%u -s --reverse -s --show-number %s %s",
+        OldLine, OldLine, OldHash.c_str(), OldFile.c_str());
```
HBelusca, line 87:
```suggestion
    return ExecuteCmd("git blame -L %u,%u -s --reverse -s --show-number %s %s",
                      OldLine, OldLine, OldHash.c_str(), OldFile.c_str());
```

sdk/tools/whitelister/whitelister.cpp
```diff
@@ -0,0 +1,301 @@
...
+    va_end(argptr);
+
+    return Execute(buffer.data());
+}
+
+string
+git_blame(string &File, int Line)
+{
+    return ExecuteCmd("git blame -L %u,%u --show-number --show-name %s",
+        Line, Line, File.c_str());
```
HBelusca, line 80:
```suggestion
    return ExecuteCmd("git blame -L %u,%u --show-number --show-name %s",
                      Line, Line, File.c_str());
```

sdk/cmake/host-tools.cmake
```diff
@@ -118,3 +118,17 @@ function(setup_host_tools)
         add_dependencies(native-${_module} host-tools ${INSTALL_DIR}/bin/${HOST_EXTRA_DIR}${_module}${HOST_MODULE_SUFFIX})
     endforeach()
 endfunction()
+
+function(generate_whitelist_code _whitelist_file)
```
zefklop, line 122: This belongs to `modules/rostests/CMakeLists.txt`

sdk/include/reactos/wine/test.h
```diff
@@ -354,6 +374,48 @@ int broken( int condition )
...
+            {
+                if (entries[j].Line == Line)
+                {
+                    return &entries[j];
+                }
+            }
+        }
+    }
+
+    return 0;
```
zefklop, line 400: `return NULL`

sdk/include/reactos/wine/test.h
```diff
@@ -213,6 +213,26 @@ extern void __winetest_cdecl winetest_print(const char* msg, ...);
 # define S5(x) (x)
 #endif
 
+enum WHITELIST_ATTRIBUTES
+{
+    WLA_OK = 0,
+    WLA_BROKEN_WIN = 1,
+    WLA_BROKEN_ROS = 2,
+    WLA_FLAKY = 4
+};
```
zefklop, line 222: How do you intend to use this ?

zefklop:
Looks fine so far. I am eager to see what the non-draft version will look like

sdk/cmake/host-tools.cmake
learn-more: Look at the cmake documentation?

sdk/tools/whitelister/whitelister.cpp
learn-more: Why?

sdk/tools/whitelister/whitelister.cpp
zefklop: iostream << operator performs so badly that it is almost always a bad idea to use it.
https://stackoverflow.com/questions/2872543/printf-vs-cout-in-c

sdk/cmake/host-tools.cmake
tkreuzer: In fact I think it needs another `` \``

sdk/include/reactos/wine/test.h
tkreuzer: These can go away anyway, they are a remainder from when I did things more complicated.

sdk/tools/whitelister/whitelister.cpp
tkreuzer: Yeah, I avoid it where I can as well. Only used it in one place for writing to a file.

HBelusca: thanks for the info.

sdk/cmake/host-tools.cmake
tkreuzer: IMO everything in here should go there.

sdk/include/reactos/wine/test.h
tkreuzer: I wanted to specify different use cases, to be able to selectively disable the test on ReactOS only. I thought about only using it for ReactOS and instead rather get the tests fixed or commented out, that fail on Windows, but then there are still a bunch of wine-tests, that fail on Windows. So I will probably add some kind of "annotation" that specifies where it fails on, like `` <...> "Test fails on #ROS and #WIN2003``  or something. Suggestions how to do it are welcome.

sdk/cmake/host-tools.cmake
zefklop: You have to be more specific on what "here" and "there" means for me to understand this last sentence :sweat_smile:

tkreuzer: Everything in this file should go to modules/rostests/CMakeLists.txt

sdk/include/reactos/wine/test.h
tkreuzer: Fixed

sdk/tools/whitelister/whitelister.cpp
tkreuzer: Fixed

sdk/tools/whitelister/whitelister.cpp
tkreuzer: Be careful what you wish for, or I slap a bunch `` nullptr``  into it.

sdk/include/reactos/wine/test.h
```diff
@@ -354,6 +382,73 @@ int broken( int condition )
...
+}
+
+static int does_whitelist_entry_apply(PWHITELIST_ENTRY entry)
+{
+    if ((entry->Attributes & WLA_BROKEN_ROS) && (strcmp(winetest_platform, "reactos") == 0))
+    {
+        return 1;
+    }
+
+    unsigned int winver = GetVersion();
```
learn-more, line 418: This will not get above 0x602 without a manifest, so better to use `RtlGetVersion`

image

tkreuzer:
I realized that the git operations take a lot of time. advapi32_winetest alone takes about 5 seconds. If we did this for all failures, we would probably add 20 minutes compile time. so I'm going to change things a bit. Instead of generating the tables at compile time, they will be generated inside the source tree with a special target, that is not part of "ALL". To avoid breaking stuff, I will add a check, whether the head of each file or maybe the folder has changed, and then require rebuild of the whitelist tables.

learn-more:
> I realized that the git operations take a lot of time. advapi32_winetest alone takes about 5 seconds. If we did this for all failures, we would probably add 20 minutes compile time. so I'm going to change things a bit. Instead of generating the tables at compile time, they will be generated inside the source tree with a special target, that is not part of "ALL". To avoid breaking stuff, I will add a check, whether the head of each file or maybe the folder has changed, and then require rebuild of the whitelist tables.

Maybe we could add a post-merge script that checks if there are changes in `modules\rostests` and updates / commits after that?

tkreuzer:
> 
> Maybe we could add a post-merge script that checks if there are changes in `modules\rostests` and updates / commits after that?

I thought about a cmake target, that does this, but is not compiled by default. Or are you thinking of some special CMake mechanism?

zefklop:
> > Maybe we could add a post-merge script that checks if there are changes in `modules\rostests` and updates / commits after that?
> 
> I thought about a cmake target, that does this, but is not compiled by default. Or are you thinking of some special CMake mechanism?

git hooks https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks look like what we need

sdk/tools/whitelister/whitelister.cpp
```diff
@@ -0,0 +1,411 @@
...
+AddWhitelistEntry(vector<WL_ENTRY> &Whitelist, string& Line, string& HeadHash, const string& WhiteListFileName)
+{
+    size_t start, length;
+
+    // Format:
+    // <Hash> <original file name> <original line> "<reason>"
+    // 8400fdc0786b rostests/apitests/advapi32/LockDatabase.c 701 "Fails on WHS"
+
+    // Start with the hash, which is 12 characters
+    string oldHash = Line.substr(0, 12);
```
zefklop, line 187: This might change if commit history grows, right ?

sdk/tools/whitelister/whitelister.cpp
```diff
@@ -0,0 +1,411 @@
...
+
+    // Check if the file has local changes
+    if (git_diff_check(newFileName))
+    {
+        string message = "ERROR: File with white-list entries '" + newFileName;
+        message += "' Has uncommitted changes in the working tree.\n";
+        message += "Please commit your changes first!";
+        fprintf(stderr, "%s", message.c_str());
+        throw exception();
+    }
```
zefklop, line 248: Does this mean that you have to commit before build each time you change something in rostests ?

sdk/tools/whitelister/whitelister.cpp
```diff
@@ -0,0 +1,411 @@
...
+}
+
+int
+AddWhitelistLine(const string &WhitelistFileName, const string &SourceFileName, int Line)
+{
+    return 0;
+}
+
+uint32_t
+GetWhiteListAttributesFromReason(string& Reason)
```
zefklop, line 127: how about comma-separated values ? `#WIN2003,#WIN8`

learn-more:
> > Maybe we could add a post-merge script that checks if there are changes in `modules\rostests` and updates / commits after that?
> 
> I thought about a cmake target, that does this, but is not compiled by default. Or are you thinking of some special CMake mechanism?

I was thinking about a github thing, where you can probably kick of a certain task based on what changed.

sdk/tools/whitelister/whitelister.cpp
tkreuzer: No. The idea is that you use the current commit hash and current line and enter that into the white-list file. Whenever the white-list is processed, this is reverse blamed to get the hash and line in the current code, even if new revisions were added.

sdk/tools/whitelister/whitelister.cpp
tkreuzer: This is how it currently works, yes, but I might change it, since compiling the lists takes quite some time. It shouldn't be an issue though, because I expect this to basically only happen for winetest sync, as apitests should be free of this. If you are working on an apitest, which has a whitelist, just fix the test first ;-).
Anyway, I will probably change it to a pre-commit hook or something, which then checks for a revision mismatch between the sources and the compiled whitelist and asks to manually recompile the whitelists in source.

sdk/tools/whitelister/whitelister.cpp
tkreuzer: Sure, you can do as you like, as long as the term is in there. "#WIN2003 and #WIN8" also works. Only "Works on #WIN7, but not on #WIN2003" wouldn't work as the text suggests.

tkreuzer:
> 
> I was thinking about a github thing, where you can probably kick of a certain task based on what changed.

Yeah, that might work, too. Gonna consider that.

sdk/tools/whitelister/whitelister.cpp
zefklop: Well, sometimes I tinker with winetests to e.g. avoid running a whole test while trying to pinpoint a bug. I'll have to live with that.

sdk/tools/whitelister/whitelister.cpp
zefklop: I mean, the string length (12).

modules/rostests/apitests/acpi/CMakeLists.txt
```diff
@@ -1,9 +1,13 @@
 
 include_directories(${REACTOS_SOURCE_DIR}/drivers/bus/acpi/acpica/include)
 
+generate_whitelist_code(acpi_apitest_whitelist.txt)
+add_definitions(-DUSE_WHITELIST)
```
ThFabba, line 5: Do you have a recommendation for how we should handle tests like this that aren't expected to succeed on Windows?
This is a real unit test, so it always runs our code, even on Windows.

We could move it to a new unittests subdirectory to make this clearer. And/or maybe exclude it from the list of tests we run on WHS?
Or do you think just having whitelist entries is a good permanent solution?

modules/rostests/apitests/acpi/acpi_apitest_whitelist.txt
```diff
@@ -0,0 +1,2 @@
+9bcd83507467 modules/rostests/apitests/acpi/Bus_PDO_QueryResourceRequirements.c 504 "Fails on #WIN2003 #WIN8"
+9bcd83507467 modules/rostests/apitests/acpi/Bus_PDO_QueryResourceRequirements.c 528 "Fails on #WIN2003"
```
ThFabba, line 2: I'm surprised that the above line includes WIN8 but this one doesn't. The results of this test should not depend on Windows version.

sdk/tools/whitelister/whitelister.cpp
tkreuzer: No, why would it change? Those are SHA1 hashes, they are 40 hex chars, but this is the default length that git shows. We don't need to be cryptographically secure, so 12 chars should be good enough.

tkreuzer: Well, I'm gonna switch  to `--porcelain`, which uses the full 40 char hash.

sdk/tools/whitelister/whitelister.cpp
tkreuzer: If it uses a git-hook, you wouldn't need to commit before building, but you would need to rebuild the whitelist before committing. I can also try a quick check feature, which would only check, whether the compiled whitelist has the hash of the last commit in that folder, and if not, recompile.

modules/rostests/apitests/acpi/CMakeLists.txt
tkreuzer: I suggest to skip those tests on Windows that are known to be broken. Or fix the code in ros, so that they succeed ;-)
The whitelists are generally not for our own tests, but for wine-tests, so we can silence them without hacking the code.
In our own tests I suggest using `todo`.

modules/rostests/apitests/acpi/acpi_apitest_whitelist.txt
tkreuzer: Nah, I was just experimenting with stuff.

sdk/include/reactos/wine/test.h
zefklop: Can we add WLA_TODO_ROS ? Broken &  flaky don't enter into that category.

As for specifying it, I'd suggest using `#REACTOS_TODO` `#REACTOS_FLAKY` etc. in the spec string. No need to add another annotation format if you already have one.

learn-more:
If you include the error output line in the whitelist,
you can detect that the same 'known bad' output occurs on another line, and suggest to rebuild the whitelist.
Optionally, for ignoring pointer like stuff, you might wanna include some pattern matching like https://stackoverflow.com/a/34824044/4928207

learn-more:

PR #3850, opened 2021-07-26
Pfn lock fix

zefklop:
## Purpose

Do not hold PFN lock when freeing section data.

JIRA issue: [CORE-17690](https://jira.reactos.org/browse/CORE-17690) [CORE-17698](https://jira.reactos.org/browse/CORE-17698) 

## Proposed changes

Always release the PFN lock in MmDereferenceSegmentWithLock
Add relevant annotations for this function & MiAcquirePfnLock & friends

Split from #3798

### DIFF
--- a/ntoskrnl/mm/section.c
+++ b/ntoskrnl/mm/section.c
@@ -987,6 +987,12 @@ FreeSegmentPage(PMM_SECTION_SEGMENT Segment, PLARGE_INTEGER Offset)
     MmReleasePageMemoryConsumer(MC_USER, Page);
 }
 
+_When_(OldIrql == MM_NOIRQL, _IRQL_requires_max_(DISPATCH_LEVEL))
+_When_(OldIrql == MM_NOIRQL, _Requires_lock_not_held_(MmPfnLock))
+_When_(OldIrql != MM_NOIRQL, _Requires_lock_held_(MmPfnLock))
+_When_(OldIrql != MM_NOIRQL, _Releases_lock_(MmPfnLock))
+_When_(OldIrql != MM_NOIRQL, _IRQL_restores_(OldIrql))
+_When_(OldIrql != MM_NOIRQL, _IRQL_requires_(DISPATCH_LEVEL))
 VOID
 NTAPI
 MmDereferenceSegmentWithLock(PMM_SECTION_SEGMENT Segment, KIRQL OldIrql)
--- a/ntoskrnl/mm/section.c
+++ b/ntoskrnl/mm/section.c
@@ -991,27 +991,21 @@ VOID
 NTAPI
 MmDereferenceSegmentWithLock(PMM_SECTION_SEGMENT Segment, KIRQL OldIrql)
 {
-    BOOLEAN HaveLock = FALSE;
-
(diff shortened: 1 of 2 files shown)


ntoskrnl/mm/section.c
```diff
@@ -987,45 +987,43 @@ FreeSegmentPage(PMM_SECTION_SEGMENT Segment, PLARGE_INTEGER Offset)
...
 NTAPI
 MmDereferenceSegmentWithLock(PMM_SECTION_SEGMENT Segment, KIRQL OldIrql)
 {
-    BOOLEAN HaveLock = FALSE;
-
     /* Lock the PFN lock because we mess around with SectionObjectPointers */
     if (OldIrql == MM_NOIRQL)
     {
-        HaveLock = TRUE;
         OldIrql = MiAcquirePfnLock();
```
zefklop, line 1003: continued from #3798 
@tkreuzer says
> Not a fan of this change. I don't see how it fixes anything. It just makes the function behave more unintuitively. What if I don't want to release the lock after calling this function? I would prefer to have 2 functions: MmDereferenceSegmentWithPfnLockHeld (which would not lock) and MmDereferenceSegment, which would acquire the lock and then call this function. That's way cleaner than all this passing of magic values.
I'm generally against lock changes in any function that isn't either called "AcquireXLock" or "ReleaseXLock".

zefklop: The issue we have here is that we use the PFN lock to protect SectionObjectPointers, Section object refcount & Segment flags. So when the section reference reaches zero, we want to dereference the segment, check if we have to free the segment and mark the segment as "in delete" under the lock. But then we have to release the lock because we may have to tinker with the FS driver in order to flush data. and reacquire it to clear SectionObjectPointers.
I also would like to see a solution with `MmDereferenceSegmentWithPfnLockHeld` but this looks tricky with the current state of things.

ntoskrnl/mm/section.c
```diff
@@ -987,45 +987,43 @@ FreeSegmentPage(PMM_SECTION_SEGMENT Segment, PLARGE_INTEGER Offset)
...
-            MiReleasePfnLock(OldIrql);
+        MiReleasePfnLock(OldIrql);
         return;
     }
 
     *Segment->Flags |= MM_SEGMENT_INDELETE;
-
-    if (HaveLock)
-        MiReleasePfnLock(OldIrql);
+    MiReleasePfnLock(OldIrql);
```
tkreuzer, line 1014: Repeating my other comment here: Just because the lock is released by all callers right after the call to this function is not argument to do it here, because it doesn't necessarily have to stay like that. It's also very unintuitive behavior. IMO no function, that isn't explicitly called "AcquireLock" or "ReleaseLock" or alike should acquire or release a lock. I'm already horrified of some Windows function names, that are called like "CheckForX", but then change X and do all kinds of stuff, no need to introduce more of that.

ntoskrnl/mm/section.c
tkreuzer: I still don't like it, but for the sake of getting this fixed asap, I guess we need to live with that for now.

tkreuzer (approved):

PR #3851, opened 2021-07-26
[SHELL32] Implement SHGetShellStyleHInstance

oleg-dubinskiy:
## Purpose

Imlement SHGetShellStyleHInstance function, which loads theme's or system shellstyle.dll.
For now, shellstyle support is completely unimplemented. This function is required only for some MS system components from XP/2003, like Add/remove programs applet (appwiz.cpl) and Picture and Fax Viewer (shimgvw.dll).
It allows to properly start MS appwiz.cpl with our shell32.dll in Windows XP/2003 at least (only from Control Panel, from system32 it still crashes), in case shellstyle.dll exists in theme/system directory, but it still does not start yet in ReactOS.
Some another issue(s) are still preventing the correct work of it in ROS, but nevertheless shell32 part is now (partially) fixed.

JIRA issue: [CORE-17707](https://jira.reactos.org/browse/CORE-17707)

## Analysis

This is maximally similar implementation to what Windows does:
shgetshellstylehinstance

Although some calls are unresolved, nevertheless it's visible what this function does in general.

## Result

As visible on the following screenshots, at least in Windows XP/2003 with our shell32.dll, MS Add/remove programs applet (appwiz.cpl) is now starting/working properly (if to launch it from Control Panel). :slightly_smiling_face: 

Before:
MS-appwiz-before

After:
MS-appwiz-after

But unfortunately, it still does not start in ReactOS, due to some other issue(s) from other component(s).

appwiz-do-launch-crash

If try to launch that from system32 directory without Control Panel (just by double-clicking), it crashes both in Windows and ReactOS. In this case, the problem is caused obviously by some another bug in our shell32, but in case of using Control Panel, the source of problem is caused by something other that causes an assert in (checked) MS duser.dll (used by appwiz.cpl) and prevents it from successful loading. :slightly_frowning_face:

### DIFF
--- a/dll/win32/shell32/shell32.cpp
+++ b/dll/win32/shell32/shell32.cpp
@@ -3,6 +3,7 @@
  *
  * Copyright 1998 Marcus Meissner
  * Copyright 1998 Juergen Schmied (jsch)  *  <juergen.schmied@metronet.de>
+ * Copyright 2021 Oleg Dubinskiy (oleg-dubinskiy)  *  <oleg.dubinskij2013@yandex.ua>
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
@@ -103,6 +104,106 @@ RegenerateUserEnvironment(LPVOID *lpEnvironment, BOOL bUpdateSelf)
     CloseHandle(hUserToken);
 
     return bResult;
+}
+
+/*
+ * Implemented
+ */
+EXTERN_C HINSTANCE
+WINAPI
+SHGetShellStyleHInstance(VOID)
+{
+    HINSTANCE hShellStyle = NULL;
+    WCHAR ThemePath[MAX_PATH] = {0}, *ShellStylePath = NULL, ExpandedShellStylePath[MAX_PATH] = {0}, ColorName[MAX_PATH] = {0};
+    HRESULT hr;
+
+    /* Get a path and a color scheme name of the current visual style */
+    hr = GetCurrentThemeName(ThemePath,
+                             MAX_PATH,
+                             ColorName,
+                             MAX_PATH,
+                             NULL,
+                             MAX_PATH);
+    if (FAILED(hr))
+    {
+        ERR("GetCurrentThemeName() failed with 0x%lx\n", hr);
+        goto next;
+    }
+
+    TRACE("Theme path is %ls\n", ThemePath);
+
+    TRACE("Color name is %ls\n", ColorName);
+
+    /* Remove a visual style filename from the string */
+    PathRemoveFileSpecW(ThemePath);
+
+    TRACE("Theme directory is %ls\n", ThemePath);
+
+    /* Add a path to shellstyle.dll */
+    lstrcatW(ThemePath, L"\\shell\\");
+    lstrcatW(ThemePath, ColorName);
+    lstrcatW(ThemePath, L"\\shellstyle.dll");
+
+    TRACE("Shellstyle path is %ls\n", ThemePath);
+
+    /* Create shellstyle path */
+    ShellStylePath = (LPWSTR)LocalAlloc(LPTR, sizeof(ThemePath));
+    hr = StringCchPrintfW(ShellStylePath,
+                          _countof(ThemePath),
+                          ThemePath);
+    if (FAILED(hr))
+    {
+        ERR("StringCchPrintfW() failed with 0x%lx\n", hr);
+        goto next;
+    }
+
+    /* Expand environment variable in the path, if any */
+    if (!ExpandEnvironmentStringsW(ShellStylePath,
+                                   ExpandedShellStylePath,
+                                   _countof(ExpandedShellStylePath)))
+    {
+        ERR("ExpandEnvironmentStringsW(%ls) failed\n", ShellStylePath);
+        goto next;
+    }
+
+next:
+
+    if (lstrlenW(ExpandedShellStylePath) != 0)
+    {
+        /* First, try expanded path, if it exists */
+        hShellStyle = LoadLibraryW(ExpandedShellStylePath);
+        if (hShellStyle == NULL)
+        {
+            ERR("LoadLibraryW(%ls) failed\n", ExpandedShellStylePath);
+            return NULL;
+        }
+    }
+    else if (ShellStylePath)
+    {
+        /* Next, try non-expanded path, if expanded was not created */
+        hShellStyle = LoadLibraryW(ShellStylePath);
+        if (hShellStyle == NULL)
+        {
+            ERR("LoadLibraryW(%ls) failed\n", ShellStylePath);
+            LocalFree(ShellStylePath);
+            return NULL;
+        }
+        LocalFree(ShellStylePath);
+    }
+    else if (lstrlenW(ExpandedShellStylePath) == 0 && !ShellStylePath)
+    {
+        /* Only if no any paths exist (assuming current visual style has no shellstyle.dll),
+         * then just try to load the one from system32 directory */
+        hShellStyle = LoadLibraryW(L"shellstyle.dll");
+        if (hShellStyle == NULL)
+        {
+            ERR("LoadLibraryW() failed\n");
+            return NULL;
+        }
+    }
+
+    /* we're done */
+    return hShellStyle;
 }
 
 /**************************************************************************
--- a/dll/win32/shell32/stubs.cpp
+++ b/dll/win32/shell32/stubs.cpp
@@ -1315,14 +1315,3 @@ SHGetAttributesFromDataObject(IDataObject *pdo,
     FIXME("SHGetAttributesFromDataObject() stub\n");
     return E_NOTIMPL;
 }
-
-/*
- * Unimplemented
- */
-EXTERN_C HINSTANCE
-WINAPI
-SHGetShellStyleHInstance(VOID)
-{
-    FIXME("SHGetShellStyleHInstance() stub\n");
-    return NULL;
-}


milawynsrealm:
There is also a shellstyle.dll in the system32 folder as well. I think this is used when the classic theme is being used.

oleg-dubinskiy:
> There is also a shellstyle.dll in the system32 folder as well. I think this is used when the classic theme is being used.

I also considered this case. Loading default shellstyle.dll from system32 is also implemented. :wink:

dll/win32/shell32/shell32.cpp
```diff
@@ -105,6 +106,106 @@ RegenerateUserEnvironment(LPVOID *lpEnvironment, BOOL bUpdateSelf)
...
+                             MAX_PATH);
+    if (FAILED(hr))
+    {
+        ERR("GetCurrentThemeName() failed with 0x%lx\n", hr);
+        goto next;
+    }
+
+    TRACE("Theme path is %ls\n", ThemePath);
+
+    TRACE("Color name is %ls\n", ColorName);
```
learn-more, line 135: Can combine this in one TRACE.

dll/win32/shell32/shell32.cpp
```diff
@@ -105,6 +106,106 @@ RegenerateUserEnvironment(LPVOID *lpEnvironment, BOOL bUpdateSelf)
...
+
+    /* Remove a visual style filename from the string */
+    PathRemoveFileSpecW(ThemePath);
+
+    TRACE("Theme directory is %ls\n", ThemePath);
```
learn-more, line 140: Additionally add this one as well.

dll/win32/shell32/shell32.cpp
```diff
@@ -105,6 +106,106 @@ RegenerateUserEnvironment(LPVOID *lpEnvironment, BOOL bUpdateSelf)
...
+
+    /* Add a path to shellstyle.dll */
+    lstrcatW(ThemePath, L"\\shell\\");
```
learn-more, line 143: Please use `StringCchcatW` instead.

dll/win32/shell32/shell32.cpp
```diff
@@ -105,6 +106,106 @@ RegenerateUserEnvironment(LPVOID *lpEnvironment, BOOL bUpdateSelf)
...
+    lstrcatW(ThemePath, ColorName);
+    lstrcatW(ThemePath, L"\\shellstyle.dll");
+
+    TRACE("Shellstyle path is %ls\n", ThemePath);
+
+    /* Create shellstyle path */
+    ShellStylePath = (LPWSTR)LocalAlloc(LPTR, sizeof(ThemePath));
+    hr = StringCchPrintfW(ShellStylePath,
+                          _countof(ThemePath),
+                          ThemePath);
```
learn-more, line 153: ???? what on earth are you doing here?

dll/win32/shell32/shell32.cpp
```diff
@@ -105,6 +106,106 @@ RegenerateUserEnvironment(LPVOID *lpEnvironment, BOOL bUpdateSelf)
...
+    if (FAILED(hr))
+    {
+        ERR("StringCchPrintfW() failed with 0x%lx\n", hr);
+        goto next;
```
learn-more, line 157: Please don't use goto.

dll/win32/shell32/shell32.cpp
```diff
@@ -105,6 +106,106 @@ RegenerateUserEnvironment(LPVOID *lpEnvironment, BOOL bUpdateSelf)
...
+
+    if (lstrlenW(ExpandedShellStylePath) != 0)
+    {
+        /* First, try expanded path, if it exists */
+        hShellStyle = LoadLibraryW(ExpandedShellStylePath);
+        if (hShellStyle == NULL)
+        {
+            ERR("LoadLibraryW(%ls) failed\n", ExpandedShellStylePath);
+            return NULL;
+        }
```
learn-more, line 179: I doubt this is correct,
I'd assume that it returns the `hShellStyle` if it is NOT null.

dll/win32/shell32/shell32.cpp
```diff
@@ -105,6 +106,106 @@ RegenerateUserEnvironment(LPVOID *lpEnvironment, BOOL bUpdateSelf)
...
+
+    /* Add a path to shellstyle.dll */
+    lstrcatW(ThemePath, L"\\shell\\");
+    lstrcatW(ThemePath, ColorName);
+    lstrcatW(ThemePath, L"\\shellstyle.dll");
+
+    TRACE("Shellstyle path is %ls\n", ThemePath);
+
+    /* Create shellstyle path */
+    ShellStylePath = (LPWSTR)LocalAlloc(LPTR, sizeof(ThemePath));
```
learn-more, line 150: Why do you allocate this?

dll/win32/shell32/shell32.cpp
```diff
@@ -105,6 +106,106 @@ RegenerateUserEnvironment(LPVOID *lpEnvironment, BOOL bUpdateSelf)
...
+        hShellStyle = LoadLibraryW(ShellStylePath);
+        if (hShellStyle == NULL)
+        {
+            ERR("LoadLibraryW(%ls) failed\n", ShellStylePath);
+            LocalFree(ShellStylePath);
+            return NULL;
+        }
+        LocalFree(ShellStylePath);
+    }
+    else if (lstrlenW(ExpandedShellStylePath) == 0 && !ShellStylePath)
```
learn-more, line 193: I doubt the fallback is based on the string length being 0,
I'd assume that the fallback is rather based on if the dll could be loaded.

HBelusca:
> This function is required only for some MS system components from XP/2003, like Add/remove programs applet (appwiz.cpl) and Picture and Fax Viewer (shimgvw.dll).

There was a discussion on mattermost where it was determined that this assertion is wrong.

HBelusca (changes requested):
This PR is completely useless, see the ~~mattermost~~ discord discussion about that.
https://discord.com/channels/374957651676626945/831490939524087888/862709027212886047

learn-more:
> This PR is completely useless, see the ~mattermost~ discord discussion about that.
> https://discord.com/channels/374957651676626945/831490939524087888/862709027212886047
> 
> EDIT: OK granted, the MS Add/Remove Programs applet does fail if there's no shellstyle.dll / the SHGetShellStyleHInstance API returns NULL.

This comment is completely useless, people need an account to read that discussion.

dll/win32/shell32/shell32.cpp
```diff
@@ -105,6 +106,106 @@ RegenerateUserEnvironment(LPVOID *lpEnvironment, BOOL bUpdateSelf)
...
+    {
+        /* Only if no any paths exist (assuming current visual style has no shellstyle.dll),
+         * then just try to load the one from system32 directory */
+        hShellStyle = LoadLibraryW(L"shellstyle.dll");
+        if (hShellStyle == NULL)
+        {
+            ERR("LoadLibraryW() failed\n");
+            return NULL;
+        }
+    }
```
HBelusca, line 203: You can merge all these different LoadLibrary calls into just one, that uses the path pointed by a pointer.

dll/win32/shell32/shell32.cpp
```diff
@@ -105,6 +106,106 @@ RegenerateUserEnvironment(LPVOID *lpEnvironment, BOOL bUpdateSelf)
...
+    else if (ShellStylePath)
+    {
+        /* Next, try non-expanded path, if expanded was not created */
+        hShellStyle = LoadLibraryW(ShellStylePath);
+        if (hShellStyle == NULL)
+        {
+            ERR("LoadLibraryW(%ls) failed\n", ShellStylePath);
+            LocalFree(ShellStylePath);
+            return NULL;
+        }
```
HBelusca, line 190: You can merge all these different LoadLibrary calls into just one, that uses the path pointed by a pointer.

dll/win32/shell32/shell32.cpp
```diff
@@ -105,6 +106,106 @@ RegenerateUserEnvironment(LPVOID *lpEnvironment, BOOL bUpdateSelf)
...
+        if (hShellStyle == NULL)
+        {
+            ERR("LoadLibraryW(%ls) failed\n", ExpandedShellStylePath);
+            return NULL;
+        }
+    }
+    else if (ShellStylePath)
+    {
+        /* Next, try non-expanded path, if expanded was not created */
+        hShellStyle = LoadLibraryW(ShellStylePath);
```
HBelusca, line 184:
```suggestion
        hShellStyle = LoadLibraryExW(ShellStylePath, NULL, LOAD_LIBRARY_AS_DATAFILE);
```

dll/win32/shell32/shell32.cpp
```diff
@@ -105,6 +106,106 @@ RegenerateUserEnvironment(LPVOID *lpEnvironment, BOOL bUpdateSelf)
...
+            LocalFree(ShellStylePath);
+            return NULL;
+        }
+        LocalFree(ShellStylePath);
+    }
+    else if (lstrlenW(ExpandedShellStylePath) == 0 && !ShellStylePath)
+    {
+        /* Only if no any paths exist (assuming current visual style has no shellstyle.dll),
+         * then just try to load the one from system32 directory */
+        hShellStyle = LoadLibraryW(L"shellstyle.dll");
```
HBelusca, line 197:
```suggestion
        hShellStyle = LoadLibraryExW(L"shellstyle.dll", NULL, LOAD_LIBRARY_AS_DATAFILE);
```

HBelusca: I wonder whether it would be worth forcing a `%SystemRoot%\\System32\\ShellStyle.dll` path there... (so that it's not going to load a shellstyle from the current directory of the app that calls that API)

PR #3859, opened 2021-07-29
[IMM32] Rewrite ImmEnumInputContext

katahiromz:
## Purpose

## Proposed changes

- Add `Imm32IsValidHimc`, `Imm32BuildHimcList` and `Imm32AllocAndBuildHimcList` helper functions.
- Rewrite `ImmEnumInputContext` function by using them.
- Modify `NtUserBuildHimcList`.

### DIFF
--- a/dll/win32/imm32/imm.c
+++ b/dll/win32/imm32/imm.c
@@ -1161,11 +1161,6 @@ static PCLIENTIMC APIENTRY Imm32GetClientImcCache(void)
     return NULL;
 }
 
-static BOOL Imm32IsValidHimc(HIMC hIMC)
-{
-    return (hIMC != NULL) && RtlValidateHeap(g_hImm32Heap, 0, hIMC);
-}
-
 static NTSTATUS APIENTRY
 Imm32BuildHimcList(DWORD dwThreadId, DWORD dwCount, HIMC *phList, LPDWORD pdwCount)
 {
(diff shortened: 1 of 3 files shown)


dll/win32/imm32/imm.c
```diff
@@ -1161,6 +1161,55 @@ static PCLIENTIMC APIENTRY Imm32GetClientImcCache(void)
     return NULL;
 }
 
+static BOOL Imm32IsValidHimc(HIMC hIMC)
+{
+    return (hIMC != NULL); // FIXME
```
learn-more, line 1166: If you give imm32 it's own heap for `Imm32HeapAlloc`, and you allocate an HIMC from this heap,
you can use `RtlValidateHeap` to check that it was allocated from the correct heap.
That seems like a good starting point to validate an HIMC?

katahiromz: OK. Thanks.

dll/win32/imm32/imm.c
```diff
@@ -3735,11 +3784,31 @@ BOOL WINAPI ImmDisableTextFrameService(DWORD dwThreadId)
...
+    TRACE("ImmEnumInputContext(%lu, %p, %p)\n", dwThreadId, lpfn, lParam);
+
+    dwCount = Imm32AllocAndBuildHimcList(dwThreadId, &phList);
+    if (!dwCount)
+        return FALSE;
+
+    for (dwIndex = 0; dwIndex < dwCount; ++dwIndex)
+    {
+        hIMC = phList[dwIndex];
+        if (!Imm32IsValidHimc(hIMC))
```
HBelusca, line 3803: Why wouldn't it be valid? that hIMC entry comes directly from that allocated phList by Imm32AllocAndBuildHimcList and initialized by Imm32BuildHimcList(), and so its contents must be valid by construction ?

katahiromz: I deleted the check.

PR #3860, opened 2021-07-29
[CONFIGURE] Make configure.sh print out an error when CMake fails

nexos-dev:
As it currently stands, configure.sh prints out a success message if CMake fails. This commit makes it print out an error message if it fails. Note that configure.sh needs reworking anyway, as configure.cmd is written a lot better


Extravert-ir:
Thanks for the contribution!

PR #3863, opened 2021-07-30
[NTOS:MM] Do not require PFN lock held in MiDeletePte

zefklop:
There is no reason to hold it during all this time.

## Purpose

Finer grain of the PFN DB locking


ntoskrnl/mm/ARM3/pagfault.c
```diff
@@ -1227,16 +1227,26 @@ MiResolveProtoPteFault(IN BOOLEAN StoreInstruction,
...
-        /* Because now we use this */
-        Pfn1 = MI_PFN_ELEMENT(PageFrameIndex);
-        MiInitializePfn(PageFrameIndex, PointerPte, TRUE);
-
         /* Fix the protection */
         Protection &= ~MM_WRITECOPY;
         Protection |= MM_READWRITE;
+
+        /* Restore our PTE to a sensible value for our PFN */
+        PointerPte->u.Long = MmProtectToPteMask[Protection];
```
tkreuzer, line 1241: Use `` MI_UPDATE_VALID_PTE``

zefklop: I don't think that we want to keep the old mapping in Pfn->OriginalPte.

ntoskrnl/mm/ARM3/virtual.c
```diff
@@ -375,9 +375,7 @@ MiDeletePte(IN PMMPTE PointerPte,
     MMPTE TempPte;
     PFN_NUMBER PageFrameIndex;
     PMMPDE PointerPde;
-
-    /* PFN lock must be held */
-    MI_ASSERT_PFN_LOCK_HELD();
```
Extravert-ir, line 380: Assert the lock not held?

zefklop: MiAcquirePfnLock does this for itself, now that recursive spinlock acquisition is forbidden everywhere.

ntoskrnl/mm/ARM3/pagfault.c
zefklop: On second thought, we don't want the hardware bits here, but the soft PTE with the proper protection. Corrected.

zefklop:
@tkreuzer can we move forward ?

ntoskrnl/mm/ARM3/pagfault.c
tkreuzer: I still think this shouldn't just be updated like this. There is a reason for using the MI_*_PTE functions. Simply writing to PTEs without any checks should not be done unless there is a real reason to do so, which should be provided in a comment.

learn-more:

PR #3872, opened 2021-08-01
[GITHUB] Remove VS 2015 build from the CI

Extravert-ir:
Let's establish a set of rules regarding compiler support in ReactOS to avoid holywars in future.

## Proposed rules
- We officially support __no more than two versions__ of each compiler.
- The term "officially supported" means there is a job on one of our CI systems (either GitHub Actions or BuildBot) checking both Debug and Release builds for this compiler for supported platforms (currently it is i386 and amd64). Every developer has to check changes against those compilers and/or fix breakages ASAP if they happen.
- Two versions usually mean having one "stable" version which is not changed often and one "latest" version. For example, VS 2017 and VS 2019 (or that will likely be VS 2022 soon) or RosBE-provided GCC 8.4.0 and the current latest release (GCC 11.1.0).
- "stable" compiler should be upgraded once in 2-3 years, or by a demand (for example, when it can't support our code anymore without too many hacks)
- Contributors can maintain support for other compilers and toolchains, but the topic of merging such changes should be treated carefully. First of all, it shouldn't require too many changes in our code base, especially for 3rd party modules.

So according to this set of rules, I suggest to leave these compilers as "officially supported":
- GCC 8.4.0 for RosBE (aka "gcc-stable")
- MSVC/VS 2019 (aka "msvc-latest")
- MSVC/VS 2017 (aka "msvc-stable")
- LLVM/Clang 12 (with MSVC and GNU-like APIs) (aka "clang-latest")

Clang does not have a "stable" version yet, as the work about its support is still in progress.

## Reasoning
- Supporting each compiler requires an extra effort from every developer contributing to the project. Considering the size of our team, I think we may spend that time on something more useful. 4 versions is already a lot.
- The more a compiler gets old, the more work is required to update the codebase for a newer version. Remember our GCC 4.7.2 -> 8.4.0 transition, it took about 1.5 years in total to make it stable.
- Old compilers don't support newer language features and static analysis hints. This is especially relevant considering our plan to incorporate libcxx (LLVM's C++ standard library) into our project.
- Supporting two versions of a compiler is motivated by a fact that the majority of developers prefer using newer tooling, but that may lead to breakages from the compiler side (I'm looking at you, MSVC). In such a case, it is always better to have a more stable base to verify that the breakage is really caused by a compiler.

Additions and remarks are welcome.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -84,24 +84,14 @@ jobs:
   build-msvc:
     strategy:
       matrix:
-        os: [windows-latest, windows-2019]
-        toolset: ['14.2', '14.1', '14.0'] # VS 2019, 2017, and 2015 (see below)
+        os: [windows-latest]
+        toolset: ['14.2', '14.1'] # VS 2019, and 2017 (see below)
         arch: [i386, amd64]
         config: [Debug, Release]
         dllver: ['0x502', '0x600']
-        exclude: # VS 2019, 2017 only with windows-latest; VS 2015 only with windows-2019
-          - os: windows-2019
-            toolset: '14.2'
-          - os: windows-2019
-            toolset: '14.1'
-          - os: windows-latest
-            toolset: '14.0'
-          - dllver: 0x600
-            os: windows-2019
+        exclude: # VS 2019, 2017 only with windows-latest
           - dllver: 0x600
             toolset: '14.1'
-          - dllver: 0x600
-            toolset: '14.0'
           - dllver: 0x600
             config: Release
       fail-fast: false
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ The code of ReactOS is licensed under [GNU GPL 2.0](https://github.com/reactos/r
 To build the system it is strongly advised to use the _ReactOS Build Environment (RosBE)._
 Up-to-date versions for Windows and for Unix/GNU-Linux are available from our download page at: ["Build Environment"](https://reactos.org/wiki/Build_Environment).
 
-Alternatively one can use Microsoft Visual C++ (MSVC) version 2015+. Building with MSVC is covered here: ["Visual Studio or Microsoft Visual C++"](https://reactos.org/wiki/CMake#Visual_Studio_or_Microsoft_Visual_C.2B.2B).
+Alternatively one can use Microsoft Visual C++ (MSVC) version 2019+. Building with MSVC is covered here: ["Visual Studio or Microsoft Visual C++"](https://reactos.org/wiki/CMake#Visual_Studio_or_Microsoft_Visual_C.2B.2B).
 
 See ["Building ReactOS"](https://reactos.org/wiki/Building_ReactOS) article for more details.
 


SergeGautherie:
Setting an explicit policy is fine.
But it still seems odd to me that such a PR/discussion looks to come out of nowhere. It reminds me a lot of #2658...

SergeGautherie:
Official: why remove 2015 support already, especially when current 2019 is broken?
GitHub: why not simply mark 14.0 build as non-blocking for the time being, as still/contributor maintained, if need be?

SergeGautherie:
Why not add 'GCC 11.1.0 (aka "gcc-latest")' to the CI?

Extravert-ir:
> why remove 2015 support already, especially when current 2019 is broken?
Because there is 2017 which works perfectly fine, and it doesn't make sense to support two compilers with the same set of features.

> why not simply mark 14.0 build as deprecated/non-blocking for the time being, as still/contributor maintained, if need be?

Contributors may support whatever they want, but that shouldn't put an obligation on code devs. That's why I want to leave only "officially supported" toolchains in the CI

> Why not add 'GCC 11.1.0 (aka "gcc-latest")' to the CI?
That fits the scheme I described, so if there will be a PR which sets up such a build and makes it green, it will be accepted

SergeGautherie:
> > why not simply mark 14.0 build as deprecated/non-blocking for the time being, as still/contributor maintained, if need be?
> 
> Contributors may support whatever they want, but that shouldn't put an obligation on code devs. That's why I want to leave only "officially supported" toolchains in the CI

I understand your goal, but why link the two things?
Move 2015 from officially-supported to contributor-supported (not unsupported), mark 14.0 jobs as not failing the whole run, until 2015 can't actually handle future ROS code.
That should be enough to remove dev obligation, while still helping 2015 contributors.

SergeGautherie:
> VS 2022 soon

`windows-2022` is now official and seems to work right (for i386 at least).

DarkFire01 (approved):

JoachimHenze (changes requested):
NO!

DarkFire01:
Leave it open, I’m taking it over.

cbialorucki (approved):

DarkFire01:
Anyway; All the drama aside let's discuss the REAL reason I'm for this.

1) The WINESYNC movement has been going strong but we're currently running into some usages of C11,
According to the Microsoft documentation 
> C11 Compiler support for C11 and C17 requires Visual Studio 2019 version 16.8 or higher.
https://learn.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=msvc-170

MSVC 2015 is a no go, although; It seems 2017 manages currently. 
This discussion will need to happen once more when we run into an issue with it, for now. we're just discussing Vs2015

2) https://github.com/dotnet/wpf/tree/main/src/Microsoft.DotNet.Wpf/src/WpfGfx
This can be used as a replacement for milcore.dll from Vista RTM, the next version we're targeting therefore for implementing WDDM this is a requirement.

This entire code base uses a bunch of C/C++ that isn't supported under VS2015.

3) ucrtbase.dll
The new ucrt requires some feature vs2015 doesn't support as well.

4) https://github.com/microsoft/SymCrypt
Symcrypt is the next step to our bcrypt.dll. it's what windows 10 currently uses, and it's under MIT as well.

ALSO it needs features vs2015 doesn't support.

These tasks bring too much benefit, and in general we've outgrown vs2015


For these reasons and more, we're dropping vs2015. 
This is more then just a troll, it's a logical choice. 

This PR will be merged in a few days :) or when another approval has been obtained.

SigmaTel71:
> 3\. The new ucrt requires some feature vs2015 doesn't support as well.

Wait, didn't VS2015 *introduce* Universal CRT? Or do you mean that UCRT in VS2017 is more feature-rich than VS2015 one?
<details>

image
</details>

DarkFire01:
> > 3. The new ucrt requires some feature vs2015 doesn't support as well.
> 
> Wait, didn't VS2015 _introduce_ Universal CRT? Or do you mean that UCRT in VS2017 is more feature-rich than VS2015 one?
> 
> image

Nah we're talking about a code base MS released under MIT Timo has been working importing.

maj113 (changes requested):
Remove the windows-2019 builder since it was used only for VS2015 (14.0) builds, after that you can also remove all the os excludes that mention windows-2019 builder since it won't be used anymore

SergeGautherie:
> Remove the windows-2019 builder

Indeed, see/update line 92.

tkreuzer (approved):

maj113 (approved):

PR #3874, opened 2021-08-02
[NTOS:IO] Improvements to OpenRegistryHandlesFromSymbolicLink

oleg-dubinskiy:
## Purpose

Add two improvements in `OpenRegistryHandlesFromSymbolicLink` function: make the returned registry handles as output parameters and introduce a BOOLEAN parameter, which allows to create or only open registry keys.

## Proposed changes

- The returned handles to registry keys should be actually output parameters, not input. Just store received handles in them instead of storing uninitialized handles before getting the data. It allows to simplify the function a bit and get rid of unnecessary variables, and is more compatible with Windows.
- Introduce new BOOLEAN parameter, which allows the caller to choose: create registry keys or just open them. The thing is, this routine is used differently by different IoDeviceInterface* functions: some of them need to create the keys, some - just open (e. g. IoGetDeviceInterfaceAlias: opening the keys without creating them fixes the kmtest results from my #3595 PR and makes them same as on Windows). Also the similarly is done on Windows, except the code which creates the last two subkeys is moved into a separate function, and that's called from here with FALSE (so the keys are always opened). But I think we don't necessarily need to completely repeat Windows routines, so everything may be just done in one function, which I partially did in this PR.
- Update apprpriately all calls of this function in other routines.
- Additionally, improve debug prints appropriately.

### DIFF
--- a/ntoskrnl/io/iomgr/deviface.c
+++ b/ntoskrnl/io/iomgr/deviface.c
@@ -32,39 +32,21 @@ static
 NTSTATUS
 OpenRegistryHandlesFromSymbolicLink(IN PUNICODE_STRING SymbolicLinkName,
                                     IN ACCESS_MASK DesiredAccess,
-                                    IN OPTIONAL PHANDLE GuidKey,
-                                    IN OPTIONAL PHANDLE DeviceKey,
-                                    IN OPTIONAL PHANDLE InstanceKey)
+                                    IN BOOLEAN bCreate,
+                                    OUT OPTIONAL PHANDLE GuidKey,
+                                    OUT OPTIONAL PHANDLE DeviceKey,
+                                    OUT OPTIONAL PHANDLE InstanceKey)
 {
     OBJECT_ATTRIBUTES ObjectAttributes;
     UNICODE_STRING BaseKeyU;
     UNICODE_STRING GuidString, SubKeyName, ReferenceString;
     PWCHAR StartPosition, EndPosition;
     HANDLE ClassesKey;
-    PHANDLE GuidKeyRealP, DeviceKeyRealP, InstanceKeyRealP;
     HANDLE GuidKeyReal, DeviceKeyReal, InstanceKeyReal;
     NTSTATUS Status;
 
     SubKeyName.Buffer = NULL;
-
-    if (GuidKey != NULL)
-        GuidKeyRealP = GuidKey;
-    else
-        GuidKeyRealP = &GuidKeyReal;
-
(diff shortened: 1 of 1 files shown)


ntoskrnl/io/iomgr/deviface.c
```diff
@@ -32,40 +32,21 @@ static
...
+                                    OUT OPTIONAL PHANDLE DeviceKey,
+                                    OUT OPTIONAL PHANDLE InstanceKey)
 {
     OBJECT_ATTRIBUTES ObjectAttributes;
     UNICODE_STRING BaseKeyU;
     UNICODE_STRING GuidString, SubKeyName, ReferenceString;
     PWCHAR StartPosition, EndPosition;
     HANDLE ClassesKey;
-    PHANDLE GuidKeyRealP, DeviceKeyRealP, InstanceKeyRealP;
     HANDLE GuidKeyReal, DeviceKeyReal, InstanceKeyReal;
```
tkreuzer, line 45: These need to be zero-initialized.

oleg-dubinskiy: Done, thanks. (see the new line below)

ntoskrnl/io/iomgr/deviface.c
```diff
@@ -32,39 +32,21 @@ static
 NTSTATUS
 OpenRegistryHandlesFromSymbolicLink(IN PUNICODE_STRING SymbolicLinkName,
                                     IN ACCESS_MASK DesiredAccess,
-                                    IN OPTIONAL PHANDLE GuidKey,
-                                    IN OPTIONAL PHANDLE DeviceKey,
-                                    IN OPTIONAL PHANDLE InstanceKey)
+                                    IN BOOLEAN bCreate,
```
GeoB99, line 35: Nit: In the realm of NT, Hungarian notation isn't used (nor accepted). You could simply rename it just as `Create`, `CreateNonExistent` or any name of you choice. Other than that, the PR is OK.

GeoB99 (approved):

ntoskrnl/io/iomgr/deviface.c
oleg-dubinskiy: Closing as superseded.

HBelusca:
Hello @oleg-dubinskiy , why was this PR closed?

oleg-dubinskiy:
> Hello @oleg-dubinskiy , why was this PR closed?

Because it was already superseded by #3510 (already a part of that). Look at the code to see the difference. :wink:

PR #3875, opened 2021-08-03
[NTFS] New defines & stub for reading encrypted files

nagayev:
## Purpose
We shoudn't read file if it's encrypted.
Also I added several new defines for future contributions.

## Proposed changes
- Add new defines for NTFS's file attributes
- Don't read file if it's encrypted

### DIFF
--- a/drivers/filesystems/ntfs/fcb.c
+++ b/drivers/filesystems/ntfs/fcb.c
@@ -144,6 +144,12 @@ NtfsFCBIsCompressed(PNTFS_FCB Fcb)
 }
 
 BOOLEAN
+NtfsFCBIsEncrypted(PNTFS_FCB Fcb)
+{
+    return ((Fcb->Entry.FileAttributes & NTFS_FILE_TYPE_ENCRYPTED) == NTFS_FILE_TYPE_ENCRYPTED);
+}
+
+BOOLEAN
 NtfsFCBIsRoot(PNTFS_FCB Fcb)
 {
     return (wcscmp(Fcb->PathName, L"\\") == 0);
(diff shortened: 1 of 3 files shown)


drivers/filesystems/ntfs/fcb.c
```diff
@@ -143,6 +143,12 @@ NtfsFCBIsCompressed(PNTFS_FCB Fcb)
     return ((Fcb->Entry.FileAttributes & NTFS_FILE_TYPE_COMPRESSED) == NTFS_FILE_TYPE_COMPRESSED);
 }
 
+BOOLEAN
+NtfsFCBIsEncrypted(PNTFS_FCB Fcb)
+{
+    return ((Fcb->Entry.FileAttributes & NTFS_FILE_TYPE_ENCRYPTED) == NTFS_FILE_TYPE_ENCRYPTED);
```
SergeGautherie, line 149: The `==` part seems overkill to check one bit only, even though some existing code does it that way.

drivers/filesystems/ntfs/ntfs.h
```diff
@@ -931,6 +935,9 @@ NtfsFCBIsReparsePoint(PNTFS_FCB Fcb);
 BOOLEAN
 NtfsFCBIsCompressed(PNTFS_FCB Fcb);
 
+BOOLEAN
+NtfsFCBIsEncrypted(PNTFS_FCB Fcb);
+
```
SergeGautherie, line 940: Move this part to the second commit.

nagayev: okey

drivers/filesystems/ntfs/fcb.c
ColinFinck: It's consistent with the existing code, so that should be fine

ColinFinck (approved):

Extravert-ir:
Don't quite get why it was needed to create a separate function for checking every individual flag, but that's not a topic of this PR

PR #3878, opened 2021-08-04
Get rid of "memory consumers" in legacy Mm

zefklop:
## Purpose

Lighten up the code.

### DIFF
--- a/ntoskrnl/cc/copy.c
+++ b/ntoskrnl/cc/copy.c
@@ -16,8 +16,6 @@
 
 /* GLOBALS *******************************************************************/
 
-static PFN_NUMBER CcZeroPage = 0;
-
 #define MAX_ZERO_LENGTH    (256 * 1024)
 
 typedef enum _CC_CAN_WRITE_RETRY
@@ -44,30 +42,6 @@ ULONG CcDataPages = 0;
 ULONG CcDataFlushes = 0;
 
 /* FUNCTIONS *****************************************************************/
-
-VOID
-NTAPI
-MiZeroPhysicalPage (
-    IN PFN_NUMBER PageFrameIndex
-);
-
-VOID
-NTAPI
-CcInitCacheZeroPage (
-    VOID)
-{
-    NTSTATUS Status;
-
-    MI_SET_USAGE(MI_USAGE_CACHE);
-    //MI_SET_PROCESS2(PsGetCurrentProcess()->ImageFileName);
-    Status = MmRequestPageMemoryConsumer(MC_SYSTEM, TRUE, &CcZeroPage);
-    if (!NT_SUCCESS(Status))
-    {
-        DbgPrint("Can't allocate CcZeroPage.\n");
-        KeBugCheck(CACHE_MANAGER);
-    }
-    MiZeroPhysicalPage(CcZeroPage);
-}
 
 VOID
 CcPostDeferredWrites(VOID)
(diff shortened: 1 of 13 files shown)


ntoskrnl/cc/copy.c
```diff
@@ -809,48 +783,51 @@ CcZeroData (
...
-        PPFN_NUMBER PfnArray;
+        PHYSICAL_ADDRESS ZeroAddress;
+        PHYSICAL_ADDRESS MaxAddress;
 
-        /* Setup our Mdl */
-        Mdl = IoAllocateMdl(NULL, min(Length, MAX_ZERO_LENGTH), FALSE, FALSE, NULL);
+        /* Get a zeroed MDL */
+        ZeroAddress.QuadPart = 0;
+        MaxAddress.QuadPart = MAXLONGLONG;
+        Mdl = MmAllocatePagesForMdl(ZeroAddress, MaxAddress, ZeroAddress, min(Length, MAX_ZERO_LENGTH));
```
Extravert-ir, line 793: Hm, MSDN recommends using `MmAllocatePagesForMdlEx`. Does it make sense here?

Extravert-ir:
Btw, according to the description on MSDN, our `CcZeroData` is not really "implemented":
https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/nf-ntifs-cczerodata

Extravert-ir (approved):

Extravert-ir:
I tested this branch, seeing some issues:
1. During the setup, there is a debug log spam with messages like
`(ntoskrnl\mm\freelist.c:196) ARM3-DEBUG: Being called with 0 7fffffffffffffff 0 9600 3 0`

2. Task manager no longer shows the current memory used. Was this number also "bogus"?

Before:
VirtualBox_reactos1_05_08_2021_05_01_25

After:
VirtualBox_reactos1_05_08_2021_04_53_33

ntoskrnl/cc/copy.c
zefklop: That would be with 0 as flags. That's still better I guess.

zefklop:
> I tested this branch, seeing some issues:
> 
>     1. During the setup, there is a debug log spam with messages like
>        `(ntoskrnl\mm\freelist.c:196) ARM3-DEBUG: Being called with 0 7fffffffffffffff 0 9600 3 0`
Will silence it.

>     2. Task manager no longer shows the current memory used. Was this number also "bogus"?
> 
Do you mean "system cache" ? Yes, it was bogus.

zefklop:
> Btw, according to the description on MSDN, our `CcZeroData` is not really "implemented":
> https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/nf-ntifs-cczerodata

What do you mean ?

Extravert-ir:
> Do you mean "system cache" ? Yes, it was bogus.

I mean Total Commit Charge

> What do you mean ?

I mean our implementation doesn't use `Wait` parameter and always returns `TRUE`

zefklop:
> > Do you mean "system cache" ? Yes, it was bogus.
> 
> I mean Total Commit Charge
Bogus too :disappointed: 
> 
> > What do you mean ?
> 
> I mean our implementation doesn't use `Wait` parameter and always returns `TRUE`
Right. I should probably change that.

ntoskrnl/ex/sysinfo.c
```diff
@@ -1568,30 +1557,9 @@ SSI_DEF(SystemDpcBehaviourInformation)
...
-           TheIdleProcess->UniqueProcessId,
-           TheIdleProcess->Pcb.KernelTime,
-           MiFreeSwapPages,
-           MiUsedSwapPages);
-
-    *Spi = MiMemoryConsumers[MC_USER].PagesUsed;
-
-    return STATUS_SUCCESS;
+    /* Unimplemented as per https://www.geoffchappell.com/studies/windows/km/ntoskrnl/api/ex/sysinfo/query.htm?tx=76&ts=0,1254 */
+    DPRINT1("NtSetSystemInformation - SystemFullMemoryInformation not implemented\n");
```
HBelusca, line 1561: Huh... are you sure? Try testing that on a **checked build** of Windows.

HBelusca: Btw. this property class uses the same implementation than SystemSummaryMemoryInformation.

zefklop: This is why I leave it like this and not with the `SI_XX` macro below.

In any case, this one doesn't provide a `ULONG` but a `SYSTEM_MEMORY_INFORMATION` structure, so this can't be more wrong than it currently is.

HBelusca: thanks for the info!

gonzoMD:
Fewer memory after the patch?

ThFabba (approved):

ntoskrnl/ex/sysinfo.c
```diff
@@ -1568,30 +1557,9 @@ SSI_DEF(SystemDpcBehaviourInformation)
...
-    DPRINT("PID: %p, KernelTime: %u PFFree: %lu PFUsed: %lu\n",
-           TheIdleProcess->UniqueProcessId,
-           TheIdleProcess->Pcb.KernelTime,
-           MiFreeSwapPages,
-           MiUsedSwapPages);
-
-    *Spi = MiMemoryConsumers[MC_USER].PagesUsed;
-
-    return STATUS_SUCCESS;
+    /* Unimplemented as per https://www.geoffchappell.com/studies/windows/km/ntoskrnl/api/ex/sysinfo/query.htm?tx=76&ts=0,1254 */
```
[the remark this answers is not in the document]
HBelusca: The question is what does the implementation actually does there.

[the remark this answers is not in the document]
ThFabba: Well, it fills a `SYSTEM_MEMORY_INFORMATION` structure, in some way. Seems easy to write a test for, if we actually need this for something. Before this PR we'd return `STATUS_INFO_LENGTH_MISMATCH` to any correct caller, so `STATUS_NOT_IMPLEMENTED` really isn't worse.

ntoskrnl/include/internal/mm.h
```diff
@@ -368,11 +368,14 @@ typedef struct _MMPFN
...
+    } u1;
+    union
+    {
+        PMMPTE PteAddress;
 
         // HACK for ROSPFN
-        SWAPENTRY SwapEntry;
-    } u1;
-    PMMPTE PteAddress;
+        PMM_RMAP_ENTRY RmapListHead;
```
tkreuzer, line 373: What's the reason for moving these around?

ntoskrnl/ex/sysinfo.c
```diff
@@ -1469,10 +1469,10 @@ QSI_DEF(SystemFileCacheInformation)
     RtlZeroMemory(Sci, sizeof(SYSTEM_FILECACHE_INFORMATION));
 
     /* Return the Byte size not the page size. */
-    Sci->CurrentSize = MiMemoryConsumers[MC_USER].PagesUsed; /* FIXME */
-    Sci->PeakSize = MiMemoryConsumers[MC_USER].PagesUsed; /* FIXME */
+    Sci->CurrentSize = 0; /* FIXME */
```
tkreuzer, line 1472: Can we get this fixed properly?

ntoskrnl/mm/freelist.c
```diff
@@ -643,11 +626,7 @@ MmAllocPage(ULONG Type)
     Pfn1->NextLRU = NULL;
     Pfn1->PreviousLRU = NULL;
 
-    if (Type == MC_USER)
-    {
-        Pfn1->u4.MustBeCached = 1; /* HACK again */
-        MmInsertLRULastUserPage(PfnOffset);
-    }
+    MmInsertLRULastUserPage(PfnOffset);
```
tkreuzer, line 629: Is this correct for cache?

tkreuzer (changes requested):
I don't like the CcZeroPage change. And it doesn't look like it's required for the rest of the code.

DarkFire01:
This PR had some unresolved stuff that was pretty important, Please reopen when you're ready to continue!
Thank you!

PR #3883, opened 2021-08-05
replace deprecated hings

fgsoftware1:
## Purpose

- replace some yacc deprecated things


learn-more (changes requested):
Please remove all unrelated changes from this PR.

What does it actually solve?

fgsoftware1:
> Please remove all unrelated changes from this PR.
> 
> What does it actually solve?

it doesn't solve anything it only replace %define pure-parser(deprecated) to %define api.pure

learn-more:
> > Please remove all unrelated changes from this PR.
> > What does it actually solve?
> 
> it doesn't solve anything it only replace %define pure-parser(deprecated) to %define api.pure

Thanks for your contibution,

however it looks like msi is synced to wine:
https://git.reactos.org/?p=reactos.git;a=blob;f=media/doc/WINESYNC.txt;hb=0bc00267df2f644d329a5b0823052293c0108384#l117

We prefer not to create a diff on upstream for things that are not fixing any bugs.


If you want to fix this, the suggested flow would be to submit a patch to wine,
and after that is merged sync our msi with wine's version (completely, not just one file)

fgsoftware1:
ok

PR #3888, opened 2021-08-07
[NTOS][HALPPC] Remove PowerPC code from the kernel and HAL

Extravert-ir:
This is a dead code anyway, was not touched for more than 10 years. The port has less chances to appear even than ARM/ARM64.
And if anybody wants to see the code, it's available forever in GIT history


HBelusca:
I suggest you also edit these two wiki pages (once this commit is merged) to say that the PPC port has been removed from master at (whatever) commit:
https://reactos.org/wiki/ReactOS_ports
https://reactos.org/wiki/PowerPC

learn-more (approved):

SergeGautherie:
Remove even more code? Like:
```
hal/CMakeLists.txt	
   7 elseif(ARCH STREQUAL "powerpc")

ntoskrnl/config/powerpc/cmhardwr.c

ntoskrnl/ntos.cmake	
 373 elseif(ARCH STREQUAL "powerpc")
```

PR #3897, opened 2021-08-11
[NTOS:SE] Document the whole subsystem in Doxygen format

GeoB99:
As the title says, write document comment headers in Doxygen format for all the functions in SRM for once and for all. The documentation headers will be updated regularly when time calls for it. In addition of documenting the subsystem module, update the file headers and annotate the remaining functions with SAL. ~~Currently WIP.~~
# TODO

- [x] access.c
- [x] accesschk.c
- [x] acl.c
- [x] audit.c
- [x] priv.c
- [x] sd.c
- [x] semgr.c
- [x] sid.c
- [x] sqos.c
- [x] srm.c
- [x] token.c

### DIFF
--- a/ntoskrnl/include/internal/se.h
+++ b/ntoskrnl/include/internal/se.h
@@ -36,7 +36,8 @@ typedef struct _TOKEN_AUDIT_POLICY_INFORMATION
 
 FORCEINLINE
 PSID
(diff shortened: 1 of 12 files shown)


Extravert-ir:
I was thinking recently that it's more practical to document functions in header files rather than near implementation. That way you may look through them all in a more compact way

PR #3898, opened 2021-08-11
[NTOS:EX] QSISystem*HandleInformation(): Fix number of handles

SergeGautherie:
## Purpose

Do not increase number of handles, when locking a handle failed.
Plus related optimizations.

JIRA issue: [CORE-11978](https://jira.reactos.org/browse/CORE-11978)

## Proposed changes

- Set number of handles once only, with correct value.
- Set error status once only, when buffer is too small.

And while there:
- Require a full struct, on input.
- Return 0 as required size, if input buffer size is less than minimal.

### DIFF
--- a/ntoskrnl/ex/sysinfo.c
+++ b/ntoskrnl/ex/sysinfo.c
@@ -1303,12 +1303,15 @@ QSI_DEF(SystemHandleInformation)
 
     DPRINT("NtQuerySystemInformation - SystemHandleInformation\n");
 
-    /* Set initial required buffer size */
-    *ReqSize = FIELD_OFFSET(SYSTEM_HANDLE_INFORMATION, Handles);
+    /* Set minimal input buffer size */
+    *ReqSize = sizeof(SYSTEM_HANDLE_INFORMATION);
 
     /* Check user's buffer size */
     if (Size < *ReqSize)
     {
+#if (NTDDI_VERSION < NTDDI_VISTA)
+        *ReqSize = 0;
+#endif
         return STATUS_INFO_LENGTH_MISMATCH;
     }
 
@@ -1324,6 +1327,9 @@ QSI_DEF(SystemHandleInformation)
         DPRINT1("Failed to lock the user buffer: 0x%lx\n", Status);
         return Status;
     }
+
+    /* Set initial required buffer size */
+    *ReqSize = FIELD_OFFSET(SYSTEM_HANDLE_INFORMATION, Handles);
 
     /* Reset of count of handles */
     HandleInformation->NumberOfHandles = 0;
@@ -2503,12 +2509,15 @@ QSI_DEF(SystemExtendedHandleInformation)
 
     DPRINT("NtQuerySystemInformation - SystemExtendedHandleInformation\n");
 
-    /* Set initial required buffer size */
-    *ReqSize = FIELD_OFFSET(SYSTEM_HANDLE_INFORMATION_EX, Handle);
+    /* Set minimal input buffer size */
+    *ReqSize = sizeof(SYSTEM_HANDLE_INFORMATION_EX);
 
     /* Check user's buffer size */
     if (Size < *ReqSize)
     {
+#if (NTDDI_VERSION < NTDDI_VISTA)
+        *ReqSize = 0;
+#endif
         return STATUS_INFO_LENGTH_MISMATCH;
     }
 
@@ -2524,6 +2533,9 @@ QSI_DEF(SystemExtendedHandleInformation)
         DPRINT1("Failed to lock the user buffer: 0x%lx\n", Status);
         return Status;
     }
+
+    /* Set initial required buffer size */
+    *ReqSize = FIELD_OFFSET(SYSTEM_HANDLE_INFORMATION_EX, Handle);
 
     /* Reset of count of handles */
     HandleInformation->Count = 0;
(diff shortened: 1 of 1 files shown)


ntoskrnl/ex/sysinfo.c
```diff
@@ -1362,7 +1356,9 @@ QSI_DEF(SystemHandleInformation)
                     /* Check user's buffer size */
                     if (*ReqSize > Size)
                     {
+                        ExUnlockHandleTableEntry(HandleTable, HandleTableEntry);
                         Status = STATUS_INFO_LENGTH_MISMATCH;
+                        goto Abort;
```
HBelusca, line 1361: Please write an apitest that shows whether this is the behaviour happening on Windows, namely, whether `HandleInformation->NumberOfHandles` will return the number of handles stored in the user buffer, or will return the number of handles in the system, regardless of the size of the user buffer.

HBelusca (dismissed):

ntoskrnl/ex/sysinfo.c
SergeGautherie: Adding that shortcut was wrong.
I'm working on extending ntdll:info winetest, separately.

ntoskrnl/ex/sysinfo.c
```diff
@@ -1303,12 +1303,15 @@ QSI_DEF(SystemHandleInformation)
 
     DPRINT("NtQuerySystemInformation - SystemHandleInformation\n");
 
-    /* Set initial required buffer size */
-    *ReqSize = FIELD_OFFSET(SYSTEM_HANDLE_INFORMATION, Handles);
+    /* Set minimal input buffer size */
+    *ReqSize = sizeof(SYSTEM_HANDLE_INFORMATION);
```
HBelusca, line 1307: What happens on Windows if you call `NtQuerySystemInformation` for that `SystemHandleInformation` class, with the `ReqSize` pointer parameter being NULL?

ntoskrnl/ex/sysinfo.c
```diff
@@ -1350,21 +1353,14 @@ QSI_DEF(SystemHandleInformation)
             if ((HandleTableEntry->Object) &&
                 (HandleTableEntry->NextFreeTableEntry != -2))
             {
-                /* Increase of count of handles */
-                ++HandleInformation->NumberOfHandles;
```
HBelusca, line 1354: Are you sure this gets incremented only if `ExpLockHandleTableEntry` succeeds? (you're now using the `Index` value to hold the number)

ntoskrnl/ex/sysinfo.c
```diff
@@ -1391,12 +1387,13 @@ QSI_DEF(SystemHandleInformation)
...
                             HandleTableEntry->GrantedAccess;
-
-                        ++Index;
                     }
 
                     /* Unlock it */
                     ExUnlockHandleTableEntry(HandleTable, HandleTableEntry);
+
+                    /* Increase count of handles, always */
+                    ++Index;
```
HBelusca, line 1396: `Index` is exclusively used for the index into the `Handles` array of the user-provided `HandleInformation` buffer. Please keep the original intent of the code and separate that from the counting of handles (the `NumberOfHandles` variable) that is also done separately, whatever the size of the user-provided information buffer.

HBelusca:
As part of the tests, please verify that the sysinternals Handle utility: https://docs.microsoft.com/en-us/sysinternals/downloads/handle , as well as the `openfiles` Windows utility, still works on ReactOS.

ntoskrnl/ex/sysinfo.c
```diff
@@ -1303,12 +1303,15 @@ QSI_DEF(SystemHandleInformation)
...
 
-    /* Set initial required buffer size */
-    *ReqSize = FIELD_OFFSET(SYSTEM_HANDLE_INFORMATION, Handles);
+    /* Set minimal input buffer size */
+    *ReqSize = sizeof(SYSTEM_HANDLE_INFORMATION);
 
     /* Check user's buffer size */
     if (Size < *ReqSize)
     {
+#if (NTDDI_VERSION < NTDDI_VISTA)
```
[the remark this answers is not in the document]
SergeGautherie: Yes.

ntoskrnl/ex/sysinfo.c
SergeGautherie: Nothing special: that parameter is optional. (See #3910.)

ntoskrnl/ex/sysinfo.c
SergeGautherie: See Jira issue:
I am not 100% sure, as I do not know how to test this very case.
Nonetheless, it feels more self-consistent, and I assume current winetest was not reported as (intermittently) failing.

ntoskrnl/ex/sysinfo.c
SergeGautherie: I will revert that, though the 2 variables feel redundant after the counting change.

SergeGautherie (commented):
> sysinternals Handle utility [...] `openfiles` Windows utility

I can try: what (steps) should I check more precisely?

ntoskrnl/ex/sysinfo.c
HBelusca: Precisely, and this is why here, if the user sets `ReqSize` to be NULL, now here and in the other places below (and in the original code as well) there would be a NULL pointer dereference.

ntoskrnl/ex/sysinfo.c
GeoB99: When it comes to kernel development, you're either sure of yourself of what you're doing or you aren't. Assumptions are to be avoided.

If you're unsure then it's at best not worth touching the parts of the kernel you're clearly not fond with.

GeoB99 (changes requested):
When it comes to kernel development, you're either sure of yourself of what you're doing or you aren't. Assumptions are to be avoided.

If you're unsure then it's at best not worth touching the parts of the kernel you're clearly not fond with.

PR #3899, opened 2021-08-11
[DSOUND] Improvements to DirectSound(Capture)EnumerateW

oleg-dubinskiy:
## Purpose

Fix and improve `DirectSoundEnumerateW` and `DirectSoundCaptureEnumerateW` implementations in our dsound.
Use the Windows-compatible way to get the device name and pass it to a callback.
This fixes incorrect detection of DirectSound audio input and output devices, so now a lot of apps are able to detect it correctly, and can play the sound properly (e. g., AIMP 2.61 and IcyTower from Rapps). :smiley: 

JIRA issue: [CORE-7535](https://jira.reactos.org/browse/CORE-7535), [CORE-10907](https://jira.reactos.org/browse/CORE-10907), [CORE-15324](https://jira.reactos.org/browse/CORE-15324), [CORE-15533](https://jira.reactos.org/browse/CORE-15533), [CORE-16340](https://jira.reactos.org/browse/CORE-16340) (Pocket Tanks 1.6, Stamina, Dave Gnukem all do improve)

## Proposed changes

- Don't use Wine-specific `DSDRIVERDESC` structure and `DRV_QUERYDSOUNDDESC` message, declared in sdk/include/dxsdk/dsdriver.h, whose are obsolete even in Wine for now.
- Instead, declare a new `WAVEINCAPSW` and `WAVEOUTCAPSW` structures (for input and output appropriately), call `waveInGetDevCapsW` and `WaveOutGetDevCapsW` for enumerated device ID and store retrieved device name in these strucutres.
- Then pass these names to a `lpDSEnumCallbackW` as-is, without Ansi to Unicode conversion (since the retrieved strings are Unicode already).
- Do this both for capture and playback functions.
- Addtionally, add `MMSYSERR_BADDEVICEID` status code to `mmErr` macro, because it also might me returned by `waveIn`/`OutGetDevCapsW` as well, in case of failure.

## Result

As is visible on the following screenshot, after my changes, AIMP 2.61 now detects the output audio device correctly, so it does not show error code -23 at start anymore, and is able to play the audio files properly. :smiley: 
AIMP2-fixed

See demo video as well also:

https://user-images.githubusercontent.com/26385117/129097904-eedd7078-de73-4dc0-ad0d-75352432159a.mp4

In general, I managed to get the sound working at least in my following apps/games: AIMP 2.61, CrystalWolf Audio Player 1.7, IcyTower 1.5.1 from Rapps, PC Radio 6.0.2, Radiocent 3.5 and many more. ^^)

### DIFF
--- a/dll/directx/wine/dsound/dsound_main.c
+++ b/dll/directx/wine/dsound/dsound_main.c
@@ -52,6 +52,7 @@ HRESULT mmErr(UINT err)
 	case MMSYSERR_INVALHANDLE:
 	case WAVERR_STILLPLAYING:
 		return DSERR_GENERIC; /* FIXME */
+	case MMSYSERR_BADDEVICEID:
 	case MMSYSERR_NODRIVER:
 		return DSERR_NODRIVER;
 	case MMSYSERR_NOMEM:
@@ -318,11 +319,9 @@ HRESULT WINAPI DirectSoundEnumerateW(
 	LPVOID lpContext )
 {
     unsigned devs, wod;
-    DSDRIVERDESC desc;
+    WAVEOUTCAPSW capsW;
     GUID guid;
     int err;
-    WCHAR wDesc[MAXPNAMELEN];
-    WCHAR wName[MAXPNAMELEN];
 
     TRACE("lpDSEnumCallback = %p, lpContext = %p\n",
 	lpDSEnumCallback, lpContext);
@@ -337,16 +336,13 @@ HRESULT WINAPI DirectSoundEnumerateW(
     devs = waveOutGetNumDevs();
     if (devs > 0) {
 	if (GetDeviceID(&DSDEVID_DefaultPlayback, &guid) == DS_OK) {
-            static const WCHAR empty[] = { 0 };
 	    for (wod = 0; wod < devs; ++wod) {
                 if (IsEqualGUID( &guid, &DSOUND_renderer_guids[wod] ) ) {
-                    err = mmErr(waveOutMessage(UlongToHandle(wod),DRV_QUERYDSOUNDDESC,(DWORD_PTR)&desc,ds_hw_accel));
+                    err = mmErr(waveOutGetDevCapsW(wod, &capsW, sizeof(WAVEOUTCAPSW)));
                     if (err == DS_OK) {
-                        TRACE("calling lpDSEnumCallback(NULL,\"%s\",\"%s\",%p)\n",
-                              "Primary Sound Driver",desc.szDrvname,lpContext);
-                        MultiByteToWideChar( CP_ACP, 0, "Primary Sound Driver", -1,
-                                             wDesc, sizeof(wDesc)/sizeof(WCHAR) );
-                        if (lpDSEnumCallback(NULL, wDesc, empty, lpContext) == FALSE)
+                        TRACE("calling lpDSEnumCallback(NULL,\"%ls\",\"%ls\",%p)\n",
+                              capsW.szPname,L"Primary Sound Driver",lpContext);
+                        if (lpDSEnumCallback(NULL, capsW.szPname, L"Primary Sound Driver", lpContext) == FALSE)
                             return DS_OK;
 		    }
 		}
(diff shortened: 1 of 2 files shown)


Extravert-ir (changes requested):
`WINESYNC.txt` states dsound.dll is "Synced to Wine-1.3.29". Have we forked that module? Why?
If not, this fix will be lost with a new winesync, as the codebase looks quite different.

I suggest to sync this module with Wine and check if the issue is fixed. Or state that dsound should not be synced at all

oleg-dubinskiy:
> `WINESYNC.txt` states dsound.dll is "Synced to Wine-1.3.29". Have we forked that module? Why?

No, we didn't actually fork it. The "Synced" statement is correct in this case.
I just added some ROS diffs those are separated properly via `#ifdef __REACTOS__`. It does not automatically mean that our dsound becomes forked.

> I suggest to sync this module with Wine and check if the issue is fixed.

I already tried that a few times. It does not makes sense at all, at least until we're re-targeted onto NT6+. The current Wine's dsound completely relies on NT6+ MMDevApi (which we don't currently have properly working), so is incompatible with us, same as with NT5. Ftr, they started switching on MMDevApi since 1.3.30, so it's a reason why we don't sync it to anything newer than 1.3.29. I checked multiple times: after such syncing, DirectSound audio input and output regresses much more: it does no longer work even with those apps, with whose it worked before. Besides impossibility to select the audio device, they can't play the sound at all, even with default (primary) device. Logging spams about missing (unregistered) MMDevApi classes. And anyway, current Wine's MMDevApi depends on Wine-specific registry keys and values (whose are missing on ReactOS/Windows, so it also will not work in any case).

> Or state that dsound should not be synced at all.

OK, I'm not against for changing the statement to "Forked" if needed, but are just a few small ROS diffs enough reason to fork the whole component? :thinking: Can't we just keep it as synced, but with some diffs? I think we can.

CecilPal:
I tried this PR with the latest master source code of ReactOS (1e62771). It makes ReactOS lag a lot when it tries to play the Logon sound when the AC97 Driver from ReactOS Application Manager is installed.

oleg-dubinskiy:
> I tried this PR with the latest master source code of ReactOS (1e62771). It makes ReactOS lag a lot when it tries to play the Logon sound when the AC97 Driver from ReactOS Application Manager is installed.

Those lags are totally random. They cannot be caused by dsound changes, because dsound.dll is not used at all when playing system sound events.
To make sure, just try to delete dsound.dll from system32 directory and reboot. After rebooting, you will same hear logon sound, and nothing will change if the lags will also appear that time.
It proves that dsound.dll is indeed not used in this case, so your guessing is not right.

CecilPal:
Okay. I wonder what is causing the lag then.

oleg-dubinskiy:
> Okay. I wonder what is causing the lag then.

It seems to be a bug(s) in other components of ROS audio subsystem (most of all it's our portcls.sys). Just tested my dsound.dll in Windows XP: it also is working properly on Windows, but no any lags there unlike ReactOS.

JoachimHenze:
The ticket also makes Pocket Tanks 1.6 play music/sound for the first time https://jira.reactos.org/browse/CORE-15324
Although It unhides other issues then in this app, see my comment in https://jira.reactos.org/browse/CORE-15324?focusedCommentId=129295&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-129295 I do still rate it as a legit fix for that ticket! I therefore added it to the listing in PR description.

JoachimHenze:
It also plays sound now in Stamina from rapps, and Dave Gnukem plays sound and music now as well. Great.
I do definitely vote for merging this PR. It's definitely an improvement
And I agree with Oleg, that it's not really needed to update the WINESYNC.txt. We have diff in many other Wine-modules already and they all still only mention the latest syncpoint in that file, and nothing else.
Yes, the diff is properly ifdef'ed which is most likely the best we can do here.
Additionally due to the restructuring in Wine, it is very unlikely anyway, that we'd ever sync further up their stack in this specific case.
What CecilPal mentioned is totally unrelated.

Extravert-ir:
Entries in WINESYNC.txt make sense only for syncing with current/future versions of the code. The fact that it is synchronized with a 10 years old version of Wine does not make any sense now, because that doesn't help synchronizing it with a fresh one.

All that `#ifdef __REACTOS__` thing is needed __only__ for synchronizing with future versions of upstream code. If that's not possible, or there will be no future versions, it does not make any sense.
(there are discussions if this thing is needed at all, but that's a different topic)

> Can't we just keep it as synced, but with some diffs?

As you said, we can't synchronize it, because latest version relies on NT6 APIs, so it is in fact _forked_ already.

So please, move dsound.dll to a forked category in WINESYNC.txt. Would be good to put some words there about why it was forked. We will probably synchronize it once ReactOS becomes compatible with those APIs.
And yeah, `#ifdef __REACTOS__` stuff is not needed anymore here.

oleg-dubinskiy:
OK, done @Extravert-ir. If this will be indeed better, then I'm not against. :slightly_smiling_face:

Extravert-ir (approved):
OK, please format the commit message a bit. You may use `-` for making lists, that looks a lot more readable when there is a list of changes.
See 798fc13b482b56a785b910f7febb396d608e14b0 as an example.

oleg-dubinskiy:
> OK, please format the commit message a bit. You may use `-` for making lists, that looks a lot more readable when there is a list of changes.
> See [798fc13](https://github.com/reactos/reactos/commit/798fc13b482b56a785b910f7febb396d608e14b0) as an example.

Done. :ok_hand:

PR #3903, opened 2021-08-12
[EXPLORER] [RUNONCEEX] Handle RunOnceEx by invoking RunOnceEx in iernonce.dll

kernelbin:
## Purpose

Handle RunOnceEx by invoking RunOnceEx in iernonce.dll

// I've already checked, RunOnceEx is not handled under Safe Mode

Looks like Windows **ONLY** checks if HKLM one exists and invokes RunOnceEx in iernonce.dll
but `iernonce` will handle both HKLM and HKCU.
currently the code imitate what Windows do
But I wonder if this is intended...

### DIFF
--- a/base/shell/explorer/startup.cpp
+++ b/base/shell/explorer/startup.cpp
@@ -3,6 +3,7 @@
  * Copyright (C) 2002 Shachar Shemesh
  * Copyright (C) 2013 Edijs Kolesnikovics
  * Copyright (C) 2018 Katayama Hirofumi MZ
+ * Copyright (C) 2021 He Yang
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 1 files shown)


GeoB99:
> But I wonder if this is intended...

My educated (or uneducated, whatever) guess is that it is Microsoft's intention to firstly check if at least one item task is in the machine key hive variant because we want at least to handle whatever remaining task is during boot stage of the machine, regardless if we haven't initiated a successful logon attempt yet.

Since the comments header states that such keys are handled after a user has fully logon then I'm unsure, perhaps as long as there's no compatibility conflict with handling both HKLM and HKCU at the Shell level (but with HKLM  taking top precedence) then you can go ahead with that.

base/shell/explorer/startup.cpp
```diff
@@ -317,6 +318,71 @@ static BOOL ProcessRunKeys(HKEY hkRoot, LPCWSTR szKeyName, BOOL bDelete,
...
+                           NULL,
+                           NULL,
+                           NULL,
+                           NULL,
+                           NULL);
+
+    if (res != ERROR_SUCCESS)
+    {
+        TRACE("RegQueryInfoKeyW failed on Software\\Microsoft\\Windows\\CurrentVersion\\RunOnceEx (%ld)\n", res);
+
```
GeoB99, line 363: Remove these unnecessary blank lines

kernelbin:
> > But I wonder if this is intended...
> 
> My educated (or uneducated, whatever) guess is that it is Microsoft's intention to firstly check if at least one item task is in the machine key hive variant because we want at least to handle whatever remaining task is during boot stage of the machine, regardless if we haven't initiated a successful logon attempt yet.
> 
> Since the comments header states that such keys are handled after a user has fully logon then I'm unsure, perhaps as long as there's no compatibility conflict with handling both HKLM and HKCU at the Shell level (but with HKLM taking top precedence) then you can go ahead with that.

just... Windows really didn't handle HKCU if HKLM is empty XD
I guess it's better to simulate Windows?

GeoB99:
I'm fine either way.

GeoB99 (approved):

HBelusca (commented):
:+1:

PR #3905, opened 2021-08-13
[OSK] On Screen Keyboard overhaul

peterooch:
## Primary changes:
Convert from dialog to an ordinary window, placing buttons according to data in keyboard.c
Character button text is now according to selected keyboard layout instead of system language.
Buttons will update* automatically when there is a layout switch (and when shift/capslock are pressed).
(Updating is most reliable when the keyboard is focused)
Removed keyboard dialog templates from resource files, they live on in keyboard.c

### Secondary changes:
Support for font change, currently saved in registry is face name and height.
Additional registry load/save functions for strings to facilitate the above change.

### Notes:
There is some buggy behavior on Windows (tested on 10) when keyboard change happens when the window is unfocused so it does not update, this does not happen in ROS (diff in handling background WM_PAINT messages).
Clicking the shift/capslock buttons (**not keys**) does not update the layout correctly (would appreciate any help with that).
There is a hack in the WM_TIMER message handler because WM_INPUTLANGCHANGE is not being handled properly in ROS.

### The new on-screen keyboard in action
https://user-images.githubusercontent.com/3612577/129358249-4d712c55-bdad-4dff-8d3e-d09ace8e7e86.mp4


peterooch:
Pinging @GeoB99

base/applications/osk/keyboard.c
```diff
@@ -0,0 +1,251 @@
...
+ * PROJECT:         ReactOS On-Screen Keyboard
+ * LICENSE:         GPL - See COPYING in the top level directory
+ * PURPOSE:         Keyboard layout data.
+ * COPYRIGHT:       Baruch Rutman (peterooch at gmail dot com)
+ * 
+ * This is partialy based on the original en-US dialog templates.
+ */
+#include "precomp.h"
+
+KEY EnhancedKeyboardKeys[] =
```
GeoB99, line 11: I am not sure if you're aware or not but certain keys like ALT are localized, depending on the locale selected within the system and whatnot. Right now you're hardcoding these keys and regardless of what locale you choose for the main keyboard keys, the rest of the keys are still remained as is. Is that intended?

base/applications/osk/settings.c
```diff
@@ -126,35 +222,49 @@ VOID LoadSettings(VOID)
     Globals.PosX = CW_USEDEFAULT;
     Globals.PosY = CW_USEDEFAULT;
 
+    /* Set font value defaults */
+    Globals.FontHeight = 15;
```
GeoB99, line 226: Is it possible to assign meaningful macro names for such values rather than hardcoding them?

base/applications/osk/main.c
```diff
@@ -116,78 +117,262 @@ DWORD WINAPI OSK_WarningDlgThread(LPVOID lpParameter)
...
     OSKIcon = LoadImageW(Globals.hInstance, MAKEINTRESOURCEW(IDI_OSK), IMAGE_ICON, 0, 0, LR_DEFAULTSIZE);
 
     /* Load the strings into the "About" dialog */
-    LoadStringW(Globals.hInstance, STRING_OSK, szTitle, countof(szTitle));
-    LoadStringW(Globals.hInstance, STRING_AUTHORS, szAuthors, countof(szAuthors));
+    LoadStringW(Globals.hInstance, STRING_AUTHORS, szAuthors, _countof(szAuthors));
 
     /* Finally, execute the "About" dialog by using the Shell routine */
-    ShellAboutW(Globals.hMainWnd, szTitle, szAuthors, OSKIcon);
+    ShellAboutW(NULL, Globals.szTitle, szAuthors, OSKIcon);
```
GeoB99, line 130: If you set the parent window to NULL, the "About" dialog box will be an orphaned window dialog. Is that what you really want?

base/applications/osk/main.c
```diff
@@ -457,6 +672,9 @@ BOOL OSK_DlgCommand(WPARAM wCommand, HWND hWndControl)
         PlaySoundW(MAKEINTRESOURCEW(IDI_SOUNDCLICK), GetModuleHandle(NULL), SND_RESOURCE | SND_ASYNC);
     }
 
+    /* FIXME */
+    PostMessageW(Globals.hMainWnd, OSK_KEYPRESS, MapVirtualKeyW(wCommand & SCANCODE_MASK, MAPVK_VSC_TO_VK), 0);
```
GeoB99, line 676: Fix me but fix me what? Please provide a more meaningful description comment of what should be fixed in the foreseeable future.

base/applications/osk/main.c
```diff
@@ -732,42 +893,65 @@ INT_PTR APIENTRY OSK_DlgProc(HWND hDlg, UINT msg, WPARAM wParam, LPARAM lParam)
...
-            break;
+            return 0;
 
         case WM_THEMECHANGED:
             /* Redraw the dialog (and its control buttons) using the new theme */
-            InvalidateRect(hDlg, NULL, FALSE);
+            InvalidateRect(hwnd, NULL, FALSE);
+            return 0;
+
+
```
GeoB99, line 932:
```suggestion

```

base/applications/osk/main.c
```diff
@@ -797,7 +982,7 @@ int WINAPI wWinMain(HINSTANCE hInstance,
     if (hMutex)
     {
         /* Check if there's already a mutex for the program */
-        dwError = GetLastError();
+        DWORD dwError = GetLastError();
```
GeoB99, line 985: Why this change? Revert this.

GeoB99 (commented):
Initial review...

base/applications/osk/main.c
peterooch: It was crashing with the window handle set as the parent

base/applications/osk/main.c
peterooch: `Clicking the shift/capslock buttons (not keys) does not update the layout correctly (would appreciate any help with that).`
Clicking buttons does the opposite of what is intended

base/applications/osk/settings.c
peterooch: I can create a macro or just leave it as 0 and let the GDI do as it wills, what would be the preferred approach?

base/applications/osk/keyboard.c
peterooch: I have noticed that some keys are translated in some translations, I guess I could add more string entries, although if there is a way to extract it from the keyboard layout would be better

petr-akhlamov:
I of course, not developer, but can you please make window size as now - little. On screens with low resolution such osk takes half-screen. I not demand and which solution will accept developers don't know.

HBelusca:
> Convert from dialog to an ordinary window

What advantage does this change provide?

peterooch:
> What advantage does this change provide?
I worked on this on and off for some time and it was the method I found to work best when I started working on this.

base/applications/osk/settings.c
GeoB99: I'm more comfortable if you create a macro specifically for this.

base/applications/osk/main.c
```diff
@@ -822,32 +1008,50 @@ int WINAPI wWinMain(HINSTANCE hInstance,
...
+    wc.hbrBackground = (HBRUSH)(COLOR_WINDOW + 1);
+    /* Set the application's icon */
+    wc.hIcon = LoadImageW(Globals.hInstance, MAKEINTRESOURCEW(IDI_OSK), IMAGE_ICON, 0, 0, LR_SHARED | LR_DEFAULTSIZE);
+    wc.hIconSm = CopyImage(wc.hIcon, IMAGE_ICON, GetSystemMetrics(SM_CXSMICON), GetSystemMetrics(SM_CYSMICON), LR_COPYFROMRESOURCE);
+
+    if (!RegisterClassExW(&wc))
+        goto quit;
+
+    if (!LoadStringW(Globals.hInstance, STRING_OSK, Globals.szTitle, _countof(Globals.szTitle)))
+        StringCchCopyW(Globals.szTitle, _countof(Globals.szTitle), L"On-Screen Keyboard");
```
GeoB99, line 1027: I am not keen of hard-coding the window title only in English locale when there's `STRING_OSK` for each translation file instead. Please use that.

GeoB99 (changes requested):

base/applications/osk/main.c
peterooch: it's fallback if `LoadStringW` fails to load the relevant string from the resources

GeoB99: Aaaaf my eyes failed hard on this one, my bad lol. Still, the possibility for `LoadStringW` to fail loading the string title from resource is slim to none. So I think having this fallback is quite redundant (unless you provide a reasonable explanation as of how this function cal fail, and if such failure rate is higher than it should be).

HBelusca: At some point people will add failure checks for stuff like `2+2` ...

binarymaster:
> @petr-akhlamov wrote:
> Of course I'm not a developer, but can you please leave the window size small.

Well, I understand that there is still a demand for small UI, however the new look of this window looks pretty good to me and even much better than before, so I'd accept this.

Size problem can be solved later by making this window resizable, with all controls and indicators being responsive (automatically adapting to the changed size). But it's something out of this PR scope I think.

HBelusca:
For information, here's the look of Windows OSK on Windows 2000 (top) and Windows Server 2003 (bottom):
image

HBelusca:
Here is how the current (master) ReactOS OSK looks like:
image

As we see, the size of the buttons is smaller than what is being shown on the PR description's video. And therefore, if this is a regression, it would be nice to have this fixed.

peterooch:
Script used for rescaling if anybody is interested:
https://gist.github.com/peterooch/0aafff20dce4e6e61a792355438c29d3

HBelusca:
@peterooch Looks great now! I wonder whether the Enter key in the numpad could also use the same arrow as the enter key of the "main" keyboard? 🤔

peterooch:
> I wonder whether the Enter key in the numpad could also use the same arrow as the enter key of the "main" keyboard?  

Anyone else for or against this proposed change?

binarymaster:
> I wonder whether the Enter key in the numpad could also use the same arrow as the enter key of the "main" keyboard? :thinking: 

See https://en.wikipedia.org/wiki/Enter_key

image

There is a difference between Enter and Return keys.

HBelusca:
> See https://en.wikipedia.org/wiki/Enter_key
> 
> There is a difference between Enter and Return keys.

I see.... Here I've a keyboard where both keys have the symbol, that's why I proposed that idea.

base/applications/osk/lang/ar-DZ.rc
```diff
@@ -61,11 +61,23 @@ END
...
+    IDS_PRN "Prn"
+    IDS_STOP "Stop"
+    IDS_ATTN "Attn"
+    IDS_INSERT "ins"
+    IDS_NUMLOCKKEY "nlk"
+    IDS_DELETE "del"
+    IDS_END "end"
+    IDS_CTRL "ctrl"
+    IDS_LEFTALT "alt"
+    IDS_RIGHTALT "alt"
```
HBelusca, line 80: hmm... Why adding translations for these? Aren't they already "translated" (and localized) from the keyboard layouts?

peterooch: `ToUnicode` does not work with these keys

HBelusca: Then how did you manage to get the screenshots (from yesterday) before you added this commit today?

peterooch:
osk_180821_1900
Welp trying to use `GetKeyNameTextW` gave mixed results...

HBelusca:
> `GetKeyNameTextW`

Yeah that API is more like for having "human-readable" name at the expense of the shortness of length of the name.

HBelusca:
> Welp trying to use `GetKeyNameTextW` gave mixed results...

Looking at osk file dump, it appears they indeed hardcode these names ("bksp", etc etc.) within osk itself... 🤔 
And by the way, on Windows (7+?) 10, they seem they could use that API for some of the keys:

image

base/applications/osk/lang/ar-DZ.rc
peterooch: The key texts are in the keyboard structs for defaults in case if ToUnicode fails (for characters) and for system keys, they are just directly being taken (if translate field in KEY struct is FALSE).
(The file itself was initially was just a scripted conversion from the dialog templates)

base/applications/osk/main.c
```diff
@@ -116,78 +131,281 @@ DWORD WINAPI OSK_WarningDlgThread(LPVOID lpParameter)
...
+
+    /* Get key states before doing anything */
+    if (!GetKeyboardState(bKeyStates))
+    {
+        DPRINT("OSK_SetKeys(): GetKeyboardState() call failed.\n");
+        return -1;
+    }
+
+    /* Keyboard language/caps change, just update the button texts */
+    if (reason == SETKEYS_LANG)
```
GeoB99, line 197: These `reason` values can be wrapped in a switch...case body code.

base/applications/osk/main.c
```diff
@@ -116,78 +131,281 @@ DWORD WINAPI OSK_WarningDlgThread(LPVOID lpParameter)
...
+                (Globals.bAlwaysOnTop ? HWND_TOPMOST : HWND_NOTOPMOST),
+                0,
+                0,
+                Globals.Keyboard->Size.cx,
+                Globals.Keyboard->Size.cy + y_pad,
+                SWP_NOMOVE);
+    
+    /* Create LEDs */
+    POINT LedPos  = Globals.Keyboard->LedStart;
+    SIZE LedSize = Globals.Keyboard->LedSize;
```
GeoB99, line 299: Please respect the current style of the code where variables are declared at the top first.

GeoB99:
> There is some buggy behavior on Windows (tested on 10) when keyboard change happens when the window is unfocused so it does not update, this does not happen in ROS (diff in handling background WM_PAINT messages).

Does this happen only in Windows 10 or also in other versions of Windows?

> Clicking the shift/capslock buttons (not keys) does not update the layout correctly (would appreciate any help with that).

Does not update the layout correctly as in what, the layout updating is not synchronized with each time the buttons you mentioned are pressed? The fact that the buttons are all windows now maybe within WM_TIMER message if such buttons are pressed as you mentioned, then force a layout update as per the state of these buttons?

peterooch:
> > There is some buggy behavior on Windows (tested on 10) when keyboard change happens when the window is unfocused so it does not update, this does not happen in ROS (diff in handling background WM_PAINT messages).
> 
> Does this happen only in Windows 10 or also in other versions of Windows?
> 
Will check on windows 2003/xp

> > Clicking the shift/capslock buttons (not keys) does not update the layout correctly (would appreciate any help with that).
> 
> Does not update the layout correctly as in what, the layout updating is not synchronized with each time the buttons you mentioned are pressed? The fact that the buttons are all windows now maybe within WM_TIMER message if such buttons are pressed as you mentioned, then force a layout update as per the state of these buttons?

Hmm will try that

peterooch:
Changed the logic behind key updates, now it just updates the key buttons anyway instead of checking various conditions.
The only side effect that I observed is in ROS that there is at times the button text 'blink', on the other hand now clicking caps/shift updates accordingly.

base/applications/osk/keyboard.c
```diff
@@ -0,0 +1,251 @@
+/*
+ * PROJECT:         ReactOS On-Screen Keyboard
+ * LICENSE:         GPL - See COPYING in the top level directory
```
GeoB99, line 3: https://reactos.org/wiki/Coding_Style#File_Structure

base/applications/osk/main.c
```diff
@@ -53,6 +54,20 @@ int OSK_SetImage(int IdDlgItem, int IdResource)
     return TRUE;
 }
 
+void OSK_SetText(int IdDlgItem, int IdResource)
```
GeoB99, line 57: Missing function comment header

GeoB99 (changes requested):
Capture
Capture2
ReactOS' Standard Keyboard style is not supposed to look like that (there shouldn't be the numpad there). See the W2k3's OSK for comparison.

base/applications/osk/keyboard.c
peterooch: This was copied from `main.c` to keep consistency

peterooch:
>  ReactOS' Standard Keyboard style is not supposed to look like that (there shouldn't be the numpad there). See the W2k3's OSK for comparison.
This feels out of the scope here, the original layout of the resources was with the numpad and it didn't bother anybody until now...

GeoB99:
> This feels out of the scope here, the original layout of the resources was with the numpad and it didn't bother anybody until now...

Capture
In case you still don't understand, I'M TALKING about the Standard layout and not the Enhanced one (this is OSK from the current master). Maybe don't do unrelated changes that don't need to be changed?

peterooch:
>In case you still don't understand, I'M TALKING about the Standard layout and not the Enhanced one (this is OSK from the current master, the numpad is actually hidden there).

Can you just not get angry about a misunderstanding?

JoachimHenze:
A potential backport of this PR would require having a fix for the regression https://jira.reactos.org/browse/CORE-18407 that it introduced.

PR #3909, opened 2021-08-14
[FASTFAT_NEW] "Fix" Flags bitfield to avoid RTC error

tkreuzer:
Since VS 16.11 the compiler sometimes emits calls to _RTC_UninitUse, when parts of a bitfield are initialized (See https://developercommunity.visualstudio.com/t/Broken-runtime-checks-with-CL-19293013/1503629). Fix this by using an ULONG instead of a bitfield.
Note: The structure uses a 24 bit bitfield plus an UCHAR, which is supposed to form a 32 bit field, but that doesn't work anyway.

### DIFF
--- a/drivers/filesystems/fastfat_new/fatstruc.h
+++ b/drivers/filesystems/fastfat_new/fatstruc.h
@@ -1370,8 +1370,12 @@ typedef struct _CCB {
     //  Define a 24bit wide field for Flags, but a UCHAR for Wild Cards Present
     //  since it is used so often.  Line these up on byte boundaries for grins.
     //
-
+#ifdef __REACTOS__
+    ULONG Flags; // Due to https://developercommunity.visualstudio.com/t/Broken-runtime-checks-with-CL-19293013/1503629
+    // Note: the following BOOLEAN will not be packed anyway!
+#else
     ULONG Flags:24;
+#endif
     BOOLEAN ContainsWildCards;
 
     //


drivers/filesystems/fastfat_new/fatstruc.h
```diff
@@ -1371,7 +1371,8 @@ typedef struct _CCB {
     //  since it is used so often.  Line these up on byte boundaries for grins.
     //
 
-    ULONG Flags:24;
+    ULONG Flags; //  :24; // Due to https://developercommunity.visualstudio.com/t/Broken-runtime-checks-with-CL-19293013/1503629
+    // Note: this will not be packed anyway!
```
binarymaster, line 1375: This either needs `#ifdef __REACTOS__` or to be sent upstream.

HBelusca: Maybe that by sending upstream they will fix it (either in fastfat or in the compiler side) quicker 🤔

HBelusca (approved):

drivers/filesystems/fastfat_new/fatstruc.h
tkreuzer: The code was already hacked, when it was imported.

tkreuzer: Ok, fixed

drivers/filesystems/fastfat_new/fatstruc.h
```diff
@@ -1370,9 +1370,13 @@ typedef struct _CCB {
...
-
-    ULONG Flags:24;
-    BOOLEAN ContainsWildCards;
+#ifdef __REACTOS__
+    ULONG Flags; // Due to https://developercommunity.visualstudio.com/t/Broken-runtime-checks-with-CL-19293013/1503629
+    // Note: the following BOOLEAN will not be packed anyway!
+#else
+    ULONG Flags; //  :24;
+#endif
+   BOOLEAN ContainsWildCards;
```
SergeGautherie, line 1379:
```suggestion

#ifdef __REACTOS__
    ULONG Flags; // Due to https://developercommunity.visualstudio.com/t/Broken-runtime-checks-with-CL-19293013/1503629
    // Note: the following BOOLEAN will not be packed anyway!
#else
    ULONG Flags:24;
#endif
    BOOLEAN ContainsWildCards;
```

SergeGautherie:
> Note: The structure uses a 24 bit bitfield plus an UCHAR, which is supposed to form a 32 bit field, but that doesn't work anyway.

Maybe upstream could use a `union` to "pack" them?

GeoB99 (approved):

binarymaster (approved):
Please fix the ros diff (Serge suggested a valid suggestion for it above), but otherwise LGTM.

Extravert-ir (approved):

drivers/filesystems/fastfat_new/fatstruc.h
```diff
@@ -1370,8 +1370,12 @@ typedef struct _CCB {
     //  Define a 24bit wide field for Flags, but a UCHAR for Wild Cards Present
     //  since it is used so often.  Line these up on byte boundaries for grins.
     //
-
```
SergeGautherie, line 1373: Why do you remove this blank line?

zefklop (approved):

drivers/filesystems/fastfat_new/fatstruc.h
wiktorek140: In my opinion its good move, because comment line above is about variable that had be changed, so it clearly mean that this block of code is connected to comment higher

PR #3910, opened 2021-08-14
[TASKMGR] Handles: Get number only, skip data

SergeGautherie:
## Purpose

Use less (useless) resources.

## Proposed changes

- Call `NtQuerySystemInformation()` once only, with a minimal struct only.
- Save number of handles only.

### DIFF
--- a/base/applications/taskmgr/perfdata.c
+++ b/base/applications/taskmgr/perfdata.c
@@ -44,7 +44,7 @@ LARGE_INTEGER                              liOldSystemTime = {{0,0}};
 SYSTEM_PERFORMANCE_INFORMATION             SystemPerfInfo;
 SYSTEM_BASIC_INFORMATION                   SystemBasicInfo;
 SYSTEM_FILECACHE_INFORMATION               SystemCacheInfo;
-SYSTEM_HANDLE_INFORMATION                  SystemHandleInfo;
+ULONG                                      SystemNumberOfHandles;
 PSYSTEM_PROCESSOR_PERFORMANCE_INFORMATION  SystemProcessorTimeInfo = NULL;
 PSID                                       SystemUserSid = NULL;
 
@@ -186,7 +186,7 @@ void PerfDataRefresh(void)
     SYSTEM_PERFORMANCE_INFORMATION             SysPerfInfo;
     SYSTEM_TIMEOFDAY_INFORMATION               SysTimeInfo;
     SYSTEM_FILECACHE_INFORMATION               SysCacheInfo;
-    LPBYTE                                     SysHandleInfoData;
+    SYSTEM_HANDLE_INFORMATION                  SysHandleInfoData;
     PSYSTEM_PROCESSOR_PERFORMANCE_INFORMATION  SysProcessorTimeInfo;
     double                                     CurrentKernelTime;
     PSECURITY_DESCRIPTOR                       ProcessSD;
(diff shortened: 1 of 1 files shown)


base/applications/taskmgr/perfdata.c
```diff
@@ -221,22 +221,13 @@ void PerfDataRefresh(void)
...
-
-        if (status == STATUS_INFO_LENGTH_MISMATCH) {
-            HeapFree(GetProcessHeap(), 0, SysHandleInfoData);
-        }
-
-    } while (status == STATUS_INFO_LENGTH_MISMATCH);
+    status = NtQuerySystemInformation(SystemHandleInformation, &SysHandleInfoData, sizeof(SysHandleInfoData), NULL);
+    // On unexpected error, reuse previous value.
+    // STATUS_SUCCESS (0-1 handle) should never happen.
+    if (status != STATUS_INFO_LENGTH_MISMATCH)
```
HBelusca, line 229: Granted this should never happen, but you can always check for the success as well.

base/applications/taskmgr/perfdata.c
```diff
@@ -221,22 +221,13 @@ void PerfDataRefresh(void)
...
-        status = NtQuerySystemInformation(SystemHandleInformation, SysHandleInfoData, BufferSize, &ulSize);
-
-        if (status == STATUS_INFO_LENGTH_MISMATCH) {
-            HeapFree(GetProcessHeap(), 0, SysHandleInfoData);
-        }
-
-    } while (status == STATUS_INFO_LENGTH_MISMATCH);
+    status = NtQuerySystemInformation(SystemHandleInformation, &SysHandleInfoData, sizeof(SysHandleInfoData), NULL);
+    // On unexpected error, reuse previous value.
+    // STATUS_SUCCESS (0-1 handle) should never happen.
```
HBelusca, line 228: You can put these comments with the other ones above.

base/applications/taskmgr/perfdata.c
```diff
@@ -44,7 +44,7 @@ LARGE_INTEGER                              liOldSystemTime = {{0,0}};
 SYSTEM_PERFORMANCE_INFORMATION             SystemPerfInfo;
 SYSTEM_BASIC_INFORMATION                   SystemBasicInfo;
 SYSTEM_FILECACHE_INFORMATION               SystemCacheInfo;
-SYSTEM_HANDLE_INFORMATION                  SystemHandleInfo;
+ULONG                                      SystemNumberOfHandles;
```
learn-more, line 47: This needs to be initialized, otherwise it will be garbage if `NtQuerySystemInformation` fails

SergeGautherie: Globals default to `0`, don't they?

base/applications/taskmgr/perfdata.c
SergeGautherie: I could, if you insist, though I would rather leave them next to the very code they apply to.

base/applications/taskmgr/perfdata.c
SergeGautherie: I could, if you insist, though that would seem useless code bloat and somewhat misleading to me.

base/applications/taskmgr/perfdata.c
binarymaster, line 228: Consistency with other comments in this file.
```suggestion
    /* On unexpected error, reuse previous value.
     * STATUS_SUCCESS (0-1 handle) should never happen.
     */
```

binarymaster: I think it's okay to have these comments just above the status check code line.

SergeGautherie: Done.

base/applications/taskmgr/perfdata.c
learn-more: Static's do, others don't.

SergeGautherie: (File) Globals, whether `static` or not, behave same as `static` (function) locals, don't they?

learn-more: It does indeed seem that file globals still have static storage without the static keyword.

binarymaster (approved):

PR #3911, opened 2021-08-15
[DSOUND] Don't force status to failure by default in primarybuffer_SetFormat

oleg-dubinskiy:
## Purpose

Don't force `HRESULT` status to `DSERR_BUFFERLOST` by default in `primarybuffer_SetFormat` function. Assign to it `S_OK` instead, as done in current Wine: https://source.winehq.org/git/wine.git/blob/HEAD:/dlls/dsound/primary.c#l440.
That variable is already set to failure in other appropriate place, when this is indeed needed: https://git.reactos.org/?p=reactos.git;a=blob;f=dll/directx/wine/dsound/primary.c;h=b04a526929e4e817ff63aa652530e5901a3253a6;hb=HEAD#l511.
The assigned failure status caused executing failure handling code even when the fail actually didn't happen. In particular, `DSOUND_ReopenDevice` failed on `waveOutOpen` call when calling this was not needed.
After my changes, it's called only when needed, i. e., only when the actual failure happens. Otherwise, if all goes good, that code is not called, so the data is transferred successfully and the function is working correctly.
It fixes non-working playback in QMMP 0.11.1 (via DirectSound ouput plugin) and AIMP 3.10 and newer versions. Also it improves the DirectSound test results from MS DxDiag: now all 4 available tests are passing successfully, instead of only 2. Other dsound issues also might be fixed as well. :smiley: 

JIRA issue: [CORE-10907](https://jira.reactos.org/browse/CORE-10907), [CORE-14783](https://jira.reactos.org/browse/CORE-14783)

### DIFF
--- a/dll/directx/wine/dsound/primary.c
+++ b/dll/directx/wine/dsound/primary.c
@@ -459,7 +459,7 @@ LPWAVEFORMATEX DSOUND_CopyFormat(LPCWAVEFORMATEX wfex)
 
 HRESULT primarybuffer_SetFormat(DirectSoundDevice *device, LPCWAVEFORMATEX wfex)
 {
-	HRESULT err = DSERR_BUFFERLOST;
+	HRESULT err = S_OK;
 	int i;
 	DWORD nSamplesPerSec, bpp, chans;
 	LPWAVEFORMATEX oldpwfx;


wiktorek140 (approved):
It look good. Probably like labbeled trivial fix but if fix audio cases, its worth of merging ;)

dll/directx/wine/dsound/primary.c
```diff
@@ -459,7 +459,7 @@ LPWAVEFORMATEX DSOUND_CopyFormat(LPCWAVEFORMATEX wfex)
 
 HRESULT primarybuffer_SetFormat(DirectSoundDevice *device, LPCWAVEFORMATEX wfex)
 {
-	HRESULT err = DSERR_BUFFERLOST;
+	HRESULT err = S_OK;
```
GeoB99, line 462: There's `DS_OK` which is the same as S_OK but strictly exclusive to DirectSound code. See https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=%23define+DS_OK

oleg-dubinskiy: Indeed, it exists. And seems to be more logically suitable here. But on my opinion, I'd keep the current one, at least to have less diffs with Wine, moreover if these values have the same effect. I think even if we forked dsound from Wine already, it still makes no sense to add additional unneeded diffs, when we can avoid using them.

GeoB99 (approved):
I am approving this PR for the mere sake that it fixes the aforementioned issues in the description, giving the benefit of the doubt. 

Don't take this approval 100% for granted as it's better to wait for others devs their judgement regarding this.

binarymaster (approved):
The change looks pretty trivial and I think I'm going to merge this today evening (some hours later), if nobody has any objections.

PR #3915, opened 2021-08-17
[NTUSER] Improve NtUserQueryWindow

katahiromz:
## Purpose

## Proposed changes

- Improve `NtUserQueryWindow` function.
- Add `IMC` structure.
- Modify `THREADINFO` structure.

## TODO

- [x] Do tests.

### DIFF
--- a/win32ss/user/ntuser/win32.h
+++ b/win32ss/user/ntuser/win32.h
@@ -75,6 +75,8 @@ typedef struct _W32THREAD
     PVOID pUMPDObj;
 } W32THREAD, *PW32THREAD;
 
+struct tagIMC;
+
 #ifdef __cplusplus
 typedef struct _THREADINFO : _W32THREAD
 {
@@ -126,6 +128,10 @@ typedef struct _THREADINFO
     INT                 iCursorLevel;
     /* Last message cursor position */
     POINT               ptLast;
+    /* Input context-related */
+    struct _WND *       spwndDefaultIme;
+    struct tagIMC *     spDefaultImc;
+    HKL                 hklPrev;
 
     INT                 cEnterCount;
     /* Queue of messages posted to the queue. */
--- a/win32ss/user/ntuser/win32.h
+++ b/win32ss/user/ntuser/win32.h
@@ -129,8 +129,8 @@ typedef struct _THREADINFO
     /* Last message cursor position */
     POINT               ptLast;
     /* Input context-related */
-    struct _WND *       spwndDefaultIme;
-    struct tagIMC *     spDefaultImc;
+    struct _WND*        spwndDefaultIme;
+    struct tagIMC*      spDefaultImc;
     HKL                 hklPrev;
 
     INT                 cEnterCount;
--- a/win32ss/user/ntuser/window.c
+++ b/win32ss/user/ntuser/window.c
@@ -4017,7 +4017,7 @@ NtUserQueryWindow(HWND hWnd, DWORD Index)
          {
              pwndActive = gpqForeground->spwndActive;
              pti = PsGetCurrentThreadWin32Thread();
-             if (pwndActive && pti->rpdesk == pwndActive->head.rpdesk)
+             if (pti->rpdesk == pwndActive->head.rpdesk)
              {
                 ptiActive = pwndActive->head.pti;
                 if (ptiActive && ptiActive->spwndDefaultIme)
--- a/win32ss/user/ntuser/window.c
+++ b/win32ss/user/ntuser/window.c
@@ -3934,6 +3934,8 @@ NtUserQueryWindow(HWND hWnd, DWORD Index)
 
    PWND pWnd;
    DWORD_PTR Result;
+   PWND pwndActive;
+   PTHREADINFO pti;
    DECLARE_RETURN(UINT);
 
    TRACE("Enter NtUserQueryWindow\n");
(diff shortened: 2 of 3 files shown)


win32ss/user/ntuser/win32.h
```diff
@@ -126,6 +128,10 @@ typedef struct _THREADINFO
     INT                 iCursorLevel;
     /* Last message cursor position */
     POINT               ptLast;
+    /* Input context-related */
+    struct _WND *       spwndDefaultIme;
+    struct tagIMC *     spDefaultImc;
```
HBelusca, line 133:
```suggestion
    struct _WND*        spwndDefaultIme;
    struct tagIMC*      spDefaultImc;
```

win32ss/user/ntuser/window.c
```diff
@@ -3996,19 +3997,27 @@ NtUserQueryWindow(HWND hWnd, DWORD Index)
          Result = (pWnd->head.pti->MessageQueue == gpqForeground);
          break;
 
-      case QUERY_WINDOW_DEFAULT_IME:
-         ERR("QUERY_WINDOW_DEFAULT_IME: FIXME\n");
-         Result = 0;
+      case QUERY_WINDOW_DEFAULT_IME: /* default IME window */
+         Result = (DWORD_PTR)UserHMGetHandle(pWnd->head.pti->spwndDefaultIme);
```
HBelusca, line 4001: The `spwndDefaultIme` may not be initialized yet (and therefore can be NULL), and therefore you _**must**_ check whether it is NULL, before calling the `UserHMGetHandle` function. See what is done e.g. for the `QUERY_WINDOW_FOCUS` case above.

win32ss/user/ntuser/window.c
```diff
@@ -3996,19 +3997,27 @@ NtUserQueryWindow(HWND hWnd, DWORD Index)
...
          break;
 
-      case QUERY_WINDOW_DEFAULT_ICONTEXT:
-         ERR("QUERY_WINDOW_DEFAULT_ICONTEXT: FIXME\n");
-         Result = 0;
+      case QUERY_WINDOW_DEFAULT_ICONTEXT: /* default input context handle */
+         Result = (DWORD_PTR)UserHMGetHandle(pWnd->head.pti->spDefaultImc);
```
HBelusca, line 4005: Same comment.

win32ss/user/ntuser/window.c
```diff
@@ -3996,19 +3997,27 @@ NtUserQueryWindow(HWND hWnd, DWORD Index)
...
          Result = 0;
+         if (gpqForeground && gpqForeground->spwndActive)
+         {
+             pwndActive = gpqForeground->spwndActive;
+             pti = PsGetCurrentThreadWin32Thread();
+             if (pwndActive && pti->rpdesk == pwndActive->head.rpdesk)
+             {
+                ptiActive = pwndActive->head.pti;
+                if (ptiActive)
+                   Result = (DWORD_PTR)UserHMGetHandle(ptiActive->spwndDefaultIme);
```
HBelusca, line 4018: Same comment.

HBelusca (changes requested):

win32ss/user/ntuser/win32.h
katahiromz: OK.

win32ss/user/ntuser/window.c
katahiromz: Added NULL check.

win32ss/user/ntuser/window.c
katahiromz: Added NULL check.

win32ss/user/ntuser/window.c
katahiromz: Added NULL check.

win32ss/user/ntuser/window.c
```diff
@@ -3996,19 +3997,33 @@ NtUserQueryWindow(HWND hWnd, DWORD Index)
...
          break;
 
       case QUERY_WINDOW_ACTIVE_IME:
-         ERR("QUERY_WINDOW_ACTIVE_IME: FIXME\n");
          Result = 0;
+         if (gpqForeground && gpqForeground->spwndActive)
+         {
+             pwndActive = gpqForeground->spwndActive;
+             pti = PsGetCurrentThreadWin32Thread();
+             if (pwndActive && pti->rpdesk == pwndActive->head.rpdesk)
```
SergeGautherie, line 4020:
```suggestion
             if (pti && pti->rpdesk == pwndActive->head.rpdesk)
```

SergeGautherie (changes requested):

win32ss/user/ntuser/window.c
HBelusca: And no check of whether `pwndActive` is null?

katahiromz: Sir, what's your command?

I thought `pwndActive` null check was necessary due to multithread. Shall I add `pti` null check?

HBelusca: @katahiromz Do as it is done in the other places of the win32k code where `spwndActive` is being used.

SergeGautherie: > Shall I add `pti` null check?

I assumed there was a typo, but, if `pti` null check is useless here, then do not add it either.

HBelusca: > > no check of whether `pwndActive` is null?
> 
> Previous `if` already checks that, doesn't it?
> 

Hehe thanks, indeed it does it (the check that is done together with `gpqForeground`)

> > multithread
> 
> If value(s) can change, then I guess that is a bigger problem...

win32k is single-threaded in the sense that, as we can see, all the exported functions are protected with the global lock (see e.g. `UserEnterShared()` etc.), so that if two(or more) separate user-mode threads call win32k functions at the "same time", all except one call gets to wait its own turn.
And for internal win32k functions that run by themselves (w/o being called from user-mode), if they are designed correctly, they should also lock whenever they are going to access a global, or some data that belongs to an exposed object (like windows, desktops, etc.)

Here, `NtUserQueryWindow()` satisfies this: it uses the global lock (see the `UserEnterShared()` call) and therefore this should protect e.g. `gpqForeground` and its data (e.g. `gpqForeground->spwndActive`), plus the other global/user-exposed stuff touched there, to be modified under the hood.
So for example, the capture of the `spwndActive` into `ptiActive` is not necessary (but it just makes the lines of code later being shorter, which is nice)...

HBelusca: Regarding the `pti` NULL check, this is redundant. Indeed, **any** win32 thread that calls **any** win32k-exported function (the `NtUser...` etc. ones), are automatically transformed into a Win32 thread, from the point of view of win32k, before proceeding further. This involves the creation + initialization of what is going to be returned by `PsGetCurrentThreadWin32Thread`,  **if it doesn't exist already**, so that, in win32k, when being called within any `NtUser...` function, `PsGetCurrentThreadWin32Thread` **always** returns a valid (and not NULL) pointer to a valid block of data.

SergeGautherie: @HBelusca, thanks for the detailed explanations.

@katahiromz, simply remove the extra `pwndActive` check.

katahiromz: Ok.

PR #3919, opened 2021-08-18
[IMM32] Implement ImmWINNLSGetIMEHotkey

katahiromz:
## Purpose

## Proposed changes

- Implement `ImmWINNLSGetIMEHotkey` function.
- Modify `imm32.spec`.

### DIFF
--- a/dll/win32/imm32/imm.c
+++ b/dll/win32/imm32/imm.c
@@ -5066,6 +5066,15 @@ Quit:
     return NtUserGetImeInfoEx(pImeInfoEx, SearchType);
 }
 
+/***********************************************************************
+ *              ImmWINNLSGetIMEHotkey (IMM32.@)
+ */
+UINT WINAPI ImmWINNLSGetIMEHotkey(HWND hwndIme)
+{
+    TRACE("(%p)\n", hwndIme);
+    return 0;
+}
+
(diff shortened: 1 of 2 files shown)


HBelusca:
"_**Implement**_ ImmWINNLSGetIMEHotkey function" ? Looks more like stubbed to me! 🤣

dll/win32/imm32/imm.c
```diff
@@ -5066,6 +5066,15 @@ ImmGetImeInfoEx(PIMEINFOEX pImeInfoEx,
     return NtUserGetImeInfoEx(pImeInfoEx, SearchType);
 }
 
+/***********************************************************************
+ *              ImmWINNLSGetIMEHotkey (IMM32.@)
+ */
+UINT WINAPI ImmWINNLSGetIMEHotkey(HWND hwndIme)
+{
+    TRACE("(%p)\n", hwndIme);
+    return 0;
```
HBelusca, line 5075: OK actually if you know since which Windows version this function just returns 0, you should mention this there, so that we know it's intentional, and is not a "stub to be implemented later in ReactOS" 😄

katahiromz: Comment added.

katahiromz: Comment added.

PR #3921, opened 2021-08-19
[USER32_DYNAMIC_APITEST] Make failure checks more consistent/explicit

SergeGautherie:
## Purpose

Minor improvements to test, while core issue remains.

JIRA issue: [CORE-10559](https://jira.reactos.org/browse/CORE-10559)

## Proposed changes

- Skip test if user32.dll is already loaded, as on Vista+.
- Rewrite 'last error' handling, to make it consistent/explicit.
- Double-check 'uninitialized' cases, with FreeLibrary().

### DIFF
--- a/modules/rostests/apitests/user32_dynamic/load.c
+++ b/modules/rostests/apitests/user32_dynamic/load.c
@@ -3,6 +3,7 @@
  * LICENSE:         LGPLv2.1+ - See COPYING.LIB in the top level directory
  * PURPOSE:         Test for loading/unloading user32.dll
  * PROGRAMMERS:     Thomas Faber <thomas.faber@reactos.org>
+ *                  Serge Gautherie <reactos-git_serge_171003@gautherie.fr>
  */
 
 #include <apitest.h>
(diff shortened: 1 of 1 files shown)


modules/rostests/apitests/user32_dynamic/load.c
```diff
@@ -18,104 +19,129 @@ START_TEST(load)
...
     ok(Peb->KernelCallbackTable == NULL, "KernelCallbackTable = %p\n", Peb->KernelCallbackTable);
     ok(Peb->PostProcessInitRoutine == NULL, "PostProcessInitRoutine = %p\n", Peb->PostProcessInitRoutine);
 
-    SetLastError(12345);
+    /* 1st load */
+    SetLastError(0xdeadbeef);
     hUser32 = LoadLibraryW(L"user32");
-    Error = GetLastError();
-    ok(hUser32 != NULL, "LoadLibrary failed\n");
-    ok(Error != 12345, "Error = %lu\n", Error);
```
ThFabba, line 30: Why remove these tests? Is behavior inconsistent between Windows versions?
(and as a rule, _this_ is the kind of thing that belongs in the PR description and/or commit message. Not just "rewrite", but "rewrite to fix problem xyz")

modules/rostests/apitests/user32_dynamic/load.c
```diff
@@ -18,104 +19,129 @@ START_TEST(load)
...
 
     /* Before init */
+    SetLastError(0xdeadbeef);
     hUser32 = GetModuleHandleW(L"user32");
-    ok(hUser32 == NULL, "hUser32 = %p\n", hUser32);
+    Error = GetLastError();
+
+    if (hUser32 != NULL)
+    {
+        win_skip("user32.dll is already loaded since Vista\n");
```
ThFabba, line 28: Any clue why? Does `msvcrt` import it maybe? If we know how it gets pulled in, we might be able to fix that.

katahiromz:
This program of master is failing as follows:
image
image

SergeGautherie (commented):
> This program of master is failing as follows

Interesting:
- Your XP fails like ReactOS, unlike Test WHS and WTB XP to S2003.
- Your 10 does not have user32 already loaded, unlike WTB Vista to 10.

I think these results should be discussed on Jira rather than here, especially as I assume my PR does not change them.

HBelusca:
> Interesting:
> 
>     * Your XP fails like ReactOS, unlike Test WHS and WTB XP to S2003.
> 
>     * Your 10 does not have user32 already loaded, unlike WTB Vista to 10.
> 
> 
> I think these results should be discussed on Jira rather than here, especially as I assume my PR does not change them.

What is WTB ?
Also, his XP is japanese and therefore, the IME is always loaded (`SRVINFO_IMM32` always set when user32 is loaded), and this is the same behaviour as ReactOS currently.
On contrast, the standard Win2003 (our Test WHS + possibly what you call WTB) is english and doesn't load the IME by default, but on-demand. I strongly suspect that if you had installed support for CJK and activated by default the advanced text services, the IME would then load by default and you would get the same behaviour as @katahiromz 's XP installation + current ReactOS.

SergeGautherie:
> What is WTB ?

Wine Test Bot.

> Also, his XP is japanese and therefore, the IME is always loaded (`SRVINFO_IMM32` always set when user32 is loaded), and this is the same behaviour as ReactOS currently.

Yes, Japanese might be the trigger. You suggest that user32 stays loaded because it loaded imm32, right?
(Meanwhile, it seems that would not explain opposite results on Vista+.)

--

With regard to this very PR, I suggest to remove the 'since Vista' log part and replace it with a comment.
What do you think?

SergeGautherie:
> You suggest that user32 stays loaded because it loaded imm32, right?

Additional test results on WTB:
* XP-S2003: Loading user32 does load imm32 too. Freeing either user32 (as currently done) or imm32 (as a test) unloads them both, as in "removes the extra reference".
I'll make the test explicit about that.
* Vista+: Even freeing user32 10 times at start does not unload it.
Investigating Vista+ (or Japanese cases) further is out of scope of this PR.

PR #3930, opened 2021-08-24
[CMAKE][BOOTDATA]  Fix root CA certificates for Live CD

julcar:
## Purpose

Starting with commit 6158207c318bccaf1767d1a5c1545de45beb113d, caroots.inf is no longer added to LiveCD image builds, this turns into a trouble when a LiveCD user tries to load TLS/SSL sites using our built-in secure channel, and what they find is that the CA chains cannot be generated and thus failing.

JIRA issue: [CORE-17739](https://jira.reactos.org/browse/CORE-17739)

## Proposed changes

1) Add caroots.inf contents directly to the registry hives when building the LiveCD ISO.
2) Do not include the file caroots.inf in the LiveCD ISO, because it is added directly into the registry.


binarymaster (approved):
Looks good for me. :+1:  I've squashed your changes into single commit removing useless merge commits from your branch :wink:

HBelusca (approved):

learn-more (approved):

PR #3934, opened 2021-08-29
[FASTFAT_NEW] Fix uninitialized use

mrmks04:
## Purpose

Fix uninitialized use

### DIFF
--- a/drivers/filesystems/fastfat_new/create.c
+++ b/drivers/filesystems/fastfat_new/create.c
@@ -629,7 +629,7 @@ Return Value:
 
     ULONG MatchFlags = 0;
 
-    CCB LocalCcb;
+    CCB LocalCcb = {0};
     UNICODE_STRING Lfn;
     UNICODE_STRING OrigLfn = {0};
 


SergeGautherie:
Where is it used uninitialized?
This should be submitted upstream...

mrmks04:
> 
> 
> Where is it used uninitialized?
> This should be submitted upstream...

uninit_use

SergeGautherie:
> uninit_use

I don't know what/how triggered this failure, but it is obviously wrong:
an assignment is not a (uninit) use!

If the failure can be reproduced, file a Jira ticket...

NB:
This PR reminds me of #3679...

HBelusca:
Ahh yes.... Actually this is due to a new bug in the RTC implementation of Visual Studio, see this discussion on Mattermost:
image
and the relevant links:
https://developercommunity.visualstudio.com/t/Broken-runtime-checks-with-CL-19293013/1503629
and the older PR that fixed the problem: PR #3909 .

tkreuzer:
Issue is fixed by PR #3909

PR #3936, opened 2021-08-31
[PORTCLS] Implement CPortPinWavePci::DeviceIoControl

oleg-dubinskiy:
## Purpose

Make a several improvements into `CPortPinWavePci::DeviceIoControl` method and other related code.
This is the first of two improvements required by some audio drivers on real hardware to play the sound properly (like Vinyl AC97 audio driver).

JIRA issue: [CORE-17742](https://jira.reactos.org/browse/CORE-17742)

## Proposed changes

- Refactor CPortPinWavePci::DeviceIoControl function:
  - use `switch` construction instead of `if`-`else` for handling IO control codes;
  - implement some more IOCTLs for whose appreopriate functions are already exist;
  - return success instead of unsuccessful status. TODO: perhaps more logical here is using `STATUS_NOT_SUPPORTED` instead?
- Implement `CPortPinWavePci::HandleKsReset` method for `IOCTL_KS_RESET_STATE` control code and call it in `DeviceIoControl`.
- Declare a new global `m_ResetState` variable, which manages `KSRESET`* states. `CPortPinWavePci::HandleKsReset` handles it. Initialize it in `CPortPinWavePci::Init`, similarly to pin wavecyclic code. Use it in `CPortPinWavePci::RequestService` and `CPortPinWavePci::HandleKsStream`.

## TODO

- [ ] The sound still is not playing after my changes. I did not figure out yet: is it due to something wrong in my implementation or something another is still causing a problem? :thinking:

### DIFF
--- a/drivers/wdm/audio/backpln/portcls/pin_wavepci.cpp
+++ b/drivers/wdm/audio/backpln/portcls/pin_wavepci.cpp
@@ -4,6 +4,7 @@
  * FILE:            drivers/wdm/audio/backpln/portcls/pin_wavepci.cpp
  * PURPOSE:         WavePci IRP Audio Pin
  * PROGRAMMER:      Johannes Anderwald
+ *                  Oleg Dubinskiy
  */
 
 #include "private.hpp"
@@ -74,9 +75,12 @@ protected:
 
     KSALLOCATOR_FRAMING m_AllocatorFraming;
 
+    KSRESET m_ResetState;
+
     LONG m_Ref;
 
     NTSTATUS NTAPI HandleKsProperty(IN PIRP Irp);
+    NTSTATUS NTAPI HandleKsReset(IN PIRP Irp);
     NTSTATUS NTAPI HandleKsStream(IN PIRP Irp);
 };
 
@@ -460,14 +464,13 @@ CPortPinWavePci::TerminatePacket()
     return STATUS_SUCCESS;
 }
 
-
 VOID
 NTAPI
 CPortPinWavePci::RequestService()
 {
     PC_ASSERT_IRQL(DISPATCH_LEVEL);
 
-    if (m_State == KSSTATE_RUN)
+    if (m_State == KSSTATE_RUN && m_ResetState == KSRESET_END)
     {
         m_Stream->Service();
         //TODO
@@ -544,6 +547,45 @@ CPortPinWavePci::HandleKsProperty(
 
 NTSTATUS
 NTAPI
+CPortPinWavePci::HandleKsReset(
+    IN PIRP Irp)
+{
+    KSRESET ResetValue;
+    NTSTATUS Status;
+
+    Status = KsAcquireResetValue(Irp, &ResetValue);
+    DPRINT("Status %x Value %u\n", Status, ResetValue);
+    /* check for success */
+    if (NT_SUCCESS(Status))
+    {
+        //determine state of reset request
+        if (ResetValue == KSRESET_BEGIN)
+        {
+            // start reset process
+            // incoming read/write requests will be rejected
+            m_ResetState = KSRESET_BEGIN;
+
+            // cancel existing buffers
+            m_IrpQueue->CancelBuffers();
+        }
+        else if (ResetValue == KSRESET_END)
+        {
+            // end of reset process
+            m_ResetState = KSRESET_END;
+        }
+    }
+
+    if (Status != STATUS_PENDING)
+    {
+        Irp->IoStatus.Status = Status;
+        IoCompleteRequest(Irp, IO_NO_INCREMENT);
+    }
+
+    return Status;
+}
+
+NTSTATUS
+NTAPI
 CPortPinWavePci::HandleKsStream(
     IN PIRP Irp)
 {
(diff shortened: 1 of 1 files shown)


drivers/wdm/audio/backpln/portcls/pin_wavepci.cpp
```diff
@@ -542,6 +545,45 @@ CPortPinWavePci::HandleKsProperty(
...
+CPortPinWavePci::HandleKsReset(
+    IN PIRP Irp)
+{
+    KSRESET ResetValue;
+    NTSTATUS Status;
+
+    Status = KsAcquireResetValue(Irp, &ResetValue);
+    DPRINT("Status %x Value %u\n", Status, ResetValue);
+    /* check for success */
+    if (NT_SUCCESS(Status))
```
Extravert-ir, line 559: Obvious comments again :)

drivers/wdm/audio/backpln/portcls/pin_wavepci.cpp
```diff
@@ -588,22 +646,33 @@ CPortPinWavePci::DeviceIoControl(
...
-    }
+        case IOCTL_KS_PROPERTY:
+            return HandleKsProperty(Irp);
 
-    UNIMPLEMENTED;
+        case IOCTL_KS_ENABLE_EVENT:
+            return PcHandleEnableEventWithTable(Irp, &m_Descriptor);
+
+        case IOCTL_KS_DISABLE_EVENT:
+            return PcHandleDisableEventWithTable(Irp, &m_Descriptor);
```
Extravert-ir, line 658: Are you sure these new IOCTLs should be handled here? I see them handled already in `CPortPinWaveCyclic::DeviceIoControl`

drivers/wdm/audio/backpln/portcls/pin_wavepci.cpp
```diff
@@ -588,22 +646,33 @@ CPortPinWavePci::DeviceIoControl(
...
+
+        case IOCTL_KS_RESET_STATE:
+            return HandleKsReset(Irp);
+
+        case IOCTL_KS_READ_STREAM:
+        case IOCTL_KS_WRITE_STREAM:
+            return HandleKsStream(Irp);
+
+        default:
+            return KsDefaultDeviceIoCompletion(DeviceObject, Irp);
```
Extravert-ir, line 668: Why using a function from ksfilter driver?

drivers/wdm/audio/backpln/portcls/pin_wavepci.cpp
```diff
@@ -588,22 +646,33 @@ CPortPinWavePci::DeviceIoControl(
...
+    }
 
     Irp->IoStatus.Information = 0;
```
Extravert-ir, line 671: The control never reaches here, you have a `return` statement in every case of `switch`

Extravert-ir (changes requested):

drivers/wdm/audio/backpln/portcls/pin_wavepci.cpp
oleg-dubinskiy: As I see, the code from pin_wavecyclic.cpp is not used at all in case with Vinyl AC97 driver.
So I'm not sure exactly about these two IOCTLs, but the driver definitely needs `IOCTL_KS_RESET` (it is called at least).
Although I think if the code from pin_wavepci.cpp and pin_wavecyclic.cpp is used separately for different audio drivers, then additional IOCTLs will not interfere in this case.

drivers/wdm/audio/backpln/portcls/pin_wavepci.cpp
oleg-dubinskiy: Because I didn't find an appropriate equivalent in portcls itself or in kernel. Is there any? :thinking:

drivers/wdm/audio/backpln/portcls/pin_wavepci.cpp
oleg-dubinskiy: Indeed. So I should use an `NTSTATUS` variable.
I tried to use it from the start, but it caused a bugcheck 0x44 (MULTIPLE_IRP_COMPLETE_REQUESTS), so then it probably needs to remove other IRP completions from all sub-functions called from `DeviceIoControl`.

ThFabba: You can just move the `default` case out of the switch. Alternatively you can probably remove this code and the compiler shouldn't complain because it's smart enough to know that every switch case returns.

drivers/wdm/audio/backpln/portcls/pin_wavepci.cpp
ThFabba: You already handle most of the cases from `KsDefaultDeviceIoCompletion` here. Seems like you'd just need a case for `IOCTL_KS_METHOD` that completes the IRP with `STATUS_PROPSET_NOT_FOUND`, and otherwise complete it with `STATUS_INVALID_DEVICE_REQUEST`.

drivers/wdm/audio/backpln/portcls/pin_wavepci.cpp
```diff
@@ -542,6 +545,45 @@ CPortPinWavePci::HandleKsProperty(
...
+            m_IrpQueue->CancelBuffers();
+        }
+        else if (ResetValue == KSRESET_END)
+        {
+            // end of reset process
+            m_ResetState = KSRESET_END;
+        }
+    }
+
+    if (Status != STATUS_PENDING)
```
ThFabba, line 578: STATUS_PENDING can't happen here so I'd recommend against this check, to avoid giving the impression that it can. Perhaps assert `Status != STATUS_PENDING`.

PR #3938, opened 2021-09-01
[NTOS:SE] Implement token effective option upon duplication

GeoB99:
This implements the effective only option where the caller may want to duplicate an access token but only with enabled parts of it, the privileges and groups respectively. Also write tests that exercise the behavior of making the token as effective only.

### DIFF
--- a/modules/rostests/apitests/ntdll/CMakeLists.txt
+++ b/modules/rostests/apitests/ntdll/CMakeLists.txt
@@ -20,6 +20,7 @@ list(APPEND SOURCE
     NtCreateThread.c
     NtDeleteKey.c
     NtDuplicateObject.c
+    NtDuplicateToken.c
     NtFreeVirtualMemory.c
     NtImpersonateAnonymousToken.c
     NtLoadUnloadKey.c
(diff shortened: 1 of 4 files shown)


GeoB99:
**Windows Server 2003**
Capture
**ReactOS**
Capture2
With token effective option implemented in code, here we are un-hiding two symptoms. First of all, in ReactOS we are still having 3 privileges left in the token instead of 2 as in Windows and that remnant privilege is `SeImpersonatePrivilege` which is what we wanted to disable it as test before we're going to duplicate the token as effective.

This privilege has an attribute of `SE_PRIVILEGE_ENABLED_BY_DEFAULT` which means the system automatically assigns the privilege to an object by default since the dawn of the object itself. As it stands out, `NtAdjustPrivilegesToken` and friends don't seem to disable privileges enabled by default because we aren't handling the bitmask flag `SE_PRIVILEGE_ENABLED_BY_DEFAULT` case. I will come up with a PR with a fix shortly for this.

As for groups/users apart that we don't include all the specific groups in the token unlike in Windows, we seem to remove a "disabled" group because its attributes is 0 for whatever reason and as such I took some quick glimpse to see what group or user is that and it seems to be the administrator user. This is something that has to be investigated later for a future PR anyway.

GeoB99:
Capture
I did some re-thinking of the privilege loop and now it can properly remove disabled privileges.

PR #3939, opened 2021-09-01
[NTDLL_APITEST] Implement NtAdjustPrivilegesToken testcase

GeoB99:
~~This is a patch that fixes a symptom that's been unveiled with #3938, where a caller can definitely disable a privilege that's been enabled by default whereas we only disable privileges that have been enabled manually with functions like `NtAdjustPrivilegesToken` and friends.~~

Implement `NtAdjustPrivilegesToken` API testcase that exercises the behavior of disabling an enabled by default privilege.

### DIFF
--- a/modules/rostests/apitests/ntdll/NtAdjustPrivilegesToken.c
+++ b/modules/rostests/apitests/ntdll/NtAdjustPrivilegesToken.c
@@ -1,0 +1,87 @@
+/*
+ * PROJECT:         ReactOS API tests
+ * LICENSE:         GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:         Tests for the NtAdjustPrivilegesToken API
+ * COPYRIGHT:       Copyright 2021 George Bișoc <george.bisoc@reactos.org>
+ */
+
+#include "precomp.h"
+
+static
+BOOL
+IsPrivilegeEnabled(
+    _In_ HANDLE TokenHandle,
+    _In_ ULONG Privilege)
+{
+    PRIVILEGE_SET PrivSet;
+    BOOL Result, Success;
+    LUID Priv;
+
+    ConvertPrivLongToLuid(Privilege, &Priv);
+
+    PrivSet.PrivilegeCount = 1;
+    PrivSet.Control = PRIVILEGE_SET_ALL_NECESSARY;
+    PrivSet.Privilege[0].Luid = Priv;
+    PrivSet.Privilege[0].Attributes = 0;
+
+    Success = PrivilegeCheck(TokenHandle, &PrivSet, &Result);
+    if (!Success)
+    {
+        skip("Failed to check the privilege (error code: %lu)\n", GetLastError());
+        return FALSE;
+    }
+
+    return Result;
+}
+
+static
+VOID
+AdjustEnableDefaultPriv(VOID)
+{
+    NTSTATUS Status;
+    HANDLE Token;
+    TOKEN_PRIVILEGES Priv;
+    BOOL Success, IsEnabled;
+    LUID PrivLuid;
+
+    Success = OpenProcessToken(GetCurrentProcess(),
+                               TOKEN_READ | TOKEN_ADJUST_PRIVILEGES,
+                               &Token);
+    if (!Success)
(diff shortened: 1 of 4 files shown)


GeoB99:
**Windows Server 2003**
Capture
**ReactOS**
Capture2

modules/rostests/apitests/ntdll/NtAdjustPrivilegesToken.c
```diff
@@ -0,0 +1,97 @@
...
+ * LICENSE:         GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:         Tests for the NtAdjustPrivilegesToken API
+ * COPYRIGHT:       Copyright 2021 George Bișoc <george.bisoc@reactos.org>
+ */
+
+#include "precomp.h"
+
+static
+LUID
+ConvertPrivLongToLuid(
```
HBelusca, line 12: Seems to be a candidate for a simple macro 🤔

ntoskrnl/se/token.c
```diff
@@ -4303,7 +4303,7 @@ SepAdjustPrivileges(
                     /* Copy SE_PRIVILEGE_ENABLED | SE_PRIVILEGE_REMOVED */
                     NewAttributes = NewState[j].Attributes;
                     NewAttributes &= (SE_PRIVILEGE_ENABLED | SE_PRIVILEGE_REMOVED);
-                    NewAttributes |= Token->Privileges[i].Attributes & ~SE_PRIVILEGE_ENABLED;
+                    NewAttributes |= Token->Privileges[i].Attributes & ~(SE_PRIVILEGE_ENABLED | SE_PRIVILEGE_ENABLED_BY_DEFAULT);
```
HBelusca, line 4306: What happens to the test results if you don't modify this line (and the other modified one below) ?
I'm not convinced this is a correct patch. This enabled_by_default flag acts like an informative flag for telling that said privilege, by default, is enabled. But this privilege can have also been disabled manually by someone, and the remaining presence of the enabled_by_default flag should not influence in any case the detection of whether a privilege is currently enabled or disabled.

HBelusca (changes requested):

ntoskrnl/se/token.c
GeoB99: Ok, it looks like ReactOS can clearly disable privileges enabled by default without problems even without these lines changed. Guess I have had a brain fart moment misunderstanding that SE_PRIVILEGE_ENABLED_BY_DEFAULT is also an indicator alongside with SE_PRIVILEGE_ENABLED where a privilege is disabled or enabled in the kernel. I will refactor this PR and only provide a testcase at this point and rethink the loops in #3938.

modules/rostests/apitests/ntdll/precomp.h
```diff
@@ -41,4 +41,12 @@ SetupLocale(
     _In_ ULONG OemCode,
     _In_ ULONG Unicode);
 
+#define ConvertPrivLongToLuid(PrivilegeVal, ConvertedPrivLuid) \
+{                                                              \
```
HBelusca, line 45:
```suggestion
do {                                                           \
```

modules/rostests/apitests/ntdll/precomp.h
```diff
@@ -41,4 +41,12 @@ SetupLocale(
...
+    LUID Luid;                                                 \
+    Luid.LowPart = PrivilegeVal;                               \
+    Luid.HighPart = 0;                                         \
+    *ConvertedPrivLuid = Luid;                                 \
+}
```
HBelusca, line 50:
```suggestion
} while (0)
```

HBelusca (approved):
LGTM (+ two suggestions in order to make the macro usable in a function-like manner).

modules/rostests/apitests/ntdll/NtAdjustPrivilegesToken.c
ThFabba: I disagree -- if anything, it's a complicated macro. What advantage does the macro give you?
The function gives you type safety, not having to worry about multiline-macro-do-while stuff, and correct syntax highlighting.

PR #3940, opened 2021-09-02
[PORTCLS] Implement CIrpQueue::GetAcquiredTagRange

oleg-dubinskiy:
## Purpose

Implement `CIrpQueue::GetAcquiredTagRange` method in IRP code of our portcls driver.
Also mark `FirstTag` and `LastTag` parameters output instead of input, since they are actually used as output parameters. Update the prototype in the header appropriately.
This method is called as well by `PinWavePciState` function from pin wavepci code, which is used by some audio drivers on real hardware (like Vinyl AC97 audio driver).

JIRA issue: [CORE-17742](https://jira.reactos.org/browse/CORE-17742)

## TODO

- [ ] The sound still does not work after my changes, so this implementation probably needs to be improved.

### DIFF
--- a/drivers/wdm/audio/backpln/portcls/irpstream.cpp
+++ b/drivers/wdm/audio/backpln/portcls/irpstream.cpp
@@ -4,6 +4,7 @@
  * FILE:            drivers/wdm/audio/backpln/portcls/irpstream.cpp
  * PURPOSE:         IRP Stream handling
  * PROGRAMMER:      Johannes Anderwald
+ *                  Oleg Dubinskiy
  */
 
 #include "private.hpp"
@@ -779,23 +780,50 @@ CIrpQueue::GetCurrentIrpOffset()
 BOOLEAN
 NTAPI
 CIrpQueue::GetAcquiredTagRange(
-    IN PVOID * FirstTag,
-    IN PVOID * LastTag)
+    OUT PVOID * FirstTag,
+    OUT PVOID * LastTag)
 {
     KIRQL OldLevel;
     BOOLEAN Ret = FALSE;
-    //PIRP Irp;
-    //PLIST_ENTRY CurEntry;
-    //PKSSTREAM_DATA StreamData;
+    PIRP Irp;
+    PLIST_ENTRY CurEntry;
+    PKSSTREAM_DATA StreamData;
 
     // lock list
     KeAcquireSpinLock(&m_IrpListLock, &OldLevel);
 
-    // initialize to zero
-    *FirstTag = NULL;
-    *LastTag = NULL;
+    CurEntry = m_IrpList.Flink;
 
-    UNIMPLEMENTED;
+    while (CurEntry != &m_IrpList)
+    {
+        // get irp from list entry
+        Irp = (PIRP)CONTAINING_RECORD(CurEntry, IRP, Tail.Overlay.ListEntry);
+
+        // get stream data
+        StreamData = (PKSSTREAM_DATA)Irp->Tail.Overlay.DriverContext[STREAM_DATA_OFFSET];
+
+        if (StreamData->StreamHeaderIndex == 0)
+        {
+            // store the first tag from stream header index
+            *FirstTag = StreamData->Tags[StreamData->StreamHeaderIndex].Tag;
+        }
+
+        // increment stream header index
+        StreamData->StreamHeaderIndex++;
+
+        if (StreamData->StreamHeaderIndex == StreamData->StreamHeaderCount)
+        {
+            // store the last tag from stream header index
+            *LastTag = StreamData->Tags[StreamData->StreamHeaderIndex].Tag;
+
+            // done
+            Ret = TRUE;
+            break;
+        }
+
+        // move to the next entry
+        CurEntry = CurEntry->Flink;
+    }
 
     // release lock
     KeReleaseSpinLock(&m_IrpListLock, OldLevel);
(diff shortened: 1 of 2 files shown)


drivers/wdm/audio/backpln/portcls/irpstream.cpp
```diff
@@ -779,23 +780,50 @@ CIrpQueue::GetCurrentIrpOffset()
...
+            // store the last tag from stream header index
+            *LastTag = StreamData->Tags[StreamData->StreamHeaderIndex].Tag;
+
+            // done
+            Ret = TRUE;
+            break;
+        }
+
+        // move to the next entry
+        CurEntry = CurEntry->Flink;
```
Extravert-ir, line 825: Comments here are useless, you can remove them without loosing anything :)

drivers/wdm/audio/backpln/portcls/irpstream.cpp
```diff
@@ -779,23 +780,50 @@ CIrpQueue::GetCurrentIrpOffset()
...
 
-    // initialize to zero
-    *FirstTag = NULL;
-    *LastTag = NULL;
+    CurEntry = m_IrpList.Flink;
+
+    while (CurEntry != &m_IrpList)
+    {
+        // get irp from list entry
+        Irp = (PIRP)CONTAINING_RECORD(CurEntry, IRP, Tail.Overlay.ListEntry);
```
Extravert-ir, line 800: Pointer cast is not needed

drivers/wdm/audio/backpln/portcls/irpstream.cpp
```diff
@@ -779,23 +780,50 @@ CIrpQueue::GetCurrentIrpOffset()
...
 
     // lock list
     KeAcquireSpinLock(&m_IrpListLock, &OldLevel);
 
-    // initialize to zero
-    *FirstTag = NULL;
-    *LastTag = NULL;
+    CurEntry = m_IrpList.Flink;
+
+    while (CurEntry != &m_IrpList)
```
Extravert-ir, line 797: This is C++, don't we have some kind of an iterator for lists?

drivers/wdm/audio/backpln/portcls/irpstream.cpp
```diff
@@ -4,6 +4,7 @@
  * FILE:            drivers/wdm/audio/backpln/portcls/irpstream.cpp
  * PURPOSE:         IRP Stream handling
  * PROGRAMMER:      Johannes Anderwald
+ *                  Oleg Dubinskiy
```
Extravert-ir, line 7: If you want to add yourself here, please format the headed according to our current coding style

Extravert-ir (changes requested):
Other than that seems good to go

drivers/wdm/audio/backpln/portcls/irpstream.cpp
```diff
@@ -779,23 +780,50 @@ CIrpQueue::GetCurrentIrpOffset()
 BOOLEAN
 NTAPI
 CIrpQueue::GetAcquiredTagRange(
-    IN PVOID * FirstTag,
-    IN PVOID * LastTag)
+    OUT PVOID * FirstTag,
+    OUT PVOID * LastTag)
```
Extravert-ir, line 784:
```suggestion
    OUT PVOID *FirstTag,
    OUT PVOID *LastTag)
```

drivers/wdm/audio/backpln/portcls/irpstream.cpp
oleg-dubinskiy: :thinking: I don't remember anything like that.

drivers/wdm/audio/backpln/portcls/irpstream.cpp
```diff
@@ -779,23 +780,50 @@ CIrpQueue::GetCurrentIrpOffset()
...
+
+        if (StreamData->StreamHeaderIndex == 0)
+        {
+            // store the first tag from stream header index
+            *FirstTag = StreamData->Tags[StreamData->StreamHeaderIndex].Tag;
+        }
 
-    UNIMPLEMENTED;
+        // increment stream header index
+        StreamData->StreamHeaderIndex++;
```
ThFabba, line 812: It doesn't seem like you should modify this. It seems to indicate a "current position" of some type, whereas this operation seems like it should be read-only. This is based on a cursory reading, so I may be wrong.

drivers/wdm/audio/backpln/portcls/irpstream.cpp
```diff
@@ -779,23 +780,50 @@ CIrpQueue::GetCurrentIrpOffset()
...
+
+        if (StreamData->StreamHeaderIndex == StreamData->StreamHeaderCount)
```
ThFabba, line 814: You don't need to loop the list if you're only interested in the first and last item.
Though it feels like these should be "smallest" and "largest" more so than first/last in the list? But I don't have a good understanding of what these tags mean, so maybe not.

ThFabba: Looked at this more, and the concept of first/last is correct the way you have it.
But I think what you want is the `Tags[0]` on the first IRP, and `Tags[StreamHeaderIndex - 1]` on the last IRP where `StreamHeaderIndex` is not zero. So something like
```
FirstTagSet = FALSE;
while (...)
{
    Irp = ...;
    StreamData = ...;
    if (StreamData->StreamHeaderIndex == 0)
    {
        break;
    }
    if (!FirstTagSet)
    {
        *FirstTag = StreamData->Tags[0].Tag;
        FirstTagSet = TRUE;
    }
    *LastTag = StreamData->Tags[StreamData->StreamHeaderIndex - 1].Tag;
    ...
}
```
Or maybe you have to go to `StreamHeaderCount` and actually do a loop. That part's not clear to me -- `ReleaseMappingWithTag` seems to look up to Index on the current IRP, up to Count on others... not sure what that's about.

PR #3943, opened 2021-09-05
[SDK:ASM][NDK] Improve and synchronize the definition of the SYNCH_LEVEL symbol.

HBelusca:
[SDK:ASM][NDK] Improve and synchronize the definition of the SYNCH_LEVEL symbol.

In particular, for a given platform (x86, AMD64...) its definition
depends for uniprocessor vs. multiprocessor NT kernel/HAL (controlled
by the CONFIG_SMP define).

Regarding the ASM-generated headers (from asm/ksx.template.h), since
they are generated once for a given platform, and since UP/MP NT builds
are generated at the same time, these headers need to contain both
SYNCH_LEVEL definitions controlled by CONFIG_SMP, instead of a fixed
value. Therefore we need to employ the geninc RAW(...) construct.

[NDK:X86] Add missing x86 CLOCK1_LEVEL and POWER_LEVEL definitions.

[SDK:ASM][NDK] Adjust the x86 definition of PCR (for UP and MP).


tkreuzer:
Makes sense. But what about just using the original definition (i.e. simply using `CONSTANT(SYNCH_LEVEL),`? I don't remember why it was done this way, but the best way to do it would be to just use whatever has been already defined by the C header files.

HBelusca:
> I don't remember why it wasn't done this way, but the best way to do it would be to just use whatever has been already defined by the C header files.

If I understand the purpose of these geninc'ed files, it's for generating these `ks<processor_arch>.inc` ASM header files. However, for a given processor architecture (thus for such a given ASM inc file) it is wishable to be able to use it when compiling both uniprocessor and multiprocessor HAL/kernel (this is not selected during configuration time, but covered both during compile time where the MP HAL / kernel get the `CONFIG_SMP` flag defined accordingly).
If we were to just use `CONSTANT(SYNCH_LEVEL)`, since the geninc(data) is generated without any `CONFIG_SMP` flag set, the recorded value would be the one suitable for monoprocessor only, and then the generated ASM inc file would only get this definition, therefore unsuitable for use when that file is used also in the MP HAL / kernel.

sdk/include/ndk/amd64/ketypes.h
```diff
@@ -276,13 +276,17 @@ typedef enum
...
+#define ISR_STACK_SIZE 0x6000
+
 //
 // Synchronization-level IRQL
 //
-#define SYNCH_LEVEL 12
-
-#define NMI_STACK_SIZE 0x2000
-#define ISR_STACK_SIZE 0x6000
+#ifndef CONFIG_SMP
```
ThFabba, line 285: I never thought this made a whole lot of sense the way it's done for x86. `CONFIG_SMP` has no meaning outside of the kernel, in most cases it won't be defined, therefore you get the UP version, no matter what type of system you're running on.
Perhaps something like `#define SYNCH_LEVEL_UP ...` / `#define SYNCH_LEVEL_MP ...` would be more useful?

HBelusca:
So, checking around the MS DDKs (Win2000, Win2003 and Win7 only...) to see how `SYNCH_LEVEL` is defined:
- always defined for both UP and MP in the ASM-generated ksXXX.inc headers;
- defined with only its MP value in wdm.h and ntddk.h in Win2k, Win2003 DDK,
- defined for both UP and MP in wdm.h and ntddk.h of WinXP DDK,
- completely absent in wdm.h (and ntddk.h it would appear) in the Win7 WDK.

Regarding the PCR define in ksXXX.inc : only defined for both UP and MP in Win2k3 DDK; otherwise just the MP definition (`PCR equ fs:`) in Win7 WDK.

sdk/include/ndk/amd64/ketypes.h
tkreuzer: NDK contains a lot of stuff that should be private to the kernel. But this is what we do and we need this definition somewhere, I guess.

tkreuzer (approved):

sdk/include/ndk/amd64/ketypes.h
HBelusca: Those definitions may also be used in the HAL.

DarkFire01 (approved):
SMP ARM actually defines SYNCH_LEVEL on CONFIG_SMP as IPI_LEVL - 2 as well! But otherwise this is all good.

PR #3944, opened 2021-09-05
[NTOS:KE] Move the SList fault check after the MmAccessFault() handling in KiTrap0EHandler().

HBelusca:
- SList faults are much less common than regular access faults.

- Breakpoint handling in SList fault handlers could be applied first
  (via the owed BPs) before handling the SList faults??
  (WIP -- under question...)

### DIFF
--- a/ntoskrnl/ke/i386/traphdlr.c
+++ b/ntoskrnl/ke/i386/traphdlr.c
@@ -1362,13 +1362,6 @@ KiTrap0EHandler(IN PKTRAP_FRAME TrapFrame)
                          TrapFrame);
     }
 
-    /* Check for S-List fault */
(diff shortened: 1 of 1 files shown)


tkreuzer (changes requested):
That doesn't make sense. MmAccessFault will do all kinds of stuff that should never be done, when it is an SList-Fault. In Kernel-Mode this could lead to a bugcheck, when IRQL is above APC_LEVEL (which is fine for SList-Faults) or when the address is not mapped at all. Windows has these checks all over the place to handle all the possible cases, but that is stupid. It was intentionally placed where it is to avoid this crazyness and rule out SList-Faults from the start.

HBelusca:
Thanks for the explanation! Closing PR as invalid.

PR #3953, opened 2021-09-10
[NTUSER] Improve Themed Scrollbars by Minimizing Updates.

Doug-Lyons:
## Improve Themed Scrollbar Performance by doing Updates only when required

_Added conditional checks to single line of scrollbar update code._

JIRA issues: [CORE-17754](https://jira.reactos.org/browse/CORE-17754) and [CORE-17755](https://jira.reactos.org/browse/CORE-17755)

## Improve Themed Scrollbar updating code

_Fix both JIRA issues noted above._

### DIFF
--- a/win32ss/user/ntuser/scrollbar.c
+++ b/win32ss/user/ntuser/scrollbar.c
@@ -686,7 +686,6 @@ co_IntSetScrollInfo(PWND Window, INT nBar, LPCSCROLLINFO lpsi, BOOL bRedraw)
                   UpdateRect.bottom -= psbi->dxyLineButton;
                }
             }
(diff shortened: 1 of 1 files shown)


JoachimHenze (approved):
We know that this patch is not entirely side-effect-free, but in sum we concluded it to be an improvement over current master head. See my detailed testing results here https://jira.reactos.org/browse/CORE-17755?focusedCommentId=129590&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-129590

JoachimHenze:
We improved further with the last update of this PR.
The themed scrollbars are still not perfect yet (I will create new JIRA tickets soon), but at least now we reached a state where I am not aware of any new regressions. The resulting state is good, without any doubt a clear improvement, and safe to be ported back into 0.4.14RC also!

PR #3960, opened 2021-09-13
Fix clang-cl boot

Extravert-ir:
Now it boots, with a patch reducing the stack usage in uniata.
PR is only for checking CI


dll/win32/kernel32/client/file/find.c
```diff
@@ -687,9 +687,10 @@ FindFirstFileExW(IN LPCWSTR lpFileName,
         /*
          * May represent many FILE_BOTH_DIR_INFORMATION
          * or many FILE_FULL_DIR_INFORMATION structures.
+         * NOTE: NtQueryDirectoryFile requires the buffer to be ULONG-aligned
          */
-        BYTE DirectoryInfo[FIND_DATA_SIZE];
-        DIR_INFORMATION DirInfo = {&DirectoryInfo};
+        DECLSPEC_ALIGN(4) BYTE DirectoryInfo[FIND_DATA_SIZE];
+        DIR_INFORMATION DirInfo = { .DirInfo = &DirectoryInfo };
```
HBelusca, line 693: Does clang complain if you don't use the named member?

.github/workflows/build.yml
```diff
@@ -173,13 +173,7 @@ jobs:
...
     - name: Install ninja
       run: choco install -y ninja
-    - name: Install LLVM (x86)
-      if: ${{ matrix.arch == 'i386' }}
-      run: |
-        choco install --x86 -y llvm
-        echo "LLVM_PATH=${env:PROGRAMFILES(X86)}\llvm\bin" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
-    - name: Install LLVM (amd64)
-      if: ${{ matrix.arch == 'amd64' }}
+    - name: Install LLVM
```
ThFabba, line 176: What's this step for? I thought the clang-cl that ships with VS worked as-is.

base/setup/welcome/CMakeLists.txt
```diff
@@ -9,7 +9,7 @@ add_importlibs(welcome gdi32 user32 shell32 msvcrt kernel32 ntdll)
 if(ARCH STREQUAL "i386")
     if(MSVC)
         # NOTE: We cannot use the following command:
-        #   add_target_link_flags(welcome "/SUBSYSTEM:WINDOWS,4.00")
+        #   target_link_options(welcome PRIVATE "/SUBSYSTEM:WINDOWS,4.00")
```
ThFabba, line 12: I'm all in favor of making this change. But does it in any way relate to fixing clang-cl boot, as its presence in this PR would have you believe?

dll/ntdll/CMakeLists.txt
```diff
@@ -60,10 +60,6 @@ set_module_type(ntdll win32dll ENTRYPOINT 0)
 set_subsystem(ntdll console)
 ################# END  HACK #################
 
-if(MSVC)
-    add_target_link_flags(ntdll "/RELEASE")
-endif()
```
ThFabba, line 65: This is non-obvious in a commit that just claims to replace the command used to add linker flags.
Is this redundant because we add it by default, or... ?

sdk/cmake/CMakeMacros.cmake
```diff
@@ -287,10 +287,10 @@ function(add_cd_file)
     endif()
 
     # do we add it to all CDs?
-    list(FIND _CD_FOR all __cd)
+    list(FIND _CD_FOR "all" __cd)
     if(NOT __cd EQUAL -1)
-        list(REMOVE_AT _CD_FOR __cd)
```
ThFabba, line 292: So should this have just been `${__cd}` instead?
I recall this function being quite performance sensitive, so I'm not sure if having it repeat the search will be problematic.

dll/ntdll/CMakeLists.txt
HBelusca: The commit that added this extra `/RELEASE` flag: https://github.com/reactos/reactos/commit/005a7acbfcf8dc2fb74944dc39882084a95740c7 , claimed it's for getting a PE checksum.
However in the meantime, for MSVC builds, this flag is systematically added, see https://github.com/reactos/reactos/blob/master/sdk/cmake/msvc.cmake#L177 , so it may be fine to remove it there, indeed.

sdk/cmake/CMakeMacros.cmake
Extravert-ir: This list is small in general, moreover for all occurrences of "all" parameter, the list contains only one element (that actually explains how it worked all the time).
So I don't think it introduces any performance impact. We can probably just construct a new list here

dll/ntdll/CMakeLists.txt
Extravert-ir: Yeah, `/RELEASE` is added globally, so I decided to remove the flag here

base/setup/welcome/CMakeLists.txt
Extravert-ir: No, it doesn't. I just pushed all the stuff laying around in one PR to check the CI

.github/workflows/build.yml
Extravert-ir: clang-cl from VS is old. This is done to use an equal version of clang on both linux and windows

dll/win32/kernel32/client/file/find.c
Extravert-ir: No, I just added it for better readability

PR #3969, opened 2021-09-22
[SHELL32] Add SS_NOPREFIX to owner labels in About dialog

wjk:
## Purpose

What the title says, really. Without these changes, owner name or organization strings containing an `&` will be displayed incorrectly.

I changed the RC files for all languages, but only tested en-US. However, I doubt there will be any problems in the other locales.


HBelusca:
I'm also curious to know whether this could be needed for the other `IDC_ABOUT_APPNAME`, `IDC_ABOUT_VERSION` and `IDC_ABOUT_OTHERSTUFF` text zones too. One could double-check whether Windows does that too, by setting some text with an `&` when calling the `ShellAbout` API.

learn-more (approved):

wjk:
@binarymaster Done. Sorry about that! For some reason, the allow-maintainer-access button is not appearing where it should be on my end.

binarymaster:
> For some reason, the allow-maintainer-access button is not appearing where it should be on my end.

@wjk see this checkbox at right bottom:

image

PR #3972, opened 2021-09-23
[NTOS:OB] Acquire the lock before setting directory's session ID

GeoB99:


HBelusca (dismissed):

JoachimHenze:
Is there anything user-observable that will get fixed by this? Do we have any synthetic test-case or known real-world-scenario to experience that improvement?

GeoB99:
> Actually you need to use the ObpAcquireDirectoryLockExclusive and ObpReleaseDirectoryLock functions

Never saw those, good idea.
> Is there anything user-observable that will get fixed by this? Do we have any synthetic test-case or known real-world-scenario to experience that improvement?

There's no user-observable with this patch. The reason due to this locking wrapping protection is so that we enforce concurrency limits on threads that access this resource, that is, when the calling thread acquires the lock and processes on setting the session ID of the directory object, no other executing threads can touch that resource until the current calling thread has finished its job and released the lock. This is an incremental bonus for the kernel.

ntoskrnl/ob/oblife.c
```diff
@@ -1830,6 +1831,9 @@ NtSetInformationObject(IN HANDLE ObjectHandle,
 
         case ObjectSessionInformation:
 
+            /* Setup a lookup context */
+            ObpInitializeLookupContext(&LookupContext);
```
HBelusca, line 1835: Please move this to the code block where the locking is actually done (in the `NT_SUCCESS(Status)` one).
The variable declaration at the top can be moved there too.

ntoskrnl/ob/oblife.c
```diff
@@ -1849,10 +1849,18 @@ NtSetInformationObject(IN HANDLE ObjectHandle,
...
+                    ObpInitializeLookupContext(&LookupContext);
+
                     /* Set its session ID */
+                    ObpAcquireDirectoryLockExclusive(Directory, &LookupContext);
                     Directory->SessionId = PsGetCurrentProcessSessionId();
+                    ObpReleaseDirectoryLock(Directory, &LookupContext);
                     ObDereferenceObject(Directory);
+
+                    /* We're done, release the context */
+                    ObpReleaseLookupContext(&LookupContext);
```
HBelusca, line 1863: This needs to be done before the Directory dereferencing.

HBelusca (approved):

PR #3977, opened 2021-09-26
[NTOS:SE] Check the privilege count if it's within the limit threshold when capturing

GeoB99:
In addition to that, the limit "magic" threshold value in `SeCaptureSidAndAttributesArray` must also be explicitly named with a define macro so that one understands what is it about.

### DIFF
--- a/ntoskrnl/se/priv.c
+++ b/ntoskrnl/se/priv.c
@@ -15,6 +15,8 @@
 
 /* GLOBALS ********************************************************************/
 
+#define SE_MAXIMUM_PRIVILEGE_LIMIT 0x3C
+
 #define CONST_LUID(x1, x2) {x1, x2}
 const LUID SeCreateTokenPrivilege = CONST_LUID(SE_CREATE_TOKEN_PRIVILEGE, 0);
 const LUID SeAssignPrimaryTokenPrivilege = CONST_LUID(SE_ASSIGNPRIMARYTOKEN_PRIVILEGE, 0);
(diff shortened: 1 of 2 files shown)


ntoskrnl/se/priv.c
```diff
@@ -15,6 +15,8 @@
 
 /* GLOBALS ********************************************************************/
 
+#define SE_MAXIMUM_PRIVILEGE_LIMIT_THRESHOLD 0x3C
```
HBelusca, line 18: If it's a "maximum" "limit", it's then already a threshold?

HBelusca (approved):
LGTM
(Side question: Do we have some sort of test for checking these values? In particular the max privilege limit value didn't appear originally in the SeCaptureLuidAndAttributesArray function)

GeoB99:
> (Side question: Do we have some sort of test for checking these values? In particular the max privilege limit value didn't appear originally in the SeCaptureLuidAndAttributesArray function)

No, we don't. And for the simple fact is that this API routine call is internal and only the kernel itself can use it, especially the security manager subsystem and nobody else. So I highly doubt it's feasible to write a kernel-mode test for that. 

I think an alternative is to write a "special" kind of test for a NT syscall like NtAdjustPrivilegesToken by cheating on the kernel on simulating an integer overflow when SeCaptureLuidAndAttributesArray actually captures the privileges and count array. Something to be done in the future I guess.

> In particular the max privilege limit value didn't appear originally

Simply because it wasn't implemented in ReactOS till now, as with many things that aren't implemented here but are in Windows. ¯\_(ツ)_/¯

PR #3984, opened 2021-09-30
[WIN32SS:GDI] Add all known function calls to NtCurrentTeb for resetting spins.

jackson2k2:
## Purpose and proposed change

I noticed missing writes to the spin count in ClientInfo within GDI32 while debugging API tests with James Tabor's recent changes to the GDI Metafile process. This doesn't appear to functionally break anything, and if I understand correctly, these lines are certainly in the original GDI subsystem for optimizing polling.

Unsure if there any additional writes of the same kind in NT 6.x.

### DIFF
--- a/win32ss/gdi/gdi32/objects/bitmap.c
+++ b/win32ss/gdi/gdi32/objects/bitmap.c
@@ -731,6 +731,8 @@ SetDIBitsToDevice(
 
     cjBmpScanSize = DIB_BitmapMaxBitsSize((LPBITMAPINFO) lpbmi, ScanLines);
 
+    ((PW32CLIENTINFO)NtCurrentTeb()->Win32ClientInfo)->cSpins = 0;
+
     pvSafeBits = RtlAllocateHeap(GetProcessHeap(), 0, cjBmpScanSize);
     if (pvSafeBits)
     {
@@ -838,6 +840,8 @@ StretchDIBits(
     }
 
     cjBmpScanSize = GdiGetBitmapBitsSize((BITMAPINFO *) pConvertedInfo);
+
+    ((PW32CLIENTINFO)NtCurrentTeb()->Win32ClientInfo)->cSpins = 0;
 
     if (lpBits)
     {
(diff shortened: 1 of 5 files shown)


win32ss/gdi/gdi32/objects/bitmap.c
```diff
@@ -731,6 +731,8 @@ SetDIBitsToDevice(
 
     cjBmpScanSize = DIB_BitmapMaxBitsSize((LPBITMAPINFO) lpbmi, ScanLines);
 
+    ((PW32CLIENTINFO)NtCurrentTeb()->Win32ClientInfo)->cSpins = 0;
```
HBelusca, line 734: Maybe do it just before `GdiGetHandleUserData` or the NtGdi call? (same question for some other places.)

jackson2k2: Tracing reveals that Windows seems to place it just before the RtlAllocateHeap calls that write to pvSafeBits, so if anything, I would judge the original developers.

binarymaster (changes requested):
Hello,
Sorry for late reply and thanks for your interest in ReactOS!

Please set up your GitHub account correctly, so maintainers can use Squash and Merge button without getting the author field modified back to nickname by GitHub, it's done in two steps:

1. Set your full name in [GitHub profile settings](https://github.com/settings/profile)
1. Uncheck `Keep my email addresses private` in [GitHub email settings](https://github.com/settings/emails)

Also it seems the email `j.p.brun@zoho.com` that you have specified in commit is not associated with your GitHub account, please associate and confirm it in [GitHub email settings](https://github.com/settings/emails).

Regards,

tkreuzer (changes requested):
What's the point of this change? What does this fix? How does it work? Why do we need/want this?
If you can't explain all of this, this should not be committed. There shouldn't be any code in ros, which nobody can explain.
The reason should be obvious.

HBelusca:
> What's the point of this change? What does this fix? How does it work? Why do we need/want this? If you can't explain all of this, this should not be committed. There shouldn't be any code in ros, which nobody can explain. The reason should be obvious.

There was this hypothesis mentioned in the PR description:
> for optimizing polling

PR #3990, opened 2021-10-03
[ROSAPPS] Remove 'man' utility

learn-more:
There is no activity on this utility, nor are there 'man' pages being written.
No-one seems to expect a 'man' utility on a win32 system.


HBelusca (approved):
OK since the format of the man files don't appear to be the "standard" one.

binarymaster (approved):

PR #3993, opened 2021-10-03
[SHELL32] Don't smartass on folders

katahiromz:
## Purpose
Selection on renaming folders shouldn't select the extension-like part of filename.
JIRA issue: [CORE-17793](https://jira.reactos.org/browse/CORE-17793)

## Proposed changes

- Modify the condition of extension non-selection.

## Screenshots
BEFORE:
before
after2

### DIFF
--- a/dll/win32/shell32/CDefView.cpp
+++ b/dll/win32/shell32/CDefView.cpp
@@ -1866,6 +1866,9 @@ SelectExtOnRename(void)
     HKEY hKey;
     LONG error;
     DWORD dwValue = FALSE, cbValue;
+    static DWORD s_dwCached = 0xDEADBEEF;
+    if (s_dwCached != 0xDEADBEEF)
+        return s_dwCached;
 
     error = RegOpenKeyExW(HKEY_CURRENT_USER,
                           L"SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Explorer",
@@ -1877,7 +1880,8 @@ SelectExtOnRename(void)
     RegQueryValueExW(hKey, L"SelectExtOnRename", NULL, NULL, (LPBYTE)&dwValue, &cbValue);
 
     RegCloseKey(hKey);
-    return !!dwValue;
+    s_dwCached = !!dwValue;
+    return s_dwCached;
 }
 
 /**********************************************************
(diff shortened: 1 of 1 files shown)


dll/win32/shell32/CDefView.cpp
```diff
@@ -1866,6 +1866,9 @@ SelectExtOnRename(void)
     HKEY hKey;
     LONG error;
     DWORD dwValue = FALSE, cbValue;
+    static DWORD s_dwCached = 0xDEADBEEF;
```
HBelusca, line 1869: Shouldn't we use a global to be initialized at startup, as it may be done for other such settings ?

JoachimHenze: Wouldn't the caching mechanism require an explorer-reboot to ever reread that setting? Isn't that over-optimization?
I base my argumentation on the assumption that SelectExtOnRename() is only called only once per renaming-request.

HBelusca: In principle, registry caching should be done together with listening for registry changes (registry notifications) on the involved registry keys. However this mechanism is currently unimplemented in ReactOS and thus cannot be used.

learn-more: Typically settings like this would be in something like `SHELLSTATE`, however this looks like a custom ROS thing, so not sure how to handle that.

JoachimHenze: Well then a vote? I'd vote for leaving the caching out for now. We could always touch it later, when Vista+ stuff designs will get more clear to us. For now just go with the most simple code. After all we also had no performance issue before PR3993. So what to fear?

katahiromz: Now deleted the caching.

learn-more: @JoachimHenze death by a thousand cuts?
There are tons of unoptimized code paths in our shell, sure.
When drawing each entry, we check the disk, we check the registry, all slowing us down a tiny bit.
Don't start complaining when explorer is slow on refreshing icons or browsing folders then.

JoachimHenze: Mark, as said, I expected that function to only be called when invoking a rename. Is that a wrong assumption?

learn-more: This particular function is indeed only called on rename, but `SHELL_FS_HideExtension` for example is not.

JoachimHenze:
I want to mention that I tried, and the current code (without the cache!) does atm also properly compile with VS2010 rls+dbg and also old RosBE2.1.6 GCC4.7.2dbg+rls, **which I will need for the backport to 0.4.14**. Please wait with the merge until @learn-more will either set his green flag for the merge, or will propose some specific change.
I didn't want to rush anything here or push into some specific direction, I just asked for a vote regarding the caching.

arnavbhatt288:
>me reading the patch name
image

dll/win32/shell32/CDefView.cpp
katahiromz: Should we modify `SHELL_FS_HideExtension`?

PR #3994, opened 2021-10-03
[RAPPS] Speed up installed application loading

learn-more:
Only load in details when we are going to display them


base/applications/rapps/include/installed.h
```diff
@@ -5,41 +5,48 @@
 
 class CInstalledApplicationInfo
 {
+private:
+    BOOL m_IsUserKey;
+    REGSAM m_WowKey;
+    HKEY m_hSubKey;
+
+    ATL::CStringW m_szKeyName;
```
HBelusca, line 13: The CStringW members below don't have `ATL::` namespace explicitly written.

base/applications/rapps/installed.cpp
```diff
@@ -2,123 +2,122 @@
...
+        CloseHandle(m_hSubKey);
+        m_hSubKey = NULL;
     }
-    else
+}
+
+void CInstalledApplicationInfo::EnsureDetailsLoaded()
+{
+    // Key not closed, so we have not loaded details yet
+    if (m_hSubKey)
```
HBelusca, line 37: You could return early and save one indentation level by `if (!m_hSubKey) return;`

learn-more: And what problem would this solve exactly?
It's not forbidden to have indentation....

wiktorek140: It will save a little of CPU time by shorting execution path. And its easier to maintain readability when there are less indentation level :)

HBelusca: It's just about readability. In this case the CPU time wouldn't certainly be modified: both assemble to a jump instruction.

learn-more: > It will save a little of CPU time by shorting execution path. And its easier to maintain readability when there are less indentation level :)

I very much doubt that this has any relevance for the execution path.

base/applications/rapps/include/installed.h
```diff
@@ -5,41 +5,48 @@
...
+    CStringW szHelpTelephone;
+    CStringW szReadme;
+    CStringW szContact;
+    CStringW szURLUpdateInfo;
+    CStringW szURLInfoAbout;
+    CStringW szInstallDate;
+    CStringW szInstallLocation;
+    CStringW szInstallSource;
+    CStringW szUninstallString;
+    CStringW szModifyPath;
```
JoachimHenze, line 49: stirp that empty line as well? No hard requirement though.

PR #4001, opened 2021-10-06
[ZIPFLDR] Show error when compression method is not supported

binarymaster:
JIRA issue: [CORE-17255](https://jira.reactos.org/browse/CORE-17255)

https://user-images.githubusercontent.com/578406/136253168-fa7080bd-2723-4eb6-acc8-087b5c49fa46.mp4


binarymaster:
> how could "continuing" make even sense?

The video I attached is the answer :) Different files in archive can be compressed with different methods.

PR #4010, opened 2021-10-08
[SETUPLIB][CPL/INPUT] Move zh-MO to ChineseTraditionalFonts group

eason329:
## Proposed changes
A trivial fix for moving Chinese (Macau S.A.R.) related definition/ code to ChineseTraditionalFonts group.

**Reason**: 
* the use of Traditional Chinese character is far more than Simplified one. (according to Wikipedia)
* the codepage used on zh-MO is 950, which is Traditional Chinese's one.
* In [this page](https://support.microsoft.com/en-us/windows/microsoft-traditional-chinese-ime-ef596ca5-aff7-4272-b34b-0ac7c2631a38#ID0EBBD=Microsoft_ChangJie/Quick), zh-MO is described as Traditional Chinese.

## P.S.
* I don't think this language will be available when ROS<s> 1.0 is officially released</s> enter beta stage lol

### DIFF
--- a/base/setup/lib/muilanguages.h
+++ b/base/setup/lib/muilanguages.h
@@ -509,7 +509,7 @@ const MUI_LANGUAGE MUILanguageList[] =
     {L"00001004", L"936", L"936", L"10008", L"Chinese (Singapore)", L"215", ChineseSimplifiedFonts, zhSGLayouts},
 #endif
 #ifdef LANGUAGE_ZH_MO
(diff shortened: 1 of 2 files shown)


eason329:
> 
> Why not?

* There is few contributors came from Macau (considering the population of Macau)
* This project is actually not so popular in Greater China area in recent days for some reason

Of course, these are just my personal feeling and I am glad to see someone brings zh-MO to this OS.

eason329:
oh I forgot to say that only the text strings like `Chinese (Macau S.A.R.)` will be changed because:

> I guess we have it like this because of `SUBLANG_CHINESE_MACAU` identifier defined by Microsoft

yea the identifier is "hard coded" by Microsoft so we won't change it.

PR #4011, opened 2021-10-09
[HHPCOMP] lz_nonslide.h: Define u_char on Linux

konradybcio:
## Purpose

Not defining the type breaks the compilation on (at least arm) Linux. Add a check for Linux-as-build-host to make it pass.

JIRA issue: none?

## Proposed changes

- define the type so that the compiler knows what on earth u_char is

### DIFF
--- a/sdk/tools/hhpcomp/lzx_compress/lz_nonslide.h
+++ b/sdk/tools/hhpcomp/lzx_compress/lz_nonslide.h
@@ -16,7 +16,7 @@
     Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 */
 
-#if defined(_WIN32) || defined(__APPLE__)
+#if defined(_WIN32) || defined(__APPLE__) || defined(__linux__)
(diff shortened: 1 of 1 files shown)


learn-more:
lzx_compress from hhpcomp is third party code,
so the preferred approach is to first check if upstream has already fixed this:
https://git.reactos.org/?p=reactos.git;a=blob;f=media/doc/3rd+Party+Files.txt;hb=da4d0545ebf9610331ec186a1275fb44ed22245f#l252

konradybcio:
> lzx_compress from hhpcomp is third party code, so the preferred approach is to first check if upstream has already fixed this: https://git.reactos.org/?p=reactos.git;a=blob;f=media/doc/3rd+Party+Files.txt;hb=da4d0545ebf9610331ec186a1275fb44ed22245f#l252

Looks like they don't even define `u_char` upstream..

SergeGautherie:
> > check if upstream
> 
> Looks like they don't even define `u_char` upstream..

I confirm I checked current upstream file before I commented.

PR #4015, opened 2021-10-09
[USER32] Detect White followed by Black in color table as monochrome

Doug-Lyons:
## Allows color table with White and then Black to be detected as monochrome

_Add additional code in is_dib_monochrome for this option._

JIRA issue: [CORE-13935](https://jira.reactos.org/browse/CORE-13935)

## Additional code testing White then Black added to original test for Black and then White

This code is a modification of code originally proposed by Fabian Maurer as a Wine Fix.
See https://bugs.winehq.org/show_bug.cgi?id=47486

### DIFF
--- a/win32ss/user/user32/windows/cursoricon.c
+++ b/win32ss/user/user32/windows/cursoricon.c
@@ -122,12 +122,12 @@ static BOOL is_dib_monochrome( const BITMAPINFO* info )
         if (((const BITMAPCOREINFO*)info)->bmciHeader.bcBitCount != 1) return FALSE;
 
         /* Check if the first color is black */
-        if ((rgb[0].rgbtRed == 0) && (rgb[0].rgbtGreen == 0) &&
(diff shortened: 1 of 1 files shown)


HBelusca:
As of today it would appear this is still not fixed on Wine side: https://bugs.winehq.org/show_bug.cgi?id=47486
Does anyone want to send this patch to them?

HBelusca (approved):

binarymaster:
@Doug-Lyons I have checked recent updates on the Wine side, it seems they need API tests for better understanding of the problem.

If you are around, it would be nice to create some tests for this, either in this or in a separate PR.

Doug-Lyons:
@binarymaster, Looking at the history in Wine, @DarkShadow44 (https://jira.reactos.org/secure/ViewProfile.jspa?name=DarkShadow44) was talking about writing tests. I will take a look at this and see what I can do as well. Thanks.

PR #4022, opened 2021-10-10
Resolve compiler warnings in host_tools

wjk:
The host_tools gave me endless amounts of warnings about `#pragma pack` usage. My version of clang apparently doesn’t understand the Windows convention of using included headers to modify the structure packing. This PR silences that and a number of other overly-pedantic clang warnings.

Two of the warnings, however, were actionable:

1. There was a `delete`/`delete[]` mismatch in txt2nls.
2. The warning in cabman actually revealed a logic error. If there is an error in the command-line syntax, the program would exit with `false`, which converts to a zero exit code. As a result, Ninja would think that the cabman call succeeded when it really didn’t, causing chaos downstream. Therefore, I changed the return value from `false` to a nonzero integer exit code.


HBelusca:
All commits except for the one in CMakeLists.txt have been committed in master.

wjk:
Unless anyone disagrees, I’m going to close this PR per @HBelusca’s statement. @learn-more Since you asked, every one of those warnings were issued on my devbox by whatever Clang version Apple bundled with Xcode at the time I last updated this PR. It apparently didn’t like the kind of cross-compilation we’re doing. 😜 Thanks!

PR #4023, opened 2021-10-11
[NTOS:PNP] Fix resource conflict detection

Extravert-ir:
Previous code did not detect equal resource ranges as conflicting.
Thanks Hervé Poussineau for pointing this out!

Meanwhile, simplify the code to make it more readable.


binarymaster:
Maybe mention ticket CORE-17789 here too.

PR #4024, opened 2021-10-11
[NOTEPAD] Modify access key for "New Window"

eason329:
## Purpose
Follow up of #3452

A trivial modification for "New Window" option's<s> hotkey</s> access key in "File" context menu
> Fun fact: I found this issue when I translating this to zh-TW lol

## Proposed changes
The recent hotkey of "New Window" option conflicts to "New" 's one. This PR change the hotkey of former from "N" to "W".

To be exact, it changes to `MENUITEM "New &Window\tCtrl+Shift+N", CMD_NEW_WINDOW`. The hotkey location can be further discussed.

**Please test the program before approval.**

The following languages is not included because they're already translated: 
> *italic text* mean  confliction occurs on that language, translator should consider change it
* *de-DE*
* hu-HU
* *pt-PT*
* *tr-TR*

### DIFF
--- a/base/applications/notepad/lang/de-DE.rc
+++ b/base/applications/notepad/lang/de-DE.rc
@@ -24,7 +24,7 @@ BEGIN
     POPUP "&Datei"
     BEGIN
         MENUITEM "&Neu\tStrg+N", CMD_NEW
-        MENUITEM "&Neues Fenster\tStrg+Umschalt+N", CMD_NEW_WINDOW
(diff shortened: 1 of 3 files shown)


binarymaster:
Note the hotkey is actually `Ctrl+Shift+N` here, while the underlined letter is a **mnemonic**, which is called an **access key** in Windows.

https://en.wikipedia.org/wiki/Mnemonics_(keyboard)

That's why Joachim was confused at first here. I renamed commit title accordingly.

PR #4027, opened 2021-10-12
[NTOS:PNP] Halfplement IoInvalidateDeviceState

Extravert-ir:
Implement the correct start-stop sequence for resource rebalancing
without the actual rebalancing. Also move IoInvalidateDeviceState
processing into the enumeration thread as it should be.

Jira issue: [CORE-17519](https://jira.reactos.org/browse/CORE-17519)

### DIFF
--- a/ntoskrnl/include/internal/io.h
+++ b/ntoskrnl/include/internal/io.h
@@ -527,7 +527,8 @@ typedef enum _DEVICE_ACTION
     PiActionEnumRootDevices,
     PiActionResetDevice,
     PiActionAddBootDevices,
-    PiActionStartDevice
+    PiActionStartDevice,
+    PiActionQueryState,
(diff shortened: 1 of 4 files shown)


Extravert-ir:
@disean I remember you had issues in this code path in isapnp. Does it work better now?

disean:
> Does it work better now?

This PR causes the same resource list (AllocatedResources) to be sent over and over again, thus creating an infinite loop

Extravert-ir:
> This PR causes the same resource list (AllocatedResources) to be sent over and over again, thus creating an infinite loop

@disean should work the same way as previous code now

disean:
Just tested it on a clean master, but I'm still getting the same infinite loop

Extravert-ir:
Merging this in the current state, please update the Jira ticket with current behaviour

PR #4033, opened 2021-10-14
[SETUP][REACTOS] Improve Japanese translation

katahiromz:
## Purpose

Improve the Japanese GUI of ReactOS Setup.
JIRA issue: N/A

## Proposed changes

- Modify Japanese resource.

## Screenshots
I got some screenshots from compiled binary by using `RisohEditor` with replacing `"MS UI Gothic"` font with `"Droid Sans Fallback"`:

### BEFORE
無題1
無題2
無題3
無題4

### AFTER
無題1
無題2
無題3
無題4


katahiromz:
> maybe it's worth to look at the runtime pics in ReactOS.

I couldn't test it due to CORE-17818.

PR #4036, opened 2021-10-14
[IMM32_APITEST] Fix himc testcase crash

katahiromz:
## Purpose
Fix hung-up in the `himc` testcase.

## Proposed changes

- Add `NULL` checks for `pIC`.

## Comparison

BEFORE:
before
AFTER:
after


katahiromz:
The IMM/Cicero mode is not implemented in our kernel. So `pIC` becomes `NULL`. No action from IMM32 side in this PR.
Win2k3 implements the Cicero mode. So `pIC` becomes non-`NULL`.

PR #3981, opened 2021-09-27
[MMIXER] Add support for WAVE_FORMAT_EXTENSIBLE audio format

oleg-dubinskiy:
## Purpose

Enable support for `WAVE_FORMAT_EXTENSIBLE` audio format in MMixer library.
In general, it allows to properly play/capture the sound by apps whose specify `WAVE_FORMAT_EXTENSIBLE` format tag.
For example, after this fix, AIMP 4.71 is playing the audio files correctly with 16 bps and 2 (stereo) channels. Debug output proves that it passes `WAVE_FORMAT_EXTENSIBLE` format in `WAVEFORMATEX.cbSize`, and with 16 bit & stereo channels (only that is supported by Intel AC97 driver), playback is working properly.

JIRA issue: [CORE-10907](https://jira.reactos.org/browse/CORE-10907)

## Proposed changes

Allow using `WAVE_FORMAT_EXTENSIBLE` format tag for newly created audio pins:
- Add it into the formats check, to pass it further for creating a new pin, so now the check does not fail with `MM_STATUS_NOT_IMPLEMENTED` when requesting this format.
- Add additional debug print in this check, which will be displayed when any other unsupported yet format will be called (besides `WAVE_FORMAT_PCM` and `WAVE_FORMAT_EXTENSIBLE`).
- Set the correct size for `WAVEFORMATEX.cbSize` member. MS says it should have the size of additional data (for additional formats), but not zero, as for standard formats, since `WAVEFORMATEXTENSIBLE` structure, in essence, is just an addition for `WAVEFORMATEX`. It stores all additional data, which has some size (stored in `cbSize` member). This data can be different, since it is specified by the caller, and the size of it also can be different appropriately.
- Do the same for `KSDATAFORMAT.FormatSize` member.

## TODO

- [x] Perhaps add additional size to `KSDATAFORMAT.FormatSize` also, instead of dumb `sizeof(WAVEFORMATEX)` (since `cbSize` is `0` by default), because it requires the actual size of the data, so logically it looks correct.

PR #4041, opened 2021-10-16
[WDMAUD.DRV] Add workaround for AC97 driver

oleg-dubinskiy:
## Purpose

Add workaround for Intel AC97 driver that is used in VirtualBox and does not support any other quality and channels except 16 bps and stereo.
If opening device fails with requested audio format, then use supported one instead: in case of failure, just force channels to stereo and bps to 16, and try to open device again. In case with AC97 driver, the 2nd opening call should not fail.
Additionally, store requested format data in audio device instance structure, which avoids uninitializer usage of it, when using it with MMixer define.
Many other old audio cards (including ones used in emulators) also don't support other quality modes and channel layouts, but support all standart sample rates (8000 - 48000 Hz). So this hack is very helpful for them also.
In particular, it allows AIMP 4.71 to play the audio properly with different audio formats chosed in settings, in VirtualBox with AC97 driver used. But even if other (unsupported) formats are requested, only default format is actually used.
**Please note that it works only for MMixer APIs (those can be enabled or disabled via MMixer define), but it does not change anything for Legacy routines (whose are used by default, without MMixer define enabled). I didn't make a fix for Legacy APIs yet.**
**Note 2: this PR gives an excepted result only in pair with #3981 applied.**

JIRA issue: [CORE-10907](https://jira.reactos.org/browse/CORE-10907)

## Result

The attached video demonstrates how AIMP works after my changes: now it plays the audio files with any audio format chosed in settings. But for unsupported formats, the default format is actually used. :slightly_smiling_face: 

https://user-images.githubusercontent.com/26385117/137594233-3f39ebd0-f084-4b74-a462-586fb7f7bc4e.mp4

### DIFF
--- a/dll/win32/wdmaud.drv/mmixer.c
+++ b/dll/win32/wdmaud.drv/mmixer.c
@@ -496,6 +496,20 @@ WdmAudSetWaveDeviceFormatByMMixer(
 
     bWaveIn = (DeviceType == WAVE_IN_DEVICE_TYPE ? TRUE : FALSE);
 
+    if (WaveFormatSize >= sizeof(WAVEFORMAT))
+    {
+        /* Store format */
+        Instance->WaveFormatEx.wFormatTag = WaveFormat->wFormatTag;
+        Instance->WaveFormatEx.nChannels = WaveFormat->nChannels;
+        Instance->WaveFormatEx.nSamplesPerSec = WaveFormat->nSamplesPerSec;
+        Instance->WaveFormatEx.nBlockAlign = WaveFormat->nBlockAlign;
+        Instance->WaveFormatEx.nAvgBytesPerSec = WaveFormat->nAvgBytesPerSec;
+    }
+
+    /* Store details */
+    Instance->WaveFormatEx.cbSize = WaveFormat->cbSize;
+    Instance->WaveFormatEx.wBitsPerSample = (WaveFormat->nAvgBytesPerSec * 8) / (WaveFormat->nSamplesPerSec * WaveFormat->nChannels);
+
     if (MMixerOpenWave(&MixerContext, DeviceId, bWaveIn, WaveFormat, NULL, NULL, &Instance->Handle) == MM_STATUS_SUCCESS)
     {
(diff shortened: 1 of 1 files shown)


oleg-dubinskiy:
> Cool. But why is there sometimes silence suddenly for a while in your video, although you didn't change any setting then? I also noticed that the powermeter is totally async to the sound-output. Is that just a problem of my player here displaying video and audio async for some reason, or is this also what you experienced when you sat in front of that and interacted with it?

It's just the lags caused by recording from host OS, they didn't happen at all on ROS side when I tested it. Don't pay attention for that. :smile:

NailYggdrago:
This PR and the one mentioned in the description of it don't work properly when using the open source AC97 driver that is available at https://github.com/thedeadfish/reactos/tree/ac97_import_new2. The version of AIMP used in the video gives an error message when playing a video while using the driver compiled from the source I linked to.

SergeGautherie:
> don't work properly

Are you saying this PR(s) is causing some kind of regression?

NailYggdrago:
In other news, the DSound Wine Tests don't seem to be crashing anymore after applying this PR when using the closed source driver. I am now getting more output from them (they are still failing in most cases though). Also, the attached program now just tells me that the open was successful but no audio plays when using it.

[waveout.zip](https://github.com/reactos/reactos/files/7360570/waveout.zip)

SergeGautherie:
> the error might be with the driver itself

> attached program

Please, unless you notice actual _regressions_, such comments do not belong in this PR: open a Jira ticket instead, if that applies.

NailYggdrago:
I will open a ticket for the driver when it gets added and will open a ticket for the program when this PR and the one it needs are added.

NailYggdrago:
Anyone want to run it through the test bots to compare it to the latest source code?

JoachimHenze:
Yes I was intentionally very verbose in my commit comment https://github.com/reactos/reactos/commit/c437ff6128a2db3dbc62a4509957338cc29c8fe1 due to the many cross-relations of that audio-commit.
If zero3k aka NailYggdrago would have read it, he wouldn't have come up with such wild hypothesis.

PR #4042, opened 2021-10-17
[FREELDR] Sort out /BASEVIDEO and /NOVESA handling

SergeGautherie:
## Purpose

Properly handle requesting VGA driver itself (Safe mode) or VGA-like resolution only (VGA mode).

JIRA issue: [CORE-10236](https://jira.reactos.org/browse/CORE-10236)

## Proposed changes

- Tweak NTUSER.
- Improve VIDEOPRT.
- Fix FREELDR.

## TODO

- [ ] `NOVESA` option seems undocumented and not to exist on WXP/WS03.
Where does it come from?
I wonder if it should be replaced by checking `BASEVIDEO` and the relevant `SAFEBOOT:...`, in VIDEOPRT?
Even though WXP (VIDEOPRT) does not seem to do it that way either.
- [ ] This PR does not try to fix VBE driver (handling): 'VGA mode' still behaves like (new) 'Safe mode'.
Ideas welcome, or will remain as a follow-up.

### DIFF
--- a/boot/freeldr/freeldr/options.c
+++ b/boot/freeldr/freeldr/options.c
@@ -84,8 +84,9 @@ enum BootOption
 
 static enum BootOption BootOptionChoice = NO_OPTION;
 static BOOLEAN BootLogging = FALSE;
-static BOOLEAN VgaMode = FALSE;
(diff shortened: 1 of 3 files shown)


DarkFire01:
Sorry as this is stale and a bit incorrect I’m closing this for now. Thank you anyway!

PR #4049, opened 2021-10-19
[KERNEL32] Implement SetFileCompletionNotificationModes

oleg-dubinskiy:
## Purpose

Implement `SetFileCompletionNotificationModes` function in our kernel32. MSDN says this function was added since Vista+, but according to @HBelusca's comment on Jira, this API is also present in Windows Server 2003 SP2: https://jira.reactos.org/browse/CORE-17821?focusedCommentId=130199&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-130199.
Call native Nt* function to do the actual work, similarly to as it done in Wine: https://source.winehq.org/git/wine.git/blob/530c1839603bd3caa1b9d17fe458b5bd341ccfc9:/dlls/kernel32/file.c#l258.
Also add/fix some declarations in internal kernel32/public NDK headers, to fix compilation.
It does not fix the actual issue, but at least makes some progress on it.

JIRA issue: [CORE-17821](https://jira.reactos.org/browse/CORE-17821)

## Proposed changes

- Implement the function itself.
- Include required header in kernel32.
- Add missing structure declaration into public NDK header.

### DIFF
--- a/dll/win32/kernel32/client/file/iocompl.c
+++ b/dll/win32/kernel32/client/file/iocompl.c
@@ -3,7 +3,8 @@
  * PROJECT:         ReactOS system libraries
  * FILE:            dll/win32/kernel32/client/file/iocompl.c
  * PURPOSE:         Io Completion functions
- * PROGRAMMER:      Ariadne ( ariadne@xs4all.nl)
+ * PROGRAMMERS:     Ariadne (ariadne@xs4all.nl)
+ *                  Oleg Dubinskiy (oleg.dubinskij2013@yandex.ua)
  * UPDATE HISTORY:
  *                  Created 01/11/98
  */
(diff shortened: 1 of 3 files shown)


oleg-dubinskiy:
MSDN says that the appropriate `FILE_IO_COMPLETION_NOTIFICATION_INFORMATION` structure of `FileIoCompletionNotificationInformation` class is "reserved for system use": https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/ne-wdm-_file_information_class. Does it mean this shouldn't be used at all? Or it should be used only internally in the system? :thinking:

GeoB99:
> 
> 
> MSDN says that the appropriate `FILE_IO_COMPLETION_NOTIFICATION_INFORMATION` structure of `FileIoCompletionNotificationInformation` class is "reserved for system use": https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/ne-wdm-_file_information_class. Does it mean this shouldn't be used at all? Or it should be used only internally in the system? 🤔

Usually when "XYZ is reserved for system use" on MSDN documentations it means that such thing is reserved only in appropriate places of the system, in this case of your PR `FileIoCompletionNotificationInformation` can be used in this routine and in whatever places in the kernel or elsewhere is permitted. So you're fine.

PR #4050, opened 2021-10-20
arm64 fixes

konradybcio:
## Purpose

Make arm(64) ports ever so closer to building

JIRA issue: 

## Proposed changes
 
- Don't force x86 specifics on everybody


binarymaster (approved):
We need to move forward, so I have dropped the first commit, and will merge only the second one.

PR #4053, opened 2021-10-20
[WIN32SS:NTUSER] Make sure to hold User Global Lock before playing with desktops.

HBelusca:
## Purpose

Should fix the bug detected by @julcar :
image
image


julcar:
Can confirm this fixes the bug. I managed to reproduce the assertion several times, then applied the patch and now the assertion is gone.

PR #4056, opened 2021-10-21
[NTOS:SE] Implement token filtering

GeoB99:
The support of token filtering allows the kernel to create "filtered" access tokens (or in other words, restricted tokens of regular ones) with disabled SIDs, deleted privileges, etc, so that the token is prevented from accessing the securable objects and such.

`NtFilterToken` is used respectively in `CreateRestrictedToken` ADVAPI32 API routine call. A separate PR for it will be published as soon as this gets merged, since our `CreateRestrictedToken` implementation is mere hacks.

### DIFF
--- a/sdk/include/ndk/sefuncs.h
+++ b/sdk/include/ndk/sefuncs.h
@@ -220,6 +220,19 @@ NtDuplicateToken(
     _In_ TOKEN_TYPE TokenType,
     _Out_ PHANDLE NewTokenHandle);
 
+_Must_inspect_result_
+__kernel_entry
+NTSYSCALLAPI
+NTSTATUS
+NTAPI
+NtFilterToken(
+    _In_ HANDLE ExistingTokenHandle,
+    _In_ ULONG Flags,
+    _In_opt_ PTOKEN_GROUPS SidsToDisable,
+    _In_opt_ PTOKEN_PRIVILEGES PrivilegesToDelete,
+    _In_opt_ PTOKEN_GROUPS RestrictedSids,
+    _Out_ PHANDLE NewTokenHandle);
+
 NTSYSCALLAPI
 NTSTATUS
 NTAPI
--- a/sdk/include/ndk/setypes.h
+++ b/sdk/include/ndk/setypes.h
@@ -109,6 +109,16 @@ typedef struct _TOKEN_ACCESS_INFORMATION
      SE_GROUP_INTEGRITY_ENABLED)
 
 //
+// Privilege token filtering flags
+//
+#define DISABLE_MAX_PRIVILEGE 0x1
+#define SANDBOX_INERT         0x2
+#if (NTDDI_VERSION >= NTDDI_LONGHORN)
+#define LUA_TOKEN             0x4
+#define WRITE_RESTRICTED      0x8
+#endif
+
+//
 // Proxy Class enumeration
 //
 typedef enum _PROXY_CLASS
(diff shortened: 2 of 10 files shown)


sdk/include/ndk/setypes.h
```diff
@@ -29,6 +29,16 @@ Header Name:
...
 
+//
+// Privilege token filtering flags
+//
+#define DISABLE_MAX_PRIVILEGE 0x1
+#define SANDBOX_INERT         0x2
+#if (NTDDI_VERSION >= NTDDI_LONGHORN)
+#define LUA_TOKEN             0x4
+#define WRITE_RESTRICTED      0x8
+#endif
```
HBelusca, line 40: Surprising this needs to be in the NDK (as they appear to be documented... at least in MSDN). Do you know whether they are in some public header in the Win2003 PSDK?

sdk/include/ndk/sefuncs.h
```diff
@@ -165,6 +165,19 @@ NtAdjustPrivilegesToken(
     _Out_writes_bytes_to_opt_(BufferLength, *ReturnLength) PTOKEN_PRIVILEGES PreviousState,
     _When_(PreviousState != NULL, _Out_) PULONG ReturnLength);
 
+_Must_inspect_result_
+__kernel_entry
+NTSYSCALLAPI
+NTSTATUS
+NTAPI
+NtFilterToken(
```
HBelusca, line 173: It appears (at least) the Nt*Token* functions are sorted alphabetically, so this one should certainly go after `NtDuplicateToken`.

sdk/include/ndk/setypes.h
GeoB99: Seemingly these appear in `winnt.h`. Thomas told me that for this case these have to be in NDK, not XDK.

ThFabba: To clarify: they should probably be in both.
* Somewhere in XDK such that they end up in `winnt.h` for UM code
* In NDK for KM code, since other KM headers don't define them and we need them in ntoskrnl

sdk/include/ndk/setypes.h
```diff
@@ -108,6 +118,16 @@ typedef struct _TOKEN_ACCESS_INFORMATION
...
+
```
HBelusca, line 130: Why duplicated? If you want to have them both available for usermode and kernelmode from the NDK itself, just keep the definition that is outside of the `NTOS_MODE_USER` define'd region.

sdk/include/xdk/setypes.h
```diff
@@ -146,6 +146,14 @@ typedef struct _SE_IMPERSONATION_STATE {
...
 
+/* Privilege token filtering flags */
+#define DISABLE_MAX_PRIVILEGE 0x1
+#define SANDBOX_INERT         0x2
+#if (NTDDI_VERSION >= NTDDI_LONGHORN)
+#define LUA_TOKEN             0x4
+#define WRITE_RESTRICTED      0x8
+#endif
+
 $endif (_WDMDDK_ || _WINNT_)
```
ThFabba, line 157: Where you currently have these definitions, they will end up in both `wdm.h` and `winnt.h`. They should move to a section that only has `$if (_WINNT_)`

sdk/include/ndk/setypes.h
```diff
@@ -29,6 +29,16 @@ Header Name:
...
```
ThFabba, line 41: You may want this in `!NTOS_MODE_USER` since UM already gets it from winnt.h

PR #4073, opened 2021-10-25
[HHPCOMP] Check return value when calling chdir

ihsinme:
Good afternoon.
I noticed that your code is using chdir without checking the return value. it may not be safe.
I suggest adding a check.


binarymaster (approved):
Personally I'm ok with this change, even a minor bugfix is an improvement.

PR #4077, opened 2021-10-28
[NTDLL_APITEST] NtDuplicateToken.c: Fix error handling

SergeGautherie:
## Purpose

Properly handle "all" error cases.

## Results

- Windows:
All succeed.
https://testbot.winehq.org/JobDetails.pl?Key=100888
- ReactOS:
Privileges are reduced to 2, as on Windows.
Groups are reduced by 1, whereas no change on Windows. (Intended or not? See also existing test comment.)
Starts with 20/8, whereas WS03 seems to start with 20/10. (Intended or not?)
NB: On LiveCD, `AdjustTokenPrivileges()` fails with Last Error `1300` (`ERROR_NOT_ALL_ASSIGNED`). (Assumed expected?)

### DIFF
--- a/modules/rostests/apitests/ntdll/NtDuplicateToken.c
+++ b/modules/rostests/apitests/ntdll/NtDuplicateToken.c
@@ -19,7 +19,7 @@ OpenTokenFromProcess(VOID)
                                &Token);
     if (!Success)
     {
-        skip("OpenProcessToken() has failed to get the process' token (error code: %lu)!\n", GetLastError());
+        ok(FALSE, "Failed to get the process' token (error code: %lu)\n", GetLastError());
         return NULL;
     }
 
@@ -27,7 +27,7 @@ OpenTokenFromProcess(VOID)
 }
 
 static
-VOID
+BOOL
 DisablePrivilege(
     _In_ HANDLE Token,
     _In_ LPCWSTR PrivilegeName)
@@ -39,25 +39,28 @@ DisablePrivilege(
     Success = LookupPrivilegeValueW(NULL, PrivilegeName, &PrivLuid);
     if (!Success)
     {
-        skip("LookupPrivilegeValueW() has failed to locate the privilege value (error code: %lu)!\n", GetLastError());
-        return;
+        ok(FALSE, "Failed to locate the privilege value (error code: %lu)\n", GetLastError());
+        return FALSE;
     }
 
     TokenPriv.PrivilegeCount = 1;
     TokenPriv.Privileges[0].Luid = PrivLuid;
     TokenPriv.Privileges[0].Attributes = 0;
 
+    SetLastError(0xdeadbeef);
     Success = AdjustTokenPrivileges(Token,
                                     FALSE,
                                     &TokenPriv,
                                     0,
                                     NULL,
                                     NULL);
-    if (!Success)
+    if (!Success || GetLastError() != ERROR_SUCCESS)
     {
-        skip("AdjustTokenPrivileges() has failed to adjust privileges of token (error code: %lu)!\n", GetLastError());
-        return;
+        ok(FALSE, "Failed to adjust privileges of token (error code: %lu)\n", GetLastError());
+        return FALSE;
     }
+
+    return TRUE;
 }
 
 static
@@ -67,9 +70,18 @@ DuplicateTokenAsEffective(VOID)
     NTSTATUS Status;
     ULONG Size;
     HANDLE TokenHandle;
-    HANDLE DuplicatedTokenHandle;
+    HANDLE DuplicatedTokenHandle = NULL;
     OBJECT_ATTRIBUTES ObjectAttributes;
-    PTOKEN_STATISTICS TokenStats;
+    PTOKEN_STATISTICS TokenStats = NULL;
+
+    /* We give a bogus invalid handle */
+    Status = NtDuplicateToken(NULL,
+                              0,
+                              NULL,
+                              TRUE,
+                              TokenPrimary,
+                              NULL);
+    ok_ntstatus(Status, STATUS_ACCESS_VIOLATION);
 
     /* Initialize the object attributes for token duplication */
     InitializeObjectAttributes(&ObjectAttributes,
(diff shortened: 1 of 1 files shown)


modules/rostests/apitests/ntdll/NtDuplicateToken.c
```diff
@@ -67,9 +70,18 @@ DuplicateTokenAsEffective(VOID)
...
+    PTOKEN_STATISTICS TokenStats = NULL;
+
+    /* We give a bogus invalid handle */
+    Status = NtDuplicateToken(NULL,
+                              0,
+                              NULL,
+                              TRUE,
+                              TokenPrimary,
+                              NULL);
+    ok_ntstatus(Status, STATUS_ACCESS_VIOLATION);
```
GeoB99, line 84: Revert it back to where it was before.

modules/rostests/apitests/ntdll/NtDuplicateToken.c
```diff
@@ -80,15 +92,11 @@ DuplicateTokenAsEffective(VOID)
...
-                              NULL,
-                              TRUE,
-                              TokenPrimary,
-                              NULL);
-    ok_hex(Status, STATUS_ACCESS_VIOLATION);
+    if (TokenHandle == NULL)
+    {
+        skip("TokenHandle is NULL\n");
+        return;
+    }
```
GeoB99, line 99: This is pointless because this is a situation that's highly unlikely to happen for the current calling thread of the process to not have the token.

modules/rostests/apitests/ntdll/NtDuplicateToken.c
```diff
@@ -98,30 +106,40 @@ DuplicateTokenAsEffective(VOID)
      * the token effective, with this potential privilege being
      * disabled by ourselves.
      */
-    DisablePrivilege(TokenHandle, L"SeImpersonatePrivilege");
+    if (!DisablePrivilege(TokenHandle, L"SeImpersonatePrivilege"))
+    {
+        skip("DisablePrivilege() failed\n");
+        goto cleanup;
+    }
```
GeoB99, line 113: Revert this.

modules/rostests/apitests/ntdll/NtDuplicateToken.c
```diff
@@ -98,30 +106,40 @@ DuplicateTokenAsEffective(VOID)
...
 
     /* Total size queried, time to allocate our buffer based on that size */
     TokenStats = RtlAllocateHeap(RtlGetProcessHeap(), 0, Size);
+    ok(TokenStats != NULL, "Failed to allocate our token statistics buffer\n");
     if (TokenStats == NULL)
     {
-        skip("Failed to allocate our token statistics buffer!\n");
-        return;
+        skip("TokenStats is NULL\n");
+        goto cleanup;
```
GeoB99, line 133: You're already printing out that token statistics buffer hasn't been allocated due to a failure, no need for a `skip()` call to print the nearly same information again.

modules/rostests/apitests/ntdll/NtDuplicateToken.c
```diff
@@ -134,7 +152,12 @@ DuplicateTokenAsEffective(VOID)
                               TRUE,
                               TokenPrimary,
                               &DuplicatedTokenHandle);
-    ok_hex(Status, STATUS_SUCCESS);
+    ok_ntstatus(Status, STATUS_SUCCESS);
+    if (!NT_SUCCESS(Status))
+    {
+        skip("NtDuplicateToken() failed\n");
+        goto cleanup;
+    }
```
GeoB99, line 160: When the testcase has reached this path `NtDuplicateToken` is supposed to always succeed and give you the duplicated effective token so this extra code is unnecessary.

modules/rostests/apitests/ntdll/NtDuplicateToken.c
```diff
@@ -144,22 +167,28 @@ DuplicateTokenAsEffective(VOID)
...
-        return;
+        skip("Failed to query the token statistics\n");
+        goto cleanup;
     }
 
     trace("Number of privileges of effective only token -- %lu\n", TokenStats->PrivilegeCount);
     trace("Number of groups of effective only token -- %lu\n", TokenStats->GroupCount);
 
+    // TODO: Actually check that counts are lower than before.
+
```
GeoB99, line 181: Remove this TODO.

GeoB99 (changes requested):
>Groups are reduced by 1, whereas no change on Windows. (Intended or not? See also existing test comment.)

`UserAndGroups` member of an access token not only holds groups but also the main user as the first element in the array and what you get in the final results of the tests is the user that's being removed. The user in this context is an administrator (with a SID equivalent to `S-1-5-21-<ID>-500`) and due to the fact that's removed because it's not enabled by default -- a similar behavior with Windows XP where one has such administrator user account disabled.

> Starts with 20/8, whereas WS03 seems to start with 20/10. (Intended or not?)

That's mainly because we aren't assigning two more mandatory groups in the access token upon its creation. I haven't looked deeper which are those but when I get the time for it I'll see.

> NB: On LiveCD, AdjustTokenPrivileges() fails with Last Error 1300 (ERROR_NOT_ALL_ASSIGNED). (Assumed expected?)

`SeImpersonatePrivilege` is a privilege that an object whose token holds it allows to impersonate a client's access token. In this context, virtually every process has it more or less and so is the Command Prompt (as `ntdll_apitest.exe`'s parent process is `cmd.exe`). If on LiveCD the public routine didn't find this specific privilege within the token's privileges of the current process, then that means the blame is on the underlying code and not on the testcase.

Most of the changes you proposed and your doubts as a whole makes me feel you barely have any idea of what you touched so it's better that you don't touch stuff you don't know. The only change that actually makes sense here is the STATUS_BUFFER_TOO_SMALL thing.

RepOfCATV95:
Revert the unnecessary changes as @GeoB99 had said. The code will surely break.

GeoB99:
I'm closing this since you haven't yet and for sure won't provide feedback for the reviews I have posted, as this PR will rot as is just like the others of yours. I seriously can't trust you that way. I'll be going to say for the last time -- **don't change stuff you have no idea about**.

PR #4079, opened 2021-10-31
[KMTESTS:MM] ZwAllocateVirtualMemory.c: Fix expected status

SergeGautherie:
## Purpose

Fix test on Windows.

JIRA issue: [ROSTESTS-262](https://jira.reactos.org/browse/ROSTESTS-262)


modules/rostests/kmtests/ntos_mm/ZwAllocateVirtualMemory.c
```diff
@@ -109,7 +109,7 @@ SimpleErrorChecks(VOID)
 
     //BASE ADDRESS TESTS
     Base = (PVOID)0x00567A20;
-    ALLOC_MEMORY_WITH_FREE(NtCurrentProcess(), Base, 0, RegionSize, (MEM_COMMIT | MEM_RESERVE), PAGE_READWRITE, STATUS_CONFLICTING_ADDRESSES, STATUS_UNABLE_TO_DELETE_SECTION);
+    ALLOC_MEMORY_WITH_FREE(NtCurrentProcess(), Base, 0, RegionSize, (MEM_COMMIT | MEM_RESERVE), PAGE_READWRITE, STATUS_CONFLICTING_ADDRESSES, STATUS_FREE_VM_NOT_AT_BASE);
```
ThFabba, line 112: I think you're likely replacing one "random" return status with another here.
What is at `0x00567A20`? We have no idea -- therefore, this test appears to be nonsensical. Same with the one below.

HBelusca: looks like an attempt at "random memory pointer" :)

SergeGautherie: > this test appears to be nonsensical. Same with the one below.

Should we simply remove these 1-2 tests? Or would you suggest other ones to replace them?
See also `CustomBaseAllocation()` which uses `0x60025000`...

ThFabba: Ideally we could do something like query virtual memory to find an empty region, then use that.
`CustomBaseAllocation` could make use of the same address(es)

PR #4081, opened 2021-11-01
[NTOS:SE] Mark the token as no longer belonging to admin group upon e…

GeoB99:
…ffective duplication

A scenario where it happens that an access token belongs to an administrators group but it's disabled (that is, SeAliasAdminsSid has no attributes or it doesn't have SE_GROUP_ENABLED turn ON), the function removes this group from the token but still has TOKEN_HAS_ADMIN_GROUP flag which can lead to erratic behavior across the kernel and security modules -- implying that the token still belongs to administrators group.

This is an oversight from my part, so I thereby apology for that.

### DIFF
--- a/ntoskrnl/se/token.c
+++ b/ntoskrnl/se/token.c
@@ -1197,12 +1197,27 @@ SepDuplicateToken(
                 (AccessToken->UserAndGroups[GroupsIndex].Attributes & SE_GROUP_ENABLED) == 0)
             {
                 /*
+                 * If this group is an administrators group
+                 * and the token belongs to such group,
+                 * we've to take away TOKEN_HAS_ADMIN_GROUP
+                 * for the fact that's not enabled and as
+                 * such the token no longer belongs to
+                 * this group.
+                 */
+                if (RtlEqualSid(SeAliasAdminsSid,
+                                &AccessToken->UserAndGroups[GroupsIndex].Sid))
+                {
+                    AccessToken->TokenFlags &= ~TOKEN_HAS_ADMIN_GROUP;
+                }
+
+                /*
                  * A group is not enabled, it's time to remove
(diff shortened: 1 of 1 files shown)


ntoskrnl/se/token.c
```diff
@@ -1196,13 +1196,28 @@ SepDuplicateToken(
...
             {
+                /*
+                 * If this group is an administrators group
+                 * and the token belongs to such group,
+                 * we've to take away TOKEN_HAS_ADMIN_GROUP
+                 * for the fact that's not enabled and as
+                 * such the token no longer belongs to
+                 * this group.
+                 */
+                if (RtlEqualSid(SeAliasAdminsSid,
```
ThFabba, line 1207: Can there be multiple entries for the same SID in the `UserAndGroups` array? If so, you should probably only remove the flag if all of them get removed.

GeoB99: No, there cannot be multiple elements of the same SID in the array, at least I have never been aware of nor Windows seems to do any checks regarding this circumstance (and personally, does it make sense to have such entries having the same SID, `SeAliasAdminsSid` in particular?).

ThFabba: No, I didn't have any particular use case in mind -- was just checking to make sure, since I noticed it was an implicit assumption that the logic depends on.

ntoskrnl/se/token.c
```diff
@@ -1228,6 +1243,7 @@ SepDuplicateToken(
                  */
                 SepRemovePrivilegeToken(AccessToken, PrivilegesIndex);
                 PrivilegesIndex--;
+                continue;
```
HBelusca, line 1246: What for-loop is this continue meant to "continue" ?

PR #4083, opened 2021-11-02
[KS] Restore old hack for MS portcls

oleg-dubinskiy:
## Purpose

Restore an old hack required by MS portcls driver.
Replace current hack from https://git.reactos.org/?p=reactos.git;a=blob;f=drivers/ksfilter/ks/irp.c;hb=22d1e7a4e4175b2d579428493434eaab28323cca#l1665 by old one that was removed in the following commit: https://git.reactos.org/?p=reactos-deprecated-gitsvn-dont-use.git;a=commitdiff;h=563682ad9a5751a7d5c3f9ffb4ac4f67bf0adc71.
This allows to boot ReactOS (and play the sound) properly again with portcls.sys driver from Windows XP/2003. :smiley: 
Note that in order to get the best result from this fix, you need to use legacy APIs in wdmaud.drv, instead of MMixer APIs, those are currently enabled by default. You can switch to them by removing (or just commenting out) `USE_MMIXER_LIB` define in dll/win32/wdmaud.drv/wdmaud.c.

JIRA issue: [CORE-17237](https://jira.reactos.org/browse/CORE-17237)

## Result

As visible on the following screenshots, ReactOS is now loading portcls.sys driver as from Windows XP SP3:
portcls-xpsp3

same with version from Windows Server 2003 SP2:
portcls-2k3sp2

### DIFF
--- a/drivers/ksfilter/ks/irp.c
+++ b/drivers/ksfilter/ks/irp.c
@@ -919,6 +919,9 @@ ProbeMdl:
             goto ProbeMdl;
     }
 
+    // HACK for MS portcls
+    HeaderSize = Length;
+
     /* probe user mode buffers */
     if (Length && ( (!HeaderSize) || (Length % HeaderSize == 0) || ((ProbeFlags & KSPROBE_ALLOWFORMATCHANGE) && (Length == sizeof(KSSTREAM_HEADER))) ) )
     {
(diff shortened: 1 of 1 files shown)


drivers/ksfilter/ks/irp.c
```diff
@@ -2032,6 +2023,9 @@ KsSetMajorFunctionHandler(
 {
     DPRINT("KsSetMajorFunctionHandler Function %x\n", MajorFunction);
 
+    // HACK for MS portcls
+    DriverObject->MajorFunction[IRP_MJ_CREATE] = KspCreate;
+
```
hpoussin, line 2028: Can it be because we don't handle KSDISPATCH_FASTIO flag ? According to MSDN, this flag "may be added to the MajorFunction identifier".
It should be a matter of setting function pointers in DriverEntry->FastIoDispatch.

oleg-dubinskiy: > Can it be because we don't handle KSDISPATCH_FASTIO flag ?

I can't say exactly, but most likely no, because MSDN also says that "This is valid only with IRP_MJ_READ, IRP_MJ_WRITE or IRP_MJ_DEVICE_CONTROL." In our case, driver calls `IRP_MJ_CREATE`, so that flag seems not be used.

PR #4085, opened 2021-11-05
[RTL] Add 3 DPRINT() when waiting for a critical section

SergeGautherie:
## Purpose

Incentive/Help to investigate.

JIRA issue: CORE 12828, for example.

## Examples

```
(/sdk/lib/rtl/critical.c:512) Thread 0000029C waits for critical section 0025714C, owned by thread 00000228
(/sdk/lib/rtl/critical.c:517) Thread 0000029C acquired critical section 0025714C

(/sdk/lib/rtl/critical.c:512) Thread 00000400 waits for critical section 00BD50EC, owned by thread 00000418
(/sdk/lib/rtl/critical.c:185) (1/2) Timeout: critical section 00BD50EC, owned by thread 00000418
ERROR:  RtlpWaitForCriticalSection at /sdk/lib/rtl/critical.c:172
(2/2) Deadlock: critical section 00BD50EC, owned by thread 00000418
```

## Note

The existing `ERROR_DBGBREAK()` logs fine, but does not trigger the debugger:
Is there a bug somewhere else? Should it be reduced to a `DPRINT1()`?

### DIFF
--- a/sdk/lib/rtl/critical.c
+++ b/sdk/lib/rtl/critical.c
@@ -168,7 +168,8 @@ RtlpWaitForCriticalSection(PRTL_CRITICAL_SECTION CriticalSection)
             /* Is this the 2nd time we've timed out? */
             if (LastChance)
             {
-                ERROR_DBGBREAK("Deadlock: 0x%p\n", CriticalSection);
+                ERROR_DBGBREAK("(2/2) Deadlock: critical section 0x%p, owned by thread 0x%p\n",
+                               CriticalSection, CriticalSection->OwningThread);
 
                 /* Yes it is, we are raising an exception */
                 ExceptionRecord.ExceptionCode    = STATUS_POSSIBLE_DEADLOCK;
@@ -179,6 +180,10 @@ RtlpWaitForCriticalSection(PRTL_CRITICAL_SECTION CriticalSection)
                 ExceptionRecord.ExceptionInformation[0] = (ULONG_PTR)CriticalSection;
                 RtlRaiseException(&ExceptionRecord);
             }
+
+            /* Incentive to check the blocking thread */
+            DPRINT1("(1/2) Timeout: critical section 0x%p, owned by thread 0x%p\n",
+                    CriticalSection, CriticalSection->OwningThread);
 
             /* One more try */
             LastChance = TRUE;
@@ -505,7 +510,10 @@ RtlEnterCriticalSection(PRTL_CRITICAL_SECTION CriticalSection)
                   OwningThread is NULL here! */
 
         /* We don't own it, so we must wait for it */
+        DPRINT("Thread 0x%p waits for critical section 0x%p, owned by thread 0x%p\n",
+               Thread, CriticalSection, CriticalSection->OwningThread);
         RtlpWaitForCriticalSection(CriticalSection);
+        DPRINT("Thread 0x%p acquired critical section 0x%p\n", Thread, CriticalSection);
     }
 
     /*


learn-more (changes requested):
I am not a fan of adding debug prints on master to all critical sections for the purpose of debugging a random ticket.

If you want to debug locks, WinDbg is far more usefull than a DPRINT that will be ignore by everyone and will just generate more noise when booting:
https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/displaying-a-critical-section

Considering this is just my opinion, I'd like to hear some more before outright closing this.

ThFabba:
The DPRINT1 is in a code path where we've already hit the timeout once (looks to be one hour by default?). I think that's a useful addition.
The others are DPRINTs, which I don't have strong feelings about either way.
So I'm okay with this.

GeoB99:
If the sole reasoning for this PR is to just investigate a specific issue regarding dead locks then I see no point on having this patch be merged in master. I whole-fully agree with Mark, it's way better investing a problem by doing a thorough step-by-step debugging with WinDBG (or just release this diff patch to that specific ticket in JIRA for investigation).

sdk/lib/rtl/critical.c
```diff
@@ -168,7 +168,8 @@ RtlpWaitForCriticalSection(PRTL_CRITICAL_SECTION CriticalSection)
             /* Is this the 2nd time we've timed out? */
             if (LastChance)
             {
-                ERROR_DBGBREAK("Deadlock: 0x%p\n", CriticalSection);
+                ERROR_DBGBREAK("(2/2) Deadlock: critical section %p, owned by thread %p\n",
```
HBelusca, line 171: Please use `0x` in front of the `%p` so that a reader of a debug log doesn't have to guesscheck what a value like `47031250` could be.

SergeGautherie: Done. (Tell me if you want that globally in this file.)

HBelusca: It's fine (was just for the new dprints).

SergeGautherie:
> adding debug prints on master to all critical sections for the purpose of debugging a random ticket.

I am not sure what you mean. This PR is not specific to any ticket, it applies to all c.s. timeouts.

> WinDbg is far more usefull than a DPRINT that will be ignore by everyone and will just generate more noise when booting: https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/displaying-a-critical-section

What about kdbg, on gcc builds?
I disagree with that, but, if noone cares about this log, let's demote the `ERROR_DBGBREAK()`, then/instead...
If booting triggers the timeout, I would think the root cause should be fixed instead.

---

> looks to be one hour by default?

When reproducing CORE 12828 case, timeout seems to happen after about a few minutes...

---

> sole reasoning for this PR is to just investigate a specific issue regarding dead locks

It is not: I already investigated 2 tickets.
It is about better reporting c.s. timeouts and deadlocks, and optionally helping to investigate them.

learn-more:
> > adding debug prints on master to all critical sections for the purpose of debugging a random ticket.
> 
> I am not sure what you mean. This PR is not specific to any ticket, it applies to all c.s. timeouts.
> 
> > WinDbg is far more usefull than a DPRINT that will be ignore by everyone and will just generate more noise when booting: https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/displaying-a-critical-section
> 
> What about kdbg, on gcc builds? I disagree with that, but, if noone cares about this log, let's demote the `ERROR_DBGBREAK()`, then/instead... If booting triggers the timeout, I would think the root cause should be fixed instead.

No clue about kdbg, and I see no reason to use that.


> > looks to be one hour by default?
> 
> When reproducing CORE 12828 case, timeout seems to happen after about a few minutes...
> 
> > sole reasoning for this PR is to just investigate a specific issue regarding dead locks
> 
> It is not: I already investigated 2 tickets. It is about better reporting c.s. timeouts and deadlocks, and optionally helping to investigate them.

k

SergeGautherie:
> No clue about kdbg, and I see no reason to use that.

Because gcc/kdbg is official, for testers. Or you may want to disable it as well?

> k

And there are various unresolved deadlock tickets, which could benefit from this PR.

PR #4086, opened 2021-11-06
[MSAFD] Don't wait on receive if socket is non-blocking

oleg-dubinskiy:
## Purpose

Improve pending check for `IOCTL_AFD_RECV` call in `WSPRecv`, as done in WSA async patch and change the socket to non-blocking to make the check succeed.
2nd change seems to be hackish.
This allows to properly run more recent versions of Mozilla Firefox (49.0 - 52.9.0 ESR) and other Gecko-based browsers in ReactOS! :smiley: 

JIRA issue: [CORE-12828](https://jira.reactos.org/browse/CORE-12828)

## Result

Here's the screenshot with Firefox 52.9.0 ESR running in ReactOS (now without any workarounds from user, like replacing any dlls):
ff-52 9 0-fixed

Websites are opening as well. :slightly_smiling_face: 

But unfortunately, after these changes, Rapps can't update its database! :frowning_face: But it also can't update it on (all versions of) Windows, so, since it is closer to the Windows behaviour, it might be correct. Probably the fix is needed for Rapps instead.

### DIFF
--- a/dll/win32/msafd/misc/sndrcv.c
+++ b/dll/win32/msafd/misc/sndrcv.c
@@ -173,6 +173,7 @@ WSPRecv(SOCKET Handle,
     HANDLE                  Event = NULL;
     HANDLE                  SockEvent;
     PSOCKET_INFORMATION     Socket;
+    BOOLEAN                 BlockMode;
 
     TRACE("Called (%x)\n", Handle);
 
@@ -202,6 +203,11 @@ WSPRecv(SOCKET Handle,
 
     if( !NT_SUCCESS(Status) )
         return -1;
+
+    /* HACK: change the Socket to Non Blocking. See CORE-12828. */
+    BlockMode = TRUE;
+    SetSocketInformation(Socket, AFD_INFO_BLOCKING_MODE, &BlockMode, NULL, NULL, NULL, NULL);
+    Socket->SharedData->NonBlocking = TRUE;
 
     /* Set up the Receive Structure */
     RecvInfo.BufferArray = (PAFD_WSABUF)lpBuffers;
@@ -292,12 +298,12 @@ WSPRecv(SOCKET Handle,
                                    0);
 
     /* Wait for completion of not overlapped */
-    if (Status == STATUS_PENDING && lpOverlapped == NULL)
+    if (Status == STATUS_PENDING)
     {
         /* It's up to the protocol to time out recv.  We must wait
          * until the protocol decides it's had enough.
          */
-        WaitForSingleObject(SockEvent, INFINITE);
+        WaitForSingleObject(SockEvent, (lpOverlapped || Socket->SharedData->NonBlocking) ? 0 : INFINITE);
         Status = IOSB->Status;
     }
 


JoachimHenze:
what would happen, if we waited for - let's say five seconds ? Given the **hypothetical assumption** that this would work, wouldn't that be a slightly smaller hack then?

oleg-dubinskiy:
> what would happen, if we waited for - let's say five seconds ? Given the **hypothetical assumption** that this would work, wouldn't that be a slightly smaller hack then?

At all, FF starts successfully even if I just change `INFINITE` to `0` in that `WaitForSingleObject` call, but then some asserts appear from tcpip sometimes. :slightly_frowning_face: With my current changes, asserts are not happening.

JoachimHenze:
And when just changing it to 5000ms? without any other change?

oleg-dubinskiy:
> And when just changing it to 5000ms? without any other change?

Just tested. It starts successfully, however with a little delay. And Rapps now updates the database again. :slightly_smiling_face:

oleg-dubinskiy:
But now FF works very slowly, even without debugging. :frowning_face:

JoachimHenze:
Well then the hack can at least be simplified a bit. Still it's a hack, although an interesting one. Question is still how big would be the risk to break anything else with that. I do recommend that you test the 5000ms variant locally for a week at least and have an eye on networking performance specifically.

Extravert-ir (changes requested):
Let's not pretend something works when it doesn't. This does not have benefits at all - the browser starts, but is not useful still.

oleg-dubinskiy:
> Let's not pretend something works when it doesn't. This does not have benefits at all - the browser starts, but is not useful still.

It actually starts and works as should with current state of this PR, but does not if we wait for 5000 ms.

dll/win32/msafd/misc/sndrcv.c
```diff
@@ -203,6 +204,11 @@ WSPRecv(SOCKET Handle,
     if( !NT_SUCCESS(Status) )
         return -1;
 
+    /* HACK: change the Socket to Non Blocking. See CORE-12828. */
+    BlockMode = TRUE;
+    SetSocketInformation(Socket, AFD_INFO_BLOCKING_MODE, &BlockMode, NULL, NULL, NULL, NULL);
+    Socket->SharedData->NonBlocking = TRUE;
```
GeoB99, line 210: I seriously doubt Windows' `WSPRecv` really tampers with the `NonBlocking` member state of the socket (do you have any user testcase that exactly shows Windows really does that?). Taking a look at the function on a first glance it seems it merely receives data from the socket so it feels kind of weird that you're toying with the socket in that way.

dll/win32/msafd/misc/sndrcv.c
```diff
@@ -292,12 +298,12 @@ WSPRecv(SOCKET Handle,
...
 
     /* Wait for completion of not overlapped */
-    if (Status == STATUS_PENDING && lpOverlapped == NULL)
+    if (Status == STATUS_PENDING)
     {
         /* It's up to the protocol to time out recv.  We must wait
          * until the protocol decides it's had enough.
          */
-        WaitForSingleObject(SockEvent, INFINITE);
+        WaitForSingleObject(SockEvent, (lpOverlapped || Socket->SharedData->NonBlocking) ? 0 : INFINITE);
```
GeoB99, line 306: With this change now the comment above feels contradictory to what the inner logic does right now as now the function returns immediately because it didn't wait due to the fact your forced the socket to be in a non blocking state.

ThFabba: In any "non-blocking" case, you shouldn't go into this code path at all, because
* the wait with zero timeout does nothing
* reading `IOSB->Status` while the operation is still in progress means you have no guarantees about what value you'll get.
So just keep the `&& lpOverlapped == NULL` condition above, and add your `NonBlocking` condition in addition to that.

oleg-dubinskiy: Thanks @ThFabba for the explanation. So, as I understood, I should just add `&& !Socket->SharedData->NonBlocking` condition to the existing check. But the problem is, in this case, that code path will be called anyway, because `NonBlocking` is actually `FALSE` when `WSPRecv` function is called. So the condition still always succeeds.
I investigated the problem a bit, and now see that Firefox first creates the socket via calling `WSPSocket` function, and then passes its handle to `WSPRecv`. We're always creating blocking sockets by default, since `NonBlocking` is always set to `FALSE` in the first function: https://github.com/reactos/reactos/blob/master/dll/win32/msafd/misc/dllmain.c#L214. After changing it to `TRUE` manually, everything seems to work properly again.
But I'm not sure whether this is correct and want to inspect this member in Windows via WinDBG somehow, to check it exactly. But I don't know how. :frowning_face:

ThFabba: Yes, the function is probably correct as-is, and your added code is just a workaround for some other limitation. My suggestion was mostly to make your workaround more concise and less questionable in terms of coding practices; it wouldn't change anything about it being a workaround.

dll/win32/msafd/misc/sndrcv.c
oleg-dubinskiy: OK, I just did some investigation. I guess that in Windows, the `NonBlocking` member is always set to `TRUE` in `WSPSocket` instead, unlike our implementation. ROS shows me the following offsets:
`fixme:(dll/win32/msafd/misc/dllmain.c:76) Offset of NonBlocking is 35, offset of SharedData is 4`
when I'm specifying the following:
`FIXME("Offset of NonBlocking is %x, offset of SharedData is %x\n", offsetof(SOCK_SHARED_INFO, NonBlocking), offsetof(SOCKET_INFORMATION, SharedData));`
How can I calculate the offset properly? :thinking: I guess I should just append the two offsets. If to append the 1st offset to the 2nd, then it seems to be 39.
The most related to this value which I found in unassemble of `WSPSocket` function, is 40. There is specified an `ebx` register with appending this offset and the `BOOLEAN` value to `TRUE` (1):
nonblocking_investigations
The total offset is higher on 1 probably because there is some additional member in any of these structures, about which we don't know. But it's just my guessing.
**This judgement is valid only in case I calculated the offset correctly.** If no, then how can I calculate it correctly instead? :thinking:

GeoB99:
I'm going to close this PR since the risk to benefits ratio is way too high for having a hack like this shipped in master branch especially as you're tampering with sockets' blocking mode state in this way. When you get the opportunity to fix MSAFD you can come back with a proper fix.

julenuri:
To quote, #5575 partially supersedes this one.

PR #4089, opened 2021-11-07
Some random improvements around timeouts for critical sections and resources.

HBelusca:
## Purpose

See PR's title.

Open for discussions.


dll/ntdll/ldr/ldrinit.c
```diff
@@ -86,7 +86,12 @@ ULONG LdrpActiveUnloadCount;
 //extern LIST_ENTRY RtlCriticalSectionList;
 
 VOID NTAPI RtlpInitializeVectoredExceptionHandling(VOID);
-VOID NTAPI RtlpInitDeferedCriticalSection(VOID);
+
+#if (NTDDI_VERSION < NTDDI_WS03)
+#define RtlpInitDeferredCriticalSection RtlpInitDeferedCriticalSection
```
learn-more, line 91: If you define both and have one just call the other there's no need for an ifdef.

dll/ntdll/ldr/ldrinit.c
```diff
@@ -1839,7 +1844,7 @@ LdrpInitializeProcess(IN PCONTEXT Context,
     /* ReactOS specific: do not clear it. (Windows starts doing the same in later versions) */
     //Peb->pShimData = NULL;
 
-    /* Save the number of processors and CS Timeout */
+    /* Save the number of processors and the critical section timeout */
```
learn-more, line 1847: Don't really see the point of comment changes like this.

dll/ntdll/ldr/ldrinit.c
tkreuzer: Why is this needed at all? It's an internal function, not exported anywhere.

tkreuzer (commented):
Looks mostly good to me, except that function rename, which seems pointless.

HBelusca:
Thanks for the review, I'll revive this PR one of these next days and get it merged.

HBelusca:
actually merged in master

PR #4091, opened 2021-11-07
[WINLOGON] Implement system sounds.

TAN-Gaming:
[WINLOGON] Implement system sounds.

Working sounds
- Logon sound (improved)
- Logoff sound
- MessageBox and MessageBeep sounds

### DIFF
--- a/base/system/winlogon/sas.c
+++ b/base/system/winlogon/sas.c
@@ -559,16 +559,16 @@ DWORD WINAPI PlayLogoffSoundThread(LPVOID lpParameter)
     return 0;
 }
 
-static
-VOID
-PlayLogonSound(
-    IN OUT PWLSESSION Session)
+static void PlayLogonSound(PWLSESSION Session)
 {
     HANDLE hThread;
 
-    hThread = CreateThread(NULL, 0, PlayLogonSoundThread, (PVOID)Session, 0, NULL);
+    hThread = CreateThread(NULL, 0, (LPTHREAD_START_ROUTINE)PlayLogonSoundThread, (LPVOID)Session, 0, NULL);
+
     if (hThread)
+    {
         CloseHandle(hThread);
+    }
 }
 
 static void PlayLogoffSound(PWLSESSION Session)
--- a/base/system/winlogon/sas.c
+++ b/base/system/winlogon/sas.c
@@ -54,6 +54,13 @@ typedef enum tagWINLOGON_SYSTEM_SOUND
 
 } WINLOGON_SYSTEM_SOUND, *PWINLOGON_SYSTEM_SOUND;
 
+typedef struct tagWINLOGON_PLAYSOUND_DATA
+{
+    WINLOGON_SYSTEM_SOUND Sound;
+    PWLSESSION Session;
+
+} WINLOGON_PLAYSOUND_DATA, *PWINLOGON_PLAYSOUND_DATA;
+
 static BOOL ExitReactOSInProgress = FALSE;
 
 LUID LuidNone = {0, 0};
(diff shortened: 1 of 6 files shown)


base/system/winlogon/sas.c
```diff
@@ -279,153 +279,305 @@ PlaySoundRoutine(
     return Ret;
 }
 
-DWORD
-WINAPI
-PlayLogonSoundThread(
-    IN LPVOID lpParameter)
+DWORD WINAPI PlayLogonSoundThread(LPVOID lpParameter)
```
HBelusca, line 282: Please revert.

base/system/winlogon/sas.c
```diff
@@ -279,153 +279,305 @@ PlaySoundRoutine(
...
+
+    hThread = CreateThread(NULL, 0, (LPTHREAD_START_ROUTINE)PlayLogonSoundThread, (LPVOID)Session, 0, NULL);
+
+    if (hThread)
+    {
+        CloseHandle(hThread);
+    }
+}
+
+static void PlayLogoffSound(PWLSESSION Session)
```
HBelusca, line 571: There seems to be quite redundancy between these functions, whose behaviour just differ by only **one** thing (the thread function)!

base/system/winlogon/sas.c
```diff
@@ -1344,38 +1497,142 @@ UnregisterHotKeys(
...
-        break;
-    case MB_ICONQUESTION:
-        EventName = L"SystemQuestion";
-        break;
-    default:
-        WARN("Unhandled type %d\n", uType);
-        EventName = L"SystemDefault";
-    }
-
-    return PlaySoundRoutine(EventName, FALSE, SND_ALIAS | SND_NOWAIT | SND_NOSTOP | SND_ASYNC);
```
HBelusca, line 1378: What's the rationale behind why it's not possible to directly use the PlaySoundRoutine as is?

base/system/winlogon/sas.c
```diff
@@ -279,153 +279,305 @@ PlaySoundRoutine(
...
+    HeapFree(GetProcessHeap(), 0, lpSndPath);
+
     return 0;
 }
 
-static
-VOID
-PlayLogonSound(
-    IN OUT PWLSESSION Session)
+static void PlayLogonSound(PWLSESSION Session)
```
HBelusca, line 559: Please revert.

base/system/winlogon/sas.c
```diff
@@ -279,153 +279,305 @@ PlaySoundRoutine(
...
+    /* Play sound */
+    PlaySoundRoutine(lpSndPath, TRUE, SND_FILENAME | SND_NODEFAULT);
+
+    HeapFree(GetProcessHeap(), 0, lpRegVal);
+    HeapFree(GetProcessHeap(), 0, lpSndPath);
+
+    return 0;
+}
+
+DWORD WINAPI PlayLogoffSoundThread(LPVOID lpParameter)
```
HBelusca, line 449: Both logon and logoff functions seem to be quite similar; isn't it possible to merge both together, and have the actual different behaviour be controlled by a parameter?

base/system/winlogon/sas.c
TAN-Gaming: PlaySound function with SND_ASYNC flag calls CreateThread function that doesn't work with user impersonation.

base/system/winlogon/sas.c
TAN-Gaming: Yep

base/system/winlogon/sas.c
TAN-Gaming: Yep

base/system/winlogon/sas.c
TAN-Gaming: Updated.

base/system/winlogon/sas.c
TAN-Gaming: Updated.

oleg-dubinskiy (commented):
As you might notice, `PlaySoundRoutine` that is called from these functions, delay-loads `PlaySoundW` from winmm.dll, which actually does all work for playing the system sound events (like done in Windows).
Please note that some of our current winmm implementations are not working correctly (in particular, `get_mmioFromProfile`, which is responsible for `SND_ALIAS` flag, receives an empty string from registry values). Although you didn't call that `SND_ALIAS` flag in your code, nevertheless it's important anyway I think. I'm not sure is it caused by winlogon issues (those might be fixed by your PR), but winmm bugs still might prevent to playing the sound events properly. Btw, I made some attempts to fix that in the past, but for now without luck. :frowning_face: 
See dll/win32/winmm/playsound.c file for the reference: https://git.reactos.org/?p=reactos.git;a=blob;f=dll/win32/winmm/playsound.c;h=a6fdbd7ed195285d9d799936000321cab7aa13f3;hb=HEAD. And feel free to test your changes locally with enabled sound events (set any wav sound files as sound events manually in Control Panel). :wink:

TAN-Gaming:
@oleg-dubinskiy I found that  `get_mmioFromProfile` doesn't work with winlogon because it reads registry values from HKEY_CURRENT_USER (predefined key) so in this case, user impersonation is needed to call `RegOpenCurrentUser` and access user's HKCU

TAN-Gaming:
I've fixed my fork and opened a new PR.

PR #4094, opened 2021-11-09
[SHELL32] Improve empty recycle bin sound code.

TAN-Gaming:
[SHELL32] Improve empty recycle bin sound code.

- Open .Default registry key if failed to open .Current
- There's no need to define _WIN32_WINNT 0x0600 for registry functions.
- No MAX_PATH limit
- Code rewrite

### DIFF
--- a/dll/win32/shell32/folders/CRecycleBin.cpp
+++ b/dll/win32/shell32/folders/CRecycleBin.cpp
@@ -982,6 +982,91 @@ TRASH_TrashFile(LPCWSTR wszPath)
     return DeleteFileToRecycleBin(wszPath);
 }
 
+static void TRASH_PlayEmptyRecycleBinSound()
+{
+    LPCWSTR lpRegSubKey = L"AppEvents\\Schemes\\Apps\\Explorer\\EmptyRecycleBin";
+    HKEY hRegKey, hRegSnd;
+    LPWSTR lpRegVal, lpSndPath;
+    DWORD dwRegValType, dwRegValSize, dwSndPathLen;
+    LONG lRegStatus;
+
+    /* Open registry key */
+    lRegStatus = RegOpenKeyExW(HKEY_CURRENT_USER, lpRegSubKey, 0, KEY_QUERY_VALUE, &hRegKey);
+
+    if (lRegStatus != ERROR_SUCCESS)
+    {
+        return;
+    }
+
+    /* Open .Current */
+    lRegStatus = RegOpenKeyExW(hRegKey, L".Current", 0, KEY_QUERY_VALUE, &hRegSnd);
+
+    if (lRegStatus != ERROR_SUCCESS)
+    {
+        /* If fail then open .Default */
+        lRegStatus = RegOpenKeyExW(hRegKey, L".Default", 0, KEY_QUERY_VALUE, &hRegSnd);
+
+        RegCloseKey(hRegKey);
+
+        if (lRegStatus != ERROR_SUCCESS)
+        {
+            return;
+        }
+    }
+
+    RegCloseKey(hRegKey);
+
+    /* Get registry value size */
+    lRegStatus = RegQueryValueExW(hRegSnd, NULL, NULL, &dwRegValType, NULL, &dwRegValSize);
+
+    /* Check whether the type is valid */
+    if (lRegStatus != ERROR_SUCCESS || (dwRegValType != REG_SZ && dwRegValType != REG_EXPAND_SZ))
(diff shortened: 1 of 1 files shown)


dll/win32/shell32/folders/CRecycleBin.cpp
```diff
@@ -982,6 +982,91 @@ TRASH_TrashFile(LPCWSTR wszPath)
...
+
+    /* Open .Current */
+    lRegStatus = RegOpenKeyExW(hRegKey, L".Current", 0, KEY_QUERY_VALUE, &hRegSnd);
+
+    if (lRegStatus != ERROR_SUCCESS)
+    {
+        /* If fail then open .Default */
+        lRegStatus = RegOpenKeyExW(hRegKey, L".Default", 0, KEY_QUERY_VALUE, &hRegSnd);
+
+        RegCloseKey(hRegKey);
```
learn-more, line 1009: You are closing this key below again, that's not valid.

dll/win32/shell32/folders/CRecycleBin.cpp
```diff
@@ -982,6 +982,91 @@ TRASH_TrashFile(LPCWSTR wszPath)
...
+
+    /* Check whether the type is valid */
+    if (lRegStatus != ERROR_SUCCESS || (dwRegValType != REG_SZ && dwRegValType != REG_EXPAND_SZ))
+    {
+        RegCloseKey(hRegSnd);
+        return;
+    }
+
+    /* Allocate buffer for registry value */
+    lpRegVal = (LPWSTR)HeapAlloc(GetProcessHeap(), 0, (SIZE_T)dwRegValSize);
```
learn-more, line 1030: Ditto, this is c++ why not use `CComHeapPtr<WCHAR>` instead of manually handling that?

oleg-dubinskiy (changes requested):
The question is still actual: is this fix really better than my #3721? Does it make more sense? Now we have two different fixes for the same issue. :frowning_face:

dll/win32/shell32/folders/CRecycleBin.cpp
TAN-Gaming: Fixed.

dll/win32/shell32/folders/CRecycleBin.cpp
```diff
@@ -982,6 +982,91 @@ TRASH_TrashFile(LPCWSTR wszPath)
...
+static void TRASH_PlayEmptyRecycleBinSound()
+{
+    LPCWSTR lpRegSubKey = L"AppEvents\\Schemes\\Apps\\Explorer\\EmptyRecycleBin";
+    HKEY hRegKey, hRegSnd;
+    LPWSTR lpRegVal, lpSndPath;
+    DWORD dwRegValType, dwRegValSize, dwSndPathLen;
+    LONG lRegStatus;
+
+    /* Open registry key */
+    lRegStatus = RegOpenKeyExW(HKEY_CURRENT_USER, lpRegSubKey, 0, KEY_QUERY_VALUE, &hRegKey);
```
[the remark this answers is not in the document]
TAN-Gaming: Isn't it depends on ATL? Why create more dependency?

dll/win32/shell32/folders/CRecycleBin.cpp
TAN-Gaming: It's depends on ATL.

TAN-Gaming:
@oleg-dubinskiy It works mostly the same with your code. I just make some improvements.

Details
- Move the code to saparated function, easier to maintain.
- Open `.Default` key if failed to open `.Current`.
- Add `SND_ASYNC` flag to the `PlaySoundW` function so the `SHEmptyRecycleBinW` function doesn't have to wait for finish playing sound.

dll/win32/shell32/folders/CRecycleBin.cpp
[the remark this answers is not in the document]
learn-more: Yes, atl's CRegKey depends on ATL.
But how is this adding a dependency?
Most of our shell code already uses ATL because having to write the same 100 lines of code for every object is not a job for a human, so we use a library for that.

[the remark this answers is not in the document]
TAN-Gaming: OK, I get it, but I don't think CRegKey will reduce the lines of my code so much compared to WinAPI functions (WinAPI just requires more function arguments) and I have more control using WinAPI.

PR #4097, opened 2021-11-09
[NTOS:SE] Validate the SID lengths when capturing them

GeoB99:
SIDs are variadic by nature which means their lengths can vary in a given amount of time and certain factors that allow for this happen. This also especially can lead to issues when capturing SIDs and attributes because SeCaptureSidAndAttributesArray might end up overwriting the buffer during the time it's been called.

Therefore when we're copying the SIDs, validate their lengths. In addition to that, update the documentation header accordingly.

### DIFF
--- a/ntoskrnl/se/sid.c
+++ b/ntoskrnl/se/sid.c
@@ -52,6 +52,12 @@ PSID SeAnonymousLogonSid = NULL;
 PSID SeLocalServiceSid = NULL;
 PSID SeNetworkServiceSid = NULL;
 
+typedef struct _SID_VALIDATE
+{
+    UCHAR SubAuthorityCount;
+    PISID ProbeSid;
+} SID_VALIDATE, *PSID_VALIDATE;
+
 /* FUNCTIONS ******************************************************************/
 
 /**
@@ -450,8 +456,29 @@ SepReleaseSid(
  * avoiding integer overflows. STATUS_INSUFFICIENT_RESOURCES
  * is returned if memory pool allocation for the captured SID has failed.
  * STATUS_BUFFER_TOO_SMALL is returned if the length of the allocated
- * buffer is less than the required size. A failure NTSTATUS code is
- * returned otherwise.
+ * buffer is less than the required size. STATUS_INVALID_SID is returned
+ * if a SID doesn't meet certain requirements to be considered a valid
+ * SID by the security manager. A failure NTSTATUS code is returned
+ * otherwise.
+ *
+ * @remarks
+ * A security identifier (SID) is a variable-length data structure that
+ * can change in size over time, depending on the factors that influence
+ * this effect in question. An attacker can take advantage of this fact
+ * and can potentially modify certain properties of a SID making it
+ * different in size than it was originally before. This is what we'd
+ * call, a TOCTOU vulnerability.
+ *
+ * For this reason, the logic of copying the SIDs and their attributes
+ * into a new buffer goes like this: first, allocate a buffer array
+ * that just holds the lengths and subauthority count of each SID.
+ * Such information is copied in the first loop. Then in a second loop,
+ * iterate over the array with SID provided and copy them into the final
+ * array. The moment we're doing this, validate the lengths of each SID
+ * basing upon the captured lengths we've got before. In this way we
+ * ensure that the SIDs have remained intact. The validation checks are
+ * done in user mode as a general rule that we just cannot trust UM and
+ * whatever data is coming from it.
  */
 NTSTATUS
 NTAPI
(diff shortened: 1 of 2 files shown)


ntoskrnl/se/sid.c
```diff
@@ -495,9 +525,29 @@ SeCaptureSidAndAttributesArray(
...
 
-    /* Check for user mode data */
     if (PreviousMode != KernelMode)
     {
+        /* We're in user mode, set up the size for the temporary array */
+        TempArrayValidate = AttributeCount * sizeof(SID_VALIDATE_LENGTHS);
+        TempLengthValidate = ALIGN_UP_BY(TempArrayValidate, sizeof(ULONG));
+
+        /*
+         * Allocate a buffer for the array that
```
HBelusca, line 535: you may be more flexible on the number of characters per line of your comments! (remember we usually use 80 chars per line)

ntoskrnl/se/sid.c
```diff
@@ -598,7 +667,32 @@ SeCaptureSidAndAttributesArray(
...
+                 *      by the security manager that it's valid.
+                 *
+                 * Any assertion violation of these conditions means
+                 * that there's been an attempt on changing the SID
+                 * properties whilst the function was copying the
+                 * SIDs into the final buffer. This can bring a lot
+                 * of issues so at this point just bail out.
+                 */
+                ASSERT(ValidateArray[i].SubAuthorityCount == Sid->SubAuthorityCount);
+                ASSERT(RtlLengthSid(SidAndAttributes[i].Sid) == ValidateArray[i].SidLength);
```
HBelusca, line 695: But shouldn't we do these checks live on release builds as well, and fail properly in case something bad happens?

ntoskrnl/se/sid.c
```diff
@@ -651,6 +746,12 @@ SeCaptureSidAndAttributesArray(
         *CapturedSidAndAttributes = NULL ;
```
HBelusca, line 746: Ah there's a space before the semicolon; could be removed.

ntoskrnl/se/sid.c
```diff
@@ -648,7 +733,13 @@ SeCaptureSidAndAttributesArray(
         }
 
         /* Set returned address to NULL */
-        *CapturedSidAndAttributes = NULL ;
+        *CapturedSidAndAttributes = NULL;
+    }
+
+    /* Free the temporary validation array */
+    if (PreviousMode != KernelMode)
```
ThFabba, line 740: Add null check, in case the allocation or anything before it raises an exception.

ntoskrnl/se/sid.c
```diff
@@ -558,7 +608,9 @@ SeCaptureSidAndAttributesArray(
                                                  TAG_SID_AND_ATTRIBUTES);
         if (SidAndAttributes == NULL)
         {
-            return STATUS_INSUFFICIENT_RESOURCES;
+            DPRINT1("SeCaptureSidAndAttributesArray(): Failed to allocate memory for SID and attributes array!\n");
```
ThFabba, line 611: Nit: printing the length of what failed to allocate is often helpful (i.e. was it a ridiculously large request vs are we out of memory)

ntoskrnl/se/sid.c
```diff
@@ -587,20 +641,50 @@ SeCaptureSidAndAttributesArray(
...
                 /* Copy the SID to the current destination address */
                 SidAndAttributes[i].Sid = (PSID)CurrentDest;
-                RtlCopyMemory(CurrentDest, SrcSidAndAttributes[i].Sid, SidLength);
+                RtlCopyMemory(CurrentDest, ValidateArray[i].ProbeSid, SidLength);
 
-                /* Sanity checks */
-                ASSERT(RtlLengthSid(SidAndAttributes[i].Sid) == SidLength);
-                ASSERT(RtlValidSid(SidAndAttributes[i].Sid));
+                /* Obtain the SID for validation */
+                Sid = SrcSidAndAttributes[i].Sid;
```
ThFabba, line 658: You want `SidAndAttributes` here, not Src -- we have to validate the data that we actually copied to kernel memory; since the UM memory may still change under us after the `RtlCopyMemory`

ntoskrnl/se/sid.c
```diff
@@ -637,6 +721,7 @@ SeCaptureSidAndAttributesArray(
         }
     }
 
+Cleanup:
     /* Check for failure */
     if (!NT_SUCCESS(Status))
     {
```
ThFabba, line 727: The line below `if (SidAndAttributes != AllocatedMem)` needs to check for NULL, which can be the case in the `STATUS_BUFFER_TOO_SMALL` code path.

PR #4099, opened 2021-11-11
[CRT] Do Not lock ioinfo when Spawning Functions

Doug-Lyons:
## Change locking ioinfo when spawning to not lock it

_Revert locking ioinfo when spawing to previous logic._

JIRA issue: [CORE-15176](https://jira.reactos.org/browse/CORE-15176)

## Change create_io_inherit_block in file.c to not lock ioinfo

_Stops GIMP Version 2.6.11 from freezing when opening a file._


JoachimHenze (approved):
The testbots were happy with that patch, and Thomas Faber also concluded this being a proper fix.
I will merge it within the next 24hours therefore.

PR #4101, opened 2021-11-12
[CRT] stdio/file.c: Import 3 Wine one-liner fixes

SergeGautherie:
A first/small step to more sync.


learn-more (approved):
changes lgtm, not sure about sync status.

PR #4104, opened 2021-11-14
[D3DWINE] wined3d_device_init_gdi(): Fix swapchain error handling

SergeGautherie:
More like `wined3d_device_init_3d()`.

Follow-up to dbd0569 (r36600).

NB:
Eventually, Wine merged `wined3d_device_init_gdi()` into `wined3d_device_init_3d()`. (Future WINESYNC...)

### DIFF
--- a/dll/directx/wine/wined3d/device.c
+++ b/dll/directx/wine/wined3d/device.c
@@ -1166,6 +1166,9 @@ HRESULT CDECL wined3d_device_init_gdi(struct wined3d_device *device,
     if (!(device->swapchains = heap_calloc(device->swapchain_count, sizeof(*device->swapchains))))
     {
         ERR("Out of memory!\n");
+#ifdef __REACTOS__
+        device->swapchain_count = 0;
+#endif
         goto err_out;
     }
     device->swapchains[0] = swapchain;
@@ -1181,7 +1184,14 @@ HRESULT CDECL wined3d_device_init_gdi(struct wined3d_device *device,
     return WINED3D_OK;
 
 err_out:
+#ifdef __REACTOS__
+    if (swapchain)
+    {
+#endif
     wined3d_swapchain_decref(swapchain);
+#ifdef __REACTOS__
+    }
+#endif
     return hr;
 }
 


JoachimHenze:
will this entirely fix CORE-17433 together with Thomas PR4103?
Have you tried that?

JoachimHenze:
gcc i386 complained about your changed code:
../src/dll/directx/wine/wined3d/device.c:1188:3: error: this 'if' clause does not guard... [-Werror=misleading-indentation]
   if (swapchain)
   ^~
../src/dll/directx/wine/wined3d/device.c:1191:5: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the 'if'
     return hr;
     ^~~~~~

JoachimHenze (changes requested):
Either change the indentation level of the return or add curly braces {} to wrap both commands.
Depends on what you had in mind.

dll/directx/wine/wined3d/device.c
```diff
@@ -1181,7 +1184,12 @@ HRESULT CDECL wined3d_device_init_gdi(struct wined3d_device *device,
     return WINED3D_OK;
 
 err_out:
+#ifndef __REACTOS__
     wined3d_swapchain_decref(swapchain);
+#else
+    if (swapchain)
+        wined3d_swapchain_decref(swapchain);
```
ThFabba, line 1191: You shouldn't need to duplicate this to fix the warning. Just have both lines at the same indent level, i.e.
```
#ifdef __REACTOS__
    if (swapchain)
#endif
    wined3d_swapchain_decref(swapchain);
```

SergeGautherie:
> will this entirely fix CORE-17433 together with Thomas PR4103? Have you tried that?

I can't test such "custom build" ticket. Sorry.

JoachimHenze:
Have you actually tried your patch vs CORE-17433 together with Thomas PR4103? Does it really/even help in any way?

dll/directx/wine/wined3d/device.c
SergeGautherie: Updated.

JoachimHenze:
I removed the link to that JIRA ticket now, because it has absolutely no relation to that.

JoachimHenze:
Since Thomas and my suggestions have been addressed now, I am neutral about committing that changed.
It is not really needed for anything actually, it only improves an academic corner case that we never saw on real life.
And has nothing to do with any JIRA ticket.
So others: close it or merge it, depending on your own preference.

PR #4105, opened 2021-11-14
[KERNEL32] Implement Data Execution Prevention (DEP) support

oleg-dubinskiy:
## Purpose

Implement support for Data Execution Prevention (DEP), which is represented by three functions: `GetProcessDEPPolicy`, `GetSystemDEPPolicy` and `SetProcessDEPPolicy` (my implementations are based on Wine's ones if what). They were introduced since Windows XP SP2.
These functions are required by some apps (like MiniBrowser 1.0), they only allow to get and set DEP policy settings of/for the operating system/current process.
This is only user mode part. To complete supporting this feature, in case of software emulation, `RtlIsValidHandler()` needs to be implemented in the kernel, for SafeSEH support: https://git.reactos.org/?p=reactos.git;a=blob;f=sdk/lib/rtl/i386/except.c;hb=c8b6abab80db2d2856dcedaa907f5b22eab3280e#l121.
But theoretically, if there is any emulated/real hardware that supports NX executable bit (required for DEP), and ROS amd64 port is able to boot there, it now might work as expected. :slightly_smiling_face: 
Also note that the appropriate tab in our System Properties (sysdm.cpl) is missing, but in Windows 2003, it is provided by ws03res.dll, so we're able to change the DEP settings in registry via MS sysdm.cpl + ws03res.dll.

JIRA issue: [CORE-16503](https://jira.reactos.org/browse/CORE-16503)

## Proposed changes

- Implement the functions itself in separate file (dll/win32/kernel32/client/dep.c).
- Add the Doxygen documentation for them.
- Include this file in CMakeLists.txt of kernel32.
- Add `DEP_SYSTEM_POLICY_TYPE` enumeration in sdk/include/psdk/winbase.h. It's required by `GetSystemDEPPolicy`.
- Include required ndk headers into internal dll/win32/kernel32/k32.h header.
- Fix exports (spec entries).

## Analysis

This is how these implementations are done in Windows:
dep

They same are getting/setting flags, calling `NtQuery`/`SetInformationProcess` for doing the actual work, and retrun an appropiate error code via `BaseSetLastNtError` in case of failure. So my implementations appear to be correct. :slightly_smiling_face: 
Regarding `GetSystemDEPPolicy`, I'm not completely sure that it's correct, but this way it's done in Wine at least: https://source.winehq.org/git/wine.git/blob/be0684dad50ffbc93b3ded4fbfebf1d1e4690589:/dlls/kernel32/process.c#l555, and in full WinDBG unassemble, global `SharedUserData` variable, which points to `_KUSER_SHARED_DATA` structure, is also visible.

## Result

After my changes, at least MiniBrowser 1.0 starts successfully, although works with a lot of errors. :slightly_frowning_face:

### DIFF
--- a/dll/win32/kernel32/client/dep.c
+++ b/dll/win32/kernel32/client/dep.c
@@ -1,0 +1,147 @@
+/*
+ * PROJECT:         ReactOS system libraries
+ * LICENSE:         GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * FILE:            dll/win32/kernel32/client/dep.c
+ * PURPOSE:         ReactOS Data Execution Prevention (DEP) functions
+ * COPYRIGHT:       Copyright 2021 Oleg Dubinskiy (oleg.dubinskij2013@yandex.ua)
+ */
+
+/* INCLUDES *******************************************************************/
+
+#include <k32.h>
+
+#define NDEBUG
+#include <debug.h>
+
+/* PUBLIC FUNCTIONS ***********************************************************/
+
+/**
+ * @brief Retrieves the data execution prevention (DEP) and thunk emulation (DEP-ATL)
+ * settings for the process.
+ * 
+ * On Windows XP SP3, the current process is handled.
+ * On Vista and newer, the process specified in hProcess parameter is handled.
+ * 
+ * @param [in] hProcess
+ * A handle to a process with PROCESS_QUERY_INFORMATION privilege.
+ * Ignored on Windows XP SP3.
+ * 
+ * @param [out] lpFlags
+ * Pointer to a variable that receives the flags:
+ * 0 - DEP is disabled.
+ * PROCESS_DEP_ENABLE - DEP is enabled.
+ * PROCESS_DEP_DISABLE_ATL_THUNK_EMULATION - DEP-ATL is disabled.
+ * 
+ * @param [out] lpPermanent
+ * Pointer to a variable that specifies whether DEP is enabled or disabled permanently.
+ * 
+ * @return
+ * TRUE on success, FALSE on failure.
+ * 
+ * @remarks
+ * Supported for 32-bit processes only. Returns ERROR_NOT_SUPPORTED for 64-bit processes.
+ * 
+ */
+BOOL
+WINAPI
+GetProcessDEPPolicy(_In_ HANDLE hProcess,
+                    _Out_ LPDWORD lpFlags,
+                    _Out_ PBOOL lpPermanent)
+{
+    ULONG Flags;
+    NTSTATUS Status;
+
+    Status = NtQueryInformationProcess(GetCurrentProcess(),
+                                       ProcessExecuteFlags,
+                                       &Flags,
+                                       sizeof(Flags),
+                                       NULL);
+    if (!NT_SUCCESS(Status))
+    {
+        BaseSetLastNTError(Status);
+        return FALSE;
+    }
+
+    if (lpFlags)
+    {
+        *lpFlags = 0;
+        if (Flags & MEM_EXECUTE_OPTION_DISABLE)
+            *lpFlags |= PROCESS_DEP_ENABLE;
+        if (Flags & MEM_EXECUTE_OPTION_DISABLE_THUNK_EMULATION)
+            *lpFlags |= PROCESS_DEP_DISABLE_ATL_THUNK_EMULATION;
+    }
+
+    if (lpPermanent)
+        *lpPermanent = (Flags & MEM_EXECUTE_OPTION_PERMANENT) != 0;
+
+    return TRUE;
+}
+
+/**
+ * @brief Sets the data execution prevention (DEP) and thunk emulation (DEP-ATL)
+ * settings for the current process.
+ * 
+ * @param [in] dwFlags
+ * Variable that specifies the flags to set:
+ * 0 - disables DEP;
+ * PROCESS_DEP_ENABLE - enables DEP permanently.
+ * PROCESS_DEP_DISABLE_ATL_THUNK_EMULATION - disables DEP-ATL. Can be specified
+ *                                           only with PROCESS_DEP_ENABLE.
+ * 
+ * @return
+ * TRUE on success, FALSE on failure.
+ * 
+ * @remarks
+ * Supported for 32-bit processes only. Returns ERROR_NOT_SUPPORTED for 64-bit processes.
+ * 
+ */
+BOOL
+WINAPI
+SetProcessDEPPolicy(_In_ DWORD dwFlags)
+{
+    ULONG Flags = 0;
+    NTSTATUS Status;
+
+    if (dwFlags & PROCESS_DEP_ENABLE)
+        Flags |= MEM_EXECUTE_OPTION_DISABLE | MEM_EXECUTE_OPTION_PERMANENT;
+    if (dwFlags & PROCESS_DEP_DISABLE_ATL_THUNK_EMULATION)
+        Flags |= MEM_EXECUTE_OPTION_DISABLE_THUNK_EMULATION;
+
+    Status = NtSetInformationProcess(GetCurrentProcess(),
+                                     ProcessExecuteFlags,
+                                     &Flags,
+                                     sizeof(Flags));
+    if (!NT_SUCCESS(Status))
+    {
+        BaseSetLastNTError(Status);
+        return FALSE;
+    }
+
+    return TRUE;
+}
+
+/**
+ * @brief Retrieves the system data execution prevention (DEP) and thunk emulation (DEP-ATL)
+ * settings.
+ * 
+ * @return
+ * One of DEP_SYSTEM_POLICY_TYPE enumeration values:
+ * AlwaysOff - DEP is always disabled system-wide.
+ * AlwaysOn - DEP is always enabled system-wide.
+ * OptIn - DEP is enabled only for system components.
+ *         Default for client Windows versions.
+ * OptOut - DEP is enabled for system components and processes.
+ *          Default for server Windows versions.
+ * 
+ * @remarks
+ * Supported for 32-bit processes only.
+ * 
+ */
+DEP_SYSTEM_POLICY_TYPE
+WINAPI
+GetSystemDEPPolicy(VOID)
+{
+    return SharedUserData->NXSupportPolicy;
+}
+
+/* EOF */
--- a/sdk/include/psdk/winbase.h
+++ b/sdk/include/psdk/winbase.h
@@ -608,10 +608,15 @@ extern "C" {
 
 #define INIT_ONCE_STATIC_INIT RTL_RUN_ONCE_INIT
 
-#if (_WIN32_WINNT >= 0x0600)
+typedef enum _DEP_SYSTEM_POLICY_TYPE {
+    AlwaysOff = 0,
+    AlwaysOn = 1,
+    OptIn = 2,
+    OptOut = 3
+} DEP_SYSTEM_POLICY_TYPE;
+
 #define PROCESS_DEP_ENABLE 0x00000001
 #define PROCESS_DEP_DISABLE_ATL_THUNK_EMULATION 0x00000002
-#endif
 
 #ifndef RC_INVOKED
 
--- a/dll/win32/kernel32/CMakeLists.txt
+++ b/dll/win32/kernel32/CMakeLists.txt
@@ -19,6 +19,7 @@ list(APPEND SOURCE
     client/atom.c
     client/compname.c
     client/debugger.c
+    client/dep.c
     client/dosdev.c
     client/dllmain.c
     client/environ.c
--- a/dll/win32/kernel32/k32.h
+++ b/dll/win32/kernel32/k32.h
@@ -38,10 +38,13 @@
 #include <ndk/iotypes.h>
 #include <ndk/kdtypes.h>
 #include <ndk/kefuncs.h>
+#include <ndk/ketypes.h>
 #include <ndk/ldrfuncs.h>
 #include <ndk/mmfuncs.h>
+#include <ndk/mmtypes.h>
 #include <ndk/obfuncs.h>
 #include <ndk/psfuncs.h>
+#include <ndk/pstypes.h>
 #include <ndk/rtlfuncs.h>
 #include <ndk/setypes.h>
 #include <ndk/umfuncs.h>
--- a/dll/win32/kernel32/kernel32.spec
+++ b/dll/win32/kernel32/kernel32.spec
@@ -557,7 +557,7 @@
 @ stdcall GetPrivateProfileStructW(wstr wstr ptr long wstr)
 @ stdcall GetProcAddress(long str)
 @ stdcall GetProcessAffinityMask(long ptr ptr)
-@ stub -version=0x600+ GetProcessDEPPolicy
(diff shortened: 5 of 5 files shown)


sdk/include/psdk/winbase.h
```diff
@@ -608,10 +608,15 @@ extern "C" {
 
 #define INIT_ONCE_STATIC_INIT RTL_RUN_ONCE_INIT
 
-#if (_WIN32_WINNT >= 0x0600)
```
SergeGautherie, line 611: Would we want something like XPsp3, instead? (Here and maybe elsewhere.)

dll/win32/kernel32/client/dep.c
```diff
@@ -0,0 +1,147 @@
...
+ * PROCESS_DEP_DISABLE_ATL_THUNK_EMULATION - DEP-ATL is disabled.
+ * 
+ * @param [out] lpPermanent
+ * Pointer to a variable that specifies whether DEP is enabled or disabled permanently.
+ * 
+ * @return
+ * TRUE on success, FALSE on failure.
+ * 
+ * @remarks
+ * Supported for 32-bit processes only. Returns ERROR_NOT_SUPPORTED for 64-bit processes.
```
SergeGautherie, line 42: The 64bit behavior looks missing for the 3 functions, isn't it?

dll/win32/kernel32/client/dep.c
```diff
@@ -0,0 +1,147 @@
...
+                                       &Flags,
+                                       sizeof(Flags),
+                                       NULL);
+    if (!NT_SUCCESS(Status))
+    {
+        BaseSetLastNTError(Status);
+        return FALSE;
+    }
+
+    if (lpFlags)
```
SergeGautherie, line 65: The 2 parameters are not optional: why check them?

dll/win32/kernel32/client/dep.c
```diff
@@ -0,0 +1,147 @@
...
+ * @return
+ * TRUE on success, FALSE on failure.
+ * 
+ * @remarks
+ * Supported for 32-bit processes only. Returns ERROR_NOT_SUPPORTED for 64-bit processes.
+ * 
+ */
+BOOL
+WINAPI
+GetProcessDEPPolicy(_In_ HANDLE hProcess,
```
SergeGautherie, line 47: Add an explicit `UNREFERENCED_PARAMETER()` call?

dll/win32/kernel32/client/dep.c
```diff
@@ -0,0 +1,147 @@
...
+ * @return
+ * TRUE on success, FALSE on failure.
+ * 
+ * @remarks
+ * Supported for 32-bit processes only. Returns ERROR_NOT_SUPPORTED for 64-bit processes.
+ * 
+ */
+BOOL
+WINAPI
+SetProcessDEPPolicy(_In_ DWORD dwFlags)
```
SergeGautherie, line 100: This (Wine) implementation seems to ignore cases like 'If the system DEP policy is AlwaysOff or AlwaysOn, SetProcessDEPPolicy returns an error.'.
A TODO comment somewhere referencing Wine and listing missing cases would be good...

SergeGautherie:
Does `modules/rostests/winetests/kernel32/virtual.c` result change?

You might want to import (additional test part)
https://source.winehq.org/git/wine.git/commit/e3fe1b09468efc9d6faf8ec05a91e3972458b293
too?

dll/win32/kernel32/client/dep.c
oleg-dubinskiy: Yes, seems to be. But regarding `GetSystemDEPPolicy`, MSDN says the following:
> If the system DEP policy is OptIn or OptOut, DEP can be selectively enabled or disabled for the current process by calling the SetProcessDEPPolicy function. This function works only for 32-bit processes.

It's not completely understandable, which function is meant in this context: `GetProcessDEPPolicy` or `GetSystemDEPPolicy`? Or both? :thinking:

dll/win32/kernel32/client/dep.c
oleg-dubinskiy: Because it was done in Wine at least, but looks like it's indeed not needed to check for them, since they are not optional at all.

sdk/include/psdk/winbase.h
```diff
@@ -608,10 +608,15 @@ extern "C" {
...
+typedef enum _DEP_SYSTEM_POLICY_TYPE {
+    AlwaysOff = 0,
+    AlwaysOn = 1,
+    OptIn = 2,
+    OptOut = 3
+} DEP_SYSTEM_POLICY_TYPE;
+
 #define PROCESS_DEP_ENABLE 0x00000001
```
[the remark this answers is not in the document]
oleg-dubinskiy: Then I also need to add the Vista ifdef around the functions? In such case, I also will need to set exports to `-version=0x600+`, otherwise it will fail to compile. But in such case, it makes no sense, since the functions will be only compiled for NT6+, when in Windows they are present since XP SP3. So this will be wrong.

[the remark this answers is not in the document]
learn-more: The functions seem present on XP SP3, but not on server 2003 SP2.

dll/win32/kernel32/client/dep.c
```diff
@@ -0,0 +1,147 @@
...
+BOOL
+WINAPI
+GetProcessDEPPolicy(_In_ HANDLE hProcess,
+                    _Out_ LPDWORD lpFlags,
+                    _Out_ PBOOL lpPermanent)
+{
+    ULONG Flags;
+    NTSTATUS Status;
+
+    Status = NtQueryInformationProcess(GetCurrentProcess(),
```
learn-more, line 54: Why not use hProcess here?

sdk/include/psdk/winbase.h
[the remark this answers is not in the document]
oleg-dubinskiy: So I'd better add XPSP3 ifdef instead.

dll/win32/kernel32/client/dep.c
oleg-dubinskiy: @learn-more see MSDN documentation:
> Gets the data execution prevention (DEP) and DEP-ATL thunk emulation settings for the specified 32-bit process.Windows XP with SP3:  Gets the DEP and DEP-ATL thunk emulation settings for the current process.

> [in] hProcess
A handle to the process. PROCESS_QUERY_INFORMATION privilege is required to get the DEP policy of a process.
Windows XP with SP3:  The hProcess parameter is ignored.

https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-getprocessdeppolicy

sdk/include/psdk/winbase.h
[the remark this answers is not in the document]
oleg-dubinskiy: Ah no, it seems to be present in 2003 SP2 also, at least because:
1. MSDN says the following:
>Data Execution Prevention (DEP) is a system-level memory protection feature that is built into the operating system starting with Windows XP **and Windows Server 2003**.

on this page: https://docs.microsoft.com/en-us/windows/win32/memory/data-execution-prevention.
2. In 2003 SP2, the DEP tab in sysdm.cpl also exists, so DEP appears to be included there too.

dll/win32/kernel32/client/dep.c
GeoB99: In this context the MSDN documentation states that setting DEP onto the current calling process with `SetProcessDEPPolicy` only works if the current calling process itself is a 32-bit process. Though indeed the documentation doesn't explicitly say how the function actually doesn't work so I'd suggest that you write an APITEST to testify the behavior.

dll/win32/kernel32/client/dep.c
GeoB99: Since these KERNEL32 API calls are public and exported, it's concise that you write some APITESTS so that you testify the behavior of these functions (in XP and in Server 2003 as well).

GeoB99 (commented):
Since these KERNEL32 API calls are public and exported, it's concise that you write some APITESTS so that you testify the behavior of these functions (in XP and in Server 2003 as well).

dll/win32/kernel32/client/dep.c
oleg-dubinskiy: Btw, we already seem to have an appropriate winetest for that: https://git.reactos.org/?p=reactos.git;a=blob;f=modules/rostests/winetests/kernel32/virtual.c;hb=f028ca5b083524cc85fd75467b6f8db4e37f0164.

GeoB99: That winetest barely tests anything about DEP implementation of Windows. So do what I've told you to do, implement a testcase on your own.

sdk/include/psdk/winbase.h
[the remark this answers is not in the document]
SergeGautherie: > These functions are required by some apps

Does MiniBrowser 1.0 work on S03sp2?

> The functions seem present on XP SP3, but not on server 2003 SP2.

Indeed!
kernel32.dll, v5.1.2600.5781: 3 names.
kernel32.dll, v5.2.3790.4480: 0 name. (Are these features available differently? Assumed not.)

Then, we may want to add them to kernel32_vista.dll instead...

> https://docs.microsoft.com/en-us/windows/win32/memory/data-execution-prevention

Yeah, meanwhile these 3 function pages do read:
```
Minimum supported client | Windows Vista with SP1, Windows XP with SP3 [desktop apps only]
Minimum supported server | Windows Server 2008 [desktop apps only]
```

dll/win32/kernel32/client/dep.c
learn-more: Indeed, msdn is not always correct ;)

sdk/include/psdk/winbase.h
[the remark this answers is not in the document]
oleg-dubinskiy: Sorry, my mistake. These APIs are indeed missing on Windows Server 2003 SP2. :sweat_smile: They exist only on XP SP3. So we indeed should separate this stuff via ifdef, but via XP SP3 ifdef still, because otherwise these APIs will be unavailable for the apps developed for XP, so they will fail.
dep-w2k3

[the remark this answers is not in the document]
SergeGautherie: > XP SP3 ifdef

MSDN reads`To compile an application that calls this function, define _WIN32_WINNT as 0x0600 or later.`...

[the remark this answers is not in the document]
HBelusca: https://github.com/adobe/chromium/blob/master/sandbox/src/dep.cc#L46
claims:
>   // Only available on Windows XP SP2 and **Windows Server 2003 SP1**.
  // For reference: http://www.uninformed.org/?v=2&a=4

🤔

dll/win32/kernel32/client/dep.c
```diff
@@ -0,0 +1,147 @@
...
+ * settings.
+ * 
+ * @return
+ * One of DEP_SYSTEM_POLICY_TYPE enumeration values:
+ * AlwaysOff - DEP is always disabled system-wide.
+ * AlwaysOn - DEP is always enabled system-wide.
+ * OptIn - DEP is enabled only for system components.
+ *         Default for client Windows versions.
+ * OptOut - DEP is enabled for system components and processes.
+ *          Default for server Windows versions.
```
HBelusca, line 134:
```suggestion
 * AlwaysOff - DEP is always disabled system-wide.
 * AlwaysOn  - DEP is always enabled system-wide.
 * OptIn     - DEP is enabled only for system components.
 *             Default for client Windows versions.
 * OptOut    - DEP is enabled for system components and processes.
 *             Default for server Windows versions.
```

dll/win32/kernel32/client/dep.c
```diff
@@ -0,0 +1,147 @@
...
+ * 
+ * @remarks
+ * Supported for 32-bit processes only.
+ *
```
HBelusca, line 138:
```suggestion
```

dll/win32/kernel32/client/dep.c
```diff
@@ -0,0 +1,147 @@
...
+ * @remarks
+ * Supported for 32-bit processes only. Returns ERROR_NOT_SUPPORTED for 64-bit processes.
+ * 
+ */
+BOOL
+WINAPI
+SetProcessDEPPolicy(_In_ DWORD dwFlags)
+{
+    ULONG Flags = 0;
+    NTSTATUS Status;
```
HBelusca, line 103:
```suggestion
    NTSTATUS Status;
    ULONG Flags = 0;
```

dll/win32/kernel32/client/dep.c
HBelusca: Also, "Windows" --> "NT" or "ReactOS"? 🤔

sdk/include/psdk/winbase.h
```diff
@@ -608,10 +608,15 @@ extern "C" {
 
 #define INIT_ONCE_STATIC_INIT RTL_RUN_ONCE_INIT
 
-#if (_WIN32_WINNT >= 0x0600)
+typedef enum _DEP_SYSTEM_POLICY_TYPE {
+    AlwaysOff = 0,
+    AlwaysOn = 1,
+    OptIn = 2,
+    OptOut = 3
+} DEP_SYSTEM_POLICY_TYPE;
```
HBelusca, line 616:
```suggestion
typedef enum _DEP_SYSTEM_POLICY_TYPE {
    DEPPolicyAlwaysOff = 0,
    DEPPolicyAlwaysOn,
    DEPPolicyOptIn,
    DEPPolicyOptOut,
    DEPTotalPolicyCount
} DEP_SYSTEM_POLICY_TYPE;
```

sdk/include/psdk/winbase.h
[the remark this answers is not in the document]
ThFabba: This comment is in a PSDK header, so it doesn't really matter what's exported where. Microsoft's header has the ifdef, so ours should, too. If that causes trouble for our kernel32, then it needs to set `_WIN32_WINNT` to Vista.

dll/win32/kernel32/client/dep.c
ThFabba: My bike-shed is gray.

dll/win32/kernel32/client/dep.c
```diff
@@ -0,0 +1,147 @@
...
+
+#define NDEBUG
+#include <debug.h>
+
+/* PUBLIC FUNCTIONS ***********************************************************/
+
+/**
+ * @brief Retrieves the data execution prevention (DEP) and thunk emulation (DEP-ATL)
+ * settings for the process.
+ *
```
HBelusca, line 21: (Here and anywhere else in this file): Please remove the trailing whitespaces.

HBelusca:
@oleg-dubinskiy What's the status of this PR?

oleg-dubinskiy:
> @oleg-dubinskiy What's the status of this PR?

Hi @HBelusca. As I remember, it was nearly almost done (only some suggestions need to be applied), but some blocking things also are **it is done like in Windows XP SP3, in Windows Server 2003 SP2 it's done in a bit different way, there are no such a functions in 2k3's kernel32.dll.** Also, writing a Rostest needs to spend more time on it too.
But if this is more actual now, I can (try to) finish it. :slightly_smiling_face:

HBelusca:
A naive question (maybe @DarkFire01 or others could tell?), is how is it done in Vista, at the level of kernel32?

PR #4107, opened 2021-11-15
Set target Windows version in config.cmake

wjk:
## Purpose

Even though ReactOS currently targets Windows XP/Server 2003 only, forks (like mine) might want to build the code that targets Vista+ on a regular basis. This PR makes that change easier to perform without affecting the default build.

## Proposed changes

- Add two new cache variables: `TARGET_WINVER` AND `TARGET_WINVER_IE`. These are the values of `_WIN32_WINNT` and `_WIN32_IE`, respectively. I must have two because they are not the same on our current configuration. They default to their current values as specified in the top-level CMakeLists.txt file.
- Pass a default `--version=` argument in spec2def calls that don’t specify an explicit `VERSION` argument. This is required so that if someone enables the newer APIs, the implibs will expose the correct symbols.


DarkFire01:
I have been requested!

This is pretty cool I actually use a similar change to this on some of my own local NT6+ branches :) as I'm sure you know to actually make use of this there's a decently long list of corrections the NT6+ definitions that need to be addressed.
Anyway here's my notes:
ntverp.h and freeldr.h need to have similar changes as well If you're going to do this! I'm on mobile at the minute but if the PR is still open in t-minus 9 hours I'll make specific changes requests showing you. That is unless you know what I mean then go for it :)

HBelusca:
What about `NTDDI_VERSION` and the like?

wjk:
> What about `NTDDI_VERSION` and the like?

[I believe I already do that.](https://github.com/reactos/reactos/pull/4107/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR236-R239) Would you like me to add to this list?

HBelusca:
> > What about `NTDDI_VERSION` and the like?
> 
> I believe I already do that.

Dunno, I don't see `NTDDI_VERSION` anywhere in the patch.
Also, there would certainly be more adjustments needed for consistency elsewhere in ROS' code: First, in the files @DarkFire01 mentioned, then in those files listed there as well: https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=define+NTDDI_VERSION

wjk:
@HBelusca Apologies for the misunderstanding; I was trying to state that I have the infrastructure ready to patch NTDDI_VERSION et al; I just need to add the appropriate variables to the list.

sdk/cmake/config.cmake
```diff
@@ -119,3 +119,40 @@ endif()
 
 set(USE_DUMMY_PSEH FALSE CACHE BOOL
 "Whether to disable PSEH support.")
+
+set(TARGET_VERSION XP CACHE STRING "The Windows platform version to target. Valid values are win10 win8.1 win8 win7 vista xp. Defaults to xp.")
```
Saibamen, line 123: Maybe commas between values?
```suggestion
set(TARGET_VERSION XP CACHE STRING "The Windows platform version to target. Valid values are win10, win8.1, win8, win7, vista, xp. Defaults to xp.")
```

sdk/cmake/config.cmake
```diff
@@ -1,5 +1,5 @@
 
-if(ARCH STREQUAL "i386")
+	if(ARCH STREQUAL "i386")
```
Saibamen, line 2: Unexpected diff?

wjk: Fixed. Will arrive next force-push. (I’m trying to keep my incremental changes from wasting time in CI.)

wjk:
This PR is ready for review now. It builds fully on my machine, but I want to wait to see what the CI says before testing it. Thanks!

sdk/cmake/CMakeMacros.cmake
```diff
@@ -577,8 +577,58 @@ list(APPEND VALID_MODULE_TYPES kernel kerneldll kernelmodedriver wdmdriver nativ
 list(APPEND KERNEL_MODULE_TYPES kernel kerneldll kernelmodedriver wdmdriver)
 list(APPEND NATIVE_MODULE_TYPES kernel kerneldll kernelmodedriver wdmdriver nativecui nativedll)
 
+function(add_ntddi_defines MODULE TARGET_VERSION)
```
Extravert-ir, line 580: Please, change the function name to `set_ntddi_defines` to follow other functions' naming convention

sdk/cmake/config.cmake
Extravert-ir, line 2: Spaces were replaces with a tab

sdk/cmake/CMakeMacros.cmake
```diff
@@ -683,6 +733,15 @@ function(set_module_type MODULE TYPE)
         set_target_properties(${MODULE} PROPERTIES SUFFIX ".cpl")
     endif()
 
+    # Set the Windows version being compiled for
+    if(TARGET_VERSION)
```
Extravert-ir, line 737: Let's do it in a little bit different way:
- `TARGET_VERSION` should always be set
- The default value is "win2003"
- If some module wants to override `TARGET_VERSION`, it should be greater than the default one, otherwise default one is used. (you may use `list(FIND ...)` cmake feature for comparing "greatness")

sdk/include/psdk/CMakeLists.txt
```diff
@@ -160,3 +160,34 @@ add_iid_library(wuguid wuapi.idl)
...
+    set(NTVERP_VERSION_W 0x0602)
+elseif(TARGET_VERSION STREQUAL "win7")
+    set(NTVERP_MAJOR_VERSION 6)
+    set(NTVERP_MINOR_VERSION 1)
+    set(NTVERP_VERSION_W 0x0601)
+elseif(TARGET_VERSION STREQUAL "vista")
+    set(NTVERP_MAJOR_VERSION 6)
+    set(NTVERP_MINOR_VERSION 0)
+    set(NTVERP_VERSION_W 0x0600)
+elseif(TARGET_VERSION STREQUAL "xp" OR TARGET_VERSION STREQUAL "")
```
Extravert-ir, line 185: xp -> win2003

sdk/lib/drivers/wdf/CMakeLists.txt
```diff
@@ -186,9 +184,13 @@ list(APPEND SOURCE_KM
 # add_registry_inf(wdf01000_reg.inf)
 
 add_library(wdf01000 ${SOURCE_SHARED} ${SOURCE_KM})
+add_ntddi_defines(wdf01000 win8.1 PUBLIC)
+
+# Any component linking against wdf01000 must also target win8.1.
+add_ntddi_defines(wdf01000 win8.1)
```
Extravert-ir, line 190: Why calling the function two times?

Extravert-ir (changes requested):
I like the idea in general, but please address the suggestions

sdk/lib/drivers/wdf/CMakeLists.txt
wjk: A mistake. Since removed.

wjk: Appears to have been a mistake. Removed.

sdk/cmake/CMakeMacros.cmake
wjk: I believe you may be misunderstanding this functionality. An explicitly specified target version for the entire OS (using `-DTARGET_VERSION=` parameter in configure.sh/cmd) is checked first. If that is not set, then the module’s manually specified target version is used, if any. Otherwise, it defaults to `win2003`.

Extravert-ir: Hm, yeah.
I wanted to say, the default value should be set in config.cmake. That way, you don't need to check for it in set_module_type

Extravert-ir:
Ah, I would also say maybe it makes sense to change the name to `TARGET_WIN_VERSION`

PR #4108, opened 2021-11-15
Implement SHUpdateRecycleBinIcon

wjk:
aka make the Recycle Bin change its icon automatically when items are added to or removed from it. Enjoy!

### DIFF
--- a/dll/win32/shell32/folders/CRecycleBin.cpp
+++ b/dll/win32/shell32/folders/CRecycleBin.cpp
@@ -983,7 +983,7 @@ TRASH_TrashFile(LPCWSTR wszPath)
 }
 
 /*************************************************************************
- * SHUpdateCRecycleBinIcon                                [SHELL32.@]
+ * SHUpdateRecycleBinIcon                                [SHELL32.@]
  *
  * Undocumented
  */
--- a/dll/win32/shell32/folders/CRecycleBin.cpp
+++ b/dll/win32/shell32/folders/CRecycleBin.cpp
@@ -989,8 +989,22 @@ TRASH_TrashFile(LPCWSTR wszPath)
  */
 EXTERN_C HRESULT WINAPI SHUpdateRecycleBinIcon(void)
 {
-    FIXME("stub\n");
+    /*
+     * This function invokes SHChangeNotify() in order to have the shell
+     * requery the recycle bin shell object for its icon. This then triggers
+     * a check in CGuidItemExtractIcon_CreateInstance, which chooses between
+     * the two icons as necessary.
+     */
 
+    CComHeapPtr<ITEMIDLIST_ABSOLUTE> pidlRecycleBin;
+    HRESULT hr = SHGetFolderLocation(NULL, CSIDL_BITBUCKET, NULL, 0, &pidlRecycleBin);
+    if (FAILED(hr))
+        return hr;
+
+    // m_pData is referenced directly to ensure that the pointer to
+    // the pidl is given to SHChangeNotify(), not a pointer to the
+    // C++ object that wraps it.
+    SHChangeNotify(SHCNE_UPDATEITEM, SHCNF_IDLIST | SHCNF_FLUSH, pidlRecycleBin.m_pData, NULL);
     return S_OK;
 }
 
--- a/dll/win32/shell32/shellrecyclebin/recyclebin.c
+++ b/dll/win32/shell32/shellrecyclebin/recyclebin.c
@@ -7,6 +7,7 @@
  */
 
 #include "recyclebin_private.h"
+#include <undocshell.h>
 
 BOOL WINAPI
 CloseRecycleBinHandle(
--- a/dll/win32/shell32/shellrecyclebin/recyclebin.c
+++ b/dll/win32/shell32/shellrecyclebin/recyclebin.c
@@ -3,7 +3,7 @@
  * LICENSE:     GPL v2 - See COPYING in the top level directory
  * FILE:        lib/recyclebin/recyclebin.c
  * PURPOSE:     Public interface
- * PROGRAMMERS: Copyright 2006-2007 Hervé Poussineau (hpoussin@reactos.org)
+ * PROGRAMMERS: Copyright 2006-2007 HervÃ© Poussineau (hpoussin@reactos.org)
  */
 
 #include "recyclebin_private.h"
@@ -79,6 +79,7 @@ DeleteFileToRecycleBinW(
 
     hr = IRecycleBin_DeleteFile(prb, FileName);
     IRecycleBin_Release(prb);
+    SHUpdateRecycleBinIcon();
 
 cleanup:
     if (SUCCEEDED(hr))
@@ -100,6 +101,7 @@ DeleteFileHandleToRecycleBin(
     TRACE("(%p)\n", hDeletedFile);
 
     hr = IRecycleBinFile_Delete(rbf);
+    SHUpdateRecycleBinIcon();
 
     if (SUCCEEDED(hr))
         return TRUE;
@@ -157,6 +159,7 @@ EmptyRecycleBinW(
 
     hr = IRecycleBin_EmptyRecycleBin(prb);
     IRecycleBin_Release(prb);
+    SHUpdateRecycleBinIcon();
 
 cleanup:
     if (SUCCEEDED(hr))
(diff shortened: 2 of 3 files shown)


dll/win32/shell32/shellrecyclebin/recyclebin.c
```diff
@@ -3,7 +3,7 @@
  * LICENSE:     GPL v2 - See COPYING in the top level directory
  * FILE:        lib/recyclebin/recyclebin.c
  * PURPOSE:     Public interface
- * PROGRAMMERS: Copyright 2006-2007 Herv� Poussineau (hpoussin@reactos.org)
+ * PROGRAMMERS: Copyright 2006-2007 Herv� Poussineau (hpoussin@reactos.org)
```
HBelusca, line 6: 👀

dll/win32/shell32/shellrecyclebin/recyclebin.c
```diff
@@ -63,6 +63,8 @@ DeleteFileToRecycleBinA(
     return ret;
 }
 
+extern HRESULT WINAPI SHUpdateRecycleBinIcon(void);
```
HBelusca, line 66: Should be moved to the undocshell.h header.

dll/win32/shell32/folders/CRecycleBin.cpp
```diff
@@ -983,14 +983,21 @@ TRASH_TrashFile(LPCWSTR wszPath)
...
-    FIXME("stub\n");
+    CComHeapPtr<ITEMIDLIST_ABSOLUTE> pidlRecycleBin;
+    HRESULT hr = SHGetFolderLocation(NULL, CSIDL_BITBUCKET, NULL, 0, &pidlRecycleBin);
+    if (FAILED(hr))
+        return hr;
 
+    // m_pData is referenced directly to ensure that the pointer to
+    // the pidl is given to SHChangeNotify(), not a pointer to the
+    // C++ object that wraps it.
+    SHChangeNotify(SHCNE_UPDATEITEM, SHCNF_IDLIST | SHCNF_FLUSH, pidlRecycleBin.m_pData, NULL);
```
HBelusca, line 1000: Can you point out which function(s) are going to actually change the recycler icon? (here it's not obviously clear since you're sending a shell notification.) -- Also, are you sure that on Windows such notification is being sent when `SHUpdateRecycleBinIcon()` is called?

wjk: ✅

wjk:
> Are you sure that on Windows such notification is being sent when `SHUpdateRecycleBinIcon()` is called?

I cannot be certain, because I do not have a copy of Windows XP or one of its variants handy to write an apitest for (and I am not am an expert at reverse engineering Windows, so I would have to first figure out how to trace the flow of SHChangeNotify calls). This is the most straightforward approach that I could create that satisfies its API contract; why make it more complicated than it has to be? Respectfully, I believe that this PR should be evaluated according to its functionality, not its ability to copy Windows’s behavior with 100% accuracy. Thanks!

HBelusca:
> > Are you sure that on Windows such notification is being sent when `SHUpdateRecycleBinIcon()` is called?
> 
> I cannot be certain, because I do not have a copy of Windows XP or one of its variants handy to write an apitest for (and I am not am an expert at reverse engineering Windows, so I would have to first figure out how to trace the flow of SHChangeNotify calls). This is the most straightforward approach that I could create that satisfies its API contract; why make it more complicated than it has to be? Respectfully, I believe that this PR should be evaluated according to its functionality, not its ability to copy Windows’s behavior with 100% accuracy. Thanks!

I think this could be checked with ApiMonitor (by Rohitab) to see when `SHChangeNotify` is called (since this is a shell32 export it can be traced this way).
Regarding the other question I had, it's in order to be sure you understand what the thing is going on (what makes the `SHChangeNotify` ultimately changing the recycler's icon) and whether or not there could be unwanted side effects when calling `SHChangeNotify` this way.... so that we can be sure the patch is actually correct and not a hack. (Because I myself doesn't know that.)

dll/win32/shell32/folders/CRecycleBin.cpp
```diff
@@ -983,14 +983,28 @@ TRASH_TrashFile(LPCWSTR wszPath)
 }
 
 /*************************************************************************
- * SHUpdateCRecycleBinIcon                                [SHELL32.@]
+ * SHUpdateRecycleBinIcon                                [SHELL32.@]
  *
  * Undocumented
  */
 EXTERN_C HRESULT WINAPI SHUpdateRecycleBinIcon(void)
```
GeoB99, line 990: `SHUpdateRecycleBinIcon`'s logic of updating the trash bin icon is like so, in the following order:

- **Determine if the trash bin is empty or not**. It first iterates all the drive partitions and check each of them that one has the location path of the bucket that is associated as the default bucket bin of the system. Afterwards it loops the bucket of potential remaining files using the traditional fashion of `FindFirstFileW ... FindNextFileW` API calls. If the file in question is a directory (pointed by the **FILE_ATTRIBUTE_DIRECTORY** attribute) then it iterates over that directory as well. If at least one or more files have been found (empty directories are taken into account as well) then the bin is **NOT** empty.
- **Query/Set the Registry SHELL32 data**. Presumably it queries some information in regard of the trash bin icons and sets the default value as either "Empty" or "Full" depending on the state of the bin, I haven't looked much deeper into that.
- **Expand the environment strings with `SHExpandEnvironmentStringsW`**.
- **Update the icon image with `SHUpdateImageW`**.
- **Invoke a notification with `SHChangeNotify`**.

Now in the grand scheme of things I'm not sure if invoking a notification straight away alone in our case would be enough so ultimately the decision is up to @yagoulas.

wjk:
I can definitely rewrite SHUpdateRecycleBin to follow @GeoB99’s spec. I based the old design upon what I found already present in CRegFolder.cpp; turns out it dates [all the way back to r72208](https://github.com/reactos/reactos/commit/43728d3f666ed2bbf94a608662c8f16354e35654). Therefore, this PR extends what is most definitely a hack. I can definitely rewrite the Recycle Bin code to follow the correct design posted above. I will need to find out where in the Registry the flag is stored, but that is something I can easily accomplish with ProcMon. I can start work on this immediately. Thank you very much for showing me the correct implementation.

wjk:
> Presumably it queries some information in regard of the trash bin icons and sets the default value as either "Empty" or "Full" depending on the state of the bin

I have looked into Windows 10’s handling of this operation. Here is what I found:

1. explorer.exe opens the following key: `HKCU\SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\CLSID\{645FF040-5081-101B-9F08-00AA002F954E}\DefaultIcon`.
2. If the trash can is full, set the default value of this key to the contents of the `full` value.
3. If the trash can is empty, set the default value of this key to the contents of the `empty` value.

Again, I don’t have a copy of Windows XP to test with, but I’m willing to bet that the behavior would be the same. Since ReactOS uses this key in CRegFolder.cpp, I will start work on porting the above algorithm. Thanks!

learn-more:
> > Presumably it queries some information in regard of the trash bin icons and sets the default value as either "Empty" or "Full" depending on the state of the bin
> 
> I have looked into Windows 10’s handling of this operation. Here is what I found:
> 
> 1. explorer.exe opens the following key: `HKCU\SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\CLSID\{645FF040-5081-101B-9F08-00AA002F954E}\DefaultIcon`.
> 2. If the trash can is full, set the default value of this key to the contents of the `full` value.
> 3. If the trash can is empty, set the default value of this key to the contents of the `empty` value.
> 
> Again, I don’t have a copy of Windows XP to test with, but I’m willing to bet that the behavior would be the same. Since ReactOS uses this key in CRegFolder.cpp, I will start work on porting the above algorithm. Thanks!

That GUID is {{CLSID_RecycleBin}}

wjk:
> That GUID is CLSID_RecycleBin

Yes, I know. 🙂

dll/win32/shell32/folders/CRecycleBin.cpp
wjk: > empty directories are taken into account as well

How? Do only empty directories signal a full bin or an empty one?

wjk:
Superseded by #4221.

wjk:
> Superseded by https://github.com/reactos/reactos/pull/4221.

Actually, that PR is unrelated. Reopening. Sorry for the ping spam!

wjk:
Due to a very unfortunate error regarding my multiple forks of ReactOS, I will have to close this PR and reopen it from a different branch. Sorry about that!

PR #4109, opened 2021-11-15
[SYSSETUP] Register type libraries before signaling we are done

learn-more:
This prevents a race condition, where hSysSetupInf could be closed before all type libraries are registered

### DIFF
--- a/dll/win32/syssetup/wizard.c
+++ b/dll/win32/syssetup/wizard.c
@@ -2041,6 +2041,12 @@ RegistrationProc(LPVOID Parameter)
     RegistrationNotify.Progress = RegistrationData->DllCount;
     RegistrationNotify.ActivityID = IDS_REGISTERING_COMPONENTS;
     RegistrationNotify.CurrentItem = NULL;
+
+    RegisterTypeLibraries(hSysSetupInf, L"TypeLibraries");
+
+    // FIXME: Move this call to a separate cleanup page!
+    RtlCreateBootStatusDataFile();
+
     SendMessage(RegistrationData->hwndDlg, PM_REGISTRATION_NOTIFY,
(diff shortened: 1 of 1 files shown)


HBelusca:
What other thread causes the race condition?

learn-more:
The main thread.

dll/win32/syssetup/wizard.c
```diff
@@ -2041,6 +2041,12 @@ RegistrationProc(LPVOID Parameter)
     RegistrationNotify.Progress = RegistrationData->DllCount;
     RegistrationNotify.ActivityID = IDS_REGISTERING_COMPONENTS;
     RegistrationNotify.CurrentItem = NULL;
+
+    RegisterTypeLibraries(hSysSetupInf, L"TypeLibraries");
+
+    // FIXME: Move this call to a separate cleanup page!
+    RtlCreateBootStatusDataFile();
```
HBelusca, line 2048: Looks like this function call could stay where it originally was 🤔

JoachimHenze:
Can you paste a bit of log "how it looks when that race condition was hit"? For later reference. Thank you!

dll/win32/syssetup/wizard.c
learn-more: it is also something that should probably prevent the process from ending.

HBelusca: in what sense? it just creates a bootstat.dat in C:\ReactOS; this should not prevent anything else? 🤔

learn-more:
> Can you paste a bit of log "how it looks when that race condition was hit"? For later reference. Thank you!
[log.txt](https://github.com/reactos/reactos/files/7547128/log.txt)

HBelusca (approved):

PR #4111, opened 2021-11-15
[NTDLL:LDR] Don't hardcode process-execute flag values.

HBelusca:
## Purpose

Don't hardcode process-execute flag values.
See also PR #4105.


dll/ntdll/ldr/ldrinit.c
```diff
@@ -16,6 +16,12 @@
 #define NDEBUG
 #include <debug.h>
 
+/* Only available in ntddk.h; used for SharedUserData->NXSupportPolicy */
+#ifndef NX_SUPPORT_POLICY_ALWAYSOFF
+#define NX_SUPPORT_POLICY_ALWAYSOFF 0
```
ThFabba, line 21: https://github.com/reactos/reactos/pull/4105/files#diff-66cbd922213de63b9781afb3ffc36b6dd0804b5fd2a772983212d046b4f63d67R140 seems to indicate that these directly translate to the Win32 type `DEP_SYSTEM_POLICY_TYPE`. Perhaps it would make sense to use that?
Alternatively, these seem like good candidates for NDK if they're only defined in KM headers but are needed in UM.

HBelusca: Using NDK sounds nice.

Extravert-ir (approved):

PR #4115, opened 2021-11-19
[USER32] GetQueueStatus() should not return 0 when passing QS_ALLINPUT flag

julcar:
## Purpose
Some versions of libgdk-win32-2.0.0.dll enters into an infinite loop when calling GetQueueStatus() with QS_ALLINPUT flag.

This happens because the RealGetQueueStatus() function performs a wrong flag check, thus calling it with QS_ALLINPUT leads to a 0 returned and setting the last error to ERROR_INVALID_FLAGS.

This seems to be fatal for [libgdk calls to GetQueueStatus()](https://gitlab.gnome.org/search?search=GetQueueStatus&group_id=8&project_id=665&scope=&search_code=true&snippets=false&repository_ref=2.24.11&nav_source=navbar) as they relies in GetQueueStatus() not returning 0.

@julenuri told me some time ago that this issue is not present in Wine, so a quick look in [Wine's implementation of GetQueueStatus](https://source.winehq.org/git/wine.git/blob/32fb017d4a22be38ca271bf387e466e958601355:/dlls/user32/input.c#l527) can confirm this.

This fixes:
[CORE-15686](https://jira.reactos.org/browse/CORE-15686), [CORE-17551](https://jira.reactos.org/browse/CORE-17551), [CORE-11850](https://jira.reactos.org/browse/CORE-11850) and [CORE-8475](https://jira.reactos.org/browse/CORE-8475).

### DIFF
--- a/win32ss/user/user32/windows/message.c
+++ b/win32ss/user/user32/windows/message.c
@@ -2899,7 +2899,7 @@ DWORD
 WINAPI
 RealGetQueueStatus(UINT flags)
 {
-   if (flags & ~(QS_SMRESULT|QS_ALLPOSTMESSAGE|QS_ALLINPUT))
+   if (flags & ~(QS_ALLINPUT|QS_ALLPOSTMESSAGE|QS_SMRESULT))
    {
       SetLastError( ERROR_INVALID_FLAGS );
       return 0;
--- a/win32ss/user/user32/windows/message.c
+++ b/win32ss/user/user32/windows/message.c
@@ -2904,6 +2904,16 @@ RealGetQueueStatus(UINT flags)
       SetLastError( ERROR_INVALID_FLAGS );
       return 0;
    }
+   /** ATM, we do not support QS_RAWINPUT, but we need to support apps that pass
+    ** this flag along, while also working around QS_RAWINPUT checks in winetests.
+    ** Just set the last error to ERROR_INVALID_FLAGS but do not fail the call.
+    **/
+   if (flags & QS_RAWINPUT)
+   {
+      SetLastError(ERROR_INVALID_FLAGS);
+      flags &= ~QS_RAWINPUT;
+   }
+   /**/
    return NtUserxGetQueueStatus(flags);
 }
 
--- a/win32ss/user/user32/windows/message.c
+++ b/win32ss/user/user32/windows/message.c
@@ -2900,12 +2900,14 @@ WINAPI
 RealGetQueueStatus(UINT flags)
 {
    #define QS_TEMPALLINPUT 255 // ATM, do not support QS_RAWINPUT
+   DWORD ret;
    if (flags & ~(QS_SMRESULT|QS_ALLPOSTMESSAGE|QS_TEMPALLINPUT))
    {
       SetLastError( ERROR_INVALID_FLAGS );
-      return 0;
+      ret = 0;
    }
-   return NtUserxGetQueueStatus(flags);
+   ret = NtUserxGetQueueStatus(flags);
+   return ret;
 }
 
 
--- a/win32ss/user/user32/windows/message.c
+++ b/win32ss/user/user32/windows/message.c
@@ -2899,15 +2899,12 @@ DWORD
 WINAPI
 RealGetQueueStatus(UINT flags)
 {
-   #define QS_TEMPALLINPUT 255 // ATM, do not support QS_RAWINPUT
-   DWORD ret;
-   if (flags & ~(QS_SMRESULT|QS_ALLPOSTMESSAGE|QS_TEMPALLINPUT))
+   if (flags & ~(QS_SMRESULT|QS_ALLPOSTMESSAGE|QS_ALLINPUT))
    {
       SetLastError( ERROR_INVALID_FLAGS );
-      ret = 0;
+      return 0;
    }
-   ret = NtUserxGetQueueStatus(flags);
-   return ret;
+   return NtUserxGetQueueStatus(flags);
 }
 
 


win32ss/user/user32/windows/message.c
```diff
@@ -2900,12 +2900,14 @@ WINAPI
 RealGetQueueStatus(UINT flags)
 {
    #define QS_TEMPALLINPUT 255 // ATM, do not support QS_RAWINPUT
+   DWORD ret;
    if (flags & ~(QS_SMRESULT|QS_ALLPOSTMESSAGE|QS_TEMPALLINPUT))
    {
       SetLastError( ERROR_INVALID_FLAGS );
-      return 0;
+      ret = 0;
```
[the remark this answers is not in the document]
binarymaster: If you don't want to return 0 in this case, the patch might be just as simple as removing `return 0`. No need for local variable, as 0 assigned here is never used.

HBelusca (commented):
So, as I asked in the associated JIRA report, please check what sort of code the compiler generates **without**, and **with** the patch, in order to understand what's actually going on, since the `NtUserxGetQueueStatus` function is declared as returning a `DWORD` already (however it's also an inline).
I won't personally accept the patch to be committed yet until this problem is not fully understood... because at the moment it just looks like a magic workaround hack.

win32ss/user/user32/windows/message.c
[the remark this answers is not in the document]
HBelusca: Keeping the original `return 0;` should still work...

win32ss/user/user32/windows/message.c
```diff
@@ -2900,12 +2900,14 @@ WINAPI
 RealGetQueueStatus(UINT flags)
 {
    #define QS_TEMPALLINPUT 255 // ATM, do not support QS_RAWINPUT
+   DWORD ret;
    if (flags & ~(QS_SMRESULT|QS_ALLPOSTMESSAGE|QS_TEMPALLINPUT))
```
Doug-Lyons, line 2904: @julcar, I compared the wine code that you linked and it appears that if you only change this statement you can leave all of the other code in place. The new line should be as follows:
```
   if (flags & ~(QS_SMRESULT|QS_ALLPOSTMESSAGE|QS_ALLINPUT))
```
It appears that your real effect was not ever returning zero (0) and in our case, this was a very expensive thing.
Thanks. Doug Lyons

julcar: Right, here seems that the magic fix here is avoiding the function to return 0 and that may explain why just casting the return to DWORD was not enough to fix the bug.

HBelusca:
Please confirm that this latest PR update indeed fixes all the mentioned JIRA reports, so that it can be pushed.

win32ss/user/user32/windows/message.c
```diff
@@ -2899,8 +2899,7 @@ DWORD
 WINAPI
 RealGetQueueStatus(UINT flags)
 {
-   #define QS_TEMPALLINPUT 255 // ATM, do not support QS_RAWINPUT
-   if (flags & ~(QS_SMRESULT|QS_ALLPOSTMESSAGE|QS_TEMPALLINPUT))
+   if (flags & ~(QS_SMRESULT|QS_ALLPOSTMESSAGE|QS_ALLINPUT))
```
SergeGautherie, line 2902: `QS_TEMPALLINPUT` hack was added on 7d1736a1a2a39ab506a3fe94fdd0e27aef1d9c46.
Was that wrong? Or what changed since then, that allows to get rid of it?

julcar (commented):
@SergeGautherie well, that hack regressed a lot of GTK applications, I guess @jimtabor could say if the hack is not needed anymore.

julcar:
> Please confirm that this latest PR update indeed fixes all the mentioned JIRA reports, so that it can be pushed.

Tested gtk-demo and Ardour and both were working.

HBelusca:
Currently being run on testbots...

SergeGautherie:
> regressed

Could this be confirmed and Jira ticket marked as such? Thanks.

win32ss/user/user32/windows/message.c
HBelusca, line 2902: Reordering of flags so that we can find the counterpart in our win32k code by trivial grepping.
```suggestion
   if (flags & ~(QS_ALLINPUT|QS_ALLPOSTMESSAGE|QS_SMRESULT))
```

julcar:
> > Currently being run on testbots...
> 
> Testbots shows (-11) failures on user32:TrackMouseEvent and (+11) failures on user32:msg_queue for a net of no change. I suspect that if the developers look at the new failures in msg_queue, one of the developers can find a fix for these.

All of those tests were GetQueueStatus(qs_all_input) returning 0, which is (confirmed by GTK) the wrong behavior

HBelusca:
I'm going to commit here a separate commit for you to test with gimp et al. and see whether they still work with it, while at the same time trying to workaround the extra winetest failures.
If everything works it could be kept, otherwise it can be ditched.

win32ss/user/user32/windows/message.c
```diff
@@ -2904,6 +2904,16 @@ RealGetQueueStatus(UINT flags)
       SetLastError( ERROR_INVALID_FLAGS );
       return 0;
    }
+   /** ATM, we do not support QS_RAWINPUT, but we need to support apps that pass
+    ** this flag along, while also working around QS_RAWINPUT checks in winetests.
+    ** Just set the last error to ERROR_INVALID_FLAGS but do not fail the call.
+    **/
+   if (flags & QS_RAWINPUT)
```
SergeGautherie, line 2911: Swap the 2 `if` blocks!

HBelusca: why?

SergeGautherie: (Disregard, I misread this.)

HBelusca:
> > > Currently being run on testbots...
> > 
> > 
> > Testbots shows (-11) failures on user32:TrackMouseEvent and (+11) failures on user32:msg_queue for a net of no change. I suspect that if the developers look at the new failures in msg_queue, one of the developers can find a fix for these.
> 
> All of those tests were GetQueueStatus(qs_all_input) returning 0, which is (confirmed by GTK) the wrong behavior

Actually, most of these newly failing tests were returning that there was a key event stuck in the queue that was not removed by the winetest because it wanted now to use some RAWINPUT behaviour to remove it (and that we don't support yet).

HBelusca:
Heh, what a surprise from QT side!
https://bugreports.qt.io/browse/QTBUG-12721

Doug-Lyons:
> I'm going to commit here a separate commit for you to test with gimp et al. and see whether they still work with it, while at the same time trying to workaround the extra winetest failures. If everything works it could be kept, otherwise it can be ditched.

@HBelusca, I have tested your latest changes and they work even better than before for Gimp 2.8.22. Previously, we had mouse clicks that were ignored and now these seem to be much improved. Thanks.

HBelusca:
> > I'm going to commit here a separate commit for you to test with gimp et al. and see whether they still work with it, while at the same time trying to workaround the extra winetest failures. If everything works it could be kept, otherwise it can be ditched.
> 
> @HBelusca, I have tested your latest changes and they work even better than before for Gimp 2.8.22. Previously, we had mouse clicks that were ignored and now these seem to be much improved. Thanks.

Perhaps visible now also as the success of the `user32:TrackMouseEvent` winetest :)
However, we now skip some tests in `user32:msg_queue`: `Tests skipped: queuing key events not supported` 🤔

PR #4117, opened 2021-11-19
** TESTS FOR TESTBOTS **

HBelusca:
Revert "[PSDK] Sort out the QS_*** definitions. Add the Win7+ QS_TOUCH and QS_POINTER."

This reverts commit d78a5b6dbf9de09d905def711cb0c3f763fcaed8.

## Purpose

Strangely after d78a5b6d, `user32:msg_queue` skips test `Tests skipped: queuing key events not supported`. Investigate why...


Extravert-ir:
beep boop :robot:

modules/rostests/winetests/user32/input.c
```diff
@@ -79,8 +79,10 @@ static struct {
     BOOL sendinput_broken;
 } key_status;
 
+#if (_WIN32_WINNT >= 0x0602)
```
SergeGautherie, line 82: Make diff more explicit?
```suggestion
#if !defined(__REACTOS__) || _WIN32_WINNT >= 0x0602
```

modules/rostests/winetests/user32/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 
 remove_definitions(-DWINVER=0x502 -D_WIN32_WINNT=0x502)
-add_definitions(-DWINVER=0x602 -D_WIN32_WINNT=0x602)
+add_definitions(-DWINVER=0x600 -D_WIN32_WINNT=0x600)
```
binarymaster, line 3: Could this be just removed now?

HBelusca: If the original add_definitions (so, with 0x602) gets used, this winetest would be compiled with 0x602 (Win8+) definitions of these:
https://github.com/reactos/reactos/commit/d78a5b6dbf9de09d905def711cb0c3f763fcaed8
that would trigger problems as explained in the PR description.

binarymaster: I mean removing these two altogether to make it respect the DLLVER we use for building everything.

JoachimHenze (approved):
Approved. Please experiment with the suggestion from @binarymaster before doing the merge. If that would actually work, it would be better than 0x600.

modules/rostests/winetests/user32/input.c
SergeGautherie: And use `_WIN32_WINNT_WIN8`.

modules/rostests/winetests/user32/CMakeLists.txt
DarkFire01: I’d rather make it respect DLLVER,
We can’t change some of the stuff these newer tests require with the NTSHIM for instance. Better to hold off enabling them until we compile ros as NT6 across the board. That’s my two cents

DarkFire01 (changes requested):

cbialorucki:
Superseded by #8086

PR #4121, opened 2021-11-20
[EXPLORER] Fix for Start Button Wrong size when Using Themes

Doug-Lyons:
## Fix for Themed Start Buttons not being full-sized in y-dimension

_Changes to code to allow different calculation for 'cy' when theme is active._

JIRA issue: [CORE-16742](https://jira.reactos.org/browse/CORE-16742)

## Make Themed Start Buttons Taller to fill space available

### DIFF
--- a/base/shell/explorer/traywnd.cpp
+++ b/base/shell/explorer/traywnd.cpp
@@ -1631,13 +1631,11 @@ ChangePos:
         HWND hwndTaskToolbar = ::GetWindow(m_TaskSwitch, GW_CHILD);
         if (hwndTaskToolbar)
         {
-            SIZE EdgeSize;
-            EdgeSize.cy = GetSystemMetrics(SM_CYEDGE);
+            LONG EdgeSizeY = GetSystemMetrics(SM_CYEDGE);
             DWORD size = SendMessageW(hwndTaskToolbar, TB_GETBUTTONSIZE, 0, 0);
 
             /* Themed button covers Edge area as well */
-            StartSize.cy = HIWORD(size) + (m_Theme ? GetSystemMetrics(SM_CYEDGE) : 0);
-            }
+            StartSize.cy = HIWORD(size) + (m_Theme ? EdgeSizeY : 0);
         }
 
         if (m_StartButton.m_hWnd != NULL)
--- a/base/shell/explorer/traywnd.cpp
+++ b/base/shell/explorer/traywnd.cpp
@@ -1562,7 +1562,7 @@ ChangePos:
         else
         {
             WndSize.cx = StartBtnSize.cx;
-            WndSize.cy = StartBtnSize.cy - EdgeSize.cx;
+            WndSize.cy = StartBtnSize.cy - EdgeSize.cy;
         }
 
         if (WndSize.cx < g_TaskbarSettings.sr.Size.cx)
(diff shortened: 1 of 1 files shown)


JoachimHenze:
https://jira.reactos.org/browse/CORE-16742?focusedCommentId=130833&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-130833 implies that it might be just a workaround and not a correct fix. Nevertheless, if it wouldn't break anything else, then I would not oppose merging it.

base/shell/explorer/traywnd.cpp
```diff
@@ -1562,7 +1562,7 @@ class CTrayWindow :
         else
         {
             WndSize.cx = StartBtnSize.cx;
-            WndSize.cy = StartBtnSize.cy - EdgeSize.cx;
+            WndSize.cy = StartBtnSize.cy - EdgeSize.cy;
```
binarymaster, line 1565: Nice catch :)

JoachimHenze (approved):
Approved. Although some slight indications point towards explorer.exe maybe not being the correct binary. But at least some parts of this PR are definitely necessary and a clear improvement. Like the "nice catch" for example.

base/shell/explorer/traywnd.cpp
```diff
@@ -1628,14 +1628,21 @@ class CTrayWindow :
...
+            if (!m_Theme)
             {
-                DWORD size = SendMessageW(hwndTaskToolbar, TB_GETBUTTONSIZE, 0, 0);
                 StartSize.cy = HIWORD(size);
             }
+            else
+            {
+                /* Themed button covers Edge area as well */
+                StartSize.cy = HIWORD(size) + EdgeSize.cy;
+            }
```
HBelusca, line 1645: Seems that all the code from the `if (!m_Theme)` above down to here could be simplified by:
```c
            /* Themed button covers Edge area as well */
           StartSize.cy = HIWORD(size) + (m_Theme ? GetSystemMetrics(SM_CYEDGE) : 0);
```

base/shell/explorer/traywnd.cpp
```diff
@@ -1628,14 +1628,21 @@ class CTrayWindow :
...
             StartSize.cx = rcClient.right;
 
-        if (!m_Theme)
+        HWND hwndTaskToolbar = ::GetWindow(m_TaskSwitch, GW_CHILD);
+        if (hwndTaskToolbar)
         {
-            HWND hwndTaskToolbar = ::GetWindow(m_TaskSwitch, GW_CHILD);
-            if (hwndTaskToolbar)
+            SIZE EdgeSize;
+            EdgeSize.cy = GetSystemMetrics(SM_CYEDGE);
```
HBelusca, line 1635:
```suggestion
            LONG EdgeSizeY = GetSystemMetrics(SM_CYEDGE);
```
This didn't need to be a full SIZE structure if you're just using its cy member.

base/shell/explorer/traywnd.cpp
Doug-Lyons: I tried this and now I have a strange build failure. Let me try this again.

Doug-Lyons: I had an extra parenthesis. Now this should be good. Thanks for your suggestions @HBelusca.

base/shell/explorer/traywnd.cpp
```diff
@@ -1628,14 +1628,14 @@ class CTrayWindow :
...
-            if (hwndTaskToolbar)
-            {
-                DWORD size = SendMessageW(hwndTaskToolbar, TB_GETBUTTONSIZE, 0, 0);
-                StartSize.cy = HIWORD(size);
-            }
+            LONG EdgeSizeY = GetSystemMetrics(SM_CYEDGE);
+            DWORD size = SendMessageW(hwndTaskToolbar, TB_GETBUTTONSIZE, 0, 0);
+
+            /* Themed button covers Edge area as well */
+            StartSize.cy = HIWORD(size) + (m_Theme ? EdgeSizeY : 0);
```
HBelusca, line 1638: Btw since you use EdgeSizeY only when m_Theme is not NULL, there is no need to systematically call the GetSystemMetrics above in all cases. (that's why in my suggestion I placed that call here directly in place of EdgeSizeY.)

Doug-Lyons: Ah, Let me fix this as well. Thanks.

Doug-Lyons: OK. Should be done now. Let's see what this looks like. Thanks.

HBelusca:
How does everything look like when themes are disabled?

JoachimHenze:
Hbelusca I posted that already before the (pretty safe) optimizations of yours: https://jira.reactos.org/browse/CORE-16742?focusedCommentId=130833&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-130833

yagoulas:
A lot of stuff that are related to themes don't necessarily make sense in the code. Please make sure that as many things as possible work. Try many different themes and the classic theme as well. Also please try changing the metrics in as many configurations as possible.

Generally speaking if our explorer is fine in windows but not in ros, most probably the problem is not explorer (as long as you make sure that there are no missing fonts and stuff like that).

Doug-Lyons:
@JoachimHenze, Thanks for the commit.

PR #4125, opened 2021-11-26
[3DTEXT] Fix 3dtext.scr using near 100% CPU

Doug-Lyons:
## Stop 3D Text ScreenSaver using near 100% CPY

_Add Timer to allow updating screen less often._

JIRA issue: [CORE-17866](https://jira.reactos.org/browse/CORE-17866)

## Add Timer to 3dtext.c to slow screen update to reasonable level

_This fixes the 3D Text ScreenSaver using near 100% CPU._

### DIFF
--- a/base/applications/screensavers/3dtext/3dtext.c
+++ b/base/applications/screensavers/3dtext/3dtext.c
@@ -41,6 +41,8 @@ GLfloat extentY = 0.0f;
 
 HINSTANCE hInstance;
 BOOL fullscreen = FALSE;
+#define APP_TIMER             1                         // Graphics Update Timer ID
+#define APP_TIMER_INTERVAL    USER_TIMER_MINIMUM * 5    // Graphics Update Interval
 
 // Build Our Bitmap Font
 GLvoid BuildFont(GLvoid)
@@ -338,6 +340,7 @@ ScreenSaverProc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam)
 
             // Initialize The GL Screen Using Screen Info
             InitGL(Screen.right, Screen.bottom);
+            SetTimer(hWnd, APP_TIMER, APP_TIMER_INTERVAL, NULL);    // Graphics update timer
             break;
 
         case WM_DESTROY:
@@ -360,11 +363,16 @@ ScreenSaverProc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam)
         case WM_PAINT:
             DrawGLScene();
             SwapBuffers(hDC);
+            ValidateRect(hWnd, NULL);   // Don't update here. Wait for timer.
             break;
 
         case WM_SIZE: // Resizing The Screen
             // Resize To The New Window Size
             ReSizeGLScene(LOWORD(lParam), HIWORD(lParam));
+            break;
+
+        case WM_TIMER:  // Used to update graphic based on timer udpate interval
+            InvalidateRect(hWnd, NULL, TRUE);
             break;
 
         default:


base/applications/screensavers/3dtext/3dtext.c
```diff
@@ -41,6 +41,8 @@ GLfloat extentY = 0.0f;
 
 HINSTANCE hInstance;
 BOOL fullscreen = FALSE;
+#define APP_TIMER             1                         // Graphics Update Timer ID
+#define APP_TIMER_INTERVAL    USER_TIMER_MINIMUM * 5    // Graphics Update Interval
```
HBelusca, line 45:
```suggestion
#define APP_TIMER_INTERVAL    (USER_TIMER_MINIMUM * 5)  // Graphics Update Interval
```

base/applications/screensavers/3dtext/3dtext.c
```diff
@@ -338,6 +340,7 @@ ScreenSaverProc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam)
 
             // Initialize The GL Screen Using Screen Info
             InitGL(Screen.right, Screen.bottom);
+            SetTimer(hWnd, APP_TIMER, APP_TIMER_INTERVAL, NULL);    // Graphics update timer
```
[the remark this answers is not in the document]
Doug-Lyons: Probably a good practice. Done. Thanks.

base/applications/screensavers/3dtext/3dtext.c
```diff
@@ -360,13 +363,18 @@ ScreenSaverProc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam)
...
             SwapBuffers(hDC);
+            ValidateRect(hWnd, NULL);   // Don't update here. Wait for timer.
             break;
 
         case WM_SIZE: // Resizing The Screen
             // Resize To The New Window Size
             ReSizeGLScene(LOWORD(lParam), HIWORD(lParam));
             break;
 
+        case WM_TIMER:  // Used to update graphic based on timer udpate interval
```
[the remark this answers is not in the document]
Doug-Lyons: Fixed. Thanks.

base/applications/screensavers/3dtext/3dtext.c
Doug-Lyons: Changed. Thanks.

base/applications/screensavers/3dtext/3dtext.c
```diff
@@ -360,13 +363,18 @@ ScreenSaverProc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam)
         case WM_PAINT:
             DrawGLScene();
             SwapBuffers(hDC);
+            ValidateRect(hWnd, NULL);   // Don't update here. Wait for timer.
```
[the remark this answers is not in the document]
Doug-Lyons: Changed. Thanks.

JoachimHenze (approved):

JoachimHenze:
I verified: The PR does also build properly with the older toolchains on 0.4.14RC
RosBE2.1.6 VS2010SP1 dbg+rls
RosBE2.1.6 VS2010SP1 dbg+rls
I like it!

base/applications/screensavers/3dtext/3dtext.c
```diff
@@ -360,13 +369,20 @@ ScreenSaverProc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam)
         case WM_PAINT:
             DrawGLScene();
             SwapBuffers(hDC);
+            // Don't update here. Wait for timer.
+            ValidateRect(hWnd, NULL);
```
learn-more, line 373: Because this screensaver never actually called 'BeginPaint/EndPaint', we never marked the `WM_PAINT` as done, so probably ReactOS was spamming `WM_PAINT`s ?

It would be interesting to investigate the behavior on windows:
Does windows automatically mark a window as `done` when `WM_PAINT` returned without calling `BeginPaint/EndPaint`, or does the one calling the ScreenSaverProc do that?

Doug-Lyons: @learn-more, I can wrap this code within ```BeginPaint/EndPaint``` if you think that it is needed. It was never there before. Do you think that it is needed? Does anyone think that it is needed? Thanks.

HBelusca: What @learn-more was asking is for someone to investigate e.g. what would already happen, without all this timing code but just by wrapping said code with the Begin/EndPrint alone, and see what happens (on both ReactOS and Windows).

learn-more: I am assuming that it does not have a near 100% processor usage on Windows.

From this assumption, I am speculating on why that would not be the case. I am observing a potential bug (the WM_PAINT) is never marked as 'done' in our code.
* Because I do not know if a screensaver is supposed to do this, it could be that the process that loaded the screensaver should be responsible for this (`the one calling the ScreenSaverProc`).
* Another option is that Windows detects this mis-behavior (not clearing the dirty flag after a WM_PAINT) and it either lowers the priority for the next drawing call, it it automatically marks the window as 'no longer dirty'.

Doug-Lyons: @learn-more, Your assumption is correct. After These changes Windows shows near 0% CPU usage on W2KSP2. Please review the video in CORE-17866 for proof. The video shows both the old and the new in sequential tests with the % CPU usage showing in the Task Manager's Performance tab. Thanks.

Doug-Lyons: Sorry that I did not make this clearer. When the WEBM shows 3dtextOld.scr selected, then it is running the unpatched version. When it shows 3dtextNew.scr selected it is running a version patched with this PR's changes.

W2KSP2_3dtext_scr_UnPatched

W2KSP2_3dtext_scr_Patched

Doug-Lyons: @HBelusca, So the answer to your question "But does our original screensaver (without any fixes whatsoever) uses ~ 100% CPU on windows itself?" is definitely "Yes". The proof is in the WEBM. Thanks.

learn-more: ok, so our screensaver is clearly bugged, which would make sense if it tries to refresh at top speed.

Thanks.

base/applications/screensavers/3dtext/3dtext.c
```diff
@@ -360,13 +369,20 @@ ScreenSaverProc(HWND hWnd, UINT message, WPARAM wParam, LPARAM lParam)
         case WM_PAINT:
             DrawGLScene();
             SwapBuffers(hDC);
+            // Don't update here. Wait for timer.
```
learn-more, line 372:
```suggestion
            // Mark this window as updated, so the OS won't ask us to update it again.
```

Doug-Lyons: Done. Thanks.

base/applications/screensavers/3dtext/3dtext.c
HBelusca: @Doug-Lyons thanks for your investigations!
What I still wonder however, based on the suggestion by @learn-more , is whether the original screensaver, with the addition of the following minimal patch (without all the WM_TIMER thing, i.e. this is NOT the current PR!) would also reduce the CPU usage on both ReactOS and Windows:
```c
        case WM_PAINT:
        {
            PAINTSTRUCT ps;
            BeginPaint(hWnd, &ps);
            DrawGLScene();
            SwapBuffers(hDC);
            EndPaint(hWnd, &ps);
            break;
```

EDIT: Heh! Other people have also encountered this problem! See for example:
https://stackoverflow.com/questions/2842319/swapbuffers-causes-redraw
https://docs.microsoft.com/en-us/windows/win32/opengl/the-program-ported-to-win32
https://www.khronos.org/opengl/wiki/Platform_specifics:_Windows

Doug-Lyons: I tried to attach a video, but received a message that the file type was not accepted.
Here is a screen shot. There is no animation and no movement of the sceensaver either in the small display or full screen.
I will try and attach the video in the CORE Jira Report.

W2KSP2_3dtextHB_scr

HBelusca: Thanks @Doug-Lyons . So the conclusion seems, that we either need to fix the screensaver the way you did it, or using an alternative like in the links I posted above.

learn-more (approved):
lgtm

Doug-Lyons:
@HBelusca, Thanks for your help in getting this committed.

PR #4126, opened 2021-11-27
[RICHED20] Fail elegantly if pRow or pPara is NULL on ME_EnsureVisible

katahiromz:
## Purpose
Baidu Browser installer used to get an assertion failure.
JIRA issue: [CORE-16578](https://jira.reactos.org/browse/CORE-16578)

## Proposed changes

- Fail elegantly on `ME_EnsureVisible` function if `pRow` or `pPara` is `NULL`.

## TODO

- [x] Do tests.

## Screenshots
BEFORE:
before
AFTER:
after


milawynsrealm:
It might be easier if you use an **#ifndef** around the assertions instead. It would simplify the code a bit, and help prevent any dead code.

SergeGautherie:
LGTM, until a future WINESYNC removes these 2 `assert()`.

PR #4127, opened 2021-11-27
[NTUSER] Fix where linking newly created window is done

katahiromz:
## Purpose
Based on I_Kill_Bugs' patch.
JIRA issue: [CORE-12052](https://jira.reactos.org/browse/CORE-12052)

## Proposed changes

- Move the code of linking the window.

## TODO

- [x] Do tests.

### DIFF
--- a/win32ss/user/ntuser/window.c
+++ b/win32ss/user/ntuser/window.c
@@ -2190,6 +2190,11 @@ co_UserCreateWindowEx(CREATESTRUCTW* Cs,
 
    Window->rcClient = Window->rcWindow;
 
+   if (Window->spwndNext || Window->spwndPrev)
+   {
+       ERR("Window 0x%p has been linked too early!\n", Window);
+   }
+
    if (!(Window->state2 & WNDS2_WIN31COMPAT))
    {
       if (Class->style & CS_PARENTDC && !(ParentWindow->style & WS_CLIPCHILDREN))
@@ -2198,10 +2203,6 @@ co_UserCreateWindowEx(CREATESTRUCTW* Cs,
 
    if ((Window->style & (WS_CHILD | WS_POPUP)) == WS_CHILD)
    {
-      if (Window->spwndNext || Window->spwndPrev)
-      {
-          ERR("Window 0x%p has been linked too early!\n", Window);
-      }
       if ( !IntIsTopLevelWindow(Window) )
       {
          if (pti != ParentWindow->head.pti)
--- a/win32ss/user/ntuser/window.c
+++ b/win32ss/user/ntuser/window.c
@@ -2204,10 +2204,10 @@ co_UserCreateWindowEx(CREATESTRUCTW* Cs,
       }
       if ( !IntIsTopLevelWindow(Window) )
       {
-         if (pti != Window->spwndParent->head.pti)
+         if (pti != ParentWindow->head.pti)
          {
             //ERR("CreateWindow Parent in.\n");
-            UserAttachThreadInput(pti, Window->spwndParent->head.pti, TRUE);
+            UserAttachThreadInput(pti, ParentWindow->head.pti, TRUE);
          }
       }
    }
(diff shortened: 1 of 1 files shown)


win32ss/user/ntuser/window.c
```diff
@@ -2226,6 +2216,16 @@ co_UserCreateWindowEx(CREATESTRUCTW* Cs,
       goto cleanup;
    }
 
+   /* Link the window */
+   if (NULL != ParentWindow)
```
binarymaster, line 2220:
```suggestion
   if (ParentWindow != NULL)
```

win32ss/user/ntuser/window.c
```diff
@@ -2226,6 +2216,16 @@ co_UserCreateWindowEx(CREATESTRUCTW* Cs,
       goto cleanup;
    }
 
+   /* Link the window */
+   if (ParentWindow != NULL)
+   {
+      /* Link the window into the siblings list */
+      if ((Cs->style & (WS_CHILD|WS_MAXIMIZE)) == WS_CHILD)
```
SergeGautherie, line 2223:
```suggestion
      if ((Cs->style & (WS_CHILD | WS_MAXIMIZE)) == WS_CHILD)
```

win32ss/user/ntuser/window.c
```diff
@@ -2190,16 +2190,6 @@ co_UserCreateWindowEx(CREATESTRUCTW* Cs,
...
-   /* Link the window */
-   if (NULL != ParentWindow)
-   {
-      /* Link the window into the siblings list */
-      if ((Cs->style & (WS_CHILD|WS_MAXIMIZE)) == WS_CHILD)
-          IntLinkHwnd(Window, HWND_BOTTOM);
-      else
-          IntLinkHwnd(Window, hwndInsertAfter);
-   }
-
```
HBelusca, line 2202: Note: Below, in (new) lines 2203 and 2206 (just below the `IntIsTopLevelWindow` check), it is suspicious to use `Window->spwndParent` instead of the original `ParentWindow`, because the window hook made above (lines around 2131) might have changed it by calling `SetParent` too early (either because the user-mode application is buggy or because it wants to trigger an error in win32k).
Also we see that code just above (e.g. lines 2180 and down) still uses the original known `ParentWindow` even after said hook.

HBelusca: It therefore may be worth somewhere to add a check to see whether `Window` has been linked too early and if so, at least now for diagnosis purposes, have some sort of DPRINT 🤔

HBelusca:
I would replace in the commit title: `Fix the timing of linking the window ` by something like `Fix where linking newly created window is done`  (since "timing" could infer there is something time-related, like a race condition or something like that, which is obviously not the case).

win32ss/user/ntuser/window.c
katahiromz: Excuse me, but how can I check the linking?

HBelusca: With some code similar to this:
```c
    if (Window->spwndNext || Window->spwndPrev)
    {
        ERR("Window 0x%p has been linked too early!\n", Window);
    }
```

katahiromz: Added the check.

win32ss/user/ntuser/window.c
```diff
@@ -2208,6 +2198,10 @@ co_UserCreateWindowEx(CREATESTRUCTW* Cs,
 
    if ((Window->style & (WS_CHILD | WS_POPUP)) == WS_CHILD)
    {
+      if (Window->spwndNext || Window->spwndPrev)
+      {
+          ERR("Window 0x%p has been linked too early!\n", Window);
+      }
```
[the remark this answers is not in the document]
katahiromz: Moved.

win32ss/user/ntuser/window.c
```diff
@@ -2208,6 +2198,10 @@ co_UserCreateWindowEx(CREATESTRUCTW* Cs,
...
       if ( !IntIsTopLevelWindow(Window) )
       {
          if (pti != Window->spwndParent->head.pti)
```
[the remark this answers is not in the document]
katahiromz: Used `ParentWindow`.

HBelusca (commented):
OK

PR #4131, opened 2021-12-01
[USETUP] Improve pt-PT translation

zecarlos1957:
[USETUP] Set menu item length to maximum string length

In pt-PT language in usetup.c BootLoaderPage  the menu bar when selected is shorter than the string length
Better solution by improving pt-PT translation

### DIFF
--- a/base/setup/usetup/usetup.c
+++ b/base/setup/usetup/usetup.c
@@ -3921,7 +3921,7 @@ BootLoaderPage(PINPUT_RECORD Ir)
     }
 
     MUIDisplayPage(BOOT_LOADER_PAGE);
-    CONSOLE_InvertTextXY(8, Line, 68, 1);
+    CONSOLE_InvertTextXY(8, Line, 60, 1);
 
     while (TRUE)
     {
--- a/base/setup/usetup/usetup.c
+++ b/base/setup/usetup/usetup.c
@@ -3930,7 +3930,7 @@ BootLoaderPage(PINPUT_RECORD Ir)
         if ((Ir->Event.KeyEvent.uChar.AsciiChar == 0x00) &&
             (Ir->Event.KeyEvent.wVirtualKeyCode == VK_DOWN))  /* DOWN */
         {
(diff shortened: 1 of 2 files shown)


binarymaster:
s/lenght/length/

zecarlos1957:
I try to find a way to get the string lenght in BootLoaderPage but without success
Maybe you prefer the menu bar set to the string lenght?

zecarlos1957:
ReactOSboot

binarymaster (commented):
It seems this PR supersedes @GeoB99's #2756 which was not finished. I see hardcoded values here, while the previous PR had a better solution without hardcoding.

GeoB99:
>which was not finished

Actually it was finished in its very own form of a fix but @HBelusca told me he's already taking care of this with his another improvements work in USETUP regarding VBR/MBR work, whatever that is, and that my patch might conflict with his work.

base/setup/usetup/usetup.c
```diff
@@ -3930,7 +3930,7 @@ BootLoaderPage(PINPUT_RECORD Ir)
         if ((Ir->Event.KeyEvent.uChar.AsciiChar == 0x00) &&
             (Ir->Event.KeyEvent.wVirtualKeyCode == VK_DOWN))  /* DOWN */
         {
-            CONSOLE_NormalTextXY(8, Line, 60, 1);
+   //         CONSOLE_NormalTextXY(8, Line, 68, 1);
```
GeoB99, line 3933: What, why?

JoachimHenze:
Wouldn't it be better and more safe to shorten the pt-PT text instead of touching that code?

zecarlos1957:
> Wouldn't it be better and more safe to shorten the pt-PT text instead of touching that code?

Sounds like a good idea to me, with the improved translation into Portuguese. Thanks.

zecarlos1957:
OPPS. I deleted this branch by mistake

binarymaster (approved):
Ah so this PR is now translation-only. LGTM.

HBelusca (approved):

PR #4135, opened 2021-12-03
[WINLOGON] Start screen saver with current user

Doug-Lyons:
## Fix Wrong Text Displayed when 3D Text ScreenSaver Times Out

_Change User Context of creating the screensaver process in winlogon.exe ._

JIRA issue: [CORE-17875](https://jira.reactos.org/browse/CORE-17875)

## Change context that 3D Text ScreenSaver runs as to Current User

_This fixes the wrong text of "ReactOS Rocks!" being displayed when this screensaver times out._

### DIFF
--- a/base/applications/screensavers/3dtext/settings.c
+++ b/base/applications/screensavers/3dtext/settings.c
@@ -47,4 +47,10 @@ VOID SaveSettings(VOID)
         RegSetValueEx(hkey, _T("DisplayString"), 0, REG_SZ, (LPBYTE)m_Text, (_tcslen(m_Text) + 1) * sizeof(TCHAR));
         RegCloseKey(hkey);
     }
+    if (RegCreateKeyEx(HKEY_USERS, _T(".DEFAULT\\Software\\Microsoft\\ScreenSavers\\Text3D"), 0,
+        _T(""), 0, KEY_WRITE, NULL, &hkey, NULL) == ERROR_SUCCESS)
+    {
+        RegSetValueEx(hkey, _T("DisplayString"), 0, REG_SZ, (LPBYTE)m_Text, (_tcslen(m_Text) + 1) * sizeof(TCHAR));
+        RegCloseKey(hkey);
+    }
 }
--- a/base/applications/screensavers/3dtext/settings.c
+++ b/base/applications/screensavers/3dtext/settings.c
@@ -101,7 +101,7 @@ VOID SaveSettings(VOID)
     size_t CbDestLength = 0;
 
     if (RegCreateKeyExW(HKEY_CURRENT_USER, L"Software\\Microsoft\\ScreenSavers\\Text3D", 0,
-    L"", 0, KEY_WRITE, NULL, &hkey, NULL) == ERROR_SUCCESS)
+        L"", 0, KEY_WRITE, NULL, &hkey, NULL) == ERROR_SUCCESS)
     {
         StringCbLengthW(m_Text, sizeof(m_Text), &CbDestLength);
         RegSetValueExW(hkey, L"DisplayString", 0, REG_SZ, (LPBYTE)m_Text, CbDestLength + sizeof(WCHAR));
--- a/base/applications/screensavers/3dtext/settings.c
+++ b/base/applications/screensavers/3dtext/settings.c
@@ -18,39 +18,102 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
  */
 
+/* This program can actually be entered from two very different contexts.
+ * One is that it can be entered when a normal user is signed into Windows.
+ * This is the case when it is called from the Desk CPL in the ScreenSaver
+ * Tab. In this context, we will be able to open a registry key for the
+ * HKEY_CURRENT_USER and use it to find the Display String to use.
+ * The other case is when is is run when the screen saver timeout occurs.
+ * Now we are running without a normal logged in user and if we try and open
+ * HKEY_CURRENT_USER, we will get a Windows Error #2 - File not Found.
+ * In this case, we will have to use the values in HKEY_USER\.Default to
+ * find the Display String to use.
+*/
+
 #include "3dtext.h"
 
 #include <winreg.h>
+#include <strsafe.h>
 
-TCHAR m_Text[MAX_PATH] = _T("ReactOS Rocks!");
+WCHAR m_Text[MAX_PATH] = L"ReactOS Rocks!";
 
 VOID LoadSettings(VOID)
 {
-	HKEY hkey;
-	DWORD len = MAX_PATH * sizeof(TCHAR);
+    HKEY hkey;
+    DWORD len = MAX_PATH * sizeof(WCHAR);
+    DWORD dwRet;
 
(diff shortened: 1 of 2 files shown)


gonzoMD:
Just curious: Isn't the Screensaver text a "per user" setting? I mean, isn't that text saved in general with this reg key? Wouldn't some HKCU key a better place?

binarymaster (changes requested):
This really looks like a hack. I don't see anywhere in the code of this screensaver where it reads settings from the system account registry.

https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=ScreenSavers%5C%5CText3D

learn-more:
> This really looks like a hack. I don't see anywhere in the code of this screensaver where it reads settings from the system account registry.
> 
> https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=ScreenSavers%5C%5CText3D

Most likely the screensaver is not started as the same user that wrote the settings,
so a good fix would probably involve looking at that.

Doug-Lyons:
I plan to make some major changes to the settings.c file, so I am setting this as a "draft" for now.

Doug-Lyons:
The single failing test shows the error as follows:
LINK : fatal error LNK1201: error writing to program database 'D:\a\reactos\reactos\build\msvc_pdb\vfd.pdb'; check for insufficient disk space, invalid path, or insufficient privilege
This does not sound like any error that my program changes could have caused.

Doug-Lyons:
> This really looks like a hack. I don't see anywhere in the code of this screensaver where it reads settings from the system account registry.
> 
> https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=ScreenSavers%5C%5CText3D

@binarymaster, With the recent changes, do you still consider this a "hack"?

binarymaster:
> With the recent changes, do you still consider this a "hack"?

As long as you are attempting to save changes in other user's (SYSTEM) registry hive, I will consider this a hack.

Doug-Lyons:
> > With the recent changes, do you still consider this a "hack"?
> 
> As long as you are attempting to save changes in other user's (SYSTEM) registry hive, I will consider this a hack.

Can you please provide suggestions on how to make this work without it being a "hack"? Thanks.

learn-more:
> > > With the recent changes, do you still consider this a "hack"?
> > 
> > 
> > As long as you are attempting to save changes in other user's (SYSTEM) registry hive, I will consider this a hack.
> 
> Can you please provide suggestions on how to make this work without it being a "hack"? Thanks.

My comment from 2 days ago does this?

> > This really looks like a hack. I don't see anywhere in the code of this screensaver where it reads settings from the system account registry.
> > https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=ScreenSavers%5C%5CText3D
> 
> Most likely the screensaver is not started as the same user that wrote the settings, so a good fix would probably involve looking at that.
> 
> Probably `StartScreenSaver` (
> 
> https://github.com/reactos/reactos/blob/3fa57b8ff7fcee47b8e2ed869aecaf4515603f3f/base/system/winlogon/screensaver.c#L257
> 
> ) needs to start this process as a specific user, but I am not sure if this should always be done or only in specific cases.

Doug-Lyons:
> > > > With the recent changes, do you still consider this a "hack"?
> > > 
> > > 
> > > As long as you are attempting to save changes in other user's (SYSTEM) registry hive, I will consider this a hack.
> > 
> > 
> > Can you please provide suggestions on how to make this work without it being a "hack"? Thanks.
> 
> My comment from 2 days ago does this?
> 
> > > This really looks like a hack. I don't see anywhere in the code of this screensaver where it reads settings from the system account registry.
> > > https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=ScreenSavers%5C%5CText3D
> > 
> > 
> > Most likely the screensaver is not started as the same user that wrote the settings, so a good fix would probably involve looking at that.
> > Probably `StartScreenSaver` (
> > https://github.com/reactos/reactos/blob/3fa57b8ff7fcee47b8e2ed869aecaf4515603f3f/base/system/winlogon/screensaver.c#L257
> > 
> > ) needs to start this process as a specific user, but I am not sure if this should always be done or only in specific cases.

@learn-more, Thank you for your earlier comments and your taking the time to point them out to me once more in your most recent message. Unfortunately, at my present level of understanding of Windows, most of your comments are incomprehensible to me. This is something that I will have to try and absorb over some period of time before I can even begin to try and do what you suggest. Thanks again for your suggestions. I just need to have lots of time to try and learn these topics to be able to deal with them. Doug Lyons

learn-more:
> @learn-more, Thank you for your earlier comments and your taking the time to point them out to me once more in your most recent message. Unfortunately, at my present level of understanding of Windows, most of your comments are incomprehensible to me. This is something that I will have to try and absorb over some period of time before I can even begin to try and do what you suggest. Thanks again for your suggestions. I just need to have lots of time to try and learn these topics to be able to deal with them. Doug Lyons

In the function pointed out, the screensaver is simply started with `CreateProcessW`.
Because this is called from `Winlogon`, the screensaver is started on the same account as `Winlogon`, which is `Local System` in our case.

My suggestion is that the process should probably be started using `CreateProcessAsUser`, specifying the current logged-in user. (This user is already available, as hinted by `if (!ImpersonateLoggedOnUser(Session->UserToken))`).

However, I do not know if this should always be done, or that there is some switch that allows to control under which user the screensaver is started.

Doug-Lyons:
@learn-more, Thanks lots. I did not know that `CreateProcessAsUser` even existed. This simple change seems to make everything work. Since currently ReactOS is only using the `administrator` user, this might require some fine tuning later on when we support multiple users.

binarymaster (approved):
Looks ok now, I guess. - Note that I didn't verified what exactly Windows does in different scenarios.

gonzoMD:
Slightly off topic, but I'd suggest to set dwCreationFlags to IDLE_PRIORITY_CLASS.
See: https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-setpriorityclass

trivalik:
@gonzoMD Where? I cannot find in commit call to SetPriorityClass. Just create a pull request is the easiest.

Doug-Lyons:
> Just curious: Isn't the Screensaver text a "per user" setting? I mean, isn't that text saved in general with this reg key? Wouldn't some HKCU key a better place?

@gonzoMD, After the latest changes, we are now using only an HKCU registry key. Thanks.

Doug-Lyons:
> Slightly off topic, but I'd suggest to set dwCreationFlags to IDLE_PRIORITY_CLASS. See: https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-setpriorityclass

@gonzoMD, I agree that the screen saver should be IDLE_PRIORITY_CLASS and I have changed this PR to reflect this. Can you add your approval to the PR? Thanks.

learn-more (approved):
lgtm

Doug-Lyons:
@binarymaster, Thanks for your help in getting this merged. Now this screen saver 'mostly' works. I have more PR's for improving it upcoming.

PR #808, opened 2018-08-24
Fixes for cmake 3.11

ariscop:
With these two commits cmake 3.11 successfully builds reactos

VSSolution doesn't work, presumably depending on rosbe's patched cmake; ninja+msvc should but has not been tested

PR #4136, opened 2021-12-04
Cmake enable exports default

HBelusca:
## Purpose

"Recent" CMake versions do not allow EXECUTABLE targets to export symbols (only libraries), unless the `ENABLE_EXPORTS` property is enabled by-target. I argue that this is unnatural, because the original behaviour (allowing EXE targets to export symbols) is what is natural for PE binaries on Windows/NT-like platforms.

Therefore, I globally enable `ENABLE_EXPORTS` to be set (`ON`, or `TRUE`), by setting the global [`CMAKE_ENABLE_EXPORTS`](https://cmake.org/cmake/help/latest/variable/CMAKE_ENABLE_EXPORTS.html) variable, only when build-configuring the main targets (not the host-tools).

Related JIRA issue: [CORE-15406](https://jira.reactos.org/browse/CORE-15406)
See also PR #1335 and #808, as well as commit 957e566.

## Proposed changes

```
[CMAKE] Globally enable executables (not only libraries) to be able to export symbols...

... following the ENABLE_EXPORTS CMake "débâcle", partly-fixed originally
by commit d8e92b5 (see PR #1335), and finally by commit 957e566.
```

```
[APITESTS][FREELDR][NTVDM][SPOOLSV] Remove now-unneeded ENABLE_EXPORTS property.
```

### DIFF
--- a/sdk/cmake/CMakeMacros.cmake
+++ b/sdk/cmake/CMakeMacros.cmake
@@ -496,6 +496,8 @@ if(NOT MSVC_IDE)
 
     function(add_executable name)
         _add_executable(${name} ${ARGN})
+        # cmake adds a module_EXPORTS define when compiling a module with exports. We don't use that.
+        set_target_properties(${name} PROPERTIES DEFINE_SYMBOL "")
         add_clean_target(${name})
     endfunction()
 elseif(USE_FOLDER_STRUCTURE)
@@ -528,6 +530,8 @@ elseif(USE_FOLDER_STRUCTURE)
 
     function(add_executable name)
         _add_executable(${name} ${ARGN})
+        # cmake adds a module_EXPORTS define when compiling a module with exports. We don't use that.
+        set_target_properties(${name} PROPERTIES DEFINE_SYMBOL "")
         string(SUBSTRING ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_SOURCE_DIR_LENGTH} -1 CMAKE_CURRENT_SOURCE_DIR_RELATIVE)
         set_property(TARGET "${name}" PROPERTY FOLDER "${CMAKE_CURRENT_SOURCE_DIR_RELATIVE}")
     endfunction()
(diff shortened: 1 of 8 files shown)


binarymaster (approved):
One downside is, without this PR it's easier to see which of executables are special to have exports. But this refactoring looks good to me. Less text in cmake scripts => faster builds. :smile: 

If this doesn't break anything, ship it. :ship:

learn-more:
What do we gain by this?



ps: I doubt that this will make the build faster.

sdk/cmake/CMakeMacros.cmake
```diff
@@ -540,6 +544,12 @@ else()
             set_target_properties(${name} PROPERTIES DEFINE_SYMBOL "")
         endif()
     endfunction()
+
+    function(add_executable name)
```
ThFabba, line 548: This does not seem worth doing just to clear out `DEFINE_SYMBOLS` -- overriding builtins has significant performance implications in CMake.

HBelusca: Maybe, maybe not... anyway `add_library` is already overloaded similarly for the same purpose.

HBelusca:
> What do we gain by this?

What do we gain? To get back the old natural behaviour of having the exports present whenever a def file is added to the list of source files for whatever is being compiled -- this is what happens already naturally when making libraries/dlls etc. , and with exes now, as this used to be, without having to manually add an ENABLE_EXPORTS option. Indeed, see how this was forgotten for the other EXE build targets?

> ps: I doubt that this will make the build faster.

It should not make it slower either.

learn-more:
> > What do we gain by this?
> 
> What do we gain? To get back the old natural behaviour of having the exports present whenever a def file is added to the list of source files for whatever is being compiled -- this is what happens already naturally when making libraries/dlls etc. , and with exes now, as this used to be, without having to manually add an ENABLE_EXPORTS option. Indeed, see how this was forgotten for the other EXE build targets?
> 
> > ps: I doubt that this will make the build faster.
> 
> It should not make it slower either.

I think Thomas' reply clearly states the opposite.

HBelusca:
> The answer is clear. This isn’t doing whats intended. Closed for now. Thank you!

The PR _**DOES**_ what's intended in its description (it was tested to work on clean configure + build), but others argued just by eye-review that it wasn't that optimal, and I opposed the observation that `add_library` was already doing the same and nobody complained about it.

PR #4139, opened 2021-12-05
CMake spec2def improve

HBelusca:
## Purpose

Do some simple cmake files cleanup, and improve some aspects of `spec2def()` function.

## Proposed changes

- Some cmake files cleanup:
```
[CMAKE] Get rid of the macro_mc() macro.

Since COMMAND_MC is the same for both MSVC and GCC, get rid of the
macro_mc() macro and directly use said command in add_message_headers().
```
```
[CMAKE] MSVC's "SPEC2DEF_ARCH" == GCC's "ARCH2", so use a common name instead. Simplify its initialization. Move __x86_64 to global CMakeLists.txt.
```
```
[CMAKE] Simplify generate_import_lib() code.

- Simplify by storing the implib def file name in a variable, as done in msvc.cmake.
- Remove deprecated STATIC_LIBRARY_FLAGS usage.
```

- Improve cmake-file `spec2def()` function:
```
[CMAKE] Move spec2def() function common implementation into CMakeMacros.cmake.

The slight different spec2def invocation command definition is stored
in a COMMAND_SPEC2DEF variable, used by both generate_import_lib() and
spec2def() functions.
```
```
[CMAKE] Allow the spec2def() function to take a list of .spec files (more than one).

When more than one .spec file are present, they get concatenated
together and the resulting file is passed to spec2def.
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -135,7 +135,9 @@ if(NOT CMAKE_CROSSCOMPILING)
         if(ARCH STREQUAL "i386")
             add_definitions(/D_X86_ /D__i386__ /DWIN32 /D_WINDOWS)
         elseif(ARCH STREQUAL "amd64")
-            add_definitions(-D_AMD64_ -D__x86_64__ /DWIN32 -D_WINDOWS)
+            # '__x86_64': This is NOT a typo.
+            # See https://software.intel.com/en-us/forums/topic/404643
+            add_definitions(-D_AMD64_ -D__x86_64__ -D__x86_64 /DWIN32 -D_WINDOWS)
         endif()
         if(MSVC_VERSION GREATER 1699)
             add_definitions(/D_ALLOW_KEYWORD_MACROS)
@@ -256,7 +258,9 @@ Enable this if the module uses typeid or dynamic_cast. You will probably need to
         if (NOT (MSVC AND CMAKE_C_COMPILER_ID STREQUAL "Clang"))
             add_compile_definitions(_M_AMD64)
         endif()
-        add_definitions(-D_AMD64_ -D__x86_64__ -D_WIN64)
+        # '__x86_64': This is NOT a typo.
+        # See https://software.intel.com/en-us/forums/topic/404643
+        add_definitions(-D_AMD64_ -D__x86_64__ -D__x86_64 -D_WIN64)
     elseif(ARCH STREQUAL "arm")
         # _M_ARM is already defined by toolchain
         add_definitions(-D_ARM_ -D__arm__ -DWIN32)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -135,7 +135,9 @@ if(NOT CMAKE_CROSSCOMPILING)
         if(ARCH STREQUAL "i386")
             add_definitions(/D_X86_ /D__i386__ /DWIN32 /D_WINDOWS)
         elseif(ARCH STREQUAL "amd64")
-            add_definitions(-D_AMD64_ -D__x86_64__ /DWIN32 -D_WINDOWS)
+            # '__x86_64': This is NOT a typo.
+            # See https://software.intel.com/en-us/forums/topic/404643
```
learn-more, line 139: Dead link

sdk/cmake/CMakeMacros.cmake
```diff
@@ -466,6 +466,74 @@ function(create_iso_lists)
...
 
+function(spec2def _dllname _spec_files)
+    cmake_parse_arguments(__spec2def "ADD_IMPORTLIB;NO_PRIVATE_WARNINGS;WITH_RELAY" "VERSION" "" ${ARGN})
+
+    # Get library basename
+    get_filename_component(_file ${_dllname} NAME_WE)
+
+    # Build a list of all the spec files.
+    list(LENGTH _spec_files _len)
+    if(_len GREATER 1)
```
learn-more, line 477: Is there a usecase for this?

CMakeLists.txt
binarymaster: It seems it was just moved from [here](https://github.com/reactos/reactos/pull/4139/files#diff-4a0d335f415e68b8558820e21f06204c347f48677fda1b26b890f12dd5fed4e8L362). But yes, can confirm, I can't retrieve it, even not archived in Web Archive.

julenuri: https://community.intel.com/t5/Software-Archive/Origins-of-X86-64-macro-used-in-header-files-of-GCC-and-Intel/td-p/967540?profile.language=ko Maybe some info here?

binarymaster: Found: https://community.intel.com/t5/Intel-C-Compiler/32-bit-intel/td-p/964398 (the link posted above has the exact thread name mentioned, that's how I found it)

sdk/cmake/CMakeMacros.cmake
HBelusca: Not yet in current master, but soon otherwise:
https://github.com/HBelusca/reactos/commit/ff5b29e25ef05e882dd11fbc121667fb1112443e
where the list of exports is modified when the DOS support is compiled in (or excluded), by adding or not adding the dos.spec file.

PR #4142, opened 2021-12-06
[SDK] Add winnt_old entry for ARM64 and start fixing ARM64 PE binaries 

DarkFire01:
## Purpose

- Grabbed definitions based on the latest SDK for arm64
- Added entries for ARM64 into pefixup and the geninc tool


sdk/include/xdk/winnt_old.h
```diff
@@ -2148,7 +2148,150 @@ typedef struct _DISPATCHER_CONTEXT
     PBYTE  NonVolatileRegisters;
     DWORD Reserved;
 } DISPATCHER_CONTEXT, *PDISPATCHER_CONTEXT;
+#elif defined(_ARM64_)
```
binarymaster, line 2151:
```suggestion
#elif defined(_ARM64_)

```

sdk/tools/CMakeLists.txt
```diff
@@ -50,7 +50,7 @@ if(NOT MSVC)
     add_subdirectory(rsym)
 
     add_host_tool(pefixup pefixup.c)
-if (ARCH STREQUAL "amd64")
+if ((ARCH STREQUAL "amd64") OR (ARCH STREQUAL "arm64"))
```
binarymaster, line 53:
```suggestion
if (ARCH STREQUAL "amd64" OR ARCH STREQUAL "arm64")
```

sdk/tools/geninc/geninc.c
```diff
@@ -10,6 +10,7 @@
 #define IMAGE_FILE_MACHINE_I386  0x014c
 #define IMAGE_FILE_MACHINE_AMD64 0x8664
 #define IMAGE_FILE_MACHINE_ARMNT 0x01c4
+#define IMAGE_FILE_MACHINE_ARM64 0xAA64
```
binarymaster, line 13:
```suggestion
#define IMAGE_FILE_MACHINE_ARM64 0xaa64
```

binarymaster (commented):
Some suggestions

sdk/include/xdk/winnt_old.h
```diff
@@ -2148,7 +2148,151 @@ typedef struct _DISPATCHER_CONTEXT
...
+    //
+
+    /* +0x000 */ DWORD ContextFlags;
+
+    //
+    // Integer registers
+    //
+
+    /* +0x004 */ DWORD Cpsr;       // NZVF + DAIF + CurrentEL + SPSel
+    /* +0x008 */ union {
```
HBelusca, line 2193: Indentation ugly; can we have the offsets on the right instead?

DarkFire01: This is just how offsets were in the SDK, We can change it if you guys would like

HBelusca: hm ok, if this is a direct copy-paste of their SDK then I understand (but we would like not direct copypaste of MS SDK code either).

DarkFire01: Sounds good I'll change it, I wanted to change the type definitions to be like the rest of our headers anyway so this works out good!

binarymaster (approved):

PR #4143, opened 2021-12-06
[VIDEOPRT] Fix video settings not persisting across reboots

Doug-Lyons:
## Video Settings should Persist across a Reboot

_Add code in the registry.c file to copy the saved settings to those about to be used._

JIRA issue: [CORE-17719](https://jira.reactos.org/browse/CORE-17719)

## Copy Saved Video Settings to Registry Key About to be Used

_This should fix the video settings not being saved after a reboot._
There are reports that not all emulators are fixed yet.
Maybe someone can review these in relation to this change and see why?

### DIFF
--- a/win32ss/drivers/videoprt/registry.c
+++ b/win32ss/drivers/videoprt/registry.c
@@ -565,10 +565,10 @@ IntCreateNewRegistryPath(
                                 OBJ_KERNEL_HANDLE | OBJ_CASE_INSENSITIVE,
                                 NULL,
                                 NULL);
-    Status = ZwOpenKey(&NewKey, KEY_READ, &ObjectAttributes);
+    Status = ZwOpenKey(&NewKey, KEY_READ | KEY_WRITE, &ObjectAttributes);
     if (!NT_SUCCESS(Status))
     {
-        ERR_(VIDEOPRT, "Failed to open settings key. Status 0x%lx\n", Status);
+        ERR_(VIDEOPRT, "Failed to open NewKey. Status 0x%lx\n", Status);
         return Status;
     }
 
@@ -578,10 +578,10 @@ IntCreateNewRegistryPath(
                                 OBJ_KERNEL_HANDLE | OBJ_CASE_INSENSITIVE,
                                 NULL,
                                 NULL);
-    Status = ZwOpenKey(&SettingsKey, KEY_READ, &ObjectAttributes);
+    Status = ZwOpenKey(&SettingsKey, KEY_READ | KEY_WRITE, &ObjectAttributes);
     if (!NT_SUCCESS(Status))
     {
-        ERR_(VIDEOPRT, "Failed to open settings key. Status 0x%lx\n", Status);
+        ERR_(VIDEOPRT, "Failed to open SettingsKey. Status 0x%lx\n", Status);
         ObCloseHandle(NewKey, KernelMode);
         return Status;
     }
--- a/win32ss/drivers/videoprt/registry.c
+++ b/win32ss/drivers/videoprt/registry.c
@@ -586,6 +586,10 @@ IntCreateNewRegistryPath(
         return Status;
     }
 
+    /* Copy the saved resolution to the source key */
+    IntCopyRegistryValue(NewKey, SettingsKey, L"DefaultSettings.XResolution");
+    IntCopyRegistryValue(NewKey, SettingsKey, L"DefaultSettings.YResolution");
+
     /* Copy the registry data from the legacy key */
     Status = IntCopyRegistryKey(SettingsKey, NewKey);
 


binarymaster (approved):
The fix looks trivial and does the thing. Instant approval. :+1:

win32ss/drivers/videoprt/registry.c
```diff
@@ -586,6 +586,10 @@ IntCreateNewRegistryPath(
         return Status;
     }
 
+    /* Copy the saved resolution to the source key */
```
SergeGautherie, line 589: Could you clarify whether goal is only to preserve these `NewKey` values, or also to actually update `SettingsKey` too?

win32ss/drivers/videoprt/registry.c
```diff
@@ -586,6 +586,10 @@ IntCreateNewRegistryPath(
...
+    IntCopyRegistryValue(NewKey, SettingsKey, L"DefaultSettings.XResolution");
+    IntCopyRegistryValue(NewKey, SettingsKey, L"DefaultSettings.YResolution");
```
SergeGautherie, line 591: Then, at line 581, you would "need" to add `KEY_SET_VALUE` or `KEY_WRITE` or the like.

win32ss/drivers/videoprt/registry.c
Doug-Lyons: @SergeGautherie, Just below the lines that I added we have the following:
`Status = IntCopyRegistryKey(SettingsKey, NewKey);`
In both functions here and above the first value is the "Source" and the second is the "Destination".
So in this statement the "NewKey" values are about to be overwritten with wrong resolution data ("SettingsKey").
My solution copies the data that had the resolution that we want to keep ("NewKey") to the source key in this statement.
So, when the source key here is copied to the destination ("NewKey") it will carry along the wanted resolution values.
I only want to carry the two resolution keys forward in this manner because other values may be important to the function that @tkreuzer was trying to fix in his previous commit.

win32ss/drivers/videoprt/registry.c
Doug-Lyons: No changes are needed there. Please see my comments above.

win32ss/drivers/videoprt/registry.c
vicmarcal: Line 571: "Failed to open settings key. Status 0x%lx\n" -> "Failed to open new settings key. Status 0x%lx\n" ? 571 vs 584, old copy-pasta?

win32ss/drivers/videoprt/registry.c
vicmarcal: Probably you want to check here the Status of both IntCopyRegistryValue calls. If in some emulators resolution is still failing is because probably one (or both) of these calls are failing and it is going unnoticed.

win32ss/drivers/videoprt/registry.c
vicmarcal: Ok, so you are preserving the data from DefaultSettings.XResolution and DefaultSettings.YResolution values stored in NewKey by copying them into SettingsKey before overwriting NewKey with SettingsKey. Now I gotcha.

SergeGautherie: > I only want to carry the two resolution keys forward

But your code is also actually updating `SettingsKey`, which seems unintended/unwanted, at least unexplained.
In other words, why do both keys need to be updated?

win32ss/drivers/videoprt/registry.c
SergeGautherie: > No changes are needed there.

That seems rather inconsistent to me: asking for read only, then doing a write.

Doug-Lyons: @vicmarcal, You may be correct, but if you look at Line-350 and Line-351 above you will see similar coding by @tkreuzer and there is no testing for success there. I normally like to follow the lead of previous developers unless I know of a good reason not to do so. Let's see what others think here.
I do have some ideas of how to trouble-shoot the other problems and that would be to add the following two lines just above the current new ones:
```
    DPRINT("SettingsKey is '%wZ'.\n", &DeviceExtension->RegistryPath);
    DPRINT("NewKey is '%wZ'.\n", &DeviceExtension->NewRegistryPath);
```
These could be "DPRINT1" as well. Maybe we might want to include these now. Let's see what others think.
 Thanks.

win32ss/drivers/videoprt/registry.c
SergeGautherie: > 571 vs 584, old copy-pasta?

True, it is not fully explicit, but correct nonetheless.
Additionally, reusing same string saves some resources.

Doug-Lyons: @SergeGautherie, Yes, you are correct in that I am updating the `SettingsKey` and the reason for this is that it is often wrong and contains the default resolution and we want it to be correct. Also, I have now figured out that your comment about needing changes for the ZwOpenKey functions was in regard to the DesiredAccess. I have made changes that now should account for both reading and writing of both values. In the added code, the Source and Destination keys are reversed from what was used previously in the statement below. Therefore, now we are both reading and writing both keys. 
@vicmarcal, I took care of the copy-pasta while here. Please see if you agree with what I am suggesting.
Thanks to both of you.

SergeGautherie: > it is often wrong

You sound as this is a hack and a proper fix for that key might be needed elsewhere/before.
Could you state what these 2 keys actually are, what they are used for, what happens if only one of them is updated (with proper values) here? Thanks.
In other words, are these key/value changes what Windows does?

> both reading and writing of both values

Right!

win32ss/drivers/videoprt/registry.c
```diff
@@ -565,10 +565,10 @@ IntCreateNewRegistryPath(
                                 OBJ_KERNEL_HANDLE | OBJ_CASE_INSENSITIVE,
                                 NULL,
                                 NULL);
-    Status = ZwOpenKey(&NewKey, KEY_READ, &ObjectAttributes);
+    Status = ZwOpenKey(&NewKey, KEY_READ | KEY_WRITE, &ObjectAttributes);
     if (!NT_SUCCESS(Status))
     {
-        ERR_(VIDEOPRT, "Failed to open settings key. Status 0x%lx\n", Status);
+        ERR_(VIDEOPRT, "Failed to open NewKey. Status 0x%lx\n", Status);
```
SergeGautherie, line 571: The change itself is fine with me, though I think these strings could stay as is (= identical).
Yet, if you change them, update the 3rd occurrence too.

win32ss/drivers/videoprt/registry.c
```diff
@@ -578,14 +578,18 @@ IntCreateNewRegistryPath(
                                 OBJ_KERNEL_HANDLE | OBJ_CASE_INSENSITIVE,
                                 NULL,
                                 NULL);
-    Status = ZwOpenKey(&SettingsKey, KEY_READ, &ObjectAttributes);
+    Status = ZwOpenKey(&SettingsKey, KEY_READ | KEY_WRITE, &ObjectAttributes);
```
SergeGautherie, line 581: That should work, though `KEY_QUERY_VALUE | KEY_WRITE` and `KEY_READ | KEY_SET_VALUE` should be enough and more expliclit.

win32ss/drivers/videoprt/registry.c
Doug-Lyons: Pleases reference a Line Number where the 3rd occurrence happens. Thanks.

win32ss/drivers/videoprt/registry.c
hpoussin: As said, this is a hack. We should never copy from 'new' to 'old' key.
See https://github.com/reactos/reactos/commit/ebd01880a353bbb3d1f110d52086e008900627da for probably a better fix.

win32ss/drivers/videoprt/registry.c
SergeGautherie: Line 332.

win32ss/drivers/videoprt/registry.c
Doug-Lyons: @hpoussin, I guess that you may not have noticed, but your "better fix" is just a revert of the last commit by @tkreuzer. So this re-introduces the problem with https://jira.reactos.org/browse/CORE-17688. Fixing this problem was the reason that @tkreuzer made this commit and it seemed to work based on feedback from the users. If you are set against copying from 'new' to 'old' key, then I have a new patch for your review in https://jira.reactos.org/browse/CORE-17719 that is called 'video_save_03.patch'. Thanks very much for your investigations and suggestions. I believe that together we will eventually fix this problem. Doug Lyons

SergeGautherie: >  "sometimes it is wrong for our purposes"

That still does not convince me your proposal is the right fix.

--

> ebd0188

It looks like that is basically reverting ecf3416.
Was the latter totally wrong? Or is an in-between solution (See @tkreuzer's commit description) needed?

oleg-dubinskiy: @Doug-Lyons as I just discussed in PM with @hpoussin, it is OK to revert that Timo's commit, since my tests prove that my recent c05a45e17eda057d8b92ba76feecf3532d8324c1 commit, which was a fix for https://jira.reactos.org/browse/CORE-17786 from the start, also fixes https://jira.reactos.org/browse/CORE-17688 properly. :slightly_smiling_face: So @tkreuzer's change is no longer needed and we now can safely revert it, to fix https://jira.reactos.org/browse/CORE-17719 too. :wink: And if your solution is hacky, it now seems no longer needed.
mm-pm

Doug-Lyons:
@oleg-dubinskiy, That's great news. I will close this PR and then we can wait for @hpoussin to make the revert commit. Thanks again. Doug Lyons

PR #4144, opened 2021-12-07
[WIN32K:NTUSER] Make sure to hold User Global Lock before mapping des…

SergeGautherie:
…ktop heap

Addendum to 8c098ee (0.4.15-dev-3001).
JIRA issue: [CORE-17890](https://jira.reactos.org/browse/CORE-17890)


learn-more (approved):
Makes sense, considering `UserHeapAlloc` which is called inside `IntMapDesktopView` asserts on the user lock being held.

Unless @jimtabor disagrees?

HBelusca (approved):

PR #4145, opened 2021-12-07
[VIDEOPRT] Match Zw*Key() DesiredAccess values

SergeGautherie:
even if kernel mode does not care.

### DIFF
--- a/win32ss/drivers/videoprt/videoprt.c
+++ b/win32ss/drivers/videoprt/videoprt.c
@@ -617,7 +617,7 @@ IntLoadRegistryParameters(VOID)
                                    NULL);
 
         Status = ZwCreateKey(&KeyHandle,
-                             KEY_WRITE,
+                             READ_CONTROL, // Non-0 placeholder: no use for this handle.
                              &ObjectAttributes,
                              0,
                              NULL,
--- a/win32ss/drivers/videoprt/registry.c
+++ b/win32ss/drivers/videoprt/registry.c
@@ -339,7 +339,7 @@ IntSetupDeviceSettingsKey(
                                OBJ_KERNEL_HANDLE | OBJ_CASE_INSENSITIVE,
                                NULL,
                                NULL);
-    Status = ZwOpenKey(&SourceKeyHandle, KEY_WRITE, &ObjectAttributes);
+    Status = ZwOpenKey(&SourceKeyHandle, KEY_READ, &ObjectAttributes);
     if (Status != STATUS_SUCCESS)
     {
         ERR_(VIDEOPRT, "ZwOpenKey failed for settings key: status 0x%lx\n", Status);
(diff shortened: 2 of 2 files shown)


learn-more:
This might conflict with #4143

win32ss/drivers/videoprt/videoprt.c
```diff
@@ -617,7 +617,7 @@ IntLoadRegistryParameters(VOID)
                                    NULL);
 
         Status = ZwCreateKey(&KeyHandle,
-                             KEY_WRITE,
+                             0,
```
ThFabba, line 620: You can't create a handle with zero access

SergeGautherie: Ok, I could leave it as is, or use a "lower" right and/or add a comment?

Meanwhile, did I miss something or was I unlucky and looked at broken calls?
https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/io/pnpmgr/pnpmgr.c;hb=HEAD#l695
https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/io/pnpmgr/pnpmgr.c;hb=HEAD#l714

ThFabba: Ah, I guess the "no 0 access handle" rule only applies to UM.
Setting a breakpoint on these examples says they succeed -- even the `ZwSetValueKey` following the second one.
`ObReferenceObjectByHandle` does
```
            if ((AccessMode == KernelMode) ||
                !(~GrantedAccess & DesiredAccess))
```
So I guess access rights are meaningless in kernel mode.
That also explains why the swapped values above don't cause failures.

No idea what best practice is, then. Principle of minimal access would suggest we should open _all_ KM handles with 0 access ^^

HBelusca: > Principle of minimal access would suggest we should open all KM handles with 0 access ^^

However this would then **not convey** the intent of what is expected to be done with said registry key (and I'm sure it's then **not** a best practice).

ThFabba: Yes, the ^^ was intended to express that I'm not serious about this approach.
But since it also doesn't have any effect whatsoever, let's not turn this into a big discussion. I'd say swap the obviously broken read/write cases, leave everything else be.

SergeGautherie: I agree not to change all to 0. Using matching values, even if useless, still feels good to me.
Then, this PR seems good as is. I'll check the other modules, especially the pnpmgr case, separately.

ThFabba: I think I'd still like to see a valid access value here. We have a test that clearly shows that if this were happening in UM, it would fail: https://git.reactos.org/?p=reactos.git;a=blob;f=modules/rostests/apitests/ntdll/NtCreateKey.c;hb=00ffe2dfff1269a0ba2331bf3e6104e68c9e4d32#l97

SergeGautherie: Ok... Reduced value and added comment.

ThFabba (approved):

HBelusca (approved):

PR #3638, opened 2021-05-03
[WIN32SS][NTDDRAW][NTGDI] Implement some DxEng* callbacks

oleg-dubinskiy:
## Purpose

Implement some `DxEng`* functions required by MS DirectDraw stack (ddraw.dll & dxg.sys) (at least the system spams about them in the log, when it loads these components).
Call appropriate functions inside them, whose actually do needed things, and update their prototypes accordingly.

JIRA issue: [CORE-17561](https://jira.reactos.org/browse/CORE-17561)

## Proposed changes

Implement the following functions:

- `DxEngAltLockSurface`:
    - call `EngLockSurface` inside it (already implemented);
    - update the prototype.
- `DxEngDeleteSurface`:
    - call `EngDeleteSurface` inside it (already implemented);
    - update the prototype.
- `DxEngReferenceHdev`:
    - call `PDEVOBJ_vReference` inside it (already implemented).
- `DxEngSelectBitmap`:
    - call `NtGdiSelectBitmap` inside it (already implemented);
    - update the prototype.
- `DxEngSetBitmapOwner`:
    - call `GreSetBitmapOwner` inside it (already implemented);
    - update the prototype.
- `DxEngUnreferenceHdev`:
    - call `PDEVOBJ_vRelease` inside it (already implemented).
- Get rid of now unused `IntGdi`(`Un`)`referencePdev`. **My analysis proves that in Windows, `DxEngReferenceHdev` and `DxEngUnreferenceHdev` call `PDEVOBJ` functions directly. So our `IntGdi` ones are not needed anymore.**
- Add some `DPRINT`s.

Although it requires some more callbacks to be implemented, but unfortunately, for others, appropriate `NtGdi`* functions are not implemented yet. Moreover, they are absent at all. :frowning_face:

PR #4146, opened 2021-12-07
[WIN32SS][ENG] EngCreateBitmap: mark created bitmaps as API bitmaps

oleg-dubinskiy:
## Purpose

Set an `API_BITMAP` flag for bitmaps created by `EngCreateBitmap`.
This avoids failure when created bitmap is passed to `NtGdiSelectBitmap`, since it checks for this flag and fails if it isn't set.
Otherwise, if failure occurs, the bitmap is hot handled properly.
Setting this flag outside of that function is not an option at all, because it is a public (exported) function, and in many cases it's called directly by the caller (in case with MS DDraw stack, it can be called also by dxg.sys from `dxgthk!EngCreateBitmap`, which directly calls `win32k!EngCreateBitmap`).
In particular, it allows Windows XP/2003 DirectDraw stack (ddraw.dll & dxg.sys) to properly work in ReactOS (in software emulation mode), even on real hardware. So now, a lot of DirectX 1-7 apps and games (whose can work without hardware acceleation), are working properly with ddraw.dll + dxg.sys replacement! Also Justin Miller (@DarkFire01) confirmed that even Direct3D software emulation now also works correctly on real hardware! :smiley: 

## Result

The following video proves that a lot of DirectX 1-7 apps and games are now working properly. :slightly_smiling_face: 

https://user-images.githubusercontent.com/26385117/145112643-558be109-d57b-4ac6-a389-d70858186ad2.mp4

In the video I tested the following:
- apps:
    - Microsoft DirectX 8.1 Diagnostic Tool (all 3 DirectDraw tests are now succeed instead of only 1).
- games:
    - Barbie Seahorse Adventures.
    - IcyTower from Rapps.
    - Mario Dash.
    - OpenSonic.
    - SimCity 3000 Demo.
    - Snoopy from Rapps.
    - Tux Racer.

I tested them many times earlier and can confirm that all of them use DirectDraw. They didn't work before my change and became work properly after them. :slightly_smiling_face:

### DIFF
--- a/win32ss/gdi/eng/surface.c
+++ b/win32ss/gdi/eng/surface.c
@@ -332,6 +332,9 @@ EngCreateBitmap(
     /* Get the handle for the bitmap */
     hbmp = (HBITMAP)psurf->SurfObj.hsurf;
 
+    /* Mark as API bitmap */
+    psurf->flags = API_BITMAP;
+
     /* Set public ownership */
     GDIOBJ_vSetObjectOwner(&psurf->BaseObject, GDI_OBJ_HMGR_PUBLIC);
 


oleg-dubinskiy:
Please note that this change will have the described effect **only in pair with #3638 PR applied!** :point_up: 
Also anyone who want to test it, needs to remove the following `ASSERT(FALSE);` from `IntGdiCreateDisplayDC`: https://git.reactos.org/?p=reactos.git;a=blob;f=win32ss/gdi/ntgdi/dclife.c;h=4671c16d4a7e4cfba440956dcbc40bb63a5af038;hb=HEAD#l1067. I'll also send a PR with removal of that. :ok_hand:

DarkFire01 (approved):
My opinion isn't the one that matters here! But i decided to pop in anyway, as a piece of evidence per say.
This PR has a massive improvement on the situation of native reactx/directx. This PR was tested on the follow graphics cards with massive success:
GTX 465
GT 330
GTX 560 ti
ATI HD7770
ATI HD5470
VMware SVGA II
NVIDIA NVS quadro 110m
and more! as @archeYR (Illen) has had some success as well

HBelusca (approved):

PR #4149, opened 2021-12-08
[W32TIME] Add Spanish (es-ES) translation

elhoir:
Spanish translation of w32time service. Created es-ES.rc file.

### DIFF
--- a/base/services/w32time/lang/es-ES.rc
+++ b/base/services/w32time/lang/es-ES.rc
@@ -1,0 +1,15 @@
+/*
+ * PROJECT:     ReactOS W32Time Service
+ * LICENSE:     GPL-2.0+ (https://spdx.org/licenses/GPL-2.0+)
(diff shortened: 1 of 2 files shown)


learn-more (changes requested):
It looks like you created this file, in that case it should also be included somewhere :)
In your case that would be:
https://github.com/reactos/reactos/blob/8dc8eef8cc304c86843112c11571a8a7ddc91fb6/base/services/w32time/w32time.rc

Should probably look like this:
```
#ifdef LANGUAGE_ES_ES
#include "lang/es-ES.rc"
#endif
```
(With the same indenting as the other languages)

elhoir:
saw it! you´re right! thank you!

learn-more:
Image from @elhoir in Mattermost:
image

elhoir:
imagen

PR #4152, opened 2021-12-10
[EXPLORER] Check fullscreen and bring the tray down or up

katahiromz:
## Purpose

The taskbar used to be blocking the full screen app.
JIRA issue: [CORE-16130](https://jira.reactos.org/browse/CORE-16130)

## Proposed changes

- Check whether the activated window is fullscreen upon `HSHELL_WINDOWCREATED`, `HSHELL_WINDOWDESTROYED`, `HSHELL_RUDEAPPACTIVATED` and `HSHELL_WINDOWACTIVATED`.
- If the window was fullscreen and not a special window, then adjust the taskbar Z-order.

## TODO

- [ ] Do tests.

### DIFF
--- a/base/shell/explorer/taskswnd.cpp
+++ b/base/shell/explorer/taskswnd.cpp
@@ -1565,6 +1565,7 @@ public:
         case HSHELL_RUDEAPPACTIVATED:
         case HSHELL_WINDOWACTIVATED:
             ActivateTask((HWND) lParam);
+            CheckFullscreen((HWND)lParam);
             break;
 
         case HSHELL_FLASH:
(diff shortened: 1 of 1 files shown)


JoachimHenze:
I have not tested the changes of the patch on ros yet,
but what is making me a bit sceptical, whether it is the ideal solution is, that
our **unpatched** explorer does not show the problem on XPSP3. On XPSP3 the **unpatched** ros explorer does properly hide the taskbar when DosBOX is switched to fullscreen.

I found that to be worth mentioning.

katahiromz:
I agree with @JoachimHenze. Closed.

sskras:
@katahiromz wrote:
> I agree with @JoachimHenze. Closed.

Does this mean that more (and detailed) tests are needed?

katahiromz:
> Does this mean that more (and detailed) tests are needed?

Nope. I believe Henze.

PR #4156, opened 2021-12-11
[HALX86] HalpDmaAllocateChildAdapter(): Fix memory leak, on error

SergeGautherie:
Simplify code, instead of adding 2 missing ObDereferenceObject().

JIRA issue: [CORE-17904](https://jira.reactos.org/browse/CORE-17904)

### DIFF
--- a/hal/halx86/generic/dma.c
+++ b/hal/halx86/generic/dma.c
@@ -477,12 +477,6 @@ HalpDmaAllocateChildAdapter(IN ULONG NumberOfMapRegisters,
                             (PVOID)&AdapterObject);
     if (!NT_SUCCESS(Status)) return NULL;
 
-    Status = ObReferenceObjectByPointer(AdapterObject,
-                                        FILE_READ_DATA | FILE_WRITE_DATA,
-                                        IoAdapterObjectType,
-                                        KernelMode);
-    if (!NT_SUCCESS(Status)) return NULL;
-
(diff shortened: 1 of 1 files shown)


ThFabba (approved):

hal/halx86/generic/dma.c
```diff
@@ -481,7 +481,11 @@ HalpDmaAllocateChildAdapter(IN ULONG NumberOfMapRegisters,
                                         FILE_READ_DATA | FILE_WRITE_DATA,
                                         IoAdapterObjectType,
                                         KernelMode);
-    if (!NT_SUCCESS(Status)) return NULL;
+    if (!NT_SUCCESS(Status))
```
ThFabba, line 484: This failure mode literally can't happen, so it's annoying that we have to have code for it at all. If you wanted to change this to a plain `ObReferenceObject`, I'd be okay with that.

SergeGautherie: Sure. Then, it seems it could be moved to (before/)after `ZwClose()` or even before `return`. Where do you prefer to have it?

ThFabba: It can't happen after `ZwClose`, since the point is to prevent `ZwClose` from destroying the adapter. But moving it after the `ObInsertObject` simplifies error handling, so that seems worth doing, yes.

SergeGautherie: Right. Done.

ThFabba (approved):

PR #4160, opened 2021-12-13
[KMTEST_DRV:NTOS_OB] ObType: Always check RtlStringCbPrintfW() status

SergeGautherie:
Do check `Status`, especially on Release build..


modules/rostests/kmtests/ntos_ob/ObType.c
```diff
@@ -290,9 +290,9 @@ ObtCreateObjects(VOID)
     {
         ASSERT(sizeof Name[i] == MAX_PATH * sizeof(WCHAR));
         Status = RtlStringCbPrintfW(Name[i], sizeof Name[i], L"\\ObtDirectory\\MyObject%d", i + 1);
-        ASSERT(Status == STATUS_SUCCESS);
+        ok_eq_hex(Status, STATUS_SUCCESS);
```
ThFabba, line 293: This does not make sense. Failure here is impossible, so this assert will only fire if there's a bug in the test -- which is exactly the point of an assert.

And, as a rule, if there was really a worry about this failing in release builds, `ok` would not magically make it safe -- you'd need a proper check & exit.

PR #4162, opened 2021-12-13
[HELP] Fixes (tr-TR) resource sporadic compilation error on VisualStudio

rm3ndozal:
## Purpose

Fixes (tr-TR) resource file sporadic compilation error RC2102 on VS2015 and VS2017

JIRA issue: [CORE-17910](https://jira.reactos.org/browse/CORE-17910)

## Proposed changes

Based on the work done at CORE-7405, moves the string split location so
the first part of the message isn't too long anymore.


binarymaster (approved):
Ah I see we have `IDS_HELP1` and `IDS_HELP2` for a reason :) There is no such problem in the other translation files, right? If so, looks good to me. :+1:

rm3ndozal:
> There is no such problem in the other translation files, right?

So far I haven't detected this problem in other resource files from this module, besides the [fr-FR.rc] file mentioned and fixed in CORE-7405. Tested VS2015, VS2017 and ROSBE 2.2.1.

PR #4168, opened 2021-12-16
Diverse Taskmgr fixes

HBelusca:
## Purpose

Mainly fix drawing issues in the performance page.
See also PR #3401.

## TODO

- [x] Finish the owner-drawn code (see console.cpl & calc).
- [ ] Remove refreshes of pages when they are not active.
- [x] Add multi-CPU support.


base/applications/taskmgr/about.c
```diff
@@ -1,9 +1,9 @@
 /*
- *  ReactOS Task Manager
+ * ReactOS Task Manager
```
binarymaster, line 2: Since you're fixing headers here, maybe it's worth to just switch to our new header syntax. Like this:
```c
/*
 * PROJECT:     ReactOS Task Manager
 * PURPOSE:     ...
 * COPYRIGHT:   ...
 */
```
See https://reactos.org/wiki/Coding_Style

HBelusca: Good idea, I will do that at some later point.

HBelusca:
Multi-CPU support in the perf stats and graphs:
image

HBelusca:
I'm going to re-create that PR to "bump" it in the list.

PR #4169, opened 2021-12-16
[TASKMGR] Make 'Hide When Minimized' non-default

katahiromz:
## Purpose

JIRA issue: N/A

## Proposed changes

- `HideWhenMinimized = FALSE;` at `LoadSettings` function.
- `AlwaysOnTop = TRUE;` at `LoadSettings` function.


HBelusca:
And why such a change? Is it the same as the default on a fresh installation of Windows **2003**?

katahiromz:
That is the default. If it is non-standard, the user will get confusedw

HBelusca:
> That is the default. If it is non-standard, the user will get confused.

Here are the defaults, by the way:
image
so indeed, 'Hide when minimized' is FALSE, but also 'Always on top' should be TRUE.

PR #4171, opened 2021-12-17
fix necessary that build runs through with removed NDEBUG

trivalik:
- fix build of ntfs/btree.c
- comment out not used variable in env_spec.cpp
- fix missed brackets in else wcache_lib.cpp

## Purpose

Fix build with removed NDEBUG

### DIFF
--- a/drivers/filesystems/udfs/Include/wcache_lib.cpp
+++ b/drivers/filesystems/udfs/Include/wcache_lib.cpp
@@ -1128,10 +1128,11 @@ WCacheUpdatePacket(
                             (PVOID)WCacheSectorAddr(block_array, Lba0),
                             BS) != BS);
             }
-            if(mod)
+            if(mod) {
                 DbgCopyMemory(tmp_buff2 + (i << BSh),
                             (PVOID)WCacheSectorAddr(block_array, Lba0),
                             BS);
+            }
         }
     }
 
--- a/drivers/filesystems/udfs/env_spec.cpp
+++ b/drivers/filesystems/udfs/env_spec.cpp
@@ -307,7 +307,7 @@ UDFPhWriteSynchronous(
     KeQuerySystemTime((PLARGE_INTEGER)&IoEnterTime);
 #endif //MEASURE_IO_PERFORMANCE
 
-#if defined UDF_DBG || defined USE_PERF_PRINT
+#ifdef USE_PERF_PRINT
     ULONG Lba = (ULONG)(Offset>>0xb);
 //    ASSERT(!(Lba & (32-1)));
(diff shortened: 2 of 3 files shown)


drivers/filesystems/udfs/env_spec.cpp
```diff
@@ -308,7 +308,7 @@ UDFPhWriteSynchronous(
 #endif //MEASURE_IO_PERFORMANCE
 
 #if defined UDF_DBG || defined USE_PERF_PRINT
-    ULONG Lba = (ULONG)(Offset>>0xb);
+//    ULONG Lba = (ULONG)(Offset>>0xb);
```
HBelusca, line 311: That's inconsistent, because now what would happen if one re-enables the NDEBUG or whatever? The PerfPrint call below needs that Lba variable.

drivers/filesystems/udfs/Include/wcache_lib.cpp
```diff
@@ -1128,10 +1128,11 @@ WCacheUpdatePacket(
                             (PVOID)WCacheSectorAddr(block_array, Lba0),
                             BS) != BS);
             }
-            if(mod)
+            else {
                 DbgCopyMemory(tmp_buff2 + (i << BSh),
                             (PVOID)WCacheSectorAddr(block_array, Lba0),
                             BS);
+			}
```
HBelusca, line 1135: Indentation problem.

drivers/filesystems/udfs/Include/wcache_lib.cpp
```diff
@@ -1128,10 +1128,11 @@ WCacheUpdatePacket(
                             (PVOID)WCacheSectorAddr(block_array, Lba0),
                             BS) != BS);
             }
-            if(mod)
+            else {
```
HBelusca, line 1131: This change should actually not be necessary (by the principle of minimal changes in 3rd-party code), even if it's true that:
```
if(mod) { /* something */} if(!mod) { /* something else */ }
```
is equivalent to
```
if(mod) { /* something */ } else { /* something else */ }
```

HBelusca (changes requested):

drivers/filesystems/udfs/Include/wcache_lib.cpp
trivalik: Sorry. Probably you are right, that it should be an "if". I just changed this that reactos was building on my PC. I will add just the brackets arround.

HBelusca (approved):

binarymaster:
@trivalik it seems the email you have set up in Git is not associated with your GitHub account. It's highly recommended to associate it.

Related to commit 9d33a205.

PR #4172, opened 2021-12-18
[NETSHELL] Improve workgroup vs. domain wording

cspiegel:
## Purpose

A string in the installer is missing the word "not", and as a result is slightly confusing.  Add the missing word.

### DIFF
--- a/dll/shellext/netshell/lang/bg-BG.rc
+++ b/dll/shellext/netshell/lang/bg-BG.rc
@@ -111,7 +111,7 @@ FONT 8, "MS Shell Dlg"
 BEGIN
     LTEXT "Do you want this computer to be a member of a domain?\n\
 (You can obtain this information from your network administrator.)", IDC_STATIC, 33, 2, 263, 50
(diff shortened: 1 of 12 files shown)


milawynsrealm:
The dialog is asking if the computer will be part of a domain. By answering no, the user is essentially saying that it will be part of a workgroup instead, since it can either be part of a domain, or part of a workstation.

gonzoMD:
> The dialog is asking if the computer will be part of a domain. By answering no, the user is essentially saying that it will be part of a workgroup instead, since it can either be part of a domain, or part of a workgroup.

Like Lee wrote, to make it clear you could add the word 'instead' after 'workgroup' in the RC files.

binarymaster (changes requested):
Computer can be either part of workgroup or in a domain. It can't belong to neither of these. Please review your changes and correct the wording accordingly.

binarymaster (approved):
LGTM now. :+1: 

VirtualBox_ReactOS_21_12_2021_11_12_39

PR #4176, opened 2021-12-19
[WIN32SS][NTDDRAW] Improve DxEngGetDCState type 2 implementation

oleg-dubinskiy:
## Purpose

Add several improvements in `case 2` of `DxEngGetDCState` function in our win32k.
This allows to properly enable and use DirectDraw and Direct3D hardware acceleration with the following MS DirectX components: ddraw.dll, d3d8.dll, d3d9.dll and dxg.sys. As in VM, same on real hardware. :smiley: 
But note that now the bugcheck 0x50 (`PAGE_FAULT_IN_NONPAGED_AREA`) occurs when executing fullscreen DDraw apps, after switching display mode. It's unhidden by this fix. **To be reported.**
Addendum to 855bf46.

## Proposed changes

- Use the `REGION_Complexity()` function to get the actual region complexity instead of just number of rectangles.
- Don't use `prgnRao` at all, because it is always uninitialized when passed to the function. So in condition between `prgnRao` and `prgnVis`, the 2nd one always succeeds.
- Update the comment accordingly.
- Additionally, add a `DPRINT` to the end of the function, which displays the returned value.

## Result

As visible on the following screenshots, DDraw acceleration (and D3D too) works properly after my changes. :slightly_smiling_face: 

Before:
ddraw-before

After:
ddraw-after

The 2 of 3 DirectDraw tests are now working with hardware acceleration, but 3rd (fullscreen) test causes a bugcheck mentioned above after display mode switch. But if to change display settings manually before starting the test, the 3rd one also works properly.
Also a lot of apps/games are now working with hardware acceleration, if they are running in window or in fullscreen, if to switch display mode to required manually before launching the app/game.
**Please note also that in order to see an effect from my changes, you need to apply all my previous ReactX improvements.**

### DIFF
--- a/win32ss/reactx/ntddraw/dxeng.c
+++ b/win32ss/reactx/ntddraw/dxeng.c
@@ -464,8 +464,8 @@ DxEngGetDCState(HDC hDC,
                 retVal = (DWORD_PTR) pDC->fs & DC_FLAG_FULLSCREEN;
                 break;
             case 2:
-                /* Return the number of rectangles in the visible region. */
-                retVal = (DWORD_PTR) pDC->prgnRao ? pDC->prgnRao->rdh.nCount : pDC->prgnVis->rdh.nCount;
+                /* Return the complexity of the visible region. */
+                retVal = (DWORD_PTR) REGION_Complexity(pDC->prgnVis);
                 break;
             case 3:
             {
@@ -480,6 +480,8 @@ DxEngGetDCState(HDC hDC,
         }
         DC_UnlockDc(pDC);
     }
+
+    DPRINT1("Return value %08lx\n", retVal);
 
     return retVal;
 }


win32ss/reactx/ntddraw/dxeng.c
```diff
@@ -465,7 +465,11 @@ DxEngGetDCState(HDC hDC,
                 break;
             case 2:
                 /* Return the number of rectangles in the visible region. */
-                retVal = (DWORD_PTR) pDC->prgnRao ? pDC->prgnRao->rdh.nCount : pDC->prgnVis->rdh.nCount;
+                /* It should not be 1, because dxg!DxDdReenableDirectDrawObject
+                 * does not accept 1. So in case it's equal to 1, don't use it.
+                 * 2 is what typically returned in Windows Server 2003 SP2.
+                 */
+                retVal = (DWORD_PTR) pDC->prgnVis->rdh.nCount != 1 ? pDC->prgnVis->rdh.nCount : 2;
```
HBelusca, line 472: This is not a number of rectangles, but a region complexity.
Try using `REGION_Complexity(pDC->prgnVis);` instead. This will also do the extra NULL check on prgnVis that was missing in your original code (and return an adequate value).
Or `pDC->prgnVis->prgnattr->iComplexity`, but I'm not sure whether this field is always in sync with what `REGION_Complexity` could return.

win32ss/reactx/ntddraw/dxeng.c
```diff
@@ -465,7 +465,11 @@ DxEngGetDCState(HDC hDC,
                 break;
             case 2:
```
HBelusca, line 466: Unrelated to this specific PR, but an idea for thought, it would be nice to have explicit names for this type values.

win32ss/reactx/ntddraw/dxeng.c
oleg-dubinskiy: Thanks @HBelusca. Committed the fix. :+1: Calling `REGION_Complexity()` works for me. And it also returns `2`, so this is indeed more correct fix. That's why it returns `2` in Windows. It's logically obvious. How I didn't understood that earlier... :slightly_smiling_face:

oleg-dubinskiy: I'm also not sure that `pDC->prgnVis->prgnattr->iComplexity` is updated correctly when it is passed to the function.

win32ss/reactx/ntddraw/dxeng.c
```diff
@@ -481,6 +481,8 @@ DxEngGetDCState(HDC hDC,
         DC_UnlockDc(pDC);
     }
 
+    DPRINT1("Return value %08lx\n", retVal);
```
HBelusca, line 484: Is this still useful?

win32ss/reactx/ntddraw/dxeng.c
oleg-dubinskiy: I wonder is it possible at all to see the aliases (names) of these types via debugging analysis somehow... :thinking:

win32ss/reactx/ntddraw/dxeng.c
oleg-dubinskiy: Yes, because the return value from type `2` is still valid, and, besides that, this `DPRINT` displays the values from all 3 types, not only from the 2nd one.

win32ss/reactx/ntddraw/dxeng.c
HBelusca: I don't think so.

HBelusca (approved):

PR #4177, opened 2021-12-19
[KEYBOARD] Switch to QWERTZ for Croatian keyboard layout

Andrej123456789:
## Purpose

Croatian keyboards are QWERTZ.

JIRA issue: unknown

## Proposed changes

Update kbdcr.c
## TODO
Nothing, I builded and tested in virtual machine and works.

### DIFF
--- a/dll/keyboard/kbdcr/kbdcr.c
+++ b/dll/keyboard/kbdcr/kbdcr.c
@@ -41,7 +41,7 @@ ROSDATA USHORT scancode_to_vk[] = {
   /* - 0f - */
   /* First Letters Row */
   VK_TAB,       'Q',          'W',          'E',
-  'R',          'T',          'Y',          'U',
+  'R',          'Z',          'Y',          'U',
   'I',          'O',          'P',
   VK_OEM_4,     VK_OEM_6,     VK_RETURN,
   /* - 1d - */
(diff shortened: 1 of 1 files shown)


dll/keyboard/kbdcr/kbdcr.c
```diff
@@ -41,7 +41,7 @@ ROSDATA USHORT scancode_to_vk[] = {
   /* - 0f - */
   /* First Letters Row */
   VK_TAB,       'Q',          'W',          'E',
-  'R',          'T',          'Y',          'U',
+  'R',          'Z',          'Y',          'U',
```
HBelusca, line 44: Shouldn't it be
```suggestion
  'R',          'T',          'Z',          'U',
```
?

Andrej123456789: Always I did something wrong; yeah I did mistake and I will fix it

HBelusca:
Since you're looking at the keyboard stuff, could you tell me whether the following keys do work:
- `?`, `'` and `¨` (they are near the `0` key in first row) ;
- `*`, `+` and `¸` (they are near the previous key in first row) ;
- `\` (in same key as `Q`) and `|` (in same key as `W`) in second row;
- `@` (in same key as `V`), `{` , `}` and `§` (in same keys as `B`, `N` and `M`) ;
- and the special characters `š`, `đ`, `č`, `ć` and `ž`

Andrej123456789:
Everything works as normal

HBelusca:
Hvala!

PR #4184, opened 2021-12-24
[NTOS:PS] Process Quota Overhaul

GeoB99:
As it currently stands, the implementation of process quotas in the kernel is a stub. We're not checking if a process' quota usage is 0 when it de-references the quota block and we're lacking in code that handles Return, Limit, Peak and Usage fields of a quota block. The Object Manager of ReactOS kernel doesn't make any attempts to charge or return quotas of a given object.

Moreover on Windows the Process Manager cooperates with the Memory Manager if quota limit raising has to occur or not, something which we don't do that. With this PR the process quota support in the kernel is now half-implemented (we currently don't support paging so process page file quota is left behind for now). Kernel consistency is also now enforced as ReactOS kernel checks every time that there's no quota usage when a process exits (the point which the process deletion cleanup method is called).

## Proposed Changes
* Implement the algorithms necessary to handle quota block fields
* Bugcheck the system with `QUOTA_UNDERFLOW` if a component is returning way too much quota to a block or process than it was previously charged
* Perform interlocked atomic operations in a platform architecture specific way (i386 or AMD64)
* Implement `MmRaisePoolQuota` and `MmReturnPoolQuota` routines and let Ps to invoke the Mm if quota raise or return has to happen
* Implement `PsChargeSharedPoolQuota` and `PsReturnSharedPoolQuota` routines, the Object Manager now keeps track of quota charge/return flow of objects in a proper way
* Miscellaneous stuff

## Remarks
As already said above, what's left to address is quota page file which is currently not totally handled for the fact that we barely support paging in any way. In addition to that, on ReactOS we are using `PspDefaultQuotaBlock` for virtually every process in the OS. This is the default block with default quota limits to -1 (which it'd imply no limits whatsoever). 

Winlogon is responsible to set up a different quota block (non-default) for all processes within the interactive session that use it -- this is something that we don't do that in ReactOS. These are two kind of topics that have to be addressed in the future.

**JIRA Issue:** [CORE-17784](https://jira.reactos.org/browse/CORE-17784)

### DIFF
--- a/modules/rostests/kmtests/ntos_ps/PsQuota.c
+++ b/modules/rostests/kmtests/ntos_ps/PsQuota.c
@@ -1,0 +1,133 @@
+/*
+ * PROJECT:     ReactOS kernel-mode tests
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Kernel mode tests for process quotas
+ * COPYRIGHT:   Copyright 2021 George Bișoc <george.bisoc@reactos.org>
+ */
+
+#include <kmt_test.h>
+
+START_TEST(PsQuota)
+{
+    NTSTATUS Status;
+    VM_COUNTERS VmCounters;
+    QUOTA_LIMITS QuotaLimits;
+    SIZE_T NonPagedUsage, PagedUsage;
+    PEPROCESS Process = PsGetCurrentProcess();
+
+    /* Guard the quota operations in a guarded region */
+    KeEnterGuardedRegion();
+
+    /* Report the current process' quota limits */
+    Status = ZwQueryInformationProcess(NtCurrentProcess(),
+                                       ProcessQuotaLimits,
+                                       &QuotaLimits,
+                                       sizeof(QuotaLimits),
+                                       NULL);
+    if (skip(NT_SUCCESS(Status), "Failed to query quota limits -- %lx\n", Status))
+    {
+        return;
+    }
+
+    trace("Process paged pool quota limit -- %lu\n", QuotaLimits.PagedPoolLimit);
+    trace("Process non paged pool quota limit -- %lu\n", QuotaLimits.NonPagedPoolLimit);
+    trace("Process page file quota limit -- %lu\n\n", QuotaLimits.PagefileLimit);
+
+    /* Query the quota usage */
+    Status = ZwQueryInformationProcess(NtCurrentProcess(),
+                                       ProcessVmCounters,
+                                       &VmCounters,
+                                       sizeof(VmCounters),
+                                       NULL);
+    if (skip(NT_SUCCESS(Status), "Failed to query quota usage -- %lx\n", Status))
+    {
+        return;
+    }
+
+    /* Test that quotas usage are within limits */
+    ok(VmCounters.QuotaNonPagedPoolUsage < QuotaLimits.NonPagedPoolLimit, "Non paged quota over limits (usage -> %lu || limit -> %lu)\n",
+       VmCounters.QuotaNonPagedPoolUsage, QuotaLimits.NonPagedPoolLimit);
+    ok(VmCounters.QuotaPagedPoolUsage < QuotaLimits.PagedPoolLimit, "Paged quota over limits (usage -> %lu || limit -> %lu)\n",
+       VmCounters.QuotaPagedPoolUsage, QuotaLimits.PagedPoolLimit);
+
+    /* Cache the quota usage pools for later checks  */
+    NonPagedUsage = VmCounters.QuotaNonPagedPoolUsage;
+    PagedUsage = VmCounters.QuotaPagedPoolUsage;
+
+    /* Charge some paged and non paged quotas */
+    Status = PsChargeProcessNonPagedPoolQuota(Process, 0x200);
+    ok_irql(PASSIVE_LEVEL);
+    ok_eq_hex(Status, STATUS_SUCCESS);
+
+    Status = PsChargeProcessPagedPoolQuota(Process, 0x500);
+    ok_irql(PASSIVE_LEVEL);
+    ok_eq_hex(Status, STATUS_SUCCESS);
+
+    /* Query the quota usage again */
+    Status = ZwQueryInformationProcess(NtCurrentProcess(),
+                                       ProcessVmCounters,
+                                       &VmCounters,
+                                       sizeof(VmCounters),
+                                       NULL);
+    if (skip(NT_SUCCESS(Status), "Failed to query quota usage -- %lx\n", Status))
+    {
+        return;
+    }
+
+
+    /* Test again the usage that's within limits */
+    ok(VmCounters.QuotaNonPagedPoolUsage < QuotaLimits.NonPagedPoolLimit, "Non paged quota over limits (usage -> %lu || limit -> %lu)\n",
+       VmCounters.QuotaNonPagedPoolUsage, QuotaLimits.NonPagedPoolLimit);
+    ok(VmCounters.QuotaPagedPoolUsage < QuotaLimits.PagedPoolLimit, "Paged quota over limits (usage -> %lu || limit -> %lu)\n",
+       VmCounters.QuotaPagedPoolUsage, QuotaLimits.PagedPoolLimit);
+
+    /*
+     * Make sure the results are consistent, that nobody else
+     * is charging quotas other than us.
+     */
+    ok_eq_size(VmCounters.QuotaNonPagedPoolUsage, NonPagedUsage + 0x200);
+    ok_eq_size(VmCounters.QuotaPagedPoolUsage, PagedUsage + 0x500);
+
+    /* Report the quota usage */
+    trace("=== QUOTA USAGE AFTER CHARGE ===\n\n");
+    trace("Process paged pool quota usage -- %lu\n", VmCounters.QuotaPagedPoolUsage);
+    trace("Process paged pool quota peak -- %lu\n", VmCounters.QuotaPeakPagedPoolUsage);
+    trace("Process non paged pool quota usage -- %lu\n", VmCounters.QuotaNonPagedPoolUsage);
+    trace("Process non paged pool quota peak -- %lu\n", VmCounters.QuotaPeakNonPagedPoolUsage);
+    trace("Process page file quota usage -- %lu\n", VmCounters.PagefileUsage);
+    trace("Process page file quota peak -- %lu\n\n", VmCounters.PeakPagefileUsage);
+
+    /* Return the quotas we've charged up */
+    PsReturnProcessNonPagedPoolQuota(Process, 0x200);
+    PsReturnProcessPagedPoolQuota(Process, 0x500);
+
+    /* Query the quota usage again */
+    Status = ZwQueryInformationProcess(NtCurrentProcess(),
+                                       ProcessVmCounters,
+                                       &VmCounters,
+                                       sizeof(VmCounters),
+                                       NULL);
+    if (skip(NT_SUCCESS(Status), "Failed to query quota usage -- %lx\n", Status))
+    {
+        return;
+    }
+
+    /*
+     * Check that nobody else has returned quotas
+     * but only us.
+     */
+    ok_eq_size(VmCounters.QuotaNonPagedPoolUsage, NonPagedUsage);
+    ok_eq_size(VmCounters.QuotaPagedPoolUsage, PagedUsage);
+
+    /* Report the usage again */
+    trace("=== QUOTA USAGE AFTER RETURN ===\n\n");
+    trace("Process paged pool quota usage -- %lu\n", VmCounters.QuotaPagedPoolUsage);
+    trace("Process paged pool quota peak -- %lu\n", VmCounters.QuotaPeakPagedPoolUsage);
+    trace("Process non paged pool quota usage -- %lu\n", VmCounters.QuotaNonPagedPoolUsage);
+    trace("Process non paged pool quota peak -- %lu\n", VmCounters.QuotaPeakNonPagedPoolUsage);
+    trace("Process page file quota usage -- %lu\n", VmCounters.PagefileUsage);
+    trace("Process page file quota peak -- %lu\n\n", VmCounters.PeakPagefileUsage);
+
+    /* We're done, leave the region */
+    KeLeaveGuardedRegion();
+}
--- a/ntoskrnl/ex/handle.c
+++ b/ntoskrnl/ex/handle.c
@@ -102,6 +102,7 @@ ExpAllocateTablePagedPool(IN PEPROCESS Process OPTIONAL,
                           IN SIZE_T Size)
 {
     PVOID Buffer;
+    NTSTATUS Status;
 
     /* Do the allocation */
     Buffer = ExAllocatePoolWithTag(PagedPool, Size, TAG_OBJECT_TABLE);
@@ -113,7 +114,13 @@ ExpAllocateTablePagedPool(IN PEPROCESS Process OPTIONAL,
         /* Check if we have a process to charge quota */
         if (Process)
         {
-            /* FIXME: Charge quota */
+            /* Charge quota */
+            Status = PsChargeProcessPagedPoolQuota(Process, Size);
+            if (!NT_SUCCESS(Status))
+            {
+                ExFreePoolWithTag(Buffer, TAG_OBJECT_TABLE);
+                return NULL;
+            }
         }
     }
 
@@ -127,6 +134,7 @@ ExpAllocateTablePagedPoolNoZero(IN PEPROCESS Process OPTIONAL,
                                 IN SIZE_T Size)
 {
     PVOID Buffer;
+    NTSTATUS Status;
 
     /* Do the allocation */
     Buffer = ExAllocatePoolWithTag(PagedPool, Size, TAG_OBJECT_TABLE);
@@ -135,7 +143,13 @@ ExpAllocateTablePagedPoolNoZero(IN PEPROCESS Process OPTIONAL,
         /* Check if we have a process to charge quota */
         if (Process)
         {
-            /* FIXME: Charge quota */
+            /* Charge quota */
+            Status = PsChargeProcessPagedPoolQuota(Process, Size);
+            if (!NT_SUCCESS(Status))
+            {
+                ExFreePoolWithTag(Buffer, TAG_OBJECT_TABLE);
+                return NULL;
+            }
         }
     }
 
@@ -153,7 +167,8 @@ ExpFreeTablePagedPool(IN PEPROCESS Process OPTIONAL,
     ExFreePoolWithTag(Buffer, TAG_OBJECT_TABLE);
     if (Process)
     {
-        /* FIXME: Release quota */
+        /* Release quota */
+        PsReturnProcessPagedPoolQuota(Process, Size);
     }
 }
 
@@ -248,7 +263,8 @@ ExpFreeHandleTable(IN PHANDLE_TABLE HandleTable)
     ExFreePoolWithTag(HandleTable, TAG_OBJECT_TABLE);
     if (Process)
     {
-        /* FIXME: TODO */
+        /* Release the quota it was taking up */
+        PsReturnProcessPagedPoolQuota(Process, sizeof(HANDLE_TABLE));
     }
 }
 
@@ -312,6 +328,7 @@ ExpAllocateHandleTable(IN PEPROCESS Process OPTIONAL,
     PHANDLE_TABLE HandleTable;
     PHANDLE_TABLE_ENTRY HandleTableTable, HandleEntry;
     ULONG i;
+    NTSTATUS Status;
     PAGED_CODE();
 
     /* Allocate the table */
@@ -323,7 +340,13 @@ ExpAllocateHandleTable(IN PEPROCESS Process OPTIONAL,
     /* Check if we have a process */
     if (Process)
     {
-        /* FIXME: Charge quota */
+        /* Charge quota */
+        Status = PsChargeProcessPagedPoolQuota(Process, sizeof(HANDLE_TABLE));
+        if (!NT_SUCCESS(Status))
+        {
+            ExFreePoolWithTag(HandleTable, TAG_OBJECT_TABLE);
+            return NULL;
+        }
     }
 
     /* Clear the table */
@@ -335,6 +358,13 @@ ExpAllocateHandleTable(IN PEPROCESS Process OPTIONAL,
     {
         /* Failed, free the table */
         ExFreePoolWithTag(HandleTable, TAG_OBJECT_TABLE);
+
+        /* Return the quota it was taking up */
+        if (Process)
+        {
+            PsReturnProcessPagedPoolQuota(Process, sizeof(HANDLE_TABLE));
+        }
+
         return NULL;
     }
 
--- a/ntoskrnl/fsrtl/notify.c
+++ b/ntoskrnl/fsrtl/notify.c
@@ -85,6 +85,7 @@ FsRtlCancelNotify(IN PDEVICE_OBJECT DeviceObject,
     PIO_STACK_LOCATION Stack;
     PNOTIFY_CHANGE NotifyChange;
     PREAL_NOTIFY_SYNC RealNotifySync;
+    BOOLEAN PoolQuotaCharged;
     PSECURITY_SUBJECT_CONTEXT _SEH2_VOLATILE SubjectContext = NULL;
 
     /* Get the NOTIFY_CHANGE struct and reset it */
@@ -165,15 +166,38 @@ FsRtlCancelNotify(IN PDEVICE_OBJECT DeviceObject,
                /* If we have a buffer length, but no buffer then allocate one */
                if (Buffer == NULL)
                {
-                   PsChargePoolQuota(NotifyChange->OwningProcess, PagedPool, BufferLength);
-                   Buffer = ExAllocatePoolWithTag(PagedPool | POOL_RAISE_IF_ALLOCATION_FAILURE, BufferLength, TAG_FS_NOTIFICATIONS);
-                   NotifyChange->AllocatedBuffer = Buffer;
-               }
+                   /* Assume we haven't charged quotas */
+                   PoolQuotaCharged = FALSE;
 
-               /* Copy data in that buffer */
-               RtlCopyMemory(Buffer, NotifyChange->Buffer, NotifyChange->DataLength);
-               NotifyChange->ThisBufferLength = BufferLength;
-               NotifyChange->Buffer = Buffer;
+                   _SEH2_TRY
+                   {
+                       /* Charge quotas */
+                       PsChargePoolQuota(NotifyChange->OwningProcess, PagedPool, BufferLength);
+                       PoolQuotaCharged = TRUE;
+
+                       /* Allocate buffer */
+                       Buffer = ExAllocatePoolWithTag(PagedPool | POOL_RAISE_IF_ALLOCATION_FAILURE, BufferLength, TAG_FS_NOTIFICATIONS);
+                       NotifyChange->AllocatedBuffer = Buffer;
+
+                       /* Copy data in that buffer */
+                       RtlCopyMemory(Buffer, NotifyChange->Buffer, NotifyChange->DataLength);
+                       NotifyChange->ThisBufferLength = BufferLength;
+                       NotifyChange->Buffer = Buffer;
+                   }
+                   _SEH2_EXCEPT(EXCEPTION_EXECUTE_HANDLER)
+                   {
+                       /* Something went wrong, have we charged quotas? */
+                       if (PoolQuotaCharged)
+                       {
+                           /* We did, return quotas */
+                           PsReturnProcessPagedPoolQuota(NotifyChange->OwningProcess, BufferLength);
+                       }
+
+                       /* And notify immediately */
+                       NotifyChange->Flags |= NOTIFY_IMMEDIATELY;
+                   }
+                   _SEH2_END;
+               }
            }
 
            /* If we have to notify immediately, ensure that any buffer is 0-ed out */
@@ -1322,10 +1346,15 @@ FsRtlNotifyFilterReportChange(IN PNOTIFY_SYNC NotifySync,
                         /* If we couldn't find one, then allocate one */
                         if (NotifyChange->Buffer == NULL)
                         {
+                            /* Assign the length buffer */
+                            NotifyChange->ThisBufferLength = NumberOfBytes;
+
+                            /* Assume we have not charged quotas */
                             PoolQuotaCharged = FALSE;
                             _SEH2_TRY
                             {
-                                PsChargePoolQuota(NotifyChange->OwningProcess, PagedPool, NumberOfBytes);
+                                /* And charge quotas */
+                                PsChargePoolQuota(NotifyChange->OwningProcess, PagedPool, NotifyChange->ThisBufferLength);
                                 PoolQuotaCharged = TRUE;
                                 OutputBuffer = ExAllocatePoolWithTag(PagedPool | POOL_RAISE_IF_ALLOCATION_FAILURE,
                                                                      NumberOfBytes, TAG_FS_NOTIFICATIONS);
@@ -1337,7 +1366,7 @@ FsRtlNotifyFilterReportChange(IN PNOTIFY_SYNC NotifySync,
                             {
                                 if (PoolQuotaCharged)
                                 {
-                                    PsReturnProcessPagedPoolQuota(NotifyChange->OwningProcess, NumberOfBytes);
+                                    PsReturnProcessPagedPoolQuota(NotifyChange->OwningProcess, NotifyChange->ThisBufferLength);
                                 }
                                 NotifyChange->Flags |= NOTIFY_IMMEDIATELY;
                             }
--- a/ntoskrnl/include/internal/mm.h
+++ b/ntoskrnl/include/internal/mm.h
@@ -28,6 +28,9 @@ extern KMUTANT MmSystemLoadLock;
 
 extern ULONG MmNumberOfPagingFiles;
 
+extern SIZE_T MmTotalNonPagedPoolQuota;
+extern SIZE_T MmTotalPagedPoolQuota;
+
 extern PVOID MmUnloadedDrivers;
 extern PVOID MmLastUnloadedDrivers;
 extern PVOID MmTriageActionTaken;
@@ -52,6 +55,14 @@ struct _KTRAP_FRAME;
 struct _EPROCESS;
 struct _MM_RMAP_ENTRY;
 typedef ULONG_PTR SWAPENTRY;
+
+//
+// Pool Quota values
+//
+#define MI_QUOTA_NON_PAGED_NEEDED_PAGES             64
+#define MI_NON_PAGED_QUOTA_MIN_RESIDENT_PAGES       200
+#define MI_CHARGE_PAGED_POOL_QUOTA                  0x80000
+#define MI_CHARGE_NON_PAGED_POOL_QUOTA              0x10000
 
 //
 // Special IRQL value (found in assertions)
@@ -646,12 +657,21 @@ MiFreePoolPages(
 
 /* pool.c *******************************************************************/
 
+_Requires_lock_held_(PspQuotaLock)
 BOOLEAN
 NTAPI
-MiRaisePoolQuota(
-    IN POOL_TYPE PoolType,
-    IN ULONG CurrentMaxQuota,
-    OUT PULONG NewMaxQuota
+MmRaisePoolQuota(
+    _In_ POOL_TYPE PoolType,
+    _In_ SIZE_T CurrentMaxQuota,
+    _Out_ PSIZE_T NewMaxQuota
+);
+
+_Requires_lock_held_(PspQuotaLock)
+VOID
+NTAPI
+MmReturnPoolQuota(
+    _In_ POOL_TYPE PoolType,
+    _In_ SIZE_T QuotaToReturn
 );
 
 /* mdl.c *********************************************************************/
--- a/ntoskrnl/include/internal/ob.h
+++ b/ntoskrnl/include/internal/ob.h
@@ -59,6 +59,11 @@
 #define ObpAccessProtectCloseBit 0x02000000L
 
 //
+// System Process Quota Block
+//
+#define OBP_SYSTEM_PROCESS_QUOTA ((PEPROCESS_QUOTA_BLOCK)(ULONG_PTR)1)
+
+//
 // Identifies a Kernel Handle
 //
 #ifdef _WIN64
--- a/ntoskrnl/include/internal/ps.h
+++ b/ntoskrnl/include/internal/ps.h
@@ -71,6 +71,12 @@
 #define PSP_JOB_SCHEDULING_CLASSES              10
 
 //
+// Process Quota Threshold Values
+//
+#define PSP_NON_PAGED_POOL_QUOTA_THRESHOLD              0x10000
+#define PSP_PAGED_POOL_QUOTA_THRESHOLD                  0x80000
+
+//
 // Thread "Set/Get Context" Context Structure
 //
 typedef struct _GET_SET_CTX_CONTEXT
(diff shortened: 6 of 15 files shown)


GeoB99:
**Windows Server 2003**
Capture
**ReactOS**
Capture2

ntoskrnl/ps/quota.c
```diff
@@ -18,123 +17,507 @@ EPROCESS_QUOTA_BLOCK PspDefaultQuotaBlock;
...
+         * the usage and the additional amount to return
+         * to the Memory Manager is that limit minus the
+         * usage.
+         */
+        if (QuotaBlockFromList->ProcessCount == 0)
+        {
+            if (QuotaBlockFromList->QuotaEntry[QuotaType].Usage <
+                QuotaBlockFromList->QuotaEntry[QuotaType].Limit)
+            {
+                InterlockedCompareExchangeSizeT(&QuotaBlockFromList->QuotaEntry[QuotaType].Limit,
```
ThFabba, line 140: Is this interlocked operation just pro forma, or is there actually a chance of concurrent access here?
If the latter, then this seems like it should use fetch-once on `Usage` and `Limit`, and actually use the InterlockedCompareExchange return value. As is, we might be fetching these values 3-4 times, so if they can change in between, that wouldn't be good.

ntoskrnl/ps/quota.c
```diff
@@ -18,123 +17,507 @@ EPROCESS_QUOTA_BLOCK PspDefaultQuotaBlock;
...
     /* Guard our quota in a spin lock */
     KeAcquireSpinLock(&PspQuotaLock, &OldIrql);
 
-    if (Process->QuotaUsage[QuotaType] < Amount)
+    /* Does the caller return way more quota than it was previously charged? */
+    if ((Process && Process->QuotaUsage[QuotaType] < Amount) ||
+        QuotaBlock->QuotaEntry[QuotaType].Usage < Amount)
     {
-        DPRINT1("WARNING: Process->QuotaUsage sanity check failed.\n");
+        /* We're returning way too much quota, kill the system! */
```
ThFabba, line 362: Nit: why "way too much"? It's any amount of too much (including 1), right?

ntoskrnl/ps/quota.c
```diff
@@ -143,45 +526,279 @@ PspInsertQuotaBlock(
...
+        }
+    }
+
+    /*
+     * As object quotas have been charged into the
+     * process' quota block, that block will gonna
+     * be used by the said process, so increment
+     * both process and reference counts.
+     */
+    InterlockedIncrementSizeT(&Process->QuotaBlock->ProcessCount);
```
ThFabba, line 733: Should we really increase `ProcessCount` here? The reference count is needed in case the object outlives the process (we don't want to delete the block while an object still references it). But the process count seems to exist mostly for `PspReturnExcessQuotas`. It would make sense that if only objects, but not processes, are referencing a quota block, then the remaining quota would be considered excess (since the objects won't suddenly request additional quota).

sdk/include/xdk/interlocked.h
```diff
@@ -173,6 +173,12 @@
 /* No release here */
 #define InterlockedExchangeNoFence64 __NF_(_InterlockedExchange64)
 
+#ifdef _WIN64
+#define InterlockedExchangeSizeT(a, b) InterlockedExchange64((LONG64 *)a, b)
```
ThFabba, line 177: Microsoft's headers doesn't seem to have this one, so we shouldn't add it either. Could just use `InterlockedExchangeAddSizeT` instead I suppose.

GeoB99: This macro is just to exchange between two values and return that , not to perform an addition operation -- which is used in `PspReturnExcessQuotas`. Unless you came up with an idea of using `InterlockedExchangeAddSizeT` but in a tricky way for that or what exactly?

ntoskrnl/ps/quota.c
```diff
@@ -18,123 +17,507 @@ EPROCESS_QUOTA_BLOCK PspDefaultQuotaBlock;
...
+    _In_ PEPROCESS_QUOTA_BLOCK QuotaBlock)
+{
+    ULONG PsQuotaTypeIndex;
+    SIZE_T QuotaToReturn;
+
+    /* Make sure that the quota block is not plain garbage */
+    ASSERT(QuotaBlock);
+
+    /* Loop over the Process quota types */
+    for (PsQuotaTypeIndex = PsQuotaTypes; PsQuotaTypeIndex <  PsQuotaTypes; PsQuotaTypeIndex++)
```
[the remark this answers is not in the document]
GeoB99: Roger that.

ntoskrnl/ps/quota.c
```diff
@@ -18,123 +17,507 @@ EPROCESS_QUOTA_BLOCK PspDefaultQuotaBlock;
...
+    {
+        /* The amount needed to return to Mm is the limit and return fields */
+        QuotaToReturn = QuotaBlock->QuotaEntry[PsQuotaTypeIndex].Limit + QuotaBlock->QuotaEntry[PsQuotaTypeIndex].Return;
+        MmReturnPoolQuota(PsQuotaTypeIndex, QuotaToReturn);
+
+        /*
+         * If that's a page file quota then don't
+         * do anything here, so just skip it outright.
```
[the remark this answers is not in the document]
GeoB99: This has to do with the fact that my knowledge of using correct English words is kind of poor, as on my thought and observation "skip it outright" implies "avoid that quota type and terminate the loop". Which I admit, I need to work myself on that.

ntoskrnl/ps/quota.c
GeoB99: The interlocked operation there is just pro forma, as race concurrences cannot happen here since `PspReturnExcessQuotas` is used in `PspChargeProcessQuotaSpecifiedPool` with the underlying code wrapped and guarded with a spin lock. No other threads can perform the said interlocked operation at the same time.

ntoskrnl/include/internal/mm.h
```diff
@@ -53,6 +56,14 @@ struct _EPROCESS;
 struct _MM_RMAP_ENTRY;
 typedef ULONG_PTR SWAPENTRY;
 
+//
+// Pool Quota values
+//
+#define MI_QUOTA_NON_PAGED_NEEDED_PAGES 64
+#define MI_NON_PAGED_QUOTA_MIN_RESIDENT_PAGES 200
+#define MI_CHARGE_PAGED_POOL_QUOTA 0x80000
+#define MI_CHARGE_NON_PAGED_POOL_QUOTA 0x10000
```
HBelusca, line 65: Code style seems to promote indenting the values.

ntoskrnl/include/internal/ob.h
```diff
@@ -58,6 +58,11 @@
 //
 #define ObpAccessProtectCloseBit 0x02000000L
 
+//
+// System Process Quota Block
+//
+#define OBJ_SYSTEM_PROCESS_QUOTA (PEPROCESS_QUOTA_BLOCK)1
```
HBelusca, line 64:
```suggestion
#define OBJ_SYSTEM_PROCESS_QUOTA    ((PEPROCESS_QUOTA_BLOCK)1)
```
The "1" value might also require an extra `(ULONG_PTR)` cast (before the pointer one)...

ntoskrnl/include/internal/ps.h
```diff
@@ -70,6 +70,17 @@
 //
 #define PSP_JOB_SCHEDULING_CLASSES              10
 
+//
+// Process Quota Threshold Values
+//
+#define PSP_NON_PAGED_POOL_QUOTA_THRESHOLD 0x10000
+#define PSP_PAGED_POOL_QUOTA_THRESHOLD 0x80000
```
HBelusca, line 77: Ditto for the indentation of the values.

ntoskrnl/mm/ARM3/pool.c
```diff
@@ -1388,6 +1375,157 @@ MiInitializeSessionPool(VOID)
...
+            {
+                DPRINT1("MmRaisePoolQuota(): Not enough pages available (current pages -- %lu)\n", MmAvailablePages);
+                return FALSE;
+            }
+
+            /*
+             * Check if there's at least some space available
+             * in the non paged pool area.
+             */
+            if (MmMaximumNonPagedPoolInPages < MmAllocatedNonPagedPool >> PAGE_SHIFT)
```
HBelusca, line 1426: I would suggest adding extra parentheses around which operator applies first.

ntoskrnl/mm/ARM3/pool.c
```diff
@@ -1388,6 +1375,157 @@ MiInitializeSessionPool(VOID)
...
+MmReturnPoolQuota(
+    _In_ POOL_TYPE PoolType,
+    _In_ SIZE_T QuotaToReturn)
+{
+    switch (PoolType)
+    {
+        case NonPagedPool:
+        {
+            /* This is a non paged pool type, decrease the non paged quota */
+            MmTotalNonPagedPoolQuota -= QuotaToReturn;
```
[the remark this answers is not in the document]
HBelusca: (same suggestion would apply as well for the next case too.)

ntoskrnl/ob/obhandle.c
```diff
@@ -459,14 +459,19 @@ ObpChargeQuotaForObject(IN POBJECT_HEADER ObjectHeader,
...
 
-        /* Charge the quota */
-        ObjectHeader->QuotaBlockCharged = (PVOID)1;
-        DPRINT("FIXME: Should charge: %lx %lx\n", PagedPoolCharge, NonPagedPoolCharge);
-#if 0
-            PsChargeSharedPoolQuota(PsGetCurrentProcess(),
-                                    PagedPoolCharge,
-                                    NonPagedPoolCharge);
-#endif
+        /* Is this a system process? */
```
HBelusca, line 462: There's only one single initial system process:
```suggestion
        /* Is this the system process? */
```
(suggestion:) You could group all comments together into a "summary" comment for conciseness.

ntoskrnl/ps/quota.c
```diff
@@ -18,123 +17,507 @@ EPROCESS_QUOTA_BLOCK PspDefaultQuotaBlock;
...
+ * paged pool quotas. Page file quota type is
+ * not permitted.
+ */
+VOID
+NTAPI
+PspReturnExcessQuotas(
+    _In_ PS_QUOTA_TYPE QuotaType,
+    _Outptr_ PSIZE_T ReturnedQuotas)
+{
+    PPSP_QUOTA_LIST PspQuotaList;
```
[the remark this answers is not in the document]
HBelusca: Ah, so that's the place where `PPSP_QUOTA_LIST` type is used. My suggestion: either move that typedef into that quota.c file, or just use here a `PLIST_ENTRY` straightforwardly (the latter seems to be the best, since all the usage done with it below clearly shows you're using such LIST_ENTRY).

sdk/include/xdk/interlocked.h
ThFabba: Yep, the tricky way is to add 0. `InterlockedExchangeAdd(&var, 0)` is the same as `InterlockedExchange(&var)`

ntoskrnl/ps/quota.c
```diff
@@ -18,123 +17,507 @@ EPROCESS_QUOTA_BLOCK PspDefaultQuotaBlock;
...
     {
-        DPRINT1("Quota exceeded, but ROS will let it slide...\n");
-        KeReleaseSpinLock(&PspQuotaLock, OldIrql);
-        return STATUS_SUCCESS;
-        //return STATUS_QUOTA_EXCEEDED; /* caller raises the exception */
+        /* We aren't... Is this a page file quota charging? */
+        if (QuotaType == PsPageFile)
+        {
+            /* It is, return the appropriate status code */
+            DPRINT1("PspChargeProcessQuotaSpecifiedPool(): Quota amount exceeds the limit on page file quota (the limit -- %lu || the given amount -- %lu)\n",
```
HBelusca, line 214: Here and in other similar comments you can shorten slightly the dprint by omitting "the" in `(the limit -- ` ... `the given amount`

ntoskrnl/ps/quota.c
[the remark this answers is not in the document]
ThFabba: But you're not avoiding that quota type at all, since you're `break`ing _after_ calling `MmReturnPoolQuota`

ntoskrnl/ps/quota.c
```diff
@@ -18,123 +17,507 @@ EPROCESS_QUOTA_BLOCK PspDefaultQuotaBlock;
...
+    if ((Process && Process->QuotaUsage[QuotaType] < Amount) ||
+        QuotaBlock->QuotaEntry[QuotaType].Usage < Amount)
     {
-        DPRINT1("WARNING: Process->QuotaUsage sanity check failed.\n");
+        /* We're returning way too much quota, kill the system! */
+        KeBugCheckEx(QUOTA_UNDERFLOW,
+                     (ULONG_PTR)Process,
+                     (ULONG_PTR)QuotaType,
+                     Process ? (ULONG_PTR)Process->QuotaUsage[QuotaType] :
+                     QuotaBlock->QuotaEntry[QuotaType].Usage,
```
HBelusca, line 367: May be worth extra indenting to visually see that it's not a separate parameter of the KeBugCheckEx:
```suggestion
                               QuotaBlock->QuotaEntry[QuotaType].Usage,
```

ntoskrnl/ps/quota.c
```diff
@@ -18,123 +17,507 @@ EPROCESS_QUOTA_BLOCK PspDefaultQuotaBlock;
...
+        if (Amount > ReturnThreshold)
+        {
+            /* It does, the new amount is the default */
+            AmountToReturn = ReturnThreshold;
+        }
+        else
+        {
+            /* Otherwise the amount is just the given one by the caller */
+            AmountToReturn = Amount;
+        }
```
HBelusca, line 391: Looks like a more concise:
```suggestion
        /* If the amount to return exceeds the threshold, the new amount becomes
         * the default, otherwise the amount is just the one given by the caller. */
        AmountToReturn = min(Amount, ReturnThreshold);
```
By the way, the part of the comment that says "the new amount is the default" might somehow need a fix, as the amount being returned is actually the threshold.

ntoskrnl/ps/quota.c
```diff
@@ -18,123 +17,507 @@ EPROCESS_QUOTA_BLOCK PspDefaultQuotaBlock;
...
+    /* Assign the default quota limits */
     PspDefaultQuotaBlock.QuotaEntry[PsNonPagedPool].Limit = (SIZE_T)-1;
     PspDefaultQuotaBlock.QuotaEntry[PsPagedPool].Limit = (SIZE_T)-1;
     PspDefaultQuotaBlock.QuotaEntry[PsPageFile].Limit = (SIZE_T)-1;
+
+    /*
+     * Set up the count references as the
+     * default block will gonna be used.
+     */
+    PspDefaultQuotaBlock.ReferenceCount = 1;
```
HBelusca, line 452: Is the value "zero" meant to say the block has to go away?

ntoskrnl/ps/quota.c
```diff
@@ -18,123 +17,507 @@ EPROCESS_QUOTA_BLOCK PspDefaultQuotaBlock;
...
+ * Nothing.
+ */
 VOID
 NTAPI
-PspInheritQuota(PEPROCESS Process, PEPROCESS ParentProcess)
+PspInheritQuota(
+    _In_ PEPROCESS Process,
+    _In_opt_ PEPROCESS ParentProcess)
 {
+    PEPROCESS_QUOTA_BLOCK ParentQuotaBlock, DefaultQuotaBlock;
```
HBelusca, line 482: It doesn't appear necessary to have two separate variables with single-usage use, so to speak.
The original `PEPROCESS_QUOTA_BLOCK QuotaBlock` was fine to be used in the two independent `if` cases.

ntoskrnl/ps/quota.c
```diff
@@ -143,45 +526,279 @@ PspInsertQuotaBlock(
...
+ * Nothing.
+ */
+VOID
+NTAPI
+PspDereferenceProcessQuotaBlock(
+    _In_ PEPROCESS Process)
+{
+    ULONG PsQuotaTypeIndex;
+
+    /* Iterate over the process quota types */
```
HBelusca, line 551: This seems to be just a debug-only sanity-check loop. (Note that the assert wouldn't do anything in release mode, and the for-loop would be therefore useless.)

ntoskrnl/ps/quota.c
```diff
@@ -143,45 +526,279 @@ PspInsertQuotaBlock(
...
+    /* Acquire the quota lock */
+    KeAcquireSpinLock(&PspQuotaLock, &OldIrql);
+
+    /* Return all the quotas back to Mm and remove the quota from list */
+    PspReturnQuotasOnDestroy(QuotaBlock);
+    RemoveEntryList(&QuotaBlock->QuotaList);
+
+    /* Release the lock and free the block */
+    KeReleaseSpinLock(&PspQuotaLock, OldIrql);
+    ExFreePool(QuotaBlock);
```
HBelusca, line 602: Not `ExFreePoolWithTag` ? Is the quota block tag always known or not?

ntoskrnl/ps/quota.c
HBelusca, line 52:
```suggestion
    for (PsQuotaTypeIndex = 0; PsQuotaTypeIndex < PsQuotaTypes; PsQuotaTypeIndex++)
```

ntoskrnl/ps/quota.c
GeoB99: If `ReferenceCount == 0` then yes, the block is no longer used by anybody and it has to go away. Here the default quota block gets initialized upon boot phase of the system, gets assigned to the initial process yet we're not starting the counters which feels wrong.

ntoskrnl/ps/quota.c
```diff
@@ -18,123 +17,499 @@ EPROCESS_QUOTA_BLOCK PspDefaultQuotaBlock;
...
 
         Process->QuotaBlock = QuotaBlock;
     }
     else
+    {
         Process->QuotaBlock = &PspDefaultQuotaBlock;
+
+        InterlockedIncrementSizeT(&Process->QuotaBlock->ProcessCount);
+        InterlockedIncrementSizeT(&Process->QuotaBlock->ReferenceCount);
+    }
```
HBelusca, line 495: This function could avoid duplicates as follows:
```c
{
    PEPROCESS_QUOTA_BLOCK QuotaBlock;

    if (ParentProcess != NULL)
    {
        QuotaBlock = ParentProcess->QuotaBlock;
        ASSERT(QuotaBlock != NULL);
    }
    else
    {
        QuotaBlock = &PspDefaultQuotaBlock;
    }

    InterlockedIncrementSizeT(&QuotaBlock->ProcessCount);
    InterlockedIncrementSizeT(&QuotaBlock->ReferenceCount);

    Process->QuotaBlock = QuotaBlock;
}
```

ntoskrnl/mm/ARM3/pool.c
```diff
@@ -1388,6 +1375,158 @@ MiInitializeSessionPool(VOID)
...
+            return TRUE;
+        }
+
+        case PagedPool:
+        {
+            /*
+             * Before raising the quota limit of a paged quota
+             * pool, make sure we've got enough space that is available.
+             * On Windows it seems it wants to check for at least 1 MB of space
+             * needed so that it'd be possible to raise the paged pool quota.
```
HBelusca, line 1457:
```suggestion
             * needed so that it would be possible to raise the paged pool quota.
```

HBelusca (commented):
Final suggestion (for this PR or later): add SAL annotations and/or extra assertions in the private PS quota functions that verify that the quota lock is correctly held.

ntoskrnl/fsrtl/notify.c
```diff
@@ -85,6 +85,7 @@ FsRtlCancelNotify(IN PDEVICE_OBJECT DeviceObject,
     PIO_STACK_LOCATION Stack;
     PNOTIFY_CHANGE NotifyChange;
     PREAL_NOTIFY_SYNC RealNotifySync;
+    BOOLEAN PoolQuotaCharged;
```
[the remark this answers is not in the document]
ThFabba: No, it does not. In a struct, it would add padding, unless you pack it. On the stack, the compiler is free to rearrange variables any way it likes, but it will ensure natural alignment.

ntoskrnl/ex/handle.c
```diff
@@ -323,7 +340,13 @@ ExpAllocateHandleTable(IN PEPROCESS Process OPTIONAL,
     /* Check if we have a process */
     if (Process)
     {
-        /* FIXME: Charge quota */
+        /* Charge quota */
+        Status = PsChargeProcessPagedPoolQuota(Process, sizeof(HANDLE_TABLE));
```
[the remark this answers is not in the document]
ThFabba: `ExAllocatePoolWithQuotaTag` will reference the process. I think the implication is that the handle table (and with it, change notifications) get cleaned up on process exit -- so if they kept references to the process, that would prevent the process from getting destroyed.

ntoskrnl/ps/quota.c
```diff
@@ -18,123 +17,499 @@ EPROCESS_QUOTA_BLOCK PspDefaultQuotaBlock;
...
     KIRQL OldIrql;
-    ASSERT(Process);
-    ASSERT(Process != PsInitialSystemProcess);
+    SIZE_T ReturnThreshold;
+    SIZE_T AmountToReturn = 0;
+    SIZE_T UpdatedQuotaUsage = 0;
+
+    /* Sanity checks */
     ASSERT(QuotaType < PsQuotaTypes);
     ASSERT(!(Amount & 0x80000000)); /* we need to be able to negate it */
```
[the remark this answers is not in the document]
ThFabba: `(SSIZE_T)Amount >= 0` should do the trick without playing bit games.

ntoskrnl/ps/quota.c
```diff
@@ -18,123 +17,499 @@ EPROCESS_QUOTA_BLOCK PspDefaultQuotaBlock;
...
+                     (ULONG_PTR)Amount);
     }
-    else
+
+    /*
+     * We need to trim the quota limits based on the
+     * amount we're going to return quotas back.
+     */
+    if (QuotaType != PsPageFile &&
+        QuotaBlock != &PspDefaultQuotaBlock)
```
ThFabba, line 376: Can we really get here with `PspDefaultQuotaBlock`? That block never actually gets anything charged against it IIUC, so I would imagine it would hit the bugcheck above, since `Usage` will always be zero

ntoskrnl/fsrtl/notify.c
```diff
@@ -1337,7 +1365,7 @@ FsRtlNotifyFilterReportChange(IN PNOTIFY_SYNC NotifySync,
                             {
                                 if (PoolQuotaCharged)
                                 {
-                                    PsReturnProcessPagedPoolQuota(NotifyChange->OwningProcess, NumberOfBytes);
+                                    PsReturnProcessPagedPoolQuota(NotifyChange->OwningProcess, NotifyChange->ThisBufferLength);
```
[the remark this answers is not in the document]
GeoB99: I might just as well assign `ThisBufferLength` outside of SEH block and be done with it.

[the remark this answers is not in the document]
HBelusca: Assigning `NotifyChange->ThisBufferLength` ? Shouldn't this be protected by SEH due to `NotifyChange` ?

ntoskrnl/ps/quota.c
```diff
@@ -18,123 +17,499 @@ EPROCESS_QUOTA_BLOCK PspDefaultQuotaBlock;
...
+    SIZE_T QuotaToReturn;
+
+    /* Make sure that the quota block is not plain garbage */
+    ASSERT(QuotaBlock);
+
+    /* Loop over the Process quota types */
+    for (PsQuotaTypeIndex = PsNonPagedPool; PsQuotaTypeIndex < PsQuotaTypes; PsQuotaTypeIndex++)
+    {
+        /* Terminate the loop if we reach here */
+        if (PsQuotaTypeIndex == PsPageFile)
```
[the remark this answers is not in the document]
GeoB99: `MmReturnPoolQuota` shouldn't be used against pagefile as for whatever reason Windows seems to handle this stuff quite differently. I'll write a comment regarding this.

ntoskrnl/fsrtl/notify.c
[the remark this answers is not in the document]
ThFabba: `NotifyChange` is a pool structure, so accessing it doesn't require SEH.
I think the main reason for SEH here is that the quota function raises? I don't see any user memory being accessed.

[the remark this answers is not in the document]
GeoB99: Thomas nailed the head, `PsChargePoolQuota` raises an exception if the caller is charging quotas that exceeds the limit. Any kernel components, FSDs, whatever, are responsible on their own to handle such exceptions.

ntoskrnl/ps/quota.c
GeoB99: I'm not sure what's exactly your point of this comment. This branch is about Return and Limit fields management of a quota block as one mustn't touch the default block's limits (otherwise it would not make sense logically to call `PspDefaultQuotaBlock` a default block with default limits if you touch them at this point). `Usage` field is handled outside of this branch.

ntoskrnl/ps/quota.c
```diff
@@ -18,123 +17,499 @@ EPROCESS_QUOTA_BLOCK PspDefaultQuotaBlock;
...
+    {
+        /* Gather that list */
+        QuotaBlockFromList = CONTAINING_RECORD(PspQuotaList, EPROCESS_QUOTA_BLOCK, QuotaList);
+
+        /*
+         * Gather any unreturned quotas and cache
+         * them to a variable. If there's nothing
+         * to return back then there's nothing to do...
+         */
+        AmountToReturn += InterlockedExchangeAddSizeT(&QuotaBlockFromList->QuotaEntry[QuotaType].Return, 0);
```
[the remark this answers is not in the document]
GeoB99: That would force me to write a macro like `InterlockedExchangeSizeT` or similar since there are `InterlockedExchange` and `InterlockedExchange64` written separately in **interlocked.h** header. Plus this trick of doing `InterlockedAdd(x, 0)` was originally suggested by you before.

ntoskrnl/ps/quota.c
ThFabba: I'm suggesting that this point in the code may be unreachable if `QuotaBlock == &PspDefaultQuotaBlock`, because either
* this function is never called with that quota block, or
* if it was called this way, it would likely have bugchecked above.

I may be making a wrong assumption somewhere, but that's my reading of the code.

ntoskrnl/ps/quota.c
```diff
@@ -18,123 +17,499 @@ EPROCESS_QUOTA_BLOCK PspDefaultQuotaBlock;
...
+                                    QuotaBlock->QuotaEntry[QuotaType].Usage + Amount,
+                                    QuotaBlock->QuotaEntry[QuotaType].Usage);
 
-    if (Process->QuotaPeak[QuotaType] < Process->QuotaUsage[QuotaType])
+    /* Are we being given a process as well? */
+    if (Process)
     {
-        Process->QuotaPeak[QuotaType] = Process->QuotaUsage[QuotaType];
+        /* We're being given, check that's not a system one  */
+        ASSERT(Process != PsInitialSystemProcess);
```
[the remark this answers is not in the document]
GeoB99: I have placed this ASSERT here so that we avoid another `if (Process)` branch block which in my opinion it's pretty much excessive at this point.

ntoskrnl/ps/quota.c
```diff
@@ -18,123 +17,483 @@ EPROCESS_QUOTA_BLOCK PspDefaultQuotaBlock;
...
+        {
+            DPRINT1("PspChargeProcessQuotaSpecifiedPool(): Quota amount exceeds the limit (limit -- %lu || amount -- %lu)\n",
+                    QuotaBlock->QuotaEntry[QuotaType].Limit, Amount);
+            return STATUS_QUOTA_EXCEEDED;
+        }
     }
 
-    InterlockedExchangeAdd((LONG*)&Process->QuotaUsage[QuotaType], Amount);
+    /* Update the quota usage */
+    InterlockedCompareExchangeSizeT(&QuotaBlock->QuotaEntry[QuotaType].Usage,
```
ThFabba, line 268: As a rule, `InterlockedCompareExchange` isn't really useful if you're not looking at its return value.
I think the logic here needs to look something like
```
while (1)
{
    Usage = QuotaEntry->Usage; // fetch once
    Limit = QuotaEntry->Limit;

    if (Usage + Amount > Limit)
    {
        // fail or acquire more quota
    }
    if (InterlockedCompareExchange(&QuotaEntry->Usage, Usage + Amount, Usage) == Usage)
    {
        // CompareExchange succeeded
        break;
    }
    // somebody updated Usage while we were doing our calculations. Start over
}
```

ntoskrnl/ps/quota.c
```diff
@@ -18,123 +17,483 @@ EPROCESS_QUOTA_BLOCK PspDefaultQuotaBlock;
...
+        }
+
+        /* Release the lock */
         KeReleaseSpinLock(&PspQuotaLock, OldIrql);
-        return STATUS_SUCCESS;
-        //return STATUS_QUOTA_EXCEEDED; /* caller raises the exception */
+
+        /* Enforce a new raised limit */
+        FinalLimit = QuotaBlock->QuotaEntry[QuotaType].Limit - UpdatedLimit;
+        InterlockedCompareExchangeSizeT(&QuotaBlock->QuotaEntry[QuotaType].Limit,
```
ThFabba, line 247: This is inconsistent with the re-fetching of `Limit` done above. I think `Limit` needs to follow one of two rules (it seems to be the first on Windows`*`, but I think the second would work fine):

Either, `Limit` updates need to always happen under lock. In this case, this CompareExchange should move before the `KeReleaseSpinLock`, but is basically safe. It could then just be an InterlockedExchange or an InterlockedExchangeAdd, I believe.

Or, `Limit` updates can happen outside the lock as long as they are interlocked. In this case:
* The code above needs to use a fetch-once approach instead of re-reading `Limit`
* This CompareExchange needs to use the once-fetched value
* There needs to be a way to recover from CompareExchange failure scenario (e.g. with a loop; or simply by not using CompareExchange. You could calculate `AddedLimit = UpdatedLimit - OldLimit` here and just use an `Interlocked(Exchange)Add`)


`* I'm actually not sure -- they seem to both hold the lock everywhere they update Limit, and use safe interlocked operations, which means I may be missing some places where Limit is updated without the lock held`

modules/rostests/kmtests/ntos_ps/PsQuota.c
```diff
@@ -0,0 +1,89 @@
...
+    ok_irql(PASSIVE_LEVEL);
+    ok_eq_hex(Status, STATUS_SUCCESS);
+
+    Status = PsChargeProcessPagedPoolQuota(Process, 0x200);
+    ok_irql(PASSIVE_LEVEL);
+    ok_eq_hex(Status, STATUS_SUCCESS);
+
+    /* Report the quota usage */
+    trace("=== QUOTA USAGE AFTER CHARGE ===\n\n");
+    QueryProcessQuotaUsage();
```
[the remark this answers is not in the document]
GeoB99: Like you mean caching the quota values in local variables and then retry recharging/returning quotas again and check if the values have remained the same (with `ok()`)? Because on Windows the results are different for obvious reasons.

ntoskrnl/ps/quota.c
```diff
@@ -18,123 +17,483 @@ EPROCESS_QUOTA_BLOCK PspDefaultQuotaBlock;
...
+        /*
+         * OK, we've now updated the quota usage of the process
+         * based upon the amount that the caller wanted to charge.
+         * Although the peak of process quota can be less than it was
+         * before so update the peaks as well accordingly.
+         */
+        if (Process->QuotaPeak[QuotaType] < Process->QuotaUsage[QuotaType])
+        {
+            Process->QuotaPeak[QuotaType] = Process->QuotaUsage[QuotaType];
+            Process->QuotaBlock->QuotaEntry[QuotaType].Peak = Process->QuotaBlock->QuotaEntry[QuotaType].Usage;
```
ThFabba, line 289: Shouldn't this be its own condition? Just because one process hits the peak of its usage doesn't mean the whole quota block is at peak.

ThFabba, line 289: I think these updates need to be interlocked as well, since the value can change between the < check and the write

modules/rostests/kmtests/ntos_ps/PsQuota.c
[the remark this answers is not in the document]
ThFabba: Yeah, I would expect that
```
ZwQuery();
NonPagedUsage = VmCounters.QuotaNonPagedPoolUsage;

PsChargeProcessNonPagedPoolQuota(Process, 0x200);

ZwQuery();
ok(VmCounters.QuotaNonPagedPoolUsage == NonPagedUsage + 0x200);

PsReturnProcessNonPagedPoolQuota(Process, 0x200);

ZwQuery();
ok(VmCounters.QuotaNonPagedPoolUsage == NonPagedUsage);
```
would work, as long as no other code is running in your process (which it shouldn't, if you have APCs disabled)

ntoskrnl/ps/quota.c
GeoB99: You're correct, perhaps a combination of `max` and `InterlockedExchangeSizeT` should do the job neatly.

ntoskrnl/ps/quota.c
```diff
@@ -18,123 +17,483 @@ EPROCESS_QUOTA_BLOCK PspDefaultQuotaBlock;
...
+         * the usage and the additional amount to return
+         * to the Memory Manager is that limit minus the
+         * usage.
+         */
+        if (QuotaBlockFromList->ProcessCount == 0)
+        {
+            if (QuotaBlockFromList->QuotaEntry[QuotaType].Usage <
+                QuotaBlockFromList->QuotaEntry[QuotaType].Limit)
+            {
+                InterlockedCompareExchangeSizeT(&QuotaBlockFromList->QuotaEntry[QuotaType].Limit,
```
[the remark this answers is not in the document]
GeoB99: I think I have already said it before that concurrency cannot happen as `PspReturnExcessQuotas` is called right in the path code of `PspChargeProcessQuotaSpecifiedPool` where a spin lock is held. I will just turn this interlocked operation into a mere `InterlockedExchangeSizeT`.

modules/rostests/kmtests/ntos_ps/PsQuota.c
```diff
@@ -0,0 +1,120 @@
...
+    /* Cache the quota usage pools for later checks  */
+    NonPagedUsage = VmCounters.QuotaNonPagedPoolUsage;
+    PagedUsage = VmCounters.QuotaPagedPoolUsage;
+
+    /* Charge some paged and non paged quotas */
+    Status = PsChargeProcessNonPagedPoolQuota(Process, 0x200);
+    ok_irql(PASSIVE_LEVEL);
+    ok_eq_hex(Status, STATUS_SUCCESS);
+
+    Status = PsChargeProcessPagedPoolQuota(Process, 0x200);
```
ThFabba, line 56: Nit: I would use a different number for each pool type. That way the test would detect the case where our code accidentally swaps them.

modules/rostests/kmtests/ntos_ps/PsQuota.c
```diff
@@ -0,0 +1,120 @@
...
+                                       sizeof(QuotaLimits),
+                                       NULL);
+    if (skip(NT_SUCCESS(Status), "Failed to query quota limits -- %lx\n", Status))
+    {
+        return;
+    }
+
+    trace("Process paged pool quota limit -- %lu\n", QuotaLimits.PagedPoolLimit);
+    trace("Process non paged pool quota limit -- %lu\n", QuotaLimits.NonPagedPoolLimit);
+    trace("Process page file quota limit -- %lu\n\n", QuotaLimits.PagefileLimit);
```
ThFabba, line 34: You could also have the test check that `Usage < Limit`.
`Usage > 0` probably also makes sense.

ntoskrnl/ps/quota.c
```diff
@@ -18,123 +17,500 @@ EPROCESS_QUOTA_BLOCK PspDefaultQuotaBlock;
...
+
+        /*
+         * If no other process is taking use of this
+         * block, then we have to check that the limit
+         * is way over the usage. If that's the case,
+         * then we'll shrink the limit up on par with
+         * the usage and the additional amount to return
+         * to the Memory Manager is that limit minus the
+         * usage.
+         */
```
ThFabba, line 140: Nit: this comment explains "what" right now. Perhaps add the "why": if ProcessCount=0, that means this block only has "shared" pool quota, which cannot increase (because there are no processes left to take it). Therefore the remaining quota isn't going to be used by anyone and we can safely take it.

ntoskrnl/ps/quota.c
```diff
@@ -18,123 +17,500 @@ EPROCESS_QUOTA_BLOCK PspDefaultQuotaBlock;
...
+        if (QuotaBlockFromList->ProcessCount == 0)
+        {
+            if (QuotaBlockFromList->QuotaEntry[QuotaType].Usage <
+                QuotaBlockFromList->QuotaEntry[QuotaType].Limit)
+            {
+                InterlockedExchangeSizeT(&QuotaBlockFromList->QuotaEntry[QuotaType].Limit,
+                                         QuotaBlockFromList->QuotaEntry[QuotaType].Usage);
+                AmountToReturn += QuotaBlockFromList->QuotaEntry[QuotaType].Limit - QuotaBlockFromList->QuotaEntry[QuotaType].Usage;
```
ThFabba, line 148: You just set `Limit` equal to `Usage` above, so `Limit - Usage` is zero.

ntoskrnl/ps/quota.c
```diff
@@ -18,123 +17,500 @@ EPROCESS_QUOTA_BLOCK PspDefaultQuotaBlock;
...
+
+            /* Try to raise the quota limits again */
+            MmRaisePoolQuota(QuotaType,
+                             QuotaBlock->QuotaEntry[QuotaType].Limit,
+                             &UpdatedLimit);
+        }
+
+        /* Enforce a new raised limit */
+        FinalLimit = QuotaBlock->QuotaEntry[QuotaType].Limit - UpdatedLimit;
+        InterlockedExchangeSizeT(&QuotaBlock->QuotaEntry[QuotaType].Limit, FinalLimit);
```
ThFabba, line 260: In non-interlocked terms, this reads as
```
Limit = Limit - UpdatedLimit;
```
I would expect UpdatedLimit to be larger than Limit, so this seems problematic. Shouldn't we in effect be doing something like
```
Limit = UpdatedLimit;
```
?

ntoskrnl/ps/quota.c
```diff
@@ -18,123 +17,500 @@ EPROCESS_QUOTA_BLOCK PspDefaultQuotaBlock;
...
+        /*
+         * If the amount to return exceeds the threshold,
+         * the new amount becomes the default, otherwise
+         * the amount is just the one given by the caller.
+         */
+        AmountToReturn = min(Amount, ReturnThreshold);
+
+        /* Return the amounting we have back to Memory Manager */
+        InterlockedExchangeAddSizeT(QuotaBlock->QuotaEntry[QuotaType].Return, AmountToReturn);
+        MmReturnPoolQuota(QuotaType, QuotaBlock->QuotaEntry[QuotaType].Return);
```
ThFabba, line 397: This does not seem right. Both `PspReturnQuotasOnDestroy` and `PspReturnExcessQuotas` take `Return` and return that quota, which implies that we shouldn't _also_ return it here.

ntoskrnl/ps/quota.c
```diff
@@ -18,123 +17,500 @@ EPROCESS_QUOTA_BLOCK PspDefaultQuotaBlock;
...
+
+        /* And try to trim the limit */
+        InterlockedExchangeSizeT(&QuotaBlock->QuotaEntry[QuotaType].Limit,
+                                 QuotaBlock->QuotaEntry[QuotaType].Limit - AmountToReturn);
```
ThFabba, line 401: Why does returning quota change the limit?

ntoskrnl/ps/quota.c
```diff
@@ -18,123 +17,500 @@ EPROCESS_QUOTA_BLOCK PspDefaultQuotaBlock;
...
+    }
+
+    /* Update the usage member of the block */
+    UpdatedQuotaUsage = QuotaBlock->QuotaEntry[QuotaType].Usage - Amount;
+    InterlockedExchangeSizeT(&QuotaBlock->QuotaEntry[QuotaType].Usage, UpdatedQuotaUsage);
```
ThFabba, line 406: Nit: Right now this is correct (cause we use the spin lock everywhere), but I'd recommend also making this an `ExchangeAdd`, just like for `QuotaUsage` below, because that's both simpler (IMHO) and will remain correct once we remove the lock (no double-fetch)

ntoskrnl/ps/quota.c
```diff
@@ -143,45 +519,257 @@ PspInsertQuotaBlock(
...
+        }
+
+        /* As the process is now gone, decrement the process count */
+        InterlockedDecrementSizeT(&QuotaBlock->ProcessCount);
+    }
+
+    /* If no one is using this block, begin to destroy it */
     if (QuotaBlock != &PspDefaultQuotaBlock &&
-        InterlockedDecrementUL(&QuotaBlock->ReferenceCount) == 0)
+        InterlockedDecrementSizeT(&QuotaBlock->ReferenceCount) == 0)
```
ThFabba, line 576: `ReferenceCount` seems to be an `ULONG`, so I think this was correct?
Same for `ProcessCount` above in L571

ntoskrnl/mm/ARM3/pool.c
```diff
@@ -1388,6 +1375,181 @@ MiInitializeSessionPool(VOID)
...
+MmRaisePoolQuota(
+    _In_ POOL_TYPE PoolType,
+    _In_ SIZE_T CurrentMaxQuota,
+    _Out_ PSIZE_T NewMaxQuota)
+{
+    /*
+     * We must be in dispatch level interrupt here
+     * as we should be under a spin lock at this point.
+     */
+    ASSERT(KeGetCurrentIrql() == DISPATCH_LEVEL);
```
Extravert-ir, line 1416: For some reason we rarely use that macro, but there is `ASSERT_IRQL_EQUAL`

Extravert-ir (changes requested):

PR #4187, opened 2021-12-26
[MSPAINT] Fix Bezier behaviour

katahiromz:
## Purpose

JIRA issue: [CORE-17930](https://jira.reactos.org/browse/CORE-17930)

## Proposed changes

- Add `GetDistancePow2` helper function.
- Calculate the norms of two control points.
- Swap two control points if the norm is less than another norm.

## TODO

- [x] Do tests.

## Screenshot
after

### DIFF
--- a/base/applications/mspaint/drawing.cpp
+++ b/base/applications/mspaint/drawing.cpp
@@ -91,7 +91,7 @@ Poly(HDC hdc, POINT * lpPoints, int nCount,  COLORREF fg,  COLORREF bg, int thic
     SetROP2(hdc, oldRop);
 }
 
-LONG GetDistancePow2(POINT p1, POINT p2)
+static inline LONG GetDistancePow2(POINT p1, POINT p2)
 {
     LONG dx = p2.x - p1.x;
     LONG dy = p2.y - p1.y;
--- a/base/applications/mspaint/drawing.cpp
+++ b/base/applications/mspaint/drawing.cpp
@@ -91,7 +91,7 @@ Poly(HDC hdc, POINT * lpPoints, int nCount,  COLORREF fg,  COLORREF bg, int thic
     SetROP2(hdc, oldRop);
 }
 
-double GetDistancePow2(POINT p1, POINT p2)
+LONG GetDistancePow2(POINT p1, POINT p2)
 {
     LONG dx = p2.x - p1.x;
     LONG dy = p2.y - p1.y;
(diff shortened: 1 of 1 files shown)


base/applications/mspaint/drawing.cpp
```diff
@@ -91,14 +91,35 @@ Poly(HDC hdc, POINT * lpPoints, int nCount,  COLORREF fg,  COLORREF bg, int thic
...
 Bezier(HDC hdc, POINT p1, POINT p2, POINT p3, POINT p4, COLORREF color, int thickness)
 {
     HPEN oldPen;
     POINT fourPoints[4];
+    double distance12 = GetDistancePow2(p1, p2);
+    double distance13 = GetDistancePow2(p1, p3);
+    double distance42 = GetDistancePow2(p4, p2);
+    double distance43 = GetDistancePow2(p4, p3);
+    double norm1 = distance42 - distance12;
+    double norm2 = distance43 - distance13;
```
HBelusca, line 111: Why using doubles? The GetDistancePow2 evaluated squared differences of LONG, so the result is naturally a LONG (if casted to a double, the decimal part would always be zero), and then you look at differences of distances, they will still be LONG.

base/applications/mspaint/drawing.cpp
```diff
@@ -91,14 +91,35 @@ Poly(HDC hdc, POINT * lpPoints, int nCount,  COLORREF fg,  COLORREF bg, int thic
     SetROP2(hdc, oldRop);
 }
 
+double GetDistancePow2(POINT p1, POINT p2)
```
HBelusca, line 94: Can return `LONG` instead (see comment below). Also, it could be marked `inline`.

vicmarcal: At first sight seems it is returning LONG * LONG + LONG * LONG, can we be sure that those LONGs are small enough that the result will fit in a LONG too?

katahiromz: Fixed.

base/applications/mspaint/drawing.cpp
katahiromz: Fixed.

base/applications/mspaint/drawing.cpp
katahiromz: 5000 * 5000 == 25000000.
`MAXLONG` == 4294967295.
Good enough!

roybaer:
To be honest, I don't quite get what this does and more importantly why it does what it does.
Does it try to enforce that a control point is close to an endpoint?
If it does, that would actually limit the tool's functionality: You can no linger create loops using the bezier tool.

katahiromz:
MS mspaint in Windows 10:
image
Sorry. My assumption is wrong. I will close this PR.

PR #4188, opened 2021-12-26
[MSPAINT] Add Zoomed and UnZoomed and use them

katahiromz:
## Purpose
Our code became too complicated. Try to simplify.

## Proposed changes

- Define `Zoomed` and `UnZoomed` helper functions.
- Use them.

### DIFF
--- a/base/applications/mspaint/common.h
+++ b/base/applications/mspaint/common.h
@@ -11,6 +11,6 @@
 
 /* FUNCTIONS ********************************************************/
 
-BOOL zoomTo(int, int, int);
+BOOL zoomTo(int newZoom, int mouseX, int mouseY);
 int Zoomed(int xy);
 int UnZoomed(int xy);
--- a/base/applications/mspaint/common.h
+++ b/base/applications/mspaint/common.h
@@ -9,8 +9,20 @@
 
 #pragma once
 
+#include "toolsmodel.h"
+
+extern ToolsModel toolsModel;
+
 /* FUNCTIONS ********************************************************/
 
 BOOL zoomTo(int newZoom, int mouseX, int mouseY);
-int Zoomed(int xy);
-int UnZoomed(int xy);
+
+static inline int Zoomed(int xy)
+{
+    return xy * toolsModel.GetZoom() / 1000;
+}
+
+static inline int UnZoomed(int xy)
+{
+    return xy * 1000 / toolsModel.GetZoom();
+}
--- a/base/applications/mspaint/common.h
+++ b/base/applications/mspaint/common.h
@@ -11,4 +11,6 @@
 
 /* FUNCTIONS ********************************************************/
 
-extern BOOL zoomTo(int, int, int);
+BOOL zoomTo(int, int, int);
+int Zoomed(int xy);
+int UnZoomed(int xy);
--- a/base/applications/mspaint/common.h
+++ b/base/applications/mspaint/common.h
@@ -5,13 +5,10 @@
  * PURPOSE:     Commonly used functions
  * PROGRAMMERS: Benedikt Freisen
  *              Stanislav Motylkov
+ *              Katayama Hirofumi MZ
  */
 
 #pragma once
-
-#include "toolsmodel.h"
-
-extern ToolsModel toolsModel;
 
 /* FUNCTIONS ********************************************************/
 
--- a/base/applications/mspaint/imgarea.cpp
+++ b/base/applications/mspaint/imgarea.cpp
@@ -76,22 +76,22 @@ LRESULT CImgAreaWindow::OnSize(UINT nMsg, WPARAM wParam, LPARAM lParam, BOOL& bH
                0,
                0, 3, 3, TRUE);
     sizeboxCenterTop.MoveWindow(
-               Zoomed(imgXRes / 2) + 3 * 3 / 4,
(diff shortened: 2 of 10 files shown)


base/applications/mspaint/imgarea.cpp
```diff
@@ -76,26 +76,26 @@ LRESULT CImgAreaWindow::OnSize(UINT nMsg, WPARAM wParam, LPARAM lParam, BOOL& bH
                0,
                0, 3, 3, TRUE);
     sizeboxCenterTop.MoveWindow(
-               imgXRes * toolsModel.GetZoom() / 2000 + 3 * 3 / 4,
+               toolsModel.Zoomed(imgXRes / 2) + 3 * 3 / 4,
```
HBelusca, line 79: Would it be more correct to do `toolsModel.Zoomed(imgXRes) / 2` (plus the rest) instead?

base/applications/mspaint/winproc.cpp
```diff
@@ -39,7 +39,7 @@ zoomTo(int newZoom, int mouseX, int mouseY)
     toolsModel.SetZoom(newZoom);
 
     selectionWindow.ShowWindow(SW_HIDE);
-    imageArea.MoveWindow(3, 3, imageModel.GetWidth() * toolsModel.GetZoom() / 1000, imageModel.GetHeight() * toolsModel.GetZoom() / 1000, FALSE);
+    imageArea.MoveWindow(3, 3, toolsModel.Zoomed(imageModel.GetWidth()) / 1000, toolsModel.Zoomed(imageModel.GetHeight()), FALSE);
```
HBelusca, line 42:
```suggestion
    imageArea.MoveWindow(3, 3, toolsModel.Zoomed(imageModel.GetWidth()), toolsModel.Zoomed(imageModel.GetHeight()), FALSE);
```

base/applications/mspaint/winproc.cpp
```diff
@@ -39,7 +39,7 @@ zoomTo(int newZoom, int mouseX, int mouseY)
     toolsModel.SetZoom(newZoom);
 
     selectionWindow.ShowWindow(SW_HIDE);
-    imageArea.MoveWindow(3, 3, imageModel.GetWidth() * toolsModel.GetZoom() / 1000, imageModel.GetHeight() * toolsModel.GetZoom() / 1000, FALSE);
+    imageArea.MoveWindow(3, 3, Zoomed(imageModel.GetWidth()) / 1000, Zoomed(imageModel.GetHeight()), FALSE);
```
HBelusca, line 42:
```suggestion
    imageArea.MoveWindow(3, 3, Zoomed(imageModel.GetWidth()), Zoomed(imageModel.GetHeight()), FALSE);
```

base/applications/mspaint/winproc.cpp
katahiromz: OK.

base/applications/mspaint/winproc.cpp
katahiromz: OK.

base/applications/mspaint/imgarea.cpp
katahiromz: Fixed.

base/applications/mspaint/common.h
```diff
@@ -11,4 +11,6 @@
 
 /* FUNCTIONS ********************************************************/
 
-extern BOOL zoomTo(int, int, int);
+BOOL zoomTo(int, int, int);
```
HBelusca, line 14:
```suggestion
BOOL zoomTo(int newZoom, int mouseX, int mouseY);
```

base/applications/mspaint/common.h
```diff
@@ -11,4 +11,6 @@
...
+int Zoomed(int xy);
+int UnZoomed(int xy);
```
HBelusca, line 16: I think that these two functions could be implemented in this file and marked as inlined.

base/applications/mspaint/common.h
katahiromz: OK.

base/applications/mspaint/common.h
katahiromz: Needs a global variable and its class definition.

HBelusca: If common.h gets included after inclusion of toolsModel in the source files that use it, then it should work (without the need for common.h itself to include that other .h) 🤔

katahiromz: Fixed.

base/applications/mspaint/precomp.h
```diff
@@ -43,5 +42,6 @@
 #include "toolsettings.h"
 #include "toolsmodel.h"
 #include "winproc.h"
+#include "common.h"
```
JoachimHenze, line 45: What was the exact reason you had to move this include?

katahiromz: Nothing.

katahiromz: To use the global variable, its type should be predefined.

katahiromz: https://github.com/reactos/reactos/pull/4188#discussion_r775185392

roybaer (approved):
Yes, this looks like sensible refactoring.
I would probably have used macros out of habit, but `static inline` functions should work at least equally well.

PR #4189, opened 2021-12-26
[MSPAINT] Fix the size of settings if too large

katahiromz:
## Purpose

If the bitmap size from the registry settings is too large, we should fix it.
JIRA issue: N/A

## Proposed changes

- Fix the initial values of `BMPHeight` and `BMPWidth`.
- Fix the values of `BMPHeight` and `BMPWidth` if too large.

### DIFF
--- a/base/applications/mspaint/registry.cpp
+++ b/base/applications/mspaint/registry.cpp
@@ -104,12 +104,11 @@ void RegistrySettings::Load()
         ReadFileHistory(files, _T("File4"), strFile4);
     }
 
-    if (BMPWidth > 4096 || BMPHeight > 2304)
-    {
-        // Too large. Fix it.
-        BMPWidth = 400;
-        BMPHeight = 300;
-    }
+    // Fix the bitmap size if too large
+    if (BMPHeight > 5000)
+        BMPHeight = 460;
+    if (BMPWidth > 5000)
+        BMPWidth = 819;
 }
 
 void RegistrySettings::Store()
--- a/base/applications/mspaint/registry.cpp
+++ b/base/applications/mspaint/registry.cpp
@@ -103,6 +103,13 @@ void RegistrySettings::Load()
         ReadFileHistory(files, _T("File3"), strFile3);
         ReadFileHistory(files, _T("File4"), strFile4);
     }
+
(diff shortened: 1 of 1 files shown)


base/applications/mspaint/registry.cpp
```diff
@@ -103,6 +103,13 @@ void RegistrySettings::Load()
         ReadFileHistory(files, _T("File3"), strFile3);
         ReadFileHistory(files, _T("File4"), strFile4);
     }
+
+    if (BMPWidth > 4096 || BMPHeight > 2304)
```
HBelusca, line 107: These values look quite artificial. Why 2304 and not, say, 2305 or 6666 ?

katahiromz: I think bigger than 4K resolution is too large.

https://en.wikipedia.org/wiki/4K_resolution

HBelusca: How does Windows' paint behave, on machines that have a lot of RAM? (I'm asking because I would expect that this limitation is calculated at runtime, if there is even one...)

katahiromz: I have no idea, but MS mspaint surely does fix the setting values if too large.

katahiromz: image
I found the limitation is 5000x5000.

katahiromz: Fixed.

base/applications/mspaint/registry.cpp
```diff
@@ -104,12 +104,11 @@ void RegistrySettings::Load()
...
-    {
-        // Too large. Fix it.
-        BMPWidth = 400;
-        BMPHeight = 300;
-    }
+    // Fix the bitmap size if too large
+    if (BMPHeight > 5000)
+        BMPHeight = 460;
+    if (BMPWidth > 5000)
+        BMPWidth = 819;
```
HBelusca, line 111: These new defaults are at odds with those from LoadPresets .
Looks like 819 and 460 were based on current VM screen size minus some margins.

katahiromz: Seems like 60%.

PR #4192, opened 2021-12-26
[MSPAINT] Define GRIP_SIZE macro and use it

katahiromz:
## Purpose
Reduce magic numbers.

## Proposed changes

- Define `GRIP_SIZE` macro (its value is `3`) in `common.h`.
- Use it.

## Screenshot

image

### DIFF
--- a/base/applications/mspaint/imgarea.cpp
+++ b/base/applications/mspaint/imgarea.cpp
@@ -76,24 +76,24 @@ LRESULT CImgAreaWindow::OnSize(UINT nMsg, WPARAM wParam, LPARAM lParam, BOOL& bH
                0,
                0, GRIP_SIZE, GRIP_SIZE, TRUE);
     sizeboxCenterTop.MoveWindow(
-               GRIP_SIZE + Zoomed(imgXRes - GRIP_SIZE) / 2,
+               GRIP_SIZE + (Zoomed(imgXRes) - GRIP_SIZE) / 2,
                0, GRIP_SIZE, GRIP_SIZE, TRUE);
     sizeboxRightTop.MoveWindow(
                GRIP_SIZE + Zoomed(imgXRes),
                0, GRIP_SIZE, GRIP_SIZE, TRUE);
     sizeboxLeftCenter.MoveWindow(
                0,
-               GRIP_SIZE + Zoomed(imgYRes - GRIP_SIZE) / 2,
+               GRIP_SIZE + (Zoomed(imgYRes) - GRIP_SIZE) / 2,
                GRIP_SIZE, GRIP_SIZE, TRUE);
     sizeboxRightCenter.MoveWindow(
                GRIP_SIZE + Zoomed(imgXRes),
-               GRIP_SIZE + Zoomed(imgYRes - GRIP_SIZE) / 2,
+               GRIP_SIZE + (Zoomed(imgYRes) - GRIP_SIZE) / 2,
                GRIP_SIZE, GRIP_SIZE, TRUE);
     sizeboxLeftBottom.MoveWindow(
                0,
                GRIP_SIZE + Zoomed(imgYRes), GRIP_SIZE, GRIP_SIZE, TRUE);
     sizeboxCenterBottom.MoveWindow(
-               GRIP_SIZE + Zoomed(imgXRes - GRIP_SIZE) / 2,
+               GRIP_SIZE + (Zoomed(imgXRes) - GRIP_SIZE) / 2,
                GRIP_SIZE + Zoomed(imgYRes), GRIP_SIZE, GRIP_SIZE, TRUE);
     sizeboxRightBottom.MoveWindow(
                GRIP_SIZE + Zoomed(imgXRes),
--- a/base/applications/mspaint/imgarea.cpp
+++ b/base/applications/mspaint/imgarea.cpp
@@ -76,26 +76,28 @@ LRESULT CImgAreaWindow::OnSize(UINT nMsg, WPARAM wParam, LPARAM lParam, BOOL& bH
                0,
                0, GRIP_SIZE, GRIP_SIZE, TRUE);
     sizeboxCenterTop.MoveWindow(
-               Zoomed(imgXRes) / 2 + 3 * 3 / 4,
+               GRIP_SIZE + Zoomed(imgXRes - GRIP_SIZE) / 2,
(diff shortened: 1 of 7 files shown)


base/applications/mspaint/imgarea.cpp
```diff
@@ -74,28 +74,28 @@ LRESULT CImgAreaWindow::OnSize(UINT nMsg, WPARAM wParam, LPARAM lParam, BOOL& bH
     int imgYRes = imageModel.GetHeight();
     sizeboxLeftTop.MoveWindow(
                0,
-               0, 3, 3, TRUE);
+               0, GRIP_SIZE, GRIP_SIZE, TRUE);
     sizeboxCenterTop.MoveWindow(
                Zoomed(imgXRes) / 2 + 3 * 3 / 4,
```
roybaer, line 79: I believe that the position of all the center grips takes the grip size into account, too, but I'm 100% sure how.
(It's been a few days...)

base/applications/mspaint/selection.cpp
```diff
@@ -65,29 +65,29 @@ ForceRefreshSelectionContents()
...
 int CSelectionWindow::IdentifyCorner(int iXPos, int iYPos, int iWidth, int iHeight)
 {
-    if (iYPos < 3)
+    if (iYPos < GRIP_SIZE)
     {
-        if (iXPos < 3)
+        if (iXPos < GRIP_SIZE)
             return ACTION_RESIZE_TOP_LEFT;
-        if ((iXPos < iWidth / 2 + 2) && (iXPos >= iWidth / 2 - 1))
+        if ((iWidth / 2 - 1 <= iXPos) && (iXPos < iWidth / 2 + 2))
```
roybaer, line 72: The -1 and +2 in here is probably also related to GRIP_SIZE.

roybaer (commented):
Getting rid of the magic numbers is definitely a good idea.
IIRC, somebody requested making them resizable a long time ago.
A GRIP_SIZE macro would simplify that, as well, either by changing the macro definition or via search+replace, but let's stick with the macro constant for now.

base/applications/mspaint/imgarea.cpp
katahiromz: What does the expression `3 * 3 / 4` mean?

base/applications/mspaint/selection.cpp
katahiromz: `2 == GRIP_SIZE - 1`. Right?

base/applications/mspaint/imgarea.cpp
roybaer: To be honest, I have no idea. That's the problem with magic numbers that are more than a decade old. ¯\_(ツ)_/¯

base/applications/mspaint/selection.cpp
roybaer: Probably more like the interval from GRIP_SIZE - floor(GRIP_SIZE / 2) to GRIP_SIZE + ceil(GRIP_SIZE / 2).
At least that would make the most sense.

katahiromz:
image

base/applications/mspaint/selection.cpp
katahiromz: Done.

base/applications/mspaint/imgarea.cpp
katahiromz: Done.

base/applications/mspaint/imgarea.cpp
```diff
@@ -74,28 +74,30 @@ LRESULT CImgAreaWindow::OnSize(UINT nMsg, WPARAM wParam, LPARAM lParam, BOOL& bH
     int imgYRes = imageModel.GetHeight();
     sizeboxLeftTop.MoveWindow(
                0,
-               0, 3, 3, TRUE);
+               0, GRIP_SIZE, GRIP_SIZE, TRUE);
     sizeboxCenterTop.MoveWindow(
-               Zoomed(imgXRes) / 2 + 3 * 3 / 4,
-               0, 3, 3, TRUE);
+               GRIP_SIZE + Zoomed(imgXRes - GRIP_SIZE) / 2,
```
roybaer, line 79: Why would you zoom the grip size? This doesn't make sense to me.

base/applications/mspaint/selection.cpp
```diff
@@ -65,31 +65,24 @@ ForceRefreshSelectionContents()
...
-    {
-        if (iXPos < 3)
-            return ACTION_RESIZE_BOTTOM_LEFT;
-        if ((iXPos < iWidth / 2 + 2) && (iXPos >= iWidth / 2 - 1))
-            return ACTION_RESIZE_BOTTOM;
-        if (iXPos >= iWidth - 3)
-            return ACTION_RESIZE_BOTTOM_RIGHT;
-    }
+    POINT pt = { iXPos, iYPos };
+    HWND hwndChild = ChildWindowFromPointEx(pt, CWP_SKIPINVISIBLE | CWP_SKIPDISABLED);
```
roybaer, line 69: Now _this_ definitely makes sense.

base/applications/mspaint/imgarea.cpp
katahiromz: Fixed.

roybaer:
Incidentally, I think that we should keep `GRIP_SIZE == 3` while it is constant, because the various Paint versions with this UI style used 3.

roybaer (approved):
Looking good.

base/applications/mspaint/main.cpp
```diff
@@ -322,7 +322,7 @@ _tWinMain (HINSTANCE hThisInstance, HINSTANCE hPrevInstance, LPTSTR lpszArgument
     sizeboxCenterBottom.Create(scrlClientWindow.m_hWnd, sizeboxPos, NULL, WS_CHILD | WS_VISIBLE);
     sizeboxRightBottom.Create(scrlClientWindow.m_hWnd, sizeboxPos, NULL, WS_CHILD | WS_VISIBLE);
     /* placing the size boxes around the image */
-    imageArea.SendMessage(WM_SIZE, 0, 0);
+    imageArea.PostMessage(WM_SIZE, 0, 0);
```
HBelusca, line 325: What's the reason of this change?

katahiromz: Reverted.
https://github.com/reactos/reactos/commit/1c163bf163d5f315e4869cff3286df946fb0a637

PR #4193, opened 2021-12-26
[NTOSKRNL] Regroup the pool allocation tags in one dedicated place

GeoB99:
We already have a dedicated header file where we store pool allocation tags that are used in NT, called `tag.h`, and yet I still see some tags around the kernel which is annoying and confusing. Everyone who discovers a new pool tag or whatever has to use `tag.h` from now on.

### DIFF
--- a/ntoskrnl/dbgk/dbgkobj.c
+++ b/ntoskrnl/dbgk/dbgkobj.c
@@ -59,7 +59,7 @@ DbgkpQueueMessage(IN PEPROCESS Process,
         /* Allocate it */
         DebugEvent = ExAllocatePoolWithTag(NonPagedPool,
                                            sizeof(DEBUG_EVENT),
-                                           'EgbD');
+                                           TAG_DEBUG_EVENT);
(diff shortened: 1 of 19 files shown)


HBelusca:
> and yet I still see some tags around the kernel which is annoying and confusing.

That's when the TAG(s) in question are only used within that only file, in which case one can argue it's better to keep the file self-contained (it would have its own tags, its own typedefs that are private to it, its own private static functions, etc.)

GeoB99:
> > and yet I still see some tags around the kernel which is annoying and confusing.
> 
> That's when the TAG(s) in question are only used within that only file, in which case one can argue it's better to keep the file self-contained (it would have its own tags, its own typedefs that are private to it, its own private static functions, etc.)

That could be said the same for whatever tags that existed on their specific files which then got moved to that `tag.h` file by others (which were also private to their own files if you ask me)... That doesn't change the fact we have some tags in `tag.h` file of a XYZ component and with remaining tags of the same XYZ component to whatever files they reside.

HBelusca (approved):

PR #4194, opened 2021-12-27
[MSPAINT] Add nearlyEqualPoints and use it

katahiromz:
## Purpose

Refactoring...

## Proposed changes

- Add `nearlyEqualPoints` helper function and use it.
- The `nearlyEqualPoints`  function uses `GetSystemMetrics(SM_CXDRAG)` and `GetSystemMetrics(SM_CYDRAG)`.

### DIFF
--- a/base/applications/mspaint/mouse.cpp
+++ b/base/applications/mspaint/mouse.cpp
@@ -53,8 +53,8 @@ roundTo8Directions(LONG x0, LONG y0, LONG& x1, LONG& y1)
 
 BOOL nearlyEqualPoints(INT x0, INT y0, INT x1, INT y1)
 {
-    INT cxThreshold = Zoomed(toolsModel.GetLineWidth()) + GetSystemMetrics(SM_CXDRAG);
-    INT cyThreshold = Zoomed(toolsModel.GetLineWidth()) + GetSystemMetrics(SM_CYDRAG);
+    INT cxThreshold = toolsModel.GetLineWidth() + GetSystemMetrics(SM_CXDRAG);
+    INT cyThreshold = toolsModel.GetLineWidth() + GetSystemMetrics(SM_CYDRAG);
     return (abs(x1 - x0) <= cxThreshold) && (abs(y1 - y0) <= cyThreshold);
 }
 
--- a/base/applications/mspaint/mouse.cpp
+++ b/base/applications/mspaint/mouse.cpp
@@ -53,9 +53,9 @@ roundTo8Directions(LONG x0, LONG y0, LONG& x1, LONG& y1)
 
 BOOL nearlyEqualPoints(INT x0, INT y0, INT x1, INT y1)
 {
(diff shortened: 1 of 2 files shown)


base/applications/mspaint/mouse.cpp
```diff
@@ -50,6 +51,13 @@ roundTo8Directions(LONG x0, LONG y0, LONG& x1, LONG& y1)
     }
 }
 
+BOOL nearlyEqualPoints(INT x0, INT y0, INT x1, INT y1)
+{
+    INT cxThreshold = toolsModel.GetLineWidth() + GetSystemMetrics(SM_CXDRAG);
```
roybaer, line 56: Wouldn't the GetSystemMetrics result have to be scaled to the current zoom level?

roybaer (commented):
I don't know what the default values for SM_CXDRAG and SM_CYDRAG are, but they are certainly adjustable.
I'd say that this change makes sense with the small exception that the drag distance -- and only that -- should not be affected by zooming.

base/applications/mspaint/mouse.cpp
katahiromz: The line width can be zoomed. The drag distance can't be.

katahiromz: OK. Fixed.

roybaer: Wait. Don't you call nearlyEqualPoints with transformed (zoomed) coordinates? Wouldn't it have to be the other way around, then?
I.e. something like UnZoomed(GetSystemMetrics(....))?

katahiromz: Umm, Reverted. `x0`, `y0`, `x1`, `y1` are unzoomed values. So, we have to calculate the threshold values in unzoomed values.

katahiromz: `GetSystemMetrics(SM_CXDRAG)` is already an unzoomed value.

roybaer: Maybe the term "unzoomed" is a bit to ambiguous. I was under the assumption that nearlyEqualPoints gets called with coordinates in the image's coordinate system, whereas GetSystemMetrics uses the screen's coordinate system, which means that one of them has to be scaled. Is this assumption correct? If both are in the screen's coordinate system, the line width would have to be scaled, though.

katahiromz: In the screen coordinate system, the pixel and the coordinate are 1:1. `GetSystemMetrics(SM_CXDRAG)` is also 1:1. This `nearlyEqualPoints` takes the image's coordinates.

katahiromz: It cannot get the accurate scaled coordinates from the image's coordinates.

katahiromz: Applied `UnZoomed` to `cxDrag` and `cyDrag`.
image
LGTM.

roybaer (approved):

PR #4195, opened 2021-12-27
[MSPAINT] Mouse Wheel support

katahiromz:
## Purpose
Expect usability improvement by mouse wheel handling.
JIRA issue: [CORE-17937](https://jira.reactos.org/browse/CORE-17937)

## Proposed changes

- If the mouse wheel message has come to the child windows, then forward it to its parent window.
- Post some scroll messages on mouse wheel message of the main window.
- Or modify the zoom rate.
- Modify `CMakeLists.txt` to use `_WIN32_WINNT=0x600`.

### DIFF
--- a/base/applications/mspaint/CMakeLists.txt
+++ b/base/applications/mspaint/CMakeLists.txt
@@ -1,5 +1,7 @@
 project(MSPAINT)
 
+remove_definitions(-D_WIN32_WINNT=0x502)
+add_definitions(-D_WIN32_WINNT=0x600) # SPI_GETWHEELSCROLLCHARS needs this
 add_definitions(-DINITGUID)
 
 list(APPEND SOURCE
--- a/base/applications/mspaint/CMakeLists.txt
+++ b/base/applications/mspaint/CMakeLists.txt
@@ -1,7 +1,5 @@
 project(MSPAINT)
 
-remove_definitions(-D_WIN32_WINNT=0x502)
-add_definitions(-D_WIN32_WINNT=0x600) # SPI_GETWHEELSCROLLCHARS needs this
 add_definitions(-DINITGUID)
 
 list(APPEND SOURCE
--- a/base/applications/mspaint/common.h
+++ b/base/applications/mspaint/common.h
@@ -26,3 +26,5 @@ static inline int UnZoomed(int xy)
 }
 
 #define GRIP_SIZE 3
+#define MIN_ZOOM 125
+#define MAX_ZOOM 8000
--- a/base/applications/mspaint/winproc.cpp
+++ b/base/applications/mspaint/winproc.cpp
@@ -157,6 +157,51 @@ void CMainWindow::InsertSelectionFromHBITMAP(HBITMAP bitmap, HWND window)
     ForceRefreshSelectionContents();
 }
 
+LRESULT CMainWindow::OnMouseWheel(UINT nMsg, WPARAM wParam, LPARAM lParam, BOOL& bHandled)
+{
+    INT xPos = GET_X_LPARAM(lParam), yPos = GET_Y_LPARAM(lParam);
+    INT zDelta = (SHORT)HIWORD(wParam);
+    //UINT fwKeys = LOWORD(wParam);
+
+    if (::GetAsyncKeyState(VK_CONTROL) < 0)
+    {
+        if (zDelta < 0)
+        {
+            if (toolsModel.GetZoom() > MIN_ZOOM)
+                zoomTo(toolsModel.GetZoom() / 2, xPos, yPos);
+        }
+        else if (zDelta > 0)
+        {
+            if (toolsModel.GetZoom() < MAX_ZOOM)
+                zoomTo(toolsModel.GetZoom() * 2, xPos, yPos);
+        }
+    }
+    else
+    {
+#define SCROLL_COUNT 5
+        for (INT i = 0; i < SCROLL_COUNT; ++i)
+        {
+            if (::GetAsyncKeyState(VK_SHIFT) < 0)
+            {
+                if (zDelta < 0)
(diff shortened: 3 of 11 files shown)


base/applications/mspaint/winproc.cpp
```diff
@@ -157,6 +157,51 @@ void CMainWindow::InsertSelectionFromHBITMAP(HBITMAP bitmap, HWND window)
...
+        else if (zDelta > 0)
+        {
+            if (toolsModel.GetZoom() < MAX_ZOOM)
+                zoomTo(toolsModel.GetZoom() * 2, xPos, yPos);
+        }
+    }
+    else
+    {
+#define SCROLL_COUNT 5
+        for (INT i = 0; i < SCROLL_COUNT; ++i)
```
roybaer, line 182: What exactly does SCROLL_COUNT do?
If it is the amount by which to scroll, isn't there a configurable system setting for that, as well?
Also, the pre-processor define here looks a bit awkward.

katahiromz:
The window hierarchy of mspaint is like this:
image

roybaer (commented):
Not that I don't find this useful, but does this replicate functionality from any existing Paint version? Particularly the VK_CONTROL and VK_SHIFT stuff?

base/applications/mspaint/winproc.cpp
katahiromz: One message had not enough effective in the scroll amount.
OK, now it uses sys params.

base/applications/mspaint/CMakeLists.txt
```diff
@@ -1,5 +1,7 @@
 project(MSPAINT)
 
+remove_definitions(-D_WIN32_WINNT=0x502)
+add_definitions(-D_WIN32_WINNT=0x600) # SPI_GETWHEELSCROLLCHARS needs this
```
roybaer, line 4: What will it do when run on Win2003?
Will it still run but use a default value or will it need a newer Windows version?
If it's the former, that should be fine. The latter sounds a bit problematic for such a trivial feature.

base/applications/mspaint/common.h
```diff
@@ -26,3 +26,5 @@ static inline int UnZoomed(int xy)
 }
 
 #define GRIP_SIZE 3
+#define MIN_ZOOM 125
+#define MAX_ZOOM 8000
```
roybaer, line 30: Makes perfect sense. Fewer magic numbers.

base/applications/mspaint/winproc.cpp
katahiromz: Windows 10 Paint does support Ctrl+Wheel and Plain Wheel but doesn't support Shift+Wheel.

roybaer (commented):
Apart from the compatibility question this looks good to me.

base/applications/mspaint/CMakeLists.txt
katahiromz: I know ReactOS supports `SPI_GETWHEELSCROLLCHARS`. But, this macro isn't defined in PSDK of `_WIN32_WINNT=0x502`.

katahiromz: I can still use a preprocessor magic...

base/applications/mspaint/winproc.cpp
roybaer: In that context Shift+Wheel looks like a sensible minor addition.

base/applications/mspaint/CMakeLists.txt
katahiromz: Done.

base/applications/mspaint/winproc.cpp
```diff
@@ -157,6 +157,56 @@ void CMainWindow::InsertSelectionFromHBITMAP(HBITMAP bitmap, HWND window)
...
+            if (toolsModel.GetZoom() < MAX_ZOOM)
+                zoomTo(toolsModel.GetZoom() * 2, 0, 0);
+        }
+    }
+    else
+    {
+        UINT nCount = 3;
+        if (::GetAsyncKeyState(VK_SHIFT) < 0)
+        {
+#ifndef SPI_GETWHEELSCROLLCHARS
```
roybaer, line 182: Ideally explain this "magic trick" for posterity.
"Needed for pre-NT6 PSDK" or something.

katahiromz: OK.

roybaer (approved):

base/applications/mspaint/CMakeLists.txt
HBelusca: And you could always fall-back to e.g. code like the scrolling code in our console, that does not use this SPI setting.

base/applications/mspaint/winproc.cpp
```diff
@@ -157,6 +157,56 @@ void CMainWindow::InsertSelectionFromHBITMAP(HBITMAP bitmap, HWND window)
     ForceRefreshSelectionContents();
 }
 
+LRESULT CMainWindow::OnMouseWheel(UINT nMsg, WPARAM wParam, LPARAM lParam, BOOL& bHandled)
```
HBelusca, line 160: How does the mouse scrolling (not zoom) behave, if all the conditions happen:
1. The system returns some usual value for `SPI_GETWHEELSCROLLCHARS` (not too much),
2. But the monitor where paint is being tested is using a super-large resolution,
3. Paint has a very large picture opened that needs scrolling.
4. Zoom is at 100%.

Question: When scrolling, is it going to look like the image is scrolling "slowly" ? Or is it going to be proportional to the screen's (not the paint window) size? And of course it also needs to be "inversely" proportional to the current zoom level (if zoomed 200%, then one needs a scrolling of half the amount of what it was when the zoom was 100%).

HBelusca:
> Modify CMakeLists.txt to use _WIN32_WINNT=0x600.

Ideally one would require to just copy-paste our paint into windows 2003 without having to change that line beforehand...

PR #4196, opened 2021-12-27
[DEVMGR] Ensure clicking on item when opening property sheet

StartForKillerMC:
## Purpose

_Fix a bug related to the device manager where double clicking anywhere when you have selected an item opens the property sheet_

JIRA issue: [CORE-17207](https://jira.reactos.org/browse/CORE-17207)

## Proposed changes

_Ensure that you are clicking an item when opening the property sheet_

### DIFF
--- a/dll/win32/devmgr/devmgmt/DeviceView.cpp
+++ b/dll/win32/devmgr/devmgmt/DeviceView.cpp
@@ -125,6 +125,27 @@ CDeviceView::OnSize(
 }
 
 LRESULT
+CDeviceView::OnDoubleClick(
+    _In_ LPNMHDR NmHdr
+    )
+{
+    TVHITTESTINFO hitInfo;
+    HTREEITEM hItem;
+
+    GetCursorPos(&hitInfo.pt);
+    ScreenToClient(m_hTreeView, &hitInfo.pt);
+
+    // Check if we are trying to double click an item
+    hItem = TreeView_HitTest(m_hTreeView, &hitInfo);
+    if (hItem != NULL && (hitInfo.flags & (TVHT_ONITEM | TVHT_ONITEMICON)))
+    {
+        DisplayPropertySheet();
+    }
+
+    return 0;
+}
+
+LRESULT
 CDeviceView::OnRightClick(
     _In_ LPNMHDR NmHdr
     )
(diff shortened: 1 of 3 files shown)


GeoB99 (dismissed):
Hello,
[Please amend your commit with your full name](https://stackoverflow.com/questions/3042437/how-to-change-the-commit-author-for-one-specific-commit/3042512#3042512), and update this PR.

```
```

```
```

Regards.

StartForKillerMC:
Ok now i rebased the last commit ammending the full name, and i changed the name in my user profile

HBelusca:
Just to be 100% sure, was this bug also happening when running **our** devmgmt on Windows? (note: for a fair test, it would need both **our** devmgmt.exe and **its** accompanying dll, put side to side in same directory.)

StartForKillerMC:
I can confirm the bug is present on windows 7 too, so the problem is related to the application itself, i will upload a video to prove it soon

StartForKillerMC:
Okay here you have a prove that in windows the problem is the same, i tested but not recorded the new version patched and works fine on windows too.
https://user-images.githubusercontent.com/21644562/147491708-25353651-3bec-4dd0-9a87-254afda2686a.mp4

dll/win32/devmgr/devmgmt/DeviceView.cpp
```diff
@@ -124,6 +124,27 @@ CDeviceView::OnSize(
...
+    _In_ LPNMHDR NmHdr
+    )
+{
+    TVHITTESTINFO hitInfo;
+    HTREEITEM hItem;
+
+    GetCursorPos(&hitInfo.pt);
+    ScreenToClient(m_hTreeView, &hitInfo.pt);
+
+    //Check if we are trying to double click an item
```
HBelusca, line 138:
```suggestion
    // Check if we are trying to double click an item
```

HBelusca (approved):

StartForKillerMC:
@GeoB99 i requested you a review and the workflow seems stopped until marked as done, it was my fault before

GeoB99 (approved):
You need to squash your commits into one, otherwise LGTM.

StartForKillerMC:
Okay @GeoB99 done

StartForKillerMC:
the only ting that is needed is the aproval of the workflow

PR #4199, opened 2021-12-27
[COMCTL32] Fix a bug related to band visibility

StartForKillerMC:
## Purpose

_This bug is present when hidding the standard buttons band causing to render it on top of the search band_

JIRA issue: [CORE-17236](https://jira.reactos.org/browse/CORE-17236)

## Proposed changes

_Change when the windows of the band visibility is changed to show or hide the window_

### DIFF
--- a/dll/win32/comctl32/rebar.c
+++ b/dll/win32/comctl32/rebar.c
@@ -1713,13 +1713,6 @@ REBAR_CommonSetupBand(HWND hwnd, const REBARBANDINFOW *lprbbi, REBAR_BAND *lpBan
 
     lpBand->fMask |= lprbbi->fMask;
 
-    if( (lprbbi->fMask & RBBIM_STYLE) &&
-        (lpBand->fStyle != lprbbi->fStyle ) )
-    {
-	lpBand->fStyle = lprbbi->fStyle;
-        uChanged |= RBBIM_STYLE;
-    }
-
     if( (lprbbi->fMask & RBBIM_COLORS) &&
        ( ( lpBand->clrFore != lprbbi->clrFore ) ||
          ( lpBand->clrBack != lprbbi->clrBack ) ) )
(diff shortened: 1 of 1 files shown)


dll/win32/browseui/internettoolbar.cpp
```diff
@@ -1086,12 +1086,8 @@ HRESULT CInternetToolbar::ToggleBandVisibility(int BandID)
     REBARBANDINFOW bandInfo = {sizeof(REBARBANDINFOW), RBBIM_STYLE};
     SendMessage(fMainReBar, RB_GETBANDINFOW, index, (LPARAM)&bandInfo);
 
-    if (bandInfo.fStyle & RBBS_HIDDEN)
-        bandInfo.fStyle &= ~RBBS_HIDDEN;
-    else
-        bandInfo.fStyle |= RBBS_HIDDEN;
-
-    SendMessage(fMainReBar, RB_SETBANDINFOW, index, (LPARAM)&bandInfo);
```
HBelusca, line 1094: The fact that the original code didn't work could also indicate a bug in the rebar code implementation in Wine's comctl32...

StartForKillerMC: Well, at least with this change the code works, it's not a bug from this problem, and with this method the code seems more cleaner

StartForKillerMC: i could check if there is a bug on rebar anyways

StartForKillerMC: and im not sure that this is a bug in rebar, if it is the change needed is to call showwindow inside RB_SETBANDINFOW callback, that the unique thing RB_SHOWBAND does

StartForKillerMC:
Anyways, any part of the application seems to use the manual way of changing visibility, so it's ok to change that, so, the pr is ok?

StartForKillerMC:
Ok, searching for different tools that use this way to toggle the band show me that the bug is within the comctl32 itself, i can update the change needed, should i make a commit?

StartForKillerMC:
Okay changes pushed
should I send a patch to wine too?

StartForKillerMC:
> Okay changes pushed should I send a patch to wine too?

Nevermind, i submited it anyways as it works for me

HBelusca:
Wine would definitively want some sort of proof that, on Windows, the rebar's RB_SETBANDINFOW with these flags indeed does change its visibility.

StartForKillerMC:
@HBelusca indeed the proof is that the intended use of the RB_SETBANDINFOW is not correct, if you specify RBBIM_STYLE and the change the visibility the visibility changes too in the window, so that's the proof

dll/win32/comctl32/rebar.c
```diff
@@ -1713,13 +1713,6 @@ REBAR_CommonSetupBand(HWND hwnd, const REBARBANDINFOW *lprbbi, REBAR_BAND *lpBan
 
     lpBand->fMask |= lprbbi->fMask;
 
-    if( (lprbbi->fMask & RBBIM_STYLE) &&
```
learn-more, line 1716: This file is winesynced,
we usually surround our changes with `#ifdef __REACTOS__` in those until it's upstreamed.

StartForKillerMC: ok, then i would update it to include the ifdef

StartForKillerMC: Done

JoachimHenze (changes requested):
PR headline and commit comment should be changed:
from [BROWSEUI] to [COMCTL32]

PR #4203, opened 2021-12-28
[MSPAINT] Add NUM_COLORS and enum PAL_TYPE

katahiromz:
## Purpose
Reduce magic numbers and improve debuggability.

## Proposed changes

- Define `NUM_COLORS` macro and `enum PAL_TYPE` in `palettemodel.h`.
- Use them.
- Add `static const` to `modernColors` and `oldColors` in `PaletteModel::SelectPalette`.
- Use `COLORREF` for some color values.

### DIFF
--- a/base/applications/mspaint/palettemodel.cpp
+++ b/base/applications/mspaint/palettemodel.cpp
@@ -54,17 +54,17 @@ void PaletteModel::SelectPalette(PAL_TYPE nPalette)
     NotifyPaletteChanged();
 }
 
-COLORREF PaletteModel::GetColor(int nIndex) const
+COLORREF PaletteModel::GetColor(UINT nIndex) const
 {
-    if (0 <= nIndex && nIndex < NUM_COLORS)
+    if (nIndex < NUM_COLORS)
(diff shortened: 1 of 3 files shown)


base/applications/mspaint/palettemodel.cpp
```diff
@@ -16,72 +17,77 @@ PaletteModel::PaletteModel()
...
+        case PAL_OLDTYPE:
+            CopyMemory(m_colors, oldColors, sizeof(m_colors));
+            break;
+    }
     m_nSelectedPalette = nPalette;
     NotifyPaletteChanged();
 }
 
-int PaletteModel::GetColor(int nIndex) const
+COLORREF PaletteModel::GetColor(int nIndex) const
```
HBelusca, line 57: Why then not forcing the nIndex to always be a positive integer? (use e.g. `UINT`) Then it would be always clear that it cannot be negative.

katahiromz: Used `UINT`.

roybaer (approved):
Looks fine to me.

PR #4204, opened 2021-12-28
[NTOS:MM] Implement partial virtual region releases.

ThFabba:
Makes MS videoprt work in ROS.
Dedicated to @hpoussin

### DIFF
--- a/ntoskrnl/mm/ARM3/virtual.c
+++ b/ntoskrnl/mm/ARM3/virtual.c
@@ -5210,8 +5210,10 @@ NtFreeVirtualMemory(IN HANDLE ProcessHandle,
     SIZE_T PRegionSize;
     PVOID PBaseAddress;
     LONG_PTR AlreadyDecommitted, CommitReduction = 0;
+    LONG_PTR FirstCommit;
     ULONG_PTR StartingAddress, EndingAddress;
     PMMVAD Vad;
+    PMMVAD NewVad;
     NTSTATUS Status;
     PEPROCESS Process;
     PMMSUPPORT AddressSpace;
@@ -5437,6 +5439,8 @@ NtFreeVirtualMemory(IN HANDLE ProcessHandle,
                 if ((EndingAddress >> PAGE_SHIFT) == Vad->EndingVpn)
                 {
                     //
+                    // Case D (freeing the entire region)
+                    //
                     // This is the easiest one to handle -- it is identical to
                     // the code path above when the caller sets a zero region size
                     // and the whole VAD is destroyed
@@ -5448,15 +5452,23 @@ NtFreeVirtualMemory(IN HANDLE ProcessHandle,
                 else
                 {
                     //
+                    // Case A (freeing a part at the beginning)
+                    //
                     // This case is pretty easy too -- we compute a bunch of
                     // pages to decommit, and then push the VAD's starting address
                     // a bit further down, then decrement the commit charge
                     //
-                    // NOT YET IMPLEMENTED IN ARM3.
-                    //
-                    DPRINT1("Case A not handled\n");
-                    Status = STATUS_FREE_VM_NOT_AT_BASE;
-                    goto FailPath;
+                    MiLockProcessWorkingSetUnsafe(Process, CurrentThread);
+                    CommitReduction = MiCalculatePageCommitment(StartingAddress,
+                                                                EndingAddress,
+                                                                Vad,
+                                                                Process);
+                    Vad->u.VadFlags.CommitCharge -= CommitReduction;
+                    // For ReactOS: shrink the corresponding memory area
+                    ASSERT(Vad->StartingVpn == MemoryArea->VadNode.StartingVpn);
+                    ASSERT(Vad->EndingVpn == MemoryArea->VadNode.EndingVpn);
+                    Vad->StartingVpn = (EndingAddress + 1) >> PAGE_SHIFT;
+                    MemoryArea->VadNode.StartingVpn = Vad->StartingVpn;
 
                     //
                     // After analyzing the VAD, set it to NULL so that we don't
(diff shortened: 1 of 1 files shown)


ntoskrnl/mm/ARM3/virtual.c
```diff
@@ -5452,11 +5454,17 @@ NtFreeVirtualMemory(IN HANDLE ProcessHandle,
...
-                    DPRINT1("Case A not handled\n");
-                    Status = STATUS_FREE_VM_NOT_AT_BASE;
-                    goto FailPath;
+                    MiLockProcessWorkingSetUnsafe(Process, CurrentThread);
+                    CommitReduction = MiCalculatePageCommitment(StartingAddress,
+                                                                EndingAddress,
+                                                                Vad,
+                                                                Process);
+                    Vad->u.VadFlags.CommitCharge -= CommitReduction;
+                    // For ReactOS: shrink the corresponding memory area
```
HBelusca, line 5463: Why "for ROS"? Does this suppose that Windows does more things there?

ntoskrnl/mm/ARM3/virtual.c
```diff
@@ -5452,11 +5454,17 @@ NtFreeVirtualMemory(IN HANDLE ProcessHandle,
                     // pages to decommit, and then push the VAD's starting address
                     // a bit further down, then decrement the commit charge
                     //
-                    // NOT YET IMPLEMENTED IN ARM3.
-                    //
-                    DPRINT1("Case A not handled\n");
```
HBelusca, line 5457: OK if this else-clause is indeed case A (or any other for that matter), please write in a comment that this is such "case". This allows people to understand what's going on in this whole algorithm and to not loose track.

ntoskrnl/mm/ARM3/virtual.c
ThFabba: `MEMORY_AREA` is a ROS-exclusive old Mm structure. We need to maintain them for memory regions owned by ARM3 so that old Mm doesn't think that the corresponding memory range is free and puts other stuff there.

ntoskrnl/mm/ARM3/virtual.c
ThFabba: Will do.

ntoskrnl/mm/ARM3/virtual.c
```diff
@@ -5436,6 +5438,8 @@ NtFreeVirtualMemory(IN HANDLE ProcessHandle,
                 //
                 if ((EndingAddress >> PAGE_SHIFT) == Vad->EndingVpn)
                 {
+                    //
+                    // Case D (freeing the entire region)
```
Extravert-ir, line 5442: Nit: the comment above `// Check for case D` is not needed anymore

ntoskrnl/mm/ARM3/virtual.c
```diff
@@ -5496,16 +5507,73 @@ NtFreeVirtualMemory(IN HANDLE ProcessHandle,
...
-                    // actually split the VAD into two new VADs and compute the
-                    // commit charges for each of them, and reinsert new charges.
+                    // Case B (freeing a part in the middle)
+                    //
+                    // This is the hardest one. Because we are removing a chunk
+                    // of memory from the very middle of the VAD, we must actually
+                    // split the VAD into two new VADs and compute the commit
+                    // charges for each of them, and reinsert new charges.
+                    //
+                    NewVad = ExAllocatePoolZero(NonPagedPool, sizeof(MMVAD_LONG), 'SdaV');
```
Extravert-ir, line 5517: Nit: I'd rather move variable declarations to this block

ntoskrnl/mm/ARM3/virtual.c
```diff
@@ -5447,16 +5451,24 @@ NtFreeVirtualMemory(IN HANDLE ProcessHandle,
...
-                    goto FailPath;
+                    MiLockProcessWorkingSetUnsafe(Process, CurrentThread);
+                    CommitReduction = MiCalculatePageCommitment(StartingAddress,
+                                                                EndingAddress,
+                                                                Vad,
+                                                                Process);
+                    Vad->u.VadFlags.CommitCharge -= CommitReduction;
+                    // For ReactOS: shrink the corresponding memory area
+                    ASSERT(Vad->StartingVpn == MemoryArea->VadNode.StartingVpn);
+                    ASSERT(Vad->EndingVpn == MemoryArea->VadNode.EndingVpn);
```
Extravert-ir, line 5469: Feels like, asserts in all 3 cases can be moved above to where `MemoryArea` is set

ntoskrnl/mm/ARM3/virtual.c
```diff
@@ -5472,8 +5484,8 @@ NtFreeVirtualMemory(IN HANDLE ProcessHandle,
                 //
                 if ((EndingAddress >> PAGE_SHIFT) == Vad->EndingVpn)
                 {
-                    PMEMORY_AREA MemoryArea;
-
+                    //
+                    // Case C (freeing a part at the end)
```
Extravert-ir, line 5488: Nit: also above `This is case B or case C. First check for case C` -> `This is case B or case C.`

Extravert-ir (approved):
LGTM

Extravert-ir:
Not related to the PR directly, but it looks like we're not reducing the process' `CommitCharge` in case D.
We just call `MiRemoveNode` which doesn't do anything with it (as far as I can see)

HBelusca (approved):

ntoskrnl/mm/ARM3/virtual.c
ThFabba: They're asserting that the before-state matches what we expect, so I'd rather have them near the state change. 🤷

ntoskrnl/mm/ARM3/virtual.c
ThFabba: Not the prevalent style anywhere around here so I'd rather not.

PR #4205, opened 2021-12-28
[REACTOS] Move our ROS-specific header-only UI libraries into a "ui" sub-folder.

HBelusca:
## Purpose

Move the ROS-specific header-only UI libraries (rosctrls.h, rosdlgs.h and layout.h) into a `ui` subfolder of `include/reactos/` so that they get sorted out of all the rest of the files already present.
And adjust the code that include these.


katahiromz (approved):
Okay.

GeoB99 (approved):

learn-more (approved):

wjk (commented):
Just curious: Why not use `rosui/rosctrls.h` instead?

HBelusca:
> Just curious: Why not use `rosui/rosctrls.h` instead? I think it would be clearer.

rosui.h needs some cleanup (see later PRs) to make it less RAPPS-specific, before being able to be used in other C++/ATL code. And also, you wouldn't use it in C projects.

binarymaster (approved):

PR #4207, opened 2021-12-28
[SDK:ATL] Implement GetDlgItemText(BSTR&), GetTopLevelParent(), GetWindowText(CSimpleString&).

HBelusca:
## Purpose

Implement few missing simple CWindow getters.
In particular, `GetWindowText(CSimpleString&)` is one missing that is needed by RAPPS (currently it's hacked into the RAPPS-specific CUiWindow class, in rosui.h).

_I suppose they are trivial enough for not necessitating any unit tests (and we don't have any for any of the CWindow functions in our `rostests/apitests/atl` module...). If you don't agree please let me know._

PR #4208, opened 2021-12-28
** WIP Refactoring aspects of rosui.h + Update RAPPS.

HBelusca:
## Purpose

Several aspects of rosui.h (the ATL-oriented UI layout manager for RAPPS) need to be cleaned up / reviewed before it could be used elsewhere. There appears to be dead or buggy code (see my comments in the review). Some other aspects of it can be simplified.

Finally I need some help from @gigaherz to explain to me some of his design choices he made back when he introduced the rosui.h library :)

### DIFF
--- a/base/applications/rapps/appview.cpp
+++ b/base/applications/rapps/appview.cpp
@@ -817,7 +817,7 @@ CAppInfoDisplay::ResizeChildren(int Width, int Height)
 
         if (hDwp)
         {
-            // hide the padding if scrnshot window width == 0
+            // hide the padding if screenshot window width == 0
             int RicheditPosX = ScrnshotWidth ? (ScrnshotWidth + INFO_DISPLAY_PADDING) : 0;
 
             hDwp = ::DeferWindowPos(hDwp, RichEdit->m_hWnd, NULL, RicheditPosX, 0, Width - RicheditPosX, Height, 0);
@@ -1411,7 +1411,7 @@ CApplicationView::ProcessWindowMessage(
         case WM_NOTIFY:
         {
             LPNMHDR pNotifyHeader = (LPNMHDR)lParam;
-            if (pNotifyHeader->hwndFrom == m_ListView->GetWindow())
+            if (pNotifyHeader->hwndFrom == m_ListView->m_hWnd)
             {
                 switch (pNotifyHeader->code)
                 {
@@ -1490,7 +1490,7 @@ CApplicationView::ProcessWindowMessage(
                     break;
                 }
             }
-            else if (pNotifyHeader->hwndFrom == m_Toolbar->GetWindow())
+            else if (pNotifyHeader->hwndFrom == m_Toolbar->m_hWnd)
             {
                 switch (pNotifyHeader->code)
                 {
@@ -1660,6 +1660,7 @@ CApplicationView::OnSize(HWND hwnd, WPARAM wParam, LPARAM lParam)
     }
 
     m_ComboBox->m_Margin.right = m_SearchBar->m_Width + m_SearchBar->m_Margin.right + TOOLBAR_PADDING;
+
     count = m_ComboBox->CountSizableChildren();
     hdwp = BeginDeferWindowPos(count);
     if (hdwp)
@@ -1677,7 +1678,7 @@ CApplicationView::OnCommand(WPARAM wParam, LPARAM lParam)
 {
     if (lParam)
     {
-        if ((HWND)lParam == m_SearchBar->GetWindow())
+        if ((HWND)lParam == m_SearchBar->m_hWnd)
         {
             CStringW szBuf;
             switch (HIWORD(wParam))
@@ -1727,7 +1728,7 @@ CApplicationView::OnCommand(WPARAM wParam, LPARAM lParam)
 
             return;
         }
-        else if ((HWND)lParam == m_ComboBox->GetWindow())
+        else if ((HWND)lParam == m_ComboBox->m_hWnd)
         {
             int NotifyCode = HIWORD(wParam);
             switch (NotifyCode)
@@ -1746,7 +1747,7 @@ CApplicationView::OnCommand(WPARAM wParam, LPARAM lParam)
 
             return;
         }
-        else if ((HWND)lParam == m_Toolbar->GetWindow())
+        else if ((HWND)lParam == m_Toolbar->m_hWnd)
         {
             // the message is sent from Toolbar. fall down to continue process
         }
@@ -1881,7 +1882,7 @@ CApplicationView::GetItemCount()
 }
 
 VOID
-CApplicationView::AppendTabOrderWindow(int Direction, ATL::CSimpleArray<HWND> &TabOrderList)
+CApplicationView::AppendTabOrderWindow(INT Direction, ATL::CSimpleArray<HWND>& TabOrderList)
 {
     m_Toolbar->AppendTabOrderWindow(Direction, TabOrderList);
     m_ComboBox->AppendTabOrderWindow(Direction, TabOrderList);
--- a/base/applications/rapps/gui.cpp
+++ b/base/applications/rapps/gui.cpp
@@ -748,7 +748,7 @@ CMainWindow::SearchTextChanged(CStringW &SearchText)
 }
 
 void
-CMainWindow::HandleTabOrder(int direction)
+CMainWindow::HandleTabOrder(INT direction)
 {
     ATL::CSimpleArray<HWND> TabOrderHwndList;
 
@@ -808,9 +808,8 @@ MainWindowLoop(CAppDB *db, INT nShowCmd)
         {
             if (Msg.message == WM_CHAR && Msg.wParam == VK_TAB)
             {
-                // Move backwards if shift is held down
-                int direction = (GetKeyState(VK_SHIFT) & 0x8000) ? -1 : 1;
-
+                /* Move backwards if shift is held down */
+                INT direction = (GetKeyState(VK_SHIFT) & 0x8000) ? -1 : 1;
                 wnd->HandleTabOrder(direction);
                 continue;
             }
--- a/base/applications/rapps/include/rosui.h
+++ b/base/applications/rapps/include/rosui.h
@@ -1,54 +1,58 @@
 /*
  * PROJECT:     ReactOS UI Layout Engine
  * LICENSE:     GPL-2.0+ (https://spdx.org/licenses/GPL-2.0+)
- * FILE:        base/applications/rapps/include/rosui.h
  * PURPOSE:     ATL Layout engine for RAPPS
  * COPYRIGHT:   Copyright 2015 David Quintana (gigaherz@gmail.com)
  */
+
 #pragma once
 
 #include <atlwin.h>
 
-template <class T, INT GrowthRate = 10> class CPointerArray
+//
+// "Basic" types / container(s)
+//
+
+template</*class TDeriv,*/ typename T, INT GrowthRate = 10>
+class CPointerArray
 {
-  protected:
+protected:
     HDPA m_hDpa;
 
-  public:
+public:
     CPointerArray()
     {
         m_hDpa = DPA_Create(GrowthRate);
     }
 
-    ~CPointerArray()
+    virtual ~CPointerArray()
     {
+        // Here, s_OnRemoveItem() can only call our own OnRemoveItem() implementation...
+        // unless we use the CRTP construct...
         DPA_DestroyCallback(m_hDpa, s_OnRemoveItem, this);
     }
 
-  private:
-    static INT CALLBACK
-    s_OnRemoveItem(PVOID ptr, PVOID context)
+private:
+    static INT CALLBACK s_OnRemoveItem(PVOID ptr, PVOID context)
     {
-        CPointerArray *self = (CPointerArray *)context;
-        return (INT)self->OnRemoveItem(reinterpret_cast<T *>(ptr));
+        // TDeriv* pThis = static_cast<TDeriv*>(context);
+        CPointerArray* pThis = (CPointerArray*)context;
+        return (INT)pThis->OnRemoveItem(reinterpret_cast<T*>(ptr));
     }
 
-    static INT CALLBACK
-    s_OnCompareItems(PVOID p1, PVOID p2, LPARAM lParam)
+    static INT CALLBACK s_OnCompareItems(PVOID p1, PVOID p2, LPARAM lParam)
     {
-        CPointerArray *self = (CPointerArray *)lParam;
-        return self->OnCompareItems(reinterpret_cast<T *>(p1), reinterpret_cast<T *>(p2));
+        CPointerArray* self = (CPointerArray*)lParam;
+        return self->OnCompareItems(reinterpret_cast<T*>(p1), reinterpret_cast<T*>(p2));
     }
 
-  public:
-    virtual BOOL
-    OnRemoveItem(T *ptr)
+public:
+    /*virtual*/ BOOL OnRemoveItem(T* ptr)
     {
         return TRUE;
     }
 
-    virtual INT
-    OnCompareItems(T *p1, T *p2)
+    virtual INT OnCompareItems(T* p1, T* p2)
     {
         INT_PTR t = (reinterpret_cast<INT_PTR>(p2) - reinterpret_cast<INT_PTR>(p1));
         if (t > 0)
@@ -58,45 +62,42 @@ template <class T, INT GrowthRate = 10> class CPointerArray
         return 0;
     }
 
-  public:
-    INT
-    GetCount() const
+public:
+    INT GetCount() const
     {
         return DPA_GetPtrCount(m_hDpa);
     }
 
-    T *
-    Get(INT i) const
+    T* Get(INT i) const
     {
-        return (T *)DPA_GetPtr(m_hDpa, i);
+        return (T*)DPA_GetPtr(m_hDpa, i);
     }
 
     BOOL
-    Set(INT i, T *ptr)
+    Set(INT i, T* ptr)
     {
         return DPA_SetPtr(m_hDpa, i, ptr);
     }
 
     INT
-    Insert(INT at, T *ptr)
+    Insert(INT at, T* ptr)
     {
         return DPA_InsertPtr(m_hDpa, at, ptr);
     }
 
     INT
-    Append(T *ptr)
+    Append(T* ptr)
     {
         return DPA_InsertPtr(m_hDpa, DA_LAST, ptr);
     }
 
     INT
-    IndexOf(T *ptr) const
+    IndexOf(T* ptr) const
     {
         return DPA_GetPtrIndex(m_hDpa, ptr);
     }
 
-    BOOL
-    Remove(T *ptr)
+    BOOL Remove(T* ptr)
     {
         INT i = IndexOf(ptr);
         if (i < 0)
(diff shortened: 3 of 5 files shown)


base/applications/rapps/appview.cpp
```diff
@@ -1524,7 +1524,7 @@ BOOL CApplicationView::ProcessWindowMessage(HWND hwnd, UINT message, WPARAM wPar
     case WM_NOTIFY:
     {
         LPNMHDR pNotifyHeader = (LPNMHDR)lParam;
-        if (pNotifyHeader->hwndFrom == m_ListView->GetWindow())
+        if (pNotifyHeader->hwndFrom == m_ListView->m_hWnd)
```
HBelusca, line 1527: Note that other parts of the code already access the **public member** `m_hWnd` of the different window objects, so having a `GetWindow()` accessor in the `CUiWindow` seems overkill.

base/applications/rapps/include/rosui.h
```diff
@@ -21,31 +25,34 @@ class CPointerArray
         m_hDpa = DPA_Create(GrowthRate);
     }
 
-    ~CPointerArray()
+    virtual ~CPointerArray()
     {
+        // Here, s_OnRemoveItem() can only call our own OnRemoveItem() implementation...
+        // unless we use the CRTP construct...
```
HBelusca, line 31: OK here and below (see `s_OnRemoveItem()` and the `class CUiCollection` implementation), we face the problem that, at the time the `~CPointerArray()` destructor is called, the `s_OnRemoveItem()` helper (and then, the `OnRemoveItem() method) being called will **never ever** call any overridden implementation of `OnRemoveItem()`, because the derived class(es) have already been destructed.

(A).
I think here, the idea was to have any derived classes from the template-specialized `CPointerArray` to have a custom "object" deleter. (I would like to know @gigaherz opinion on this.)
I'm not sure this is of great use, unless there is also some sort of callback during the `Set/Insert/Append` calls to balance this. My point being, at least as far as this class (and the `CUiCollection` that derives from it) is concerned in RAPPS, UI objects are first allocated and stored separately, then pointer to them are added into this pointer-array, but, these objects are later manually deleted one after the other in RAPPS code implementation. And then the `CUiCollection` (and therefore the underlying `CPointerArray`) instance(s) get destructed, and at this point we don't really need the "collection"/array to do the deletion for us (see more on that later in this comment).

(B).
In order to have the `CPointerArray` template class be able, from this `s_OnRemoveItem` helper, to call the overridden implementation from the derived class (`CUiCollection` in our case), the only simple technique I've found is this so-called _Curiously recurring template pattern_ (CRTP), aka. what the ATL classes do, to specify as the first template "parameter" the class being derived, so that the specialization of the base class "knows" what derived class (and its methods) to access.
This is the reason of the (currently commented-out) `TBase` parameter you can see above in the definition, and the commented `TDeriv* pThis = static_cast<TDeriv*>(context);` line in `s_OnRemoveItem()`.
Then, `CUiCollection` needs to be defined instead as:
```c++
class CUiCollection :
    public CPointerArray<CUiCollection, CUiPrimitive>
```
and the whole construct will work: the overridden `CUiCollection::OnRemoveItem()` helper will be called during destruction.

However, the currently suggested implementation of `CUiCollection::OnRemoveItem()` (to explicitly `delete` the passed pointer) will be inadequate, due to the reason I've given in point (A). Indeed, with the modifications explained above, if we keep the current implementation of deleting the passed pointers, then we trigger a crash, since we would delete a pointer to a window object, that has already been separately deleted by some RAPPS cleanup code...

base/applications/rapps/include/rosui.h
```diff
@@ -21,31 +25,34 @@ class CPointerArray
...
     static INT CALLBACK s_OnCompareItems(PVOID p1, PVOID p2, LPARAM lParam)
     {
-        CPointerArray * self = (CPointerArray*) lParam;
+        CPointerArray* self = (CPointerArray*)lParam;
         return self->OnCompareItems(reinterpret_cast<T*>(p1), reinterpret_cast<T*>(p2));
     }
 
 public:
-    virtual BOOL OnRemoveItem(T * ptr)
+    /*virtual*/ BOOL OnRemoveItem(T* ptr)
```
HBelusca, line 50: The commenting of `virtual` was part of my attempts at studying the above-mentioned problem.

base/applications/rapps/include/rosui.h
```diff
@@ -97,17 +104,14 @@ class CPointerArray
...
         PVOID ptr = DPA_DeletePtr(m_hDpa, i);
-        if (ptr != NULL)
-        {
-            OnRemoveItem(reinterpret_cast<T*>(ptr));
-            return TRUE;
-        }
-        return FALSE;
+        if (!ptr)
+            return FALSE;
+        return OnRemoveItem(reinterpret_cast<T*>(ptr));
```
HBelusca, line 109: This makes the code slightly more concise and makes the return value of `RemoveAt` more accurate.

base/applications/rapps/include/rosui.h
```diff
@@ -97,17 +104,14 @@ class CPointerArray
...
     }
 
     BOOL Clear()
     {
-        DPA_EnumCallback(s_OnRemoveItem, this);
+        DPA_EnumCallback(m_hDpa, s_OnRemoveItem, this);
```
HBelusca, line 114: Missing parameter, this was an unseen compilation error because `Clear()` was never used in RAPPS.

base/applications/rapps/include/rosui.h
```diff
@@ -116,46 +120,37 @@ class CPointerArray
         return DPA_Sort(m_hDpa, s_OnCompareItems, (LPARAM)this);
     }
 
-    INT Search(T* item, INT iStart, UINT uFlags)
+    INT Search(T* item, INT iStart = 0, UINT uFlags = 0)
     {
-        return DPA_Search(m_hDpa, item, 0, s_OnCompareItems, (LPARAM)this, 0);
+        return DPA_Search(m_hDpa, item, iStart, s_OnCompareItems, (LPARAM)this, uFlags);
```
HBelusca, line 125: Make use of the `iStart` and `uFlags` parameters, giving them sane default values.

base/applications/rapps/include/rosui.h
```diff
@@ -116,46 +120,37 @@ class CPointerArray
...
     }
 };
 
-class CUiRect
```
HBelusca, line 125: OK this is a class that doesn't serve any purpose, since it just duplicates the `RECT` structure (and that structure is manipulated anyways anywhere else in the code).
Since this rosui.h header is also used exclusively in ATL apps it could, at best, use the `CRect` structure from `atltypes.h`, in case of need.

base/applications/rapps/include/rosui.h
```diff
@@ -116,46 +120,37 @@ class CPointerArray
...
-        bottom = b;
-    }
-};
+//
+// Units/Ways of UI measure
+//
 
 class CUiMargin
-    : public CUiRect
+    : public RECT // CRect
```
HBelusca, line 135: `CUiMargin` now directly inherits from the basic `RECT` structure. I have kept (commented) the code that would be needed if instead we make it derive from ATL's `CRect` class.
Suggestions are welcome.

base/applications/rapps/include/rosui.h
```diff
@@ -175,16 +170,14 @@ class CUiMeasure
     INT m_Value;
 
 public:
-    CUiMeasure()
+    CUiMeasure() :
```
HBelusca, line 173: In both constructors I went fancy and used initialization lists instead.

base/applications/rapps/include/rosui.h
```diff
@@ -200,10 +193,10 @@ class CUiMeasure
         case Type_FitParent:
             return parent;
         }
-
-        return 0;
+        UNREACHABLE;
```
HBelusca, line 196: That's indeed unreachable since the `switch` statement covers all four possibilities for type.

base/applications/rapps/gui.cpp
```diff
@@ -891,9 +890,8 @@ VOID MainWindowLoop(INT nShowCmd)
             if (Msg.message == WM_CHAR &&
                 Msg.wParam == VK_TAB)
             {
-                // Move backwards if shift is held down
-                int direction = (GetKeyState(VK_SHIFT) & 0x8000) ? -1 : 1;
-
+                /* Move backwards if shift is held down */
+                INT direction = (GetKeyState(VK_SHIFT) & 0x8000) ? -1 : 1;
```
katahiromz, line 894:
```suggestion
                INT direction = (GetKeyState(VK_SHIFT) < 0) ? -1 : 1;
```

base/applications/rapps/include/rosui.h
```diff
@@ -1,15 +1,19 @@
...
 #pragma once
 
 #include <atlwin.h>
 
-template<class T, INT GrowthRate = 10>
+//
+// "Basic" types / container(s)
+//
+
+template</*class TDeriv,*/ typename T, INT GrowthRate = 10>
```
katahiromz, line 16:
```suggestion
template<typename T, INT GrowthRate = 10>
```

base/applications/rapps/include/rosui.h
```diff
@@ -302,54 +300,78 @@ class CUiBox
         *newRect = currentRect;
     }
 
-
 public:
+    //virtual ~CUiBox() {} // Needed for current CUiCollection implementation.
```
HBelusca, line 304: This is a leftover of tests back when I made `CUiBox` derive from `CUiCollection` in order to simplify code later on (regarding the `AsBox()` returned values checks).

base/applications/rapps/include/rosui.h
```diff
@@ -302,54 +300,78 @@ class CUiBox
...
+    }
 
+    // Override in subclass
     virtual HDWP OnParentSize(RECT parentRect, HDWP hDwp)
     {
-        // Override in subclass
-        return NULL;
-    };
+        /* Pass hDwp down the call chain */
+        return hDwp;
```
HBelusca, line 328: This was a bug, I think.

base/applications/rapps/include/rosui.h
```diff
@@ -21,31 +25,34 @@ class CPointerArray
...
         DPA_DestroyCallback(m_hDpa, s_OnRemoveItem, this);
     }
 
 private:
     static INT CALLBACK s_OnRemoveItem(PVOID ptr, PVOID context)
     {
-        CPointerArray * self = (CPointerArray*) context;
-        return (INT) self->OnRemoveItem(reinterpret_cast<T*>(ptr));
+        // TDeriv* pThis = static_cast<TDeriv*>(context);
+        CPointerArray* pThis = (CPointerArray*)context;
```
katahiromz, line 39: `reinterpret_cast<CPointerArray *>`???

base/applications/rapps/include/rosui.h
```diff
@@ -302,54 +300,78 @@ class CUiBox
...
+    // Override in subclass
     virtual VOID ComputeContentBounds(RECT* rect)
     {
-        // Override in subclass
-    };
+    }
 
-    virtual DWORD_PTR CountSizableChildren()
+    // Override in subclass
+    virtual INT CountSizableChildren()
```
HBelusca, line 319: I want to uniformly change the returned type of this function (here and in the derived classes and where it gets used) to `UINT`, as it is obvious a number of objects won't be negative.

base/applications/rapps/include/rosui.h
```diff
@@ -21,31 +25,34 @@ class CPointerArray
...
-        return (INT) self->OnRemoveItem(reinterpret_cast<T*>(ptr));
+        // TDeriv* pThis = static_cast<TDeriv*>(context);
+        CPointerArray* pThis = (CPointerArray*)context;
+        return (INT)pThis->OnRemoveItem(reinterpret_cast<T*>(ptr));
     }
 
     static INT CALLBACK s_OnCompareItems(PVOID p1, PVOID p2, LPARAM lParam)
     {
-        CPointerArray * self = (CPointerArray*) lParam;
+        CPointerArray* self = (CPointerArray*)lParam;
```
katahiromz, line 45: `reinterpret_cast<CPointerArray *>`???

base/applications/rapps/include/rosui.h
```diff
@@ -374,173 +396,150 @@ class CUiPanel :
...
     {
-        if (T::IsWindow())
-        {
-            T::DestroyWindow();
-        }
+        UNREFERENCED_PARAMETER(Direction);
+        TabOrderList.Add(TBase::m_hWnd);
     }
 
-    VOID GetWindowTextW(ATL::CStringW& szText)
```
HBelusca, line 514: This is the ATL-ish function that becomes unnecessary after PR #4207.

base/applications/rapps/include/rosui.h
```diff
@@ -21,31 +25,34 @@ class CPointerArray
...
-    virtual BOOL OnRemoveItem(T * ptr)
+    /*virtual*/ BOOL OnRemoveItem(T* ptr)
     {
         return TRUE;
     }
 
-    virtual INT OnCompareItems(T * p1, T * p2)
+    virtual INT OnCompareItems(T* p1, T* p2)
     {
         INT_PTR t = (reinterpret_cast<INT_PTR>(p2) - reinterpret_cast<INT_PTR>(p1));
```
katahiromz, line 57:
```c
if (p1 < p2)
    return 1;
if (p1 > p2)
    return -1;
return 0;
```

base/applications/rapps/include/rosui.h
```diff
@@ -63,7 +70,7 @@ class CPointerArray
 
     T* Get(INT i) const
     {
-        return (T*) DPA_GetPtr(m_hDpa, i);
+        return (T*)DPA_GetPtr(m_hDpa, i);
```
katahiromz, line 73:
```suggestion
        return reinterpret_cast<T*>(DPA_GetPtr(m_hDpa, i));
```

base/applications/rapps/include/rosui.h
```diff
@@ -374,173 +396,150 @@ class CUiPanel :
...
+    virtual ~CUiPanel() {}
 
-    virtual CUiBox * AsBox() { return this; }
+    // virtual CUiBox* AsBox() { return this; }
 
     virtual VOID ComputeMinimalSize(SIZE* size)
     {
         for (INT i = 0; i < m_Children.GetCount(); i++)
         {
-            CUiBox * box = m_Children.Get(i)->AsBox();
```
HBelusca, line 387: Here (and elsewhere) I have the impression that this indirection looks quite unnecessary.

Indeed: why would we have elements in the children collection of `CUiPrimitive`s that would not be measurable UI boxes? In what sort of situations this could happen? (I'm asking in a generic manner, not in regards to the current usage of this UI library by RAPPS). What sort of so-called `CUiPrimitive`s could be added into the children collection, yet not have a box? Would this make any sense?
Cc @gigaherz for any clues...

base/applications/rapps/include/rosui.h
```diff
@@ -374,173 +396,150 @@ class CUiPanel :
...
     public CUiPrimitive,
-    public CUiBox,
-    public T
+    // public CUiBox,
+    public TBase
 {
 public:
-    virtual CUiBox * AsBox() { return this; }
-
-    HWND GetWindow() { return T::m_hWnd; }
```
HBelusca, line 458: This is the function that is not necessary since we have (by default) `m_hWnd` that is public (by ATL's `CWindow` or derived classes), and the implementation of `CUiWindow` makes it implicit, by the usage of this member as well as other window functions, that we will only accept template specialization with other classes that "look like" ATL's `CWindow`.

base/applications/rapps/include/rosui.h
```diff
@@ -302,54 +300,78 @@ class CUiBox
...
         delete ptr;
         return TRUE;
     }
+#endif
 };
 
+//
+// Question: Does it need more? If not, we could just use
+// CUiCollection's directly, in lieu of CUiContainer's.
+//
```
HBelusca, line 374: So yeah, why do we need to have this distinction of UI "container" and UI "collection" ?
Here it seems apparent they would be the same. What was the reason behind this distinction @gigaherz ?

base/applications/rapps/include/rosui.h
```diff
@@ -302,54 +300,78 @@ class CUiBox
...
 
 class CUiCollection :
-    public CPointerArray < CUiPrimitive >
+    public CPointerArray</*CUiCollection,*/ CUiPrimitive>
 {
-    virtual BOOL OnRemoveItem(CUiPrimitive * ptr)
+public:
+    virtual ~CUiCollection() {}
+
+#if 0
```
HBelusca, line 362: This is related to the problem I (extensively) exposed in the `CPointerArray` class.

base/applications/rapps/include/rosui.h
```diff
@@ -302,54 +300,78 @@ class CUiBox
...
+#define CUiPrimitive CUiBox
+#endif
 
 class CUiCollection :
-    public CPointerArray < CUiPrimitive >
+    public CPointerArray</*CUiCollection,*/ CUiPrimitive>
 {
-    virtual BOOL OnRemoveItem(CUiPrimitive * ptr)
+public:
+    virtual ~CUiCollection() {}
```
HBelusca, line 360: The presence or absence of this constructor doesn't seem to affect anything regarding the problem exposed in the `CPointerArray` class.

base/applications/rapps/include/rosui.h
```diff
@@ -302,54 +300,78 @@ class CUiBox
...
+// This is actually an interface of sorts...
+// Actually it could be helpful if it's a reference count
+// so that the UiCollection could implement the ref uncount and delete if needed.
+//
+class CUiPrimitive : public CUiBox
 {
 protected:
-    CUiPrimitive * m_Parent;
+#if 0
+    CUiPrimitive* m_Parent;
```
HBelusca, line 344: For what purpose would this be used?

base/applications/rapps/appview.cpp
katahiromz, line 1527:
```suggestion
        if (pNotifyHeader->hwndFrom == m_ListView)
```

base/applications/rapps/appview.cpp
```diff
@@ -1604,7 +1604,7 @@ BOOL CApplicationView::ProcessWindowMessage(HWND hwnd, UINT message, WPARAM wPar
             break;
             }
         }
-        else if (pNotifyHeader->hwndFrom == m_Toolbar->GetWindow())
+        else if (pNotifyHeader->hwndFrom == m_Toolbar->m_hWnd)
```
katahiromz, line 1607:
```suggestion
        else if (pNotifyHeader->hwndFrom == m_Toolbar)
```

base/applications/rapps/appview.cpp
```diff
@@ -1782,7 +1782,7 @@ VOID CApplicationView::OnCommand(WPARAM wParam, LPARAM lParam)
 {
     if (lParam)
     {
-        if ((HWND)lParam == m_SearchBar->GetWindow())
+        if ((HWND)lParam == m_SearchBar->m_hWnd)
```
katahiromz, line 1785:
```suggestion
        if (reinterpret_cast<HWND>(lParam) == m_SearchBar)
```

base/applications/rapps/appview.cpp
```diff
@@ -1832,7 +1832,7 @@ VOID CApplicationView::OnCommand(WPARAM wParam, LPARAM lParam)
 
             return;
         }
-        else if ((HWND)lParam == m_ComboBox->GetWindow())
+        else if ((HWND)lParam == m_ComboBox->m_hWnd)
```
katahiromz, line 1835:
```suggestion
        else if (reinterpret_cast<HWND>(lParam) == m_ComboBox)
```

base/applications/rapps/appview.cpp
```diff
@@ -1851,7 +1851,7 @@ VOID CApplicationView::OnCommand(WPARAM wParam, LPARAM lParam)
 
             return;
         }
-        else if ((HWND)lParam == m_Toolbar->GetWindow())
+        else if ((HWND)lParam == m_Toolbar->m_hWnd)
```
katahiromz, line 1854:
```suggestion
        else if (reinterpret_cast<HWND>(lParam) == m_Toolbar)
```

base/applications/rapps/include/rosui.h
katahiromz, line 39:
```suggestion
        auto pThis = reinterpret_cast<CPointerArray*>(context);
```

base/applications/rapps/include/rosui.h
katahiromz, line 45:
```suggestion
        auto self = reinterpret_cast<CPointerArray*>(lParam);
```

base/applications/rapps/include/rosui.h
```diff
@@ -302,54 +300,78 @@ class CUiBox
...
-class CUiPrimitive
+
+
+#if 1
+//
+// This is actually an interface of sorts...
+// Actually it could be helpful if it's a reference count
+// so that the UiCollection could implement the ref uncount and delete if needed.
+//
+class CUiPrimitive : public CUiBox
```
HBelusca, line 340: This is related also to what the `CUiCollection` / `CUiContainer` classes can contain: What is, in this context, a "UI primitive", and why would instances of it be present in these collections/containers that are used by e.g. `CUiPanel` , if some of these are not measurable ?
Does making these inherit from `CUiBox` makes sense, or does there exist some situations where one would not want that ?
Cc @gigaherz for some clues about this.

base/applications/rapps/include/rosui.h
HBelusca: Well, this was about the question I had about this class, with regards to the `CUiContainer` (see large comment down below)

base/applications/rapps/gui.cpp
HBelusca: No, the `& 0x8000` is the actual test (test for bit flag)

base/applications/rapps/appview.cpp
HBelusca: This suggestion doesn't work (I tested it)

base/applications/rapps/appview.cpp
HBelusca: The reinterpret cast on the scalar lParam is not necessary.

base/applications/rapps/include/rosui.h
HBelusca: I won't use `auto` as I want this code to be compilable with e.g. MSVC 2010.

base/applications/rapps/include/rosui.h
HBelusca: ditto.

base/applications/rapps/gui.cpp
katahiromz: The return value of `GetKeyState` is a `SHORT` (a 16-bit signed integer). `0x8000` means its most significant bit (MSB).

base/applications/rapps/appview.cpp
learn-more: I don't like this change, is it just changing code for the sake of changing it?
What actual problem does this solve?

base/applications/rapps/include/rosui.h
learn-more: It does not duplicate the RECT structure, it provides c++ initialization for the RECT structure.

base/applications/rapps/include/rosui.h
learn-more: no, it was telling to implement it in a subclass, not to use it as-is.

JoachimHenze:
Imho rapps was better, more modular and the binary was much smaller back when it was just using plain C and no fancy cpp rosui library.

base/applications/rapps/include/rosui.h
learn-more: If you wanna refactor, nuke the `CPointerArray` and use either `CSimpleArray` or `CArray`,
both should delete their items.
The whole DPA_XXX stuff is a thing you would use when writing in C, and you don't want to write too much boilerplate.

base/applications/rapps/include/rosui.h
HBelusca: Then it should be possible to just use the CRect from ATL? (but again, this doesn't look like it's necessary to be used as other code just uses the RECT without any problems)

base/applications/rapps/include/rosui.h
HBelusca: Sure, however it's a default implementation, and suppose that for some reason a UI object doesn't need resizing, then it wouldn't have to override that function, and would just need to pass the resize-handle down the call chain (that would be a sane behaviour). With the old code, it would need to override the function just to do that (even if it doesn't need to implement its resize), otherwise it would break the defer-window handle passing around.

base/applications/rapps/include/rosui.h
HBelusca: I hope this wouldn't call some overriden `operator<` or `operator>` of the `T` class.

DarkFire01:
As this seems to be stalled for now, feel free to reopen when you wish for it to continue. 

Thank you!

PR #4211, opened 2021-12-28
[MSAFD] Handle option levels properly in WSPGetSockOpt(), following WSPSetSockOpt()

vicmarcal:
:: WSPGetSockOpt()

-Handle option levels properly in WSPGetSockOpt following the WSPSetSockOpt

## Purpose

I've reviewed all the patches in this following JIRA:

JIRA issue: [CORE-12104](https://jira.reactos.org/browse/CORE-12104)

In order to see if everything is already in master and close the Jira report if so.

I detected a patch which seems to be missing in master being the counterpart of the WSPSetSockOpt.
The creator is "tambre", according to the Jira History, 2 years later, and whom shares same location as the author Peter Hater.

I've recreated the patch for Git and current master. All the other patches seems to be included in master.

## Proposed changes

This seems to properly manage option levels in the Get counterpart. Now Get and Set shares the same structure, and lets lower options to be properly handled with a default case which can help to understand which other options could be used.

### DIFF
--- a/dll/win32/msafd/misc/dllmain.c
+++ b/dll/win32/msafd/misc/dllmain.c
@@ -2752,10 +2752,19 @@ WSPGetSockOpt(IN SOCKET Handle,
             RtlCopyMemory(OptionValue, Buffer, BufferSize);
 
             return 0;
+        /* These are handled at a lower level */
+        case IPPROTO_IP:
+        case IPPROTO_IPV6:
+        case IPPROTO_TCP:
+        case IPPROTO_UDP:
+            goto SendToHelper;
 
         default:
+        {
+            DbgPrint("Invalid option level (%x)\n", Level);
             if (lpErrno) *lpErrno = WSAEINVAL;
             return SOCKET_ERROR;
+        }
     }
 
 SendToHelper:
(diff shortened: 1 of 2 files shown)


binarymaster (changes requested):
Please follow our current [commit style](https://github.com/reactos/reactos/blob/master/.gitmessage).

Only `[MSAFD]` on the first line of commit message is not descriptive.

dll/win32/msafd/misc/dllmain.c
```diff
@@ -2752,10 +2752,19 @@ WSPGetSockOpt(IN SOCKET Handle,
             RtlCopyMemory(OptionValue, Buffer, BufferSize);
 
             return 0;
+        /* These are handled at a lower level */
```
HBelusca, line 2755:
```suggestion

        /* These are handled at a lower level */
```

dll/win32/msafd/misc/dllmain.c
```diff
@@ -2796,166 +2805,181 @@ WSPSetSockOpt(
...
+                case SO_KEEPALIVE:
+                case SO_DONTROUTE:
+                    /* These go directly to the helper dll */
+                    goto SendToHelper;
+
+                default:
+                    /* Obviously this is a hack */
+                    ERR("MSAFD: Set SOL_SOCKET unknown optname %x\n", optname);
+                    return NO_ERROR;
+            }
```
HBelusca, line 2968: This bracket is the end of the `switch ()` statement from the `case SOL_SOCKET:`
Why isn't there any "break" here at the end? Is it because we have to fallback into these other cases that just `goto SendToHelper;` ? or ? (Note that the `WSPGetSockOpt` function does **NOT** make its `case SOL_SOCKET:` handling ultimately fall back to these cases.)

dll/win32/msafd/misc/dllmain.c
```diff
@@ -2796,166 +2805,181 @@ WSPSetSockOpt(
...
+        case IPPROTO_TCP:
+        case IPPROTO_UDP:
+            goto SendToHelper;
+
+        default:
+        {
+            DbgPrint("Invalid option level (%x)\n", level);
+            if (lpErrno) *lpErrno = WSAEINVAL;
+            return SOCKET_ERROR;
+         }
```
HBelusca, line 2982: Indentation there is not terrible.

vicmarcal:
> Please follow our current [commit style](https://github.com/reactos/reactos/blob/master/.gitmessage).
> 
> Only `[MSAFD]` on the first line of commit message is not descriptive.

Thanks! The finger slipped...

dll/win32/msafd/misc/dllmain.c
vicmarcal: Perfect! Thanks!

dll/win32/msafd/misc/dllmain.c
vicmarcal: Nice appreciation. So the case SOL_SOCKET in WSPSetSockOpt has an internal switch(optname), all the cases handled in switch(optname) ends returning something or gotoing, so in case SOL_SOCKET we shouldn't be falling to its next sibling case (line 2972). 

If you wish we can add a "return 0" or a "break;" to enhace readibility but Coverity will sign it as unreachable code because all the previous switch..case paths are closed. (If you see a not-closed path, please tell me)

Regarding WSPGetSockOpt, iiuc, SOL_SOCKET won't fall into IPPROTO_IP because the line 2574 returns 0.

dll/win32/msafd/misc/dllmain.c
vicmarcal: Thanks! Fixing!

PR #4214, opened 2021-12-29
[RAPPS] Centralize the names of the RAPPS' registry setting sub-key, and user local settings & offline apps database sub-directories.

HBelusca:
## Purpose

The primary goal is the replace (in user's local AppData) `rapps\rapps`, which looks a bit ridiculous and could appear as a bug of RAPPS (why creating two subdirectories with the same name??), by instead `RAPPS\AppsDb`, which (to my opinion) looks more decent and tells exactly what it is for -- the offline apps database. This would also allow for future sub-directories into the `RAPPS` settings folder.

And also I take the opportunity to consolidate in one single place the defines for these directory names.

## Proposed changes

- Change the name of the application database sub-directory from "rapps\rapps" to "RApps\appdb".
- Adjust the documentation to reflect this.
- Simplify CConfigParser constructor by removing GetINIFullPath().

NOTE: In case this is accepted, https://reactos.org/wiki/RAPPS will need minor adjustments as well (following the README update).

### DIFF
--- a/base/applications/rapps/configparser.cpp
+++ b/base/applications/rapps/configparser.cpp
@@ -7,6 +7,7 @@
  *              Copyright 2017 Alexander Shaposhnikov (sanchaez@reactos.org)
  *              Copyright 2021 Mark Jansen <mark.jansen@reactos.org>
  */
+
 #include "rapps.h"
 #include <debug.h>
 
@@ -24,19 +25,7 @@ struct CSectionNames
 };
 static CSectionNames g_Names;
 
-static CStringW
-GetINIFullPath(const CStringW &FileName)
-{
-    CStringW szDir;
-    CStringW szBuffer;
-
-    GetStorageDirectory(szDir);
-    szBuffer.Format(L"%ls\\rapps\\%ls", szDir.GetString(), FileName.GetString());
-
-    return szBuffer;
-}
-
-CConfigParser::CConfigParser(const CStringW &FileName) : szConfigPath(GetINIFullPath(FileName))
+CConfigParser::CConfigParser(const CStringW &FilePath) : szConfigPath(FilePath)
 {
     CacheINI();
 }
--- a/base/applications/rapps/include/defines.h
+++ b/base/applications/rapps/include/defines.h
@@ -29,6 +29,15 @@
 #include "resource.h"
 #include "winmain.h"
 
-#define APPLICATION_DATABASE_URL L"https://rapps.reactos.org/rappmgr2.cab"
-#define APPLICATION_DATABASE_NAME L"rappmgr2.cab"
+/* Name of the registry sub-key where RAPPS settings are stored, and
+ * of the local AppData sub-directory where the RAPPS files are stored. */
+#define RAPPS_NAME                  L"RApps"
+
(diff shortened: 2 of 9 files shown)


base/applications/rapps/configparser.cpp
```diff
@@ -28,11 +29,9 @@ static CSectionNames g_Names;
 static
 ATL::CStringW GetINIFullPath(const ATL::CStringW& FileName)
 {
-    ATL::CStringW szDir;
-    ATL::CStringW szBuffer;
-
-    GetStorageDirectory(szDir);
-    szBuffer.Format(L"%ls\\rapps\\%ls", szDir.GetString(), FileName.GetString());
+    ATL::CStringW szBuffer = CAvailableApps::m_Strings.szAppsPath;
+    PathAppendW(szBuffer.GetBuffer(MAX_PATH), FileName.GetString());
```
learn-more, line 33: This re-introduces a MAX_PATH limit.

base/applications/rapps/include/defines.h
```diff
@@ -29,9 +29,18 @@
 #include "resource.h"
 #include "winmain.h"
 
-#define APPLICATION_DATABASE_URL L"https://rapps.reactos.org/rappmgr2.cab"
-#define APPLICATION_DATABASE_NAME L"rappmgr2.cab"
-#define MAX_STR_LEN              256
+/* Name of the registry sub-key where RAPPS settings are stored, and
+ * of the local AppData sub-directory where the RAPPS files are stored. */
+#define RAPPS_NAME                  L"RAPPS"
```
learn-more, line 34: I prefer `rapps` or `RApps`

learn-more (dismissed):
I don't like the re-introduction of the `MAX_PATH` limit,
other than that lgtm.

base/applications/rapps/include/defines.h
JoachimHenze: RApps looks like a typo to my eyes (although I know it maybe wasn't), what about capital R, small apps:
Rapps

base/applications/rapps/configparser.cpp
```diff
@@ -28,11 +29,8 @@ static CSectionNames g_Names;
 static
 ATL::CStringW GetINIFullPath(const ATL::CStringW& FileName)
 {
-    ATL::CStringW szDir;
-    ATL::CStringW szBuffer;
-
-    GetStorageDirectory(szDir);
-    szBuffer.Format(L"%ls\\rapps\\%ls", szDir.GetString(), FileName.GetString());
+    ATL::CStringW szBuffer = CAvailableApps::m_Strings.szAppsPath;
+    szBuffer += L"\\" + FileName;
```
learn-more, line 33: What problem does this solve?

HBelusca: In your previous comment you complained that I used `PathAppendW` that necessitated the explicit `MAX_PATH` limit.

learn-more: yes, but the original code worked fine?

HBelusca: The original code part that was concatenating `szDir` + the hardcoded `rapps` was in effect reconstructing the already-constructed `szAppsPath`.

PR #4215, opened 2021-12-29
[ATTRIB] De-duplicate code; handle reparse points; verify path concatenation

HBelusca:
## Purpose and Proposed changes

- Simplify duplicated code by having one EnumFiles() function using
  callbacks to perform actual action on enumerated files.

  PrintAttribute() and ChangeAttribute() duplicate their functionality,
  except for the actual difference of displaying the file attributes vs.
  changing them.
  Instead, move that per-file action into two callbacks, create a file
  enumerator function that contains the common functionality, and make
  it call the callback passed in parameter to it.

- Reparse points can be encountered during enumeration. In this case,
  don't try to go into too many reparse-point levels (maximum two).

- Verify that paths/file names concatenation don't overflow (with the
  StringCch* functions returning an error), as this could cause an
  infinite loop in case of recursive-enumeration, and the position where
  the new path part would be appended turns out to be just at the end
  of the buffer.

## TODO

- [x] Optimize the stack usage of the recursive enumerator function.
- [x] Finish using string-safe functions.

### DIFF
--- a/base/applications/cmdutils/attrib/attrib.c
+++ b/base/applications/cmdutils/attrib/attrib.c
@@ -1,35 +1,10 @@
 /*
- *  ATTRIB.C - attrib internal command.
- *
- *
- *  History:
- *
- *    04-Dec-1998 Eric Kohl
- *        started
- *
- *    09-Dec-1998 Eric Kohl
- *        implementation works, except recursion ("attrib /s").
- *
- *    05-Jan-1999 Eric Kohl
- *        major rewrite.
- *        fixed recursion ("attrib /s").
- *        started directory support ("attrib /s /d").
- *        updated help text.
- *
- *    14-Jan-1999 Eric Kohl
- *        Unicode ready!
- *
- *    19-Jan-1999 Eric Kohl
- *        Redirection ready!
- *
- *    21-Jan-1999 Eric Kohl
- *        Added check for invalid filenames.
- *
- *    23-Jan-1999 Eric Kohl
- *        Added handling of multiple filenames.
- *
- *    02-Apr-2005 (Magnus Olsen <magnus@greatlord.com>)
- *        Remove all hardcoded strings in En.rc
+ * PROJECT:     ReactOS Attrib Command
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Displays or changes file attributes recursively.
+ * COPYRIGHT:   Copyright 1998-2019 Eric Kohl <eric.kohl@reactos.org>
+ *              Copyright 2021 Doug Lyons <douglyons@douglyons.com>
+ *              Copyright 2021-2023 Hermès Bélusca-Maïto <hermes.belusca-maito@reactos.org>
  */
 
 #include <stdio.h>
@@ -39,15 +14,22 @@
 #include <winbase.h>
 #include <wincon.h>
 #include <winuser.h>
+#include <strsafe.h>
 
 #include <conutils.h>
 
 #include "resource.h"
 
+/* Enable to support extended attributes.
+ * See https://ss64.com/nt/attrib.html for an exhaustive list. */
+// TODO: If you enable this, translations need to be updated as well!
+//#define EXTENDED_ATTRIBUTES
+
+#define ALL_FILES_PATTERN   L"*.*" // It may also be possible to use L"*" (shorter)
+
 CON_SCREEN StdOutScreen = INIT_CON_SCREEN(StdOut);
 
-static
-VOID
+static VOID
 ErrorMessage(
     _In_ DWORD dwErrorCode,
     _In_opt_ PCWSTR pszMsg,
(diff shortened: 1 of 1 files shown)


base/applications/cmdutils/attrib/attrib.c
```diff
@@ -274,14 +239,15 @@ ChangeAttribute(
             continue;
         }
 
+// FIXME: This was done for ChangeAttribute(), but not for PrintAttribute() ??
+#if 0
         if (bRecurse && bIsDir && !bDirectories)
             continue;
+#endif
```
HBelusca, line 246: @Doug-Lyons : Do you happen to know why (from your committed PR), the `ChangeAttribute()` function had this extra piece of code, while the `PrintAttribute()` function did not?

Doug-Lyons: @HBelusca, It has been some time since I have worked on this code, so I am not sure that my memory is correct on this. But I believe that this was related to Thomas Fabers comment: "Directories seem to be excluded whenever wildcards are used, so attributes are only applied to them if mentioned specifically by name"(https://github.com/reactos/reactos/pull/2142). I think that the regression tests will reflect the correct usage, so running your changes against the cmd:attrib test should tell you whether your changes are correct or not.

Doug-Lyons: @HBelusca, I added your "if 0/#endif" around these two lines in my source for attrib.c and ran the regression test for cmd:attrib. These new results show 12 failures with this code removed as you can see here: 

new_attrib_fails_01

HBelusca: > I added your "if 0/#endif" around these two lines in my source for attrib.c 

In the one currently in master?

And out of curiosity, what would happen instead if you were to **add** this extra code (that is #if-0'ed in this PR), not only in `ChangeAttribute()` as it currently is, but in `PrintAttribute()` as well?

Doug-Lyons: @HBelusca, Yes, I tested this on the 'attrib.c' file that is currently in master. I changed the code as you asked and added this additional code to `PrintAttributes()` and it seems not to cause any failures. So it looks safe to add it. Here is the result of running with the added code:

new_attrib_OK_01

HBelusca: Thanks for the tests!

base/applications/cmdutils/attrib/attrib.c
```diff
@@ -40,10 +14,20 @@
...
 
 #include "resource.h"
 
+#define DEBUG_TMP
+
+/* Enable to support extended attributes.
+ * See https://ss64.com/nt/attrib.html for an exhaustive list. */
+//#define EXTENDED_ATTRIBUTES
+
+#define ALL_FILES_PATTERN   L"*.*" // L"*"
```
learn-more, line 29: Debug leftover?

base/applications/cmdutils/attrib/attrib.c
```diff
@@ -77,217 +61,286 @@ ErrorMessage(
...
-                  (findData.dwFileAttributes & FILE_ATTRIBUTE_SYSTEM) ? L'S' : L' ',
-                  (findData.dwFileAttributes & FILE_ATTRIBUTE_HIDDEN) ? L'H' : L' ',
-                  (findData.dwFileAttributes & FILE_ATTRIBUTE_READONLY) ? L'R' : L' ',
-                  szFullName);
-        bFound = TRUE;
-    }
-    while (FindNextFileW(hFind, &findData));
-    FindClose(hFind);
+    /* The full path buffer the function will act recursively */
+    // PWSTR FullPath;
```
learn-more, line 147: Debug leftover?

base/applications/cmdutils/attrib/attrib.c
```diff
@@ -77,217 +61,286 @@ ErrorMessage(
...
+                // increasing its size by _countof(EnumCtx->findData.cFileName) + 1
+                // to satisfy this copy, as well as the one made in the loop below.
+            }
+            // else
+            ConPrintf(StdErr, L"Directory level too deep: %s\n", EnumCtx->FullPathBuffer);
+            return FALSE;
+        }
 
-        hFind = FindFirstFileW(szFullName, &findData);
+        hFind = FindFirstFileW(/*EnumCtx->FullPathBuffer*/pPath, &EnumCtx->findData);
```
learn-more, line 203: I am not sure you should re-use the `findData` while another one is still in progress?

base/applications/cmdutils/attrib/attrib.c
```diff
@@ -77,217 +61,286 @@ ErrorMessage(
...
-            Error = GetLastError();
-            if ((Error != ERROR_DIRECTORY) && (Error != ERROR_SHARING_VIOLATION)
-                  && (Error != ERROR_FILE_NOT_FOUND))
+            DWORD Error = GetLastError();
+            if ((Error != ERROR_DIRECTORY) &&
+                (Error != ERROR_SHARING_VIOLATION) &&
+                (Error != ERROR_FILE_NOT_FOUND))
             {
-                ErrorMessage(Error, pszFile);
+                ErrorMessage(Error, EnumCtx->FullPathBuffer);
```
learn-more, line 211: `pPath` or `EnumCtx->FullPathBuffer`?

base/applications/cmdutils/attrib/attrib.c
```diff
@@ -77,217 +61,286 @@ ErrorMessage(
...
-    DWORD  dwAttribute;
-    DWORD  Error;
-
-    /* prepare full file name buffer */
-    wcscpy(szFullName, pszPath);
-    pszFileName = szFullName + wcslen(szFullName);
-
-    /* display all subdirectories */
-    if (bRecurse)
+    PWSTR findFileName = EnumCtx->findData.cFileName;
```
learn-more, line 165: I don't think this `findFileName` makes the code more readable

base/applications/cmdutils/attrib/attrib.c
HBelusca: Added comment in code.

base/applications/cmdutils/attrib/attrib.c
HBelusca: Added explanation in code.

binarymaster (approved):
Looks nice 🙂

base/applications/cmdutils/attrib/attrib.c
```diff
@@ -77,217 +59,260 @@ ErrorMessage(
...
-            if (!(findData.dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY))
-                continue;
-
-            if (!wcscmp(findData.cFileName, L".") ||
-                !wcscmp(findData.cFileName, L".."))
-            {
-                continue;
-            }
+    DWORD dwAttributes = pFindData->dwFileAttributes;
+
```
katahiromz, line 75: extra empty line!

katahiromz (approved):

Doug-Lyons (approved):
I suggest that you add a comment in the code near the new statement about Line 25 which has '//#define EXTENDED_ATTRIBUTES' that the language files will need to be improved to cover this new functionality when it is enabled.
LGTM. This has been a long time project of yours as I remember. Congratulations on its completion.

base/applications/cmdutils/attrib/attrib.c
HBelusca: ? You mean the line between dwAttributes and the UNREFERENCED_PARAMETER thing? well yes of course there is an empty line there, since UNREFERENCED_PARAMETER is not part of the list of local variables...
(dunno why there was even a need to explain this triviality)

PR #4217, opened 2021-12-29
[EXPLORER] Implement CDesktopThread::Destroy()

TAN-Gaming:
## Purpose

Implement CDesktopThread::Destroy()


base/shell/explorer/desktop.cpp
```diff
@@ -22,107 +22,160 @@
...
 
-    HRESULT Initialize(IN OUT ITrayWindow *pTray)
-    {
-        HANDLE hThread;
-        HANDLE Handles[2];
+    m_hThread = CreateThread(NULL, 0, s_DesktopThreadProc, (LPVOID)this, 0, NULL);
+
+    if (!m_hThread)
+    {   
+        return E_FAIL;
```
HBelusca, line 83: The function creates a handle, but then something else fails: it needs to do its own cleanup.
```suggestion
        CloseHandle(m_hEvent);
        m_hEvent = NULL;
        return E_FAIL;
```
Because otherwise, what if somebody who uses the CDesktopThread() does:
```
ret = someThread->Initialize(tray1);
if (ret != S_OK)
{
    ret = someThread->Initialize(tray2);
}
```
?
You would get leakage of the event handle.

base/shell/explorer/desktop.cpp
```diff
@@ -22,107 +22,160 @@
...
+    }
+
+    Handles[0] = m_hThread;
+    Handles[1] = m_hEvent;
 
-        m_Tray = pTray;
+    while (true)
```
HBelusca, line 89: usually some compilers warn that you're using a constant in the while test, but they don't say anything when using the `for(;;)` construct, so I would suggest you don't do unnecessary changes here and keep the original.

base/shell/explorer/desktop.cpp
```diff
@@ -22,107 +22,160 @@
...
+    {
+        DWORD WaitResult = MsgWaitForMultipleObjects(_countof(Handles), Handles, FALSE, INFINITE, QS_ALLEVENTS);
 
-        m_hEvent = CreateEventW(NULL, FALSE, FALSE, NULL);
-        if (!m_hEvent)
-            return E_FAIL;
+        if (WaitResult == WAIT_OBJECT_0 + _countof(Handles))
+        {
+            TrayProcessMessages(m_Tray);
+        }
```
HBelusca, line 97:
```suggestion
```

base/shell/explorer/desktop.cpp
```diff
@@ -22,107 +22,160 @@
...
+        return E_FAIL;
+    }
+
+    Handles[0] = m_hThread;
+    Handles[1] = m_hEvent;
 
-        m_Tray = pTray;
+    while (true)
+    {
+        DWORD WaitResult = MsgWaitForMultipleObjects(_countof(Handles), Handles, FALSE, INFINITE, QS_ALLEVENTS);
```
HBelusca, line 92:
```suggestion
```

base/shell/explorer/desktop.cpp
```diff
@@ -22,107 +22,160 @@
...
-            {
-                break;
-            }
+            TerminateThread(m_hThread, 0);
         }
+    }
+}
 
-        CloseHandle(hThread);
-        CloseHandle(m_hEvent);
```
HBelusca, line 106: You're sure you don't need to close the handles?

base/shell/explorer/desktop.cpp
```diff
@@ -22,107 +22,160 @@
...
+
+void CDesktopThread::Destroy()
+{
+    if (m_hThread)
+    {
+        DWORD dwExitCode;
 
-        for (;;)
+        GetExitCodeThread(m_hThread, &dwExitCode);
+
```
HBelusca, line 114:
```suggestion
```

base/shell/explorer/desktop.cpp
ThFabba: That's invalid usage for every `Initialize` function I've ever seen. It'd be easier to just assert `m_hEvent == nullptr` at the beginning of `Initialize` or something.

base/shell/explorer/desktop.cpp
TAN-Gaming: I moved them to the destructor.

TAN-Gaming:
Handles leak fixed.

base/shell/explorer/desktop.cpp
```diff
@@ -22,107 +22,184 @@
...
-        hThread = CreateThread(NULL, 0, s_DesktopThreadProc, (PVOID)this, 0, NULL);
-        if (!hThread)
+        if (WaitResult == WAIT_OBJECT_0 + _countof(Handles))
         {
-            CloseHandle(m_hEvent);
-            return E_FAIL;
+            TrayProcessMessages(m_Tray);
         }
+        
+        else if (WaitResult != WAIT_FAILED && WaitResult != WAIT_OBJECT_0)
```
ThFabba, line 111: I know you didn't change this, but maybe you understand: why is this loop not exiting in the `WAIT_OBJECT_0` case? If the thread encounters a failure, it seems to exit without setting the event, so it seems like we'd loop forever here in that case.

TAN-Gaming: I think I need to rewrite the class because an old one used ExitProcess() hack.

TAN-Gaming:
I think I need to rewrite the class because an old one used ExitProcess() hack.

base/shell/explorer/desktop.cpp
```diff
@@ -22,107 +22,184 @@
...
-            DWORD WaitResult = MsgWaitForMultipleObjects(_countof(Handles), Handles, FALSE, INFINITE, QS_ALLEVENTS);
-            if (WaitResult == WAIT_OBJECT_0 + _countof(Handles))
-            {
-                TrayProcessMessages(m_Tray);
-            }
-            else if (WaitResult != WAIT_FAILED && WaitResult != WAIT_OBJECT_0)
-            {
-                break;
-            }
+            TerminateThread(m_hThread, 0);
```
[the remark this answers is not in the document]
TAN-Gaming: Thanks for your suggestion.

ThFabba:
Looks pretty good to me! Does ROS provide a way to exit explorer, or how did you test this?

HBelusca:
> Does ROS provide a way to exit explorer

It does: open the shutdown dialog, maintain Ctrl, Alt, Shift pressed while clicking on the dialog's Cancel button.

TAN-Gaming:
Loop cases handled, unnecessary checks removed.

PR #4218, opened 2021-12-29
[COMCTL32] Invalidate LVIS_CUT selected items too

StartForKiller:
## Purpose

_Hidden folders are not drawn with transparent icon until the filebrowser window is resized_

JIRA issue: [CORE-16722](https://jira.reactos.org/browse/CORE-16722)

## Proposed changes

_Seems that window doesn't send a setfocus event until the icon list are displayed so no transparent icon is displayed because the cut property is not applied to non focused listview, so the change proposed is to redraw the LVIS_CUT istead of just LVIS_SELECTED when window is focused_

### DIFF
--- a/dll/win32/comctl32/listview.c
+++ b/dll/win32/comctl32/listview.c
@@ -2276,7 +2276,7 @@ static void LISTVIEW_InvalidateSelectedItems(const LISTVIEW_INFO *infoPtr)
     iterator_frameditems(&i, infoPtr, &infoPtr->rcList); 
     while(iterator_next(&i))
     {
-	if (LISTVIEW_GetItemState(infoPtr, i.nItem, LVIS_SELECTED))
+	if (LISTVIEW_GetItemState(infoPtr, i.nItem, LVIS_SELECTED | LVIS_CUT))
 	    LISTVIEW_InvalidateItem(infoPtr, i.nItem);
     }
     iterator_destroy(&i);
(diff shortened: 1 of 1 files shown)


HBelusca:
OK if it's not a bug intrinsic to our shell32/browseui.
I think this patch should also be sent to Wine.

StartForKillerMC:
> OK if it's not a bug intrinsic to our shell32/browseui. I think this patch should also be sent to Wine.

How can i send a patch to wine?

gonzoMD:
> > OK if it's not a bug intrinsic to our shell32/browseui. I think this patch should also be sent to Wine.
> 
> How can i send a patch to wine?

Check this page, it's a bit more complicated like with github: https://wiki.winehq.org/Submitting_Patches

StartForKillerMC:
Anyone can check if the explorer with our dll works the same on windows, if it's present on windows too? if not this could mean i should send it to wine

StartForKillerMC:
Ok, time to add the ifdef reactos, and then it should be ready for approval right?

JoachimHenze:
It is a pity that nobody seems to have proposed that 2 year-old one-liner-fix to Wine yet. @whindsaks @StartForKiller @BinaryMaster

dll/win32/comctl32/listview.c
```diff
@@ -2276,7 +2276,7 @@ static void LISTVIEW_InvalidateSelectedItems(const LISTVIEW_INFO *infoPtr)
     iterator_frameditems(&i, infoPtr, &infoPtr->rcList); 
     while(iterator_next(&i))
     {
-	if (LISTVIEW_GetItemState(infoPtr, i.nItem, LVIS_SELECTED))
+	if (LISTVIEW_GetItemState(infoPtr, i.nItem, LVIS_SELECTED | LVIS_CUT))
```
HBelusca, line 2279:
```suggestion
#ifdef __REACTOS__
	if (LISTVIEW_GetItemState(infoPtr, i.nItem, LVIS_SELECTED | LVIS_CUT))
#else
	if (LISTVIEW_GetItemState(infoPtr, i.nItem, LVIS_SELECTED))
#endif
```

PR #4219, opened 2021-12-29
[SETUPAPI] Fix SetupInstallServicesFromInfSectionExW returning TRUE even if the Section was not found

vicmarcal:
- A section has to be found in order to return TRUE.
- It  was returning true even if the section was not found.

## Purpose

This patch is a slim version  created by @ThFabba based on my patch in 
JIRA issue: [CORE-9398](https://jira.reactos.org/browse/CORE-9398)

## Proposed changes

SetupInstallServicesFromInfSectionExW is returning TRUE in current Master even if the section can't be found by SetupFindFirstLineW. Add a flag to control the section name has been found and check against it.

### DIFF
--- a/dll/win32/setupapi/install.c
+++ b/dll/win32/setupapi/install.c
@@ -2,7 +2,7 @@
  * Setupapi install routines
  *
  * Copyright 2002 Alexandre Julliard for CodeWeavers
- *           2005-2006 Hervé Poussineau (hpoussin@reactos.org)
+ *           2005-2006 Hervï¿½ Poussineau (hpoussin@reactos.org)
  *
  * This library is free software; you can redistribute it and/or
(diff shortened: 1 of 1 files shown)


dll/win32/setupapi/install.c
```diff
@@ -2,7 +2,7 @@
  * Setupapi install routines
  *
  * Copyright 2002 Alexandre Julliard for CodeWeavers
- *           2005-2006 Herv� Poussineau (hpoussin@reactos.org)
+ *           2005-2006 Herv� Poussineau (hpoussin@reactos.org)
```
learn-more, line 5: oops?

vicmarcal: Thanks Fixed!

PR #4223, opened 2021-12-30
[PSDK][SHELL32][SHELLEXT] Add SHCreateFileExtractIconW() to the PSDK headers, and SAL-annotate it.

HBelusca:
Note that even if the MS PSDK and MSDN documents an hypothetical
ANSI version SHCreateFileExtractIconA(), this one never existed
exported in any Windows version!

### DIFF
--- a/sdk/include/psdk/shlobj.h
+++ b/sdk/include/psdk/shlobj.h
@@ -2092,6 +2092,28 @@ SHBindToParent(
   _Outptr_opt_ PCUITEMID_CHILD *ppidlLast);
 
 /****************************************************************************
+ * SHCreateFileExtractIcon API
+ */
+#if (NTDDI_VERSION >= NTDDI_WINXP)
+
+// NOTE: Even if documented on MSDN, the SHCreateFileExtractIconA()
+// ANSI function never existed on Windows!
+
+HRESULT
+WINAPI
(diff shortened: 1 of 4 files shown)


HBelusca:
Cc @StartForKillerMC too.

sdk/include/psdk/shlobj.h
```diff
@@ -2091,6 +2091,36 @@ SHBindToParent(
   _Outptr_ LPVOID *ppv,
   _Outptr_opt_ PCUITEMID_CHILD *ppidlLast);
 
+/****************************************************************************
+ * SHCreateFileExtractIcon API
+ */
+#if (NTDDI_VERSION >= NTDDI_WINXP)
+
+#if 0 // NOTE: Even if documented on MSDN, the ANSI function never existed on Windows!
```
learn-more, line 2099: The just don't include it?

HBelusca: OK I will just leave the comment (as documentation so that if someone compares with MSDN or PSDK and sees the ansi version, (s)he shouldn't be surprised it doesn't exist in our header).

HBelusca: Fixed.

learn-more (approved):

PR #4229, opened 2021-12-31
Refactor Cloud Workspace support to support Codespaces and VS Remote for Containers

sr229:
## Purpose

Currently, ROS only supports Gitpod workspaces. This brings the support level of what we have from Gitpod to GitHub Codespaces and VS Remote for Containers as well as making the configuration platform-agnostic, giving contributors the choice between Codespaces and Gitpod.

## Proposed changes

- Add workspace extension recommendations for local, GH Codespaces, and Gitpod workspaces.
- Launch, settings, and Tasks has been redone to be more vendor-neutral, as both Gitpod and Codespaces reads from `.vscode` nowadays.
- Introduced a Codespace-specific Dockerfile that mirrors the Gitpod version but remains functionally similar.


learn-more (changes requested):
By adding .vscode files here, and removing it from the .gitignore list, you are forcing everyone to have the same .vscode configuration,
and you are breaking existing configurations from everyone.

sr229:
> By adding .vscode files here, and removing it from the .gitignore list, you are forcing everyone to have the same .vscode configuration, and you are breaking existing configurations from everyone.

The compromise was necessary however, since Gitpod nowadays uses VSCode as the default experience, so there's no point in using `.theia`, when even Theia itself reads from `.vscode` too. If you have your own configuration, I think it's sufficient to say that you can always unstage the change yourself, so you don't commit it.

.vscode/settings.json
```diff
@@ -0,0 +1,10 @@
+{
```
wjk, line 1: Perhaps rename this file to `settings.json.example`?

HBelusca:
This PR may need to be revisited (as of March 2026), because: gitpod became Ona; the Unix RosBE being used is outdated compared to what is currently required to build ReactOS, etc.
Cc @learn-more @zefklop

PR #4239, opened 2022-01-03
[WINDOWSCODECS] Add converter for WICPixelFormat32bppRGBA

thc13:
## Purpose

This fixes CORE-15708 "PdfSam 3.3.5 setup can not decode bmp"
https://jira.reactos.org/browse/CORE-15708

Wine does not have this issue, but it did not have it back when last sync to
WineStaging-4.18 was done. This commit is as near as possible to actual
wine-7.0-rc3 version. This wine code uses reverse_bgr8 instead of own
convert_rgba_to_bgra, but it leads to wrong colors.

JIRA issue: [CORE-15708](https://jira.reactos.org/browse/CORE-15708)

## Proposed changes

- Add converter for WICPixelFormat32bppRGBA

### DIFF
--- a/dll/win32/windowscodecs/converter.c
+++ b/dll/win32/windowscodecs/converter.c
@@ -722,6 +722,15 @@ static HRESULT copypixels_to_32bppBGRA(struct FormatConverter *This, const WICRe
                     pbBuffer[cbStride*y+4*x+3] = 0xff;
         }
         return S_OK;
+    case format_32bppRGBA:
+        if (prc)
+        {
+            HRESULT res;
+            res = IWICBitmapSource_CopyPixels(This->source, prc, cbStride, cbBufferSize, pbBuffer);
+            if (FAILED(res)) return res;
+            convert_rgba_to_bgra(4, pbBuffer, prc->Width, prc->Height, cbStride);
+        }
+        return S_OK;
     case format_32bppBGRA:
         if (prc)
             return IWICBitmapSource_CopyPixels(This->source, prc, cbStride, cbBufferSize, pbBuffer);
@@ -859,6 +868,7 @@ static HRESULT copypixels_to_32bppBGRA(struct FormatConverter *This, const WICRe
         }
         return S_OK;
     default:
+        FIXME("Unimplemented conversion path!\n");
         return WINCODEC_ERR_UNSUPPORTEDOPERATION;
     }
 }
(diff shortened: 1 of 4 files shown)


dll/win32/windowscodecs/main.c
```diff
@@ -228,6 +228,28 @@ void reverse_bgr8(UINT bytesperpixel, LPBYTE bits, UINT width, UINT height, INT
...
+    {
+        pixel = bits + stride * y;
+
+        for (x=0; x<width; x++)
+        {
+            temp = pixel[3];
+            pixel[3] = pixel[0];
+            pixel[0] = pixel[1];
+            pixel[1] = pixel[2];
+//            pixel[2] = pixel[3];
```
SergeGautherie, line 246: Remove this comment line.

dll/win32/windowscodecs/converter.c
```diff
@@ -859,6 +868,7 @@ static HRESULT copypixels_to_32bppBGRA(struct FormatConverter *This, const WICRe
         }
         return S_OK;
     default:
+        FIXME("Unimplemented conversion path!\n");
```
SergeGautherie, line 871: Add `source_format`.

thc13: This is a take over from wine. So i would prefer to keep it.

dll/win32/windowscodecs/main.c
thc13: I can if need, no big deal, but it is a better reminder then writing a ton of words to see what was done and why temp had to be used. I screwed up the conversion, because at first attempt i did not take into account that i need temp and order is also important. But if others recommend it too, i will do a new commit.

dll/win32/windowscodecs/converter.c
SergeGautherie: I understand.
NiceToHave: Improve Wine first and/or use `#ifndef __REACTOS__`?

dll/win32/windowscodecs/main.c
learn-more: The code is fairly obvious in this regard, the commented line seems more like a testing leftover now.

thc13: Done

dll/win32/windowscodecs/converter.c
thc13: Keeping it is better i think. Similar problems like this will be visible in log files in future, not only while testing apps. This should pinpoint us to missing feature.

SergeGautherie: > pinpoint

Isn't that exactly why adding `source_format` value would be useful?

learn-more (approved):

PR #4241, opened 2022-01-03
[XDK] MmBadPointer: Exists on NT6+ only

SergeGautherie:
Add an `#ifdef`, plus a comment.

JIRA issue: [CORE-15443](https://jira.reactos.org/browse/CORE-15443)


ThFabba:
Negative, this contradicts what's in Microsoft's headers.

PR #4244, opened 2022-01-04
[CMAKE] Only warn about base addresses for release builds

ThFabba:
## Purpose

Let's not bother people with pointless warnings -- let's regenerate base addresses right before each release branch cut, then the warning actually has meaning.

JIRA issue: [CORE-17968](https://jira.reactos.org/browse/CORE-17968)


JoachimHenze:
No, please do not mute!
I am fine with your proposal of recreating each time before the branching though.
**But that does not require us to mute the warning!**
The question to clarify is who will be the responsible person to regen each time. (I cannot as I don't have any python.)
But I could request/remember each time, if someone who is willing to take that responsibility.

ThFabba:
The reason to mute the warning is to stop wasting developer time due to people complaining about the warnings. There is no reason to be bothered by these warnings, it's a waste of time to regenerate base addresses all the time.

HBelusca:
LGTM for me, but OK 😂

JoachimHenze:
Well we should have all those addresses for each release we aim to, and it should be created before the branching.
The real question is: Who will do it?

PR #4245, opened 2022-01-04
[GDI32] Don't make pointless copies of bitmap buffers.

ThFabba:


win32ss/gdi/gdi32/objects/bitmap.c
```diff
@@ -767,14 +742,12 @@ SetDIBitsToDevice(
      pConvertedInfo->bmiHeader.biCompression  == BI_PNG )) )*/
     {
         LinesCopied = NtGdiSetDIBitsToDeviceInternal(hdc, XDest, YDest, Width, Height, XSrc, YSrc,
-            StartScan, ScanLines, (LPBYTE) pvSafeBits, (LPBITMAPINFO) pConvertedInfo, ColorUse,
+            StartScan, ScanLines, (LPBYTE) Bits, (LPBITMAPINFO) pConvertedInfo, ColorUse,
```
learn-more, line 745: `NtGdiSetDIBitsToDeviceInternal->GreCreateBitmapEx->DecompressBitmap` does access this under SEH,
but it never propagates an error so it will continue on uninitialized data afterwards, is that expected behavior?

ThFabba: Uh, yeah that whole code path seems off. If the bitmap is not compressed, the UM pointer is directly saved to the surface object, it seems; if they're compressed then we go through this code path, but I think we go through this one first:  https://github.com/reactos/reactos/blob/master/win32ss/gdi/eng/surface.c#L265-L275 -- which means the destination buffer is zero-initialized. So at least that `DecompressBitmap` aspect should be safe (no idea whether it's correct, though).

HBelusca, line 745: Note: the cast in front of `pConvertedInfo` shouldn't be necessary, as it is already of the correct type.

win32ss/gdi/gdi32/objects/bitmap.c
```diff
@@ -731,29 +729,6 @@ SetDIBitsToDevice(
 
     cjBmpScanSize = DIB_BitmapMaxBitsSize((LPBITMAPINFO) lpbmi, ScanLines);
 
-    pvSafeBits = RtlAllocateHeap(GetProcessHeap(), 0, cjBmpScanSize);
```
HBelusca, line 734: This sort of reallocation + copy might be needed only if the Bits pointer is not DWORD-aligned....

win32ss/gdi/gdi32/objects/bitmap.c
```diff
@@ -731,29 +729,6 @@ SetDIBitsToDevice(
...
-            // space.....
-            DPRINT1("SetDIBitsToDevice fail to read BitMapInfo: %p or Bits: %p & Size: %u\n",
-                pConvertedInfo, Bits, cjBmpScanSize);
-        }
-        DPRINT("SetDIBitsToDevice Allocate Bits %u!!!\n", cjBmpScanSize);
-    }
-
     if (!GdiGetHandleUserData(hdc, GDI_OBJECT_TYPE_DC, (PVOID) & pDc_Attr))
     {
         SetLastError(ERROR_INVALID_PARAMETER);
```
HBelusca, line 734: This failure code path is leaking the allocated `pConvertedInfo` (and was leaking previously the `pvSafeBits` as well).
Other code above did, instead of the `return 0;` the following:
```c
    LinesCopied = 0;
    goto Exit;
```

win32ss/gdi/gdi32/objects/bitmap.c
```diff
@@ -839,32 +810,6 @@ StretchDIBits(
...
-                DPRINT1("StretchDIBits fail to read BitMapInfo: %p or Bits: %p & Size: %u\n",
-                    pConvertedInfo, lpBits, cjBmpScanSize);
-            }
-            DPRINT("StretchDIBits Allocate Bits %u!!!\n", cjBmpScanSize);
-        }
-    }
-
     if (!GdiGetHandleUserData(hdc, GDI_OBJECT_TYPE_DC, (PVOID) & pDc_Attr))
     {
         SetLastError(ERROR_INVALID_PARAMETER);
```
HBelusca, line 815: Same comment as above regarding leakage of allocated stuff.

win32ss/gdi/gdi32/objects/bitmap.c
```diff
@@ -839,32 +810,6 @@ StretchDIBits(
 
     cjBmpScanSize = GdiGetBitmapBitsSize((BITMAPINFO *) pConvertedInfo);
 
-    if (lpBits)
```
HBelusca, line 842: Similar comment to above: This might still need the reallocation if `lpBits` is not DWORD-aligned.

win32ss/gdi/gdi32/objects/bitmap.c
ThFabba: Yep, except our NtGdiSetDIBits doesn't require alignment, either.

HBelusca: *our*, perhaps... but then what would happen if you put our gdi32.dll with MS' win32k.sys? :trollface:

PR #4248, opened 2022-01-04
[NTUSER] Plan B: UserDereferenceObject in ExitThreadCallback

katahiromz:
## Purpose

## Proposed changes

- Call `UserDereferenceObject` function in `ExitThreadCallback` function to release the default IMC of the thread.

## TODO

- [ ] Do tests.

### DIFF
--- a/win32ss/user/ntuser/main.c
+++ b/win32ss/user/ntuser/main.c
@@ -811,6 +811,12 @@ ExitThreadCallback(PETHREAD Thread)
     if (ptiCurrent->KeyboardLayout)
         UserDereferenceObject(ptiCurrent->KeyboardLayout);
 
+    if (ptiCurrent->spDefaultImc)
+    {
+        UserDereferenceObject(ptiCurrent->spDefaultImc);
+        ptiCurrent->spDefaultImc = NULL;
+    }
+
     if (gptiForeground == ptiCurrent)
     {
 //       IntNotifyWinEvent(EVENT_OBJECT_FOCUS, NULL, OBJID_CLIENT, CHILDID_SELF, 0);


win32ss/user/ntuser/main.c
```diff
@@ -815,6 +815,12 @@ ExitThreadCallback(PETHREAD Thread)
     if (ptiCurrent->KeyboardLayout)
         UserDereferenceObject(ptiCurrent->KeyboardLayout);
 
+    if (ptiCurrent->spDefaultImc)
+    {
+        UserDereferenceObject(ptiCurrent->spDefaultImc);
+        ptiCurrent->spDefaultImc = NULL;
+    }
```
binarymaster, line 822: I think this way it will be dereferenced only a single time, am I right? :thinking:

katahiromz:
1. Use Plan A and Don't use Plan B.
2. Don't use Plan A but use Plan B.
3. Use both Plan A and B.
4. Something is wrong in Plan B.

binarymaster:
Plan A seems to be enough, at least for me.

katahiromz:
We won't release the default IMC at thread exit?

binarymaster:
Hm actually I'm not sure, maybe you're right and we need both A and B.

I don't have much knowledge in this area.

ThFabba:
As-is, the default IMC will be released on process exit when the handle table is cleared out. So releasing it on thread exit (which is earlier) makes sense.
I _think_ this UserDereferenceObject call will remove the handle table entry, and therefore prevent a double free on process exit... right?

win32ss/user/ntuser/main.c
```diff
@@ -815,6 +815,12 @@ ExitThreadCallback(PETHREAD Thread)
     if (ptiCurrent->KeyboardLayout)
         UserDereferenceObject(ptiCurrent->KeyboardLayout);
 
+    if (ptiCurrent->spDefaultImc)
+    {
+        UserDereferenceObject(ptiCurrent->spDefaultImc);
```
SergeGautherie, line 820: As object deref already happened, would it be possible/more_explicit/cleaner to close the (internal) handle instead?

SergeGautherie (commented):
I assume 'Use both Plan A and B' is wanted.

win32ss/user/ntuser/main.c
HBelusca: As can be seen from the (currently failing) tests, doing that extra dereference is unnecessary since the object is actually already destroyed.
```
*** Assertion failed: ObjHead->cLockObj < 0x10000
***   Source File: /win32ss/user/ntuser/object.c, line 615
```

katahiromz:
```txt
*** Assertion failed: ObjHead->cLockObj < 0x10000
***   Source File: /win32ss/user/ntuser/object.c, line 615
Break repeatedly, break Once, Ignore, terminate Process or terminate Thread (boipt)? 
kdb:>
```
Something is wrong. Closed.

PR #4249, opened 2022-01-04
[NTOS:KE/x64] Detect CPU vendor properly and store value in PRCB

binarymaster:
Also generate processor identifier properly based on this value
on the Configuration Manager machine-dependent initialization.

Update processor driver INF file accordingly.

JIRA issue: [CORE-17970](https://jira.reactos.org/browse/CORE-17970)

### DIFF
--- a/ntoskrnl/config/i386/cmhardwr.c
+++ b/ntoskrnl/config/i386/cmhardwr.c
@@ -14,14 +14,7 @@
 
 /* GLOBALS *******************************************************************/
 
-#ifdef _M_IX86
-PCHAR CmpID1 = "80%u86-%c%x";
-PCHAR CmpID2 = "x86 Family %u Model %u Stepping %u";
-#else
-PCHAR CmpID1 = "EM64T Family %u Model %u Stepping %u";
-PCHAR CmpID2 = "AMD64 Family %u Model %u Stepping %u";
-PCHAR CmpID3 = "VIA64 Family %u Model %u Stepping %u";
-#endif
+PCHAR CmpFullCpuID = "%s Family %u Model %u Stepping %u";
 PCHAR CmpBiosStrings[] =
 {
     "Ver",
@@ -353,7 +346,7 @@ CmpInitializeMachineDependentConfiguration(IN PLOADER_PARAMETER_BLOCK LoaderBloc
         for (i = 0; i < KeNumberProcessors; i++)
         {
 #ifdef _M_AMD64
-            PCHAR CmpID;
+            PCHAR FamilyId;
 #endif
             /* Get the PRCB */
             Prcb = KiProcessorBlock[i];
@@ -370,18 +363,19 @@ CmpInitializeMachineDependentConfiguration(IN PLOADER_PARAMETER_BLOCK LoaderBloc
             /* Check if the CPU doesn't support CPUID */
             if (!Prcb->CpuID)
             {
-                /* Build ID1-style string for older CPUs */
+                /* Build 80x86-style string for older CPUs */
                 sprintf(Buffer,
-                        CmpID1,
+                        "80%u86-%c%x",
                         Prcb->CpuType,
                         (Prcb->CpuStep >> 8) + 'A',
                         Prcb->CpuStep & 0xff);
             }
             else
             {
-                /* Build ID2-style string for newer CPUs */
+                /* Build full ID string for newer CPUs */
                 sprintf(Buffer,
-                        CmpID2,
+                        CmpFullCpuID,
+                        "x86",
                         Prcb->CpuType,
                         (Prcb->CpuStep >> 8),
                         Prcb->CpuStep & 0xff);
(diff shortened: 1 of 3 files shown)


ntoskrnl/config/i386/cmhardwr.c
```diff
@@ -376,6 +386,32 @@ CmpInitializeMachineDependentConfiguration(IN PLOADER_PARAMETER_BLOCK LoaderBloc
                         (Prcb->CpuStep >> 8),
                         Prcb->CpuStep & 0xff);
             }
+#elif defined(_M_AMD64)
```
HBelusca, line 389: I was looking at Geoff Chappell's website to see the differences between these `CpuID`, `CpuVendor` etc. fields.
Turns out that `CpuID` field in x64 should always be TRUE: I'm wondering whether we could add somewhere some
`ASSERT(Prcb->CpuID == TRUE);` 🤔

ntoskrnl/config/i386/cmhardwr.c
```diff
@@ -376,6 +386,32 @@ CmpInitializeMachineDependentConfiguration(IN PLOADER_PARAMETER_BLOCK LoaderBloc
...
+            }
+
+            /* Build ID2-style string for all 64-bit CPUs */
+            sprintf(Buffer,
+                    CmpID,
+                    Prcb->CpuType,
+                    (Prcb->CpuStep >> 8),
+                    Prcb->CpuStep & 0xff);
+#else
+            #error Unknown architecture
```
HBelusca, line 413:
```suggestion
#error Unknown architecture
```

ntoskrnl/ke/amd64/cpu.c
```diff
@@ -89,25 +86,25 @@ KiGetCpuVendor(VOID)
     /* Now check the CPU Type */
     if (!strcmp((PCHAR)Prcb->VendorString, CmpIntelID))
     {
-        return CPU_INTEL;
+        Prcb->CpuVendor = CPU_INTEL;
```
HBelusca, line 89: Interesting: the `CpuVendor` field already existed for x64 in NT 5.2, but it had to wait until NT 6.0 to appear in the x86 KPRCB: this should explain why our x86 `KiGetCpuVendor()` doesn't assign anything to this field.

HBelusca (approved):

ntoskrnl/ke/amd64/cpu.c
binarymaster: Yes, I noticed that too.

ntoskrnl/config/i386/cmhardwr.c
binarymaster: Don't have a checked build x64 kernel at hand, so don't know if assertion like that exists. Also in [this function](https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/ke/amd64/cpu.c;hb=1f6cd380509e7eab8cb653a748ff947fbf8d6974#l43) the code permanently sets it to TRUE.

HBelusca: ah ok then no need for any assertion.

ntoskrnl/config/i386/cmhardwr.c
```diff
@@ -357,6 +366,7 @@ CmpInitializeMachineDependentConfiguration(IN PLOADER_PARAMETER_BLOCK LoaderBloc
             ConfigData.ComponentEntry.AffinityMask = AFFINITY_MASK(i);
             ConfigData.ComponentEntry.Identifier = Buffer;
 
+#if defined(_M_IX86)
             /* Check if the CPU doesn't support CPUID */
             if (!Prcb->CpuID)
```
Extravert-ir, line 371: We don't support CPUs without CPUID, so this should be removed completely

Extravert-ir:
I don't see the point - why complicate the code and detect CPU vendor for that string? Where is it going to be used?
They share the same instruction set

ntoskrnl/config/i386/cmhardwr.c
```diff
@@ -241,6 +247,9 @@ CmpInitializeMachineDependentConfiguration(IN PLOADER_PARAMETER_BLOCK LoaderBloc
     CPU_INFO CpuInfo;
     ULONG ExtendedId;
     PKPRCB Prcb;
+#ifdef _M_AMD64
+    PCHAR CmpID;
```
Extravert-ir, line 251: I don't see the point - why complicate the code and detect CPU vendor for that string? Where is it going to be used?
They share the same instruction set

Extravert-ir (changes requested):

ntoskrnl/config/i386/cmhardwr.c
HBelusca: what string?

ntoskrnl/config/i386/cmhardwr.c
binarymaster: I didn't touched that code anyway :wink:

ntoskrnl/config/i386/cmhardwr.c
binarymaster: This string affects ACPI identifiers, and they affect what CPU driver will be assigned to the device. We don't have specialized CPU drivers though, but this is to match Windows behavior.

Extravert-ir: So we are going to have one driver for all CPUs on 32bit and 3 drivers on 64bit? Where is the logic here?

binarymaster: We will have one driver for both cases, until that changes for some reason. Note that Windows also have different CPU drivers on 32-bit.

HBelusca (approved):

PR #4255, opened 2022-01-07
[REACTOS] AppHelp: Replace INVALID_HANDLE_VALUE by NULL

SergeGautherie:
`NULL` is already used at some places.

JIRA issue: [CORE-10187](https://jira.reactos.org/browse/CORE-10187)


learn-more:
Did you verify that this matches what windows is doing?

SergeGautherie:
> Did you verify that this matches what windows is doing?

Yes.

learn-more:
> > Did you verify that this matches what windows is doing?
> 
> Yes, test succeeds/skips as intended from XP to 8.1.

I see skips where it would run tests successfully without your changes.

PR #4257, opened 2022-01-08
[MSPAINT] Prepare for debugging

katahiromz:
## Purpose
This PR helps the developers debug our Paint software.
JIRA issue: [CORE-17969](https://jira.reactos.org/browse/CORE-17969)

## Proposed changes

- Add `#include <debug.h>` and add link to `ntdll.dll` for debugging purpose.
- Use `DPRINT` for debugging.
- Add `CopyDIBImage` helper function for simplification.
- Refactor some codes.

## Screenshots
image
LGTM.


base/applications/mspaint/dib.cpp
```diff
@@ -208,3 +208,8 @@ HBITMAP DoLoadImageFile(HWND hwnd, LPCTSTR name, BOOL fIsMainFile)
 
     return hBitmap;
 }
+
+HBITMAP CopyDIBImage(HBITMAP hbm, INT cx, INT cy)
```
HBelusca, line 212: Could be made as an inlined function.

base/applications/mspaint/precomp.h
```diff
@@ -21,6 +21,17 @@
...
+#define NDEBUG
+#include <reactos/debug.h>
+
+#if DBG
+    #define __LOCATION__  LPCSTR file, INT line
+    #define __location__  __FILE__, __LINE__
+#else
+    #define __LOCATION__  /*empty*/
+    #define __location__  /*empty*/
+#endif
```
HBelusca, line 33: This is quite an invasive debugging way! I would suggest to keep this PR not as something that goes into master, but as a "tool" to apply only when local debugging of Paint is needed, and that needs to be removed afterwards once debugging has reached its goals 🤔

base/applications/mspaint/dib.cpp
katahiromz: OK.

base/applications/mspaint/precomp.h
katahiromz: `__LOCATION__` and `__location__` removed.

base/applications/mspaint/precomp.h
```diff
@@ -21,6 +21,9 @@
 #include <shellapi.h>
 #include <htmlhelp.h>
 
+#define NDEBUG
+#include <reactos/debug.h>
```
HBelusca, line 25: `#include <debug.h>` should be sufficient? or would this break compilation? (if so, then just keep using `reactos/debug.h`)

base/applications/mspaint/history.cpp
```diff
@@ -183,9 +186,9 @@ void ImageModel::StretchSkew(int nStretchPercentX, int nStretchPercentY, int nSk
 {
     int oldWidth = GetWidth();
     int oldHeight = GetHeight();
-    Insert((HBITMAP) CopyImage(hBms[currInd], IMAGE_BITMAP,
-           GetWidth() * nStretchPercentX / 100,
-           GetHeight() * nStretchPercentY / 100, 0));
+    INT newWidth = GetWidth() * nStretchPercentX / 100;
+    INT newHeight = GetHeight() * nStretchPercentY / 100;
```
HBelusca, line 190:
```suggestion
    INT newWidth = oldWidth * nStretchPercentX / 100;
    INT newHeight = oldHeight * nStretchPercentY / 100;
```

base/applications/mspaint/history.cpp
```diff
@@ -44,8 +44,9 @@ ImageModel::ImageModel()
 
 void ImageModel::CopyPrevious()
 {
+    DPRINT("CopyPrevious: %d\n", currInd);
```
HBelusca, line 47: That's ok; also, you could use `__FUNCTION__` if you don't want to hardcode the name of the function.

base/applications/mspaint/precomp.h
katahiromz: Used `<debug.h>`.

base/applications/mspaint/history.cpp
katahiromz: OK.

base/applications/mspaint/history.cpp
katahiromz: OK.

PR #4258, opened 2022-01-08
[SYSDM] Fix fancy edit Environment dialog box

arnavbhatt288:
## Purpose

Fix the issue of the changes of the environment values not being set sometimes through fancy dialog box.

## Proposed changes

- Make sure the environment value buffer is always reallocated before taking the values.

### DIFF
--- a/dll/cpl/sysdm/environment.c
+++ b/dll/cpl/sysdm/environment.c
@@ -155,7 +155,9 @@ GatherDataFromListView(HWND hwndListView,
             return 0;
     }
 
-    /* Copy the variable values while seperating them with a semi-colon except for the last value */
+    /* First reinitialize the value buffer then copy the variable values while
+       seperating them with a semi-colon except for the last value */
+    VarData->lpRawValue[0] = _T('\0');
     for (i = 0; i < NumberOfItems; i++)
     {
         if (i > 0)
--- a/dll/cpl/sysdm/environment.c
+++ b/dll/cpl/sysdm/environment.c
@@ -155,8 +155,8 @@ GatherDataFromListView(HWND hwndListView,
             return 0;
     }
 
-    /* First reinitialize the value buffer then copy the variable values while
-       seperating them with a semi-colon except for the last value */
+    /* First reinitialize the value buffer, then copy the variable values while
+     * separating them with a semi-colon, except for the last value. */
     VarData->lpRawValue[0] = _T('\0');
     for (i = 0; i < NumberOfItems; i++)
     {


dll/cpl/sysdm/environment.c
```diff
@@ -144,16 +144,13 @@ GatherDataFromListView(HWND hwndListView,
...
+    if (VarData->lpRawValue)
+        GlobalFree(VarData->lpRawValue);
 
-        VarData->lpRawValue = GlobalAlloc(GPTR, (dwValueLength + 1) * sizeof(TCHAR));
-        if (!VarData->lpRawValue)
-            return 0;
-    }
+    VarData->lpRawValue = GlobalAlloc(GPTR, (dwValueLength + 1) * sizeof(TCHAR));
+    if (!VarData->lpRawValue)
+        return 0;
```
HBelusca, line 153: So, what was the fundamental reason why the previous code didn't work?

arnavbhatt288: @HBelusca The reason was that since, if the length of changed environment values are of same or smaller than previous one then this will not reallocate the buffer hence if any new value will be tried to get copied to it, it will outright fail.
This was especially a problem when you either deleted the value(s) or even just changed the order of it.

HBelusca: > this will not reallocate the buffer

Indeed that was the original intent.

> hence if any new value will be tried to get copied to it, it will outright fail.

That's the actual problem: why would it fail?

arnavbhatt288: > > this will not reallocate the buffer
> 
> Indeed that was the original intent.
> 
> > hence if any new value will be tried to get copied to it, it will outright fail.
> 
> That's the actual problem: why would it fail? I would expect that one would just copy the new data (of smaller size, in the case the buffer didn't need to be reallocated) by replacing whatever is already present in the existing buffer (since we don't care about it anyway: now with your PR you just directly free it and replace by a fresh one).

Since the way the values are getting copied is by taking the values from the listview control one to a temporary buffer and then copying it to the actual buffer one by one since it is not possible to directly take all the values from it.

JoachimHenze:
With "fancy dialog box" you mean the resizable one that we recently added during 0.4.15-dev'ing?

arnavbhatt288:
> With "fancy dialog box" you mean the resizable one that we recently added during 0.4.15-dev'ing?

@JoachimHenze Yes.

dll/cpl/sysdm/environment.c
```diff
@@ -144,16 +144,13 @@ GatherDataFromListView(HWND hwndListView,
...
```
HBelusca, line 154: Simpler fix than all these changes:
1. Keep original code,
2. Add the actual missing `VarData->lpRawValue[0] = _T('\0');` just after that block, and before the copy.

The actual problem was, that since you are using the concatenation functions below, the code was trying to concatenate onto the existing data.

dll/cpl/sysdm/environment.c
```diff
@@ -144,16 +144,13 @@ GatherDataFromListView(HWND hwndListView,
...
-    /* Reallocate the value buffer, regrowing it if necessary */
-    if (!VarData->lpRawValue || (_tcslen(VarData->lpRawValue) < dwValueLength))
-    {
-        if (VarData->lpRawValue)
-            GlobalFree(VarData->lpRawValue);
+    /* Reallocate the value buffer */
+    if (VarData->lpRawValue)
+        GlobalFree(VarData->lpRawValue);
 
-        VarData->lpRawValue = GlobalAlloc(GPTR, (dwValueLength + 1) * sizeof(TCHAR));
```
HBelusca, line 153: (Note that we could have used the `GlobalReAlloc` function there instead for the reallocation)

dll/cpl/sysdm/environment.c
```diff
@@ -144,16 +144,13 @@ GatherDataFromListView(HWND hwndListView,
...
-        if (!VarData->lpRawValue)
-            return 0;
-    }
+    VarData->lpRawValue = GlobalAlloc(GPTR, (dwValueLength + 1) * sizeof(TCHAR));
+    if (!VarData->lpRawValue)
+        return 0;
 
     /* Copy the variable values while seperating them with a semi-colon except for the last value */
     for (i = 0; i < NumberOfItems; i++)
```
HBelusca, line 156:
```suggestion
    VarData->lpRawValue[0] = _T('\0'); // Reinitialize the value buffer.
    for (i = 0; i < NumberOfItems; i++)
```

dll/cpl/sysdm/environment.c
arnavbhatt288: > Simpler fix than all these changes:
> 
> 1. Keep original code,
> 2. Add the actual missing `VarData->lpRawValue[0] = _T('\0');` just after that block, and before the copy.
> 
> The actual problem was, that since you are using the concatenation functions below, the code was trying to concatenate onto the existing data.

@HBelusca The reason why I didn't do this since I thought that since we aren't required the old values at all and going for the new values, it would be much simpler to just reallocating it again. (since even if it is bigger or smaller, it would get the desired size)
Nonetheless I am taking your suggestion.

dll/cpl/sysdm/environment.c
arnavbhatt288: Ahh this crossed my mind entirely!

dll/cpl/sysdm/environment.c
```diff
@@ -155,7 +155,9 @@ GatherDataFromListView(HWND hwndListView,
             return 0;
     }
 
-    /* Copy the variable values while seperating them with a semi-colon except for the last value */
+    /* First reinitialize the value buffer then copy the variable values while
+       seperating them with a semi-colon except for the last value */
```
HBelusca, line 159:
```suggestion
    /* First reinitialize the value buffer, then copy the variable values while
     * separating them with a semi-colon, except for the last value. */
```

PR #4259, opened 2022-01-08
[NTOS:KE] Fully implement FPU state Save/Restore mechanism

GeoB99:
On ReactOS we are merely doing a **FNSAVE** when saving the non-volatile data of floating point unit state, while on Windows it does a lot more than that.

FPU state saving is done thorough this rule -- check the NPX state of the current thread. If NPX is loaded then the mechanism grabs the NPX registers, otherwise the non-volatile data is obtained from NPX frame. Windows performs a `FNSAVE` or `FXSAVE` only if NPX state is loaded and that the FPU state context is being held in a pool allocation.

If the NPX interrupt level is no longer equal to the current interrupt level of the calling thread, it checks if the level in question is APC_LEVEL and if that's the case then from that moment memory pool is allocated to hold the FX save area, otherwise the said area is obtained from the current processor control region.

When restoring the FPU state back, we aren't checking if the calling thread is right the same as the one that performed a save FPU state context. We're only checking on an ASSERT if the NPX interrupt level is the same as the current one but on release builds that's a NOP, which is pretty useless. There's `INVALID_FLOATING_POINT_STATE` bugcheck for this kind of stuff. If it's deemed that the saved FPU state context is not valid, ReactOS will crash.

**JIRA Issue:** [CORE-12904](https://jira.reactos.org/browse/CORE-12904)

## TODO

- [x] ~~Investigate why `IRQL_NOT_LESS_OR_EQUAL` occurs when the KMTEST has finished~~

### DIFF
--- a/ntoskrnl/include/internal/i386/intrin_i.h
+++ b/ntoskrnl/include/internal/i386/intrin_i.h
@@ -45,7 +45,6 @@ Ke386FxSave(IN PFX_SAVE_AREA SaveArea)
     asm volatile ("fxsave (%0)" : : "r"(SaveArea));
 }
 
-
 FORCEINLINE
 VOID
 Ke386FnSave(IN PFLOATING_SAVE_AREA SaveArea)
@@ -66,6 +65,28 @@ Ke386SaveFpuState(IN PFX_SAVE_AREA SaveArea)
     {
         __asm__ __volatile__ ("fnsave %0\n wait\n" : : "m"(*SaveArea));
     }
+}
+
+FORCEINLINE
+VOID
+Ke386RestoreFpuState(_In_ PFX_SAVE_AREA SaveArea)
+{
+    extern ULONG KeI386FxsrPresent;
+    if (KeI386FxsrPresent)
+    {
+        __asm__ __volatile__ ("fxrstor %0\n" : : "m"(*SaveArea));
+    }
+    else
+    {
+        __asm__ __volatile__ ("frstor %0\n\t" : "=m" (*SaveArea));
+    }
+}
+
+FORCEINLINE
+VOID
+Ke386ClearFpExceptions(VOID)
+{
+    __asm__ __volatile__ ("fnclex");
 }
 
 FORCEINLINE
@@ -161,6 +182,14 @@ __fxrstor(IN PFX_SAVE_AREA SaveArea)
 {
     __asm mov eax, SaveArea
     __asm fxrstor [eax]
+}
+
+FORCEINLINE
+VOID
+__frstor(_In_ PFX_SAVE_AREA SaveArea)
+{
+    __asm mov eax, SaveArea
+    __asm frstor [eax]
 }
 
 FORCEINLINE
@@ -314,16 +343,35 @@ Ke386SaveFpuState(IN PVOID SaveArea)
     }
 }
 
+FORCEINLINE
+VOID
+Ke386RestoreFpuState(_In_ PVOID SaveArea)
+{
+    if (KeI386FxsrPresent)
+    {
+        __fxrstor((PFX_SAVE_AREA)SaveArea);
+    }
+    else
+    {
+        __frstor((PFX_SAVE_AREA)SaveArea);
+    }
+}
+
+FORCEINLINE
+VOID
+Ke386ClearFpExceptions(VOID)
+{
+    __asm fnclex;
+}
+
 #define Ke386FnSave __fnsave
 #define Ke386FxSave __fxsave
 // The name suggest, that the original author didn't understand what frstor means
 #define Ke386FxStore __fxrstor
 
-
 #else
 #error Unknown compiler for inline assembler
 #endif
-
 
 #define Ke386GetGlobalDescriptorTable __sgdt
 #define Ke386SetGlobalDescriptorTable __lgdt
@@ -332,6 +380,5 @@ Ke386SaveFpuState(IN PVOID SaveArea)
 #ifdef __cplusplus
 } // extern "C"
 #endif
-
 
 /* EOF */
--- a/ntoskrnl/include/internal/i386/ke.h
+++ b/ntoskrnl/include/internal/i386/ke.h
@@ -235,6 +235,17 @@ typedef struct _LARGE_IDENTITY_MAP
     PVOID PagesList[30];
 } LARGE_IDENTITY_MAP, *PLARGE_IDENTITY_MAP;
 
+//
+// Floating Point Internal Context Structure
+//
+typedef struct _FLOATING_SAVE_CONTEXT
+{
+    PKTHREAD CurrentThread;
+    KIRQL OldNpxIrql;
+    PFX_SAVE_AREA Buffer;
+    PFX_SAVE_AREA PfxSaveArea;
+} FLOATING_SAVE_CONTEXT, *PFLOATING_SAVE_CONTEXT;
+
 /* Diable interrupts and return whether they were enabled before */
 FORCEINLINE
 BOOLEAN
--- a/ntoskrnl/ke/i386/cpu.c
+++ b/ntoskrnl/ke/i386/cpu.c
@@ -12,6 +12,8 @@
 #define NDEBUG
 #include <debug.h>
 
+#include <xmmintrin.h>
+
 /* GLOBALS *******************************************************************/
 
 /* The TSS to use for Double Fault Traps (INT 0x9) */
@@ -75,6 +77,9 @@ typedef union _CPU_SIGNATURE
     ULONG AsULONG;
 } CPU_SIGNATURE;
 
+/* FX area alignment size */
+#define FXSAVE_ALIGN 15
+
 /* SUPPORT ROUTINES FOR MSVC COMPATIBILITY ***********************************/
 
 /* NSC/Cyrix CPU configuration register index */
@@ -96,7 +101,6 @@ setCx86(UCHAR reg, UCHAR data)
     WRITE_PORT_UCHAR((PUCHAR)(ULONG_PTR)0x22, reg);
     WRITE_PORT_UCHAR((PUCHAR)(ULONG_PTR)0x23, data);
 }
-
 
 /* FUNCTIONS *****************************************************************/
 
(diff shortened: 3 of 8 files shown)


ntoskrnl/ke/i386/cpu.c
```diff
@@ -1359,62 +1425,392 @@ KiCoprocessorError(VOID)
...
+    /* Set the current NPX IRQL and store the NPX state */
+    FsContext->CurrentThread->Header.NpxIrql = KeGetCurrentIrql();
+    FsContext->NpxState = FsContext->CurrentThread->NpxState;
+
+    /* Zero out the NPX CR0 state of the frame */
+    FxSaveAreaFrame->Cr0NpxState = 0;
+
+    /* Initialize the FPU and set the default FP control word */
+    Ke386FnInit();
+    Ke386SetCwState(0x27F);
```
HBelusca, line 1665: Naive question: are these values (here and the one below, and maybe elsewhere) flags? (If so it would be of interest to convert them into OR'ed flag values with human-readable names.)

GeoB99: Nope, these are just the "raw" (or whatever term it fits for this context of the post) values that are used also when initializing the context of a thread -- see e.g. -> https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/ke/i386/thrdini.c;hb=fdcefaa7cbd419aef8796b2f690c90d1cd968431#l133

HBelusca: quite interesting.... I wonder whether they have a special meaning (flags?) or it's really just, as you call them, "raw", like when the FPU sees this value (and not another one) he will do something.

ntoskrnl/ke/i386/cpu.c
```diff
@@ -1359,62 +1425,392 @@ KiCoprocessorError(VOID)
...
-    UNIMPLEMENTED_ONCE;
 
-    /* check if we are doing software emulation */
+    /* Check if we are doing software emulation */
     if (!KeI386NpxPresent) return STATUS_ILLEGAL_FLOAT_CONTEXT;
 
-    FpState = ExAllocatePool(NonPagedPool, sizeof (FNSAVE_FORMAT));
-    if (!FpState) return STATUS_INSUFFICIENT_RESOURCES;
+    /* Cache the FS context */
+    FsContext = *((PVOID *) Save);
```
HBelusca, line 1476: Weird cast. `Save` is a pointer to a struct with perfectly defined layout, that doesn't correspond to the layout of any of the members of `FLOATING_SAVE_CONTEXT`.

GeoB99: This is a trick used to initialize the floating point context from the parameter as a pointer that points to anything, as the members are opaque to the caller (one's not supposed to access them anyway). I don't think this would cause big issues down the road, but alternatively I can store the internal context by allocating a buffer for it with `ExAllocatePoolWithTag`, at the expense of taking up some bit of non paged memory though.

MVoloshin:
I'll retest ReactOS with AMD Radeon driver if once this is complete because I had the same issue as in CORE-12904

ntoskrnl/ke/i386/cpu.c
```diff
@@ -1359,62 +1438,397 @@ KiCoprocessorError(VOID)
...
+           FsContext->FxSaveArea->Cr0NpxState = FxSaveAreaFrame->Cr0NpxState;
+           FsContext->FxSaveArea->NpxSavedCpu = 0;
+        }
+        else
+        {
+            /* It's not, save the control word then */
+            Ke386SaveCwState(FsControlWord);
+
+            /* Get MXCsr if XMMI is supported */
+            if (KeI386XMMIPresent && KeI386FxsrPresent)
```
Extravert-ir, line 1615: Can't we just obtain that info from CurrentPrcb->FeatureBits? I don't get why a separate variable is needed.
In future, other extended state support will be added here

ntoskrnl/ke/i386/cpu.c
```diff
@@ -1359,62 +1438,397 @@ KiCoprocessorError(VOID)
...
+            Ke386RestoreFpuState(FsContext->FxSaveArea);
+        }
+        else
+        {
+            /* It's not, initialize the FPU and restore the control word */
+            Ke386FnInit();
+            Ke386SetCwState(FsContext->ControlWord);
+
+            /* Restore MXCsr if FX is supported */
+            if ((KeI386FxsrPresent) && (FsContext->MXCsr != 0))
```
HBelusca, line 1762:
```suggestion
            if (KeI386FxsrPresent && (FsContext->MXCsr != 0))
```

ntoskrnl/ke/i386/cpu.c
```diff
@@ -1359,62 +1438,397 @@ KiCoprocessorError(VOID)
...
+         */
+        if (FsContext->PoolAllocated)
+        {
+            /* It is, restore the saved state to frame */
+            RtlCopyMemory(FxSaveAreaFrame, FsContext->FxSaveArea, sizeof(FX_SAVE_AREA));
+        }
+        else
+        {
+            /* It's not, we've got to restore the state by hand */
+            if ((KeI386FxsrPresent) && (FsContext->MXCsr != 0))
```
HBelusca, line 1785:
```suggestion
            if (KeI386FxsrPresent && (FsContext->MXCsr != 0))
```

ntoskrnl/include/internal/i386/ke.h
```diff
@@ -235,6 +235,21 @@ typedef struct _LARGE_IDENTITY_MAP
...
+typedef struct _FLOATING_SAVE_CONTEXT
+{
+    PKTHREAD CurrentThread;
+    BOOLEAN PoolAllocated;
+    ULONG Cr0NpxState;
+    KIRQL OldNpxIrql;
+    UCHAR NpxState;
+    USHORT ControlWord;
+    ULONG MXCsr;
+    DECLSPEC_ALIGN(16) PFX_SAVE_AREA FxSaveArea;
```
Extravert-ir, line 250: Here you're aligning a pointer, not the area itself

ntoskrnl/ke/i386/cpu.c
HBelusca: Wouldn't `CurrentPrcb` be the per-processor state?

Using the `KeI386...Present` *global* variables ensures consistency of usage with respect to all these other Ke386Save* and similar functions, where the initial *Present -variables state is the one of the boot processor:
https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/ke/i386/kiinit.c;hb=75d33bf1481d328788e9898743421e907e7fc5fc#l469
and it is **expected** that all the other processors on an SMP system *have* the same characteristics as the boot processor (i.e. if the boot processor has FPU / XMM / whatever support, then **all** the other processors must have it).

ntoskrnl/include/internal/i386/intrin_i.h
```diff
@@ -195,6 +260,35 @@ __sldt(PVOID Descriptor)
...
 }
 
+#define Ke386FnSave __fnsave
+#define Ke386FxSave __fxsave
+// The name suggest, that the original author didn't understand what frstor means
+#define Ke386FxStore __fxrstor
+
+FORCEINLINE
+VOID
+Ke396XStore(_In_ PVOID Address, _In_ ULONGLONG Mask)
```
Extravert-ir, line 270: 396 haha

GeoB99: What a typo, fuck me. I will fix that.

ntoskrnl/ke/i386/cpu.c
```diff
@@ -1270,59 +1274,259 @@ KiCoprocessorError(VOID)
...
+
+    /* Give the saved NPX IRQL back to the NPX thread */
+    FsContext->CurrentThread->Header.NpxIrql = FsContext->OldNpxIrql;
+
+    /* Enable interrupts back */
+    _enable();
+
+    /* We're done, free the allocated area and context */
+    ExFreePoolWithTag(FsContext->Buffer, TAG_FLOATING_POINT_FX);
+    FsContext->Buffer = NULL;
```
HBelusca, line 1525: That NULL assign is not necessary since you're freeing `FsContext` just after.

ntoskrnl/ke/i386/cpu.c
```diff
@@ -1270,59 +1274,259 @@ KiCoprocessorError(VOID)
...
+
+    ExFreePoolWithTag(FsContext, TAG_FLOATING_POINT_CONTEXT);
+    FsContext = NULL;
```
HBelusca, line 1528: This NULL assign is not needed as well since it's not being used further in this function.

ntoskrnl/ke/i386/cpu.c
```diff
@@ -1270,59 +1274,259 @@ KiCoprocessorError(VOID)
...
+
+    /* Sanity checks */
+    ASSERT(Save);
     ASSERT(KeGetCurrentIrql() <= DISPATCH_LEVEL);
-    UNIMPLEMENTED_ONCE;
 
-    FpState = ExAllocatePool(NonPagedPool, sizeof (FNSAVE_FORMAT));
-    if (!FpState) return STATUS_INSUFFICIENT_RESOURCES;
+    /* Check if we are doing software emulation */
+    if (!KeI386NpxPresent) return STATUS_ILLEGAL_FLOAT_CONTEXT;
```
HBelusca, line 1322:
```suggestion
    if (!KeI386NpxPresent)
        return STATUS_ILLEGAL_FLOAT_CONTEXT;
```

HBelusca:
Testbots are OK: https://reactos.org/testman/compare.php?ids=82781,82783

ntoskrnl/ke/i386/cpu.c
```diff
@@ -1270,59 +1274,259 @@ KiCoprocessorError(VOID)
...
+    Ke386ClearFpExceptions();
+
+    /* Restore the state */
+    Ke386RestoreFpuState(FsContext->PfxSaveArea);
+
+    /* Restore the previous MXCsr if we saved it before */
+    if (KeI386XMMIPresent && KeI386FxsrPresent)
+    {
+        _mm_setcsr(FsContext->MXCsr);
+    }
```
Extravert-ir, line 1515: As far as I know, all this machinery can be replaced with `Ke386RestoreFpuState(FsContext->PfxSaveArea);`
Both MXCSR and x87 state are being overwritten by FNSAVE/FXSAVE instructions

Extravert-ir (changes requested):

ntoskrnl/ke/i386/cpu.c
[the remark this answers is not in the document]
HBelusca: an ASSERT, perhaps?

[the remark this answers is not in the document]
GeoB99: I can promote that check into an ASSERT, other than that I avoid any unrelated requests or whatever you're trying to suggest me to do. Thanks.

HBelusca (approved):

PR #4263, opened 2022-01-09
[3DTEXT] Rearrange settings dialog window to look nicer

blueghost-tbs:
## Purpose

Rearrange settings dialog window of the 3d text screensaver to look nicer.

JIRA issue: none

## Proposed changes

The settings dialog window for the 3d text screensaver is unnecessarily big. No code change, just a small visual fix. I also checked that all the elements are wide enough for all language strings.

Before:
before

After:
after


JoachimHenze:
I guess it was designed that big, because the MS one has a lot more options, that may (or may not) be intended by us to add one day:
image

blueghost-tbs:
> I guess it was designed that big, because the MS one has a lot more options, that may (or may not) be intended by us to add one day

That was my thought, too. But if a new setting comes along, one can easily resize the window anytime to fit in there. It just looks a bit silly now with all the empty space.

PR #4265, opened 2022-01-10
[CRT] _mbsstr should return NULL if the substring is longer than source string

julcar:
## Purpose

_mbsstr should return NULL when is requested to compare a substring whose length is greater than the source string, until now it tries to return a pointer and this leads to heap corruption

JIRA issue: [Core 16933](https://jira.reactos.org/browse/Core 16933)

Software affected: BabyWeb Server

### DIFF
--- a/sdk/lib/crt/mbstring/mbsstr.c
+++ b/sdk/lib/crt/mbstring/mbsstr.c
@@ -8,6 +8,9 @@ unsigned char *_mbsstr(const unsigned char *src1,const unsigned char *src2)
 {
   size_t len;
 
+  if (_mbslen(src2) > _mbslen(src1))
+    return NULL;
+
   if (src2 == NULL || *src2 == 0)
     return (unsigned char *)src1;
 
--- a/sdk/lib/crt/mbstring/mbsstr.c
+++ b/sdk/lib/crt/mbstring/mbsstr.c
@@ -8,16 +8,17 @@ unsigned char *_mbsstr(const unsigned char *src1,const unsigned char *src2)
 {
   size_t len;
 
-  if (src2 ==NULL || *src2 == 0)
+  if (src2 == NULL || *src2 == 0)
     return (unsigned char *)src1;
 
   len = _mbslen(src2);
 
   while(*src1)
-    {
-      if ((*src1 == *src2) && (_mbsncmp(src1,src2,len) == 0))
-	return (unsigned char *)src1;
-      src1 = (unsigned char *)_mbsinc(src1);
-    }
+  {
+    if ((*src1 == *src2) && (_mbsncmp(src1,src2,len) == 0))
+      return (unsigned char *)src1;
+
+    src1 = (unsigned char *)_mbsinc(src1);
+  }
   return NULL;
 }


sdk/lib/crt/mbstring/mbsstr.c
```diff
@@ -8,16 +8,20 @@ unsigned char *_mbsstr(const unsigned char *src1,const unsigned char *src2)
 {
   size_t len;
 
-  if (src2 ==NULL || *src2 == 0)
+  if (_mbslen(src2) > _mbslen(src1))
+    return NULL;
+
+  if (src2 == NULL || *src2 == 0)
```
SergeGautherie, line 14:
```suggestion
  if (src2 == NULL || *src2 == ANSI_NULL)
```

sdk/lib/crt/mbstring/mbsstr.c
```diff
@@ -8,16 +8,20 @@ unsigned char *_mbsstr(const unsigned char *src1,const unsigned char *src2)
...
   len = _mbslen(src2);
 
   while(*src1)
-    {
-      if ((*src1 == *src2) && (_mbsncmp(src1,src2,len) == 0))
-	return (unsigned char *)src1;
-      src1 = (unsigned char *)_mbsinc(src1);
-    }
+  {
+    if ((*src1 == *src2) && (_mbsncmp(src1,src2,len) == 0))
```
SergeGautherie, line 21:
```suggestion
    if (*src1 == *src2 && _mbsncmp(src1, src2, len) == 0)
```

sdk/lib/crt/mbstring/mbsstr.c
```diff
@@ -8,16 +8,20 @@ unsigned char *_mbsstr(const unsigned char *src1,const unsigned char *src2)
...
+      return (unsigned char *)src1;
+
+    src1 = (unsigned char *)_mbsinc(src1);
+  }
```
SergeGautherie, line 25:
```suggestion
  }

```

sdk/lib/crt/mbstring/mbsstr.c
```diff
@@ -8,16 +8,20 @@ unsigned char *_mbsstr(const unsigned char *src1,const unsigned char *src2)
 {
   size_t len;
 
-  if (src2 ==NULL || *src2 == 0)
+  if (_mbslen(src2) > _mbslen(src1))
```
SergeGautherie, line 11: `_mbslen()` does not seem to support `NULL`.
Current function does not seem to handle `NULL` `src1` either...

SergeGautherie (commented):
As well rewrite the whole file:
use 4-space indentation
rename the 2 parameters
...

JoachimHenze:
> As well rewrite the whole file: use 4-space indentation rename the 2 parameters ...

Part of CRT is 3rd party code (Wine/MinGW). We should not do massive white-space-changes in such parts therefore.
I am not sure whether mbsstr.c falls into that category.

sdk/lib/crt/mbstring/mbsstr.c
katahiromz, line 11:
```suggestion
  if (_mbslen(src1) < _mbslen(src2))
```

ThFabba:
I disagree with the premise of this. The function looks correct as it is.
`_mbsncmp(src1,src2,len)` should not return 0 if either of the strings is shorter than `len` -- it is not memcmp and should therefore respect terminating null characters. If that's not the case, then `_mbsncmp` is where the bug is.

PR #4256, opened 2022-01-08
[SDK] Add <debug/rappdbg.h> header

katahiromz:
## Purpose
Make debugging ReactOS applications easier.
JIRA issue: [CORE-17969](https://jira.reactos.org/browse/CORE-17969)

## Proposed changes

- Add header file `sdk/include/reactos/debug/rappdbg.h`.
- Add `ASSERT`, and `DPRINT` macros in `<debug/rappdbg.h>`.

## TODO

- [ ] Get consensus.

PR #4269, opened 2022-01-10
[SDK][APPLICATIONS][DRIVERS] Unlink ntdll.dll and use modified <debug.h>

katahiromz:
## Purpose
We decided that the excessive `ntdll.dll` dependency of the user-mode programs should be reduced if possible.

JIRA issue: [CORE-17969](https://jira.reactos.org/browse/CORE-17969)
Revival of PR #4256

## Proposed changes

- Modify `<debug.h>` to support user-mode debugging.
- Define `NTOS_MODE_USER` macro in the top of some modules.
- Unlink `ntdll.dll` of some modules.

## Screenshots
Using `gdb`:
after


base/applications/mspaint/precomp.h
```diff
@@ -21,8 +21,9 @@
 #include <shellapi.h>
 #include <htmlhelp.h>
 
+#define NO_STRSAFE
```
HBelusca, line 24: What's the problem with using stringsafe here?

sdk/include/reactos/debug/rappdbg.h
```diff
@@ -0,0 +1,71 @@
...
+static inline ULONG rappdbg_printfA(const char *file, int line, const char *fmt, ...)
+{
+    char buf[APPDBG_BUFSIZE];
+    va_list va;
+    const char *fname;
+    size_t cch;
+
+    va_start(va, fmt);
+
+    fname = strrchr(file, '\\');
```
HBelusca, line 44: Why doing these conversions?

HBelusca, line 44: Why not keeping the original file path as it comes out? (this is what the regular DPRINT etc. macros do)

base/applications/mspaint/precomp.h
katahiromz: The `<strsafe.h>` implementation can refuse incompatible function calls.

https://github.com/reactos/reactos/blob/47f6745bcdfb833b0311e052f04e6cd5634660eb/sdk/include/psdk/strsafe.h#L1921

So, some programs are incompatible to `<strsafe.h>`.

sdk/include/reactos/debug/rappdbg.h
katahiromz: Fixed.

sdk/include/reactos/debug.h
```diff
@@ -58,6 +60,44 @@ RtlAssert(
     _In_opt_z_ PCHAR Message
 );
 
+#else /* def NTOS_MODE_USER */
+
+typedef _Return_type_success_(return >= 0) LONG NTSTATUS;
```
HBelusca, line 65: For what code is the NTSTATUS needed?

sdk/include/reactos/debug.h
```diff
@@ -30,6 +30,8 @@
 extern "C" {
 #endif
 
+#ifndef NTOS_MODE_USER
```
HBelusca, line 33: Actually it would be nice to keep using the original DbgPrint and RtlAssert in case `NTOS_MODE_USER` is defined: this means the (user-mode) app will use ntdll functions too, and so it can call these direct ntdll-exported debug functions.

I would suggest something like this (adapted from a similar situation for the `TerminateCurrentProcess` macro at the end of this file):
```c
#if defined(_NTDDK_) || defined(NTOS_MODE_USER)
/* Kernel-Mode, or Native User-Mode */

// Here, the existing DbgPrint etc. definitions...

#elif defined(_WIN32)
/* Win32 User-Mode */

// Here, your new function and definitions...

#else
    #error "Unsupported target."
#endif
```

sdk/include/reactos/debug.h
```diff
@@ -58,6 +60,78 @@ RtlAssert(
...
+          "File '%s', Line %ld:\n\n%s\n\n%s",
+          (LPCSTR)FileName, LineNumber, (LPCSTR)FailedAssertion, Message);
+#endif
+
+    id = MessageBoxA(NULL, szBuff, "Assertion Failure", MB_ICONERROR | MB_ABORTRETRYIGNORE);
+
+    if (id == IDABORT)
+        ExitProcess((UINT)-1);
+
+    if (id == IDRETRY)
```
HBelusca, line 123:
```suggestion
    else if (id == IDRETRY)
```

sdk/include/reactos/debug.h
```diff
@@ -58,6 +60,78 @@ RtlAssert(
...
+__cdecl
+W32DbgPrint(_In_z_ _Printf_format_string_ PCSTR Format, ...)
+{
+    va_list va;
+    char szBuff[512];
+
+    va_start(va, Format);
+
+#ifdef NO_STRSAFE
+    wvsprintfA(szBuff, Format, va);
```
HBelusca, line 82: Why not using the standard C runtime function _vsnprintf instead? (This wvsprintfA would enforce a dependence on user32.dll, which is not always present for win32 user-mode apps.)

sdk/include/reactos/debug.h
katahiromz: Deleted.

sdk/include/reactos/debug.h
katahiromz: Used `_vsnprintf`.

sdk/include/reactos/debug.h
katahiromz: OK.

sdk/include/reactos/debug.h
```diff
@@ -18,6 +18,10 @@
 #define __RELFILE__ __FILE__
 #endif
 
+#ifdef NDEBUG
+#undef DBG
+#endif
```
HBelusca, line 23: Why doing this? `DBG` is the general macro that tells whether a program is compiled in debug mode. `˙NDEBUG` means, that we don't want the debug macros to show debug output when using debug.h.

katahiromz: Some codes manually define `NDEBUG` and disable the debugging output... I followed this manner.

https://github.com/reactos/reactos/blob/47f6745bcdfb833b0311e052f04e6cd5634660eb/win32ss/gdi/ntgdi/freetype.c#L36-L37

https://github.com/reactos/reactos/blob/47f6745bcdfb833b0311e052f04e6cd5634660eb/win32ss/gdi/ntgdi/bitblt.c#L10-L11

HBelusca: Yes but this means, that if they do an explicit `DPRINT1` (not `DPRINT`) then the DPRINT1 output will be output, which is what is expected. When `NDEBUG` is not defined, which is e.g. usually done when testing something locally, both `DPRINT` and `DPRINT1` will be output (this is one thing I forgot to tell in the previous message). But in all cases, the generic `DBG` define symbol still remains as it is (untouched).

base/applications/mspaint/dialogs.cpp
```diff
@@ -320,15 +320,15 @@ void CFontsDialog::InitFontSizes()
...
+        StringCchPrintf(szText, _countof(szText), TEXT("%d"), s_sizes[i]);
         INT iItem = ComboBox_AddString(hwndSizes, szText);
         if (s_sizes[i] == (INT)registrySettings.PointSize)
             ComboBox_SetCurSel(hwndSizes, iItem);
     }
 
     if (ComboBox_GetCurSel(hwndSizes) == CB_ERR)
     {
-        wsprintf(szText, TEXT("%d"), (INT)registrySettings.PointSize);
+        StringCchPrintf(szText, _countof(szText), TEXT("%d"), (INT)registrySettings.PointSize);
```
HBelusca, line 331: I would suggest you do that in a very separate commit, outside of this PR.

katahiromz: Without these changes, the build will fail.

HBelusca: ah yes, the mspaint would need an explicit `#include <strsafe.h>` in its precompiled header.

sdk/include/reactos/debug.h
```diff
@@ -58,6 +61,73 @@ RtlAssert(
...
+#include <strsafe.h>
+#endif
+
+static inline
+ULONG
+__cdecl
+W32DbgPrint(_In_z_ _Printf_format_string_ PCSTR Format, ...)
+{
+    va_list va;
+    char szBuff[512];
```
HBelusca, line 77: The idea of a user-adjustable `APPDBG_BUFSIZE` value (used here, and in the W32Assert below) sounded nice.

katahiromz: `DEBUG_BUFSIZE`

sdk/include/reactos/debug.h
```diff
@@ -58,6 +61,77 @@ RtlAssert(
...
+{
+    INT id;
+    char szBuff[DEBUG_BUFSIZE];
+
+#ifdef NO_STRSAFE
+    _snprintf(szBuff, _countof(szBuff),
+#else
+    StringCchPrintfA(szBuff, _countof(szBuff),
+#endif
+                     "File '%s',\nLine %ld:\n\n%s\n\n%s",
```
katahiromz, line 118: Is this assertion message alright?

HBelusca: For information, here is what the ntdll RtlAssert calls when displaying the assertion:
```c
    /* Print the assertion */
    DbgPrint("\n*** Assertion failed: %s%s\n"
             "***   Source File: %s, line %lu\n\n",
             Message != NULL ? Message : "",
             (PSTR)FailedAssertion,
             (PSTR)FileName,
             LineNumber);
```
(note the `Message` NULL check)

HBelusca: You can also observe that the win32 CRT has equivalent functions, and you may be inspired by its assertion message:
https://git.reactos.org/?p=reactos.git;a=blob;f=sdk/lib/crt/misc/dbgrpt.cpp;hb=47f6745bcdfb833b0311e052f04e6cd5634660eb#l79
(Note: these can only be used in a win32 user-mode app that explicitly uses the CRT.)

sdk/include/reactos/debug.h
```diff
@@ -58,6 +61,73 @@ RtlAssert(
...
+W32Assert(
+    _In_ PVOID FailedAssertion,
+    _In_ PVOID FileName,
+    _In_ ULONG LineNumber,
+    _In_opt_z_ PCHAR Message)
+{
+    INT id;
+    char szBuff[DEBUG_BUFSIZE];
+
+#ifdef _STRSAFE_H_INCLUDED_
```
HBelusca, line 109: Nice :)

sdk/include/reactos/debug.h
```diff
@@ -58,6 +61,73 @@ RtlAssert(
...
+#else
+    _snprintf(szBuff, _countof(szBuff),
+#endif
+              "File '%s',\nLine %ld:\n\n%s\n\n%s",
+              (LPCSTR)FileName, LineNumber, (LPCSTR)FailedAssertion, Message);
+
+    id = MessageBoxA(NULL, szBuff, "Assertion Failure", MB_ICONERROR | MB_ABORTRETRYIGNORE);
+
+    if (id == IDABORT)
+        ExitProcess((UINT)-1);
```
HBelusca, line 120: or use `ERROR_ASSERTION_FAILURE`

sdk/include/reactos/debug.h
```diff
@@ -58,6 +61,73 @@ RtlAssert(
...
+    INT id;
+    char szBuff[DEBUG_BUFSIZE];
+
+#ifdef _STRSAFE_H_INCLUDED_
+    StringCchPrintfA(szBuff, _countof(szBuff),
+#else
+    _snprintf(szBuff, _countof(szBuff),
+#endif
+              "File '%s',\nLine %ld:\n\n%s\n\n%s",
+              (LPCSTR)FileName, LineNumber, (LPCSTR)FailedAssertion, Message);
```
HBelusca, line 115:
```suggestion
              (PCSTR)FileName, LineNumber, (PCSTR)FailedAssertion, Message);
```

base/applications/mspaint/dialogs.cpp
HBelusca: OK now, you can move both this change and the one in the mspaint's precompiler header into a separate commit, I guess... ? 😄

sdk/include/reactos/debug.h
katahiromz: OK.

sdk/include/reactos/debug.h
katahiromz: Used.

sdk/include/reactos/debug.h
```diff
@@ -58,6 +61,74 @@ RtlAssert(
...
+#ifdef _STRSAFE_H_INCLUDED_
+    StringCchPrintfA(szBuff, _countof(szBuff),
+#else
+    _snprintf(szBuff, _countof(szBuff),
+#endif
+              "*** Assertion failed: %s%s\n"
+              "***   Source File: %s, line %lu",
+              (Message ? Message : ""), (PSTR)FailedAssertion, (PSTR)FileName, LineNumber);
+
+    id = MessageBoxA(NULL, szBuff, "Assertion Failure", MB_ICONERROR | MB_ABORTRETRYIGNORE);
```
HBelusca, line 118: For the messagebox flags I would suggest: `MB_ABORTRETRYIGNORE | MB_ICONERROR | MB_SETFOREGROUND | MB_TASKMODAL` (like in the CRT debugging).
The 3rd flag puts the msgbox in front when it appears, and the last flag makes that "all the top-level windows belonging to the current thread [of the current application] are disabled if the hWnd parameter is NULL", and since the app is crashing, it should "prevent input to other windows in the calling thread without suspending other threads."
(I'm quoting MSDN here.)

sdk/include/reactos/debug.h
```diff
@@ -58,6 +61,74 @@ RtlAssert(
...
+#ifdef _STRSAFE_H_INCLUDED_
+    StringCchVPrintfA(szBuff, _countof(szBuff), Format, va);
+#else
+    _vsnprintf(szBuff, _countof(szBuff), Format, va);
+#endif
+
+    OutputDebugStringA(szBuff);
+    va_end(va);
+
+    return 0; /* Return STATUS_SUCCESS, since we are supposed to mimic DbgPrint */
```
HBelusca, line 90: Win32 equivalent is `ERROR_SUCCESS`.

sdk/include/reactos/debug.h
```diff
@@ -58,6 +61,74 @@ RtlAssert(
...
+
+#ifdef _STRSAFE_H_INCLUDED_
+    StringCchPrintfA(szBuff, _countof(szBuff),
+#else
+    _snprintf(szBuff, _countof(szBuff),
+#endif
+              "*** Assertion failed: %s%s\n"
+              "***   Source File: %s, line %lu",
+              (Message ? Message : ""), (PCSTR)FailedAssertion, (PCSTR)FileName, LineNumber);
+
```
HBelusca, line 117: It could be of interest to also send this message to dprint (besides popping up the message box) in case the application runs under a debugger.

sdk/include/reactos/debug.h
```diff
@@ -58,6 +61,78 @@ RtlAssert(
...
+    StringCchPrintfA(szBuff, _countof(szBuff),
+#else
+    _snprintf(szBuff, _countof(szBuff),
+#endif
+              "*** Assertion failed: %s%s\n"
+              "***   Source File: %s, line %lu\n\n"
+              "(Press Retry to debug the application)\n",
+              (Message ? Message : ""), (PCSTR)FailedAssertion, (PCSTR)FileName, LineNumber);
+
+    OutputDebugStringA(szBuff);
```
HBelusca, line 119: This is what I was thinking about regarding "sending this message to dprint in case the app runs under a debugger":
```suggestion
    if (IsDebuggerPresent())
        OutputDebugStringA(szBuff);
```
(inspired from the CRT debug helpers)

base/applications/mspaint/dialogs.cpp
katahiromz: Reverted.

dll/cpl/mmsys/mmsys.h
```diff
@@ -6,6 +6,7 @@
 #define WIN32_NO_STATUS
 #define _INC_WINDOWS
 #define COM_NO_WINDOWS_H
+#define NTOS_MODE_USER
```
HBelusca, line 9: The CPL is a GUI program, it should not use the NTDLL debugging stuff.

win32ss/drivers/miniport/vbe/vbemp.h
```diff
@@ -23,6 +23,7 @@
 
 /* INCLUDES *******************************************************************/
 
+#define NTOS_MODE_USER
```
HBelusca, line 26: No! This is a driver!

win32ss/drivers/miniport/pc98vid/pc98vid.h
```diff
@@ -8,6 +8,7 @@
 #ifndef _PC98VID_PCH_
 #define _PC98VID_PCH_
 
+#define NTOS_MODE_USER
```
HBelusca, line 11: No! This is a driver!

win32ss/drivers/font/ftfd/ftfd.h
```diff
@@ -8,6 +8,7 @@
 #ifndef _FTFD_PCH_
 #define _FTFD_PCH_
 
+#define NTOS_MODE_USER
```
HBelusca, line 11: No! This is a driver!

win32ss/drivers/miniport/vga/vgamp.h
```diff
@@ -23,6 +23,7 @@
 
 /* INCLUDES *******************************************************************/
 
+#define NTOS_MODE_USER
```
HBelusca, line 26: No! This is a driver!

win32ss/drivers/miniport/vga_new/vga.h
```diff
@@ -10,6 +10,7 @@
 #ifndef _VGA_NEW_PCH_
 #define _VGA_NEW_PCH_
 
+#define NTOS_MODE_USER
```
HBelusca, line 13: No! This is a driver!

win32ss/drivers/miniport/vmx_svga/precomp.h
```diff
@@ -6,6 +6,7 @@
  * PROGRAMMERS:     ReactOS Portable Systems Group
  */
 
+#define NTOS_MODE_USER
```
HBelusca, line 9: No! This is a driver!

win32ss/drivers/miniport/xboxvmp/xboxvmp.h
```diff
@@ -15,6 +15,7 @@
  * FIXME: specify headers properly in the triangle brackets and rearrange them
  * in a way so it would be simpler to add NDK and other headers for debugging.
  */
+#define NTOS_MODE_USER
```
HBelusca, line 18: No! This is a driver!

HBelusca (changes requested):

dll/win32/kernel32/k32.h
```diff
@@ -11,6 +11,7 @@
 
 /* INCLUDES ******************************************************************/
 
+#define NTOS_MODE_USER
```
HBelusca, line 14: Please move this just above the first `#include <ndk/...>` below (line 35)

dll/win32/loadperf/loadperf_main.c
```diff
@@ -18,7 +18,7 @@
  * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
  */
 
-
+#define NTOS_MODE_USER
```
HBelusca, line 21: Should not be necessary.

sdk/lib/cportlib/cport.c
```diff
@@ -30,6 +30,7 @@
 
 /* INCLUDES *******************************************************************/
 
+#define NTOS_MODE_USER
```
HBelusca, line 33: Problematic, as this library is only used in kernel mode.

dll/win32/authz/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@ add_library(authz MODULE
     ${CMAKE_CURRENT_BINARY_DIR}/authz.def)
 
 set_module_type(authz win32dll UNICODE)
-add_importlibs(authz advapi32 msvcrt kernel32 ntdll)
+add_importlibs(authz advapi32 msvcrt kernel32 user32 ntdll)
```
HBelusca, line 17: This dll should never depend on user32.

dll/win32/authz/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@ add_library(authz MODULE
...
+target_link_libraries(authz wine)
```
HBelusca, line 18: Should not be needed (we don't use any wine stuff in authz)

dll/win32/mswsock/CMakeLists.txt
```diff
@@ -16,6 +16,6 @@ add_library(mswsock MODULE
     ${CMAKE_CURRENT_BINARY_DIR}/mswsock.def)
 
 set_module_type(mswsock win32dll UNICODE)
-add_importlibs(mswsock ws2_32 advapi32 dnsapi msvcrt kernel32 ntdll)
+add_importlibs(mswsock ws2_32 advapi32 dnsapi msvcrt kernel32 user32)
```
HBelusca, line 19: On Windows: there is a ntdll dependence (it is there because the NT IO functions are used), and user32 is delay-loaded, only when necessary.

drivers/wdm/audio/filters/kmixer/kmixer.h
```diff
@@ -1,6 +1,7 @@
 #ifndef _KMIXER_PCH_
 #define _KMIXER_PCH_
 
+#define NTOS_MODE_USER
```
HBelusca, line 4: That's a driver!

HBelusca (commented):
Remarks:
- I've added commits that fix aspects of some of the previous commits of this PR.
- the CPORTLIB and kernel32 changes have been manually merged to master: your branch will need to be rebased (and the potential minimal conflicts solved).
- It would be nice to clean up all the "fix build" etc. commits into an intelligible set of commits that touch logically the different modules.

sdk/include/reactos/debug.h
```diff
@@ -18,9 +18,6 @@
 #define __RELFILE__ __FILE__
```
learn-more, line 18: I modified this header to only use ntdll, and not kernel32.dll so that we can use this in all modules, instead of having 2 or 3 different versions of this file.

Shims and verifier dll's cannot import from kernel32!

What is the reason to add the kernel32 dependency back?

learn-more (changes requested):
Please verify that both shims and verifier dll's still work correctly, and do NOT link against kernel32.

dll/win32/mswsock/nsplookup.c
```diff
@@ -14,8 +14,8 @@
 
 #include "mswhelper.h"
 
-#define NDEBUG
-#include <debug.h>
+#include <assert.h>
+#define ASSERT assert
```
ThFabba, line 18: Not a fan of this. `msvcrt`'s assert results in annoying popups instead of just breaking into the debugger. Using it instead of `RtlAssert` may result in testbot timeouts without a backtrace, rather than an immediate debugger break that our test infrastructure knows how to handle.

sdk/include/reactos/debug.h
ThFabba: Yeah I'm not a huge fan of the idea, either. `ntdll` is always safe to import from, `kernel32` is not. So you're adding extra complexity to the code just to avoid ntdll imports. Because it's "cleaner", by some definition thereof?
Dependencies on ntdll don't hurt because it's always loaded in every process anyway. This fact isn't officially documented for Win32 applications, but (a) it's not something Microsoft can change at this point without breaking a ton of apps, (b) we're writing an OS anyway and a ton of our code relies on undocumented OS internals.

An argument can be made that applications should use kernel32 instead of ntdll, because they make sense on their own, even if you take them away from ROS. But for system DLLs (which seems to be what most of the affected modules here are) I see more potential problems than benefits.

katahiromz: `IsDebuggerPresent` and `OutputDebugString` are of `kernel32`.

katahiromz:
> do NOT link against kernel32.

Impossible. Closed.

PR #4273, opened 2022-01-11
[NEWDEV][SHELL32][DEVMGR]  Update / Small improvements to pt-PT translation

zecarlos1957:
[NEWDEV][SHELL32][DEVMGR]  Update / Small improvements to pt-PT translation


binarymaster (approved):

dll/win32/devmgr/lang/pt-PT.rc
```diff
@@ -7,10 +7,10 @@ BEGIN
     IDS_NAME "Nome"
     IDS_TYPE "Tipo"
     IDS_MANUFACTURER "Fabricante: %1"
-    IDS_LOCATION "Local: %1"
-    IDS_STATUS "Estado do dispositivo: %1"
+    IDS_LOCATION "Localização: %1"
+    IDS_STATUS "Estado do\ndispositivo: %1"
```
binarymaster, line 11: Are you sure about using "\n" line break here? Other translations does not seem to do that:

image

binarymaster, line 11: Are you sure about using "\n" line break here? Other translations does not seem to do that:

image

zecarlos1957: ReactOS
ReactOS2

Improved visual arrangement

binarymaster: The "before" looks better for me. It's also the reason why this label supports word wrapping.

zecarlos1957: OK. I accept the majority opinion. I like the altered version as it is more aligned .

binarymaster: I'm pretty sure this is how it's done in Windows, might be a good idea to double check.

zecarlos1957: Maybe I have a better solution for both.
"Estado disp.." Like in Win7 pt-PT

binarymaster: Maybe the best option would be to resize the underlying dialog box itself.

zecarlos1957: I like that solution.

binarymaster: Hmm... I actually tried to resize the `IDD_HARDWARE` dialog box, and it seems it's not possible, as the property sheet is then resized automatically.

While checking, I looked up the dialog in Restorator, and noticed there's also problem with "fix conflicts" button, which is usually hidden. Here's the fix for it:
```
PUSHBUTTON "&Resolução de problemas...", IDC_TROUBLESHOOT, 42, 140, 97, 14, BS_PUSHBUTTON | WS_CHILD | WS_DISABLED | WS_TABSTOP
```

binarymaster:
Ok let's leave that label as is for now, I also pushed the fix for the button.

PR #4274, opened 2022-01-11
[NEWDEV] Fix Update Driver Wizard Dialog position

zecarlos1957:
Observed behavior:
        The Update Driver Wizard Dialog is not completely visible.
		Part of the window is off the screen.

Expected behavior:
        As in Windows XP the dialog box is positioned slightly below
		and to the right of the parent window (Z order).

### DIFF
--- a/dll/win32/newdev/wizard.c
+++ b/dll/win32/newdev/wizard.c
@@ -25,32 +25,8 @@
 #include <shlobj.h>
 #include <shlwapi.h>
 
+
 HANDLE hThread;
-
-static VOID
-CenterWindow(
-    IN HWND hWnd)
-{
-    HWND hWndParent;
-    RECT rcParent;
-    RECT rcWindow;
-
-    hWndParent = GetParent(hWnd);
-    if (hWndParent == NULL)
-        hWndParent = GetDesktopWindow();
-
-    GetWindowRect(hWndParent, &rcParent);
-    GetWindowRect(hWnd, &rcWindow);
-
-    SetWindowPos(
-        hWnd,
-        HWND_TOP,
-        ((rcParent.right - rcParent.left) - (rcWindow.right - rcWindow.left)) / 2,
-        ((rcParent.bottom - rcParent.top) - (rcWindow.bottom - rcWindow.top)) / 2,
-        0,
-        0,
-        SWP_NOSIZE);
-}
 
 static BOOL
 SetFailedInstall(
@@ -482,9 +458,6 @@ WelcomeDlgProc(
 
             hwndControl = GetParent(hwndDlg);
 
-            /* Center the wizard window */
-            CenterWindow(hwndControl);
-
             /* Hide the system menu */
             dwStyle = GetWindowLongPtr(hwndControl, GWL_STYLE);
             SetWindowLongPtr(hwndControl, GWL_STYLE, dwStyle & ~WS_SYSMENU);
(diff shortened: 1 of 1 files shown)


HBelusca:
> The Update Driver Wizard Dialog is not completely visible.

Do you have a screenshot to prove that?

zecarlos1957:
Issue
Fixed

dll/win32/newdev/wizard.c
```diff
@@ -25,32 +25,8 @@
...
-
-    SetWindowPos(
-        hWnd,
-        HWND_TOP,
-        ((rcParent.right - rcParent.left) - (rcWindow.right - rcWindow.left)) / 2,
-        ((rcParent.bottom - rcParent.top) - (rcWindow.bottom - rcWindow.top)) / 2,
-        0,
-        0,
-        SWP_NOSIZE);
-}
```
binarymaster, line 53: This function was introduced by @christophvw back in 2005 in commit 775e2dbd070abe62b04c244db1ee82bc4d4aac77 (r19070). Maybe it was needed as a hack to make the window centered.

HBelusca: The point is that windows never appear centered, unless being told so (with such code, for example). The question is whether we want this; does Windows also show the wizard always centered?

HBelusca: Also here I think that the function should be more clever.

https://git.reactos.org/?p=reactos.git;a=blob;f=base/applications/msconfig_new/utils.c;hb=9a523c81356bb0349b7e545611bcad8ca3325b69#l161

https://social.msdn.microsoft.com/Forums/vstudio/en-US/54f7f131-5d15-4b2d-863b-bef329e89a32/how-to-display-the-window-center-win32-api?forum=vcgeneral

https://git.reactos.org/?p=reactos.git;a=blob;f=modules/rosapps/applications/explorer-old/utility/utility.cpp;hb=9a523c81356bb0349b7e545611bcad8ca3325b69#l48

https://www.codeproject.com/Tips/250672/CenterWindow-in-WIN32

HBelusca:
(Screenshots)
Hmm OK the first picture looks unusual. I have never seen the wizard popping up on the top left like that with cropped left part; it always showed up centered on the screen for me, and this was the intent of the CenterWindow function that you removed....
.... except, indeed, that this function will first attempt to center the wizard _**with respect to its parent window**_, and only if there is no parent window, it will center it _**with respect to the desktop**_ . So the problem may be due to the first part ("centering with respect to its parent window") where a little bit of more non-trivial work may be required to ensure the wizard doesn't show up partly cropped.

zecarlos1957:
Sem título
On my WindowsXP the driver update dialog never appears centered

dll/win32/newdev/wizard.c
```diff
@@ -42,6 +42,11 @@ CenterWindow(
     GetWindowRect(hWndParent, &rcParent);
     GetWindowRect(hWnd, &rcWindow);
 
+    /// Check if the child window fits inside the parent window
+    if (rcWindow.left < rcParent.left || rcWindow.top < rcParent.top ||
+        rcWindow.right > rcParent.right || rcWindow.bottom > rcParent.bottom)
+        return;
```
HBelusca, line 48:
```suggestion
    /* Check if the child window fits inside the parent window */
    if (rcWindow.left < rcParent.left || rcWindow.top < rcParent.top ||
        rcWindow.right > rcParent.right || rcWindow.bottom > rcParent.bottom)
    {
        return;
    }
```

HBelusca (approved):

PR #4278, opened 2022-01-12
[ADVAPI32] Registry: Use inline functions instead of macros

TAN-Gaming:
## Purpose

Use inline functions instead of macros


JoachimHenze:
I verified these changes to compile also with the following older toolchains:
RosBEWin2.1.6 GCC4.7.2 dbg
RosBEWin2.1.6 GCC4.7.2 rls
RosBEWin2.1.6 MSVC2010 dbg
RosBEWin2.1.6 MSVC2010 rls
**Can not say anything about even older compilers though.**
That specific file reg.c does not exist in Wine head. In case parts of it would have been taken from Wine, then it would be from different filenames in Wines advapi32. No urgent need to upstream therefore.

I do still wonder why we'd want to change that just for the sake of change.
Macros aren't necessarily bad. Those used here should be rather safe.
So I see no immediate benefit from that refactoring proposal.

HBelusca:
What do we gain with this?

dll/win32/advapi32/reg/reg.c
```diff
@@ -39,30 +39,41 @@ static BOOLEAN DllInitialized = FALSE; /* HACK */
...
+}
+
+static inline BOOL IsPredefKey(HKEY HKey)
+{
+    return (((ULONG_PTR)(HKey) & 0xF0000000) == 0x80000000);
+}
+
+static inline ULONG GetPredefKeyIndex(HKEY HKey)
+{
+    return ((ULONG_PTR)(HKey) & 0x0FFFFFFF);
```
JoachimHenze, line 66: The retval is ULONG.
Is that implicit ULONG_PTR to ULONG cast for the retval even x64 safe? If not, then it's a risk for crashes / random behavior. cc Timo Kreuzer.

JoachimHenze, line 66: The retval is ULONG.
Is that implicit ULONG_PTR to ULONG cast for the retval even x64 safe? If not, then it's a risk for crashes / random behavior. cc Timo Kreuzer.

JoachimHenze: Maybe it is, because the 0x0FFFFFFF is short enough. But still, it does not really seem to make anything safer for me.

ThFabba: Yes, this is safe and correct. Casting `HKey` (a pointer) to `ULONG` would give you a warning on x64.
You _could_ either double-cast (`(ULONG)(ULONG_PTR)HKey & 0x0FFFFFFF`) or cast the result (`(ULONG)((ULONG_PTR)HKey & 0x0FFFFFFF)`), but neither is necessary because MSVC knows that the result of the `&` will fit in a `ULONG`, so it won't issue a warning.

dll/win32/advapi32/reg/reg.c
```diff
@@ -39,30 +39,41 @@ static BOOLEAN DllInitialized = FALSE; /* HACK */
...
-    ((ULONG_PTR)(HKey) & 0x0FFFFFFF)
-
 static NTSTATUS OpenClassesRootKey(PHANDLE KeyHandle);
-static NTSTATUS OpenLocalMachineKey (PHANDLE KeyHandle);
-static NTSTATUS OpenUsersKey (PHANDLE KeyHandle);
+static NTSTATUS OpenLocalMachineKey(PHANDLE KeyHandle);
+static NTSTATUS OpenUsersKey(PHANDLE KeyHandle);
 static NTSTATUS OpenCurrentConfigKey(PHANDLE KeyHandle);
 
+/* INLINE FUNCTIONS *********************************************************/
```
HBelusca, line 49: Not sure it's useful to have this distinction between "inline functions" vs. "functions" in these decorative titles.

TAN-Gaming: Maybe not, I'll revert.

dll/win32/advapi32/reg/reg.c
JoachimHenze: You say MSVC does understand that it is safe, okay.
But what about GCC x64? Will that be that smart as well to **not warn**, or does GCC x64 need the explicit double-cast for silence?

learn-more:
No input from author.

PR #4279, opened 2022-01-12
[SHELL32] Folders: Adjust column sequence in details view

thc13:
## Purpose

Adjust column sequence in folder view to match WinXP and Win2k3 order.

JIRA issue: [CORE-11846](https://jira.reactos.org/browse/CORE-11846)

**Before:**
image

**After:**
image

**Win2k3:**
image

**WinXP:**
image

## ToDo

- [ ] Do not use magic numbers for column sequence number. **Does any dev have preference or statement, which header file should be used?**

### DIFF
--- a/dll/win32/shell32/folders/CControlPanelFolder.cpp
+++ b/dll/win32/shell32/folders/CControlPanelFolder.cpp
@@ -333,7 +333,7 @@ HRESULT WINAPI CControlPanelFolder::CompareIDs(LPARAM lParam, PCUIDLIST_RELATIVE
         case 0:        /* name */
             result = wcsicmp(pData1->szName + pData1->offsDispName, pData2->szName + pData2->offsDispName);
             break;
-        case 1:        /* comment */
+        case 4:        /* comment */
             result = wcsicmp(pData1->szName + pData1->offsComment, pData2->szName + pData2->offsComment);
             break;
         default:
@@ -581,7 +581,7 @@ HRESULT WINAPI CControlPanelFolder::GetDetailsOf(PCUITEMID_CHILD pidl, UINT iCol
         {
             case 0:        /* name */
                 return SHSetStrRet(&psd->str, pCPanel->szName + pCPanel->offsDispName);
-            case 1:        /* comment */
+            case 4:        /* comment */
                 return SHSetStrRet(&psd->str, pCPanel->szName + pCPanel->offsComment);
         }
     }
--- a/dll/win32/shell32/folders/CDrivesFolder.cpp
+++ b/dll/win32/shell32/folders/CDrivesFolder.cpp
@@ -528,10 +528,10 @@ class CDrivesFolderEnum :
 
 static const shvheader MyComputerSFHeader[] = {
     {IDS_SHV_COLUMN_NAME, SHCOLSTATE_TYPE_STR | SHCOLSTATE_ONBYDEFAULT, LVCFMT_LEFT, 15},
-    {IDS_SHV_COLUMN_COMMENTS, SHCOLSTATE_TYPE_STR, LVCFMT_LEFT, 10},
     {IDS_SHV_COLUMN_TYPE, SHCOLSTATE_TYPE_STR | SHCOLSTATE_ONBYDEFAULT, LVCFMT_LEFT, 10},
     {IDS_SHV_COLUMN_DISK_CAPACITY, SHCOLSTATE_TYPE_STR | SHCOLSTATE_ONBYDEFAULT, LVCFMT_RIGHT, 10},
     {IDS_SHV_COLUMN_DISK_AVAILABLE, SHCOLSTATE_TYPE_STR | SHCOLSTATE_ONBYDEFAULT, LVCFMT_RIGHT, 10},
+    {IDS_SHV_COLUMN_COMMENTS, SHCOLSTATE_TYPE_STR, LVCFMT_LEFT, 10},
 };
 
 #define MYCOMPUTERSHELLVIEWCOLUMNS 5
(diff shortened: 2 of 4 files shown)


HBelusca:
I would ask you to thoroughly investigate how this is done on Windows. There are certainly registry values that control this, and if you find these, you can check what the behaviour becomes when you remove them (and thus letting the shell use its defaults).
As long as you haven't investigated this, I would say that it's better keeping our defaults as they are.

thc13:
> I would ask you to thoroughly investigate how this is done on Windows. There are certainly registry values that control this, and if you find these, you can check what the behaviour becomes when you remove them (and thus letting the shell use its defaults). As long as you haven't investigated this, I would say that it's better keeping our defaults as they are.

Please look into the ticket, i am aware of that registry is used, to store column size, order, ..., but as written in ticket it is out of scope of the ticket and is an improvement or feature request (FR). I tried to only focus on what ticket requested. I am willing to do the investigation for an improvement or FR. May actual goal is to address bugs and to reduce bug report count, as often read on chat, that this is (should be) one of the biggest goals atm. Additionally FR where closed in the past as "wont fix" iirc, explained with ticket count.

I also agree to this goal, because it makes project management much easier and votes or watchers also indicate somehow priority of FR. Do not get me wrong, i do not say that all devs should do so, because everybody volunteers and should pick his own favorite task. So let us focus on ticket content and do improvements later. Especially in this task doing registry handling according to WinXP or/and Win2k3 is a bigger task, other folder views have multiple possible other columns and every column has an option to hide/unhide. So i really think to do it correctly is out of scope of the ticket, but i am willing to do it in future.

HBelusca:
> I tried to only focus on what ticket requested.

Nowhere in the ticket was mentioned the question of the ordering of columns. The ticket was about that the contents of the columns was wrong, and this was fixed long ago.

The ordering of columns was mentioned in a comment (by Pi_User5) in passing as an after-thought, and therefore was not the actual point of the ticket.

> I am willing to do the investigation for an improvement or FR. May actual goal is to address bugs and to reduce bug report count, as often read on chat, that this is (should be) one of the biggest goals atm.

The default ordering of columns is not a bug; it's merely a "convenience".

> I also agree to this goal, because it makes project management much easier and votes or watchers also indicate somehow priority of FR. Do not get me wrong, i do not say that all devs should do so, because everybody volunteers and should pick his own favorite task. So let us focus on ticket content and do improvements later. Especially in this task doing registry handling according to WinXP or/and Win2k3 is a bigger task, other folder views have multiple possible other columns and every column has an option to hide/unhide. So i really think to do it correctly is out of scope of the ticket, but i am willing to do it in future.

We actually want to reduce the number of hacks, as well as the number of modifications done that may have to be reverted later because it is discovered that they were wrong; and that if a more thorough investigation was done to begin with, such modifications would have not been necessary.

For example (and in regards to my previous comment), suppose that we discover that the default behaviour of Windows, when all the shell registry settings related to the columns are removed, is exactly the same as what we currently do. And that the observed so-called "default ordering" seen when you install Windows is just because there are added registry settings that change the internal defaults of the shell.
What would we say then, in that case?

thc13:
> > I tried to only focus on what ticket requested.
> 
> Nowhere in the ticket was mentioned the question of the ordering of columns. The ticket was about that the contents of the columns was wrong, and this was fixed long ago.
Correct and it is kept open, because of an comment what is missing, and that part i added.
> 
> The ordering of columns was mentioned in a comment (by Pi_User5) in passing as an after-thought, and therefore was not the actual point of the ticket.
> 
> > I am willing to do the investigation for an improvement or FR. May actual goal is to address bugs and to reduce bug report count, as often read on chat, that this is (should be) one of the biggest goals atm.
> 
> The default ordering of columns is not a bug; it's merely a "convenience".
> 
Correct, no reason to keep the ticket open, a new ticket with better and correct description is needed to track the rest.
> > I also agree to this goal, because it makes project management much easier and votes or watchers also indicate somehow priority of FR. Do not get me wrong, i do not say that all devs should do so, because everybody volunteers and should pick his own favorite task. So let us focus on ticket content and do improvements later. Especially in this task doing registry handling according to WinXP or/and Win2k3 is a bigger task, other folder views have multiple possible other columns and every column has an option to hide/unhide. So i really think to do it correctly is out of scope of the ticket, but i am willing to do it in future.
> 
> We actually want to reduce the number of hacks, as well as the number of modifications done that may have to be reverted later because it is discovered that they were wrong; and that if a more thorough investigation was done to begin with, such modifications would have not been necessary.
> 
This may or may not be a hack and as i stated a leastl magic numbers should be avoided, because switch case relying on that is spread over multiple files, which sooner or later lead to problems. But it is up to you to close this PR and the ticket, as you said it is a hack and ticket is solved since years. IMHO keeping "useless tickets open, because in a comment there maybe an information which is not done yet is against the policy to focus on 1 problem/issue per ticket.
> For example (and in regards to my previous comment), suppose that we discover that the default behaviour of Windows, when all the shell registry settings related to the columns are removed, is exactly the same as what we currently do. And that the observed so-called "default ordering" seen when you install Windows is just because there are added registry settings that change the internal defaults of the shell. What would we say then, in that case?

I fully agree, that it needs more investigation, but this is out of scope of the ticket. I wanted to offer my time and knowledge to clean up things in code and tickets, but if you think it is not needed, then i have no problem to focus on other things and spend my time elsewhere.

binarymaster:
@thc13 recently I started to check commit author fields that are being merged into master branch, and I noticed you have an autogenerated email address by GitHub.

You need to uncheck this option in your [GitHub email settings](https://github.com/settings/emails):

image

So when maintainers use "Squash and merge" button, the commit would end up merged with correct fields.

PR #4281, opened 2022-01-13
[CMD][WINSRV] Workaround when Asian fixed-pitch font was absent

katahiromz:
## Purpose
If the user was CJK (Chinese, Japanese, or Korean; Asian), CJK Command Prompt has a readability problem when the CJK fixed-pitch font was absent.

JIRA issue: N/A

## Proposed changes

- Set a property `"ReactOSFontFallback"` to the console window if the CJK fixed-pitch font was not found.
- If the property `"ReactOSFontFallback"` of the console window is set, then show the warning message, and then set the thread locale to English.

## Screenshots
BEFORE:
before
AFTER:
after

### DIFF
--- a/base/shell/cmd/cmd.c
+++ b/base/shell/cmd/cmd.c
@@ -2311,7 +2311,7 @@ Initialize(VOID)
 
     /* ReactOS extension: Inform if the Asian font is not found */
     hwndConsole = GetConsoleWindow();
-    if (GetPropW(hwndConsole, L"ReactOSCJKFontFallback"))
+    if (GetPropW(hwndConsole, L"ReactOSFontFallback"))
     {
         ConOutPuts(_T("WARNING: The Asian fixed-pitch font not found. Falling back to English.\n"));
 
@@ -2320,7 +2320,7 @@ Initialize(VOID)
         SetThreadLocale(MAKELCID(ENGLISH_LANGID, SORT_DEFAULT));
 #undef ENGLISH_LANGID
 
-        RemovePropW(hwndConsole, L"ReactOSCJKFontFallback");
+        RemovePropW(hwndConsole, L"ReactOSFontFallback");
     }
 
     if (AutoRun)
--- a/base/shell/cmd/cmd.c
+++ b/base/shell/cmd/cmd.c
@@ -2140,6 +2140,7 @@ Initialize(VOID)
     TCHAR option = 0;
     BOOL AutoRun = TRUE;
     TCHAR ModuleName[MAX_PATH + 1];
+    HWND hwndConsole;
 
     /* Get version information */
     InitOSVersion();
@@ -2306,6 +2307,21 @@ Initialize(VOID)
                         _T(KERNEL_VERSION_STR),
                         _T(KERNEL_VERSION_BUILD_STR));
         ConOutPuts(_T("(C) Copyright 1998-") _T(COPYRIGHT_YEAR) _T(" ReactOS Team.\n"));
+    }
+
+    /* ReactOS extension: Inform if the Asian font is not found */
+    hwndConsole = GetConsoleWindow();
+    if (GetPropW(hwndConsole, L"ReactOSCJKFontFallback"))
+    {
+        if (!*ptr)
+        {
+            ConOutPuts(_T("WARNING: The Asian fixed-pitch font was not found. It goes English.\n"));
+        }
+
+        /* Set the thread locale to English */
+#define ENGLISH_LANGID MAKELANGID(LANG_ENGLISH, SUBLANG_ENGLISH_US)
+        SetThreadLocale(MAKELCID(ENGLISH_LANGID, SORT_DEFAULT));
+#undef ENGLISH_LANGID
     }
 
     if (AutoRun)
--- a/base/shell/cmd/cmd.c
+++ b/base/shell/cmd/cmd.c
@@ -2315,7 +2315,7 @@ Initialize(VOID)
     {
         if (!*ptr)
         {
-            ConOutPuts(_T("WARNING: The Asian fixed-pitch font was not found. It goes English.\n"));
+            ConOutPuts(_T("WARNING: The Asian fixed-pitch font not found. Falling back to English.\n"));
         }
 
         /* Set the thread locale to English */
--- a/base/shell/cmd/cmd.c
+++ b/base/shell/cmd/cmd.c
@@ -2126,6 +2126,18 @@ GetCmdLineCommand(
     _tcscpy(commandline, ptr);
 }
 
+/* Is the user Chinese, Japanese or Korean? */
+static BOOL IsUserCJK(VOID)
+{
+    switch (PRIMARYLANGID(GetUserDefaultLangID()))
+    {
+        case LANG_CHINESE:
+        case LANG_JAPANESE:
+        case LANG_KOREAN:
+            return TRUE;
+    }
+    return FALSE;
+}
 
 /*
  * Set up global initializations and process parameters.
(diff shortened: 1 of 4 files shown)


base/shell/cmd/cmd.c
```diff
@@ -2308,6 +2309,23 @@ Initialize(VOID)
         ConOutPuts(_T("(C) Copyright 1998-") _T(COPYRIGHT_YEAR) _T(" ReactOS Team.\n"));
     }
 
+    /* ReactOS extension: Inform if the Asian font is not found */
+    hwndConsole = GetConsoleWindow();
+    if (GetPropW(hwndConsole, L"ReactOSCJKFontFallback"))
+    {
+        if (!*ptr)
+        {
```
binarymaster, line 2317: The fallback doesn't seem to depend on this variable, so you can put the warning message without checking it.

base/shell/cmd/cmd.c
```diff
@@ -2308,6 +2309,23 @@ Initialize(VOID)
...
+            ConOutPuts(_T("WARNING: The Asian fixed-pitch font was not found. It goes English.\n"));
```
binarymaster, line 2318:
```suggestion
            ConOutPuts(_T("WARNING: The Asian fixed-pitch font not found. Falling back to English.\n"));
```

win32ss/user/winsrv/concfg/font.c
```diff
@@ -40,16 +40,20 @@ CodePageToCharSet(
 }
 
 HFONT
-CreateConsoleFontEx(
+CreateConsoleFontExEx(
```
binarymaster, line 43: Suggestion to call the function `CreateConsoleFontInternal` (instead of ExEx) 😂

binarymaster:
And also a question, does Windows perform a fallback when no monospace Asian font found?

win32ss/user/winsrv/concfg/font.c
katahiromz: OK.

katahiromz:
> does Windows perform a fallback?

I think no. WinMe didn't fallback nor showed message.
VirtualBox_Windows Me_13_01_2022_21_11_08

katahiromz:
However, the command `chcp 437` of Japanese Windows makes the fallback to English.

base/shell/cmd/cmd.c
```diff
@@ -2308,6 +2309,20 @@ Initialize(VOID)
...
+    {
+        ConOutPuts(_T("WARNING: The Asian fixed-pitch font not found. Falling back to English.\n"));
+
+        /* Set the thread locale to English */
+#define ENGLISH_LANGID MAKELANGID(LANG_ENGLISH, SUBLANG_ENGLISH_US)
+        SetThreadLocale(MAKELCID(ENGLISH_LANGID, SORT_DEFAULT));
+#undef ENGLISH_LANGID
+
+        RemovePropW(hwndConsole, L"ReactOSFontFallback");
+    }
```
HBelusca, line 2324: The problem with that is, what if a user doesn't run cmd.exe, but just run any other command-line utility?

win32ss/user/winsrv/concfg/font.c
```diff
@@ -70,9 +74,10 @@ CreateConsoleFontEx(
 
     if (!IsValidConsoleFont(FaceName, CodePage))
     {
-        StringCchCopyW(FaceName, LF_FACESIZE, L"Terminal");
```
HBelusca, line 73: So now, what font is it going to use if `IsValidConsoleFont` fails? The `lf.lfFaceName` below would be initialized with the name of that invalid font.

win32ss/user/winsrv/consrv/frontends/gui/conwnd.c
```diff
@@ -3,7 +3,7 @@
  * PROJECT:         ReactOS Console Server DLL
  * FILE:            win32ss/user/winsrv/consrv/frontends/gui/conwnd.c
  * PURPOSE:         GUI Console Window Class
- * PROGRAMMERS:     G� van Geldorp
+ * PROGRAMMERS:     Gé van Geldorp
```
HBelusca, line 6: Was this changed due to UTF8? (if so, it's OK)

HBelusca:
> However, the command `chcp 437` of Japanese Windows makes the font fallback to English. 

But on Windows, the console server (winsrv) knows what fonts are available and what fonts are not (and whether they support CJK or not), and from this, can do appropriate action when chcp tries to change the codepage (including, refusing to change it altogether if there is no font to satisfy the codepage).

katahiromz:
> Better check on XP or 2003 by deleting the fonts.

Sorry, I can't.

JoachimHenze:
Katayama, Just an idea: What about instead
in 2nd stage setup do perform a check whether CJK was selected for install, and if yes, then make it automatically download the proper fonts like the gecko download is currently done automatically. Wouldn't that solve the problem more elegantly (and even more user friendly)?
And that even without having to touch integral parts of the OS, where some of them are constantly loaded in memory?

katahiromz:
Henze, that is a good idea. I want to realize it. But Chinese and Korean packages are not finished, and you know the CJK font is big. The user might cancel the download or might be disconnected to Internet. So, the workaround like this PR is needed.

HBelusca:
We should also note, that even if the fonts were present, one could always do the "little experiment" of manually deleting the fonts afterwards and see how the ROS console would behave when it couldn't use said fonts, and we are back to the problem of "what should be the fallback behaviour" in this case, which was in a sense the point of this PR. Excepting that the fallback behaviour would be done by winsrv/consrv, no cmd.exe messages would be displayed, and winsrv/consrv would refuse to change the codepage to one for which no suitable font for display does exist.

win32ss/user/winsrv/consrv/frontends/gui/conwnd.c
katahiromz: Yes.

win32ss/user/winsrv/concfg/font.c
katahiromz:
```txt
    if (!IsValidConsoleFont(FaceName, CodePage))
    {
        if (IsCJKCodePage(CodePage))
        {
            MessageBoxW(NULL, FaceName, NULL, 0); // Added
            if (pbCJKFontFallback)
                *pbCJKFontFallback = TRUE;
            lf.lfCharSet = ANSI_CHARSET;
        }
    }
```

base/shell/cmd/cmd.c
katahiromz: Trying some executables, I found it doesn't affect the other command-line utilities:
VirtualBox_ReactOS_14_01_2022_08_38_38

win32ss/user/winsrv/concfg/font.c
HBelusca: Let's please stop piling hacks in top of my code, plz...

JoachimHenze:
I don't know... One could argue that would require some kind of fallback mechanism being needed then for virtually anything not being present. Any dll, any .txt file, whatever then. I think this is overkill. If a font doesn't render correctly, like in this case, then it's pretty obvious that some font is missing. Just my 2 cents.

katahiromz:
> pretty obvious that some font is missing

The required font is Asian fixed-pitch font. Requiring "fixed-pitch" is not obvious from normal users. And this PR critically affects usability of Asian users.

katahiromz:
See real CJK users:
https://twitter.com/driver1998/status/1248918696392708096

HBelusca:
Just for your information, I did start working on [CORE-17601](https://jira.reactos.org/browse/CORE-17601) and one of its aim was also to have a correct implementation for what this PR is trying to do (namely: the console falling back to a default codepage, and setting the app's thread locale to a default value, in case there is no font in the system for displaying text in the chosen codepage).

base/shell/cmd/cmd.c
```diff
@@ -2308,6 +2320,23 @@ Initialize(VOID)
         ConOutPuts(_T("(C) Copyright 1998-") _T(COPYRIGHT_YEAR) _T(" ReactOS Team.\n"));
     }
 
+    /* ReactOS extension: Inform if the Asian font is not found */
```
HBelusca, line 2323: PR #4301 should allow removing this CMD hack.

JoachimHenze (changes requested):
Ok so now that Hermes has proposed something for cmd, we should not longer touch cmd from within this PR anymore.

katahiromz:
Closed.

PR #4096, opened 2021-11-09
[NET] Some improvements

nagayev:
## Purpose

Several net improvements.
JIRA issue: none

## Proposed changes

- Add translation for net command
- Simplify conditions

PR #4282, opened 2022-01-13
Fix net command output for PR #4096

binarymaster:
- [SAMSRV] Return correct status for existing groups and users
- [NETAPI32] NetpNtStatusToApiStatus: Update accordingly

Maybe it even fixes some API tests, but I didn't checked.


HBelusca (approved):

binarymaster:
Test results: https://reactos.org/testman/compare.php?ids=80737,80740

Nothing about netapi or samsrv has changed, probably because tests don't cover user account management.

PR #4283, opened 2022-01-14
[USER32] Adjust the threshold in CascadeWindows

katahiromz:
## Purpose
Not to make the windows too small.
JIRA issue: N/A

## Proposed changes

- Improve the threshold of `CascadeWindows` function not to make the windows too small.

## TODO

- [x] Do tests.

## Screenshots

VirtualBox_ReactOS_14_01_2022_21_04_47
VirtualBox_ReactOS_14_01_2022_21_00_53
Still working.

### DIFF
--- a/win32ss/user/user32/windows/mdi.c
+++ b/win32ss/user/user32/windows/mdi.c
@@ -2149,7 +2149,7 @@ CascadeWindows(HWND hwndParent, UINT wFlags, LPCRECT lpRect,
         if (IsZoomed(hwnd))
         {
             if (!SendMessageTimeoutW(hwnd, WM_SYSCOMMAND, SC_RESTORE, 0,
-                                     SMTO_ABORTIFHUNG, 100, NULL))
+                                     SMTO_ABORTIFHUNG, 300, NULL))
             {
                 continue;
             }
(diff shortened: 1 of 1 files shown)


win32ss/user/user32/windows/mdi.c
```diff
@@ -2147,7 +2147,13 @@ CascadeWindows(HWND hwndParent, UINT wFlags, LPCRECT lpRect,
             continue;
 
         if (IsZoomed(hwnd))
-            ShowWindow(hwnd, SW_RESTORE | SW_SHOWNA);
+        {
```
HBelusca, line 2150: Technically the show window should be enough; if the window to show doesn't respond, it should just be ghosted by win32k. So it seems that here this is a workaround for an actual still unfixed bug.

win32ss/user/user32/windows/mdi.c
```diff
@@ -2330,7 +2338,13 @@ TileWindows(HWND hwndParent, UINT wFlags, LPCRECT lpRect,
         hwnd = info.ahwnd[i];
 
         if (IsZoomed(hwnd))
-            ShowWindow(hwnd, SW_RESTORE | SW_SHOWNA);
+        {
```
HBelusca, line 2341: Same comment here.

win32ss/user/user32/windows/mdi.c
katahiromz: Reverted.

win32ss/user/user32/windows/mdi.c
katahiromz: Reverted.

PR #4285, opened 2022-01-15
[NTOS:KE] Check for vital features being present on i586

Extravert-ir:
## Purpose

Explicitly do a BugCheck if a CPU does not support features we require (cpuid, x87 FPU and some others).
Also, remove unused code paths for FPU emulation.

For any commenters, please take into account:
**This PR does not change current behaviour.** We don't support such CPUs anyway.

FIXME: what to put as parameters to `KeBugCheckEx(UNSUPPORTED_PROCESSOR, ...)`?

### DIFF
--- a/ntoskrnl/ke/i386/cpu.c
+++ b/ntoskrnl/ke/i386/cpu.c
@@ -24,7 +24,7 @@ UCHAR KiNMITSS[KTSS_IO_MAPS];
 ULONG KeI386CpuType;
 ULONG KeI386CpuStep;
 ULONG KiFastSystemCallDisable = 0;
-ULONG KeI386NpxPresent = 0;
+ULONG KeI386NpxPresent = TRUE;
 ULONG KiMXCsrMask = 0;
 ULONG MxcsrFeatureMask = 0;
 ULONG KeI386XMMIPresent = 0;
@@ -90,66 +90,30 @@ VOID
 NTAPI
 KiSetProcessorType(VOID)
 {
-    ULONG EFlags, NewEFlags;
     CPU_INFO CpuInfo;
     ULONG Stepping, Type;
 
-    /* Start by assuming no CPUID data */
-    KeGetCurrentPrcb()->CpuID = 0;
+    /* Do CPUID 1 now */
+    KiCpuId(&CpuInfo, 1);
 
-    /* Save EFlags */
-    EFlags = __readeflags();
+    /*
+     * Get the Stepping and Type. The stepping contains both the
+     * Model and the Step, while the Type contains the returned Type.
+     * We ignore the family.
+     *
+     * For the stepping, we convert this: zzzzzzxy into this: x0y
+     */
+    Stepping = CpuInfo.Eax & 0xF0;
+    Stepping <<= 4;
+    Stepping += (CpuInfo.Eax & 0xFF);
+    Stepping &= 0xF0F;
+    Type = CpuInfo.Eax & 0xF00;
+    Type >>= 8;
 
-    /* XOR out the ID bit and update EFlags */
-    NewEFlags = EFlags ^ EFLAGS_ID;
-    __writeeflags(NewEFlags);
-
-    /* Get them back and see if they were modified */
-    NewEFlags = __readeflags();
-    if (NewEFlags != EFlags)
-    {
-        /* The modification worked, so CPUID exists. Set the ID Bit again. */
-        EFlags |= EFLAGS_ID;
-        __writeeflags(EFlags);
-
-        /* Peform CPUID 0 to see if CPUID 1 is supported */
-        KiCpuId(&CpuInfo, 0);
-        if (CpuInfo.Eax > 0)
-        {
-            /* Do CPUID 1 now */
-            KiCpuId(&CpuInfo, 1);
-
-            /*
-             * Get the Stepping and Type. The stepping contains both the
-             * Model and the Step, while the Type contains the returned Type.
-             * We ignore the family.
-             *
-             * For the stepping, we convert this: zzzzzzxy into this: x0y
-             */
-            Stepping = CpuInfo.Eax & 0xF0;
-            Stepping <<= 4;
-            Stepping += (CpuInfo.Eax & 0xFF);
-            Stepping &= 0xF0F;
-            Type = CpuInfo.Eax & 0xF00;
-            Type >>= 8;
-
-            /* Save them in the PRCB */
-            KeGetCurrentPrcb()->CpuID = TRUE;
-            KeGetCurrentPrcb()->CpuType = (UCHAR)Type;
-            KeGetCurrentPrcb()->CpuStep = (USHORT)Stepping;
-        }
-        else
-        {
-            DPRINT1("CPUID Support lacking\n");
-        }
-    }
-    else
-    {
-        DPRINT1("CPUID Support lacking\n");
-    }
-
-    /* Restore EFLAGS */
-    __writeeflags(EFlags);
+    /* Save them in the PRCB */
+    KeGetCurrentPrcb()->CpuID = TRUE;
+    KeGetCurrentPrcb()->CpuType = (UCHAR)Type;
+    KeGetCurrentPrcb()->CpuStep = (USHORT)Stepping;
 }
 
 CODE_SEG("INIT")
@@ -159,10 +123,6 @@ KiGetCpuVendor(VOID)
 {
     PKPRCB Prcb = KeGetCurrentPrcb();
     CPU_INFO CpuInfo;
-
-    /* Assume no Vendor ID and fail if no CPUID Support. */
-    Prcb->VendorString[0] = 0;
-    if (!Prcb->CpuID) return 0;
 
     /* Get the Vendor ID */
     KiCpuId(&CpuInfo, 0);
@@ -465,7 +425,6 @@ NTAPI
 KiGetCacheInformation(VOID)
 {
     PKIPCR Pcr = (PKIPCR)KeGetPcr();
-    ULONG Vendor;
     CPU_INFO CpuInfo;
     ULONG CacheRequests = 0, i;
     ULONG CurrentRegister;
@@ -476,12 +435,8 @@ KiGetCacheInformation(VOID)
     /* Set default L2 size */
     Pcr->SecondLevelCacheSize = 0;
 
-    /* Get the Vendor ID and make sure we support CPUID */
-    Vendor = KiGetCpuVendor();
-    if (!Vendor) return;
-
     /* Check the Vendor ID */
-    switch (Vendor)
+    switch (KiGetCpuVendor())
     {
         /* Handle Intel case */
         case CPU_INTEL:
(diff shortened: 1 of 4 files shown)


ntoskrnl/ke/i386/kiinit.c
```diff
@@ -392,6 +340,122 @@ KiInitializePcr(IN ULONG ProcessorNumber,
...
+    /* XOR out the ID bit and update EFlags */
+    ULONG NewEFlags = EFlags ^ EFLAGS_ID;
+    __writeeflags(NewEFlags);
+
+    /* Get them back and see if they were modified */
+    NewEFlags = __readeflags();
+
+    if (NewEFlags == EFlags)
+    {
+        /* The modification did not work, so CPUID is not supported. Try Windows 95. */
```
HBelusca, line 362: "Try Windows 95?" what does that mean?

ntoskrnl/ke/i386/kiinit.c
```diff
@@ -392,6 +340,122 @@ KiInitializePcr(IN ULONG ProcessorNumber,
...
+        KeBugCheckEx(UNSUPPORTED_PROCESSOR, 0, 0, 0, 0); // TODO: what to put here?
```
HBelusca, line 363: What to put: is the VendorString initialized at this stage? if so, something similar to what's done in the other bugchecks below?

ntoskrnl/ke/i386/cpu.c
```diff
@@ -1208,9 +1122,6 @@ KiIsNpxErrataPresent(VOID)
     INT ErrataPresent;
     ULONG Cr0;
 
-    /* Disable interrupts */
-    _disable();
```
HBelusca, line 1212: Why not necessary?

ntoskrnl/ke/i386/kiinit.c
```diff
@@ -85,58 +85,6 @@ KiInitMachineDependent(VOID)
     /* Check for PAT support and enable it */
     if (KeFeatureBits & KF_PAT) KiInitializePAT();
 
-    /* Assume no errata for now */
-    SharedUserData->ProcessorFeatures[PF_FLOATING_POINT_PRECISION_ERRATA] = 0;
-
-    /* Check if we have an NPX */
-    if (KeI386NpxPresent)
```
HBelusca, line 92: Since you now assume always we support CPUs with an **NPX** :P (aka math coprocessor aka FPU), that FPU errata seems to be important to be controlled.

ntoskrnl/ke/i386/kiinit.c
```diff
@@ -392,6 +340,122 @@ KiInitializePcr(IN ULONG ProcessorNumber,
...
+        EFlags |= EFLAGS_ID;
+        __writeeflags(EFlags);
+    }
+
+    /* Peform CPUID 0 to see if CPUID 1 is supported */
+    KiCpuId(&CpuInfo, 0);
+    if (CpuInfo.Eax == 0)
+    {
+        /* Maybe Windows 95 after all? */
+        KeBugCheckEx(UNSUPPORTED_PROCESSOR, 0, 0, 0, 0); // TODO: what to put here?
```
HBelusca, line 377: Same comment.

ntoskrnl/ke/i386/kiinit.c
```diff
@@ -392,6 +340,122 @@ KiInitializePcr(IN ULONG ProcessorNumber,
...
+    }
+
+    // 2. Detect and set the CPU Type
+    KiSetProcessorType();
+
+    // We recommend Windows 3.11 for i386 CPU.
```
HBelusca, line 383: ok that's a nice recurring joke for the PR but should be removed in final version.

ntoskrnl/ke/i386/kiinit.c
```diff
@@ -392,6 +340,122 @@ KiInitializePcr(IN ULONG ProcessorNumber,
...
+
+    // 4. Verify it supports everything we need.
+    if (!(FeatureBits & KF_RDTSC))
+    {
+        ULONG Vendor[3];
+
+        RtlCopyMemory(Vendor, Prcb->VendorString, sizeof(Vendor));
+
+        KeBugCheckEx(UNSUPPORTED_PROCESSOR,
+                     (1 << 24 ) | (Prcb->CpuType << 16) | Prcb->CpuStep,
```
HBelusca, line 398:
```suggestion
                     (1 << 24) | (Prcb->CpuType << 16) | Prcb->CpuStep,
```
(here and below in the 3 other copypasted lines).

ntoskrnl/ke/i386/kiinit.c
```diff
@@ -392,6 +340,122 @@ KiInitializePcr(IN ULONG ProcessorNumber,
...
+                     Vendor[0],
+                     Vendor[1],
+                     Vendor[2]);
+    }
+
+    // Check x87 FPU is present. FIXME: put this into FeatureBits?
+    KiCpuId(&CpuInfo, 1);
+
+    if (!(CpuInfo.Edx & 0x00000001))
```
HBelusca, line 420: What does this bit correspond to?

ntoskrnl/ke/i386/kiinit.c
HBelusca: Do you think there is one of those four bugcheck fields that could be sacrified and used instead as a hint to which feature bit support is missing?

ntoskrnl/ke/i386/kiinit.c
```diff
@@ -432,31 +485,28 @@ KiInitializeKernel(IN PKPROCESS InitProcess,
     {
         /* Set it always on */
         SharedUserData->NXSupportPolicy = NX_SUPPORT_POLICY_ALWAYSON;
-        FeatureBits |= KF_NX_ENABLED;
+        KeFeatureBits |= KF_NX_ENABLED;
```
HBelusca, line 488: Now you are modifying the global `KeFeatureBits` with state from every processor, instead of it being the bits of the boot CPU only. Not sure this is what you want, this changes what this global variable was designed for.

HBelusca:
You may be interested by https://www.geoffchappell.com/studies/windows/km/bugchecks/3e.htm

JoachimHenze:
Out of curiosity, can you point towards some code **most close** to the bare core of ros where we do rely on x87 FPU?

ntoskrnl/ke/i386/kiinit.c
```diff
@@ -392,6 +340,122 @@ KiInitializePcr(IN ULONG ProcessorNumber,
...
+    KiCpuId(&CpuInfo, 1);
+
+    if (!(CpuInfo.Edx & 0x00000001))
+    {
+        ULONG Vendor[3];
+
+        RtlCopyMemory(Vendor, Prcb->VendorString, sizeof(Vendor));
+
+        KeBugCheckEx(UNSUPPORTED_PROCESSOR,
+                     (1 << 24 ) | (Prcb->CpuType << 16) | Prcb->CpuStep,
```
Saibamen, line 427:
```suggestion
                     (1 << 24) | (Prcb->CpuType << 16) | Prcb->CpuStep,
```

ntoskrnl/ke/i386/kiinit.c
```diff
@@ -392,6 +340,122 @@ KiInitializePcr(IN ULONG ProcessorNumber,
...
+    }
+
+    if (!(FeatureBits & KF_CMPXCHG8B))
+    {
+        ULONG Vendor[3];
+
+        RtlCopyMemory(Vendor, Prcb->VendorString, sizeof(Vendor));
+
+        KeBugCheckEx(UNSUPPORTED_PROCESSOR,
+                     (1 << 24 ) | (Prcb->CpuType << 16) | Prcb->CpuStep,
```
Saibamen, line 411:
```suggestion
                     (1 << 24) | (Prcb->CpuType << 16) | Prcb->CpuStep,
```

ntoskrnl/ke/i386/kiinit.c
```diff
@@ -392,6 +340,122 @@ KiInitializePcr(IN ULONG ProcessorNumber,
...
+    __writecr0(Cr0);
+
+    if (KiIsNpxErrataPresent())
+    {
+        ULONG Vendor[3];
+
+        RtlCopyMemory(Vendor, Prcb->VendorString, sizeof(Vendor));
+
+        KeBugCheckEx(UNSUPPORTED_PROCESSOR,
+                     (1 << 24 ) | (Prcb->CpuType << 16) | Prcb->CpuStep,
```
Saibamen, line 449:
```suggestion
                     (1 << 24) | (Prcb->CpuType << 16) | Prcb->CpuStep,
```

HBelusca:
> Out of curiosity, can you point towards some code **most close** to the bare core of ros where we do rely on x87 FPU? And KF_CMPXCHG8B.

@JoachimHenze For CMPXCHG8B the answer is very simple: All the InterlockedExchange*** and InterlockedCompare*** intrinsics used to do lockless operations (just grep in the code and see for yourself).

Extravert-ir:
> where we do rely on x87 FPU?

We don't rely on it in the kernel, we have to set it up all surrounding stuff anyway

ntoskrnl/ke/i386/kiinit.c
Extravert-ir: Yeah, I'm aware of that. I left it like that, because I'm going to do more refactorings of Ke initialization code in preparation for SMP. Want to unify Ke stuff across architectures and make it less messy

ntoskrnl/ke/i386/kiinit.c
Extravert-ir: That's an x87 FPU bit

ntoskrnl/ke/i386/kiinit.c
Extravert-ir: I'd say vendor field is not that much important at all :) The best way is to indicate which feature we didn't like here, yes

ntoskrnl/ke/i386/kiinit.c
JoachimHenze: I think each commit/PR should strive to give a sane state afterwards already, if possible.

JoachimHenze:
> > where we do rely on x87 FPU?
> 
> We don't rely on it in the kernel, we have to set it up all surrounding stuff anyway
Don't worry, I won't oppose your PR because of that. I really would just welcome some fingerpoint to such code in question.

ntoskrnl/ke/i386/cpu.c
Extravert-ir: This function is now called from `KiVerifyCpuFeatures`, which has interrupts disabled already

ThFabba: I would typically indicate that with an `ASSERT(!(__readeflags() & EFLAGS_INTERRUPT_MASK))`

ntoskrnl/ke/i386/kiinit.c
Extravert-ir: That is what `KiIsNpxErrataPresent` function is used for

HBelusca: Ah ok, so my understanding correct, that in a sense, that loop done below (and deleted) is in a way "done", now when each CPU is being initialized separately and KiVerifyCpuFeatures() is called?

Extravert-ir: Yes. `KiVerifyCpuFeatures` is called on each CPU, so we verify all of them don't have that FPU bug

ntoskrnl/ke/i386/kiinit.c
```diff
@@ -392,6 +340,92 @@ KiInitializePcr(IN ULONG ProcessorNumber,
...
+
+    // 3. Finally, obtain CPU features.
+    ULONG FeatureBits = KiGetFeatureBits();
+
+    // 4. Verify it supports everything we need.
+    if (!(FeatureBits & KF_RDTSC))
+    {
+        // 0x2 - missing CPUID features
+        // second paramenter - edx flag which is missing
+        KeBugCheckEx(UNSUPPORTED_PROCESSOR, 0x2, 0x00000010, 0, 0);
```
HBelusca, line 394: Use the human-readable flag names (here and in the other places) instead of their hardcoded value.

Extravert-ir: Oh please, we don't use them anywhere in the kernel

HBelusca: I was thinking about using (for example) "KF_RDTSC" and co. ... unless there is not a one-to-one mapping between the actual CPU flag bit, and the "feature bits" stored in the eponymous name.

Extravert-ir: Check `KeGetFeatureBits` https://github.com/reactos/reactos/blob/463d312169f6763fa51663e87d71e5abbd297331/ntoskrnl/ke/i386/cpu.c#L214 - we don't use any defines when accessing CPUID flags. That may be refactored, but not in this PR

HBelusca: ok. Then I wonder, whether it's better to show up the actual CPU flag bit in the BSOD, or the NT `KF_xxx` flag instead 🤔

Extravert-ir: I've choose CPUID flag because we don't have a feature bit for x87

PR #4294, opened 2022-01-16
[FREELDR] LoadModule(): Add error handling

SergeGautherie:
## Purpose

Better fail gracefully, than later load default file(s) silently...

Another testcase:
Select Screen or File entries!

JIRA issue: [CORE-17645](https://jira.reactos.org/browse/CORE-17645)

## Proposed changes

- LoadWindowsCore(): Fix KD dll name handling
- LoadModule(): Check/log errors.
- LoadWindowsCore(): Check/log/return LoadModule() failures.

### DIFF
--- a/boot/freeldr/freeldr/ntldr/winldr.c
+++ b/boot/freeldr/freeldr/ntldr/winldr.c
@@ -583,14 +583,17 @@ LoadWindowsCore(IN USHORT OperatingSystemVersion,
              * Parse the port name.
              * Format: /DEBUGPORT=COM[1-9]
              * or: /DEBUGPORT=FILE:\Device\HarddiskX\PartitionY\debug.log
+             * or: /DEBUGPORT=SCREEN
              * or: /DEBUGPORT=FOO
              * If we only have /DEBUGPORT= (i.e. without any port name),
              * defaults it to "COM".
              */
             RtlStringCbCopyA(KdTransportDllName, sizeof(KdTransportDllName), "KD");
-            if (_strnicmp(Option, "COM", 3) == 0 && '0' <= Option[3] && Option[3] <= '9')
+            if ((OptionLength == 4 && _strnicmp(Option, "COM", 3) == 0 && '1' <= Option[3] && Option[3] <= '9') ||
+                (OptionLength > 5 && _strnicmp(Option, "FILE", 4) == 0 && Option[4] == ':') ||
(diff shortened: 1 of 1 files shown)


SergeGautherie:
> Select kdscreen.dll entry on a msvc build!

Actually, gcc build is affected too.
Only kdcom and kdvbox exist: kdscreen and kdfile do not, so kdcom is loaded later/instead.
Then, why is there code to try to load such nonexistent filenames?

boot/freeldr/freeldr/ntldr/winldr.c
```diff
@@ -583,14 +595,17 @@ LoadWindowsCore(IN USHORT OperatingSystemVersion,
              * Parse the port name.
              * Format: /DEBUGPORT=COM[1-9]
              * or: /DEBUGPORT=FILE:\Device\HarddiskX\PartitionY\debug.log
+             * or: /DEBUGPORT=SCREEN
              * or: /DEBUGPORT=FOO
```
HBelusca, line 599: These are **just** examples, not the only actual list of options supported. (and "FOO" is definitively placeholder for any name).

boot/freeldr/freeldr/ntldr/winldr.c
```diff
@@ -583,14 +595,17 @@ LoadWindowsCore(IN USHORT OperatingSystemVersion,
...
              * If we only have /DEBUGPORT= (i.e. without any port name),
              * defaults it to "COM".
              */
             RtlStringCbCopyA(KdTransportDllName, sizeof(KdTransportDllName), "KD");
-            if (_strnicmp(Option, "COM", 3) == 0 && '0' <= Option[3] && Option[3] <= '9')
+            if ((OptionLength == 4 && _strnicmp(Option, "COM", 3) == 0 && '1' <= Option[3] && Option[3] <= '9') ||
+                (OptionLength > 5 && _strnicmp(Option, "FILE", 4) == 0 && Option[4] == ':') ||
+                (OptionLength == 6 && _strnicmp(Option, "SCREEN", 6) == 0))
```
HBelusca, line 606: Now, if we happen (or some user happen) to introduce a genuine "kdscreen.dll" (for example) debug transport dll file, you are forbidding it from loading?

Please don't change code you don't fully understand why it is there.

HBelusca (changes requested):

HBelusca:
Parts of this PR have been merged into master, see commit 15e1489c7f.

PR #4301, opened 2022-01-21
Sync thread's language/locale with console's preferred language

HBelusca:
## Purpose

For a given non-unicode display code page, the console may not support showing the characters of the corresponding character set, if no suitable monospace font is available.
(For example, this is especially true in CJK ReactOS installations where no default _monospace_ font with CJK characters is installed; one needs an external fonts pack. But this observation is not limited to ReactOS.)

In this case, the console may change the code page to a default one. Or it may not, and might instead keep the code page, even if the selected font does not fully support displaying the corresponding characters.
On the other hand, a suitable font may exist and the console would keep the output code page.

In any of these cases, it is advisable that the console also tells which language it advises to be used for output.
On Windows, this is done whenever a program creates/attaches to a console, and when it sets a new output code page, and is mostly used in CJK installations. The calling thread's locale (in Windows <= 2003, since there was no concept of a separate thread UI language) is then modified with the corresponding language.

JIRA issue: [CORE-17601](https://jira.reactos.org/browse/CORE-17601), [CORE-17803](https://jira.reactos.org/browse/CORE-17803)
Replaces PR #4281

### NOTE
This PR should be tested together with PR #4337

## Proposed changes

```
[KERNEL32][CONSRV] Implement retrieving the best-suited language ID corresponding to the active console output code page.

Implement SrvGetConsoleLangId() (server-side) to retrieve the console's preferred language ID,
and (client-side) set the new current thread's locale after connecting to a console, or changing its output code page.

Based on comments gathered, and code, from: https://github.com/microsoft/terminal
```
```
[KERNEL32] SetThreadUILanguage(): Check for parameter (we and Win2k3 only support '0'). Use LANGIDFROMLCID() macro.
```

## TODO

- [x] Write tests.
- [x] Show the matrix of test results.
- [x] What about Windows 10 CJK? (ask @katahiromz)

## Test results matrix

**How to test:** Add a pause (like a `Sleep` or an infinite loop `while (1);`) just after the `test_CP_ThreadLang()` test in `modules/rostests/apitests/kernel32/ConsoleCP.c`, so that you can have time to see the test results (that appear in the separate spanned console window, that would otherwise just vanish quickly).

### Non-CJK systems

_**OBSERVATIONS:**_
- *(Cases A,E):* Both unfixed ReactOS and Windows 10 switch to a console CJK codepage (second test line), irrespectively of whether the system is CJK or not (and has or hasn't suitable fonts).
- *(Cases B,D):* Both ReactOS with this PR + PR #4337 and Windows 2003 will refuse to switch the console to a (otherwise valid) CJK codepage. _**BUT:**_ Windows 2003 will always do that, while ReactOS will do it only if there is no suitable CJK font present *(Case B, compare with Case C)*.
- *(Cases C,E)* Fixed ReactOS will switch to a console CJK codepage (even when the system is non-CJK) when it has the CJK fonts, and thus will behave similarly to Windows 10.

| Test OS      | Results |
| :-----:      | :-----: |
| _**A:**_ ReactOS _(No fixes or just this PR)_ | reactos_nofixes_or_withfixes_USACP |
| _**B:**_ ReactOS _(This PR + PR #4337, No CJK fonts)_ | reactos_withfixes_and_PR4337_NoCJKFonts_USACP |
| _**C:**_ ReactOS _(This PR + PR #4337, With CJK fonts)_ | reactos_withfixes_and_PR4337_USACP |
| _**D:**_ Windows 2003 | win2k3_USACP |
| _**E:**_ Windows 10   | win10_USACP  |

### CJK systems

_**OBSERVATIONS:**_
- *(Case C):* ReactOS with this PR + PR #4337 _*but without CJK fonts*_ will refuse to change to a console CJK codepage and therefore will behave similarly to the case of a non-CJK system.
- *(Cases B/D/E/F vs. A):* This PR allows ReactOS to sync a console program's thread lang ID to the one suitable for the console, as on Windows 2003+.
- *(Cases D,E,F):* Both ReactOS with this PR + PR #4337 with CJK fonts, and Windows 2003/Windows 10 behave correctly.

| Test OS      | Results |
| :-----:      | :-----: |
| _**A:**_ ReactOS _(No fixes, with or without CJK fonts)_ | reactos_nofixes_CJK |
| _**B:**_ ReactOS _(This PR only, with or without CJK fonts)_ | reactos_withfixes_CJK |
| _**C:**_ ReactOS _(This PR + PR #4337, No CJK fonts)_ | reactos_withfixes_and_PR4337_CJK_1 |
| _**D:**_ ReactOS _(This PR + PR #4337, With CJK fonts)_ | reactos_withfixes_and_PR4337_CJK_2 |
| _**E:**_ Windows 2003 | win2k3_CJK_system |
| _**F:**_ Windows 10 | <pre><br>ConsoleCP.c:1531: Running on CJK system (codepage 932)<br>ConsoleCP.c:152: Thread LangId 1041, expecting 1041...<br>ConsoleCP.c:157: Thread LangId 1033, expecting 1033...<br>ConsoleCP.c:161: Thread LangId 1041, expecting 1041...</pre> |

### DIFF
--- a/dll/win32/kernel32/client/console/console.c
+++ b/dll/win32/kernel32/client/console/console.c
@@ -1375,8 +1375,6 @@ AllocConsole(VOID)
     ULONG AppNameLength = 128 * sizeof(WCHAR);
     ULONG CurDirLength  = (MAX_PATH + 1) * sizeof(WCHAR);
 
-    LCID lcid;
-
     RtlEnterCriticalSection(&ConsoleLock);
 
     if (NtCurrentPeb()->ProcessParameters->ConsoleHandle)
@@ -1427,8 +1425,8 @@ AllocConsole(VOID)
         /* Initialize Console Ctrl Handling */
         InitializeCtrlHandling();
 
-        /* Sets the current console locale for this thread */
-        SetTEBLangID(lcid);
+        /* Sync the current thread's LangId with the console's one */
+        SetTEBLangID();
     }
 
 Quit:
@@ -2500,6 +2498,9 @@ SetConsoleOutputCP(UINT wCodePageID)
         return FALSE;
     }
 
+    /* Sync the current thread's LangId with the console's one */
+    SetTEBLangID();
+
     return TRUE;
 }
 
@@ -2676,9 +2677,7 @@ AttachConsole(DWORD dwProcessId)
 {
     BOOL Success;
     CONSOLE_START_INFO ConsoleStartInfo;
-
     DWORD dummy;
-    LCID lcid;
 
     RtlEnterCriticalSection(&ConsoleLock);
 
@@ -2711,8 +2710,8 @@ AttachConsole(DWORD dwProcessId)
         /* Initialize Console Ctrl Handling */
         InitializeCtrlHandling();
 
-        /* Sets the current console locale for this thread */
-        SetTEBLangID(lcid);
+        /* Sync the current thread's LangId with the console's one */
+        SetTEBLangID();
     }
 
 Quit:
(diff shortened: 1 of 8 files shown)


dll/win32/kernel32/client/console/console.c
```diff
@@ -3067,6 +3066,48 @@ UnregisterConsoleIME(VOID)
...
+     * modifying its locale -- Windows <= 2003 does not have
+     * the concept of a separate thread UI language.
+     * Ignore the returned value.
+     */
+    if (!SetThreadLocale(MAKELCID(LangIdRequest->LangId, SORT_DEFAULT)))
+    {
+        DPRINT1("SetTEBLangID: Could not set thread locale to console lang ID %lu\n",
+                LangIdRequest->LangId);
+    }
+}
```
katahiromz, line 3110: 1. How to detect whether the target font is absent or not.
2. If absent, what should we have to do?

win32ss/user/winsrv/consrv/console.c
```diff
@@ -2014,10 +2014,51 @@ CSR_API(SrvSetConsoleNlsMode)
...
+    UINT CurrentACP = GetACP();
+    if (!IsCJKCodePage(CurrentACP))
+        return STATUS_NOT_SUPPORTED;
+#endif
+
+    /* Return a "best-suited" language ID corresponding
+     * to the active console output code page. */
+    switch (Console->OutputCodePage)
+    {
+    case CP_JAPANESE:
```
katahiromz, line 2042:
```txt
2022-01-21T20:08:44.9101026Z ../src/win32ss/user/winsrv/consrv/console.c:2042:10: error: 'CP_JAPANESE' undeclared (first use in this function); did you mean 'LANG_JAPANESE'?
2022-01-21T20:08:44.9101405Z      case CP_JAPANESE:
```

dll/win32/kernel32/client/console/console.c
HBelusca: For the font, this will be in a separate PR.

win32ss/user/winsrv/consrv/console.c
HBelusca: fixed

katahiromz (approved):

katahiromz:
I have tested with PR 4301 and 4337 applied.
Fallbacking font and switching font looks good.
VirtualBox_ReactOS_06_02_2022_11_23_26
VirtualBox_ReactOS_06_02_2022_11_24_36

katahiromz (approved):

PR #4303, opened 2022-01-21
[ROSTESTS] Fix crash in winhttp:winhttp regression test

Doug-Lyons:
## Improve wine_dbgstr_wn function by partial sync to Wine-6.14

_The winhttp:winhttp test was often crashing on wine_dbgstr_wn._

JIRA issue: [ROSTESTS-377](https://jira.reactos.org/browse/ROSTESTS-377)

## Sync our test.h file's wine_dbgstr_wn as much as possible to Wine's

_This brings most of Wine's current wine_dbgstr_wn code into ReactOS._
It is not possible to completely sync this to current Wine.
This is because there is Wine-specific code regarding "debug_info" and their TEB in remaining code.
I confirmed this with @ThFabba.

### DIFF
--- a/sdk/include/reactos/wine/test.h
+++ b/sdk/include/reactos/wine/test.h
@@ -634,7 +634,7 @@ const char *wine_dbgstr_an( const CHAR *str, intptr_t n )
     return res;
 }
 
-const char *wine_dbgstr_wn( const WCHAR *str, int n )
+const char *wine_dbgstr_wn( const WCHAR *str, intptr_t n )
 {
     char *res;
     static const char hex[16] = {'0','1','2','3','4','5','6','7','8','9','a','b','c','d','e','f'};
--- a/sdk/include/reactos/wine/test.h
+++ b/sdk/include/reactos/wine/test.h
@@ -634,30 +634,24 @@ const char *wine_dbgstr_an( const CHAR *str, intptr_t n )
     return res;
 }
 
-const char *wine_dbgstr_wn( const WCHAR *str, intptr_t n )
+const char *wine_dbgstr_wn( const WCHAR *str, int n )
 {
-    char *dst, *res;
-    size_t size;
+    char *res;
+    static const char hex[16] = {'0','1','2','3','4','5','6','7','8','9','a','b','c','d','e','f'};
+    char buffer[300], *dst = buffer;
 
+    if (!str) return "(null)";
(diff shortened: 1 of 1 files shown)


sdk/include/reactos/wine/test.h
```diff
@@ -634,30 +634,24 @@ const char *wine_dbgstr_an( const CHAR *str, intptr_t n )
     return res;
 }
 
-const char *wine_dbgstr_wn( const WCHAR *str, intptr_t n )
+const char *wine_dbgstr_wn( const WCHAR *str, int n )
```
HBelusca, line 637: No idea what Wine drank, but this should certainly be `intptr_t` or `size_t` for x64 compliance.

Doug-Lyons: Changed. Thanks.

sdk/include/reactos/wine/test.h
```diff
@@ -668,14 +662,15 @@ const char *wine_dbgstr_wn( const WCHAR *str, intptr_t n )
...
-            else
+            if (c < ' ' || c >= 127)
             {
                 *dst++ = '\\';
-                sprintf(dst,"%04x",c);
-                dst+=4;
+                *dst++ = hex[(c >> 12) & 0x0f];
+                *dst++ = hex[(c >> 8) & 0x0f];
+                *dst++ = hex[(c >> 4) & 0x0f];
+                *dst++ = hex[c & 0x0f];
```
learn-more, line 671: This code is written for a 4-bytes WCHAR, which we don't have.
So the upper 2 digits will always be 0.

Doug-Lyons: Mark, Can you please review https://reactos.org/testman/diff.php?id1=56015182&id2=56024828&type=1&strip=0 and look at line #4279 on the left and the right and see if it makes sense to remove any of these lines? Thanks. Doug Lyons

learn-more: Sorry, I misread the code.
It is good as-is.

learn-more (approved):

HBelusca (approved):

Doug-Lyons:
Thanks for your help in getting this merged, @binarymaster. We will have to watch the winhttp:winhttp test for a while and see what it shows. I do have another patch in [ROSTESTS-377](https://jira.reactos.org/browse/ROSTESTS-377) that may be needed.

PR #4304, opened 2022-01-22
[PORTCLS] Centralize AddRef/Release implementation and make it thread-safe.

ThFabba:
Just cleaning up all those broken interlocked operations...


drivers/wdm/audio/backpln/portcls/miniport_dmus.cpp
```diff
@@ -67,12 +67,11 @@ const ULONG      kMPUInputBufferSize = 128;
  * reference counting and aggregation support.
  */
 class CMiniportDMusUART
-:   public IMiniportDMus,
-    public IMusicTechnology,
-    public IPowerNotify
+:   public CUnknownImpl<IMiniportDMus,
+                        IMusicTechnology,
+                        IPowerNotify>
```
HBelusca, line 72: So that I understand the need of the (C++11 😡) variadic template, is it because if there is inheritance from other classes implementing these interfaces, they will share the same reference counter?

ThFabba: Yep. If you inherited from three different `CUnknownImpl`s, you'd get three refcounts. If you only make the first interface a `CUnknownImpl`, the other interfaces would be lacking AddRef/Release implementations.
Happy to solve this another way if there is one -- but the only one I found was to put the implementation behind a macro. Or I guess you could have `CUnknownImpl1`, `CUnknownImpl2`, `CUnknownImpl3`, but that's pretty ridiculous, too.

HBelusca: Would using virtual inheritance be another way?
`class CFooBar : virtual public IInterface1, virtual public IInterface2`

ColinFinck (approved):
Thanks for doing that cleanup!
Nice to see that we can finally solve entire classes of bugs once and for all, instead of watching and fixing AddRef/Release implementations over and over again..

drivers/wdm/audio/backpln/portcls/miniport_dmus.cpp
ThFabba: As far as I can tell, that doesn't help -- https://godbolt.org/z/soMfn396c
But if you can get it to work in that snippet somehow, we can probably get it to work in the real world, too.

HBelusca: I got it working by adding some missing "virtual" to the inheritances of the intermediate classes: see https://godbolt.org/z/dorxxcz1n

JoachimHenze: I would love if we could keep the compatibility to the other compilers here!

learn-more (approved):

JoachimHenze (changes requested):
What about hbeluscas proposal?

Extravert-ir:
Guys, let's just merge it. I think there are better things to do than trying to replace a compiler feature with a pile of hacks (for completely no reason)

drivers/wdm/audio/backpln/portcls/port_wavecyclic.cpp
```diff
@@ -16,28 +16,13 @@
...
-    {
-        InterlockedIncrement(&m_Ref);
-        return m_Ref;
-    }
-    STDMETHODIMP_(ULONG) Release()
-    {
-        InterlockedDecrement(&m_Ref);
-        if (!m_Ref)
-        {
-            //delete this;
```
ThFabba, line 36: Uh... well that's a problem.

drivers/wdm/audio/backpln/portcls/private.hpp
```diff
@@ -439,4 +439,60 @@ typedef struct
...
+};
+
+template<typename Interface1, typename Interface2>
+class CUnknownImpl2 : public Interface1, public Interface2
+{
+    IMPLEMENT_CUNKNOWNIMPL(CUnknownImpl2)
+};
+
+template<typename Interface1, typename Interface2, typename Interface3>
+class CUnknownImpl3 : public Interface1, public Interface2, public Interface3
```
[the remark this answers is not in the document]
learn-more: Think of it like politics, a minority screams the loudest, and they usually get what they want because of that.

ColinFinck (changes requested):
Now we have two competing implementations, 3b8301e9358ef846ca34c92a31e33678d634da89 and 2a8ab7b5d8c0ef3d4b94b77471cea11909071642:
* 3b8301e9358ef846ca34c92a31e33678d634da89 provides a template that can derive from an arbitrary number of interfaces.
* 2a8ab7b5d8c0ef3d4b94b77471cea11909071642 replaces that template by a macro and implements a template for deriving from 1 interface, another one for deriving from 2 interfaces, ... That makes up 4 implementations in total.

I am very much opposed to the current 2a8ab7b5d8c0ef3d4b94b77471cea11909071642:
* Long multi-line macros should be avoided where possible. They are harder to write and harder to debug as there is no source-level debugging for them.
* We introduce another level of complexity by first requiring a macro and then one template per number of interfaces to derive from.
* There is a better way to do the same as proven by 3b8301e9358ef846ca34c92a31e33678d634da89.

Templates with variadic type parameters have been supported for well over a decade. All compilers we target in ReactOS (as [confirmed in several votings in 2020/2021](https://reactos.org/archives/public/ros-dev/2021-January/019173.html)) support C++11 and thereby also that feature.

We would set a really bad precedent by replacing good code with inferior code. Even more so when the good code already got a majority of proponents.
Keep in mind that everything we do here is public and serves as an example to potential future contributors.

PR #4246, opened 2022-01-04
[AC97] Add Open Source AC'97 Driver

Obaid51:
This PR imports the AC'97 Driver that was originally modified by Michael Stamper. His PR can be found at https://github.com/reactos/reactos/pull/3141
I am not sure about zero3k which you are referring to, but I was hired by some guy on a freelancing platform for some coding. Also as this is working without installing the audio driver explicitly, I guess it's good for the pull request.
Proof of working can be found at the following video link
https://youtu.be/Pv3q_ZLAYt8

PR #4305, opened 2022-01-22
[PORTCLS] Fix crashes because of zero initialization weirdness

Vayne2500:
## Purpose

This fixes BSODs on startup and when running programs that play audio when using the AC’97 Audio Driver that is available at https://github.com/reactos/reactos/pull/4246. This fix was done by Michael Stamper (aka TheDeadFish).

### DIFF
--- a/drivers/wdm/audio/backpln/portcls/dma_slave.cpp
+++ b/drivers/wdm/audio/backpln/portcls/dma_slave.cpp
@@ -17,19 +17,6 @@
 class CDmaChannelInit : public IDmaChannelInit
 {
 public:
-    inline
-    PVOID
-    operator new(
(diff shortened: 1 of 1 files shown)


Vayne2500:
Obaid won't do it (or he might just approve this one).

gonzoMD:
[PORTCLS] Close PR because of Zero3k weirdness.

Vayne2500:
Why? Obaid won't make it and its needed to keep the driver from crashing the OS.

Vayne2500:
Weird, I cannot get it to crash after running it for 30+ minutes with just the driver being used and it playing audio. Maybe another change was done that fixes it or it happens randomly.

PR #4306, opened 2022-01-23
[EXPLORER] Enable _DEBUG and fix assertion failures

katahiromz:
## Purpose

JIRA issue: [CORE-18013](https://jira.reactos.org/browse/CORE-18013)

## Proposed changes

- Enable `_DEBUG` if debugging.
- Fix assertion failures.

### DIFF
--- a/base/shell/explorer/syspager.cpp
+++ b/base/shell/explorer/syspager.cpp
@@ -1231,8 +1231,10 @@ void CNotifyToolbar::Initialize(HWND hWndParent, CBalloonQueue * queue)
         TBSTYLE_FLAT | TBSTYLE_TOOLTIPS | TBSTYLE_WRAPABLE | TBSTYLE_TRANSPARENT |
         CCS_TOP | CCS_NORESIZE | CCS_NOPARENTALIGN | CCS_NODIVIDER;
 
-    // HACK & FIXME: CORE-17505
-    SubclassWindow(CToolbar::Create(hWndParent, styles));
+    // HACK & FIXME
+    HWND hwnd = CToolbar::Create(hWndParent, styles);
+    m_hWnd = NULL;
+    SubclassWindow(hwnd);
 
     // Force the toolbar tooltips window to always show tooltips even if not foreground
     HWND tooltipsWnd = (HWND)SendMessageW(TB_GETTOOLTIPS);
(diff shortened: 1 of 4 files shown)


base/shell/explorer/traywnd.cpp
```diff
@@ -200,8 +200,10 @@ class CStartButton
 
     VOID Initialize()
     {
-        // HACK & FIXME: CORE-17505
-        SubclassWindow(m_hWnd);
+        // HACK & FIXME
+        HWND hWnd = m_hWnd;
+        m_hWnd = NULL;
+        SubclassWindow(hWnd);
```
HBelusca, line 206: Kind of unnatural (here and in the other places). What does the SubclassWindow function expect from `m_hWnd` ?

katahiromz: `SubclassWindow` does assert `m_hWnd == NULL`.

base/shell/explorer/traywnd.cpp
```diff
@@ -231,6 +231,7 @@ class CStartButton
 
         DWORD dwStyle = WS_CHILD | WS_VISIBLE | WS_CLIPSIBLINGS | BS_PUSHBUTTON | BS_LEFT | BS_VCENTER;
 
+        // Hack
```
HBelusca, line 234: It would seem we "loose" the reason why this is a hack. Is it still because the actual fix is to superclass? (as suggested at the top?) Or another reason?

base/shell/explorer/syspager.cpp
```diff
@@ -1231,8 +1231,10 @@ void CNotifyToolbar::Initialize(HWND hWndParent, CBalloonQueue * queue)
         TBSTYLE_FLAT | TBSTYLE_TOOLTIPS | TBSTYLE_WRAPABLE | TBSTYLE_TRANSPARENT |
         CCS_TOP | CCS_NORESIZE | CCS_NOPARENTALIGN | CCS_NODIVIDER;
 
-    // HACK & FIXME: CORE-17505
-    SubclassWindow(CToolbar::Create(hWndParent, styles));
+    // HACK & FIXME
+    HWND hwnd = CToolbar::Create(hWndParent, styles);
+    m_hWnd = NULL;
```
HBelusca, line 1236: Is it the `CToolbar::Create` that sets the `m_hWnd` member?

katahiromz: `CToolbar::Create` does create the window but it doesn't subclass itself. Maybe its class hierarchy has a bug that the `SubclassWindow` method can't be called from `CToolbar`.

base/shell/explorer/traywnd.cpp
katahiromz: Superclassing is not relevant.

PR #4308, opened 2022-01-24
[NTUSER] Improve SPI_(GET/SET)NONCLIENTMETRICS

katahiromz:
## Purpose

`SystemParametersInfo` function didn't check in the cases of `SPI_GETNONCLIENTMETRICS` and `SPI_SETNONCLIENTMETRICS` correctly.

JIRA issue: [CORE-18022](https://jira.reactos.org/browse/CORE-18022)

## Proposed changes

- Strengthen `user32_apitest` `SystemParametersInfo` testcase.
- Return `1` if `SPI_GETNONCLIENTMETRICS` is successful.
- Consider Vista-sized `NONCLIENTMETRICS`.

## TODO

- [x] Do tests.

## Screenshots
WinXP (with disabling theme):
WinXP
Successful.

Win2k3:
Win2k3
Successful.

REACTOS BEFORE:
before
17 failures.

REACTOS AFTER:
after
17 failures. No change.

### DIFF
--- a/modules/rostests/apitests/user32/SystemParametersInfo.c
+++ b/modules/rostests/apitests/user32/SystemParametersInfo.c
@@ -106,10 +106,129 @@ MSG_ENTRY CaptionHeight_chain[]={
 static void Test_NonClientMetrics()
 {
     NONCLIENTMETRICS NonClientMetrics;
+    NONCLIENTMETRICSA ncmA;
+    NONCLIENTMETRICSW ncmW;
+    BOOL ret;
 
     /* WARNING: this test requires themes and dwm to be disabled */
 
     SetCursorPos(0,0);
+
+    ncmA.cbSize = 0;
+    ret = SystemParametersInfo(SPI_GETNONCLIENTMETRICS, 0, &ncmA, 0);
+    ok_int(ret, 0);
+    FlushMessages();
+    COMPARE_CACHE(empty_chain);
+
+    ncmA.cbSize = sizeof(ncmA);
+    ret = SystemParametersInfoA(SPI_GETNONCLIENTMETRICS, 0, &ncmA, 0);
+    ok_int(ret, 1);
+    FlushMessages();
+    COMPARE_CACHE(empty_chain);
+
+    ncmA.cbSize = sizeof(ncmA);
+    ret = SystemParametersInfoA(SPI_GETNONCLIENTMETRICS, 1, &ncmA, 0);
+    ok_int(ret, 1);
+    FlushMessages();
+    COMPARE_CACHE(empty_chain);
+
+    /* ncmA */
+    ncmA.cbSize = 0;
+    ret = SystemParametersInfoA(SPI_GETNONCLIENTMETRICS, sizeof(ncmA), &ncmA, 0);
+    ok_int(ret, 0);
+    FlushMessages();
+    COMPARE_CACHE(empty_chain);
+
+    ncmA.cbSize = sizeof(ncmA);
+    ret = SystemParametersInfoA(SPI_GETNONCLIENTMETRICS, sizeof(ncmA), &ncmA, 0);
+    ok_int(ret, 1);
+    FlushMessages();
+    COMPARE_CACHE(empty_chain);
+
+    ncmA.cbSize = sizeof(ncmA);
+    ret = SystemParametersInfoW(SPI_GETNONCLIENTMETRICS, 0, &ncmA, 0);
+    ok_int(ret, 0);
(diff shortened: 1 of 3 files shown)


HBelusca:
Yep this needs some apitests.

win32ss/user/ntuser/sysparams.c
```diff
@@ -942,18 +1022,57 @@ SpiGetSet(UINT uiAction, UINT uiParam, PVOID pvParam, FLONG fl)
 
         case SPI_GETNONCLIENTMETRICS:
         {
-            return SpiGet(pvParam, &gspv.ncm, sizeof(NONCLIENTMETRICSW), fl);
+            if (uiParam == 0 && !SpiGetInt(pvParam, &uiParam, fl))
+                return 0;
+
+            if (uiParam == sizeof(NONCLIENTMETRICSA))
```
HBelusca, line 1028: Regarding that size check: I'm not sure this constraint (passing a redundant `sizeof(NONCLIENTMETRICS)` as parameter of `SystemParametersInfo` is systematically verified, on Windows 2003 (which is the version we are being compatible with).
This definitively requires an API test.

win32ss/user/ntuser/sysparams.c
```diff
@@ -797,6 +797,86 @@ SpiNotifyNCMetricsChanged(VOID)
...
+static void
+SpiLogFontWideToAnsi(const LOGFONTW *pLFW, LPLOGFONTA pLFA)
+{
+    RtlCopyMemory(pLFA, pLFW, offsetof(LOGFONTA, lfFaceName));
+    RtlUnicodeToMultiByteN(pLFA->lfFaceName, sizeof(pLFA->lfFaceName), NULL,
+                           pLFW->lfFaceName, sizeof(pLFW->lfFaceName));
+}
+
+static UINT_PTR
+SpiNonClientMetricsAnsiToWide(const NONCLIENTMETRICSA *pNCMA, LPNONCLIENTMETRICSW pNCMW)
```
HBelusca, line 817: The necessary ANSI/WIDE conversions are already done on client-side.
The kernel win32k (i.e. the function `NtUserSystemParametersInfo`) only deals with the UNICODE version of the NONCLIENTMETRICSW.

HBelusca (changes requested):

modules/rostests/apitests/user32/SystemParametersInfo.c
```diff
@@ -106,11 +106,130 @@ MSG_ENTRY CaptionHeight_chain[]={
 static void Test_NonClientMetrics()
 {
     NONCLIENTMETRICS NonClientMetrics;
```
HBelusca, line 108: That one has a problem, namely it's not explicitly told whether it's the unicode or ANSI version here (it should be unicode because we compile everything in unicode, but it makes the test look as if it was written with preconceptions).

modules/rostests/apitests/user32/SystemParametersInfo.c
```diff
@@ -106,11 +106,130 @@ MSG_ENTRY CaptionHeight_chain[]={
...
+    NONCLIENTMETRICSA ncmA;
+    NONCLIENTMETRICSW ncmW;
+    BOOL ret;
 
     /* WARNING: this test requires themes and dwm to be disabled */
 
     SetCursorPos(0,0);
 
+    ncmA.cbSize = 0;
+    ret = SystemParametersInfo(SPI_GETNONCLIENTMETRICS, 0, &ncmA, 0);
```
HBelusca, line 118: See comment above.

modules/rostests/apitests/user32/SystemParametersInfo.c
```diff
@@ -106,11 +106,130 @@ MSG_ENTRY CaptionHeight_chain[]={
...
+    ok_int(ret, 0);
+    FlushMessages();
+    COMPARE_CACHE(empty_chain);
+
+    ncmA.cbSize = sizeof(ncmA);
```
HBelusca, line 123: I would suggest storing the test data into an array, and then looping on the array, calling `SystemParametersInfoA` or SystemParametersInfoW` depending on whether an ANSI or unicode systemparametersinfo is tested.

learn-more:
The ticket seems to ask for a hybrid approach, where we accept a vista-sized struct in our NT 5.2 system (without recompilation).
I am not strictly against that idea, but your current approach does not do that.

katahiromz:
I can't do test on Vista because I don't have Vista environment.
Retired. Closed.

PR #3141, opened 2020-09-06
Ac97 import

TheDeadFish:
## Purpose
Import of AC97 sample driver

## Proposed changes
Import AC97 sample driver into drivers\wdm\audio\drivers\ac97
Insert missing definitions required to build into ddk

-- Explanation of commit "fixed buffer issue" --
Every time the DMA enable bit is written to 1, the emulated hardware also increments the buff index. The DMA enable bit was being set repeatedly during playback which screwed up the buffering. The fix checks if DMA is enabled, and if so it does not try and enable it again. 

## issues
Garbled buffers (sometimes it works perfectly)
Crashes

PR #4310, opened 2022-01-25
Import AC97 driver sample

Extravert-ir:
Reopening #4246. Also it makes sense to reference the "original" PR #3141


Vayne2500:
By the way, this is needed to be made into another PR:

https://github.com/reactos/reactos/pull/4305

Without it, it will crash the OS when booting.

binarymaster:
> Also it makes sense to reference the "original" PR #3141

I think the real origin of this code is this branch: https://github.com/TheDeadFish/reactos/commits/ac97_import_new2

media/doc/3rd Party Files.txt
```diff
@@ -151,6 +151,12 @@ Used Version: ???
 License: GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later.html)
 URL: http://www.vmware.com/download/open_sources.html
 
+Title: AC97 Sample Driver
+Path: drivers/wdm/audio/drivers/AC97
```
binarymaster, line 155:
```suggestion
Path: drivers/wdm/audio/drivers/ac97
```

JoachimHenze: I am no fan of frequent restructuring in the source tree.
Atm this folder does contain 3 subdirs: cmiDriver, mpu401, sb16. All of them are sound related. Fine.
I don't see any real benefit to move them around. I mean we are within root/drivers already. It's not like that they would be heavily misplaced within root/dll or something.

Extravert-ir: Audio folders are a mess, `wdm` is a useless folder there. It needs more than just a renaming

Extravert-ir:
This PR waits for portcls driver fixes to be merged.

JoachimHenze:
Attention: This PR did unhide:
https://jira.reactos.org/browse/CORE-18678

PR #4141, opened 2021-12-06
[TASKMGR] Preserve graphs history on resizes

z4v4l:
* also the graph stays preserved on the show/hide kernel time toggle
* an attempt to clean up the collected cruft was made as well

PR #4311, opened 2022-01-25
[TASKMGR]: Fixing the pool corruption bug

z4v4l:
* improper adjustment of the array index in graphctl.c, introduced in PR #4141 lead to an off-by-one pool corruption. fixing this.

JIRA issue: [CORE-18015](https://jira.reactos.org/browse/CORE-18015)

### DIFF
--- a/base/applications/taskmgr/graphctl.c
+++ b/base/applications/taskmgr/graphctl.c
@@ -150,7 +150,7 @@ GraphCtrl_AddPoint(PTM_GRAPH_CONTROL inst, BYTE val0, BYTE val1)
     t = inst->PointBuffer;
     Prev0 = *(t + inst->CurrIndex);
     Prev1 = *(t + inst->CurrIndex + inst->NumberOfPoints);
-    if (inst->CurrIndex < inst->NumberOfPoints)
+    if (inst->CurrIndex < inst->NumberOfPoints - 1)
     {
         inst->CurrIndex++;
     }


HBelusca:
From Discord convo:
```
ThFabba: 04:26
I just reorganized it as
    inst->CurrIndex++;
    if (inst->CurrIndex >= inst->NumberOfPoints)
    {
        inst->CurrIndex = 0;
    }

That make sense?

val: 04:34
I am tired af now, this looks ok, but why just not to change the line 153 into if (inst->CurrIndex < inst->NumberOfPoints - 1)?

ThFabba: 04:35
I tend to think the "CurrIndex == NumberOfPoints is not allowed" is expressed in a more obvious way in my version

val: 04:36
matter of taste, but since, you are the boss, yours wins. I guess. :smile:

ThFabba: 04:37
Heh. Yeah I'm getting tired, too. Maybe my opinion will differ in the morning :wink:
```

ThFabba (approved):
It's heap, not pool, but LGTM

z4v4l:
`NumberOfPoints` is the size of the subarray. hence `NumberOfPoints - 1` is the maximum index. Checking against it and then incrementing `CurrIndex` if it's less than that is not any less "readable" and made this way, it will modify `CurrIndex` only once, not twice as in that other variant. Hermes, let's not turn this tiny thing into another 100 comments worth debates, please.

base/applications/taskmgr/graphctl.c
```diff
@@ -150,7 +150,7 @@ GraphCtrl_AddPoint(PTM_GRAPH_CONTROL inst, BYTE val0, BYTE val1)
     t = inst->PointBuffer;
     Prev0 = *(t + inst->CurrIndex);
     Prev1 = *(t + inst->CurrIndex + inst->NumberOfPoints);
-    if (inst->CurrIndex < inst->NumberOfPoints)
```
SergeGautherie, line 153:
```
    if (++inst->CurrIndex == inst->NumberOfPoints)
    {
        inst->CurrIndex = 0;
    }

```

PR #4319, opened 2022-01-27
[TASKMGR] Don't set "Always On Top" option by default

TAN-Gaming:
## Purpose

- Don't set "Always On Top" option by default, like Windows does.

### DIFF
--- a/base/applications/taskmgr/taskmgr.c
+++ b/base/applications/taskmgr/taskmgr.c
@@ -863,7 +863,7 @@ void LoadSettings(void)
     TaskManagerSettings.ActiveTabPage = 0;
 
     /* Options menu settings */
-    TaskManagerSettings.AlwaysOnTop = TRUE;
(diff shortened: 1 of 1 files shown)


JoachimHenze:
Windows does remember that setting. What is set on a freshly installed 2k3sp2 is the question?

binarymaster:
This partially reverts commit e8a7e0da, and I'm totally not sure it's correct.

From what I remember, Task Manager is always on top of other windows **by default**.

TAN-Gaming:
> This partially reverts commit [e8a7e0d](https://github.com/reactos/reactos/commit/e8a7e0da2bcbd842f95e4224e72bfb5d748dbd05), and I'm totally not sure it's correct.
> 
> From what I remember, Task Manager is always on top of other windows **by default**.

I'm not sure about Win2003, but Win7 and Win10 don't set it by default.

HBelusca:
PR is invalid: See the discussion in PR #4169 that investigated what the actual defaults were.

PR #4323, opened 2022-01-29
[TASKMGR] Process page: Add "Properties" and "Open File Location" context menu item

TAN-Gaming:
## Purpose

- Add "Properties" and "Open File Location" context menu item

VirtualBox_ReactOS_08_02_2022_15_16_32

VirtualBox_ReactOS_08_02_2022_15_18_44


sashaqwert:
The idea is good. This is missing in Windows Server 2003.

katahiromz:
You can use `GetModuleFileNameEx` function.

TAN-Gaming:
> You can use `GetModuleFileNameEx` function.

Yeah, I used it at line 1011.

JoachimHenze:
I do absolutely love the intended functionality!
Small question though: Could it be done without the ole32 import somehow?

base/applications/taskmgr/procpage.c
```diff
@@ -931,3 +933,160 @@ int CALLBACK ProcessPageCompareFunc(LPARAM lParam1, LPARAM lParam2, LPARAM lPara
...
+    }
+
+    hProcess = OpenProcess(PROCESS_QUERY_INFORMATION, FALSE, dwProcessId);
+
+    if (!hProcess)
+    {
+        return FALSE;
+    }
+
+    hLibrary = LoadLibraryW(L"kernel32_vista.dll");
```
JoachimHenze, line 961: I don't think the taskmgr ever should use anything from kernel32_vista. Neither today, nor in the future.
Other applications, like Mozilla Firefox can absolutely offer similar functionality, without using that.
What is the benefit of this GetProcessExecutablePath() implementation compared to the other one?
Would it make sense to mention its specific strengths within its CORE-16145-comment maybe?

JoachimHenze: Unless there is really some benefit i here that I would miss, I would recommend to simply strip that second implementation entirely.

base/applications/taskmgr/CMakeLists.txt
```diff
@@ -24,6 +24,6 @@ file(GLOB taskmgr_rc_deps res/*.*)
 add_rc_deps(taskmgr.rc ${taskmgr_rc_deps})
 add_executable(taskmgr ${SOURCE} taskmgr.rc)
 set_module_type(taskmgr win32gui UNICODE)
-add_importlibs(taskmgr advapi32 user32 gdi32 shell32 shlwapi comctl32 msvcrt kernel32 ntdll)
+add_importlibs(taskmgr advapi32 user32 gdi32 shell32 shlwapi comctl32 ole32 msvcrt kernel32 ntdll)
```
JoachimHenze, line 27: ole32? What is it used for? Any alternatives?

base/applications/taskmgr/procpage.c
TAN-Gaming: > I don't think the taskmgr ever should use anything from kernel32_vista. Neither today, nor in the future. Other applications, like Mozilla Firefox can absolutely offer similar functionality, without using that. What is the benefit of this GetProcessExecutablePath() implementation compared to the other one? Would it make sense to mention its specific strengths within its CORE-16145-comment maybe?

It's just easier to use and don't requires path conversion.

TAN-Gaming: > Unless there is really some benefit in here that I would miss, I would recommend to simply strip that second implementation entirely.

OK, I'll remove.

base/applications/taskmgr/CMakeLists.txt
TAN-Gaming: > ole32? What is it used for? Any alternatives?

For `SHOpenFolderAndSelectItems`

[https://docs.microsoft.com/en-us/windows/win32/api/shlobj_core/nf-shlobj_core-shopenfolderandselectitems](url)

An alternative maybe `ShellExecute`

HBelusca: Taskmgr should only use functions from kernel32/ntdll for retrieving a suitable Win32 path, and then *dynamically load* shell32 to use that API.

base/applications/taskmgr/procpage.c
```diff
@@ -931,3 +933,156 @@ int CALLBACK ProcessPageCompareFunc(LPARAM lParam1, LPARAM lParam2, LPARAM lPara
...
+            size_t NewLen = wcslen(pszExePath) - 4;
+            memmove(pszExePath, pszExePath + 4, (NewLen + 1) * sizeof(WCHAR));
+        }
+        else if (_wcsnicmp(pszExePath, L"\\SystemRoot", 11) == 0)
+        {
+            WCHAR szWinDir[MAX_PATH];
+            HRESULT hr;
+
+            bSuccess = FALSE;
+            hr = SHGetFolderPathW(NULL, CSIDL_WINDOWS, NULL, 0, szWinDir);
```
HBelusca, line 1019: Please use NT path functions to deal with all that.

base/applications/taskmgr/procpage.c
```diff
@@ -931,3 +933,156 @@ int CALLBACK ProcessPageCompareFunc(LPARAM lParam1, LPARAM lParam2, LPARAM lPara
...
+                    memcpy(pszExePath, szWinDir, WinDirLen * sizeof(WCHAR));
+
+                    bSuccess = TRUE;
+                }
+            }
+        }
+
+        if (bSuccess)
+        {
+            memcpy(lpExePath, pszExePath, dwLength * sizeof(WCHAR));
```
HBelusca, line 1039: Use stringsafe functions instead (here and elsewhere, for example above).

HBelusca (changes requested):

base/applications/taskmgr/CMakeLists.txt
TAN-Gaming: > Taskmgr should only use functions from kernel32/ntdll for retrieving a suitable Win32 path, and then _dynamically load_ shell32 to use that API.

Should I use `CreateProcess` instead of `ShellExecute` ?

base/applications/taskmgr/procpage.c
TAN-Gaming: OK.

base/applications/taskmgr/CMakeLists.txt
HBelusca: > An alternative maybe using ShellExecute to start explorer.exe with /select,"<EXE_PATH>" option.

What does Windows' taskmgr do there? (you can see what it does by monitoring it with e.g. ApiMonitor from Rohitab)

JoachimHenze: taskmgr of XPSP3 has no static imports to OLE32 (only dynamic). But it does have static imports to shell32.
So I do vote for hbeluscas proposal to prefer "ShellExectute /select" over adding the OLE32 import
image

HBelusca: Note that taskmgr has a *delayed-import* to ole32 though.

TAN-Gaming: > > An alternative maybe using ShellExecute to start explorer.exe with /select,"<EXE_PATH>" option.
> 
> What does Windows' taskmgr do there? (you can see what it does by monitoring it with e.g. ApiMonitor from Rohitab)

I used Dependency Walker to inspect `taskmgr.exe` from Windows 7, I saw
- ole32.dll and shell32.dll imported
- `SHOpenFolderAndSelectItems`, `ShellExecuteExW` and more shell functions used
- `QueryFullProcessImageNameW` function used

but according to @sashaqwert , Windows 2003 doesn't have this functionality.

HBelusca: Do you have a fallback for GetModuleFileNameExW ? (because if so this means that using that function was not necessary to begin with)

TAN-Gaming: > Do you have a fallback for GetModuleFileNameExW ? (because if so this means that using that function was not necessary to begin with)

Maybe `NtQueryInformationProcess` but it returns path in device form and requires conversion.

TAN-Gaming: > Note that taskmgr has a _delayed-import_ to ole32 though. But I would prefer if all that stuff could be even manually loaded so that one could use taskmgr in an extreme minimal environment that does not contain any shell stuff.

I think we can simply disable this functionality in that case, because we can't open explorer or open properties dialog without shell stuffs anyway.

JoachimHenze: Yes: "disabling the feature whenever delay-load is not found", is a good idea to have good user experience even in modded setups.
Still the shell32-way ShellExecuteExW() is better than the ole32-way for several reasons:
1.) We have a dependency to shell32 already, while we have none to ole32.
2.) it keeps the set of required dlls for full functionality closer to XP/2k3
3.) imho it is more likely to find a "super-slimified" version of ros without ole32, than one without shell32.

TAN-Gaming: @JoachimHenze I agree with you to remove ole32 dependency for portability, but I have a question.
Are there any ways to open file properties dialog that work on ROS and don't require ole32? I tried  `ShellExecuteEx` method but it doesn't work for me.

JoachimHenze: Maybe you did wrong with the flags: I just googled, but did not try it:

SHELLEXECUTEINFO ShExecInfo ={0};
ShExecInfo.cbSize = sizeof(SHELLEXECUTEINFO);
ShExecInfo.fMask = SEE_MASK_INVOKEIDLIST ;
ShExecInfo.hwnd = NULL;
ShExecInfo.lpVerb = "properties";
ShExecInfo.lpFile = "c:\\"; //can be a file as well
ShExecInfo.lpParameters = ""; 
ShExecInfo.lpDirectory = NULL;
ShExecInfo.nShow = SW_SHOW;
ShExecInfo.hInstApp = NULL; 
ShellExecuteEx(&ShExecInfo);

above taken from https://www.codeproject.com/Articles/1842/A-newbie-s-elementary-guide-to-spawning-processes
If it wouldn't work out of the box, make sure to try in first when running our taskmgr on Windows. Maybe some other bug in ros (outside of taskmgr) is in your way still.


(I found also something that looks pretty hacky https://stackoverflow.com/questions/33472585/open-windows-file-folder-properties-dialog-from-c (scroll down a bit, the one with that Sleep()...). But that would for sure be no proper solution.

TAN-Gaming: > Maybe you did wrong with the flags: I just googled, but did not try it:
> 
> SHELLEXECUTEINFO ShExecInfo ={0}; ShExecInfo.cbSize = sizeof(SHELLEXECUTEINFO); ShExecInfo.fMask = SEE_MASK_INVOKEIDLIST ; ShExecInfo.hwnd = NULL; ShExecInfo.lpVerb = "properties"; ShExecInfo.lpFile = "c:\"; //can be a file as well ShExecInfo.lpParameters = ""; ShExecInfo.lpDirectory = NULL; ShExecInfo.nShow = SW_SHOW; ShExecInfo.hInstApp = NULL; ShellExecuteEx(&ShExecInfo);
> 
> above taken from https://www.codeproject.com/Articles/1842/A-newbie-s-elementary-guide-to-spawning-processes If it wouldn't work out of the box in ros, make sure to try it first when running our taskmgr on Windows. Maybe some other bug in ros (outside of taskmgr) is in your way still.
> 
> (I found also something that looks pretty hacky https://stackoverflow.com/questions/33472585/open-windows-file-folder-properties-dialog-from-c (scroll down a bit, the one with that Sleep()...). But that would for sure be no proper solution.

Thanks. I'll try it.

base/applications/taskmgr/lang/en-US.rc
```diff
@@ -135,6 +135,9 @@ BEGIN
             MENUITEM "&Low", ID_PROCESS_PAGE_SETPRIORITY_LOW
         END
         MENUITEM "Set &Affinity...", ID_PROCESS_PAGE_SETAFFINITY
+        MENUITEM SEPARATOR
+        MENUITEM "&Properties", ID_PROCESS_PAGE_PROPERTIES
+        MENUITEM "&Open File Location", ID_PROCESS_PAGE_OPENFILELOCATION
```
JoachimHenze, line 140: This needs to be done/duplicated for all other languages too ofc.

JoachimHenze (changes requested):

base/applications/taskmgr/lang/en-US.rc
TAN-Gaming: OK.

base/applications/taskmgr/procpage.c
TAN-Gaming: Fixed.

base/applications/taskmgr/procpage.c
```diff
@@ -931,3 +933,140 @@ int CALLBACK ProcessPageCompareFunc(LPARAM lParam1, LPARAM lParam2, LPARAM lPara
...
+    DWORD dwProcessId;
+    WCHAR szExePath[MAX_PATH];
+
+    dwProcessId = GetSelectedProcessId();
+
+    if (GetProcessExecutablePath(dwProcessId, szExePath, _countof(szExePath)))
+    {
+        return;
+    }
+}
```
JoachimHenze, line 1054: unfinished still. I guess it's WIP.

base/applications/taskmgr/lang/de-DE.rc
```diff
@@ -137,6 +137,9 @@ BEGIN
             MENUITEM "N&iedrig", ID_PROCESS_PAGE_SETPRIORITY_LOW
         END
         MENUITEM "&Zugehörigkeit festlegen...", ID_PROCESS_PAGE_SETAFFINITY
+        MENUITEM SEPARATOR
+        MENUITEM "&Properties", ID_PROCESS_PAGE_PROPERTIES
+        MENUITEM "&Open File Location", ID_PROCESS_PAGE_OPENFILELOCATION
     END
```
JoachimHenze, line 143: MENUITEM "&Eigenschaften", ID_PROCESS_PAGE_PROPERTIES
MENUITEM "Übergeordneten Ordner ö&ffnen", ID_PROCESS_PAGE_OPENFILELOCATION

JoachimHenze (changes requested):

base/applications/taskmgr/CMakeLists.txt
learn-more: Why the obsession with ole32?
Just make it a delay-import and call it a day, no need to make tons of unreadable code / hacks to prevent loading a system dll that most likely will be loaded by shell32 anyway...

JoachimHenze: shell32 on XP also loads ole32, but it **delay-loads** it. So there is at least a possibility, that shell32 can offer that specific functionality (file prop dlg) without ole32.
But yes: Generally the delay-loading at all is the more important part of the two. The question whether shell32 or ole32 is not **that** important. But still worth a thought, given that shell32 is already a dependency here historically, but ole32 not directly.

base/applications/taskmgr/lang/de-DE.rc
TAN-Gaming: Updated.

base/applications/taskmgr/procpage.c
TAN-Gaming: I tried both `ShellExecute` method but still not working.

JoachimHenze: Is it only non-working in ros? Or is it also non-working in Windows? If one of the solutions would work on Windows, then it would be fine to use that. And let (someone else) fix the remaining bug otherwhere in ros later.

TAN-Gaming: Not tried it on Windows yet. I'll try it and use a working solution.

TAN-Gaming: Solution 1 works, tested on WinXP SP3

base/applications/taskmgr/procpage.c
```diff
@@ -931,3 +933,152 @@ int CALLBACK ProcessPageCompareFunc(LPARAM lParam1, LPARAM lParam2, LPARAM lPara
...
+        static const WCHAR szKernelPath[] = L"\\SystemRoot\\System32\\ntoskrnl.exe";
+
+        if (dwLength >= _countof(szKernelPath))
+        {
+            StringCchCopyW(pszExePath, dwLength, szKernelPath);
+
+            bSuccess = TRUE;
+        }
+    }
+    else
```
JoachimHenze, line 992: special treatment for SYSTEM? Is that necessary also on XP? What of "the default way of doing things below" would fail for ntoskrnl.exe on Windows?

JoachimHenze (approved):

base/applications/taskmgr/procpage.c
```diff
@@ -931,3 +933,152 @@ int CALLBACK ProcessPageCompareFunc(LPARAM lParam1, LPARAM lParam2, LPARAM lPara
...
+
+    HMODULE hLibrary;
+    PFN_GETMODULEFILENAMEEXW func;
+    DWORD dwRet = 0;
+
+    hLibrary = LoadLibraryW(L"psapi.dll");
+
+    if (hLibrary)
+    {
+        func = (PFN_GETMODULEFILENAMEEXW)GetProcAddress(hLibrary, "GetModuleFileNameExW");
```
JoachimHenze, line 949: Is it possible to use NtQuerySystemInformation() from ntdll here instead maybe?
https://docs.microsoft.com/en-us/windows/win32/api/winternl/nf-winternl-ntquerysysteminformation
Windows XPSP3 taskmgr also uses NtQuerySystemInformation() from ntdll.dll, but does not import from psapi.dll
image

JoachimHenze: Hmm, maybe not. Otherwise we would have that information at hand already from other parts of taskmgr.
Guess the current solution to use psapi.dll might be the best then already here. Other peoples ideas are appreciated though.

JoachimHenze: Oh I found something else that may be of (future) interest:
"GetModuleFileNameEx dosen't work with 32-bit app in 64-bit systems, so I've use GetProcessImageFileName with _fullpath and _splitpath_s and its work fine." from
https://stackoverflow.com/questions/30254793/getmodulefilenameex-split-output

base/applications/taskmgr/procpage.c
TAN-Gaming: Yes, `GetModuleFileNameEx` will fail with ntoskrnl.exe and I found that Windows even hardcoded the path for it.
[https://stackoverflow.com/questions/51985730/not-able-to-get-executable-path-for-process-id-4-ntoskrnl-exe](url)

WinXP's Taskmgr doesn't have this functionality, on Win7 it will redirect to My Computer, but on Win10 it goes to ntoskrnl.exe

base/applications/taskmgr/procpage.c
TAN-Gaming: > Oh I found something else that may be of (future) interest: "GetModuleFileNameEx dosen't work with 32-bit app in 64-bit systems, so I've use GetProcessImageFileName with _fullpath and _splitpath_s and its work fine." from https://stackoverflow.com/questions/30254793/getmodulefilenameex-split-output

Interesting, if it can convert device path to normal path without using `QueryDosDevice` method which is not performance-friendly so it will be a fix for CORE-16145 as well.

HBelusca: > without using QueryDosDevice method which is not performance-friendly

It's not as if you were calling that function in a tight loop every fraction of a second...

TAN-Gaming: @HBelusca I'll definitely use `GetProcessImageFileName` or `NtQueryInformationProcess` instead of `GetModuleFileNameEx` after finished working on path conversion.

JoachimHenze: The reason we do PRs is for aiming to get things more correct right from the start.

TAN-Gaming: I mean portability improvements for shell32 (as suggested by @HBelusca), but for `GetProcessExecutablePath` I'll fix it.

TAN-Gaming:
I've opened a JIRA issue for "Properties" problem.
[https://jira.reactos.org/browse/CORE-18035](https://jira.reactos.org/browse/CORE-18035)

base/applications/taskmgr/procpage.c
TAN-Gaming: `GetProcessExecutablePath` fixed.

base/applications/taskmgr/procpage.c
```diff
@@ -931,3 +935,202 @@ int CALLBACK ProcessPageCompareFunc(LPARAM lParam1, LPARAM lParam2, LPARAM lPara
...
+    }
+
+    pszPath = HeapAlloc(GetProcessHeap(), 0, dwLength * sizeof(WCHAR));
+
+    if (!pszPath)
+    {
+        return FALSE;
+    }
+
+    for (WCHAR cDrive = L'A'; cDrive <= L'Z'; cDrive++)
```
JoachimHenze, line 957: Please move the declaration WCHAR cDrive out of the for-loop to the top of the function to make this code compatible to more compilers.

TAN-Gaming: OK.

learn-more: All those compilers are unsupported, and Joachim does not send PR's himself so feel free to ignore this request.

learn-more:
@TAN-Gaming the blocker _should_ be resolved now in master.

TAN-Gaming:
> @TAN-Gaming the blocker _should_ be resolved now in master.

Thanks.

learn-more:
@TAN-Gaming I have changed the PR to a draft since you keep pushing new changes, please let us know when it is ready for review.

TAN-Gaming:
The PR is ready.

base/applications/taskmgr/procpage.c
```diff
@@ -931,3 +936,300 @@ int CALLBACK ProcessPageCompareFunc(LPARAM lParam1, LPARAM lParam2, LPARAM lPara
...
+    WCHAR cDrive;
+    BOOL bSuccess = FALSE;
+
+    /* Check if lpPath is a device path */
+    if (_wcsnicmp(lpPath, L"\\Device\\", 8) != 0)
+    {
+        return FALSE;
+    }
+
+    pszPath = HeapAlloc(GetProcessHeap(), 0, dwLength * sizeof(WCHAR));
```
HBelusca, line 952: Seeing how you use `pszPath` and `lpPath`, you could directly skip that and in the `StringCchPrintfW` call below, use `lpPath`.

base/applications/taskmgr/procpage.c
```diff
@@ -931,3 +936,300 @@ int CALLBACK ProcessPageCompareFunc(LPARAM lParam1, LPARAM lParam2, LPARAM lPara
...
+
+    pszExePath = HeapAlloc(GetProcessHeap(), 0, ImagePath->Length + sizeof(WCHAR));
+
+    if (!pszExePath)
+    {
+        goto Cleanup;
+    }
+
+    memcpy(pszExePath, ImagePath->Buffer, ImagePath->Length);
+    pszExePath[ImagePath->Length / sizeof(WCHAR)] = UNICODE_NULL;
```
HBelusca, line 1042:
```suggestion
    StringCbCopyNW(pszExePath, ImagePath->Length + sizeof(WCHAR), ImagePath->Buffer, ImagePath->Length);
```

base/applications/taskmgr/procpage.c
```diff
@@ -931,3 +936,300 @@ int CALLBACK ProcessPageCompareFunc(LPARAM lParam1, LPARAM lParam2, LPARAM lPara
     }
     return ret;
 }
+
+static BOOL DevicePathToDosPath(LPWSTR lpPath, DWORD dwLength)
```
HBelusca, line 940: It may be of interest to have the dwLength in number of bytes instead (for ease of usage of this function into the other one).

base/applications/taskmgr/procpage.c
```diff
@@ -931,3 +936,300 @@ int CALLBACK ProcessPageCompareFunc(LPARAM lParam1, LPARAM lParam2, LPARAM lPara
...
+    {
+        HeapFree(GetProcessHeap(), 0, DynamicBuffer);
+    }
+
+    return dwRet;
+}
+
+static DWORD GetProcessExecutablePathById(DWORD dwProcessId, LPWSTR lpExePath, DWORD dwLength)
+{
+    DWORD dwRet = 0;
```
HBelusca, line 1075: Question for here and the other function above: What does ret == 0 mean? and -1 ? Success or failure?

base/applications/taskmgr/procpage.c
```diff
@@ -931,3 +936,300 @@ int CALLBACK ProcessPageCompareFunc(LPARAM lParam1, LPARAM lParam2, LPARAM lPara
...
+
+    if (dwProcessId == 0)
+    {
+        return 0;
+    }
+    
+    /* PID = 4 or "System" */
+    if (dwProcessId == 4)
+    {
+        static const WCHAR szKernelExe[] = L"\\ntoskrnl.exe";
```
HBelusca, line 1085:
```suggestion
        static const WCHAR* szKernelExe = L"\\ntoskrnl.exe";
```

base/applications/taskmgr/procpage.c
```diff
@@ -931,3 +936,300 @@ int CALLBACK ProcessPageCompareFunc(LPARAM lParam1, LPARAM lParam2, LPARAM lPara
     }
     return ret;
 }
+
+static BOOL DevicePathToDosPath(LPWSTR lpPath, DWORD dwLength)
+{
+    WCHAR *pszPath;
```
HBelusca, line 942: `WCHAR *` --> `PWSTR` here and below (6 more changes).

base/applications/taskmgr/procpage.c
TAN-Gaming: Is it safe to use `StringCbCopyNW` with `UNICODE_STRING` ? The documentation says that for `StringCbCopyNW` the source string must be null-terminated.

base/applications/taskmgr/procpage.c
TAN-Gaming: If the function succeeds, the return value is the path length, not including the terminating null character.

If the function fails, the return value is zero.

If the buffer is not large enough, the return value is the required size of the buffer, in characters, including the terminating null character.

base/applications/taskmgr/procpage.c
TAN-Gaming: What's wrong with this?

base/applications/taskmgr/procpage.c
TAN-Gaming: What's wrong with this?

base/applications/taskmgr/procpage.c
TAN-Gaming: OK.

base/applications/taskmgr/procpage.c
TAN-Gaming: OK.

base/applications/taskmgr/procpage.c
TAN-Gaming: What's wrong with this?

base/applications/taskmgr/procpage.c
TAN-Gaming: OK.

base/applications/taskmgr/procpage.c
TAN-Gaming: What's wrong with this?

base/applications/taskmgr/procpage.c
TAN-Gaming: OK.

base/applications/taskmgr/procpage.c
TAN-Gaming: OK.

base/applications/taskmgr/procpage.c
TAN-Gaming: OK.

base/applications/taskmgr/procpage.c
TAN-Gaming: What's wrong with this?

base/applications/taskmgr/procpage.c
TAN-Gaming: What's wrong with this?

base/applications/taskmgr/procpage.c
TAN-Gaming: What's wrong with this?

base/applications/taskmgr/procpage.c
TAN-Gaming: OK.

base/applications/taskmgr/procpage.c
TAN-Gaming: If I change it to a pointer, I won't be able to use `sizeof(szKernelExe)` and `_countof(szKernelExe)` .

base/applications/taskmgr/procpage.c
HBelusca: yes but actually you specify what the actual number of characters is in the source string, so it doesn't matter.

PR #4325, opened 2022-01-30
[NTOS:CC] Unintialize private cache maps before purging the cache sec…

GeoB99:
…tion

Before purging the data cache of a certain section of a file from system cache, we have to unintialize the private cache maps of that section if a filesystem or any other component prompts the kernel to do so.

I have noticed this whilst working on #4259, getting viciously spammed by the debugger when copying a directory or file from a shared folder to ReactOS, that `CcPurgeCacheSection` makes no attempt on unintializing private cache maps.

### DIFF
--- a/ntoskrnl/cc/fs.c
+++ b/ntoskrnl/cc/fs.c
@@ -161,6 +161,7 @@ CcPurgeCacheSection (
     IN BOOLEAN UninitializeCacheMaps)
 {
     PROS_SHARED_CACHE_MAP SharedCacheMap;
+    PPRIVATE_CACHE_MAP PrivateCacheMap;
     LONGLONG StartOffset;
     LONGLONG EndOffset;
     LIST_ENTRY FreeList;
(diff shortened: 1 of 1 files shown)


ntoskrnl/cc/fs.c
```diff
@@ -173,18 +174,36 @@ CcPurgeCacheSection (
...
+    if (UninitializeCacheMaps)
+    {
+        /*
+         * We have gotten the acknowledgement that
+         * the caller wants to unintialize the private
+         * cache maps so let's do this. Since we already
+         * have the shared cache map from above, iterate
+         * over that cache's private lists.
+         */
+        while (SharedCacheMap && (!IsListEmpty(&SharedCacheMap->PrivateList)))
```
ThFabba, line 194: It doesn't look like SharedCacheMap can be or become NULL here

ThFabba (approved):

PR #4328, opened 2022-01-30
[USER32] Don't delay-load-link to imm32.dll

katahiromz:
## Purpose
Reduce binary size.

## Proposed changes

- Don't directly call the `IMM32` functions. Use `IMM_FN` instead.
- Modify `CMakeLists.txt` to unlink `imm32.dll`.

## Comparison
BEFORE:
before
1,553,920 bytes.

AFTER:
after
1,552,384 bytes.

`1,552,384 - 1,553,920 == -1536 bytes.`

### DIFF
--- a/win32ss/user/user32/CMakeLists.txt
+++ b/win32ss/user/user32/CMakeLists.txt
@@ -87,7 +87,7 @@ if(MSVC AND (ARCH STREQUAL "i386"))
     target_sources(user32 PRIVATE $<TARGET_OBJECTS:ftol2_sse>)
 endif()
 
-add_delay_importlibs(user32 imm32 usp10)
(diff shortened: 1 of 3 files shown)


JoachimHenze:
Does it work with a gcc build? We do sometimes have issues with delay-loads in GCC. But not everywhere. So it might be okay here. see https://jira.reactos.org/browse/CORE-6504 for more info. If it works for GCC then thumbs up from my side.

katahiromz:
I used GCC compiler to build.

JoachimHenze:
And confirmed: It does bring us closer to XPSP3 here also:
image

JoachimHenze (approved):

katahiromz:
working-well
Working well.

PR #4334, opened 2022-02-03
[NTUSER] Implement NtUserAssociateInputContext

katahiromz:
## Purpose

## Proposed changes

- Add `IntReAllocatePoolWithTag` function in `window.c`.
- Define `WINDOWLIST` structure in `"window.h"`.
- Add `IntBuildHwndList` and `IntFreeHwndList` helper functions in `window.c`.
- Add `IntAssociateInputContext` and `IntAssociateInputContextEx` helper functions in `ime.c`.
- Implement `NtUserAssociateInputContext` function.


sdk/include/xdk/exfuncs.h
```diff
@@ -785,6 +785,24 @@ ExFreePoolWithTag(
...
+{
+    PVOID pNew = ExAllocatePoolWithTag(PoolType, cbNew, Tag);
+    if (!pNew)
+        return NULL;
+
+    RtlCopyMemory(pNew, pOld, ((cbOld < cbNew) ? cbOld : cbNew));
+    ExFreePoolWithTag(pOld, Tag);
+    return pNew;
+}
+
```
GeoB99, line 805: MS headers do not have such kind of function exported nor it has ever existed anywhere else so please revert this and implement it something like `IntReAllocatePool` or whatever since it seem it's exclusively for NTUSER here.

GeoB99 (changes requested):

sdk/include/xdk/exfuncs.h
katahiromz: Reverted and implemented as `IntReAllocatePoolWithTag`.

PR #4336, opened 2022-02-03
[SHELL32] ShellExecute: Implement SEE_MASK_INVOKEIDLIST

learn-more:
JIRA issue: [CORE-18035](https://jira.reactos.org/browse/CORE-18035)


dll/win32/shell32/shlexec.cpp
```diff
@@ -1431,60 +1428,128 @@ static HRESULT shellex_run_context_menu_default(IShellExtInit *obj,
     return r;
 }
 
-static HRESULT shellex_load_object_and_run(HKEY hkey, LPCGUID guid, LPSHELLEXECUTEINFOW sei)
+namespace
```
HBelusca, line 1431: Might be of interest (for later) to have that thing available on the whole shell32 🤔 
But why does it need to reside in a nameless namespace?

learn-more:
The case with the extension cut off is not handled correctly yet.
Depending on how complicated this is, I will merge the PR without that.

yagoulas:
The code makes sense. I don't remember anything about how ShellExecute works though.

dll/win32/shell32/shlexec.cpp
```diff
@@ -1431,60 +1428,139 @@ static HRESULT shellex_run_context_menu_default(IShellExtInit *obj,
...
+        return hr;
 
-    r = obj->QueryInterface(IID_PPV_ARG(IObjectWithSite, &ows));
-    if (FAILED(r))
-        goto end;
+    return shf->GetUIObjectOf(NULL, 1, &pidllast, IID_NULL_PPV_ARG(IContextMenu, &cm));
+}
 
-    ows->SetSite(NULL);
+static HRESULT IContextMenu_exec(LPSHELLEXECUTEINFOW sei)
```
yagoulas, line 1514: Perhaps a different name could be chosen here. We already have a different code path that invokes the IContextMenu of shell extensions and this could become confusing.

learn-more: Thanks, will do

PR #3609, opened 2021-04-13
[HALX86] Disable Lazy IRQL in APIC HAL

DarkFire01:
## Purpose
-First couple steps to get ACPI APIC working.

## Proposed changes

_Describe what you propose to change/add/fix with this pull request._
  -Cleaned APIC header files 
  -Created files for building the respective HALs
  -Disabled LazyIRQL until a solution for it is found
         -APIC_LAZY_IRQL being enabled stops virtualbox from being able to boot at the current time, during my time attempting to solve the issue I discovered that it occurs after a raise IRQL event; this doesn't succeed and causes an asset fail before finishing the initialization of the later stages of the kernel.

PR #4339, opened 2022-02-05
[HAL:APIC] Ensure the interrupt gets requested immediately in ApicRequestSelfInterrupt.

ThFabba:
## Purpose
First, wait for the APIC to be idle before requesting an interrupt. Then, wait until we actually see the interrupt request bit set.
This will ensure that if we're at a low enough IRQL, the software interrupt will happen immediately (which is an invariant that Windows's IRQL mechanism relies on).

JIRA issue: [CORE-17663](https://jira.reactos.org/browse/CORE-17663)

### DIFF
--- a/hal/halx86/apic/apic.c
+++ b/hal/halx86/apic/apic.c
@@ -138,7 +138,18 @@ FORCEINLINE
 VOID
 ApicRequestSelfInterrupt(IN UCHAR Vector, UCHAR TriggerMode)
 {
+    ULONG Flags;
     APIC_INTERRUPT_COMMAND_REGISTER Icr;
+    APIC_INTERRUPT_COMMAND_REGISTER IcrStatus;
+
+    /*
+     * The IRR registers are spaced 16 bytes apart and hold 32 status bits each.
+     * Pre-compute the register and bit that match our vector.
+     */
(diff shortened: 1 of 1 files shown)


ThFabba:
@DarkFire01 I can't add you as a reviewer, but would appreciate if you had a look.

HBelusca:
Could it be possible that it fixes what was (temporarily?) disabled by PR #3609 ?

HBelusca (approved):
lgtm

ThFabba:
> Could it be possible that it fixes what was (temporarily?) disabled by PR #3609 ?

I'm not sure if the two are related, it's possible. But this PR fixes an active bug in master, so if the two are related then it would appear that disabling lazy IRQLs actually did not completely fix whatever the issue was.

DarkFire01 (approved):
Ah cool :D that looks great to me!
You're going to have situations where interrupts targeting other groups (IPIs) will still do this, but honestly if we do something similar in nature to what you do here we'll be fine when it comes to master.
Thank you for looking at this!

Extravert-ir:
> Could it be possible that it fixes what was (temporarily?) disabled by PR #3609 ?

This change is not relevant anymore - lazy IRQL was fixed since than and turned back on.

PR #4340, opened 2022-02-05
[NTOS:SE] Access Check Overhaul -- Deny access to the caller if not permitted

GeoB99:
There are two fundamental problems when it comes to access checks in ReactOS. First, the internal function `SepAccessCheck` which is the heart and brain of the whole access checks logic of the kernel warrants access to the calling thread of a process to an object even though access could not be given. 

This can potentially leave security issues as we literally leave objects to be touched indiscriminately by anyone regardless of their ACEs in the DACL of a security descriptor. Second, the current access check code doesn't take into account the fact that an access token can have restricted SIDs. In such scenario we must perform additional access checks by iterating over the restricted SIDs of the primary token by comparing the SID equality and see if the group can be granted certain rights based on the ACE policy that represents the same SID.

Part of `SepAccessCheck`'s code logic will be split for a separate private kernel routine, `SepAnalyzeAcesFromDacl`. The reasons for this are primarily two -- such code is subject to grow eventually as we'll support different type ACEs and handle them accordingly -- and we avoid further code duplicates. On Windows Server 2003 there are 5 different type of ACEs that are supported for access checks:

- **ACCESS_DENIED_ACE_TYPE** (supported by ReactOS)
- **ACCESS_ALLOWED_ACE_TYPE** (supported by ReactOS)
- **ACCESS_DENIED_OBJECT_ACE_TYPE**
- **ACCESS_ALLOWED_OBJECT_ACE_TYPE**
- **ACCESS_ALLOWED_COMPOUND_ACE_TYPE**

Once the foundation of access checks is properly solid by getting rid of the ugly hack present in the code, this allows for more opportunity to continue further on implementing object type access checks, compound ACE handling, and whatnot. For now we'll live by the basics.
## TODO

- [x] Make ReactOS bootable again
- [x] Investigate most of the "failed to grant access rights" occurrences on ReactOS

**JIRA Issue:** [CORE-9184](https://jira.reactos.org/browse/CORE-9184)

### DIFF
--- a/base/services/dhcpcsvc/dhcp/pipe.c
+++ b/base/services/dhcpcsvc/dhcp/pipe.c
@@ -36,6 +36,243 @@ DWORD PipeSend( HANDLE CommPipe, COMM_DHCP_REPLY *Reply ) {
     return Success ? Written : -1;
 }
 
+/**
+ * @brief
+ * Creates a security descriptor for the DHCP pipe
+ * service.
+ *
+ * @param[out] SecurityDescriptor
+ * A pointer to an allocated security descriptor
+ * for the DHCP pipe.
+ *
+ * @return
+ * ERROR_SUCCESS is returned if the function has
+ * successfully created the descriptor otherwise
+ * a Win32 error code is returned.
+ *
+ * @remarks
+ * Both admins and local system are given full power
+ * over the DHCP pipe whereas authenticated users
+ * and network operators can only read over this pipe.
+ * They can also execute it.
+ */
+DWORD CreateDhcpPipeSecurity( PSECURITY_DESCRIPTOR *SecurityDescriptor ) {
+    DWORD ErrCode;
+    PACL Dacl;
+    ULONG DaclSize, RelSDSize = 0;
+    PSECURITY_DESCRIPTOR AbsSD = NULL, RelSD = NULL;
+    PSID AuthenticatedUsersSid = NULL, NetworkOpsSid = NULL, AdminsSid = NULL, SystemSid = NULL;
+    static SID_IDENTIFIER_AUTHORITY NtAuthority = {SECURITY_NT_AUTHORITY};
+
+    if (!AllocateAndInitializeSid(&NtAuthority,
+                                  1,
+                                  SECURITY_AUTHENTICATED_USER_RID,
+                                  0, 0, 0, 0, 0, 0, 0,
+                                  &AuthenticatedUsersSid))
+    {
+        DPRINT1("CreateDhcpPipeSecurity(): Failed to create Authenticated Users SID (error code %d)\n", GetLastError());
+        return GetLastError();
+    }
+
+    if (!AllocateAndInitializeSid(&NtAuthority,
+                                  2,
+                                  SECURITY_BUILTIN_DOMAIN_RID,
+                                  DOMAIN_ALIAS_RID_NETWORK_CONFIGURATION_OPS,
+                                  0, 0, 0, 0, 0, 0,
+                                  &NetworkOpsSid))
+    {
+        DPRINT1("CreateDhcpPipeSecurity(): Failed to create Network Ops SID (error code %d)\n", GetLastError());
+        ErrCode = GetLastError();
+        goto Quit;
+    }
+
+    if (!AllocateAndInitializeSid(&NtAuthority,
+                                  2,
+                                  SECURITY_BUILTIN_DOMAIN_RID,
+                                  DOMAIN_ALIAS_RID_ADMINS,
+                                  0, 0, 0, 0, 0, 0,
+                                  &AdminsSid))
+    {
+        DPRINT1("CreateDhcpPipeSecurity(): Failed to create Admins SID (error code %d)\n", GetLastError());
+        ErrCode = GetLastError();
+        goto Quit;
+    }
+
+    if (!AllocateAndInitializeSid(&NtAuthority,
+                                  1,
+                                  SECURITY_LOCAL_SYSTEM_RID,
+                                  0, 0, 0, 0, 0, 0, 0,
+                                  &SystemSid))
+    {
+        DPRINT1("CreateDhcpPipeSecurity(): Failed to create Local System SID (error code %d)\n", GetLastError());
+        ErrCode = GetLastError();
+        goto Quit;
+    }
+
+    AbsSD = HeapAlloc(GetProcessHeap(), HEAP_ZERO_MEMORY, sizeof(SECURITY_DESCRIPTOR));
+    if (!AbsSD)
+    {
+        DPRINT1("CreateDhcpPipeSecurity(): Failed to allocate absolute security descriptor!\n");
+        ErrCode = ERROR_OUTOFMEMORY;
+        goto Quit;
+    }
+
+    if (!InitializeSecurityDescriptor(AbsSD, SECURITY_DESCRIPTOR_REVISION))
+    {
+        DPRINT1("CreateDhcpPipeSecurity(): Failed to initialize absolute security descriptor (error code %d)\n", GetLastError());
+        ErrCode = GetLastError();
+        goto Quit;
+    }
+
+    DaclSize = sizeof(ACL) +
+               sizeof(ACCESS_ALLOWED_ACE) + GetLengthSid(AuthenticatedUsersSid) +
+               sizeof(ACCESS_ALLOWED_ACE) + GetLengthSid(NetworkOpsSid) +
+               sizeof(ACCESS_ALLOWED_ACE) + GetLengthSid(AdminsSid) +
+               sizeof(ACCESS_ALLOWED_ACE) + GetLengthSid(SystemSid);
+
+    Dacl = HeapAlloc(GetProcessHeap(), HEAP_ZERO_MEMORY, DaclSize);
+    if (!Dacl)
+    {
+        DPRINT1("CreateDhcpPipeSecurity(): Failed to allocate DACL!\n");
+        ErrCode = ERROR_OUTOFMEMORY;
+        goto Quit;
+    }
+
+    if (!InitializeAcl(Dacl, DaclSize, ACL_REVISION))
+    {
+        DPRINT1("CreateDhcpPipeSecurity(): Failed to initialize DACL (error code %d)\n", GetLastError());
+        ErrCode = GetLastError();
+        goto Quit;
+    }
+
+    if (!AddAccessAllowedAce(Dacl,
+                             ACL_REVISION,
+                             GENERIC_READ | GENERIC_EXECUTE,
+                             AuthenticatedUsersSid))
+    {
+        DPRINT1("CreateDhcpPipeSecurity(): Failed to set up ACE for Authenticated Users SID (error code %d)\n", GetLastError());
+        ErrCode = GetLastError();
+        goto Quit;
+    }
+
+    if (!AddAccessAllowedAce(Dacl,
+                             ACL_REVISION,
+                             GENERIC_READ | GENERIC_EXECUTE,
+                             NetworkOpsSid))
+    {
+        DPRINT1("CreateDhcpPipeSecurity(): Failed to set up ACE for Network Ops SID (error code %d)\n", GetLastError());
+        ErrCode = GetLastError();
+        goto Quit;
+    }
+
+    if (!AddAccessAllowedAce(Dacl,
+                             ACL_REVISION,
+                             GENERIC_ALL,
+                             AdminsSid))
+    {
+        DPRINT1("CreateDhcpPipeSecurity(): Failed to set up ACE for Admins SID (error code %d)\n", GetLastError());
+        ErrCode = GetLastError();
+        goto Quit;
+    }
+
+    if (!AddAccessAllowedAce(Dacl,
+                             ACL_REVISION,
+                             GENERIC_ALL,
+                             SystemSid))
+    {
+        DPRINT1("CreateDhcpPipeSecurity(): Failed to set up ACE for Local System SID (error code %d)\n", GetLastError());
+        ErrCode = GetLastError();
+        goto Quit;
+    }
+
+    if (!SetSecurityDescriptorDacl(AbsSD, TRUE, Dacl, FALSE))
+    {
+        DPRINT1("CreateDhcpPipeSecurity(): Failed to set up DACL to absolute security descriptor (error code %d)\n", GetLastError());
+        ErrCode = GetLastError();
+        goto Quit;
+    }
+
+    if (!SetSecurityDescriptorOwner(AbsSD, AdminsSid, FALSE))
+    {
+        DPRINT1("CreateDhcpPipeSecurity(): Failed to set up owner to absolute security descriptor (error code %d)\n", GetLastError());
+        ErrCode = GetLastError();
+        goto Quit;
+    }
+
+    if (!SetSecurityDescriptorGroup(AbsSD, SystemSid, FALSE))
+    {
+        DPRINT1("CreateDhcpPipeSecurity(): Failed to set up group to absolute security descriptor (error code %d)\n", GetLastError());
+        ErrCode = GetLastError();
+        goto Quit;
+    }
+
+    if (!MakeSelfRelativeSD(AbsSD, NULL, &RelSDSize) && GetLastError() != ERROR_INSUFFICIENT_BUFFER)
+    {
+        DPRINT1("CreateDhcpPipeSecurity(): Unexpected error code (error code %d -- must be ERROR_INSUFFICIENT_BUFFER)\n", GetLastError());
+        ErrCode = GetLastError();
+        goto Quit;
+    }
+
+    RelSD = HeapAlloc(GetProcessHeap(), HEAP_ZERO_MEMORY, RelSDSize);
+    if (RelSD == NULL)
+    {
+        DPRINT1("CreateDhcpPipeSecurity(): Failed to allocate relative SD!\n");
+        ErrCode = ERROR_OUTOFMEMORY;
+        goto Quit;
+    }
+
+    if (!MakeSelfRelativeSD(AbsSD, RelSD, &RelSDSize) && GetLastError() == ERROR_INSUFFICIENT_BUFFER)
+    {
+        DPRINT1("CreateDhcpPipeSecurity(): Failed to allocate relative SD, buffer too smal (expected size %lu)\n", RelSDSize);
+        ErrCode = ERROR_INSUFFICIENT_BUFFER;
+        goto Quit;
+    }
+
+    *SecurityDescriptor = RelSD;
+    ErrCode = ERROR_SUCCESS;
+
+Quit:
+    if (ErrCode != ERROR_SUCCESS)
+    {
+        if (RelSD)
+        {
+            HeapFree(GetProcessHeap(), 0, RelSD);
+        }
+    }
+
+    if (AuthenticatedUsersSid)
+    {
+        FreeSid(AuthenticatedUsersSid);
+    }
+
+    if (NetworkOpsSid)
+    {
+        FreeSid(NetworkOpsSid);
+    }
+
+    if (AdminsSid)
+    {
+        FreeSid(AdminsSid);
+    }
+
+    if (SystemSid)
+    {
+        FreeSid(SystemSid);
+    }
+
+    if (Dacl)
+    {
+        HeapFree(GetProcessHeap(), 0, Dacl);
+    }
+
+    if (AbsSD)
+    {
+        HeapFree(GetProcessHeap(), 0, AbsSD);
+    }
+
+    return ErrCode;
+}
+
 DWORD WINAPI PipeThreadProc( LPVOID Parameter ) {
     DWORD BytesRead;
     COMM_DHCP_REQ Req;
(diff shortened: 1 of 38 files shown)


GeoB99:
And yes in case it's not quite obvious for someone, the fact that ROS is now unbootable and lots of failed attempts to warrant rights occur is just a clear message to me that security rights in general have been neglected in ROS for years. Therefore...
warning

GeoB99:
**16/04/2022 UPDATE** -- ReactOS can now boot to the 1st stage setup (ROS literally couldn't boot to USETUP for installation at all!!!), it can also boot to desktop and you can interact with the system. However there are 3 things that currently don't work anymore:

- Shared folders won't show up on "My Computer" anymore, Vbox service fails to create certain Vbox internals files due to access check failures upon boot. You can however access such shared folders thorough "My Network Places".

- Network services can't open the TCP device object anymore. This is primarily due to LSASS. We currently lack code that handles services logged with special accounts such as NetworkService. Unfortunately there's nothing I can do with this.

-- CSRSS cannot set up SE_SHUTDOWN_PRIVILEGE (CsrSetToShutdownPriority fails) during shutdown. I have not seen any grave effects however, ReactOS still shutdowns normally.

These kind of issues have to live with us for a while since I cannot hold all this burden all by myself. ROS at least is in a tolerable state.

base/services/dhcpcsvc/dhcpcsvc.c
```diff
@@ -47,6 +47,8 @@ DhcpCApiInitialize(LPDWORD Version)
     if (PipeHandle == INVALID_HANDLE_VALUE)
     {
         /* We didn't */
+        HeapFree(GetProcessHeap(), 0, DhcpPipeSD);
+        DhcpPipeSD = NULL;
```
HBelusca, line 51: You're sure about these heapfree's here? This function is not the one that has managed the pipe creation and its security.

base/system/services/security.c
```diff
@@ -54,6 +54,7 @@ static
 DWORD
 ScmCreateSids(VOID)
 {
+    SID_IDENTIFIER_AUTHORITY WorldAuthority = {SECURITY_WORLD_SID_AUTHORITY};
     SID_IDENTIFIER_AUTHORITY NullAuthority = {SECURITY_NULL_SID_AUTHORITY};
```
HBelusca, line 58:
```suggestion
    SID_IDENTIFIER_AUTHORITY NullAuthority = {SECURITY_NULL_SID_AUTHORITY};
    SID_IDENTIFIER_AUTHORITY WorldAuthority = {SECURITY_WORLD_SID_AUTHORITY};
```
(sorted by increased SID values, see e.g. https://git.reactos.org/?p=reactos.git;a=blob;f=base/services/svchost/globals.c;hb=2f2795ab4cf7fdddf6c9d88d885a338a7678a888#l22 or the headers declarations)

base/system/services/services.c
```diff
@@ -107,6 +109,26 @@ CheckForLiveCD(VOID)
...
+                               L"SystemSetupInProgress",
+                               NULL,
+                               &dwType,
+                               (LPBYTE)&dwSetupInProgress,
+                               &dwSize);
+    if (dwError != ERROR_SUCCESS ||
+        dwType != REG_DWORD ||
+        dwSize != sizeof(DWORD) ||
+        dwSetupType == 0)
+        goto done;
```
HBelusca, line 124:
```suggestion
        dwSetupType == 0)
    {
        goto done;
    }
```

base/system/services/services.c
```diff
@@ -107,6 +109,26 @@ CheckForLiveCD(VOID)
...
+
+    if (dwSetupInProgress == 1)
```
HBelusca, line 126: what if it's non-zero but not equal to one either?

base/system/winlogon/winlogon.c
```diff
@@ -5,7 +5,7 @@
  * PURPOSE:         Logon
  * PROGRAMMERS:     Thomas Weidenmueller (w3seek@users.sourceforge.net)
  *                  Filip Navara
- *                  Herv� Poussineau (hpoussin@reactos.org)
+ *                  Hervé Poussineau (hpoussin@reactos.org)
```
HBelusca, line 8: untouched file otherwise?

dll/win32/kernel32/kernel32.spec
```diff
@@ -148,7 +148,7 @@
 @ stdcall CreateMutexW(ptr long wstr)
 @ stdcall CreateNamedPipeA(str long long long long long long ptr)
 @ stdcall CreateNamedPipeW(wstr long long long long long long ptr)
-@ stdcall -version=0x501-0x502 CreateNlsSecurityDescriptor(ptr long long)
```
HBelusca, line 151: Which Windows version have you checked? Because on 2003 the function has 3 parameters.

dll/win32/lsasrv/security.c
```diff
@@ -599,4 +599,159 @@ LsapCreateSecretSd(PSECURITY_DESCRIPTOR *SecretSd,
...
+    if (!NT_SUCCESS(Status))
+    {
+        if (RelativeSd != NULL)
+            RtlFreeHeap(RtlGetProcessHeap(), 0, RelativeSd);
+    }
+
+    return Status;
+}
+
+
```
HBelusca, line 756:
```suggestion
```

ntoskrnl/include/internal/se.h
```diff
@@ -38,6 +45,12 @@ typedef struct _TOKEN_AUDIT_POLICY_INFORMATION
 #define TOKEN_DUPLICATE_METHOD 0xDUL
 #define TOKEN_FILTER_METHOD    0xFUL
 
+typedef enum _ACCESS_CHECK_RIGHT_TYPE
```
HBelusca, line 48: Why not next to ACCESS_CHECK_RIGHTS declaration?

subsystems/win/basesrv/basesrv.h
```diff
@@ -58,9 +58,8 @@ typedef BOOL (WINAPI *PGET_NLS_SECTION_NAME)(UINT   CodePage,
                                              ULONG  ResultSize);
 
 typedef BOOL (WINAPI *PVALIDATE_LOCALE)(IN ULONG LocaleId);
-typedef NTSTATUS (WINAPI *PCREATE_NLS_SECURTY_DESCRIPTOR)(IN PVOID Buffer,
-                                                          IN ULONG BufferSize,
-                                                          IN ULONG AceType);
+typedef NTSTATUS (WINAPI *PCREATE_NLS_SECURTY_DESCRIPTOR)(_In_ ULONG AccessMask,
```
HBelusca, line 61: Which Windows version does this conform to?

win32ss/user/ntuser/security.h
```diff
@@ -0,0 +1,92 @@
...
+ */
+
+#pragma once
+
+//
+// USER objects security rights
+//
+
+/* Desktop access rights */
+#define DESKTOP_READ       STANDARD_RIGHTS_READ      | \
```
HBelusca, line 15: For correctness it would be better to have opening and closing parentheses around the whole definition block:
`#define FOO (BAR1 | ... | BARN)`

dll/win32/kernel32/winnls/string/nls.c
```diff
@@ -2237,13 +2272,339 @@ IsDBCSLeadByte(BYTE TestByte)
...
+ * that is given to the caller.
+ *
+ * @return
+ * STATUS_SUCCESS is returned if the function has successfully
+ * created a security descriptor for a NLS section name. Otherwise
+ * a NTSTATUS failure code is returned.
+ *
+ * @remarks
+ * The way we implement the function diverges with the Windows' one.
+ * On Windows Server 2003 this function takes 3 parameters. The first one
```
HBelusca, line 2470: And what is the problem of just keeping the 3 existing parameters? First one was your SecurityDescriptor, second one was your AccessMask, and you wouldn't need to do anything with the 3rd one.

base/system/services/services.c
GeoB99: `SystemSetupInProgress` can never be above 1, it's either 1 (== TRUE the setup is in progress) or 0 (== FALSE). The key value is set 0 when the setup is complete. See all the use cases of this value -> https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=SystemSetupInProgress

I see no point on adding unnecessary validity checks as it's just unwanted bikeshedding really.

GeoB99:
Test results: https://reactos.org/testman/compare.php?ids=82508,82510
KVM: https://build.reactos.org/#/builders/6/builds/31563
Vbox: https://build.reactos.org/#/builders/3/builds/19044

base/services/dhcpcsvc/dhcp/pipe.c
```diff
@@ -36,6 +38,245 @@ DWORD PipeSend( HANDLE CommPipe, COMM_DHCP_REPLY *Reply ) {
...
+
+    if (!InitializeSecurityDescriptor(AbsSD, SECURITY_DESCRIPTOR_REVISION))
+    {
+        DPRINT1("CreateDhcpPipeSecurity(): Failed to initialize absolute security descriptor (error code %d)\n", GetLastError());
+        ErrCode = GetLastError();
+        goto Quit;
+    }
+
+    DaclSize = sizeof(ACL) +
+               sizeof(ACCESS_ALLOWED_ACE) - sizeof(ULONG) + GetLengthSid(AuthenticatedUsersSid) +
```
HBelusca, line 132: Instead of subtracting the `sizeof(ULONG)`, you could instead do:
`sizeof(ACE) + GetLengthSid(...)` for each of these -- here and in the other parts of this PR.

base/services/dhcpcsvc/include/rosdhcp.h
```diff
@@ -78,6 +78,8 @@ typedef struct _DHCP_ADAPTER {
 
 typedef DWORD (*PipeSendFunc)(HANDLE CommPipe, COMM_DHCP_REPLY *Reply );
 
+extern PSECURITY_DESCRIPTOR DhcpPipeSD;
```
HBelusca, line 81: Should now be not necessary.

base/services/dhcpcsvc/dhcp/pipe.c
```diff
@@ -15,6 +15,8 @@
 #define COMM_PIPE_INPUT_BUFFER sizeof(COMM_DHCP_REPLY)
 #define COMM_PIPE_DEFAULT_TIMEOUT 1000
 
+PSECURITY_DESCRIPTOR DhcpPipeSD = NULL;
```
HBelusca, line 18: This could be moved within the PipeThreadProc function (as it is not used elsewhere).

base/services/dhcpcsvc/dhcp/pipe.c
```diff
@@ -56,10 +309,12 @@ DWORD WINAPI PipeThreadProc( LPVOID Parameter ) {
           COMM_PIPE_OUTPUT_BUFFER,
           COMM_PIPE_INPUT_BUFFER,
           COMM_PIPE_DEFAULT_TIMEOUT,
-          NULL );
+          &SecurityAttributes );
     if (CommPipe == INVALID_HANDLE_VALUE)
     {
         DbgPrint("DHCP: Could not create named pipe\n");
+        HeapFree(GetProcessHeap(), 0, DhcpPipeSD);
+        DhcpPipeSD = NULL;
```
HBelusca, line 317: I wonder whether it is possible to free the `DhcpPipeSD` at once just after the `CreateNamedPipeW` call (thus avoiding the repetitive HeapFree calls below).
In principle the kernel has captured the pipe's SD and doesn't need the info (on user-mode side) to stay around anymore.
But please test first whether this suggestion doesn't break anything.

base/services/dhcpcsvc/dhcp/pipe.c
```diff
@@ -36,6 +38,245 @@ DWORD PipeSend( HANDLE CommPipe, COMM_DHCP_REPLY *Reply ) {
     return Success ? Written : -1;
 }
 
+DWORD CreateDhcpPipeSecurity( PSECURITY_DESCRIPTOR *SecurityDescriptor ) {
```
HBelusca, line 41: I would suggest you add a documentation comment telling us what access rights are set for which users/groups.
(Suggestion being the same for all other such functions you've added in this PR.)

win32ss/user/ntuser/security.c
```diff
@@ -0,0 +1,493 @@
...
+                                     TokenUser,
+                                     UserToken,
+                                     BufferLength,
+                                     &BufferLength);
+    if (!NT_SUCCESS(Status))
+    {
+        /* We failed, bail out */
+        ERR("IntQueryUserSecurityIdentification(): Failed to query token user (Status 0x%08lx)\n", Status);
+        ZwClose(Token);
+        IntFreeSecurityBuffer(UserToken);
```
HBelusca, line 227: unwinding in reverse order of init:
```suggestion
        IntFreeSecurityBuffer(UserToken);
        ZwClose(Token);
```

ntoskrnl/se/sid.c
```diff
@@ -412,6 +412,77 @@ SepReleaseSid(
...
+    /* Initialize the SID */
+    Sid = NULL;
+
+    /* Obtain the SID based upon ACE type */
+    switch (AceType)
+    {
+        case ACCESS_DENIED_ACE_TYPE:
+        {
+            Sid = (PSID)&((PACCESS_DENIED_ACE)Ace)->SidStart;
+            break;
```
HBelusca, line 458: May be simplified by:
```suggestion
            return (PSID)&((PACCESS_DENIED_ACE)Ace)->SidStart;
```
etc. (for each cases; the default just doing `return NULL;`)

ntoskrnl/se/access.c
```diff
@@ -106,8 +102,15 @@ SepSidInTokenEx(
...
-                if ((!(i) && !(SidAndAttributes->Attributes & SE_GROUP_USE_FOR_DENY_ONLY)) ||
+                /*
+                 * Check if the group is enabled, or used for deny only.
+                 * Otherwise we have to check if this is the first user.
+                 * We understand that by looking if this SID is not
+                 * restricted, this is the first element we are iterating
+                 * and that it doesn't have SE_GROUP_USE_FOR_DENY_ONLY
+                 * attribute.
+                 */
+                if ((!Restricted && !(i) && !(SidAndAttributes->Attributes & SE_GROUP_USE_FOR_DENY_ONLY)) ||
```
HBelusca, line 113: `!(i)` --> `(i != 0)`

ntoskrnl/include/internal/se.h
```diff
@@ -24,6 +24,19 @@ typedef struct _KNOWN_COMPOUND_ACE
...
 } KNOWN_COMPOUND_ACE, *PKNOWN_COMPOUND_ACE;
 
+typedef struct _ACCESS_CHECK_RIGHTS
+{
+    ACCESS_MASK RemainingAccessRights;
+    ACCESS_MASK GrantedAccessRights;
+    ACCESS_MASK DeniedAccessRights;
+} ACCESS_CHECK_RIGHTS, *PACCESS_CHECK_RIGHTS;
+
+typedef enum _ACCESS_CHECK_RIGHT_TYPE
```
HBelusca, line 34: Are these a REACTOS-only addition, or do they also exist on some form in Windows as well?

base/system/winlogon/security.c
```diff
@@ -0,0 +1,1490 @@
...
+    /* Allow window station access to this user within this session */
+    if (!AllowWinstaAccessToUser(Session->InteractiveWindowStation, LogonSid))
+    {
+        ERR("AllowAccessOnSession(): Failed to allow winsta access to the logon user!\n");
+        goto Quit;
+    }
+
+/*
+ * FIXME: Desktop security management is broken. The application desktop gets created
+ * with CreateDesktopW() API call on wlx.file with an initial and defined security
```
HBelusca, line 1451: > on wlx.file

What do you mean there?

base/system/winlogon/security.c
```diff
@@ -0,0 +1,1490 @@
...
+ * descriptor for it yet when we are giving access to the logged in user, SetUserObjectSecurity()
+ * API call fails to set new security because the desktop in question has no prior security
```
HBelusca, line 1453: Do you know where the bug may reside?

base/services/dhcpcsvc/dhcp/pipe.c
```diff
@@ -36,6 +38,245 @@ DWORD PipeSend( HANDLE CommPipe, COMM_DHCP_REPLY *Reply ) {
...
+    PACL Dacl;
+    ULONG DaclSize, RelSDSize = 0;
+    PSECURITY_DESCRIPTOR AbsSD = NULL, RelSD = NULL;
+    PSID AuthenticatedUsersSid = NULL, NetworkOpsSid = NULL, AdminsSid = NULL, SystemSid = NULL;
+    static SID_IDENTIFIER_AUTHORITY NtAuthority = {SECURITY_NT_AUTHORITY};
+
+    if (!AllocateAndInitializeSid(&NtAuthority,
+                                  1,
+                                  SECURITY_AUTHENTICATED_USER_RID,
+                                  0,
```
HBelusca, line 52: You may fold all these zeroes on a single line, as it is usually done with `(Rtl)AllocateAndInitializeSid`.

dll/win32/rpcrt4/rpc_transport.c
```diff
@@ -107,23 +110,267 @@ static void release_np_event(RpcConnection_np *connection, HANDLE event)
...
                                         PIPE_TYPE_MESSAGE | PIPE_READMODE_MESSAGE,
                                         PIPE_UNLIMITED_INSTANCES,
                                         RPC_MAX_PACKET_SIZE, RPC_MAX_PACKET_SIZE, 5000, NULL);
+#endif
     if (connection->pipe == INVALID_HANDLE_VALUE)
     {
         WARN("CreateNamedPipe failed with error %d\n", GetLastError());
         if (GetLastError() == ERROR_FILE_EXISTS)
+        {
+#ifdef __REACTOS__
```
HBelusca, line 362: Same kind of comment for the security description freeing as I had for dhcpsvc.

dll/win32/kernel32/winnls/string/nls.c
```diff
@@ -2237,13 +2272,338 @@ IsDBCSLeadByte(BYTE TestByte)
...
+                                         &AdminsSid);
+    if (!NT_SUCCESS(Status))
+    {
+        DPRINT1("CreateNlsDirectorySecurity(): Failed to create admins SID (Status 0x%08x)\n", Status);
+        goto Quit;
+    }
+
+    /* Build up the size of our DACL, including the World and admins SIDs */
+    DaclSize = sizeof(ACL) +
+               (FIELD_OFFSET(ACCESS_ALLOWED_ACE, SidStart) + RtlLengthSid(WorldSid)) +
```
HBelusca, line 2328: What's being usually done is to do `sizeof(ACE) + RtlLengthSid(...)`, which is shorter and exact.

ntoskrnl/se/accesschk.c
```diff
@@ -922,17 +1369,20 @@ NtAccessCheck(
...
                        DesiredAccess,
                        NULL,
                        0,
                        PreviouslyGrantedAccess,
-                       &PrivilegeSet, //FIXME
                        GenericMapping,
                        PreviousMode,
+                       FALSE,
+                       FALSE,
+                       &PrivilegeSet,
```
HBelusca, line 1383: The whole ordeal with this PrivilegeSet pointer is partially broken in this function. (Missing capture of the buffer, etc etc.)

ntoskrnl/include/internal/se.h
GeoB99: These are specific ReactOS-only stuff, on my own personal way.

HBelusca: I was thinking about some `#ifdef _REACTOS_`... also the first member of the enum being "...Maximum" sounded unnatural, because usually these "Maximum" are always the last ones in an enum.

GeoB99: I'm not sure what you are talking about, that doesn't make sense. This is a private Se header so someone is free to do whatever they want so I dunno what you're trying to convey here.

base/services/dhcpcsvc/dhcp/pipe.c
GeoB99: You're correct, as once Se of the kernel has captured the descriptor and Ob assigning it to a securable object the information allocated in user mode is pretty much useless afterwards. As long as the allocated descriptor pointer in memory that you give to `CreateNamedPipeW` (and any other function in Win32 that accepts a security attribute with descriptor) is valid then the kernel happily captures it and you're good to go.

base/system/winlogon/security.c
GeoB99: No, I am not sure nor I am aware where's the root problem of this bug yet, I will look into this in the future when I can, for now this code path has to stay disabled (otherwise you ROS won't be able to log in anymore).

ntoskrnl/se/accesschk.c
GeoB99: This `PrivilegeSet` is completely redundant as `NtAccessCheck` already allocates a privilege set buffer by the `SePrivilegePolicyCheck` call anyway. I'm going to remove `PrivilegeSet`from there.

PR #4341, opened 2022-02-06
[REGEDIT] Fix issue at Find registry key

zecarlos1957:
Searching the registry for a key or value gives no results 

[REGEDIT] Fix issue at Find registry key
If we dont select a value to [pszValueName] as a starting point to search from,  we need set [pszValueName]  to the first item in current subkey

[[REGEDIT] Set focus to subkey when we search for it.

[[REGEDIT] Check pszSubKey length before calling RegFindRecurse.

### DIFF
--- a/base/applications/regedit/find.c
+++ b/base/applications/regedit/find.c
@@ -160,9 +160,6 @@ BOOL RegFindRecurse(
     if (lResult != ERROR_SUCCESS)
         return FALSE;
 
-    if (pszValueName == NULL)
-        pszValueName = s_empty;
-
     lResult = RegQueryInfoKeyW(hSubKey, NULL, NULL, NULL, NULL, NULL, NULL,
                               &c, NULL, NULL, NULL, NULL);
     if (lResult != ERROR_SUCCESS)
@@ -195,7 +192,7 @@ BOOL RegFindRecurse(
 
     qsort(ppszNames, c, sizeof(LPWSTR), compare);
 
-    if (pszValueName == s_empty)
+    if (pszValueName == NULL)
         pszValueName = ppszNames[0];
 
     for(i = 0; i < c; i++)
@@ -373,7 +370,8 @@ BOOL RegFindWalk(
     LPWSTR *ppszNames = NULL;
 
     hBaseKey = *phKey;
-    if (wcslen(pszSubKey) >= MAX_PATH)
+
+    if (wcslen(pszSubKey) >= _countof(szSubKey))
         return FALSE;
 
     if (RegFindRecurse(hBaseKey, pszSubKey, pszValueName, ppszFoundSubKey,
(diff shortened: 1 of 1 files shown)


zecarlos1957:
If we are looking for a non-existent string and press F3 key, this give a error: 
[The instruction at [addr] referenced memory at  [addr]. The memory could not be read]. 
This issue is not in RegEdit code. I tested our RegEdit.exe in WinXP and works fine with no errors.

base/applications/regedit/find.c
```diff
@@ -370,13 +373,13 @@ BOOL RegFindWalk(
     LPWSTR *ppszNames = NULL;
 
     hBaseKey = *phKey;
+    if (wcslen(pszSubKey) >= MAX_PATH)
```
HBelusca, line 376:
```suggestion

    if (wcslen(pszSubKey) >= _countof(szSubKey))
```

base/applications/regedit/find.c
```diff
@@ -687,10 +690,17 @@ BOOL FindNext(HWND hWnd)
...
-        SetValueName(g_pChildWnd->hListWnd, pszFoundValueName);
+
+        if (pszFoundValueName != NULL)
+        {
+            SetValueName(g_pChildWnd->hListWnd, pszFoundValueName);
+            free(pszFoundValueName);
+            SetFocus(g_pChildWnd->hListWnd);
+        }
+        else
+            SetFocus(g_pChildWnd->hTreeWnd);
```
HBelusca, line 701:
```suggestion
        else
        {
            SetFocus(g_pChildWnd->hTreeWnd);
        }
```

base/applications/regedit/find.c
```diff
@@ -687,10 +690,17 @@ BOOL FindNext(HWND hWnd)
     {
         GetKeyName(szFullKey, COUNT_OF(szFullKey), hKeyRoot, pszFoundSubKey);
         SelectNode(g_pChildWnd->hTreeWnd, szFullKey);
-        SetValueName(g_pChildWnd->hListWnd, pszFoundValueName);
+
```
HBelusca, line 693:
```suggestion
        free(pszFoundSubKey);

```

base/applications/regedit/find.c
```diff
@@ -687,10 +690,17 @@ BOOL FindNext(HWND hWnd)
...
+        if (pszFoundValueName != NULL)
+        {
+            SetValueName(g_pChildWnd->hListWnd, pszFoundValueName);
+            free(pszFoundValueName);
+            SetFocus(g_pChildWnd->hListWnd);
+        }
+        else
+            SetFocus(g_pChildWnd->hTreeWnd);
+
         free(pszFoundSubKey);
```
HBelusca, line 703:
```suggestion
```

base/applications/regedit/find.c
```diff
@@ -195,6 +195,9 @@ BOOL RegFindRecurse(
 
     qsort(ppszNames, c, sizeof(LPWSTR), compare);
 
+    if (pszValueName == s_empty)
```
HBelusca, line 198: It looks to me that you could even delete the previous `if (pszValueName == NULL) ...` case (that makes the pointer point to `s_empty`, and directly change here your new 'if'-case by `if (pszValueName == NULL)  pszValueName = ppszNames[0];`.

HBelusca (approved):

PR #4343, opened 2022-02-06
[REACTOS] Enforce 'Biosinfo' reg key name exact case

SergeGautherie:
Use 1 consistent case, instead of 3 different ones.
No functional change, yet helps reading/searching code.
https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=Biosinfo&sr=1

Noticed while looking into [CORE-18040](https://jira.reactos.org/browse/CORE-18040).


learn-more:
This solves no actual problem, and with the amount of open PR's that we have I see no reason to keep this open.

PR #2182, opened 2019-12-27
[FREELDR] Don't read a full FAT12/16 table into [7000:0000]

disean:
## Proposed changes

BEFORE:
```...```
```|-----------| 0000:F800```
```|freeldr.sys|```
```|-----------|```
```...```
```|-----------| 7000:000```
```| FAT table | (whole)```
```|-----------|```
```...```

AFTER:
```...```
```|-----------| 0000:XXXX```
```| FAT table | (2 sectors cached)```
```|-----------| 0000:F800```
```|freeldr.sys|```
```|-----------|```
```...```

`XXXX = 0xF800 - BytesPerSector * 2`

Ensuing consequences: This would affect disk performance because `sizeof(FAT12 table, 1.44MB FDD)` ≈ 9 sectors, `sizeof(FAT16 table, 490MB HDD)` ≈ 245 sectors.

JIRA issue: [CORE-15427](https://jira.reactos.org/browse/CORE-15427)

PR #4358, opened 2022-02-12
[FREELDR] Don't load a full FAT12/16 table when loading FREELDR.SYS.

HBelusca:
## Purpose

At least the way we currently do it and have our data layed out in
memory. Otherwise we risk corrupting either the FAT-loading code,
or the contents of FREELDR.SYS being loaded, when said FAT table is
quite large.

Supersedes PR #2182.

JIRA issues: [CORE-15427](https://jira.reactos.org/browse/CORE-15427), [CORE-14558](https://jira.reactos.org/browse/CORE-14558), [CORE-16195](https://jira.reactos.org/browse/CORE-16195), [CORE-13740](https://jira.reactos.org/browse/CORE-13740)

## Proposed changes

- Adapt the simple "cache" code from FAT32.
- But read two FAT sectors everytime in order to cover the case where one FAT12 entry data crosses sector boundary (since FAT12 entry data is 12 bytes, and therefore _**NOT**_ byte / 2-byte aligned).

## TODO

- [ ] Fully test this under Bochs+IDA Pro.
- [ ] Test the FAT16 modifications.


HBelusca:
Cc @disean too (I cannot add you to the list of reviewers.)

boot/freeldr/freeldr/arch/realmode/fathelp.inc
```diff
@@ -158,40 +170,101 @@ GetFatEntry16:
...
+/*
+ * This code is directly adapted from fat32.S
+ * BUT: Use DX:AX instead of EAX for the FAT logical sector number.
+ * CX: FAT entry offset.
+ */
+LoadFatSector:
+    push  cx                        // Save FAT entry offset
+
+//
+// FIXME: Why loading the sectors at 7000h do fail?
```
HBelusca, line 220: This is the first problem to understand.

boot/freeldr/freeldr/arch/realmode/fathelp.inc
```diff
@@ -158,40 +170,101 @@ GetFatEntry16:
...
+// In the meantime, do as with FAT32 and load the sectors
+// at 9000h instead; this seems to work fine...
+//
+    mov   bx, HEX(9000)             // We will load it to [9000:0000h]
+    mov   es, bx
+
+    // DX:AX holds logical FAT sector number
+    // Check if we have already loaded it
+//
+// FIXME: Why using cs:[...] does not work?
```
HBelusca, line 230: This is the second problem to understand and fix. I think that as this code currently stands, it even doesn't cache this data at the correct intended place either. I will have to run this through Bochs+IDA Pro to confirm my suppositions there...

boot/freeldr/freeldr/arch/realmode/fathelp.inc
```diff
@@ -158,40 +170,101 @@ GetFatEntry16:
...
+    cmp   ax, word ptr [bp+FatSectorInCache]
+    jne   DoLoad
+    cmp   dx, word ptr [bp+FatSectorInCache+2]
+    je    LoadFatSectorAlreadyLoaded
+
+DoLoad:
+    mov word ptr [bp+FatSectorInCache], ax
+    mov word ptr [bp+FatSectorInCache+2], dx
+    xor bx, bx
+    // Read 2 consecutive sectors, so that, in FAT12, entry data
```
disean, line 241: If an entry starts on the cache boundary you only neeed a `rep movsw // (move cached sector 2 to the cached sector 1)` and read a single sector to the cache.

boot/freeldr/freeldr/arch/realmode/fathelp.inc
```diff
@@ -1,6 +1,11 @@
-// fathelp.S
-// FAT12/16 Boot Sector Helper Code
-// Copyright (c) 1998, 2001, 2002, 2003 Brian Palmer
+/*
+ * PROJECT:     FreeLoader
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     FAT12/16 Boot Sector Helper Code
+ * COPYRIGHT:   Copyright 1998-2003 Brian Palmer (brianp@reactos.org)
+ * COPYRIGHT:   Copyright 2019 Dmitry Borisov (di.sean@protonmail.com)
```
HBelusca, line 6: Remark: twice "COPYRIGHT."

boot/freeldr/freeldr/arch/realmode/fathelp.inc
HBelusca: The FAT entry data being 12 bits, there is the possibility that the evaluated FAT sector offset points to the very last byte of the 1st cached sector, so 8 of the 12 bits will be there, but the last 4 bits will be in the next sector.

boot/freeldr/freeldr/arch/realmode/fathelp.inc
iLauncherDev: memory collision maybe?

HBelusca: > but this area is not fully used by the bios so it's not a problem

Well, the point is that you don't know that in advance, since this can depend on the BIOS / hardware. (And our freeldr looks at some EBDA data later on, so it's better not to corrupt it.)

boot/freeldr/freeldr/arch/realmode/fathelp.inc
iLauncherDev: You may want to check the bootsector functions

learn-more:

PR #4359, opened 2022-02-12
[FREELDR:PC98] Don't do the HighRes check everytime a character is written with writechr.

HBelusca:
## Purpose

See this PR's title. I suppose that this setting won't change between calls (when e.g. done from the other `writestr` helpers).

## Proposed changes

Do the HighRes check and cache the computed value of `VramSegment` only once, then reuse that latter one.

## TODO

- [x] Retest again once other blocker PR #4358 gets fixed.

### DIFF
--- a/boot/freeldr/freeldr/arch/realmode/helpers_pc98.inc
+++ b/boot/freeldr/freeldr/arch/realmode/helpers_pc98.inc
@@ -56,12 +56,12 @@ writestr:
     jmp short .writestr_loop
 
 .writestr_cr:
-    mov ax, word ptr VramOffset
+    mov ax, word ptr ds:[VramOffset]
     mov dl, 80 * 2
     div dl
     inc ax
     mul dl
-    mov word ptr VramOffset, ax
+    mov word ptr ds:[VramOffset], ax
 
     /* Skip the next LF character */
     inc si
@@ -80,26 +80,38 @@ writechr:
     pushf
     pusha
 
+    /* Check if the VRAM segment was initialized */
+    mov dx, word ptr ds:[VramSegment]
+    test dx, dx
+    jnz short .writechr_write
+
     /* High-resolution mode check */
     test byte ptr ds:[HEX(501)], HEX(08)
-    jz .writechr_normal
-    push HEX(0E000)
-    jmp short .writechr_test_done
-.writechr_normal:
-    push HEX(0A000)
-
+    mov dx, HEX(0A000) // Suppose normal mode
+    jz short .writechr_test_done
+    mov dh, HEX(E0)    // Change 0xA000 to 0xE000, use hi-res mode
 .writechr_test_done:
-    pop es
-    mov di, word ptr VramOffset
+    mov word ptr ds:[VramSegment], dx
+
+.writechr_write:
+    /* Load the VRAM segment and offset (ES:DI) */
+    les di, dword ptr ds:[VramSegOff]
+
+    /* Write ASCII directly to the VRAM (two bytes per character) */
     xor ah, ah
     stosw
-    mov word ptr VramOffset, di
+
+    /* Update the start position */
+    mov word ptr ds:[VramOffset], di
 
     popa
     popf
     ret
 
+VramSegOff:
 VramOffset:
+    .word 0
+VramSegment:
     .word 0
 
 /*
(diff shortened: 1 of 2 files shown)


HBelusca:
Cc @disean.

boot/freeldr/freeldr/arch/realmode/helpers_pc98.inc
```diff
@@ -80,15 +80,22 @@ writechr:
...
 .writechr_normal:
-    push HEX(0A000)
-
+    mov ax, HEX(0A000)
 .writechr_test_done:
+    mov word ptr VramSegment, ax
+
+.writechr_test_done2:
+    push ax
     pop es
```
disean, line 99: `mov es, ax`?

disean (approved):

binarymaster (approved):

boot/freeldr/freeldr/arch/realmode/helpers_pc98.inc
```diff
@@ -80,15 +80,22 @@ writechr:
     pushf
     pusha
 
+    mov ax, word ptr VramSegment
```
disean, line 83: There is a nice `LES` instruction you can use. This suggestion has been tested to work correctly:
```writechr:
    pushf
    pusha

    /* Check if the VRAM segment was initialized */
    mov dx, word ptr ds:[VramSegment]
    test dx, dx
    jnz .writechr_write

    /* High-resolution mode check */
    test byte ptr ds:[HEX(501)], HEX(08)
    mov dx, HEX(0A000)
    jz .writechr_test_done
    mov dh, HEX(0E)

.writechr_test_done:
    mov word ptr ds:[VramSegment], dx

.writechr_write:
    /* Load the VRAM segment and offset */
    les di, dword ptr ds:[VramOffset]

    /* Two bytes per character */
    xor ah, ah

    /* Write ASCII directly to the VRAM */
    stosw

    /* Update the start position */
    mov word ptr ds:[VramOffset], di

    popa
    popf
    ret

VramOffset:
    .word 0
VramSegment:
    .word 0

**Comment by disean** [comment]
Come to think of it, I wrote the hi-reso video support code with little knowledge of the PC-98 architecture, and I just realized that hi-reso models don't meet the minimum hardware requirements for ReactOS (https://www.pc-9800.net/db_98/dbh98_index.htm). ROS PC-98 targets PC-9821 hardware with PCI bus. I suggest to remove the VRAM check at all.

**Comment by binarymaster** [comment]
They all are non-Pentium? Machines without PCI should be possible (even in PC/AT case), though I don't rememeber if our Standard HAL could handle these properly.

**Comment by disean** [comment]
Yeah, they use processors older than Pentium.

File: boot/freeldr/freeldr/arch/realmode/helpers_pc98.inc (hunk 2 above)
Comment (HBelusca) [reply boot/freeldr/freeldr/arch/realmode/helpers_pc98.inc]: I've updated the code accordingly, please review :)

**Comment by HBelusca** [comment]
Hi @disean and @binarymaster , I've updated the PR, could you please retest it? (I've also updated the bootsector with some of the optimizations @disean suggested.)

File: boot/freeldr/freeldr/arch/realmode/helpers_pc98.inc [hunk 3]
```diff
@@ -80,27 +80,39 @@ writechr:
...
+
+    /* Update the start position */
+    mov word ptr ds:[VramOffset], di
 
     popa
     popf
     ret
 
+VramSegOff:
 VramOffset:
```
disean, line 112: Why is there two symbols in the first place?

disean (approved):

boot/freeldr/freeldr/arch/realmode/helpers_pc98.inc
HBelusca: The redundant `VramSegOff` is just to understand what's going to be read by the `les di, dword ptr ds:[VramSegOff]` instruction (mentioning only "VramOffset" in it instead could look weird at first).

boot/freeldr/freeldr/arch/realmode/helpers_pc98.inc
```diff
@@ -80,27 +80,39 @@ writechr:
...
     test byte ptr ds:[HEX(501)], HEX(08)
-    jz .writechr_normal
-    push HEX(0E000)
-    jmp short .writechr_test_done
-.writechr_normal:
-    push HEX(0A000)
-
+    mov dx, HEX(0A000) // Suppose normal mode
+    jz short .writechr_test_done
+    mov dh, HEX(E0)    // Change 0xA000 to 0xE000, use hi-res mode
```
HBelusca, line 92: @disean Note that here I fixed the value to be E0, instead of 0E as in your proposal, because it had to change the A0 to E0.

PR #4364, opened 2022-02-15
[NTOSKRNL] Implement JobObjectBasicProcessIdList

learn-more:


ntoskrnl/ps/job.c
```diff
@@ -707,6 +708,60 @@ NtQueryInformationJobObject (
...
+                    Status = STATUS_SUCCESS;
+
+                    for (NextEntry = Job->ProcessListHead.Flink;
+                        NextEntry != &Job->ProcessListHead;
+                        NextEntry = NextEntry->Flink)
+                    {
+                        if (Total == Job->ActiveProcesses)
+                        {
+                            DPRINT1("ERROR: Job->ActiveProcesses did not match Job->ProcessListHead!");
+                            Status = STATUS_PROCESS_IN_JOB;
```
learn-more, line 741: Not sure what to return here.

oleg-dubinskiy (commented):
Suggestion for better commit message:
`[NTOS:PS] Implement JobObjectBasicProcessIdList `

ntoskrnl/ps/job.c
GeoB99: I do not think you're actually supposed to return a status code in this kind of situation. Whenever a process' unique ID is not associated within a job, `NumberOfAssignedProcesses` field is decremented and the operation moves on until you have looped thorough all the present processes in the job.

With that said, `NumberOfAssignedProcesses` has to be set to active processes and when a process' ID is indeed associated with that job, you increment `NumberOfProcessIdsInList` and submit the process ID to the list.

learn-more: This `ActiveProcesses` count is relied upon earlier to ensure the buffer was big enough.
If there is a mismatch between this number and the LIST_ENTRY there is probably a serious problem?
Might even warrant an assert.

PR #4365, opened 2022-02-15
[SHELL32] CDefaultContextMenu: Provide a fallback property sheet

learn-more:
JIRA issue: [CORE-17691](https://jira.reactos.org/browse/CORE-17691)


yagoulas:
What is shown in the properties dialog with this?

learn-more:
> What is shown in the properties dialog with this?

image

yagoulas:
Huh nice. The way all pieces come together is absurdly complicated but it makes sense.

learn-more:
> Huh nice. The way all pieces come together is absurdly complicated but it makes sense.

I think there should be default handling for more cases, but let's not block this improvement on that and wait until we actually have a use case for it.

PR #4367, opened 2022-02-16
[WIN32SS] Store the scrollbar theming enabled flag in the scrollbar

learn-more:
This ensure we do not rely on usermode for always passing this flag in,
which is a ReactOS specific flag.
Thanks to Doug Lyons for finding the source of the problem.

JIRA issue: [CORE-17780](https://jira.reactos.org/browse/CORE-17780)

### DIFF
--- a/win32ss/user/ntuser/scrollbar.c
+++ b/win32ss/user/ntuser/scrollbar.c
@@ -528,6 +528,11 @@ co_IntSetScrollInfo(PWND Window, INT nBar, LPCSCROLLINFO lpsi, BOOL bRedraw)
    Info = IntGetScrollInfoFromWindow(Window, nBar);
    pSBData = IntGetSBData(Window, nBar);
 
+   if (lpsi->fMask & SIF_THEMED && !(Info->fMask & SIF_THEMED))
+   {
+       Info->fMask |= SIF_THEMED;
+   }
+
    /* Set the page size */
    if (lpsi->fMask & SIF_PAGE)
    {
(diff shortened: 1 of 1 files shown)


win32ss/user/ntuser/scrollbar.c
```diff
@@ -528,6 +528,11 @@ co_IntSetScrollInfo(PWND Window, INT nBar, LPCSCROLLINFO lpsi, BOOL bRedraw)
    Info = IntGetScrollInfoFromWindow(Window, nBar);
    pSBData = IntGetSBData(Window, nBar);
 
+   if (lpsi->fMask & SIF_THEMED && !(Info->fMask & SIF_THEMED))
```
HBelusca, line 531: It looks like the second condition is somehow redundant.

HBelusca (approved):

win32ss/user/ntuser/scrollbar.c
learn-more: It prevents setting the flag when it is already set?

JoachimHenze: Just a question: why is this flag not set for comctl32 treeview, but most likely is already set properly for comctl32 listview?
Wouldn't that actually show us the root-cause of the bug, instead of "fixing-up-later"?
I am just asking this in the hope, we would find something even more correct than your proposal.
Still it is by no means meant as a DoNotCommit, as your proposal is much better than all previous approaches.

learn-more: see jira.

JoachimHenze (approved):
Yes, that patch works well for me. Approved.

binarymaster (approved):
Suggesting `[NTUSER]` or `[WIN32SS:NTUSER]` as module prefix, other than that LGTM. :+1:

learn-more:
> Suggesting `[NTUSER]` or `[WIN32SS:NTUSER]` as module prefix, other than that LGTM. 👍

I kept the style as previous commits in this area.

JoachimHenze:
It turns out that this PR was just a first workaround, and we most likely can revert that in master It later got superseded by a better fix: https://github.com/reactos/reactos/commit/977c129f331cd5b92e0c7a4925e85fc14e147ec9

I will open a PR for reverting that first workaround at some point.

PR #4378, opened 2022-03-03
[NtUser] Fix Toolbars not being re-painted after his parent Rebar is altered.

julcar:
## Purpose

When altering the state of a Rebar window using a ShowWindow() or UpdateWindow() call, his children Toolbars are not re-painted due to a missing RDW_ALLCHILDREN flag to IntInvalidateWindows()

JIRA issue: [CORE-12342](https://jira.reactos.org/browse/CORE-12342)

## Proposed changes

Add the RDW_ALLCHILDREN flag to IntInvalidateWindows() call to force children Toolbar re-paint.


win32ss/user/ntuser/winpos.c
```diff
@@ -2100,8 +2100,15 @@ co_WinPosSetWindowPos(
...
                 {
-                   IntInvalidateWindows( Parent, DirtyRgn, RDW_ERASE | RDW_INVALIDATE);
-                   co_IntPaintWindows(Parent, RDW_NOCHILDREN, FALSE);
+                   /* HACK: For some reason, Rebar windows matches this condition after being altered
+                    * by ShowWindow() or UpdateWindow(), and as Rebar windows usually have children
+                    * Toolbars, these are not re-painted due to a missing RDW_ALLCHILDREN flag.
+                    * This might not have sense here, but fixes CORE-12342, CORE-15349, CORE-15917 and
+                    * very likely many others.
+                    */
+                   IntInvalidateWindows( Parent, DirtyRgn, RDW_ERASE | RDW_INVALIDATE | RDW_ALLCHILDREN);
```
JoachimHenze, line 2109: Please remove the space character before Parent, according to our coding-guidelines.

Then Please intentionally touch the 2nd IntInvalidateWindows() a few lines below in the else-branch the same way, so that it becomes visible in this review and will be visible also in a future git blame. To make it clearly visible that both code-paths will add RDW_ALLCHILDREN in the future!

And please readd in the comment that the change in Line2019 (adding RDW_ALLCHILDREN) does reduce the amount of executed tests of user32:msg_messages and user32:msg_winpos
like we had it clearly visible in
https://jira.reactos.org/secure/attachment/61682/61682_rebar_fixup4.patch

The future developers might be thankful for that information, because at some point in the future we will definitely have to touch that again, when something more correct was found.

JoachimHenze (changes requested):

win32ss/user/ntuser/winpos.c
```diff
@@ -2100,8 +2100,15 @@ co_WinPosSetWindowPos(
 
                 if ( (Window->style & WS_CHILD) && (Parent) && !(Parent->style & WS_CLIPCHILDREN))
                 {
-                   IntInvalidateWindows( Parent, DirtyRgn, RDW_ERASE | RDW_INVALIDATE);
-                   co_IntPaintWindows(Parent, RDW_NOCHILDREN, FALSE);
+                   /* HACK: For some reason, Rebar windows matches this condition after being altered
+                    * by ShowWindow() or UpdateWindow(), and as Rebar windows usually have children
+                    * Toolbars, these are not re-painted due to a missing RDW_ALLCHILDREN flag.
+                    * This might not have sense here, but fixes CORE-12342, CORE-15349, CORE-15917 and
+                    * very likely many others.
```
JoachimHenze, line 2107: strip the "and very likely many others" as that does not add any value at all. Even if it wasn't just pure speculation, it still wouldn't add any value.

win32ss/user/ntuser/winpos.c
```diff
@@ -2100,8 +2100,15 @@ co_WinPosSetWindowPos(
...
+                    */
+                   IntInvalidateWindows( Parent, DirtyRgn, RDW_ERASE | RDW_INVALIDATE | RDW_ALLCHILDREN);
+                   /* This call is kinda redundant, because the window already has QS_PAINT set by IntInvalidateWindows() */
+                   //co_IntPaintWindows(Parent, RDW_NOCHILDREN, FALSE);
```
JoachimHenze, line 2112: I would rather add that information to the commit comment than leaving this comment and the commented out call inside, but that is more a personal preference. Wait for others up-/downvotes first before acting on that one.

binarymaster (commented):
There are some commits in your branch which don't seem related to this PR:

image

JoachimHenze:
I closed this PR, because julenuri has a better version in the making already: https://jira.reactos.org/browse/CORE-12342?focusedCommentId=132865&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-132865

PR #4379, opened 2022-03-03
ReactOS Update

silvicky:
## Purpose
ReactOS Update.
It seems that nobody has tried yet, so I made this draft.
JIRA issue: [CORE-12934](https://jira.reactos.org/browse/CORE-12934)
## Proposed changes
Online update the system with a simple command.
Need an update server.
Just a draft now, can't really work.
No automatic update please!
## TODO
Finish the interface, add MD5 checking function, polish the code and a lot.

### DIFF
--- a/base/applications/rosupd/rosupd.cpp
+++ b/base/applications/rosupd/rosupd.cpp
@@ -1,0 +1,121 @@
+/*
+ * PROJECT:     ReactOS Update
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     To update or fix system files easily
+ * COPYRIGHT:   Copyright 2022 Vicky Silviana         (vickyisnotaboy@gmail.com)
+ */
+#include<stdio.h>
+#include<stdlib.h>
+#include<fstream>
+#include<iostream>
+#include<windows.h>
+#include<string.h>
+#include<string>
+#include<vector>
+#include<map>
+typedef HRESULT (WINAPI *UDProc) (LPUNKNOWN,LPCTSTR,LPCTSTR,DWORD,LPBINDSTATUSCALLBACK);
+using namespace std;
+//Remember to change the path!
+string sysdrv,sysdir;
+bool fmatch(string fname,char fmd5sum[16])
+{
+	ifstream fin;
+	fin.open((sysdir+"/"+fname).c_str(),ios_base::in);
+	fin.close();
+	return 1;
+}
+int main(int argc, char* argv[])
+{
+	int choicesum,filecnt;
+	ifstream fin;
+	string sname,vname,fpath;
+	map<string,int>mfile;
+	map<int,string>rmfile;
+	vector<char[16]>fmd5sum;
+	vector<bool>fpass;
+	HINSTANCE hurlmon;
+	UDProc murldownload;
+	sysdrv=getenv("SystemDrive");
+	sysdir=getenv("SystemRoot");
+	printf("ReactOS Update Version 0.1\n");
+	printf("Before updating your system, please make sure that ");
+	printf("you have enough free space on your system partition and a stable connection.\n");
+	printf("Are you sure you want to continue?");
+	while (choicesum=system("choice")-1)
+	{
+		if (choicesum==1)return 0;
+	}
+	printf("Stage 1: Reading source file...");
+	
+	fin.open((sysdir+"\\source.cfg").c_str(),ios_base::in);
+	fin>>sname>>vname;
+	fin.close();
+	if (sname.length()<=1)
+	{
+		printf("Failed!\n");
+		printf("Can\'t find source file!");
+		return 1;
+	}
+	printf("Done!\n");
+	printf("Stage 2: Downloading index file...");
+	system(("if exist "+sysdir+"\\index.cfg "
+			+"del /f /q "+sysdir+"\\index.cfg > nul").c_str());
+	hurlmon=LoadLibrary("urlmon.dll");
+	murldownload=(UDProc)GetProcAddress(hurlmon,"URLDownloadToFileA");
+	if (S_OK!=(murldownload)(NULL,
+							 (sname+vname+"/index.cfg").c_str(),
+							 (sysdir+"\\index.cfg").c_str(),
+							 0,
+							 NULL))
+	{
+		printf("Failed!\n");
+		printf("No internet or wrong source!");
+		FreeLibrary(hurlmon);
+		return 2;
+	}
+	printf("Done!\n");
+	printf("Stage 3: Comparing files...");
+	fin.open((sysdir+"\\index.cfg").c_str(),ios_base::in);
+	for (filecnt=0;;filecnt++)
+	{
+		fin>>rmfile[filecnt];
+		if (rmfile[filecnt].length()<=1)break;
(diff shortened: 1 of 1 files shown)


julenuri:
Actually, there is an experimental upgrade option in the first stage of the setup (that might work or might not). But this is interesting. Nice aim!

base/applications/rosupd/rosupd.cpp
```diff
@@ -0,0 +1,121 @@
...
+	string sname,vname,fpath;
+	map<string,int>mfile;
+	map<int,string>rmfile;
+	vector<char[16]>fmd5sum;
+	vector<bool>fpass;
+	HINSTANCE hurlmon;
+	UDProc murldownload;
+	sysdrv=getenv("SystemDrive");
+	sysdir=getenv("SystemRoot");
+	printf("ReactOS Update Version 0.1\n");
```
JoachimHenze, line 40: Why hardcoding a version here?

JoachimHenze (changes requested):
See some of my arbitrary review comments. I didn't put any love into that even.
But first and foremost: Why not simply using the existing update that we do have planned for usetup?

base/applications/rosupd/rosupd.cpp
```diff
@@ -0,0 +1,121 @@
...
+	}
+	printf("Done!\n");
+	printf("Stage 3: Comparing files...");
+	fin.open((sysdir+"\\index.cfg").c_str(),ios_base::in);
+	for (filecnt=0;;filecnt++)
+	{
+		fin>>rmfile[filecnt];
+		if (rmfile[filecnt].length()<=1)break;
+		mfile[rmfile[filecnt]]=filecnt;
+		for (int j=0;j<16;j++)
```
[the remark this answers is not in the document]
silvicky: This is online update, with which we can simply substitute different files.
Take it easy, it's just a draft!

katahiromz:
The system files can be locked. You should use `MoveFileEx` to avoid locking.

base/applications/rosupd/rosupd.cpp
```diff
@@ -0,0 +1,121 @@
...
+#include<fstream>
+#include<iostream>
+#include<windows.h>
+#include<string.h>
+#include<string>
+#include<vector>
+#include<map>
+typedef HRESULT (WINAPI *UDProc) (LPUNKNOWN,LPCTSTR,LPCTSTR,DWORD,LPBINDSTATUSCALLBACK);
+using namespace std;
+//Remember to change the path!
```
katahiromz, line 18: Don't omit std.

base/applications/rosupd/rosupd.cpp
```diff
@@ -0,0 +1,121 @@
+/*
+ * PROJECT:     ReactOS Update
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     To update or fix system files easily
+ * COPYRIGHT:   Copyright 2022 Vicky Silviana         (vickyisnotaboy@gmail.com)
+ */
+#include<stdio.h>
+#include<stdlib.h>
```
katahiromz, line 8:
```suggestion
#include <stdlib.h>
```

base/applications/rosupd/rosupd.cpp
```diff
@@ -0,0 +1,121 @@
...
+#include<fstream>
+#include<iostream>
+#include<windows.h>
+#include<string.h>
+#include<string>
+#include<vector>
+#include<map>
+typedef HRESULT (WINAPI *UDProc) (LPUNKNOWN,LPCTSTR,LPCTSTR,DWORD,LPBINDSTATUSCALLBACK);
```
katahiromz, line 16: Don't omit function name.

base/applications/rosupd/rosupd.cpp
```diff
@@ -0,0 +1,121 @@
...
+	ifstream fin;
+	fin.open((sysdir+"/"+fname).c_str(),ios_base::in);
+	fin.close();
+	return 1;
+}
+int main(int argc, char* argv[])
+{
+	int choicesum,filecnt;
+	ifstream fin;
+	string sname,vname,fpath;
```
katahiromz, line 31: Insert space after comma.

base/applications/rosupd/rosupd.cpp
```diff
@@ -0,0 +1,121 @@
...
+	printf("you have enough free space on your system partition and a stable connection.\n");
+	printf("Are you sure you want to continue?");
+	while (choicesum=system("choice")-1)
+	{
+		if (choicesum==1)return 0;
+	}
+	printf("Stage 1: Reading source file...");
+	
+	fin.open((sysdir+"\\source.cfg").c_str(),ios_base::in);
+	fin>>sname>>vname;
```
katahiromz, line 51:
```suggestion
	fin >> sname >> vname;
```

base/applications/rosupd/rosupd.cpp
```diff
@@ -0,0 +1,121 @@
...
+									 NULL))
+						   				{
+						   					printf("Succeed!\n");
+						   					fpass[i]=1;
+						   				}
+			else printf("Failed!\n");
+		}
+	}
+	printf("Done!\n");
+//	printf("Stage 5: Changing freeldr.ini...");
```
JoachimHenze, line 118: Why changing freeldr.ini and not using pendmoves maybe to replace files upon rebooting?
Would that introduce any limitations? E.g. for replacing the ntoskrnl itself?

base/applications/rosupd/rosupd.cpp
```diff
@@ -0,0 +1,121 @@
...
+	}
+	printf("Stage 1: Reading source file...");
+	
+	fin.open((sysdir+"\\source.cfg").c_str(),ios_base::in);
+	fin>>sname>>vname;
+	fin.close();
+	if (sname.length()<=1)
+	{
+		printf("Failed!\n");
+		printf("Can\'t find source file!");
```
JoachimHenze, line 56: is that backslash for escaping the ' really needed?

base/applications/rosupd/rosupd.cpp
```diff
@@ -0,0 +1,121 @@
...
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     To update or fix system files easily
+ * COPYRIGHT:   Copyright 2022 Vicky Silviana         (vickyisnotaboy@gmail.com)
+ */
+#include<stdio.h>
+#include<stdlib.h>
+#include<fstream>
+#include<iostream>
+#include<windows.h>
+#include<string.h>
```
JoachimHenze, line 12: what is string.h used for here?

silvicky:
I'm really sorry for the bad code, it's my first time to make a real program outside algorithm contests.
Will make it better tomorrow.

HBelusca:
I think that first, someone (e.g. you?) has to explicitly write down what you expect the updater to do and how it is supposed to work.

learn-more:
Before this is nitpicked to death (again) by @HBelusca and @JoachimHenze , it makes sense to discuss this particular subject.
In my opinion, at this point in time an application like this has no added value.
- There is not a 'stable' version to start with
- No-one is going to maintain lists of files that can be 'downloaded'
- In-place upgrade means each version of each downloadable component needs to be tested against each version of each other component.
-- Considering we hardly even test master, this is simply not going to happen

PR #4386, opened 2022-03-10
[SHELL32] Copy without FILE_ATTRIBUTE_READONLY from a CDROM

learn-more:
JIRA issue: [CORE-18089](https://jira.reactos.org/browse/CORE-18089)

### DIFF
--- a/dll/win32/shell32/shlfileop.cpp
+++ b/dll/win32/shell32/shlfileop.cpp
@@ -755,6 +755,21 @@ static DWORD SHNotifyMoveFileW(FILE_OPERATION *op, LPCWSTR src, LPCWSTR dest, BO
 #endif
 }
 
+static BOOL SHIsCdRom(LPCWSTR path)
+{
+    WCHAR tmp[] = { L"A:\\" };
+
+    if (!path || !path[0])
+        return FALSE;
+
+    if (path[1] != UNICODE_NULL && path[1] != ':')
+        return FALSE;
+
+    tmp[0] = path[0];
+
+    return GetDriveTypeW(tmp) == DRIVE_CDROM;
+}
+
 /************************************************************************
  * SHNotifyCopyFile          [internal]
  *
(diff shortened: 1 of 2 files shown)


dll/win32/shell32/shlfileop.cpp
```diff
@@ -755,6 +755,21 @@ static DWORD SHNotifyMoveFileW(FILE_OPERATION *op, LPCWSTR src, LPCWSTR dest, BO
 #endif
 }
 
+static BOOL SHIsCdRom(LPCWSTR path)
+{
+    WCHAR tmp[] = { 'A',':','\\',0 };
```
HBelusca, line 760: you may be able to get rid of this wine-style string?

dll/win32/shell32/shlfileop.cpp
```diff
@@ -795,6 +810,14 @@ static DWORD SHNotifyCopyFileW(FILE_OPERATION *op, LPCWSTR src, LPCWSTR dest, BO
     ret = CopyFileExW(src, dest, SHCopyProgressRoutine, op, &op->bCancelled, bFailIfExists);
     if (ret)
     {
+        // We are copying from a CD-ROM volume, which is readonly
+        if (SHIsCdRom(src))
```
HBelusca, line 814: If a ton of files are moved from the same origin (CD-ROM), is this function going to be called such a ton of times?

learn-more: yes, but we cannot just call this on the first file, because it's possible to copy files from multiple locations

dll/win32/shell32/shlfileop.cpp
learn-more: sure

dll/win32/shell32/shlfileop.cpp
HBelusca: ok if the shell doesn't group the file operations also by source folder/device, then indeed I don't see how we can improve that.

JoachimHenze (approved):

PR #4389, opened 2022-03-12
[SYSSETUP] Add new Env section to unattend.inf

learn-more:
This allows to add environment variables during unattended setup

### DIFF
--- a/dll/win32/syssetup/wizard.c
+++ b/dll/win32/syssetup/wizard.c
@@ -2831,6 +2831,53 @@ ProcessUnattendSection(
     }
 
     RegCloseKey(hKey);
+
+    if (SetupFindFirstLineW(pSetupData->hSetupInf,
+        L"Env",
+        NULL,
+        &InfContext))
+    {
+        if (RegCreateKeyExW(
+                HKEY_LOCAL_MACHINE, L"SYSTEM\\CurrentControlSet\\Control\\Session Manager\\Environment", 0, NULL,
+                REG_OPTION_NON_VOLATILE, KEY_WRITE | KEY_READ, NULL, &hKey, NULL) != ERROR_SUCCESS)
+        {
+            DPRINT1("Error: failed to open HKLM\\SYSTEM\\CurrentControlSet\\Control\\Session Manager\\Environment\n");
+            return;
+        }
+        do
+        {
+            if (!SetupGetStringFieldW(&InfContext,
+                0,
+                szName,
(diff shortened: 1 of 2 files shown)


dll/win32/syssetup/wizard.c
```diff
@@ -2831,6 +2831,53 @@ ProcessUnattendSection(
     }
 
     RegCloseKey(hKey);
+
+    if (SetupFindFirstLineW(pSetupData->hSetupInf,
+        L"Env",
+        NULL,
+        &InfContext))
+    {
+        if (RegCreateKeyExW(
```
HBelusca, line 2840: That registry key should already exist; if it doesn't then it should be an error.

dll/win32/syssetup/wizard.c
```diff
@@ -2831,6 +2831,53 @@ ProcessUnattendSection(
     }
 
     RegCloseKey(hKey);
+
+    if (SetupFindFirstLineW(pSetupData->hSetupInf,
```
HBelusca, line 2835: Indent seems inconsistent wrt. the rest of the code in this file.

HBelusca (commented):
Interesting feature; I wonder how one does the same with Windows.

dll/win32/syssetup/wizard.c
learn-more: Indeed, but I don't feel like manually spacing that stuff out

dll/win32/syssetup/wizard.c
learn-more: The only reason why I would see this fail is when the key cannot be opened for some reason?

learn-more:
@HBelusca do you want to investigate if there is a more windows-like way to do this, or shall I go ahead and commit this?

HBelusca:
> @HBelusca do you want to investigate if there is a more windows-like way to do this, or shall I go ahead and commit this?

I did search around a bit to find whether there was a windows-like way to do this, but I haven't found succeeded to find anything.
So I would say, go ahead and commit the PR, and in the future, if I or someone else is able to find whether Windows has a better way of doing this, one could always fix the code.

(Personally I don't think they have one: unattend.inf is just a *response* file, i.e. to automate giving the answers for the installer, and not a file that contains new settings; otherwise one would have to edit the registry template hive files, for example this seems to be what happens for e.g. WinPE or Windows installation ISOs when they are customized).

PR #4390, opened 2022-03-13
[NTGDI] Check if face->charmap is not zero before accessing its encoding

julcar:
## Purpose

Attempting to start DevExpress Ribbon Notepad makes crash ReactOS due to null pointer access, after some investigation, I found that `(face->charmap)` might return 0 sometimes, and we are not making this check.

This allows DevExpress Ribbon Notepad sample to start.

JIRA issue: [CORE-18091](https://jira.reactos.org/browse/CORE-18091) [CORE-18558](https://jira.reactos.org/browse/CORE-18558)

### DIFF
--- a/win32ss/gdi/ntgdi/freetype.c
+++ b/win32ss/gdi/ntgdi/freetype.c
@@ -4501,6 +4501,8 @@ ftGetFontUnicodeRanges(PFONTGDI Font, PGLYPHSET glyphset)
     DWORD num_ranges = 0;
     FT_Face face = Font->SharedFace->Face;
 
+    if (!(face->charmap)) return 0;
+
     if (face->charmap->encoding == FT_ENCODING_UNICODE)
     {
         FT_UInt glyph_code = 0;
--- a/win32ss/gdi/ntgdi/freetype.c
+++ b/win32ss/gdi/ntgdi/freetype.c
@@ -4501,7 +4501,11 @@ ftGetFontUnicodeRanges(PFONTGDI Font, PGLYPHSET glyphset)
     DWORD num_ranges = 0;
     FT_Face face = Font->SharedFace->Face;
 
-    if (!(face->charmap)) return 0;
+    if (face->charmap == NULL)
+    {
+        DPRINT1("FIXME: No charmap selected! This is a BUG!\n");
+        return 0;
+    }
 
     if (face->charmap->encoding == FT_ENCODING_UNICODE)
     {


learn-more:
There are a few places where it's checked for NULL already,
but all these places also assign a charmap in that case.
Sadly, there are some changes between how they manage that, but I suspect that they should all do the same.
(And that it should happen in all cases before `->charmap` is accessed.

https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&sr=1&s=charmap+%3D%3D+NULL

vicmarcal:
> There are a few places where it's checked for NULL already, but all these places also assign a charmap in that case. Sadly, there are some changes between how they manage that, but I suspect that they should all do the same. (And that it should happen in all cases before `->charmap` is accessed.
> 
> https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&sr=1&s=charmap+%3D%3D+NULL

Mmm comparing all the uses, seems the first result in that link is following a different pattern (encoding checking) to assign a charmap  compared with the other 2 (which are blindly assigning the first one found). Any reasoning behind that? Just curious.

katahiromz (approved):

KRosUser:
Does it fixes the refered Jira ticket ? Or is it a side fix ?

julcar:
> Does it fixes the refered Jira ticket ? Or is it a side fix ?

It is a real fix, the program affected started to run after applying the fix

KRosUser:
So good to merge since July, right ?

KRosUser2:
@julcar : Still OK to go with this PR ?

learn-more (approved):

KRosUser2 (approved):

KRosUser2:
Please note it also fixes https://jira.reactos.org/browse/CORE-18558

win32ss/gdi/ntgdi/freetype.c
```diff
@@ -4501,6 +4501,8 @@ ftGetFontUnicodeRanges(PFONTGDI Font, PGLYPHSET glyphset)
     DWORD num_ranges = 0;
     FT_Face face = Font->SharedFace->Face;
 
+    if (!(face->charmap)) return 0;
```
tkreuzer, line 4504:
```suggestion
    if (face->charmap == NULL)
    {
        DPRINT1("FIXME: No charmap selected! This is a BUG!\n");
        return 0;
    }
```

tkreuzer (changes requested):
This is at best a band-aid. There is a lot wrong in the code. 

- With this fix ftGetFontUnicodeRanges would simply fail, instead of selecting the proper charmap.
- In some places the code alls FT_Set_Charmap, first acquiring, then instantly releasing the freetype lock. This is wrong. Either we set the charset only once and keep it immutable in the rest of the code, or all code that relies on it must run with the freetype lock help.
- Code duplication.

I'm ok with a quick-fix, but put at least a fat DPRINT1 there.

JoachimHenze (approved):

tkreuzer (approved):

PR #4393, opened 2022-03-14
[MSI] Winesync up to WineStaging-7.3

learn-more:
- It's probably easiest to compare this against current wine.
- Please note that `sdk\include\reactos\wine\test.h` is also changed.
- `sdk\include\crt\{stdio.h|wchar.h}` are changed (and the root `CMakeLists.txt`)


JoachimHenze:
did you run the testbots against that yet? Can you post the result comparison?

learn-more:
It's running:
 https://build.reactos.org/#/builders/9/builds/31436

modules/rostests/apitests/compiler/pseh.c
```diff
@@ -27,6 +27,7 @@ extern "C" {
 #endif
 
 #include <wine/test.h>
+#undef subtest
```
HBelusca, line 30: "subtest" is used in this file? o_O

dll/win32/msi/msi.spec
```diff
@@ -294,3 +294,5 @@
 @ stdcall -private DllGetClassObject(ptr ptr ptr)
 @ stdcall -private DllRegisterServer()
 @ stdcall -private DllUnregisterServer()
+
+@ cdecl __wine_msi_call_dll_function(long ptr)
```
HBelusca, line 298: Hmmm.... do we accept these kind of things? Is this function necessary? for what? If not, could we remove it?

JoachimHenze:
Are the increased failure counts in msi:action and msi:package all failing in newly added tests or are there also new failures in the formerly working tests? (Same question in different words: Would we have no additional failing tests if we *theoretically* just updated msi.dll but did not touch the msi winetests at all?)

dll/win32/msi/msi.spec
learn-more:
```
static DWORD CALLBACK custom_action_thread(void *arg)
{
    GUID guid = *(GUID *)arg;
    heap_free(arg);
    return __wine_msi_call_dll_function(client_pid, &guid);
}
```

dll/win32/msi/CMakeLists.txt
```diff
@@ -71,16 +80,17 @@ if(MSVC)
...
 
-add_idl_headers(msi_idlheader msiserver.idl)
-add_typelib(msiserver.idl)
-set_source_files_properties(msi.rc PROPERTIES OBJECT_DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/msiserver.tlb)
+add_idl_headers(msi_idlheader msiserver.idl winemsi.idl)
+add_typelib(msiserver.idl winemsi.idl)
+set_source_files_properties(msi.rc PROPERTIES OBJECT_DEPENDS "${CMAKE_CURRENT_BINARY_DIR}/msiserver.tlb;${CMAKE_CURRENT_BINARY_DIR}/winemsi.tlb")
 set_module_type(msi win32dll)
-target_link_libraries(msi uuid ${PSEH_LIB} wine)
+target_link_libraries(msi uuid ${PSEH_LIB} wine_dll_register oldnames)
```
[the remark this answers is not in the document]
learn-more: the non-underscore prefixed functions:
https://github.com/reactos/reactos/commit/64c3e59f62c238eda0425370479f56ad38cf99dd

dll/win32/msi/msi.spec
HBelusca: OK so basically this is (yet another example of) a piece of Wine hackish code... 😒😒

learn-more: yup

learn-more: at least they are starting to share code between modules now, instead of copy-pasting it every time...

learn-more:
The increase in `msi:action` test failures is due to the same bugs being checked differently.
All these failures come from the `test_envvar` part.

binarymaster:
This brought a regresso :coffee: https://jira.reactos.org/browse/CORE-18707

PR #4395, opened 2022-03-15
[RAPPS] Fix uninstalling applications

learn-more:
- 'WindowsInstaller' setups do not fill the UninstallString and ModifyPath with the needed info, so manually calculate those.
- Fix the case where REG_EXPAND_SZ is used instead of REG_SZ
- Take NoModify into account

### DIFF
--- a/base/applications/rapps/installed.cpp
+++ b/base/applications/rapps/installed.cpp
@@ -71,8 +71,40 @@ void CInstalledApplicationInfo::EnsureDetailsLoaded()
         }
         GetApplicationRegString(L"InstallLocation", szInstallLocation);
         GetApplicationRegString(L"InstallSource", szInstallSource);
-        GetApplicationRegString(L"UninstallString", szUninstallString);
-        GetApplicationRegString(L"ModifyPath",szModifyPath);
+        DWORD dwWindowsInstaller = 0;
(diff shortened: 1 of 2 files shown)


base/applications/rapps/installed.cpp
```diff
@@ -71,8 +71,40 @@ void CInstalledApplicationInfo::EnsureDetailsLoaded()
...
+        {
+            GetApplicationRegString(L"UninstallString", szUninstallString);
+        }
+        DWORD dwNoModify = 0;
+        if (!GetApplicationRegDword(L"NoModify", &dwNoModify))
+        {
+            CStringW Tmp;
+            if (GetApplicationRegString(L"NoModify", Tmp))
+            {
+                dwNoModify = Tmp.GetLength() > 0 ? (Tmp[0] == '1') : 0;
```
HBelusca, line 90: What if the string is "2" or something else? If it's considered to be like a BOOL variable, usually it's FALSE when it's zero, and TRUE when it's anything else.

HBelusca (approved):

base/applications/rapps/installed.cpp
learn-more: This is already a fallback, so let's not write a fallback on top of a fallback :)

PR #4272, opened 2022-01-11
[FREELDR] Add support for Physical Address Extension (PAE) paging mode memory management.

vgalnt:
## Purpose

[FREELDR] Add support for Physical Address Extension (PAE) paging mode memory management.

JIRA issue: [CORE-17986](https://jira.reactos.org/browse/CORE-17986)

## Proposed changes

- Added "LiveCD PAE (Debug)" in livecd.ini.
- Changes have been made to "ndk/i386/mmtypes.h" so that PAE\Non-PAE structures are visible at the same time.
- Added defines for PAE in "boot/freeldr/freeldr/include/arch/i386/i386.h"
- Added WinLdrIsPaeSupported() - a stub for checking compatibility and enabling PAE mode.
- Add support for PAE in "boot/freeldr/freeldr/ntldr/arch/i386/winldr.c".

PR #4398, opened 2022-03-20
[FREELDR] Add support for Physical Address Extension (PAE) paging mode memory management.

HBelusca:
### _**This is a Rebasing of PR #4272 on top of master, that I cannot update in-place due to repository rights.**_

Cc @vgalnt

### DIFF
--- a/boot/bootdata/livecd.ini
+++ b/boot/bootdata/livecd.ini
@@ -9,6 +9,7 @@ MinimalUI=Yes
 [Operating Systems]
 LiveCD="LiveCD"
 LiveCD_Debug="LiveCD (Debug)"
+LiveCD_PAE_Debug="LiveCD PAE (Debug)"
 LiveCD_Aacpi="LiveCD ACPI APIC (Debug)"
 LiveCD_VBoxDebug="LiveCD (VBox Debug)"
 LiveCD_Screen="LiveCD (Screen)"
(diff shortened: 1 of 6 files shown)


binarymaster (commented):
amd64 build not happy:
```
../src/boot/freeldr/freeldr/ntldr/winldr.c:1242:9: error: use of undeclared identifier 'PaeModeOn'
```

tkreuzer:
Why does freeldr need to support PAE? IMO this is something only the kernel should care about.

DotNetTester:
In Windows, whether PAE is enabled or not is determined by the hidden "boot.ini" file, the "/pae" parameter in Windows 2000. I *believe* the freeldr plays a similar role to boot.ini in ReactOS.

tkreuzer:
Whether the system is PAE or not depends on what kernel is being loaded (based on boot.ini / freeldr.ini). A PAE bootloader makes no sense. If you want PAE, make the kernel handle PAE (Mm needs to build a new PAE page directory from the freeldr non-PAE mappings.

HBelusca:
> Whether the system is PAE or not depends on what kernel is being loaded (based on boot.ini / freeldr.ini). A PAE bootloader makes no sense. If you want PAE, make the kernel handle PAE (Mm needs to build a new PAE page directory from the freeldr non-PAE mappings.

Perhaps. But if you are then unable to load a windows 2003 PAE kernel from our freeldr (or, vice versa, Windows' NTLDR cannot load our ROS PAE kernel), whereas both can load each other's non-PAE kernels, then we have to ask ourselves whether indeed the freeldr still needs to do some PAE preparation.

tkreuzer:
>

PR #4402, opened 2022-03-22
Get ARM64 compiling further and start ke.h

DarkFire01:
## Purpose
Get closer to building ARM64 applications 

JIRA issue: [CORE-17518](https://jira.reactos.org/browse/CORE-17518)

## Proposed changes

-  Add in xdk/arm64/ke.h
- Add most of the missing one line definitions
- Make memset_fallback build for arm64 as well.


sdk/include/xdk/ntddk.template.h
```diff
@@ -118,6 +118,8 @@ typedef PIMAGE_NT_HEADERS32 PIMAGE_NT_HEADERS;
 #elif defined(_M_ARM)
 $include(arm/ke.h)
 $include(arm/mm.h)
+#elif defined(_M_ARM64)
+$include(arm/ke.h)
```
binarymaster: There should be `$include(arm64/ke.h)` - you have fixed it in the second commit, however the fix should be in this first commit, otherwise it wouldn't be correct.

PR #4403, opened 2022-03-22
Control Panel applets should be started in single instance

RaymondCzerny:
## Purpose

In MS Windows all control panel applets are single instance applications. This prevents the collision of multiple accesses to the configuration data.

## Proposed changes

In the initialization phase of the applets, a preinstance is now searched for after the dialog window. If the search is successful, this dialog is brought to the foreground and the second instance is aborted.

### DIFF
--- a/dll/cpl/access/CMakeLists.txt
+++ b/dll/cpl/access/CMakeLists.txt
@@ -8,7 +8,9 @@ list(APPEND SOURCE
     keyboard.c
     mouse.c
     sound.c
-    access.h)
+    access.h
+    ../common/cpl_common.c
+)
 
 add_rc_deps(access.rc ${CMAKE_CURRENT_SOURCE_DIR}/resources/applet.ico)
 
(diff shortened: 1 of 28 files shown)


HBelusca (changes requested):
This behaviour is controlled not by the individual applets, but by the shell itself.
You could verify this by observing that the original applets (without your changes) were working in "single-instance" mode when you try them on Windows (by e.g. correctly installing one of these as an actual CPL with an entry within the control panel folder).

RaymondCzerny:
Okay, I get it.

My new suggestion:
Step 1: Discard the branch 'cpl'.
Step 2: Instance check in the function 'Control_DoLaunch' in 
reactos/dll/win32/shell32/wine/control.c'.

What do you think about this procedure?

HBelusca:
> Okay, I get it.
> 
> My new suggestion: Step 1: Discard the branch 'cpl'. Step 2: Instance check in the function 'Control_DoLaunch' in reactos/dll/win32/shell32/wine/control.c'.
> 
> What do you think about this procedure?

That would do the trick certainly. Please check that Wine doesn't have code for that already (if not, just go ahead and do the implementation). You could keep your cpl branch, but implement the new functionality and remove the old commits with git (and then force-push to the corresponding remote branch): the PR should automatically update itself with the new stuff.

PR #4405, opened 2022-03-23
[SHELL32] Make Control Panel applets start in single instance

RaymondCzerny:
## Purpose

In MS Windows all control panel applets are single instance applications. This prevents the collision of multiple accesses to the configuration data.

Jira tickets: [CORE-7921](https://jira.reactos.org/browse/CORE-7921), [CORE-17025](https://jira.reactos.org/browse/CORE-17025)

## Proposed changes

Before start of the applets, is now searched the dialog window of the preinstance. If the search is successful, this dialog is brought to the foreground and a second instance not started.

Supersedes #4404.

### DIFF
--- a/dll/win32/shell32/wine/control.c
+++ b/dll/win32/shell32/wine/control.c
@@ -2,6 +2,7 @@
  *
  * Copyright 2001 Eric Pouech
  * Copyright 2008 Owen Rudge
+ * Copyright 2022 Raymond Czerny
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
@@ -736,6 +737,75 @@ static	void	Control_DoWindow(CPanel* panel, HWND hWnd, HINSTANCE hInst)
 }
 #endif
 
+#ifdef __REACTOS__
+
+/** Structure for in and out data when
+ *  search for the cpl dialog of first instance
+ */
+typedef struct tagAppDlgFindData
+{
+    PCWSTR    szAppFile;  /**< Full path to applet library as search parameter */
+    UINT_PTR  sAppletNo;  /**< Number of applet in a system control library as search parameter */
+    ATOM      aCPLName;   /**< to read window property 'CPLName' */
+    ATOM      aCPLFlags;  /**< to read window property 'CPLFlags'*/
+    HWND      hRunDLL;    /**< to skip self instance */
+    HWND      hDlgResult; /**< Returned dialog handle or NULL if not found */
+} AppDlgFindData;
+
+/**
+ * Callback function to search applet dialog
+ * @param hwnd A handle to a top-level window.
+ * @param lParam Pointer of AppDlgFindData
+ * @return TRUE: continue enumeration, FALSE: stop enumeration
+ */
+static BOOL CALLBACK
+Control_EnumWinProc(
+    _In_ HWND   hwnd,
+    _In_ LPARAM lParam)
+{
+    AppDlgFindData* pData = (AppDlgFindData*)lParam;
+    WCHAR szClassName[256] = L"";
+
+    if (pData->hRunDLL == hwnd)
+    {
+        // Skip self instance
+        return TRUE;
+    }
+
+    if (GetClassNameW(hwnd, szClassName, _countof(szClassName)))
+    {
+        // Note: A comparison on identical is not possible, the class names are different.
+        // ReactOS: 'rundll32_window'
+        // WinXP: 'RunDLL'
+        // other OS: not checked
+        if (StrStrIW(szClassName, L"rundll32") != NULL)
+        {
+            UINT_PTR sAppletNo;
+
+            sAppletNo = (UINT_PTR)GetPropW(hwnd, (LPTSTR)MAKEINTATOM(pData->aCPLFlags));
+            if (sAppletNo == pData->sAppletNo)
+            {
+                HANDLE hRes;
+                WCHAR szAppFile[MAX_PATH];
+
+                hRes = GetPropW(hwnd, (LPTSTR)MAKEINTATOM(pData->aCPLName));
+                GlobalGetAtomNameW((ATOM)HandleToUlong(hRes), szAppFile, _countof(szAppFile));
+                if (wcscmp(szAppFile, pData->szAppFile) == 0)
+                {
+                    HWND hDialog = GetLastActivePopup(hwnd);
+                    if (IsWindow(hDialog))
+                    {
+                        pData->hDlgResult = hDialog;
+                        return FALSE; // stop enumeration
+                    }
+                }
+            }
+        }
+    }
+    return TRUE; // continue enumeration
+}
+#endif /* __REACTOS__ */
+
 static	void	Control_DoLaunch(CPanel* panel, HWND hWnd, LPCWSTR wszCmd)
    /* forms to parse:
     *	foo.cpl,@sp,str
@@ -828,6 +898,10 @@ static	void	Control_DoLaunch(CPanel* panel, HWND hWnd, LPCWSTR wszCmd)
 #ifdef __REACTOS__
     ULONG_PTR cookie;
     BOOL bActivated;
+    ATOM aCPLName;
+    ATOM aCPLFlags;
+    ATOM aCPLPath;
+    AppDlgFindData findData;
 #endif
         /* we've been given a textual parameter (or none at all) */
         if (sp == -1) {
@@ -846,10 +920,51 @@ static	void	Control_DoLaunch(CPanel* panel, HWND hWnd, LPCWSTR wszCmd)
 
 #ifdef __REACTOS__
         bActivated = (applet->hActCtx != INVALID_HANDLE_VALUE ? ActivateActCtx(applet->hActCtx, &cookie) : FALSE);
+
+        aCPLPath = GlobalFindAtomW(applet->cmd);
+        if (!aCPLPath)
+        {
+            aCPLPath = GlobalAddAtomW(applet->cmd);
+        }
+
+        aCPLName = GlobalFindAtomW(L"CPLName");
+        if (!aCPLName)
+        {
+            aCPLName = GlobalAddAtomW(L"CPLName");
+        }
+
+        aCPLFlags = GlobalFindAtomW(L"CPLFlags");
+        if (!aCPLFlags)
+        {
+            aCPLFlags = GlobalAddAtomW(L"CPLFlags");
+        }
+
+        findData.szAppFile = applet->cmd;
+        findData.sAppletNo = (UINT_PTR)(sp + 1);
+        findData.aCPLName = aCPLName;
+        findData.aCPLFlags = aCPLFlags;
+        findData.hRunDLL = applet->hWnd;
+        findData.hDlgResult = NULL;
+        // Find the dialog of this applet in the first instance.
+        // Note: The simpler functions "FindWindow" or "FindWindowEx" does not find this type of dialogs.
+        EnumWindows(Control_EnumWinProc, (LPARAM)&findData);
+        if (findData.hDlgResult)
+        {
+            BringWindowToTop(findData.hDlgResult);
+        }
+        else
+        {
+            SetPropW(applet->hWnd, (LPTSTR)MAKEINTATOM(aCPLName), (HANDLE)MAKEINTATOM(aCPLPath));
+            SetPropW(applet->hWnd, (LPTSTR)MAKEINTATOM(aCPLFlags), UlongToHandle(sp + 1));
 #endif
 
         if (!applet->proc(applet->hWnd, CPL_STARTWPARMSW, sp, (LPARAM)extraPmts))
             applet->proc(applet->hWnd, CPL_DBLCLK, sp, applet->info[sp].data);
+#ifdef __REACTOS__
+            RemovePropW(applet->hWnd, applet->cmd);
+            GlobalDeleteAtom(aCPLPath);
+        }
+#endif
 
         Control_UnloadApplet(applet);
 


learn-more:
Please do not create a new PR for every change, but update your PR after receiving feedback.
and use a title that is descriptive.

dll/win32/shell32/wine/control.c
```diff
@@ -736,6 +736,53 @@ static	void	Control_DoWindow(CPanel* panel, HWND hWnd, HINSTANCE hInst)
...
+    {
+        // check of first search value
+        if (wcscmp(L"#32770", szClassName) == 0)
+        {
+            WCHAR szCaption[256] = L"";
+            if (GetWindowTextW(hwnd, szCaption, _countof(szCaption)))
+            {
+                AppDlgFindData* pData = (AppDlgFindData*)lParam;
+                // check of second search value
+                if(wcsstr(szCaption, pData->szCaption) != NULL)
```
HBelusca, line 773:
```suggestion
                if (wcsstr(szCaption, pData->szCaption) != NULL)
```

HBelusca (commented):
Looks fine as a first implementation.
I may think of a way Windows goes further than that: What would happen if two same applets are run, but each in a different language?

RaymondCzerny:
I have since found a language-independent method to solve the single instance problem.

Question: How should I technically attach this re-implentation to the current PR.

HBelusca:
> Question: How should I technically attach this re-implentation to the current PR.

You just make a new commit in your branch and push it. This will update the PR.

dll/win32/shell32/CMakeLists.txt
```diff
@@ -114,7 +114,7 @@ set_source_files_properties(shell32.rc PROPERTIES OBJECT_DEPENDS ${CMAKE_CURRENT
 set_module_type(shell32 win32dll UNICODE)
 target_link_libraries(shell32 shellmenu shelldesktop wine uuid recyclebin cpprt atl_classes)
 add_delay_importlibs(shell32 powrprof shdocvw devmgr winspool.drv winmm mpr uxtheme ole32 oleaut32 userenv browseui version fmifs)
-add_importlibs(shell32 advapi32 gdi32 user32 comctl32 comdlg32 shlwapi msvcrt kernel32 ntdll)
+add_importlibs(shell32 advapi32 gdi32 user32 comctl32 comdlg32 shlwapi msvcrt kernel32 ntdll psapi)
```
learn-more, line 117: `msvcrt kernel32 ntdll` should be the last imports (and in that order)

dll/win32/shell32/wine/control.c
```diff
@@ -736,6 +737,96 @@ static	void	Control_DoWindow(CPanel* panel, HWND hWnd, HINSTANCE hInst)
...
+{
+    WCHAR szClassName[256] = L"";
+    if (GetClassNameW(hwnd, szClassName, _countof(szClassName)))
+    {
+        // check of first search value
+        if (wcscmp(L"#32770", szClassName) == 0)
+        {
+            AppDlgFindData* pData = (AppDlgFindData*)lParam;
+            // check of second search value
+            if (Control_FindAndCheckAppFile(hwnd, pData->szAppFile))
```
learn-more, line 818: I don't think this is a good way to determine if a control panel applet is already running,
because now you are opening **each process** that spawns a dialog, to check if it has a certain dll loaded.

And a small nitpick:
A comment like `// check of second search value` does not really add value to the code, either explain _why_ something is done, but just explaining what can be read in the next line does not add much value.

learn-more (changes requested):
The current solution seems over-engineered, and there are various problems with this implementation from a security point of view.

dll/win32/shell32/wine/control.c
RaymondCzerny: Control_FindAndCheckAppFile is called only for processes whose dialog class is '#32770'. So **not all processes** will be opened.
However, I can move the class check from Control_EnumWinProc to Control_FindAndCheckAppFile to eliminate this problem, which can occur when Control_FindAndCheckAppFile is called from another location.

dll/win32/shell32/CMakeLists.txt
RaymondCzerny: Okay

dll/win32/shell32/wine/control.c
learn-more: > Control_FindAndCheckAppFile is called only for processes whose dialog class is '#32770'. So **not all processes** will be opened. However, I can move the class check from Control_EnumWinProc to Control_FindAndCheckAppFile to eliminate this problem, which can occur when Control_FindAndCheckAppFile is called from another location.

In practice this means any application that opens a dialog.

To me it does not make sense to open every process, and look for loaded modules for something like this.
I would assume that windows attaches a window property to a dialog or something like that,
which can be retrieved by all applications without having to open the process.

HBelusca: It may be more probable that the Windows shell maintains a list of the control panels it opened, and when a "new" one gets opened, it is checked against that maintained list, and if present, the existing CPL dialog gets focused instead.

RaymondCzerny: How many dialogs of class '#32770' will be open at the same time in practice? I guess between 1 and 5. Does browsing 5 processes bring the entire operating system to its knees?
Dialog class '#32770' is a system class used by operating system applications. Normal programs (Notepad, Wordpad, Paint, ...) use their own classes for their dialogs, except for those coming from comdlg32.dll (FileOpen, FileSave, Print, ...).

My examination of the CPL dialogs concerning the attributes GWL_STYLE and GWL_EXSTYLE did not reveal any peculiarities compared to the other dialogs of the class '#32770'.

The GWL_USERDATA element is too unspecific and may already be used for the tasks of the applet itself.

How do we want to proceed now?
a) Discard the code and ignore the problem of individual instances and the possible problems of parallel access to configuration data.
b) Return to the initial implementation with the multi-language problem.
c) Take over after moving the class check to Control_FindAndCheckAppFile.
d) Unchanged takeover.

No decision is a bad decision.

RaymondCzerny: > It may be more probable that the Windows shell maintains a list of the control panels it opened, and when a "new" one gets opened, it is checked against that maintained list, and if present, the existing CPL dialog gets focused instead.

A typical chicken - egg problem.

Only after the Windows shell has given control to the CPL applet, this applet itself creates the dialog. When the dialog is closed, control returns to the Windows shell for cleanup.

How should the applet pass its dialog handle to the Windows shell so that it can keep a list of open dialogs.

--- Code snipped ---
            if (!applet->proc(applet->hWnd, CPL_STARTWPARMSW, sp, (LPARAM)extraPmts))
                applet->proc(applet->hWnd, CPL_DBLCLK, sp, applet->info[sp].data);
---
The applet call is blocking.

learn-more: image

These atoms appear on Windows Server 2003 SP2 when opening the first .cpl file,
so they will most likely be used with `SetProp` and `GetProp`:
https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setpropW
https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getpropw

learn-more: > How many dialogs of class '#32770' will be open at the same time in practice? I guess between 1 and 5. Does browsing 5 processes bring the entire operating system to its knees? Dialog class '#32770' is a system class used by operating system applications. 

'#32770' is a system class, but not in the way that it is only in use by the operating system.
It is used for ANY window that is created as a Dialog instead of as a Window (So not using `CreateWindow`, but any of the `DialogBoxXXX` functions



And no, this does not 'bring the operating system to it's knees', which you would have seen when testing,
but opening a process like this is simply not always possible.
What if the process is elevated / has different permissions?

RaymondCzerny: > These atoms appear on Windows Server 2003 SP2 when opening the first .cpl file, so they will most likely be used with `SetProp` and `GetProp`: https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setpropW https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getpropw

The functions `SetProp` and `GetProp` are new to me. The only method I know to piggyback data to a window is based on GWL_USERDATA.

I will need a little time to get to grips with the functions. Maybe there is indeed a more elegant solution.

learn-more: Apimonitor might be of use here, if you add window related functions (SetProp, GetProp, FindWindow, EnumWindows, GetWindowClassname, ....) and all atom related functions, you can probably deduce what / how windows this is doing.

HBelusca: In addition to what Mark pointed (the SetProp/GetProp stuff), one should also consider the window argument passed to the `Control_RunDll` call (which corresponds to the main window of the controlling application for the CPL).

JoachimHenze: > It may be more probable that the Windows shell maintains a list of the control panels it opened, and when a "new" one gets opened, it is checked against that maintained list, and if present, the existing CPL dialog gets focused instead.

Unless it could be proven that this would be exactly what Windows does, I do think a stateless solution is preferable.

dll/win32/shell32/CMakeLists.txt
JoachimHenze, line 117: I checked with MSDepends: Win 2k3sp2 shell32.dll does not directly import anything from psapi.dll at all. Neither static nor delay-loaded import. A perfect solution would be avoiding that import altogether therefore.
In case you wouldn't be able to avoid it entirely in the final solution, then a delay-load is the weaker form of violation.

HBelusca: Indeed, Windows' shell32 doesn't require enumerating processes in order to find the opened control panels windows.

learn-more: This is already a moot point based on my comments?

learn-more (approved):

learn-more:
Please note that there is still one commit in there (the PSDK commit) that does not belong there, so that should be removed either by the person merging this, or by the author.

binarymaster (approved):

dll/win32/shell32/wine/control.c
```diff
@@ -736,6 +737,107 @@ static	void	Control_DoWindow(CPanel* panel, HWND hWnd, HINSTANCE hInst)
...
 
+#ifdef __REACTOS__
+
+/** Structure for in and out data when
+ *  search for the cpl dialog of first instance
+ */
+typedef struct tagAppDlgFindData
+{
+    PCWSTR  szAppFile;  /**< Full path to applet library as search parameter */
+    signed  sAppletNo;  /**< Number of applet in a system control library as search parameter */
```
HBelusca, line 748: normally a number of elements is not signed (and the type would be something like UINT or something similar),

dll/win32/shell32/wine/control.c
```diff
@@ -736,6 +737,107 @@ static	void	Control_DoWindow(CPanel* panel, HWND hWnd, HINSTANCE hInst)
...
+    ATOM    aCPLFlags;  /**< to read window property 'CPLFlags'*/
+    HWND    hRunDLL;    /**< to skip self instance */
+    HWND    hDlgResult; /**< Returned dialog handle or NULL if not found */
+} AppDlgFindData;
+
+/*
+ * Case insensitive variant of wcsstr
+ */
+static
+wchar_t * wcsistr(const wchar_t *s, const wchar_t *b)
```
HBelusca, line 759: Since this is shell code, you could use StrStrIW: https://docs.microsoft.com/en-us/windows/win32/api/shlwapi/nf-shlwapi-strstriw

dll/win32/shell32/wine/control.c
```diff
@@ -736,6 +737,107 @@ static	void	Control_DoWindow(CPanel* panel, HWND hWnd, HINSTANCE hInst)
...
+ * Callback function to search applet dialog
+ * @param hwnd A handle to a top-level window.
+ * @param lParam Pointer of AppDlgFindData
+ * @return TRUE: continue enumeration, FALSE: stop enumeration
+ */
+static BOOL CALLBACK
+Control_EnumWinProc(
+    _In_ HWND   hwnd,
+    _In_ LPARAM lParam
+)
```
HBelusca, line 794:
```suggestion
    _In_ LPARAM lParam)
```

dll/win32/shell32/wine/control.c
```diff
@@ -736,6 +737,107 @@ static	void	Control_DoWindow(CPanel* panel, HWND hWnd, HINSTANCE hInst)
...
+{
+    AppDlgFindData* pData;
+    pData = (AppDlgFindData*)lParam;
+    
+    if(pData->hRunDLL == hwnd)
```
HBelusca, line 799:
```suggestion
    if (pData->hRunDLL == hwnd)
```

dll/win32/shell32/wine/control.c
```diff
@@ -736,6 +737,107 @@ static	void	Control_DoWindow(CPanel* panel, HWND hWnd, HINSTANCE hInst)
...
+                HANDLE hRes;
+                WCHAR szAppFile[MAX_PATH];
+
+                hRes = GetPropW(hwnd, (LPTSTR)MAKELONG(pData->aCPLName, 0));
+                GlobalGetAtomNameW((ATOM)(LONG)hRes, szAppFile, MAX_PATH);
+                if (wcscmp(szAppFile, pData->szAppFile) == 0)
+                {
+                    HWND hDialog;
+
+                    hDialog = GetLastActivePopup(hwnd);
```
HBelusca, line 827:
```suggestion
                    HWND hDialog = GetLastActivePopup(hwnd);
```

dll/win32/shell32/wine/control.c
```diff
@@ -736,6 +737,107 @@ static	void	Control_DoWindow(CPanel* panel, HWND hWnd, HINSTANCE hInst)
...
+                    if (IsWindow(hDialog))
+                    {
+                        pData->hDlgResult = hDialog;
+                        return FALSE;// stop enumeration
```
HBelusca, line 831:
```suggestion
                        return FALSE; // stop enumeration
```

dll/win32/shell32/wine/control.c
```diff
@@ -846,11 +952,50 @@ static	void	Control_DoLaunch(CPanel* panel, HWND hWnd, LPCWSTR wszCmd)
...
+            SetPropW(applet->hWnd, (LPTSTR)MAKELONG(aCPLName, 0), (HANDLE)MAKELONG(aCPLPath, 0));
+            SetPropW(applet->hWnd, (LPTSTR)MAKELONG(aCPLFlags, 0), (HANDLE)(sp + 1));
+#endif
+            if (!applet->proc(applet->hWnd, CPL_STARTWPARMSW, sp, (LPARAM)extraPmts))
+                applet->proc(applet->hWnd, CPL_DBLCLK, sp, applet->info[sp].data);
+#ifdef __REACTOS__
+            RemovePropW(applet->hWnd, applet->cmd);
+            GlobalDeleteAtom(aCPLPath);
+        }
+#endif
```
HBelusca, line 998:
```suggestion
#endif

```

dll/win32/shell32/wine/control.c
```diff
@@ -736,6 +737,107 @@ static	void	Control_DoWindow(CPanel* panel, HWND hWnd, HINSTANCE hInst)
...
+                        pData->hDlgResult = hDialog;
+                        return FALSE;// stop enumeration
+                    }
+                }
+            }
+        }
+    }
+    return TRUE; // continue enumeration
+}
+#endif
```
HBelusca, line 839: (Putting what the closing endif refers to; see what is usually done in our codebase.)
```suggestion
#endif /* __REACTOS__ */
```

dll/win32/shell32/wine/control.c
```diff
@@ -736,6 +737,77 @@ static	void	Control_DoWindow(CPanel* panel, HWND hWnd, HINSTANCE hInst)
...
+    pData = (AppDlgFindData*)lParam;
+
+    if (pData->hRunDLL == hwnd)
+    {
+        // Skip self instance
+        return TRUE;
+    }
+
+    WCHAR szClassName[256] = L"";
+
```
HBelusca, line 776:
```suggestion
```

dll/win32/shell32/wine/control.c
```diff
@@ -736,6 +737,77 @@ static	void	Control_DoWindow(CPanel* panel, HWND hWnd, HINSTANCE hInst)
...
+ * @param lParam Pointer of AppDlgFindData
+ * @return TRUE: continue enumeration, FALSE: stop enumeration
+ */
+static BOOL CALLBACK
+Control_EnumWinProc(
+    _In_ HWND   hwnd,
+    _In_ LPARAM lParam)
+{
+    AppDlgFindData* pData;
+    pData = (AppDlgFindData*)lParam;
```
HBelusca, line 767:
```suggestion
    AppDlgFindData* pData = (AppDlgFindData*)lParam;
    WCHAR szClassName[256] = L"";
```

dll/win32/shell32/wine/control.c
```diff
@@ -736,6 +737,77 @@ static	void	Control_DoWindow(CPanel* panel, HWND hWnd, HINSTANCE hInst)
...
+        // Skip self instance
+        return TRUE;
+    }
+
+    WCHAR szClassName[256] = L"";
+
+    if (GetClassNameW(hwnd, szClassName, _countof(szClassName)))
+    {
+        // Note: A comparison on identical is not possible, the class names are different.
+        // ReactOS: 'rundll32_window'
```
HBelusca, line 780: Could be of interest to investigate and maybe fix that later.

dll/win32/shell32/wine/control.c
```diff
@@ -736,6 +737,77 @@ static	void	Control_DoWindow(CPanel* panel, HWND hWnd, HINSTANCE hInst)
...
+        // WinXP: 'RunDLL'
+        // other OS: not checked
+        if (StrStrIW(szClassName, L"rundll32") != NULL)
+        {
+            UINT sAppletNo;
+
+            sAppletNo = (UINT)GetPropW(hwnd, (LPTSTR)MAKELONG(pData->aCPLFlags, 0));
```
HBelusca, line 787:
```suggestion
            sAppletNo = (UINT)GetPropW(hwnd, (LPTSTR)MAKEINTATOM(pData->aCPLFlags));
```

dll/win32/shell32/wine/control.c
```diff
@@ -736,6 +737,77 @@ static	void	Control_DoWindow(CPanel* panel, HWND hWnd, HINSTANCE hInst)
...
+            if (sAppletNo == pData->sAppletNo)
+            {
+                HANDLE hRes;
+                WCHAR szAppFile[MAX_PATH];
+
+                hRes = GetPropW(hwnd, (LPTSTR)MAKELONG(pData->aCPLName, 0));
```
HBelusca, line 793:
```suggestion
                hRes = GetPropW(hwnd, (LPTSTR)MAKEINTATOM(pData->aCPLName));
```

dll/win32/shell32/wine/control.c
```diff
@@ -846,10 +922,50 @@ static	void	Control_DoLaunch(CPanel* panel, HWND hWnd, LPCWSTR wszCmd)
...
+        // Find the dialog of this applet in the first instance.
+        // Note: The simpler functions "FindWindow" or "FindWindowEx" does not find this type of dialogs.
+        EnumWindows(Control_EnumWinProc, (LPARAM)&findData);
+        if (findData.hDlgResult)
+        {
+            BringWindowToTop(findData.hDlgResult);
+        }
+        else
+        {
+            SetPropW(applet->hWnd, (LPTSTR)MAKELONG(aCPLName, 0), (HANDLE)MAKELONG(aCPLPath, 0));
```
HBelusca, line 959:
```suggestion
            SetPropW(applet->hWnd, (LPTSTR)MAKEINTATOM(aCPLName), (HANDLE)MAKEINTATOM(aCPLPath));
```
(Should we need to use `UlongToHandle` for the `aCPLPath`?)

dll/win32/shell32/wine/control.c
```diff
@@ -846,10 +922,50 @@ static	void	Control_DoLaunch(CPanel* panel, HWND hWnd, LPCWSTR wszCmd)
...
+            SetPropW(applet->hWnd, (LPTSTR)MAKELONG(aCPLFlags, 0), (HANDLE)(sp + 1));
```
HBelusca, line 960:
```suggestion
            SetPropW(applet->hWnd, (LPTSTR)MAKEINTATOM(aCPLFlags), (HANDLE)(sp + 1));
```
The `(HANDLE)(sp + 1)` part may need actually `UlongToHandle(sp + 1)`.

dll/win32/shell32/wine/control.c
```diff
@@ -736,6 +737,77 @@ static	void	Control_DoWindow(CPanel* panel, HWND hWnd, HINSTANCE hInst)
...
+            UINT sAppletNo;
+
+            sAppletNo = (UINT)GetPropW(hwnd, (LPTSTR)MAKELONG(pData->aCPLFlags, 0));
+            if (sAppletNo == pData->sAppletNo)
+            {
+                HANDLE hRes;
+                WCHAR szAppFile[MAX_PATH];
+
+                hRes = GetPropW(hwnd, (LPTSTR)MAKELONG(pData->aCPLName, 0));
+                GlobalGetAtomNameW((ATOM)(LONG)hRes, szAppFile, MAX_PATH);
```
HBelusca, line 794:
```suggestion
                GlobalGetAtomNameW((ATOM)(LONG)hRes, szAppFile, _countof(szAppFile));
```

binarymaster:
I guess everything is addressed now. Will be merged as soon as we resolve some build problems on master.

dll/win32/shell32/wine/control.c
binarymaster: Ok let's do it like this for consistency.

HBelusca:
OK so we've got
```
control.c(785): warning C4311: 'type cast': pointer truncation from 'HANDLE' to 'UINT'
control.c(957): error C4047: 'function': 'const long' differs in levels of indirection from 'LPTSTR'
control.c(957): warning C4024: 'ULongToHandle': different types for formal and actual parameter 1
```

dll/win32/shell32/wine/control.c
```diff
@@ -846,10 +920,51 @@ static	void	Control_DoLaunch(CPanel* panel, HWND hWnd, LPCWSTR wszCmd)
...
+        // Find the dialog of this applet in the first instance.
+        // Note: The simpler functions "FindWindow" or "FindWindowEx" does not find this type of dialogs.
+        EnumWindows(Control_EnumWinProc, (LPARAM)&findData);
+        if (findData.hDlgResult)
+        {
+            BringWindowToTop(findData.hDlgResult);
+        }
+        else
+        {
+            SetPropW(applet->hWnd, (LPTSTR)MAKEINTATOM(aCPLName), UlongToHandle(MAKEINTATOM(aCPLPath)));
```
HBelusca, line 957:
```suggestion
            SetPropW(applet->hWnd, (LPTSTR)MAKEINTATOM(aCPLName), (HANDLE)MAKEINTATOM(aCPLPath));
```
Here the direct cast was OK (should fix the x64 build).

dll/win32/shell32/wine/control.c
```diff
@@ -736,6 +737,75 @@ static	void	Control_DoWindow(CPanel* panel, HWND hWnd, HINSTANCE hInst)
...
+    {
+        // Note: A comparison on identical is not possible, the class names are different.
+        // ReactOS: 'rundll32_window'
+        // WinXP: 'RunDLL'
+        // other OS: not checked
+        if (StrStrIW(szClassName, L"rundll32") != NULL)
+        {
+            UINT sAppletNo;
+
+            sAppletNo = (UINT)GetPropW(hwnd, (LPTSTR)MAKEINTATOM(pData->aCPLFlags));
```
HBelusca, line 785: We could fix the x64 build error if we used an UINT_PTR for sAppletNo (here + in the structure definition above).

dll/win32/shell32/wine/control.c
```diff
@@ -846,10 +920,51 @@ static	void	Control_DoLaunch(CPanel* panel, HWND hWnd, LPCWSTR wszCmd)
...
+        }
+
+        aCPLFlags = GlobalFindAtomW(L"CPLFlags");
+        if (!aCPLFlags)
+        {
+            aCPLFlags = GlobalAddAtomW(L"CPLFlags");
+        }
+
+        findData.szAppFile = applet->cmd;
+        findData.sAppletNo = sp + 1;
```
HBelusca, line 943: Here, (sp + 1) might need a cast to UINT(_PTR).

dll/win32/shell32/wine/control.c
HBelusca: That's precisely what will make the build fail. Because you are going to truncate a "HANDLE" (which has the size of a platform-sized integer, i.e. 32 or 64 bits), to a fixed size.

PR #4407, opened 2022-03-25
Move boot-related commands from boot/CMakeLists.txt into a boot_images.cmake sub-file.

HBelusca:
## Purpose

This allows the CMakeLists.txt file to be less cluttered by unrelated stuff.


boot/boot_images.cmake
```diff
@@ -0,0 +1,165 @@
+## efisys.bin
+
+# EFI platform ID, used in environ/CMakelists.txt for bootmgfw filename naming also.
+if(ARCH STREQUAL "amd64")
```
learn-more, line 4: Would it make sense to do this in `/CMakeLists.txt` next to where `WINARCH` is defined.

boot/boot_images.cmake
```diff
@@ -0,0 +1,165 @@
...
+    file(APPEND ${_image_filelist} "${_rootdir}/${_username}/Templates=${CMAKE_CURRENT_BINARY_DIR}/empty\n")
+endfunction()
+
+
+## BootCD
+# Create the file list
+file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/bootcd.cmake.lst "")
+file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/bootcd.cmake.lst "${CMAKE_CURRENT_BINARY_DIR}/empty\n")
+
+add_custom_target(bootcd
```
learn-more, line 102: I know that you just moved this thing, and I am having a hard time reading the mkisofs code,
but would it be possible to write the command line in a 'response file' like for example compilers do, so that we don't have command lines that span 4 screens?

learn-more: So that the resulting commandline would be something like:
`native-mkisofs @bootcd.txt`

boot/boot_images.cmake
HBelusca: It would, maybe, if we needed this identifier elsewhere than just in this very precise place.

boot/boot_images.cmake
HBelusca: > and I am having a hard time reading the mkisofs code,
but would it be possible to write the command line in a 'response file'

Response files have to be supported by the tool in question, but it's not the case for mkisofs (as far as I can see from its help file). The best I can do here, is to just improve the looks of the command line in that cmake file by going to a newline for each argument (or set of arguments that make sense).

HBelusca: As we agreed on chat, I'm going to commit this PR without any supplemental changes in the mkisofs options; instead such changes will be done in a separate PR.

PR #4410, opened 2022-03-27
[NTOS] A list of Misc Changes needed to boot ReactOS with SMP

DarkFire01:
## Purpose

This PR is mostly to get conversation going about all the small changes me and  @HBelusca 
discovered while using ReactOS with SMP support.
This is a draft as I may add more to it the coming days.

However:
This isn't every problem with the multiprocessor kernel it's only the small stuff that kills the BSP when its called by the other CPUs

## Proposed changes

- [NTOS] Swap MAXIMUM_PROCESSORS with NUMBER_POOL_LOOKASIDE_LISTS
- [NTOS] fix timer lock data and hardcode
- [NTOS] Increment ActiveProcessors accurately
- [NTOS] Set SwapBusy properly for i386

### DIFF
--- a/ntoskrnl/ke/i386/thrdini.c
+++ b/ntoskrnl/ke/i386/thrdini.c
@@ -427,6 +427,9 @@ KiSwapContextEntry(IN PKSWITCHFRAME SwitchFrame,
     /* Get the old thread and set its kernel stack */
     OldThread->KernelStack = SwitchFrame;
 
+    /* Set swapbusy to false for the new thread */
+    NewThread->SwapBusy = FALSE;
+
     /* ISRs can change FPU state, so disable interrupts while checking */
     _disable();
 
--- a/ntoskrnl/ex/lookas.c
+++ b/ntoskrnl/ex/lookas.c
@@ -20,8 +20,8 @@ LIST_ENTRY ExpPagedLookasideListHead;
 KSPIN_LOCK ExpPagedLookasideListLock;
 LIST_ENTRY ExSystemLookasideListHead;
 LIST_ENTRY ExPoolLookasideListHead;
-GENERAL_LOOKASIDE ExpSmallNPagedPoolLookasideLists[MAXIMUM_PROCESSORS];
-GENERAL_LOOKASIDE ExpSmallPagedPoolLookasideLists[MAXIMUM_PROCESSORS];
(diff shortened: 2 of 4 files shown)


ntoskrnl/ke/i386/thrdini.c
```diff
@@ -438,6 +438,10 @@ KiSwapContextEntry(IN PKSWITCHFRAME SwitchFrame,
     /* Get the old thread and set its kernel stack */
     OldThread->KernelStack = SwitchFrame;
 
+    /* If it isn't already set, set swapbusy to false for the new thread */
+    if(NewThread->SwapBusy != FALSE)
```
HBelusca, line 442:
```suggestion
    if (NewThread->SwapBusy != FALSE)
```

ntoskrnl/ke/i386/thrdini.c
```diff
@@ -505,6 +509,9 @@ KiDispatchInterrupt(VOID)
         NewThread->State = Running;
         OldThread->WaitReason = WrDispatchInt;
 
+        /* Lock the PRCB */
+        KiAcquirePrcbLock(Prcb);
```
HBelusca, line 513: Maybe you can mention where it's going to be unlocked?

DarkFire01:
image

Just for the record these changes don't seem to affect the UP kernel at all in any visible way!
But :P that's why this is opened! Just in case i missed something

ntoskrnl/ke/i386/thrdini.c
DarkFire01: Removed! It seems I fixed it another way in the SMP branch.

ntoskrnl/ke/i386/thrdini.c
```diff
@@ -427,6 +427,10 @@ KiSwapContextEntry(IN PKSWITCHFRAME SwitchFrame,
     /* Get the old thread and set its kernel stack */
     OldThread->KernelStack = SwitchFrame;
 
+    /* If it isn't already set, set swapbusy to false for the new thread */
+    if (NewThread->SwapBusy != FALSE)
```
SergeGautherie, line 431: Superfluous `if`?

tkreuzer, line 431: `if (a != 0) a= 0;`
Pretty silly, isn't it?

tkreuzer (approved):

tkreuzer:
Please make sure the builds are not broken!

PR #4411, opened 2022-03-28
PSDK:IsReactOS

RaymondCzerny:
## Purpose

The old implementation of the IsReactOS function is unsafe because it relies on the installation directory being named ReactOS.

## Proposed changes

The new implementation reads the 'ProductName' field from the VersionInformation resource of the shell32.dll library.


dll/win32/shell32/wine/control.c
```diff
@@ -736,6 +737,96 @@ static	void	Control_DoWindow(CPanel* panel, HWND hWnd, HINSTANCE hInst)
...
+
+
+/**
+ * Callback function to search applet dialog
+ * @param hwnd A handle to a top-level window.
+ * @param lParam Pointer of AppDlgFindData
+ * @return TRUE: continue enumeration, FALSE: stop enumeration
+ */
+static BOOL CALLBACK
+Control_EnumWinProc(
```
binarymaster, line 805: It seems you have branched from other PR branch, so this PR contains unrelated changes.

binarymaster: How to fix:
- Do a hard reset to the master branch state
- Cherry-pick new commits on top of it, and do a force push

binarymaster:
You forgot cherry pick:
```
git cherry-pick 12895c11e20d2c409841820a95be728a83e85b9c
git push --force-with-lease
```

sdk/include/psdk/versionhelpers.h
```diff
@@ -148,9 +149,58 @@ IsActiveSessionCountLimited()
...
 {
-    // FIXME: Find a better method!
-    WCHAR szWinDir[MAX_PATH];
-    GetWindowsDirectoryW(szWinDir, _countof(szWinDir));
-    return (wcsstr(szWinDir, L"ReactOS") != NULL);
+    BOOL bResult = FALSE;
+    BOOL bLibFree = FALSE;
+    HMODULE hMod;
+
+    hMod = GetModuleHandleW(L"shell32");
```
HBelusca, line 156: It would be more reliable to use kernel32.dll instead (which is guaranteed to be always present, and loaded for win32 applications, contrary to shell32), or better, ntdll.dll ?

Also, have you thought about looking at some registry settings as an alternative/simpler way?
For example, the following registry key/value:
`HKLM\SOFTWARE\Microsoft\Windows NT\CurrentVersion` value `ProductName`

sdk/include/psdk/versionhelpers.h
```diff
@@ -148,9 +149,58 @@ IsActiveSessionCountLimited()
...
+        WCHAR szFullPath[MAX_PATH];
+
+        dwLen = GetModuleFileNameW(hMod, szFullPath, MAX_PATH);
+        if (dwLen)
+        {
+            DWORD dwSize;
+
+            szFullPath[dwLen] = 0; // required by Windows XP
+            dwSize = GetFileVersionInfoSizeW(szFullPath, NULL);
+            if(dwSize)
```
HBelusca, line 175:
```suggestion
            if (dwSize)
```

sdk/include/psdk/versionhelpers.h
```diff
@@ -148,9 +149,58 @@ IsActiveSessionCountLimited()
...
+            {
+                BOOL bProduct;
+                UINT cchVer = 0;
+                LPCWSTR lszValue = NULL;
+                HANDLE hMem = NULL;
+                LPVOID VerInfo = NULL;
+
+                hMem = GlobalAlloc(GMEM_MOVEABLE, dwSize);
```
HBelusca, line 183: Prefer using HeapAlloc/HeapFree.

sdk/include/psdk/versionhelpers.h
RaymondCzerny: The entries in the registry can be easily changed by the user or AD administrator.
Is this secure enough?

HBelusca: If you think it's better to have something that always tells "it's ReactOS" even if the user masquerades it by changing the registry entries manually, then it may be better to stay by checking the vendor of a core system dll indeed.

sdk/include/psdk/versionhelpers.h
RaymondCzerny: Do the GlobalLock and GlobalUnlock functions work or do alternative functions like HeapLock and HeapUnlock exist?

HBelusca: Why do you need "lock"/"unlock" stuff?

sdk/include/psdk/versionhelpers.h
learn-more: This is completely not about security, the only reason this exists is for unittests.
So a simple registry check on product name is more than enough, and way simpler.

RaymondCzerny: It was not obvious that these functions are only for unit tests. Why is this file not in the unit tests directory tree?

Currently, ReactOS is a replica of MS Windows. In the distant future, could ReactOS contain features that Windows does not? Application developers who want to use these features would appreciate such a public feature.

learn-more: > It was not obvious that these functions are only for unit tests. Why is this file not in the unit tests directory tree?
> 
> Currently, ReactOS is a replica of MS Windows. In the distant future, could ReactOS contain features that Windows does not? Application developers who want to use these features would appreciate such a public feature.

We do not support feature detection for stuff like that.
If this ever becomes a thing we might want to support, there should be checks on the new functionality, and not on what OS the code is running. (As should all things like this, but that is sadly not always possible)

learn-more: > It was not obvious that these functions are only for unit tests. Why is this file not in the unit tests directory tree?

I assumed you would have at least checked where a function is used before you start modifying it.
Did you check that this works at all?

RaymondCzerny: > I assumed you would have at least checked where a function is used before you start modifying it. Did you check that this works at all?

I have not checked if the IsReactOS function is used anywhere.

Whether this function works? With a miniprogram that prints the return value to the console. However still with the shell32.dll.
Now I have detected the problem with the VersionInformation resource.
shell32.dll => english => 040904B0
ntdll.dll => neutral => 000004B0

Ergo, I still have to change that.

learn-more: > > I assumed you would have at least checked where a function is used before you start modifying it. Did you check that this works at all?
> 
> I have not checked if the IsReactOS function is used anywhere.
> 
> Whether this function works? With a miniprogram that prints the return value to the console. However still with the shell32.dll. Now I have detected the problem with the VersionInformation resource. shell32.dll => english => 040904B0 ntdll.dll => neutral => 000004B0
> 
> Ergo, I still have to change that.

Right, but please re-read my comment:
In my opinion this is the wrong way to go, reading the registry is simpler, and more than sufficient for our needs.

HBelusca:
Commit history fixed by using the good old: `git fetch --all && git rebase upstream/master` (i.e. without using any git merge or git pull) : Removing the unrelated IMM commits and squashing the double-commits together (there were each of the commits twice with same contents, for unknown reasons...)

RaymondCzerny:
The original author of this feature was aware of how prone to failure it was. 
What is the use of error-prone tests?
The method with the registry is a good compromise.

learn-more:
> The original author of this feature was aware of how prone to failure it was. What is the use of error-prone tests? The method with the registry is a good compromise.

It is not meant to be robust, it is meant to detect the normal usage of a test run.

RaymondCzerny:
I give up and discard my efforts at this function.

PR #4412, opened 2022-03-28
[SHELL32]  Remove horizontal scrollbar from IDC_FILETYPES_LISTVIEW in FileTypes … 

zecarlos1957:
[SHELL32]  Remove horizontal scrollbar from IDC_FILETYPES_LISTVIEW
    in FileTypes Dialog

### DIFF
--- a/dll/win32/shell32/dialogs/filetypes.cpp
+++ b/dll/win32/shell32/dialogs/filetypes.cpp
@@ -1482,7 +1482,7 @@ FileTypesDlg_InitListView(HWND hwndDlg, HWND hListView)
     LVCOLUMNW col;
     WCHAR szName[50];
     DWORD dwStyle;
-    INT columnSize = 140;
+    INT columnSize;
 
     if (!LoadStringW(shell32_hInstance, IDS_COLUMN_EXTENSION, szName, _countof(szName)))
     {
@@ -1512,7 +1512,7 @@ FileTypesDlg_InitListView(HWND hwndDlg, HWND hListView)
     }
 
     col.iSubItem    = 1;
-    col.cx          = clientRect.right - clientRect.left - columnSize - GetSystemMetrics(SM_CYVTHUMB);
+    col.cx          = clientRect.right - clientRect.left - columnSize - GetSystemMetrics(SM_CYVSCROLL);
     col.cchTextMax  = wcslen(szName);
     col.pszText     = szName;
     SendMessageW(hListView, LVM_INSERTCOLUMNW, 1, (LPARAM)&col);
--- a/dll/win32/shell32/dialogs/filetypes.cpp
+++ b/dll/win32/shell32/dialogs/filetypes.cpp
@@ -1495,7 +1495,7 @@ FileTypesDlg_InitListView(HWND hwndDlg, HWND hListView)
 
     GetClientRect(hListView, &clientRect);
     ZeroMemory(&col, sizeof(LV_COLUMN));
-    columnSize      = 120;
+    columnSize      = (clientRect.right - clientRect.left) / 4;
     col.iSubItem    = 0;
     col.mask        = LVCF_WIDTH | LVCF_TEXT | LVCF_SUBITEM | LVCF_FMT;
     col.fmt         = LVCFMT_FIXED_WIDTH;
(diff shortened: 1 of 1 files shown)


zecarlos1957:
Before
BeforeOption


After
AfterOption

JoachimHenze:
Is "Interface de Friomware Extensivel Executavael" the longest string within the list above?
If not, then please attach a fresh "after-picture" with the very longest string in there. Is it worth the risk to risk string truncation if there will be any longer strings than that? The version that limits this text strictly leftmost of the "Alterara" nd "Avancadas" buttons (aka the "old state") is slightly more resilient to longer texts there. Also it is more in sync with other languages.
Is it really worth the change?

binarymaster:
I'd prefer a variant where text description does not go in the middle of two buttons at right.

JoachimHenze:
Alos please take into account that the text "Ficheros com..." and  (Abre Com) "Reactos" is not only under the control of ReactOS sources, but they are dynamically filled with the text from 3rd party applications, which sometimes use damn long strings in there when they do register their file types.
Your proposed newstyle is therefore more likely to cut off strings in such cases.

Therefore I do feel like binarymaster, and would prefer to close this PR without merging it. No matter what MS is doing here, because even in XPSP3 you might see string truncation here once in a while for the very same reason.
Our current approach is imho superior here for that reason.

JoachimHenze:
I do think that if we really wanted to improve something here in this dialog, then it should be about finding out how to hide the horizontal scrollbar by default, and show that only in case it is necessary by obscure string lengths.

HBelusca:
That's how it looks in Windows 2003:
image
On the left: file with extension; on the right: file without extension.

We see that for files with extension, the information given in the text is somewhat redundant.
My suggestion, if we want to improve the string, would be that two different strings get used, like that:
- for files with extension: `Files with extension 'xxx' are of type 'Super-long Type'.\nTo change settings that affect all such files, click Advanced.`
- for files without extension: `To change settings that affect all 'FileType without extension' files, click Advanced.`

zecarlos1957:
@JoachimHenze  It seems that most people like the current version of this dialog.
So can I change the subject of this PR to JoachimHenze's idea or is it preferable in a new PR?

zecarlos1957:
Thanks to @JoachimHenze  and @HBelusca  for  the tips

binarymaster (approved):
Latest changes are LGTM. :+1: Can you update before/after screenshots please?

dll/win32/shell32/dialogs/filetypes.cpp
```diff
@@ -1512,7 +1512,7 @@ FileTypesDlg_InitListView(HWND hwndDlg, HWND hListView)
     }
 
     col.iSubItem    = 1;
-    col.cx          = clientRect.right - clientRect.left - columnSize;
+    col.cx          = clientRect.right - clientRect.left - columnSize - GetSystemMetrics(SM_CXHTHUMB);
```
HBelusca, line 1515: How does the `SM_CXHTHUMB` part affect the horizontal bar if it is removed?

dll/win32/shell32/dialogs/filetypes.cpp
```diff
@@ -1512,7 +1512,7 @@ FileTypesDlg_InitListView(HWND hwndDlg, HWND hListView)
     }
 
     col.iSubItem    = 1;
-    col.cx          = clientRect.right - clientRect.left - columnSize;
+    col.cx          = clientRect.right - clientRect.left - columnSize - GetSystemMetrics(SM_CYVTHUMB);
```
HBelusca, line 1515:
```suggestion
    col.cx          = clientRect.right - clientRect.left - columnSize - GetSystemMetrics(SM_CYVSCROLL);
```

dll/win32/shell32/dialogs/filetypes.cpp
```diff
@@ -1495,7 +1495,7 @@ FileTypesDlg_InitListView(HWND hwndDlg, HWND hListView)
 
     GetClientRect(hListView, &clientRect);
     ZeroMemory(&col, sizeof(LV_COLUMN));
-    columnSize      = 120;
+    columnSize      = (clientRect.right - clientRect.left) / 4;
```
HBelusca, line 1498: You can remove the initial value of `columnSize` above at line 1485.

PR #4416, opened 2022-03-30
[IMM32] Renaming: s/pImmHeap/ghImmHeap/

katahiromz:
## Purpose

I had confirmed the name of `pImmHeap` on WinDBG, but @HBelusca thinks it's bad name. Renaming correctly...

## Proposed changes

- Rename `pImmHeap` as `ghImmHeap`.

### DIFF
--- a/dll/win32/imm32/precomp.h
+++ b/dll/win32/imm32/precomp.h
@@ -75,7 +75,7 @@ extern PIMEDPI g_pImeDpiList;
 extern PSERVERINFO gpsi;
 extern SHAREDINFO g_SharedInfo;
 extern BYTE g_bClientRegd;
-extern HANDLE ghImmHeap;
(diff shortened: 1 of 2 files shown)


learn-more:
Is the name exported somewhere, or is it only present in symbol files?
@HBelusca what's your take on this?

katahiromz:
It's present in symbol files.

HBelusca:
> Is the name exported somewhere, or is it only present in symbol files? @HBelusca what's your take on this?

Turns out it's only from symbol files, as katayama mentions, but not an exported variable. That's why I was doubtful about the usefulness of the change in master. (The only time when matching exactly a symbol name is necessary is when we know e.g. WinDbg -- via one of its standard commands -- expects to see such symbol in the module in question, in order to implement some debugging support for something.)

katahiromz:
Go or not?

PR #4425, opened 2022-04-01
[BOOTDATA] Reorganize the mkisofs ISO boot-related options, making them modular.

HBelusca:
## Purpose

The mkisofs options are quite long and repetitive (mostly duplicated four times for the current four different ISO targets), and are also hard to read (i.e. what group of options go together).
Reorganize the mkisofs ISO boot-related options in different thematic sets, and make them modular too.
This also allow to conditionally add/remove ISO boot entries for different platforms, for example:
- BIOS-based PC builds (only x86/x64) get the BIOS bootsector (excluded from IA64 and ARM32/64);
- EFI-based architectures get the EFI boot entry.

Based on a suggestion by Mark Jansen (see PR #4407).
Dedicated to Justin Miller ;)

## Proposed changes

- Reorganize the mkisofs ISO boot-related options;
- Adjust some old comments of mine in the process.


HBelusca:
Cc @DarkFire01 too.

boot/boot_images.cmake
```diff
@@ -59,6 +61,53 @@ ${_efisys_file} 1
...
+set(ISO_COMMON_OPTIONS
+    -iso-level 4 -publisher ${ISO_MANUFACTURER} -preparer ${ISO_MANUFACTURER} -volid ${ISO_VOLNAME} -volset ${ISO_VOLNAME})
+
+set(ISO_LAYOUT_OPTIONS
+    -duplicates-once -no-cache-inodes -graft-points)
+
+## "El Torito" ISO boot options.
+#
+set(ISO_BOOT_OPTIONS)
+set(ISO_BOOT_REGTEST_OPTIONS) # Useful only for BootCDRegTest
```
learn-more, line 74: Don't need to forward declare these.

boot/boot_images.cmake
```diff
@@ -59,6 +61,53 @@ ${_efisys_file} 1
...
+# BIOS-based PC boot entry (x86/x64 only)
+if(ARCH STREQUAL "i386" OR ARCH STREQUAL "amd64")
+    set(ISO_BOOT_OPTIONS
+        -eltorito-platform x86 -eltorito-boot loader/isoboot.bin -no-emul-boot -boot-load-size 4)
+    set(ISO_BOOT_REGTEST_OPTIONS
+        -eltorito-platform x86 -eltorito-boot loader/isobtrt.bin -no-emul-boot -boot-load-size 4)
+endif()
+
+# EFI boot entry
+list(APPEND EFI_SUPPORTED_ARCHS "i386" "amd64" "ia64" "arm" "arm64")
```
learn-more, line 85: Why `list(APPEND`?
Why not just set it?

boot/boot_images.cmake
```diff
@@ -59,6 +61,53 @@ ${_efisys_file} 1
...
+
+## "El Torito" ISO boot options.
+#
+set(ISO_BOOT_OPTIONS)
+set(ISO_BOOT_REGTEST_OPTIONS) # Useful only for BootCDRegTest
+
+# BIOS-based PC boot entry (x86/x64 only)
+if(ARCH STREQUAL "i386" OR ARCH STREQUAL "amd64")
+    set(ISO_BOOT_OPTIONS
+        -eltorito-platform x86 -eltorito-boot loader/isoboot.bin -no-emul-boot -boot-load-size 4)
```
learn-more, line 79: I think `x86` is the default option, do we need to explicitly specify it?
Also please take care, the order of these things can matter according to the documentation I found online:
> If  the  option -eltorito-platform appears before	the first -el-torito-boot option, it sets the platform id for the default boot entry.
 If the option -eltorito-platform appears after an -eltorito-boot option and sets the platform id to a value  different from  the previous value, it starts a new set of boot entries.

https://www.freebsd.org/cgi/man.cgi?query=mkisofs&sektion=8&format=html

binarymaster:
Cc @konradybcio for review, as you worked on #4050, and some EFI stuff is happening here :)

boot/boot_images.cmake
HBelusca: I copied the technique from https://git.reactos.org/?p=reactos.git;a=blob;f=sdk/cmake/CMakeMacros.cmake;hb=64290dd2c98f430c23a47100f655ce7e6849c3ee#l575
that seems to avoid the syntax with the explicit semicolons between the elements.

boot/boot_images.cmake
HBelusca: Yep that's expected. And explicitly specify the default should only clarify things.

boot/boot_images.cmake
Extravert-ir, line 85: What's the point in this list? We don't support other architectures

boot/boot_images.cmake
```diff
@@ -59,6 +61,53 @@ ${_efisys_file} 1
...
+if(ARCH STREQUAL "i386" OR ARCH STREQUAL "amd64")
+    set(ISO_BOOT_OPTIONS ${ISO_BOOT_OPTIONS}
+        -eltorito-alt-boot ${ISO_BOOT_EFI_OPTIONS}
+        ${ISO_BOOT_COMMON_OPTIONS})
+
+    set(ISO_BOOT_REGTEST_OPTIONS ${ISO_BOOT_REGTEST_OPTIONS}
+        -eltorito-alt-boot ${ISO_BOOT_EFI_OPTIONS}
+        ${ISO_BOOT_COMMON_OPTIONS})
+else()
+    set(ISO_BOOT_OPTIONS ${ISO_BOOT_EFI_OPTIONS} ${ISO_BOOT_COMMON_OPTIONS})
```
Extravert-ir, line 107: I don't think it makes sense to create an ISO image for non-x86 architectures. Better generate fat32 image for them right away

Extravert-ir (changes requested):

boot/boot_images.cmake
DarkFire01: They make sense for VMs and hardware with ACPI/UEFI, They're easier to debug with and say someone wanted to deploy our image on a surface RT for instance or even the dev kit Microsoft sells.

ISOs make sense and they save a lot of hassle 

I understand for making machines without UEFI/ACPI just using a fat32 image

but our generic ISOs for arm32 - arm64 need to be ISOs

boot/boot_images.cmake
HBelusca: See the one that defines the EFIP_PLATFORM_ID above in the file.

HBelusca: Indeed I may simplify this by always including the EFI boot entry (or by just checking whether EFI_PLATFORM_ID variable is defined).

boot/boot_images.cmake
HBelusca: Fixed

boot/boot_images.cmake
HBelusca: Fixed: I use `if(DEFINED EFI_PLATFORM_ID)` as the test instead.

PR #4432, opened 2022-04-07
[MSVCRT_WINETEST] Mute _CRT_NON_CONFORMING_SWPRINTFS

learn-more:


modules/rostests/winetests/msvcrt/precomp.h
```diff
@@ -6,7 +6,9 @@
 #define _INC_WINDOWS
 #define COM_NO_WINDOWS_H
 
+#if !defined(_CRT_NON_CONFORMING_SWPRINTFS)
```
HBelusca, line 9: ifndef should be shorter ^^

JoachimHenze: One day he is gonna kill you :D

learn-more: just block his github access

JoachimHenze (approved):
[build-msvc (14.1, i386, Debug)](https://github.com/reactos/reactos/runs/5876890077?check_suite_focus=true#logs)
does not longer warn.
(I also think ifndef is better, but I won't block due to that)

PR #4448, opened 2022-04-13
[TASKMGR] Application Page: Fix "Go To Process" item selection

TAN-Gaming:
## Purpose

Fix "Go To Process" selecting wrong process in the process page

### DIFF
--- a/base/applications/taskmgr/applpage.c
+++ b/base/applications/taskmgr/applpage.c
@@ -944,6 +944,8 @@ void ApplicationPage_OnGotoProcess(void)
         i = ProcGetIndexByProcessId(dwProcessId);
         if (i != -1)
         {
+            ListView_SetSelectionMark(hProcessPageListCtrl, i);
+
             ListView_SetItemState(hProcessPageListCtrl,
                                   i,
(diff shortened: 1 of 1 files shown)


base/applications/taskmgr/applpage.c
```diff
@@ -944,6 +944,8 @@ void ApplicationPage_OnGotoProcess(void)
         i = ProcGetIndexByProcessId(dwProcessId);
         if (i != -1)
         {
+            ListView_SetSelectionMark(hProcessPageListCtrl, i);
```
HBelusca, line 947: How does this differ with the `SetItemState` from below? (is there also some sort of visual difference, with and without that line?)

TAN-Gaming: Yes.

binarymaster: I cannot reproduce this problem on unpatched Task Manager in Windows 10. Need to check on XP/2003.

TAN-Gaming:
The problem maybe in another module, taskmgr works fine on Windows.

PR #4451, opened 2022-04-15
[NTUSER][IMM32_APITEST] Associate HIMC to WND

katahiromz:
## Purpose

## Proposed changes

- Set the default input context to `WND` at `IntCreateWindow` function.
- Add more tests to the `himc` testcase of `imm32_apitest`.

### DIFF
--- a/win32ss/user/ntuser/window.c
+++ b/win32ss/user/ntuser/window.c
@@ -1866,6 +1866,10 @@ PWND FASTCALL IntCreateWindow(CREATESTRUCTW* Cs,
    pWnd->ExStyle = Cs->dwExStyle;
    pWnd->cbwndExtra = pWnd->pcls->cbwndExtra;
    pWnd->pActCtx = acbiBuffer;
+
+   if (pti->spDefaultImc && Class->atomClassName != gpsi->atomSysClass[ICLS_BUTTON])
+      pWnd->hImc = UserHMGetHandle(pti->spDefaultImc);
+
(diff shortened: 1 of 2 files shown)


win32ss/user/ntuser/window.c
```diff
@@ -1866,6 +1866,10 @@ PWND FASTCALL IntCreateWindow(CREATESTRUCTW* Cs,
    pWnd->ExStyle = Cs->dwExStyle;
    pWnd->cbwndExtra = pWnd->pcls->cbwndExtra;
    pWnd->pActCtx = acbiBuffer;
+
+   if (pti->spDefaultImc && Class->atomClassName != gpsi->atomSysClass[ICLS_BUTTON])
```
HBelusca, line 1870: why button is excluded?

katahiromz:
BEFORE:
before
AFTER:
after
2 failures reduced.

win32ss/user/ntuser/window.c
katahiromz: `ImmGetContext` returns `NULL` against the buttons.
The button doesn't need the input context.

HBelusca: But this is far from being the only control that a priori doesn't need input context...

katahiromz: https://jira.reactos.org/secure/attachment/61923/CtrlContext.zip
image
The buttton is special.

PR #4454, opened 2022-04-16
[GITHUB] Use the correct worker to get MSBuild VS2019 working.

HBelusca:
## Purpose

Fix the MSVC build workers.

JIRA issue: [CORE-18146](https://jira.reactos.org/browse/CORE-18146)

## Proposed changes

```
[GITHUB] Use the correct worker to get MSBuild VS2019 working. (#4454)
CORE-18146
```

```
[GITHUB] Fix MSVC 2015 (v14.0) build, until a clear consensus is reached for PR #3872 (#4454)
CORE-18146

Use the matrix.os + include/exclude technique, as suggested
by Serge Gautherie :)
```

NOTE: aminya's msvc-dev-cmd fork (`aminya/msvc-dev-cmd@vs-version`)
supports providing an explicit VS version for vswhere, via the `vsversion` property,
thus not having to always get the latest VS.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -256,7 +256,7 @@ jobs:
 
   build-msbuild-i386:
     name: MSBuild (i386)
-    runs-on: windows-latest
+    runs-on: windows-2019
     steps:
     - name: Install Flex and Bison
       run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -78,14 +78,19 @@ jobs:
   build-msvc:
     strategy:
       matrix:
-        toolset: ['14.2', '14.1'] # VS 2019, 2017
+        os: [windows-latest, windows-2019]
+        toolset: ['14.2', '14.1', '14.0'] # VS 2019, 2017, and 2015 (see below)
         arch: [i386, amd64]
         config: [Debug, Release]
-        include:
-        - arch: i386 # Not compiling on amd64 prompt
-          toolset: '14.0' # VS 2015
+        exclude: # VS 2019, 2017 only with windows-latest; VS 2015 only with windows-2019
+          - os: windows-2019
+            toolset: '14.2'
+          - os: windows-2019
+            toolset: '14.1'
+          - os: windows-latest
+            toolset: '14.0'
       fail-fast: false
-    runs-on: windows-latest
+    runs-on: ${{matrix.os}}
     steps:
     - name: Install ninja
       run: choco install -y ninja
@@ -115,7 +120,7 @@ jobs:
     - name: Configure
       run: cmake -S src -B build -G Ninja -DCMAKE_TOOLCHAIN_FILE:FILEPATH=toolchain-msvc.cmake -DARCH:STRING=${{matrix.arch}} -DCMAKE_BUILD_TYPE=${{matrix.config}} -DENABLE_ROSTESTS=1 -DENABLE_ROSAPPS=1
     - name: Build
-      run: cmake --build build  -- -k0
+      run: cmake --build build -- -k0
     - name: Generate ISOs
       run: cmake --build build --target bootcd --target livecd
     - name: Upload ISOs


binarymaster:
Ticket id: [CORE-18146](https://jira.reactos.org/browse/CORE-18146)

.github/workflows/build.yml
```diff
@@ -132,14 +129,69 @@ jobs:
...
+        arch: amd64
+        toolset: ${{matrix.toolset}}
+    - name: Source checkout
+      uses: actions/checkout@v2
+      with:
+        path: src
+    - name: Configure
+      run: cmake -S src -B build -G Ninja -DCMAKE_TOOLCHAIN_FILE:FILEPATH=toolchain-msvc.cmake -DARCH:STRING=${{matrix.arch}} -DCMAKE_BUILD_TYPE=${{matrix.config}} -DENABLE_ROSTESTS=1 -DENABLE_ROSAPPS=1
+    - name: Build
+      run: cmake --build build  -- -k0
```
SergeGautherie, line 170: (Formatting: Remove double space. And at old line 118 too?)

.github/workflows/build.yml
```diff
@@ -132,14 +129,69 @@ jobs:
         name: reactos-syms-msvc${{matrix.toolset}}-${{matrix.arch}}-${{matrix.config}}-${{github.sha}}
         path: build/msvc_pdb
 
+  build-msvc_win2019:
+    runs-on: windows-2019
+    strategy:
+      matrix:
+        toolset: ['14.0'] # VS 2015
+        # vsversion: ['2019'] # Use VS2019 with older toolset
+        arch: [i386, amd64]
```
SergeGautherie, line 138: amd64 jobs still fail: why add it, yet?

binarymaster: It seems amd64 build has been broken in 93d8a1b7b.

HBelusca: I wanted to test whether amd64 works, and it generally works except somewhere for the generation of the 16-bit asm files....

HBelusca: it's now fixed.

binarymaster (approved):
:white_check_mark: All green

.github/workflows/build.yml
```diff
@@ -132,14 +129,69 @@ jobs:
         name: reactos-syms-msvc${{matrix.toolset}}-${{matrix.arch}}-${{matrix.config}}-${{github.sha}}
         path: build/msvc_pdb
 
+  build-msvc_win2019:
```
learn-more, line 132: Why the new builder?

HBelusca: The windows-2019 one seems to be the only one with an VS installation that contains the v14.0 build tools. If you replace the `runs-on:` with windows-latest (equivalently, that jobs section is merged back with the previous `build-msvc` one), the VS installation on that windows-latest doesn't have the v14.0 build tools and you get the builders errors that we currently have in master, see example below:

image

learn-more: Ah so this one is split off from the other win builder that used to do all 3?

HBelusca: Exactly. And I tried before to see whether, in the original `build-msvc` job, I could override the `runs-on` property (I tried by re-specifying a new `runs-on: windows-2019` in the matrix specification where the VS2015/v14.0 build-tools were specified), but without success. The job always went to the windows-latest runner...

learn-more (approved):

.github/workflows/build.yml
SergeGautherie: > override the `runs-on` property
> without success

Odd: `matrix.os` worked for me last time I used it...

HBelusca: Do you have an example of that @SergeGautherie ?

.github/workflows/build.yml
```diff
@@ -256,7 +256,7 @@ jobs:
 
   build-msbuild-i386:
     name: MSBuild (i386)
-    runs-on: windows-latest
+    runs-on: windows-2019
```
SergeGautherie, line 311: Add a comment to explain why `windows-2019` only?
Can't this job (be made to) work with VS 2022?

SergeGautherie: Ah, msvc jobs use VS 2022 but not its toolset `14.3` yet. Then, fine as is, for the time being.

HBelusca: it's similar to the 14.0 toolset problem, but different:
```
CMake Error at CMakeLists.txt:17 (project):
  Generator

    Visual Studio 16 2019
```

SergeGautherie: > `Visual Studio 16 2019`

Sure: from old line 279.
Trying to force `14.2` toolset here seems overkill, hence fine with your "image downgrade" fix.

HBelusca: iirc it was unsuccessful by just forcing an older toolset.

.github/workflows/build.yml
SergeGautherie: > example

https://github.com/SergeGautherie/reactos/commit/9435cb982f26b66d02aa3fdb99ff3392e27f70e3

HBelusca: The `matrix.os` suggestion works great together with the `include`/`exclude` trick.

binarymaster (approved):
It looks even better now :)

.github/workflows/build.yml
```diff
@@ -78,14 +78,19 @@ jobs:
...
-        include:
-        - arch: i386 # Not compiling on amd64 prompt
-          toolset: '14.0' # VS 2015
+        exclude: # VS 2019, 2017 only with windows-latest; VS 2015 only with windows-2019
+          - toolset: '14.2'
+            os: windows-2019
+          - toolset: '14.1'
+            os: windows-2019
+          - toolset: '14.0'
+            os: windows-latest
```
SergeGautherie, line 91: Nice!

```suggestion
          - os: windows-latest
            toolset: '14.0'
          - os: windows-2019
            toolset: '14.2'
          - os: windows-2019
            toolset: '14.1'
```

SergeGautherie (approved):

PR #4458, opened 2022-04-18
[NTOS:SE] Add token debug code

GeoB99:
Implement initial token debug code. For now debug information that is being tracked are: process image file name, process and thread client IDs and token creation method. More specific debug code can be added later only if needed.

As for the token creation method, this follows the same principle as on Windows where the creation method is defined by a value denoting the first letter of the said method of creation. That is, `0xC` is for token creation, `0xD` is for token duplication and `0xF` is for token filtering.


ntoskrnl/se/token.c
```diff
@@ -1086,6 +1086,17 @@ SepDuplicateToken(
         goto Quit;
     }
 
+    /* Fill in token debug information */
+#if DBG
+    RtlCopyMemory(AccessToken->ProcessName,
+                  PsGetCurrentProcess()->ImageFileName,
+                  16);
```
binarymaster, line 1093:
```suggestion
                  sizeof(AccessToken->ProcessName));
```

ntoskrnl/include/internal/se.h
```diff
@@ -34,6 +34,10 @@ typedef struct _TOKEN_AUDIT_POLICY_INFORMATION
     } Policies[1];
 } TOKEN_AUDIT_POLICY_INFORMATION, *PTOKEN_AUDIT_POLICY_INFORMATION;
 
+#define TOKEN_CREATE_METHOD    0xCUL
+#define TOKEN_DUPLICATE_METHOD 0xDUL
+#define TOKEN_FILTER_METHOD    0xFUL
```
binarymaster, line 39: CUL DUL FUL! :smile: Sounds nice

ntoskrnl/se/token.c
```diff
@@ -1086,6 +1086,17 @@ SepDuplicateToken(
         goto Quit;
     }
 
+    /* Fill in token debug information */
+#if DBG
+    RtlCopyMemory(AccessToken->ProcessName,
+                  PsGetCurrentProcess()->ImageFileName,
```
HBelusca, line 1092: Regarding the bug mentioned on Mattermost, there exists the case, where tokens start to be first used in the kernel, before any process (including the "initial" kernel process) get created; therefore in that case, `PsGetCurrentProcess()` would return `NULL` and you're in trouble.

ntoskrnl/se/token.c
```diff
@@ -1844,6 +1860,43 @@ SepCreateToken(
...
+        /*
+         * Set up the process and thread CIDs.
+         *
+         * !!!IMPORTANT NOTE!!! -- ProcessCid, ThreadCid and CreateMethod
+         * are member field names taken from Windows Server 2003 SP2 checked
+         * build kernel symbols. The reason why we have to have to such names
+         * is for WinDBG debug extension commands support purposes (such as
+         * !logonsession command respectively). THESE FIELD NAMES MUST NOT BE
+         * CHANGED!!!
+         */
```
learn-more, line 1891: It makes more sense to have a comment like this in the header where it's defined?

PR #4461, opened 2022-04-20
[CSRSRV] Acquire the necessary SE_INC_BASE_PRIORITY_PRIVILEGE at shutdown

HBelusca:
## Purpose

When increasing the base priority at shutdown, acquire the necessary
SE_INC_BASE_PRIORITY_PRIVILEGE in CsrSetToShutdownPriority(),
called by CsrShutdownProcesses(), which is necessary due to the increase
of process base priority (i.e. increase of scheduling) being done.

Dedicated to George Bișoc :)

## Explanation

The Shutdown privilege is however unnecessary (or unwanted) since this
has been already dealt with by the shutdown caller. Also, it will be
needed only by the caller of NtShutdownSystem() -- that we don't call
at all here.

### DIFF
--- a/subsystems/win32/csrsrv/procsup.c
+++ b/subsystems/win32/csrsrv/procsup.c
@@ -64,8 +64,8 @@ CsrSetToNormalPriority(VOID)
  * @name CsrSetToShutdownPriority
  *
  * The CsrSetToShutdownPriority routine sets the current NT Process'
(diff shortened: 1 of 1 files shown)


GeoB99 (approved):
Can confirm Windows enables `SE_INC_BASE_PRIORITY_PRIVILEGE` in order to increase the base priority of a process, not `SE_SHUTDOWN_PRIVILEGE`.

PR #4465, opened 2022-04-21
[ASM x64] Fix UNIMPLEMENTED macro for MSVC

learn-more:
Should make tracking down issues like [CORE-17639](https://jira.reactos.org/browse/CORE-17639) easier

### DIFF
--- a/sdk/include/asm/asm.inc
+++ b/sdk/include/asm/asm.inc
@@ -189,8 +189,23 @@ data32 MACRO opcode:VARARG
     opcode
 ENDM
 
-UNIMPLEMENTED MACRO name
+UNIMPLEMENTED2 MACRO file, line, func
+    jmp UNIMPLEMENTED2_IMPL
+UNIMPLEMENTED_MSG: .ascii "WARNING:  %s at %s:%d is UNIMPLEMENTED!", 10, 0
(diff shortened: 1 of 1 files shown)


HBelusca (approved):
huhuh, interesting that for whatever obscure reason, only the GAS version was actually implemented 🤣

sdk/include/asm/asm.inc
```diff
@@ -189,8 +189,23 @@ data32 MACRO opcode:VARARG
...
+EXTERN DbgPrint:PROC
+UNIMPLEMENTED2_IMPL:
+    sub rsp, 28h
+    mov r9, line
+    lea r8, UNIMPLEMENTED_FILE
+    lea rdx, UNIMPLEMENTED_FUNC
+    lea rcx, UNIMPLEMENTED_MSG
+    call DbgPrint
+    add rsp, 28h
+    xor eax, eax
```
learn-more, line 206: The GAS version does not do this, but it seems safer to not leave DbgPrint's result in there.

HBelusca: Could then be of interest to port it back to GAS 🤔

learn-more: I don't have a gcc64 build setup, and I don't really like editing code without properly testing it.

PR #4467, opened 2022-04-22
[IMM32] Add ValidateHandle function

katahiromz:
## Purpose

`imm32_winetest` expects some OS error code. Setting it properly...

## Proposed changes

- Set the last error on failure of handle validation.


katahiromz:
BEFORE:
before
14 failures.
AFTER:
after
11 failures.

PR #4468, opened 2022-04-24
[NTUSER] Add locking at IntSendFocusMessage

katahiromz:
## Purpose
Fix access violation on `IntFocusSetInputContext`.
JIRA issue: [CORE-18163](https://jira.reactos.org/browse/CORE-18163)

## Proposed changes

- Add locking on `pWndPrev`.

### DIFF
--- a/win32ss/user/ntuser/focus.c
+++ b/win32ss/user/ntuser/focus.c
@@ -614,7 +614,7 @@ co_IntSendActivateMessages(PWND WindowPrev, PWND Window, BOOL MouseActivate, BOO
 VOID FASTCALL
 IntSendFocusMessages( PTHREADINFO pti, PWND pWnd)
 {
-   PWND pWndPrev, pRealWndPrev;
+   PWND pWndPrev;
    PUSER_MESSAGE_QUEUE ThreadQueue = pti->MessageQueue; // Queue can change...
    HWND hwndPrev;
    USER_REFERENCE_ENTRY Ref;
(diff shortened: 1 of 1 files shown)


win32ss/user/ntuser/focus.c
```diff
@@ -614,17 +614,24 @@ co_IntSendActivateMessages(PWND WindowPrev, PWND Window, BOOL MouseActivate, BOO
...
    {
       ThreadQueue->QF_flags |= QF_FOCUSNULLSINCEACTIVE;
    }
 
-   pWndPrev = ThreadQueue->spwndFocus;
+   pRealWndPrev = pWndPrev = ThreadQueue->spwndFocus;
+   if (pRealWndPrev)
+   {
+      UserRefObjectCo(pRealWndPrev, &Ref);
+      if (!ValidateHwndNoErr(UserHMGetHandle(pRealWndPrev)))
```
HBelusca, line 632: Why is that handle stuff needed? Shouldn't it be that from within running win32k code, all these stored non-NULL PWNDs are automatically valid?

katahiromz: `spwndFocus` might be invalid.

HBelusca: In what circumstances? I would expect it to be either NULL or a pointer to a valid PWND ?

katahiromz: Ideally `spwndFocus` should be assignment-locked but ReactOS has different impl.

katahiromz: This function is called from `co_UserSetFocus`.

katahiromz: The below code sending {{WM_KILLFOCUS}}.

katahiromz: This code was useless. Deleted.

katahiromz: This code was useless. Deleted.

HBelusca (commented):
If this solves the reported problem, without breaking anything else, then LGTM :)

win32ss/user/ntuser/focus.c
learn-more: `co_` functions do a callback to usermode code, so the entire world can change during those ;)

PR #4469, opened 2022-04-26
Compile ReactOS with -fno-common

Extravert-ir:
That is a default option on modern compilers


sdk/cmake/gcc.cmake
```diff
@@ -39,7 +39,8 @@ if(STACK_PROTECTOR)
 endif()
 
 # Compiler Core
-add_compile_options(-pipe -fms-extensions -fno-strict-aliasing)
+# note: -fno-common is default since GCC 10
```
HBelusca, line 42: Is it supported by clang as well?

modules/rostests/apitests/compiler/ms/seh/seh0001.c
```diff
@@ -6,7 +6,6 @@
 #include "seh.h"
 
 char test[] = "SEH0001.c";
-int perfect;
```
HBelusca, line 9: I wonder what was their idea behind this.
Do you think it's possible to send that upstream?

HBelusca (commented):
Looks nice; does the PR fix all our codebase or is there other modifications to be done?

Extravert-ir:
> Some wine code modifs that may be worth sending upstream.

Wine is already compiled like that

sdk/cmake/gcc.cmake
Extravert-ir: Yes

modules/rostests/apitests/compiler/ms/seh/seh0001.c
Extravert-ir: No, I have already spent enough time on this :P

PR #4485, opened 2022-05-04
[NTUSER][USER32] Make NtUserBuildHwndList exclusive

katahiromz:
## Purpose
Fix hung-up and memory leaks.
JIRA issue: [CORE-18173](https://jira.reactos.org/browse/CORE-18173)

## Proposed changes

- Modify the prototype of `NtUserBuildHwndList` function.
- Wrap the code by `UserEnterExclusive();` and `UserLeave();`.
- Fix calling of `NtUserBuildHwndList` in `user32.dll`.

## TODO

- [x] Do tests.

### DIFF
--- a/win32ss/user/ntuser/window.c
+++ b/win32ss/user/ntuser/window.c
@@ -1560,7 +1560,7 @@ NtUserBuildHwndList(
          {
             if (bGoDown)
             {
-               if(dwCount++ < *pcHwndNeeded && phwndList)
+               if(dwCount++ < cHwnd && phwndList)
                {
                   _SEH2_TRY
                   {
@@ -1638,7 +1638,7 @@ NtUserBuildHwndList(
             Window = ValidateHwndNoErr(List[i]);
             if (Window && Window->head.pti == W32Thread)
             {
-               if (dwCount < *pcHwndNeeded && phwndList)
+               if (dwCount < cHwnd && phwndList)
                {
                   _SEH2_TRY
                   {
--- a/win32ss/user/ntuser/window.c
+++ b/win32ss/user/ntuser/window.c
@@ -1514,10 +1514,7 @@ NtUserBuildHwndList(
    ULONG dwCount = 0;
 
    if (pcHwndNeeded == NULL)
-   {
-       UserLeave();
        return ERROR_INVALID_PARAMETER;
-   }
 
    UserEnterExclusive();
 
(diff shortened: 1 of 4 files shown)


win32ss/user/ntuser/window.c
```diff
@@ -1511,8 +1511,13 @@ NtUserBuildHwndList(
    NTSTATUS Status;
    ULONG dwCount = 0;
 
+   UserEnterExclusive();
```
HBelusca, line 1514: Could be moved after the `pBufSize` check.

win32ss/user/ntuser/window.c
```diff
@@ -1523,7 +1528,8 @@ NtUserBuildHwndList(
       {
          if(hDesktop == NULL && !(Desktop = IntGetActiveDesktop()))
          {
-            return ERROR_INVALID_HANDLE;
+            Status = ERROR_INVALID_HANDLE;
```
HBelusca, line 1531: Should we call the `SetLastNtError` in case of error for these? (basically all those you've modified).

win32ss/user/ntuser/window.c
```diff
@@ -1511,8 +1511,13 @@ NtUserBuildHwndList(
    NTSTATUS Status;
    ULONG dwCount = 0;
 
+   UserEnterExclusive();
+
    if (pBufSize == 0)
```
HBelusca, line 1516:
```suggestion
   if (pBufSize == NULL)
```

win32ss/user/ntuser/window.c
```diff
@@ -1504,7 +1504,7 @@ NtUserBuildHwndList(
    HWND hwndParent,
    BOOLEAN bChildren,
    ULONG dwThreadId,
-   LPARAM lParam,
+   ULONG lParam,
```
HBelusca, line 1507: yes indeed it's actually not an lparam.

win32ss/user/ntuser/window.c
katahiromz: OK.

win32ss/user/ntuser/window.c
katahiromz: Moved `SetLastNtError` call to the end.

win32ss/user/ntuser/window.c
katahiromz: OK.

PR #4488, opened 2022-05-05
[SHELL32] Fix of visual glitch problem after CORE-18137 shell32 fix

RaymondCzerny:
## Purpose

Adding the Control Panel applet to the taskbar caused a visual glitch in the installation wizard via the RunDLL window 

JIRA issue: [CORE-18175](https://jira.reactos.org/browse/CORE-18175)

## Proposed changes

Implemented a check if the parent window of RunDLL is the desktop to add the system control applet to the taskbar only if it is.

In case of a problem, the installer window is the parent window of RunDLL, so now the RunDLL window is not manipulated.

### DIFF
--- a/dll/win32/shell32/wine/control.c
+++ b/dll/win32/shell32/wine/control.c
@@ -819,6 +819,7 @@ Control_EnumWinProc(
 static void
 Control_ShowAppletInTaskbar(CPlApplet* applet, UINT index)
 {
+    HICON hSmallIcon;
     ITaskbarList* pTaskbar = NULL;
 
     /* Try to add a taskbar button only if the applet's parent window is the desktop */
@@ -828,9 +829,22 @@ Control_ShowAppletInTaskbar(CPlApplet* applet, UINT index)
     }
 
     SetWindowTextW(applet->hWnd, applet->info[index].name);
-    if (applet->info[index].icon)
+
+    /* try loading the 16x16 icon for the taskbar button */
+    hSmallIcon = (HICON)LoadImageW(applet->hModule,
+                                   MAKEINTRESOURCEW(applet->info[index].idIcon),
+                                   IMAGE_ICON,
+                                   16, 16, 0);
+    if (hSmallIcon)
     {
-        SendMessageW(applet->hWnd, WM_SETICON, ICON_SMALL, (LPARAM)applet->info[index].icon);
+        SendMessageW(applet->hWnd, WM_SETICON, ICON_SMALL, (LPARAM)hSmallIcon);
+    }
+    else
+    {
+        if (applet->info[index].icon)
+        {
+            SendMessageW(applet->hWnd, WM_SETICON, ICON_SMALL, (LPARAM)applet->info[index].icon);
+        }
     }
 
     /* Add button to the taskbar */
--- a/dll/win32/shell32/wine/control.c
+++ b/dll/win32/shell32/wine/control.c
@@ -830,11 +830,13 @@ Control_ShowAppletInTaskbar(CPlApplet* applet, UINT index)
 
     SetWindowTextW(applet->hWnd, applet->info[index].name);
 
-    /* try loading the 16x16 icon for the taskbar button */
+    /* Try loading the small icon for the taskbar button */
     hSmallIcon = (HICON)LoadImageW(applet->hModule,
                                    MAKEINTRESOURCEW(applet->info[index].idIcon),
                                    IMAGE_ICON,
-                                   16, 16, 0);
+                                   GetSystemMetrics(SM_CXSMICON),
+                                   GetSystemMetrics(SM_CYSMICON),
+                                   0);
     if (hSmallIcon)
     {
         SendMessageW(applet->hWnd, WM_SETICON, ICON_SMALL, (LPARAM)hSmallIcon);
--- a/dll/win32/shell32/wine/control.c
+++ b/dll/win32/shell32/wine/control.c
@@ -821,26 +821,29 @@ Control_ShowAppletInTaskbar(CPlApplet* applet, UINT index)
 {
     ITaskbarList* pTaskbar = NULL;
 
-    SetWindowTextW(applet->hWnd, applet->info[index].name);
-    SendMessageW(applet->hWnd, WM_SETICON, ICON_SMALL, (LPARAM)applet->info[index].icon);
-
-    /* Add button to the taskbar */
-    ShowWindow(applet->hWnd, SW_SHOWMINNOACTIVE);
-
-    /* Activate the corresponding button in the taskbar */
-    CoInitialize(NULL);
-    if (CoCreateInstance(&CLSID_TaskbarList,
-                         NULL, CLSCTX_INPROC_SERVER,
-                         &IID_ITaskbarList,
-                         (LPVOID*)&pTaskbar) == S_OK)
+    if (GetParent(applet->hWnd) == NULL) // Is desktop the parent window?
     {
-        if (ITaskbarList_HrInit(pTaskbar) == S_OK)
+        SetWindowTextW(applet->hWnd, applet->info[index].name);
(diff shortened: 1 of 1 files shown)


HBelusca (commented):
Please show a screenshot of how this looks when:
1. during 2nd-stage installation, you open the keyboard settings (for example, in order to change keyboard layout);
2. during normal usage, how it looks in the shell;
3. when you force-kill explorer.exe (and it's not restarted), and you open a control panel (use taskmgr for that).

RaymondCzerny:
> Please show a screenshot of how this looks when:
> 
> 1. during 2nd-stage installation, you open the keyboard settings (for example, in order to change keyboard layout);
Setup2_keyboard

> 2. during no
normal
rmal usage, how it looks in the shell;

> 3. when you force-kill explorer.exe (and it's not restarted), and you open a control panel (use taskmgr for that).
noexplorer

dll/win32/shell32/wine/control.c
```diff
@@ -821,26 +821,29 @@ Control_ShowAppletInTaskbar(CPlApplet* applet, UINT index)
...
-    ShowWindow(applet->hWnd, SW_SHOWMINNOACTIVE);
-
-    /* Activate the corresponding button in the taskbar */
-    CoInitialize(NULL);
-    if (CoCreateInstance(&CLSID_TaskbarList,
-                         NULL, CLSCTX_INPROC_SERVER,
-                         &IID_ITaskbarList,
-                         (LPVOID*)&pTaskbar) == S_OK)
+    if (GetParent(applet->hWnd) == NULL) // Is desktop the parent window?
     {
```
HBelusca, line 825: By the way, why is it necessary to call the `SetWindowTextW` function? In principle the applet's window will set its title, so it should not be force-changed manually like this.
Besides, the set-icon call below can be ill-defined. For example, the "icon" handle can be NULL:
image

dll/win32/shell32/wine/control.c
```diff
@@ -821,26 +821,29 @@ Control_ShowAppletInTaskbar(CPlApplet* applet, UINT index)
...
-        if (ITaskbarList_HrInit(pTaskbar) == S_OK)
+        SetWindowTextW(applet->hWnd, applet->info[index].name);
+        SendMessageW(applet->hWnd, WM_SETICON, ICON_SMALL, (LPARAM)applet->info[index].icon);
+
```
HBelusca, line 828: You may be interested in investigating
https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-itaskbarlist-addtab

RaymondCzerny: About the idea of using the method ITaskbarList::AddTab:

What happens if the explorer crashes and restarts automatically?
Then with this method the switch in the taskbar is gone.

With my method, the switches of all open applications including the system control applets are restored.

I can implement the idea to use the return value of ITaskbarList_HrInit as condition to add.

HBelusca: > What happens if the explorer crashes and restarts automatically?
Then with this method the switch in the taskbar is gone.

Ah yes, good point.

RaymondCzerny: > > What happens if the explorer crashes and restarts automatically?
> > Then with this method the switch in the taskbar is gone.
> 
> Ah yes, good point. But the same sort of observation could be said about the following: with this PR, start a control panel while explorer is not running, then start the explorer. Would the applet show in the taskbar? How does this behave on Windows?

Yes, and Win 7 also reconstructs the switch.

I have uploaded a video. [CORE-18175](https://jira.reactos.org/browse/CORE-18175)

RaymondCzerny: The idea to use the return value of ITaskbarList_HrInit is not so good after all.

The following scenario:
- Taskbar exists
- an application uses a system control applet like the installer. (application window = RunDLL parent window)

Result: visual glitch in application, but no button in taskbar

dll/win32/shell32/wine/control.c
RaymondCzerny: Without these function calls it looks like this.
nocall

dll/win32/shell32/wine/control.c
```diff
@@ -821,26 +821,29 @@ Control_ShowAppletInTaskbar(CPlApplet* applet, UINT index)
...
-    /* Add button to the taskbar */
-    ShowWindow(applet->hWnd, SW_SHOWMINNOACTIVE);
-
-    /* Activate the corresponding button in the taskbar */
-    CoInitialize(NULL);
-    if (CoCreateInstance(&CLSID_TaskbarList,
-                         NULL, CLSCTX_INPROC_SERVER,
-                         &IID_ITaskbarList,
-                         (LPVOID*)&pTaskbar) == S_OK)
+    if (GetParent(applet->hWnd) == NULL) // Is desktop the parent window?
```
HBelusca, line 824: Nit: In order to avoid the useless indentation change, you could do:
```suggestion
    /* Try to add a taskbar button only if the applet's parent window is the desktop */
    if (GetParent(applet->hWnd) != NULL)
        return;
```
(then follows the original unchanged code).

dll/win32/shell32/wine/control.c
RaymondCzerny: In case of an empty icon handle, default icon for taskbar is used automatically.

dll/win32/shell32/wine/control.c
RaymondCzerny: Yes

HBelusca (approved):

binarymaster (approved):
Looks mostly good, see comments below. Yes, it does fix the problem for me.

image image

Here is something to improve, maybe in other PR:
- The icon on the taskbar randomly (but in most cases) is not being set
- If the icon is set, it's using 32x32 icon instead of small 16x16, see comparison with the window caption icon

image

image

RaymondCzerny:
I have tried to load the 16x16 icon with LoadImageW to use it for RunDLL.

However, this does not change the initial misrepresentation on the taskbar button. This is a problem with the taskbar.

binarymaster:
> This is a problem with the taskbar.

However such programs as Explorer or RAPPS do set a small icon correctly on the taskbar. Ok, I think this can be improved later.

dll/win32/shell32/wine/control.c
```diff
@@ -819,10 +819,33 @@ Control_EnumWinProc(
...
+    }
+
     SetWindowTextW(applet->hWnd, applet->info[index].name);
-    SendMessageW(applet->hWnd, WM_SETICON, ICON_SMALL, (LPARAM)applet->info[index].icon);
+
+    /* try loading the 16x16 icon for the taskbar button */
+    hSmallIcon = (HICON)LoadImageW(applet->hModule,
+                                   MAKEINTRESOURCEW(applet->info[index].idIcon),
+                                   IMAGE_ICON,
+                                   16, 16, 0);
```
binarymaster, line 837:
```suggestion
                                   IMAGE_ICON,
                                   GetSystemMetrics(SM_CXSMICON),
                                   GetSystemMetrics(SM_CYSMICON),
                                   0);
```

dll/win32/shell32/wine/control.c
```diff
@@ -819,10 +819,33 @@ Control_EnumWinProc(
...
+    /* Try to add a taskbar button only if the applet's parent window is the desktop */
+    if (GetParent(applet->hWnd) != NULL)
+    {
+        return;
+    }
+
     SetWindowTextW(applet->hWnd, applet->info[index].name);
-    SendMessageW(applet->hWnd, WM_SETICON, ICON_SMALL, (LPARAM)applet->info[index].icon);
+
+    /* try loading the 16x16 icon for the taskbar button */
```
binarymaster, line 833:
```suggestion
    /* Try loading the small icon for the taskbar button */
```

PR #4489, opened 2022-05-05
[SHELL32_APITEST] ShellExecCmdLine: Do **NOT** use TerminateProcess to close the opened windows.

HBelusca:
## Purpose

This design, introduced in 418edcd2b, is fundamentally flawed as it
can also close windows unrelated to the running test (e.g. windows
of programs that the user can start, while the test is running).

But since we cannot do much better, mitigate instead the other main
problem of this design: Just use PostMessageW(WM_CLOSE), as it used
to be, instead of TerminateProcess().

Indeed, using TerminateProcess() otherwise could kill unrelated
processes the test hasn't created. In particular it could kill the
csrss.exe system process if, during the testing procedure, a hard-error
popup shows up.
And this is precisely the case when this test runs with limited memory,
and a hard-error
  "Insufficient system resources exist to complete the requested service."
arises.

## Proposed changes

Effectively partly revert commit 418edcd2b.

### DIFF
--- a/modules/rostests/apitests/shell32/ShellExecCmdLine.cpp
+++ b/modules/rostests/apitests/shell32/ShellExecCmdLine.cpp
@@ -609,13 +609,7 @@ static void CleanupNewlyCreatedWindows(void)
             }
         }
         if (!bFound)
-        {
-            DWORD dwPID;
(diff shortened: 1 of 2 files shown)


katahiromz (approved):

HBelusca:
> All the newly-opened windows were closed?

As far as I could see, yes.

PR #4491, opened 2022-05-05
[PCIX] DriverEntry(): Check PciOpenKey(L"Parameters", ...) result

SergeGautherie:
## Purpose

Minimal safety, until investigated/fixed further.
While here, move 2 ZwClose() where they belong.

Addendum to 29f98a6 (r48851).
JIRA issue: [CORE-7056](https://jira.reactos.org/browse/CORE-7056)

## TODO

I assume next step would be to actually resume work on:
```
- Fix PciGetHackFlags for setup found by sir_richard "Setup currently doesn't have a correct registry"
- Fix DriverEntry for setup like PciGetHackFlags
```

### DIFF
--- a/drivers/bus/pcix/init.c
+++ b/drivers/bus/pcix/init.c
@@ -712,8 +712,6 @@ DriverEntry(IN PDRIVER_OBJECT DriverObject,
 
     /* Setup initial loop variables */
     KeyHandle = NULL;
-    ParametersKey = NULL;
-    DebugKey = NULL;
     ControlSetKey = NULL;
     do
     {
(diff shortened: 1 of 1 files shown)


drivers/bus/pcix/init.c
```diff
@@ -869,8 +876,6 @@ DriverEntry(IN PDRIVER_OBJECT DriverObject,
     /* Close all opened keys, return driver status to PnP Manager */
     if (KeyHandle) ZwClose(KeyHandle);
     if (ControlSetKey) ZwClose(ControlSetKey);
-    if (ParametersKey) ZwClose(ParametersKey);
-    if (DebugKey) ZwClose(DebugKey);
```
GeoB99, line 873: And what's the point of moving two `ZwClose` calls whilst keeping others? What is the outcome benefit of doing this?

PR #4495, opened 2022-05-07
[NTOS:KE] Fix CPU family and model detection

BUGinator86:
## Purpose

Fix CPU-Revision and Type

JIRA issue: [CORE-17974](https://jira.reactos.org/browse/CORE-17974)

use correct amount of nibbles to determine CPU-Revision and Type

### DIFF
--- a/ntoskrnl/ke/amd64/cpu.c
+++ b/ntoskrnl/ke/amd64/cpu.c
@@ -98,7 +98,11 @@ KiSetProcessorType(VOID)
 {
     CPU_INFO CpuInfo;
     CPU_SIGNATURE CpuSignature;
-    ULONG Stepping, Type;
+    BOOLEAN ExtendModel;
+    ULONG Stepping, Type, Vendor;
+
+    /* This initializes Prcb->CpuVendor */
+    Vendor = KiGetCpuVendor();
 
     /* Do CPUID 1 now */
     KiCpuId(&CpuInfo, 1);
@@ -111,8 +115,29 @@ KiSetProcessorType(VOID)
      */
     CpuSignature.AsULONG = CpuInfo.Eax;
     Stepping = CpuSignature.Model;
+    ExtendModel = (CpuSignature.Family == 15);
+#if ( (NTDDI_VERSION >= NTDDI_WINXPSP2) && (NTDDI_VERSION < NTDDI_WS03) ) || (NTDDI_VERSION >= NTDDI_WS03SP1)
+    if (CpuSignature.Family == 6)
+    {
+        ExtendModel |= (Vendor == CPU_INTEL);
+#if (NTDDI_VERSION >= NTDDI_WIN8)
+        ExtendModel |= (Vendor == CPU_CENTAUR);
+#endif
+    }
+#endif
+    if (ExtendModel)
+    {
+        /* Add ExtendedModel to distinguish from non-extended values. */
+        Stepping |= (CpuSignature.ExtendedModel << 4);
+    }
     Stepping = (Stepping << 8) | CpuSignature.Step;
     Type = CpuSignature.Family;
+    if (CpuSignature.Family == 15)
+    {
+        /* Add ExtendedFamily to distinguish from non-extended values.
+         * It must not be larger than 0xF0 to avoid overflow. */
+        Type += min(CpuSignature.ExtendedFamily, 0xF0);
+    }
 
     /* Save them in the PRCB */
     KeGetCurrentPrcb()->CpuID = TRUE;
@@ -130,7 +155,7 @@ KiGetFeatureBits(VOID)
     CPU_INFO CpuInfo;
 
     /* Get the Vendor ID */
-    Vendor = KiGetCpuVendor();
+    Vendor = Prcb->CpuVendor;
 
     /* Make sure we got a valid vendor ID at least. */
     if (!Vendor) return FeatureBits;
--- a/ntoskrnl/ke/amd64/cpu.c
+++ b/ntoskrnl/ke/amd64/cpu.c
@@ -35,6 +35,21 @@ static const CHAR CmpIntelID[]       = "GenuineIntel";
 static const CHAR CmpAmdID[]         = "AuthenticAMD";
 static const CHAR CmpCentaurID[]     = "CentaurHauls";
 
+typedef union _CPU_SIGNATURE
+{
+    struct
+    {
+        ULONG Step : 4;
+        ULONG Model : 4;
+        ULONG Family : 4;
+        ULONG Unused : 4;
+        ULONG ExtendedModel : 4;
+        ULONG ExtendedFamily : 8;
+        ULONG Unused2 : 4;
+    };
+    ULONG AsULONG;
+} CPU_SIGNATURE;
+
 /* FUNCTIONS *****************************************************************/
 
 ULONG
@@ -82,6 +97,7 @@ NTAPI
 KiSetProcessorType(VOID)
 {
     CPU_INFO CpuInfo;
+    CPU_SIGNATURE CpuSignature;
     ULONG Stepping, Type;
 
     /* Do CPUID 1 now */
@@ -89,17 +105,14 @@ KiSetProcessorType(VOID)
 
     /*
      * Get the Stepping and Type. The stepping contains both the
-     * Model and the Step, while the Type contains the returned Type.
-     * We ignore the family.
+     * Model and the Step, while the Type contains the returned Family.
      *
      * For the stepping, we convert this: zzzzzzxy into this: x0y
      */
-    Stepping = CpuInfo.Eax & 0xF0;
-    Stepping <<= 4;
-    Stepping += (CpuInfo.Eax & 0xFF);
-    Stepping &= 0xF0F;
-    Type = CpuInfo.Eax & 0xF00;
-    Type >>= 8;
+    CpuSignature.AsULONG = CpuInfo.Eax;
+    Stepping = CpuSignature.Model;
+    Stepping = (Stepping << 8) | CpuSignature.Step;
+    Type = CpuSignature.Family;
 
     /* Save them in the PRCB */
     KeGetCurrentPrcb()->CpuID = TRUE;
--- a/ntoskrnl/ke/amd64/cpu.c
+++ b/ntoskrnl/ke/amd64/cpu.c
@@ -37,36 +37,6 @@ static const CHAR CmpCentaurID[]     = "CentaurHauls";
 
 /* FUNCTIONS *****************************************************************/
 
-VOID
-NTAPI
-KiSetProcessorType(VOID)
-{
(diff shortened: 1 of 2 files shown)


ntoskrnl/ke/amd64/cpu.c
```diff
@@ -52,14 +52,16 @@ KiSetProcessorType(VOID)
...
     Stepping = CpuInfo.Eax & 0xF0;
     Stepping <<= 4;
-    Stepping += (CpuInfo.Eax & 0xFF);
-    Stepping &= 0xF0F;
+    Stepping += ((CpuInfo.Eax & 0x000F0000) >> 4);
+    Stepping += (CpuInfo.Eax & 0xF);
     Type = CpuInfo.Eax & 0xF00;
     Type >>= 8;
+    Type += ((CpuInfo.Eax & 0x0FF00000) >> 16);
+
```
HBelusca, line 64:
```suggestion
```

HBelusca (approved):

ntoskrnl/ke/i386/cpu.c
```diff
@@ -101,14 +101,15 @@ KiSetProcessorType(VOID)
...
      */
     Stepping = CpuInfo.Eax & 0xF0;
     Stepping <<= 4;
-    Stepping += (CpuInfo.Eax & 0xFF);
-    Stepping &= 0xF0F;
+    Stepping += ((CpuInfo.Eax & 0x000F0000) >> 4);
+    Stepping += (CpuInfo.Eax & 0xF);
     Type = CpuInfo.Eax & 0xF00;
     Type >>= 8;
+    Type += ((CpuInfo.Eax & 0x0FF00000) >> 16);
```
ThFabba, line 112: The top of the three nibbles is going to be ignored since we store this in an UCHAR. I would suggest making that explicit by masking out those bits, as well as adding a comment noting that there are additional family bits but we can't make use of them.

ThFabba: Actually the way Windows encodes these seems to be:
If the "family" (0xf00) value is less than 15, use only that value
If it is 15, use "extended family" (0x0ff00000) but add 15 to distinguish it from non-extended values.

ntoskrnl/ke/amd64/cpu.c
```diff
@@ -52,14 +52,15 @@ KiSetProcessorType(VOID)
...
      *
-     * For the stepping, we convert this: zzzzzzxy into this: x0y
+     * For the stepping, we convert this: zzzxzzxy into this: xx0y
      */
     Stepping = CpuInfo.Eax & 0xF0;
     Stepping <<= 4;
-    Stepping += (CpuInfo.Eax & 0xFF);
-    Stepping &= 0xF0F;
+    Stepping += ((CpuInfo.Eax & 0x000F0000) >> 4);
+    Stepping += (CpuInfo.Eax & 0xF);
```
SergeGautherie, line 60: Nit pick:
```suggestion
    Stepping |= (CpuInfo.Eax & 0x0F);
```
While here, everywhere, `|=` seems more consistent.
Are the outer parentheses really needed?

ThFabba: Guidance on nit picks: if they're the only thing you have to contribute to a PR, stay quiet.

ntoskrnl/ke/i386/cpu.c
HBelusca: Mentioning this reference (URL) in either the commit log and/or the comment in source code would be nice :)

jeditobe:
https://jira.reactos.org/browse/CORE-17974

binarymaster (dismissed):
Playing with bits is fun of course, but we can make it a lot more readable by using a bitfield structure. As already pointed out by @ThFabba in review comment above, the algorithm has some subtleties.

I would suggest doing it this way:
- In first commit, add the structure and update the code to use it without changing the algorithm
- In second commit, update the algorithm itself

The result would be a lot more clean and readable.

Here is an example of bitfield structure: https://git.reactos.org/?p=reactos.git;a=blob;f=boot/freeldr/freeldr/include/arch/pc/pcbios.h;hb=a84912594754b52343d5cac0426adbaba764d2bf#l43

BUGinator86:
> Playing with bits is fun of course, but we can make it a lot more readable by using a bitfield structure. As already pointed out by @ThFabba in review comment above, the algorithm has some subtleties.
> 
> I would suggest doing it this way:
> 
> * In first commit, add the structure and update the code to use it without changing the algorithm
> * In second commit, update the algorithm itself
> 
> The result would be a lot more clean and readable.
> 
> Here is an example of bitfield structure: https://git.reactos.org/?p=reactos.git;a=blob;f=boot/freeldr/freeldr/include/arch/pc/pcbios.h;hb=a84912594754b52343d5cac0426adbaba764d2bf#l43

With the recent change a local struct with a union is used to store the the CPU-Info.
I also added the ticket in the comment as reference for the algorithm used. If the site changes its content or is down we can still see the quote from @HBelusca for reference if the implementation is/was right. :)

ntoskrnl/ke/amd64/cpu.c
```diff
@@ -35,37 +35,54 @@ static const CHAR CmpIntelID[]       = "GenuineIntel";
...
+        {
+            USHORT SteppingAsUSHORT;
+            struct
+            {
+                UCHAR Stepping;
+                UCHAR Model;
+            };
+        };
+        UCHAR Family;
+    } CPU_REVISION;
```
binarymaster, line 63: This structure is not a bitfield structure :sweat_smile: You can make it bitfield for the CPUID EAX register, and then you will not need to use the macros.

BUGinator86: Ah, I think I got it now. :)
Updated the PR.

ntoskrnl/ke/amd64/cpu.c
```diff
@@ -35,36 +35,68 @@ static const CHAR CmpIntelID[]       = "GenuineIntel";
...
+            {
+                UCHAR Stepping : 4;
+                UCHAR Model : 4;
+                UCHAR Family : 4;
+                UCHAR Unused : 4;
+                UCHAR ModelExt : 4;
+                UCHAR FamilyExt;
+            };
+        };
+    } CPUINFO_EAX;
```
binarymaster, line 62: Yes, now that's it :slightly_smiling_face: I think I'm going to help with rearranging commits here, and then it will be good to go.

binarymaster (approved):
I think it looks perfect now. Ping reviewers to take one more look before merging.

ntoskrnl/ke/amd64/cpu.c
```diff
@@ -35,31 +35,68 @@ static const CHAR CmpIntelID[]       = "GenuineIntel";
...
+    if (ExtendModel)
+    {
+        /* Add ExtendedModel to distinct from non-extended values. */
+        Stepping |= (CpuSignature.ExtendedModel << 4);
+    }
+    Stepping = (Stepping << 8) | CpuSignature.Step;
+    Type = CpuSignature.Family;
+    if (CpuSignature.Family == 15)
+    {
+        /* Add ExtendedFamily to distinct from non-extended values.
```
HBelusca, line 95:
```suggestion
        /* Add ExtendedFamily to distinguish from non-extended values.
```

ntoskrnl/ke/i386/cpu.c
```diff
@@ -91,24 +110,46 @@ NTAPI
...
+    if (ExtendModel)
+    {
+        /* Add ExtendedModel to distinct from non-extended values. */
+        Stepping |= (CpuSignature.ExtendedModel << 4);
+    }
+    Stepping = (Stepping << 8) | CpuSignature.Step;
+    Type = CpuSignature.Family;
+    if (CpuSignature.Family == 15)
+    {
+        /* Add ExtendedFamily to distinct from non-extended values.
```
HBelusca, line 148:
```suggestion
        /* Add ExtendedFamily to distinguish from non-extended values.
```

ntoskrnl/ke/i386/cpu.c
```diff
@@ -91,24 +110,46 @@ NTAPI
...
+        ULONG Vendor = KiGetCpuVendor();
+        ExtendModel |= (Vendor == CPU_INTEL);
+#if (NTDDI_VERSION >= NTDDI_WIN8)
+        ExtendModel |= (Vendor == CPU_CENTAUR);
+#endif
+    }
+#endif
+    if (ExtendModel)
+    {
+        /* Add ExtendedModel to distinct from non-extended values. */
```
HBelusca, line 141:
```suggestion
        /* Add ExtendedModel to distinguish from non-extended values. */
```

ntoskrnl/ke/i386/cpu.c
```diff
@@ -91,24 +110,46 @@ NTAPI
...
-    Stepping &= 0xF0F;
-    Type = CpuInfo.Eax & 0xF00;
-    Type >>= 8;
+    CpuSignature.AsULONG = CpuInfo.Eax;
+    Stepping = CpuSignature.Model;
+    ExtendModel = (CpuSignature.Family == 15);
+#if ( (NTDDI_VERSION >= NTDDI_WINXPSP2) && (NTDDI_VERSION < NTDDI_WS03) ) || (NTDDI_VERSION >= NTDDI_WS03SP1)
+    if (CpuSignature.Family == 6)
+    {
+        ULONG Vendor = KiGetCpuVendor();
```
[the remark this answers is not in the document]
binarymaster: Yes, it's amd64-specific, x86 doesn't have it, I remember we discussed it somewhere earlier.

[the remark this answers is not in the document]
HBelusca: Ah yes, a `CpuVendor` field has been added to the x86 KPRCB starting Vista.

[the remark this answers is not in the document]
binarymaster: But I don't remember if `Prcb->CpuVendor` is already initialized at this point of time. Needs to be explicitly confirmed.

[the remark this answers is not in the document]
binarymaster: Indeed `Prcb->CpuVendor` was not initialized in function `KiSetProcessorType` on amd64. I have corrected the code to fix it.

ntoskrnl/ke/amd64/cpu.c
```diff
@@ -117,7 +153,7 @@ KiGetFeatureBits(VOID)
     CPU_INFO CpuInfo;
 
     /* Get the Vendor ID */
-    Vendor = KiGetCpuVendor();
+    Vendor = Prcb->CpuVendor;
```
HBelusca, line 156: Here, it is already initialized?

binarymaster: Yes, here it is. I think I'm going to add a comment on the first call.

PR #4496, opened 2022-05-07
[WIN32K:NTUSER] Assign a security descriptor when parsing the desktop

GeoB99:
The problem ReactOS currently faces is this -- whenever the desktop is being parsed we aren't assigning a security descriptor to it. As a matter of fact when Winlogon tries to assign new security information to the application desktop when a user logs in, Winlogon fails because no prior descriptor has been created for it even though we already do this when initializing security buffers in Winlogon.

With that said, we must assign a descriptor when parsing the desktop as well. This fixes a hack in Winlogon where security assigning of application desktop during a log in is disabled (which we can now enable such code path back).

### DIFF
--- a/win32ss/user/ntuser/security.c
+++ b/win32ss/user/ntuser/security.c
@@ -237,6 +237,72 @@ IntQueryUserSecurityIdentification(
 
 /**
  * @brief
+ * Assigns a security descriptor to the desktop
+ * object during a desktop object parse procedure.
+ *
+ * @param[in] WinSta
+ * A pointer to a window station object, of which
+ * such object contains its own security descriptor
+ * that will be captured.
+ *
+ * @param[in] Desktop
+ * A pointer to a desktop object that is created
+ * during a parse procedure.
+ *
+ * @param[in] AccessState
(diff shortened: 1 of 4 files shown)


win32ss/user/ntuser/security.c
```diff
@@ -235,6 +235,73 @@ IntQueryUserSecurityIdentification(
...
+                              ExDesktopObjectType);
+    if (!NT_SUCCESS(Status))
+    {
+        ERR("IntAssignDesktopSecurityOnParse(): Failed to assign security information to the desktop object (Status 0x%08lx)\n", Status);
+        ObReleaseObjectSecurity(CapturedDescriptor, MemoryAllocated);
+        return Status;
+    }
+
+    ObReleaseObjectSecurity(CapturedDescriptor, MemoryAllocated);
+    return STATUS_SUCCESS;
```
HBelusca, line 302: Could be simplified by:
```suggestion
    }

    ObReleaseObjectSecurity(CapturedDescriptor, MemoryAllocated);
    return Status;
```

HBelusca (approved):

PR #4497, opened 2022-05-08
[FRAGINATOR] Fix some uninitialized variables and make it work again

BUGinator86:
…Fraginator

## Purpose

_Set BitmapSize to zero so DeviceIOControl doesn't fail_

JIRA issue: [ROSAPPS-332](https://jira.reactos.org/browse/ROSAPPS-332)

## Proposed changes

_Make Fraginator working again_

### DIFF
--- a/modules/rosapps/applications/fraginator/DriveVolume.cpp
+++ b/modules/rosapps/applications/fraginator/DriveVolume.cpp
@@ -214,9 +214,8 @@ bool DriveVolume::GetBitmap (void)
 
     if (reallBitmap == NULL)
     {
-        //Fail "miserably"
-        //We don't have enought memory left
-        wprintf(L"\nNot enought memory to read volume bitmap\n");
+        // Fail "miserably"
+        wprintf(L"\nNot enough memory to read volume bitmap\n");
         HeapFree(GetProcessHeap(), 0, Bitmap);
         return (false);
     }
--- a/modules/rosapps/applications/fraginator/DriveVolume.cpp
+++ b/modules/rosapps/applications/fraginator/DriveVolume.cpp
@@ -1,11 +1,11 @@
 #include "DriveVolume.h"
 
 
-DriveVolume::DriveVolume ()
+DriveVolume::DriveVolume()
 {
     Handle = INVALID_HANDLE_VALUE;
     BitmapDetail = NULL;
-    return;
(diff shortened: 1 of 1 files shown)


modules/rosapps/applications/fraginator/DriveVolume.cpp
```diff
@@ -1,11 +1,11 @@
 #include "DriveVolume.h"
 
 
-DriveVolume::DriveVolume ()
+DriveVolume::DriveVolume()
 {
     Handle = INVALID_HANDLE_VALUE;
     BitmapDetail = NULL;
-    return;
+    ZeroMemory(&Geometry, sizeof(Geometry));
```
learn-more, line 8: So this is the actual fix, the rest is just changing code to your preferred style?

BUGinator86: No. There are several instances where structs needed to be initialized with ZERO. Instead of writing malloc() and zeroMemory() in all these cases I just used HeapAlloc with ZeroMemory flag. :)
If there is another way, we could use the alternative. AFAIK a flag in VS/MSBuild exists to zero-initialize vars by default. If we could use a similar flag, then we would solve the problems without changing the code at all. :)

learn-more (approved):

modules/rosapps/applications/fraginator/DriveVolume.cpp
```diff
@@ -200,16 +200,28 @@ bool DriveVolume::GetBitmap (void)
...
     BitmapSize = sizeof (VOLUME_BITMAP_BUFFER) + (Bitmap->BitmapSize.QuadPart / 8) + 1;
-    Bitmap = (VOLUME_BITMAP_BUFFER *) realloc (Bitmap, BitmapSize);
+
+    void *reallBitmap = HeapReAlloc(GetProcessHeap(), 0, Bitmap, BitmapSize);
+
+    if (reallBitmap == NULL)
+    {
+        //Fail "miserably"
+        //We don't have enought memory left
+        wprintf(L"\nNot enought memory to read volume bitmap\n");
```
binarymaster, line 219:
```suggestion
        // Fail "miserably"
        wprintf(L"\nNot enough memory to read volume bitmap\n");
```
Fix typo in "enought", fix comment style, remove one obvious duplicated line :)

jeditobe:
Jira issue link is broken

PR #4510, opened 2022-05-13
Fix some build bugs

wjk:
## Purpose

During the build several places in the host-tools assume that the build is being performed an Intel processor that has a little-endian byte order. Since modern Macs use the ARM architecture and are big-endian, this caused problems. With this patch applied, I can build `calc` (my usual smoke test) from a fresh clone without problem.

JIRA issue: (none)

## Proposed changes

See the commit messages.

One note about bec8b07: On my dev box the RosBE tools are only in the PATH when needed. Since my IDE inherits the PATH from the system and not from my shell, I needed to be able to run `configure.sh` with the tools in my PATH and then be able to successfully reconfigure when building from a “regular” command prompt.


wjk:
Unfortunately, I started this PR from a fork under a GitHub org that no longer exists. 😔 Will properly open shortly. Thanks!

PR #4511, opened 2022-05-13
Pr/fix arm build

wjk:
## Purpose

During the build several places in the host-tools assume that the build is being performed an Intel processor that has a little-endian byte order. Since modern Macs use the ARM architecture and are big-endian, this caused problems. With this patch applied, I can build calc (my usual smoke test) from a fresh clone without problem.

JIRA issue: (none)  
Redo of #4510.

## Proposed changes

See the commit messages.

One note about https://github.com/reactos/reactos/commit/bec8b079c310cd205de22ce93b453ab4a893f80b: On my dev box the RosBE tools are only in the PATH when needed. Since my IDE inherits the PATH from the system and not from my shell, I needed to be able to run configure.sh with the tools in my PATH and then be able to successfully reconfigure when building from a “regular” command prompt.


sdk/tools/mkisofs/schilytools/libmdigest/byte_order.h
```diff
@@ -72,12 +72,12 @@ extern "C" {
 	defined(__POWERPC__) || defined(POWERPC) || defined(__powerpc) || \
 	defined(__powerpc__) || defined(__powerpc64__) || defined(__ppc__) || \
 	defined(__hpux) || defined(_MIPSEB) || defined(mc68000) || \
-	defined(__s390__) || defined(__s390x__) || defined(sel)
+	defined(__s390__) || defined(__s390x__) || defined(sel) || (defined(__arm64__) && defined(__APPLE__))
 #define	CPU_BIG_ENDIAN
 #define	IS_BIG_ENDIAN 1
 #define	IS_LITTLE_ENDIAN 0
```
binarymaster, line 78: > Per my research all M1 processors are big endian.

It's actually Little Endian:

From https://developer.apple.com/documentation/apple-silicon/porting-your-macos-apps-to-apple-silicon
> Both Apple silicon and Intel-based Mac computers use the little-endian format for data, so you don’t need to make endian conversions in your code.

From https://www.quora.com/Is-Apples-new-M1-processor-big-endian-like-the-original-Motorola-microprocessor-or-little-endian-like-the-Intel-processor
> ARMv8 instruction set architecture, used in Apples CPUs, like all modern architectures is a little-endian architecture, just like x86, and unlike 68000 or ancient Apples.

wjk: In that case I have no idea why Google told me they are big-endian when I asked it the question. How strange.

dll/win32/dbghelp/symbol.c
```diff
@@ -109,7 +109,7 @@ DWORD             symt_ptr2index(struct module* module, const struct symt* sym)
 
 struct symt*      symt_index2ptr(struct module* module, DWORD id)
 {
-#ifdef __x86_64__
+#if defined(__x86_64__) || (defined(__arm64__) && defined(__APPLE__))
```
binarymaster, line 112: I think this should be common for any arm64:
```suggestion
#if defined(__x86_64__) || (defined(__arm64__)
```

sdk/tools/mkisofs/schilytools/libmdigest/byte_order.h
```diff
@@ -61,7 +61,7 @@ extern "C" {
 		__BYTE_ORDER == __LITTLE_ENDIAN) || \
 	defined(CPU_IA32) || defined(CPU_X64) || \
 	defined(__ia64) || defined(__ia64__) || defined(__alpha__) || defined(_M_ALPHA) || \
-	defined(vax) || defined(MIPSEL) || defined(_ARM_)
+	defined(vax) || defined(MIPSEL) || defined(_ARM_) || (defined(__arm64__) && defined(__APPLE__))
```
binarymaster, line 64: Same here, but maybe also it's worth to check whether `_M_ARM64` or `_ARM64_` are defined here too:
```suggestion
	defined(vax) || defined(MIPSEL) || defined(_ARM_) || (defined(__arm64__)
```

sdk/tools/mkisofs/schilytools/libmdigest/byte_order.h
```diff
@@ -73,11 +73,12 @@ extern "C" {
 	defined(__powerpc__) || defined(__powerpc64__) || defined(__ppc__) || \
 	defined(__hpux) || defined(_MIPSEB) || defined(mc68000) || \
 	defined(__s390__) || defined(__s390x__) || defined(sel)
+|| (defined(__arm64__) && defined(__APPLE__))
```
binarymaster, line 76: Leftover
```suggestion
```

toolchain-gcc.cmake
```diff
@@ -28,15 +28,15 @@ set(CMAKE_SYSTEM_NAME Windows)
 set(CMAKE_SYSTEM_PROCESSOR i686)
 
 # Which tools to use
-set(CMAKE_C_COMPILER ${MINGW_TOOLCHAIN_PREFIX}gcc${MINGW_TOOLCHAIN_SUFFIX} CACHE FILEPATH "The C Compiler")
-set(CMAKE_CXX_COMPILER ${MINGW_TOOLCHAIN_PREFIX}g++${MINGW_TOOLCHAIN_SUFFIX} CACHE FILEPATH "The C++ Compiler")
-set(CMAKE_ASM_COMPILER ${MINGW_TOOLCHAIN_PREFIX}gcc${MINGW_TOOLCHAIN_SUFFIX} CACHE FILEPATH "The ASM Compiler")
+find_program(CMAKE_C_COMPILER ${MINGW_TOOLCHAIN_PREFIX}gcc${MINGW_TOOLCHAIN_SUFFIX} REQUIRED)
```
Extravert-ir, line 31: the "REQUIRED" keyword seems to be CMake 3.18+ so won't work on on RosBE default CMake

Extravert-ir (changes requested):
Otherwise, LGTM

binarymaster (approved):

toolchain-gcc.cmake
```diff
@@ -28,15 +35,15 @@ set(CMAKE_SYSTEM_NAME Windows)
 set(CMAKE_SYSTEM_PROCESSOR i686)
 
 # Which tools to use
-set(CMAKE_C_COMPILER ${MINGW_TOOLCHAIN_PREFIX}gcc${MINGW_TOOLCHAIN_SUFFIX} CACHE FILEPATH "The C Compiler")
-set(CMAKE_CXX_COMPILER ${MINGW_TOOLCHAIN_PREFIX}g++${MINGW_TOOLCHAIN_SUFFIX} CACHE FILEPATH "The C++ Compiler")
-set(CMAKE_ASM_COMPILER ${MINGW_TOOLCHAIN_PREFIX}gcc${MINGW_TOOLCHAIN_SUFFIX} CACHE FILEPATH "The ASM Compiler")
+require_program(CMAKE_C_COMPILER ${MINGW_TOOLCHAIN_PREFIX}gcc${MINGW_TOOLCHAIN_SUFFIX})
```
HBelusca, line 38: i wonder whether similar improvements can be applied to the other toolchain files?

PR #4514, opened 2022-05-14
[NTOS:MM] Fix memory leak in MiMapViewOfDataSection

JohnyTheCarrot:
## Purpose

If inserting the allocated VAD fails, MiMapViewOfDataSection will make no attempt to free the allocated VAD.
This PR should fix the leaking of memory in such situations.

### DIFF
--- a/ntoskrnl/mm/ARM3/section.c
+++ b/ntoskrnl/mm/ARM3/section.c
@@ -1494,6 +1494,11 @@ MiMapViewOfDataSection(IN PCONTROL_AREA ControlArea,
     if (!NT_SUCCESS(Status))
     {
         ExFreePoolWithTag(Vad, 'ldaV');
+        MiDereferenceControlArea(ControlArea);
+
+        KeAcquireGuardedMutex(&MmSectionCommitMutex);
+        Segment->NumberOfCommittedPages -= QuotaCharge;
+        KeReleaseGuardedMutex(&MmSectionCommitMutex);
         return Status;
     }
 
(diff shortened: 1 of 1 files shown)


ntoskrnl/mm/ARM3/section.c
```diff
@@ -1498,6 +1498,8 @@ MiMapViewOfDataSection(IN PCONTROL_AREA ControlArea,
                            AllocationType);
     if (!NT_SUCCESS(Status))
     {
+        MiDereferenceControlArea(ControlArea);
+        ExFreePoolWithTag(Vad, 'ldaV');
```
HBelusca, line 1502: VAD was created after, so it has to disappear first.
```suggestion
        ExFreePoolWithTag(Vad, 'ldaV');
        MiDereferenceControlArea(ControlArea);
```

HBelusca: There is also (at least one) modification that has been made on the `Segment`, above:
`Segment->NumberOfCommittedPages += QuotaCharge;`
that probably needs to be rolled back too (to be done under `MmSectionCommitMutex` held).

HBelusca: Another (perhaps better?) way would be to insert the VAD before doing the operations on the corresponding Section...

JohnyTheCarrot:
Just to be clear, did I miss a change that was required?

GeoB99 (commented):
`NumberOfCommittedPages` of a segment has to be rolled back to its initial state when you fail to charge process' quota as well. Apart from that, your PR has conflicts which need to be resolved.

JohnyTheCarrot:
> `NumberOfCommittedPages` of a segment has to be rolled back to its initial state when you fail to charge process' quota as well. Apart from that, your PR has conflicts which need to be resolved.

both have already been addressed

GeoB99:
I do not see anywhere you're rolling `NumberOfCommittedPages` back if `PsChargeProcessNonPagedPoolQuota` fails.
Also...
Capture

learn-more:
> I do not see anywhere you're rolling `NumberOfCommittedPages` back if `PsChargeProcessNonPagedPoolQuota` fails. Also... Capture

`Segment->NumberOfCommittedPages -= QuotaCharge;` is in the diff, if this is not what you expect to see then it might make sense to elaborate a bit more.

GeoB99:
In case it is still not self explanatory enough.
Capture
Here you're accordingly rolling `NumberOfCommittedPages` back as it should be, however, suppose `PsChargeProcessNonPagedPoolQuota` fails because of overreaching the quota limits, you aren't rolling the number of committed pages back to initial state.

What happens here is that `NumberOfCommittedPages` of a control area is being left touched on that failure path which is wrong.

GeoB99 (approved):

PR #4516, opened 2022-05-14
[NOTEPAD] Less flickering title bar

katahiromz:
## Purpose

When typing text on NotePad, then NotePad title bar flickers.
JIRA issue: N/A

## Proposed changes

- Save previous modification flag.
- If no change in modification flag, do not set the title bar text.

### DIFF
--- a/base/applications/notepad/dialog.c
+++ b/base/applications/notepad/dialog.c
@@ -84,6 +84,8 @@ void UpdateWindowCaption(BOOL clearModifyAlert)
     {
         StringCbPrintf(szCaption, sizeof(szCaption), _T("%s - %s"),
                        szFilename, szNotepad);
+
+        Globals.bWasModified = FALSE;
     }
     else
     {
--- a/base/applications/notepad/dialog.c
+++ b/base/applications/notepad/dialog.c
@@ -68,7 +68,6 @@ void UpdateWindowCaption(BOOL clearModifyAlert)
     TCHAR szCaption[MAX_STRING_LEN];
     TCHAR szNotepad[MAX_STRING_LEN];
     TCHAR szFilename[MAX_STRING_LEN];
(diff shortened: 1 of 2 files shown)


HBelusca:
When does this flickering happen?

katahiromz:
>When does this flickering happen?

Typing text into NotePad.

base/applications/notepad/dialog.c
```diff
@@ -68,6 +68,7 @@ void UpdateWindowCaption(BOOL clearModifyAlert)
     TCHAR szCaption[MAX_STRING_LEN];
     TCHAR szNotepad[MAX_STRING_LEN];
     TCHAR szFilename[MAX_STRING_LEN];
+    static TCHAR s_szOldCaption[MAX_STRING_LEN] = TEXT("");
```
HBelusca, line 71: As you can see, even if the caption is not going to be changed, that whole function is called at each key press. Since that function does quite a lot of operations, it is slowing down stuff.
You may need to consider the alternative way to verify whether or not the window caption needs to be modified. This code would have to be inserted here at the very beginning of this function, before doing all the rest.

```c
    BOOL isModified;
    // It might be better to save this variable in the Globals structure...
    static BOOL old_isModified = FALSE;

    if (!clearModifyAlert)
    {
        /* Check whether we need to update the window caption */
        isModified = ... ; // The SendMessage() call of the original code below.

        if (isModified == old_isModified)
        {
            /* We are in the same state as before, don't change the caption */
            return;
        }

        /* We will modify the caption below */
        old_isModified = isModified;
    }

    // Now comes the rest of the function: the LoadString() call, etc.
```

learn-more:
Is this also required on windows?

katahiromz:
@learn-more Yep.
image

PR #4519, opened 2022-05-16
[WIN32SS][ENG][NTDDRAW] Manage DirectDraw instances when switching display mode

oleg-dubinskiy:
## Purpose

[ENG] Implement DirectDraw management in switch display mode functions (e. g. whose change resolution, color depth, display frequency etc.):
- Switch DirectDraw instances between the two `PDEV`s (the current one and the new one allocated by ourself) by calling dxg!`DxDdDynamicModeChange` function.
- Suspend them before and resume after the display mode switch, by calling dxg!`DxDdsuspendDirectDraw` and dxg!`DxDdResumeDirectDraw` appropriately.
We currently don't have these functions implemented, but MS DXG has, so it allows to propely manage DirectDraw `PDEV`s using this driver, similarly to Windows.
My analysis confirms that these functions are always called in XP/2k3 on display mode switch, even when there is no any DirectX app running at the moment. So I'm doing the same.
I've also analyzed their prototypes and see that my guessings are right.
- Initialize `hDev` and `dhpdev` members for `EDD_DIRECTDRAW_GLOBAL` for newly created surfaces, switch them during mode change and re-initialize after it also. They are commonly used by DirectDraw stack.
Also enable DirectDraw for an old and new PDEVs, by calling dxg!`DxDdEnableDirectDraw` function.

[NTDDRAW] Additionally, fix usage of DirectDraw lock count in `PDEVOBJ` structure.
- Enable `cDirectDrawDisableLocks` member for storing its value, instead of `DxDd_nCount`, which is marked as ROS-specific.
- Use it in win32k!`DxEngGet`/`SetHdevData` for getting/setting DirectDraw count aprropriately.
My analysis also shows that in Windows, `PDEVOBJ::cDirectDrawDisableLocks` method calls `DxEngGetHdevData` with type `8`, which corresponds to our `DxDd_nCount`. So there is no any doubts that this member is used there.
- Rename `DxEngGetHdevData_dd_count` alias of type `8` to `DxEngGetHdevData_dd_locks`, to more match to an actual member name. Update the enumeration and fix all code parts appropriately.

All these changes allow to properly change display mode during executing DirectDraw applications, when they try to switch in fullscreen mode.
At least bugcheck that happened before my changes, does no longer appear.
Although some games still don't run correctly like there is no 3D acceleration (which actually exists). It needs further investigations.

And please note that my patch causes the problem when switching display mode with our current ddraw.dll (that is imported from Wine). Wine's implementation doesn't relay on DXG kernel APIs and uses WineD3D OpenGL wrapper instead. After trying to do that, appears an exception due to something uninitialized. But the MS ddraw.dll and dxg.sys are now functioning much better. **So it is not recommend to apply this PR until that problem is fixed** (by patching Wine's ddraw or switching to our ddraw_new instead). **UPDATE: it happens also in master, so seems unrelated to my changes.**
And don't forget that my #3638 PR also is badly needed to be applied, to see any effect from my current changes.

JIRA issue: [CORE-17932](https://jira.reactos.org/browse/CORE-17932)

## Proposed changes

- Switch DirectDraw instances between the two `PDEV`s (the current one and the new one allocated by ourself) by calling dxg!`DxDdDynamicModeChange` function.
- Suspend them before and resume after the display mode switch, by calling dxg!`DxDdsuspendDirectDraw` and dxg!`DxDdResumeDirectDraw` appropriately.
- Initialize `hDev` and `dhpdev` members for `EDD_DIRECTDRAW_GLOBAL` for newly created surfaces, switch them during mode change and re-initialize after it also. They are commonly used by DirectDraw stack.
- Enable `cDirectDrawDisableLocks` member for storing its value, instead of `DxDd_nCount`, which is marked as ROS-specific.
- Use it in win32k!`DxEngGet`/`SetHdevData` for getting/setting DirectDraw count appropriately.
My analysis also shows that in Windows, `PDEVOBJ::cDirectDrawDisableLocks` method calls `DxEngGetHdevData` with type `8`, which corresponds to our `DxDd_nCount`. So there is no any doubts that this member is used there.
- Rename `DxEngGetHdevData_dd_count` alias of type `8` to `DxEngGetHdevData_dd_locks`, to more match to an actual member name. Update the enumeration and fix all code parts appropriately.

## TODO

- [ ] Investigate why some DirectDraw games still don't run properly in fullscreen mode, like without hardware acceleration.
- [x] Find a possible fix for the problem with using Wine's DirectDraw after my changes. **UPDATE: this is also happening in master, so is unrelated to my changes.**
- [x] **UPDATE**: Now fixed display mode switch problem, which was discovered recently. Exception during manual attempt to switch the mode happened due to some uninitialized data in `EDD_DIRECTDRAW_GLOBAL` structure. Fixed it by calling dxg!`DxDdEnableDirectDraw` function. :slightly_smiling_face: 
- [x] **UPDATE 2**: Don't disable DirectDraw for an old PDEV. It simply doesn't make sense, because it is released and not used anymore anyway. Doing that only causes de-initialization of the common PDEV stuff (when using MS DXG) followed in memory, since dxg!`DxDdDisableDirectDraw` actually reverses all things that dxg!`DxDdEnableDirectDraw` does. Without doing it, both PDEVs (old and new ones) are initialized correctly. My further analysis confirms that Windows XP/2003 also doesn't do that. This fixes an exception 0xc00000005 when display resolution is switched by DirectDraw/Direct3D application without terminating (when the app does not exiting). For example, when switching display mode from the game menu in Rome: Total War or starting the game from menu with the different resolution in Battlefield 1942 Singleplayer Demo (the games tested by me, but it can be also reproduced with any others). :slightly_smiling_face: 
- [x] **UPDATE 3**: Don't exchange PDEVs in `EDD_DIRECTDRAW_GLOBAL` structure. Doing this manually appears to be bogus, because it already seems to be done properly via dxg!`DxDdDynamicModeChange` function instead, which is also called there. It fixes assert `(((Process)->Header.Type & KOBJECT_TYPE_MASK) == ProcessObject)` in nt!`KeAttachProcess` called from dxg!`DxDdResumeDirectDraw` for new PDEV when trying to launch any DirectDraw/Direct3D application in fullscreen mode 2nd or more times. So now they can be launched multiple times without issues. :slightly_smiling_face: 

## Result

Sorry guys, but I'm not able to make a demo video, because my recorder captures simply a black screen instead of contents, due to stupid open source Linux drivers for my video card. :frowning_face: Feel free to test my PR with by yourself and record a video if you can. :wink: 
**UPDATE**: video test is now recorded and published on my YouTube channel. :smiley: 
**UPDATE 2**: here's a video URL: https://youtu.be/H8hAfJhksdQ.


tkreuzer (approved):

learn-more:
Buildbot build:
https://build.reactos.org/#/builders/9/builds/33331

win32ss/gdi/eng/pdevobj.c
```diff
@@ -336,6 +337,55 @@ PDEVOBJ_pSurface(
     return ppdev->pSurface;
 }
 
+BOOL
+PDEVOBJ_bEnableDirectDraw(
+    _Inout_ PPDEVOBJ ppdev)
+{
+    PGD_DXDDENABLEDIRECTDRAW pfnDdEnableDirectDraw = (PGD_DXDDENABLEDIRECTDRAW)gpDxFuncs[DXG_INDEX_DxDdEnableDirectDraw].pfn;
+    BOOL assertVal;
```
HBelusca, line 345: why called "assertVal"? Maybe just "Success" ?

win32ss/gdi/eng/pdevobj.c
```diff
@@ -548,6 +600,15 @@ PDEVOBJ_Create(
         return NULL;
     }
 
+    /* Enable DirectDraw */
+    if (!PDEVOBJ_bEnableDirectDraw(ppdev))
+    {
+        DPRINT1("Failed to enable DirectDraw\n");
```
HBelusca, line 606: DPRINTs or ERR? (the debug channels are available in this file)

win32ss/reactx/ntddraw/dxeng.c
```diff
@@ -413,9 +413,10 @@ DxEngSetHdevData(HDEV hDev,
 
     DPRINT1("ReactX Calling : DxEngSetHdevData DXEGSHDEVDATA : %ld\n", Type);
 
-    if ( Type == DxEGShDevData_dd_nCount )
+    if ( Type == DxEGShDevData_dd_locks )
```
HBelusca, line 416:
```suggestion
    if (Type == DxEGShDevData_dd_locks)
```

win32ss/reactx/ntddraw/intddraw.h
```diff
@@ -108,6 +108,11 @@ typedef VOID (APIENTRY *PGD_ENGFREEPRIVATEUSERMEM)(PDD_SURFACE_LOCAL, PVOID);
 typedef PDD_SURFACE_LOCAL (APIENTRY *PGD_ENGLOCKDIRECTDRAWSURFACE)(HANDLE);
 typedef BOOL (APIENTRY *PGD_ENGUNLOCKDIRECTDRAWSURFACE)(PDD_SURFACE_LOCAL);
 
+/* Other internal functions */
+typedef VOID (APIENTRY *PGD_DXDDDYNAMICMODECHANGE)(HDEV, HDEV, ULONG);
+typedef VOID (APIENTRY *PGD_DXDDRESUMEDIRECTDRAW)(HDEV, ULONG);
+typedef VOID (APIENTRY *PGD_DXDDSUSPENDDIRECTDRAW)(HDEV, ULONG);
```
HBelusca, line 114: Question: is this last parameter (the ULONG) truly a ULONG, or ULONG_PTR (or PVOID) ? (i.e. something whose size depends on 32 vs. 64 bits)

HBelusca:
Test results (merged with current master):
Test KVM: https://reactos.org/testman/compare.php?ids=87478,87481
Test VBox: https://reactos.org/testman/compare.php?ids=87480,87482
Looks ok to me.

win32ss/gdi/eng/pdevobj.c
oleg-dubinskiy: I decided to name it `assertVal` based on other code style in this file. Look at other routines those are calling driver functions. Should I rename them to something like `Success` too? Definitely not in this PR.

win32ss/gdi/eng/pdevobj.c
oleg-dubinskiy: Okay, `ERR` seems better here. :slightly_smiling_face:

win32ss/reactx/ntddraw/intddraw.h
oleg-dubinskiy: As I know, `ULONG_PTR` is like a pointer to `ULONG`, same as `PVOID` is a pointer to `VOID`. In debugger, when I analysed this stuff, I saw only `00000000`, or `00000002` passed as the 2nd parameter into `DxDdSuspendDirectDraw` and `DxDdResumeDirectDraw`, and the 3rd one to `DxDdDynamicModeChange`. If it would be a pointer, it could look like a some hexadecimal address (combination of symbols), e. g. `k3ud58jm`, not just a decimal number, like `0` or `2`. Such pointer needs to be unassembled one more time to see its actual data. So it's looks more like a simple `ULONG` or `DWORD` flags, those are not documented as well.

HBelusca: `ULONG_PTR` is a scalar value (a number) whose size corresponds to the size of pointers in the current platform.
For example, a `ULONG_PTR` is 4 bytes when you compile for 32-bits , but is 8 bytes when you compile for 64-bits.
This is why such a value can hold pointer addresses without problems.

On the contrary, using `ULONG` values to convey pointer addresses, would truncate such addresses when compiling code for 64-bits.

The type for a pointer to `ULONG` , i.e. `ULONG *`, is named `PULONG` usually, and not `ULONG_PTR`.

oleg-dubinskiy: Okay, then using `ULONG_PTR` is better here.

win32ss/gdi/eng/pdevobj.c
oleg-dubinskiy: Okay, changed the variable name to `Success`, but only in this function.

win32ss/reactx/ntddraw/dxeng.c
oleg-dubinskiy: Done, thanks. :slightly_smiling_face:

JoachimHenze (approved):

HBelusca (approved):
I would strongly suggest that for all those (mostly internal and/or undocumented) DirectX functions, you write a documentation for them in doxygen format.
(For examples, see https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/io/iomgr/iorsrce.c;h=01f2b104a846cf1764a774886543f0dac2db88fd;hb=HEAD#l1157 or some commits by @GeoB99 )

jeditobe:
Are these changes self sufficient? Or some extra pathes are needed?

oleg-dubinskiy:
> Are these changes self sufficient? Or some extra pathes are needed?

#4551 is badly required by #4519, to fix regression caught by this PR. #5329 is independent from any of them.

PR #4524, opened 2022-05-21
[HALX86] Print the correct size of the BAR in HalpDebugPciDumpBus()

disean:
## Purpose

BEFORE:
```
00:01.1 IDE interface [0101]: Intel Corporation 82371AB/EB/MB PIIX4 IDE [8086:7111] (rev 01)
        Subsystem: Unknown [0000:0000]
        Flags: bus master, fast devsel, latency 0
        I/O ports at d000 [size=4K]
```
AFTER:
```
00:01.1 IDE interface [0101]: Intel Corporation 82371AB/EB/MB PIIX4 IDE [8086:7111] (rev 01)
        Subsystem: Unknown [0000:0000]
        Flags: bus master, fast devsel, latency 0
        I/O ports at d000 [size=16]
```
Note: 64-bit BARs are not supported yet.

### DIFF
--- a/hal/halx86/legacy/bussupp.c
+++ b/hal/halx86/legacy/bussupp.c
@@ -787,7 +787,9 @@ ShowSize(ULONG x)
 CODE_SEG("INIT")
 VOID
 NTAPI
-HalpDebugPciDumpBus(IN ULONG i,
+HalpDebugPciDumpBus(IN PBUS_HANDLER BusHandler,
+                    IN PCI_SLOT_NUMBER PciSlot,
+                    IN ULONG i,
                     IN ULONG j,
                     IN ULONG k,
                     IN PPCI_COMMON_CONFIG PciData)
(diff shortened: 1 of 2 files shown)


hal/halx86/legacy/bussupp.c
```diff
@@ -963,12 +965,30 @@ HalpDebugPciDumpBus(IN ULONG i,
...
+            HalpReadPCIConfig(BusHandler,
+                              PciSlot,
+                              &PciBar,
+                              FIELD_OFFSET(PCI_COMMON_HEADER, u.type0.BaseAddresses[b]),
+                              sizeof(ULONG));
+            HalpWritePCIConfig(BusHandler,
+                               PciSlot,
+                               &Mem,
+                               FIELD_OFFSET(PCI_COMMON_HEADER, u.type0.BaseAddresses[b]),
+                               sizeof(ULONG));
```
binarymaster, line 984: Ah, so that means it's not possible to obtain correct BAR size without writing to it?

binarymaster (approved):
Yes, this is correct. It's documented in https://wiki.osdev.org/PCI#Address_and_size_of_the_BAR :

> To determine the amount of address space needed by a PCI device, you must save the original value of the BAR, write a value of all 1's to the register, then read it back.

PR #4526, opened 2022-05-24
[WIN32SS] EnumDisplayDevices: Fill DeviceID field properly

binarymaster:
Since this is kernel mode stuff, and especially Win32SS, let's review this together.

- [VIDEOPRT] Return the real PhysicalDeviceObject in Win32k callbacks.

- [WIN32SS:ENG] Pass PhysDeviceObject into the GraphicsDevice structure
  and rename the destination field accordingly.

- [WIN32SS:NTUSER] Request hardware identifiers from device PDO and
  fill DISPLAY_DEVICE's DeviceID field with the first identifier.

Now it's correctly passed to the usermode, and Desktop Propertes applet
can open the video adapter device properties.

Thanks to @hpoussin for the help.


binarymaster:
Here's how it looks:

VirtualBox_ReactOS_24_05_2022_03_27_56

Previously the "Properties" button in Adapter tab was disabled, because it couldn't determine hardware identifier for the device.

win32ss/gdi/eng/pdevobj.h
```diff
@@ -61,7 +61,7 @@ typedef struct _GRAPHICS_DEVICE
     struct _GRAPHICS_DEVICE * pNextGraphicsDevice;
     struct _GRAPHICS_DEVICE * pVgaDevice;
     PDEVICE_OBJECT   DeviceObject;
-    PVOID            pDeviceHandle;
+    PDEVICE_OBJECT   PhysDeviceHandle;
```
HBelusca, line 64: device "Handle" sounds like it would have to be some sort of a handle, not a direct `PDEVICE_OBJECT` 🤔

win32ss/user/ntuser/display.c
```diff
@@ -280,9 +281,53 @@ UserEnumDisplayDevices(
     RtlStringCbCopyW(pdispdev->DeviceName, sizeof(pdispdev->DeviceName), pGraphicsDevice->szWinDeviceName);
     RtlStringCbCopyW(pdispdev->DeviceString, sizeof(pdispdev->DeviceString), pGraphicsDevice->pwszDescription);
     pdispdev->StateFlags = pGraphicsDevice->StateFlags;
-    // FIXME: fill in DEVICE ID
     pdispdev->DeviceID[0] = UNICODE_NULL;
 
+    if (pGraphicsDevice->PhysDeviceHandle != NULL)
```
HBelusca, line 286: Maybe add a comment explaining what is going to be done there, and the reason why only the first hardware ID is the one considered?

win32ss/user/ntuser/display.c
```diff
@@ -280,9 +281,53 @@ UserEnumDisplayDevices(
...
+
+        if (Status == STATUS_BUFFER_TOO_SMALL)
+        {
+            pHardwareId = ExAllocatePoolWithTag(PagedPool,
+                                                dwLength,
+                                                USERTAG_DISPLAYINFO);
+            if (!pHardwareId)
+            {
+                EngSetLastError(ERROR_NOT_ENOUGH_MEMORY);
+                return STATUS_UNSUCCESSFUL;
```
HBelusca, line 302: or `STATUS_INSUFFICIENT_RESOURCES` ?

win32ss/gdi/eng/pdevobj.h
binarymaster: Yes, dunno why it's called like that. `DeviceObject` contains just DO, i.e. link to the video driver itself. VideoPrt previously returned pointer to the same DO object, so both `DeviceObject` and `PhysDeviceHandle` members were equal - which was incorrect and fixed :)

win32ss/user/ntuser/display.c
HBelusca: Do these different IDs have different formats?

binarymaster: image

Just a bit.

binarymaster: I have added missing bits regarding the monitor handling. As it turned out, there is a lot of work needs to be done for it, and I don't want to interfere with Hervé's work on multiple monitors. :smile:

binarymaster:
Testbot results:
- :white_check_mark: KVM: https://build.reactos.org/#/builders/6/builds/31775
- :white_check_mark: VBox: https://build.reactos.org/#/builders/3/builds/19254

PR #4527, opened 2022-05-25
[WALLPAPER] Add new wallpapers from Daniel Myslivets

veselcraft:
Title says the purpose of it all. All wallpapers are licensed by CC_BY-SA.

_Co-authored with @myslivets_


HBelusca:
This should go into rapps instead.

myslivets:
> This should go into rapps instead.

RAPPS? That's cool, of course, but why? It''s possible to throw in a couple of wallpapers (~3 or 5 wallpapers) as a demonstration of the desktop background change functionality. _[why not lol]_
And the others are to upload to RAPPS as a supplement. Although, it's up to you, I'm just a suggestion :)

HBelusca:
> > This should go into rapps instead.
> 
> RAPPS? That's cool, of course, but why? It''s possible to throw in a couple of wallpapers (for example: 3-5 wallpapers) as a demonstration of the desktop background change functionality. _[why not lol]_ And the others are to upload to RAPPS as a supplement. Although, it's up to you, I'm just a suggestion :)

That's because we already have demo wallpapers in our source code that gets added in into the releases (whenever the release engineer doesn't forget to add them, that is ^^ ).
See https://github.com/reactos/reactos/tree/master/modules/wallpapers
They get added when (re)configuring a build with `-DENABLE_WALLPAPERS=1` option, it was already explained in the ReadMe.txt file:
https://github.com/reactos/reactos/blob/master/modules/wallpapers/ReadMe.txt

So we already have the couple of wallpapers as a demo of desktop background change functionality.

PR #4528, opened 2022-05-25
[KERNEL32] Fix BaseDefaultPath containing extra '.'

learn-more:
2k3:

image

ROS before patch:
image

ROS after patch:
image


learn-more:
The first path being different is because I patched Rapps to display this, and on ROS it is running from system32, on Win it was running from the desktop

HBelusca (approved):

binarymaster (approved):

PR #4530, opened 2022-05-25
[ROSAPPS] Remove 'stats' utility

learn-more:


HBelusca:
i suppose one could use any other line-counting tool out there?

HBelusca (approved):

learn-more:
> i suppose one could use any other line-counting tool out there?

Or if one was inclined to write something, use a tool more suited for this like python, where it would be around 10 lines of code.

Extravert-ir (approved):

PR #4531, opened 2022-05-28
[NTOS:SE] Reorganize the security manager component

GeoB99:
The current state of Security manager's code is kind of a mess. Mainly, there's code scattered around places where they shouldn't belong and token implementation (token.c) is already of a bloat in itself as it is. The file has over 6k lines and it's subject to grow exponentially with improvements, features, whatever that is.

With that being said, the token implementation code in the kernel will be split accordingly and rest of the code moved to appropriate places. The new layout will look as follows (excluding the already existing files):

- client.c (Client security implementation code)
- objtype.c (Object type list implementation code -- more code related to object types will be put here when I'm going to implement object type access checks in the future)
- subject.c (Subject security context support)

The token implementation in the kernel will be split in 4 distinct files as shown:

- token.c (Base token support routines)
- tokenlif.c (Life management of a token object -- that is Duplication, Creation and Filtering)
- tokencls.c (Token Query/Set Information Classes support)
- tokenadj.c (Token privileges/groups adjusting support)

In addition to that, tidy up the internal header and reorganize it as well.


ntoskrnl/ntos.cmake
```diff
@@ -275,13 +275,19 @@ list(APPEND SOURCE
...
     ${REACTOS_SOURCE_DIR}/ntoskrnl/se/semgr.c
     ${REACTOS_SOURCE_DIR}/ntoskrnl/se/sid.c
     ${REACTOS_SOURCE_DIR}/ntoskrnl/se/sqos.c
     ${REACTOS_SOURCE_DIR}/ntoskrnl/se/srm.c
-    ${REACTOS_SOURCE_DIR}/ntoskrnl/se/token.c
+    ${REACTOS_SOURCE_DIR}/ntoskrnl/se/subject.c
+    ${REACTOS_SOURCE_DIR}/ntoskrnl/se/tknadj.c
+    ${REACTOS_SOURCE_DIR}/ntoskrnl/se/tkncls.c
+    ${REACTOS_SOURCE_DIR}/ntoskrnl/se/tknlife.c
+    ${REACTOS_SOURCE_DIR}/ntoskrnl/se/tknsup.c
```
HBelusca, line 290: looks like these "tkn*" files' names are short enough so that you could use "token" in their names instead (making them a bit more readable), without going over the 8.3 length :)

HBelusca, line 290: dunno.... "tknsup.c" could stay as "token.c" or sthg (as in, having the generic miscellaneous token manipulation functions).

HBelusca (approved):
Splitting these files makes perfect sense for maintainability.

PR #4535, opened 2022-06-03
[NTOS:MM] Fix memory leak in NtAllocateVirtualMemory

JohnyTheCarrot:
When an allocated VAD's insertion fails, the VAD is not freed. This commit attempts to fix this behaviour.


HBelusca (approved):
looks fine.

GeoB99 (approved):

PR #4536, opened 2022-06-03
[CRT] Implement mmintrin.h and xmmintrin.h

tkreuzer:
## Purpose

This change implements all intrinsics in mmintrin.h (MMX) and xmmintrin.h (SSE)
Note: the previous GCC definition of __m128 was broken and didn't work.
It's preparation work for importing some SSE math functions for x64

## Proposed changes

* Implement all mmintrin.h/xmmintrin.h functions as inline wrappers around GCC/Clang builtins
* Move __INTRIN_INLINE definition to mingw32.h, which is always included
* Allow attribute artificial for Clang, too

## TODO

- [ ] Find someone to write tests

### DIFF
--- a/sdk/include/crt/xmmintrin.h
+++ b/sdk/include/crt/xmmintrin.h
@@ -1,25 +1,42 @@
-/**
- * This file has no copyright assigned and is placed in the Public Domain.
- * This file is part of the w64 mingw-runtime package.
- * No warranty is given; refer to the file DISCLAIMER within this package.
+/*
+ * xmmintrin.h
+ *
+ * This file is part of the ReactOS CRT package.
+ *
+ * Contributors:
+ *   Timo Kreuzer (timo.kreuzer@reactos.org)
+ *
+ * THIS SOFTWARE IS NOT COPYRIGHTED
+ *
+ * This source code is offered for use in the public domain. You may
+ * use, modify or distribute it freely.
+ *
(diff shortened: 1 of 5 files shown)


sdk/include/crt/xmmintrin.h
```diff
@@ -1,25 +1,42 @@
...
-#ifdef __clang__
+#if defined(_MM2_FUNCTIONALITY) && !defined(_MM_FUNCTIONALITY)
+#define _MM_FUNCTIONALITY
+#endif
 
-typedef        float __v4sf __attribute__((__vector_size__(16)));
-typedef   signed int __v4si __attribute__((__vector_size__(16)));
-typedef unsigned int __v4su __attribute__((__vector_size__(16)));
+#if !defined _VCRT_BUILD && !defined _INC_MALLOC
+//#include <malloc.h> // FIXME: This breaks build
```
HBelusca, line 32: Why the malloc inclusion? Is it for these `__mm_malloc` and free down the file? (also I see those latter are defined only when `__ICL` is defined; maybe worth adding that here too? Or you didn't do this in order to have the header compatible with an equivalent one from MSVC or another compiler?)

tkreuzer: The malloc inclusion is what MS headers include. I don't see any actual reason for it, other than "his is what Windows does".
But in this case it's what public headers do, so I think it's worth to try to be conforming.

PR #4537, opened 2022-06-04
[TASKMGR] Process page: Fix DevicePathToDosPath

TAN-Gaming:
## Proposed changes
- Fix `DevicePathToDosPath`, don't use the same in-out buffer.
- Also simplify functions related to [59dcec1](https://github.com/reactos/reactos/commit/59dcec1fafcb27ad18b246d34531e3373c8c1012)

### DIFF
--- a/base/applications/taskmgr/procpage.c
+++ b/base/applications/taskmgr/procpage.c
@@ -966,7 +966,7 @@ static DWORD DevicePathToDosPath(LPCWSTR lpDevicePath, LPWSTR lpDosPath, DWORD d
                 /* Get required length, including the NULL terminator */
                 dwRet = _countof(szDrive) + wcslen(lpDevicePath + len);
 
-                if (lpDosPath && dwDosPathLength >= dwRet)
+                if (dwDosPathLength >= dwRet)
                 {
                     StringCchPrintfW(lpDosPath, dwDosPathLength, L"%s%s", szDrive, lpDevicePath + len);
                     
@@ -986,8 +986,8 @@ static DWORD GetProcessExecutablePath(HANDLE hProcess, LPWSTR lpExePath, DWORD d
     BYTE StaticBuffer[sizeof(UNICODE_STRING) + (MAX_PATH * sizeof(WCHAR))];
     PVOID DynamicBuffer = NULL;
     PUNICODE_STRING ExePath;
-    ULONG SizeNeeded;
     LPWSTR pszExePath = NULL;
+    ULONG SizeNeeded;
     NTSTATUS Status;
     DWORD dwRet = 0;
 
@@ -1034,22 +1034,13 @@ static DWORD GetProcessExecutablePath(HANDLE hProcess, LPWSTR lpExePath, DWORD d
     StringCbCopyNW(pszExePath, ExePath->Length + sizeof(WCHAR), ExePath->Buffer, ExePath->Length);
 
     /* HACK: Convert device path format into Win32 path format
-     * 
      * Use ProcessImageFileNameWin32 instead if the kernel
-     * supports it
-     */
+     * supports it */
     dwRet = DevicePathToDosPath(pszExePath, lpExePath, dwLength);
 
 Cleanup:
-    if (pszExePath)
-    {
-        HeapFree(GetProcessHeap(), 0, pszExePath);
-    }
-
-    if (DynamicBuffer)
-    {
-        HeapFree(GetProcessHeap(), 0, DynamicBuffer);
-    }
+    HeapFree(GetProcessHeap(), 0, pszExePath);
+    HeapFree(GetProcessHeap(), 0, DynamicBuffer);
 
     return dwRet;
 }
@@ -1072,11 +1063,11 @@ static DWORD GetProcessExecutablePathById(DWORD dwProcessId, LPWSTR lpExePath, D
 
         uLength = GetSystemDirectoryW(szSystemDir, _countof(szSystemDir));
 
-        if (uLength != 0)
+        if (uLength != 0 && uLength < _countof(szSystemDir))
         {
             dwRet = uLength + _countof(szKernelExe);
 
-            if (lpExePath && dwLength >= dwRet)
+            if (dwLength >= dwRet)
             {
                 StringCchPrintfW(lpExePath, dwLength, L"%s%s", szSystemDir, szKernelExe);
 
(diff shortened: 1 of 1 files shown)


base/applications/taskmgr/procpage.c
```diff
@@ -1150,69 +1149,66 @@ void ProcessPage_OnProperties(void)
...
-    }
-    else
-    {
-        pszExePath = szPath;
-        dwLength += 1;
+        dwLength -= 1;
     }
-
-    pszCmdLine = HeapAlloc(GetProcessHeap(), 0, (dwLength + 10) * sizeof(WCHAR));
+
```
HBelusca, line 1198:
```suggestion

```

base/applications/taskmgr/procpage.c
```diff
@@ -1003,52 +1012,37 @@ static DWORD GetProcessExecutablePath(HANDLE hProcess, LPWSTR lpExePath, DWORD d
...
-    {
-        StringCchCopyW(lpExePath, dwLength, pszExePath);
-
-        dwRet -= 1;
-    }
+    /* HACK: Convert device path format into Win32 path format
+     * 
+     * Use ProcessImageFileNameWin32 instead if the kernel
+     * supports it
+     */
```
HBelusca, line 1040:
```suggestion
    /* HACK: Convert device path format into Win32 path format.
     * Use ProcessImageFileNameWin32 instead if the kernel
     * supports it. */
```

base/applications/taskmgr/procpage.c
```diff
@@ -1003,52 +1012,37 @@ static DWORD GetProcessExecutablePath(HANDLE hProcess, LPWSTR lpExePath, DWORD d
...
+        ExePath = (PUNICODE_STRING)StaticBuffer;
     }
 
     if (!NT_SUCCESS(Status))
     {
         goto Cleanup;
     }
 
-    pszExePath = HeapAlloc(GetProcessHeap(), 0, ImagePath->Length + sizeof(WCHAR));
+    pszExePath = HeapAlloc(GetProcessHeap(), 0, ExePath->Length + sizeof(WCHAR));
```
HBelusca, line 1028:
```suggestion
```

base/applications/taskmgr/procpage.c
```diff
@@ -1103,38 +1097,43 @@ static DWORD GetProcessExecutablePathById(DWORD dwProcessId, LPWSTR lpExePath, D
...
-    LPWSTR pszExePath = NULL;
+    LPWSTR pszExePath;
     DWORD dwLength;
     SHELLEXECUTEINFOW info = { 0 };
 
     dwProcessId = GetSelectedProcessId();
-    dwLength = GetProcessExecutablePathById(dwProcessId, szPath, _countof(szPath));
 
-    if (dwLength == 0)
+    /* Allocate a buffer for the path */
```
HBelusca, line 1106: Since in all cases (this call path, or the one below if `GetProcessExecutablePathById()` returns a longer length than MAX_PATH) you dynamically allocate a buffer, why not just more simply do one single call to `GetProcessExecutablePathById()` in order to retrieve that length ?

base/applications/taskmgr/procpage.c
```diff
@@ -1150,69 +1149,66 @@ void ProcessPage_OnProperties(void)
...
+    LPWSTR pszExePath;
     LPWSTR pszCmdLine = NULL;
     DWORD dwLength;
 
     dwProcessId = GetSelectedProcessId();
-    dwLength = GetProcessExecutablePathById(dwProcessId, szPath, _countof(szPath));
 
-    if (dwLength == 0)
+    /* Allocate a buffer for the path */
+    pszExePath = HeapAlloc(GetProcessHeap(), 0, MAX_PATH * sizeof(WCHAR));
```
HBelusca, line 1165: Since in all cases (this call path, or the one below if `GetProcessExecutablePathById()` returns a longer length than MAX_PATH) you dynamically allocate a buffer, why not just more simply do one single call to `GetProcessExecutablePathById()` in order to retrieve that length ?

base/applications/taskmgr/procpage.c
```diff
@@ -1150,69 +1149,66 @@ void ProcessPage_OnProperties(void)
...
             goto Cleanup;
         }
 
-        pszExePath = pszPath;
-    }
-    else
-    {
-        pszExePath = szPath;
-        dwLength += 1;
+        dwLength -= 1;
```
HBelusca, line 1196: What's this for?

base/applications/taskmgr/procpage.c
```diff
@@ -1103,116 +1095,48 @@ static DWORD GetProcessExecutablePathById(DWORD dwProcessId, LPWSTR lpExePath, D
...
-    LPWSTR pszPath = NULL;
-    LPWSTR pszExePath = NULL;
-    LPWSTR pszCmdLine = NULL;
+    WCHAR szExePath[MAX_PATH];
+    WCHAR szCmdLine[MAX_PATH + 10];
     DWORD dwLength;
 
     dwProcessId = GetSelectedProcessId();
-    dwLength = GetProcessExecutablePathById(dwProcessId, szPath, _countof(szPath));
+    dwLength = GetProcessExecutablePathById(dwProcessId, szExePath, _countof(szExePath));
```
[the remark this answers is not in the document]
TAN-Gaming: It seems pointless to add long path support here because Explorer won't accept long paths anyway.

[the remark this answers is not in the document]
learn-more: That one place is not fixed yet does not mean we shouldn't aim to fix other places,
otherwise this would always be stuck in a 'but some other place is not fixed yet' loop.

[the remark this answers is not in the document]
TAN-Gaming: I understand that MAX_PATH limit is an annoying issue of the Windows platform and it can't be fixed without writing a new set of API functions that support long path so adding support for long path now will make the execution slower and it won't work anyway.

But if you think that using dynamic allocation now is better than rely on MAX_PATH until ROS supports long path, I'll change.

base/applications/taskmgr/procpage.c
```diff
@@ -937,48 +937,57 @@ int CALLBACK ProcessPageCompareFunc(LPARAM lParam1, LPARAM lParam2, LPARAM lPara
...
-                
-                return TRUE;
+                /* Get required length, including the NULL terminator */
+                dwRet = _countof(szDrive) + wcslen(lpDevicePath + len);
+
+                if (dwDosPathLength >= dwRet)
+                {
+                    StringCchPrintfW(lpDosPath, dwDosPathLength, L"%s%s", szDrive, lpDevicePath + len);
+                    
+                    dwRet -= 1;
```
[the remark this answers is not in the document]
TAN-Gaming: Don't count the NULL terminator.

PR #4542, opened 2022-06-06
[WIN32K] Conditionally enable IMM32

learn-more:
MFC42 applications only expect an IME window when on a DBCS system,
so they will capture this IME window as their 'main' window on non-DBCS systems.

JIRA issue: [CORE-18212](https://jira.reactos.org/browse/CORE-18212)

### DIFF
--- a/win32ss/user/ntuser/simplecall.c
+++ b/win32ss/user/ntuser/simplecall.c
@@ -124,7 +124,16 @@ NtUserCallNoParam(DWORD Routine)
             break;
 
         case NOPARAM_ROUTINE_UPDATEPERUSERIMMENABLING:
-            gpsi->dwSRVIFlags |= SRVINFO_IMM32; // Always set.
+            // TODO: This should also check the registry!
+            // see https://www.pctipsbox.com/fix-available-for-ie7-memory-leaks-on-xp-sp3/ for more information
+            if (NLS_MB_CODE_PAGE_TAG)
+            {
+                gpsi->dwSRVIFlags |= SRVINFO_IMM32;
+            }
+            else
+            {
+                gpsi->dwSRVIFlags &= ~SRVINFO_IMM32;
+            }
             Result = TRUE; // Always return TRUE.
             break;
 


katahiromz:
VirtualBox_Windows Server 2003_07_06_2022_08_41_29
I can't see the condition.

win32ss/user/ntuser/simplecall.c
```diff
@@ -124,7 +124,10 @@ NtUserCallNoParam(DWORD Routine)
             break;
 
         case NOPARAM_ROUTINE_UPDATEPERUSERIMMENABLING:
-            gpsi->dwSRVIFlags |= SRVINFO_IMM32; // Always set.
+            if (NLS_MB_CODE_PAGE_TAG)
+            {
+                gpsi->dwSRVIFlags |= SRVINFO_IMM32;
+            }
```
katahiromz, line 130:
```suggestion
            if (NLS_MB_CODE_PAGE_TAG)
                gpsi->dwSRVIFlags |= SRVINFO_IMM32;
            else
                gpsi->dwSRVIFlags &= ~SRVINFO_IMM32;
```

learn-more:
>  I can't see the condition.

I am not sure if that is the same condition as is checked by MFC:

```
AFX_STATIC BOOL PASCAL _AfxInitDBCS()
{
    CPINFO info;
    GetCPInfo(GetOEMCP(), &info);
    return info.MaxCharSize > 1;
}
const AFX_DATADEF BOOL _afxDBCS = _AfxInitDBCS();
```

And it _might_ require a reboot.

At least on my 2k3 there is no IME window created...

jimtabor:
Still need to check the registry if NLS_MB_CODE_PAGE_TAG is false.
Just saying...

jimtabor (approved):
Good start moving from a hack state.

learn-more:
> Still need to check the registry if NLS_MB_CODE_PAGE_TAG is false. Just saying...
> 
> Good read; https://www.pctipsbox.com/fix-available-for-ie7-memory-leaks-on-xp-sp3/

My goal was to fix a compatibility issue, not completely implement IMM,
but thanks for the link!
I'll add that with a comment to the code, to indicate that it needs more work.

HBelusca (approved):

katahiromz (approved):

PR #4545, opened 2022-06-11
[NTOSKRNL] Force a probe against ReturnLength on query & Misc ICIF stuff

GeoB99:
`NtQueryInformationToken` is by far the only system call in NT where `ReturnLength` simply cannot be optional. On Windows this parameter is always probed and an argument to NULL directly leads to an access violation exception.
This is due to the fact of how tokens work, as its information contents (token user, owner, primary group, et al) are dynamic and can vary throughout over time in memory.

What happens on current ReactOS master however is that `ReturnLength` is only probed if the parameter is not NULL. On a NULL case scenario the probing checks succeed and `NtQueryInformationToken` fails later. For this, just get rid of `CompleteProbing`
parameter and opt in for a bit mask flag based approach, with **ICIF_FORCE_RETURN_LENGTH_PROBE** being set on `DefaultQueryInfoBufferCheck` which `NtQueryInformationToken` calls it to do sanity checks.

In addition to that...

- Document the ICIF probe helpers
- Annotate the ICIF prope helpers with SAL
- With the riddance of `CompleteProbing` and adoption of flags based approach, add **ICIF_PROBE_READ_WRITE** and **ICIF_PROBE_READ** flags alongside with **ICIF_FORCE_RETURN_LENGTH_PROBE**

### DIFF
--- a/ntoskrnl/include/internal/icif.h
+++ b/ntoskrnl/include/internal/icif.h
@@ -2,8 +2,7 @@
  * PROJECT:     ReactOS Kernel
  * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
  * PURPOSE:     Internal header for information classes info interface
- * COPYRIGHT:   Copyright ???
- *              Copyright 2020-2021 George Bișoc <george.bisoc@reactos.org>
+ * COPYRIGHT:   Copyright 2020-2022 George Bișoc <george.bisoc@reactos.org>
  */
 
 #pragma once
@@ -21,6 +20,10 @@
 #define ICIF_QUERY_SIZE_VARIABLE 0x4
 #define ICIF_SET_SIZE_VARIABLE   0x8
 #define ICIF_SIZE_VARIABLE (ICIF_QUERY_SIZE_VARIABLE | ICIF_SET_SIZE_VARIABLE)
+
+#define ICIF_PROBE_READ_WRITE          0x0
+#define ICIF_PROBE_READ                0x1
+#define ICIF_FORCE_RETURN_LENGTH_PROBE 0x2
 
 typedef struct _INFORMATION_CLASS_INFO
 {
--- a/ntoskrnl/include/internal/probe.h
+++ b/ntoskrnl/include/internal/probe.h
@@ -1,16 +1,79 @@
+/*
+ * PROJECT:     ReactOS Kernel
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Internal header containing information class probing helpers
+ * COPYRIGHT:   Copyright 2022 George Bișoc <george.bisoc@reactos.org>
+ */
+
 #pragma once
 
 #include <reactos/probe.h>
 
+/**
+ * @brief
+ * Probe helper that validates the provided parameters whenever
+ * a NtSet*** system call is invoked from user or kernel mode.
+ *
+ * @param[in] Class
+ * The specific class information that the caller explicitly
+ * requested information to be set into an object.
+ *
+ * @param[in] ClassList
+ * An internal INFORMATION_CLASS_INFO consisting of a list array
+ * of information classes checked against the requested information
+ * classes given in Class parameter.
+ *
+ * @param[in] ClassListEntries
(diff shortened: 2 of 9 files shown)


GeoB99:
**BEFORE**
Capture2
**AFTER**
Capture

ntoskrnl/include/internal/probe.h
```diff
@@ -1,16 +1,80 @@
+/*
+ * PROJECT:     ReactOS Kernel
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Internal header containing information class probing helpers
+ * COPYRIGHT:   Copyright ???
```
HBelusca, line 5: I guess you were really the first one to tentatively write properly these validation functions, so you may just put your own copyright, skipping that "unknown" entry 😂

ntoskrnl/include/internal/probe.h
```diff
@@ -105,10 +269,11 @@ DefaultQueryInfoBufferCheck(ULONG Class,
                         }
                     }
 
-                    if (ReturnLength != NULL)
+                    if ((Flags & ICIF_FORCE_RETURN_LENGTH_PROBE) || ReturnLength != NULL)
```
HBelusca, line 272:
```suggestion
                    if ((Flags & ICIF_FORCE_RETURN_LENGTH_PROBE) || (ReturnLength != NULL))
```

ntoskrnl/include/internal/icif.h
```diff
@@ -3,7 +3,7 @@
  * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
  * PURPOSE:     Internal header for information classes info interface
  * COPYRIGHT:   Copyright ???
```
HBelusca, line 5: See similar comment in probe.h

ntoskrnl/include/internal/probe.h
```diff
@@ -1,16 +1,80 @@
...
+ * this function will only validate the properties of the information class itself
+ * such as the required information length size. If the calling mode is UM, the
+ * pointer buffer provided by Buffer parameter is also validated.
+ *
+ * @return
+ * The outcome of the probe validation is based upon the returned NTSTATUS code.
+ * That is, STATUS_SUCCESS is returned if the validation in question has succeeded.
+ * Otherwise one of the following failure status codes is returned:
+ *
+ * STATUS_INVALID_INFO_CLASS -- This status code indicates the given information
```
HBelusca, line 53: You can improve the doc by removing "This status code" in each of these below, and start by just "Indicates ...", since we know it's a status code -- it's told so in the explanation text above.

ntoskrnl/include/internal/probe.h
```diff
@@ -1,16 +1,80 @@
...
+ * The processor calling level mode. This level mode determines the procedure
+ * of probing validation in action. If the level calling mode is KM (aka kernel mode)
+ * this function will only validate the properties of the information class itself
+ * such as the required information length size. If the calling mode is UM, the
+ * pointer buffer provided by Buffer parameter is also validated.
+ *
+ * @return
+ * The outcome of the probe validation is based upon the returned NTSTATUS code.
+ * That is, STATUS_SUCCESS is returned if the validation in question has succeeded.
+ * Otherwise one of the following failure status codes is returned:
```
HBelusca, line 51:
```suggestion
 * STATUS_SUCCESS is returned if the validation succeeded. Otherwise, one of the
 * following failure status codes is returned:
```

ntoskrnl/include/internal/probe.h
```diff
@@ -53,18 +117,118 @@ DefaultSetInfoBufferCheck(ULONG Class,
...
+ * The processor calling level mode. This level mode determines the procedure
+ * of probing validation in action. If the level calling mode is KM (aka kernel mode)
+ * this function will only validate the properties of the information class itself
+ * such as the required information length size. If the calling mode is UM, the
+ * pointer buffer provided by Buffer parameter is also validated as well as
+ * the return length parameter.
+ *
+ * @return
+ * The outcome of the probe validation is based upon the returned NTSTATUS code.
+ * That is, STATUS_SUCCESS is returned if the validation in question has succeeded.
```
HBelusca, line 179: Same doc comments as for the other function.

PR #4546, opened 2022-06-12
[NTOS:SE] Implement token sandbox inert querying

GeoB99:
Since there's support for restricted tokens in the kernel which also supports for sandbox inert tokens as well, just implement the querying part outright, aka `TokenSandBoxInert` class.

### DIFF
--- a/modules/rostests/apitests/ntdll/NtQueryInformationToken.c
+++ b/modules/rostests/apitests/ntdll/NtQueryInformationToken.c
@@ -19,7 +19,7 @@ OpenCurrentToken(VOID)
                                &Token);
     if (!Success)
     {
-        ok(0, "OpenProcessToken() has failed to get the process' token (error code: %lu)!\n", GetLastError());
+        ok(FALSE, "OpenProcessToken() has failed to get the process' token (error code: %lu)!\n", GetLastError());
         return NULL;
     }
 
@@ -52,7 +52,7 @@ QueryTokenUserTests(
     UserToken = RtlAllocateHeap(RtlGetProcessHeap(), 0, BufferLength);
     if (!UserToken)
     {
-        ok(0, "Failed to allocate from heap for token user (required buffer length %lu)!\n", BufferLength);
+        ok(FALSE, "Failed to allocate from heap for token user (required buffer length %lu)!\n", BufferLength);
         return;
     }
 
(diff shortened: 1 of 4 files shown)


GeoB99:
**Windows Server 2003 SP2**
Capture
**ReactOS**
Capture2

modules/rostests/apitests/ntdll/NtQueryInformationToken.c
```diff
@@ -617,6 +617,59 @@ QueryTokenSessionIdTests(
...
+     * Query the sandbox inert token information,
+     * it must not be inert.
+     */
+    Status = NtQueryInformationToken(Token,
+                                     TokenSandBoxInert,
+                                     &IsTokenInert,
+                                     sizeof(ULONG),
+                                     &BufferLength);
+    ok_ntstatus(Status, STATUS_SUCCESS);
+    ok(IsTokenInert == FALSE, "The token mustn't be a sandbox inert one!\n");
```
HBelusca, line 640:
```suggestion
    ok(IsTokenInert == FALSE, "The token must not be a sandbox inert one!\n");
```

modules/rostests/apitests/ntdll/NtQueryInformationToken.c
```diff
@@ -617,6 +617,59 @@ QueryTokenSessionIdTests(
...
+     */
+    Status = NtFilterToken(Token,
+                           SANDBOX_INERT,
+                           NULL,
+                           NULL,
+                           NULL,
+                           &FilteredToken);
+    if (!NT_SUCCESS(Status))
+    {
+        ok(0, "Failed to filter the current token (Status code %lx)!\n", Status);
```
HBelusca, line 654:
```suggestion
        ok(FALSE, "Failed to filter the current token (Status code %lx)!\n", Status);
```
or use skip() ?

HBelusca (approved):

modules/rostests/apitests/ntdll/NtQueryInformationToken.c
GeoB99: `skip()` is the first thing that came up in my mind for such stuff but someone else suggested me to rather fail than skip the test with `ok(0, ...)` when I wrote the tests... :trollface:

HBelusca: yes I may understand the logic there: skip() would be more appropriate in the cases of a test to be skipped because e.g. an API is missing in ROS/Windows version being tested, or as a temporary skip until the test is fixed.
While ok(FALSE) would be more like for something that should never fail.

PR #4549, opened 2022-06-18
Crt math/win libm

tkreuzer:
## Purpose

Implement SSE math functions b importing AMD's libm.

## Proposed changes

- Import win-libm from AMD
- Make it build
- Convert .asm tables to C
- Write a tool to convert MASM style syntax to GAS compatible syntax

## Motivation
On x64 we lacked many of these functions and they need to be SSE(2). Windows is using libm, too (probably a slightly customized version). Windows also adds these functions for x86 msvcrt in Vista+ (__libm_sse2_*), where we can reuse them later.


encodedpr (approved):

julenuri:
VirtualBox_ReactOS 64_02_09_2022_21_05_01
Under MSVC 14.0. But a good WIP. 
Lot of my x64 tickets involved, and lot of them fixed under MSVC 14.2. 
At least: 

- CORE-17639
- CORE-18156
- CORE-18155
- CORE-18234
- CORE-18310

Pending for more.

sdk/lib/crt/math/libm_sse2/libm_errno.h
```diff
@@ -0,0 +1,35 @@
...
+/** THE SOFTWARE. **/
+/***********************************************************************************/
+
+#ifndef LIBM_ERRNO_AMD_H_INCLUDED
+#define LIBM_ERRNO_AMD_H_INCLUDED 1
+
+#include <stdio.h>
+#include <errno.h>
+#ifndef __set_errno
+//#define __set_errno(x) errno = (x)
```
learn-more, line 32: Why is this commented out?

sdk/tools/asmpp/asmpp.sln
```diff
@@ -0,0 +1,31 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
```
learn-more, line 2: Did you leave those in for easier debugging, or did you forget to delete em?

julenuri:
> VirtualBox_ReactOS 64_02_09_2022_21_05_01 
> Under MSVC 14.0. But a good WIP. 


Now fixed the MSVC 14.0, 14.1 Freeldr Image Corruption. I must do some tests, and I don't know how much pollish needs this amount of commits, but well done @tkreuzer !
msvc140

boot/freeldr/freeldr/freeldr.c
```diff
@@ -98,3 +98,29 @@ int __cdecl iswctype(wint_t wc, wctype_t wctypeFlags)
...
 }
+
+#ifdef _MSC_VER
+#pragma warning(disable:4164)
+#pragma function(pow)
+#pragma function(log)
+#pragma function(log10)
+#endif
+
+// Stubs to avoid pulling in data from CRT
```
HBelusca, line 109: What about building a CRT tailored for the kernel and/or freeldr, that doesn't use any of those functions?

sdk/tools/asmpp/asmpp.cpp
```diff
@@ -0,0 +1,1152 @@
...
+ * COPYRIGHT:   Copyright 2021 Timo Kreuzer <timo.kreuzer@reactos.org>
+ */
+
+// Optimize even on debug builds, because otherwise it's ridiculously slow
+#ifdef _MSC_VER
+#pragma optimize("gst", on)
+#pragma auto_inline(on)
+#else
+#pragma GCC optimize("O3,inline")
+#endif
```
HBelusca, line 14: Can this be added within the CMakeLists file?

sdk/tools/asmpp/asmpp.cpp
```diff
@@ -0,0 +1,1152 @@
...
+    { TOKEN_TYPE::Identifier, R"((@@))" FOLLOWED_BY(SEPARATOR)},
+    { TOKEN_TYPE::Identifier, R"((@[a-z_][a-z0-9_]*))" FOLLOWED_BY(SEPARATOR)},
+    { TOKEN_TYPE::Identifier, R"(([a-z_][a-z0-9_]*))" FOLLOWED_BY(SEPARATOR)},
+
+};
+
+// FIXME: use context?
+unsigned int g_label_number = 0;
+
+vector<string> g_identifiers;
```
HBelusca, line 311:
```suggestion
std::vector<string> g_identifiers;
```
(the other vector above had the std namespaec)
Or remove the `std::` in that other vector.

sdk/tools/asmpp/asmpp.cpp
```diff
@@ -0,0 +1,1152 @@
...
+            index += 3;
+            break;
+        }
+
+        case TOKEN_TYPE::KW_RECORD:
+            index = translate_record(tokens, index, macro_params);
+            break;
+
+        default:
+            // We don't know what it is, assume it's a macro and trat it like an instruction
```
HBelusca, line 906:
```suggestion
            // We don't know what it is, assume it's a macro and treat it like an instruction
```

sdk/tools/asmpp/asmpp.cpp
```diff
@@ -0,0 +1,1152 @@
...
+        case TOKEN_TYPE::KW_ALIGN:
+            index = translate_construct_one_param(".align", tokens, index, macro_params);
+            break;
+
+        case TOKEN_TYPE::KW_allocstack:
+            index = translate_construct_one_param(".seh_stackalloc", tokens, index, macro_params);
+            break;
+
+        case TOKEN_TYPE::KW_code:
+#ifdef TARGET_amd64
```
HBelusca, line 938: Make the target a runtime option?

sdk/tools/asmpp/asmpp.cpp
```diff
@@ -0,0 +1,1152 @@
...
+            printf("//");
+            return complete_line(tokens, index, macro_params);
+        }
+
+        case TOKEN_TYPE::KW_if:
+        case TOKEN_TYPE::KW_ifdef:
+        case TOKEN_TYPE::KW_ifndef:
+        case TOKEN_TYPE::KW_else:
+        case TOKEN_TYPE::KW_endif:
+            // FIXME: not really good
```
HBelusca, line 981: What would be one better solution?

sdk/tools/asmpp/asmpp.cpp
```diff
@@ -0,0 +1,1152 @@
...
+}
+
+size_t
+translate_macro(TokenList& tokens, size_t index)
+{
+    vector<string> macro_params;
+
+    printf(".macro %s", tokens[index].str().c_str());
+
+    // parse marameters
```
HBelusca, line 1043:
```suggestion
    // Parse parameters
```

sdk/tools/asmpp/asmpp.cpp
```diff
@@ -0,0 +1,1152 @@
...
+            case TOKEN_TYPE::WhiteSpace:
+            case TOKEN_TYPE::Operator:
+                index = translate_token(tokens, index, macro_params);
+                continue;
+        }
+
+        break;
+    }
+
+    // parse content
```
HBelusca, line 1070:
```suggestion
    // Parse content
```

sdk/tools/asmpp/asmpp.cpp
```diff
@@ -0,0 +1,1152 @@
...
+
+#include "tokenizer.hpp"
+#include <cstdlib>
+#include <cstdio>
+#include <sstream>
+#include <ctime>
+
+#ifdef _MSC_VER
+#define throw __debugbreak(); throw
+#endif
```
HBelusca, line 24: This is already defined in the tokenizer.hpp file.

sdk/tools/asmpp/tokenizer.hpp
```diff
@@ -0,0 +1,248 @@
...
+
+#include <string>
+#include <vector>
+#include <fstream>
+#include <regex>
+#include <ctime>
+
+#ifdef _MSC_VER
+#define throw __debugbreak(); throw
+#endif
```
HBelusca, line 16: Maybe surround that with an `#if _DEBUG` too?
(also, should this be `#if` or `#ifdef` ?)

sdk/lib/crt/math/libm_sse2/libm_errno.h
tkreuzer: Good question. I'll revert that.

sdk/tools/asmpp/asmpp.cpp
JoachimHenze: If you add the explicit namespace for the vector, you should do the same for the string also.

boot/freeldr/freeldr/freeldr.c
tkreuzer: I guess we could do that, but that would mean 3 CRTs: CRT, LIBCNT_UM, LIBCNT_KM. Preferably this should avoid recompiling code, so you would have chains: CRT_COMMON<-CRT / CRT_COMMON<-LIBCNT_COMMON<-LIBCNT_[KM|UM]. That can quickly get messy. I might look at this in the future, but for this PR it's just too much extra hassle.

JoachimHenze:
Could you please add to the description what is the intended usage of those SSE2 math functions? Just because they are faster, or is this any requirement for some specific follow-up that you have in mind? Thx.

sdk/tools/asmpp/asmpp.cpp
tkreuzer: Probably, but what's the argument for doing that?

sdk/tools/asmpp/asmpp.cpp
tkreuzer: I'll remove the std consistently.

sdk/tools/asmpp/asmpp.cpp
tkreuzer: What's the use case?

sdk/tools/asmpp/asmpp.cpp
tkreuzer: It's a preliminary solution that works with the code I have, but there are some differences in how MASM"if" and GAS ".if" work with the parameters. I'll improve the comment to make that more clear.

sdk/tools/asmpp/tokenizer.hpp
tkreuzer: I'll use `#if 0`, since it's really only when you actively debug it.

sdk/tools/asmpp/asmpp.sln
tkreuzer: Easier debugging. I guess it should work with CMaking a solution file as well. So if prefered, I can remove them.

tkreuzer:
@JoachimHenze 
> Could you please add to the description what is the intended usage of those SSE2 math functions? Just because they are faster, or is this any requirement for some specific follow-up that you have in mind? Thx.
> (Alternatively) referencing a specific CORE-ID which would get fixed in the description would serve the same purpose of "justification".
1. We lacked most of these functions on x64
2. They need to be SSE2
3. There is no point in re-implementing them, when there is already a proper OSS version
4. We'll need them for x86 for Vista+ later
5. Windows uses libm, too, so they should behave more or less exactly the same, including the error handling
Is that enough "justification" for you?

sdk/tools/asmpp/asmpp.sln
learn-more: Up to you.

sdk/cmake/gcc.cmake
```diff
@@ -467,8 +467,37 @@ function(allow_warnings __module)
...
+        if (${_extension} STREQUAL ".asm")
+            convert_asm_file(${_source_file} ${_source_file}.s)
+            list(APPEND ${_target} ${CMAKE_CURRENT_BINARY_DIR}/${_source_file}.s)
+        elseif (${_extension} STREQUAL ".inc")
+            convert_asm_file(${_source_file} ${_source_file}.h)
+            list(APPEND ${_target} ${CMAKE_CURRENT_BINARY_DIR}/${_source_file}.h)
+        else()
+            list(APPEND ${_target} ${_source_file})
+        endif()
+    endforeach()
```
learn-more, line 500: Does this track dependencies correctly?
e.g. does it rebuild the target if one of the source files changes?

tkreuzer: Yes, it does.

PR #4551, opened 2022-06-18
[WIN32SS][NTUSER] Load DirectX graphics driver at system startup

oleg-dubinskiy:
## Purpose

Load DirectX graphics kernel driver (dxg.sys) by our win32k when the system starts up. Keep it always loaded in memory, like done in Windows, instead of loading only by DirectX dlls. It fixes the problem with accessing to this driver, so then we need only call `DxDdEnableDirectDraw` and do other stuff when DirectDraw/Direct3D is required by anything. In other cases, it is called from win32k PDEV* functions when changing display mode (as in Windows). Since it's used by other things too, it needs to be always loaded.
Otherwise, if it's not loaded, its APIs are not accessible when needed, and then execution fails.
It fixes display mode change problem in VMWare, when a new mode fails to be applied, because when it manages DirectDraw stuff, it calls DXG routines, and therefore fails if dxg.sys isn't loaded in memory at the moment.

JIRA issue: [CORE-18221](https://jira.reactos.org/browse/CORE-18221)

## Proposed changes

- Implement `InitializeGreCSRSS()` initialization routine. Call win32k!`DxDdStartupDxGraphics` inside it, which loads dxg.sys.
Additionally, move fonts and language ID initialization into `InitializeGreCSRSS()` from win32k!`DriverEntry`, as it is implemented in Windows. My analysis confirms that `DriverEntry()` does not initialize fonts stuff and language ID. It's done in `IntitializeGreCSRSS()` instead.
- Call `InitializeGreCSRSS()` in `NtUserInitialize()` main initialization routine. My analysis confirms that this is correct, regarding the Windows implementation. :slightly_smiling_face: 

**NOTE 2: To see the bug, #4519 PR should also be applied.**

### DIFF
--- a/win32ss/gdi/ntgdi/init.c
+++ b/win32ss/gdi/ntgdi/init.c
@@ -7,10 +7,9 @@
  */
 
 #include <win32k.h>
+DBG_DEFAULT_CHANNEL(UserMisc);
 
-#define NDEBUG
-#include <debug.h>
-#include <kdros.h>
+USHORT gusLanguageID;
 
 BOOL NTAPI GDI_CleanupForProcess(struct _EPROCESS *Process);
 
@@ -76,6 +75,29 @@ GdiThreadDestroy(PETHREAD Thread)
     return STATUS_SUCCESS;
 }
 
+
+BOOL
+InitializeGreCSRSS(VOID)
+{
+    /* Initialize DirectX graphics driver */
+    if (DxDdStartupDxGraphics(0, NULL, 0, NULL, NULL, gpepCSRSS) != STATUS_SUCCESS)
+    {
+        ERR("Unable to initialize DirectX graphics\n");
+        return FALSE;
+    }
+
+    /* Get global language ID */
+    gusLanguageID = UserGetLanguageID();
+
+    /* Initialize FreeType library */
+    if (!InitFontSupport())
+    {
+        ERR("Unable to initialize font support\n");
+        return FALSE;
+    }
+
+    return TRUE;
+}
 
 /*
  * @implemented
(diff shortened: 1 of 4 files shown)


win32ss/user/ntuser/csr.c
```diff
@@ -63,6 +63,20 @@ ResetCsrApiPort(VOID)
...
+BOOL
+InitializeGreCSRSS(VOID)
+{
+    /* Initialize DirectX graphics driver */
+    if (DxDdStartupDxGraphics(0, NULL, 0, NULL, NULL, gpepCSRSS) != STATUS_SUCCESS)
+        return FALSE;
+
+    // FIXME: it's not the full implementation of the function.
+    //        We should call UserGetLanguageID() and sets its data.
+    //        Also enable fonts drivers, initialize font table & stock fonts.
```
HBelusca, line 75: Currently, `UserGetLanguageID()` and the font support is initialized at the end of our win32k's `DriverEntry`, see https://git.reactos.org/?p=reactos.git;a=blob;f=win32ss/user/ntuser/main.c;hb=568383c9b91c2e529ae4a3609e416ec8044cfc23#l1053

HBelusca: In Windows there's still some global GRE initialization that happens in DriverEntry() (and not in the NtUserInitialize); you have to check what it is, though...

oleg-dubinskiy: Also additional note: I did not see any problems/regressions after my latest changes, at least in VMWare.

HBelusca (approved):

win32ss/gdi/ntgdi/init.c
```diff
@@ -7,10 +7,13 @@
...
 
 #include <win32k.h>
+DBG_DEFAULT_CHANNEL(UserMisc);
 
-#define NDEBUG
-#include <debug.h>
-#include <kdros.h>
+//#define NDEBUG
+//#include <debug.h>
+//#include <kdros.h>
```
HBelusca, line 14: It looks like those things are unneeded in this file. If compilation succeeds without these, I'll clean them up.

PR #4557, opened 2022-06-23
Remove references to ReactOS Foundation

garyexplains:
## Purpose

Remove leftover references to "ReactOS Foundation"

JIRA issue: [CORE-18191](https://jira.reactos.org/browse/CORE-18191)

## Proposed changes

Changed references to "ReactOS Foundation" to "ReactOS Project"

## TODO

Some files use "ReactOS Foundation" in reference to legal trademarks, like "ReactOS(R) is a registered trademark of the ReactOS Foundation." I didn't change those yet, I will ask in IRC.

### DIFF
--- a/boot/boot_images.cmake
+++ b/boot/boot_images.cmake
@@ -56,8 +56,8 @@ ${_efisys_file} 1
 ")
 
 # ISO image identifier names
-set(ISO_MANUFACTURER "ReactOS Foundation")  # For both the publisher and the preparer
-set(ISO_VOLNAME      "ReactOS")             # For both the Volume ID and the Volume set ID
+set(ISO_MANUFACTURER "ReactOS Project") # For both the publisher and the preparer
+set(ISO_VOLNAME      "ReactOS")         # For both the Volume ID and the Volume set ID
 
 
 # Create user profile directories in the LiveImage
--- a/drivers/processor/processr/cpu.inf
+++ b/drivers/processor/processr/cpu.inf
@@ -120,7 +120,7 @@ LoadOrderGroup = Extended Base
 ;-------------------------------- STRINGS -------------------------------
 
 [Strings]
-ReactOS = "ReactOS Foundation"
+ReactOS = "ReactOS Team"
 ProcessorClassName = "Processors"
 
 CyrixMfg = "Cyrix"
--- a/media/themes/Blackshade/blackshade.msstyles/lang/de-DE.rc
+++ b/media/themes/Blackshade/blackshade.msstyles/lang/de-DE.rc
@@ -28,7 +28,7 @@ STRINGTABLE
 {
 5000, "Black Shade"
 5001, "Visueller Stil"
-5002, "ReactOS Foundation"
+5002, "ReactOS Project"
 5003, "Grafiken erstellt durch David Quintana <gigaherz@gmail.com>"
 5004, "Grafiken veröffentlicht unter CC-BY-SA 4.0"
 5005, "https://reactos.org/"
--- a/media/themes/Blackshade/blackshade.msstyles/lang/en-US.rc
+++ b/media/themes/Blackshade/blackshade.msstyles/lang/en-US.rc
@@ -28,7 +28,7 @@ STRINGTABLE
 {
 5000, "Black Shade"
 5001, "Visual Style"
-5002, "ReactOS Foundation"
+5002, "ReactOS Project"
 5003, "Images by David Quintana <gigaherz@gmail.com>"
 5004, "Images released under CC-BY-SA 4.0"
 5005, "https://reactos.org/"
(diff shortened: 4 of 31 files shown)


boot/boot_images.cmake
```diff
@@ -56,7 +56,7 @@ ${_efisys_file} 1
 ")
 
 # ISO image identifier names
-set(ISO_MANUFACTURER "ReactOS Foundation")  # For both the publisher and the preparer
+set(ISO_MANUFACTURER "ReactOS Project")  # For both the publisher and the preparer
 set(ISO_VOLNAME      "ReactOS")             # For both the Volume ID and the Volume set ID
```
HBelusca, line 60:
```suggestion
set(ISO_MANUFACTURER "ReactOS Project") # For both the publisher and the preparer
set(ISO_VOLNAME      "ReactOS")         # For both the Volume ID and the Volume set ID
```

HBelusca:
> Some files use "ReactOS Foundation" in reference to legal trademarks, like "ReactOS(R) is a registered trademark of the ReactOS Foundation." I didn't change those yet, I will ask in IRC.

These files are:
https://git.reactos.org/?p=reactos.git;a=blob;f=sdk/include/psdk/common.ver;hb=3a72a52ce886c5a1bfa1e87b1a9b759e85a5c3d4#l19
https://git.reactos.org/?p=reactos.git;a=blob;f=sdk/include/psdk/ntverp.h;hb=3a72a52ce886c5a1bfa1e87b1a9b759e85a5c3d4#l150

Cc @ColinFinck for advice there.

drivers/processor/processr/cpu.inf
```diff
@@ -121,7 +121,7 @@ LoadOrderGroup = Extended Base
 
 [Strings]
 
-ReactOS = "ReactOS Foundation"
+ReactOS = "ReactOS Project"
```
binarymaster, line 124: In the INF files we're using "ReactOS Team" in all other places, so let's keep this in sync.
```suggestion
ReactOS = "ReactOS Team"
```
Example: https://github.com/reactos/reactos/blob/1225d4d3922343b6590e85f2c74d4edac8c61c6f/media/inf/usbport.inf#L119

drivers/processor/processr/cpu.inf
```diff
@@ -121,7 +121,7 @@ LoadOrderGroup = Extended Base
 
 [Strings]
 
-ReactOS = "ReactOS Foundation"
+ReactOS = "ReactOS Team"
```
ColinFinck, line 124: See above. A consistent "ReactOS Project" everywhere is better after all.

drivers/processor/processr/cpu.inf
ColinFinck: On the contrary, the idea of CORE-18191 is to have a single consistent naming everywhere to hopefully never deal with this negligible issue ever again.
I therefore like to see all occurrences of "ReactOS Team" being replaced with "ReactOS Project" as well.

Sorry for the confusion that happened here!

ColinFinck (changes requested):
Thanks for taking care of this thankless issue!
I left a few comments in the hope to bring some clarity into this.

Regarding `ntverp.h`: My suggestion would be as follows:
```
#define VER_COMPANYNAME_STR                 \
    "ReactOS Project"
#define VER_PRODUCTNAME_STR                 \
    "ReactOS Operating System"
#define VER_LEGALTRADEMARKS_STR             \
    "ReactOS is a registered trademark of ReactOS Deutschland e.V."
```

(I don't think the "(R)" adds anything beneficial; it rather hampers reading for anyone who really reads looks into these strings)

drivers/processor/processr/cpu.inf
binarymaster: We already have most translations for "ReactOS Team" in the INF files, this change would be breaking one. If we want to keep consistency so much, I think it would be better to do in a separate commit.

drivers/processor/processr/cpu.inf
binarymaster: We already have most translations for "ReactOS Team" in the INF files, this change would be breaking one. If we want to keep consistency so much, I think it would be better to do in a separate commit.

ColinFinck: I see. My point is: We currently put "ReactOS Team" or "ReactOS Foundation" where original Windows has "Microsoft Corporation".
That "Microsoft Corporation" string is never translated, and I want to establish "ReactOS Project" as the same consistent never-translated "company name". Sure, we have ReactOS Deutschland e.V., but even after 13 years, our association exists mostly next to the project whenever a legal entity is required (e.g. for the trademark), but it doesn't _own_ the ReactOS Project.

As such, it may have been a mistake in the first place to translate the ReactOS/ReactOS Team string.

PR #4558, opened 2022-06-25
[CDMAKE] Remove module

binarymaster:
As said by Colin in PR #4557:

> We've been using mkisofs for 5 years now and are not looking back.

If nobody has objections, I'm going to merge this in 2 days.


learn-more (approved):

HBelusca (approved):
OK let's delete our past ^^
Btw @binarymaster remember to update the "retired modules" wiki page after this gets committed.

binarymaster:
> OK let's delete our past ^^

Git history remembers everything anyway.

> remember to update the "retired modules" wiki page after this gets committed.

Sure :+1:

ColinFinck (approved):
@HBelusca wrote:
> OK let's delete our past ^^

Sadly, that's the only option for projects that have been part of a larger monorepo for their entire lifetime.
If cdmake was invented after our move to GitHub, it would have probably lived in its own repo. Then we could just flip GitHub's "Archive" option and signal to everyone that development on cdmake has ceased, but keep the source code easily available. Might be another argument in favor of an eventual split of our monorepo.

For now, I'm fine with removing cdmake. I don't see it providing any value over mkisofs.
And if someone ever misses it, they may resurrect the source from the repo and put it in an own repo.

PR #4559, opened 2022-06-25
[CRT] Add emmintrin.h

tkreuzer:
## Purpose

Implement intrinsics in emmintrin.h

## TODO

- [x] Make sure it compiles everywhere
- [x] Remove hack


HBelusca:
It seems it compiles at least :+1:
Does LLVM have a test suite for these kind of functions?

PR #4567, opened 2022-07-01
[SHELL32] Fix PrivateExtractIconExW behavior

rndtrash:
## Purpose

This PR fixes PrivateExtractIconExW behavior, thus makes the shell32:ExtractIconEx test suite complete without any failed tests (https://reactos.org/testman/detail.php?id=59598012&prev=0)

## Proposed changes

- Make PrivateExtractIconExW return the number of icons if neither of the icon pointers are valid.

I have changed the nIndex in https://github.com/reactos/reactos/blob/master/modules/rostests/apitests/shell32/ExtractIconEx.cpp#L34 to arbitrary values from -999 to 999, ran the executable on Windows XP SP3 and Windows 10 and I came to conclusion that, in case both phIconLarge and phIconSmall are NULL, Windows actually doesn't care about the value of nIndex at all.

### DIFF
--- a/win32ss/user/user32/misc/exticon.c
+++ b/win32ss/user/user32/misc/exticon.c
@@ -814,7 +814,14 @@ UINT WINAPI PrivateExtractIconExW (
 	TRACE("%s %d %p %p %d\n",
 	debugstr_w(lpwstrFile),nIndex,phIconLarge, phIconSmall, nIcons);
 
-	if (nIndex == -1)
+        /*
+         * https://docs.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-extracticonexa
+         * "If this value is –1 and phiconLarge and phiconSmall are **both** NULL, the function returns the total number of icons in the specified file.
+         * If the file is an executable file or DLL, the return value is the number of RT_GROUP_ICON resources."
+         *
+         * BUT apparently it doesn't matter what number is stored in nIndex as long as neither of the icon pointers are valid.
+         */
+	if (nIndex == -1 || !(phIconLarge || phIconSmall))
 	  /* get the number of icons */
 	  return ICO_ExtractIconExW(lpwstrFile, NULL, 0, 0, 0, 0, NULL, LR_DEFAULTCOLOR);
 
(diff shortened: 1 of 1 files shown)


SergeGautherie:
Your change might be correct, but it looks odd as is: your code uses 'or' while MS doc says 'and'. And what about line 802 comment?

win32ss/user/user32/misc/exticon.c
```diff
@@ -814,7 +814,14 @@ UINT WINAPI PrivateExtractIconExW (
...
+        /*
+         * https://docs.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-extracticonexa
+         * "If this value is –1 and phiconLarge and phiconSmall are **both** NULL, the function returns the total number of icons in the specified file.
+         * If the file is an executable file or DLL, the return value is the number of RT_GROUP_ICON resources."
+         *
+         * BUT apparently it doesn't matter what number is stored in nIndex as long as neither of the icon pointers are valid.
+         */
+	if (nIndex == -1 || !(phIconLarge || phIconSmall))
 	  /* get the number of icons */
 	  return ICO_ExtractIconExW(lpwstrFile, NULL, 0, 0, 0, 0, NULL, LR_DEFAULTCOLOR);
```
SergeGautherie, line 826:
```suggestion
	if (nIndex == -1 || !(phIconLarge || phIconSmall))
	{
	  /* get the number of icons */
	  return ICO_ExtractIconExW(lpwstrFile, NULL, 0, 0, 0, 0, NULL, LR_DEFAULTCOLOR);
	}
```

SergeGautherie:
In addition, this is Wine code, which needs `#ifdef __REACTOS__`.

rndtrash:
> Your change might be correct, but it looks odd as is: your code uses 'or' while MS doc says 'and'.

Yeah, that seems odd to me, but that is what the Windows does. Here is a modified version of the shell32:ExtractIconEx test that succeeds on Windows XP:

```cpp
/*
 * PROJECT:     ReactOS API Tests
 * PURPOSE:     Tests for ExtractIconEx routine
 * COPYRIGHT:   Copyright 2019 George Bișoc (george.bisoc@reactos.org)
 */

#include "shelltest.h"

typedef struct
{
    PCWSTR pszFilePath;
    UINT nIcons;
} EXTRACTICONTESTS;

EXTRACTICONTESTS IconTests[] =
{
    /* Executable file with icon */
    {L"%SystemRoot%\\System32\\cmd.exe", 1},

    /* Executable file without icon */
    {L"%SystemRoot%\\System32\\autochk.exe", 0},

    /* Non-existing files */
    {L"%SystemRoot%\\non-existent-file.sdf", 0}
};

START_TEST(ExtractIconEx)
{
    UINT i, nReturnedIcons;

    HICON decoy = NULL;

    for (i = 0; i < _countof(IconTests); ++i)
    {
        nReturnedIcons = ExtractIconExW(IconTests[i].pszFilePath, -1, &decoy, NULL, IconTests[i].nIcons);
        ok(nReturnedIcons == IconTests[i].nIcons, "ExtractIconExW(%u): Expected %u icons, got %u\n", i, IconTests[i].nIcons, nReturnedIcons);
        printf("decoy pointer: %p\n", decoy);
    }

    printf("shell32.dll icon count: %u\n", ExtractIconExW(L"%SystemRoot%\\System32\\shell32.dll", 999, NULL, NULL, -5));
}
```

Снимок экрана_2022-07-02_14-05-23

I've got a feeling that the MS docs are not correct.

> As in: maybe simply merge and simplify these comments to tell what your code actually do.

Oh.

> And what about line 802 comment?

Statement on the line 802 is still correct.

> In addition, this is Wine code, which needs `#ifdef __REACTOS__`. Or to be fixed upstream first.

Well, then I guess I'll close this PR and try submitting this patch to Wine.

PR #4569, opened 2022-07-02
[NTOSKRNL] Fix a bug in MiDeleteVirtualAddresses

tkreuzer:
When a PDE gets empty, we skip the address to the next PDE boundary, which might introduce an AddressGap, which wasn't handled before.

### DIFF
--- a/ntoskrnl/mm/ARM3/virtual.c
+++ b/ntoskrnl/mm/ARM3/virtual.c
@@ -714,10 +714,15 @@ MiDeleteVirtualAddresses(IN ULONG_PTR Va,
                 if (MiDecrementPageTableReferences((PVOID)Va) == 0)
                 {
                     ASSERT(PointerPde->u.Long != 0);
+
                     /* Delete the PDE proper */
                     MiDeletePde(PointerPde, CurrentProcess);
-                    /* Jump */
+
+                    /* Continue with the next PDE */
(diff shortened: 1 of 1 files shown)


GeoB99 (approved):
LGTM

zefklop (approved):

ntoskrnl/mm/ARM3/virtual.c
```diff
@@ -733,8 +738,8 @@ MiDeleteVirtualAddresses(IN ULONG_PTR Va,
 
         if (Va > EndingAddress) return;
 
-        /* Otherwise, we exited because we hit a new PDE boundary, so start over */
-        AddressGap = FALSE;
+        /* Check if we exited the loop regularly */
+        AddressGap = (PointerPte != MiAddressToPte(Va));
```
HBelusca, line 742: Question: If this is just for handling the `Use this to detect address gaps` part in the modified if-block above, wouldn't it then be simpler to do a `AddressGap = TRUE;` inside that block?
Or, is this condition also covering the fact that `PointerPte` is also increased in the main do-while loop, and an address gap can develop there as well?

tkreuzer: First there is not neccessarily a gap, so you can't just set it to TRUE, but you need to compare the incremented PointerPte with the new address. Also when you set it there, it will break out of the loop and then reset it back to false.

PR #3932, opened 2021-08-28
[CMLIB] Registry transactional writes.

mrmks04:
## Purpose

Implemented restore registry from log file

JIRA issue: [CORE-6762](https://jira.reactos.org/browse/CORE-6762)

## Proposed changes

- Changed create flags for log files
- Changed log write function

PR #4571, opened 2022-07-03
[NTOS:CM] Implement registry validation checks & registry healing

GeoB99:
`CmCheckRegistry` is a vital kernel routine that provides the brains and heart mechanism of registry validation, namely the CM hive, the hive and corresponding bins and cells and whatnot. `CmCheckRegistry` comes into action when logs have been replayed and more specifically when certain actions have occurred such as hive initialization, key flushing, key saving, key restoring and whatnot.

`CmCheckRegistry` is an essential routine that provides the bulk of registry healing and a piece of puzzle among others to achieve that. This routine works dependently based upon a bit flag that is passed to **Flags** parameter. The following flags are:
- **CM_CHECK_REGISTRY_DONT_PURGE_VOLATILES** -- Don't do any volatile purges
- **CM_CHECK_REGISTRY_PURGE_VOLATILES** -- Purge out volatile information data from a registry hive, on-demand
- **CM_CHECK_REGISTRY_BOOTLOADER_PURGE_VOLATILES** -- Purge out bootloader related volatile data, whatever that is
- **CM_CHECK_REGISTRY_VALIDATE_HIVE** -- Perform hive validation checks and thorough analyzing of hive's bins and cells, `HvValidateHive` will be triggered in this case
- **CM_CHECK_REGISTRY_FIX_HIVE** -- Fix a broken registry hive on-demand. This flag is typically used by an offline registry repair tool such as **ReactOS Check Registry Utility**, where one could repair a hive without the need to boot the ReactOS Usetup to repair it.

`CmCheckRegistry` performs the following actions in this strict order:
- Validate the hive if requested by submitting the **CM_CHECK_REGISTRY_VALIDATE_HIVE** flag, from there each bin internal structure header is validated for correctness (non-corrupt), size and if it's mapped, that cells sizes make sense and cells pointers, etc.
- Validate and analyze the security descriptor of each hive block, in accordance with the requirements imposed by the Security subsystem (aka Se) with `RtlValidSecurityDescriptor` routine. A security descriptor that went to shit will be reported as such and security information will be defaulted accordingly. From there, security caching comes into play.
- Perform deep recursive registry checking, the key is also validated. Lexicographical order is also validated. 

KVM https://reactos.org/testman/compare.php?ids=85306,85313

## TODO
For the last point, apparently there's quite some stuff in our Configuration Manager of the kernel that are a stub or barely implemented at all in the kernel, namely map view of hives and security caching support and stuff. All of this further work has to be shipped separately in different PRs.

- [x] Implement `HvValidateHive`
- [x] Implement `HvValidateBin`
- ~~[ ] Implement security descriptor validation~~ (to be implemented on a separate PR in the future)
- [x] Implement deep checking of the registry
- [x] Implement key validation, value lists, lexicographical order, etc
- [x] Implement registry self healing (`CmSelfHeal` and `CmpBootType` switch indicators)
- [x] Implement bootloader hive healing & recovery
- [x] Implement log transaction & other necessary code
- [x] Fix registry lazy flushing (merged separately in #4955)
- [x] Do more investigations for missing implementations of stuff in Configuration Manager that prevent the further progress of the implementation of registry checking itself
- [x] Run testbot / perform further tests
- [x] Fix / work around freeldr size issue on x64

**JIRA Issues:** [CORE-9195](https://jira.reactos.org/browse/CORE-9195), [CORE-6762](https://jira.reactos.org/browse/CORE-6762), [CORE-18303](https://jira.reactos.org/browse/CORE-18303), [CORE-19207](https://jira.reactos.org/browse/CORE-19207)


HBelusca:
It could be of interest to move the cmcheck file (+ the associated defines) in the cmlib instead. The reasons:
- it's going to be used in the bootloader (for fixing the SYSTEM hive);
- it's of course also used in the kernel;
- it would help in testing the code in user-mode (and debugging it with e.g. VS or whatever);
- it would help in creating an offline registry repair tool, similar to Windows' chkreg.exe : https://www.microsoft.com/en-us/download/details.aspx?id=20068

HBelusca:
Build has been fixed. I let you re-order + squash the commits as needed:
- the ` [CMLIB][NTOS:CM] Deduplicate other common definitions between CMLIB and the NTOS CM` commit must be moved at the very first position, and never be squashed (ideally it should be committed in master right now);
- the ` ** TO BE SQUASHED **` commit should be squashed with your current last commit `Implement registry check validation".

sdk/lib/cmlib/cmcheck.c
```diff
@@ -0,0 +1,62 @@
...
+#endif
+
+    /*
+     * Obtain the hive and check if the caller wants
+     * that the hive has to be validated.
+     */
+    Hive = GET_HHIVE(RegistryHive);
+
+    UNREFERENCED_PARAMETER(CmStatusCode);
+    UNREFERENCED_PARAMETER(Hive);
```
HBelusca, line 57: These two lines have been added to fix the build for now. When you implement the "TODO" and therefore, use those variables, these `UNREFERENCED_PARAMETER` should go away.

sdk/lib/cmlib/cmcheck.c
```diff
@@ -0,0 +1,62 @@
...
+#include "cmlib.h"
+#define NDEBUG
+#include <debug.h>
+
+/* DEFINES  ******************************************************************/
+
+#define GET_HHIVE(CmHive)   (&((CmHive)->Hive))
+
+#if !defined(CMLIB_HOST) && !defined(_BLDR_)
+extern PCMHIVE CmiVolatileHive;
```
HBelusca, line 19: Since that variable is used only once only there, it's cleaner to put it there, instead of polluting the global cmlib header with it.

sdk/lib/cmlib/cmcheck.c
```diff
@@ -0,0 +1,236 @@
...
+
+    return CM_CHECK_REGISTRY_GOOD;
+}
+
+CM_CHECK_REGISTRY_STATUS
+NTAPI
+HvValidateHive(
+    _In_ PHHIVE Hive)
+{
+    ULONG CmStatusCode;
```
HBelusca, line 142:
```suggestion
    CM_CHECK_REGISTRY_STATUS CmStatusCode;
```

sdk/lib/cmlib/cmlib.h
```diff
@@ -500,6 +532,34 @@ CmPrepareHive(
 
 /* NT-style Public Cm functions */
 
+//
+// Check Registry Routines
+//
+BOOLEAN
+NTAPI
+CmpValidateSDsOfHive(
```
HBelusca, line 540: If this function is going to be used only within cmcheck.c, then it doesn't need to be present in this global header.

sdk/lib/cmlib/cmcheck.c
```diff
@@ -0,0 +1,236 @@
...
+    ASSERT(Bin);
+    ASSERT(Hive);
+    ASSERT(StorageLength != 0);
+
+    /*
+     * The block offset is the current storage space
+     * times the block size.
+     */
+    FileOffset = Hive->Storage[StorageIndex].Length *
+                 HBLOCK_SIZE;
```
HBelusca, line 61:
```suggestion
    FileOffset = Hive->Storage[StorageIndex].Length * HBLOCK_SIZE;
```

sdk/lib/cmlib/cmcheck.c
```diff
@@ -0,0 +1,236 @@
...
+    PHCELL Cell, Basket;
+
+    /* Sanity checks */
+    ASSERT(Bin);
+    ASSERT(Hive);
+    ASSERT(StorageLength != 0);
+
+    /*
+     * The block offset is the current storage space
+     * times the block size.
```
HBelusca, line 58: I guess we could have understood that by just reading the formula 🤣

sdk/lib/cmlib/cmcheck.c
```diff
@@ -0,0 +1,341 @@
...
+     * Begin looping and validate each subkey now.
+     * From there we also check the order consistency
+     * (aka lexicographical order).
+     */
+    ParentCell = RootCell;
+    Sibling = HCELL_NIL;
+    for (SubKeyIndex = 0; SubKeyIndex < RootNode->SubKeyCounts[Stable]; SubKeyIndex++)
+    {
+        /* Obtain the subkey at the index we're in */
+        SubKeyCell = CmpFindSubKeyByNumber(Hive, RootNode, SubKeyIndex);
```
HBelusca, line 89: Not sure you can use this function. It expects that the referenced data is valid (and would assert/crash if not); however here, we are validating and repairing possibly corrupted data.

sdk/lib/cmlib/cmcheck.c
```diff
@@ -0,0 +1,341 @@
...
+    }
+
+    /*
+     * Begin looping and validate each subkey now.
+     * From there we also check the order consistency
+     * (aka lexicographical order).
+     */
+    ParentCell = RootCell;
+    Sibling = HCELL_NIL;
+    for (SubKeyIndex = 0; SubKeyIndex < RootNode->SubKeyCounts[Stable]; SubKeyIndex++)
```
HBelusca, line 86: Are you sure the SubKeyCounts contents is valid? etc. (recall that we repair the hive, so its yet-non-validated data has to be considered as potentially invalid/corrupted).

sdk/lib/cmlib/cmcheck.c
```diff
@@ -0,0 +1,341 @@
...
+
+    FileOffset = Hive->Storage[StorageIndex].Length * HBLOCK_SIZE;
+
+    /* Ensure that this bin we got has valid header */
+    if (Bin->Signature != HV_HBIN_SIGNATURE ||
+        Bin->Size > StorageLength ||
+        Bin->FileOffset != FileOffset)
+    {
+        DPRINT1("HvValidateBin(): The bin's header is corrupt!\n");
+        return CM_CHECK_REGISTRY_BIN_HEADER_CORRUPT;
```
HBelusca, line 159: When is the repair done?

binarymaster:
```
D:\a\reactos\reactos\src\boot\freeldr\freeldr\ntldr\registry.c(275): error C4312: 'type cast': conversion from 'ULONG' to 'PULONG' of greater size
```
MSVC x64 not happy.

boot/freeldr/freeldr/ntldr/wlregistry.c
```diff
@@ -141,7 +141,7 @@ WinLdrInitSystemHive(
     }
 
     /* Import what was loaded */
-    Success = RegImportBinaryHive(VaToPa(LoaderBlock->RegistryBase), LoaderBlock->RegistryLength);
+    Success = RegImportBinaryHive(LoaderBlock);
```
HBelusca, line 144: This was originally done such that registry.c doesn't depend on the rest of definitions needed for the whole LOADER_PARAMETER_BLOCK structure. So please revert.
Also, what happens to the `RegistryLength` value now?

boot/freeldr/freeldr/ntldr/registry.c
```diff
@@ -89,11 +111,362 @@ RegImportBinaryHive(
...
+ *
+ * @return
+ * Returns TRUE if data copying has succeeded without
+ * problems, FALSE otherwise.
+ */
+BOOLEAN
+RegCopyDataFromSystemLog(
+    _In_ PHBASE_BLOCK LogBaseBlock,
+    _In_ ULONG HiveLength,
+    _In_ PUCHAR SystemHive,
```
HBelusca, line 249: Please use PVOIDs.

boot/freeldr/freeldr/ntldr/registry.c
```diff
@@ -89,11 +111,362 @@ RegImportBinaryHive(
...
+     * we can read the dirty vector
+     * and write up data to the hive after
+     * the header.
+     */
+    HeaderSize = ROUND_UP(HV_LOG_HEADER_SIZE, LogBaseBlock->Cluster * HSECTOR_SIZE);
+    Offset = LogBaseBlock->Cluster * HSECTOR_SIZE;
+    Offset = ROUND_UP(Offset, HeaderSize);
+
+    /* Now check the dirty vector signature */
+    DirtyVector = *((PULONG)(SystemLog + Offset));
```
HBelusca, line 273: `(ULONG_PTR)SystemLog + Offset` etc.

GeoB99:
As of right now, most of TODOs in this PR are finally done and it should be ready for testing and whatnot. Here's a list of proposed changes:

- Don't queue the registry lazy flusher during a registry unlocking operation. This fixes a bug with the lazy flusher not working properly (actually, not flushing the registry at all). It didn't work for almost a decade.
- Implemented the registry validation function, aka `CmCheckRegistry`. The heart of registry checks algorithms, it also performs registry repairing if needed.
- Implemented registry self-healing routines, mostly needed for `CmCheckRegistry`.
- Move the `CmCheckRegistry` function to our dedicated ReactOS Configuration Manager Library (CMLIB). This allows the function to be exported to user mode as well, such as developing an offline registry repairing tool and whatnot.
- Removed the `CmPrepareHive` hack which purged volatile data in an improper way. Instead use `CmCheckRegistry` which purges volatiles more correctly.
- Slight refactoring of `HvInitialize` (thanks to the point above)
- Perform registry validation checks against the SYSTEM hive on FreeLdr. Perform repairing procedures if need be. Also the necessary algorithms to recover the SYSTEM hive using its respective LOG in case it's too corrupt is also implemented, so that FreeLdr can do a "boot recover". 
- Enable & implement log transaction writes (the initial work is an inspiration from PR #3932, I'd like to honor @mrmks04 for being an awesome guy). I have made some code tweaks as well as continuing the work.
- Document the code and other parts of CMLIB.
- Etc.

Now ReactOS provides some resilience and basic protection against the most usual registry corruptions. Namely, the infamous CTRL+ALT+DEL dialog box during boot-up should no longer occur on first time clean installation of the system **(1)** thanks to the lazy flusher working again. 

With the help of hive LOGs and system recovery algorithms, ReactOS should be able to recover a broken hive by obtaining data from a LOG. If the LOG is corrupt too, the system will perform resuscitation at the desperation attempt to heal the system, at the cost of data loss. If the SYSTEM hive is too corrupt, FreeLdr will recover it from a LOG.

By the end of this month as well as the whole November will be a test-fest time. So I'll be asking if you guys can give it a shot. This should help me understanding how effective is registry resiliency and catching bugs.

## IMPORTANT NOTES (READ THIS MY FELLOW DUMMY HUMAN)

1* --  While the CTRL+ALT+DEL dialog box should no longer appear, it's still possible to get it if one happens to force reboot the system too early when the desktop taskbar isn't even displayed. It's because this phenomena isn't a registry corruption per se but rather the local user profile wasn't even prevent physically on the registry hive file.

Technically speaking, the registry is loaded in memory (aka your RAM) from registry hives and the local user profile is created on first clean installation. The user profile, since still residing in memory, it hasn't been flushed to disk yet and first lazy flushing kicks in after 5 seconds. This is the normal behavior of the lazy flusher. Once 5 seconds have elapsed (the moment ROS is on desktop) it will flush NTUSER.DAT hive to disk, preserving the local profile. On Windows they implement local profile restoration/recreation on either MSGINA or USERENV, whichever of these two components.

Furthermore, **REACTOS IS STILL PRONE TO REGISTRY FS CORRUPTIONS**! That is, if a registry hive is corrupt down to filesystem level, the registry self-healing mechanisms won't be able to recover it. This only can be helped with a FS that has journaling and/or TxR like NTFS and BTRFS (unfortunately the BTRFS boot on ROS has currently regressed). FASTFAT with CHKDSK may probably help fixing a corrupted file down to FS level.

binarymaster:
> I have made some code tweaks as well as continuing the work.

Does that mean #3932 will be superseded by this PR?

GeoB99:
> > I have made some code tweaks as well as continuing the work.
> 
> Does that mean #3932 will be superseded by this PR?

Yes.

binarymaster:
> Yes.

No way to rebase anything on top of it, so it could be merged as well? It would be an overkill probably though...

sdk/lib/cmlib/hivecell.c
```diff
@@ -5,10 +5,22 @@
  *            Copyright 2001 - 2005 Eric Kohl
  */
 
+/* INCLUDES *****************************************************************/
+
 #include "cmlib.h"
```
JoachimHenze, line 10: Which purpose does this added comment have? Is the additional "s" after "include" followed by 66 stars really adding any value? I mean: Look what the first word at line 10 is telling you what is comming next... Surprise: The includes.

boot/freeldr/freeldr/ntldr/registry.c
```diff
@@ -89,11 +111,363 @@ RegImportBinaryHive(
...
+    /* Allocate and initialize the hive */
+    CmSystemHive = FrLdrTempAlloc(sizeof(CMHIVE), 'eviH');
+    Success = RegInitializeSystemHive(CmSystemHive, ChunkBase);
+    if (!Success)
+    {
+        /*
+         * The SYSTEM hive is corrupt. Recover
+         * it with the SYSTEM log that we have
+         * at disposal. Free the system hive buffer
+         * first.
```
JoachimHenze, line 435: A general note: I won't highlight this everywhere, but when scrolling through your code I do often think that your code would be better readable if you would avoid a few line-breaks here and there within your comments.
Don't get me wrong: Commenting is good, but being too excessive and consuming too many lines with it will make less **actual code** fit to the screen, which makes it harder for readers to grasp the wider context of things.
Please take that thought with you for future PRs.

ntoskrnl/config/cmlazy.c
```diff
@@ -9,7 +9,7 @@
 /* INCLUDES *******************************************************************/
 
 #include "ntoskrnl.h"
-#define NDEBUG
+//#define NDEBUG
```
JoachimHenze, line 12: intended? Or leftover?

ntoskrnl/config/cmlazy.c
```diff
@@ -100,7 +100,7 @@ CmpDoFlushNextHive(_In_  BOOLEAN ForceFlush,
             /* Use another lazy flusher for this hive */
             ASSERT(CmHive->FlushCount == CmpLazyFlushCount);
             *DirtyCount += CmHive->Hive.DirtyCount;
-            DPRINT("CmHive %wZ already uptodate.\n", &CmHive->FileFullPath);
+            DPRINT("CmHive %wZ already up to date.\n", &CmHive->FileFullPath);
```
JoachimHenze, line 103: superfluous change. Addressing this and my comment above would allow you to not touch that file at all!

ntoskrnl/config/cmsysini.c
```diff
@@ -1329,8 +1336,19 @@ CmpLoadHiveThread(IN PVOID StartContext)
...
             //}
 
-            /* Another thing we don't support is NTLDR-recovery */
-            if (CmHive->Hive.BaseBlock->BootRecover) ASSERT(FALSE);
+            /*
+             * FreeLdr has recovered the hive. We have
+             * to flush every dirty data that's been
+             * written to the hive by FreeLdr to
+             * the backing storage.
+             */
```
JoachimHenze, line 1344: you are wasting many lines with your commenting style of "beginning and ending each comment on its own line".

sdk/lib/cmlib/cmcheck.c
```diff
@@ -0,0 +1,1800 @@
...
+    {
+        ChildString.Buffer = &(ChildNode->Name[0]);
+        ChildString.Length = ChildNode->NameLength;
+        ChildString.MaximumLength = ChildNode->NameLength;
+        Result = CmpCompareCompressedName(&ChildString,
+                                          SiblingNode->Name,
+                                          SiblingNode->NameLength);
+        if (Result <= 0)
+        {
+            DPRINT1("CmpValidateLexicalOrder(): The compressed sibling node name is lesser or equal to that of the child!\n");
```
JoachimHenze, line 154: By grepping your code for !\n and counting the results you will see how many bytes you are wasting. Also long lines make it more likely to generate line-breaks in the logoutput due to limited window size, making the log harder to read. Each logged char consumes time and bytes. I would recommend stripping all your ! at the end of log lines therefore.

sdk/lib/cmlib/cmcheck.c
```diff
@@ -0,0 +1,1800 @@
...
+             * is still here. We cannot trust it so ultimately the
+             * value will be killed.
+             *
+             * CASE DataSize > 0:
+             *
+             * - This registry hive still has some data as the length
+             * tells us but we want to be sure this kind of data is
+             * not "cancerous", aka, is the cell allocated?
+             *
+             */
```
JoachimHenze, line 436: One superfluous line even in your commenting style. At least 3 too much for my style.

sdk/lib/cmlib/cmcheck.c
```diff
@@ -0,0 +1,1800 @@
...
+                 * damage done to the root is catastrophically
+                 * fatal.
+                 */
+                if (CurrentCell == RootCell)
+                {
+                    /*
+                     * IT IS THE ROOT!!! There's no hope for
+                     * this registry hive, it will go straight
+                     * to hell.
+                     */
```
JoachimHenze, line 1323: Nobody would miss that entire 5lines comment block. It doesn't transport any meaning that wouldn't be obvious already by the surrounding debug-prints and all the surrounding code.

sdk/lib/cmlib/cmcheck.c
```diff
@@ -0,0 +1,1800 @@
...
+                /* Assign the prior sibling for upcoming iteration */
+                WorkState[StackDepth - CMP_PRIOR_STACK].Sibling = CurrentCell;
+            }
+        }
+
+        /* Obtain a node for this key */
+        KeyNode = (PCM_KEY_NODE)HvGetCell(Hive, CurrentCell);
+        if (!KeyNode)
+        {
+            /* We couldn't be able to get the node, bail out */
```
JoachimHenze, line 1406: superfluous comment. Try to comment less! Focus on the "why" and not the "what". Especially when the "what" is obvious.

sdk/lib/cmlib/cmheal.c
```diff
@@ -0,0 +1,913 @@
+/*
+ * PROJECT:     ReactOS Kernel
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Configuration Manager Library - Registry Self-Heal Routines
+ * COPYRIGHT:   Copyright 2022 George Bișoc <george.bisoc@reactos.org>
+ */
+
+/* INCLUDES ******************************************************************/
```
JoachimHenze, line 8: <sarcasm> Thank god this received a comment, people might have searched for the includes at the end of the file otherwise! </sarcasm> I don't want to troll, I want to open your eyes what others might be thinking when reading such verbosely commented code.

sdk/lib/cmlib/cmheal.c
```diff
@@ -0,0 +1,913 @@
...
+{
+    PCM_KEY_INDEX Leaf;
+    PCM_KEY_FAST_INDEX FastIndex;
+    HCELL_INDEX LeafCell;
+    ULONG RootCountIndex;
+    ULONG LeafCountIndex;
+
+    PAGED_CODE();
+
+    /* Sanity check */
```
JoachimHenze, line 184: Have seen that exact kind of comment several times. Superfluous. An assert at a non-empty function: What else than a sanity check should that be?

sdk/lib/cmlib/cmheal.c
```diff
@@ -0,0 +1,913 @@
...
+    _In_ HCELL_INDEX TargetKey,
+    _In_ HCELL_INDEX ParentKey)
+{
+    PCM_KEY_INDEX KeyIndex;
+    PCM_KEY_NODE KeyNode;
+    BOOLEAN ParentRepaired;
+
+    PAGED_CODE();
+
+    DPRINT1("CmpRepairParentKey(): HIVE DAMAGED!!! Attempting repair...\n");
```
JoachimHenze, line 439: Why 3 (and not 7) exclamation marks?

sdk/lib/cmlib/cmheal.c
```diff
@@ -0,0 +1,913 @@
...
+BOOLEAN
+CMAPI
+CmpRepairKeyNodeSignature(
+    _Inout_ PHHIVE Hive,
+    _In_ HCELL_INDEX CurrentCell,
+    _Inout_ PCELL_DATA CellData)
+{
+    PAGED_CODE();
+
+    DPRINT1("CmpRepairKeyNodeSignature(): HIVE DAMAGED!!! Attempting repair...\n");
```
JoachimHenze, line 607: Same here: global text replacement !!! -> ! recommended.

sdk/lib/cmlib/cmlib.h
```diff
@@ -202,13 +205,75 @@
...
+#define CM_CHECK_REGISTRY_CORRUPT_SUBKEYS_INDEX        7010
+#define CM_CHECK_REGISTRY_BAD_SUBKEY_COUNT             7015
+#define CM_CHECK_REGISTRY_KEY_INDEX_CELL_UNALLOCATED   7020
+#define CM_CHECK_REGISTRY_CORRUPT_LEAF_ON_ROOT         7025
+#define CM_CHECK_REGISTRY_CORRUPT_LEAF_SIGNATURE       7030
+#define CM_CHECK_REGISTRY_CORRUPT_KEY_INDEX_SIGNATURE  7035
+
+//
+// Check Registry success macro
+//
```
JoachimHenze, line 274: You do use CPP style comments here, although that is C code otherwise. I do **not** criticize that.
But it makes me wonder why you don't use the CPP style comments everywhere then? They would allow many of your comments to be written in a more compact manner.

sdk/lib/cmlib/hiveinit.c
```diff
@@ -332,20 +437,34 @@ HvpInitializeMemoryHive(
...
- * @name HvpInitializeFlatHive
+ * @brief
+ * Initializes a hive descriptor for an already loaded hive
+ * that is stored in memory. This descriptor serves to denote
+ * such hive as being "flat", that is, the data and properties
+ * can be only read and not written into.
+ *
+ * @param[in] Hive
+ * A pointer to a registry hive descriptor where
+ * the internal structures field are to be initialized
```
JoachimHenze, line 448: field -> field**s**

sdk/lib/cmlib/hiveinit.c
```diff
@@ -357,98 +476,651 @@ HvpInitializeFlatHive(
...
     Result = Hive->FileRead(Hive,
                             HFILE_TYPE_PRIMARY,
-                            &Offset,
+                            &FileOffset,
                             BaseBlock,
                             Hive->Cluster * HSECTOR_SIZE);
+    if (!Result)
+    {
+        /*
+         * Don't assume the hive is ultiamtely destroyed
```
JoachimHenze, line 564: ultiamtely -> ultimately

sdk/lib/cmlib/hiveinit.c
```diff
@@ -357,98 +476,651 @@ HvpInitializeFlatHive(
...
+                             HFILE_TYPE_LOG,
+                             &FileOffset,
+                             DirtyVector,
+                             HSECTOR_SIZE);
+    if (!Success)
+    {
+        /*
+         * We failed? See if self-heal mode is
+         * enabled on this system.
+         */
```
JoachimHenze, line 899: again a 4 line-comment-block that does not add **any** information at all. Strip it, help your readers and their mousewheels.

ntoskrnl/config/cmlazy.c
GeoB99: Intended, as it enables verbose debugging of the lazy flusher since it's now in action after a couple of years not doing anything so it's best to examine the behavior of the lazy flusher as ROS matures with time, until it'll be disabled back.

ntoskrnl/config/cmsysini.c
GeoB99: Have you bothered to check other parts of the kernel where multi-line comments (e.g. https://github.com/reactos/reactos/blob/master/ntoskrnl/ob/obname.c#L234 and etc) before calling it "your commenting style" before reviewing?

sdk/lib/cmlib/hivecell.c
GeoB99: I have the feeling that you clearly have no idea what you're actually reviewing which, granted some reviews may make sense but most of them are just unnecessary noise. Here's a following list of links:

https://github.com/reactos/reactos/blob/master/ntoskrnl/ob/obsecure.c
https://github.com/reactos/reactos/blob/master/boot/freeldr/freeldr/freeldr.c
https://github.com/reactos/reactos/blob/master/base/system/winlogon/winlogon.c
https://github.com/reactos/reactos/blob/master/base/services/shsvcs/shsvcs.c
https://github.com/reactos/reactos/blob/master/drivers/filesystems/ntfs/btree.c
(and so on)

And we have been following such style for ages and nobody has questioned it. I am going to say it for the last time, think twice before reviewing.

JoachimHenze: https://reactos.org/wiki/Coding_Style dosn't mention or enforce anything like that. Other places having chosen questionable practices is not requiring you or should encourage you to do the same.
Nevertheless most other code that you pointed to doesn't suffer as much from commentitis as it is the case for the code that is under review here.
People nitpicking mostly about comments is a logical consequence when >50% of the proposed code-changes do consist of comments just.
But let's not fight, instead please take away something for the future from this discussion.

ntoskrnl/config/cmsysini.c
JoachimHenze: Same thing: Other places do not comment as verbosely as you do. I do not really care about the style of your commenting, but I am ciriticizing the sheer amount of comments in relation to the code in your PR. So I added some specific suggestions about how you could improve. You can chose other ways to reduce their amount, I don't really care how you decide to improve, I do just point to the fact that you comment too excessively.
You should ask yourself the question for each and every comment that you put into your code:
Does it really help others to better understand the code? Does it really add some value? Or is it it just saying the same thing that can be easily seen by anybody just looking at the code directly.

Writing too many comments is a bad practice in the same way as writing too much code is. For the same reason: both is error-prone and each line you write today will get viewed/reviewed/changed and later updated a thousand times by others. So what you add, should really offer some value.
It is definitely no good idea to comment each and every line of code you write!

tkreuzer:
This still breaks freeldr on x64 due to size issues. My suggestion to fix this:
1. Move newly added functions in "[SDK][CMLIB] Implement log transaction writes & Resuscitation" to a separate file, comment out their call for amd64 for now
2. Do the same thing for "[FREELDR] Implement SYSTEM hive recovery at bootloader level & use CmCheckRegistry for registry validation"
This should be a temporary hack and later superceded with a dll loaded by freeldr, preferably containing all ntldr code. This should fix size issues once and for all.
Also I would appreciate if you could move all formatting changes to existing code into a separate commit. That would make reviewing easier.
Another thing would be to split off the code up to "[NTOS:CM] Don't lazy flush the registry during unlocking operation" into a separate PR, which IIUC should fix the main issue we have and should be good to merge straight away.

SergeGautherie:
'help wanted': What kind of help are you looking for?

GeoB99:
It's been finally merged. I want to thank @mrmks04, @HBelusca, testers and everybody who helped in and made this possible!

JoachimHenze:
Attention: This PR introduced regression:
https://jira.reactos.org/browse/CORE-19337

PR #4577, opened 2022-07-15
[CRT] Add round and roundf

tkreuzer:
## Purpose

Add round and roundf and add it to msvcrtex to make clang v14 happy


sdk/lib/crt/msvcrtex.cmake
```diff
@@ -13,6 +13,7 @@ list(APPEND MSVCRTEX_SOURCE
 if(CMAKE_C_COMPILER_ID STREQUAL "Clang")
     # Clang performs some optimizations requiring those funtions
     list(APPEND MSVCRTEX_SOURCE
+        math/round.c
```
HBelusca, line 16: round**f**.c not added however?

learn-more: it is now :)

PR #4600, opened 2022-08-08
Move KDBG kernel debugger from ntoskrnl to its own kdcom.dll (CORE-10749)

tkreuzer:
## Purpose

Move KDBG kernel debugger from ntoskrnl to its own kdcom like kdrosdbg.dll 
The goal is to turn KDBG completely into a kdcom like extension dll, which would make compilation of _WINKD_ vs KDBG obsolete and turn KDBG into a KD64 client, which can be selected at boot-time.

JIRA issue: [CORE-10749](https://jira.reactos.org/browse/CORE-10749)

## Proposed changes
- Always use kdcom.dll functions, even when KDBG is enabled
- Extract all I/O (serial port I/O, screen output, keyboard input) from ntoskrnl to kdrosdbg.dll

## TODO
- [ ] Check if x64 still works
- [ ] Check if /DEBUGPORT=FILE still works (probably no...)
- [ ] Update KDBG commands (still in ntoskrnl/kdbg) to not directly use ntoskrnl internal functions/structures
- [ ] Move KDBG command handling from ntoskrnl to kdrosdbg
- [ ] Determine how to register and handle "plugin commands" (like for win32k and fastfat)
- [ ] Remove usage of _WINKD_ and KDBG defines in ntoskrnl


HBelusca:
Is this going to partially collide with the other [winkd_merge](https://github.com/reactos/reactos/commits/winkd_merge) branch?

PR #4612, opened 2022-08-16
[NTUSER][USER32] Define NtUserSetWindowLongPtr for non-Win64

katahiromz:
## Purpose
Improve code uniformity and backward compatibility of `NtUserSetWindowLongPtr`.

## Proposed changes

- Add `NtUserSetWindowLongPtr` macro on non-Win64.
- Use `NtUserSetWindowLongPtr` instead of `NtUserSetWindowLong`.

### DIFF
--- a/win32ss/include/ntuser.h
+++ b/win32ss/include/ntuser.h
@@ -3366,11 +3366,7 @@ NtUserSetWindowLongPtr(
     LONG_PTR NewValue,
     BOOL Ansi);
 #else
-static inline LONG_PTR NTAPI
-NtUserSetWindowLongPtr(HWND hWnd, DWORD Index, LONG_PTR NewValue, BOOL Ansi)
-{
-    return NtUserSetWindowLong(hWnd, Index, (LONG)NewValue, Ansi);
-}
+#define NtUserSetWindowLongPtr NtUserSetWindowLong
 #endif // _WIN64
 
 BOOL
--- a/win32ss/include/ntuser.h
+++ b/win32ss/include/ntuser.h
@@ -3366,8 +3366,11 @@ NtUserSetWindowLongPtr(
     LONG_PTR NewValue,
     BOOL Ansi);
 #else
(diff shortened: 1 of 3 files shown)


win32ss/include/ntuser.h
```diff
@@ -3365,6 +3365,9 @@ NtUserSetWindowLongPtr(
     DWORD Index,
     LONG_PTR NewValue,
     BOOL Ansi);
+#else
+#define NtUserSetWindowLongPtr(hWnd, Index, NewValue, Ansi) \
+    ((LONG_PTR)NtUserSetWindowLong((hWnd), (Index), (LONG)(NewValue), (Ansi)))
```
HBelusca: I think that in 32-bits it is sufficient to do
```suggestion
#define NtUserSetWindowLongPtr  NtUserSetWindowLong
```
(as it is done in user-mode for Get/SetWindowLongPtr), since the `LONG_PTR` type will automatically become `LONG`.

win32ss/user/user32/windows/class.c
```diff
@@ -1856,7 +1856,7 @@ SetWindowLongW(
   int nIndex,
   LONG dwNewLong)
 {
-    return NtUserSetWindowLong(hWnd, nIndex, dwNewLong, FALSE);
+    return (LONG)NtUserSetWindowLongPtr(hWnd, nIndex, dwNewLong, FALSE);
```
HBelusca, line 1859: I'm not sure this is useful to do those three changes, since it is OK to manipulate those WORD/LONG with the NtUserSetWindowLong (not Ptr) function even in x64.

win32ss/include/ntuser.h
katahiromz: OK.

win32ss/user/user32/windows/class.c
katahiromz: It's for code uniformity.

tkreuzer (approved):

PR #4614, opened 2022-08-19
[RTL][NTDLL_APITEST] Implement RtlRemovePrivileges

RatinCN:
Implement RtlRemovePrivileges

### DIFF
--- a/sdk/include/ndk/rtlfuncs.h
+++ b/sdk/include/ndk/rtlfuncs.h
@@ -1568,6 +1568,19 @@ RtlReleasePrivilege(
     _In_ PVOID ReturnedState
 );
 
+#if (NTDDI_VERSION >= NTDDI_VISTA)
+
+NTSYSAPI
+NTSTATUS
+NTAPI
+RtlRemovePrivileges(
+    _In_ HANDLE TokenHandle,
+    _In_reads_opt_(PrivilegeCount) _When_(PrivilegeCount != 0, _Notnull_) PULONG PrivilegesToKeep,
+    _In_ ULONG PrivilegeCount
+);
+
+#endif /* (NTDDI_VERSION >= NTDDI_VISTA) */
+
 _IRQL_requires_max_(APC_LEVEL)
 NTSYSAPI
 NTSTATUS
--- a/sdk/include/ndk/rtlfuncs.h
+++ b/sdk/include/ndk/rtlfuncs.h
@@ -1575,7 +1575,8 @@ NTSTATUS
 NTAPI
 RtlRemovePrivileges(
     _In_ HANDLE TokenHandle,
-    _In_reads_opt_(PrivilegeCount) _When_(PrivilegeCount != 0, _Notnull_) PULONG PrivilegesToKeep,
+    _In_reads_opt_(PrivilegeCount) _When_(PrivilegeCount != 0, _Notnull_)
+         PULONG PrivilegesToKeep,
     _In_ ULONG PrivilegeCount
 );
 
--- a/sdk/lib/rtl/priv.c
+++ b/sdk/lib/rtl/priv.c
@@ -488,6 +488,7 @@ RtlAdjustPrivilege(IN ULONG Privilege,
 }
 
 #if (NTDDI_VERSION >= NTDDI_VISTA)
+
 /**
  * @brief
  * Removes all privileges in the specified access token.
@@ -514,7 +515,7 @@ NTAPI
 RtlRemovePrivileges(
     _In_ HANDLE TokenHandle,
     _In_reads_opt_(PrivilegeCount) _When_(PrivilegeCount != 0, _Notnull_)
-        PULONG PrivilegesToKeep,
+         PULONG PrivilegesToKeep,
     _In_ ULONG PrivilegeCount)
 {
     NTSTATUS Status;
@@ -585,4 +586,5 @@ RtlRemovePrivileges(
 
     return Status;
 }
+
 #endif /* (NTDDI_VERSION >= NTDDI_VISTA) */
--- a/sdk/lib/rtl/priv.c
+++ b/sdk/lib/rtl/priv.c
@@ -486,3 +486,103 @@ RtlAdjustPrivilege(IN ULONG Privilege,
 
     return STATUS_SUCCESS;
 }
+
+#if (NTDDI_VERSION >= NTDDI_VISTA)
+/**
+ * @brief
+ * Removes all privileges in the specified access token.
+ *
+ * @param[in] TokenHandle
+ * A handle to the access token that contains the privileges to be removed.
+ *
+ * @param[in] PrivilegesToKeep
+ * A pointer to an array of privilege values (defined as SE_XXX_PRIVILEGE) that specify
+ * the privileges to keep in the token.
+ *
+ * @param[in] PrivilegeCount
+ * Specifies the number of entries in the PrivilegesToKeep array.
+ *
+ * @return
+ * Returns STATUS_SUCCESS if privileges removed successfully.
+ * STATUS_INVALID_PARAMETER is returned if input privilege value greater than
+ * SE_MAX_WELL_KNOWN_PRIVILEGE. STATUS_NOT_ALL_ASSIGNED is returned if The token does
+ * not have one or more of the privileges specified in the PrivilegesToKeep parameter,
+ * and no privileges were removed. A failure NTSTATUS code is returned otherwise.
+ */
+NTSTATUS
+NTAPI
+RtlRemovePrivileges(
+    _In_ HANDLE TokenHandle,
+    _In_reads_opt_(PrivilegeCount) _When_(PrivilegeCount != 0, _Notnull_)
+        PULONG PrivilegesToKeep,
+    _In_ ULONG PrivilegeCount)
+{
+    NTSTATUS Status;
+    UINT64 PrivilegesToKeepBitmap;
+    ULONG i, ReturnLength;
+    UCHAR Buffer[sizeof(TOKEN_PRIVILEGES) +
+                 sizeof(LUID_AND_ATTRIBUTES) * (SE_MAX_WELL_KNOWN_PRIVILEGE - SE_MIN_WELL_KNOWN_PRIVILEGE)];
+    PTOKEN_PRIVILEGES Privileges;
+
+    C_ASSERT(SE_MAX_WELL_KNOWN_PRIVILEGE < 64);
+
+    DPRINT("RtlRemovePrivileges(%p, %p, %u)\n", TokenHandle, PrivilegesToKeep, PrivilegeCount);
+
+    /* Save privileges that should be keep */
+    PrivilegesToKeepBitmap = 0;
+    if (PrivilegeCount)
+    {
+        for (i = 0; i < PrivilegeCount; i++)
+        {
+            if (PrivilegesToKeep[i] > SE_MAX_WELL_KNOWN_PRIVILEGE)
+            {
+                return STATUS_INVALID_PARAMETER;
+            }
+            PrivilegesToKeepBitmap |= (1ULL << PrivilegesToKeep[i]);
+        }
+    }
+
+    /* Get token privileges information */
+    Status = ZwQueryInformationToken(TokenHandle,
+                                     TokenPrivileges,
+                                     Buffer,
+                                     sizeof(Buffer),
+                                     &ReturnLength);
+    if (!NT_SUCCESS(Status))
+    {
+        return Status;
+    }
+
+    /* Remove all privileges that we don't need to keep */
+    Privileges = (PTOKEN_PRIVILEGES)Buffer;
+    for (i = 0; i < Privileges->PrivilegeCount; i++)
+    {
+        LARGE_INTEGER Privilege = *(LARGE_INTEGER*)&Privileges->Privileges[i].Luid;
+        ASSERT(Privilege.QuadPart <= SE_MAX_WELL_KNOWN_PRIVILEGE);
+        if (PrivilegesToKeepBitmap & (1ULL << Privilege.QuadPart))
+        {
+            PrivilegesToKeepBitmap &= ~(1ULL << Privilege.QuadPart);
+        }
+        else
+        {
+            Privileges->Privileges[i].Attributes = SE_PRIVILEGE_REMOVED;
+        }
+    }
+
+    if (PrivilegesToKeepBitmap)
+    {
+        Status = STATUS_NOT_ALL_ASSIGNED;
+    }
+    else
+    {
+        Status = ZwAdjustPrivilegesToken(TokenHandle,
+                                         FALSE,
+                                         (PTOKEN_PRIVILEGES)Buffer,
+                                         sizeof(Buffer),
+                                         NULL,
+                                         NULL);
+    }
+
+    return Status;
+}
+#endif /* (NTDDI_VERSION >= NTDDI_VISTA) */
--- a/dll/ntdll/def/ntdll.spec
+++ b/dll/ntdll/def/ntdll.spec
@@ -1111,7 +1111,7 @@
 @ stdcall -stub -version=0x600+ RtlReleaseSRWLockExclusive(ptr)
 @ stdcall -stub -version=0x600+ RtlReleaseSRWLockShared(ptr)
 @ stdcall RtlRemoteCall(ptr ptr ptr long ptr long long)
-@ stub -version=0x600+ RtlRemovePrivileges
+@ stdcall -version=0x600+ RtlRemovePrivileges(ptr ptr long)
 @ stdcall RtlRemoveVectoredContinueHandler(ptr)
 @ stdcall RtlRemoveVectoredExceptionHandler(ptr)
 @ stub -version=0x600+ RtlReportException
--- a/modules/rostests/apitests/ntdll/CMakeLists.txt
+++ b/modules/rostests/apitests/ntdll/CMakeLists.txt
@@ -85,6 +85,7 @@ list(APPEND SOURCE
     RtlpEnsureBufferSize.c
     RtlQueryTimeZoneInfo.c
     RtlReAllocateHeap.c
+    RtlRemovePrivileges.c
     RtlUnicodeStringToAnsiString.c
     RtlUnicodeStringToCountedOemString.c
     RtlUnicodeToOemN.c
--- a/modules/rostests/apitests/ntdll/RtlRemovePrivileges.c
+++ b/modules/rostests/apitests/ntdll/RtlRemovePrivileges.c
@@ -1,0 +1,111 @@
+/*
+ * PROJECT:         ReactOS api tests
+ * LICENSE:         See COPYING in the top level directory
+ * PURPOSE:         Test for RtlRemovePrivileges
+ * PROGRAMMER:      Ratin Gao <ratin@knsoft.org>
+ */
+
+#include "precomp.h"
+
+START_TEST(RtlRemovePrivileges)
+{
+#if (NTDDI_VERSION >= NTDDI_VISTA)
+    NTSTATUS Status;
+    HANDLE TokenHandle, TestTokenHandle;
+    ULONG ReturnLength;
+    UCHAR Buffer
+        [sizeof(TOKEN_PRIVILEGES) +
+         sizeof(LUID_AND_ATTRIBUTES) * (SE_MAX_WELL_KNOWN_PRIVILEGE - SE_MIN_WELL_KNOWN_PRIVILEGE)];
+    PTOKEN_PRIVILEGES Privileges;
+    ULONG PrivilegesToKeep[2];
+
+    /* Duplicate current process token to run this test */
+    Status = NtOpenProcessToken(GetCurrentProcess(), TOKEN_DUPLICATE, &TokenHandle);
+    if (!NT_SUCCESS(Status))
+    {
+        ok(0, "Failed to open current process token with TOKEN_DUPLICATE access (Status code %lx)!\n", Status);
+        return;
+    }
+
+    Status = NtDuplicateToken(TokenHandle, TOKEN_ALL_ACCESS, NULL, FALSE, TokenPrimary, &TestTokenHandle);
+    NtClose(TokenHandle);
+    if (!NT_SUCCESS(Status))
+    {
+        ok(0, "Failed to duplicate current process token (Status code %lx)!\n", Status);
+        return;
+    }
+
+    /* Retrieve token privileges, we need at least 3 privileges to run following tests */
+    Status = NtQueryInformationToken(TestTokenHandle, TokenPrivileges, Buffer, sizeof(Buffer), &ReturnLength);
+    if (!NT_SUCCESS(Status))
+    {
+        NtClose(TestTokenHandle);
+        ok(0, "Failed to retrieve token privileges (Status code %lx)!\n", Status);
+        return;
+    }
+    Privileges = (PTOKEN_PRIVILEGES)Buffer;
+    if (Privileges->PrivilegeCount < 3)
+    {
+        NtClose(TestTokenHandle);
+        ok(0, "No enough privileges to run the test (Number of privilege: %lu)!\n", Privileges->PrivilegeCount);
+        return;
+    }
+
+    /* Remove all privileges except 2nd and 3rd privileges, this should succeed */
+    PrivilegesToKeep[0] = Privileges->Privileges[1].Luid.LowPart;
+    PrivilegesToKeep[1] = Privileges->Privileges[2].Luid.LowPart;
+    Status = RtlRemovePrivileges(TestTokenHandle, PrivilegesToKeep, ARRAYSIZE(PrivilegesToKeep));
+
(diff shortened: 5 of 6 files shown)


sdk/lib/rtl/priv.c
```diff
@@ -486,3 +486,78 @@ RtlAdjustPrivilege(IN ULONG Privilege,
...
+    /* Get token privileges information */
+    ULONG ulReturnLength;
+    BYTE tpBuffer[sizeof(TOKEN_PRIVILEGES) +
+                  sizeof(LUID_AND_ATTRIBUTES) * (SE_MAX_WELL_KNOWN_PRIVILEGE - SE_MIN_WELL_KNOWN_PRIVILEGE)];
+    NTSTATUS Status = ZwQueryInformationToken(TokenHandle,
+                                              TokenPrivileges,
+                                              tpBuffer,
+                                              sizeof(tpBuffer),
+                                              &ulReturnLength);
+    if (NT_SUCCESS(Status))
```
HBelusca, line 526: Please do a return-early here instead, so that the (long) success code path doesn't get indented more.

sdk/lib/rtl/priv.c
```diff
@@ -486,3 +486,78 @@ RtlAdjustPrivilege(IN ULONG Privilege,
...
+    {
+
+        /* Remove all privileges that we don't need to keep */
+        PTOKEN_PRIVILEGES tp = (PTOKEN_PRIVILEGES)tpBuffer;
+        DWORD i;
+        for (i = 0; i < tp->PrivilegeCount; i++)
+        {
+            UINT64 ulPrivilege = ((UINT64)tp->Privileges[i].Luid.HighPart << 32) | tp->Privileges[i].Luid.LowPart;
```
HBelusca, line 534:
```suggestion
            LARGE_INTEGER Privilege = (LARGE_INTEGER)tp->Privileges[i].Luid;
```
Then the `ulPrivilege` below can be replaced by `Privilege.QuadPart`.

sdk/lib/rtl/priv.c
```diff
@@ -486,3 +486,78 @@ RtlAdjustPrivilege(IN ULONG Privilege,
...
+            if (*pulPrivilegesToKeep++ > SE_MAX_WELL_KNOWN_PRIVILEGE)
+            {
+                return STATUS_INVALID_PARAMETER;
+            }
+        }
+    }
+
+    /* Get token privileges information */
+    ULONG ulReturnLength;
+    BYTE tpBuffer[sizeof(TOKEN_PRIVILEGES) +
```
HBelusca, line 519: Don't use `BYTE`; use `UCHAR`

sdk/lib/rtl/priv.c
```diff
@@ -486,3 +486,78 @@ RtlAdjustPrivilege(IN ULONG Privilege,
 
     return STATUS_SUCCESS;
 }
+
+/*
```
HBelusca, line 490: Maybe add a doxygen-type documentation? (since this function is undocumented on the internet, it better be documented here in our code).

HBelusca (changes requested):

sdk/lib/rtl/priv.c
RatinCN: Of couse, I gonna fix :)

sdk/lib/rtl/priv.c
RatinCN: It's seems like `LUID` cannot be cast to `LARGE_INTEGER` directly due to they are structrues, shall I use:

`PLARGE_INTEGER Privilege = (PLARGE_INTEGER)&Privileges->Privileges[i].Luid;` instead OR:

`LARGE_INTEGER Privilege;`
`Privilege.HighPart = Privileges->Privileges[i].Luid.HighPart;`
`Privilege.LowPart = Privileges->Privileges[i].Luid.LowPart;`？

HBelusca: `PLARGE_INTEGER Privilege = (PLARGE_INTEGER)&Privileges->Privileges[i].Luid;` may be simpler indeed, and it's close to something I was testing (which was:
```c
LARGE_INTEGER Privilege = *(LARGE_INTEGER*)&Privileges->Privileges[i].Luid;
```
)

HBelusca:
Ah, and since this is also a new function, please write an apitest for it in order to document and test its functioning, following the existing examples of
https://git.reactos.org/?p=reactos.git;a=tree;f=modules/rostests/apitests/ntdll;h=1b71170096ae03425f27561db49de82d05391f4d;hb=HEAD

sdk/lib/rtl/priv.c
RatinCN: Okay, I get it

GeoB99: Honestly the whole code inside the loop looks fishy to me. Why do we have to toy around with privileges in this way, and especially with LARGE_INTEGERs and 64-bit unsigned integers?  This looks quite unnatural and not so human readable.

sdk/lib/rtl/priv.c
```diff
@@ -486,3 +486,78 @@ RtlAdjustPrivilege(IN ULONG Privilege,
...
+                                              TokenPrivileges,
+                                              tpBuffer,
+                                              sizeof(tpBuffer),
+                                              &ulReturnLength);
+    if (NT_SUCCESS(Status))
+    {
+
+        /* Remove all privileges that we don't need to keep */
+        PTOKEN_PRIVILEGES tp = (PTOKEN_PRIVILEGES)tpBuffer;
+        DWORD i;
```
GeoB99, line 531: For NT code we usually use `ULONG` for loop indices.

GeoB99 (changes requested):

sdk/lib/rtl/priv.c
RatinCN: This implement equals to Windows (as Win11 I currently use)

This is a simple and effective way to save all the privileges specified in `PrivilegesToKeep` input parameter, and then we determine each privilege should be keep or not by testing the corresponding bit. If we do that by enumerating `PrivilegesToKeep` again and again, it's not a good idea I think.

sdk/include/ndk/rtlfuncs.h
```diff
@@ -1568,6 +1568,15 @@ RtlReleasePrivilege(
...
 
+NTSYSAPI
+NTSTATUS
+NTAPI
+RtlRemovePrivileges(
+    IN HANDLE TokenHandle,
+    IN PULONG PrivilegesToKeep,
+    IN ULONG PrivilegeCount
+);
+
```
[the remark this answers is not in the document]
RatinCN: This RTL function exported from ntdll.dll in user-mode, I not found NTDDI_VERSION macro used like this in ntdll.dll, shall I do so?

sdk/lib/rtl/priv.c
```diff
@@ -486,3 +486,78 @@ RtlAdjustPrivilege(IN ULONG Privilege,
...
+ * @implemented
+ */
+NTSTATUS
+NTAPI
+RtlRemovePrivileges(IN HANDLE TokenHandle,
+                    IN PULONG PrivilegesToKeep,
+                    IN ULONG PrivilegeCount)
+{
+    DPRINT("RtlRemovePrivileges(%p, %p, %u)\n", TokenHandle, PrivilegesToKeep, PrivilegeCount);
+
```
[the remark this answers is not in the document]
RatinCN: This RTL function exported from ntdll.dll in user-mode, in this case, PAGED_CODE_RTL() is not necessary?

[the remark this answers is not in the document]
GeoB99: There can be NTDLL functions that can be used in kernel mode on-demand. Is this `RtlRemovePrivileges` strictly used for user mode system components or can it have its usage in kernel mode too? If the former then you might not need `PAGED_CODE_RTL()`, otherwise you have to use it.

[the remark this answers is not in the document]
RatinCN: Yes, `RtlRemovePrivileges` is exported in user-mode only (in Windows), not in kernel-mode, like `RtlAdjustPrivilege` but new for NT 6.0+

sdk/include/ndk/rtlfuncs.h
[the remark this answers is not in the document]
tkreuzer: What's the point of using NTDDI_VERSION? It will just make things more complicated. If you want to use it from anywhere else, you now have to either redefine NTDDI_VERSION or duplicate the prototype.

sdk/lib/rtl/priv.c
```diff
@@ -486,3 +486,78 @@ RtlAdjustPrivilege(IN ULONG Privilege,
...
+    /* Save privileges that should be keep */
+    UINT64 PrivilegesToKeepBitmap = 0;
+    if (PrivilegeCount)
+    {
+        ULONG ulCount;
+        PULONG pulPrivilegesToKeep = PrivilegesToKeep;
+        for (ulCount = 0; ulCount < PrivilegeCount; ulCount++)
+        {
+            PrivilegesToKeepBitmap |= (1ULL << *pulPrivilegesToKeep);
+            if (*pulPrivilegesToKeep++ > SE_MAX_WELL_KNOWN_PRIVILEGE)
```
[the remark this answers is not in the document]
tkreuzer: No, this completely changes the logic.

sdk/lib/rtl/priv.c
```diff
@@ -486,3 +486,78 @@ RtlAdjustPrivilege(IN ULONG Privilege,
...
+                    IN ULONG PrivilegeCount)
+{
+    DPRINT("RtlRemovePrivileges(%p, %p, %u)\n", TokenHandle, PrivilegesToKeep, PrivilegeCount);
+
+    /* Save privileges that should be keep */
+    UINT64 PrivilegesToKeepBitmap = 0;
+    if (PrivilegeCount)
+    {
+        ULONG ulCount;
+        PULONG pulPrivilegesToKeep = PrivilegesToKeep;
```
tkreuzer, line 506: If you used something like `PrivilegesToKeep[Ii]` you wouldn't need another pointer, you wouldn't need the `if (PrivilegeCount)` and the code would look less like straight out of IDA.

sdk/lib/rtl/priv.c
```diff
@@ -486,3 +486,78 @@ RtlAdjustPrivilege(IN ULONG Privilege,
...
+            }
+            else
+            {
+                tp->Privileges[i].Attributes = SE_PRIVILEGE_REMOVED;
+            }
+        }
+
+        if (PrivilegesToKeepBitmap)
+        {
+            Status = STATUS_NOT_ALL_ASSIGNED;
```
tkreuzer, line 547: `STATUS_NOT_ALL_ASSIGNED` is a success code. So it should probably still remove the privileges.

sdk/lib/rtl/priv.c
ThFabba: Yeah, which is why the whole LowPart/HighPart dance makes no sense. So, assuming the logic is correct and should stay like this, just use LowPart for the bit shift and skip this step if HighPart is set (and/or assert that HighPart is zero). Then we can forget about the "how to convert LUID to LARGE_INTEGER" discussion.

sdk/lib/rtl/priv.c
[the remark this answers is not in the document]
ThFabba: `*p++` and `*(p++)` are indeed the same.  Another reason why this should just use `PrivilegesToKeep[i]`

sdk/lib/rtl/priv.c
```diff
@@ -486,3 +486,78 @@ RtlAdjustPrivilege(IN ULONG Privilege,
...
+NTSTATUS
+NTAPI
+RtlRemovePrivileges(IN HANDLE TokenHandle,
+                    IN PULONG PrivilegesToKeep,
+                    IN ULONG PrivilegeCount)
+{
+    DPRINT("RtlRemovePrivileges(%p, %p, %u)\n", TokenHandle, PrivilegesToKeep, PrivilegeCount);
+
+    /* Save privileges that should be keep */
+    UINT64 PrivilegesToKeepBitmap = 0;
```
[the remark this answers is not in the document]
ThFabba: And no Hungarian notation please (`Count`, not `ulCount` etc)

sdk/lib/rtl/priv.c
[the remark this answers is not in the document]
RatinCN: Yes, PrivilegesToKeep[i] would be better

sdk/lib/rtl/priv.c
RatinCN: I intend to keep the same with Windows, so here `HighPart` is involved. But in fact, `HighPart` is always 0 in this case, privilege values won't greater than 63 yet.

sdk/lib/rtl/priv.c
RatinCN: Here keeps the same with Windows, in this code path, `RtlRemovePrivileges` returns `STATUS_NOT_ALL_ASSIGNED` but do nothing actually.

RatinCN:
Ok to go? @HBelusca @GeoB99 @tkreuzer @ThFabba

RatinCN:
Screenshot
ntdll.dll and ntdll_apitest.exe complied without NTDDK_VERSION (temporarily removed to run this test) and run on ReactOS. In formal build (NTDDK_VERSION lower than Vista), this function will not involved and api test will skipped.

sdk/include/ndk/rtlfuncs.h
```diff
@@ -1568,6 +1568,19 @@ RtlReleasePrivilege(
...
 );
 
+#if (NTDDI_VERSION >= NTDDI_VISTA)
+
+NTSYSAPI
+NTSTATUS
+NTAPI
+RtlRemovePrivileges(
+    _In_ HANDLE TokenHandle,
+    _In_reads_opt_(PrivilegeCount) _When_(PrivilegeCount != 0, _Notnull_) PULONG PrivilegesToKeep,
```
HBelusca, line 1578:
```suggestion
    _In_reads_opt_(PrivilegeCount) _When_(PrivilegeCount != 0, _Notnull_)
        PULONG PrivilegesToKeep,
```

sdk/lib/rtl/priv.c
```diff
@@ -486,3 +486,97 @@ RtlAdjustPrivilege(IN ULONG Privilege,
...
+ * STATUS_INVALID_PARAMETER is returned if input privilege value greater than
+ * SE_MAX_WELL_KNOWN_PRIVILEGE. STATUS_NOT_ALL_ASSIGNED is returned if The token does
+ * not have one or more of the privileges specified in the PrivilegesToKeep parameter,
+ * and no privileges were removed. A failure NTSTATUS code is returned otherwise.
+ */
+NTSTATUS
+NTAPI
+RtlRemovePrivileges(_In_ HANDLE TokenHandle,
+                    _In_reads_opt_(PrivilegeCount) _When_(PrivilegeCount != 0, _Notnull_) PULONG PrivilegesToKeep,
+                    _In_ ULONG PrivilegeCount)
```
HBelusca, line 516:
```suggestion
RtlRemovePrivileges(
    _In_ HANDLE TokenHandle,
    _In_reads_opt_(PrivilegeCount) _When_(PrivilegeCount != 0, _Notnull_)
        PULONG PrivilegesToKeep,
    _In_ ULONG PrivilegeCount)
```

sdk/lib/rtl/priv.c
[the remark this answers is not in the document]
HBelusca: @GeoB99 : Normally the RTL should be compiled twice: once for the kernel (making it use possible kernel-specific defines), and once for user-mode (for ntdll). We don't do that yet, and as a result we may end up (in non-optimized builds) with user-mode-only functions (like this one) being added in the kernel but being never used.

sdk/lib/rtl/priv.c
```diff
@@ -486,3 +486,97 @@ RtlAdjustPrivilege(IN ULONG Privilege,
...
+{
+    UINT64 PrivilegesToKeepBitmap;
```
HBelusca, line 518:
```suggestion
    NTSTATUS Status;
    UINT64 PrivilegesToKeepBitmap;
```

sdk/lib/rtl/priv.c
```diff
@@ -486,3 +486,97 @@ RtlAdjustPrivilege(IN ULONG Privilege,
...
+            }
+        }
+    }
+
+    /* Get token privileges information */
+    NTSTATUS Status = ZwQueryInformationToken(TokenHandle,
+                                              TokenPrivileges,
+                                              Buffer,
+                                              sizeof(Buffer),
+                                              &ReturnLength);
```
HBelusca, line 545:
```suggestion
    Status = ZwQueryInformationToken(TokenHandle,
                                     TokenPrivileges,
                                     Buffer,
                                     sizeof(Buffer),
                                     &ReturnLength);
```

sdk/lib/rtl/priv.c
```diff
@@ -486,3 +486,97 @@ RtlAdjustPrivilege(IN ULONG Privilege,
...
+    DPRINT("RtlRemovePrivileges(%p, %p, %u)\n", TokenHandle, PrivilegesToKeep, PrivilegeCount);
+
+    /* Save privileges that should be keep */
+    PrivilegesToKeepBitmap = 0;
+    if (PrivilegeCount)
+    {
+        for (i = 0; i < PrivilegeCount; i++)
+        {
+            PrivilegesToKeepBitmap |= (1ULL << PrivilegesToKeep[i]);
+            if (PrivilegesToKeep[i] > SE_MAX_WELL_KNOWN_PRIVILEGE)
```
HBelusca, line 533: This validity check should be done before adding the value in the bitmap
(this would also allow the compiler to understand that `PrivilegesToKeep[i]` cannot be greater than 63 and thus make the shift operation undefined).

sdk/lib/rtl/priv.c
```diff
@@ -486,3 +486,97 @@ RtlAdjustPrivilege(IN ULONG Privilege,
...
+    {
+        return Status;
+    }
+
+    /* Remove all privileges that we don't need to keep */
+    Privileges = (PTOKEN_PRIVILEGES)Buffer;
+    for (i = 0; i < Privileges->PrivilegeCount; i++)
+    {
+        LARGE_INTEGER Privilege = *(LARGE_INTEGER*)&Privileges->Privileges[i].Luid;
+        if (PrivilegesToKeepBitmap & (1ULL << Privilege.QuadPart))
```
HBelusca, line 556: Before these shifts: a `if (Privilege.QuadPart > SE_MAX_WELL_KNOWN_PRIVILEGE)` validation check similar to the one above seems to be missing.

sdk/lib/rtl/priv.c
```diff
@@ -486,3 +486,97 @@ RtlAdjustPrivilege(IN ULONG Privilege,
...
+RtlRemovePrivileges(_In_ HANDLE TokenHandle,
+                    _In_reads_opt_(PrivilegeCount) _When_(PrivilegeCount != 0, _Notnull_) PULONG PrivilegesToKeep,
+                    _In_ ULONG PrivilegeCount)
+{
+    UINT64 PrivilegesToKeepBitmap;
+    ULONG i, ReturnLength;
+    UCHAR Buffer[sizeof(TOKEN_PRIVILEGES) +
+                 sizeof(LUID_AND_ATTRIBUTES) * (SE_MAX_WELL_KNOWN_PRIVILEGE - SE_MIN_WELL_KNOWN_PRIVILEGE)];
+    PTOKEN_PRIVILEGES Privileges;
+
```
HBelusca, line 523: Optional: compile-time check of an assumption used in this function:
```suggestion

    C_ASSERT(SE_MAX_WELL_KNOWN_PRIVILEGE < 64);

```

HBelusca:
> ntdll.dll and ntdll_apitest.exe complied without NTDDK_VERSION (temporarily removed to run this test) and run on ReactOS. In formal build (NTDDK_VERSION lower than Vista), this function will not involved and api test will skipped.

Does this test pass OK on Windows 7+ ?

sdk/lib/rtl/priv.c
RatinCN: The first validation checks input parameter, and checks `SE_MAX_WELL_KNOWN_PRIVILEGE` only, doesn't check `SE_MIN_WELL_KNOWN_PRIVILEGE`.
Here the value returned by `ZwQueryInformationToken`, so the check in here is not necessary if `ZwQueryInformationToken` returns correct value which lower than `SE_MAX_WELL_KNOWN_PRIVILEGE`.
The logic above is the same with Windows.

HBelusca: Hm OK, so maybe just an `ASSERT(Privilege.QuadPart <= SE_MAX_WELL_KNOWN_PRIVILEGE);` would be enough.
Also,
>  The logic above is the same with Windows.

Perhaps but you know, if said logic was "buggy" and then it turned out that sometime from now they issue a fix for (e.g.) this function, while we could have found said fix by ourselves much before.... It would have then been better to "improve" the logic from windows instead of making it "the same" (but buggy).

RatinCN:
> > ntdll.dll and ntdll_apitest.exe complied without NTDDK_VERSION (temporarily removed to run this test) and run on ReactOS. In formal build (NTDDK_VERSION lower than Vista), this function will not involved and api test will skipped.
> 
> Does this test pass OK on Windows 7+ ?

Yes, this apitest also passed in my Win7 and Win11:
Screenshot2
Screenshot1

sdk/lib/rtl/priv.c
```diff
@@ -486,3 +486,103 @@ RtlAdjustPrivilege(IN ULONG Privilege,
...
+                                         FALSE,
+                                         (PTOKEN_PRIVILEGES)Buffer,
+                                         sizeof(Buffer),
+                                         NULL,
+                                         NULL);
+    }
+
+    return Status;
+}
+#endif /* (NTDDI_VERSION >= NTDDI_VISTA) */
```
HBelusca, line 588:
```suggestion

#endif /* (NTDDI_VERSION >= NTDDI_VISTA) */
```

sdk/lib/rtl/priv.c
```diff
@@ -486,3 +486,103 @@ RtlAdjustPrivilege(IN ULONG Privilege,
...
+ * SE_MAX_WELL_KNOWN_PRIVILEGE. STATUS_NOT_ALL_ASSIGNED is returned if The token does
+ * not have one or more of the privileges specified in the PrivilegesToKeep parameter,
+ * and no privileges were removed. A failure NTSTATUS code is returned otherwise.
+ */
+NTSTATUS
+NTAPI
+RtlRemovePrivileges(
+    _In_ HANDLE TokenHandle,
+    _In_reads_opt_(PrivilegeCount) _When_(PrivilegeCount != 0, _Notnull_)
+        PULONG PrivilegesToKeep,
```
HBelusca, line 517:
```suggestion
         PULONG PrivilegesToKeep,
```

HBelusca (approved):

sdk/lib/rtl/priv.c
RatinCN: OK, and add new line after `#if` too?

sdk/lib/rtl/priv.c
RatinCN: This one more space for align `PULONG` and `HANDLE`? I saw many functions alignment like this:
_IRQL_requires_max_(PASSIVE_LEVEL)
```
NTSYSAPI
NTSTATUS
NTAPI
ZwOpenProcessToken(
    _In_ HANDLE ProcessHandle,
    _In_ ACCESS_MASK DesiredAccess,
    _Out_ PHANDLE TokenHandle);
```

sdk/lib/rtl/priv.c
HBelusca: iirc it is already there.

sdk/lib/rtl/priv.c
HBelusca: ok let's not worry about it then.

sdk/lib/rtl/priv.c
RatinCN: Nope, but doxygen comments are followed:
```
#if (NTDDI_VERSION >= NTDDI_VISTA)
/**
 * @brief
 * Removes all privileges in the specified access token.
 ...
```

HBelusca:
Please ensure that
image
is enabled in the PRs so that if there are (for example) last-time minor formatting/typos suggestions to be applied to it, they can be applied by the person who will commit the PR into master, instead of waiting for PR author / making a separate branch and then manually closing the PR.

RatinCN:
> Please ensure that image is enabled in the PRs so that if there are (for example) last-time minor formatting/typos suggestions to be applied to it, they can be applied by the person who will commit the PR into master, instead of waiting for PR author / making a separate branch and then manually closing the PR.

Done

sdk/lib/rtl/priv.c
```diff
@@ -486,3 +486,103 @@ RtlAdjustPrivilege(IN ULONG Privilege,
 
     return STATUS_SUCCESS;
 }
+
+#if (NTDDI_VERSION >= NTDDI_VISTA)
```
HBelusca, line 490:
```suggestion
#if (NTDDI_VERSION >= NTDDI_VISTA)

```

HBelusca (approved):

PR #4618, opened 2022-08-21
[USETUP] Hardcode PARTITION_MAXSIZE

silvicky:
## Purpose

Fix partition max size on versions without pow.s, like amd64.

JIRA issue: [CORE-18310](https://jira.reactos.org/browse/CORE-18310)

## Proposed changes

Turn the PARTITION_MAXSIZE macro from an expression into a fixed sum, so it no longer requires unimplemented calculation.


base/setup/usetup/usetup.c
```diff
@@ -1874,7 +1874,8 @@ SelectPartitionPage(PINPUT_RECORD Ir)
 
 #define PARTITION_SIZE_INPUT_FIELD_LENGTH 9
 /* Restriction for MaxSize */
-#define PARTITION_MAXSIZE (pow(10, (PARTITION_SIZE_INPUT_FIELD_LENGTH - 1)) - 1)
+/* #define PARTITION_MAXSIZE(pow(10, (PARTITION_SIZE_INPUT_FIELD_LENGTH - 1)) - 1)*/
+#define PARTITION_MAXSIZE 99999999
```
GeoB99, line 1878: If partition maximum size calculation in AMD64 build of ROS doesn't work then it is probably because `pow` isn't actually implemented for the AMD64 architecture? So the legit fix would be to actually implement `pow` rather than rely on hacks like this?

GeoB99 (dismissed):

base/setup/usetup/usetup.c
silvicky: Sure, but calculating such a simple constant expression might not be necessary, and this value has only one usage.

base/setup/usetup/usetup.c
```diff
@@ -1874,7 +1874,8 @@ SelectPartitionPage(PINPUT_RECORD Ir)
 
 #define PARTITION_SIZE_INPUT_FIELD_LENGTH 9
 /* Restriction for MaxSize */
-#define PARTITION_MAXSIZE (pow(10, (PARTITION_SIZE_INPUT_FIELD_LENGTH - 1)) - 1)
+/* #define PARTITION_MAXSIZE(pow(10, (PARTITION_SIZE_INPUT_FIELD_LENGTH - 1)) - 1)*/
```
HBelusca, line 1877:
```suggestion
// #define PARTITION_MAXSIZE (pow(10, (PARTITION_SIZE_INPUT_FIELD_LENGTH - 1)) - 1)
```

base/setup/usetup/usetup.c
HBelusca, line 1878:
```suggestion
#define PARTITION_MAXSIZE 99999999UL
```

GeoB99: My point is, this piece of code never bothered anyone and the real fix should be implementing that math function. Since your patch is a mere workaround it must be treated and explicitly explained in a comment (including the aforementioned CORE-ID ticket) as such.

silvicky: I added a comment.
Should I rebase?

HBelusca: You can just add a second commit. We'll merge everything when committing the PR.

silvicky:
we have pow now; abandon this

PR #4619, opened 2022-08-21
[BOOT][SDK:CMLIB] Compile a CMLIB for FreeLdr / NT bootloader with correct definitions.

HBelusca:
This also allows using the CMLIB with any reduced functionality
that could be required at boot-time.


GeoB99 (approved):
Thanks to this patch, now the relationship between NTOS and CMLIB is harmonic and I can use kernel functions easier.

PR #4620, opened 2022-08-21
[NTUSER] Use assignment-lock against THREADINFO.KeyboardLayout

katahiromz:
## Purpose
Improve stability when switching languages.
JIRA issue: [CORE-11700](https://jira.reactos.org/browse/CORE-11700), [CORE-18317](https://jira.reactos.org/browse/CORE-18317)

## Proposed changes

- When the system is to assign a pointer to `THREADINFO.KeyboardLayout`, it uses `UserAssignmentLock`.
- When the system is to detach a pointer from `THREADINFO.KeyboardLayout`, it uses `UserAssignmentUnlock`.
- Synchronize `ClientInfo->hKL` with `pti->KeyboardLayout->hkl`.

## TODO

- [x] A small test.

### DIFF
--- a/win32ss/user/ntuser/kbdlayout.c
+++ b/win32ss/user/ntuser/kbdlayout.c
@@ -670,11 +670,16 @@ co_UserActivateKeyboardLayout(
     if (pti->TIF_flags & TIF_CSRSSTHREAD)
     {
         UserAssignmentLock((PVOID*)&pti->KeyboardLayout, pKL);
-        ClientInfo->CodePage = pKL->CodePage;
+
+        if (!(pti->TIF_flags & TIF_INCLEANUP))
+        {
+            ClientInfo->CodePage = pKL->CodePage;
+            ClientInfo->hKL = pKL->hkl;
+        }
     }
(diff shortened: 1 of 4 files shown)


win32ss/user/ntuser/kbdlayout.c
```diff
@@ -583,12 +583,10 @@ co_UserActivateKbl(PTHREADINFO pti, PKL pKl, UINT Flags)
     PWND pWnd;
 
     pklPrev = pti->KeyboardLayout;
-    if (pklPrev)
-        UserDereferenceObject(pklPrev);
 
-    pti->KeyboardLayout = pKl;
+    UserAssignmentLock((PVOID*)&(pti->KeyboardLayout), pKl);
```
HBelusca, line 587: No `UserAssignmentUnlock` call for the previous value of `pti->KeyboardLayout` ?

katahiromz: The previous value will released by assignment-lock if not same nor null.

katahiromz:
VirtualBox_ReactOS_23_08_2022_17_36_26
No assertion failure any more.

PR #4625, opened 2022-08-23
[USER32] Get/SetWindowLongPtr about PIMEUI

katahiromz:
## Purpose

## Proposed changes

- In `ImeWndProc_common`, correctly get/set the `PIMEUI` value by `Get/SetWindowLongPtrW`.

### DIFF
--- a/win32ss/user/user32/misc/imm.c
+++ b/win32ss/user/user32/misc/imm.c
@@ -948,32 +948,36 @@ ImeWndProc_common(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lParam, BOOL unicod
                 return ret;
         }
 
-       if (!pWnd->fnid)
-       {
-          if (msg != WM_NCCREATE)
-          {
-             if (unicode)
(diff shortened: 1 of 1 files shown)


win32ss/user/user32/misc/imm.c
```diff
@@ -967,7 +968,7 @@ ImeWndProc_common(HWND hwnd, UINT msg, WPARAM wParam, LPARAM lParam, BOOL unicod
              ERR("Wrong window class for Ime! fnId 0x%x\n",pWnd->fnid);
              return 0;
           }
-          pimeui = ((PIMEWND)pWnd)->pimeui;
```
learn-more, line 970: Can `pimeui` be removed from the `PIMEWND` structure?

katahiromz: The `WND` structure can have extra bytes of `WNDCLASS[EX].cbWndExtra` bytes.
On user land, it is not safe, I think.

PR #4629, opened 2022-08-24
[KBSWITCH] Fix GetNextLayout for three or more languages

katahiromz:
## Purpose

Even when the user installed three or more languages into the system, `Shift+Alt` key combination should work correctly.
JIRA issue: [CORE-11737](https://jira.reactos.org/browse/CORE-11737)

VirtualBox_ReactOS Debug_24_08_2022_14_24_06

For example, we assume the user installed three Latin languages (English, French, and German), then typing `Alt+Shift` should work like EN → FR → DE → EN → FR → DE → ... (cyclically).

## Proposed changes

- Add `GetMaxLayoutNum` helper function.
- Fix `GetNextLayout` function to make `Shift+Alt` cyclic.

## TODO

- [x] Do small test.

### DIFF
--- a/base/applications/kbswitch/kbswitch.c
+++ b/base/applications/kbswitch/kbswitch.c
@@ -331,6 +331,40 @@ BuildLeftPopupMenu(VOID)
     return hMenu;
 }
 
+static ULONG
+GetMaxLayoutNum(VOID)
+{
+    HKEY hKey;
+    ULONG dwIndex, dwSize, uLayoutNum, uMaxLayoutNum = 0;
+    TCHAR szLayoutNum[CCH_ULONG_DEC + 1], szLayoutID[CCH_LAYOUT_ID + 1];
+
+    /* Add the keyboard layouts to the popup menu */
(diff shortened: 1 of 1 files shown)


base/applications/kbswitch/kbswitch.c
```diff
@@ -278,7 +278,7 @@ ActivateLayout(HWND hwnd, ULONG uLayoutNum)
     TCHAR szLayoutNum[CCH_ULONG_DEC + 1], szLCID[CCH_LAYOUT_ID + 1], szLangName[MAX_PATH];
     LANGID LangID;
 
-    if (uLayoutNum == (ULONG)-1 || uLayoutNum > 0xFF) /* Invalid */
+    if (uLayoutNum == 0 || uLayoutNum > 0xFF) /* Invalid */
```
HBelusca, line 281: Why is the value "0" somewhat special?

katahiromz: The layout number begins from one.

katahiromz: Added a comment.

PR #4631, opened 2022-08-24
[CPL] Use MAKEINTRESOURCE instead of LoadString

TAN-Gaming:
## Purpose
- Follow-up of [0ed302e](https://github.com/reactos/reactos/commit/0ed302ef656ed39c932e084508ffa506b47c463b)

## Proposed changes
- Replace unnecessary `LoadString` calls with `MAKEINTRESOURCE`

### DIFF
--- a/dll/cpl/access/access.c
+++ b/dll/cpl/access/access.c
@@ -189,14 +189,11 @@ SystemApplet(HWND hwnd, UINT uMsg, LPARAM wParam, LPARAM lParam)
     PGLOBAL_DATA pGlobalData;
     PROPSHEETPAGE psp[5];
     PROPSHEETHEADER psh;
-    TCHAR Caption[1024];
     INT nPage = 0;
     INT ret;
 
     if (uMsg == CPL_STARTWPARMSW && lParam != 0)
         nPage = _wtoi((PWSTR)lParam);
-
-    LoadString(hApplet, IDS_CPLSYSTEMNAME, Caption, sizeof(Caption) / sizeof(TCHAR));
 
     pGlobalData = HeapAlloc(GetProcessHeap(), HEAP_ZERO_MEMORY, sizeof(GLOBAL_DATA));
     if (pGlobalData == NULL)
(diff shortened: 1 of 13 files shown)


SergeGautherie:
Some cases swap between `TCHAR` and `WCHAR`: are these changes intended?

--

While here, would `MAKEINTRESOURCE` apply to some other CPL cases, even to some other modules?
[https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=LoadString.%2Baption&sr=1]

TAN-Gaming:
@SergeGautherie 

> Some cases swap between TCHAR and WCHAR: are these changes intended?

There's a lot of mixed uses between `TCHAR` and `WCHAR` here (which is incorrect but it's unrelated to this change) so I'll keep `MAKEINTRESOURCE` of `pszCaption` the same as `pszIcon` for readability.

> While here, would MAKEINTRESOURCE apply to some other CPL cases, even to some other modules?

This change can be applied to `PROPSHEETPAGE` and `PROPSHEETHEADER` structure.
[https://docs.microsoft.com/en-us/windows/win32/controls/pss-propsheetpage](https://docs.microsoft.com/en-us/windows/win32/controls/pss-propsheetpage)
[https://docs.microsoft.com/en-us/windows/win32/controls/pss-propsheetheader](https://docs.microsoft.com/en-us/windows/win32/controls/pss-propsheetheader)

learn-more:
Did you verify that this works correctly?

HBelusca (approved):
ok if it works.

TAN-Gaming:
@learn-more It works and it's valid (according to the documentation)

HBelusca:
> @learn-more Tested, it works and it's valid (according to the documentation).

The MSDN doc is not always reliable, so you always need to test it yourself.

TAN-Gaming:
@HBelusca Sorry my bad English, what I meant was

- I tested it, it works.
- It's valid (according to the documentation).

PR #4635, opened 2022-08-26
[WINMM] Winmm fixes for system sound

TAN-Gaming:
## Proposed changes
- Fix incorrect registry key in `get_mmioFromProfile`
- `get_mmioFromProfile` : Expand environment strings if registry value type is `REG_EXPAND_SZ`
- Use `RegOpenCurrentUser` to access the appropriate `HKEY_CURRENT_USER` key. This allows functions that use `HKEY_CURRENT_USER` to be called from system processes while impersonating, for example: winlogon
- Fix `SystemDefault` event and `SND_NODEFAULT` flag

#### Also:
- `get_mmioFromProfile` code formatting
- `PlaySound_IsString` code formatting

### DIFF
--- a/dll/win32/winmm/playsound.c
+++ b/dll/win32/winmm/playsound.c
@@ -121,15 +121,16 @@ static HMMIO	get_mmioFromProfile(UINT uFlags, LPCWSTR lpszName)
     err = RegOpenKeyW(hRegApp, lpszName, &hScheme);
     RegCloseKey(hRegApp);
     if (err != 0) goto none;
-    /* what's the difference between .Current and .Default ? */
-    err = RegOpenKeyW(hScheme, wszDotDefault, &hSnd);
-    if (err != 0)
+    
+    err = RegOpenKeyW(hScheme, wszDotCurrent, &hSnd);
+
+    RegCloseKey(hScheme);
+
+    if (err != ERROR_SUCCESS)
     {
-        err = RegOpenKeyW(hScheme, wszDotCurrent, &hSnd);
-        RegCloseKey(hScheme);
-        if (err != 0)
-            goto none;
+        goto none;
     }
+
     count = sizeof(str);
     err = RegQueryValueExW(hSnd, NULL, 0, &type, (LPBYTE)str, &count);
     RegCloseKey(hSnd);
--- a/dll/win32/winmm/playsound.c
+++ b/dll/win32/winmm/playsound.c
@@ -61,7 +61,7 @@ static HMMIO	get_mmioFromFile(LPCWSTR lpszName)
 static HMMIO	get_mmioFromProfile(UINT uFlags, LPCWSTR lpszName)
 {
     WCHAR	str[128];
-    LPWSTR	ptr;
+    LPWSTR	ptr, pszSnd;
     HMMIO  	hmmio;
     HKEY        hRegSnd, hRegApp, hScheme, hSnd;
     DWORD       err, type, count;
@@ -135,7 +135,42 @@ static HMMIO	get_mmioFromProfile(UINT uFlags, LPCWSTR lpszName)
     err = RegQueryValueExW(hSnd, NULL, 0, &type, (LPBYTE)str, &count);
     RegCloseKey(hSnd);
     if (err != 0 || !*str) goto none;
-    hmmio = mmioOpenW(str, NULL, MMIO_ALLOCBUF | MMIO_READ | MMIO_DENYWRITE);
+
+    if (type == REG_EXPAND_SZ)
+    {
+        /* Get full path length, including the NULL terminator */
+        count = ExpandEnvironmentStringsW(str, NULL, 0);
+        if (count == 0)
+            goto none;
+
+        /* Allocate a buffer for the path */
+        pszSnd = HeapAlloc(GetProcessHeap(), 0, count * sizeof(WCHAR));
+        if (!pszSnd)
+            goto none;
+
+        /* Get the path */
+        if (ExpandEnvironmentStringsW(str, pszSnd, count) == 0)
+        {
+            HeapFree(GetProcessHeap(), 0, pszSnd);
+            goto none;
+        }
+    }
+    else if (type == REG_SZ)
+    {
+        /* The type is REG_SZ, no need to expand */
+        pszSnd = str;
+    }
+    else
+    {
+        /* Invalid type */
+        goto none;
+    }
+
+    hmmio = mmioOpenW(pszSnd, NULL, MMIO_ALLOCBUF | MMIO_READ | MMIO_DENYWRITE);
+
+    if (type == REG_EXPAND_SZ)
+        HeapFree(GetProcessHeap(), 0, pszSnd);
+
     if (hmmio) return hmmio;
  none:
     WARN("can't find SystemSound=%s !\n", debugstr_w(lpszName));
--- a/dll/win32/winmm/playsound.c
+++ b/dll/win32/winmm/playsound.c
@@ -54,42 +54,37 @@ static HMMIO	get_mmioFromFile(LPCWSTR lpszName)
     return 0;
 }
 
-static HMMIO	get_mmioFromProfile(UINT uFlags, LPCWSTR lpszName)
+static HMMIO get_mmioFromProfile(UINT uFlags, LPCWSTR lpszName)
 {
-    WCHAR	str[128];
-    LPWSTR	ptr, pszSnd;
-    HMMIO  	hmmio;
-    HKEY        hUserKey, hRegSnd, hRegApp, hScheme, hSnd;
-    DWORD       err, type, count;
+    WCHAR str[128];
+    LPWSTR ptr, pszSnd;
+    HMMIO hmmio;
+    HKEY hUserKey, hRegSnd, hRegApp, hScheme, hSnd;
+    DWORD err, type, count;
     BOOL bIsDefault;
-
-    static const WCHAR  wszSounds[] = {'S','o','u','n','d','s',0};
-    static const WCHAR  wszDefault[] = {'D','e','f','a','u','l','t',0};
-    static const WCHAR  wszKey[] = {'A','p','p','E','v','e','n','t','s','\\',
-                                    'S','c','h','e','m','e','s','\\',
-                                    'A','p','p','s',0};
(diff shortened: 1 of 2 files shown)


dll/win32/winmm/playsound.c
```diff
@@ -92,7 +92,21 @@ static HMMIO	get_mmioFromProfile(UINT uFlags, LPCWSTR lpszName)
      *      HKCU\AppEvents\Schemes\Apps\.Default
      *      HKCU\AppEvents\Schemes\Apps\<AppName>
      */
-    if (RegOpenKeyW(HKEY_CURRENT_USER, wszKey, &hRegSnd) != 0) goto none;
+    err = RegOpenCurrentUser(KEY_READ, &hUserKey);
```
HBelusca, line 95: For information, the reason why `RegOpenKey(Ex)(HKEY_CURRENT_USER, ...` is not completely equivalent to `RegOpenCurrentUser`, is because in the case of `RegOpenKey`, there is a caching being done for the "predefined" handles (like HKEY_CURRENT_USER), and when impersonating, the cached value is the one still used, instead of a new one corresponding to the impersonated user.

You can add this command ^ above in your commit log, this will help understand what's going on.

HBelusca: Another HKEY_CURRENT_USER that may be buggy for the same reasons, in winmm, is this one:
https://git.reactos.org/?p=reactos.git;a=blob;f=dll/win32/winmm/midimap/midimap.c;hb=83442919e4df0067a7bc72f351fefed667c1737a#l207

HBelusca:
I would suggest having a look at wine's code and try keeping close to it if possible.

dll/win32/winmm/playsound.c
TAN-Gaming: Thanks for the info, I found that Windows has two undocumented functions: `WinmmLogon` and `WinmmLogoff`

winlogon.exe from WinXP SP3:
WinXP_WinMM

`WinmmLogon` maybe the way Windows gets user token directly from winlogon and use it to access the appropriate key.

TAN-Gaming:
@HBelusca 

> I would suggest having a look at wine's code and try keeping close to it if possible.

But, Wine doesn't have to deal with security stuff, does it?

dll/win32/winmm/playsound.c
HBelusca: May also be related to managing the audio service (ensuring it's started)?

HBelusca:
> Wine doesn't have to deal with security stuff, does it?

I was thinking about any possible other fixes for PlaySoundA/W that they may have added (in addition to replacing those "winish" unicode strings into sane ones (with the `L""` syntax) ).

dll/win32/winmm/playsound.c
TAN-Gaming: Based on my research, `WinmmLogon` is used to initialize the audio system on Windows. This function doesn't take user token as a parameter.

I think implementing it is not necessary. A better solution maybe to implement ROS-specific functions in winmm and dynamically load them from winlogon. This way we can use Wine's code with least modification.

Sources:
[https://forum.thegamecreators.com/thread/176905](https://forum.thegamecreators.com/thread/176905)

[https://www.writebug.com/git/greenlight/win32-asm-calculator/src/branch/master/src/Calculator(源程序)/Calculator/Calculator/winmm.inc](https://www.writebug.com/git/greenlight/win32-asm-calculator/src/branch/master/src/Calculator(源程序)/Calculator/Calculator/winmm.inc)

[https://rsdn.org/forum/info/FAQ.mmedia.winmminit](https://rsdn.org/forum/info/FAQ.mmedia.winmminit)

TAN-Gaming: The better solution may be to implement ROS-specific functions in winmm and dynamically load them from winlogon, for example:

```
BOOL WINAPI WinmmRosLogon(_In_ HANDLE hUserToken);
VOID WINAPI WinmmRosLogoff(VOID);
```

HBelusca: But then, why adding those functions at all to begin with? How is this going to help with having PlaySound reading the correct user data from the registry?

TAN-Gaming: `WinmmRosLogon` will be the way winmm gets user token directly from the calling process and will call `RegOpenCurrentUser` internally, if succeed then store the returned `HKEY` as a global variable `hUserKey`.

`WinmmRosLogoff` will close the shared `hUserKey` if needed and set it back to the default (predefined HKEY_CURRENT_USER).

TAN-Gaming: Or we can just use `RegOverridePredefKey` from the winlogon side?

HBelusca: Then since you're going to call RegOpenCurrentUser from winmm, and you're going to use a `hUserKey`, this means that you're going to remove the `RegOpenKeyW(HKEY_CURRENT_USER` call that is currently present, therefore modifying that PlaySound function anyways.

Also, we want (in ReactOS) to *remove* hacks and do actual fixes, not adding new hacks, so this WinmmRosLogon/Logoff ROS-specific exports solution won't be accepted. As mentioned in a past comment, it can be seen that in Vista+ (and Windows 7) there is no such exports anymore, yet sounds work during logon etc.

HBelusca: > Can we just use `RegOverridePredefKey` from winlogon side?

And then if you replace Windows' winmm.dll by ours, on Windows, logon sounds and messagebox sounds will stop working correctly?

TAN-Gaming: > it can be seen that in Vista+ (and Windows 7) there is no such exports anymore, yet sounds work during logon etc.

Is that requires WASAPI to be implemented?

HBelusca: perhaps, but not really. winlogon still uses PlaySound.

TAN-Gaming: Winmm_W7

So `RegOpenCurrentUser` may be the way Win7 does this thing.

TAN-Gaming: > Another HKEY_CURRENT_USER that may be buggy for the same reasons, in winmm, is this one: https://git.reactos.org/?p=reactos.git;a=blob;f=dll/win32/winmm/midimap/midimap.c;hb=83442919e4df0067a7bc72f351fefed667c1737a#l207

Which system process needs this function?

HBelusca: **Any** process (can be a user-mode app started by the user) that impersonates a different user (e.g. one that could play the sounds for different users... ?) and calls any of these winmm functions.

TAN-Gaming: OK, I'll add a fix for it.

dll/win32/winmm/playsound.c
```diff
@@ -121,15 +135,16 @@ static HMMIO	get_mmioFromProfile(UINT uFlags, LPCWSTR lpszName)
     err = RegOpenKeyW(hRegApp, lpszName, &hScheme);
     RegCloseKey(hRegApp);
     if (err != 0) goto none;
-    /* what's the difference between .Current and .Default ? */
```
HBelusca, line 124: Isn't this question still valid?

dll/win32/winmm/playsound.c
```diff
@@ -92,7 +92,21 @@ static HMMIO	get_mmioFromProfile(UINT uFlags, LPCWSTR lpszName)
...
+    }
+
+    err = RegOpenKeyW(hUserKey, wszKey, &hRegSnd);
+
+    RegCloseKey(hUserKey);
+
+    if (err != ERROR_SUCCESS)
+    {
+        goto none;
+    }
```
HBelusca, line 108: Could be simplified (as you did in the other function) by:
```suggestion
    if (err == ERROR_SUCCESS)
    {
        err = RegOpenKeyW(hUserKey, wszKey, &hRegSnd);
        RegCloseKey(hUserKey);
    }

    if (err != ERROR_SUCCESS)
    {
        goto none;
    }
```

dll/win32/winmm/playsound.c
TAN-Gaming: Based on my analysis, .Default means "Windows Default" profile as seen in mmsys.cpl and has nothing to do with the `PlaySound` function.

dll/win32/winmm/playsound.c
TAN-Gaming: OK.

dll/win32/winmm/playsound.c
HBelusca: (To clarify it's about the `.Current` vs. `.Default` values, the one selected in the screenshot below)

So your hypothesis, is that, for each windows event, (e.g. the "WindowsLogon"), there are the different sound profiles (e.g. "Afternoon", "Quirky" etc.), as well as the default Windows profile that would be that `.Default` one?
And so, would the value inside `.Current` reflect whatever value corresponding to the selected profile (or a custom sound if the user changed it via the mmsys.cpl GUI) ?

image

TAN-Gaming: win_sounds

This is how Windows store the sound profiles.

In the example, I created a new profile "ReactOS".

HBelusca: OK that's what I understand. (Unrelated: dunno how you created the new profile, but it seems there's an off-by-one bug in the string length, and it shows up some garbage/the NULL terminator.)

TAN-Gaming: > would the value inside .Current reflect whatever value corresponding to the selected profile (or a custom sound if the user changed it via the mmsys.cpl GUI)

Yes, that's what I found.

> dunno how you created the new profile

Using Control Panel (mmsys.cpl) on WinXP.

> but it seems there's an off-by-one bug in the string length, and it shows up some garbage/the NULL terminator

That may be Windows' bug.

TAN-Gaming:
Thread impersonation works now.

TAN-Gaming:
It's OK now.

learn-more:
@TAN-Gaming What is the status of this PR, does it work yet?
Is this still a blocker for #4755 ?

TAN-Gaming:
> @TAN-Gaming What is the status of this PR, does it work yet? Is this still a blocker for #4755 ?

This PR works fine and yes PR #4755 requires it.

HBelusca (approved):
Final cleanup

PR #4636, opened 2022-08-27
[REGEDIT] Fix possible buffer overflow

learn-more:


base/applications/regedit/find.c
```diff
@@ -416,7 +416,7 @@ BOOL RegFindWalk(
             if (DoEvents())
                 goto err;
 
-            s_cbName = MAX_PATH * sizeof(WCHAR);
+            s_cbName = MAX_PATH;
```
katahiromz, line 419: Can you rename `s_cbName` as s_cchName`?

KRosUser (changes requested):
Same bug at L177 and L270 (and s_cbName has to be renamed s_cchName there too)

KRosUser:
Regarding comment from @katahiromz, you should also apply the same to "cbName" in base/applications/regedit/edit.c in CopyKey function. It represent a character count, not a byte count (logic is OK, but naming is wrong)

learn-more:
> Regarding comment from @katahiromz, you should also apply the same to "cbName" in base/applications/regedit/edit.c in CopyKey function. It represent a character count, not a byte count (logic is OK, but naming is wrong)

I don't intend to make the diff any bigger than it needs to be.

KRosUser:
> > Regarding comment from @katahiromz, you should also apply the same to "cbName" in base/applications/regedit/edit.c in CopyKey function. It represent a character count, not a byte count (logic is OK, but naming is wrong)
> 
> I don't intend to make the diff any bigger than it needs to be.

Worth that i propose a short PR for that ? or is coding style fix not relevant enougth ?

learn-more:
> > > Regarding comment from @katahiromz, you should also apply the same to "cbName" in base/applications/regedit/edit.c in CopyKey function. It represent a character count, not a byte count (logic is OK, but naming is wrong)
> > 
> > 
> > I don't intend to make the diff any bigger than it needs to be.
> 
> Worth that i propose a short PR for that ? or is coding style fix not relevant enougth ?

No, please don't create a PR for such minor stylistic changes.
In general, stuff like that should be handled by a maintainer of the component, although I think there is none at the moment.

KRosUser:
> > > > Regarding comment from @katahiromz, you should also apply the same to "cbName" in base/applications/regedit/edit.c in CopyKey function. It represent a character count, not a byte count (logic is OK, but naming is wrong)
> > > 
> > > 
> > > I don't intend to make the diff any bigger than it needs to be.
> > 
> > 
> > Worth that i propose a short PR for that ? or is coding style fix not relevant enougth ?
> 
> No, please don't create a PR for such minor stylistic changes. In general, stuff like that should be handled by a maintainer of the component, although I think there is none at the moment.

OK !

HBelusca (approved):

katahiromz (approved):

binarymaster (approved):

PR #4642, opened 2022-08-29
[DOC] 3rd Party Files.txt: Add 3rdparty/ to some sdk/lib/ paths

SergeGautherie:
Fix wrong paths.

### DIFF
--- a/media/doc/3rd Party Files.txt
+++ b/media/doc/3rd Party Files.txt
@@ -190,13 +190,13 @@ License: GPL-3.0-or-later (https://spdx.org/licenses/GPL-3.0-or-later.html)
 URL: http://savannah.gnu.org/projects/freefont/
 
 Title: GNU adns
-Path: sdk/lib/adns
+Path: sdk/lib/3rdparty/adns
 Used Version: 1.0 REV 5
 License: GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later.html)
 URL: http://www.gnu.org/software/adns/, http://adns.jgaa.com
 
 Title: CardLib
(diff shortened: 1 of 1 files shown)


media/doc/3rd Party Files.txt
```diff
@@ -147,7 +147,7 @@ URL: http://alter.org.ua/soft/win/uni_ata/
 
 Title: Miniport driver for the Buslogic BT 958 SCSI Controller
 Path: drivers/storage/port/buslogic
-Used Version: v1.2.0.2 (2005-08-17)
+Used Version: 1.2.0.2 (2005-08-17)
```
learn-more, line 150: This is VMWare's numbering scheme, not ours.

learn-more (changes requested):

media/doc/3rd Party Files.txt
SergeGautherie: > VMWare's numbering scheme

Really? I don't know where you got that from.

See
https://git.reactos.org/?p=reactos.git;a=blob;f=drivers/storage/port/buslogic/BusLogic958.rc;hb=HEAD#l26

learn-more: https://git.reactos.org/?p=reactos.git;a=blob;f=drivers/storage/port/buslogic/BusLogic958.c;h=76b6d8b024f8fe3614d836372326a130202ee1fb;hb=HEAD#l75

SergeGautherie: I would rather trust an active line in .rc than a comment in .c.
Also, 'v' looks like an abbreviation to me, not a numbering part.

learn-more (approved):

PR #4640, opened 2022-08-28
[USRMGR] Replace old (leftover) debug MessageBox with Wine macros

KRosUser:
## Purpose

- Current code uses "MessageBox" with leftover of old debug code
- Proposed modification is to use DPRINT instead

JIRA issue: [CORE-18276](https://jira.reactos.org/browse/CORE-18276)

PR #4643, opened 2022-08-29
[EXPLORER] Fix Start Menu context menu actions

KRosUser:
## Purpose

_ Current design is not processing actions verbs correctly for some Start Menu context menu actions (Properties, Open all users, Explore all users) despite associated code being implemented
_ This bug is cased by an incorrect actions id filtering not routing to the appropriate processing

JIRA issue: [CORE-18336](https://jira.reactos.org/browse/CORE-18336)

_ before : Properties, Open all users, Explore all users have no effect
_ after : Properties, Open all users, Explore all users open the appropriate windows

### DIFF
--- a/base/shell/explorer/startctxmnu.cpp
+++ b/base/shell/explorer/startctxmnu.cpp
@@ -185,7 +185,7 @@ public:
         UINT uiCmdId = PtrToUlong(lpici->lpVerb);
         if (uiCmdId != 0)
         {
-            if ((uiCmdId >= ID_SHELL_CMD_FIRST) && (uiCmdId <= ID_SHELL_CMD_LAST))
+            if ((uiCmdId < ID_SHELL_CMD_PROPERTIES ) || (uiCmdId > ID_SHELL_CMD_RESTORE_ALL))
             {
                 CMINVOKECOMMANDINFO cmici = { 0 };
                 CHAR szDir[MAX_PATH];
--- a/base/shell/explorer/startctxmnu.cpp
+++ b/base/shell/explorer/startctxmnu.cpp
@@ -151,7 +151,6 @@ public:
         HRESULT hRet;
 
         psfDesktop = NULL;
-        m_Inner = NULL;
 
         pidlStart = SHCloneSpecialIDList(m_Owner, CSIDL_STARTMENU, TRUE);
 
@@ -169,7 +168,15 @@ public:
                     if (SUCCEEDED(hRet))
                     {
                         CreateContextMenuFromShellFolderPidl(hPopup);
-                        m_idCmdCmLast = ID_SHELL_CMD_FIRST + GetMenuItemCount(hPopup);
+
+                        m_idCmdCmLast = m_Inner->QueryContextMenu(
+                            hPopup,
+                            0,
+                            ID_SHELL_CMD_FIRST,
+                            ID_SHELL_CMD_LAST,
+                            CMF_VERBSONLY);
+                        if(!SUCCEEDED(m_idCmdCmLast))
+                            m_idCmdCmLast = ID_SHELL_CMD_LAST;
                         AddStartContextMenuItems(hPopup);
                     }
                 }
--- a/base/shell/explorer/startctxmnu.cpp
+++ b/base/shell/explorer/startctxmnu.cpp
@@ -136,7 +136,6 @@ public:
     {
         m_TrayWnd = pTrayWnd;
         m_Owner = hWndOwner;
-        m_idCmdCmLast = ID_SHELL_CMD_LAST;
         return S_OK;
     }
 
@@ -227,6 +226,7 @@ public:
 
     CStartMenuBtnCtxMenu()
     {
+        m_idCmdCmLast = ID_SHELL_CMD_LAST;
     }
 
     virtual ~CStartMenuBtnCtxMenu()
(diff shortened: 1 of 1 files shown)


base/shell/explorer/startctxmnu.cpp
```diff
@@ -185,7 +185,7 @@ class CStartMenuBtnCtxMenu :
         UINT uiCmdId = PtrToUlong(lpici->lpVerb);
         if (uiCmdId != 0)
         {
-            if ((uiCmdId >= ID_SHELL_CMD_FIRST) && (uiCmdId <= ID_SHELL_CMD_LAST))
+            if ((uiCmdId < ID_SHELL_CMD_PROPERTIES ) || (uiCmdId > ID_SHELL_CMD_RESTORE_ALL))
```
learn-more, line 188: This is wrong.
As you can observe below, `uiCmdId - ID_SHELL_CMD_FIRST` depends on the fact that it falls in this range.
If it would be a string, with your code you would subtract `ID_SHELL_CMD_FIRST` from the address of the string, and then try to use it as string again, which will crash.

learn-more (changes requested):
This might look like a fix, but it is introducing bad behavior.

KRosUser:
> This might look like a fix, but it is introducing bad behavior.

Behavior was OK (robustness check were present one layer below) but i've re-added the necessary boundary check at the right level.

learn-more:
> > This might look like a fix, but it is introducing bad behavior.
> 
> Behavior was OK (robustness check were present one layer below) but i've re-added the necessary boundary check at the right level and re-ran tests on all menu items

You code will still crash when passed a verb instead of an ID.
You are still passing almost every command to m_Inner, which did not ask for that.

Most likely you will have to catch the HRESULT from the `m_Inner->QueryContextMenu`, extract the number of added items from that using `HRESULT_CODE` and use that as upper bound instead of the hardcoded `ID_SHELL_CMD_LAST`

KRosUser:
> > > This might look like a fix, but it is introducing bad behavior.
> > 
> > 
> > Behavior was OK (robustness check were present one layer below) but i've re-added the necessary boundary check at the right level and re-ran tests on all menu items
> 
> You code will still crash when passed a verb instead of an ID. You are still passing almost every command to m_Inner, which did not ask for that.
> 
> Most likely you will have to catch the HRESULT from the `m_Inner->QueryContextMenu`, extract the number of added items from that using `HRESULT_CODE` and use that as upper bound instead of the hardcoded `ID_SHELL_CMD_LAST`

OK, i will rework.

KRosUser:
@learn-more : Change as proposed (in a simpler manner)

learn-more:
I am curious how you tested this if it doesn't even compile.

base/shell/explorer/startctxmnu.cpp
```diff
@@ -135,6 +136,7 @@ class CStartMenuBtnCtxMenu :
     {
         m_TrayWnd = pTrayWnd;
         m_Owner = hWndOwner;
+        m_idCmdCmLast = ID_SHELL_CMD_LAST;
```
learn-more, line 139: this is what a constructor is for...

KRosUser:
> I am curious how you tested this if it doesn't even compile.

Compiles correctly. Compilation error is on https://github.com/reactos/reactos/pull/4640

base/shell/explorer/startctxmnu.cpp
```diff
@@ -168,6 +170,7 @@ class CStartMenuBtnCtxMenu :
                     if (SUCCEEDED(hRet))
                     {
                         CreateContextMenuFromShellFolderPidl(hPopup);
+                        m_idCmdCmLast = ID_SHELL_CMD_FIRST + GetMenuItemCount(hPopup);
```
learn-more, line 173: There is a reason `QueryContextMenu` returns the amount of id's needed.....

learn-more:
> > I am curious how you tested this if it doesn't even compile.
> 
> Compiles correctly. Compilation error is on #4640

Not with the code you pushed, that's visible even from the diff here.

KRosUser:
> > > I am curious how you tested this if it doesn't even compile.
> > 
> > 
> > Compiles correctly. Compilation error is on #4640
> 
> Not with the code you pushed, that's visible even from the diff here.

Oups, my mistake i'm really sorry. Changed a variable name at the last minute and did not launch a last compile check....

KRosUser:
It's now OK

KRosUser:
@learn-more : Any change still requested before approval ?

learn-more:
Yes, you are still using `GetMenuItemCount`

KRosUser:
> Yes, you are still using `GetMenuItemCount`

This is finally done.
Tested, working for all Start Context menu actions

base/shell/explorer/startctxmnu.cpp
```diff
@@ -169,7 +168,15 @@ class CStartMenuBtnCtxMenu :
                     if (SUCCEEDED(hRet))
                     {
                         CreateContextMenuFromShellFolderPidl(hPopup);
-                        m_idCmdCmLast = ID_SHELL_CMD_FIRST + GetMenuItemCount(hPopup);
+
+                        m_idCmdCmLast = m_Inner->QueryContextMenu(
```
learn-more, line 172: You should not re-query this, since it is already queried on L49! (Depending on how the shell extension is implemented, you now have double items in that menu)
Also, you are using the return value directly, which is incorrect. I explicitly said to use `HRESULT_CODE ` for that.

learn-more (approved):

PR #4644, opened 2022-08-29
[NTUSER] Move TL structure definition to ntuser.h

katahiromz:
## Purpose
Define type in proper place.

# Proposed changes

- Move `struct _TL` and `TL` definitions from `win32.h` to `ntuser.h`.
- Modify the type of `TL.pfnFree` as newly-defined `TL_FN_FREE` function pointer.


katahiromz:
Looks successful. Let's go.

PR #4650, opened 2022-09-01
[NTOSKRNL] Implement PeakCommitment (MmPeakCommitment, MmTotalCommittedPages)

KRosUser:
## Purpose

PeakComitment is currently not implemented, forced to 0
image

Proposal to implement it 
image


HBelusca:
Can you show Windows tasmkgr's switch-to in the very same kind of setup (more than one task selected)?

KRosUser:
> Can you show Windows tasmkgr's switch-to in the very same kind of setup (more than one task selected)?

Yes, it follow the principle of the proposal :
image

KRosUser:
UI aspects carved out to [PR](https://github.com/reactos/reactos/pull/4652)

ntoskrnl/mm/balance.c
```diff
@@ -81,6 +93,7 @@ MmReleasePageMemoryConsumer(ULONG Consumer, PFN_NUMBER Page)
     }
 
     (void)InterlockedDecrementUL(&MiMemoryConsumers[Consumer].PagesUsed);
+    UpdatePeakCommitment();
```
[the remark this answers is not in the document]
KRosUser: I don't think so. Fixed !

ntoskrnl/mm/balance.c
```diff
@@ -39,6 +39,16 @@ static LONG PageOutThreadActive;
 
 /* FUNCTIONS ****************************************************************/
 
+static inline VOID UpdatePeakCommitment(VOID)
+{
+    // HACK: MmTotalCommittedPages should be adjusted consistently with
+    // other counters at different places.
+    MmTotalCommittedPages = MiMemoryConsumers[MC_SYSTEM].PagesUsed + MiMemoryConsumers[MC_USER].PagesUsed + MiUsedSwapPages;
```
SergeGautherie, line 46: Why no update call after `MiUsedSwapPages` increase?

KRosUser:
> Can you show what Windows TaskMgr shows at the same time?

Same data :
image

ntoskrnl/mm/balance.c
KRosUser: Good point. Fixed

KRosUser:
@HBelusca : I see you still have the "change requested" flag. Did I miss something ?

ntoskrnl/mm/balance.c
```diff
@@ -39,6 +39,16 @@ static LONG PageOutThreadActive;
 
 /* FUNCTIONS ****************************************************************/
 
+static inline VOID UpdatePeakCommitment(VOID)
+{
+    // HACK: MmTotalCommittedPages should be adjusted consistently with
+    // other counters at different places.
```
SergeGautherie, line 45: Is this comment still relevant? If yes, which other counters/places would you have in mind, for example?

KRosUser: Was added by @HBelusca . @HBelusca : What's your view on this comment now ?

ntoskrnl/mm/balance.c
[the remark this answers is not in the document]
KRosUser: Right

ntoskrnl/mm/balance.c
```diff
@@ -50,6 +60,8 @@ MmInitializeBalancer(ULONG NrAvailablePages, ULONG NrSystemPages)
     MiMinimumAvailablePages = 256;
     MiMinimumPagesPerRun = 256;
     MiMemoryConsumers[MC_USER].PagesTarget = NrAvailablePages / 2;
+    MmPeakCommitment = 0;
+    UpdatePeakCommitment();
```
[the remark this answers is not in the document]
KRosUser: Because i didn't knew globals were implicitely initialized at 0x0

ntoskrnl/mm/balance.c
```diff
@@ -50,6 +60,8 @@ MmInitializeBalancer(ULONG NrAvailablePages, ULONG NrSystemPages)
     MiMinimumAvailablePages = 256;
     MiMinimumPagesPerRun = 256;
     MiMemoryConsumers[MC_USER].PagesTarget = NrAvailablePages / 2;
+    MmPeakCommitment = 0;
```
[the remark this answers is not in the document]
KRosUser: I didn't knew. fixed.

ntoskrnl/mm/balance.c
HBelusca: My view: Since we are not experts in Mm, and this PR isn't an actual serious attempt at getting all the Mm counters consistently implemented, the purpose of this comment is to tell any future Mm-knowledgeable dev to double-check and possibly fix where MmTotalCommittedPages (and the rest) needs to be updated.

HBelusca:
Asking review to @ThFabba @tkreuzer @zefklop regarding the correctness of where the MmTotal...Pages counter needs to be modified in order to fairly represent the current total number of pages allocated.

ntoskrnl/mm/balance.c
KRosUser: OK so let's keep this comment anyway

KRosUser:
> Asking review to @ThFabba @tkreuzer @zefklop regarding the correctness of where the MmTotal...Pages counter needs to be modified in order to fairly represent the current total number of pages allocated.

Yes, keeping in mind that this  PR isn't an actual serious attempt at getting all the Mm counters consistently implemented,

ntoskrnl/ex/sysinfo.c
```diff
@@ -718,17 +718,15 @@ QSI_DEF(SystemPerformanceInformation)
      *   Add up all the used "Committed" memory + pagefile.
      *   Not sure this is right. 8^\
      */
-    Spi->CommittedPages = MiMemoryConsumers[MC_SYSTEM].PagesUsed +
-                          MiMemoryConsumers[MC_USER].PagesUsed +
-                          MiUsedSwapPages;
+    Spi->CommittedPages = MmTotalCommittedPages;
```
SergeGautherie, line 721: Calculation is moved, so should its related comment.

ntoskrnl/mm/balance.c
```diff
@@ -39,6 +39,18 @@ static LONG PageOutThreadActive;
 
 /* FUNCTIONS ****************************************************************/
 
+static inline VOID UpdatePeakCommitment(VOID)
```
SergeGautherie, line 42:
```suggestion
static inline VOID UpdateTotalCommittedPages(VOID)
```
PeakCommitment is a consequence here.

ntoskrnl/mm/balance.c
```diff
@@ -39,6 +39,18 @@ static LONG PageOutThreadActive;
...
+{
+    // HACK: MmTotalCommittedPages should be adjusted consistently with
+    // other counters at different places.
+    MmTotalCommittedPages = MiMemoryConsumers[MC_SYSTEM].PagesUsed +
+                          MiMemoryConsumers[MC_USER].PagesUsed +
+                          MiUsedSwapPages;
```
SergeGautherie, line 48:
```suggestion
    MmTotalCommittedPages = MiMemoryConsumers[MC_SYSTEM].PagesUsed +
                            MiMemoryConsumers[MC_USER].PagesUsed +
                            MiUsedSwapPages;
```

ntoskrnl/mm/pagefile.c
```diff
@@ -89,6 +89,16 @@ static BOOLEAN MmSystemPageFileLocated = FALSE;
 
 /* FUNCTIONS *****************************************************************/
 
+static inline VOID UpdatePeakCommitment(VOID)
+{
+    // HACK: MmTotalCommittedPages should be adjusted consistently with
+    // other counters at different places.
+    MmTotalCommittedPages = MiMemoryConsumers[MC_SYSTEM].PagesUsed + MiMemoryConsumers[MC_USER].PagesUsed + MiUsedSwapPages;
```
SergeGautherie, line 96: Since you duplicate the function, keep it in sync!

ThFabba: Don't duplicate it, please. You can put a `static inline` function in a header file if needed.

ntoskrnl/mm/balance.c
```diff
@@ -39,6 +39,18 @@ static LONG PageOutThreadActive;
 
 /* FUNCTIONS ****************************************************************/
 
+static inline VOID UpdatePeakCommitment(VOID)
+{
+    // HACK: MmTotalCommittedPages should be adjusted consistently with
+    // other counters at different places.
+    MmTotalCommittedPages = MiMemoryConsumers[MC_SYSTEM].PagesUsed +
```
ThFabba, line 46: Does the caller of this function guarantee that some kind of global lock is held? If yes, please document that (or assert in code that the lock is held if possible, as well as use SAL to annotate the function).
If there's no lock, this should likely be an `InterlockedCompareExchange` loop. Depending on how critical the accuracy of this value is (is it used to enforce memory limits or only for returning information to user mode callers?), the loop might have to repeat until it observes that none of the three source values have changed during the calculation.

ThFabba:
> Asking review to @ThFabba @tkreuzer @zefklop regarding the correctness of where the MmTotal...Pages counter needs to be modified in order to fairly represent the current total number of pages allocated.

I don't know about Windows comaptibility -- but given the three source values we currently calculate it from, updating it in the locations where those values change, as seems to be done here, makes sense to me.

ntoskrnl/mm/balance.c
```diff
@@ -284,6 +296,7 @@ MmRequestPageMemoryConsumer(ULONG Consumer, BOOLEAN CanWait,
 
     /* Update the target */
     InterlockedIncrementUL(&MiMemoryConsumers[Consumer].PagesUsed);
+    UpdatePeakCommitment();
```
ThFabba, line 299: Actually, I think we're missing a matching update in `MmReleasePageMemoryConsumer` where this gets decremented again.

ntoskrnl/mm/pagefile.c
```diff
@@ -330,7 +340,7 @@ MmAllocSwapPage(VOID)
             MiUsedSwapPages++;
             MiFreeSwapPages--;
             KeReleaseGuardedMutex(&MmPageFileCreationLock);
-
+            UpdatePeakCommitment();
```
ThFabba, line 343: `MmFreeSwapPage` should also call this, to keep the `Total` value accurate

ntoskrnl/mm/balance.c
ThFabba: Actually, given how the `PagesUsed` and `MiUsedSwapPages` counters are updated, perhaps this function should just take a parameter to indicate increment/decrement, and do an `InterlockedIncrement` or `InterlockedDecrement` on the Total value. Much simpler than turning this into a loop or adding a lock.

ntoskrnl/mm/balance.c
```diff
@@ -39,6 +39,18 @@ static LONG PageOutThreadActive;
...
+static inline VOID UpdatePeakCommitment(VOID)
+{
+    // HACK: MmTotalCommittedPages should be adjusted consistently with
+    // other counters at different places.
+    MmTotalCommittedPages = MiMemoryConsumers[MC_SYSTEM].PagesUsed +
+                          MiMemoryConsumers[MC_USER].PagesUsed +
+                          MiUsedSwapPages;
+
+    if (MmTotalCommittedPages > MmPeakCommitment)
+        MmPeakCommitment = MmTotalCommittedPages;
```
ThFabba, line 51: Related to the locking question above... atomically updating a maximum value looks something like this:
```
SIZE_T TotalCommittedPages = ...; // obtain this value atomically above instead of re-reading from memory
SIZE_T PeakCommitment = MmPeakCommitment;
while (TotalCommittedPages > PeakCommitment &&
       InterlockedCompareExchangeSizeT(&MmPeakCommitment, TotalCommittedPages, PeakCommitment) != PeakCommitment)
{
    PeakCommitment = MmPeakCommitment;
}
```

ntoskrnl/ex/sysinfo.c
KRosUser: Yes

ntoskrnl/mm/balance.c
KRosUser: ok

ntoskrnl/mm/balance.c
KRosUser: yes

ntoskrnl/mm/pagefile.c
KRosUser: yes

ntoskrnl/mm/pagefile.c
KRosUser: I used it once now with a ` static inline` declaration in a new mm.h but now it no longer compiles due to : 

```
In file included from ../ntoskrnl/mm/balance.c:18:
../ntoskrnl/mm/mm.h:2:20: error: 'UpdateTotalCommittedPages' used but never defined [-Werror]
 static inline VOID UpdateTotalCommittedPages(VOID);
                    ^~~~~~~~~~~~~~~~~~~~~~~~~
cc1.exe: all warnings being treated as errors
```

Could you help me please ?

katahiromz: inline function `UpdateTotalCommittedPages` needs its definition. Where is the contents?

ntoskrnl/mm/balance.c
KRosUser: the "lock" discussion goes beyond my understanding of this module.
The current criticity is very low (display of "peak" data in TaskMgr), so i'd propose to go on with this "first simple" implementation and let another more experienced dev continue with another PR to make it safer and reader for more critical use case, if any. Would it make sense to you @ThFabba & @HBelusca ? Otherwise, i'd need more explicit contributions from you to this PR on this specific item :-)

ntoskrnl/mm/balance.c
KRosUser: See previous comment / question regarding lock

ntoskrnl/mm/pagefile.c
KRosUser: defined in mm.h and implemented in balance.c

KRosUser: (sorry i'm not familiar with `static inline`)

KRosUser: Done

ntoskrnl/mm/mm.h
```diff
@@ -0,0 +1,2 @@
+#pragma once
+static inline VOID UpdateTotalCommittedPages(VOID);
```
HBelusca, line 2: Why creating this file, instead of using the existing `ntoskrnl/include/internal/mm.h` ?
Also, it's in that .h that the implementation of the inline UpdateTotalCommittedPages() function must be placed.

ntoskrnl/mm/pagefile.c
```diff
@@ -31,6 +31,7 @@
 #include <ntoskrnl.h>
 #define NDEBUG
 #include <debug.h>
+#include "mm.h"
```
HBelusca, line 34:
```suggestion
```

ntoskrnl/mm/balance.c
```diff
@@ -15,8 +15,7 @@
 #include <debug.h>
 
 #include "ARM3/miarm.h"
-
-
+#include "mm.h"
```
HBelusca, line 18: Please revert this change as well.

ntoskrnl/mm/balance.c
```diff
@@ -39,6 +38,23 @@ static LONG PageOutThreadActive;
 
 /* FUNCTIONS ****************************************************************/
 
+static inline VOID UpdateTotalCommittedPages(VOID)
+{
+    /*
+     *   Add up all the used "Committed" memory + pagefile.
+     *   Not sure this is right. 8^\
+     *   MmTotalCommittedPages should be adjusted consistently with
+    *    other counters at different places.
```
HBelusca, line 47:
```suggestion
     * Add up all the used "Committed" memory + pagefile.
     * Not sure this is right. 8^\
     * MmTotalCommittedPages should be adjusted consistently with
     * other counters at different places.
```

ntoskrnl/mm/balance.c
```diff
@@ -39,6 +38,23 @@ static LONG PageOutThreadActive;
 
 /* FUNCTIONS ****************************************************************/
 
+static inline VOID UpdateTotalCommittedPages(VOID)
```
HBelusca, line 41: That inline function should be moved into the mm.h (and the "static" be removed).

ntoskrnl/mm/mm.h
KRosUser: ok

ntoskrnl/mm/pagefile.c
KRosUser: ok

ntoskrnl/mm/balance.c
KRosUser: ok

ntoskrnl/mm/balance.c
KRosUser: ok

ntoskrnl/mm/pagefile.c
```diff
@@ -31,6 +31,7 @@
 #include <ntoskrnl.h>
 #define NDEBUG
 #include <debug.h>
+#include "../include/internal/mm.h"
```
HBelusca, line 34: This header should already be automatically included, so this line should be not needed.

ntoskrnl/mm/balance.c
```diff
@@ -15,8 +15,7 @@
 #include <debug.h>
 
 #include "ARM3/miarm.h"
-
-
+#include "../include/internal/mm.h"
```
HBelusca, line 18: Ditto.

KRosUser: Done

ntoskrnl/mm/pagefile.c
KRosUser: Done

ntoskrnl/mm/balance.c
ThFabba: You can just use my second suggestion, it's much simpler. Something like this should do the trick -- and then you call it with either +1 or -1 depending on whether the calling code just increased or decreased the commitment.
```
inline VOID UpdateTotalCommittedPages(LONG Increase)
{
    SIZE_T TotalCommittedPages = InterlockedExchangeAddSizeT(&MmTotalCommittedPages, Increase) + Increase;

    SIZE_T PeakCommitment = MmPeakCommitment;
    while (TotalCommittedPages > PeakCommitment &&
           InterlockedCompareExchangeSizeT(&MmPeakCommitment, TotalCommittedPages, PeakCommitment) != PeakCommitment)
    {
        PeakCommitment = MmPeakCommitment;
    }
}
```

ntoskrnl/mm/balance.c
```diff
@@ -15,8 +15,6 @@
 #include <debug.h>
 
 #include "ARM3/miarm.h"
```
HBelusca, line 17: Restoring original newlines
```suggestion
#include "ARM3/miarm.h"


```

HBelusca (commented):
Missing: the lock-free considerations from Thomas :thinking:

zefklop (approved):
I'd say it's good enough for now. There are other places where we should update those values, but that would be for later.

KRosUser:
> I'd say it's good enough for now.

You mean no need for the lock stuff ? @HBelusca ?

zefklop:
> 

Oh I didn't see this part of the conversation. Reading.

ntoskrnl/mm/balance.c
zefklop: Yes, please use @ThFabba suggestion. At first I thought this was done under PFN lock, but that's obviously not the case.

KRosUser:
> > 
> 
> Oh I didn't see this part of the conversation. Reading.

OK

ntoskrnl/mm/pagefile.c
```diff
@@ -329,8 +330,9 @@ MmAllocSwapPage(VOID)
             }
             MiUsedSwapPages++;
             MiFreeSwapPages--;
-            KeReleaseGuardedMutex(&MmPageFileCreationLock);
+            UpdateTotalCommittedPages();
```
ThFabba, line 333: These two calls were good outside of the lock, like you had them before. That is, assuming we're making the updates interlocked.

ntoskrnl/mm/balance.c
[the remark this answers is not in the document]
SergeGautherie: My bad for leading you to remove this call, as we want to recalculate the total nonetheless, as Thomas pointed out.

ntoskrnl/mm/balance.c
SergeGautherie: > LONG Increase

Good point. To further optimize, we could have 2 functions, as decrease never updates peak.

HBelusca: That's a pointless optimization. Just doing `if (Increase < 0) return;` before the while() loop should be sufficient.

SergeGautherie: My suggestion is to reduce compiled code size and increase execution speed. (And I find it more explicit / simpler.)
As always, do it as you prefer..

HBelusca: > to reduce compiled code size and increase execution speed

So we understand you actually did all those verifications before proposing your suggestions?

ntoskrnl/mm/balance.c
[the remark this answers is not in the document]
KRosUser: No problem, I put it back since then.

ntoskrnl/mm/balance.c
KRosUser: Implementation done. I'm testing and i'll commit after if works fine.

KRosUser:
"Locked" implementation as proposed by @ThFabba has been implemented with the optimization suggested by @HBelusca 
Tested and ready to go !

ntoskrnl/include/internal/mm.h
```diff
@@ -866,6 +866,36 @@ NTAPI
 MmDeleteKernelStack(PVOID Stack,
                     BOOLEAN GuiStack);
 
+/* balance.c / pagefile.c******************************************************/
+
+inline VOID UpdateTotalCommittedPages(LONG Increase)
```
zefklop, line 871: Can you call that `Delta` or similar please ? Negative increase  is not an increase anymore 🙂

ntoskrnl/include/internal/mm.h
```diff
@@ -866,6 +866,36 @@ NTAPI
...
+                            MiUsedSwapPages;
+     
+     */
+    
+    /* Update Commitment */
+    SIZE_T TotalCommittedPages = InterlockedExchangeAddSizeT(&MmTotalCommittedPages, Increase) + Increase;
+
+    SIZE_T PeakCommitment = MmPeakCommitment;
+    
+    if (Increase < 0) return; //if we decrement, we exit here to optimize processing (peak not impacted)
```
zefklop, line 889: If that's a nagative, yout guaranteed not to enter the while loop below.

KRosUser: Done

ntoskrnl/include/internal/mm.h
KRosUser: Done

ntoskrnl/mm/pagefile.c
KRosUser: Done

zefklop (approved):
Approving again. Looks better now.

ntoskrnl/include/internal/mm.h
```diff
@@ -866,6 +866,34 @@ NTAPI
...
+    SIZE_T TotalCommittedPages = InterlockedExchangeAddSizeT(&MmTotalCommittedPages, Delta) + Delta;
+
+    SIZE_T PeakCommitment = MmPeakCommitment;
+
+    /* Update Peak */
+    while (TotalCommittedPages > PeakCommitment &&
+           InterlockedCompareExchangeSizeT(&MmPeakCommitment, TotalCommittedPages, PeakCommitment) != PeakCommitment)
+    {
+        PeakCommitment = MmPeakCommitment;
+    }
```
HBelusca, line 894: Side remark: an explanation of this pattern can be found e.g. at
https://devblogs.microsoft.com/oldnewthing/20040915-00/?p=37863

ntoskrnl/include/internal/mm.h
```diff
@@ -866,6 +866,34 @@ NTAPI
...
+     * MmTotalCommittedPages should be adjusted consistently with
+     * other counters at different places.
+         MmTotalCommittedPages = MiMemoryConsumers[MC_SYSTEM].PagesUsed +
+                            MiMemoryConsumers[MC_USER].PagesUsed +
+                            MiUsedSwapPages;
+     
+     */
+    
+    /* Update Commitment */
+    SIZE_T TotalCommittedPages = InterlockedExchangeAddSizeT(&MmTotalCommittedPages, Delta) + Delta;
```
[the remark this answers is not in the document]
ThFabba: `InterlockedExchangeAdd` returns the "old" value. We want the new value, which is larger by `Delta`

ntoskrnl/include/internal/mm.h
ThFabba: Thanks, I was struggling to come up with a better name haha

ntoskrnl/mm/pagefile.c
```diff
@@ -329,8 +330,9 @@ MmAllocSwapPage(VOID)
             }
             MiUsedSwapPages++;
             MiFreeSwapPages--;
-            KeReleaseGuardedMutex(&MmPageFileCreationLock);
+            UpdateTotalCommittedPages(1);
```
ThFabba, line 333: As I think I said somewhere before, this can be called outside of the page file creation lock. But NBD either way.

ThFabba (approved):

ntoskrnl/include/internal/mm.h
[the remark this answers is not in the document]
HBelusca: Aha good point :+1:

ntoskrnl/mm/pagefile.c
KRosUser: Would you like me to swap both lines or is it ok for approval?

ntoskrnl/include/internal/mm.h
KRosUser: Do you want me to add this as comment in code ?

ntoskrnl/mm/pagefile.c
HBelusca: I'm personally fine with updating all counters at once (instead of having some counters with the lock held, then the other counter(s) with the lock unheld).

KRosUser: So I keep it as is since it's ok for @ThFabba too.

I let you answer on the comment and we're done !

ntoskrnl/include/internal/mm.h
HBelusca: No, it was just for your (and possible other readers) personal knowledge.

KRosUser: Ok, thanks!

HBelusca (approved):

ntoskrnl/mm/balance.c
SergeGautherie: > you actually did all those verifications

No, I simply assume that less parameters/variables/code in C results in less binary code.

ThFabba: Are you forgetting that you'd also be duplicating the function prologue/epilogue, alignment and the InterlockedExchangeAdd? It's not at all obvious that it would result in smaller code size - and it definitely wouldn't result in less C code, so I'm not sure how you get to this conclusion.

PR #4651, opened 2022-09-01
[EXPLORER] Fix zombie window created by CTRL+ALT+DEL

KRosUser:
Fix for :
JIRA issue: [CORE-18258](https://jira.reactos.org/browse/CORE-18258)

Based on patch by [I_Kill_Bugs](https://jira.reactos.org/secure/ViewProfile.jspa?name=I_Kill_Bugs) validated by @HBelusca as a suitable solution

### DIFF
--- a/base/shell/explorer/traywnd.cpp
+++ b/base/shell/explorer/traywnd.cpp
@@ -2110,7 +2110,7 @@ ChangePos:
 
     BOOL STDMETHODCALLTYPE IsSpecialHWND(IN HWND hWnd)
     {
-        return (m_hWnd == hWnd ||
+        return (m_hWnd == hWnd || hWnd == GetDesktopWindow() ||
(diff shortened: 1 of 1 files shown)


learn-more:
@HBelusca I don't understand why `m_DesktopWnd` does not catch this case, and I don't have the time to investigate that so I won't be reviewing this.

KRosUser:
> @HBelusca I don't understand why `m_DesktopWnd` does not catch this case, and I don't have the time to investigate that so I won't be reviewing this.

This precise question was discussed between @HBelusca and I_Kill_Bugs here : https://jira.reactos.org/browse/CORE-18258?focusedCommentId=134027&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-134027

PR #4653, opened 2022-09-02
[USER32] Fix CascadeWindows()

KRosUser:
## Purpose

- Current implementation of CascadeWindows is broken, the sequence of Z-Order is not correct when finally rendered

JIRA issue: [CORE-18343](https://jira.reactos.org/browse/CORE-18343)

## Proposed changes

- The root cause is about the use of NtUserEndDeferWindowPosEx(hDWP, TRUE) which does an ASYNC processing of the defered positionning. 

## Before

image

## After

TBD

### DIFF
--- a/win32ss/user/user32/windows/mdi.c
+++ b/win32ss/user/user32/windows/mdi.c
@@ -2386,7 +2386,7 @@ TileWindows(HWND hwndParent, UINT wFlags, LPCRECT lpRect,
         ++ret;
     }
 
-    NtUserEndDeferWindowPosEx(hDWP, TRUE);
+    EndDeferWindowPos(hDWP);
 
     if (hwndPrev)
         SetForegroundWindow(hwndPrev);
--- a/win32ss/user/user32/windows/mdi.c
+++ b/win32ss/user/user32/windows/mdi.c
@@ -2192,7 +2192,7 @@ CascadeWindows(HWND hwndParent, UINT wFlags, LPCRECT lpRect,
         ++ret;
     }
 
-    NtUserEndDeferWindowPosEx(hDWP, TRUE);
+    EndDeferWindowPos(hDWP);
 
     if (hwndPrev)
         SetForegroundWindow(hwndPrev);


KRosUser:
@katahiromz : For review, please.

katahiromz:
Investigating by API Monitor:
無題
Looks like correct.

You can also fix `TileWindows` function.

KRosUser:
Fix for TileWindow added, even if less functional impact (as no "Z order challenge" in tiles). But at least it makes design consistent and deterministic.
Thanks !

@katahiromz : Would you approve the PR as such ?

katahiromz (approved):

HBelusca:
> Investigating by API Monitor: [...]

As you can see, you haven't told it to display the actual NtUser*** functions, so you don't know whether NtUserEndDeferWindowPosEx was actually directly called, and with which parameters.

HBelusca:
> The root cause is about the use of NtUserEndDeferWindowPosEx(hDWP, TRUE) which does an ASYNC processing of the defered positionning. Need to have a SYNC processing to ensure sequence is preserved.

This does *not* explain why an ASYNC does not preserve the windows sequence, which seems to be the actual root of the problem...

KRosUser:
> > Investigating by API Monitor: [...]
> 
> As you can see, you haven't told it to display the actual NtUser*** functions, so you don't know whether NtUserEndDeferWindowPosEx was actually directly called, and with which parameters.

KRosUser:
Sorry, my mistake, I closed the PR, while I wanted to post a comment... reopened

KRosUser:
> > The root cause is about the use of NtUserEndDeferWindowPosEx(hDWP, TRUE) which does an ASYNC processing of the defered positionning. Need to have a SYNC processing to ensure sequence is preserved.
> 
> This does _not_ explain why an ASYNC does not preserve the windows sequence, which seems to be the actual root of the problem...

I do agree on the fact that something has to be investigated, but design with a SYNC update is totally fine here. Maybe should we need a separate PR & ticket ? Or would you prefer to keep the whole stuff open until someone make it clear about NtUserEndDeferWindowPosEx ? (to be honest, this is beyond my skills)

learn-more:
> > > The root cause is about the use of NtUserEndDeferWindowPosEx(hDWP, TRUE) which does an ASYNC processing of the defered positionning. Need to have a SYNC processing to ensure sequence is preserved.
> > 
> > 
> > This does _not_ explain why an ASYNC does not preserve the windows sequence, which seems to be the actual root of the problem...
> 
> I do agree on the fact that something has to be investigated, but design with a SYNC update is totally fine here. Maybe should we need a separate PR & ticket ? Or would you prefer to keep the whole stuff open until someone make it clear about NtUserEndDeferWindowPosEx ? (to be honest, this is beyond my skills)
> 
> @HBelusca : Would you accept to go on with this specific PR as is ?

Why is a SYNC update totally fine here?
That it appears to work does not mean it is correct.

KRosUser:
> > > > The root cause is about the use of NtUserEndDeferWindowPosEx(hDWP, TRUE) which does an ASYNC processing of the defered positionning. Need to have a SYNC processing to ensure sequence is preserved.
> > > 
> > > 
> > > This does _not_ explain why an ASYNC does not preserve the windows sequence, which seems to be the actual root of the problem...
> > 
> > 
> > I do agree on the fact that something has to be investigated, but design with a SYNC update is totally fine here. Maybe should we need a separate PR & ticket ? Or would you prefer to keep the whole stuff open until someone make it clear about NtUserEndDeferWindowPosEx ? (to be honest, this is beyond my skills)
> > @HBelusca : Would you accept to go on with this specific PR as is ?
> 
> Why is a SYNC update totally fine here? That it appears to work does not mean it is correct.

Yes something may be wrong in the lower level NtUserEndDeferWindowPosEx. But alternatively, why would NtUserEndDeferWindowPosEx be needed here and not the standard EndDeferWindowPos which finally works totally fine ?

HBelusca:
> But alternatively, why would NtUserEndDeferWindowPosEx be needed here and not the standard EndDeferWindowPos which finally works totally fine ?

Precisely in order *not* to hang the caller of the Cascade/TileWindows functions, by doing an ASYNC window positioning (by specifying the flag to the NtUserEndDeferWindowPosEx function. The standard EndDeferWindowPos doesn't do ASYNC repositioning.)

KRosUser:
> > But alternatively, why would NtUserEndDeferWindowPosEx be needed here and not the standard EndDeferWindowPos which finally works totally fine ?
> 
> Precisely in order _not_ to hang the caller of the Cascade/TileWindows functions, by doing an ASYNC window positioning (by specifying the flag to the NtUserEndDeferWindowPosEx function. The standard EndDeferWindowPos doesn't do ASYNC repositioning.)
> 
> And if a function X (here, the Cascade/TileWindows) calls a buggy function Y that doesn't fully do what it's supposed to do, you don't "fix" X. Instead you have to fix Y .

OK, got it. In this case, I have no clue at where the problem lies with WM_ASYNC_SETWINDOWPOS. Should I close this PR and leave the floor to somebody else ? Or does anyone has a clue to guide me ?

KRosUser:
Second thoughts on the topic looking at how CascadeWindow is built, it's a sequence of positionning with HWND_TOP with no relative Z-Order between each windows

As a conseqence, i'm not even sure that using a ASYNC way to manage the processing of defered Windows position would work, as we would loose the guarantee of the sequence, isn't it ?

Furthermore : Looking at @katahiromz capture of API calls, Windows used EndDeferWindowPos, which means the "sync" implementation, likely for this reason. Furthermore, would using a different API interface lead to other inconsistent dynamic behavior ?

Not sure, so please let me know if I miss something obvious.

HBelusca:
One cannot conclude anything just by the screenshot, because it doesn't tell whether CascadeWindows/TileWindows was called.

katahiromz:
On Win2k3:
```txt
(Break by WinDBG)
kd> !process 0 0
...(process list)...
kd> .process (The process address of explorer.exe)
kd> .reload /user
kd> bp USER32!EndDeferWindowPos
kd> bp win32k!NtUserEndDeferWindowPosEx
kd> g
...(Choose Cascade)...
Breakpoint 0 hit
kd> kb
```
無題

katahiromz (changes requested):
Thus, we have to fix `NtUserEndDeferWindowPosEx(***, TRUE)`.

KRosUser:
To be continued by @katahiromz

katahiromz:
Closed.

PR #4656, opened 2022-09-02
[RAPPS] Fix view mode error management

KRosUser:
## Purpose

When unable to change view mode, current design :
- show an error message
- but switch the view mode indication to the mode that just failed

Expected design would be to revert the view mode to the last mode that was successfully activated.

JIRA issue: [CORE-17274](https://jira.reactos.org/browse/CORE-17274)

Error message proposed to be localized as it may be required even after tiled mode gets implements (eg : SetViewMode may fail for whatever reason)


learn-more:
You are still localizing a workaround for a bug, which makes no sense.
A better approach would be to fix the real bug.

PR #4660, opened 2022-09-02
[AUDIO] Device management code refactoring

oleg-dubinskiy:
## Proposed changes
[WDMAUD.DRV]
Completely rewrite Legacy and MMixer routines, to make them more compatible with Windows.
- Make it independent from MMEBuddy library and make it using its own MME routines instead of MMEBuddy's.
- Get rid from SOUND_DEVICE_INSTANCE structure usage, use WDMAUD_DEVICE_INFO instead. SOUND_DEVICE_INSTANCE has the same purpose as WDMAUD_DEVICE_INFO, and Windows does not use multiple structures, so we also don't need that, to be more compatible with MS. Separate all device state related stuff into WDMAUD_DEVICE_STATE structure. Rework them t the look similar to Windows's (based on analysis). Set alignment 1 for both of them, to make it of the same size as in Windows XP/2003. This gets us better compatibility (and therefore ability to work together) between our/MS wdmaud.drv/sys ring. Although it still doesn't work, but the major steps are done.
- Rewrite all Legacy rotines to make them more similar to MS. Add their own WdmAudIoControl, and don't use SyncOverlappedDeviceIoControl from MMEBuddy.
- Don't use separate members for each data buffer in WDMAUD_DEVICE_INFO structure, declare one buffer member, which can store several types of data passed by the caller (as in Windows). Add a second member which indicates the size of passed data.
- Adapt all MMixer routines to other changes. Use WDMAUD_DEVICE_INFO structure for all of them too, instead of SOUND_DEVICE_INSTANCE.
- Implement proper resampling support. Rename WriteFileEx_Remixer to WdmAudResampleStream, and rewrite it to only do the resampling, remove all streaming code from it, which is buggy anyway. Call it in WdmAudSubmitWaveHeader (reworked from WdmAudCommitWaveBuffer), to properly resample wave header data before sending it. Add RESAMPLING_ENABLED define which allows to enable/disable resampling support.
- Don't use MMFUNCTION_TABLE structure from MMEBuddy, extend usage of FUNC_NAME macro instead, which redirects to Legacy or MMixer API appropriatedly, depending on USE_MMIXER_LIB define.
- Remove WdmAudQueryDeviceInterfaceString, since it does not exist in Windows. The device interface string is already received by calling SetupApi routines and is stored in WDMAUD_DEVICE_INFO structure.
- Refactor opening audio device mechanism. Add WdmAudOpenKernelSoundDevice (previously WdmAudOpenSoundDevice), which does the initial initialization of Wdm Audio driver mapper, and WdmAudOpenSoundDevice, which opens the device itself (previously SetWaveDeviceFormat). This is more similar to Windows.
- Implement thread management, represented by WdmAudCreate/DestroyCompletionThread.
- Completely rewrite Wave Header extension management. In Windows, header extension has two members: streaming device info pointer (aka sound private handle), allocated and used by SubmitWaveHeader for I/O operations, and a pointer to OVERLAPPED structure which contains steaming event handle. There is no committed and completed bytes counts.

[WDMAUD]
Fix all kernel-mode routines, whose are used by Legacy APIs. Adapt them to current changes also.
- Fix all IOCTL declarations. Add new/update existing IOCTLs to Windows-compatible state. I investigated them by the following way: 1. Analyze them via debugger, when they are passed to DeviceIoControl. 2. Copy their handle and paste it to any IOCTL decoder (I used OSR online IOCTL Decoder for example: https://www.osronline.com/article.cfm%5Earticle=229.htm). 3. Decode them ang get their actual declarations, those can be used by us.
- Use Buffer member of WDMAUD_DEVICE_INFO when possible for storing user data received from/passed to the caller.
- Implement some more IOCTLs whose exist in Windows. Add initialization code, and rewrite streaming routine. Use KsStreamIo for that, instead of dumb calling the driver via IoCallDriver. That KS routine is buggy, so it also needs a fix, which I will send in a separate PR.
- Extend usage of MMixer library. Implement MMixerGetWavePosition, and use it for both Legacy and MMixer routines (in wdmaud.sys and wdmaud.drv appropriately).

[MMIXER]
- Add support for WAVE_FORMAT_EXTENSIBLE audio format. Don't fail when it is requested. Initialize required structure fields for it. It already is supported properly by us, since it is just and extended representation of WAVE_FORMAT_PCM.
- Handle some more Mixer control types, to fix sndvol32.exe Mixer open failures after my changes. Remove unneeded duplicating of MIXERCONTROL_CONTROLTYPE_MUX handling.
- Implement MMixerGetWavePosition.

[KS]
- Fix bug in KsStreamIo
Properly set output buffer length in IO Stack Location of the current IRP, since it is passed to KsProbeStreamIrp when calling KsStreamIo, so it fails if the length isn't set properly.
Don't set an input buffer length and the buffer itself, since it isn't passed anywhere, so setting it makes no sense. Moreover, MSDN says that for IOCTL_KS_READ/WRITE_STREAM, only output buffer (and its length) is needed to be set, but not an input one. So it indeed is more correct.
It fixes buffer overflow in KsProbeStreamIrp when attempting to perform the streaming via KsStreamIo. I discovered this bug during my audio refactoring from PR https://github.com/reactos/reactos/pull/4660.
- Restore old hack for MS portcls.

## TODO

- [x] Fix remaining bugs in the code, to make it basically working. Currently, sound is not working properly in most of cases.
- [ ] Probably implement MIDI support.
- [x] Fix a bug in our winmm if possible, preventing the proper work of streaming. Currently, the sound works only with MS winmm.dll, so it is badly needed to be replaced. **UPD: it was a mistake. Actually our winmm is OK. As appeared, the problem was caused by my previous changes in wdmaud.drv. After further changes the streaming works properly, both with MS and our winmm.dll.**
- [x] Revert some unnecessary changes.
- [ ] Remove MMEbuddy library and its depending NT4 SoundBlaster driver and MmeNT4 library, since they are unused anyway and are very incomplete. They seem to be not compatible with NT5, and are not used by our wdmaud.drv or anything other anymore. Do it in a separate PR.
- [x] Send a fix in Kernel Streaming driver, which also can be done in a separate PR (done now in #4663).

### DIFF
--- a/dll/win32/wdmaud.drv/legacy.c
+++ b/dll/win32/wdmaud.drv/legacy.c
@@ -1,89 +1,292 @@
 /*
- * PROJECT:     ReactOS Sound System
- * LICENSE:     GPL - See COPYING in the top level directory
- * FILE:        dll/win32/wdmaud.drv/wdmaud.c
- *
- * PURPOSE:     WDM Audio Driver (User-mode part)
- * PROGRAMMERS: Andrew Greenwood (silverblade@reactos.org)
-                Johannes Anderwald
- *
- * NOTES:       Looking for wodMessage & co? You won't find them here. Try
- *              the MME Buddy library, which is where these routines are
- *              actually implemented.
- *
+ * PROJECT:     ReactOS Audio Subsystem
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     WDM Audio Driver Mapper (User-mode part)
+ * COPYRIGHT:   Copyright 2007-2008 Andrew Greenwood (silverblade@reactos.org)
+ *              Copyright 2009-2010 Johannes Anderwald
+ *              Copyright 2022 Oleg Dubinskiy (oleg.dubinskij30@gmail.com)
  */
 
 #include "wdmaud.h"
 
-#define NDEBUG
+#include <mmixer.h>
+
+#define YDEBUG
 #include <debug.h>
-#include <mmebuddy_debug.h>
 
-#define KERNEL_DEVICE_NAME      L"\\\\.\\wdmaud"
-
+extern MIXER_CONTEXT MixerContext;
 HANDLE KernelHandle = INVALID_HANDLE_VALUE;
 DWORD OpenCount = 0;
 
+/*
+    This is a wrapper around DeviceIoControl which provides control over
+    instantiated sound devices. It waits for I/O to complete (since an
+    instantiated sound device is opened _In_ overlapped mode, this is necessary).
+*/
+MMRESULT
+WdmAudIoControl(
+    _In_ PWDMAUD_DEVICE_INFO DeviceInfo,
+    _In_opt_ ULONG BufferSize,
+    _In_opt_ PVOID Buffer,
+    _In_ DWORD IoControlCode)
+{
+    MMRESULT Result = MMSYSERR_NOERROR;
+    OVERLAPPED Overlapped;
+    DWORD Transferred = 0;
+    BOOL IoResult;
+
+    /* Overlapped I/O is done here - this is used for waiting for completion */
+    ZeroMemory(&Overlapped, sizeof(OVERLAPPED));
+    Overlapped.hEvent = CreateEvent(NULL, FALSE, FALSE, NULL);
+
+    if ( ! Overlapped.hEvent )
+        return Win32ErrorToMmResult(GetLastError());
+
+    if (DeviceInfo->DeviceType != AUX_DEVICE_TYPE &&
+        DeviceInfo->DeviceType != MIXER_DEVICE_TYPE)
+    {
+        EnterCriticalSection(DeviceInfo->DeviceState->QueueCriticalSection);
+    }
+
+    /* Store input data */
+    DeviceInfo->BufferSize = BufferSize;
+    DeviceInfo->Buffer = Buffer;
+
+    /* Talk to the device */
+    IoResult = DeviceIoControl(KernelHandle,
+                               IoControlCode,
+                               DeviceInfo,
+                               sizeof(WDMAUD_DEVICE_INFO),
+                               DeviceInfo,
+                               sizeof(WDMAUD_DEVICE_INFO),
+                               &Transferred,
+                               &Overlapped);
+
+    /* If failure occurs, make sure it's not just due to the overlapped I/O */
+    if ( ! IoResult )
+    {
+        if (GetLastError() == ERROR_IO_PENDING)
+        {
+            /* Wait for I/O complete */
+            WaitForSingleObject(Overlapped.hEvent, INFINITE);
+        }
+        Result = Win32ErrorToMmResult(GetLastError());
+    }
+
+    if (DeviceInfo->DeviceType != AUX_DEVICE_TYPE &&
+        DeviceInfo->DeviceType != MIXER_DEVICE_TYPE)
+    {
+        LeaveCriticalSection(DeviceInfo->DeviceState->QueueCriticalSection);
+    }
+
+    /* Don't need this any more */
+    CloseHandle(Overlapped.hEvent);
+
+    DPRINT("Transferred %d bytes in Sync overlapped I/O\n", Transferred);
+
+    return Result;
+}
+
 DWORD
 WINAPI
-MixerEventThreadRoutine(
+WaveThreadRoutine(
     LPVOID Parameter)
 {
-    HANDLE WaitObjects[2];
-    DWORD dwResult;
+    PWDMAUD_DEVICE_INFO DeviceInfo = (PWDMAUD_DEVICE_INFO)Parameter;
+    PWAVEHDR_EXTENSION HeaderExtension;
+    DWORD dwResult = WAIT_FAILED;
+    PWAVEHDR WaveHeader;
+#ifndef USE_MMIXER_LIB
     MMRESULT Result;
-    WDMAUD_DEVICE_INFO DeviceInfo;
-    PSOUND_DEVICE_INSTANCE Instance = (PSOUND_DEVICE_INSTANCE)Parameter;
+#endif
+    HANDLE hEvent;
 
-    /* setup wait objects */
-    WaitObjects[0] = Instance->hNotifyEvent;
-    WaitObjects[1] = Instance->hStopEvent;
-
-    /* zero device info */
-    ZeroMemory(&DeviceInfo, sizeof(WDMAUD_DEVICE_INFO));
-
-    DeviceInfo.hDevice = Instance->Handle;
-    DeviceInfo.DeviceType = MIXER_DEVICE_TYPE;
-
-    do
+    while (TRUE)
     {
-        dwResult = WaitForMultipleObjects(2, WaitObjects, FALSE, INFINITE);
-
-        if (dwResult == WAIT_OBJECT_0 + 1)
+        DPRINT("WaveQueue %p\n", DeviceInfo->DeviceState->WaveQueue);
+        EnterCriticalSection(DeviceInfo->DeviceState->QueueCriticalSection);
+        WaveHeader = DeviceInfo->DeviceState->WaveQueue;
+        if (WaveHeader && (WaveHeader->dwFlags & (WHDR_BEGINLOOP | WHDR_DONE |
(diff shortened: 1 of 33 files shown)


JoachimHenze:
Doesn't build on MSVC yet. Regarding "Implement proper resampling support", Does it fix playback of the 22kHz mp3 in https://jira.reactos.org/browse/CORE-17744 ?

oleg-dubinskiy:
> Doesn't build on MSVC yet. Regarding "Implement proper resampling support", Does it fix playback of the 22kHz mp3 in https://jira.reactos.org/browse/CORE-17744 ?

Not yet. The resampling should work after successful opening audio device, when the playback starts. But the problem is, the device fails to open if the format is unsupported. This happens even after such major refactoring. So not fixed yet, sorry. :slightly_frowning_face:

dll/win32/wdmaud.drv/legacy.c
```diff
@@ -1,89 +1,306 @@
...
+                               sizeof(WDMAUD_DEVICE_INFO),
+                               DeviceInfo,
+                               sizeof(WDMAUD_DEVICE_INFO),
+                               &Transferred,
+                               &Overlapped);
+
+    /* If failure occurs, make sure it's not just due to the overlapped I/O */
+    if ( ! IoResult )
+    {
+        if (GetLastError() == ERROR_IO_PENDING)
```
ThFabba, line 71: `GetOverlappedResult` might be more appropriate than hand-rolling this? In particular, the `Transferred` variable will not currently be set correctly in the async case.

dll/win32/wdmaud.drv/legacy.c
```diff
@@ -1,89 +1,306 @@
...
+    Overlapped.hEvent = CreateEvent(NULL, FALSE, FALSE, NULL);
+
+    if ( ! Overlapped.hEvent )
+        return Win32ErrorToMmResult(GetLastError());
+
+    if (DeviceInfo->DeviceType != AUX_DEVICE_TYPE &&
+        DeviceInfo->DeviceType != MIXER_DEVICE_TYPE)
+    {
+        /* Enter critical section */
+        EnterCriticalSection(&WdmAudCriticalSection);
```
ThFabba, line 51: What exactly does this critical section protect? You seem to acquire it for just about everything.

dll/win32/wdmaud.drv/legacy.c
```diff
@@ -1,89 +1,306 @@
...
+                Result = WdmAudIoControl(DeviceInfo,
+                                         0,
+                                         NULL,
+                                         DeviceInfo->DeviceType == WAVE_OUT_DEVICE_TYPE ?
+                                         IOCTL_PAUSE_PLAYBACK : IOCTL_PAUSE_CAPTURE);
+                if (!MMSUCCESS(Result))
+                {
+                    DPRINT1("Call to %ls failed with %d\n",
+                            DeviceInfo->DeviceType == WAVE_OUT_DEVICE_TYPE ?
+                            L"IOCTL_PAUSE_PLAYBACK" : L"IOCTL_PAUSE_CAPTURE",
```
ThFabba, line 174: Using wide strings here isn't ideal, DbgPrint is natively ansi

dll/win32/wdmaud.drv/legacy.c
```diff
@@ -1,89 +1,306 @@
...
-    PSOUND_DEVICE_INSTANCE Instance = (PSOUND_DEVICE_INSTANCE)Parameter;
+#endif
+
+    /* Enter critical section */
+    EnterCriticalSection(&WdmAudCriticalSection);
+
+    while(TRUE)
+    {
+        DPRINT("WaveQueue %p\n", DeviceInfo->DeviceState.WaveQueue);
+        if (DeviceInfo->DeviceState.WaveQueue)
```
ThFabba, line 113: You're checking this again below (L116), so it seems unnecessary here.

dll/win32/wdmaud.drv/legacy.c
```diff
@@ -1,89 +1,306 @@
...
+            /* Done */
+            break;
+        }
+    }
 
-    do
+    CloseHandle(DeviceInfo->DeviceState.hNotifyEvent);
+    DeviceInfo->DeviceState.hNotifyEvent = NULL;
+    DeviceInfo->DeviceState.bStartInThread = FALSE;
+    SetEvent(DeviceInfo->DeviceState.hStopEvent);
```
ThFabba, line 196: This event seems misleading/unnecessary. You generally want to wait on the thread handle instead, which will signal that the thread actually stopped running.

dll/win32/wdmaud.drv/legacy.c
```diff
@@ -95,175 +312,130 @@ WdmAudCleanupByLegacy()
...
+    DPRINT("WDMAUD - AddRemoveDeviceNode DeviceType %u\n", DeviceType);
 
-    Result = SyncOverlappedDeviceIoControl(KernelHandle,
-                                           IOCTL_GETNUMDEVS_TYPE,
-                                           (LPVOID) &DeviceInfo,
-                                           sizeof(WDMAUD_DEVICE_INFO),
-                                           (LPVOID) &DeviceInfo,
-                                           sizeof(WDMAUD_DEVICE_INFO),
-                                           NULL);
+    DeviceInfo = HeapAlloc(GetProcessHeap(), HEAP_ZERO_MEMORY, sizeof(WDMAUD_DEVICE_INFO));
```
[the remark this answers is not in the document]
oleg-dubinskiy: As I tested, just zeroing memory without allocating a pointer (like `WDMAUD_DEVICE_INFO DeviceInfo; ZeroMemory(&DeviceInfo, sizeof(WDMAUD_DEVICE_INFO);`) doesn't change anything, so it doesn't matter what to use from the technical point. But regarding to analysis, Windows does the allocation in all cases when a new structure pointer is needed to be used, so I prefer to do the same. :slightly_smiling_face:

dll/win32/wdmaud.drv/legacy.c
oleg-dubinskiy: This call existed from the start in my code, but I've removed it, because it creates an endless wait with deadlocks, so initialization hangs. Moreover, Windows does not do it. I don't see any `GetOverlappedResult` call in an unassemble of `wdmaudIoControl`. However, that routine does the wait only if `bWait` is `TRUE`. If I'll set it to 'FALSE`, then waiting will not be performed, and the problem will be gone, **but does calling it make any sense then?** :thinking:

oleg-dubinskiy: Regarding `Transferred` variable, your guessing is correct, but if it causes such problems, then maybe I can to not use it at all, because it is used only for a debug print to view the passed buffer size... :thinking:

dll/win32/wdmaud.drv/legacy.c
oleg-dubinskiy: I did this previously because Windows XP/2003 also manages a critical section in wdmaud.drv. But if it's not necessarily, I removed it now. :slightly_smiling_face:

dll/win32/wdmaud.drv/legacy.c
oleg-dubinskiy: Okay, changed all strings to ANSI here and in all other places. :slightly_smiling_face:

dll/win32/wdmaud.drv/legacy.c
oleg-dubinskiy: Fixed, thanks. :+1:

dll/win32/wdmaud.drv/legacy.c
```diff
@@ -1,89 +1,306 @@
...
+                {
+                    DPRINT1("Call to %ls failed with %d\n",
+                            DeviceInfo->DeviceType == WAVE_OUT_DEVICE_TYPE ?
+                            L"IOCTL_PAUSE_PLAYBACK" : L"IOCTL_PAUSE_CAPTURE",
+                            GetLastError());
+                    break;
+                }
+#endif
+                DeviceInfo->DeviceState.bStart = FALSE;
+		    }
```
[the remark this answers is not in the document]
oleg-dubinskiy: Fixed the indentation, thanks.

oleg-dubinskiy:
Disabled Wave Header Extension management for now and added a workaround, because I still can't figure out why the correct code does not work properly. :slightly_frowning_face:  But with the current changes, streaming works fine. WaveOut playback is nearly fine, DirectSound badly needs MS portcls replacement, and works even a bit better than WaveOut with replaced driver. :slightly_smiling_face:

PR #4662, opened 2022-09-03
[NETSHELL] Fix "ghost" network activity when opening Network Status page

KRosUser:
## Purpose

Opening Network Status page creates a systematic "ghost" network activity, activating RX+TX activity icon both on property page and tray, while no Rx/Tx.
This is due to pContext->dwIn{out}Octets being initialized to 0 and compared to refreshed interface data.

## Proposed changes

On first Update, copy refreshed interface data to the Context

### DIFF
--- a/dll/shellext/netshell/lanstatusui.cpp
+++ b/dll/shellext/netshell/lanstatusui.cpp
@@ -15,7 +15,6 @@ CLanStatus::CLanStatus() :
 {
 }
 
-BOOL bFirst;
 
 VOID
 UpdateLanStatusUiDlg(
@@ -132,7 +131,7 @@ UpdateLanStatus(HWND hwndDlg,  LANSTATUSUI_CONTEXT * pContext)
         return;
     }
 
-    if(bFirst)
+    if(pContext->Status == (UINT)-1)
     {
         /*
         On first execution, pContext->dwIn{out}Octets will be null while IF data are already refreshed with non null data so a gap is normal
@@ -140,7 +139,6 @@ UpdateLanStatus(HWND hwndDlg,  LANSTATUSUI_CONTEXT * pContext)
         */
         pContext->dwInOctets = IfEntry.dwInOctets;
         pContext->dwOutOctets = IfEntry.dwOutOctets;
-        bFirst = FALSE;
     }
 
     hIcon = NULL;
@@ -276,7 +274,6 @@ InitializeLANStatusUiDlg(HWND hwndDlg, LANSTATUSUI_CONTEXT * pContext)
 
     /* update adapter info */
     pContext->Status = -1;
-    bFirst = TRUE;
     UpdateLanStatus(hwndDlg, pContext);
     NcFreeNetconProperties(pProperties);
 }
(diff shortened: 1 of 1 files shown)


dll/shellext/netshell/lanstatusui.cpp
```diff
@@ -130,6 +132,17 @@ UpdateLanStatus(HWND hwndDlg,  LANSTATUSUI_CONTEXT * pContext)
         return;
     }
 
+    if(bFirst)
```
HBelusca, line 135: I see that you can determine whether you're in "first" case by just looking at the `pContext->Status` value, that conveniently has the value `-1` at the first call (see where you set bFirst to TRUE below).
This would allow not using this ugly global variable ^^

KRosUser: Nice idea ! I'll do it this way

dll/shellext/netshell/lanstatusui.cpp
```diff
@@ -130,6 +130,16 @@ UpdateLanStatus(HWND hwndDlg,  LANSTATUSUI_CONTEXT * pContext)
         return;
     }
 
+    if(pContext->Status == (UINT)-1)
```
HBelusca, line 133:
```suggestion
    if (pContext->Status == (UINT)-1)
```

dll/shellext/netshell/lanstatusui.cpp
```diff
@@ -130,6 +130,16 @@ UpdateLanStatus(HWND hwndDlg,  LANSTATUSUI_CONTEXT * pContext)
...
+    {
+        /*
+        On first execution, pContext->dwIn{out}Octets will be null while IF data are already refreshed with non null data so a gap is normal
+        and does not corresponds to an effective packet tx or rx
+        */
```
HBelusca, line 138:
```suggestion
        /*
         * On first execution, pContext->dwIn{out}Octets will be null while IF data is already refreshed
         * with non-null data, so a gap is normal and does not correspond to an effective tx or rx packet.
         */
```

SergeGautherie: Moreover:
* dwIn{out}Octets: it's In and Out, not In and Inout.
* null: '0' rather.

dll/shellext/netshell/lanstatusui.cpp
KRosUser: ok

dll/shellext/netshell/lanstatusui.cpp
KRosUser: OK

KRosUser:
@SergeGautherie @HBelusca : Comment updated accordingly.
OK to go ?

dll/shellext/netshell/lanstatusui.cpp
```diff
@@ -130,6 +130,16 @@ UpdateLanStatus(HWND hwndDlg,  LANSTATUSUI_CONTEXT * pContext)
         return;
     }
 
+    if (pContext->Status == (UINT)-1)
+    {
+        /*
+        On first execution, pContext->dwInOctets/dwInOctets will be at 0 while interface data is already refreshed with non-null data
+        so a gap is normal and does not correspond to an effective tx or rx packet
+        */
```
HBelusca, line 138:
```suggestion
        /*
         * On first execution, pContext->dw[In|Out]Octets will be zero while
         * the interface info is already refreshed with non-null data, so a
         * gap is normal and does not correspond to an effective TX or RX packet.
         */
```

PR #4663, opened 2022-09-03
[KS] Fix bug in KsStreamIo

oleg-dubinskiy:
## Purpose

Properly set output buffer length in IO Stack Location of the current IRP, since it is passed to KsProbeStreamIrp when calling KsStreamIo, so it fails if the length isn't set properly.
Don't set an input buffer length and the buffer itself, since it isn't passed anywhere, so setting it makes no sense. Moreover, MSDN says that for IOCTL_KS_READ/WRITE_STREAM, only output buffer (and its length) is needed to be set, but not an input one. So it indeed is more correct.
It fixes buffer overflow in KsProbeStreamIrp when attempting to perform the streaming via KsStreamIo. I discovered this bug during my audio refactoring from PR #4660.

JIRA issue: None.

### DIFF
--- a/drivers/ksfilter/ks/irp.c
+++ b/drivers/ksfilter/ks/irp.c
@@ -634,8 +634,7 @@ KsStreamIo(
     IoStack = IoGetNextIrpStackLocation(Irp);
     /* setup stack parameters */
     IoStack->FileObject = FileObject;
-    IoStack->Parameters.DeviceIoControl.InputBufferLength = Length;
-    IoStack->Parameters.DeviceIoControl.Type3InputBuffer = StreamHeaders;
+    IoStack->Parameters.DeviceIoControl.OutputBufferLength = Length;
     IoStack->Parameters.DeviceIoControl.IoControlCode = (Flags == KSSTREAM_READ ? IOCTL_KS_READ_STREAM : IOCTL_KS_WRITE_STREAM);
 
     if (CompletionRoutine)


GeoB99:
I reopened the PR back and ran a testbot against it.
https://build.reactos.org/#/builders/9/builds/33390

By the way could you please show me where is that `"no input buffer needed"` stated? Because https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ks/ni-ks-ioctl_ks_read_stream seems to tell otherwise.
```
Input buffer
The application places a pointer to an array of [KSSTREAM_HEADER](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ks/ns-ks-ksstream_header) structures in the InBuffer parameter. Each describes a packet of data to be read. The headers can also contain the presentation time and duration.

Input buffer length
The InLength parameter specifies the size, in bytes, of the input buffer contents.
```

GeoB99 (approved):

ThFabba:
I find this very hard to follow. `IoBuildSynchronousFsdRequest` sets the buffers in the IRP based on the DeviceObject's `DO_XXX_IO` flags. This is totally inappropriate for IOCTLs since the correct I/O mode is determined by the CtlCode in that case.
So the IRP we get has some random irrelevant buffer member set, then we set the correct one in addition to that.

Any reason we can't just switch this code to `IoBuildDeviceIoControlRequest` and make it explicit that we want an input buffer of NULL and to only set the output buffer?

PR #4682, opened 2022-09-12
[NTFS] PrintAllVCNs(): Fix a typo in an ASSERT()

SergeGautherie:
Obvious typo.

### DIFF
--- a/drivers/filesystems/ntfs/btree.c
+++ b/drivers/filesystems/ntfs/btree.c
@@ -56,7 +56,7 @@ PrintAllVCNs(PDEVICE_EXTENSION Vcb,
 
     BytesRead = ReadAttribute(Vcb, IndexAllocationContext, 0, (PCHAR)Buffer, BufferSize);
 
-    ASSERT(BytesRead = BufferSize);
+    ASSERT(BytesRead == BufferSize);
 
     CurrentNode = Buffer;
 


drivers/filesystems/ntfs/btree.c
```diff
@@ -56,7 +56,7 @@ PrintAllVCNs(PDEVICE_EXTENSION Vcb,
 
     BytesRead = ReadAttribute(Vcb, IndexAllocationContext, 0, (PCHAR)Buffer, BufferSize);
 
-    ASSERT(BytesRead = BufferSize);
+    ASSERT(BytesRead == BufferSize);
```
binarymaster, line 59: This was added by @ThFabba in 5579428b.

learn-more: Trevor Thompson actually.

learn-more (approved):

drivers/filesystems/ntfs/btree.c
JoachimHenze: ouch, introduced in 0.4.7-dev-590-g5579428. I will put that on my list for backmerging. Good catch Serge!

JoachimHenze (approved):

drivers/filesystems/ntfs/btree.c
JoachimHenze: It luckily has no real world impact because the local variable BytesRead is currently not used afterwards anymore.

PR #4675, opened 2022-09-10
[SHELL32] Fix Assertion when opening "File" menu in MyComputer twice

KRosUser:
## Purpose

_ ReactOS asserts when opening File menu in MyComputer for the second time
image
_ This is due to the cached copy of IContextMenu not being released before attempting to cache it again.

JIRA issue: [CORE-18353](https://jira.reactos.org/browse/CORE-18353)

## Proposed changes

_ Fix based on the patch proposed by user https://jira.reactos.org/secure/ViewProfile.jspa?name=I_Kill_Bugs

PR #4683, opened 2022-09-12
[SHELL32] Fix Assert when hitting DEL or F2 in Recycle Bin (or copying files)

KRosUser:
Due to improper return without calling "cleanup" code

## Purpose

_ Remplace incorrect "return" call to call for "cleanup" section

JIRA issue: [CORE-18361](https://jira.reactos.org/browse/CORE-18361)
JIRA issue: [CORE-18345](https://jira.reactos.org/browse/CORE-18345)
JIRA issue: [CORE-18321](https://jira.reactos.org/browse/CORE-18321)

Next steps will be to implement Delete (and fix CM Deleted action on multiple files as per CORE-18359) action using keyboard "del" key which is not supported (currently asserts without action) in **another** PR.

### DIFF
--- a/dll/win32/shell32/CDefView.cpp
+++ b/dll/win32/shell32/CDefView.cpp
@@ -1656,14 +1656,21 @@ LRESULT CDefView::OnExplorerCommand(UINT uCommand, BOOL bUseSelection)
     HRESULT hResult;
     HMENU hMenu = NULL;
 
+    if (m_pCM)
+    {
+        IUnknown_SetSite(m_pCM, NULL);
+        m_pCM.Release();
+    }
+
     hResult = GetItemObject( bUseSelection ? SVGIO_SELECTION : SVGIO_BACKGROUND, IID_PPV_ARG(IContextMenu, &m_pCM));
     if (FAILED_UNEXPECTEDLY( hResult))
         goto cleanup;
+
     if ((uCommand != FCIDM_SHVIEW_DELETE) && (uCommand != FCIDM_SHVIEW_RENAME))
     {
         hMenu = CreatePopupMenu();
         if (!hMenu)
-            return 0;
+            goto cleanup;
 
         hResult = m_pCM->QueryContextMenu(hMenu, 0, FCIDM_SHVIEWFIRST, FCIDM_SHVIEWLAST, CMF_NORMAL);
         if (FAILED_UNEXPECTEDLY(hResult))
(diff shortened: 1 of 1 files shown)


learn-more:
I am not sure this is the proper solution.
Some other code paths leave the `m_pCM` filled in,
so maybe the actual fix would be cleaning up the object before initializing.....

KRosUser:
> I am not sure this is the proper solution. Some other code paths leave the `m_pCM` filled in, so maybe the actual fix would be cleaning up the object before initializing.....

That's indeed the way it has been managed for https://github.com/reactos/reactos/pull/4675 
I'll propose a change with this design pattern

learn-more (approved):

KRosUser:
@learn-more : I've placed the cleanup BEFORE the call to GetItemObject like we did on https://github.com/reactos/reactos/pull/4675 but I also left it in the cleanup code in the end, not assuming what other function would do. Would you think this make sense ? Or should I remove the "trailing" cleanup now that's it's art the beginning ?

Btw this "cleanup" is missing in 2 other functions (unrelated to the bug I propose to fix) : OpenSelectedItems and OnContextMenu. Would you like me to add this there too as defensive programming ?

learn-more:
> @learn-more : I've placed the cleanup BEFORE the call to GetItemObject like we did on #4675 but I also left it in the cleanup code in the end, not assuming what other function would do. Would you think this make sense ? Or should I remove the "trailing" cleanup now that's it's art the beginning ?
> 
> Btw this "cleanup" is missing in 2 other functions (unrelated to the bug I propose to fix) : OpenSelectedItems and OnContextMenu. Would you like me to add this there too as defensive programming ?

That this needs copy-pasting code in multiple functions is a code smell, so I'd rather not that you make matters worse.
Instead, investigate if this object should be a class variable at all, or that it should be initialized with a helper instead that does cleanup before initializing.

KRosUser:
> > @learn-more : I've placed the cleanup BEFORE the call to GetItemObject like we did on #4675 but I also left it in the cleanup code in the end, not assuming what other function would do. Would you think this make sense ? Or should I remove the "trailing" cleanup now that's it's art the beginning ?
> > Btw this "cleanup" is missing in 2 other functions (unrelated to the bug I propose to fix) : OpenSelectedItems and OnContextMenu. Would you like me to add this there too as defensive programming ?
> 
> That this needs copy-pasting code in multiple functions is a code smell, so I'd rather not that you make matters worse. Instead, investigate if this object should be a class variable at all, or that it should be initialized with a helper instead that does cleanup before initializing.

OK understood. So is current PR "as is" OK for you ?

KRosUser:
Now fully superseeded by https://github.com/reactos/reactos/pull/4702

PR #4686, opened 2022-09-13
[USER32] Load keyboard layouts on logon correctly

katahiromz:
## Purpose
JIRA issue: [CORE-16600](https://jira.reactos.org/browse/CORE-16600)

## Proposed changes

- Add `IntLoadPreloadKeyboardLayouts` helper function to load the layouts on log-on.
- In `UpdatePerUserSystemParameters` function, call `CliImmInitializeHotKeys` and `IntLoadPreloadKeyboardLayouts` functions.

## TODO

- [x] Do tests.

### DIFF
--- a/win32ss/user/user32/misc/logon.c
+++ b/win32ss/user/user32/misc/logon.c
@@ -7,6 +7,7 @@
  */
 
 #include <user32.h>
+#include <strsafe.h>
 
 WINE_DEFAULT_DEBUG_CHANNEL(user32);
 
@@ -83,6 +84,69 @@ Logon(BOOL IsLogon)
                          sizeof(*LogonRequest));
 }
 
+/* Win: LoadPreloadKeyboardLayouts */
+BOOL IntLoadPreloadKeyboardLayouts(VOID)
+{
+    UINT nNumber, uFlags;
+    DWORD cbValue, dwType;
+    WCHAR szNumber[32], szValue[9];
+    HKEY hPreloadKey;
+    BOOL ret = FALSE;
+    HKL hKL, hDefaultKL = NULL;
+
+    if (RegOpenKeyW(HKEY_CURRENT_USER,
+                    L"Keyboard Layout\\Preload",
+                    &hPreloadKey) != ERROR_SUCCESS)
+    {
+        return FALSE;
+    }
+
+    for (nNumber = 1; nNumber <= 1000; ++nNumber)
+    {
+        StringCchPrintfW(szNumber, _countof(szNumber), L"%u", nNumber);
+
+        cbValue = sizeof(szValue);
+        if (RegQueryValueExW(hPreloadKey,
+                             szNumber,
+                             NULL,
+                             &dwType,
+                             (LPBYTE)szValue,
+                             &cbValue) != ERROR_SUCCESS)
+        {
+            break;
+        }
+
+        if (dwType != REG_SZ)
+            continue;
+
+        if (nNumber == 1)
+            uFlags = KLF_SUBSTITUTE_OK | KLF_ACTIVATE | KLF_RESET;
+        else
+            uFlags = KLF_SUBSTITUTE_OK | KLF_NOTELLSHELL | KLF_REPLACELANG;
+
+        hKL = LoadKeyboardLayoutW(szValue, uFlags);
+        if (hKL)
+        {
+            ret = TRUE;
+            if (nNumber == 1)
+                hDefaultKL = hKL;
+        }
+    }
+
+    RegCloseKey(hPreloadKey);
+
+    if (hDefaultKL)
+        SystemParametersInfoW(SPI_SETDEFAULTINPUTLANG, 0, &hDefaultKL, 0);
+
+    if (!ret)
+    {
+        /* Default to USA */
+        LoadKeyboardLayoutW(L"04090409", KLF_SUBSTITUTE_OK | KLF_ACTIVATE | KLF_RESET);
+    }
+
+    return ret;
+}
+
 /*
  * @implemented
  */
@@ -103,6 +167,12 @@ UpdatePerUserSystemParameters(DWORD dwReserved,
 {
     // Update Imm support and load Imm32.dll.
     UpdatePerUserImmEnabling();
+
+    /* Delete all the IME hotkeys */
+    CliImmInitializeHotKeys(SETIMEHOTKEY_DELETEALL, NULL);
+
+    /* Load Preload keyboard layouts */
+    IntLoadPreloadKeyboardLayouts();
 
     return NtUserUpdatePerUserSystemParameters(dwReserved, bEnable);
 }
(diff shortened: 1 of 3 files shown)


katahiromz:
Pre-Reboot:
prereboot
I added French and german. And I made French default.

Post-Reboot
postreboot
Correctly restored.

HBelusca:
@katahiromz (Something for a possible separate PR): Could you check why manipulating the layouts would require a reboot?

win32ss/user/user32/misc/logon.c
```diff
@@ -83,6 +84,69 @@ Logon(BOOL IsLogon)
                          sizeof(*LogonRequest));
 }
 
+/* Win: LoadPreloadKeyboardLayouts */
+BOOL IntLoadPreloadKeyboardLayouts(VOID)
+{
+    UINT nNumber, uFlags;
+    DWORD cbValue, dwType;
+    WCHAR szNumber[32], szValue[9];
```
HBelusca, line 92: 9 is certainly `KL_NAMELENGTH` ?

win32ss/user/user32/misc/logon.c
```diff
@@ -83,6 +84,69 @@ Logon(BOOL IsLogon)
...
+                             (LPBYTE)szValue,
+                             &cbValue) != ERROR_SUCCESS)
+        {
+            break;
+        }
+
+        if (dwType != REG_SZ)
+            continue;
+
+        if (nNumber == 1)
```
HBelusca, line 122: Having a comment explaining what this particular case is and why we need different flags here, would be useful.

win32ss/user/user32/misc/logon.c
```diff
@@ -83,6 +84,69 @@ Logon(BOOL IsLogon)
...
+            uFlags = KLF_SUBSTITUTE_OK | KLF_ACTIVATE | KLF_RESET;
+        else
+            uFlags = KLF_SUBSTITUTE_OK | KLF_NOTELLSHELL | KLF_REPLACELANG;
+
+        hKL = LoadKeyboardLayoutW(szValue, uFlags);
+        if (hKL)
+        {
+            ret = TRUE;
```
HBelusca, line 130: It looks like one could get rid of that "ret" variable, and: 1. initialize hKL originally to NULL (where it's defined), and 2. replace all those "ret" checks by suitable checks on the value of hKL; finally, 3. replacing `return ret;` by `return (hKL != NULL);`

win32ss/user/user32/misc/logon.c
```diff
@@ -83,6 +84,69 @@ Logon(BOOL IsLogon)
...
+    if (RegOpenKeyW(HKEY_CURRENT_USER,
+                    L"Keyboard Layout\\Preload",
+                    &hPreloadKey) != ERROR_SUCCESS)
+    {
+        return FALSE;
+    }
+
+    for (nNumber = 1; nNumber <= 1000; ++nNumber)
+    {
+        StringCchPrintfW(szNumber, _countof(szNumber), L"%u", nNumber);
```
HBelusca, line 106: Shouldn't using `_ultow` be enough? With your 32-character-sized `szNumber` buffer, there is no way the function could ever do a buffer overrun on it, unless that function itself is buggy (which it shouldn't be).

katahiromz:
> Could you check why manipulating the layouts would require a reboot?

The keyboard layout can be loaded by `LoadKeyboardLayout...` functions. The keyboard layout can be unloaded by `UnloadKeyboardLayout` function. The loaded layouts are stored as the linked list of `gspklBaseLayout` of the kernel subsystem. The user-side application can access the layout list by calling `GetKeyboardLayoutList` function. We can change the input language of the windows by `WM_INPUTLANGCHANGEREQUEST` message broadcasting. The default keyboard layout can be changed `SystemParametersInfo.SPI_SETDEFAULTINPUTLANG`.

I think rebooting is not required to change the layout list.

win32ss/user/user32/misc/logon.c
katahiromz: Used `_ultow` function.

win32ss/user/user32/misc/logon.c
katahiromz: Fixed.

win32ss/user/user32/misc/logon.c
katahiromz: Added a comment.

HBelusca:
> I think rebooting is not required to change the layout list.

One example why requiring reboot to change the layouts is problematic is, for example, during ROS installation. For example, you install ROS in some language (say, japanese), but you don't have the corresponding ideal keyboard layout, and instead you need to use a german keyboard layout to navigate in the installation process (and to use ROS later on). In the installer you can change the keyboard layout, but currently it would ask you to reboot, which is obviously ridiculous at this point, since this would signify you're also aborting the installation.

win32ss/user/user32/misc/logon.c
HBelusca: > I can't agree that. How to check `hKL` without assignment?

> 1. initialize hKL originally to NULL (where it's defined)

i.e. `HKL hKL = NULL, hDefaultKL = NULL;` that is at the top of the function. Then the hKL variable is always correctly defined: to NULL if nothing has been loaded, or to a successfully loaded layout.

katahiromz: Difficult to see your opinion.

- There can be zero, single or multiple layout(s).
- `LoadKeyboardLayoutW` can be called in multiple times. Not just once.

KRosUser:
Looking at the task bar it seems that the icon is incorrect. Do you confirm ?

katahiromz:
> it seems that the icon is incorrect.

I made French default.

katahiromz:
> I meant task bar not tray icon

Surely. But unrelated to this PR. I will fix that later.

KRosUser:
> > I meant task bar not tray icon
> 
> Surely. But unrelated to this PR. I will fix that later.

Ok, super clear. Thanks!

win32ss/user/user32/misc/logon.c
HBelusca: > LoadKeyboardLayoutW can be called in multiple times. Not just once.

Oh indeed! So you would return TRUE if at least one LoadKeyboardLayoutW call succeed? (even if some others didn't?) If so, then my proposition doesn't make sense, and indeed you should keep the separate `ret` variable.

katahiromz: The return value is not important because the caller ignores it.

HBelusca: Then why is there a `ret` variable altogether? The IntLoadPreloadKeyboardLayouts should then just be "void" :smile:

katahiromz: Now return void. OK.

katahiromz:
@KRosUser FYI:
https://github.com/reactos/reactos/commit/976d48baeb1a98d620107885022b44b24bad77a1
The icon is fixed.

PR #4688, opened 2022-09-14
[INPUT] Fix taskbar icon and remove small icon hack

TAN-Gaming:
## Proposed changes
- Use `int` for applet index (like other CPLs)
- Revert some changes from https://github.com/reactos/reactos/commit/976d48baeb1a98d620107885022b44b24bad77a1

### DIFF
--- a/dll/cpl/input/input.c
+++ b/dll/cpl/input/input.c
@@ -39,22 +39,13 @@ static int CALLBACK
 PropSheetProc(HWND hwndDlg, UINT uMsg, LPARAM lParam)
 {
     // NOTE: This callback is needed to set large icon correctly.
-    // FIXME: Handle leaks
-    HICON hIcon, hIconSmall;
+    HICON hIcon;
     switch (uMsg)
     {
         case PSCB_INITIALIZED:
         {
             hIcon = LoadIconW(hApplet, MAKEINTRESOURCEW(IDI_CPLSYSTEM));
             SendMessageW(hwndDlg, WM_SETICON, ICON_BIG, (LPARAM)hIcon);
-
-            hIconSmall = (HICON)LoadImageW(hApplet,
-                                           MAKEINTRESOURCEW(IDI_CPLSYSTEM),
-                                           IMAGE_ICON,
-                                           GetSystemMetrics(SM_CXSMICON),
-                                           GetSystemMetrics(SM_CYSMICON),
-                                           0);
-            SendMessageW(hwndDlg, WM_SETICON, ICON_SMALL, (LPARAM)hIconSmall);
             break;
         }
     }
@@ -96,9 +87,9 @@ LONG CALLBACK
 CPlApplet(HWND hwndCPl, UINT uMsg, LPARAM lParam1, LPARAM lParam2)
 {
     CPLINFO *CPlInfo;
-    DWORD i;
+    int i;
 
-    i = (DWORD)lParam1;
+    i = (int)lParam1;
 
     switch (uMsg)
     {


dll/cpl/input/input.c
```diff
@@ -39,22 +39,13 @@ static int CALLBACK
...
             hIcon = LoadIconW(hApplet, MAKEINTRESOURCEW(IDI_CPLSYSTEM));
             SendMessageW(hwndDlg, WM_SETICON, ICON_BIG, (LPARAM)hIcon);
-
-            hIconSmall = (HICON)LoadImageW(hApplet,
-                                           MAKEINTRESOURCEW(IDI_CPLSYSTEM),
-                                           IMAGE_ICON,
-                                           GetSystemMetrics(SM_CXSMICON),
-                                           GetSystemMetrics(SM_CYSMICON),
-                                           0);
-            SendMessageW(hwndDlg, WM_SETICON, ICON_SMALL, (LPARAM)hIconSmall);
```
learn-more, line 57: Why did you revert this?

JoachimHenze: This looks like it was not intentionally reverted, but was a merge-oopsie.
Those code-pieces was merged a day ago just and was committed by Katayama in
https://git.reactos.org/?p=reactos.git;a=commit;h=976d48baeb1a98d620107885022b44b24bad77a1

JoachimHenze (changes requested):
Please restore Katayamas code!

dll/cpl/input/input.c
TAN-Gaming: This is not needed. All other CPLs that use `PROPSHEETHEADER` don't do this.
The small icon is already set by `PROPSHEETHEADER` structure.

The problem comes from the applet index.

TAN-Gaming: Not sure why that happens. Maybe because the index returned by `CPL_INQUIRE` can be a nagative value?

TAN-Gaming:
VirtualBox_ReactOS_14_09_2022_19_13_29

It works, after changing the applet index data type.

dll/cpl/input/input.c
katahiromz: @TAN-Gaming saying is correct.
https://github.com/reactos/reactos/blob/3647f6a5eb633b52ef4bf1db6e43fc2b3fc72969/dll/win32/comctl32/propsheet.c#L3553

```c
SendMessageW(hwnd, WM_SETICON, 0, (LPARAM)hIcon);
```
and
```c
#define ICON_SMALL             0
```

katahiromz (approved):

dll/cpl/input/input.c
JoachimHenze: If you find a solution that does satisfy you and Katayama, and the icon remains in working state after that, then I am ofc fine with whatever you'll decide together.
>>The problem comes from the applet index.
I don't understand.
Does the icon reliably show with the code as it currently is in your PR?

katahiromz: If the applet index is negative, then what will happen?
Why is the applet index negative?

katahiromz: BEFORE and AFTER screenshots please.

katahiromz (changes requested):
The difference between `INT` and `DWORD` is signedness.
Thus, we have to consider the negative value.

dll/cpl/input/input.c
```diff
@@ -96,9 +87,9 @@ LONG CALLBACK
 CPlApplet(HWND hwndCPl, UINT uMsg, LPARAM lParam1, LPARAM lParam2)
 {
     CPLINFO *CPlInfo;
-    DWORD i;
+    int i;
 
-    i = (DWORD)lParam1;
+    i = (int)lParam1;
```
katahiromz, line 92: The variable `i` should be in `[0, CPL_GETCOUNT – 1]`.
(See `CPL_INQUIRE` and `CPL_DBLCLK`)
And `CPL_GETCOUNT` is one.
Thus, the variable `i` should be one.

dll/cpl/input/input.c
TAN-Gaming: @JoachimHenze 

> Does the icon reliably show with the code as it currently is in your PR?

Yes, other CPLs also use `int`.

dll/cpl/input/input.c
JoachimHenze: We still don't have a before-picture, that was shot with Katayamas code. Does it show the exact same icon or another one?

binarymaster: @JoachimHenze the "before" picture was posted by Katayama in comments of recent commit: https://user-images.githubusercontent.com/2107452/189984046-b5c516ed-6a37-47ad-a40e-a0c79651cebc.png

There should be no visual difference between "before" and "after".

katahiromz: Can you elaborate the difference?

TAN-Gaming: @JoachimHenze 

> We still don't have a before-picture, that was shot with Katayamas code. Does it show the exact same icon or another one?

Katayama's code works, the result is the same.
This PR is from my analysis and comparison with other CPLs.

TAN-Gaming: Other CPLs don't have to set the small icon manually and works just fine with `int`.

So `LoadImage` code is a hack?

katahiromz: > So LoadImage code is a hack?

That was improper code.

TAN-Gaming: I think it would be better to set the icon the same way as other CPLs do.

katahiromz (approved):
It's a go.

PR #4689, opened 2022-09-14
[HOTPLUG] Add missing string resources

TAN-Gaming:
## Purpose
- Add missing string resources for hotplug.dll

This also fixes the missing title text on the taskbar.

Before:
ROS_hotplug_before

After:
ROS_hotplug_after


binarymaster (commented):
Keep translation in sync with window caption. Also, maybe there is a way to set the caption to the `IDS_CPLNAME` string to avoid duplication?

binarymaster:
> Also, maybe there is a way to set the caption to the `IDS_CPLNAME` string to avoid duplication?

Ok I did this part on my own.

binarymaster (approved):

PR #4691, opened 2022-09-14
[GDI32] Fix ExtTextOutA with ETO_GLYPH_INDEX

learn-more:
JIRA issue: [CORE-18365](https://jira.reactos.org/browse/CORE-18365)

after fix:
image

### DIFF
--- a/win32ss/gdi/gdi32/objects/text.c
+++ b/win32ss/gdi/gdi32/objects/text.c
@@ -465,8 +465,15 @@ ExtTextOutA(
     UNICODE_STRING StringU;
     BOOL ret;
 
-    RtlInitAnsiString(&StringA, (LPSTR)lpString);
+    if (fuOptions & ETO_GLYPH_INDEX)
+        return ExtTextOutW(hdc, x, y, fuOptions, lprc, (LPCWSTR)lpString, cch, lpDx);
+
+    StringA.Buffer = (PCHAR)lpString;
+    StringA.Length = StringA.MaximumLength = cch;
     RtlAnsiStringToUnicodeString(&StringU, &StringA, TRUE);
+
+    if (StringU.Length != StringA.Length * sizeof(WCHAR))
+        DPRINT1("ERROR: Should convert lpDx properly!\n");
 
     ret = ExtTextOutW(hdc, x, y, fuOptions, lprc, StringU.Buffer, cch, lpDx);
 


win32ss/gdi/gdi32/objects/text.c
```diff
@@ -465,9 +465,16 @@ ExtTextOutA(
...
 
-    RtlInitAnsiString(&StringA, (LPSTR)lpString);
+    if (fuOptions & ETO_GLYPH_INDEX)
+        return ExtTextOutW(hdc, x, y, fuOptions, lprc, (LPCWSTR)lpString, cch, lpDx);
+
+    StringA.Buffer = (PCHAR)lpString;
+    StringA.Length = StringA.MaximumLength = cch;
     RtlAnsiStringToUnicodeString(&StringU, &StringA, TRUE);
 
+    if (StringU.Length != (StringA.Length * 2))
```
SergeGautherie, line 475:
```suggestion
    if (StringU.Length != StringA.Length * sizeof(WCHAR))
```

win32ss/gdi/gdi32/objects/text.c
```diff
@@ -465,9 +465,16 @@ ExtTextOutA(
...
+        DPRINT1("ERROR: Should convert lpDx properly!\n");
```
HBelusca, line 476: Is this more like a "todo" for this ExtTextOutA function?

learn-more: yeah, this should receive some special handling, but was not hit when running some tests and this application so I left it.

katahiromz: Can `StringA` be a multi-byte string?

HBelusca: > Can `StringA` be a multi-byte string?

If so, then `RtlMultiByteToUnicodeN` and `RtlMultiByteToUnicodeSize` may be of interest.

katahiromz: In general, a multi-byte character (of multiple bytes) can be converted to one `WCHAR`. Now, the `A` suffix of `ExtTextOutA` means multi-byte version instead of ANSI version.

learn-more: I have no interest in fixing this now, but it should probably be converted manually here, because the `lpDx` should be kept in-sync with it.

HBelusca (approved):

PR #4694, opened 2022-09-14
Add back driver to access FATX partitions

hpoussin:
## Purpose

When switching from fastfat to fastfat_new driver (https://github.com/reactos/reactos/commit/c05f3d8d1ce5641f1768ecd775a375a34e368cf9), we lost the ability to access FATX partitions (usually found on XBOX).

JIRA issue: [CORE-16373](https://jira.reactos.org/browse/CORE-16373)

## Proposed changes

- add back fastfat driver when compiling for XBOX architecture, under the name vfatfs.sys.
- add a FATX recognizer in fs_rec.sys to only load vfatfs.sys when dealing with FATX partitions
- fix some bugs found while testing


HBelusca:
> under the name vfatfs.sys

Or `fastfatx.sys` ?

HBelusca:
For an independent reference, see also https://github.com/HBelusca/reactos/commits/fatx_fastfat

hpoussin:
> > under the name vfatfs.sys
> 
> Or `fastfatx.sys` ?

vfatfs.sys is the historical name of this driver.
As long as this driver also handles FAT, I don't think it should be named fastfatx.sys
The rename can come later, if/when we remove FAT support from it.

drivers/filesystems/CMakeLists.txt
```diff
@@ -2,7 +2,9 @@
 add_subdirectory(btrfs)
 add_subdirectory(cdfs)
 add_subdirectory(ext2)
-#add_subdirectory(fastfat)
+if(SARCH STREQUAL "xbox")
```
learn-more, line 5: If this would be enabled for other architectures, could we read xbox disks on those?

hpoussin: If the driver is enabled for other architectures, yes, you will be able to read FATX partitions.
However, if you want to read XBOX-partitioned disks (hard-coded partition sizes), you need to use the halxbox.dll HAL.

learn-more: Thanks for the explanation.

boot/bootdata/hivesys.inf
```diff
@@ -2191,6 +2191,13 @@ HKLM,"SYSTEM\CurrentControlSet\Services\fastfat","ImagePath",0x00020000,"system3
 HKLM,"SYSTEM\CurrentControlSet\Services\fastfat","Start",0x00010001,0x00000003
 HKLM,"SYSTEM\CurrentControlSet\Services\fastfat","Type",0x00010001,0x00000002
 
+; Virtual FATX filesystem driver
+HKLM,"SYSTEM\CurrentControlSet\Services\vfatfs","ErrorControl",0x00010001,0x00000001
+HKLM,"SYSTEM\CurrentControlSet\Services\vfatfs","Group",0x00000000,"Boot File System"
+HKLM,"SYSTEM\CurrentControlSet\Services\vfatfs","ImagePath",0x00020000,"system32\drivers\vfatfs.sys"
+HKLM,"SYSTEM\CurrentControlSet\Services\vfatfs","Start",0x00010001,0x00000003
+HKLM,"SYSTEM\CurrentControlSet\Services\vfatfs","Type",0x00010001,0x00000002
```
zefklop, line 2199: Wouldn't this fail on non-xbox systems ?

drivers/filesystems/fastfat/volume.c
```diff
@@ -100,13 +100,14 @@ FsdGetFsAttributeInformation(
...
+    switch (DeviceExt->FatInfo.FatType)
     {
-        pName = L"FAT32";
-    }
-    else
-    {
-        pName = L"FAT";
+        case FAT12: pName = L"FAT"; break;
+        case FAT16: pName = L"FAT"; break;
+        case FAT32: pName = L"FAT32"; break;
```
zefklop, line 107: Would this driver load for regular FAT partitions as well ? Maybe we should prefer the other one.

boot/bootdata/hivesys.inf
hpoussin: 3 is SERVICE_DEMAND_START. If a FATX partition is detected on non-XBOX system, kernel will try to start and fail to load the driver, and the partition would not be mounted.
If these lines are not present, kernel would fail to locate the vfatfs service announced by fs_rec, and the partition would not be mounted.

drivers/filesystems/fastfat/volume.c
learn-more: `FsRecIsFatxVolume` only accepts `FATX` partitions

binarymaster: Speaking of filesystem recognition, I wonder... how Windows implements it? Is our implementation different from it? Because for example I'm pretty sure Windows does not have BTRFS filesystem detection code, however it's possible to use that filesystem by just supplying the FS driver.

HBelusca: A FS recognizer is just a FS (declares as such with the kernel), it's just that it's super-lightweight and when a suitable FS is detected by it, it asks the kernel to load the actual FS driver.
If BTRFS works obviously on Windows (and Windows' fs_rec doesn't have BTRFS support), it's just because either the fullblown BTRFS driver is always loaded, or, because WinBtrfs also includes a small BTRFS-specific recognizer (that would also be a FS).

HBelusca: So, in ReactOS' case, instead of having also a separate smaller BTRFS driver whose purpose only would be to recognize Btrfs volumes and load the fullfledged BTRFS driver, we instead incorporate BTRFS detection within our single fs_rec.

boot/bootdata/hivesys.inf
HBelusca: Yes, those lines should be present like for example, any other filesystem driver supported (fastfat, ntfs, etc.)

drivers/filesystems/CMakeLists.txt
```diff
@@ -2,7 +2,9 @@
...
+add_subdirectory(fastfat)
```
HBelusca, line 6: I then wonder whether you could directly rename that fastfat directory into vfatfs, and the fastfat_new into fastfat. (Maybe in a separate commit?)

drivers/filesystems/fs_rec/fatx.c
```diff
@@ -0,0 +1,132 @@
...
+#include <pshpack1.h>
+typedef struct _FATX_BOOT_SECTOR
+{
+    UCHAR SysType[4];
+    ULONG VolumeId;
+    ULONG SectorsPerCluster;
+    USHORT FatCount;
+    ULONG Reserved;
+    UCHAR Unused[4078];
+
```
HBelusca, line 27:
```suggestion
```

drivers/filesystems/fs_rec/fs_rec.c
```diff
@@ -468,6 +474,17 @@ DriverEntry(IN PDRIVER_OBJECT DriverObject,
                              0);
     if (NT_SUCCESS(Status)) DeviceCount++;
 
+    /* Register FATX */
+    Status = FsRecRegisterFs(DriverObject,
+                             NULL,
+                             NULL,
+                             L"\\FatX",
```
HBelusca, line 481: Regarding this, in the (now old) fastfat, I think this name should be changed (to "FatX") as well:
https://git.reactos.org/?p=reactos.git;a=blob;f=drivers/filesystems/fastfat/iface.c;hb=e5ebbc8afaf104eff2db274a73131e83e6bed4dc#l55
(and we can get rid of the "RosFat" case in that function as well.)

boot/bootdata/hivesys.inf
disean: You might want to keep the registry data in a dedicated .inf file (vfatfs_reg.inf).

binarymaster (commented):
Suggesting these changes to make `vfatfs` driver building by all CI compilers, but integrated into system only for `SARCH=xbox`.

drivers/filesystems/CMakeLists.txt
```diff
@@ -2,7 +2,9 @@
 add_subdirectory(btrfs)
 add_subdirectory(cdfs)
 add_subdirectory(ext2)
-#add_subdirectory(fastfat)
+if(SARCH STREQUAL "xbox")
+add_subdirectory(vfatfs)
+endif()
```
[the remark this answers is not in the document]
binarymaster: And also, it should be on last line here now (alphabet order), because it was renamed.

binarymaster:
FATX tests on hardware are successful! :tada: But we didn't tried saving files to disk yet.

https://twitter.com/x86corez/status/1571595532623126528

binarymaster (commented):
Please add this line to `txtsetup.sif` file, because otherwise `vfatfs.sys` driver wouldn't be copied over to the hard drive after 1st stage setup:
```ini
vfatfs.sys   = 1,,,,,,x,4,,,,1,4
```
You can add it right after `fastfat.sys`

binarymaster (approved):
Only thing left is `txtsetup.sif` change that I suggested, but otherwise LGTM. :+1:

HBelusca:
As discussed with @binarymaster on Mattermost (and in the corresponding JIRA report), there needs to be an additional change to FreeLdr if we want to be able to boot directly from FATX volumes.
However, this may not be necessary for this PR, since we still have conceptual problems to solve before fully supporting such scenario (in particular, how to handle "long-file names" of more than 31 characters on FATX, required for SxS files...)

binarymaster:
I will probably do FreeLdr change separately, and associated with another ticket regarding booting.

PR #4697, opened 2022-09-15
Removed rapps.com

Artem13327:
Deleted rapps.com because it does almost nothing


learn-more:
Please inform yourself a bit better before just deleting shit.

PR #4698, opened 2022-09-15
[SHELL32] Fix Assert of OnContextMenu on Right click spamming while cancelling folder copy

KRosUser:
## Purpose

Missinng Release cached IContextMenu

JIRA issue: [CORE-18366](https://jira.reactos.org/browse/CORE-18366)

## Proposed changes

Same design pattern as 
https://github.com/reactos/reactos/pull/4683
https://github.com/reactos/reactos/pull/4683

### DIFF
--- a/dll/win32/shell32/CDefView.cpp
+++ b/dll/win32/shell32/CDefView.cpp
@@ -1579,6 +1579,13 @@ LRESULT CDefView::OnContextMenu(UINT uMsg, WPARAM wParam, LPARAM lParam, BOOL &b
 
     m_cidl = m_ListView.GetSelectedCount();
 
+    /* Release cached IContextMenu */
+    if (m_pCM)
+    {
+        IUnknown_SetSite(m_pCM, NULL);
+        m_pCM.Release();
+    }
+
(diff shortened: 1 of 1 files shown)


learn-more:
> ## Proposed changes
> Same design pattern as #4683 #4683

And after the last comment from #4683 you thought, heck, let's copy-paste this once more,
because copypasting the same code 6 times is for sure a great pattern!

And because most developers are lazy, here is the comment from said PR:

> > @learn-more : I've placed the cleanup BEFORE the call to GetItemObject like we did on #4675 but I also left it in the cleanup code in the end, not assuming what other function would do. Would you think this make sense ? Or should I remove the "trailing" cleanup now that's it's art the beginning ?
> > Btw this "cleanup" is missing in 2 other functions (unrelated to the bug I propose to fix) : OpenSelectedItems and OnContextMenu. Would you like me to add this there too as defensive programming ?
> 
> That this needs copy-pasting code in multiple functions is a code smell, so I'd rather not that you make matters worse. Instead, investigate if this object should be a class variable at all, or that it should be initialized with a helper instead that does cleanup before initializing.

PR #4706, opened 2022-09-16
[CPL] Boundary check of Control Panel applets

katahiromz:
## Purpose

Improve stability of the system.

## Proposed changes

- Check the boundary `(i < NUM_APPLETS)` of the variable `i` in `CPlApplet` functions.
- Use `UINT` type for the variable `i`.

### DIFF
--- a/dll/cpl/access/access.c
+++ b/dll/cpl/access/access.c
@@ -240,7 +240,7 @@ CPlApplet(HWND hwndCPl,
           LPARAM lParam1,
           LPARAM lParam2)
 {
-    INT i = (INT)lParam1;
+    UINT i = (UINT)lParam1;
 
     switch (uMsg)
     {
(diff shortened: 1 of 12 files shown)


TAN-Gaming:
Tested using `UINT`, the icon correctly displayed so the index can't be a negative value.

Tested modules list: ACCESS DESK INPUT INTL JOY MAIN MMSYS POWERCFG SYSDM TIMEDATE USRMGR

The documentation also uses the variable named `uAppNum`.
[https://docs.microsoft.com/en-us/windows/win32/shell/cpl-inquire](https://docs.microsoft.com/en-us/windows/win32/shell/cpl-inquire)

I think the correct data type is `UINT`.

learn-more:
> The boundary check is unnecessary, `CPL_INQUIRE` message sender should take care of it.

No.
We are writing an OS, not a hobby application that will run only on your own PC.
For Windows (and thus, ReactOS) everyone and his grandmother can (and has) written broken applications.
It is our job to ensure that even those (given, if they run on Windows) also run on ReactOS.

TAN-Gaming:
> > The boundary check is unnecessary, `CPL_INQUIRE` message sender should take care of it.
> 
> No. We are writing an OS, not a hobby application that will run only on your own PC. For Windows (and thus, ReactOS) everyone and his grandmother can (and has) written broken applications. It is our job to ensure that even those (given, if they run on Windows) also run on ReactOS.

OK.

TAN-Gaming (approved):

PR #4711, opened 2022-09-16
[EXPLORER] Fix lazy Clock Tray (CORE-12987)

KRosUser:
Implement fix to cover 4 cases refered in CORE-12987

## Purpose

Clock tray time indication, when displaying hh:mm (not second) is not updated at the right pace.

It "calibrates" a 60sec timer based on init and therefore gets out of sync if time (in particular seconds) is modified by NTP, API, user (from cpl), ...

This calibrated timer is overcomplex and not robust compared to a basic 1 Hz timer.

JIRA issue: [CORE-12987](https://jira.reactos.org/browse/CORE-12987)
JIRA issue: [CORE-11443](https://jira.reactos.org/browse/CORE-11443)

## Proposed changes

Simplify and always stick to a 1 Hz pace regardless of seconds being displayed or not
but optimize the condition of the UpdateWnd call to get the save performance saving (no unnecessary redraw of the clock tray area)

### DIFF
--- a/base/shell/explorer/trayclock.cpp
+++ b/base/shell/explorer/trayclock.cpp
@@ -426,8 +426,6 @@ UINT CTrayClockWnd::CalculateDueTime()
 
     GetLocalTime(&LocalTime);
     uiDueTime = 1000 - (UINT) LocalTime.wMilliseconds;
-    if (!g_TaskbarSettings.bShowSeconds)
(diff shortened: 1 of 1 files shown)


SergeGautherie:
I wonder : is the (Windows) tray clock supposed to poll to discover changes, or should it be receiving some system messages when needed?
As discussed in [CORE-11443](https://jira.reactos.org/browse/CORE-11443).

PR #4715, opened 2022-09-20
[EXPLORER] Add 'Show Desktop' button at right edge of taskbar

katahiromz:
## Purpose
This button allows the user to access "Show Desktop" feature by mouse.
JIRA issue: [CORE-15369](https://jira.reactos.org/browse/CORE-15369)

after1

## Proposed changes

- Add `CTrayShowDesktopButton` window class.
- Toggle the desktop show/hide on button action.

## TODO

- [x] Drawing over the non-client area of task bar.
- [x] Hovering.
- [x] Toggle by registry.
- [x] Theme.
- [x] Do tests.

## Screenshots

BEFORE:
before1

AFTER:
after1

BEFORE:
before2

AFTER:
after2

HOVERING:
hover

THEME:
theme


learn-more:
Can you also try this with one of our builtin themes?

base/shell/explorer/traywnd.cpp
```diff
@@ -264,6 +264,95 @@ class CStartButton
 
 };
 
+static const WCHAR szTrayShowDesktopButton[] = L"TrayShowDesktopButtonWClass";
+
+// The 'Show Desktop' button at edge of taskbar
+class CTrayShowDesktopButton :
+    public CComCoClass<CTrayShowDesktopButton>,
+    public CComObjectRootEx<CComMultiThreadModelNoCS>,
```
learn-more, line 272: Are the `CComCoClass` and `CComObjectRootEx` required here?
I think you only need the `CWindowImpl` in this case?

base/shell/explorer/traywnd.cpp
```diff
@@ -264,6 +264,95 @@ class CStartButton
...
+    public CWindowImpl<CTrayShowDesktopButton, CWindow, CControlWinTraits>
+{
+public:
+    DECLARE_WND_CLASS_EX(szTrayShowDesktopButton, CS_DBLCLKS | CS_HREDRAW | CS_VREDRAW,
```
learn-more, line 276: Do you need `CS_DBLCLKS` ?

base/shell/explorer/traywnd.cpp
katahiromz: Removed `CComCoClass` and `CComObjectRootEx`.

base/shell/explorer/traywnd.cpp
katahiromz: The user might want to double-click my head.

learn-more: Did you test that this works?

katahiromz:
a
Theming is not working yet.

base/shell/explorer/traywnd.cpp
katahiromz: Quickly repeating clicking is working well.

JoachimHenze: And what would happen *exacly*, if you would not have that CD_DBL_CLKS? Can you try to describe the samptom with words?

katahiromz: `CS_DBLCLKS` and handler are removed.

HBelusca:
Interestingly, in classic mode, Windows (7+) has this button within the expandable traybar:
image
image
I suppose it could be possible to do that too here for ROS' explorer. Notice how the "button" is flat, originally, and then becomes 3D when it is hovered upon (like toolbar buttons).

base/shell/explorer/traywnd.cpp
```diff
@@ -264,6 +264,115 @@ class CStartButton
...
+    DECLARE_WND_CLASS_EX(szTrayShowDesktopButton, CS_HREDRAW | CS_VREDRAW, COLOR_3DFACE)
+
+    CTrayShowDesktopButton() : m_nClickedTime(0)
+    {
+    }
+
+    INT WidthOrHeight() const
+    {
+        INT cxy = GetSystemMetrics(SM_CXEDGE);
+        return (cxy ? (2 * cxy) : 6);
```
HBelusca, line 285: What is "6" ? Also what would happen if SM_CXEDGE returned "1" (i.e. a very small value)?

base/shell/explorer/traywnd.cpp
```diff
@@ -264,6 +264,115 @@ class CStartButton
...
+        if (::SetWindowTheme(m_hWnd, L"Start", NULL) != S_OK)
+            ::SetWindowTheme(m_hWnd, L"Explorer", NULL);
+        return S_OK;
+    }
+
+    LRESULT OnClick(UINT uMsg, WPARAM wParam, LPARAM lParam, BOOL& bHandled)
+    {
+        LONG nTime0 = m_nClickedTime;
+        LONG nTime1 = ::GetMessageTime();
+        if (nTime1 - nTime0 >= 800) // 0.8sec
```
HBelusca, line 305: Don't hesitate to write comments in code, they exist for a reason. In particular, what is the reason for this thing.

base/shell/explorer/traywnd.cpp
```diff
@@ -264,6 +264,115 @@ class CStartButton
 
 };
 
+static const WCHAR szTrayShowDesktopButton[] = L"TrayShowDesktopButtonWClass";
```
HBelusca, line 267: Is it an invented name? Or is this the class name of the corresponding button on Windows 7+?

base/shell/explorer/traywnd.cpp
```diff
@@ -264,6 +264,115 @@ class CStartButton
...
+public:
+    DECLARE_WND_CLASS_EX(szTrayShowDesktopButton, CS_HREDRAW | CS_VREDRAW, COLOR_3DFACE)
+
+    CTrayShowDesktopButton() : m_nClickedTime(0)
+    {
+    }
+
+    INT WidthOrHeight() const
+    {
+        INT cxy = GetSystemMetrics(SM_CXEDGE);
```
HBelusca, line 284:
```suggestion
        INT cxy = ::GetSystemMetrics(SM_CXEDGE);
```

binarymaster (commented):
Uhm... not that I against this feature, but it's Windows 7+ shell feature. Maybe we should care more about Quick Launch panel, which has "Show Desktop" shortcut.

JoachimHenze:
> Uhm... not that I against this feature, but it's Windows 7+ shell feature. Maybe we should care more about Quick Launch panel, which has "Show Desktop" shortcut.

That is technically correct, but I think we can make an exception here. The Win7 way is simply more untuitive and it is extremely unlikely that anything depends on the "real" legacy way here. But I do agree, that our quicklaunch (which we do have since 2017 already) is lacking the general love and fixes it deserves, for too long already. But I definitely do not mean the .scf stuff with that. Let's be honest, who needs that?
I do mean all the other subtle bugs and shortcomings in it.

JoachimHenze (changes requested):
But what I do request here, is a feature toggle to turn it off. Just in registry would be ok as a start. A GUI way of switching could be done also later. Ideally we'd use the same switch as MS Windows.

HBelusca:
> But I definitely do not mean the .scf stuff with that. Let's be honest, who needs that?

Certainly now only 1 or 2 people on the whole Earth, since it has been deprecated in Vista+.

JoachimHenze:
> Also, regarding "windows 7+" feature, we have some sort of snap-to sides of screen (in win32k), which is kind of vista+/7, but I don't remember whether we have a registry switch to turn that off (for which purpose, by the way?)

https://jira.reactos.org/browse/CORE-16379 is the request to have a feature toggle for that, and it is also not implemented yet unfortunately. Reasons are many:
1.) usually a new feature brings also new bugs and liabilities with it usually, that we then have to endure for many years before they'll be solved, see everything linked against https://jira.reactos.org/browse/CORE-12845 as an example for the snapping feature. Having a toggle is a safe way out in case such things would evolve.

In case of the "Show desktop" expectable follow-up requests after initial implementation will be "make it hide all windows temporarilly already while hovering the mouse over the button, aka Desktop SneakPeak", or "Fix crash xyz quickly that was introduced together with the new feature"

2.) Some people may simply not like the feature and value some pixels extra in the taskbar for other things. E.g. personally I don't ever use that feature, but I just do use the keyboard shortcuts. MS often gives that freedom of choice for their features and allow things to be configurable.

But let's not get too offtopic here,and stop discussing the Window-Snap.

JoachimHenze:
Question:
On the start-button we do have a feature, that the button can even be invoked, if the mouse is moved to the very very very bottom left corner of the screen, even if that few Pixels are is actually not part of the start-button.
Can you make that work for your feature in a similar way?

base/shell/explorer/traywnd.cpp
katahiromz: Added comments.

base/shell/explorer/traywnd.cpp
katahiromz: It's confirmed by WinHier on Win10.

image

base/shell/explorer/traywnd.cpp
katahiromz: OK.

katahiromz:
> even if that few Pixels are actually not part of the start-button.
> Can you make that work for your feature in a similar way?

That is already done.

base/shell/explorer/traywnd.cpp
katahiromz: I meant `"6"` is a minimum width or height. Fixed.

katahiromz:
> a feature toggle to turn it off. A flag just in registry would be ok as a start

As far as I know, there is no registry value to disable to this feature.

base/shell/explorer/traywnd.cpp
katahiromz: 6:


5:


4:


3:


Which is the best choice?

learn-more: I'd say 3 or 4 is fine

base/shell/explorer/traywnd.cpp
```diff
@@ -264,6 +264,199 @@ class CStartButton
...
+        // But a too late action is an unexpected behaviour.
+        LONG nTime0 = m_nClickedTime;
+        LONG nTime1 = ::GetMessageTime();
+        if (nTime1 - nTime0 >= 600) // Ignore after 0.6 sec
+            return 0;
+
+        INPUT inputs[4];
+        ZeroMemory(inputs, sizeof(inputs));
+
+        // Emulate Win+D
```
HBelusca, line 319: I don't know why, but there is already an existing function in this very same file, that is called by clicking on the "Show desktop" taskbar context menu item, whose handling is there:
https://git.reactos.org/?p=reactos.git;a=blob;f=base/shell/explorer/traywnd.cpp;hb=34635e1585e97539358d975c4fe0e03758c3c14d#l612
(and related: https://git.reactos.org/?p=reactos.git;a=blob;f=base/shell/explorer/traywnd.cpp;hb=34635e1585e97539358d975c4fe0e03758c3c14d#l568 )
Why not calling those functions directly instead?

base/shell/explorer/traywnd.cpp
```diff
@@ -2224,6 +2461,9 @@ class CTrayWindow :
         /* Create the Start button */
         m_StartButton.Create(m_hWnd);
 
+        /* Create the 'Show Desktop' button */
+        m_ShowDesktopButton.DoCreate(m_hWnd, IsShowDesktopButtonNeeded());
```
HBelusca, line 2465: I was under the impression that the feature was meant (in case it's not enabled) to not create the button at all, i.e. _*disable*_ completely the feature.
Here you always create it, but just show or hide it, meaning that one could still send messages to its window (programmatically, thus still being able to use the feature) if one knows it exists there.

base/shell/explorer/traywnd.cpp
katahiromz: `CTrayShowDesktopButton` is a child window.
Now we use `WM_COMMAND:TRAYCMD_TOGGLE_DESKTOP` instead.

base/shell/explorer/traywnd.cpp
katahiromz: OK. Fixed.

base/shell/explorer/traywnd.cpp
HBelusca: In any case you could call those `ShowDesktop()` etc. functions?

base/shell/explorer/traywnd.cpp
HBelusca: Ah, and yes, as shown in your screenshot (which, by the way, gives the same window hierarchy as on Windows 7), we can see that the show-desktop button is *within* the trayNotifyWnd, i.e. next to the clock etc. within that tray-window, as shown in my screenshots elsewhere in this PR. Not outside as seems to be from your screenshots in the PR description.

base/shell/explorer/traywnd.cpp
katahiromz: `PostMessage` function does add a message onto the message queue. Thus, it can be delayed the action when the system is too busy. It is good for the user to queue mouse input messages as expected behaviour.

katahiromz: `ShowDesktop()` is in class `CTrayWindow`. This code is in class `CTrayShowDesktopButton`.

base/shell/explorer/traywnd.cpp
katahiromz: Should we touch `TrayNotifyWnd`? It's a complicated hierarchy.

base/shell/explorer/traywnd.cpp
HBelusca: How does it behave with double-clicks?

katahiromz: Now, we have no `CS_DBLCLKS` flag. So double-clicks are converted to single clicks.

base/shell/explorer/traywnd.cpp
katahiromz: Applied 3.

base/shell/explorer/traywnd.cpp
```diff
@@ -264,6 +264,200 @@ class CStartButton
...
+        return S_OK;
+    }
+
+    LRESULT OnClick(UINT uMsg, WPARAM wParam, LPARAM lParam, BOOL& bHandled)
+    {
+        // The actual action can be delayed as an expected behaviour.
+        // But a too late action is an unexpected behaviour.
+        LONG nTime0 = m_nClickedTime;
+        LONG nTime1 = ::GetMessageTime();
+        if (nTime1 - nTime0 >= 600) // Ignore after 0.6 sec
```
learn-more, line 309: are you looking for `GetDoubleClickTime ` ?

base/shell/explorer/traywnd.cpp
```diff
@@ -264,6 +264,200 @@ class CStartButton
...
+    VOID Click()
+    {
+        // The actual action can be delayed as an expected behaviour.
+        m_nClickedTime = ::GetMessageTime();
+        PostMessage(TSDB_CLICK, 0, 0);
+    }
+
+    LRESULT OnLButtonDown(UINT uMsg, WPARAM wParam, LPARAM lParam, BOOL& bHandled)
+    {
+        Click(); // Left-click
```
learn-more, line 329: On windows this responds to the button up I think

base/shell/explorer/traywnd.cpp
katahiromz: Nope. That is not related to double-clicks.

base/shell/explorer/traywnd.cpp
katahiromz: This behavior looks quicker.

JoachimHenze: Mark is correct, and we should stick to what Windows does.

learn-more:
>  E.g: Marks comment that it should actually use LBUTTON_UP instead of LBUTTON_DOWN has not been addressed yet.

I reviewed that after it was merged, because it went from 'not complete' to 'merged' overnight.

katahiromz:
> I presume this isn't supposed to happen

Nice catch. Next PR will fix it.

PR #4716, opened 2022-09-21
[MMSYS] Fix "Speaker settings" image

TAN-Gaming:
## Purpose
- Fix "Speaker settings" image ( `IDC_SPEAKIMG` )
- Addendum to [781c247](https://github.com/reactos/reactos/commit/781c247bd3c5aeee799527db2ce06e15f794b0a4)

## Proposed changes
- Use 72x72 icon instead of bitmap
- Use correct control position and size for `IDC_SPEAKIMG`
- Also add NULL check for `DestroyIcon` and fix cleanup order in `VolumeDlgProc`

Mizu Before:
ROS_MMSYS_before

Mizu After:
ROS_MMSYS_after_2

Classic theme:
ROS_MMSYS_classic_2

## Notes
- This fix also applies to Lautus, Lunar and Modern theme
- The original file is from Tango Icon Theme 0.8.90


katahiromz:
We can reduce code by using a huge icon instead of color  bitmap and mask bitmap.

katahiromz (approved):

PR #4722, opened 2022-09-23
[EXPLORER] Cleanup after last commit CORE-15369

JoachimHenze:
Improves code that was added by 0.4.15-dev-5078-g 46de3a22acc964f27a5260e144089125f83586ff without any review yesterday.

JIRA issue: [CORE-15369](https://jira.reactos.org/browse/CORE-15369)

### DIFF
--- a/base/shell/explorer/traywnd.cpp
+++ b/base/shell/explorer/traywnd.cpp
@@ -3045,10 +3045,8 @@ HandleTrayContextMenu:
 
     LRESULT OnNcLButtonUp(UINT uMsg, WPARAM wParam, LPARAM lParam, BOOL& bHandled)
     {
-        if (CheckShowDesktopButtonClick(lParam, bHandled))
-            return 0;
-
-        return 0;
+        CheckShowDesktopButtonClick(lParam, bHandled);
+        return FALSE;
     }
 
     LRESULT OnAppTrayDestroy(UINT uMsg, WPARAM wParam, LPARAM lParam, BOOL& bHandled)


katahiromz (approved):

base/shell/explorer/traywnd.cpp
```diff
@@ -3045,10 +3045,8 @@ class CTrayWindow :
 
     LRESULT OnNcLButtonUp(UINT uMsg, WPARAM wParam, LPARAM lParam, BOOL& bHandled)
     {
-        if (CheckShowDesktopButtonClick(lParam, bHandled))
-            return 0;
-
-        return 0;
+        CheckShowDesktopButtonClick(lParam, bHandled);
+        return FALSE;
```
learn-more, line 3049: Why FALSE?
The return is an LRESULT, not a BOOLEAN

binarymaster: I think this is following existing code in this file. There are other functions which were coded the same way.

HBelusca (approved):

base/shell/explorer/traywnd.cpp
JoachimHenze: @learn-more yes, I did orienedt at all the surrounding functions here that returned LRESULT as TRUE/FALSE.
Do you want me to change them all? Or can I push it like that? I don't have any preference in that regard.

learn-more: TBH this PR already cost more time than it will ever gain.

PR #4724, opened 2022-09-24
[DDRAW] Resize the window after setting the new display mode

masnuvahaque:
## Purpose

This gets the taskbar to not show when running a DirectDraw based program in Full-Screen Mode.

JIRA issue: [CORE-11242](https://jira.reactos.org/browse/CORE-11242)

This fix is taken from code found in https://source.winehq.org/git/wine.git/commitdiff/84413298de371a9e9b9ecfe0dd01f580b7521222


dll/directx/wine/ddraw/ddraw.c
```diff
@@ -1149,6 +1150,10 @@ static HRESULT WINAPI ddraw7_SetDisplayMode(IDirectDraw7 *iface, DWORD width, DW
 
     InterlockedCompareExchange(&ddraw->device_state, DDRAW_DEVICE_STATE_NOT_RESTORED, DDRAW_DEVICE_STATE_OK);
 
+    if (ddraw->cooperative_level & DDSCL_EXCLUSIVE)
+       SetWindowPos(ddraw->dest_window, HWND_TOP, 0, 0, width, height, SWP_SHOWWINDOW | SWP_NOACTIVATE);
+       TRACE("DirectDraw window has been resized\n");
+
```
JoachimHenze, line 1156: Please guard this block by #ifdef __REACTOS__
as ddraw is code from Wine, not ours.

dll/directx/wine/ddraw/ddraw.c
```diff
@@ -1,4 +1,4 @@
-/*
+﻿/*
```
JoachimHenze, line 1: This line should not be changed. That can be achieved by using an editor, that doesn't touch the ByteOrderMark, e.g. Notepad++.
You must not add BOM.

dll/directx/wine/ddraw/ddraw.c
```diff
@@ -1132,6 +1132,7 @@ static HRESULT WINAPI ddraw7_SetDisplayMode(IDirectDraw7 *iface, DWORD width, DW
 
     /* TODO: The possible return values from msdn suggest that the screen mode
      * can't be changed if a surface is locked or some drawing is in progress. */
+
```
JoachimHenze, line 1135: This change should be reverted before merging, but please wait with that until the rest of your changes have been reviewed by multiple devs. You intended to highlight the comment above I guess. Good idea.

JoachimHenze (changes requested):

dll/directx/wine/ddraw/ddraw.c
gonzoMD: The code is taken from Wine, so why guarding it?

JoachimHenze: https://github.com/wine-mirror/wine/blob/master/dlls/ddraw/ddraw.c at line 1170
shows the line InterlockedCompareExchange(&ddraw->device_state, DDRAW_DEVICE_STATE_NOT_RESTORED, DDRAW_DEVICE_STATE_OK);
after the 
if (ddraw->cooperative_level & DDSCL_EXCLUSIVE)

while for us it is the other way round. Why is that? Bug or intended? That is at least a diff

dll/directx/wine/ddraw/ddraw.c
JoachimHenze: LetsHaveALook

dll/directx/wine/ddraw/ddraw.c
JoachimHenze: LetsHaveALook

JoachimHenze: Does your patch also work if you placed it, where Wine placed its corresponding block?

masnuvahaque:
Closing this pull request due conflict with commits

masnuvahaque:
Reopening this pull request

JoachimHenze:
According to "theDarkFire" this patch broke its effectiveness by 0.4.15-dev-5382-ga2c6af0 https://github.com/reactos/reactos/commit/a2c6af0da4acbba9c254d003e9d9f4ea6e03ed63 . I have not double-checked that claim yet, but found it worth to add that information here.

PR #4725, opened 2022-09-24
[FREELDR] Allow to boot from FATX volume using vfatfs driver

binarymaster:
This fixes JIRA issue: [CORE-16329](https://jira.reactos.org/browse/CORE-16329)

Creating a PR to run on testbots, to make sure 1st stage setup passes on non-Xbox.

### DIFF
--- a/boot/freeldr/freeldr/lib/fs/fat.c
+++ b/boot/freeldr/freeldr/lib/fs/fat.c
@@ -1547,6 +1547,16 @@ const DEVVTBL FatFuncTable =
     L"fastfat",
 };
 
+const DEVVTBL FatXFuncTable =
+{
+    FatClose,
+    FatGetFileInformation,
+    FatOpen,
+    FatRead,
+    FatSeek,
+    L"vfatfs",
+};
+
 const DEVVTBL* FatMount(ULONG DeviceId)
 {
     PFAT_VOLUME_INFO Volume;
@@ -1634,5 +1644,5 @@ const DEVVTBL* FatMount(ULONG DeviceId)
     // Return success
     //
     TRACE("FatMount(%lu) success\n", DeviceId);
-    return &FatFuncTable;
+    return (ISFATX(Volume->FatType) ? &FatXFuncTable : &FatFuncTable);
 }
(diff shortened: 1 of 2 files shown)


binarymaster:
I'm getting this in VirtualBox when testing manually:

VirtualBox_ReactOS_24_09_2022_17_03_38

Probably some flags needs to be adjusted to make this file optional.

boot/freeldr/freeldr/lib/fs/fat.c
```diff
@@ -1590,10 +1601,12 @@ const DEVVTBL* FatMount(ULONG DeviceId)
     //
     // Check if BootSector is valid. If no, return early
     //
+    IsFatX = RtlEqualMemory(BootSectorX->FileSystemType, "FATX", 4);
+
     if (!RtlEqualMemory(BootSector->FileSystemType, "FAT12   ", 8) &&
         !RtlEqualMemory(BootSector->FileSystemType, "FAT16   ", 8) &&
         !RtlEqualMemory(BootSector32->FileSystemType, "FAT32   ", 8) &&
-        !RtlEqualMemory(BootSectorX->FileSystemType, "FATX", 4))
+        !IsFatX)
```
HBelusca, line 1609: Better fix is to keep the original code here, and in the last return statement for the function table, you do instead:
`return (ISFATX(Volume->FatType) ? &FatXFuncTable : &FatFuncTable);`

HBelusca:
> I'm getting this in VirtualBox when testing manually:

Is it with a XBOX-configured image? or a usual one from reatos.org/getbuilds ?
Is this error showing when start 1st-stage setup? or 2nd-stage? Was there a FATX disk attached to the VM?

binarymaster:
> Is it with a XBOX-configured image? or a usual one from reatos.org/getbuilds ?

This is usual manual build for PC/AT.

> Is this error showing when start 1st-stage setup? or 2nd-stage?

Right before 1st stage.

> Was there a FATX disk attached to the VM?

No it was not.

HBelusca:
OK the problem why in a usual PC/AT setting you get this error is because of this:
(vfatfs CMakeFiles.txt)
```
  35 if(SARCH STREQUAL "xbox")
  36     add_cd_file(TARGET vfatfs DESTINATION reactos/system32/drivers NO_CAB FOR all)
  37     add_registry_inf(vfatfs_reg.inf)
  38 endif()
```
it's only added to the ISO (and in the registry) only if you build an XBOX ISO, and not a usual one (which also means that in a usual PC/AT ReactOS installation, if you ever wanted to mount and read a FATX volume, you couldn't, because the vfatfs.sys driver wouldn't have been present in your ROS installation to begin with).

binarymaster:
Ok so removing "x" seems to solve both problems.

PR #4726, opened 2022-09-25
[CPL][INPUT] Make the ListView a TreeView

katahiromz:
## Purpose

Improve UI/UX compatibility to WinXP/2k3.

BEFORE:
before

AFTER:
after

JIRA issue: N/A

## Proposed changes

- Add `InputList_Sort` for sorting the list.
- Modify `IDD_PROPPAGESETTINGS` resource dialog for compatibility.
- Make the list a TreeView.
- Add some icons.

## TODO

- [x] Fix resource and translation.
- [ ] Disable TreeView collapse (Unable to fix due to COMCTL32's bug).
- [ ] Consistent selection (Unable to fix due to COMCTL32's bug).


HBelusca:
Is Windows using a treeview as well for that? Or just a usual ListView but with indented items?

HBelusca:
No idea why the keyboard icon looks kind of ugly in your screenshot, but is it the icon from dll/cpl/main/resources/keyboard.ico ? Because if it isn't, then I would strongly suggest you use that latter one instead.

binarymaster:
Indeed, small keyboard icon looks quite meh... but I think it's out of context here :)

binarymaster:
@katahiromz please use this keyboard icon: https://github.com/reactos/reactos/blob/master/dll/cpl/main/resources/keyboard.ico

dll/cpl/input/input_list.h
```diff
@@ -60,6 +60,8 @@ InputList_SetDefault(INPUT_LIST_NODE *pNode);
 VOID
 InputList_Remove(INPUT_LIST_NODE *pNode);
 
+VOID InputList_Sort(VOID);
```
HBelusca, line 63:
```suggestion
VOID
InputList_Sort(VOID);
```

dll/cpl/input/lang/en-US.rc
```diff
@@ -1,19 +1,23 @@
 LANGUAGE LANG_ENGLISH, SUBLANG_ENGLISH_US
 
 IDD_PROPPAGESETTINGS DIALOGEX 0, 0, 254, 228
-STYLE DS_SHELLFONT | WS_CHILD | WS_DISABLED | WS_CAPTION
 CAPTION "Settings"
+STYLE DS_SHELLFONT | WS_CHILD | WS_DISABLED | WS_CAPTION
```
HBelusca, line 5: In all those translations, please keep the "STYLE" line before the "CAPTION".

dll/cpl/input/lang/en-US.rc
```diff
@@ -1,19 +1,23 @@
...
-    LTEXT "Select the services that you want for each input language shown in the list.\nUse the Add and Remove buttons to modify this list.", -1, 9, 6, 238, 25
-    CONTROL "TEXT", IDC_KEYLAYOUT_LIST, "SYSLISTVIEW32", WS_BORDER | WS_VSCROLL | WS_TABSTOP | LVS_SINGLESEL | LVS_NOSORTHEADER | LVS_SHOWSELALWAYS | LVS_REPORT, 8, 36, 237, 101
-    PUSHBUTTON "&Set Default", IDC_SET_DEFAULT, 101, 159, 144, 14
-    PUSHBUTTON "A&dd...", IDC_ADD_BUTTON, 27, 142, 70, 14
-    PUSHBUTTON "&Remove", IDC_REMOVE_BUTTON, 101, 142, 70, 14
-    PUSHBUTTON "&Properties...", IDC_PROP_BUTTON, 175, 142, 70, 14
-    GROUPBOX "Preferences", -1, 7, 185, 240, 36
-    PUSHBUTTON "&Key Settings...", IDC_KEY_SET_BTN, 14, 198, 110, 14
-END
+{
```
HBelusca, line 7: Please keep "BEGIN" / "END" instead of the curly brackets.

dll/cpl/input/lang/he-IL.rc
```diff
@@ -7,14 +7,18 @@ STYLE DS_SHELLFONT | WS_CHILD | WS_DISABLED | WS_CAPTION
...
+    GROUPBOX "Default input &language", -1, 5, 5, 246, 55
+    LTEXT "Select one of the installed input languages to use when your computer starts.", -1, 15, 15, 230, 23
+    COMBOBOX IDC_DEFAULT_LANGUAGE, 15, 40, 230, 300, CBS_HASSTRINGS | CBS_AUTOHSCROLL | CBS_DROPDOWNLIST | WS_VSCROLL | WS_TABSTOP
+    GROUPBOX "&Installed services", -1, 5, 65, 246, 125
+    LTEXT "Select the services that you want for each input language shown in the list. Use the Add and Remove buttons to modify this list.", -1, 15, 78, 230, 25
+    CONTROL "", IDC_KEYLAYOUT_LIST, "SysTreeView32", TVS_SHOWSELALWAYS | WS_BORDER | WS_VSCROLL | WS_TABSTOP, 15, 105, 155, 80
+    PUSHBUTTON "הוסף...", IDC_ADD_BUTTON, 175, 105, 70, 14
+    PUSHBUTTON "הסר", IDC_REMOVE_BUTTON, 175, 125, 70, 14
+    PUSHBUTTON "מאפיניים...", IDC_PROP_BUTTON, 175, 145, 70, 14
+    GROUPBOX "העדפות", -1, 5, 195, 245, 30
```
HBelusca, line 19: For some reason the groupbox line here doesn't display correctly.

dll/cpl/input/resource.h
```diff
@@ -37,6 +38,8 @@
 #define IDC_CTRL_LAYOUT          1018
 #define IDC_LEFT_ALT_LANG        1019
 #define IDC_LEFT_ALT_LAYOUT      1020
+#define IDC_LANGUAGE_BAR         1021
+#define IDC_DEFAULT_LANGUAGE     1022
```
HBelusca, line 42: It seems to me you could just recycle the value of the old `IDC_SET_DEFAULT` here.

dll/cpl/input/settings_page.c
```diff
@@ -11,192 +11,341 @@
...
-                    hBmpOld = SelectObject(hdc, hBitmap);
-
-                    if (hBmpOld != NULL)
-                    {
-                        RECT rect;
-
-                        SetRect(&rect, 0, 0, width / 2, height);
+        StringCchCopyW(szBuf, ARRAYSIZE(szBuf), L"??");
+    }
+    szBuf[2] = 0; /* "ENG" --> "EN" etc. */
```
HBelusca, line 37:
```suggestion
    szBuf[2] = UNICODE_NULL; /* Truncate the identifiers to two characters: "ENG" --> "EN" etc. */
```

dll/cpl/input/settings_page.c
```diff
@@ -11,192 +11,341 @@
...
+
+    /* Prepare for DIB (device-independent bitmap) */
+    ZeroMemory(&bmi, sizeof(bmi));
+    bmi.bmiHeader.biSize = sizeof(BITMAPINFOHEADER);
```
HBelusca, line 41:
```suggestion
    bmi.bmiHeader.biSize = sizeof(bmi.bmiHeader);
```

dll/cpl/input/settings_page.c
```diff
@@ -11,192 +11,341 @@
...
+        insert.item         = item;
+        hItem = TreeView_InsertItem(hwndList, &insert);
+
+        // Keyboard
+        ZeroMemory(&item, sizeof(item));
+        item.mask           = TVIF_TEXT | TVIF_IMAGE | TVIF_PARAM | TVIF_SELECTEDIMAGE | TVIF_STATE;
+        item.pszText        = szKeyboard;
+        item.iImage         = KeyboardImageIndex;
+        item.iSelectedImage = KeyboardImageIndex;
+        item.lParam         = 0xFFFF;
```
HBelusca, line 236: Not `((LPARAM)-1)` ?

dll/cpl/input/settings_page.c
```diff
@@ -11,192 +11,341 @@
...
-    item.mask    = LVIF_TEXT | LVIF_PARAM | LVIF_IMAGE;
-    item.pszText = pInputNode->pLocale->pszName;
-    item.iItem   = ListView_GetItemCount(hwndList);
-    item.lParam  = (LPARAM)pInputNode;
-    item.iImage  = ImageIndex;
-    ItemIndex = ListView_InsertItem(hwndList, &item);
+    // Input method
+    if (hItem)
+    {
+        // FIXME: IME icon
```
HBelusca, line 279: In which aspect this needs to be fixed?

dll/cpl/input/settings_page.c
```diff
@@ -453,8 +598,7 @@ SettingsPageProc(HWND hwndDlg, UINT uMsg, WPARAM wParam, LPARAM lParam)
             break;
 
         case WM_NOTIFY:
-            OnNotifySettingsPage(hwndDlg, lParam);
-            break;
+            return OnNotifySettingsPage(hwndDlg, lParam);
```
HBelusca, line 601: Does this function need to return values that are potentially different from zero?

dll/cpl/input/settings_page.c
```diff
@@ -11,192 +11,341 @@
 #include "locale_list.h"
 #include "input_list.h"
 
+static INT s_nLeavesCount = 0;
```
HBelusca, line 14: Isn't this somehow related to the number of items/"nodes" in the `_InputList` ?

dll/cpl/input/settings_page.c
```diff
@@ -11,192 +11,341 @@
...
+    SetRect(&rect, 0, 0, cxIcon, cyIcon);
+
+    /* Draw hbmColor */
+    hBmpOld = SelectObject(hdc, hbmColor);
+    SetDCBrushColor(hdc, GetSysColor(COLOR_HIGHLIGHT));
+    FillRect(hdc, &rect, (HBRUSH)GetStockObject(DC_BRUSH));
+    hFontOld = SelectObject(hdc, hFont);
+    SetTextColor(hdc, GetSysColor(COLOR_HIGHLIGHTTEXT));
+    SetBkMode(hdc, TRANSPARENT);
+    DrawText(hdc, szBuf, 2, &rect, DT_SINGLELINE | DT_CENTER | DT_VCENTER);
```
HBelusca, line 81:
```suggestion
    DrawTextW(hdc, szBuf, 2, &rect, DT_SINGLELINE | DT_CENTER | DT_VCENTER);
```

dll/cpl/input/settings_page.c
```diff
@@ -11,192 +11,341 @@
...
-    BOOL bCanProp = (iSelected != -1);
-
-    LV_ITEM item = { LVIF_PARAM, iSelected };
-    if (ListView_GetItem(hwndList, &item))
+    INT LangImageIndex = -1, KeyboardImageIndex = -1, ImeImageIndex = -1;
+    HICON hLangIcon, hKeyboardIcon, hInputMethodIcon;
+    TV_ITEM item;
+    TV_INSERTSTRUCT insert;
+    HIMAGELIST hImageList = TreeView_GetImageList(hwndList, TVSIL_NORMAL);
+    TCHAR szKeyboard[] = TEXT("Keyboard");
```
HBelusca, line 186: Should this need to be localized? or not?
Also it seems you could just have used something like:
```suggestion
    PCWSTR pszKeyboard = L"Keyboard";
```
or even directly do `item.pszText = "Keyboard";` below :thinking:

dll/cpl/input/settings_page.c
```diff
@@ -11,192 +11,341 @@
...
+        ZeroMemory(&item, sizeof(item));
+        item.mask           = TVIF_STATE | TVIF_HANDLE;
+        item.hItem          = hItem;
+        item.stateMask      = TVIS_BOLD;
+        if (TreeView_GetItem(hwndList, &item) && bBold && !(item.state & TVIS_BOLD))
         {
-            ImageIndex = ImageList_AddIcon(hImageList, hLayoutIcon);
-            DestroyIcon(hLayoutIcon);
+            item.mask = TVIF_STATE | TVIF_HANDLE;
+            item.hItem = hItem;
```
HBelusca, line 270: Should both of these attributes (the mask and hItem) be reset yet again?

HBelusca:
As a general information:
- Windows 7 reworked that dialog a little bit and added some useful buttons:
image
- The contents of the two dialogs regarding the "Language Bar" and the "Key Settings" became separate tabs. I don't know whether this can be useful for us too.

A question about the design: I know Windows XP+ added that chonky "Default Input Language" groupbox with explanation and combobox.... I don't know, this looks a bit redundant, wouldn't it be better to keep the "Set Default" button and just have the default layout directly selected in the treeview (and maybe labeled in bold / italic / whatever)?

dll/cpl/input/settings_page.c
katahiromz: The `HIWORD` value is reserved.

dll/cpl/input/settings_page.c
katahiromz: Loading 1st resource icon from IME file...

dll/cpl/input/settings_page.c
katahiromz: In order to prevent Collapse in future.

dll/cpl/input/settings_page.c
HBelusca: ok this needs to be spelled out in the comment (unless you expect to work on that in the near future)

dll/cpl/input/input_list.h
katahiromz: Fixed.

dll/cpl/input/lang/en-US.rc
katahiromz: Fixed.

dll/cpl/input/resource.h
katahiromz: Fixed.

dll/cpl/input/settings_page.c
katahiromz: OK.

dll/cpl/input/settings_page.c
katahiromz: Fixed.

dll/cpl/input/settings_page.c
katahiromz: Fixed.

dll/cpl/input/settings_page.c
katahiromz: Used `LoadStringW`. Done.

dll/cpl/input/lang/en-US.rc
katahiromz: Fixed.

dll/cpl/input/lang/he-IL.rc
katahiromz: It's a Right-to-Left text.

dll/cpl/input/settings_page.c
katahiromz: See `SetControlsState` function.

dll/cpl/input/settings_page.c
katahiromz: Added comment.

dll/cpl/input/settings_page.c
```diff
@@ -11,192 +11,343 @@
...
+    // Input method
+    if (hItem)
+    {
+        // FIXME: Load 1st resource icon from "IME File"...
+        hInputMethodIcon = (HICON)LoadImageW(hApplet,
+                                             MAKEINTRESOURCEW(IDI_DOT),
+                                             IMAGE_ICON,
+                                             GetSystemMetrics(SM_CXSMICON),
+                                             GetSystemMetrics(SM_CYSMICON),
+                                             0);
```
binarymaster, line 287: It might be better to draw the dot icon manually (using some font), because a black dot would be invisible on black background (i.e. when some themes are applied).

dll/cpl/input/settings_page.c
katahiromz: The root "Locale" items can be shared some leaves.
Zero or one leaves can be default.

dll/cpl/input/settings_page.c
katahiromz: That icon is a black dot surrounded by white border. No problem.

dll/cpl/input/settings_page.c
katahiromz: This is the alive leaf count. Aliveness can be changed anytime.
I don't want to manage it in `INPUT_LIST_NODE` structure.

dll/cpl/input/lang/he-IL.rc
HBelusca: Sure, except that why does the problem show there only here and not for the other strings above?

katahiromz:
> wouldn't it be better to keep the "Set Default" button and just have the default layout directly selected in the treeview

The light user cannot set the default item due to operation difficulty. The combo box is easier.

HBelusca:
> The light user cannot set the default item due to operation difficulty. The combo box is easier.

Wouldn't the user have to select, in any case, something from the list? In this way I don't see why this would be more difficult. (The only possible difficulty I would imagine is to find in the treelist the current selected default :thinking:)

dll/cpl/input/lang/he-IL.rc
katahiromz: https://www.compart.com/en/unicode/U+05D4
The character "ה" is a right-to-left letter.

HBelusca: But aren't they **all** RTL ? :smile:

katahiromz: Umm... Dunno.

katahiromz:
> please use this keyboard icon:

Used.

JoachimHenze (approved):

dll/cpl/input/settings_page.c
katahiromz: For example:
ex

katahiromz:
Japanese and IME:
japanese

dll/cpl/input/settings_page.c
katahiromz: NOTE: `comctl32!TREEVIEW_Collapse` function has a bug. That function didn't check the return value of `TREEVIEW_SendExpanding`.

katahiromz: https://bugs.winehq.org/show_bug.cgi?id=53727

dll/cpl/input/settings_page.c
katahiromz: Like this:
like-this

dll/cpl/input/settings_page.c
katahiromz: Win2k3:


Windows doesn't bold the keyboard item text...

katahiromz: Fixed.

dll/cpl/input/settings_page.c
HBelusca: The bullet looks nice this way :)

dll/cpl/input/settings_page.c
```diff
@@ -394,18 +563,28 @@ BOOL EnableProcessPrivileges(LPCWSTR lpPrivilegeName, BOOL bEnable)
...
         {
-            if (header->idFrom == IDC_KEYLAYOUT_LIST)
+            SetControlsState(hwndDlg);
+        }
+        break;
+
+        case TVN_ITEMEXPANDING:
+        {
+            // FIXME: Prevent collapse (COMCTL32 is buggy)
+            // https://bugs.winehq.org/show_bug.cgi?id=53727
```
HBelusca, line 582: Question is whether they have it fixed in their current code or not (since we don't use the latest synced comctl32)

dll/cpl/input/settings_page.c
binarymaster: Heh I just checked, Windows does not recolor the bullet in "high contrast black" theme. So we do this better :) However maybe make it just a bit smaller? Looks quite bold.

dll/cpl/input/lang/en-US.rc
```diff
@@ -5,14 +5,18 @@ STYLE DS_SHELLFONT | WS_CHILD | WS_DISABLED | WS_CAPTION
...
-    LTEXT "Select the services that you want for each input language shown in the list.\nUse the Add and Remove buttons to modify this list.", -1, 9, 6, 238, 25
-    CONTROL "TEXT", IDC_KEYLAYOUT_LIST, "SYSLISTVIEW32", WS_BORDER | WS_VSCROLL | WS_TABSTOP | LVS_SINGLESEL | LVS_NOSORTHEADER | LVS_SHOWSELALWAYS | LVS_REPORT, 8, 36, 237, 101
-    PUSHBUTTON "&Set Default", IDC_SET_DEFAULT, 101, 159, 144, 14
-    PUSHBUTTON "A&dd...", IDC_ADD_BUTTON, 27, 142, 70, 14
-    PUSHBUTTON "&Remove", IDC_REMOVE_BUTTON, 101, 142, 70, 14
-    PUSHBUTTON "&Properties...", IDC_PROP_BUTTON, 175, 142, 70, 14
-    GROUPBOX "Preferences", -1, 7, 185, 240, 36
-    PUSHBUTTON "&Key Settings...", IDC_KEY_SET_BTN, 14, 198, 110, 14
+    GROUPBOX "Default input &language", -1, 5, 5, 246, 55
+    LTEXT "Select one of the installed input languages to use when your computer starts.", -1, 15, 15, 230, 23
```
HBelusca, line 9: It may be better to use the formulation from Windows 7+ and say instead:
`"Select one of the installed input languages to use as the default for all input fields."`
since:
1. It's obvious it will be used when the computer starts (since it will be used when it's running as well);
2. Especially in case we are able to change the default input language **without having to restart the computer** (OK I know there's still a limitation for IMEs?), then it is definitively not useful to mention it's going to be used "when the computer starts".

I'll let you judge about that (especially if the two points above are true in the current state of ReactOS).

dll/cpl/input/lang/fr-FR.rc
```diff
@@ -5,14 +5,18 @@ STYLE DS_SHELLFONT | WS_CHILD | WS_DISABLED | WS_CAPTION
...
-    PUSHBUTTON "Par &défaut", IDC_SET_DEFAULT, 101, 159, 144, 14
-    PUSHBUTTON "A&jouter...", IDC_ADD_BUTTON, 27, 142, 70, 14
-    PUSHBUTTON "Supp&rimer", IDC_REMOVE_BUTTON, 101, 142, 70, 14
-    PUSHBUTTON "&Propriétés...", IDC_PROP_BUTTON, 175, 142, 70, 14
-    GROUPBOX "Pré&férences", -1, 7, 185, 240, 36
-    PUSHBUTTON "Para&mètres des touches...", IDC_KEY_SET_BTN, 14, 198, 110, 14
+    GROUPBOX "Default input &language", -1, 5, 5, 246, 55
+    LTEXT "Select one of the installed input languages to use when your computer starts.", -1, 15, 15, 230, 23
+    COMBOBOX IDC_DEFAULT_LANGUAGE, 15, 40, 230, 300, CBS_HASSTRINGS | CBS_AUTOHSCROLL | CBS_DROPDOWNLIST | WS_VSCROLL | WS_TABSTOP
+    GROUPBOX "&Installed services", -1, 5, 65, 246, 125
```
HBelusca, line 11:
```suggestion
    GROUPBOX "Services &installés", -1, 5, 65, 246, 125
```

dll/cpl/input/input_list.c
```diff
@@ -662,6 +680,73 @@ InputList_Create(VOID)
...
+            {
+                pMinimum = pNode;
+            }
+            else
+            {
+                if (InputList_Compare(pNode, pMinimum) < 0)
+                {
+                    pMinimum = pNode;
+                }
+            }
```
HBelusca, line 716:
```suggestion
            else if (InputList_Compare(pNode, pMinimum) < 0)
            {
                pMinimum = pNode;
            }
```

dll/cpl/input/input_list.c
```diff
@@ -662,6 +680,73 @@ InputList_Create(VOID)
...
+    INT nCompare = lstrcmpiW(pNode1->pszIndicator, pNode2->pszIndicator);
+    if (nCompare != 0)
+        return nCompare;
+
+    return lstrcmpiW(pNode1->pLayout->pszName, pNode2->pLayout->pszName);
+}
+
+VOID InputList_Sort(VOID)
+{
+    INPUT_LIST_NODE *pNode, *pNext, *pPrev, *pMinimum, *pList = _InputList;
```
HBelusca, line 694: Splitting the long line in different "units" (makes it more readable I think?)
```suggestion
    INPUT_LIST_NODE *pList = _InputList;
    INPUT_LIST_NODE *pMinimum, *pNode;
    INPUT_LIST_NODE *pNext, *pPrev;
```

dll/cpl/input/settings_page.c
```diff
@@ -394,18 +563,28 @@ BOOL EnableProcessPrivileges(LPCWSTR lpPrivilegeName, BOOL bEnable)
...
 
     switch (header->code)
     {
-        case LVN_ITEMCHANGED:
+        case TVN_SELCHANGED:
         {
-            if (header->idFrom == IDC_KEYLAYOUT_LIST)
+            SetControlsState(hwndDlg);
+        }
+        break;
```
HBelusca, line 577:
```suggestion
            break;
        }
```

dll/cpl/input/settings_page.c
```diff
@@ -394,18 +563,28 @@ BOOL EnableProcessPrivileges(LPCWSTR lpPrivilegeName, BOOL bEnable)
...
+            // https://bugs.winehq.org/show_bug.cgi?id=53727
+            NM_TREEVIEW* pTreeView = (NM_TREEVIEW*)lParam;
+            if ((pTreeView->action & TVE_TOGGLE) == TVE_COLLAPSE)
             {
-                SetControlsState(hwndDlg);
+                SetWindowLongPtrW(hwndDlg, DWLP_MSGRESULT, TRUE);
+                return TRUE;
             }
         }
         break;
```
HBelusca, line 590:
```suggestion
            break;
        }
```

dll/cpl/input/settings_page.c
```diff
@@ -432,8 +611,9 @@ OnNotifySettingsPage(HWND hwndDlg, LPARAM lParam)
...
```
HBelusca, line 612:
```suggestion
            break;
        }
```

dll/cpl/input/settings_page.c
```diff
@@ -10,195 +10,358 @@
...
         {
-            bCanDefault = FALSE;
+            item.state = item.stateMask = TVIS_BOLD;
         }
+        insert.hParent      = TVI_ROOT;
+        insert.hInsertAfter = TVI_LAST;
+        insert.item         = item;
+        hItem = TreeView_InsertItem(hwndList, &insert);
+
+        // Keyboard
```
HBelusca, line 233: Note: This is not necessarily always "Keyboard": indeed you can have "Other" (or maybe some other description): this is the "type" of the input method. "Other" is for example used for pen-driven inputs.

dll/cpl/input/settings_page.c
```diff
@@ -283,15 +426,32 @@ OnCommandSettingsPage(HWND hwndDlg, WPARAM wParam)
...
+                        SetControlsState(hwndDlg);
+                        PropSheet_Changed(GetParent(hwndDlg), hwndDlg);
+                    }
+                    else // Root?
+                    {
+                        InputList_RemoveByLang(LOWORD(item.lParam));
+                        UpdateInputListView(hwndList);
+                        SetControlsState(hwndDlg);
+                        PropSheet_Changed(GetParent(hwndDlg), hwndDlg);
+                    }
```
HBelusca, line 454: It seems this can be simplified to:
```suggestion
                    if (HIWORD(item.lParam)) // Leaf?
                        InputList_Remove((INPUT_LIST_NODE*)item.lParam);
                    else // Root?
                        InputList_RemoveByLang(LOWORD(item.lParam));

                    UpdateInputListView(hwndList);
                    SetControlsState(hwndDlg);
                    PropSheet_Changed(GetParent(hwndDlg), hwndDlg);
```

dll/cpl/input/settings_page.c
```diff
@@ -10,195 +10,358 @@
 #include "layout_list.h"
 #include "locale_list.h"
 #include "input_list.h"
+#include <shellapi.h>
```
HBelusca, line 13: For which function(s) is this header needed?

dll/cpl/input/settings_page.c
```diff
@@ -10,195 +10,358 @@
...
-                HFONT hFont;
-
-                hFont = CreateFontIndirectW(&lf);
-
-                if (hFont != NULL)
-                {
-                    HBITMAP hBmpOld;
+        StringCchCopyW(szBuf, ARRAYSIZE(szBuf), L"??");
+    }
+    szBuf[2] = UNICODE_NULL; /* Truncate the identifiers to two characters: "ENG" --> "EN" etc. */
```
HBelusca, line 41:
```suggestion
    szBuf[2] = UNICODE_NULL; /* Truncate the identifier to two characters: "ENG" --> "EN" etc. */
```

dll/cpl/input/settings_page.c
```diff
@@ -10,195 +10,358 @@
...
+
+    /* Prepare for DIB (device-independent bitmap) */
+    ZeroMemory(&bmi, sizeof(bmi));
+    bmi.bmiHeader.biSize = sizeof(bmi.bmiHeader);
+    bmi.bmiHeader.biWidth = cxIcon;
+    bmi.bmiHeader.biHeight = cyIcon;
+    bmi.bmiHeader.biPlanes = 1;
+    bmi.bmiHeader.biBitCount = 24;
```
HBelusca, line 49: I wonder whether this should always be "24", or be adjusted depending on the underlying DC (that is however currently created below), with `GetDeviceCaps(hdc, BITSPIXEL);`

dll/cpl/input/settings_page.c
katahiromz: `ExtractIconExW`.

HBelusca: Isn't it possible to use LoadImageW or LoadIconW instead? Those two functions are in user32 and this would allow not getting an otherwise useless dependency on the shell. It looks to me those functions also allow a path to a file to be used.

learn-more: Correct, that Isn't possible.
`ExtractIconExW` is doing a lot more than just calling `LoadIconW`

dll/cpl/input/settings_page.c
katahiromz: The 24-bpp DIB (device-independent bitmap) is a full color bitmap. So it doesn't need the color table if conversion is correct. The DDB (device-dependent bitmap) has a fear that conversion might be failed if the device was changed.

dll/cpl/input/settings_page.c
katahiromz: Not fixed yet in the latest WineHQ.
```c
    if (bUser)
	TREEVIEW_SendExpanding(infoPtr, item, action);
```

dll/cpl/input/input_list.c
```diff
@@ -662,6 +680,73 @@ InputList_Create(VOID)
     free(pLayoutList);
 }
 
+static INT InputList_Compare(INPUT_LIST_NODE *pNode1, INPUT_LIST_NODE *pNode2)
+{
+    INT nCompare = lstrcmpiW(pNode1->pszIndicator, pNode2->pszIndicator);
+    if (nCompare != 0)
+        return nCompare;
+
+    return lstrcmpiW(pNode1->pLayout->pszName, pNode2->pLayout->pszName);
```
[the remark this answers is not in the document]
katahiromz: Fixed.

dll/cpl/input/input_list.c
```diff
@@ -662,6 +680,73 @@ InputList_Create(VOID)
...
+        {
+            pMinimum->pPrev = pMinimum->pNext = NULL;
+            _InputList = pMinimum;
+        }
+        else
+        {
+            /* Find last node */
+            for (pNode = _InputList; pNode->pNext; pNode = pNode->pNext)
+            {
+                ;
```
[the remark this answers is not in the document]
katahiromz: `NOTHING` is not defined. I defined it in this file.

dll/cpl/input/settings_page.c
katahiromz: Fixed.

dll/cpl/input/settings_page.c
katahiromz: Fixed.

dll/cpl/input/settings_page.c
```diff
@@ -10,195 +10,358 @@
...
+    hItem = TreeView_GetRoot(hwndList);
+    while (hItem)
+    {
+        szText[0] = 0;
+        item.mask       = TVIF_TEXT | TVIF_HANDLE;
+        item.pszText    = szText;
+        item.cchTextMax = _countof(szText);
+        item.hItem      = hItem;
+        TreeView_GetItem(hwndList, &item);
+        if (lstrcmpi(szText, pszLangName) == 0)
```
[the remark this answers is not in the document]
katahiromz: `_wcsicmp` is used.

dll/cpl/input/settings_page.c
katahiromz: `InputList_GetAliveCount` function is added.

katahiromz: `InputList_GetAliveCount` function is added.

dll/cpl/input/settings_page.c
katahiromz: Fixed.
black

after

dll/cpl/input/settings_page.c
katahiromz: Fixed. Unlinked `shell32`.

dll/cpl/input/input_list.c
katahiromz: Fixed.

dll/cpl/input/input_list.c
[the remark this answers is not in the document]
HBelusca: Shouldn't it be in the winnt.h? I don't remember :thinking:
Otherwise the loop could be: `pNode = _InputList; while (pNode->pNext) { pNode = pNode->pNext; }`
Anyway, I'm ok with the define as well.

dll/cpl/input/input_list.c
katahiromz: Fixed.

dll/cpl/input/settings_page.c
katahiromz: Fixed.

dll/cpl/input/settings_page.c
katahiromz: Fixed.

dll/cpl/input/settings_page.c
katahiromz: Fixed.

dll/cpl/input/lang/en-US.rc
katahiromz: I think setting the current default input language might need reboot.

dll/cpl/input/settings_page.c
katahiromz: Comments fixed.

dll/cpl/input/settings_page.c
katahiromz: See also: https://github.com/reactos/reactos/pull/4723#discussion_r981796819

PR #4727, opened 2022-09-25
Ddraw fullscreen fix

masnuvahaque:
## Purpose

This gets the taskbar to not show when running a DirectDraw based program in Full-Screen Mode.

JIRA issue: [CORE-11242](https://jira.reactos.org/browse/CORE-11242)

This fix is taken from code found in https://source.winehq.org/git/wine.git/commitdiff/84413298de371a9e9b9ecfe0dd01f580b7521222


JoachimHenze:
According to "theDarkFire" this patch broke its effectiveness by 0.4.15-dev-5382-ga2c6af0 https://github.com/reactos/reactos/commit/a2c6af0da4acbba9c254d003e9d9f4ea6e03ed63 . I have not double-checked that claim yet, but found it worth to add that information here.

PR #4730, opened 2022-09-27
[NTUSER] IntVkToVsc should return zero if first parameter is zero

julcar:
## Purpose

Investigated using ApiMonitor, removes the "F17" key shortcut in menu items. Fixes CORE-17906 and should fix CORE-3903 too.

JIRA issue: [CORE-17906](https://jira.reactos.org/browse/CORE-17906)

### DIFF
--- a/win32ss/user/ntuser/keyboard.c
+++ b/win32ss/user/ntuser/keyboard.c
@@ -1270,6 +1270,9 @@ IntMapVirtualKeyEx(UINT uCode, UINT Type, PKBDTABLES pKbdTbl)
 {
     UINT uRet = 0;
 
+    if (uCode == 0)
+        return uRet;
+
     switch (Type)
     {
         case MAPVK_VK_TO_VSC:
--- a/win32ss/user/ntuser/keyboard.c
+++ b/win32ss/user/ntuser/keyboard.c
@@ -528,14 +528,11 @@ IntVkToVsc(WORD wVk, PKBDTABLES pKbdTbl)
 
     ASSERT(pKbdTbl);
 
-    if (wVk == 0)
-        return 0;
-
     /* Check standard keys first */
     for (i = 0; i < pKbdTbl->bMaxVSCtoVK; i++)
     {
         if ((pKbdTbl->pusVSCtoVK[i] & 0xFF) == wVk)
-            return i;
+            return pKbdTbl->pusVSCtoVK[i] & 0xFF;
     }
 
     /* Check extended keys now */
--- a/win32ss/user/ntuser/keyboard.c
+++ b/win32ss/user/ntuser/keyboard.c
@@ -528,6 +528,9 @@ IntVkToVsc(WORD wVk, PKBDTABLES pKbdTbl)
 
     ASSERT(pKbdTbl);
 
+    if (wVk == 0)
+        return 0;
+
     /* Check standard keys first */
     for (i = 0; i < pKbdTbl->bMaxVSCtoVK; i++)
     {
@@ -1269,9 +1272,6 @@ static UINT
 IntMapVirtualKeyEx(UINT uCode, UINT Type, PKBDTABLES pKbdTbl)
 {
     UINT uRet = 0;
-
-    if (uCode == 0)
-        return uRet;
 
     switch (Type)
     {


win32ss/user/ntuser/keyboard.c
```diff
@@ -1270,6 +1270,9 @@ IntMapVirtualKeyEx(UINT uCode, UINT Type, PKBDTABLES pKbdTbl)
 {
     UINT uRet = 0;
 
+    if (uCode == 0)
+        return uRet;
```
binarymaster, line 1274: Is this correct for all types, i.e. does not depend on `Type` value? Is there an API test for this which we can compare with Windows?

binarymaster (commented):
MapVirtualKeyW / MapVirtualKeyExW => NtUserMapVirtualKeyEx => IntMapVirtualKeyEx

This Winetest does use `MapVirtualKeyW` function, you can modify it to test different cases when `uCode` is equal to zero: https://git.reactos.org/?p=reactos.git;a=blob;f=modules/rostests/winetests/user32/input.c;hb=22d8c0fd541eaf92b3462ae1d252cdbba16e30db

win32ss/user/ntuser/keyboard.c
julcar: Is non-sense to attempt to translate a VK of value zero, tested with API Monitor in windows XP using the application that shows the glitch

julcar:
> MapVirtualKeyW / MapVirtualKeyExW => NtUserMapVirtualKeyEx => IntMapVirtualKeyEx
> 
> This Winetest does use `MapVirtualKeyW` function, you can modify it to test different cases when `uCode` is equal to zero: https://git.reactos.org/?p=reactos.git;a=blob;f=modules/rostests/winetests/user32/input.c;hb=22d8c0fd541eaf92b3462ae1d252cdbba16e30db

Please test the executable file I've attached to the jira report CORE-17906

binarymaster:
@julcar what is the problem with writing API tests?

binarymaster:
Also your test written in Visual Basic does not test all possible values.

win32ss/user/ntuser/keyboard.c
binarymaster: Ok I'm going to make API test for you in a few days.

julcar: According to: 
https://learn.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes
Values start from 0x01, so does not make any sense to attempt to translate 0x0

julcar:
By the way, here is the code portion that triggers the glitch in Sqliteadmin:
https://github.com/mpryanikov/sourseide-delphi7/blob/263796ccf7b0e8cfd51b4bfdacfb6c9ca511aec6/Source/Vcl/Menus.pas#L492
It is clear that the function expects a non-zero value from MapVirtualKey() to make the call to GetKeyNameText()

HBelusca:
> It is clear that the function expects a non-zero value from MapVirtualKey() to make the call to GetKeyNameText()

It is clearer that MapVirtualKey can return zero, and then, GetKeyNameText is going to be called only if the returned scancode was not zero.

binarymaster (commented):
Here are the test results for before and after:

## Before

VirtualBox_ReactOS_02_10_2022_04_37_01

## After

VirtualBox_ReactOS_02_10_2022_04_38_32

win32ss/user/ntuser/keyboard.c
binarymaster: So according to test results, you need to investigate `IntVkToVsc` function - it should not return `128` scan code for a `0` virtual key.

win32ss/user/ntuser/keyboard.c
```diff
@@ -532,7 +532,7 @@ IntVkToVsc(WORD wVk, PKBDTABLES pKbdTbl)
     for (i = 0; i < pKbdTbl->bMaxVSCtoVK; i++)
     {
         if ((pKbdTbl->pusVSCtoVK[i] & 0xFF) == wVk)
-            return i;
+            return pKbdTbl->pusVSCtoVK[i] & 0xFF;
```
binarymaster, line 535: This introduces a bug :) Explained in the chat

binarymaster (commented):
Interesting: https://git.reactos.org/?p=reactos.git;a=blob;f=dll/keyboard/kbdus/kbdus.c;h=ef103cdc1b2df9bf1071d06a6487a63580293073;hb=12526424d5e44865e55478436763bfbfc4b36fc7

I just followed @HBelusca advice to look into our keyboard layout DLLs and here is it:
```c
#define VK_EMPTY  0xff   /* The non-existent VK */

...

ROSDATA USHORT scancode_to_vk[] = {
  /* Numbers Row */
  /* - 00 - */
  /* 1 ...         2 ...         3 ...         4 ... */
  VK_EMPTY,     VK_ESCAPE,    '1',          '2',
```

And it's placed on the zero scan code position, and then...

```c
  /* - 80 - */
  0
};
```

There is zero right at the 128 position, which makes the VkToVsc function return that 128 for a 0.

binarymaster (commented):
Finally, I was able to confirm that with replacing our `kbd*.dll` keyboard layout files with the files from Win2k3 an offending test failure is gone!

So there is no problem on win32ss/ntuser side. See in chat: https://chat.reactos.org/reactos/pl/637q435313g4xfa9b9ka3jmosh

PR #4734, opened 2022-09-30
[OSK] Release Super (Windows) key on exit

silvicky:
## Purpose

Fix the bug that ROS(windows) key is still pressed on exit.

## Proposed changes

While closing the window, press ctrl(to prevent opening main menu), release two ROS keys then release ctrl.

### DIFF
--- a/base/applications/osk/main.c
+++ b/base/applications/osk/main.c
@@ -495,17 +495,17 @@ int OSK_Close(void)
     OSK_ReleaseKey(SCAN_CODE_44); // Left shift
     OSK_ReleaseKey(SCAN_CODE_57); // Right shift
 
-    /* As releasing ROS calls main menu, press ctrl first */
+    /* As releasing Super key opens Start menu, press ctrl first */
     INPUT Input;
     Input.type = INPUT_KEYBOARD;
     Input.ki.wVk = 0;
-    Input.ki.wScan = SCAN_CODE_58;
+    Input.ki.wScan = SCAN_CODE_58; // Left ctrl
     Input.ki.time = GetTickCount();
     Input.ki.dwExtraInfo = GetMessageExtraInfo();
     Input.ki.dwFlags = KEYEVENTF_SCANCODE;
     SendInput(1, &Input, sizeof(Input));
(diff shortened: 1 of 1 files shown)


silvicky:
I wrote like this because make a real call to OSK_Command is too complex and unnecessary.
Also, as the "press" is fake, I can't use OSK_ReleaseKey to release it.

silvicky:
I took the name from keyboard.c so should we also change the names inside as such?

binarymaster:
> I took the name from keyboard.c so should we also change the names inside as such?

At some point, I think yes. Not in this PR though :slightly_smiling_face:

binarymaster (commented):
I'm ok with this change, however I'm not qualified enough to approve it, because I didn't investigated how Windows' OSK works.

peterooch:
> I'm ok with this change, however I'm not qualified enough to approve it, because I didn't investigated how Windows' OSK works.

I did investigate (while trying to fix other osk issues) the windows osk doesn't do the ctrl hold 'hack' (I assume the extra code with `SendInput` is that), so if superkey is being held and osk is closed then it's released and the start menu opens. (tested on win2003 osk and win10 osk)

peterooch:
it's 50/50, the superkey release is a proper fix as I also encountered that bug, the ctrl hold can be called a hackfix

silvicky:
> > I'm ok with this change, however I'm not qualified enough to approve it, because I didn't investigated how Windows' OSK works.
> 
> I did investigate (while trying to fix other osk issues) the windows osk doesn't do the ctrl hold 'hack' (I assume the extra code with `SendInput` is that), so if superkey is being held and osk is closed then it's released and the start menu opens. (tested on win2003 osk and win10 osk)

Not true in latest win11 though, so it's a really old bug that is fixed just now.

KRosUser2:
@silvicky : Does this fix makes to situation better regarding OSK defects like https://jira.reactos.org/browse/CORE-12431 or https://jira.reactos.org/browse/CORE-18529 ?

KRosUser2:
@silvicky : did you test against these 2 tickets ?

PR #4735, opened 2022-09-30
[FASTFAT][VFATFS] Rename FF directory, Restrict VFAT compilation

SergeGautherie:
## Purpose

2 minor additional changes, after 14c3936 (0.4.15-dev-5090).

JIRA issue: [CORE-16373](https://jira.reactos.org/browse/CORE-16373)

## Proposed changes

- [FASTFAT] Rename its source directory from fastfat_new
- [VFATFS] Compile it for xbox only


drivers/filesystems/CMakeLists.txt
```diff
@@ -2,12 +2,14 @@
...
+add_subdirectory(fastfat)
 add_subdirectory(fs_rec)
 add_subdirectory(msfs)
 add_subdirectory(mup)
 add_subdirectory(nfs)
 add_subdirectory(npfs)
 add_subdirectory(ntfs)
 add_subdirectory(udfs)
-add_subdirectory(vfatfs)
+if(SARCH STREQUAL "xbox")
```
HBelusca, line 13: This was made to compile for everything also for a reason....

binarymaster: Double that.

binarymaster (commented):
I don't see a reason for any changes in this PR. If nobody objects, I will close it in 5 days.

HBelusca:
The fastfat_new rename however is OK I think.

drivers/filesystems/CMakeLists.txt
SergeGautherie: > This was made to compile for everything also for a reason....

What is that reason?
It seems to me to be wasting resources/time on usual builds.

binarymaster: Judging from that logic we shouldn't add new code and new modules to avoid getting build times bigger. As for the reason, look at discussion/review of the FATX PR.

SergeGautherie: > Judging from that logic we shouldn't add new code and new modules to avoid getting build times bigger.

I never wrote such a thing.
My point is: why compile something which is not going to be packaged?

> As for the reason, look at discussion/review of the FATX PR.

That PR seems to be  #4694.
Always-compile was explicitly intentional.
Nonetheless, the reason is still not obvious to me.

Anyway...

SergeGautherie:
> The fastfat_new rename however is OK I think.

Superseded by 23b7c7b8234e48ed35ce7c40ce80a04d43432d3d.

drivers/filesystems/CMakeLists.txt
JoachimHenze: I do concur that it would make sense to exclude that from the build when not needed.

JoachimHenze: I don't want to fight neither. I just expressed my vote here, concuring Serges. I did neither reopen anything nor created a PR myself. It is jsut expressing my opinion. Which is as good or as bad as anybody elses opinion. In FOSS projects I do perceive that as important, to let democracy / swarm-intelligence find the best solutions/answers.

binarymaster: As Xbox port maintainer I'd say... ideally ReactOS should be able to mount, read, and write FATX partitions regardless of being booted from Xbox or not.

PR #4737, opened 2022-10-01
[WLANCONF] Fix getting the interface GUID value

binarymaster:
**JIRA issue:** [**CORE-18032**](https://jira.reactos.org/browse/CORE-18032)

## Description

GetInterfaceInfo returns interface name in Windows XP and 2003 in this format: `\DEVICE\TCPIP_{GUID}`.

MSDN says that the `Name` member of the `IP_ADAPTER_INDEX_MAP` may start with '{' character on Windows Vista and later.

https://docs.microsoft.com/en-us/windows/win32/api/ipexport/ns-ipexport-ip_adapter_index_map

Change the code to support both cases.

## Result

Tested with RTL8187-based USB adapter:

VirtualBox_ReactOS_28_09_2022_19_47_09

### DIFF
--- a/base/applications/network/wlanconf/wlanconf.c
+++ b/base/applications/network/wlanconf/wlanconf.c
@@ -1,9 +1,8 @@
 /*
  * PROJECT:     ReactOS WLAN command-line configuration utility
- * LICENSE:     GPL - See COPYING in the top level directory
- * FILE:        base/applications/network/wlanconf/wlanconf.c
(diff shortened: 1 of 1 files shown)


HBelusca (approved):

obsolete-is-better:
this patch seems to be working for me too
wlanconf

obsolete-is-better (approved):

PR #4739, opened 2022-10-01
[NTOS:CC] Let Cc make some space in case of memory pressure

zefklop:
## Purpose

Have lazy writer leave some space after it's done writing in case we badly need it to.
Also, free one VACB when we need to make some space for a new one instead of freeing them all. Seriously...

### DIFF
--- a/ntoskrnl/cc/copy.c
+++ b/ntoskrnl/cc/copy.c
@@ -381,6 +381,16 @@ CcCanIWrite (
         return FALSE;
     }
 
+    /* Otherwise, if there are no deferred writes yet, start the lazy writer */
+    if (IsListEmpty(&CcDeferredWrites))
+    {
+        KIRQL OldIrql;
+
+        OldIrql = KeAcquireQueuedSpinLock(LockQueueMasterLock);
+        CcScheduleLazyWriteScan(TRUE);
+        KeReleaseQueuedSpinLock(LockQueueMasterLock, OldIrql);
+    }
+
     /* Initialize our wait event */
     KeInitializeEvent(&WaitEvent, NotificationEvent, FALSE);
 
@@ -407,12 +417,6 @@ CcCanIWrite (
                                     &Context.DeferredWriteLinks,
                                     &CcDeferredWriteSpinLock);
     }
-
-    /* Now make sure that the lazy scan writer will be active */
-    OldIrql = KeAcquireQueuedSpinLock(LockQueueMasterLock);
-    if (!LazyWriter.ScanActive)
-        CcScheduleLazyWriteScan(TRUE);
-    KeReleaseQueuedSpinLock(LockQueueMasterLock, OldIrql);
 
 #if DBG
     DPRINT1("Actively deferring write for: %p\n", FileObject);
--- a/ntoskrnl/cc/view.c
+++ b/ntoskrnl/cc/view.c
@@ -595,34 +595,26 @@ CcRosUnmarkDirtyVacb (
     }
 }
 
-static
 BOOLEAN
-CcRosFreeUnusedVacb (
-    PULONG Count)
+CcRosFreeOneUnusedVacb (
+    VOID)
 {
-    ULONG cFreed;
-    BOOLEAN Freed;
     KIRQL oldIrql;
-    PROS_VACB current;
-    LIST_ENTRY FreeList;
     PLIST_ENTRY current_entry;
-
-    cFreed = 0;
-    Freed = FALSE;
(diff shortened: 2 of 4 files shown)


ntoskrnl/cc/view.c
```diff
@@ -595,34 +595,26 @@ CcRosUnmarkDirtyVacb (
     }
 }
 
-static
 BOOLEAN
-CcRosFreeUnusedVacb (
-    PULONG Count)
+CcRosFreeOneUnusedVacb (
+    void)
```
HBelusca, line 600:
```suggestion
CcRosFreeOneUnusedVacb(VOID)
```

ntoskrnl/cc/view.c
```diff
@@ -634,47 +626,32 @@ CcRosFreeUnusedVacb (
             ASSERT(!current->MappedCount);
             ASSERT(Refs == 1);
 
-            /* Reset and move to free list */
+            /* Reset it, this is he one we want to free */
```
HBelusca, line 629:
```suggestion
            /* Reset it, this is the one we want to free */
```

ntoskrnl/cc/view.c
```diff
@@ -711,30 +687,27 @@ CcRosCreateVacb (
...
+         * If it succeeds, retry to map, otherwise just fail.
+         */
+        if (!CcRosFreeOneUnusedVacb())
+        {
+            ExFreeToNPagedLookasideList(&VacbLookasideList, current);
+            return Status;
+        }
     }
 
+
```
HBelusca, line 710:
```suggestion
```

ntoskrnl/include/internal/cc.h
```diff
@@ -509,3 +509,7 @@ CcRosVacbDecRefCount(
 }
 #define CcRosVacbGetRefCount(vacb) InterlockedCompareExchange((PLONG)&(vacb)->ReferenceCount, 0, 0)
 #endif
+
+BOOLEAN
+CcRosFreeOneUnusedVacb (
+    void);
```
HBelusca, line 515:
```suggestion
CcRosFreeOneUnusedVacb(VOID);

```

HBelusca (commented):
Interesting PR :)

ntoskrnl/cc/view.c
zefklop: I'l uppercase the void but I'll leave the indentation, as that follows the surrounding code.

ntoskrnl/include/internal/cc.h
zefklop: Same. Uppercase void, left indentation.

JoachimHenze:
I cannot say whether the PR is bad or good, when reviewing the code,
but what I can definitely say, is, that it does not fix https://jira.reactos.org/browse/CORE-17624 yet.
See the test outcome https://jira.reactos.org/browse/CORE-17624?focusedCommentId=134760&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-134760
Needs more work still, either in here, or in another PR.

JoachimHenze:
For the record: the Darkfire reported here, that this PR allowed him booting with 36MB RAM just.
https://chat.reactos.org/reactos/pl/ge4se91hcjfzpnoffijmzdsteh
That would be a major achievement without that file-copy-improvement even, and would transform this PR into a valuable gift nevertheless. But it should not link against CORE-17624 then, (but maybe better against https://jira.reactos.org/browse/CORE-18010 then????)
Testing that myself right now....

ntoskrnl/cc/copy.c
```diff
@@ -381,6 +381,16 @@ CcCanIWrite (
         return FALSE;
     }
 
+    /* Otherwise, if there are no deferred writes yet, start the lazy writer */
+    if (IsListEmpty(&CcDeferredWrites))
+    {
+        KIRQL OldIrql;
+
+        OldIrql = KeAcquireQueuedSpinLock(LockQueueMasterLock);
+        CcScheduleLazyWriteScan(TRUE);
```
ThFabba, line 390: The rules around this aren't clear to me. If you understand them, could you explain (and ideally, document in the code)?

* Do we need to check `LazyWriter.ScanActive`?
* What is the master lock protecting here?
* Don't we need to hold `CcDeferredWriteSpinLock` before checking `CcDeferredWrites`?
* Is this call broken since it _doesn't_ acquire the master lock? https://github.com/reactos/reactos/pull/4739/files#diff-e76ba4c26f0c04317701b9f1891af3634bfce59f91c80468c469eb18e58432a3L189

PR #4742, opened 2022-10-01
[SYSSETUP] Add support for shortcut arguments

learn-more:
+ [SHORTCUTS] Create the readme shortcut with readme as argument


HBelusca (approved):

media/inf/shortcuts.inf
```diff
@@ -15,7 +15,7 @@ GamesShortcuts=2, %GAMES%
 QuickLaunchShortcuts=26, Microsoft\Internet Explorer\Quick Launch
 
 [DesktopShortcuts]
-%SystemRoot%\readme.txt, %README_TITLE%, %README_DESC%
+%SystemRoot%\notepad.exe, %README_TITLE%, %README_DESC%, 0, %SystemRoot%, readme.txt
```
binarymaster, line 18: Hmm, I wonder if the shortcut now will have the Notepad icon instead of TXT-file icon :slightly_smiling_face:

binarymaster (approved):
Nice ^^)

media/inf/shortcuts.inf
learn-more: image

binarymaster: Yes indeed, it has changed. Here is how it used to look:

image

HBelusca: That's why, when I added this shortcut back in the day, I made it against readme.txt directly.

PR #4743, opened 2022-10-01
[WINLOGON] Protect function calls to '3rd-party' DLLs by SEH.

HBelusca:
## Purpose

3rd-party code (e.g. Gina notification handlers DLLs), or related code (e.g. winmm.dll when playing sound events via sound files) can run in the context of WinLogon.exe. If this code crashes (because it contains a bug, or plays a corrupted sound file, etc.), then WinLogon would crash, and since this is a marked system process, this would trigger a BSoD.

Protect calls to this 3rd-party code by SEH.
(Remark: LoadLibrary calls, that run dll's DllMain() functions, don't need to be under SEH, since LoadLibrary already handles that. Thanks to Mark for noticing this.)

## Proposed changes

This includes:
- Notification dll calling in CallNotificationDll().
- winmm.dll API calling (e.g. PlaySound) in PlaySoundRoutine().

Also:
- Fix dwKeyName usage in RegEnumKeyExW() specifying a number of *characters*.

### DIFF
--- a/base/system/winlogon/notify.c
+++ b/base/system/winlogon/notify.c
@@ -278,7 +278,7 @@ InitNotifications(VOID)
     dwIndex = 0;
     for(;;)
     {
-        dwKeyName = 80 * sizeof(WCHAR);
+        dwKeyName = ARRAYSIZE(szKeyName);
         lError = RegEnumKeyExW(hNotifyKey,
                                dwIndex,
                                szKeyName,
@@ -312,11 +312,8 @@ CallNotificationDll(
     NOTIFICATION_TYPE Type,
     PWLX_NOTIFICATION_INFO pInfo)
 {
-    HKEY hDllKey = NULL;
-    HMODULE hModule = NULL;
+    HMODULE hModule;
     CHAR szFuncBuffer[128];
-    DWORD dwSize;
-    DWORD dwType;
     DWORD dwError = ERROR_SUCCESS;
     PWLX_NOTIFY_HANDLER pNotifyHandler;
 
@@ -338,6 +335,10 @@ CallNotificationDll(
     }
     else
     {
+        HKEY hDllKey;
+        DWORD dwSize;
+        DWORD dwType;
+
         dwError = RegOpenKeyExW(hNotifyKey,
                                 NotificationDll->pszKeyName,
                                 0,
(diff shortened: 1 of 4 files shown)


base/system/winlogon/notify.c
```diff
@@ -356,23 +357,38 @@ CallNotificationDll(
...
 
-    if (dwError == ERROR_SUCCESS)
+    if (dwError != ERROR_SUCCESS)
+        return;
+
+    _SEH2_TRY
     {
-        hModule = LoadLibraryW(NotificationDll->pszDllName);
-        if (hModule != NULL)
+        _SEH2_TRY
```
binarymaster, line 369: Wow, nested SEH2, now I've seen everything :smile:

HBelusca: Yes, this exists (see also msdn). That's when you want to have both the "finally" and the "except" functionality of SEH at the same time.

learn-more:
Are you sure this is what winlogon is doing?

base/system/winlogon/notify.c
learn-more: It should not be needed around LoadLibraryW, because that itself already uses SEH.
https://git.reactos.org/?p=reactos.git;a=blob;f=dll/ntdll/ldr/ldrapi.c;hb=cd2d284142c1fcaefc9759412c51c6d2fb2ca461#l356
and
https://git.reactos.org/?p=reactos.git;a=blob;f=dll/ntdll/ldr/ldrutils.c;hb=cd2d284142c1fcaefc9759412c51c6d2fb2ca461#l2497

HBelusca: Fixed!

base/system/winlogon/notify.c
```diff
@@ -356,23 +357,31 @@ CallNotificationDll(
...
 
-            FreeLibrary(hModule);
-        }
+    hModule = LoadLibraryW(NotificationDll->pszDllName);
+    if (!hModule)
+        return;
+
+    _SEH2_TRY
+    {
+        pNotifyHandler = (PWLX_NOTIFY_HANDLER)GetProcAddress(hModule, szFuncBuffer);
```
learn-more, line 373: nit: I would move this outside the SEH block.

learn-more (approved):

binarymaster (approved):

base/system/winlogon/notify.c
HBelusca: Moved :+1:

JoachimHenze:
> Are you sure this is what winlogon is doing?
HBelusca you reacted with thumbsup to Marks question. Does that mean: "Yes, I know exactly that this is what Windows does!" or does it mean "Thumbs up! You hit the nail! I know that this is not what Windows does, but it is okay for now to prevent us from BSOD'ing here."
I will not block your PR whatever you will reply to this question. I just want to know.

HBelusca:
> I just want to know.

It means that I have discussed in private about that with him, and confirmed that Windows' winlogon.exe does protect some API calls to 3rd-party loaded DLLs with SEH (which is obvious, from a security standpoint; and if it happened to not do that, it would have been a bug on their part).

learn-more (approved):

PR #4745, opened 2022-10-03
Minor cleanup for EventVwr

HBelusca:
## Purpose

- Update file headers.
- Simplify some duplicated code.

### DIFF
--- a/base/applications/mscutils/eventvwr/eventvwr.c
+++ b/base/applications/mscutils/eventvwr/eventvwr.c
@@ -33,8 +33,8 @@ static const LPCWSTR SystemLogs[] =
 };
 
 /* MessageFile message buffer size */
-#define EVENT_MESSAGE_EVENTTEXT_BUFFER  1024*10                             // NOTE: Used by evtdetctl.c
-#define EVENT_MESSAGE_FILE_BUFFER       1024*10
+#define EVENT_MESSAGE_EVENTTEXT_BUFFER  (1024*10)                           // NOTE: Used by evtdetctl.c
+#define EVENT_MESSAGE_FILE_BUFFER       (1024*10)
 #define EVENT_DLL_SEPARATOR             L";"
 #define EVENT_CATEGORY_MESSAGE_FILE     L"CategoryMessageFile"
 #define EVENT_MESSAGE_FILE              L"EventMessageFile"
@@ -2034,7 +2034,7 @@ EnumEventsThread(IN LPVOID lpParameter)
         // (EventLogFilter->NumOfEventLogs > 1)
         MessageBoxW(hwndMainWindow,
                     L"Many-logs filtering is not implemented yet!!",
-                    L"Event Log",
+                    szTitle,
                     MB_OK | MB_ICONINFORMATION);
     }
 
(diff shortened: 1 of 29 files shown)


base/applications/mscutils/eventvwr/eventvwr.c
```diff
@@ -1,29 +1,10 @@
...
- * PURPOSE:         Event Log Viewer main file
- * PROGRAMMERS:     Marc Piulachs (marc.piulachs at codexchange [dot] net)
- *                  Eric Kohl
- *                  Hermes Belusca-Maito
+ * PROJECT:     ReactOS Event Log Viewer
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Event Log Viewer main file.
+ * COPYRIGHT:   Copyright 2007 Marc Piulachs (marc.piulachs at codexchange [dot] net)
+ *              Copyright 2008-2016 Eric Kohl
+ *              Copyright 2016-2022 Hermès Bélusca-Maïto
```
binarymaster, line 7: I'd go without an ending dot in the 'PURPOSE' field. Also missing emails for last two authors :)
```suggestion
 * PURPOSE:     Event Log Viewer main file
 * COPYRIGHT:   Copyright 2007 Marc Piulachs <marc.piulachs@codexchange.net>
 *              Copyright 2008-2016 Eric Kohl <eric.kohl@reactos.org>
 *              Copyright 2016-2022 Hermès Bélusca-Maïto <hermes.belusca-maito@reactos.org>
```

HBelusca: What if I (or others) don't want to specify their emails? :oof:

JoachimHenze (approved):

base/applications/mscutils/eventvwr/evtdetctl.c
```diff
@@ -878,20 +878,15 @@ EventDetailsCtrl(HWND hDlg, UINT uMsg, WPARAM wParam, LPARAM lParam)
...
                 case IDC_BYTESRADIO:
-                    if (pData->EventLogFilter)
-                    {
-                        pData->bDisplayWords = FALSE;
-                        DisplayEventData(hDlg, pData->bDisplayWords);
-                    }
-                    return (INT_PTR)TRUE;
-
                 case IDC_WORDRADIO:
+                {
```
JoachimHenze, line 882: Personally I am no fan of introducing such curly-braces for switch-cases unless there is some variable declaration within that, which is not the case here. But I guess this is a matter of personal preference.
Same applies for the other places where you added those.

learn-more (approved):

base/applications/mscutils/eventvwr/eventvwr.c
binarymaster: This is not an option, according to our coding style: https://reactos.org/wiki/Coding_Style

HBelusca: Note that nothing says there whether email is mandatory or not. It just gives an example (in case there is a mail).

binarymaster: Alright, but still I would prefer to stay uniform here. I had a similar discussion with Hervé, he didn't wanted to put an email because of too much spam coming to the inbox, but I don't see this as a solution of the problem.

HBelusca: Yes this was exactly my intent. (Especially since I also don't consult reactos.org on a regular basis, but more like twice a year.)

JoachimHenze: I do agree to binarymasters argumentation that it is better to omit the fullstop on the purpose line. Here and in the other files as well. But I won't block your PR just for that reason.

base/applications/mscutils/eventvwr/evtdetctl.c
HBelusca: I put brackets also for case blocks when they start to become large (>= 4 lines) as this allows to fold them in any code editor.

binarymaster (approved):

PR #4746, opened 2022-10-03
[EVENTVWR] Improve behaviour on event item selection.

HBelusca:
## Purpose

When clicking outside a selected event item in the list, or right clicking on a non-selected line,
we used to get "No Items in ListView" error.
This error came from a check for code paths that should, in principle, be not taken if no event item is selected.

JIRA issue: [CORE-18438](https://jira.reactos.org/browse/CORE-18438)

## Proposed changes

Rework how some aspects of item selection retrieval is done.

- Each event detail control stores its own "current" item index, so
  that there can be different event detail dialogs showing different
  events concurrently (half-plemented; this is a Win7-like feature).
  As such, give the index of the selected event item when sending
  the EVT_DISPLAY message, instead of having the details dialog
  retrieve everytime by itself the current selected item (that may
  change in-between calls, and can trigger the "No Items in ListView"
  error).

- When pressing "Prev"/"Next" buttons, detect whether we already are
  at the top/bottom of the event log, and if so, prompt the user to
  continue around.
  Clear up any selected event in the list, before selecting the new
  one. (Note: the event list supports multiple selection, for future
  functionality.)

### DIFF
--- a/base/applications/mscutils/eventvwr/eventvwr.c
+++ b/base/applications/mscutils/eventvwr/eventvwr.c
@@ -3362,16 +3362,44 @@ WndProc(HWND hWnd, UINT uMsg, WPARAM wParam, LPARAM lParam)
                         if ( (pnmv->uChanged  & LVIF_STATE) && /* The state has changed */
                              (pnmv->uNewState & LVIS_SELECTED) /* The item has been (de)selected */ )
                         {
-                            if (hwndEventDetails)
-                                SendMessageW(hwndEventDetails, EVT_DISPLAY, 0, 0);
+                            if (!hwndEventDetails)
+                                break;
+
+                            /* Verify the index of selected item */
+                            if (pnmv->iItem == -1)
+                            {
+                                MessageBoxW(hWnd,
+                                            L"No selected items!",
+                                            szTitle,
+                                            MB_OK | MB_ICONERROR);
+                                break;
+                            }
+                            SendMessageW(hwndEventDetails, EVT_DISPLAY, 0, (LPARAM)pnmv->iItem);
                         }
                         break;
                     }
 
+#ifdef LVN_ITEMACTIVATE
+                    case LVN_ITEMACTIVATE:
+                    {
+                        /* Get the index of the single focused selected item */
+                        LPNMITEMACTIVATE lpnmitem = (LPNMITEMACTIVATE)lParam;
+                        INT iItem = lpnmitem->iItem;
+                        if (iItem != -1)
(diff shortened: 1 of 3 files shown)


base/applications/mscutils/eventvwr/eventvwr.c
```diff
@@ -3381,16 +3381,42 @@ WndProc(HWND hWnd, UINT uMsg, WPARAM wParam, LPARAM lParam)
...
+                                            szTitle,
+                                            MB_OK | MB_ICONERROR);
+                                break;
+                            }
+                            SendMessageW(hwndEventDetails, EVT_DISPLAY, 0, (LPARAM)pnmv->iItem);
                         }
                         break;
                     }
 
+                #ifdef LVN_ITEMACTIVATE
```
JoachimHenze, line 3401: I would recommend to place all #ifdef, #else and #endif in your fix leftmost, at column0.

base/applications/mscutils/eventvwr/eventvwr.c
```diff
@@ -3381,16 +3381,42 @@ WndProc(HWND hWnd, UINT uMsg, WPARAM wParam, LPARAM lParam)
...
+                        LPNMITEMACTIVATE lpnmitem = (LPNMITEMACTIVATE)lParam;
+                        INT iItem = lpnmitem->iItem;
+                #else
+                        INT iItem = ListView_GetNextItem(hwndListView, -1, LVNI_FOCUSED | LVNI_SELECTED);
+                #endif // LVN_ITEMACTIVATE
+
+                        if (iItem != -1)
+                            SendMessageW(hWnd, WM_COMMAND, IDM_EVENT_DETAILS, (LPARAM)iItem);
                         break;
+                    }
```
learn-more, line 3419: I would split this in 2 different distinct cases.
Having the LVN_ITEMACTIVATE and NM_DBLCLICK code intermingled is not making it any clearer, and you don't combine any big code blocks either.

HBelusca: Actually I'm conditionally compiling the code here. If we support LVN_ITEMACTIVATE then I'm using it, otherwise it would fallback to handling NM_DBLCLK/NM_RETURN.

HBelusca: Fixed.

JoachimHenze (approved):

binarymaster (approved):
Looks good, and it looks like there's stuff to be localized later :slightly_smiling_face:

PR #4748, opened 2022-10-03
[MSGINA] Automatize creation of branding banner at the top of the system dialogs.

HBelusca:
## Purpose

We don't need to manually place the banner on each dialog of choice
in all languages and size it accordingly. It's now done automatically.

This also fixes the problem of placing them for dialogs whose size
differ due to different dialog font size being used (e.g. CJK, or
because the user changed the default system font).

(See also PR #3661.)

## Changes

Code adaptation, and:
Modify dialogs: Remove logo/bar (now dynamically added).
_+ Correct/harmonize styles, etc.


JoachimHenze:
Can you attach a picture of how the result will look?

dll/win32/msgina/gui.c
```diff
@@ -57,57 +64,219 @@ DlgData_Create(HWND hwndDlg, PGINA_CONTEXT pgContext)
...
+
+    /* If the logo failed to be created, don't bother creating the bar and moving anything */
+    if (!hwndLogo)
         return;
+
+    /* Create and place the bar control just beneath the logo */
+    hwndBar = CreateWindowExW(0,
+                              WC_STATICW,
+                              ID2STR(IDI_BAR),
+                              WS_VISIBLE | WS_CHILD | SS_BITMAP /*| SS_OWNERDRAW*/,
```
JoachimHenze, line 179: Why keeping the commented SS_OWNERDRAW?

dll/win32/msgina/gui.c
```diff
@@ -57,57 +64,219 @@ DlgData_Create(HWND hwndDlg, PGINA_CONTEXT pgContext)
...
+#endif
+
+    /* Count the number of children windows in this dialog */
+    for (hWnd = GetWindow(hwndDlg, GW_CHILD);
+         hWnd;
+         hWnd = GetNextWindow(hWnd, GW_HWNDNEXT))
+    {
+        /* Skip the logo and bar controls */
+        // LONG_PTR nID = GetWindowLongPtrW(hWnd, GWLP_ID);
+        // if (nID == IDC_ROSLOGO || nID == IDC_BAR)
```
JoachimHenze, line 216: Why keeping those 2 commented lines?

dll/win32/msgina/gui.c
```diff
@@ -57,57 +64,219 @@ DlgData_Create(HWND hwndDlg, PGINA_CONTEXT pgContext)
...
+    /* Move each child window down */
+    hdwp = BeginDeferWindowPos(nNumChildren);
+
+    for (hWnd = GetWindow(hwndDlg, GW_CHILD);
+         hWnd;
+         hWnd = GetNextWindow(hWnd, GW_HWNDNEXT))
+    {
+        /* Skip the logo and bar controls */
+        // LONG_PTR nID = GetWindowLongPtrW(hWnd, GWLP_ID);
+        // if (nID == IDC_ROSLOGO || nID == IDC_BAR)
```
JoachimHenze, line 232: Same here, why keeping them?

dll/win32/msgina/gui.c
HBelusca: Notice how the PR is in *draft* state. This means that there are a lot of unfinished/untested/in-progress stuff ...

dll/win32/msgina/gui.c
binarymaster: Calm down, it's a draft PR yet :smile:

dll/win32/msgina/gui.c
```diff
@@ -498,9 +606,11 @@ WelcomeDialogProc(
                 return FALSE;
 
             DlgData_LoadBitmaps(pDlgData);
+            DlgData_MoveControls(hwndDlg, pDlgData);
             return TRUE;
         }
 
+#if 0
```
JoachimHenze, line 613: Why #if 0 and not deleting the code enclosed by that? Same question for the other occurences of the same thing.

dll/win32/msgina/msgina.rc
```diff
@@ -60,7 +60,8 @@ IDB_IMAGE_STRIP BITMAP "resources/image_strip.bmp"
 #ifdef LANGUAGE_JA_JP
     #include "lang/ja-JP.rc"
 #endif
-#ifdef LANGUAGE_NB_NO
+//#ifdef LANGUAGE_NB_NO
+#ifdef LANGUAGE_NN_NO
```
JoachimHenze, line 64: delete the commented line, and that diff will vanish entirely!

JoachimHenze (changes requested):
the msgina.rc change doesn't make any sense. Rest are mostly questions.

dll/win32/msgina/msgina.rc
HBelusca: Can you just stop reviewing this draft PR? For your msgina.rc "doesn't make any change" statement, you could just have looked at the first "temp" commit!

HBelusca:
> Can you attach a picture of how the result will look?

Like this (currently this is a MSPaint mockup):
Left: like normal in default settings ; Right: case of different font -> larger dialog: logo should be centered and filling color should be put to its left and right, and the bar is stretched.
image

See for comparison, how such a scenario looks like on Windows (here, I've artificially increased the size of the login dialog in the msgina resources, changed the logo to add a thin border around it, and put some markers on the bar):
image

JoachimHenze:
I do definitely favor the approach of PR4748 over the approach in PR3661. So if thisone is superseding the other one, you'll get my ok. Our foremost concern should be to get the hardcoded year 2022 out of those bitmaps to prevent recurrent maintenance.
Duplicating the bitmaps before that is reached is definitely a no-go. And the less dupes will be needed in the end, the better.

dll/win32/msgina/branding.c
```diff
@@ -0,0 +1,61 @@
+/*
+ * PROJECT:     ReactOS Logon GINA DLL
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Branding support for dialog boxes.
+ * COPYRIGHT:   Copyright 2006 Hervé Poussineau <hpoussin@reactos.org>
+ *              Copyright 2019 Stanislav Motylkov <x86corez@gmail.com>
```
binarymaster, line 6: Heh I already forgot that I've contributed here :sweat_smile:

dll/win32/msgina/branding.c
```diff
@@ -0,0 +1,61 @@
+/*
+ * PROJECT:     ReactOS Logon GINA DLL
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Branding support for dialog boxes.
```
binarymaster, line 4: Drop the full stop at the end :)

dll/win32/msgina/branding.c
```diff
@@ -58,4 +63,261 @@ Brand_Cleanup(
...
+    _In_ UINT uMsg,
+    _In_ LPARAM lParam)
+{
+    if (!pDlgData->hBarBitmap)
+        return;
+
+    switch (uMsg)
+    {
+#if 0
+        case WM_INITDIALOG:
```
binarymaster, line 241: Oh no, the bar animation thing is still being defined out, even with nested `#if 0`'s

binarymaster (commented):
Looks very promising, looking forward to it. :+1: Also this probably would be needed to do in shell32 (About window) and syssetup (initial driver installation window).

dll/win32/msgina/branding.c
HBelusca: Yes, the working code is still in gui.c . I was here thinking whether I could move it out too.

HBelusca:
> Looks very promising, looking forward to it. +1 Also this probably would be needed to do in shell32 (About window) and syssetup (initial driver installation window).

My intent with moving the stuff into branding.c is that we could also share it with shell and syssetup, in order to copy-paste the code.

HBelusca:
Another interesting remark is that one can tell msgina to add some extra custom strings (by editing msgina resources; one can also change the used font/size, but not the colors):
image
(Only when using a remote Terminal Services session does an extra copyright string appear; not shown on this picture.)

dll/win32/msgina/prototype/prototype.c
```diff
@@ -0,0 +1,196 @@
+// prototype.cpp : définit le point d'entrée pour l'application.
```
learn-more, line 1: Are you going to remove the prototype again?

HBelusca: Prototype is a temporary thing for me to quickly test/code the functionality, and once it's done, I plug it in msgina/shell32/syssetup , and remove the prototype, of course. (And clean up the commit history ^^)

dll/win32/msgina/branding.c
```diff
@@ -58,4 +63,261 @@ Brand_Cleanup(
...
+    _In_ PBRAND pBrand,
+    _In_ INT xDelta,
+    _In_ INT yDelta)
+{
+    INT xOld, yOld, cxOld, cyOld;
+    INT xNew, yNew, cxNew, cyNew;
+    RECT rc;
+    // DWORD style, exstyle;
+
+    /* Get "current" (-> old) window position and size */
```
binarymaster: Hmm I wonder what would be a non-hack way to move all controls.

HBelusca: "HACK" was refering to the previous purpose of the function (that was introduced by Katahiro to fixup the design of the dialog in japanese language only. Then I repurposed the function but it still needed cleanup, and I kept my own comment that it was a hack. I think I've now sanitized it, if you look at the code as a whole now (not at some intermediate commit).

sdk/include/reactos/ui/branding.h
```diff
@@ -0,0 +1,594 @@
...
+ * PROJECT:     ReactOS headers
+ * LICENSE:     LGPL-2.0-or-later (https://spdx.org/licenses/LGPL-2.0-or-later)
+ * PURPOSE:     Branding support for dialog boxes.
+ * COPYRIGHT:   Copyright 2006 Hervé Poussineau <hpoussin@reactos.org>
+ *              Copyright 2019 Stanislav Motylkov <x86corez@gmail.com>
+ *              Copyright 2020 Katayama Hirofumi MZ <katayama.hirofumi.mz@gmail.com>
+ *              Copyright 2022 Hermès Bélusca-Maïto <hermes.belusca-maito@reactos.org>
+ */
+
+#pragma once
```
binarymaster, line 11: I see you started to move branding into SDK. Is it possible to move actual resources (bitmaps) also there to avoid duplicating them in source tree?

HBelusca: This could be interesting to do. But then it would be useful to move them into the "media" folder... as long as they are indeed duplicated (aka. the very same files).

learn-more:
Closing this to reduce the backlog of stale PR's.
Feel free to reopen it when you are working on it again.

PR #4750, opened 2022-10-04
[USETUP][VGAFONTS] Improve Romanian character display

katahiromz:
## Purpose

JIRA issue: [CORE-18446](https://jira.reactos.org/browse/CORE-18446)

## Proposed changes

- Modify `media/vgafonts/28606-8x8.bin` Romanian VGA font to display special characters.
- Add global variables for special characters `Char...` for Romanian VGA font.
- Toggle the global variables `Char...` in `SetConsoleCodePage` function.
- Improve `LanguagePage` function.

## TODO

- [ ] Connect the lines to be connected.


katahiromz:
BEFORE:
before1

AFTER:
after1

katahiromz:
BEFORE:
before2

AFTER:
after2

katahiromz:
BEFORE:
before3

AFTER:
after3

katahiromz:
VGA font.

BEFORE:
before

AFTER:
after

binarymaster:
> I don't know why the adjacent lines are not connected.

This is because the video mode is 720x400. In this mode the VGA BIOS would glue together only linedrawing characters.

Current approach uses different key codes for them, that's why they are not being glued.

katahiromz:
>because the video mode is 720x400

Is 640 x 400 available for us?

binarymaster:
Yes, 640x400 mode should work without the magic glue, however I don't know if it's possible to switch to it in VirtualBox.

katahiromz:
Maybe `IOCTL_CONSOLE_SET_MODE`...

HBelusca:
> Yes, 640x400 mode should work without the magic glue, however I don't know if it's possible to switch to it in VirtualBox.

Is that available via VGA IO port operations?

binarymaster:
I don't actually remember whether it's really possible to choose between these two. From my own observations:
1. Virtual PC uses 640x400 for text mode
2. VirtualBox does use 720x400
3. QEMU (some old versions) use 640x400
4. QEMU (latest) uses 720x400
5. VMware uses 720x400 too if I remember correctly

base/setup/usetup/mui.c
```diff
@@ -535,6 +537,62 @@ SetConsoleCodePage(VOID)
...
+
+        case 932: /* Japanese */
+            /* FIXME */
+            break;
+    }
+    return Ch;
+}
+
+VOID
+MUIUpdateVideoMode(VOID)
```
HBelusca, line 583: This kind of functionality has to be moved in the console.c file, as part of the SetConsoleOutputCP() implementation.

base/setup/usetup/mui.c
```diff
@@ -30,6 +30,8 @@
 #define NDEBUG
 #include <debug.h>
 
+WORD SelectedCodePage = 437;
```
HBelusca, line 33: The idea of this PR looks nice, but wouldn't it be possible instead to have those character translation (incl. MUITranslateChar) in the console.c code, where the console functions could adjust the characters being emitted to the video (this would be helpful for when also we redirect to e.g. a serial port).

base/setup/usetup/usetup.c
```diff
@@ -663,6 +663,15 @@ LanguagePage(PINPUT_RECORD Ir)
                 /* Load the font */
                 SetConsoleCodePage();
 
+                /* Update video mode */
+                MUIUpdateVideoMode();
```
HBelusca, line 667: This should be part of the `SetConsoleCodePage` call, see comment above.

katahiromz: OK. Fixed.

base/setup/usetup/mui.c
katahiromz: Added global variables for special characters.

HBelusca:
Btw. the japanese box characters may be improved in a similar way.

PR #4752, opened 2022-10-04
[EVENTVWR] Fix settings loading. Remove usage of Rtl functions. Use Win32 CRT assert.

HBelusca:
## Purpose

Exactly what's written in the title.

## Proposed changes

- Don't create the settings registry key when trying to load them.
  Defer its creation when saving the settings.

- Use more restricted access rights when opening/creating the key.

- Ensure read and written string settings are NULL-terminated.

- Use the Win32 CRT assert so that if an assertion fails we get a
  message on the screen, not via the kernel debugger.

### DIFF
--- a/base/applications/mscutils/eventvwr/eventvwr.c
+++ b/base/applications/mscutils/eventvwr/eventvwr.c
@@ -346,116 +346,135 @@ Quit:
 BOOL
 LoadSettings(int nDefCmdShow)
 {
-    HKEY hKeyEventVwr;
     LONG Result;
-    DWORD dwSize;
-    DWORD dwType;
-    DWORD Value;
-    UNICODE_STRING ValueU;
+    HKEY hKeyEventVwr;
+    DWORD dwType, cbData;
     WCHAR buffer[100];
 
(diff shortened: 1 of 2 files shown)


base/applications/mscutils/eventvwr/eventvwr.c
```diff
@@ -346,116 +346,135 @@ ProcessCmdLine(IN LPWSTR lpCmdLine)
...
+    if (Result == ERROR_SUCCESS)
+    {
+        if (dwType == REG_SZ)
+        {
+            buffer[cbData / sizeof(WCHAR) - 1] = UNICODE_NULL;
+            Settings.bSaveSettings = !!(DWORD)_wtoi(buffer);
+        }
+        else if (dwType == REG_DWORD && cbData == sizeof(DWORD))
+        {
+            Settings.bSaveSettings = !!*(PDWORD)buffer;
```
binarymaster, line 392: Funny way to convert integer or pointer to a boolean :smile:

binarymaster (approved):
Looks good and clear.

PR #4762, opened 2022-10-06
[GIT] Add .DS_Store files to .gitignore

gonzoMD:
## Purpose

When working on macOS, the system creates a .DS_Store file in every accessed directory. They are marked as new files
in git, which is very annoying. This change will exclude these files.

## Proposed changes

- Add .DS_Store files to .gitignore


binarymaster (approved):

HBelusca:
ah yes those files that creep everywhere.

PR #4764, opened 2022-10-07
[RTL] RtlReAllocateHeap(): Promote a DPRINT1() to ASSERTMSG()

SergeGautherie:
for "Realloc in place failed, but it was the only option"

to help notice/investigate this case.

JIRA issues: https://jira.reactos.org/browse/CORE-15840?jql=resolution%20is%20EMPTY%20AND%20text%20~%20%22Realloc%20in%20place%20failed%22
6 open out of 11 total. (It could be unnoticed on some other tickets...)


ThFabba:
This could make sense for debugging those issues, but only for debugging. Generally, we don't want to assert on memory allocation failures.

PR #4775, opened 2022-10-13
Enable framebuf_new driver

hpoussin:
## Purpose

Replace framebuf display driver by framebuf_new display driver

## Proposed changes

- Compile framebuf_new as framebuf.dll, and add it to bootcd/livecd
- Remove old framebuf from compilation, to not conflict
## TODO

Later, remove framebuf driver if no regression is found.


HBelusca:
Ah, since we're here :) Could you tell us quickly what's the difference between framebuffer and framebuffer_new ? (as well as vga and vga_new)

binarymaster:
Looks like framebuffer_new is genuine(tm) and comes from NT4 DDK samples.

binarymaster (approved):

SergeGautherie:
> NT4 DDK samples

vga_new too.
https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=NT4+DDK

Should they both/all be added to
https://reactos.org/wiki/Problematic_Licenses
as atapi is
?
Or should they even be deleted rather than used?

PR #4778, opened 2022-10-14
[RUNDLL32] Add comctl32 manifest to rundll32 (CORE-16669)

katahiromz:
## Purpose

Support `Ctrl+A` (Select All) on `appwiz` (`Create Shortcut`) wizard.

JIRA issue: [CORE-16669](https://jira.reactos.org/browse/CORE-16669)

## Proposed changes

- Add exe manifest to `rundll32.exe`.

## Comparison

BEFORE:
before
`Ctrl+A` won't work.

after
`Ctrl+A` works well.

### DIFF
--- a/base/system/rundll32/rundll32.rc
+++ b/base/system/rundll32/rundll32.rc
@@ -1,4 +1,5 @@
 #include <windef.h>
+#include <winuser.h>
 
 #include "resource.h"
 
@@ -6,6 +7,8 @@
 #define REACTOS_STR_INTERNAL_NAME     "rundll32"
 #define REACTOS_STR_ORIGINAL_FILENAME "rundll32.exe"
 #include <reactos/version.rc>
+
+#include <reactos/manifest_exe.rc>
 
 IDI_MAIN ICON "res/main.ico"
 


HBelusca:
The title of the commit (and of the PR) is misleading, because it doesn't explain the connection between what's claimed ("enable Ctrl-A") and what's actually being done: adding the comctl32 manifest to rundll32.
Please explain why this is needed. Also please add to your commit message the JIRA report number.

katahiromz:
> Should the system edit boxes (available without comctl32 manifest) also support Ctrl-A

The non-comctl32 edit box won't process `Ctrl+A` key combination.
The comctl32 edit box does process.

`Ctrl+C` and `Ctrl+V` are supported in both non-comctl32 and comctl32.

katahiromz:
> Please explain why this is needed.

The `Edit` controls in `user32` won't handle `Ctrl+A` key combination.
The `Edit` controls in `comctl32` handles `Ctrl+A`.

`rundll32` loads a control panel applet and execute the procedure of the applet in the context of the `rundll32` module.
Thus, in order to enable `Ctrl+A` in `appwiz.cpl`, we have to add exe manifest to the `rundll32` module.

HBelusca:
Technically now, since the comctl32 manifest is added to rundll32, this should allow any control panel applet, too, as well as (in particular) this create-shortcut wizard, to be correctly themed when themes are being used in ReactOS :thinking:

JoachimHenze:
I do want to mention at least that adding manifests to even more of our dlls might unhide / intensify problems like https://jira.reactos.org/browse/CORE-13940
As that was also unhidden, when we started to add manifests.
We should really try to solve that issue, or the curse will get worse. @learn-more

learn-more (changes requested):
Windows 2k3's rundll32 does not have this manifest at all,
and Windows 10's rundll32 has a manifest, but only for comctl v5!

JoachimHenze (changes requested):
rejection based in Marks findings.

katahiromz:
Closed due to wrong implementation.

PR #4779, opened 2022-10-15
[USER32][NTUSER] Fill by white in DrawFrameControl:DFC_MENU

katahiromz:
## Purpose

`DrawFrameControl:DFC_MENU` draws the monochrome image of menu arrow or checkmark. However, the function didn't draw the entire rectangle correctly when the rectangle was not a square.
`DrawFrameControl:DFC_MENU` should fill the background rectangle by white.

JIRA issue: [CORE-18417](https://jira.reactos.org/browse/CORE-18417)

## Comparison

BEFORE:
before

AFTER:
after

## TODO

- [x] Do tests.

### DIFF
--- a/win32ss/user/user32/windows/draw.c
+++ b/win32ss/user/user32/windows/draw.c
@@ -978,6 +978,7 @@ static BOOL UITOOLS95_DrawFrameMenu(HDC dc, LPRECT r, UINT uFlags)
     TCHAR Symbol;
     RECT myr;
     INT cxy;
+    FillRect(dc, r, (HBRUSH)GetStockObject(WHITE_BRUSH));
     cxy = UITOOLS_MakeSquareRect(r, &myr);
     switch(uFlags & 0x1f)
     {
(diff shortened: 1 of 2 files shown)


win32ss/user/user32/windows/draw.c
```diff
@@ -978,6 +978,7 @@ static BOOL UITOOLS95_DrawFrameMenu(HDC dc, LPRECT r, UINT uFlags)
     TCHAR Symbol;
     RECT myr;
     INT cxy;
+    FillRect(dc, r, (HBRUSH)GetStockObject(WHITE_BRUSH));
```
learn-more, line 981: Are you sure `WHITE_BRUSH` is the correct brush?

learn-more (commented):
There is also an `UITOOLS95_DrawFrameMenu` in `win32ss/user/ntuser/draw.c`, this probably also needs fixing?

win32ss/user/user32/windows/draw.c
katahiromz: Yes, must be white. I confirmed.

learn-more (approved):

learn-more:
An optimization would be to not draw this unless it is required, or disable this drawing operation while drawing the arrow (set background mode to transparent)

PR #4780, opened 2022-10-15
[SHELL32] Fix using the wrong context menu

learn-more:
This got broken when when resolving a conflict between two changes in the same function.


binarymaster (approved):
Fixes the problem for me :+1:

PR #4790, opened 2022-10-20
[NTDLL] Implement LdrInitShimEngineDynamic.

learn-more:
This is the final step required to properly support SE_DynamicShim 

JIRA issue: [CORE-13283](https://jira.reactos.org/browse/CORE-13283)

### DIFF
--- a/dll/ntdll/ldr/ldrapi.c
+++ b/dll/ntdll/ldr/ldrapi.c
@@ -1672,4 +1672,23 @@ LdrSetAppCompatDllRedirectionCallback(
     return STATUS_NOT_IMPLEMENTED;
 }
 
+BOOLEAN
+NTAPI
+LdrInitShimEngineDynamic(IN PVOID BaseAddress)
+{
+    ULONG_PTR Cookie;
+    NTSTATUS Status = LdrLockLoaderLock(0, NULL, &Cookie);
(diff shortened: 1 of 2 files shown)


dll/ntdll/ldr/ldrapi.c
```diff
@@ -1672,4 +1672,23 @@ LdrSetAppCompatDllRedirectionCallback(
...
+    {
+        if (!g_pShimEngineModule)
+        {
+            g_pShimEngineModule = BaseAddress;
+            LdrpGetShimEngineInterface();
+        }
+        LdrUnlockLoaderLock(0, Cookie);
+        return TRUE;
+    }
+    return FALSE;
```
binarymaster, line 1691: If you return FALSE early here, you can reduce indentation level :smile:

learn-more: But why?

binarymaster: Well if there's nothing to be done additionaly in a failure case, this would simplify the code. Simpler is usually better, not always though, but in this case why not.

learn-more: Introducing multiple exit points does not make the code simpler imo.

HBelusca (approved):

PR #4792, opened 2022-10-20
[APPHELP] Stop shimeng being initialized multiple times

learn-more:
[CORE-18532](https://jira.reactos.org/browse/CORE-18532)
[CORE-13283](https://jira.reactos.org/browse/CORE-13283)


HBelusca:
OK just to understand: is this a "real" fix, or a hack?

learn-more:
> OK just to understand: is this a "real" fix, or a hack?

This is a hack, to fix problem caused by the hack that is used to shim msi files.

HBelusca:
OK then maybe it needs to be told so in the commit message (besides the fact that you've added the DPRINT in the code).

HBelusca (approved):

JoachimHenze (approved):

PR #4793, opened 2022-10-20
[BROWSEUI] Fix CAutoComplete use after free

learn-more:
[CORE-18395](https://jira.reactos.org/browse/CORE-18395)

This was very visible when running with DPH:
Open an explorer window,
and close it again.

### DIFF
--- a/dll/win32/browseui/CAutoComplete.cpp
+++ b/dll/win32/browseui/CAutoComplete.cpp
@@ -1,7 +1,7 @@
 /*
  *    AutoComplete interfaces implementation.
  *
- *    Copyright 2004    Maxime Bellengé <maxime.bellenge@laposte.net>
+ *    Copyright 2004    Maxime BellengÃ© <maxime.bellenge@laposte.net>
  *    Copyright 2009  Andrew Hill
  *    Copyright 2020-2021 Katayama Hirofumi MZ <katayama.hirofumi.mz@gmail.com>
  *
@@ -1549,7 +1549,7 @@ LRESULT CAutoComplete::OnCreate(UINT uMsg, WPARAM wParam, LPARAM lParam, BOOL &b
     m_hFont = reinterpret_cast<HFONT>(::GetStockObject(DEFAULT_GUI_FONT));
     m_hwndList.SetFont(m_hFont);
 
-    // add reference to CAutoComplete::m_hWnd
+    // add reference so we won't be deleted during message processing
     AddRef();
     return 0; // success
 }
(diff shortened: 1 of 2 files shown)


HBelusca (approved):

dll/win32/browseui/CAutoComplete.h
```diff
@@ -284,6 +284,8 @@ class CAutoComplete
     LRESULT OnAutoCompStart(UINT uMsg, WPARAM wParam, LPARAM lParam, BOOL &bHandled);
     LRESULT OnAutoCompFinish(UINT uMsg, WPARAM wParam, LPARAM lParam, BOOL &bHandled);
 
+    virtual VOID OnFinalMessage(HWND);
```
katahiromz, line 287:
```cxx
virtual VOID OnFinalMessage(HWND) override;
```

katahiromz (approved):

dll/win32/browseui/CAutoComplete.cpp
```diff
@@ -1,7 +1,7 @@
 /*
  *    AutoComplete interfaces implementation.
  *
- *    Copyright 2004    Maxime Belleng� <maxime.bellenge@laposte.net>
+ *    Copyright 2004    Maxime Bellengé <maxime.bellenge@laposte.net>
```
JoachimHenze, line 4: If this is a real change, and no fuckup of the gut review webinterface, then please do revert it!
Same in the 2nd file that was touched.

JoachimHenze (changes requested):

JoachimHenze:
Which commit did introduce that Use-After-Free?

dll/win32/browseui/CAutoComplete.cpp
HBelusca: Perhaps it's an ANSI -> UTF8 conversion made in the file, in which case I would say it's fine.

JoachimHenze: Why does it look worse than before then, if you switch into githubs "Files changed"-tab?
If it is really just wrong in the webview, then please ignore my rejection.

learn-more: it used to be ansi indeed, now it's utf8

learn-more:
> Which commit did introduce that Use-After-Free?

probably the commit introducing CAutoComplete

learn-more:
https://reactos.org/testman/compare.php?ids=84462,84466

ThFabba (approved):

PR #4796, opened 2022-10-22
[NTFS] Add a few casts to ULONGLONG from ULONG/LONGLONG

SergeGautherie:
## Purpose

This is probably not a fix for this very ticket, yet a small step.

JIRA issue: [CORE-12101](https://jira.reactos.org/browse/CORE-12101)

## Proposed changes

in
AttributeAllocatedLength() and AttributeDataLength(),
NtfsReadFile() and NtfsWriteFile().

- Be explicit about 32/64 bits and signs..

### DIFF
--- a/drivers/filesystems/ntfs/mft.c
+++ b/drivers/filesystems/ntfs/mft.c
@@ -249,9 +249,9 @@ ULONGLONG
 AttributeAllocatedLength(PNTFS_ATTR_RECORD AttrRecord)
 {
     if (AttrRecord->IsNonResident)
-        return AttrRecord->NonResident.AllocatedSize;
+        return (ULONGLONG)AttrRecord->NonResident.AllocatedSize; // From LONGLONG.
     else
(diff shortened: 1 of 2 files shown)


JoachimHenze:
Is there any compiler complaining about the current code state, or what is the motivation for those explicit casts?

drivers/filesystems/ntfs/mft.c
```diff
@@ -249,19 +249,19 @@ ULONGLONG
 AttributeAllocatedLength(PNTFS_ATTR_RECORD AttrRecord)
 {
     if (AttrRecord->IsNonResident)
-        return AttrRecord->NonResident.AllocatedSize;
+        return (ULONGLONG)AttrRecord->NonResident.AllocatedSize; // From LONGLONG.
```
learn-more, line 252: Are you editing in notepad, or why else do you feel the need to annotate what type it is being cast from?

PR #4800, opened 2022-10-22
[SHELL32] Delete old unused implementation of OLD_SHAppBarMessage()

JoachimHenze:
SHAppBarMessage() was reimplemented at a different place by 0.4.14-dev-1209-g 10d1afea18db0ef06942239867cbc090e2330b90
A reimport from Wine.

Cleanup-Addendum to CORE-14439


dll/win32/shell32/wine/shell32_main.c
```diff
@@ -965,72 +965,6 @@ typedef struct
     HICON hIcon;
 } ABOUT_INFO;
 
-#define DROP_FIELD_TOP    (-15)
```
binarymaster, line 968: Note the code in this file is guarded with `#ifdef __REACTOS__` in some places. If this is Wine code, you cannot simply delete it, but undefine it instead.

JoachimHenze: see https://github.com/wine-mirror/wine/blob/master/dlls/shell32/shell32_main.c for your reference.
That define is solely used in Wines
static void paint_dropline( HDC hdc, HWND hWnd )
which we do not even have

binarymaster: Ok so the code was added by @gedmurphy here: https://github.com/reactos/reactos/commit/6ced137ce2f6f415c2106ce8b57d936002cf5941#diff-59157aaaa82be46a8d33fad61bdd91186ce6f10d3c0c3bbd332ac883ed22c504

JoachimHenze: yes. But like I wrote it my description already: That code is not gone, but was moved by jimtabors
0.4.14-dev-1209-g https://github.com/reactos/reactos/commit/10d1afea18db0ef06942239867cbc090e2330b90
analog to where Wine stores it today:
https://github.com/wine-mirror/wine/blob/master/dlls/shell32/appbar.c

He just left the old duplicate existing and renamed it. Effectively reducing the sync to Wine.
We should not keep old code by copy-pasting it. If one needs a reference to the former state, that is what a version control system is used for.

HBelusca (approved):

binarymaster (approved):

PR #4801, opened 2022-10-23
[NTOS:MM] MmPurgeSegment: Fix wrong return value

archeYR:
Return TRUE instead of NTSTATUS code which has a value of FALSE and may confuse caller. Fixes sporadic 0x7B bugcheck when booting from corrupted NTFS volume using WinXP ntfs.sys.

### DIFF
--- a/ntoskrnl/mm/section.c
+++ b/ntoskrnl/mm/section.c
@@ -4739,7 +4739,7 @@ MmPurgeSegment(
     if (!Segment)
     {
         /* Nothing to purge */
-        return STATUS_SUCCESS;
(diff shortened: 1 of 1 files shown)


HBelusca (approved):

ThFabba (approved):
LGTM -- though TBH, functions in the kernel shouldn't return BOOLEAN in the first place, unless maybe they have "Is" or "Does" in the name.

binarymaster (approved):
LGTM, also testbots are happy: https://reactos.org/testman/compare.php?ids=84462,84463

PR #4802, opened 2022-10-23
CSR maintenance: Moving it to its own subdirectory.

HBelusca:
## Purpose

The rationale is that the CSR subsystem can theoretically be used for other subsystems than Windows (Win32) -- up until Windows 7 included, that is
(see https://www.geoffchappell.com/studies/windows/win32/csrsrv/api/apireqst/api_connectinfo.htm , the presence of `ObjectDirectory` and `NumberOfServerDllNames`, that allows this freedom),
and so, it shouldn't be too strongly related by name with "Win".
Needless to say that NT 6.2 aka Windows 8 dumbed CSR down by removing those key values.
(Not sure we want to follow such a dumb path as well, but anyway...)

I also move the client side of CSR together with it, into a library "CSRLIB" that is then compiled into NTDLL.

In ReactOS, this could allow us to do something similar to what was previously attempted in antediluvian times:
https://reactos.org/archives/public/ros-diffs/2005-August/007776.html
for example:
```
HKLM,"SYSTEM\CurrentControlSet\Control\Session Manager\Subsystems","Posix",0x00020000,
"%SystemRoot%\system32\csrss.exe ObjectDirectory=\POSIX SharedSection=128 Windows=Off SubSystemType=Text ServerDll=psxsrv,1"
```
by keeping delegating to CSRSRV all the basic management of client/server message passing that is common to all API subsystems one could implement. (And thus, removing the need to duplicate boilerplate code. An alternative would be to move said boilerplate code to some shared library, compiled both in CSRSRV and the other subsystems...)

NOTA: The two other commits are kind of stand-alone and could be committed separately out of this PR.

## Proposed changes

See each separate commit.


JoachimHenze:
> I also move the client side of CSR together with it, into a library "CSRDLL" that is then compiled into NTDLL.

Wouldn't CSRLIB be a better name then, if it's always/only statically compiled into other modules?

sdk/include/reactos/subsys/csr/csr.h
```diff
@@ -1,10 +1,9 @@
 /*
- * COPYRIGHT:       See COPYING in the top level directory
- * PROJECT:         ReactOS Client/Server Runtime SubSystem
- * FILE:            include/reactos/subsys/csr/csr.h
- * PURPOSE:         Public definitions for CSR Clients
- * PROGRAMMERS:     Alex Ionescu (alex.ionescu@reactos.org)
- *                  Hermes Belusca-Maito (hermes.belusca@sfr.fr)
+ * PROJECT:     ReactOS Client/Server Runtime SubSystem
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Public definitions for CSR Clients.
```
binarymaster, line 4: Please drop the ending full stop in `PURPOSE` fields, unless there are more than one sentence describing it (here and in other places).

binarymaster (approved):
This looks ok, however the idea of `CSRLIB` is more appealing to me.

binarymaster (approved):

subsystems/csr/csrlib/csrdll.spec
```diff
@@ -0,0 +1,17 @@
+@ stdcall CsrAllocateCaptureBuffer(long long)
```
learn-more, line 1: What is this file used for?
I don't see it referenced anywhere

learn-more (approved):

subsystems/csr/csrlib/csrdll.spec
HBelusca: I added it for reference and later usage.

GeoB99 (approved):
I love the idea 👍

PR #4805, opened 2022-10-24
[NTUSER] Strengthen Input Context tracing

katahiromz:
## Purpose
Improve debuggability on kernel input context.

## Proposed changes

- Add more `ERR`, `WARN`, and `TRACE`.


HBelusca:
In principle, "debuggability" should be done with more usage of an actual debugger like WinDbg, and much less by polluting debug logs...

katahiromz:
> In principle, "debuggability" should be done with more usage of an actual debugger
I believe this PR helps us find the logical error and reduce debugging time.

PR #4808, opened 2022-10-24
[SHELL32] Update German translation

gonzoMD:


dll/win32/shell32/lang/de-DE.rc
```diff
@@ -843,8 +843,8 @@ BEGIN
     IDS_FORMAT_COMPLETE "Formatieren abgeschlossen."
 
     /* Warning format system drive dialog strings */
-    IDS_NO_FORMAT_TITLE "Cannot format this volume"
-    IDS_NO_FORMAT "This volume cannot be formatted! It contains important system files in order for ReactOS to run."
+    IDS_NO_FORMAT_TITLE "Formatieren des Laufwerks nicht möglich"
+    IDS_NO_FORMAT "Dieses Laufwerk kann nicht formatiert werden! \nEs enthält wichtige Systemdateien, welche zur Ausführung von ReactOS benötigt werden."
```
JoachimHenze, line 847: There is an unintended space in the string before the newline.

dll/win32/shell32/lang/de-DE.rc
```diff
@@ -952,7 +952,7 @@ BEGIN
     IDS_EDIT_VERB "Bearbeiten"
     IDS_FIND_VERB "Finden"
     IDS_PRINT_VERB "Drucken"
-    IDS_CMD_VERB "Command Prompt hier"
+    IDS_CMD_VERB "Eingabeaufforderung hier"
```
JoachimHenze, line 955: I would really love if we wouldn't translate that further than it currently is. Reasons:
1.) C... is easily associated in the brain with cmd, while E... is not.
2.) The context menu might get wider than necessary with the long german word Eingabeaufforderung.
3.) The widespread Powertoys tool also calls this "Command prompt here" without localization.

If you really have to change that out of some intrinsic motivation, then please at least use
"CMD hier" or "Cmd hier".
But please not "Eingabeaufforderung hier". Thank you!

JoachimHenze (changes requested):
just the two things. Beside that LGTM.

dll/win32/shell32/lang/de-DE.rc
HBelusca: > Eingabeaufforderung 

Indeed, the noun is not long enough for a good translation :rofl:

binarymaster: Just to note, if we want to be consistent, these cases would need to be fixed: https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=Eingabeaufforderung

JoachimHenze: interesting idea. I would have expected more matches actually. That would indeed also solve uglyness like that:
image

dll/win32/shell32/lang/de-DE.rc
gonzoMD: fixed

dll/win32/shell32/lang/de-DE.rc
gonzoMD: I changed it to Cmd.
I second your opinion that "Eingabeaufforderung" is too long. ("lovely" German language).
But "Command Prompt hier" sounds Denglisch.
The opinion that most ppl are looking for "cmd" do I share only partially. For me and you it's cmd, but for average Joe it might not. (In fact I never do click somewhere when opening a prompt, I press win+R and type cmd)

So I think it's a good compromise.

JoachimHenze (approved):

binarymaster (approved):
Regarding the Cmd thing, I hope you'll create a follow up PR making it consistent :slightly_smiling_face:

PR #4824, opened 2022-10-29
[ADVAPI32] RegCreateKeyExW(): Log parameters

SergeGautherie:
## Purpose

Give clues about failures.

JIRA issue: [CORE-15471](https://jira.reactos.org/browse/CORE-15471)

## Proposed changes

- Enhance a `TRACE()`.
- Add a `WARN()`.


JoachimHenze:
Why not just executing with that diff locally? Why has this be committed to master also?

SergeGautherie:
> Why not just executing with that diff locally? Why has this be committed to master also?

To be available when needed, so testers are not required/blocked to do a custom build each time.

learn-more:
We don't need yet another implementation of something that already exists:
https://reactos.org/wiki/Debugging#Log_api_calls_to_a_specific_ReactOS_dll

PR #4825, opened 2022-10-30
[SHELL32] Canonicalize path in CShellLink

katahiromz:
## Purpose

Fix LibreOffice icon.
JIRA issue: [CORE-18080](https://jira.reactos.org/browse/CORE-18080)

## Proposed changes

- Use `GetFullPathName` to canonicalize the path in `CShellLink::SetTargetFromPIDLOrPath`.

## Comparison

BEFORE:
before
The icon was blank.

AFTER:
after
Successful.


HBelusca:
PR is invalid. Reread the bug report. I mentioned (and the WinDbg trace on Windows made it clear) that it was PathResolve that had to collapse the backslashes, but didn't do so in ReactOS (hence bug in ROS' PathResolve).

PR #4827, opened 2022-10-30
[SETUP][INF] Delete obsolete font substitutions "DejaVu Serif"

JoachimHenze:
## Purpose

Cleanup of leftovers.

These 2 lines are nonsense now for 2 different reasons even:

Since 0.4.13-dev-764-g dbb4432b25dc65b9d65f02bfc09d47d63b8cd5e5 we do have a proper "Times New Roman" font, so there is no need for substitution any longer.

And since 0.4.14-dev-20-g 2f4fb903b46207ab380365908401ded63752a108 we don't even have the substitution-target "DejaVu Serif" anymore.


JoachimHenze:
cc @peterooch also as potential reviewer. Sorry can not add you otherwise.

katahiromz (approved):
Nice catch!

peterooch (approved):
I checked to see if the Hebrew glyphs are there and they are indeed there.

binarymaster:
Gave PR a better title :wink:

HBelusca:
You should double-check the other instances of "DejaVu Serif" remaining in our code base:
https://git.reactos.org/?p=reactos.git&a=search&h=9c21d0c12438b9dc3462100dae37e88c0f522c60&st=grep&s=DejaVu+Serif

JoachimHenze:
> You should double-check the other instances of "DejaVu Serif" remaining in our code base: https://git.reactos.org/?p=reactos.git&a=search&h=9c21d0c12438b9dc3462100dae37e88c0f522c60&st=grep&s=DejaVu+Serif

I checked them @HBelusca .
I will not touch the "media/fonts/doc/Deja Vu" because the txt files in there are shared for the different fonts "DejaVu Sans" "DejaVu SansMono" and "DejaVu Serif".
We stripped "DejaVu Sans" and "DejaVu Serif", but we do still use "DejaVu SansMono". At least today.

But I thinned out the gdi32:GetTextMetrics test. See next commit.
And please do not come with greps for the other dropped font (Ubuntu). That should happen in a different PR definitely!

The rest of the matches for "Dejavu Serif" I covered already.

JoachimHenze:
KVM https://reactos.org/testman/compare.php?ids=84582,84591 LGTM 3 tests less skipped, as expected
VBox https://reactos.org/testman/compare.php?ids=84581,84590 LGTM 3 tests less skipped, as expected

Both bots do show significantly faster test execution times of ~2 minutes!

PR #4829, opened 2022-10-31
[SETUP][INF][ROSTESTS] Delete font substitutions and tests for "DejaVu Sans"

JoachimHenze:
Since 0.4.14-dev-6-g f45dd6577321355752e29ccd5f49039cfdaa761e we do have a proper "Verdana" font, so there is no need for substitution with "DejaVu Sans" any longer. That is the proper MS name for such a font.

And with 0.4.14-dev-20-g 2f4fb903b46207ab380365908401ded63752a108 we even deleted the "DejaVu Sans" font, so it makes no sense to keep it as a substitution target.

So adapt fonts.inf and muifonts.h, and also a test where it was still referenced. MS 2k3sp2 does neither have such a font substitute, nor such a font.


JoachimHenze:
cc @yagoulas as a greek tester and cc @peterooch as a Hebrew tester. All your magic characters still working? Or at least working as good as before?

JoachimHenze:
The grep looks clean now to my eyes: https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=DejaVu+Sans
because
* "DejaVu Sans **Mono**" is to date still a legit font that we do use,your eyes have to filter those out
* The textfiles within media/fonts/doc/DejaVu shall not be touched, as they are shared for "DejaVu Sans Mono" (which we still have) and "DejaVu Sans" (which we don't have any longer), and "DejaVu Serif" (which we don't have any longer)
* sdk/lib/3rdparty/freetype/src/truetype/ttsubpix.c is 3rd party and therefore I will not touch it

peterooch (approved):

JoachimHenze:
The greek text rendering changed by this PR:
Before:
greek_before

After:
greek_after

Rendering got more sharp, but the text weight seems to be too bold now. Since I don't know which is better, or how the font substitutions actually have to look like on a Greek system, I will push another commit to mitigate that.
The "old" state might be a workaround by abusing font substitutions, I cannot say for sure.
But I don't intend to change anything here visually. So PR is WIP for a moment.

HBelusca:
Is it falling back to some dejavu bold or sthg?

JoachimHenze:
> Is it falling back to some dejavu bold or sthg?
I cannot say for sure to what it falls back, but from my eyes it looks like Tahoma Bold.
**I do suspect that both renderings: before and after are wrong.**

JoachimHenze:
This is how it would look like if I deleted all substitutes to "Tahoma" in the greek section of
[base/setup/lib/muifonts.h](https://github.com/reactos/reactos/pull/4829/files#diff-41826f6ac79d18709c624f29c7c6729851a4808c05232826d945e74d9f25cf33)
and
[media/inf/font.inf](https://github.com/reactos/reactos/pull/4829/files#diff-37ba06c846a96b553eedd639aab50417fcb2f49151d17c221c934091fd154e00)
greek_this_Is_How_it_would_look_like_when_deleting_all_Tahoma_substitutes
This is kind of an inbetween-both-worlds-state.
Most dialogs are displaying then the blurry unbold font.
But some parts, e.g. within the startmenu do still use the bold Tahoma (non-blurry).
That is still not like it was before.
I will try next whether substituting all of them to Verdana will render as before. (Although that is most likely wrong regarding the substitutions correctness, and will also not give best visual result.). But that would be visually like before the PR I guess.

JoachimHenze:
Restoring the hack of using "Verdana" wherever I used "Tahoma" before looks like that:
Verdana_everywhere
I left a comment, that it is a hack in the code at keast, abusing the font substitutions as workaround for Tahoma drawing everything bold erroneously otherwise.

JoachimHenze:
Good to go now?

media/inf/font.inf
```diff
@@ -48,20 +48,23 @@ HKLM,"SOFTWARE\Microsoft\Windows NT\CurrentVersion\FontSubstitutes","Times",0x00
 HKLM,"SOFTWARE\Microsoft\Windows NT\CurrentVersion\FontSubstitutes","Tms Rmn",0x00000000,"Times New Roman"
 
 [Font.Greek.Reg]
+; FIXME: I do suspect that all substitutions to "Verdana" in this section and in the corresponding Greek section in
+; base/setup/lib/muifonts.h are a hack actually. I suspect that actually "Tahoma" would be correct for all of them.
+; Using "Tahoma" looks bold erroneously. Using "Verdana" looks blurry, but (at least) not bold.
+; see https://github.com/reactos/reactos/pull/4829 for some pictures
```
HBelusca, line 54:
```suggestion
; See https://github.com/reactos/reactos/pull/4829 for some pictures.
```

HBelusca:
> Good to go now?

Wait for katayama to review, at least...

katahiromz (approved):

HBelusca (approved):

JoachimHenze:
cc'ing @yagoulas was the very first action that I did when I created this PR. See my first comment.
Unfortunately he didn't react yet.
I am sure, that he is aware of the glitch. If I remember correctly, he brought up the topic that shell for Greek looks "bold unintentionally" (with Tahoma) 10 years ago already in IRC. I will merge the PR tomorrow.

media/inf/font.inf
```diff
@@ -48,20 +48,23 @@ HKLM,"SOFTWARE\Microsoft\Windows NT\CurrentVersion\FontSubstitutes","Times",0x00
 HKLM,"SOFTWARE\Microsoft\Windows NT\CurrentVersion\FontSubstitutes","Tms Rmn",0x00000000,"Times New Roman"
 
 [Font.Greek.Reg]
+; FIXME: I do suspect that all substitutions to "Verdana" in this section and in the corresponding Greek section in
+; base/setup/lib/muifonts.h are a hack actually. I suspect that actually "Tahoma" would be correct for all of them.
+; Using "Tahoma" looks bold erroneously. Using "Verdana" looks blurry, but (at least) not bold.
+; See https://github.com/reactos/reactos/pull/4829 for some pictures.
```
binarymaster, line 54:
```suggestion
; FIXME: It's suspected that all substitutions to "Verdana" in this section and in the corresponding Greek section in
; base/setup/lib/muifonts.h are a hack actually. Perhaps actually "Tahoma" would be correct for all of them.
; Using "Tahoma" looks bold erroneously. Using "Verdana" looks blurry, but (at least) not bold.
; See https://github.com/reactos/reactos/pull/4829 for some pictures.
```

binarymaster (approved):
Minor fix for the comment, otherwise LGTM.

katahiromz:
#4837 is merged. Please retry with new ReactOS.

JoachimHenze:
Greek draws perfectly fine now, when applying Tahoma now wherever I applied Verdana before, since (#4837)
greek_perfectly_draws_with_Tahoma_now

I will close this PR now and push the result by hand, because I am too stupid to make git do what I want within this PR, since I updated my local "DejavuSans" branch with the latest changes from master. I will reference this PR in the direct Push.

media/inf/font.inf
JoachimHenze: superseded by the direct push to master. The comment is no longer necessary at all.

JoachimHenze:
KVM https://reactos.org/testman/compare.php?ids=84697,84700 LGTM
VBox https://reactos.org/testman/compare.php?ids=84696,84699 LGTM

PR #4832, opened 2022-10-31
[PSDK] SECURITY_ATTRIBUTES: Remove size_is(nLength)

SergeGautherie:
nLength is the size of the struct itself,
not a number of descriptors.

Addendum to 1f12113 (r26428).
JIRA issue: noticed while investigating CORE 15471.


JoachimHenze:
What does that fix?

SergeGautherie:
> What does that fix?

Compatibility with MS SDK, wrong "self documentation", ...

HBelusca (approved):

HBelusca:
What it actually fixes is the MIDL code that gets generated out of this.

GeoB99 (approved):

PR #4841, opened 2022-11-02
[REGEDIT] Use shell icons for the tree view

binarymaster:
# Purpose 

Use shell icons for the tree view and do not hardcode them.

JIRA issue: [CORE-10764](https://jira.reactos.org/browse/CORE-10764)

## Proposed changes

- This makes regedit.exe smaller
- Reduces resource duplication in the source tree
- Potentially improves support for theming


HBelusca:
"Is it what Windows does?" (TM)
Also what happens in the scenario where you would use regedit.exe in a PE-like environment that doesn't incorporate shell32.dll?

binarymaster:
## Demo

| ReactOS | Windows XP | Windows 8.1 | Windows 10 |
| --- | --- | --- | --- |
| reactos | winxp | win8 | win10 |

binarymaster:
> "Is it what Windows does?" (TM)

Yes, but only since Windows Vista or so.

> Also what happens in the scenario where you would use regedit.exe in a PE-like environment that doesn't incorporate shell32.dll?

I don't think there's such environment exists. Also Regedit requires shell open/save dialogs.

HBelusca:
> Also Regedit requires shell open/save dialogs

We don't use the comdlg32 dialogs? (that themselves, show the shell32 ones when the shell is present.)

HBelusca:
> I don't think there's such environment exists

WinPE 1.x

binarymaster:
> We don't use the comdlg32 dialogs? (that themselves, show the shell32 ones when the shell is present.)

Hmm... didn't checked that.

> WinPE 1.x

Well I have checked, and our regedit has a strong shell32 dependency (`CommandLineToArgvW` and `ShellAboutW`):

image

Windows XP regedit does depend on shell32 as well:

image

base/applications/regedit/treeview.c
```diff
@@ -441,42 +446,32 @@ static BOOL InitTreeViewImageLists(HWND hwndTV)
...
     /* Add the open file, closed file, and document bitmaps.  */
-    hico = LoadImageW(hInst,
-                      MAKEINTRESOURCEW(IDI_OPEN_FILE),
-                      IMAGE_ICON,
-                      GetSystemMetrics(SM_CXSMICON),
-                      GetSystemMetrics(SM_CYSMICON),
-                      0);
+    wcscpy(szShell32, L"%SystemRoot%\\system32\\shell32.dll");
+    hico = NULL;
+    ExtractIconExW(szShell32, -IDI_SHELL_FOLDER_OPEN, NULL, &hico, 1);
```
binarymaster, line 458: Actually this could be simplified by using the same `LoadImageW`, as shell32 should be already loaded in the process.

katahiromz (approved):

milawynsrealm:
> Also what happens in the scenario where you would use regedit.exe in a PE-like environment that doesn't incorporate shell32.dll?

https://en.wikipedia.org/wiki/Server_Core

base/applications/regedit/treeview.c
```diff
@@ -441,41 +446,42 @@ static BOOL InitTreeViewImageLists(HWND hwndTV)
     HICON hico;       /* handle to icon  */
     INT cx = GetSystemMetrics(SM_CXSMICON);
     INT cy = GetSystemMetrics(SM_CYSMICON);
+    HMODULE hShell32 = GetModuleHandleW(L"SHELL32");
```
HBelusca, line 449: upper, or lowercase? with `.dll` extension ?

base/applications/regedit/treeview.c
```diff
@@ -441,41 +446,42 @@ static BOOL InitTreeViewImageLists(HWND hwndTV)
...
-    hico = LoadImageW(hInst,
-                      MAKEINTRESOURCEW(IDI_OPEN_FILE),
+    hico = LoadImageW(hShell32,
+                      MAKEINTRESOURCEW(IDI_SHELL_FOLDER_OPEN),
                       IMAGE_ICON,
-                      GetSystemMetrics(SM_CXSMICON),
-                      GetSystemMetrics(SM_CYSMICON),
+                      cx,
+                      cy,
                       0);
```
HBelusca, line 461: 0 --> `LR_DEFAULTCOLOR`

base/applications/regedit/treeview.c
binarymaster: Seems to work fine without extension, but ok I have added it and made it lowercase.

HBelusca (approved):

PR #4847, opened 2022-11-04
[NTOS:MM] NtCreatePagingFile(): Do not delete file on close

SergeGautherie:
Restore previous/default behavior.

Addendum to 3814a82 (0.4.10-dev-470).
JIRA issue: [CORE-14924](https://jira.reactos.org/browse/CORE-14924)

### DIFF
--- a/ntoskrnl/mm/pagefile.c
+++ b/ntoskrnl/mm/pagefile.c
@@ -544,14 +544,14 @@ NtCreatePagingFile(
                           FILE_ATTRIBUTE_SYSTEM | FILE_ATTRIBUTE_HIDDEN,
                           FILE_SHARE_WRITE,
                           FILE_SUPERSEDE,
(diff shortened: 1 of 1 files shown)


HBelusca:
PR is wrong: on Windows, the first `IoCreateFile` sets the `CreateOptions` parameter to value `0x9008`, which decomposes into `0x8000 | 0x1000 | 0x0008`, which are, respectively, `FILE_NO_COMPRESSION | FILE_DELETE_ON_CLOSE | FILE_NO_INTERMEDIATE_BUFFERING`, and if this call fails, the second `IoCreateFile` call uses the value `0x8008`, that decomposes into `0x8000 | 0x0008` i.e. `FILE_NO_COMPRESSION | FILE_NO_INTERMEDIATE_BUFFERING`.

PR #4852, opened 2022-11-06
[SETUP][FONTS][INF][GDI32_APITEST] Delete FreeMono and re-map

katahiromz:
## Purpose

Reduce memory usage and bundle size and standardize the fonts.

JIRA issue: [CORE-18605](https://jira.reactos.org/browse/CORE-18605)

## Proposed changes

- Delete `FreeMono` fonts.
- Re-map `"Courier"` font substitutes to `"Courier New"` fonts.

## TODO

- [x] Do small tests.


JoachimHenze (changes requested):
The substitutions in muifonts.h still need to be changed in a similar manner as in font.inf. Both should be kept synchron.

JoachimHenze:
Yes definitely better than before now. But...
Here is the font substitutions list from my german XPSP3ger
my_XPSP3ger_font_substitutions
(I do hope that it is still stock).
I don't see any Courier->anything in there. What do we need them for? Can they be stipped entirely maybe?
I am not sure about that. That is a question.

katahiromz:
In Windows, `"Courier"` is the font collection of the real raster fonts. You can see them in `Fonts` folder.

JoachimHenze:
Okay, then I suppose we will need those substitutions until we do support raster fonts, ok
https://git.reactos.org/?p=reactos.git;a=commit;h=200665fb89115b47c8c0b5f228413c95a6a90784 is the commit that introduced the "Courier" substitutions by @bragin in 2008. The referenced ticket ID cannot be watched anymore, because we didn't use JIRA yet back then.

I do give my consent now. As this is as good as it will get today.

JoachimHenze (approved):

HBelusca:
> The referenced ticket ID cannot be watched anymore, because we didn't use JIRA yet back then.

This is why you can search by "External ID" in JIRA...

HBelusca (approved):

PR #4853, opened 2022-11-06
[SHELL32][SDK] Check drop target for allowed effects

learn-more:
This fixes 'paste as link' being enabled in the recycler

JIRA issue: [CORE-16811](https://jira.reactos.org/browse/CORE-16811)

### DIFF
--- a/dll/win32/shell32/CDefViewBckgrndMenu.cpp
+++ b/dll/win32/shell32/CDefViewBckgrndMenu.cpp
@@ -24,7 +24,6 @@ class CDefViewBckgrndMenu :
         UINT m_LastFolderCMId;
 
         BOOL _bIsDesktopBrowserMenu();
-        BOOL _bCanPaste();
     public:
         CDefViewBckgrndMenu();
         ~CDefViewBckgrndMenu();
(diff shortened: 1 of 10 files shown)


KRosUser2:
Just tested : [reactos-gcc-i386-Debug-334c869b716b51617a9808577b29ba7beca49d5f](https://github.com/reactos/reactos/suites/9152226746/artifacts/425860975)
- "Paste" in Recycle Bin is indeed active only if "cut" is used on a file first
BUT
- "Paste" does nothing (expects move to Recycle Bin)
- If the file is "Pasted" somewhere else, Paste remains available in Recycle Bin, this is WRONG (eg : nothing to paste as not "cut" anymore) <<<==== **Remains to be fixed in this PR**
- If the file is Cut then deleted, Paste remains available in Recycle Bin, this is CORRECT (eg : nothing to paste as not "cut" anymore) but Paste operations is expected to fail as "file no longer available"

dll/win32/shell32/CDefViewBckgrndMenu.cpp
```diff
@@ -318,12 +281,7 @@ CDefViewBckgrndMenu::HandleMenuMsg2(UINT uMsg, WPARAM wParam, LPARAM lParam, LRE
 {
     if(m_folderCM)
```
HBelusca, line 282: Could take the opportunity for adding space after "if" (and removing the curly brackets?)

sdk/include/reactos/wine/debug.h
```diff
@@ -64,6 +64,12 @@ struct __wine_debug_channel
 };
 
 #define UNIMPLEMENTED WINE_FIXME("%s is UNIMPLEMENTED!\n", __FUNCTION__)
+#define UNIMPLEMENTED_DBGBREAK                                                                                         \
```
HBelusca, line 67: You may reduce the number of spaces before the backslash.

dll/win32/shell32/CDefViewBckgrndMenu.cpp
learn-more: no.

KRosUser2:
May fix this ticket too : CORE-18953
To be checked.

HBelusca:
@whindsaks Do you know whether these changes/fixes are still valid with the current shell state?

PR #4863, opened 2022-11-08
[SHELL32] CDrivesFolder: Unhardcode pqcminfo->idCmdFirst value

binarymaster:
DFM_MERGECONTEXTMENU handler works properly now.

Addendum to 64657051c3c.

Following @learn-more's suggestion :slightly_smiling_face: 
> The fix is the same as for the recently fixed 'Send to' fix, so please just send a PR for that.

JIRA issue: [CORE-18577](https://jira.reactos.org/browse/CORE-18577)

Cc @KRosUser2 to retest.

### DIFF
--- a/dll/win32/shell32/folders/CDrivesFolder.cpp
+++ b/dll/win32/shell32/folders/CDrivesFolder.cpp
@@ -286,38 +286,52 @@ HRESULT CALLBACK DrivesContextMenuCallback(IShellFolder *psf,
     GetVolumeInformationA(szDrive, NULL, 0, NULL, NULL, &dwFlags, NULL, 0);
 
 // custom command IDs
+#if 0 // Disabled until our menu building system is fixed
+#define CMDID_FORMAT        0
+#define CMDID_EJECT         1
+#define CMDID_DISCONNECT    2
+#else
+/* FIXME: These IDs should start from 0, however there is difference
+ * between ours and Windows' menu building systems, which should be fixed. */
 #define CMDID_FORMAT        1
 #define CMDID_EJECT         2
 #define CMDID_DISCONNECT    3
+#endif
 
     if (uMsg == DFM_MERGECONTEXTMENU)
     {
         QCMINFO *pqcminfo = (QCMINFO *)lParam;
 
         UINT idCmdFirst = pqcminfo->idCmdFirst;
+        UINT idCmd = 0;
         if (!(dwFlags & FILE_READ_ONLY_VOLUME) && nDriveType != DRIVE_REMOTE)
         {
             /* add separator and Format */
-            UINT idCmd = idCmdFirst + CMDID_FORMAT;
+            idCmd = idCmdFirst + CMDID_FORMAT;
             _InsertMenuItemW(pqcminfo->hmenu, pqcminfo->indexMenu++, TRUE, 0, MFT_SEPARATOR, NULL, 0);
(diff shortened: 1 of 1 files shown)


HBelusca (approved):
looks reasonable.

KRosUser2:
I'll test

dll/win32/shell32/folders/CDrivesFolder.cpp
```diff
@@ -295,29 +295,31 @@ HRESULT CALLBACK DrivesContextMenuCallback(IShellFolder *psf,
...
             /* add separator and Disconnect */
-            UINT idCmd = idCmdFirst + CMDID_DISCONNECT;
+            idCmd = idCmdFirst + CMDID_DISCONNECT;
             _InsertMenuItemW(pqcminfo->hmenu, pqcminfo->indexMenu++, TRUE, 0, MFT_SEPARATOR, NULL, 0);
             _InsertMenuItemW(pqcminfo->hmenu, pqcminfo->indexMenu++, TRUE, idCmd, MFT_STRING, MAKEINTRESOURCEW(IDS_DISCONNECT), MFS_ENABLED);
         }
 
-        pqcminfo->idCmdFirst += 3;
+        if (idCmd)
+            pqcminfo->idCmdFirst = (idCmd + 2);
```
learn-more, line 322: If you update the CMDID's to be 0-based, then  you can do +1 here,
which makes more sense since this is supposed to point to the next _free_ item.

binarymaster: I have applied your suggestion, it seems to break menu item position:
VirtualBox_ReactOS_08_11_2022_22_48_52

KRosUser2:
Seems OK now for "Loaded" CD Drive : 
image

but I now realize that it displays the "Format" option for an empty CD Drive, which is a non sense (and not displayed in Win XPSP3 btw)
image

binarymaster:
> but I now realize that it displays the "Format" option for an empty CD Drive, which is a non sense (and not displayed in Win XPSP3 btw)

I noticed it too, but in ReactOS it was the same for quite some time before :slightly_smiling_face:

dll/win32/shell32/folders/CDrivesFolder.cpp
learn-more: That this does not break more often is a game of chance,
looks like we do not treat those menu's correctly at all.
We regularly confuse index and id...

binarymaster: I guess I'll commit this one in current state, to avoid introducing one more regresso. 
:coffee:  :sweat_smile:  This has to be investigated deeper at some point.

learn-more: This perfectly highlights a difference between ours and windows' menu building system that I knew existed, but did not understand the reason for yet :)

I'll work on that but it might take some time, so this hackfix is good enough for now.

binarymaster: I have added `#if 0` with a comment for it. Also mentioned in the commit message. :+1:

KRosUser2:
> > but I now realize that it displays the "Format" option for an empty CD Drive, which is a non sense (and not displayed in Win XPSP3 btw)
> 
> I noticed it too, but in ReactOS it was the same for quite some time before 🙂

If not fixed by this PR i'll open a dedicated ticket.

PR #4864, opened 2022-11-08
[SETUP][INF] Use the standard "Helv" -> "MS Sans Serif" substitution

HBelusca:
## Purpose & Proposed changes

Use the standard "Helv" -> "MS Sans Serif" substitution, and let the latter one substitute to whatever we want.
Hindi however had "Helv" -> "Tahoma", but "MS Sans Serif" -> "FreeSans".
Now its "Helv" will go to "FreeSans".

Cc @arnavbhatt288 for testing Hindi changes.


arnavbhatt288:
I will gonna do the tests as soon as possible.

katahiromz:
Win2k Japanese:
Win2k

WinXP Japanese:
WinXP

Win2k3 English:
Win2k3

Win10 Japanese:
Win10

The PR seems legit.

katahiromz (approved):

JoachimHenze (changes requested):
The PR is legit ofc from a pure "registry-only" perspective. But "MS Sans Serif" is a raster font. And since we don't have any rasterfont support yet in ReactOS, it may not render correctly maybe. It should be tested first when selecting Helv in wordpad and writing a few words. If that would look shitty, then we should wait with merging that PR until we do have rasterfont support. Rejecting the PR until such a wordpad screenshot would be posted.

HBelusca:
> The PR is legit ofc from a pure "registry-only" perspective. But "MS Sans Serif" is a raster font. And since we don't have any rasterfont support yet in ReactOS, it may not render correctly maybe. It should be tested first when selecting Helv in wordpad and writing a few words. If that would look shitty, then we should wait with merging that PR until we do have rasterfont support. Rejecting the PR until such a wordpad screenshot would be posted.

I don't think you notice that in current master, our "Helv" is redirected to "Tahoma", and "MS Sans Serif" itself is redirected to none other than... the same "Tahoma" font. So I'm not sure I follow your argument here.

JoachimHenze (approved):
You are right. I missed that 2nd redirection. Fine then.

binarymaster (approved):

PR #4868, opened 2022-11-11
[SHELL32] Show 'comments' column for folders, make it rightmost CORE-18615

JoachimHenze:
In explorer filebrowser 'details' view
the column 'comments' had invalid length of zero, and therefore was hidden under the 'attributes' column.

I guess even after this patch, it will not display any sane contents in there yet.
Therefore I chose a small width of 10 only, to reduce the chance for it to trigger an undesired horizontal scrollbar.

JIRA issue: [CORE-18615](https://jira.reactos.org/browse/CORE-18615)

As far as I see we don't support manual showing and hiding of additional columns.
By default my Windows does not show this column. But when activated manually, it is always
shown on the right hand side of the Atributes column.

For a video of the patched state, check the JIRA ticket.


HBelusca (approved):
LGTM. No idea what's the standard ordering in Win2k3 though.

learn-more (approved):

JoachimHenze:
Is it safe to ignore the failed check and to squashnMerge anyway? I doubt that it comes from my changes.
Please let me push the button myself, I want to fix the typo in the commit comment when doing that "columns column" -> "comment" column when htting the squashnMerge by editing it.

PR #4871, opened 2022-11-11
[SETUP][INF][MEDIA] Get rid of the last Ubuntu references CORE-18607

JoachimHenze:
An addendum to 0.4.14-dev-20-g 2f4fb903b46207ab380365908401ded63752a108 because since then we don't have the Ubuntu font anymore.

The substitutes are also not needed any longer.
We can use Tahoma here without causing any change in the current rendering, because for all languages the former
Ubuntu substitutions did point either to Tahoma,
or to the same thing, that Tahoma atm points to
(for those language that do require additional glyphs).

This way we do not only get the substitutions closer to 2k3sp2, but will also simplify our maintenance and testing, because the same font is guaranteed to be used then for all themes: Classic, Blackshade, Modern and Lautus: The font which has the needed glyphs for that specific language.

E.g.
"FreeSans" for Hindi,
"Tahoma" for most Western languages, and
"Droid Sans Fallback" for Chinese and Japanese language.

JIRA issue: [CORE-18607](https://jira.reactos.org/browse/CORE-18607)

### DIFF
--- a/base/setup/lib/muifonts.h
+++ b/base/setup/lib/muifonts.h
@@ -122,7 +122,6 @@ MUI_SUBFONT HebrewFonts[] =
     { L"Times New Roman Greek,161",    L"Times New Roman,161" },
     { L"Times New Roman TUR,162",      L"Times New Roman,162" },
     { L"Tms Rmn",                L"Times New Roman" },
-    { L"Ubuntu",                 L"Tahoma" },
     { NULL, NULL }
 };
 
@@ -169,7 +168,6 @@ MUI_SUBFONT ChineseSimplifiedFonts[] =
     { L"Times New Roman TUR,162",      L"Times New Roman,162" },
     { L"Tms Rmn",                L"Times New Roman" },
     { L"Trebuchet MS",           L"Droid Sans Fallback" },
-    { L"Ubuntu",                 L"Droid Sans Fallback" },
     /* localized names */
     { CSF_LocalName0,            L"Droid Sans Fallback" },
     { CSF_LocalName1,            L"Droid Sans Fallback" },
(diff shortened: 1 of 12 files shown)


JoachimHenze:
patched displaying of Lautus and Modern:
Lautus
Modern
which is the same as before this PR.

katahiromz (approved):

HBelusca:
Ah ok, if the themes that needed the Ubuntu font, work fine without it ("patched as you say") and indeed we didn't have the ubuntu ttf file since some time now, and it was anyway substituted, then this should be fine.

base/setup/lib/muifonts.h
```diff
@@ -169,7 +168,6 @@ MUI_SUBFONT ChineseSimplifiedFonts[] =
     { L"Times New Roman TUR,162",      L"Times New Roman,162" },
     { L"Tms Rmn",                L"Times New Roman" },
     { L"Trebuchet MS",           L"Droid Sans Fallback" },
-    { L"Ubuntu",                 L"Droid Sans Fallback" },
```
HBelusca, line 172: Can you double-check, that things look "ok" (as good or as bad) when you install ROS in either one of chinese, japanese or korean, with those themes enabled?

HBelusca (approved):

base/setup/lib/muifonts.h
JoachimHenze: Yes they will not be changed because
MUI_SUBFONT ChineseSimplifiedFonts[] =
...
    { L"Tahoma",                 L"Droid Sans Fallback" },

and
MUI_SUBFONT JapaneseFonts[] =
    { L"Tahoma",                 L"Droid Sans Fallback" },

and
MUI_SUBFONT KoreanFonts[] =
    { L"Tahoma",                 L"Droid Sans Fallback" },

I cannot test that, because the "Droid Sans Fallback" font gets copied into the iso only for releases, and not for master builds. But like I wrote in the description already: They won't be changed by the PR!

PR #4873, opened 2022-11-12
[DISKPART] Fix help for multi-word command crashing.

TheNNX:
## Purpose

When the HELP command is executed for a multi-word command, for example "HELP CREATE PARTITION PRIMARY", diskpart crashes. This happens because before the desired form of the command is encountered, its single word form is found (in case of this example - "CREATE") - since the first word of the command matches the one HELP is looking for, it tries to compare the second word of the input command with the second (nonexistent) word of the encountered single word command, which is NULL. This results in diskpart crashing.

## Proposed changes

This fix makes HELP check if the to-be-compared word is not NULL before calling wcsicmp.

### DIFF
--- a/base/system/diskpart/help.c
+++ b/base/system/diskpart/help.c
@@ -60,7 +60,7 @@ HelpCommand(
     {
         if (pCommand->cmd1 != NULL && pCommand->cmd2 == NULL && pCommand->cmd3 == NULL)
         {
-            if ((wcsicmp(pCommand->cmd1, cmdptr->cmd1) == 0) &&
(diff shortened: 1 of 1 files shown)


HBelusca (approved):

JoachimHenze (approved):

JoachimHenze:
@TheNNX Can you please disable anonymous email for your future contributions? Thank you!

binarymaster:
Yes, please disable "Keep my email addresses private" in (email settings)[https://github.com/settings/emails].

Because this has broken email when merging this PR:

image

JoachimHenze:
at least for our own future reference in this oopsie-case:
image

PR #4877, opened 2022-11-13
[IMM32] Introduce 'security' code

katahiromz:
## Purpose
Add 'security' code for future use.

## Proposed changes

- Add `Imm32InitSecurity` and `Imm32GenerateSecurityCookie` helper functions.
- Add `__security_cookie` and `__security_cookie_complement` global variables.
- Add `__security_check_cookie` function to detect stack corruption.

## TODO

- [x] Do small tests.

### DIFF
--- a/dll/win32/imm32/imm.c
+++ b/dll/win32/imm32/imm.c
@@ -1350,7 +1350,7 @@ VOID FASTCALL __security_check_cookie(DWORD_PTR ecx)
 #ifndef NDEBUG
     DebugBreak();
 #endif
-    TerminateProcess(GetCurrentProcess(), 0xC0000409); // STATUS_STACK_BUFFER_OVERRUN
+    TerminateProcess(GetCurrentProcess(), STATUS_STACK_BUFFER_OVERRUN);
 }
 #endif /* def IMM_SECURITY */
 
--- a/dll/win32/imm32/imm.c
+++ b/dll/win32/imm32/imm.c
@@ -1310,9 +1310,9 @@ BOOL WINAPI ImmSetActiveContextConsoleIME(HWND hwnd, BOOL fFlag)
 }
 
 /************************************************************************/
-/* IMM_SECURITY */
+/* STACK_CHECK */
 
-#ifdef IMM_SECURITY
+#ifdef STACK_CHECK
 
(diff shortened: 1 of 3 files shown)


dll/win32/imm32/imm.c
```diff
@@ -1296,6 +1309,73 @@ BOOL WINAPI ImmSetActiveContextConsoleIME(HWND hwnd, BOOL fFlag)
...
 
+/************************************************************************/
+/* STACK_CHECK */
+
+#ifdef STACK_CHECK
+
+DWORD_PTR __security_cookie = 0x0000BB40;
+DWORD_PTR __security_cookie_complement = 0xFFFF44BF;
+
+static VOID Imm32GenerateSecurityCookie(VOID)
```
learn-more, line 1320: Why are you re-implementing the stack checks the compiler inserts?

katahiromz: How can we insert the stack checks into `IMM32`?

katahiromz: https://learn.microsoft.com/en-us/archive/msdn-magazine/2017/december/c-visual-c-support-for-stack-based-buffer-protection

katahiromz:
Closed.

dll/win32/imm32/imm.c
HBelusca: But all that is automatically done when you compile any module and the CRT is linked to it...
Now, I've tried to look at our compilation settings and I see that by default, `/GS` appears to be disabled (see the msvc.cmake results here:
https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=%2FGS
)
However I don't know whether they get reenabled somehow later.

The `_security_cookie` stuff in the CRT is implemented here:
https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=security_cookie

learn-more: Here is what {{Imm32LoadImeDpi}} looks like on current master (or close to it anyway) in msvc:

image

image

HBelusca: This is indeed RTC "only". The GuardStack feature (with "security cookies") doesn't seem to be enabled indeed.

PR #4878, opened 2022-11-13
[KBDTOOL] Enforce VirtualKey type and max value

SergeGautherie:
Fix broken max value handling.
Optimize a bit.

JIRA issue: [CORE-16075](https://jira.reactos.org/browse/CORE-16075)

### DIFF
--- a/sdk/tools/kbdtool/output.c
+++ b/sdk/tools/kbdtool/output.c
@@ -714,6 +714,7 @@ kbd_c(IN ULONG StateCount,
         if ((Layout->Entry[i].ScanCode & 0xFF00) == 0xE000)
         {
             /* Which are valid */
+            // FIXME: This check is always true.
             if (Layout->Entry[i].ScanCode != 0xFF)
             {
                 /* And mapped */
@@ -743,6 +744,7 @@ kbd_c(IN ULONG StateCount,
         if ((Layout->Entry[i].ScanCode & 0xFF00) == 0xE100)
         {
             /* Which are valid */
+            // FIXME: This check is always true.
             if (Layout->Entry[i].ScanCode != 0xFF)
             {
                 /* And mapped */
--- a/sdk/tools/kbdtool/parser.c
+++ b/sdk/tools/kbdtool/parser.c
@@ -62,7 +62,7 @@ isKeyWord(PCHAR p)
 }
 
 PCHAR
-getVKName(IN ULONG VirtualKey,
+getVKName(IN UCHAR VirtualKey,
           IN BOOLEAN Prefix)
 {
     ULONG i;
@@ -108,7 +108,7 @@ getVKName(IN ULONG VirtualKey,
     return gVKeyName;
 }
 
-ULONG
+UCHAR
 getVKNum(IN PCHAR p)
 {
     ULONG Length;
@@ -719,7 +719,6 @@ DoLAYOUT(IN PLAYOUT LayoutData,
     ULONG KeyWord;
     ULONG ScanCode, CurrentCode;
     ULONG TokenCount;
-    ULONG VirtualKey;
     ULONG i;
     ULONG Count;
     BOOLEAN FullEntry;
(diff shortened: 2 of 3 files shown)


HBelusca:
What are you trying to actually fix? Nobody is using the KBDTOOL yet.

SergeGautherie:
> What are you trying to actually fix? Nobody is using the KBDTOOL yet.

I am trying to resolve the Jira ticket about the LGTM.com report,
by reviewing and fixing related code.
That should help whenever someone start using this tool.

sdk/tools/kbdtool/output.c
```diff
@@ -714,6 +714,7 @@ kbd_c(IN ULONG StateCount,
         if ((Layout->Entry[i].ScanCode & 0xFF00) == 0xE000)
         {
             /* Which are valid */
+            // FIXME: This check is always true.
             if (Layout->Entry[i].ScanCode != 0xFF)
```
SergeGautherie, line 718: Guess: missing `& 0x00FF`? To be investigated (as a follow-up)...

HBelusca:
> > What are you trying to actually fix? Nobody is using the KBDTOOL yet.
> 
> I am trying to resolve the Jira ticket about the LGTM.com report, by reviewing and fixing related code. That should help whenever someone start using this tool.

OK but then, there are the patches of CORE-9261 to consider first. (And someone else may already have started to deal with those.)

SergeGautherie:
> there are the patches of CORE-9261 to consider first. (And someone else may already have started to deal with those.)

It looks rather like that issue and patch have been abandoned 7+ years ago.
Nonetheless, I searched its kbdtool_UTF.patch and it does not touch any of what my PR does.

SergeGautherie:
Looking for a reviewer.

SergeGautherie:
Looking for a reviewer.

SergeGautherie:
Looking for a reviewer.

sdk/tools/kbdtool/parser.c
```diff
@@ -146,8 +146,8 @@ getVKNum(IN PCHAR p)
     if ((*p == '0') && ((*(p + 1) == 'x') || (*(p + 1) == 'X')))
     {
         /* Get the key number from the hex string */
-        *(p + 1) = 'x';
-        if (sscanf(p, "0x%x", &KeyNumber) == 1) return KeyNumber;
+        if (sscanf(p + 2, "%x", &KeyNumber) == 1 && KeyNumber <= 0xFF)
+            return KeyNumber;
```
katahiromz, line 150:
```suggestion
            return (UCHAR)KeyNumber;
```

katahiromz (approved):

SergeGautherie:
(Rebased, to trigger current master builds.)

sdk/tools/kbdtool/output.c
```diff
@@ -714,6 +714,7 @@ kbd_c(IN ULONG StateCount,
         if ((Layout->Entry[i].ScanCode & 0xFF00) == 0xE000)
         {
             /* Which are valid */
+            // FIXME: This check is always true.
```
HBelusca, line 717: How so? If someone makes a layout with a scancode value `0xE0FF` ?
(And same question for the similar fixme below.)

SergeGautherie: After line 714 check, we know `ScanCode >= 0xE000`, hence it cannot be `0x00FF`.

sdk/tools/kbdtool/parser.c
JoachimHenze, line 150: Last and only question from my side before I will give my approval for this PR: What exactly (which symptom) do the changes in lines149+150 do fix actually? I am okay with every other change proposed in this PR, but don't understand this one. Thx.

JoachimHenze: would it make sense here to do the cast proposed by Hermes, and stip the last condition of the if in return? Or would that be wrong?

sdk/tools/kbdtool/output.c
HBelusca: Unless the purpose of the second test was to check the lowest byte of the scan code? (And they forgot the `& 0xFF`)

sdk/tools/kbdtool/parser.c
JoachimHenze: so you say your intent is to actively filter everything >= 0xFF and not return that as a KeyNumber. Why?

SergeGautherie:
Looking for a committer.

DarkFire01:
Since this isn't used right now i'm going to close it. feel free to reopen if you wish to work on it still. 
As this PR has clearly died.

PR #4881, opened 2022-11-14
[DOCS] Update contributing rules and PR management

binarymaster:
This PR is a result of discussion (links: [one](https://chat.reactos.org/reactos/pl/s1xsp4b1s38o9kx3dc41fy871e), [two](https://chat.reactos.org/reactos/pl/da3ozt58k3bnukpj96ip9n119a)) in the chat with @JoachimHenze and @ThFabba regarding noreply e-mails being used by some GitHub contributors.

It has been experimentally verified that using anonymous GitHub e-mail does not allow to deliver mail to the commit author. GitHub does replace author's real email when `Keep my email addresses private` option is enabled in the account's e-mail settings.

Additionally I have replaced old IRC links with the references to our chat.

Please approve if you are okay with the changes I made in this PR. Suggestions are also welcome.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,14 +64,12 @@ Our commit style is defined in a __[commit template]__. Use it as a reference or
 ### Rules and Recommendations
 
 - *Use your __real name__ and __real email__.* We do not accept anonymous code contributions!
-  - Every commit that changes the code or translations should have author's full legal name (in latin letters, diacritics allowed) with real e-mail address.
-  - Same applies to the author's GitHub profile:
-  1. It should have full name set (in the [Name field here][GitHub Profile Settings]) that matches one specified in the commits.
-  1. In the [e-mail settings][GitHub Email Settings] the checkbox "Keep my email addresses private" must be unchecked.
-- There is an exception for media changes, such as changes of the art (wallpapers, themes, icons, sounds) and out-of-code documentation. In these specific cases it's allowed to use a nickname or alias. Few rules still apply here:
-  1. Artist must have an alias set (in the [Name field here][GitHub Profile Settings]) and it should match one specified in the commits.
-  1. Real e-mail address is still required, so the checkbox "Keep my email addresses private" must be unchecked in the [e-mail settings][GitHub Email Settings].
-- In order to *keep your privacy*, select appropriate "Primary email address" that will be applied to your commits in GitHub [e-mail settings][GitHub Email Settings].
+  - Every commit that changes the code or translations should have author's full legal name (in latin letters, diacritics allowed).
+  - Same applies to the author's GitHub profile, it should have full name set (in the [Name field here][GitHub Profile Settings]) that matches one specified in the commits.
+- There is an exception for media changes, such as changes of the art (wallpapers, themes, icons, sounds) and out-of-code documentation.
+  - In these specific cases it's allowed to use a nickname or alias, however the artist name (in the [Name field here][GitHub Profile Settings]) should match one specified in the commits.
+- In any case the author must use a real e-mail address, this includes git commits (`user.email` setting) and GitHub [e-mail settings][GitHub Email Settings] - the checkbox "Keep my email addresses private" must be unchecked there.
+  - In order to *keep your privacy*, select appropriate "Primary email address" that will be applied to your commits in GitHub [e-mail settings][GitHub Email Settings].
 - *Ensure your contribution is properly described.* Include the relevant issue number if applicable.
 - *Put only related changes.* It will make reviewing easier as the reviewer needs to recall less information about the existing source code that is changed.
 - *Search for similar pull requests/patches before submitting.* It may be that a similar pull request or issue was opened previously. Comment and review on that one instead.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,10 +64,10 @@ Our commit style is defined in a __[commit template]__. Use it as a reference or
 ### Rules and Recommendations
 
 - *Use your __real name__ and __real email__.* We do not accept anonymous code contributions!
-  - Every commit that changes the code or translations should have author's full legal name (in latin letters, diacritics allowed).
-  - Same applies to the author's GitHub profile, it should have full name set (in the [Name field here][GitHub Profile Settings]) that matches one specified in the commits.
-- There is an exception for media changes, such as changes of the art (wallpapers, themes, icons, sounds) and out-of-code documentation.
-  - In these specific cases it's allowed to use a nickname or alias, however the artist name (in the [Name field here][GitHub Profile Settings]) should match one specified in the commits.
+  - Every commit that changes code or translations should have author's full legal name (in latin letters, diacritics allowed).
+  - It's recommended to have the same full name set in GitHub profile (in the [Name field here][GitHub Profile Settings]) that matches one specified in commits.
+- There is an exception for media changes, such as changes of art (wallpapers, themes, icons, sounds) and out-of-code documentation.
+  - In these specific cases it's allowed to use a nickname or alias as author's name, and it's recommended to have the same name set in GitHub profile (in the [Name field here][GitHub Profile Settings]) matching one specified in commits.
 - In any case the author must use a real e-mail address, this includes git commits (`user.email` setting) and GitHub [e-mail settings][GitHub Email Settings] - the checkbox "Keep my email addresses private" must be unchecked there.
   - In order to *keep your privacy*, select appropriate "Primary email address" that will be applied to your commits in GitHub [e-mail settings][GitHub Email Settings].
 - *Ensure your contribution is properly described.* Include the relevant issue number if applicable.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -64,12 +64,16 @@ Our commit style is defined in a __[commit template]__. Use it as a reference or
 ### Rules and Recommendations
 
 - *Use your __real name__ and __real email__.* We do not accept anonymous contributions!
+  - Every commit should have author's full legal name (in latin letters, diacritics allowed) with real e-mail address.
+  - Same applies to the author's GitHub profile:
+  1. It should have full name set (in the [Name field here](https://github.com/settings/profile)) that matches one specified in the commits.
+  1. In the [e-mail settings](https://github.com/settings/emails) the checkbox "Keep my email addresses private" must be unchecked.
 - *Ensure your contribution is properly described.* Include the relevant issue number if applicable.
 - *Put only related changes.* It will make reviewing easier as the reviewer needs to recall less information about the existing source code that is changed.
 - *Search for similar pull requests/patches before submitting.* It may be that a similar pull request or issue was opened previously. Comment and review on that one instead.
 - *Keep your contribution small and focused on the topic.* It can be tempting to fix existing issues as you come across them while reading the source code. Resist the temptation and put in a note in the source code instead, or (even better) put the issue in the issue tracking system.
 - *Respect our __[Coding Style]__ and __[Programming Guidelines]__.*
-- *Do not be afraid to ask questions.* Ask our developers on JIRA or [IRC] channel.
+- *Do not be afraid to ask questions.* Ask our developers in the [chat].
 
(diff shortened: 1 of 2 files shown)


PULL_REQUEST_MANAGEMENT.md
```diff
@@ -14,3 +14,8 @@ In addition, in order to avoid coming off as rude to helpful contributors, pleas
 - Asking the contributor to do unrelated work
 - Closing without providing a reason
 - Merging with the intention to rewrite that code soon after
+
+Before merging a PR, make sure it follows the [contributing rules](CONTRIBUTING.md#rules-and-recommendations), but more importantly:
+- Make sure the author is not using a nickname or alias, but a full legal name instead, and have specified a real e-mail in all commits
```
JoachimHenze, line 19: I would vote to weaken that sentence on line19 to:
Make sure the author is using a real e-mail in all commits

I think real names instead of aliases should be preferred yet, but I think it is too strict to forbid aliases entirely.

binarymaster: From the legal point, it would be better to forbid aliases, however some projects do allow aliases in case the author has sent a letter/document to the project's legal team, asking to associate an alias with the author.

CONTRIBUTING.md
```diff
@@ -64,12 +64,16 @@ Our commit style is defined in a __[commit template]__. Use it as a reference or
 ### Rules and Recommendations
 
 - *Use your __real name__ and __real email__.* We do not accept anonymous contributions!
+  - Every commit should have author's full legal name (in latin letters, diacritics allowed) with real e-mail address.
```
GeoB99, line 67: As it currently stands, this rule regarding the exposure of your real name is optional for people such as artists and translators whereas your wording of this PR makes the said rule compulsory for everyone. That's too rigorous and I'm not so keen about it (plus we had someone who ranted about it on Twitter sometime ago but I would rather not want go into specifics).

I would suggest that you at least make such rule mandatory only for people who contribute with code to the project and if someone is too paranoid to use their legal name for contribution due to understandable reasons such as privacy, we should adopt the approach of accepting user aliases where a contributor signs a legal document (which should include their real name for legal reasons obviously) so they can use an alias that is easily recognizable so no stupid crap nicknames like `trololololo` or `Xxx CykaBloat xxX`. The ReactOS project is fully responsible to not disclose such document at this point.

GeoB99 (changes requested):

CONTRIBUTING.md
binarymaster: Such thing as "optional for people such as artists and translators" is actually documented nowhere. I believe it was in an open discussion without the final decision, and that is why current contributing rules do not include it and just require the real name.

> plus we had someone who ranted about it on Twitter sometime ago

If I remember correctly, that was because PR was merged, but we required the full name from a person only after that. Good reason to make a final decision on that, to avoid these rants in future.

CONTRIBUTING.md
```diff
@@ -64,12 +64,16 @@ Our commit style is defined in a __[commit template]__. Use it as a reference or
...
+  - Same applies to the author's GitHub profile:
+  1. It should have full name set (in the [Name field here](https://github.com/settings/profile)) that matches one specified in the commits.
+  1. In the [e-mail settings](https://github.com/settings/emails) the checkbox "Keep my email addresses private" must be unchecked.
```
SergeGautherie, line 70: Then, contributing to ReactOS would force to disclose all addresses declared on GitHub? Yuck!

PULL_REQUEST_MANAGEMENT.md
```diff
@@ -14,3 +14,9 @@ In addition, in order to avoid coming off as rude to helpful contributors, pleas
 - Asking the contributor to do unrelated work
 - Closing without providing a reason
 - Merging with the intention to rewrite that code soon after
+
+Before merging a PR, make sure it follows the [contributing rules](CONTRIBUTING.md#rules-and-recommendations), but more importantly:
+- Make sure the author does not use a nickname or alias, but a full legal name instead, and have specified a real e-mail in all PR commits
+- Make sure the author does have full name set in GitHub profile that matches one in commits
+- By pressing "Squash and merge" button in a PR you can make sure the author does not use no-reply e-mail -
```
SergeGautherie, line 21: As discussed on an older PR of mine, real name and address used to be forwarded from PR commits to official repository.
Why not ask GitHub to fix current mess instead?

Also, what about even worse `committer GitHub <noreply@github.com>`?

CONTRIBUTING.md
binarymaster: There is always a possibility to remove addresses from GitHub that you don't want to disclose. Personally I have only two emails associated with GitHub, one of which I use for signing code changes, and another for translations.

PULL_REQUEST_MANAGEMENT.md
binarymaster: If all rules are met by the author (profile), there should be no problem regardless of the merge method used by a maintainer.

> Also, what about even worse `committer GitHub <noreply@github.com>`?

It's a corner case which is allowed.

binarymaster: > Why not ask GitHub to fix current mess instead?

If GitHub decides to change interface with a feature to manually specify commit author's email, or select it from existing commits in the PR, then the rules could be updated.

Asking something always takes some time. We have limited manpower with limited time here.

JoachimHenze: "We have limited manpower with limited time here, and despite that we must keep progressing."

Yes, and that is exactly the reason why I think we shouldn't make it too hard for external people to help us.

binarymaster: I know there are some people who still helping us under an alias (e.g. I_Kill_Bugs), however they don't create PRs directly, instead maintainers themselves create commits based on their patches.

SergeGautherie: > If all rules are met by the author (profile), there should be no problem regardless of the merge method used by a maintainer.

Which rules? A "no privacy" one? Select a process that rather does not require that...

> It's a corner case which is allowed, only for the committer of course.

Corner case? I would rather call it major, given the high number of affected commits.
Then, committer remains totally anonymous (as in that commit field is 100% useless), while author privacy is forbidden: NOT great!

> If GitHub decides to change interface with a feature to manually specify commit author's email, or select it from existing commits in the PR, then the rules could be updated.

Such feature behavior seems wanted, indeed.
In the meantime, that is nonsense: lost privacy could not be recovered afterward.

> Asking something always takes some time. It takes even more time to actually get things changed. We have limited manpower with limited time here, and despite that we must keep progressing.

So what? Did you even ask GitHub first? Why not just not use that feature if it doesn't work as needed (anymore/yet)?

CONTRIBUTING.md
HBelusca: We obviously don't care at all about your own private email; However, how would one contact you without knowing *some* email? You use your email to contact privately some ReactOS devs there, so the reverse should be possible. Use/create an email address that is meant to be public and tell github to use that latter one as the "public" one (while keeping the one used to manage your own github settings private)...

HBelusca: image

image

ColinFinck: What @HBelusca said. It's fine to set up one public e-mail address to work on the ReactOS repo, and this should be a requirement.
But disclosing all addresses connected to a GitHub account serves no additional purpose.

HBelusca:
Any news on this PR?

PULL_REQUEST_MANAGEMENT.md
binarymaster: I have added a "Keep your privacy" point to the README.md.

CONTRIBUTING.md
binarymaster: I have added an exception point for the artists (changes related to wallpapers, themes, icons, sounds), please review. 🙂

binarymaster:
> Any news on this PR?

@HBelusca yes, please take a look at it again. I have updated some points per review.

CONTRIBUTING.md
```diff
@@ -63,13 +63,19 @@ Our commit style is defined in a __[commit template]__. Use it as a reference or
 
 ### Rules and Recommendations
 
-- *Use your __real name__ and __real email__.* We do not accept anonymous contributions!
+- *Use your __real name__ and __real email__.* We do not accept anonymous code contributions!
+  - Every commit that changes the code or translations should have author's full legal name (in latin letters, diacritics allowed).
+  - Same applies to the author's GitHub profile, it should have full name set (in the [Name field here][GitHub Profile Settings]) that matches one specified in the commits.
+- There is an exception for media changes, such as changes of the art (wallpapers, themes, icons, sounds) and out-of-code documentation.
```
JoachimHenze, line 69: I don't understand why there should be any exception for artists: A media file crafted with evil intention can also be used to exploit a security vulnerability within the OS. I don't see any good reason to treat such files more relaxed than any other contribution. Also it is more text, and weakens a rule that could be more general and therefore easier to remember and easier to be established.
The exception might also lower our standards in those files, causing bloat and other useless eye-candy to sneak into the project more easily.

binarymaster: This is valid critics, but see George's comment: https://github.com/reactos/reactos/pull/4881#discussion_r1022073965

GeoB99: Media files with exploit code crafted on them could be checked before merging a PR. It is quite cumbersome if there are multiple malicious files but it is not that hard, also the chance of it happening is quite slim so I don't see what is the paranoia here.

We haven't applied the mandatory real name usage rule to artists retroactively and as such we should not apply it now and beyond after changing the contribution rules either. If you want it mandatory for everyone regardless of the nature of a contribution, then I would suggest a creation of a legal document where a person could request to have their real name hidden from all of their contributions (see my comment above). Otherwise I'm just simply totally against it.

GeoB99 (approved):

CONTRIBUTING.md
```diff
@@ -63,13 +63,19 @@ Our commit style is defined in a __[commit template]__. Use it as a reference or
 
 ### Rules and Recommendations
 
-- *Use your __real name__ and __real email__.* We do not accept anonymous contributions!
+- *Use your __real name__ and __real email__.* We do not accept anonymous code contributions!
+  - Every commit that changes the code or translations should have author's full legal name (in latin letters, diacritics allowed).
+  - Same applies to the author's GitHub profile, it should have full name set (in the [Name field here][GitHub Profile Settings]) that matches one specified in the commits.
```
learn-more, line 68: Why do we want to enforce people to fill in their github profile?

HBelusca: If I understand correctly, it's to be able to press the "Squash & Merge" button in their PRs (from the github webinterface) so as to get their name OK. Instead to do the whole work of checking their PR locally and manually do the adjustments.

binarymaster: Yes, for the same reason the author's GitHub name display has been added into the [tampermonkey script](https://github.com/learn-more/tampermonkey) for ReactOS PRs.

learn-more: I am not in favor of forcing people to set that.

learn-more (approved):

PULL_REQUEST_MANAGEMENT.md
SergeGautherie: > Did you even ask GitHub first?

Yesterday, I eventually asked GitHub about this privacy with regard to multiple emails issue.
Hoping for an answer...

CONTRIBUTING.md
SergeGautherie: > Use/create an email address that is meant to be public and tell github to use that latter one as the "public" one (while keeping the one used to manage your own github settings private)...

My Git commits (except the ones I created through GitHub web interface on a couple recent PRs...) do have a usable address.

Your images show a setup where all addresses, including the primary one, are public, which is not the privacy I want to use.

> Also for unknown reasons, this doesn't seem to cause problems with Wine developers, they manage to retain their private emails private, but are able to use public ones.

Interesting. It would be worth asking them how to do that...

PULL_REQUEST_MANAGEMENT.md
SergeGautherie: > Hoping for an answer...

GitHub support response:
```
As you have discovered, if you have email privacy enabled, your noreply email address is the default commit author email address.
Unfortunately you cannot choose which email addresses to keep private and which email addresses to make public.
```

CONTRIBUTING.md
SergeGautherie: > set up one public e-mail address to work on the ReactOS repo,

As far as I remember, noone explained how to make that work on GitHub.
See https://github.com/reactos/reactos/pull/4881#discussion_r1295884940, which confirms what I wrote all along.

> disclosing all addresses connected to a GitHub account serves no additional purpose.

This is why I consider new ROS rules to be an infringement and I will not comply with them.

PR #4888, opened 2022-11-19
[SHELL32] Prevent a second call to  Drive Properties Dialog

zecarlos1957:
Clicking on the Properties item in the drives context menu will open two Drives Properties dialogs one after the other.

CDefaultContextMenu::DoProperties provides a fallback call to property sheet testing the return value of the _DoCallback method, which is ultimately the return value of SH_ShowDriveProperties(...).
_Do Callback returns 0 (on error) and 1(on success), S_FALSE is defined as 1. 

https://jira.reactos.org/browse/CORE-18613

### DIFF
--- a/dll/win32/shell32/CDefaultContextMenu.cpp
+++ b/dll/win32/shell32/CDefaultContextMenu.cpp
@@ -842,7 +842,7 @@ CDefaultContextMenu::DoProperties(
     HRESULT hr = _DoCallback(DFM_INVOKECOMMAND, DFM_CMD_PROPERTIES, NULL);
 
     // We are asked to run the default property sheet
-    if (hr != S_FALSE)
+    if (hr == S_FALSE)
     {
         return Shell_DefaultContextMenuCallBack(m_psf, m_pDataObj);
     }
--- a/dll/win32/shell32/CDefaultContextMenu.cpp
+++ b/dll/win32/shell32/CDefaultContextMenu.cpp
@@ -842,7 +842,7 @@ CDefaultContextMenu::DoProperties(
     HRESULT hr = _DoCallback(DFM_INVOKECOMMAND, DFM_CMD_PROPERTIES, NULL);
 
     // We are asked to run the default property sheet
-    if (hr == S_FALSE)
+    if (hr != S_FALSE)
     {
         return Shell_DefaultContextMenuCallBack(m_psf, m_pDataObj);
     }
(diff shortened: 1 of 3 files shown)


KRosUser2:
Link to Jira ticket is missing

dll/win32/shell32/CDefaultContextMenu.cpp
```diff
@@ -842,7 +842,7 @@ CDefaultContextMenu::DoProperties(
     HRESULT hr = _DoCallback(DFM_INVOKECOMMAND, DFM_CMD_PROPERTIES, NULL);
 
     // We are asked to run the default property sheet
-    if (hr == S_FALSE)
+    if (hr == FALSE)
```
learn-more, line 845: `_DoCallback` returns an `HRESULT`, and `FALSE` is not an `HRESULT` so this is not a fix but a hack.

learn-more (changes requested):
This is not a fix, but rather a workaround or a hack.

dll/win32/shell32/CDefaultContextMenu.cpp
hpoussin: `SH_ShowDriveProperties` is marked as returning a `BOOL`, but sometimes returns a `HRESULT`.
Then, `DrivesContextMenuCallback` handles this result as an `HRESULT`.
This needs to be fixed.

dll/win32/shell32/folders/CDrivesFolder.cpp
```diff
@@ -345,7 +345,7 @@ HRESULT CALLBACK DrivesContextMenuCallback(IShellFolder *psf,
         {
             // pdtobj should be valid at this point!
             ATLASSERT(pdtobj);
-            hr = SH_ShowDriveProperties(wszBuf, pdtobj);
+            hr = SH_ShowDriveProperties(wszBuf, pdtobj) ? 0 : (HRESULT)-1;
```
learn-more, line 348:
```suggestion
            hr = SH_ShowDriveProperties(wszBuf, pdtobj) ? `S_OK` : `E_UNEXPECTED`;
```
Alternatively, make `SH_ShowDriveProperties` return an `HRESULT`

learn-more (approved):

binarymaster (approved):
I have tested this PR, but according to the test results, it **does not** fix [CORE-18613](https://jira.reactos.org/browse/CORE-18613), but fixes [CORE-18537](https://jira.reactos.org/browse/CORE-18537) instead. :smile:

HBelusca (approved):

PR #4890, opened 2022-11-19
[LIBXML2][LIBXSLT] Sync to 2.9.14 / 1.1.35

ThFabba:
Syncing using automation... PR should help find any bugs in my script.

JIRA issue: [CORE-17766](https://jira.reactos.org/browse/CORE-17766)

Sync scripts I used:
* libxml2: https://gist.github.com/ThFabba/ef1866e447d0c928f37643cfaddad8e3
* libxslt: https://gist.github.com/ThFabba/a6574d04b353e56bec7d07d613799bfc

Things to watch for:
* Check the "Improve sync with upstream" commits, since they make changes to ROS-diffs etc
* Check headers, especially config.h-type files
* Check for any unintentional changes to ROS-diffs (you can see all the ROS-diffs in my script, even though the sed/perl stuff is difficult to read)
* In libxslt, check `precomp.h`, which is entirely ours and has been updated to reflect upstream changes


sdk/lib/3rdparty/libxml2/config.h
```diff
@@ -138,56 +90,30 @@
...
 #define HAVE_STAT 1
 
-/* Define to 1 if you have the <stdarg.h> header file. */
-#define HAVE_STDARG_H 1
-
 /* Define to 1 if you have the <stdint.h> header file. */
 #define HAVE_STDINT_H 1
 
+/* Define to 1 if you have the <stdio.h> header file. */
+#undef HAVE_STDIO_H
```
learn-more, line 103: This seems weird?

ThFabba: Yeah it does... it's not present in `config.h.cmake.in`, which is checked into their git repository. I'm basing this off `config.h.in`, which is shipped in the release archives, and indeed this line was added there between 2.10.1 and 2.10.2.
Guess I should inquire where that file even comes from.

ThFabba: Found it... `config.h.in` is generated by autoheader, and the autoconf version used was updates with 2.10.2. `HAVE_STDIO_H` was indeed added, by https://git.savannah.gnu.org/gitweb/?p=autoconf.git;a=commitdiff;h=b045574cb2ea1c8441be7a7b56fd471d704ebf3a
So this is correct and intentional.

Thanks for catching it, though, this is exactly the kind of stuff I'm looking for!

sdk/include/reactos/libs/libxml/xmlversion.h
```diff
@@ -475,6 +484,15 @@ XMLPUBFUN void XMLCALL xmlCheckVersion(int version);
...
  */
 #define LIBXML_ATTR_FORMAT(fmt,args)
+/**
+ * XML_DEPRECATED:
+ *
+ * Macro used to indicate that a function, variable, type or struct member
+ * is deprecated.
+ */
+#ifndef XML_DEPRECATED
+#define XML_DEPRECATED
```
learn-more, line 494: Not the purpose of a sync, but would it make sense to add `__declspec(deprecated)` or `[[deprecated]]` here?

sdk/lib/3rdparty/libxml2/config.h
learn-more: But we have `stdio.h`, so shouldn't this be defined to `1` ?

sdk/include/reactos/libs/libxml/xmlversion.h
```diff
@@ -171,7 +171,7 @@ XMLPUBFUN void XMLCALL xmlCheckVersion(int version);
  *
  * Whether the FTP support is configured in
  */
-#if 1
+#if 0
 #define LIBXML_FTP_ENABLED
```
learn-more, line 175: This one makes sense to have disabled, but it was enabled before.
Did you disable this intentionally?

sdk/lib/3rdparty/libxml2/config.h
ThFabba: Yes, it should be. Good catch!

sdk/include/reactos/libs/libxml/xmlversion.h
ThFabba: Version 2.10.0 changed the default from 1 to 0, so I didn't do it explicitly, it just happened.

sdk/include/reactos/libs/libxml/xmlversion.h
ThFabba: Yes, it would. The libxml2 2.10.0 sync actually breaks libxslt because it deprecates a bunch of stuff (libxslt 1.1.36 fixes that) -- but only on GCC. So it would indeed be nice to give MSVC a hint too.

ThFabba:
Since the deprecation of functions in libxml2 2.10.0 also breaks msxml3 and therefore requires a partial Wine sync, I'm going to split things up and reduce this PR to everything before the 2.10.0 sync.

PR #4892, opened 2022-11-20
[NTOS:KD][KDBG] Rework the BootPhase >= 2 initialization of the KD/KDBG kernel debugger.

HBelusca:
## Purpose

Fix the way how (and when) we initialize KD(BG) functionality that depends on the NT I/O Manager to be up & running.
This includes, creating the debug log file + its thread, and for KDBG, loading the KDBinit file.

JIRA issue: [CORE-17470](https://jira.reactos.org/browse/CORE-17470)

## Proposed changes

**_NOTE: A hack for fixing FILE_APPEND_DATA support appears to be needed, because ReactOS doesn't fully honour it (files get overwritten instead). This will potentially be fixed in a separate PR, see [CORE-18789](https://jira.reactos.org/browse/CORE-18789)_**

Remove the ReactOS-specific (NT-incompatible) callback we do in the middle of IoInitSystem(), for a runtime mechanism that should work on Windows as well.

The idea of the new mechanism is loosely inspired by the TDL4 rootkit :sunglasses:
http://blog.w4kfu.com/public/tdl4_article/draft_tdl4article.html
but contrary to it (which uses [IoRegisterPlugPlayNotification](https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/nf-wdm-ioregisterplugplaynotification) which is not reliable to detect when the storage stack is up)
we use instead a specific hook and the technique of driver reinitialization:
https://web.archive.org/web/20211021050515/https://driverentry.com.br/en/blog/?p=261
```
We want to be able to perform I/O-related initialization (starting a
logger thread for file log debugging, loading KDBinit file for KDBG,
etc.). A good place for this would be as early as possible, once the
I/O Manager has started the disk drivers and the boot filesystem driver.

Here is an overview of the initialization steps of the NT Kernel and
Executive:

KiSystemStartup(KeLoaderBlock)
    if (Cpu == 0) KdInitSystem(0, KeLoaderBlock);
    KiSwitchToBootStack() -> KiSystemStartupBootStack()
    -> KiInitializeKernel() -> ExpInitializeExecutive(Cpu, KeLoaderBlock)

(NOTE: Any unexpected debugger break will call KdInitSystem(0, NULL); )
KdInitSystem(0, LoaderBlock) -> KdDebuggerInitialize0(LoaderBlock);

ExpInitializeExecutive(Cpu == 0):    ExpInitializationPhase = 0;
    HalInitSystem(0, KeLoaderBlock); <-- Sets HalInitPnpDriver callback.
    ...
    PsInitSystem(LoaderBlock)
        PsCreateSystemThread(Phase1Initialization)

Phase1Initialization(Discard):       ExpInitializationPhase = 1;
    HalInitSystem(1, KeLoaderBlock);
    ...
    Early initialization of Ob, Ex, Ke.
    KdInitSystem(1, KeLoaderBlock);
    ...
    KdDebuggerInitialize1(LoaderBlock);
    ...
    IoInitSystem(LoaderBlock);
    ...

As we can see, KdDebuggerInitialize1() is the last KD initialization
routine the kernel calls, and is called *before* the I/O Manager starts.
Thus, direct Nt/ZwCreateFile ... calls done there would fail. Also,
we want to do our I/O initialization as soon as possible. There does
not seem to be any exported way to be notified about the I/O manager
initialization steps... that is, unless we somehow become a driver and
insert ourselves in the flow!

Since we are not a natural driver, we need to invoke IoCreateDriver()
to create one. However, remember that we are currently running *before*
IoInitSystem(), the I/O subsystem is not initialized yet. Due to this,
calling IoCreateDriver(), much like any other IO functions, would lead
to a crash, because it calls
ObCreateObject(..., IoDriverObjectType, ...), and IoDriverObjectType
is non-initialized yet (it's NULL).

The chosen solution here is to hook a "known" exported callback: namely,
the HalInitPnpDriver() callback (initializes the HAL Root Bus Driver).
It is set very early on by the HAL via the HalInitSystem(0, ...) call,
and is called early on by IoInitSystem() before any driver is loaded,
but after the I/O Manager has been minimally set up so that new drivers
can be created.
When our hooked HalInitPnpDriver() is called, we create our driver with
IoCreateDriver(), specifying its entrypoint KdpDriverEntry, then restore
and call the original HalInitPnpDriver() callback.

Another possible unexplored alternative, could be to insert ourselves
in the KeLoaderBlock->LoadOrderListHead boot modules list, or in the
KeLoaderBlock->BootDriverListHead boot-driver list. (Note that while
we may be able to do this, because boot-drivers are resident in memory,
much like we are, we cannot insert ourselves in the system-driver list
however, since those drivers are expected to come from PE image files.)

Once our "driver" KdpDriverEntry() entrypoint is called, we register
KdpDriverReinit() for re-initialization with the I/O Manager, in order
to provide more initialization points. KdpDriverReinit() calls the KD
providers at BootPhase >= 2, and reschedule further reinitializations
if any of those fail with either STATUS_OBJECT_NAME_NOT_FOUND or
STATUS_OBJECT_PATH_NOT_FOUND.
```

## TODO

- [x] More testing in cooperation with @binarymaster, @julenuri, @DarkFire01 and others who have also encountered the problem.

### DIFF
--- a/ntoskrnl/io/iomgr/iomgr.c
+++ b/ntoskrnl/io/iomgr/iomgr.c
@@ -588,14 +588,6 @@ IoInitSystem(IN PLOADER_PARAMETER_BLOCK LoaderBlock)
      * We can finally load other drivers from the boot volume. */
     PnPBootDriversInitialized = TRUE;
 
-#if !defined(_WINKD_) && defined(KDBG)
-    /* Read KDB Data */
-    KdbpCliInit();
-
-    /* I/O is now setup for disk access, so phase 3 */
-    KdInitSystem(3, LoaderBlock);
-#endif
-
     /* Load system start drivers */
     IopInitializeSystemDrivers();
     PnpSystemInit = TRUE;
--- a/ntoskrnl/kd/kdio.c
+++ b/ntoskrnl/kd/kdio.c
@@ -19,7 +19,6 @@
 
 #define KdpBufferSize  (1024 * 512)
 static BOOLEAN KdpLoggingEnabled = FALSE;
-static BOOLEAN KdpLoggingStarting = FALSE;
 static PCHAR KdpDebugBuffer = NULL;
 static volatile ULONG KdpCurrentPosition = 0;
 static volatile ULONG KdpFreeBytes = 0;
@@ -39,14 +38,14 @@ static ULONG KdpScreenLineBufferPos = 0, KdpScreenLineLength = 0;
 
 KDP_DEBUG_MODE KdpDebugMode;
 LIST_ENTRY KdProviders = {&KdProviders, &KdProviders};
-KD_DISPATCH_TABLE DispatchTable[KdMax];
+KD_DISPATCH_TABLE DispatchTable[KdMax] = {0};
 
 PKDP_INIT_ROUTINE InitRoutines[KdMax] =
 {
     KdpScreenInit,
     KdpSerialInit,
     KdpDebugLogInit,
-#ifdef KDBG
+#ifdef KDBG // See kdb_cli.c
     KdpKdbgInit
 #endif
 };
@@ -158,7 +157,6 @@ KdpPrintToLogFile(PCHAR String,
 {
     KIRQL OldIrql;
     ULONG beg, end, num;
-    BOOLEAN DoReinit = FALSE;
 
     if (KdpDebugBuffer == NULL) return;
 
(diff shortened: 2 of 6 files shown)


binarymaster (commented):
Good outcome on two setups - 128 MB RAM and 512 MB RAM :+1: However the debug log begins with the arbiter prints (that usually come after HAL initialization debug print), while I expect it to begin from the very start - like it was before regresso.

HBelusca:
Yes it doesn't seem to be possible, at this point, to start the capture earlier, because it depends on the posssibility of allocating the shadow memory buffer (+ creating the pumping thread).
I think one can improve that, by doing these earlier (i think it should be possible to do that at initialization step 0 or 1), and only wait for the IO initialization step (2+) to create the actual file only, and then stop doing the debug log to file only if that fails.

HBelusca:
> like it was before regresso.

I think that before, we were cheating by manually printing the banner whenever the different debug "outputs" (screen, serial, file...) were fully initialized. And so for file debug log you had the banner, that's true, but there might have been some intermediate stuff between it and the arbiters warnings that could have gone missing.

julenuri:
(../src/boot/freeldr/freeldr/ntldr/winldr.c:525) err: PeLdrLoadImage('kdfile.dll') failed

When I started from freeldr, in the common COM1 port this line appears. I expect this to be the first chance of trying to create a debug log, and that is why we haven't the beginning of the log.

Qemu testing:
The log is writing after reboots adding the info to the previous log.  :heavy_check_mark:
The log is able to be opened and later it continues the debugging in log. :heavy_check_mark:
Tab+k works, and I can continue the command. :heavy_check_mark:

RealHardware:
The log is writing after reboots adding the info to the previous log.  :heavy_check_mark:
The log is able to be opened and later it continues the debugging in log. :heavy_check_mark:
Tab+k works, and I can continue the command. :heavy_check_mark:

Out of the ticket and maybe from the PR:
This iso build artifact in the real hardware computer with the intel graphic card, in Screen mode, if I press Tab+k, now makes an assert in miarm. Also, the back space case generates an square. 
photo1669458976

binarymaster (approved):
I have retested, nothing has broken so far. :slightly_smiling_face: Ship it.

HBelusca:
> (../src/boot/freeldr/freeldr/ntldr/winldr.c:525) err: PeLdrLoadImage('kdfile.dll') failed

Not a problem, will go away once @hpoussin and I rework the debugport in kdrosdbg.

> Also, the back space case generates an square

See PR #4914 .

zefklop:
Looks good to me. Thanks for taking care of this. At some point, we might want to move KDBG out of the kernel and put all that into KDCOM, and this is a step in the right direction.

zefklop (approved):

HBelusca:
Yes, with @hpoussin we are slowly preparing this move (to KDROSDBG)

ntoskrnl/io/iomgr/iomgr.c
```diff
@@ -584,18 +584,10 @@ IoInitSystem(IN PLOADER_PARAMETER_BLOCK LoaderBlock)
         return FALSE;
     }
 
-    // the disk subsystem is initialized here and the SystemRoot is set too
-    // we can finally load other drivers from the boot volume
+    /* The disk subsystem is initialized here and the SystemRoot is set too.
```
tkreuzer, line 587: Please put formatting / annotation / spelling fixes into a separate commit to keep the actual changes as minimal as possible.

tkreuzer (dismissed):
Please separate formatting/header/annotations/typos from the actual changes. Maybe you can split those, too (e.g. one for the VOID -> NTSTATUS change). This would make reviewing much easier. Also please coordinate with @hpoussin if possible (see #4600).

ntoskrnl/io/iomgr/iomgr.c
HBelusca: Will do :+1:

ntoskrnl/kd/kdio.c
```diff
@@ -267,26 +264,68 @@ KdpDebugLogInit(PKD_DISPATCH_TABLE DispatchTable,
                                    NULL);
 
         /* Create the log file */
+        DPRINT1("Create/open '%wZ'\n", &FileName);
```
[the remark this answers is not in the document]
HBelusca: Yep, this is a temporary debugging aid.

PR #4895, opened 2022-11-23
[NTOS:EX][NTOS:MM] Minor code improvements, trivial change of behavior only, based on CORE-11958 patch

SergeGautherie:
## Purpose

Commit good parts of the abandoned patch,
plus some related improvements.

JIRA issue: [CORE-11958](https://jira.reactos.org/browse/CORE-11958)

Co-authored-by: "vshcherbyna"

## Proposed changes

Minor code improvements, trivial change of behavior only.

- Move a few RtlZeroMemory() calls around.
- Match a few variable name/types.
- A few trivial code rewrites.

## TODO

- [ ] Are the 2 'TODO?' code wanted, to match other places?

### DIFF
--- a/ntoskrnl/ex/handle.c
+++ b/ntoskrnl/ex/handle.c
@@ -683,7 +683,7 @@ NTAPI
 ExpAllocateHandleTableEntry(IN PHANDLE_TABLE HandleTable,
                             OUT PEXHANDLE NewHandle)
 {
-    ULONG OldValue, NewValue, NewValue1;
+    ULONG OldValue, NewValue, OldValue1;
     PHANDLE_TABLE_ENTRY Entry;
     EXHANDLE Handle, OldHandle;
     BOOLEAN Result;
@@ -760,7 +760,7 @@ ExpAllocateHandleTableEntry(IN PHANDLE_TABLE HandleTable,
 
         /* Now get the next value and do the compare */
         NewValue = *(volatile ULONG*)&Entry->NextFreeTableEntry;
-        NewValue1 = InterlockedCompareExchange((PLONG) &HandleTable->FirstFree,
+        OldValue1 = InterlockedCompareExchange((PLONG) &HandleTable->FirstFree,
                                                NewValue,
                                                OldValue);
 
@@ -769,19 +769,16 @@ ExpAllocateHandleTableEntry(IN PHANDLE_TABLE HandleTable,
         KeLeaveCriticalRegion();
 
         /* Check if the compare was successful */
-        if (NewValue1 == OldValue)
+        if (OldValue1 == OldValue)
         {
             /* Make sure that the new handle is in range, and break out */
             ASSERT((NewValue & FREE_HANDLE_MASK) <
                    HandleTable->NextHandleNeedingPool);
             break;
         }
-        else
-        {
-            /* The compare failed, make sure we expected it */
-            ASSERT((NewValue1 & FREE_HANDLE_MASK) !=
-                   (OldValue & FREE_HANDLE_MASK));
-        }
+
+        /* The compare failed, make sure we expected it, then try again */
+        ASSERT((OldValue1 & FREE_HANDLE_MASK) != (OldValue & FREE_HANDLE_MASK));
     }
 
     /* Increase the number of handles */
--- a/ntoskrnl/ex/handle.c
+++ b/ntoskrnl/ex/handle.c
@@ -114,9 +114,6 @@ ExpAllocateTablePagedPool(IN PEPROCESS Process OPTIONAL,
     Buffer = ExAllocatePoolWithTag(PagedPool, Size, TAG_OBJECT_TABLE);
     if (Buffer)
     {
-        /* Clear the memory */
-        RtlZeroMemory(Buffer, Size);
-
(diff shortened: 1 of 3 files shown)


ntoskrnl/ex/handle.c
```diff
@@ -728,16 +728,11 @@ ExpAllocateHandleTableEntry(IN PHANDLE_TABLE HandleTable,
...
-                {
-                    /* We're still the only thread around, so fail */
-                    NewHandle->GenericHandleOverlay = NULL;
-                    return NULL;
-                }
+            /* Check if allocation failed and nobody else went through here */
+            if (!Result && !OldValue)
+                /* We're still the only thread around, so fail */
+                NewHandle->GenericHandleOverlay = NULL;
+                return NULL;
```
learn-more, line 735: You probably introduced a bug here.

learn-more (changes requested):
'Minor code improvements, trivial change of behavior' does not at all describe the purpose of a PR.

This is the last time I am going to ask you:
- If you create a PR, please describe the PURPOSE of the PR, not the literal code or some generic statements.

JoachimHenze:
I do think the part of "moving downwards the 2 calls to RtlZeroMemory(Buffer, Size);" of this PR is fine and worth committing.

Any other change done in here, I do not understand and would request a convincing justification for. If you cannot give such convincing justification, they should be reverted.
That includes your added comments about where you raised the question "whether more calls of RtlZeroMemory would be beneficial". Either you are sure about that, and you can prove them to be missing, then do so and prove it.
Or you have no clue about ít, in that case it makes no sense to lure other people into the same thinking by adding those comments.

Same for the added casts: Does any compiler complain, or is proven to do the implicit cast wrong? If not, I see no benefit in adding them.

ntoskrnl/ex/handle.c
katahiromz: { }

binarymaster:
Also the build is rekt.

ntoskrnl/ex/handle.c
SergeGautherie: Sure. I re-added the `{` line which was unintentionaly removed.

SergeGautherie:
> added casts

gcc-i386-Debug, without casts:
```
2022-11-24T02:01:06.4836853Z ../src/ntoskrnl/mm/ARM3/virtual.c: In function 'NtAllocateVirtualMemory':
2022-11-24T02:01:06.4837490Z ../src/ntoskrnl/mm/ARM3/virtual.c:4797:32: error: passing argument 1 of 'MiInsertVadEx' from incompatible pointer type [-Werror=incompatible-pointer-types]
2022-11-24T02:01:06.4837911Z          Status = MiInsertVadEx(Vad,
2022-11-24T02:01:06.4838135Z                                 ^~~
2022-11-24T02:01:06.4838390Z In file included from ../src/ntoskrnl/mm/ARM3/virtual.c:16:
2022-11-24T02:01:06.4838939Z ../src/ntoskrnl/mm/ARM3/miarm.h:2237:20: note: expected 'PMMVAD' {aka 'struct _MMVAD *'} but argument is of type 'PMMVAD_LONG' {aka 'struct _MMVAD_LONG *'}
2022-11-24T02:01:06.4839286Z      _Inout_ PMMVAD Vad,
2022-11-24T02:01:06.4839483Z              ~~~~~~~^~~
2022-11-24T02:01:06.4839821Z ../src/ntoskrnl/mm/ARM3/virtual.c: In function 'NtFreeVirtualMemory':
2022-11-24T02:01:06.4840408Z ../src/ntoskrnl/mm/ARM3/virtual.c:5595:33: error: passing argument 1 of 'MiInsertVad' from incompatible pointer type [-Werror=incompatible-pointer-types]
2022-11-24T02:01:06.4840894Z                      MiInsertVad(NewVad, &Process->VadRoot);
2022-11-24T02:01:06.4841136Z                                  ^~~~~~
2022-11-24T02:01:06.4841381Z In file included from ../src/ntoskrnl/mm/ARM3/virtual.c:16:
2022-11-24T02:01:06.4841907Z ../src/ntoskrnl/mm/ARM3/miarm.h:2231:20: note: expected 'PMMVAD' {aka 'struct _MMVAD *'} but argument is of type 'PMMVAD_LONG' {aka 'struct _MMVAD_LONG *'}
2022-11-24T02:01:06.4842240Z      _Inout_ PMMVAD Vad,
2022-11-24T02:01:06.4842436Z              ~~~~~~~^~~
```

ThFabba:
I am against moving these `RtlZeroMemory` calls. They belong right after (or inside, with ExAllocatePoolZero) the allocation for clarity and to avoid future bugs. That is significantly more important than optimizing the performance of error cases.

JoachimHenze:
Ok Thomas, I do trust your educated judgement more than my own feeling regarding the moving of the RtlZeroMemory calls.
I am fine with your opinion superseding mine here therefore. Follow Thomas advice then.

Are you fine with the casts to silence the GCC build? Serges justification looks legit to me, but I cannot judge the correctness of those changes,

ThFabba:
The casts are necessary if we're doing the type change. Why was the type change necessary? 🤷

PR #4896, opened 2022-11-24
Kddata fix clang

HBelusca:
## Purpose

OK it would appear that Clang, even in MSVC mode, behaves as stubbornly as GCC regarding 32-bit pointer extension to 64 bits...


binarymaster:
> HBelusca closed this 10 hours ago

But the build is still rekt.

HBelusca:
Indeed :+1: But I realized that I could instead just look at the build status of branches in my fork directly, instead of polluting the PR list to see how builds go.

binarymaster:
Ah ok, makes sense.

PR #4898, opened 2022-11-24
[MMSYS] Prevent the same sound from being added multiple times

TAN-Gaming:
## Proposed changes
- Check if the list already contains the sound

Before:
ros_mmsys_before

After:
ros_mmsys_after

### DIFF
--- a/dll/cpl/mmsys/sounds.c
+++ b/dll/cpl/mmsys/sounds.c
@@ -896,6 +896,33 @@ FreeSoundFiles(HWND hwndDlg)
     }
 }
 
+static
+LRESULT
+FindSoundFileInList(HWND hwndDlg, LPCWSTR pSoundPath)
+{
+    LRESULT lCount, lIndex, lResult;
+    LPWSTR pszPath;
+    HWND hwndComboBox;
+
+    hwndComboBox = GetDlgItem(hwndDlg, IDC_SOUND_LIST);
+    lCount = ComboBox_GetCount(hwndComboBox);
+    if (lCount == CB_ERR)
+        return CB_ERR;
+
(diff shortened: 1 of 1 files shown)


dll/cpl/mmsys/sounds.c
```diff
@@ -1182,7 +1209,13 @@ SoundsDlgProc(HWND hwndDlg,
 
                     if (GetOpenFileNameW(&ofn))
                     {
-                        // FIXME search if list already contains that sound
+                        // search if list already contains that sound
+                        lResult = FindSoundFileInList(hwndDlg, filename);
+                        if (lResult != CB_ERR)
```
learn-more, line 1214:
```suggestion
                        if (lResult >= 0)
```
Seems a bit more intuitive, but fine either way.

learn-more (approved):

dll/cpl/mmsys/sounds.c
TAN-Gaming: I think `CB_ERR` is more readable.

JoachimHenze:
Please perform the following test before committing:
Make sure that the combobox is completely empty. Then try to add a first entry via the browse button.
Does that still work after this patch?

In case it wouldn't work yet, then please fix it before merging!

TAN-Gaming:
It still works after this PR applied.

[ReactOS-screen0.webm](https://user-images.githubusercontent.com/86486978/203904894-72d89326-22ab-47ae-bd2e-3ac45a3e8a69.webm)

JoachimHenze (approved):

binarymaster (approved):

PR #4899, opened 2022-11-24
[MMSYS] Minor fixes for FreeSoundFiles

TAN-Gaming:
## Purpose
- Addendum to 781c247bd3c5aeee799527db2ce06e15f794b0a4

## Proposed changes
- Skip empty (None) item
- Remove unnecessary `pSoundPath` variable

### DIFF
--- a/dll/cpl/mmsys/sounds.c
+++ b/dll/cpl/mmsys/sounds.c
@@ -875,7 +875,6 @@ VOID
 FreeSoundFiles(HWND hwndDlg)
 {
     LRESULT lCount, lIndex, lResult;
-    PWCHAR pSoundPath;
     HWND hwndComboBox;
 
     hwndComboBox = GetDlgItem(hwndDlg, IDC_SOUND_LIST);
@@ -891,8 +890,7 @@ FreeSoundFiles(HWND hwndDlg)
             continue;
         }
 
-        pSoundPath = (PWCHAR)lResult;
-        free(pSoundPath);
+        free((PVOID)lResult);
     }
 }
 
--- a/dll/cpl/mmsys/sounds.c
+++ b/dll/cpl/mmsys/sounds.c
@@ -886,7 +886,7 @@ FreeSoundFiles(HWND hwndDlg)
     for (lIndex = 0; lIndex < lCount; lIndex++)
     {
         lResult = ComboBox_GetItemData(hwndComboBox, lIndex);
-        if (lResult == CB_ERR)
+        if (lResult == CB_ERR || lResult == 0)
         {
             continue;
         }


learn-more:
Does this fix a bug?
What is the purpose of this PR?
Please describe this in your description.

TAN-Gaming:
> Does this fix a bug? What is the purpose of this PR? Please describe this in your description.

I tried debugging (using text output). `free()` is called on an empty item (None).

It doesn't make sense to free an empty item (None) so just skip it.

dll/cpl/mmsys/sounds.c
```diff
@@ -886,12 +886,12 @@ FreeSoundFiles(HWND hwndDlg)
...
         lResult = ComboBox_GetItemData(hwndComboBox, lIndex);
-        if (lResult == CB_ERR)
+        if (lResult == CB_ERR || lResult == 0)
         {
             continue;
         }
 
-        pSoundPath = (PWCHAR)lResult;
+        pSoundPath = (LPWSTR)lResult;
         free(pSoundPath);
```
HBelusca, line 895: Could have been directly: `free((PVOID)lResult)`;

TAN-Gaming: Done.

learn-more:
> If ptr is a null pointer, the function does nothing.

https://en.cppreference.com/w/c/memory/free
So not really fixing a bug, hence my question.

TAN-Gaming:
> > If ptr is a null pointer, the function does nothing.
> 
> https://en.cppreference.com/w/c/memory/free So not really fixing a bug, hence my question.

It won't fix any bug, but the purpose of the function is to free all the allocated items.
This change doesn't introduce more `if`.
The `lResult == 0` check is widely used too.

dll/cpl/mmsys/sounds.c
```diff
@@ -886,13 +885,12 @@ FreeSoundFiles(HWND hwndDlg)
...
         lResult = ComboBox_GetItemData(hwndComboBox, lIndex);
-        if (lResult == CB_ERR)
+        if (lResult == CB_ERR || lResult == 0)
         {
             continue;
         }
 
-        pSoundPath = (PWCHAR)lResult;
-        free(pSoundPath);
+        free((PVOID)lResult);
```
binarymaster, line 893: I would prefer the former code, it's more readable and self-descriptive.

TAN-Gaming:
Waiting for reviews ...

learn-more:
Sorry for the long wait.
Considering @binarymaster prefers the previous code, and I am not convinced either that this is an improvement, let's close it.

PR #4902, opened 2022-11-25
[RPCRT4] Sync msvc.S with WINE's inline assembly,

learn-more:
as suggested by Thomas

[CORE-18505](https://jira.reactos.org/browse/CORE-18505)

## Purpose

This fixes the VirtualBox 2.0.0 installer.

### DIFF
--- a/dll/win32/rpcrt4/msvc.S
+++ b/dll/win32/rpcrt4/msvc.S
@@ -13,24 +13,25 @@ EXTERN _ndr_client_call:PROC
 PUBLIC _call_stubless_func
 _call_stubless_func:
 
-    mov ecx,[esp+4]        /* This Pointer */
-    mov ecx,[ecx]          /* This->lpVtbl */
-    mov ecx,[ecx-8]        /* MIDL_STUBLESS_PROXY_INFO */
-    mov edx,[ecx+8]        /* Info->FormatStringOffset */
-    mov edx,[edx+eax*2]    /* FormatStringOffset[index] */
-    and edx, 0000FFFFh
-    add edx,[ecx+4]        /* info->ProcFormatString + offset */
-    mov eax, [edx+8]       /* arguments size */
-    and eax, 0000FFFFh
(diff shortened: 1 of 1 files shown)


JoachimHenze (approved):
Approval given based on positive test outcome. Not from understanding the ASM changes.

JoachimHenze:
Do we need something similar also for MSVC x64?

ThFabba (approved):

learn-more:
> Do we need something similar also for MSVC x64?

Probably, but I don't have a testcase nor a working gcc x64, and I don't fancy spending a lot of time deciphering that AT&T syntax.

dll/win32/rpcrt4/msvc.S
```diff
@@ -13,24 +13,25 @@ EXTERN _ndr_client_call:PROC
...
+    shr eax, 1
+    movzx eax, word ptr [edx+eax+4] /* arguments size */
     push eax
-    lea eax, [esp+8]       /* &This */
+    lea eax, [esp+8]                /* &This */
     push eax
-    push edx               /* format string */
-    push [ecx]             /* info->pstubdesc */
+    push edx                        /* format string */
+    push [ecx]                      /* info->pstubdesc */
```
tkreuzer, line 30: You might want to add a `dword ptr` here. Not sure it it's needed though.

tkreuzer (approved):

julenuri:
I have a testing GCC x64 (not of current master). It's not based on the current GCC 8 that we use, but GCC 11, that have few hacks to load (made by Simone Lombardo). 
I don't believe RPCRT4 works inside of it, but I had to retest - since there are VBox 2.0.0 XP64 / amd64 versions.

julenuri:
Fixed for me under x86 MSVC, so the ticket might be closed as merged :) Thanks you
Captura desde 2022-11-26 19-46-17
.

binarymaster (approved):

PR #4903, opened 2022-11-25
[COMDLG32] Enable saving files with different extension

zecarlos1957:
[COMDLG32] Enable saving files with different extension

like on Windows, gives precedence to the user's extension in file names if there is one.
Otherwise appends the selected extension. If not selected, append default extension.

### DIFF
--- a/dll/win32/comdlg32/filedlg.c
+++ b/dll/win32/comdlg32/filedlg.c
@@ -3099,7 +3099,7 @@ BOOL FILEDLG95_OnOpen(HWND hwnd)
         /* Attach the file extension with file name*/
         ext = PathFindExtensionW(lpstrPathAndFile);
 #ifdef __REACTOS__
-         if (! *ext)
+        if (*ext == UNICODE_NULL && fodInfos->defext)
         {
             LPWSTR filterExt = NULL, lpstrFilter = NULL, pch, pchNext;
             LPCWSTR the_ext = NULL;
--- a/dll/win32/comdlg32/filedlg.c
+++ b/dll/win32/comdlg32/filedlg.c
@@ -3099,6 +3099,7 @@ BOOL FILEDLG95_OnOpen(HWND hwnd)
         /* Attach the file extension with file name*/
         ext = PathFindExtensionW(lpstrPathAndFile);
 #ifdef __REACTOS__
+         if (! *ext)
         {
             LPWSTR filterExt = NULL, lpstrFilter = NULL, pch, pchNext;
             LPCWSTR the_ext = NULL;


JoachimHenze:
Will saving something as "hello" still save it as "hello.txt" after your change?

zecarlos1957:
@JoachimHenze 
If you want a file without extension, you should change the file types  combo box to (*.*) mask (Like Windows).
Now, with my changes notepad has the same behavior as Windows.

JoachimHenze (changes requested):
I disagree: I just tested on XPSP3: If I do select "*.*" as the mask and then save a file as "hello", the file will get saved as "hello.txt", not as "hello".
In other words: I interpret this result as: We are required to attach ".txt" whenever no extension was added explicitly in the editbox. Even if "*.*" is the current mask.

zecarlos1957:
@JoachimHenze 
Please open the command line in WinXP in the directory where you saved the hellow file, type the DIR command and check the file name. Probably the explorer has the hidden extensions. I await your confirmation.

JoachimHenze:
No this is not the cause for the result I saw with MS notepad on XPSP3.
For 2 reasons:
1.) My OS is always set up to never hide any extensions.
2.) Hiding the extensions would only hide extension, but would not add extensions.

zecarlos1957:
you right

zecarlos1957:
@JoachimHenze  your comment are addressed.
 Thanks for your review

dll/win32/comdlg32/filedlg.c
```diff
@@ -3096,9 +3096,10 @@ BOOL FILEDLG95_OnOpen(HWND hwnd)
 	else
 	  fodInfos->ofnInfos->Flags &= ~OFN_READONLY;
 
-        /* Attach the file extension with file name*/
+        /* if the user don't supply a extension, attach default extension to file name*/
```
binarymaster, line 3099: This is Wine code. If you want to modify only comment here, it's better to do inside a `__REACTOS__` guard below.

dll/win32/comdlg32/filedlg.c
```diff
@@ -3096,9 +3096,10 @@ BOOL FILEDLG95_OnOpen(HWND hwnd)
...
         ext = PathFindExtensionW(lpstrPathAndFile);
 #ifdef __REACTOS__
+         if (! *ext)
```
binarymaster, line 3102:
```suggestion
        if (*ext == UNICODE_NULL)
```

dll/win32/comdlg32/filedlg.c
zecarlos1957: yes

zecarlos1957:
@JoachimHenze 
Yours requests are addressed. Can you review please ?
https://github.com/reactos/reactos/pull/4903

dll/win32/comdlg32/filedlg.c
```diff
@@ -3099,6 +3099,7 @@ BOOL FILEDLG95_OnOpen(HWND hwnd)
         /* Attach the file extension with file name*/
         ext = PathFindExtensionW(lpstrPathAndFile);
 #ifdef __REACTOS__
+        if (*ext == UNICODE_NULL)
```
learn-more, line 3102: Should this also take into account `fodInfos->defext` like the original code does?

JoachimHenze (approved):
Approved. We are within a ReactOS block here, so we are free to experiment.
Ftr whenever you explicitly want to (re)request from me a review for any of your PRs, it is best to contact me in chat.reactos.org. Reaching me via email or github is much more likely to miss it.

dll/win32/comdlg32/filedlg.c
zecarlos1957: Yes it should. Thanks.

PR #4906, opened 2022-11-26
[SHELL32] Fix My Documents properties

TAN-Gaming:
## Purpose

JIRA issue: [CORE-17509](https://jira.reactos.org/browse/CORE-17509)

Before: Nothing happens.

After:
ROS_MyDocs

### DIFF
--- a/dll/win32/shell32/folders/CRegFolder.cpp
+++ b/dll/win32/shell32/folders/CRegFolder.cpp
@@ -82,6 +82,11 @@ HRESULT CALLBACK RegFolderContextMenuCallback(IShellFolder *psf,
         if (!SH_ShowRecycleBinProperties(L'C'))
             hr = E_FAIL;
     }
+    else
+    {
+        /* Tell the caller to run the default action */
+        hr = S_FALSE;
+    }
 
     SHFree(pidlFolder);
     _ILFreeaPidl(apidl, cidl);


binarymaster:
If I remember correctly, it should also have an additional property sheet for the My Documents folder location.

But it's improvement already. :slightly_smiling_face:

HBelusca (approved):

learn-more:
I am pretty sure that this is a hack.

dll/win32/shell32/folders/CRegFolder.cpp
```diff
@@ -82,6 +82,17 @@ HRESULT CALLBACK RegFolderContextMenuCallback(IShellFolder *psf,
...
+    {
+        WCHAR szPath[MAX_PATH];
+
+        /* FIXME: Windows has the special "Target" page for My Documents properties dialog */
+        hr = SHGetFolderPathW(NULL, CSIDL_MYDOCUMENTS, NULL, SHGFP_TYPE_CURRENT, szPath);
+        if (SUCCEEDED(hr))
+        {
+            hr = (SH_ShowPropertiesDialog(szPath, pdtobj) ? S_OK : E_FAIL);
+        }
+    }
```
learn-more, line 95: I'd have rather that you turn this into and `else` case that sets `hr = S_FALSE`,
which will tell the calling code to run the default action.

TAN-Gaming: The result will be:
Capture

I've added else case to tell the caller to run the default action.

learn-more: But you still left the specialization for the 'My Documents' case.

TAN-Gaming: At least `else if (_ILIsMyDocuments(apidl[0]))` is correct because My Documents has a special dialog.
The dialog can be implemented later.

learn-more: No, special dialogs don't have to be hardcoded like that, they are most likely from a shell extension. (With some exceptions)

For example, the 'compatibility' dialog is not hardcoded either.

TAN-Gaming: OK.

learn-more (approved):

binarymaster (approved):

PR #4914, opened 2022-11-27
KD/KDBG: Fix/improve KdbpReadCommand() and KdReceivePacket() for DBGKD_DEBUG_IO

HBelusca:
## Purpose

@julenuri noticed that debugger input is a bit "raw" when KD handles input for DbgPrompt() calls, and e.g. shows squares if backspaces are used, etc.:
image

Take also the opportunity to add other fixes.

## Proposed changes

- Validate DEBUG_IO API call.

- Take the LengthOfStringRead into account; use KdbpReadCommand() to read
  the input, so that correct line edition is available (backspace, etc.)

For KdbpReadCommand():
- Don't read anything and return immediately if the buffer size is zero.

- Allow the controlling keys (up/down arrows, backspace) even if the
  buffer is full! (especially backspace if too much has been written).

- Return the number of characters stored, not counting the NULL terminator.

### DIFF
--- a/ntoskrnl/kd/kdio.c
+++ b/ntoskrnl/kd/kdio.c
@@ -37,7 +37,7 @@ CPPORT SerialPortInfo   = {0, DEFAULT_DEBUG_BAUD_RATE, 0};
 static CHAR KdpScreenLineBuffer[KdpScreenLineLengthDefault + 1] = "";
 static ULONG KdpScreenLineBufferPos = 0, KdpScreenLineLength = 0;
 
-const ULONG KdpDmesgBufferSize = 128 * 1024; // 512*1024; // 5*1024*1024;
+const ULONG KdpDmesgBufferSize = 128 * 1024; // 512*1024;
 PCHAR KdpDmesgBuffer = NULL;
 volatile ULONG KdpDmesgCurrentPosition = 0;
 volatile ULONG KdpDmesgFreeBytes = 0;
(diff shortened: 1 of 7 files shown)


binarymaster:
From commit message:
> and return immediately of the buffer size is zero.

of -> if

ntoskrnl/kd/kdio.c
```diff
@@ -723,100 +741,51 @@ KdReceivePacket(
...
 
+    DebugIo = (PDBGKD_DEBUG_IO)MessageHeader->Buffer;
+
+#if 0
+    /* Validate API call */
+    if (MessageHeader->MaximumLength != sizeof(DBGKD_DEBUG_IO))
+        return KdPacketNeedsResend;
+    if (DebugIo->ApiNumber != DbgKdGetStringApi)
+        return KdPacketNeedsResend;
+#endif
```
binarymaster, line 752: Any reason why this is disabled?

ntoskrnl/ps/kill.c
```diff
@@ -42,23 +42,20 @@ PspCatchCriticalBreak(IN PCHAR Message,
...
+            /* A debugger is active, prompt for action */
+            DbgPrompt("Break, or Ignore (bi)? ", Action, sizeof(Action));
             switch (Action[0])
             {
                 /* Break */
                 case 'B': case 'b':
-
-                    /* Do a breakpoint */
                     DbgBreakPoint();
+                    /* Fall back */
```
binarymaster, line 52: You probably meant "Fall through".

sdk/lib/rtl/assert.c
```diff
@@ -43,52 +43,38 @@ RtlAssert(IN PVOID FailedAssertion,
...
-
-                /* Do a breakpoint, then prompt again */
+            case 'O': case 'o':
                 DbgPrint("Execute '.cxr %p' to dump context\n", &Context);
+                /* Do a breakpoint, then prompt again or return */
                 DbgBreakPoint();
-                break;
+                if ((Action[0] == 'B') || (Action[0] == 'b'))
+                    break;
+                /* else ('O','o'): fall back */
```
binarymaster, line 60: And here.

binarymaster (approved):

PR #4915, opened 2022-11-27
[RTL] Correctly initialize segments larger than MAXUSHORT. CORE-18196

ThFabba:
GuardEntry->PreviousSize would overflow in this case. RtlpInsertFreeBlock already handles this correctly by creating multiple free entries. So we just let it return the last entry and use that for the PreviousSize.

JIRA issue: [CORE-18196](https://jira.reactos.org/browse/CORE-18196)

This is not a very elegant fix and possibly incomplete. Would love your opinion on whether this is a good approach @zefklop.

### DIFF
--- a/sdk/lib/rtl/heap.c
+++ b/sdk/lib/rtl/heap.c
@@ -353,7 +353,7 @@ RtlpInsertFreeBlockHelper(PHEAP Heap,
     Heap->FreeHints[HintIndex] = &FreeEntry->FreeList;
 }
 
-VOID NTAPI
+PHEAP_FREE_ENTRY NTAPI
 RtlpInsertFreeBlock(PHEAP Heap,
                     PHEAP_FREE_ENTRY FreeEntry,
                     SIZE_T BlockSize)
@@ -361,6 +361,7 @@ RtlpInsertFreeBlock(PHEAP Heap,
     USHORT Size, PreviousSize;
     UCHAR SegmentOffset, Flags;
     PHEAP_SEGMENT Segment;
+    PHEAP_FREE_ENTRY LastEntry;
 
     DPRINT("RtlpInsertFreeBlock(%p %p %x)\n", Heap, FreeEntry, BlockSize);
 
@@ -368,6 +369,7 @@ RtlpInsertFreeBlock(PHEAP Heap,
     Heap->TotalFreeSize += BlockSize;
 
     /* Remember certain values */
+    LastEntry = FreeEntry;
     Flags = FreeEntry->Flags;
     PreviousSize = FreeEntry->PreviousSize;
     SegmentOffset = FreeEntry->SegmentOffset;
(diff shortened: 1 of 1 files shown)


sdk/lib/rtl/heap.c
```diff
@@ -1100,11 +1124,10 @@ RtlpInitializeHeapSegment(IN OUT PHEAP Heap,
             FreeEntry->SegmentOffset = SegmentIndex;
             FreeEntry->Flags = HEAP_ENTRY_LAST_ENTRY;
             FreeEntry->Size = (SegmentCommit >> HEAP_ENTRY_SHIFT) - Segment->Entry.Size;
-        }
 
-        /* Register the Free Heap Entry */
-        if (FreeEntry)
+            /* Register the Free Heap Entry */
             RtlpInsertFreeBlock(Heap, (PHEAP_FREE_ENTRY)FreeEntry, FreeEntry->Size);
```
HBelusca, line 1129: Ignoring the new return value here is OK?

sdk/lib/rtl/heap.c
```diff
@@ -1087,8 +1100,19 @@ RtlpInitializeHeapSegment(IN OUT PHEAP Heap,
...
+                    FreeEntry->Size = PreviousSize;
                     FreeEntry->Flags = 0;
+
+                    /* Register the Free Heap Entry */
+                    FreeEntry = (PHEAP_ENTRY)RtlpInsertFreeBlock(Heap, (PHEAP_FREE_ENTRY)FreeEntry, PreviousSize);
+
+                    /* Fill guard entry */
+                    ASSERT(FreeEntry->Size != 0);
+                    ASSERT(FreeEntry != GuardEntry);
+                    ASSERT(FreeEntry->Size != 1); /* FIXME: we need to double the guard entry */
```
zefklop, line 1112: I don't get the FIXME comment. The guard entry needs to be doubled in case the free entry before it is too small to be suitable for an alloc, which is size == 1. The assert is exactly the contrary of that.

zefklop:
This looks fine to me. I don't really find that the guard entry approach is very elegant either, but I couldn't come up with anything better anyway.

sdk/lib/rtl/heap.c
ThFabba: The assert is written such that it fires if size == 1, because that's the unimplemented code path.
If the general approach is acceptable, this would turn into `if (FreeEntry->Size == 1) { /* double the guard entry */ }`

zefklop: In that case, please use UNIMPLEMENTED_DBGBREAK
ReactOS code is crippled with asserts where we're not sure if those are asserts or unimplemented code paths.
But in that case, I do think this is a valid assert, see in RtlpInsertFreeBlock:
```
            /* Special compensation if it goes above limit just by 1 */
            if (BlockSize == (HEAP_MAX_BLOCK_SIZE + 1))
                Size -= 16;
```

sdk/lib/rtl/heap.c
ThFabba: Yes, most places don't need to do anything with it (which is why it was void before).

JoachimHenze:
A different approach, that is shorter but also effective in fixing CORE-18196 has been proposed by Doug in https://github.com/reactos/reactos/pull/5750
I let you decide which PR is more accurate.

Doug-Lyons:
> A different approach, that is shorter but also effective in fixing CORE-18196 has been proposed by Doug in #5750 I let you decide which PR is more accurate. Maybe we can use some tickets that are a bit more tricky to decide which of both PRs is better? E.g. https://jira.reactos.org/browse/CORE-18014 started with the same guilty rev of Jerome as https://jira.reactos.org/browse/CORE-18196 . Maybe one of both PRs behaves better in that scenario than the other?

I tested for the [CORE-18014](https://jira.reactos.org/browse/CORE-18014) leaking memory and both patches had the same results that neither affected the leak. It still remains with either patch.

SergeGautherie:
_needs rebase_ after #5750.

learn-more:

PR #4918, opened 2022-11-29
[NTOS:MM/PS] Remove code duplication between LookupEntryPoint/MiLocateExportName/MiFindExportedRoutineByName.

HBelusca:
## Purpose

Remove code duplication between LookupEntryPoint/MiLocateExportName/MiFindExportedRoutineByName.

As it turns out, those three functions were duplicating the same code
between each other. Reimplement these in terms of a common helper,
RtlFindExportedRoutineByName().
Indeed: MiFindExportedRoutineByName() was just MiLocateExportName()
but taking a PANSI_STRING instead of a NULL-terminated string.

A similar state of affairs also existed in Windows <= 2003, and the
MS guys also noticed it. Both routines have been then merged and renamed
to MiFindExportedRoutineByName() on Windows 8 (taking a PCSTR instead),
and finally renamed and exported as RtlFindExportedRoutineByName()
on Windows 10. (See e.g. Geoff Chappell.)

## Proposed changes

I introduced the common helper RtlFindExportedRoutineByName() (+ a private RtlpFindExportedRoutineByName() that provides more granularity on the returned error code). The private helper is used for LookupEntryPoint(), in order to obtain a similar behaviour as the original code (Note: on Windows 10+ this Lookup... helper has been completely removed and RtlFindExportedRoutineByName() is used directly in its place.)

The common helper RtlFindExportedRoutineByName() is what is exported in Windows 10, and what replaces both MiLocateExportName() and MiFindExportedRoutineByName() there. For the current ReactOS purposes it is kept "private" however. All calls to MiLocateExportName and MiFindExportedRoutineByName are replaced by the helper - however in order to reduce the diffs I decided instead to #define those names as an alias to the Rtl function.

Finally I took the opportunity to simplify one block of code in MiSnapThunk() by directly calling the NameToOrdinal() private helper. That latter was originally implemented in psmgr.c but in a recursive way; I replaced that implementation by the better non-recursive one found in the other functions.


ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -213,43 +213,31 @@ MiLoadImageSection(_Inout_ PSECTION *SectionPtr,
...
-    High = ExportDirectory->NumberOfNames - 1;
+    High = NumberOfNames - 1;
     while (High >= Low)
     {
         /* Get new middle value */
         Mid = (Low + High) >> 1;
 
         /* Compare name */
-        Ret = strcmp(ExportName, (PCHAR)DllBase + NameTable[Mid]);
+        Ret = strcmp(ExportName, (PCHAR)((ULONG_PTR)DllBase + NameTable[Mid]));
```
learn-more, line 240: Why did you change this?

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -268,28 +256,168 @@ MiLocateExportName(IN PVOID DllBase,
...
+                            ExportDirectory->NumberOfNames,
+                            NameTable,
+                            OrdinalTable);
+
+    /* Check if we couldn't find it */
+    if (Ordinal == -1)
+        return STATUS_ORDINAL_NOT_FOUND;
+
+    /* Validate the ordinal */
+    if (Ordinal >= ExportDirectory->NumberOfFunctions)
```
learn-more, line 344: Do you need to take into account `IMAGE_EXPORT_DIRECTORY.Base` here?

image

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -268,28 +256,168 @@ MiLocateExportName(IN PVOID DllBase,
...
+    if (CheckIfForwarder)
     {
-        /* It is, fail */
-        return NULL;
+        /* Check if the function is actually a forwarder */
+        if (((ULONG_PTR)FunctionAddress > (ULONG_PTR)ExportDirectory) &&
+            ((ULONG_PTR)FunctionAddress < ((ULONG_PTR)ExportDirectory + ExportSize)))
+        {
+            ASSERT((FunctionAddress <= (PVOID)ExportDirectory) ||
+                   (FunctionAddress >= (PVOID)((ULONG_PTR)ExportDirectory + ExportSize)));
```
learn-more, line 362: This assert seems weird?

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -268,28 +256,168 @@ MiLocateExportName(IN PVOID DllBase,
...
+    PVOID Function;
+    NTSTATUS Status;
+
+    PAGED_CODE();
+
+    /* Call the internal function */
+    Status = RtlpFindExportedRoutineByName(DllBase,
+                                           ExportName,
+                                           &Function,
+                                           TRUE);
```
learn-more, line 412: Why should this fail on a forwarder?

ntoskrnl/ps/psmgr.c
```diff
@@ -62,106 +62,31 @@ BOOLEAN PspDoingGiveBacks;
...
-        return STATUS_PROCEDURE_NOT_FOUND;
-    }
+    /* Call the internal function */
+    NTSTATUS Status = RtlpFindExportedRoutineByName(DllBase,
+                                                    Name,
+                                                    EntryPoint,
+                                                    FALSE);
+    /* Adjust the returned status if we failed */
+    if (!NT_SUCCESS(Status) && (Status != STATUS_INVALID_PARAMETER))
+        Status = STATUS_PROCEDURE_NOT_FOUND;
```
learn-more, line 80: Is this something that should be logged?

ntoskrnl/mm/ARM3/sysldr.c
HBelusca: No; this PR code is functionally equivalent to the old one we had (and what appears to be even in Win10).

ntoskrnl/mm/ARM3/sysldr.c
HBelusca: The assert is the negation of the if-test for whether function is a forwarder. (I didn't use ASSERT(FALSE) in order to have a more descriptive error in case it's hit.) The question is then, why the forwarder-check test is what it is (namely, function address being strictly greater than export directory address, and strictly less than export directory + its size).
This appears to be a recurring way of checking, see:

ntoskrnl/ps/psmgr.c
HBelusca: I don't think so. Original code didn't log here in case the entry point was not found. (Not sure it is useful as well, since this function in psmgr, and the one that calls it, are used so far only when loading exports from ntdll.dll.)

ntoskrnl/mm/ARM3/sysldr.c
learn-more: No clue why `>` is used instead of `>=`

But the range check is because this is how a 'forwarder' is defined (and in practice, is checked + implemented)
See https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#export-address-table
> The pointer to a null-terminated ASCII string in the export section. This string must be within the range that is given by the export table data directory entry. See [Optional Header Data Directories (Image Only)](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#optional-header-data-directories-image-only). This string gives the DLL name and the name of the export (for example, "MYDLL.expfunc") or the DLL name and the ordinal number of the export (for example, "MYDLL.#27").

ntoskrnl/mm/ARM3/sysldr.c
HBelusca: The original functions MiFindExportedRoutineByName and MiLocateExportName failed on that case (second one really did, first one kind of asserted, meaning that in release builds this would get unnoticed).

Starting Win8 and Win10 where that exported RtlFindExportedRoutineByName is used in place of those old internal functions, it fails in this condition as well.
I could write a Win10-specific kernel-mode test to show this.
A test that could work on current ROS and Win2k3 would be with MmGetSystemRoutineAddress() (that internally uses this ***FindExportedRoutineByName), to see whether it supports forwarders or not.
But that latter one is limited to looking only at ntoskrnl and hal modules. Also, if an ntos function forwards to a like-named one in hal, the way it is written is that it will succeed, since: a) it will search it in ntos first (and potentially fail if it was a forwarder), then b) search it in hal.

Why this has been done so? One possible reason is that the original internal functions were potentially designed to locate those system routines that must not be forwarders ("DllInitialize", "DllUnload", and all the entry points listed in https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/ps/psmgr.c;hb=4d0d22aee22a56b89990c9e18b2a4cb3cf8d8af5#l172).

ntoskrnl/ps/psmgr.c
```diff
@@ -62,106 +62,31 @@ BOOLEAN PspDoingGiveBacks;
...
-                            OrdinalTable);
-
-    /* Make sure the ordinal is valid */
-    if (Ordinal >= ExportDirectory->NumberOfFunctions)
-    {
-        /* It's not, fail */
-        return STATUS_PROCEDURE_NOT_FOUND;
-    }
+    /* Call the internal function */
+    NTSTATUS Status = RtlpFindExportedRoutineByName(DllBase,
```
tkreuzer, line 74: Variable declaration separately? Then the call firs in one line.

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -268,28 +256,168 @@ MiLocateExportName(IN PVOID DllBase,
...
-        ((ULONG_PTR)Function < ((ULONG_PTR)ExportDirectory + ExportSize)))
+    if (CheckIfForwarder)
     {
-        /* It is, fail */
-        return NULL;
+        /* Check if the function is actually a forwarder */
+        if (((ULONG_PTR)FunctionAddress > (ULONG_PTR)ExportDirectory) &&
+            ((ULONG_PTR)FunctionAddress < ((ULONG_PTR)ExportDirectory + ExportSize)))
+        {
+            ASSERT((FunctionAddress <= (PVOID)ExportDirectory) ||
```
tkreuzer, line 361: This should probably be `<`.

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -268,28 +256,168 @@ MiLocateExportName(IN PVOID DllBase,
...
+ * The function does not handle forwarders, and will fail with
+ * STATUS_PROCEDURE_NOT_FOUND if the export is as such.
+ *
+ * @note
+ * See RtlFindExportedRoutineByName() below for more remarks.
+ * Used by psmgr.c LookupEntryPoint() as well.
+ **/
+NTSTATUS
+NTAPI
+RtlpFindExportedRoutineByName(
```
tkreuzer, line 302: Move this to RTL?

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -268,28 +256,168 @@ MiLocateExportName(IN PVOID DllBase,
...
+ *
+ * @note
+ * This routine was originally named MiLocateExportName(), with a separate
+ * duplicated MiFindExportedRoutineByName() one (taking a PANSI_STRING)
+ * on Windows <= 2003. Both routines have been then merged and renamed
+ * to MiFindExportedRoutineByName() on Windows 8 (taking a PCSTR instead),
+ * and finally renamed and exported as RtlFindExportedRoutineByName() on
+ * Windows 10.
+ **/
+#define MiLocateExportName              RtlFindExportedRoutineByName
```
tkreuzer, line 395: Do we really need this? Why not simply rename those 3 instances?

ntoskrnl/mm/ARM3/sysldr.c
HBelusca: The assert is the negation of the if-test (in order to avoid having an `ASSERT(FALSE);` instead). See discussion with Mark above.

ntoskrnl/mm/ARM3/sysldr.c
HBelusca: We could just rename the 3 instances. (I just wanted to keep their original names for documentation purposes, and to not add extra changes to the diff.)

ntoskrnl/include/internal/mm.h
```diff
@@ -1653,15 +1653,30 @@ MmCheckSystemImage(
...
     IN PLDR_DATA_TABLE_ENTRY LdrEntry);
 
+/* ReactOS-only, used by psmgr.c LookupEntryPoint() as well */
+NTSTATUS
+NTAPI
+RtlpFindExportedRoutineByName(
+    _In_ PVOID DllBase,
+    _In_ PCSTR ExportName,
+    _Out_ PVOID* Function,
+    _In_opt_ BOOLEAN CheckIfForwarder);
```
learn-more, line 1671: I don't think this can be an `opt` argument, considering it is not a pointer.

ntoskrnl/include/internal/mm.h
```diff
@@ -1653,15 +1653,30 @@ MmCheckSystemImage(
...
 NTAPI
 MmFreeDriverInitialization(
     IN PLDR_DATA_TABLE_ENTRY LdrEntry);
 
+/* ReactOS-only, used by psmgr.c LookupEntryPoint() as well */
+NTSTATUS
+NTAPI
+RtlpFindExportedRoutineByName(
+    _In_ PVOID DllBase,
+    _In_ PCSTR ExportName,
```
learn-more, line 1669: `_In_z` probably?

ntoskrnl/include/internal/mm.h
learn-more: And maybe something like `AllowForwarder` or `DisallowForwarder` is a better name,
`CheckIfForwarder` does not describe what is happening when it is a forwarder at all.

HBelusca: This has been re-architectured.

ntoskrnl/mm/ARM3/sysldr.c
HBelusca: Will do, maybe, in a separate PR.

PR #4922, opened 2022-12-01
[NTOS:MM] Nt*VirtualMemory(): Match 2 variable actual types

SergeGautherie:
## Purpose

Self-documentation, consistency within function and with others using the same pattern, allows to use `sizeof(*varname)` syntax...

JIRA issue: split off from #4895.

## Proposed changes

- Use `PMMVAD_LONG` instead of `PMMVAD`.

### DIFF
--- a/ntoskrnl/mm/ARM3/virtual.c
+++ b/ntoskrnl/mm/ARM3/virtual.c
@@ -4492,7 +4492,8 @@ NtAllocateVirtualMemory(IN HANDLE ProcessHandle,
 {
     PEPROCESS Process;
     PMEMORY_AREA MemoryArea;
-    PMMVAD Vad = NULL, FoundVad;
+    PMMVAD_LONG Vad = NULL;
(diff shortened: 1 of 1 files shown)


tkreuzer (approved):

ntoskrnl/mm/ARM3/virtual.c
```diff
@@ -4793,7 +4794,7 @@ NtAllocateVirtualMemory(IN HANDLE ProcessHandle,
         //
         // Insert the VAD
         //
-        Status = MiInsertVadEx(Vad,
+        Status = MiInsertVadEx((PMMVAD)Vad,
```
HBelusca, line 4797: But then, this requires to make a cast? Not sure whether this is better...

ThFabba (approved):

ThFabba:
Please only suggest changes that are objective improvements. This is a waste of time.

PR #4924, opened 2022-12-03
[MSPAINT] Don't allow to set image as wallpaper if it has the wrong format

Cleverking2003:
## Purpose

Our Paint allows user to try to set a .ico file as a wallpaper, which isn't possible. Different Windows versions have different behaviour, so it was decided that the simplest fix would be to just grey out "Set as wallpaper" buttons as in 2K3 (See the Jira ticket).

JIRA issue: [CORE-18661](https://jira.reactos.org/browse/CORE-18661)

## Proposed changes

Check the file extension and deactivate the buttons if needed.

image

### DIFF
--- a/base/applications/mspaint/winproc.cpp
+++ b/base/applications/mspaint/winproc.cpp
@@ -280,14 +280,18 @@ LRESULT CMainWindow::OnInitMenuPopup(UINT nMsg, WPARAM wParam, LPARAM lParam, BO
     BOOL trueSelection =
         (::IsWindowVisible(selectionWindow) &&
          ((toolsModel.GetActiveTool() == TOOL_FREESEL) || (toolsModel.GetActiveTool() == TOOL_RECTSEL)));
+    BOOL isBMP;
     switch (lParam)
     {
         case 0: /* File menu */
             if ((HMENU)wParam != GetSubMenu(menu, 0))
                 break;
-            EnableMenuItem(menu, IDM_FILEASWALLPAPERPLANE,     ENABLED_IF(isAFile));
-            EnableMenuItem(menu, IDM_FILEASWALLPAPERCENTERED,  ENABLED_IF(isAFile));
-            EnableMenuItem(menu, IDM_FILEASWALLPAPERSTRETCHED, ENABLED_IF(isAFile));
+
+            isBMP = _wcsicmp(PathFindExtensionW(filepathname), L".bmp") == 0;
+            EnableMenuItem(menu, IDM_FILEASWALLPAPERPLANE,     ENABLED_IF(isAFile && isBMP));
+            EnableMenuItem(menu, IDM_FILEASWALLPAPERCENTERED,  ENABLED_IF(isAFile && isBMP));
+            EnableMenuItem(menu, IDM_FILEASWALLPAPERSTRETCHED, ENABLED_IF(isAFile && isBMP));
+
             RemoveMenu(menu, IDM_FILE1, MF_BYCOMMAND);
             RemoveMenu(menu, IDM_FILE2, MF_BYCOMMAND);
             RemoveMenu(menu, IDM_FILE3, MF_BYCOMMAND);


katahiromz (approved):

KRosUser2:
Are you sure that .ico is the only file format that can be opened by Paint but can NOT be used as wallpaper ?

base/applications/mspaint/winproc.cpp
```diff
@@ -280,14 +280,18 @@ LRESULT CMainWindow::OnInitMenuPopup(UINT nMsg, WPARAM wParam, LPARAM lParam, BO
...
     switch (lParam)
     {
         case 0: /* File menu */
             if ((HMENU)wParam != GetSubMenu(menu, 0))
                 break;
-            EnableMenuItem(menu, IDM_FILEASWALLPAPERPLANE,     ENABLED_IF(isAFile));
-            EnableMenuItem(menu, IDM_FILEASWALLPAPERCENTERED,  ENABLED_IF(isAFile));
-            EnableMenuItem(menu, IDM_FILEASWALLPAPERSTRETCHED, ENABLED_IF(isAFile));
+
+            notIco = wcscmp(PathFindExtensionW(filepathname), L".ico") != 0;
```
learn-more, line 290: So you allow `.ICO` and `.Ico` etc to be set?
I would compare case insensitive here ;)

JoachimHenze:
@Krosuser. No we are not sure about that. But the code could be expanded later for other similar formats. That doesn't invalidate the PR.
Feel free to provide us a list of formats that are not available for desktop backgrounding.

Cleverking2003:
I tested right now and it seems that only .bmp can be set as a wallpaper on ReactOS

KRosUser2:
> I tested right now and it seems that only .bmp can be set as a wallpaper on ReactOS

Then should be =bmp rather than !=ico

base/applications/mspaint/winproc.cpp
Cleverking2003: Fixed

Cleverking2003:
Fixed the PR, now we don't allow anything other than .bmp

base/applications/mspaint/winproc.cpp
```diff
@@ -280,14 +280,18 @@ LRESULT CMainWindow::OnInitMenuPopup(UINT nMsg, WPARAM wParam, LPARAM lParam, BO
...
     switch (lParam)
     {
         case 0: /* File menu */
             if ((HMENU)wParam != GetSubMenu(menu, 0))
                 break;
-            EnableMenuItem(menu, IDM_FILEASWALLPAPERPLANE,     ENABLED_IF(isAFile));
-            EnableMenuItem(menu, IDM_FILEASWALLPAPERCENTERED,  ENABLED_IF(isAFile));
-            EnableMenuItem(menu, IDM_FILEASWALLPAPERSTRETCHED, ENABLED_IF(isAFile));
+
+            isBMP = _wcsicmp(PathFindExtensionW(filepathname), L".bmp") == 0;
```
binarymaster, line 290: I think there should be a way to try setting the wallpaper, and check error code in case it fails.

JoachimHenze (approved):

base/applications/mspaint/winproc.cpp
JoachimHenze: I think the current way of making paint check that itself is better (more defensive) compared to passing garbage data to other components and relying on their correct behavior for all of them.
It is more likely that we will have to adapt the code again in the future to support other formats (like jpg), yes. But unless you could provide demo code for the check you have in mind, I do like the current approach.

I still recommend to double-check the beahvior in our unpatched build for .jpg and .png before merging this PR.

JoachimHenze: For the record: On my XPSP3 the desk.cpl applet allows to set as wallpaper *.bmp, *.gif, *.jpg, *.jpeg, *.dib, *.png
And it calls that set of formats "Alle Bilddateien".
We do know that different Windows versions do support a different set of formats. Most likely they get more with each version of Windows.

Furthermore my XPSP3 allows *.htm and *.html but since we decided to not support ActiveDesktop that part doesn't apply to us. And I think this is a good decision. Active Desktop sucks.

Cleverking2003: I investigated that and our desk.cpl is actually able to set more formats as wallpapers (e.g. gif, png, jpg) than our Paint. So it seems that Paint does something wrong when setting images as wallpapers, so it only works with bmp.
As for the list of formats, our desk.cpl calls gdi to find supported formats, but this approach also isn't correct (for example, it also counts .ico, but you can't set an icon as a wallpaper in desk.cpl).

mr-sihc:
does this just gray the option out?
if so, that really would confuse the user
try making a tooltip if you hover over the disabled option or automatically convert to .bmp

learn-more:
> does this just gray the option out? if so, that really would confuse the user try making a tooltip if you hover over the disabled option or automatically convert to .bmp

Why would this confuse the user?
That is a very common way of indicating that something is not available.

Cleverking2003:
> Wrong suggestion by @KRosUser2 . If e.g. a .jpg or .png .... are opened in mspaint, they can be set as wallpapers. So the comparison definitively must _NOT_ be "== bmp".

But the thing is that our Paint allows _only_ bmp to be set as a wallpaper, all other formats give an error.

HBelusca:
> > Wrong suggestion by @KRosUser2 . If e.g. a .jpg or .png .... are opened in mspaint, they can be set as wallpapers. So the comparison definitively must _NOT_ be "== bmp".
> 
> But the thing is that our Paint allows _only_ bmp to be set as a wallpaper, all other formats give an error.

Ah yes indeed you are correct (and then KRosUser too), the way our MSPaint RegistrySettings::SetWallpaper is coded indeed only allows bmps as wallpapers.

HBelusca (approved):

mr-sihc:
> > does this just gray the option out? if so, that really would confuse the user try making a tooltip if you hover over the disabled option or automatically convert to .bmp
> 
> Why would this confuse the user? That is a very common way of indicating that something is not available.

but why is it not availabe?
the user would notice sometimes, (the rare times you still use .bmp images) that it would work, and i dont think the user would know why it is working sometimes and sometimes not (i dont think a user would realize it is the image format, i wouldnt think of that when i cant set wallpapers as a general option is not available)

JoachimHenze:
"Greying out" unsupported formats does not require us to add new strings that we then would have to translate into 38 different languages, which requires a big load of follow-up-commits and also bloats the binary. And in the JIRA ticket someone mentioned that greying out unsupported formats would also be the 2k3sp2 solution to the problem. So i think this is good enough.
But improving mspaint to support also other formats, that is definitely something we should fix in the future.

KRosUser2:
The simpler the better
Lgtm, ready to go.

KRosUser2:
3 approved. Ready for merge ?

PR #4932, opened 2022-12-06
[SHLWAPI] No need extension to execute '.bat' files CORE-17612

zecarlos1957:
When no extension is provided, add default extension with correct precedence
using PathFileExistsDefExtW.

Use PathFindOnPathExW when searching in current directory and PATH environment variable
Fix Width String terminator to UNICODE_NULL
Fix all failed tests on PathFindPathExW KVM test

JIRA issue: [CORE-17612](https://jira.reactos.org/browse/CORE-17612)

### DIFF
--- a/dll/win32/shell32/shlexec.cpp
+++ b/dll/win32/shell32/shlexec.cpp
@@ -2531,6 +2531,8 @@ HRESULT WINAPI ShellExecCmdLine(
     }
     else
     {
+        PCWSTR apPathList[2];
+
         pchParams = SplitParams(lpCommand, szFile, _countof(szFile));
         if (szFile[0] != UNICODE_NULL && szFile[1] == L':' &&
             szFile[2] == UNICODE_NULL)
@@ -2551,30 +2553,10 @@ HRESULT WINAPI ShellExecCmdLine(
         {
             StringCchCopyW(szFile, _countof(szFile), szFile2);
         }
-        else if (SearchPathW(NULL, szFile, NULL, _countof(szFile2), szFile2, NULL) ||
-                 SearchPathW(NULL, szFile, L".exe", _countof(szFile2), szFile2, NULL) ||
-                 SearchPathW(NULL, szFile, L".com", _countof(szFile2), szFile2, NULL) ||
-                 SearchPathW(pwszStartDir, szFile, NULL, _countof(szFile2), szFile2, NULL) ||
-                 SearchPathW(pwszStartDir, szFile, L".exe", _countof(szFile2), szFile2, NULL) ||
-                 SearchPathW(pwszStartDir, szFile, L".com", _countof(szFile2), szFile2, NULL))
-        {
-            StringCchCopyW(szFile, _countof(szFile), szFile2);
-        }
-        else if (SearchPathW(NULL, lpCommand, NULL, _countof(szFile2), szFile2, NULL) ||
-                 SearchPathW(NULL, lpCommand, L".exe", _countof(szFile2), szFile2, NULL) ||
-                 SearchPathW(NULL, lpCommand, L".com", _countof(szFile2), szFile2, NULL) ||
-                 SearchPathW(pwszStartDir, lpCommand, NULL, _countof(szFile2), szFile2, NULL) ||
-                 SearchPathW(pwszStartDir, lpCommand, L".exe", _countof(szFile2), szFile2, NULL) ||
-                 SearchPathW(pwszStartDir, lpCommand, L".com", _countof(szFile2), szFile2, NULL))
-        {
-            StringCchCopyW(szFile, _countof(szFile), szFile2);
-            pchParams = NULL;
-        }
 
-        if (pwszStartDir)
-        {
-            SetCurrentDirectoryW(szCurDir);
-        }
+        apPathList[0] = pwszStartDir;
+        apPathList[1] = NULL;
+        PathFindOnPathExW(szFile, apPathList, WHICH_DEFAULT);
 
         if (!(dwSeclFlags & SECL_ALLOW_NONEXE))
         {
--- a/dll/win32/shlwapi/path.c
+++ b/dll/win32/shlwapi/path.c
@@ -1151,6 +1151,7 @@ BOOL WINAPI PathFileExistsDefExtW(LPWSTR lpszPath,DWORD dwWhich)
 #ifdef __REACTOS__
         if (dwWhich & 0x1)
         {
+        if (GetFileAttributes(lpszPath) != FILE_ATTRIBUTE_DIRECTORY)
 #endif
         lstrcpyW(lpszPath + iLen, pszExts[iChoose]);
         if (PathFileExistsW(lpszPath))
--- a/dll/win32/shlwapi/path.c
+++ b/dll/win32/shlwapi/path.c
@@ -1353,7 +1353,7 @@ BOOL WINAPI PathFindOnPathExW(LPWSTR lpszFile,LPCWSTR *lppszOtherDirs,DWORD dwWh
 
   TRACE("(%s,%p,%08x)\n", debugstr_w(lpszFile), lppszOtherDirs, dwWhich);
 
-  if (!lpszFile || !PathIsFileSpecW(lpszFile))
+  if (!PathIsFileSpecW(lpszFile) || !lpszFile)
     return FALSE;
 
   /* Search provided directories first */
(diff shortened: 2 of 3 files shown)


learn-more:
I'd like to run the testbot against this PR, will do so later (unless someone beats me to it)

dll/win32/shell32/shlexec.cpp
```diff
@@ -2531,6 +2531,9 @@ HRESULT WINAPI ShellExecCmdLine(
     }
     else
     {
+        PCWCHAR m_pPathList[2];
+        PCWCHAR *ppPathList;
```
katahiromz, line 2535:
```suggestion
        PCWCHAR apPathList[2];
```

HBelusca (approved):

HBelusca:
Test VBox: https://reactos.org/testman/compare.php?ids=85251,85254
Test KVM: https://reactos.org/testman/compare.php?ids=85250,85255 suspicious shell32:shlexec behaviour: https://reactos.org/testman/diff.php?id1=61758223&id2=61764208&type=1&strip=0 (doesn't show on Test VBox)

learn-more (changes requested):
Those look like test regressions.
@zecarlos1957 please investigate / fix those.

In case you are absolutely sure that those regressions are not a problem, please explain that here.

HBelusca:
Testing PR with the temporary commit, merged on top of master:
Test VBox: https://reactos.org/testman/compare.php?ids=85295,85297
Test KVM: https://reactos.org/testman/compare.php?ids=85294,85296 again, only here does the shell32:shlexec test behaviour changes:
https://reactos.org/testman/diff.php?id1=61810799&id2=61813193&type=1&strip=0
But we now have: `1679 tests executed (0 marked as todo, 43 failures), 0 skipped`, contrary to in master where `1677 tests executed (0 marked as todo, 46 failures), 0 skipped`. So there it would now appear there are some improvements.

zecarlos1957:
@learn-more 
When we digit the first letter in RunDlg  combobox to write some command, the PathFileExistsW function is called by six times with meaningless arguments, called from CLSID_ACListISF object created by  SHAutoComplete(hwndEdit, SHACF_FILESYSTEM | SHACF_FILESYS_ONLY | SHACF_URLALL);
called when processing WM_INITDIALOG message from RunDlgProc. You can view the result teste when we comment  the  line 566 of the dll/win32/shell32/dialogs/dialog.cpp file (in order not to execute the instruction SHAutoComplete(hwndEdit, SHACF_FILESYSTEM | SHACF_FILESYS_ONLY | SHACF_URLALL);
Do you want me to open a new jira ticket about this issue? So what is not directly related to this PR?

zecarlos1957:
PathFindOnPath
@learn-more  I have fix all failed testes on PathFindOnPath apitest

dll/win32/shlwapi/path.c
```diff
@@ -1353,7 +1353,7 @@ BOOL WINAPI PathFindOnPathExW(LPWSTR lpszFile,LPCWSTR *lppszOtherDirs,DWORD dwWh
 
   TRACE("(%s,%p,%08x)\n", debugstr_w(lpszFile), lppszOtherDirs, dwWhich);
 
-  if (!lpszFile || !PathIsFileSpecW(lpszFile))
+  if (!PathIsFileSpecW(lpszFile) || !lpszFile)
```
learn-more, line 1356: Why did you reorder this?

zecarlos1957: @learn-more  We take advantage of the fact that the statements inside the if are executed from right to left.
First it checks if the pointer is valid (not NULL) if TRUE executes the left instruction.
This prevents the pointer from being dereferenced if it is null.

zecarlos1957: elegant isn't it?

JoachimHenze: wouldn't it be much better than to turn that from an OR condition into an AND condition where we do have the guarantee that stuff is evaluated from left to right? that would be much more readable.
You remember that each condition of the kind
(!A) OR (!B)
can be written logically equivalent also as
!(A AND B).

JoachimHenze (changes requested):
see my comment above.

dll/win32/shlwapi/path.c
learn-more: > @learn-more We take advantage of the fact that the statements inside the if are executed from right to left. First it checks if the pointer is valid (not NULL) if TRUE executes the left instruction. This prevents the pointer from being dereferenced inside PathIsFileSpecW if it is null.

Except the statements are executed left to right.

learn-more: and `PathIsFileSpecW` returns `FALSE` for a nullptr, so it wouldn't make any difference anyway.

JoachimHenze: learn-more: "Except the statements are executed left to right."
I always thought that there is no guarantee at all about the order of evaluations for OR conditions.
Only AND conditions are guaranteed to be evaluated from left to right iirc. Therefore my proposal for the change.

learn-more: "and PathIsFileSpecW returns FALSE for a nullptr, so it wouldn't make any difference anyway."
I have not double-checked that, and therefore can neither confirm nor want to put that into question.

JoachimHenze: Given the following answer on stackoverflow https://stackoverflow.com/a/2456415 it seems like Mark Jansen is right and both OR and AND should guarantee evaluation from left to right.
But that doesn't reflect what zecarlos claimed (that exchanging the conditions in the OR would make a difference), and it also contradicts what I do remember personally for some reason.

zecarlos1957: I don't understand. I made this change because I had a failed test due to an exception (dereferencing a NULL pointer).
But now I see that this exception does not occur???

dll/win32/shlwapi/path.c
```diff
@@ -1353,7 +1353,7 @@ BOOL WINAPI PathFindOnPathExW(LPWSTR lpszFile,LPCWSTR *lppszOtherDirs,DWORD dwWh
 
   TRACE("(%s,%p,%08x)\n", debugstr_w(lpszFile), lppszOtherDirs, dwWhich);
 
-  if (!PathIsFileSpecW(lpszFile) || !lpszFile)
+  if (!(lpszFile && PathIsFileSpecW(lpszFile)))
```
learn-more, line 1356: The original here was way more readable.

JoachimHenze:
Now that you reverted some parts of your fix, I guess it would make sense to strip
[SHLWAPI] Prevent calling a NULL pointer taking advantage of compiler executing instructions from right to lef
from the PRs description.
I do assume that at least one unittest is failing now again, based on what you wrote earlier.
Since he pushed for that revert, I do assume Mark Jansen has a better idea to address that UT failure?

JoachimHenze:
VBOX https://reactos.org/testman/compare.php?ids=85908,85920 LGTM
KVM https://reactos.org/testman/compare.php?ids=85909,85921 shows additional failures in shell32:shlexec, but I am not sure whether they are related or whether that testsuite is just flaky.
shlwapi:PathFindOnPath-tests do look fine now.

JoachimHenze:
Okay, here is the proof that shell32:shlexec failures are random on KVM:
https://reactos.org/testman/compare.php?ids=85921,85923
That outcome makes me approve the PR in its current form.

JoachimHenze (approved):

learn-more (approved):

dll/win32/shlwapi/path.c
```diff
@@ -1151,6 +1151,7 @@ BOOL WINAPI PathFileExistsDefExtW(LPWSTR lpszPath,DWORD dwWhich)
 #ifdef __REACTOS__
         if (dwWhich & 0x1)
         {
+        if (GetFileAttributes(lpszPath) != FILE_ATTRIBUTE_DIRECTORY)
```
HBelusca, line 1154: Just to confirm: By this new if-case, your intent is to just have the lstrcpyW call done below called only when the file is not a directory? (while the PathFileExistsW is always being done?)

HBelusca (approved):

dll/win32/shlwapi/path.c
zecarlos1957: Yes ,correct. 
Prevent adding  executable extension to directory name

PR #4934, opened 2022-12-08
[SHELL32][USER32][USERENV] Improve Portuguese (pt-PT) translation

zecarlos1957:
[SHELL32][USER32] Improve pt-PT translation
[USERENV] Don't translate Documents and Settings folder

### DIFF
--- a/base/applications/msconfig/lang/pt-PT.rc
+++ b/base/applications/msconfig/lang/pt-PT.rc
@@ -66,13 +66,13 @@ IDD_GENERAL_PAGE DIALOGEX 0, 0, 362, 175
 STYLE DS_SHELLFONT | DS_CONTROL | WS_CHILD | WS_CLIPCHILDREN
 FONT 8, "MS Shell Dlg"
 BEGIN
-    GROUPBOX "Selecção de inicialização", -1, 10, 10, 340, 150, 0, WS_EX_TRANSPARENT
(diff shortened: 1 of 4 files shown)


KRosUser2:
[USERENV] Dont translate Documents and Settings folder => Shouldn't this be fixed for all language ? Looks like an error and should be "%SystemDrive%\\Documents and Settings" regardless of the langage isn't it ?

https://git.reactos.org/?p=reactos.git&a=search&st=grep&s=IDS_PROFILEPATH

HBelusca:
I confirm that in both portuguese from Portugal and Brazil Windows XP and 2003 versions, the "Documents and Settings" is not translated:
winxp_pt_docs_settings
Maybe it was translated in Windows 2000 (this is what internet seems to suggest, see e.g. http://www.josevalter.com.br/dicas_windows_2000.htm )

HBelusca:
Correction for commit log: "Dont translate" --> "Don't translate"

HBelusca (approved):

JoachimHenze:
> [USERENV] Dont translate Documents and Settings folder => Shouldn't this be fixed for all language ? Looks like an error and should be "%SystemDrive%\Documents and Settings" regardless of the langage isn't it ?
> 
> https://git.reactos.org/?p=reactos.git&a=search&st=grep&s=IDS_PROFILEPATH
> 
> Or at least to be checked for other localization...

Done. There **are** other languages where **MS** does indeed translate that folders name, e.g. German where it is called "Dokumente und Einstellungen".

PR #4936, opened 2022-12-09
[OLD_WORDPAD] Dump it

JoachimHenze:
To speed up the build and shrink the releases
iso size a bit. It doesn't work anymore.
And we have the well-working wordpad from Wine.


JoachimHenze:
old_wordpad3
oldwordpad2

HBelusca:
> To speed up the build and shrink the releases

Speed up by 1 second? Otherwise, since it's in rosapps only, it's only built iff `-DENABLE_ROSAPPS=1` is set, which is, only when an actual ROS release is built (and not the nightlies).

JoachimHenze:
This is correct. That is why I wrote: shrink the **releases** iso size.
But the argument of having less code to maintain does apply always.

HBelusca:
The "templates" directory has been excluded from build in master; this PR isn't necessary anymore.

JoachimHenze:
Okay, I am fine with that solution.

PR #4941, opened 2022-12-13
[WIN32K:NTGDI] PATH_PathToRegion(): Update documentation

SergeGautherie:
## Purpose

Missed change.

Addendum to 215e0d9 (r64177).

## Proposed changes

- `s/*pHrgn/Rgn/`

### DIFF
--- a/win32ss/gdi/ntgdi/path.c
+++ b/win32ss/gdi/ntgdi/path.c
@@ -1442,9 +1442,8 @@ PATH_FlattenPath(PPATH pPath)
 
 /* PATH_PathToRegion
  *
- * Creates a region from the specified path using the specified polygon
- * filling mode. The path is left unchanged. A handle to the region that
- * was created is stored in *pHrgn.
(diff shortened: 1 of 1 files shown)


win32ss/gdi/ntgdi/path.c
```diff
@@ -1444,7 +1444,7 @@ PATH_FlattenPath(PPATH pPath)
  *
  * Creates a region from the specified path using the specified polygon
  * filling mode. The path is left unchanged. A handle to the region that
```
zefklop, line 1446: That would be more `Fills Rgn from the specified path etc.` There is no handle allocated here.

zefklop (changes requested):

zefklop:
Yolo behemoth commits without peer-review. Those were good old days 👴

win32ss/gdi/ntgdi/path.c
SergeGautherie: Updated.

zefklop (approved):

PR #4950, opened 2022-12-19
[SHELL32] Give a chance to rename new file or folder when created

zecarlos1957:
[SHELL32] Give a chance to rename new file or folder when created 

OnChangeNotify  on event SHCNE_CREATE and SHCNE_MKDIR call SelectItem with SVSI_EDIT flag set.

### DIFF
--- a/dll/win32/shell32/wine/pidl.c
+++ b/dll/win32/shell32/wine/pidl.c
@@ -520,6 +520,10 @@ BOOL _ILHACKCompareSimpleIds(LPCITEMIDLIST pidltemp1, LPCITEMIDLIST pidltemp2)
         if (!_ILIsDrive(pidltemp1) || !_ILIsDrive(pidltemp2) || pdata1->u.drive.szDriveName[0] != pdata2->u.drive.szDriveName[0])
             return FALSE;
     }
+    else
+    {
+        return FALSE;
+    }
 
     return TRUE;
 }
--- a/dll/win32/shell32/wine/pidl.c
+++ b/dll/win32/shell32/wine/pidl.c
@@ -520,10 +520,6 @@ BOOL _ILHACKCompareSimpleIds(LPCITEMIDLIST pidltemp1, LPCITEMIDLIST pidltemp2)
         if (!_ILIsDrive(pidltemp1) || !_ILIsDrive(pidltemp2) || pdata1->u.drive.szDriveName[0] != pdata2->u.drive.szDriveName[0])
             return FALSE;
     }
-    else
-    {
-        return FALSE;
-    }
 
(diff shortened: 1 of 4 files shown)


dll/win32/shell32/wine/pidl.c
```diff
@@ -555,8 +555,11 @@ BOOL WINAPI ILIsEqual(LPCITEMIDLIST pidl1, LPCITEMIDLIST pidl2)
     while (pidltemp1->mkid.cb && pidltemp2->mkid.cb)
     {
         if (!_ILHACKCompareSimpleIds(pidltemp1, pidltemp2))
+#ifndef __REACTOS__
             return FALSE;
-
+#else
+            return TRUE;
+#endif
```
JoachimHenze, line 562: That part looks like a hack. What is the symptom without that change?

zecarlos1957: ILIsEqual must return TRUE when objects are equal
If IsEqual returns FALSE the rename function is aborted.

JoachimHenze: But I do assume that this code is not only run when renaming the Network Connections object.

learn-more, line 562: So is this a hack on top of a hack?

zecarlos1957: I'm pretty sure it's a bug. I'll run the ILIsEqual rostest to make sure

learn-more: This is a function that is pretty crucial to the inner workings of the shell, and a change like this is hard to guesstimate the impact.
It would help a lot if you can prove your changes with an apitest.

dll/win32/shell32/wine/pidl.c
```diff
@@ -520,10 +520,6 @@ BOOL _ILHACKCompareSimpleIds(LPCITEMIDLIST pidltemp1, LPCITEMIDLIST pidltemp2)
         if (!_ILIsDrive(pidltemp1) || !_ILIsDrive(pidltemp2) || pdata1->u.drive.szDriveName[0] != pdata2->u.drive.szDriveName[0])
             return FALSE;
     }
-    else
-    {
-        return FALSE;
-    }
```
learn-more, line 526: I very much doubt that this is correct, it basically classifies everything it does not know as 'equal'

learn-more (approved):

yagoulas:
I think that with this, opening two times the same dir and creating a file in one of them, will make both views enter rename mode

learn-more:
Confirmed:

image

PR #4951, opened 2022-12-20
[MSI] Revert to previous working solution for disk usage display in dialog.c

muthu90tech:
## Purpose

There is a bug, which displays empty optical drive as a valid Drive where applications can be installed, which is wrong. This change will skip displaying any drives that are not fixed in the disk usage table in the msi installer UI.

JIRA issue: [CORE-18758](https://jira.reactos.org/browse/CORE-18758)

## Proposed changes

Revert to old solution from Doug Lyons which was working fine.
see [diff1](https://git.reactos.org/?p=reactos.git;a=commitdiff;h=d5265b07bbfb6ac6da190b8213cf2f3f8d9cd675;hp=05fb0f1c392484d774d0fafbb6ca6e0317a69fa7), [diff2](https://git.reactos.org/?p=reactos.git;a=commitdiff;h=864e20b881130cbc9668b8e253a31b8c325e9b99;hp=83fcd65700e8cf6d0c07c81d85a07e68c5d87cdc)

### DIFF
--- a/dll/win32/msi/dialog.c
+++ b/dll/win32/msi/dialog.c
@@ -3206,7 +3206,7 @@ static void msi_dialog_vcl_add_drives( msi_dialog *dialog, msi_control *control 
     WCHAR cost_text[MAX_PATH];
     LPWSTR drives, ptr;
     LVITEMW lvitem;
-    DWORD size, flags;
+    DWORD size;
     int i = 0;
 
     cost = msi_vcl_get_cost(dialog);
@@ -3223,13 +3223,13 @@ static void msi_dialog_vcl_add_drives( msi_dialog *dialog, msi_control *control 
     ptr = drives;
     while (*ptr)
     {
-        if (GetVolumeInformationW(ptr, NULL, 0, NULL, 0, &flags, NULL, 0) &&
-            flags & FILE_READ_ONLY_VOLUME)
+#ifdef __REACTOS__
+        if (GetDriveTypeW(ptr) != DRIVE_FIXED)
         {
             ptr += lstrlenW(ptr) + 1;
             continue;
         }
-
+#endif
         lvitem.mask = LVIF_TEXT;
         lvitem.iItem = i;
         lvitem.iSubItem = 0;


KRosUser2:
Lgtm as discussed in the Jira ticket

learn-more (dismissed):
Did you check that `GetVolumeInformationW` is actually returning something correct, or are you just hiding a bug?

KRosUser2:
> Did you check that `GetVolumeInformationW` is actually returning something correct, or are you just hiding a bug?

Same comment on GetDiskFreeSpaceExW below

KRosUser2:
And final patch to be sent upstream (see link to Wine in associated JIRA ticket)

TAN-Gaming:
> given that Windows XP also forbids MSI packages being installed to **writable** USB sticks.

Why this restriction needs to be ported to ReactOS? For compatibility?

learn-more:
> @learn-more Checking any functions return values is not even needed to prove Wines solution being wrong. they do check for FILE_READ_ONLY_VOLUME whcih is obviously the wrong check, given that Windows XP also forbids MSI packages being installed to **writable** USB sticks. It is obvious that their code is wrong.


Then I am curious to see how wine behaves in this regard.

KRosUser2:
Is any change still required ? Or is this one OK now ?

KRosUser2:
> > @learn-more Checking any functions return values is not even needed to prove Wines solution being wrong. they do check for FILE_READ_ONLY_VOLUME whcih is obviously the wrong check, given that Windows XP also forbids MSI packages being installed to **writable** USB sticks. It is obvious that their code is wrong.
> 
> Then I am curious to see how wine behaves in this regard.

@Doug-Lyons @JoachimHenze : Is it something you tested in the past when proposing the design leading to this PR ?

learn-more:
At least ensure that the diff is a proper diff, and not just a plain revert,
because now the actual code is just deleted again instead of `#ifdef`d out

KRosUser2:
> At least ensure that the diff is a proper diff, and not just a plain revert, because now the actual code is just deleted again instead of `#ifdef`d out

Right, I fully agree. @muthu90tech would you please #ifndef the origine code (or #else) rather than deleting it ?

dll/win32/msi/dialog.c
```diff
@@ -3206,7 +3206,7 @@ static void msi_dialog_vcl_add_drives( msi_dialog *dialog, msi_control *control
     WCHAR cost_text[MAX_PATH];
     LPWSTR drives, ptr;
     LVITEMW lvitem;
-    DWORD size, flags;
+    DWORD size;
```
binarymaster, line 3209: Non-guarded diff.

KRosUser2:
@Doug-Lyons : this reverts to your original design.  Would you like to take over the final change ?

atharva1910:
The remote branch was deleted so I couldn't continue here. Please can we move to [5194](https://github.com/reactos/reactos/pull/5194)

JoachimHenze:
superseded by https://github.com/reactos/reactos/pull/5194

PR #4954, opened 2022-12-22
[SHELLFIND] Show 'Relevance' column by default.

Cleverking2003:
## Purpose

Show 'Relevance' column by default.

JIRA issue: [CORE-18629](https://jira.reactos.org/browse/CORE-18629)

image


JoachimHenze (changes requested):
Rejected. My XPSP3 does not show this column by default. And I also see no need to waste any screenspace with that at all.

Cleverking2003:
> Rejected. My XPSP3 does not show this column by default. And I also see no need to waste any screenspace with that at all.

Maybe then we should just close the ticket?

learn-more:
Using '30' as width here is still a legit fix.

PR #4955, opened 2022-12-22
[NTOS:CM] Don't lazy flush the registry during unlocking operation

GeoB99:
This patch fixes a problem with the lazy flusher not being able to flush the registry hives. This fix is a split from #4571, as requested by @tkreuzer.
[CORE-18303](https://jira.reactos.org/browse/CORE-18303)
## NOTE
This PR does not implement registry validation checks in any form, nor does it implement recoverable corrupted registry routines! For those, you have to look for #4571.


JoachimHenze (approved):
I do very much like the split-off of this minimal part from the bigger PR, to get it into production earlier.

Extravert-ir:
Hm, you are saying that the problem is with `CmpLazyFlush`, which resets the timer with every call.
Why not fix it instead so it doesn't reset the timer when it is already set? Doesn't look like a big deal

GeoB99:
> Hm, you are saying that the problem is with `CmpLazyFlush`, which resets the timer with every call. Why not fix it instead so it doesn't reset the timer when it is already set? Doesn't look like a big deal

`KeSetTimer` (which is called by `CmpLazyFlush`) resets the timer if it was already inserted into the timer table, and that's the exact proper implementation of that function. `CmpLazyFlush` simply should not be called when locking/unlocking the registry as this operation can occur for various reasons other than writing into a registry hive with lock held.

PR #4958, opened 2022-12-23
[IP][TCPIP] Update TCP and UDP dynamic port ranges

SergeGautherie:
## Purpose

Use higher ports, as updated Windows S2003 does.

JIRA issue: [CORE-18766](https://jira.reactos.org/browse/CORE-18766)

## Proposed changes

- Use IANA range (49152-65535), only.
- Add TODOs about `MaxUserPort` registry key.

### DIFF
--- a/drivers/network/tcpip/include/udp.h
+++ b/drivers/network/tcpip/include/udp.h
@@ -7,8 +7,9 @@
 
 #pragma once
 
-#define UDP_STARTING_PORT 0x8000
-#define UDP_DYNAMIC_PORTS 0x8000
+// TODO: Support MaxUserPort registry key.
+#define UDP_STARTING_PORT 49152
+#define UDP_DYNAMIC_PORTS (65535 - UDP_STARTING_PORT + 1)
 
 /* UDPv4 header structure */
 #include <pshpack1.h>
--- a/drivers/network/tcpip/ip/transport/tcp/tcp.c
+++ b/drivers/network/tcpip/ip/transport/tcp/tcp.c
@@ -638,8 +638,9 @@ UINT TCPAllocatePort(const UINT HintPort)
             return (UINT)-1;
         }
     }
-    else
-        return AllocatePortFromRange( &TCPPorts, 1024, 5000 );
+
+    // TODO: Support MaxUserPort registry key.
+    return AllocatePortFromRange(&TCPPorts, 49152, 65535);
 }
 
 VOID TCPFreePort(const UINT Port)


KRosUser2:
Note that full TCP port range (up to 0xFFFF) will be added when the patch from https://jira.reactos.org/browse/CORE-18371 gets PR'd

learn-more:
And we are not going to hit the asserts now, due to this?

KRosUser2:
> And we are not going to hit the asserts now, due to this?

Well, we ALREADY hit the ASSERT anyway.

KRosUser2:
This PR will likely fix https://jira.reactos.org/browse/CORE-18806 too.

SergeGautherie:
> This PR will likely fix https://jira.reactos.org/browse/CORE-18806 too.

What makes you think that?

sdk/lib/drivers/ip/transport/tcp/tcp.c
```diff
@@ -632,8 +632,9 @@ UINT TCPAllocatePort(const UINT HintPort)
             return (UINT)-1;
         }
     }
-    else
-        return AllocatePortFromRange( &TCPPorts, 1024, 5000 );
+
+    // TODO: Support MaxUserPort registry key.
+    return AllocatePortFromRange(&TCPPorts, 49152, 65535);
```
KRosUser2, line 637: 49152 or UDP_STARTING_PORT

KRosUser2:
@SergeGautherie any documentation justifying this change?

SergeGautherie (commented):
> And we are not going to hit the asserts now, due to this?

That separate issue(s) was fixed in the meantime.

SergeGautherie:
Looking for an (active) reviewer...

DarkFire01 (approved):

SergeGautherie:
@ThFabba, Could you approve and commit this PR, which has stalled for about 1/5 month(s)? Thanks.

ThFabba:
I'm not approving anything in the networking stack that's not backed by tests.

SergeGautherie:
> (API- or Wine-) tests.

Ok... What kind of tests would you want? Simply checking that the port number is in the expected range (or failing)?

ThFabba:
A test that fails without your change and passes with it would be ideal. I'm guessing something like this would work?
```
for (i = 0; i < 1000000; i++)
{
    bind(socket, localhost, port=0);
    getsockname(socket, &sockname);
    if (sockname.port < min_port) min_port = sockname.port;
    if (sockname.port > max_port) max_port = sockname.port;
}
ok(min_port == 49152);
ok(max_port == 65535);
```

SergeGautherie:
> ```c
>     bind(socket, localhost, port=0);
> ```

I did a minimal check addition to existing ws2_32_apitest:bind code (existing cases).
Results on WXP and WS2003 are identical: port cycles through the whole 1-65535 range. (Next port = previous port + 256)
That is kind of unexpected: not old nor new ranges.

To double-check, I made a few ftp connections on WXP, and got 4nnn ports, (Next port = previous port + 1or2)
no matter the latest small range apitest reported.

Then, `bind()` would not seem right/enough to test the expected range(s).
Suggestion?

SergeGautherie:
Downgrading to Draft, until I/someone come up with a useful test (idea).

DarkFire01:
No tests. Closed for now. Thank you anyway!! Please re open if you wish to continue

PR #4960, opened 2022-12-26
[COMCTL32] Do not start tracking mouse input in combobox on the opening click

Cleverking2003:
## Purpose

On small screens there is a problem that a big combobox closes itself as if it was clicked.
Fix this issue by enabling mouse capture only after the opening click is complete.

JIRA issue: [CORE-18769](https://jira.reactos.org/browse/CORE-18769)

### DIFF
--- a/dll/win32/comctl32/combo.c
+++ b/dll/win32/comctl32/combo.c
@@ -1610,8 +1610,6 @@ static void COMBO_LButtonDown( LPHEADCOMBO lphc, LPARAM lParam )
        else
        {
 	   /* drop down the listbox and start tracking */
-
-           lphc->wState |= CBF_CAPTURE;
            SetCapture( hWnd );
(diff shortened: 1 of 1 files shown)


HBelusca:
This needs to be sent upstream to Wine.

Cleverking2003:
Pushed a small fix to move `SetCapture` call back to where it was. Otherwise, the arrow button wouldn't become unpressed.

DarkShadow44:
Wine doesn't have this issue, they won't accept the change.

learn-more:
Your comments do not align with your code at all.
You suggest that you only start tracking when the thing is opened, but you did not change tracking at all.
The only thing you changed was change when a _flag_ that is supposed to track the state of tracking or not,
so that it does not actually reflect what it was supposed to do anymore.

PR #4962, opened 2022-12-28
Amd64/some fixes

tkreuzer:
## Purpose
A number or misc fixes.

Testbot: https://reactos.org/testman/compare.php?ids=85356,85365,85367,85378,85384 (rightmost is this PR)


dll/win32/winmm/mci.c
```diff
@@ -37,24 +37,32 @@
...
+#include "winreg.h"
+#include "wownt32.h"
+
+#include "digitalv.h"
 #include "winemm.h"
 
-#include <mmsystem.h>
-#include <wownt32.h>
-#include <digitalv.h>
+#include "wine/debug.h"
```
HBelusca, line 57: @TAN-Gaming Does this sync conflict with one of your WINMM PRs?

win32ss/user/ntuser/window.h
```diff
@@ -124,4 +124,21 @@ BOOL FASTCALL IntBroadcastImeShowStatusChange(PWND pImeWnd, BOOL bShow);
 VOID FASTCALL IntNotifyImeShowStatus(PWND pImeWnd);
 VOID FASTCALL IntCheckImeShowStatusInThread(PWND pImeWnd);
 
+static inline
+VOID
+WndSetOwner(PWND pwnd, PWND pwndOwner)
```
HBelusca, line 129:
```suggestion
WndSetOwner(_In_ PWND pwnd, _In_opt_ PWND pwndOwner)
```

win32ss/user/ntuser/winpos.c
```diff
@@ -436,12 +436,12 @@ co_WinPosActivateOtherWindow(PWND Wnd)
 
    // Find any window to bring to top. Works Okay for wine since it does not see X11 windows.
    WndTo = UserGetDesktopWindow();
-   WndTo = WndTo->spwndChild;
-   if ( WndTo == NULL )
+   if ( WndTo == NULL || WndTo->spwndChild == NULL)
```
HBelusca, line 439:
```suggestion
   if (WndTo == NULL || WndTo->spwndChild == NULL)
```

win32ss/user/ntuser/desktop.c
```diff
@@ -1788,6 +1788,9 @@ BOOL IntDeRegisterShellHookWindow(HWND hWnd)
     PLIST_ENTRY ListEntry;
     PSHELL_HOOK_WINDOW Current;
 
+    if (Desktop == NULL)
```
HBelusca, line 1791: In which situation did you hit that Desktop == NULL case?

win32ss/user/ntuser/window.h
yagoulas: For the record this kind of assignment in windows is called an assignment lock and this scheme is used in a lot of places, not just owner windows.

dll/win32/shell32/CDefView.cpp
```diff
@@ -867,11 +867,15 @@ BOOL CDefView::_Sort()
 
 PCUITEMID_CHILD CDefView::_PidlByItem(int i)
 {
+    if (!m_ListView)
```
yagoulas, line 870: I suspect that this could be hiding another issue here. Under what conditions can it happen? I ask because Initialization of the CDefView is supposed to fail if this is null and this gets destroyed in the destructor of the class.

dll/win32/winmm/mci.c
TAN-Gaming: No, I didn't make any change to this file so it's fine.

dll/win32/winmm/winemm.h
```diff
@@ -205,4 +203,9 @@ extern HANDLE psStopEvent;
 
 INT LoadRegistryMMEDrivers(char* key);
 
+// REACTOS:
+#define ARRAY_SIZE(a) (sizeof(a) / sizeof((a)[0]))
```
TAN-Gaming, line 207: How about using `_countof` ?

KRosUser2:
Looking at all the fixes around spwndOwner management, you might give a look at : https://jira.reactos.org/browse/CORE-18768 and https://jira.reactos.org/browse/CORE-18659

win32ss/user/ntuser/window.h
learn-more: When our win32k is compiled in c++ we can create an object for this.

win32ss/user/ntuser/window.h
```diff
@@ -124,4 +124,21 @@ BOOL FASTCALL IntBroadcastImeShowStatusChange(PWND pImeWnd, BOOL bShow);
...
+{
+    if (pwnd->spwndOwner != NULL)
+    {
+        UserDereferenceObject(pwnd->spwndOwner);
+    }
+
+    if (pwndOwner != NULL)
+    {
+        UserReferenceObject(pwndOwner);
```
zefklop, line 138: I May be paranoid here but please reference before dereferencing. In case that's the same window, this might lead it to be freed.

dll/win32/shell32/CDefView.cpp
tkreuzer: I don't think so. m_ListView is an object (not a pointer) and it is default-initialized.  The listview is "created" in CDefView::CreateList(), which is called from CDefView::OnCreate(), i.e. after the view window was created by IShellView_CreateViewWindow. The problem is an assertion failure that happens during shell32_winetest shlview in test_IFolderView.
```
00 00000000`0012f910 000007ff`a45bbb66     shell32!ATL::CWindow::SendMessageW+0x60 [E:\amd64\reactos\sdk\lib\atl\atlwin.h @ 1112] 
01 00000000`0012f950 000007ff`a45bbc10     shell32!CListView::GetItem+0x26 [E:\amd64\reactos\sdk\include\reactos\ui\rosctrls.h @ 109] 
02 00000000`0012f980 000007ff`a45c7995     shell32!CListView::GetItemData+0x60 [E:\amd64\reactos\sdk\include\reactos\ui\rosctrls.h @ 209] 
03 00000000`0012fa20 000007ff`a45bdf88     shell32!CDefView::_PidlByItem+0x25 [E:\amd64\reactos\dll\win32\shell32\CDefView.cpp @ 871] 
04 00000000`0012fa50 00000001`4007e668     shell32!CDefView::Item+0x28 [E:\amd64\reactos\dll\win32\shell32\CDefView.cpp @ 2819] 
05 00000000`0012fa90 00000001`4007ac48     shell32_winetest!test_IFolderView+0x258 [E:\amd64\reactos\modules\rostests\winetests\shell32\shlview.c @ 747] 
06 00000000`0012fcf0 00000001`400822a9     shell32_winetest!func_shlview+0x28 [E:\amd64\reactos\modules\rostests\winetests\shell32\shlview.c @ 1524] 
07 00000000`0012fd20 00000001`40082193     shell32_winetest!run_test+0xa9 [E:\amd64\reactos\sdk\include\reactos\wine\test.h @ 879] 
```

IFolderView_Item() is supposed to return E_INVALIDARG here, because there is no window and no list yet. CDefView::Item() does exactly this, if _PidlByItem() returns NULL.

win32ss/user/ntuser/window.h
tkreuzer: Yes, optimally all window pointers should be `shared_ptr`. But I fear to get there is a long way and probably means rewriting half of the USER code in C++.

win32ss/user/ntuser/desktop.c
tkreuzer: It happens during user32_winetest msg_focus:
```
nt!KiPageFault+0x150
win32k!IntDeRegisterShellHookWindow+0x28 [E:\amd64\reactos\win32ss\user\ntuser\desktop.c @ 1791] 
win32k!co_UserFreeWindow+0x1c9 [E:\amd64\reactos\win32ss\user\ntuser\window.c @ 603] 
win32k!co_UserDestroyWindow+0x8a5 [E:\amd64\reactos\win32ss\user\ntuser\window.c @ 3006] 
win32k!UserDestroyObjectsForOwner+0x9e [E:\amd64\reactos\win32ss\user\ntuser\object.c @ 773] 
win32k!ExitThreadCallback+0x3e0 [E:\amd64\reactos\win32ss\user\ntuser\main.c @ 782] 
win32k!Win32kThreadCallback+0xe5 [E:\amd64\reactos\win32ss\user\ntuser\main.c @ 898] 
nt!PspExitThread+0xb67 [E:\amd64\reactos\ntoskrnl\ps\kill.c @ 751] 
```
For some reason the thread doesn't have a desktop at this point. I do not know why, but I know it's better to not crash the system.

dll/win32/winmm/winemm.h
tkreuzer: I copied this from wine. It's the same thing, really.

win32ss/user/ntuser/desktop.c
yagoulas: Interesting. This could only happen for the desktop thread as I understand. The desktop thread owns all desktop windows for all desktops, and the desktop windows are allocated from the desktop heap. The real question then is why this hasn't happened before.

dll/win32/shell32/CDefView.cpp
yagoulas: Hm indeed. This change is perfectly fine. All the rest of the methods of this class probably have the same issue.

yagoulas (approved):
The changes in ntuser and shell32 are all reasonable. At a later point the same kind of fixes should be applied more broadly in both.

dll/win32/winmm/winemm.h
TAN-Gaming: OK.
Should I replace `_countof` in my winmm PR with `ARRAY_SIZE` after this PR merged.

tkreuzer: I suggest doing so, because it will probably make it easier to get changes merged upstream. Btw, did you try to sync the files first, before changing them? I only synced mci, because it was easy (the rest is more complicated) and it had all the fixes I needed at this point. But generally we should try to sync it as much as possible.

TAN-Gaming: > did you try to sync the files first, before changing them?

No, I didn't do that.
My changes are completely different from Wine because I have to do some refactoring to get the system sound to work. The reason behind the changes is to support user impersonation that Wine doesn't have to deal with.

Note: I only changed `midimap.c` and `playsound.c`.

PR #4963, opened 2022-12-29
[USBSTOR_NEW] Add missing poppack.h

learn-more:


ThFabba (approved):

drivers/usb/usbstor_new/usbstor.h
```diff
@@ -547,4 +547,6 @@ USBSTOR_TimerRoutine(
     PDEVICE_OBJECT DeviceObject,
      PVOID Context);
 
+#include <poppack.h>
```
HBelusca, line 550: This poppack must be moved much more earlier above, to line 314, just after the `UFI_UNKNOWN_CMD` structure and before the hardware-unrelated `IRP_CONTEXT`.

HBelusca (changes requested):

drivers/usb/usbstor_new/usbstor.h
learn-more: Why?

HBelusca: **Why** would you want the non-hardware-related IRP_CONTEXT and other structures that follow it to be packed on 1-byte boundary, instead of the default platform one?

learn-more: Because that is what was done previously.

HBelusca (approved):

binarymaster (approved):

PR #4964, opened 2022-12-29
[SDK] Improve gen_baseaddress.py and its output files

JoachimHenze:
## Purpose

[SDK:TOOLS] Improve gen_baseaddress.py

Adds some kbd layout modules into the excludes-section:
that we do lack for todays master: e.g: kbdeo, kbdsf, kbdrost

and also some that we lacked in the past
(for improving backwards compatibility of the script to older rls-branches):
kbdgm, kbdes, kbdgrist, kbdja, kbdko, kbdsk, kbdsk1

[SDK:CMAKE] Highlight in the output files where the script does misbehave

## TODO for future PRs
- at some point in the future we have to fix the py script from misbehaving for fusion, fusion_1_1, fusion_2_0, localspl_apitest.dll

## Parts that were voted out by Mark Jansen and therefore are no longer part of this PR:
switch back the padding to 22 instead of 30 spaces,
like we used it in older branches until
0.4.10-dev-307-g 82351bc6d410f020b5589cfdda628073753f78f4 for the outputs
as that makes the output-files a bit smaller again and is enough padding
for even the longest string 'baseaddress_msvcrt_crt_dll_startup'.
It is the exact same amount of padding we used earlier.


sdk/tools/gen_baseaddress.py
```diff
@@ -252,7 +262,7 @@ def gen_baseaddress(self, output_file):
             postfix = ' # should be above 0x%08x' % self.address
         elif self._reserved:
             postfix = ' # reserved'
-        output_file.write('set(baseaddress_%-30s 0x%08x)%s\n' % (name, self.address, postfix))
+        output_file.write('set(baseaddress_%-22s 0x%08x)%s\n' % (name, self.address, postfix))
```
learn-more, line 265: What is the point of this change?

JoachimHenze: See what I wrote in the PRs description:
> Also switch back the padding to 22 instead of 30 spaces,
> like we used it in older branches until
> 0.4.10-dev-307-g https://github.com/reactos/reactos/commit/82351bc6d410f020b5589cfdda628073753f78f4 for the outputs
> as that makes the output-files a bit smaller again and is enough padding
> for even the longest string 'baseaddress_msvcrt_crt_dll_startup'.
> It is the exact same amount of padding we used earlier.

learn-more: right, so it is a change for the sake of changing it.

learn-more (dismissed):

HBelusca:
> at some point in the future we have to fix the py script from misbehaving for fusion [...]

Can you (re)explain what is this misbehaviour about?

sdk/cmake/baseaddress.cmake
```diff
@@ -196,7 +196,7 @@ set(baseaddress_load_notifications             0x73cd0000)
 set(baseaddress_loadperf                       0x73ca0000)
 set(baseaddress_localmon                       0x73c40000)
 set(baseaddress_localspl                       0x73b50000)
-set(baseaddress_localspl_apitest               0x73b20000)
+set(baseaddress_localspl_apitest.dll           0x73b20000) #module name corrected manually
```
HBelusca, line 199: Why does this one need the extension?

JoachimHenze:
> > at some point in the future we have to fix the py script from misbehaving for fusion [...]
> 
> Can you (re)explain what is this misbehaviour about?

Sure, we don't set a base address for them https://jira.reactos.org/browse/CORE-14831

sdk/cmake/baseaddress.cmake
JoachimHenze: because otherwise it won't get a base address, see https://jira.reactos.org/browse/CORE-14831
I m not sure, why this one is so special, but we touched and renamed that stuff multiple times in the past, and composed part of its name. That was around 0.4.7 to 0.4.9 iirc. Thomas Faber might remember, he was involved in that. Our python script was never able to cover that dll correctly.

HBelusca: Ah, this is because the `baseaddress_*` names there are based on the cmake build targets. And the generated `localspl_apitest` dll cmake build target name contains the extension (in order to disentangle it from the other build target that would otherwise have the same name, but builds the test .exe): see the first results there:
https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=localspl_apitest

sdk/cmake/baseaddress.cmake
```diff
@@ -181,9 +181,9 @@ set(baseaddress_itss                           0x745e0000)
 set(baseaddress_joy                            0x745a0000)
 set(baseaddress_jscript                        0x74400000)
 set(baseaddress_jsproxy                        0x743d0000)
-set(baseaddress_kbdeo                          0x743b0000)
-set(baseaddress_kbdrost                        0x74390000)
-set(baseaddress_kbdsf                          0x74370000)
+#set(baseaddress_kbdeo                          0x743b0000) #unwanted, should vanish when rerunning the script
+#set(baseaddress_kbdrost                        0x74390000) #unwanted, should vanish when rerunning the script
+#set(baseaddress_kbdsf                          0x74370000) #unwanted, should vanish when rerunning the script
```
HBelusca, line 186: Then, why not just manually deleting those lines? (in this baseaddress.cmake file and the others)

JoachimHenze: I could have done this also yes. But by commenting them, we do still see the "gaps" in the base-addresses-ranges. Next time the script will be run, they will be gone.

JoachimHenze: The lines are left for verifying the effectiveness of the changes when running the script later.

JoachimHenze:
@learn-more I waited 2months for approval here. Is there anything wrong with this PR?

learn-more (approved):
Don't really see the point in manually editing those .cmake files without fixing up the `.py` script, but sure.

PR #4966, opened 2022-12-30
[MSI] Fix empty disc costing (Terragen, ...)

rexxDigital:
## Purpose

Temporary workaround for empty disc cost on terragen, etc.
Changing return to continue fixes the issue.

Before:
image

After:
image

JIRA issue: [CORE-18749](https://jira.reactos.org/browse/CORE-18749)

## Proposed changes

- change `return;` to `continue;`


oleg-dubinskiy:
If you have the same problem in Wine, then it should be sent to upstream.
https://source.winehq.org/git/wine.git/blob/HEAD:/dlls/msi/dialog.c#l3163

KRosUser2:
> If you can reproduce the same problem in Wine, then it should be sent to upstream. https://source.winehq.org/git/wine.git/blob/HEAD:/dlls/msi/dialog.c#l3163

Already reported upstream : https://bugs.winehq.org/show_bug.cgi?id=54164

dll/win32/msi/dialog.c
```diff
@@ -3160,7 +3160,11 @@ static void msi_dialog_vcl_add_columns( msi_dialog *dialog, msi_control *control
          */
         if ( !wcsncmp( num, L"-", 1 ) || !str_is_number( num ) ) {
             msi_free( num );
+            #ifndef __REACTOS__
```
SergeGautherie, line 3163: Move `#...` to the first column.

dll/win32/msi/dialog.c
```diff
@@ -3160,7 +3160,11 @@ static void msi_dialog_vcl_add_columns( msi_dialog *dialog, msi_control *control
...
             return;
+            #else
+            continue;
```
SergeGautherie, line 3166: At least, the existing comment does not match anymore...

learn-more (changes requested):
This is a workaround or even an hack, not a fix for the problem, and thus it should be clearly marked as such.

KRosUser2:
> This is a workaround or even an hack, not a fix for the problem, and thus it should be clearly marked as such.

This is a fix for a bug, but a complementary "improvement" PR is indeed required. @rexxDigital : I suggest a simple comment about it in the change.

rexxDigital:
> This is a fix for a bug, but a complementary "improvement" PR is indeed required. @rexxDigital : I suggest a simple comment about it in the change.

Added a comment in the code + changed purpose text

learn-more (changes requested):
Considering that this is a hack, and it's a visual thing only I don't like this being merged.

KRosUser2:
> Considering that this is a hack, and it's a visual thing only I don't like this being merged.

This is a total nonsense given the current weak design, but you're the boss, that's your project.
Please close the ticket as Won't Do then.

dll/win32/msi/dialog.c
```diff
@@ -3160,7 +3160,14 @@ static void msi_dialog_vcl_add_columns( msi_dialog *dialog, msi_control *control
          */
         if ( !wcsncmp( num, L"-", 1 ) || !str_is_number( num ) ) {
             msi_free( num );
+/* Temporary workaround for CORE-18749
+ * new issue will be opened for improvement on jira
```
SergeGautherie, line 3164: This line does not seem wanted in code.
Moreover, you have still not explained why this "workaround", as is, is not a regression with regard to existing feature: what happens if `num[0] == '-'`, for example?

learn-more:
> > Considering that this is a hack, and it's a visual thing only I don't like this being merged.
> 
> This is a total nonsense given the current weak design, but you're the boss, that's your project. Please close the ticket as Won't Do then.

As already mentioned on mattermost:
> Continuing on after something that you could not parse is inherently unsafe,
> and because there is no gain other than a small visual win here I don't see the gains outweigh the cost of another wine-diff

dll/win32/msi/dialog.c
```diff
@@ -3132,6 +3132,9 @@ static void msi_dialog_vcl_add_columns( msi_dialog *dialog, msi_control *control
     WCHAR *num;
     LVCOLUMNW lvc;
     DWORD count = 0;
+#ifdef __REACTOS__
+    BOOL bFirst = TRUE;
```
learn-more, line 3136: Why the new boolean, `count` already keeps track of the index?

dll/win32/msi/dialog.c
```diff
@@ -3160,7 +3160,19 @@ static void msi_dialog_vcl_add_columns( msi_dialog *dialog, msi_control *control
          */
         if ( !wcsncmp( num, L"-", 1 ) || !str_is_number( num ) ) {
             msi_free( num );
+#ifdef __REACTOS__
+            if ((count == 0) && (wcsncmp( num, L"\\", 1 ) || wcsncmp( num, L"&", 1 )))
```
learn-more, line 3164: The condition `(wcsncmp( num, L"\\", 1 ) || wcsncmp( num, L"&", 1 ))` is always true, so effectively this code is back to the first iteration where anything invalid is ignored.

You probably meant `(!wcsncmp( num, L"\\", 1 ) || !wcsncmp( num, L"&", 1 ))`

learn-more (changes requested):

SergeGautherie:
> new fix

Morphing this PR: I think a new PR would be cleaner to follow. Anyway...
At the very least, update the PR description too.

--

Moreover, I am not sure why such broken patches get PR'ed.
@rexxDigital, Do you understand and test these changes, before you submit them?

dll/win32/msi/dialog.c
SergeGautherie: Moreover, why is `wcsncmp()` needed, to compare a single character?

KRosUser2:
> > new fix
> 
> Morphing this PR: I think a new PR would be cleaner to follow. Anyway... At the very least, update the PR description too.
> 
> --
> 
> Moreover, I am not sure why such broken patches get PR'ed. @rexxDigital, Do you understand and test these changes, before you submit them?

Keep cool Serge, broken patch was mine and he did fix it here. Many thanks to @rexxDigital for the good job, looking forward for approval and merge

dll/win32/msi/dialog.c
```diff
@@ -3159,8 +3159,22 @@ static void msi_dialog_vcl_add_columns( msi_dialog *dialog, msi_control *control
          * if a width is invalid, all remaining columns are hidden
          */
         if ( !wcsncmp( num, L"-", 1 ) || !str_is_number( num ) ) {
+#ifdef __REACTOS__
+            if ((count == 0) && !(wcsncmp( num, L"\\", 1 ) || !wcsncmp( num, L"&", 1 )))
```
learn-more, line 3163:
```suggestion
            if ((count == 0) && (!wcsncmp( num, L"\\", 1 ) || !wcsncmp( num, L"&", 1 )))
```

Please review your own code before submitting it as a PR.....

learn-more (changes requested):

learn-more (approved):

dll/win32/msi/dialog.c
```diff
@@ -3159,8 +3159,22 @@ static void msi_dialog_vcl_add_columns( msi_dialog *dialog, msi_control *control
...
          */
         if ( !wcsncmp( num, L"-", 1 ) || !str_is_number( num ) ) {
+#ifdef __REACTOS__
+            if ((count == 0) && (!wcsncmp( num, L"\\", 1 ) || !wcsncmp( num, L"&", 1 )))
+            {
+                msi_free( num );
+                FIXME("Style prefix not supported. Skipping.\n");
+                continue;
+            }
+            else
```
SergeGautherie, line 3169: Why this redundant `else` block?

SergeGautherie:
> he did fix it here. Many thanks to @rexxDigital for the good job

I disagree, obviously.

KRosUser2:
> > he did fix it here. Many thanks to @rexxDigital for the good job
> 
> I disagree, obviously.

Then please add review comment on what is not fixed

SergeGautherie:
> add review comment on what is not fixed

Isn't it what I am already doing??

dll/win32/msi/dialog.c
```diff
@@ -3159,8 +3159,22 @@ static void msi_dialog_vcl_add_columns( msi_dialog *dialog, msi_control *control
          * if a width is invalid, all remaining columns are hidden
          */
         if ( !wcsncmp( num, L"-", 1 ) || !str_is_number( num ) ) {
+#ifdef __REACTOS__
+            if ((count == 0) && (!wcsncmp( num, L"\\", 1 ) || !wcsncmp( num, L"&", 1 )))
```
HBelusca, line 3163: @KRosUser2 Looking at your "better fix" patch in the WineHQ Bugzilla, that one has the redundant bFirst variable, and also the bug that you call `msi_free(num);` before having those `wcsncmp(num, ...)` calls. As such I think it should be instead fixed with something as being done in this PR.

dll/win32/msi/dialog.c
HBelusca: In what sense is this "else" block redundant?

HBelusca (approved):

dll/win32/msi/dialog.c
binarymaster: I think he means the diff can be reduced by removing else case from the condition, but also from the diff guard.

tkreuzer: Yes, please do so.

KRosUser2:
@rexxDigital : All pending comments are taken into account in the new patch provided in https://jira.reactos.org/browse/CORE-18749

dll/win32/msi/dialog.c
```diff
@@ -3159,8 +3159,22 @@ static void msi_dialog_vcl_add_columns( msi_dialog *dialog, msi_control *control
...
+            {
+                msi_free( num );
+                FIXME("Style prefix not supported. Skipping.\n");
```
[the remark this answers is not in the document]
learn-more: Please stop giving stylistic suggestions to suit your personal taste.

KRosUser2:
At least 3 apps identified in the associated ticket that would benefit from this fix.

dll/win32/msi/dialog.c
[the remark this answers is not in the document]
SergeGautherie: > stylistic suggestions [...] personal taste

This is fool-proofing, not what you write: people already got it wrong on this ticket...

JoachimHenze:
superseded by the better (#5201)

PR #4968, opened 2022-12-31
[IP] Avoid declaring 0xFFFF as invalid port in ReactOS

muthu90tech:
## Purpose

Do not declare 0XFFFF as invalid in ReactOS. Applications should be able to use 0XFFFF as it is valid. Thanks to @KRosUser for detailed analysis.

JIRA issue: 
[CORE-18371](https://jira.reactos.org/browse/CORE-18371)
[CORE-18764](https://jira.reactos.org/browse/CORE-18764)

## Proposed changes

The proposed fix is similar to what is already implemented in ReactOS for existing TCPAlloc failure in [FileOpenAddress](https://git.reactos.org/?p=reactos.git;a=blob;f=drivers/network/tcpip/tcpip/fileobjs.c#l475 )

### DIFF
--- a/drivers/network/tcpip/tcpip/fileobjs.c
+++ b/drivers/network/tcpip/tcpip/fileobjs.c
@@ -404,6 +404,7 @@ NTSTATUS FileOpenAddress(
   PVOID Options)
 {
   PADDRESS_FILE AddrFile;
+  UINT nPort;
 
   TI_DbgPrint(MID_TRACE, ("Called (Proto %d).\n", Protocol));
 
@@ -472,14 +473,15 @@ NTSTATUS FileOpenAddress(
       if (Address->Address[0].Address[0].sin_port)
       {
           /* The client specified an explicit port so we force a bind to this */
-          AddrFile->Port = TCPAllocatePort(Address->Address[0].Address[0].sin_port);
+          nPort = TCPAllocatePort(Address->Address[0].Address[0].sin_port);
 
           /* Check for bind success */
-          if (AddrFile->Port == 0xffff)
+          if (nPort == (UINT)-1)
           {
               ExFreePoolWithTag(AddrFile, ADDR_FILE_TAG);
               return STATUS_ADDRESS_ALREADY_EXISTS;
           }
+          AddrFile->Port = nPort;
 
           /* Sanity check */
           ASSERT(Address->Address[0].Address[0].sin_port == AddrFile->Port);
@@ -487,14 +489,15 @@ NTSTATUS FileOpenAddress(
       else if (!AddrIsUnspecified(&AddrFile->Address))
       {
           /* The client is trying to bind to a local address so allocate a port now too */
-          AddrFile->Port = TCPAllocatePort(0);
+          nPort = TCPAllocatePort(0);
 
           /* Check for bind success */
-          if (AddrFile->Port == 0xffff)
+          if (nPort == (UINT)-1)
           {
               ExFreePoolWithTag(AddrFile, ADDR_FILE_TAG);
               return STATUS_ADDRESS_ALREADY_EXISTS;
           }
+          AddrFile->Port = nPort;
       }
       else
       {
@@ -509,16 +512,16 @@ NTSTATUS FileOpenAddress(
 
   case IPPROTO_UDP:
       TI_DbgPrint(MID_TRACE,("Allocating udp port\n"));
-      AddrFile->Port =
-	  UDPAllocatePort(Address->Address[0].Address[0].sin_port);
+      nPort = UDPAllocatePort(Address->Address[0].Address[0].sin_port);
 
       if ((Address->Address[0].Address[0].sin_port &&
-           AddrFile->Port != Address->Address[0].Address[0].sin_port) ||
-           AddrFile->Port == 0xffff)
+           nPort != Address->Address[0].Address[0].sin_port) ||
+           nPort == (UINT)-1)
       {
           ExFreePoolWithTag(AddrFile, ADDR_FILE_TAG);
           return STATUS_ADDRESS_ALREADY_EXISTS;
       }
+      AddrFile->Port = nPort;
 
       TI_DbgPrint(MID_TRACE,("Setting port %d (wanted %d)\n",
                              AddrFile->Port,
--- a/sdk/lib/drivers/ip/network/ports.c
+++ b/sdk/lib/drivers/ip/network/ports.c
@@ -39,6 +39,12 @@ VOID DeallocatePort( PPORT_SET PortSet, ULONG Port ) {
     ASSERT(Port >= PortSet->StartingPort);
     ASSERT(Port < PortSet->StartingPort + PortSet->PortsToOversee);
 
+    if ((Port < PortSet->StartingPort) ||
+        (Port >= PortSet->StartingPort + PortSet->PortsToOversee))
+    {
+       return;
+    }
+
     KeAcquireSpinLock( &PortSet->Lock, &OldIrql );
     RtlClearBits( &PortSet->ProtoBitmap, Port - PortSet->StartingPort, 1 );
     KeReleaseSpinLock( &PortSet->Lock, OldIrql );
--- a/sdk/lib/drivers/ip/transport/tcp/accept.c
+++ b/sdk/lib/drivers/ip/transport/tcp/accept.c
@@ -77,10 +77,10 @@ NTSTATUS TCPListen(PCONNECTION_ENDPOINT Connection, UINT Backlog)
             if (NT_SUCCESS(Status))
             {
                 /* Allocate the port in the port bitmap */
-                Connection->AddressFile->Port = TCPAllocatePort(LocalAddress.Address[0].Address[0].sin_port);
-
+                UINT nPort = TCPAllocatePort(LocalAddress.Address[0].Address[0].sin_port);
                 /* This should never fail */
-                ASSERT(Connection->AddressFile->Port != 0xFFFF);
+                ASSERT(nPort != (UINT)-1);
+                Connection->AddressFile->Port = nPort;
             }
         }
     }
--- a/sdk/lib/drivers/ip/transport/tcp/tcp.c
+++ b/sdk/lib/drivers/ip/transport/tcp/tcp.c
@@ -214,7 +214,7 @@ NTSTATUS TCPStartup(VOID)
 {
     NTSTATUS Status;
 
-    Status = PortsStartup( &TCPPorts, 1, 0xfffe );
+    Status = PortsStartup(&TCPPorts, 1, 0xffff);
     if (!NT_SUCCESS(Status))
     {
         return Status;
@@ -370,6 +370,8 @@ NTSTATUS TCPConnect
     /* Check if we had an unspecified port */
     if (!Connection->AddressFile->Port)
     {
+        UINT nPort;
+
         /* We did, so we need to copy back the port */
         Status = TCPGetSockAddress(Connection, (PTRANSPORT_ADDRESS)&LocalAddress, FALSE);
         if (!NT_SUCCESS(Status))
@@ -379,10 +381,10 @@ NTSTATUS TCPConnect
         }
 
         /* Allocate the port in the port bitmap */
-        Connection->AddressFile->Port = TCPAllocatePort(LocalAddress.Address[0].Address[0].sin_port);
-
+        nPort = TCPAllocatePort(LocalAddress.Address[0].Address[0].sin_port);
         /* This should never fail */
-        ASSERT(Connection->AddressFile->Port != 0xFFFF);
+        ASSERT(nPort != (UINT)-1);
+        Connection->AddressFile->Port = nPort;
     }
 
     connaddr.addr = RemoteAddress.Address.IPv4Address;


sdk/lib/drivers/ip/network/ports.c
```diff
@@ -36,8 +36,11 @@ VOID DeallocatePort( PPORT_SET PortSet, ULONG Port ) {
     KIRQL OldIrql;
 
     Port = htons(Port);
-    ASSERT(Port >= PortSet->StartingPort);
-    ASSERT(Port < PortSet->StartingPort + PortSet->PortsToOversee);
```
learn-more, line 40: DeallocatePort can only be called after a port has been created?
In that case it's perfectly fine to assert on an invalid port, because that should never happen

sdk/lib/drivers/ip/transport/tcp/accept.c
```diff
@@ -77,10 +77,18 @@ NTSTATUS TCPListen(PCONNECTION_ENDPOINT Connection, UINT Backlog)
             if (NT_SUCCESS(Status))
             {
                 /* Allocate the port in the port bitmap */
-                Connection->AddressFile->Port = TCPAllocatePort(LocalAddress.Address[0].Address[0].sin_port);
-
-                /* This should never fail */
-                ASSERT(Connection->AddressFile->Port != 0xFFFF);
+                UINT nPort = TCPAllocatePort(LocalAddress.Address[0].Address[0].sin_port);
+
+                if (nPort == (UINT)-1)
```
learn-more, line 82: If this should never fail, then an assert on this would be a good idea?
Otherwise, please elaborate on why it's allowed to fail now.

sdk/lib/drivers/ip/transport/tcp/tcp.c
```diff
@@ -379,10 +379,18 @@ NTSTATUS TCPConnect
         }
 
         /* Allocate the port in the port bitmap */
-        Connection->AddressFile->Port = TCPAllocatePort(LocalAddress.Address[0].Address[0].sin_port);
+        UINT nPort = TCPAllocatePort(LocalAddress.Address[0].Address[0].sin_port);
 
-        /* This should never fail */
-        ASSERT(Connection->AddressFile->Port != 0xFFFF);
+        if (nPort == (UINT)-1)
```
learn-more, line 384: Same as above here, why is this suddenly OK now?

sdk/lib/drivers/ip/network/ports.c
KRosUser2: Yes that's true now that the proposed change fully allow full range to be supported up to 0xFFFF. @muthu90tech : please re-enable these 2 ASSERTs

sdk/lib/drivers/ip/transport/tcp/accept.c
KRosUser2: Now that we properly check against (UINT)-1 and no longer 0xFFFF, we could indeed put the ASSERT back. But then i'd strongly advise to re-run the test with the same app provided in the ticket.

sdk/lib/drivers/ip/transport/tcp/tcp.c
KRosUser2: Same answer as above.

KRosUser2:
All comments from @learn-more have been tested OK against the use case raising the ticket for this PR. New patch taking into account these remarks is available here : https://jira.reactos.org/secure/attachment/64805/64805_net2.patch
@muthu90tech : comments to be integrated accordingly, please !

learn-more:
@KRosUser2 you choose not to create a PR yourself, so please don't ask other people to do your bidding this way.

KRosUser2:
> @KRosUser2 you choose not to create a PR yourself, so please don't ask other people to do your bidding this way.

Happy new year too.

SergeGautherie:
> existing TCPAlloc failure in [FileOpenAddress](https://git.reactos.org/?p=reactos.git;a=blob;f=drivers/network/tcpip/tcpip/fileobjs.c#l475)

Shouldn't you fix all such broken checks too?

KRosUser2:
> > existing TCPAlloc failure in [FileOpenAddress](https://git.reactos.org/?p=reactos.git;a=blob;f=drivers/network/tcpip/tcpip/fileobjs.c#l475)
> 
> Shouldn't you fix all such broken checks too?

Yes, that's the same flaw. I'd suggest the same design pattern to be used @muthu90tech on top of the proposed patch from yesterday

KRosUser2:
For information i've been testing this patch locally for 12+ days with permanent network activity : works perfectly, no more ASSERT and no adverse effect (no loss of comm, no performance degradation).
image

learn-more (approved):

SergeGautherie:
> [learn-more](https://github.com/learn-more) approved

Really? What is the plan for the missing fixes?

KRosUser2:
> > [learn-more](https://github.com/learn-more) approved
> 
> Really? What is the plan for the missing fixes?

The current proposal "does the job" for the ASSERT issue faced in the scenario of the ticket.
This being done, your suggestion to extend the same pattern to FileOpenAddress is a good one and :
- either @muthu90tech is able to do it (but seems offline these days....)
- or could be added by the dev at merge time (marginally compliant with PR rules)
- or could be added by a complementary PR (but likely under-optimal....)

KRosUser2:
@muthu90tech : Request from @SergeGautherie is available and tested in the new patch attached to https://jira.reactos.org/browse/CORE-18371 (see fileobj.c diff)

muthu90tech:
> @muthu90tech : Request from @SergeGautherie is available and tested in the new patch attached to https://jira.reactos.org/browse/CORE-18371 (see fileobj.c diff)

Done, apologies for delay on this.

KRosUser2:
Thanks @muthu90tech 
@SergeGautherie : do you confirm it's now fully ok for you ?

drivers/network/tcpip/tcpip/fileobjs.c
```diff
@@ -472,29 +473,39 @@ NTSTATUS FileOpenAddress(
...
       {
           /* The client specified an explicit port so we force a bind to this */
-          AddrFile->Port = TCPAllocatePort(Address->Address[0].Address[0].sin_port);
+          nPort = TCPAllocatePort(Address->Address[0].Address[0].sin_port);
 
           /* Check for bind success */
-          if (AddrFile->Port == 0xffff)
+          if (nPort == (UINT)-1)
           {
+              AddrFile->Port = 0xffff;
```
learn-more, line 481: This is not needed, since AddrFile will be removed anyway

drivers/network/tcpip/tcpip/fileobjs.c
```diff
@@ -472,29 +473,39 @@ NTSTATUS FileOpenAddress(
...
       {
           /* The client is trying to bind to a local address so allocate a port now too */
-          AddrFile->Port = TCPAllocatePort(0);
+          nPort = TCPAllocatePort(0);
 
           /* Check for bind success */
-          if (AddrFile->Port == 0xffff)
+          if (nPort == (UINT)-1)
           {
+              AddrFile->Port = 0xffff;
```
learn-more, line 501: Same here, it does not make sense to set it.

drivers/network/tcpip/tcpip/fileobjs.c
```diff
@@ -472,29 +473,39 @@ NTSTATUS FileOpenAddress(
...
               ExFreePoolWithTag(AddrFile, ADDR_FILE_TAG);
               return STATUS_ADDRESS_ALREADY_EXISTS;
           }
+          else
```
SergeGautherie, line 485: `else` is redundant, after `return`.

SergeGautherie (commented):
> it's now fully ok

Obviously not! Especially, `UDPAllocatePort()` call needs the same fix.

KRosUser2:
@muthu90tech : new patch added to https://jira.reactos.org/browse/CORE-18371 covering : 
- Comment from @learn-more on unnecessary "Port" member update before free
- Comment from @SergeGautherie about unnecessary else condition supported by @JoachimHenze 
- Request from @SergeGautherie to extend the fix to the UDPAllocate Port

drivers/network/tcpip/tcpip/fileobjs.c
KRosUser2: OK, good point.

KRosUser2:
> > it's now fully ok
> 
> Obviously not! Especially, `UDPAllocatePort()` call needs the same fix.

patch updated with UDPAllocate

drivers/network/tcpip/tcpip/fileobjs.c
KRosUser2: Yes, right, my mistake (work fine, but useless)

drivers/network/tcpip/tcpip/fileobjs.c
KRosUser2: same.

learn-more:
> @muthu90tech : new patch added to https://jira.reactos.org/browse/CORE-18371 covering :
> 
>     * Comment from @learn-more on unnecessary "Port" member update before free
> 
>     * Comment from @SergeGautherie about unnecessary else condition supported by @JoachimHenze
> 
>     * Request from @SergeGautherie to extend the fix to the UDPAllocate Port

Please stop mentioning everyone, and using github as messaging platform...

drivers/network/tcpip/tcpip/fileobjs.c
```diff
@@ -509,16 +512,16 @@ NTSTATUS FileOpenAddress(
 
   case IPPROTO_UDP:
       TI_DbgPrint(MID_TRACE,("Allocating udp port\n"));
-      AddrFile->Port =
-	  UDPAllocatePort(Address->Address[0].Address[0].sin_port);
+      nPort = UDPAllocatePort(Address->Address[0].Address[0].sin_port);
 
       if ((Address->Address[0].Address[0].sin_port &&
            AddrFile->Port != Address->Address[0].Address[0].sin_port) ||
```
HBelusca, line 518: This should be a comparison with `nPort` now ?

sdk/lib/drivers/ip/transport/tcp/tcp.c
```diff
@@ -214,7 +214,7 @@ NTSTATUS TCPStartup(VOID)
 {
     NTSTATUS Status;
 
-    Status = PortsStartup( &TCPPorts, 1, 0xfffe );
+    Status = PortsStartup( &TCPPorts, 1, 0xffff );
```
HBelusca, line 217:
```suggestion
    Status = PortsStartup(&TCPPorts, 1, 0xffff);
```

sdk/lib/drivers/ip/transport/tcp/tcp.c
```diff
@@ -379,10 +379,10 @@ NTSTATUS TCPConnect
         }
 
         /* Allocate the port in the port bitmap */
-        Connection->AddressFile->Port = TCPAllocatePort(LocalAddress.Address[0].Address[0].sin_port);
-
+        UINT nPort = TCPAllocatePort(LocalAddress.Address[0].Address[0].sin_port);
```
HBelusca, line 382: Please move the `UINT nPort;` at the top of the `if (!Connection->...)` block.

drivers/network/tcpip/tcpip/fileobjs.c
KRosUser2: Yes, my mistake.

sdk/lib/drivers/ip/transport/tcp/tcp.c
KRosUser2: cosmetic, but OK

sdk/lib/drivers/ip/transport/tcp/tcp.c
KRosUser2: OK, will be done in next commit

drivers/network/tcpip/tcpip/fileobjs.c
```diff
@@ -472,29 +473,31 @@ NTSTATUS FileOpenAddress(
       if (Address->Address[0].Address[0].sin_port)
       {
           /* The client specified an explicit port so we force a bind to this */
-          AddrFile->Port = TCPAllocatePort(Address->Address[0].Address[0].sin_port);
+          nPort = TCPAllocatePort(Address->Address[0].Address[0].sin_port);
```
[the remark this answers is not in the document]
learn-more: no need for that.

sdk/lib/drivers/ip/transport/tcp/tcp.c
KRosUser2: Yes, see discussion with Mark in ticket's comment

drivers/network/tcpip/tcpip/fileobjs.c
```diff
@@ -472,29 +473,31 @@ NTSTATUS FileOpenAddress(
...
 
           /* Check for bind success */
-          if (AddrFile->Port == 0xffff)
+          if (nPort == (UINT)-1)
           {
               ExFreePoolWithTag(AddrFile, ADDR_FILE_TAG);
               return STATUS_ADDRESS_ALREADY_EXISTS;
           }
+          AddrFile->Port = nPort;
```
SergeGautherie, line 484:
```suggestion

          AddrFile->Port = nPort;
```

drivers/network/tcpip/tcpip/fileobjs.c
```diff
@@ -472,29 +473,31 @@ NTSTATUS FileOpenAddress(
...
 
           /* Sanity check */
           ASSERT(Address->Address[0].Address[0].sin_port == AddrFile->Port);
```
SergeGautherie, line 487: Why is this check needed? Why is it not identical everywhere: why sometimes `if()` or missing?
If actually wanted, why not do it in `*PAllocatePort()` themselves instead?

drivers/network/tcpip/tcpip/fileobjs.c
```diff
@@ -509,16 +512,16 @@ NTSTATUS FileOpenAddress(
...
-	  UDPAllocatePort(Address->Address[0].Address[0].sin_port);
+      nPort = UDPAllocatePort(Address->Address[0].Address[0].sin_port);
 
       if ((Address->Address[0].Address[0].sin_port &&
-           AddrFile->Port != Address->Address[0].Address[0].sin_port) ||
-           AddrFile->Port == 0xffff)
+           nPort != Address->Address[0].Address[0].sin_port) ||
+           nPort == (UINT)-1)
       {
           ExFreePoolWithTag(AddrFile, ADDR_FILE_TAG);
```
SergeGautherie, line 521: No port cleanup?

sdk/lib/drivers/ip/network/ports.c
```diff
@@ -39,6 +39,12 @@ VOID DeallocatePort( PPORT_SET PortSet, ULONG Port ) {
     ASSERT(Port >= PortSet->StartingPort);
     ASSERT(Port < PortSet->StartingPort + PortSet->PortsToOversee);
```
SergeGautherie, line 40: Why assert when deallocating but not when allocating?

drivers/network/tcpip/tcpip/fileobjs.c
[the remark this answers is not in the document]
HBelusca: Suggestion doesn't make sense. The idea was to store the port in AddrFile only when it has been validated.

sdk/lib/drivers/ip/network/ports.c
HBelusca: Isn't it because the TCPAllocatePort function already ensures that? While here the assert is part of checking whether the caller of the DeallocatePort gives a valid Port value?

sdk/lib/drivers/ip/transport/tcp/tcp.c
JoachimHenze: Resolved.

drivers/network/tcpip/tcpip/fileobjs.c
JoachimHenze: Resolved.

drivers/network/tcpip/tcpip/fileobjs.c
KRosUser2: Why would it be needed here ?

drivers/network/tcpip/tcpip/fileobjs.c
KRosUser2: Why ?

ericsomto (approved):

KRosUser2:
The last comments Serge seem to be general evolution fixes in the Port management logics going beyond the scope of this PR (fixing the ASSERT, allowing use of the full range of ports). Should this be fixed here or in a separate PR to come afterwards (to reduce the risk of regression and keep a good atomicity ?)

sdk/lib/drivers/ip/network/ports.c
SergeGautherie: > ... TCPAllocatePort ... DeallocatePort ...

I don't think your "questions" explain why alloc and dealloc (should continue to) behave differently with regard to input validation(s).

SergeGautherie (commented):
> scope of this PR

I think actual scope is to fix the involved functions and their callers, ideally without the need for a follow-up.

sdk/lib/drivers/ip/network/ports.c
```diff
@@ -39,6 +39,12 @@ VOID DeallocatePort( PPORT_SET PortSet, ULONG Port ) {
...
 
+    if ((Port < PortSet->StartingPort) ||
+        (Port >= PortSet->StartingPort + PortSet->PortsToOversee))
```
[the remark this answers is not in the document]
KRosUser2: Right.

drivers/network/tcpip/tcpip/fileobjs.c
```diff
@@ -404,6 +404,7 @@ NTSTATUS FileOpenAddress(
   PVOID Options)
 {
   PADDRESS_FILE AddrFile;
+  UINT nPort;
```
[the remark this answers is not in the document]
KRosUser2: OK

KRosUser2:
The patch in https://jira.reactos.org/secure/attachment/64907/net.patch has been updated with all latest changes requested by Thomas.
@muthu90tech : Would you please commit here ?

KRosUser2:
This PR will likely fix https://jira.reactos.org/browse/CORE-18806 too.

KRosUser2:
@muthu90tech : Please note in the description that your PR will also fix https://jira.reactos.org/browse/CORE-18806 if it's someday merged.

KRosUser2:
@muthu90tech : please note this will contribute to this ticket too : https://jira.reactos.org/browse/CORE-12024?focusedCommentId=138138&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-138138

learn-more (changes requested):
Please fix the issues indicated by Thomas.

KRosUser2:
This PR needs to be rebased following https://github.com/reactos/reactos/pull/5025
Net.Patch in https://jira.reactos.org/browse/CORE-18371 has been updated accordingly to reflect the needed change.

zecarlos1957:
@KRosUser2 It's one of the hardest things, when it's not even impossible to make everyone happy

KRosUser2:
> @KRosUser2 It's one of the hardest things, when it's not really impossible to make everyone happy.

This fix is quite easy indeed

Doug-Lyons:
Since there has been no update from the creator of this PR in over a month, I would like to pick it up and create a new PR because of the inactivity. I intend to create a new PR shortly and will reference it here after it is created. I have a strong interest in the TCP/IP functionality of ReactOS and this seems to be a move in the right direction. Thanks.

gonzoMD:
This is superseeded by #5074

PR #4969, opened 2022-12-31
[NTOS:CM] Cleanup the hive in case linking it to master fails

GeoB99:
Currently the failure code path doesn't do any kind of cleanup against the hive that was being linked to master. The cleanup is pretty straightforward as you just simply close the hive file handles and free the registry kernel structures.

**JIRA Issues:** [CORE-5772](https://jira.reactos.org/browse/CORE-5772), [CORE-13559](https://jira.reactos.org/browse/CORE-13559), [CORE-17263](https://jira.reactos.org/browse/CORE-17263)

### DIFF
--- a/ntoskrnl/config/cmapi.c
+++ b/ntoskrnl/config/cmapi.c
@@ -2121,17 +2121,23 @@ CmLoadKey(IN POBJECT_ATTRIBUTES TargetKey,
         /* Release the hive */
         CmHive->Hive.HiveFlags &= ~HIVE_IS_UNLOADING;
         CmHive->CreatorOwner = NULL;
-
-        /* Allow loads */
-        ExReleasePushLock(&CmpLoadHiveLock);
     }
     else
     {
         DPRINT1("CmpLinkHiveToMaster failed, Status %lx\n", Status);
-        /* FIXME: TODO */
-        // ASSERT(FALSE); see CORE-17263
-        ExReleasePushLock(&CmpLoadHiveLock);
+
+        /* We're touching this hive, set the loading flag */
+        CmHive->HiveIsLoading = TRUE;
+
+        /* Close associated file handles */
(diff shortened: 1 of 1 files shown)


ntoskrnl/config/cmapi.c
```diff
@@ -2128,8 +2128,12 @@ CmLoadKey(IN POBJECT_ATTRIBUTES TargetKey,
...
         DPRINT1("CmpLinkHiveToMaster failed, Status %lx\n", Status);
-        /* FIXME: TODO */
-        // ASSERT(FALSE); see CORE-17263
+
+        /* Close file handles of this hive */
+        CmpCloseHiveFiles(CmHive);
+
+        /* Perform a cleanup against this hive */
+        CmpDestroyHive(CmHive);
         ExReleasePushLock(&CmpLoadHiveLock);
```
SergeGautherie, line 2137: While here, could merge identical line in both cases.

HBelusca (commented):
Looks OK, but this needs a test, besides testing whether CORE-17263 is fixed with this (ask @Extravert-ir for how to test that new storage stack).
The test would be, to manually fail that `CmpLinkHiveToMaster` call in that code above (e.g. take unconditionally that newly-implemented failure branch, or, return an error status instead of success in the CmpLinkHiveToMaster implementation), and test with windbg whether that new cleanup path works correctly and doesn't trigger further errors/asserts.

HBelusca (approved):
Testing the failure code path seems to work as expected (offline test with George).

ntoskrnl/config/cmapi.c
Extravert-ir: I'd say this is a legit comment. The lock is acquired outside the condition, and should be released outside it too.

HBelusca (approved):

PR #4972, opened 2023-01-01
[AFD] ReceiveActivity: Don't return STATUS_FILE_CLOSED in case of FCB overread

TAN-Gaming:
## Purpose
Fix FileZilla 3.8 `ECONNRESET` error (based on patch by [KRosUser](https://jira.reactos.org/secure/ViewProfile.jspa?name=KRosUser))

JIRA issue: [CORE-18328](https://jira.reactos.org/browse/CORE-18328)

### DIFF
--- a/drivers/network/afd/afd/read.c
+++ b/drivers/network/afd/afd/read.c
@@ -175,16 +175,8 @@ static NTSTATUS ReceiveActivity( PAFD_FCB FCB, PIRP Irp ) {
                                     TotalBytesCopied));
             UnlockBuffers( RecvReq->BufferArray,
                            RecvReq->BufferCount, FALSE );
-            if (FCB->Overread && FCB->LastReceiveStatus == STATUS_SUCCESS)
-            {
-                /* Overread after a graceful disconnect */
-                AFD_DbgPrint(MID_TRACE,("FCB overread detected\n"));
-            }
 
-            /* Overread shall not lead to situation of "ECONNRESET" errors
-             * See CORE-18328 and CORE-11650 */
-
-            /* Unexpected disconnect by the remote host or initial read after a graceful disconnect */
+            /* Unexpected disconnect by the remote host or graceful disconnect */
             Status = FCB->LastReceiveStatus;
 
             NextIrp->IoStatus.Status = Status;
@@ -193,7 +185,6 @@ static NTSTATUS ReceiveActivity( PAFD_FCB FCB, PIRP Irp ) {
             if( NextIrp->MdlAddress ) UnlockRequest( NextIrp, IoGetCurrentIrpStackLocation( NextIrp ) );
             (void)IoSetCancelRoutine(NextIrp, NULL);
             IoCompleteRequest( NextIrp, IO_NETWORK_INCREMENT );
-            FCB->Overread = TRUE;
         }
     } else {
         /* Kick the user that receive would be possible now */
--- a/drivers/network/afd/afd/read.c
+++ b/drivers/network/afd/afd/read.c
@@ -177,14 +177,16 @@ static NTSTATUS ReceiveActivity( PAFD_FCB FCB, PIRP Irp ) {
                            RecvReq->BufferCount, FALSE );
             if (FCB->Overread && FCB->LastReceiveStatus == STATUS_SUCCESS)
             {
-                /* Overread after a graceful disconnect so complete with an error */
-                Status = STATUS_FILE_CLOSED;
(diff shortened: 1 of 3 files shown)


KRosUser2:
Lgtm, thanks !

KRosUser2:
Tested 20+ times with Filezilla on a VM running this patch locally for 12+days with continuous network activity : 
- Problem is correctly fixed (see below)
- No adverse effect on other network related functions (browsing, SMTP, FTP, HTTP)
image

KRosUser2 (approved):

ThFabba:
Do we have anything in our test suite that confirms that this change is correct? If not, it would be helpful to see an added test that we can refer back to to avoid regressions.

KRosUser2:
In fact the question is :
- Is it appropriate to "force" the status to STATUS_FILE_CLOSED in case of OverRead when the FCB returns a succesful status ?
- Is it legit to raise an "Overread" situation in this case ?

This patch fixes the 1st point and makes FileZilla operative in ReactOS.

Regarding the test you need, I undestand your point as "we need test in testman" to check against possible future regressions, is it correct ?
@TAN-Gaming : Do you know how to do this ? (I don't)

KRosUser2:
This may be a hack, but at least it enables FileZilla client in ReactOS and logs an ERR which would allow the "full fix" later if it's not.

ThFabba:
Let's get a testman comparison of before and after please. Has anyone done a buildbot run against this?

KRosUser2:
@ThFabba : results posted here : https://jira.reactos.org/browse/CORE-18328?focusedId=138889&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-138889 by @Doug-Lyons

drivers/network/afd/afd/read.c
```diff
@@ -177,14 +177,16 @@ static NTSTATUS ReceiveActivity( PAFD_FCB FCB, PIRP Irp ) {
                            RecvReq->BufferCount, FALSE );
             if (FCB->Overread && FCB->LastReceiveStatus == STATUS_SUCCESS)
```
ThFabba, line 178: This is the only use of the `Overread` member. Please remove the member entirely.

ThFabba, line 178: The new test in `ws2_32:recv` confirms that this special behavior is never needed, and this is the only use of the `Overread` member. Please remove it entirely.

drivers/network/afd/afd/read.c
```diff
@@ -177,14 +177,16 @@ static NTSTATUS ReceiveActivity( PAFD_FCB FCB, PIRP Irp ) {
...
-            else
-            {
-                /* Unexpected disconnect by the remote host or initial read after a graceful disconnect */
-                Status = FCB->LastReceiveStatus;
+                /* Overread after a graceful disconnect */
+                AFD_DbgPrint(MID_TRACE,("FCB overread detected\n"));
             }
+
+            /* Overread shall not lead to situation of "ECONNRESET" errors
+             * See CORE-18328 and CORE-11650 */
```
ThFabba, line 185: No need for this comment, it's backed by tests (and we won't have to track "Overread" at all)

drivers/network/afd/afd/read.c
```diff
@@ -177,14 +177,16 @@ static NTSTATUS ReceiveActivity( PAFD_FCB FCB, PIRP Irp ) {
...
+
+            /* Unexpected disconnect by the remote host or initial read after a graceful disconnect */
```
ThFabba, line 187: The `initial read after a` part is now superfluous -- graceful disconnect always returns success as shown by the test.

KRosUser2:
Tested OK :
image

atharva1910:
I agree with @KRosUser2, forcing a file close status seems strange.

I see you have also deleted the Overread flag. I have a question: 
Do we not care about over reading? If we are not returning an error does that not mean the client might potentially continue to read after EOF (the check above checks for EOF)? Would it not be apt to return a less harsh status? One which keeps does not force bugzilla (or any other application) to reset the connection?

KRosUser2:
@ThFabba : Do you need anything more before approval ?

atharva1910:
LGTM

atharva1910 (approved):

JoachimHenze (approved):

Doug-Lyons:
Thanks @GeoB99 for the merge.

KRosUser2:
Thanks @GeoB99 and @TAN-Gaming for this PR and its merge !

PR #4973, opened 2023-01-01
[RAPPS] Fix Installed Apps information display

rexxDigital:
## Purpose

Trivial fix to installed apps information so it displays properly

JIRA issue: [CORE-18397](https://jira.reactos.org/browse/CORE-18397) [CORE-18774](https://jira.reactos.org/browse/CORE-18774)

## Proposed changes

- ReleaseBuffer()

### DIFF
--- a/base/applications/rapps/installed.cpp
+++ b/base/applications/rapps/installed.cpp
@@ -88,7 +88,7 @@ void CInstalledApplicationInfo::EnsureDetailsLoaded()
         if (GetApplicationRegDword(L"WindowsInstaller", &dwWindowsInstaller) && dwWindowsInstaller)
         {
             // MSI has the same info in Uninstall / modify, so manually build it
-            szUninstallString.Format(L"msiexec /x %s", m_szKeyName.GetString());
+            szUninstallString.Format(L"msiexec /x%s", m_szKeyName.GetString());
         }
         else
         {
@@ -111,7 +111,7 @@ void CInstalledApplicationInfo::EnsureDetailsLoaded()
         {
             if (dwWindowsInstaller)
             {
-                szModifyPath.Format(L"msiexec /i %s", m_szKeyName.GetString());
+                szModifyPath.Format(L"msiexec /i%s", m_szKeyName.GetString());
             }
             else
             {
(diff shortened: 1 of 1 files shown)


KRosUser2:
Lgtm, thanks

base/applications/rapps/installed.cpp
```diff
@@ -88,7 +88,7 @@ void CInstalledApplicationInfo::EnsureDetailsLoaded()
         if (GetApplicationRegDword(L"WindowsInstaller", &dwWindowsInstaller) && dwWindowsInstaller)
         {
             // MSI has the same info in Uninstall / modify, so manually build it
-            szUninstallString.Format(L"msiexec /x%s", m_szKeyName.GetString());
+            szUninstallString.Format(L"msiexec /x %s", m_szKeyName.GetString());
```
learn-more, line 91: There is a special case in our (and wine's) code to handle exactly this case,
were any negative effects observed from the two options being concatenated?
https://git.reactos.org/?p=reactos.git;a=blob;f=base/system/msiexec/msiexec.c;hb=95e5f07084088129e3c121568ea06c7f4d0a8d41#l257

KRosUser2: Sorry I missed this specific case being properly managed by ReactOS/Wine. 

Do you then consider this change is un-nnecesasry and to be removed from this PR (and associated ticket closed as Won't do) ? Or is the change "cleaner" (but only cosmetic) ?

learn-more: This looks like a change for the sake of changing it, hence my original question in [CORE-18774](https://jira.reactos.org/browse/CORE-18774):
> And this is a problem because?

KRosUser2: This is not a problem and I didn't knew ReactOS specifically handle this case. Not sure Windows does but likely not a sufficient reason for change. 

In this case I propose to trim this PR to the Hyperlink fix only

@rexxDigital : ok for you ?

rexxDigital: Yeah thats ok with me

learn-more: This has worked fine for years as far as I know, and I have used rapps both on windows and on ros to test stuff.

JoachimHenze:
The decription is not up to date anymore / yet.

learn-more (approved):

PR #4975, opened 2023-01-03
[RAPPS] Add tooltip for "Refresh" and "Update Database"

TAN-Gaming:
## Purposed changes
- Add tooltip for "Refresh" and "Update Database" in the main toolbar

VirtualBox_ReactOS_DBG_03_01_2023_11_09_18

VirtualBox_ReactOS_DBG_03_01_2023_11_09_28

### DIFF
--- a/base/applications/rapps/include/appview.h
+++ b/base/applications/rapps/include/appview.h
@@ -285,6 +285,8 @@ class CMainToolbar :
     WCHAR szUninstallBtn[MAX_STR_LEN];
     WCHAR szModifyBtn[MAX_STR_LEN];
     WCHAR szSelectAll[MAX_STR_LEN];
+    WCHAR szRefreshBtn[MAX_STR_LEN];
+    WCHAR szUpdateDbBtn[MAX_STR_LEN];
 
     VOID AddImageToImageList(HIMAGELIST hImageList, UINT ImageIndex);
 
--- a/base/applications/rapps/appview.cpp
+++ b/base/applications/rapps/appview.cpp
@@ -66,6 +66,8 @@ CMainToolbar::CMainToolbar()
     memset(szUninstallBtn, 0, sizeof(szUninstallBtn));
     memset(szModifyBtn, 0, sizeof(szModifyBtn));
     memset(szSelectAll, 0, sizeof(szSelectAll));
+    memset(szRefreshBtn, 0, sizeof(szRefreshBtn));
+    memset(szUpdateDbBtn, 0, sizeof(szUpdateDbBtn));
 }
 
 VOID CMainToolbar::OnGetDispInfo(LPTOOLTIPTEXT lpttt)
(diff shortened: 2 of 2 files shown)


base/applications/rapps/include/appview.h
```diff
@@ -285,6 +285,8 @@ class CMainToolbar :
     WCHAR szUninstallBtn[MAX_STR_LEN];
     WCHAR szModifyBtn[MAX_STR_LEN];
     WCHAR szSelectAll[MAX_STR_LEN];
+    WCHAR szRefreshBtn[MAX_STR_LEN];
+    WCHAR szUpdateDbBtn[MAX_STR_LEN];
```
learn-more, line 289: Ideally those should be CString's, but that would require some refactoring, so continuing on as is would be good enough for now.

learn-more (approved):

base/applications/rapps/include/appview.h
GGLinnk: Since ``SendMessageW`` (which is used by ``AddButtons``) requires the type to be ``WCHAR``, it would require additional non-optimized conversions.
When dealing with the ~~Windows~~ ReactOS API, it's always better to use embedded types like ``WCHAR``.

An actual good idea would be to migrate to ``TCHAR`` and ``SendMessage`` macros for better maintainability and compatibility.
But I think at this stage it's not required.

HBelusca: > migrate to TCHAR and SendMessage macros for better maintainability and compatibility.

Better maintainability and compatibility with what?

HBelusca (approved):

base/applications/rapps/include/appview.h
GGLinnk: > Better maintainability and compatibility with what?

Let's imagine that in the future we need to change the compiler settings to another text encoding system than UNICODE (which is unlikely in the near future but not impossible).
As a new standard that will say that the new standard is "NEWTYPECODE", we technically do not need to refactor the code for all applications. The only thing that will need to be changed will be the libraries like rosctrls.h.

HBelusca: This may be applicable only in like 200 years from now or whatnot. Doesn't make sense at all.

learn-more: > Since `SendMessageW` (which is used by `AddButtons`) requires the type to be `WCHAR`, it would require additional non-optimized conversions. When dealing with the ~Windows~ ReactOS API, it's always better to use embedded types like `WCHAR`.
> 
> An actual good idea would be to migrate to `TCHAR` and `SendMessage` macros for better maintainability and compatibility. But I think at this stage it's not required.

Sorry but I have to disagree with almost everything you said.
In cases like this, where there are a ton of details, and you are not in a hot code path, it's way better to go for the safety of something like a CString, instead of fiddling around with manually loading resource strings.
Besides, it's not like there is a huge overhead in those CStrings, other than the (usually single) allocation done for them.

As for using TCHAR: that might have been sound advice in 1998, but there is no good reason at all nowadays to use it.
Internally (almost) everything in ReactOS or Windows is using Unicode (or UTF-16)

GGLinnk: I've just checked and I did make a mistake.
I confused ATL/MFC CString with C++ std::string which I thought you were talking about.
So you are right.

ATL/MFC CString is actually very similar to TCHAR, WCHAR and CHAR.
CStringW would be similar to ``wchar_t`` (``WCHAR``).
CStringA would be similar to ``char`` (``CHAR``).
And finally CString would be like the ``TCHAR'' macro.

The real difference is that it requires the use of an additional library, which is not a big deal...

My bad, sorry :/

GGLinnk (approved):

PR #4977, opened 2023-01-04
[CMAKE] Do not use the dwarf baseaddress file for GCC Release builds

JoachimHenze:
Explanation:
There are currently two seperate cases where we do use the baseaddress_**dwarf**.cmake file:
1.) When configuring a GCC dbg build with -DNO_ROSSYM:BOOL=TRUE
2.) When configuring a GCC rls build with -DCMAKE_BUILD_TYPE=Release

The first one makes sense, large binaries will be the result, and therefore the standard baseaddress.cmake file is not giving sufficient space for those binaries.
But for the 2nd case it does not make sense to use the dwarf baseaddress file. The generated binaries can be expected to be even smaller than the regular debug buillds with RSYM. The dwarf baseaddress file being used here as well is an unintended side effect of the ancient commit SVN r68147 == git 74c80a51dd88d30d346928c04d98382319131507 

This is a late addendum to SVN r68147 == git 74c80a51dd88d30d346928c04d98382319131507

Back then we set NO_ROSSYM to TRUE for the release builds, that did lead us to unintentionally always using the baseaddress_dwarf.cmake file for our release builds, which is not optimal. Neither from the perspective of efficiency (the base addresses are too sparse for them, since dwarf generates much bigger binaries) nor from a testing perspective: We want the base addresses used for our releases to be tested frequently and not only very rarely. So it is good to use then what is frequently tested. And that are the addresses from the 'regular' debug build.

As discussed with @ThFabba and others in chat https://chat.reactos.org/reactos/pl/f1nafrgagjb1tm8a8ihzow6t9r

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -335,7 +335,7 @@ Enable this if the module uses typeid or dynamic_cast. You will probably need to
     add_subdirectory(sdk/include/reactos/mc)
     add_subdirectory(sdk/include/asm)
 
-    if(NO_ROSSYM)
+    if(DEFINED NO_ROSSYM AND NOT CMAKE_BUILD_TYPE STREQUAL "Release")
         include(sdk/cmake/baseaddress_dwarf.cmake)
     elseif(MSVC)
         if (ARCH STREQUAL "amd64")
--- a/sdk/cmake/gcc.cmake
+++ b/sdk/cmake/gcc.cmake
@@ -220,8 +220,12 @@ if(SEPARATE_DBG)
         "${CMAKE_STRIP} --only-keep-debug <TARGET> -o ${REACTOS_BINARY_DIR}/symbols/${SYMBOL_FILE}"
         ${strip_debug})
 elseif(NO_ROSSYM)
-    # Dwarf-based build
-    message(STATUS "Generating a dwarf-based build (no rsym)")
+    # Either a GCC Dwarf-based build, or a GCC build in release config
+    if(CMAKE_BUILD_TYPE STREQUAL "Release")
+        message(STATUS "Generating a Release build (no rsym)")
+    else()
+        message(STATUS "Generating a dwarf-based build (no rsym)")
+    endif()
     set(CMAKE_C_LINK_EXECUTABLE "<CMAKE_C_COMPILER> ${CMAKE_C_FLAGS} <CMAKE_C_LINK_FLAGS> <LINK_FLAGS> <OBJECTS> -o <TARGET> <LINK_LIBRARIES>")
     set(CMAKE_CXX_LINK_EXECUTABLE "<CMAKE_CXX_COMPILER> ${CMAKE_CXX_FLAGS} <CMAKE_CXX_LINK_FLAGS> <LINK_FLAGS> <OBJECTS> -o <TARGET> <LINK_LIBRARIES>")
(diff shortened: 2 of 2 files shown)


JoachimHenze:
You might be wondering why I came up with that "fancy" solution and did not simply revert the guilty rev... As that would be simpler. And even better: It would make the GCC Release builds binaries much smaller (~50% of the size then only!!!)

I tried that as well before, and it would also solve the wrong base address file being used, but it makes the rsym.exe process crash in the middle of the compilation for a GCC release build then:
[2955/12901] Linking CXX executable base\applications\network\ftp\ftp.exe
FAILED: base/applications/network/ftp/ftp.exe
cmd.exe /C "cd . && C:\RosBE2_1_6\i386\bin\g++.exe   -D__RELFILE__="&__FILE__[__FILE__[0] == '.' ? sizeof \"../0414rls/reactos\" - 1 : sizeof REACTOS_SOURCE_DIR
]" -pipe -fms-extensions -fno-strict-aliasing -nostdinc -mstackrealign -fno-rtti -fno-exceptions -fdebug-prefix-map="C:\0414rls\reactos"=ReactOS -march=pentium
-mtune=i686 -Wall -Wpointer-arith -Wno-char-subscripts -Wno-multichar -Wno-unused-value -Wno-unused-local-typedefs -Wno-deprecated -Wno-maybe-uninitialized -O2
-DNDEBUG -fno-optimize-sibling-calls -fno-omit-frame-pointer -mpreferred-stack-boundary=3 -fno-set-stack-executable -momit-leaf-frame-pointer  -nostdlib -Wl,--e
nable-auto-image-base,--disable-auto-import,--disable-stdcall-fixup,--gc-sections    -Wl,--subsystem,console:5.01 -Wl,--major-image-version,5 -Wl,--minor-image-
version,01 -Wl,--major-os-version,5 -Wl,--minor-os-version,01 -Wl,-entry,_mainCRTStartup base/applications/network/ftp/CMakeFiles/ftp.dir/cmds.c.obj base/applic
ations/network/ftp/CMakeFiles/ftp.dir/cmdtab.c.obj base/applications/network/ftp/CMakeFiles/ftp.dir/domacro.c.obj base/applications/network/ftp/CMakeFiles/ftp.d
ir/fake.c.obj base/applications/network/ftp/CMakeFiles/ftp.dir/ftp.c.obj base/applications/network/ftp/CMakeFiles/ftp.dir/main.c.obj base/applications/network/f
tp/CMakeFiles/ftp.dir/ruserpass.c.obj base/applications/network/ftp/CMakeFiles/ftp.dir/ftp.rc.obj -o base\applications\network\ftp\ftp.exe  dll/win32/ws2_32/lib
ws2_32.a dll/win32/iphlpapi/libiphlpapi.a sdk/lib/crt/libmsvcrtex.a dll/win32/msvcrt/libmsvcrt.a dll/win32/kernel32/libkernel32.a sdk/lib/crt/oldnames.a -lgcc &
& C:/0414GCCrlsRevert/host-tools/rsym.exe -s C:/0414rls/reactos base\applications\network\ftp\ftp.exe base\applications\network\ftp\ftp.exe && cd ."
ninja: build stopped: subcommand failed.

Maybe it would be an even better idea to find out what is causing that crash and then simply revert the guilty rev!

PR #4978, opened 2023-01-04
Fix SEH in Cc

tkreuzer:
## Purpose

This PR fixes broken usage of _SEH2_FINALLY.
1. While variable assignment is handled properly with MSVC, GCC/PSEH does not know anything about exceptions and does not guarantee that your variables are set where you think they are set, because the compiler will happily remove any assignments that it considers redundant or move them to a different place. If you need to check a variable within a finally or except block, which is set in a try block, you must mark that variable with _SEH2_VOLATILE.
2. Finally handlers are - unlike except blocks - not part of the function they are in, but separate functions, which are called during unwind. PSEH implements them on GCC using nested functions. While "return" from a finally handler is allowed with native SEH, it's handled by the compiler through an extra unwinding operation using _local_unwind, WHICH IS NOT SUPPORTED BY PSEH! With PSEH, returning from a finally handler does not return from the function, instead it will only return from the finally handler and the function will continue below the finally handler as if there was no return at all. To work around this issue, _SEH2_EXCEPT can be used together with a variable that is set in the try and the except block and checked after it.
3. Since there is no point in raising an exception from CcRosEnsureVacbResident in combination with returning a BOOLEAN result, and it simply makes everything more complicated and prone to errors, remove all SEH entirely and simply use the returned BOOLEAN value.

Also change PSEH to make return from finally handler a compile error.

## TODO

- [x] Run on testbot; https://reactos.org/testman/compare.php?ids=85480,85482,85484,85486,85497 (rightmost is this PR)

### DIFF
--- a/ntoskrnl/cc/pin.c
+++ b/ntoskrnl/cc/pin.c
@@ -221,7 +221,7 @@ CcpPinData(
     KIRQL OldIrql;
     ULONG VacbOffset;
     NTSTATUS Status;
-    BOOLEAN Result;
+    _SEH2_VOLATILE BOOLEAN Result;
 
     VacbOffset = (ULONG)(FileOffset->QuadPart % VACB_MAPPING_GRANULARITY);
 
@@ -302,15 +302,19 @@ CcpPinData(
             CCTRACE(CC_API_DEBUG, "FileObject=%p FileOffset=%p Length=%lu Flags=0x%lx -> FALSE\n",
                             SharedCacheMap->FileObject, FileOffset, Length, Flags);
             CcUnpinData(&NewBcb->PFCB);
-            return FALSE;
+            *Bcb = NULL;
+            *Buffer = NULL;
         }
     }
     _SEH2_END;
 
-    *Bcb = &NewBcb->PFCB;
-    *Buffer = (PVOID)((ULONG_PTR)NewBcb->Vacb->BaseAddress + VacbOffset);
+    if (Result)
+    {
+        *Bcb = &NewBcb->PFCB;
+        *Buffer = (PVOID)((ULONG_PTR)NewBcb->Vacb->BaseAddress + VacbOffset);
+    }
 
-    return TRUE;
+    return Result;
 }
 
 /*
@@ -332,7 +336,7 @@ CcMapData (
     PROS_SHARED_CACHE_MAP SharedCacheMap;
     ULONG VacbOffset;
     NTSTATUS Status;
-    BOOLEAN Result;
+    _SEH2_VOLATILE BOOLEAN Result;
 
     CCTRACE(CC_API_DEBUG, "CcMapData(FileObject 0x%p, FileOffset 0x%I64x, Length %lu, Flags 0x%lx,"
            " pBcb 0x%p, pBuffer 0x%p)\n", FileObject, FileOffset->QuadPart,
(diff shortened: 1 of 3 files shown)


ntoskrnl/cc/pin.c
```diff
@@ -286,26 +286,19 @@ CcpPinData(
...
     }
 
-    Result = FALSE;
-    _SEH2_TRY
+    /* Ensure the pages are resident */
+    Result = CcRosEnsureVacbResident(NewBcb->Vacb,
+                                     BooleanFlagOn(Flags, PIN_WAIT),
+                                     BooleanFlagOn(Flags, PIN_NO_READ),
+                                     VacbOffset,
+                                     Length);
```
GeoB99, line 294: While this function is no longer wrapped within a SEH block, it actually raises a status exception. I don't know exactly how dramatic the situation can be if a VACB couldn't be set as resident so perhaps removing `ExRaiseStatus` in that function should be more than enough?

tkreuzer: Yes it should just return TRUE/FALSE. Fixed that.

zefklop:
`Since there is no point in raising an exception from CcRosEnsureVacbResident in combination with returning a BOOLEAN result, and it simply makes everything more complicated and prone to errors, remove all SEH entirely and simply use the returned BOOLEAN value.`
Well, yes. But Cc is designed in such a way that callers of those functions get a NTSTATUS by wrapping the calls in try/except blocks. So raising here is actually required. I know it's lame, but it is what it is.
Finally handlers were here to perform cleanup in case this happened. I'm fine if raising is done in the Cc API rather than in the worker function, but please keep the raise!

GeoB99:
@zefklop Except that your `CcRosEnsureVacbResident` isn't properly handled in all cases, e.g. https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/cc/copy.c;hb=84e32e4e90109b3bce75ddb3eee2d3dfc42ab4be#l209 and https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/cc/copy.c;hb=84e32e4e90109b3bce75ddb3eee2d3dfc42ab4be#l237, supposedly `CcRosEnsureVacbResident` raises an exception these linked code paths won't be able to handle it. And the worst case of all is that this code (among others in Cc unfortunately) is that it's not even documented so one has to resort to guesses when trying to understand what's going on.

If you have something to talk about, please use the #kernel-dev channel on Mattermost and we'll continue our discussion from there so we don't pollute this PR.

zefklop:
@GeoB99 you're right, this function and its use are sloppy,  and there is something to fix. I don't see how commenting on the PR changes pollutes the said PR though.

tkreuzer:
> Well, yes. But Cc is designed in such a way that callers of those functions get a NTSTATUS by wrapping the calls in try/except blocks. So raising here is actually required. I know it's lame, but it is what it is. Finally handlers were here to perform cleanup in case this happened. I'm fine if raising is done in the Cc API rather than in the worker function, but please keep the raise!

I understand that Cc throws exceptions for certain errors, but that is how the external/exported functions work, it is IMO a very bad design decision, and IMO we should avoid this mechanism internally to avoid problems like in the 2 calls that are not wrapped in SEH (and additionally treat the returned value as an NTSTATUS, but that is a different thing.)
Anyway, I brought back the Raise and fixed the way _SEH_FINALLY was used. I also fixed the 2 broken calls.

zefklop:
> 

Tanks, drama averted.

zefklop (approved):

learn-more (approved):

HBelusca:
For the second commit " [NTOS/CC] Fix broken usage of _SEH2_FINALLY ", I would suggest you fold the commit log text at around 74 characters, otherwise its display would showed messed up in e.g. git.reactos.org or other git tools (with the need to scroll a lot horizontally etc.)

HBelusca (approved):

PR #4983, opened 2023-01-06
[THEMES] "Open Sans" -> "Trebuchet MS"

JoachimHenze:
## Purpose

Eliminates the last references to "Open Sans" font.

This is an addendum to ( 0.4.13-dev-765-g 04a361d091095a96bb3af376e46daf4c5fe81c63 )
and 0.4.14-dev-20-g 2f4fb903b46207ab380365908401ded63752a108
"Open Sans" has been replaced by "Trebuchet MS" back then.

### DIFF
--- a/media/doc/3rd Party Files.txt
+++ b/media/doc/3rd Party Files.txt
@@ -178,7 +178,7 @@ License: media/fonts/doc/DejaVu/LICENSE.txt
 URL: http://dejavu.sourceforge.net
 
 Title: Open Sans Fonts
-Path: media/fonts
+Path: media/fonts/trebuc*.ttf
 Used Version: 1.10
 License: Apache-2.0 (https://spdx.org/licenses/Apache-2.0.html)
 URL: http://www.google.com/fonts/specimen/Open+Sans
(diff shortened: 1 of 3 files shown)


JoachimHenze:
Here is an after-pic how it renders with "Trebuchet MS". As you can see the result is not visually ideal, I guess using Tahoma (as it is used in the other font rendering sizes would look better) but I reused the same font again, that was used here historically. (Lautus ExtraLarge)
Lautus_extralarge

In case you want me to use Tahoma instead for consistency reasons to the other FontRendering sizes, just let me know. It is easy to be changed.

SergeGautherie:
Update '3rd Party Files.txt' too'?

JoachimHenze:
It seems 'Open Sans' string is found in 'trebuc*.ttf'. Is that intended?

Yes, as it was copied from that file. See the 2 commits that I linked.
Regarding '3rd Party Files.txt' I wouldn't know what to do in there except deleting that OpenSans section entirely. But what would be the benefit of that? Maybe they will improve some glyph in the future and we want to pick that via fontforge. In such a case it would be good to know where to find it.

katahiromz:
image

katahiromz (approved):

HBelusca (approved):

SergeGautherie:
> I wouldn't know what to do

If not deleting Open Sans, then reference Trebuchet MS at least.

JoachimHenze:
> > I wouldn't know what to do
> 
> If not deleting Open Sans, then reference Trebuchet MS at least.

done.

binarymaster (approved):

PR #4984, opened 2023-01-06
[HDAUDBUS] Add support for GUID_HDAUDIO_BUS_INTERFACE_BDL

oleg-dubinskiy:
## Purpose

FIX 1: Handle `GUID_HDAUDIO_BUS_INTERFACE_BDL` in `HDA_PDOHandleQueryInterface`.
This HD audio version is used by Realtek High Definition audio driver as well.

FIX 2: Break out from the device enumeration loop when max codec responses number is reached.
Stop the execution of device enumeration in `HDA_DpcForIsr` DPC routine if `ResponseCount` is equal or more of `MAX_CODEC_RESPONSES` definition.
Since the maximal number of items in Responses array is declared with this constant, it definitely makes no any sense to store more items there.
Simply switching to next loop iteration causes endless loop execution until the timeout is reached and bugcheck is occured. So just break when the array is full.
Don't dumbly repeat Haiku OS behaviour.

The Intel HDA specification (to which MSDN references), declares 3 possible amounts of Responses for RIRB: 2, 16 and 256. It's available at https://www.intel.com/content/dam/www/public/us/en/documents/product-specifications/high-definition-audio-specification.pdf.

It should fix bugcheck 0x7e when installing Realtek HD audio driver via Device Manager. Although another timer assert is still here, but after ignoring it, the installation should finish successfully at least.

JIRA issue: [CORE-18776](https://jira.reactos.org/browse/CORE-18776)

## Proposed changes

- Handle `GUID_HDAUDIO_BUS_INTERFACE_BDL` in `HDA_PDOHandleQueryInterface`.
- Properly return success for `GUID_HDAUDIO_BUS_INTERFACE_V2`, to avoid executing failure case upon finishing.
- Add some stubs for `HDAUDIO_BUS_INTERFACE_BDL` specific routines: `AllocateContiguousDmaBuffer`, `FreeContiguousDmaBuffer` and `SetupDmaEngineWithBdl`. Set them together with other ones in `HDA_PDOHandleQueryInterface`.
- Break out from the device enumeration loop when max codec responses number is reached. **TODO: make sure this fix is correct.**

## TODO

- [ ] Perhaps implement the HDA routines as much as possible. The driver seems to call them, so it needs the correct implementation of them.
- [ ] Test it on a compatible hardware (cc @julenuri). Unfortunately, I have no compatible controller(s) right now.

### DIFF
--- a/drivers/wdm/audio/hdaudbus/businterface.cpp
+++ b/drivers/wdm/audio/hdaudbus/businterface.cpp
@@ -294,6 +294,48 @@ HDA_UnregisterNotificationEvent(
 	return STATUS_NOT_IMPLEMENTED;
 }
 
+NTSTATUS
+NTAPI
+HDA_AllocateContiguousDmaBuffer(
+    IN PVOID _context,
+    IN HANDLE Handle,
+    ULONG RequestedBufferSize,
+    OUT PVOID *DataBuffer,
+    OUT PHDAUDIO_BUFFER_DESCRIPTOR *BdlBuffer)
+{
+	UNIMPLEMENTED;
+	ASSERT(FALSE);
+	return STATUS_NOT_IMPLEMENTED;
+}
+
+NTSTATUS
+NTAPI
+HDA_FreeContiguousDmaBuffer(
+    IN PVOID _context,
+    IN HANDLE Handle)
+{
+	UNIMPLEMENTED;
+	ASSERT(FALSE);
+	return STATUS_NOT_IMPLEMENTED;
+}
+
+NTSTATUS
+NTAPI
+HDA_SetupDmaEngineWithBdl(
+    IN PVOID _context,
+    IN HANDLE Handle,
+    ULONG BufferLength,
+    IN ULONG Lvi,
+    IN PHDAUDIO_BDL_ISR Isr,
+    PVOID Context,
+    OUT PUCHAR StreamId,
+    OUT PULONG FifoSize)
+{
+	UNIMPLEMENTED;
+	ASSERT(FALSE);
+	return STATUS_NOT_IMPLEMENTED;
+}
+
 
 NTSTATUS
 HDA_PDOHandleQueryInterface(
@@ -302,6 +344,7 @@ HDA_PDOHandleQueryInterface(
 {
     PIO_STACK_LOCATION IoStack;
     PHDAUDIO_BUS_INTERFACE_V2 InterfaceHDA;
+    PHDAUDIO_BUS_INTERFACE_BDL InterfaceBDL;
     PHDA_PDO_DEVICE_EXTENSION DeviceExtension;
 
     /* get device extension */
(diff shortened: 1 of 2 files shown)


drivers/wdm/audio/hdaudbus/businterface.cpp
```diff
@@ -363,10 +406,36 @@ HDA_PDOHandleQueryInterface(
...
+        InterfaceBDL->AllocateContiguousDmaBuffer = HDA_AllocateContiguousDmaBuffer;
+        InterfaceBDL->FreeContiguousDmaBuffer = HDA_FreeContiguousDmaBuffer;
+        InterfaceBDL->SetupDmaEngineWithBdl = HDA_SetupDmaEngineWithBdl;
+        return STATUS_SUCCESS;
     }
 
-    // FIXME
-    // implement support for GUID_HDAUDIO_BUS_INTERFACE_BDL
-    UNIMPLEMENTED;
+    DPRINT1("Unsupported HDAUDIO GUID requested\n");
```
SergeGautherie, line 439: Would it be worth to print the GUID too?

drivers/wdm/audio/hdaudbus/businterface.cpp
```diff
@@ -294,6 +294,48 @@ HDA_UnregisterNotificationEvent(
 	return STATUS_NOT_IMPLEMENTED;
 }
 
+NTSTATUS
+NTAPI
+HDA_AllocateContiguousDmaBuffer(
+    IN PVOID _context,
```
SergeGautherie, line 300:
```suggestion
    _In_ PVOID _context,
```

drivers/wdm/audio/hdaudbus/businterface.cpp
oleg-dubinskiy: Yes. But I doubt how to print it better. :thinking:

SergeGautherie: Something like ba1ed89f?

drivers/wdm/audio/hdaudbus/fdo.cpp
```diff
@@ -120,7 +120,7 @@ HDA_DpcForIsr(
         if (Codec->ResponseCount >= MAX_CODEC_RESPONSES)
         {
             DPRINT1("too many responses for codec %x Response %x ResponseFlags %x\n", Cad, Response, ResponseFlags);
-            continue;
+            break;
```
Extravert-ir, line 123: This is not a device enumeration code, but a reading RIRB code. If one item is wrong here, we shouldn't stop reading it as other entries may be correct.
If this code loops for you, the reason is definitely something else.

> Don't dumbly repeat Haiku OS behaviour.

Are you sure Haiku does the same here? :)

Extravert-ir (changes requested):

drivers/wdm/audio/hdaudbus/businterface.cpp
oleg-dubinskiy: Ah, yes. Then this is only one way to print the passed GUID.

drivers/wdm/audio/hdaudbus/businterface.cpp
```diff
@@ -294,6 +294,48 @@ HDA_UnregisterNotificationEvent(
...
+    IN HANDLE Handle,
+    ULONG RequestedBufferSize,
+    OUT PVOID *DataBuffer,
+    OUT PHDAUDIO_BUFFER_DESCRIPTOR *BdlBuffer)
+{
+	UNIMPLEMENTED;
+	ASSERT(FALSE);
```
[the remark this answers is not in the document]
oleg-dubinskiy: Okay, will remove them.

drivers/wdm/audio/hdaudbus/fdo.cpp
oleg-dubinskiy: > This is not a device enumeration code, but a reading RIRB code. If one item is wrong here, we shouldn't stop reading it as other entries may be correct.
If this code loops for you, the reason is definitely something else.

Ah, okay then. As I've suspected. I think the problem is too small max count of RIRB items defined by Haiku's driver.h:
https://github.com/reactos/reactos/blob/master/drivers/wdm/audio/hdaudbus/driver.h#L43
Intel HD Audio specification declares 3 possible counts for RIRB responses:
RIRB-size
https://www.intel.com/content/dam/www/public/us/en/documents/product-specifications/high-definition-audio-specification.pdf
2, 16 and 256.
Haiku's one is limited to 16, but I think it's not enough. Realtek HD Audio seems require 256, and in general, the maximum possible amount of items should be 256, to succeed for all types of the controllers. **Indeed, with 256, the installation doesn't fail anymore, although it installs 3 minutes in debug mode, and the count overflow case isn't reached.** :smiley: So yes, this is not a correct fix, so I should remove it, but set `MAX_CODEC_RESMPONSES` to 256.
> Are you sure Haiku does the same here? :)

Yep, just look at this code:
https://github.com/haiku/haiku/blob/ee4577937404e1f320bfde245f4dea1739735a92/src/add-ons/kernel/drivers/audio/hda/hda_controller.cpp#L401
It's obviously looks similar. :smile:

Extravert-ir: You are mixing `MAX_CODEC_RESPONSES` and RIRB size. RIRB size is represented by `DeviceExtension->RirbLength` variable, and is indeed set into one of 3 values inside `HDA_InitCorbRirbPos`.
`MAX_CODEC_RESPONSES` is a queue size where responses are stored for a specific codec. The fact that the `Codec->ResponseCount >= MAX_CODEC_RESPONSES` condition hits here, mean that the queue is full and nobody takes elements from it.

SergeGautherie:
I would suggest to complete the fix 1 part here, then move the fix 2 part to a future PR.

PR #4986, opened 2023-01-08
[ZLIB] Update to v1.2.13. CORE-18752

GGLinnk:
## Purpose

Updated zlib to v1.2.13 from v1.2.12.
Added 3rd party minizip LICENSE file.

JIRA issue: [CORE-18752](https://jira.reactos.org/browse/CORE-18752)

Source : https://github.com/madler/zlib/tree/04f42ceca40f73e2978b50e93806c2a18c1281fc

### Note
I have restored and disabled some of the code in ``zconf.h`` that defined ``WIN32_LEAN_AND_MEAN``.
I just tested it quickly and it seems to work fine, but I don't know if I should restore it completely or leave it disabled.

### Changes in 1.2.13 (13 Oct 2022) - From ChangeLog
- Fix configure issue that discarded provided CC definition
- Correct incorrect inputs provided to the CRC functions
- Repair prototypes and exporting of new CRC functions
- Fix inflateBack to detect invalid input with distances too far
- Have infback() deliver all of the available output up to any error
- Fix a bug when getting a gzip header extra field with inflate()
- Fix bug in block type selection when Z_FIXED used
- Tighten deflateBound bounds
- Remove deleted assembler code references
- Various portability and appearance improvements

### DIFF
--- a/sdk/include/reactos/libs/zlib/zconf.h
+++ b/sdk/include/reactos/libs/zlib/zconf.h
@@ -38,6 +38,9 @@
 #  define crc32                 z_crc32
 #  define crc32_combine         z_crc32_combine
 #  define crc32_combine64       z_crc32_combine64
+#  define crc32_combine_gen     z_crc32_combine_gen
+#  define crc32_combine_gen64   z_crc32_combine_gen64
+#  define crc32_combine_op      z_crc32_combine_op
 #  define crc32_z               z_crc32_z
 #  define deflate               z_deflate
 #  define deflateBound          z_deflateBound
@@ -349,6 +352,9 @@
 #    ifdef FAR
 #      undef FAR
 #    endif
+#    ifndef WIN32_LEAN_AND_MEAN
+#      define WIN32_LEAN_AND_MEAN
+#    endif
 #    include <windows.h>
      /* No need for _export, use ZLIB.DEF instead. */
      /* For complete Windows compatibility, use WINAPI, not __stdcall. */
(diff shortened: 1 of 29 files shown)


sdk/include/reactos/libs/zlib/zconf.h
```diff
@@ -349,6 +352,11 @@
 #    ifdef FAR
 #      undef FAR
 #    endif
+#    ifndef __REACTOS__
```
tkreuzer, line 355: Why is this needed?

GGLinnk: It was manually removed from previous release.
I've restored it so it's easier to maintain, track what is modified by REACTOS or not.
I've disabled it so it's stays disabled.
I'm not sur we should keep it disabled though. It's working fine with WIN32_LEAN_AND_MEAN.

binarymaster:
Minor thing, before merging, you need to fix caps lock in your last name (author field + GitHub profile) :sweat_smile: 

image

GGLinnk:
> Minor thing, before merging, you need to fix caps lock in your last name (author field + GitHub profile) 😅
> 
> image

I will update my name for the commits/project.

About my GitHub profile is it really required ?
I've always preferred my name in capital letters.

GGLinnk:
> Minor thing, before merging, you need to fix caps lock in your last name (author field + GitHub profile) 😅
> 
> image

This should be fixed.
I can't do anything for the GPG key though...

binarymaster:
This is okay for your GPG Key.

HBelusca (approved):

sdk/include/reactos/libs/zlib/zconf.h
GGLinnk: @tkreuzer
Does this answer your question ?

KRosUser2 (approved):

sdk/include/reactos/libs/zlib/zconf.h
GGLinnk: I consider this solved.

learn-more: This ifdef guard should be used if this was kept in the original code, but it needs to be removed for ROS.
If it's removed from the original code and we don't need it either, it can just be removed entirely.

GGLinnk: Ok, let me rephrase.

This piece of code was included with the original ZLIB source code but it was removed by a ROS dev from our version of code. It have been NOT removed from original ZLIB source code.

The ifdef guard have been put here to restore this removed piece of code and keep it disabled while I don't know why it was removed in the first place.

My real question is: It is really required to keep it guarded ?
So far my tests have shown no issue unguarded.

GGLinnk: Indeed... I probably compared it with a wrong reference...

Extravert-ir (approved):

learn-more (approved):

tkreuzer (approved):

PR #4989, opened 2023-01-10
[NTOS:MM] Support image prefixing in MmLoadSystemImage

RigoLigoRLC:
## Purpose

The `MmLoadSystemImage` function has a parameter called `IN PUNICODE_STRING NamePrefix OPTIONAL`. In the current ReactOS codebase, there's nowhere that this parameter is used. In Windows, this parameter is used in crash dump storage stack, where the special "crash mode" drivers would be loaded by kernel directly through MmLoadSystemImage, with a "dump_"(for crash dump)/"hiber_"(for hibernation) (see [reference post on OSR](https://community.osr.com/discussion/comment/97932/)).

When this parameter is specified, the actual driver's image to be loaded would be loaded under the prefixed name. The driver would appear in module list when running `lm` in WinDbg.

This isn't a urgently needed feature but I think it's better to have a solid foundation for my experiments on crash dump mechanism. And I should also begin with smaller kernel patches.

Jira ticket: [CORE-376](https://jira.reactos.org/browse/CORE-376)

In Windows XP (5.1.2600): 
图片
图片

This PR (with my experiments on crash dump):
图片
图片

## Proposed changes

- Added image prefix support in MmLoadSystemImage.

### DIFF
--- a/ntoskrnl/mm/ARM3/sysldr.c
+++ b/ntoskrnl/mm/ARM3/sysldr.c
@@ -2900,7 +2900,7 @@ MmLoadSystemImage(IN PUNICODE_STRING FileName,
     ULONG EntrySize, DriverSize;
     PLOAD_IMPORTS LoadedImports = MM_SYSLDR_NO_IMPORTS;
     PCHAR MissingApiName, Buffer;
-    PWCHAR MissingDriverName;
+    PWCHAR MissingDriverName, PrefixedBuffer = NULL;
     HANDLE SectionHandle;
     ACCESS_MASK DesiredAccess;
     PSECTION Section = NULL;
@@ -2964,7 +2964,52 @@ MmLoadSystemImage(IN PUNICODE_STRING FileName,
     PrefixName = *FileName;
 
     /* Check if we have a prefix */
-    if (NamePrefix) DPRINT1("Prefixed images are not yet supported!\n");
+    if (NamePrefix)
+    {
+        /* Check if "directory + prefix" is too long for the string */
+        Status = RtlUShortAdd(BaseDirectory.Length,
+                              NamePrefix->Length,
+                              &PrefixName.MaximumLength);
+        if (!NT_SUCCESS(Status))
+        {
+            Status = STATUS_INVALID_PARAMETER;
+            goto Quickie;
+        }
+
+        /* Check if "directory + prefix + basename" is too long for the string */
+        Status = RtlUShortAdd(PrefixName.MaximumLength,
+                              BaseName.Length,
+                              &PrefixName.MaximumLength);
+        if (!NT_SUCCESS(Status))
+        {
+            Status = STATUS_INVALID_PARAMETER;
+            goto Quickie;
+        }
+
+        /* Allocate the buffer exclusively used for prefixed name */
+        PrefixedBuffer = ExAllocatePoolWithTag(PagedPool,
+                                               PrefixName.MaximumLength,
+                                               TAG_LDR_WSTR);
+        if (!PrefixedBuffer)
+        {
+            Status = STATUS_INSUFFICIENT_RESOURCES;
+            goto Quickie;
+        }
+
+        /* Clear out the prefixed name string */
+        PrefixName.Buffer = PrefixedBuffer;
+        PrefixName.Length = 0;
+
+        /* Concatenate the strings */
+        RtlAppendUnicodeStringToString(&PrefixName, &BaseDirectory);
+        RtlAppendUnicodeStringToString(&PrefixName, NamePrefix);
+        RtlAppendUnicodeStringToString(&PrefixName, &BaseName);
+
+        /* Now the base name of the image becomes the prefixed version */
+        BaseName.Buffer = &(PrefixName.Buffer[BaseDirectory.Length / sizeof(WCHAR)]);
+        BaseName.Length += NamePrefix->Length;
+        BaseName.MaximumLength = (PrefixName.MaximumLength - BaseDirectory.Length);
+    }
 
     /* Check if we already have a name, use it instead */
     if (LoadedName) BaseName = *LoadedName;
@@ -3406,8 +3451,8 @@ Quickie:
     /* If we have a file handle, close it */
     if (FileHandle) ZwClose(FileHandle);
 
-    /* Check if we had a prefix (not supported yet - PrefixName == *FileName now) */
-    /* if (NamePrefix) ExFreePool(PrefixName.Buffer); */
+    /* If we have allocated a prefixed name buffer, free it */
+    if (PrefixedBuffer) ExFreePoolWithTag(PrefixedBuffer, TAG_LDR_WSTR);
 
     /* Free the name buffer and return status */
     ExFreePoolWithTag(Buffer, TAG_LDR_WSTR);


ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -2964,7 +2964,52 @@ MmLoadSystemImage(IN PUNICODE_STRING FileName,
     PrefixName = *FileName;
 
     /* Check if we have a prefix */
-    if (NamePrefix) DPRINT1("Prefixed images are not yet supported!\n");
+    if (NamePrefix)
+    {
+        /* Check if "directory + prefix" is too long for the string */
+        if(PrefixName.MaximumLength + BaseDirectory.Length < PrefixName.MaximumLength)
```
HBelusca, line 2970:
```suggestion
        if (PrefixName.MaximumLength + BaseDirectory.Length < PrefixName.MaximumLength)
```

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -2964,7 +2964,52 @@ MmLoadSystemImage(IN PUNICODE_STRING FileName,
...
+        {
+            Status = STATUS_INSUFFICIENT_RESOURCES;
+            goto Quickie;
+        }
+
+        /* Set maximum length to the size of "directory + prefix" */
+        PrefixName.MaximumLength += BaseDirectory.Length;
+
+        /* Check if "directory + prefix + basename" is too long for the string */
+        if(PrefixName.MaximumLength + BaseName.Length < PrefixName.MaximumLength)
```
HBelusca, line 2980:
```suggestion
        if (PrefixName.MaximumLength + BaseName.Length < PrefixName.MaximumLength)
```

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -2964,7 +2964,52 @@ MmLoadSystemImage(IN PUNICODE_STRING FileName,
...
+        }
+
+        /* Set the final maximum length to the size of "directory + prefix + basename" */
+        PrefixName.MaximumLength += BaseName.Length;
+
+        /* Allocate the buffer exclusively used for prefixed name */
+        PrefixedBuffer = ExAllocatePoolWithTag(PagedPool,
+                                               PrefixName.MaximumLength,
+                                               TAG_LDR_WSTR);
+        if(!PrefixedBuffer)
```
HBelusca, line 2993:
```suggestion
        if (!PrefixedBuffer)
```

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -2964,7 +2964,52 @@ MmLoadSystemImage(IN PUNICODE_STRING FileName,
...
+        PrefixName.Buffer = PrefixedBuffer;
+        PrefixName.Length = 0;
+
+        /* Concatenate the strings */
+        RtlAppendUnicodeStringToString(&PrefixName, &BaseDirectory);
+        RtlAppendUnicodeStringToString(&PrefixName, NamePrefix);
+        RtlAppendUnicodeStringToString(&PrefixName, &BaseName);
+
+        /* Now the base name of the image would become the prefixed version */
+        BaseName.Buffer = &(PrefixName.Buffer[BaseDirectory.Length * 2]);
```
HBelusca, line 3009: What is this "2" about? Is it meant for `sizeof(WCHAR)` or something else?

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -2964,7 +2964,52 @@ MmLoadSystemImage(IN PUNICODE_STRING FileName,
...
+    {
+        /* Check if "directory + prefix" is too long for the string */
+        if(PrefixName.MaximumLength + BaseDirectory.Length < PrefixName.MaximumLength)
+        {
+            Status = STATUS_INSUFFICIENT_RESOURCES;
+            goto Quickie;
+        }
+
+        /* Set maximum length to the size of "directory + prefix" */
+        PrefixName.MaximumLength += BaseDirectory.Length;
```
HBelusca, line 2977: I see what you are doing here (with the previous if-test): you are checking whether the addition is going to overflow.
I wonder whether we already have the following function, that you could use there:
https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ntintsafe/nf-ntintsafe-rtlushortadd

ntoskrnl/mm/ARM3/sysldr.c
RigoLigoRLC: Yes, this is the length of a single character. I forgot to use sizeof for this index.

RigoLigoRLC: Actually I just found out the buffer is of type `PWCH`, it points to WCHARs and should increment by 2 bytes each index. I'm not sure if that's a bug on my side and the fact the load completed without issue is kind of confusing. I'll investigate further later.

ntoskrnl/mm/ARM3/sysldr.c
learn-more: Can you please leave the style nitpicking for the last iteration, it makes the entire PR unreadable if you add a comment on every line...

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -2964,7 +2964,52 @@ MmLoadSystemImage(IN PUNICODE_STRING FileName,
...
 
     /* Check if we have a prefix */
-    if (NamePrefix) DPRINT1("Prefixed images are not yet supported!\n");
+    if (NamePrefix)
+    {
+        /* Check if "directory + prefix" is too long for the string */
+        if(PrefixName.MaximumLength + BaseDirectory.Length < PrefixName.MaximumLength)
+        {
+            Status = STATUS_INSUFFICIENT_RESOURCES;
+            goto Quickie;
```
GeoB99, line 2973: You aren't allocating chunk of memory here but the caller submitted an invalid argument. Use `STATUS_INVALID_PARAMETER`.

RigoLigoRLC: I have changed the return values when integer overflow happens to `STATUS_INVALID_PARAMETER`.

ntoskrnl/mm/ARM3/sysldr.c
RigoLigoRLC: I have changed the checks to use `RtlUShortAdd` now.

ntoskrnl/mm/ARM3/sysldr.c
RigoLigoRLC: The indexing bug was fixed and it looks right in debugger now. 
图片

ntoskrnl/mm/ARM3/sysldr.c
RigoLigoRLC: The if statement styles is changed to the proposed style now.

binarymaster:
I have added related Jira ticket to your PR: [CORE-376](https://jira.reactos.org/browse/CORE-376)

Also don't forget to mention ticket ID in your commits.

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -2964,7 +2964,52 @@ MmLoadSystemImage(IN PUNICODE_STRING FileName,
...
+    if (NamePrefix)
+    {
+        /* Check if "directory + prefix" is too long for the string */
+        Status = RtlUShortAdd(PrefixName.MaximumLength,
+                              BaseDirectory.Length,
+                              &PrefixName.MaximumLength);
+        if (!NT_SUCCESS(Status))
+        {
+            Status = STATUS_INVALID_PARAMETER;
+            goto Quickie;
```
learn-more, line 2976: In this case you are accessing an uninitialized pointer when you are calling `ExFreePoolWithTag(PrefixedBuffer, TAG_LDR_WSTR);`.
You should probably initialize `PrefixedBuffer` to `NULL`, and check that it is not null instead of checking for `NamePrefix` in the cleanup.

RigoLigoRLC: Changes have been pushed.

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -2964,7 +2964,52 @@ MmLoadSystemImage(IN PUNICODE_STRING FileName,
...
+        /* Clear out the prefixed name string */
+        PrefixName.Buffer = PrefixedBuffer;
+        PrefixName.Length = 0;
+
+        /* Concatenate the strings */
+        RtlAppendUnicodeStringToString(&PrefixName, &BaseDirectory);
+        RtlAppendUnicodeStringToString(&PrefixName, NamePrefix);
+        RtlAppendUnicodeStringToString(&PrefixName, &BaseName);
+
+        /* Now the base name of the image would become the prefixed version */
```
HBelusca, line 3008:
```suggestion
        /* Now the base name of the image becomes the prefixed version */
```

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -3406,8 +3451,8 @@ MmLoadSystemImage(IN PUNICODE_STRING FileName,
     /* If we have a file handle, close it */
     if (FileHandle) ZwClose(FileHandle);
 
-    /* Check if we had a prefix (not supported yet - PrefixName == *FileName now) */
-    /* if (NamePrefix) ExFreePool(PrefixName.Buffer); */
+    /* Check if we had a prefix, and free it if it's has been allocated */
+    if (NamePrefix && PrefixedBuffer) ExFreePoolWithTag(PrefixedBuffer, TAG_LDR_WSTR);
```
HBelusca, line 3455: Here, the `NamePrefix` check is now redundant.

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -2964,7 +2964,52 @@ MmLoadSystemImage(IN PUNICODE_STRING FileName,
     PrefixName = *FileName;
 
     /* Check if we have a prefix */
-    if (NamePrefix) DPRINT1("Prefixed images are not yet supported!\n");
+    if (NamePrefix)
+    {
+        /* Check if "directory + prefix" is too long for the string */
+        Status = RtlUShortAdd(PrefixName.MaximumLength,
```
HBelusca, line 2970: Shouldn't we look at `PrefixName.Length` here only? (equivalently, `FileName->Length`)

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -2964,7 +2964,52 @@ MmLoadSystemImage(IN PUNICODE_STRING FileName,
...
+
+        /* Check if "directory + prefix + basename" is too long for the string */
+        Status = RtlUShortAdd(PrefixName.MaximumLength,
+                              BaseName.Length,
+                              &PrefixName.MaximumLength);
+        if (!NT_SUCCESS(Status))
+        {
+            Status = STATUS_INVALID_PARAMETER;
+            goto Quickie;
+        }
```
HBelusca, line 2987: Shorter version:
```suggestion
        /* Get the length of "BaseDirectory + NamePrefix + BaseName" without overflow */
        Status = RtlUShortAdd(BaseDirectory.Length,
                              NamePrefix->Length,
                              &PrefixName.MaximumLength);
        if (NT_SUCCESS(Status))
        {
            Status = RtlUShortAdd(PrefixName.MaximumLength,
                                  BaseName.Length,
                                  &PrefixName.MaximumLength);
        }
        if (!NT_SUCCESS(Status))
        {
            Status = STATUS_INVALID_PARAMETER;
            goto Quickie;
        }
```

ntoskrnl/mm/ARM3/sysldr.c
RigoLigoRLC: hmm, I might first thought I need to "enlarge" the buffer based on the buffer size of the file name. And yeah we're only copying the part of buffer that's been used by the string so using the buffer size is not necessary. I'll change it to Length later.

ntoskrnl/mm/ARM3/sysldr.c
RigoLigoRLC: I felt this shorter version sacrifices a bit of readability.

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -2964,7 +2964,52 @@ MmLoadSystemImage(IN PUNICODE_STRING FileName,
     PrefixName = *FileName;
 
     /* Check if we have a prefix */
-    if (NamePrefix) DPRINT1("Prefixed images are not yet supported!\n");
+    if (NamePrefix)
+    {
+        /* Check if "directory + prefix" is too long for the string */
+        Status = RtlUShortAdd(NamePrefix->Length,
+                              BaseDirectory.Length,
```
HBelusca, line 2971: Putting in the order the elements show up in the string (directory, then nameprefix, then basename)
```suggestion
        Status = RtlUShortAdd(BaseDirectory.Length,
                              NamePrefix->Length,
```

HBelusca (approved):

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -2964,7 +2964,52 @@ MmLoadSystemImage(IN PUNICODE_STRING FileName,
...
+                                               TAG_LDR_WSTR);
+        if (!PrefixedBuffer)
+        {
+            Status = STATUS_INSUFFICIENT_RESOURCES;
+            goto Quickie;
+        }
+
+        /* Clear out the prefixed name string */
+        PrefixName.Buffer = PrefixedBuffer;
+        PrefixName.Length = 0;
```
learn-more, line 3001: You forgot to set PrefixName.MaxLength

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -2964,7 +2964,52 @@ MmLoadSystemImage(IN PUNICODE_STRING FileName,
     PrefixName = *FileName;
 
     /* Check if we have a prefix */
-    if (NamePrefix) DPRINT1("Prefixed images are not yet supported!\n");
+    if (NamePrefix)
+    {
+        /* Check if "directory + prefix" is too long for the string */
+        Status = RtlUShortAdd(BaseDirectory.Length,
```
learn-more, line 2970: Neither here nor below do you reserve place for a null terminator.

ntoskrnl/mm/ARM3/sysldr.c
```diff
@@ -2964,7 +2964,52 @@ MmLoadSystemImage(IN PUNICODE_STRING FileName,
...
+
+        /* Concatenate the strings */
+        RtlAppendUnicodeStringToString(&PrefixName, &BaseDirectory);
+        RtlAppendUnicodeStringToString(&PrefixName, NamePrefix);
+        RtlAppendUnicodeStringToString(&PrefixName, &BaseName);
+
+        /* Now the base name of the image becomes the prefixed version */
+        BaseName.Buffer = &(PrefixName.Buffer[BaseDirectory.Length / sizeof(WCHAR)]);
+        BaseName.Length += NamePrefix->Length;
+        BaseName.MaximumLength += NamePrefix->Length;
```
learn-more, line 3011: This can be incorrect, because you do not know the `MaximumLength` from the BaseName. (It does not have to have a relation to the used `Length` there,
this should actually be the remaining bytes inside your own buffer (PrefixName.MaximumLength - offset)

ntoskrnl/mm/ARM3/sysldr.c
HBelusca: The prefixed name is a counted string, it therefore doesn't necessarily need a null terminator.

RigoLigoRLC: I wasn't clear about this, originally assumed a string structure with a length property won't strictly need a null terminator. I checked `RtlInitUnicodeString` and it did reserve the place for null terminator. This may need more changes than just a few lines...

HBelusca: RtlInitUnicodeString does that because the C-string passed in order to initialize the UNICODE_STRING has a null-terminator.

RigoLigoRLC: According to MSDN https://learn.microsoft.com/en-us/windows/win32/api/ntdef/ns-ntdef-_unicode_string ,

> If the string is null-terminated, Length does not include the trailing null character.

So the `UNICODE_STRING` doesn't force the string to be null-terminated, nor does the `Length` cares about it. Am I wrong about RtlInitUnicodeString or is the implementation inaccurate?

HBelusca: That's correct. And RtlInitUnicodeString is perfectly fine. This does not conflict with those two remarks:
> If the string is null-terminated, Length does not include the trailing null character.

and
> The MaximumLength is used to **indicate the length of Buffer** so that if the string is passed to a conversion routine such as RtlAnsiStringToUnicodeString the returned string does not exceed the buffer size.

ntoskrnl/mm/ARM3/sysldr.c
RigoLigoRLC: Right. This is one more thing left from the previous "enlarging the buffer" mindset and forgot to be changed.

ntoskrnl/mm/ARM3/sysldr.c
RigoLigoRLC: Oh yeah it's MaximumLength not Length *facepalm*

ntoskrnl/mm/ARM3/sysldr.c
RigoLigoRLC: Related change is in place. Please let me know if it's appropriate.

ntoskrnl/mm/ARM3/sysldr.c
learn-more: it doesn't _need_ it, but i think it's good practice.

RigoLigoRLC:
I've squashed the review changes and changed commit author name as suggested now.

HBelusca (approved):

ThFabba (approved):

GeoB99 (approved):

PR #4990, opened 2023-01-10
Reduce crt heap diff

tkreuzer:
## Purpose

Reduce differences of heap.c to wine (version 5.0rc5)
This basically reverts most of the nasty hacks that have been applied.
The goal is to make the file auto-syncable with the winesync script.
It also serves as a lesson for everyone how to make 3rd-party code compile and work without hacking every other line.

## Proposed changes
- Get back MSVCRT prefix
- Use the original wine heap functions
- Remove some unneccessary changes
- Add some functions from wine
- Sync some wine commits

## Tests
https://reactos.org/testman/compare.php?ids=85614,85621,85626,85628,85630

### DIFF
--- a/sdk/lib/crt/wine/heap.c
+++ b/sdk/lib/crt/wine/heap.c
@@ -62,7 +62,7 @@ typedef int (CDECL *MSVCRT_new_handler_func)(MSVCRT_size_t size);
 static MSVCRT_new_handler_func MSVCRT_new_handler;
 static int MSVCRT_new_mode;
 
-/* FIXME - According to documentation it should be 8*1024, at runtime it returns 16 */
+/* FIXME - According to documentation it should be 8*1024, at runtime it returns 16 */ 
 static unsigned int MSVCRT_amblksiz = 16;
 /* FIXME - According to documentation it should be 480 bytes, at runtime default is 0 */
 static MSVCRT_size_t MSVCRT_sbh_threshold = 0;
@@ -74,6 +74,7 @@ void* CDECL MSVCRT_operator_new(MSVCRT_size_t size)
 {
   void *retval;
   int freed;
+  MSVCRT_new_handler_func handler;
 
   do
   {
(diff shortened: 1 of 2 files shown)


sdk/lib/crt/wine/heap.c
```diff
@@ -703,11 +703,9 @@ int CDECL MSVCRT_strncpy_s(char *dest, MSVCRT_size_t numberOfElements,
     if(!count)
         return 0;
 
-    if (!MSVCRT_CHECK_PMT(dest != NULL) || !MSVCRT_CHECK_PMT(src != NULL) ||
-        !MSVCRT_CHECK_PMT(numberOfElements != 0)) {
-        *_errno() = EINVAL;
```
HBelusca, line 589: Shouldn't errno be updated as well?

HBelusca (approved):
Dunno why both msi:msi and user32:menu wine tests appear to generate one or two more errors, but otherwise this looks fine.

Sad that Wine had to pollute their code with those "MSVCRT_" prefixes...

sdk/lib/crt/wine/heap.c
```diff
@@ -35,51 +60,131 @@
...
+static HANDLE heap, sb_heap;
 
-typedef int (CDECL *MSVCRT_new_handler_func)(size_t size);
+typedef int (CDECL *MSVCRT_new_handler_func)(MSVCRT_size_t size);
 
 static MSVCRT_new_handler_func MSVCRT_new_handler;
 static int MSVCRT_new_mode;
 
-/* FIXME - According to documentation it should be 8*1024, at runtime it returns 16 */
+/* FIXME - According to documentation it should be 8*1024, at runtime it returns 16 */
```
binarymaster, line 72: image

HBelusca: Potentially said 3rd-party (wine) code had a trailing whitespace?

binarymaster: It's okay in this case, which I didn't checked. :sweat_smile:

tkreuzer:
> Dunno why both msi:msi and user32:menu wine tests appear to generate one or two more errors, but otherwise this looks fine.

Those are flaky.

> Sad that Wine had to pollute their code with those "MSVCRT_" prefixes...

In fact they removed them. They will go away after syncing the file.

sdk/lib/crt/wine/heap.c
tkreuzer: It's updated in MSVCRT_CHECK_PMT

JoachimHenze:
Attention, one of the commits in this PR did introduce regression https://jira.reactos.org/browse/CORE-19505

PR #4991, opened 2023-01-12
[HDAUDBUS_NEW] Import new SkyLake/KabyLake/ApolloLake HD Audio Bus Driver (Apache 2.0 License)

oleg-dubinskiy:
## Purpose

Import new sklhdaudbus driver written by @coolstar community member from his GitHub repository: https://github.com/coolstar/sklhdaudbus.
Licensed under Apache 2.0 Public License: https://www.apache.org/licenses/LICENSE-2.0, so we can use it without any copyright restrictions. :slightly_smiling_face: 

JIRA issue: [CORE-18776](https://jira.reactos.org/browse/CORE-18776)

## Proposed changes

- Import the driver.
- Add License.
- Restore an inf file from the old hdaudbus.
- Remove SkyLake prefix in some file names and internal names.
- Restore some removed implementations and improve them.
- Reenable some disabled changes, as we want to use it as normal hdaudbus driver, without any ***Lake-specific features.
- Disable and remove some Windows 7/8/10-specific functionality. Use XP-compatible one where possible.
- Enable debugging for the driver.
- Fix all build issues (at least for GCC and Clang, MSVC and MSBuild are still failing, see TODO).
- Enable it by default and disable an old driver, to make possible to test it.

## TODO

- [ ] Fix MSVC and MSBuild compilation (help me do it correctly please! :pray: ).
- [ ] Update 3rd party modules list (where exactly should I update it? :thinking: ).
- [ ] **Test it with 3rd-party audio codecs and controllers (cc @julenuri )**.


coolstar:
So the original driver you're basing off of does not have working HD Audio under Windows yet (hence it was disabled). The work item also is disabled because it was causing crashes on Tiger Lake in Windows 11. As the person who wrote the original driver, I think it's far too early to import into ReactOS seeing as only the DSP portion of this driver currently works properly.

Also some of the changes you made to change MAXULONG64 -> MAXULONG for example, could cause problems on 64 bit systems (the driver was never tested on 32 bit, so it might be broken on both now with your changes?)

IMO it'd be best to fix and test the original driver under Windows Vista or up itself first (ideally Windows 10 or 11, as that's what the original driver was written for and tested to at least partially work under). Then import that into ReactOS once it's known to work. It's far too early to make changes like this to deviate from the original codebase before the original driver even works

Extravert-ir:
Closing the PR, as the code is in too unstable state.

PR #4994, opened 2023-01-14
[NTOS:PNP] Avoid recursion when walking the device tree.

ThFabba:
Let's not abuse the kernel stack...

### DIFF
--- a/ntoskrnl/io/pnpmgr/plugplay.c
+++ b/ntoskrnl/io/pnpmgr/plugplay.c
@@ -18,6 +18,12 @@ typedef struct _PNP_EVENT_ENTRY
     PLUGPLAY_EVENT_BLOCK Event;
 } PNP_EVENT_ENTRY, *PPNP_EVENT_ENTRY;
 
+typedef struct _IOP_FIND_DEVICE_INSTANCE_TRAVERSE_CONTEXT
+{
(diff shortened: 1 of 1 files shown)


ntoskrnl/io/pnpmgr/plugplay.c
```diff
@@ -136,7 +135,17 @@ IopGetDeviceObjectFromDeviceInstance(PUNICODE_STRING DeviceInstance)
             return NULL;
```
GGLinnk, line 135: Could this be simplified with:
```c
if (IopRootDeviceNode->PhysicalDeviceObject)
    ObReferenceObject(IopRootDeviceNode->PhysicalDeviceObject);
    
return IopRootDeviceNode->PhysicalDeviceObject;
```

Extravert-ir (approved):
Meh, need to finish the refactoring of device traversal operations and move everything to io/pnpmgr/devnode.c

HBelusca (approved):

ntoskrnl/io/pnpmgr/plugplay.c
ThFabba: Yes, that would be valid. Though I don't consider it so much simpler that I feel a need to touch it.

PR #4998, opened 2023-01-17
Fix butterflies screensaver

ThFabba:
* Fix a memory leak in gdi32
* Fix the butterflies screensaver not to release the DC it passes to OpenGL

The screensaver now works smoothly in ROS, and without blowing up our memory manager.

JIRA issue: [CORE-18498](https://jira.reactos.org/browse/CORE-18498)

### DIFF
--- a/modules/rosapps/applications/screensavers/butterflies/butterflies.c
+++ b/modules/rosapps/applications/screensavers/butterflies/butterflies.c
@@ -12,6 +12,7 @@
 HINSTANCE		hInstance;			// Holds The Instance Of The Application
 
 GLuint texture[3];	                //stores texture objects and display list
+HDC hdcOpenGL;
 
 LPCTSTR registryPath = _T("Software\\Microsoft\\ScreenSavers\\Butterflies");
 BOOL dRotate;
@@ -136,7 +137,7 @@ HGLRC InitOGLWindow(HWND hWnd)
 	hRC = wglCreateContext(hDC);
 	wglMakeCurrent(hDC, hRC);
 
-	ReleaseDC(hWnd, hDC);
+	hdcOpenGL = hDC;
 
 	return hRC;
 }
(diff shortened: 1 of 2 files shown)


learn-more (approved):
lgtm

JoachimHenze (approved):

modules/rosapps/applications/screensavers/butterflies/butterflies.c
```diff
@@ -285,8 +286,9 @@ LRESULT WINAPI ScreenSaverProc(HWND hWnd, UINT message,
 		ReleaseDC(hWnd, hDC);
 		break;
 	case WM_DESTROY:
-		wglMakeCurrent(NULL, NULL);
+		wglMakeCurrent(hdcOpenGL, NULL);
 		wglDeleteContext(hRC);
+		ReleaseDC(hWnd, hdcOpenGL);
```
binarymaster, line 291: It could be done without adding global `hdcOpenGL` variable, by using `wglGetCurrentDC()`.

So you can do this:
```c
		hDC = wglGetCurrentDC(); // hDC is a local variable in this function
		wglMakeCurrent(NULL, NULL); // this is correct
		wglDeleteContext(hRC);
		ReleaseDC(hWnd, hDC);
```

binarymaster (commented):
See for reference: https://learn.microsoft.com/en-us/windows/win32/opengl/win32-rendering-context-code-sample

modules/rosapps/applications/screensavers/butterflies/butterflies.c
learn-more: We don't do that in our other screensavers, does it add an advantage?

hope2586:
The first commit fixes https://jira.reactos.org/browse/CORE-16227.

PR #5000, opened 2023-01-17
[COMCTL32] Fix issue at BUTTON_CheckAutoRadioButton"

zecarlos1957:
[COMCTL32] Fix issue at BUTTON_CheckAutoRadioButton

     When we modify the selection in a group of AutoRadioButton  objects, the new button is not selected.

This happen when the WM_GETDLGCODE message returns bits other than (DLGC_BUTTON|DLGC_RADIOBUTTON) such DLGC_WANTARROWS, wich is quite common.

Adopted Solution: The comparison must be done using a mask that rejects the non-mandatory bits

Affected objects:
    Accessibility/Mouse Keys Settings/(On/off).
    Properties for Console/Options/(Cursor Size)
                                                       (Display Options)

### DIFF
--- a/dll/win32/comctl32/button.c
+++ b/dll/win32/comctl32/button.c
@@ -1667,7 +1667,8 @@ static void BUTTON_CheckAutoRadioButton( HWND hwnd )
     {
         if (!sibling) break;
 #ifdef __REACTOS__
-        if (SendMessageW( sibling, WM_GETDLGCODE, 0, 0 ) == (DLGC_BUTTON | DLGC_RADIOBUTTON))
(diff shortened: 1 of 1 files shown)


JoachimHenze:
> Does this help for https://jira.reactos.org/browse/CORE-6542 in any way?

According to @zecarlos1957 test it did fix the issue, so I mention that CORE-ID now in the tickets description.

zecarlos1957:
@JoachimHenze 
The logic is different from ReactOS to Wine, and I am not able to test these changes in wine code

zecarlos1957:
Current state of reactos does not show this issue. 
So i close the Pull Request

PR #503, opened 2018-04-19
Ignore IMAGE_SCN_CNT_UNINITIALIZED_DATA

learn-more:
## Purpose

This fixes loading binaries with some packers (like MPRESS)

JIRA issue: [CORE-14556](https://jira.reactos.org/browse/CORE-14556)

PR #5002, opened 2023-01-19
[NTOS:MM] really ignore explicit BSS sections.

roytam1:
## Purpose

VC1 Linker will fill BSS.SizeOfRawData with data in IMAGE_OPTIONAL_HEADER.SizeOfUninitializedData.

JIRA issue: [CORE-18797](https://jira.reactos.org/browse/CORE-18797)

## Proposed changes

This patch should really ignore any BSS sections.


KRosUser2:
This PR succesfully fixes https://jira.reactos.org/browse/CORE-18685 (please add it to the description)

KRosUser2 (approved):
Lgtm and tested ok on 
GnuChess : https://jira.reactos.org/browse/CORE-18685
Doom95 : https://jira.reactos.org/browse/CORE-18724 (To be added to the PR description)

To be checked for possible side effect and regression

ThFabba (changes requested):
We've concluded in the past that the kernel likely ignores all `IMAGE_SCN_CNT_*` flags. See https://github.com/reactos/reactos/pull/503
So I have to assume that this is not correct, and that the check needs to use a different criterion.

We have a test for this stuff here: https://git.reactos.org/?p=reactos.git;a=blob;f=modules/rostests/apitests/ntdll/NtMapViewOfSection.c;hb=e89675768be212fca2e59be68a4276cd308abd82#l1170  so the right way forward is to add one or more sections to that test with the characteristics you're interested in, and see what it says. We can't make any more changes here that aren't supported by regression tests or we'll just flip back and forth between incorrect implementations.

KRosUser2:
Also fixes https://jira.reactos.org/browse/CORE-18724 (to be added to description)

JoachimHenze:
>We can't make any more changes here that aren't supported by regression tests or we'll just flip back and forth between incorrect implementations.

I do agree that adding a test would be good.
But do we really actually know anything that this patch would break? That ancient packer MPress maybe, that Mark intended to fix back then? Has anybody ever tested that with this PR even? To me the results do look promising so far, so I it makes me a bit sad that you are so extra-careful here.

roytam1:
> We've concluded in the past that the kernel likely ignores all `IMAGE_SCN_CNT_*` flags. See #503

alright, can we confirm `PointerToRawData` = 0 is invalid value? if so, we can have another way to fix VC1 programs.

ThFabba (approved):

binarymaster (changes requested):
Hello,

```
git config --global user.name "Roy Tam"
git config --global user.email roytam@gmail.com
```

```
git config user.name "Roy Tam"
git config user.email roytam@gmail.com
```

Also please set up your GitHub account accordingly, so maintainers can use Squash & Merge button without getting the author field modified back to nickname by GitHub:

* Set your full name in [GitHub profile settings](https://github.com/settings/profile) (it should match your `user.name` setting)

Regards,

binarymaster:
Currently there is a nickname in author field of both commits:

image

binarymaster (approved):

PR #5004, opened 2023-01-21
[OLEAUT32] Return proper value from OaBuildVersion

RatinCN:
## Purpose
Return proper value from OaBuildVersion 
JIRA issue: [CORE-18419](https://jira.reactos.org/browse/CORE-18419)

## Proposed changes
- Returns MAKELONG(0xffff, 50) for Win2K3 (and ReactOS)
- No useless FIXME output

OaBuildVersion_On_2K3_W11

### DIFF
--- a/dll/win32/oleaut32/oleaut.c
+++ b/dll/win32/oleaut32/oleaut.c
@@ -660,6 +660,9 @@ ULONG WINAPI OaBuildVersion(void)
     case 0x00000005:  /* W2K */
 		return MAKELONG(0xffff, 40);
     case 0x00000105:  /* WinXP */
+#ifdef __REACTOS__
+    case 0x00000205:  /* Win2K3 */
+#endif /* __REACTOS__ */
     case 0x00000006:  /* Vista */
     case 0x00000106:  /* Win7 */
 		return MAKELONG(0xffff, 50);


KRosUser2 (approved):
lgtm, tested against the ticket case and now OK

dll/win32/oleaut32/oleaut.c
```diff
@@ -667,6 +668,9 @@ ULONG WINAPI OaBuildVersion(void)
 		FIXME("Version value not known yet. Please investigate it !\n");
 		return MAKELONG(0xffff, 40);  /* for now return the same value as for w2k */
     }
+#else
+    return MAKELONG(5014, 50); /* same value as for Win2K3 and later */
+#endif /* __REACTOS__ */
```
binarymaster, line 673: I guess you can just add a `case 0x00000205:  /* Win2K3 */` to the switch above.

KRosUser2: It was the proposal in my patch. This would be easily be pushed upstream to Wine too (for Win2K3 compat)....

RatinCN: Yes, I consider that:
 - If the return value other than Win2K3, it may be "incorrect" or "incompatible"
 - We're currently targeting Win2K3, so we needn't call `GetVersion`, and further, the same value also returned on Win11

If take them into account, add a `case 0x00000205:  /* Win2K3 */` and return `MAKELONG(0xffff, 50)` is recommended for this patch, I'd like to do it

gonzoMD: Imo, if it's possible to send a fix with this additional case to Wine without adding a nasty `#ifndef __REACTOS__`, that would be the better way. I believe that it would be accepted well, because you can prove it's correctness with a test program.

RatinCN: Committed, and comment of this PR has updated

dll/win32/oleaut32/oleaut.c
```diff
@@ -660,6 +660,7 @@ ULONG WINAPI OaBuildVersion(void)
     case 0x00000005:  /* W2K */
 		return MAKELONG(0xffff, 40);
     case 0x00000105:  /* WinXP */
+    case 0x00000205:  /* Win2K3 */
```
binarymaster, line 663: It's not guarded anymore. Did you sent a Merge Request to Wine GitLab?

HBelusca (approved):
Approved (for the concept, but yes, sending this to Wine should be done)

dll/win32/oleaut32/oleaut.c
RatinCN: Nope, I'm new to contribute files from wine, what should I do now? guarded or sent to wine?

gonzoMD: Send it to Wine. At least open a bug report and attach your Patch as well as your test program. After that you can create a merge request which refers to your Bug report.
Maybe our Wine ambassador Fabian Maurer can help to get it in?

KRosUser2: I proposed it as a Wine ticket here : https://bugs.winehq.org/show_bug.cgi?id=54257
Just update with your final implementation.

RatinCN: Committed, `#ifdef __REACTOS__` guarded now

JoachimHenze (approved):

dll/win32/oleaut32/oleaut.c
```diff
@@ -660,6 +660,9 @@ ULONG WINAPI OaBuildVersion(void)
     case 0x00000005:  /* W2K */
 		return MAKELONG(0xffff, 40);
     case 0x00000105:  /* WinXP */
+#ifdef __REACTOS__
+    case 0x00000205:  /* Win2003, ReactOS */
```
binarymaster, line 664:
```suggestion
    case 0x00000205:  /* Win2K3 */
```

RatinCN: committed

binarymaster (approved):

PR #5008, opened 2023-01-23
[SHELL32] Improve StartButton context menu

katahiromz:
## Purpose

JIRA issue: [CORE-16544](https://jira.reactos.org/browse/CORE-16544)

## Proposed changes

- Add `uFlags` argument to `CDefaultContextMenu::AddShellExtensionsToMenu`.
- `CCopyToMenu`, `CMoveToMenu`, and `CSendToMenu` check the `uFlags` against `(CMF_NOVERBS | CMF_VERBSONLY)`.

## TODO

- [x] Do tests.

## Screenshots

BEFORE:
before-1

AFTER:
after-1

### DIFF
--- a/dll/win32/shell32/CCopyToMenu.cpp
+++ b/dll/win32/shell32/CCopyToMenu.cpp
@@ -271,6 +271,11 @@ CCopyToMenu::QueryContextMenu(HMENU hMenu,
     TRACE("CCopyToMenu::QueryContextMenu(%p, %u, %u, %u, %u)\n",
           hMenu, indexMenu, idCmdFirst, idCmdLast, uFlags);
 
+    if ((uFlags & (CMF_NOVERBS | CMF_VERBSONLY)) != 0) // CORE-16544
+    {
+        return MAKE_HRESULT(SEVERITY_SUCCESS, 0, 0);
+    }
+
     m_idCmdFirst = m_idCmdLast = idCmdFirst;
 
     // insert separator if necessary
--- a/dll/win32/shell32/CCopyToMenu.cpp
+++ b/dll/win32/shell32/CCopyToMenu.cpp
@@ -271,10 +271,8 @@ CCopyToMenu::QueryContextMenu(HMENU hMenu,
     TRACE("CCopyToMenu::QueryContextMenu(%p, %u, %u, %u, %u)\n",
           hMenu, indexMenu, idCmdFirst, idCmdLast, uFlags);
 
-    if ((uFlags & (CMF_NOVERBS | CMF_VERBSONLY)) != 0) // CORE-16544
-    {
+    if (uFlags & (CMF_NOVERBS | CMF_VERBSONLY))
         return MAKE_HRESULT(SEVERITY_SUCCESS, 0, 0);
-    }
 
     m_idCmdFirst = m_idCmdLast = idCmdFirst;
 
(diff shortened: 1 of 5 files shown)


dll/win32/shell32/CSendToMenu.cpp
```diff
@@ -295,6 +295,11 @@ CSendToMenu::QueryContextMenu(HMENU hMenu,
     TRACE("%p %p %u %u %u %u\n", this,
           hMenu, indexMenu, idCmdFirst, idCmdLast, uFlags);
 
+    if ((uFlags & (CMF_NOVERBS | CMF_VERBSONLY)) != 0) // CORE-16544
```
JoachimHenze, line 298: If you believe your PR to be a proper solution I would recommend to strip the mentioning of CORE-16544 here and at the other places where you added that.
We do use such comments in the code only for temporary hacks and workarounds. (If you believe your PR would fall into that category, then it would be okay to keep them, but then the PRs title/description should be changed to reflect it being a hack.)
For now I do assume your PR is a proper fix. (?!?)

dll/win32/shell32/CCopyToMenu.cpp
```diff
@@ -271,6 +271,11 @@ CCopyToMenu::QueryContextMenu(HMENU hMenu,
     TRACE("CCopyToMenu::QueryContextMenu(%p, %u, %u, %u, %u)\n",
           hMenu, indexMenu, idCmdFirst, idCmdLast, uFlags);
 
+    if ((uFlags & (CMF_NOVERBS | CMF_VERBSONLY)) != 0) // CORE-16544
+    {
+        return MAKE_HRESULT(SEVERITY_SUCCESS, 0, 0);
+    }
```
JoachimHenze, line 277: I would omit the opening and closing curly braces for conditions where the commands do need only a single line.
Here and also at the other 2 places where such braces were added. That style is a giant waste of vertical space in a source file. Those can be added whenever a 2nd command would get added. And no: Developers won't forget that: There is static code analysis, we are not in the 80ies anymore. ;-)

dll/win32/shell32/CCopyToMenu.cpp
```diff
@@ -271,6 +271,11 @@ CCopyToMenu::QueryContextMenu(HMENU hMenu,
     TRACE("CCopyToMenu::QueryContextMenu(%p, %u, %u, %u, %u)\n",
           hMenu, indexMenu, idCmdFirst, idCmdLast, uFlags);
 
+    if ((uFlags & (CMF_NOVERBS | CMF_VERBSONLY)) != 0) // CORE-16544
```
JoachimHenze, line 274: != 0
 Can be stripped here and at the other 2 locations.

JoachimHenze:
I just tested: Our unpatched explorer does not show the erroneous SendTo-menu when being executed on XPSP3. I cannot say whether that renders this PR a hack or not, because it does changes in both: explorer.exe and shell32.dll.
And I used only explorer.exe for the cross-test, but did not replace shell32.dll as well, so I was testing with XPSP3 shell32.dll.
Maybe others can deduce better than myself whether that outcome does strengthen or weaken the trust in this PRs correctness.

katahiromz:
I have reverted `explorer`.

JoachimHenze (approved):
LGTM.
Wait until 3 approvals (without rejection) for immediate merge.
Merge in one week otherwise.

HBelusca (approved):

PR #5011, opened 2023-01-25
[SHELL32] Fix context menu command ids

ssafar:
## Purpose

JIRA issue: [CORE-18585](https://jira.reactos.org/browse/CORE-18585)

Right clicking & selecting "Properties" from background context menus now works.

## Proposed changes

We're mainly cleaning up how CDefaultContextMenu merges menus, with some random minor fixes.

The bug itself seems to have been caused by multiple things:

- CDefView was applying an offset of -1 on the command ids; context menu implementations assume that what they return in QueryContextMenu stays the same. We're no longer doing this.
- it was (to me?) not entirely obvious whether the "last" command id fields represent actual last ids or the next ids after (as per the checks, it was the latter); I renamed them to "end" to follow the STL convention of "end means a non-inclusive upper bound"
- these bounds are also always set now, even if the ranges are empty (e.g. "ids 7 to 7"); this allows the removal of some extra checks (for equality)
- it also wasn't super easy to see how we were supposed to know what ids the callback in CDefaultContextMenu added (which we would then need to skip) as it doesn't really return anything; this got replaced with just giving them 48 (0x30) ids to pick from.
- there is also a minor fix in CDefView: we were using the & operator to produce a pointer to a list of children, but the ATL class implements "&" in a way where we go from holding a NULL pointer to holding something else, so in debug builds it'll fail an assertion. Now we're handing out an actual separate list.

## TODO

- [x] more detailed testing? I haven't found any automated tests yet (are there any) & there might be some corner cases?

### DIFF
--- a/dll/win32/shell32/CDefaultContextMenu.cpp
+++ b/dll/win32/shell32/CDefaultContextMenu.cpp
@@ -24,6 +24,9 @@ typedef struct _StaticShellEntry_
     HKEY hkClass;
 } StaticShellEntry, *PStaticShellEntry;
 
+/* Maximum number of items that the callback can add. See also the comment where
+   this is being used. */
+#define MAX_MENU_ITEMS_ADDED_BY_CALLBACK 0x30
 
 //
 // verbs for InvokeCommandInfo
@@ -67,16 +70,20 @@ class CDefaultContextMenu :
         UINT m_cKeys;
         PIDLIST_ABSOLUTE m_pidlFolder;
         DWORD m_bGroupPolicyActive;
+
+        // Ids and entries of commands added by various sources; "first" is the
+        // first valid id added, while "end" is the first id not included in the
+        // range.
         CAtlList<DynamicShellEntry> m_DynamicEntries;
         UINT m_iIdSHEFirst; /* first used id */
-        UINT m_iIdSHELast; /* last used id */
+        UINT m_iIdSHEEnd;
         CAtlList<StaticShellEntry> m_StaticEntries;
-        UINT m_iIdSCMFirst; /* first static used id */
-        UINT m_iIdSCMLast; /* last static used id */
-        UINT m_iIdCBFirst; /* first callback used id */
-        UINT m_iIdCBLast;  /* last callback used id */
-        UINT m_iIdDfltFirst; /* first default part id */
-        UINT m_iIdDfltLast; /* last default part id */
+        UINT m_iIdSCMFirst; /* static */
+        UINT m_iIdSCMEnd;
+        UINT m_iIdCBFirst; /* from callback */
+        UINT m_iIdCBEnd;
+        UINT m_iIdDfltFirst; /* default part */
+        UINT m_iIdDfltEnd;
 
         HRESULT _DoCallback(UINT uMsg, WPARAM wParam, LPVOID lParam);
         void AddStaticEntry(const HKEY hkeyClass, const WCHAR *szVerb);
@@ -144,13 +151,13 @@ CDefaultContextMenu::CDefaultContextMenu() :
     m_pidlFolder(NULL),
     m_bGroupPolicyActive(0),
     m_iIdSHEFirst(0),
-    m_iIdSHELast(0),
+    m_iIdSHEEnd(0),
     m_iIdSCMFirst(0),
-    m_iIdSCMLast(0),
+    m_iIdSCMEnd(0),
     m_iIdCBFirst(0),
-    m_iIdCBLast(0),
+    m_iIdCBEnd(0),
     m_iIdDfltFirst(0),
-    m_iIdDfltLast(0)
+    m_iIdDfltEnd(0)
 {
 }
 
@@ -617,9 +624,10 @@ CDefaultContextMenu::QueryContextMenu(
     UINT uFlags)
 {
     HRESULT hr;
+    /* The next available free command id. */
     UINT idCmdNext = idCmdFirst;
-    UINT cIds = 0;
 
+    UINT added = 0;
     TRACE("BuildShellItemContextMenu entered\n");
 
     /* Load static verbs and shell extensions from registry */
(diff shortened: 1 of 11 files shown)


dll/win32/shell32/CDefaultContextMenu.cpp
```diff
@@ -144,13 +151,13 @@ CDefaultContextMenu::CDefaultContextMenu() :
     m_pidlFolder(NULL),
     m_bGroupPolicyActive(0),
     m_iIdSHEFirst(0),
-    m_iIdSHELast(0),
+    m_iIdSHEEnd(0),
```
HBelusca, line 154: Why not keeping the "Last" qualifier?

ssafar: since, at least to me, "last" sounds like it's the id of the last command included even though it's not (e.g. first 5 last 8 includes 5, 6, 7, but not 8), while "end" is already used in the "smaller than this" sense e.g. in STL. I can change it back though if you think it's clear enough already!

(at some point I was convinced that this is where the one-off is coming from... it wasn't though :))

ssafar:
Actually, a question:

is my assumption correct where... if `QueryContextMenu` gets called with a nonzero `idCmdFirst`, then subsequent `InvokeCommand` calls will get back the exact menu values submitted, and thus will need to remember the offset to translate them back?

Since... after additional testing, it looks like there are quite some places where this isn't being done yet (e.g. netshell, Recycle Bin, etc.), which would then acquire a one-off with this patch and would need to be fixed, too.

(As per the [docs](https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-icontextmenu-invokecommand):

>If the verb, specified either by a canonical verb name or the command ID is not recognized by the context menu handler, it must return a failure (E_FAIL) so that the verb can be passed on to other context menu handlers that might implement it.

... implying that multiple context menu handlers would share the same ID space, meaning no translation.)

Going with this for the time being; will submit these fixes soon!

ssafar:
... done; looks like this also fixes https://jira.reactos.org/browse/CORE-18349 as a byproduct?

learn-more:
> Actually, a question:
> 
> is my assumption correct where... if `QueryContextMenu` gets called with a nonzero `idCmdFirst`, then subsequent `InvokeCommand` calls will get back the exact menu values submitted, and thus will need to remember the offset to translate them back?
> 
> Since... after additional testing, it looks like there are quite some places where this isn't being done yet (e.g. netshell, Recycle Bin, etc.), which would then acquire a one-off with this patch and would need to be fixed, too.
> 
> As per the [docs](https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-icontextmenu-invokecommand):
> 
> > If the verb, specified either by a canonical verb name or the command ID is not recognized by the context menu handler, it must return a failure (E_FAIL) so that the verb can be passed on to other context menu handlers that might implement it.
> 
> ... implying that multiple context menu handlers would share the same ID space, meaning no translation.)
> 
> Going with this for the time being; will submit these fixes soon!

No, those offsets should not be received back.

e.g. the shell extension just does:
`id = idCmdFirst + HARDCODED_SHELL_EXTENSION_ID1`
And later it will receive just `HARDCODED_SHELL_EXTENSION_ID1`

learn-more (changes requested):
Shell extensions should not need to remember their 'initial' offset, this should be handled by the shell itself.

katahiromz:
@ssafar Thanks for providing PR. Enjoy coding! :)

Some questions:
1. Which do the `IContextMenu` objects remember `idCmdFirst`? Which don't?
2. What is the definition of the 1st parameter of `IContextMenu::InvokeCommand`?
3. If the `IContextMenu` object don't remember `idCmdFirst`, then how does the shell invoke the verbs.

ssafar:
Sorry for the latency & thanks everyone for the comments!

Sounds like I was wrong about the offsets (I'm actually happy about this since having to store the first one would make the API a lot more uglier, by design); re-reading the docs did eventually result in finding...

> A common practice is to set the first command identifier to idCmdFirst (an **offset** of zero) and increment the offset for each additional command by one.

(from QueryContextMenu)

along with

> where **offset** is the menu-identifier offset of the command to carry out.

(from CMINVOKECOMMANDINFO)

making it pretty clear that it's the offset **compared to** idCmdFirst, which is itself not the offset. Thanks @learn-more for pointing it out!

Meanwhile... not sure of this, but CNewMenu.cpp is a bit suspicious to me that it does offsetting itself:
```
        SHELLNEW_ITEM *pItem = FindItemFromIdOffset(LOWORD(lpici->lpVerb));
```
```
CNewMenu::SHELLNEW_ITEM *CNewMenu::FindItemFromIdOffset(UINT IdOffset)
{
    /* Folder */
    if (m_idCmdFirst + IdOffset == m_idCmdFolder)
```
... but then it seems to be doing the right thing later? Everyone else seems to be going with the now-obviously-correct model.

I did get some enjoyment out of trying to dig up how TortoiseSVN/GIT does things though... they have an ID map for this, and each time they add something to it... [example](https://github.com/TortoiseGit/TortoiseGit/blob/master/src/TortoiseShell/ContextMenu.cpp#L655):
```
  myIDMap[id - idCmdFirst] = com;
  myIDMap[id] = com;
```
... just in case I guess? :)

Anyway: will get rid of the "fixes" & get it working again!

katahiromz:
Some menus are comming from resource.
How will you deal with them?

HBelusca:
@katahiromz @whindsaks Could you guys verify whether this PR still needs to be addressed, or whether similar fixes have already been merged in master since?

whindsaks:
> @katahiromz @whindsaks Could you guys verify whether this PR still needs to be addressed, or whether similar fixes have already been merged in master since?

The verb fix in CRecycleBin.cpp is still needed, the rest is hard to say without testing but as already commented, the shell extension itself is not the one that has to remember the offset.

PR #5014, opened 2023-01-26
[NTUSER] Allow Window Snap to be disabled

katahiromz:
## Purpose
Allow the user to disable Window Snap for usability.

## Proposed changes

- Add `IntIsWindowSnapEnabled` helper function that reads registry value `WindowArrangementActive`.
- Check the registry and store the value to newly-added `g_bWindowSnapEnabled` global variable on startup.
- Check the global variable before Window Snap.

## Screenshot
after
`Win+Left`, `Win+Right`, `Win+Up`, and `Win+Down` can be disabled by registry value `WindowArrangementActive`.
Snapping mouse can be also disabled.


JoachimHenze:
Ftr, this is not implemented like it is implemented on Windows 7:
I double-checked on Win7 build 7600:
- "killing and restarting explorer" is not enough to make it have an effect
- logging out and back on is enough to make it have an effect.

As far as I can anticipate from Katayamas implementation, his solution would react immediately to the registry change, without any need for logoff/logon.
I do not reject the PR today, despite that difference.
But if we would ever detect any visible performance degradation due to that difference in implementation, it should be changed to only be checked during logon, and afterwards held as a cached value instead of always rereading from registry.

win32ss/user/ntuser/defwnd.c
```diff
@@ -529,6 +529,19 @@ DefWndScreenshot(PWND pWnd)
     UserCloseClipboard();
 }
 
+/* Is Window Snap enabled? */
+BOOL IsWindowSnapEnabled(VOID)
+{
+    WCHAR szValue[16];
```
JoachimHenze, line 535: Why 16 elements? An arbitrary choice? We need only "0" / "1" plus maybe some UNICODE_NULL here, I guess.

JoachimHenze (approved):

win32ss/user/ntuser/defwnd.c
katahiromz: Fixed as `2`.

win32ss/user/ntuser/defwnd.c
```diff
@@ -529,6 +529,19 @@ DefWndScreenshot(PWND pWnd)
...
 
+/* Is Window Snap enabled? */
+BOOL IsWindowSnapEnabled(VOID)
+{
+    WCHAR szValue[2];
+    if (RegReadUserSetting(L"Control Panel\\Desktop", L"WindowArrangementActive",
+                           REG_SZ, szValue, sizeof(szValue)))
+    {
+        szValue[RTL_NUMBER_OF(szValue) - 1] = UNICODE_NULL; /* Avoid buffer overrun */
+        return _wtoi(szValue);
```
HBelusca, line 540:
```suggestion
        return (_wtoi(szValue) != 0);
```

win32ss/user/ntuser/defwnd.c
```diff
@@ -529,6 +529,19 @@ DefWndScreenshot(PWND pWnd)
...
+    }
+    return TRUE;
```
HBelusca, line 542: So window-snap is "by default" enabled?

JoachimHenze: This is what is implemented here, yes.
Since CORE-16379 (request for option to disable it) is from me, I do hereby state that I am okay with that chosen default. Even if it is not the default in 2k3sp2 (which does not have this feature at all yet).

In case you would request to turn it off by default, I would also be okay with that (I am neutral), but that was not explicitly requested by me. So the chosen default will not make **me** reject this PR.

JoachimHenze:
Katayama will your PR disable both "Snapping by mouse movements" and the Keyboard-accelerators (same like the flag works in Win7), or does it just disable one of both?

win32ss/user/ntuser/defwnd.c
HBelusca: Adding the double exclamation point should work perfectly as well. I don't have preference between the two versions. The "!= 0" is more common though.

katahiromz: Added paren and added `!= 0`.

katahiromz:
> Katayama will your PR disable both "Snapping by mouse movements" and the Keyboard-accelerators (same like the flag works in Win7), or does it just disable one of both?

Snapping by mouse is not disabled yet.

katahiromz:
Now, snapping by mouse can be disabled.

Extravert-ir (changes requested):
I'd say accessing the registry each time would consume too much resources. Let's cache it somehow.

katahiromz:
Now cached.

win32ss/user/ntuser/sysparams.c
```diff
@@ -346,11 +345,25 @@ SpiUpdatePerUserSystemParameters(VOID)
...
+    }
+    return TRUE;
+}
+
 BOOL
 InitSysParams(VOID)
 {
     SpiUpdatePerUserSystemParameters();
     gbSpiInitialized = TRUE;
+    g_bWindowSnapEnabled = IntIsWindowSnapEnabled();
```
learn-more, line 366: Considering it reads from the user setting, should it go into `SpiUpdatePerUserSystemParameters` ?

katahiromz: Moved.

HBelusca (approved):

learn-more (approved):

PR #5015, opened 2023-01-27
Reduce diff of wine crt code to wine-5.0-rc5

tkreuzer:
## Purpose

Reduce diff between crt wine code and wine-5.0-rc5

## Proposed changes

- Remove white space changes
- Merge a number of missing commits

## Tests
Rightmost 2 are this PR
KVM: https://reactos.org/testman/compare.php?ids=85666,85673,85682,85678,85681
VBox: https://reactos.org/testman/compare.php?ids=85667,85671,85683,85679,85680

### DIFF
--- a/sdk/lib/crt/wine/heap.c
+++ b/sdk/lib/crt/wine/heap.c
@@ -175,7 +175,7 @@ void* CDECL MSVCRT_operator_new(MSVCRT_size_t size)
 
   TRACE("(%ld) out of memory\n", size);
 #if _MSVCR_VER >= 80
-  throw_bad_alloc("bad allocation");
+  throw_exception(EXCEPTION_BAD_ALLOC, 0, "bad allocation");
 #endif
   return NULL;
 }
--- a/sdk/lib/crt/wine/heap.c
+++ b/sdk/lib/crt/wine/heap.c
@@ -418,7 +418,15 @@ size_t CDECL _aligned_msize(void *p, MSVCRT_size_t alignment, MSVCRT_size_t offs
  */
 void* CDECL MSVCRT_calloc(MSVCRT_size_t count, MSVCRT_size_t size)
 {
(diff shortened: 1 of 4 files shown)


sdk/lib/crt/wine/heap.c
```diff
@@ -779,10 +822,11 @@ int CDECL MSVCRT_memcpy_s(void *dest, MSVCRT_size_t numberOfElements, const void
         return MSVCRT_ERANGE;
     }
 
-    memcpy(dest, src, count);
```
Extravert-ir, line 782: WTF, why do they call `memmove` inside `memcpy`?

learn-more: * inside memcpy_s

Extravert-ir: whatever :) question still remains

ColinFinck: Quoting the msvcrt:string test:
https://github.com/reactos/reactos/blob/d5999c6b66c4779bc71db51df4189e2a465b0ee5/modules/rostests/winetests/msvcrt/string.c#L3927-L3928

Actually, we got that wrong a few times in ReactOS history:
- The 2005 edition: https://github.com/reactos/reactos/commit/1b17f93d3a6b8cea4e6d51baac033abdbae2826d - "MSVCRT memcpy behaves like memmove for overlapping moves"
- The 2009 edition: https://github.com/reactos/reactos/commit/d41ad9a51b5bbe86e24466795584f471c131d038 - Make memcpy a duplicate of a memmove (confirmed by testing), there is a dependency on this behaviour.
- The 2012 edition: https://github.com/reactos/reactos/commit/ff6c1ce97291b7f5830c700a108608655f5ded5f - memcpy should behave exactly like memmove for Windows compatibility.

ColinFinck (approved):

PR #5018, opened 2023-01-28
[DNSAPI][LDNS] Replacing ADNS with LDNS

mann3616:
## Purpose

Removed ADNS in DNSAPI and added LDNS library to sdk/lib/3rdparty/

JIRA issue: [CORE-11410](https://jira.reactos.org/browse/CORE-11410)

## Proposed changes

I propose to remove ADNS which is an outdated library and switch it out with the [LDNS](https://www.nlnetlabs.nl/documentation/ldns/dir_fbc3dd5c1925e6acb2daf846ece5427b.html) library. 

## TODO

- [x] Fix no internet connection error.


dll/win32/dnsapi/CMakeLists.txt
```diff
@@ -2,16 +2,16 @@
 include_directories(
     include
     ${REACTOS_SOURCE_DIR}/sdk/include/reactos/idl
-    ${REACTOS_SOURCE_DIR}/sdk/lib/3rdparty/adns/src
-    ${REACTOS_SOURCE_DIR}/sdk/lib/3rdparty/adns/adns_win32)
```
HBelusca, line 6: You should then remove those corresponding files as well from the source tree? Or are they also used elsewhere in our code?

Extravert-ir:
Two questions:
- Does it fix the issue mentioned?
- Why did you choose LDNS particularly?

mann3616:
> Two questions:
> 
> * Does it fix the issue mentioned?
I put the wrong core, but this does fix the issue mentioned as it is a maintained and better version of ADNS.
> * Why did you choose LDNS particularly?
I chose LDNS because it also offers a bunch of other utilities that ADNS does not have, which means it will prove to be more useful as the network functionality of ReactOS improves. Such utilities are DNSSEC signing and verification + TSIG, and requires OpenSSL for it's cryptographic functions. ADNS does not support all these utilities so I believe LDNS is an equivalent fit as of now, and will be a better fit as ReactOS grows in functionality.

@Extravert-ir

PR #5019, opened 2023-01-30
[REACTOS][CMAKE][BOOTDATA] Replace reactos with ReactOS for bootcd/livecd

katahiromz:
## Purpose

JIRA issue: [CORE-14005](https://jira.reactos.org/browse/CORE-14005)

## Proposed changes

- Replace `'reactos'` with `'ReactOS'` in `bootcd`/`livecd` paths.

## TODO

- [x] Do small tests.
- [x] Do big tests.


katahiromz:
bootcd
I have booted on bootcd.

ericsomto (approved):
:-)

katahiromz:
Test KVM BEFORE: c3644fd https://reactos.org/testman/compare.php?ids=85704
Test KVM AFTER: 0457746 Testing...
https://reactos.org/testman/compare.php?ids=85704,???

Test VBox BEFORE: c3644fd https://reactos.org/testman/compare.php?ids=85705
Test VBox AFTER: 0457746 Testing...
https://reactos.org/testman/compare.php?ids=85705,???

Extravert-ir (approved):

learn-more (changes requested):
I am not sure which problem this is fixing?

katahiromz:
> I am not sure which problem this is fixing?

Consistency and speed (not confirmed yet).

HBelusca (changes requested):
Not convinced this makes any sense. Indeed, those paths in the bootcd/livecd are only used when ReactOS is running from the CD itself, in which case e.g. BTRFS is not used to read the system files from the CD.
And when ReactOS is installed, it will use whatever the user has specified as the installation path during the 1st-stage installation, and it can be anything (`ReactOS`, `rEaCtOs`, `MyLittlePony`, etc.)

katahiromz:
OK. Won't Fix. Closed.

PR #5031, opened 2023-01-31
[SHELL32_APITEST] Add exception for w2k3's version of SHGetAttributesFromDataObject

learn-more:


learn-more:
Need to fix this with a VM that reproduces.

PR #5036, opened 2023-02-01
[APPHELP_APITEST] dwMinor indicates the current date on Windows below vista....

learn-more:


modules/rostests/apitests/apphelp/db.cpp
```diff
@@ -223,10 +223,16 @@ static void test_GetDatabaseInformationEmpty(PDB pdb)
         {
             ok(pInfo->dwMajor == 2, "Expected pInfo->dwMajor to be 2, was: %d\n", pInfo->dwMajor);
             if (g_WinVersion >= _WIN32_WINNT_VISTA)
+            {
```
JoachimHenze, line 226: Yuck!

tkreuzer (approved):

modules/rostests/apitests/apphelp/db.cpp
learn-more: unbalanced braces is even more yuck

PR #5044, opened 2023-02-03
[USER32_WINETEST] Change module type to win32cui

tkreuzer:
## Purpose

Change module type to win32cui to be able to see the test results.

## Tests
Test WHS: https://reactos.org/testman/compare.php?ids=85895,85896,85897,85899,85900

### DIFF
--- a/modules/rostests/winetests/user32/CMakeLists.txt
+++ b/modules/rostests/winetests/user32/CMakeLists.txt
@@ -47,7 +47,6 @@ if(CMAKE_C_COMPILER_ID STREQUAL "GNU")
 endif()
 
 add_pch(user32_winetest precomp.h "${PCH_SKIP_SOURCE}")
-# some tests need to be run from an app compiled as GUI
-set_module_type(user32_winetest win32gui)
+set_module_type(user32_winetest win32cui)
 add_importlibs(user32_winetest user32 gdi32 advapi32 msvcrt kernel32)
 add_rostests_file(TARGET user32_winetest)


ColinFinck (approved):

modules/rostests/winetests/user32/CMakeLists.txt
```diff
@@ -48,6 +48,6 @@ endif()
 
 add_pch(user32_winetest precomp.h "${PCH_SKIP_SOURCE}")
 # some tests need to be run from an app compiled as GUI
```
ThFabba, line 50: If this is no longer true, you should remove this comment.
That said, has anything changed to make this unnecessary now?

tkreuzer: TBH, I don't know and this comment / the commit message don't help. I don't see any regression on Test WHS though, so I think it should be ok.

SergeGautherie: Current PR reverts 559bf9639468e83ff34a6f06d3501ccb1efdccb3.

HBelusca (approved):
Regarding the "some tests..." comment, as soon as a "windows"-type app (incl. CUI ones) make one first GUI call, it is treated as such in the user32/win32k, so I don't see why those tests would still fail.

modules/rostests/winetests/user32/CMakeLists.txt
HBelusca: @zefklop could shed light on that, then...

katahiromz (approved):

zefklop:
I distinctly remember that this fixed some tests at some point. Maybe the interactive ones, I can't remember. Let's merge it, maybe something in win32k was fixed since then.

zefklop (approved):

PR #5045, opened 2023-02-03
[WIN32SS] Improve UserHMGetHandle definition

TAN-Gaming:
## Proposed changes
Add UserHMGetHandleSafe macro that does the NULL check and use it instead of writing NULL check multiple times

### DIFF
--- a/win32ss/include/ntuser.h
+++ b/win32ss/include/ntuser.h
@@ -228,6 +228,7 @@ typedef struct _PROCMARKHEAD
 } PROCMARKHEAD, *PPROCMARKHEAD;
 
 #define UserHMGetHandle(obj) ((obj)->head.h)
+#define UserHMGetHandleSafe(obj) ((obj) ? (obj)->head.h : NULL)
 
 /* Window Client Information structure */
 struct _ETHREAD;
--- a/win32ss/include/ntuser.h
+++ b/win32ss/include/ntuser.h
@@ -227,6 +227,12 @@ typedef struct _PROCMARKHEAD
     struct _PROCESSINFO *ppi;
 } PROCMARKHEAD, *PPROCMARKHEAD;
 
+/**
+ * @note
+ * Use UserHMGetHandleSafe() by default if obj has not been checked for NULL already.
+ * Use UserHMGetHandle() if obj has been checked for NULL already, or if it contains
+ * a function-call (to not call that function twice).
+ **/
 #define UserHMGetHandle(obj) ((obj)->head.h)
 #define UserHMGetHandleSafe(obj) ((obj) ? (obj)->head.h : NULL)
 
--- a/win32ss/user/ntuser/focus.c
+++ b/win32ss/user/ntuser/focus.c
@@ -34,7 +34,7 @@ HWND FASTCALL
 IntGetCaptureWindow(VOID)
 {
    PUSER_MESSAGE_QUEUE ForegroundQueue = IntGetFocusMessageQueue();
-   return ( ForegroundQueue ? (ForegroundQueue->spwndCapture ? UserHMGetHandle(ForegroundQueue->spwndCapture) : 0) : 0);
+   return (ForegroundQueue ? UserHMGetHandleSafe(ForegroundQueue->spwndCapture) : NULL);
 }
 
 HWND FASTCALL
@@ -47,7 +47,7 @@ IntGetThreadFocusWindow(VOID)
    ThreadQueue = pti->MessageQueue;
    if (!ThreadQueue)
      return NULL;
-   return ThreadQueue->spwndFocus ? UserHMGetHandle(ThreadQueue->spwndFocus) : 0;
+   return UserHMGetHandleSafe(ThreadQueue->spwndFocus);
 }
 
 BOOL FASTCALL IntIsWindowFullscreen(PWND Window)
@@ -598,7 +598,7 @@ co_IntSendActivateMessages(PWND WindowPrev, PWND Window, BOOL MouseActivate, BOO
       co_IntSendMessage( UserHMGetHandle(Window),
                          WM_ACTIVATE,
                          MAKEWPARAM(MouseActivate ? WA_CLICKACTIVE : WA_ACTIVE, (Window->style & WS_MINIMIZE) != 0),
-                        (LPARAM)(WindowPrev ? UserHMGetHandle(WindowPrev) : 0));
+                        (LPARAM)UserHMGetHandleSafe(WindowPrev));
 
       if (Window->style & WS_VISIBLE)
          UpdateShellHook(Window);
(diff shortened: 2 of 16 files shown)


JoachimHenze:
The part that is visible in this PRs diff looks good to me, but:
Warning: There are other places where UserHMGetHandle is called in win32SS/user/*.*, like callproc.c and others, that are not visible in this PRs diff.
Those should be reviewed as well, for anyone reviewing this PR.

TAN-Gaming:
> Warning: There are other places where UserHMGetHandle is called in win32SS/user/_._, like callproc.c and others, that are not visible in this PRs diff.

Looked at them. I think they are OK, no need to change.

TAN-Gaming:
More redundant NULL checks removed.

win32ss/user/ntuser/focus.c
```diff
@@ -1325,9 +1325,9 @@ co_UserSetFocus(PWND Window)
    ThreadQueue = pti->MessageQueue;
    ASSERT(ThreadQueue != 0);
 
-   TRACE("Enter SetFocus hWnd 0x%p pti 0x%p\n",Window ? UserHMGetHandle(Window) : 0, pti );
+   TRACE("Enter SetFocus hWnd 0x%p pti 0x%p\n", UserHMGetHandle(Window), pti );
```
HBelusca, line 1328:
```suggestion
   TRACE("Enter SetFocus hWnd 0x%p pti 0x%p\n", UserHMGetHandle(Window), pti);
```

HBelusca (approved):

win32ss/user/ntuser/focus.c
TAN-Gaming: Done.

learn-more:
I am not sure this is an improvement?

katahiromz (changes requested):
We believe in zero cost in OS development. If you need a NULL check, check it, if not, don't check it. You might be better to split the function macro.

For example, `UserGetHandle` and `UserGetHandleSafe` etc.

HBelusca:
> We believe in zero cost in OS development. If you need a NULL check, check it, if not, don't check it. You might be better to split the function macro.
> 
> For example, `UserHMGetHandle` and `UserGetHandleSafe` etc.

Now I don't know about you, but there were maybe only two places where the UserHMGetHandle was used without checking whether the pointer was null beforehand, while all the others had the check already...

TAN-Gaming:
> UserGetHandleSafe

Good idea.

win32ss/user/ntuser/misc.c
```diff
@@ -308,8 +308,7 @@ NtUserGetThreadState(
          ret = (gpqForeground == GetW32ThreadInfo()->MessageQueue);
          break;
       case THREADSTATE_GETCURSOR:
-         ret = (DWORD_PTR) (GetW32ThreadInfo()->MessageQueue->CursorObject ?
-                            UserHMGetHandle(GetW32ThreadInfo()->MessageQueue->CursorObject) : 0);
+         ret = (DWORD_PTR)UserHMGetHandleSafe(GetW32ThreadInfo()->MessageQueue->CursorObject);
```
katahiromz, line 311: This code will call `GetW32ThreadInfo` twice.

TAN-Gaming: Can inline function fix this?

JoachimHenze: Using the old macro will fix it.

TAN-Gaming: Fixed.

JoachimHenze:
Are there any calls to the old macro UserHMGetHandle() left now?
Why / Why not?
(A legit justification for a "Why" could be, that the NULL check would be redundant at some place then.)

If you compile your win32k.sys with and without your change, does the binary size change in any way?
Which compiler did you use for that test?

TAN-Gaming:
> Are there any calls to the old macro UserHMGetHandle() left now? Why / Why not? (A legit justification for a "Why" could be, that the NULL check would be redundant at some place when using the new inline func.)

Yes, there are many of them left. The reason is it already does the NULL check before calling UserHMGetHandle() .

JoachimHenze:
Okay,
Since you changed from a macro to a function you **wish** to be inlined, I would still be interested in the result of the compilation test and the resulting filesize of win32k.sys before and after the changes. And your used compiler.
I do understand the improvements from code aesthetics point of view, but want to see whether it has any effect or not.
Don't worry I will stay neutral for your PR even if it increases, but it is interesting for me to decide on whether I will ever port your changes back or not.

HBelusca:
Why would you ever port those changes back?

JoachimHenze:
> Why would you ever port those changes back?

For the same reason that you do consider accepting the PR to master: It slightly beautifies the code.
And after a while many new and old implementations will depend on it.
So I will at some point have to decide whether I want to port back the func, or adapt the callers back to old-style for backports.

win32ss/user/ntuser/misc.c
TAN-Gaming: This cannot be inlined because UserHMGetHandle can be used with any object type (PVOID).

TAN-Gaming:
Reverted to macro.

HBelusca:
> Reverted back to macro as it cannot be inlined.

Why can't it be inlined?

win32ss/user/ntuser/misc.c
HBelusca: Note that in case this would be a problem, a fix could be:
```c
case THREADSTATE_GETCURSOR:
{
    PCURICON_OBJECT CursorObject = GetW32ThreadInfo()->MessageQueue->CursorObject;
    ret = (DWORD_PTR)UserHMGetHandleSafe(CursorObject);
    break;
}
```

TAN-Gaming: Fixed,

win32ss/include/ntuser.h
```diff
@@ -228,6 +228,7 @@ typedef struct _PROCMARKHEAD
 } PROCMARKHEAD, *PPROCMARKHEAD;
 
 #define UserHMGetHandle(obj) ((obj)->head.h)
```
JoachimHenze, line 230: Maybe add a comment above to make users aware of the differences:
//use UserHMGetHandleSafe by default and if (obj) has not been checked for NULL already
//use UserHMGetHandle if (obj) has been checked for NULL already, or if (obj) is a function-call, to not call func twice

JoachimHenze, line 230: Maybe add a comment above to make users aware of the differences:
//use UserHMGetHandleSafe by default and if (obj) has not been checked for NULL already
//use UserHMGetHandle if (obj) has been checked for NULL already, or if (obj) is a function-call, to not call func twice

JoachimHenze, line 230: Maybe add a comment at line 229 to make users aware of the differences between both macros:
//use UserHMGetHandleSafe by default and if (obj) has not been checked for NULL already
//use UserHMGetHandle if (obj) has been checked for NULL already, or if (obj) is a function-call, to not call func twice

katahiromz (approved):
LGTM.

win32ss/include/ntuser.h
HBelusca: Then do it with doxygen-style comment:
```c
/**
 * @note
 * Use UserHMGetHandleSafe() by default if obj has not been checked for NULL already.
 * Use UserHMGetHandle() if obj has been checked for NULL already, or if it contains a function-call
 * (to not call that function twice).
 **/
```

TAN-Gaming: Comment added.

HBelusca (approved):

JoachimHenze (approved):

TAN-Gaming:
I've changed my mind. The old style looks somewhat more readable and easier to understand the code.
The better fix is to remove actual redundant checks.

PR #5047, opened 2023-02-05
[WINMM] MMIO_ParseExtA: Don't parse extension if the file name is empty

TAN-Gaming:
## Purpose

JIRA issue: [CORE-18810](https://jira.reactos.org/browse/CORE-18810)

### DIFF
--- a/dll/win32/winmm/mmio.c
+++ b/dll/win32/winmm/mmio.c
@@ -374,6 +374,12 @@ static FOURCC MMIO_ParseExtA(LPCSTR szFileName)
     if (!szFileName)
 	return ret;
 
+#ifdef __REACTOS__
+    /* Check if the file name is empty (CORE-18810) */
+    if (!*szFileName)
+        return ret;
+#endif
+
     /* Find the last '.' */
     extStart = strrchr(szFileName,'.');
 


JoachimHenze:
Can you send this upstream?

KRosUser2:
Tested with build artifact and correctly fixes the regression.

KRosUser2 (approved):

TAN-Gaming:
> Can you send this upstream?

Dunno how to do that.

learn-more (approved):

gonzoMD:
> > Can you send this upstream?
> 
> Dunno how to do that. Where should the patch goes, WineHq or Wine-staging?

Winehq, but dont refer to ReactOS when you send it.

SergeGautherie:
Why not import WINE code, instead of adding a ROS diff?

KRosUser2:
Serge, do you see a similar fix in current Wine master ?

Extravert-ir (changes requested):
This is already fixed in Wine. Please, either import their commits win winmm, or we may merge this change as-is, but leave a comment about this `#ifdef __REACTOS__` block not needed to be synced in future

TAN-Gaming:
superseded by #5050

PR #5048, opened 2023-02-05
[NTUSER] co_UserDestroyWindow: Check Window->pcls is not NULL

TAN-Gaming:
## Purpose

JIRA issue: [CORE-18821](https://jira.reactos.org/browse/CORE-18821)

### DIFF
--- a/win32ss/user/ntuser/window.c
+++ b/win32ss/user/ntuser/window.c
@@ -2881,7 +2881,7 @@ BOOLEAN co_UserDestroyWindow(PVOID Object)
       }
    }
 
-   if (Window->pcls->atomClassName != gpsi->atomSysClass[ICLS_IME])
+   if (Window->pcls && (Window->pcls->atomClassName != gpsi->atomSysClass[ICLS_IME]))
    {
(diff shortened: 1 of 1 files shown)


JoachimHenze (approved):

learn-more (changes requested):
This seems like a hack?
How can a window be destroyed when it has no class?
Then it should not even be created.

KRosUser2:
> This seems like a hack? How can a window be destroyed when it has no class? Then it should not even be created.

Yes but here we have access to a member of Null.  Maybe an ERR message to be added since this is not supposed to happen ?

TAN-Gaming:
> How can a window be destroyed when it has no class?

From JIRA issue: This happens after WM_CREATE message failed.

TAN-Gaming:
@KRosUser2 I agree that an ERR message is better than a BSOD.

yagoulas:
Yes, but why does it happen? Not understanding wtf is going on in our code is recipe for disaster.

TAN-Gaming:
Incorrect fix, closed.

PR #5049, opened 2023-02-05
[SYSSETUP] Fix a space-typo in IDS_WZD_PASSWORDEMPTY for *.rc

JoachimHenze:
## Purpose

Fix a space-typo in IDS_WZD_PASSWORDEMPTY for *.rc
Also update any existing old-school-headers to more compact newstyle.


JoachimHenze:
Please let me know if I should strip "Eric Kohl" and "Filip Navala" as it is highly unlikely that they translated all of those. That header was just copy-pasted in most of the cases, and contains only noise actually.

dll/win32/syssetup/lang/fr-FR.rc
```diff
@@ -213,7 +213,7 @@ BEGIN
     IDS_WZD_NAME "L'installation ne peut pas continuer tant que vous n'avez pas entré votre nom."
     IDS_WZD_SETCOMPUTERNAME "L'installation a échoué lors de la définition du nom de votre ordinateur."
     IDS_WZD_COMPUTERNAME "L'installation ne peut continuer tant que vous n'avez pas entré le nom de votre ordinateur."
-    IDS_WZD_PASSWORDEMPTY "Vous devez entrer un mot de passe !"
+    IDS_WZD_PASSWORDEMPTY "Vous devez entrer un mot de passe!"
```
SergeGautherie, line 216: No, french needs a space.

KRosUser2: No space afaik. Check rules

JoachimHenze: I asked @KRosUser2 who is a native french speaker, and he confirmed that no space would be needed before ! in french.
Only a space before a ":" character is required for french.

JoachimHenze: That being said, ofc I will happily apply the way for french where majority of french native speakers would feel most comfortable with.

SergeGautherie: > No space afaik. Check rules ("espace insécable")

Where are these rules defined/explained?

Check, for example,
https://storylearning.com/learn/french/french-tips/french-punctuation

HBelusca: Guys you're all wrong (except Serge). French puts a space between interrogation point, exclamation point, double-point (:) and semi-colon (;) - The list given by the last link is correct by the way.

HBelusca, line 216:
```suggestion
    IDS_WZD_PASSWORDEMPTY "Vous devez entrer un mot de passe !"
```

HBelusca (changes requested):

dll/win32/syssetup/lang/fr-FR.rc
JoachimHenze: Okay, "2 vs 1 of native french speakers" voted for changing back the space for french. I will do so therefore.

HBelusca (approved):
Now OK; is the file header of the french resource OK though?

JoachimHenze:
> Now OK; is the file header of the french resource OK though?

French had no header at all. And that is fine. Because the headers do not add much value. I updated only those files who had a header already with the wrong outdated style. I do see no urgent need to add any additional bloat into the other files that future developers will have to groom then everytime the rules for that changes. Me touching the headers was mostly motivated to get rid of the redundant GPL-copy-paste-noise everywhere.

PR #5051, opened 2023-02-07
[Scsiport] Return correct status for IOCTL FT_BALANCED_READ_MODE

atharva1910:
JIRA issue: [CORE-18820](https://jira.reactos.org/browse/CORE-18820)
Add support for IOCTL FT_BALANCED_READ_MODE and return STATUS_NOT_IMPLEMENTED

### DIFF
--- a/drivers/storage/port/scsiport/ioctl.c
+++ b/drivers/storage/port/scsiport/ioctl.c
@@ -8,6 +8,7 @@
  */
 
 #include "scsiport.h"
+#include "ntddft.h" // FT_BALANCED_READ_MODE
 
 #define NDEBUG
 #include <debug.h>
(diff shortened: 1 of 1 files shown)


drivers/storage/port/scsiport/ioctl.c
```diff
@@ -538,6 +539,11 @@ ScsiPortDeviceControl(
             status = STATUS_NOT_IMPLEMENTED;
             break;
 
+        case FT_BALANCED_READ_MODE:
+            DPRINT1("FT_BALANCED_READ_MODE unimplemented!\n");
```
JoachimHenze, line 543: Omit the final exclamation mark.

JoachimHenze (changes requested):
To be honest I don't see any gain in this PR at all. It just changes one unimplemented logging into another one. And that by using more code and more includes. Strpping the ! is the minimal thing to make me undo my rejection. I will never give active approval for this PR though.

KRosUser2:
> To be honest I don't see any gain in this PR at all. It just changes one unimplemented logging into another one. And that by using more code and more includes. Strpping the ! is the minimal thing to make me undo my rejection. I will never give active approval for this PR though.

Purpose of my ticket leading to this minimal PR is to add clarity when analysing logs. No functional gain but clearer logging since this IOCTL code is used as a command but not properly managed as unsupported code, leading to "unknown" command received while it's intentionnally used by ReactOS itself.

KRosUser2 (approved):

Extravert-ir:
This IOCTL is targeted to a driver we don't have (ftmgr or volmgr in never versions). So I'm against putting the message anywhere in scsiport/disk/partmgr/whetever driver we currently have in storage stack. Let is stay like it goes right now

Extravert-ir:
@atharva1910 Thanks for the contribution, but it looks like you need to pick some other `starter-project` :)

PR #5052, opened 2023-02-08
[BROWSEUI] Enable address bar focus key (Alt+D)

katahiromz:
## Purpose

Improve keyboard interface usability.
JIRA issue: [CORE-18823](https://jira.reactos.org/browse/CORE-18823)

## Proposed changes

- Fix some `TranslateAcceleratorIO` methods so that `TranslateAcceleratorIO` returns `S_OK` if and only if the message was translated.
- Implement `CAddressBand::TranslateAcceleratorIO` to process `Alt+D`.
- Add `CAddressBand::GetFocusKey` helper function and `CAddressBand::m_chFocusKey` member.
- Fix `CShellBrowser::TranslateAcceleratorSB` to process the key messages on the client bars.

## TODO

- [x] Do small tests.

## Screenshot
VirtualBox_ReactOS_08_02_2023_12_41_26
The key combination `Alt+D` is working.


dll/win32/browseui/lang/de-DE.rc
```diff
@@ -330,6 +330,7 @@ BEGIN
     IDS_BACK "Zurück zu %s"
     IDS_FORWARD "Vorwärts zu %s"
     IDS_FOLDER_OPTIONS "Ordneroptionen"
+    IDS_ADDRESSBARFOCUSKEY "D"
```
JoachimHenze, line 333: Use "S" as the german accelerator.

JoachimHenze (changes requested):
Use "S" as the german accelerator. Aside from that LGTM.

dll/win32/browseui/lang/de-DE.rc
katahiromz: OK.

dll/win32/browseui/addressband.cpp
```diff
@@ -259,24 +259,52 @@ HRESULT STDMETHODCALLTYPE CAddressBand::HasFocusIO()
...
 
-HRESULT STDMETHODCALLTYPE CAddressBand::TranslateAcceleratorIO(LPMSG lpMsg)
+WCHAR CAddressBand::GetAddressBarFocusKey()
 {
-    if (lpMsg->hwnd == fEditControl)
+    static WCHAR s_szKey[2] = L"";
+    if (s_szKey[0] == UNICODE_NULL)
     {
-        switch (lpMsg->message)
+        if (!LoadStringW(_AtlBaseModule.GetResourceInstance(), IDS_ADDRESSBARFOCUSKEY,
```
learn-more, line 267: Are you sure that this should be special cased like this,
and not just be a 'regular' accelerator?

katahiromz: image
`WM_SYSKEYDOWN` hits `LoadStringW` and there is `"D"` resource string in `shell32`.

katahiromz: The normal accelerator generates `WM_COMMAND` message.

katahiromz: image
`browseui` doesn't have `RT_ACCELERATOR`.

image
`shell32` doesn't have `Alt+D` accelerator.

katahiromz: We don't have `BROWSEUI!MLLoadStringW`.

JoachimHenze (approved):

dll/win32/browseui/basebarsite.cpp
```diff
@@ -287,12 +287,11 @@ HRESULT STDMETHODCALLTYPE CBaseBarSite::TranslateAcceleratorIO(LPMSG lpMsg)
 {
     if (!fCurrentActiveBar)
     {
-        if (lpMsg)
+        if (lpMsg && TranslateMessage(lpMsg))
```
HBelusca, line 290: Why is this fix needed? (that one and the similar one below)

HBelusca (changes requested):
I don't fully think all aspects of this PR are correct.
Indeed, the accelerator "d" is the one that corresponds to the underlined one in the "Address" label.
And this should act as usual window/dialog navigation: The focus (here, given to the address bar) is given to the **next control** that follows the static label "Address".
While this PR obviously hardcodes only some aspects of it, namely re-hardcoding the keyboard shortcut "d", etc.

When a standard dialog has two controls: the static text "A&ddress", and following it a combobox (or an editbox) (and other controls as well), then by pressing "alt+d" the focus will be given to the control following the static text that contains the keyboard accelerator "d", and this will be the combobox (in this example). No extra heavy code will be necessary in the app.

dll/win32/browseui/basebarsite.cpp
katahiromz: `TranslateAcceleratorIO` has to return `S_OK` if and only if the message was translated. There is some chances that the message isn't translated.

dll/win32/browseui/addressband.cpp
```diff
@@ -259,24 +259,52 @@ HRESULT STDMETHODCALLTYPE CAddressBand::HasFocusIO()
...
+                         s_szKey, _countof(s_szKey)))
         {
+            s_szKey[0] = L'D'; /* Alt+D */
+        }
+    }
+    return s_szKey[0];
```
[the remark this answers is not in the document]
katahiromz: A private member `WCHAR m_chAddressBarFocusKey` is added.

katahiromz:
The "Address" label is implemented as the inner component of Rebar control, thus it is not a `STATIC` control.

katahiromz:
Now `IDS_ADDRESSBARFOCUSKEY` is deleted.

dll/win32/browseui/basebarsite.cpp
katahiromz: See also `CInternetToolbar::TranslateAcceleratorIO`.

```cxx
HRESULT STDMETHODCALLTYPE CInternetToolbar::TranslateAcceleratorIO(LPMSG lpMsg)
{
    HRESULT hr = S_FALSE;

    if (fMenuBar)
        hr = IUnknown_TranslateAcceleratorIO(fMenuBar, lpMsg);
    if (hr == S_OK)
        return hr;

    if (fControlsBar)
        hr = IUnknown_TranslateAcceleratorIO(fControlsBar, lpMsg);
    if (hr == S_OK)
        return hr;

    if (fNavigationBar)
        hr = IUnknown_TranslateAcceleratorIO(fNavigationBar, lpMsg);
    if (hr == S_OK)
        return hr;

    return S_FALSE;
}
```

dll/win32/browseui/addressband.h
```diff
@@ -50,6 +50,8 @@ class CAddressBand :
 private:
     void FocusChange(BOOL bFocus);
     void CreateGoButton();
+    WCHAR GetFocusKey();
+    WCHAR m_chFocusKey;
```
HBelusca, line 54: Please move the member up with the other ones.

dll/win32/browseui/addressband.cpp
```diff
@@ -259,24 +260,53 @@ HRESULT STDMETHODCALLTYPE CAddressBand::HasFocusIO()
...
+        return m_chFocusKey;
+
+    CStringW strLabel(MAKEINTRESOURCEW(IDS_ADDRESSBANDLABEL));
+    INT ich = strLabel.Find(L'&');
+    if (ich == -1)
+        m_chFocusKey = L'D'; /* Alt+D */
+    else
+        m_chFocusKey = (WCHAR)(INT_PTR)CharUpperW((LPWSTR)(INT_PTR)strLabel[ich + 1]);
+
+    return m_chFocusKey;
```
HBelusca, line 275: Ah yes, whatever way we do, because the key is cached and not re-updated everytime, when one changes the UI language of ReactOS at runtime (with an explorer window opened), say, from english to german, the UI would change (and the "Address" label would change to the german corresponding one), but the cached accelerator would still be (Alt+)D instead of another one. This is a bug.

dll/win32/browseui/addressband.h
katahiromz: OK.

dll/win32/browseui/addressband.cpp
katahiromz: I tried to fix this UI bug by using newly-added `m_strAddressLabel`.

HBelusca:
I tried to investigate a little bit more on Win2k3 and Win7. Here are my remarks:
1. In both OSes, pressing Ctrl+F4 or F4, also puts the focus on the address bar, but also expands the combobox. This seems to be the most robust way since it doesn't depend on any sort of localization. I was able to find the candidate keyboard accelerators (at least, the one for Ctrl+F4) in (the MUI file of) Win7's explorerframe.dll: `VK_F4, 40993, NOINVERT, CONTROL, VIRTKEY` ; however I wasn't able to find the equivalent on Win2k3 (neither in explorer.exe, nor browseui.dll nor browselc.dll).
2. On both Win2k3 and Win7, explorer.exe has a "suspiciously-looking" keyboard accelerator, see screenshots:
image
I suppose this is how they handle the Alt+D shortcut. If this is indeed the correct one, what @JoachimHenze should see there on his Win2k3 german instead, is a shortcut for Alt+S. Please confirm :)

katahiromz:
Something is broken. Retired.

JoachimHenze:
> I suppose this is how they handle the Alt+D shortcut. If this is indeed the correct one, what @JoachimHenze should see there on his Win2k3 german instead, is a shortcut for Alt+S. Please confirm :)

hbelusca, I don't have a german 2k3sp2.
I do have an english 2k3sp2, which uses Alt+D. Confirmed.
And I do have a german XPSP3, which uses Alt+S. Confirmed.
And I am sure, that a german 2k3sp2 (if such a thing exists, e.g. the MUI package) would not use Alt+D for german, because that is a very well known and reserved thing for the file (&Datei) toolbar menu.

HBelusca:
@JoachimHenze you can readily look with any resource hacker the explorer.exe in your german xpsp3 and see whether its accelerator table contains the Alt+"S" or Alt+"D" . But yes, definitively re. the &Datei thing, they had to use something different.

PR #5053, opened 2023-02-08
UniATA: Fix Memory leak while claiming PCI IDE controller

atharva1910:
HalAssignSlotResources will [allocate a new resource list which the caller must free](https://git.reactos.org/?p=reactos.git;a=blob;f=hal/halx86/legacy/bus/pcibus.c;h=1098eb0629ca1d5df47dd793fa7e6f141fa82b8c;hb=HEAD#l862)

According to the [code comment](https://git.reactos.org/?p=reactos.git;a=blob;f=drivers/storage/ide/uniata/id_probe.cpp;h=1ca49d248ecdfc0b2f3379849776cf2f8a6b3a85;hb=HEAD#l1984), in some cases HalAssignSlotResources fails on some HW so a resource list is pre-allocated

This fix plugs the memory leak by freeing both the resource lists if allocated

### DIFF
--- a/drivers/storage/ide/uniata/id_probe.cpp
+++ b/drivers/storage/ide/uniata/id_probe.cpp
@@ -1942,6 +1942,7 @@ UniataClaimLegacyPCIIDE(
 {
     NTSTATUS status;
     PCM_RESOURCE_LIST resourceList = NULL;
+    PCM_RESOURCE_LIST oldResList   = NULL;
     UNICODE_STRING devname;
 
     KdPrint2((PRINT_PREFIX "UniataClaimLegacyPCIIDE:\n"));
@@ -1967,11 +1968,10 @@ UniataClaimLegacyPCIIDE(
     if (!resourceList) {
         KdPrint2((PRINT_PREFIX "!resourceList\n"));
         status = STATUS_INSUFFICIENT_RESOURCES;
-del_do:
-        IoDeleteDevice(BMList[i].PciIdeDevObj);
-        BMList[i].PciIdeDevObj          = NULL;
-        return status;
+        goto del_do;
     }
+
+    oldResList = resourceList;
 
     RtlZeroMemory(
         resourceList,
@@ -2001,14 +2001,22 @@ del_do:
 
     if (!NT_SUCCESS(status)) {
         KdPrint2((PRINT_PREFIX "HalAssignSlotResources failed %#x\n", status));
-        // this is always deallocated inside HalAssignSlotResources() implementation
-        //ExFreePool(resourceList);
         goto del_do;
     }
+
+    ExFreePool(oldResList);
+    ExFreePool(resourceList);
 
     KdPrint2((PRINT_PREFIX "ok %#x\n", status));
     BMList[i].ChanInitOk |= 0x80;
 
+    return status;
+
+del_do:
+    IoDeleteDevice(BMList[i].PciIdeDevObj);
+    BMList[i].PciIdeDevObj          = NULL;
+    if (oldResList)
+        ExFreePool(oldResList);
     return status;
 } // end UniataClaimLegacyPCIIDE()
 


HBelusca:
> HalAssignSlotResources will [allocate a new resource list which the caller must free](https://git.reactos.org/?p=reactos.git;a=blob;f=hal/halx86/legacy/bus/pcibus.c;h=1098eb0629ca1d5df47dd793fa7e6f141fa82b8c;hb=HEAD#l862)

Don't be fooled necessarily by our own code (which may be buggy or leak memory). If HalAssignSlotResources failed and no stuff has been added to the would-be-returned resource list, HalAssignSlotResources must free the resource list it allocated, instead of returning it uninitialized.

ericsomto (approved):

atharva1910:
>  HalAssignSlotResources must free the resource list it allocated, instead of returning it uninitialized.

That's correct, In the failure case we do not free the new list, just the old list

drivers/storage/ide/uniata/id_probe.cpp
```diff
@@ -1942,6 +1942,7 @@ UniataClaimLegacyPCIIDE(
 {
     NTSTATUS status;
     PCM_RESOURCE_LIST resourceList = NULL;
+    PCM_RESOURCE_LIST oldResList   = NULL;
```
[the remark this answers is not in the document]
atharva1910: Ah apologies this is my first time contributing, I do have some questions though

1. This fix is not specific to ReactOS, The memory leaks would be present even if this code were to run on Windows NT or ReactOS. Do we add #ifdef to 3rd party drivers for every change? If it was something related specific to ReactOS I would understand but this is just a dangling pointer which is freed.

2.  Unless we push these changes upstream (to UniATA) this part of the code is now part of ReactOS yes? Why do we add #ifdef(s)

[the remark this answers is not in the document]
JoachimHenze: The best thing to do would be upstream integration ofc, especially if the drivers failed similarly on Windows. We would then not need the ifdefs on our side.
But often upstream integration is taking considerable amount of time, or isn't possible at all (uniata maintainer named "Altar" seems to be not that active anymore).
In such cases we do use the #ifdef guards to highlight our diff to the upstream repo.
Hoping for them to be a temporary thing. But more often than we like, they are getting a permanent thing.
That is still better than not adding them.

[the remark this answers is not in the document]
HBelusca: In any case, Alter is alive because http://alter.org.ua/ru/  shows updates of some of his stuff on this year (2023).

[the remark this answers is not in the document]
HBelusca: Cc @Alter-1 could you check this UniATA bugfix? :)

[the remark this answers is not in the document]
roytam1: > In any case, Alter is alive because http://alter.org.ua/ru/ shows updates of some of his stuff on this year (2023).
> Cc @Alter-1 could you check this UniATA bugfix? :)

I wonder if this is not on his priority list since there is war and electricity is not stable in his state.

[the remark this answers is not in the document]
atharva1910: Done! The code is now under `#ifdef __REACTOS__` block

drivers/storage/ide/uniata/id_probe.cpp
```diff
@@ -2006,10 +2017,24 @@ UniataClaimLegacyPCIIDE(
...
+
     KdPrint2((PRINT_PREFIX "ok %#x\n", status));
     BMList[i].ChanInitOk |= 0x80;
 
     return status;
+
+#if __REACTOS__
+del_do:
+        IoDeleteDevice(BMList[i].PciIdeDevObj);
+        BMList[i].PciIdeDevObj          = NULL;
```
JoachimHenze, line 2033: Omit that space padding before the = please. Only one space beforehand.

drivers/storage/ide/uniata/id_probe.cpp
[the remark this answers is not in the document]
JoachimHenze: Thanks. We will wait one or two weeks, hoping for @Alter-1 to show up. Then we will merge it most likely.
Question: What are the **exact reproduction steps** to make it leak, and how did you confirm it to stop leaking then?

drivers/storage/ide/uniata/id_probe.cpp
```diff
@@ -2006,10 +2017,24 @@ UniataClaimLegacyPCIIDE(
...
+    ExFreePool(resourceList);
+    ExFreePool(oldResList);
+#endif
+
     KdPrint2((PRINT_PREFIX "ok %#x\n", status));
     BMList[i].ChanInitOk |= 0x80;
 
     return status;
+
+#if __REACTOS__
```
JoachimHenze, line 2030: why #ifdef at the other locations but only #if here?

katahiromz:
```suggestion
#ifdef __REACTOS__
```

drivers/storage/ide/uniata/id_probe.cpp
atharva1910: Done 👍

drivers/storage/ide/uniata/id_probe.cpp
atharva1910: Done 👍

drivers/storage/ide/uniata/id_probe.cpp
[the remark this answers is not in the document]
atharva1910: > Question: What are the exact reproduction steps to make it leak, and how did you confirm it to stop leaking then?

To reproduce:  Create VM with storage attached to PCI IDE controller (PXII3/4 under VBox). Memory should leak while UniATA is loaded and enumerates on the PCI bus

To confirm leak is plugged: Ran under WinDBG, observing the dangling pointer and the new pointer allocated by HalAssignSlotResources, ensure both pointers are freed

drivers/storage/ide/uniata/id_probe.cpp
```diff
@@ -1,3 +1,4 @@
+
```
SergeGautherie, line 1: Assumed unwanted change?

drivers/storage/ide/uniata/id_probe.cpp
```diff
@@ -2006,10 +2018,24 @@ UniataClaimLegacyPCIIDE(
...
+#endif
+
     KdPrint2((PRINT_PREFIX "ok %#x\n", status));
     BMList[i].ChanInitOk |= 0x80;
 
     return status;
+
+#ifdef __REACTOS__
+del_do:
+        IoDeleteDevice(BMList[i].PciIdeDevObj);
```
SergeGautherie, line 2033: Reduce indentation.

drivers/storage/ide/uniata/id_probe.cpp
```diff
@@ -1943,6 +1944,9 @@ UniataClaimLegacyPCIIDE(
     NTSTATUS status;
     PCM_RESOURCE_LIST resourceList = NULL;
     UNICODE_STRING devname;
+#ifdef __REACTOS__
```
SergeGautherie, line 1947: Move this block 1 line up?

drivers/storage/ide/uniata/id_probe.cpp
```diff
@@ -1967,16 +1971,24 @@ UniataClaimLegacyPCIIDE(
     if (!resourceList) {
         KdPrint2((PRINT_PREFIX "!resourceList\n"));
         status = STATUS_INSUFFICIENT_RESOURCES;
+#ifdef __REACTOS__
+        goto del_do;
+#else
```
SergeGautherie, line 1976: Why "move" this block, instead of simply updating it here?

drivers/storage/ide/uniata/id_probe.cpp
JoachimHenze: Could be a change in the ByteOrderMark? We do want our files to be UTF8 without BOM.
Please make sure to only edit with editors that won't break the format. Notepad++ is fine. Notepad is not,

drivers/storage/ide/uniata/id_probe.cpp
HBelusca: To group it together with the other resourceList?

drivers/storage/ide/uniata/id_probe.cpp
HBelusca: So you think moving the (now placed-at-the-end) new cleanup code back there, could be better?
Would the extra `ExFreePool(oldResList)` call be needed there then?

atharva1910: I feel its much cleaner to have the cleaner code at the bottom of the function. Jumping randomly to the middle of the function to cleanup is not "pretty" tbh. 
The code was originally changed without the `#ifdef`s so it looked much cleaner. Now its just ugly, but I guess we can't avoid it. Hopefully the UniATA maintainer approves these changes so we can remove the `#ifdef`

> Would the extra ExFreePool(oldResList) call be needed there then?

Yes if the HalAssignSlotResources  fails we must still cleanup the "old" allocated resource list. The new resource list will have be cleaned up by HalAssignSlotResources but the old resource list must be cleaned by us, since its allocated in this function.

drivers/storage/ide/uniata/id_probe.cpp
atharva1910: I wanted to have the `#ifdef` declaration block after the 'old' local variables, having it between the local variables seems ugly(er).
Changed to suit both our requirments

drivers/storage/ide/uniata/id_probe.cpp
atharva1910: Done :+1:

drivers/storage/ide/uniata/id_probe.cpp
atharva1910: Strange, removed :+1:

Extravert-ir (approved):

drivers/storage/ide/uniata/id_probe.cpp
JoachimHenze: No. Your new solution is worse to both: Your old solution and also worse than what Serge requested.
I do request that to be changed, and I think Serges proposal was ok.
I do reject your current solution as sicne you moved, now an ifdef guard is missing.

JoachimHenze (changes requested):
I do request you to change your current solution of the "Move this block 1 line up?" dicussion. It is definitely worse than Serges proposal, and also worse to what you had before!

drivers/storage/ide/uniata/id_probe.cpp
JoachimHenze: I do understand your motivation not wanting to not jump into the middle of the func for cleanup. A good rule when coding is to "not surprise other developers", e.g.: with your code-flow.
Resolved.

drivers/storage/ide/uniata/id_probe.cpp
SergeGautherie: > ugly(er)

Less "beautiful" as a ROS diff, but more "logical" as an upstream fix.

drivers/storage/ide/uniata/id_probe.cpp
SergeGautherie: > not "pretty"

I agree for project own code, but, for upstream code, minimal diff is preferable.

> UniATA maintainer approves

That would be great, yet I do not expect it: no answer from him for a long time...

Extravert-ir:
Guys, you're wasting too much time here. This fixes the issue, and is OK for merging.
UniATA is not the best codebase in the world and nitpicks like in this discussion will not make it better.

drivers/storage/ide/uniata/id_probe.cpp
```diff
@@ -1967,16 +1970,24 @@ UniataClaimLegacyPCIIDE(
...
         resourceList,
         sizeof(CM_RESOURCE_LIST));
 
+#ifdef __REACTOS__
+    oldResList = resourceList;
+#endif
+
     // IoReportDetectedDevice() should be used for WDM OSes
 
     // TODO: check if resourceList is actually used inside HalAssignSlotResources()
```
Extravert-ir, line 1993: OK, I've looked through this more careful and here is the thing: `resourceList` is not actually used inside `HalAssignSlotResources`. So as we're touching this code already, let's just remove all that resource allocation (the one which is now `oldResList`) stuff as not needed.

I don't see it being used on my version of Windows as well.

Extravert-ir (changes requested):

drivers/storage/ide/uniata/id_probe.cpp
atharva1910: Thats correct, `resourceList` is not used inside `HalAssignSlotResources` but according to the [code comment](https://git.reactos.org/?p=reactos.git;a=blob;f=drivers/storage/ide/uniata/id_probe.cpp;h=1ca49d248ecdfc0b2f3379849776cf2f8a6b3a85;hb=HEAD#l1984), in some cases HalAssignSlotResources fails on some HW so a resource list is pre-allocated.
Now I would love to remove it, I really would but that would require extensive testing on all the supported hardware (virtual or otherwise). Which can be taken up as a task/bug later.
So for now I have settled on cleaning up the leaking memory

drivers/storage/ide/uniata/id_probe.cpp
atharva1910: > I do reject your current solution as since you moved, now an ifdef guard is missing.
Um I have moved one of the "old" declarations (resouceList) down so that it is bundled up with the other resourceList as was suggested.
The new declaration (oldResList) is still guarded by an `ifdef`. I'm not sure I understand the question here.

If variable ordering is a concern here (and I mean really?). If variable ordering is a concern here. Having a new variable between the old ones re-orders the variables anyways. I mean its variables

drivers/storage/ide/uniata/id_probe.cpp
HBelusca: Indeed, what's actually "used" by HalAssignSlotResources, is the pointer to pointer to the resource list, so that it knows where to store the pointer to the new allocated list. So, strange, but all the initialization code for resourceList done just before HalAssignSlotResources in principle is thus "dead" code.

drivers/storage/ide/uniata/id_probe.cpp
atharva1910: Alright, Alright. I moved it back

drivers/storage/ide/uniata/id_probe.cpp
atharva1910: Yep. Its mostly some other issue and not if the resource list is pre-allocated or not but I don't want to break the code

PR #5054, opened 2023-02-09
[NTUSER] co_UserDestroyWindow: Validate window before destroying it

TAN-Gaming:
## Purpose

JIRA issue: [CORE-18821](https://jira.reactos.org/browse/CORE-18821)

## Proposed changes
Do window validation for some functions in win32k that call co_UserDestroyWindow directly.

Thanks @sdever and @yagoulas for the help.

### DIFF
--- a/win32ss/user/ntuser/window.c
+++ b/win32ss/user/ntuser/window.c
@@ -2855,6 +2855,15 @@ BOOLEAN co_UserDestroyWindow(PVOID Object)
 
    ASSERT_REFS_CO(Window); // FIXME: Temp HACK?
 
+   /* NtUserDestroyWindow does check if the window has already been destroyed
+      but co_UserDestroyWindow can be called from more paths which means
+      that it can also be called for a window that has already been destroyed. */
+   if (!IntIsWindow(UserHMGetHandle(Window)))
+   {
+      TRACE("Tried to destroy a window twice\n");
+      return TRUE;
+   }
+
    hWnd = Window->head.h;
    ti = PsGetCurrentThreadWin32Thread();
 


yagoulas:
Can you also add a comment why this is needed? NtUserDestroyWindow does check if the window has already been destroyed but co_UserDestroyWindow can be called from more paths which means that it can also be called for a window that has already been destroyed.

TAN-Gaming:
> Can you also add a comment why this is needed? NtUserDestroyWindow does check if the window has already been destroyed but co_UserDestroyWindow can be called from more paths which means that it can also be called for a window that has already been destroyed.

Comment added.

win32ss/user/ntuser/window.c
```diff
@@ -2855,6 +2855,15 @@ BOOLEAN co_UserDestroyWindow(PVOID Object)
...
+   /* NtUserDestroyWindow does check if the window has already been destroyed
+      but co_UserDestroyWindow can be called from more paths which means
+      that it can also be called for a window that has already been destroyed. */
+   if (!IntIsWindow(UserHMGetHandle(Window)))
+   {
+      TRACE("Tried to call co_UserDestroyWindow() twice\n");
+      return TRUE;
+   }
+
    hWnd = Window->head.h;
```
HBelusca, line 2867: Please note that in your PR #5045 this could have become a `UserHMGetHandle(Window);` .

TAN-Gaming: I'll update this PR later if PR [#5045](https://github.com/reactos/reactos/pull/5045) is merged.

TAN-Gaming: For small changes, I think updating PR is easier than opening a new PR.

Since this PR fixes the bug, I'll update PR [#5045](https://github.com/reactos/reactos/pull/5045) after this PR is merged or open a new PR if PR [#5045](https://github.com/reactos/reactos/pull/5045) is merged before this PR.

TAN-Gaming: Not sure if a NULL check is needed here, if not then using UserHMGetHandle macro is fine.

sdever: This patch looks like a hack for me, the problem seems to be a lot deeper: this patch doesn't prevent use-after-free scenarios, where kernel code uses invalid WND pointer destroyed by co_UserDestroyWindow.
For instance, the first co_UserDestroyWindow call (for instance by DestroyWindow) will then really destroy the window, leading to huge use-after-frees when returning to kernel-mode code which uses now-invalid PWND pointer. A proper fix would be to ensure nobody holds a reference to the window before really destroying the window, and only set a status flag like WINDOW_DESTROYING (which is more or less refcounting), or checking if the window handle still exists after every user-mode call.

yagoulas: WND pointers are refcounted and they are freed only after all references are released. There is a problem that the class of the window can be NULL as it is right now, but win32k has the tools to prevent PWND from getting freed like that. Generally speaking whenever we call a co_ function (co_ functions can release the user lock and do a callback to user mode), object pointers can get invalid. This is correct. This is why before calling a co_ function we need to call UserRefObjectCo for all the user objects that we want to be still alive after a call to a co_ function. Note: in windows these functions have the prefix xxx in their names.

TAN-Gaming: > There is indeed a problem that the class of the window can be NULL as it is right now.

This is a use-after-free problem, the class is set to NULL by the first user-mode DestroyWindow call (WM_CREATE message failed)
https://github.com/reactos/reactos/blob/eb7550f76730105582625e84210a376cf676c112/win32ss/user/ntuser/window.c#L764

then co_UserCreateWindowEx tried to destroy the same window again by calling co_UserDestroyWindow
https://github.com/reactos/reactos/blob/eb7550f76730105582625e84210a376cf676c112/win32ss/user/ntuser/window.c#L2581

The problem is co_UserDestroyWindow tried to access the destroyed window pointer without doing any validation.

katahiromz (approved):

HBelusca (approved):

JoachimHenze (approved):

PR #5058, opened 2023-02-09
[BROWSEUI] Explorer: Show context menu on Right-Clicking icon of title bar

katahiromz:
## Purpose

JIRA issue: [CORE-18827](https://jira.reactos.org/browse/CORE-18827)

## Proposed changes

- Handle `WM_CONTEXTMENU` message for Right-Clicking the icon of title bar.
- Delete unwanted menu items.
- Display the context menu of the folder.

## TODO

- [x] Implement perfectly.
- [ ] Do tests.
- [ ] Fix bugs.
- [x] Vote.

## Comparison

Win2k3 English:
VirtualBox_Windows Server 2003_09_02_2023_17_07_18

BEFORE:
before

AFTER:
after


dll/win32/browseui/shellbrowser.cpp
```diff
@@ -3551,6 +3554,60 @@ LRESULT CShellBrowser::RelayMsgToShellView(UINT uMsg, WPARAM wParam, LPARAM lPar
...
+    HRESULT hr;
+    LPCITEMIDLIST pidlChild;
+    CComPtr<IShellFolder> pFolder;
+    CComPtr<IContextMenu> pContextMenu;
+
+    hr = SHBindToParent(fCurrentDirectoryPIDL, IID_IShellFolder, (void **)&pFolder, &pidlChild);
+    if (FAILED_UNEXPECTEDLY(hr))
+        return hr;
+
+    hr = pFolder->GetUIObjectOf(m_hWnd, 1, &pidlChild, IID_IContextMenu, NULL, (void **)&pContextMenu);
```
learn-more, line 3568: Shouldn't you call this on `fCurrentShellFolder` instead?

The screenshots are a bit confusing to me, because they are mixed japanese (which I can't read) and english.
I am not convinced that the menu that is shown when clicking the top left icon is 100% the same as the folder menu.

btw, on my Win10 machine when I right-click that icon in explorer I get the 'normal' restore/move/etc menu.

katahiromz: Deleted Japanese screenshots.

roytam1: > btw, on my Win10 machine when I right-click that icon in explorer I get the 'normal' restore/move/etc menu.

this is removed since vista.

katahiromz:
after
Let's go!

katahiromz:
Now started voting:
https://chat.reactos.org/reactos/pl/gsghurcu9pg1urnjt49mayyp9e

dll/win32/browseui/shellbrowser.cpp
tkreuzer: On Windows 2003 there is left- and right-click, with left-click being the restore/move/etc menu and right-click being the close/search/explore/etc menu. On Windows 10 the right-click menu is the same as the left-click one.

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -3551,6 +3554,151 @@ LRESULT CShellBrowser::RelayMsgToShellView(UINT uMsg, WPARAM wParam, LPARAM lPar
...
+    if (FAILED_UNEXPECTEDLY(hr))
+        return hr;
+    hr = pFolder->GetUIObjectOf(m_hWnd, 1, &pidlChild, IID_IContextMenu, NULL, (void **)&pContextMenu);
+    if (FAILED_UNEXPECTEDLY(hr))
+        return hr;
+
+    // Query context menu
+#define CMDID_FIRST 0x1000
+#define CMDID_LAST  0x7000
+#define CMDID_CLOSE 0x7001
```
tkreuzer, line 3625: Shouldn't these go to a header?

katahiromz: No, these are local definitions.

dll/win32/browseui/shellbrowser.cpp
JoachimHenze: Thank you Timo. "On Windows 2003 there is left- and right-click, with left-click being the restore/move/etc menu and right-click being the close/search/explore/etc menu." <- That is also exactly the same for my XPSP3.
Katayama, what is the result after your PR for the **left**-click? Did that change in any way by your PR?

katahiromz: I think Left-Clicking won't be affected in this PR. Dragging as a link can be implemented in later work.

JoachimHenze:
Why not adding the "Map network drive" and "Disconnect network drive" as well. This is imho the most useful shortcuts in this menu. Especially for those who prefer to disable the toolbar. By implementing those two as well, you might even be able to delete your "DeleteDuplicatedSeparators"-function. That part looks a bit hand-knitted to me.

KRosUser2:
Is your "properties" menu action effective ? Or affected by this bug https://jira.reactos.org/browse/CORE-18585 and needing PR https://github.com/reactos/reactos/pull/5011 ?

katahiromz:
> Why not adding the "Map network drive" and "Disconnect network drive" as well.

That depends on the current implementation of `CDrivesFolder::CallBack`. See below:

Win2k3:
win2k3
ReactOS:
ros

katahiromz:
> Is your "properties" menu action effective ?

https://user-images.githubusercontent.com/2107452/218068557-a5ca3a90-9648-4293-a1dc-abc49b13a815.mp4
Sorry, nope. I will fix it before merge.

JoachimHenze:
> > Why not adding the "Map network drive" and "Disconnect network drive" as well.
> 
> That depends on the current implementation of `CDrivesFolder::CallBack`. See below:

I see. So your point is that this would have to be implemented at another place, and therefore should be split off from this PR. I do understand.
I guess I will approve as soon as the "properties" will work. Hoping for others maybe to give any suggestions about how/whether this PRs code could be written in a slightly more compact form.

dll/win32/browseui/lang/bg-BG.rc
```diff
@@ -300,6 +300,7 @@ STRINGTABLE
 BEGIN
     IDS_ADDRESSMENUTEXT "&Address"
     IDS_HISTORYTEXT "&Дневник\tCtrl+H"
+    IDS_CLOSE "&Close\tAlt+F4"
```
JoachimHenze, line 303: Is it possible to extract that string programmatically at runtime from a real system-menu, instead of adding a duplicate of it into the browseui resources?

katahiromz: Possible. With no check:
```c
HMENU hSysMenu = ::GetSystemMenu(hWnd, FALSE);
MENUITEMINFOW mii = { sizeof(mii), MIIM_TYPE };
WCHAR szText[64];
mii.dwTypeData = szText;
mii.cch = _countof(szText);
GetMenuItemInfoW(hSysMenu, &mii);
... // Use szText
```

JoachimHenze: To me that looks superior, as it is shorter and would improve the locality of this menus implementation, which requires a lot of "manual work" anyway in the code. Would also reduce translation-follow-up-commits. Do you agree?

katahiromz: Fixed.

JoachimHenze: Thanks. Looks much better already.
Introduction of define CMDID_CLOSE is inevitable?

katahiromz: still-working
Confirmed still working.

learn-more:
> Wait... This PR is blocked by #5011 . Wait and see.

As commented in #5011 already, that PR is incorrect.

HBelusca:
I wonder why this feature even exists on Windows (by the way, it has been removed in Vista+).

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -350,6 +350,7 @@ class CShellBrowser :
     HRESULT UpdateUpState();
     void UpdateGotoMenu(HMENU theMenu);
     void UpdateViewMenu(HMENU theMenu);
+    HRESULT ShowSysMenuContextMenu(INT xPos, INT yPos);
```
HBelusca, line 353: Or "ShowSysContextMenu" would be shorter.

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -3551,6 +3554,156 @@ LRESULT CShellBrowser::RelayMsgToShellView(UINT uMsg, WPARAM wParam, LPARAM lPar
...
+
+    // Get 'Close' menu item string from system menu
+    WCHAR szClose[128];
+    HMENU hSysMenu = ::GetSystemMenu(m_hWnd, FALSE);
+    MENUITEMINFOW mii = { sizeof(mii), MIIM_TYPE };
+    mii.dwTypeData = szClose;
+    mii.cch = _countof(szClose);
+    if (!::GetMenuItemInfoW(hSysMenu, SC_CLOSE, FALSE, &mii))
+        StringCchCopyW(szClose, _countof(szClose), L"&Close\tAlt+F4");
+    // Insert 'Close' menu item and make it default
```
HBelusca, line 3649: Adding some spacing to avoid the big-messy-looking code blob effect...
```suggestion

    // Insert 'Close' menu item and make it default
```

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -3551,6 +3554,156 @@ LRESULT CShellBrowser::RelayMsgToShellView(UINT uMsg, WPARAM wParam, LPARAM lPar
...
+    mii.fMask = MIIM_ID | MIIM_FTYPE | MIIM_STATE | MIIM_STRING | MIIM_BITMAP;
+    mii.fType = MFT_STRING;
+    mii.fState = MFS_DEFAULT;
+    mii.wID = SC_CLOSE;
+    mii.dwTypeData = szClose;
+    mii.hbmpItem = HBMMENU_POPUP_CLOSE;
+    ::InsertMenuItemW(hMenu, 0, TRUE, &mii);
+    // Insert a separator
```
HBelusca, line 3657:
```suggestion

    // Insert a separator
```

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -3551,6 +3554,156 @@ LRESULT CShellBrowser::RelayMsgToShellView(UINT uMsg, WPARAM wParam, LPARAM lPar
...
+    mii.fMask = MIIM_FTYPE;
+    mii.fType = MFT_SEPARATOR;
+    ::InsertMenuItemW(hMenu, 1, TRUE, &mii);
+    // No left-side space for checkmarks
```
HBelusca, line 3661:
```suggestion

    // No left-side space for checkmarks
```

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -3551,6 +3554,156 @@ LRESULT CShellBrowser::RelayMsgToShellView(UINT uMsg, WPARAM wParam, LPARAM lPar
...
+    MENUINFO mi = { sizeof(mi), MIM_STYLE };
+    ::GetMenuInfo(hMenu, &mi);
+    mi.dwStyle |= MNS_NOCHECK;
+    ::SetMenuInfo(hMenu, &mi);
+
+    // Display the menu
+    ::SetForegroundWindow(m_hWnd);
+    UINT id = ::TrackPopupMenu(hMenu, TPM_LEFTALIGN | TPM_RETURNCMD | TPM_RIGHTBUTTON,
+                               xPos, yPos, 0, m_hWnd, NULL);
+    // Action!
```
HBelusca, line 3671:
```suggestion
```

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -3551,6 +3554,156 @@ LRESULT CShellBrowser::RelayMsgToShellView(UINT uMsg, WPARAM wParam, LPARAM lPar
...
+#undef CMDID_LAST
+
+    // Get 'Close' menu item string from system menu
+    WCHAR szClose[128];
+    HMENU hSysMenu = ::GetSystemMenu(m_hWnd, FALSE);
+    MENUITEMINFOW mii = { sizeof(mii), MIIM_TYPE };
+    mii.dwTypeData = szClose;
+    mii.cch = _countof(szClose);
+    if (!::GetMenuItemInfoW(hSysMenu, SC_CLOSE, FALSE, &mii))
+        StringCchCopyW(szClose, _countof(szClose), L"&Close\tAlt+F4");
```
HBelusca, line 3648: Not sure this is necessary (especially since you are forced to hardcode an english string there). Just do the close-item menu insertion + separator below only if that GetMenuItemInfoW call succeeded.

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -3551,6 +3554,156 @@ LRESULT CShellBrowser::RelayMsgToShellView(UINT uMsg, WPARAM wParam, LPARAM lPar
...
+    }
+
+    // Delete some menu items
+    DeleteUnwantedVerbItems(pContextMenu, hMenu, CMDID_FIRST, CMDID_LAST);
+    DeleteDuplicateSeparators(hMenu);
+    C_ASSERT(!(CMDID_FIRST <= SC_CLOSE && SC_CLOSE <= CMDID_LAST));
+#undef CMDID_FIRST
+#undef CMDID_LAST
+
+    // Get 'Close' menu item string from system menu
```
HBelusca, line 3641: General question: Isn't it possible instead to call GetSystemMenu first, and then remove everything from it except for the "close" item, and then append the (edited) contents of the context menu as done above? (perhaps this would mean, calling pContextMenu->QueryContextMenu() on that system menu handle? :thinking: )
If not, why?

HBelusca (changes requested):

dll/win32/browseui/shellbrowser.cpp
HBelusca: By the way, why not moving those action things *after* destroying the menu (i.e. after that DestroyMenu call below) ?

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -3551,6 +3554,156 @@ LRESULT CShellBrowser::RelayMsgToShellView(UINT uMsg, WPARAM wParam, LPARAM lPar
...
+    {
+        CMINVOKECOMMANDINFO info =
+        {
+            sizeof(info), CMIC_MASK_FLAG_NO_UI | SEE_MASK_ASYNCOK, m_hWnd, MAKEINTRESOURCEA(id)
+        };
+        pContextMenu->InvokeCommand(&info);
+    }
+
+    // Clean up
+    ::PostMessageW(m_hWnd, WM_NULL, 0, 0);
```
HBelusca, line 3686: Isn't that WM_NULL trick only necessary when displaying context menus for notification icons?

HBelusca:
Re-analysing on Windows 2003, the "Close" item that shows in this specific right-click "system context menu" is *NOT* the "Close" item that comes from the usual system menu, but the item that is in the File menu of the explorer window:

Left: the "File" menu ; Center: the context menu when right-clicking on the icon ; Right: the standard system menu when either left-clicking on the icon, or, right-clicking elsewhere on the window title bar.
image

dll/win32/browseui/shellbrowser.cpp
katahiromz: Renamed.

dll/win32/browseui/shellbrowser.cpp
katahiromz: Fixed.

dll/win32/browseui/shellbrowser.cpp
katahiromz: OK.

dll/win32/browseui/shellbrowser.cpp
katahiromz: The original system menu is still used for when R-Clicking non-icon space of title bar.

dll/win32/browseui/shellbrowser.cpp
katahiromz: Deleted.

katahiromz:
"Postponed" due to the background menu issue.
Please reopen if the issue got fixed.

PR #5061, opened 2023-02-11
[NTUSER] Optimize Window Snap Disabling

JoachimHenze:
## Purpose

I experimented with porting back the code-changes of (#5014) locally and while doing so,
I saw that it can be optimized a lot for the case that Window Snapping is turned off:

In defwnd.c we can save a call to UserGetForegroundWindow(), a call to UserGetWindowObject(hwndTop), assignments, ifs, traces.
In nonclient.c we can save a call to UserSystemParametersInfo(), assignments, ifs, traces.
In nonclient.c we are within a tight for(;;) even!

And I also think the code is much more readable like that,
as it separates the Window-Snapped-paths much leaner from the non-Window-Snapped-paths.

That gets very obvious when having a look again at the initial commit that introduced Window-Snapping:
https://github.com/reactos/reactos/commit/7e396787ed00a3939031447725f79697a41878e1
We can see that its code begins much further upwards than where Katayama decided to check his new flag.

**All relevant code changes are in the first commit, the 2nd commit is only non-functional formatting improvements**

### DIFF
--- a/win32ss/user/ntuser/defwnd.c
+++ b/win32ss/user/ntuser/defwnd.c
@@ -792,25 +792,20 @@ IntDefWindowProc(
          if (g_bWindowSnapEnabled && (IS_KEY_DOWN(gafAsyncKeyState, VK_LWIN) || IS_KEY_DOWN(gafAsyncKeyState, VK_RWIN)))
          {
             BOOL IsTaskBar;
-            DWORD Style;
-            DWORD ExStyle;
+            DWORD StyleTB;
+            DWORD ExStyleTB;
             HWND hwndTop = UserGetForegroundWindow();
             PWND topWnd = UserGetWindowObject(hwndTop);
 
-            /* Test for typical TaskBar ExStyle Values */
-            ExStyle = (topWnd->ExStyle & WS_EX_TOOLWINDOW);
-            TRACE("ExStyle=%x\n", ExStyle);
+            // We want to forbid snapping operations on the TaskBar
+            // We use a heuristic for detecting the TaskBar Wnd by its typical Style & ExStyle Values
+            ExStyleTB = (topWnd->ExStyle & WS_EX_TOOLWINDOW);
+            StyleTB = (topWnd->style & (WS_POPUP | WS_VISIBLE | WS_CLIPSIBLINGS | WS_CLIPCHILDREN));
+            IsTaskBar = (StyleTB == (WS_POPUP | WS_VISIBLE | WS_CLIPSIBLINGS | WS_CLIPCHILDREN))
+                        && (ExStyleTB == WS_EX_TOOLWINDOW);
+            TRACE("ExStyle=%x Style=%x IsTaskBar=%d\n", StyleTB, ExStyleTB, IsTaskBar);
 
-            /* Test for typical TaskBar Style Values */
-            Style = (topWnd->style & (WS_POPUP | WS_VISIBLE | WS_CLIPSIBLINGS | WS_CLIPCHILDREN));
-            TRACE("Style=%x\n", Style);
-
-            /* Test for masked typical TaskBar Style and ExStyles to detect TaskBar */
-            IsTaskBar = (Style == (WS_POPUP | WS_VISIBLE | WS_CLIPSIBLINGS | WS_CLIPCHILDREN))
-                        && (ExStyle == WS_EX_TOOLWINDOW);
-            TRACE("IsTaskBar=%d\n", IsTaskBar);
-
-            if (topWnd && !IsTaskBar)  /* Second test is so we are not touching the Taskbar */
+            if (topWnd && !IsTaskBar)
             {
                if ((topWnd->style & WS_THICKFRAME) == 0)
                   return 0;
--- a/win32ss/user/ntuser/defwnd.c
+++ b/win32ss/user/ntuser/defwnd.c
@@ -408,7 +408,7 @@ UserPaintCaption(PWND pWnd, INT Flags)
      * When themes are not enabled we can go on and paint the non client area.
      * However if we do that with themes enabled we will draw a classic frame.
      * This is solved by sending a themes specific message to notify the themes
-     * engine that the caption needs to be redrawn
+     * engine that the caption needs to be redrawn.
      */
       if (gpsi->dwSRVIFlags & SRVINFO_APIHOOK)
       {
--- a/win32ss/user/ntuser/defwnd.c
+++ b/win32ss/user/ntuser/defwnd.c
@@ -792,7 +792,7 @@ IntDefWindowProc(
                co_IntSendMessage(UserHMGetHandle(Wnd), WM_CONTEXTMENU, (WPARAM)UserHMGetHandle(Wnd), MAKELPARAM(-1, -1));
             }
          }
-         if (IS_KEY_DOWN(gafAsyncKeyState, VK_LWIN) || IS_KEY_DOWN(gafAsyncKeyState, VK_RWIN))
+         if (g_bWindowSnapEnabled && (IS_KEY_DOWN(gafAsyncKeyState, VK_LWIN) || IS_KEY_DOWN(gafAsyncKeyState, VK_RWIN)))
          {
             HWND hwndTop = UserGetForegroundWindow();
             PWND topWnd = UserGetWindowObject(hwndTop);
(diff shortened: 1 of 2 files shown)


JoachimHenze:
cc @Getequ

win32ss/user/ntuser/defwnd.c
```diff
@@ -792,7 +792,7 @@ IntDefWindowProc(
                co_IntSendMessage(UserHMGetHandle(Wnd), WM_CONTEXTMENU, (WPARAM)UserHMGetHandle(Wnd), MAKELPARAM(-1, -1));
             }
          }
-         if (IS_KEY_DOWN(gafAsyncKeyState, VK_LWIN) || IS_KEY_DOWN(gafAsyncKeyState, VK_RWIN))
+         if (g_bWindowSnapEnabled && (IS_KEY_DOWN(gafAsyncKeyState, VK_LWIN) || IS_KEY_DOWN(gafAsyncKeyState, VK_RWIN)))
          {
             HWND hwndTop = UserGetForegroundWindow();
```
katahiromz, line 797: I can confirm code equivalency.
You can make below code a newly-defined function.

win32ss/user/ntuser/nonclient.c
```diff
@@ -393,8 +393,13 @@ DefWndDoSizeMove(PWND pwnd, WORD wParam)
       if (!co_IntGetPeekMessage(&msg, 0, 0, 0, PM_REMOVE, TRUE)) break;
       if (IntCallMsgFilter( &msg, MSGF_SIZE )) continue;
 
-      /* Exit on button-up */
-      if (msg.message == WM_LBUTTONUP)
+      if (!g_bWindowSnapEnabled) /* If no WindowSnapEnabled: Exit on button-up, Return, or Esc */
+      {
+         if((msg.message == WM_LBUTTONUP) ||
```
katahiromz, line 398: You can move the following code above here:
```c
      /* Exit on Return or Esc */
      if (msg.message == WM_KEYDOWN &&
          (msg.wParam == VK_RETURN || msg.wParam == VK_ESCAPE))
      {
         break;
      }
```

win32ss/user/ntuser/defwnd.c
JoachimHenze: Yes, I could do that and I just experimented with it. But I do like the current version more, especially because the func is only and will only ever be called once, and the current version makes the differences in the key-handling more visible.
I do think just using a **block** in the current function is superior and allows more flexibility at a more flat callstack.

But an aspect of your suggestion is very good, and I do think that part is what motivated your suggestion:
I can move the declaration of the 3 Window-SnapOnly-variables "IstTaskBar, Style, Exstyle" into the Window-Snap-block.
It limits their scope and it keeps all parts of the WindowSnapping in this func within the Window-Snap-block!
It builds also with VC2010, I double-checked that. **Thanks for the input!** I will push that with commit3.

JoachimHenze: Suggestion Resolved?

katahiromz: Ok.

win32ss/user/ntuser/nonclient.c
JoachimHenze: Possible that I am just not clever enough to understand the code entirely, but are you really 100% sure that this would be safe? I mean we wouldn't call then any longer code like
```
pwnd->InternalPos.NormalRect = origRect;
co_IntSendMessage(UserHMGetHandle(pwnd), WM_SYSCOMMAND, SC_MAXIMIZE, 0);
```
and so on...
pwnd is a pointer from outside the func. So we may end up with a different state of that Window when exiting (which would most likely be the state of the previous iteration).
Are you 100% sure that this will not change the behavior?

Is it still guaranteed for the Window to properly get back to its original position?
Sorry for the question, but I really don't want to break inside here. :-)

katahiromz: I can say that it depends `UserIsEnteredExclusive()` or not. If exclusive mode, the execution keeps the `PWND` data.

katahiromz: It'll be better if we wrote `NT_ASSERT(UserIsEnteredExclusive());` in some places.

katahiromz: You can increment the reference counter of `WND` by calling `UserRefObjectCo` function.

win32ss/user/ntuser/defwnd.c
```diff
@@ -529,6 +529,122 @@ DefWndScreenshot(PWND pWnd)
...
+    ExStyle = (topWnd->ExStyle & WS_EX_TOOLWINDOW);
+    /* Test for typical TaskBar Style Values */
+    Style = (topWnd->style & (WS_POPUP | WS_VISIBLE | WS_CLIPSIBLINGS | WS_CLIPCHILDREN));
+    /* Test for masked typical TaskBar Style and ExStyles to detect TaskBar */
+    IsTaskBar = (Style == (WS_POPUP | WS_VISIBLE | WS_CLIPSIBLINGS | WS_CLIPCHILDREN)) &&
+                (ExStyle == WS_EX_TOOLWINDOW);
+
+    TRACE("ExStyle=%x, Style=%x, IsTaskBar=%d\n", ExStyle, Style, IsTaskBar);
+
+    /* Second test is so we are not touching the Taskbar */
```
JoachimHenze, line 556: The new comment is misleading. You mean the **first** test, not the second test. I do recommend to strip the comment entirely, because it is of Captain Obvious style.

win32ss/user/ntuser/nonclient.c
JoachimHenze: Okay, we tested and worked in parallel on moving up the checks. And both came to the conclusion that it works.
Thank you for your input. And your idea for fusing the TRACEs.
Suggestion resolved? (see latest commits)

JoachimHenze:
Finished now. Ready for review!

win32ss/user/ntuser/defwnd.c
```diff
@@ -792,32 +789,27 @@ IntDefWindowProc(
...
-            /* Test for typical TaskBar Style Values */
-            Style = (topWnd->style & (WS_POPUP | WS_VISIBLE |
-                        WS_CLIPSIBLINGS | WS_CLIPCHILDREN));
-            TRACE("Style is '%x'.\n", Style);
-
-            /* Test for masked typical TaskBar Style and ExStyles to detect TaskBar */
-            IsTaskBar = (Style == (WS_POPUP | WS_VISIBLE | WS_CLIPSIBLINGS | WS_CLIPCHILDREN))
-                        && (ExStyle == WS_EX_TOOLWINDOW);
-            TRACE("This %s the TaskBar.\n", IsTaskBar ? "is" : "is not");
+            // We want to forbid snapping operations on the TaskBar
```
HBelusca, line 800: Question, for here and the other file:
Can you explain what this "forbidding snapping operations on the taskbar" is all about?
When you test on Windows, by moving the taskbar to e.g. the top, the left or the right of the screen (instead of the standard bottom position), you will notice that the snapping action is forbidden only at the bottom, independently of where the taskbar is.

image

Dragging any window to the bottom (of the "workarea"/screen) won't make the snapping work.

HBelusca (changes requested):
Let's fix the damn thing since we're at it ^^

win32ss/user/ntuser/defwnd.c
```diff
@@ -404,15 +404,15 @@ UserPaintCaption(PWND pWnd, INT Flags)
   {
       if (pWnd->state & WNDS_HASCAPTION && pWnd->head.pti->MessageQueue == gpqForeground)
          Flags |= DC_ACTIVE;
-    /* 
+    /*
      * When themes are not enabled we can go on and paint the non client area.
      * However if we do that with themes enabled we will draw a classic frame.
      * This is solved by sending a themes specific message to notify the themes
-     * engine that the caption needs to be redrawn 
+     * engine that the caption needs to be redrawn
```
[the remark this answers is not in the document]
JoachimHenze: Oof, you are a punctuation freak. Accepted.

win32ss/user/ntuser/defwnd.c
JoachimHenze: Although your question is totally out of the scope of this PR, I still want to answer your question: see the resolved https://jira.reactos.org/browse/CORE-16289 to understand what this code/fix of Doug was about.

learn-more (approved):

HBelusca (approved):

katahiromz (approved):

PR #5063, opened 2023-02-11
[MCIAVI32] Media Player: Don't open empty files

atharva1910:
Opening Empty files in Mediaplayer leads to a crash
JIRA issue: [CORE-18669](https://jira.reactos.org/browse/CORE-18669)

### DIFF
--- a/dll/win32/mciavi32/mmoutput.c
+++ b/dll/win32/mciavi32/mmoutput.c
@@ -401,6 +401,7 @@ BOOL MCIAVI_GetInfo(WINE_MCIAVI* wma)
 
     /* Empty file */
     if (alb.numVideoFrames == 0) {
+        WARN("NumVideoFrames: %u, Empty or possibly corrupt video file");
         return FALSE;
     }
 
@@ -619,8 +620,10 @@ double MCIAVI_PaintFrame(WINE_MCIAVI* wma, HDC hDC)
 
     if (wma->dwCurrVideoFrame != wma->dwCachedFrame)
     {
-        if (wma->dwCurrVideoFrame >= wma->dwPlayableVideoFrames)
+        if (wma->dwCurrVideoFrame >= wma->dwPlayableVideoFrames) {
+            ERR("Invalid frame requested. Current : %u Total Playable %u\n", wma->dwCurrVideoFrame, wma->dwPlayableVideoFrames);
             return 0;
+        }
 
         if (!wma->lpVideoIndex[wma->dwCurrVideoFrame].dwOffset)
 	    return 0;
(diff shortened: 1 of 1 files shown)


KRosUser2 (approved):

KRosUser2:
Lgtm. Please check if proposed patch is not needed anyway for the sake of robustness against Exception

katahiromz:
It don't open zero-byte-sized files.
Then, how is an one-byte-sized file?

atharva1910:
> It don't open zero-byte-sized files. Then, how is an one-byte-sized file?

I'm not sure I understand the question. 
The file is basically just an AVI header and no payload. So it contains no playable data

dll/win32/mciavi32/mmoutput.c
```diff
@@ -398,6 +398,12 @@ BOOL MCIAVI_GetInfo(WINE_MCIAVI* wma)
 
 	mmioAscend(wma->hFile, &mmckInfo, 0);
     }
+
+    /* Empty file*/
```
katahiromz, line 402:
```suggestion
    /* Empty file */
```

atharva1910: Done 👍

KRosUser2:
> It don't open zero-byte-sized files. Then, how is an one-byte-sized file?

@katahiromz : I don't get your point. We don't talk about 0 byte file here but 0 FRAME file.

dll/win32/mciavi32/mmoutput.c
```diff
@@ -398,6 +398,12 @@ BOOL MCIAVI_GetInfo(WINE_MCIAVI* wma)
 
 	mmioAscend(wma->hFile, &mmckInfo, 0);
     }
+
+    /* Empty file */
+    if (alb.numVideoFrames == 0) {
+        return FALSE;
```
HBelusca, line 404: Maybe add a WARN trace?

HBelusca:
Shouldn't this fix be sent upstream to Wine?

dll/win32/mciavi32/mmoutput.c
KRosUser2: Good idea, same for the other place too (was the idea of the initial patch btw)

atharva1910: Done :+1:

atharva1910:
> Shouldn't this fix be sent upstream to Wine?

As I understand, this is not part of the Wine Win32 DLLs but rather the MediaPlayer which is a part of ReactOS

learn-more (approved):

HBelusca:
> > Shouldn't this fix be sent upstream to Wine?
> 
> As I understand, this is not part of the Wine Win32 DLLs but rather the MediaPlayer which is a part of ReactOS

I don't know; I see Wine authors listed at the beginning of the file, as well as its presence there:
https://github.com/wine-mirror/wine/tree/master/dlls/mciavi32
as well as the fact it's mentioned it's synced with Wine there (though a bit outdated): https://github.com/reactos/reactos/blob/master/media/doc/WINESYNC.txt#L95

HBelusca (approved):

PR #5065, opened 2023-02-11
[ATL] Remove some code that doesn't exist in the native ATL

learn-more:
## Purpose

This retry does not happen in the native ATL, so we should not do it either.

### DIFF
--- a/sdk/lib/atl/atlbase.h
+++ b/sdk/lib/atl/atlbase.h
@@ -1077,11 +1077,6 @@ public:
 
         HKEY hKey = NULL;
         LONG lRes = ::RegOpenKeyEx(hKeyParent, lpszKeyName, 0, samDesired, &hKey);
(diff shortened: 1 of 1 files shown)


JoachimHenze:
Sure, why not. But don't forget that you once committed that yourself with the aim to fix https://jira.reactos.org/browse/CORE-11746
Do you want to reopen that ticket then?

learn-more:
> Sure, why not. But don't forget that you once committed that yourself with the aim to fix https://jira.reactos.org/browse/CORE-11746 Do you want to reopen that ticket then?

Only if I delete CRegKey

JoachimHenze (approved):
Anyway. I just wanted to make you aware of the historic relation.
I guess you know what you are doing, and I am not aware of any real world case relying on that retry.
If code can be deleted, and tests prove that retry to be wrong, I am all up for it!

HBelusca (approved):

PR #2600, opened 2020-04-19
[AUDIOSRV] logmsg(): Skip it on non-debug builds. CORE-16912

SergeGautherie:
Near mininal workaround for
JIRA issue: [CORE-16912](https://jira.reactos.org/browse/CORE-16912)

Co-authored-by: Hermès BÉLUSCA - MAÏTO <hermes.belusca-maito@reactos.org>

PR #5067, opened 2023-02-12
[AUDIOSRV] Overhaul logging interfaces CORE-16912

JoachimHenze:
fixes [CORE-16912](https://jira.reactos.org/browse/CORE-16912) 'Starting "Audio Service" logs debug-level data in file in nominal execution'

Since 0.4.15-dev-445-g 353b5440474f3bf2c9dd16eee11bd42f11ea5d6d our audio service does not longer log into a file. But since then it doesn't log at all anymore.

We decided back then to switch from the specific apps logmsg() to our regular DPRINT logging ultimately. This is what this PR does provide.

I could successfully compile the PR with both: NDEBUG defined and NDEBUG undefined for the following toolchains: RosBEWin2.1.6 GCC4.7.2 dbg-config (by copying my result into releases/0.4.14 for a test)
RosBEWin2.1.6 GCC4.7.2 rls-config (by copying my result into releases/0.4.14 for a test)
RosBEWin2.1.6 MSVC2010 dbg-config (by copying my result into releases/0.4.14 for a test)
RosBEWin2.1.6 MSVC2010 rls-config (by copying my result into releases/0.4.14 for a test)
RosBEWin2.2.2 GCC8.4.0 dbg-config
RosBEWin2.2.2 GCC8.4.0 rls-config

I decided to keep NDEBUG defined for now, so we will see no change to the current behavior to master head.

But to prove it to be working also when NDEBUG is not defined, I attached also a log of bootup of master head.

Since I touched the files anyway, I tweaked the headers to new-style also.


JoachimHenze:
[z_NDEBUGuncommented_would_log_like_that.log](https://github.com/reactos/reactos/files/10715044/z_NDEBUGuncommented_would_log_like_that.log)
e.g.
```
(base/services/audiosrv/main.c:168) Audio Service main()
(base/services/audiosrv/main.c:83) * Service starting
(base/services/audiosrv/main.c:84) Registering service control handler
(base/services/audiosrv/main.c:89) Service status handle 2316824
(base/services/audiosrv/main.c:108) Creating audio device list
(base/services/audiosrv/pnp_list_manager.c:107) Creating file mapping
(base/services/audiosrv/pnp_list_manager.c:125) Mapping view of file
(base/services/audiosrv/pnp_list_manager.c:155) Device list created
(base/services/audiosrv/main.c:119) Registering for device notifications

```

base/services/audiosrv/services.c
```diff
@@ -1,13 +1,15 @@
 /*
- * PROJECT:          ReactOS
- * LICENSE:          GPL - See COPYING in the top level directory
- * FILE:             base/services/audiosrv/services.c
- * PURPOSE:          Audio Service Plug and Play
- * COPYRIGHT:        Copyright 2009 Johannes Anderwald
+ * PROJECT:     ReactOS
+ * LICENSE:     GPL - See COPYING in the top level directory
```
katahiromz, line 3:
```suggestion
 * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
```

katahiromz (commented):
LGTM.

base/services/audiosrv/services.c
JoachimHenze: I will have to do it in all files. Therefore I will not accept the suggestion via webinterface, but I add it myself.

JoachimHenze: But is "GPL - See COPYING in the top level directory" really GPL2+? I will add @ColinFinck for certainty here.

base/services/audiosrv/pnp_list_manager.c
```diff
@@ -90,19 +85,17 @@ AppendAudioDeviceToList(PnP_AudioDevice* device)
...
 
 BOOL
 CreateAudioDeviceList(DWORD max_size)
 {
-/*    printf("Initializing memory device list lock\n");*/
-
     if (!InitializeAudioDeviceListLock())
     {
-        /*printf("Failed!\n");*/
+        /*DPRINT("Failed\n");*/
```
SergeGautherie, line 98:
```suggestion
        /*DPRINT("InitializeAudioDeviceListLock() failed\n");*/
```

JoachimHenze: not needed. it is the only failed call in here. It is even commented out. and when commented in it will contain the line number even. I just kept that dead code to highlight that this is a failure path and no good-path.

base/services/audiosrv/services.c
learn-more: That file has an exception at the footer of it, so moving it to play GPL will make the code less usable.

KRosUser2 (approved):
LGTM, better than original https://github.com/reactos/reactos/pull/2600

learn-more (approved):

base/services/audiosrv/services.c
JoachimHenze: @learn-more , you wanted to help resolving this discussion. Unfortunately I do not understand entirely yet in which direction you want to push. You want me to **not** change it, right?

learn-more: I am not a lawyer, but I am pretty sure the last paragraph in that file is not plain gpl, so you probably want something slightly less evil like the LGPL.

It might be a good idea to see with which other occurrences of this were replaced.

JoachimHenze: That is still not the clear answer I hoped for. I will wait for @ColinFinck s reply before doing any action on this.
If the doubt would remain after Colins reply, then my personal preference would be to simply not change it.

ColinFinck: @JoachimHenze The primary problem here is that "GPL - See COPYING in the top level directory" doesn't equal "GPL-2.0-**or-later**".
The license text is just about GPLv2 and several developers have expressed their objections against GPLv3 in the past, explicitly making their code GPL-2.0-only. Unless you ask the developers of this source, you may only change this line to "GPL-2.0-only".

@learn-more You are right that our "COPYING" file in the top-level directory indeed has additional lines compared to a regular GPLv2. I have laid out in https://www.youtube.com/watch?v=OTVVz9VLisU why I think that these lines are superfluous _for ReactOS_ and don't grant any additional rights over the regular GPLv2. I even went as far as saying that the effects of LGPL and GPL are identical _when implementing a Windows-compatible component for ReactOS_.

That said, I'm not a lawyer. You can again only be sure about the intended license when asking the developers of this particular source. Chances are high that they didn't even know about the additional exception in our "COPYING" file when writing this code. They may also agree to relicense it under LGPL or even more permissive if you ask them.

Note that despite my claim that the effects of LGPL and GPL are identical _for us_, you can't just take a GPL-licensed source and slap an LGPL license on it. The source may later be used in a different context where the effects of LGPL and GPL are different.
You could however always restrict the license even more, so replacing "GPL - See COPYING in the top level directory" by "GPL-2.0-only" is definitely possible (effectively removing the additional exceptions from our "COPYING" file).


**TL;DR:** If in doubt, leave this unchanged. If you really need an SPDX line, choose "GPL-2.0-only". If you need more than that, ask the original developers.

HBelusca (approved):

base/services/audiosrv/services.c
JoachimHenze: Ok, I will leave it like it is then.

PR #5069, opened 2023-02-12
[NTUSER] IntWinListOwnedPopups: Exclude default IME window from the list

TAN-Gaming:
## Purpose
- Fix `ASSERT(pWnd->spwndOwner == Window)` in `IntDestroyOwnedWindows`.
- Also remove window owner check hack in `IntShowOwnedPopups`.

Based on my observation, the owner of the default IME window can change and it depends on the window that is being focused.

JIRA issue: [CORE-18659](https://jira.reactos.org/browse/CORE-18659) [CORE-18768](https://jira.reactos.org/browse/CORE-18768)

### DIFF
--- a/win32ss/user/ntuser/window.c
+++ b/win32ss/user/ntuser/window.c
@@ -314,6 +314,15 @@ IntWinListChildren(PWND Window)
     return List;
 }
 
+static BOOL
+IntWndIsDefaultIme(_In_ PWND Window)
+{
+    PTHREADINFO pti = Window->head.pti;
+
+    return (IS_IMM_MODE() && !(pti->TIF_flags & TIF_INCLEANUP) &&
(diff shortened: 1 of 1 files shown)


learn-more:
I am not sure this change is correct.
Why is this window listed as owner when it should not be?

Note that `IntDestroyOwnedWindows` retrieves a list of windows...

learn-more (approved):
These changes make sense.

@katahiromz can you please take a look since this touches IMM code?

katahiromz (approved):
`IntWinListOwnedPopups` is also used in `IntShowOwnedPopups`.
The default IME window is normally hidden, so we don't have to show the default IME window in `IntShowOwnedPopups`.

HBelusca (approved):

PR #5071, opened 2023-02-12
[BROWSEUI] Relevance column shows file names

atharva1910:
[CORE-18630](https://jira.reactos.org/browse/CORE-18630)

Make Relevance column empty and not show file names


KRosUser2:
Why all these commits ??

atharva1910:
Created a cleaner PR : https://github.com/reactos/reactos/pull/5071

PR #5072, opened 2023-02-12
[BROWSEUI] Don't show file names in Relevance column

atharva1910:
[CORE-18630](https://jira.reactos.org/browse/CORE-18630)

Make Relevance column empty and not show file names


dll/win32/browseui/shellfind/CFindFolder.cpp
```diff
@@ -671,11 +671,17 @@ STDMETHODIMP CFindFolder::GetDetailsOf(PCUITEMID_CHILD pidl, UINT iColumn, SHELL
...
         return SHSetStrRet(&pDetails->str, _ILGetPath(pidl));
     }
 
+    if (iColumn == IDS_COL_RELEVANCE_COLNUM)
+    {
+        /* TODO: Fill once the relevance is calculated */
+        return SHSetStrRet(&pDetails->str, "");
+    }
+
     return GetDisplayNameOf(pidl, SHGDN_NORMAL, &pDetails->str);
```
HBelusca, line 685: Shouldn't this even be some kind of `if (iColumn == the_name_column_index) { return GetDisplayNameOf(...); }` , in order to ensure always that the file name will only be shown in the "Name" column only?

dll/win32/browseui/resource.h
```diff
@@ -161,22 +161,25 @@
...
+#define IDS_COL_NAME             8976
+#define IDS_COL_LOCATION         8977
+#define IDS_COL_RELEVANCE        8989
+#define IDS_SEARCH_FILES_FOUND   9232
+#define IDS_SEARCH_FOLDER        9234
+#define IDS_SEARCH_RESULTS       30520
+#define IDS_SEARCH_OPEN_FOLDER   40960
+#define IDS_COL_NAME_COLNUM      0
+#define IDS_COL_LOCATION_COLNUM  1
+#define IDS_COL_RELEVANCE_COLNUM 2
```
JoachimHenze, line 182: Why does the github web review show so much diff here, while in fact only 3 lines have been added at the end? (as far as I can see at least)

KRosUser2 (approved):
Lgtm

dll/win32/browseui/shellfind/CFindFolder.cpp
```diff
@@ -671,11 +671,17 @@ STDMETHODIMP CFindFolder::GetDetailsOf(PCUITEMID_CHILD pidl, UINT iColumn, SHELL
...
 
-    if (iColumn == 1)
+    if (iColumn == IDS_COL_LOCATION_COLNUM)
     {
         return SHSetStrRet(&pDetails->str, _ILGetPath(pidl));
     }
 
+    if (iColumn == IDS_COL_RELEVANCE_COLNUM)
+    {
+        /* TODO: Fill once the relevance is calculated */
```
JoachimHenze, line 681: this is a cpp file. You should not hesitate the slightest bit to favor CPP // commenting style over C /* */

dll/win32/browseui/resource.h
JoachimHenze: I do see now. The padding has been changed without any need. Revert that! Rejected.

JoachimHenze (changes requested):
I do see now. The padding has been changed without any need. Revert that! Rejected.

dll/win32/browseui/resource.h
JoachimHenze: And COLNUM should be COLUMN by the way, or something shorter.

dll/win32/browseui/shellfind/CFindFolder.cpp
ThFabba: I see some other `GetDetailsOf` implementations return a failure status for unknown column ids. That seems appropriate, but the current implementation doesn't seem like it would really hurt, either. This is pre-existing behavior so I'd be okay with just leaving it as is for this PR.

dll/win32/browseui/resource.h
ThFabba: I don't agree with "just use shorter names". Use descriptive names please. If that means the longest name gets longer, so be it.
That said, I think it would be more descriptive not to start these constant names with IDS_ because they are not string resource indexes. And, I don't think they should be grouped with all the other ID* constants, since those are resource identifiers, and these new ones aren't. That solves the alignment issue.

Just COLNUM_NAME or COLUMN_ID_NAME seems fine to me.

atharva1910: It's the ID of the column eg (IDS_COL_NAME) appended with COLNUM to create a define which describes the column number for the Column. This is done so that there can be seen a strict dependency between the COLNUM and the column names.

I fail to see how padding is a cause for concern here. The new code is in alignment with the old code, the new definitions are in alignment with the old (as explained above). Even the padding is added to be in alignment with what the file looked like before.

dll/win32/browseui/shellfind/CFindFolder.cpp
atharva1910: Oh good catch, that will prevent name being added to any new columns which are introduced.

Although you are right, i would prefer to do it as a seperate commit since this PR targets a specific issue.

dll/win32/browseui/resource.h
ThFabba: I'd be okay with the current names, but they should be grouped separately from the resource identifiers and have a comment explaining why they're called what they are.

JoachimHenze: I do agree to Thomas suggestion to separate them from the rest. Good recommendation!
colNUM is still a really bad choice. Sooner or later somebody will be tempted to rename them to column (erroneously).
I do suggest something like
COLUMN_ID_LOCATION
COLUMN_ID_RELEVANCE
COLUMN_ID_NAME
and separation from the block above

dll/win32/browseui/shellfind/CFindFolder.cpp
atharva1910: Done

dll/win32/browseui/resource.h
atharva1910: Done

learn-more (approved):
lgtm.

If you don't mind, could you also update `CFindFolder::CompareIDs` with the new ID's?
Not a requirement for merging tho.

HBelusca (approved):

JoachimHenze (approved):

PR #5073, opened 2023-02-13
[SHELL32] Enable Network Connection Rename CORE-18750

zecarlos1957:
[SHELL32] Enable Network Connection Rename 

Add Network Connection pidl type 0x99
Value hard coded at netshell/enumlist.cpp in function ILCreateNetConnectionItem(...)

ms say:
ILIsEqual performs a binary comparison of the item data.

JIRA issue: [CORE-18750](https://jira.reactos.org/browse/CORE-18750)

### DIFF
--- a/dll/win32/shell32/debughlp.cpp
+++ b/dll/win32/shell32/debughlp.cpp
@@ -379,7 +379,6 @@ BOOL pcheck( LPCITEMIDLIST pidl )
                 case PT_YAGUID:
                 case PT_IESPECIAL2:
                 case PT_SHARE:
-                case 0x99:      /* Network Connection pidl type */
                     break;
                 default:
                     ERR("unknown IDLIST %p [%p] size=%u type=%x\n",
--- a/dll/win32/shell32/debughlp.cpp
+++ b/dll/win32/shell32/debughlp.cpp
@@ -379,6 +379,7 @@ BOOL pcheck( LPCITEMIDLIST pidl )
                 case PT_YAGUID:
                 case PT_IESPECIAL2:
                 case PT_SHARE:
+                case 0x99:      /* Network Connection pidl type */
                     break;
                 default:
                     ERR("unknown IDLIST %p [%p] size=%u type=%x\n",
--- a/dll/win32/shell32/wine/pidl.c
+++ b/dll/win32/shell32/wine/pidl.c
@@ -476,13 +476,51 @@ LPITEMIDLIST WINAPI ILGlobalClone(LPCITEMIDLIST pidl)
     return newpidl;
 }
 
-BOOL _ILCompareSimpleIDs(LPCITEMIDLIST pidl1, LPCITEMIDLIST pidl2)
+BOOL _ILHACKCompareSimpleIds(LPCITEMIDLIST pidltemp1, LPCITEMIDLIST pidltemp2)
 {
-    if (pidl1->mkid.cb != pidl2->mkid.cb)
+    LPPIDLDATA pdata1 = _ILGetDataPointer(pidltemp1);
+    LPPIDLDATA pdata2 = _ILGetDataPointer(pidltemp2);
+
+    IID *iid1 = _ILGetGUIDPointer(pidltemp1);
+    IID *iid2 = _ILGetGUIDPointer(pidltemp2);
+
+    FileStructW* pDataW1 = _ILGetFileStructW(pidltemp1);
+    FileStructW* pDataW2 = _ILGetFileStructW(pidltemp2);
+
+    if (_ILIsDesktop(pidltemp1) && _ILIsDesktop(pidltemp2))
+    {
+        return TRUE;
+    }
+    else if (_ILIsDesktop(pidltemp1) || _ILIsDesktop(pidltemp2))
     {
         return FALSE;
     }
-    else if (!RtlEqualMemory((BYTE*)&pidl1->mkid, (BYTE*)&pidl2->mkid, pidl1->mkid.cb))
+    else if (iid1 || iid2)
+    {
+        if (!iid1 || !iid2 || memcmp(iid1, iid2, sizeof(GUID)))
+            return FALSE;
+    }
+    else if (pDataW1 || pDataW2)
+    {
+        if (!pDataW1 || !pDataW2 || wcsicmp(pDataW1->wszName, pDataW2->wszName))
+            return FALSE;
+    }
+    else if (_ILIsFolder(pidltemp1) || _ILIsFolder(pidltemp2))
+    {
+        if (!_ILIsFolder(pidltemp1) || !_ILIsFolder(pidltemp2) || strcmp(pdata1->u.file.szNames, pdata2->u.file.szNames))
+            return FALSE;
+    }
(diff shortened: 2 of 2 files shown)


dll/win32/shell32/wine/pidl.c
```diff
@@ -527,6 +528,21 @@ BOOL _ILHACKCompareSimpleIds(LPCITEMIDLIST pidltemp1, LPCITEMIDLIST pidltemp2)
 
     return TRUE;
 }
+#else
+BOOL _ILCompareSimpleIDs(LPCITEMIDLIST pidl1, LPCITEMIDLIST pidl2)
```
SergeGautherie, line 532: This function can be rewritten to be a single return expression.

dll/win32/shell32/debughlp.cpp
```diff
@@ -379,6 +379,7 @@ BOOL pcheck( LPCITEMIDLIST pidl )
                 case PT_YAGUID:
                 case PT_IESPECIAL2:
                 case PT_SHARE:
+                case 0x99:      /* Network Connection pidl type */
```
SergeGautherie, line 382: Make it a `#define`?

dll/win32/shell32/wine/pidl.c
```diff
@@ -554,7 +570,11 @@ BOOL WINAPI ILIsEqual(LPCITEMIDLIST pidl1, LPCITEMIDLIST pidl2)
 
     while (pidltemp1->mkid.cb && pidltemp2->mkid.cb)
     {
+#ifdef __REACTOS__
+        if (!_ILCompareSimpleIDs(pidltemp1, pidltemp2))
```
SergeGautherie, line 574: `#define` an alias, instead.

dll/win32/shell32/wine/pidl.c
learn-more: It could, but that does not make it better or more readable.

SergeGautherie: > not [...] more readable

I have a different point of view:
```
if (false cond) return false
else if (false cond) return false
return true
```
looks quite bloated to me, even more with the additional braces.

dll/win32/shell32/debughlp.cpp
learn-more: Netshell itself does not even have a define for this, so this is following the existing style.

dll/win32/shell32/wine/pidl.c
learn-more: I don't like this suggestion.

learn-more (approved):
lgtm

dll/win32/shell32/wine/pidl.c
```diff
@@ -476,6 +476,7 @@ LPITEMIDLIST WINAPI ILGlobalClone(LPCITEMIDLIST pidl)
     return newpidl;
 }
 
+#ifndef __REACTOS__
 BOOL _ILHACKCompareSimpleIds(LPCITEMIDLIST pidltemp1, LPCITEMIDLIST pidltemp2)
```
HBelusca, line 480: This "HACKCompare..." function was introduced there: https://github.com/reactos/reactos/commit/2af89ef8f01423c913f47df6e321cc1cec07b3bc
by @yagoulas . So it doesn't need to be "ifndef __REACTOS__"'ed.

yagoulas:
I'm not really confident that we can remove all these special cases for the pidls that are internal to shell32. If I remember correctly there were applications that passed pidls for CFSFolder with forms from older versions of windows (like the ones written in a shortcut file). Some tests may do crazy things like that.

 However I'm totally for adding an RtlEqualMemory call in the last else clause that now returns just FALSE.

learn-more:
> I'm not really confident that we can remove all these special cases for the pidls that are internal to shell32. If I remember correctly there were applications that passed pidls for CFSFolder with forms from older versions of windows (like the ones written in a shortcut file). Some tests may do crazy things like that.
> 
> However I'm totally for adding an RtlEqualMemory call in the last else clause that now returns just FALSE.

Any chance you could dig up an issue that was fixed by those special cases?

KRosUser2:
@yagoulas @learn-more : Do you consider a change is finally stil needed here or is the PR "approved" ?

katahiromz:
Cc @reactos/lang-russian

tkreuzer:
Please provide testbot results.

Doug-Lyons (approved):
LGTM.

zecarlos1957:
If we don't interpret any part of the pidls and leave only the binary comparison, when we delete a file (only in delete) the file is only deleted when we close the browser window.
Deleting a file triggers the error (win32ss/user/ntuser/msgqueue.c:2398): MQ Cleanup Post Messages E134E4F8. Having as a side effect a failure in the logic of the _ILCompareIds function, which always returns FALSE.

If I remember correctly, if we keep the GUIID test beyond the binary comparison solves this problem, said error is not triggered.

Are there advantages in this perspective?

zecarlos1957:
If we dont interpret any part of the pidls and we let just the binary comparation, when we delete a file (only on delete) the file is delected just when, and if we close the browser window.
Deleting a file, triggers the Error (win32ss/user/ntuser/msgqueue.c:2398) err: MQ Cleanup Post Messages E134E4F8.
Having as a side effect failure in the logic of the _ILCompareIds function , which always returns FALSE

zecarlos1957:
If we do not interpret any part of the pidls and we let just the binary comparison, when we delete a file (only on delete) the file is deleted just when, and if we close the browser window.

zecarlos1957:
_ILCompareIDs definitely doesn't work without the special cases in place.
The present proposal really brings an improvement because it enables the renaming of the shell item "Network Connections" and simultaneously can delete a file without problems, which does not happen in the previous proposal.

learn-more (approved):

yagoulas (approved):

dll/win32/shell32/wine/pidl.c
```diff
@@ -522,7 +522,14 @@ BOOL _ILHACKCompareSimpleIds(LPCITEMIDLIST pidltemp1, LPCITEMIDLIST pidltemp2)
     }
     else
     {
-        return FALSE;
+        if (pidltemp1->mkid.cb != pidltemp2->mkid.cb)
+        {
+            return FALSE;
+        }
+        else if (!RtlEqualMemory((BYTE*)&pidltemp1->mkid, (BYTE*)&pidltemp2->mkid, pidltemp1->mkid.cb))
```
HBelusca, line 529:
```suggestion
        if ((pidltemp1->mkid.cb != pidltemp2->mkid.cb) ||
            !RtlEqualMemory((BYTE*)&pidltemp1->mkid, (BYTE*)&pidltemp2->mkid, pidltemp1->mkid.cb))
```

HBelusca:
@GeoB99 Why "no squash merge" ?

dll/win32/shell32/wine/pidl.c
SergeGautherie: Eventually improved/done.

dll/win32/shell32/wine/pidl.c
SergeGautherie: Renaming was eventually reverted.

PR #5074, opened 2023-02-13
[IP] Allow 0xFFFF as Valid Network Port Number

Doug-Lyons:
## Purpose

Do not declare 0XFFFF as invalid in ReactOS. Applications should be able to use 0XFFFF as it is valid. Thanks to @KRosUser for detailed analysis.

JIRA issue:
[CORE-18371](https://jira.reactos.org/browse/CORE-18371)
[CORE-18764](https://jira.reactos.org/browse/CORE-18764)

## Proposed changes

The proposed fix is similar to what is already implemented in ReactOS for existing TCPAlloc failure in [FileOpenAddress](https://git.reactos.org/?p=reactos.git;a=blob;f=drivers/network/tcpip/tcpip/fileobjs.c#l475)

This PR replaces https://github.com/reactos/reactos/pull/4968 for which the creator has been inactive for over a month.

### DIFF
--- a/drivers/network/tcpip/ip/transport/tcp/accept.c
+++ b/drivers/network/tcpip/ip/transport/tcp/accept.c
@@ -77,10 +77,10 @@ NTSTATUS TCPListen(PCONNECTION_ENDPOINT Connection, UINT Backlog)
             if (NT_SUCCESS(Status))
             {
                 /* Allocate the port in the port bitmap */
-                Connection->AddressFile->Port = TCPAllocatePort(LocalAddress.Address[0].Address[0].sin_port);
-
-                /* This should never fail */
-                ASSERT(Connection->AddressFile->Port != 0xFFFF);
+                UINT AllocatedPort = TCPAllocatePort(LocalAddress.Address[0].Address[0].sin_port);
+                /* This should never fail unless all ports are in use */
+                ASSERT(AllocatedPort != (UINT)-1);
+                Connection->AddressFile->Port = AllocatedPort;
             }
         }
     }
--- a/drivers/network/tcpip/ip/transport/tcp/accept.c
+++ b/drivers/network/tcpip/ip/transport/tcp/accept.c
@@ -81,7 +81,7 @@ NTSTATUS TCPListen(PCONNECTION_ENDPOINT Connection, UINT Backlog)
                 /* This should never fail unless all ports are in use */
                 if (AllocatedPort == (UINT) -1)
                 {
-                    ERR("No more ports available.\n");
+                    DbgPrint("ERR: No more ports available.\n");
                     return STATUS_TOO_MANY_ADDRESSES;
                 }
                 Connection->AddressFile->Port = AllocatedPort;
--- a/drivers/network/tcpip/ip/transport/tcp/accept.c
+++ b/drivers/network/tcpip/ip/transport/tcp/accept.c
@@ -79,7 +79,11 @@ NTSTATUS TCPListen(PCONNECTION_ENDPOINT Connection, UINT Backlog)
                 /* Allocate the port in the port bitmap */
                 UINT AllocatedPort = TCPAllocatePort(LocalAddress.Address[0].Address[0].sin_port);
                 /* This should never fail unless all ports are in use */
-                ASSERT(AllocatedPort != (UINT)-1);
+                if (AllocatedPort == (UINT) -1)
+                {
+                    ERR("No more ports available.\n");
+                    return STATUS_TOO_MANY_ADDRESSES;
+                }
                 Connection->AddressFile->Port = AllocatedPort;
             }
         }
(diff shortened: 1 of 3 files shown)


drivers/network/tcpip/ip/transport/tcp/tcp.c
```diff
@@ -379,10 +381,10 @@ NTSTATUS TCPConnect
         }
 
         /* Allocate the port in the port bitmap */
-        Connection->AddressFile->Port = TCPAllocatePort(LocalAddress.Address[0].Address[0].sin_port);
-
+        AllocatedPort = TCPAllocatePort(LocalAddress.Address[0].Address[0].sin_port);
         /* This should never fail */
```
KRosUser2, line 385: /* This should never fail unless all ports are in use */

KRosUser2 (approved):
This code is good. Tested for 34 days with constant SMTP+POP3+HTTP+FTP without a single problem.

image

drivers/network/tcpip/ip/transport/tcp/tcp.c
GGLinnk:
```suggestion
        /* This should never fail unless all ports are in use */
```

GGLinnk (approved):

Doug-Lyons:
Testbot Results:
net.patch JID65050 on top of 0.4.15-dev-5680-g5ad5737

VBox: https://reactos.org/testman/compare.php?ids=85980,86009 LGTM
KVM: https://reactos.org/testman/compare.php?ids=85981,86008 LGTM

drivers/network/tcpip/ip/transport/tcp/accept.c
```diff
@@ -77,10 +77,10 @@ NTSTATUS TCPListen(PCONNECTION_ENDPOINT Connection, UINT Backlog)
             if (NT_SUCCESS(Status))
             {
                 /* Allocate the port in the port bitmap */
-                Connection->AddressFile->Port = TCPAllocatePort(LocalAddress.Address[0].Address[0].sin_port);
-
-                /* This should never fail */
-                ASSERT(Connection->AddressFile->Port != 0xFFFF);
+                UINT AllocatedPort = TCPAllocatePort(LocalAddress.Address[0].Address[0].sin_port);
+                /* This should never fail unless all ports are in use */
+                ASSERT(AllocatedPort != (UINT)-1);
```
sdever, line 82: Is the assert needed here ? Maybe the function should only return a status code to usermode (exhausting all local ports can happen in expected system lifecycle, for instance if server is target of DDoS or runs poorly-coded usermode application)

drivers/network/tcpip/ip/transport/tcp/tcp.c
```diff
@@ -379,10 +381,10 @@ NTSTATUS TCPConnect
         }
 
         /* Allocate the port in the port bitmap */
-        Connection->AddressFile->Port = TCPAllocatePort(LocalAddress.Address[0].Address[0].sin_port);
-
-        /* This should never fail */
-        ASSERT(Connection->AddressFile->Port != 0xFFFF);
+        AllocatedPort = TCPAllocatePort(LocalAddress.Address[0].Address[0].sin_port);
+        /* This should never fail unless all ports are in use */
+        ASSERT(AllocatedPort != (UINT)-1);
```
sdever, line 386: Same comment as above, reporting correct NTSTATUS to caller instead of asserting seems more correct (exhausting all local ports is a scenario that can happen in a "normal" lifecycle)

sdever (changes requested):

drivers/network/tcpip/ip/transport/tcp/accept.c
KRosUser2: This was a request from @learn-more in the previous PR, but based on the assumption that "this should never fail" which was the code comment.
But you're right, the error case is correctly manage and this CAN happen under very loaded use cases (or poorly written networking app) so the best option is NOT to ASSERT but to fail gracefully (and i'd suggest to log an ERR since it should be very unlikely)

drivers/network/tcpip/ip/transport/tcp/tcp.c
KRosUser2: This was a request from @learn-more in the previous PR, but based on the assumption that "this should never fail" which was the code comment.
But you're right, the error case is correctly manage and this CAN happen under very loaded use cases (or poorly written networking app) so the best option is NOT to ASSERT but to fail gracefully (and i'd suggest to log an ERR since it should be very unlikely)

drivers/network/tcpip/ip/transport/tcp/accept.c
Doug-Lyons: Latest commit attempts to address this issue. Please review and approve if appropriate.

drivers/network/tcpip/ip/transport/tcp/tcp.c
Doug-Lyons: Latest commit attempts to address this issue. Please review and approve if appropriate.

learn-more (approved):

sdever (approved):

tkreuzer (approved):
lgtm

Doug-Lyons:
Thanks for everyone's help in getting this merged.

PR #5078, opened 2023-02-17
[NOTEPAD] Avoid middle status of settings

katahiromz:
## Purpose
If there is the target registry key and the settings are incomplete, then the settings would become the middle status. This PR will avoid this status.

## Proposed changes

- Avoid buffer overrun in `QueryString` helper function.
- Improve `NOTEPAD_LoadSettingsFromRegistry` function.

### DIFF
--- a/base/applications/notepad/settings.c
+++ b/base/applications/notepad/settings.c
@@ -140,7 +140,8 @@ void NOTEPAD_LoadSettingsFromRegistry(void)
     /* FIXME: Globals.fSaveWindowPositions = FALSE; */
     /* FIXME: Globals.fMLE_is_broken = FALSE; */
 
-    RegOpenKey(HKEY_CURRENT_USER, s_szRegistryKey, &hKey);
+    if (RegOpenKey(HKEY_CURRENT_USER, s_szRegistryKey, &hKey) != ERROR_SUCCESS)
+        hKey = NULL;
 
     QueryByte(hKey, _T("lfCharSet"), &Globals.lfFont.lfCharSet);
     QueryByte(hKey, _T("lfClipPrecision"), &Globals.lfFont.lfClipPrecision);
--- a/base/applications/notepad/settings.c
+++ b/base/applications/notepad/settings.c
@@ -101,12 +101,12 @@ static BOOL QueryBool(HKEY hKey, LPCTSTR pszValueName, BOOL *pbResult)
     return TRUE;
 }
 
-static BOOL QueryString(HKEY hKey, LPCTSTR pszValueName, LPTSTR pszResult, DWORD dwResultLen)
+static BOOL QueryString(HKEY hKey, LPCTSTR pszValueName, LPTSTR pszResult, DWORD dwResultLength)
 {
-    if (!QueryGeneric(hKey, pszValueName, REG_SZ, pszResult, dwResultLen * sizeof(TCHAR)))
+    if (!QueryGeneric(hKey, pszValueName, REG_SZ, pszResult, dwResultLength * sizeof(TCHAR)))
         return FALSE;
-    assert(dwResultLen > 0);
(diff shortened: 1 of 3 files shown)


base/applications/notepad/settings.c
```diff
@@ -100,9 +101,54 @@ static BOOL QueryBool(HKEY hKey, LPCTSTR pszValueName, BOOL *pbResult)
     return TRUE;
 }
 
-static BOOL QueryString(HKEY hKey, LPCTSTR pszValueName, LPTSTR pszResult, DWORD dwResultSize)
+static BOOL QueryString(HKEY hKey, LPCTSTR pszValueName, LPTSTR pszResult, DWORD cchResult)
```
JoachimHenze, line 104: Why changing/breaking the hungarian notation?

katahiromz: http://www.imm.dtu.dk/~alan/hungarian.html
`cch` is also in Hungarian notation.
Renamed `cchResult` as `dwResultLen`.

katahiromz: There is a confusion between string size and length. I want to use the word 'length' (or 'len') to specify the length of a string.

JoachimHenze: Using dwResultLength should be fine though, if you think it is more descriptive. I just find dw for DWORD much more common and logical here than cch

katahiromz: OK.

katahiromz: OK.

base/applications/notepad/settings.c
```diff
@@ -100,9 +101,54 @@ static BOOL QueryBool(HKEY hKey, LPCTSTR pszValueName, BOOL *pbResult)
     return TRUE;
 }
 
-static BOOL QueryString(HKEY hKey, LPCTSTR pszValueName, LPTSTR pszResult, DWORD dwResultSize)
+static BOOL QueryString(HKEY hKey, LPCTSTR pszValueName, LPTSTR pszResult, DWORD dwResultLen)
 {
-    return QueryGeneric(hKey, pszValueName, REG_SZ, pszResult, dwResultSize * sizeof(TCHAR));
+    if (!QueryGeneric(hKey, pszValueName, REG_SZ, pszResult, dwResultLen * sizeof(TCHAR)))
+        return FALSE;
+    assert(dwResultLen > 0);
```
JoachimHenze: Can we bailout here instead (and handle that at the callers sideif needed), instead of adding an assert? That would also avoid the need for the additional include.

katahiromz: Deleted `#include <assert.h>`.

katahiromz:
If there is the registry key and the settings are incomplete, then the settings will become the middle status. This PR will avoid this status.

base/applications/notepad/settings.c
learn-more: cch is an actual useful notation that perfectly communicates the actual intent, as opposed to dw, which does not add anything of value.

HBelusca:
> If there is the target registry key and the settings are incomplete, then the settings would become the middle status. This PR will avoid this status.

What do you mean by that?

katahiromz:
For example, when registry key `HKCU\Software\Microsoft\Notepad` existed but `"szHeader"` registry value didn't exist, then `Globals.szHeader` becomes middle status.

HBelusca:
OK what do you mean by "middle status"? A setting is either initialized or not initialized.

katahiromz:
> what do you mean by "middle status"?

At first, The variable `Globals` is initialized by zeros. `szHeader` is also zeros. And if `RegOpenKey` succeeded and there is no `szHeader` registry value, `szHeader` will be still zeros.

learn-more:
> OK what do you mean by "middle status"? A setting is either initialized or not initialized.

Half initialized, some settings are initialized and others are not.

PR #5082, opened 2023-02-17
[NOTEPAD] Improve DIALOG_GoTo

katahiromz:
## Purpose

Improve user interface.

## Proposed changes

- Don't use text directly in `DIALOG_GoTo` function. Rely the `EM_*` message handlers.
- Fix and improve `DIALOG_GOTO` resource dialog (that was wrong in some points).
- Add `STRING_LINE_NUMBER_OUT_OF_RANGE` resource string to display error message.

## TODO

- [x] Do tests.


base/applications/notepad/dialog.c
```diff
@@ -1117,82 +1117,79 @@ VOID DIALOG_Replace(VOID)
...
+        case WM_COMMAND:
         {
             if (LOWORD(wParam) == IDOK)
             {
-                hTextBox = GetDlgItem(hwndDialog, ID_LINENUMBER);
-                GetWindowText(hTextBox, szText, ARRAY_SIZE(szText));
-                EndDialog(hwndDialog, _ttoi(szText));
-                bResult = TRUE;
+                INT iLine = GetDlgItemInt(hwndDialog, ID_LINENUMBER, NULL, FALSE);
+                // TODO: Should we display error message?
```
HBelusca, line 1144: What error message? If we want to go further than the end of file?
That would be the only reason for the "total number of lines" (cLines) to be passed via the goto_data structure.

katahiromz: No error message is better?

HBelusca: Personally an error would be better. Windows' notepad has one too: "The line number is larger than the total number of lines." (something like that; it looks as follows in french, I don't have an english version to see the actual error message)
image

katahiromz: Fixed.

learn-more: image

katahiromz: The line number zero is out of range.

learn-more: Zero and a too big number display the same text on the notepad on Win10.

base/applications/notepad/dialog.c
```diff
@@ -1117,82 +1117,90 @@ VOID DIALOG_Replace(VOID)
...
+
+        case WM_COMMAND:
         {
             if (LOWORD(wParam) == IDOK)
             {
-                hTextBox = GetDlgItem(hwndDialog, ID_LINENUMBER);
-                GetWindowText(hTextBox, szText, ARRAY_SIZE(szText));
-                EndDialog(hwndDialog, _ttoi(szText));
-                bResult = TRUE;
+                INT iLine = GetDlgItemInt(hwndDialog, ID_LINENUMBER, NULL, FALSE);
```
HBelusca, line 1143: It would be better to work with positive unsigned (UINT) line numbers (check https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getdlgitemint too)

base/applications/notepad/dialog.c
```diff
@@ -1117,82 +1117,90 @@ VOID DIALOG_Replace(VOID)
     DIALOG_SearchDialog(ReplaceText);
 }
 
+typedef struct tagGOTO_DATA
+{
+    INT iLine;
+    INT cLines;
```
HBelusca, line 1123: Prefer using UINT for both (see comment below).

base/applications/notepad/lang/en-US.rc
```diff
@@ -167,6 +167,7 @@ task.\nClose one or more applications to increase the amount of\nfree memory."
     STRING_LINE_COLUMN "Line %d, column %d"
     STRING_PRINTERROR "Cannot print the file '%s'.\n\nMake sure that the printer is turned on and is configured properly."
     STRING_DEFAULTFONT "Lucida Console"
+    STRING_LINE_NUMBER_OUT_OF_RANGE "The specified line number is out of range."
```
HBelusca, line 170: :+1: That's a better translation (and straight to the point) than Windows' one.

base/applications/notepad/dialog.c
katahiromz: OK.

base/applications/notepad/dialog.c
katahiromz: OK.

base/applications/notepad/lang/en-US.rc
katahiromz: Yes, we are making better Windows.

HBelusca (approved):

PR #4929, opened 2022-12-05
[KBSWITCH][SHLWAPI] Rewrite kbswitch

katahiromz:
## Purpose
JIRA issue: [CORE-10667](https://jira.reactos.org/browse/CORE-10667)

## Proposed changes

- `shlwapi!SHLoadRegUIStringW` expands the environmental strings.
- Load all the keyboard layouts at first.
- Remember console window's `HKL` by `SetProp`.
- Don't use Layout numbers but just use the `HKL` value.
- Use the layout list by `GetKeyboardLayoutList`.

## TODO

- [x] Fix hooks.
- [ ] Do tests.

PR #5085, opened 2023-02-18
[KBSWITCH] Really enable system hooking

katahiromz:
## Purpose
The current `master`'s implementation of `kbswitch` were not working in real system hooking. That old code just did a hack by using  `RegisterShellHookWindow`, that is not real hooking.

JIRA issue: [CORE-10667](https://jira.reactos.org/browse/CORE-10667)

Split from https://github.com/reactos/reactos/pull/4929.

## Proposed changes

- Use shared variables in `kbsdll.dll`.
- Delete the `RegisterShellHookWindow` hack.
- Enable the real system hooking.

## TODO

- [ ] Do tests.


katahiromz:
Unfortunately, global hook is still unstable.

HBelusca:
How does Windows (2000? 2003?) do?

katahiromz:
Postponed merging. Closed.

PR #5088, opened 2023-02-20
[NTOS:CM] Implement registry cache lookup & enforce cache consistency

GeoB99:
## Purpose
ReactOS implements a cache feature for the registry thanks to `CmpCacheTable` and various functions that handle it. With this, registry data can be cached on respective key control blocks (KCBs) so that such data can be acquired quickly. Except that the cached data isn't quite consistent.

The reason why the cache can become inconsistent is because we don't properly clean the subkey's information, such as when we are creating a link node or a new key. As a result the cached data can get out of sync and such data in a KCB might not reflect with the actual data of the subkey itself. Moreover, we barely do any cache lookup which means if a key path component is already in cache we will not ever acknowledge that. This PR, together with #4571, it embodies the "registry data integrity" goal which is what I'm trying to achieve.
## Proposed Changes
* Implement `CmpLookInCache`, the heart and brains of cache lookup;
* Implement `CmpBuildAndLockKcbArray` and `CmpComputeHashValue`. Together with these two, we can compute the hash values of each path name component of a registry key and lock the KCBs of each key component by their hash CONVKEY;
* Implement `CmpBuildHashStackAndLookupCache`;
* Don't touch read-only KCBs when creating a link node or just a mere key;
* Do a soft rewrite of `CmpDoOpen`, when using a KCB from cache we would want to expect a shared lock here. Also don't touch the registry node when we already have a KCB. Otherwise the passed KCB must be held with lock exclusively on new KCB creation;
* Add missing `CmpCleanUpSubKeyInfo` calls when performing a `create` procedure of a registry key object (key or link node);
* Add missing cleanup code paths in `CmpDoCreate`;
* Lock the hive flusher where appropriate, in `create` procedure, we don't want the lazy flusher to kick in when we are creating something on the target hive;
* Add missing KCB lock assertions;
* Fix a push lock acquiring recursion in `CmpSecurityMethod`, where it can happen that we might acquire a push lock (whether exclusive or shared) twice which it hangs the calling thread due to the nature of push locks;
* Document the code.
## TODO (for future PRs)
* Currently we don't handle `CM_KCB_SYM_LINK_FOUND`, a flag that indicates a symbolic link has been found (or resolved, whatever that is) which is necessary if a KCB is cached but such block is also a symbolic link. Perhaps `CmpGetSymbolicLink` might need some improvements.
* The registry parser of ReactOS doesn't handle fake keys. We are good because we don't create them anywhere anyway but when the time comes, `CmpParseKey` must require code adaptations to handle fake keys differently than the usual keys. FIXME comments have been added as reminders.
## JIRA Issues
[CORE-10581](https://jira.reactos.org/browse/CORE-10581)
[ROSTESTS-198](https://jira.reactos.org/browse/ROSTESTS-198)
## TODO
- [x] Fix a recursive lock in `CmpBuildAndLockKcbArray` (and other potential bugs I may find)
- [x] Run a testbot

### DIFF
--- a/ntoskrnl/config/cmkcbncb.c
+++ b/ntoskrnl/config/cmkcbncb.c
@@ -1134,6 +1134,13 @@ DelistKeyBodyFromKCB(IN PCM_KEY_BODY KeyBody,
     if (!LockHeld) CmpReleaseKcbLock(KeyBody->KeyControlBlock);
 }
 
+/**
+ * @brief
+ * Unlocks a number of KCBs provided by a KCB array.
+ *
+ * @param[in] KcbArray
+ * A pointer to an array of KCBs to be unlocked.
+ */
 VOID
 CmpUnLockKcbArray(
     _In_ PULONG KcbArray)
@@ -1147,6 +1154,24 @@ CmpUnLockKcbArray(
     }
 }
 
+/**
+ * @brief
+ * Locks a given number of KCBs.
+ *
+ * @param[in] KcbArray
+ * A pointer to an array of KCBs to be locked.
+ * The count of KCBs to be locked is defined by the
+ * first element in the array.
+ *
+ * @param[in] KcbLockFlags
+ * Define a lock flag to lock the KCBs.
+ *
+ * CMP_LOCK_KCB_ARRAY_EXCLUSIVE -- indicates the KCBs are locked
+ * exclusively and owned by the calling thread.
+ *
+ * CMP_LOCK_KCB_ARRAY_SHARED --  indicates the KCBs are locked
+ * in shared mode by the owning threads.
+ */
 static
 VOID
 CmpLockKcbArray(
@@ -1169,6 +1194,18 @@ CmpLockKcbArray(
     }
 }
 
+/**
+ * @brief
+ * Sorts an array of KCB hashes in ascending order
+ * and removes any key indices that are duplicates.
+ * The purpose of sorting the KCB elements is to
+ * ensure consistent and proper locking order, so
+ * that we can prevent a deadlock.
+ *
+ * @param[in,out] KcbArray
+ * A pointer to an array of KCBs of which the key
+ * indices are to be sorted.
+ */
 static
 VOID
 CmpSortKcbArray(
@@ -1216,6 +1253,50 @@ CmpSortKcbArray(
     KcbArray[0] = KcbCount;
 }
 
+/**
+ * @brief
+ * Builds an array of KCBs and locks them. Whether these
+ * KCBs are locked exclusively or in shared mode by the calling
+ * thread, is specified by the KcbLockFlags parameter. The array
+ * is sorted.
+ *
+ * @param[in] HashCacheStack
+ * A pointer to a hash cache stack. This stack parameter
+ * stores the convkey hashes of interested KCBs of a
+ * key path name that need to be locked.
+ *
+ * @param[in] KcbLockFlags
+ * Define a lock flag to lock the KCBs. Consult the
+ * CmpLockKcbArray documentation for more information.
+ *
+ * @param[in] Kcb
+ * A pointer to a key control block to be given. This
+ * KCB is included in the array for locking, that is,
+ * given by the CmpParseKey from the parser object.
+ *
+ * @param[in,out] OuterStackArray
+ * A pointer to an array that lives on the caller's
+ * stack. It acts like an auxiliary array used by
+ * the function to store the KCB elements for locking.
+ * The expected size of the array is up to 32 elements,
+ * which is the imposed limit by CMP_HASH_STACK_LIMIT.
+ * This limit also corresponds to the maximum depth of
+ * subkey levels.
+ *
+ * @param[in] TotalRemainingSubkeys
+ * The number of total remaining subkey levels.
+ *
+ * @param[in] MatchRemainSubkeyLevel
+ * The number of remaining subkey levels that match.
+ *
+ * @return
+ * Returns a pointer to an array of KCBs that have been
+ * locked.
+ *
+ * @remarks
+ * The caller HAS THE RESPONSIBILITY to unlock the KCBs
+ * after the necessary operations are done!
+ */
 PULONG
 NTAPI
 CmpBuildAndLockKcbArray(
--- a/ntoskrnl/config/cmkcbncb.c
+++ b/ntoskrnl/config/cmkcbncb.c
@@ -1135,6 +1135,154 @@ DelistKeyBodyFromKCB(IN PCM_KEY_BODY KeyBody,
 }
 
 VOID
+CmpUnLockKcbArray(
+    _In_ PULONG KcbArray)
+{
+    ULONG i;
+
+    /* Release the locked KCBs in reverse order */
+    for (i = KcbArray[0]; i > 0; i--)
+    {
+        CmpReleaseKcbLockByIndex(KcbArray[i]);
+    }
+}
+
+static
+VOID
+CmpLockKcbArray(
+    _In_ PULONG KcbArray,
+    _In_ ULONG KcbLockFlags)
+{
+    ULONG i;
+
+    /* Lock the KCBs */
+    for (i = 1; i <= KcbArray[0]; i++)
+    {
+        if (KcbLockFlags & CMP_LOCK_KCB_ARRAY_EXCLUSIVE)
+        {
+            CmpAcquireKcbLockExclusiveByIndex(KcbArray[i]);
+        }
+        else // CMP_LOCK_KCB_ARRAY_SHARED
+        {
+            CmpAcquireKcbLockSharedByIndex(KcbArray[i]);
+        }
+    }
+}
+
+static
+VOID
+CmpSortKcbArray(
+    _Inout_ PULONG KcbArray)
+{
+    ULONG i, j, k, KcbCount;
+
+    /* Ensure we don't go above the limit of KCBs we can hold */
+    KcbCount = KcbArray[0];
+    ASSERT(KcbCount < CMP_KCBS_IN_ARRAY_LIMIT);
+
+    /* Exchange-Sort the array in ascending order. Complexity: O[n^2] */
+    for (i = 1; i <= KcbCount; i++)
+    {
+        for (j = i + 1; j <= KcbCount; j++)
+        {
+            if (KcbArray[i] > KcbArray[j])
+            {
+                ULONG Temp = KcbArray[i];
+                KcbArray[i] = KcbArray[j];
+                KcbArray[j] = Temp;
+            }
+        }
+    }
+
+    /* Now remove any duplicated indices on the sorted array if any */
+    for (i = 1; i <= KcbCount; i++)
+    {
+        for (j = i + 1; j <= KcbCount; j++)
+        {
+            if (KcbArray[i] == KcbArray[j])
+            {
+                for (k = j; k <= KcbCount; k++)
+                {
+                    KcbArray[k - 1] = KcbArray[k];
+                }
+
+                j--;
+                KcbCount--;
+            }
+        }
+    }
+
+    /* Update the KCB count */
+    KcbArray[0] = KcbCount;
+}
+
+PULONG
+NTAPI
+CmpBuildAndLockKcbArray(
+    _In_ PCM_HASH_CACHE_STACK HashCacheStack,
+    _In_ ULONG KcbLockFlags,
+    _In_ PCM_KEY_CONTROL_BLOCK Kcb,
+    _Inout_ PULONG OuterStackArray,
+    _In_ ULONG TotalRemainingSubkeys,
+    _In_ ULONG MatchRemainSubkeyLevel)
+{
+    ULONG KcbIndex = 1, HashStackIndex, TotalRemaining;
+    PULONG LockedKcbs = NULL;
+    PCM_KEY_CONTROL_BLOCK ParentKcb = Kcb->ParentKcb;;
+
+    /* These parameters are expected */
+    ASSERT(HashCacheStack != NULL);
+    ASSERT(Kcb != NULL);
+    ASSERT(OuterStackArray != NULL);
+
+    /*
+     * Ensure when we build an array of KCBs to lock, that
+     * we don't go beyond the boundary the limit allows us
+     * to. 1 is the current KCB we would want to lock
+     * alongside with the remaining key levels in the formula.
+     */
+    TotalRemaining = (1 + TotalRemainingSubkeys) - MatchRemainSubkeyLevel;
+    ASSERT(TotalRemaining <= CMP_KCBS_IN_ARRAY_LIMIT);
+
+    /* Count the parent if we have one */
+    if (ParentKcb)
+    {
+        /* Ensure we are still below the limit and add the parent to KCBs to lock */
+        if (TotalRemainingSubkeys == MatchRemainSubkeyLevel)
+        {
+            TotalRemaining++;
+            ASSERT(TotalRemaining <= CMP_KCBS_IN_ARRAY_LIMIT);
+            OuterStackArray[KcbIndex++] = GET_HASH_INDEX(ParentKcb->ConvKey);
+        }
+    }
+
+    /* Add the current KCB */
+    OuterStackArray[KcbIndex++] = GET_HASH_INDEX(Kcb->ConvKey);
+
+    /* Loop over the hash stack and grab the hashes for locking (they will be converted to indices) */
+    for (HashStackIndex = 0;
+         HashStackIndex < TotalRemainingSubkeys;
+         HashStackIndex++)
+    {
+        OuterStackArray[KcbIndex++] = GET_HASH_INDEX(HashCacheStack[HashStackIndex].ConvKey);
+    }
+
+    /*
+     * Store how many KCBs we need to lock and sort the array.
+     * Remove any duplicated indices from the array if any.
+     */
+    OuterStackArray[0] = KcbIndex - 1;
+    CmpSortKcbArray(OuterStackArray);
+
+    /* Lock them */
+    CmpLockKcbArray(OuterStackArray, KcbLockFlags);
+
+    /* Give the locked KCBs to caller now */
+    LockedKcbs = OuterStackArray;
+    return LockedKcbs;
+}
+
+VOID
 NTAPI
 CmpFlushNotifiesOnKeyBodyList(IN PCM_KEY_CONTROL_BLOCK Kcb,
                               IN BOOLEAN LockHeld)
--- a/ntoskrnl/config/cmparse.c
+++ b/ntoskrnl/config/cmparse.c
@@ -284,6 +284,7 @@ CmpDoCreateChild(IN PHHIVE Hive,
     KeyBody = (PCM_KEY_BODY)(*Object);
     KeyBody->Type = CM_KEY_BODY_TYPE;
     KeyBody->KeyControlBlock = NULL;
+    KeyBody->KcbLocked = FALSE;
 
     /* Check if we had a class */
     if (ParseContext->Class.Length > 0)
@@ -702,6 +703,15 @@ CmpDoOpen(IN PHHIVE Hive,
         /* Link to the KCB */
         EnlistKeyBodyWithKCB(KeyBody, 0);
 
+        /*
+         * We are already holding a lock against the KCB that is assigned
+         * to this key body. This is to prevent a potential deadlock on
+         * CmpSecurityMethod as ObCheckObjectAccess will invoke the Object
+         * Manager to call that method, of which CmpSecurityMethod would
+         * attempt to acquire a lock again.
+         */
+        KeyBody->KcbLocked = TRUE;
+
         if (!ObCheckObjectAccess(*Object,
                                  AccessState,
                                  FALSE,
@@ -711,6 +721,12 @@ CmpDoOpen(IN PHHIVE Hive,
             /* Access check failed */
             ObDereferenceObject(*Object);
         }
+
+        /*
+         * We are done, the lock we are holding will be released
+         * once the registry parsing is done.
+         */
+        KeyBody->KcbLocked = FALSE;
     }
     else
     {
--- a/ntoskrnl/config/cmparse.c
+++ b/ntoskrnl/config/cmparse.c
@@ -1111,6 +1111,31 @@ CmpHandleExitNode(IN OUT PHHIVE *Hive,
     }
 }
 
+/**
+ * @brief
+ * Computes the hashes of each subkey in key path name
+ * and stores them in a hash stack for cache lookup.
+ *
+ * @param[in] RemainingName
+ * A Unicode string structure consisting of the remaining
+ * registry key path name.
+ *
+ * @param[in] ConvKey
+ * The hash convkey of the current KCB to be supplied.
+ *
+ * @param[in,out] HashCacheStack
+ * An array stack. This function uses this array to store
+ * all the computed hashes of a key pathname.
+ *
+ * @param[out] TotalSubKeys
+ * The number of total subkeys that have been found, returned
+ * by this function to the caller. If no subkey levels are found
+ * the function returns 0.
+ *
+ * @return
+ * Returns the number of remaining subkey levels to caller.
+ * If no subkey levels are found then this function returns 0.
+ */
 static
 ULONG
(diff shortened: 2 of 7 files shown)


ntoskrnl/config/cmparse.c
```diff
@@ -502,8 +524,21 @@ CmpDoCreate(IN PHHIVE Hive,
         {
             /* Fail, this shouldn't happen */
             ASSERT(FALSE);
```
HBelusca, line 526: Now that you've implemented that failure code path, that ASSERT(FALSE) may be removed.

ntoskrnl/config/cmparse.c
```diff
@@ -539,6 +574,16 @@ CmpDoCreate(IN PHHIVE Hive,
             {
                 /* This shouldn't happen */
                 ASSERT(FALSE);
```
HBelusca, line 576: Ditto

ntoskrnl/config/cmkcbncb.c
```diff
@@ -1134,6 +1134,209 @@ DelistKeyBodyFromKCB(IN PCM_KEY_BODY KeyBody,
...
+         */
+        if (KcbIndex >= CMP_HASH_STACK_LIMIT)
+        {
+            DPRINT1("Limit to store KCBs for locking has been REACHED\n");
+            break;
+        }
+    }
+
+    /* Lock them now (store how many KCBs we need to lock) */
+    OuterStackArray[0] = KcbIndex;
```
ThFabba, line 1332: Should be `KcbIndex - 1`, since `KcbIndex` is one more than the count of items.

ntoskrnl/config/cmkcbncb.c
```diff
@@ -1134,6 +1134,209 @@ DelistKeyBodyFromKCB(IN PCM_KEY_BODY KeyBody,
...
+        /* Go to the next hash */
+        HashStackIndex++;
+
+        /*
+         * Make sure we don't blow up the outer stack as we
+         * are adding up hashes for KCBs to lock.
+         */
+        if (KcbIndex >= CMP_HASH_STACK_LIMIT)
+        {
+            DPRINT1("Limit to store KCBs for locking has been REACHED\n");
```
ThFabba, line 1326: I think this should be impossible because keys nested that deeply are forbidden? I.e. this can probably be an ASSERT?

ntoskrnl/config/cmse.c
```diff
@@ -278,14 +279,26 @@ CmpSecurityMethod(IN PVOID ObjectBody,
...
-    /* Acquire the KCB lock */
+    /*
+     * Acquire the KCB lock but only if it has not been
+     * ever locked before (this is to avoid push lock
+     * recursion).
+     */
     if (OperationCode == QuerySecurityDescriptor)
     {
-        CmpAcquireKcbLockShared(Kcb);
+        if (!CmpIsKcbLocked(Kcb))
```
ThFabba, line 289: This check implies that locking is optional. If so, just don't lock.
But I'm guessing it's not really optional and this is working around some bug. Could you elaborate?

GeoB99: There is the risk that we can deadlock inside the `CmpSecurityMethod` function if it happens that the KCB of the object body already has a push lock held and somebody still wants to acquire it. The chance of a potential deadlock to occur is higher, for example, when we are holding a lock ourselves while doing a cache lookup during a parse procedure of the registry, or in the other words, somebody wants to query security information of a key object meanwhile we are holding the lock. On top of that, registry parsing is a very common occurrence further increasing such risk.

Why is this of a problem now but not before is because we are going to treat KCB locks seriously, since ReactOS misses some necessary locks here and there in the registry parser. And there's code that lives up on the fact that nobody was holding a lock. This is exactly what it happened to me with `CmpSecurityMethod`, when I was trying to debug the first deadlock I noticed that this function was called in `CmpDoCreateChild` with a KCB lock already held.

Now that I am discussing this, perhaps the correct fix would be is to remove the `CmpSecurityMethod` call in `CmpDoCreateChild`  and directly invoke `CmpAssignSecurityDescriptor`. It makes sense because `CmpSecurityMethod` is a construct used by OB when creating the key type object with `ObCreateObjectType` so nobody else should be calling `CmpSecurityMethod` but only OB when it sees fit. Also it seems like Windows doesn't use `CmpSecurityMethod` when creating a link node but rather it just calls `CmpAssignSecurityDescriptor`. Although you would still need to check if someone holds a lock against the KCB in `CmpSecurityMethod`, if the requested operation is to query info from a security descriptor that is.

ThFabba: Yeah, removing direct calls to `CmpSecurityMethod` definitely makes sense to me.
If you wanted (and were able) to check "is the current thread already holding this lock" to avoid acquiring it recursively, that could make sense (though I would hope that there'd be more elegant solutions).
But the current check is: if anyone is already holding this lock, skip acquiring it -- which entirely negates the purpose of the lock, so it can't possibly be a correct fix.

GeoB99: We have `CmpIsKcbLockedExclusive` which checks if the ownership of the lock is held by the calling thread, I don't see any other solutions that could see fit. If you have anything else that comes in your mind then suggestions are welcome.

ntoskrnl/config/cmparse.c
```diff
@@ -602,34 +651,90 @@ CmpDoOpen(IN PHHIVE Hive,
...
+                     * The real KCB is gone, do a reparse. We used to lock the KCB in
+                     * shared mode as others may have taken use of it but since we
+                     * must do a reparse of the key the only thing that matter is us.
+                     * Lock the KCB exclusively so nobody is going to mess with the KCB.
+                     */
+                    DPRINT1("The real KCB is deleted, attempt a reparse\n");
+                    CmpUnLockKcbArray(KcbsLocked);
+                    CmpAcquireKcbLockExclusiveByKey((*CachedKcb)->ConvKey);
+                    CmpCleanUpKcbValueCache(*CachedKcb);
+                    KcbsLocked[0] = 1;
```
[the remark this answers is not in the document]
GeoB99: You're correct. The first position of the array holds the number of locked KCBs, the rest of the elements in the array are the KCB convkey hashes.

ntoskrnl/config/cmkcbncb.c
```diff
@@ -1134,6 +1134,294 @@ DelistKeyBodyFromKCB(IN PCM_KEY_BODY KeyBody,
...
+VOID
+NTAPI
+CmpUnLockKcbArray(
+    _In_ PULONG KcbArray)
+{
+    ULONG i, KcbCount;
+
+    /* Release the locked KCBs */
+    KcbCount = KcbArray[0];
+    for (i = KcbCount; i > 0; i--)
```
HBelusca, line 1153:
```suggestion
    for (i = KcbArray[0]; i > 0; i--)
```

ntoskrnl/config/cmkcbncb.c
```diff
@@ -1134,6 +1134,294 @@ DelistKeyBodyFromKCB(IN PCM_KEY_BODY KeyBody,
...
+VOID
+NTAPI
+CmpLockKcbArray(
+    _In_ PULONG KcbArray,
+    _In_ ULONG KcbLockFlags)
+{
+    ULONG i;
+
+    /* Lock the KCBs */
+    for (i = 0; i < KcbArray[0]; i++)
```
HBelusca, line 1183: What about
```suggestion
    for (i = 1; i <= KcbArray[0]; i++)
```
and just using `KcbArray[i]` in the calls below?

GeoB99:
BEFORE:
Capture2
BEFORE (after ignoring the assert):
Capture
AFTER:
Capture

Test KVM: https://build.reactos.org/#/builders/6/builds/32959
Test Vbox: https://build.reactos.org/#/builders/3/builds/20428

It looks like the bots are complaining for whatever reason.

ntoskrnl/include/internal/cm.h
```diff
@@ -119,6 +125,12 @@
 #define CM_DELAYS_PER_PAGE                              \
     ((PAGE_SIZE - FIELD_OFFSET(CM_ALLOC_PAGE, AllocPage)) / sizeof(CM_DELAY_ALLOC))
 
+//
+// Cache Lookup & KCB Array constructs
+//
+#define CMP_SUBKEY_LEVELS_DEPTH_LIMIT   32
+#define CMP_KCBS_IN_ARRAY_LIMIT         CMP_SUBKEY_LEVELS_DEPTH_LIMIT + 2
```
HBelusca, line 132:
```suggestion
#define CMP_KCBS_IN_ARRAY_LIMIT         (CMP_SUBKEY_LEVELS_DEPTH_LIMIT + 2)
```

ntoskrnl/config/cmparse.c
```diff
@@ -989,40 +1127,711 @@ CmpHandleExitNode(IN OUT PHHIVE *Hive,
...
+    LONG RemainingSubkeys;
+    ULONG TotalLevels;
+    BOOLEAN SubkeysMatch;
+    PCM_KEY_CONTROL_BLOCK CurrentKcb, ParentKcb;
+    PCM_KEY_HASH HashEntry = NULL;
+    BOOLEAN KeyFoundInCache = FALSE;
+    PULONG LockedKcbs = NULL;
+
+    /* Reference the KCB */
+    (VOID)CmpReferenceKeyControlBlock(*Kcb);
```
HBelusca, line 1490: I'm not sure that ignoring the returned value is a good thing (especially since on return, the kcb is dereferenced).

ntoskrnl/config/cmparse.c
```diff
@@ -989,40 +1129,731 @@ CmpHandleExitNode(IN OUT PHHIVE *Hive,
...
+        }
+    }
+
+    /*
+     * Check if we have a full match of remaining levels.
+     *
+     * FIXME: It is possible we can catch a fake key from the cache
+     * when we did the lookup, in such case we should not do any
+     * locking as such KCB does not point to any real information.
+     * Currently ReactOS doesn't create fake KCBs so we are good
```
HBelusca, line 1823: What are those "fake" KCBs about?

GeoB99: A key control block has to belong to a registry key that actually must exist in the registry tree, fake keys are the opposite in the spectrum in the sense that they are not real (call them "imaginary" or whatever). We have a very few mentions of this such as `CM_KCB_KEY_NON_EXIST` and `IsFake`constructs (e.g. grep in https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/config/cmkcbncb.c;hb=f8c7bd6898ec15bb8420b7c7a2683759b67ddcd7#l146).

I don't know exactly what's the point of these, perhaps they are simply just dummy KCBs that serve only specific internal kernel purposes. I just discovered that such fake keys are handled in the Windows' registry parser whilst doing some reverse engineering so I thought of adding this FIXME comment as a remainder for whoever wants to work on such thing in the future.

ntoskrnl/include/internal/cm_x.h
```diff
@@ -83,6 +83,14 @@
     (GET_HASH_ENTRY(CmpCacheTable,                                  \
                     (k)->ConvKey)->Owner == KeGetCurrentThread())
 
+//
```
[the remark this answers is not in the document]
GeoB99: Yes, sure.

GeoB99:
Ok, after fixing some bugs here are the new test results (the previous screenshots don't matter anymore).

BEFORE (assertion hit, continuable):
Capture2
AFTER:
Capture

Test KVM: https://build.reactos.org/#/builders/6/builds/32970
Test Vbox: https://build.reactos.org/#/builders/3/builds/20439

Test results: https://reactos.org/testman/compare.php?ids=86161,86165

ntoskrnl/config/cmparse.c
```diff
@@ -989,40 +1113,731 @@ CmpHandleExitNode(IN OUT PHHIVE *Hive,
...
-                                OUT PULONG *LockedKcbs)
+CmpLookInCache(
+    _In_ PCM_HASH_CACHE_STACK HashCacheStack,
+    _In_ BOOLEAN LockKcbsExclusive,
+    _In_ ULONG TotalRemainingSubkeys,
+    _Inout_ PUNICODE_STRING RemainingName,
+    _Inout_ PULONG OuterStackArray,
+    _Inout_ PCM_KEY_CONTROL_BLOCK *Kcb,
+    _Out_ PHHIVE *Hive,
+    _Out_ HCELL_INDEX *Cell,
```
HBelusca, line 1465:
```suggestion
    _Out_ PHCELL_INDEX Cell,
```

ntoskrnl/config/cmparse.c
```diff
@@ -989,40 +1113,731 @@ CmpHandleExitNode(IN OUT PHHIVE *Hive,
...
+        if (!CmpReferenceKeyControlBlock(*Kcb))
+        {
+            /* This key is opened too many times, bail out */
+            DPRINT1("Could not reference the KCB, too many references (KCB 0x%p)\n", Kcb);
+            return STATUS_UNSUCCESSFUL;
+        }
 
-    /* Reference it */
-    (VOID)CmpReferenceKeyControlBlock(*Kcb);
+        CmpAcquireKcbLockSharedByIndex(GET_HASH_INDEX((*Kcb)->ConvKey));
```
HBelusca, line 1749: Does the Kcb change within this whole CmpBuildHashStackAndLookupCache ? If not, it seems you could reuse that `ConvKey` variable you introduced above and in which you have stored that `(*Kcb)->ConvKey` value.

ntoskrnl/config/cmkcbncb.c
```diff
@@ -1134,6 +1134,271 @@ DelistKeyBodyFromKCB(IN PCM_KEY_BODY KeyBody,
...
+    for (i = 1; i <= KcbCount; i++)
+    {
+        for (j = i + 1; j <= KcbCount; j++)
+        {
+            if (KcbArray[i] == KcbArray[j])
+            {
+                for (k = j; k <= KcbCount; k++)
+                {
+                    KcbArray[k] = KcbArray[k + 1];
+                }
```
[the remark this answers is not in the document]
GeoB99: Is this a _hypothetical_  problem or an actual problem you think that could practically raise to bugs? If duplicated indices happen to be at the end of the array, the duplicated KCB index will be moved one location ahead out of the array. Looping the KCB array before and after the removal of duplicated indices, and displaying the actual indices, reveals the following scenario:

```
(ntoskrnl\config\cmkcbncb.c:1232) ===
(ntoskrnl\config\cmkcbncb.c:1235) ARRAY ELEMENT BEFORE REMOVAL 1227
(ntoskrnl\config\cmkcbncb.c:1235) ARRAY ELEMENT BEFORE REMOVAL 1235
(ntoskrnl\config\cmkcbncb.c:1235) ARRAY ELEMENT BEFORE REMOVAL 1688
(ntoskrnl\config\cmkcbncb.c:1235) ARRAY ELEMENT BEFORE REMOVAL 1688
(ntoskrnl\config\cmkcbncb.c:1237) ===

(ntoskrnl\config\cmkcbncb.c:1265) ===
(ntoskrnl\config\cmkcbncb.c:1268) ARRAY ELEMENT AFTER REMOVAL 1227
(ntoskrnl\config\cmkcbncb.c:1268) ARRAY ELEMENT AFTER REMOVAL 1235
(ntoskrnl\config\cmkcbncb.c:1268) ARRAY ELEMENT AFTER REMOVAL 1688
(ntoskrnl\config\cmkcbncb.c:1270) ===
```

ntoskrnl/config/cmkcbncb.c
```diff
@@ -1134,6 +1134,271 @@ DelistKeyBodyFromKCB(IN PCM_KEY_BODY KeyBody,
...
 
+/**
+ * @brief
+ * Unlocks a number of KCBs provided by a KCB array.
+ *
+ * @param[in] KcbArray
+ * A pointer to an array of KCBs to be unlocked.
+ */
+VOID
+NTAPI
```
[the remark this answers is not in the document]
GeoB99: `CmpUnLockKcbArray` is used outside of `cmkcbncb.c` so I would argue keeping the NT annotation. I'm just pretty much following the exact same style that is currently in force in CM.

I do definitely agree functions like `CmpLockKcbArray` and `CmpSortKcbArray` which are used only here could be promoted as static functions though.

ntoskrnl/config/cmkcbncb.c
```diff
@@ -1134,6 +1134,271 @@ DelistKeyBodyFromKCB(IN PCM_KEY_BODY KeyBody,
...
+ *
+ * @param[in] TotalRemainingSubkeys
+ * The number of total remaining subkey levels.
+ *
+ * @param[in] MatchRemainSubkeyLevel
+ * The number of remaining subkey levels that match.
+ */
+VOID
+NTAPI
+CmpValidateKeyPathForKcbArray(
```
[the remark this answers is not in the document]
GeoB99: As the name implies, it validates the key path ensuring that it does have a proper key depth level (aka the key must not be super long in length). I should probably move these asserts in `CmpBuildAndLockKcbArray`.

ntoskrnl/config/cmkcbncb.c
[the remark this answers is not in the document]
HBelusca: Hypothetical or not is not the question, as long as it is a possibility not forbidden by any other constraint.
When you run the code on your example-array, using step-by-step in windbg, what is happening in that loop during the loop? Is the k index going through 4 or not?

ntoskrnl/config/cmparse.c
```diff
@@ -284,6 +284,7 @@ CmpDoCreateChild(IN PHHIVE Hive,
     KeyBody = (PCM_KEY_BODY)(*Object);
     KeyBody->Type = CM_KEY_BODY_TYPE;
     KeyBody->KeyControlBlock = NULL;
+    KeyBody->KcbLocked = FALSE;
```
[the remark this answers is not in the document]
GeoB99: As we talked privately recently, the `KcbLocked` field was introduced when we discussed about an approach in kernel-dev channel. I plan to have this kind of stuff rewritten to follow Windows 10 way, by using flags.

ntoskrnl/config/cmkcbncb.c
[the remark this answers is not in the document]
HBelusca: ASSERTs are just debugging helpers happening only for debug builds. Meaning that if you compile in Release mode, this whole "validation" function is basically doing nothing.
(I remember that a similar discussion already happened in a past PR of yours for either Cm or Se.)

ntoskrnl/config/cmsysini.c
```diff
@@ -138,7 +138,7 @@ CmpDeleteKeyObject(PVOID DeletedObject)
         if (Kcb)
         {
             /* Delist the key */
-            DelistKeyBodyFromKCB(KeyBody, FALSE);
+            DelistKeyBodyFromKCB(KeyBody, KeyBody->KcbLocked ? TRUE : FALSE);
```
HBelusca, line 141: If the KcbLocked is already a boolean, there shouldn't be any need to do the `? TRUE : FALSE`.

ntoskrnl/config/cmkcbncb.c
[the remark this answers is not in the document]
HBelusca: > CmpUnLockKcbArray is used outside of cmkcbncb.c so I would argue keeping the NT annotation

The "NTAPI" thing is a calling convention that doesn't need to be enforced if the functions are not exported and not need a precise signature as in the cases I mentioned above.

ntoskrnl/config/cmparse.c
```diff
@@ -989,40 +1113,731 @@ CmpHandleExitNode(IN OUT PHHIVE *Hive,
...
+        else
+        {
+            /* We have a leading slash here, skip it */
+            while (*RemainingNameBuffer == OBJ_NAME_PATH_SEPARATOR)
+            {
+                RemainingNameBuffer++;
+                RemainingNameLength -= sizeof(WCHAR);
+            }
+
+            /* Take into account this subkey */
```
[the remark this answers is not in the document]
GeoB99: Generally this should never happen. `RemainingNameLength` is decreased only when there are characters in a subkey whose hash values are still to be computed, When the length reaches 0 we have computed the hash values of the whole key pathname. 

Perhaps adding an ASSERT there could be worthwhile.

ntoskrnl/config/cmparse.c
```diff
@@ -989,40 +1113,731 @@ CmpHandleExitNode(IN OUT PHHIVE *Hive,
...
+
+                        /* Remove the subkeys in the remaining name and stop looking in the cache */
+                        CmpRemoveSubkeysInRemainingName(HashCacheStack, RemainingSubkeys, RemainingName);
+                        break;
+                    }
+                }
+            }
+
+            /* Go to the next hash */
+            HashEntry = HashEntry->NextHash;
```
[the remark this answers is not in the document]
GeoB99: I don't think so.

ntoskrnl/config/cmparse.c
```diff
@@ -989,40 +1113,731 @@ CmpHandleExitNode(IN OUT PHHIVE *Hive,
...
+ * @return
+ * Returns TRUE if all the subkey levels match, otherwise
+ * FALSE is returned.
+ */
+BOOLEAN
+NTAPI
+CmpCompareSubkeys(
+    _In_ PCM_HASH_CACHE_STACK HashCacheStack,
+    _In_ PCM_KEY_CONTROL_BLOCK CurrentKcb,
+    _In_ LONG RemainingSubkeys,
```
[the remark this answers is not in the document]
GeoB99: To summarize:

`TotalSubkeys` -- The number of total subkeys in a key pathname (e.g. `Machine\BlaBla\lololo` would have 3 subkeys in total);

`TotalRemainingSubkeys` -- Same as `TotalSubkeys`, although this variable is used in conjunction with the hash cache stack. Each subkey that is added into the stack is counted in this variable. See `CmpComputeHashValue`;

`MatchRemainSubkeyLevel` -- The number of subkey levels that match with each other.

With that said, think of `RemainingSubkeys` as a construct that is used to walk through the cache stack.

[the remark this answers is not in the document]
GeoB99: To summarize:

`TotalSubkeys` -- The number of total subkeys in a key pathname (e.g. `Machine\BlaBla\lololo` would have 3 subkeys in total);

`TotalRemainingSubkeys` -- Same as `TotalSubkeys`, although this variable is used in conjunction with the hash cache stack. Each subkey that is added into the stack is counted in this variable. See `CmpComputeHashValue`;

`MatchRemainSubkeyLevel` -- The number of subkey levels that match with each other.

With that said, think of `RemainingSubkeys` as a construct that is used to walk through the cache stack.

ntoskrnl/config/cmparse.c
```diff
@@ -989,40 +1113,731 @@ CmpHandleExitNode(IN OUT PHHIVE *Hive,
...
+        RemainingNameBuffer++;
+        RemainingNameLength -= sizeof(WCHAR);
+    }
+
+    /* Now set up the hash stack entries and compute the hashes */
+    SubkeysInTotal = 0;
+    RemainingSubkeysInTotal = 0;
+    KeyNameLength = 0;
+    CopyConvKey = ConvKey;
+    HashCacheStack[RemainingSubkeysInTotal].NameOfKey.Buffer = RemainingNameBuffer;
```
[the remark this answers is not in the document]
HBelusca: > Suppose a key name with no path-separators was passed such as DisplayController for example

Yes of course. But now suppose that name is passed instead: `\\\\` ?  I mean I suppose other functions called before that one should have already handled such a case. But if such name is directly fed into that function, you get:
- original `RemainingNameLength` being greater than zero;
- the while-loop skips over the slashes and decreases `RemainingNameLength`. At the end of the while-loop, all slashes are skept over and ``RemainingNameLength` is now == 0 ;
- then that code (below the loop) gets run.

I guess you could also add an ASSERT to get some possible troubleshooting in debug builds (similarly to what you suggested in the other comment below).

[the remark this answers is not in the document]
GeoB99: I added an ASSERT in the `else` branch below, this should detect this kind of situation where a malformed path was passed. In any case this is a serious bug in the kernel anyway.

jeditobe:
amazing

PR #5091, opened 2023-02-20
[RAPPS] Style-addendum, no functional change

JoachimHenze:
## Purpose

This is a small style-addendum to 33c2903e6d4fda32215f553c2189c13d52ed31de
Absolutely no functional change.

## Proposed changes

- The 4 files have been renamed today by the mentioned commit 33c2903e6d4fda32215f553c2189c13d52ed31de . Therefore the timepoint is ideal now to give the files and classes a shorter name which doesn't sacrifice any of their distinct meaning.
- Benefit of that is that we have shorter logs, therefore less linebreaks in the GCC logs also.
- In the long run this will also avoid linebreaks within the code itself
- It will shrink the binary sizes (due to function and classnames being compiled into them in dbg)
- In last commit I followed Mark Jansens example to not space-pad before the email addresses, because that will lead to frequent grooming of all lines whenever a developer with a long name would contribute anything in the future

## Sidenote

Choosing good names is no premature optimization btw.
Optimization gets easier the more close you are at the destination of the final thing function-wise.
Contrary 'changing names' gets harder, the more code you wrote already.
After todays renaming the timepoint is ideal for this addendum PR.


JoachimHenze:
Ready for review.

katahiromz (approved):

learn-more (dismissed):
As I already mentioned to you, I think this kinda renaming is pointless and does not improve readability.

JoachimHenze:
Mark I do kind of understand, that you 'do not really care about such minor things', but I did clearly point out the objective benefits. So unless you can prove there to be some bug/mistake that I did, the least thing that I should be able to expect from you is neutrality for my PR. I don't think that rejection (change-request without me being able to action on it) would be justified in any way. Can we behave like adults please? Will that PR make anything worse for you? If not, why blocking it?

PR #5094, opened 2023-02-21
[INBV] Omit black screen fill at epilogue of boot

katahiromz:
## Purpose

Speed up!
JIRA issue: N/A

## Proposed changes

- Do not fill the screen by black at `FinalizeBootLogo` function.

## TODO

- [x] Do tests.

### DIFF
--- a/ntoskrnl/inbv/bootanim.c
+++ b/ntoskrnl/inbv/bootanim.c
@@ -755,12 +755,6 @@ FinalizeBootLogo(VOID)
 {
     /* Acquire lock and check the display state */
     InbvAcquireLock();
-    if (InbvGetDisplayState() == INBV_DISPLAY_STATE_OWNED)
-    {
-        /* Clear the screen */
-        VidSolidColorFill(0, 0, SCREEN_WIDTH-1, SCREEN_HEIGHT-1, BV_COLOR_BLACK);
-    }
-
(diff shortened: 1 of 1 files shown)


ntoskrnl/inbv/bootanim.c
```diff
@@ -755,12 +755,6 @@ FinalizeBootLogo(VOID)
 {
     /* Acquire lock and check the display state */
     InbvAcquireLock();
-    if (InbvGetDisplayState() == INBV_DISPLAY_STATE_OWNED)
-    {
-        /* Clear the screen */
-        VidSolidColorFill(0, 0, SCREEN_WIDTH-1, SCREEN_HEIGHT-1, BV_COLOR_BLACK);
-    }
-
```
GeoB99, line 763: How does this optimize the speeds of the boot phase? Have you recorded a performance sample against this or something?

This is just my consideration but I don't see what would we gain from not clearing the display, plus it is kind of odd you don't want to clear the display once the bootstrap has been finalized. If you suspect `VidSolidColorFill` is poorly optimized then I would suggest that you fix the root of the problem instead (aka in the function itself).

HBelusca (changes requested):
PR is invalid. The screen cleanup is there for a reason.

katahiromz:
Closed.

PR #5095, opened 2023-02-21
[USETUP][INFLIB] Don't use HEAP_ZERO_MEMORY for malloc

katahiromz:
## Purpose

The `malloc` function doesn't need `HEAP_ZERO_MEMORY` flag.
JIRA issue: [CORE-18838](https://jira.reactos.org/browse/CORE-18838)

## TODO

- [x] Do tests.

### DIFF
--- a/base/setup/usetup/spapisup/cabinet.c
+++ b/base/setup/usetup/spapisup/cabinet.c
@@ -140,7 +140,7 @@ typedef struct _CFDATA
 void *__cdecl
 malloc(size_t size)
 {
-    return RtlAllocateHeap(ProcessHeap, HEAP_ZERO_MEMORY, size);
+    return RtlAllocateHeap(ProcessHeap, 0, size);
 }
 
 void __cdecl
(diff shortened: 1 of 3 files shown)


learn-more:
For this change you'd need to audit _all_ code in usetup, because there might be code that depends on the fact that data was zero'd.

katahiromz:
I grep'ed `malloc` in `usetup`, then I found no `malloc` usage in the usetup module source code.

```txt
target_link_libraries(usetup inflib setuplib zlib_solo ext2lib vfatlib btrfslib chkstk)
```

`inflib`, `zlib_solo`, and `vfatlib` use `malloc`.
`setuplib`, `ext2lib`, and `chkstk` don't use `malloc`.

`zlib_solo` should be ok (this module was independent).
The problem might be in `inflib` and `vfatlib`.

katahiromz:
I found some problems in `inflib`. Fixes are added.

HBelusca:
"cabinet.c" mentions "Needed by zlib, but we don't want the dependency on the CRT", which is the malloc implementation that in principle would be only used by the linked zlib (since, inflib would here not use malloc but RtlAllocateHeap, via the macros it uses).
I suppose the zlib already correctly uses malloc and, if needed, manually zeroes the allocated buffer? (or directly uses calloc)

katahiromz:
Time up. Closed.

PR #5100, opened 2023-02-24
[FLTLIB] Reimplement FilterHResultFromNtStatus and FilterpDeviceIoControl

RatinCN:
## Purpose

Reimplement `FilterHResultFromNtStatus` and `FilterpDeviceIoControl` correctly

`FilterHResultFromNtStatus`: Translate `NTSTATUS` returned from Filter Manager to Win32 error style `HRESULT`
`FilterpDeviceIoControl`: Send IOCTL to Filter Manager, very similar to `DeviceIoControl`

JIRA issue: None

## Proposed changes

- Add Filter Manager related definitions in `ntstatus.h`
- Reimplement `FilterHResultFromNtStatus` and `FilterpDeviceIoControl` correctly
- Apply new style header format

## Test

Before:
Before
After:
After
They're the same, `fltmc` load minifilter by calling `FltLib.dll` works well without regression.

This PR have no functional changes but do the work in right way,
after this PR, the rest of unimplement FltLib APIs will be easy to complete.


JoachimHenze:
@RatinCN I do think this ticket may have a slight relation to https://jira.reactos.org/browse/CORE-14182 (most likely without being able to resolve that ticket). If you would agree, you could mention that JIRA ID in the description.

RatinCN:
> Out of curiosity: is this PR going to indirectly fix the behaviour of some apps that use the filter manager?

Sadly, effect is not obvious. As far as I know, following components are required to make our Minifilter mechanism works:

1. FltMgr.sys: The degree of completion seems not enough
2. FltLib.dll: User-mode interface, most of unimplement APIs are base on `FilterHResultFromNtStatus` and `FilterpDeviceIoControl`, after this PR, FltLib.dll will be easier to accomplish, and I'll move on it
3. Fltmc.exe: Cmdutil, we already have it and LGTM

sdk/include/psdk/ntstatus.h
```diff
@@ -234,6 +235,8 @@ extern "C" {
 #define STATUS_CLUSTER_NETWORK_ALREADY_OFFLINE  ((NTSTATUS)0x80130004)
 #define STATUS_CLUSTER_NODE_ALREADY_MEMBER      ((NTSTATUS)0x80130005)
 
+#define STATUS_FLT_BUFFER_TOO_SMALL             ((NTSTATUS)0x801C0001L)
```
JoachimHenze, line 238: Why the empty line at line237? Is there anything that would separate those 2 kind of STATUS_ defines into different classes? (Same question for the line 239)

HBelusca: Obviously. Above you have some status codes for the cluster component, then that one for the FS filter component.

RatinCN: I see empty line was used to separate different kind of STATUS, so do I.
BTW I noticed `L` should be removed to make it tidy, I'll commit it later.

RatinCN:
> @RatinCN I do think this ticket may have a slight relation to https://jira.reactos.org/browse/CORE-14182 (most likely without being able to resolve that ticket). If you would agree, you could mention that JIRA ID in the description.

How about to summary all the Minifilter related tickets into one, and create tasks or goals for fltmgr.sys and fltlib.dll? Nearly all the JIRA tickets about minifilter show they are incomplete, as Ged commented in this JIRA

dll/win32/fltlib/fltlib.c
```diff
@@ -109,48 +156,135 @@ FilterLoadUnload(_In_z_ LPCWSTR lpFilterName,
...
+    /* Check for asynchronous operation */
+    if (lpOverlapped)
     {
-        Status = NtWaitForSingleObject(Handle, FALSE, NULL);
-        if (NT_SUCCESS(Status))
+        /* Set pending status */
+        lpOverlapped->Internal = STATUS_PENDING;
+
+        /* Check if there's an APC context */
+        ApcContext = (ULONG_PTR)lpOverlapped->hEvent & 0x1 ? NULL : lpOverlapped;
```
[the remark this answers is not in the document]
RatinCN: I have no idea, but `FilterpDeviceIoControl` did, and [DeviceIoControl](https://github.com/reactos/reactos/blob/master/dll/win32/kernel32/client/file/deviceio.c#L166) also did

dll/win32/fltlib/fltlib.c
```diff
@@ -109,48 +156,135 @@ FilterLoadUnload(_In_z_ LPCWSTR lpFilterName,
...
+         */
+        if (!NT_ERROR(Status))
         {
-            Status = IoStatusBlock.Status;
+            *lpBytesReturned = (DWORD)IoStatusBlock.Information;
         }
     }
 
-    return RtlNtStatusToDosError(Status);
+    return Status == STATUS_SUCCESS ? S_OK : FilterHResultFromNtStatus(Status);
```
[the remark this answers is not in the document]
RatinCN: Committed

dll/win32/fltlib/fltlib.c
```diff
@@ -109,48 +156,135 @@ FilterLoadUnload(_In_z_ LPCWSTR lpFilterName,
...
+            Status = NtWaitForSingleObject(hFltMgr, FALSE, NULL);
+            if (NT_SUCCESS(Status))
+            {
+                Status = IoStatusBlock.Status;
+            }
+        }
+
+        /* Return the number of bytes transferred if no error,
+         * but seems Windows didn't check this
+         */
```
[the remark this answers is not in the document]
RatinCN: Committed

dll/win32/fltlib/fltlib.c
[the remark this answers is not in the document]
RatinCN: Committed

dll/win32/fltlib/fltlib.c
```diff
@@ -21,17 +21,24 @@
 
 /* DATA ****************************************************************************/
 
+#define NT_FACILITY(Status) ((Status >> 16) & 0xFFF)
```
learn-more, line 24: This looks a lot like `SCODE_FACILITY` (but not entirely, that uses `0x1FFF`) is that intentional?
Can't you use `SCODE_FACILITY` here?

dll/win32/fltlib/fltlib.c
```diff
@@ -1,10 +1,10 @@
 /*
-* PROJECT:         Filesystem Filter Manager library
-* LICENSE:         GPL - See COPYING in the top level directory
-* FILE:            dll/win32/fltlib/fltlib.c
-* PURPOSE:
-* PROGRAMMERS:     Ged Murphy (ged.murphy@reactos.org)
-*/
+ * PROJECT:     ReactOS system libraries
```
learn-more, line 2: The previous project name was more descriptive?

dll/win32/fltlib/fltlib.c
```diff
@@ -56,15 +63,53 @@ FilterUnload(_In_ LPCWSTR lpFilterName)
...
+        DosError = RtlNtStatusToDosError(Status);
+
+        /* If RtlNtStatusToDosError didn't translate the status, we return the return value as-is */
+        if ((LONG)(DosError) <= 0)
+        {
+            return DosError;
+        }
+
+        /* Convert Win32 error to HRESULT */
+        Result = HRESULT_FROM_WIN32(DosError);
```
learn-more, line 109: Why translate it via dos to NT, instead of using `HRESULT_FROM_NT` ?
Are there apitests that show it needs to do the dance with the DosError?

dll/win32/fltlib/fltlib.c
RatinCN: `SCODE_FACILITY` should not be used here, because:
- Status is not HResult, `SCODE_FACILITY` should be use to resolve HResult only.
- Facility code takes 12bits, a higher bit have special meaning, should not be ignored. See: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/87fba13e-bf06-450e-83b1-9241dc81e781

Windows SDK have no macro to get facility code of status, so we need to do something like Process Hacker here:
https://processhacker.sourceforge.io/doc/ntbasic_8h.html#ab922cd64630fe14a55627b1d987f644a

dll/win32/fltlib/fltlib.c
RatinCN: Previous project name has moved to "PURPOSE", any suggestion about "PROJECT" and "PURPOSE" here?

dll/win32/fltlib/fltlib.c
HBelusca: Note that the "SCODE" stuff is a status code used in MAPI interface only.

HBelusca, line 24:
```suggestion
#define NT_FACILITY(Status) ((((ULONG)(Status)) >> 16) & 0xFFF)
```

dll/win32/fltlib/fltlib.c
RatinCN: `HRESULT_FROM_NT(Status)` and `HRESULT_FROM_NT(RtlNtStatusToDosError(Status))` are totally different.

`HRESULT_FROM_NT` just wrap the Status in HResult format, no translation.
`FilterHResultFromNtStatus` returns HResult that contains Win32 error, not HResult that contains NTStatus.

And previous code did so.

HBelusca: > HRESULT_FROM_NT(RtlNtStatusToDosError(Status))

You meant `HRESULT_FROM_WIN32(RtlNtStatusToDosError(Status))`

dll/win32/fltlib/fltlib.c
HBelusca: In principle, "Project" is the overall description for (here) fltlib, while "Purpose" would be the description/purpose of the specific .c/.h/.rc etc. file.

dll/win32/fltlib/fltlib.c
RatinCN: You are right, I confused `SCODE` and`HRESULT`, previous my comment fixed

dll/win32/fltlib/fltlib.c
RatinCN: > > HRESULT_FROM_NT(RtlNtStatusToDosError(Status))
> 
> You meant `HRESULT_FROM_WIN32(RtlNtStatusToDosError(Status))`

You're right, fixed

dll/win32/fltlib/fltlib.c
RatinCN: Thanks, new header committed

dll/win32/fltlib/fltlib.c
RatinCN: Committed

dll/win32/fltlib/fltlib.c
```diff
@@ -109,48 +156,134 @@ FilterLoadUnload(_In_z_ LPCWSTR lpFilterName,
...
+            {
+                Status = IoStatusBlock.Status;
+            }
+        }
+
+        /* Return the number of bytes transferred if no error,
+         * but seems Windows didn't check this. */
+        if (!NT_ERROR(Status))
+        {
+            *lpBytesReturned = (DWORD)IoStatusBlock.Information;
```
ThFabba, line 282: I understand that whether `Information` is meaningful is based on `NT_ERROR`, but it still seems a bit weird in this function to set bytes returned in some but not all failure cases. Perhaps set it to 0 at the beginning or in the `else` block?

dll/win32/fltlib/fltlib.c
```diff
@@ -109,48 +156,134 @@ FilterLoadUnload(_In_z_ LPCWSTR lpFilterName,
...
+        }
+    }
+    else
     {
-        Status = NtWaitForSingleObject(Handle, FALSE, NULL);
-        if (NT_SUCCESS(Status))
+        /* Send file system control or device control according to the device type */
+        if (DeviceType == FILE_DEVICE_FILE_SYSTEM)
+        {
+            Status = NtFsControlFile(hFltMgr,
```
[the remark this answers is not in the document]
RatinCN: I tried this way before, and found we had to write `if (lpOverlapped)` again after FsControl/IoControl call, like:
```
if (lpOverlapped) {
    xxx;
} else {
    xxx;
}
FsControl/IoControl
if (lpOverlapped) {
    yyy;
} else {
    yyy;
}
```
seems incoherent, so I use the style like [DeviceIoControl](https://github.com/reactos/reactos/blob/master/dll/win32/kernel32/client/file/deviceio.c#L160)

sdk/include/psdk/ntstatus.h
```diff
@@ -40,6 +40,7 @@ extern "C" {
 #define FACILITY_CLUSTER_ERROR_CODE     0x13
 #define FACILITY_ACPI_ERROR_CODE        0x14
 #define FACILITY_SXS_ERROR_CODE         0x15
+#define FACILITY_FILTER_MANAGER         0x1C
```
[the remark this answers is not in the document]
RatinCN: Committed

dll/win32/fltlib/fltlib.c
RatinCN: Seems Windows neither set it to 0 at the beginning, nor check `if (!NT_ERROR(Status))`...
I guess so, `lpBytesReturned` output is valid only if this function success, so there is no big problem, but a bit weird.

dll/win32/fltlib/fltlib.c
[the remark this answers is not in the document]
RatinCN: Aha, finally I found the truth, in [GetQueuedCompletionStatus](https://learn.microsoft.com/en-us/windows/win32/api/ioapiset/nf-ioapiset-getqueuedcompletionstatus#parameters), `lpOverlapped` description said:
>Even if you have passed the function a file handle associated with a completion port and a valid [OVERLAPPED](https://learn.microsoft.com/en-us/windows/desktop/api/minwinbase/ns-minwinbase-overlapped) structure, an application can prevent completion port notification. This is done by specifying a valid event handle for the hEvent member of the OVERLAPPED structure, and setting its low-order bit. **A valid event handle whose low-order bit is set keeps I/O completion from being queued to the completion port.**

Comment updated!

dll/win32/fltlib/fltlib.c
RatinCN: Sorry for my misunderstanding, there are some details (may include your question) seems missed by me, more investigations and code fixes are in progress, this PR convert to draft temporarily, my fault

RatinCN:
Sorry, my fault. Current implement have a bit different from Windows, so more investigations and minor code fixes are needed and in progress, convert to draft temporarily

dll/win32/fltlib/fltlib.c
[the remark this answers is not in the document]
HBelusca: Ah, thanks for the info!

dll/win32/fltlib/fltlib.c
RatinCN: I set it to 0 at the beginning and removed `if (!NT_ERROR(Status))`.
`IoStatusBlock.Information` is request-depended and may meaningful when failed, like `STATUS_BUFFER_OVERFLOW` (Warning), `STATUS_FLT_BUFFER_TOO_SMALL` (Warning), `STATUS_BUFFER_TOO_SMALL` (Error), ...

RatinCN:
- Add missing SEH
- `LPOVERLAPPED ApcContext` to `PVOID ApcContext`
- Set `IoStatusBlock.Information` to 0 at the beginning, and always return it to `lpBytesReturned`

Okay to go now. @ThFabba

PR #5101, opened 2023-02-25
[NOTEPAD] Shift+F3 for backward search

katahiromz:
## Purpose

JIRA issue: [CORE-17064](https://jira.reactos.org/browse/CORE-17064)

## Proposed changes

- Add `Shift+F3` accelerator as command `CMD_SEARCH_PREV`.
- Extend `DIALOG_SearchNext` function with `bDown` argument for the action of command `CMD_SEARCH_PREV`.
- Modify the message loop for effective `F3` key on Find/Replace dialog.

## TODO

- [x] Do tests.


base/applications/notepad/dialog.c
```diff
@@ -1072,6 +1072,16 @@ VOID DIALOG_Search(VOID)
...
 VOID DIALOG_SearchNext(VOID)
 {
+    Globals.find.Flags |= FR_DOWN;
+    if (Globals.find.lpstrFindWhat != NULL)
+        NOTEPAD_FindNext(&Globals.find, FALSE, TRUE);
+    else
+        DIALOG_Search();
+}
+
+VOID DIALOG_SearchPrev(VOID)
```
HBelusca, line 1082: Both functions could be merged together, with a BOOL parameter that just specifies the direction of the search.

katahiromz: Fixed.

base/applications/notepad/main.c
```diff
@@ -649,6 +651,14 @@ int WINAPI _tWinMain(HINSTANCE hInstance, HINSTANCE prev, LPTSTR cmdline, int sh
 
     while (GetMessage(&msg, 0, 0, 0))
     {
+        /* Make F3 key effective on Find/Replace dialog */
```
HBelusca, line 654: Why is this hack needed? Shouldn't this be handled correctly by the IsDialogMessage+TranslateAccelerator + DispatchMessage?

katahiromz: The target window is different from the main window, so the `CMD_SEARCH_NEXT` command won't reach the main window.

HBelusca: In standard notepad, (and notepad++ etc.), pressing F3 while the find dialog is focused, does not work (one needs to press e.g. ENTER on the "Search next" button).

katahiromz: Fixed the calling order of `TranslateAccelerator` and `IsDialogMessage`.

HBelusca (approved):

PR #5103, opened 2023-02-26
[NOTEPAD] Fix NOTEPAD_FindTextAt

katahiromz:
## Purpose
The whole-word search of Notepad had a bug around punctuation. For example, the text `"Windows"` didn't match in the text `"MS-DOS,Windows,ReactOS"`.

## Proposed changes

- Use `_istalnum` instead of `_istspace`.
- Fix the position to check.

`_istalnum` matches an alphabet letter or numeric digit.

## TODO

- [x] Do tests.


HBelusca:
> Use _istalnum instead of _istspace.

Well yes, but what is it actually changing? Please describe the symptom(s) of the bug in the commit log.

PR #5104, opened 2023-02-26
[NOTEPAD] Don't reset Find/Replace if lpstrFindWhat is set

katahiromz:
## Purpose

We don't want to reset the status of the checkboxes even if the Find/Replace dialog was closed.

## Proposed changes

- If `lpstrFindWhat` is set, do not reset the Find/Replace settings at `DIALOG_SearchDialog`.

## TODO

- [x] Do tests.


learn-more:
> If set, do not reset the Find/Replace settings at DIALOG_SearchDialog.

I can see that in the code, what I'm not seeing in the code nor in your PR description is _why_.

katahiromz:
> why

The checkboxes will also be reset if the structure was reset.

PR #5109, opened 2023-03-01
[MSPAINT] Treat as a file even if the bitmap is empty

atharva1910:
JIRA issue: [CORE-18508](https://jira.reactos.org/browse/CORE-18508)

Currently if a file is empty, while loading the file the global variable isAFile is set as FALSE.
This leads to a save as dialogue while saving thinking that no file is loaded

### DIFF
--- a/base/applications/mspaint/dib.cpp
+++ b/base/applications/mspaint/dib.cpp
@@ -118,7 +118,6 @@ HBITMAP SetBitmapAndInfo(HBITMAP hBitmap, LPCTSTR name, DWORD dwFileSize, BOOL i
 
         fileHPPM = fileVPPM = 2834;
         ZeroMemory(&fileTime, sizeof(fileTime));
-        isFile = FALSE;
     }
     else
     {


katahiromz (approved):

katahiromz (changes requested):
Can you add a filename check please?

katahiromz:
You can use `PathFileExists` function.

learn-more:
> Can you add a filename check please?

Can you please elaborate what you expect the author to do when a file is found or not found?

atharva1910:
> Can you add a filename check please?

That information is already passed to the function
HBITMAP SetBitmapAndInfo(HBITMAP hBitmap, LPCTSTR name, DWORD dwFileSize, BOOL isFile)
The last paramter indicates the same

katahiromz:
```
if (!PathFileExists(name))
    isFile = FALSE;
```

roybaer:
The two-liner suggested by @katahiromz does indeed look like a somewhat better solution, because it only catches the exact problem described.
Off the top of my head I would say that simply omitting `isFile = FALSE;` could potentially cause problems with newly created pictures and the list of recently used files.

atharva1910:
The application already checks if a file is being loaded , opens it, checks if the HANDLE is valid and then passes a true value to the function in the last parameter. Why do we wish to do that again. It seems pointless to add the check again

> Off the top of my head I would say that simply omitting isFile = FALSE; could potentially cause problems with newly created pictures and the list of recently used files.
I have ran multiple scenarios using the fix and it works fine in the new file case as well as opening files

katahiromz (approved):
I did test actually, it is successful. Approved.

PR #5112, opened 2023-03-02
[NTOS:SE] Mute Access check output

learn-more:
This mutes the access check debug output that was recently added.

### DIFF
--- a/ntoskrnl/se/debug.c
+++ b/ntoskrnl/se/debug.c
@@ -69,7 +69,7 @@ SepDumpAceFlags(
 #define ACE_FLAG_PRINT(x)  \
     if (AceFlags & x)      \
     {                      \
-        DbgPrint(#x "\n"); \
+        DPRINT(#x "\n"); \
     }
 
     ACE_FLAG_PRINT(OBJECT_INHERIT_ACE);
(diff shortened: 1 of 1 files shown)


ntoskrnl/se/debug.c
```diff
@@ -107,24 +107,23 @@ SepDumpAces(
...
              * debugging of ACEs.
              */
-            DbgPrint("SepDumpAces(): Failed to find the next ACE, stop dumping info...\n");
+            DPRINT("SepDumpAces(): Failed to find the next ACE, stop dumping info...\n");
             return;
         }
 
-        DbgPrint("================== %lu# ACE DUMP INFO ==================\n", AceIndex);
-        DbgPrint("Ace -> 0x%p\n", Ace);
-        DbgPrint("Ace->Header -> 0x%p\n", Ace->Header);
```
learn-more, line 116: @GeoB99 I removed this line, because you were printing an object as a pointer.
You probably wanted to print the address of Header, but that's the same as `Ace` since it's the first item.

GeoB99 (approved):
Ideally the Security subsystem would need a debugging system similarly seen to other components like the Object Manager with its `OBTRACE` and such in the future. This is kind of a low priority for me now though.

I'm OK for the change by the way.

ntoskrnl/se/debug.c
```diff
@@ -69,7 +69,7 @@ SepDumpAceFlags(
 #define ACE_FLAG_PRINT(x)  \
     if (AceFlags & x)      \
     {                      \
-        DbgPrint(#x "\n"); \
+        DPRINT(#x "\n"); \
```
HBelusca, line 72: Using DPRINT(1) would make the dumps (when debugging is enabled at compile-time) rather long with the useless file-name+line numbers. Wouldn't it be preferrable to have something like:
```c
#ifdef NDEBUG
#define ACE_FLAG_PRINT(x)
#else
#define ACE_FLAG_PRINT(x) { /* its actual definition */ }
#endif /* NDEBUG */
```
and also have those SepDump*** functions defined only when `NDEBUG` is not defined? (and otherwise defined to nothing if `NDEBUG` is defined?)
This may help also reducing the size of the PR diff.

learn-more: My aim was not to create the best debugging system ever,
but if you want to implement something like this feel free to do so.

GeoB99:
I disabled this spammy debug output by having it enabled only if `NDEBUG` is not defined on that file. Before anyone of you asks me why did I do like this -- because it's the quickest way.

I will come up with a debugging system for Se when I have time for this, for now we will live with it.

PR #5115, opened 2023-03-04
[KERNEL32] GlobalMemoryStatusEx should handle error and return FALSE when failed

RatinCN:
## Purpose

`GlobalMemoryStatusEx` should return FALSE and set Last Error when failed

JIRA issue: None

## Proposed changes

`GlobalMemoryStatusEx` return FALSE and set Last Error when failed.
Before this PR, `GlobalMemoryStatusEx` always return TRUE and didn't inspect result of calling `Nt*`.

### DIFF
--- a/dll/win32/kernel32/client/heapmem.c
+++ b/dll/win32/kernel32/client/heapmem.c
@@ -1275,6 +1275,7 @@ GlobalMemoryStatusEx(LPMEMORYSTATUSEX lpBuffer)
     VM_COUNTERS VmCounters;
     QUOTA_LIMITS QuotaLimits;
     ULONGLONG PageFile, PhysicalMemory;
+    NTSTATUS Status;
 
     if (lpBuffer->dwLength != sizeof(*lpBuffer))
     {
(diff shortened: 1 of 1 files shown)


GeoB99 (approved):

learn-more:
Is there an application that has problems due to this?
Does this fix an apitest?

RatinCN:
> Is there an application that has problems due to this? Does this fix an apitest?

No, this PR only add error-handling that should have done, Windows also check them.

learn-more:
> > Is there an application that has problems due to this? Does this fix an apitest?
> 
> No, this PR only add error-handling that should have done, Windows also check them.

Then how did you come to the conclusion that this change is needed?

RatinCN:
> > > Is there an application that has problems due to this? Does this fix an apitest?
> > 
> > 
> > No, this PR only add error-handling that should have done, Windows also check them.
> 
> Then how did you come to the conclusion that this change is needed?

Many legal ways other than existing app/apitest to know or infer this:

- MSDoc shows those `Nt*` are not always success and `GlobalMemoryStatusEx` may return FALSE
- Most of WinAPI obey the error-handling practice like this, however, `GlobalMemoryStatusEx` always return `TRUE` even if `Nt*` failed is odd and incorrect obviously
- Write an app call `GlobalMemoryStatusEx` after hooking `Nt*` to make them failed and observe return value in Windows
- Respect clean room reverse engineering practices
- ...

PR #5117, opened 2023-03-04
[KERNEL32] Remove a pointless SEH in VirtualAllocEx

RatinCN:
## Purpose

This SEH wraps a syscall only, seems pointless and should be removed

JIRA issue: None

## Proposed changes

Remove this SEH

### DIFF
--- a/dll/win32/kernel32/client/virtmem.c
+++ b/dll/win32/kernel32/client/virtmem.c
@@ -37,22 +37,13 @@ VirtualAllocEx(IN HANDLE hProcess,
         return NULL;
     }
 
-    /* Handle any possible exceptions */
-    _SEH2_TRY
-    {
-        /* Allocate the memory */
-        Status = NtAllocateVirtualMemory(hProcess,
-                                         &lpAddress,
-                                         0,
-                                         &dwSize,
-                                         flAllocationType,
-                                         flProtect);
-    }
-    _SEH2_EXCEPT(EXCEPTION_EXECUTE_HANDLER)
-    {
-        Status = _SEH2_GetExceptionCode();
-    }
-    _SEH2_END;
+    /* Allocate the memory */
+    Status = NtAllocateVirtualMemory(hProcess,
+                                     &lpAddress,
+                                     0,
+                                     &dwSize,
+                                     flAllocationType,
+                                     flProtect);
 
     /* Check for status */
     if (!NT_SUCCESS(Status))


SergeGautherie:
> SEH [...] pointless

SEH was added as part of 6d97180 (r52899).
Was it a superfluous addition or did something change in the meantime?

JoachimHenze (approved):
I cannot really judge whether it is pointless. But I think we should give the PR a chance to find it out.
The commit linked by Serge also didn't answer this question to me. Thx still for mentioning it.
Pr is very local and therefore easily revertable, in case it would introduce any problems.

RatinCN:
- **Win2K3 has SEH frame here** but was removed since Win7 (and later), seems is an improvement
- No more info I found in [6d97180](https://github.com/reactos/reactos/commit/6d971807f700945345b150167b62294b30ae83e2), just explicitly added SEH (because SEH is present in 2K3?)
Win2K3 is currently target, so I'm not sure this PR shoud be merged or closed now, maybe we need discuss more?
@JoachimHenze @SergeGautherie

JoachimHenze:
I doubt that any participant in the discussion has enough information to actually answer the guesses and questions here.
We should run the bots on it, and if they are neutral, I think we should just merge it as an experiment, and revert it if the experiment would prove it to break anything. I don't think we should close it.

RatinCN:
> I doubt that any participant in the discussion has enough information to actually answer the guesses and questions here. We should run the bots on it, and if they are neutral, I think we should just merge it as an experiment, and revert it if the experiment would prove it to break anything. I don't think we should close it.

Agree, if this SEH was optimized out in Win7 as an improvement and safe, we can take it

HBelusca:
> * **Win2K3 has SEH frame here** but was removed since Win7 (and later), seems is an improvement
> 
>     * No more info I found in [6d97180](https://github.com/reactos/reactos/commit/6d971807f700945345b150167b62294b30ae83e2), just explicitly added SEH (because SEH is present in 2K3?)

I'm going to run this PR on the testbots in order to see whether the existing VirtualAllocEx winetests don't break.

HBelusca:
Test results:
KVM: https://reactos.org/testman/compare.php?ids=86171,86174
VBox: https://reactos.org/testman/compare.php?ids=86172,86173

I will re-run them again to be sure.

SergeGautherie:
> * **Win2K3 has SEH frame here** but was removed since Win7 (and later), seems is an improvement

If in doubt (as in: does that removal "depended" on other Win7 changes?), I would suggest to only add some "#if NT <= 6.0" instead.
If 100% sure no exceptions can be triggered, then I would suggest to also add an explicit reminder code comment about this change.

ThFabba (approved):
The only way SEH here should make a difference here is if the stack pointer is close to an inaccessible region. You _could_ write a test for that but I think it's pointless. Yes, Win2003 has SEH here but I think that's some kind of leftover or bug. There's no reason to have it and no observable effect for a normal application -- as evidenced by the fact that Vista removed it.

HBelusca (approved):
Maybe add "Reverts commit xxxxx" (with the one Serve mentioned) in the commit log.

RatinCN:
Both of code comment and commit comment added.

RatinCN:
> The only way SEH here should make a difference here is if the stack pointer is close to an inaccessible region.
If there are something wrong with stack pointer, an exception will occur sooner or later or when return from `VirtualAllocEx`, seems nothing benefit from this SEH in this case.

dll/win32/kernel32/client/virtmem.c
```diff
@@ -37,22 +37,17 @@ VirtualAllocEx(IN HANDLE hProcess,
...
-                                         flProtect);
-    }
-    _SEH2_EXCEPT(EXCEPTION_EXECUTE_HANDLER)
-    {
-        Status = _SEH2_GetExceptionCode();
-    }
-    _SEH2_END;
+    /* Allocate the memory */
+
+    /* Win2K3 has a SEH just wrap this syscall only,
```
HBelusca, line 42: This should maybe just be part of the commit log.

JoachimHenze: Yes, I would also omit the comment entirely, and just mention that in the commit comment.

RatinCN: Commit log updated

dll/win32/kernel32/client/virtmem.c
```diff
@@ -37,22 +37,17 @@ VirtualAllocEx(IN HANDLE hProcess,
...
+       which seems pointless and was removed since NT6.0,
+       we remove it too as an experiment. */
```
HBelusca, line 44:
```suggestion
```

RatinCN: Committed

PR #5119, opened 2023-03-05
[REGEDIT] Avoid buffer overflow in SelectNode

ThFabba:
JIRA issue: [CORE-18602](https://jira.reactos.org/browse/CORE-18602)

### DIFF
--- a/base/applications/regedit/treeview.c
+++ b/base/applications/regedit/treeview.c
@@ -781,7 +781,7 @@ BOOL SelectNode(HWND hwndTV, LPCWSTR keyPath)
 
     /* Load "My Computer" string... */
     LoadStringW(hInst, IDS_MY_COMPUTER, szBuffer, ARRAY_SIZE(szBuffer));
-    wcscat(szBuffer, L"\\");
+    StringCbCatW(szBuffer, sizeof(szBuffer), L"\\");
 
     /* ... and remove it from the key path */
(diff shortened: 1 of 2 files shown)


base/applications/regedit/treeview.c
```diff
@@ -795,24 +795,33 @@ BOOL SelectNode(HWND hwndTV, LPCWSTR keyPath)
...
         s = wcschr(keyPath, L'\\');
-        lstrcpynW(szPathPart, keyPath, s ? s - keyPath + 1 : wcslen(keyPath) + 1);
+        if (s != NULL)
+        {
+            copyLength = (s - keyPath) * sizeof(WCHAR);
+        }
+        else
+        {
+            copyLength = sizeof(szPathPart);
+        }
```
JoachimHenze, line 807: I am no fan of those 4 superfluous curly braces. Both: the command and the condition do fit into a single line and therefore they do not contribute to improved readability.. We are just increasing LOC-count here and reduce amount of code that can be viewed at the same time without scrolling.

JoachimHenze: Maybe the best would be even to avoid the additional variable and just write it with the "?" operator again. That would save 9 lines of code. And would even "limit the scope" of the variable to the max by eliminating it.

ThFabba: [Our coding style](https://reactos.org/wiki/Coding_Style#Braces) expressly permits braces around all blocks.

PR #5123, opened 2023-03-06
[NTOS:SE] Do not allocate memory pool just for the access rights

GeoB99:
Access check is an expensive operation, that is, whenever an access to an object is performed an access check has to be done to ensure the access can be allowed to the calling thread who attempts to access such object.

Currently `SepAnalyzeAcesFromDacl` allocates a block of pool memory for access check rights, nagging the Memory Manager like a desperate naughty creep. So instead initialize the access rights as a simple variable in `SepAccessCheck` and pass it out as an address to `SepAnalyzeAcesFromDacl `so that the function will fill it up with access rights. This helps with performance, avoiding wasting a few bits of memory just to hold these access rights.

In addition to that, add a few asserts and fix the copyright header on both `se.h` and `accesschk.c,` to reflect the Coding Style rules.

### DIFF
--- a/ntoskrnl/include/internal/se.h
+++ b/ntoskrnl/include/internal/se.h
@@ -1,9 +1,9 @@
 /*
- * PROJECT:         ReactOS Kernel
- * LICENSE:         GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
- * PURPOSE:         Internal header for the Security Manager
- * COPYRIGHT:       Copyright Eric Kohl
- *                  Copyright 2022 George Bișoc <george.bisoc@reactos.org>
+ * PROJECT:     ReactOS Kernel
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Internal header for the Security Manager
+ * COPYRIGHT:   Copyright Eric Kohl
+ *              Copyright 2022 George Bișoc <george.bisoc@reactos.org>
  */
 
 #pragma once
@@ -303,7 +303,7 @@ SepDumpTokenDebugInfo(
 
 VOID
 SepDumpAccessRightsStats(
-    _In_opt_ PACCESS_CHECK_RIGHTS AccessRights);
+    _In_ PACCESS_CHECK_RIGHTS AccessRights);
 #endif // DBG
 
 //
--- a/ntoskrnl/se/accesschk.c
+++ b/ntoskrnl/se/accesschk.c
@@ -1,10 +1,10 @@
 /*
- * PROJECT:         ReactOS Kernel
- * LICENSE:         GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
- * PURPOSE:         Security access check control implementation
- * COPYRIGHT:       Copyright 2014 Timo Kreuzer <timo.kreuzer@reactos.org>
- *                  Copyright 2014 Eric Kohl
- *                  Copyright 2022 George Bișoc <george.bisoc@reactos.org>
+ * PROJECT:     ReactOS Kernel
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Security access check control implementation
+ * COPYRIGHT:   Copyright 2014 Timo Kreuzer <timo.kreuzer@reactos.org>
+ *              Copyright 2014 Eric Kohl
+ *              Copyright 2022-2023 George Bișoc <george.bisoc@reactos.org>
  */
 
 /* INCLUDES *******************************************************************/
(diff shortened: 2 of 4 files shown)


ntoskrnl/se/accesschk.c
```diff
@@ -191,22 +131,14 @@ SepAnalyzeAcesFromDacl(
...
-     * restricted SIDs so we have allocated this
-     * pointer before.
-     */
-    if (!AccessRightsAllocated)
+    /* Initialize the access rights if we have not done it before */
+    if (!AccessRightsArrayInitialized)
     {
-        AccessRights = SepInitAccessCheckRights();
-        if (!AccessRights)
+        for (AccessRightsIndex = 0;
```
learn-more, line 137: RtlZeroMemory would be easier here?

ntoskrnl/include/internal/se.h
```diff
@@ -71,6 +61,14 @@ typedef struct _TOKEN_AUDIT_POLICY_INFORMATION
 #define TOKEN_DUPLICATE_METHOD 0xDUL
 #define TOKEN_FILTER_METHOD    0xFUL
 
+//
+// Access Rights Types Array
+//
+#define ACCESS_RIGHTS_REMAINING_RIGHTS_ARRAY  0
+#define ACCESS_RIGHTS_GRANTED_RIGHTS_ARRAY    1
+#define ACCESS_RIGHTS_DENIED_RIGHTS_ARRAY     2
+#define ACCESS_RIGHTS_MAX_ARRAY               3
```
learn-more, line 70: I can understand why you would not want an allocation, but what was wrong with the struct?
Imo this is less sensitive to errors than an array?

ntoskrnl/se/accesschk.c
```diff
@@ -486,7 +424,8 @@ SepAccessCheck(
     BOOLEAN Defaulted;
     NTSTATUS Status;
     PACCESS_TOKEN Token = NULL;
-    PACCESS_CHECK_RIGHTS AccessCheckRights = NULL;
+    ACCESS_MASK AccessRightsArray[ACCESS_RIGHTS_MAX_ARRAY];
```
learn-more, line 427: If you just initialize your data here, you don't need to pass the whole 'initialize' thingy to the function.

ntoskrnl/include/internal/se.h
```diff
@@ -71,6 +61,14 @@ typedef struct _TOKEN_AUDIT_POLICY_INFORMATION
 #define TOKEN_DUPLICATE_METHOD 0xDUL
 #define TOKEN_FILTER_METHOD    0xFUL
 
+//
+// Access Rights Types Array
+//
+#define ACCESS_RIGHTS_REMAINING_RIGHTS_ARRAY  0
```
HBelusca, line 67: Names are quite long, maybe use "ACCESS_RIGHTS_REMAINING" etc. (no need to say "RIGHTS_ARRAY"

ntoskrnl/se/accesschk.c
HBelusca, line 427: Observation: This array contains only 3 (== ACCESS_RIGHTS_MAX_ARRAY) ACCESS_MASK elements, not more, not less.
So, you could have kept your original `ACCESS_CHECK_RIGHTS` structure, and here, declare a variable of type `ACCESS_CHECK_RIGHTS` (instead of that array, or a pointer as it was before). This would allow to get rid of the ugly indices define names from above too, and to keep the original code almost unchanged.
The only thing that would be removed is, the pool allocation + the pool free at the end.

HBelusca (changes requested):
Code can be made a bit more optimal.

ntoskrnl/se/accesschk.c
GeoB99: Ok, it turns out ditching the structure that I introduced is a bad idea. I'm going to refactor this as per your observation.

ntoskrnl/se/debug.c
```diff
@@ -325,8 +325,13 @@ VOID
 SepDumpAccessRightsStats(
     _In_opt_ PACCESS_CHECK_RIGHTS AccessRights)
```
HBelusca, line 326: Now, this should become `_In_`  only (+ in the .h file).

HBelusca (approved):

ntoskrnl/se/accesschk.c
```diff
@@ -486,7 +394,7 @@ SepAccessCheck(
     BOOLEAN Defaulted;
     NTSTATUS Status;
     PACCESS_TOKEN Token = NULL;
-    PACCESS_CHECK_RIGHTS AccessCheckRights = NULL;
+    ACCESS_CHECK_RIGHTS AccessCheckRights = {0};
```
learn-more, line 397: In this case it's probably not going to go towards usermode, but in case it ever is:
This can leak stack data if there is padding in the struct, because the padding is not zero'd out.

PR #5129, opened 2023-03-08
[NOTEPAD] Arabic also uses Right-to-Left (RTL) layout

katahiromz:
## Purpose
The Arabic user uses Right-to-Left (RTL) layout. Notepad should support RTL.

## Proposed changes

- Use `PRIMARYLANGID` macro to get the primary language ID.
- Add Arabic case.

### DIFF
--- a/base/applications/notepad/main.c
+++ b/base/applications/notepad/main.c
@@ -573,6 +573,7 @@ int WINAPI _tWinMain(HINSTANCE hInstance, HINSTANCE prev, LPTSTR cmdline, int sh
     static const TCHAR className[] = _T("Notepad");
     static const TCHAR winName[] = _T("Notepad");
 
+    /* Arabic and Hebrew are Right-to-Left */
     switch (PRIMARYLANGID(GetUserDefaultUILanguage()))
     {
         case LANG_ARABIC:
--- a/base/applications/notepad/main.c
+++ b/base/applications/notepad/main.c
@@ -573,14 +573,12 @@ int WINAPI _tWinMain(HINSTANCE hInstance, HINSTANCE prev, LPTSTR cmdline, int sh
     static const TCHAR className[] = _T("Notepad");
     static const TCHAR winName[] = _T("Notepad");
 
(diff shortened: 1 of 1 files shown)


base/applications/notepad/main.c
```diff
@@ -573,14 +573,13 @@ int WINAPI _tWinMain(HINSTANCE hInstance, HINSTANCE prev, LPTSTR cmdline, int sh
...
-    case MAKELANGID(LANG_HEBREW, SUBLANG_DEFAULT):
-        SetProcessDefaultLayout(LAYOUT_RTL);
-        break;
-
-    default:
-        break;
+        case LANG_ARABIC:
+        case LANG_HEBREW:
+            SetProcessDefaultLayout(LAYOUT_RTL);
+            break;
```
learn-more, line 582: Why is this required, this seems like something that should be done by the OS?
Does it need this when running our Notepad on an RTL Windows machine?

HBelusca: It's automatically done by the OS **only if** the file description string version resource contains two left-to-right marks, as done e.g. in https://git.reactos.org/?p=reactos.git;a=blob;f=base/setup/welcome/lang/he-IL.rc;hb=5bc045c041f0c90d54f009c622d085422c80b662
See also https://devblogs.microsoft.com/oldnewthing/20100611-00/?p=13743

katahiromz: https://github.com/reactos/reactos/blob/2f43d9c18338722999f18630dabd77821a1d79cb/base/applications/notepad/rsrc.rc#L26

katahiromz: https://github.com/reactos/reactos/blob/35a816a24e3b7b5752a287686c9df833af6d07a9/sdk/include/reactos/version.rc#L121

HBelusca:
Actually there is no point in having this PR, because literally **no** ReactOS application is translated in arabic so far, except for OSK.

katahiromz:
Superceded by #5138. Closed.

PR #5132, opened 2023-03-08
[NOTEPAD] Use Globals.hMainWnd in ShowLastError

katahiromz:
## Purpose
Improving UI/UX by making `Globals.hMainWnd` the owner of the message box.

## Proposed changes

- Use `Globals.hMainWnd` variable in `ShowLastError` function.

## Comparison
BEFORE: Try to open a folder by dragging & dropping.
image
Since the owner of the messsage box was `NULL`, the message box could be overlapped by the main window.

### DIFF
--- a/base/applications/notepad/dialog.c
+++ b/base/applications/notepad/dialog.c
@@ -75,7 +75,7 @@ VOID ShowLastError(VOID)
                       0,
                       NULL);
 
-        MessageBox(NULL, lpMsgBuf, szTitle, MB_OK | MB_ICONERROR);
+        MessageBox(Globals.hMainWnd, lpMsgBuf, szTitle, MB_OK | MB_ICONERROR);
         LocalFree(lpMsgBuf);
     }
 }


katahiromz:
Wrong implementation. Closed.

JoachimHenze (changes requested):
No this change is not correct.
We do have code in main.c of the kind:

    if (!Globals.hMainWnd)
    {
        ShowLastError();
        ExitProcess(1);
    }

katahiromz:
We can specify NULL to the 1st parameter of `MessageBox`. Nothing wrong.

learn-more:
> No this change is not correct. We do have code in main.c of the kind:
> 
> ```
> if (!Globals.hMainWnd)
> {
>     ShowLastError();
>     ExitProcess(1);
> }
> ```

So you are saying that `MessageBox(NULL, ...` is fine,
but 
```
Globals.hMainWnd = NULL;
MessageBox(Globals.hMainWnd, ...
```
is not fine?

HBelusca (approved):

PR #5135, opened 2023-03-08
[NOTEPAD] Refactor around _tWinMain

katahiromz:
## Purpose
Refactoring to improve prospects.

## Proposed changes

- Add `hInstance` parameter to `NOTEPAD_InitData`.
- Move some code in `_tWinMain` into `NOTEPAD_InitData`.
- Move some code in `_tWinMain` into `WM_CREATE` handling.
- Move some code in `WM_CLOSE` handling into `WM_DESTROY` handling.
- Fix the exit code of `_tWinMain`.

## TODO

- [x] Do tests.

### DIFF
--- a/base/applications/notepad/main.c
+++ b/base/applications/notepad/main.c
@@ -614,7 +614,6 @@ int WINAPI _tWinMain(HINSTANCE hInstance, HINSTANCE prev, LPTSTR cmdline, int sh
                                         GetSystemMetrics(SM_CXSMICON),
                                         GetSystemMetrics(SM_CYSMICON),
                                         0);
-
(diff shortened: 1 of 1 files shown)


JoachimHenze:
You explained what you did here. But you didn't explain yet why the new state is better than the old one.

katahiromz:
At first, the `WM_DESTROYCLIPBOARD` handling is needless in Notepad.

Second, the message handler has its specific role. We should conceal the detailed implementation from outside.

katahiromz:
It sounds like `"Flat Model vs. Encapsulated Model"`.
There're some room to debate about it.

PR #5136, opened 2023-03-08
[NTOS:CM] Set alternative system architecture field on PC-98

disean:
## Purpose

This is needed by some 3rd party drivers.

## Proposed changes

- Initialize the alternative system architecture field

### DIFF
--- a/ntoskrnl/config/cmconfig.c
+++ b/ntoskrnl/config/cmconfig.c
@@ -333,6 +333,11 @@ CmpInitializeHardwareConfiguration(IN PLOADER_PARAMETER_BLOCK LoaderBlock)
     ULONG Disposition;
     UNICODE_STRING KeyName;
 
+    /* Set the alternative system architecture information */
+#if defined(SARCH_PC98)
+    SharedUserData->AlternativeArchitecture = NEC98x86;
+#endif
+
     /* Setup the key name */
     RtlInitUnicodeString(&KeyName,
                          L"\\Registry\\Machine\\Hardware\\DeviceMap");


ntoskrnl/config/i386/cmhardwr.c
```diff
@@ -251,6 +252,11 @@ CmpInitializeMachineDependentConfiguration(IN PLOADER_PARAMETER_BLOCK LoaderBloc
     extern UNICODE_STRING KeRosProcessorName, KeRosBiosDate, KeRosBiosVersion;
     extern UNICODE_STRING KeRosVideoBiosDate, KeRosVideoBiosVersion;
 
+    /* Set the alternative system architecture information */
```
HBelusca, line 255: Technically this is done in CmpInitializeHardwareConfiguration, in case the configuration tree contains the "NEC PC-98" identifier (set by the NT loader, see https://git.reactos.org/?p=reactos.git;a=blob;f=boot/freeldr/freeldr/arch/i386/pc98/pc98hw.c;hb=8e0c920fd2649b1127abe37b88f5753647df5d7d#l1200 ).

disean: Moved into the CmpInitializeHardwareConfiguration function. An identifier check would be pointless since all of this information is already known to the kernel

julenuri (changes requested):

ntoskrnl/config/cmconfig.c
```diff
@@ -333,6 +333,11 @@ CmpInitializeHardwareConfiguration(IN PLOADER_PARAMETER_BLOCK LoaderBlock)
     ULONG Disposition;
     UNICODE_STRING KeyName;
 
+    /* Set the alternative system architecture information */
+#if defined(SARCH_PC98)
+    SharedUserData->AlternativeArchitecture = NEC98x86;
```
julenuri, line 338: A user suggest to replace 
"SharedUserData->AlternativeArchitecture = NEC98x86;" 
with a macro 
"SetNEC_98;"

julenuri (approved):

ntoskrnl/config/cmconfig.c
julenuri, line 338: A experienced user suggest under the debate field to replace the thing with the macro
```suggestion
    SetNEC_98;
```

julenuri (changes requested):

HBelusca (approved):

ntoskrnl/config/cmconfig.c
julenuri: Ok :)

PR #5138, opened 2023-03-08
[APPLICATIONS] Enable mirroring for right-to-left languages from resources.

HBelusca:
Currently done only for Hebrew, since ReactOS doesn't have the other ones (e.g. Arabic, Persian, etc.)

Supersedes PR #5129

### DIFF
--- a/base/applications/charmap/lang/he-IL.rc
+++ b/base/applications/charmap/lang/he-IL.rc
@@ -1,6 +1,14 @@
 /* Translated by Baruch Rutman */
 LANGUAGE LANG_HEBREW, SUBLANG_DEFAULT
 
+/*
+ * Enable RTL mirroring by adding two left-to-right marks (LRMs)
+ * in front of the FileDescription value of the version stamping.
+ */
+#undef REACTOS_STR_FILE_DESCRIPTION
+#define REACTOS_STR_FILE_DESCRIPTION    L"\x200e\x200e" L"ReactOS Character Map"
+#include <reactos/version.rc>
+
 IDD_CHARMAP DIALOGEX 6, 6, 292, 224
 FONT 8, "MS Shell Dlg", 0, 0
 STYLE DS_SHELLFONT | WS_CHILD | WS_VISIBLE
--- a/base/applications/charmap/charmap.c
+++ b/base/applications/charmap/charmap.c
@@ -680,17 +680,6 @@ wWinMain(HINSTANCE hInst,
 
     hInstance = hInst;
 
-    /* Mirroring code for the titlebar */
-    switch (GetUserDefaultUILanguage())
-    {
-        case MAKELANGID(LANG_HEBREW, SUBLANG_DEFAULT):
-            SetProcessDefaultLayout(LAYOUT_RTL);
-            break;
-
-        default:
-            break;
-    }
-
     iccx.dwSize = sizeof(INITCOMMONCONTROLSEX);
     iccx.dwICC = ICC_TAB_CLASSES;
     InitCommonControlsEx(&iccx);
--- a/base/applications/clipbrd/clipbrd.c
+++ b/base/applications/clipbrd/clipbrd.c
@@ -675,16 +675,6 @@ int WINAPI wWinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPWSTR lpCmdLi
         return 0;
     }
 
-    switch (GetUserDefaultUILanguage())
-    {
-        case MAKELANGID(LANG_HEBREW, SUBLANG_DEFAULT):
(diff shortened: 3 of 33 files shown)


HBelusca:
Cc @peterooch @reactos/lang-hebrew

base/applications/charmap/lang/he-IL.rc
```diff
@@ -1,6 +1,14 @@
 /* Translated by Baruch Rutman */
 LANGUAGE LANG_HEBREW, SUBLANG_DEFAULT
 
+/*
+ * Enable RTL mirroring by adding two left-to-right marks (LRMs)
+ * in front of the FileDescription value of the version stamping.
+ */
+#undef REACTOS_STR_FILE_DESCRIPTION
```
katahiromz, line 8: I think undefining `REACTOS_STR_FILE_DESCRIPTION` is an improper manner.
We can change `<reactos/version.rc>`.

HBelusca: And you would change it to what?

katahiromz:
```
VALUE "FileDescription",  L"\x200e\x200e"  REACTOS_STR_FILE_DESCRIPTION
```

HBelusca: But version.rc is the template file also used for non RTL languages. That's why you don't want to do that change always.

One way around, could be to keep the `REACTOS_STR_FILE_DESCRIPTION` define, and for RTL languages (in their resource file), define a `REACTOS_STR_FILE_DESCRIPTION_RTL`, or maybe some other macro name (e.g. `REACTOS_RTL_RESOURCE`), and in the version.rc template, do something along the lines of:
```
#ifdef REACTOS_RTL_RESOURCE
VALUE "FileDescription",  L"\x200e\x200e"  REACTOS_STR_FILE_DESCRIPTION
#undef REACTOS_RTL_RESOURCE
#else
VALUE "FileDescription",  REACTOS_STR_FILE_DESCRIPTION
#endif
```
Also, note that having unicode L string tag added in front of the description string would be mandatory. With the current (or yours) proposition, it is not, and that would break compilation with some compilers.

katahiromz: > In any case, one would have to re-include version.rc in the language resource file...

Why? Reason?

katahiromz: Define `REACTOS_RTL_RESOURCE` in Hebrew or Arabib language files, and later include `version.rc`.

HBelusca: The reason is that the first inclusion of version.rc done in the main resource file, creates the language-neutral version resource. To get the mechanism of this PR to work, one would also need a language-specific version resource, and the only way to do that is to re-include version.rc from within the localized (per-language) resource file, e.g. he-IL.rc , so that this lang-specific version resource gets the localized file version description string containing the two left-to-right symbols.

katahiromz: > need a language-specific version resource

I see. Thanks.

base/applications/charmap/lang/he-IL.rc
```diff
@@ -1,6 +1,14 @@
 /* Translated by Baruch Rutman */
 LANGUAGE LANG_HEBREW, SUBLANG_DEFAULT
 
+/*
+ * Enable RTL mirroring by adding two left-to-right marks (LRMs)
+ * in front of the FileDescription value of the version stamping.
+ */
```
katahiromz, line 7:
```suggestion
/*
 * This is a language-specific version info.
 * Enable RTL mirroring by adding two left-to-right marks (LRMs).
 */
```

base/applications/charmap/lang/he-IL.rc
katahiromz: If we want some language-specific version info, then some properties might be changed:

https://github.com/reactos/reactos/blob/139809b747aed5f0afc34734d5703845efbcd435/sdk/include/reactos/version.rc#L118

https://github.com/reactos/reactos/blob/139809b747aed5f0afc34734d5703845efbcd435/sdk/include/reactos/version.rc#L138

peterooch:
I was thinking about doing something like this a while ago as hardcoding languages in code is not a good idea and better to have it in the language specific parts, like how mirroring dialog boxes its in the resources only.
I would prefer to have it as a much less lines per file so to not over complicate it for translators, like
```
#include <rtl_version.rc>
``` 
or 
```
#define REACTOS_RTL_RESOURCE
#include <version.rc>
```

learn-more:
@HBelusca do you want to continue this work, or can we close it?

learn-more:

PR #5128, opened 2023-03-07
[BOOT] Support compiling freeldr as a UEFI loader as well - Implement UI functions

DarkFire01:
## Purpose

First batch of changes to implement a UEFI version of freeldr

## Proposed changes

- Compile freeldr as EFI binary on top of the existing loader
- Stub out various functions so we can create a UEFI mach type in freeldr
- Implement all of the video output function so we can display a pretty freeldr BSoD

PR #5141, opened 2023-03-09
Freeldr - Preparation for UEFI

HBelusca:
## Purpose

Move platform-specific build targets into dedicated .cmake files.
Future UEFI port will use such layout.
Prerequisite for PR #5128 

## Proposed changes

```
[FREELDR] Move common x64 runtime-specific ASM helpers to a misc file.
```
and
```
[FREELDR] Move platform-specific build targets into dedicated .cmake files.

- Keep the common definitions and  file lists in CMakeLists.txt
- Move PC-AT & compatibles (PC-98, XBOX) definitions in pcat.cmake
- Future UEFI-specific definitions will be in an uefi.cmake
```

### DIFF
--- a/boot/freeldr/freeldr/CMakeLists.txt
+++ b/boot/freeldr/freeldr/CMakeLists.txt
@@ -163,6 +163,7 @@ elseif(ARCH STREQUAL "amd64")
     list(APPEND FREELDR_COMMON_ASM_SOURCE
         arch/amd64/entry.S
         arch/amd64/int386.S
+        arch/amd64/misc.S
         arch/amd64/pnpbios.S
         arch/amd64/linux.S)
 
--- a/boot/freeldr/freeldr/CMakeLists.txt
+++ b/boot/freeldr/freeldr/CMakeLists.txt
@@ -1,30 +1,39 @@
+##
+## PROJECT:     FreeLoader
+## LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+## PURPOSE:     Common build definitions
+## COPYRIGHT:   Copyright 2003 Brian Palmer <brianp@sginet.com>
+##              Copyright 2011-2014 Amine Khaldi <amine.khaldi@reactos.org>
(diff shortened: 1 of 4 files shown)


HBelusca:
Cc @DarkFire01 and @disean too :) (as you're not in the reviewers list)

DarkFire01 (approved):
I've been using this for awhile in all my UEFI stuff; I'll rebase the PR whenever this gets merged. I haven't seen any down sides in my testing!

boot/freeldr/freeldr/CMakeLists.txt
```diff
@@ -252,91 +128,10 @@ list(APPEND FREELDR_BASE_SOURCE
...
-    add_custom_target(setupldr ALL DEPENDS freeldr_pe)
-endif()
+# if(ARCH STREQUAL "i386")
+#     # Must be included together with disk/scsiport.c
+#     list(APPEND FREELDR_BASE_SOURCE
+#         ${CMAKE_CURRENT_BINARY_DIR}/freeldr_pe.def)
+# endif()
 
-add_cd_file(TARGET setupldr FILE ${CMAKE_CURRENT_BINARY_DIR}/setupldr.sys DESTINATION loader NO_CAB FOR bootcd regtest)
+include(pcat.cmake)
```
julenuri, line 137: ``pcat.cmake`` filename does not fit well, because it includes PC-98 which is not PC/AT compatible, and even some ARM stuff; Better name it ``legacy.cmake`` or ``lgcyboot.cmake`` (contrary to UEFI boot)

HBelusca: True. Well (apart from ARM which is here "legacy" because it doesn't use uefi and could be in its own file as well), the pc-98 and xbox are bios-based architectures.

julenuri: > For info, I chose "pcat" also because this is the name given to windows bootmgr when it's not efi

In Windows they switched to bootmgr since Vista, and it never supported anything besides PC/AT, so it makes sense for the bootmgr; But our FreeLoader is meant to be a NTLDR replacement; Maybe `bios.cmake` then?

PR #5143, opened 2023-03-10
[NTOS:KD] Move handling of Dmesg buffer from screen provider to KDBG provider.

HBelusca:
## Purpose

The dmesg command is now available even if screen output is disabled.

JIRA issue: [CORE-10749](https://jira.reactos.org/browse/CORE-10749)

----

Commit isolated from https://github.com/reactos/reactos/pull/4600/commits/aeb843e01476f104c0e5ac5c6214a2673275a15a of PR #4600 (branch "CORE-10749"), cleaned up and rebased on top of master.

### DIFF
--- a/ntoskrnl/include/internal/kd.h
+++ b/ntoskrnl/include/internal/kd.h
@@ -53,6 +53,17 @@ VOID
 
 /* INIT ROUTINES *************************************************************/
 
+KIRQL
+NTAPI
+KdbpAcquireLock(
+    _In_ PKSPIN_LOCK SpinLock);
+
+VOID
+NTAPI
+KdbpReleaseLock(
+    _In_ PKSPIN_LOCK SpinLock,
+    _In_ KIRQL OldIrql);
+
 VOID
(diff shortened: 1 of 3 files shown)


ntoskrnl/kdbg/kdb_symbols.c
```diff
@@ -330,15 +330,15 @@ KdbSymProcessSymbols(
     KeSetEvent(&SymbolsToLoadEvent, IO_NO_INCREMENT, FALSE);
 }
 
+
 VOID
```
HBelusca, line 334: Note: All the following functions (incl. KdbInitialize) should be moved out of kdb_symbols, since they are unrelated in general with symbols management. This will be done elsewhere (not in this PR).

ericsomto (changes requested):
Untested but I think it should not be harmful to merge

HBelusca:
> Untested but I think it should not be harmful to merge

Well, I usually test what I commit, and here, I can confirm that it works as intended.

PR #5145, opened 2023-03-10
[MSPAINT] Remember status bar visibility

katahiromz:
## Purpose
Some users might want to keep the status bar hidden.

## Proposed changes

- Add `ShowStatusBar` registry setting.
- Save `ShowStatusBar` status and restore the status on startup.

## TODO

- [x] Do tests.

### DIFF
--- a/base/applications/mspaint/registry.cpp
+++ b/base/applications/mspaint/registry.cpp
@@ -74,6 +74,7 @@ void RegistrySettings::LoadPresets()
     FontsPositionX = 0;
     FontsPositionY = 0;
     ShowTextTool = TRUE;
+    ShowStatusBar = TRUE;
 
     LOGFONT lf;
     GetObject(GetStockObject(DEFAULT_GUI_FONT), sizeof(lf), &lf);
@@ -100,6 +101,7 @@ void RegistrySettings::Load()
         ReadDWORD(view, _T("ThumbXPos"),     ThumbXPos,     TRUE);
         ReadDWORD(view, _T("ThumbYPos"),     ThumbYPos,     TRUE);
         ReadDWORD(view, _T("UnitSetting"),   UnitSetting,   FALSE);
+        ReadDWORD(view, _T("ShowStatusBar"), ShowStatusBar, FALSE);
 
         ULONG pnBytes = sizeof(WINDOWPLACEMENT);
         view.QueryBinaryValue(_T("WindowPlacement"), &WindowPlacement, &pnBytes);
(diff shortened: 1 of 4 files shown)


base/applications/mspaint/winproc.cpp
```diff
@@ -730,6 +730,7 @@ LRESULT CMainWindow::OnCommand(UINT nMsg, WPARAM wParam, LPARAM lParam, BOOL& bH
             break;
         case IDM_VIEWSTATUSBAR:
             ::ShowWindow(hStatusBar, ::IsWindowVisible(hStatusBar) ? SW_HIDE : SW_SHOW);
+            registrySettings.ShowStatusBar = ::IsWindowVisible(hStatusBar);
```
HBelusca, line 733: You may exchange those two lines, and use in the `ShowWindow` call the resulting value of `registrySettings.ShowStatusBar` ? :thinking:

base/applications/mspaint/registry.h
```diff
@@ -41,6 +41,7 @@ class RegistrySettings
     DWORD FontsPositionX;
     DWORD FontsPositionY;
     DWORD ShowTextTool;
+    DWORD ShowStatusBar;
```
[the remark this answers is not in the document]
katahiromz: Fixed with type casts `(DWORD&)`.

base/applications/mspaint/winproc.cpp
katahiromz: Fixed.

base/applications/mspaint/registry.h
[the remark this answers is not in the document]
HBelusca: that sounds ugly

base/applications/mspaint/registry.cpp
```diff
@@ -124,7 +126,7 @@ void RegistrySettings::Load()
         ReadDWORD(text, _T("PointSize"),    PointSize,      FALSE);
         ReadDWORD(text, _T("PositionX"),    FontsPositionX, FALSE);
         ReadDWORD(text, _T("PositionY"),    FontsPositionY, FALSE);
-        ReadDWORD(text, _T("ShowTextTool"), ShowTextTool,   FALSE);
+        ReadDWORD(text, _T("ShowTextTool"), (DWORD&)ShowTextTool, FALSE);
```
HBelusca, line 129: I suppose that if no typecast, compilation emits an error? Then I think I'll revert my original suggestion, and you may actually keep those as DWORDs instead of BOOL :sweat_smile:
In any case, this shouldn't change the rest of your code (the usages in ShowWindow)

katahiromz: `../base/applications/mspaint/registry.cpp:129:45: error: cannot bind non-const lvalue reference of type 'DWORD&' {aka 'long unsigned int&'} to an rvalue of type 'DWORD' {aka 'long unsigned int'}`

katahiromz: Reverted.

base/applications/mspaint/registry.h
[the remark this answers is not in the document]
katahiromz: Reverted.

HBelusca (approved):

PR #5148, opened 2023-03-11
[INCLUDE][CMLIB][CABMAN][XML2SDB] Adapt C_ASSERT macro to C++

katahiromz:
## Purpose

The `C_ASSERT` macro in Win32 is known as the "compile-time assertion" macro. 
Using `C_ASSERT` macro in `mspaint` as below caused some linkage errors:

```txt
C_ASSERT(16 == TOOL_MAX);
```

```txt
sdk/include/psdk/winnt.h:807:38: error: conflicting declaration of 'char (* c_assert())[1]' with 'C' linkage
 #define C_ASSERT(expr) extern char (*c_assert(void)) [(expr) ? 1 : -1]
                                      ^~~~~~~~
../sdk/include/psdk/wincon.h:251:1: note: in expansion of macro 'C_ASSERT'
 C_ASSERT(FIELD_OFFSET(KEY_EVENT_RECORD, uChar) == 0xA);
 ^~~~~~~~
sdk/include/psdk/winnt.h:807:38: note: previous declaration with 'C++' linkage
 #define C_ASSERT(expr) extern char (*c_assert(void)) [(expr) ? 1 : -1]
                                      ^~~~~~~~
../base/applications/mspaint/toolsmodel.h:31:1: note: in expansion of macro 'C_ASSERT'
 C_ASSERT(16 == TOOL_MAX);
```

Make `C_ASSERT` macro fully available in C++.

## Proposed changes

- Use an array type in a function declaration.

## TODO

- [x] Do tests.

### DIFF
--- a/sdk/include/xdk/ntbasedef.h
+++ b/sdk/include/xdk/ntbasedef.h
@@ -788,7 +788,7 @@ $endif(_WINNT_)
 #endif /* _M_AMD64 */
 
 /* C_ASSERT Definition */
-#define C_ASSERT(expr) int __C_ASSERT__(int c_assert[(expr) ? 1 : -1])
+#define C_ASSERT(expr) typedef char __C_ASSERT__[(expr) ? 1 : -1]
 
 /* Eliminate Microsoft C/C++ compiler warning 4715 */
 #if defined(_MSC_VER)
--- a/sdk/include/xdk/ntbasedef.h
+++ b/sdk/include/xdk/ntbasedef.h
@@ -788,7 +788,7 @@ $endif(_WINNT_)
 #endif /* _M_AMD64 */
 
 /* C_ASSERT Definition */
-#define C_ASSERT(expr) EXTERN_C char (*c_assert(void)) [(expr) ? 1 : -1]
+#define C_ASSERT(expr) int __C_ASSERT__(int c_assert[(expr) ? 1 : -1])
 
 /* Eliminate Microsoft C/C++ compiler warning 4715 */
 #if defined(_MSC_VER)
--- a/sdk/include/xdk/ntbasedef.h
+++ b/sdk/include/xdk/ntbasedef.h
@@ -788,7 +788,11 @@ $endif(_WINNT_)
 #endif /* _M_AMD64 */
 
 /* C_ASSERT Definition */
-#define C_ASSERT(expr) extern char (*c_assert(void)) [(expr) ? 1 : -1]
+#ifdef __cplusplus
+    #define C_ASSERT(expr) extern "C" char (*c_assert(void)) [(expr) ? 1 : -1]
+#else
+    #define C_ASSERT(expr) extern char (*c_assert(void)) [(expr) ? 1 : -1]
+#endif
 
 /* Eliminate Microsoft C/C++ compiler warning 4715 */
 #if defined(_MSC_VER)
--- a/sdk/include/xdk/ntbasedef.h
+++ b/sdk/include/xdk/ntbasedef.h
@@ -788,11 +788,7 @@ $endif(_WINNT_)
 #endif /* _M_AMD64 */
 
 /* C_ASSERT Definition */
-#ifdef __cplusplus
-    #define C_ASSERT(expr) extern "C" char (*c_assert(void)) [(expr) ? 1 : -1]
-#else
-    #define C_ASSERT(expr) extern char (*c_assert(void)) [(expr) ? 1 : -1]
-#endif
+#define C_ASSERT(expr) EXTERN_C char (*c_assert(void)) [(expr) ? 1 : -1]
 
 /* Eliminate Microsoft C/C++ compiler warning 4715 */
 #if defined(_MSC_VER)
--- a/sdk/tools/cabman/cabinet.h
+++ b/sdk/tools/cabman/cabinet.h
@@ -33,7 +33,7 @@
 #endif
 
 #if !defined(C_ASSERT)
-#define C_ASSERT(expr) int __C_ASSERT__(int c_assert[(expr) ? 1 : -1])
+#define C_ASSERT(expr) typedef char __C_ASSERT__[(expr) ? 1 : -1]
 #endif
 
 #if defined(_WIN32)
--- a/sdk/tools/cabman/cabinet.h
+++ b/sdk/tools/cabman/cabinet.h
@@ -33,7 +33,7 @@
 #endif
 
 #if !defined(C_ASSERT)
-#define C_ASSERT(expr) extern char (*c_assert(void)) [(expr) ? 1 : -1]
+#define C_ASSERT(expr) int __C_ASSERT__(int c_assert[(expr) ? 1 : -1])
 #endif
 
 #if defined(_WIN32)
--- a/sdk/tools/xml2sdb/xml2sdb.cpp
+++ b/sdk/tools/xml2sdb/xml2sdb.cpp
@@ -17,7 +17,7 @@ static const GUID GUID_NULL = { 0 };
 static const char szCompilerVersion[] = "1.7.0.1";
 
 #if !defined(C_ASSERT)
-#define C_ASSERT(expr) extern char (*c_assert(void)) [(expr) ? 1 : -1]
+#define C_ASSERT(expr) int __C_ASSERT__(int c_assert[(expr) ? 1 : -1])
 #endif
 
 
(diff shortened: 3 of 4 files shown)


katahiromz:
```txt
2023-03-11T05:39:15.0933892Z In file included from sdk/include/ddk/ntddk.h:47,
2023-03-11T05:39:15.0934413Z                  from ../src/sdk/lib/drivers/wdf/shared/inc/primitives/km/mxkm.h:28,
2023-03-11T05:39:15.0934959Z                  from ../src/sdk/lib/drivers/wdf/shared/inc/primitives/common/mx.h:42,
2023-03-11T05:39:15.0935501Z                  from ../src/sdk/lib/drivers/wdf/shared/core/coreprivshared.hpp:46,
2023-03-11T05:39:15.0936040Z                  from ../src/sdk/lib/drivers/wdf/shared/core/fxcxdeviceinit.cpp:26:
2023-03-11T05:39:15.0937026Z ../src/sdk/lib/drivers/wdf/shared/inc/primitives/km/mxeventkm.h: In constructor 'MxEvent::MxEvent()':
2023-03-11T05:39:15.0937875Z ../src/sdk/include/psdk/basetyps.h:10:27: error: expected unqualified-id before string constant
2023-03-11T05:39:15.0938363Z  #  define EXTERN_C extern "C"
2023-03-11T05:39:15.0938830Z                            ^~~
2023-03-11T05:39:15.0939906Z sdk/include/psdk/ntdef.h:872:24: note: in expansion of macro 'EXTERN_C'
2023-03-11T05:39:15.0940986Z  #define C_ASSERT(expr) EXTERN_C char (*c_assert(void)) [(expr) ? 1 : -1]
2023-03-11T05:39:15.0941332Z                         ^~~~~~~~
```

katahiromz:
https://stackoverflow.com/questions/30440369/error-expected-unqualified-id-on-extern-c

sdk/include/xdk/ntbasedef.h
```diff
@@ -788,7 +788,7 @@ extern "C++" { \
 #endif /* _M_AMD64 */
 
 /* C_ASSERT Definition */
-#define C_ASSERT(expr) extern char (*c_assert(void)) [(expr) ? 1 : -1]
+#define C_ASSERT(expr) typedef char C_ASSERT_failed_Please_check_expr_is_TRUE[(expr) ? 1 : -1]
```
HBelusca, line 791: I think it would be better to just copy those already-existing definitions, see e.g.
https://git.reactos.org/?p=reactos.git;a=blob;f=sdk/include/psdk/intsafe.h;hb=072503c1fc133f48c4d710362b42b4cc4fb7f192#l70

katahiromz: Fixed.

sdk/include/xdk/ntbasedef.h
```diff
@@ -787,8 +787,10 @@ extern "C++" { \
  #define RotateRight16 _rotr16
 #endif /* _M_AMD64 */
 
-/* C_ASSERT Definition */
-#define C_ASSERT(expr) extern char (*c_assert(void)) [(expr) ? 1 : -1]
+/* Static assert */
+#ifndef C_ASSERT
```
ThFabba, line 791: Our headers are supposed to be compatible with Microsoft's, so I wouldn't recommend adding checks like this "just because" (you may end up with applications that compile fine with ours but not with MS's)

ThFabba (approved):

sdk/include/xdk/ntbasedef.h
```diff
@@ -787,8 +787,10 @@ extern "C++" { \
...
+# define C_ASSERT(e) typedef char __C_ASSERT__[(e) ? 1 : -1]
+#endif /* C_ASSERT */
```
HBelusca, line 793: Note: the intsafe.h has this MSC_VER conditional. Would it be nice to have it there too?
```suggestion
#ifdef _MSC_VER
# define C_ASSERT(e) typedef char __C_ASSERT__[(e)?1:-1]
#else
# define C_ASSERT(e) extern void __C_ASSERT__(int [(e)?1:-1])
#endif
```

katahiromz: This `extern` function declaration had a conflict on C++.

HBelusca: That's interesting.... Would this mean that e.g. using intsafe.h (from which that extern version comes from) would generate a compile error as well?

HBelusca: Seriously, what a mess: see for example those commits:
https://git.reactos.org/?p=reactos.git;a=commit;h=f8e59069e2301e3dd26cce1cadc1d9ac23f932a8
https://git.reactos.org/?p=reactos.git;a=commit;h=f73d776d30f1d4c11f1845d5bfc2a7a7589217b7
> New definition of C_ASSERT and _STATIC_ASSERT using a function prototype. This way we avoid duplicate definition errors like with a typedef and unused function warning like with the extern.

@ThFabba @tkreuzer Can you shed some light on this?

katahiromz: Yep if `C_ASSERT`'s of C code and C++ code are mixed in a module that uses `<intsafe.h>`.

sdk/include/xdk/ntbasedef.h
katahiromz: Fixed.

HBelusca:
I will re-verify myself this PR before approving. Official Windows headers use the macro without the LINE thing, and we know they are able to compile their C++ code without problems.

HBelusca:
Here are the results of my investigations:
1. Definitions like what we have in intsafe.h will break compilation with GCC:
```c
#ifdef _MSC_VER
# define C_ASSERT(e) typedef char __C_ASSERT__[(e)?1:-1]
#else
# define C_ASSERT(e) extern void __C_ASSERT__(int [(e)?1:-1])   // <---- this is the definition used when compiling with GCC.
#endif
```
2. Unconditionally using the first definition, i.e. both with MSVC and GCC, will work when compiling the C++ with GCC.

sdk/tools/cabman/cabinet.h
```diff
@@ -33,7 +33,7 @@
 #endif
 
 #if !defined(C_ASSERT)
-#define C_ASSERT(expr) extern char (*c_assert(void)) [(expr) ? 1 : -1]
+#define C_ASSERT(expr) int __C_ASSERT__(int c_assert[(expr) ? 1 : -1])
```
HBelusca, line 36:
```suggestion
#define C_ASSERT(expr) typedef char __C_ASSERT__[(expr) ? 1 : -1]
```

sdk/tools/xml2sdb/xml2sdb.cpp
```diff
@@ -17,7 +17,7 @@ static const GUID GUID_NULL = { 0 };
 static const char szCompilerVersion[] = "1.7.0.1";
 
 #if !defined(C_ASSERT)
-#define C_ASSERT(expr) extern char (*c_assert(void)) [(expr) ? 1 : -1]
+#define C_ASSERT(expr) int __C_ASSERT__(int c_assert[(expr) ? 1 : -1])
```
HBelusca, line 20:
```suggestion
#define C_ASSERT(expr) typedef char __C_ASSERT__[(expr) ? 1 : -1]
```

HBelusca (changes requested):
Those are the MS PSDK-compatible definitions, and indeed they work when compiling C and C++ with both MSVC and GCC.

sdk/tools/xml2sdb/xml2sdb.cpp
```diff
@@ -17,7 +17,7 @@ static const GUID GUID_NULL = { 0 };
 static const char szCompilerVersion[] = "1.7.0.1";
 
 #if !defined(C_ASSERT)
-#define C_ASSERT(expr) extern char (*c_assert(void)) [(expr) ? 1 : -1]
+#define C_ASSERT(expr) typedef char __C_ASSERT__[(expr) ? 1 : -1]
```
learn-more, line 20: Why does this need to change?
Didn't it work before?

katahiromz: mspaint uses C++. The target macro declares extern function. An extern function is either C function or C++. If they are mixed in a module, that causes linkage errors.

katahiromz:
Ok, we use standard `static_assert` in C++11.
Closed.

PR #5149, opened 2023-03-11
[KERNEL32][KERNEL32_APITEST] Implement user-mode UEFI / Firmware API

RatinCN:
## Purpose

Implement user-mode UEFI / Firmware API export from kernel32.dll

JIRA issue: None

## Proposed changes

Implement those APIs and related definitions will be added together

## TODO

- [x] Implement firmware environment variable APIs, `(Get/Set)FirmwareEnvironmentVariable[Ex][A/W]`
- [x] ~Implement `GetFirmwareType`~
- [x] Test those implement in Windows (because kernel-mode interfaces are unimplemented yet)
- [x] Add apitest

apitest cooperated with binarymaster


learn-more:
How do you test / validate this, are there apitests for this?
Did you write tests?

RatinCN:
> How do you test / validate this, are there apitests for this? Did you write tests?

In TODO list, I'll compile this DLL and write a program (not apitest) to call those API, and also call Windows API, compare the result to prove the implement is right.

Those APIs are depend on KM interfaces and UEFI environment, they're not ready. So I'm not planned to write apitests in this PR.

dll/win32/kernel32/client/sysinfo.c
```diff
@@ -1,14 +1,14 @@
...
+ * LICENSE:         GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
  * PURPOSE:         System Information Functions
- * PROGRAMMERS:     Emanuele Aliberti
+ * COPYRIGHT:       Emanuele Aliberti
  *                  Christoph von Wittich
  *                  Thomas Weidenmueller
  *                  Gunnar Andre Dalsnes
  *                  Stanislav Motylkov (x86corez@gmail.com)
  *                  Mark Jansen (mark.jansen@reactos.org)
+ *                  Copyright 2023 Ratin Gao <ratin@knsoft.org>
```
JoachimHenze, line 11: when adding yourself to this already long list, I would recommend to strip 2 inactive "Thomas Weidenmueller" and "Gunnar Andre Dalsnes" in return. (Alternatively don't extend the list even further).

dll/win32/kernel32/client/sysinfo.c
```diff
@@ -78,11 +78,11 @@ GetSystemInfoInternal(IN PSYSTEM_BASIC_INFORMATION BasicInfo,
...
-BaseQuerySystemFirmware(IN DWORD FirmwareTableProviderSignature,
-                        IN DWORD FirmwareTableID,
-                        OUT PVOID pFirmwareTableBuffer,
-                        IN DWORD BufferSize,
-                        IN SYSTEM_FIRMWARE_TABLE_ACTION Action)
+BaseQuerySystemFirmware(_In_ DWORD FirmwareTableProviderSignature,
+                        _In_ DWORD FirmwareTableID,
+                        _Out_writes_bytes_to_opt_(BufferSize, return) PVOID pFirmwareTableBuffer,
+                        _In_ DWORD BufferSize,
+                        _In_ SYSTEM_FIRMWARE_TABLE_ACTION Action)
```
HBelusca, line 85:
```suggestion
BaseQuerySystemFirmware(
    _In_ DWORD FirmwareTableProviderSignature,
    _In_ DWORD FirmwareTableID,
    _Out_writes_bytes_to_opt_(BufferSize, return) PVOID pFirmwareTableBuffer,
    _In_ DWORD BufferSize,
    _In_ SYSTEM_FIRMWARE_TABLE_ACTION Action)
```

dll/win32/kernel32/client/sysinfo.c
```diff
@@ -557,6 +729,36 @@ GetSystemFirmwareTable(IN DWORD FirmwareTableProviderSignature,
...
 }
 
+#if (_WIN32_WINNT >= _WIN32_WINNT_WIN8)
+_Success_(return != FALSE)
+BOOL
+WINAPI
+GetFirmwareType(
+    _Out_ PFIRMWARE_TYPE FirmwareType)
+{
+  NTSTATUS Status;
```
HBelusca, line 739: Indentation in this function is broken, please fix.

ntoskrnl/ex/sysinfo.c
```diff
@@ -562,25 +562,26 @@ NtEnumerateSystemEnvironmentValuesEx(IN ULONG InformationClass,
...
-NtQuerySystemEnvironmentValueEx(IN PUNICODE_STRING VariableName,
-                                IN LPGUID VendorGuid,
-                                IN PVOID Value,
-                                IN OUT PULONG ReturnLength,
-                                IN OUT PULONG Attributes)
+NtQuerySystemEnvironmentValueEx(_In_ PUNICODE_STRING VariableName,
+                                _In_ LPGUID VendorGuid,
+                                _Out_opt_ PVOID Value,
+                                _Inout_ PULONG ReturnLength,
+                                _Out_opt_ PULONG Attributes)
```
HBelusca, line 572:
```suggestion
NtQuerySystemEnvironmentValueEx(
    _In_ PUNICODE_STRING VariableName,
    _In_ LPGUID VendorGuid,
    _Out_opt_ PVOID Value,
    _Inout_ PULONG ReturnLength,
    _Out_opt_ PULONG Attributes)
```

ntoskrnl/ex/sysinfo.c
```diff
@@ -562,25 +562,26 @@ NtEnumerateSystemEnvironmentValuesEx(IN ULONG InformationClass,
...
-NtSetSystemEnvironmentValueEx(IN PUNICODE_STRING VariableName,
-                              IN LPGUID VendorGuid,
-                              IN PVOID Value,
-                              IN OUT PULONG ReturnLength,
-                              IN OUT PULONG Attributes)
+NtSetSystemEnvironmentValueEx(_In_ PUNICODE_STRING VariableName,
+                              _In_ LPGUID VendorGuid,
+                              _In_reads_bytes_opt_(ValueLength) PVOID Value,
+                              _In_ ULONG ValueLength,
+                              _In_ ULONG Attributes)
```
HBelusca, line 584:
```suggestion
NtSetSystemEnvironmentValueEx(
    _In_ PUNICODE_STRING VariableName,
    _In_ LPGUID VendorGuid,
    _In_reads_bytes_opt_(ValueLength) PVOID Value,
    _In_ ULONG ValueLength,
    _In_ ULONG Attributes)
```

sdk/include/psdk/winbase.h
```diff
@@ -3156,8 +3174,79 @@ SetFirmwareEnvironmentVariableW(
   _In_reads_bytes_opt_(nSize) PVOID pValue,
   _In_ DWORD nSize);
 
+#ifdef UNICODE
+#define SetFirmwareEnvironmentVariable SetFirmwareEnvironmentVariableW
+#else
+#define SetFirmwareEnvironmentVariable SetFirmwareEnvironmentVariableA
 #endif
 
+#endif
```
HBelusca, line 3183: To which ifdef/... this corresponds?

ntoskrnl/ex/sysinfo.c
```diff
@@ -562,25 +562,26 @@ NtEnumerateSystemEnvironmentValuesEx(IN ULONG InformationClass,
     return STATUS_NOT_IMPLEMENTED;
 }
 
+_Success_(NT_SUCCESS(return))
```
HBelusca, line 565: Isn't it already understood? (annotation included in the typedef of NTSTATUS)

sdk/include/ndk/exfuncs.h
```diff
@@ -390,15 +390,16 @@ NtQuerySystemEnvironmentValue(
     PULONG ReturnLength
 );
 
+_Success_(NT_SUCCESS(return))
```
HBelusca, line 393: Understood already via NTSTATUS typedef.

sdk/include/psdk/winbase.h
```diff
@@ -3156,8 +3174,79 @@ SetFirmwareEnvironmentVariableW(
...
+    _In_ DWORD nSize,
+    _In_ DWORD dwAttributes);
+
+#ifdef UNICODE
+#define SetFirmwareEnvironmentVariableEx SetFirmwareEnvironmentVariableExW
+#else
+#define SetFirmwareEnvironmentVariableEx SetFirmwareEnvironmentVariableExA
+#endif
+
+_Success_(return != FALSE)
```
HBelusca, line 3241: Already understood because it's BOOL as return type, and it's not a function where FALSE means "success"...

HBelusca (changes requested):

dll/win32/kernel32/client/sysinfo.c
RatinCN: Committed.

dll/win32/kernel32/client/sysinfo.c
RatinCN: Fixed.

ntoskrnl/ex/sysinfo.c
RatinCN: Committed.

ntoskrnl/ex/sysinfo.c
RatinCN: Committed.

sdk/include/psdk/winbase.h
RatinCN: Comment added.

ntoskrnl/ex/sysinfo.c
RatinCN: Yes it is, thanks for your infomation, superfluous SAL notations are removed.

sdk/include/ndk/exfuncs.h
RatinCN: Superfluous SAL notations are removed.

sdk/include/psdk/winbase.h
RatinCN: Superfluous SAL notations are removed.

dll/win32/kernel32/client/sysinfo.c
```diff
@@ -413,68 +414,238 @@ GetNumaAvailableMemoryNode(IN UCHAR Node,
...
+    _In_ LPCWSTR lpGuid,
+    _In_reads_bytes_opt_(nSize) PVOID pValue,
+    _In_ DWORD nSize,
+    _In_ DWORD dwAttributes)
 {
-    STUB;
-    return 0;
+    NTSTATUS Status;
+    UNICODE_STRING VariableName, Namespace;
+    GUID NamespaceGuid;
```
HBelusca, line 544:
```suggestion
    GUID VendorGuid;
```

dll/win32/kernel32/client/sysinfo.c
```diff
@@ -413,68 +414,238 @@ GetNumaAvailableMemoryNode(IN UCHAR Node,
...
+
+    /* Check input parameters and build NT strings */
+    if (!lpName || !lpGuid)
+    {
+        SetLastError(ERROR_INVALID_PARAMETER);
+        return FALSE;
+    }
+    RtlInitUnicodeString(&VariableName, lpName);
+    RtlInitUnicodeString(&Namespace, lpGuid);
+    Status = RtlGUIDFromString(&Namespace, &NamespaceGuid);
```
HBelusca, line 554:
```suggestion
    Status = RtlGUIDFromString(&Namespace, &VendorGuid);
```

dll/win32/kernel32/client/sysinfo.c
```diff
@@ -413,68 +414,238 @@ GetNumaAvailableMemoryNode(IN UCHAR Node,
...
+    if (!NT_SUCCESS(Status))
+    {
+        BaseSetLastNTError(Status);
+        return FALSE;
+    }
+
+    /* Set firmware system environment variable value */
+    Status = NtSetSystemEnvironmentValueEx(&VariableName, &NamespaceGuid, pValue, nSize, dwAttributes);
```
HBelusca, line 562:
```suggestion
    Status = NtSetSystemEnvironmentValueEx(&VariableName, &VendorGuid, pValue, nSize, dwAttributes);
```

dll/win32/kernel32/client/sysinfo.c
```diff
@@ -413,68 +414,238 @@ GetNumaAvailableMemoryNode(IN UCHAR Node,
...
+    NTSTATUS Status;
+    BOOL Result;
+    UNICODE_STRING VariableName, Namespace;
+    ANSI_STRING AnsiVariableName, AnsiNamespaceGuid;
+
+    /* Check input parameters and build NT strings */
+    if (!lpName || !lpGuid)
+    {
+        SetLastError(ERROR_INVALID_PARAMETER);
+        return 0;
```
HBelusca, line 590:
```suggestion
        return FALSE;
```

dll/win32/kernel32/client/sysinfo.c
```diff
@@ -413,68 +414,238 @@ GetNumaAvailableMemoryNode(IN UCHAR Node,
...
+    }
+    RtlInitString(&AnsiVariableName, lpName);
+    Status = RtlAnsiStringToUnicodeString(&VariableName, &AnsiVariableName, TRUE);
+    if (!NT_SUCCESS(Status))
+    {
+        BaseSetLastNTError(Status);
+        return 0;
```
HBelusca, line 597:
```suggestion
        return FALSE;
```

dll/win32/kernel32/client/sysinfo.c
```diff
@@ -413,68 +414,238 @@ GetNumaAvailableMemoryNode(IN UCHAR Node,
...
+    }
+    RtlInitString(&AnsiNamespaceGuid, lpGuid);
+    Status = RtlAnsiStringToUnicodeString(&Namespace, &AnsiNamespaceGuid, TRUE);
+    if (!NT_SUCCESS(Status))
+    {
+        RtlFreeUnicodeString(&VariableName);
+        BaseSetLastNTError(Status);
+        return 0;
```
HBelusca, line 605:
```suggestion
        return FALSE;
```

dll/win32/kernel32/client/sysinfo.c
```diff
@@ -413,68 +414,238 @@ GetNumaAvailableMemoryNode(IN UCHAR Node,
...
+    _In_ LPCSTR lpName,
+    _In_ LPCSTR lpGuid,
+    _In_reads_bytes_opt_(nSize) PVOID pValue,
+    _In_ DWORD nSize,
+    _In_ DWORD dwAttributes)
+{
+    NTSTATUS Status;
+    BOOL Result;
+    UNICODE_STRING VariableName, Namespace;
+    ANSI_STRING AnsiVariableName, AnsiNamespaceGuid;
```
HBelusca, line 584:
```suggestion
    ANSI_STRING AnsiVariableName, AnsiNamespace;
```

dll/win32/kernel32/client/sysinfo.c
```diff
@@ -413,68 +414,238 @@ GetNumaAvailableMemoryNode(IN UCHAR Node,
...
+    }
+    RtlInitString(&AnsiVariableName, lpName);
+    Status = RtlAnsiStringToUnicodeString(&VariableName, &AnsiVariableName, TRUE);
+    if (!NT_SUCCESS(Status))
+    {
+        BaseSetLastNTError(Status);
+        return 0;
+    }
+    RtlInitString(&AnsiNamespaceGuid, lpGuid);
+    Status = RtlAnsiStringToUnicodeString(&Namespace, &AnsiNamespaceGuid, TRUE);
```
HBelusca, line 600:
```suggestion
    RtlInitString(&AnsiNamespace, lpGuid);
    Status = RtlAnsiStringToUnicodeString(&Namespace, &AnsiNamespace, TRUE);
```

dll/win32/kernel32/client/sysinfo.c
```diff
@@ -413,68 +414,238 @@ GetNumaAvailableMemoryNode(IN UCHAR Node,
...
+
+BOOL
+WINAPI
+SetFirmwareEnvironmentVariableW(
+    _In_ LPCWSTR lpName,
+    _In_ LPCWSTR lpGuid,
+    _In_reads_bytes_opt_(nSize) PVOID pValue,
+    _In_ DWORD nSize)
+{
+    return SetFirmwareEnvironmentVariableExW(lpName, lpGuid, pValue, nSize, VARIABLE_ATTRIBUTE_NON_VOLATILE);
```
HBelusca, line 629: May place each parameter on its own line (see other functions).

dll/win32/kernel32/client/sysinfo.c
HBelusca, line 584: Maybe you can use shorter variable names, like "Name, Guid" and "AnsiName, AnsiGuid".

dll/win32/kernel32/client/sysinfo.c
```diff
@@ -413,68 +414,238 @@ GetNumaAvailableMemoryNode(IN UCHAR Node,
...
+    Status = RtlGUIDFromString(&Namespace, &NamespaceGuid);
+    if (!NT_SUCCESS(Status))
+    {
+        BaseSetLastNTError(Status);
+        return 0;
+    }
+
+    /* Query firmware system environment variable value */
+    Length = nSize;
+    Status = NtQuerySystemEnvironmentValueEx(&VariableName, &NamespaceGuid, pBuffer, &Length, pdwAttribubutes);
```
HBelusca, line 449:
```suggestion
    Status = NtQuerySystemEnvironmentValueEx(&VariableName,
                                             &NamespaceGuid,
                                             pBuffer,
                                             &Length,
                                             pdwAttribubutes);
```

dll/win32/kernel32/client/sysinfo.c
HBelusca, line 562:
```suggestion
    Status = NtSetSystemEnvironmentValueEx(&VariableName,
                                           &NamespaceGuid,
                                           pValue,
                                           nSize,
                                           dwAttributes);
```

dll/win32/kernel32/client/sysinfo.c
learn-more: Why?

dll/win32/kernel32/client/sysinfo.c
RatinCN: Committed

dll/win32/kernel32/client/sysinfo.c
RatinCN: Committed.

dll/win32/kernel32/client/sysinfo.c
RatinCN: Committed.

dll/win32/kernel32/client/sysinfo.c
RatinCN: Committed.

dll/win32/kernel32/client/sysinfo.c
RatinCN: Committed.

dll/win32/kernel32/client/sysinfo.c
RatinCN: Committed.

dll/win32/kernel32/client/sysinfo.c
RatinCN: Committed.

dll/win32/kernel32/client/sysinfo.c
RatinCN: Committed.

dll/win32/kernel32/client/sysinfo.c
RatinCN: Committed. Of course, it exceeded 100 slightly.

dll/win32/kernel32/client/sysinfo.c
RatinCN: I suggest use `AnsiVariableName` and `AnsiNamespace`, won't be so long

dll/win32/kernel32/client/sysinfo.c
RatinCN: Committed.

dll/win32/kernel32/client/sysinfo.c
RatinCN: Committed.

RatinCN:
apitest added, run on Win11 UEFI system and passed:
image

modules/rostests/apitests/kernel32/UEFIFirmware.c
```diff
@@ -0,0 +1,221 @@
...
+                                                   _A2W(EFI_TEST_GUID_STRING),
+                                                   &dwValue,
+                                                   sizeof(dwValue));
+        ok(dwLength,
+           "GetFirmwareEnvironmentVariableW failed with error: 0x%08lX\n",
+           GetLastError());
+        if (dwLength)
+        {
+            ok(dwLength == sizeof(EfiVariableValue) && dwValue == EfiVariableValue,
+               "The variable we get is different from what we set, "
```
HBelusca, line 128:
```suggestion
               "Retrieved variable different from what we set, "
```

modules/rostests/apitests/kernel32/UEFIFirmware.c
```diff
@@ -0,0 +1,221 @@
...
+                                                   EFI_TEST_GUID_STRING,
+                                                   &dwValue,
+                                                   sizeof(dwValue));
+        ok(dwLength,
+           "GetFirmwareEnvironmentVariableA failed with error: 0x%08lX\n",
+           GetLastError());
+        if (dwLength)
+        {
+            ok(dwLength == sizeof(EfiVariableValue) && dwValue == EfiVariableValue,
+               "The variable we get is different from what we set, "
```
HBelusca, line 159:
```suggestion
               "Retrieved variable different from what we set, "
```

modules/rostests/apitests/kernel32/UEFIFirmware.c
```diff
@@ -0,0 +1,221 @@
...
+       GetLastError());
+    if (bResult)
+    {
+        dwLength = GetFirmwareEnvironmentVariableW(_A2W(EFI_TEST_VARIABLE_NAME),
+                                                   _A2W(EFI_TEST_GUID_STRING),
+                                                   &dwValue,
+                                                   sizeof(dwValue));
+        ok(dwLength == 0, "SetFirmwareEnvironmentVariableW didn't delete the variable!\n");
+    }
+    /* Set variable back and test ANSI version API */
```
HBelusca, line 184:
```suggestion
    /* Restore variable and test ANSI function */
```

modules/rostests/apitests/kernel32/UEFIFirmware.c
```diff
@@ -0,0 +1,221 @@
...
+    bResult = SetFirmwareEnvironmentVariableW(_A2W(EFI_TEST_VARIABLE_NAME),
+                                              _A2W(EFI_TEST_GUID_STRING),
+                                              &EfiVariableValue,
+                                              sizeof(EfiVariableValue));
+    if (!bResult)
+    {
+        skip("SetFirmwareEnvironmentVariableW failed to set variable back with error: 0x%08lX\n",
```
HBelusca, line 191:
```suggestion
        skip("SetFirmwareEnvironmentVariableW failed to restore variable with error: 0x%08lX\n",
```

modules/rostests/apitests/kernel32/UEFIFirmware.c
```diff
@@ -0,0 +1,221 @@
...
+                                               NULL,
+                                               0);
+    dwError = GetLastError();
+    ok(dwLength == 0, "dwLength = %lu, expected 0\n", dwLength);
+    if (dwLength || dwError == ERROR_INVALID_FUNCTION)
+    {
+        skip("The rest of tests require UEFI environment, test abort.\n");
+        return;
+    }
+    /* Test ANSI version API too */
```
HBelusca, line 50:
```suggestion
    /* Test ANSI function too */
```

modules/rostests/apitests/kernel32/UEFIFirmware.c
```diff
@@ -0,0 +1,221 @@
...
+    bResult = SetFirmwareEnvironmentVariableW(_A2W(EFI_TEST_VARIABLE_NAME),
+                                              _A2W(EFI_TEST_GUID_STRING),
+                                              &EfiVariableValue,
+                                              sizeof(EfiVariableValue));
+    dwError = GetLastError();
+    ok(!bResult && dwError == ERROR_PRIVILEGE_NOT_HELD,
+       "bResult = %d, LastError = %lu, expected bResult = 0, LastError = ERROR_PRIVILEGE_NOT_HELD\n",
+       bResult,
+       dwError);
+    /* Test ANSI version API too */
```
HBelusca, line 78:
```suggestion
    /* Test ANSI function too */
```

modules/rostests/apitests/kernel32/UEFIFirmware.c
```diff
@@ -0,0 +1,221 @@
...
+        return;
+    }
+    Privilege.PrivilegeCount = 1;
+    Privilege.Privileges[0].Luid = (LUID){ SE_SYSTEM_ENVIRONMENT_PRIVILEGE, 0 };
+    Privilege.Privileges[0].Attributes = SE_PRIVILEGE_ENABLED;
+    Status = NtAdjustPrivilegesToken(hToken, FALSE, &Privilege, sizeof(Privilege), NULL, &ReturnLength);
+    NtClose(hToken);
+    if (Status != STATUS_SUCCESS) {
+        skip("NtAdjustPrivilegesToken failed with status: 0x%08lX, test abort.\n", Status);
+        return;
```
HBelusca, line 104: Shouldn't you restore the old privileges?

modules/rostests/apitests/kernel32/UEFIFirmware.c
```diff
@@ -0,0 +1,221 @@
...
+    /* Set variable back and test ANSI version API */
+    bResult = SetFirmwareEnvironmentVariableW(_A2W(EFI_TEST_VARIABLE_NAME),
+                                              _A2W(EFI_TEST_GUID_STRING),
+                                              &EfiVariableValue,
+                                              sizeof(EfiVariableValue));
+    if (!bResult)
+    {
+        skip("SetFirmwareEnvironmentVariableW failed to set variable back with error: 0x%08lX\n",
+             GetLastError());
+        return;
```
HBelusca, line 193: Shouldn't you restore the old privileges?

modules/rostests/apitests/kernel32/UEFIFirmware.c
```diff
@@ -0,0 +1,221 @@
...
+       GetLastError());
+    if (bResult)
+    {
+        dwLength = GetFirmwareEnvironmentVariableA(EFI_TEST_VARIABLE_NAME,
+                                                   EFI_TEST_GUID_STRING,
+                                                   &dwValue,
+                                                   sizeof(dwValue));
+        ok(dwLength == 0, "SetFirmwareEnvironmentVariableA didn't delete the variable!\n");
+    }
+
```
HBelusca, line 210: Shouldn't you restore the old privileges?

modules/rostests/apitests/kernel32/UEFIFirmware.c
```diff
@@ -0,0 +1,221 @@
...
+               "The variable we get is different from what we set, "
+               "dwLength = %lu, dwValue = %lu, expected dwLength = %u, dwValue = %lu",
+               dwLength,
+               dwValue,
+               sizeof(EfiVariableValue),
+               EfiVariableValue);
+        }
+    }
+
+    /* Change variable value and test ANSI version API */
```
HBelusca, line 137:
```suggestion
    /* Change variable value and test ANSI function */
```

modules/rostests/apitests/kernel32/UEFIFirmware.c
```diff
@@ -0,0 +1,221 @@
...
+                                                   EFI_TEST_GUID_STRING,
+                                                   &dwValue,
+                                                   sizeof(dwValue));
+        ok(dwLength == 0, "SetFirmwareEnvironmentVariableA didn't delete the variable!\n");
+    }
+
+#if (_WIN32_WINNT >= 0x0602)
+    /* Test GetFirmwareType, should returns FirmwareTypeUefi if test reached here */
+    FIRMWARE_TYPE FirmwareType = FirmwareTypeUnknown;
+    bResult = GetFirmwareType(&FirmwareType);
```
HBelusca, line 214: Regarding all the `return;` on failures done above: Why not jumping down to testing this function still?

modules/rostests/apitests/kernel32/UEFIFirmware.c
```diff
@@ -0,0 +1,221 @@
...
+    ok(bResult && FirmwareType == FirmwareTypeUefi,
```
HBelusca, line 215: Not really: the firmware type can be "Bios" (see MSDN).

HBelusca (changes requested):

dll/win32/kernel32/client/sysinfo.c
```diff
@@ -413,68 +414,250 @@ GetNumaAvailableMemoryNode(IN UCHAR Node,
...
+
+BOOL
+WINAPI
+SetFirmwareEnvironmentVariableA(
+    _In_ LPCSTR lpName,
+    _In_ LPCSTR lpGuid,
+    _In_reads_bytes_opt_(nSize) PVOID pValue,
+    _In_ DWORD nSize)
+{
+    return SetFirmwareEnvironmentVariableExA(lpName, lpGuid, pValue, nSize, VARIABLE_ATTRIBUTE_NON_VOLATILE);
```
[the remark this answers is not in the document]
RatinCN: Committed.

dll/win32/kernel32/client/sysinfo.c
```diff
@@ -413,68 +414,250 @@ GetNumaAvailableMemoryNode(IN UCHAR Node,
...
+    }
+
+    /* Call unicode version interface */
+    Length = GetFirmwareEnvironmentVariableExW(VariableName.Buffer,
+                                               Namespace.Buffer,
+                                               pBuffer,
+                                               nSize,
+                                               pdwAttribubutes);
+
+    /* Clean and return */
```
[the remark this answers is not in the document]
RatinCN: Committed.

modules/rostests/apitests/kernel32/UEFIFirmware.c
```diff
@@ -0,0 +1,221 @@
...
+                                                  sizeof(EfiVariableValue));
+    dwErrorTemp = GetLastError();
+    ok(bResultTemp == bResult && dwErrorTemp == dwError,
+       "bResult = %d, LastError = %lu, expected bResult = %d, LastError = %lu\n",
+       bResultTemp,
+       dwErrorTemp,
+       bResult,
+       dwError);
+
+    /* Enable SE_SYSTEM_ENVIRONMENT_NAME privilege that required by following tests */
```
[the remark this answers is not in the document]
RatinCN: Committed.

modules/rostests/apitests/kernel32/UEFIFirmware.c
```diff
@@ -0,0 +1,221 @@
...
+    {
+        dwLength = GetFirmwareEnvironmentVariableA(EFI_TEST_VARIABLE_NAME,
+                                                   EFI_TEST_GUID_STRING,
+                                                   &dwValue,
+                                                   sizeof(dwValue));
+        ok(dwLength == 0, "SetFirmwareEnvironmentVariableA didn't delete the variable!\n");
+    }
+
+#if (_WIN32_WINNT >= 0x0602)
+    /* Test GetFirmwareType, should returns FirmwareTypeUefi if test reached here */
```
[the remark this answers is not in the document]
RatinCN: Committed.

modules/rostests/apitests/kernel32/UEFIFirmware.c
```diff
@@ -0,0 +1,221 @@
...
+    FIRMWARE_TYPE FirmwareType = FirmwareTypeUnknown;
+    bResult = GetFirmwareType(&FirmwareType);
+    ok(bResult && FirmwareType == FirmwareTypeUefi,
+       "GetFirmwareType failed, bResult = %d, FirmwareType = %d, expected bResult = 1, FirmwareType = %d\n",
+       bResult,
+       FirmwareType,
+       FirmwareTypeUefi);
```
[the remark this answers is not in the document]
RatinCN: Committed.

modules/rostests/apitests/kernel32/UEFIFirmware.c
```diff
@@ -0,0 +1,221 @@
...
+    BOOL bResult, bResultTemp;
+    DWORD dwError, dwErrorTemp, dwLength, dwLengthTemp, dwValue;
+    HANDLE hToken;
+    TOKEN_PRIVILEGES Privilege;
+    NTSTATUS Status;
+    ULONG ReturnLength;
+
+    /* Check this test run in legacy BIOS-based or UEFI system
+     * by calling GetFirmwareEnvironmentVariable with dummy name and GUID,
+     * should be failed with ERROR_INVALID_FUNCTION on the former or others on the latter. */
```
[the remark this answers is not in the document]
RatinCN: Committed. Note: on UEFI system, this calling won't success, because dummy namespace `{00000000-0000-0000-0000-000000000000}` should not exists, even if the namespace exists, it still fails because the dummy variable name is empty string `""`.

modules/rostests/apitests/kernel32/UEFIFirmware.c
```diff
@@ -0,0 +1,221 @@
...
+       dwErrorTemp,
+       dwLength,
+       dwError);
+
+    /* Generate a random as variable value to be used in this test */
+    RandomSeed = GetTickCount();
+    EfiVariableValue = RtlRandom(&RandomSeed);
+
+    /* Try to set firmware variable, should be failed with ERROR_PRIVILEGE_NOT_HELD,
+     * because no SE_SYSTEM_ENVIRONMENT_NAME privilege by default. */
```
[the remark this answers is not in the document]
RatinCN: Committed.

modules/rostests/apitests/kernel32/UEFIFirmware.c
```diff
@@ -0,0 +1,221 @@
...
+    ok(bResultTemp == bResult && dwErrorTemp == dwError,
+       "bResult = %d, LastError = %lu, expected bResult = %d, LastError = %lu\n",
+       bResultTemp,
+       dwErrorTemp,
+       bResult,
+       dwError);
+
+    /* Enable SE_SYSTEM_ENVIRONMENT_NAME privilege that required by following tests */
+    bResult = OpenProcessToken(GetCurrentProcess(), TOKEN_ADJUST_PRIVILEGES, &hToken);
+    if (!bResult) {
```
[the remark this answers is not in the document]
RatinCN: Committed.

modules/rostests/apitests/kernel32/UEFIFirmware.c
```diff
@@ -0,0 +1,221 @@
...
+        skip("OpenProcessToken failed with error: 0x%08lX, test abort.\n", GetLastError());
```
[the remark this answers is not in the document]
RatinCN: Committed.

modules/rostests/apitests/kernel32/UEFIFirmware.c
```diff
@@ -0,0 +1,221 @@
...
+        return;
+    }
+    Privilege.PrivilegeCount = 1;
+    Privilege.Privileges[0].Luid = (LUID){ SE_SYSTEM_ENVIRONMENT_PRIVILEGE, 0 };
+    Privilege.Privileges[0].Attributes = SE_PRIVILEGE_ENABLED;
+    Status = NtAdjustPrivilegesToken(hToken, FALSE, &Privilege, sizeof(Privilege), NULL, &ReturnLength);
+    NtClose(hToken);
+    if (Status != STATUS_SUCCESS) {
+        skip("NtAdjustPrivilegesToken failed with status: 0x%08lX, test abort.\n", Status);
```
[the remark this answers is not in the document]
RatinCN: Committed.

modules/rostests/apitests/kernel32/UEFIFirmware.c
RatinCN: Committed.

modules/rostests/apitests/kernel32/UEFIFirmware.c
RatinCN: Committed.

modules/rostests/apitests/kernel32/UEFIFirmware.c
RatinCN: Committed.

modules/rostests/apitests/kernel32/UEFIFirmware.c
RatinCN: Committed.

modules/rostests/apitests/kernel32/UEFIFirmware.c
RatinCN: Committed.

modules/rostests/apitests/kernel32/UEFIFirmware.c
RatinCN: In this code path, only one privilege was requested and fail, no privilege has been changed. So nothing to restore here.

modules/rostests/apitests/kernel32/UEFIFirmware.c
RatinCN: Committed, it will jump to exit and restore the old privilege state now.

modules/rostests/apitests/kernel32/UEFIFirmware.c
RatinCN: Committed, old privilege state will be restored after that now.

modules/rostests/apitests/kernel32/UEFIFirmware.c
RatinCN: Committed.

modules/rostests/apitests/kernel32/UEFIFirmware.c
RatinCN: Committed, now this function will be tested at the beginning.

modules/rostests/apitests/kernel32/UEFIFirmware.c
RatinCN: Committed, both of `FirmwareTypeBios` and `FirmwareTypeUefi` are okay now.

RatinCN:
Merged patch from binarymaster in chat, it check for result and firmware type separately.

dll/win32/kernel32/kernel32.spec
```diff
@@ -482,7 +482,10 @@
 @ stub -version=0x600+ GetFinalPathNameByHandleA
 @ stub -version=0x600+ GetFinalPathNameByHandleW
 @ stdcall GetFirmwareEnvironmentVariableA(str str ptr long)
+@ stdcall -version=0x602+ GetFirmwareEnvironmentVariableExA(str str ptr long ptr)
+@ stdcall -version=0x602+ GetFirmwareEnvironmentVariableExW(wstr wstr ptr long ptr)
 @ stdcall GetFirmwareEnvironmentVariableW(wstr wstr ptr long)
+@ stdcall -version=0x602+ GetFirmwareType(ptr)
```
HBelusca, line 488: Question for @tkreuzer / @learn-more (perhaps) : what's currently the best place to get the new _implemented_ (not stubbed) NT6+ functions? Here, with the version thing, or in kernel32_vista ?

sdk/include/xdk/extypes.h
```diff
@@ -346,3 +346,14 @@ typedef struct _ZONE_HEADER {
 #define EX_PUSH_LOCK ULONG_PTR
 #define PEX_PUSH_LOCK PULONG_PTR
 $endif (_NTIFS_)
+
+$if (_WINNT_ || _WDMDDK_)
+#if (NTDDI_VERSION >= NTDDI_VISTA)
```
HBelusca, line 351: Isn't this introduced only on Windows 8+?

sdk/include/ndk/extypes.h
```diff
@@ -1424,7 +1424,21 @@ typedef struct _SYSTEM_HANDLE_INFORMATION_EX
...
+#if (NTDDI_VERSION >= NTDDI_LONGHORN)
+typedef struct _SYSTEM_BOOT_ENVIRONMENT_INFORMATION
+{
+    GUID BootIdentifier;
+    FIRMWARE_TYPE FirmwareType;
+#if (NTDDI_VERSION >= NTDDI_WIN8)
+    ULONGLONG BootFlags;
+#endif
+} SYSTEM_BOOT_ENVIRONMENT_INFORMATION, *PSYSTEM_BOOT_ENVIRONMENT_INFORMATION;
+#endif
```
HBelusca, line 1439: See also Geoff Chappell:

> The SYSTEM_BOOT_ENVIRONMENT_INFORMATION is 0x20 bytes in both 32-bit and 64-bit Windows 8 and higher, but is only 0x18 bytes in earlier versions. _**The older, smaller structure is retained in Windows 8 and higher as SYSTEM_BOOT_ENVIRONMENT_V1.**_

modules/rostests/apitests/kernel32/UEFIFirmware.c
```diff
@@ -0,0 +1,253 @@
...
+
+    if (!bResult)
+        return;
+
+    FirmwareExpect = (bIsUEFI ? FirmwareTypeUefi : FirmwareTypeBios);
+    ok(FirmwareType == FirmwareExpect,
+       "FirmwareType is %d, but %d is expected.\n",
+       FirmwareType, FirmwareExpect);
+#else
+    skip("This test can be ran only when compiled for NT >= 6.2.\n");
```
HBelusca, line 48:
```suggestion
    skip("This test can be run only when compiled for NT >= 6.2.\n");
```

HBelusca: Note that alternatively, you could try loading at runtime the GetFirmwareType function with LoadLibrary("kernel32") + GetProcAddress("GetFirmwareType") and work from there.

modules/rostests/apitests/kernel32/UEFIFirmware.c
```diff
@@ -0,0 +1,253 @@
...
+
+#include "precomp.h"
+
+#include <ndk/psfuncs.h>
+#include <ndk/setypes.h>
+#include <ndk/sefuncs.h>
+#include <ndk/obfuncs.h>
+
+#define _A2W(quote) __A2W(quote)
+#define __A2W(quote) L##quote
```
HBelusca, line 16: Isn't it already done via `_T` or `TEXT` macro?

modules/rostests/apitests/kernel32/UEFIFirmware.c
```diff
@@ -0,0 +1,253 @@
...
+                                                   0);
+    dwErrorTemp = GetLastError();
+    ok(dwLengthTemp == dwLength && dwErrorTemp == dwError,
+       "dwLength = %lu, LastError = %lu, expected bResult = %lu, LastError = %lu\n",
+       dwLengthTemp,
+       dwErrorTemp,
+       dwLength,
+       dwError);
+
+    /* Generate a random as variable value to be used in this test */
```
HBelusca, line 95:
```suggestion
    /* Generate a random variable value to be used in this test */
```

modules/rostests/apitests/kernel32/UEFIFirmware.c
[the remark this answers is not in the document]
RatinCN: Yes, `bIsUEFI = (dwLength == 0 && dwError != ERROR_INVALID_FUNCTION);` did.

modules/rostests/apitests/kernel32/UEFIFirmware.c
RatinCN: `#define TEXT(s) L##s` is already defined in `k32.h`, but `_T` or `TEXT` seems depends on ANSI build or Unicode build.
Here convert `"xxx"` to `L"xxx"` explicitly.

sdk/include/xdk/extypes.h
RatinCN:
```C
#if (NTDDI_VERSION >= NTDDI_LONGHORN)
typedef struct _SYSTEM_BOOT_ENVIRONMENT_INFORMATION
{
    GUID BootIdentifier;
    FIRMWARE_TYPE FirmwareType;
#if (NTDDI_VERSION >= NTDDI_WIN8)
    ULONGLONG BootFlags;
#endif
} SYSTEM_BOOT_ENVIRONMENT_INFORMATION, *PSYSTEM_BOOT_ENVIRONMENT_INFORMATION;
#endif
```

`FIRMWARE_TYPE` is available since NT 6.0 in KM (wdm.h), and NT 6.2 in UM (winnt.h), so how about:
```C
$if (_WDMDDK_)
#if (NTDDI_VERSION >= NTDDI_VISTA)
$endif (_WDMDDK_)
$if (_WINNT_)
#if (NTDDI_VERSION >= NTDDI_WIN8)
$endif (_WINNT_)
$if (_WINNT_ || _WDMDDK_)
typedef enum _FIRMWARE_TYPE {
    FirmwareTypeUnknown,
    FirmwareTypeBios,
    FirmwareTypeUefi,
    FirmwareTypeMax
} FIRMWARE_TYPE, *PFIRMWARE_TYPE;
#endif /* (NTDDI_VERSION >= NTDDI_VISTA) or (NTDDI_VERSION >= NTDDI_WIN8) */
$endif (_WINNT_ || _WDMDDK_)
```
Or no `#if` guarded like Windows SDK/WDK?

HBelusca: > FIRMWARE_TYPE is available since NT 6.0 in KM (wdm.h) [...]

ok I see. Because I actually checked for the existence of GetFirmwareType function instead in kernel32/kernelbase, and it wasn't there on windows 7.

modules/rostests/apitests/kernel32/UEFIFirmware.c
HBelusca: ah yes correct :+1:

sdk/include/xdk/extypes.h
RatinCN: Some NT6.0 build (xxx_vista) have `SYSTEM_BOOT_ENVIRONMENT_INFORMATION`, they also need `FIRMWARE_TYPE`. They should be guarded by `(NTDDI_VERSION >= NTDDI_VISTA)` or no guard together.

sdk/include/ndk/extypes.h
RatinCN: Committed, `SYSTEM_BOOT_ENVIRONMENT_V1` added after `SYSTEM_BOOT_ENVIRONMENT_INFORMATION` with `#if (NTDDI_VERSION >= NTDDI_WIN8)`

modules/rostests/apitests/kernel32/UEFIFirmware.c
RatinCN: Committed.

sdk/include/xdk/extypes.h
HBelusca: ok let's just keep it simple as it currently is, with the ntddi_version >= vista.

sdk/include/ndk/extypes.h
```diff
@@ -1532,6 +1554,17 @@ typedef struct _SYSTEM_MEMORY_LIST_INFORMATION
...
+//
+// Firmware variable attributes
+//
+#define VARIABLE_ATTRIBUTE_NON_VOLATILE                             0x00000001
+#define VARIABLE_ATTRIBUTE_BOOTSERVICE_ACCESS                       0x00000002
+#define VARIABLE_ATTRIBUTE_RUNTIME_ACCESS                           0x00000004
+#define VARIABLE_ATTRIBUTE_HARDWARE_ERROR_RECORD                    0x00000008
+#define VARIABLE_ATTRIBUTE_AUTHENTICATED_WRITE_ACCESS               0x00000010
+#define VARIABLE_ATTRIBUTE_TIME_BASED_AUTHENTICATED_WRITE_ACCESS    0x00000020
+#define VARIABLE_ATTRIBUTE_APPEND_WRITE                             0x00000040
```
HBelusca, line 1566: Those attributes are not exposed by the standard wdk in the MS DDK (or even in the PSDK)?
If they are exposed then they should be moved to the xdk template files.

RatinCN: No, they introduced in MSDoc only (and in UEFI Specification named EFI_VARIABLE_XXX), WDK/DDK didn't expose them.

dll/win32/kernel32/kernel32.spec
RatinCN: Well, I'll :
+ Move `GetFirmwareType` to kernel32_vista
+ **Copy** (Get/Set)FirmwareEnvironmentVariableEx[A/W] to kernel32_vista, because (Get/Set)FirmwareEnvironmentVariable call their Ex version internally, code will have redundance.

Is that okay? @HBelusca

modules/rostests/apitests/kernel32/UEFIFirmware.c
learn-more: If that is possible it would be preferred, because that means the test can also run if ReactOS is not NT >= 6.2, but has the function available anyway.

sdk/include/psdk/winbase.h
RatinCN: > Already understood because it's BOOL as return type, and it's not a function where FALSE means "success"...

Something I found:
`BOOL` is `int`, and has no builtin `_Success_` annotation like `NTSTATUS` and `HRESULT`.
```
typedef int                 BOOL;
```
MS also use `_Success_` to evaluate `BOOL` or `int` return value, see [Best practices and examples (SAL) - TRUE in _Success_](https://learn.microsoft.com/en-us/cpp/code-quality/best-practices-and-examples-sal?view=msvc-170#true-in-_success_):
>If the function succeeds when the return value is nonzero, use return != 0 as the success condition instead of return == TRUE

And some API in Windows SDK have `_Success_(return != FALSE)` annotation, like:
```
WINBASEAPI
_Success_(return != FALSE)
BOOL
WINAPI
GetSystemTimeAdjustment(
    _Out_ PDWORD lpTimeAdjustment,
    _Out_ PDWORD lpTimeIncrement,
    _Out_ PBOOL lpTimeAdjustmentDisabled
    );
```
But, seems MSVC compiler didn't concerned about `_Success_` annotation for a `int` return value, no warning on code analysis or compile-time if I ignore it:
```
DWORD dw1, dw2, dw3;
GetSystemTimeAdjustment(&dw1, &dw2, &dw3);
return dw1 + dw2 + dw3;
```
or
```
char sz[5];
sprintf_s(sz, _countof(sz), "xxxxxxxxxxxxxxxxxxxxxxx");
printf(sz);
```

However, `_Success_(return != FALSE)` is unnecessary for now.

dll/win32/kernel32/kernel32.spec
RatinCN: Seems hard to do so:

In `k32_vista.h`, `NTDDI_VERSION` is lower than `NTDDI_LONGHORN`:
```
/* Redefine NTDDI_VERSION to 2K3 SP1 to get correct NDK definitions */
#undef NTDDI_VERSION
#define NTDDI_VERSION NTDDI_WS03SP1
```
So kernel32_vista cannot has `SYSTEM_BOOT_ENVIRONMENT_INFORMATION` definition, that's `GetFirmwareType` required. Define it again in kernel32_vista seems is not a good programming practice.

And also, re-implement the "Ex" version in kernel32_vista, code will have a big redundance, I prefer to give up exporting them and keep them in kernel32 as private functions for now.

BTW, those UM API is a part of UEFI, that's NT6.2+ stuff. If we support UEFI, could we export them whatever NT >= 6.2 or not?

DarkFire01: If it were me, I'd create a new k32_vista (k32_win8 lol?) header that only is included in the file you make to store these APIs in. 
But maybe that's not a great solution either to be honest as that implies quite a bit of redundancy for something a single thing.

But to answer your other question here from my opinion;
I don't think ROS will have a habit of using these anytime soon. There's not really any app one should be running in ROS anyway that needs to modify the firmware variables. So I don't think it's worth exporting no matter the NT version. 

However in the not to distant road these would eventually need to be used if we want to install on a modern system the "Correct way". Cross that bridge when we get there I'd say.

RatinCN: > 

I'm planned to remove NT6+ functions in this PR:
+ Remove `GetFirmwareType` and `_FIRMWARE_TYPE`
+ Do not export `(Get/Set)FirmwareEnvironmentVariableEx[A/W]` but keep as internal functions used by `(Get/Set)FirmwareEnvironmentVariable[A/W]`
+ Keep `_SYSTEM_BOOT_ENVIRONMENT_INFORMATION` and API tests for future

How about this plan? BTW, some commands to edit boot sequence, and "Boot to BIOS setup next time" feature will use them. For example, "shutdown /r **/fw**" is a real OS app call them to modify firmware variables, to change the boot sequence make machine enter BIOS setup next time. @DarkFire01

RatinCN:
NT6+ exports are removed now

modules/rostests/apitests/kernel32/UEFIFirmware.c
```diff
@@ -0,0 +1,253 @@
...
+
+    /* Enable SE_SYSTEM_ENVIRONMENT_NAME privilege required by the following tests */
+    bResult = OpenProcessToken(GetCurrentProcess(), TOKEN_ADJUST_PRIVILEGES, &hToken);
+    if (!bResult)
+    {
+        skip("OpenProcessToken failed with error: 0x%08lX, aborting.\n", GetLastError());
+        return;
+    }
+    Privilege.PrivilegeCount = 1;
+    Privilege.Privileges[0].Luid = (LUID){ SE_SYSTEM_ENVIRONMENT_PRIVILEGE, 0 };
```
HBelusca, line 132: Is the cast needed?

Note: otherwise, `Privilege.Privileges[0].Luid = RtlConvertLongToLuid(SE_SYSTEM_ENVIRONMENT_PRIVILEGE);`

HBelusca (approved):
LGTM otherwise

modules/rostests/apitests/kernel32/UEFIFirmware.c
RatinCN: Yes, the cast is needed, because here is not a variable declaration like `LUID Luid = { SE_SYSTEM_ENVIRONMENT_PRIVILEGE, 0 }`.
`RtlConvertLongToLuid` is better but seems it exists in XDK/DDK, for KM only.

GeoB99: Please avoid ugly casts like these as there is no reason for it, thanks.

You could simply use `RtlConvertUlongToLuid` as it is defined in NDK and since you are writing an apitest in Kernel32 it already imports the RTL functions from here -> https://git.reactos.org/?p=reactos.git;a=blob;f=modules/rostests/apitests/kernel32/precomp.h;hb=961893a712a0ba3a508b743aac2a769500397d0d#l16

GeoB99 (changes requested):

modules/rostests/apitests/kernel32/UEFIFirmware.c
RatinCN: Committed.

GeoB99 (approved):

binarymaster (approved):

PR #5150, opened 2023-03-11
[REGEDIT] Fix ListView selection and finding

katahiromz:
## Purpose

Follow up to #5146.

- Fix ListView Selection.
- Search a value correctly.

The `wcslen` call against non-`UNICODE_NULL`-terminated data had caused buffer overrun. We will check the data size correctly, instead of 3 `NUL` byte appending hack.

JIRA issue: [CORE-15986](https://jira.reactos.org/browse/CORE-15986), [CORE-18230](https://jira.reactos.org/browse/CORE-18230)

## Proposed changes

- Add `bSelectNone` parameter to `UpdateAddress` and `RefreshListView` functions.
- If `bSelectNone` is `TRUE`, then select nothing of ListView.
- Fix item selection of ListView.
- Rename `CompareData` helper function as `MatchData` and improve it.
- Improve the search algorithm.
- If the item selection of ListView changed, scroll down to the item.

## Screenshots

`ReportAs`:
after-ReportAs

`DisplayVersion`:
after-DisplayVersion

## TODO

- [x] Do tests.

### DIFF
--- a/base/applications/regedit/find.c
+++ b/base/applications/regedit/find.c
@@ -36,6 +36,18 @@ static DWORD s_cchName;
 static const WCHAR s_empty[] = L"";
 static const WCHAR s_backslash[] = L"\\";
 
+/* TODO: Hack. To be removed */
+#define wcslen(s) wcslen_hack(s, __LINE__)
+size_t wcslen_hack(LPCWSTR psz, INT line)
+{
+    {
+        WCHAR sz[256];
+        wsprintfW(sz, L"%d: %10s\n", line, psz);
+        OutputDebugStringW(sz);
+    }
+    return lstrlenW(psz);
+}
+
 extern VOID SetValueName(HWND hwndLV, LPCWSTR pszValueName);
 
 BOOL DoEvents(VOID)
@@ -91,11 +103,18 @@ static BOOL
 CompareData(
     DWORD   dwType,
     LPCWSTR psz1,
-    LPCWSTR psz2)
+    size_t cb1,
+    LPCWSTR psz2,
+    size_t cb2)
 {
-    INT i, cch1 = wcslen(psz1), cch2 = wcslen(psz2);
+    ssize_t i;
+    size_t cch1, cch2;
     if (dwType == REG_SZ || dwType == REG_EXPAND_SZ)
     {
+        StringCbLengthW(psz1, cb1 / sizeof(WCHAR), &cb1);
+        StringCbLengthW(psz2, cb2 / sizeof(WCHAR), &cb2);
+        cch1 = cb1 / sizeof(WCHAR);
+        cch2 = cb2 / sizeof(WCHAR);
         if (s_dwFlags & RSF_WHOLESTRING)
         {
             if (s_dwFlags & RSF_MATCHCASE)
@@ -106,6 +125,7 @@ CompareData(
                                           NORM_IGNORECASE, psz1, cch1, psz2, cch2);
         }
 
+        if (cch1 > cch2)
         for(i = 0; i <= cch1 - cch2; i++)
         {
             if (s_dwFlags & RSF_MATCHCASE)
@@ -193,14 +213,14 @@ BOOL RegFindRecurse(
     qsort(ppszNames, c, sizeof(LPWSTR), compare);
 
     if (pszValueName == NULL)
-        pszValueName = ppszNames[0];
+        fPast = TRUE;
 
     for(i = 0; i < c; i++)
     {
         if (DoEvents())
             goto err;
 
-        if (!fPast && _wcsicmp(ppszNames[i], pszValueName) == 0)
+        if (!fPast && pszValueName && _wcsicmp(ppszNames[i], pszValueName) == 0)
         {
             fPast = TRUE;
             continue;
@@ -237,7 +257,7 @@ BOOL RegFindRecurse(
         pb[cb] = pb[cb + 1] = pb[cb + 2] = 0;
 
         if ((s_dwFlags & RSF_LOOKATDATA) &&
-                CompareData(type, (LPWSTR) pb, s_szFindWhat))
+                CompareData(type, (LPWSTR) pb, cb, s_szFindWhat, wcslen(s_szFindWhat)))
         {
             *ppszFoundSubKey = _wcsdup(szSubKey);
             if (ppszNames[i][0] == 0)
--- a/base/applications/regedit/find.c
+++ b/base/applications/regedit/find.c
@@ -108,7 +108,7 @@ static BOOL MatchString(LPCWCH pch1, INT cch1, LPCWCH pch2, INT cch2)
     return FALSE;
 }
 
-static BOOL MatchData(DWORD dwType, LPCVOID pv1, size_t cb1)
+static BOOL MatchData(DWORD dwType, LPCVOID pv1, DWORD cb1)
 {
     if (dwType == REG_SZ || dwType == REG_EXPAND_SZ || dwType == REG_MULTI_SZ)
         return MatchString(pv1, (INT)(cb1 / sizeof(WCHAR)), s_szFindWhat, lstrlenW(s_szFindWhat));
--- a/base/applications/regedit/find.c
+++ b/base/applications/regedit/find.c
@@ -88,9 +88,9 @@ static BOOL CompareName(LPCWSTR pszName1, LPCWSTR pszName2)
 }
 
 /* We don't assume that pch1 is UNICODE_NULL-terminated */
-static BOOL MatchString(LPCWCH pch1, size_t cch1, LPCWCH pch2, size_t cch2)
+static BOOL MatchString(LPCWCH pch1, INT cch1, LPCWCH pch2, INT cch2)
 {
-    size_t i;
+    INT i;
     DWORD dwNorm = ((s_dwFlags & RSF_MATCHCASE) ? NORM_IGNORECASE : 0);
 
     if (s_dwFlags & RSF_WHOLESTRING)
@@ -111,7 +111,7 @@ static BOOL MatchString(LPCWCH pch1, size_t cch1, LPCWCH pch2, size_t cch2)
 static BOOL MatchData(DWORD dwType, LPCVOID pv1, size_t cb1)
 {
     if (dwType == REG_SZ || dwType == REG_EXPAND_SZ || dwType == REG_MULTI_SZ)
-        return MatchString(pv1, cb1 / sizeof(WCHAR), s_szFindWhat, wcslen(s_szFindWhat));
+        return MatchString(pv1, (INT)(cb1 / sizeof(WCHAR)), s_szFindWhat, lstrlenW(s_szFindWhat));
 
     return FALSE;
 }
--- a/base/applications/regedit/find.c
+++ b/base/applications/regedit/find.c
@@ -88,8 +88,7 @@ static BOOL CompareName(LPCWSTR pszName1, LPCWSTR pszName2)
 }
 
 /* We don't assume that pch1 is UNICODE_NULL-terminated */
-static BOOL
-MatchString(LPCWCH pch1, size_t cch1, LPCWCH pch2, size_t cch2)
+static BOOL MatchString(LPCWCH pch1, size_t cch1, LPCWCH pch2, size_t cch2)
 {
     size_t i;
     DWORD dwNorm = ((s_dwFlags & RSF_MATCHCASE) ? NORM_IGNORECASE : 0);
@@ -109,8 +108,7 @@ MatchString(LPCWCH pch1, size_t cch1, LPCWCH pch2, size_t cch2)
     return FALSE;
 }
 
-static BOOL
-MatchData(DWORD dwType, LPCVOID pv1, size_t cb1)
+static BOOL MatchData(DWORD dwType, LPCVOID pv1, size_t cb1)
 {
     if (dwType == REG_SZ || dwType == REG_EXPAND_SZ || dwType == REG_MULTI_SZ)
         return MatchString(pv1, cb1 / sizeof(WCHAR), s_szFindWhat, wcslen(s_szFindWhat));
--- a/base/applications/regedit/find.c
+++ b/base/applications/regedit/find.c
@@ -160,7 +160,7 @@ BOOL RegFindRecurse(
         goto err;
     ZeroMemory(ppszNames, c * sizeof(LPWSTR));
 
-    /* Get the value names of the current key */
+    /* Retrieve the value names associated with the current key */
     for(i = 0; i < c; i++)
     {
         if (DoEvents())
@@ -184,10 +184,10 @@ BOOL RegFindRecurse(
 
     qsort(ppszNames, c, sizeof(LPWSTR), compare);
 
-    /* If pszValueName is not specified, then search the values entirely in the key */
+    /* If pszValueName is NULL, the function will search for all values within the key */
     fPast = (pszValueName == NULL);
 
-    /* Search in the value entries */
+    /* Search within the values */
     for (i = 0; i < c; i++)
     {
         if (DoEvents())
@@ -239,7 +239,7 @@ BOOL RegFindRecurse(
     }
     ppszNames = NULL;
 
-    /* Get the count of the sub-keys */
+    /* Retrieve the number of sub-keys */
     lResult = RegQueryInfoKeyW(hSubKey, NULL, NULL, NULL, &c, NULL, NULL,
                               NULL, NULL, NULL, NULL, NULL);
     if (lResult != ERROR_SUCCESS)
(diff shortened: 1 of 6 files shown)


base/applications/regedit/find.c
```diff
@@ -223,16 +231,19 @@ BOOL RegFindRecurse(
                                   NULL, &cb);
         if (lResult != ERROR_SUCCESS)
             goto err;
-        pb = malloc(cb);
+        pb = malloc(cb + 3); /* To avoid buffer overrun, append 3 NULs */
```
Doug-Lyons, line 234: You should only need to add two (2) more bytes so that you can put a UNICODE NULL at the end.

base/applications/regedit/find.c
```diff
@@ -223,16 +231,19 @@ BOOL RegFindRecurse(
...
         if (pb == NULL)
             goto err;
         lResult = RegQueryValueExW(hSubKey, ppszNames[i], NULL, &type,
                                   pb, &cb);
         if (lResult != ERROR_SUCCESS)
             goto err;
 
+        /* To avoid buffer overrun, append 3 NULs */
+        pb[cb] = pb[cb + 1] = pb[cb + 2] = 0;
```
Doug-Lyons, line 243: Suggestion if you follow above comment: pb[cb] = pb[cb + 1] = 0;

Note: It is interesting that I have been working on this exact same area of the code. I have found that there are some values in the registry that are not NULL terminated and this causes crashes. With this approach, it should be fixed.

Doug-Lyons (commented):
Thank you for looking into this. I have been investigating crashes in regedit as well and came to the same conclusion to add a UNICODE NULL at the end of all of the values read this way.

base/applications/regedit/find.c
```diff
@@ -91,11 +91,18 @@ static BOOL
...
     DWORD   dwType,
     LPCWSTR psz1,
-    LPCWSTR psz2)
+    size_t  cb1,
+    LPCWSTR psz2,
+    size_t  cb2)
 {
-    INT i, cch1 = wcslen(psz1), cch2 = wcslen(psz2);
+    ssize_t i;
+    size_t cch1, cch2;
```
Doug-Lyons, line 99: Could you please consider adding this statement here. These types of keys should be excluded.

   if (dwType == REG_BINARY) return FALSE;

katahiromz: Rejected. Our regedit should search the binary data by string.

Doug-Lyons: Do you know if W2K3 can handle this data in its search?

base/applications/regedit/find.c
katahiromz: `cb` can be an odd number. 2 bytes is not enough.

base/applications/regedit/find.c
katahiromz: `cb` can be an odd number. 2 bytes is not enough.

base/applications/regedit/find.c
katahiromz: 2k3canfind
Win2k3 can find the binary data by string.

Doug-Lyons: Since you are using by definition LPCWSTR for psz1 and psz2, there will not be a case of these types of variables where the BYTE count will be odd. I have seen the byte count from a REG_BINARY data be odd. If you intend to handle the REG_BINARY type, then I believe that you will need quite a bit more code. You may want to make this another PR?

katahiromz: VirtualBox_Windows Server 2003_11_03_2023_19_55_05

Win2k3 **cannot** find the binary data by string.

katahiromz: unicode
Win2k3 *cannot* search Unicode text of the binary data.

katahiromz: Okay, your suggestion is accepted. We prohibit to search in binary data by string.

katahiromz: `(dwType == REG_SZ || dwType == REG_EXPAND_SZ)`
is enough condition.

katahiromz: > definition LPCWSTR for psz1 and psz2

`CompareData` parameters are not guaranteed to follow the definition of `LPCWSTR`.

katahiromz: Our registry can be broken, so we need strict checking...

katahiromz: > I believe that you will need quite a bit more code

Now simplified by deleting data size paramters.

learn-more: It is perfectly valid for `RegQueryValue` to return a non-null terminated string.

HBelusca: > `(dwType == REG_SZ || dwType == REG_EXPAND_SZ)` is enough condition.

And what about `REG_MULTI_SZ` ?

base/applications/regedit/find.c
```diff
@@ -223,16 +231,23 @@ BOOL RegFindRecurse(
...
         if (pb == NULL)
             goto err;
         lResult = RegQueryValueExW(hSubKey, ppszNames[i], NULL, &type,
                                   pb, &cb);
         if (lResult != ERROR_SUCCESS)
             goto err;
 
+        /* To avoid buffer overrun, append 3 NUL bytes.
+           NOTE: cb can be an odd number although UNICODE_NULL is two bytes.
+                 Two bytes at odd position is not enough to avoid buffer overrun. */
```
HBelusca, line 244:
```suggestion
           Two bytes at odd position is not enough to avoid buffer overrun. */
```

base/applications/regedit/find.c
```diff
@@ -93,9 +93,16 @@ CompareData(
     LPCWSTR psz1,
     LPCWSTR psz2)
 {
-    INT i, cch1 = wcslen(psz1), cch2 = wcslen(psz2);
+    ssize_t i;
```
HBelusca, line 96: Using (unsigned) `size_t` here wouldn't work? Since you added the `cch1 >= cch2` test below, this should be OK I think.

katahiromz:
If we search `REG_MULTI_SZ`, then we need the data size parameters.

Doug-Lyons:
@katahiromz, Your last commit broke F3 in some cases. Start from the top of the treeview at "My Computer" and do a find for the single lower case letter "a". It will find "mspaint.exe". Pressing F3 should move down to "notepad.exe", but now the F3 key is ingored. Thanks for working on this. So far there have been many inprovements.

base/applications/regedit/find.c
ThFabba: I believe `RegQueryValueExW` will null-terminate the string for you, even if it's stored without a null-terminator in the registry. So checking for type seems reasonably safe.
Multi-SZ and binary would need special handling if we want to support them, since wcsstr won't do the trick. I'm in favor of supporting Multi-SZ, but wouldn't make it a requirement here, since the functionality is already broken anyway.

HBelusca: https://learn.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regqueryvalueexw
> To ensure that any string values (REG_SZ, REG_MULTI_SZ, and REG_EXPAND_SZ) returned are null-terminated, use the [RegGetValue](https://learn.microsoft.com/en-us/windows/desktop/api/winreg/nf-winreg-reggetvaluea) function.

However the RegGetValue function is Vista+.

Doug-Lyons: The RegQueryValueExW is not without its problems though. See
https://devblogs.microsoft.com/oldnewthing/20040824-00/?p=38063

Doug-Lyons: Regarding RegGetValue I see "Minimum supported server | Windows Server 2008, Windows Server 2003 with SP1"

ThFabba: Huh I guess I misremembered. RegQueryValueExW is still fine -- just either null-terminate the buffer yourself or pass the size to all callees and use strsafe functions.

base/applications/regedit/find.c
katahiromz: Fixed.

katahiromz:
Let's go.

base/applications/regedit/find.c
tkreuzer: If cb is an odd number, there is probably something wrong. Maybe ASSERT this after the type check?

base/applications/regedit/find.c
```diff
@@ -38,6 +38,18 @@ static const WCHAR s_backslash[] = L"\\";
...
+void DebugPrintf(LPCSTR fmt, ...)
+{
+    char buf[512];
+    va_list va;
+    va_start(va, fmt);
+    vsprintf(buf, fmt, va);
+    OutputDebugStringA(buf);
+    va_end(va);
+}
+#endif
```
tkreuzer, line 51: Clean up?

base/applications/regedit/framewnd.c
```diff
@@ -877,15 +877,20 @@ static BOOL CreateNewValue(HKEY hRootKey, LPCWSTR pszKeyPath, DWORD dwType)
...
     /* locate the newly added value, and get ready to rename it */
     memset(&lvfi, 0, sizeof(lvfi));
     lvfi.flags = LVFI_STRING;
     lvfi.psz = szNewValue;
     iIndex = ListView_FindItem(g_pChildWnd->hListWnd, -1, &lvfi);
     if (iIndex >= 0)
+    {
+        ListView_SetItemState(g_pChildWnd->hListWnd, iIndex,
+                              LVIS_SELECTED | LVIS_FOCUSED, LVIS_SELECTED | LVIS_FOCUSED);
+        ListView_EnsureVisible(g_pChildWnd->hListWnd, iIndex, FALSE);
```
tkreuzer, line 891: What does this change do?

base/applications/regedit/find.c
```diff
@@ -663,7 +687,10 @@ BOOL FindNext(HWND hWnd)
     }
     s_bAbort = FALSE;
 
-    pszValueName = GetValueName(g_pChildWnd->hListWnd, -1);
+    if (ListView_GetNextItem(g_pChildWnd->hListWnd, -1, LVNI_SELECTED) == 0)
+        pszValueName = L""; /* The default name */
+    else
+        pszValueName = GetValueName(g_pChildWnd->hListWnd, -1);
```
tkreuzer, line 693: Why is this needed?

tkreuzer (commented):
Looks generally ok, but can you please clean up the commits and separate the buffer overflow fixes from the other changes?

base/applications/regedit/find.c
```diff
@@ -90,11 +102,15 @@ static BOOL CompareName(LPCWSTR pszName1, LPCWSTR pszName2)
...
-    LPCWSTR psz2)
+    LPCWSTR psz1, size_t cch1,
+    LPCWSTR psz2, size_t cch2)
 {
-    INT i, cch1 = wcslen(psz1), cch2 = wcslen(psz2);
-    if (dwType == REG_SZ || dwType == REG_EXPAND_SZ)
+    size_t i;
+
+    if (dwType == REG_BINARY)
+        return FALSE; /* We don't search the binary data by string */
```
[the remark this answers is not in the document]
katahiromz: Deleted.

base/applications/regedit/find.c
```diff
@@ -106,6 +122,7 @@ CompareData(
                                           NORM_IGNORECASE, psz1, cch1, psz2, cch2);
         }
 
+        if (cch1 >= cch2)
```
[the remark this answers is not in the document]
katahiromz: OK. Fixed.

base/applications/regedit/find.c
katahiromz: Deleted.

base/applications/regedit/framewnd.c
katahiromz: Automatically scrolling down to the `iIndex`-th item.

base/applications/regedit/find.c
katahiromz: Reverted.

Doug-Lyons:
> Looks generally ok, but can you please clean up the commits and separate the buffer overflow fixes from the other changes?

@tkreuzer, Based on your request, I have modified my previous PR to only fix the buffer overflows here:
https://github.com/reactos/reactos/pull/5146

Now it should be easy to review and possible can get quickly approved independent of @katahiromz other fixes here. Thanks.

Doug-Lyons:
@katahiromz, Can you review my revised https://github.com/reactos/reactos/pull/5146 which is just the buffer read past end changes and a possible check for an odd number of bytes for a wide string? It would be great if you could offer your approval on it. Thanks.

katahiromz:
Let's go.

base/applications/regedit/find.c
```diff
@@ -97,7 +97,7 @@ CompareData(DWORD dwType, LPCVOID pv1, size_t cb1)
         size_t cch1 = cb1 / sizeof(WCHAR), cch2 = wcslen(s_szFindWhat), i;
 
         if (cb1 & 1)
-            DPRINT1("Odd size for string (%lu)!\n", dwType);
```
HBelusca: Most importantly it would be of interest to indicate the name of the corresponding registry **value** (and its path in the registry) so that we can identify these.

base/applications/regedit/find.c
```diff
@@ -88,14 +88,22 @@ static BOOL CompareName(LPCWSTR pszName1, LPCWSTR pszName2)
...
 {
-    INT i, cch1 = wcslen(psz1), cch2 = wcslen(psz2);
-    if (dwType == REG_SZ || dwType == REG_EXPAND_SZ)
+    if (dwType == REG_SZ || dwType == REG_EXPAND_SZ || dwType == REG_MULTI_SZ)
     {
+        LPCWSTR psz1 = pv1; /* We assume pv1 is UNICODE_NULL-terminated */
+        LPCWSTR psz2 = s_szFindWhat;
+        size_t cch1 = cb1 / sizeof(WCHAR), cch2 = wcslen(s_szFindWhat), i;
+
+#ifdef _DEBUG
```
ThFabba, line 99: Regedit is not the right place for this. Ignore the extra byte and move on.

base/applications/regedit/find.c
```diff
@@ -88,14 +88,22 @@ static BOOL CompareName(LPCWSTR pszName1, LPCWSTR pszName2)
...
 static BOOL
-CompareData(
-    DWORD   dwType,
-    LPCWSTR psz1,
-    LPCWSTR psz2)
+CompareData(DWORD dwType, LPCVOID pv1, size_t cb1)
 {
-    INT i, cch1 = wcslen(psz1), cch2 = wcslen(psz2);
-    if (dwType == REG_SZ || dwType == REG_EXPAND_SZ)
+    if (dwType == REG_SZ || dwType == REG_EXPAND_SZ || dwType == REG_MULTI_SZ)
```
ThFabba, line 93: Same comment as in the first PR: please don't add Multi-SZ unless you _actually_ support it. That would entail (1) ensuring a _double_ null-terminator, and (2) looping through all strings instead of just the first one.
Finding things in only the first element of multi-sz strings is inconsistent and confusing for the user.

base/applications/regedit/find.c
JoachimHenze: May be ok as a **temporary** debug thing. Feels a bit like abusing regedit as a debugger though.
But please at least without trailing ! then. :-)

base/applications/regedit/find.c
katahiromz: Cc @tkreuzer

base/applications/regedit/find.c
HBelusca: In which case, this has to be done only by the interested person, in his **own** local branch/repo, in order for him to investigate what is adding said odd-counted unicode string, and fix (with a PR) that underlying problem.

base/applications/regedit/find.c
```diff
@@ -88,14 +88,25 @@ static BOOL CompareName(LPCWSTR pszName1, LPCWSTR pszName2)
...
-    DWORD   dwType,
-    LPCWSTR psz1,
-    LPCWSTR psz2)
+CompareData(LPCWSTR pszSubKey, LPCWSTR pszValueName, DWORD dwType, LPCVOID pv1, size_t cb1)
 {
-    INT i, cch1 = wcslen(psz1), cch2 = wcslen(psz2);
-    if (dwType == REG_SZ || dwType == REG_EXPAND_SZ)
+    if (dwType == REG_SZ || dwType == REG_EXPAND_SZ || dwType == REG_MULTI_SZ)
     {
+        LPCWSTR psz1 = pv1; /* We assume pv1 is UNICODE_NULL-terminated */
```
ThFabba, line 95: If you can assume that, why are you passing `cb1` to the function?

base/applications/regedit/find.c
katahiromz: We can actually support `REG_MULTI_SZ` in this PR.
This function searches the substring beyond `UNICODE_NULL`.

base/applications/regedit/find.c
katahiromz: Assumption is removed.

base/applications/regedit/find.c
katahiromz: Removed.

base/applications/regedit/find.c
katahiromz: Removed.

katahiromz:
Time up. It should be good. Force.

PR #5151, opened 2023-03-11
[NOTEPAD] Use _CrtSetDbgFlag to check memory leak

katahiromz:
## Purpose
We can borrow the power of CRT debug.

## Proposed changes

These changes are effective for debug version only.
- Insert `#include <crtdbg.h>` at `main.c`.
- Call `_CrtSetDbgFlag(_CRTDBG_ALLOC_MEM_DF | _CRTDBG_LEAK_CHECK_DF)` at the prologue of `_tWinMain`.

## TODO

- [x] Do tests.

## See also

- `_CrtSetDbgFlag` function: https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/crtsetdbgflag?view=msvc-170

### DIFF
--- a/base/applications/notepad/main.c
+++ b/base/applications/notepad/main.c
@@ -661,7 +661,7 @@ int WINAPI _tWinMain(HINSTANCE hInstance, HINSTANCE prev, LPTSTR cmdline, int sh
     DestroyAcceleratorTable(hAccel);
 
 #if defined(_MSC_VER) && !defined(NDEBUG)
-    // For detecting memory leak (MSVC only)
+    /* For detecting memory leak (MSVC only) */
     _CrtSetDbgFlag(_CRTDBG_ALLOC_MEM_DF | _CRTDBG_LEAK_CHECK_DF);
 #endif
 
--- a/base/applications/notepad/main.c
+++ b/base/applications/notepad/main.c
@@ -572,8 +572,8 @@ int WINAPI _tWinMain(HINSTANCE hInstance, HINSTANCE prev, LPTSTR cmdline, int sh
     static const TCHAR className[] = _T("Notepad");
     static const TCHAR winName[] = _T("Notepad");
 
-#if defined(_MSC_VER) && defined(_DEBUG)
-    /* Report any memory leaks at exit */
+#ifdef _DEBUG
+    /* Report any memory leaks on exit */
     _CrtSetDbgFlag(_CRTDBG_ALLOC_MEM_DF | _CRTDBG_LEAK_CHECK_DF);
 #endif
 
--- a/base/applications/notepad/main.c
+++ b/base/applications/notepad/main.c
@@ -27,7 +27,7 @@
 #include <shlobj.h>
 #include <strsafe.h>
 
-#if defined(_MSC_VER) && !defined(NDEBUG)
+#if defined(_MSC_VER) && defined(_DEBUG)
     #include <crtdbg.h> /* For _CrtSetDbgFlag (MSVC only) */
 #endif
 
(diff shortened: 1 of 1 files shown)


base/applications/notepad/main.c
```diff
@@ -656,5 +660,10 @@ int WINAPI _tWinMain(HINSTANCE hInstance, HINSTANCE prev, LPTSTR cmdline, int sh
 
     DestroyAcceleratorTable(hAccel);
 
-    return (int) msg.wParam;
+#if defined(_MSC_VER) && !defined(NDEBUG)
+    // For detecting memory leak (MSVC only)
```
HBelusca:
```suggestion
    /* Report any memory leaks */
```

base/applications/notepad/main.c
```diff
@@ -27,6 +27,10 @@
 #include <shlobj.h>
 #include <strsafe.h>
 
+#if defined(_MSC_VER) && !defined(NDEBUG)
```
HBelusca, line 30: NDEBUG is mostly for "debug output" . I think you need to use `_DEBUG` instead?

katahiromz: Fixed.

base/applications/notepad/main.c
katahiromz: Fixed.

base/applications/notepad/main.c
```diff
@@ -656,5 +660,10 @@ int WINAPI _tWinMain(HINSTANCE hInstance, HINSTANCE prev, LPTSTR cmdline, int sh
 
     DestroyAcceleratorTable(hAccel);
 
-    return (int) msg.wParam;
+#if defined(_MSC_VER) && defined(_DEBUG)
+    /* Report any memory leaks */
+    _CrtSetDbgFlag(_CRTDBG_ALLOC_MEM_DF | _CRTDBG_LEAK_CHECK_DF);
```
learn-more, line 665: By just setting it here you omit the error reporting on all failure cases (f.e. `HandleCommandLine` returning FALSE)

katahiromz: Added `NOTEPAD_Main` and fixed the problem you mensioned.

learn-more: It would have been easier to just call that function at the start of winmain?

katahiromz: OK.

base/applications/notepad/main.c
```diff
@@ -27,6 +27,10 @@
 #include <shlobj.h>
 #include <strsafe.h>
 
+#if defined(_MSC_VER) && defined(_DEBUG)
+    #include <crtdbg.h> /* For _CrtSetDbgFlag (MSVC only) */
+#endif
```
HBelusca, line 32:
```suggestion
#ifdef _DEBUG
#include <crtdbg.h>
#endif
```

base/applications/notepad/main.c
```diff
@@ -565,10 +569,14 @@ int WINAPI _tWinMain(HINSTANCE hInstance, HINSTANCE prev, LPTSTR cmdline, int sh
...
     INT x, y;
     RECT rcIntersect;
-
     static const TCHAR className[] = _T("Notepad");
     static const TCHAR winName[] = _T("Notepad");
 
+#if defined(_MSC_VER) && defined(_DEBUG)
+    /* Report any memory leaks at exit */
+    _CrtSetDbgFlag(_CRTDBG_ALLOC_MEM_DF | _CRTDBG_LEAK_CHECK_DF);
+#endif
```
HBelusca, line 578:
```suggestion
#ifdef _DEBUG
    /* Report any memory leaks at exit */
    _CrtSetDbgFlag(_CRTDBG_ALLOC_MEM_DF | _CRTDBG_LEAK_CHECK_DF);
#endif
```

HBelusca: Shouldn't we say "on exit" instead in english?

HBelusca (commented):
_CrtSetDbgFlag and the rest of the functionality is something implemented in the CRT, so in principle it's not really compiler-dependent.
If the program is linked dynamically to the CRT dlls, then the feature should be there, whether you compiled with GCC or MSVC, as long as the CRT headers being used have the correct prototypes for these functions (and the import libs too).
If the program is linked statically with the CRT, well since this is a ROS app, use ROS's (or MSVC) CRT and you get that feature as well.

HBelusca (approved):

PR #5152, opened 2023-03-12
[NOTEPAD] Use full path for non-existent file

katahiromz:
## Purpose

JIRA issue: [CORE-18259](https://jira.reactos.org/browse/CORE-18259), [CORE-18837](https://jira.reactos.org/browse/CORE-18837)

## Proposed changes

- In `HandleCommandLine` function, use `GetFullPathName` API to get the full path of a non-existent file.

## TODO

- [x] Do tests.


KRosUser2 (approved):
tested OK on original failure scenario from CORE-18259

PR #5153, opened 2023-03-12
[GITHUB] Automatically add label TRANSLATION

Artem13327:
Written using Google Translate
Automatically adds the TRANSLATION label to the Pull Request if any *.rc file in the lang folder is changed there. Will not work for strings that are translated in inf files

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -16,4 +16,3 @@ ROSTESTS:
 
 TRANSLATION:
   - "***/lang/*.rc"
-  


GeoB99:
> However, this can lead to a problem when adding a new application, as localization files will be added there.

This shouldn't be of a problem because writing new apps for the OS doesn't necessarily count as "translating files in new locale".

GeoB99 (approved):
LGTM, but please fix the merge conflicts before we can ship this PR to master.

PR #5156, opened 2023-03-12
[NTOS:SE] Fill the access check rights array with zeros

GeoB99:
GCC optimizes the code for the sake of performance purposes, by adding padding to structures so that the access of the member fields are memory aligned. However, this padding is not zero'd out which means we can leak critical kernel data into the stack to UM, which is the access check mask rights.

Probably MSVC doesn't zero out the padding either, if they ever add padding to begin with. Whatever that is, we have to zero out the padding ourselves by filling the access check rights array with zeros, as it fills the whole memory range that holds the array with zeros.

This bug was originally spotted by @learn-more, so kudos to him.
For more information -> https://research.nccgroup.com/2019/10/30/padding-the-struct-how-a-compiler-optimization-can-disclose-stack-memory/

### DIFF
--- a/ntoskrnl/se/accesschk.c
+++ b/ntoskrnl/se/accesschk.c
@@ -394,12 +394,15 @@ SepAccessCheck(
     BOOLEAN Defaulted;
     NTSTATUS Status;
     PACCESS_TOKEN Token = NULL;
-    ACCESS_CHECK_RIGHTS AccessCheckRights = {0};
+    ACCESS_CHECK_RIGHTS AccessCheckRights;
 
     PAGED_CODE();
 
     /* A security descriptor must be expected for access checks */
     ASSERT(SecurityDescriptor);
+
+    /* Fill the whole access rights array with zeros */
+    RtlZeroMemory(&AccessCheckRights, sizeof(ACCESS_CHECK_RIGHTS));
 
     /* Check for no access desired */
     if (!DesiredAccess)


ntoskrnl/se/accesschk.c
```diff
@@ -394,13 +394,16 @@ SepAccessCheck(
...
-    ACCESS_CHECK_RIGHTS AccessCheckRights = {0};
+    ACCESS_CHECK_RIGHTS AccessCheckRights;
 
     PAGED_CODE();
 
     /* A security descriptor must be expected for access checks */
     ASSERT(SecurityDescriptor);
 
+    /* Fill the whole access rights array with zeros */
+    RtlZeroMemory(&AccessCheckRights, sizeof(ACCESS_CHECK_RIGHTS));
```
HBelusca, line 405:
```suggestion
    RtlZeroMemory(&AccessCheckRights, sizeof(AccessCheckRights));
```

ntoskrnl/se/accesschk.c
```diff
@@ -394,13 +394,16 @@ SepAccessCheck(
...
     PACCESS_TOKEN Token = NULL;
-    ACCESS_CHECK_RIGHTS AccessCheckRights = {0};
+    ACCESS_CHECK_RIGHTS AccessCheckRights;
 
     PAGED_CODE();
 
     /* A security descriptor must be expected for access checks */
     ASSERT(SecurityDescriptor);
 
+    /* Fill the whole access rights array with zeros */
```
HBelusca, line 404: Not sure the comment is necessary, the code is already self-explainable.
But again, the structure is local to that function and used only within it, so the story of security and leaks is not that relevant here.
It would however be relevant, if somehow you then did a copy of the contents of this structure to an external buffer given as parameter to that SepAccessCheck function.

ThFabba: Agreed, this isn't useful for this particular variable. If we want to preemptively zero stack space, we can do it, but we should let the compiler do it, no pollute our code: https://msrc.microsoft.com/blog/2020/05/solving-uninitialized-stack-memory-on-windows/

ThFabba:
> Probably MSVC doesn't zero out the padding either, if they ever add padding to begin with.

FWIW, padding is part of the ABI, which means it's consistent between all compilers we use (if it's not, we either have a bug or are using the wrong compiler options).

PR #5158, opened 2023-03-14
[MSPAINT] Restore the main window saved show state

katahiromz:
## Purpose
Our `mspaint` didn't remember the maximized status of the main window.

## Proposed changes

- Initialize `RegistrySettings::WindowPlacement` correctly.
- Fix `mainWindowPos` values.
- Add `nCmdShow` parameter to `RegistrySettings::Load` and `RegistrySettings::LoadPresets`.
- Use `registrySettings.WindowPlacement.showCmd` for `ShowWindow` call.

## TODO

- [x] Do tests.

### DIFF
--- a/base/applications/mspaint/main.cpp
+++ b/base/applications/mspaint/main.cpp
@@ -143,7 +143,7 @@ OFNHookProc(HWND hwnd, UINT uMsg, WPARAM wParam, LPARAM lParam)
 /* entry point */
 
 int WINAPI
-_tWinMain (HINSTANCE hThisInstance, HINSTANCE hPrevInstance, LPTSTR lpszArgument, int nFunsterStil)
+_tWinMain (HINSTANCE hThisInstance, HINSTANCE hPrevInstance, LPTSTR lpszArgument, INT nShowCmd)
 {
     HWND hwnd;               /* This is the handle for our window */
     MSG messages;            /* Here messages to the application are saved */
--- a/base/applications/mspaint/main.cpp
+++ b/base/applications/mspaint/main.cpp
@@ -194,7 +194,9 @@ _tWinMain (HINSTANCE hThisInstance, HINSTANCE hPrevInstance, LPTSTR lpszArgument
     imageModel.Crop(registrySettings.BMPWidth, registrySettings.BMPHeight);
 
     /* create main window */
-    RECT mainWindowPos = {0, 0, 544, 375};	// FIXME: use equivalent of CW_USEDEFAULT for position
+    RECT mainWindowPos = { CW_USEDEFAULT, CW_USEDEFAULT };
+    mainWindowPos.right = mainWindowPos.left + 544;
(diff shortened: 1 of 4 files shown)


base/applications/mspaint/main.cpp
```diff
@@ -340,7 +340,7 @@ _tWinMain (HINSTANCE hThisInstance, HINSTANCE hPrevInstance, LPTSTR lpszArgument
     mainWindow.SetWindowPlacement(&(registrySettings.WindowPlacement));
 
     /* Make the window visible on the screen */
-    ShowWindow (hwnd, nFunsterStil);
```
HBelusca, line 343: You may replace as well "nFunsterStil" in the _tWinMain prototype above:
https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=nFunsterStil
by `nShowCmd`  (standard name).

HBelusca (approved):
According to the fix, the actual title of this PR (+ associated commit) would be: "Restore the main window saved show state"

base/applications/mspaint/main.cpp
katahiromz: Renamed as `nCmdShow`.

base/applications/mspaint/main.cpp
```diff
@@ -337,10 +339,19 @@ _tWinMain (HINSTANCE hThisInstance, HINSTANCE hPrevInstance, LPTSTR lpszArgument
...
+    {
+        RECT rc = registrySettings.WindowPlacement.rcNormalPosition;
+        mainWindow.MoveWindow(rc.left, rc.top,
+                              rc.right - rc.left, rc.bottom - rc.top,
+                              TRUE);
+    }
 
     /* Make the window visible on the screen */
-    ShowWindow (hwnd, nFunsterStil);
+    if (registrySettings.WindowPlacement.showCmd == SW_MAXIMIZE)
```
HBelusca, line 351: I know for a fact, that this code (with all those ifs etc.) might be simplified.
For example:
- intializing to defaults the position: https://git.reactos.org/?p=reactos.git;a=blob;f=base/applications/mscutils/eventvwr/eventvwr.c;hb=de892d5bc715d68af757a36173e9f99e7927115f#l361
(and later, getting them from registry);
- next, creating the main window already at its correct position:
https://git.reactos.org/?p=reactos.git;a=blob;f=base/applications/mscutils/eventvwr/eventvwr.c;hb=de892d5bc715d68af757a36173e9f99e7927115f#l3041
- then just calling ShowWindow with the correct parameter:
https://git.reactos.org/?p=reactos.git;a=blob;f=base/applications/mscutils/eventvwr/eventvwr.c;hb=de892d5bc715d68af757a36173e9f99e7927115f#l3261

katahiromz: Fixed.

PR #5163, opened 2023-03-15
[MSPAINT] Refactor about Recent Files

katahiromz:
## Purpose
Improve code flexibility.

## Proposed changes

- Remove `strFile1`, ..., `strFile4` settings and add `strFiles[4]` instead of it for Most Recently Used (MRU) files.
- Use for-loops to process the MRU files.
- Add `CMainWindow::ProcessFileMenu` helper function.

## TODO

- [x] Do tests.

### DIFF
--- a/base/applications/mspaint/registry.cpp
+++ b/base/applications/mspaint/registry.cpp
@@ -110,10 +110,12 @@ void RegistrySettings::Load()
     CRegKey files;
     if (files.Open(HKEY_CURRENT_USER, _T("Software\\Microsoft\\Windows\\CurrentVersion\\Applets\\Paint\\Recent File List"), KEY_READ) == ERROR_SUCCESS)
     {
-        ReadString(files, _T("File1"), strFile1);
-        ReadString(files, _T("File2"), strFile2);
-        ReadString(files, _T("File3"), strFile3);
-        ReadString(files, _T("File4"), strFile4);
+        TCHAR szName[64];
+        for (INT i = 0; i < MAX_RECENT_FILES; ++i)
+        {
+            wsprintf(szName, _T("File%u"), i + 1);
+            ReadString(files, szName, strFiles[i]);
+        }
     }
 
     CRegKey text;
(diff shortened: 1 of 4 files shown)


base/applications/mspaint/winproc.cpp
```diff
@@ -275,53 +275,59 @@ LRESULT CMainWindow::OnClose(UINT nMsg, WPARAM wParam, LPARAM lParam, BOOL& bHan
...
+    INT cItems = GetMenuItemCount(hPopupMenu);
+
+    for (INT iItem = 0; iItem < MAX_RECENT_FILES; ++iItem)
+    {
+        CString& strFile = registrySettings.strFiles[iItem];
+        if (strFile.IsEmpty())
+            break;
+
+        CPath pathFile(strFile);
+        pathFile.CompactPathEx(40);
```
JoachimHenze, line 310: A lot of magic numbers in here. Is 40 now, was 30 before. Why?
I do assume that the wsprintf at line 313 should not overwrite the szText buffer. Is that 100% sure currently?
Do we already have strsafe included here? If yes, then maybe we should use that insead.
If not included already, then maybe someone else who is better at counting digits than my brain can confirm that the wsprintf is guaranteed here to not overflow the buffer.

base/applications/mspaint/winproc.h
```diff
@@ -29,6 +29,7 @@ class CMainWindow : public CWindowImpl<CMainWindow>
         MESSAGE_HANDLER(WM_MOUSEWHEEL, OnMouseWheel)
     END_MSG_MAP()
 
+protected:
```
JoachimHenze, line 32: You decided to limit public visibility here. Generally I am neutral to that.
But why not use **private** then instead of **protected**?
Does anyone inherit from CMainWindow curently? I don't see any further inheritance atm. Or is anyone expected to do so in the future?

base/applications/mspaint/registry.cpp
```diff
@@ -189,39 +191,32 @@ void RegistrySettings::Store()
...
+    if (strFiles[0].CompareNoCase(szPathName) != 0)
     {
-        strFile4 = strFile3;
-        strFile3 = strFile2;
-        strFile2 = strFile1;
-        strFile1 = szPathName;
+        for (INT i = MAX_RECENT_FILES - 1; i > 0; --i)
+        {
+            strFiles[i] = strFiles[i - 1];
+        }
```
JoachimHenze, line 218: curly braces could be omitted here, allowing for more code to fit the screen vertically at once.

base/applications/mspaint/winproc.h
katahiromz: `s/protected:/private:/`
Fixed.

base/applications/mspaint/registry.cpp
```diff
@@ -189,39 +191,32 @@ void RegistrySettings::Store()
 
 void RegistrySettings::SetMostRecentFile(LPCTSTR szPathName)
 {
+    /* Register the file to the system 'Recent' folder */
```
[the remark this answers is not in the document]
katahiromz: Used C++ comments.

base/applications/mspaint/winproc.cpp
katahiromz: > Is 40 now, was 30 before. Why?

Reverted to 30.

> should not overwrite the szText buffer. Is that 100% sure currently?

Added an assertion.

> Do we already have strsafe included here?

No.

JoachimHenze (approved):

PR #5175, opened 2023-03-19
[NTOS:OB] Implement name information querying on ObjectInformation case

GeoB99:
Currently there is no implementation support for object name information querying when someone queries information about an object with `ObjectBasicInformation` class case. Some device drivers may take use of such information.

`NameInfoSize` is a size that is the sum of all name directories of an object, including path separators. `TypeInfoSize` is a size that is pointed by the size of a type object itself, therefore this size remains pretty much fixed depending on the kind of an object.

https://www.geoffchappell.com/studies/windows/km/ntoskrnl/inc/api/ntobapi/object_basic_information.htm
http://undocumented.ntinternals.net/index.html?page=UserMode%2FUndocumented%20Functions%2FNT%20Objects%2FType%20independed%2FNtQueryObject.html

### DIFF
--- a/modules/rostests/kmtests/ntos_ob/ObQuery.c
+++ b/modules/rostests/kmtests/ntos_ob/ObQuery.c
@@ -1,0 +1,71 @@
+/*
+ * PROJECT:     ReactOS kernel-mode tests
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Kernel mode tests for object information querying
+ * COPYRIGHT:   Copyright 2023 George Bișoc <george.bisoc@reactos.org>
+ */
+
+#include <kmt_test.h>
+
+#define OBJ_WINSTA_DIRECTORY_NAME_INFO_SIZE (sizeof(UNICODE_STRING) + sizeof(L"\\Windows"))
+#define OBJ_DIRECTORY_TYPE_INFO_SIZE (sizeof(OBJECT_TYPE_INFORMATION) + (wcslen(L"\\Directory") * sizeof(WCHAR)))
+
+static
+VOID
+ObjectBasicInformationTests(VOID)
+{
+    NTSTATUS Status;
+    HANDLE WinStaDirHandle;
+    OBJECT_BASIC_INFORMATION BasicInfo;
+    ULONG ReturnLength;
+    OBJECT_ATTRIBUTES ObjectAttributes;
+    static UNICODE_STRING WinStaDir = RTL_CONSTANT_STRING(L"\\Windows");
+
+    /* We must be in PASSIVE_LEVEL to do all of this stuff */
+    ok_irql(PASSIVE_LEVEL);
+
+    /* Create a path to \Windows directory */
+    InitializeObjectAttributes(&ObjectAttributes,
+                               &WinStaDir,
+                               OBJ_CASE_INSENSITIVE | OBJ_OPENIF | OBJ_KERNEL_HANDLE,
+                               NULL,
+                               NULL);
+    Status = ZwOpenDirectoryObject(&WinStaDirHandle,
+                                   DIRECTORY_QUERY | DIRECTORY_TRAVERSE,
+                                   &ObjectAttributes);
+    if (!NT_SUCCESS(Status))
+    {
+        ok(FALSE, "Failed to open \\Windows directory (Status 0x%lx)\n", Status);
+        return;
+    }
+
+    /* Give 0 as information length, this must fail */
+    Status = ZwQueryObject(WinStaDirHandle,
+                           ObjectBasicInformation,
+                           &BasicInfo,
+                           0,
+                           &ReturnLength);
+    ok_eq_hex(Status, STATUS_INFO_LENGTH_MISMATCH);
+
+    /* Do a proper query now */
+    Status = ZwQueryObject(WinStaDirHandle,
+                           ObjectBasicInformation,
+                           &BasicInfo,
+                           sizeof(BasicInfo),
+                           &ReturnLength);
+    ok_eq_hex(Status, STATUS_SUCCESS);
+
+    /* \Windows is currently used */
+    ok(BasicInfo.HandleCount != 0, "\\Windows is in use but HandleCount is 0!\n");
+    ok(BasicInfo.PointerCount != 0, "\\Windows is in use but PointerCount is 0!\n");
+
+    ok_eq_ulong(BasicInfo.NameInfoSize, OBJ_WINSTA_DIRECTORY_NAME_INFO_SIZE);
+    ok_eq_ulong(BasicInfo.TypeInfoSize, OBJ_DIRECTORY_TYPE_INFO_SIZE);
+
+    ZwClose(WinStaDirHandle);
+}
+
+START_TEST(ObQuery)
+{
+    ObjectBasicInformationTests();
+}
--- a/ntoskrnl/ob/oblife.c
+++ b/ntoskrnl/ob/oblife.c
@@ -869,6 +869,79 @@ ObpAllocateObject(IN POBJECT_CREATE_INFORMATION ObjectCreateInfo,
     return STATUS_SUCCESS;
 }
 
+/**
+ * @brief
+ * Queries the name info size of a given resource object.
+ * The function loops through all the parent directories
+ * of the object and computes the name size.
+ *
+ * @param[in] ObjectHeader
+ * A pointer to an object header, of which name and
+ * directory info are to be retrieved.
+ *
+ * @return
+ * Returns the name info size that is pointed by the
+ * given object by the caller of this function. If
+ * an object does not have a name or no directories,
+ * it returns 0.
+ */
+static
+ULONG
+ObpQueryNameInfoSize(
+    _In_ POBJECT_HEADER ObjectHeader)
+{
+    ULONG NameSize = 0;
+    POBJECT_DIRECTORY ParentDirectory;
+    POBJECT_HEADER_NAME_INFO NameInfo;
+    PAGED_CODE();
+
+    /* Get the name info */
+    NameInfo = OBJECT_HEADER_TO_NAME_INFO(ObjectHeader);
+    if (!NameInfo)
+    {
+        return 0;
+    }
+
+    /* Get the parent directory from the object name too */
+    ParentDirectory = NameInfo->Directory;
+    if (!ParentDirectory)
+    {
+        return 0;
+    }
+
+    /* Take into account the name size of this object and loop for all parent directories */
+    NameSize = sizeof(OBJ_NAME_PATH_SEPARATOR) + NameInfo->Name.Length;
+    for (;;)
+    {
+        /* Get the name info from the parent directory */
+        NameInfo = OBJECT_HEADER_TO_NAME_INFO(
+            OBJECT_TO_OBJECT_HEADER(ParentDirectory));
+        if (!NameInfo)
+        {
+            /* Stop looking if this is the last one */
+            break;
+        }
+
+        /* Get the parent directory */
+        ParentDirectory = NameInfo->Directory;
+        if (!ParentDirectory)
+        {
+            /* This is the last directory, stop looking */
+            break;
+        }
+
+        /*
+         * Take into account the size of this name info,
+         * keep looking for other parent directories.
+         */
+        NameSize += sizeof(OBJ_NAME_PATH_SEPARATOR) + NameInfo->Name.Length;
+    }
+
+    /* Include the size of the object name information as well as the NULL terminator */
+    NameSize += sizeof(OBJECT_NAME_INFORMATION) + sizeof(UNICODE_NULL);
+    return NameSize;
+}
+
 NTSTATUS
 NTAPI
 ObQueryTypeInfo(IN POBJECT_TYPE ObjectType,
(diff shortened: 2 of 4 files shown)


GeoB99:
**Windows Server 2003 SP2**
Capture
**ReactOS**
Capture2

ntoskrnl/ob/oblife.c
```diff
@@ -869,6 +869,94 @@ ObpAllocateObject(IN POBJECT_CREATE_INFORMATION ObjectCreateInfo,
...
+    POBJECT_HEADER DirectoryHeader;
+    POBJECT_HEADER_NAME_INFO NameInfo;
+    PAGED_CODE();
+
+    /* We are grabing the name of this object so reference it */
+    NameInfo = ObpReferenceNameInfo(ObjectHeader);
+    if (!NameInfo)
+    {
+        return 0;
+    }
```
JoachimHenze, line 904: curly braces could be omitted here.

ntoskrnl/ob/oblife.c
```diff
@@ -869,6 +869,94 @@ ObpAllocateObject(IN POBJECT_CREATE_INFORMATION ObjectCreateInfo,
...
+ObpQueryNameInfoSize(
+    _In_ POBJECT_HEADER ObjectHeader)
+{
+    ULONG NameSize = 0;
+    POBJECT_DIRECTORY PrevDirectory, Directory;
+    POBJECT_HEADER DirectoryHeader;
+    POBJECT_HEADER_NAME_INFO NameInfo;
+    PAGED_CODE();
+
+    /* We are grabing the name of this object so reference it */
```
HBelusca, line 899:
```suggestion
    /* We are grabbing the name of this object so reference it */
```

JoachimHenze (approved):
LGTM, aside my curly-braces nitpick.

ntoskrnl/ob/oblife.c
```diff
@@ -869,6 +869,94 @@ ObpAllocateObject(IN POBJECT_CREATE_INFORMATION ObjectCreateInfo,
...
+    /* Grab the directory from the object name too */
+    Directory = NameInfo->Directory;
+    if (!Directory)
+    {
+        ObpDereferenceNameInfo(NameInfo);
+        return 0;
+    }
+
+    /*
+     * We are going to compute the name size of this object of each directory
```
HBelusca, line 915: Sentence not very clear.

ntoskrnl/ob/oblife.c
```diff
@@ -869,6 +869,94 @@ ObpAllocateObject(IN POBJECT_CREATE_INFORMATION ObjectCreateInfo,
...
+        {
+            /* Stop looking if this is the last one */
+            ObDereferenceObject(PrevDirectory);
+            break;
+        }
+
+        /* Take into account the size of this name info */
+        NameSize += sizeof(OBJ_NAME_PATH_SEPARATOR) + NameInfo->Name.Length;
+
+        /* Cache the next directory */
```
HBelusca, line 938:
```suggestion
        /* Get the next directory */
```

modules/rostests/kmtests/ntos_ob/ObQuery.c
```diff
@@ -0,0 +1,113 @@
...
+    /* These must not be 0 because the registry is currently used */
+    ok(BasicInfo.HandleCount != 0, "\\Registry is in use but HandleCount is 0!\n");
+    ok(BasicInfo.PointerCount != 0, "\\Registry is in use but PointerCount is 0!\n");
+
+    /*
+     * NameInfoSize must be 0 because \Registry does not have
+     * directories, instead it has root keys (aka HKEYs).
+     * TypeInfoSize is fixed because it is defined by the
+     * size of an object type though, the size of a registry
+     * key object is 104.
```
HBelusca, line 67: Isn't that different on x64?
Also, should that size correlate with the size of the data retrieved with ObjectTypeInformation ?

modules/rostests/kmtests/ntos_ob/ObQuery.c
```diff
@@ -0,0 +1,113 @@
...
+                           sizeof(BasicInfo),
+                           &ReturnLength);
+    ok_eq_hex(Status, STATUS_SUCCESS);
+
+    /* \GLOBAL?? is currently used */
+    ok(BasicInfo.HandleCount != 0, "\\GLOBAL?? is in use but HandleCount is 0!\n");
+    ok(BasicInfo.PointerCount != 0, "\\GLOBAL?? is in use but PointerCount is 0!\n");
+
+    /*
+     * \\GLOBAL?? has directories but the number of them
```
HBelusca, line 101:
```suggestion
     * \GLOBAL?? has directories but the number of them
```

modules/rostests/kmtests/ntos_ob/ObQuery.c
```diff
@@ -0,0 +1,113 @@
...
+     * varies depending on the system.
+     */
+    trace("NameInfoSize of \\GLOBAL?? directory -- %lu\n", BasicInfo.NameInfoSize);
+    ok_eq_ulong(BasicInfo.TypeInfoSize, OBJ_GLOBAL_DIR_TYPE_INFO_SIZE);
```
HBelusca, line 105: Same question as for the registry type size.

modules/rostests/kmtests/ntos_ob/ObQuery.c
```diff
@@ -0,0 +1,113 @@
...
+                           sizeof(BasicInfo),
+                           &ReturnLength);
+    ok_eq_hex(Status, STATUS_SUCCESS);
+
+    /* These must not be 0 because the registry is currently used */
+    ok(BasicInfo.HandleCount != 0, "\\Registry is in use but HandleCount is 0!\n");
+    ok(BasicInfo.PointerCount != 0, "\\Registry is in use but PointerCount is 0!\n");
+
+    /*
+     * NameInfoSize must be 0 because \Registry does not have
```
HBelusca, line 63: Isn't an object directory an object itself with a name? If so, why isn't this counted in the nameinfosize?

ntoskrnl/ob/oblife.c
```diff
@@ -869,6 +869,94 @@ ObpAllocateObject(IN POBJECT_CREATE_INFORMATION ObjectCreateInfo,
...
+    {
+        ObpDereferenceNameInfo(NameInfo);
+        return 0;
+    }
+
+    /*
+     * We are going to compute the name size of this object of each directory
+     * so each directory that is traversed has to be referenced.
+     */
+    ObReferenceObject(Directory);
```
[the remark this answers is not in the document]
ThFabba: For reference, existing code that already does this same thing: https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/ob/obname.c;hb=942b0221e85bdcf5a4bca7865fb4c72ffa105309#l1296

modules/rostests/kmtests/ntos_ob/ObQuery.c
GeoB99: > Isn't an object directory an object itself with a name? If so, why isn't this counted in the nameinfosize?

Technically yes, an object that has a name provided by the object attributes should be included in the name info but registry keys are special in this regard, as `CmpParseKey` (the registry parser) doesn't take into account the object attributes during an open or create procedure. So whenever a key is parsed like here, we allocate a key body object that is nameless. The Windows' behavior is similar as ours, which is why that test doesn't fail. I guess I should probably update the comment to give an explanation about this.

See those `ObCreateObject` calls in the parser:
https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/config/cmparse.c;h=52a3219babb72077cd3cca42dc610408e1254da3;hb=HEAD#l270
https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/config/cmparse.c;h=52a3219babb72077cd3cca42dc610408e1254da3;hb=HEAD#l682

So basically, as a TL;DR...

- Initialize object attributes, with `\Registry` as object name path
- Open the key with `ZwOpenKey`
- Go all the way down until `CmpParseKey`
- Parse the key, open it with `CmpDoOpen`
- Allocate a key body object with `ObCreateObject`, no object attributes given
- Query basic information with `ZwQueryObject`
- `NameInfoSize == 0`

Now the question is, does `\Registry` have a name if at all? It does, when the registry root is create we don't invoke the registry parser to do it but instead we directly call OB to do the job for us.
https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/config/cmsysini.c;hb=7c14c0ed7346c81311d8179562d489439273dca1#l1090

modules/rostests/kmtests/ntos_ob/ObQuery.c
GeoB99: The length of a name type is arch dependent, so whoops. Also yeah, `TypeInfoSize` correlates with `ObjectTypeInformation` class in one way or another. Just read the article by Geoff I linked.

modules/rostests/kmtests/ntos_ob/ObQuery.c
ThFabba: The `\Registry` object has a name, and it's `\Registry` -- you can query `ObjectNameInformation` (or call `ObQueryNameString`) to find out. I believe `NameInfoSize` is intended to give the length of the buffer required for `ObjectNameInformation`, except that for some reason it just returns 0 for objects in the root directory?

GeoB99: Apparently querying the name with `ObjectNameInformation` does indeed return you the name of the registry object, I guess the discussion in regard of the registry parser above is irrelevant at this point. So I dunno, I'm out of ideas, all I can say is that I am getting a NULL `NameInfo` when grabbing it from the object header, hence `NameInfoSize` will become 0. Everyone is free to volunteer why does it behave like that.

modules/rostests/kmtests/ntos_ob/ObQuery.c
```diff
@@ -0,0 +1,76 @@
...
+ * PROJECT:     ReactOS kernel-mode tests
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Kernel mode tests for object information querying
+ * COPYRIGHT:   Copyright 2023 George Bișoc <george.bisoc@reactos.org>
+ */
+
+#include <kmt_test.h>
+
+#ifdef _WIN64
+#define OBJ_DIRECTORY_NAME_INFO_SIZE  34
```
ThFabba, line 11: I find this a little confusingly named since it's the size for the "Windows" directory specifically, not for any directory.
(I'd still find `sizeof(UNICODE_STRING) + sizeof(L"\\Windows")` most useful, but just numbers work if you prefer 🤷)

ThFabba (approved):

modules/rostests/kmtests/ntos_ob/ObQuery.c
HBelusca: The sizeof solution would indeed be more portable, and this would remove the question "why this value 34 and not e.g. 42 or sthg.".

GeoB99: I refactored the defines, opting for a "sizeof" approach instead. Let me know if this is OK for you.

modules/rostests/kmtests/ntos_ob/ObQuery.c
```diff
@@ -0,0 +1,71 @@
...
+ * PROJECT:     ReactOS kernel-mode tests
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Kernel mode tests for object information querying
+ * COPYRIGHT:   Copyright 2023 George Bișoc <george.bisoc@reactos.org>
+ */
+
+#include <kmt_test.h>
+
+#define OBJ_WINSTA_DIRECTORY_NAME_INFO_SIZE sizeof(UNICODE_STRING) + sizeof(L"\\Windows")
+#define OBJ_WINSTA_DIRECTORY_TYPE_INFO_SIZE sizeof(OBJECT_TYPE_INFORMATION) + sizeof(L"\\Windows") + sizeof(UNICODE_NULL)
```
ThFabba, line 11: Shouldn't this one be based on the string "Directory", not the string "\Windows"?

modules/rostests/kmtests/ntos_ob/ObQuery.c
```diff
@@ -0,0 +1,71 @@
...
+ * PROJECT:     ReactOS kernel-mode tests
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Kernel mode tests for object information querying
+ * COPYRIGHT:   Copyright 2023 George Bișoc <george.bisoc@reactos.org>
+ */
+
+#include <kmt_test.h>
+
+#define OBJ_WINSTA_DIRECTORY_NAME_INFO_SIZE (sizeof(UNICODE_STRING) + sizeof(L"\\Windows"))
+#define OBJ_DIRECTORY_TYPE_INFO_SIZE (sizeof(OBJECT_TYPE_INFORMATION) + (wcslen(L"\\Directory") * sizeof(WCHAR)))
```
ThFabba, line 11: The missing null terminator here is because the type name string is just "Directory" with no backslash, FWIW. So `sizeof(L"Directory")` would work

PR #5180, opened 2023-03-21
[EXPLORER] Improve Clock text color for High Contrast

katahiromz:
## Purpose

Improve usability for High Contrast users.
Make the clock visible.
JIRA issue: [CORE-18886](https://jira.reactos.org/browse/CORE-18886)

## Proposed changes

- Use `GetSysColor(COLOR_BTNTEXT)` to get the system color in `CTrayClockWnd::OnPaint`.

## TODO

- [x] Do tests on themed.
- [x] Do tests on non-themed.

## Screenshots

`High Contrast 1`:
HighCon1

`High Contrast 2`:
HighCon2

`High Contrast Black`:
HighConBlack

### DIFF
--- a/base/shell/explorer/trayclock.cpp
+++ b/base/shell/explorer/trayclock.cpp
@@ -75,7 +75,7 @@ public:
     virtual ~CTrayClockWnd();
 
 private:
-    VOID DoUpdateAppearance();
+    LRESULT OnThemeChanged();
     LRESULT OnThemeChanged(UINT uMsg, WPARAM wParam, LPARAM lParam, BOOL& bHandled);
 
     BOOL MeasureLines();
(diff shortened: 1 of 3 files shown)


base/shell/explorer/trayclock.cpp
```diff
@@ -659,6 +658,9 @@ LRESULT CTrayClockWnd::OnCreate(UINT uMsg, WPARAM wParam, LPARAM lParam, BOOL& b
         ResetTime();
     }
 
+    /* Update hFont and textColor */
```
JoachimHenze, line 661: cpp file. We can use //

katahiromz: Fixed.

base/shell/explorer/trayclock.cpp
```diff
@@ -173,34 +173,19 @@ LRESULT CTrayClockWnd::OnThemeChanged()
...
         if (this->hFont != NULL)
             DeleteObject(this->hFont);
 
         SetFont(hFont, FALSE);
-    }
-    else
-    {
-        /* We don't need to set a font here, our parent will use
-            * WM_SETFONT to set the right one when themes are not enabled. */
-        textColor = RGB(0, 0, 0);
```
HBelusca, line 200: Why wouldn't it be possible to have the new `GetSysColor(COLOR_BTNTEXT);` made here instead? Naively it would seem that this would allow to get the color initialization done at just one place.

katahiromz: Message `WM_SYSCOLORCHANGE` won't reach the child windows.
Message `WM_SYSCOLORCHANGE` just invalidate the window area.
So, getting the system color at `OnPaint` is reasonable.

JoachimHenze (approved):

PR #5186, opened 2023-03-22
[SHELL32] CommandLineToArgvW: fixed vulnerability by incorrect parsing

sebres:
## Purpose

This PR fixes a vulnerability by incorrect parsing of command line arguments.
The way how a triple quote chars (or double quote chars inside a quoted string) will be currently parsed in `CommandLineToArgvW` of ReactOS is wrong, therefore opens an attack vector (using artificially created command line validated with different rules matching default parsing mechanism of Windows), so could be even considered as a vulnerability.

## Proposed changes

`CommandLineToArgvW` will parse command line arguments correct and more similar to Windows.
Here is the diff illustrating wrong behavior:
```diff
# # test is an alias for
# # python.exe -c "from sys import argv; del argv[0]; print(str(len(argv)) + ' | ' + ' | '.join(argv))"
- # ReactOS's Shell32
+ # Windows's Shell32 (and fixed variant)
# double in quoted string:
  > test "abc""def" ghi" xxx
- 2 | abc"def ghi | xxx
+ 2 | abc"def | ghi xxx
  > test "abc"""def" ghi" xxx
- 3 | abc"def | ghi" xxx
+ 2 | abc"def ghi | xxx
# triple in unquoted string:
  > test abc"""def" ghi" xxx
- 2 | abc"def ghi | xxx
+ 2 | abc"def | ghi xxx
  > test abc""""def" ghi" xxx
- 2 | abc"def | ghi xxx
+ 2 | abc"def ghi | xxx
```
As a consequence: 
- the arguments can be parsed incorrectly (compared to default behavior of Windows parser) in the way that parts of quoted arguments may become unquoted and vice versa, as well as swim between different args;
- an attacker may create artificial command line that pass validation rules of nominal condition of Windows, but vulnerable here (inject, data steal, etc), because the arguments would deviate between validation and execution phase
- special tokens like pipe `|`, ampersand `&` or redirecting tokens like `>` that normally included in quoted string (and validated as a string) could abrupt get different meaning and used for piping, redirecting etc, that beside the injection possibility opens still worse attacking vector that can even cause RCE or used to create persistent exploits.

<details><summary>Here is the nominal condition how arguments will be parsed in Windows...</summary>

cmd line|arg1|arg2
---|---|---
abc" "def|abc def|
abc\\" \\"def|abc"|"def
"abc\\" \\"def"|abc" "def|
"abc"" ""def"|abc" "def|
abc"" ""def|abc|def
abc""" """def|abc" "def|
abc\\""" \\"""def|abc"|"def
abc\\\\""" \\\\"""def|abc\\" \\"def
"abc"""def" ghi"</sup>|abc"def ghi|
"abc"""def" ghi<br><sup>* missing close qoute at end</sup>|abc"def ghi|
"abc""def" ghi|abc"def|ghi
"abc\\"""def" ghi|abc""def|ghi
"abc\\\\"""def" ghi"|abc\\"def ghi|
"abc\\\\\\"""def" ghi|abc\\""def|ghi
"abc\\\\\\\\"""def" ghi"|abc\\\\"def ghi|
</details>

For PoC one could use any lang like tcl, python, etc, or even a self-written executable getting the argc/argv from main.
The above diff and nominal condition was generated using this python script on Windows box:
```python
from sys import argv; del argv[0]; print(str(len(argv)) + ' | ' + ' | '.join(argv))
```

### References

Related wine merge request - https://gitlab.winehq.org/wine/wine/-/merge_requests/2472

### DIFF
--- a/dll/win32/shell32/wine/shell32_main.c
+++ b/dll/win32/shell32/wine/shell32_main.c
@@ -176,7 +176,6 @@ LPWSTR* WINAPI CommandLineToArgvW(LPCWSTR lpCmdline, int* numargs)
                     qcount=1;
                 s++;
             }
-            qcount=qcount % 3;
             if (qcount==2)
                 qcount=0;
         }
--- a/dll/win32/shell32/wine/shell32_main.c
+++ b/dll/win32/shell32/wine/shell32_main.c
@@ -172,7 +172,8 @@ LPWSTR* WINAPI CommandLineToArgvW(LPCWSTR lpCmdline, int* numargs)
             /* consecutive quotes, see comment in copying code below */
             while (*s=='"')
             {
-                qcount++;
+                if (++qcount==3)
+                    qcount=1;
                 s++;
             }
             qcount=qcount % 3;
@@ -290,7 +291,7 @@ LPWSTR* WINAPI CommandLineToArgvW(LPCWSTR lpCmdline, int* numargs)
                 if (++qcount==3)
                 {
                     *d++='"';
-                    qcount=0;
+                    qcount=1;
                 }
                 s++;
             }


HBelusca (approved):
:+1: Now the counting loop and the copying loop are identical in terms of how qcount is checked/incremented. As it should be.

You may try to supplement https://git.reactos.org/?p=reactos.git;a=blob;f=modules/rostests/apitests/shell32/CommandLineToArgvW.cpp;hb=16decc61be526657a8ff92cab458a934ecb1367e with new tests validating the fixed behaviour.

Notice also that Wine itself (from which this code comes from) appear vulnerable as well:
https://github.com/wine-mirror/wine/blob/67c67703311f75540896f29ec576fb8fac427121/dlls/shcore/main.c#L376

katahiromz:
Thanks for reporting!

Did you test `shell32_apitest` or `shell32_winetest` in `C:\ReactOS\bin`?

sebres:
> You may try to supplement

Yep, I'd still wait for CI (in order to fix some broken tests too)...

> Notice also that Wine itself (from which this code comes from) appear vulnerable as well

Yeah, will provide a fix too... (lets fix it everywhere, IIRC it is 5th vulni this week:)

By the way, MS has another [similar vulni](https://github.com/microsoft/terminal/issues/250) (initially CRM:0461062796) in a batch processor that they reject to fix - but maybe it could be properly fixed in ReactOS.

HBelusca:
> By the way, MS has another [similar vulni](https://github.com/microsoft/terminal/issues/250) (initially CRM:0461062796) in a batch processor that they reject to fix - but maybe it could be properly fixed in ReactOS.

Regarding that one, they indeed have a policy of not breaking old behaviour:
> Changing how cmd handles arguments isn’t feasible because it could break many existing applications.

I would be interested to know whether you can reproduce it with our cmd.exe, since I've tried lately to make its parser behave the closest to windows' own cmd.exe (I've also spotted these sorts of bugs that I've ifdef'ed so that one could select at compile time a given behaviour.) But this discussion would better be moved either to Mattermost / to another suitable PR.

sebres:
Here is the related wine merge request - https://gitlab.winehq.org/wine/wine/-/merge_requests/2472

katahiromz:
If parsing of Windows has vuln, we can adopt another implementation.

sebres:
@HBelusca

> I would be interested to know whether you can reproduce it with our cmd.exe

You can download the [test-dump-details.cmd](https://github.com/sebres/PoC/blob/master/SB-0D-001-win-exec/test-dump-details.cmd) and [test-dump.exe](https://github.com/sebres/PoC/raw/master/SB-0D-001-win-exec/test-dump.exe) or compile it from `test-exec.c` with `gcc -mconsole test-exec.c -o test-dump.exe`.
And test it in with your cmd with args **containing unbalanced quote-chars in single parameter**, for example [from here](https://github.com/microsoft/terminal/issues/250) and then inspect console output (and `test-result.txt` file created from, just to illustrate the deviation of exe-invocation to `echo` from batch).
<details><summary>For instance ...</summary>

the output of
`test-dump-details.cmd "A B\"" "C D" "\"E F"`
and 
`test-dump.exe "\"E F" "C D" "A B\"" & echo ^ - revers & test-dump.exe "A B\"" "C D" "\"E F" & echo ^ - normal`
must produce the same result for that invocation regardless from batch with interpolated `%1 ... %3` or from comspec directly, but on windows it'd generate this:
```diff
-   `test-dump.exe´ `D "E´ `F A´ `B" C´ - revers (* from batch)
+   `test-dump.exe´ `"E F´ `C D´ `A B"´ - revers (* direct)
    `test-dump.exe´ `A B"´ `C D´ `"E F´ - normal
```
that's about mis-escape of `%1 ... %9`, or rather it becomes visible in reversed order `%9 ... %1`.
</details>

However the actual "solution" is pretty easy - simply don't use windows batch for something serious. There are enough script langs across the world doing safe a parameter escape and substitution.

@katahiromz 
> If parsing of Windows has vuln, we can adopt another implementation.

I doubt that this batch misconception (inclusive different handling between `%1 ... %9` and `%*` interpolation) could be indeed fixed in a fully compatible way, but one could surely close the vulni, for instance either with a fix for `%...` in a quoted substitution or with a throwing of error in case of unbalanced quotes-chars within single parameter by usage of `%...` (or adjusting a parameter with additional closing quote char or extra escape sequence). 
I guess the substitution of args inside the batch file is simply inappropriate or insufficient escaped.

sebres:
Trying to write new tests here (and fix tests for wine) I got confused much more now.

Weird, but everyone seems really to use own parser for splitting of command line parameters.
The tests are good (and this PR is then wrong), if one tries it with a small [test-cltaw.exe](https://github.com/sebres/PoC/raw/master/SB-0D-001-win-exec/test-cltaw.exe) (from [test-cltaw.c](https://github.com/sebres/PoC/blob/master/SB-0D-001-win-exec/test-cltaw.c)) using `CommandLineToArgvW` on windows. Exactly in the same way it does with [test-dump.exe](https://github.com/sebres/PoC/raw/master/SB-0D-001-win-exec/test-dump.exe) using `args/argv` pair from `main`, but...

But it completely deviates from `python`, `tcl`, whatever script lang, that seem to use own command line parser for arguments if built for windows (and then the PR would be correct).

<details><summary>Here is small table illustrating the issue...</summary>

<table>
<tr><th rowspan="2">command&nbsp;line</th><th colspan="4">CommandLineToArgvW&nbsp;/&nbsp;main</th><th colspan="4">python,&nbsp;tcl,&nbsp;etc</th></tr>
<tr><th>C</th><th>A1</th><th>A2</th><th>A3</th><th>C</th><th>A1</th><th>A2</th><th>A3</th></tr>
<tr><td>three"""quotes&nbsp;next</td><td>2</td><td>three"quotes</td><td>next</td><td></td>  <td>1</td><td>three"quotes&nbsp;next</td><td></td><td></tr>
<tr><td>four""""&nbsp;quotes"&nbsp;next&nbsp;4%3=1</td><td>3</td><td>four"&nbsp;quotes</td><td>next</td><td>4%3=1</td>  <td>2</td><td>four"</td><td>quotes&nbsp;next&nbsp;4%3=1</td><td></tr>
<tr><td>five"""""quotes&nbsp;next</td><td>2</td><td>five"quotes</td><td>next</td><td></td>  <td>1</td><td>five""quotes&nbsp;next</td><td></td><td></tr>
<tr><td>seven"""""""&nbsp;quotes"&nbsp;next&nbsp;7%3=1</td><td>3</td><td>seven""&nbsp;quotes</td><td>next</td><td>7%3=1</td>  <td>3</td><td>seven"""&nbsp;quotes</td><td>next</td><td>7%3=1</tr>
<tr><td>twelve""""""""""""quotes&nbsp;next</td><td>2</td><td>twelve""""quotes</td><td>next</td><td></td>  <td>2</td><td>twelve"""""quotes</td><td>next</td><td></tr>
<tr><td>thirteen"""""""""""""&nbsp;quotes"&nbsp;next&nbsp;13%3=1</td><td>3</td><td>thirteen""""&nbsp;quotes</td><td>next</td><td>13%3=1</td>  <td>3</td><td>thirteen""""""&nbsp;quotes</td><td>next</td><td>13%3=1</tr>
<tr><td>"two""quotes&nbsp;next</td><td>2</td><td>two"quotes</td><td>next</td><td></td>  <td>1</td><td>two"quotes&nbsp;next</td><td></td><td></tr>
<tr><td>"two""&nbsp;next</td><td>2</td><td>two"</td><td>next</td><td></td>  <td>1</td><td>two"&nbsp;next</td><td></td><td></tr>
<tr><td>"three"""&nbsp;quotes"&nbsp;next&nbsp;4%3=1</td><td>3</td><td>three"&nbsp;quotes</td><td>next</td><td>4%3=1</td>  <td>2</td><td>three"</td><td>quotes&nbsp;next&nbsp;4%3=1</td><td></tr>
<tr><td>"four""""quotes&nbsp;next</td><td>2</td><td>four"quotes</td><td>next</td><td></td>  <td>1</td><td>four""quotes&nbsp;next</td><td></td><td></tr>
<tr><td>"six""""""&nbsp;quotes"&nbsp;next&nbsp;7%3=1</td><td>3</td><td>six""&nbsp;quotes</td><td>next</td><td>7%3=1</td>  <td>3</td><td>six"""&nbsp;quotes</td><td>next</td><td>7%3=1</tr>
<tr><td>"eleven"""""""""""quotes&nbsp;next</td><td>2</td><td>eleven""""quotes</td><td>next</td><td></td>  <td>2</td><td>eleven"""""quotes</td><td>next&nbsp;|</tr>
<tr><td>"twelve""""""""""""&nbsp;quotes"&nbsp;next&nbsp;13%3=1</td><td>3</td><td>twelve""""&nbsp;quotes</td><td>next</td><td>13%3=1</td>  <td>3</td><td>twelve""""""&nbsp;quotes</td><td>next</td><td>13%3=1</tr>
<tr><td>"the&nbsp;crazy&nbsp;\\"""\\"&nbsp;quotes</td><td>2</td><td>the&nbsp;crazy&nbsp;\"\</td><td>quotes</td><td></td>  <td>1</td><td>the&nbsp;crazy&nbsp;\"\&nbsp;quotes</td><td></td><td></tr>
</table>
</details>

Just since it is supposed to be a fix for `CommandLineToArgvW` (to behave more similar on windows), I'll close it now (and gonna open few PRs for the mentioned langs for windows).

Thus closed, sorry for noise.

HBelusca:
> Just since it is supposed to be a fix for CommandLineToArgvW (to behave more similar on windows)

Yes let's just focus on that function (+ adding possibly some extra apitests for it) for now. By the way the original fix of yours (making the "counting" loop similar to the actual copying loop) was correct IMO.

PR #5187, opened 2023-03-22
KDGDB: Don't make it a separate citizen + fix some build errors.

HBelusca:
## Purpose

KDGDB: Don't make it a separate citizen + fix some build errors.


drivers/base/kdgdb/i386_sup.c
```diff
@@ -95,7 +95,7 @@ thread_to_reg(PETHREAD Thread, enum reg_name reg_name, unsigned short* size)
             case EIP:
                 KDDBGPRINT("Returning NULL for register %d.\n", reg_name);
                 *size = 4;
-                return &NullValue;
+                return (void*)&NullValue;
```
zefklop, line 98: You can as well change the function return type to be const. That would be more invasive though.

zefklop (approved):

drivers/base/CMakeLists.txt
```diff
@@ -4,13 +4,10 @@ add_subdirectory(bootvid)
...
 if(_WINKD_)
-    if(GDB)
-        add_subdirectory(kdgdb)
-    else()
-        add_subdirectory(kdcom)
-        if(NOT ARCH STREQUAL "arm")
-            add_subdirectory(kdvm)
-        endif()
+    add_subdirectory(kdcom)
+    add_subdirectory(kdgdb)
```
learn-more, line 8: This only means it will be created, it will not always be loaded?

HBelusca: Indeed, not until you modify the boot option to load kdgdb instead (`/DEBUGPORT=GDB` instead of `/DEBUGPORT=COMn` with `n` an integer), or you manually rename kdgdb.dll to kdcom.dll (and change the other's name) in your install.

drivers/base/kdgdb/i386_sup.c
HBelusca: I finally decided to go along with const'-ifying the return value of `ctx_to_reg` and `thread_to_reg` functions.

drivers/base/CMakeLists.txt
HBelusca: For Gitpod (whose support needs to be updated, by the way....) I've kept the GDB build switch, and added a workaround to rename the dll file name (kdgdb becomes kdcom, and the original kdcom gets another name).

PR #5188, opened 2023-03-23
[NTOS:KDBG] Isolate terminal and input-related routines from the rest of KDBG.

HBelusca:
## Purpose

This is done in preparation for moving all this functionality in a separate KDTERM "KD Terminal Driver" DLL.
It would allow KDBG to not be strongly tied to serial (or PS/2) port for its input/output, instead relying on an helper DLL taking the form of a KD transport DLL. This would permit to instead direct KDBG output to... WinDbg (!!) by using as an alternative terminal DLL the standard KDCOM (or KDNET etc.).
Furthermore, such a DLL could be used as an alternative "debugport" transport dll to **only** view some DbgPrint output (and interact with prompts from DbgPrompt) while running Windows.

Additionally in this PR:

- Flush the terminal input before sending ANSI escape sequences.

- Use a specific KdbpSendCommandSerial() function to send these sequences, instead of KdpDprintf()! Otherwise, we send them to **ALL** the display providers, including for example dmesg. Replaying the listing with dmesg would then cause the terminal to misbehave later. For example, it would send the answer of a "Query Device Attributes" command, as the response to a query for terminal size...


learn-more:
To be honest I don't see the point of redirecting kdbg output to windbg while we already have something that can communicate with windbg?

learn-more:

PR #5189, opened 2023-03-23
Tag Win32SS related PRs with "Win32SS" label

GeoB99:
Let the GitHub bot automatically tag PRs that touch Win32k code or Win32 client DLLs (namely GDI, USER, whatever). With this new label we can browse through these kind of PRs easily, similarly with those that touch the kernel which are labeled "Kernel & Hal".

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -13,3 +13,6 @@ freeldr:
 
 ROSTESTS:
   - modules/rostests/**
+
+Win32SS:
+  - win32ss/**


HBelusca (approved):

oleg-dubinskiy (approved):
LGTM. :slightly_smiling_face: 
Perhaps it makes sense to label the base/applications, dll/win32 etc. also by the same way (in separate PRs)? :thinking: To have a labels for more system parts. On my opinion, it's a good idea.

PR #5196, opened 2023-03-26
[FREELDR][SDK] Build UEFI bootloader for ARM32

DarkFire01:
## Purpose

In the beginning of the ARM port FreeLoader's UI functions were too slow on targeted devices. As such the solution was to split off from that and minimize what those ARM devices needed to do to achieve the FreeLoader Ui. Now in the present world we use MSVC which breaks compatibility with the devices we used to target; None of these newer device have issues with the way FreeLoader draws so we are going to attempt to unify this here today. This PR lays nearly everything we need for ARM32 to boot up to FreeLoader from EFI ARM devices. 

-> This PR will have a single link error left that'll be solved in another PR due to its complexity 

JIRA issue: [CORE-17604](https://jira.reactos.org/browse/CORE-17604)

## Proposed changes

- Disable some redundant function defined elsewhere on ARM32
- Remove some unused link options
- link _controlfp to the float cmake target
- Add the missing ARM specific code into the mach file of ARM freeeldr
- Add a hard coded UART address for QEMU ARM32/ARM64 so we can have serial debugging in freeldr
- Unify the UI drawing calls

### DIFF
--- a/boot/freeldr/freeldr/arch/arm/debug.c
+++ b/boot/freeldr/freeldr/arch/arm/debug.c
@@ -1,0 +1,35 @@
+/*
+ * PROJECT:     Freeldr ARM32
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Arch specific debug
+ * COPYRIGHT:   Copyright 2022 Justin Miller <justinmiller100@gmail.com>
+ */
+
+#include <freeldr.h>
+#include <debug.h>
+
+#define QEMUUART 0x09000000
+volatile unsigned int * UART0DR = (unsigned int *) QEMUUART;
+
+BOOLEAN
+Rs232PortInitialize(IN ULONG ComPort,
+                    IN ULONG BaudRate)
+{
+    return TRUE;
+}
+
+VOID
+Rs232PortPutByte(UCHAR ByteToSend)
+{
+    *UART0DR = ByteToSend;
+}
+
+VOID
+FrLdrBugCheckWithMessage(
+    ULONG BugCode,
+    PCHAR File,
+    ULONG Line,
+    PSTR Format,
+    ...)
+{
+}
--- a/boot/freeldr/freeldr/arch/arm/macharm.c
+++ b/boot/freeldr/freeldr/arch/arm/macharm.c
@@ -18,7 +18,9 @@ ULONG gDiskReadBuffer, gFileSysBuffer;
 BOOLEAN ArmHwDetectRan;
 PCONFIGURATION_COMPONENT_DATA RootNode;
 
+#ifndef UEFIBOOT
 BOOLEAN AcpiPresent = FALSE;
+#endif
 
 ULONG FirstLevelDcacheSize;
 ULONG FirstLevelDcacheFillSize;
@@ -174,6 +176,7 @@ ArmHwIdle(VOID)
     /* UNIMPLEMENTED */
 }
 
+#ifndef UEFIBOOT
 VOID
 MachInit(IN PCCH CommandLine)
 {
@@ -229,3 +232,4 @@ MachInit(IN PCCH CommandLine)
     MachVtbl.HwDetect = ArmHwDetect;
     MachVtbl.HwIdle = ArmHwIdle;
 }
+#endif
--- a/boot/freeldr/freeldr/arch/uefi/uefildr.c
+++ b/boot/freeldr/freeldr/arch/uefi/uefildr.c
@@ -31,7 +31,9 @@ EfiEntry(
     return 0;
 }
 
+#ifndef _M_ARM
 VOID __cdecl Reboot(VOID)
 {
 
 }
+#endif
--- a/boot/freeldr/freeldr/lib/debug.c
+++ b/boot/freeldr/freeldr/lib/debug.c
@@ -20,7 +20,7 @@
 #include <freeldr.h>
 #include <debug.h>
 
-#if DBG && !defined(_M_ARM)
+#if DBG
 
 // #define DEBUG_ALL
 // #define DEBUG_WARN
@@ -53,6 +53,11 @@ ULONG ComPort  = 0; // The COM port initializer chooses the first available port
 ULONG PortIrq  = 0; // Not used at the moment.
 
 BOOLEAN DebugStartOfLine = TRUE;
+
+#ifdef UEFIBOOT
+VOID
+ARMWriteToUART(UCHAR Data);
+#endif
 
 VOID DebugInit(IN ULONG_PTR FrLdrSectionId)
 {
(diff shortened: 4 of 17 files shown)


boot/freeldr/freeldr/lib/debug.c
```diff
@@ -203,10 +210,17 @@ VOID DebugPrintChar(UCHAR Character)
...
     if (DebugPort & RS232)
     {
+#ifndef _M_ARM
         if (Character == '\n')
             Rs232PortPutByte('\r');
 
         Rs232PortPutByte(Character);
+#else
+        if (Character == '\n')
+            ARMWriteToUART('\r');
```
HBelusca, line 220: I wonder whether it could be of interest to alias `Rs232PortPutByte` to `ARMWriteToUART` and `Rs232PortInitialize` to a dummy function, in order to avoid those if(n)defs here -- at the expense of having them in the rs232.c file or, in the corresponding header? (that latter option may be slightly less invasive.)

boot/freeldr/freeldr/arch/arm/macharm.c
```diff
@@ -229,3 +231,4 @@ MachInit(IN PCCH CommandLine)
     MachVtbl.HwDetect = ArmHwDetect;
     MachVtbl.HwIdle = ArmHwIdle;
 }
+#endif
```
HBelusca, line 234:
```suggestion
#endif

```

boot/freeldr/freeldr/arch/uefi/uefildr.c
```diff
@@ -31,7 +31,9 @@ EfiEntry(
     return 0;
 }
 
+#ifndef _M_ARM
 VOID __cdecl Reboot(VOID)
 {
 
 }
+#endif
```
HBelusca, line 39:
```suggestion
#endif

```

boot/freeldr/freeldr/arch/arm/debug.c
```diff
@@ -0,0 +1,29 @@
+
+/* INCLUDES ***************************************************************/
+
+#include <freeldr.h>
+#include <debug.h>
+
+#define QEMUUART 0x09000000
+
+volatile unsigned int * UART0DR = (unsigned int *) QEMUUART;
```
HBelusca, line 9: Is that something that works *just* in QEMU?

HBelusca (commented):
So now with these changes, you are able to get UI output on those ARM32 devices?

boot/freeldr/freeldr/arch/arm/debug.c
DarkFire01: This is for ARM64 and ARM32 QEMU yes, they share the same UART controller. I don't plan on adding support for any specific ARM32 devices UARTs inside freeldr beyond this.

sdk/lib/crt/float/float.cmake
```diff
@@ -40,8 +40,15 @@ elseif(ARCH STREQUAL "arm")
...
+    list(APPEND LIBCNTPR_FLOAT_SOURCE
+        float/arm/_controlfp.c
+    )
     list(APPEND CRT_FLOAT_ASM_SOURCE
         float/arm/__getfp.s
         float/arm/__setfp.s
     )
+    list(APPEND LIBCNTPR_FLOAT_ASM_SOURCE
+        float/arm/__getfp.s
+        float/arm/__setfp.s
```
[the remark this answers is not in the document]
DarkFire01: It gets spammed as a compiler error without it, so it's used for multiple modules.

boot/freeldr/freeldr/ui/ui.c
```diff
@@ -56,16 +54,13 @@ const PCSTR UiMonthNames[12] = { "January", "February", "March", "April", "May",
...
-#endif // _M_ARM
-
 /*
  * Loading progress bar, based on the NTOS Inbv one.
  * Supports progress within sub-ranges, used when loading
  * with an unknown number of steps.
  */
 UI_PROGRESS_BAR UiProgressBar = {{0}};
 
-#ifndef _M_ARM
```
JoachimHenze, line 64: strip one empty line here additionally.

boot/freeldr/freeldr/ui/ui.c
```diff
@@ -386,7 +381,6 @@ UCHAR UiTextToFillStyle(PCSTR FillStyleText)
     return UiVtbl.TextToFillStyle(FillStyleText);
 }
 
-#endif // _M_ARM
```
JoachimHenze, line 384: strip one empty line here additionally.

boot/freeldr/freeldr/arch/arm/macharm.c
```diff
@@ -173,7 +175,7 @@ ArmHwIdle(VOID)
 {
     /* UNIMPLEMENTED */
 }
-
+#ifndef UEFIBOOT
```
JoachimHenze, line 178: readd an empty line here to separate both functions.

boot/freeldr/freeldr/uefi.cmake
```diff
@@ -84,7 +87,11 @@ set_target_properties(uefildr PROPERTIES SUFFIX ".efi")
 target_compile_definitions(uefildr PRIVATE UEFIBOOT)
 
 if(MSVC)
+if(ARCH STREQUAL "arm")
+    target_link_options(uefildr PRIVATE /ignore:4078 /ignore:4254 /DRIVER)
```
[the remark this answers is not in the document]
DarkFire01: from MSDN:
> Because ASLR can't be disabled on ARM, ARM64, or ARM64EC architectures, /DYNAMICBASE:NO isn't supported for these targets.

boot/freeldr/freeldr/ui/tuimenu.c
```diff
@@ -231,43 +206,16 @@ TuiDrawMenu(
         TuiDrawMenuItem(MenuInfo, i);
     }
 
-#ifndef _M_ARM // FIXME: Theme-specific
-
     /* Update the status bar */
```
HBelusca, line 209:
```suggestion
    // FIXME: Theme-specific
    /* Update the status bar */
```

boot/freeldr/freeldr/uefi.cmake
```diff
@@ -28,6 +28,9 @@ if(ARCH STREQUAL "i386")
 elseif(ARCH STREQUAL "amd64")
     #TBD
 elseif(ARCH STREQUAL "arm")
+    list(APPEND UEFILDR_ARC_SOURCE
+        arch/arm/macharm.c
+        arch/arm/debug.c)
```
HBelusca, line 33: Are those files used only in *UEFI* builds of ARM32 (and not in non-UEFI builds)?

DarkFire01: We talked about it else where putting it here for record;
I put these together specifically for UEFI/QEMU, we can adapt them to normal freeldr if we ever fix it one day for arm.

PR #5197, opened 2023-03-27
[NTOS:IO] Fix CurrentByteOffset to FILE_APPEND_DATA DesiredAccess

zecarlos1957:
As the NtWriteFile test indicates for FILE_APPEND_DATA access right, the file pointer must be set to the end of the file in NtWriteFile before the write operation.

JIRA issue: [CORE-18789](https://jira.reactos.org/browse/CORE-18789)

### DIFF
--- a/ntoskrnl/io/iomgr/iofunc.c
+++ b/ntoskrnl/io/iomgr/iofunc.c
@@ -3907,6 +3907,8 @@ NtWriteFile(IN HANDLE FileHandle,
         /* Give the drivers something to understand */
         CapturedByteOffset.u.LowPart = FILE_WRITE_TO_END_OF_FILE;
         CapturedByteOffset.u.HighPart = -1;
+        FileObject->CurrentByteOffset.u.LowPart = FILE_WRITE_TO_END_OF_FILE;
+        FileObject->CurrentByteOffset.u.HighPart = -1;
     }
 
     /* Check for event */


ntoskrnl/io/iomgr/iofunc.c
```diff
@@ -3907,6 +3907,8 @@ NtWriteFile(IN HANDLE FileHandle,
         /* Give the drivers something to understand */
         CapturedByteOffset.u.LowPart = FILE_WRITE_TO_END_OF_FILE;
         CapturedByteOffset.u.HighPart = -1;
+        FileObject->CurrentByteOffset.u.LowPart = FILE_WRITE_TO_END_OF_FILE;
```
HBelusca, line 3910: Certainly not. The CurrentByteOffset of a FileObject always contains the actual offset.

HBelusca (changes requested):
Your suggestion doesn't appear to be correct.
Please check while debugging that the handling of file writing to end in the functions [FsRtlCopyWrite](https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/fsrtl/fastio.c;hb=7c755d9b849a75382cc3579fe146c05c3ac08f14#l262)
and [FsRtlPrepareMdlWriteDev](https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/fsrtl/fastio.c;hb=7c755d9b849a75382cc3579fe146c05c3ac08f14#l1270) behave correctly.

PR #5202, opened 2023-03-27
[SHELL32] Add support for shell: folder shortcuts

unia909:
## Purpose

Add support for shell: folder shortcuts
JIRA issue: [CORE-16939](https://jira.reactos.org/browse/CORE-16939)

## Proposed changes

- Add `_ILCreateGuidFromShellUrlW` function to resolve shell: folder shortcuts and GUIDs

### DIFF
--- a/dll/win32/shell32/wine/pidl.c
+++ b/dll/win32/shell32/wine/pidl.c
@@ -1682,11 +1682,25 @@ LPITEMIDLIST _ILCreateBitBucket(void)
     return _ILCreateGuid(PT_GUID, &CLSID_RecycleBin);
 }
 
+#ifdef __REACTOS__
+LPITEMIDLIST _ILCreateFonts(void)
+{
+    TRACE("()\n");
+    return _ILCreateGuid(PT_GUID, &CLSID_FontsFolderShortcut);
+}
+
 LPITEMIDLIST _ILCreateAdminTools(void)
 {
     TRACE("()\n");
     return _ILCreateGuid(PT_GUID, &CLSID_AdminFolderShortcut); //FIXME
 }
+
+LPITEMIDLIST _ILCreateConnections(void)
+{
+    TRACE("()\n");
+    return _ILCreateGuid(PT_GUID, &CLSID_ConnectionFolder);
+}
+#endif
 
 LPITEMIDLIST _ILCreateGuid(PIDLTYPE type, REFIID guid)
 {
@@ -1724,7 +1738,34 @@ LPITEMIDLIST _ILCreateGuidFromStrA(LPCSTR szGUID)
     }
     return _ILCreateGuid(PT_GUID, &iid);
 }
-#endif
+#endif // __REACTOS__
+
+#ifdef __REACTOS__
+UINT _SHGetCSIDLByNameW(LPCWSTR lpszName);
+
+// If szUrl is '::{GUID}' then returns a GUID
+// otherwise it tries to resolve the shortcut.
+LPITEMIDLIST _ILCreateGuidFromShellUrlW(LPCWSTR szUrl)
+{
+    UINT idx;
+    LPITEMIDLIST pidl;
+
+    if (szUrl[0] == L':' && szUrl[1] == L':')
+    {
+        return _ILCreateGuidFromStrW(szUrl + 2);
+    }
+
+    idx = _SHGetCSIDLByNameW(szUrl);
+    if (idx == UINT_MAX)
+    {
+        ERR("%s is not a shell shortcut\n", debugstr_w(szUrl));
+        return NULL;
+    }
+
+    SHGetSpecialFolderLocation(NULL, idx, &pidl);
+    return pidl;
+}
+#endif // __REACTOS__
 
 LPITEMIDLIST _ILCreateGuidFromStrW(LPCWSTR szGUID)
 {
--- a/dll/win32/shell32/wine/pidl.h
+++ b/dll/win32/shell32/wine/pidl.h
@@ -269,6 +269,10 @@ LPITEMIDLIST	_ILCreateGuid(PIDLTYPE type, REFIID guid) DECLSPEC_HIDDEN;
 LPITEMIDLIST	_ILCreateGuidFromStrA(LPCSTR szGUID) DECLSPEC_HIDDEN;
 LPITEMIDLIST	_ILCreateGuidFromStrW(LPCWSTR szGUID) DECLSPEC_HIDDEN;
 
+#ifdef __REACTOS__
+LPITEMIDLIST    _ILCreateGuidFromShellUrlW(LPCWSTR szUrl) DECLSPEC_HIDDEN;
+#endif
+
 /* Commonly used PIDLs representing file system objects. */
 LPITEMIDLIST	_ILCreateDesktop	(void) DECLSPEC_HIDDEN;
 LPITEMIDLIST	_ILCreateFromFindDataW(const WIN32_FIND_DATAW *stffile) DECLSPEC_HIDDEN;
@@ -283,7 +287,9 @@ LPITEMIDLIST	_ILCreatePrinters	(void) DECLSPEC_HIDDEN;
 LPITEMIDLIST	_ILCreateNetwork	(void) DECLSPEC_HIDDEN;
 LPITEMIDLIST	_ILCreateNetHood	(void) DECLSPEC_HIDDEN;
 #ifdef __REACTOS__
+LPITEMIDLIST	_ILCreateFonts		(void);
 LPITEMIDLIST	_ILCreateAdminTools	(void);
+LPITEMIDLIST	_ILCreateConnections(void);
 #endif
 LPITEMIDLIST	_ILCreateBitBucket	(void) DECLSPEC_HIDDEN;
 LPITEMIDLIST	_ILCreateDrive		(LPCWSTR) DECLSPEC_HIDDEN;
--- a/dll/win32/shell32/wine/shellpath.c
+++ b/dll/win32/shell32/wine/shellpath.c
@@ -1110,14 +1110,14 @@ static const CSIDL_DATA CSIDL_Data[] =
     { /* 0x03 - CSIDL_CONTROLS (.CPL files) */
         &FOLDERID_ControlPanelFolder,
         CSIDL_Type_SystemPath,
-        NULL,
+        L"ControlPanelFolder",
         NULL,
         -IDI_SHELL_CONTROL_PANEL
     },
     { /* 0x04 - CSIDL_PRINTERS */
         &FOLDERID_PrintersFolder,
         CSIDL_Type_SystemPath,
-        NULL,
+        L"PrintersFolder",
         NULL,
         -IDI_SHELL_PRINTERS_FOLDER
     },
@@ -1157,7 +1157,7 @@ static const CSIDL_DATA CSIDL_Data[] =
     { /* 0x0a - CSIDL_BITBUCKET - Recycle Bin */
         &FOLDERID_RecycleBinFolder,
         CSIDL_Type_Disallowed,
-        NULL,
+        L"RecycleBinFolder",
         NULL
     },
     { /* 0x0b - CSIDL_STARTMENU */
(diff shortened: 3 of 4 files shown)


sdk/include/psdk/shlguid.h
```diff
@@ -146,6 +146,7 @@ DEFINE_GUID(CLSID_NetworkPlaces, 0x208D2C60, 0x3AEA, 0x1069, 0xA2, 0xD7, 0x08, 0
 DEFINE_GUID(CLSID_NetworkDomain, 0x46e06680, 0x4bf0, 0x11d1, 0x83, 0xee, 0x00, 0xa0, 0xc9, 0x0d, 0xc8, 0x49);
 DEFINE_GUID(CLSID_NetworkServer, 0xc0542a90, 0x4bf0, 0x11d1, 0x83, 0xee, 0x00, 0xa0, 0xc9, 0x0d, 0xc8, 0x49);
 DEFINE_GUID(CLSID_NetworkShare, 0x54a754c0, 0x4bf0, 0x11d1, 0x83, 0xee, 0x00, 0xa0, 0xc9, 0x0d, 0xc8, 0x49);
+DEFINE_GUID(CLSID_NetworkConnections, 0x7007ACC7, 0x3202, 0x11D1, 0xAA, 0xD2, 0x00, 0x80, 0x5F, 0xC1, 0x27, 0x0E);
```
HBelusca, line 149: Check this commit: https://github.com/reactos/reactos/commit/bea0b47db66519925a707cf67bbbcde9c8adf180 as well as include/reactos/shlguid_undoc.h regarding this extra GUID.

dll/win32/shell32/wine/pidl.c
```diff
@@ -1722,6 +1722,80 @@ LPITEMIDLIST _ILCreateGuidFromStrA(LPCSTR szGUID)
 }
 #endif
 
+// Create GUID from shortcut(desktop, Personal...)
+// returns TRUE if succefull and FALSE otherwise
+BOOL _ILCreateGuidFromShortcutW(LPCWSTR szShortcut, GUID *fidOut)
+{
+	if (!StrCmpIW(szShortcut, L"ADMINISTRATIVE TOOLS"))
```
HBelusca, line 1729: I wonder whether (especially if more to be added in the future) you could use a kind of table "lookup" that would avoid code repetition, something like this:
```c
static const struct {
    LPCWSTR pszName;
    GUID guid;
} ShellShortcuts[] =
{
    {L"CONTROLPANELFOLDER", CLSID_ControlPanel},
    {L"FONTS", CLSID_FontsFolderShortcut},
/* et caetera, you see the idea... */
};

    UINT i;
    for (i = 0; i < _countof(ShellShortcuts); ++i)
    {
        if (StrCmpIW(szShortcut, ShellShortcuts[i].pszName) == 0)
        {
            *fidOut = ShellShortcuts[i].guid;
            return TRUE;
            return TRUE;
        }
    }

    return FALSE;
```

dll/win32/shell32/wine/pidl.c
```diff
@@ -1722,6 +1722,80 @@ LPITEMIDLIST _ILCreateGuidFromStrA(LPCSTR szGUID)
...
+	{
+		*fidOut = CLSID_AdminFolderShortcut;
+		return TRUE;
+	}
+	else if (!StrCmpIW(szShortcut, L"CONTROLPANELFOLDER"))
```
HBelusca, line 1734: Also I noticed you used tabs for indentation. Please use 4-space (this is mentioned in our code guidelines).

dll/win32/shell32/wine/pidl.c
```diff
@@ -1722,6 +1722,80 @@ LPITEMIDLIST _ILCreateGuidFromStrA(LPCSTR szGUID)
 }
 #endif
 
+// Create GUID from shortcut(desktop, Personal...)
+// returns TRUE if succefull and FALSE otherwise
+BOOL _ILCreateGuidFromShortcutW(LPCWSTR szShortcut, GUID *fidOut)
```
HBelusca, line 1727: Could be made "static" if only used in this file.

dll/win32/shell32/wine/pidl.c
```diff
@@ -1722,6 +1722,80 @@ LPITEMIDLIST _ILCreateGuidFromStrA(LPCSTR szGUID)
 }
 #endif
 
+// Create GUID from shortcut(desktop, Personal...)
+// returns TRUE if succefull and FALSE otherwise
```
HBelusca, line 1726:
```suggestion
// Returns TRUE if successful and FALSE otherwise.
```

dll/win32/shell32/wine/pidl.c
```diff
@@ -1722,6 +1722,80 @@ LPITEMIDLIST _ILCreateGuidFromStrA(LPCSTR szGUID)
...
+	else if (!StrCmpIW(szShortcut, L"RECYCLEBINFOLDER"))
+	{
+		*fidOut = CLSID_RecycleBin;
+		return TRUE;
+	}
+	return FALSE;
+}
+
+// if szUrl is '::{GUID}' then returns a GUID
+// else it tries to resolve shortcut
```
HBelusca, line 1778:
```suggestion
// If szUrl is '::{GUID}' then returns a GUID
// otherwise it tries to resolve the shortcut.
```

dll/win32/shell32/wine/pidl.c
```diff
@@ -1722,6 +1722,80 @@ LPITEMIDLIST _ILCreateGuidFromStrA(LPCSTR szGUID)
...
+			return NULL;
+		}
+	}
+	else if (!_ILCreateGuidFromShortcutW(szUrl, &iid))
+	{
+		ERR("%s is not a shell shortcut\n", debugstr_w(szUrl));
+		return NULL;
+	}
+    return _ILCreateGuid(PT_GUID, &iid);
+}
```
HBelusca, line 1797: Should those two functions placed in file of "wine" subdirectory?
At least they should be surrounded by a `#ifdef __REACTOS__  ...  #endif` guard, since they are not in Wine.

HBelusca:
Thanks for your PR, but please fix your commit name, as explained in the https://github.com/reactos/reactos/blob/master/CONTRIBUTING.md#rules-and-recommendations
(Reading this file was mentioned during our discussion on mattermost before you started making the PR.)

dll/win32/shell32/wine/pidl.c
```diff
@@ -1722,6 +1722,62 @@ LPITEMIDLIST _ILCreateGuidFromStrA(LPCSTR szGUID)
...
+    return FALSE;
+}
+
+// If szUrl is '::{GUID}' then returns a GUID
+// otherwise it tries to resolve the shortcut.
+LPITEMIDLIST _ILCreateGuidFromShellUrlW(LPCWSTR szUrl)
+{
+    GUID iid;
+
+    if (szUrl[0] == L':' && szUrl[1] == L':')
```
KRosUser2, line 1764: What if only 1 character? Eg [1] does not exists ?

KRosUser2 (changes requested):
See review comment on [1]

dll/win32/shell32/wine/pidl.c
unia909: what is 1 character? it checks [0] and [1]

HBelusca: What KRosUser2 means, is what happens if for some reason the helper function is called with a one-character NULL-terminated (or just an empty NULL-terminated) string. In that very latter case, only szUrl[0] would be valid (and be the NULL char) while szUrl[1] would be undefined and you could crash the code.

viciious: Their initial point probably was that if the string is empty the szUrl[0] == L':' check will fail and the szUrl[1] == L':' check isn't going to happen

dll/win32/shell32/wine/pidl.c
```diff
@@ -1722,6 +1722,62 @@ LPITEMIDLIST _ILCreateGuidFromStrA(LPCSTR szGUID)
 }
 #endif
 
+#ifdef __REACTOS__
+static const struct {
+    LPCWSTR pszName;
+    LPCGUID guid;
+} ShellShortcuts[] =
```
learn-more, line 1729: Any reason why you don't use `CSIDL_Data` from `dll/win32/shell32/wine/shellpath.c` ?

binarymaster:
Ticket [CORE-16939](https://jira.reactos.org/browse/CORE-16939) looks like what this PR attempts to implement, needs to be added to the description and commits.

binarymaster (changes requested):
Hello,

```
```

```
```

Also please set up your GitHub account accordingly, so maintainers can use Squash & Merge button without getting the author field modified back to nickname by GitHub, it's done in two steps:

1. Set your full name in [GitHub profile settings](https://github.com/settings/profile) (it should match your `user.name` setting)
1. Uncheck `Keep my email addresses private` in [GitHub email settings](https://github.com/settings/emails)

Regards,

binarymaster:
GitHub profile looks fine now, however there are different emails: `ussunk918@gmail.com` is used in commits, and `yrakrulasov@gmail.com` is a squash & merge email, this probably needs to be corrected.

dll/win32/shell32/wine/pidl.c
unia909: Thanks for the tip, I'll upload the changes soon.

binarymaster:
Linking problem:
```
/home/runner/work/reactos/reactos/RosBE-CI/i386/lib/gcc/i686-w64-mingw32/8.4.0/../../../../i686-w64-mingw32/bin/ld: dll/win32/shell32/CMakeFiles/shell32.dir/folders/CDesktopFolder.cpp.obj: in function `ZN14CDesktopFolder16ParseDisplayNameEP6HWND__P8IBindCtxPwPmPP11_ITEMIDLISTS5_@28':
/home/runner/work/reactos/reactos/build/../src/dll/win32/shell32/folders/CDesktopFolder.cpp:335: undefined reference to `_ILCreateGuidFromShellUrlW'
```

dll/win32/shell32/wine/pidl.c
```diff
@@ -1682,12 +1682,24 @@ LPITEMIDLIST _ILCreateBitBucket(void)
     return _ILCreateGuid(PT_GUID, &CLSID_RecycleBin);
 }
 
+LPITEMIDLIST _ILCreateFonts(void)
```
HBelusca, line 1685: If those two new functions are _not_ in Wine's shell32 as well, they should certainly be surrounded inside a `#ifdef __REACTOS__  .... #endif` block.

dll/win32/shell32/wine/shellpath.c
```diff
@@ -3155,6 +3171,18 @@ HRESULT WINAPI SHGetFolderLocation(
             *ppidl = _ILCreateNetwork();
             break;
 
+        case CSIDL_FONTS:
```
HBelusca, line 3174: If those three new cases are not in original Wine shell32 code, those should be placed inside a `#ifdef __REACTOS__` block.

dll/win32/shell32/wine/pidl.h
```diff
@@ -268,6 +268,7 @@ LPITEMIDLIST	_ILCreateGuid(PIDLTYPE type, REFIID guid) DECLSPEC_HIDDEN;
 /* Like _ILCreateGuid, but using the string szGUID. */
 LPITEMIDLIST	_ILCreateGuidFromStrA(LPCSTR szGUID) DECLSPEC_HIDDEN;
 LPITEMIDLIST	_ILCreateGuidFromStrW(LPCWSTR szGUID) DECLSPEC_HIDDEN;
+LPITEMIDLIST    _ILCreateGuidFromShellUrlW(LPCWSTR szUrl) DECLSPEC_HIDDEN;
```
HBelusca, line 271: This needs to be inside a `#ifdef __REACTOS__` block.

dll/win32/shell32/wine/shellpath.c
```diff
@@ -1110,14 +1110,14 @@ static const CSIDL_DATA CSIDL_Data[] =
     { /* 0x03 - CSIDL_CONTROLS (.CPL files) */
         &FOLDERID_ControlPanelFolder,
         CSIDL_Type_SystemPath,
-        NULL,
+        L"ControlPanelFolder",
```
learn-more, line 1113: This entire struct is updated on the wine side, including `.parsing` entries, possibly they have another fix for this issue?

unia909: Wine 8.0 which I tried from the Debian repositories does not recognize any shell links like `shell:::{GUID}` or eg. `shell:Personal`. Updating this structure from wine does not make sense, since the code using the new fields also needs to be transferred, and this is not my goal.

HBelusca:
@katahiromz @whindsaks Could you guys have a look at this PR?

katahiromz:
This PR is superseded.

PR #5217, opened 2023-04-03
[KDGDB] Macro for converting a ULPTR64 to a ULONG64

johannesthoma:
This depends on the architecture and compiler. If this is 32 bit and gcc / clang ULPTR64 is a union which cannot be simply cast to an integer value. The macro is an architecture independent way of converting a ULPTR64 to a ULONG64.

This should fix a compilation error when configuring with:

    -D_WINKD_:BOOL=TRUE -DGDB:BOOL=TRUE -DSEPARATE_DBG:BOOL=TRUE -DKDBG:BOOL=FALSE

as documented on https://reactos.org/wiki/GDB

## Purpose

ReactOS currently does not build with

    -D_WINKD_:BOOL=TRUE -DGDB:BOOL=TRUE -DSEPARATE_DBG:BOOL=TRUE -DKDBG:BOOL=FALSE

on 32 bit / gcc. This patch should fix that. Moreover it introduces a macro which can be used in
future references of ULPTR64 (so one does not need to #ifdef in the .c file)

### DIFF
--- a/drivers/base/kdgdb/kdpacket.c
+++ b/drivers/base/kdgdb/kdpacket.c
@@ -259,8 +259,8 @@ GetVersionSendHandler(
     RtlCopyMemory(&KdVersion, &State->u.GetVersion64, sizeof(KdVersion));
     DebuggerDataList = (LIST_ENTRY*)(ULONG_PTR)KdVersion.DebuggerDataList;
     KdDebuggerDataBlock = CONTAINING_RECORD(DebuggerDataList->Flink, KDDEBUGGER_DATA64, Header.List);
-    ProcessListHead = (LIST_ENTRY*)(ULONG_PTR)KdDebuggerDataBlock->PsActiveProcessHead;
-    ModuleListHead = (LIST_ENTRY*)(ULONG_PTR)KdDebuggerDataBlock->PsLoadedModuleList;
+    ProcessListHead = (LIST_ENTRY*)(ULONG_PTR)ULPTR64_TO_ULONG64(KdDebuggerDataBlock->PsActiveProcessHead);
+    ModuleListHead = (LIST_ENTRY*)(ULONG_PTR)ULPTR64_TO_ULONG64(KdDebuggerDataBlock->PsLoadedModuleList);
 
     /* Now we can get the context for the current state */
     KdpSendPacketHandler = NULL;
--- a/sdk/include/psdk/wdbgexts.h
+++ b/sdk/include/psdk/wdbgexts.h
@@ -182,9 +182,11 @@ typedef union _ULPTR64
     ULONG_PTR ptr;
     ULONG64 ptr64;
 } ULPTR64;
+#define ULPTR64_TO_ULONG64(x) (x).ptr64
 #else
 // #define ULPTR64 PVOID64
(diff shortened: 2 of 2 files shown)


HBelusca:
> If this is 32 bit and gcc / clang ULPTR64 is a union which cannot be simply cast to an integer value.

Then why do the existing builders succeed in building the code?

johannesthoma:
Hi, I don't know which configurations the builders build.

But on 32-bit gcc *and* 

    -D_WINKD_:BOOL=TRUE -DGDB:BOOL=TRUE -DSEPARATE_DBG:BOOL=TRUE -DKDBG:BOOL=FALSE commit 

ce562727b8c566 does not build. The error message is:
```
../drivers/base/kdgdb/kdpacket.c: In function 'GetVersionSendHandler':
../drivers/base/kdgdb/kdpacket.c:262:5: error: aggregate value used where an int
eger was expected
     ProcessListHead = (LIST_ENTRY*)(ULONG_PTR)(KdDebuggerDataBlock->PsActivePro
cessHead);
     ^~~~~~~~~~~~~~~
../drivers/base/kdgdb/kdpacket.c:263:5: error: aggregate value used where an int
eger was expected
     ModuleListHead = (LIST_ENTRY*)(ULONG_PTR)(KdDebuggerDataBlock->PsLoadedModu
leList);
     ^~~~~~~~~~~~~~

```

With my patch building succeeds also for this configuration.

johannesthoma:
I introduced the macro to avoid #ifdefs in the C source code where this
is referenced.

johannesthoma:
Thanks for the fast reaction by the way.
:)

HBelusca:
> Ah ok, just read your message.
> 
> ```
> ModuleListHead = (PLIST_ENTRY)*(ULONG_PTR*)&KdDebuggerData->PsLoadedModuleList;
> ```
> 
> does something different than:
> 
> ```
> ModuleListHead = (PLIST_ENTRY)*(ULONG_PTR*)KdDebuggerData->PsLoadedModuleList.ptr64;
> ```
> 
> (which would be the macro expansion).
> 
> The pointer value would differ when you do &KdDebuggerData->PsLoadedModuleList. It compiles because there are casts but it probably would crash at runtime since &KdDebuggerData->PsLoadedModuleList points to somewhere else than KdDebuggerData->PsLoadedModuleList.ptr64.

Note that the original code does:
```c
ModuleListHead = (LIST_ENTRY*)(ULONG_PTR)KdDebuggerDataBlock->PsLoadedModuleList;
```
(with `LIST_ENTRY* === PLIST_ENTRY`), while the short version I proposed (that I claimed worked, because I actually tested it and got the correct values), was:
```c
ModuleListHead = (PLIST_ENTRY)*(ULONG_PTR*)&KdDebuggerData->PsLoadedModuleList; 
```
This is actually the same thing. (Notice where the `*` operations are.)

johannesthoma:
Ok sorry, I did not see the extra '*' in your code. Personally I find the
macro solution cleaner and more easy to read but as ever you wish
:)
Best regards,

- Johannes

HBelusca:
A fix is in master, but thanks for having reported this issue nonetheless!

PR #5221, opened 2023-04-04
[NTOS:OB] Implement ObjectTypesInformation information query case

GeoB99:
This implements the `ObjectTypesInformation` query case, which queries the type info of every defined object type pointed by `ObpObjectTypes`. The number of existing object types is counted and given to caller in form of a `OBJECT_TYPES_INFORMATION` structure. For more information, check the commits descriptions.


GeoB99:
**Windows Server 2003 SP2 (32-bit)**
Capture
**ReactOS (32-bit)**
Capture2
**ReactOS BEFORE (32-bit -- `ntdll_winetest.exe om`**
Capture4
**ReactOS AFTER (32-bit -- `ntdll_winetest.exe om`**
Capture3
**ReactOS (64-bit -- `ntdll_winetest.exe om`**
Capture5

For whatever reason I'm getting some NULL or empty type strings in AMD64 build of ROS...

sdk/include/ndk/obtypes.h
```diff
@@ -307,11 +307,10 @@ typedef struct _OBJECT_TYPE_INFORMATION
     ULONG DefaultNonPagedPoolCharge;
 } OBJECT_TYPE_INFORMATION, *POBJECT_TYPE_INFORMATION;
 
-typedef struct _OBJECT_ALL_TYPES_INFORMATION
+typedef struct _OBJECT_TYPES_INFORMATION
 {
     ULONG NumberOfTypes;
-    //OBJECT_TYPE_INFORMATION TypeInformation[1];
-} OBJECT_ALL_TYPES_INFORMATION, *POBJECT_ALL_TYPES_INFORMATION;
+} OBJECT_TYPES_INFORMATION, *POBJECT_TYPES_INFORMATION;
```
HBelusca, line 313:
```suggestion
    /* Followed by an array of OBJECT_TYPE_INFORMATION */
} OBJECT_TYPES_INFORMATION, *POBJECT_TYPES_INFORMATION;
```

ntoskrnl/ob/oblife.c
```diff
@@ -1558,6 +1558,9 @@ NtQueryObject(IN HANDLE ObjectHandle,
     POBJECT_HEADER ObjectHeader = NULL;
     POBJECT_HANDLE_ATTRIBUTE_INFORMATION HandleFlags;
     POBJECT_BASIC_INFORMATION BasicInfo;
+    POBJECT_TYPES_INFORMATION AllTypesInfo;
+    POBJECT_TYPE_INFORMATION ObjectTypeInfo;
+    ULONG ObjectTypeIndex;
```
HBelusca, line 1563: By curly-bracketing the whole "case" you are implementing, you could move these declarations inside the case-block.

ntoskrnl/ob/oblife.c
```diff
@@ -1722,9 +1725,60 @@ NtQueryObject(IN HANDLE ObjectHandle,
...
+                {
+                    if (ObpObjectTypes[ObjectTypeIndex] != NULL)
+                    {
+                        Status = ObQueryTypeInfo(ObpObjectTypes[ObjectTypeIndex],
+                                                 ObjectTypeInfo,
+                                                 Length,
+                                                 &InfoLength);
+                        if (NT_SUCCESS(Status))
+                        {
+                            /* We have enough space to hold this type, move the pointer to the next spot */
```
HBelusca, line 1775: In case of failure, what does the ObQueryTypeInfo function do? Does it continue increasing the InfoLength accordingly so that one gets a total needed buffer size at the end?

ntoskrnl/ob/oblife.c
GeoB99: I would rather want to keep the current code structure style consistent with the rest of the `NtQueryObject` implementation, than diverging from the said current style. It would make sense to curly-bracketing all the information cases and have specific info case variables on each case body in a separate PR.

ntoskrnl/ob/oblife.c
GeoB99: `ObQueryTypeInfo` will fail with `STATUS_INFO_LENGTH_MISMATCH` if the caller did not supply an adequate amount of buffer space to hold every type info. The algorithm here will keep querying for type info for each object type and the function will increase the `InfoLength` variable upon calculation so that the caller can use the returned length to allocate a pool of memory (or a heap or whatever) for the buffer to do a successful `ObjectTypesInformation` query case.

GeoB99:
I will revisit this information class later so I'm going to close this and split the `ObQueryTypeInfo` commit into a new PR instead.

PR #5224, opened 2023-04-05
[NTDLL:LDR] Rename parameter "SearchPath" to "DllSearchPath"

RatinCN:
## Purpose

Rename parameter "SearchPath" to "DllSearchPath".

`SearchPath` is a macro defined as `SearchPathA` or `SearchPathW`, improper to be used as a parameter name

JIRA issue: None

## Proposed changes

"SearchPath" has renamed to "DllSearchPath".

### DIFF
--- a/dll/ntdll/ldr/ldrapi.c
+++ b/dll/ntdll/ldr/ldrapi.c
@@ -307,7 +307,7 @@ LdrLockLoaderLock(IN ULONG Flags,
 NTSTATUS
 NTAPI
 DECLSPEC_HOTPATCH
-LdrLoadDll(IN PWSTR SearchPath OPTIONAL,
+LdrLoadDll(IN PWSTR DllSearchPath OPTIONAL,
            IN PULONG DllCharacteristics OPTIONAL,
            IN PUNICODE_STRING DllName,
            OUT PVOID *BaseAddress)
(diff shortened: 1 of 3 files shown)


learn-more:
That macro should not be defined in these areas, so I am unsure why this is a problem?

RatinCN:
> That macro should not be defined in these areas, so I am unsure why this is a problem?

But the macro is defined in this scope, if use `SearchPath` inside function and compile, `SearchPath` will be evaluated as a function pointer of `SearchPathA`:
image

learn-more:
Not sure what you are doing, but neither on the builders nor locally do I encounter this issue?

Are you adding non-ntdll includes to this file, or using this file directly in a different project?

RatinCN:
> Not sure what you are doing, but neither on the builders nor locally do I encounter this issue?
> 
> Are you adding non-ntdll includes to this file, or using this file directly in a different project?

Use macro as parameter is not a good practice, that's all.
The reply above just a way to prove the macro `#define SearchPath SearchPathA` is take effect in this scope, to answer your question "That macro should not be defined in these areas".

RatinCN:
As mark and hbelusca said in chat, that code (Ldr) is part of NT and should not worry about what the win32 subsystem does, or that macro should not be defined in these areas, rename it is not an improvement, PR shoud be close.

PR #5230, opened 2023-04-07
[ADVAPI32] Fix RegCreateKeyExW with lpSecurityAttributes->nLength of 0

Doug-Lyons:
## Purpose

_Provide graceful handling of RegCreateKeyExW when lpSecurityAttributes->nLength is zero._

JIRA issue: [CORE-15471](https://jira.reactos.org/browse/CORE-15471)

## Proposed changes

_When lpSecurityAttributes->nLength == 0 then set lpSecurityAttributes to NULL and continue._
_Provide error message showing what was found._

This allows the Win98 DDK to install without any failures.

### DIFF
--- a/dll/win32/advapi32/reg/reg.c
+++ b/dll/win32/advapi32/reg/reg.c
@@ -1108,9 +1108,6 @@ RegCreateKeyExW(
 
     TRACE("RegCreateKeyExW() called\n");
 
-    if (lpSecurityAttributes && lpSecurityAttributes->nLength != sizeof(SECURITY_ATTRIBUTES))
-        return ERROR_INVALID_USER_BUFFER;
-
     /* get the real parent key */
     Status = MapDefaultKey(&ParentKey,
                            hKey);


dll/win32/advapi32/reg/reg.c
```diff
@@ -1108,6 +1108,12 @@ RegCreateKeyExW(
 
     TRACE("RegCreateKeyExW() called\n");
 
+    if (lpSecurityAttributes && lpSecurityAttributes->nLength == 0)
```
HBelusca, line 1111: Maybe the question is whether the **other** check (line 1117 that returns ERROR_INVALID_USER_BUFFER) is actually correct.
(The reason I'm asking is that it would appear that's the only function that does such check, while all others just expect the kernel to do the verification.)
Note that this was added in commit https://github.com/reactos/reactos/commit/cc0716b65d18871f3124a02dcbb2c717a6c38c96 .

JoachimHenze:
Can we prove with a test that this change is correct? (instead of potentially just being a workaround)

Doug-Lyons:
> Can we prove with an added unittest that this change is correct? (instead of potentially just being a workaround)

Let me try and add another test to the advapi32_apitest in RegEnumKey. It should be easy to do.

dll/win32/advapi32/reg/reg.c
Doug-Lyons: See new tests in [CORE-15471](https://jira.reactos.org/browse/CORE-15471). Seems to prove that this is correct.

dll/win32/advapi32/reg/reg.c
```diff
@@ -1108,6 +1108,12 @@ RegCreateKeyExW(
...
+    {
+        lpSecurityAttributes = NULL;
```
GeoB99, line 1113: The caller is expected to properly set the length of the security attributes structure to `sizeof(SECURITY_ATTRIBUTES)` but if they don't do that then too bad, they're lazy. But the problem here now is that you are totally ignoring the security descriptor that was passed by the caller if they did not set the right structure size, thus the newly created key will get the default security descriptor instead of the one set up by the caller. Which is a hack.

Just remove the line below that returns `ERROR_INVALID_USER_BUFFER` -- it is completely pointless. `SECURITY_ATTRIBUTES` is a Win32 construct, the NT kernel only cares about a security descriptor and it is responsible to validate all of its contents in memory.

GeoB99 (changes requested):

dll/win32/advapi32/reg/reg.c
Doug-Lyons: Thanks, @GeoB99. That's an even better solution. Fewer lines of code and simpler. DONE!

dll/win32/advapi32/reg/reg.c
Doug-Lyons: @HBelusca, it seems that @GeoB99 was able to make me understand what your original idea was here about whether the __other__ check was actually correct. See what you think of the latest changes. It passes my new test as well. Thanks.

HBelusca:
> I have added a proof-of-concept code stand-alone test program in [CORE-15471](https://jira.reactos.org/browse/CORE-15471).
I plan to integrate this into the normal regression tests in another PR soon.

Would be better to have this together with that PR and/or in master first.

GeoB99 (approved):
It looks much better now. Just please squash your commits, edit the commit title to `[ADVAPI32] Do not fail on creating a key if the security attributes has invalid length` or something similar, provide an explanation in the description of why the change is done like that and incorporate the testcase that proves `RegCreateKeyExW` doesn't fail with a length of the security attributes set to anything else other than `sizeof(SECURITY_ATTRIBUTES)`.

HBelusca:
> It looks much better now. Just please squash your commits, edit the commit title to `[ADVAPI32] Do not fail on creating a key if the security attributes has invalid length` or something similar [...]

and add that this reverts commit `cc0716b65` (r24491).

JoachimHenze (approved):

HBelusca (approved):

Doug-Lyons:
Thanks, @JoachimHenze, @GeoB99, and @HBelusca for all of your help with this and getting it merged.

ThFabba:
> and add that this reverts commit `cc0716b65` (r24491).

For future reference, this fact should typically make us hesitate and wait for a test. If we make a change and revert it, both without tests to back them up, we may just end up in an infinite cycle.
The original change had a reason behind it, the revert has a reason behind it -- this could mean that neither is quite correct. A test will make sure we don't introduce regressions.

PR #5235, opened 2023-04-09
[SHELL32] Move to/Copy to folder should not be displayed when unable to retrieve the file name

TAN-Gaming:
## Purpose
JIRA issue: [CORE-18909](https://jira.reactos.org/browse/CORE-18909)

### DIFF
--- a/dll/win32/shell32/CCopyToMenu.cpp
+++ b/dll/win32/shell32/CCopyToMenu.cpp
@@ -192,7 +192,7 @@ HRESULT CCopyToMenu::DoRealCopy(LPCMINVOKECOMMANDINFO lpici, LPCITEMIDLIST pidl)
 
 CStringW CCopyToMenu::DoGetFileTitle()
 {
-    CStringW ret = L"(file)";
+    CStringW ret;
 
     CDataObjectHIDA pCIDA(m_pDataObject);
     if (FAILED_UNEXPECTEDLY(pCIDA.hr()))
@@ -240,6 +240,9 @@ HRESULT CCopyToMenu::DoCopyToFolder(LPCMINVOKECOMMANDINFO lpici)
     }
 
     CStringW strFileTitle = DoGetFileTitle();
+    if (strFileTitle.IsEmpty())
+        return hr;
+
     CStringW strTitle;
     strTitle.Format(IDS_COPYTOTITLE, static_cast<LPCWSTR>(strFileTitle));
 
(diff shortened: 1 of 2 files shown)


KRosUser2 (approved):
Lgtm, better than my patch

dll/win32/shell32/CCopyToMenu.cpp
```diff
@@ -192,7 +192,7 @@ HRESULT CCopyToMenu::DoRealCopy(LPCMINVOKECOMMANDINFO lpici, LPCITEMIDLIST pidl)
 
 CStringW CCopyToMenu::DoGetFileTitle()
 {
-    CStringW ret = L"(file)";
+    CStringW ret; // Empty
```
learn-more, line 195: What is the point of the comment?

TAN-Gaming: It's easier to understand the code: returns empty string in case of failure.

TAN-Gaming: Removed.

JoachimHenze (approved):
LGTM now. Retval hr is preintialized to E_FAIL in both cases.

dll/win32/shell32/CCopyToMenu.cpp
```diff
@@ -240,6 +240,9 @@ HRESULT CCopyToMenu::DoCopyToFolder(LPCMINVOKECOMMANDINFO lpici)
     }
 
     CStringW strFileTitle = DoGetFileTitle();
+    if (strFileTitle.IsEmpty())
```
learn-more, line 243: I am not convinced that the _title_ of a dialog should dictate whether or not we can copy or move a file.

KRosUser2: @learn-more , we refer to file title (name without extension) not dialog title... Too fast in your comments guy !

KRosUser2: Ok so everything is fine.

learn-more: You didn't convince me with your weird arguments, so no, I am still not convinced that the title used for a dialog should be used to check if we can copy a file.

learn-more (changes requested):
I am not convinced that something that is used as title of a dialog should dictate whether or not we can copy or move a file.

I suspect that it should be more something like is being done in `CDefaultContextMenu::QueryContextMenu`, where actual attributes (https://learn.microsoft.com/en-us/windows/win32/shell/sfgao) are being checked.

KRosUser2:
@TAN-Gaming : Reference to my patch to be removed from the PR description as it's no longer being used nor the best option.

TAN-Gaming:
Maybe next time.

PR #5238, opened 2023-04-13
Fix kernel crash, when copying large files

tkreuzer:
## Purpose

This PR fixes the kernel crash, when copying large files.
Suggested to be reviewed by commit.

## Details
There are multiple attempts to fix the bug:
- PR #4739 by Jerome (releases VACBs on memory pressure) (currently included)
- [trim_cache_patch_2.patch](https://jira.reactos.org/secure/attachment/65635/65635_trim_cache_patch_2.patch) by Thomas (well, trims individual cache pages, when needed)
- PR #5237 (based on trim_cache_patch_2)
- [ntos_mm_higher_bottom.patch](https://jira.reactos.org/secure/attachment/65511/65511_ntos_mm_higher_bottom.patch)  (only starts paging earlier)
- [modified_standby_pages_3.diff](https://jira.reactos.org/secure/attachment/65559/65559_modified_standby_pages_3.diff) by Jerome (lets the page fault handler search for standby pages. This might help, but isn't entirely correct either, because there is no guarantee, that there are any standby pages available and the code cannot page out any more pages because it runs at `DISPATCH_LEVEL`, so it can still randomly hang the system entirely)

None of them fixes the problem, that the page fault handler can randomly bugcheck, when temporarily running out of pages. Instead they just hide the underlying problem, by making it's occurrence really unlikely.
They are partly still valuable though, as they can improve the situation (in fact either PR #4739 (currently included) or trim_cache_path / #5237 is needed for this PR to fully work).
This PR fixes the issue by letting the page fault handler wait for new pages at `PASSIVE_LEVEL`. This allows the balancer to page out more pages and doesn't stall the system unneccessarily. Additionally it fixes a number of subtle bugs that are only triggered on high memory pressure.
This fixes the file copy crash and also any random crashes, when temporarily running out of free/standby pages. It can still hang, when more user memory is committed than is available in both physical memory and the page file, but fixing this requires to implement commitment accounting.

## Proposed changes

- Make sure the balancer thread doesn't bugcheck, when there is nothing to page out.
- Implement ShareCount handling for RosMm
- Fail gracefully in the PFN allocation functions, when there are no more pages available.
- Implement `MmRebalanceMemoryConsumersAndWait()` to run the balancer on demand and wait for it to (hopefully) return free pages.
- Use `MmRebalanceMemoryConsumersAndWait()` in the page fault handler, when no page could be allocated and retry page fault handling.
- Fix locking and a race condition between the balancer and the section code, when handling RMAPs.

See commit messages for more details.

This PR includes PR #4739 by Jerome, which makes the whole thing work by giving back cache pages, when there is need for memory. Without the PR the crash is fixed, but the system will run out of pages and then hang, when the balancer cannot free anything anymore.

Note that this fixes copying, but it doesn't entirely prevent running out of pages. The problem is that we do not track committed pages and can therefore overcommit (i.e. commit more pages than are together in physical RAM plus page file) and eventually run out of pages and pagefile space, at which point the system would hang.

[ReactOS total commander copy large file.webm](https://user-images.githubusercontent.com/313067/231837590-25d746d2-e630-41c7-b656-6031912a9e6e.webm)

## TODO
- [ ] Run on test bots
- [ ] Have some people test it and check performance.

## More work
These things still need to be done to fix paging entirely, but are not within the scope of this PR
- Implement accounting (respect system commit limit)
- Implement working set handling for ARM³
- Migrate rest of RosMm PFN handling to ARM³
- Migrate file sections to ARM³

### DIFF
--- a/ntoskrnl/cc/copy.c
+++ b/ntoskrnl/cc/copy.c
@@ -381,6 +381,16 @@ CcCanIWrite (
         return FALSE;
     }
 
+    /* Otherwise, if there are no deferred writes yet, start the lazy writer */
+    if (IsListEmpty(&CcDeferredWrites))
+    {
+        KIRQL OldIrql;
+
+        OldIrql = KeAcquireQueuedSpinLock(LockQueueMasterLock);
+        CcScheduleLazyWriteScan(TRUE);
+        KeReleaseQueuedSpinLock(LockQueueMasterLock, OldIrql);
+    }
+
     /* Initialize our wait event */
     KeInitializeEvent(&WaitEvent, NotificationEvent, FALSE);
 
@@ -407,12 +417,6 @@ CcCanIWrite (
                                     &Context.DeferredWriteLinks,
                                     &CcDeferredWriteSpinLock);
     }
-
-    /* Now make sure that the lazy scan writer will be active */
-    OldIrql = KeAcquireQueuedSpinLock(LockQueueMasterLock);
-    if (!LazyWriter.ScanActive)
-        CcScheduleLazyWriteScan(TRUE);
-    KeReleaseQueuedSpinLock(LockQueueMasterLock, OldIrql);
 
 #if DBG
     DPRINT1("Actively deferring write for: %p\n", FileObject);
--- a/ntoskrnl/cc/view.c
+++ b/ntoskrnl/cc/view.c
@@ -595,34 +595,26 @@ CcRosUnmarkDirtyVacb (
     }
 }
 
-static
 BOOLEAN
-CcRosFreeUnusedVacb (
-    PULONG Count)
+CcRosFreeOneUnusedVacb(
+    VOID)
 {
-    ULONG cFreed;
-    BOOLEAN Freed;
     KIRQL oldIrql;
-    PROS_VACB current;
-    LIST_ENTRY FreeList;
     PLIST_ENTRY current_entry;
-
-    cFreed = 0;
-    Freed = FALSE;
-    InitializeListHead(&FreeList);
+    PROS_VACB to_free = NULL;
 
     oldIrql = KeAcquireQueuedSpinLock(LockQueueMasterLock);
 
     /* Browse all the available VACB */
     current_entry = VacbLruListHead.Flink;
-    while (current_entry != &VacbLruListHead)
+    while ((current_entry != &VacbLruListHead) && (to_free == NULL))
     {
         ULONG Refs;
+        PROS_VACB current;
 
         current = CONTAINING_RECORD(current_entry,
                                     ROS_VACB,
                                     VacbLruListEntry);
-        current_entry = current_entry->Flink;
 
         KeAcquireSpinLockAtDpcLevel(&current->SharedCacheMap->CacheMapLock);
 
@@ -634,47 +626,32 @@ CcRosFreeUnusedVacb (
             ASSERT(!current->MappedCount);
             ASSERT(Refs == 1);
 
-            /* Reset and move to free list */
+            /* Reset it, this is the one we want to free */
             RemoveEntryList(&current->CacheMapVacbListEntry);
+            InitializeListHead(&current->CacheMapVacbListEntry);
             RemoveEntryList(&current->VacbLruListEntry);
             InitializeListHead(&current->VacbLruListEntry);
-            InsertHeadList(&FreeList, &current->CacheMapVacbListEntry);
+
+            to_free = current;
         }
 
         KeReleaseSpinLockFromDpcLevel(&current->SharedCacheMap->CacheMapLock);
 
+        current_entry = current_entry->Flink;
     }
 
     KeReleaseQueuedSpinLock(LockQueueMasterLock, oldIrql);
 
-    /* And now, free any of the found VACB, that'll free memory! */
-    while (!IsListEmpty(&FreeList))
+    /* And now, free the VACB that we found, if any. */
+    if (to_free == NULL)
     {
-        ULONG Refs;
-
-        current_entry = RemoveHeadList(&FreeList);
-        current = CONTAINING_RECORD(current_entry,
-                                    ROS_VACB,
-                                    CacheMapVacbListEntry);
-        InitializeListHead(&current->CacheMapVacbListEntry);
-        Refs = CcRosVacbDecRefCount(current);
-        ASSERT(Refs == 0);
-        ++cFreed;
+        return FALSE;
     }
 
-    /* If we freed at least one VACB, return success */
-    if (cFreed != 0)
-    {
-        Freed = TRUE;
-    }
+    /* This must be its last ref */
+    NT_VERIFY(CcRosVacbDecRefCount(to_free) == 0);
 
-    /* If caller asked for free count, return it */
-    if (Count != NULL)
-    {
-        *Count = cFreed;
-    }
-
-    return Freed;
+    return TRUE;
 }
 
 static
@@ -690,7 +667,6 @@ CcRosCreateVacb (
     NTSTATUS Status;
     KIRQL oldIrql;
     ULONG Refs;
-    BOOLEAN Retried;
     SIZE_T ViewSize = VACB_MAPPING_GRANULARITY;
 
     ASSERT(SharedCacheMap);
@@ -711,28 +687,24 @@ CcRosCreateVacb (
 
     CcRosVacbIncRefCount(current);
 
-    Retried = FALSE;
-Retry:
-    /* Map VACB in system space */
-    Status = MmMapViewInSystemSpaceEx(SharedCacheMap->Section, &current->BaseAddress, &ViewSize, &current->FileOffset, 0);
-
-    if (!NT_SUCCESS(Status))
+    while (TRUE)
     {
-        ULONG Freed;
-        /* If no space left, try to prune unused VACB
-         * to recover space to map our VACB
-         * If it succeed, retry to map, otherwise
-         * just fail.
-         */
-        if (!Retried && CcRosFreeUnusedVacb(&Freed))
+        /* Map VACB in system space */
+        Status = MmMapViewInSystemSpaceEx(SharedCacheMap->Section, &current->BaseAddress, &ViewSize, &current->FileOffset, 0);
+        if (NT_SUCCESS(Status))
         {
-            DPRINT("Prunned %d VACB, trying again\n", Freed);
-            Retried = TRUE;
-            goto Retry;
+            break;
         }
 
-        ExFreeToNPagedLookasideList(&VacbLookasideList, current);
-        return Status;
+        /*
+         * If no space left, try to prune one unused VACB to recover space to map our VACB.
+         * If it succeeds, retry to map, otherwise just fail.
+         */
+        if (!CcRosFreeOneUnusedVacb())
+        {
+            ExFreeToNPagedLookasideList(&VacbLookasideList, current);
+            return Status;
+        }
     }
 
 #if DBG
--- a/ntoskrnl/include/internal/cc.h
+++ b/ntoskrnl/include/internal/cc.h
@@ -509,3 +509,7 @@ CcRosVacbDecRefCount(
 }
 #define CcRosVacbGetRefCount(vacb) InterlockedCompareExchange((PLONG)&(vacb)->ReferenceCount, 0, 0)
 #endif
+
+BOOLEAN
+CcRosFreeOneUnusedVacb(
+    VOID);
--- a/ntoskrnl/include/internal/mm.h
+++ b/ntoskrnl/include/internal/mm.h
@@ -1135,6 +1135,14 @@ MmCreateVirtualMappingUnsafe(
     PFN_NUMBER Page
 );
 
+NTSTATUS
+NTAPI
+MmCreatePhysicalMapping(
+    _Inout_opt_ PEPROCESS Process,
+    _In_ PVOID Address,
+    _In_ ULONG flProtect,
+    _In_ PFN_NUMBER Page);
+
 ULONG
 NTAPI
 MmGetPageProtect(
@@ -1291,9 +1299,18 @@ MmGetExecuteOptions(IN PULONG ExecuteOptions);
 _Success_(return)
 BOOLEAN
 MmDeleteVirtualMapping(
-    _In_opt_ PEPROCESS Process,
+    _Inout_opt_ PEPROCESS Process,
     _In_ PVOID Address,
     _Out_opt_ BOOLEAN* WasDirty,
+    _Out_opt_ PPFN_NUMBER Page
+);
+
+_Success_(return)
+BOOLEAN
+MmDeletePhysicalMapping(
+    _Inout_opt_ PEPROCESS Process,
+    _In_ PVOID Address,
+    _Out_opt_ BOOLEAN * WasDirty,
     _Out_opt_ PPFN_NUMBER Page
 );
 
--- a/ntoskrnl/mm/ARM3/pagfault.c
+++ b/ntoskrnl/mm/ARM3/pagfault.c
@@ -15,6 +15,10 @@
 #define MODULE_INVOLVED_IN_ARM3
 #include <mm/ARM3/miarm.h>
 
+VOID
+NTAPI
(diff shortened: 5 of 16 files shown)


ntoskrnl/cc/view.c
```diff
@@ -595,34 +595,26 @@ CcRosUnmarkDirtyVacb (
     }
 }
 
-static
 BOOLEAN
-CcRosFreeUnusedVacb (
-    PULONG Count)
+CcRosFreeOneUnusedVacb (
```
JoachimHenze, line 599:
```suggestion
CcRosFreeOneUnusedVacb(
```

ntoskrnl/include/internal/cc.h
```diff
@@ -509,3 +509,7 @@ CcRosVacbDecRefCount(
 }
 #define CcRosVacbGetRefCount(vacb) InterlockedCompareExchange((PLONG)&(vacb)->ReferenceCount, 0, 0)
 #endif
+
+BOOLEAN
+CcRosFreeOneUnusedVacb (
```
JoachimHenze, line 514:
```suggestion
CcRosFreeOneUnusedVacb(
```

ntoskrnl/mm/balance.c
```diff
@@ -245,15 +251,30 @@ MmTrimUserMemory(ULONG Target, ULONG Priority, PULONG NrFreedPages)
...
         }
 
         CurrentPage = MmGetLRUNextUserPage(CurrentPage, TRUE);
+        if (FirstPage == 0)
+        {
+            FirstPage = CurrentPage;
+        }
+        else if (CurrentPage == FirstPage)
+        {
+            DPRINT1("We are back at the start, abort!\n");
```
JoachimHenze, line 275:
```suggestion
            DPRINT("We are back at the start, abort\n");
```

ntoskrnl/mm/section.c
```diff
@@ -1689,7 +1698,7 @@ MmNotPresentFaultSectionView(PMMSUPPORT AddressSpace,
          * Just map the desired physical page
          */
         Page = (PFN_NUMBER)(Offset.QuadPart >> PAGE_SHIFT);
-        Status = MmCreateVirtualMappingUnsafe(Process,
+        Status = MmCreatePhysicalMapping(Process,
                                               PAddress,
```
JoachimHenze, line 1702:
```suggestion
                                         PAddress,
```

ntoskrnl/mm/section.c
```diff
@@ -1689,7 +1698,7 @@ MmNotPresentFaultSectionView(PMMSUPPORT AddressSpace,
...
                                               Region->Protect,
```
JoachimHenze, line 1703:
```suggestion
                                         Region->Protect,
```

ntoskrnl/mm/section.c
```diff
@@ -1689,7 +1698,7 @@ MmNotPresentFaultSectionView(PMMSUPPORT AddressSpace,
...
                                               Page);
```
JoachimHenze, line 1704:
```suggestion
                                         Page);
```

ntoskrnl/cc/view.c
```diff
@@ -711,28 +687,24 @@ CcRosCreateVacb (
...
-            DPRINT("Prunned %d VACB, trying again\n", Freed);
-            Retried = TRUE;
-            goto Retry;
+            break;
         }
 
-        ExFreeToNPagedLookasideList(&VacbLookasideList, current);
-        return Status;
+        /*
+         * If no space left, try to prune  one unused VACB to recover space to map our VACB.
```
JoachimHenze, line 700:
```suggestion
         * If no space left, try to prune one unused VACB to recover space to map our VACB.
```

ntoskrnl/mm/balance.c
tkreuzer: This one is actually worth a DPRINT1, since it really only happens when the balancer struggles to get any new pages. It likely would never show up with the cache patch by Jerome, unless there is too much memory committed.

JoachimHenze: Ok, I restored the 1 in the suggestion.

ntoskrnl/include/internal/cc.h
tkreuzer: In fact I wish we allowed this style, because it allows to distinguish header definitions from function invocations when grepping. This one is from Jerome's PR, btw.

ntoskrnl/mm/mmfault.c
```diff
@@ -252,16 +257,25 @@ MmAccessFault(IN ULONG FaultCode,
         return MmArmAccessFault(FaultCode, Address, Mode, TrapInformation);
     }
 
+Retry:
```
learn-more, line 260: I think this retry block is small enough that it would be better readable to have a while loop here.

ntoskrnl/mm/section.c
tkreuzer: Hint: you can mark multiple lines and suggest a correction for all of them in one comment.

ntoskrnl/mm/mmfault.c
tkreuzer: I don't think that makes it any more readable. It's just an infinite loop with an if/break.

ntoskrnl/mm/section.c
```diff
@@ -1335,9 +1335,16 @@ MmMakeSegmentResident(
...
             {
-                /* MmRequestPageMemoryConsumer succeeds or bugchecks */
-                (void)MmRequestPageMemoryConsumer(MC_USER, FALSE, &Pages[i]);
+                Status = MmRequestPageMemoryConsumer(MC_USER, FALSE, &Pages[i]);
+                if (!NT_SUCCESS(Status))
+                {
+                    /* Damn. Roll-back. */
+                    for (UINT j = 0; j < i; j++)
+                        MmReleasePageMemoryConsumer(MC_USER, Pages[j]);
+                    goto Failed;
```
learn-more, line 1344: Should `Pages` be cleared?

tkreuzer: That's not needed. This uses the same clean-up logic as is used before the "Failed" path, only limited to the number of pages that were actually allocated. After that we only need to free the MDL, which doesn't care about the contents.

ntoskrnl/mm/ARM3/pagfault.c
```diff
@@ -653,7 +657,7 @@ MiResolveDemandZeroFault(IN PVOID Address,
     ASSERT(PointerPte->u.Hard.Valid == 0);
 
     /* Assert we have enough pages */
-    ASSERT(MmAvailablePages >= 32);
+    //ASSERT(MmAvailablePages >= 32);
```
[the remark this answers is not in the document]
tkreuzer: If we correctly counted available pages, the check below is eqivalent to `MmAvailablePages == 0`. And yes at some point we might run out of pages. I thought about bailing out on 32, but the "problem" is that we only have one way to allocate pages anyway (well, 2, but they behave the same), so even if we "reserved" 32 pages, we don't have a path that can allocate them, so why reserve them in the first place? If we have code that requires a certain amount of pages no matter what, even at IRQL >= DISPATCH_LEVEL, then we should fix that by pre-allocating the pages instead of expecting the PFN database to respect other components needs by randomly reserving an arbitrary amount of pages. I haven't seen that happening though.

Daniel7689:
I have followed the steps at https://jira.reactos.org/browse/CORE-17125 after applying this PR and PR #5237. They result in the commands taking significantly less time but end with a segmentation fault error. I have made a video showing what happens. It is in the list of attachments shown at that link.

Daniel7689:
I have noticed an improvement in the file copying part of the setup of ReactOS. Here are the videos showing the before and after:

https://user-images.githubusercontent.com/24794949/233167724-88ef35aa-b4b7-449b-a7f9-591604f96eb7.mp4
https://user-images.githubusercontent.com/24794949/233167727-5c56ecc0-2ae4-457f-af08-c3fcb6a2aca2.mp4

Daniel7689:
I have noticed an improvement in running the benchmarks of CrystalDiskMark 8.0.1. Here are the videos showing the before and after:

Upl

https://user-images.githubusercontent.com/24794949/233233647-eac44b72-1f33-4152-afe0-885fd3f1072b.mp4

oading 2048_MB_RAM_VM_-_CrystalDiskMark_running_benchmarks_-_No_PRs_applied.mp4…

Daniel7689:
I have noticed an improvement in the file copying part of the setup of ReactOS. Here are the videos showing the before and after:


https://user-images.githubusercontent.com/24794949/233235406-98a150ef-c5a9-4202-b87b-22e69b08a6db.mp4

https://user-images.githubusercontent.com/24794949/233235413-dd2629ee-8632-44d4-9225-5976904c5fbb.mp4

Daniel7689:
I have noticed an improvement in running the benchmarks of CrystalDiskMark 8.0.1. Here are the videos showing the before and after:

https://user-images.githubusercontent.com/24794949/233236224-97cc2f3b-dade-475d-91f0-c67125d50544.mp4



https://user-images.githubusercontent.com/24794949/233236227-6587f167-6988-4ff9-89ec-81dd2fe6f4b6.mp4

ntoskrnl/cc/copy.c
```diff
@@ -381,6 +381,16 @@ CcCanIWrite (
         return FALSE;
     }
 
+    /* Otherwise, if there are no deferred writes yet, start the lazy writer */
+    if (IsListEmpty(&CcDeferredWrites))
+    {
+        KIRQL OldIrql;
```
HBelusca, line 387: Not sure you need this local variable, since there is already one available in the function, and the previous acquire-release spinlock has been already done (see lines 331-354).

ntoskrnl/mm/i386/page.c
```diff
@@ -667,6 +699,26 @@ MmCreateVirtualMappingUnsafe(PEPROCESS Process,
     return(STATUS_SUCCESS);
 }
 
+NTSTATUS
+NTAPI
+MmCreateVirtualMappingUnsafe(PEPROCESS Process,
```
HBelusca, line 722: Use SAL annotations (for this function and the other one)?

ntoskrnl/mm/i386/page.c
```diff
@@ -588,10 +615,10 @@ MmCreatePageFileMapping(PEPROCESS Process,
 
 NTSTATUS
 NTAPI
-MmCreateVirtualMappingUnsafe(PEPROCESS Process,
+MmCreateVirtualMappingUnsafeEx(PEPROCESS Process,
                              PVOID Address,
                              ULONG flProtect,
-                             PFN_NUMBER Page)
+                             PFN_NUMBER Page, BOOLEAN IsPhysical)
```
HBelusca, line 632: SAL ?

ntoskrnl/include/internal/mm.h
```diff
@@ -1135,6 +1135,14 @@ MmCreateVirtualMappingUnsafe(
     PFN_NUMBER Page
 );
 
+NTSTATUS
+NTAPI
+MmCreatePhysicalMapping(
+    PEPROCESS Process,
+    PVOID Address,
+    ULONG flProtect,
+    PFN_NUMBER Page);
```
HBelusca, line 1144: SAL?

ntoskrnl/mm/ARM3/pfnlist.c
```diff
@@ -1227,6 +1231,13 @@ MiDecrementReferenceCount(IN PMMPFN Pfn1,
     /* PFN lock must be held */
     MI_ASSERT_PFN_LOCK_HELD();
 
+    /* Hack for RosMm */
```
HBelusca, line 1242: Can you maybe quickly explain what this does in the comment? (the "AWE allocation")

ntoskrnl/include/internal/cc.h
```diff
@@ -509,3 +509,7 @@ CcRosVacbDecRefCount(
...
+    VOID);
```
HBelusca, line 515: Missing end-of-file newline:
```suggestion
    VOID);

```

Daniel7689:
@tkreuzer The first bullet point in the Details section is pointing to the wrong PR. It should be pointing to #4739 instead. Also, your latest comment in https://jira.reactos.org/browse/CORE-17624 should mention #4739 and #5237 instead of #5237 twice.

JoachimHenze (approved):
All my comments have been addressed properly by Timo, and also all that I resolved which were requested by Hermes. Thank you very much Timo! Approved.
Please mention JIRA-17624 in the commit comment when merging, as it is not referenced by any commit atm.

Daniel7689:
@JoachimHenze you forgot to change "in fact either PR #5237" to "in fact either PR #4739" and "This PR includes PR #5237 by Jerome," to "This PR includes PR #4739 by Jerome,"

HBelusca (approved):
Trivial things to fix:
- See Daniel's remarks;
- 4th commit message: `[...] to handle pyhsical memory [...]` : `pyhsical` -> `physical`
- In case the commit `NTOS:CC Free some VACBs [...]` gets merged too: add the brackets around the module name: `[NTOS:CC]`

ntoskrnl/mm/ARM3/pagfault.c
```diff
@@ -1925,7 +1944,7 @@ _WARN("Session space stuff is not implemented yet!")
                 return STATUS_IN_PAGE_ERROR | 0x10000000;
             }
         }
-
+RetryKernel:
```
HBelusca, line 1947:
```suggestion

RetryKernel:
```

ntoskrnl/mm/balance.c
```diff
@@ -276,27 +297,41 @@ MmRebalanceMemoryConsumers(VOID)
...
+        *AllocatedPage = 0;
+        return STATUS_NO_MEMORY;
     }
     *AllocatedPage = Page;
 
+    /* Update the target */
+    InterlockedIncrementUL(&MiMemoryConsumers[Consumer].PagesUsed);
+    UpdateTotalCommittedPages(1);
+
     return(STATUS_SUCCESS);
```
HBelusca, line 335:
```suggestion
    return STATUS_SUCCESS;
```

ntoskrnl/mm/section.c
```diff
@@ -1335,9 +1335,16 @@ MmMakeSegmentResident(
...
             for (UINT i = 0; i < BYTES_TO_PAGES(ReadLength); i++)
             {
-                /* MmRequestPageMemoryConsumer succeeds or bugchecks */
-                (void)MmRequestPageMemoryConsumer(MC_USER, FALSE, &Pages[i]);
+                Status = MmRequestPageMemoryConsumer(MC_USER, FALSE, &Pages[i]);
+                if (!NT_SUCCESS(Status))
+                {
+                    /* Damn. Roll-back. */
+                    for (UINT j = 0; j < i; j++)
+                        MmReleasePageMemoryConsumer(MC_USER, Pages[j]);
```
HBelusca, line 1343:
```suggestion
                    for (UINT j = 0; j < i; j++)
                    {
                        MmReleasePageMemoryConsumer(MC_USER, Pages[j]);
                    }
```

ntoskrnl/mm/i386/page.c
```diff
@@ -560,7 +598,7 @@ MmCreatePageFileMapping(PEPROCESS Process,
     /* And we don't support creating for other process */
     ASSERT(Process == PsGetCurrentProcess());
 
-    if (SwapEntry & (1 << 31))
+    if (SwapEntry & (1 << (sizeof(SWAPENTRY) - 1)))
```
HBelusca, line 601: Same problem might also be present there:
https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/mm/i386/pagepae.c;hb=59e74584ac5c48e9edb2bf494c2224bd5ca6fa90#l1072

JoachimHenze:
> @JoachimHenze you forgot to change "in fact either PR #5237" to "in fact either PR #4739" and "This PR includes PR #5237 by Jerome," to "This PR includes PR #4739 by Jerome,"

addressed now. thx.

zefklop (approved):
That's a tremendous amount of work. Looks good. Thanks for taking care of that.

PR #5240, opened 2023-04-14
[NTOS:KDBG] Only load symbols on x86

tkreuzer:
## Purpose
Fix x64 boot with KDBG enabled (rosautotest).

## Proposed changes
- Only load symbols on x86

### DIFF
--- a/ntoskrnl/kdbg/kdb_symbols.c
+++ b/ntoskrnl/kdbg/kdb_symbols.c
@@ -355,8 +355,10 @@ KdbSymInit(
         SHORT Found = FALSE;
         CHAR YesNo;
 
-        /* By default, load symbols in DBG builds, but not in REL builds */
-#if DBG
+        /* By default, load symbols in DBG builds, but not in REL builds
+           or anything other than x86, because they only work on x86
+           and can cause the system to hang on x64. */
+#if DBG && defined(_M_IX86)
         LoadSymbols = TRUE;
 #else
         LoadSymbols = FALSE;


HBelusca (approved):

HBelusca:
I am suspecting that it fails because the rossym format is "different" on x64, but do you know whether this is the actual reason, or whether it's due to something else?

tkreuzer:
> I am suspecting that it fails because the rossym format is "different" on x64, but do you know whether this is the actual reason, or whether it's due to something else?
> 
> (I also know that there are other problems in symbols loading that could make it fail similarly even on x86.)

This is an MSVC build, so it doesn't have rossym. I tested x86 MSVC build, because I suspected it might also fail, but it doesn't. So maybe it's just not compatible with 64 bit.

ntoskrnl/kdbg/kdb_symbols.c
```diff
@@ -356,7 +356,7 @@ KdbSymInit(
         CHAR YesNo;
 
         /* By default, load symbols in DBG builds, but not in REL builds */
-#if DBG
+#if DBG && defined(_M_IX86)
```
GeoB99, line 359: I would prefer that an explanation in the comments (or just in the commit description) is present alongside with the code change so that people don't have to rely on guesses why symbols loading has to be done in x86 only.

tkreuzer: Done. TLDR: "It doesn't work"

HBelusca (approved):

GeoB99 (approved):

PR #5245, opened 2023-04-16
[EXT2FS] Align with upstream and mark ReactOS diff

Vince1171:
## Purpose
add ifdefs to distinguish ReactOS code/fixes from the upstream code.
NOTE: there are no new features in this PR, we stay with upstream version 0.69.

JIRA issue: [CORE-18645](https://jira.reactos.org/browse/CORE-18645)

## Proposed changes

 * surround ReactOS modifications with ifdef
 * if upstream code was override, reintroduce it in #else section
 * re-align .rc version info with upstream (0.69)

### DIFF
--- a/drivers/filesystems/ext2/inc/linux/module.h
+++ b/drivers/filesystems/ext2/inc/linux/module.h
@@ -79,8 +79,8 @@ FASTCALL
 RtlUlonglongByteSwap(
     IN ULONGLONG Source
 );
-#endif
 #endif //#ifndef __REACTOS__
+#endif
 
 #define __swab16(x) RtlUshortByteSwap(x)
 #define __swab32(x) RtlUlongByteSwap(x)
--- a/drivers/filesystems/ext2/inc/linux/module.h
+++ b/drivers/filesystems/ext2/inc/linux/module.h
@@ -59,7 +59,8 @@ extern "C" {
 #define RtlUlongByteSwap(_x)     _byteswap_ulong((_x))
 #define RtlUlonglongByteSwap(_x) _byteswap_uint64((_x))
 
-#elif !defined(__REACTOS__)
+#else
+#ifndef __REACTOS__
 
 USHORT
 FASTCALL
@@ -79,6 +80,7 @@ RtlUlonglongByteSwap(
     IN ULONGLONG Source
 );
 #endif
+#endif //#ifndef __REACTOS__
 
 #define __swab16(x) RtlUshortByteSwap(x)
 #define __swab32(x) RtlUlongByteSwap(x)
@@ -429,10 +431,18 @@ extern struct task_struct *current;
 //
 
 
+#ifdef __REACTOS__
 static inline int cond_resched() {
+#else
+static inline cond_resched() {
+#endif
     return FALSE;
 }
+#ifdef __REACTOS__
 static inline int need_resched() {
+#else
+static inline need_resched() {
+#endif
     return FALSE;
 }
 
--- a/drivers/filesystems/ext2/inc/linux/types.h
+++ b/drivers/filesystems/ext2/inc/linux/types.h
@@ -27,14 +27,18 @@ typedef	signed   __int64	__s64;
 typedef	unsigned __int64	__u64;
 
 
+#ifdef __REACTOS__
 typedef __s8         s8;
 typedef __u8         u8;
+#endif
+#if defined(_MSC_VER) || defined(__REACTOS__)
 typedef __s16       s16;
 typedef __u16       u16;
 typedef __s32       s32;
 typedef __u32       u32;
 typedef __s64       s64;
 typedef __u64       u64;
+#endif
 
 #define __le16      u16
 #define __le32      u32
--- a/drivers/filesystems/ext2/src/devctl.c
+++ b/drivers/filesystems/ext2/src/devctl.c
@@ -17,7 +17,11 @@ extern PEXT2_GLOBAL Ext2Global;
 
 /* DEFINITIONS *************************************************************/
 
+#ifdef __REACTOS__
 NTSTATUS NTAPI
+#else
+NTSTATUS
+#endif
 Ext2DeviceControlCompletion (
     IN PDEVICE_OBJECT   DeviceObject,
     IN PIRP             Irp,
@@ -38,7 +42,11 @@ Ext2DeviceControlCompletion (
 #endif
 
 
+#ifdef __REACTOS__
 NTSTATUS NTAPI
+#else
+NTSTATUS
+#endif
 Ext2DeviceControlCompletion (
     IN PDEVICE_OBJECT   DeviceObject,
     IN PIRP             Irp,
@@ -426,14 +434,22 @@ Ext2ProcessVolumeProperty(
         case APP_CMD_SET_PROPERTY2:
 
             RtlZeroMemory(Vcb->sHidingPrefix, HIDINGPAT_LEN);
+#ifdef __REACTOS__
             if (Vcb->bHidingPrefix == Property2->bHidingPrefix) {
+#else
+            if (Vcb->bHidingPrefix = Property2->bHidingPrefix) {
+#endif
                 RtlCopyMemory( Vcb->sHidingPrefix,
                                Property2->sHidingPrefix,
                                HIDINGPAT_LEN - 1);
             }
 
             RtlZeroMemory(Vcb->sHidingSuffix, HIDINGPAT_LEN);
+#ifdef __REACTOS__
             if (Vcb->bHidingSuffix == Property2->bHidingSuffix) {
+#else
+            if (Vcb->bHidingSuffix = Property2->bHidingSuffix) {
+#endif
                 RtlCopyMemory( Vcb->sHidingSuffix,
                                Property2->sHidingSuffix,
                                HIDINGPAT_LEN - 1);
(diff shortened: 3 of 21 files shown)


drivers/filesystems/ext2/inc/linux/module.h
```diff
@@ -79,6 +80,7 @@ RtlUlonglongByteSwap(
     IN ULONGLONG Source
 );
 #endif
```
JoachimHenze, line 82:
```suggestion
#endif //#ifndef __REACTOS__
```

drivers/filesystems/ext2/src/devctl.c
```diff
@@ -511,7 +527,11 @@ Ext2ProcessVolumeProperty(
             RtlCopyMemory(Property2->UUID, Vcb->SuperBlock->s_uuid, 16);
             Property2->DrvLetter = Vcb->DrvLetter;
 
+#ifdef __REACTOS__
             if (Property2->bHidingPrefix == Vcb->bHidingPrefix) {
+#else
+            if (Property2->bHidingPrefix = Vcb->bHidingPrefix) {
```
JoachimHenze, line 533: Is that a bug in the original code, that we fixed in ReadctOS? Or is the original code correct here, and we just tried to mute some static code check and did it wrong?

drivers/filesystems/ext2/src/ext3/generic.c
```diff
@@ -919,6 +924,77 @@ Ext2SaveBuffer( IN PEXT2_IRP_CONTEXT    IrpContext,
 }
 
 
+#else
```
JoachimHenze, line 927: Since the above is within #ifdef __REACTOS__ we could strip the 2empty lines 925 and 926.

drivers/filesystems/ext2/src/fsctl.c
```diff
@@ -1799,7 +1799,7 @@ Ext2DeleteReparsePoint (IN PEXT2_IRP_CONTEXT IrpContext)
 
     BOOLEAN             FcbLockAcquired = FALSE;
     BOOLEAN             MainResourceAcquired = FALSE;
-
+
```
JoachimHenze, line 1802: Technically correct maybe, but I wouldn't do that. Open a PR in the source repo and remove it instead.

drivers/filesystems/ext2/src/devctl.c
Vince1171: it's a bug in the original code that we fixed indeed. After this PR, I'll try to upstream those fixes. but unfortunately, as stated in the Jira ticket, upstream seems to be abandoned. only https://github.com/bobranten/Ext4Fsd seems to still be around

drivers/filesystems/ext2/src/fsctl.c
Vince1171: Indeed

Vince1171: removed

drivers/filesystems/ext2/src/ext3/generic.c
Vince1171: done :)

drivers/filesystems/ext2/inc/linux/module.h
Vince1171: done

drivers/filesystems/ext2/src/devctl.c
HBelusca: It would be worth looking at whether the ext4fsd of Bo Branten also covers ext2 and 3, and if so, how much does the code change wrt. that ext2/3 driver we have. In case of positive observations, maybe one could think about updating to that ext4fsd ?

Vince1171: Yes, I'll try to have a look into it :)

gonzoMD: > it's a bug in the original code that we fixed indeed. After this PR, I'll try to upstream those fixes. but unfortunately, as stated in the Jira ticket, upstream seems to be abandoned. only https://github.com/bobranten/Ext4Fsd seems to still be around

This also have this bug, so it's worth to send a PR. See https://github.com/bobranten/Ext4Fsd/blob/d9da413ad28afa3c0519d730da60f8598bbaea32/Ext4Fsd/devctl.c#L514

Vince1171: after this PR is merged, I'll open a PR on ext4fsd with ReactOS fixes. Then I'll see to use Ext4fsd in ROS :)

drivers/filesystems/ext2/inc/linux/module.h
```diff
@@ -59,7 +59,8 @@ extern "C" {
 #define RtlUlongByteSwap(_x)     _byteswap_ulong((_x))
 #define RtlUlonglongByteSwap(_x) _byteswap_uint64((_x))
 
-#elif !defined(__REACTOS__)
+#else
+#ifndef __REACTOS__
```
HBelusca, line 63: why this change?

drivers/filesystems/ext2/inc/linux/module.h
```diff
@@ -79,6 +80,7 @@ RtlUlonglongByteSwap(
     IN ULONGLONG Source
 );
 #endif
+#endif //#ifndef __REACTOS__
```
HBelusca, line 83:
```suggestion
#endif //#ifndef __REACTOS__
#endif
```

drivers/filesystems/ext2/inc/linux/types.h
```diff
@@ -27,14 +27,18 @@ typedef	signed   __int64	__s64;
 typedef	unsigned __int64	__u64;
 
 
+#ifdef __REACTOS__
```
HBelusca, line 30: these typedefs weren't in the original code?

drivers/filesystems/ext2/src/ext4/ext4_extents.c
```diff
@@ -14,9 +14,13 @@
  */
 
 #include "ext2fs.h"
+#ifdef __REACTOS__
 #include <linux/ext4.h>
```
HBelusca, line 18:
```suggestion
#include "linux/ext4.h"
```

drivers/filesystems/ext2/src/ext4/ext4_extents.c
```diff
@@ -83,10 +87,16 @@ static inline void ext_debug(char *str, ...)
 {
 }
 #if TRUE
+#ifdef __REACTOS__
 #define EXT4_ERROR_INODE(inode, str, ...) do {                      \
             DbgPrint("inode[%p]: " str "\n", inode, ##__VA_ARGS__); \
         } while(0)
 #else
+#define EXT4_ERROR_INODE(inode, str, ...) do {                      \
```
HBelusca, line 95: Isn't that definition more correct than the ifdef'ed ReactOS one (minus the va_args that better need the ## in front of it)?
(or, is it possible to find why the ROS one removed the token pasting ## operator in front of str?)

drivers/filesystems/ext2/src/fileinfo.c
```diff
@@ -10,8 +10,13 @@
 /* INCLUDES *****************************************************************/
 
 #include "ext2fs.h"
+#ifdef __REACTOS__
 #include <linux/ext4.h>
```
HBelusca, line 14:
```suggestion
#include "linux/ext4.h"
```

drivers/filesystems/ext2/inc/linux/module.h
Vince1171: oops, indeed

drivers/filesystems/ext2/inc/linux/types.h
Vince1171: nope https://github.com/matt-wu/Ext3Fsd/blob/master/Ext3Fsd/include/linux/types.h#L31

drivers/filesystems/ext2/src/ext4/ext4_extents.c
Vince1171: I didn't wanted to change anything in our code, except the ifdef.
but I suppose I can do that change

drivers/filesystems/ext2/inc/linux/module.h
HBelusca: ok keep it as you have done.

drivers/filesystems/ext2/src/ext4/ext4_extents.c
Vince1171: I'm not familiar enough (yet) with this part of ext2fsd to answer this question.
all I can see is that it has been part of the integration of ext2fsd in our codebase https://github.com/reactos/reactos/commit/897634ce5bbb47b0f7c4ec48200baa77eb30b4f4 .
And I can trace back its origin to a patch in the related jira ticket https://jira.reactos.org/browse/CORE-10272 
https://jira.reactos.org/secure/attachment/31814/ext2fsd.patch  
but in the different conversation I haven't found an explanation about why they did that change

Vince1171: @HBelusca we cannot put ##str as str is always just a single string, one token, so no concatenation of token is possible here

HBelusca: yes actually you're correct. I've looked at the usages of `EXT4_ERROR_INODE` and it's used like:
```c
EXT4_ERROR_INODE(inode, "Some string with %s formatting", "something");
```
and so indeed, we cannot use the stringizer operator on `str`, otherwise it would re-stringize the string ^^

HBelusca (approved):

HBelusca:
I suppose that those changes (restoring the proper ifdefs wrt. upstream) doesn't break reading EXT2/3 partitions with ReactOS.

Vince1171:
> I suppose that those changes (restoring the proper ifdefs wrt. upstream) doesn't break reading EXT2/3 partitions with ReactOS.


no break indeed ^^

Screenshot from 2023-04-25 16-02-06

HBelusca:
Thanks for the sync!

PR #5247, opened 2023-04-16
[REGEDIT] Various fixes and improvements

JoachimHenze:
## Purpose

Fixes accelerator collisions for some languages (recommended to review the commits one by one to understand them, I explained them in each commits title).
Shrinks the binary size a bit.

JIRA issue: [CORE-18876](https://jira.reactos.org/browse/CORE-18876)


base/applications/regedit/treeview.c
```diff
@@ -781,7 +781,7 @@ BOOL SelectNode(HWND hwndTV, LPCWSTR keyPath)
 
     /* Load "My Computer" string... */
     LoadStringW(hInst, IDS_MY_COMPUTER, szBuffer, ARRAY_SIZE(szBuffer));
-    StringCbCatW(szBuffer, sizeof(szBuffer), L"\\");
+    wcscat(szBuffer, L"\\");
```
ThFabba, line 784: No.

base/applications/regedit/treeview.c
```diff
@@ -798,30 +798,26 @@ BOOL SelectNode(HWND hwndTV, LPCWSTR keyPath)
         size_t copyLength;
         s = wcschr(keyPath, L'\\');
         if (s != NULL)
-        {
```
ThFabba, line 801: You trying to start an edit war over this?

ThFabba (changes requested):

base/applications/regedit/clb/lang/tr-TR.rc
```diff
@@ -19,7 +19,7 @@ BEGIN
    CHECKBOX "&Esnek Sütunlar", 1715, 79, 50, 66, 10, BS_AUTOCHECKBOX | WS_GROUP | WS_TABSTOP
    GROUPBOX "Temel Biçimler", -1, 6, 80, 158, 34
    CHECKBOX "Gö&rünür", 1701, 10, 92, 34, 10, BS_AUTOCHECKBOX | WS_GROUP | WS_TABSTOP
-   CHECKBOX "Dev&re Dışı", 1702, 10, 102, 41, 10, BS_AUTOCHECKBOX | WS_GROUP | WS_TABSTOP
+   CHECKBOX "Edi&lgin", 1702, 10, 102, 41, 10, BS_AUTOCHECKBOX | WS_GROUP | WS_TABSTOP
```
HBelusca, line 22: Where do these changes come from?

base/applications/regedit/treeview.c
JoachimHenze: Why not? it generates shorter object code. And is safe, because IDS_MY_COMPUTER is under our control at compile time. Not under users control at runtime.

base/applications/regedit/treeview.c
JoachimHenze: I can change that back if you want.

base/applications/regedit/lang/pt-PT.rc
```diff
@@ -149,7 +149,7 @@ BEGIN
         MENUITEM "Modificar dados binários", ID_EDIT_MODIFY_BIN
         MENUITEM SEPARATOR
         MENUITEM "&Excluir\tDel", ID_EDIT_DELETE
-        MENUITEM "&Mudar o nome", ID_EDIT_RENAME
+        MENUITEM "&Renomear", ID_EDIT_RENAME
```
HBelusca, line 152: Where do these translation changes come from?

base/applications/regedit/childwnd.c
```diff
@@ -129,84 +128,6 @@ static void finish_splitbar(HWND hWnd, int x)
...
 }
 
-/*******************************************************************************
- *
- *  Key suggestion
- */
-
-#define MIN(a,b)    ((a < b) ? (a) : (b))
-
-static void SuggestKeys(HKEY hRootKey, LPCWSTR pszKeyPath, LPWSTR pszSuggestions,
```
HBelusca, line 139: Where was this "keys suggestion" feature used? (do you have a quick screenshot?)

base/applications/regedit/treeview.c
JoachimHenze: I added them back. Resolved.

base/applications/regedit/treeview.c
GeoB99: `wcscat` doesn't understand how much of space is left for the destination buffer to uphold the concatenated string, resulting into buffer overruns. The Safe String API explicitly serves the purpose to prevent this, among other things, and I would strongly recommend that Win32 built-in applications use these functions. So revert this.

ThFabba: 1. By forcing the programmer to specify the buffer size, safe string functions force the programmer to think about buffer size. This in itself avoids bugs.
2. The mere presence of an strsafe function call indicates "this has been reviewed for buffer overflows within the last decade". A `wcscat` call is always an uncertainty to the reader between "this has been deemed safe" and "the author of this code didn't know what they were doing".
3. Because of (1) and (2), I would like to some day be able to `#define wcscat do_not_use_this_function`. Any new code using this function is counter to that goal, and reverting already modernized code is even worse.

This is not about code size or binary size. It is about knowing that humans make mistakes, and making it harder for them to do so.

base/applications/regedit/lang/pt-PT.rc
JoachimHenze: From an older branch. Intention is to avoid accelerator collision, see commit comment.

base/applications/regedit/clb/lang/tr-TR.rc
JoachimHenze: From an older branch. Intention is to avoid accelerator collision, CHECKBOX "Dev&re Dışı" vs. CHECKBOX "Gö&rünür". See commits comment.

base/applications/regedit/lang/pt-PT.rc
ThFabba: So does that imply that a language maintainer has since updated these translations to be more correct/natural/consistent and you're reverting their decision? Or why are the translations different in the older branch?

base/applications/regedit/treeview.c
ThFabba: Thank you.

base/applications/regedit/childwnd.c
JoachimHenze: They were intended to be added to be added to the right-click-menu, see the code in the bottom of childwnd and framewnd. Have never seen this been working in a contributional way. Therefore stripped.

base/applications/regedit/lang/pt-PT.rc
JoachimHenze: If you prefer me to keep the current translation and just change the accelerator position within the string, just let me know. That would be possible as well, but results in longer strings. I had the impression that the older translators invested more love into the proper translations thn their younger colleagues. Therefore I relied on "the ancient wisdom of the elders" in some cases.

base/applications/regedit/childwnd.c
HBelusca: https://github.com/reactos/reactos/commit/32cf34124f5277185970b7aaf20055f03f1f1fa8

base/applications/regedit/lang/pt-PT.rc
HBelusca: Obviously you should keep the current translation from current master, and just fix the keyboard accelerators. You are in no position to judge whether the old translations were better or whatever. If they have changed in master, this proves that the old ones were not that great.

base/applications/regedit/treeview.c
JoachimHenze: Offer: I could add a comment that it has been reviewed for being safe.

Contrary if you request me to change it back to the safe func, you cannot expect me to merge the changes in question back into older branches, because the code got less efficient here without actually gaining **any** quality at all at these places. That is only true for the places I changed back. That other location you changed ofc fixed a real bug and is necessary therefore ofc.

base/applications/regedit/childwnd.c
JoachimHenze: It's not worth the additional complexity, and Windows doesn't have it. I checked on XPSP3.

JoachimHenze:
Okay, I see you seem to dislike most of the PR as it seems (stripping the suggestions). I will just close it. I don't care for the accelerator mess enough to creep through all that another time.

base/applications/regedit/lang/pt-PT.rc
JoachimHenze: The new ones just prove that the translator did not know a thing about what accelerators were and how to use them without causing duplicates and therefore breaking 2 accelerators from working at the same time with his changes. They are definitely not "better", otherwise they would work in master head. And they don't, due to the collisions.

HBelusca: Text translation and keyboard accelerators are two different things, even though they meet each other at the same place. The translator improved the text, but indeed at the detriment of the accelerators.

base/applications/regedit/treeview.c
learn-more: I am sorry, but if you can't understand why it's a bad idea to revert to hacky functions then you should not be modifying the code in the first place.

base/applications/regedit/lang/pt-PT.rc
dreimer1986: I still don't understand why you don't just move the & to the correct letter then? "Muda&r o nome" and all is set. I hope your reply is not "Because the older translation is 4 Bytes shorter"... If that is the case I feel the urge to review any and every german translation for shortening out of no reason.

base/applications/regedit/childwnd.c
dreimer1986: Windows does not have it is NO reason to remove a useful feature IMO. We are NOT cloning a OS here, we are improving, too.

PR #5248, opened 2023-04-17
[NTOS:MM] MmCanFileBeTruncated: Check whether second (optional) param…

archeYR:
Second parameter is optional, so mark it as such and check whether it was passed. Fixes a sporadic 0x24 bugcheck caused by access violation when running ReactOS on NTFS volume with WinXP ntfs.sys.

### DIFF
--- a/ntoskrnl/mm/section.c
+++ b/ntoskrnl/mm/section.c
@@ -4209,9 +4209,11 @@ MmMapViewOfSection(IN PVOID SectionObject,
 /*
  * @unimplemented
  */
-BOOLEAN NTAPI
-MmCanFileBeTruncated (IN PSECTION_OBJECT_POINTERS SectionObjectPointer,
-                      IN PLARGE_INTEGER   NewFileSize)
+BOOLEAN
+NTAPI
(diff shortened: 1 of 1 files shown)


GeoB99 (approved):

learn-more (approved):

ntoskrnl/mm/section.c
```diff
@@ -4248,6 +4250,12 @@ MmCanFileBeTruncated (IN PSECTION_OBJECT_POINTERS SectionObjectPointer,
         }
 #endif
     }
+    else
+    {
+        DPRINT("ERROR: File can't be truncated because it has references held to its data section\n");
```
HBelusca, line 4255: Is it ok to leave that as DPRINT? Or would it be of interest to promote it to DPRINT1 for some time? (if it's not too noisy during regular ROS usage?)

HBelusca (approved):

ntoskrnl/mm/section.c
archeYR: Should be fine with DPRINT1 actually, this case isn't very common.

PR #5250, opened 2023-04-19
[ATL] Take _ATL_FREE_THREADED into account

learn-more:
When it is not defined, COM should not be initialize multithreaded
`COINIT_MULTITHREADED` can cause problems when COM objects expect `Apartment` threading, because they will be serialized.

### DIFF
--- a/sdk/lib/atl/atlbase.h
+++ b/sdk/lib/atl/atlbase.h
@@ -849,7 +849,12 @@ public:
 
     static HRESULT InitializeCom()
     {
-        return ::CoInitializeEx(NULL, COINIT_MULTITHREADED);
+#if defined(_ATL_FREE_THREADED)
+        constexpr COINIT coInit = COINIT_MULTITHREADED;
+#else
+        constexpr COINIT coInit = COINIT_APARTMENTTHREADED;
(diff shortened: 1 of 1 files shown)


sdk/lib/atl/atlbase.h
```diff
@@ -766,7 +766,11 @@ class CAtlExeModuleT : public CAtlModuleT<T>
 
     static HRESULT InitializeCom()
     {
+#if defined(_ATL_FREE_THREADED)
         return ::CoInitializeEx(NULL, COINIT_MULTITHREADED);
+#else
+        return ::CoInitialize(NULL);
+#endif
```
HBelusca, line 773: What about just keeping the original CoInitializeEx call but just using `COINIT_APARTMENTTHREADED` instead as the flag? Something like
```suggestion
#ifdef _ATL_FREE_THREADED
        static const COINIT coInit = COINIT_MULTITHREADED;
#else
        static const COINIT coInit = COINIT_APARTMENTTHREADED;
#endif
        return ::CoInitializeEx(NULL, coInit);
```

HBelusca (approved):

PR #5251, opened 2023-04-19
[SHELL32] Implement SHQueryRecycleBinW

learn-more:


dll/win32/shell32/folders/CRecycleBin.cpp
```diff
@@ -1255,5 +1255,33 @@ HRESULT WINAPI SHQueryRecycleBinW(LPCWSTR pszRootPath, LPSHQUERYRBINFO pSHQueryR
...
+        CComPtr<IRecycleBinFile> spFile;
+        hr = spEnumList->Next(1, &spFile, NULL);
+        if (hr == S_FALSE)
+            return S_OK;
+
+        if (FAILED_UNEXPECTEDLY(hr))
+            return hr;
+
+        ULARGE_INTEGER Size = {};
+        if (FAILED_UNEXPECTEDLY(spFile->GetFileSize(&Size)))
```
HBelusca, line 1279:
```suggestion
        hr = spFile->GetFileSize(&Size);
        if (FAILED_UNEXPECTEDLY(hr))
```

HBelusca (approved):

JoachimHenze (approved):
Approved. How to trigger a call to that func manually from the GUI?

HBelusca:
> Approved. How to trigger a call to that func manually from the GUI?

Check whether the first results in https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=SHQueryRecycleBin
would give you the answer.

dll/win32/shell32/folders/CRecycleBin.cpp
```diff
@@ -1255,5 +1255,33 @@ HRESULT WINAPI SHQueryRecycleBinW(LPCWSTR pszRootPath, LPSHQUERYRBINFO pSHQueryR
...
+    hr = spRecycleBin->EnumObjects(&spEnumList);
+    if (!SUCCEEDED(hr))
+        return hr;
+
+    while (TRUE)
+    {
+        CComPtr<IRecycleBinFile> spFile;
+        hr = spEnumList->Next(1, &spFile, NULL);
+        if (hr == S_FALSE)
+            return S_OK;
```
GeoB99, line 1273: Would not it make sense to break the loop here, rather than immediately returning? Because ` return S_OK;` at the bottom of the function will never be reached.

learn-more: Both work, something like this is really up to personal preference.

PR #5257, opened 2023-04-26
[RICHED20] RichEdit: CompForm and CandForm

katahiromz:
## Purpose

Correctly display the composition window and the candidate window.

## Proposed changes
The changes of this PR are on `RichEdit` controls.

- Delete `imeStartIndex` member.
- At `update_caret` function, set the position and font of the composition window.
- We don't use internal composition string. Rely on the composition window.
- Improve `WM_IME_SETCONTEXT`, `WM_IME_CONTROL`, `WM_IME_SELECT`, `WM_IME_STARTCOMPOSITION`, `WM_IME_COMPOSITION` and `WM_IME_ENDCOMPOSITION` message handlings.

## Comparison

BEFORE:
before

AFTER:
after

## TODO

- [x] Do tests.

### DIFF
--- a/dll/win32/riched20/caret.c
+++ b/dll/win32/riched20/caret.c
@@ -342,7 +342,7 @@ void update_caret(ME_TextEditor *editor)
       if (fmt.dwMask & CFM_SIZE)
       {
         HDC hdc = CreateCompatibleDC(NULL);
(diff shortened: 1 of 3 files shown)


HBelusca:
May be worth sending (parts of it) to Wine :thinking:

PR #5258, opened 2023-04-27
[INPUT] Remove SHLoadRegUIStringW hack

katahiromz:
## Purpose
Remove the needless hack and reduce the maintenance cost. Because `shlwapi!SHLoadRegUIStringW` function is already implemented, so we don't need the hack any more.

## Proposed change

- Simplify `FakeSHLoadRegUIStringW` without using hack.

## TODO

- [x] Do tests.

### DIFF
--- a/dll/cpl/input/layout_list.c
+++ b/dll/cpl/input/layout_list.c
@@ -87,51 +87,8 @@ LayoutList_Destroy(VOID)
 
 typedef HRESULT (WINAPI *FN_SHLoadRegUIStringW)(HKEY, LPCWSTR, LPWSTR, DWORD);
 
-/* FIXME: Use shlwapi!SHLoadRegUIStringW instead when it is fully implemented */
 HRESULT FakeSHLoadRegUIStringW(HKEY hkey, LPCWSTR value, LPWSTR buf, DWORD size)
 {
-#if 1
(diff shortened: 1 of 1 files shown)


katahiromz:
> Does it still work on XPSP3 afterwards?

Yep, it works.

VirtualBox_Windows XP_27_04_2023_18_39_07

dll/cpl/input/layout_list.c
```diff
@@ -87,51 +87,8 @@ LayoutList_Destroy(VOID)
 
 typedef HRESULT (WINAPI *FN_SHLoadRegUIStringW)(HKEY, LPCWSTR, LPWSTR, DWORD);
 
-/* FIXME: Use shlwapi!SHLoadRegUIStringW instead when it is fully implemented */
```
HBelusca, line 90: Note that this is also in https://git.reactos.org/?p=reactos.git;a=blob;f=base/applications/kbswitch/kbswitch.c;hb=4cdcea5e65326cbea77458a3e9daa335d2e530b7#l121

katahiromz: https://github.com/reactos/reactos/pull/5260

HBelusca (approved):

PR #5259, opened 2023-04-27
[NTOS:SE][SDK] Pretend the Administrator user is elevated

katahiromz:
## Purpose

Preparation for UAC control.
JIRA issue: [CORE-11357](https://jira.reactos.org/browse/CORE-11357)

## Proposed changes

- Pretend Administrator with power that is always elevated.

## TODO

- [ ] Do big tests.

## Screenshot

VirtualBox_ReactOS_27_04_2023_19_50_25

### DIFF
--- a/ntoskrnl/se/tokencls.c
+++ b/ntoskrnl/se/tokencls.c
@@ -16,6 +16,13 @@
 
 /* INFORMATION CLASSES ********************************************************/
 
+/* FIXME: Delete this */
+typedef enum _TOKEN_ELEVATION_TYPE {
+    TokenElevationTypeDefault = 1,
+    TokenElevationTypeFull,
+    TokenElevationTypeLimited,
+} TOKEN_ELEVATION_TYPE, *PTOKEN_ELEVATION_TYPE;
+
 static const INFORMATION_CLASS_INFO SeTokenInformationClass[] = {
 
     /* Class 0 not used, blame MS! */
@@ -55,6 +62,12 @@ static const INFORMATION_CLASS_INFO SeTokenInformationClass[] = {
     IQS_SAME(TOKEN_AUDIT_POLICY_INFORMATION, ULONG, ICIF_SET | ICIF_SET_SIZE_VARIABLE),
     /* TokenOrigin */
     IQS_SAME(TOKEN_ORIGIN, ULONG, ICIF_QUERY | ICIF_SET),
+    /* TokenElevationType */
+    IQS_SAME(TOKEN_ELEVATION_TYPE, ULONG, ICIF_QUERY | ICIF_SET),
+    /* TokenLinkedToken */
+    IQS_NONE, /* FIXME */
+    /* TokenElevation */
+    IQS_SAME(TOKEN_ELEVATION, ULONG, ICIF_QUERY | ICIF_SET),
 };
 
 /* PUBLIC FUNCTIONS *****************************************************************/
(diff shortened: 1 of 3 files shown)


ntoskrnl/se/tokencls.c
```diff
@@ -16,6 +16,13 @@
 
 /* INFORMATION CLASSES ********************************************************/
 
+/* FIXME: Delete this */
+typedef enum _TOKEN_ELEVATION_TYPE {
```
HBelusca, line 20: Does this exist on Windows XP or 2003? Or is this Vista+?

katahiromz: I found this code in our `winnt_old.h`.

ntoskrnl/se/tokencls.c
```diff
@@ -16,6 +16,13 @@
 
 /* INFORMATION CLASSES ********************************************************/
 
+/* FIXME: Delete this */
```
JoachimHenze, line 19: Is it possible to give a more specific hint **when** this can be deleted? Any specific precondition to be met? Anything must have been implemented anywhere? Where?

ntoskrnl/se/tokencls.c
katahiromz: https://learn.microsoft.com/en-us/windows/win32/api/winnt/ne-winnt-token_elevation_type
image

ntoskrnl/se/tokencls.c
katahiromz: `<winnt.h>` should provide this definition.

ntoskrnl/se/tokencls.c
HBelusca: OK then why this whole ordeal of elevation for the AV is even needed, if this doesn't exist on Win2k3 ?

HBelusca: > I couldn't download the target AV software.

See link in updated description of the JIRA report.

katahiromz: VirtualBox_ReactOS_27_04_2023_21_16_59
It displayed this window for a while. And some minutes later, it vanished.

HBelusca: What about Win2k3 ?

sdk/include/xdk/setypes.h
```diff
@@ -1053,6 +1053,12 @@ typedef struct _TOKEN_ELEVATION {
   $ULONG TokenIsElevated;
 } TOKEN_ELEVATION, *PTOKEN_ELEVATION;
 
+typedef enum _TOKEN_ELEVATION_TYPE {
```
HBelusca, line 1056: I therefore wonder whether we need to surround this typedef by `#if (NTDDI_VERSION >= NTDDI_VISTA)  .... #endif`

GeoB99:
I'm no fan of pilling the Security subsystem up with hacks just to make an app work. The elevation of an access token depends on multiple factors such as if the token is restricted or filtered (if so then how restricted is it), in which session is the token at, privileges, if it belongs to a higher SID group such as Administrators, etc. It is LSASS that sets the appropriate elevation for a token once it is being created or its security properties have changed, also whether UAC is enabled or disabled as this can influence on the elevation of a token.

In your hack you're telling the caller a token is always elevated -- this can lead to erratic behaviors. For instance a restricted token is by no means elevated as it can have stripped privileges, disabled SIDs or deny restrict SIDs added into it so suppose an app looks for a token that is by chance, restricted, it maybe improperly. This is just an example among others on why I don't want hacks here.

By the way...
Capture
Capture2
I had to run the installer by toggling Windows XP compatibility mode on Properties in order to successfully install that AV, otherwise on normal execution the installer runs but after some minutes it exits prematurely with no premonition or anything. This furthermore proves that real work has to be done elsewhere and certainly not in Se.

HBelusca:
> I had to run the installer by toggling Windows XP compatibility mode on Properties in order to successfully install that AV, otherwise on normal execution the installer runs but after some minutes it exits prematurely with no premonition or anything. [...]

This shows that it's yet again an AV that wants to run on a "workstation"-type Windows OS. The "exits prematurely" appears to reproduce what @katahiromz observed when he tried to run the AV on ReactOS.... Maybe he would have observed something different if he had tried to run it under XP compatibility / made ReactOS report as "Workstation" ?

PR #5267, opened 2023-05-04
[FREELDR] Enable UEFI boot for i386 and amd64. CORE-11954

DarkFire01:
## Purpose

Finally allow UEFI FreeLoader to boot ReactOS kernel on our 2 primary architectures.

JIRA issue: [CORE-11954](https://jira.reactos.org/browse/CORE-11954)

## Proposed changes

- Implement remaining code needed to start kernel for i386 and amd64

## TODO

- [X]  Figure out if it's possible to just use the stack given by UEFI so we don't have to repeat the last few lines of freeldr before exit inside another file

### DIFF
--- a/boot/freeldr/freeldr/arch/uefi/amd64/uefiasm.S
+++ b/boot/freeldr/freeldr/arch/uefi/amd64/uefiasm.S
@@ -1,0 +1,85 @@
+/*
+ * PROJECT:     FreeLoader UEFI Support
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     x64 assembly UEFI escape helper functions
+ * COPYRIGHT:   Copyright 2023 Justin Miller <justinmiller100@gmail.com>
+ */
+
+#include <asm.inc>
+#include <arch/pc/x86common.h>
+#include <arch/pc/pcbios.h>
+
+EXTERN UefiServiceStack:QWORD
+EXTERN BasicStack:QWORD
+EXTERN ExecuteLoaderCleanly:PROC
+EXTERN UefiExitBootServices:PROC
+
+.code64
+
+// void _exituefi(VOID)
+PUBLIC _exituefi
+_exituefi:
+    /* Save non-volatile registers */
+    push rbp
+    push rsi
+    push rdi
+    push rbx
+
+    /* Save the old stack */
+    mov rbx, rsp
+
+    /* Load the new stack */
+    xor rbp, rbp
+    mov rsp, qword ptr UefiServiceStack[rip]
+
+    /* Call the entry routine, passing the parameters */
+    mov rax, UefiExitBootServices[rip]
+    call rax
+
+    /* Retore old stack */
+    mov rsp, rbx
+
+    /* Retore non-volatiles */
+    pop rbx
+    pop rdi
+    pop rsi
+    pop rbp
+
+#ifdef _USE_ML
+    lgdt fword ptr [_gdtptr]
+#else
+    lgdt cs:[_gdtptr][rip] /* GAS isn't my friend - avoid letting it generate absolute addressing */
+#endif
+
+    /* All done */
+    ret
+
+// void _changestack(VOID)
+PUBLIC _changestack
+_changestack:
+    mov rax, rsp
+    mov rsp, BasicStack[rip]
+    push rax
+    call ExecuteLoaderCleanly[rip]
+    ret
+
+.align 8
+gdt:
+    .word HEX(0000), HEX(0000), HEX(0000), HEX(0000) /* 00: NULL descriptor */
+    .word HEX(0000), HEX(0000), HEX(0000), HEX(0000) /* 08: */
+    .word HEX(0000), HEX(0000), HEX(9800), HEX(0020) /* 10: long mode CS */
+    .word HEX(FFFF), HEX(0000), HEX(F300), HEX(00CF) /* 18: long mode DS */
+    .word HEX(FFFF), HEX(0000), HEX(9E00), HEX(0000) /* 20: 16-bit real mode CS */
+    .word HEX(FFFF), HEX(0000), HEX(9200), HEX(0000) /* 28: 16-bit real mode DS */
+    .word HEX(FFFF), HEX(0000), HEX(9B00), HEX(00CF) /* 30: compat mode CS */
+
+/* GDT table pointer */
+_gdtptr:
+    .word HEX(37) /* Limit */
+#ifdef _USE_ML
+    .quad gdt     /* Base Address */
+#else
+    .quad gdt, 0  /* Base Address */
+#endif
+
+END
--- a/boot/freeldr/freeldr/arch/uefi/i386/uefiasm.S
+++ b/boot/freeldr/freeldr/arch/uefi/i386/uefiasm.S
@@ -1,0 +1,109 @@
+/*
+ * PROJECT:     FreeLoader UEFI Support
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     i386 assembly UEFI escape helper functions
+ * COPYRIGHT:   Copyright 2023 Justin Miller <justinmiller100@gmail.com>
+ */
+
+#include <asm.inc>
+#include <arch/pc/x86common.h>
+#include <arch/pc/pcbios.h>
+
+PUBLIC _gdtptr
+PUBLIC _i386idtptr
+PUBLIC __changestack
+EXTERN _UefiServiceStack:DWORD
+EXTERN _i386Idt:DWORD
+EXTERN _ExecuteLoaderCleanly:PROC
+EXTERN _UefiExitBootServices:PROC
+EXTERN _BasicStack:DWORD
+.code32
+
+// void __exituefi(VOID)
+PUBLIC __exituefi
+__exituefi:
+    push ebp
+    push esi
+    push edi
+    push ebx
+
+    /* Save the old stack */
+    mov ebx, esp
+
+    /* Load the new stack */
+    xor ebp, ebp
+    mov esp, _UefiServiceStack
+
+    /* Call the entry routine, passing the parameters */
+    call _UefiExitBootServices
+
+    /* Retore old stack */
+    mov esp, ebx
+
+    /* Retore non-volatiles */
+    pop ebx
+    pop edi
+    pop esi
+    pop ebp
+
+#ifdef _USE_ML
+    lidt fword ptr ds:[_i386idtptr]
+    lgdt fword ptr [_gdtptr]
+#else
+    lgdt cs:[_gdtptr]
+    lidt _i386idtptr
+#endif
+
+    /* All done */
+    ret
+
+// void __reloadsegment(VOID)
+PUBLIC __changestack
+__changestack:
+    mov eax, esp
+    mov esp, _BasicStack
+    push eax
+    call _ExecuteLoaderCleanly
+    ret
+
+    .align 4    /* force 4-byte alignment */
+gdt:
+    /* NULL Descriptor */
+   .word HEX(0000)
+   .word HEX(0000)
+   .word HEX(0000)
+   .word HEX(0000)
+
+    /* 32-bit flat CS */
+    .word HEX(FFFF)
+    .word HEX(0000)
+    .word HEX(9A00)
+    .word HEX(00CF)
+
+    /* 32-bit flat DS */
+    .word HEX(FFFF)
+    .word HEX(0000)
+    .word HEX(9200)
+    .word HEX(00CF)
+
+    /* 16-bit real mode CS */
+    .word HEX(FFFF)
+    .word HEX(0000)
+    .word HEX(9E00)
+    .word HEX(0000)
+
+    /* 16-bit real mode DS */
+    .word HEX(FFFF)
+    .word HEX(0000)
+    .word HEX(9200)
+    .word HEX(0000)
+
+/* GDT table pointer */
+_gdtptr:
+    .word HEX(27)  /* Limit */
+    .long gdt, 0   /* Base Address */
+
+_i386idtptr:
+    .word 255      /* Limit */
+    .long _i386Idt /* Base Address */
+END
--- a/boot/freeldr/freeldr/arch/uefi/ueficon.c
+++ b/boot/freeldr/freeldr/arch/uefi/ueficon.c
@@ -138,6 +138,5 @@ UefiConsGetCh(VOID)
     /* UEFI will stack input requests, we have to clear it */
     Key.UnicodeChar = 0;
     Key.ScanCode = 0;
-    GlobalSystemTable->ConIn->Reset(GlobalSystemTable->ConIn, FALSE);
     return KeyOutput;
 }
--- a/boot/freeldr/freeldr/arch/uefi/uefihw.c
+++ b/boot/freeldr/freeldr/arch/uefi/uefihw.c
@@ -1,0 +1,149 @@
+/*
+ * PROJECT:     FreeLoader UEFI Support
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Hardware detection routines
+ * COPYRIGHT:   Copyright 2022 Justin Miller <justinmiller100@gmail.com>
+ */
+
+/* INCLUDES ******************************************************************/
+
+#include <uefildr.h>
+
+#include <debug.h>
+DBG_DEFAULT_CHANNEL(WARNING);
+
+/* GLOBALS *******************************************************************/
+
+extern EFI_SYSTEM_TABLE * GlobalSystemTable;
+extern EFI_HANDLE GlobalImageHandle;
+extern UCHAR PcBiosDiskCount;
+extern EFI_MEMORY_DESCRIPTOR* EfiMemoryMap;
+extern UINT32 FreeldrDescCount;
+
+BOOLEAN AcpiPresent = FALSE;
+
+/* FUNCTIONS *****************************************************************/
+
+static
+PRSDP_DESCRIPTOR
+FindAcpiBios(VOID)
+{
+    UINTN i;
+    RSDP_DESCRIPTOR* rsdp = NULL;
+    EFI_GUID acpi2_guid = EFI_ACPI_20_TABLE_GUID;
+
+    for (i = 0; i < GlobalSystemTable->NumberOfTableEntries; i++)
+    {
+        if (!memcmp(&GlobalSystemTable->ConfigurationTable[i].VendorGuid,
+                    &acpi2_guid, sizeof(acpi2_guid)))
+        {
+            rsdp = (RSDP_DESCRIPTOR*)GlobalSystemTable->ConfigurationTable[i].VendorTable;
+            break;
+        }
+    }
+
+    return rsdp;
+}
+
+VOID
+DetectAcpiBios(PCONFIGURATION_COMPONENT_DATA SystemKey, ULONG *BusNumber)
+{
+    PCONFIGURATION_COMPONENT_DATA BiosKey;
+    PCM_PARTIAL_RESOURCE_LIST PartialResourceList;
+    PCM_PARTIAL_RESOURCE_DESCRIPTOR PartialDescriptor;
+    PRSDP_DESCRIPTOR Rsdp;
+    PACPI_BIOS_DATA AcpiBiosData;
+    ULONG TableSize;
+
+    Rsdp = FindAcpiBios();
+
+    if (Rsdp)
+    {
+        /* Set up the flag in the loader block */
+        AcpiPresent = TRUE;
+
+        /* Calculate the table size */
+        TableSize = FreeldrDescCount * sizeof(BIOS_MEMORY_MAP) +
+            sizeof(ACPI_BIOS_DATA) - sizeof(BIOS_MEMORY_MAP);
+
+        /* Set 'Configuration Data' value */
+        PartialResourceList = FrLdrHeapAlloc(sizeof(CM_PARTIAL_RESOURCE_LIST) +
+                                             TableSize, TAG_HW_RESOURCE_LIST);
+        if (PartialResourceList == NULL)
+        {
+            ERR("Failed to allocate resource descriptor\n");
+            return;
+        }
+
+        RtlZeroMemory(PartialResourceList, sizeof(CM_PARTIAL_RESOURCE_LIST) + TableSize);
+        PartialResourceList->Version = 0;
+        PartialResourceList->Revision = 0;
+        PartialResourceList->Count = 1;
+
+        PartialDescriptor = &PartialResourceList->PartialDescriptors[0];
+        PartialDescriptor->Type = CmResourceTypeDeviceSpecific;
+        PartialDescriptor->ShareDisposition = CmResourceShareUndetermined;
+        PartialDescriptor->u.DeviceSpecificData.DataSize = TableSize;
+
+        /* Fill the table */
+        AcpiBiosData = (PACPI_BIOS_DATA)&PartialResourceList->PartialDescriptors[1];
+
+        if (Rsdp->revision > 0)
+        {
+            TRACE("ACPI >1.0, using XSDT address\n");
+            AcpiBiosData->RSDTAddress.QuadPart = Rsdp->xsdt_physical_address;
+        }
+        else
+        {
+            TRACE("ACPI 1.0, using RSDT address\n");
+            AcpiBiosData->RSDTAddress.LowPart = Rsdp->rsdt_physical_address;
+        }
+
+        AcpiBiosData->Count = FreeldrDescCount;
+        memcpy(AcpiBiosData->MemoryMap, EfiMemoryMap,
+            FreeldrDescCount * sizeof(BIOS_MEMORY_MAP));
+
+        TRACE("RSDT %p, data size %x\n", Rsdp->rsdt_physical_address, TableSize);
+
+        /* Create new bus key */
+        FldrCreateComponentKey(SystemKey,
+                               AdapterClass,
+                               MultiFunctionAdapter,
+                               0x0,
+                               0x0,
+                               0xFFFFFFFF,
+                               "ACPI BIOS",
+                               PartialResourceList,
+                               sizeof(CM_PARTIAL_RESOURCE_LIST) + TableSize,
+                               &BiosKey);
+
+        /* Increment bus number */
+        (*BusNumber)++;
+    }
+}
+
+PCONFIGURATION_COMPONENT_DATA
+UefiHwDetect(VOID)
+{
+    PCONFIGURATION_COMPONENT_DATA SystemKey;
+    ULONG BusNumber = 0;
+
+    TRACE("DetectHardware()\n");
+
+    /* Create the 'System' key */
+#if defined(_M_IX86) || defined(_M_AMD64)
+    FldrCreateSystemKey(&SystemKey, "AT/AT COMPATIBLE");
+#elif defined(_M_IA64)
+    FldrCreateSystemKey(&SystemKey, "Intel Itanium processor family");
+#elif defined(_M_ARM) || defined(_M_ARM64)
+    FldrCreateSystemKey(&SystemKey, "ARM processor family");
+#else
+    #error Please define a system key for your architecture
+#endif
+
+    /* Detect ACPI */
+    DetectAcpiBios(SystemKey, &BusNumber);
+
+    TRACE("DetectHardware() Done\n");
+    return SystemKey;
+}
--- a/boot/freeldr/freeldr/arch/uefi/uefildr.c
+++ b/boot/freeldr/freeldr/arch/uefi/uefildr.c
@@ -8,11 +8,16 @@
 #include <uefildr.h>
 
 #include <debug.h>
+DBG_DEFAULT_CHANNEL(WARNING);
 
 /* GLOBALS ********************************************************************/
 
 EFI_HANDLE GlobalImageHandle;
 EFI_SYSTEM_TABLE *GlobalSystemTable;
+PVOID UefiServiceStack;
+PVOID BasicStack;
+
+void _changestack(VOID);
 
 /* FUNCTIONS ******************************************************************/
 
@@ -25,15 +30,61 @@ EfiEntry(
     GlobalImageHandle = ImageHandle;
     GlobalSystemTable = SystemTable;
 
-    BootMain(NULL);
+    /* Needed for default settings */
+    CmdLineParse("");
+
+    /* Debugger pre-initialization */
+    DebugInit(0);
+
+    MachInit("");
+
+    /* UI pre-initialization */
+    if (!UiInitialize(FALSE))
+    {
+        UiMessageBoxCritical("Unable to initialize UI.");
+        goto Quit;
+    }
+
+    /* Initialize memory manager */
+    if (!MmInitializeMemoryManager())
+    {
+        UiMessageBoxCritical("Unable to initialize memory manager.");
+        goto Quit;
+    }
+
+    /* Initialize I/O subsystem */
+    FsInit();
+
+    /* 0x32000 is what UEFI defines, but we can go smaller if we want */
+    BasicStack = (PVOID)((ULONG_PTR)0x32000 + (ULONG_PTR)MmAllocateMemoryWithType(0x32000, LoaderOsloaderStack));
+    _changestack();
+
+Quit:
+    /* If we reach this point, something went wrong before, therefore reboot */
+    Reboot();
 
     UNREACHABLE;
     return 0;
 }
 
+void
+ExecuteLoaderCleanly(PVOID PreviousStack)
+{
+    TRACE("ExecuteLoaderCleanly Entry\n");
+    UefiServiceStack = PreviousStack;
+
+    RunLoader();
+    UNREACHABLE;
+}
+
 #ifndef _M_ARM
 VOID __cdecl Reboot(VOID)
 {
-
+    //TODO: Replace with a true firmware reboot eventually
+    WARN("Something has gone wrong - halting FreeLoader\n");
+    for (;;)
+    {
+        NOTHING;
+    }
 }
 #endif
--- a/boot/freeldr/freeldr/arch/uefi/uefimem.c
+++ b/boot/freeldr/freeldr/arch/uefi/uefimem.c
@@ -27,6 +27,7 @@ AddMemoryDescriptor(
 
 /* GLOBALS *******************************************************************/
 
+extern ULONG LoaderPagesSpanned;
 extern EFI_SYSTEM_TABLE* GlobalSystemTable;
 extern EFI_HANDLE GlobalImageHandle;
 extern REACTOS_INTERNAL_BGCONTEXT framebufferData;
@@ -38,6 +39,8 @@ SIZE_T OsLoaderSize;
 EFI_HANDLE PublicBootHandle;
 PVOID ExitStack;
 PVOID EndofExitStack;
+
+void _exituefi(VOID);
 
 /* FUNCTIONS *****************************************************************/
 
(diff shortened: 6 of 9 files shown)


boot/freeldr/freeldr/arch/uefi/uefihw.c
```diff
@@ -0,0 +1,151 @@
...
+
+BOOLEAN AcpiPresent = FALSE;
+
+/* FUNCTIONS *****************************************************************/
+
+static
+PRSDP_DESCRIPTOR
+FindAcpiBios(VOID)
+{
+    UINT32 i;
```
HBelusca, line 31: Could be `UINTN` in order to match the type of `NumberOfTableEntries`.

boot/freeldr/freeldr/arch/uefi/uefihw.c
```diff
@@ -0,0 +1,151 @@
...
+    RSDP_DESCRIPTOR* rsdp = NULL;
+    EFI_GUID acpi2_guid = EFI_ACPI_20_TABLE_GUID;
+
+    for (i = 0; i < GlobalSystemTable->NumberOfTableEntries; i++)
+    {
+        if (!memcmp(&GlobalSystemTable->ConfigurationTable[i].VendorGuid, &acpi2_guid, sizeof(EFI_GUID)))
```
HBelusca, line 37:
```suggestion
        if (!memcmp(&GlobalSystemTable->ConfigurationTable[i].VendorGuid, &acpi2_guid, sizeof(acpi2_guid)))
```

boot/freeldr/freeldr/arch/uefi/uefihw.c
```diff
@@ -0,0 +1,151 @@
...
+    /* Create the 'System' key */
+    FldrCreateSystemKey(&SystemKey);
+#if defined(_M_IX86) || defined(_M_AMD64) /* Taken from Windows 11 */
+    FldrSetIdentifier(SystemKey, "AT/AT COMPATIBLE");
+#elif defined(_M_IA64) /* Taken from Windows XP 64-bit*/
+    FldrSetIdentifier(SystemKey, "Intel Itanium processor family");
+#elif defined(_M_ARM) || defined(_M_ARM64) /* Taken from Windows RT/11 */
+    FldrSetIdentifier(SystemKey, "ARM processor family");
+#else
+    #error "Please define a system key for your architecture"
```
HBelusca, line 143: Quotes are not required.
```suggestion
    #error Please define a system key for your architecture
```

boot/freeldr/freeldr/arch/uefi/arch/i386/uefiasm.S
```diff
@@ -0,0 +1,84 @@
...
+
+#include <asm.inc>
+#include <arch/pc/x86common.h>
+#include <arch/pc/pcbios.h>
+
+.code32
+PUBLIC _gdtptr
+PUBLIC _i386idtptr
+EXTERN _i386Idt:DWORD
+EXTERN _EndofExitStack:DWORD
```
HBelusca, line 16: I was looking at existing code in master. This stack is a new piece of memory allocated in UefiPrepareForReactOS() with LoaderOsloaderStack type. I was wondering: When exiting the UEFI boot services, what happens to the existing stack FreeLdr is using? Is it still valid? If so, why wouldn't it be possible to use that latter one instead of a new one?

boot/freeldr/freeldr/ntldr/arch/i386/winldr.c
```diff
@@ -744,3 +744,45 @@ MempDump(VOID)
...
+                   KERNEL_ENTRY_POINT KiSystemStartup)
+{
+    TRACE("Preparing to jump to kernel\n");
+    PubLoaderBlockVA = LoaderBlockVA;
+    PubKiSystemStartup = KiSystemStartup;
+    _LeaveUEFIStack();
+}
+
+VOID
+UefiExitToKernel(VOID)
```
HBelusca, line 767: Hmm.... this looks redundant and duplicates what's already done in winldr.c .
It's not possible to return execution back to the caller? Is it because you've changed the stack?

boot/freeldr/freeldr/arch/uefi/arch/i386/uefiasm.S
HBelusca: Thinking more about that, in PC-BIOS architecture we reserve when freeldr starts, a zone for the freeldr stack:
https://git.reactos.org/?p=reactos.git;a=blob;f=boot/freeldr/freeldr/arch/i386/pc/pcmem.c;hb=f18111b64120806529e93404f21b6320f211ec39#l546
It may be worth investigating that and doing something similar for UEFI too. The stack used by FreeLdr there should be already known in the main() function.

boot/freeldr/freeldr/ntldr/arch/i386/winldr.c
DarkFire01: It’s not possible to return from callee because I’ve changed the stack. But yes I’m trying to work on figuring out a reliable way where I don’t have to change the stack, if that’s possible. I feel like it should be

boot/freeldr/freeldr/arch/uefi/arch/i386/uefiasm.S
DarkFire01: Problem is you still have to have the UEFI provided stack In order to exit boot services. So I can’t just change it before calling the main function

boot/freeldr/freeldr/arch/uefi/ueficon.c
```diff
@@ -16,6 +16,8 @@ extern EFI_SYSTEM_TABLE* GlobalSystemTable;
 static unsigned CurrentCursorX = 0;
 static unsigned CurrentCursorY = 0;
 static unsigned CurrentAttr = 0x0f;
+static EFI_INPUT_KEY Key;
+static BOOLEAN scancode;
```
binarymaster, line 20: Unused vars
```
../src/boot/freeldr/freeldr/arch/uefi/ueficon.c:20:16: error: 'scancode' defined but not used [-Werror=unused-variable]
 static BOOLEAN scancode;
                ^~~~~~~~
../src/boot/freeldr/freeldr/arch/uefi/ueficon.c:19:22: error: 'Key' defined but not used [-Werror=unused-variable]
 static EFI_INPUT_KEY Key;
                      ^~~
```

DarkFire01: yeah sorry! Im currently playing with this, with the other 2 architectures on top of it locally right now. I'm trying to make sure what I'm going to do will work across the board as is

This will be pushed with pretty large changes to it asap.

binarymaster:
image

binarymaster:
gcc and clang on amd64 are not happy:
```
boot/freeldr/freeldr/CMakeFiles/uefifreeldr_common.dir/arch/uefi/arch/amd64/uefiasm.S.obj:/home/runner/work/reactos/reactos/build/../src/boot/freeldr/freeldr/arch/uefi/arch/amd64/uefiasm.S:41:(.text+0x47): relocation truncated to fit: R_X86_64_32S against `.text'
boot/freeldr/freeldr/CMakeFiles/uefifreeldr_common.dir/arch/uefi/arch/amd64/uefiasm.S.obj:/home/runner/work/reactos/reactos/build/../src/boot/freeldr/freeldr/arch/uefi/arch/amd64/uefiasm.S:43:(.text+0x4f): relocation truncated to fit: R_X86_64_32S against symbol `EndofExitStack' defined in .bss section in boot/freeldr/freeldr/CMakeFiles/uefifreeldr_common.dir/arch/uefi/uefimem.c.obj
collect2: error: ld returned 1 exit status
```

boot/freeldr/freeldr/ntldr/winldr.h
```diff
@@ -172,6 +172,14 @@ MempSetupPaging(IN PFN_NUMBER StartPage,
                 IN PFN_NUMBER NumberOfPages,
                 IN BOOLEAN KernelMapping);
 
+VOID
+WinldrFinalizeBoot(PLOADER_PARAMETER_BLOCK LoaderBlockVA,
+                   KERNEL_ENTRY_POINT KiSystemStartup);
+
+VOID
+WinLdrExitToNtoskrnl(PLOADER_PARAMETER_BLOCK LoaderBlockVA,
+                     KERNEL_ENTRY_POINT KiSystemStartup);
```
JoachimHenze, line 181:
```suggestion
    KERNEL_ENTRY_POINT KiSystemStartup);
```

boot/freeldr/freeldr/ntldr/winldr.c
```diff
@@ -1238,11 +1238,25 @@ LoadAndBootWindowsCommon(
...
+
+    /* Unify UEFI and BIOS back to the same exit point */
+    WinLdrExitToNtoskrnl(LoaderBlockVA, KiSystemStartup);
+
+    UNREACHABLE; // return ESUCCESS;
+}
+
+VOID
+WinLdrExitToNtoskrnl(PLOADER_PARAMETER_BLOCK LoaderBlockVA,
+                     KERNEL_ENTRY_POINT KiSystemStartup)
```
JoachimHenze, line 1253:
```suggestion
    KERNEL_ENTRY_POINT KiSystemStartup)
```

boot/freeldr/freeldr/ntldr/winldr.h
```diff
@@ -172,6 +172,14 @@ MempSetupPaging(IN PFN_NUMBER StartPage,
                 IN PFN_NUMBER NumberOfPages,
                 IN BOOLEAN KernelMapping);
 
+VOID
+WinldrFinalizeBoot(PLOADER_PARAMETER_BLOCK LoaderBlockVA,
+                   KERNEL_ENTRY_POINT KiSystemStartup);
```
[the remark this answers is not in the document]
binarymaster: Fixed formatting in 31b95a6b.

boot/freeldr/freeldr/arch/uefi/arch/i386/uefiasm.S
HBelusca: It looks like now you can use a way similar to the one in AMD64, using a parameter of the `__ExitUefi` function to pass the end of exit stack, and getting rid of this `_EndofExitStack` variable.

DarkFire01:
Welp. That about does it :)
I have 2 places where MempSetupPaging are used and i need to fix that. 
it's a small issue and it just requires some discussion at most. 

But that's the major changes planned for this PR. As of the state of this PR UEFI boot will operate as long as you have a serial port you should be able to test it.

Some more boot fixes will come down the road but this covers 85% of devices we've ran into

boot/freeldr/freeldr/arch/uefi/ueficon.c
```diff
@@ -138,6 +138,5 @@ UefiConsGetCh(VOID)
     /* UEFI will stack input requests, we have to clear it */
     Key.UnicodeChar = 0;
     Key.ScanCode = 0;
-    GlobalSystemTable->ConIn->Reset(GlobalSystemTable->ConIn, FALSE);
```
HBelusca, line 141: What does this removal fix?

boot/freeldr/freeldr/ntldr/winldr.h
```diff
@@ -172,6 +172,7 @@ MempSetupPaging(IN PFN_NUMBER StartPage,
                 IN PFN_NUMBER NumberOfPages,
                 IN BOOLEAN KernelMapping);
 
+
```
HBelusca, line 175:
```suggestion
```

boot/freeldr/freeldr/ntldr/wlmemory.c
```diff
@@ -176,7 +176,10 @@ MempSetupPagingForRegion(
 #ifdef _M_ARM
 #define PKTSS PVOID
 #endif
-
+#ifdef UEFIBOOT
```
HBelusca, line 179:
```suggestion

#ifdef UEFIBOOT
```

boot/freeldr/freeldr/ntldr/wlmemory.c
```diff
@@ -176,7 +176,10 @@ MempSetupPagingForRegion(
...
+extern PVOID OsLoaderBase;
+extern SIZE_T OsLoaderSize;
+#endif
```
HBelusca, line 182:
```suggestion
#endif

```

boot/freeldr/freeldr/uefi.cmake
```diff
@@ -25,9 +26,6 @@ list(APPEND UEFILDR_ARC_SOURCE
 if(ARCH STREQUAL "i386")
     list(APPEND UEFILDR_COMMON_ASM_SOURCE
         arch/i386/i386trap.S)
-
-elseif(ARCH STREQUAL "amd64")
```
HBelusca, line 29: dunno, maybe this will be needed later on, so it may be kept...

boot/freeldr/freeldr/ntldr/wlmemory.c
```diff
@@ -330,6 +333,9 @@ WinLdrSetupMemoryLayout(IN OUT PLOADER_PARAMETER_BLOCK LoaderBlock)
 
     WinLdrpDumpMemoryDescriptors(LoaderBlock); //FIXME: Delete!
 
+#ifdef UEFIBOOT
+    MempSetupPaging((ULONG_PTR)OsLoaderBase >> MM_PAGE_SHIFT, OsLoaderSize >> MM_PAGE_SHIFT, FALSE);
```
HBelusca, line 337: Looks like you can merge that new thing with the next commented code :)

boot/freeldr/freeldr/arch/uefi/uefimem.c
```diff
@@ -303,10 +314,21 @@ UefiExitBootServices(VOID)
...
 UefiPrepareForReactOS(VOID)
 {
+    ULONG StackVar = 0;
+    StackPtrNew = &StackVar;
+    StackPtrEnd = (&StackVar + 0x2000); /* Technically UEFI defines that stack space as 32kb*/
     UefiExitBootServices();
-    ExitStack = MmAllocateMemoryWithType(EXIT_STACK_SIZE, LoaderOsloaderStack);
-    EndofExitStack = (PVOID)((ULONG_PTR)ExitStack + EXIT_STACK_SIZE);
+    //TODO: Let's take another look at this
+    MempSetupPaging((ULONG_PTR)StackPtrEnd >> MM_PAGE_SHIFT, 0x2000 >> MM_PAGE_SHIFT, FALSE);
```
HBelusca, line 333: This call cannot be done before the UefiExitBootServices call?

DarkFire01: Doesn’t matter either way

boot/freeldr/freeldr/arch/uefi/ueficon.c
DarkFire01: One some hardware I’ve found this isn’t fully supported as an operation so it’s solely just to lower minimal requirements,

also it’s a bit buggy in virtualbox (sometimes it makes long and press not register

boot/freeldr/freeldr/arch/uefi/uefimem.c
HBelusca: I was asking, because maybe in one case you could move those StackPtr and StackPtrEnd variables into the function itself (instead of keeping them as globals)

boot/freeldr/freeldr/arch/uefi/uefimem.c
```diff
@@ -303,10 +314,21 @@ UefiExitBootServices(VOID)
...
+                IN BOOLEAN KernelMapping);
+
+ULONG* StackPtrNew;
+ULONG* StackPtrEnd;
 VOID
 UefiPrepareForReactOS(VOID)
 {
+    ULONG StackVar = 0;
+    StackPtrNew = &StackVar;
+    StackPtrEnd = (&StackVar + 0x2000); /* Technically UEFI defines that stack space as 32kb*/
```
binarymaster, line 330:
```suggestion
    ULONG StackVar = 0;

    /* Technically UEFI defines that stack space as 32kb */
    StackPtrNew = &StackVar;
    StackPtrEnd = (&StackVar + 0x2000);
```

boot/freeldr/freeldr/arch/uefi/uefimem.c
```diff
@@ -303,10 +314,23 @@ UefiExitBootServices(VOID)
...
 VOID
 UefiPrepareForReactOS(VOID)
 {
+    ULONG* StackPtrNew;
+    ULONG* StackPtrEnd;
+    ULONG StackVar = 0;
+
+    /* Technically UEFI defines that stack space as 32kb */
+    StackPtrNew = &StackVar;
+    StackPtrEnd = (&StackVar + 0x2000);
```
binarymaster, line 332:
```suggestion
    StackPtrEnd = (StackPtrNew + 0x2000 / sizeof(StackPtrNew));
```
In order to fix error:
```
../src/boot/freeldr/freeldr/arch/uefi/uefimem.c:326:12: error: variable 'StackPtrNew' set but not used [-Werror=unused-but-set-variable]
     ULONG* StackPtrNew;
            ^~~~~~~~~~~
```

boot/freeldr/freeldr/arch/uefi/uefildr.c
```diff
@@ -25,12 +28,50 @@ EfiEntry(
...
+        UiMessageBoxCritical("Unable to initialize memory manager.");
+        //goto Quit;
+    }
+
+    /* Initialize I/O subsystem */
+    FsInit();
+
+    /* 0x32000 is what UEFI defines, but we can go smaller if we want */
+    BasicStack = (PVOID)((ULONG_PTR)0x32000 + (ULONG_PTR)MmAllocateMemoryWithType(0x32000, LoaderLoadedProgram));
+    _changestack();
```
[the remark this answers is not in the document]
DarkFire01: We NEED to use the UEFI stack for every firmware call back it was way less janky to do this way.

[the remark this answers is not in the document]
DarkFire01: Here lemme be more detailed:
We call into a callback that needs the UEFI stack 5+ times before we call RunLoader();
afterword we need to call ExitBootServices

The idea here is that instead of calling that function directly we restore the UEFI stack and return to how we were in one assembly call.

This removes a lot of the instability, and ensures that we can modify uefimem.c later down the line.

From what i understand we can still use uefidisk.c on our new stack; As this is seemingly what GRUB and Limine do.

boot/freeldr/freeldr/arch/uefi/amd64/uefiasm.S
```diff
@@ -0,0 +1,90 @@
+/*
+ * PROJECT:     FreeLoader UEFI Support
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     UEFI Assembly helpers
```
binarymaster, line 4:
```suggestion
 * PURPOSE:     x64 assembly helpers
```

binarymaster: Maybe: `[ARCH] assembly UEFI escape helper functions`

boot/freeldr/freeldr/arch/uefi/uefimem.c
```diff
@@ -264,11 +278,13 @@ UefiMemGetMemoryMap(ULONG *MemoryMapSize)
         MapEntry = NEXT_MEMORY_DESCRIPTOR(MapEntry, DescriptorSize);
     }
 
+    /* mark the first page as on some UEFI systems it can be free. */
```
binarymaster, line 281:
```suggestion
    /* Reserve the first page, as on some UEFI systems it can be free */
```

boot/freeldr/freeldr/arch/uefi/amd64/uefiasm.S
```diff
@@ -0,0 +1,90 @@
...
+    .word HEX(FFFF), HEX(0000), HEX(9200), HEX(0000) /* 28: 16-bit real mode DS */
+    .word HEX(FFFF), HEX(0000), HEX(9B00), HEX(00CF) /* 30: compat mode CS */
+
+/* GDT table pointer */
+_gdtptr:
+    .word HEX(37)       /* Limit */
+#ifdef _USE_ML
+    .quad gdt    /* Base Address */
+#else
+    .quad gdt, 0           /* Base Address */
```
JoachimHenze, line 87:
```suggestion
    .word HEX(37) /* Limit */
#ifdef _USE_ML
    .quad gdt     /* Base Address */
#else
    .quad gdt, 0  /* Base Address */
```

boot/freeldr/freeldr/arch/uefi/i386/uefiasm.S
```diff
@@ -0,0 +1,119 @@
...
+    .word HEX(0000)
+
+/* GDT table pointer */
+_gdtptr:
+    .word HEX(27)       /* Limit */
+    .long gdt, 0          /* Base Address */
+
+_i386idtptr:
+    .word 255           /* Limit */
+    .long _i386Idt     /* Base Address */
```
JoachimHenze, line 118:
```suggestion
    .word HEX(27)  /* Limit */
    .long gdt, 0   /* Base Address */

_i386idtptr:
    .word 255      /* Limit */
    .long _i386Idt /* Base Address */
```

boot/freeldr/freeldr/arch/uefi/uefildr.c
```diff
@@ -25,12 +30,50 @@ EfiEntry(
...
+    /* Debugger pre-initialization */
+    DebugInit(0);
+
+    MachInit("");
+
+    /* UI pre-initialization */
+    if (!UiInitialize(FALSE))
+    {
+        UiMessageBoxCritical("Unable to initialize UI.");
+        //goto Quit;
```
JoachimHenze, line 42:
```suggestion
        //goto Quit;
```
why leaving that commented out code inside?
If it's needed, uncomment it.
**If it's not needed, strip it.**
If it would be needed actually, but causes issues due to bugs otherwhere, please add a comment about what would go wrong if it was enabled and where you believe a fix should go.

boot/freeldr/freeldr/arch/uefi/uefildr.c
```diff
@@ -25,12 +30,50 @@ EfiEntry(
...
+    }
+
+    /* Initialize memory manager */
+    if (!MmInitializeMemoryManager())
+    {
+        UiMessageBoxCritical("Unable to initialize memory manager.");
+        //goto Quit;
```
JoachimHenze, line 49: same here.

boot/freeldr/freeldr/arch/uefi/uefildr.c
```diff
@@ -25,12 +30,50 @@ EfiEntry(
...
 }
 
+void
+ExecuteLoaderCleanly(PVOID PreviousStack)
+{
+    UefiServiceStack = PreviousStack;
+    if (UefiServiceStack == NULL)
+    {
+        TRACE("Assigning stack has failed\n");
+    }
```
JoachimHenze, line 70:
```suggestion
        TRACE("Assigning stack has failed\n");

```

boot/freeldr/freeldr/arch/uefi/uefihw.c
```diff
@@ -0,0 +1,149 @@
...
+            TRACE("ACPI 1.0, using RSDT address\n");
+            AcpiBiosData->RSDTAddress.LowPart = Rsdp->rsdt_physical_address;
+        }
+
+        AcpiBiosData->Count = FreeldrDescCount;
+        memcpy(AcpiBiosData->MemoryMap, EfiMemoryMap,
+            FreeldrDescCount * sizeof(BIOS_MEMORY_MAP));
+
+        TRACE("RSDT %p, data size %x\n", Rsdp->rsdt_physical_address,
+            TableSize);
```
JoachimHenze, line 106:
```suggestion
        TRACE("RSDT %p, data size %x\n", Rsdp->rsdt_physical_address, TableSize);
```
Not line-breaking here (because you are too afraid of using 82 columns here, does not make sense if you at the same time and in the same file accept using till column 108 at line 37. That is much longer, and that is what induces the need for a horizontal slider first. Not this TRACE.

boot/freeldr/freeldr/arch/uefi/amd64/uefiasm.S
```diff
@@ -0,0 +1,90 @@
...
+    lgdt fword ptr [_gdtptr]
+#else
+    lgdt cs:[_gdtptr][rip] /* GAS isn't my friend - avoid letting it generate absolute addressing */
+#endif
+    ret
+
+.align 8
+gdt:
+    .word HEX(0000), HEX(0000), HEX(0000), HEX(0000) /* 00: NULL descriptor */
+    .word HEX(0000), HEX(0000), HEX(0000), HEX(0000) /* 08:  */
```
JoachimHenze, line 74:
```suggestion
    .word HEX(0000), HEX(0000), HEX(0000), HEX(0000) /* 08: */
```
The last thing, promised. I do approve already.

JoachimHenze (approved):

boot/freeldr/freeldr/uefi.cmake
```diff
@@ -87,6 +91,7 @@ add_executable(uefildr ${UEFILDR_BASE_SOURCE})
 set_target_properties(uefildr PROPERTIES SUFFIX ".efi")
 
 target_compile_definitions(uefildr PRIVATE UEFIBOOT)
+set_image_base(uefildr 0x10000)
```
HBelusca, line 94: Is this needed?

boot/freeldr/freeldr/arch/uefi/uefildr.c
```diff
@@ -25,15 +30,65 @@ EfiEntry(
...
 
     UNREACHABLE;
     return 0;
 }
 
+void
+ExecuteLoaderCleanly(PVOID PreviousStack)
+{
+    UefiServiceStack = PreviousStack;
+    if (UefiServiceStack == NULL)
```
HBelusca, line 71: This would be NULL if PreviousStack was actually NULL I guess. But if in your call you call ExecuteLoaderCleanly by passing a non-NULL parameter, I don't see how this would ever happen.

boot/freeldr/freeldr/arch/uefi/i386/uefiasm.S
```diff
@@ -0,0 +1,119 @@
...
+    /* All done */
+    ret
+
+// void __reloadsegment(VOID)
+PUBLIC __changestack
+__changestack:
+    mov eax, esp
+    mov esp, _BasicStack
+    push eax
+    call _ExecuteLoaderCleanly
```
HBelusca, line 60: Here -- for example -- you call the ExecuteLoaderCleanly, but I don't see how the old esp (now stored in eax) parameter could have been NULL.

boot/freeldr/freeldr/uefi.cmake
DarkFire01: so badly, UEFI threw freeloader into 8gb+ and I had to change loader pages span to adjust for 8gb!!! This is absolutely ridiculous.

and ReactOS explodes when there’s that many pages marked like this.

boot/freeldr/freeldr/arch/uefi/i386/uefiasm.S
DarkFire01: It’s not physically possible, I was trying to sanity check but you’re right this wouldn’t remotely happen

HBelusca: ok then this should allow you to simplify further that execute.... function.

boot/freeldr/freeldr/uefi.cmake
HBelusca: That's interesting.... Worth its own investigation.

boot/freeldr/freeldr/arch/uefi/amd64/uefiasm.S
```diff
@@ -0,0 +1,90 @@
...
+// void _exituefi(VOID)
+PUBLIC _exituefi
+_exituefi:
+    /* Save non-volatile registers */
+    push rbp
+    push rsi
+    push rdi
+    push rbx
+
+
```
HBelusca, line 28:
```suggestion
```

boot/freeldr/freeldr/arch/uefi/i386/uefiasm.S
```diff
@@ -0,0 +1,119 @@
...
+    /* Retore old stack */
+    mov esp, ebx
+
+    /* Retore non-volatiles */
+    pop ebx
+    pop edi
+    pop esi
+    pop ebp
+
+
```
HBelusca, line 50:
```suggestion
```

boot/freeldr/freeldr/arch/uefi/uefildr.c
```diff
@@ -25,15 +30,58 @@ EfiEntry(
...
+    UNREACHABLE;
+}
+
 #ifndef _M_ARM
 VOID __cdecl Reboot(VOID)
 {
+    //TODO: Replace with a true firmware reboot eventually
+    WARN("Stopping freeloader something has gone wrong - Freezing");
+    for(;;)
+    {
```
HBelusca, line 84:
```suggestion
        NOTHING;
```

boot/freeldr/freeldr/uefi.cmake
DarkFire01: Well UEFI is suppose to respect the base address of the PE file, and when the AMD64 linker doesn’t have one set it’ll assign it pretty high due to ASLR

boot/freeldr/freeldr/arch/uefi/uefimem.c
```diff
@@ -306,7 +326,6 @@ UefiExitBootServices(VOID)
 VOID
 UefiPrepareForReactOS(VOID)
 {
-    UefiExitBootServices();
-    ExitStack = MmAllocateMemoryWithType(EXIT_STACK_SIZE, LoaderOsloaderStack);
-    EndofExitStack = (PVOID)((ULONG_PTR)ExitStack + EXIT_STACK_SIZE);
+    _exituefi();
+    _changegdt();
```
HBelusca, line 330: It looks like these two helper functions could be combined into one? Or do you keep them separated because you think you may need to use them sometimes separately, in the future?

boot/freeldr/freeldr/arch/uefi/uefimem.c
```diff
@@ -256,19 +260,35 @@ UefiMemGetMemoryMap(ULONG *MemoryMapSize)
...
-                      MapEntry->PhysicalStart,
-                      MapEntry->NumberOfPages,
-                      MemoryType);
+        /* Sometimes our loader can be loaded into higher memory than we ever allocate */
+        if (MemoryType == LoaderLoadedProgram)
+        {
+            if (((MapEntry->PhysicalStart + (MapEntry->NumberOfPages * PAGE_SIZE)) >> EFI_PAGE_SHIFT) > LoaderPagesSpanned)
+            {
+                /* This value needs to be adjusted if this occurs */
+                LoaderPagesSpanned = ((MapEntry->PhysicalStart + (MapEntry->NumberOfPages * PAGE_SIZE)) >> EFI_PAGE_SHIFT);
```
HBelusca, line 269: The comments don't really explain what this hack is trying to achieve (counteracting what? etc.). You've just told _**in which case**_ it is needed.

DarkFire01: How is it a hack? The loader spans the pages of the LoaderLoadedProgram? This is exactly what BIOS does but it knows it’s full range immediately

boot/freeldr/freeldr/arch/uefi/uefimem.c
```diff
@@ -256,19 +260,35 @@ UefiMemGetMemoryMap(ULONG *MemoryMapSize)
...
+                          MapEntry->PhysicalStart,
+                          MapEntry->NumberOfPages,
+                          MemoryType);
+        }
 
         MapEntry = NEXT_MEMORY_DESCRIPTOR(MapEntry, DescriptorSize);
     }
 
+    /* Reserve the first page, as on some UEFI systems it can be free */
+    UefiSetMemory(FreeldrMem, 0x000000, 1, LoaderFirmwarePermanent);
```
HBelusca, line 286: I know.... the UEFI version of SetMemory takes the 3rd parameter as a size in number of pages, instead of size in number of bytes, as done in the Setmemory functions of the other "platforms" (BIOS, PC-98, XBOX).
Why this difference? Is it possible to go back to size in bytes (and then the function would divide by the size of one page for the UEFI platform :thinking: )

boot/freeldr/freeldr/arch/uefi/uefimem.c
HBelusca: Ah yes thanks for pointing. Actually this variable is indeed checked (and re-calculated if needed) like every time there is a memory allocation.

boot/freeldr/freeldr/arch/uefi/uefimem.c
DarkFire01: UEFI gives you the size in pages, and this is UefiSetMemory. I just wanted it to be clean instead of adding * PAGE_SIZE to the UefiSetMemory calls in the loader map.

boot/freeldr/freeldr/arch/uefi/amd64/uefiasm.S
```diff
@@ -46,6 +45,12 @@ _exituefi:
     pop rsi
     pop rbp
 
+#ifdef _USE_ML
+    lgdt fword ptr [_gdtptr]
+#else
+    lgdt cs:[_gdtptr][rip] /* GAS isn't my friend - avoid letting it generate absolute addressing */
```
HBelusca, line 51: Nice that you've found how to fix that error. Will be helpful in other cases as well.

HBelusca (approved):

boot/freeldr/freeldr/arch/uefi/uefildr.c
```diff
@@ -25,15 +30,58 @@ EfiEntry(
...
+        UiMessageBoxCritical("Unable to initialize memory manager.");
+        goto Quit;
+    }
+
+    /* Initialize I/O subsystem */
+    FsInit();
+
+    /* 0x32000 is what UEFI defines, but we can go smaller if we want */
+    BasicStack = (PVOID)((ULONG_PTR)0x32000 + (ULONG_PTR)MmAllocateMemoryWithType(0x32000, LoaderLoadedProgram));
```
binarymaster, line 56: Shouldn't this be `LoaderOsloaderStack`?

boot/freeldr/freeldr/arch/uefi/uefildr.c
```diff
@@ -25,15 +30,58 @@ EfiEntry(
...
+    RunLoader();
+    UNREACHABLE;
+}
+
 #ifndef _M_ARM
 VOID __cdecl Reboot(VOID)
 {
-
+    //TODO: Replace with a true firmware reboot eventually
+    WARN("Stopping freeloader something has gone wrong - Freezing");
```
binarymaster, line 81:
```suggestion
    WARN("Something has gone wrong - halting FreeLoader\n");
```

boot/freeldr/freeldr/arch/uefi/uefildr.c
```diff
@@ -25,15 +30,58 @@ EfiEntry(
...
+    for(;;)
```
binarymaster, line 82:
```suggestion
    for (;;)
```

boot/freeldr/freeldr/arch/uefi/uefimem.c
```diff
@@ -256,19 +259,35 @@ UefiMemGetMemoryMap(ULONG *MemoryMapSize)
...
+            UefiSetMemory(FreeldrMem,
+                          MapEntry->PhysicalStart,
+                          MapEntry->NumberOfPages,
+                          MemoryType);
+        }
 
         MapEntry = NEXT_MEMORY_DESCRIPTOR(MapEntry, DescriptorSize);
     }
 
+    /* Reserve the first page, as on some UEFI systems it can be free */
```
binarymaster, line 284: The comment is not very clear, since a free page is rather should be utilized by the system, than using it as justification for reserving that page as permanently occupied by firmware.

DarkFire01: We hard coded this is ntoskrnl as a reserved space like windows 2003 does, if we let windows use this as free it’ll assert, both win2k3 and ROS

binarymaster:
```suggestion
    /* Windows expects the first page to be reserved, otherwise it asserts.
     * However it can be just a free page on some UEFI systems. */
```

HBelusca: By the way, I've just seen we have a "ReserveMemory" in uefi. Is it used?

Regarding the first page reservation, this is indeed done in PC-BIOS as well:
https://git.reactos.org/?p=reactos.git;a=blob;f=boot/freeldr/freeldr/arch/i386/pc/pcmem.c;hb=a6b281c2280710e684a208deb07abd9838ce7cb6#l654
as well as in PC-98:
https://git.reactos.org/?p=reactos.git;a=blob;f=boot/freeldr/freeldr/arch/i386/pc98/pc98mem.c;hb=a6b281c2280710e684a208deb07abd9838ce7cb6#l75
but I wonder where exactly this is done for XBOX (and if not, how ROS can cope with that); I guess it's one of these calls but not sure:
https://git.reactos.org/?p=reactos.git;a=blob;f=boot/freeldr/freeldr/arch/i386/xbox/xboxmem.c;hb=a6b281c2280710e684a208deb07abd9838ce7cb6#l250

binarymaster (approved):
I think this is fine.

boot/freeldr/freeldr/arch/uefi/i386/uefiasm.S
```diff
@@ -0,0 +1,126 @@
...
+__changestack:
+    mov eax, esp
+    mov esp, _BasicStack
+    push eax
+    call _ExecuteLoaderCleanly
+    ret
+
+
+PUBLIC __changegdt
+__changegdt:
```
HBelusca, line 72: This function can now be removed.

HBelusca (approved):
Minor thing here, but otherwise LGTM so I approve!

HBelusca:
In commit log title: [UEFI] --> [FREELDR]  ,  since this is a compilable module name.

PR #5282, opened 2023-05-12
[SHELL32] Fix for shlexec.cpp regressions

Doug-Lyons:
## Purpose

_Revert https://github.com/reactos/reactos/commit/0f16d44b660266130f5c0807c01d01cfaa6e685d._

JIRA issue: [CORE-18967](https://jira.reactos.org/browse/CORE-18967)

## Proposed changes

_Add conditional 'if' to code so that it is only executed when needed._

Testbot results:

[shlexec-cpp-05.patch](https://jira.reactos.org/secure/attachment/65772/65772_shlexec-cpp-05.patch) JID65772 on top of 0.4.15-dev-6010-g30bc0b6

VBox: https://reactos.org/testman/compare.php?ids=87117,87120 LGTM
KVM: https://reactos.org/testman/compare.php?ids=87116,87119 LGTM

All three (3) "blocks" issues checked and appear OK.

### DIFF
--- a/dll/win32/shell32/shlexec.cpp
+++ b/dll/win32/shell32/shlexec.cpp
@@ -2125,7 +2125,7 @@ static BOOL SHELL_execute(LPSHELLEXECUTEINFOW sei, SHELL_ExecuteW32 execfunc)
             lpFile = wfileName;
         }
         /* We have to test sei instead of sei_tmp because sei_tmp had its
-         * input fMask modifed above in SHELL_translate_idlist.
+         * input fMask modified above in SHELL_translate_idlist.
          * This code is needed to handle the case where we only have an
          * lpIDList with multiple CLSID/PIDL's (not 'My Computer' only) */
         else if ((sei->fMask & SEE_MASK_IDLIST) == SEE_MASK_IDLIST)
--- a/dll/win32/shell32/shlexec.cpp
+++ b/dll/win32/shell32/shlexec.cpp
@@ -2125,25 +2125,32 @@ static BOOL SHELL_execute(LPSHELLEXECUTEINFOW sei, SHELL_ExecuteW32 execfunc)
             lpFile = wfileName;
         }
         else if (!PathIsDirectoryW(wszApplicationName) &&
-            !PathIsExeW(wszApplicationName) && 
-            (sei_tmp.fMask & SEE_MASK_INVOKEIDLIST) != SEE_MASK_INVOKEIDLIST &&
-            (sei_tmp.fMask & SEE_MASK_HASLINKNAME) != SEE_MASK_HASLINKNAME &&
-            (sei_tmp.fMask & SEE_MASK_FLAG_NO_UI) != SEE_MASK_FLAG_NO_UI)
+                 !PathIsExeW(wszApplicationName) &&
+                 (sei_tmp.fMask & SEE_MASK_INVOKEIDLIST) != SEE_MASK_INVOKEIDLIST &&
+                 (sei_tmp.fMask & SEE_MASK_HASLINKNAME) != SEE_MASK_HASLINKNAME &&
+                 (sei_tmp.fMask & SEE_MASK_FLAG_NO_UI) != SEE_MASK_FLAG_NO_UI)
         {
-            /* If the executable name is not quoted, we have to use this search loop here,
-               that in CreateProcess() is not sufficient because it does not handle shell links. */
+            /* This handles the case where wszApplicationName contains a
+             * non-double-quoted executable followed by one or more spaces
+             * and then followed by a double-quoted parameter.
+             * Example: (single quotes used to enclose strings)
+             * (In)  wszApplicationName = 'explorer.exe "C:\Program Files"'
+             * (In)  wszParameters      = ''
+             * (out) wszApplicationName = 'explorer.exe'
+             * (Out) wszParameters      = '"C:\Program Files"'
+             */
             WCHAR buffer[MAX_PATH], xlpFile[MAX_PATH];
             LPWSTR space, s;
 
-            LPWSTR beg = wszApplicationName/*sei_tmp.lpFile*/;
+            LPWSTR beg = wszApplicationName;
             for(s = beg; (space = const_cast<LPWSTR>(strchrW(s, L' '))); s = space + 1)
             {
                 int idx = space - sei_tmp.lpFile;
                 memcpy(buffer, sei_tmp.lpFile, idx * sizeof(WCHAR));
                 buffer[idx] = '\0';
 
-                /*FIXME This finds directory paths if the targeted file name contains spaces. */
-                if (SearchPathW(*sei_tmp.lpDirectory ? sei_tmp.lpDirectory : NULL, buffer, L".exe", sizeof(xlpFile) / sizeof(xlpFile[0]), xlpFile, NULL))
+                if (SearchPathW(*sei_tmp.lpDirectory ? sei_tmp.lpDirectory : NULL,
+                    buffer, L".exe", _countof(xlpFile), xlpFile, NULL))
                 {
                     /* separate out command from parameter string */
                     LPCWSTR p = space + 1;
--- a/dll/win32/shell32/shlexec.cpp
+++ b/dll/win32/shell32/shlexec.cpp
@@ -2124,21 +2124,12 @@ static BOOL SHELL_execute(LPSHELLEXECUTEINFOW sei, SHELL_ExecuteW32 execfunc)
             *end = L'\0';
             lpFile = wfileName;
         }
-        else if (!PathIsDirectoryW(wszApplicationName) &&
-                 !PathIsExeW(wszApplicationName) &&
-                 (sei_tmp.fMask & SEE_MASK_INVOKEIDLIST) != SEE_MASK_INVOKEIDLIST &&
-                 (sei_tmp.fMask & SEE_MASK_HASLINKNAME) != SEE_MASK_HASLINKNAME &&
-                 (sei_tmp.fMask & SEE_MASK_FLAG_NO_UI) != SEE_MASK_FLAG_NO_UI)
+        /* We have to test sei instead of sei_tmp because sei_tmp had its
+         * input fMask modifed above in SHELL_translate_idlist.
+         * This code is needed to handle the case where we only have an
+         * lpIDList with multiple CLSID/PIDL's (not 'My Computer' only) */
+        else if ((sei->fMask & SEE_MASK_IDLIST) == SEE_MASK_IDLIST)
         {
-            /* This handles the case where wszApplicationName contains a
-             * non-double-quoted executable followed by one or more spaces
-             * and then followed by a double-quoted parameter.
-             * Example: (single quotes used to enclose strings)
-             * (In)  wszApplicationName = 'explorer.exe "C:\Program Files"'
-             * (In)  wszParameters      = ''
-             * (out) wszApplicationName = 'explorer.exe'
-             * (Out) wszParameters      = '"C:\Program Files"'
-             */
             WCHAR buffer[MAX_PATH], xlpFile[MAX_PATH];
             LPWSTR space, s;
 
--- a/dll/win32/shell32/shlexec.cpp
+++ b/dll/win32/shell32/shlexec.cpp
@@ -2124,6 +2124,41 @@ static BOOL SHELL_execute(LPSHELLEXECUTEINFOW sei, SHELL_ExecuteW32 execfunc)
             *end = L'\0';
             lpFile = wfileName;
         }
+        else if (!PathIsDirectoryW(wszApplicationName) &&
+            !PathIsExeW(wszApplicationName) && 
+            (sei_tmp.fMask & SEE_MASK_INVOKEIDLIST) != SEE_MASK_INVOKEIDLIST &&
+            (sei_tmp.fMask & SEE_MASK_HASLINKNAME) != SEE_MASK_HASLINKNAME &&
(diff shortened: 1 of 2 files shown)


KRosUser2 (approved):
Tested OK against the regression raised by the JIRA ticket

dll/win32/shell32/shlexec.cpp
```diff
@@ -2124,6 +2124,41 @@ static BOOL SHELL_execute(LPSHELLEXECUTEINFOW sei, SHELL_ExecuteW32 execfunc)
             *end = L'\0';
             lpFile = wfileName;
         }
+        else if (!PathIsDirectoryW(wszApplicationName) &&
+            !PathIsExeW(wszApplicationName) &&
```
JoachimHenze, line 2128:
```suggestion
            !PathIsExeW(wszApplicationName) &&
```

dll/win32/shell32/shlexec.cpp
```diff
@@ -2124,6 +2124,41 @@ static BOOL SHELL_execute(LPSHELLEXECUTEINFOW sei, SHELL_ExecuteW32 execfunc)
             *end = L'\0';
             lpFile = wfileName;
         }
+        else if (!PathIsDirectoryW(wszApplicationName) &&
+            !PathIsExeW(wszApplicationName) && 
+            (sei_tmp.fMask & SEE_MASK_INVOKEIDLIST) != SEE_MASK_INVOKEIDLIST &&
+            (sei_tmp.fMask & SEE_MASK_HASLINKNAME) != SEE_MASK_HASLINKNAME &&
+            (sei_tmp.fMask & SEE_MASK_FLAG_NO_UI) != SEE_MASK_FLAG_NO_UI)
```
HBelusca, line 2131:
```suggestion
        else if (!PathIsDirectoryW(wszApplicationName) &&
                 !PathIsExeW(wszApplicationName) &&
                 (sei_tmp.fMask & SEE_MASK_INVOKEIDLIST) != SEE_MASK_INVOKEIDLIST &&
                 (sei_tmp.fMask & SEE_MASK_HASLINKNAME) != SEE_MASK_HASLINKNAME &&
                 (sei_tmp.fMask & SEE_MASK_FLAG_NO_UI) != SEE_MASK_FLAG_NO_UI)
```

dll/win32/shell32/shlexec.cpp
```diff
@@ -2124,6 +2124,41 @@ static BOOL SHELL_execute(LPSHELLEXECUTEINFOW sei, SHELL_ExecuteW32 execfunc)
...
+
+            LPWSTR beg = wszApplicationName/*sei_tmp.lpFile*/;
+            for(s = beg; (space = const_cast<LPWSTR>(strchrW(s, L' '))); s = space + 1)
+            {
+                int idx = space - sei_tmp.lpFile;
+                memcpy(buffer, sei_tmp.lpFile, idx * sizeof(WCHAR));
+                buffer[idx] = '\0';
+
+                /*FIXME This finds directory paths if the targeted file name contains spaces. */
+                if (SearchPathW(*sei_tmp.lpDirectory ? sei_tmp.lpDirectory : NULL, buffer, L".exe", sizeof(xlpFile) / sizeof(xlpFile[0]), xlpFile, NULL))
```
HBelusca, line 2146:
```suggestion
                if (SearchPathW(*sei_tmp.lpDirectory ? sei_tmp.lpDirectory : NULL, buffer, L".exe", _countof(xlpFile), xlpFile, NULL))
```

dll/win32/shell32/shlexec.cpp
```diff
@@ -2124,6 +2124,41 @@ static BOOL SHELL_execute(LPSHELLEXECUTEINFOW sei, SHELL_ExecuteW32 execfunc)
...
+            LPWSTR space, s;
+
+            LPWSTR beg = wszApplicationName/*sei_tmp.lpFile*/;
+            for(s = beg; (space = const_cast<LPWSTR>(strchrW(s, L' '))); s = space + 1)
+            {
+                int idx = space - sei_tmp.lpFile;
+                memcpy(buffer, sei_tmp.lpFile, idx * sizeof(WCHAR));
+                buffer[idx] = '\0';
+
+                /*FIXME This finds directory paths if the targeted file name contains spaces. */
```
HBelusca, line 2145: This does not explain why it's a fixme...

dll/win32/shell32/shlexec.cpp
```diff
@@ -2124,6 +2124,41 @@ static BOOL SHELL_execute(LPSHELLEXECUTEINFOW sei, SHELL_ExecuteW32 execfunc)
...
+            (sei_tmp.fMask & SEE_MASK_INVOKEIDLIST) != SEE_MASK_INVOKEIDLIST &&
+            (sei_tmp.fMask & SEE_MASK_HASLINKNAME) != SEE_MASK_HASLINKNAME &&
+            (sei_tmp.fMask & SEE_MASK_FLAG_NO_UI) != SEE_MASK_FLAG_NO_UI)
+        {
+            /* If the executable name is not quoted, we have to use this search loop here,
+               that in CreateProcess() is not sufficient because it does not handle shell links. */
+            WCHAR buffer[MAX_PATH], xlpFile[MAX_PATH];
+            LPWSTR space, s;
+
+            LPWSTR beg = wszApplicationName/*sei_tmp.lpFile*/;
```
HBelusca, line 2138: which one to use?

HBelusca (commented):
Does Wine have some fixes regarding this code path?

dll/win32/shell32/shlexec.cpp
Doug-Lyons: Fixed. Thanks.

dll/win32/shell32/shlexec.cpp
Doug-Lyons: Fixed. Thanks.

dll/win32/shell32/shlexec.cpp
Doug-Lyons: Added comments showing what this section of code does. Removed this old comment because it no longer applies.

dll/win32/shell32/shlexec.cpp
Doug-Lyons: The 'wszApplicationName' is correct, but the 'sei_tmp.lpFile' is always equal to wszApplicationName if we get to this point.

Doug-Lyons:
> Does Wine have some fixes regarding this code path?

Wine has many changes through its current version of 8.8, but none in this area of the code. I spent about 3 hours doing a partial sync with the wine code and after testing it, there were no improvement here. Thanks.

dll/win32/shell32/shlexec.cpp
Doug-Lyons: Fixed. Thanks.

JoachimHenze (approved):
Approved. I would welcome too if we could write some unit-tests for that tough. I do like that you put a big comment at least in there explaining the cases this code is still needed for to be handled correctly.

dll/win32/shell32/shlexec.cpp
```diff
@@ -2124,6 +2124,48 @@ static BOOL SHELL_execute(LPSHELLEXECUTEINFOW sei, SHELL_ExecuteW32 execfunc)
             *end = L'\0';
             lpFile = wfileName;
         }
+        else if (!PathIsDirectoryW(wszApplicationName) &&
+                 !PathIsExeW(wszApplicationName) &&
```
learn-more, line 2128: These checks seem weird,
what happens on a command line like this?
`"test.exe example.exe"`

Doug-Lyons: @learn-more, I have been researching this a bit and in my most recent revision (included here now) I have only a single condition. From what I learned, this code is only used when we have an fMask of 4 (SEE_MASK_IDLIST) and then we have the SHELLEXECUTEINFOW structure with only the lpIDList active. Otherwise, hwnd = NULL, lpFile = NULL, lpParameters = NULL, lpDirectory = NULL, and lpClass is 'not used', so this is a pretty narrow range of use. I have a proof-of-concept regression test for this and I plan to create a PR for it soon. Thanks.

dll/win32/shell32/shlexec.cpp
```diff
@@ -2124,6 +2124,39 @@ static BOOL SHELL_execute(LPSHELLEXECUTEINFOW sei, SHELL_ExecuteW32 execfunc)
...
+         * input fMask modifed above in SHELL_translate_idlist.
+         * This code is needed to handle the case where we only have an
+         * lpIDList with multiple CLSID/PIDL's (not 'My Computer' only) */
+        else if ((sei->fMask & SEE_MASK_IDLIST) == SEE_MASK_IDLIST)
+        {
+            WCHAR buffer[MAX_PATH], xlpFile[MAX_PATH];
+            LPWSTR space, s;
+
+            LPWSTR beg = wszApplicationName;
+            for(s = beg; (space = const_cast<LPWSTR>(strchrW(s, L' '))); s = space + 1)
```
learn-more, line 2137: How will this handle e.g.:
`{B4BFCC3A-DB2C-424C-B029-7FE99A87C641} something notepad`.

dll/win32/shell32/shlexec.cpp
```diff
@@ -2124,6 +2124,39 @@ static BOOL SHELL_execute(LPSHELLEXECUTEINFOW sei, SHELL_ExecuteW32 execfunc)
...
+            {
+                int idx = space - sei_tmp.lpFile;
+                memcpy(buffer, sei_tmp.lpFile, idx * sizeof(WCHAR));
+                buffer[idx] = '\0';
+
+                if (SearchPathW(*sei_tmp.lpDirectory ? sei_tmp.lpDirectory : NULL,
+                    buffer, L".exe", _countof(xlpFile), xlpFile, NULL))
```
learn-more, line 2144: Should this only be `.exe` or also `.bat`, `.cmd` etc?

dll/win32/shell32/shlexec.cpp
learn-more: Or is that not possible?

Doug-Lyons: When I test this I get either a "Bad command or filename - shell:::{CLSID}" or it is ignored with this patch. Do you think that this section of code should support the shell CLSID's?
image

dll/win32/shell32/shlexec.cpp
Doug-Lyons: I am not sure about this. My intent was just to fix the issue of https://jira.reactos.org/browse/CORE-18967 which was caused by a previous deletion of parts of this section of code in https://github.com/reactos/reactos/commit/0f16d44b660266130f5c0807c01d01cfaa6e685d. Do you need more than this? Thanks.

dll/win32/shell32/shlexec.cpp
Doug-Lyons: I have added a new regression test into this PR for shell32:ShellExecuteEx for 'test_sei_lpIDList()'. If this test is added to the current master, it will just hang the test until the rosautotest times out after about 3 minutes. The extra test succeeds when it is committed along with the others changes in this PR.

dll/win32/shell32/shlexec.cpp
```diff
@@ -2124,6 +2124,39 @@ static BOOL SHELL_execute(LPSHELLEXECUTEINFOW sei, SHELL_ExecuteW32 execfunc)
             *end = L'\0';
             lpFile = wfileName;
         }
+        /* We have to test sei instead of sei_tmp because sei_tmp had its
+         * input fMask modifed above in SHELL_translate_idlist.
```
JoachimHenze, line 2128:
```suggestion
         * input fMask modified above in SHELL_translate_idlist.
```

modules/rostests/apitests/shell32/ShellExecuteEx.cpp
```diff
@@ -467,6 +467,48 @@ static void test_properties()
...
+    ShellExecInfo.cbSize = sizeof(ShellExecInfo);
+    ShellExecInfo.fMask = SEE_MASK_IDLIST;
+    ShellExecInfo.hwnd = NULL;
+    ShellExecInfo.nShow = SW_SHOWNORMAL;
+    ShellExecInfo.lpFile = NULL;
+    ShellExecInfo.lpDirectory = NULL;
+    ShellExecInfo.lpIDList = lpBytes;
+
+    Result = ShellExecuteExW(&ShellExecInfo);
+    ok(Result == TRUE, "ShellExecuteEx lpIDList 'C:\\' failed.\n");
```
JoachimHenze, line 501:
```suggestion
    ok(Result == TRUE, "ShellExecuteEx lpIDList 'C:\\' failed\n");
```

modules/rostests/apitests/shell32/ShellExecuteEx.cpp
```diff
@@ -467,6 +467,48 @@ static void test_properties()
...
+    trace("sei_lpIDList returned: %s.\n", Result ? "SUCCESS" : "FAILURE");
```
JoachimHenze, line 502:
```suggestion
    trace("sei_lpIDList returned: %s\n", Result ? "SUCCESS" : "FAILURE");
```

dll/win32/shell32/shlexec.cpp
Doug-Lyons: Fixed. Thanks.

modules/rostests/apitests/shell32/ShellExecuteEx.cpp
Doug-Lyons: Changed. Thanks.

modules/rostests/apitests/shell32/ShellExecuteEx.cpp
Doug-Lyons: Changed. Thanks.

PR #5283, opened 2023-05-12
[NTOS:KE/x64] Fix KiConvertToGuiThread

tkreuzer:
## Purpose
This PR fixes some issues on x64, where threads that already had a large GUI stack were switched to a new large stack, deallocating the previous stack, as if it was a normal stack. With the current stack allocator, that uses system PTEs and doesn't keep track on the size of the allocation, this would have leaked system PTEs.
It also fixes a broken success check, which would always succeed, and in case of a failure execute the systemcall over and over again.

## Proposed changes
- Do not allocate a new stack, if the thread already has a large one. This prevents the function from freeing a large stack as a normal stack and subsequently leaking system PTEs.
- Fix the check for failure of PsConvertToGuiThread (test eax, not rax, for being negative, because by default rax is zero extended from eax, not sign extended). This fixes an infinite loop on failure.

### DIFF
--- a/ntoskrnl/ke/amd64/trap.S
+++ b/ntoskrnl/ke/amd64/trap.S
@@ -1041,6 +1041,11 @@ FUNC KiConvertToGuiThread
     .allocstack 40
     .endprolog
 
+    /* Check if we already have a large stack */
+    mov rax, gs:[PcCurrentThread]
+    cmp byte ptr [rax + KTHREAD_LargeStack], 0
+    jnz AlreadyLargeStack
+
     // NewStack = (ULONG_PTR)MmCreateKernelStack(TRUE, 0);
     mov cl, 1
     xor rdx, rdx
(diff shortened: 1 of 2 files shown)


Extravert-ir:
Why `KiConvertToGuiThread` is written in assembly at all? Looking at it, I'd say it should be rewritten in C

tkreuzer:
> Why `KiConvertToGuiThread` is written in assembly at all? Looking at it, I'd say it should be rewritten in C

Switching the stack from C code is not legal. Yes, we do it for x86, but it's technically wrong. It's only allowed, if you start fresh and keep the old stack (e.g. `KeExpandKernelStackAndCallout`)

Extravert-ir:
I thought the switch is done inside `KiSystemCall64Again`, isn't it?

tkreuzer:
> I thought the switch is done inside `KiSystemCall64Again`, isn't it?

You obviously didn't even view this PR, otherwise you would know the answer.

PR #5289, opened 2023-05-18
[UTILMAN] Fix content order in IDC_LISTBOX.

zecarlos1957:
Fix the content order in IDC_LISTBOX.

OSK et Magnify are mixed at initialization (language dependent)
LBS_STANDARD includes LBS_SORT style that changes the content in the listbox

JIRA issue: [CORE-18274](https://jira.reactos.org/browse/CORE-18274)
Patch by KRosUser.


HBelusca (approved):
lgtm, but let's @GeoB99 think whether using LB_INSERTSTRING with wParam == -1 instead of LB_ADDSTRING would be also better for robustness.

KRosUser2 (approved):

GeoB99 (approved):
I don't have a strong preference over the one or the other, both LB_INSERTSTRING and LB_ADDSTRING work identically since the `LBS_STANDARD` style is removed the latter message ListBox message won't sort the elements in the said ListBox.

So use whichever.

julenuri:
Hi @zecarlos1957 , Could you please update `pt-PT.rc` file with the same change too. It has been added recently. Thanks.

julenuri (changes requested):
Please, change all the headers to:
 
```
* PROJECT:     ReactOS Win32 BrowseUI
```

julenuri (approved):

PR #5291, opened 2023-05-18
[USER32] RegisterClassExWOWW/ExA: NULL or IS_ATOM class are treated the same.

HBelusca:
## Purpose

Fixes NULL pointer access when e.g. class.lpszClassName == NULL and the
class is registered. The RegisterClass(ExA/W/...) should return 0
instead of throw an invalid access exception.

Noticed by @julenuri.

JIRA issue: [CORE-18978](https://jira.reactos.org/browse/CORE-18978)

### DIFF
--- a/win32ss/user/user32/windows/class.c
+++ b/win32ss/user/user32/windows/class.c
@@ -1310,32 +1310,28 @@ RegisterClassExWOWW(WNDCLASSEXW *lpwcx,
     RtlCopyMemory(&WndClass, lpwcx, sizeof(*lpwcx));
 
     RtlInitEmptyAnsiString(&AnsiMenuName, NULL, 0);
-    if (WndClass.lpszMenuName != NULL)
+    if (!IS_INTRESOURCE(WndClass.lpszMenuName))
     {
-        if (!IS_INTRESOURCE(WndClass.lpszMenuName))
+        if (WndClass.lpszMenuName[0])
         {
-            if (WndClass.lpszMenuName[0])
-            {
-               RtlInitUnicodeString(&MenuName, WndClass.lpszMenuName);
-               RtlUnicodeStringToAnsiString( &AnsiMenuName, &MenuName, TRUE);
-            }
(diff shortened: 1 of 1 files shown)


julenuri:
To investigate.
The PR is still failing:
image
I bet that the issue is more with some gcc "way of being". The code is not failing under MSVC master.
I remember that GCC before the https://github.com/reactos/reactos/commit/72ad37c1cd437c3ef433c01edd4aab2957bea656 is behaving well. Let's see if it's not only a User32 bug.

win32ss/user/user32/windows/class.c
```diff
@@ -1466,32 +1466,29 @@ RegisterClassExWOWW(WNDCLASSEXW *lpwcx,
...
     {
-        ClassName.Length =
-        ClassName.MaximumLength = 0;
-        ClassName.Buffer = (LPWSTR)WndClass.lpszClassName;
+        RtlInitUnicodeString(&ClassName, WndClass.lpszClassName);
     }
     else
     {
-        RtlInitUnicodeString(&ClassName, WndClass.lpszClassName);
+        ClassName.Length =
```
tkreuzer, line 1489: Please put this on one line.

win32ss/user/user32/windows/class.c
```diff
@@ -1466,32 +1466,29 @@ RegisterClassExWOWW(WNDCLASSEXW *lpwcx,
     }
 */
     RtlInitEmptyAnsiString(&AnsiMenuName, NULL, 0);
-    if (WndClass.lpszMenuName != NULL)
+    if (WndClass.lpszMenuName && !IS_INTRESOURCE(WndClass.lpszMenuName))
```
tkreuzer, line 1469: This doesn't seem to change anything. If `WndClass.lpszMenuName` is `NULL`, `MenuName` and `AnsiMenuName` now get assigned a `NULL` buffer, which they were already initialized with.

tkreuzer (approved):

PR #5297, opened 2023-05-19
[WIN32K] Do not try to free a unicode string that is an atom

tkreuzer:
## Purpose

Fixes a bug on an error path, where the code tried to free the buffer of a UNICODE_STRING, even if the string was an ATOM, i.e. the Buffer didn't point anywhere.

### DIFF
--- a/win32ss/user/ntuser/cursoricon.c
+++ b/win32ss/user/ntuser/cursoricon.c
@@ -1665,7 +1665,8 @@ Exit:
     /* Additional cleanup on failure */
     if (bResult == FALSE)
     {
-        if (ustrRsrc.Buffer != NULL)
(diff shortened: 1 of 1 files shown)


win32ss/user/ntuser/cursoricon.c
```diff
@@ -1665,7 +1665,8 @@ NtUserSetCursorIconData(
     /* Additional cleanup on failure */
     if (bResult == FALSE)
     {
-        if (ustrRsrc.Buffer != NULL)
+        if ((ustrRsrc.Buffer != NULL) &&
+            !IS_ATOM(ustrRsrc.Buffer))
```
HBelusca, line 1669: `IS_ATOM` --> `IS_INTRESOURCE` . Where do you see it could be an atom?

PR #5301, opened 2023-05-22
[MSPAINT] Use better units for image resolution

katahiromz:
## Purpose
Our paint program used the pixels per meter (PPM) unit but this unit is hardly used in the world. Let's use DPI (dots per inch) and/or pixels per centimeter (PPCM).

## Proposed changes

- Modify `IDD_ATTRIBUTESTEXT8` to specify the units for image resolution.
- Modify `IDS_PRINTRES` resource strings.
- Use `dpi` for English/American people.
- Use `ppcm` for the other countries / areas.
- Reorder `IDD_ATTRIBUTES` controls (for keyboard usability).

## Comparison

BEFORE:
before
AFTER:
after

## TODO

- [x] Legal settlement.
- [x] Localization.


HBelusca (commented):
I'm still thinking whether you could do a bit different than having that hardcoded IDS_LEGALUNITS string :thinking:

base/applications/mspaint/dib.cpp
```diff
@@ -19,7 +19,13 @@ SYSTEMTIME fileTime;
 // Convert DPI (dots per inch) into PPM (pixels per meter)
 float PpmFromDpi(float dpi)
```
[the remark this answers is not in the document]
katahiromz: Fixed.

base/applications/mspaint/dialogs.cpp
```diff
@@ -98,8 +96,18 @@ LRESULT CAttributesDialog::OnInitDialog(UINT nMsg, WPARAM wParam, LPARAM lParam,
...
+    strUnit.LoadString(hProgInstance, IDS_LEGALUNITS);
+
     CString strRes;
-    strRes.Format(IDS_PRINTRES, (INT)PpmFromDpi(g_xDpi), (INT)PpmFromDpi(g_yDpi));
+    if (strUnit == L"dot/inch")
+        strRes.Format(IDS_PRINTRES, ROUND(g_xDpi), ROUND(g_yDpi));
+    else if (strUnit == L"pixel/cm")
+        strRes.Format(IDS_PRINTRES, ROUND(PpcmFromDpi(g_xDpi)), ROUND(PpcmFromDpi(g_yDpi)));
+    else
+        ATLASSERT(0);
```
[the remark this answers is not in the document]
katahiromz: OK. `ATLASSERT(0);` is deleted.

base/applications/mspaint/dialogs.cpp
```diff
@@ -198,7 +204,7 @@ LRESULT CAttributesDialog::OnEdit2(WORD wNotifyCode, WORD wID, HWND hWndCtl, BOO
         else if (IsDlgButtonChecked(IDD_ATTRIBUTESRB2))
```
HBelusca, line 204: Side-question: What's the difference between those two OnEdit1 and OnEdit2 functions? It seems their code is identical -- except for the control ID of one single edit control.
(It seems possible one of those functions can be removed, and take an extra parameter to specify the edit control ID to act upon.)

katahiromz: There has one more difference: `newWidth` and `newHeight`.
I think no need to commonize this functions.

base/applications/mspaint/lang/en-US.rc
```diff
@@ -242,4 +241,6 @@ BEGIN
     IDS_ITALIC "Italic"
     IDS_UNDERLINE "Underline"
     IDS_VERTICAL "Vertical"
+    IDS_LEGALRESUNITS "dot/inch" /* Don't translate this line */
```
HBelusca, line 244: Looking again at the code, it would appear possible to have that string (that remains untranslated) as the default (resource) string of within the corresponding dialog static text control. That would allow not growing the string table unnecessarily.
Then in the `CAttributesDialog::OnInitDialog` function could do a simple `GetDlgItemText` call on the corresponding static text control, instead of calling `strUnit.LoadString(hProgInstance, IDS_LEGALRESUNITS);`
(Note however that this would work only in the OnInitDialog function -- when the dialog is created and default contents being initialized -- and nowhere else, since the static text would then change.)

zecarlos1957: dot per inch (dpi) better no?

katahiromz: OK, fixed.

PR #5307, opened 2023-05-28
[KMTEST][NTOS] Fix ExHardError test failures on Windows

ThFabba:
## Purpose

JIRA issue: [ROSTESTS-382](https://jira.reactos.org/browse/ROSTESTS-382)

### DIFF
--- a/modules/rostests/kmtests/ntos_ex/ExHardError.c
+++ b/modules/rostests/kmtests/ntos_ex/ExHardError.c
@@ -81,6 +81,27 @@ TestHardError(
     UNICODE_STRING String2 = RTL_CONSTANT_STRING(StringBuffer2);
     ULONG_PTR HardErrorParameters[6];
     BOOLEAN Ret;
+    ULONG FailCriticalErrors = SEM_FAILCRITICALERRORS;
+    ULONG HardErrorMode = FailCriticalErrors;
+
+    /* Without SEM_FAILCRITICALERRORS, most of the calls in this
+     * test will just return ResponseReturnToCaller without doing anything.
+     * TODO: We should have test coverage for that.
+     */
+    Status = ZwQueryInformationProcess(ZwCurrentProcess(),
+                                       ProcessDefaultHardErrorMode,
+                                       &HardErrorMode,
+                                       sizeof(HardErrorMode),
+                                       NULL);
+    ok_eq_hex(Status, STATUS_SUCCESS);
+    if (HardErrorMode != FailCriticalErrors)
+    {
+        Status = ZwSetInformationProcess(ZwCurrentProcess(),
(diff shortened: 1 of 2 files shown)


modules/rostests/kmtests/ntos_ex/ExHardError.c
```diff
@@ -94,7 +115,12 @@ TestHardError(
     CheckHardError(0x40000003,                  0, OptionOk,                STATUS_SUCCESS,            ResponseNotHandled,     6, 1, 2, 3, 4, 5, 6);           // TODO: interactive on ROS
     }
 
-    CheckHardError(0x40000004,                  0, OptionShutdownSystem,    STATUS_PRIVILEGE_NOT_HELD, ResponseNotHandled,     0, 0);
+    /* Note: in the following test, getting a Response other than ResponseReturnToCaller
```
HBelusca, line 118: By the way, what do we get on ReactOS?

ThFabba: This is the code path that I fix in the ntoskrnl commit. So currently we return `ResponseNotHandled`, with this PR, we return `ResponseReturnToCaller`.

HBelusca: How could a returned value of `ResponseNotHandled` by NtRaiseHardError be a sign it could be exploited (as you suggest in your comment)?

ThFabba: In the buggy code path, the return value comes from an uninitialized variable (namely, that `SafeResponse` local var).
So ROS doesn't currently have a bug -- the reason I'm changing it in ROS code is just to make it behave the same as newer versions of Windows where the bug is fixed, rather than implementing its own fix (which was made earlier than Microsoft's, they just happened to choose a different default value)

HBelusca: OK then please refer to this commit https://github.com/reactos/reactos/commit/6cefd1242d0ab72ca3cb5e5f0dde0a3953f4b80e that fixed the thing in ROS originally.
Also note that there is one code path in ExRaiseHardError
```c
            if (!NT_SUCCESS(Status))
            {
                /* Return failure */
                *Response = ResponseNotHandled;
                return Status;
            }
```
where that "default" return value is also specified. I don't know whether Windows also use that one in such similar failure path, or not.

DarkFire01:
Going to run this against the testbots

WHS:
Test_KVM:

HBelusca:
Hi @ThFabba I'll handle this PR when we've dealt with PR #8342

tkreuzer:
These have been fixed by PR #8342

PR #5308, opened 2023-05-28
[KMTEST][NTOS][NDK] Fix SeQueryInfoToken test failures

ThFabba:
Review commit by commit please.

### DIFF
--- a/modules/rostests/kmtests/ntos_se/SeQueryInfoToken.c
+++ b/modules/rostests/kmtests/ntos_se/SeQueryInfoToken.c
@@ -328,6 +328,10 @@ START_TEST(SeQueryInfoToken)
     //      Testing SeFreePrivileges                                  //
     //----------------------------------------------------------------//
 
+    // FIXME: KernelMode will automatically get all access granted without
+    // getting Privileges filled in. For UserMode, Privileges will only get
+    // filled if either WRITE_OWNER or ACCESS_SYSTEM_SECURITY is requested
+    // and granted. So this doesn't really test SeFreePrivileges.
     Privileges = NULL;
     Checker = SeAccessCheck(
         AccessState->SecurityDescriptor,
@@ -342,7 +346,7 @@ START_TEST(SeQueryInfoToken)
         &Status
         );
     ok(Checker, "Checker is NULL\n");
-    ok((Privileges != NULL), "Privileges is NULL\n");
+    ok(Privileges == NULL, "Privileges is not NULL\n");
     if (Privileges)
     {
         trace("AuxData->PrivilegesUsed->PrivilegeCount = %d ; Privileges->PrivilegeCount = %d\n",
@@ -390,6 +394,8 @@ START_TEST(SeQueryInfoToken)
 
     // Call SeFreePrivileges again
 
+    // FIXME: See other SeAccessCheck call above, we're not really testing
+    // SeFreePrivileges here.
     Privileges = NULL;
     Checker = SeAccessCheck(
         AccessState->SecurityDescriptor,
@@ -404,7 +410,7 @@ START_TEST(SeQueryInfoToken)
         &Status
         );
     ok(Checker, "Checker is NULL\n");
-    ok((Privileges != NULL), "Privileges is NULL\n");
+    ok(Privileges == NULL, "Privileges is not NULL\n");
     if (Privileges)
     {
         trace("AuxData->PrivilegesUsed->PrivilegeCount = %d ; Privileges->PrivilegeCount = %d\n",
--- a/modules/rostests/kmtests/ntos_se/SeQueryInfoToken.c
+++ b/modules/rostests/kmtests/ntos_se/SeQueryInfoToken.c
@@ -229,6 +229,7 @@ START_TEST(SeQueryInfoToken)
     NTSTATUS Status = STATUS_SUCCESS;
     PAUX_ACCESS_DATA AuxData = NULL;
     PPRIVILEGE_SET NewPrivilegeSet;
+    ULONG InitialPrivilegeCount;
     BOOLEAN Checker;
     PPRIVILEGE_SET Privileges = NULL;
     PSECURITY_SUBJECT_CONTEXT SubjectContext = NULL;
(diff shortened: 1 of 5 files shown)


modules/rostests/kmtests/ntos_se/SeQueryInfoToken.c
```diff
@@ -240,14 +239,14 @@ START_TEST(SeQueryInfoToken)
...
+    // AUX_ACCESS_DATA gets larger in newer Windows version.
+    // This is the largest known size, found in Windows 10/11.
+    AuxData = ExAllocatePool(PagedPool, 0xe0);
 
     Status = SeCreateAccessState(AccessState,
-                                 (PVOID)AuxData,
+                                 AuxData,
                                  DesiredAccess,
-                                 GenericMapping
+                                 &(*PsProcessType)->TypeInfo.GenericMapping
```
tkreuzer, line 249: This will fail/crash on Windows Vista SP1 and later (see http://terminus.rewolf.pl/terminus/structures/ntdll/_OBJECT_TYPE_x86.html). Why not just put a copy of the GENERIC_MAPPING into this file and use it here?

HBelusca: Shouldn't also the kernel tests be rebuilt with suitable `NTDDI_VERSION` values and be run only on the windows version for which they have been compiled?

ntoskrnl/se/access.c
```diff
@@ -23,10 +23,10 @@
  * @param[in] Process
  * Valid process object where subject context is to be captured.
  *
- * @param[in,out] AccessState
+ * @param[out] AccessState
  * An initialized returned parameter to an access state.
  *
- * @param[in] AuxData
+ * @param[out] AuxData
```
GeoB99, line 29: `_Out_` implies that the routine is fully responsible to initialize and return the contents to the caller on that parameter, whereas the function actually expects a valid auxiliary data to be given to the access state as done by `AccessState->AuxData = AuxData;` and the used privileges being assigned to that auxiliary data. So I would argue using `_Inout_` is more fitting for this case? 🤷

ntoskrnl/se/access.c
```diff
@@ -101,10 +101,10 @@ SeCreateAccessStateEx(
  * @brief
  * Creates an access state.
  *
- * @param[in,out] AccessState
+ * @param[out] AccessState
  * An initialized returned parameter to an access state.
  *
- * @param[in] AuxData
+ * @param[out] AuxData
```
GeoB99, line 107: Ditto.

GeoB99 (commented):
The specific test code that exercises `SeFreePrivileges`  and `SeAccessCheck` has to be adjusted so that it can actually return a set of granted privileges to the untrusted caller. Of course this is something I should do this when I'm going to look more in-depth on this KMTEST in the future so I am fine with this change for time being.

ntoskrnl/se/access.c
HBelusca: Note: in terms of doxygen comments, this would translate into `@param[in,out]`

ntoskrnl/se/access.c
```diff
@@ -43,8 +43,8 @@ NTAPI
 SeCreateAccessStateEx(
     _In_ PETHREAD Thread,
     _In_ PEPROCESS Process,
-    _Inout_ PACCESS_STATE AccessState,
-    _In_ PAUX_ACCESS_DATA AuxData,
+    _Out_ PACCESS_STATE AccessState,
+    _Out_ __drv_aliasesMem PAUX_ACCESS_DATA AuxData,
```
HBelusca, line 47: What's the meaning of the `__drv_aliasesMem` annotation?

modules/rostests/kmtests/ntos_se/SeQueryInfoToken.c
```diff
@@ -240,14 +239,14 @@ START_TEST(SeQueryInfoToken)
     //----------------------------------------------------------------//
 
     AccessState = ExAllocatePool(PagedPool, sizeof(ACCESS_STATE));
-    PsProcessType = ExAllocatePool(PagedPool, sizeof(OBJECT_TYPE));
-    AuxData = ExAllocatePool(PagedPool, 0xC8);
-    GenericMapping = ExAllocatePool(PagedPool, sizeof(GENERIC_MAPPING));
+    // AUX_ACCESS_DATA gets larger in newer Windows version.
+    // This is the largest known size, found in Windows 10/11.
+    AuxData = ExAllocatePool(PagedPool, 0xe0);
```
HBelusca, line 244: Not using ExAllocatePoolZero or WithTag?

modules/rostests/kmtests/ntos_se/SeQueryInfoToken.c
ThFabba: The original design for kmtests had them running on Windows Server 2003 SP1 only and nothing else. But it's been shown that running them on other versions can be useful. There are a bunch of tests that make assumptions about internal data structures... but there's no reason this has to be one of them. So yeah, I'll go with Timo's fix.

ntoskrnl/se/access.c
ThFabba: I've thought about this a bit, but I think `_Out_` really only means "contents not valid before the call, contents valid after the call". And the function completely initializes the structure and doesn't expect any of it to be valid on input.
The `__drv_aliasesMem` expresses the "this pointer is copied to elsewhere", so I think the combination of the two expresses the right thing.

modules/rostests/kmtests/ntos_se/SeQueryInfoToken.c
ThFabba: Yeah, might as well since I'm already touching it.

ThFabba:
> The specific test code that exercises `SeFreePrivileges` and `SeAccessCheck` has to be adjusted so that it can actually return a set of granted privileges to the untrusted caller.

Yeah, I tried to get it to work as (probably) intended by switching to UserMode and adding WRITE_OWNER, but that just made the function fail instead, probably cause `SeTakeOwnershipPrivilege` isn't enabled. So I didn't bother digging more and just flipped the expectation.

tkreuzer:
✅ Test WHS: https://reactos.org/testman/compare.php?ids=87343,87345,87351,87359
✅ Test Windows 2003 x64: https://reactos.org/testman/compare.php?ids=87315,87342,87346,87360

JoachimHenze:
0.4.15-3-g5f1fcb5 vs. 0.4.15-4-g 3780e42ff997745241aca908afc8c6f8f9785a9c (2024-09-15):
VBox https://reactos.org/testman/compare.php?ids=97858,97887
KVM  https://reactos.org/testman/compare.php?ids=97859,97888

PR #5309, opened 2023-05-30
[FRAMEBUF] Return the "old" driver back

Extravert-ir:
It has a better license (GPL) and does not have obvious drawbacks compared to framebuf_new.


learn-more:
@hpoussin recently enabled this framebuf_new, did you contact him with regards to this change @Extravert-ir ?

binarymaster:
> hpoussin recently enabled this framebuf_new, did you contact him with regards to this change?

framebuf_new PR #4775 was merged by Katayama in 1bdefebc rather accidentally.

JoachimHenze:
Since I got no reply frome extravert34 what was better with the old driver:
In chat it was mentioned by other people that the new framebuf broke Win32K frankensteining.
It's not the only thing which broke that. IME commits may also have broken that additionally during 0.4.15-deving.

PR #5311, opened 2023-06-01
[GETUNAME] Update Romanian translation

AndreiMiloiu:


reactos-contributor:
_"29,271 additions, 8,927 deletions not shown because the diff is too large. Please use a local Git client to view these changes."_
Question: How could we comment on the PR's content if we cannot view it here?

AndreiMiloiu:
I don't know. The file was already very large. I can create a new file and add a part of my translation, we can comment it, then I can add another part of the file in it and so on. Or you can download the two versions of the file: the original one and the modified one. I know there are some plugins in Notepad++ that help you to compare two files.

AndreiMiloiu:
I didn't find an easy way to solve this problem in this PR, so I made a new one named [GETUNAME] Update Romanian translation (2nd attempt). Please, look at it. I'm going to close this.

PR #5314, opened 2023-06-02
[NTOS:SE] Refactor NtOpenThreadTokenEx + fix a pool corruption in SepDuplicateToken

GeoB99:
Check the commit descriptions for explanation.

[CORE-18986](https://jira.reactos.org/browse/CORE-18986)

### DIFF
--- a/ntoskrnl/se/token.c
+++ b/ntoskrnl/se/token.c
@@ -1103,6 +1103,252 @@ SepFindPrimaryGroupAndDefaultOwner(
 
 /**
  * @brief
+ * Internal private function that returns an opened handle
+ * of an access token associated with a thread.
+ *
+ * @param[in] Thread
+ * A pointer to a Executive thread. This parameter is used to
+ * validate that the newly obtained thread in this function
+ * hasn't diverged. This could potentially lead to a scenario
+ * that we might get an access token from a different token
+ * which is not what we want. The validation is performed
+ * if the token has to copied and can't be opened directly.
+ *
+ * @param[in] ThreadHandle
+ * A handle to a thread, of which an access token is to be opened
+ * and given from that thread.
+ *
+ * @param[in] ThreadToken
+ * A pointer to an access token associated with the specific thread.
+ * The function assumes that the token is an impersonation one
+ * prior the calling of this function.
+ *
+ * @param[in] DesiredAccess
+ * The desired access rights for the access token.
+ *
+ * @param[in] HandleAttributes
+ * Handle attributes of which they are used for the newly creation
+ * of the opened thread token. The function assumes that they have
+ * been validated prior the calling of this function.
+ *
+ * @param[in] EffectiveOnly
+ * If set to TRUE, the function will copy a new access token with
+ * privileges and groups that are effectively enabled. Any disabled
+ * privilege or group is removed from the copied token. Otherwise
+ * if set to FALSE, the function retains all the enabled and disabled
+ * privielges and groups.
+ *
+ * @param[in] CopyOnOpen
+ * If set to TRUE, it tells the function that the access token cannot
+ * be directly opened due to the security impersonation info of the
+ * associated thread being enforced. In this case the function will
+ * make a copy of the said token by duplicating it. Otherwise if set
+ * to FALSE, the function will just open the access token directly.
+ *
+ * @param[in] ImpersonationLevel
+ * The security impersonation level, at which it is allowed to
+ * access the token.
+ *
+ * @param[in] PreviousMode
+ * The processor request level mode.
+ *
+ * @param[out] OpenedTokenHandle
+ * A pointer to an opened access token handle associated with the
+ * specific thread, returned to the caller. Initially this parameter
+ * is set to NULL and if the function fails to open the thread's token,
+ * it will stay NULL.
+ *
+ * @return
+ * Returns STATUS_SUCCESS if the function has successfully opened the thread's
+ * token. STATUS_OBJECT_TYPE_MISMATCH is returned if the obtained thread object
+ * no longer matches with the other thread that has been obtained previously.
+ * STATUS_NO_TOKEN is returned if the associated thread's process has no
+ * primary access token. A failure NTSTATUS code is returned otherwise.
+ */
+static
+NTSTATUS
+SepOpenThreadToken(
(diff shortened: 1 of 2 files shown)


ntoskrnl/se/token.c
```diff
@@ -2119,138 +2374,82 @@ NtOpenThreadTokenEx(
...
-        }
+        DPRINT1("Failed to open the thread's token (Status 0x%lx)\n", Status);
+        ObDereferenceObject(Token);
+        ObDereferenceObject(Thread);
+        return Status;
     }
 
-    if (NewToken) ObDereferenceObject(NewToken);
-
+    /* Dereference all the stuff and give the token handle to caller */
```
HBelusca, line 2440: You could move both those ObDereferenceObject calls before the `if (!NT_SUCCESS(Status))` above: this would allow to not duplicate them.

ntoskrnl/se/token.c
```diff
@@ -1101,6 +1101,264 @@ SepFindPrimaryGroupAndDefaultOwner(
     return STATUS_SUCCESS;
 }
 
+/**
+ * @brief
+ * Internal private function that returns an opened handle
+ * of an access token associated with a thread.
+ *
+ * @param[in] Thread
+ * A pointer to a Executive thread (ETHREAD). This parameter
```
HBelusca, line 1110:
```suggestion
 * A pointer to a Executive thread. This parameter
```
Kernel devs know what it is (and its type, it's just written below in the function parameters list).

ntoskrnl/se/token.c
```diff
@@ -1101,6 +1101,264 @@ SepFindPrimaryGroupAndDefaultOwner(
...
+        *OpenedTokenHandle = TokenHandle;
+        return STATUS_SUCCESS;
+    }
+
+    /*
+     * The caller asks to do a copy of that token whilst it's opened.
+     * Obtain a thread object again but this time we have to obtain
+     * it in our side, kernel mode.
+     */
+    Status = ObReferenceObjectByHandle(ThreadHandle,
```
HBelusca, line 1222: Do we need to do that, because the original Thread object psased also as parameter doesn't have all the required rights in order to do the token copy?
(This aspect of explanation is missing in your comment above.)

ntoskrnl/se/token.c
```diff
@@ -1101,6 +1101,264 @@ SepFindPrimaryGroupAndDefaultOwner(
...
+    ObFastDereferenceObject(&Thread2->ThreadsProcess->Token, PrimaryToken);
+    if (!NT_SUCCESS(Status))
+    {
+        DPRINT1("Failed to create an impersonation token DACL (Status 0x%lx)\n", Status);
+        ObDereferenceObject(Thread2);
+        return Status;
+    }
+
+    /* Create a security descriptor with the DACL we got */
+    if (Dacl)
```
HBelusca, line 1262: Weird, so is it possible that the `SepCreateImpersonationTokenDacl` call above could has succeeded (thus the previous `if (!NT_SUCCESS(Status))` check was FALSE and not taken), **_even if_** the returned `Dacl` pointer is still NULL ? This looks wrong.

ntoskrnl/se/tokenlif.c
```diff
@@ -665,18 +665,60 @@ SepDuplicateToken(
...
      * token as effective only.
      */
     if (EffectiveOnly)
     {
-        /* Begin querying the groups and search for disabled ones */
-        for (GroupsIndex = 0; GroupsIndex < AccessToken->UserAndGroupCount; GroupsIndex++)
+        /*
+         * Begin querying the groups and search for disabled ones. Do not touch the
+         * user which is at the first position because it cannot be disabled, no
+         * matter what attributes does it have.
```
HBelusca, line 716:
```suggestion
         * matter what attributes it does have.
```
(or "it has")

HBelusca (changes requested):
Some stuff to check...

GeoB99:
Test KVM: https://reactos.org/testman/compare.php?ids=87396,87399
Test Vbox: https://reactos.org/testman/compare.php?ids=87397,87400

ntoskrnl/se/token.c
GeoB99: It's a copy from the previous code from `NtOpenThreadTokenEx`, `SepCreateImpersonationTokenDacl` will always return a DACL for you. I'm gonna change that.

ntoskrnl/se/token.c
GeoB99: Yes we do, because the original thread only has the query access needed only to query information for token validation. On a CoO (Copy-On-Open) scenario we would need to grab the thread again with full access needed to duplicate a token. I'm gonna expand the comment on that.

ntoskrnl/se/token.c
```diff
@@ -1101,6 +1101,252 @@ SepFindPrimaryGroupAndDefaultOwner(
     return STATUS_SUCCESS;
 }
 
+/**
+ * @brief
+ * Internal private function that returns an opened handle
+ * of an access token associated with a thread.
+ *
+ * @param[in] Thread
+ * A pointer to a Executive thread. This parameter is used
```
JoachimHenze, line 1110:
```suggestion
 * A pointer to a Executive thread. This parameter is used to
```

ntoskrnl/se/token.c
```diff
@@ -1101,6 +1101,252 @@ SepFindPrimaryGroupAndDefaultOwner(
...
+                                       KernelMode,
+                                       (PVOID*)&Thread2,
+                                       NULL);
+    if (!NT_SUCCESS(Status))
+    {
+        DPRINT1("Failed to reference the object thread (Status 0x%lx)\n", Status);
+        return Status;
+    }
+
+    /* Check that one of the threads haven't diverged */
```
JoachimHenze, line 1236:
```suggestion
    /* Check that one of the threads hasn't diverged */
```

JoachimHenze (approved):
My acceptance is based on the reported positive result of the tests, not on any deep understanding of the code.
Sidenote: Commenting is pretty verbose, and focuses a bit too much on the what instead of the why.

HBelusca (commented):
ok for the merging, my "stuff" to check was addressed in between.

PR #5315, opened 2023-06-02
[GITHUB] Temporarily disable MSBuild

GeoB99:
Temporarily disable MSBuild until a proper fix is shipped, having a single failing worker just because MS did an oopsie with MSBuild is annoying.

[CORE-18911](https://jira.reactos.org/browse/CORE-18911)


HBelusca (approved):
approved... oopsie :rofl:

Extravert-ir:
I think we can downgrade the builder version to "windows-2019"

Extravert-ir (approved):
This is caused by a change in CMake btw https://gitlab.kitware.com/cmake/cmake/-/issues/24639

PR #5317, opened 2023-06-03
[OBJ2BIN] Support MSVC-specific text sections

disean:
## Purpose

An ASM file can turn into an empty binary file in certain case (MSVC, 32-bit code)


HBelusca:
Can you tell us how you are able to get those `.text$mn` sections sometimes?

disean:
See [this commit](https://github.com/disean/reactos/commit/c998a5ca12c401b3e814239e7f898cb5f2d89f93#diff-b23e889e54a8a629075d20cf1160e4a77ca0237add7b9af10e73c74a21d8c556R13) for the details.

HBelusca:
> See [this commit](https://github.com/disean/reactos/commit/c998a5ca12c401b3e814239e7f898cb5f2d89f93#diff-b23e889e54a8a629075d20cf1160e4a77ca0237add7b9af10e73c74a21d8c556R13) for the details.

Yes well it seems to give an example of when this happens, but ... why?
The new startrom.S doesn't seem, at first sight, to differ much from other .S files compiled for freeldr...

sdk/include/asm/asm.inc
```diff
@@ -98,7 +98,7 @@ ENDM
 ENDM
 
 .code32 MACRO
-    .code
+    .text SEGMENT use32 PUBLIC 'CODE'
```
HBelusca, line 101: That's interesting. Why the usual `.code` section does not work is a mystery :thinking:

HBelusca:
@disean Just to clarify regarding this startrom.S, the OLPC platform calls this entry-point code directly in 32-bit mode?

sdk/include/asm/asm.inc
HBelusca: interesting...

disean:
**startrom.S**
pr
**arcname.c**
pr2
**command.com**
pr3

binarymaster (approved):

HBelusca:
Interesting... @disean what's the name of this tool you use in your screenshots?

HBelusca:
Before approving the PR I will do extra tests to understand the _root cause_ of why MSVC suddenly generate those .text$mn sections. In your case it seems to be related to the fact you compile/assemble *32-bit* C code that is not yet linked, and also possibly because of its base address == 0.

Does not happen with our standard 16-bit stubs? It would be of interest if you could also show how our MSVC/ML-assembled bootsectors look like. What's interesting is that the MSVC command.com (still in object form) has _*both*_ sections, .text and .text$mn. The latter being empty, our operation of extracting only .text didn't break anything indeed.

disean:
16-bit object files (16-bit stubs, bootsectors) all have similar sections pattern like `command.com`: the .text and empty .text$mn.

> what's the name of this tool you use in your screenshots?

7-Zip :)

learn-more (approved):

learn-more:
Started a test build:
https://build.reactos.org/#/builders/6/builds/37895

cbialorucki (approved):

sdk/tools/obj2bin/obj2bin.c
```diff
@@ -140,8 +140,9 @@ int main(int argc, char *argv[])
     /* Loop all sections */
     for (i = 0; i < pFileHeader->NumberOfSections; i++)
     {
-        /* Check if this is '.text' section */
-        if ((strcmp((char*)pSectionHeader->Name, ".text") == 0) &&
+        /* Check if this is a section with the executable code */
```
HBelusca:
```suggestion
        /* Check if this is a section with executable code */
```

PR #5321, opened 2023-06-04
[NTOS:PNP] Assign correct PnP IDs to the devices detected by bootloader

HBelusca:
_**Made on behalf of @binarymaster**_ 

Based on a commit by Vadim Galyant:
https://github.com/vgalnt/reactos/commit/5ef5c11e7fcca44ab31606ab762e3700f1ae24f4

JIRA issues: [CORE-18963](https://jira.reactos.org/browse/CORE-18963), [CORE-17977](https://jira.reactos.org/browse/CORE-17977)


HBelusca:
cc @disean too :)

ntoskrnl/io/pnpmgr/pnpmap.c
```diff
@@ -29,48 +35,114 @@ static UNICODE_STRING ConfigurationDataU = RTL_CONSTANT_STRING(L"Configuration D
...
     {
-        RtlInitUnicodeString(&CmpId, PnPMap[i].DetectId);
+        CurId = PnPMap[i].DetectId;
+        RtlInitUnicodeString(&CmpId, CurId);
 
         if (RtlCompareUnicodeString(DetectId, &CmpId, FALSE) == 0)
         {
             *DeviceIndex = PnPMap[i].Counter++;
+
+            if (wcscmp(CurId, L"KeyboardPeripheral") == 0)
```
ThFabba, line 142: Why not
1. compare to a UNICODE_STRING, since you already have one in `CmpId` anyway, or
2. maybe even avoid this string comparison entirely by making the `PnPMap` predictable so that this can be an index check, or
3. maybe the most elegant: just put the KeyboardMap/PointerMap pointer inside the `PnPMap` array as a struct member?

(Side note, Windows doesn't seem to need a loop here at all since they just use the `CONFIGURATION_TYPE` returned by `IoQueryDeviceDescription`? Not super familiar with this code, so not sure if that idea is even applicable to ROS.)

ntoskrnl/io/pnpmgr/pnpmap.c
```diff
@@ -29,48 +35,114 @@ static UNICODE_STRING ConfigurationDataU = RTL_CONSTANT_STRING(L"Configuration D
...
+    { NULL, NULL }
+};
+
 /* FUNCTIONS *****************************************************************/
 
+static
+CODE_SEG("INIT")
+PWSTR
+IopMapPeripheralId(
+    _In_ PUNICODE_STRING Value,
```
ThFabba, line 98: `PCUNICODE_STRING` should work here

ntoskrnl/io/pnpmgr/pnpmap.c
HBelusca: IoQueryDeviceDescription only takes its parameters as *inputs*: the pointer to `CONFIGURATION_TYPE` given to it is an input. (It's done that way, instead of giving the direct value, in order to tell the function what it should enumerate and what it shouldn't.)

ntoskrnl/io/pnpmgr/pnpmap.c
```diff
@@ -29,33 +36,89 @@ static UNICODE_STRING ConfigurationDataU = RTL_CONSTANT_STRING(L"Configuration D
 static UNICODE_STRING BootConfigU = RTL_CONSTANT_STRING(L"BootConfig");
 static UNICODE_STRING LogConfU = RTL_CONSTANT_STRING(L"LogConf");
 
-/* FIXME: There should be two sets of hardcoded PnP identifiers
- * for the keyboard and pointer peripherals (see CORE-18963).
- * They also can be parsed from a LegacyXlate sections of driver INF files.
- */
+/* FIXME: Trailing \0 in structures below are hacks, should be removed.
```
disean, line 39: An identifier string will always have a length of 7

ntoskrnl/io/pnpmgr/pnpmap.c
```diff
@@ -29,33 +36,89 @@ static UNICODE_STRING ConfigurationDataU = RTL_CONSTANT_STRING(L"Configuration D
...
+ * Hardware identifiers also can be mapped using "LegacyXlate" sections
+ * of driver INF files. */
+
```
disean, line 42:
```suggestion
DATA_SEG("INITDATA")
```

ntoskrnl/io/pnpmgr/pnpmap.c
HBelusca: > 19:46 Stas'M:
This is needed for the underlying code which works with multi-line registry strings

binarymaster (approved):
Looks good to me! 🏅 image

Tests passing. 👍

HBelusca (commented):
lgtm

PR #5324, opened 2023-06-06
[NTOSKRNL] Fix a double reference on client impersonation and refactor SeTokenCanImpersonate

GeoB99:
Back then I was working on #3605 that would determine if the server is allowed to impersonate a client's token or not, although after learning more about how impersonation works there are several issues concerning with `SeTokenCanImpersonate`, that is...

- `ImpersonationLevel` is basically what the server asks to impersonate a client based on that level, having there failing if the level was either given `SecurityAnonymous` or `SecurityIdentification` doesn't make much sense. If the server wants to identify the details of a client then let it do so.
- The function does not check for `SeImpersonatePrivilege`. Certain system components of the OS or the system process itself might need to impersonate on behalf of a client to do specific OS tasks, so give them leverage.
- The function does not check if the server is authenticated on the same logon session of the client, nor it checks if it requests an impersonation that is beyond the acceptable level that the client permits.
- The code path that looks for the client's token having an anonymous authentication doesn't make much sense, specifically where it fails impersonation if an anonymous token wasn't created by the system. On contrary, anonymous tokens are only created by the system and as their name imply, they have hidden security attributes with no privileges attached to them so one could freely impersonate such token.

In addition to that there are two bugs in `PsImpersonateClient`, that is, we're trying to reference a token after we filled out thread impersonation info. However we aren't taking into account that if impersonation is not possible we are making a copy of the client token and `SeCopyClientToken` already holds a reference which means we are gaining an extra reference count. Therefore if the server was to cancel impersonation the previously copied token still has an extra reference and the token won't go away, WHICH IS WRONG! 

Also, if impersonation fails we are assigning the impersonation level to the thread that is given by the caller, instead of assigning it as `SecurityIdentification`. This could lead to erratic behaviors or worse, as the token doesn't allow for impersonation yet the thread thinks it can actually impersonate as per its impersonation info.

Some interesting stuff:
https://www.youtube.com/watch?v=QRpfvmMbDMg
https://learn.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/impersonate-a-client-after-authentication
https://juggernaut-sec.com/seimpersonateprivilege/

### DIFF
--- a/ntoskrnl/ps/security.c
+++ b/ntoskrnl/ps/security.c
@@ -615,6 +615,7 @@ PsImpersonateClient(IN PETHREAD Thread,
 {
     PPS_IMPERSONATION_INFORMATION Impersonation, OldData;
     PTOKEN OldToken = NULL, ProcessToken = NULL;
+    BOOLEAN CopiedToken = FALSE;
     PACCESS_TOKEN NewToken, ImpersonationToken;
     PEJOB Job;
     NTSTATUS Status;
@@ -706,7 +707,13 @@ PsImpersonateClient(IN PETHREAD Thread,
                 return Status;
             }
 
-            /* Since we cannot impersonate, assign the newly copied token */
+            /*
+             * Since we cannot impersonate, assign the newly copied token.
+             * SeCopyClientToken already holds a reference to the copied token,
+             * let the code path below know that it must not reference it twice.
+             */
+            CopiedToken = TRUE;
+            ImpersonationLevel = SecurityIdentification;
             ImpersonationToken = NewToken;
         }
 
@@ -721,6 +728,11 @@ PsImpersonateClient(IN PETHREAD Thread,
             if ((Job->SecurityLimitFlags & JOB_OBJECT_SECURITY_NO_ADMIN) &&
                 SeTokenIsAdmin(ImpersonationToken))
             {
+                if (CopiedToken)
+                {
+                    ObDereferenceObject(ImpersonationToken);
+                }
+
                 return STATUS_ACCESS_DENIED;
             }
 
@@ -728,6 +740,11 @@ PsImpersonateClient(IN PETHREAD Thread,
             if ((Job->SecurityLimitFlags & JOB_OBJECT_SECURITY_RESTRICTED_TOKEN) &&
                 SeTokenIsRestricted(ImpersonationToken))
             {
+                if (CopiedToken)
+                {
+                    ObDereferenceObject(ImpersonationToken);
+                }
+
                 return STATUS_ACCESS_DENIED;
             }
 
@@ -758,7 +775,12 @@ PsImpersonateClient(IN PETHREAD Thread,
         Impersonation->CopyOnOpen = CopyOnOpen;
         Impersonation->EffectiveOnly = EffectiveOnly;
         Impersonation->Token = ImpersonationToken;
-        ObReferenceObject(ImpersonationToken);
+
+        /* Do not reference the token again if we copied it */
+        if (!CopiedToken)
+        {
+            ObReferenceObject(ImpersonationToken);
+        }
 
         /* Unlock the thread */
         PspUnlockThreadSecurityExclusive(Thread);
--- a/ntoskrnl/se/token.c
+++ b/ntoskrnl/se/token.c
@@ -3,7 +3,7 @@
  * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
  * PURPOSE:     Security access token implementation base support routines
  * COPYRIGHT:   Copyright David Welch <welch@cwcom.net>
- *              Copyright 2021-2022 George Bișoc <george.bisoc@reactos.org>
+ *              Copyright 2021-2023 George Bișoc <george.bisoc@reactos.org>
  */
 
 /* INCLUDES *******************************************************************/
@@ -1752,16 +1752,13 @@ PTOKEN
 NTAPI
 SepCreateSystemProcessToken(VOID)
 {
-    LUID_AND_ATTRIBUTES Privileges[25];
     ULONG GroupAttributes, OwnerAttributes;
-    SID_AND_ATTRIBUTES Groups[32];
     LARGE_INTEGER Expiration;
     SID_AND_ATTRIBUTES UserSid;
     ULONG GroupsLength;
     PSID PrimaryGroup;
     OBJECT_ATTRIBUTES ObjectAttributes;
     PSID Owner;
-    ULONG i;
     PTOKEN Token;
     NTSTATUS Status;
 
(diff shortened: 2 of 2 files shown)


ntoskrnl/ps/security.c
```diff
@@ -689,25 +691,38 @@ PsImpersonateClient(IN PETHREAD Thread,
...
 
-        /* Make sure we can impersonate */
-        if (!SeTokenCanImpersonate(ProcessToken,
-                                   Token,
-                                   ImpersonationLevel))
+        /*
+         * The server may want to obtain identification details of a client
+         * instead of impersonating so just give the server a pass.
+         */
+        if (Level > SecurityIdentification)
```
HBelusca, line 698: Shouldn't this inequality check better be done by `SeTokenCanImpersonate` ? (It takes `Level` as parameter, and original code already tried to do validation checks on `Level`.)

ntoskrnl/se/token.c
```diff
@@ -2220,78 +2244,83 @@ SeTokenCanImpersonate(
...
-
     /* Time to lock our tokens */
     SepAcquireTokenLockShared(ProcessToken);
     SepAcquireTokenLockShared(TokenToImpersonate);
 
-    /* What kind of authentication ID does the token have? */
-    if (RtlEqualLuid(&TokenToImpersonate->AuthenticationId,
-                     &SeAnonymousAuthenticationId))
+    /* Allow impersonation for the process server if it's granted the impersonation privilege */
+    if ((ProcessToken->TokenFlags & TOKEN_HAS_IMPERSONATE_PRIVILEGE) != 0)
```
HBelusca, line 2254: Are you sure this is the correct ordering of checks (verifying this flag, before doing the equalluid for anonymous auth on the TokenToImpersonate)?

ntoskrnl/se/token.c
```diff
@@ -2220,78 +2244,83 @@ SeTokenCanImpersonate(
...
     /*
-     * Make sure the tokens aren't diverged in terms of
-     * restrictions, that is, one token is restricted
-     * but the other one isn't.
+     * Deny impersonation for the server if it wants to impersonate a client
+     * beyond what the impersonation level originally permits.
      */
-    if (SeTokenIsRestricted(ProcessToken) !=
-        SeTokenIsRestricted(TokenToImpersonate))
+    if (ImpersonationLevel > TokenToImpersonate->ImpersonationLevel)
```
HBelusca, line 2280: Is this what Windows does?

ntoskrnl/se/token.c
```diff
@@ -1856,7 +1856,10 @@ SepCreateSystemProcessToken(VOID)
 
     Privileges[i].Attributes = 0;
     Privileges[i++].Luid = SeSystemtimePrivilege;
-    ASSERT(i == 20);
+
```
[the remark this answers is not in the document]
GeoB99: I might as well add these while I'm at it.

ntoskrnl/se/token.c
GeoB99: No, nor does it mean that it is automatically disqualified because it is not what Windows does. Having the check there makes sense because you want a server that respects the impersonation limitations imposed by a client, e.g. if the client forbids anyone to delegate its security credentials through remote impersonation but only impersonate the client, you can't simply ask to impersonate other than what you've been limited to.

ntoskrnl/se/token.c
```diff
@@ -1875,7 +1878,7 @@ SepCreateSystemProcessToken(VOID)
                             3,
                             Groups,
```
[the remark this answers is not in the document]
GeoB99: All the free space in the array is a waste indeed, since I'm touching this I might as well address that.

GeoB99:
> Does our testsuite have tests that exercise those functions?

Currently our testsuite doesn't cover tests for `PsImpersonateClient`, a testcase for it is planned in the future. The aforementioned bugs are pretty obvious anyway.

HBelusca:
Some related interesting readings:
https://stackoverflow.com/a/45624473/21852502 (the links mentioned should be safe)
https://malware.news/t/dont-use-system-tokens-for-sandboxing-part-1-of-n/36625
Slides 31-32+ in https://i.blackhat.com/USA-22/Wednesday/US-22-Forshaw-Taking-Kerberos-To-The-Next-Level.pdf
Slides 24-26 (and the rest) in https://i.blackhat.com/USA-22/Wednesday/US-22-Forshaw-Taking-Kerberos-To-The-Next-Level.pdf
https://www.exploit-db.com/exploits/44630
These may be useful to check whether we would be vulnerable to those types of problems mentioned there, as well as things regarding PsImpersonateClient and SeTokenCanImpersonate.

ntoskrnl/se/token.c
```diff
@@ -1856,7 +1856,10 @@ SepCreateSystemProcessToken(VOID)
 
     Privileges[i].Attributes = 0;
     Privileges[i++].Luid = SeSystemtimePrivilege;
-    ASSERT(i == 20);
+
+    Privileges[i].Attributes = SE_PRIVILEGE_ENABLED_BY_DEFAULT | SE_PRIVILEGE_ENABLED;
+    Privileges[i++].Luid = SeImpersonatePrivilege;
+    ASSERT(i == 21);
```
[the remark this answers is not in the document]
GeoB99: Adding the missing privileges that you mentioned into the array will total the count up to 24 privileges, so 25 is reasonable.

GeoB99:
I can only speak up for `PsImpersonateClient` and its function `SeTokenCanImpersonate`, they do not pose a vulnerability on the system, at least as far as I'm aware. The checks as they currently stand are reasonable, we can add more security checks if needed be.

The vulnerability is not on the privileges themselves or the implemented checks but on the current logged user. You only need a very slight of power in order to circumvent the mechanisms and take over the whole system. Microsoft made a big mistake in the long past where the newly created user upon installation was granted admin rights and as a result many application developers followed suit. Microsoft later realized the mistake but damage was already done.

We as a project have the bonus that we can take immediate action if a vulnerability plagues us. Our weapon at disposal is **filtered tokens** and such, so that you can strip off the user from rights and privileged groups. At the cost of breaking compatibility with many Win32 apps of course. We could implement a **"trusted app database"** where only trusted programs can be executed normally, whereas any other unknown program is prohibited from executing or given very limited to no rights at all.

ntoskrnl/se/token.c
Extravert-ir, line 1862: Oh, this is just a statically initialized array. Please, do it the proper way:
```
LUID_AND_ATTRIBUTES Privileges[] = {
{
    .Luid = SeTcbPrivilege,
    .Attributes = SE_PRIVILEGE_ENABLED_BY_DEFAULT | SE_PRIVILEGE_ENABLED
},
{
    .Luid = SeCreateTokenPrivilege,
},
{
    .Luid = SeTakeOwnershipPrivilege,
},
...
};
```
and then when you need to take the amount of it, just use `RTL_NUMBER_OF(Privileges)`

Extravert-ir (changes requested):

ntoskrnl/se/token.c
JoachimHenze: Please do not artificially worsen Georges code by giving recommendations extravert34 that will break the compatibility of essential ros components like the kernel to older compilers with your C99-isms. E.g. VS2010 will not accept that way of initializing the struct, which you suggested. Neither in a .cpp file nor in a .c file. Your suggestion is very unportable code with poor compiler compatibility.

That RTL_NUMBER_OF(Priviliges) part is unaffected by my comment, therefore I did not resolve this conversation yet.

ntoskrnl/se/token.c
```diff
@@ -1875,7 +1884,7 @@ SepCreateSystemProcessToken(VOID)
                             3,
```
HBelusca, line 1884:
```suggestion
                            RTL_NUMBER_OF(Groups),
```

ntoskrnl/se/token.c
```diff
@@ -1754,14 +1754,13 @@ SepCreateSystemProcessToken(VOID)
 {
     LUID_AND_ATTRIBUTES Privileges[25];
```
HBelusca, line 1755:
```suggestion
    LUID_AND_ATTRIBUTES Privileges[24];
```

ntoskrnl/se/token.c
```diff
@@ -2193,22 +2202,48 @@ SeTokenIsWriteRestricted(
...
  * @return
  * Returns TRUE if the conditions checked are met for token impersonation,
  * FALSE otherwise.
+ *
+ * @remarks
+ * The server has to meet the following criteria in order to impersonate
+ * a client, that is:
+ *
+ * - The server must not impersonate a client beyond the level that
+ *   the client imposed on itself
```
HBelusca, line 2229:
```suggestion
 *   the client imposed on itself.
```

ntoskrnl/se/token.c
```diff
@@ -2193,22 +2202,48 @@ SeTokenIsWriteRestricted(
...
+ *
+ * - The server must be authenticated on the same logon session of
+ *   the target client
```
HBelusca, line 2232:
```suggestion
 *   the target client.
```

ntoskrnl/se/token.c
```diff
@@ -2193,22 +2202,48 @@ SeTokenIsWriteRestricted(
...
+ *
+ * - IF NOT then the server's user has to match to that of the
+ *   target client
```
HBelusca, line 2235:
```suggestion
 *   target client.
```

ntoskrnl/se/token.c
```diff
@@ -2193,22 +2202,48 @@ SeTokenIsWriteRestricted(
...
+ * The server has to meet the following criteria in order to impersonate
+ * a client, that is:
+ *
+ * - The server must not impersonate a client beyond the level that
+ *   the client imposed on itself
+ *
+ * - The server must be authenticated on the same logon session of
+ *   the target client
+ *
+ * - IF NOT then the server's user has to match to that of the
```
HBelusca, line 2234: user ID ?

ntoskrnl/se/token.c
```diff
@@ -2193,22 +2202,48 @@ SeTokenIsWriteRestricted(
...
+ *   target client
+ *
+ * - The server must not be restricted in order to impersonate a
+ *   client that's not restricted
```
HBelusca, line 2238:
```suggestion
 *   client that is not restricted
```

ntoskrnl/se/token.c
```diff
@@ -2221,73 +2256,87 @@ SeTokenCanImpersonate(
...
+        if (!RtlEqualSid(ProcessToken->UserAndGroups[0].Sid,
+                         TokenToImpersonate->UserAndGroups[0].Sid))
+        {
+            DPRINT1("Server and client aren't the same user!\n");
+            CanImpersonate = FALSE;
+            goto Quit;
+        }
+
+        /*
+         * Make sure the tokens aren't diverged in terms of restrictions
```
HBelusca, line 2318:
```suggestion
         * Make sure the tokens haven't diverged in terms of restrictions
```

HBelusca (commented):
lgtm

ntoskrnl/se/token.c
Extravert-ir: @JoachimHenze would you stop repeating that bullshit about MSVC2010 and "not portable code" and finally make your own project with VC6 and Turbo C++ support?

Why should everyone satisfy you just because of your stupid arguments?
Instead of writing clean and correct code, you're trying to sabotage the project for the sake of "compatibility".

HBelusca (approved):

PR #5325, opened 2023-06-09
[WINMINE] Rename WineMine to Minesweeper

Artem13327:
I also made a translation somewhere right away, as I was able to take it from the shortcuts.inf file and check it for correctness. I will also have to ask for help with line 327 in the shortcuts.inf file from those who know Czech, and with line 844 from those who know Indonesian


ThFabba:
Does this fix something? If not, I'd rather leave it as WineMine, because that's what it is...

Matriks404:
I agree with @Artem13327 name change, since it makes it more clear what game is about, although it definitely doesn't make that much of difference.

GeoB99 (changes requested):
I'm definitely against this change because WineMine isn't ours (it is literally in the name) and I don't think the license explicitly allows it. Well, even though if it may allow it I don't think we're giving good impressions to Wine with this move.

boot/bootdata/CMakeLists.txt
```diff
@@ -103,7 +103,7 @@ add_livecd_shortcut("Notepad" notepad.exe "Profiles/All Users/Start Menu/Program
 add_livecd_shortcut("WordPad" wordpad.exe "Profiles/All Users/Start Menu/Programs/Accessories")
 add_livecd_shortcut("Solitaire" sol.exe "Profiles/All Users/Start Menu/Programs/Games")
 add_livecd_shortcut("Spider Solitaire" spider.exe "Profiles/All Users/Start Menu/Programs/Games")
-add_livecd_shortcut("WineMine" winmine.exe "Profiles/All Users/Start Menu/Programs/Games")
+add_livecd_shortcut("Minesweeper" winmine.exe "Profiles/All Users/Start Menu/Programs/Games")
```
binarymaster, line 106: If you limit the changes to only renaming the shortcuts, I think this could be acceptable.

MesaBlack:
https://www.youtube.com/watch?v=NIgfiSzCy1o

PR #5326, opened 2023-06-10
[NTOS:PNP] Refactor remaining functions after PnP manager rewrite. CORE-18961

olase23:
## Purpose

Not all IRP sending functions are refactored after PnP manager rewrite
namely, IopQueryDeviceCapabilities, IopQueryHardwareIds, IopQueryCompatibleIds
They should have a function in pnpirp.c file that sends an IRP, and other related code in devaction.c

JIRA issue: [CORE-18961](https://jira.reactos.org/browse/CORE-18961)

## Proposed changes

- Intruducing two new IRP functions to query HardwareIds, CompatibleIds and device node capabilities.
- Separate device_node capabilities query and set functionality. 
- Centralize storing device_node capabilities in the PiInitializeDevNode() when the device_node enumeration occurs.
- Also adapt function names to NT5.2-like schema.

### DIFF
--- a/ntoskrnl/include/internal/io.h
+++ b/ntoskrnl/include/internal/io.h
@@ -1436,7 +1436,7 @@ NTSTATUS
 PiIrpQueryPnPDeviceId(
     _In_ PDEVICE_NODE DeviceNode,
     _In_ BUS_QUERY_ID_TYPE IdType,
-    _Out_ PWCHAR *Id);
+    _Out_ PZZWSTR *Id);
 
 //
 // Global I/O Data
--- a/ntoskrnl/include/internal/io.h
+++ b/ntoskrnl/include/internal/io.h
@@ -629,8 +629,9 @@ IopFreeDeviceNode(
 
 NTSTATUS
 NTAPI
-IopQueryDeviceCapabilities(PDEVICE_NODE DeviceNode,
-                           PDEVICE_CAPABILITIES DeviceCaps);
+PiSetDeviceCapabilities(
+    _In_ PDEVICE_NODE DeviceNode,
+    _Out_ PDEVICE_CAPABILITIES DeviceCaps);
 
 NTSTATUS
 IopSynchronousCall(
@@ -1425,6 +1426,11 @@ NTSTATUS
 PiIrpQueryPnPDeviceState(
     _In_ PDEVICE_NODE DeviceNode,
     _Out_ PPNP_DEVICE_STATE DeviceState);
+
+NTSTATUS
+PiIrpQueryPnPDeviceCapabilities(
+    _In_ PDEVICE_NODE DeviceNode,
+    _Out_ PDEVICE_CAPABILITIES DeviceCaps);
 
 //
 // Global I/O Data
--- a/ntoskrnl/include/internal/io.h
+++ b/ntoskrnl/include/internal/io.h
@@ -1432,6 +1432,12 @@ PiIrpQueryPnPDeviceCapabilities(
     _In_ PDEVICE_NODE DeviceNode,
     _Out_ PDEVICE_CAPABILITIES DeviceCaps);
 
+NTSTATUS
+PiIrpQueryPnPDeviceId(
+    _In_ PDEVICE_NODE DeviceNode,
+    _In_ BUS_QUERY_ID_TYPE IdType,
+    _Out_ PWCHAR *Id);
+
 //
 // Global I/O Data
 //
--- a/ntoskrnl/io/pnpmgr/devaction.c
+++ b/ntoskrnl/io/pnpmgr/devaction.c
@@ -1595,7 +1595,9 @@ PiStartDeviceFinal(
     if (NT_SUCCESS(Status))
     {
         PiSetDeviceCapabilities(DeviceNode, &DeviceCapabilities);
-    } else {
+    }
+    else
+    {
         DPRINT("PiIrpQueryPnPDeviceCapabilities() failed (Status 0x%08lx)\n", Status);
     }
 
--- a/ntoskrnl/io/pnpmgr/devaction.c
+++ b/ntoskrnl/io/pnpmgr/devaction.c
@@ -931,7 +931,7 @@ NTSTATUS
 IopQueryHardwareIds(PDEVICE_NODE DeviceNode,
                     HANDLE InstanceKey)
 {
-    PWCHAR HwId;
+    PZZWSTR HwIds;
     PWSTR Ptr;
     UNICODE_STRING ValueName;
     NTSTATUS Status;
@@ -940,10 +940,10 @@ IopQueryHardwareIds(PDEVICE_NODE DeviceNode,
 
     DPRINT("IopQueryHardwareIds() query BusQueryHardwareIDs from device stack\n");
 
-    Status = PiIrpQueryPnPDeviceId(DeviceNode, BusQueryHardwareIDs, &HwId);
+    Status = PiIrpQueryPnPDeviceId(DeviceNode, BusQueryHardwareIDs, &HwIds);
     if (NT_SUCCESS(Status))
     {
-        IsValidID = IopValidateID(HwId, BusQueryHardwareIDs);
+        IsValidID = IopValidateID(HwIds, BusQueryHardwareIDs);
 
         if (!IsValidID)
         {
@@ -952,7 +952,7 @@ IopQueryHardwareIds(PDEVICE_NODE DeviceNode,
 
         TotalLength = 0;
 
-        Ptr = (PWSTR)HwId;
+        Ptr = (PWSTR)HwIds;
         DPRINT("Hardware IDs:\n");
         while (*Ptr)
         {
@@ -970,15 +970,15 @@ IopQueryHardwareIds(PDEVICE_NODE DeviceNode,
                                &ValueName,
                                0,
                                REG_MULTI_SZ,
-                               (PVOID)HwId,
+                               (PVOID)HwIds,
                                (TotalLength + 1) * sizeof(WCHAR));
         if (!NT_SUCCESS(Status))
         {
             DPRINT1("ZwSetValueKey() failed (Status %lx)\n", Status);
         }
 
-        if (HwId)
-            ExFreePool(HwId);
+        if (HwIds)
+            ExFreePool(HwIds);
     }
     else
     {
--- a/ntoskrnl/io/pnpmgr/devaction.c
+++ b/ntoskrnl/io/pnpmgr/devaction.c
@@ -940,7 +940,6 @@ IopQueryHardwareIds(PDEVICE_NODE DeviceNode,
 
     DPRINT("IopQueryHardwareIds() query BusQueryHardwareIDs from device stack\n");
 
-     ;
     Status = PiIrpQueryPnPDeviceId(DeviceNode, BusQueryHardwareIDs, &HwId);
     if (NT_SUCCESS(Status))
     {
--- a/ntoskrnl/io/pnpmgr/devaction.c
+++ b/ntoskrnl/io/pnpmgr/devaction.c
@@ -248,12 +248,12 @@ IopCreateDeviceInstancePath(
 
     DPRINT("Sending IRP_MN_QUERY_CAPABILITIES to device stack (after enumeration)\n");
 
-    Status = IopQueryDeviceCapabilities(DeviceNode, &DeviceCapabilities);
+    Status = PiIrpQueryPnPDeviceCapabilities(DeviceNode, &DeviceCapabilities);
     if (!NT_SUCCESS(Status))
     {
         if (Status != STATUS_NOT_SUPPORTED)
         {
-            DPRINT1("IopQueryDeviceCapabilities() failed (Status 0x%08lx)\n", Status);
+            DPRINT1("PiIrpQueryPnPDeviceCapabilities() failed (Status 0x%08lx)\n", Status);
         }
         RtlFreeUnicodeString(&DeviceId);
         return Status;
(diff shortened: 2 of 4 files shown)


ntoskrnl/include/internal/io.h
```diff
@@ -629,7 +629,7 @@ IopFreeDeviceNode(
 
 NTSTATUS
 NTAPI
-IopQueryDeviceCapabilities(PDEVICE_NODE DeviceNode,
+PiSetDeviceCapabilities(PDEVICE_NODE DeviceNode,
                            PDEVICE_CAPABILITIES DeviceCaps);
```
JoachimHenze, line 633:
```suggestion
    PDEVICE_CAPABILITIES DeviceCaps);
```

ntoskrnl/io/pnpmgr/devaction.c
```diff
@@ -851,40 +851,13 @@ PiCallDriverAddDevice(
 
 NTSTATUS
 NTAPI
-IopQueryDeviceCapabilities(PDEVICE_NODE DeviceNode,
+PiSetDeviceCapabilities(PDEVICE_NODE DeviceNode,
                            PDEVICE_CAPABILITIES DeviceCaps)
```
JoachimHenze, line 855:
```suggestion
    PDEVICE_CAPABILITIES DeviceCaps)
```

ntoskrnl/io/pnpmgr/pnpirp.c
```diff
@@ -306,3 +306,61 @@ PiIrpQueryPnPDeviceState(
...
+    PAGED_CODE();
+
+    ASSERT(DeviceNode);
+    ASSERT(IdType == BusQueryDeviceID || IdType == BusQueryHardwareIDs ||
+           IdType == BusQueryCompatibleIDs || IdType == BusQueryInstanceID ||
+           IdType == BusQueryDeviceSerialNumber || IdType == BusQueryContainerID);
+
+    IO_STACK_LOCATION stack = {
+        .MajorFunction = IRP_MJ_PNP,
+        .MinorFunction = IRP_MN_QUERY_ID
```
JoachimHenze, line 352: Do not initialize this struct in C99-style! It is not compatible to older compilers.

JoachimHenze (changes requested):

ntoskrnl/include/internal/io.h
HBelusca: This suggestion doesn't make much more sense.

ntoskrnl/io/pnpmgr/devaction.c
```diff
@@ -957,25 +930,20 @@ NTSTATUS
...
-    DPRINT("Sending IRP_MN_QUERY_ID.BusQueryHardwareIDs to device stack\n");
+    DPRINT("IopQueryHardwareIds() query BusQueryHardwareIDs from device stack\n");
 
-    RtlZeroMemory(&Stack, sizeof(Stack));
-    Stack.Parameters.QueryId.IdType = BusQueryHardwareIDs;
-    Status = IopInitiatePnpIrp(DeviceNode->PhysicalDeviceObject,
-                               &IoStatusBlock,
-                               IRP_MN_QUERY_ID,
-                               &Stack);
+     ;
```
HBelusca, line 942:
```suggestion
```

ntoskrnl/io/pnpmgr/devaction.c
HBelusca: idem

ntoskrnl/io/pnpmgr/devaction.c
```diff
@@ -1612,10 +1590,12 @@ PiStartDeviceFinal(
...
     DPRINT("Sending IRP_MN_QUERY_CAPABILITIES to device stack (after start)\n");
 
-    Status = IopQueryDeviceCapabilities(DeviceNode, &DeviceCapabilities);
-    if (!NT_SUCCESS(Status))
+    Status = PiIrpQueryPnPDeviceCapabilities(DeviceNode, &DeviceCapabilities);
+    if (NT_SUCCESS(Status))
     {
-        DPRINT("IopInitiatePnpIrp() failed (Status 0x%08lx)\n", Status);
+        PiSetDeviceCapabilities(DeviceNode, &DeviceCapabilities);
+    } else {
```
HBelusca, line 1597:
```suggestion
    }
    else
    {
```

ntoskrnl/io/pnpmgr/pnpirp.c
```diff
@@ -306,3 +306,61 @@ PiIrpQueryPnPDeviceState(
...
+    PVOID Dummy;
+    return IopSynchronousCall(DeviceNode->PhysicalDeviceObject, &stack, &Dummy);
+}
+
+// IRP_MN_QUERY_ID (0x13)
+NTSTATUS
+PiIrpQueryPnPDeviceId(
+    _In_ PDEVICE_NODE DeviceNode,
+    _In_ BUS_QUERY_ID_TYPE IdType,
+    _Out_ PWCHAR *Id)
```
HBelusca, line 341: Is it a string that is guaranteed to be always NULL-terminated?

ntoskrnl/include/internal/io.h
JoachimHenze: It does. Because it never ever has to be changed again, no matter to which name the function would be renamed to.

ntoskrnl/io/pnpmgr/pnpirp.c
```diff
@@ -306,3 +306,61 @@ PiIrpQueryPnPDeviceState(
...
+
+    RtlZeroMemory(DeviceCaps, sizeof(DEVICE_CAPABILITIES));
+    DeviceCaps->Size = sizeof(DEVICE_CAPABILITIES);
+    DeviceCaps->Version = 1;
+    DeviceCaps->Address = -1;
+    DeviceCaps->UINumber = -1;
+
+    IO_STACK_LOCATION stack = {
+        .MajorFunction = IRP_MJ_PNP,
+        .MinorFunction = IRP_MN_QUERY_CAPABILITIES
```
JoachimHenze, line 328: C99 struct init should be avoided to improve compiler compat. Same at line351.

JoachimHenze (changes requested):

ntoskrnl/include/internal/io.h
olase23: I admit the indentation in my commit was not the best. But the suggestion to drop the indentation for the second parameter completely looks odd since all other function declarations use it.

binarymaster: Well, this change introduces an odd indentation for the function arguments, as you can see:

```c
PiSetDeviceCapabilities(PDEVICE_NODE DeviceNode,
                           PDEVICE_CAPABILITIES DeviceCaps);
```

The first argument is located more to the left, and the second is more to the right. In order to make this consistent, I would suggest doing it like in other places:

```c
PiSetDeviceCapabilities(
    PDEVICE_NODE DeviceNode,
    PDEVICE_CAPABILITIES DeviceCaps);
```

And ideally add SAL2 annotation here as well.

ntoskrnl/io/pnpmgr/devaction.c
binarymaster: Same here + SAL2 annotations.

```c
PiSetDeviceCapabilities(
    PDEVICE_NODE DeviceNode,
    PDEVICE_CAPABILITIES DeviceCaps)
```

ntoskrnl/io/pnpmgr/pnpirp.c
```diff
@@ -306,3 +306,61 @@ PiIrpQueryPnPDeviceState(
...
+    };
```
binarymaster, line 329: While C99-style variable initialization is allowed (we have these in the kernel already), you need to place them at the start of the code block -- this is still just plain C code, not C++.  It's my own preference though, others may have different opinion.

ntoskrnl/io/pnpmgr/devaction.c
```diff
@@ -1002,16 +969,19 @@ IopQueryHardwareIds(PDEVICE_NODE DeviceNode,
...
-                               (PVOID)IoStatusBlock.Information,
+                               (PVOID)HwId,
                                (TotalLength + 1) * sizeof(WCHAR));
         if (!NT_SUCCESS(Status))
         {
             DPRINT1("ZwSetValueKey() failed (Status %lx)\n", Status);
         }
+
+        if (HwId)
+            ExFreePool(HwId);
```
GeoB99, line 980: As I currently understand the `PiIrpQueryPnPDeviceId` routine initializes the `Id` pointer into the stack, I do not see anywhere that the routine allocates this pointer in a pool. Are you sure you really intend to free the pool there?

ntoskrnl/io/pnpmgr/pnpirp.c
```diff
@@ -306,3 +306,61 @@ PiIrpQueryPnPDeviceState(
 
     return status;
 }
+
+// IRP_MN_QUERY_CAPABILITIES (0x09)
+NTSTATUS
+PiIrpQueryPnPDeviceCapabilities(
```
GeoB99, line 312: It is a good practice that newly added functions in the kernel have to be documented using Doxygen format, so that people can understand more or less how the newly implemented routines in question work or their purpose.

https://reactos.org/wiki/Documentation_Guidelines

See an example -> https://github.com/reactos/reactos/blob/master/ntoskrnl/se/tokenadj.c#L17

GeoB99 (commented):
Thank you very much for participating in the kernel development! I have left two review comments that may be worthwhile, as for the rest I do not have much to say.

ntoskrnl/io/pnpmgr/pnpirp.c
binarymaster: Pushed a commit 77b4d0f8 to fix remaining C++'ish things.

ntoskrnl/io/pnpmgr/pnpirp.c
```diff
@@ -306,3 +306,62 @@ PiIrpQueryPnPDeviceState(
...
+
+// IRP_MN_QUERY_ID (0x13)
+NTSTATUS
+PiIrpQueryPnPDeviceId(
+    _In_ PDEVICE_NODE DeviceNode,
+    _In_ BUS_QUERY_ID_TYPE IdType,
+    _Out_ PWCHAR *Id)
+{
+    NTSTATUS Status;
+    ULONG_PTR LongId;
```
HBelusca, line 344: Why using that intermediate variable? Why not just passing the `Id` parameter directly? (as in the suggestion below).

ntoskrnl/io/pnpmgr/pnpirp.c
```diff
@@ -306,3 +306,62 @@ PiIrpQueryPnPDeviceState(
...
+
+    ASSERT(DeviceNode);
+    ASSERT(IdType == BusQueryDeviceID || IdType == BusQueryHardwareIDs ||
+           IdType == BusQueryCompatibleIDs || IdType == BusQueryInstanceID ||
+           IdType == BusQueryDeviceSerialNumber || IdType == BusQueryContainerID);
+
+    Stack.Parameters.QueryId.IdType = IdType;
+    *Id = NULL;
+
+    Status = IopSynchronousCall(DeviceNode->PhysicalDeviceObject, &Stack, (PVOID)&LongId);
```
HBelusca, line 360:
```suggestion
    Status = IopSynchronousCall(DeviceNode->PhysicalDeviceObject, &Stack, Id);
```

ntoskrnl/io/pnpmgr/devaction.c
```diff
@@ -957,25 +931,19 @@ NTSTATUS
 IopQueryHardwareIds(PDEVICE_NODE DeviceNode,
                     HANDLE InstanceKey)
 {
-    IO_STACK_LOCATION Stack;
-    IO_STATUS_BLOCK IoStatusBlock;
+    PWCHAR HwId;
```
HBelusca, line 934: Use `PWSTR` if the pointer will point to a string that is NULL-terminated.
Also, if the string actually points to a (null-termianted) **multi-string**, then the `PZZWSTR` type may be better (and the name be: `HwIds` in plural).

ntoskrnl/io/pnpmgr/devaction.c
HBelusca: `HwId` allocated by some function called by `PiIrpQueryPnPDeviceId` (`IRP_MN_QUERY_ID` handling by some driver called by `IopSynchronousCall`).

olase23: A driver returns ID(s) in response to the IRP, it allocates a WCHAR structure from paged pool to contain the ID(s). The PnP manager frees the structure when it is no longer needed.

ntoskrnl/io/pnpmgr/pnpirp.c
olase23: Well, this is what MSDN says:
If a driver returns ID(s) in response to this IRP, it allocates a WCHAR structure from paged pool to contain the ID(s). The PnP manager frees the structure when it is no longer needed.

HBelusca: Of course. But if it doesn't also return the size of that string allocation, this certainly means that this string has to be null-terminated. Otherwise nobody would know where to stop its parsing.

olase23: What do you suggest?

HBelusca: If it's just one null-terminated string, then `PWSTR* Id`; if it's a multi-string, then `PZZWSTR* Id` (if there can be many, then `Ids` in plural maybe?)

ntoskrnl/io/pnpmgr/devaction.c
```diff
@@ -248,12 +248,12 @@ IopCreateDeviceInstancePath(
...
 
-    Status = IopQueryDeviceCapabilities(DeviceNode, &DeviceCapabilities);
+    Status = PiIrpQueryPnPDeviceCapabilities(DeviceNode, &DeviceCapabilities);
     if (!NT_SUCCESS(Status))
     {
         if (Status != STATUS_NOT_SUPPORTED)
         {
-            DPRINT1("IopQueryDeviceCapabilities() failed (Status 0x%08lx)\n", Status);
+            DPRINT1("PiIrpQueryPnPDeviceCapabilities() failed (Status 0x%08lx)\n", Status);
         }
```
Extravert-ir, line 257: Please, just remove this DPRINT. I'm going to implement some kind of central logging inside pnpirp.c

ntoskrnl/io/pnpmgr/devaction.c
```diff
@@ -1612,10 +1590,14 @@ PiStartDeviceFinal(
 
     DPRINT("Sending IRP_MN_QUERY_CAPABILITIES to device stack (after start)\n");
```
Extravert-ir, line 1591: This dprint should also go away

ntoskrnl/io/pnpmgr/devaction.c
```diff
@@ -248,12 +248,12 @@ IopCreateDeviceInstancePath(
 
     DPRINT("Sending IRP_MN_QUERY_CAPABILITIES to device stack (after enumeration)\n");
 
-    Status = IopQueryDeviceCapabilities(DeviceNode, &DeviceCapabilities);
+    Status = PiIrpQueryPnPDeviceCapabilities(DeviceNode, &DeviceCapabilities);
```
Extravert-ir, line 251: Hm, let's just pass device capabilities into `IopCreateDeviceInstancePath` as a parameter, so `PiInitializeDevNode` makes only one `IRP_MN_QUERY_CAPABILITIES` IRP

Extravert-ir (changes requested):
Hey!
I'm finally here, reverted some coding style stuff (sorry). Regarding doxygen comments - this is not the right time to do them. They may be added later for all IRP functions at once.

Now, real feedback:
Let's first finish with `IRP_MN_QUERY_CAPABILITIES`/`PiIrpQueryPnPDeviceCapabilities`. I've left a couple of comments about it, and also there is one raw usage of `IRP_MN_QUERY_CAPABILITIES` left in `pnpmgr/plugplay.c`. Other than that, it should be good to go.

Regarding `IRP_MN_QUERY_ID`, there are still places where this IRP is sent avoiding a `PiIrpQueryPnPDeviceId` call (in `devaction.c` and some other places). Also, I'd like to put `IopValidateID` right into `pnpirp.c`, so it is a static function there.

I'd like to clarify, my ultimate goal here is to remove `IopSynchronousCall` and `IopInitiatePnpIrp` calls outside of `pnpirp.c` file completely, so all IRP sending code is located in a single file.

P.S. Thank you for the contribution, and sorry for the mess with reviews. You can reach me in mattermost/matrix if you have more questions.

ntoskrnl/io/pnpmgr/pnpirp.c
Extravert-ir: `IRP_MN_QUERY_ID` returns different strings depending on the ID type requested. So let's just keep it `PWSTR`.

ntoskrnl/io/pnpmgr/devaction.c
olase23: Ok, i'll have a look on that.

PR #5328, opened 2023-06-10
[NTUSER] Fix an unintialized user's token variable case

GeoB99:
And remove the `!NT_SUCCESS(Status` check which is excessive, the expected status will always be **STATUS_BUFFER_TOO_SMALL** anyway. This should fix some compilation warnings spotted by GCC. 

Courtesy goes to Hermes for letting me know of these warnings.

### DIFF
--- a/win32ss/user/ntuser/security.c
+++ b/win32ss/user/ntuser/security.c
@@ -2,7 +2,7 @@
  * PROJECT:         ReactOS Win32k subsystem
  * LICENSE:         GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
  * PURPOSE:         Security infrastructure of NTUSER component of Win32k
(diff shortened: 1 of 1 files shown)


win32ss/user/ntuser/security.c
```diff
@@ -196,7 +196,7 @@ IntQueryUserSecurityIdentification(
                                      NULL,
                                      0,
                                      &BufferLength);
-    if (!NT_SUCCESS(Status) && Status == STATUS_BUFFER_TOO_SMALL)
+    if (Status == STATUS_BUFFER_TOO_SMALL)
```
HBelusca, line 199: at line 214 (the end of this if-block), you may add a
```c
else if (!NT_SUCCESS(Status))
{
    //some err...
    ZwClose(Token);
    return Status;
}
```
in case that first ZwQueryInformationToken fails but for a reason diffferent from buffer too small.

PR #5333, opened 2023-06-12
[MMSYS] Use PWSTR instead of PWCHAR for string

TAN-Gaming:


dll/cpl/mmsys/sounds.c
```diff
@@ -82,7 +82,7 @@ LPWSTR MakeFilter(LPWSTR psz)
     return psz;
 }
 
-PLABEL_MAP FindLabel(PGLOBAL_DATA pGlobalData, PAPP_MAP pAppMap, PWCHAR szName)
+PLABEL_MAP FindLabel(PGLOBAL_DATA pGlobalData, PAPP_MAP pAppMap, PWSTR szName)
```
HBelusca, line 85: If those strings are passed _**as inputs only**_ to the function, and are _**not**_ modified by it, then you could even use `PCWSTR`.

TAN-Gaming: Done.

HBelusca (approved):

PR #5337, opened 2023-06-14
[MSPAINT] Improve CMiniatureWindow

katahiromz:
## Purpose
Fix the miniature window.

## Proposed changes

- Save the position and size of the miniature window.
- Improve drawing of the miniature window.
- Sync with the canvas.

## Screenshot

image

## TODO

- [x] Do tests.

### DIFF
--- a/base/applications/mspaint/history.cpp
+++ b/base/applications/mspaint/history.cpp
@@ -18,7 +18,7 @@ void ImageModel::NotifyImageChanged()
     if (canvasWindow.IsWindow())
         canvasWindow.Invalidate(FALSE);
     if (miniature.IsWindow())
-        miniature.Invalidate(TRUE);
+        miniature.Invalidate(FALSE);
 }
 
 ImageModel::ImageModel()
--- a/base/applications/mspaint/history.cpp
+++ b/base/applications/mspaint/history.cpp
@@ -17,6 +17,8 @@ void ImageModel::NotifyImageChanged()
 {
     if (canvasWindow.IsWindow())
         canvasWindow.Invalidate(FALSE);
+    if (miniature.IsWindow())
+        miniature.Invalidate(TRUE);
 }
 
 ImageModel::ImageModel()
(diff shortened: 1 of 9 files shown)


base/applications/mspaint/miniature.cpp
```diff
@@ -32,23 +33,72 @@ HWND CMiniatureWindow::DoCreate(HWND hwndParent)
...
+    // Start painting
     PAINTSTRUCT ps;
     HDC hDC = BeginPaint(&ps);
-    StretchBlt(hDC, 0, 0, rc.right, rc.bottom,
-               imageModel.GetDC(), 0, 0, imageModel.GetWidth(), imageModel.GetHeight(),
-               SRCCOPY);
+    if (!hDC)
+        return 0;
+
+    // Use a memory bitmap to reduce flickering
```
HBelusca, line 79: Does this indeed give a visual difference when run on Windows?

base/applications/mspaint/history.cpp
```diff
@@ -17,6 +17,8 @@ void ImageModel::NotifyImageChanged()
 {
     if (canvasWindow.IsWindow())
         canvasWindow.Invalidate(FALSE);
+    if (miniature.IsWindow())
+        miniature.Invalidate(TRUE);
```
HBelusca, line 21: Is the "TRUE" parameter here used to make the miniature window redraw completely?

katahiromz: This `TRUE` is for `::InvalidateRect` function's 3rd parameter `bErase`.
`FALSE` is also OK for this implementation.

base/applications/mspaint/miniature.cpp
katahiromz: Yep. The miniature window does flickering.

base/applications/mspaint/history.cpp
HBelusca: ok yes, I wanted to understand why the canvas got FALSE while the miniature got TRUE.

katahiromz: Fixed.

PR #5338, opened 2023-06-14
[MSPAINT] Add CheckHandleLeaks function

katahiromz:
## Purpose
Detect handle leaks easily.

## Proposed changes

- Add `CheckHandleLeaks` function to detect handle leaks.

## Screenshot

image

## TODO

- [ ] Check if it is working.

### DIFF
--- a/base/applications/mspaint/main.cpp
+++ b/base/applications/mspaint/main.cpp
@@ -182,11 +182,9 @@ HWND CMainWindow::DoCreate()
 #ifdef _DEBUG
 void CheckHandleLeaks(void)
 {
-#if (WINVER >= 0x0500)
     HANDLE hProcess = ::GetCurrentProcess();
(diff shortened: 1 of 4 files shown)


base/applications/mspaint/main.cpp
```diff
@@ -179,13 +179,28 @@ HWND CMainWindow::DoCreate()
     return Create(HWND_DESKTOP, rc, strTitle, WS_OVERLAPPEDWINDOW, WS_EX_ACCEPTFILES);
 }
 
+#ifdef _DEBUG
+void CheckHandleLeaks(void)
+{
+#if (WINVER >= 0x0500)
```
HBelusca, line 185: Not sure this is important. Everything we implement is of NT >= 5.

katahiromz: Deleted.

katahiromz:
Because GDI handles are sometimes cached, we cannot rely on the number of handles to detect handle leaks.
Closed.

PR #5342, opened 2023-06-15
[SCSIPORT] Simplify storage of device type identifiers.

HBelusca:
Addendum to commits 87d9dfae0 (r74595) and a97c6e0aa .


Extravert-ir:
I don't think this is better. Will reject that, sorry

PR #5344, opened 2023-06-15
[COMCTL32][MSHTML][QCAP][MSGINA][UXTHEME][MSHTML_WINETEST][NTGDI][NTUSER] DeleteObject(hdc...) are wrong!

katahiromz:
## Purpose

Using `DeleteObject` against a DC is totally wrong. Use `DeleteDC` instead to delete a DC.

JIRA issue: N/A

## Proposed changes

- Use `DeleteDC` instead of `DeleteObject` against DCs.

## TODO

- [x] Build.
- [ ] Big tests.

### DIFF
--- a/win32ss/gdi/ntgdi/dclife.c
+++ b/win32ss/gdi/ntgdi/dclife.c
@@ -930,7 +930,7 @@ IntGdiDeleteDC(HDC hDC, BOOL Force)
 
     if (GreIsHandleValid(hDC))
     {
-        if (!IntGdiDeleteDC(hDC, FALSE))
+        if (!GreDeleteObject(hDC))
         {
             DPRINT1("DC_FreeDC failed\n");
             return FALSE;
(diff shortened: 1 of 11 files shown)


win32ss/gdi/ntgdi/dclife.c
```diff
@@ -930,7 +930,7 @@ IntGdiDeleteDC(HDC hDC, BOOL Force)
 
     if (GreIsHandleValid(hDC))
     {
-        if (!GreDeleteObject(hDC))
+        if (!IntGdiDeleteDC(hDC, FALSE))
```
HBelusca, line 933: hm you're doing a recursive call there.

HBelusca (changes requested):

win32ss/gdi/ntgdi/dclife.c
katahiromz: Reverted.

tkreuzer:
No, it's not "totally wrong". While it's not documented, you can call DeleteObject with a DC just fine on Windows. A quick look at our implementation of DeleteObject also reveals that.
I'm fine with changing this in our own code, but not in wine, since it fixes nothing and only causes useless deviation from the original code.

tkreuzer (changes requested):
Don't change wine code, since the "fix" doesn't actually fix anything.

katahiromz:
Sorry my half-rotten brain. My brain is weak in undocumented things. Closed.

PR #5346, opened 2023-06-15
[NTOS:PNP] Assert the presence of a device extension when handling PnP root power IRPs

GeoB99:
Handling PnP root driver power IRPs requires that a device object must come up with a device extension to determine whether it is a function driver and if so, handle the IRP accordingly.

**JIRA Issue:** [CORE-18989](https://jira.reactos.org/browse/CORE-18989)

### DIFF
--- a/ntoskrnl/io/pnpmgr/pnproot.c
+++ b/ntoskrnl/io/pnpmgr/pnproot.c
@@ -1473,6 +1473,12 @@ PnpRootPowerControl(
     Status = Irp->IoStatus.Status;
     IrpSp = IoGetCurrentIrpStackLocation(Irp);
 
+    /*
+     * We must handle power IRPs based on whether it is a function driver
+     * or not from the device extension, so it cannot be NULL.
+     */
+    ASSERT(DeviceExtension);
+
     if (DeviceExtension->Common.IsFDO)
     {
         ASSERT(!DeviceExtension->Common.IsFDO);


ntoskrnl/io/pnpmgr/pnproot.c
```diff
@@ -1473,7 +1473,7 @@ PnpRootPowerControl(
     Status = Irp->IoStatus.Status;
     IrpSp = IoGetCurrentIrpStackLocation(Irp);
 
-    if (DeviceExtension->Common.IsFDO)
+    if (DeviceExtension && DeviceExtension->Common.IsFDO)
```
HBelusca, line 1476: Looking at this and the comments from @Extravert-ir from #kernel-dev, it may be of interest to have an `ASSERT(DeviceExtension);` if the `DeviceExtension` actually must *not* be NULL there.

ThFabba: Yeah, this doesn't make sense to me. The "pnp root" driver is in complete control over whether its device objects have a device extension (and the device extension is part of the same allocation as the device object, so there is no case where it's null due to OOM), and it can't get requests for device objects it doesn't own.
Something else is going on here...

binarymaster (approved):
LGTM, one note though, make sure to prefix this as `[NTOS:PNP]`.

ntoskrnl/io/pnpmgr/pnproot.c
```diff
@@ -1473,6 +1473,12 @@ PnpRootPowerControl(
     Status = Irp->IoStatus.Status;
     IrpSp = IoGetCurrentIrpStackLocation(Irp);
 
+    /*
+     * We must handle power IRPs based on whether it is a function driver
+     * or not from the device extension, so it cannot be be NULL.
```
SergeGautherie, line 1478:
```suggestion
     * or not from the device extension, so it cannot be NULL.
```

PR #5357, opened 2023-06-19
[MSPAINT][ATL] Send mail with image

katahiromz:
## Purpose
Implement sending an image file via e-mail on `IDM_FILESEND` command action.

## Proposed changes

- Include `<mapi.h>` and use `mapi32.dll` dynamically.
- Add `OpenMailer`, `ShowError` and `SetFileInfo` helper functions.
- By using `MAPISendMail` function, open the mailer with attachment.
- Add `IDS_SAVEERROR` and `IDS_CANTSENDMAIL` resource strings.
- Add `g_szTempFile` global variable.
- Rename `DoLoadImageFile` as `LoadDIBFromFile`.
- Add `ImageModel::LoadImage`.

## TODO

- [x] Do tests.


base/applications/mspaint/winproc.cpp
```diff
@@ -657,6 +658,16 @@ LRESULT CMainWindow::OnCommand(UINT nMsg, WPARAM wParam, LPARAM lParam, BOOL& bH
             if (pd.hDevNames)
                 GlobalFree(pd.hDevNames);
             break;
+        case IDM_FILESEND:
+            if (PathFileExists(g_szFileName))
+            {
+                // FIXME: URL escape and localization
+                TCHAR szCmdLine[1024];
```
HBelusca, line 665: Use a `CString` variable.

base/applications/mspaint/winproc.cpp
```diff
@@ -657,6 +689,21 @@ LRESULT CMainWindow::OnCommand(UINT nMsg, WPARAM wParam, LPARAM lParam, BOOL& bH
...
+            {
+                CStringW strSubject;
+                strSubject.LoadString(IDS_MAILSUBJECT);
+
+                // NOTE: "%22" is for a double quote
+                CString strCmdLine = _T("mailto:?subject=%22");
+                strCmdLine += URLEscapeW(strSubject);
+                strCmdLine += _T("%22&attach=%22");
+                strCmdLine += URLEscapeW(g_szFileName);
+                strCmdLine += _T("%22");
```
HBelusca, line 703: Alternatively:
```suggestion
                CString strCmdLine;
                strCmdLine.Format(_T("mailto:?subject=%%22%s%%22&attach=%%22%s%%22"),
                                  URLEscapeW(strSubject), URLEscapeW(g_szFileName));
```
Notice that you are mixing TCHARs with explicit unicode strings (from URLEscape**W**) .

base/applications/mspaint/winproc.cpp
```diff
@@ -575,6 +576,37 @@ LRESULT CMainWindow::OnSysColorChange(UINT nMsg, WPARAM wParam, LPARAM lParam, B
...
 }
 
+// NOTE: The standard isalnum is locale-dependent. We don't use it in this case.
+static BOOL my_isalnum(char ch)
+{
+    return (('a' <= ch && ch <= 'z') || ('A' <= ch && ch <= 'Z') ||
+            ('0' <= ch && ch <= '9'));
+}
+
+static CStringW URLEscapeW(LPCWSTR psz)
```
HBelusca, line 586: It would be better to use the existing shlwapi `UrlEscape` function.

base/applications/mspaint/winproc.cpp
```diff
@@ -575,6 +576,20 @@ LRESULT CMainWindow::OnSysColorChange(UINT nMsg, WPARAM wParam, LPARAM lParam, B
...
+    char utf8[256];
+    ::WideCharToMultiByte(CP_UTF8, 0, psz, -1, utf8, _countof(utf8), NULL, NULL);
+
+    char url[256];
+    DWORD cch = _countof(url);
+    UrlEscapeA(utf8, url, &cch, URL_ESCAPE_PERCENT | URL_ESCAPE_SEGMENT_ONLY);
+
+    WCHAR wide[256];
+    ::MultiByteToWideChar(CP_ACP, 0, url, -1, wide, _countof(wide));
+    return CStringW(wide);
```
HBelusca, line 590: I don't understand why you go to UTF8 then go back to UTF16, instead of just calling `UrlEscapeW` .

base/applications/mspaint/winproc.cpp
```diff
@@ -657,6 +672,20 @@ LRESULT CMainWindow::OnCommand(UINT nMsg, WPARAM wParam, LPARAM lParam, BOOL& bH
...
                 GlobalFree(pd.hDevNames);
             break;
+        case IDM_FILESEND:
+            if (PathFileExists(g_szFileName))
+            {
+                CStringW strSubject;
+                strSubject.LoadString(IDS_MAILSUBJECT);
+
+                // NOTE: "%22" is for a double quote
+                CString strCmdLine;
```
HBelusca, line 682: Why not using directly unicode strings here `CStringW` instead of the TCHAR thing?

base/applications/mspaint/winproc.cpp
```diff
@@ -657,6 +666,20 @@ LRESULT CMainWindow::OnCommand(UINT nMsg, WPARAM wParam, LPARAM lParam, BOOL& bH
...
+            if (PathFileExists(g_szFileName))
+            {
+                CStringW strSubject;
+                strSubject.LoadString(IDS_MAILSUBJECT);
+
+                // NOTE: "%22" is for a double quote
+                CStringW strCmdLine;
+                strCmdLine.Format(L"mailto:?subject=%%22%s%%22&attach=%%22%s%%22",
+                                  (LPCWSTR)EscapeURLW(strSubject),
+                                  (LPCWSTR)EscapeURLW(g_szFileName));
```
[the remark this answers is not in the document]
katahiromz: Thanks.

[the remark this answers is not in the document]
katahiromz: @binarymaster Can you test whether the command opens the mailer?

base/applications/mspaint/winproc.cpp
```diff
@@ -575,6 +577,26 @@ LRESULT CMainWindow::OnSysColorChange(UINT nMsg, WPARAM wParam, LPARAM lParam, B
...
+    char szUtf8[256];
+    ::WideCharToMultiByte(CP_UTF8, 0, psz, -1, szUtf8, _countof(szUtf8), NULL, NULL);
+    int cchUtf8 = ::lstrlenA(szUtf8);
+
+    // Convert to Base64
+    DWORD cchRet;
+    CStringW strRet;
+    if (!::CryptBinaryToStringW((const BYTE*)szUtf8, cchUtf8, CRYPT_STRING_BASE64, NULL, &cchRet))
+        return strRet;
+    ::CryptBinaryToStringW((const BYTE*)szUtf8, cchUtf8, CRYPT_STRING_BASE64,
```
HBelusca, line 592: Seriously. This is MSPAINT, not a webbrowser or a mail client here.
The shell function UtfEscape didn't work for those strings, or what?

base/applications/mspaint/winproc.cpp
HBelusca: Any care to explain instead of closing as resolved?

HBelusca (changes requested):

HBelusca:
By the way: attachments specification with mailto protocol doesn't seem to be a standard thing or a possibility, see for example: https://mailtrap.io/blog/mailto-links-explained/#Add-an-attachment

base/applications/mspaint/winproc.cpp
binarymaster: Maybe it's possible to use some APIs of sendmail shell extension?

base/applications/mspaint/winproc.cpp
```diff
@@ -575,6 +577,42 @@ LRESULT CMainWindow::OnSysColorChange(UINT nMsg, WPARAM wParam, LPARAM lParam, B
...
+
+        message.nFileCount = 1;
+        message.lpFiles = &attachment;
+    }
+
+    ULONG result = MAPI_E_NOT_SUPPORTED;
+    HINSTANCE hMAPI32 = LoadLibraryW(L"mapi32.dll");
+    if (hMAPI32)
+    {
+        FN_MAPISendMail pMAPISendMail = (FN_MAPISendMail)GetProcAddress(hMAPI32, "MAPISendMail");
```
HBelusca, line 606: Why not using the unicode function directly: MAPISendMailW ?
This would avoid using that inelegant explicit unicode to ansi conversion you do above.

base/applications/mspaint/winproc.cpp
```diff
@@ -575,6 +577,42 @@ LRESULT CMainWindow::OnSysColorChange(UINT nMsg, WPARAM wParam, LPARAM lParam, B
     return 0;
 }
 
+typedef ULONG (WINAPI *FN_MAPISendMail)(LHANDLE, ULONG_PTR, lpMapiMessage, FLAGS, ULONG);
+
+ULONG CMainWindow::OpenMailerWithAttachmentW(LPCWSTR pszFileName)
+{
+    if (!PathFileExists(pszFileName))
+        return MAPI_E_ATTACHMENT_NOT_FOUND;
```
HBelusca, line 585: Notice the behaviour MS' own mspaint has:
If the image hasn't been saved already, save it in the user's own temporary-files folder (but doesn't make mspaint remember it saved the file), then attach that temporarily-saved file to the mail.
(This can be observed from the parameters screenshot I attached.)

base/applications/mspaint/winproc.cpp
katahiromz: Fixed.

base/applications/mspaint/winproc.cpp
katahiromz: Fixed. Added `g_szTempFile` global variable.

base/applications/mspaint/winproc.cpp
```diff
@@ -575,6 +577,44 @@ LRESULT CMainWindow::OnSysColorChange(UINT nMsg, WPARAM wParam, LPARAM lParam, B
     return 0;
 }
 
+typedef ULONG (WINAPI *FN_MAPISendMailW)(LHANDLE, ULONG_PTR, lpMapiMessageW, FLAGS, ULONG);
+
+ULONG CMainWindow::OpenMailerWithAttachmentW(LPCWSTR pszFileName)
+{
+    if (!PathFileExists(pszFileName))
+    {
+        ::ExpandEnvironmentStrings(_T("%TEMP%\\Image.png"), g_szTempFile, _countof(g_szTempFile));
```
HBelusca, line 586: Why doesn't mspaint use explicit unicode functions and strings by now?

base/applications/mspaint/winproc.cpp
```diff
@@ -384,6 +385,7 @@ void CMainWindow::ProcessFileMenu(HMENU hPopupMenu)
     EnableMenuItem(hPopupMenu, IDM_FILEASWALLPAPERPLANE,     ENABLED_IF(g_isAFile && isBMP));
     EnableMenuItem(hPopupMenu, IDM_FILEASWALLPAPERCENTERED,  ENABLED_IF(g_isAFile && isBMP));
     EnableMenuItem(hPopupMenu, IDM_FILEASWALLPAPERSTRETCHED, ENABLED_IF(g_isAFile && isBMP));
+    EnableMenuItem(hPopupMenu, IDM_FILESEND,                 ENABLED_IF(PathFileExists(g_szFileName)));
```
HBelusca, line 388: Wouldn't this enable the menu only if the file has been already saved once?

base/applications/mspaint/winproc.cpp
katahiromz: Someone wants to port to Win9x?

base/applications/mspaint/winproc.cpp
katahiromz: Fixed.

base/applications/mspaint/winproc.cpp
HBelusca: none wants to do that. Win9x already has an mspaint. Also, the fact ours use gdiplus makes me think it wouldn't be usable on win9x.

HBelusca (changes requested):
There are some bugs present, when testing the x64 version on Windows:
1. The created temporary file doesn't have a uniquely generated name;
2. The mail client doesn't open when clicking on "File / Send"
3. Finally, there is no "Do you want to save the file?" message dialog when closing MSPaint on a newly created and unsaved image on which the Step 2. above has been attempted.

base/applications/mspaint/winproc.cpp
```diff
@@ -575,6 +576,44 @@ LRESULT CMainWindow::OnSysColorChange(UINT nMsg, WPARAM wParam, LPARAM lParam, B
...
+        {
+            g_szTempFile[0] = UNICODE_NULL;
+            return MAPI_E_ATTACHMENT_NOT_FOUND;
+        }
+        pszFileName = g_szTempFile;
+    }
+
+    MapiFileDescW attachment = { 0 };
+    attachment.nPosition = (ULONG)-1;
+    attachment.lpszPathName = (LPWSTR)pszFileName;
```
HBelusca, line 596: You need also to set the `lpszFileName` member of the `attachment` structure (see the screenshot attached).

katahiromz: `MAPISendMailW` is Win8+. Are you OK?

JoachimHenze: We are definitely not ok with breaking our MSpaint on 2k3sp2. I wouldn't block any optional feature though that would be only loaded at runtime when it detects a Win8 is present, and falling back gracefully otherwise.
Iirc even 2k3sp2/XP do offer a way to invoke mailto: via the shell somehow.

katahiromz: Fallback to ANSI added.

base/applications/mspaint/winproc.cpp
```diff
@@ -575,6 +576,100 @@ LRESULT CMainWindow::OnSysColorChange(UINT nMsg, WPARAM wParam, LPARAM lParam, B
...
+        if (!ret)
+        {
+            g_szTempFile[0] = UNICODE_NULL;
+            return MAPI_E_FAILURE; // Failure
+        }
+
+        // Use the temporary file 
+        pszPathName = g_szTempFile;
+        strFileTitle.LoadString(IDS_DEFAULTFILENAME);
+        strFileTitle += L".png";
```
HBelusca, line 615: This extension must correspond to the format the image is saved by the SaveDIBToFile -- This is not evident whether this is the case, knowing that saving images with mspaint can be done with different formats.

base/applications/mspaint/winproc.cpp
```diff
@@ -575,6 +576,100 @@ LRESULT CMainWindow::OnSysColorChange(UINT nMsg, WPARAM wParam, LPARAM lParam, B
...
+    attachment.lpszPathName = szPathNameA;
+    attachment.lpszFileName = szFileTitleA;
+
+    MapiMessage message = { 0 };
+    message.lpszSubject = NULL;
+    message.nFileCount = 1;
+    message.lpFiles = &attachment;
+
+    // Try again but in ANSI version
+    FN_MAPISendMail pMAPISendMail = (FN_MAPISendMail)::GetProcAddress(hMAPI, "MAPISendMail");
```
HBelusca, line 661: OK with that (after reading the comments, MSDN and also https://github.com/microsoft/win32metadata/blob/81344326b62096e7c7d62bc9905d4258103eef1d/generation/WinSDK/RecompiledIdlHeaders/um/MapiUnicodeHelp.h#L8 )

HBelusca:
image
However, the saved image is empty...

base/applications/mspaint/winproc.cpp
```diff
@@ -575,6 +576,100 @@ LRESULT CMainWindow::OnSysColorChange(UINT nMsg, WPARAM wParam, LPARAM lParam, B
...
+typedef ULONG (WINAPI *FN_MAPISendMailW)(LHANDLE, ULONG_PTR, lpMapiMessageW, FLAGS, ULONG);
+
+ULONG CMainWindow::OpenMailerWithAttachmentW(LPCWSTR pszPathName)
+{
+    CStringW strFileTitle;
+    if (PathFileExistsW(pszPathName)) // File does exist?
+    {
+        strFileTitle = PathFindFileNameW(pszPathName);
+    }
+    else // Not existing
```
HBelusca, line 589: Does this cover the case of the file that already exists, *but* has been modified in the meantime (and not saved) before sending it by mail?

base/applications/mspaint/dib.cpp
```diff
@@ -143,41 +143,43 @@ GetDIBHeight(HBITMAP hBitmap)
     return bm.bmHeight;
 }
 
-BOOL SaveDIBToFile(HBITMAP hBitmap, LPCTSTR FileName, HDC hDC)
+BOOL SaveDIBToFile(HBITMAP hBitmap, LPCTSTR FileName, BOOL fIsMainFile)
 {
     CImageDx img;
     img.Attach(hBitmap);
-    img.SaveDx(FileName, GUID_NULL, g_xDpi, g_yDpi); // TODO: error handling
+    img.SaveDx(FileName, GUID_NULL, g_xDpi, g_yDpi);
```
HBelusca, line 150: I don't see any error handling here btw.

base/applications/mspaint/dib.cpp
```diff
@@ -268,12 +268,9 @@ HBITMAP DoLoadImageFile(HWND hwnd, LPCTSTR name, BOOL fIsMainFile)
 
     HBITMAP hBitmap = img.Detach();
```
HBelusca, line 270:
```suggestion
```

base/applications/mspaint/winproc.cpp
```diff
@@ -575,6 +576,100 @@ LRESULT CMainWindow::OnSysColorChange(UINT nMsg, WPARAM wParam, LPARAM lParam, B
...
+    {
+        // Delete the temporary file if any
+        if (g_szTempFile[0])
+            ::DeleteFileW(g_szTempFile);
```
HBelusca, line 593:
```suggestion
        if (g_szTempFile[0])
        {
            ::DeleteFileW(g_szTempFile);
            g_szTempFile[0] = UNICODE_NULL;
        }
```

base/applications/mspaint/winproc.cpp
katahiromz: In this situation, the file is not saved yet, thus we  use a temp file. The best format for unknown image is png.

HBelusca: Yes, but where in the code is that output format specified? Since currently you save the image into the temp file (that has the name afxWHATEVER.tmp) ?

base/applications/mspaint/winproc.cpp
katahiromz: It may need confirmation if save or not.

base/applications/mspaint/winproc.cpp
katahiromz: Can you actually send me an e-mail with attachment by this code?
My address is katayama.hirofumi.mz@gmail.com .

base/applications/mspaint/dib.cpp
katahiromz: Added check.

base/applications/mspaint/winproc.cpp
katahiromz: Sorry.

```c
        // if the file type is null, get the file type from extension
        const GUID *FileType = &guidFileType;
        if (IsGuidEqual(guidFileType, GUID_NULL))
        {
            LPCWSTR pszExt = GetFileExtension(pszNameW);
            FileType = FileTypeFromExtension(pszExt);
        }
```
This code cannot send it.

katahiromz:
```c
        if (IsGuidEqual(guidFileType, GUID_NULL))
        {
            LPCWSTR pszExt = GetFileExtension(pszNameW);
            FileType = FileTypeFromExtension(pszExt);
            if (!FileType)
            {
                FileType = &Gdiplus::ImageFormatPNG;
            }
        }
```
Now is OK.

base/applications/mspaint/winproc.cpp
```diff
@@ -584,6 +584,9 @@ ULONG CMainWindow::OpenMailerWithAttachmentW(LPCWSTR pszPathName)
     CStringW strFileTitle;
     if (PathFileExistsW(pszPathName)) // File does exist?
     {
+        if (!ConfirmSave())
```
HBelusca, line 588: I've checked with Windows' own MSPaint, and when the file exists, but has been modified and not yet saved, and one tries to send it by mail, there is *_NO_* prompt for save confirmation, *but* the file with modifications is nevertheless attached: a temporary file is created.

katahiromz: Always temp file?

HBelusca: According to observations, it appears it's always, yes, **UNLESS** the file hasn't been modified (or has been manually saved just prior to sending it).

katahiromz: Now FileType is fixed.

katahiromz: Now file title is fixed.

base/applications/mspaint/atlimagedx.h
```diff
@@ -98,10 +100,15 @@ class CImageDx : public CImage
...
             FileType = FileTypeFromExtension(pszExt);
+            if (!FileType)
+            {
+                FileType = &Gdiplus::ImageFormatPNG;
+            }
         }
 
         // get CLSID from file type
         CLSID clsid;
+
```
HBelusca, line 111:
```suggestion
```

base/applications/mspaint/atlimagedx.h
```diff
@@ -17,6 +17,8 @@ class CImageDx : public CImage
         BitmapSetResolution = NULL;
     }
 
+    using CImage::FileTypeFromExtension;
```
HBelusca, line 20: Why is this needed?

base/applications/mspaint/winproc.cpp
```diff
@@ -580,6 +582,120 @@ LRESULT CMainWindow::OnSysColorChange(UINT nMsg, WPARAM wParam, LPARAM lParam, B
...
+typedef ULONG (WINAPI *FN_MAPISendMailW)(LHANDLE, ULONG_PTR, lpMapiMessageW, FLAGS, ULONG);
+
+ULONG CMainWindow::OpenMailerWithAttachmentW(LPCWSTR pszPathName)
+{
+    CStringW strFileTitle;
+    if (PathFileExistsW(pszPathName) && imageModel.IsImageSaved())
+    {
+        strFileTitle = PathFindFileNameW(pszPathName);
+    }
+    else // Not existing
```
HBelusca, line 595:
```suggestion
    else // Not existing or not saved
```

base/applications/mspaint/winproc.cpp
```diff
@@ -580,6 +582,120 @@ LRESULT CMainWindow::OnSysColorChange(UINT nMsg, WPARAM wParam, LPARAM lParam, B
...
+        ::GetTempPathW(_countof(szTempDir), szTempDir);
+        if (!::GetTempFileNameW(szTempDir, L"afx", 0, g_szTempFile))
+            return MAPI_E_FAILURE;
+
+        // Get the file type if already saved
+        const GUID *FileType = &GUID_NULL;
+        if (imageModel.IsImageSaved())
+        {
+            CImageDx img;
+            FileType = img.FileTypeFromExtension(PathFindExtensionW(g_szFileName));
```
HBelusca, line 615: It is weird to create a CImageDx object just to do some parsing with the function.
It would be better if, for the CImageDx class, the FileTypeFromExtension was some kind of static function, so that one could just do instead: `CImageDx::FileTypeFromExtension(etc.);`

base/applications/mspaint/winproc.cpp
```diff
@@ -580,6 +582,120 @@ LRESULT CMainWindow::OnSysColorChange(UINT nMsg, WPARAM wParam, LPARAM lParam, B
...
+        pszPathName = g_szTempFile;
+
+        // Set the file title
+        if (imageModel.IsImageSaved())
+        {
+            strFileTitle = PathFindFileNameW(g_szFileName);
+        }
+        else
+        {
+            strFileTitle.LoadString(IDS_DEFAULTFILENAME);
```
HBelusca, line 639: iirc when the file *exists* but may not be saved (again) yet, it's the actual name of the file that is used for the attachment. You can test yourself all that with standard Windows' MSPaint, by the way.

base/applications/mspaint/atlimagedx.h
katahiromz: That is a `protected` member.

base/applications/mspaint/winproc.cpp
katahiromz: Fixed.

base/applications/mspaint/winproc.cpp
katahiromz: Fixed.

base/applications/mspaint/atlimagedx.h
HBelusca: `friend` keyword wouldn't work here?

katahiromz: Doing that will expose a design flaw.

katahiromz:
```c
#ifdef _MSPAINT_H
    friend ULONG CMainWindow::OpenMailerWithAttachmentW(LPCWSTR pszPathName);
#endif
```
This code won't work because `CMainWindow` is not declared.

katahiromz:
```c
#ifdef _MSPAINT_H
    public:
#else
    protected:
#endif
```
This will work.

katahiromz: `using CImage::FileTypeFromExtension;`

This is simple and the best.

HBelusca: > ```c
> #ifdef _MSPAINT_H
>     public:
> #else
>     protected:
> #endif
> ```
> 
> This will work.

But this is inelegant. atlimage.h should stay MS-compatible, not have hacks for internal applications.

katahiromz: https://jira.reactos.org/browse/CORE-19008

katahiromz:
Time up. I will retry with smaller changes...

PR #5358, opened 2023-06-20
[MSPAINT] Fix ShapeTool with Shift key

katahiromz:
## Purpose
Fix the behavior of the `ShapeTool` in pressing `Shift` key.

## Proposed changes

- Correctly call `roundTo8Directions` helper function on mouse actions.

## TODO

- [x] Do tests.


katahiromz:
> Can you quickly remind us what the Shift key does when the tool is selected?

On `ShapeTool` (polygon), the direction of the drawing line is restricted in 8 directions.

PR #5359, opened 2023-06-21
[HAL][MINIHAL] Minor cleanup in reboot.c files.

HBelusca:
And remove pic/processor.c from minihal compilation.


HBelusca:
Cc @disean as well for review.

hal/halx86/xbox/reboot.c
```diff
@@ -76,22 +75,31 @@ HalpXboxPowerAction(IN UCHAR Action)
...
-HalReturnToFirmware(IN FIRMWARE_REENTRY Action)
+HalReturnToFirmware(
+    _In_ FIRMWARE_REENTRY Action)
 {
     /* Check what kind of action this is */
     switch (Action)
     {
-        /* All recognized actions */
+        /* All recognized actions: call the internal power function */
+        case HalHaltRoutine:
```
binarymaster, line 100: I wonder if this addition would make ReactOS turn Xbox off properly - currently it halts on the "it's safe to shutdown" screen, like on a non-ACPI machine.

hal/halx86/xbox/reboot.c
```diff
@@ -63,10 +62,10 @@ SMBusWriteByte(UCHAR Address, UCHAR Register, UCHAR Data)
     }
 }
 
-VOID
 DECLSPEC_NORETURN
```
disean, line 65:
```suggestion
static
DECLSPEC_NORETURN
```

hal/halarm/generic/reboot.c
```diff
@@ -9,35 +9,41 @@
 /* INCLUDES ******************************************************************/
 
 #include <hal.h>
-#define NDEBUG
-#include <debug.h>
```
JoachimHenze, line 13: Question, no request for change: Here you do remove the debug include. I do like that. But then later the source calls stuff like DbgPrint(). Isn't that a problem? Or where does DbgPrint() come from? Thx for explanation in advance.

HBelusca: DbgPrint is standard NT function -> it's in the usual WDK/DDK/XDK/NDK headers. reactos' own debug.h is for the ROS-specific debug wrapper macros DPRINT(1) and ERR/WARN/TRACE .

hal/halx86/xbox/reboot.c
HBelusca: I don't think so, because we don't have any explicit `HalReturnToFirmware(HalHaltRoutine);` call in our code base currently.

hal/halx86/xbox/reboot.c
```diff
@@ -63,35 +62,47 @@ SMBusWriteByte(UCHAR Address, UCHAR Register, UCHAR Data)
...
+static DECLSPEC_NORETURN
 VOID
-DECLSPEC_NORETURN
-NTAPI
-HalpXboxPowerAction(IN UCHAR Action)
+HalpXboxPowerAction(
+    _In_ UCHAR Action)
 {
+    /* Disable interrupts */
+    _disable();
```
HBelusca, line 71: @binarymaster @disean I've added this explicit interrupt disable. Please tell me whether this is OK or not, for XBOX.

binarymaster: I think this should be fine.

binarymaster (approved):

PR #5375, opened 2023-06-29
[NTOS:PNP] Remove firmware mapper

Extravert-ir:
Remove the ARC tree parsing code from the kernel, and leave it legacy x86 HAL (it was enabled only there anyway).
Those legacy devices from the ARC tree will now be attached to HAL bus, instead of root one.

Should fix [CORE-17463](https://jira.reactos.org/browse/CORE-17463)


binarymaster:
From MSVC 14.0 i386:
```
D:\a\reactos\reactos\src\hal\halx86\legacy\halpnpdd.c(538): error C2983: 'HaliInitPnpDriver': all declarations must have an identical __declspec(code_seg(...))
d:\a\reactos\reactos\src\hal\halx86\include\halp.h(504): note: see declaration of 'HaliInitPnpDriver'
```

hal/halx86/legacy/arcmap.c
```diff
@@ -44,57 +50,65 @@ DATA_SEG("INITDATA")
...
-    { L"SerialController", L"*PNP0501\0", NULL, 0 },
+    { L"SerialController", L"PNP0501", NULL, 0 },
     //{ L"KeyboardController", L"*PNP0303\0", NULL, 0 },
     //{ L"PointerController", L"*PNP0F13\0", NULL, 0 },
     { L"KeyboardPeripheral", NULL, KeyboardMap, 0 },
     { L"PointerPeripheral", NULL, PointerMap, 0 },
-    { L"ParallelController", L"*PNP0400\0", NULL, 0 },
-    { L"FloppyDiskPeripheral", L"*PNP0700\0", NULL, 0 },
+    { L"ParallelController", L"PNP0400", NULL, 0 },
+    { L"FloppyDiskPeripheral", L"PNP0700", NULL, 0 },
```
HBelusca, line 98: (generic comment for all these lists above): Wasn"t there a discussion somewhere (iirc either by Stas'm or disean) who mentioned that these 2nd strings could be MULTI_SZ ?

binarymaster: I think he modified the underlying code already, so this is not needed anymore.

Extravert-ir: Yes, the MULTI_SZ is generated in `HalpPdoQueryId`

binarymaster (commented):
I don't have a strong preference on how legacy devices should be attached to the PnP manager - this way it would look more logical from the hardware standpoint, as siblings of the computer/HAL and not the PnP root. However while this fixes PS/2 input with Standard HAL at the 3rd stage, this also weakens cross-compatiblity with Windows 2003 legacy HALs, which I'd like to preserve and improve. I say 'weakens' here and not 'breaks' because we have not reached that kind of compatiblity yet. 🙂 I'd like to hear opinions on this from @disean and @r3mendozal.

Extravert-ir:
> this also weakens cross-compatiblity with Windows 2003 legacy HALs, which I'd like to preserve and improve

I have said numerous times, **there will be no compatibility with Windows 2003 HALs**. If you want a new HAL for Windows 2003, start a new project.

hal/halx86/legacy/arcmap.c
```diff
@@ -28,13 +35,12 @@ typedef struct _PNP_DETECT_IDENTIFIER_MAP
...
 
+extern LIST_ENTRY HalpPdoListHead;
+
 /* DATA **********************************************************************/
 
 static UNICODE_STRING IdentifierU = RTL_CONSTANT_STRING(L"Identifier");
-static UNICODE_STRING HardwareIDU = RTL_CONSTANT_STRING(L"HardwareID");
 static UNICODE_STRING ConfigurationDataU = RTL_CONSTANT_STRING(L"Configuration Data");
-static UNICODE_STRING BootConfigU = RTL_CONSTANT_STRING(L"BootConfig");
-static UNICODE_STRING LogConfU = RTL_CONSTANT_STRING(L"LogConf");
```
HBelusca, line 37: Which module is going to deal with those values now?

HBelusca:
> **there will be no compatibility with Windows 2003 HALs**

To be honest this has been a unipersonal decision from you alone without asking the other kernel core developers about what they really wanted...

hal/halx86/legacy/arcmap.c
Extravert-ir: These registry keys are written by the PnP manager

HBelusca:
> [...] I say 'weakens' here and not 'breaks' because we have not reached that kind of compatiblity yet. slightly_smiling_face I'd like to hear opinions on this from @disean and @r3mendozal.

There was also a discussion on whether this would break support for stuff like PnP ISA; SoundBlaster drivers etc.

binarymaster:
> If you want a new HAL for Windows 2003, start a new project.

No thanks, this is one of reasons for ReactOS to exist.

> I have said numerous times, **there will be no compatibility with Windows 2003 HALs**.

While I can accept this for the ACPI HALs specifically, given the reasons you have explained about the complexity of acpi.sys and halacpi interaction, this is unacceptable for the legacy HALs. I can only confirm Hermes' words above regarding "unipersonal decision from you alone".

disean:
I'm not sure about this patch but I think it will make it more difficult to update the ESCD storage area (not currently implemented yet) and may discourage the HAL to be replaced by user (this replacement has never worked right before).

hal/halx86/legacy/halpnpdd.c
```diff
@@ -1,931 +1,544 @@
...
     EXTENSION_TYPE ExtensionType;
-    struct _PDO_EXTENSION* Next;
-    PDEVICE_OBJECT PhysicalDeviceObject;
-    PFDO_EXTENSION ParentFdoExtension;
-    PDO_TYPE PdoType;
-    PDESCRIPTION_HEADER WdTable;
-    LONG InterfaceReferenceCount;
+    PDEVICE_OBJECT DeviceObject;
+    SINGLE_LIST_ENTRY ListEntry;
+    WCHAR PnPID[8];
```
disean, line 31:
```suggestion
    WCHAR PnPID[7 + 1];
```

hal/halx86/legacy/halpnpdd.c
```diff
@@ -1,931 +1,544 @@
...
-    else
-    {
-        /* The only other thing we support is a target relation for the PDO */
-        if ((RelationType == TargetDeviceRelation) &&
-            (ExtensionType == PdoExtensionType))
+        PFDO_EXTENSION fdoExtension = DeviceObject->DeviceExtension;
+        PDEVICE_RELATIONS fdoRelations = ExAllocatePoolWithTag(
+            PagedPool,
+            FIELD_OFFSET(DEVICE_RELATIONS, Objects)
+                + (sizeof(PDEVICE_OBJECT) * fdoExtension->PdoCount),
```
disean, line 63: Maybe something like this might be more suitable:
```suggestion
            FIELD_OFFSET(DEVICE_RELATIONS, Objects[fdoExtension->PdoCount]),
```

hal/halx86/legacy/halpnpdd.c
```diff
@@ -1,931 +1,544 @@
...
-        return STATUS_NOT_SUPPORTED;
-    }
-    else
-    {
-        /* This shouldn't happen */
-        return STATUS_UNSUCCESSFUL;
-    }
-}
+    PPDO_EXTENSION pdoExtension = DeviceObject->DeviceExtension;
+    WCHAR id[30];
```
disean, line 192: Why `30`?
```suggestion
    WCHAR id[sizeof("PCI_HAL\\PNP1234\0*PNP1234\0")];
```

hal/halx86/legacy/halpnpdd.c
```diff
@@ -1,931 +1,544 @@
...
+CODE_SEG("PAGE")
 NTSTATUS
 NTAPI
-HalpDispatchWmi(IN PDEVICE_OBJECT DeviceObject,
-                IN PIRP Irp)
+HalpAddDevice(
+    _In_ PDRIVER_OBJECT DriverObject,
+    _In_ PDEVICE_OBJECT TargetDevice)
 {
-    UNIMPLEMENTED_DBGBREAK("HAL: PnP Driver WMI!\n");
```
disean, line 840: PAGED_CODE() is missing here.

hal/halx86/legacy/halpnpdd.c
```diff
@@ -1,931 +1,544 @@
...
+    // Must not be called
+    ASSERT(FALSE);
     return STATUS_SUCCESS;
 }
 
+CODE_SEG("PAGE")
```
disean, line 357: This section specifier doesn't seem correct.
```suggestion
```

hal/halx86/legacy/halpnpdd.c
```diff
@@ -1,931 +1,544 @@
...
+    _In_ PCM_RESOURCE_LIST BootResources,
+    _In_ UINT32 BootResourcesLength,
+    _In_ PDEVICE_OBJECT Fdo)
+{
+    PFDO_EXTENSION fdoExtension = Fdo->DeviceExtension;
+    PDEVICE_OBJECT devObject;    
+
+    NTSTATUS status = IoCreateDevice(
+        HalpDriverObject,
+        sizeof(PDO_EXTENSION),
```
disean, line 396: You could declare the `pdoExtension` variable first
```suggestion
        sizeof(*pdoExtension),
```

hal/halx86/legacy/halpnpdd.c
```diff
@@ -1,931 +1,544 @@
...
-    /* This is us */
     HalpDriverObject = DriverObject;
 
-    /* Set up add device */
     DriverObject->DriverExtension->AddDevice = HalpAddDevice;
-
-    /* Set up the callouts */
     DriverObject->MajorFunction[IRP_MJ_PNP] = HalpDispatchPnp;
     DriverObject->MajorFunction[IRP_MJ_POWER] = HalpDispatchPower;
-    DriverObject->MajorFunction[IRP_MJ_SYSTEM_CONTROL] = HalpDispatchWmi;
```
disean, line 887: This handler is not needed anymore?

hal/halx86/legacy/halpnpdd.c
```diff
@@ -1,931 +1,544 @@
...
+    pdoExtension->BootResourcesLength = BootResourcesLength;
+    PushEntryList(&fdoExtension->PdoListHead, &pdoExtension->ListEntry);
+    fdoExtension->PdoCount++;
+
+    devObject->Flags &= ~DO_DEVICE_INITIALIZING;
+
+    DPRINT("HAL: Created a PDO %.7S\n", pdoExtension->PnPID);
+
+    return status;
 }
```
disean, line 424: Perhaps the `extern` keyword is missing here.

hal/halx86/legacy/halpnpdd.c
HBelusca: With MSVC sometimes this syntax works, and sometimes it doesn't (but I don't remember exactly in which condition).

Extravert-ir:
> it more difficult to update the ESCD storage area (not currently implemented yet)

How so?

> and may discourage the HAL to be replaced by user (this replacement has never worked right before)

And it will never work right, so it will not work at all :)

PR #5377, opened 2023-06-29
[NETSTAT] -b flag implies -o flag on Windows XP/2003. CORE-19006

JoachimHenze:
When calling 'netstat -abn'
Win 2k3sp2 and XPSP3 do show both: the processes name and the PID.
Contrary Win 7 and Win 8.1 would show only the process name then without the PID.

The newer Windows versions would require you to explicitly pass -o if you want to see the PID also.

We do follow 2k3sp2 because it is our target, but also because that is a more sane default choice.
The process name is not of much use without having the PID as well,
especially if multiple processes with the same name
do run on a system, e.g.: multiple 'svchost.exe' processes.

JIRA issue: [CORE-19006](https://jira.reactos.org/browse/CORE-19006)

Attached is a picture of the fixed state:
PID_shown

### DIFF
--- a/base/applications/network/netstat/netstat.c
+++ b/base/applications/network/netstat/netstat.c
@@ -98,10 +98,9 @@ BOOL ParseCmdline(int argc, wchar_t* argv[])
                         // UNIMPLEMENTED.
                         ConPuts(StdErr, L"'b' option is FIXME (Accepted option though unimplemented feature).\n");
                         bDoShowProcName = TRUE;
-
-                        // 2k3sp2 and XPSP3: -b does imply -o (superior)
-                        // Win7 and 8.1: -b and -o were decoupled (inferior)
+#if _WIN32_WINNT < 0x600
                         bDoShowProcessId = TRUE;
+#endif
                         break;
                     case L'e':
                         bDoShowEthStats = TRUE;
--- a/base/applications/network/netstat/netstat.c
+++ b/base/applications/network/netstat/netstat.c
@@ -98,7 +98,7 @@ BOOL ParseCmdline(int argc, wchar_t* argv[])
                         // UNIMPLEMENTED.
                         ConPuts(StdErr, L"'b' option is FIXME (Accepted option though unimplemented feature).\n");
                         bDoShowProcName = TRUE;
-#if _WIN32_WINNT < 0x600
+#if (_WIN32_WINNT < _WIN32_WINNT_VISTA || TRUE) /* prefer the old behavior */
                         bDoShowProcessId = TRUE;
 #endif
                         break;
--- a/base/applications/network/netstat/netstat.c
+++ b/base/applications/network/netstat/netstat.c
@@ -98,6 +98,10 @@ BOOL ParseCmdline(int argc, wchar_t* argv[])
                         // UNIMPLEMENTED.
                         ConPuts(StdErr, L"'b' option is FIXME (Accepted option though unimplemented feature).\n");
                         bDoShowProcName = TRUE;
+
+                        // 2k3sp2 and XPSP3: -b does imply -o (superior)
+                        // Win7 and 8.1: -b and -o were decoupled (inferior)
+                        bDoShowProcessId = TRUE;
                         break;
                     case L'e':
                         bDoShowEthStats = TRUE;
--- a/base/applications/network/netstat/netstat.c
+++ b/base/applications/network/netstat/netstat.c
@@ -98,7 +98,7 @@ BOOL ParseCmdline(int argc, wchar_t* argv[])
                         // UNIMPLEMENTED.
                         ConPuts(StdErr, L"'b' option is FIXME (Accepted option though unimplemented feature).\n");
                         bDoShowProcName = TRUE;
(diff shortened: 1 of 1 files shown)


base/applications/network/netstat/netstat.c
```diff
@@ -98,6 +98,10 @@ BOOL ParseCmdline(int argc, wchar_t* argv[])
                         // UNIMPLEMENTED.
                         ConPuts(StdErr, L"'b' option is FIXME (Accepted option though unimplemented feature).\n");
                         bDoShowProcName = TRUE;
+
+                        // 2k3sp2 and XPSP3: -b does imply -o (superior)
+                        // Win7 and 8.1: -b and -o were decoupled (inferior)
```
HBelusca, line 103: Not sure having "superior" or "inferior" helps in anything here.
You could instead say that we follow what netstat does on Windows <= 2003, and that it is more logical to have -b show the processes IDs, since -b itself shows the process image that created the connection.

(Also, comment style for non-temporary/FIXMEs/TODOs here is with `/* ... */`).

SergeGautherie: Would some `#if _WIN32_WINNT` or the likes be applicable here?

binarymaster: Yes, using `#if _WIN32_WINNT` macros would be preferable.

JoachimHenze: To be honest, I don't like the fact that the ifdef will automatically switch back to the inferior state when ros will update that define in several years to target more recent versions of Windows. But if that is the preference of the majority of devs then I can do that.

JoachimHenze: Fixed heavy-heart-edly.

binarymaster (approved):

base/applications/network/netstat/netstat.c
```diff
@@ -98,6 +98,9 @@ BOOL ParseCmdline(int argc, wchar_t* argv[])
                         // UNIMPLEMENTED.
                         ConPuts(StdErr, L"'b' option is FIXME (Accepted option though unimplemented feature).\n");
                         bDoShowProcName = TRUE;
+#if _WIN32_WINNT < 0x600
```
SergeGautherie, line 101:
```suggestion
#if _WIN32_WINNT < _WIN32_WINNT_VISTA
```

HBelusca: `#if (_WIN32_WINNT < _WIN32_WINNT_VISTA)`
even though I would prefer, moe like Joachim, that this behaviour stays even when we compile ROS in Vista+ mode.

binarymaster:
From Joachim:
> The process name is not of much use without having the PID as well,
especially if multiple processes with the same name

From Hermes:
> even though I would prefer, more like Joachim, that this behaviour stays even when we compile ROS in Vista+ mode.

Guys, are you really that lazy to specify one more flag, i.e. `-ob` ?

On Vista+ netstat is just outputting what the user asked it to, not more, not less. XP/2003 behavior is looking more like a bug to me. 🐞

base/applications/network/netstat/netstat.c
JoachimHenze: Applied

JoachimHenze:
> Guys, are you really that lazy to specify one more flag, i.e. `-ob` ?

Yes, we are.

base/applications/network/netstat/netstat.c
```diff
@@ -98,7 +98,7 @@ BOOL ParseCmdline(int argc, wchar_t* argv[])
                         // UNIMPLEMENTED.
                         ConPuts(StdErr, L"'b' option is FIXME (Accepted option though unimplemented feature).\n");
                         bDoShowProcName = TRUE;
-#if _WIN32_WINNT < 0x600
+#if (_WIN32_WINNT < _WIN32_WINNT_VISTA || TRUE) /* prefer the old behavior */
```
binarymaster, line 101: This will break compatibility with scripts, once we bump the target.

JoachimHenze: Link to some real-world-example or "it didn't happen"(TM). Currently 2 vs 1 votes for keeping the current state.

binarymaster: If you want to use voting, better create a poll on Mattermost. Very few people look into PRs.

JoachimHenze: Not needed to split the voting from the PR, we can vote at Hermes post https://github.com/reactos/reactos/pull/5377#discussion_r1248249573

learn-more: I'd prefer the newer behaviour as well.

JoachimHenze: Ok, I do summarize the votings as follows:
People who think we should stick with the 2k3sp2 beahvior even after version bump:
JoachimHenze, hbelusca

People who think we should automatically toggle to Win7 behavior after version bump:
binarymaster, learn-more, Darkfire, GeoB99

2nd fraction won democratically and I will remove the "|| TRUE" and the comment.

binarymaster (approved):

learn-more (approved):

HBelusca (approved):

PR #5378, opened 2023-06-30
[NTOS:KE] Report x86 CPU features in a separate function

binarymaster:
Print out the CPU features only when Kernel Debugger has been initialized and `KF_*` features and policies has been applied.

Also add this feature to AMD64.

JIRA issue: [CORE-18023](https://jira.reactos.org/browse/CORE-18023)

### DIFF
--- a/ntoskrnl/include/internal/amd64/ke.h
+++ b/ntoskrnl/include/internal/amd64/ke.h
@@ -420,6 +420,9 @@ VOID KiGetCacheInformation(VOID);
 VOID KiSetProcessorType(VOID);
 ULONG KiGetFeatureBits(VOID);
 VOID KiInitializeCpuFeatures(VOID);
+#if DBG
+VOID KiReportCpuFeatures(IN PKPRCB Prcb);
+#endif
 
 ULONG KeAllocateGdtSelector(ULONG Desc[2]);
 VOID KeFreeGdtSelector(ULONG Entry);
--- a/ntoskrnl/include/internal/i386/ke.h
+++ b/ntoskrnl/include/internal/i386/ke.h
@@ -459,6 +459,12 @@ ULONG
 NTAPI
 KiGetFeatureBits(VOID);
 
+#if DBG
+CODE_SEG("INIT")
+VOID
+KiReportCpuFeatures(VOID);
+#endif
+
 VOID
 NTAPI
 KiThreadStartup(VOID);
--- a/ntoskrnl/include/internal/i386/ke.h
+++ b/ntoskrnl/include/internal/i386/ke.h
@@ -22,6 +22,30 @@ extern "C"
 #define KD_BREAKPOINT_SIZE        sizeof(UCHAR)
 #define KD_BREAKPOINT_VALUE       0xCC
 
+/* CPUID 1 - EDX flags */
+#define X86_FEATURE_FPU         0x00000001 /* x87 FPU is present */
+#define X86_FEATURE_VME         0x00000002 /* Virtual 8086 Extensions are present */
+#define X86_FEATURE_DBG         0x00000004 /* Debugging extensions are present */
+#define X86_FEATURE_PSE         0x00000008 /* Page Size Extension is present */
+#define X86_FEATURE_TSC         0x00000010 /* Time Stamp Counters are present */
+#define X86_FEATURE_PAE         0x00000040 /* Physical Address Extension is present */
+#define X86_FEATURE_CX8         0x00000100 /* CMPXCHG8B instruction present */
+#define X86_FEATURE_APIC        0x00000200 /* APIC is present */
+#define X86_FEATURE_SYSCALL     0x00000800 /* SYSCALL/SYSRET support present */
+#define X86_FEATURE_MTTR        0x00001000 /* Memory type range registers are present */
+#define X86_FEATURE_PGE         0x00002000 /* Page Global Enable */
+#define X86_FEATURE_CMOV        0x00008000 /* "Conditional move" instruction supported */
+#define X86_FEATURE_PAT         0x00010000 /* Page Attribute Table is supported */
+#define X86_FEATURE_DS          0x00200000 /* Debug Store is present */
+#define X86_FEATURE_MMX         0x00800000 /* MMX extension present */
+#define X86_FEATURE_FXSR        0x01000000 /* FXSAVE/FXRSTOR instructions present */
+#define X86_FEATURE_SSE         0x02000000 /* SSE extension present */
+#define X86_FEATURE_SSE2        0x04000000 /* SSE2 extension present */
+#define X86_FEATURE_HT          0x10000000 /* Hyper-Threading present */
+
+/* CPUID 0x80000001 - EDX extended flags */
+#define X86_FEATURE_NX          0x00100000 /* NX support present */
+
 //
 // One-liners for getting and setting special purpose registers in portable code
 //
--- a/ntoskrnl/ke/amd64/cpu.c
+++ b/ntoskrnl/ke/amd64/cpu.c
@@ -234,6 +234,60 @@ KiGetFeatureBits(VOID)
     return FeatureBits;
 }
 
+#if DBG
+VOID
+KiReportCpuFeatures(IN PKPRCB Prcb)
+{
+    ULONG CpuFeatures = 0;
+    CPU_INFO CpuInfo;
+
+    if (Prcb->CpuVendor)
+    {
+        KiCpuId(&CpuInfo, 1);
+        CpuFeatures = CpuInfo.Edx;
+    }
+
+    DPRINT1("Supported CPU features: ");
+
+#define print_kf_bit(kf_value) if (Prcb->FeatureBits & kf_value) DbgPrint(#kf_value " ")
+    print_kf_bit(KF_V86_VIS);
+    print_kf_bit(KF_RDTSC);
+    print_kf_bit(KF_CR4);
+    print_kf_bit(KF_CMOV);
+    print_kf_bit(KF_GLOBAL_PAGE);
+    print_kf_bit(KF_LARGE_PAGE);
+    print_kf_bit(KF_MTRR);
+    print_kf_bit(KF_CMPXCHG8B);
+    print_kf_bit(KF_CMPXCHG16B);
+    print_kf_bit(KF_MMX);
+    print_kf_bit(KF_WORKING_PTE);
+    print_kf_bit(KF_PAT);
+    print_kf_bit(KF_FXSR);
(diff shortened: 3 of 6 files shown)


ntoskrnl/include/internal/amd64/ke.h
```diff
@@ -420,6 +420,9 @@ VOID KiGetCacheInformation(VOID);
 VOID KiSetProcessorType(VOID);
 ULONG KiGetFeatureBits(VOID);
 VOID KiInitializeCpuFeatures(VOID);
+#if DBG
+VOID KiReportCpuFeatures(IN PKPRCB Prcb);
```
SergeGautherie, line 424:
```suggestion
VOID NTAPI KiReportCpuFeatures(_In_ PKPRCB Prcb);
```

ntoskrnl/ke/i386/cpu.c
```diff
@@ -430,35 +425,61 @@ KiGetFeatureBits(VOID)
...
+    ULONG CpuFeatures = 0;
+    CPU_INFO CpuInfo;
+
+    if (KiGetCpuVendor())
+    {
+        KiCpuId(&CpuInfo, 1);
+        CpuFeatures = CpuInfo.Edx;
+    }
+
+    DPRINT1("Supported CPU features: ");
```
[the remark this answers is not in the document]
binarymaster: I keep `DPRINT1` here to keep a print of the source file path and line number.

ntoskrnl/include/internal/amd64/ke.h
binarymaster: AMD64 functions in `cpu.c` do not have that, so I'm following existing conventions.

binarymaster: Just following other function declarations in `ke.h` here.

ntoskrnl/include/internal/i386/ke.h
```diff
@@ -22,6 +22,30 @@ extern "C"
 #define KD_BREAKPOINT_SIZE        sizeof(UCHAR)
 #define KD_BREAKPOINT_VALUE       0xCC
 
+/* CPUID 1 - EDX flags */
+#define X86_FEATURE_FPU         0x00000001 /* x87 FPU is present */
```
Extravert-ir, line 26: These defines are usually separated by the register (eax/ebx/ecx/edx) and the leaf. So this would be something like `X86_CPUID_STD_EDX_FPU`. And NX bit below would be `X86_CPUID_EXT_EDX_NX`

binarymaster: Yes, this would be more clear, however I decided to follow how Timo did it for AMD64: https://git.reactos.org/?p=reactos.git;a=blob;f=ntoskrnl/include/internal/amd64/ke.h;hb=40864bc15cb77ceb07e73f8625486c120192bb18#l29

JoachimHenze (approved):
Glad to see it back. It was one of the most important loggings that we have.

ntoskrnl/ke/i386/cpu.c
```diff
@@ -416,7 +416,7 @@ KiGetFeatureBits(VOID)
                 KiCpuId(&CpuInfo, 0x80000001);
 
                 /* Check if NX-bit is supported */
-                if (CpuInfo.Edx & 0x00100000) FeatureBits |= KF_NX_BIT;
+                if (CpuInfo.Edx & X86_FEATURE_NX) FeatureBits |= KF_NX_BIT;
```
HBelusca, line 414: Maybe worth putting the FeatureBits setting in its own line.

ntoskrnl/include/internal/i386/ke.h
```diff
@@ -22,6 +22,30 @@ extern "C"
...
+#define X86_FEATURE_MTTR        0x00001000 /* Memory type range registers are present */
+#define X86_FEATURE_PGE         0x00002000 /* Page Global Enable */
+#define X86_FEATURE_CMOV        0x00008000 /* "Conditional move" instruction supported */
+#define X86_FEATURE_PAT         0x00010000 /* Page Attribute Table is supported */
+#define X86_FEATURE_DS          0x00200000 /* Debug Store is present */
+#define X86_FEATURE_MMX         0x00800000 /* MMX extension present */
+#define X86_FEATURE_FXSR        0x01000000 /* FXSAVE/FXRSTOR instructions present */
+#define X86_FEATURE_SSE         0x02000000 /* SSE extension present */
+#define X86_FEATURE_SSE2        0x04000000 /* SSE2 extension present */
+#define X86_FEATURE_HT          0x10000000 /* Hyper-Threading present */
```
HBelusca, line 44: Aren't there some of these flags the same as in AMD64? I was wondering whether it could make sense to have them in the same file if so :thinking:

ntoskrnl/ke/i386/cpu.c
```diff
@@ -430,35 +425,61 @@ KiGetFeatureBits(VOID)
...
-#undef print_supported
-
     /* Return the Feature Bits */
     return FeatureBits;
 }
 
+#if DBG
+CODE_SEG("INIT")
+VOID
+NTAPI
```
HBelusca, line 435: If the function is internal and not a callback, the NTAPI is just superfluous.

ntoskrnl/ke/i386/kiinit.c
```diff
@@ -480,6 +480,12 @@ KiInitializeKernel(IN PKPROCESS InitProcess,
     /* Save CPU state */
     KiSaveProcessorControlState(&Prcb->ProcessorState);
 
+#if DBG
+    /* Print applied kernel features/policies and boot CPU features */
```
HBelusca, line 484: It would be interesting to show the info also for each other CPU present in the system (since in SMP we can have CPUs with slightly different features, but not too different or incompatible though), whenever those are attempted to be started.

ntoskrnl/ke/i386/cpu.c
```diff
@@ -430,35 +425,61 @@ KiGetFeatureBits(VOID)
...
+
+    if (KiGetCpuVendor())
+    {
+        KiCpuId(&CpuInfo, 1);
+        CpuFeatures = CpuInfo.Edx;
+    }
+
+    DPRINT1("Supported CPU features: ");
+
+#define print_kf_bit(kf_value) if (KeFeatureBits & kf_value) DbgPrint(#kf_value " ")
```
HBelusca, line 449: Why don't you use the old type of `DPRINT1("Supported CPU features : %s %s %s %s %s %s %s %s %s %s %s %s %s %s %s %s %s %s %s %s %s\n",` ? Is it to remove visible extra spaces that could show up if the CPU doesn't have some of the bits set?

(Alternatively you could use a format string like `%s%s%s%s` etc. without intermediate spaces, and in the printf_kf_bit macro, use `#kf_value " "` ?)

ntoskrnl/ke/i386/cpu.c
```diff
@@ -430,35 +425,61 @@ KiGetFeatureBits(VOID)
...
+NTAPI
+KiReportCpuFeatures(VOID)
+{
+    ULONG CpuFeatures = 0;
+    CPU_INFO CpuInfo;
+
+    if (KiGetCpuVendor())
+    {
+        KiCpuId(&CpuInfo, 1);
+        CpuFeatures = CpuInfo.Edx;
```
HBelusca, line 444: We don't cache those in some "Ke" variable, contrary to the KeFeatureBits ?

ntoskrnl/ke/amd64/cpu.c
```diff
@@ -234,6 +234,61 @@ KiGetFeatureBits(VOID)
     return FeatureBits;
 }
 
+#if DBG
+VOID
+NTAPI
```
HBelusca, line 239: As said in one previous comment, if function is internal and not a callback, NTAPI is superfluous.

HBelusca (commented):
Some things to think about :)

ntoskrnl/include/internal/i386/ke.h
binarymaster: Yes, I copied most of them from AMD64 `ke.h`, but also I had to add APIC one here, because on AMD64 the APIC is always assumed present.

ntoskrnl/ke/i386/cpu.c
binarymaster: Following existing style here, also the same Timo did on AMD64.

ntoskrnl/ke/i386/cpu.c
binarymaster: Looking at `KiGetFeatureBits()`... no we currently don't.

ntoskrnl/ke/i386/cpu.c
binarymaster: Yes, to remove spaces, and also this construction doesn't look flexible enough to me (rather looks hacky).

ntoskrnl/ke/i386/kiinit.c
binarymaster: Indeed it would be interesting feature, hope we'll get SMP working one day. 🙂

binarymaster (commented):
Commented, and resolved review comments that has been addressed.

ntoskrnl/ke/amd64/cpu.c
```diff
@@ -234,6 +234,60 @@ KiGetFeatureBits(VOID)
...
+KiReportCpuFeatures(IN PKPRCB Prcb)
+{
+    ULONG CpuFeatures = 0;
+    CPU_INFO CpuInfo;
+
+    if (Prcb->CpuVendor)
+    {
+        KiCpuId(&CpuInfo, 1);
+        CpuFeatures = CpuInfo.Edx;
+    }
```
GeoB99, line 248: There is `KiGetCpuVendor` in AMD64 which already returns you the CPU vendor so you do not have to pass a PCB here. So why not use that?

binarymaster: On AMD64 `Prcb->CpuVendor` is being populated by `KiGetCpuVendor()` which is already has been called in phase 0 even before the Kernel Debugger initialization. Why we should call it when the value is already evaluated and cached in Prcb?

ntoskrnl/ke/i386/cpu.c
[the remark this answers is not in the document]
SergeGautherie: Right.

ntoskrnl/include/internal/amd64/ke.h
SergeGautherie: For what it's worth, `KiSetTrapContext` has `_In_`.

While here, `KiGetUserModeStackAddress` could have `s/void/VOID/`.

PR #5387, opened 2023-07-02
[SHELL32] Translated "type" in Recycle Bin

3ts75:
## Purpose

Folders and files to be translated.
And even unknown file extensions are translated.

JIRA issue: [CORE-18437](https://jira.reactos.org/browse/CORE-18437)

### DIFF
--- a/dll/win32/shell32/lang/cs-CZ.rc
+++ b/dll/win32/shell32/lang/cs-CZ.rc
@@ -925,6 +925,7 @@ BEGIN
     IDS_NONE "(None)"
 
     /* Friendly File Type Names */
+    IDS_FILE "Soubor"
     IDS_DIRECTORY "Složka"
     IDS_BAT_FILE "ReactOS dávkový soubor"
     IDS_CMD_FILE "ReactOS příkazový skript"
--- a/dll/win32/shell32/wine/shell32_main.c
+++ b/dll/win32/shell32/wine/shell32_main.c
@@ -568,28 +568,30 @@ DWORD_PTR WINAPI SHGetFileInfoW(LPCWSTR path,DWORD dwFileAttributes,
         else
         {
             if (dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY)
-                strcatW (psfi->szTypeName, L"Folder");
+                LoadStringW(shell32_hInstance, IDS_DIRECTORY, psfi->szTypeName, 80);
             else 
             {
-                WCHAR sTemp[64];
+                WCHAR sType[64];
 
-                lstrcpyW(sTemp,PathFindExtensionW(szFullPath));
-                if (sTemp[0] == 0 || (sTemp[0] == '.' && sTemp[1] == 0))
+                lstrcpyW(sType,PathFindExtensionW(szFullPath));
+                if (sType[0] == 0 || (sType[0] == '.' && sType[1] == 0))
                 {
                     /* "name" or "name." => "File" */
-                    lstrcpynW (psfi->szTypeName, L"File", 64);
+                    LoadStringW(shell32_hInstance, IDS_FILE, psfi->szTypeName, 80);
                 }
-                else if (!( HCR_MapTypeToValueW(sTemp, sTemp, 64, TRUE) &&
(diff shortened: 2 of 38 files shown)


KRosUser2 (approved):

dll/win32/shell32/wine/shell32_main.c
```diff
@@ -568,28 +568,30 @@ DWORD_PTR WINAPI SHGetFileInfoW(LPCWSTR path,DWORD dwFileAttributes,
...
-                        strcatW (psfi->szTypeName, L" file");
+                        LoadStringW(shell32_hInstance, IDS_ANY_FILE, sTemp, 64);
+                        StringCchPrintfW(psfi->szTypeName, 80, sTemp, sType);
                     }
                     else
                     {
-                        lstrcpynW (psfi->szTypeName, L"File", 64);
+                        LoadStringW(shell32_hInstance, IDS_FILE, psfi->szTypeName, 80);
                     }
                 }
```
binarymaster, line 596: This is Wine code, either use `#ifdef __REACTOS__` guard, or send changes upstream, so it could be then synced.

binarymaster (changes requested):
You cannot change Wine code directly, because it's being kept in sync with WineHQ project.

In order to solve this:
- Send changes upstream to WineHQ project
- Once the changes have been accepted by upstream, perform a Wine-sync

Alternative solution:
- Use ros-diff guards, `#ifdef __REACTOS__`, to mark the changes that we don't want to send upstream

dll/win32/shell32/lang/cs-CZ.rc
```diff
@@ -925,6 +925,7 @@ BEGIN
     IDS_NONE "(None)"
 
     /* Friendly File Type Names */
+    IDS_FILE "Soubor"
```
GeoB99, line 928: It is best to leave the resource string left untranslated for other locales and let other maintainers translate it, whereas you can just simply translate it in your own native language(s).

dll/win32/shell32/wine/shell32_main.c
JoachimHenze: Sending to Wine is better than guarding. Especially for such a minor translation thingy.

dll/win32/shell32/wine/shell32_main.c
```diff
@@ -568,28 +568,30 @@ DWORD_PTR WINAPI SHGetFileInfoW(LPCWSTR path,DWORD dwFileAttributes,
         else
         {
             if (dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY)
-                strcatW (psfi->szTypeName, L"Folder");
+                LoadStringW(shell32_hInstance, IDS_DIRECTORY, psfi->szTypeName, 80);
```
GeoB99, line 571: It is good practice to not hardcode the length of the buffer, instead of `80` replace it with `_countof(psfi->szTypeName)` to compute the count of elements of the string array. Same thing for the `StringCchPrintfW` call below.

dll/win32/shell32/wine/shell32_main.c
3ts75: Now, let's revert the changes to WINE and only change the resources.

dll/win32/shell32/lang/cs-CZ.rc
3ts75: I see, This time I diverted from IDS_ANY_FILE.

dll/win32/shell32/wine/shell32_main.c
3ts75: @JoachimHenze 
I can only imagine how to guard against it using "#ifdef __REACTOS__".
Is that still a problem?

JoachimHenze: That would definitely be better than the current state of the PR, yes.

But the best would be if you opened a PR in the Wine repository instead of in the ReactOS repository with your intended changes. We could then later sync against their code once your proposal was accepted by Wine.
And therefore have the optimal/shortest code without any guards, and with the best sync to Wine codebase.
I think the issue discussed is definitely trivial enough to allow such an additional delay. We are not discussing any crash here,  but just a trivial translation glitch after all.

PR #5396, opened 2023-07-03
[RTL][ADVAPI32_APITEST] RtlIsTextUnicode: Don't change weight unless lead-byte flag

katahiromz:
## Purpose

Fix the garbled characters (a.k.a. mojibake) in some Japanese text files.

JIRA issue: [CORE-19016](https://jira.reactos.org/browse/CORE-19016)

## Proposed changes

- `weight` will remain `3` unless `IS_TEXT_UNICODE_DBCS_LEADBYTE` flag is set.
- Strengthen `advapi32_apitest IsTextUnicode` testcase.

## TODO

- [x] Do small tests.
- [x] Do big tests.

## Comparison

BEFORE:
before

AFTER:
after

BEFORE:
before

AFTER:
after

### DIFF
--- a/modules/rostests/apitests/advapi32/IsTextUnicode.c
+++ b/modules/rostests/apitests/advapi32/IsTextUnicode.c
@@ -128,6 +128,9 @@ START_TEST(IsTextUnicode)
     };
 
     const char japanese_with_lead[] = "ABC" "\x83\x40" "D";
+    const char japanese_sjis[] = "\x93\xFA\x96\x7B\x8C\xEA\x31\x32\x33\x93\xFA\x96\x7B\x8C\xEA";
+    const char japanese_utf8[] = "\xE6\x97\xA5\xE6\x9C\xAC\xE8\xAA\x9E\x31\x32\x33\xE6\x97\xA5"
+                                 "\xE6\x9C\xAC\xE8\xAA\x9E";
     const char simplfied_chinese_with_lead[] = "ABC" "\xC5\xC5" "D";
     const char korean_with_lead[] = "ABC" "\xBF\xAD" "D";
     const char traditional_chinese_with_lead[] = "ABC" "\xB1\xC1" "D";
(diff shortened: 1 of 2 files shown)


HBelusca:
Do we have an apitest or a winetest that backs this change?

katahiromz:
> Do we have an apitest or a winetest that backs this change?

I will add it.

sdk/lib/rtl/unicode.c
```diff
@@ -1367,7 +1367,7 @@ RtlIsTextUnicode(CONST VOID* buf, INT len, INT* pf)
             }
         }
 
-        if (lead_byte)
+        if (lead_byte && pf && (*pf & IS_TEXT_UNICODE_DBCS_LEADBYTE))
```
SergeGautherie, line 1370: Moved conditions can move to line 1359 even.

HBelusca: Those need to be done only on lead_byte.

katahiromz:
The tests are strengthened.

WinXP:
WinXP
Successful.

Win2k3:
Win2k3
Successful.

katahiromz:
`Test KVM` BEFORE: 16882dc https://reactos.org/testman/compare.php?ids=87967
`Test KVM` AFTER: 8d262fa
https://reactos.org/testman/compare.php?ids=87967,???

`Test VBox` BEFORE: 16882dc https://reactos.org/testman/compare.php?ids=87968
`Test VBox` AFTER: 8d262fa
https://reactos.org/testman/compare.php?ids=87968,???

modules/rostests/apitests/advapi32/IsTextUnicode.c
```diff
@@ -152,6 +155,24 @@ START_TEST(IsTextUnicode)
     ok(!IsTextUnicode(japanese_with_lead, sizeof(japanese_with_lead), &Result), "IsTextUnicode() returned TRUE, expected FALSE\n");
     ok(Result == IS_TEXT_UNICODE_DBCS_LEADBYTE, "Result returned 0x%x, expected 0x%x\n", Result, IS_TEXT_UNICODE_DBCS_LEADBYTE);
 
+    Result = IS_TEXT_UNICODE_STATISTICS | IS_TEXT_UNICODE_REVERSE_STATISTICS;
+    ok(!IsTextUnicode(japanese_sjis, sizeof(japanese_sjis) - 1, &Result), "IsTextUnicode() returned TRUE, expected FALSE\n");
+    ok(Result == 0, "Result returned 0x%x, expected 0x%x\n", Result, 0);
```
SergeGautherie, line 160:
```suggestion
    ok(Result == 0, "Result returned 0x%x, expected 0\n", Result);
```

sdk/lib/rtl/unicode.c
SergeGautherie: > Those need to be done only on lead_byte.

PR code looks the opposite to me : `lead_byte` is useless unless the flag is present.

katahiromz: Partially moved.

katahiromz:
`Test KVM` BEFORE: 5ee97b9 https://reactos.org/testman/compare.php?ids=88055
`Test KVM` AFTER: e5783e4
https://reactos.org/testman/compare.php?ids=88055,???

`Test VBox` BEFORE: 5ee97b9 https://reactos.org/testman/compare.php?ids=88056
`Test VBox` AFTER: e5783e4
https://reactos.org/testman/compare.php?ids=88056,???

HBelusca (approved):
looks ok

HBelusca:
As a side-note on this "DBCS-leadbyte" flag, see http://archives.miloush.net/michkap/archive/2005/01/30/363308.html /
> A bit of trivia -- the code for a flag that used to be documented (IS_TEXT_UNICODE_DBCS_LEADBYTE) is still there (and it is still in the header file, obviously -- the PSDK never breaks people like that). But the flag does not work well, so it is probably just as well that it is not documented any more. I highly recommend not passing it. Or ignoring when it is returned. The flag not dangerous or anything; it's just not too terribly useful for its intended purpose (detecting text that is actually DBCS).

sdk/lib/rtl/unicode.c
```diff
@@ -1356,7 +1356,7 @@ RtlIsTextUnicode(CONST VOID* buf, INT len, INT* pf)
         }
     }
 
-    if (NlsMbCodePageTag)
+    if (NlsMbCodePageTag && pf && (*pf & IS_TEXT_UNICODE_DBCS_LEADBYTE))
```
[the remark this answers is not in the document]
HBelusca: Maybe, but note that `flags` gets set to a possible non-zero value (with `IS_TEXT_UNICODE_DBCS_LEADBYTE` possibly) only if `pf != NULL`, see the top of the function.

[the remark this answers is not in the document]
katahiromz:
```c
unsigned int flags = MAXULONG, out_flags = 0;
```
It's unexpected `MAXLONG` if `pf == NULL`.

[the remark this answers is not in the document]
HBelusca: Ah yes, you're correct.
Then (from MSDN):
> If lpiResult [called "pf" in our code] is NULL, the function uses all available tests to determine if the data in the buffer is likely to be Unicode text.

But indeed that dbcs_leadbyte is about checking for dbcs, not unicode, so indeed, checking for `pf` may be more logical in that regard.

PR #5403, opened 2023-07-04
Fixed configure.cmd not always creating REACTOS_OUTPUT_PATH

whindsaks:
## Purpose

The configure batch does not always create the REACTOS_OUTPUT_PATH directory when configure is executed in the ROS source directory because it does a case-sensitive path comparison to detect this state.

## Proposed changes

- Perform a case-insensitive path comparison

### DIFF
--- a/configure.cmd
+++ b/configure.cmd
@@ -157,7 +157,7 @@ if "%VS_SOLUTION%" == "1" (
     set REACTOS_OUTPUT_PATH=%REACTOS_OUTPUT_PATH%-sln
 )
 
-if "%REACTOS_SOURCE_DIR%" == "%CD%\" (
+if /I "%REACTOS_SOURCE_DIR%" == "%CD%\" (
     set CD_SAME_AS_SOURCE=1
     echo Creating directories in %REACTOS_OUTPUT_PATH%
 


JoachimHenze:
And in which cases does the case differ? That path is created by the same script.
Can this actually ever fail on Windows?
If not, then the actual bug is otherwhere in the ReactOS sources!
I do suspect that you do actually hide a bug in the way ReactOS deals with the filesystem by this change!

whindsaks:
It looks like %~dp0 is upper-cased for reasons I don't understand. Does not matter if I'm in the directory or not. 

ROS

(Renaming the folder to "DeskTOP" has the same result)

That being said, I don't think Microsoft specifies the case of %0. Does it come from the command line? From FindFirstFile?

Paths on Windows are case-insensitive so any batch file that does not use /I is broken IMHO.

HBelusca:
Problem seems to happen from ROS' own cmd.exe itself indeed:

- Comparison with Windows 2000 and 2003 cmd.exe's running in ROS:
image

- Comparison when the same is run on the ROS virtual hard disk image mounted on Windows 7:
image

But the key point to observe is that the reason why "BUREAU" (or "DESKTOP") is in caps appears to be related to the directory in particular and the fact it is on a FAT partition. ROS cmd.exe doesn't always show the last directory in full caps though.

HBelusca:
Debugging cmd.exe in Windows:
image
For whatever reason (one of the conditions in the if-block is wrong) it uses the cAlternateFileName (aka. the 8.3-format short name), while it should NOT.
As you can see, the 8.3 name is in all caps.

HBelusca:
The root cause of this problem has been fixed in master. Please retest again, but I don't think adding the `/I` flag adds any robustness in this case, since now the `%~dp0` has been fixed, and the pseudo-environment variable `%CD%` should also behave consistently (without any difference of case between the two).

HBelusca:
Superseded by PR #5433, but thanks for having reported the problem!

PR #5407, opened 2023-07-06
[NETSTAT] pragma once, reorder, fix pl-PL.rc

JoachimHenze:
- add a pragma once into the header
- order -o option alphabetically also for the global variables, not only in usage-help
- pl-PL.rc: fix wrong alignment between IDS_ETHERNET_THEADER and its data
- pl-PL.rc: also don't pad the data with 2 spaces here in the beginning, which we don't do in any other locale, and MS netstat also doesn't do that here. pl-PL.rc most likely got this wrong because it was created while en-US wasn't finished yet.

here is a pic of the **fixed** polish, or let's better say: 'polished' state:
netstat_polish_after


katahiromz (approved):

binarymaster (approved):
absolutely polished 🇵🇱

PR #5420, opened 2023-07-10
[SHELL32] Update the status bar asynchronously

katahiromz:
## Purpose
Selecting all in `system32` was too slow.
JIRA issue: [CORE-18663](https://jira.reactos.org/browse/CORE-18663)

## Proposed changes

- Make `CDefView::UpdateStatusbar` function asynchronized by using a thread.
- It makes selecting all very quick (within one second).
- Add `m_hUpdateStatusbarThread` member variable in order to asynchronize.

after

## TODO

- [x] Do tests.

### DIFF
--- a/dll/win32/shell32/CDefView.cpp
+++ b/dll/win32/shell32/CDefView.cpp
@@ -173,7 +173,7 @@ class CDefView :
         HRESULT IncludeObject(PCUITEMID_CHILD pidl);
         HRESULT OnDefaultCommand();
         HRESULT OnStateChange(UINT uFlags);
-        static unsigned __stdcall UpdateStatusbarProc(void *args);
+        static unsigned __stdcall _UpdateStatusbarProc(void *args);
         void UpdateStatusbarWorker(HANDLE hThread);
         void UpdateStatusbar();
         void CheckToolbar();
@@ -625,7 +625,7 @@ void CDefView::UpdateStatusbarWorker(HANDLE hThread)
     }
 }
 
-unsigned __stdcall CDefView::UpdateStatusbarProc(void *args)
+unsigned __stdcall CDefView::_UpdateStatusbarProc(void *args)
 {
     CDefView* pView = static_cast<CDefView*>(args);
     pView->UpdateStatusbarWorker(pView->m_hUpdateStatusbarThread);
@@ -639,7 +639,7 @@ void CDefView::UpdateStatusbar()
     // We have to initialize m_hUpdateStatusbarThread before the thread starts up.
     // Thus, we use CREATE_SUSPENDED.
     m_hUpdateStatusbarThread =
-        reinterpret_cast<HANDLE>(_beginthreadex(NULL, 0, UpdateStatusbarProc, this,
+        reinterpret_cast<HANDLE>(_beginthreadex(NULL, 0, _UpdateStatusbarProc, this,
                                                 CREATE_SUSPENDED, NULL));
     ::ResumeThread(m_hUpdateStatusbarThread);
 
--- a/dll/win32/shell32/CDefView.cpp
+++ b/dll/win32/shell32/CDefView.cpp
@@ -545,7 +545,7 @@ void CDefView::CheckToolbar()
 
 void CDefView::UpdateStatusbarWorker(HANDLE hThread)
 {
-    WCHAR szFormat[MAX_PATH], szPartText[MAX_PATH], szPartText2[MAX_PATH];
+    WCHAR szFormat[MAX_PATH], szPartText[MAX_PATH];
     UINT cSelectedItems;
 
     cSelectedItems = m_ListView.GetSelectedCount();
@@ -591,29 +591,24 @@ void CDefView::UpdateStatusbarWorker(HANDLE hThread)
             }
         }
 
-        szPartText[0] = szPartText2[0] = UNICODE_NULL;
-
         /* Don't show the file size text if there is 0 bytes in the folder
          * OR we only have folders selected. */
+        szPartText[0] = UNICODE_NULL;
         if ((cSelectedItems && !bIsOnlyFoldersSelected) || uTotalFileSize)
             StrFormatByteSizeW(uTotalFileSize, szPartText, _countof(szPartText));
 
-        if (hThread != m_hUpdateStatusbarThread)
-            return;
+        m_pShellBrowser->SendControlMsg(FCW_STATUS, SB_SETTEXT, 1, (LPARAM)szPartText, &lResult);
 
         /* If we are in a Recycle Bin folder then show no text for the location part. */
+        szPartText[0] = UNICODE_NULL;
         if (!_ILIsBitBucket(m_pidlParent))
         {
-            LoadStringW(shell32_hInstance, IDS_MYCOMPUTER, szPartText2, _countof(szPartText2));
+            LoadStringW(shell32_hInstance, IDS_MYCOMPUTER, szPartText, _countof(szPartText));
             pIcon = (LPARAM)m_hMyComputerIcon;
         }
 
-        if (hThread != m_hUpdateStatusbarThread)
-            return;
-
-        m_pShellBrowser->SendControlMsg(FCW_STATUS, SB_SETTEXT, 1, (LPARAM)szPartText, &lResult);
         m_pShellBrowser->SendControlMsg(FCW_STATUS, SB_SETICON, 2, pIcon, &lResult);
-        m_pShellBrowser->SendControlMsg(FCW_STATUS, SB_SETTEXT, 2, (LPARAM)szPartText2, &lResult);
+        m_pShellBrowser->SendControlMsg(FCW_STATUS, SB_SETTEXT, 2, (LPARAM)szPartText, &lResult);
     }
 }
 
--- a/dll/win32/shell32/CDefView.cpp
+++ b/dll/win32/shell32/CDefView.cpp
@@ -605,9 +605,6 @@ void CDefView::UpdateStatusbarWorker(HANDLE hThread)
             *szPartText = 0;
         }
 
-        if (hThread != m_hUpdateStatusbarThread)
-            return;
-
         m_pShellBrowser->SendControlMsg(FCW_STATUS, SB_SETTEXT, 1, (LPARAM)szPartText, &lResult);
 
         /* If we are in a Recycle Bin folder then show no text for the location part. */
--- a/dll/win32/shell32/CDefView.cpp
+++ b/dll/win32/shell32/CDefView.cpp
@@ -175,7 +175,7 @@ class CDefView :
         HRESULT OnDefaultCommand();
         HRESULT OnStateChange(UINT uFlags);
         void UpdateStatusbarInner();
-        void UpdateStatusbarAsync();
+        void UpdateStatusbar();
         void CheckToolbar();
         BOOL CreateList();
         void UpdateListColors();
@@ -634,7 +634,7 @@ static unsigned __stdcall UpdateStatusProc(void *args)
     return 0;
 }
 
-void CDefView::UpdateStatusbarAsync()
+void CDefView::UpdateStatusbar()
 {
     HANDLE hOldThread = m_hUpdateStatusbarThread;
     m_hUpdateStatusbarThread = (HANDLE)_beginthreadex(NULL, 0, UpdateStatusProc, this, 0,
@@ -1334,7 +1334,7 @@ LRESULT CDefView::OnCreate(UINT uMsg, WPARAM wParam, LPARAM lParam, BOOL &bHandl
         _ForceStatusBarResize();
     }
 
-    UpdateStatusbarAsync();
+    UpdateStatusbar();
 
     return S_OK;
 }
@@ -1774,7 +1774,7 @@ LRESULT CDefView::OnSize(UINT uMsg, WPARAM wParam, LPARAM lParam, BOOL &bHandled
     _DoFolderViewCB(SFVM_SIZE, 0, 0);
 
     _HandleStatusBarResize(wWidth);
-    UpdateStatusbarAsync();
+    UpdateStatusbar();
 
     return 0;
 }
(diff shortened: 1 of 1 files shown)


dll/win32/shell32/CDefView.cpp
```diff
@@ -607,11 +619,29 @@ void CDefView::UpdateStatusbar()
...
+{
+    static_cast<CDefView*>(args)->UpdateStatusbarInner();
+    return 0;
+}
+
+void CDefView::UpdateStatusbar()
+{
+    HANDLE hOldThread = m_hUpdateStatusbarThread;
+    m_hUpdateStatusbarThread = reinterpret_cast<HANDLE>(
+        _beginthreadex(NULL, 0, UpdateStatusbarProc, this, 0, &m_uUpdateStatusbarThreadId));
```
HBelusca, line 640: Does shell32 already use that function elsewhere to create threads?

dll/win32/shell32/CDefView.cpp
```diff
@@ -153,6 +154,8 @@ class CDefView :
         SFVM_CUSTOMVIEWINFO_DATA  m_viewinfo_data;
 
         HICON                     m_hMyComputerIcon;
+        HANDLE                    m_hUpdateStatusbarThread;
+        unsigned                  m_uUpdateStatusbarThreadId;
```
HBelusca, line 158:
```suggestion
        unsigned int              m_uUpdateStatusbarThreadId;
```

dll/win32/shell32/CDefView.cpp
katahiromz: There is `SHCreateThread` function in `shlwapi`, but it has no thread id paramter.

HBelusca: Do we need to keep a thread ID around for this PR functionality, or is it sufficient to keep the handle?

katahiromz: Using IDs is better in mechanism. That will distinguish threads.

HBelusca: imo thread handles better distinguish them: they are guaranteed to be unique and you can control the thread by using handles, not by using IDs.

dll/win32/shell32/CDefView.cpp
```diff
@@ -539,11 +544,12 @@ void CDefView::CheckToolbar()
     }
 }
 
-void CDefView::UpdateStatusbar()
+void CDefView::UpdateStatusbarInner()
```
HBelusca, line 547: Using standard naming would be:
```suggestion
void CDefView::UpdateStatusbarWorker()
```
that would be called from the thread function.

dll/win32/shell32/CDefView.cpp
```diff
@@ -577,6 +583,9 @@ void CDefView::UpdateStatusbar()
 
         while ((nItem = m_ListView.GetNextItem(nItem, uFileFlags)) >= 0)
         {
+            if (dwThreadId != m_uUpdateStatusbarThreadId)
+                return;
```
HBelusca, line 587: Is this done in order to avoid this worker function to be called by two differently launched threads?
If so, in which conditions would there be more than one thread launched?

dll/win32/shell32/CDefView.cpp
katahiromz: Renamed. Added parameter `hThread`.

dll/win32/shell32/CDefView.cpp
```diff
@@ -607,11 +618,31 @@ void CDefView::UpdateStatusbar()
...
 
+        if (hThread != m_hUpdateStatusbarThread)
+            return;
+
         m_pShellBrowser->SendControlMsg(FCW_STATUS, SB_SETICON, 2, pIcon, &lResult);
         m_pShellBrowser->SendControlMsg(FCW_STATUS, SB_SETTEXT, 2, (LPARAM)szPartText, &lResult);
     }
 }
 
+static unsigned __stdcall UpdateStatusbarProc(void *args)
```
HBelusca, line 629: This should be a static method of CDefView. This would certainly allow you to access the `m_hUpdateStatusbarThread` member of `pView` even if that member was private (or protected?)

dll/win32/shell32/CDefView.cpp
katahiromz: Calculating total size is slow operation and it takes time.
`UpdateStatusbar` is triggered on every selection.
The thread is different from the variable, then the function aborts it.

HBelusca: OK. Question: Suppose I am in system32 and I select all the files, then very quickly I cancel the selection (by e.g. clicking somewhere else in the file list) or I change the selection to a different set of selected files (for example, one single file) while staying in that same directory.
Or alternatively, suppose I first select all files in system32, then very quickly open a subbdirectory in system32 (while being in the same explorer window).

--> Would the existing update thread be stopped in a timely fashion, and the status bar reflect the new change?
-

dll/win32/shell32/CDefView.cpp
katahiromz: OK. Thanks.

dll/win32/shell32/CDefView.cpp
katahiromz: A selection change will trigger `LVN_ITEMCHANGED` message and `UpdateStatusbar` call.
Going to another view will trigger `DestroyViewWindow` and it will cancel `UpdateStatusbar` task.

dll/win32/shell32/CDefView.cpp
katahiromz: Renamed as `_UpdateStatusbarProc`.

dll/win32/shell32/CDefView.cpp
katahiromz: Used thread handles. Dismissed thread IDs.

dll/win32/shell32/CDefView.cpp
HBelusca: I see. Wouldn't it be useful then to stop the previous update thread in UpdateStatusbar() , before starting the new one?

dll/win32/shell32/CDefView.cpp
```diff
@@ -2582,6 +2617,14 @@ HRESULT WINAPI CDefView::DestroyViewWindow()
 {
     TRACE("(%p)\n", this);
 
+    if (m_hUpdateStatusbarThread)
+    {
+        HANDLE hOldThread = m_hUpdateStatusbarThread;
+        m_hUpdateStatusbarThread = NULL;
+        ::WaitForSingleObject(hOldThread, INFINITE);
```
learn-more, line 2624: Will this block browsing until the thread is killed?

katahiromz: Assigning `NULL` to `m_hUpdateStatusbarThread` will terminate the thread immediately.

dll/win32/shell32/CDefView.cpp
katahiromz: The modern computer is very fast. Assigning a handle to `m_hUpdateStatusbarThread` may come after the beginning of the new thread.

HBelusca: It would be better to not allow bugs to be present just because "the modern computer" hides them. ReactOS is also run in virtual machines with different speeds, and e.g. running ROS in QEmu on Windows is not the fastest thing in existence. There, you could get into troubles.

dll/win32/shell32/CDefView.cpp
learn-more: Not immediately, because then there would be no need for a thread (if all operations in the thread return immediately).

dll/win32/shell32/CDefView.cpp
```diff
@@ -607,11 +614,39 @@ void CDefView::UpdateStatusbar()
             pIcon = (LPARAM)m_hMyComputerIcon;
         }
 
+        if (hThread != m_hUpdateStatusbarThread)
```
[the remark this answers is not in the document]
katahiromz: If the target variable has changed In the iteration of selection items, it might be not updated.
If it was outside of the iteration, it might be updated.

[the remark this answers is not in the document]
katahiromz: Refactored and optimized.

dll/win32/shell32/CDefView.cpp
```diff
@@ -587,28 +591,54 @@ void CDefView::UpdateStatusbar()
...
 
         /* If we are in a Recycle Bin folder then show no text for the location part. */
         if (!_ILIsBitBucket(m_pidlParent))
         {
-            LoadStringW(shell32_hInstance, IDS_MYCOMPUTER, szPartText, _countof(szPartText));
+            LoadStringW(shell32_hInstance, IDS_MYCOMPUTER, szPartText2, _countof(szPartText2));
             pIcon = (LPARAM)m_hMyComputerIcon;
         }
 
+        if (hThread != m_hUpdateStatusbarThread)
```
HBelusca, line 611: Not sure it's useful to have that check just after loading the string, while the thread check has been done just above.

dll/win32/shell32/CDefView.cpp
```diff
@@ -587,28 +591,54 @@ void CDefView::UpdateStatusbar()
...
 
-        m_pShellBrowser->SendControlMsg(FCW_STATUS, SB_SETTEXT, 1, (LPARAM)szPartText, &lResult);
+        if (hThread != m_hUpdateStatusbarThread)
+            return;
 
         /* If we are in a Recycle Bin folder then show no text for the location part. */
         if (!_ILIsBitBucket(m_pidlParent))
         {
-            LoadStringW(shell32_hInstance, IDS_MYCOMPUTER, szPartText, _countof(szPartText));
+            LoadStringW(shell32_hInstance, IDS_MYCOMPUTER, szPartText2, _countof(szPartText2));
```
HBelusca, line 607: I don't think I understand why we need yet another variable there. (I know this is related to the comment below, but still...)

Why not having:
```c++
        if (hThread != m_hUpdateStatusbarThread)
            return;

        // Now do all the "UI" update:
        //
        m_pShellBrowser->SendControlMsg(FCW_STATUS, SB_SETTEXT, 1, (LPARAM)szPartText, &lResult);
        //
        // and the bitbucket stuff:
        // if (!_ILIsBitBucket(m_pidlParent)) { /* load string etc. */ }
        m_pShellBrowser->SendControlMsg(FCW_STATUS, SB_SETICON, 2, pIcon, &lResult);
        m_pShellBrowser->SendControlMsg(FCW_STATUS, SB_SETTEXT, 2, (LPARAM)szPartText2, &lResult);
```

dll/win32/shell32/CDefView.cpp
katahiromz: Deleted.

dll/win32/shell32/CDefView.cpp
katahiromz: Fixed.

dll/win32/shell32/CDefView.cpp
```diff
@@ -612,6 +612,31 @@ void CDefView::UpdateStatusbar()
     }
 }
 
+unsigned __stdcall CDefView::_UpdateStatusbarProc(void *args)
+{
+    CDefView* pView = static_cast<CDefView*>(args);
+    pView->UpdateStatusbarWorker(pView->m_hUpdateStatusbarThread);
```
learn-more, line 618: I'd expect an `AddRef()` before starting this thread, and a `Release()` as last thing in this function

dll/win32/shell32/CDefView.cpp
```diff
@@ -612,6 +612,33 @@ void CDefView::UpdateStatusbar()
     }
 }
 
+unsigned __stdcall CDefView::_UpdateStatusbarProc(void *args)
+{
+    CDefView* pView = static_cast<CDefView*>(args);
+    pView->AddRef();
```
learn-more, line 618: This `AddRef` is not before starting the thread, it should be before calling `_beginthreadex`, so that the object stays valid as long as the thread is alive

dll/win32/shell32/CDefView.cpp
```diff
@@ -612,6 +612,33 @@ void CDefView::UpdateStatusbar()
...
+    return 0;
+}
+
+void CDefView::UpdateStatusbar()
+{
+    HANDLE hOldThread = m_hUpdateStatusbarThread;
+
+    // We have to initialize m_hUpdateStatusbarThread before the target thread begins.
+    // So, we use CREATE_SUSPENDED.
+    HANDLE hNewThread = reinterpret_cast<HANDLE>(_beginthreadex(NULL, 0, _UpdateStatusbarProc,
```
learn-more, line 630: Ideally the `AddRef` should be before this call

katahiromz: OK. Thank you.

dll/win32/shell32/CDefView.cpp
katahiromz: Fixed.

dll/win32/shell32/CDefView.cpp
learn-more: The purpose is to ensure that the object cannot be destroyed while the thread is starting/running.

dll/win32/shell32/CDefView.cpp
HBelusca: > Confirmed: Going to another view will trigger DestroyViewWindow and it will cancel UpdateStatusbar task.

To be clear, the hypothetical situation I was presenting in those questions was such that the **SAME** explorer window was being used to go through different subdirectories. Not by killing the window.

dll/win32/shell32/CDefView.cpp
```diff
@@ -2582,6 +2614,17 @@ HRESULT WINAPI CDefView::DestroyViewWindow()
 {
     TRACE("(%p)\n", this);
 
+    if (m_hUpdateStatusbarThread)
+    {
+        HANDLE hOldThread = m_hUpdateStatusbarThread;
+
+        // Assigning NULL to m_hUpdateStatusbarThread will terminate the target thread immediately
```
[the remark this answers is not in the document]
katahiromz: Deleted: https://github.com/reactos/reactos/commit/74ca0a12ad5df16195622c015b628f1f6cb10c8a

PR #5422, opened 2023-07-10
[FONTEXT] Update Romanian (ro-RO) translation

AndreiMiloiu:


AndreiMiloiu:
Cc @reactos/lang-romanian

PR #5423, opened 2023-07-10
[EXPLORER] Improve the minimize-condition on Win+D

katahiromz:
## Purpose

JIRA issue: [CORE-18427](https://jira.reactos.org/browse/CORE-18427)

## Proposed changes

- Add `CanBeMinimized` helper function to determine whether the window should be minimized.
- Use them in `FindEffectiveProc` and `MinimizeWindowsProc` functions.
- Improve the Minimize code.

## TODO

- [x] Do tests.

## Comparison

BEFORE:
before
Magnifier does minimize on `Win+D`.

AFTER:
after
Magnifier won't minimize on `Win+D`.

AFTER:
after2
Task Manager does minimize on `Win+D`.

### DIFF
--- a/base/shell/explorer/traywnd.cpp
+++ b/base/shell/explorer/traywnd.cpp
@@ -111,9 +111,9 @@ BOOL ShouldMinimize(HWND hwnd)
     if (::IsWindowVisible(hwnd) && !::IsIconic(hwnd) && ::IsWindowEnabled(hwnd))
     {
         DWORD style = static_cast<DWORD>(::GetWindowLongPtrW(hwnd, GWL_STYLE));
-        if (style & WS_MINIMIZEBOX)
+        if ((style & WS_MINIMIZEBOX) && (style & WS_CAPTION) == WS_CAPTION)
         {
-            if ((style & (WS_CAPTION | WS_SYSMENU)) == (WS_CAPTION | WS_SYSMENU))
+            if (style & WS_SYSMENU)
             {
                 HMENU hSysMenu = ::GetSystemMenu(hwnd, FALSE);
                 if (hSysMenu)
--- a/base/shell/explorer/traywnd.cpp
+++ b/base/shell/explorer/traywnd.cpp
@@ -106,12 +106,9 @@ BOOL CanBeMinimized(HWND hwnd)
         if (::GetClassLongPtrW(hwnd, GCW_ATOM) == (ULONG_PTR)WC_DIALOG)
             return TRUE;
 
-        DWORD style = (DWORD)::GetWindowLongPtrW(hwnd, GWL_STYLE);
-        if (style & WS_MINIMIZEBOX)
-        {
-            if ((style & WS_CAPTION) == WS_CAPTION)
-                return TRUE;
-        }
+        DWORD exstyle = (DWORD)::GetWindowLongPtrW(hwnd, GWL_EXSTYLE);
+        if (!(exstyle & WS_EX_TOPMOST))
+            return TRUE;
     }
     return FALSE;
 }
--- a/base/shell/explorer/traywnd.cpp
+++ b/base/shell/explorer/traywnd.cpp
@@ -3289,7 +3289,7 @@ HandleTrayContextMenu:
 
             info->pMinimizedAll->Add(mwp);
 
-            ::ShowWindowAsync(hwnd, SW_MINIMIZE);
+            ::ShowWindowAsync(hwnd, SW_SHOWMINNOACTIVE);
         }
 
         return TRUE;
--- a/base/shell/explorer/traywnd.cpp
+++ b/base/shell/explorer/traywnd.cpp
@@ -3338,9 +3338,7 @@ HandleTrayContextMenu:
         {
             HWND hwnd = g_MinimizedAll[i].hwnd;
             if (::IsWindowVisible(hwnd) && ::IsIconic(hwnd))
-            {
                 ::SetWindowPlacement(hwnd, &g_MinimizedAll[i].wndpl);
-            }
         }
 
         g_MinimizedAll.RemoveAll();
--- a/base/shell/explorer/traywnd.cpp
+++ b/base/shell/explorer/traywnd.cpp
@@ -99,30 +99,18 @@ VOID RestoreWindowPos()
     g_WindowPosBackup.RemoveAll();
 }
 
-BOOL CanDialogSysMinimize(HWND hwnd)
-{
-    if (::GetClassLongPtrW(hwnd, GCW_ATOM) != (ULONG_PTR)WC_DIALOG)
-        return FALSE;
-    return (::IsWindowVisible(hwnd) && !::IsIconic(hwnd) && ::IsWindowEnabled(hwnd));
-}
-
-BOOL ShouldMinimize(HWND hwnd)
+BOOL HasToBeMinimized(HWND hwnd)
 {
     if (::IsWindowVisible(hwnd) && !::IsIconic(hwnd) && ::IsWindowEnabled(hwnd))
     {
+        if (::GetClassLongPtrW(hwnd, GCW_ATOM) == (ULONG_PTR)WC_DIALOG)
+            return TRUE;
+
         DWORD style = (DWORD)::GetWindowLongPtrW(hwnd, GWL_STYLE);
-        if ((style & WS_CAPTION) == WS_CAPTION)
+        if (style & WS_MINIMIZEBOX)
         {
-            if (style & WS_SYSMENU)
-            {
-                HMENU hSysMenu = ::GetSystemMenu(hwnd, FALSE);
-                if (hSysMenu)
-                    return !(::GetMenuState(hSysMenu, SC_MINIMIZE, MF_BYCOMMAND) & MF_DISABLED);
-            }
-            else
-            {
+            if ((style & WS_CAPTION) == WS_CAPTION)
                 return TRUE;
-            }
         }
     }
     return FALSE;
@@ -142,7 +130,7 @@ FindEffectiveProc(HWND hwnd, LPARAM lParam)
 {
     EFFECTIVE_INFO *pei = (EFFECTIVE_INFO *)lParam;
 
-    if (!ShouldMinimize(hwnd) && !CanDialogSysMinimize(hwnd))
+    if (!HasToBeMinimized(hwnd))
         return TRUE;    // continue
 
     if (pei->hTrayWnd == hwnd || pei->hwndDesktop == hwnd ||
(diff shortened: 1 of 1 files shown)


learn-more (approved):

base/shell/explorer/traywnd.cpp
```diff
@@ -159,13 +163,6 @@ IsThereAnyEffectiveWindow(BOOL bMustBeInMonitor)
...
 
     EnumWindows(FindEffectiveProc, (LPARAM)&ei);
-    if (ei.hwndFound && FALSE)
-    {
-        WCHAR szClass[64], szText[64];
-        GetClassNameW(ei.hwndFound, szClass, _countof(szClass));
-        GetWindowTextW(ei.hwndFound, szText, _countof(szText));
-        MessageBoxW(NULL, szText, szClass, 0);
-    }
     return ei.hwndFound != NULL;
```
JoachimHenze, line 166: Do I assume correctly, that the deletion of that block above has nothing to do with the fix, but is unrelated cleanup?

JoachimHenze (approved):

HBelusca:
Can you verify that this is the case also when taskmgr is opened in "Always on Top" mode? (which should set the `WS_EX_TOPMOST` style to its window)

binarymaster:
> Can you verify that this is the case also when taskmgr is opened in "Always on Top" mode?

I've tested it using Process Hacker with "Always on Top" enabled in Windows XP -- Win+D minimizes it anyway despite this feature enabled. Can't test with standard Task Manager atm.

HBelusca (changes requested):
Testing `Win+D` on Windows 2003 with taskmgr having the `WS_EX_TOPMOST` style clearly invalidates the test you did and the PR:

https://github.com/reactos/reactos/assets/1969829/cfe40a46-4459-4901-ac94-51a43efc610a

I suppose that the magnifier actively checks whether it receives a `WM_MINIMIZE` message and ignores it (thus maintaining its window on the screen).

EDIT 2: You may be able to also test and observe the following two things:
- Windows' own magnifier can be reduced (its dialog has the minimize button and a button is present in the taskbar);
- You can test our ReactOS' magnifier in Windows, and verify how it behaves there.

EDIT 3: I just went ahead and tested ROS' magnifier on Win2k3, and here it is:

https://github.com/reactos/reactos/assets/1969829/680028cf-5b37-476e-b576-9e77503dd213

Two things can be observed:
- `Win+D` _**DOES**_ minimize the magnifier's windows. They are however restored when pressing again `Win+D`, along with all the other minimized windows. They are _**also**_ listed in the `Alt+TAB` window and are restored by pressing `Alt+TAB` many times.
- Unrelated to this PR, ReactOS' magnifier is kind of not really usable in Windows due to unresponsiveness of the "advice-dialog" and of the other window, regarding button presses. (This should be visible in the video.)

base/shell/explorer/traywnd.cpp
katahiromz: `((...) && FALSE)` means `FALSE`.
The above code will be ignored and useless.

katahiromz:
I see that the top-most flag was unrelated on this case.
Some code is fixed.

HBelusca:
So, do you know (or have) a test app that we could test on Windows in order to see whether this is what happens?
Also, is this minimization condition only happening just for the Win+D handling in explorer, but does not happen in *all* cases, namely, whenever one always sends a WM_MINIMIZE message programmatically to the window/app in question ?

katahiromz:
There's no `WM_MINIMIZE` message that has such name.

On my Windows environment, our magnifier doesn't minimize on `Win+D`.

JoachimHenze:
Based on @HBelusca s findings I would assume that the PR is guaranteed to be wrong until at least some changes in our magnify binaries would be within. Therefore I will **not** restore my approval yet. I will wait until we solved the riddle that is standing in the room atm. Until then tendency for rejection on my side.

katahiromz:
https://jira.reactos.org/secure/attachment/66024/DontMinimize.zip
This is a sample console application.

katahiromz:
Let's go!

base/shell/explorer/traywnd.cpp
```diff
@@ -99,6 +99,35 @@ VOID RestoreWindowPos()
     g_WindowPosBackup.RemoveAll();
 }
 
+BOOL CanDialogSysMinimize(HWND hwnd)
+{
+    if (::GetClassLongPtrW(hwnd, GCW_ATOM) != reinterpret_cast<ULONG_PTR>(WC_DIALOG))
```
HBelusca, line 104: Just using `(ULONG_PTR)WC_DIALOG` wouldn't work?

base/shell/explorer/traywnd.cpp
```diff
@@ -3262,16 +3284,23 @@ class CTrayWindow :
...
+                ::ShowWindowAsync(hwnd, SW_SHOWMINNOACTIVE)) // Minimize
+            {
+                info->pMinimizedAll->Add(mwp);
+            }
+        }
+        else if (CanDialogSysMinimize(hwnd))
+        {
+            MINWNDPOS mwp = { hwnd, { sizeof(mwp.wndpl) } };
+            if (::GetWindowPlacement(hwnd, &mwp.wndpl) && // Save the position and status
+                ::PostMessageW(hwnd, WM_SYSCOMMAND, SC_MINIMIZE, -1)) // Minimize
```
HBelusca, line 3300: Doesn't `::ShowWindowAsync(hwnd, SW_SHOWMINNOACTIVE)` work for dialog boxes as well?

base/shell/explorer/traywnd.cpp
```diff
@@ -99,6 +99,35 @@ VOID RestoreWindowPos()
...
+    if (::IsWindowVisible(hwnd) && !::IsIconic(hwnd) && ::IsWindowEnabled(hwnd))
+    {
+        DWORD style = static_cast<DWORD>(::GetWindowLongPtrW(hwnd, GWL_STYLE));
+        if ((style & WS_MINIMIZEBOX) && (style & WS_CAPTION) == WS_CAPTION)
+        {
+            if (style & WS_SYSMENU)
+            {
+                HMENU hSysMenu = ::GetSystemMenu(hwnd, FALSE);
+                if (hSysMenu)
+                    return !(::GetMenuState(hSysMenu, SC_MINIMIZE, MF_BYCOMMAND) & MF_DISABLED);
```
HBelusca, line 120: Does your interactive test app also show that this is a condition that needs to be respected?

base/shell/explorer/traywnd.cpp
katahiromz: In C++, C++-typecast is recommended. But we can use the simpler expression. OK.

base/shell/explorer/traywnd.cpp
katahiromz: I'm not sure why it is this code... But it is not `ShouldMinimize`.

HBelusca: Do you mean that doing:
```c
        if (ShouldMinimize(hwnd) || CanDialogSysMinimize(hwnd))
        {
            MINWNDPOS mwp = { hwnd, { sizeof(mwp.wndpl) } };
            if (::GetWindowPlacement(hwnd, &mwp.wndpl) && // Save the position and status
/*** Here, either just using that ShowWindowAsync, or the PostMessage version only ***/
                ::ShowWindowAsync(hwnd, SW_SHOWMINNOACTIVE)) // Minimize
            {
                info->pMinimizedAll->Add(mwp);
            }
        }
```
wouldn't work?

HBelusca:
By the way I have a question:
those windows that look similar to the magnify (on ROS) are similar to those "tool-windows" we can see there for example:
image
Those tool windows are children of the main parent window. (That main parent window has the usual window style that can be minimized, etc.)
The question is: In this scenario, would these tool windows (and the main parent window) be reduced with Win+D with your PR?

base/shell/explorer/traywnd.cpp
katahiromz: Deleted `CanDialogSysMinimize`.

katahiromz:
They are "owner" windows.

An overlapped window is a top-level window that has a title bar, border, and client area.

A pop-up window is a special type of overlapped window used for dialog boxes, message boxes, and other temporary windows that appear outside an application's main window. 

Only an overlapped or pop-up window can be an owner window; a child window cannot be one.

`EnumWindows` enumerates the top-level windows. A top-level window is an owner window of top level.

base/shell/explorer/traywnd.cpp
```diff
@@ -99,6 +99,23 @@ VOID RestoreWindowPos()
     g_WindowPosBackup.RemoveAll();
 }
 
+BOOL HasToBeMinimized(HWND hwnd)
```
HBelusca, line 102: `CanBeMinimized`  :thinking:

base/shell/explorer/traywnd.cpp
```diff
@@ -99,6 +99,23 @@ VOID RestoreWindowPos()
...
+{
+    if (::IsWindowVisible(hwnd) && !::IsIconic(hwnd) && ::IsWindowEnabled(hwnd))
+    {
+        if (::GetClassLongPtrW(hwnd, GCW_ATOM) == (ULONG_PTR)WC_DIALOG)
```
HBelusca, line 106: Oh, so now *all* dialog windows can be minimized? Even those without minimize button or sthg?

katahiromz: Visible, non-minimized, and enabled dialog windows can be.

base/shell/explorer/traywnd.cpp
katahiromz: Fixed.

JoachimHenze:
How can the currrent fix proposal be correct, if it doesn't change anything on hbeluscas observsation:
"the ReactOS mangifier (with whatever window styles it uses), is indeed minimized when using Win+D on Windows."

It is supposed to not be minimized. How can we ever fix that, if we decide to touch our explorer instead?

base/shell/explorer/traywnd.cpp
HBelusca: But is this because they have a button in the taskbar ?

HBelusca:
> How can the currrent fix proposal be correct, if it doesn't change anything on hbeluscas observsation: "the ReactOS mangifier (with whatever window styles it uses), is indeed minimized when using Win+D on Windows."
> 
> It is supposed to not be minimized. How can we ever fix that, if we decide to touch our explorer instead?

A priori the same magnifier on Windows 2003 (but "Enterprise Edition") reduced in Katayama's test (see his video). Which is strange, as I said, because basically the same thing happened to behave differently...

katahiromz:
I seems lik: inconsistency of Magnifier behaviour depends on AppBar (attached to desktop side) or not...

base/shell/explorer/traywnd.cpp
katahiromz: 1. Created a window of the combination of `WS_POPUP` window style and `WS_EX_TOOLWINDOW` extended style.
2. It won't be shown on taskbar.
3. But `Win+D` works on it.

katahiromz:
I've found that there are some really minimized windows and visually-minimized-look windows.
`IsIconic` function won't give me the direct answer.

katahiromz:
Oh, I found the trick!

https://github.com/reactos/reactos/blob/cdfd8f4e1bb8be767a1941bd4e1d50bdd0ac152c/base/applications/magnify/magnifier.c#L282

The `WS_EX_TOPMOST` flag can be enabled depending on the magnifier settings. This flag affects whether the window is minimized or not.

base/shell/explorer/traywnd.cpp
HBelusca: > 3. But Win+D works on it.

Even if that dialog, with those styles, has its minimize button disabled or hidden?

katahiromz:
> to be done in a separate PR?

Yes.

base/shell/explorer/traywnd.cpp
katahiromz: Yes, it is. I did test.

https://jira.reactos.org/secure/attachment/66044/DialogTest.zip

PR #5428, opened 2023-07-12
[RAPPS][MC] Add Romanian (ro-RO) translation to message table

AndreiMiloiu:


AndreiMiloiu:
Cc @reactos/lang-romanian

base/applications/rapps/rappsmsg.mc
```diff
@@ -31,4 +35,7 @@ Removal of the program "%1" is called.
 Language=Polish
 Wywołano usunięcie programu "%1".
 .
+Language=Romanian
+Este apelată ștergerea programului "%1".
```
reactos-contributor, line 39: Some other verb besides "a șterge" (which is the closest translation for "to delete") might have fit better. However, this is not a deal-breaker.

reactos-contributor (approved):

PR #5430, opened 2023-07-12
[MYEVENTPROVIDER_DLL][MC] Add Romanian (ro-RO) translation to message table

AndreiMiloiu:


AndreiMiloiu:
Cc @reactos/lang-romanian

AndreiMiloiu:
I also don't know how to translate the word "server" in Romanian. Should I leave it as it is?

reactos-contributor:
> I also don't know how to translate the word "server" in Romanian.

Apparently, you should. Here is a resource with your name on "translators": [base/applications/mstsc/lang/ro-RO.rc](https://github.com/reactos/reactos/blob/master/base/applications/mstsc/lang/ro-RO.rc#L19)

modules/rostests/win32/advapi32/eventlog/MyEventProvider.mc
```diff
@@ -117,6 +140,9 @@ quarts%0
 Language=Polish
 kwarty%0
 .
+Language=Romanian
+litre%0
```
reactos-contributor, line 144: Just for the record, the "quart" here refers to the non SI unit of volume measurement (see the [Wikipedia page on it](https://en.wikipedia.org/wiki/Quart)), whose use in other languages is limited. Of course, there is no easy translatable option in these cases. The closest route to something in Romanian I could find is through French ([Wikipedia equivalent page to English "quart"](https://fr.wikipedia.org/wiki/Pinte)), ending up to "pintă": https://dexonline.ro/definitie/pint%C4%83
Please consider the above suggestion.

reactos-contributor (approved):
Looks good to me (regardless of the outcome related to given note).

modules/rostests/win32/advapi32/eventlog/MyEventProvider.mc
AndreiMiloiu: I modified it. For the word "server" is it good to translate it as "servitor informatic"?

reactos-contributor: That would be... bold, to say the least. So far, in Romanian, it was taken as is (although with inflections applied when it comes to that). However, it's not unprecedented for this term to be translated in other languages (e.g. [Croatian](https://hr.wikipedia.org/wiki/Poslu%C5%BEitelj), [Basque](https://eu.wikipedia.org/wiki/Zerbitzari), [Finnish](https://fi.wikipedia.org/wiki/Palvelin), [Macedonian](https://mk.wikipedia.org/wiki/%D0%9E%D0%BF%D1%81%D0%BB%D1%83%D0%B6%D1%83%D0%B2%D0%B0%D1%87), [Norwegian Nynorsk](https://nn.wikipedia.org/wiki/Tenarmaskin), [Turkish](https://tr.wikipedia.org/wiki/Sunucu_(bili%C5%9Fim)), and others). It's up to you.
Please do not forget update it everywhere, if you decide to change it.

PR #5431, opened 2023-07-12
[NETEVENT][MC] Add Romanian (ro-RO) translation to message table

AndreiMiloiu:


AndreiMiloiu:
Cc @reactos/lang-romanian

reactos-contributor (changes requested):
The translation needs to be reworked a bit. I'll review this after my comments get addressed.
@AndreiMiloiu: For overall body of this work, please note where the English content gets limited to lines of up to 80 characters in length and keep the same limit for translated ones.

sdk/include/reactos/mc/neteventmsg.mc
```diff
@@ -119,6 +132,9 @@ Un paquet de pilote reçu par le sous-système IO était invalide. La donnée es
 Language=Polish
 Pakiet sterownika otrzymany z podsystemu wejścia/wyjścia jest niepoprawny. Przedstawione dane to otrzymany pakiet.
 .
+Language=Romanian
+Un pachet primit al modulului pilot din subsistemul de intrare/ieșire a fost nevalid. Datele sunt pachetul însuși.
```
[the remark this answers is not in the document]
reactos-contributor: The above comment is not a call for change.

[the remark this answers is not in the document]
AndreiMiloiu: Shouldn't it be "In/Ie", instead of "In/Ex"? It seems to be more explicit."Ex" reffers to "Exit".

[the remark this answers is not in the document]
AndreiMiloiu: I remember that somewhere I translated with "In/Ie".

sdk/include/reactos/mc/neteventmsg.mc
```diff
@@ -1140,6 +1338,9 @@ Le service %1 ne s'est pas arrêté proprement après avoir reçu un contrôle d
 Language=Polish
 Usługa %1 nie została poprawnie wyłączona po otrzymaniu kodu sterownika przed zamknięciem systemu.
 .
+Language=Romanian
+Serviciul %1 nu a fost oprit corect după primirea controlului de preoprire.
```
reactos-contributor, line 1342: Another case of reflexive that may fit better than passive, since it's clear that what was expected (but didn't happen) for the given service to gracefully shut itself down.

sdk/include/reactos/mc/neteventmsg.mc
```diff
@@ -1161,6 +1362,9 @@ Language=French
 Language=Polish
 %2: nie można zarejestrować w otoce NDIS.
 .
+Language=Romanian
+%2-ul însuși nu a putut fi înregistrat cu emulatorul NDIS.
```
reactos-contributor, line 1366: A mere switch to reflexive (and removal of "însuși") in this case should both simplify and convey better the meaning of the English message: "%2 nu s-a putut înregistra..."

reactos-contributor (changes requested):

sdk/include/reactos/mc/neteventmsg.mc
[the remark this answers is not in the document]
reactos-contributor: My reason, like I mentioned above, was about the direction of the communication. "In/Ex" are short for "Interior/Exterior", and there are quite a few resources translated this way (like [machine.inf](https://github.com/reactos/reactos/blob/master/media/inf/machine.inf#L613), [cmd](https://github.com/reactos/reactos/blob/master/base/shell/cmd/lang/ro-RO.rc#L107), [dxdiag](https://github.com/reactos/reactos/blob/master/base/applications/dxdiag/lang/ro-RO.rc#L255), [taskmgr](https://github.com/reactos/reactos/blob/master/base/applications/taskmgr/lang/ro-RO.rc#L311), [devmgr](https://github.com/reactos/reactos/blob/master/dll/win32/devmgr/lang/ro-RO.rc#L42), [errcodes.mc](https://github.com/reactos/reactos/blob/master/sdk/include/reactos/mc/errcodes.mc#L1708), and probably more). Since you took on maintenance of ReactOS Romanian resources, now it's up to you to decide if it makes sense to you this way or you'd rather want to improve (be it with "In/Ie" or with something else). The important thing is (as you'd already guess) to keep things consistent across the body of translation work.

AndreiMiloiu:
Please look at lines 1002-1004. I don't know why it doesn't show it's outdated. I already corrected the mistakes. Also, shouldn't it be "In/Ie", instead of "In/Ex"? It seems to be more explicit."Ex" reffers to "Exit".

AndreiMiloiu:
Sorry, I mistakenly closed it. I reopened it now.

reactos-contributor:
> Please look at lines 1002-1004. I don't know why it doesn't show it's outdated. I already corrected the mistakes. Also, shouldn't it be "In/Ie", instead of "In/Ex"? It seems to be more explicit."Ex" reffers to "Exit".

https://github.com/reactos/reactos/pull/5431#discussion_r1263145270

sdk/include/reactos/mc/neteventmsg.mc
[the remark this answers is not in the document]
AndreiMiloiu: Yes, you're right. I completely forgot about it.

sdk/include/reactos/mc/neteventmsg.mc
```diff
@@ -213,6 +241,9 @@ ReactOS %1 %2 %3 %4.
 Language=Polish
 ReactOS %1 %2 %3 %4.
 .
+Language=Romanian
+ReactOS %1 %2 %3 %4.
```
reactos-contributor, line 245: Just a trivial note - I do not understand why was this string duplicated in other languages.

sdk/include/reactos/mc/neteventmsg.mc
```diff
@@ -230,6 +261,9 @@ Le service %1 n'a pas réussi à installer une source d'évènement.
 Language=Polish
 Usługa %1 nie mogła skonfigurować źródła zdarzenia.
 .
+Language=Romanian
+Serviciul %1 nu a putut configura o sursă a evenimentului.
```
reactos-contributor, line 265: English test says "an event", whereas the Romanian translates it as "the event".
...but maybe this is not that important.

sdk/include/reactos/mc/neteventmsg.mc
```diff
@@ -835,6 +971,9 @@ Le gestionnaire de contrôle des services a tenté de prendre une action correct
 Language=Polish
 Menedżer sterowania usługami próbował podjąć akcję korekcyjną (%2) po nieoczekiwanym zakończeniu usługi %3, ale ta akcja nie powiodła się przy następującym błędzie: %n%4.
 .
+Language=Romanian
+Administratorul de control al serviciului a încercat să ia o acțiune corectivă (%2) după o închidere neanunțată a serviciului %3, dar această acțiune a eșuat cu următoarea eroare: %n%4
```
reactos-contributor, line 975: In the previous translated resource, "unexpected" got translated as "neașteptat" whereas here is translated as "neanunțată" (which is less accurate). Please revise.

sdk/include/reactos/mc/neteventmsg.mc
```diff
@@ -1012,6 +1179,13 @@ proces %2, a połączył się proces %3.%n%n
 
 Jeśli ta usługa jest skonfigurowana do uruchomienia z debugerem, takie zachowanie jest oczekiwane.
 .
+Language=Romanian
+Procesul serviciului, altul decât ceea ce este pornit în Administratorul de control al serviciilor
+e conectat odată cu pornirea serviciului %1. Administratorul de control al serviciilor a lansat, în schimb
+procesul %2 și procesul %3 conectat.%n%n
```
reactos-contributor, line 1185: This is still confusing and needs rework. Some hints:
- on first sentence, English tells that "a service process" "connected" (itself), whereas Romanian says "the service process" "is connected"
- on second sentence, English states clearly that "(SCM) launched process %2" (to connect) but "process %3 connected (itself) instead", whereas the Romanian sentence does not convey this (as clearly)
- it also may help to use reflexive when it's about that service/process that is connecting itself instead of the expected one

sdk/include/reactos/mc/neteventmsg.mc
```diff
@@ -1098,6 +1275,22 @@ i prawo użytkownika wygląda na usunięte, obiekt zasad grupy powiązany z tym
 węzłem może usuwać to prawo. Skontaktuj się z administratorem domeny by sprawdzić,
 czy to się dokonuje.
 .
+Language=Romanian
+Serviciul %1 nu a putut fi logat ca %2 cu parola configurată în prezent, din cauza următoarei erori:%n
```
reactos-contributor, line 1279: Another omission here with "logat".

reactos-contributor (changes requested):
First two comments are not necessarily requesting change.

sdk/include/reactos/mc/neteventmsg.mc
```diff
@@ -859,6 +998,11 @@ Language=Polish
 Menedżer sterowania usługami nie został zainicjowany pomyślnie. Nie udało się zainicjować serwera
 konfiguracji zabezpieczeń (scesrv.dll) z powodu błędu %1. Trwa ponowne uruchamianie systemu...
 .
+Language=Romanian
+Administratorul de control al serviciului s-a putut inițializa cu succes. Securitatea
+servitorului de configurație (scesrv.dll) s-a putut inițializa cu eroarea %1.
```
reactos-contributor, line 1003: Also, at the moment of writing this comment, I see:
`Administratorul de control al serviciului s-a putut inițializa cu succes. Securitatea
servitorului de configurație (scesrv.dll) s-a putut inițializa cu eroarea %1.`
...which reports something positive, whereas the correct message should report a failure instead.

sdk/include/reactos/mc/neteventmsg.mc
```diff
@@ -262,7 +262,7 @@ Language=Polish
 Usługa %1 nie mogła skonfigurować źródła zdarzenia.
 .
 Language=Romanian
-Serviciul %1 nu a putut configura o sursă a evenimentului.
+Serviciul %1 nu a putut configura o sursă a unui eveniment.
```
reactos-contributor, line 265: To be frank, "sursă a unui eveniment" sounds worse than "sursă de eveniment" which [you've chosen before](https://github.com/reactos/reactos/pull/5431#discussion_r1262780104). The improvement I was hinting at was "sursă de evenimente".

sdk/include/reactos/mc/neteventmsg.mc
```diff
@@ -1180,9 +1180,9 @@ proces %2, a połączył się proces %3.%n%n
 Jeśli ta usługa jest skonfigurowana do uruchomienia z debugerem, takie zachowanie jest oczekiwane.
 .
 Language=Romanian
-Procesul serviciului, altul decât ceea ce este pornit în Administratorul de control al serviciilor
-e conectat odată cu pornirea serviciului %1. Administratorul de control al serviciilor a lansat, în schimb
-procesul %2 și procesul %3 conectat.%n%n
+Un proces al serviciului, altul decât ceea ce este pornit în Administratorul de control al serviciilor
```
reactos-contributor, line 1183: Here, the English "service process" is describing the type of the process, not that the belongs to a service. So here's a suggestion: "Un proces de tip serviciu" instead of "Un proces al serviciului".

reactos-contributor (approved):
Despite the remaining two notes, this looks good to me.

sdk/include/reactos/mc/neteventmsg.mc
AndreiMiloiu: If I don't duplicate this string, the file will crash on compiling. Hence, just some checks will pass, all the others will fail.

AndreiMiloiu:
I'll revise this file, too in order to limit every row to less than 80 characters.

reactos-contributor:
> Now as I know how to do it, I'll revise this file, too in order to limit every row to less than 80 characters.

There is no need to have this for **every row**. Lines get automatically wrapped into new lines most of the time, so the need to wrap them manually arises only when there are some other considerations involved. A rule of thumb is to look for the English text that you are translating, if it is or not limited to 80 characters.

> Is anything else to do?

To me it looks good enough as is (confirmed by the fact that I've approved it already).

AndreiMiloiu:
Good! I understand!

PR #5445, opened 2023-07-15
[BROWSEUI] Hide status bar in the Folders and Search panes when it's not enabled

cbialorucki:
## Purpose

Hides the status bar in the folder and search pane when status bar is hidden.

image

JIRA issue: [CORE-19027](https://jira.reactos.org/browse/CORE-19027)

## Proposed changes

- Do not create the status bar unless it is visible. Previously, the status bar was always created, then hidden if the status bar was hidden.
- Move the status bar creation code to a separate function.


dll/win32/browseui/shellbrowser.cpp
```diff
@@ -786,11 +787,10 @@ HRESULT CShellBrowser::Initialize()
...
 
-
-    // create status bar
-    fStatusBar = CreateWindow(STATUSCLASSNAMEW, NULL, WS_CHILD | WS_VISIBLE | WS_CLIPSIBLINGS |
-                    SBT_NOBORDERS | SBT_TOOLTIPS, 0, 0, 500, 20, m_hWnd, (HMENU)0xa001,
-                    _AtlBaseModule.GetModuleInstance(), 0);
+    if(fStatusBarVisible)
+    {
+        CreateStatusBar();
+    }
```
binarymaster, line 793:
```suggestion
    if (fStatusBarVisible)
        CreateStatusBar();
```

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -3659,11 +3659,12 @@ LRESULT CShellBrowser::OnOrganizeFavorites(WORD wNotifyCode, WORD wID, HWND hWnd
 LRESULT CShellBrowser::OnToggleStatusBarVisible(WORD wNotifyCode, WORD wID, HWND hWndCtl, BOOL &bHandled)
 {
     fStatusBarVisible = !fStatusBarVisible;
-    if (fStatusBar)
-    {
-        ::ShowWindow(fStatusBar, fStatusBarVisible ? SW_SHOW : SW_HIDE);
-        RepositionBars();
-    }
+    
+    if(fStatusBarVisible)
```
binarymaster, line 3663:
```suggestion
    if (fStatusBarVisible)
```

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -3676,6 +3677,17 @@ LRESULT CShellBrowser::OnToggleStatusBarVisible(WORD wNotifyCode, WORD wID, HWND
...
 
+LRESULT CShellBrowser::CreateStatusBar()
+{
+    if(!fStatusBar)
+    {
+        fStatusBar = CreateWindow(STATUSCLASSNAMEW, NULL, WS_CHILD | WS_VISIBLE | WS_CLIPSIBLINGS |
+                    SBT_NOBORDERS | SBT_TOOLTIPS, 0, 0, 500, 20, m_hWnd, (HMENU)0xa001,
+                    _AtlBaseModule.GetModuleInstance(), 0);
+    }
+    return 0;
```
binarymaster, line 3688:
```suggestion
    if (!fStatusBar)
    {
        fStatusBar = CreateWindow(STATUSCLASSNAMEW, NULL, WS_CHILD | WS_VISIBLE | WS_CLIPSIBLINGS |
                                  SBT_NOBORDERS | SBT_TOOLTIPS, 0, 0, 500, 20, m_hWnd, (HMENU)0xa001,
                                  _AtlBaseModule.GetModuleInstance(), 0);
    }
    return 0;
```

binarymaster (commented):
Formatting...

dll/win32/browseui/shellbrowser.cpp
cbialorucki: Absolutely :)

JoachimHenze (approved):

binarymaster (approved):

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -3676,6 +3675,17 @@ LRESULT CShellBrowser::OnToggleStatusBarVisible(WORD wNotifyCode, WORD wID, HWND
     return 0;
 }
 
+LRESULT CShellBrowser::CreateStatusBar()
+{
+    if (!fStatusBar)
+    {
+        fStatusBar = CreateWindow(STATUSCLASSNAMEW, NULL, WS_CHILD | WS_VISIBLE | WS_CLIPSIBLINGS |
+                                  SBT_NOBORDERS | SBT_TOOLTIPS, 0, 0, 500, 20, m_hWnd, (HMENU)0xa001,
```
HBelusca, line 3683: Do we know why "0xa001" needs to be used there?

HBelusca, line 3683: Not sure `SBT_NOBORDERS` is a valid statusbar window style. Also, replace `SBT_TOOLTIPS` by `SBARS_TOOLTIPS`.
It might be desirable to add `CCS_BOTTOM` too just to be sure.

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -3676,6 +3675,17 @@ LRESULT CShellBrowser::OnToggleStatusBarVisible(WORD wNotifyCode, WORD wID, HWND
     return 0;
 }
 
+LRESULT CShellBrowser::CreateStatusBar()
+{
+    if (!fStatusBar)
+    {
+        fStatusBar = CreateWindow(STATUSCLASSNAMEW, NULL, WS_CHILD | WS_VISIBLE | WS_CLIPSIBLINGS |
```
HBelusca, line 3682: You should be able to use here `::CreateWindowExW(0, STATUSCLASSNAMEW, NULL, etc.`.

dll/win32/browseui/shellbrowser.cpp
Doug-Lyons: Per https://github.com/wine-mirror/wine/blob/master/include/shlobj.h it shows:
```
/* undocumented toolbar items from stddlg's*/
#define FCIDM_TB_UPFOLDER       0xA001
#define FCIDM_TB_NEWFOLDER      0xA002
#define FCIDM_TB_SMALLICON      0xA003
#define FCIDM_TB_REPORTVIEW     0xA004
#define FCIDM_TB_DESKTOP        0xA005  /* FIXME */
````

Comment (binarymaster) [reply dll/win32/browseui/shellbrowser.cpp:3683]: https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=FCIDM_TB_UPFOLDER

Comment (cbialorucki) [reply dll/win32/browseui/shellbrowser.cpp:3683]: ```SBT_NOBORDERS``` is defined here: https://github.com/wine-mirror/wine/blob/master/include/commctrl.h#L294

I will work on your other suggestions though.

Comment (HBelusca) [reply dll/win32/browseui/shellbrowser.cpp:3683]: Yes of course it is defined, like any other random define out there. This doesn't mean it is a valid status bar window style, though.
For example, `SBT_NOBORDERS` is a style to be used together with the `SB_SETTEXT` message, as explained there:
https://learn.microsoft.com/fr-fr/windows/win32/controls/sb-settext

Comment (cbialorucki) [reply dll/win32/browseui/shellbrowser.cpp:3683]: Good call, ```SBT_NOBORDERS``` was a hack to add the size gripper on the status bar. I will replace it with the more appropriately named ```SBARS_SIZEGRIP```.
https://learn.microsoft.com/en-us/windows/win32/controls/status-bar-styles

File: dll/win32/browseui/shellbrowser.cpp [hunk 6]
```diff
@@ -3676,6 +3675,18 @@ LRESULT CShellBrowser::OnToggleStatusBarVisible(WORD wNotifyCode, WORD wID, HWND
...
+LRESULT CShellBrowser::CreateStatusBar()
+{
+    if (!fStatusBar)
+    {
+        fStatusBar = CreateWindowExW(0, STATUSCLASSNAMEW, NULL, WS_CHILD | WS_VISIBLE |
+                                     WS_CLIPSIBLINGS | SBARS_SIZEGRIP | SBARS_TOOLTIPS,
+                                     0, 0, 500, 20, m_hWnd, (HMENU)FCIDM_TB_UPFOLDER,
+                                     _AtlBaseModule.GetModuleInstance(), 0);
+    }
+    return 0;
```
SergeGautherie, line 3687: This `return` value looks useless.

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -3659,11 +3657,12 @@ LRESULT CShellBrowser::OnOrganizeFavorites(WORD wNotifyCode, WORD wID, HWND hWnd
 LRESULT CShellBrowser::OnToggleStatusBarVisible(WORD wNotifyCode, WORD wID, HWND hWndCtl, BOOL &bHandled)
 {
     fStatusBarVisible = !fStatusBarVisible;
-    if (fStatusBar)
```
SergeGautherie, line 3662: Why is this check not wanted anymore?

SergeGautherie (commented):
Would it make sense to destroy the bar even, instead of hiding it only?

dll/win32/browseui/shellbrowser.cpp
cbialorucki: Because there is no possible condition where ```fStatusBar``` would not be initialized.

If it is enabled on launch, then ```fStatusBar``` will be initialized and you can show or hide it when toggling it.

If it is disabled on launch, then it will be created before calling the show window function.

dll/win32/browseui/shellbrowser.cpp
cbialorucki: This return value is required to use the standard ```LRESULT``` return value. Using a plain void is bad practice in Win32.

cbialorucki:
> Would it make sense to destroy the bar even, instead of hiding it only?

Calling ```::DestroyWindow(fStatusBar)``` destroys the contents of the entire file explorer window, so I kept the existing behavior for now.

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -786,11 +787,8 @@ HRESULT CShellBrowser::Initialize()
 
     fToolbarProxy.Initialize(m_hWnd, clientBar);
 
-
-    // create status bar
-    fStatusBar = CreateWindow(STATUSCLASSNAMEW, NULL, WS_CHILD | WS_VISIBLE | WS_CLIPSIBLINGS |
-                    SBT_NOBORDERS | SBT_TOOLTIPS, 0, 0, 500, 20, m_hWnd, (HMENU)0xa001,
-                    _AtlBaseModule.GetModuleInstance(), 0);
+    if (fStatusBarVisible)
+        CreateStatusBar();
```
binarymaster, line 791: Looking at the Serge's review comments I'm thinking now, maybe keep the status bar window creation unconditional, but then immediately hide it, if it's not enabled? This way it would not be needed to introduce new function.

JoachimHenze: I would vote for keeping it like proposed by the author for 3 reasons:
1.) the current state has been tested already
2.) I don't think it is good design to create something and then immediately hide it. Might also be less efficient from a performance / and or rendering perspective
3.) The function is not exported via the header even, so I think it is the appropriate solution chosen: If you want to call a piece of code at two positions, have a function for it to not repeat yourself.

Avoiding 2 calls would be good, sure. If that would be possible somehow: Fine.
But create+hide seems too hacky to me.

HBelusca: That could make sense, and instead of "immediately hiding it", you could remove that `WS_VISIBLE` style passed in the window creation, so that the status bar is initially hidden. Then you call the ShowWindow suitably.

binarymaster: ^ just wanted to say this :)

dll/win32/browseui/shellbrowser.cpp
HBelusca: Alternatively the function could return a BOOL that is TRUE in case the status bar creation succeeded, and FALSE if not.

dll/win32/browseui/shellbrowser.cpp
binarymaster: Reduced the diff in 0a3b5d04 using the flags, tested - it works. 🙂

JoachimHenze: I admit: It is the better solution, since we had a flag for for visibility within aleady anyway. Solves also the quetion about the new funcs ignored/superfluous retval.

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -786,11 +786,13 @@ HRESULT CShellBrowser::Initialize()
...
     // create status bar
-    fStatusBar = CreateWindow(STATUSCLASSNAMEW, NULL, WS_CHILD | WS_VISIBLE | WS_CLIPSIBLINGS |
-                    SBT_NOBORDERS | SBT_TOOLTIPS, 0, 0, 500, 20, m_hWnd, (HMENU)0xa001,
-                    _AtlBaseModule.GetModuleInstance(), 0);
+    DWORD dwStatusStyle = WS_CHILD | WS_CLIPSIBLINGS | SBARS_SIZEGRIP | SBARS_TOOLTIPS;
+    if (fStatusBarVisible)
+        dwStatusStyle |= WS_VISIBLE;
+    fStatusBar = CreateWindowExW(0, STATUSCLASSNAMEW, NULL, dwStatusStyle,
+                                 0, 0, 500, 20, m_hWnd, (HMENU)FCIDM_TB_UPFOLDER,
+                                 _AtlBaseModule.GetModuleInstance(), 0);
```
binarymaster, line 795: Btw why CreateWindow got replaced by CreateWindowExW? Arguments look the same except the first one added 0.

JoachimHenze: That was a review request of Hermes. Look upwards.

binarymaster: Hermes' suggestion in https://github.com/reactos/reactos/pull/5445#discussion_r1264544598 was to use an in-class call (with `::` preceeding it), here I see just a normal call. I wonder if this changes behavior somehow.

HBelusca: As long as no inherited class (nor the current one) implement a CreateWindow(Ex) function, this would be ok, but since we are never sure, and if the expectation here is to call the actual user32 CreateWindow(Ex) function, then one would better use the scope operator on the global namespace. i.e. `::CreateWindowExW` in the actual example above.

dll/win32/browseui/shellbrowser.cpp
SergeGautherie: I still do not see the difference with existing code. Is the existing check useless (with regard to existing code), then?
If it is, then would it be useless at https://github.com/reactos/reactos/blob/0a3b5d04899a40fbfe77d31dc7cad9d42c5cbcfd/dll/win32/browseui/shellbrowser.cpp#L1444 too?

HBelusca: > Because there is no possible condition where fStatusBar would not be initialized.

Unless, due to a low resources (i.e. memory) condition, the status bar window creation failed.

dll/win32/browseui/shellbrowser.cpp
HBelusca, line 795:
```suggestion
    fStatusBar = ::CreateWindowExW(0, STATUSCLASSNAMEW, NULL, dwStatusStyle,
                                   0, 0, 500, 20, m_hWnd, (HMENU)FCIDM_TB_UPFOLDER,
                                   _AtlBaseModule.GetModuleInstance(), 0);
```

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -786,11 +786,13 @@ HRESULT CShellBrowser::Initialize()
...
 
-
     // create status bar
-    fStatusBar = CreateWindow(STATUSCLASSNAMEW, NULL, WS_CHILD | WS_VISIBLE | WS_CLIPSIBLINGS |
-                    SBT_NOBORDERS | SBT_TOOLTIPS, 0, 0, 500, 20, m_hWnd, (HMENU)0xa001,
-                    _AtlBaseModule.GetModuleInstance(), 0);
+    DWORD dwStatusStyle = WS_CHILD | WS_CLIPSIBLINGS | SBARS_SIZEGRIP | SBARS_TOOLTIPS;
+    if (fStatusBarVisible)
+        dwStatusStyle |= WS_VISIBLE;
+    fStatusBar = CreateWindowExW(0, STATUSCLASSNAMEW, NULL, dwStatusStyle,
```
HBelusca: Unrelated to your actual PR, but using "f" as prefix for fStatusBar is obviously very error-inducing, because this is a window handle that is being returned, not a flag or boolean value. So `hStatusBar` would be more correct there.
(And if this is also a member of the CShellBrowser class, it should have been "m_hStatusBar"... But I think this goes beyond the scope of this PR. This was just my 1-cent rant of the day xD)

dll/win32/browseui/shellbrowser.cpp
cbialorucki: I just pushed this change 😅

HBelusca: For note: the (HMENU) magic value we were mentioning in a previous comment is the window control ID corresponding to the toolbar. Not sure that `FCIDM_TB_UPFOLDER` (while this has the same numerical value as seen by Doug) actually helps in that regard. (TB == toobar? why "upfolder" while we are dealing with a status-bar, etc.)

dll/win32/browseui/shellbrowser.cpp
cbialorucki: Good point @HBelusca. I reverted this change.

dll/win32/browseui/shellbrowser.cpp
cbialorucki: You are right, this is incorrect. Let's make a new define for the status bar. Should I do something like ```#define IDC_STATUSBAR 0xA001```

HBelusca: `IDC_STATUSBAR` (since this is not an actual menu, but a control)

binarymaster: Since `0xA001` should be an identifier of a parent, shouldn't it be something else? I mean, we're creating a status bar, it couldn't be a child of `IDC_STATUSBAR`. 🤔

HBelusca: > Type: HMENU
>
A handle to a menu, **or specifies a child-window identifier**, depending on the window style. For an overlapped or pop-up window, hMenu identifies the menu to be used with the window; it can be NULL if the class menu is to be used. **For a child window, hMenu specifies the child-window identifier, an integer value used by a dialog box control** to notify its parent about events. The application determines the child-window identifier; it must be unique for all child windows with the same parent window.

In other words, the same kind of resource ID one uses in dialog resource files. See existing code in ReactOS.

dll/win32/browseui/shellbrowser.cpp
SergeGautherie: While here, would the 3 existing calls without an `fStatusBar` check need one?

HBelusca (approved):
Looks quite a reduced patch :+1: Minor thing is to use something else than the "FCIDM_TB_UPFOLDER" define (see the corresponding comments), but otherwise LGTM.

dll/win32/browseui/shellbrowser.cpp
binarymaster: Regarding `0xA001` constant, it's also used in other places of the same file, probably it needs to be updated in these places as well.

learn-more (approved):

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -788,9 +788,12 @@ HRESULT CShellBrowser::Initialize()
...
-                    SBT_NOBORDERS | SBT_TOOLTIPS, 0, 0, 500, 20, m_hWnd, (HMENU)0xa001,
-                    _AtlBaseModule.GetModuleInstance(), 0);
+    DWORD dwStatusStyle = WS_CHILD | WS_CLIPSIBLINGS | SBARS_SIZEGRIP | SBARS_TOOLTIPS;
+    if (fStatusBarVisible)
+        dwStatusStyle |= WS_VISIBLE;
+    fStatusBar = ::CreateWindowExW(0, STATUSCLASSNAMEW, NULL, dwStatusStyle,
+                                   0, 0, 500, 20, m_hWnd, (HMENU)FCIDM_TB_UPFOLDER,
+                                   _AtlBaseModule.GetModuleInstance(), 0);
 
     ShowWindow(SW_SHOWNORMAL);
```
learn-more, line 798: Is this not going to invalidate the missing WS_VISIBLE?

HBelusca: Doesn't this one act on the main window though? Or are you thinking, that by showing the main window it would force all the children (including the status bar) to become visible?

dll/win32/browseui/shellbrowser.cpp
binarymaster: Added a constant, now waiting for CI...

dll/win32/browseui/shellbrowser.cpp
learn-more: Right, I missed that.
Doesn't matter since it is merged anyway.

PR #5450, opened 2023-07-17
[ZIPFLDR][BOOTDATA] Support UTF-8 Zip packing

katahiromz:
## Purpose

Enable ReactOS to compress files in UTF-8 filenames. It will improve interoperability on the other codepages.

JIRA issue: [CORE-16668](https://jira.reactos.org/browse/CORE-16668)

## Overview

New ReactOS can zip/unzip files and folders in UTF-8 filenames. You can also choose the codepage of filenames to zip/unzip via the `ZipCodePage` / `UnZipCodePage` registry values on the registry key `HKEY_CURRENT_USER\Software\ReactOS`.
Windows 8 or later also support UTF-8 zipped folders. You can also use 3rd party software to zip/unzip in older Windows.

## Comparison

I created a Zip Folder in Chinese ReactOS and then, I extracted it in Japanese Win10.

BEFORE:
before
FAILED.

AFTER:
after
Successful.

## Proposed changes

- Use `<atlconv.h>` for string conversion.
- Use `zipOpenNewFileInZip4_64` instead of `zipOpenNewFileInZip3_64`, and then add `MINIZIP_UTF8_FLAG` flag.
- Set the filenames in UTF-8 by using `CP_UTF8` codepage.
- Codepage is user selectable via registry settings.

## TODO

- [x] Do tests.

### DIFF
--- a/dll/shellext/zipfldr/CZipCreator.cpp
+++ b/dll/shellext/zipfldr/CZipCreator.cpp
@@ -3,7 +3,7 @@
  * LICENSE:     GPL-2.0+ (https://spdx.org/licenses/GPL-2.0+)
  * PURPOSE:     Create a zip file
  * COPYRIGHT:   Copyright 2019 Mark Jansen (mark.jansen@reactos.org)
- *              Copyright 2019 Katayama Hirofumi MZ (katayama.hirofumi.mz@gmail.com)
+ *              Copyright 2019-2023 Katayama Hirofumi MZ (katayama.hirofumi.mz@gmail.com)
  */
 
 #include "precomp.h"
@@ -11,6 +11,8 @@
 #include "minizip/zip.h"
 #include "minizip/iowin32.h"
 #include <process.h>
+
+#define PACK_AS_UTF8_ZIP
 
 static CStringW DoGetZipName(PCWSTR filename)
 {
@@ -34,10 +36,10 @@ static CStringW DoGetZipName(PCWSTR filename)
     return ret;
 }
 
-static CStringA DoGetAnsiName(PCWSTR filename)
+static CStringA DoGetAnsiName(PCWSTR filename, UINT nCodePage)
 {
     CHAR buf[MAX_PATH];
-    WideCharToMultiByte(CP_ACP, 0, filename, -1, buf, _countof(buf), NULL, NULL);
+    WideCharToMultiByte(nCodePage, 0, filename, -1, buf, _countof(buf), NULL, NULL);
     return buf;
 }
 
@@ -65,7 +67,11 @@ DoGetNameInZip(const CStringW& basename, const CStringW& filename)
 
     ret.Replace(L'\\', L'/');
 
-    return DoGetAnsiName(ret);
+#ifdef PACK_AS_UTF8_ZIP
+    return DoGetAnsiName(ret, CP_UTF8);
+#else
+    return DoGetAnsiName(ret, CP_ACP);
+#endif
 }
 
 static BOOL
@@ -299,7 +305,7 @@ unsigned CZipCreatorImpl::JustDoIt()
         }
 
         CStringA strNameInZip = DoGetNameInZip(strBaseName, strFile);
-        err = zipOpenNewFileInZip3_64(zf,
+        err = zipOpenNewFileInZip4_64(zf,
                                       strNameInZip,
                                       &zi,
                                       NULL,
@@ -315,6 +321,12 @@ unsigned CZipCreatorImpl::JustDoIt()
                                       Z_DEFAULT_STRATEGY,
                                       password,
                                       crc,
+                                      36,
+#ifdef PACK_AS_UTF8_ZIP
+                                      MINIZIP_UTF8_FLAG,
+#else
+                                      0,
+#endif
                                       zip64);
         if (err)
         {
--- a/dll/shellext/zipfldr/CZipCreator.cpp
+++ b/dll/shellext/zipfldr/CZipCreator.cpp
@@ -44,7 +44,7 @@ static CStringW DoGetBaseName(PCWSTR filename)
 }
 
 static CStringA
-DoGetNameInZip(const CStringW& basename, const CStringW& filename, BOOL bUtf8)
+DoGetNameInZip(const CStringW& basename, const CStringW& filename, UINT nCodePage)
 {
     CStringW basenameI = basename, filenameI = filename;
     basenameI.MakeUpper();
@@ -58,10 +58,7 @@ DoGetNameInZip(const CStringW& basename, const CStringW& filename, BOOL bUtf8)
 
     ret.Replace(L'\\', L'/');
 
-    if (!bUtf8)
-        return CStringA(ret);
-
-    return CStringA(CW2AEX<MAX_PATH>(ret, CP_UTF8));
+    return CStringA(CW2AEX<MAX_PATH>(ret, nCodePage));
 }
 
 static BOOL
@@ -275,7 +272,7 @@ unsigned CZipCreatorImpl::JustDoIt()
 
     int err = 0;
     CStringW strTarget, strBaseName = DoGetBaseName(m_items[0]);
-    const BOOL bUtf8 = TRUE;
+    UINT uCodePage = GetZipCodePage(FALSE);
     for (INT iFile = 0; iFile < files.GetSize(); ++iFile)
     {
         const CStringW& strFile = files[iFile];
@@ -295,7 +292,7 @@ unsigned CZipCreatorImpl::JustDoIt()
             // TODO: crc = ...;
         }
 
-        CStringA strNameInZip = DoGetNameInZip(strBaseName, strFile, bUtf8);
+        CStringA strNameInZip = DoGetNameInZip(strBaseName, strFile, uCodePage);
         err = zipOpenNewFileInZip4_64(zf,
                                       strNameInZip,
                                       &zi,
@@ -313,7 +310,7 @@ unsigned CZipCreatorImpl::JustDoIt()
                                       password,
                                       crc,
                                       MINIZIP_COMPATIBLE_VERSION,
-                                      (bUtf8 ? MINIZIP_UTF8_FLAG : 0),
+                                      (uCodePage == CP_UTF8 ? MINIZIP_UTF8_FLAG : 0),
                                       zip64);
         if (err)
         {
--- a/dll/shellext/zipfldr/CZipCreator.cpp
+++ b/dll/shellext/zipfldr/CZipCreator.cpp
@@ -58,10 +58,10 @@ DoGetNameInZip(const CStringW& basename, const CStringW& filename, BOOL bUtf8)
 
     ret.Replace(L'\\', L'/');
 
-    if (bUtf8)
-        return CStringA(CW2AEX<MAX_PATH>(ret, CP_UTF8));
-    else
+    if (!bUtf8)
         return CStringA(ret);
+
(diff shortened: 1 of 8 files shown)


dll/shellext/zipfldr/CZipCreator.cpp
```diff
@@ -12,6 +12,8 @@
 #include "minizip/iowin32.h"
 #include <process.h>
 
+#define PACK_AS_UTF8_ZIP
```
learn-more, line 15: Why do we need this define?

katahiromz: Deleted.

dll/shellext/zipfldr/CZipCreator.cpp
```diff
@@ -65,7 +65,7 @@ DoGetNameInZip(const CStringW& basename, const CStringW& filename)
 
     ret.Replace(L'\\', L'/');
 
-    return DoGetAnsiName(ret);
+    return DoGetAnsiName(ret, CP_UTF8);
```
binarymaster, line 68: This function is for getting ANSI name of the file, and this change effectively making it always return UTF-8 name instad. I think this is rather should be configurable by the user.

binarymaster: ...configurable by the user, or at least detected in run-time by checking file names to be packed.

dll/shellext/zipfldr/CZipCreator.cpp
```diff
@@ -315,6 +315,8 @@ unsigned CZipCreatorImpl::JustDoIt()
                                       Z_DEFAULT_STRATEGY,
                                       password,
                                       crc,
+                                      36,
```
binarymaster, line 318: This needs a define somewhere.

katahiromz: Added `MINIZIP_COMPATIBLE_VERSION` macro.

dll/shellext/zipfldr/CZipCreator.cpp
```diff
@@ -279,6 +292,7 @@ unsigned CZipCreatorImpl::JustDoIt()
 
     int err = 0;
     CStringW strTarget, strBaseName = DoGetBaseName(m_items[0]);
+    BOOL bUtf8 = IsUtf8ZipEnabled();
```
Extravert-ir, line 295: Why would anyone need to disable this? Also the knowledge about this setting will be lost immediately

binarymaster: I think runtime detection would be better, e.g. if one of filenames cannot be converted to ANSI losslessly, UTF-8 needs to be enabled.

katahiromz: Configuration is deleted.

dll/shellext/zipfldr/CZipCreator.cpp
katahiromz: Configuration is deleted.

dll/shellext/zipfldr/CZipCreator.cpp
```diff
@@ -65,7 +65,7 @@ DoGetNameInZip(const CStringW& basename, const CStringW& filename)
 
     ret.Replace(L'\\', L'/');
 
-    return DoGetAnsiName(ret);
+    return DoGetAnsiName(ret, (bUtf8 ? CP_UTF8 : CP_ACP));
```
binarymaster, line 68:
```suggestion
    return DoEncodeName(ret, (bUtf8 ? CP_UTF8 : CP_ACP));
```

binarymaster (commented):
DoGetAnsiName -> DoEncodeName

HBelusca:
> DoGetAnsiName -> DoEncodeName

Or just "EncodeName" , since "Encode" is already describing the action.

dll/shellext/zipfldr/CZipCreator.cpp
katahiromz: Fixed as `EncodeName`.

dll/shellext/zipfldr/CZipCreator.cpp
```diff
@@ -34,10 +34,10 @@ static CStringW DoGetZipName(PCWSTR filename)
     return ret;
 }
 
-static CStringA DoGetAnsiName(PCWSTR filename)
+static CStringA DoGetAnsiName(PCWSTR filename, UINT nCodePage)
```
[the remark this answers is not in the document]
katahiromz: Fixed as `EncodeName`.

katahiromz:
Fixed.

JoachimHenze (approved):

dll/shellext/zipfldr/CZipCreator.cpp
```diff
@@ -12,7 +12,7 @@
 #include "minizip/iowin32.h"
 #include <process.h>
 
-static CStringW DoGetZipName(PCWSTR filename)
+static CStringW GetZipName(PCWSTR filename)
```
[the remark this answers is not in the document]
JoachimHenze: I do think the PRs actual changes are small enough to allow an exception here in this specific case.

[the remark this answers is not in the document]
katahiromz: Reverted some names.

katahiromz:
The test is successful...
after
(Japanese --> Chinese)

Let's go!

dll/shellext/zipfldr/CZipCreator.cpp
```diff
@@ -65,7 +58,10 @@ DoGetNameInZip(const CStringW& basename, const CStringW& filename)
 
     ret.Replace(L'\\', L'/');
 
-    return DoGetAnsiName(ret);
+    if (!bUtf8)
+        return CStringA(ret);
+
+    return CStringA(CW2AEX<MAX_PATH>(ret, CP_UTF8));
```
HBelusca, line 64: Just to recall, if the necessary internal buffer for the conversion happens to be larger than the default `MAX_PATH`, it would be reallocated with corret size?

dll/shellext/zipfldr/CZipCreator.cpp
```diff
@@ -279,6 +275,7 @@ unsigned CZipCreatorImpl::JustDoIt()
 
     int err = 0;
     CStringW strTarget, strBaseName = DoGetBaseName(m_items[0]);
+    const BOOL bUtf8 = TRUE;
```
HBelusca, line 278: Why not then just use directly "TRUE" whenever this is used?

dll/shellext/zipfldr/CZipCreator.cpp
```diff
@@ -315,6 +312,8 @@ unsigned CZipCreatorImpl::JustDoIt()
                                       Z_DEFAULT_STRATEGY,
                                       password,
                                       crc,
+                                      MINIZIP_COMPATIBLE_VERSION,
+                                      (bUtf8 ? MINIZIP_UTF8_FLAG : 0),
```
HBelusca, line 316: Seems here that bUtf8 would always be TRUE (first because you declared it as "const", and because DoGetNameInZip doesn't modify bUtf8 as reference...)

dll/shellext/zipfldr/CZipCreator.cpp
katahiromz: When the source string was too long, then It will use a heap memory.

dll/shellext/zipfldr/CZipCreator.cpp
katahiromz: In the future, some customers might want to use non-UTF-8. Just a memorial reason.

dll/shellext/zipfldr/CZipCreator.cpp
katahiromz: In the previous PR, some reviewer wanted to use non-UTF-8 zip. Just a memorial reason.

dll/shellext/zipfldr/CZipCreator.cpp
binarymaster: I'd suggest to loop through all files to be packed first, check if their names contain non-latin (non-ANSI) characters, and then enable UTF-8 only in this case.

katahiromz: A Latin character in the different codepage also causes mojibake.
A plain ASCII string has the same size of the UTF-8 string of it.

katahiromz: A Latin character in the different codepage also causes mojibake.
A plain ASCII string has the same size of the UTF-8 string.

binarymaster: I wonder how these UTF-8 zip files would look like in Windows XP/2003 zip folder.

katahiromz: You can check UTF-8 Zip Folders:
[新建文件夹.zip](https://github.com/reactos/reactos/files/12105683/default.zip)
[新しいフォルダ.zip](https://github.com/reactos/reactos/files/12105684/default.zip)
* Chinese and Japanese fonts required.

binarymaster: To avoid complexity with fonts, I tried this scenario:

- Create zip-file from ReactOS, with non-ANSI filenames (I used Cyrillic chars here)
- Open the same zip-file in Windows XP/2003, result is mojibake:

| ReactOS | Windows XP |
| --- | --- |
| image | image |

So the problem that I expected does happen. This supports the idea of registry key you had added earlier here.

katahiromz: Now the registry values are effective. Fixed.

katahiromz: Now codepage is user selectable.

dll/shellext/zipfldr/zipfldr.cpp
```diff
@@ -94,6 +87,28 @@ GetDefaultUserSendTo(PWSTR pszPath)
...
 
+UINT GetZipCodePage(BOOL bUnZip)
+{
+    WCHAR szValue[16];
+    DWORD dwType, cbValue = sizeof(szValue);
+    UINT nDefaultCodePage = (bUnZip ? CP_ACP : CP_UTF8);
+
+    LONG error = SHGetValueW(HKEY_CURRENT_USER, L"Software\\ReactOS",
+                             (bUnZip ? L"UnZipCodePage" : L"ZipCodePage"),
+                             &dwType, szValue, &cbValue);
```
binarymaster, line 98: I think unzip code page is not needed, because it will be automatically detected by the UTF-8 flag.

katahiromz: In default settings, it uses the `CP_ACP` codepage. `CP_ACP` is zero but the meaning of `CP_ACP` has many variation.

- The Japanese `CP_ACP` codepage and the Chinese `CP_ACP` codepage are different.
- The Japanese `CP_ACP` codepage and the Korean `CP_ACP` codepage are different.
- The Chinese `CP_ACP` codepage and the Russian `CP_ACP` codepage are different.

etc.

katahiromz: if you want to extract from a Chinese `CP_ACP` codepage Zip file in the Japanese system, this setting is useful.

binarymaster: Ah I see, this is to force the codepage on extraction when UTF-8 is not used, nice.

PR #5452, opened 2023-07-18
[BROWSEUI] Use details view by default

cbialorucki:
## Purpose
Use the details view by default in the file explorer, except in the Control Panel, My Computer, and Desktop folders. This follows the default behavior in Windows Vista+.

JIRA issue: [CORE-18904](https://jira.reactos.org/browse/CORE-18904)

## Proposed changes
- Use the icon view by default for My Computer, Control Panel, and Desktop
- Use the details view by default in all other folders

## TODO (in separate PRs)
- Call the "Start Navigation" sound event in ```CShellBrowser::BrowseToPIDL```
- Implement the "Display the full path in the title bar" option including setting persistence
- Properly implement ShellBags to allow the user to change the default view for each folder
- Implement ```Tiles``` view and set that view as the default for My Computer and Desktop

### DIFF
--- a/dll/win32/browseui/shellbrowser.cpp
+++ b/dll/win32/browseui/shellbrowser.cpp
@@ -815,8 +815,8 @@ HRESULT CShellBrowser::BrowseToPIDL(LPCITEMIDLIST pidl, long flags)
     if (FAILED_UNEXPECTEDLY(hResult))
         return hResult;
     IUnknown_GetClassID(newFolder, &clsid);
-    HasIconViewType = clsid == CLSID_MyComputer || clsid == CLSID_ControlPanel
-                      || clsid == CLSID_ShellDesktop;
+    HasIconViewType = clsid == CLSID_MyComputer || clsid == CLSID_ControlPanel ||
+                      clsid == CLSID_ShellDesktop;
 
     if (HasIconViewType)
         newFolderSettings.ViewMode = FVM_ICON;
--- a/dll/win32/browseui/shellbrowser.cpp
+++ b/dll/win32/browseui/shellbrowser.cpp
@@ -801,6 +801,20 @@ HRESULT CShellBrowser::Initialize()
     return S_OK;
 }
 
+BOOL _ILIsSpecialFolder(LPCITEMIDLIST pidl, int CSIDL)
+{
+    LPITEMIDLIST                            SpecialFolder;
+    CComPtr<IShellFolder>                   ShellFolder;
+    HRESULT                                 hr;
+
+    hr = SHBindToFolder(pidl, &ShellFolder);
+    if (FAILED_UNEXPECTEDLY(hr))
+        return false;
+    SHGetSpecialFolderLocation(0, CSIDL, &SpecialFolder);
+    hr = ShellFolder->CompareIDs(NULL, pidl, SpecialFolder);
+    return SUCCEEDED(hr) && HRESULT_CODE(hr) == 0;
+}
+
 HRESULT CShellBrowser::BrowseToPIDL(LPCITEMIDLIST pidl, long flags)
 {
     CComPtr<IShellFolder>                   newFolder;
@@ -813,7 +827,11 @@ HRESULT CShellBrowser::BrowseToPIDL(LPCITEMIDLIST pidl, long flags)
     if (FAILED_UNEXPECTEDLY(hResult))
         return hResult;
 
-    newFolderSettings.ViewMode = FVM_ICON;
+    if (_ILIsSpecialFolder(pidl, CSIDL_DRIVES) || _ILIsSpecialFolder(pidl, CSIDL_CONTROLS)
+        || _ILIsSpecialFolder(pidl, CSIDL_DESKTOP))
+        newFolderSettings.ViewMode = FVM_ICON;
+    else
+        newFolderSettings.ViewMode = FVM_DETAILS;
     newFolderSettings.fFlags = 0;
     hResult = BrowseToPath(newFolder, pidl, &newFolderSettings, flags);
     if (FAILED_UNEXPECTEDLY(hResult))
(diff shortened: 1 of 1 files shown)


dll/win32/browseui/shellbrowser.cpp
```diff
@@ -813,7 +827,11 @@ HRESULT CShellBrowser::BrowseToPIDL(LPCITEMIDLIST pidl, long flags)
     if (FAILED_UNEXPECTEDLY(hResult))
         return hResult;
 
-    newFolderSettings.ViewMode = FVM_ICON;
+    if (_ILIsSpecialFolder(pidl, CSIDL_DRIVES) || _ILIsSpecialFolder(pidl, CSIDL_CONTROLS)
+        || _ILIsSpecialFolder(pidl, CSIDL_DESKTOP))
```
binarymaster, line 831: I wonder if there's simpler way to check if the folder is special. 🤔

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -801,6 +801,20 @@ HRESULT CShellBrowser::Initialize()
     return S_OK;
 }
 
+BOOL _ILIsSpecialFolder(LPCITEMIDLIST pidl, int CSIDL)
```
learn-more, line 804: I doubt that we would need another copy of this hack.

dll/win32/browseui/shellbrowser.cpp
cbialorucki: In Wine, they compare the GUIDs to see if they are equal but getting the GUID from the PIDL was complicated as well. [Raymond Chen recommends using ```IShellFolder::CompareIDs```](https://stackoverflow.com/questions/3572220/recognize-windows-shell-special-folder-i-e-get-its-csidl-via-its-pidl-now-de#comment9256003_7624031) for this purpose so I opted to use that.

learn-more: I don't think that this is the way it should be determined what view this folder needs, it should probably be some property from the IShellFolder or one of its related IShellFolder2 or whatever

cbialorucki:
I refactored the folder detection code and I think it is much better than the original PR. Big thanks to @learn-more for pushing me to, well, learn more!

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -806,14 +806,28 @@ HRESULT CShellBrowser::BrowseToPIDL(LPCITEMIDLIST pidl, long flags)
...
     // also called by explorer band to navigate to new folder
     hResult = SHBindToFolder(pidl, &newFolder);
     if (FAILED_UNEXPECTEDLY(hResult))
         return hResult;
+    hResult = newFolder->QueryInterface(IID_PPV_ARG(IPersist, &pp));
+    if (SUCCEEDED(hResult))
+    {
+        CLSID pclsid;
+        hResult = pp->GetClassID(&pclsid);
+        pp->Release();
```
learn-more, line 822: There is a helper function for this: `IUnknown_GetClassID`, that also handles another case where `IPersistFolder` is exported, but `IPersist` is not.

dll/win32/browseui/shellbrowser.cpp
learn-more: There is an `GetViewModeAndIconSize` in the `IFolderView2` class, and I assume that this is what should be used here.
How that determines the values is something I don't know yet.

cbialorucki: ```IFolderView2``` is NT6+ https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nn-shobjidl_core-ifolderview2#requirements

I would love to get that implemented eventually, it just goes well beyond the scope of this PR.

dll/win32/browseui/shellbrowser.cpp
cbialorucki: Updated to use ```IUnknown_GetClassID```

JoachimHenze (approved):

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -806,14 +806,22 @@ HRESULT CShellBrowser::BrowseToPIDL(LPCITEMIDLIST pidl, long flags)
...
+    CLSID                                   clsid;
+    BOOL                                    HasIconViewType;
 
     // called by shell view to browse to new folder
     // also called by explorer band to navigate to new folder
     hResult = SHBindToFolder(pidl, &newFolder);
     if (FAILED_UNEXPECTEDLY(hResult))
         return hResult;
+    IUnknown_GetClassID(newFolder, &clsid);
+    HasIconViewType = clsid == CLSID_MyComputer || clsid == CLSID_ControlPanel ||
```
JoachimHenze, line 818: Personally I would have left out CLSID_MyComputer from this list, but since I don't want that PR to be delayed by a big discussion about that tiny detail, I added this only as a comment. I consider this PR to be a giant leap for our filebrowsers usability, although we all know that this will be just an interim solution until we will have the shellbags implemented properly. If you also think that `CLSID_MyComputer` should be excluded here with the aim to also make that use the details-view, please leave me a thumbs up.

cbialorucki: If anything, I thought that ```CLSID_ShellDesktop``` would be up for debate.

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -806,14 +806,22 @@ HRESULT CShellBrowser::BrowseToPIDL(LPCITEMIDLIST pidl, long flags)
...
     HRESULT                                 hResult;
+    CLSID                                   clsid;
+    BOOL                                    HasIconViewType;
 
     // called by shell view to browse to new folder
     // also called by explorer band to navigate to new folder
     hResult = SHBindToFolder(pidl, &newFolder);
     if (FAILED_UNEXPECTEDLY(hResult))
         return hResult;
+    IUnknown_GetClassID(newFolder, &clsid);
```
learn-more, line 817: @cbialorucki Would you mind adding a comment that this is most likely a hack to get it working?

cbialorucki: Added a hack/fix me comment

PR #5456, opened 2023-07-19
[APPWIZ] 'Create Shortcut' wizard: Accept arguments

katahiromz:
## Purpose
Allow command line as `Create Shortcut` wizard's target.
JIRA issue: [CORE-5866](https://jira.reactos.org/browse/CORE-5866)

## Proposed changes

- Add `szArguments` member to `CREATE_LINK_CONTEXT` structure.
- Use `IShellLink::SetArguments`.
- Split arguments from command line text by using `PathGetArgsW` and `PathRemoveArgsW`.
- Find the first parameter by using `PathFindOnPathExW` and `FindExecutableW` and use the found file path.

## TODO

- [x] Do tests.

## Comparison

Win2k3:
win2k3
The Win2k3 wizard accepts command line and creates shortcut files.

BEFORE:
before
The wizard won't accept command line.

after
The wizard accepts command line and creates shortcut files.

### DIFF
--- a/dll/cpl/appwiz/createlink.c
+++ b/dll/cpl/appwiz/createlink.c
@@ -14,6 +14,8 @@
 #include <commctrl.h>
 #include <shellapi.h>
 #include <strsafe.h>
+#include <shlwapi_undoc.h>
+#include <undocshell.h>
 
 BOOL
 IsShortcut(HKEY hKey)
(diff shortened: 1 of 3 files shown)


dll/cpl/appwiz/createlink.c
```diff
@@ -14,6 +14,7 @@
 #include <commctrl.h>
 #include <shellapi.h>
 #include <strsafe.h>
+#include <shlwapi_undoc.h>
```
JoachimHenze, line 17: Which specific function call does require that include?

katahiromz: It's for `PathFindOnPathExW`. Added a comment.

PR #5402, opened 2023-07-04
[GDI32_APITEST] Remove "todo_ros" for SetDIBitsToDevice tests that work now

Doug-Lyons:
## Purpose

_Remove some outdated 'todo_ros' qualifiers._

JIRA issue: [ROSTESTS-384](https://jira.reactos.org/browse/ROSTESTS-384)

## Proposed changes

_Remove 'todo_ros' in six (6) places._

Testbot results:
SetDIBitsToDevice_fix.patch JID65977 on top of 0.4.15-dev-6253-g7fafeb6

VBox: https://reactos.org/testman/compare.php?ids=88092,88095 LGTM
KVM:  https://reactos.org/testman/compare.php?ids=88091,88094 LGTM

PR #4401, opened 2022-03-21
[SHELL32] Add multiple status bar parts to the file browser

EncryptedData:
## Purpose

The file browser now has a status bar like Windows Server 2003 does. This includes the sizes of the files, as well as the location. In certain folders (Networks, Desktop folder, etc) the status bar is hidden.

JIRA issue: [CORE-17603](https://jira.reactos.org/browse/CORE-17603)

## Known Issue:
In Windows Server 2003 the multipart status bar does not show when in a Control Panel. In ReactOS I do not see a way to check for that like I can for the Desktop folder. That or ```_ILIsCPanelStruct()``` is the correct function and it doesn't work correctly.

## Screenshots
Screenshot_20220320_211302
Screenshot_20220320_211342
Screenshot_20220320_211401

PR #5457, opened 2023-07-19
[SHELL32] Communicate with GUI parts in non-main thread

katahiromz:
## Purpose

#5420 (https://github.com/reactos/reactos/commit/9a1487f3a8997e3ea10328749dada70222868e54) causes regression. This PR will fix it.

JIRA issue: [CORE-19044](https://jira.reactos.org/browse/CORE-19044)

after

## Proposed changes

- Use `InterlockedExchangePointer` to set the thread handle.
- Modify wait timeout as `3000` from `INFINITE`.
- Don't operate GUI parts from non-main thread. Use `PostMessage` and `SendMessageTimeout` to communicate with the GUI parts from non-main thread.

## TODO

- [x] Do small tests.
- [ ] Do big tests.

### DIFF
--- a/dll/win32/shell32/CDefView.cpp
+++ b/dll/win32/shell32/CDefView.cpp
@@ -558,12 +558,12 @@ void CDefView::UpdateStatusbarWorker(HANDLE hThread)
         m_dwTotalSize = 0; // Don't calculate file size
         m_bIsOnlyFoldersSelected = false;
 
-        // It's a non-main thread. To communicate GUI parts, use posting message
+        // Here is in a non-main thread. To communicate GUI parts, use posting message
         PostMessage(SHV_UPDATE_STATUSBAR, 0, 0);
         return;
     }
 
-    // Sending message from a non-main thread to get the number of items
+    // Send a message from a non-main thread to get the number of items
     DWORD_PTR dwResult = 0;
     if (!::SendMessageTimeoutW(m_ListView, LVM_GETSELECTEDCOUNT, 0, 0,
                                SMTO_ABORTIFHUNG | SMTO_BLOCK, 500, &dwResult))
@@ -582,7 +582,7 @@ void CDefView::UpdateStatusbarWorker(HANDLE hThread)
         if (hThread != m_hUpdateStatusbarThread)
             return;
 
-        // Sending message from a non-main thread to get the next item
+        // Send a message from a non-main thread to get the next item
         dwResult = (DWORD)-1;
         if (!::SendMessageTimeoutW(m_ListView, LVM_GETNEXTITEM, nItem, uFileFlags,
                                    SMTO_ABORTIFHUNG | SMTO_BLOCK, 500, &dwResult))
@@ -605,7 +605,7 @@ void CDefView::UpdateStatusbarWorker(HANDLE hThread)
     m_dwTotalSize = uTotalFileSize;
     m_bIsOnlyFoldersSelected = bIsOnlyFoldersSelected;
 
-    // Here is a non-main thread. To communicate GUI parts, use posting message
+    // Here is in a non-main thread. To communicate GUI parts, use posting message
     PostMessage(SHV_UPDATE_STATUSBAR, 0, 0);
 }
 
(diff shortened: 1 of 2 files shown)


dll/win32/shell32/CDefView.cpp
```diff
@@ -2616,11 +2667,11 @@ HRESULT WINAPI CDefView::DestroyViewWindow()
...
     if (m_hUpdateStatusbarThread)
     {
-        HANDLE hOldThread = m_hUpdateStatusbarThread;
+        HANDLE hOldThread = NULL;
 
         // Assigning NULL to m_hUpdateStatusbarThread will terminate the target thread
-        m_hUpdateStatusbarThread = NULL;
-        ::WaitForSingleObject(hOldThread, INFINITE);
+        hOldThread = ::InterlockedExchangePointer(&m_hUpdateStatusbarThread, hOldThread);
+        ::WaitForSingleObject(hOldThread, 3000);
```
learn-more, line 2674: `if (hOldThread)`

dll/win32/shell32/CDefView.cpp
```diff
@@ -2616,11 +2667,11 @@ HRESULT WINAPI CDefView::DestroyViewWindow()
 
     if (m_hUpdateStatusbarThread)
     {
-        HANDLE hOldThread = m_hUpdateStatusbarThread;
+        HANDLE hOldThread = NULL;
 
         // Assigning NULL to m_hUpdateStatusbarThread will terminate the target thread
-        m_hUpdateStatusbarThread = NULL;
-        ::WaitForSingleObject(hOldThread, INFINITE);
+        hOldThread = ::InterlockedExchangePointer(&m_hUpdateStatusbarThread, hOldThread);
```
learn-more, line 2673: For portability you should use interlocked functions at all points where this is accessed, since you imply here that there could be race conditions

dll/win32/shell32/CDefView.cpp
```diff
@@ -2340,6 +2342,55 @@ static BOOL ILIsParentOrSpecialParent(PCIDLIST_ABSOLUTE pidl1, PCIDLIST_ABSOLUTE
     return FALSE;
 }
 
+LRESULT CDefView::OnUpdateStatusBar(UINT uMsg, WPARAM wParam, LPARAM lParam, BOOL &bHandled)
+{
+    // Here is in the main thread. We can operate GUI parts directly
+    LRESULT lResult;
+    WCHAR szFormat[MAX_PATH], szPartText[MAX_PATH];
```
learn-more, line 2349: Why don't you use CString's ?

katahiromz: Fixed.

dll/win32/shell32/CDefView.cpp
katahiromz:
```cxx
    CDefView* pView = static_cast<CDefView*>(args);
    pView->UpdateStatusbarWorker(pView->m_hUpdateStatusbarThread);
    pView->Release();
```

How can I fix it?

dll/win32/shell32/CDefView.cpp
katahiromz: Fixed.

yagoulas:
I'm starting to believe that the root of the problem is PR #4401 . The more stuff like that we add to the CDefView, the more chaotic it becomes.  In fact I don't like PR #4401, PR #5402 nor this one. Initially we said [here](https://github.com/reactos/reactos/pull/4401#issuecomment-1080334480) that we shouldn't try to be perfect from the get go but seeing this, with more and more problems piling up, it is obvious that this is getting out of hand. Now I would prefer a proper solution and no more hacks on top of hacks. 

Let me repeat: the CDefView class should not care about what kind of stuff it shows. It should be completely indifferent to the contents. Otherwise it will quickly fill with hacks for 20 or 30 different kinds of folders here and there and will be a mess.

@katahiro This all started with PR #4401. The reason I'm against these changes is that they build upon the wrong start that was done with that. I think it is time we fix that. I find it silly that we started adding hacks here and there in the shell for such a minor feature of the shell, the useless statusbar.

@learn-more thoughts?

katahiromz:
Closed due to wrong implementation.

Doug-Lyons:
Should the reference to PR #5402 have been PR#5420 instead?

learn-more:
probably

PR #5459, opened 2023-07-19
[NTOS:KE] Implement KeFreezeExecution and KeThawExecution for SMP

DarkFire01:
## Purpose

Provide the functionality needed to get every other processor in the system to freeze and thaw when requested.
This PR requires one of the SMP HAL PRs to be merged and my IPI work which ill submit in separate request to work. I'll try to get that done and I'll updated the description here when I have time.

## Proposed changes
- Create KiFreezeTargetExecution which holds all the other cores in a specific state
- Implement the ability to freeze and thaw processors using the correct PRCB elements so it shows up in Windbg with the !ipi command.

## TODO

- [x]  Check how windows uses KiFreezeTargetExecution to handle context switching (Will be PRed at a later date for simplicity of just getting this done)
- [x]  Check to make sure this will work on x64 when that has it's bootup code implemented.

### DIFF
--- a/ntoskrnl/ke/freeze.c
+++ b/ntoskrnl/ke/freeze.c
@@ -15,9 +15,37 @@
 
 /* GLOBALS ********************************************************************/
 
+extern KSPIN_LOCK KiFreezeExecutionLock;
+
 /* Freeze data */
 KIRQL KiOldIrql;
 ULONG KiFreezeFlag;
+
+/* PRIVATE FUNCTIONS ***********************************************************/
+
+VOID
+NTAPI
+KiFreezeTargetExecution(_In_ PKTRAP_FRAME TrapFrame,
+                        _In_ PKEXCEPTION_FRAME ExceptionFrame)
+{
+    PKPRCB Prcb;
+
+    Prcb = KeGetCurrentPrcb();
+    Prcb->IpiFrozen = IPI_FROZEN_HALTED;
+
+    /* Wait for triggering AP to give the go ahead to thaw */
+    while (Prcb->IpiFrozen == IPI_FROZEN_HALTED)
+    {
+        YieldProcessor();
+        KeMemoryBarrier();
+    }
+
+    /* Cleanup CPU caches */
+    KeFlushCurrentTb();
+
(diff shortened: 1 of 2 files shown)


ntoskrnl/ke/freeze.c
```diff
@@ -49,7 +80,27 @@ KeFreezeExecution(IN PKTRAP_FRAME TrapFrame,
...
+    TargetAffinity = KeActiveProcessors;
+    TargetAffinity &= ~Prcb->SetMember;
+    if (TargetAffinity)
+    {
+        for (i = 0, Current = 1; i < KeNumberProcessors; i++, Current <<= 1)
+        {
+            if (TargetAffinity & Current)
+            {
+                /* stop target processor */
+                KiIpiSend(Current, IPI_FREEZE);
```
tkreuzer, line 93: Why not send all IPIs at once?

DarkFire01: To verify each processor updated its field before moving on.

But you asking this has made me realize if I want to verify that I can just parse through the list of processors and check instead of sending one IPI at a time and checking.

ntoskrnl/ke/freeze.c
```diff
@@ -64,7 +115,31 @@ NTAPI
...
+
+    /* Loop through every processor */
+    for (i = 0, Current = 1; i < KeNumberProcessors; i++, Current <<= 1)
+    {
+        if (TargetAffinity & Current)
+        {
+            TargetPrcb = KiProcessorBlock[i];
+
+            /* Multiple processors can write this value */
+            InterlockedExchange((LONG*)&TargetPrcb->IpiFrozen, IPI_FROZEN_THAWING);
```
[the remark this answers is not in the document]
DarkFire01: working on these today;

If this assert in theory hits wouldn't it cause a problem since it requires a debug break yet we're already thawing from one?
If I understand correctly we're going to hold KiFreezeExecutionLock still at this point, only to unlock after this is completed.

DarkFire01:
Super seeded, will address comments in new PR.

PR #5462, opened 2023-07-20
[SHELL32] Rename file on copy-paste path collision

katahiromz:
## Purpose

JIRA issue: [CORE-18599](https://jira.reactos.org/browse/CORE-18599)

## Proposed changes

- If the source and destination of copying file are on the same directory, then enable `FOF_RENAMEONCOLLISION` flag. End if.
- Don't show error message `IDS_COPYERRORSUBFOLDER` or `IDS_MOVEERRORSUBFOLDER` if the source and the destination are the same.

## Comparison

Do `Ctrl+C` and `Ctrl+V` on same file.

BEFORE:
before

AFTER:
after

## TODO

- [x] Do tests.

### DIFF
--- a/dll/win32/shell32/droptargets/CFSDropTarget.cpp
+++ b/dll/win32/shell32/droptargets/CFSDropTarget.cpp
@@ -42,6 +42,7 @@ HRESULT CFSDropTarget::_CopyItems(IShellFolder * pSFFrom, UINT cidl,
     PWCHAR pwszListPos = pwszSrcPathsList;
     STRRET strretFrom;
     SHFILEOPSTRUCTW fop;
+    BOOL bRenameOnCollision = FALSE;
 
     /* Build a double null terminated list of C strings from source paths */
     for (UINT i = 0; i < cidl; i++)
@@ -64,7 +65,18 @@ HRESULT CFSDropTarget::_CopyItems(IShellFolder * pSFFrom, UINT cidl,
     if (FAILED(ret))
         goto cleanup;
 
-    wszDstPath[lstrlenW(wszDstPath) + 1] = L'\0';
+    wszDstPath[lstrlenW(wszDstPath) + 1] = UNICODE_NULL;
+
+    /* Set TRUE to bRenameOnCollision if necesssary */
+    if (bCopy)
+    {
+        WCHAR szPath1[MAX_PATH], szPath2[MAX_PATH];
+        GetFullPathNameW(pwszSrcPathsList, _countof(szPath1), szPath1, NULL);
+        GetFullPathNameW(wszDstPath, _countof(szPath2), szPath2, NULL);
+        PathRemoveFileSpecW(szPath1);
+        if (_wcsicmp(szPath1, szPath2) == 0)
+            bRenameOnCollision = TRUE;
+    }
 
     ZeroMemory(&fop, sizeof(fop));
     fop.hwnd = m_hwndSite;
@@ -72,6 +84,9 @@ HRESULT CFSDropTarget::_CopyItems(IShellFolder * pSFFrom, UINT cidl,
     fop.pFrom = pwszSrcPathsList;
     fop.pTo = wszDstPath;
     fop.fFlags = FOF_ALLOWUNDO | FOF_NOCONFIRMMKDIR;
+    if (bRenameOnCollision)
+        fop.fFlags |= FOF_RENAMEONCOLLISION;
+
     ret = S_OK;
 
     if (SHFileOperationW(&fop))
--- a/dll/win32/shell32/droptargets/CFSDropTarget.cpp
+++ b/dll/win32/shell32/droptargets/CFSDropTarget.cpp
@@ -67,7 +67,7 @@ HRESULT CFSDropTarget::_CopyItems(IShellFolder * pSFFrom, UINT cidl,
 
     wszDstPath[lstrlenW(wszDstPath) + 1] = UNICODE_NULL;
 
-    /* Set TRUE to bRenameOnCollision if necesssary */
+    /* Set bRenameOnCollision to TRUE if necesssary */
     if (bCopy)
     {
         WCHAR szPath1[MAX_PATH], szPath2[MAX_PATH];
--- a/dll/win32/shell32/shlfileop.cpp
+++ b/dll/win32/shell32/shlfileop.cpp
@@ -1940,11 +1940,7 @@ validate_operation(LPSHFILEOPSTRUCTW lpFileOp, FILE_LIST *flFrom, FILE_LIST *flT
             {
                 size_t cchFrom = PathAddBackslashW(szFrom) - szFrom;
                 size_t cchTo = PathAddBackslashW(szTo) - szTo;
-#ifdef __REACTOS__
                 if (cchFrom < cchTo)
-#else
-                if (cchFrom <= cchTo)
-#endif
                 {
                     WCHAR ch = szTo[cchFrom];
                     szTo[cchFrom] = 0;
--- a/dll/win32/shell32/shlfileop.cpp
+++ b/dll/win32/shell32/shlfileop.cpp
@@ -1940,7 +1940,11 @@ validate_operation(LPSHFILEOPSTRUCTW lpFileOp, FILE_LIST *flFrom, FILE_LIST *flT
             {
                 size_t cchFrom = PathAddBackslashW(szFrom) - szFrom;
                 size_t cchTo = PathAddBackslashW(szTo) - szTo;
+#ifdef __REACTOS__
+                if (cchFrom < cchTo)
+#else
                 if (cchFrom <= cchTo)
+#endif
                 {
                     WCHAR ch = szTo[cchFrom];
                     szTo[cchFrom] = 0;


binarymaster:
Question, does this PR fix this scenario?

- Go to "My Documents"
- Create an empty text file
- Press Ctrl+C and Ctrl+V
- Expected result: A new copy of empty text file

katahiromz:
@binarymaster Yes, exactly.

dll/win32/shell32/droptargets/CFSDropTarget.cpp
```diff
@@ -83,12 +83,26 @@ HRESULT CFSDropTarget::_CopyItems(IShellFolder * pSFFrom, UINT cidl,
...
 
+    BOOL bRenameOnCollision = FALSE;
+    if (bCopy)
+    {
+        WCHAR szPath1[MAX_PATH], szPath2[MAX_PATH];
+        lstrcpynW(szPath1, pszSrcList, _countof(szPath1));
+        PathRemoveFileSpecW(szPath1);
+        lstrcpynW(szPath2, m_sPathTarget, _countof(szPath2));
+        PathRemoveFileSpecW(szPath2);
+        if (lstrcmpiW(szPath1, szPath2) == 0)
```
HBelusca, line 94: Maybe the following may be more efficient and less stack-space-consuming:
```suggestion
        PCWSTR pszFileName1 = PathFindFileNameW(pszSrcList);
        PCWSTR pszFileName2 = PathFindFileNameW(m_sPathTarget);
        if (lstrcmpiW(pszFileName1, pszFileName2) == 0)
```

katahiromz: Did you think the case that two directories are different and two file titles are the same?

HBelusca, line 94: Oh i misread the code, sorry. You are _**keeping**_ the directories in the paths, not the file names. Please disregard my above suggestion. It may be possible to find where the file name specification currently starts in both paths (keeping those `PathFindFileNameW` calls), and do the string comparison for the substring that starts from the beginning and up to the file part (using lstrncmpiW or sthg):

```suggestion
        PCWSTR pszFileName1 = PathFindFileNameW(pszSrcList);
        PCWSTR pszFileName2 = PathFindFileNameW(m_sPathTarget);
        SIZE_T cchPath1 = pszFileName1 - pszSrcList;
        SIZE_T cchPath2 = pszFileName2 - m_sPathTarget;

        if ((cchPath1 == cchPath2) && lstrncmpiW(szPath1, szPath2) == 0)
```
_**(NOTE: This would require to copy-paste this helper function from https://git.reactos.org/?p=reactos.git;a=blob;f=dll/win32/comctl32/listview.c;hb=0e8cf6ffd58cc7a46db8bfe46cfb361aceb15153#l545  since lstrncmpiW is not an existing export... -- That is, if you want to really use those `lstr*W` functions. Otherwise, using `_wcsnicmp` might be sufficient.)**_

Also, I've the separate question: Are the paths, at this stage, already normalized, i.e. that for example:
`"C:\Some\Path\..\where\file.txt"` has already become `"C:\Some\where\file.txt"` ?

katahiromz: > Also, I've the separate question: Are the paths, at this stage, already normalized, i.e. that for example: "C:\Some\Path\..\where\file.txt" has already become "C:\Some\where\file.txt" ?

```txt
CDefaultContextMenu::InvokeCommand
    --> CDefaultContextMenu::DoPaste
        --> shlwapi!SHSimulateDrop
            --> CFSDropTarget::Drop
                --> CFSDropTarget::_DoDrop
                    --> CFSDropTarget::_CopyItems
```

`DoPaste` comes from clipboard. The PIDL of `DoPaste` can be a relative PIDL.

dll/win32/shell32/droptargets/CFSDropTarget.cpp
```diff
@@ -83,12 +83,27 @@ HRESULT CFSDropTarget::_CopyItems(IShellFolder * pSFFrom, UINT cidl,
...
+    }
+
     SHFILEOPSTRUCTW op = {0};
     op.pFrom = pszSrcList;
     op.pTo = wszTargetPath;
     op.hwnd = m_hwndSite;
     op.wFunc = bCopy ? FO_COPY : FO_MOVE;
     op.fFlags = FOF_ALLOWUNDO | FOF_NOCONFIRMMKDIR;
+    if (bRenameOnCollision)
+        op.fFlags |= FOF_RENAMEONCOLLISION;
```
HBelusca, line 106: I wonder now: What would happen if the `FOF_RENAMEONCOLLISION` flag is *always* set in the `op.fFlags` (when `bCopy` is TRUE) ? Would this give a correct behaviour ?

dll/win32/shell32/droptargets/CFSDropTarget.cpp
```diff
@@ -83,12 +83,27 @@ HRESULT CFSDropTarget::_CopyItems(IShellFolder * pSFFrom, UINT cidl,
...
+            bRenameOnCollision = TRUE;
+    }
+
     SHFILEOPSTRUCTW op = {0};
     op.pFrom = pszSrcList;
     op.pTo = wszTargetPath;
     op.hwnd = m_hwndSite;
     op.wFunc = bCopy ? FO_COPY : FO_MOVE;
     op.fFlags = FOF_ALLOWUNDO | FOF_NOCONFIRMMKDIR;
+    if (bRenameOnCollision)
```
HBelusca, line 105: The question from above would be equivalent to removing the whole added code block from above, and replacing this condition by just `if (bCopy)  op.fFlags |= FOF_RENAMEONCOLLISION;` .

dll/win32/shell32/droptargets/CFSDropTarget.cpp
katahiromz: > Would this still give a correct behaviour?

Nope. Copying from another directory shouldn't rename automatically.

dll/win32/shell32/droptargets/CFSDropTarget.cpp
katahiromz: Copying from another directory shouldn't rename automatically. It should confirm overwrite.

dll/win32/shell32/droptargets/CFSDropTarget.cpp
```diff
@@ -83,12 +83,27 @@ HRESULT CFSDropTarget::_CopyItems(IShellFolder * pSFFrom, UINT cidl,
     if (!pszSrcList)
         return E_OUTOFMEMORY;
 
+    BOOL bRenameOnCollision = FALSE;
+    if (bCopy)
```
HBelusca, line 87: It may be worth checking whether you are copying only one item or more. Because this `_CopyItems` seems to cover the case of many files, it would seem so. (Please correct me if I'm wrong.)

dll/win32/shell32/droptargets/CFSDropTarget.cpp
HBelusca: OK (iirc this overwrite confirmation should happen also in the many-copied-files case?)

katahiromz: > should happen also in the many-copied-files case?)

Yep.

dll/win32/shell32/droptargets/CFSDropTarget.cpp
katahiromz: `SHFileOperationW` should cover multiple source files.

katahiromz: Now the tests on multiple files successful:

after

katahiromz:
Please wait for getting #5481 merged.

katahiromz:
Okey, let's restart.

katahiromz:
after
Successfully working...!

dll/win32/shell32/shlfileop.cpp
```diff
@@ -1940,7 +1940,11 @@ validate_operation(LPSHFILEOPSTRUCTW lpFileOp, FILE_LIST *flFrom, FILE_LIST *flT
             {
                 size_t cchFrom = PathAddBackslashW(szFrom) - szFrom;
                 size_t cchTo = PathAddBackslashW(szTo) - szTo;
+#ifdef __REACTOS__
```
HBelusca, line 1943: This `validate_operation` function is a ReactOS-specific function already (see above, already surrounded by the guards), that you have introduced in this commit https://github.com/reactos/reactos/commit/0e2bb5b0b80edac9e1078388c9d0b60fba4a55e1 .
So if what you're doing there is a bugfix, you can directly fix it without adding more REACTOS guards.

dll/win32/shell32/droptargets/CFSDropTarget.cpp
```diff
@@ -64,14 +65,28 @@ HRESULT CFSDropTarget::_CopyItems(IShellFolder * pSFFrom, UINT cidl,
     if (FAILED(ret))
         goto cleanup;
 
-    wszDstPath[lstrlenW(wszDstPath) + 1] = L'\0';
+    wszDstPath[lstrlenW(wszDstPath) + 1] = UNICODE_NULL;
+
+    /* Set TRUE to bRenameOnCollision if necesssary */
```
HBelusca, line 70:
```suggestion
    /* Set bRenameOnCollision to TRUE if necesssary */
```

dll/win32/shell32/droptargets/CFSDropTarget.cpp
```diff
@@ -64,14 +65,28 @@ HRESULT CFSDropTarget::_CopyItems(IShellFolder * pSFFrom, UINT cidl,
...
+    if (bCopy)
+    {
+        WCHAR szPath1[MAX_PATH], szPath2[MAX_PATH];
+        GetFullPathNameW(pwszSrcPathsList, _countof(szPath1), szPath1, NULL);
+        GetFullPathNameW(wszDstPath, _countof(szPath2), szPath2, NULL);
+        PathRemoveFileSpecW(szPath1);
```
HBelusca, line 76: I don't remember whether this has already been a past question: `szPath2` doesn't need to have a PathRemoveFileSpecW call on it too?

dll/win32/shell32/shlfileop.cpp
katahiromz: Removed that guard.

dll/win32/shell32/droptargets/CFSDropTarget.cpp
katahiromz: Fixed that comment.

dll/win32/shell32/droptargets/CFSDropTarget.cpp
katahiromz: `wszDstPath` is already a directory.

PR #5466, opened 2023-07-20
[SDK:RTL] Don't overwrite Source UString when writing to Destination UString.

zecarlos1957:
## Purpose

Fixes the buggy observed behaviour described below, causing the crash described in the following JIRA issue.

JIRA issue: [CORE-18988](https://jira.reactos.org/browse/CORE-18988)

## Observed behavior:

When ScmAutoStartService  call RtlQueryRegistryValues, the code reserve 4096 bytes to the KEY_VALUE_FULL_INFORMATION KeyValueInfo structure.
After a call to ZwQueryValueKey, the Data member of KeyValueInfo give us a string with the ImagePath to  system folder that contains the respective service.
This string contain a environment variable that needs to be expanded with a call to RtlExpandEnvironmentStrings.
The caller provides three UNICODE_STRING, that is Environment, Source and Destination.
The Environment is NULL, (that is ok).
The Source.Buffer is set to &KeyValueInfo+DataOffset. (That is correct).
The Destination.Buffer is set to SpareData pointer that point to &KeyValueInfo address  (Problems  here).

When RtlExpandEnvironmentStrings starts writing to the Destination buffer, it will overwrite not only the Source UString but also the entire KeyValueInfo structure.
This can be proven by looking at the attached DPrint1.txt which shows the address of both Strings as well as the final result on the Source UString.

## Proposed changes

In the `RtlpCallQueryRegistryRoutine()` helper function (called by `RtlQueryRegistryValues()`), `SpareData` and `SpareLength` need to be set correctly, as they are used later on as well.

### DIFF
--- a/base/setup/usetup/lang/bg-BG.h
+++ b/base/setup/usetup/lang/bg-BG.h
@@ -2071,9 +2071,7 @@ MUI_ERROR bgBGErrorEntries[] =
         // ERROR_DIRECTORY_NAME,
         "\215\245\242\240\253\250\244\255\256 \250\254\245 \255\240 \244\250\340\245\252\342\256\340\250\357\342\240.\n"
         "\n"
-        "The maximum length for the directory name is 31 characters.\n"
-        "\n"
-        "  * Press Enter to continue."
+        "  * \215\240\342\250\341\255\245\342\245 \257\340\256\250\247\242\256\253\245\255 \252\253\240\242\250\350, \247\240 \244\240 \257\340\256\244\352\253\246\250\342\245."
     },
     {
         // ERROR_INSUFFICIENT_PARTITION_SIZE,
--- a/base/setup/usetup/lang/bg-BG.h
+++ b/base/setup/usetup/lang/bg-BG.h
@@ -2071,7 +2071,9 @@ MUI_ERROR bgBGErrorEntries[] =
         // ERROR_DIRECTORY_NAME,
         "\215\245\242\240\253\250\244\255\256 \250\254\245 \255\240 \244\250\340\245\252\342\256\340\250\357\342\240.\n"
         "\n"
-        "  * \215\240\342\250\341\255\245\342\245 \257\340\256\250\247\242\256\253\245\255 \252\253\240\242\250\350, \247\240 \244\240 \257\340\256\244\352\253\246\250\342\245."
+        "The maximum length for the directory name is 31 characters.\n"
+        "\n"
+        "  * Press Enter to continue."
     },
     {
         // ERROR_INSUFFICIENT_PARTITION_SIZE,
--- a/base/setup/usetup/usetup.c
+++ b/base/setup/usetup/usetup.c
@@ -3251,12 +3251,16 @@ IsValidPath(
     // TODO: Add check for 8.3 too.
 
     /* Path must be at least 2 characters long */
-//    if (Length < 2)
-//        return FALSE;
+    if (Length < 2)
+        return FALSE;
 
     /* Path must start with a backslash */
-//    if (InstallDir[0] != L'\\')
-//        return FALSE;
+    if (InstallDir[0] != L'\\')
+        return FALSE;
+
+    /* ISO-9660 filenames longer than 31 may cause buffer overflows in the OS. */
+    if (Length > 31)
+        return FALSE;
 
     /* Path must not end with a backslash */
     if (InstallDir[Length - 1] == L'\\')
--- a/base/setup/usetup/usetup.c
+++ b/base/setup/usetup/usetup.c
@@ -3251,16 +3251,12 @@ IsValidPath(
     // TODO: Add check for 8.3 too.
 
     /* Path must be at least 2 characters long */
-    if (Length < 2)
-        return FALSE;
+//    if (Length < 2)
+//        return FALSE;
 
     /* Path must start with a backslash */
-    if (InstallDir[0] != L'\\')
-        return FALSE;
-
-    /* ISO-9660 filenames longer than 31 may cause buffer overflows in the OS. */
-    if (Length > 31)
-        return FALSE;
+//    if (InstallDir[0] != L'\\')
+//        return FALSE;
 
     /* Path must not end with a backslash */
     if (InstallDir[Length - 1] == L'\\')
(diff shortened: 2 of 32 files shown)


zecarlos1957:
https://github.com/reactos/reactos/assets/49964946/127fce0c-5f3f-43bd-a981-f008038db38d

JoachimHenze (approved):
A workaround, sure.
But I think it is worth getting merged as it protects the user.
And I doubt that all existing 3rd party applications would work flawlessly with extremely long SYSTEMROOT-path either. Many of them do have even problems when SYSTEMROOT is called C:\WinNT because they accidentally hardcoded it to be C:\Windows

So ros not being compatible (yet) to 256 chars for that path isn't a big deal imho. Because it would be a foolish choice for users anyway to use any such name,

Last thought/idea: It might be more memory efficient and would require less translation efforts, if we would not pop up that error-box, but instead would just limit the amount of characters the user can enter in that edit-field to 31 instead. Would also minimize the diff.

katahiromz (changes requested):
There is no change in `base/setup/usetup/lang/ja-JP.h`.

katahiromz (approved):
LGTM from my side.

base/setup/usetup/usetup.c
```diff
@@ -3251,12 +3251,16 @@ IsValidPath(
...
 
     /* Path must be at least 2 characters long */
-//    if (Length < 2)
-//        return FALSE;
+    if (Length < 2)
+        return FALSE;
 
     /* Path must start with a backslash */
-//    if (InstallDir[0] != L'\\')
-//        return FALSE;
```
binarymaster, line 3259: These commented out lines were added in 9c64b57dc by Hermes, needs to be reviewed by him.

HBelusca: These actually came from one of those commits shown in the blame there:
https://github.com/reactos/reactos/blame/14284ddc618661fc8c8e64acde2238ca3dfd7897/base/setup/usetup/filesup.c#L133

binarymaster: The second commit by Eric is so descriptive...

HBelusca:
Why 31? Not 30 or something else? The JIRA report mentions 31 because this is a limit length for directory/file names in ISO images, but that's unrelated to ReactOS installation on a hard-disk proper...

base/setup/usetup/lang/bg-BG.h
```diff
@@ -2071,7 +2071,9 @@ MUI_ERROR bgBGErrorEntries[] =
         // ERROR_DIRECTORY_NAME,
         "\215\245\242\240\253\250\244\255\256 \250\254\245 \255\240 \244\250\340\245\252\342\256\340\250\357\342\240.\n"
         "\n"
-        "  * \215\240\342\250\341\255\245\342\245 \257\340\256\250\247\242\256\253\245\255 \252\253\240\242\250\350, \247\240 \244\240 \257\340\256\244\352\253\246\250\342\245."
+        "The maximum length for the directory name is 31 characters.\n"
+        "\n"
+        "  * Press Enter to continue."
```
HBelusca, line 2076: Does only "enter" work there, or any key press? If the latter then you could keep the original (already-translated) message there.

JoachimHenze: I would also love a solution which wouldn't have to touch the ressources at all.

zecarlos1957:
[dprint1.txt](https://github.com/reactos/reactos/files/12269967/dprint1.txt)

In this "DPrint1.txt" file, we can observe the address of Source and Destination UNICODE_STRING and the effect of a write operation on Destination Ustring over Source Ustring.

HBelusca:
Can you please provide an example how to reproduce the problem? (actual ROS installation directory name, etc.)

sdk/lib/rtl/registry.c
```diff
@@ -330,8 +330,8 @@ RtlpCallQueryRegistryRoutine(IN PRTL_QUERY_REGISTRY_TABLE QueryTable,
                 RtlInitEmptyUnicodeString(&Source, Data, (USHORT)Length);
                 Source.Length = Source.MaximumLength - sizeof(UNICODE_NULL);
 
-                /* Setup the desination string */
-                RtlInitEmptyUnicodeString(&Destination, (PWCHAR)SpareData, 0);
+                /* Setup the destination string */
+                RtlInitEmptyUnicodeString(&Destination, ((PWCHAR)Data + Length + sizeof(UNICODE_NULL)), 0);
```
katahiromz, line 334: The unit of `PWCHAR` is `WCHAR`.
Adding `sizeof(UNICODE_NULL) * sizeof(WCHAR)` bytes is questionable.

zecarlos1957:
@HBelusca i'm trying to make a test program for RtlpCallQueryRegistryRoutine.

sdk/lib/rtl/registry.c
tkreuzer: Also `Length` is in bytes. And you can use `Source.MaximumLength`, which already contains the `UNICODE_NULL`:
`RtlInitEmptyUnicodeString(&Destination, (PWCHAR)((PCHAR)Data + Source.MaximumLength), 0);`
or
`RtlInitEmptyUnicodeString(&Destination, Add2Ptr(Data, Length + Source.MaximumLength), 0);`

tkreuzer (changes requested):
This fix isn't correct. Instead `SpareData` ans `SpareLength` need to be set correctly, as they are used later on as well. To do this, the code between line 202 and 128 should move to the top, where they are initialized, so that the value is correct in all cases.

EDIT (hbelusca): Should be between lines 202 and 228.

JoachimHenze (changes requested):
This PR looks entirely different to what it looked like when I approved it. It even touches entirely different files now, and is wrong as others have pointed out already. I would not have approved the current state.
Only good thing about the current state is the "desination"->"destination part" and that it points towards some buffers that where too short before.

tkreuzer (approved):
LGTM now

sdk/lib/rtl/registry.c
```diff
@@ -196,36 +196,37 @@ RtlpCallQueryRegistryRoutine(IN PRTL_QUERY_REGISTRY_TABLE QueryTable,
...
-            /* Check if we have space to copy the data */
-            RequiredLength = KeyValueInfo->NameLength + sizeof(UNICODE_NULL);
-            if ((SpareData > DataEnd) || (SpareLength < RequiredLength))
-            {
-                /* Fail and return the missing length */
-                *InfoSize = (ULONG)(SpareData - (PCHAR)KeyValueInfo) + RequiredLength;
-                return STATUS_BUFFER_TOO_SMALL;
-            }
+        /* Check if we have space to copy the data */
+        RequiredLength = KeyValueInfo->NameLength + sizeof(UNICODE_NULL);
```
HBelusca, line 219:
```suggestion
```

HBelusca (approved):

PR #5470, opened 2023-07-22
Misc compilation fixes

zefklop:
## Purpose

Some quick fixes for compilation issues I faced on Fedora 37

## Proposed changes

See individual commits

### DIFF
--- a/dll/win32/framedyn/chstring.cpp
+++ b/dll/win32/framedyn/chstring.cpp
@@ -24,6 +24,8 @@
 
 /* INCLUDES ******************************************************************/
 
+#define CHSTRING_BUILD
+
 #include <chstring.h>
 #define NDEBUG
 #include <debug.h>
--- a/ntoskrnl/include/internal/mm.h
+++ b/ntoskrnl/include/internal/mm.h
@@ -868,7 +868,7 @@ MmDeleteKernelStack(PVOID Stack,
 
 /* balance.c / pagefile.c******************************************************/
 
-inline VOID UpdateTotalCommittedPages(LONG Delta)
+FORCEINLINE VOID UpdateTotalCommittedPages(LONG Delta)
 {
     /*
      * Add up all the used "Committed" memory + pagefile.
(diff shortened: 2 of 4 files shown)


JoachimHenze:
Can you paste the compilers output wof the unfixed ithin the PR please? Thx.

zefklop:
> Can you paste the compilers output of the unfixed within the PR please? And tell us the RosBELinux version you used for compilation? Thx.

RosBE is the latest one, 2.1.
Change in txt2nls is because stdint16_t isn't declared.
Framedyn stuff is because I was trying to compile with less optimizations and linker complained about undefined symbols. I didn't dive too much, just those aliases are already defined in the spec when building. 
inline stuff is the usual stupidity of C standard allowing compiler to actually completely dismiss function definition and expecting the linker to catch up

JoachimHenze:
Thx for explanation.
RosBELin 2.1 is from 2012. Are you sure that is what you use?
Or do you use the latest of today, which would be:
https://sourceforge.net/projects/reactos/files/RosBE-Unix/2.2.1/

zefklop:
> Thx for explanation. RosBELin 2.1 is from 2012. Are you sure that is what you use? Or do you use the latest of today, which would be: https://sourceforge.net/projects/reactos/files/RosBE-Unix/2.2.1/

2.2.1 yes 🙂

JoachimHenze (approved):

HBelusca:
>  Change in txt2nls is because stdint16_t isn't declared.

This is interesting. Because I remember that recently (like 1 or 2 weeks ago) somewhere else found that problem.
Do you know why our current linux build bots succeed the compilation there?

HBelusca (approved):

zefklop:
> > Change in txt2nls is because stdint16_t isn't declared.
> 
> This is interesting. Because I remember that recently (like 1 or 2 weeks ago) somewhere else found that problem. Do you know why our current linux build bots succeed the compilation there?
> 
> (EDIT: Added "no-squash-merge" label because it appears the 3 separate commits are better be merged separately.)

Different c++ headers ? 🤷‍♂️

PR #5471, opened 2023-07-22
[WIN32SS:USER] Always reference a window when using it as parent/child etc.

zefklop:
CORE-18811

## Purpose

Explicitly reference window objects when using it as sibling in some other window object

JIRA issue: [CORE-18811](https://jira.reactos.org/browse/CORE-18811)

## Proposed changes

Very much like what @tkreuzer did recently for owner window

### DIFF
--- a/win32ss/user/ntuser/window.h
+++ b/win32ss/user/ntuser/window.h
@@ -127,21 +127,64 @@ VOID FASTCALL IntCheckImeShowStatusInThread(PWND pImeWnd);
 
 static inline
 VOID
+ReplaceWndPtr(_Inout_ PWND* ppwnd, _In_opt_ PWND pwndNew)
+{
+    /* First reference the new one */
+    if (pwndNew != NULL)
+    {
+        UserReferenceObject(pwndNew);
+    }
+
+    /* Then dereference the previous one */
+    if (*ppwnd != NULL)
+    {
+        UserDereferenceObject(*ppwnd);
+    }
+
+    /* And set */
+    *ppwnd = pwndNew;
+}
+
+static inline
+VOID
 WndSetOwner(_Inout_ PWND pwnd, _In_opt_ PWND pwndOwner)
 {
-    /* First reference the new owner window */
-    if (pwndOwner != NULL)
-    {
-        UserReferenceObject(pwndOwner);
-    }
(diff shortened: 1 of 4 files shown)


JoachimHenze (approved):
Approved. Please wait for others a bit before merging. Thx for working on the BSOD.

binarymaster:
Minor suggestion for commit title: `[WIN32SS:USER]` -> `[NTUSER][USER32]`

HBelusca (approved):

zefklop:
> Minor suggestion for commit title: `[WIN32SS:USER]` -> `[NTUSER][USER32]`

That's kernel component, hence the prefix

win32ss/user/ntuser/window.h
```diff
@@ -127,21 +127,64 @@ VOID FASTCALL IntCheckImeShowStatusInThread(PWND pImeWnd);
...
+    /* First reference the new one */
+    if (pwndNew != NULL)
     {
-        UserReferenceObject(pwndOwner);
+        UserReferenceObject(pwndNew);
     }
 
-    /* Now dereference the previous owner window */
-    if (pwnd->spwndOwner != NULL)
+    /* The dereference the previous one */
```
katahiromz, line 138: The dereference?

katahiromz (approved):

binarymaster:
> That's kernel component, hence the prefix

Well, `[USER32]` is the user-mode part, and I'm kinda already used to `[NTUSER]` prefix since we're using it for quite some time already - it starts with "NT", it's unique, and everybody knows what is it.

However the changes here cover both, so perhaps it makes sense to use upper module prefix.

win32ss/user/ntuser/window.h
zefklop: Then :sweat_smile:

zefklop:
> > That's kernel component, hence the prefix
> 
> Well, `[USER32]` is the user-mode part, and I'm kinda already used to `[NTUSER]` prefix since we're using it for quite some time already - it starts with "NT", it's unique, and everybody knows what is it.
> 
> However the changes here cover both, so perhaps it makes sense to use upper module prefix.

In fact, those should be two commits. Good catch. I amended.

binarymaster (approved):
Looks good! Don't forget to add CORE-18811 to commit messages :)

PR #5472, opened 2023-07-23
[DMILIB][UDMIHELP] Favor pragma once

JoachimHenze:
## Purpose

Favor pragma once over the include guards. Is shorter.

### DIFF
--- a/sdk/lib/udmihelp/udmihelp.h
+++ b/sdk/lib/udmihelp/udmihelp.h
@@ -5,8 +5,7 @@
  * COPYRIGHT:   Copyright 2018 Stanislav Motylkov
  */
 
-#ifndef UDMIHELP_H
-#define UDMIHELP_H
+#pragma once
 
 #ifdef __cplusplus
 extern "C" {
@@ -37,4 +36,3 @@ FreeSMBiosData(
 } // extern "C"
 #endif
 
-#endif /* UDMIHELP_H */
--- a/sdk/lib/udmihelp/udmihelp.h
+++ b/sdk/lib/udmihelp/udmihelp.h
@@ -35,4 +35,3 @@ FreeSMBiosData(
 #ifdef __cplusplus
 } // extern "C"
 #endif
-
--- a/sdk/lib/dmilib/dmilib.h
+++ b/sdk/lib/dmilib/dmilib.h
@@ -1,13 +1,11 @@
 /*
  * PROJECT:     ReactOS DMI/SMBIOS Library
  * LICENSE:     GPL - See COPYING in the top level directory
- * FILE:        dmilib.h
  * PURPOSE:     SMBIOS table parsing functions
  * PROGRAMMERS: Timo Kreuzer (timo.kreuzer@reactos.org)
  */
 
-#ifndef DMILIB_H
-#define DMILIB_H
+#pragma once
 
 enum _ID_STRINGS
 {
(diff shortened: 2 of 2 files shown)


sdk/lib/udmihelp/udmihelp.h
```diff
@@ -37,4 +36,3 @@ FreeSMBiosData(
 } // extern "C"
 #endif
 
-#endif /* UDMIHELP_H */
```
binarymaster, line 40: You can remove remaining empty line from the EOF.

JoachimHenze: done.

binarymaster (approved):
No strong preference on this, so I think it's ok.

HBelusca:
Personally I go like this:
- If "public" headers for a lib: use the ifdef guards + pragma once (for max compatibility);
- If private headers within a module: just use pragma once.

JoachimHenze:
> * [in public headers use both] (for max compatibility)

Can you please elaborate, because I absolutely cannot understand how this would be any more compatible in our codebase. Everything beginning from MS VC6 does understand "pragma once". And as we all know our code does not even compile anymore with VS2013 currently. So I do not understand how this extra-length would improve any compatiblity anywhere. Even our oldest branches on the very finest selected few builds did never manage to compile with anything older than 2008 (let alone speaking of actually running it).
So unless you could prove me to mess in any way with PCH here, I cannot take that serious.
**But if there would be any potential issue with PCH, then please tell me**, because that is something that I usually have turned off in my own projects.

"Just pragma once" is shortest and builds fastest, because the compiler can skip the header the moment he parses that and does not need to read until then #endifs of the include-guards. And it avoids typos.
Also see https://reactos.org/wiki/Coding_Style#Other_points

HBelusca:
This *may* help dealing with two situations at once:
1. "backwards" compatibility with compilers not knowing the pragma-once stuff (but as you said, very rare now and we would't support them anyway); and
2. Because in some cases it can be useful to know whether a given (optional) header has been included, and if not, take a certain different route.
Examples:
```c
#ifndef SOME_DEFINE_FROM_HEADER
#error Header xxxx hasn't been included!
#endif
```
or:
```c
#ifndef SOME_DEFINE_FROM_HEADER
// Just define this struct that we use, that would otherwise be available from
// some header together with other bunch of stuff we may not need...
typedef struct _Whatever { /* ... */ } Whatever;
#endif
```

Of course one could object that just having in the header:
```c
#pragma once

#ifndef SOME_DEFINE_FROM_HEADER
#define SOME_DEFINE_FROM_HEADER
#endif

// the rest of the header
```
does the job of point 2 "better" than:
```c
#ifndef SOME_DEFINE_FROM_HEADER
#define SOME_DEFINE_FROM_HEADER

#pragma once

// the rest of the header

#endif /* SOME_DEFINE_FROM_HEADER */
```

JoachimHenze:
For me alone the possibility of accidentally copy-pasting the #ifndef SOME_DEFINE_FROM_HEADER from one header into another one when copying the file, is not worth that risk. But thanks for not rejecting, as discussed in chat.
I will stick with https://reactos.org/wiki/Coding_Style#Other_points,
but think your examples are still good to be posted here for educational purposes and specific exceptional cases.
Another point where pragma once may behave differently than the include guards is when including the same header twice. But that is a code-smell in itself, in > 99.99% of the cases I do claim.

PR #5473, opened 2023-07-23
[ATL] Overload const-return-versions of FindChar/FindString/FindOneOf

katahiromz:
## Purpose
Fixing type problems.
JIRA issue: N/A

## Proposed changes

- Overload the non-const-return-version of `CSimpleStringT::GetString`.
- Overload the const-return-versions of `FindChar/FindCharReverse/FindString/FindOneOf`.
- Modify `CStringT::Replace`.

## TODO

- [x] Do build.

### DIFF
--- a/sdk/lib/atl/atlsimpstr.h
+++ b/sdk/lib/atl/atlsimpstr.h
@@ -358,6 +358,10 @@ public:
         return GetData()->nDataLength;
     }
 
+    PXSTR GetString() throw()
+    {
+        return m_pszData;
+    }
     PCXSTR GetString() const throw()
     {
         return m_pszData;
--- a/sdk/lib/atl/cstringt.h
+++ b/sdk/lib/atl/cstringt.h
@@ -102,28 +102,28 @@ public:
         _In_z_ LPCWSTR pszSource,
         _In_z_ LPCWSTR pszSub)
     {
-        return ::wcsstr(pszSource, pszSub);
+        return const_cast<LPWSTR>(::wcsstr(pszSource, pszSub));
     }
 
     static LPWSTR __cdecl FindChar(
         _In_z_ LPCWSTR pszSource,
         _In_ WCHAR ch)
     {
-        return ::wcschr(pszSource, ch);
+        return const_cast<LPWSTR>(::wcschr(pszSource, ch));
     }
 
     static LPWSTR __cdecl FindCharReverse(
         _In_z_ LPCWSTR pszSource,
         _In_ WCHAR ch)
     {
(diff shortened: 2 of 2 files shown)


JoachimHenze:
That looks like a really bad idea to me!
Why can't the return values be changed instead?
What do you want to achieve.
Please do not merge this before you have at least 3 approvals!

HBelusca (approved):

JoachimHenze:
After your 2nd commits "bigFix" it looks better than before, agreed. Now less additional const_casts are needed.
But personally I will not give approval for any PR that adds any const_cast anywhere.
I do not reject though. Neutrality now.
I do recommend to wait a week before the merging. Maybe others have ideas whether it can be improved further?

sdk/lib/atl/cstringt.h
```diff
@@ -858,7 +858,7 @@ class CStringT :
         PXSTR pszResult = CThisSimpleString::GetBuffer(nResultLength);
         PXSTR pszNext;
         int nCount = 0, nRemaining = nLength;
-        while (nRemaining && (pszNext = StringTraits::FindString(pszResult, pszOld)))
+        while (nRemaining && (pszNext = const_cast<PXSTR>(StringTraits::FindString(pszResult, pszOld))))
```
ThFabba, line 861: To avoid the need for this, you could have two overloads of FindString -- much like wcsstr/strstr: one with const input and const return and one with non-const input and non-const return.

katahiromz: Fixed. Thanks!

katahiromz: Now we could fix this issue without using `const_cast`.

sdk/lib/atl/atlsimpstr.h
```diff
@@ -358,6 +358,10 @@ class CSimpleStringT
         return GetData()->nDataLength;
     }
 
+    PXSTR GetString() throw()
```
HBelusca, line 361: Does **MS** ATL have such const + non-const overloaded versions?

katahiromz: MS ATL is using const versions only:

image

katahiromz: And in traits:

image

HBelusca: Personally I think it would be just preferrable to follow the MS ATL there.

katahiromz: Cc @JoachimHenze

PR #5481, opened 2023-07-25
[SHELL32] Improve CopyItems logic

HuwCampbell:
This change addresses a longstanding TODO in copying and moving to shell folders.

It's a simple change, but it removes a bit of code and makes things simpler.

Corresponding [wine PR](https://gitlab.winehq.org/wine/wine/-/merge_requests/3360).

### DIFF
--- a/dll/win32/shell32/droptargets/CFSDropTarget.cpp
+++ b/dll/win32/shell32/droptargets/CFSDropTarget.cpp
@@ -36,6 +36,9 @@ HRESULT CFSDropTarget::_CopyItems(IShellFolder * pSFFrom, UINT cidl,
     HRESULT ret;
     WCHAR wszDstPath[MAX_PATH + 1] = {0};
     PWCHAR pwszSrcPathsList = (PWCHAR) HeapAlloc(GetProcessHeap(), 0, MAX_PATH * sizeof(WCHAR) * cidl + 1);
+    if (!pwszSrcPathsList)
+        return E_OUTOFMEMORY;
+
     PWCHAR pwszListPos = pwszSrcPathsList;
     STRRET strretFrom;
     SHFILEOPSTRUCTW fop;
--- a/dll/win32/shell32/droptargets/CFSDropTarget.cpp
+++ b/dll/win32/shell32/droptargets/CFSDropTarget.cpp
@@ -34,7 +34,7 @@ HRESULT CFSDropTarget::_CopyItems(IShellFolder * pSFFrom, UINT cidl,
                                   LPCITEMIDLIST * apidl, BOOL bCopy)
 {
     HRESULT ret;
-    WCHAR wszDstPath[MAX_PATH+1];
+    WCHAR wszDstPath[MAX_PATH + 1];
     WCHAR *wszSrcPathsList = (WCHAR *) HeapAlloc(GetProcessHeap(), 0, MAX_PATH * sizeof(WCHAR) * cidl + 1);
     WCHAR *wszListPos = wszSrcPathsList;
     STRRET strretFrom;
--- a/dll/win32/shell32/droptargets/CFSDropTarget.cpp
+++ b/dll/win32/shell32/droptargets/CFSDropTarget.cpp
@@ -24,82 +24,57 @@
 
 WINE_DEFAULT_DEBUG_CHANNEL (shell);
 
-/****************************************************************************
- * BuildPathsList
- *
- * Builds a list of paths like the one used in SHFileOperation from a table of
- * PIDLs relative to the given base folder
- */
-static WCHAR* BuildPathsList(LPCWSTR wszBasePath, int cidl, LPCITEMIDLIST *pidls)
-{
-    WCHAR *pwszPathsList = (WCHAR *)HeapAlloc(GetProcessHeap(), 0, MAX_PATH * sizeof(WCHAR) * cidl + 1);
-    WCHAR *pwszListPos = pwszPathsList;
-
-    for (int i = 0; i < cidl; i++)
-    {
-        FileStructW* pDataW = _ILGetFileStructW(pidls[i]);
-        if (!pDataW)
-        {
-            ERR("Mistreating a pidl:\n");
-            pdump_always(pidls[i]);
-            continue;
-        }
-
-        PathCombineW(pwszListPos, wszBasePath, pDataW->wszName);
-        pwszListPos += wcslen(pwszListPos) + 1;
-    }
-    *pwszListPos = 0;
-    return pwszPathsList;
-}
 
 /****************************************************************************
  * CFSDropTarget::_CopyItems
  *
- * copies items to this folder
- * FIXME: We should not ask the parent folder: 'What is your path', and then manually build paths assuming everything is a simple pidl!
- * We should be asking the parent folder: Give me a full name for this pidl (for each child!)
+ * copies or moves items to this folder
  */
 HRESULT CFSDropTarget::_CopyItems(IShellFolder * pSFFrom, UINT cidl,
                                   LPCITEMIDLIST * apidl, BOOL bCopy)
 {
-    LPWSTR pszSrcList;
-    HRESULT hr;
-    WCHAR wszTargetPath[MAX_PATH + 1];
-
-    wcscpy(wszTargetPath, m_sPathTarget);
-    //Double NULL terminate.
-    wszTargetPath[wcslen(wszTargetPath) + 1] = '\0';
-
-    TRACE ("(%p)->(%p,%u,%p)\n", this, pSFFrom, cidl, apidl);
-
+    HRESULT ret;
+    WCHAR wszDstPath[MAX_PATH+1];
+    WCHAR *wszSrcPathsList = (WCHAR *) HeapAlloc(GetProcessHeap(), 0, MAX_PATH * sizeof(WCHAR) * cidl + 1);
+    WCHAR *wszListPos = wszSrcPathsList;
     STRRET strretFrom;
-    hr = pSFFrom->GetDisplayNameOf(NULL, SHGDN_FORPARSING, &strretFrom);
-    if (hr != S_OK)
-        return hr;
+    SHFILEOPSTRUCTW fop;
 
-    pszSrcList = BuildPathsList(strretFrom.pOleStr, cidl, apidl);
-    TRACE("Source file (just the first) = %s, target path = %s, bCopy: %d\n", debugstr_w(pszSrcList), debugstr_w(m_sPathTarget), bCopy);
-    CoTaskMemFree(strretFrom.pOleStr);
-    if (!pszSrcList)
-        return E_OUTOFMEMORY;
+    /* Build a double null terminated list of C strings from source paths */
+    for (UINT i = 0; i < cidl; i++) {
+        ret = pSFFrom->GetDisplayNameOf(apidl[i], SHGDN_FORPARSING, &strretFrom);
+        if (FAILED(ret))
+            goto cleanup;
 
-    SHFILEOPSTRUCTW op = {0};
-    op.pFrom = pszSrcList;
-    op.pTo = wszTargetPath;
-    op.hwnd = m_hwndSite;
-    op.wFunc = bCopy ? FO_COPY : FO_MOVE;
-    op.fFlags = FOF_ALLOWUNDO | FOF_NOCONFIRMMKDIR;
+        ret = StrRetToBufW(&strretFrom, NULL, wszListPos, MAX_PATH);
+        if (FAILED(ret))
+            goto cleanup;
 
-    int res = SHFileOperationW(&op);
-
-    HeapFree(GetProcessHeap(), 0, pszSrcList);
-
-    if (res)
-    {
-        ERR("SHFileOperationW failed with 0x%x\n", res);
-        return E_FAIL;
+        wszListPos += lstrlenW(wszListPos) + 1;
     }
-    return S_OK;
+    /* Append the final null. */
+    *wszListPos=0;
+
+    /* Build a double null terminated target (this path) */
+    lstrcpynW(wszDstPath, m_sPathTarget, MAX_PATH);
+    wszDstPath[wcslen(wszDstPath) + 1] = '\0';
+
+    ZeroMemory(&fop, sizeof(fop));
+    fop.hwnd = m_hwndSite;
+    fop.wFunc = bCopy ? FO_COPY : FO_MOVE;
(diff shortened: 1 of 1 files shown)


dll/win32/shell32/droptargets/CFSDropTarget.cpp
```diff
@@ -24,82 +24,57 @@
...
-    wcscpy(wszTargetPath, m_sPathTarget);
-    //Double NULL terminate.
-    wszTargetPath[wcslen(wszTargetPath) + 1] = '\0';
-
-    TRACE ("(%p)->(%p,%u,%p)\n", this, pSFFrom, cidl, apidl);
-
+    HRESULT ret;
+    WCHAR wszDstPath[MAX_PATH+1];
+    WCHAR *wszSrcPathsList = (WCHAR *) HeapAlloc(GetProcessHeap(), 0, MAX_PATH * sizeof(WCHAR) * cidl + 1);
+    WCHAR *wszListPos = wszSrcPathsList;
```
HBelusca, line 39: Please check
1. Use 4-space indent instead of TABs;
2. Since those wszList thingies are pointers, it would be better to call them pwszListPos, etc.
3. Also, `WCHAR *` --> `PWCHAR` for pointer to non-NULL-terminated buffer; `PWSTR` for pointer to NULL-terminated string; `PCWSTR` for pointer to NULL-terminated unmodifiable string, etc.
4. opening and closing curly brackets for if-, for-, and other code blocks are on their own lines.

dll/win32/shell32/droptargets/CFSDropTarget.cpp
```diff
@@ -24,82 +24,57 @@
...
+    SHFILEOPSTRUCTW fop;
 
-    SHFILEOPSTRUCTW op = {0};
-    op.pFrom = pszSrcList;
-    op.pTo = wszTargetPath;
-    op.hwnd = m_hwndSite;
-    op.wFunc = bCopy ? FO_COPY : FO_MOVE;
-    op.fFlags = FOF_ALLOWUNDO | FOF_NOCONFIRMMKDIR;
+    /* Build a double null terminated list of C strings from source paths */
+    for (UINT i = 0; i < cidl; i++) {
```
HBelusca, line 44:
```suggestion
    for (UINT i = 0; i < cidl; i++)
    {
```

dll/win32/shell32/droptargets/CFSDropTarget.cpp
```diff
@@ -24,82 +24,57 @@
...
-
-    if (res)
-    {
-        ERR("SHFileOperationW failed with 0x%x\n", res);
-        return E_FAIL;
+        wszListPos += lstrlenW(wszListPos) + 1;
     }
-    return S_OK;
+    /* Append the final null. */
+    *wszListPos=0;
```
HBelusca, line 56:
```suggestion
    /* Append the final null */
    *wszListPos = 0;
```
You could use `'\0'` or better, since this is a WCHAR string, use `L'\0'` .

dll/win32/shell32/droptargets/CFSDropTarget.cpp
```diff
@@ -24,82 +24,57 @@
...
+
+    ZeroMemory(&fop, sizeof(fop));
+    fop.hwnd = m_hwndSite;
+    fop.wFunc = bCopy ? FO_COPY : FO_MOVE;
+    fop.pFrom = wszSrcPathsList;
+    fop.pTo = wszDstPath;
+    fop.fFlags = FOF_ALLOWUNDO | FOF_NOCONFIRMMKDIR;
+    ret = S_OK;
+
+    if(SHFileOperationW(&fop)) {
```
HBelusca, line 70:
```suggestion
    if (SHFileOperationW(&fop))
    {
```

HBelusca (commented):
Those comments could also be applied to the corresponding Wine patch you've sent to them.

dll/win32/shell32/droptargets/CFSDropTarget.cpp
```diff
@@ -24,82 +24,57 @@
...
-    WCHAR wszTargetPath[MAX_PATH + 1];
-
-    wcscpy(wszTargetPath, m_sPathTarget);
-    //Double NULL terminate.
-    wszTargetPath[wcslen(wszTargetPath) + 1] = '\0';
-
-    TRACE ("(%p)->(%p,%u,%p)\n", this, pSFFrom, cidl, apidl);
-
+    HRESULT ret;
+    WCHAR wszDstPath[MAX_PATH+1];
```
learn-more, line 37: This needs to be zero-initialized, since `lstrcpynW` does not null-terminate if there is not enough room to do that.

dll/win32/shell32/droptargets/CFSDropTarget.cpp
```diff
@@ -24,82 +24,57 @@
...
-        return E_FAIL;
+        wszListPos += lstrlenW(wszListPos) + 1;
     }
-    return S_OK;
+    /* Append the final null. */
+    *wszListPos=0;
+
+    /* Build a double null terminated target (this path) */
+    lstrcpynW(wszDstPath, m_sPathTarget, MAX_PATH);
+    wszDstPath[wcslen(wszDstPath) + 1] = '\0';
```
learn-more, line 60: `lstrcpynW` does not null-terminate when the buffer is too small, so this could overwrite the data after `wszDstPath`.

HBelusca: I wonder why StringCchCopyNW couldn't be used there.

dll/win32/shell32/droptargets/CFSDropTarget.cpp
HuwCampbell: No worries.

dll/win32/shell32/droptargets/CFSDropTarget.cpp
HuwCampbell: I left this as is (see line 71 previously), but I agree [StringCchCopyW](https://learn.microsoft.com/en-us/windows/win32/api/strsafe/nf-strsafe-stringcchcopyw) would probably be better.

I leave this one to you guys as I won't have access to a machine which builds it for a few days.

HuwCampbell:
```
    /* Build a double null terminated target (this path) */
    ret = StringCchCopyW(wszDstPath, MAX_PATH, m_sPathTarget);
    if (FAILED(ret))
        goto cleanup;

    wszDstPath[wcslen(wszDstPath) + 1] = L'\0';
```
should do it.

dll/win32/shell32/droptargets/CFSDropTarget.cpp
```diff
@@ -24,82 +24,59 @@
...
+    fop.pFrom = pwszSrcPathsList;
+    fop.pTo = wszDstPath;
+    fop.fFlags = FOF_ALLOWUNDO | FOF_NOCONFIRMMKDIR;
+    ret = S_OK;
+
+    if (SHFileOperationW(&fop))
     {
-        ERR("SHFileOperationW failed with 0x%x\n", res);
-        return E_FAIL;
+        ERR("SHFileOperationW failed\n");
```
katahiromz, line 73: The return value handling of `SHFileOperationW` is wrong.

katahiromz (changes requested):
Did you test?

dll/win32/shell32/droptargets/CFSDropTarget.cpp
katahiromz: My info source was wrong. Never mind.

katahiromz (approved):

dll/win32/shell32/droptargets/CFSDropTarget.cpp
```diff
@@ -24,82 +24,59 @@
...
-
-    wcscpy(wszTargetPath, m_sPathTarget);
-    //Double NULL terminate.
-    wszTargetPath[wcslen(wszTargetPath) + 1] = '\0';
-
-    TRACE ("(%p)->(%p,%u,%p)\n", this, pSFFrom, cidl, apidl);
-
+    HRESULT ret;
+    WCHAR wszDstPath[MAX_PATH + 1] = {0};
+    PWCHAR pwszSrcPathsList = (PWCHAR) HeapAlloc(GetProcessHeap(), 0, MAX_PATH * sizeof(WCHAR) * cidl + 1);
```
katahiromz, line 38: `HeapAlloc` can be failed. You should check the return value of `HeapAlloc`.

dll/win32/shell32/droptargets/CFSDropTarget.cpp
katahiromz: OK, let's do it.

dll/win32/shell32/droptargets/CFSDropTarget.cpp
HuwCampbell: Done

dll/win32/shell32/droptargets/CFSDropTarget.cpp
HuwCampbell: I'll see if I can find some time to test it this evening after the kids are in bed.

HuwCampbell: Compiles and tests well.

dll/win32/shell32/droptargets/CFSDropTarget.cpp
```diff
@@ -24,82 +24,65 @@
...
 
-    pszSrcList = BuildPathsList(strretFrom.pOleStr, cidl, apidl);
-    TRACE("Source file (just the first) = %s, target path = %s, bCopy: %d\n", debugstr_w(pszSrcList), debugstr_w(m_sPathTarget), bCopy);
-    CoTaskMemFree(strretFrom.pOleStr);
-    if (!pszSrcList)
-        return E_OUTOFMEMORY;
+        pwszListPos += lstrlenW(pwszListPos) + 1;
+    }
+    /* Append the final null. */
+    *pwszListPos = L'\0';
```
katahiromz, line 60:
```suggestion
    *pwszListPos = UNICODE_NULL;
```

HuwCampbell: This is as suggested by @HBelusca .

JoachimHenze (approved):
See no problem of applying this for us **after** Wine applied it on their side.

HBelusca (approved):
Out of curiosity, does Wine have a test case for this functionality?

HuwCampbell:
> Out of curiosity, does Wine have a test case for this functionality?

Wine's shell is pretty immature in this in these parts. I didn't see a test (and the suite passed when I deleted the function body).

Furthermore, wine explorer doesn't seem to actually copy and paste properly, and they don't use drag simulation, or a background thread, like what I wrote for ReactOS here almost a decade ago (and is what windows does).

I wouldn't wait for wine, the shell doesn't appear to be their priority, but I'm going to poke that PR and see what I can do over time.

HuwCampbell:
@yagoulas Giannis any thoughts?

learn-more:
> > Out of curiosity, does Wine have a test case for this functionality?
> 
> Wine's shell is pretty immature in these parts. I didn't see a test (and the suite passed when I deleted the function body).
> 
> Furthermore, wine explorer doesn't seem to actually copy and paste properly, and they don't use drag simulation, or a background thread, like what I wrote for ReactOS here almost a decade ago (and is what windows does).
> 
> I wouldn't wait for wine, the shell doesn't appear to be their priority, but I'm going to poke that PR and see what I can do over time.

Did you test that dragging / dropping, copying etc still works in our shell?

HuwCampbell:
Yes of course

dll/win32/shell32/droptargets/CFSDropTarget.cpp
```diff
@@ -24,82 +24,65 @@
...
  */
 HRESULT CFSDropTarget::_CopyItems(IShellFolder * pSFFrom, UINT cidl,
                                   LPCITEMIDLIST * apidl, BOOL bCopy)
 {
-    LPWSTR pszSrcList;
-    HRESULT hr;
-    WCHAR wszTargetPath[MAX_PATH + 1];
+    HRESULT ret;
+    WCHAR wszDstPath[MAX_PATH + 1] = {0};
+    PWCHAR pwszSrcPathsList = (PWCHAR) HeapAlloc(GetProcessHeap(), 0, MAX_PATH * sizeof(WCHAR) * cidl + 1);
```
learn-more, line 38: For future reference:
We have `CComHeapPtr<>` now, that can be used to automatically clean up memory when going out of scope.
That might simplify the function (saves having to do `goto` everywhere).

dll/win32/shell32/droptargets/CFSDropTarget.cpp
```diff
@@ -24,82 +24,65 @@
...
+    PWCHAR pwszListPos = pwszSrcPathsList;
+    STRRET strretFrom;
+    SHFILEOPSTRUCTW fop;
 
-    TRACE ("(%p)->(%p,%u,%p)\n", this, pSFFrom, cidl, apidl);
+    /* Build a double null terminated list of C strings from source paths */
+    for (UINT i = 0; i < cidl; i++)
+    {
+        ret = pSFFrom->GetDisplayNameOf(apidl[i], SHGDN_FORPARSING, &strretFrom);
+        if (FAILED(ret))
```
learn-more, line 50: If this is an error that _should not_ happen, we have `FAILED_UNEXPECTEDLY()` that behaves the same as `FAILED()`, but it also prints a message in the failure case.

learn-more (approved):
lgtm.

Left some minor comments that might aid you in future PR's, I am fine with merging this with or without those improvements.

> Yes of course

You would be surprised by the amount of people that do not test their work, and shoot pr's at us from the web interface or by guessing what happens.

dll/win32/shell32/droptargets/CFSDropTarget.cpp
HuwCampbell: Good to know.

I think there's a few leaks around the shell, so it might be worth a separate PR to use smart pointers more broadly.

JoachimHenze:
@HuwCampbell please have a look at the regressions https://jira.reactos.org/browse/CORE-19210 and https://jira.reactos.org/browse/CORE-19211 that your change did introduce/unhide!

binarymaster:
This PR also fixed [CORE-18487](https://jira.reactos.org/browse/CORE-18487).

PR #5483, opened 2023-07-25
[FREELDR] Implement bus mouse detection

disean:
## Purpose

Automatically install the inport device driver. (this is possible since 7d5e1591313df39d0b2a81c20ab2af9e8d76252a). Besides, some people have reported PS/2 mouse issues with 86Box. For now, the best work around is to use the inport driver.

References:
https://raw.githubusercontent.com/86Box/86Box/master/src/device/mouse_bus.c
https://bochs.sourceforge.io/cgi-bin/lxr/source/iodev/busmouse.cc

inp

### DIFF
--- a/boot/freeldr/freeldr/arch/i386/irqsup.S
+++ b/boot/freeldr/freeldr/arch/i386/irqsup.S
@@ -1,0 +1,33 @@
+/*
+ * PROJECT:     FreeLoader
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Interrupt handling
+ * COPYRIGHT:   Copyright 2023 Dmitry Borisov <di.sean@protonmail.com>
+ */
+
+#include <asm.inc>
+
+#define PIC1_CONTROL_PORT   HEX(20)
+#define PIC_EOI             HEX(20)
+
+.code32
+
+PUBLIC _HwIrqHandler
+_HwIrqHandler:
+    push ax
+
+    /* Increment the interrupt count */
+    inc dword ptr ds:[_HwIrqCount]
+
+    /* Dismiss the interrupt */
+    mov al, PIC_EOI
+    out PIC1_CONTROL_PORT, al
+
+    pop ax
+    iret
+
+PUBLIC _HwIrqCount
+_HwIrqCount:
+    .long 0
+
+END
--- a/boot/freeldr/freeldr/arch/i386/pc/machpc.c
+++ b/boot/freeldr/freeldr/arch/i386/pc/machpc.c
@@ -38,6 +38,20 @@ DBG_DEFAULT_CHANNEL(HWDETECT);
 /* Mouse Systems Mouse */
 #define MOUSE_TYPE_MOUSESYSTEMS    4
 
+#define INPORT_REGISTER_CONTROL    0x00
+#define INPORT_REGISTER_DATA       0x01
+#define INPORT_REGISTER_SIGNATURE  0x02
+
+#define INPORT_REG_MODE            0x07
+#define INPORT_RESET               0x80
+#define INPORT_MODE_BASE           0x10
+#define INPORT_TEST_IRQ            0x16
+#define INPORT_SIGNATURE           0xDE
+
+#define PIC1_CONTROL_PORT          0x20
+#define PIC1_DATA_PORT             0x21
+#define PIC2_CONTROL_PORT          0xA0
+#define PIC2_DATA_PORT             0xA1
 
 /* PS2 stuff */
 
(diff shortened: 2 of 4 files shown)


binarymaster:
From MSVC i386:

```
D:\a\reactos\reactos\build\boot\freeldr\freeldr\asm\entry_freeldr_common_asm.asm(791) : error A2006:undefined symbol : stack32
D:\a\reactos\reactos\build\boot\freeldr\freeldr\asm\entry_freeldr_common_asm.asm(795) : error A2006:undefined symbol : i386idtptr
D:\a\reactos\reactos\build\boot\freeldr\freeldr\asm\entry_freeldr_common_asm.asm(810) : error A2006:undefined symbol : _FrldrBootDrive
D:\a\reactos\reactos\build\boot\freeldr\freeldr\asm\entry_freeldr_common_asm.asm(813) : error A2006:undefined symbol : _FrldrBootPartition
D:\a\reactos\reactos\build\boot\freeldr\freeldr\asm\entry_freeldr_common_asm.asm(817) : error A2006:undefined symbol : SwitchToReal16Address
D:\a\reactos\reactos\build\boot\freeldr\freeldr\asm\entry_freeldr_common_asm.asm(851) : error A2006:undefined symbol : SwitchToReal
D:\a\reactos\reactos\build\boot\freeldr\freeldr\asm\entry_freeldr_common_asm.asm(901) : error A2006:undefined symbol : SwitchToReal
D:\a\reactos\reactos\build\boot\freeldr\freeldr\asm\entry_freeldr_common_asm.asm(934) : error A2006:undefined symbol : i386CallRealMode
D:\a\reactos\reactos\build\boot\freeldr\freeldr\asm\entry_freeldr_common_asm.asm(958) : error A2006:undefined symbol : i386CallRealMode
```

boot/freeldr/freeldr/arch/i386/entry.S
```diff
@@ -199,6 +199,33 @@ _PxeCallApi:
...
+    call i386CallRealMode
+
+    /* Restore registers */
+    popa
+
+    mov eax, dword ptr ds:[BSS_IrqBitmap]
+
+    ret
+
+END
```
HBelusca, line 227:
```suggestion
    /* Restore registers and return the IRQ */
    popa
    mov eax, dword ptr ds:[BSS_IrqBitmap]
    ret
```

boot/freeldr/freeldr/arch/realmode/bmouse.inc
```diff
@@ -0,0 +1,166 @@
+/*
+ * PROJECT:     FreeLoader
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Bus Mouse Interrupt Detection
+ * COPYRIGHT:   Copyright 2023 Dmitry Borisov <di.sean@protonmail.com>
+ */
+
+#define PIC1_BASE         HEX(20)
```
HBelusca, line 8: Question: What requires this code to be run in real mode and ASM? Couldn't it be done in C like some of our other detection routines?

disean: In protected mode, the IRQs 0 to 32 are reserved for CPU generated exceptions. Ok, let's hope this code won't cause any exceptions.

binarymaster: You probably meant interrupt vectors, IRQs are different, usually having a more narrow range (but depending on the type of PIC).

boot/freeldr/freeldr/arch/i386/pc/machpc.c
```diff
@@ -1291,6 +1305,200 @@ DetectPS2Mouse(PCONFIGURATION_COMPONENT_DATA BusKey)
...
+    PartialDescriptor->u.Port.Start.LowPart = IoBase;
+    PartialDescriptor->u.Port.Start.HighPart = 0;
+    PartialDescriptor->u.Port.Length = 4;
+
+    /* Set Interrupt */
+    PartialDescriptor = &PartialResourceList->PartialDescriptors[1];
+    PartialDescriptor->Type = CmResourceTypeInterrupt;
+    PartialDescriptor->ShareDisposition = CmResourceShareUndetermined;
+    PartialDescriptor->Flags = CM_RESOURCE_INTERRUPT_LATCHED;
+    PartialDescriptor->u.Interrupt.Level = IrqVector;
```
HBelusca, line 1352: Is the IRQ level == IRQ vector? :thinking: :thinking:

boot/freeldr/freeldr/arch/i386/pc/machpc.c
```diff
@@ -1291,6 +1305,200 @@ DetectPS2Mouse(PCONFIGURATION_COMPONENT_DATA BusKey)
...
+
+    HwIrqCount = 0;
+
+    /* Reset the device */
+    WRITE_PORT_UCHAR((PUCHAR)IoBase + INPORT_REGISTER_CONTROL, INPORT_RESET);
+    WRITE_PORT_UCHAR((PUCHAR)IoBase + INPORT_REGISTER_CONTROL, INPORT_REG_MODE);
+
+    Vector = IrqVector + 8;
+
+    OldOffset = i386Idt[Vector].Offset;
```
HBelusca, line 1414:
```suggestion
    /* Save the old interrupt vector and replace it by ours */
    OldOffset = i386Idt[Vector].Offset;
```
Question: Should we touch that i386Idt table here, but not touch the real-mode IVT at address 0x00000000 and beyond (a bit like what was done here: https://git.reactos.org/?p=reactos.git;a=blob;f=boot/freeldr/freeldr/arch/i386/drivemap.c;h=e9fa105b889c9adccc9d7afa87bca8ead3a078da;hb=HEAD#l177 )  ?

boot/freeldr/freeldr/arch/i386/entry.S
```diff
@@ -199,6 +202,20 @@ _PxeCallApi:
 
     ret
 
+PUBLIC _HwMouseIrqHandler
+_HwMouseIrqHandler:
```
HBelusca, line 206: I'm thinking whether it could be better to move this mouse-specific code into a new .S file, much like what is done in arch/i386/drvmap.S
(OK this requires creating a new file, but this may help having more tidy code...)

HBelusca (commented):
The C code way looks cleaner than a pure ASM solution. And if it indeed works as intended, then go for it!

boot/freeldr/freeldr/arch/i386/entry.S
```diff
@@ -21,6 +21,9 @@
 #include <arch/pc/x86common.h>
 #include <arch/pc/pcbios.h>
 
+#define PIC1_COMMAND      HEX(20)
+#define PIC_EOI           HEX(20)
```
JoachimHenze, line 25: is it correct to use HEX(20) for both defines here?

boot/freeldr/freeldr/arch/i386/pc/machpc.c
disean: This doesn't apply in this case. The IVT cannot be used in protected mode.

boot/freeldr/freeldr/arch/i386/entry.S
disean: Yeah, there is no difference between 2 of those

boot/freeldr/freeldr/arch/i386/irqsup.S
```diff
@@ -0,0 +1,33 @@
+/*
+ * PROJECT:     FreeLoader
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Interrupt handling
```
HBelusca, line 4: but interrupt handling for _**InPort mouse recognition**_? (If so, maybe "moudtect.S" or "inport.S" or sthg like that)

HBelusca (commented):
Question: If no mouse detection is done, is ROS on 86box falling back to ps/2 mouse?

disean:
You cannot select more than 1 mouse device. ROS boots up, but it doesn't take input from the mouse. The really strange thing is that in the registry one `PointerPeripheral` key was created https://git.reactos.org/?p=reactos.git;a=blob;f=boot/freeldr/freeldr/arch/i386/pc/machpc.c;h=f20fa2b9a52ba2f231fce0ced9fce78bfa5312d1;hb=refs/heads/master#l1126

boot/freeldr/freeldr/arch/i386/irqsup.S
disean: I'm going to use this code for CardBus detection as well.

binarymaster:
> If no mouse detection is done, is ROS on 86box falling back to ps/2 mouse?

Mouse detection is always done in Freeloader, including PS/2 controller and peripheral detection.

In order to fallback to PS/2 mouse it has to be detected first. It's not the case when booting with ACPI HAL though.

boot/freeldr/freeldr/include/arch/pc/hardware.h
```diff
@@ -72,6 +72,9 @@ VOID
 DetectBiosDisks(PCONFIGURATION_COMPONENT_DATA SystemKey,
                 PCONFIGURATION_COMPONENT_DATA BusKey);
 
+VOID __cdecl HwIrqHandler(VOID);
```
HBelusca, line 75: within an `#ifdef _M_IX86` perhaps?

HBelusca (approved):

boot/freeldr/freeldr/include/arch/pc/hardware.h
```diff
@@ -72,6 +72,11 @@ VOID
 DetectBiosDisks(PCONFIGURATION_COMPONENT_DATA SystemKey,
                 PCONFIGURATION_COMPONENT_DATA BusKey);
 
+#if defined(_M_IX86)
+VOID __cdecl HwIrqHandler(VOID);
+extern volatile ULONG HwIrqCount;
+#endif
```
HBelusca, line 78: I wonder whether these two things need to be in the global hardware.h header, or could be placed in a localized one (within arch/i386/pc) or even just in the (single??) .c file that would use those.

HBelusca (approved):

boot/freeldr/freeldr/include/arch/pc/hardware.h
disean: Moved into a single .c file

binarymaster (approved):

PR #5482, opened 2023-07-25
[KERNEL32][MSAFD][WS2HELP] Add winsock socket inheritance. CORE-17554

bigdude9:
## Purpose

Implement winsock socket inheritance in ReactOS

JIRA issue: [CORE-17554](https://jira.reactos.org/browse/CORE-17554)

## Proposed changes

1 - set winsock handle object creation properties to include OBJ_INHERIT [WS2HELP]
2 - add code in kernel32.dll to obtain the socket information, allocate space in child and inject the information using the PEB in CreateProcessInteralW, prepare winsock structures in BaseProcessStarup, and clean up winsock data in ExitProcess. [KERNEL32]
3 - adapt WsaDuplicateSocket and WsaSocket to detect and use the new inheritance information when present. [MSAFD]

NOTE: code uses shared structure areas currently not referenced anywhere else in ReactOS code
PEB areas used by this code are SpareULong and SparePtr2 - these carry socket count and child allocation pointer
Also makes use of Socket Shared data area item lpProtocolInfo->dwProviderReserved to carry socket handle value

PR #5484, opened 2023-07-25
[KERNEL32][MSAFD] Add winsock socket handle duplication to DuplicateHandle. CORE-19045

bigdude9:
## Purpose

Revise DuplicateHandle to function for Winsock socket handles (partial implementation)
- dependent on PR #5482

JIRA issue: [CORE-19045](https://jira.reactos.org/browse/CORE-19045)

## Proposed changes

Added code in DuplicateHandle to detect Winsock Handles and treat them differently. Current implementation depends on the presence of PR-5482 for winsock inheritance. Current code can duplicate winsock handles within a process or duplicate socket handles that were marked as inheritable from a process' parent process. Duplication is handled in process by calling WsaDuplicateSocket and WsaSocket to create new socket. MSAFD dllmain.c is also updated to remove the use of DuplicateHandle in WspDuplicateSocket and replcae it with direct calls to NtDuplicateObject to avoid recursion or other dependency issues. Code does attempt to use the same method for arbitrary source process duplication, but it fails (FIXME)


dllmain.c
```diff
@@ -0,0 +1,4098 @@
+/*
+ * COPYRIGHT:   See COPYING in the top level directory
+ * PROJECT:     ReactOS Ancillary Function Driver DLL
+ * FILE:        dll/win32/msafd/misc/dllmain.c
```
SergeGautherie, line 4: This file was added to PR without its path...

bigdude9: I'm looking for a way to fix that now.. I merged two patches entered via the web interface and the second one lost its path.

bigdude9: fixed it in the original patch I created the PR form and it looks like it is now in the correct path in the PR. Please let me know if it is not visible to you.

dll/win32/kernel32/client/handle.c
```diff
@@ -187,6 +221,145 @@ DuplicateHandle(IN HANDLE hSourceProcessHandle,
...
+				{
+					DPRINT1("DuplicateHandle: SOCKET DUPLICATION exception (%lx) for duplicating handle(%lx) returning source handle (BADIMPL FIXME).\n", _SEH2_GetExceptionCode(), hSourceHandle);			
+				}
+				_SEH2_END;		
+			}
+			DPRINT("DuplicateHandle: SOCKET DUPLICATION completed, returning true.\n");									
+			return TRUE;																
+		} else 
+        {
+			if (isSocket && winsock_module && !canDupe)
```
SergeGautherie, line 322: `else` is superfluous after `return`.
You could do 1 `if (isSocket)` instead of repeating that condition multiple times. Probably for `winsock_module` too.

dll/win32/msafd/misc/dllmain.c
```diff
@@ -491,26 +491,38 @@ WSPDuplicateSocket(
...
+                                  FALSE, // DLB this turned off inheritability on all duplicated sockets. Is this correct ??   bInheritHandle ? OBJ_INHERIT : 0,
                                   DUPLICATE_SAME_ACCESS);
+    if (NT_SUCCESS(Status)) {
+        bDuplicated = TRUE;
+    } else {
+        bDuplicated = FALSE;
+	WSASetLastError(Status); 
+    }
     if (!bDuplicated)
     {
```
SergeGautherie, line 509:
```suggestion
    bDuplicated = NT_SUCCESS(Status);
    if (!bDuplicated)
    {
        WSASetLastError(Status);
```

dll/win32/msafd/misc/dllmain.c
```diff
@@ -491,26 +491,38 @@ WSPDuplicateSocket(
...
                                   0,
-                                  FALSE,
+                                  FALSE, // DLB this turned off inheritability on all duplicated sockets. Is this correct ??   bInheritHandle ? OBJ_INHERIT : 0,
                                   DUPLICATE_SAME_ACCESS);
+    if (NT_SUCCESS(Status)) {
+        bDuplicated = TRUE;
+    } else {
+        bDuplicated = FALSE;
+	    WSASetLastError(Status); 
+    }
```
SergeGautherie, line 525:
```suggestion
    bDuplicated = NT_SUCCESS(Status);
    if (!bDuplicated)
        WSASetLastError(Status);
```

dll/win32/kernel32/client/handle.c
```diff
@@ -187,6 +221,145 @@ DuplicateHandle(IN HANDLE hSourceProcessHandle,
         return FALSE;
     }
 
+	// winsock socket handles code currently only dupes from self or inherited sockets from parent are implemented
```
HBelusca, line 224: But.... does Windows have all that supplemental code here? I strongly doubt that.

HBelusca (changes requested):
All the modifications in kernel32 are a hack and are absent in Windows.
Adding those modifications there would break compatibility.

dll/win32/kernel32/client/handle.c
bigdude9: the else is outside the code block with the return.. I will look at it in context but it looks valid (but badly indented) at first blush.

The repeated if takes place in a for loop and the first if ensures that section of code in it executes only on the first pass of the loop, while the second if essentially skips that section of code if first section fails. Thinking about it again though it may be possible to get rid of the second if and just break out of the loop if the code inside the first if section fails.

dll/win32/kernel32/client/handle.c
bigdude9: I doubt it too, but it has 'something' undocumented that makes this work. There are some MSDN docs suggesting it is done through WSHGetSocketInformation and WSHSetSocketInformation Helper DLL functions both of which are unimplemented in ReactOS. [MSDN WSHGetSocketInformation page](https://learn.microsoft.com/en-us/previous-versions/windows/hardware/network/ff565992(v=vs.85))

HBelusca: I think that this would be worth investigating further. (Also note there exists a [WSPDuplicateSocket](https://learn.microsoft.com/en-us/previous-versions/windows/hardware/network/ff566282(v=vs.85)) function.)

bigdude9:
HBelusca,

The Kernel32 changes are definitely a hack. Even so they actually create a degree of compatibility rather than break it by adding existing windows functionality where ReactOS does not have it now. My prior comment/reply included a link for the only hint Microsoft has published (that I know of), about how this actually works. It is definitely an 'undocumented' feature within Winsock, but it has been there since Winsock v1.1 and does not currently work at all in ReactOS. This particular implementation is probably not how windows does it and in that respect it breaks binary compatibility but it does create new functional compatibility. I'm definitely open to doing it a better more windows compatible way, but this is a cleanroom workalike hack implementation that does work. Even so, I don't doubt there is a better way, but my current knowledge could only take me this far. Implementing the helper functions WSHGetSocketInformation and WSHSetSocketInformation would probably be a better approach, but I could not figure out how to make them work (and apparently to date nobody else could either).

dll/win32/kernel32/client/handle.c
bigdude9: WSPDuplicateSocket was the first thing I looked at and it is used in the Inheritance PR [absolutely also a hack] which this PR depends on (in that PR it is called from CreateProcessInternalW). The problem with that function is that It can only work from the process owning the original socket and DuplicateHandle is more often called from the 'target' process. That's the problem here.. this is a cross process operation and there is currently no provision for handling that IPC I can find anywhere in ReactOS (apparently if the MSDN page is correct, Windows handles it somewhere in the helper DLL).

HBelusca:
Those Get/SetSocketInformation seem to be implemented by the different wshtcpip, wshirda modules. (This looks like potentially going to duplicate code or stuff, not centralized into one single dll.)

dll/win32/kernel32/client/handle.c
bigdude9: You are right about the if.. the first one for !winsock_module is definitely superfluous. The code is not in a loop here (it is in the other PR). I'll look it over more carefully, but the intent was to treat three cases in separate code blocks because one is very likely to need future revision (the one for !canDupe).

dll/win32/kernel32/client/handle.c
```diff
@@ -187,6 +221,138 @@ DuplicateHandle(IN HANDLE hSourceProcessHandle,
...
+                } else DPRINT1("Error calling getsockopt for inheritable socket.\n");
+            } else DPRINT1("Error GetProcAddressA() could not Dynlink getsockopt!\n");			
+        } else DPRINT1("Error GetModuleHandleA() could not find WS2_32 module handle.\n");
+        // NOTE: currently this code duplicates the handle/socket but DOES NOT CHAGE THE HANDLE ACCESS OR OPTIONS IN ANY WAY (ignores dwOption and access rights changes)
+        // TODO: revise code to respect access changes requested in parameters
+        if (isSocket && GetCurrentProcessId() == GetProcessId(hSourceProcessHandle)) 
+        {
+            canDupe = TRUE; // this is the only case that is fairly certain to work
+        }
+        if (isSocket && hGetParentProcessId() == GetProcessId(hSourceProcessHandle)) { // handle is from parent process, it can be directly duplicated if inheritable
```
SergeGautherie, line 273:
```suggestion
        else if (isSocket && hGetParentProcessId() == GetProcessId(hSourceProcessHandle)) { // handle is from parent process, it can be directly duplicated if inheritable
```

dll/win32/kernel32/client/handle.c
```diff
@@ -187,6 +221,138 @@ DuplicateHandle(IN HANDLE hSourceProcessHandle,
...
+                    }
+                }			
+                _SEH2_EXCEPT(EXCEPTION_EXECUTE_HANDLER)
+                {
+                    DPRINT1("DuplicateHandle: SOCKET DUPLICATION exception (%lx) for duplicating handle(%lx) returning source handle (BADIMPL FIXME).\n", _SEH2_GetExceptionCode(), hSourceHandle);			
+                }
+                _SEH2_END;		
+            }
+            DPRINT("DuplicateHandle: SOCKET DUPLICATION completed, returning true.\n");									
+            return TRUE;
```
SergeGautherie, line 316:
```suggestion
```

dll/win32/kernel32/client/handle.c
```diff
@@ -187,6 +221,138 @@ DuplicateHandle(IN HANDLE hSourceProcessHandle,
...
+                    {
+                        DPRINT1("DuplicateHandle: CROSS PROCESS SOCKET DUPLICATION exception (%lx) for duplicating handle(%lx) reutrning source handle.\n", _SEH2_GetExceptionCode(), hSourceHandle);			
+                    }
+                    _SEH2_END;	
+                }
+                DPRINT("DuplicateHandle: SOCKET DUPLICATION completed, returning true.\n");							
+            } else DPRINT1("DuplicateHandle: SOCKET DUPLICATION SKIPPED (no ws2_32), returning true and source handle. (NOIMPL FIXME)\n");					
+            return TRUE;																			
+        }
+    } else DPRINT("DuplicateHandle: handle is not a socket, proceeding with NtDuplicateObject.\n");
```
SergeGautherie, line 354:
```suggestion
        }
        return TRUE;
    }
    DPRINT("DuplicateHandle: handle is not a socket, proceeding with NtDuplicateObject\n");
```

dll/win32/kernel32/client/handle.c
bigdude9: Rebuilt the if structure. I saw what you meant about the superfluous parts after I reoriented things as you suggested. Please look at the revisions and see if they match what you expected now. Sorry, I have looked at it so many times I'm code blind.

bigdude9:
the Get/SetSocketInformation are implemented with a layered approach and the scant docs there are seem to suggest the helper portion is supposed to be the 'last stop' for the oddments that cannot be done anywhere else. MSDN is the one that points at this handle duplication work being there. It's not a particularly clean place for it I agree, but winsock attempts to emulate berkeley sockets, and windows is nothing like BSD internally.. so there are bound to be massive kludges hidden in there somewhere (or everywhere!). Please review again and let me know if more changes are needed (for correctness). The horrible hackiness of it is not lost on me, but this has worked when tested(lightly) with SAMBA4 porting. I have a partially working SMBD server on ReactOS with these hacks in place (and it will not run without them).

dll/win32/kernel32/client/handle.c
```diff
@@ -187,6 +221,135 @@ DuplicateHandle(IN HANDLE hSourceProcessHandle,
...
+                        if (result != 0) {
+                            DPRINT1("DuplicateHandle: IN Process call to WSADuplicateSocketW failed, returned:%x.\n", result);
+                        } else {
+                            // now call wsasocket to finish the process and create new socket
+                            targetSocket = DynWSASocket(SharedSocketInfo.iAddressFamily, SharedSocketInfo.iSocketType, SharedSocketInfo.iProtocol, &SharedSocketInfo, 0, WSA_FLAG_OVERLAPPED); 
+                            // 	not sure we want overlapped set, it's the default for winsock 1.x but winsock 2 stopped setting it by default	
+                            if (targetSocket != INVALID_SOCKET) {
+                                DPRINT("DuplicateHandle: winsock socket handle succesfully duplicated, returning handle(%lx).\n", targetSocket);
+                                *lpTargetHandle = (HANDLE)targetSocket;							
+                            } else DPRINT1("DuplicateHandle: WSASocketW on cross process call returned INVALID_SOCKET, returning input handle(%lx) (BADIMPL FIXME).\n", hSourceHandle);
```
JoachimHenze, line 306:
```suggestion
                            } else DPRINT1("DuplicateHandle: WSASocketW on cross process call returned INVALID_SOCKET, returning input handle(%lx) (BADIMPL FIXME)\n", hSourceHandle);
```

dll/win32/kernel32/client/handle.c
```diff
@@ -187,6 +221,135 @@ DuplicateHandle(IN HANDLE hSourceProcessHandle,
...
+                            // 	not sure we want overlapped set, it's the default for winsock 1.x but winsock 2 stopped setting it by default	
+                            if (targetSocket != INVALID_SOCKET) {
+                                DPRINT("DuplicateHandle: winsock socket handle succesfully duplicated, returning handle(%lx).\n", targetSocket);
+                                *lpTargetHandle = (HANDLE)targetSocket;							
+                            } else DPRINT1("DuplicateHandle: WSASocketW on cross process call returned INVALID_SOCKET, returning input handle(%lx) (BADIMPL FIXME).\n", hSourceHandle);						
+                        }
+                    }			
+                    _SEH2_EXCEPT(EXCEPTION_EXECUTE_HANDLER)
+                    {
+                        DPRINT1("DuplicateHandle: SOCKET DUPLICATION exception (%lx) for duplicating handle(%lx) returning source handle (BADIMPL FIXME).\n", _SEH2_GetExceptionCode(), hSourceHandle);
```
JoachimHenze, line 311:
```suggestion
                        DPRINT1("DuplicateHandle: SOCKET DUPLICATION exception (%lx) for duplicating handle(%lx) returning source handle (BADIMPL FIXME)\n", _SEH2_GetExceptionCode(), hSourceHandle);
```

dll/win32/kernel32/client/handle.c
```diff
@@ -187,6 +221,135 @@ DuplicateHandle(IN HANDLE hSourceProcessHandle,
...
+                            } else DPRINT1("DuplicateHandle: WSASocketW on cross process call returned INVALID_SOCKET, returning input handle(%lx) (BADIMPL FIXME).\n", hSourceHandle);						
+                        }
+                    }			
+                    _SEH2_EXCEPT(EXCEPTION_EXECUTE_HANDLER)
+                    {
+                        DPRINT1("DuplicateHandle: SOCKET DUPLICATION exception (%lx) for duplicating handle(%lx) returning source handle (BADIMPL FIXME).\n", _SEH2_GetExceptionCode(), hSourceHandle);			
+                    }
+                    _SEH2_END;		
+                }
+                DPRINT("DuplicateHandle: SOCKET DUPLICATION completed, returning true.\n");
```
JoachimHenze, line 315:
```suggestion
                DPRINT("DuplicateHandle: SOCKET DUPLICATION completed, returning true\n");
```

dll/win32/kernel32/client/handle.c
```diff
@@ -187,6 +221,135 @@ DuplicateHandle(IN HANDLE hSourceProcessHandle,
...
+                    }			
+                    _SEH2_EXCEPT(EXCEPTION_EXECUTE_HANDLER)
+                    {
+                        DPRINT1("DuplicateHandle: SOCKET DUPLICATION exception (%lx) for duplicating handle(%lx) returning source handle (BADIMPL FIXME).\n", _SEH2_GetExceptionCode(), hSourceHandle);			
+                    }
+                    _SEH2_END;		
+                }
+                DPRINT("DuplicateHandle: SOCKET DUPLICATION completed, returning true.\n");									
+            } else { // !canDupe - winsock handle, but not from self or parent process, THIS WILL PROBABLY FAIL (FIXME)
+                DPRINT("DuplicateHandle: canDupe is false, attempting CROSS-PROCESS socket duplication.\n");
```
JoachimHenze, line 317:
```suggestion
                DPRINT("DuplicateHandle: canDupe is false, attempting CROSS-PROCESS socket duplication\n");
```

dll/win32/kernel32/client/handle.c
```diff
@@ -187,6 +221,135 @@ DuplicateHandle(IN HANDLE hSourceProcessHandle,
...
+                        if (result != 0) 
+                        {
+                            DPRINT1("DuplicateHandle: CROSS Process call to WSADuplicateSocketW failed, returned:%x.\n", result);
+                        } else {
+                            // now call wsasocket to finish the process and create new socket for us
+                            targetSocket = DynWSASocket(SharedSocketInfo.iAddressFamily, SharedSocketInfo.iSocketType, SharedSocketInfo.iProtocol, &SharedSocketInfo, 0, WSA_FLAG_OVERLAPPED); 
+                            // 	not sure we want overlapped set, it's the default for winsock 1.x but winsock 2 stopped setting it by default	
+                            if (targetSocket != INVALID_SOCKET) 
+                            {
+                                DPRINT("DuplicateHandle: winsock socket handle succesfully duplicated, returning handle(%lx).\n", targetSocket);
```
JoachimHenze, line 335:
```suggestion
                                DPRINT("DuplicateHandle: winsock socket handle succesfully duplicated, returning handle(%lx)\n", targetSocket);
```

dll/win32/kernel32/client/handle.c
```diff
@@ -187,6 +221,135 @@ DuplicateHandle(IN HANDLE hSourceProcessHandle,
...
+                                *lpTargetHandle = (HANDLE)targetSocket;							
+                            } else DPRINT1("DuplicateHandle: WSASocketW on cross process call returned INVALID_SOCKET, returning input handle(%lx) (BADIMPL FIXME).\n", hSourceHandle);
```
JoachimHenze, line 337:
```suggestion
                            } else DPRINT1("DuplicateHandle: WSASocketW on cross process call returned INVALID_SOCKET, returning input handle(%lx) (BADIMPL FIXME).\n", hSourceHandle);						
```
```suggestion
                            } else DPRINT1("DuplicateHandle: WSASocketW on cross process call returned INVALID_SOCKET, returning input handle(%lx) (BADIMPL FIXME)\n", hSourceHandle);
```

dll/win32/kernel32/client/handle.c
```diff
@@ -187,6 +221,135 @@ DuplicateHandle(IN HANDLE hSourceProcessHandle,
...
+                            if (targetSocket != INVALID_SOCKET) 
+                            {
+                                DPRINT("DuplicateHandle: winsock socket handle succesfully duplicated, returning handle(%lx).\n", targetSocket);
+                                *lpTargetHandle = (HANDLE)targetSocket;							
+                            } else DPRINT1("DuplicateHandle: WSASocketW on cross process call returned INVALID_SOCKET, returning input handle(%lx) (BADIMPL FIXME).\n", hSourceHandle);						
+                        }
+                    }			
+                    _SEH2_EXCEPT(EXCEPTION_EXECUTE_HANDLER)
+                    {
+                        DPRINT1("DuplicateHandle: CROSS PROCESS SOCKET DUPLICATION exception (%lx) for duplicating handle(%lx) reutrning source handle.\n", _SEH2_GetExceptionCode(), hSourceHandle);
```
JoachimHenze, line 342:
```suggestion
                        DPRINT1("DuplicateHandle: CROSS PROCESS SOCKET DUPLICATION exception (%lx) for duplicating handle(%lx) returning source handle\n", _SEH2_GetExceptionCode(), hSourceHandle);
```

dll/win32/kernel32/client/handle.c
```diff
@@ -187,6 +221,135 @@ DuplicateHandle(IN HANDLE hSourceProcessHandle,
...
+                            } else DPRINT1("DuplicateHandle: WSASocketW on cross process call returned INVALID_SOCKET, returning input handle(%lx) (BADIMPL FIXME).\n", hSourceHandle);						
+                        }
+                    }			
+                    _SEH2_EXCEPT(EXCEPTION_EXECUTE_HANDLER)
+                    {
+                        DPRINT1("DuplicateHandle: CROSS PROCESS SOCKET DUPLICATION exception (%lx) for duplicating handle(%lx) reutrning source handle.\n", _SEH2_GetExceptionCode(), hSourceHandle);			
+                    }
+                    _SEH2_END;	
+                }
+                DPRINT("DuplicateHandle: SOCKET DUPLICATION completed, returning true.\n");
```
JoachimHenze, line 346:
```suggestion
                DPRINT("DuplicateHandle: SOCKET DUPLICATION completed, returning true\n");
```

JoachimHenze (changes requested):
See my suggestions please. Also fix the wrong indentation character everywhere, and remove all the trailing whitespace (best to do that in Notepad++). Fix also some typos.

binarymaster (commented):
Next time please apply all suggestions in a single commit, because every single commit triggers the build, which takes almost a hour. I had to cancel a lot of builds to free the limited resources GitHub Actions share with us. Also, every commit sends an email to all participants (I had over 150 today).

dll/win32/kernel32/client/handle.c
```diff
@@ -13,8 +13,26 @@
...
 
+// private items for winsock socket handle duplication
+#define SystemHandleInformation 0x10
+#define SystemHandleInformationSize 3427780 * 2 // VALUE FROM WIN10 headers doubled, seems to rise every call in ROS
+
+ULONG_PTR hGetParentProcessId() // derived form Napalm @ NetCore2K from https://stackoverflow.com/questions/185254/how-can-a-win32-process-get-the-pid-of-its-parent
+{
+	ULONG_PTR pbi[6];
+	ULONG ulSize = 0;
+	if(NtQueryInformationProcess(GetCurrentProcess(), 0,
```
SergeGautherie, line 28: And we use `if ()`, not `if()`. Everywhere...

dll/win32/kernel32/client/handle.c
```diff
@@ -187,6 +221,134 @@ DuplicateHandle(IN HANDLE hSourceProcessHandle,
...
+        // we rely on a side effect of a failed call.. we get the name even though call failed
+        NtQueryObject(hSourceHandle, ObjectNameInformation, nameFull, sizeof(nameFull), &returnedLength);
+    }
+    _SEH2_EXCEPT(EXCEPTION_EXECUTE_HANDLER)
+    {
+        int excd = _SEH2_GetExceptionCode();
+        DPRINT1("DuplicateHandle: NtQueryObject generated an exception (%lx) for handle(%lx)\n", excd, hSourceHandle);			
+    }
+    _SEH2_END;
+    if (wcscmp((LPWSTR)(nameFull + (2 * sizeof(WCHAR))), L"\\Device\\Afd") == 0)
```
SergeGautherie, line 241: Document why `+ (2 * sizeof(WCHAR))`.

dll/win32/kernel32/client/handle.c
```diff
@@ -187,6 +221,134 @@ DuplicateHandle(IN HANDLE hSourceProcessHandle,
         return FALSE;
     }
 
+    // winsock socket handles code currently only dupes from self or inherited sockets from parent are implemented
+    // TODO: add socket duplication from arbitrary source process (works in Windows 2003)
+    WCHAR nameFull[80];
```
SergeGautherie, line 226: Why `80`?

dll/win32/kernel32/client/handle.c
```diff
@@ -187,6 +221,134 @@ DuplicateHandle(IN HANDLE hSourceProcessHandle,
...
+            if (getsockopt) 
+            {
+                // getsockopt is our gatekeeper. if it fails winsock access is not in a reliable state
+                int result = getsockopt((SOCKET)hSourceHandle, SOL_SOCKET, SO_TYPE, (char*)&optVal, &optLen);
+                if (result == 0) 
+                {
+                    if(optVal == SOCK_STREAM)
+                        DPRINT("(PID %lx) DuplicateHandle: socket is a TCP socket\n");
+                    else if(optVal == SOCK_DGRAM)
+                        DPRINT("DuplicateHandle: socket is a UDP socket\n");
```
SergeGautherie, line 260: Is `if(optVal == SOCK_DGRAM)` superfluous?
Or would we want an additional `else` case?

dll/win32/kernel32/client/handle.c
```diff
@@ -187,6 +221,134 @@ DuplicateHandle(IN HANDLE hSourceProcessHandle,
...
+                    dupwinsock = (wsdskcall)GetProcAddress(winsock_module, "WSADuplicateSocketW");
+                    isSocket = TRUE; // makes sockets fall back to normal handle duplication if we cannot dynlink
+                } else DPRINT1("Error calling getsockopt for inheritable socket\n");
+            } else DPRINT1("Error GetProcAddressA() could not Dynlink getsockopt\n");
+        } else DPRINT1("Error GetModuleHandleA() could not find WS2_32 module handle\n");
+        // NOTE: currently this code duplicates the handle/socket but DOES NOT CHAGE THE HANDLE ACCESS OR OPTIONS IN ANY WAY (ignores dwOption and access rights changes)
+        // TODO: revise code to respect access changes requested in parameters
+        if (isSocket) {
```
SergeGautherie, line 268: We put `{` on its own line. Everywhere...
For `else` too.

dll/win32/kernel32/client/handle.c
```diff
@@ -187,6 +221,134 @@ DuplicateHandle(IN HANDLE hSourceProcessHandle,
...
+            DynWSASocket = (wsskcall)GetProcAddress(winsock_module, "WSASocketW");
+            getsockopt = (wsgsocall)GetProcAddress(winsock_module, "getsockopt");
+            if (getsockopt) 
+            {
+                // getsockopt is our gatekeeper. if it fails winsock access is not in a reliable state
+                int result = getsockopt((SOCKET)hSourceHandle, SOL_SOCKET, SO_TYPE, (char*)&optVal, &optLen);
+                if (result == 0) 
+                {
+                    if(optVal == SOCK_STREAM)
+                        DPRINT("(PID %lx) DuplicateHandle: socket is a TCP socket\n");
```
SergeGautherie, line 258: Unmatched `%lx`.

ThFabba:
I'm removing all the style nitpick spam. This code isn't going to make it into master as-is so please stop obsessing over what it looks like. Instead, let's focus on how to solve this in a more appropriate way.

SergeGautherie:
> This code isn't going to make it into master as-is

Then, could this PR be marked as Draft?

bigdude9:
did general cleanup so the formatting questions should be resolved. also removed explanatory comments and dprints used for code dev tracing. I'm definitely open to a better way, this is kludgy but functional, but far from elegant.

bigdude9:
sorry about applying suggestions individually, didn't see a way to group them up..

binarymaster:
> sorry about applying suggestions individually, didn't see a way to group them up..

It can be done in the "Files changed" tab, there you can add suggestions to batch, and then commit the batch.

dll/win32/kernel32/client/handle.c
bigdude9: Removed whole section, it was mostly traces for testing.

PR #5488, opened 2023-07-26
[EXPLORER] Open windows with the correct position

cbialorucki:
## Purpose

Opens minimized windows from the correct position when using the taskbar. This also fixes an issue where minimizing and opening task manager causes that window to become blank when running our shell in Windows Server 2003.

JIRA issue: [CORE-19055](https://jira.reactos.org/browse/CORE-19055)

## Proposed changes

- Stop storing and setting window positions from the taskbar and instead rely on the window's existing position.

### DIFF
--- a/base/shell/explorer/taskswnd.cpp
+++ b/base/shell/explorer/taskswnd.cpp
@@ -82,7 +82,6 @@ typedef struct _TASK_ITEM
     PTASK_GROUP Group;
     INT Index;
     INT IconIndex;
-    WINDOWPLACEMENT wndpl;
 
     union
     {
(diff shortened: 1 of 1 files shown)


JoachimHenze (approved):

JoachimHenze:
The code that is removed here was added by e6bced7a353132121387c77ed16d67bd10e9d8d5
So how do https://jira.reactos.org/browse/CORE-13895 and especially https://jira.reactos.org/browse/CORE-18350 do behave now with your patch?

learn-more (approved):
These changes make sense in the context of the ticket, but I cannot guarantee that this won't break something else.

katahiromz (approved):
Please check if you can restore `desk.cpl` by clicking on Task button when `desk.cpl` got minimized by `Win+D`.

cbialorucki:
> Please check if you can restore `desk.cpl` by clicking on Task button when `desk.cpl` got minimized by `Win+D`.

Control panel applets, including `desk.cpl`, restore as normal after using `Win+D`.

katahiromz:
@cbialorucki Let's go.

PR #5491, opened 2023-07-27
[SHLWAPI][SHLWAPI_APITEST][SDK] Implement SHPropertyBag_ReadGUID etc.

katahiromz:
## Purpose

## Proposed changes

- Add `VariantArrayToBuffer` helper function.
- Implement `SHPropertyBag_ReadGUID` and `SHPropertyBag_ReadStream` functions.
- Modify `shlwapi.spec`.
- Modify `shlwapi_undoc.h`.
- Strengthen `SHPropertyBag` testcase in `shlwapi_apitest.exe`.
- Link `uuid.dll` and `ole32.dll` to `shlwapi_apitest.exe`.

## TODO

- [x] Do tests.

## Comparison

WinXP:
WinXP
Successful.

Win2k3:
Win2k3
Successful.

Win10:
Win10
Successful.

ReactOS:
after
Successful.

### DIFF
--- a/dll/win32/shlwapi/ordinal.c
+++ b/dll/win32/shlwapi/ordinal.c
@@ -5356,7 +5356,7 @@ BOOL VariantToBuffer(const VARIANT *varIn, void *pvDest, UINT cb)
     LONG LowerBound, UpperBound;
     SAFEARRAY *pArray;
 
-    if (varIn && V_VT(varIn) == (VT_ARRAY | VT_UI1))
+    if (varIn && V_VT(varIn) == (VT_ARRAY | VT_UI1)) /* Byte Array */
     {
         pArray = V_ARRAY(varIn);
         if (SafeArrayGetDim(pArray) == 1 &&
--- a/dll/win32/shlwapi/ordinal.c
+++ b/dll/win32/shlwapi/ordinal.c
@@ -5350,6 +5350,37 @@ DoDefault:
     return hr;
 }
 
+BOOL VariantToBuffer(const VARIANT *varIn, void *pv, UINT cb)
+{
+    void *pvData;
+    LONG lBound, uBound;
+
+    if (!varIn || V_VT(varIn) != (VT_ARRAY | VT_UI1) ||
+        SafeArrayGetDim(V_ARRAY(varIn)) != 1 ||
+        FAILED(SafeArrayGetLBound(V_ARRAY(varIn), 1, &lBound)) ||
+        FAILED(SafeArrayGetUBound(V_ARRAY(varIn), 1, &uBound)) ||
+        (uBound - lBound + 1 < cb) ||
+        FAILED(SafeArrayAccessData(V_ARRAY(varIn), &pvData)))
+    {
+        return FALSE;
+    }
+
+    CopyMemory(pv, pvData, cb);
+    SafeArrayUnaccessData(V_ARRAY(varIn));
+    return TRUE;
+}
+
+BOOL VariantToGUID(const VARIANT *varIn, GUID *pguid)
+{
+    if (VariantToBuffer(varIn, pguid, sizeof(*pguid)))
+        return TRUE;
+
+    if (varIn && V_VT(varIn) == VT_BSTR && V_BSTR(varIn))
+        return GUIDFromStringW(V_BSTR(varIn), pguid);
+
+    return FALSE;
+}
+
 /**************************************************************************
  *  SHPropertyBag_ReadType (SHLWAPI.493)
  */
--- a/dll/win32/shlwapi/ordinal.c
+++ b/dll/win32/shlwapi/ordinal.c
@@ -5671,6 +5671,54 @@ HRESULT WINAPI SHPropertyBag_ReadRECTL(IPropertyBag *ppb, LPCWSTR pszPropName, R
 }
 
 /**************************************************************************
+ *  SHPropertyBag_ReadGUID (SHLWAPI.505)
+ */
+HRESULT WINAPI SHPropertyBag_ReadGUID(IPropertyBag *ppb, LPCWSTR pszPropName, GUID *pguid)
+{
+    HRESULT hr;
+    BOOL bRet;
+    VARIANT vari;
+
+    if (!ppb || !pszPropName || !pguid)
+        return E_INVALIDARG;
+
+    hr = SHPropertyBag_ReadType(ppb, pszPropName, &vari, VT_EMPTY);
+    if (FAILED(hr))
+        return hr;
+
+    if (V_VT(&vari) == (VT_UI1 | VT_ARRAY)) /* Byte Array */
+        bRet = VariantToGUID(&vari, pguid);
+    else if (V_VT(&vari) == VT_BSTR)
+        bRet = GUIDFromStringW(V_BSTR(&vari), pguid);
+    else
+        bRet = TRUE;
+
+    VariantClear(&vari);
+    return (bRet ? S_OK : E_FAIL);
+}
+
+/**************************************************************************
+ *  SHPropertyBag_ReadStream (SHLWAPI.531)
+ */
+HRESULT WINAPI SHPropertyBag_ReadStream(IPropertyBag *ppb, LPCWSTR pszPropName, IStream **ppStream)
+{
+    HRESULT hr;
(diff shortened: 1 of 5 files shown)


dll/win32/shlwapi/ordinal.c
```diff
@@ -5670,6 +5699,70 @@ HRESULT WINAPI SHPropertyBag_ReadRECTL(IPropertyBag *ppb, LPCWSTR pszPropName, R
...
+        ERR("%p %s %p\n", ppb, debugstr_w(pszPropName), pguid);
+        return hr;
+    }
+
+    if (V_VT(&vari) == (VT_UI1 | VT_ARRAY)) /* Byte Array */
+        bRet = VariantToBuffer(&vari, pguid, sizeof(*pguid));
+    else if (V_VT(&vari) == VT_BSTR)
+        bRet = GUIDFromStringW(V_BSTR(&vari), pguid);
+    else
+        bRet = TRUE; /* This is by design */
```
HBelusca, line 5731: Maybe, but what could be the reason behind this?
Is this behaviour tested by your tests?

katahiromz: Reason I don't know. Yeah, tested.

JoachimHenze (approved):

dll/win32/shlwapi/ordinal.c
```diff
@@ -5350,6 +5350,35 @@ HRESULT VariantChangeTypeForRead(_Inout_ VARIANTARG *pvarg, _In_ VARTYPE vt)
     return hr;
 }
 
+BOOL
+VariantToBuffer(
```
HBelusca, line 5354: More like "VariantArrayToBuffer" (since it's only dealing with that?)
If this is really so, it seems you could do very simply:
```c
    /* Only supports byte array */
    if (!varIn || V_VT(varIn) != (VT_UI1 | VT_ARRAY))
        return FALSE;

    /* Boundary check and access */
    // Continue here as usual...
    return TRUE;
```

dll/win32/shlwapi/ordinal.c
HBelusca: OK, but in that case, what does the data buffer pointed by `pguid` contains on return? Garbage? or all zeroes?

modules/rostests/apitests/shlwapi/SHPropertyBag.cpp
```diff
@@ -168,6 +177,12 @@ static void SHPropertyBag_ReadTest(void)
     ok_long(hr, E_FAIL);
     ok_int(s_cRead, 2);
     ok_int(s_cWrite, 0);
+
+    ResetTest(VT_EMPTY, L"GUID1");
+    hr = SHPropertyBag_ReadGUID(&dummy, L"GUID1", &guid);
```
HBelusca, line 182: Is this the test that checks the `/* This is by design */` thing from above? If so it would be interesting to initialize the `guid` buffer to something different from zero, then check after the function is called, whether it still contains the original data, or whether it has been zeroed out ....

dll/win32/shlwapi/ordinal.c
katahiromz: It's a bug in WinXP/Win2k3. Win10 won't reproduce this behavior.
It's correctly returning a GUID value.

modules/rostests/apitests/shlwapi/SHPropertyBag.cpp
katahiromz: Added tests to the `SHPropertyBag` testcase.

dll/win32/shlwapi/ordinal.c
katahiromz: Improved.

dll/win32/shlwapi/ordinal.c
HBelusca: Hmmmm.... I don't know, but I would prefer in that case, that we actually fix the bug and return whatever correct GUID in that case (a null one perhaps).
Cc @ThFabba @learn-more for some advice on that.

modules/rostests/apitests/shlwapi/SHPropertyBag.cpp
```diff
@@ -29,6 +30,30 @@ static void ResetTest(VARTYPE vt,
...
+        SafeArrayDestroy(pArray);
+        return NULL;
+    }
+
+    CopyMemory(pvData, pvSrc, cbSize);
+    SafeArrayUnaccessData(pArray);
+
+    return pArray;
+}
+
```
HBelusca, line 56: (I cannot add github comments on the code below)
Why using `ok_int(0, 1);` in the `CDummyPropertyBag` functions below? (No comment is present to explain what's going on.)

katahiromz: `ok_int(0, 1)` is a just failure.

katahiromz: Added comments.

HBelusca: OK then please use instead the more understandable syntax:
`ok(FALSE, "An understandable error message that shows in the tests results screen\n");`
(and with a suitable error message)

modules/rostests/apitests/shlwapi/SHPropertyBag.cpp
```diff
@@ -64,13 +89,36 @@ class CDummyPropertyBag : public IPropertyBag
...
+                    V_VT(pvari) = (VT_UI1 | VT_ARRAY);
+                    V_ARRAY(pvari) = CreateByteArray(&IID_IShellLink, sizeof(IID));
+                    return S_OK;
+                }
+
+                if (lstrcmpiW(pszPropName, L"GUID2") == 0)
+                {
+                    V_VT(pvari) = VT_BSTR;
+                    V_BSTR(pvari) =
+                        SysAllocString(L"{00000000-0000-0000-C000-000000000046}"); // IID_IUnknown
```
HBelusca, line 107: why this value specifically?

katahiromz: Fixed with using `StringFromGUID2`.

modules/rostests/apitests/shlwapi/SHPropertyBag.cpp
katahiromz: Fixed.

modules/rostests/apitests/shlwapi/SHPropertyBag.cpp
HBelusca: what I meant is, why using this GUID and not another one?

katahiromz: Another GUID is also okay if the tests could be performed.

katahiromz: Another GUID is also okay if the tests could be performed.

katahiromz: Another GUID is also okay if the tests could be performed.

katahiromz: Another GUID is also okay if the tests could be performed.

katahiromz: Another GUID is also okay if the tests could be performed.

katahiromz: Another GUID is also okay if the tests could be performed.

katahiromz: Using another GUID is also okay.

katahiromz: I intended to choose famous and available one.

learn-more: Is there something wrong with using this guid?

HBelusca: I suppose nothing is wrong. I was just asking in order to know whether any GUID value was as good as any other for the present test, or whether that value was chosen because e.g. this was a value that was expected from testing on Windows.

PR #5494, opened 2023-07-28
[SHELL32] Fix error code for GetDisplayNameOf for the Control Panel

HuwCampbell:
S_False is a success response, but was being used as a fail condition. It just isn't right to do that.

## Purpose

I noticed (and I think others may have as well) that the response from the control panel was deceptive.

https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-ishellfolder-getdisplaynameof

JIRA issue: [CORE-18841](https://jira.reactos.org/browse/CORE-18841)

## Proposed changes

Just fix the response code

### DIFF
--- a/dll/win32/shell32/folders/CControlPanelFolder.cpp
+++ b/dll/win32/shell32/folders/CControlPanelFolder.cpp
@@ -488,8 +488,8 @@ HRESULT WINAPI CControlPanelFolder::GetUIObjectOf(HWND hwndOwner,
 */
 HRESULT WINAPI CControlPanelFolder::GetDisplayNameOf(PCUITEMID_CHILD pidl, DWORD dwFlags, LPSTRRET strRet)
 {
-    if (!pidl)
-        return S_FALSE;
+    if (!strRet || !pidl)
+        return E_INVALIDARG;
 
     PIDLCPanelStruct *pCPanel = _ILGetCPanelPointer(pidl);
 


binarymaster:
Question, does this mean that d02cb099c commit is incorrect and should be reverted?

Cc @Doug-Lyons for review also.

HBelusca:
This precisely reminds me of this comment by @learn-more in PR #5444:
https://github.com/reactos/reactos/pull/5444#pullrequestreview-1533544638
> This only hides an issue, so I am not in favor of doing this.
>
> Maybe CControlPanelFolder::GetDisplayNameOf is wrong, returning S_FALSE (which is a success code), and CFSDropTarget::_CopyItems is wrong for sure, how it builds paths is plainly incorrect (as indicated by the comment above it).

katahiromz (dismissed):

dll/win32/shell32/folders/CControlPanelFolder.cpp
```diff
@@ -488,8 +488,8 @@ HRESULT WINAPI CControlPanelFolder::GetUIObjectOf(HWND hwndOwner,
 */
 HRESULT WINAPI CControlPanelFolder::GetDisplayNameOf(PCUITEMID_CHILD pidl, DWORD dwFlags, LPSTRRET strRet)
 {
-    if (!pidl)
-        return S_FALSE;
+    if (!strRet || !pidl)
+        return E_INVALIDARG;
```
HBelusca, line 492: Are there other GetDisplayNameOf functions in our code base that may require such a similar fix?

Doug-Lyons:
@HuwCampbell, In spite of what the Microsoft article that you quoted says, what is actually done by Microsoft in their code is different and the ReactOS team and I have seen this type of mistake many times in the past. When I created the solution to #5444, Mark Jansen created a unit test to make sure that the ReactOS code matches what Microsoft does so that we can be sure that we are following the right standard. I am in the process of creating a PR to submit this to the ReactOS code base, but have not finished it yet. I have included the test in [CORE-18841](https://jira.reactos.org/browse/CORE-18841)  as https://jira.reactos.org/secure/attachment/66125/shell32_GetDisplayNameOf.patch. What it proves is that the Control Panel's return from a GetDisplayNameOf function with a NULL first parameter is actually 'S_FALSE', which is a success indication. I will post the link after I have created the PR here. This PR is not correct. Sorry :(

dll/win32/shell32/folders/CControlPanelFolder.cpp
HuwCampbell: Not that I spotted in a brief sweep.

learn-more: the PIDL return value here is actually correct, passing in a null pidl is not an error but a request to get the item name itself.

HuwCampbell:
I tested the test on Windows 10 and it looks like it's not just an XP bug.

I think that #5495 should include a comment in the code for the control panel which justifies this out of spec behaviour.

PR #5496, opened 2023-07-29
[BROWSEUI] Refactor settings code

cbialorucki:
## Purpose
Refactor the code for storing browseui settings. This works similarly to how our shell stores its settings from a software design standpoint. 

## Proposed changes
- Add a `settings.cpp` file and `BrowseUISettings` struct to load and save settings.
- Add a registry value to `hivedef.inf` for the locked toolbar state. This prevents a bug where the associated registry key cannot be opened or saved to.
- Add a new `BWM_SETTINGCHANGE` window message to refresh the UI on setting changes.
- Add a new `BWM_GETSETTINGPTR` window message to allow child windows to get a pointer to the settings struct in `shellbrowser`.
- Send the `BWM_SETTINGCHANGE` window message to every open window when saving settings to the registry.
- Send the `BWM_GETSETTINGPTR` window message from child windows to `shellbrowser` to get a pointer to its settings struct.

## TODO (in future PRs)
- Add additional settings to `browseui`.


dll/win32/browseui/precomp.h
```diff
@@ -65,4 +65,16 @@
...
+    BOOL fShowGoButton;
+    BOOL fStatusBarVisible;
+
+    BOOL Load();
+    BOOL Save();
+};
+
+extern BrowseUISettings gSettings;
+
+#endif /* _BROWSEUI_PCH_ */
```
binarymaster, line 80: No newline at end of file.

dll/win32/browseui/settings.cpp
```diff
@@ -0,0 +1,38 @@
...
+                                             L"StatusBarOther", FALSE, FALSE);
+
+    fShowGoButton = SHRegGetBoolUSValueW(L"Software\\Microsoft\\Internet Explorer\\Main",
+                                   L"ShowGoButton", FALSE, TRUE);
+
+    fLocked = SHRegGetBoolUSValueW(L"Software\\Microsoft\\Internet Explorer\\Toolbar",
+                                   L"Locked", FALSE, TRUE);
+
+    return TRUE;
+}
```
binarymaster, line 38: No newline at end of file.

dll/win32/browseui/CMakeLists.txt
```diff
@@ -36,7 +36,8 @@ list(APPEND SOURCE
     utility.cpp
     CProgressDialog.cpp
     CUserAssist.cpp
-    CUserAssist.h)
+    CUserAssist.h
+    settings.cpp)
```
binarymaster, line 40: I'd suggest to move `settings.cpp` here before `shellbrowser.cpp`.

dll/win32/browseui/settings.cpp
```diff
@@ -0,0 +1,38 @@
+/*
+ * PROJECT:     ReactOS Explorer
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
```
binarymaster, line 3:
```suggestion
 * PROJECT:     ReactOS browseui
 * LICENSE:     LGPL-2.1-or-later (https://spdx.org/licenses/LGPL-2.1-or-later)
```

cbialorucki: Done.

dll/win32/browseui/settings.cpp
```diff
@@ -0,0 +1,38 @@
+/*
+ * PROJECT:     ReactOS BrowseUI
+ * LICENSE:     LGPL-2.1-or-later (https://spdx.org/licenses/LGPL-2.1-or-later)
+ * PURPOSE:     Stores settings for the file explorer UI.
+ * COPYRIGHT:   Copyright 2023 Carl Bialorucki <cbialo2@outlook.com>
+ */
+
+#include "precomp.h"
+
+BrowseUISettings gSettings;
```
learn-more, line 10: This needs to be initialized here, or the struct needs a ctor.
Otherwise you are using uninitialized memory.

dll/win32/browseui/settings.cpp
binarymaster: Please keep it lowercase as in my suggestion, see:

- https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=PROJECT%3A+++++ReactOS+browseui
- https://git.reactos.org/?p=reactos.git&a=search&h=HEAD&st=grep&s=PROJECT%3A+++++ReactOS+BrowseUI

dll/win32/browseui/settings.cpp
cbialorucki: Added a constructor

HBelusca: Question: Should this be a global? Or a member of CShellBrowser?

cbialorucki: I think it should be global because in addition to `CShellBrowser` I already also need it in `CInternetToolbar`.  I would expect many other settings to be littered throughout browseui. The setting controlling if control panel is shown under 'My Computer' will probably also have to be done outside of `CShellBrowser`.

dll/win32/browseui/settings.cpp
cbialorucki: Changed to lowercase

binarymaster (approved):

dll/win32/browseui/settings.cpp
HBelusca: Since you're making it a global, does this mean that the constructor is called _**very early on**_ (by the CRT) when the dll is being loaded? Should it be so? Or should we wait until the DllMain() gets called (or even after gModule.Init() is called, see https://git.reactos.org/?p=reactos.git;a=blob;f=dll/win32/browseui/browseui.cpp;h=8634429bfc95b174a58fd397ddf8ec8f0f70a609;hb=HEAD#l170 )

learn-more: If it's a simple ctor (just setting values) it is fine to just have it as global, when it is doing complex stuff that _might_ depend on initialization in the rest of the DLL, it should not be a global because it runs before DllMain gets a chance to initialize stuff indeed.

HBelusca: When my comment was made, the constructor called the Load() function, which were doing reads from the registry...

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -3681,21 +3675,14 @@ LRESULT CShellBrowser::OnOrganizeFavorites(WORD wNotifyCode, WORD wID, HWND hWnd
 
 LRESULT CShellBrowser::OnToggleStatusBarVisible(WORD wNotifyCode, WORD wID, HWND hWndCtl, BOOL &bHandled)
```
HBelusca, line 3676: Is this the function called from the explorer menu when you click on "show/hide status bar" (or some similarly-named menu item)? If so, have you verified whether, on Windows, when clicking on that item, it applies automatically to *all* other explorer windows as well as any new future ones that will be opened?

dll/win32/browseui/internettoolbar.cpp
```diff
@@ -720,16 +715,10 @@ HRESULT CInternetToolbar::LockUnlockToolbars(bool locked)
     CDockSite                               *dockSite;
```
HBelusca, line 715: Should this lock-unlock state apply to all toolbars at once and the status saved so that all future opened explorer windows get the new state?

HBelusca (commented):
I have some little concerns about the reliance on the global settings state, see below.

dll/win32/browseui/shellbrowser.cpp
cbialorucki: In Windows Server 2003, after showing or hiding the status bar it applies to all open windows (after they navigate to a new location) and to all future explorer windows.
In ReactOS, this code only applies to new windows that opened.

dll/win32/browseui/internettoolbar.cpp
cbialorucki: The locked/unlocked status should be saved, this is a setting that applies to new browseui windows that are opened and matches the behavior of Windows Server 2003. However, I don't think we need to refresh all the open browseui windows. Even Windows Server 2003 doesn't do that.

dll/win32/browseui/settings.cpp
cbialorucki: I decided to use the default constructor to initialize the settings struct and to call `Load()` when shellbrowser is initialized.

dll/win32/browseui/internettoolbar.cpp
```diff
@@ -1727,9 +1715,8 @@ LRESULT CInternetToolbar::OnContextMenu(UINT uMsg, WPARAM wParam, LPARAM lParam,
     SHCheckMenuItem(contextMenu, IDM_TOOLBARS_ADDRESSBAR, IsBandVisible(ITBBID_ADDRESSBAND) == S_OK);
     SHCheckMenuItem(contextMenu, IDM_TOOLBARS_LINKSBAR, FALSE);
     SHCheckMenuItem(contextMenu, IDM_TOOLBARS_CUSTOMIZE, FALSE);
-    SHCheckMenuItem(contextMenu, IDM_TOOLBARS_LOCKTOOLBARS, fLocked);
-    goButtonChecked = SHRegGetBoolUSValueW(L"Software\\Microsoft\\Internet Explorer\\Main", L"ShowGoButton", FALSE, TRUE);
-    SHCheckMenuItem(contextMenu, IDM_TOOLBARS_GOBUTTON, goButtonChecked);
+    SHCheckMenuItem(contextMenu, IDM_TOOLBARS_LOCKTOOLBARS, gSettings.fLocked);
```
HBelusca, line 1718: If one clicks on the "lock toolbars" in one explorer window, this appears to retro-affect how the corresponding menu item is going to be shown **in other already-existing opened windows** (once one opens the corresponding menu). Is this a behaviour that also happens on Windows?

dll/win32/browseui/settings.cpp
learn-more: Which default constructor?
You did not define one, you just call it (which should be called implicitly already when constructing the object), now you also invoke the copy constructor or assignment operator, whilst still not initializing your member variables.

dll/win32/browseui/internettoolbar.cpp
cbialorucki: This does not affect existing windows because it only reads the registry when initialized. This is the same behavior as Windows Server 2003.

dll/win32/browseui/settings.cpp
HBelusca: If you have 100 member functions that have to act on all the members, for sure that adding yet a new member may require modifying all those functions :thinking:

learn-more: > I added a constructor and initialized each member variable in it. Is there a better way to do this? Now to add a new setting you have to update four places (the header, the ctor, the load function, and the save function).

If this module is compiled as c++11 code you can use in-class initializers, which means you can drop the entire constructor.

Then it would look like:
```
struct BrowseUISettings
{
    BOOL fLocked = FALSE;
    BOOL fShowGoButton = FALSE;
    BOOL fStatusBarVisible = FALSE;

    BOOL Load();
    BOOL Save();
};
```

learn-more (approved):

dll/win32/browseui/settings.cpp
learn-more: Some people prefer more work for the developer and less for the compiler, but most don't.

dll/win32/browseui/shellbrowser.cpp
HBelusca: Confirmed in Win2k3.

dll/win32/browseui/internettoolbar.cpp
HBelusca: I mean, the check-menu-item call looks now at the global `gSettings.fLocked` .... So if another window has already deselected this item, it seems now that for the current window, opening again the menu would show the menu item deselected as well, even if that current window still has its toolbars still shown in the old state (and same for the status bar too).
See picture below:

image

In addition (not shown here), when navigating to a new folder (in the window "2" above), the status bar will disappear (as discussed in another comment of ours below), *but* the toolbars stay in their original state (here, locked).

cbialorucki: Since each browseui window has its own copy of the settings struct and it doesn't reload its settings on navigation; this implementation allows the other window to keep its setting for the status bar and locked toolbar state. It does not cause a mismatch between the checked state of either item and the actual UI in that window.

HBelusca: How is that each window has its own copy of the settings struct? Aren't they all using the same global gSettings variable?

cbialorucki: I see what you are saying now. My bad.

binarymaster:
@cbialorucki please document the [Add refresh global UI method](https://github.com/reactos/reactos/pull/5496/commits/9cdb043261d765be159524cdfc3c8e5f1b6f94fc) commit behavior in the PR description.

I think this addition is more than just refactoring and changes the shell behavior.

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -3677,12 +3688,7 @@ LRESULT CShellBrowser::OnOrganizeFavorites(WORD wNotifyCode, WORD wID, HWND hWnd
...
 {
     gSettings.fStatusBarVisible = !gSettings.fStatusBarVisible;
-    if (fStatusBar)
-    {
-        ::ShowWindow(fStatusBar, gSettings.fStatusBarVisible ? SW_SHOW : SW_HIDE);
-        RepositionBars();
-    }
-
+    CShellBrowser::RefreshGlobalUISettings();
```
HBelusca, line 3691: The "CShellBrowser::" prefix may not be needed (since you call that function from within a member function of that class)

dll/win32/browseui/settings.cpp
```diff
@@ -0,0 +1,38 @@
...
+    return TRUE;
+}
+
+BOOL BrowseUISettings::Load()
+{
+    fStatusBarVisible = SHRegGetBoolUSValueW(L"Software\\Microsoft\\Internet Explorer\\Main",
+                                             L"StatusBarOther", FALSE, FALSE);
+
+    fShowGoButton = SHRegGetBoolUSValueW(L"Software\\Microsoft\\Internet Explorer\\Main",
+                                   L"ShowGoButton", FALSE, TRUE);
```
HBelusca, line 32:
```suggestion
    fShowGoButton = SHRegGetBoolUSValueW(L"Software\\Microsoft\\Internet Explorer\\Main",
                                         L"ShowGoButton", FALSE, TRUE);
```

dll/win32/browseui/internettoolbar.cpp
HBelusca: > but the toolbar state does not refresh

Also note that when I opened the context menu in window 2, it still shows the menu item that says "Lock tool bars" being enabled (checked) , contrary to what's in window 1.
Which makes me tell that your usages of `gSettings.fLocked`, for example this `SHCheckMenuItem(contextMenu, IDM_TOOLBARS_LOCKTOOLBARS, gSettings.fLocked);` call above (and some others down below), look "weird" and should instead use a per-window `fLocked` variable instead. (Which is of course initially initialized with the value of `gSettings.fLocked`, but then that works independently.)

cbialorucki: If we are really concerned about this, I think the best way is to use a local settings struct for shellbrowser and get rid of the `RefreshGlobalUI` hack.

cbialorucki: I don't think a separate `fLocked` variable is a good idea. The purpose of this PR is to make it easier to add setting persistence to BrowseUI. Maybe what we really need is some sort of refresh message when `gSettings.save()` is called.

cbialorucki:
> @cbialorucki please document the [Add refresh global UI method]
Updated the PR documentation. I ended up going a different route.

dll/win32/browseui/internettoolbar.cpp
```diff
@@ -714,45 +709,41 @@ HRESULT CInternetToolbar::CreateMenuBar(IShellMenu **pMenuBar)
 }
 
 HRESULT CInternetToolbar::LockUnlockToolbars(bool locked)
+{
+    if (locked != pSettings->fLocked)
+    {
+        pSettings->fLocked = locked;
+        pSettings->Save();
+        RefreshLockedToolbarState();
+        // TODO: refresh view menu?
```
HBelusca, line 718: This should happen automatically if:
- the `pSettings` pointer points to the cached settings member of the window to which the toolbar belongs;
- the window' view menu "lock-toolbar" item state is set according to what is stored in the window's cached settings member.

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -319,14 +318,14 @@ class CShellBrowser :
         return CComObjectRootEx<CComMultiThreadModelNoCS>::InternalRelease();
     }
 #endif
-
+    BrowseUISettings m_settings;
```
HBelusca, line 321: Can't it be kept amongst the other private members of the CShellBrowser class above?

cbialorucki: Moved.

dll/win32/browseui/internettoolbar.cpp
cbialorucki: What are you trying to say? Every call to lock or unlock the toolbar goes through this function. At no other point in `browseui` is `fLocked` ever set to anything (except during initialization). `RefreshToolbarState()` only refreshes the toolbar window view after setting the locked state, it doesn't do any locking on its own.

HBelusca: yeah this is what should happen (if pSettings is set to the correct thing as I explained).

cbialorucki: Right. I left that comment there because it was there before. No big deal.

dll/win32/browseui/internettoolbar.h
```diff
@@ -92,6 +92,7 @@ class CInternetToolbar :
     BOOL                                    fSizing;
     POINT                                   fStartPosition;
     LONG                                    fStartHeight;
+    BrowseUISettings                        *pSettings;
```
learn-more, line 95: Where is this initialized?

cbialorucki: `pSettings` is initialized using a `BWM_SETTINGCHANGE` window message. The lparam contains a pointer to the actual settings struct in its `shellbrowser` parent.

HBelusca: Doesn't it need to be initialized to NULL initially in the constructor?

dll/win32/browseui/internettoolbar.cpp
```diff
@@ -980,7 +970,7 @@ HRESULT STDMETHODCALLTYPE CInternetToolbar::ResizeBorderDW(LPCRECT prcBorder,
     ::GetWindowRect(fMainReBar, &availableBorderSpace);
     neededBorderSpace.left = 0;
     neededBorderSpace.top = availableBorderSpace.bottom - availableBorderSpace.top;
-    if (!fLocked)
+    if (pSettings->fLocked && !pSettings->fLocked)
```
HBelusca, line 973: Are you sure about that ??

cbialorucki: This is embarrassing 😅

dll/win32/browseui/internettoolbar.h
cbialorucki: Fixed.

dll/win32/browseui/internettoolbar.cpp
```diff
@@ -1928,3 +1917,14 @@ LRESULT CInternetToolbar::OnWinIniChange(UINT uMsg, WPARAM wParam, LPARAM lParam
 
     return lres;
 }
+
+LRESULT CInternetToolbar::OnBrowseUISettingChanged(UINT uMsg, WPARAM wParam, LPARAM lParam, BOOL& bHandled)
+{
+    if (!pSettings)
+        pSettings = (BrowseUISettings*)lParam;
```
HBelusca, line 1924: Note, that ideally it would be nice to have the internettoolbar's pSettings pointer initialized to a valid non-NULL value as soon as the internettoolbar object gets created, which should be, in principle, done when a CShellBrowser window gets created too.

cbialorucki: I agree, but I don't think it's possible given the architecture of our browseui. This is why I was so against moving the settings object to shell browser. The internettoolbar object is created using `CInternetToolbar_CreateInstance` and we only have access to its window handle.

dll/win32/browseui/precomp.h
```diff
@@ -65,4 +67,14 @@
 
 WINE_DEFAULT_DEBUG_CHANNEL(browseui);
 
+struct BrowseUISettings
+{
+    BOOL fLocked = FALSE;
```
tkreuzer, line 72: 😍

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -319,14 +319,13 @@ class CShellBrowser :
...
 #endif
-
     CShellBrowser();
     ~CShellBrowser();
     HRESULT Initialize();
 public:
     HRESULT BrowseToPIDL(LPCITEMIDLIST pidl, long flags);
     HRESULT BrowseToPath(IShellFolder *newShellFolder, LPCITEMIDLIST absolutePIDL,
-        FOLDERSETTINGS *folderSettings, long flags);
+                         FOLDERSETTINGS *folderSettings, long flags);
```
binarymaster, line 328:
```suggestion
        FOLDERSETTINGS *folderSettings, long flags);
```

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -319,14 +319,13 @@ class CShellBrowser :
         return CComObjectRootEx<CComMultiThreadModelNoCS>::InternalRelease();
     }
 #endif
-
     CShellBrowser();
```
binarymaster, line 322:
```suggestion

    CShellBrowser();
```

binarymaster (commented):
Diff reducen

dll/win32/browseui/internettoolbar.cpp
```diff
@@ -250,11 +250,11 @@ HRESULT CDockSite::GetRBBandInfo(REBARBANDINFOW &bandInfo)
         bandInfo.fStyle |= RBBS_BREAK;
     if (fDeskBandInfo.dwModeFlags & DBIMF_TOPALIGN)
         bandInfo.fStyle |= RBBS_TOPALIGN;
-    if (fFlags & ITF_NOGRIPPER || fToolbar->fLocked)
+    if (fFlags & ITF_NOGRIPPER || (fToolbar->pSettings && fToolbar->pSettings->fLocked))
         bandInfo.fStyle |= RBBS_NOGRIPPER;
     if (fFlags & ITF_NOTITLE)
         bandInfo.fStyle |= RBBS_HIDETITLE;
-    if (fFlags & ITF_GRIPPERALWAYS && !fToolbar->fLocked)
+    if (fFlags & ITF_GRIPPERALWAYS && (fToolbar->pSettings && !fToolbar->pSettings->fLocked))
```
HBelusca, line 257: I wonder whether this code can be called with the `fToolbar->pSettings` being NULL while `fToolBar` being "valid" (not NULL) ....
Otherwise adding an `ASSERT(fToolbar && fToolbar->pSettings);` to inform that this cannot happen _normally_ may be considered :thinking:

dll/win32/browseui/internettoolbar.cpp
```diff
@@ -714,45 +710,40 @@ HRESULT CInternetToolbar::CreateMenuBar(IShellMenu **pMenuBar)
 }
 
 HRESULT CInternetToolbar::LockUnlockToolbars(bool locked)
+{
+    if (locked != pSettings->fLocked)
```
HBelusca, line 714: Can this code be called with `pSettings` still being NULL ? (I hope it shouldn't)

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -783,11 +786,12 @@ HRESULT CShellBrowser::Initialize()
 
     fToolbarProxy.Initialize(m_hWnd, clientBar);
 
-    LoadSettings();
+    LoadCabinetState();
+    ::SendMessageW(fClientBars[BIInternetToolbar].hwnd, BWM_SETTINGCHANGE, 0, (LPARAM)&m_settings);
```
HBelusca, line 790: An alternative could be for the toolbar itself to query its parent window for the settings structure (via a window message) :thinking: (when e.g. it gets "attached" to it via the SetSite call or sthg....)

cbialorucki: That could be a better solution, I'll look into it. Thanks!

dll/win32/browseui/internettoolbar.cpp
cbialorucki: It's not possible. `LockUnlockToolbars` is only called in two places. One of those places is the context menu on the control itself. The other is in the view context menu for the parent window.

dll/win32/browseui/shellbrowser.cpp
cbialorucki: I've got something like this up and running, I will clean it up real quick.

dll/win32/browseui/internettoolbar.cpp
cbialorucki: Should I resolve this now? pSettings now gets set during `SiteSet()`

dll/win32/browseui/internettoolbar.cpp
cbialorucki: This check is no longer needed now that the settings for the InternetToolbar gets set on `SetSite()`

dll/win32/browseui/precomp.h
```diff
@@ -6,6 +6,8 @@
 #define WIN32_NO_STATUS
 #define _INC_WINDOWS
 #define COM_NO_WINDOWS_H
+#define BWM_SETTINGCHANGE (WM_USER + 300)
+#define BWM_GETSETTINGS (WM_USER + 301)
```
HBelusca, line 10: Can you move those two defines together with the struct BrowseUISettings definition below?

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -3800,6 +3788,26 @@ LRESULT CShellBrowser::RelayCommands(UINT uMsg, WPARAM wParam, LPARAM lParam, BO
...
+        ::ShowWindow(fStatusBar, m_settings.fStatusBarVisible ? SW_SHOW : SW_HIDE);
+        RepositionBars();
+    }
+
+    return 0;
+}
+
+LRESULT CShellBrowser::OnGetBrowseUISettings(UINT uMsg, WPARAM wParam, LPARAM lParam, BOOL& bHandled)
+{
+    return (LRESULT)&m_settings;
```
HBelusca, line 3808: This is bad usage of LRESULT. (See my previous comment.)
You should instead implement this handler so that it returns the pointer via e.g. the LPARAM.
```suggestion
    if (!lParam)
        return ERROR_INVALID_PARAMETER;

    *(BrowseUISettings**)lParam = &m_settings;
    return NO_ERROR;
```

dll/win32/browseui/internettoolbar.cpp
```diff
@@ -980,7 +971,7 @@ HRESULT STDMETHODCALLTYPE CInternetToolbar::ResizeBorderDW(LPCRECT prcBorder,
     ::GetWindowRect(fMainReBar, &availableBorderSpace);
     neededBorderSpace.left = 0;
     neededBorderSpace.top = availableBorderSpace.bottom - availableBorderSpace.top;
-    if (!fLocked)
+    if (pSettings && !pSettings->fLocked)
```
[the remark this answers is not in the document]
cbialorucki: Removed.

dll/win32/browseui/internettoolbar.cpp
```diff
@@ -1356,6 +1347,9 @@ HRESULT STDMETHODCALLTYPE CInternetToolbar::SetSite(IUnknown *pUnkSite)
         if (ownerWindow == NULL)
             return E_FAIL;
 
+        // Get browseui settings from owner window
+        pSettings = (BrowseUISettings*)::SendMessageW(ownerWindow, BWM_GETSETTINGS, 0, 0);
```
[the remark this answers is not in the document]
cbialorucki: Done.

HBelusca (approved):

binarymaster (approved):

dll/win32/browseui/internettoolbar.h
```diff
@@ -221,6 +224,7 @@ class CInternetToolbar :
         MESSAGE_HANDLER(WM_LBUTTONUP, OnLUp)
         MESSAGE_HANDLER(WM_MOUSEMOVE, OnMouseMove)
         MESSAGE_HANDLER(WM_WININICHANGE, OnWinIniChange)
+        MESSAGE_HANDLER(BWM_SETTINGCHANGE, OnBrowseUISettingChanged)
```
binarymaster, line 227:
```suggestion
        MESSAGE_HANDLER(BWM_SETTINGCHANGE, OnSettingsChange)
```

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -306,6 +305,7 @@ class CShellBrowser :
     IBindCtx                                *fHistoryBindContext;
     HDSA menuDsa;
     HACCEL m_hAccel;
+    BrowseUISettings m_settings;
```
binarymaster, line 308:
```suggestion
    ShellSettings m_settings;
```

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -624,6 +624,8 @@ class CShellBrowser :
     LRESULT OnRefresh(WORD wNotifyCode, WORD wID, HWND hWndCtl, BOOL &bHandled);
     LRESULT OnExplorerBar(WORD wNotifyCode, WORD wID, HWND hWndCtl, BOOL &bHandled);
     LRESULT RelayCommands(UINT uMsg, WPARAM wParam, LPARAM lParam, BOOL &bHandled);
+    LRESULT OnBrowseUISettingChanged(UINT uMsg, WPARAM wParam, LPARAM lParam, BOOL& bHandled);
+    LRESULT OnGetBrowseUISettingsPtr(UINT uMsg, WPARAM wParam, LPARAM lParam, BOOL& bHandled);
```
binarymaster, line 628:
```suggestion
    LRESULT OnSettingsChange(UINT uMsg, WPARAM wParam, LPARAM lParam, BOOL& bHandled);
    LRESULT OnGetSettingsPtr(UINT uMsg, WPARAM wParam, LPARAM lParam, BOOL& bHandled);
```

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -674,6 +676,8 @@ class CShellBrowser :
         COMMAND_RANGE_HANDLER(IDM_GOTO_TRAVEL_FIRSTTARGET, IDM_GOTO_TRAVEL_LASTTARGET, OnGoTravel)
         COMMAND_RANGE_HANDLER(IDM_EXPLORERBAND_BEGINCUSTOM, IDM_EXPLORERBAND_ENDCUSTOM, OnExplorerBar)
         MESSAGE_HANDLER(WM_COMMAND, RelayCommands)
+        MESSAGE_HANDLER(BWM_SETTINGCHANGE, OnBrowseUISettingChanged)
+        MESSAGE_HANDLER(BWM_GETSETTINGSPTR, OnGetBrowseUISettingsPtr)
```
binarymaster, line 680:
```suggestion
        MESSAGE_HANDLER(BWM_SETTINGCHANGE, OnSettingsChange)
        MESSAGE_HANDLER(BWM_GETSETTINGSPTR, OnGetSettingsPtr)
```

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -3800,6 +3788,30 @@ LRESULT CShellBrowser::RelayCommands(UINT uMsg, WPARAM wParam, LPARAM lParam, BO
...
+
+    return 0;
+}
+
+LRESULT CShellBrowser::OnGetBrowseUISettingsPtr(UINT uMsg, WPARAM wParam, LPARAM lParam, BOOL& bHandled)
+{
+     if (!lParam)
+        return ERROR_INVALID_PARAMETER;
+
+    *(BrowseUISettings**)lParam = &m_settings;
```
binarymaster, line 3811:
```suggestion
LRESULT CShellBrowser::OnSettingsChange(UINT uMsg, WPARAM wParam, LPARAM lParam, BOOL& bHandled)
{
    /* Refresh child windows */
    ::SendMessageW(fClientBars[BIInternetToolbar].hwnd, uMsg, wParam, lParam);

    /* Refresh status bar */
    if (fStatusBar)
    {
        ::ShowWindow(fStatusBar, m_settings.fStatusBarVisible ? SW_SHOW : SW_HIDE);
        RepositionBars();
    }

    return 0;
}

LRESULT CShellBrowser::OnGetSettingsPtr(UINT uMsg, WPARAM wParam, LPARAM lParam, BOOL& bHandled)
{
    if (!lParam)
        return ERROR_INVALID_PARAMETER;

    *(ShellSettings**)lParam = &m_settings;
```

PR #5498, opened 2023-07-29
[FASTFAT][NTDLL_APITEST] Add a semicolon after _SEH2_END

SergeGautherie:
## Purpose

Trivial code consistency.

## Proposed changes

- Write these uses like everywhere else.
Note: it seems to me no semicolon should be wanted for this macro (`_SEH2_END`). Anyway.

### DIFF
--- a/drivers/filesystems/fastfat/read.c
+++ b/drivers/filesystems/fastfat/read.c
@@ -87,7 +87,7 @@ FatOverflowPagingFileRead (
         RtlZeroMemory((AT), (BYTE_COUNT));                         \
     } _SEH2_EXCEPT(EXCEPTION_EXECUTE_HANDLER) {                    \
          FatRaiseStatus( IrpContext, STATUS_INVALID_USER_BUFFER ); \
-    } _SEH2_END                                                    \
+    } _SEH2_END;                                                   \
 }
 
 //
--- a/modules/rostests/apitests/ntdll/RtlMultipleAllocateHeap.c
+++ b/modules/rostests/apitests/ntdll/RtlMultipleAllocateHeap.c
@@ -23,7 +23,7 @@ static FN_RtlMultipleFreeHeap g_free = NULL;
     } _SEH2_EXCEPT(EXCEPTION_EXECUTE_HANDLER) { \
         threw = _SEH2_GetExceptionCode(); \
     } \
-    _SEH2_END \
+    _SEH2_END; \
     ok((ret) == (ret_expected), "ret excepted %d, but %d\n", (ret_expected), (ret)); \
     ok((err) == (err_expected), "err excepted %d, but %d\n", (err_expected), (err)); \
     ok((threw) == (threw_excepted), "threw excepted %d, but %d\n", (threw_excepted), (threw));
@@ -37,7 +37,7 @@ static FN_RtlMultipleFreeHeap g_free = NULL;
     } _SEH2_EXCEPT(EXCEPTION_EXECUTE_HANDLER) { \
         threw = _SEH2_GetExceptionCode(); \
     } \
-    _SEH2_END \
+    _SEH2_END; \
     ok((err) == (err_expected), "err excepted %d, but %d", (err_expected), (err)); \
     ok((threw) == (threw_excepted), "threw excepted %d, but %d\n", (threw_excepted), (threw));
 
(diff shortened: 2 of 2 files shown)


HBelusca (approved):

JoachimHenze (approved):

binarymaster:
I think this was done on purpose, so when you write the call like `MACRO(x, y, z);` you use semicolon at the end of line, and shouldn't be able to omit it.

This change is unnecessary as it doesn't fix anything.

SergeGautherie:
> I think this was done on purpose, so when you write the call like `MACRO(x, y, z);` you use semicolon at the end of line, and shouldn't be able to omit it.

Sorry, I don't understand what you mean:
What is different with these few calls? `_SEH2_END` macro does not take any parameter...

> This change is unnecessary as it doesn't fix anything.

This PR does fix consistency, as in bad examples. Exactly as 12e1919 included...
Please, reopen.

binarymaster:
What I mean is this:

```c
// macro definition
#define MACRO_NAME(X, Y, Z) {                                      \
    _SEH2_TRY {                                                    \
    } _SEH2_EXCEPT(EXCEPTION_EXECUTE_HANDLER) {                    \
    } _SEH2_END                                                    \
}

// macro usage
MACRO_NAME(X, Y, Z); // should compile correctly
MACRO_NAME(X, Y, Z)  // without semicolon, should throw compile time error
```

Adding a semicolon to the macro definition would allow the second usage scenario without error.

SergeGautherie:
@binarymaster, what you are pointing at about the local macro definitions seems unrelated to this PR.
About this PR, my comment would be exactly what Hermès wrote.
Please, reopen.

HBelusca:
> What I mean is this: [...]

When the examples are expanded, you get:
```c
// MACRO_NAME(X, Y, Z); // expands to the following:
{
     _SEH2_TRY {
     } _SEH2_EXCEPT(EXCEPTION_EXECUTE_HANDLER) {
     } _SEH2_END
};
```
which compiles fine, because:
- _SEH2_END with or without semicolon compiles OK (but it's more natural to have it with a semicolon);
- since the macro definition had the opening-closing brackets for code-block, those are still present after expansion, thus making a perfectly valid code block;
- the trailing semicolon gets added because it was present after the macro usage. This represents an empty C statement, which is valid.

The second example without the trailing semicolon expands to:
```c
// MACRO_NAME(X, Y, Z)  // without semicolon, should throw compile time error .... or does it?? (it doesn't and expands to...)

{
     _SEH2_TRY {
     } _SEH2_EXCEPT(EXCEPTION_EXECUTE_HANDLER) {
     } _SEH2_END
}
```
Again, for the very same reasons as stated previously, this compiles perfectly fine. The trailing semicolon, absent in that example, doesn't change anything, since now its corresponding "empty C statement" is just absent.

Where this whole story of semicolons would matter, is if the macro was defined like:
```c
#define MACRO_NAME(X, Y, Z) \
    _SEH2_TRY { \
   } _SEH2_EXCEPT(EXCEPTION_EXECUTE_HANDLER) { \
   } _SEH2_END
```
(which in that precise case, would _**still**_ work with or without trailing name in its usage point, because the SEH2 try-except-end block is also valid as is with or without semicolon).

In order to have a macro definition that requires a function-like usage syntax, you need the famous do{}while(0) construct:
```c
#define MACRO_NAME(X, Y, Z) \
do { \
    _SEH2_TRY { \
   } _SEH2_EXCEPT(EXCEPTION_EXECUTE_HANDLER) { \
   } _SEH2_END // Here, with or without semicolon, doesn't matter
} while(0)
```
Now, using it without a trailing semicolon: `MACRO_NAME(X, Y, Z)` would be invalid, because the while(0) of the do-while loop needs a trailing semicolon.

binarymaster:
Ok with the curly braces you've convinced me.

binarymaster (approved):

HBelusca:
@SergeGautherie your default primary github account mail needs to be changed.

SergeGautherie:
> [SergeGautherie](https://github.com/SergeGautherie) closed this

No : I renamed the branch through the web interface...

Please, reopen.

HBelusca:
Ah, if you renamed the branch **in your remote fork**, sadly github cannot follow the change and closes the PR (and disconnects it from the branch). That's a limitation. And as a result you need to recreate a new PR from the new renamed branch.

HBelusca:
Ah, if you renamed the branch **in your remote fork**, sadly github cannot follow the change and closes the PR (and disconnects it from the branch). That's a limitation. And as a result you need to recreate a new PR from the new renamed branch.

----

This being said I can try to reopen but not sure the PR will operate on your renamed branch though....

SergeGautherie:
GitHub did not warn about that consequence.

Initial branch restored...

tkreuzer:
Reopen

tkreuzer:
@SergeGautherie Can't reopen. You have to create a new PR.

SergeGautherie:
Damn GitHub. Thanks for trying.

SergeGautherie:
> your default primary github account mail needs to be changed.

I assume you are referring to https://github.com/reactos/reactos/pull/4881#discussion_r1022091027.
Then, my answer is "No: my GitHub Primary address is the one I want it to be".
If there is another reason for your request, please contact me privately.

PR #5500, opened 2023-07-29
[AFD] Add UDP vectored I/O capability - packet assembly in Afd*SocketWriteData, scatter code in SatisfyPacketRecvRequest. CORE-17526

bigdude9:
## Purpose

When afd gets a request to send or receive a UDP packet based on multiple parts (vectored i/o scatter/gather) using the request array, the parts must be combined before packet can be sent, but current code sends only the first buffer. Similar issue exists for UDP receive, it lacks code for scattering buffers and simply returns buffer overflow error if multiple array elements are used because it only processes first element. NOTE: this is a bugfix, UDP vectored I/O works in Windows

JIRA issue: [CORE-17526](https://jira.reactos.org/browse/CORE-17526)

Added a local buffer and loop to assemble packet data, copying the elements from userland in AfdPacketSocketWriteData and AfdConnectedSocketWriteData to bugfix the ReactOS functions to work like windows does. Added a loop to handle buffer scatter in SatisfyPacketRecvRequest as well. These changes add support for vectored i/o in UDP to ReactOS

### DIFF
--- a/drivers/network/afd/afd/write.c
+++ b/drivers/network/afd/afd/write.c
@@ -50,9 +50,11 @@ static NTSTATUS NTAPI SendComplete
     FCB->SendIrp.InFlightRequest = NULL;
     /* Request is not in flight any longer */
 
-    if( FCB->State == SOCKET_STATE_CLOSED ) {
+    if( FCB->State == SOCKET_STATE_CLOSED )
+    {
         /* Cleanup our IRP queue because the FCB is being destroyed */
-        while( !IsListEmpty( &FCB->PendingIrpList[FUNCTION_SEND] ) ) {
+        while( !IsListEmpty( &FCB->PendingIrpList[FUNCTION_SEND] ) )
+        {
             NextIrpEntry = RemoveHeadList(&FCB->PendingIrpList[FUNCTION_SEND]);
             NextIrp = CONTAINING_RECORD(NextIrpEntry, IRP, Tail.Overlay.ListEntry);
             NextIrpSp = IoGetCurrentIrpStackLocation( NextIrp );
@@ -71,10 +73,12 @@ static NTSTATUS NTAPI SendComplete
         return STATUS_FILE_CLOSED;
     }
 
-    if( !NT_SUCCESS(Status) ) {
+    if( !NT_SUCCESS(Status) )
+    {
         /* Complete all following send IRPs with error */
 
-        while( !IsListEmpty( &FCB->PendingIrpList[FUNCTION_SEND] ) ) {
+        while( !IsListEmpty( &FCB->PendingIrpList[FUNCTION_SEND] ) )
+        {
             NextIrpEntry =
                 RemoveHeadList(&FCB->PendingIrpList[FUNCTION_SEND]);
             NextIrp =
@@ -108,7 +112,8 @@ static NTSTATUS NTAPI SendComplete
     TotalBytesProcessed = 0;
     SendLength = Irp->IoStatus.Information;
     HaltSendQueue = FALSE;
-    while (!IsListEmpty(&FCB->PendingIrpList[FUNCTION_SEND]) && SendLength > 0) {
+    while (!IsListEmpty(&FCB->PendingIrpList[FUNCTION_SEND]) && SendLength > 0)
+    {
         NextIrpEntry = RemoveHeadList(&FCB->PendingIrpList[FUNCTION_SEND]);
         NextIrp = CONTAINING_RECORD(NextIrpEntry, IRP, Tail.Overlay.ListEntry);
         NextIrpSp = IoGetCurrentIrpStackLocation( NextIrp );
@@ -158,7 +163,8 @@ static NTSTATUS NTAPI SendComplete
 
     ASSERT(SendLength == 0);
 
-   if ( !HaltSendQueue && !IsListEmpty( &FCB->PendingIrpList[FUNCTION_SEND] ) ) {
+   if ( !HaltSendQueue && !IsListEmpty( &FCB->PendingIrpList[FUNCTION_SEND] ) )
+   {
         NextIrpEntry = FCB->PendingIrpList[FUNCTION_SEND].Flink;
         NextIrp = CONTAINING_RECORD(NextIrpEntry, IRP, Tail.Overlay.ListEntry);
         NextIrpSp = IoGetCurrentIrpStackLocation( NextIrp );
@@ -202,7 +208,8 @@ static NTSTATUS NTAPI SendComplete
 
         if (NextIrp != NULL)
         {
-            for( i = 0; i < SendReq->BufferCount; i++ ) {
+            for( i = 0; i < SendReq->BufferCount; i++ )
+            {
                 BytesCopied = MIN(SendReq->BufferArray[i].len, SpaceAvail);
 
                 Map[i].BufferAddress =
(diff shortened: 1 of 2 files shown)


drivers/network/afd/afd/write.c
```diff
@@ -569,6 +569,8 @@ AfdPacketSocketWriteData(PDEVICE_OBJECT DeviceObject, PIRP Irp,
     PAFD_FCB FCB = FileObject->FsContext;
     PAFD_SEND_INFO_UDP SendReq;
     KPROCESSOR_MODE LockMode;
+    INT FullSendLen, LoopIdx; // accumulator for full packet length
+    char pktbuf[4096];  // local packet assembly buffer
```
HBelusca, line 573: What justifies that 4096 is a good length? Why not 2048 or 65536 ?

drivers/network/afd/afd/write.c
```diff
@@ -645,22 +647,55 @@ AfdPacketSocketWriteData(PDEVICE_OBJECT DeviceObject, PIRP Irp,
...
-            if (Status != STATUS_PENDING)
-            {
-                NT_VERIFY(RemoveHeadList(&FCB->PendingIrpList[FUNCTION_SEND]) == &Irp->Tail.Overlay.ListEntry);
-                Irp->IoStatus.Status = Status;
-                Irp->IoStatus.Information = 0;
-                (void)IoSetCancelRoutine(Irp, NULL);
-                UnlockBuffers(SendReq->BufferArray, SendReq->BufferCount, FALSE);
-                UnlockRequest(Irp, IoGetCurrentIrpStackLocation(Irp));
-                IoCompleteRequest(Irp, IO_NETWORK_INCREMENT);
+            if (SendReq->BufferCount > 1) {
```
HBelusca, line 650: Please keep the code style of the rest of the file.

drivers/network/afd/afd/write.c
```diff
@@ -645,22 +647,55 @@ AfdPacketSocketWriteData(PDEVICE_OBJECT DeviceObject, PIRP Irp,
...
+                AFD_DbgPrint(MID_TRACE,("Assembling local packet buffer from %u Buffer Array elements, clearing buffer\n", SendReq->BufferCount));
+                RtlZeroMemory((&pktbuf[0]), 4096 );
```
HBelusca, line 652: Buffer size has been hardcoded here.

drivers/network/afd/afd/write.c
bigdude9: That is a valid question. I chose it based on remembering buffer sizes used in network code I have seen in the past which used it because other MAC layers supported it as max routable packet size(token ring I think). In truth it is more or less arbitrary because most UDP packets using it are small. I did consider 65536 (max allowable packet size in many places) or using 1550 (prevalent ethernet MTU size), but 2048 may be a better choice being the round base 2 size just above the ethernet MTU. As for hard coding, maybe a constant definition would be better ?   Make a suggestion with your preference and I'll accept it.

drivers/network/afd/afd/write.c
```diff
@@ -10,6 +10,8 @@
 
 #include "afd.h"
 
+#define PAD_BUFFER 2048;
```
HBelusca, line 13:
```suggestion
#define PAD_BUFFER 2048
```

drivers/network/afd/afd/write.c
HBelusca: The point of my comment is that there is a hardcoded size being used there, but then in the loop, you are copying (and appending) into that buffer data blocks, whose different sizes are known only at runtime. And so it can happen that crafted blocks could be sent with their sizes made in such a way that the copy operation would overflow that large _on-stack_ buffer and corrupt everything.

drivers/network/afd/afd/write.c
bigdude9: saw that right after I comitted and fixed

drivers/network/afd/afd/write.c
```diff
@@ -334,6 +336,8 @@ AfdConnectedSocketWriteData(PDEVICE_OBJECT DeviceObject, PIRP Irp,
     PAFD_SEND_INFO SendReq;
     UINT TotalBytesCopied = 0, i, SpaceAvail = 0, BytesCopied, SendLength;
     KPROCESSOR_MODE LockMode;
+    INT FullSendLen, LoopIdx; // accumulator for full packet length
+    char pktbuf[PAD_BUFFER];  // local packet assembly buffer
```
JoachimHenze, line 340:
```suggestion
    char pktbuf[PAD_BUFFER];  // local packet assembly buffer
```
I do recommend to use the Notepad++ function again, so there is no chance to miss any EOL whitespace.
I will not point to further hits, I used this suggestion just as an example to highlight at least one hit.

JoachimHenze (changes requested):

bigdude9:
this PR also needs bugfix tag

drivers/network/afd/afd/write.c
bigdude9: ran it through notepad plus and cleaned that as well as other training spaces. also rechecked to make sure there are no tab indents.

drivers/network/afd/afd/write.c
bigdude9: The code style of this file was already very mixed and most did not match the coding style guide. I went through the file and fixed the control structures throughout the file to match the coding style. I had originally followed the content style of the file rather than the guide.. hopefully this is what you intended.

bigdude9:
Please remove changes requested flag and add bugfix tag. This functionality exists in Windows 2003 and not ReactOS and the requested formatting changes have been made.

binarymaster:
@bigdude9 I've added `bugfix` label as you requested, but for the "changes requested" you need to contact @JoachimHenze (`@reactosfanboy` in the chat) to make sure you've addressed his comments.

drivers/network/afd/afd/write.c
JoachimHenze: I do confirm that you fixed that fine. Resolved.

JoachimHenze:
My previous review comments have been properly addressed, which brings me personally from rejection to **neutrality** for this PR.
I will nevertheless not actively approve it for 2 reasons.
1.) I do not understand the actual semantic changes and I currently am not aware of a specific tests which would strengthen my trust of these changes. You can change that by either writing a test, or do nothing and hope that someone else will approve without tests who actually does understand the code, and the implications of your changes.
2.) The PR is full of (needless/unrelated) changes changing the position of opening curly-braces. The old way of placing those braces is not forbidden by our coding style. And the new one is not perfect neither, because the non-curly-braces are still padded with spaces inside (which actually **is** violating our style). So you touched a lot of code here, which is suboptimal before **and** after your changes.
3.) What makes it even harder for me to see the point in doing that, is that the new code that you added to write.c also does not place the opening curly-braces in its own line neither. So you change some (apparently random) code here in style, and at the same time violate that desired style with the code you do add. Seems a bit strange to me.
image
I would hold it like @hbelusca suggested to keep that kind of style changes out entirely in lines that you do not necessarily have to touch for your fix.

bigdude9:
I took the comments about coding style to mean you wanted all of it reformatted to fit the style guide. I have no issue at all with reverting that part of the changes. Originally I had placed the braces on the opening line in my code in order to match the style of the rest of the file which though somewhat mixed, was predominantly in that style...

bigdude9:
Sorry for the confusion. In part the changes to style in other lines was the result of using the automatic formatting tools suggested in notepad++ which changed formatting globally. I'll go back and manually reverse changes outside my additions and do some more cleanup to get it closer to what you describe. 

Honestly I'm more concerned with the functionality than anything else. I have a highly patched up version of Cygwin 2.5.2 and samba 4.11.0 that will  run on ReactOS with all of my PRs in place (as well as on windows). It's only lightly tested at this point, but I have been able to get and put files from a share running smbd on ReactOS/FAT32.. I would really like to move forward with it since it's useful functionality currently missing in ReactOS and I have spent many hours working on it.

PR #5501, opened 2023-07-29
[GITHUB] build.yml: Upgrade LLVM version to 14

SergeGautherie:
## Purpose

Minimal fix for GitHub clang build error.
It looks like requested v13 conflicts with already installed(?) v14. Whatever.

Note:
LLVM 15 and 16 already exist. To be tried as a followup PR.

## Proposed changes

- clang: 13 -> 14, as fix.
- clang-cl: 13.0.1 -> 14.0.6, to match.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,7 +34,7 @@ jobs:
     - name: Install LLVM
       if: ${{ matrix.compiler == 'clang' }}
       run: |
-        export LLVM_VERSION=13
+        export LLVM_VERSION=14
         wget https://apt.llvm.org/llvm.sh
         chmod +x llvm.sh
         sudo ./llvm.sh $LLVM_VERSION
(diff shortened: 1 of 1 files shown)


HBelusca (approved):

binarymaster:
Since Serge never participates in the chat, it has to be mentioned here explicitly...

Problem was reported to upstream:
- https://github.com/llvm/llvm-project/issues/64120
- https://github.com/llvm/llvm-project/issues/64182

We would likely prefer the problem to be solved in upstream, so we could then just re-run failed workflows without updating configuration at our side.

It's not clear yet if we would want to upgrade LLVM.

SergeGautherie:
> We would likely prefer the problem to be solved in upstream

It looks like that happened somehow.

> It's not clear yet if we would want to upgrade LLVM.

I am not aware of reasons not to upgrade clang*. I'll let you decide.

binarymaster:
Postponed this for now. Could be reopened once we want to upgrade.

PR #5502, opened 2023-07-30
[GDI32] CreateDIBSection: Remove erroneous assignation.

HBelusca:
## Purpose

bmBits is only used and assigned on output. It points (holds the address) to the array of DIB bit values. The "Bits" parameter is however a pointer to a variable that will receive the address of that array. So it makes no sense to initially assign bmBits to the value of the Bits parameter...

### DIFF
--- a/win32ss/gdi/gdi32/objects/bitmap.c
+++ b/win32ss/gdi/gdi32/objects/bitmap.c
@@ -255,34 +255,36 @@ CreateDIBSection(
     HBITMAP hBitmap = NULL;
     PVOID bmBits = NULL;
 
-    pConvertedInfo = ConvertBitmapInfo(BitmapInfo, Usage, &ConvertedInfoSize,
-    FALSE);
-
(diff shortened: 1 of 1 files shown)


SergeGautherie (approved):
While here, I would suggest to move the `if (!hBitmap)` check to immediately after the assignment
and to drop its `{}`.
This check seems useless and could even be removed, unless this extra safety is really wanted for some reason.

SergeGautherie:
bugfix: Current code is useless, but not a bug, isn't it?

HBelusca:
> bugfix: Current code is useless, but not a bug, isn't it?

It's a potential bug waiting to happen.

HBelusca:
Neither Test KVM: https://reactos.org/testman/compare.php?ids=88595,88598
and Test VBox: https://reactos.org/testman/compare.php?ids=88594,88599
doesn't show any visible regressions.

katahiromz (approved):

JoachimHenze (approved):

PR #5504, opened 2023-07-31
[VIDEOPRT] Fix legacy videoprt drivers bus interrupts

StartForKiller:
## Purpose

InterruptVector is obtained inside VideoPortGetAccessRanges, that's called inside HwFindAdapter,
but IntVideoPortFindAdapter is setting BusInterruptVector before HwFindAdapter call.

This was causing legacy drivers to not get interrupts. This maybe fixes some video drivers not working right now.
I don't see any JIRA issue for this so i'm directly doing this PR.

Edit: I found another bug on VideoPortGetBusData and VideoPortSetBusData that had the legacy vs pnp logic reversed.

JIRA issue: 

## Proposed changes

- Move the BusInterruptVector and BusInterruptLevel write right below the HwFindAdapter call
- Edit: change VideoPortGetBusData and VideoPortSetBusData if not equal to if equal PhysicalDeviceObject to NULL

### DIFF
--- a/win32ss/drivers/videoprt/resource.c
+++ b/win32ss/drivers/videoprt/resource.c
@@ -875,6 +875,10 @@ VideoPortGetAccessRanges(
         {
             DeviceExtension->InterruptLevel = Descriptor->u.Interrupt.Level;
             DeviceExtension->InterruptVector = Descriptor->u.Interrupt.Vector;
+            if(DriverExtension->InitializationData.HwInterrupt != NULL) {
+                DeviceExtension->ConfigInfo.BusInterruptLevel = DeviceExtension->InterruptLevel;
+                DeviceExtension->ConfigInfo.BusInterruptVector = DeviceExtension->InterruptVector;
+            }
             if (Descriptor->ShareDisposition == CmResourceShareShared)
                 DeviceExtension->InterruptShared = TRUE;
             else
--- a/win32ss/drivers/videoprt/resource.c
+++ b/win32ss/drivers/videoprt/resource.c
@@ -1189,7 +1189,7 @@ VideoPortGetBusData(
    if (BusDataType != Cmos)
    {
       /* Legacy vs. PnP behaviour */
-      if (DeviceExtension->PhysicalDeviceObject != NULL)
+      if (DeviceExtension->PhysicalDeviceObject == NULL)
          SlotNumber = DeviceExtension->SystemIoSlotNumber;
    }
 
@@ -1224,7 +1224,7 @@ VideoPortSetBusData(
    if (BusDataType != Cmos)
    {
       /* Legacy vs. PnP behaviour */
-      if (DeviceExtension->PhysicalDeviceObject != NULL)
+      if (DeviceExtension->PhysicalDeviceObject == NULL)
          SlotNumber = DeviceExtension->SystemIoSlotNumber;
    }
 
--- a/win32ss/drivers/videoprt/videoprt.c
+++ b/win32ss/drivers/videoprt/videoprt.c
@@ -437,8 +437,6 @@ IntVideoPortFindAdapter(
     ConfigInfo.DriverRegistryPath = DriverExtension->RegistryPath.Buffer;
     ConfigInfo.VideoPortGetProcAddress = IntVideoPortGetProcAddress;
     ConfigInfo.SystemIoBusNumber = DeviceExtension->SystemIoBusNumber;
-    ConfigInfo.BusInterruptLevel = DeviceExtension->InterruptLevel;
-    ConfigInfo.BusInterruptVector = DeviceExtension->InterruptVector;
 
     Status = ZwQuerySystemInformation(SystemBasicInformation,
                                       &SystemBasicInfo,
@@ -525,6 +523,9 @@ IntVideoPortFindAdapter(
         Status = STATUS_UNSUCCESSFUL;
         goto Failure;
     }
+
+    ConfigInfo.BusInterruptLevel = DeviceExtension->InterruptLevel;
+    ConfigInfo.BusInterruptVector = DeviceExtension->InterruptVector;
 
     /*
      * Now we know the device is present, so let's do all additional tasks
--- a/win32ss/drivers/videoprt/videoprt.c
+++ b/win32ss/drivers/videoprt/videoprt.c
@@ -419,24 +419,26 @@ IntVideoPortFindAdapter(
     PVIDEO_PORT_DEVICE_EXTENSION DeviceExtension;
     NTSTATUS Status;
     VP_STATUS vpStatus;
-    VIDEO_PORT_CONFIG_INFO ConfigInfo;
+    PVIDEO_PORT_CONFIG_INFO ConfigInfo;
     SYSTEM_BASIC_INFORMATION SystemBasicInfo;
     UCHAR Again = FALSE;
     BOOL LegacyDetection = FALSE;
 
     DeviceExtension = (PVIDEO_PORT_DEVICE_EXTENSION)DeviceObject->DeviceExtension;
 
+    ConfigInfo = &DeviceExtension->ConfigInfo;
+
     /* Setup a ConfigInfo structure that we will pass to HwFindAdapter. */
-    RtlZeroMemory(&ConfigInfo, sizeof(VIDEO_PORT_CONFIG_INFO));
-    ConfigInfo.Length = sizeof(VIDEO_PORT_CONFIG_INFO);
-    ConfigInfo.AdapterInterfaceType = DeviceExtension->AdapterInterfaceType;
-    if (ConfigInfo.AdapterInterfaceType == PCIBus)
-        ConfigInfo.InterruptMode = LevelSensitive;
+    RtlZeroMemory(ConfigInfo, sizeof(VIDEO_PORT_CONFIG_INFO));
+    ConfigInfo->Length = sizeof(VIDEO_PORT_CONFIG_INFO);
+    ConfigInfo->AdapterInterfaceType = DeviceExtension->AdapterInterfaceType;
+    if (ConfigInfo->AdapterInterfaceType == PCIBus)
+        ConfigInfo->InterruptMode = LevelSensitive;
     else
-        ConfigInfo.InterruptMode = Latched;
-    ConfigInfo.DriverRegistryPath = DriverExtension->RegistryPath.Buffer;
-    ConfigInfo.VideoPortGetProcAddress = IntVideoPortGetProcAddress;
-    ConfigInfo.SystemIoBusNumber = DeviceExtension->SystemIoBusNumber;
+        ConfigInfo->InterruptMode = Latched;
+    ConfigInfo->DriverRegistryPath = DriverExtension->RegistryPath.Buffer;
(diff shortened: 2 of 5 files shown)


HBelusca (changes requested):
https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/video/nc-video-pvideo_hw_find_adapter

> If the driver does not handle interrupts, HwVidFindAdapter should set both BusInterruptLevel and BusInterruptVector in the VIDEO_PORT_CONFIG_INFO structure to zero after its call to VideoPortGetAccessRanges. If both members are zero, the video port driver does not connect the interrupt for the miniport driver. Explicitly setting both BusInterruptLevel and BusInterruptVector to zero in HwVidFindAdapter disables the [HwVidInterrupt](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/video/nc-video-pvideo_hw_interrupt) entry point, if any, that was set up by the miniport driver's DriverEntry routine.

What you are doing here in your patch is to overwrite what the video miniport driver might have set inside `ConfigInfo.BusInterruptLevel` and `ConfigInfo.BusInterruptVector` by the "default" values from the device extension, that should have instead be set _before_ the call to the HwVidFindAdapter callback (in other words, as it currently being done before the PR).

Did you notice any "improvement" in behaviour in your own tests after your patch?

StartForKiller:
@HBelusca 
Yeah I found that now the interrupts work.

Anyways, I can add a manual check for that

StartForKiller:
@HBelusca  As you said yeah, it should be set to zero if HwInterrupt is NULL but actually that is also check on the Io connection itself, so even if the BusInterrupt* is set to some value the function won't connect the IO

win32ss/drivers/videoprt/videoprt.c
```diff
@@ -526,6 +524,12 @@ IntVideoPortFindAdapter(
         goto Failure;
     }
 
+    if(DriverExtension->InitializationData.HwInterrupt != NULL)
```
HBelusca, line 527:
```suggestion
    if (DriverExtension->InitializationData.HwInterrupt != NULL)
```

win32ss/drivers/videoprt/videoprt.c
```diff
@@ -526,6 +524,12 @@ IntVideoPortFindAdapter(
...
+    {
+        ConfigInfo.BusInterruptLevel = DeviceExtension->InterruptLevel;
+        ConfigInfo.BusInterruptVector = DeviceExtension->InterruptVector;
```
HBelusca, line 530: You're still erasing what the miniport driver has set into the ConfigInfo.BusInterrupt*** members with something else. This is wrong!

StartForKiller: > BusInterruptLevel
> 
> The bus-relative interrupt request level. Storport makes no assumptions about the HBA's interrupt usage, so the default value is zero. Storport initializes this member and miniport drivers must not modify it.

StartForKiller: @HBelusca As yo can see miniport must not modify this

HBelusca: But you're reading me the documentation of the storage ports, **AND NOT** the one of the **video** miniport drivers :rofl:

https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/video/ns-video-_video_port_config_info
clearly says:
```
BusInterruptLevel

This member is irrelevant if the video hardware does not generate interrupts, indicated by setting the HwInterrupt member in the [VIDEO_HW_INITIALIZATION_DATA](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/video/ns-video-_video_hw_initialization_data) structure to NULL. Otherwise, it specifies the bus-relative IRQL that corresponds to the interrupt request on Isa or MicroChannel type buses. _**The preset default value for this member is zero. A miniport driver must supply the correct value if it handles interrupts for a video adapter on an Isa or MicroChannel type bus, or for a video adapter configured for level-sensitive interrupts on an Eisa type bus.**_

If a miniport driver's [HwVidFindAdapter](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/video/nc-video-pvideo_hw_find_adapter) function finds that the video adapter does not generate interrupts or that it cannot determine a valid interrupt vector/level for the adapter, _**HwVidFindAdapter should set both BusInterruptLevel and BusInterruptVector to zero.**_

BusInterruptVector

This member is irrelevant if the video hardware does not generate interrupts, indicated by setting the HwInterrupt member in the VIDEO_HW_INITIALIZATION_DATA structure to NULL. Otherwise, it specifies the bus-relative vector used by the video hardware on I/O buses that use interrupt vectors, such as PCI buses. The preset default value for this member is zero.

If a miniport driver's HwVidFindAdapter function finds that the video adapter does not generate interrupts or that it cannot determine a valid interrupt vector/level for the adapter, HwVidFindAdapter should set both BusInterruptVector and BusInterruptLevel to zero.
```

(Also, searching for legacy video miniports, e.g. those in the NT4 DDK, would show you _actual_ usage examples.)

StartForKiller: @HBelusca I'm reading windows 2000 samples drivers and headers. The sample drivers that uses interrupts don't touch the BysInterruptVector and Level, maybe the behaviour has changed since then?

StartForKiller: @HBelusca the latest changes should make the difference

StartForKiller:
Okey, review it when you have time @HBelusca :D

win32ss/drivers/videoprt/videoprt.c
```diff
@@ -419,34 +419,34 @@ IntVideoPortFindAdapter(
     PVIDEO_PORT_DEVICE_EXTENSION DeviceExtension;
     NTSTATUS Status;
     VP_STATUS vpStatus;
-    VIDEO_PORT_CONFIG_INFO ConfigInfo;
+    PVIDEO_PORT_CONFIG_INFO ConfigInfo;
```
oleg-dubinskiy, line 422: Any reason to change the variable to a pointer here? You storing the memory address of the `VIDEO_PORT_CONFIG_INFO` variable from `VIDEO_PORT_DEVICE_EXTENSION` at line 429, not even a pointer, and you don't do any allocation from dynamical memory. You can simply pass its memory address via the `&` operator where it is needed, as it was done before. So it seems needless.

oleg-dubinskiy (commented):
Still one thing to be improved...

win32ss/drivers/videoprt/resource.c
```diff
@@ -875,6 +875,10 @@ VideoPortGetAccessRanges(
         {
             DeviceExtension->InterruptLevel = Descriptor->u.Interrupt.Level;
             DeviceExtension->InterruptVector = Descriptor->u.Interrupt.Vector;
+            if(DriverExtension->InitializationData.HwInterrupt != NULL) {
```
binarymaster, line 878:
```suggestion
            if (DriverExtension->InitializationData.HwInterrupt != NULL)
            {
```

win32ss/drivers/videoprt/videoprt.c
binarymaster: I don't get your review comment. The memory address of the variable is a pointer by definition. He changed the structure on stack to the pointer and then using it.

oleg-dubinskiy: > He changed the structure on stack to the pointer and then using it.

@binarymaster: but he didn't do it in the device extension. Just look at line 113 in videoprt.h:
image
It's declared not as `PVIDEO_PORT_CONFIG_INFO`, simply as `VIDEO_PORT_CONFIG_INFO` instead. So it's not a pointer and cannot be such, otherwise it also will need the additional allocation/freeing.
Then, he simply assigns its memory address via `&` to the locally declared pointer:
image
It makes sense only when assigning pointer to pointer, e. g.
```
PVIDEO_PORT_CONFIG_INFO ConfigInfo;
ConfigInfo = DeviceExtension->ConfigInfo;
```
in case `DeviceExtension->ConfigInfo` is a pointer too. But it actually isn't.
My point is, it's not necessary needed, and with my suggestion, I simply mean the following:
```
// declare it as a normal variable instead:
VIDEO_PORT_CONFIG_INFO ConfigInfo;
// then assign it without & operator:
ConfigInfo = DeviceExtension->ConfigInfo;
// use it as normal variable (with & and use . instead of ->):
    RtlZeroMemory(&ConfigInfo, sizeof(VIDEO_PORT_CONFIG_INFO));
    ConfigInfo.Length = sizeof(VIDEO_PORT_CONFIG_INFO);
    ConfigInfo.AdapterInterfaceType = DeviceExtension->AdapterInterfaceType;
    if (ConfigInfo.AdapterInterfaceType == PCIBus)
        ConfigInfo.InterruptMode = LevelSensitive;
// and pass its memory address via the & in the calls which require the pointer:
            vpStatus = DriverExtension->InitializationData.HwFindAdapter(
                         &DeviceExtension->MiniPortDeviceExtension,
                         DriverExtension->HwContext,
                         NULL,
                         &ConfigInfo,
                         &Again);
if (!IntVideoPortSetupInterrupt(DeviceObject, DriverExtension, &ConfigInfo))
```
Like it was done before.
I guess it's needless using a pointer when doing that is not required at all.
Everything other in this PR is fine to me.

binarymaster: > It's declared not as PVIDEO_PORT_CONFIG_INFO, simply as VIDEO_PORT_CONFIG_INFO instead. So it's not a pointer

That's right, it's not a pointer, but the structure data, and it is stored within the extension data (on the stack or whatever).

> So it's not a pointer and cannot be such, otherwise it also will need the additional allocation/freeing.

You access the data by memory address, which IS a pointer and which could be assigned to a pointer variable of the same type.

> Then, he simply assigns its memory address via & to the locally declared pointer

Yes, and it totally makes sense.

> It makes sense only when assigning pointer to pointer

But again, the memory address is what a pointer contains. Please update your knowledge base.

> ```c
> // declare it as a normal variable instead:
> VIDEO_PORT_CONFIG_INFO ConfigInfo;
> // then assign it without & operator:
> ConfigInfo = DeviceExtension->ConfigInfo;
> ```

This is not permitted in C language, you cannot assign a struct to struct. You either assign struct members manually or use memcpy / RtlCopyMemory.

tkreuzer: It's an alias. A "shortcut" name. So you can write `ConfigInfo` instead of `&DeviceExtension->ConfigInfo`.
That is all there is to it.

tkreuzer: > This is not permitted in C language, you cannot assign a struct to struct. You either assign struct members manually or use memcpy / RtlCopyMemory.

That is simply wrong.

binarymaster: Alright, but then I don't see a reason to keep structs in both places. Using a pointer (or a "shortcut" as you called it) in the function makes perfect sense and avoids duplicating the struct.

StartForKiller: That's why I used a pointer. Anyways, i tried to keep my code the most tiny as possible

DarkFire01:
Reopen if its still relevant

PR #5509, opened 2023-08-01
[WALLPAPERS] Add a new Wallpaper "Underwater"

user0-tb:
## Purpose

Add a new Wallpaper with the name "Underwater"

## Proposed changes

A better Wallpaper selection


JoachimHenze:
I would recommend to delete another wallpaper in return, that we do consider the least professional one.
If we wouldn't come to any conclusion, I would vote with a thumbs down for adding another one.

ace-dent:
This is quite a large file. Please consider whether a more optimized version is appropriate; does it need ~700 bytes of colorprofile metadata? (Check with `exiftool`). Artistically speaking, the photograph also seems rotated?

user0-tb:
@JoachimHenze and @ace-dent, the File is now replaced with a smaller one.

HBelusca:
We should consider having a separate github repo for the wallpapers.....

cbialorucki:
> We should consider having a separate github repo for the wallpapers.....

Perhaps better still, what if we added a "Wallpaper" section to RAPPS?

HBelusca:
> > We should consider having a separate github repo for the wallpapers.....
> 
> Perhaps better still, we could add a "Wallpaper" section to RAPPS.

Having a Themes/Wallpapers section in RAPPS would be a plus, yes. But those themes/wallpapers (if they are for ReactOS) should also be stored somewhere (hence the repo suggestion too).

PR #5511, opened 2023-08-02
[SHLWAPI][SHLWAPI_APITEST][SDK] SHCreatePropertyBagOnMemory

katahiromz:
## Purpose

## Proposed changes

- Add `dll/win32/shlwapi/propbag.cpp`.
- Implement `CBasePropertyBag` and `CMemPropertyBag` classes.
- Implement `SHCreatePropertyBagOnMemory` function by using these classes.
- Strengthen `SHPropertyBag` testcase in `shlwapi_apitest.exe`.
- Modify `shlwapi.spec`.
- Modify `shlwapi_undoc.h`.

## TODO

- [x] Do small tests.

## Comparison

WinXP:
WinXP
Successful.

Win2k3:
Win2k3
Successful.

Win10:
Win10
Successful.

ReactOS (AFTER):
after
Successful.

### DIFF
--- a/dll/win32/shlwapi/CMakeLists.txt
+++ b/dll/win32/shlwapi/CMakeLists.txt
@@ -24,6 +24,7 @@ list(APPEND SOURCE
 
 list(APPEND PCH_SKIP_SOURCE
     assoc.c
+    propbag.cpp
     wsprintf.c
     ${CMAKE_CURRENT_BINARY_DIR}/shlwapi_stubs.c)
 
@@ -36,7 +37,9 @@ add_library(shlwapi MODULE
 
 # our C++ atlbase.h conflicts with the one from wine, so only use wine includes for C
 # Unfortunately, we can't use different includes for C & C++ in VS generator, so use an object library to achieve this
-target_include_directories(shlwapi BEFORE PRIVATE ${REACTOS_SOURCE_DIR}/sdk/include/reactos/wine)
+target_include_directories(shlwapi BEFORE PRIVATE
+    ${REACTOS_SOURCE_DIR}/sdk/include/reactos/wine
+    ${REACTOS_SOURCE_DIR}/sdk/lib/atl)
 
 add_library(shlwapi_autocomp OBJECT autocomp.cpp)
 target_link_libraries(shlwapi_autocomp PRIVATE atl_classes)
--- a/dll/win32/shlwapi/propbag.cpp
+++ b/dll/win32/shlwapi/propbag.cpp
@@ -1,0 +1,253 @@
+/*
+ * PROJECT:     ReactOS Shell
+ * LICENSE:     LGPL-2.0-or-later (https://spdx.org/licenses/LGPL-2.0-or-later)
+ * PURPOSE:     Implement shell property bags
+ * COPYRIGHT:   Copyright 2023 Katayama Hirofumi MZ <katayama.hirofumi.mz@gmail.com>
+ */
+
+#define _ATL_NO_EXCEPTIONS
+#include "precomp.h"
+#include <shlwapi.h>
+#include <shlwapi_undoc.h>
+#include <atlstr.h>         // for CStringW
+#include <atlsimpcoll.h>    // for CSimpleMap
+#include <atlcomcli.h>      // for CComVariant
+
+WINE_DEFAULT_DEBUG_CHANNEL(shell);
+
+class CBasePropertyBag
+    : public IPropertyBag
+#if (_WIN32_WINNT < _WIN32_WINNT_VISTA)
+    , public IPropertyBag2
+#endif
+{
+protected:
+    LONG m_cRefs;   // reference count
+    DWORD m_dwMode; // STGM_* flags
+
+public:
+    CBasePropertyBag(DWORD dwMode)
+        : m_cRefs(1)
+        , m_dwMode(dwMode)
+    {
+    }
+
+    virtual ~CBasePropertyBag() { }
+
+    // IUnknown interface
+    STDMETHODIMP QueryInterface(REFIID riid, void **ppvObject) override
+    {
+#if (_WIN32_WINNT < _WIN32_WINNT_VISTA)
+        if (::IsEqualGUID(riid, IID_IPropertyBag2))
+        {
+            AddRef();
+            *ppvObject = dynamic_cast<IPropertyBag2*>(this);
+            return S_OK;
+        }
+#endif
+        if (::IsEqualGUID(riid, IID_IUnknown) || ::IsEqualGUID(riid, IID_IPropertyBag))
+        {
+            AddRef();
+            *ppvObject = dynamic_cast<IPropertyBag*>(this);
+            return S_OK;
+        }
+        return E_NOTIMPL;
+    }
+    STDMETHODIMP_(ULONG) AddRef() override
+    {
+        return ::InterlockedIncrement(&m_cRefs);
+    }
+    STDMETHODIMP_(ULONG) Release() override
+    {
+        if (::InterlockedDecrement(&m_cRefs) == 0)
+        {
+            delete this;
+            return 0;
+        }
+        return m_cRefs;
+    }
+
+#if (_WIN32_WINNT < _WIN32_WINNT_VISTA)
+    // IPropertyBag2 interface (stubs)
+    STDMETHODIMP Read(
+        _In_ ULONG cProperties,
+        _In_ PROPBAG2 *pPropBag,
+        _In_opt_ IErrorLog *pErrorLog,
+        _Out_ VARIANT *pvarValue,
+        _Out_ HRESULT *phrError) override
+    {
+        return E_NOTIMPL;
+    }
+    STDMETHODIMP Write(
+        _In_ ULONG cProperties,
+        _In_ PROPBAG2 *pPropBag,
+        _In_ VARIANT *pvarValue)
+    {
+        return E_NOTIMPL;
+    }
+    STDMETHODIMP CountProperties(_Out_ ULONG *pcProperties) override
+    {
+        return E_NOTIMPL;
+    }
+    STDMETHODIMP GetPropertyInfo(
+        _In_ ULONG iProperty,
+        _In_ ULONG cProperties,
+        _Out_ PROPBAG2 *pPropBag,
+        _Out_ ULONG *pcProperties) override
+    {
+        return E_NOTIMPL;
+    }
+    STDMETHODIMP LoadObject(
+        _In_z_ LPCWSTR pstrName,
+        _In_ DWORD dwHint,
+        _In_ IUnknown *pUnkObject,
+        _In_opt_ IErrorLog *pErrorLog) override
+    {
+        return E_NOTIMPL;
+    }
+#endif
+};
+
+struct CPropMapEqual
+{
+    static bool IsEqualKey(const ATL::CStringW& k1, const ATL::CStringW& k2)
+    {
+        return k1.CompareNoCase(k2) == 0;
+    }
+
+    static bool IsEqualValue(const ATL::CComVariant& v1, const ATL::CComVariant& v2)
+    {
+        return false;
+    }
+};
+
+class CMemPropertyBag : public CBasePropertyBag
+{
+protected:
+    ATL::CSimpleMap<ATL::CStringW, ATL::CComVariant, CPropMapEqual> m_PropMap;
+
+public:
+    CMemPropertyBag(DWORD dwFlags) : CBasePropertyBag(dwFlags) { }
+
+    STDMETHODIMP Read(_In_z_ LPCWSTR pszPropName, _Inout_ VARIANT *pvari,
+                      _Inout_opt_ IErrorLog *pErrorLog) override;
+    STDMETHODIMP Write(_In_z_ LPCWSTR pszPropName, _In_ VARIANT *pvari) override;
+};
+
+STDMETHODIMP
+CMemPropertyBag::Read(
+    _In_z_ LPCWSTR pszPropName,
+    _Inout_ VARIANT *pvari,
+    _Inout_opt_ IErrorLog *pErrorLog)
+{
+    UNREFERENCED_PARAMETER(pErrorLog);
+
+    TRACE("%p: %s %p %p\n", this, debugstr_w(pszPropName), pvari, pErrorLog);
+
+    VARTYPE vt = V_VT(pvari);
+
+    ::VariantInit(pvari);
+
+#if (_WIN32_WINNT < _WIN32_WINNT_VISTA)
+    if ((m_dwMode & (STGM_READ | STGM_WRITE | STGM_READWRITE)) == STGM_WRITE)
+    {
+        ERR("%p: 0x%X\n", this, m_dwMode);
+        return E_ACCESSDENIED;
+    }
+#endif
+
+    if (!pszPropName || !pvari)
+    {
+        ERR("%p: %s %p %p\n", this, debugstr_w(pszPropName), pvari, pErrorLog);
+        return E_INVALIDARG;
+    }
+
+    INT iItem = m_PropMap.FindKey(pszPropName);
+    if (iItem == -1)
+    {
+        ERR("%p: %s %p %p\n", this, debugstr_w(pszPropName), pvari, pErrorLog);
+        return E_FAIL;
+    }
+
+    HRESULT hr = ::VariantCopy(pvari, &m_PropMap.GetValueAt(iItem));
+    if (FAILED(hr))
+    {
+        ERR("%p: 0x%08X %p\n", this, hr, pvari);
+        return hr;
+    }
+
+    hr = ::VariantChangeTypeForRead(pvari, vt);
+    if (FAILED(hr))
+    {
+        ERR("%p: 0x%08X %p\n", this, hr, pvari);
+        return hr;
+    }
+
+    return hr;
+}
+
+STDMETHODIMP
+CMemPropertyBag::Write(
+    _In_z_ LPCWSTR pszPropName,
+    _In_ VARIANT *pvari)
+{
+    TRACE("%p: %s %p\n", this, debugstr_w(pszPropName), pvari);
+
+#if (_WIN32_WINNT < _WIN32_WINNT_VISTA)
+    if ((m_dwMode & (STGM_READ | STGM_WRITE | STGM_READWRITE)) == STGM_READ)
+    {
+        ERR("%p: 0x%X\n", this, m_dwMode);
+        return E_ACCESSDENIED;
+    }
+#endif
+
+    if (!pszPropName || !pvari)
+    {
+        ERR("%p: %s %p\n", this, debugstr_w(pszPropName), pvari);
+        return E_INVALIDARG;
+    }
+
+    ATL::CComVariant vari;
+    HRESULT hr = vari.Copy(pvari);
+    if (FAILED(hr))
+    {
+        ERR("%p: %s %p: 0x%08X\n", this, debugstr_w(pszPropName), pvari, hr);
+        return hr;
+    }
+
+    if (!m_PropMap.SetAt(pszPropName, vari))
+    {
+        ERR("%p: %s %p\n", this, debugstr_w(pszPropName), pvari);
+        return E_FAIL;
+    }
+
+    return hr;
+}
+
+/**************************************************************************
+ *  SHCreatePropertyBagOnMemory (SHLWAPI.477)
+ *
+ * Creates a property bag object on memory.
+ *
+ * @param dwMode  Specifies either STGM_READ, STGM_WRITE or STGM_READWRITE. Ignored on Vista+.
+ * @param riid    Specifies either IID_IUnknown, IID_IPropertyBag or IID_IPropertyBag2.
+ *                Vista+ rejects IID_IPropertyBag2.
+ * @param ppvObj  Receives an IPropertyBag pointer.
+ * @return        An HRESULT value. S_OK on success, non-zero on failure.
+ */
+EXTERN_C HRESULT WINAPI
+SHCreatePropertyBagOnMemory(_In_ DWORD dwMode, _In_ REFIID riid, _Out_ void **ppvObj)
+{
+    TRACE("0x%08X, %s, %p\n", dwMode, debugstr_guid(&riid), ppvObj);
+
+    *ppvObj = NULL;
+
+    CMemPropertyBag* pMemBag = new CMemPropertyBag(dwMode);
+
+    HRESULT hr = pMemBag->QueryInterface(riid, ppvObj);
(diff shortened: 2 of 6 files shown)


dll/win32/shlwapi/CMakeLists.txt
```diff
@@ -50,6 +51,6 @@ add_dependencies(shlwapi_autocomp psdk)
 set_module_type(shlwapi win32dll UNICODE)
 target_link_libraries(shlwapi uuid wine cpprt)
 add_delay_importlibs(shlwapi userenv oleaut32 ole32 comdlg32 mpr mlang urlmon shell32 winmm version)
-add_importlibs(shlwapi user32 gdi32 advapi32 wininet msvcrt kernel32 ntdll)
+add_importlibs(shlwapi comctl32 user32 gdi32 advapi32 wininet msvcrt kernel32 ntdll)
```
katahiromz, line 54: NOTE: Delay loading on ordinal function won't work correctly.

HBelusca: Won't work correctly where? In MSVC builds? In GCC builds? On both?

dll/win32/shlwapi/ordinal.c
```diff
@@ -5729,7 +5729,7 @@ HRESULT WINAPI SHPropertyBag_ReadGUID(IPropertyBag *ppb, LPCWSTR pszPropName, GU
     else if (V_VT(&vari) == VT_BSTR)
         bRet = GUIDFromStringW(V_BSTR(&vari), pguid);
     else
-#if (_WIN32_WINNT >= _WIN32_WINNT_VISTA)
+#if (WINVER >= _WIN32_WINNT_VISTA)
```
HBelusca, line 5732: Huh why? (this is incorrect).

sdk/include/reactos/shlwapi_undoc.h
```diff
@@ -122,6 +122,8 @@ HRESULT WINAPI SHPropertyBag_WritePOINTL(IPropertyBag *ppb, LPCWSTR pszPropName,
 HRESULT WINAPI SHPropertyBag_WritePOINTS(IPropertyBag *ppb, LPCWSTR pszPropName, const POINTS *ppts);
 HRESULT WINAPI SHPropertyBag_WriteRECTL(IPropertyBag *ppb, LPCWSTR pszPropName, const RECTL *prcl);
 
+HRESULT WINAPI SHCreatePropertyBagOnMemory(DWORD dwMode, REFIID riid, IPropertyBag **pppb);
```
HBelusca, line 125: Maybe adding some documentation link like https://www.geoffchappell.com/studies/windows/shell/shlwapi/api/propbag/getviewstate.htm
or something (also it seems it's exported in vista+, in which case having a _WIN32_VERSION guard would be needed)

dll/win32/shlwapi/CMakeLists.txt
katahiromz: GCC build.

dll/win32/shlwapi/propbag.cpp
```diff
@@ -0,0 +1,318 @@
...
+        }
+        return m_cRefs;
+    }
+
+    // IPropertyBag interface (pure)
+    STDMETHODIMP Read(LPCWSTR pszPropName, VARIANT *pvari, IErrorLog *pErrorLog) override = 0;
+    STDMETHODIMP Write(LPCWSTR pszPropName, VARIANT *pvari) override = 0;
+
+    // IPropertyBag2 interface (stubs)
+    STDMETHODIMP Read(
```
HBelusca, line 83: Since this is new ROS code, please add SAL2 annotations.

dll/win32/shlwapi/ordinal.c
katahiromz: It seems `_WIN32_WINNT` definition was incorrect.

HBelusca: "Incorrect": well, it is redefined, for a reason : https://git.reactos.org/?p=reactos.git;a=blob;f=dll/win32/shlwapi/CMakeLists.txt;h=af8ec781824f46c03d8c336f1a59b74cab4818b7;hb=HEAD#l3

dll/win32/shlwapi/CMakeLists.txt
learn-more: not related to ordinal, but the way the symbol is imported.
https://jira.reactos.org/browse/CORE-6504

dll/win32/shlwapi/propbag.cpp
```diff
@@ -0,0 +1,318 @@
...
+    static void operator delete(void* ptr)
+    {
+        ::LocalFree(ptr);
+    }
+};
+
+class CMemPropertyBag : public CBasePropertyBag
+{
+protected:
+    HDSA m_hDSA; // a DSA of PBAGENTRY's
```
learn-more, line 134: Why use a DSA, this is c++ code so you can just use one of the high level collections?

dll/win32/shlwapi/propbag.cpp
```diff
@@ -0,0 +1,318 @@
...
+    STDMETHODIMP LoadObject(
+        LPCWSTR pstrName,
+        DWORD dwHint,
+        IUnknown *pUnkObject,
+        IErrorLog *pErrorLog) override
+    {
+        return E_NOTIMPL;
+    }
+
+    static void* operator new(size_t size)
```
learn-more, line 120: Is there a reason that you need to overload this?

katahiromz: We cannot use both `<new>` and `nothrow` yet.

dll/win32/shlwapi/propbag.cpp
katahiromz: Windows uses DSA at here. I think that is optimal.

dll/win32/shlwapi/propbag.cpp
katahiromz: Added.

sdk/include/reactos/shlwapi_undoc.h
katahiromz: `SHCreatePropertyBagOnMemory` is also exported in Win2k3 as ordinal function.

katahiromz: Added comments.

dll/win32/shlwapi/propbag.cpp
learn-more: I don't see the reason to add a bunch of extra code unless you want to use this from C code, or you need to stay compatible with the memory layout of ms' code (which should not be needed in the internal implementation of something like this, unless this is shared memory that other processes can read or something like it)

dll/win32/shlwapi/propbag.cpp
```diff
@@ -0,0 +1,336 @@
...
+ */
+
+#include "precomp.h"
+#include <shlwapi.h>
+#include <shlwapi_undoc.h>
+
+WINE_DEFAULT_DEBUG_CHANNEL(shell);
+
+// Dynamic structure array (DSA); See COMCTL32
+struct _DSA
```
learn-more, line 15: This struct does not seem to be used here?

dll/win32/shlwapi/propbag.cpp
```diff
@@ -0,0 +1,336 @@
...
+        if (::InterlockedDecrement(&m_cRefs) == 0)
+        {
+            delete this;
+            return 0;
+        }
+        return m_cRefs;
+    }
+
+    // IPropertyBag interface (pure)
+    STDMETHODIMP Read(_In_z_ LPCWSTR pszPropName, _Inout_ VARIANT *pvari,
```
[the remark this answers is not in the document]
katahiromz: Deleted.

dll/win32/shlwapi/propbag.cpp
katahiromz: `HDSA` is a pointer to `struct _DSA`. This is the definition.

sdk/include/psdk/winreg.h
```diff
@@ -197,7 +197,7 @@ RegEnumValueW(
   _Out_writes_bytes_to_opt_(*lpcbData, *lpcbData) __out_data_source(REGISTRY) LPBYTE lpData,
   _Inout_opt_ LPDWORD lpcbData);
 
-#if (_WIN32_WINNT >= 0x0600)
+#if (_WIN32_WINNT >= 0x0600) || defined(__REACTOS__)
```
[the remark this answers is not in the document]
learn-more: it seems there are a few already...

dll/win32/shlwapi/propbag.cpp
katahiromz: Used `CStringW`, `CSimpleMap`, and `CComVariant`.

dll/win32/shlwapi/propbag.cpp
katahiromz: Deleted.

sdk/include/psdk/winreg.h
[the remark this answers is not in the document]
ThFabba: Agreed, though the point is to let arbitrary apps build with our SDK -- so this seems fairly harmless.
It does break the opposite direction though, i.e. now ROS can't build with MS SDK. So I think the usual approach is to just define a newer _WIN32_WINNT for the file(s) where you need this (which we already do for lots of Wine code).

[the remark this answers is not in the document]
HBelusca: > So I think the usual approach is to just define a newer _WIN32_WINNT for the file(s)

That's precisely why we redefined _WIN32_WINNT to Vista+ value in the shlwapi cmakefiles.txt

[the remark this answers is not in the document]
JoachimHenze: "redefined _WIN32_WINNT to Vista+ value in the shlwapi cmakefiles.txt" is definitely not optimal, but what was implemented in this PR instead currently, is even worse. So I do also vote with sticking with the "usual way" here:
* redefining _WIN32_WINNT in cmakefiles.txt
* using _WIN32_WINNT where the code requires it to be used
* **not touching the PSDK with that line** (as that would make the cancer spread out even more

dll/win32/shlwapi/propbag.cpp
katahiromz: Deleted.

sdk/include/psdk/winreg.h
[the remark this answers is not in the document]
katahiromz: Reverted.

dll/win32/shlwapi/CMakeLists.txt
katahiromz: `comctl32` link is deletedw

dll/win32/shlwapi/ordinal.c
katahiromz: Any solution?

JoachimHenze: Can you identify all the parts that do require it to be redefined? (The compiler should help when looking at the build issues when **locally** undefining it). Is it okay to split that question from your current PR?

dll/win32/shlwapi/propbag.cpp
JoachimHenze: Now we do have a lot of additional dependencies. I felt Katayamas initial approach to be superior.

dll/win32/shlwapi/ordinal.c
katahiromz: At least, `RegGetValueW` prototype is missing.

katahiromz: Adding `RegGetValueW` prototype to `assoc.c` does fix this issue.

dll/win32/shlwapi/CMakeLists.txt
```diff
@@ -1,9 +1,6 @@
 
 project(SHELL)
 
-remove_definitions(-D_WIN32_WINNT=0x502)
-add_definitions(-D_WIN32_WINNT=0x600)
-
```
JoachimHenze, line 6: I do like your current approach of encapsulating that existing hack to an even more isolated/distinct part of shlwapi (assoc.c).
Can we split off that part from your PR5511 "SHCreatePropertyBagOnMemory" and commit that on its own?
I think it is a decent improvement in itself already.

sdk/include/psdk/winreg.h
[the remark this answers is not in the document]
katahiromz: Fixed `_WIN32_WINNT`.

dll/win32/shlwapi/ordinal.c
katahiromz: https://github.com/reactos/reactos/pull/5514

sdk/include/psdk/winreg.h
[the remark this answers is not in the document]
katahiromz: https://github.com/reactos/reactos/pull/5514

dll/win32/shlwapi/CMakeLists.txt
katahiromz: https://github.com/reactos/reactos/pull/5514

katahiromz:
Please wait for #5514 patiantly.

dll/win32/shlwapi/propbag.cpp
learn-more: The horror, we have less code and it is a lot simpler now....

sdk/include/reactos/shlwapi_undoc.h
katahiromz: Documentation link is not found.

HBelusca: Oh yes correct, it was one of the others.
This one should work: http://undoc.airesoft.co.uk/shlwapi.dll/SHCreatePropertyBagOnMemory.php
In addition, https://www.geoffchappell.com/studies/windows/shell/shlwapi/api/propbag/index.htm has links for the other functions.

katahiromz: http://undoc.airesoft.co.uk/shlwapi.dll/SHCreatePropertyBagOnMemory.php is added.

katahiromz: Should I add `SHPropertyBag_...` links?

HBelusca: If you personally find it useful, why not (I'm not forcing anything here as I don't have any preference). But if you want to add the links for all the other SHPropertyBag_ functions, just use the probag/index.htm one, no need I think to add each of the sublinks for each function.

dll/win32/shlwapi/propbag.cpp
```diff
@@ -0,0 +1,254 @@
...
+#include <atlstr.h>         // for CStringW
+#include <atlsimpcoll.h>    // for CSimpleMap
+#include <atlcomcli.h>      // for CComVariant
+
+WINE_DEFAULT_DEBUG_CHANNEL(shell);
+
+class CBasePropertyBag
+    : public IPropertyBag
+#if (_WIN32_WINNT < _WIN32_WINNT_VISTA)
+    , public IPropertyBag2
```
HBelusca, line 21: Oh, that's interesting. So (for once!) they merged both interfaces into a single one in Vista+ ? or...?

sdk/include/reactos/shlwapi_undoc.h
katahiromz: https://github.com/reactos/reactos/pull/5515

katahiromz: Done.

dll/win32/shlwapi/propbag.cpp
katahiromz: @HBelusca Any idea?

dll/win32/shlwapi/propbag.cpp
katahiromz: I didn't test it on Win7 nor Win8.
Vista is my expectation.

dll/win32/shlwapi/propbag.cpp
HBelusca: If the internals (and layout) of the underlying class that implements the "on-memory" property bag doesn't need to be known by applications using the SHCreatePropertyBagOnMemory API and the associated helpers, then using the ATL may be better here.

dll/win32/shlwapi/propbag.cpp
HBelusca: You should try checking on Win7 and 10 out of curiosity / documentation.

katahiromz: Win10 is tested on my PC.

modules/rostests/apitests/shlwapi/SHPropertyBag.cpp
```diff
@@ -325,8 +325,140 @@ static void SHPropertyBag_WriteTest(void)
...
+        ok_long(hr, E_ACCESSDENIED);
+        ok_int(V_VT(&vari), VT_EMPTY);
+        ok_long(V_UI4(&vari), 0xFEEDF00D);
+    }
+    VariantClear(&vari);
+
+    pPropBag->Release();
+
+    hr = SHCreatePropertyBagOnMemory(STGM_READWRITE, IID_IPropertyBag2, (void**)&pPropBag);
+    if (IsWindowsVistaOrGreater())
```
HBelusca, line 426: Can you make a temporary commit where this check is removed, so as to see whether this one fails on Windows 7 or 10? (same for the two other IsWindowsVistaOrGreater checks below).

modules/rostests/apitests/shlwapi/SHPropertyBag.cpp
```diff
@@ -325,8 +325,140 @@ static void SHPropertyBag_WriteTest(void)
...
+    ok_long(hr, S_OK);
+    VariantClear(&vari);
+
+    VariantInit(&vari);
+    V_VT(&vari) = VT_UI4;
+    V_UI4(&vari) = 0xFEEDF00D;
+    hr = pPropBag->Read(L"Name1", &vari, NULL);
+    if (IsWindowsVistaOrGreater())
+    {
+        ok_long(hr, S_OK);
```
HBelusca, line 411: Is this behaviour change also happening on Windows 7? (If you make a temporary commit for those, in addition to the one I suggested previously, I could test that on Win7)

modules/rostests/apitests/shlwapi/SHPropertyBag.cpp
katahiromz: Win10 is already tested.
Win7 is greater than WinVista.
No need to test so.

dll/win32/shlwapi/propbag.cpp
```diff
@@ -0,0 +1,254 @@
...
+        _In_opt_ IErrorLog *pErrorLog,
+        _Out_ VARIANT *pvarValue,
+        _Out_ HRESULT *phrError) override
+    {
+        return E_NOTIMPL;
+    }
+    STDMETHODIMP Write(
+        _In_ ULONG cProperties,
+        _In_ PROPBAG2 *pPropBag,
+        _In_ VARIANT *pvarValue)
```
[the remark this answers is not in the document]
katahiromz: Yep.

modules/rostests/apitests/shlwapi/SHPropertyBag.cpp
katahiromz: Win7 is greater than WinVista.
So, it is also happening.

HBelusca (approved):

dll/win32/shlwapi/propbag.cpp
```diff
@@ -0,0 +1,254 @@
...
+    virtual ~CBasePropertyBag() { }
+
+    // IUnknown interface
+    STDMETHODIMP QueryInterface(REFIID riid, void **ppvObject) override
+    {
+#if (_WIN32_WINNT < _WIN32_WINNT_VISTA)
+        if (::IsEqualGUID(riid, IID_IPropertyBag2))
+        {
+            AddRef();
+            *ppvObject = dynamic_cast<IPropertyBag2*>(this);
```
learn-more, line 44: I'd prefer a `static_cast` here, because `dynamic_cast` depends on runtime type info, which can be 'interesting' in certain builds.

dll/win32/shlwapi/propbag.cpp
```diff
@@ -0,0 +1,254 @@
...
+ * @see           http://undoc.airesoft.co.uk/shlwapi.dll/SHCreatePropertyBagOnMemory.php
+ */
+EXTERN_C HRESULT WINAPI
+SHCreatePropertyBagOnMemory(_In_ DWORD dwMode, _In_ REFIID riid, _Out_ void **ppvObj)
+{
+    TRACE("0x%08X, %s, %p\n", dwMode, debugstr_guid(&riid), ppvObj);
+
+    *ppvObj = NULL;
+
+    CMemPropertyBag* pMemBag = new CMemPropertyBag(dwMode);
```
learn-more, line 246: Dropping this directly into a `CComPtr` here will simplify cleanup because you don't have to call 'Release' anymore.

learn-more (approved):
some small points, lgtm otherwise

dll/win32/shlwapi/propbag.cpp
katahiromz: Used `static_cast`.

dll/win32/shlwapi/propbag.cpp
katahiromz: Used `CComPtr`. Thanks.

dll/win32/shlwapi/propbag.cpp
```diff
@@ -243,12 +243,12 @@ SHCreatePropertyBagOnMemory(_In_ DWORD dwMode, _In_ REFIID riid, _Out_ void **pp
 
     *ppvObj = NULL;
 
-    CMemPropertyBag* pMemBag = new CMemPropertyBag(dwMode);
+    CComPtr<CMemPropertyBag> pMemBag;
+    pMemBag.Attach(new CMemPropertyBag(dwMode));
```
HBelusca, line 247:
```suggestion
    CComPtr<CMemPropertyBag> pMemBag(new CMemPropertyBag(dwMode));
```

katahiromz: Needed a fix in `m_cRefs` initialization.

HBelusca (approved):

dll/win32/shlwapi/propbag.cpp
HBelusca: If the Attach method had been used, there wouldn't have been the need for fixing the m_cRefs? interesting...

katahiromz: `CComPtr` ctor is adding the reference count...
```c
    CComPtr(T *lp)
    {
        p = lp;
        if (p != NULL)
            p->AddRef();
    }
```

PR #5517, opened 2023-08-04
[SHELL32][SHELL32_APITEST][SDK] Implement SheRemoveQuotesA/W

katahiromz:
## Purpose

## Proposed changes

- Implement `SheRemoveQuotesA` and `SheRemoveQuotesW` functions.
- Add `She` testcase into `shell32_apitest.exe`.
- Add `SheRemoveQuotesA` and `SheRemoveQuotesW` prototypes into `undocshell.h`

## TODO

- [x] Do small tests.

## Comparison

WinXP:
WinXP
Successful.

Win2k3:
Win2k3
Successful.

Win10:
Win10
Skipped because there are no `SheRemoveQuotesA/W` functions.

ReactOS (AFTER):
after
Successful.

### DIFF
--- a/dll/win32/shell32/stubs.cpp
+++ b/dll/win32/shell32/stubs.cpp
@@ -460,7 +460,10 @@ SheRemoveQuotesW(LPWSTR psz)
     if (*psz == L'"')
     {
         for (pch = psz + 1; *pch && *pch != L'"'; ++pch)
+        {
             *(pch - 1) = *pch;
+        }
+
         if (*pch == L'"')
             *(pch - 1) = UNICODE_NULL;
     }
@@ -477,7 +480,10 @@ SheRemoveQuotesA(LPSTR psz)
     if (*psz == '"')
     {
         for (pch = psz + 1; *pch && *pch != '"'; ++pch)
+        {
             *(pch - 1) = *pch;
+        }
+
         if (*pch == '"')
             *(pch - 1) = ANSI_NULL;
     }
--- a/dll/win32/shell32/stubs.cpp
+++ b/dll/win32/shell32/stubs.cpp
@@ -451,26 +451,38 @@ SheSetCurDrive(INT iIndex)
     return 1;
 }
 
-/*
- * Unimplemented
- */
 EXTERN_C LPWSTR
 WINAPI
-SheRemoveQuotesW(LPWSTR lpInput)
+SheRemoveQuotesW(LPWSTR psz)
 {
-    FIXME("SheRemoveQuotesW() stub\n");
-    return NULL;
+    PWCHAR pch;
+
+    if (*psz == L'"')
+    {
+        for (pch = psz + 1; *pch && *pch != L'"'; ++pch)
+            *(pch - 1) = *pch;
+        if (*pch == L'"')
+            *(pch - 1) = UNICODE_NULL;
+    }
+
+    return psz;
 }
 
-/*
- * Unimplemented
- */
 EXTERN_C LPSTR
 WINAPI
-SheRemoveQuotesA(LPSTR lpInput)
+SheRemoveQuotesA(LPSTR psz)
 {
-    FIXME("SheRemoveQuotesA() stub\n");
-    return NULL;
+    PCHAR pch;
+
+    if (*psz == '"')
+    {
+        for (pch = psz + 1; *pch && *pch != '"'; ++pch)
+            *(pch - 1) = *pch;
+        if (*pch == '"')
+            *(pch - 1) = ANSI_NULL;
+    }
+
+    return psz;
 }
 
 /*
--- a/modules/rostests/apitests/shell32/She.cpp
+++ b/modules/rostests/apitests/shell32/She.cpp
@@ -6,7 +6,6 @@
  */
 
 #include "shelltest.h"
-#include <versionhelpers.h>
 #include <undocshell.h>
 
 typedef LPSTR (WINAPI *FN_SheRemoveQuotesA)(LPSTR psz);
--- a/modules/rostests/apitests/shell32/She.cpp
+++ b/modules/rostests/apitests/shell32/She.cpp
@@ -20,6 +20,18 @@ static void test_SheRemoveQuotesA(void)
     CHAR sz1[] = "\"Test\"";
     CHAR sz2[] = "\"Test\"123";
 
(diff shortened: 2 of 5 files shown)


JoachimHenze (approved):
Approved. Can confirm those funcx are exported even on XPSP3 already.
Implementation does look straigthforward to me.
Question: Where do you intend to use them Katayama?

katahiromz:
> Question: Where do you intend to use them Katayama?

This function is used in `shell32!ExtractAssociatedIconExW`.

dll/win32/shell32/stubs.cpp
```diff
@@ -451,26 +451,38 @@ SheSetCurDrive(INT iIndex)
...
 EXTERN_C LPWSTR
 WINAPI
-SheRemoveQuotesW(LPWSTR lpInput)
+SheRemoveQuotesW(LPWSTR psz)
 {
-    FIXME("SheRemoveQuotesW() stub\n");
-    return NULL;
+    PWCHAR pch;
+
+    if (*psz == L'"')
```
JoachimHenze, line 460:
```suggestion
    if (psz && *psz == L'"')
```
Wouldn't your initial implemenation crash when being called with a nullptr? I would recommend to add a test with NULL being passed also.

dll/win32/shell32/stubs.cpp
```diff
@@ -451,26 +451,38 @@ SheSetCurDrive(INT iIndex)
...
 EXTERN_C LPSTR
 WINAPI
-SheRemoveQuotesA(LPSTR lpInput)
+SheRemoveQuotesA(LPSTR psz)
 {
-    FIXME("SheRemoveQuotesA() stub\n");
-    return NULL;
+    PCHAR pch;
+
+    if (*psz == '"')
```
JoachimHenze, line 477:
```suggestion
    if (psz && *psz == '"')
```

Wouldn't your initial implemenation crash when being called with a nullptr? I would recommend to add a test with NULL being passed also.

JoachimHenze (changes requested):

dll/win32/shell32/stubs.cpp
katahiromz: The original function doesn't check it.

dll/win32/shell32/stubs.cpp
katahiromz: The original function doesn't check it.

dll/win32/shell32/stubs.cpp
JoachimHenze: So you say if you pass a nullptr to that func in Windows it will crash?
Does that still apply to Windows 10?

katahiromz: Added `NULL` tests.
Confirmed in Win2k3 and WinXP.
Win10 skips the tests.

dll/win32/shell32/stubs.cpp
katahiromz: Added `NULL` tests.
Confirmed in Win2k3 and WinXP.
Win10 skips the tests.

modules/rostests/apitests/shell32/She.cpp
```diff
@@ -0,0 +1,87 @@
...
+    CHAR sz2[] = "\"Test\"123";
+
+    BOOL bGotException = FALSE;
+    _SEH2_TRY
+    {
+        pSheRemoveQuotesA(NULL);
+    }
+    _SEH2_EXCEPT(EXCEPTION_EXECUTE_HANDLER)
+    {
+        bGotException = TRUE;
```
katahiromz, line 30: Here is the `NULL` test.

dll/win32/shell32/stubs.cpp
JoachimHenze: I don't think we should follow Microsofts behavior in crashing to such an extent. I do really vote for checking for NULL here. I don't think this would break any compatibility. I don't think there is any application out there that requires to crash "to behave correctly".

modules/rostests/apitests/shell32/She.cpp
```diff
@@ -0,0 +1,87 @@
+/*
+ * PROJECT:     ReactOS api tests
+ * LICENSE:     GPL-2.0-or-later (https://spdx.org/licenses/GPL-2.0-or-later)
+ * PURPOSE:     Test for She* functions
```
HBelusca, line 4: Is it going to serve for other ones (like SheSetCurDrive etc.), or only for the RemoveQuote functions?

dll/win32/shell32/stubs.cpp
```diff
@@ -451,26 +451,38 @@ SheSetCurDrive(INT iIndex)
...
+    {
+        for (pch = psz + 1; *pch && *pch != '"'; ++pch)
+            *(pch - 1) = *pch;
```
HBelusca, line 480:
```suggestion
        {
            *(pch - 1) = *pch;
        }
```

modules/rostests/apitests/shell32/She.cpp
katahiromz: It's to serve for other ones.

dll/win32/shell32/stubs.cpp
katahiromz: It should be "correctly crashing". That is the expected behavior.

Why does every PC crash on system failure? To stop the wrong process. To find the fatal mistake(s) on the system.

dll/win32/shell32/stubs.cpp
```diff
@@ -451,26 +451,40 @@ SheSetCurDrive(INT iIndex)
...
+SheRemoveQuotesW(LPWSTR psz)
 {
-    FIXME("SheRemoveQuotesW() stub\n");
-    return NULL;
+    PWCHAR pch;
+
+    if (*psz == L'"')
+    {
+        for (pch = psz + 1; *pch && *pch != L'"'; ++pch)
+            *(pch - 1) = *pch;
```
HBelusca, line 463: for-loop braces here too (I didn't want to repeat my previous comment)

HBelusca (approved):
lgtm otherwise

dll/win32/shell32/stubs.cpp
katahiromz: Fixed. Sorry.

PR #5519, opened 2023-08-04
[APPLICATIONS][DEVMGR] Do not free shared icon for ShellAbout() calls

SergeGautherie:
## Purpose

Do not free shared icons, ReactOS ShellAbout() calls part.

JIRA issue: [CORE-18369](https://jira.reactos.org/browse/CORE-18369)

## Proposed changes

- Remove unwanted `DeleteObject()` and `DestroyIcon(()` calls.

## TODO

Wine ieframe case will be handled separately/later.

### DIFF
--- a/base/applications/calc/winmain.c
+++ b/base/applications/calc/winmain.c
@@ -1945,9 +1945,9 @@ int WINAPI _tWinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPTSTR lpCmdL
                     hInstance,
                     MAKEINTRESOURCE(IDI_CALC),
                     IMAGE_ICON,
-                    GetSystemMetrics(SM_CXICON),
-                    GetSystemMetrics(SM_CYICON),
-                    0);
+                    0,
+                    0,
+                    LR_DEFAULTSIZE | LR_SHARED);
 
     calc.hSmIcon = LoadImage(
                     hInstance,
@@ -1955,7 +1955,7 @@ int WINAPI _tWinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPTSTR lpCmdL
                     IMAGE_ICON,
                     GetSystemMetrics(SM_CXSMICON),
                     GetSystemMetrics(SM_CYSMICON),
-                    0);
+                    LR_SHARED);
 
     do {
         /* ignore hwnd: dialogs are already visible! */
@@ -1984,12 +1984,6 @@ int WINAPI _tWinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPTSTR lpCmdL
 
         save_config();
     } while (calc.action != IDC_STATIC);
-
-    if (calc.hBgIcon != NULL)
-        DestroyIcon(calc.hBgIcon);
-
-    if (calc.hSmIcon != NULL)
-        DestroyIcon(calc.hSmIcon);
 
     stop_rpn_engine();
 
--- a/base/applications/osk/main.c
+++ b/base/applications/osk/main.c
@@ -138,19 +138,14 @@ DWORD WINAPI OSK_WarningDlgThread(LPVOID lpParameter)
 VOID OSK_About(VOID)
 {
     WCHAR szAuthors[MAX_PATH];
-    HICON OSKIcon;
-
-    /* Load the icon */
-    OSKIcon = LoadImageW(Globals.hInstance, MAKEINTRESOURCEW(IDI_OSK), IMAGE_ICON, 0, 0, LR_DEFAULTSIZE);
 
     /* Load the strings into the "About" dialog */
     LoadStringW(Globals.hInstance, IDS_AUTHORS, szAuthors, _countof(szAuthors));
 
+    /* Load the icon */
     /* Finally, execute the "About" dialog by using the Shell routine */
-    ShellAboutW(Globals.hMainWnd, Globals.szTitle, szAuthors, OSKIcon);
-
-    /* Once done, destroy the icon */
-    DestroyIcon(OSKIcon);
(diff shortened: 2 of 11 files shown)


base/applications/osk/main.c
```diff
@@ -138,19 +138,14 @@ DWORD WINAPI OSK_WarningDlgThread(LPVOID lpParameter)
...
     LoadStringW(Globals.hInstance, IDS_AUTHORS, szAuthors, _countof(szAuthors));
 
+    /* Load the icon */
     /* Finally, execute the "About" dialog by using the Shell routine */
-    ShellAboutW(Globals.hMainWnd, Globals.szTitle, szAuthors, OSKIcon);
-
-    /* Once done, destroy the icon */
-    DestroyIcon(OSKIcon);
+    ShellAboutW(Globals.hMainWnd, Globals.szTitle, szAuthors,
+                LoadImageW(Globals.hInstance, MAKEINTRESOURCEW(IDI_OSK), IMAGE_ICON, 0, 0, LR_DEFAULTSIZE | LR_SHARED));
```
katahiromz, line 148: Why not using `LoadIconW`?

SergeGautherie: For minimal change and because MS doc reads
'This function has been superseded by the [LoadImage](https://learn.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-loadimagea) function (with LR_DEFAULTSIZE and LR_SHARED flags set).'.
Other than that, nothing is really wrong with`LoadIconW`.
I'll change it if you prefer.

katahiromz: Will Windows drop `LoadIconW` support? I don't think so.

base/applications/calc/winmain.c
```diff
@@ -1945,17 +1945,17 @@ int WINAPI _tWinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPTSTR lpCmdL
...
+                    0,
+                    0,
+                    LR_DEFAULTSIZE | LR_SHARED);
 
     calc.hSmIcon = LoadImage(
                     hInstance,
                     MAKEINTRESOURCE(IDI_CALC),
                     IMAGE_ICON,
                     GetSystemMetrics(SM_CXSMICON),
                     GetSystemMetrics(SM_CYSMICON),
```
JoachimHenze, line 1957: I tend to approve, just one very last question for my understanding, before I'll do:
Why do we use the GetSystemMetrics() calls for the small icon, but do change those to 0 for the large icon a few lines above?

base/applications/osk/main.c
GeoB99, line 148: I don't get the point of this. `LoadImageW` in the original will load the icon each time a call is invoked to it, thus a uniquely handle is given for it. It is not a shared icon so it is correct to call `DestroyIcon` so I don't get what are you trying to achieve here.

GeoB99 (changes requested):

base/applications/osk/main.c
learn-more: `LR_SHARED` does not give a unique handle every time,
as it implies: it is a shared resource.

base/applications/calc/winmain.c
SergeGautherie: Functionally equivalent: `0 and LR_DEFAULTSIZE == SM_CXICON and SM_CYICON`, like other calls which do not needed non-default sizes.

base/applications/osk/main.c
SergeGautherie: GeoB99, yes, this case existing code is self-consistent. I am simply switching it to the same shared pattern as the other calls.

base/applications/calc/winmain.c
JoachimHenze: Interesting. The version without the GetSystemMetrics()-calls looks cheaper then. Nvm. Approved also in its current state. Don't want to add any risk with something I am not familiar with.

JoachimHenze (approved):

HBelusca (approved):

base/applications/osk/main.c
GeoB99: I'm not a fan of randomly changing code on stuff that actually is fine, the ticket on itself says about that there might be calls that create shared images and then they're freed afterwards, when in OSK this is not actually the case. But fine, you get a pass.

HBelusca:
@SergeGautherie Could you fix your primary mail in your github account to be the one "reactos-git_serge_171003@ ...." ? instead of using the default no-reply one from github.

SergeGautherie:
> Could you fix your primary mail in your github account to be the one "reactos-git_serge_171003@ ...." ?

No, that address is not meant to be my GitHub Primary one.
See https://github.com/reactos/reactos/pull/4881#discussion_r1022091027.

HBelusca:
> > Could you fix your primary mail in your github account to be the one "reactos-git_serge_171003@ ...." ?
> 
> No, that address is not meant to be my GitHub Primary one. See [#4881 (comment)](https://github.com/reactos/reactos/pull/4881#discussion_r1022091027).

The one shown as "Public email" on your profile (and that is going to be used, when us there press the Squash-and-merge button) can be different from your GitHub primary address.

HBelusca:
> "Squash and merge" uses the primary email, not the one specified as "Public email" in profile.

:oof: then what can Serge do there?

SergeGautherie:
> The one shown as "Public email" on your profile [...] can be different from your GitHub primary address.

"No": Public profile email is disabled while private option is enabled. Not relevant anyway...

binarymaster:
> then what can Serge do there?

Enable edits from maintainers, so we could merge manually.

PR #5520, opened 2023-08-04
[SHELL32][SDK] Implement SHIsBadInterfacePtr

katahiromz:
## Purpose

## Proposed changes

- Implement `SHIsBadInterfacePtr` function in `dll/win32/shell32/misc.cpp`.
- Add `SHIsBadInterfacePtr` prototype into `<undocshell.h>`.

### DIFF
--- a/dll/win32/shell32/wine/shellole.c
+++ b/dll/win32/shell32/wine/shellole.c
@@ -810,3 +810,13 @@ HRESULT WINAPI SHCreateQueryCancelAutoPlayMoniker(IMoniker **moniker)
     if (!moniker) return E_INVALIDARG;
     return CreateClassMoniker(&CLSID_QueryCancelAutoPlay, moniker);
 }
+
+/*************************************************************************
+ *  SHIsBadInterfacePtr [SHELL32.84]
+ */
+BOOL WINAPI SHIsBadInterfacePtr(IUnknown *punk, UINT cbVtbl)
+{
+    return (IsBadReadPtr(punk, sizeof(LPVOID)) ||
+            IsBadReadPtr(punk->lpVtbl, cbVtbl) ||
+            IsBadCodePtr((FARPROC)punk->lpVtbl->Release));
+}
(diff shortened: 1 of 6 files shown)


dll/win32/shell32/wine/shellole.c
```diff
@@ -810,3 +810,13 @@ HRESULT WINAPI SHCreateQueryCancelAutoPlayMoniker(IMoniker **moniker)
     if (!moniker) return E_INVALIDARG;
     return CreateClassMoniker(&CLSID_QueryCancelAutoPlay, moniker);
 }
+
+/*************************************************************************
+ *  SHIsBadInterfacePtr [SHELL32.84]
+ */
+BOOL WINAPI SHIsBadInterfacePtr(IUnknown *punk, UINT cbVtbl)
```
HBelusca, line 817: You've put it in a wine-synced file, but this function is not implemented in wine. Please either move it somewhere else, or surround its implementation in `#ifdef __REACTOS__` .

katahiromz: Added `#ifdef __REACTOS__` ... `#endif`.

learn-more:
Do you need this function for anything?
Or did you find some random application depending on it?

katahiromz:
> Do you need this function for anything?
> Or did you find some random application depending on it?

Nope. Nope.

dll/win32/shell32/wine/shellole.c
tkreuzer: Please don't put new functions into wine synced files.

katahiromz: Moved to `misc.cpp`.

JoachimHenze:
`Nope. Nope.`
We had a rule once to only implement what we really needed for something or where we could prove that any 3rd party app depends on. I always found that to be a very good rule. I therefore recommend to not merge this PR 'just because we can'. But to wait instead until such a need would evolve. It has a proper description so we would be able to find it later.

katahiromz:
I found it was needless. Closed.

learn-more:
> `Nope. Nope.` We had a rule once to only implement what we really needed for something or where we could prove that any 3rd party app depends on. I always found that to be a very good rule. I therefore recommend to not merge this PR 'just because we can'. But to wait instead until such a need would evolve. It has a proper description so we would be able to find it later.

Really now, where was that rule defined?

PR #5521, opened 2023-08-04
[SHELL32][SDK] Implement SHLimitInputCombo

katahiromz:
## Purpose

## Overview

`SHLimitInputCombo` function sets limits on valid characters for a combobox control.
This function works like `SHLimitInputEdit`, but the target is a combobox instead of a textbox.

## Proposed changes

- Implement `SHLimitInputCombo` function.

### DIFF
--- a/dll/win32/shell32/stubs.cpp
+++ b/dll/win32/shell32/stubs.cpp
@@ -51,6 +51,10 @@ static BOOL CALLBACK _FindTheEditBox(HWND hwndEdit, LPARAM lParam)
 
 /*************************************************************************
  *  SHLimitInputCombo [SHELL32.748]
+ *
+ * Sets limits on valid characters for a combobox control.
+ * This function works like SHLimitInputEdit, but the target is a combobox
+ * instead of a textbox.
  */
 EXTERN_C BOOL
 WINAPI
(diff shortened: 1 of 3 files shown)


HBelusca (changes requested):
Please explain somewhere (in commit log / in the code) what the purpose of this function is.

dll/win32/shell32/stubs.cpp
```diff
@@ -36,15 +36,33 @@ SHFindComputer(LPCITEMIDLIST pidl1, LPCITEMIDLIST pidl2)
...
  */
 EXTERN_C BOOL
 WINAPI
-SHLimitInputCombo(HWND hWnd, LPVOID lpUnknown)
+SHLimitInputCombo(HWND hWnd, IShellFolder *psf)
 {
-    FIXME("SHLimitInputCombo() stub\n");
-    return FALSE;
+    LIMIT_INPUT_COMBO LimitInput = { E_FAIL, psf };
+    EnumChildWindows(hWnd, _FindTheEditBox, (LPARAM)&LimitInput);
```
learn-more, line 64: This function seems to suggest that there is only ever one child window, so why enumerate it and not just fetch the child window directly?
Something like `::GetWindow(hWnd, GW_CHILD);`

katahiromz: Simplified by using `GetTopWindow`.

learn-more (approved):

PR #5526, opened 2023-08-05
[NTUSER] Fix SPI_SETFONTSMOOTHING behavior

cbialorucki:
## Purpose

Fixes the behavior of the `SPI_SETFONTSMOOTHING` system parameter. This also fixes a bug when trying to select "Use the following method to smooth edges of screen fonts" in `desk.cpl`.

JIRA issue: [CORE-19092](https://jira.reactos.org/browse/CORE-19092)

## Proposed changes
- Instead of checking if `uiParam` is equal to two, treat `uiParam` as a Boolean that sets font smoothing on or off. This is the correct behavior according to [Microsoft documentation](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-systemparametersinfow#SPI_SETFONTSMOOTHING).
- Use `SpiStoreSzInt` method instead of `SpiStoreSz` to store the value in the registry.

## TODO
- Get font smoothing working in more windows

### DIFF
--- a/win32ss/user/ntuser/sysparams.c
+++ b/win32ss/user/ntuser/sysparams.c
@@ -1404,7 +1404,7 @@ SpiGetSet(UINT uiAction, UINT uiParam, PVOID pvParam, FLONG fl)
             return SpiGetInt(pvParam, &gspv.bFontSmoothing, fl);
 
         case SPI_SETFONTSMOOTHING:
-            gspv.bFontSmoothing = uiParam;
+            gspv.bFontSmoothing = !!uiParam;
             if (fl & SPIF_UPDATEINIFILE)
             {
                 SpiStoreSzInt(KEY_DESKTOP, VAL_FONTSMOOTHING, gspv.bFontSmoothing ? 2 : 0);
--- a/win32ss/user/ntuser/sysparams.c
+++ b/win32ss/user/ntuser/sysparams.c
@@ -1404,10 +1404,10 @@ SpiGetSet(UINT uiAction, UINT uiParam, PVOID pvParam, FLONG fl)
             return SpiGetInt(pvParam, &gspv.bFontSmoothing, fl);
 
         case SPI_SETFONTSMOOTHING:
-            gspv.bFontSmoothing = (uiParam == 2);
+            gspv.bFontSmoothing = uiParam;
             if (fl & SPIF_UPDATEINIFILE)
             {
-                SpiStoreSz(KEY_DESKTOP, VAL_FONTSMOOTHING, (uiParam == 2) ? L"2" : L"0");
+                SpiStoreSzInt(KEY_DESKTOP, VAL_FONTSMOOTHING, gspv.bFontSmoothing ? 2 : 0);
             }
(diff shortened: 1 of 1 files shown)


win32ss/user/ntuser/sysparams.c
```diff
@@ -1404,10 +1404,10 @@ SpiGetSet(UINT uiAction, UINT uiParam, PVOID pvParam, FLONG fl)
             return SpiGetInt(pvParam, &gspv.bFontSmoothing, fl);
 
         case SPI_SETFONTSMOOTHING:
-            gspv.bFontSmoothing = (uiParam == 2);
+            gspv.bFontSmoothing = uiParam;
```
binarymaster, line 1407: You can use this to convert it into a BOOL:
```suggestion
            gspv.bFontSmoothing = !!uiParam;
```

binarymaster (commented):
See also line 308 in the same file, it might need to be changed as well.

win32ss/user/ntuser/sysparams.c
```diff
@@ -1404,10 +1404,10 @@ SpiGetSet(UINT uiAction, UINT uiParam, PVOID pvParam, FLONG fl)
...
             if (fl & SPIF_UPDATEINIFILE)
             {
-                SpiStoreSz(KEY_DESKTOP, VAL_FONTSMOOTHING, (uiParam == 2) ? L"2" : L"0");
+                SpiStoreSzInt(KEY_DESKTOP, VAL_FONTSMOOTHING, gspv.bFontSmoothing ? 2 : 0);
```
binarymaster, line 1410: Doesn't the "2" value imply that there should be two algorithms of font smoothing?

HBelusca: Here the "2" seems to be an internal value (stored by the win32k) that is not directly related to the `FE_FONTSMOOTHING***` defined values. (Those latter ones are for the `SPI_GET/SETFONTSMOOTHINGTYPE` messages.)

cbialorucki: Perhaps at some point in Window's development Microsoft considered using the font smoothing registry value to also store the font smoothing type, but after my own investigation I only ever found it to be either 2 or 0. Since SPI_SETFONTSMOOTHING is only supposed to take in a Boolean I felt this was the right way to do it. Feel free to check on your own Windows installs, the registry key is in `HKCU\Control Panel\Desktop\FontSmoothing`.

win32ss/user/ntuser/sysparams.c
cbialorucki: I think I would rather do this if we want to be more explicit, though I don't think its necessary:
`gspv.bFontSmoothing = uiParam ? TRUE : FALSE`
I'll leave this up for debate.

HBelusca (approved):
The `!!uiParam` needs to be applied though.

cbialorucki:
> The `!!uiParam` needs to be applied though.

Applied.

binarymaster (approved):

PR #5527, opened 2023-08-05
[NTOS:CC] Fix locking of CcRosVacbIncRefCount in CcRosCreateVacb

Doug-Lyons:
## Purpose
I apologize for the massive changes here in advance.
Based on Developer comments, I have decided to change this PR to only cover the lock fixes.
Therefore, comments below may seem out-of-place.
Thanks for understanding.
I am leaving the older information for context only.
=================================
MM/CC Improvements by Thomas Faber. Captured by Doug Lyons.

_Add back CcRosTrimCache removed in [0.4.15-dev-1717-gd8cdb89](https://github.com/reactos/reactos/commit/d8cdb89fb03006595dc40ac23db5267b8d9d9c09)._
_Fix locking of CcRosVacbIncRefCount in CcRosCreateVacb._

## Proposed changes

_This PR replaces PR #5237 which needed a rebase and I was impatient to get this done._
_It also captures a new improvement suggested to me by Thomas Faber just today._

Testbot Results:
PR5237-05.patch JID66166 on top of 0.4.15-dev-6415-g21e30f3

VBox: https://reactos.org/testman/compare.php?ids=88675,88680 LGTM
KVM:  https://reactos.org/testman/compare.php?ids=88676,88679 LGTM

I have had very good results running this patch 20 times on Total Commander with no fails from the memory manager.
Maybe others can provide their results here. Thanks.


tkreuzer:
@ThFabba In PR #5237 is says "as there might not be enough pages for the balancer itself (ASSERT(MmAvailablePages >= 32);) later on." - Is this still valid in the sense that the paging path relies on pages being available? In that case we need even more changes, because even if the thread runs at higher priority, it can not prevent other threads from eating up the remaining pages before the paging is complete, especially in an SMP environment.
We would need to either fix the paging path to preallocate any needed pages or reserve pages that will only be available in the paging path. I would prefer the former, because the latter seems like hard to prove to be correct, but it's probably also extremely difficult.

tkreuzer:
@JoachimHenze Please test.

ntoskrnl/cc/view.c
```diff
@@ -449,6 +449,132 @@ CcRosFlushDirtyPages (
...
+                                    VacbLruListEntry);
+        current_entry = current_entry->Flink;
+
+        KeAcquireSpinLockAtDpcLevel(&current->SharedCacheMap->CacheMapLock);
+
+        /* Reference the VACB */
+        CcRosVacbIncRefCount(current);
+
+        /* Check if it's mapped and not dirty */
+        if (InterlockedCompareExchange((PLONG)&current->MappedCount, 0, 0) > 0 && !current->Dirty)
```
tkreuzer, line 498: This looks wrong. This doesn't provide any syncronization, since it doesn't change the value. It's basically a volatile read.

ntoskrnl/cc/view.c
```diff
@@ -449,6 +449,132 @@ CcRosFlushDirtyPages (
...
+
+            RemoveEntryList(&current->CacheMapVacbListEntry);
+            RemoveEntryList(&current->VacbLruListEntry);
+            InitializeListHead(&current->VacbLruListEntry);
+            InsertHeadList(&FreeList, &current->CacheMapVacbListEntry);
+
+            /* Calculate how many pages we freed for Mm */
+            PagesFreed = min(VACB_MAPPING_GRANULARITY / PAGE_SIZE, Target);
+            Target -= PagesFreed;
+            (*NrFreed) += PagesFreed;
```
tkreuzer, line 535: Is there a reason we report the minimum, rather than the actual number?

binarymaster:
> This PR replaces another PR which needed a rebase and I was impatient to get this done.

Why not just ask maintainers to rebase a PR?

Doug-Lyons:
> > This PR replaces another PR which needed a rebase and I was impatient to get this done.
> 
> Why not just ask maintainers to rebase a PR?

@binarymaster, Honestly, it is not something that I thought about in this context. Next time I will be aware of this option though. Do you want to rebase the other PR and let me move my input to it now? Thanks for the suggestion.

binarymaster:
> Do you want to rebase the other PR and let me move my input to it now?

Yes, let's do this. I have rebased the other PR now.

ntoskrnl/cc/view.c
Doug-Lyons: At one time this IF statement was as follows:
```
    if (current->MappedCount > 0 && !current->Dirty)
```
It was changed by Pierre Schweitzer to what it is now in [14b05e6](https://github.com/reactos/reactos/commit/40017a54f958093ee6a12cd73412aedbe44f0ea1) on 3/24/2018.

Suggestions?

Doug-Lyons:
> @ThFabba In PR #5237 is says "as there might not be enough pages for the balancer itself (ASSERT(MmAvailablePages >= 32);) later on." - Is this still valid in the sense that the paging path relies on pages being available? In that case we need even more changes, because even if the thread runs at higher priority, it can not prevent other threads from eating up the remaining pages before the paging is complete, especially in an SMP environment. We would need to either fix the paging path to preallocate any needed pages or reserve pages that will only be available in the paging path. I would prefer the former, because the latter seems like hard to prove to be correct, but it's probably also extremely difficult.

I believe that you are probably correct here. When I was working with @JoachimHenze he hit a lockup condition that seemed to be related to this. This seems fairly complicated to me. Should it possibly go into another PR? Thanks.

GeoB99 (commented):
If `CcRosTrimCache` was removed from the codebase perhaps because it was done for a reason.  Have you looked into that?

Both the Memory and Cache Managers are so messy it hurts my heart pretty bad, so I'm fine with this being a placebo solution to the problem of memory pages but I want that you provide an explanation or comment in the code explaining why this hack has to exist and why it's not the ultimate permanent solution.

SergeGautherie:
> Fix locking of CcRosVacbIncRefCount in CcRosCreateVacb.

It looks like this part could/should have its own PR, at least its own commit.

JoachimHenze (approved):
I completed my testing. I tested this PR applied on top of 0.4.15-dev-6416-g2d7ff7e. I could with it successfully complete 8 copy-operations witn my file, followed by 8 successful compare operations.
On the 9th try the copy completed, but during the compare I got a "read error".
I could cancel and retry the compare, and it finished successfully then (proving that the file was at least properly written).
Still it had a problem with reading, so it is not perfect. The OS did not entirely freeze yet, though, which is better than the two patches I tested last night.

Yes, that still isn't perfectly stable, but to understand why **I do** still **accept the PR**, is because I want to put things into perspective:

With current un-PR'ed master head it reliably fails during the first compare-attempt already and freezes the OS.
With the 2 patches that I tested last night, the results were also worse:
1.) PR5237 modified with 4 lines less and 2 seconds trim-interval allowed me to complete 4 copy-compare runs successfully, and then it failed on the 5th compare  with read-error, then the OS froze
2.) PR5237 modified with 4 lines less and 1 seconds trim-interval allowed me to complete 5 copy-compare runs successfully and then it failed on the 6th compare with read-error, then the OS froze

So this PR gives me the best result, that I ever had since MM was touched with early 0.4.15-dev.
We are still not on par with 0.4.14-release.
**But we do definitely massively improve with this PR**
Ftr: A sawtooth-like memory restoring/trimming can be observed within the taskmgr again.

Doug-Lyons:
> If `CcRosTrimCache` was removed from the codebase perhaps because it was done for a reason. Have you looked into that?
> 
> Both the Memory and Cache Managers are so messy it hurts my heart pretty bad, so I'm fine with this being a placebo solution to the problem of memory pages but I want that you provide an explanation or comment in the code explaining why this hack has to exist and why it's not the ultimate permanent solution.

@GeoB99, Please see https://github.com/reactos/reactos/pull/5237#issuecomment-1666540389. Thanks.

Doug-Lyons:
> > Fix locking of CcRosVacbIncRefCount in CcRosCreateVacb.
> 
> It looks like this part could/should have its own PR, at least its own commit.

@SergeGautherie, Now this PR is revised to only cover the Lock Fix. Thanks.

JoachimHenze:
Excellent: Investing a whole day into the testing just to find my testresults becoming totally useless just 5minutes after I posted them, because the PR got changed entirely.

ThFabba:
> Excellent: Investing a whole day into the testing just to find my testresults becoming totally useless just 5minutes after I posted them, because the PR got changed entirely.

You were testing the combination of the current state of this PR and #5237. Since this PR is now trivial, your tests should cover #5237 once this gets merged.

GeoB99 (approved):

tkreuzer (approved):

HBelusca (approved):

ThFabba (approved):

Doug-Lyons:
Thanks for the merge @HBelusca and for the approvals from @ThFabba, @tkreuzer, @HBelusca, @JoachimHenze, and @GeoB99.

PR #5528, opened 2023-08-05
[ATL] CImage: Hack fix for getting filter string

katahiromz:
## Purpose

Fix the crash on opening/saving file.
To retrieve the filename extension info from CODECs, we have to keep `gdiplus.dll` loaded.
The perfect fix will take time.
JIRA issue: [CORE-19093](https://jira.reactos.org/browse/CORE-19093), [CORE-19094](https://jira.reactos.org/browse/CORE-19094)

## Proposed changes

- By adding `CImage dummy` local variables, ensure keeping loaded the `gdiplus.dll`.

## TODO

- [x] Do tests.
- [ ] Make `CImage` compatible (in the future).

## Comparison

BEFORE:
0 4 15-dev-6416-g2d7ff7e-FAILED
Crashed!

AFTER:
after
Successful.

### DIFF
--- a/sdk/lib/atl/atlimage.h
+++ b/sdk/lib/atl/atlimage.h
@@ -811,7 +811,7 @@ public:
         DWORD dwExclude = excludeDefaultLoad,
         TCHAR chSeparator = TEXT('|'))
     {
-        CImage dummy; // Initialize common
+        CImage dummy; // HACK: Initialize common
         UINT cDecoders = 0;
         Gdiplus::ImageCodecInfo* pDecoders = _getAllDecoders(cDecoders);
         HRESULT hr = BuildCodecFilterString(pDecoders,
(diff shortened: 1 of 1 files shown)


sdk/lib/atl/atlimage.h
```diff
@@ -1057,6 +1059,8 @@ class CImage
     // Deprecated. Don't use this
     static const GUID *FileTypeFromExtension(LPCTSTR dotext)
     {
+        CImage dummy; // Initialize common
+
```
HBelusca, line 1063:
```suggestion
        CImage dummy; // Initialize common
```
If this "dummy" thing is the actual hack, maybe you should prefix `HACK: ` in from of these comments.

katahiromz: `"HACK:"` is added.

sdk/lib/atl/atlimage.h
```diff
@@ -1057,6 +1059,8 @@ class CImage
     // Deprecated. Don't use this
     static const GUID *FileTypeFromExtension(LPCTSTR dotext)
     {
+        CImage dummy; // HACK: Initialize common
+
```
HBelusca, line 1063:
```suggestion
```

HBelusca (approved):

PR #5529, opened 2023-08-05
[SHELL32] Move SheRemoveQuotesA/W to iconcache.cpp

katahiromz:
## Purpose

Follow-up to #5517 (https://github.com/reactos/reactos/commit/358e45d33a80a8db868324cf1f477c9e24408e53). `stubs.cpp` is for stub functions only.

## Proposed changes

- Move `SheRemoveQuotesA` and `SheRemoveQuotesW` into `iconcache.cpp`.

### DIFF
--- a/dll/win32/shell32/iconcache.cpp
+++ b/dll/win32/shell32/iconcache.cpp
@@ -988,6 +988,54 @@ HICON WINAPI ExtractAssociatedIconW(HINSTANCE hInst, LPWSTR lpIconPath, LPWORD l
 }
 
 /*************************************************************************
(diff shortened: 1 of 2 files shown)


binarymaster (approved):
Mark as addendum to previous commit in master.

HBelusca:
But why moving it to **iconcache**.cpp ? iconcache has **nothing** to do with She functions and adding/removing quotes from string. @katahiromz Please change that and move those to some "utils.cpp" file or sthg else in shell32 (where it makes **more** sense).

PR #5532, opened 2023-08-05
[NMIDEBUG] Fix a bugcheck in the driver when the CPU gets an NMI

disean:
## Proposed changes

- Fix an invalid pointer dereference.
- Rework the logic such that the driver handles the NMI safely on debug builds only


hal/halx86/generic/nmi.c
```diff
@@ -99,7 +99,7 @@ HalHandleNMI(
 
     /* Enter the debugger if possible */
     KiBugCheckData[0] = (ULONG_PTR)KeServiceDescriptorTable; /* NMI Corruption? */
-    if (!KdDebuggerNotPresent && KdDebuggerEnabled)
+    if (!KD_DEBUGGER_NOT_PRESENT && KD_DEBUGGER_ENABLED)
```
HBelusca, line 102: Good catch!

drivers/base/nmidebug/nmidebug.c
```diff
@@ -13,49 +12,31 @@
 
 /* FUNCTIONS ******************************************************************/
 
-PCHAR NmiBegin = "NMI4NMI@";
-
-FORCEINLINE
-VOID
-NmiClearFlag(VOID)
```
HBelusca, line 20: What about that old code? Why didn't it work?

disean: The old code has a pointer dereference bug in two places that I found while running the code in debugger
https://github.com/reactos/reactos/blob/master/drivers/base/nmidebug/nmidebug.c#L28
https://github.com/reactos/reactos/blob/master/drivers/base/nmidebug/nmidebug.c#L46
That NMI callback is doing strange stuff that I didn't quite understand, it looks like just an incomplete code/driver.

HBelusca: In any case, shouldn't one need to read/write to some IO port or do something for the CPU to ignore the NMI that has happened? (It seemed that the idea of the original code was to do that.) With the new code it appears that we break into the debugger (OK) but that's it.

drivers/base/nmidebug/nmidebug.c
```diff
@@ -13,49 +12,31 @@
...
-    NmiClearFlag();
+    /* Handle the NMI safely if a debugger is connected */
+    if (!KD_DEBUGGER_NOT_PRESENT && KD_DEBUGGER_ENABLED)
+    {
+        DbgPrint("NMI Callback entered!\n");
 
-    /* Get NMI status signature */
-    __indwordstring(0x80, (PULONG)NmiBegin, 1);
-    ((void(*)())&KiBugCheckData[4])();
+        DbgBreakPoint();
```
HBelusca, line 26: In https://community.osr.com/discussion/comment/286830/#Comment_286830 I see
> Since an NMI can interrupt anything (and there is almost no code that can safely be called from an NMI callback), synchronizing with any code that is called to handle an NMI without risk of deadlock etc. is difficult and cumbersome, not to mention that NMI code typically cannot be debugged in typical fashion, as the debugger itself relies on NMIs to function (at least on AMD64).

I cannot judge of the exactness of this claim but, if this is true, how is this breakpoint breaking in windbg going to behave?

HBelusca:
(Interesting side-reading about the NMI handler in Windows 8+: http://hypervsir.blogspot.com/2014/10/an-os-kernel-bug-in-windows-81-32-bit-os.html )

drivers/base/nmidebug/nmidebug.c
HBelusca: OK this commit gives slightly more context: https://github.com/reactos/reactos/commit/475b0fd7cb55b963b206d9eddcbfbb2e3bd88748

drivers/base/nmidebug/nmidebug.c
```diff
@@ -13,49 +12,31 @@
...
 {
-    /* Clear the NMI flag */
-    NmiClearFlag();
+    /* Handle the NMI safely if a debugger is connected */
+    if (!KD_DEBUGGER_NOT_PRESENT && KD_DEBUGGER_ENABLED)
+    {
+        DbgPrint("NMI Callback entered!\n");
 
-    /* Get NMI status signature */
-    __indwordstring(0x80, (PULONG)NmiBegin, 1);
```
HBelusca, line 45: This reads *one* (1) dword from port 0x80, which is (on PT-AT compatibles) reading the port for POST error codes (see https://stackoverflow.com/questions/6793899/what-does-the-0x80-port-address-connect-to ).
It stores it in the first 4 bytes in the NmiBegin buffer (erases the first "NMI4" part).

Or... is it used as "dummy" port for delay? (see https://groups.google.com/g/linux.kernel/c/UTKlqyBFiyU )

drivers/base/nmidebug/nmidebug.c
```diff
@@ -13,49 +12,31 @@
...
-    ((void(*)())&KiBugCheckData[4])();
+        DbgBreakPoint();
+
+        return TRUE;
+    }
 
-    /* Handle the NMI safely */
-#ifdef _M_IX86
-    KiEnableTimerWatchdog = (RtlCompareMemory(NmiBegin, NmiBegin + 4, 4) != 4);
```
HBelusca, line 50: Somehow, this checks whether the modified `"NMI4"` part of NmiBegin is equal to the second (untouched) `"NMI@"` part.

drivers/base/nmidebug/nmidebug.c
```diff
@@ -13,49 +12,31 @@
 
 /* FUNCTIONS ******************************************************************/
 
-PCHAR NmiBegin = "NMI4NMI@";
-
-FORCEINLINE
-VOID
-NmiClearFlag(VOID)
-{
-    ((PCHAR)&KiBugCheckData[4])[0] -= (NmiBegin[3] | NmiBegin[7]);
```
HBelusca, line 22: `NmiBegin[3] == '4' == 0x34` and `NmiBegin[7] == '@' == 0x40`, thus `NmiBegin[3] | NmiBegin[7] == 0x74 == 't'`

drivers/base/nmidebug/nmidebug.c
disean: >Or... is it used as "dummy" port for delay?

or to crash the system at certain condition...

drivers/base/nmidebug/nmidebug.c
disean: The kernel trap handler is responsible for clearing the NMI iirc

drivers/base/nmidebug/nmidebug.c
disean: That's interesting... On ROS this patch makes the debugger to break at every NMI invocation.
I have tested also on checked windows 2003 x86 but it doesn't break into the debugger and only prints the message "Port lock was not acquired!".

drivers/base/nmidebug/nmidebug.c
tkreuzer: This code has a very specific purpose, which might not be obvious to the uninitiated reader, and that is the purpose :)

drivers/base/nmidebug/nmidebug.c
securityandstuff: It does seem correct as is.

I was not aware of win2k3 hyper-v, so this is interesting behavior for sure.

drivers/base/nmidebug/nmidebug.c
disean: Removed call to DbgBreakPoint.

learn-more:
@HBelusca please update

PR #5536, opened 2023-08-06
[APPWIZ] CORE-5737: Move focus before SW_HIDE

katahiromz:
## Purpose
Arrow keys were not working.
JIRA issue: [CORE-5737](https://jira.reactos.org/browse/CORE-5737)

## Proposed changes

- Move focus before `SW_HIDE`.
- Swap `ID_DWL_INSTALL` and `IDCANCEL` controls.

## TODO

- [x] Do tests.

## Comparison

BEFORE:
before
Arrow keys were not working. Two buttons swapped.

AFTER:
after
Arrow keys are working. Two buttons correct.


JoachimHenze:
Why do we have to swap the buttons positions for that?

katahiromz:
> Why do we have to swap the buttons positions for that?

It looks like a mistake of specifying the coordinates.
In Windows UI/UX, `[Cancel]` button should be put on the right side of the dialog.

dll/cpl/appwiz/addons.c
```diff
@@ -393,10 +393,16 @@ static DWORD WINAPI download_proc(PVOID arg)
...
-    HWND hwndInstallButton;
+    HWND hwndProgress, hwndInstallButton;
     switch(msg) {
     case WM_INITDIALOG:
-        ShowWindow(GetDlgItem(hwnd, ID_DWL_PROGRESS), SW_HIDE);
+        hwndProgress = GetDlgItem(hwnd, ID_DWL_PROGRESS);
+
+        /* CORE-5737: Move focus before SW_HIDE */
+        if (hwndProgress == GetFocus())
+            SendMessageW(hwnd, WM_NEXTDLGCTL, 0, FALSE);
```
HBelusca, line 403: This looks like a hack. Is the problem also fixed by just keeping the `WS_GROUP | WS_TABSTOP` style for the `DEFPUSHBUTTON` control?

Also, the `WM_INITDIALOG` documentation https://learn.microsoft.com/en-us/windows/win32/dlgbox/wm-initdialog
clearly states:
> The control to receive the default keyboard focus is always the first control in the dialog box that is visible, not disabled, and that has the WS_TABSTOP style. When the dialog box procedure returns TRUE, the system checks the control to ensure that the procedure has not disabled it. If it has been disabled, the system sets the keyboard focus to the next control that is visible, not disabled, and has the WS_TABSTOP.

It may be worth adding a `NOT WS_VISIBLE` to the progress control in the dialog resource too. But it is strange though, because the progress control (while being "visible" by just observing the resources), doesn't have the `WS_TABSTOP` style -- at least explicitly.

HBelusca (changes requested):
The current solution looks like a hack. Could you investigate a bit more, in regards to the comment I mention there?

dll/cpl/appwiz/addons.c
katahiromz: You can ignore here `WS_GROUP` and `WS_TABSTOP` is default style of `PUSHBUTTON`. Deleting them is harmless.

HBelusca: https://learn.microsoft.com/en-us/windows/win32/menurc/pushbutton-control
> style
> Styles for the pushbutton, which can be a combination of the BS_PUSHBUTTON style and the following styles: WS_TABSTOP, WS_DISABLED, and WS_GROUP.
>
> _**If you do not specify a style, the default style is BS_PUSHBUTTON | WS_TABSTOP.**_

WS_TABSTOP yes, but WS_GROUP doesn't appear to be in the list.

katahiromz: IIRC, `WS_GROUP` is the radiobutton-related value. This control isn't a radiobutton.

HBelusca: No, WS_GROUP is focus/navigation (with TAB)-related. It is, of course, especially important for groups of radiobuttons, but not only for that.

katahiromz: We studied the focus problem from CORE-17550.
We have to move the focus before hiding the focus control.

HBelusca: Ah ok i see, you're implementing also the trick mentioned in https://devblogs.microsoft.com/oldnewthing/20040804-00/?p=38243

HBelusca (approved):
In view of the present discussion, it looks OK.

PR #5537, opened 2023-08-06
[MSPAINT] Fix focus and radiobuttons of some dialogs

katahiromz:
## Purpose
Improve keyboard usability.
JIRA issue: [CORE-18011](https://jira.reactos.org/browse/CORE-18011)

## Proposed changes

- Return `TRUE` on `WM_INITDIALOG` handling to set focus.
- Check if the radiobutton is checked on the radiobutton action (considering `BN_SETFOCUS` etc.).

## TODO

- [x] Do tests

### DIFF
--- a/base/applications/mspaint/dialogs.cpp
+++ b/base/applications/mspaint/dialogs.cpp
@@ -74,7 +74,7 @@ LRESULT CMirrorRotateDialog::OnCancel(WORD wNotifyCode, WORD wID, HWND hWndCtl, 
 
 LRESULT CMirrorRotateDialog::OnRadioButton3(WORD wNotifyCode, WORD wID, HWND hWndCtl, BOOL& bHandled)
 {
-    if (!IsDlgButtonChecked(IDD_MIRRORROTATERB3) != BST_CHECKED)
+    if (IsDlgButtonChecked(IDD_MIRRORROTATERB3) != BST_CHECKED)
         return 0;
 
     ::EnableWindow(GetDlgItem(IDD_MIRRORROTATERB4), TRUE);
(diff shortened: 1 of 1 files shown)


JoachimHenze (approved):

JoachimHenze:
gcc chokes on those:```

../src/base/applications/mspaint/dialogs.cpp:77:50: error: logical not is only applied to the left hand side of comparison [-Werror=logical-not-parentheses]
     if (!IsDlgButtonChecked(IDD_MIRRORROTATERB3) != BST_CHECKED)
                                                  ^~
../src/base/applications/mspaint/dialogs.cpp:77:9: note: add parentheses around left hand side expression to silence this warning
     if (!IsDlgButtonChecked(IDD_MIRRORROTATERB3) != BST_CHECKED)
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
         (                                       )
```

File: base/applications/mspaint/dialogs.cpp [hunk 1]
```diff
@@ -74,6 +74,9 @@ LRESULT CMirrorRotateDialog::OnCancel(WORD wNotifyCode, WORD wID, HWND hWndCtl,
 
 LRESULT CMirrorRotateDialog::OnRadioButton3(WORD wNotifyCode, WORD wID, HWND hWndCtl, BOOL& bHandled)
 {
+    if (!IsDlgButtonChecked(IDD_MIRRORROTATERB3) != BST_CHECKED)
```
JoachimHenze, line 77:
```suggestion
    if (IsDlgButtonChecked(IDD_MIRRORROTATERB3) == BST_CHECKED)
```
maybe change all of the IsDlgButtonChecked occurances to something like that?
Multiple hits, I commented just on this one.

JoachimHenze, line 77:
```suggestion
    if (IsDlgButtonChecked(IDD_MIRRORROTATERB3))
```
or just like that?

base/applications/mspaint/dialogs.cpp
```diff
@@ -74,6 +74,9 @@ LRESULT CMirrorRotateDialog::OnCancel(WORD wNotifyCode, WORD wID, HWND hWndCtl,
 
 LRESULT CMirrorRotateDialog::OnRadioButton3(WORD wNotifyCode, WORD wID, HWND hWndCtl, BOOL& bHandled)
 {
+    if (IsDlgButtonChecked(IDD_MIRRORROTATERB3) != BST_CHECKED)
```
HBelusca, line 77: Hmmmm.... How could these OnRadioButtonX be called, to begin with, if these IDD_MIRRORROTATE** radiobutons be not checked (or disabled)?

katahiromz: Have a use the Tab key on your keyboard.
That will move the focus.

base/applications/mspaint/dialogs.cpp
katahiromz: `IsDlgButtonChecked` has a tri-state value. I would use a strict style.

base/applications/mspaint/dialogs.cpp
katahiromz: That will generate `BN_SETFOCUS` without checked.

PR #5551, opened 2023-08-09
[APPWIZ] Don't forget to CoInitializeEx on COM thread

katahiromz:
## Purpose

JIRA issue: [CORE-19110](https://jira.reactos.org/browse/CORE-19110), [CORE-15786](https://jira.reactos.org/browse/CORE-15786)

## Proposed changes

- Do `CoInitializeEx` and `CoUninitialize` in `download_proc`.

### DIFF
--- a/dll/cpl/appwiz/addons.c
+++ b/dll/cpl/appwiz/addons.c
@@ -358,7 +358,9 @@ static DWORD WINAPI download_proc(PVOID arg)
 {
     WCHAR message[256];
     WCHAR tmp_dir[MAX_PATH], tmp_file[MAX_PATH];
-    HRESULT hres;
+    HRESULT hres, hrCoInit;
+
+    hrCoInit = CoInitializeEx(NULL, COINIT_MULTITHREADED);
 
     GetTempPathW(sizeof(tmp_dir)/sizeof(WCHAR), tmp_dir);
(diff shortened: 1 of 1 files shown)


dll/cpl/appwiz/addons.c
```diff
@@ -358,7 +358,9 @@ static DWORD WINAPI download_proc(PVOID arg)
 {
     WCHAR message[256];
     WCHAR tmp_dir[MAX_PATH], tmp_file[MAX_PATH];
-    HRESULT hres;
+    HRESULT hres, hrCoInit;
+
+    hrCoInit = CoInitializeEx(NULL, COINIT_MULTITHREADED);
```
HBelusca, line 363: Not sure we should ignore the return value there.....

HBelusca (commented):
Does this fix the crash problem, or is there more to it?

katahiromz:
> Does this fix the crash

I'm not sure. The crash backtrace is abnormal and not analyzable.

katahiromz:
https://stackoverflow.com/questions/9286600/when-do-i-need-to-call-coinitialize-in-this-scenario

> CoInitialize has to be called in every single thread that uses COM, regardless of what thread it is, or whether it has a parent thread or child threads. If the thread uses COM, it must call CoInitialize.

PR #5557, opened 2023-08-10
[CRT][WINESYNC] Sync crt wine 5.0

tkreuzer:
## Purpose

Syncs some CRT files shared with wine to wine-5.0 and reduces the difference to wine code.

# Tests

x86 KVM:  https://reactos.org/testman/compare.php?ids=88774,88776,88791,88802,88820

### DIFF
--- a/sdk/lib/crt/except/i386/cpp.s
+++ b/sdk/lib/crt/except/i386/cpp.s
@@ -5,6 +5,8 @@
 .code
 .align 4
 
+EXTERN _cxx_frame_handler:PROC
+
 MACRO(DEFINE_THISCALL_ALIAS, cxxname, target)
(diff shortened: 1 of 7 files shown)


JoachimHenze (approved):
Approved based on KVM results and the default-assumption for a sync that newer-Wine == better-Wine, until proven to be wrong.
I would welcome adding the VBox testbots results as well. Especially in the hypothetical case they would look worse!
I would welcome an "explanation for dummies" why this will benefit us for NT6.

DarkFire01 (approved):

binarymaster (commented):
It seems WINESYNC.txt hasn't been updated with new information in this PR.

PR #5559, opened 2023-08-10
[COMCTL32] Fix balloon tooltips and system pager alerts

cbialorucki:
## Purpose

Before
image

After
image

JIRA issue: [CORE-19109](https://jira.reactos.org/browse/CORE-19109)

## Proposed changes
- Set a maximum width for balloons coming from the system pager.
- Limit balloon tooltips from extending past the edges of the monitor, not the edges of the work area.
- Instead of simply repositioning the main rectangle when the balloon is too far towards one edge of the screen, try flipping the balloon the other way. This is the same behavior as Windows Server 2003.
- Tweak some values used to draw the balloon tooltips to more closely follow the Windows balloon tooltip style.
- Removing trailing whitespace.
- While the new changes are guarded, consider cancelling our Wine sync for common controls. Our common controls are responsible for many graphical issues and lack of features throughout ReactOS.

## TODO (in future PRs)
- Add a close button for balloons that come from the system tray.
- Add options to use either NT6+ balloon styling or NT5.x balloon styling.
- Support NT6+ balloon flags.
- More system pager fixes. For example, there is a bug where if an application displays a notification immediately after being added to the pager the balloon is drawn in the wrong place. This existed in our explorer before this pull request.

### DIFF
--- a/base/shell/explorer/syspager.cpp
+++ b/base/shell/explorer/syspager.cpp
@@ -21,6 +21,8 @@
 
 #include "precomp.h"
 
+#define BALLOON_MAXWIDTH 340
+
 struct InternalIconData : NOTIFYICONDATA
 {
     // Must keep a separate copy since the original is unioned with uTimeout.
@@ -617,19 +619,19 @@ void CBalloonQueue::Show(Info& info)
 
     // TODO: NIF_REALTIME, NIIF_NOSOUND, other Vista+ flags
 
-    const int index = IndexOf(info.pSource);
+    m_current = info.pSource;
     RECT rc;
-    m_toolbar->GetItemRect(index, &rc);
+    m_toolbar->GetItemRect(IndexOf(m_current), &rc);
     m_toolbar->ClientToScreen(&rc);
     const WORD x = (rc.left + rc.right) / 2;
     const WORD y = (rc.top + rc.bottom) / 2;
 
     m_tooltips->SetTitle(info.szInfoTitle, info.uIcon);
     m_tooltips->TrackPosition(x, y);
+    m_tooltips->SetMaxTipWidth(BALLOON_MAXWIDTH);
     m_tooltips->UpdateTipText(m_hwndParent, reinterpret_cast<LPARAM>(m_toolbar->m_hWnd), info.szInfo);
     m_tooltips->TrackActivate(m_hwndParent, reinterpret_cast<LPARAM>(m_toolbar->m_hWnd));
 
-    m_current = info.pSource;
     int timeout = info.uTimeout;
     if (timeout < MinTimeout) timeout = MinTimeout;
     if (timeout > MaxTimeout) timeout = MaxTimeout;
--- a/dll/win32/comctl32/tooltips.c
+++ b/dll/win32/comctl32/tooltips.c
@@ -162,10 +162,17 @@ typedef struct
 #define BALLOON_TEXT_MARGIN (NORMAL_TEXT_MARGIN+8)
 /* value used for CreateRoundRectRgn that specifies how much
  * each corner is curved */
+#ifdef __REACTOS__
+#define BALLOON_ROUNDEDNESS 16
+#define BALLOON_STEMHEIGHT 18
+#define BALLOON_STEMWIDTH 18
+#define BALLOON_STEMINDENT 16
+#else
 #define BALLOON_ROUNDEDNESS 20
 #define BALLOON_STEMHEIGHT 13
 #define BALLOON_STEMWIDTH 10
 #define BALLOON_STEMINDENT 20
+#endif // __REACTOS__
 
 #define BALLOON_ICON_TITLE_SPACING 8 /* horizontal spacing between icon and title */
 #define BALLOON_TITLE_TEXT_SPACING 8 /* vertical spacing between icon/title and main text */
@@ -741,6 +748,40 @@ TOOLTIPS_Show (TOOLTIPS_INFO *infoPtr, BOOL track_activate)
     mon_info.cbSize = sizeof(mon_info);
     GetMonitorInfoW( monitor, &mon_info );
 
+#ifdef __REACTOS__
+    if (rect.right > mon_info.rcMonitor.right)
+    {
+        rect.left -= size.cx - (BALLOON_STEMINDENT + BALLOON_STEMWIDTH);
+        rect.right -= size.cx - (BALLOON_STEMINDENT + BALLOON_STEMWIDTH);
+        if (rect.right > mon_info.rcMonitor.right)
+        {
+            rect.left -= (rect.right - mon_info.rcMonitor.right);
+            rect.right = mon_info.rcMonitor.right;
+        }
+    }
+
+    if (rect.left < mon_info.rcMonitor.left)
+    {
+        rect.right += abs(rect.left);
+        rect.left = 0;
+    }
+
+    if (rect.bottom > mon_info.rcMonitor.bottom)
+    {
+        RECT rc;
+        if (toolPtr->uFlags & TTF_IDISHWND)
+        {
+            GetWindowRect((HWND)toolPtr->uId, &rc);
+        }
+        else
+        {
+            rc = toolPtr->rect;
+            MapWindowPoints(toolPtr->hwnd, NULL, (LPPOINT)&rc, 2);
+        }
+	    rect.bottom = rc.top - 2;
+    	rect.top = rect.bottom - size.cy;
+    }
+#else
     if( rect.right > mon_info.rcWork.right ) {
         rect.left -= rect.right - mon_info.rcWork.right + 2;
         rect.right = mon_info.rcWork.right - 2;
@@ -759,6 +800,7 @@ TOOLTIPS_Show (TOOLTIPS_INFO *infoPtr, BOOL track_activate)
 	rect.bottom = rc.top - 2;
     	rect.top = rect.bottom - size.cy;
     }
+#endif // __REACTOS__
 
     AdjustWindowRectEx (&rect, GetWindowLongW (infoPtr->hwndSelf, GWL_STYLE),
 			FALSE, GetWindowLongW (infoPtr->hwndSelf, GWL_EXSTYLE));
@@ -775,7 +817,11 @@ TOOLTIPS_Show (TOOLTIPS_INFO *infoPtr, BOOL track_activate)
         {
           pts[0].x = ptfx;
           pts[0].y = 0;
+#ifdef __REACTOS__
+          pts[1].x = max(BALLOON_STEMINDENT, ptfx - BALLOON_STEMWIDTH);
+#else
           pts[1].x = max(BALLOON_STEMINDENT, ptfx - (BALLOON_STEMWIDTH / 2));
+#endif
           pts[1].y = BALLOON_STEMHEIGHT;
           pts[2].x = pts[1].x + BALLOON_STEMWIDTH;
           pts[2].y = pts[1].y;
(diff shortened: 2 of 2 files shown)


HBelusca:
Suddenly the balloon looks much less shitty. :+1:

HBelusca:
It may be worth sending the comctl32 fix to Wine, I think they could accept it. @DarkShadow44 (aka. Fabian Maurer) might be able to help there.

win32ss/gdi/ntgdi/region.c
```diff
@@ -3662,10 +3662,6 @@ NtGdiCreateRoundRectRgn(
         bottom = tmp;
     }
 
-    /* the region is for the rectangle interior, but only at right and bottom for some reason */
```
HBelusca, line 3665: We can observe that this piece of code has been added (and is still present) in Wine:
https://github.com/wine-mirror/wine/blame/b80ea4153b096970514f86f385a37c9ceaceada2/dlls/win32u/region.c#L704
from this commit https://github.com/wine-mirror/wine/commit/3e51dd751568730d500b7ed7402fbf308a99f028

dll/win32/comctl32/tooltips.c
```diff
@@ -741,22 +741,36 @@ TOOLTIPS_Show (TOOLTIPS_INFO *infoPtr, BOOL track_activate)
...
+    {
+        RECT rc;
+        if (toolPtr->uFlags & TTF_IDISHWND)
+        {
+            GetWindowRect ((HWND)toolPtr->uId, &rc);
+        }
+        else
+        {
+            rc = toolPtr->rect;
+            MapWindowPoints (toolPtr->hwnd, NULL, (LPPOINT)&rc, 2);
```
JoachimHenze, line 771:
```suggestion
            MapWindowPoints(toolPtr->hwnd, NULL, (LPPOINT)&rc, 2);
```

dll/win32/comctl32/tooltips.c
```diff
@@ -741,22 +741,36 @@ TOOLTIPS_Show (TOOLTIPS_INFO *infoPtr, BOOL track_activate)
...
-	    rc = toolPtr->rect;
-	    MapWindowPoints (toolPtr->hwnd, NULL, (LPPOINT)&rc, 2);
-	}
-	rect.bottom = rc.top - 2;
+    if (rect.bottom > mon_info.rcMonitor.bottom)
+    {
+        RECT rc;
+        if (toolPtr->uFlags & TTF_IDISHWND)
+        {
+            GetWindowRect ((HWND)toolPtr->uId, &rc);
```
JoachimHenze, line 766:
```suggestion
            GetWindowRect((HWND)toolPtr->uId, &rc);
```

JoachimHenze (changes requested):
A great PR. 2 style-nitpicks just. My inner compass strongly points towards approval afterwards, unless the Wine guys would prove the comctl32 change to be wrong.
We should aim to stay in sync with Wine in comctl32. Ideally that should happen by them accepting the comctl32 change. If that would be impossible for any reason, then we should at least guard by an ifndef REACTOS.

win32ss/gdi/ntgdi/region.c
cbialorucki: It's wrong. You can't have a balloon tooltip without it.

cbialorucki:
I submitted it to Wine.

https://gitlab.winehq.org/wine/wine/-/merge_requests/3547

cbialorucki:
After some additional testing I was able to determine that the bug was entirely in common controls and did not require an update to our GDI. I have updated this PR to exclusively deal with our common controls.

win32ss/gdi/ntgdi/region.c
cbialorucki: Reverted change.

dll/win32/comctl32/tooltips.c
```diff
@@ -819,7 +819,7 @@ TOOLTIPS_Show (TOOLTIPS_INFO *infoPtr, BOOL track_activate)
         hRgn = CreateRoundRectRgn(0,
                                   (infoPtr->bToolBelow ? BALLOON_STEMHEIGHT : 0),
                                   rect.right - rect.left,
-                                  (infoPtr->bToolBelow ? rect.bottom - rect.top : rect.bottom - rect.top - BALLOON_STEMHEIGHT),
+                                  (infoPtr->bToolBelow ? rect.bottom - rect.top : rect.bottom - rect.top - BALLOON_STEMHEIGHT + 1),
```
HBelusca, line 822: you say "with our GDI", but what happens if you try to test a "frankenstein ReactOS" where only the win32k/gdi32/user32/winsrv.dll (and maybe one or two other things) are changed with their Windows 2003 SP2 counterparts?

HBelusca:
> After some additional testing I was able to determine that the bug was entirely in common controls and did not require an update to our GDI. I have updated this PR to exclusively deal with our common controls.

You may update the PR description to reflect that.

cbialorucki:
> You may update the PR description to reflect that.

Done.

dll/win32/comctl32/tooltips.c
cbialorucki: When I tested it, it seemed that the issue was our GDI; however I may have done the Frankenstein incorrectly. When another person Frankensteined it, it seemed that the GDI did not need to be adjusted; only common controls.

cbialorucki:
Since Wine is not interested in this fix, I guarded my changes using `#ifdef __REACTOS__`

base/shell/explorer/syspager.cpp
```diff
@@ -617,7 +619,8 @@ void CBalloonQueue::Show(Info& info)
 
     // TODO: NIF_REALTIME, NIIF_NOSOUND, other Vista+ flags
 
-    const int index = IndexOf(info.pSource);
+    m_current = info.pSource;
+    const int index = IndexOf(m_current);
```
HBelusca, line 623: note: I even wonder whether it could be possible to call the IndexOf(...) directly as the 1st parameter of the GetItemRect call below (optional).

dll/win32/comctl32/tooltips.c
```diff
@@ -162,10 +162,17 @@ typedef struct
...
+#define BALLOON_ROUNDEDNESS 16
+#define BALLOON_STEMHEIGHT 18
+#define BALLOON_STEMWIDTH 18
+#define BALLOON_STEMINDENT 16
+#else
 #define BALLOON_ROUNDEDNESS 20
 #define BALLOON_STEMHEIGHT 13
 #define BALLOON_STEMWIDTH 10
 #define BALLOON_STEMINDENT 20
+#endif
```
HBelusca, line 175:
```suggestion
#endif // __REACTOS__
```

dll/win32/comctl32/tooltips.c
```diff
@@ -759,6 +800,7 @@ TOOLTIPS_Show (TOOLTIPS_INFO *infoPtr, BOOL track_activate)
 	rect.bottom = rc.top - 2;
     	rect.top = rect.bottom - size.cy;
     }
+    #endif
```
HBelusca, line 803:
```suggestion
    #endif // __REACTOS__
```

dll/win32/comctl32/tooltips.c
HBelusca: ok so here the "+ 1" in the REACTOS-specific case is still needed?

HBelusca (approved):
minor questions, otherwise LGTM

JoachimHenze (approved):

dll/win32/comctl32/tooltips.c
cbialorucki: It is still needed. I believe it's required for Windows' GDI also.

base/shell/explorer/syspager.cpp
cbialorucki: Done.

DarkShadow44:
Eh, it's not rejected per se, it often just takes an eternity for someone to review/answer. You should probably clean up your commit history and rebase, but otherwise you can only wait.

whindsaks:
I know DPI scaling is not really a priority but is adding even more magic values unavoidable?

Perhaps take a step in the right direction by doing:

```
template<class T> inline T SHELL_DpiScale(HWND hWnd, T Value)
{
    // FIXME: GetDpiForWindow + MulDiv
    return Value;
}

...

m_tooltips->SetMaxTipWidth(SHELL_DpiScale(m_hwndParent, BALLOON_MAXWIDTH));
```

dll/win32/comctl32/tooltips.c
```diff
@@ -775,7 +817,11 @@ TOOLTIPS_Show (TOOLTIPS_INFO *infoPtr, BOOL track_activate)
         {
           pts[0].x = ptfx;
           pts[0].y = 0;
+          #ifdef __REACTOS__
```
JoachimHenze, line 820:
```suggestion
#ifdef __REACTOS__
```
Please wait with the merging! These were not here yet when I watched and approved the PR last time!
Please move all of them and the #else and #endif s everywhere to column 0!

JoachimHenze (changes requested):

HBelusca:
> I know DPI scaling is not really a priority but is adding even more magic values unavoidable?
> ```
> [...]
> m_tooltips->SetMaxTipWidth(SHELL_DpiScale(BALLOON_MAXWIDTH, m_hwndParent));
> ```

Isn't the original size value passed in the dpiscale helper less magic than what was done without it?

whindsaks:
> Isn't the original size value passed in the dpiscale helper less magic than what was done without it?

The magic pixel size roughly relates to how many characters fits on a line before it wraps.

In a theoretical world where ROS supports per-monitor (or even per-window) DPI, if the DPI is higher the font is usually larger and you cannot fit as many characters per line if you don't also increase the width of the window.

cbialorucki:
> > > I know DPI scaling is not really a priority but is adding even more magic values unavoidable?
> > Isn't the original size value passed in the dpiscale helper less magic than what was done without it?
> The magic pixel size roughly relates to how many characters fits on a line before it wraps.

I don't think we need to worry about this yet. We aren't getting rid of any "magic" values by doing this and we don't know how explorer will evolve to become DPI aware. It's very likely that doing this will just create more work for the person wanting to implement DPI scaling in the future.

whindsaks:
> We aren't getting rid of any "magic" values by doing this

We are because the balloon would have the same "physical" size no matter the DPI instead of the balloon being tiny of the DPI is very high (400% scaling etc).

> It's very likely that doing this will just create more work for the person wanting to implement DPI scaling in the future.

My suggested `SHELL_DpiScale` is the same as the example @ [High DPI Desktop Application Development on Windows](https://learn.microsoft.com/en-us/windows/win32/hidpi/high-dpi-desktop-application-development-on-windows#example) is doing in its `UpdateButtonLayoutForDpi`.

cbialorucki:
Wouldn't any generic DPI conversion logic happen somewhere else in explorer, not in `syspager.cpp`? You are just creating additional work for anyone that wants to properly add DPI scaling.

whindsaks:
> Wouldn't any generic DPI conversion logic happen somewhere else in explorer, not in `syspager.cpp`?

`SHELL_DpiScale` (the function that scales your pixel value) can be placed in util.cpp, that does not change the fact that `SetMaxTipWidth` should not be called with a hardcoded pixel value!

binarymaster:
⚠️ Merge of this PR resulted in two badly formatted commits.

- @cbialorucki please use proper style for commit messages.
- @GeoB99 please use Squash & Merge when PR commit history does not follow the rules.

PR #5560, opened 2023-08-10
[EXPLORER] Fix icon adjustment issue

cbialorucki:
## Purpose
Fixes an issue where when toggling between large and small taskbar icons toolbars ends up on top of the task items. This is most easily seen using quick launch and toggling between large and small icons.

JIRA issue: [CORE-19112](https://jira.reactos.org/browse/CORE-19112)

## Proposed changes
- When setting the taskbar to use small icons, shrink the taskbar down to use only one line.

### DIFF
--- a/base/shell/explorer/traywnd.cpp
+++ b/base/shell/explorer/traywnd.cpp
@@ -3528,7 +3528,7 @@ HandleTrayContextMenu:
             AlignControls(NULL);
         }
 
-        /* Forces taskbar to use one line when using small icons. */
+        /* Force taskbar to use one line when using small icons */
         if (g_TaskbarSettings.bSmallIcons)
             pTaskbarRect->top = pTaskbarRect->bottom;
 
--- a/base/shell/explorer/traywnd.cpp
+++ b/base/shell/explorer/traywnd.cpp
@@ -3488,6 +3488,7 @@ HandleTrayContextMenu:
     LRESULT OnTaskbarSettingsChanged(UINT uMsg, WPARAM wParam, LPARAM lParam, BOOL& bHandled)
     {
         TaskbarSettings* newSettings = (TaskbarSettings*)lParam;
+        RECT *pTaskbarRect = &m_TrayRects[m_Position];
 
         /* Propagate the new settings to the children */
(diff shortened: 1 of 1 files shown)


base/shell/explorer/traywnd.cpp
```diff
@@ -3527,7 +3528,10 @@ class CTrayWindow :
             AlignControls(NULL);
         }
 
-        /* Adjust taskbar size */
+        /* Forces taskbar to use one line when using small icons. */
```
HBelusca, line 3531:
```suggestion
        /* Forces taskbar to use one line when using small icons */
```

HBelusca (approved):

yagoulas:
Can you explain why this change has this result? I don't really understand how this works.

JoachimHenze (approved):
Approved. +1 for Hermes suggestion with the comment.

cbialorucki:
> Can you explain why this change has this result? I don't really understand how this works.

`CheckTrayWndPosition()` calls `FitToRebar()` using `m_TrayRects[m_Position]`. By setting the height of `m_TrayRects[m_Position]` to zero, `FitToRebar()` calculates how to place the toolbars with the minimum height needed and resizes the taskbar. The rest of the `CheckTrayWndPosition()` ensures that the taskbar is placed in the correct position.

Maybe I can do better reworking the FitToRebar method.

binarymaster (commented):
I feel like the problem should be fixed another way:
1. Assume the initial taskbar has small icons
2. Apply big icons - this will adjust the height of the taskbar in order to fit all elements, but the minimum wanted size should remain unchanged
3. Apply small icons - initial minimum wanted size should be restored

This way we could respect the size selected by the user (taskbar could be resized when unlocked, then locked with specific minimum wanted size).

Looking at the screen capture in the ticket, it seems the minimum wanted size is being modified by enabling big icons.

cbialorucki:
Somehow our rebar gets set with the `RBS_VARHEIGHT` flag which I think is the root cause of this. Problem is I can't find where this is being set.

image

https://learn.microsoft.com/en-us/windows/win32/controls/rebar-control-styles

PR #5554, opened 2023-08-10
[MSPAINT] Support converting to black and white

katahiromz:
## Purpose

JIRA issue: [CORE-19094](https://jira.reactos.org/browse/CORE-19094)

無題

## Proposed changes

- Add `ImageModel::IsBlackAndWhite` and `ImageModel::PushBlackAndWhite` helper functions.
- Add `CAttributesDialog::m_bBlackAndWhite`.
- If `IDD_ATTRIBUTESRB4` is checked, then make the bitmap black and white.
- Add `IDS_LOSECOLORINFO` resource string to show warning message.

## TODO

- [x] Do tests.

PR #5563, opened 2023-08-11
[MSPAINT] Speed up for black and white

katahiromz:
## Purpose

Follow-up to #5554.
JIRA issue: [CORE-19094](https://jira.reactos.org/browse/CORE-19094)

## Proposed changes

- Speed up `ImageModel::PushBlackAndWhite` by using `GetDIBits` and `SetDIBits`.

## TODO

- [x] Do tests.

## Test results

The bitmap conversion of 5000x5000 pixels takes 3 seconds on x64 Win10.

### DIFF
--- a/base/applications/mspaint/dib.cpp
+++ b/base/applications/mspaint/dib.cpp
@@ -538,7 +538,7 @@ BOOL IsBitmapBlackAndWhite(HBITMAP hbm)
 
     BITMAPINFOEX bmi;
     ZeroMemory(&bmi, sizeof(bmi));
-    bmi.bmiHeader.biSize = sizeof(BITMAPINFOHEADER);
+    bmi.bmiHeader.biSize = sizeof(bmi.bmiHeader);
     bmi.bmiHeader.biWidth = bm.bmWidth;
     bmi.bmiHeader.biHeight = bm.bmHeight;
     bmi.bmiHeader.biPlanes = 1;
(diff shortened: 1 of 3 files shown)


katahiromz:
> CImage::GetBits

This bits doesn't assume 24bpp (bits per pixels). Useless.

HBelusca:
> > CImage::GetBits
> 
> This bits doesn't assume 24bpp (bits per pixels). Useless.

Well, suppose you are opening an already monochrome bitmap in mspaint. Why would one need to convert it first to 24bpp and then do the loop over pixels in order to determine whether it's a black-and-white image?
Same question if you open a 4 (or 8)-bit bitmap (like the ROS bootscreens).

katahiromz:
> Why would one need to convert it first to 24bpp

Because handling the bitmap bits in 24bpp is the easiest way.

katahiromz:
Now conversion to 24bpp is omitted.

base/applications/mspaint/dib.cpp
```diff
@@ -519,3 +526,89 @@ HBITMAP BitmapFromHEMF(HENHMETAFILE hEMF)
...
+    BITMAP bm;
+    if (!::GetObjectW(hbm, sizeof(bm), &bm))
+        return FALSE;
+
+    if (bm.bmBitsPixel == 1)
+        return TRUE;
+
+    BITMAPINFOEX bmi;
+    ZeroMemory(&bmi, sizeof(bmi));
+    bmi.bmiHeader.biSize = sizeof(BITMAPINFOHEADER);
```
HBelusca, line 541: It should be possible to do:
```suggestion
    bmi.bmiHeader.biSize = sizeof(bmi.bmiHeader);
```

HBelusca (commented):
Hmm, so finally you rely on the GDI to do the black-and-white conversion?

PR #5569, opened 2023-08-12
[GITHUB] Update Github actions to V3 and replace deprecated set-output

maj113:
## Purpose

GitHub is starting to Discontinue Actions V2 (the date was postponed), when they enforce Actions V3 builds will probably fail to run, so its better to change this sooner rather than later.
This also silences/fixes the warnings on the builds/artifacts page

JIRA issue: None

## Proposed changes

- Update Actions (`cache`, `checkout`, `upload-artifact`) to V3
- Replace `set-output` with `$GITHUB_OUTPUT`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,8 +21,8 @@ jobs:
       id: get_rosbe_spec
       run: |
         gcc -march=native -Q --help=target | grep "\-march= " | awk '{print $NF}'
-        echo name=march-sha::$(gcc -march=native -Q --help=target | sha1sum | awk '{print $1}') >> $GITHUB_OUTPUT
-        echo name=git-sha::$(git ls-remote https://github.com/zefklop/RosBE.git | grep unix_amd64 | awk '{print $1}') >> $GITHUB_OUTPUT
+        echo march-sha=$(gcc -march=native -Q --help=target | sha1sum | awk '{print $1}') >> $GITHUB_OUTPUT
+        echo git-sha=$(git ls-remote https://github.com/zefklop/RosBE.git | grep unix_amd64 | awk '{print $1}') >> $GITHUB_OUTPUT
         wget https://gist.githubusercontent.com/zefklop/b2d6a0b470c70183e93d5285a03f5899/raw/build_rosbe_ci.sh
     - name: Get RosBE
       id: get_rosbe
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,8 +21,8 @@ jobs:
       id: get_rosbe_spec
       run: |
         gcc -march=native -Q --help=target | grep "\-march= " | awk '{print $NF}'
-        echo ::set-output name=march-sha::$(gcc -march=native -Q --help=target | sha1sum | awk '{print $1}')
-        echo ::set-output name=git-sha::$(git ls-remote https://github.com/zefklop/RosBE.git | grep unix_amd64 | awk '{print $1}')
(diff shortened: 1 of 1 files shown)


katahiromz:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

katahiromz:
https://stackoverflow.com/questions/73473617/what-is-the-difference-between-github-actions-checkoutv3-and-v2

katahiromz:
https://github.com/actions/cache

maj113:
The changes between the two versions are more or less minimal, the only major thing being the nodejs upgrade from 12 to 16 so this shouldn't break anything.

Node JS 12 will apparently stop being supported this summer...

.github/workflows/build.yml
```diff
@@ -21,12 +21,12 @@ jobs:
       id: get_rosbe_spec
       run: |
         gcc -march=native -Q --help=target | grep "\-march= " | awk '{print $NF}'
-        echo ::set-output name=march-sha::$(gcc -march=native -Q --help=target | sha1sum | awk '{print $1}')
-        echo ::set-output name=git-sha::$(git ls-remote https://github.com/zefklop/RosBE.git | grep unix_amd64 | awk '{print $1}')
+        echo name=march-sha::$(gcc -march=native -Q --help=target | sha1sum | awk '{print $1}') >> $GITHUB_OUTPUT
+        echo name=git-sha::$(git ls-remote https://github.com/zefklop/RosBE.git | grep unix_amd64 | awk '{print $1}') >> $GITHUB_OUTPUT
```
SergeGautherie, line 25: According to documentation, syntax changed a bit too.
Please, test and comment about that.
```suggestion
        echo march-sha=$(gcc -march=native -Q --help=target | sha1sum | awk '{print $1}') >> $GITHUB_OUTPUT
        echo git-sha=$(git ls-remote https://github.com/zefklop/RosBE.git | grep unix_amd64 | awk '{print $1}') >> $GITHUB_OUTPUT
```

SergeGautherie (commented):
https://github.com/actions/upload-artifact
Note: It looks like 'What's new' list is still the one from v2.

.github/workflows/build.yml
maj113: > According to documentation, syntax changed a bit too. Please, test and comment about that.

Ah i see, i read the docs wrong, thanks for pointing it out

maj113: I'll check how the builds behave and then resolve this conversation accordingly

SergeGautherie: > I am still slightly concerned about why its not finding a cache hit

New/fixed code looks good to me, although I did not check/test cache behavior.

SergeGautherie (approved):

.github/workflows/build.yml
maj113: Seems it just a coincidence, cache works for this branch on my fork https://github.com/maj113/reactos/actions/runs/5846326122/job/15851388771

`Cache restored from key: RosBE-CI-Linux-8ce17749606c36b104f9dd3fe630f8de8d828498-f94ba214bd0582e7cbd2ef5d0d60cd749a17aa21-78dcab8d2d31b8dfd0f36bc45490b7a9a3c227659db3919bbb8ac9b121ccdf95`

JoachimHenze (approved):
Approved. I can confirm those warnings in the Artifacts section of other PRs. And I do confirm, that they are gone with this PR.

binarymaster (approved):

PR #5570, opened 2023-08-13
[BOOTDATA][NVME][DDK] Bare Bones NVMe

atharva1910:
### Overview
A Working NVMe driver. The driver is takes some inspiration from NVMe OF driver (https://github.com/gigaherz/nvmewin). 
The OpenFabric driver is written for NVMe over Fabrics and loads over StorPort. Its a huge driver with compliance to spec which ROS wouldn't need for generations to come and frankly a little ugly (no offense intended to anyone)

This driver was rewritten from scratch for ScsiPort Initialization and teardown which was about 80% of the code. Since there wasn't much needed for ROS to work with this driver at that point, I decided to rewrite the IO flow as well. In all honesty the driver is rewritten entirely for ROS. The structures are borrowed from the OF driver and thats it. Wherever the code is borrowed I have included the licenses.

### Testing [VBox] :
- Boot up
The OS is able to boot up with the NVMe driver loaded on boot. This includes if a NVMe target is present or absent
If a NVMe target is present under a controller. The driver loads and initializes the target, Creates Admin and IO queues and enables the hardware.
If no NVMe target is present, the driver will fail at initialization and continue with ROS boot as before

- Power Down
The OS is able to power down with the NVMe driver loaded. This includes if a NVMe target is present or absent.
If no NVMe target is present, the driver will not be loaded at boot and the power down will work as before.
TODO: If a NVMe device is present under a controller. The driver should destroys Admin and IO queues and enables the hardware. Currently the Filesystem does not send a Flush Cache and SRB Shutdown. This should be included to a clean shutdown.

- Read/Writes
The NVMe driver is able to perform reads and writes on the underlying target. This was tested with some instrumentation to the code

- Boot Device
Currently VBox BIOS firmware does not support NVMe driver and ROS does not support UEFI boot. So this will be tested once UEFI is merged in

For now ROS isn't able to create a "good" volume from a Raw disk. The diskpart utility is able to create a volume but the volume is not in correct state, so the read/writes aren't allowed (this was tested with NVMe driver and the UniAta driver). The next steps would be to get the diskpart working so we are able to perform IOs on the volume.

### Testing [QEMU}:
Qemu isn't tested with NVMe target attached to the VM
VMs on QEMU work fine with no NVMe target. There is no change to VMs. As long as people dont attach a NVMe target, things should work *as is*

### TODO [Order of Priority]:
- [ ] Fix the Diskpart to create a good partition/volume capable of IOs
- [ ] Enable Interrupts 
- [ ] Support for Flush Cache
- [ ] Support for SRB Shutdown
- [ ] Allow Multiple Namespaces
- [ ] Support for Unmap
- [ ] Support for IO queues/CPU (doesn't matter now since only one CPU)
- [ ] Fix VirtualBox to comply to spec


boot/bootdata/txtsetup.sif
```diff
@@ -91,6 +91,7 @@ acpi.sys     = 1,,,,,,,4,1,,,1,4
 nmidebug.sys = 1,,,,,,x,4,,,,1,4
 sacdrv.sys   = 1,,,,,,x,4,,,,1,4
 uniata.sys   = 1,,,,,,x,4,,,,1,4
+nvme.sys   = 1,,,,,,x,4,,,,1,4
```
binarymaster, line 94:
```suggestion
nvme.sys     = 1,,,,,,x,4,,,,1,4
```

boot/bootdata/txtsetup.sif
```diff
@@ -157,6 +158,7 @@ PCI\CC_0601 = isapnp
 PCI\CC_0604 = pci
 PCI\VEN_104B&CC_0100 = buslogic
 PCI\CC_0101 = pciide
+PCI\CC_010802=nvme
```
binarymaster, line 161:
```suggestion
PCI\CC_010802 = nvme
```

drivers/storage/CMakeLists.txt
```diff
@@ -4,3 +4,4 @@ add_subdirectory(ide)
 add_subdirectory(mountmgr)
 add_subdirectory(partmgr)
 add_subdirectory(port)
+add_subdirectory(nvme)
```
binarymaster, line 7: Move to line 5 (alphabet order).

drivers/storage/nvme/nvme.inf
```diff
@@ -0,0 +1,80 @@
...
+DefaultDestDir = 11
+
+[ControlFlags]
+ExcludeFromSelect = *
+
+[Manufacturer]
+%ROS%=NVME_COMM
+
+[NVME_COMM]
+%NVME_MGMT_PORT.DeviceDesc%=nvme_inst, PCI\CC_0108
```
binarymaster, line 20: Inconsistent hardware id with txtsetup.sif entry, should be fixed here or there.

drivers/storage/nvme/nvme.inf
```diff
@@ -0,0 +1,80 @@
+; Copyright (c) 2002-2012 ROS group
```
binarymaster, line 1: Needs to be corrected.

drivers/storage/nvme/NvmeStd.c
```diff
@@ -0,0 +1,330 @@
+#include "precomp.h"
```
binarymaster, line 1: I'd suggest to add standard headers per our coding style, at least where there are no header specified.

binarymaster (commented):
General comment: Add entry to "3rd Party files.txt" and fix "no new line at end of file" problem in all related files.

drivers/storage/nvme/NvmeProbe.c
```diff
@@ -0,0 +1,645 @@
...
+    pDevExt->NSInfo.IsNamespaceReadOnly = FALSE;
+    pDevExt->NSInfo.nsReady = TRUE;
+
+    return TRUE;
+}
+
+BOOLEAN
+NVMeInitIOQueues(PNVME_DEVICE_EXTENSION pDevExt)
+{
+    if (pDevExt == NULL)
```
ThFabba, line 550: Why/how would this be null? Same question for the two other functions above.

drivers/storage/nvme/NvmeProbe.c
```diff
@@ -0,0 +1,645 @@
...
+                  sizeof(ADMIN_IDENTIFY_CONTROLLER));
+
+    DbgPrint("Sucessfully identified NVMe controller 0x%x 0x%x\n", pDevExtension->controllerIdentifyData.VID, pDevExtension->controllerIdentifyData.SSVID);
+
+    /* TODO: Currenly we are limiting IOs to 1MB but if the controller somehow demands lower max transfer we need to set that up here */
+    return TRUE;
+}
+
+BOOLEAN
+CreateNS(PNVME_DEVICE_EXTENSION pDevExt)
```
ThFabba, line 430: Is this called from anywhere?

ThFabba (commented):
> In all honesty the driver is rewritten entirely for ROS.

This is not really a good thing, for two reasons:
* If we import a third-party driver, we can benefit from future work done by the upstream driver's maintainers. If we fork or do our own thing, maintenance is completely on us -- and we're already short on man-power as-is
* We want to be compatible with Windows. To achieve that, drivers should be developed such that they work on Windows (with Driver Verifier), then imported into ROS. If the driver then doesn't work, that means another ROS component needs to be fixed. If we hack our drivers to work on ROS, we end up with a mess of incompatible assumption where neither the drivers nor the OS are compatible with their Windows counterpart.

SergeGautherie:
This PR looks like a Draft one...

>     * If we import a third-party driver

As in: if your driver does not match ROS repository goal, it could live elsewhere and be used by people needing it.

>     * We want to be compatible with Windows.

As in: which Windows (XP/S2003/7/...) does it work on?

drivers/storage/nvme/nvme.inf
atharva1910: Oops good catch. Will fix

drivers/storage/nvme/NvmeIo.c
```diff
@@ -0,0 +1,99 @@
...
+    pCmpQ->CplQHeadPtr++;
+    if (pCmpQ->CplQHeadPtr > pCmpQ->CplQEntries) {
+        pCmpQ->CplQHeadPtr = 0;
+    }
+
+    /* Copy the completion */
+    if (pRetCmpEntry)
+        RtlCopyMemory(pRetCmpEntry, pCmpEntry, sizeof(NVMe_COMPLETION_QUEUE_ENTRY));
+
+    /* What a fucking waste. Fuck you Virtual Box */
```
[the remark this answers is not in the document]
atharva1910: Okay, will remove

drivers/storage/nvme/NvmeProbe.c
```diff
@@ -0,0 +1,645 @@
...
+    PHYSICAL_ADDRESS phySubQ = MmGetPhysicalAddress(AdminSubQ);
+    WRITE_REGISTER_ULONG(&pAE->pCtrlRegister->ASQ.LowPart, phySubQ.LowPart);
+    WRITE_REGISTER_ULONG(&pAE->pCtrlRegister->ASQ.HighPart, phySubQ.HighPart);
+    PHYSICAL_ADDRESS phyComQ = MmGetPhysicalAddress(AdminComQ);
+    WRITE_REGISTER_ULONG(&pAE->pCtrlRegister->ACQ.LowPart, phyComQ.LowPart);
+    WRITE_REGISTER_ULONG(&pAE->pCtrlRegister->ACQ.HighPart, phyComQ.HighPart);
+    
+    return TRUE;
+
+ERR_FREE_MEM:
```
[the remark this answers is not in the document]
atharva1910: We use that PRP list for Admin commands later and in removed later in unload

drivers/storage/nvme/NvmeProbe.c
```diff
@@ -0,0 +1,645 @@
...
+    if (barSize == 0){
+        DbgPrint("Failed to get barsize of NVMe Device\n");
+        return FALSE;
+    }
+    pAE->CtrlRegSize = barSize;
+
+    /* TODO: We should ideally check with what the scsiport has offered as ranges */
+
+    /* MMap BAR0 */
+    PNVMe_CONTROLLER_REGISTERS pNvmeCtrlRegs = (PNVMe_CONTROLLER_REGISTERS)MmMapIoSpace(barAddress, barSize, FALSE);
```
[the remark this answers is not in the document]
atharva1910: Good catch will fix

drivers/storage/nvme/NvmeStd.c
```diff
@@ -0,0 +1,330 @@
...
+
+    /* Set required extension sizes. */
+    hwInitData.DeviceExtensionSize = sizeof(NVME_DEVICE_EXTENSION);
+
+    /* Lets see if we have NVMe device */
+    do {
+        Status = ScanBusAndFindNvme(&isNvmeDevPresent, &pciNvmeDeviceInfo.Bus, &pciNvmeDeviceInfo.Slot);
+
+        if (!NT_SUCCESS(Status) || !isNvmeDevPresent)
+           break;
```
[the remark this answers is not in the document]
atharva1910: Yes driverentry should return a status from ScsiPortInitialize

drivers/storage/nvme/precomp.h
```diff
@@ -0,0 +1,60 @@
...
+ ** Integrated Device Technology Inc., or Sandforce Corporation.              **
+ **                                                                           **
+ *******************************************************************************
+**/
+
+/*
+ * File: precomp.h
+ */
+#include <ntddk.h>
+#include <scsi.h>               // includes scsi.h
```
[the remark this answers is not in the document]
atharva1910: From the Open Frabic repo, will remove

sdk/include/ddk/scsi.h
```diff
@@ -2322,7 +2322,7 @@ typedef struct _SCSI_EXTENDED_MESSAGE {
 #ifndef _INQUIRYDATA_DEFINED /* also in minitape.h */
 #define _INQUIRYDATA_DEFINED
 
-#define INQUIRYDATABUFFERSIZE 36
+#define INQUIRYDATABUFFERSIZE 0x60
```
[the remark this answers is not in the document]
atharva1910: The size in the public header is wrong. Its fixed here.

drivers/storage/nvme/NvmeIo.c
```diff
@@ -0,0 +1,99 @@
...
+    RtlCopyMemory(pCmd, pNvmeCmd, sizeof(NVMe_COMMAND));
+
+    /* Ring The Door bell */ 
+    WRITE_REGISTER_ULONG(pSubQ->pSubTDBL, pSubQ->SubQTailPtr);
+
+    /* Poll for completion */ 
+    PNVMe_COMPLETION_QUEUE_ENTRY pCmpEntry = (PNVMe_COMPLETION_QUEUE_ENTRY)pCmpQ->CplQStartVirtual;
+    pCmpEntry += pCmpQ->CplQHeadPtr;
+
+    while(TRUE) {
```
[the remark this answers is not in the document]
atharva1910: I want to keep it this way because I want to add checkes for Phase and timeout here

drivers/storage/nvme/NvmeProbe.c
atharva1910: Its not currently but this is not guaranteed to be only called from the same function. Also checking passed parameters is always a good thing when it comes to writting C.
Its a personal prefrence really. And dosen't hurt

drivers/storage/nvme/NvmeProbe.c
atharva1910: Yes this should be called during Init if there's no namespace on the target. Currently its not called from anywhere

atharva1910:
> This is not really a good thing, for two reasons:
> 
>     * If we import a third-party driver, we can benefit from future work done by the upstream driver's maintainers. If we fork or do our own thing, maintenance is completely on us -- and we're already short on man-power as-is
There is no third party driver which is compatible with ROS. The driver works on 
>     * We want to be compatible with Windows. To achieve that, drivers should be developed such that they work on Windows (with Driver Verifier), then imported into ROS. If the driver then doesn't work, that means another ROS component needs to be fixed. If we hack our drivers to work on ROS, we end up with a mess of incompatible assumption where neither the drivers nor the OS are compatible with their Windows counterpart.

drivers/storage/nvme/NvmeIo.c
```diff
@@ -0,0 +1,99 @@
...
+        return FALSE;
+    
+    if (pCmpEntry->DW3.SF.SC == 0 && pCmpEntry->DW3.SF.SCT == 0) {
+        return TRUE;
+    }
+
+    /* TODO: Check for retryable error we should not be failing commands */
+    ASSERT(FALSE);
+
+    DbgPrint("Failed Command CID:0x%x on SQID:0x%x SQDH:0x%x. SC: 0x%x SCT:0x%x\n", pCmpEntry->DW3.CID,
```
JoachimHenze, line 16: This line contains EOL-whitespaces. I will not point to all of them in each and every file you touched, but I do recommend you to use the Notepad++ feature which can fix all of them by a single click.

JoachimHenze (changes requested):

drivers/storage/nvme/NvmeProbe.c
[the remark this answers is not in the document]
ThFabba: I'm talking about the error path below. You're freeing the other two allocations but not this one.

sdk/include/ddk/scsi.h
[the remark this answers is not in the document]
ThFabba: No, `INQUIRYDATABUFFERSIZE` is defined as 36 is Microsoft's version of this header.

drivers/storage/nvme/NvmeIo.c
atharva1910: Thanks for the suggestion, will do

gonzoMD:
> > In all honesty the driver is rewritten entirely for ROS.
> 
> This is not really a good thing, for two reasons:
> 
> * If we import a third-party driver, we can benefit from future work done by the upstream driver's maintainers. If we fork or do our own thing, maintenance is completely on us -- and we're already short on man-power as-is

The linked origin repo is in an archived state, so it's unlikely that there is any future work.
The origin repo belongs to @gigaherz. Maybe He wants to join this PR?

gigaherz:
Hi!
Just wanted to point out that the repo was a mirror I made of what looked to be a dead project which was hosted in a dying server. I have nothing to do with that project and I have absolutely no idea how it works. :P
It seems that original site and svn repository dosn't even exist anymore which makes me think I was lucky to come across it when I did, but yeah I don't think I can be of any help here.
Sorry!

atharva1910:
Hi Guys, Apologies for the delay, I'm working on getting the driver working with Win XP. Once validated that its in compliance with windows I shall resubmit the PR again

HBelusca:
yay @gigaherz you're alive!

PR #5571, opened 2023-08-13
[TASKMGR] Do not disable status bar when opening system menu 

Cleverking2003:
## Purpose

We do not need to clean the task manager status bar when opening the system menu as it does not show any hints. Also fix a bug where WM_INITMENUPOPUP wouldn't recognise the system menu. 

JIRA issue: [CORE-19061](https://jira.reactos.org/browse/CORE-19061)

image


KRosUser2:
Did you check other places for similar bug in the code base : https://git.reactos.org/?p=reactos.git&a=search&st=grep&s=WM_INITMENUPOPUP ? clipbrd, mscutils, mspaint, notepad, winhlp32, wordpad, ...

JoachimHenze:
Why is that coupled with the NTUSER changes? What would be the symptom/bug if we only applied the taskmgr changes, but would leave NTUSER unchanged?

Cleverking2003:
> Why is that coupled with the NTUSER changes? What would be the symptom/bug if we only applied the taskmgr changes, but would leave NTUSER unchanged?

Without the NTUSER changes, the task manager would stiil act as if we didn't apply the fix at all (though the fix works on Windows on its own)

SergeGautherie (commented):
> the fix works on Windows on its own

I concur with Joachim: splitting this PR in two separate ones should ease review process.

win32ss/user/ntuser/menu.c
```diff
@@ -4587,7 +4587,7 @@ BOOL WINAPI IntTrackPopupMenuEx( PMENU menu, UINT wFlags, int x, int y,
        /* Send WM_INITMENUPOPUP message only if TPM_NONOTIFY flag is not specified */
        if (!(wFlags & TPM_NONOTIFY))
        {
-          co_IntSendMessage( UserHMGetHandle(pWnd), WM_INITMENUPOPUP, (WPARAM) UserHMGetHandle(menu), 0);
+          co_IntSendMessage( UserHMGetHandle(pWnd), WM_INITMENUPOPUP, (WPARAM) UserHMGetHandle(menu), MAKELPARAM(0, IS_SYSTEM_MENU(menu)));
```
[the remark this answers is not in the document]
Cleverking2003: No, this one should actually return 0 or 1

win32ss/user/ntuser/menu.c
```diff
@@ -58,7 +58,7 @@ BOOL fInEndMenu = FALSE;
 #define MII_STATE_MASK (MFS_GRAYED|MFS_CHECKED|MFS_HILITE|MFS_DEFAULT)
 
 #define IS_SYSTEM_MENU(MenuInfo)  \
```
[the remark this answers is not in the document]
Cleverking2003: Yes, I think we could remove the unused macros

base/applications/taskmgr/taskmgr.c
```diff
@@ -482,14 +482,19 @@ TaskManagerWndProc(HWND hDlg, UINT message, WPARAM wParam, LPARAM lParam)
...
         break;
 
-    case WM_ENTERMENULOOP:
-        TaskManager_OnEnterMenuLoop(hDlg);
+    case WM_INITMENUPOPUP:
+        /* Do not disable the status bar if we opened the system menu */
+        if (!HIWORD(lParam))
+            TaskManager_DisableStatusbar(hDlg);
+        else
+            TaskManager_EnableStatusbar(hDlg);
```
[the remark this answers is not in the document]
Cleverking2003: I think there was some edge case which broke without this `else`, need to recheck that

JoachimHenze:
> I concur with Joachim: splitting this PR in two separate ones should ease review process.

I just requested for an explanation, which was given by the author.
I **didn't** explicitly request for splitting it up. Ftr it is not my position to split them up.

base/applications/taskmgr/taskmgr.c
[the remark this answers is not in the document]
Cleverking2003: Without this `else` clause, if you open the menu and then click on the app icon to open the system menu, the status bar won't be restored.

base/applications/taskmgr/taskmgr.c
```diff
@@ -482,14 +482,19 @@ TaskManagerWndProc(HWND hDlg, UINT message, WPARAM wParam, LPARAM lParam)
...
         break;
     case WM_EXITMENULOOP:
-        TaskManager_OnExitMenuLoop(hDlg);
+        TaskManager_EnableStatusbar(hDlg);
         break;
     case WM_MENUSELECT:
-        TaskManager_OnMenuSelect(hDlg, LOWORD(wParam), HIWORD(wParam), (HMENU)lParam);
+        if (!(HIWORD(wParam) & MF_SYSMENU))
+            TaskManager_OnMenuSelect(hDlg, LOWORD(wParam), HIWORD(wParam), (HMENU)lParam);
```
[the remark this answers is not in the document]
Cleverking2003: I don't know what's the purpose of this change other than adding a second `break`

win32ss/user/ntuser/menu.c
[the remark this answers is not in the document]
Cleverking2003: Done

base/applications/taskmgr/taskmgr.c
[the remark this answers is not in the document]
SergeGautherie: Ok, maybe add a comment about that.
I know little about UI, so this code looks to me more linked to sysmenu opening than to appmenu closing which I would expect. All right if that is how it should be handled.

Cleverking2003:
> Did you check other places for similar bug in the code base : https://git.reactos.org/?p=reactos.git&a=search&st=grep&s=WM_INITMENUPOPUP ? clipbrd, mscutils, mspaint, notepad, winhlp32, wordpad, ...

I checked these apps and they don't seem to depend on that behaviour.

win32ss/user/ntuser/menu.c
```diff
@@ -4617,7 +4617,7 @@ BOOL WINAPI IntTrackPopupMenuEx( PMENU menu, UINT wFlags, int x, int y,
           if (!(wFlags & TPM_NONOTIFY))
           {
              co_IntSendMessage( UserHMGetHandle(pWnd), WM_UNINITMENUPOPUP, (WPARAM)UserHMGetHandle(menu),
-                                            MAKELPARAM(0, IS_SYSTEM_MENU(menu)) );
+                                            MAKELPARAM(0, IS_SYSTEM_MENU(menu) ? MF_SYSMENU : 0) );
```
HBelusca, line 4620: Question: Are those modifications for `WM_UNINITMENUPOPUP` are for implementing the comment made in
https://learn.microsoft.com/en-us/windows/win32/menurc/wm-uninitmenupopup ? Namely:
> lParam
>
> The high-order word identifies the menu that was destroyed. Currently, this parameter can only be MF_SYSMENU (the window menu).

base/applications/taskmgr/taskmgr.c
```diff
@@ -898,7 +903,7 @@ void TaskManager_OnEnterMenuLoop(HWND hWnd)
     SendMessageW(hStatusWnd, SB_SETTEXT, (WPARAM)0, (LPARAM)L"");
 }
 
-void TaskManager_OnExitMenuLoop(HWND hWnd)
+void TaskManager_EnableStatusbar(HWND hWnd)
```
HBelusca, line 906:
```suggestion
void TaskManager_EnableStatusBar(HWND hWnd)
```
and similar change for the Disable function above.

HBelusca (commented):
I would really like to see a separate PR from yours, for just the ntuser/menu.c changes, or, in any case, add a small apitest (adapted probably from an existing one or from a winetest we have) to test those changes.

win32ss/user/ntuser/menu.c
Cleverking2003: Yes, they are implemeting this comment

Cleverking2003:
Now this PR contains only the taskmgr fix, the ntuser fix is in #5573

base/applications/taskmgr/taskmgr.c
```diff
@@ -482,14 +482,19 @@ TaskManagerWndProc(HWND hDlg, UINT message, WPARAM wParam, LPARAM lParam)
...
         break;
 
-    case WM_ENTERMENULOOP:
-        TaskManager_OnEnterMenuLoop(hDlg);
+    case WM_INITMENUPOPUP:
+        /* Do not disable the status bar if we opened the system menu */
+        if (!HIWORD(lParam))
+            TaskManager_DisableStatusBar(hDlg);
+        else
+            TaskManager_EnableStatusBar(hDlg);
```
binarymaster, line 490: I think you need to preserve `WM_ENTERMENULOOP` handler and move `bInMenuLoop` variable assignment there.

```suggestion
            TaskManager_EnableStatusBar(hDlg);
        break;
    case WM_ENTERMENULOOP:
        bInMenuLoop = TRUE;
        break;
```

base/applications/taskmgr/taskmgr.c
```diff
@@ -482,14 +482,19 @@ TaskManagerWndProc(HWND hDlg, UINT message, WPARAM wParam, LPARAM lParam)
...
         break;
     case WM_EXITMENULOOP:
-        TaskManager_OnExitMenuLoop(hDlg);
+        TaskManager_EnableStatusBar(hDlg);
```
binarymaster, line 493:
```suggestion
        bInMenuLoop = FALSE;
        TaskManager_EnableStatusBar(hDlg);
```

binarymaster (commented):
By applying code suggestions you can remove `bInMenuLoop` assignments inside the renamed functions.

base/applications/taskmgr/taskmgr.c
```diff
@@ -482,14 +482,23 @@ TaskManagerWndProc(HWND hDlg, UINT message, WPARAM wParam, LPARAM lParam)
...
+            TaskManager_EnableStatusBar(hDlg);
+        break;
     case WM_ENTERMENULOOP:
-        TaskManager_OnEnterMenuLoop(hDlg);
+        bInMenuLoop = TRUE;
         break;
     case WM_EXITMENULOOP:
-        TaskManager_OnExitMenuLoop(hDlg);
+        TaskManager_EnableStatusBar(hDlg);
+        bInMenuLoop = FALSE;
```
binarymaster, line 497: I think `bInMenuLoop = FALSE;` should be done before calling `TaskManager_EnableStatusBar()`, because the latter invokes `RefreshPerformancePage()`, which reads the value of that variable and behaves differently depending on it.

binarymaster (approved):

base/applications/taskmgr/taskmgr.c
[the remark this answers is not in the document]
SergeGautherie: > this code looks to me more linked to sysmenu opening than to appmenu closing which I would expect.

Shouldn't this live in `WM_UNINITMENUPOPUP` rather?
Would that allow to simplify `WM_EXITMENULOOP` too?

base/applications/taskmgr/taskmgr.h
```diff
@@ -94,8 +94,8 @@ void FillSolidRect(HDC hDC, LPCRECT lpRect, COLORREF clr);
 void LoadSettings(void);
 void SaveSettings(void);
 void TaskManager_OnRestoreMainWindow(void);
-void TaskManager_OnEnterMenuLoop(HWND hWnd);
-void TaskManager_OnExitMenuLoop(HWND hWnd);
+void TaskManager_DisableStatusBar(HWND hWnd);
+void TaskManager_EnableStatusBar(HWND hWnd);
```
SergeGautherie, line 98: Move these 1 line up, to keep them somewhat sorted.

base/applications/taskmgr/taskmgr.c
```diff
@@ -482,14 +482,23 @@ TaskManagerWndProc(HWND hDlg, UINT message, WPARAM wParam, LPARAM lParam)
         TrayIcon_UpdateIcon();
         break;
 
+    case WM_INITMENUPOPUP:
+        /* Do not disable the status bar if we opened the system menu */
+        if (!HIWORD(lParam))
+            TaskManager_DisableStatusBar(hDlg);
+        else
+            TaskManager_EnableStatusBar(hDlg);
```
HBelusca, line 490: Shouldn't the re-enabling be done in an `WM_UNINITMENUPOPUP`? Or maybe the call done in `WM_EXITMENULOOP` is sufficient?

Cleverking2003: WM_EXITMENULOOP is sufficient

Cleverking2003: If you're talking about the `else` clause then see one of the comments above

Cleverking2003: Yes, it does happen on Windows too

HBelusca: OK actually this also handles "smoothly" the transition between the "user" menu and the system one (and back).

HBelusca (approved):

PR #5576, opened 2023-08-16
[KERNEL32] Fix startup context of initial thread

tkreuzer:
When creating a new process the monster function CreateProcessInternalW calls BaseInitializeContext to set up the context for the initial thread. One of the parameters is the PEB pointer. Instead of passing the pointer to the newly created process' PEB, the function was passing its own PEB address, leading to test failures. How this is not an issue otherwise is a mystery to be resolved by someone else. Also remove ros_skip_flaky marks in kernel32_winetest:process (yes, flaky, because sometimes the PEB addresses actually match)

Test KVM: https://reactos.org/testman/compare.php?ids=88924,88926,88931,88934,88939

### DIFF
--- a/dll/win32/kernel32/client/proc.c
+++ b/dll/win32/kernel32/client/proc.c
@@ -4181,7 +4181,7 @@ StartScan:
 
     /* Create the Thread's Context */
     BaseInitializeContext(&Context,
-                          Peb,
+                          RemotePeb,
                           ImageInformation.TransferAddress,
                           InitialTeb.StackBase,
                           0);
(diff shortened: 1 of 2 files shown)


HBelusca (approved):

HBelusca:
Nice catch!

binarymaster (approved):

dll/win32/kernel32/client/proc.c
```diff
@@ -4181,7 +4181,7 @@ CreateProcessInternalW(IN HANDLE hUserToken,
 
     /* Create the Thread's Context */
     BaseInitializeContext(&Context,
-                          Peb,
+                          RemotePeb,
```
SergeGautherie, line 4184: While here, it looks like [line 3558](https://github.com/reactos/reactos/blob/bdf7e0a990722cee28b106426fc82970c5e0848f/dll/win32/kernel32/client/proc.c#L3558) could/should use `Peb`, for trivial consistency/optimization.

PR #5577, opened 2023-08-16
[NTOS:KE/x64] Fix bug in KiInitializeContextThread

tkreuzer:
The function set CtxSwitchFrame->ApcBypass to FALSE, preventing APCs (like when user mode sets the context while the thread is suspended) from being delivered as soon as the thread lowers IRQL to PASSIVE_LEVEL. This resulted in the SetContext APC to be delivered only after the user mode APC was initialized, overwriting the user mode APC context in the trap frame. This caused kernel32_winetest process to break.

### DIFF
--- a/ntoskrnl/ke/amd64/thrdini.c
+++ b/ntoskrnl/ke/amd64/thrdini.c
@@ -134,7 +134,7 @@ KiInitializeContextThread(IN PKTHREAD Thread,
 
     /* Set up the Context Switch Frame */
     CtxSwitchFrame->Return = (ULONG64)KiThreadStartup;
-    CtxSwitchFrame->ApcBypass = FALSE;
+    CtxSwitchFrame->ApcBypass = TRUE;
 
     StartFrame->P1Home = (ULONG64)StartRoutine;
     StartFrame->P2Home = (ULONG64)StartContext;


ntoskrnl/ke/amd64/thrdini.c
```diff
@@ -134,7 +134,7 @@ KiInitializeContextThread(IN PKTHREAD Thread,
 
     /* Set up the Context Switch Frame */
     CtxSwitchFrame->Return = (ULONG64)KiThreadStartup;
-    CtxSwitchFrame->ApcBypass = FALSE;
+    CtxSwitchFrame->ApcBypass = TRUE;
```
HBelusca, line 137: The x64 ApcBypass is not a BOOLEAN, but a KIRQL:
https://git.reactos.org/?p=reactos.git;a=blob;f=sdk/include/ndk/amd64/ketypes.h;hb=5645f7cf8f5b9f41de782535081983db0a7cfb35#l1054
so it should be:
```suggestion
    CtxSwitchFrame->ApcBypass = APC_LEVEL;
```

binarymaster (commented):
Suggesting `[NTOS:KE/x64]` prefix, and Hermes' review comment makes sense.

PR #5582, opened 2023-08-17
[CONSRV] Only call CoUninitialize if initialized

whindsaks:
Only call CoUninitialize on SUCCEEDED

### DIFF
--- a/win32ss/user/winsrv/consrv/console.c
+++ b/win32ss/user/winsrv/consrv/console.c
@@ -3,7 +3,7 @@
  * PROJECT:         ReactOS Console Server DLL
  * FILE:            win32ss/user/winsrv/consrv/console.c
  * PURPOSE:         Console Management Functions
- * PROGRAMMERS:     Gé van Geldorp
+ * PROGRAMMERS:     GÃ© van Geldorp
  *                  Jeffrey Morlan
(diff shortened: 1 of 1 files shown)


SergeGautherie (approved):

win32ss/user/winsrv/consrv/console.c
```diff
@@ -3,7 +3,7 @@
  * PROJECT:         ReactOS Console Server DLL
  * FILE:            win32ss/user/winsrv/consrv/console.c
  * PURPOSE:         Console Management Functions
- * PROGRAMMERS:     G� van Geldorp
+ * PROGRAMMERS:     Gé van Geldorp
```
JoachimHenze, line 6: We do have an encoding change here. Is it to the good (towards UTF8 wo BOM)? Or to the bad (away from UTF8woBOM)?

SergeGautherie: Good, according to Hermès's today commit.

whindsaks: Github did that, from 8...something to UTF-8

HBelusca (approved):
I confirm that the original file was ANSI-encoded.

PR #5584, opened 2023-08-18
[BROWSEUI] More settings and window refresh

cbialorucki:
## Purpose

image

Adds the option to set the address edit box to use the display name or the full path. Also refreshes the window title and edit box in all open explorer windows when changing these settings using the folder options dialog.

## Proposed changes
- Create a new `CabinetStateSettings` type that inherits from the `CABINETSTATE` type. This allows us to add additional cabinet state settings not exposed in the `CABINETSTATE` type as well as adding a `Load()` method to easily populate the cabinet state settings.
- Add a global cabinet state settings object. While most settings in browseui are stored independently in each shellbrowser window, cabinet state settings are global and apply to every shellbrowser window. This can be confirmed on Windows Server 2003 and Windows 7.
- When receiving the `WM_SETTINGCHANGE` window message from the folder options dialog, refresh the title of the window and the text in the address edit box. This is the same behavior as Windows Server 2003 and Windows 7.
- Add a DWORD registry value to `HKCU\...\Explorer\CabinetState\FullPathAddress` to allow users to toggle this setting on or off in our folder options.

### DIFF
--- a/dll/win32/browseui/CMakeLists.txt
+++ b/dll/win32/browseui/CMakeLists.txt
@@ -30,6 +30,7 @@ list(APPEND SOURCE
     internettoolbar.cpp
     parsecmdline.cpp
     regtreeoptions.cpp
+    settings_cabinet.cpp
     settings.cpp
     shellbrowser.cpp
     toolsband.cpp
--- a/dll/win32/browseui/addresseditbox.cpp
+++ b/dll/win32/browseui/addresseditbox.cpp
@@ -130,17 +130,22 @@ HRESULT CAddressEditBox::RefreshAddress()
 
     /* Set the path if filesystem; otherwise use the name */
     WCHAR szPathOrName[MAX_PATH];
-    if (!SHGetPathFromIDListW(absolutePIDL, szPathOrName))
-    {
-        STRRET ret;
-        hr = pShellFolder->GetDisplayNameOf(pidlChild, SHGDN_FORADDRESSBAR | SHGDN_FORPARSING, &ret);
-        if (FAILED_UNEXPECTEDLY(hr))
-            return hr;
+    SHGDNF flags;
 
-        hr = StrRetToBufW(&ret, pidlChild, szPathOrName, _countof(szPathOrName));
-        if (FAILED_UNEXPECTEDLY(hr))
-            return hr;
-    }
+    if (gCabinetState.fFullPathAddress)
+        flags = SHGDN_FORADDRESSBAR | SHGDN_FORPARSING;
+    else
+        flags = SHGDN_FORADDRESSBAR;
+
+    STRRET ret;
+    hr = pShellFolder->GetDisplayNameOf(pidlChild, flags, &ret);
+    if (FAILED_UNEXPECTEDLY(hr))
+        return hr;
+
+    hr = StrRetToBufW(&ret, pidlChild, szPathOrName, _countof(szPathOrName));
+    if (FAILED_UNEXPECTEDLY(hr))
+        return hr;
+
     item.pszText = szPathOrName;
 
     /* Ownership of absolutePIDL will be moved to fCombobox. See CBEN_DELETEITEM */
@@ -643,4 +648,10 @@ LPITEMIDLIST CAddressEditBox::GetItemData(int index)
     item.iItem = index;
     SendMessageW(hComboBoxEx, CBEM_GETITEMW, 0, (LPARAM)&item);
     return (LPITEMIDLIST)item.lParam;
+}
+
+LRESULT CAddressEditBox::OnSettingChange(UINT uMsg, WPARAM wParam, LPARAM lParam, BOOL &bHandled)
+{
+    RefreshAddress();
+    return NO_ERROR;
 }
--- a/dll/win32/browseui/precomp.h
+++ b/dll/win32/browseui/precomp.h
@@ -81,7 +81,6 @@ struct ShellSettings
 
 struct CabinetStateSettings : CABINETSTATE
 {
-    /* Additional CabinetState settings */
     BOOL fFullPathAddress = TRUE;
 
     BOOL Load();
--- a/dll/win32/browseui/precomp.h
+++ b/dll/win32/browseui/precomp.h
@@ -79,4 +79,14 @@ struct ShellSettings
     BOOL Load();
 };
 
+struct CabinetStateSettings : CABINETSTATE
+{
+    /* Additional CabinetState settings */
+    BOOL fFullPathAddress = TRUE;
+
+    BOOL Load();
+};
+
+extern CabinetStateSettings gCabinetState;
+
 #endif /* _BROWSEUI_PCH_ */
--- a/dll/win32/browseui/settings_cabinet.cpp
+++ b/dll/win32/browseui/settings_cabinet.cpp
@@ -1,0 +1,22 @@
+/*
+ * PROJECT:     ReactOS browseui
+ * LICENSE:     LGPL-2.1-or-later (https://spdx.org/licenses/LGPL-2.1-or-later)
+ * PURPOSE:     Stores the global cabinet state for the file explorer UI.
+ * COPYRIGHT:   Copyright 2023 Carl Bialorucki <cbialo2@outlook.com>
+ */
+
+#include "precomp.h"
+
+CabinetStateSettings gCabinetState;
+
+BOOL CabinetStateSettings::Load()
+{
+    /* Ideally we would use ReadCabinetState, but unfortunately Wine's implementation is incomplete. */
+    fFullPathTitle = SHRegGetBoolUSValueW(L"Software\\Microsoft\\Windows\\CurrentVersion\\Explorer\\CabinetState",
+                                          L"FullPath", FALSE, FALSE);
+
+    fFullPathAddress = SHRegGetBoolUSValueW(L"Software\\Microsoft\\Windows\\CurrentVersion\\Explorer\\CabinetState",
+                                            L"FullPathAddress", FALSE, TRUE);
+
+    return TRUE;
+}
--- a/dll/win32/browseui/shellbrowser.cpp
+++ b/dll/win32/browseui/shellbrowser.cpp
@@ -296,7 +296,6 @@ private:
     barInfo                                 fClientBars[3];
     CComPtr<ITravelLog>                     fTravelLog;
     HMENU                                   fCurrentMenuBar;
-    CABINETSTATE                            fCabinetState;
     GUID                                    fCurrentVertBar;             //The guid of the built in vertical bar that is being shown
     // The next three fields support persisted history for shell views.
     // They do not need to be reference counted.
@@ -350,7 +349,8 @@ public:
     HRESULT UpdateUpState();
     void UpdateGotoMenu(HMENU theMenu);
     void UpdateViewMenu(HMENU theMenu);
-    void LoadCabinetState();
+    void RefreshCabinetState();
+    void UpdateWindowTitle();
 
 /*    // *** IDockingWindowFrame methods ***
     virtual HRESULT STDMETHODCALLTYPE AddToolbar(IUnknown *punkSrc, LPCWSTR pwszItem, DWORD dwAddFlags);
@@ -719,6 +719,7 @@ CShellBrowser::CShellBrowser()
     fHistoryStream = NULL;
     fHistoryBindContext = NULL;
     m_settings.Load();
+    gCabinetState.Load();
 }
 
 CShellBrowser::~CShellBrowser()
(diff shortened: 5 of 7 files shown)


dll/win32/browseui/addresseditbox.cpp
```diff
@@ -644,3 +649,9 @@ LPITEMIDLIST CAddressEditBox::GetItemData(int index)
     SendMessageW(hComboBoxEx, CBEM_GETITEMW, 0, (LPARAM)&item);
     return (LPITEMIDLIST)item.lParam;
 }
+
+LRESULT CAddressEditBox::OnSettingChange(UINT uMsg, WPARAM wParam, LPARAM lParam, BOOL &bHandled)
+{
+    RefreshAddress();
+    return NO_ERROR;
+}
```
SergeGautherie, line 657:
```suggestion
}

```

dll/win32/browseui/settings_cabinet.cpp
```diff
@@ -0,0 +1,22 @@
...
+{
+    /* Ideally we would use ReadCabinetState, but unfortunately Wine's implementation is incomplete. */
+    fFullPathTitle = SHRegGetBoolUSValueW(L"Software\\Microsoft\\Windows\\CurrentVersion\\Explorer\\CabinetState",
+                                          L"FullPath", FALSE, FALSE);
+
+    fFullPathAddress = SHRegGetBoolUSValueW(L"Software\\Microsoft\\Windows\\CurrentVersion\\Explorer\\CabinetState",
+                                            L"FullPathAddress", FALSE, TRUE);
+
+    return TRUE;
+}
```
SergeGautherie, line 22:
```suggestion
}

```

dll/win32/browseui/shellbrowser.cpp
```diff
@@ -3816,3 +3794,23 @@ HRESULT CShellBrowser_CreateInstance(REFIID riid, void **ppv)
...
+    SHGDNF flags;
+
+    if (gCabinetState.fFullPathTitle)
+        flags = SHGDN_FORADDRESSBAR | SHGDN_FORPARSING;
+    else
+        flags = SHGDN_FORADDRESSBAR;
+
+    if (SUCCEEDED(IEGetNameAndFlags(fCurrentDirectoryPIDL, flags, title, MAX_PATH, NULL)))
+        SetWindowText(title);
+}
```
SergeGautherie, line 3816:
```suggestion
}

```

dll/win32/browseui/CMakeLists.txt
```diff
@@ -30,6 +30,7 @@ list(APPEND SOURCE
     internettoolbar.cpp
     parsecmdline.cpp
     regtreeoptions.cpp
+    settings_cabinet.cpp
     settings.cpp
```
HBelusca, line 34:
```suggestion
    settings.cpp
    settings_cabinet.cpp
```

dll/win32/browseui/precomp.h
```diff
@@ -79,4 +79,13 @@ struct ShellSettings
     BOOL Load();
 };
 
+struct CabinetStateSettings : CABINETSTATE
+{
+    BOOL fFullPathAddress = TRUE;
+
+    BOOL Load();
+};
```
HBelusca, line 87: At some point it could be cleaner to have all those settings structures definitions in their own header file.

dll/win32/browseui/CMakeLists.txt
cbialorucki: Do you think it would be better to move the cabinet state settings method implementation to `settings.cpp`?

HBelusca: yes probably

cbialorucki: Done 👍

dll/win32/browseui/precomp.h
cbialorucki: Added a header file for settings.

whindsaks:
Why extend CABINETSTATE? There are plenty of free bits in `fUnusedFlags : 7` (Microsoft has not appended to this struct in 25 years, the last bit is probably safe to use)

You should still call `ReadCabinetState` when reading the settings because the single values are overrides for the struct:

````C++
ReadCabinetState(&cs);
DWORD value, error, defval;
if (ERROR_SUCCESS == SHGetValue(..., "FullPath", &value)) cs.fFullPathTitle = value != 0;
error = SHGetValue(..., "FullPathAddress", &value), defval = TRUE;
cs.newBitYouInventHere = error ? defval != 0 : value != 0;
````

Updating hivedef.inf should not be required (IMHO), the reader code should use the correct default.

**Comment by cbialorucki** [comment]
>Updating hivedef.inf should not be required (IMHO), the reader code should use the correct default.

That's how our folder options dialog is designed. If the registry value cannot be opened in HKCU the setting is disabled in the UI. @katahiromz designed it from what I understand, and you can confirm with him.

>You should still call `ReadCabinetState` when reading the settings because the single values are overrides for the struct.

Ok. Done.

>Why [not] extend CABINETSTATE?

This is an extension of `CABINETSTATE`; the new `CabinetStateSettings` type inherits `CABINETSTATE`. Using those extra flags is not Window's behavior. Adding this variable is far easier to work with than doing some magic on an unused field.

**Comment by whindsaks** [comment]
> That's how our folder options dialog is designed.

But that is not how Windows works. Even if a user deletes HKCU\Software\Microsoft\Windows\Explorer, it is not supposed to disable anything, it just reverts everything to their default values. These types of settings should be disabled by a value in the Policy key if disabling them has any purpose at all.

**Comment by cbialorucki** [comment]
> > That's how our folder options dialog is designed.
> 
> But that is not how Windows works. 

Correct. Unfortunately this PR isn't about refactoring Shell32.

**Comment by cbialorucki** [comment]
> There are plenty of free bits in `fUnusedFlags : 7` (Microsoft has not appended to this struct in 25 years, the last bit is probably safe to use)

Out of curiosity I did write a program to see if the unused flags changed when setting the address bar to use a full path or not. The unused flags did not change.
image

**Comment by HBelusca** [comment]
> Add a DWORD registry value to HKCU\...\Explorer\CabinetState\FullPathAddress to allow users to toggle this setting on or off in our folder options.

Is it the actual value Windows shell uses for that? (This wasn't fully clear from the comments above...)

**Comment by cbialorucki** [comment]
> > Add a DWORD registry value to HKCU...\Explorer\CabinetState\FullPathAddress to allow users to toggle this setting on or off in our folder options.
> 
> Is it the actual value Windows shell uses for that? (This wasn't fully clear from the comments above...)

This is the actual value used on Windows Server 2003.

File: dll/win32/browseui/addresseditbox.cpp [hunk 6]
```diff
@@ -130,17 +130,22 @@ HRESULT CAddressEditBox::RefreshAddress()
...
-            return hr;
+    SHGDNF flags;
+
+    if (gCabinetState.fFullPathAddress)
+        flags = SHGDN_FORADDRESSBAR | SHGDN_FORPARSING;
+    else
+        flags = SHGDN_FORADDRESSBAR;
+
+    STRRET ret;
+    hr = pShellFolder->GetDisplayNameOf(pidlChild, flags, &ret);
```
HBelusca, line 141: This kinds of look similar to IEGetNameAndFlags but also a bit different. Do you know whether using that other function would make the code behave differently?

HBelusca (approved):
LGTM, but:
> Add a DWORD registry value to HKCU\...\Explorer\CabinetState\FullPathAddress to allow users to toggle this setting on or off in our folder options.

Is this what @whindsaks was talking about earlier, namely that on Windows, the explicit presence or absence of such value in the registry does not lead to the disabling of the associated item in the folders options?

cbialorucki:
> LGTM, but:
> 
> > Add a DWORD registry value to HKCU...\Explorer\CabinetState\FullPathAddress to allow users to toggle this setting on or off in our folder options.
> 
> Is this what @whindsaks was talking about earlier, namely that on Windows, the explicit presence or absence of such value in the registry does not lead to the disabling of the associated item in the folders options?

@whindsaks is correct that Windows does not disable the option in folder options if the registry value is missing, ***but ROS 
 does***. I presume this was done because most settings in folder options still do not work and we only wanted to enable settings that actually do something. I pointed to the offending code in our shell32.
 
All folder option settings can be found in `HKLM\Software\Microsoft\Windows\CurrentVersion\Explorer\Advanced\Folder`. When showing the folder options dialog, the dialog opens each registry key under `HKLM\..\Advanced\Folder` then checks if the value it references can be queried. If it can, it is enabled. In our shell32 we also check for a ROS-specific "Grayed" registry value. If it is enabled, it disables the corresponding folder options setting in the dialog.

We have a couple options for fixing this, but I felt that it was outside the scope of this PR because it requires a lot of discussion. I feel confident that I can implement any of these fixes.

1. Stop checking if the registry key can be queried and enable these settings by default. Only disable these settings if the ROS-specific 'Grayed' registry value is present.
- Pros: Straightforward implementation, disables unavailable settings to users.
- Cons: Additional registry bloat. We must add the "Grayed" registry value to every setting that still does not work (which is most of them.) As more settings are added we have to remove the "Grayed" registry value from `hivedef.inf` to enable the setting in folder options. To remove these keys requires a clean install of the OS with our current updater. So, you will have to clean install the OS to test these features.

2. Enable all folder option settings by default anyways and don't query the ROS-specific "Grayed" registry value.
- Pros: Even more straightforward implementation. Easier to add new features and does not require a clean install to enable new folder option features. We can also remove some file explorer registry bloat from our `hivedef.inf` that were put in because of the old behavior.
- Cons: Users may be confused that some folder options settings do not work.

As far as I know, group policy on Windows allows administrators to disable access to the entire folder options dialog but does not allow individual folder option settings to be disabled. If that is true, there is no Windows-specific way to disable individual folder option settings outside of blocking the entire dialog.

I am in favor of fixing this using option 2.

whindsaks:
> As far as I know, group policy on Windows allows administrators to disable access to the entire folder options dialog but does not allow individual folder option settings to be disabled.

Incorrect, Windows checks a policy key for each option. I will PR this later.

cbialorucki:
Since @whindsaks is working on fixing shell32 and has found the registry values required to disable and enable individual folder option settings, should I remove the extra registry key?

@whindsaks - When submitting your PR make sure to review `hivedef.inf` to remove any unneeded registry keys/values that were added because of the old behavior.

dll/win32/browseui/settings.cpp
```diff
@@ -34,3 +36,17 @@ BOOL ShellSettings::Load()
...
+    ReadCabinetState(this, this->cLength);
+
+    /* Overrides */
+    fFullPathTitle = SHRegGetBoolUSValueW(L"Software\\Microsoft\\Windows\\CurrentVersion\\Explorer\\CabinetState",
+                                          L"FullPath", FALSE, FALSE);
+
+    fFullPathAddress = SHRegGetBoolUSValueW(L"Software\\Microsoft\\Windows\\CurrentVersion\\Explorer\\CabinetState",
+                                            L"FullPathAddress", FALSE, TRUE);
+
+    return TRUE;
```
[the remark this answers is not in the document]
cbialorucki: I converted these functions to `void`s.

dll/win32/browseui/addresseditbox.cpp
cbialorucki: `IEGetNameAndFlags` looks like a wrapper for `GetDisplayNameOf`. I can expose the `IEGetNameAndFlags` function through a header and use it here instead to avoid code duplication.

whindsaks:
> should I remove the added registry value?

No, just leave it.


> When submitting your PR make sure to review `hivedef.inf` to remove any unneeded registry keys/values that were added because of the old behavior.

Sure.

dll/win32/browseui/settings.h
```diff
@@ -0,0 +1,29 @@
+/*
+ * PROJECT:     ReactOS browseui
+ * LICENSE:     LGPL-2.1-or-later (https://spdx.org/licenses/LGPL-2.1-or-later)
+ * PURPOSE:     Settings header file
+ * COPYRIGHT:   Copyright 2023 Carl Bialorucki <cbialo2@outlook.com>
+ */
+#include "precomp.h"
```
binarymaster, line 7: Always keep an empty line after the header:
```suggestion

#include "precomp.h"
```

dll/win32/browseui/browseui.h
```diff
@@ -24,3 +24,5 @@ HRESULT CTravelLog_CreateInstance(REFIID riid, void **ppv);
 HRESULT CBaseBar_CreateInstance(REFIID riid, void **ppv, BOOL vertical);
 HRESULT CBaseBarSite_CreateInstance(REFIID riid, void **ppv, BOOL bVertical);
 HRESULT CToolsBand_CreateInstance(REFIID riid, void **ppv);
+
+long IEGetNameAndFlags(LPITEMIDLIST pidl, SHGDNF uFlags, LPWSTR pszBuf, UINT cchBuf, SFGAOF *rgfInOut);
```
binarymaster, line 28: No newline at end of file. Also this function return type should be HRESULT.

cbialorucki: Fixed :)

binarymaster (approved):

HBelusca (approved):
Waiting for @whindsaks to make a separate PR regarding the shell settings and policy stuff :)

PR #5586, opened 2023-08-18
[KMTEST:TcpIp] Use 10 second timeout value instead of INFINITE

tkreuzer:
On x64 the test sometimes fails to connect and then times out on the testbot, causing the system to be rebooted.

### DIFF
--- a/modules/rostests/kmtests/tcpip/TcpIp_user.c
+++ b/modules/rostests/kmtests/tcpip/TcpIp_user.c
@@ -113,7 +113,7 @@ START_TEST(TcpIpConnect)
     Error = KmtSendToDriver(IOCTL_TEST_CONNECT);
     ok_eq_ulong(Error, ERROR_SUCCESS);
 
-    WaitForSingleObject(AcceptThread, INFINITE);
+    WaitForSingleObject(AcceptThread, 10 * 1000);
 
     UnloadTcpIpTestDriver();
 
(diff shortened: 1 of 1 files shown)


modules/rostests/kmtests/tcpip/TcpIp_user.c
```diff
@@ -113,7 +113,7 @@ START_TEST(TcpIpConnect)
     Error = KmtSendToDriver(IOCTL_TEST_CONNECT);
     ok_eq_ulong(Error, ERROR_SUCCESS);
 
-    WaitForSingleObject(AcceptThread, INFINITE);
+    WaitForSingleObject(AcceptThread, 10 * 1000);
```
binarymaster, line 116: Maybe add a test check whether it timed out:
```suggestion
    Error = WaitForSingleObject(AcceptThread, 10 * 1000);
    ok(Error == WAIT_OBJECT_0, "AcceptThread timed out\n");
```

binarymaster (approved):
LGTM + a small idea in review comment.

JoachimHenze:
If it fails with the connect in the test for x64, but succeeds for x86, then we can be pretty sure that this is a hack.

HBelusca:
I somewhat concur with @JoachimHenze : do we know the underlying reason why the test sometimes fails to "connect"? (connect to what, by the way?)

SergeGautherie:
> On x64 the test sometimes fails to connect and then times out on the testbot

ReactOS and/or Windows? Which testbot(s) (runs)?

> this is a hack.

Suggestions: `#ifdef _WIN64` + `HACK` comment with a Jira ticket.

tkreuzer:
This is a test. It's entire purpose is to test whether something works. It cannot test, whether it works, if it never finishes, when it doesn't work. The test has a timeout anyway: when the testbot hard resets the machine after 5 minutes or so, but that is not how it's supposed to be.
When the connect takes more than 10 seconds, you can safely assume that something is wrong and the test should fail.
So this change is not a hack, it's an improvement.

ThFabba (approved):

learn-more (approved):

PR #5587, opened 2023-08-18
[SHELL32] Don't display missing applications in Open with. CORE-19117

whindsaks:
## Purpose

Don't display applications that don't exist.

JIRA issue: [CORE-19117](https://jira.reactos.org/browse/CORE-19117)

I verified the behavior on XP. Exe files that no longer exist are hidden but their registry entry is not deleted.

### DIFF
--- a/dll/win32/shell32/COpenWithMenu.cpp
+++ b/dll/win32/shell32/COpenWithMenu.cpp
@@ -240,12 +240,12 @@ HICON COpenWithList::GetIcon(SApp *pApp)
         if (!ExtractIconExW(wszPath, 0, NULL, &pApp->hIcon, 1))
         {
             SHFILEINFO fi;
-            /* FIXME: Ideally we should pass SHGFI_ICON|SHGFI_USEFILEATTRIBUTES because
+            /* FIXME: Ideally we should include SHGFI_USEFILEATTRIBUTES because
             ** we already know the file has no icons but SHGetFileInfo is broken in that case.
             ** Without SHGFI_USEFILEATTRIBUTES we needlessly hit the disk again but it will
             ** return the correct default .exe icon.
             */
-            SHGetFileInfoW(wszPath, 0, &fi, sizeof(fi), SHGFI_ICON);
+            SHGetFileInfoW(wszPath, 0, &fi, sizeof(fi), SHGFI_ICON|SHGFI_SMALLICON|SHGFI_SHELLICONSIZE);
             pApp->hIcon = fi.hIcon;
         }
     }
--- a/dll/win32/shell32/COpenWithMenu.cpp
+++ b/dll/win32/shell32/COpenWithMenu.cpp
@@ -78,7 +78,7 @@ class COpenWithList
 
         SApp *AddInternal(LPCWSTR pwszFilename);
         static BOOL LoadInfo(SApp *pApp);
-        static VOID GetPathFromCmd(LPWSTR pwszAppPath, LPCWSTR pwszCmd);
+        static BOOL GetPathFromCmd(LPWSTR pwszAppPath, LPCWSTR pwszCmd);
         BOOL LoadProgIdList(HKEY hKey, LPCWSTR pwszExt);
         static HANDLE OpenMRUList(HKEY hKey);
         BOOL LoadMRUList(HKEY hKey);
@@ -217,7 +217,12 @@ LPCWSTR COpenWithList::GetName(SApp *pApp)
         {
             WARN("Failed to load %ls info\n", pApp->wszFilename);
             StringCbCopyW(pApp->wszName, sizeof(pApp->wszName), pApp->wszFilename);
-            return NULL;
+
+            WCHAR wszPath[MAX_PATH];
+            if (!GetPathFromCmd(wszPath, pApp->wszCmd))
+            {
+                return NULL;
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


dll/win32/shell32/COpenWithMenu.cpp
```diff
@@ -363,7 +380,10 @@ BOOL COpenWithList::LoadInfo(COpenWithList::SApp *pApp)
     /* Query name */
     swprintf(wszBuf, L"\\StringFileInfo\\%04x%04x\\FileDescription", wLang, wCode);
     if (VerQueryValueW(pBuf, wszBuf, (LPVOID *)&pResult, &cchLen))
+    {
         StringCchCopyNW(pApp->wszName, _countof(pApp->wszName), pResult, cchLen);
+        success = cchLen > 1;
+    }
```
HBelusca, line 386: I wonder whether you could rearrange this block as follows:
```suggestion
    success = VerQueryValueW(pBuf, wszBuf, (LPVOID *)&pResult, &cchLen) && (cchLen > 1);
    if (success)
        StringCchCopyNW(pApp->wszName, _countof(pApp->wszName), pResult, cchLen);
```
in case retrieving the name is what really differentiates between a successful or a failed call (independently of whether the StringCchCopyNW call succeeded or not).

dll/win32/shell32/COpenWithMenu.cpp
```diff
@@ -396,8 +416,10 @@ VOID COpenWithList::GetPathFromCmd(LPWSTR pwszAppPath, LPCWSTR pwszCmd)
 
     /* Expand evn vers and optionally search for path */
     ExpandEnvironmentStrings(wszBuf, pwszAppPath, MAX_PATH);
-    if (!PathFileExists(pwszAppPath))
-        SearchPath(NULL, pwszAppPath, NULL, MAX_PATH, pwszAppPath, NULL);
+    if (PathFileExists(pwszAppPath))
+        return TRUE;
+    else
+        return SearchPath(NULL, pwszAppPath, NULL, MAX_PATH, pwszAppPath, NULL);
```
HBelusca, line 422: Minimal diff:
```suggestion
    if (!PathFileExists(pwszAppPath))
        return SearchPath(NULL, pwszAppPath, NULL, MAX_PATH, pwszAppPath, NULL);
    return TRUE;
```

dll/win32/shell32/COpenWithMenu.cpp
```diff
@@ -862,6 +884,7 @@ BOOL COpenWithDialog::IsNoOpen(HWND hwnd)
 VOID COpenWithDialog::AddApp(COpenWithList::SApp *pApp, BOOL bSelected)
 {
     LPCWSTR pwszName = m_pAppList->GetName(pApp);
+    if (!pwszName) return ;
```
HBelusca, line 887:
```suggestion
    if (!pwszName) return;
```
and same in the other AddApp function below.

dll/win32/shell32/COpenWithMenu.cpp
```diff
@@ -217,6 +217,12 @@ LPCWSTR COpenWithList::GetName(SApp *pApp)
         {
             WARN("Failed to load %ls info\n", pApp->wszFilename);
             StringCbCopyW(pApp->wszName, sizeof(pApp->wszName), pApp->wszFilename);
+
+            WCHAR wszPath[MAX_PATH];
+            if (!GetPathFromCmd(wszPath, pApp->wszCmd))
+            {
+                return NULL;
```
HBelusca, line 224: Actually I wonder why those missing applications are _not_ filtered already much before those functions (GetName(), and the other ones that call GetName()) are called; in other words, I wonder why `SApp` objects have been constructed for those missing applications....

whindsaks: Most of them (except any added from the Browse dialog later) come from the registry.

dll/win32/shell32/COpenWithMenu.cpp
whindsaks: Not checking StringCchCopyNW because we don't care if the name fits in the provided buffer (250 characters) or if it is truncated (SUCCEEDED will not help us) and the proper logic is not worth it, we will end up with something usable either way.

HBelusca: You mean we don't care because that name is something that will be only displayed, so it would be just a minor inconvenience if it gets truncated?

dll/win32/shell32/COpenWithMenu.cpp
learn-more: But is the `GetName` function supposed to filter them, or someone else?

KRosUser2: Same question, and it was part of the description of my original ticket

whindsaks: > But is the `GetName` function supposed to filter them, or someone else?

You could in theory move the validation to the part where it loads from the registry but that would make the code less generic. The current code gives you some options to filter away hidden and/or "not recommended" items without having to hit the disk during enumeration.

I suppose the function should be called something like `LoadName` or something like that, it is not a simple getter and it only gets called once per item (by the open with dialog or by the open with menu).

dll/win32/shell32/COpenWithMenu.cpp
```diff
@@ -231,7 +237,17 @@ HICON COpenWithList::GetIcon(SApp *pApp)
...
 
         GetPathFromCmd(wszPath, pApp->wszCmd);
-        ExtractIconExW(wszPath, 0, NULL, &pApp->hIcon, 1);
+        if (!ExtractIconExW(wszPath, 0, NULL, &pApp->hIcon, 1))
+        {
+            SHFILEINFO fi;
+            /* FIXME: Ideally we should include SHGFI_USEFILEATTRIBUTES because we already
+            ** know the file has no icons but SHGetFileInfo is broken in that case (CORE-19122).
+            ** Without SHGFI_USEFILEATTRIBUTES we needlessly hit the disk again but it will
+            ** return the correct default .exe icon.
```
JoachimHenze, line 246: Rather uncommon way to comment a block. I would recommend to transfom all "**" to " *" like we do anywhere else.

dll/win32/shell32/COpenWithMenu.cpp
```diff
@@ -875,7 +895,7 @@ VOID COpenWithDialog::AddApp(COpenWithList::SApp *pApp, BOOL bSelected)
         tvins.hParent = tvins.hInsertAfter = m_hOther;
 
     tvins.item.mask = TVIF_TEXT|TVIF_PARAM;
-    tvins.item.pszText = (LPWSTR)pwszName;
+    tvins.item.pszText = const_cast<LPWSTR>(pwszName);
```
JoachimHenze, line 898: superfluous change. I also don't think it is "more speaking" in any way.

dll/win32/shell32/COpenWithMenu.cpp
```diff
@@ -1189,14 +1209,15 @@ VOID COpenWithMenu::AddApp(PVOID pApp)
...
+    if (!pwszName) return;
 
     ZeroMemory(&mii, sizeof(mii));
     mii.cbSize = sizeof(mii);
     mii.fMask = MIIM_ID | MIIM_TYPE | MIIM_STATE | MIIM_DATA;
     mii.fType = MFT_STRING;
     mii.fState = MFS_ENABLED;
     mii.wID = m_idCmdLast;
-    mii.dwTypeData = (LPWSTR)pwszName;
+    mii.dwTypeData = const_cast<LPWSTR>(pwszName);
```
JoachimHenze, line 1220: superfluous change. I also don't think it is "more speaking" in any way.

learn-more (approved):

PR #5589, opened 2023-08-18
[RTL][NTDLL] Implement and export RtlInitializeCriticalSectionEx

tkreuzer:
## Purpose

Implement and export RtlInitializeCriticalSectionEx

## Tests

- Test WHS: https://reactos.org/testman/compare.php?ids=89057
- Test Win2003_x64: 
- Build GCCLin_x86 on Test KVM: 
- Build MSVC_x64 on Test KVM_x64:


sdk/lib/rtl/critical.c
```diff
@@ -575,62 +583,95 @@ RtlInitializeCriticalSectionAndSpinCount(PRTL_CRITICAL_SECTION CriticalSection,
...
-           CritcalSectionDebugData,
-           NtCurrentTeb()->ClientId.UniqueProcess);
-
-    if (!CritcalSectionDebugData)
+    if (Flags & RTL_CRITICAL_SECTION_FLAG_NO_DEBUG_INFO)
     {
-        /* This is bad! */
-        DPRINT1("Couldn't allocate Debug Data for: %p\n", CriticalSection);
-        return STATUS_NO_MEMORY;
+        CriticalSection->DebugInfo = LongToPtr(-1);
```
HBelusca, line 588: I wonder why DebugInfo set to NULL is _not enough_ to signal that the critical section indeed doesn't have debug information....

(Side remark: if you early-return success there, this would allow to now reindent the whole existing code below, thus reducing the overall diff.)

sdk/lib/rtl/critical.c
```diff
@@ -575,62 +583,95 @@ RtlInitializeCriticalSectionAndSpinCount(PRTL_CRITICAL_SECTION CriticalSection,
...
+ *     STATUS_SUCCESS.
+ *
+ * Remarks:
+ *     SpinCount is ignored on single-processor systems.
+ *
+ *--*/
+NTSTATUS
+NTAPI
+RtlInitializeCriticalSectionAndSpinCount(PRTL_CRITICAL_SECTION CriticalSection,
+    ULONG SpinCount)
```
HBelusca, line 670: You may add the missing SAL annotations.

sdk/lib/rtl/critical.c
whindsaks: Guess #1: Is there some way to delay allocate the debug info?

Guess #2: Compatibility thing because the CS internals have changed a couple of times. Maybe some apps initialize directly with {0} instead of calling InitializeCriticalSection?

sdk/lib/rtl/critical.c
```diff
@@ -28,6 +28,10 @@ extern HANDLE LdrpShutdownThreadId;
 
 /* FUNCTIONS *****************************************************************/
 
+#define CRITSECT_HAS_DEBUG_INFO(CriticalSection) \
+    (((CriticalSection)->DebugInfo != NULL) && \
```
learn-more, line 32: Isn't this also the field those wine-hacked critical sections have the filename stored in?

sdk/lib/rtl/critical.c
tkreuzer: I added a test. That test shows it is like that.

sdk/lib/rtl/critical.c
tkreuzer: No, they store it in a field in the debug info: https://git.reactos.org/?p=reactos.git;a=blob;f=sdk/include/psdk/winbase.h;hb=56d95154eeaf2e96d5f3c6cd47baa03e50435235#l879

modules/rostests/apitests/ntdll/RtlCriticalSection.c
```diff
@@ -0,0 +1,362 @@
...
+    hEventThread2Cont = CreateEvent(NULL, TRUE, FALSE, NULL);
+
+    // Create thread 1 and wait to it time to try to acquire the critical section
+    CreateThread(NULL, 0, ThreadProc1, NULL, 0, &dwThreadId1);
+    Sleep(10);
+
+    ok_long(CritSect.LockCount, -6);
+    ok_long(CritSect.RecursionCount, 2);
+    ok_ptr(CritSect.OwningThread, UlongToHandle(GetCurrentThreadId()));
+    //ok_ptr(CritSect.LockSemaphore, LongToPtr(-1));
```
HBelusca, line 304: What happens with those commented tests? In which cases they work and don't?

modules/rostests/apitests/ntdll/RtlCriticalSection.c
```diff
@@ -0,0 +1,362 @@
+/*
+ * PROJECT:         ReactOS api tests
+ * LICENSE:         LGPLv2.1+ - See COPYING.LIB in the top level directory
+ * PURPOSE:         Test for Rtl Critical Section API
+ * PROGRAMMERS:     Thomas Faber <thomas.faber@reactos.org>
```
HBelusca, line 5: This may be adjusted.

modules/rostests/apitests/ntdll/RtlCriticalSection.c
```diff
@@ -0,0 +1,362 @@
...
+    else
+    {
+        ok(CritSect.DebugInfo == LongToPtr(-1), "DebugInfo is %p\n", CritSect.DebugInfo);
+    }
+
+    Status = RtlInitializeCriticalSectionAndSpinCount(&CritSect, 0xFF000000);
+    ok_ntstatus(Status, STATUS_SUCCESS);
+    ok_size_t(CritSect.SpinCount, DefaultSpinCount);
+
+    if (pfnRtlInitializeCriticalSectionEx != NULL)
```
HBelusca, line 87: What about:
```suggestion
    if (pfnRtlInitializeCriticalSectionEx == NULL)
    {
        skip("RtlInitializeCriticalSectionEx not available.\n");
        return;
    }
```

modules/rostests/apitests/ntdll/RtlCriticalSection.c
tkreuzer: They behave differently on different Windows versions. Didn't bother to investigate.

modules/rostests/apitests/ntdll/RtlCriticalSection.c
tkreuzer: No, because then someone adds tests at the end, and they will never run.

HBelusca: yes maybe, if those other tests have nothing to do with the RtlInitialize***SectionEx function.

DarkFire01 (approved):

modules/rostests/apitests/ntdll/RtlCriticalSection.c
```diff
@@ -0,0 +1,374 @@
...
+        ok_ntstatus(Status, STATUS_ACCESS_VIOLATION);
+
+        Status = pfnRtlInitializeCriticalSectionEx(&CritSect, 0, 0);
+        ok_ntstatus(Status, STATUS_SUCCESS);
+        ok_long(CritSect.LockCount, -1);
+        ok_long(CritSect.RecursionCount, 0);
+        ok_ptr(CritSect.OwningThread, NULL);
+        ok_ptr(CritSect.LockSemaphore, NULL);
+        ok_size_t(CritSect.SpinCount, DefaultSpinCount);
+        if (HasDebugInfo)
```
ThFabba, line 120: This branch seems unreachable. `RtlInitializeCriticalSectionEx` only exists on NT 6.x, `HasDebugInfo` is never true for NT 6.x. Is there some kind of compatibility mode setting involved in making this possible?

modules/rostests/apitests/ntdll/RtlCriticalSection.c
```diff
@@ -0,0 +1,374 @@
...
+        {
+            Status = _SEH2_GetExceptionCode();
+        }
+        _SEH2_END;
+        ok_ntstatus(Status, STATUS_INVALID_PARAMETER_2);
+
+        for (ULONG i = 0; i < 32; i++)
+        {
+            ULONG Flags = 1UL << i;
+            ULONG AllowedFlags = 0x07000000;
```
ThFabba, line 174: Nit: I'd find this more straightforward if you added the `0xFFFFFF` bits here instead of below.

modules/rostests/apitests/ntdll/RtlCriticalSection.c
```diff
@@ -0,0 +1,374 @@
...
+
+    RtlInitializeCriticalSection(&CritSect);
+
+    // Acquire once
+    RtlEnterCriticalSection(&CritSect);
+    ok_long(CritSect.LockCount, -2);
+    ok_long(CritSect.RecursionCount, 1);
+    ok_ptr(CritSect.OwningThread, UlongToHandle(GetCurrentThreadId()));
+    ok_ptr(CritSect.LockSemaphore, NULL);
+    //ok_size_t(CritSect.SpinCount, 0x20007d0);
```
ThFabba, line 294: Might want to comment as to why this is commented out.

If the reason is that it depends on UP vs MP -- maybe just save the initial value and then compare with that to show "acquire doesn't change spin count"?

modules/rostests/apitests/ntdll/RtlCriticalSection.c
```diff
@@ -0,0 +1,374 @@
...
+
+    hEventThread1Ready = CreateEvent(NULL, TRUE, FALSE, NULL);
+    hEventThread1Cont = CreateEvent(NULL, TRUE, FALSE, NULL);
+    hEventThread2Ready = CreateEvent(NULL, TRUE, FALSE, NULL);
+    hEventThread2Cont = CreateEvent(NULL, TRUE, FALSE, NULL);
+
+    // Create thread 1 and wait to it time to try to acquire the critical section
+    CreateThread(NULL, 0, ThreadProc1, NULL, 0, &dwThreadId1);
+    Sleep(10);
```
ThFabba, line 311: This seems short. Perhaps poll on `LockCount` with a timeout, to avoid flakiness?

modules/rostests/apitests/ntdll/RtlCriticalSection.c
ThFabba: Would you mind saying that in a comment in the code?

modules/rostests/apitests/ntdll/RtlCriticalSection.c
```diff
@@ -0,0 +1,374 @@
...
+    CreateThread(NULL, 0, ThreadProc2, NULL, 0, &dwThreadId2);
+    Sleep(10);
+
+    ok_long(CritSect.LockCount, -10);
+    ok_long(CritSect.RecursionCount, 2);
+    ok_ptr(CritSect.OwningThread, UlongToHandle(GetCurrentThreadId()));
+    //ok_ptr(CritSect.LockSemaphore, LongToPtr(-1));
+    //ok_size_t(CritSect.SpinCount, 0x20007d0);
+
+    RtlLeaveCriticalSection(&CritSect);
```
ThFabba, line 329: Perhaps test that this decreases RecursionCount but not LockCount? If we ever introduced a bug there that would be very hard to track down outside of a test.

modules/rostests/apitests/ntdll/RtlCriticalSection.c
```diff
@@ -0,0 +1,374 @@
...
+
+    ok_long(CritSect.LockCount, -2);
+    ok_long(CritSect.RecursionCount, 1);
+    ok_ptr(CritSect.OwningThread, UlongToHandle(dwThreadId2));
+    //ok_ptr(CritSect.LockSemaphore, LongToPtr(-1));
+    //ok_size_t(CritSect.SpinCount, 0x20007cf);
+
+    // Release thread 2
+    SetEvent(hEventThread2Cont);
+}
```
ThFabba, line 353: Any point in waiting on the two thread handles for a "cleaner" exit?

modules/rostests/apitests/ntdll/RtlCriticalSection.c
tkreuzer: `BOOL HasDebugInfo = (g_Version <= 0x600);`

Smaller or equal :)

modules/rostests/apitests/ntdll/RtlCriticalSection.c
tkreuzer: Fixed SpinCount, added a comment for LockSemaphore.

modules/rostests/apitests/ntdll/RtlCriticalSection.c
ThFabba: Ohhh, that dang Vista! :p

modules/rostests/apitests/ntdll/RtlCriticalSection.c
```diff
@@ -0,0 +1,411 @@
...
+            if (g_Version > 0x600) AllowedFlags |= 0x18000000;
+            NTSTATUS ExpectedStatus = (Flags & ~AllowedFlags) ?
+                STATUS_INVALID_PARAMETER_3 : STATUS_SUCCESS;
+            Status = pfnRtlInitializeCriticalSectionEx(&CritSect, 0, Flags);
+            ok(Status == ExpectedStatus, "Wrong Status (0x%lx) for Flags 0x%lx\n",
+                Status, Flags);
+            if (NT_SUCCESS(Status))
+            {
+                ULONG SetFlags = Flags & 0x08000000;
+                if (g_Version > 0x600) SetFlags |= (Flags & 0x03000000);
```
HBelusca, line 175:
```suggestion
                if (g_Version > _WIN32_WINNT_VISTA)
                    SetFlags |= (Flags & 0x03000000);
```

HBelusca (commented):
suggestion for using version number defines.

modules/rostests/apitests/ntdll/RtlCriticalSection.c
learn-more: I assume (but can't recall exactly) that application verifier might be using this field.

And there is this appcompat shim flag:
```
FLAG: AllocDebugInfoForCritSections
  NAME: AllocDebugInfoForCritSections
  FIX_ID: (Binary data)
  FLAG_MASK_KERNEL: 4194304
  GENERAL
```

We don't have those yet, but I could add them?

tkreuzer: OK, but why?

learn-more: Windows uses it for emulating older systems:
image

tkreuzer: OK, but this has nothing to do with this PR. It would also be easier and less hacky to simply use `if (OsVersion <= _WIN32_WINNT_VISTA)` or something like that.

modules/rostests/apitests/ntdll/RtlCriticalSection.c
```diff
@@ -0,0 +1,411 @@
...
+START_TEST(RtlCriticalSection)
+{
+    HMODULE hmodNtDll = GetModuleHandleA("ntdll.dll");
+    pfnRtlInitializeCriticalSectionEx = (FN_RtlInitializeCriticalSectionEx*)
+        GetProcAddress(hmodNtDll, "RtlInitializeCriticalSectionEx");
+
+    g_VerInfo.dwOSVersionInfoSize = sizeof(g_VerInfo);
+    GetVersionExA((LPOSVERSIONINFOA)&g_VerInfo);
+    g_Version = g_VerInfo.dwMajorVersion << 8 | g_VerInfo.dwMinorVersion;
+    printf("g_VerInfo: %lu.%lu.%lu ('%s')\n ",
```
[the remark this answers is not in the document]
tkreuzer: yes

modules/rostests/apitests/ntdll/RtlCriticalSection.c
tkreuzer: Not really, but I added it anyway.

modules/rostests/apitests/ntdll/RtlCriticalSection.c
```diff
@@ -0,0 +1,411 @@
...
+            Status = _SEH2_GetExceptionCode();
+        }
+        _SEH2_END;
+        ok_ntstatus(Status, STATUS_INVALID_PARAMETER_2);
+
+        for (ULONG i = 0; i < 32; i++)
+        {
+            ULONG Flags = 1UL << i;
+            ULONG AllowedFlags = 0x07FFFFFF;
+            if (g_Version > 0x600) AllowedFlags |= 0x18000000;
```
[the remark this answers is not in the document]
tkreuzer: Changed it

modules/rostests/apitests/ntdll/RtlCriticalSection.c
tkreuzer: Fixed

sdk/lib/rtl/critical.c
```diff
@@ -562,10 +568,41 @@ RtlInitializeCriticalSection(PRTL_CRITICAL_SECTION CriticalSection)
...
+
+    /* These flags generally allowed */
+    AllowedFlags = RTL_CRITICAL_SECTION_FLAG_NO_DEBUG_INFO |
+                   RTL_CRITICAL_SECTION_FLAG_DYNAMIC_SPIN |
+                   RTL_CRITICAL_SECTION_FLAG_STATIC_INIT;
+
+    /* Flags for Windows 7+ (CHECKME) */
+    OsVersion = NtCurrentPeb()->OSMajorVersion << 8 |
+                NtCurrentPeb()->OSMinorVersion;
+    if (OsVersion >= 0x601)
```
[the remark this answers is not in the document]
tkreuzer: No, NTDDI is way different and relates to the kernel, not user mode.

ThFabba (approved):

HBelusca (approved):

PR #5594, opened 2023-08-19
[WINESYNC][SDK][DWMAPI] Import DWMAPI from WINE 8.14

DarkFire01:
## Purpose

Directly import the DWMAPI dll from wine.
This dll won't be included at the moment until DLL_EXPORT_VERSION is set to 0x600 or greater.

While this dll is mostly a stub in latest WINE, this is to make sure I understand the WINE import process correctly and pitch my idea of throwing all the vista+ specific DLLs into a separate directory so they can be enabled or disabled without dirtying up the main win32 dll cmakelist.

This dll is required for pretty much every app since 2010. :trollface: 

## Proposed changes

- Directly import the DWMAPI dll from wine.


HBelusca:
> This dll is required for pretty much every app since 2010. :trollface:

I suppose the dwm.exe should be part of that thing? (It would make sense to put that dll inside win32ss as it will certainly be tied up to the rest of win32k and stuff)

DarkFire01:
> > This dll is required for pretty much every app since 2010. :trollface:
> 
> I suppose the dwm.exe should be part of that thing? (It would make sense to put that dll inside win32ss as it will certainly be tied up to the rest of win32k and stuff)

Well, let's keep it as a WINE thing for now. 
though admittedly even in windows DWMAPI works without dwm.exe even running or in the system. 

This DLL is just for win32 apps.

down the line the suggestion should make sense, but I'd like to keep all this stuff quarantined for now.

sdk/include/psdk/winerror.h
```diff
@@ -3311,6 +3319,8 @@
 #define WINCODEC_ERR_WIN32ERROR                            _HRESULT_TYPEDEF_(0x88982f94)
 #define WINCODEC_ERR_INVALIDPROGRESSIVELEVEL               _HRESULT_TYPEDEF_(0x88982f95)
 
+#define MILERR_MISMATCHED_SIZE                             _HRESULT_TYPEDEF_(0x88980090)
```
HBelusca, line 3322: I'll check official Windows SDK to see whether they put the MIL/DWM errors also in winerror.h

HBelusca: Yhile the DWM_*** errors above can indeed be found in Windows PSDK (7+, and perhaps Vista too), I cannot find this MILERR error (and others) in there, not even in Windows 10 PSDK.
I know that Wine added lots of these there too: https://source.winehq.org/patches/data/235259

Do you know whether these defines are actually new (like Windows 11+)? or just undocumented? (in which case they may have to go somewhere, like in dwmapi.h ...)

DarkFire01: I don't think they're particularly new, WINE puts them in winerror it seems. 

I don't mind putting them in dwmapi.h, I don't think we will sync this ever again.

HBelusca: Actually yes you're correct.
I've found them in https://raw.githubusercontent.com/tpn/winsdk-10/master/Include/10.0.16299.0/shared/winerror.h for example (which means that github search is still broken and doesn't systematically search within files)

HBelusca (approved):
OK with the sync per se. The details of the actual placement of the module (win32_vista? somewhere else?...) should be discussed with @tkreuzer @learn-more perhaps.

Question: Do you plan to fork dwmapi at some point? If not it may be of interest to have a look at this "winesync.py" script of ours.

DarkFire01:
> OK with the sync per se. The details of the actual placement of the module (win32_vista? somewhere else?...) should be discussed with @tkreuzer @learn-more perhaps.
> 
> Question: Do you plan to fork dwmapi at some point? If not it may be of interest to have a look at this "winesync.py" script of ours.

Sort of, apps really need this to do anything. So this needed to happen now. Of course WINE has no interest in implementing DWM.

As I discussed in the discord and other places at one point; DWMCORE has been open sourced under MIT so implementing a correct DWM implementation isn't out of the question anymore, But this isn't something im personally interested in doing.

Other people have shown interest so it'll likely be forked sooner rather then later.

wjk:
> DWMCORE has been open sourced under MIT

If this is actually true, can you post a link to the source?

DarkFire01:
> OK with the sync per se. The details of the actual placement of the module (win32_vista? somewhere else?...) should be discussed with @tkreuzer @learn-more perhaps.
> 
> Question: Do you plan to fork dwmapi at some point? If not it may be of interest to have a look at this "winesync.py" script of ours.

Suggested I move it back to under win32 directory and just keep everything organized as it is,  I’ll do this when I get home. Then it’ll be ready for merge again whenever.

media/doc/WINESYNC.txt
```diff
@@ -230,6 +230,8 @@ dll/win32/xinput9_1_0         # Synced to WineStaging-2.9
 dll/win32/xmllite             # Synced to WineStaging-4.18
 dll/win32/xolehlp             # Synced to WineStaging-3.21
 
+dll/win32_vista/dwmapi        # Synced to Wine-8.14
```
wjk, line 233: This should be updated to `dll/win32/dwmapi`.

wjk (changes requested):

PR #5598, opened 2023-08-20
[COMCTL32] Set TBBUTTON::fsState   before  RB_SETBANDINFOW message returns.

zecarlos1957:
## Observed behavior:
In BrowserUI window, when deselecting the visibility of the standard buttons from  View menu , the toolbar is hidden, but if we move the mouse over the address bar, the buttons become visible.

## Proposed change:
The correct fix should be done  in commctl32/rebar.c.
As on Windows, the state of toolbar buttons must be set before the RB_SETBANDINFOW message returns. This solves the two previously listed problems in CORE-18742

JIRA issue: [CORE-18742](https://jira.reactos.org/browse/CORE-18742)

### DIFF
--- a/dll/win32/browseui/internettoolbar.cpp
+++ b/dll/win32/browseui/internettoolbar.cpp
@@ -1109,12 +1109,7 @@ HRESULT CInternetToolbar::ToggleBandVisibility(int BandID)
         {
             TBBUTTONINFO BtnInfo = {sizeof(TBBUTTONINFO), TBIF_BYINDEX | TBIF_STATE};
             ::SendMessage(fToolbarWindow, TB_GETBUTTONINFO, idx, (LPARAM)&BtnInfo);
-
-            if (bandInfo.fStyle & RBBS_HIDDEN)
-                BtnInfo.fsState |= TBSTATE_HIDDEN;
-            else
-                BtnInfo.fsState &= ~TBSTATE_HIDDEN;
-
+            BtnInfo.fsState = bandInfo.fStyle & RBBS_HIDDEN ? BtnInfo.fsState | TBSTATE_HIDDEN : BtnInfo.fsState & ~TBSTATE_HIDDEN;
             ::SendMessage(fToolbarWindow, TB_SETBUTTONINFO, idx, (LPARAM)&BtnInfo);
         }
     }
--- a/dll/win32/browseui/internettoolbar.cpp
+++ b/dll/win32/browseui/internettoolbar.cpp
@@ -1109,7 +1109,12 @@ HRESULT CInternetToolbar::ToggleBandVisibility(int BandID)
         {
             TBBUTTONINFO BtnInfo = {sizeof(TBBUTTONINFO), TBIF_BYINDEX | TBIF_STATE};
             ::SendMessage(fToolbarWindow, TB_GETBUTTONINFO, idx, (LPARAM)&BtnInfo);
-            BtnInfo.fsState = bandInfo.fStyle & RBBS_HIDDEN ? BtnInfo.fsState | TBSTATE_HIDDEN : BtnInfo.fsState & ~TBSTATE_HIDDEN;
+
+            if (bandInfo.fStyle & RBBS_HIDDEN)
+                BtnInfo.fsState |= TBSTATE_HIDDEN;
+            else
+                BtnInfo.fsState &= ~TBSTATE_HIDDEN;
+
             ::SendMessage(fToolbarWindow, TB_SETBUTTONINFO, idx, (LPARAM)&BtnInfo);
         }
     }
--- a/dll/win32/browseui/internettoolbar.cpp
+++ b/dll/win32/browseui/internettoolbar.cpp
@@ -1101,6 +1101,19 @@ HRESULT CInternetToolbar::ToggleBandVisibility(int BandID)
     else
         bandInfo.fStyle |= RBBS_HIDDEN;
 
+    if (BandID == ITBBID_TOOLSBAND)
+    {
+        int BtnCount = ::SendMessage(fToolbarWindow, TB_BUTTONCOUNT, 0, 0);
+
+        for (int idx = 0; idx < BtnCount; idx++)
+        {
+            TBBUTTONINFO BtnInfo = {sizeof(TBBUTTONINFO), TBIF_BYINDEX | TBIF_STATE};
+            ::SendMessage(fToolbarWindow, TB_GETBUTTONINFO, idx, (LPARAM)&BtnInfo);
+            BtnInfo.fsState = bandInfo.fStyle & RBBS_HIDDEN ? BtnInfo.fsState | TBSTATE_HIDDEN : BtnInfo.fsState & ~TBSTATE_HIDDEN;
+            ::SendMessage(fToolbarWindow, TB_SETBUTTONINFO, idx, (LPARAM)&BtnInfo);
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


HBelusca:
Cc @cbialorucki and @whindsaks too for review.

dll/win32/browseui/internettoolbar.cpp
```diff
@@ -1101,6 +1101,19 @@ HRESULT CInternetToolbar::ToggleBandVisibility(int BandID)
...
 
+    if (BandID == ITBBID_TOOLSBAND)
+    {
+        int BtnCount = ::SendMessage(fToolbarWindow, TB_BUTTONCOUNT, 0, 0);
+
+        for (int idx = 0; idx < BtnCount; idx++)
+        {
+            TBBUTTONINFO BtnInfo = {sizeof(TBBUTTONINFO), TBIF_BYINDEX | TBIF_STATE};
+            ::SendMessage(fToolbarWindow, TB_GETBUTTONINFO, idx, (LPARAM)&BtnInfo);
+            BtnInfo.fsState = bandInfo.fStyle & RBBS_HIDDEN ? BtnInfo.fsState | TBSTATE_HIDDEN : BtnInfo.fsState & ~TBSTATE_HIDDEN;
```
HBelusca, line 1112:
```suggestion
            BtnInfo.fsState = (bandInfo.fStyle & RBBS_HIDDEN) ? (BtnInfo.fsState | TBSTATE_HIDDEN) : (BtnInfo.fsState & ~TBSTATE_HIDDEN);
```
or in case this actually makes the code more readable:
```suggestion
            if (bandInfo.fStyle & RBBS_HIDDEN)
                BtnInfo.fsState |= TBSTATE_HIDDEN
            else
                BtnInfo.fsState &= ~TBSTATE_HIDDEN;
```

dll/win32/browseui/internettoolbar.cpp
```diff
@@ -1101,6 +1101,19 @@ HRESULT CInternetToolbar::ToggleBandVisibility(int BandID)
     else
         bandInfo.fStyle |= RBBS_HIDDEN;
 
+    if (BandID == ITBBID_TOOLSBAND)
```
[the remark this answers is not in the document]
zecarlos1957: This corresponds  just to the BrowserUI toolbar, which can be hidden by unchecking the Show/Toolbars/Standard Buttons menu item checkbox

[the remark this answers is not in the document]
zecarlos1957: After testing on windows it seems this is not the correct fix. 
keep investigating

whindsaks:
Just hide the toolbar window instead of its buttons.

cbialorucki (commented):
Some code style suggestions

dll/win32/comctl32/rebar.c
```diff
@@ -2732,6 +2732,11 @@ REBAR_strdifW( LPCWSTR a, LPCWSTR b )
     return ( (a && !b) || (b && !a) || (a && b && lstrcmpW(a, b) ) );
 }
 
+#ifdef __REACTOS__
+static LRESULT
+REBAR_ShowBand (REBAR_INFO *infoPtr, INT iBand, BOOL show);
+#endif // __REACTOS__
```
cbialorucki, line 2738:
```suggestion
#endif
```

dll/win32/comctl32/rebar.c
```diff
@@ -2775,7 +2780,13 @@ REBAR_SetBandInfoT(REBAR_INFO *infoPtr, INT iBand, const REBARBANDINFOW *lprbbi,
 	  REBAR_Layout(infoPtr);
 	  InvalidateRect(infoPtr->hwndSelf, NULL, TRUE);
     }
-
+#ifdef __REACTOS__
+    if (uChanged & RBBIM_STYLE)
+    {
+        BOOL bVisible = (lprbbi->fStyle & RBBS_HIDDEN) != RBBS_HIDDEN;
+        REBAR_ShowBand(infoPtr, iBand, bVisible == TRUE);
```
cbialorucki, line 2787:
```suggestion
        REBAR_ShowBand(infoPtr, iBand, bVisible);
```

zecarlos1957:
@cbialorucki  Done. Thanks.

dll/win32/comctl32/rebar.c
```diff
@@ -2775,7 +2780,13 @@ REBAR_SetBandInfoT(REBAR_INFO *infoPtr, INT iBand, const REBARBANDINFOW *lprbbi,
 	  REBAR_Layout(infoPtr);
 	  InvalidateRect(infoPtr->hwndSelf, NULL, TRUE);
     }
-
+#ifdef __REACTOS__
```
HBelusca, line 2783: Should this be done *after* (as you currently did), or *before* the uChanged thing done just above?
Or maybe your new REBAR_ShowBand call be done **inside** REBAR_Layout ?

zecarlos1957: You're absolutely right. The functions REBAR_Layout(infoPtr) and InvalidateRect(...) are already called when finishing the REBAR_SHowBand() function, so just call REBAR_ShowBand().

HBelusca: Not sure you should handle any changed flag by a call to REBAR_ShowBand (even if that function also calls the REBAR_Layout + InvalidateRect at its end), since it also calls every time ShowWindow.

This is why I mentioned that maybe the REBAR_ShowBand may be done _inside_ REBAR_Layout instead, of course only under some circumstances.

Irrespectively of this, as the current code is, you may have noticed the following:
REBAR_Layout is (amongst others) called by REBAR_InsertBandT, which is the handler for RB_INSERTBAND(A/W) too. That message also takes a REBARBANDINFO(A/W) structure. This structure can also specify, much like what is done in the RB_SETBANDINFO(A/W) handler you are modifying here (REBAR_SetBandInfoT), the initial style of the rebar, in particular whether it is going to be hidden. However, handling that visibility is currently absent in REBAR_InsertBandT. I think it may be necessary to add it there too.

dll/win32/comctl32/rebar.c
```diff
@@ -2732,6 +2732,11 @@ REBAR_strdifW( LPCWSTR a, LPCWSTR b )
     return ( (a && !b) || (b && !a) || (a && b && lstrcmpW(a, b) ) );
 }
 
+#ifdef __REACTOS__
+static LRESULT
+REBAR_ShowBand (REBAR_INFO *infoPtr, INT iBand, BOOL show);
```
JoachimHenze, line 2737:
```suggestion
REBAR_ShowBand(REBAR_INFO *infoPtr, INT iBand, BOOL show);
```

HBelusca:
I don't know why @zecarlos1957 closed this PR, but @cbialorucki @DarkFire01 , do you know whether this PR is still valid and if so, could be applied after the comctl32 wine sync?

binarymaster:
@HBelusca [CORE-18742](https://jira.reactos.org/browse/CORE-18742) has been marked as duplicate of [CORE-17236](https://jira.reactos.org/browse/CORE-17236), which in turn got fixed by @whindsaks in 674136bcd02a766dd8baa8a168551850eec83abc.

whindsaks:
My fix is Explorer specific. This might still be valid and could fix other applications using the ReBar.

HBelusca:
A simple interactive win32 test app for this would then be useful.

PR #5600, opened 2023-08-20
[MKISOFS] Use prototypes in all function definitions

wjk:
## Purpose

The C syntax in the mkisofs components is so ridiculously ancient it causes my compiler (a version of Clang, which is notoriously picky) to output warnings about every function definition in this part of the codebase.

JIRA issue: none

## Proposed changes

- Change all legacy non-prototype function definitions in the components of mkisofs to standard C syntax.


SergeGautherie:
This is 3rd-party code, so I am not sure it should be done, at least not that way.

wjk:
Closing in favor of #5603.

SergeGautherie:
Note:
From the quick look I had, I would suggest to look into using the `PROTOTYPES` define where it is missing and send that upstream.

PR #5601, opened 2023-08-21
[CONSRV] Don't blink cursor if the console window is not active

whindsaks:
The cursor is not supposed to blink if the console window is not active.

Ideally the blink timer would be turned completely off when the window is inactive but that would require the hack in `GuiWriteStream` to change to a `PostMessage` with a private message?  The scrolling code that uses the same timer would also need changes if scrolling is possible when the window is not active.

A compromise would perhaps be to set the timer to a larger value when the window is not active? `4 * CURSOR_BLINK_TIME`?

### DIFF
--- a/win32ss/user/winsrv/consrv/frontends/gui/conwnd.c
+++ b/win32ss/user/winsrv/consrv/frontends/gui/conwnd.c
@@ -642,6 +642,7 @@ OnNcCreate(HWND hWnd, LPCREATESTRUCTW Create)
 
     GuiData->hWindow = hWnd;
     GuiData->hSysMenu = GetSystemMenu(hWnd, FALSE);
+    GuiData->IsWindowActive = FALSE;
 
     /* Initialize the fonts */
     if (!InitFonts(GuiData,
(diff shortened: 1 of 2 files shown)


HBelusca:
(For note, the hackish refresh handling of the screen within the Timer function that was originally for the cursor blinking, was added in https://github.com/reactos/reactos/commit/9cb27f40af5720a054c01c76188d4f953fe34107 .)

JoachimHenze (approved):

win32ss/user/winsrv/consrv/frontends/gui/conwnd.c
```diff
@@ -725,6 +726,8 @@ OnActivate(PGUI_CONSOLE_DATA GuiData, WPARAM wParam)
 
     DPRINT("WM_ACTIVATE - ActivationState = %d\n", ActivationState);
 
+    GuiData->IsWindowActive = ActivationState != WA_INACTIVE;
```
binarymaster, line 729:
```suggestion
    GuiData->IsWindowActive = (ActivationState != WA_INACTIVE);
```

katahiromz (approved):

whindsaks:
> This PR is stale because it received no updates in the last 4 months. Without removing the stale label, or commenting on this ticket it will be closed in 2 weeks.

Go away stupid bot

PR #5603, opened 2023-08-21
[MKISOFS] Silence prototype warnings

wjk:
## Purpose

mkisofs and its components uses K&R-style function definitions to support very old compilers. Modern compilers consider K&R syntax to be deprecated. Clang therefore emits a large number of warnings over this. Supplants #5600.

JIRA issue: none

## Proposed changes

- Add `-Wno-deprecated-non-prototype` compiler flag to mkisofs and its support libraries


HBelusca (approved):

binarymaster (approved):

ColinFinck (approved):
Thank you for going the extra mile and finding an alternative to #5600.

These tools are now developed at https://codeberg.org/schilytools/schilytools, but a quick look confirmed that even the latest versions are still using K&R-style C to this date.
If we want to change that, we need to submit a patch there.

PR #5604, opened 2023-08-21
Implement some missing functionality for Dxg.sys

NiDragon:
## Purpose

My work is comprised mainly of implementing a large number of stubs in dxg.sys along with various helper functions to make them work. Reworking our existing understanding of the DirectDraw internal structures introducing a few missing structures for utility functions. There is still a few missing core functionalities like releasing certain DirectDraw objects and implementing Blt for surfaces.

## Proposed changes

- Implemented about 40 something stubbed functions
- Implemented a large portion of setup and shutdown of dxg.sys
- Filled in some blank spaces in existing structures
- Added new structures for dxg.sys

## TODO

- [ Implement Surface Blting ] 
- [ Implement Linear and Rectangular VideoMemory Region Allocation and Release. ] 
- [ Handle Notification Events for state of DirectDraw objects. ]
- [ Reorganize all implemented functions into their respective source files. ]
- [ Implement necessary SEH2 Exception handling ]
- [ Add SAL2 Annotations ]
- [ Research EDD_VIDEOPORT and EDD_MOTIONCOMP objects ]

PR #5607, opened 2023-08-22
[DXGTHK] Export DriverEntry

oleg-dubinskiy:
## Purpose

Export driver entry point function `DriverEntry` from our dxgthk.sys. It does not improve anything so much, it just mimics Windows behaviour. Just do it the same in ReactOS as it's done in Windows. Confirmed that it is exported on Windows XP SP3 at least. See attached screenshot. :slightly_smiling_face: 
Small addendum to #5604 PR.

dxgthk_xp

JIRA issue: None

### DIFF
--- a/win32ss/reactx/dxgthk/dxgthk.spec
+++ b/win32ss/reactx/dxgthk/dxgthk.spec
@@ -1,3 +1,4 @@
+@ stdcall DriverEntry(ptr ptr)
 @ stdcall EngAcquireSemaphore(ptr) win32k.EngAcquireSemaphore
 @ stdcall EngAllocMem(long long long) win32k.EngAllocMem
 @ stdcall EngAllocUserMem(long long) win32k.EngAllocUserMem


GeoB99:
> It does not improve anything so much, it just mimics Windows behaviour.

As you said, it doesn't do anything much, I don't see the point of mimicking Windows' behavior unless it actually benefits something or somebody does actually call this function (which I kind of doubt).

HBelusca:
> Confirmed that it is exported on Windows XP SP3 at least

Also we are more targeting Windows Server 2003 SP2 (for NT5.x) and with forward-compatibility with NT 6+

win32ss/reactx/dxgthk/dxgthk.spec
```diff
@@ -1,3 +1,4 @@
+@ stdcall DriverEntry(ptr ptr)
```
HBelusca, line 1: Isn't the entry-point usually automatically "exported" ?

oleg-dubinskiy: No, it isn't. If to open our dxgthk.sys on recent ROS master via Dependency Walker, it isn't exported at all, unlike Windows:
ros_no_export

oleg-dubinskiy:
> > Confirmed that it is exported on Windows XP SP3 at least
> 
> Also we are more targeting Windows Server 2003 SP2 (for NT5.x) and with forward-compatibility with NT 6+

@HBelusca: I can confirm it's also exported on Windows Server 2003 as well:
2k3-exported

And most likely it's also done on Vista and newer. Although I have no any of newer Windows versions to check it on, but nevertheless.
So doing this indeed makes sense. Perhaps there is some stuff existing which calls this export...

oleg-dubinskiy:
> And most likely it's also done on Vista and newer. Although I have no any of newer Windows versions to check it on, but nevertheless.

Oh, just checked! As it appears, Vista and newer have no dxgthk.sys at all! It only has dxgkrnl.sys instead! My mistake, sorry. :sweat_smile:

tkreuzer (approved):

HBelusca (approved):
OK this is a rather minor trivial change so I guess we can live with it.

PR #5585, opened 2023-08-18
[PSDK] enum FILE_INFO_BY_HANDLE_CLASS: Add NT version conditions

SergeGautherie:
## Purpose

Improve/Fix `enum FILE_INFO_BY_HANDLE_CLASS` definition.

## Proposed changes

-  Check/Update value list and related NT versions.

## TODO

- [ ] Helpwanted, to sort out `FILE_INFO_BY_HANDLE_CLASS` based on MS header(s). Thanks.
- [ ] Remove ShellBtrfs commit, which is there to give context only.

PR #5611, opened 2023-08-23
[SHELLBTRFS] Downgrade requirement to _WIN32_WINNT=0x601

SergeGautherie:
## Purpose

Lower extra requirement.
Somewhat related to NT6 work.

## Proposed changes

- shellext.h: Fix define condition of NT6.x+ features
Note: Refused upstream as https://github.com/maharmstone/btrfs/pull/588, so fix this for ROS only.
- CMakeLists.txt: Downgrade to _WIN32_WINNT=0x601
Part 1/2 of reverting this requirement upgrade on https://github.com/reactos/reactos/commit/c032d89e77509055ab647f51b82b028e0e6e2fce (r75793), after code fix.
Note: Part 2/2 will need #5585.

### DIFF
--- a/dll/shellext/shellbtrfs/shellext.h
+++ b/dll/shellext/shellbtrfs/shellext.h
@@ -193,6 +193,7 @@ typedef struct _REPARSE_DATA_BUFFER {
 #define SYMLINK_FLAG_RELATIVE 1
 #endif
 
+#ifndef __REACTOS__
 #ifndef FILE_SUPPORTS_BLOCK_REFCOUNTING
 
 typedef struct _DUPLICATE_EXTENTS_DATA {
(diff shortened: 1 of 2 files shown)


dll/shellext/shellbtrfs/shellext.h
```diff
@@ -222,6 +223,41 @@ typedef struct _FSCTL_SET_INTEGRITY_INFORMATION_BUFFER {
 #define FSCTL_SET_INTEGRITY_INFORMATION CTL_CODE(FILE_DEVICE_FILE_SYSTEM, 160, METHOD_BUFFERED, FILE_READ_DATA | FILE_WRITE_DATA)
 
 #endif
+#else // __REACTOS__
+#ifndef DUPLICATE_EXTENTS_DATA // NT6.3+ feature.
```
SergeGautherie, line 227: I am not sure about NT6.3+ comment: that is according to ReactOS, though MS docs read 'Windows Server 2016'.

learn-more:
And what problem does this fix?

SergeGautherie:
> what problem does this fix?

Currently, I would call it a code improvement, not a behavior fix.
Make it more explicit that a NT6.3+ kernel is not really needed. Allow to set 6.1 or 6.2 too.
Prevent breaking compilation, as 6.3 at least, when the upstream commit is imported.

SergeGautherie:
Looking for a reviewer.

DarkFire01:
I don't think we should be doing this to our drivers, never the less it does fix the problem when building ReactOS as NT6.0 kernel target and i probably was just going to do this same thing anyway

DarkFire01 (approved):

SergeGautherie:
Looking for a committer.

dll/shellext/shellbtrfs/shellext.h
learn-more: `#ifndef DUPLICATE_EXTENTS_DATA` cannot be used to detect the presence of the struct `DUPLICATE_EXTENTS_DATA`, so this check makes no sense.

learn-more (changes requested):

PR #5617, opened 2023-08-25
[RAPPS] Improve existing window re-use

whindsaks:
- There is a small race between `CreateMutex` and `CMainWindow` creating the window, another instance might get `NULL` in its `FindWindow`. This PR retries multiple times and if the window does not appear within a reasonable, it starts another instance.
- If the existing window is minimized, it should be restored. If the existing window is maximized, it should **not** be "normalized". `SwitchToThisWindow` does all this in addition to calling `SetForegroundWindow`.

### DIFF
--- a/base/applications/rapps/unattended.cpp
+++ b/base/applications/rapps/unattended.cpp
@@ -269,6 +269,10 @@ ParseCmdAndExecute(LPWSTR lpCmdLine, BOOL bIsFirstLaunch, int nCmdShow)
                 SwitchToThisWindow(hWindow, TRUE);
                 if (bAppwizMode)
                     PostMessage(hWindow, WM_COMMAND, ID_ACTIVATE_APPWIZ, 0);
+
+                if (hMutex)
+                    CloseHandle(hMutex);
+
                 return FALSE;
             }
         }
--- a/base/applications/rapps/unattended.cpp
+++ b/base/applications/rapps/unattended.cpp
@@ -253,15 +253,20 @@ ParseCmdAndExecute(LPWSTR lpCmdLine, BOOL bIsFirstLaunch, int nCmdShow)
         hMutex = CreateMutexW(NULL, FALSE, szWindowClass);
         if ((!hMutex) || (GetLastError() == ERROR_ALREADY_EXISTS))
         {
-            /* If already started, find its window */
-            HWND hWindow = FindWindowW(szWindowClass, NULL);
+           /* If already started, find its window */
+            HWND hWindow;
+            for (int wait = 2500, inter = 250; wait > 0; wait -= inter)
+                if ((hWindow = FindWindowW(szWindowClass, NULL)) != 0)
+                    break;
+                else
+                    Sleep(inter);
 
-            /* Activate window */
-            ShowWindow(hWindow, SW_SHOWNORMAL);
-            SetForegroundWindow(hWindow);
+            /* Activate the window in the other instance */
+            SwitchToThisWindow(hWindow, TRUE);
             if (bAppwizMode)
                 PostMessage(hWindow, WM_COMMAND, ID_ACTIVATE_APPWIZ, 0);
-            return FALSE;
+            if (hWindow)
+                return FALSE;
         }
 
         CMainWindow wnd(&db, bAppwizMode);
--- a/base/applications/rapps/unattended.cpp
+++ b/base/applications/rapps/unattended.cpp
@@ -253,20 +253,24 @@ ParseCmdAndExecute(LPWSTR lpCmdLine, BOOL bIsFirstLaunch, int nCmdShow)
         hMutex = CreateMutexW(NULL, FALSE, szWindowClass);
         if ((!hMutex) || (GetLastError() == ERROR_ALREADY_EXISTS))
         {
-           /* If already started, find its window */
+            /* If already started, find its window */
             HWND hWindow;
             for (int wait = 2500, inter = 250; wait > 0; wait -= inter)
-                if ((hWindow = FindWindowW(szWindowClass, NULL)) != 0)
+            {
+                if ((hWindow = FindWindowW(szWindowClass, NULL)) != NULL)
                     break;
                 else
                     Sleep(inter);
+            }
 
-            /* Activate the window in the other instance */
-            SwitchToThisWindow(hWindow, TRUE);
-            if (bAppwizMode)
-                PostMessage(hWindow, WM_COMMAND, ID_ACTIVATE_APPWIZ, 0);
             if (hWindow)
+            {
+                /* Activate the window in the other instance */
+                SwitchToThisWindow(hWindow, TRUE);
+                if (bAppwizMode)
+                    PostMessage(hWindow, WM_COMMAND, ID_ACTIVATE_APPWIZ, 0);
                 return FALSE;
+            }
         }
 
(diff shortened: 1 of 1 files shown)


base/applications/rapps/unattended.cpp
```diff
@@ -253,15 +253,20 @@ ParseCmdAndExecute(LPWSTR lpCmdLine, BOOL bIsFirstLaunch, int nCmdShow)
         hMutex = CreateMutexW(NULL, FALSE, szWindowClass);
         if ((!hMutex) || (GetLastError() == ERROR_ALREADY_EXISTS))
         {
-            /* If already started, find its window */
-            HWND hWindow = FindWindowW(szWindowClass, NULL);
-
-            /* Activate window */
-            ShowWindow(hWindow, SW_SHOWNORMAL);
-            SetForegroundWindow(hWindow);
+           /* If already started, find its window */
```
SergeGautherie, line 256:
```suggestion
            /* If already started, find its window */
```

base/applications/rapps/unattended.cpp
```diff
@@ -253,15 +253,20 @@ ParseCmdAndExecute(LPWSTR lpCmdLine, BOOL bIsFirstLaunch, int nCmdShow)
...
+            HWND hWindow;
+            for (int wait = 2500, inter = 250; wait > 0; wait -= inter)
+                if ((hWindow = FindWindowW(szWindowClass, NULL)) != 0)
```
SergeGautherie, line 259:
```suggestion
                if ((hWindow = FindWindowW(szWindowClass, NULL)) != NULL)
```

base/applications/rapps/unattended.cpp
```diff
@@ -253,15 +253,20 @@ ParseCmdAndExecute(LPWSTR lpCmdLine, BOOL bIsFirstLaunch, int nCmdShow)
...
         {
-            /* If already started, find its window */
-            HWND hWindow = FindWindowW(szWindowClass, NULL);
-
-            /* Activate window */
-            ShowWindow(hWindow, SW_SHOWNORMAL);
-            SetForegroundWindow(hWindow);
+           /* If already started, find its window */
+            HWND hWindow;
+            for (int wait = 2500, inter = 250; wait > 0; wait -= inter)
```
SergeGautherie, line 258:
```suggestion
            for (int wait = 2500, inter = 250; wait > 0; wait -= inter)
            {
```

base/applications/rapps/unattended.cpp
```diff
@@ -253,15 +253,20 @@ ParseCmdAndExecute(LPWSTR lpCmdLine, BOOL bIsFirstLaunch, int nCmdShow)
...
+                if ((hWindow = FindWindowW(szWindowClass, NULL)) != 0)
+                    break;
+                else
+                    Sleep(inter);
```
SergeGautherie, line 262:
```suggestion

                Sleep(inter);
            }
```

base/applications/rapps/unattended.cpp
```diff
@@ -253,15 +253,20 @@ ParseCmdAndExecute(LPWSTR lpCmdLine, BOOL bIsFirstLaunch, int nCmdShow)
...
+
+            /* Activate the window in the other instance */
+            SwitchToThisWindow(hWindow, TRUE);
             if (bAppwizMode)
                 PostMessage(hWindow, WM_COMMAND, ID_ACTIVATE_APPWIZ, 0);
```
SergeGautherie, line 267: Are these correct if `hWindow == NULL`?

whindsaks: Correct is a strong word, they will fail do nothing. The code was already like this but I suppose it could be better.

base/applications/rapps/unattended.cpp
SergeGautherie: `else` is superfluous after `break`.

base/applications/rapps/unattended.cpp
```diff
@@ -254,14 +254,23 @@ ParseCmdAndExecute(LPWSTR lpCmdLine, BOOL bIsFirstLaunch, int nCmdShow)
         if ((!hMutex) || (GetLastError() == ERROR_ALREADY_EXISTS))
         {
             /* If already started, find its window */
```
SergeGautherie, line 256: The `hMutex` check at old line 270 should be moved, or at least copied, here.
Moreover, I assume no mutex should mean to fail: otherwise, how could `ERROR_ALREADY_EXISTS` happen then?
Note: See #3200 for a similar case.

base/applications/rapps/unattended.cpp
```diff
@@ -254,14 +254,23 @@ ParseCmdAndExecute(LPWSTR lpCmdLine, BOOL bIsFirstLaunch, int nCmdShow)
...
-            HWND hWindow = FindWindowW(szWindowClass, NULL);
-
-            /* Activate window */
-            ShowWindow(hWindow, SW_SHOWNORMAL);
-            SetForegroundWindow(hWindow);
-            if (bAppwizMode)
-                PostMessage(hWindow, WM_COMMAND, ID_ACTIVATE_APPWIZ, 0);
-            return FALSE;
+            HWND hWindow;
+            for (int wait = 2500, inter = 250; wait > 0; wait -= inter)
```
JoachimHenze, line 258: Is that added for-loop also required to make it run properly on MS Windows? Or is that a blessing needed only on ReactOS to make it run reliably?

whindsaks: It has nothing to do with ROS vs Windows, it is about speed and quantum because there is a race here.

|Process A|Process B|
| ---- | ---- |
|CreateMutex OK|CreateMutex fails|
|...|FindWindow returns NULL|
|CreateWindow|...|

base/applications/rapps/unattended.cpp
whindsaks: I figured the original code was like this because the mutex is not required for the program to function and I just kept this behavior.

whindsaks:
> This PR is stale because it received no updates in the last 4 months. Without removing the stale label, or commenting on this ticket it will be closed in 2 weeks.

Go away bot

katahiromz (approved):

base/applications/rapps/unattended.cpp
SergeGautherie: > When another process creates the mutex first

Yeah, but what if the first process failed to create the mutex, for some reason?

--

To confirm, can multiple instances of RApps work fine simultaneously?

While here, note:
https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-createmutexexa
`If you are using a named mutex to limit your application to a single instance, a malicious user can create this mutex before you do and prevent your application from starting. To prevent this situation, create a randomly named mutex and store the name so that it can only be obtained by an authorized user. Alternatively, you can use a file for this purpose. To limit your application to one instance per user, create a locked file in the user's profile directory.`

whindsaks: RApps works with multiple instances for the most part. The critical part is probably the database update but that happens before the mutex anyway (this PR does not change that). The database update should probably have its own mutex.

screenshot


You cannot protect yourself from a process running as the same user at the same (or higher) privilege level. Named kernel objects are per session unless you use the Global prefix. The only scenario where a lockfile is better than a mutex is when the malicious program is running at Low Integrity Level (UAC feature).

PR #5621, opened 2023-08-27
[CALC] Use CW_USEDEFAULT16 for Dialog X

katahiromz:
## Purpose
Use default position for main dialogs.
JIRA issue: [CORE-19141](https://jira.reactos.org/browse/CORE-19141)

## Proposed changes

- Use `CW_USEDEFAULT16` (`0x8000`) for the X coordinate of `IDD_DIALOG_SCIENTIFIC`, `IDD_DIALOG_STANDARD`, and `IDD_DIALOG_CONVERSION` dialogs.

## TODO

- [x] Do tests.


binarymaster:
How does it behave before and after? (a video would be nice)

katahiromz:
BEFORE:
https://github.com/reactos/reactos/assets/2107452/e16ce819-05a9-483a-aacb-b93caaacff5a

AFTER:
(wait)

katahiromz:
Video added.

binarymaster (approved):
I can confirm this problem was also on Windows with our calc.exe and now it's fixed. You can reference this solution: https://stackoverflow.com/a/76201122/4381493 (also I grepped our source and found no previous usage of this, perhaps it could be used in more places)

PR #5632, opened 2023-08-30
[REGEDIT] Update Romanian (ro-RO) translation

AndreiMiloiu:


AndreiMiloiu:
Cc @reactos/lang-romanian

reactos-contributor (approved):
The translation looks OK, but for now none of these strings are being used. As a general advice, a quality translation is made taking into account the context in which it resides, yet here the context is not defined.

AndreiMiloiu:
So, what should I do? Isn't it a good translation? I understand nothing.

reactos-contributor:
> So, what should I do? Isn't it a good translation? I understand nothing.

We don't know how good it is (for now) as it isn't used at all yet. You may either let it be merged into master as is (without knowing much about how it will look in use), or wait until the translated resources will get to be used (and then you'll see if it may need adjustments or not).

AndreiMiloiu:
I understand now.

PR #5633, opened 2023-08-30
[USETUP] Install bootloader (MBR and VBR) by default

cbialorucki:
## Purpose

Don't prompt the user asking how they would like to install the bootloader. Lock advanced bootloader setup settings behind unattended setup. This has numerous benefits including being easier to install by end users and making it easier to test changes for developers.

## Proposed changes
- Use the `Install Bootloader (MBR and VBR)` option by default without prompting the user when installing or upgrading ROS. Windows Server 2003 also installs the bootloader by default without prompting the user during the first-stage setup. Other options are still available using unattended setup.

### DIFF
--- a/base/setup/usetup/usetup.c
+++ b/base/setup/usetup/usetup.c
@@ -3853,8 +3853,6 @@ RegistryPage(PINPUT_RECORD Ir)
     else
     {
         CONSOLE_SetStatusText(MUIGetString(STRING_DONE));
-        //if (RepairUpdateFlag)
-            //return SUCCESS_PAGE;
         return BOOT_LOADER_PAGE;
     }
 }
--- a/base/setup/usetup/usetup.c
+++ b/base/setup/usetup/usetup.c
@@ -3853,8 +3853,8 @@ RegistryPage(PINPUT_RECORD Ir)
     else
     {
         CONSOLE_SetStatusText(MUIGetString(STRING_DONE));
-        if (RepairUpdateFlag)
-            return SUCCESS_PAGE;
+        //if (RepairUpdateFlag)
+            //return SUCCESS_PAGE;
         return BOOT_LOADER_PAGE;
     }
 }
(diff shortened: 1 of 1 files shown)


cbialorucki:
> Why did reinstalling the bootloader disturb the quick "Enter+Enter+Enter,U" way of doing things?

Currently, after copying the binaries and updating the registry, it then asks for how you would like to install the bootloader. It will wait forever on that screen until you select an option. So if you leave the room, the upgrade will not complete until you sit back down, hit enter again, then wait for the OS to reboot.

If we skip the bootloader install when performing an upgrade, the installer will automatically reboot the OS after it has finished upgrading, and by the time you come back the OS will be ready to use with your new changes.

HBelusca:
Well.... Repair upgrades/updates do an update of the files, but here, if you skip the bootloader install screen, you don't give it a chance to.... update the bootloader :confused: :confused:

JoachimHenze:
Ah, I do understand now, because the bootloader question comes after the file copy, and not before it.
Why not moving that page before the file copy?

cbialorucki:
> Ah, I do understand now, because the bootloader question comes after the file copy, and not before it. Why not moving that page before the file copy?

I will experiment with this, thanks for the suggestion.

cbialorucki:
> Ah, I do understand now, because the bootloader question comes after the file copy, and not before it. Why not moving that page before the file copy?

I looked into it, we would have to make significant changes.

Why do we prompt the user at all for the bootloader? Windows doesn't. We should set `Install Bootloader (MBR and VBR)` as the default and require unattended setup for the other options.

TAN-Gaming:
> Why do we prompt the user at all about the bootloader?

The user maybe using an alternative bootloader so the setup should ask the user before overwriting it.

HBelusca:
> > Ah, I do understand now, because the bootloader question comes after the file copy, and not before it. Why not moving that page before the file copy?
> 
> I looked into it, we would have to make significant changes.

Except that this is going to be part of the soon-to-start paid job of _someone_ who will also finish the 1st-stage GUI setup... (and whom have already started working on that)

PR #5635, opened 2023-08-31
[BROWSEUI][SDK] Half-implement CShellBrowser::GetPropertyBag

katahiromz:
## Purpose

## Proposed changes

- Add `_ILIsNetworkPlace` helper function.
- Implement `CShellBrowser::GetPropertyBag` function a little by using `SHGetViewStatePropertyBag`.
- Add `SHGetPathFromIDListWrapW` prototype to `<shlwapi_undoc.h>`.

### DIFF
--- a/dll/win32/browseui/shellbrowser.cpp
+++ b/dll/win32/browseui/shellbrowser.cpp
@@ -351,7 +351,6 @@ public:
     void UpdateGotoMenu(HMENU theMenu);
     void UpdateViewMenu(HMENU theMenu);
     void LoadCabinetState();
-    static BOOL _DoesPidlRoam(LPCITEMIDLIST pidl);
 
 /*    // *** IDockingWindowFrame methods ***
     virtual HRESULT STDMETHODCALLTYPE AddToolbar(IUnknown *punkSrc, LPCWSTR pwszItem, DWORD dwAddFlags);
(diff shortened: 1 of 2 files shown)


dll/win32/browseui/shellbrowser.cpp
```diff
@@ -2387,12 +2388,33 @@ HRESULT STDMETHODCALLTYPE CShellBrowser::QueryService(REFGUID guidService, REFII
     return E_NOINTERFACE;
 }
 
+BOOL CShellBrowser::_DoesPidlRoam(LPCITEMIDLIST pidl)
+{
+    WCHAR szPath[MAX_PATH];
+    return SHGetPathFromIDListWrapW(pidl, szPath) && PathIsUNCW(szPath);
```
learn-more, line 2394: Would it make sense to create something like `_ILIsNetworkPlaces` that can be used?
We have f.e. `BOOL _ILIsBitBucket(LPCITEMIDLIST pidl)` which does exactly the same, but for the bitbucket.

That would save translating the pidl to string, and parsing that string.
@yagoulas any input on this?

katahiromz: Hmm, there is no such function named `_ILIsNetworkPlaces`.
The target is PIDL, so `IL...` functions can be used.

PR #5637, opened 2023-08-31
[I8042PRT] Do not try to flush controller's input buffer

serrox:
Do not try to flush controller's input buffer - in fact, current code will try to flush very same controller's output port that above in code, but looking to wrong status flag

As I googled about i8042 - it seems that OS has no way flush controller's input buffer. And I found that linux happy with just flushing controller's output buffer: 
https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/drivers/input/serio/i8042.c?h=v6.4.13#n263

### DIFF
--- a/drivers/input/i8042prt/readwrite.c
+++ b/drivers/input/i8042prt/readwrite.c
@@ -28,12 +28,6 @@ i8042Flush(
 		KeStallExecutionProcessor(50);
 		TRACE_(I8042PRT, "Output data flushed\n");
 	}
-
-	/* Flush input buffer */
-	while (NT_SUCCESS(i8042ReadData(DeviceExtension, KBD_IBF, &Ignore))) {
-		KeStallExecutionProcessor(50);
-		TRACE_(I8042PRT, "Input data flushed\n");
-	}
 }
 
 BOOLEAN


HBelusca:
Is this fixing an actual bug on virtual machines or hardware?

JoachimHenze (approved):
Deleting code gets my approval by default, unless the deletion can be proven to be wrong ofc. ;-)

serrox:
> Is this fixing an actual bug on virtual machines or hardware?

No, I found it while just looking around code. Reading from port using a wrong flag as data ready flag looks strange

binarymaster (changes requested):
Hi, just a few things to be corrected regarding commit author field:

image

1. [ ] Make sure your first name comes first (currently it seems you specified last name as first)
1. [x] Set your full name in [GitHub profile settings](https://github.com/settings/profile) (it should match your author name from commit)

Regards,

HBelusca:
> in fact, current code will try to flush very same controller's output port that above in code, but looking to wrong status flag

Where do you see in the code, that it's going to actually flush the controller output port?

serrox:
> Hi, just a few things to be corrected regarding commit author field:
> 
> image
> 
> 1. [ ]  Make sure your first name comes first (currently it seems you specified last name as first)
> 2. [x]  Set your full name in [GitHub profile settings](https://github.com/settings/profile) (it should match your author name from commit)
> 
> Regards,

done

binarymaster:
Reportedly there was a problem with PS/2 mouse on 86Box emulator, I wonder if this improves that somehow.

tkreuzer:
This should be tested by the "official ReactOS laptos" and other hardware before merge.

